PR #1188, opened 2018-03-08
Gossip enhance prework

rustyrussell:
I've been working on trying to test the gossip enhancements proposed, and it requires some more infrastructure around features and gossip messages.

In particular, we'll be relaying more messages for gossipd, so that needs to be enhanced: we've currently been ignoring errors when we should have been closing connections according to the spec.

### DIFF
--- a/channeld/channel.c
+++ b/channeld/channel.c
@@ -306,11 +306,16 @@ static void enqueue_peer_msg(struct peer *peer, const u8 *msg TAKES)
 static void gossip_in(struct peer *peer, const u8 *msg)
 {
 	u8 *gossip;
+	u64 gossip_index;
 
-	if (!fromwire_gossip_send_gossip(msg, msg, &peer->gossip_index, &gossip))
+	if (!fromwire_gossip_send_gossip(msg, msg, &gossip_index, &gossip))
 		status_failed(STATUS_FAIL_GOSSIP_IO,
 			      "Got bad message from gossipd: %s",
 			      tal_hex(msg, msg));
+
+	/* Zero is a special index meaning this is unindexed gossip. */
+	if (gossip_index != 0)
+		peer->gossip_index = gossip_index;
 
 	if (is_msg_for_gossipd(gossip))
 		enqueue_peer_msg(peer, gossip);
--- a/channeld/channel.c
+++ b/channeld/channel.c
@@ -306,22 +306,19 @@ static void enqueue_peer_msg(struct peer *peer, const u8 *msg TAKES)
 static void gossip_in(struct peer *peer, const u8 *msg)
 {
 	u8 *gossip;
-	u16 type;
 
-	if (!fromwire_gossip_send_gossip(msg, msg,
-					 &peer->gossip_index, &gossip))
+	if (!fromwire_gossip_send_gossip(msg, msg, &peer->gossip_index, &gossip))
 		status_failed(STATUS_FAIL_GOSSIP_IO,
 			      "Got bad message from gossipd: %s",
 			      tal_hex(msg, msg));
-	type = fromwire_peektype(gossip);
 
-	if (type == WIRE_CHANNEL_ANNOUNCEMENT || type == WIRE_CHANNEL_UPDATE ||
-	    type == WIRE_NODE_ANNOUNCEMENT)
+	if (is_msg_for_gossipd(gossip))
 		enqueue_peer_msg(peer, gossip);
 	else
 		status_failed(STATUS_FAIL_GOSSIP_IO,
 			      "Got bad message type %s from gossipd: %s",
-			      wire_type_name(type), tal_hex(msg, msg));
+			      wire_type_name(fromwire_peektype(gossip)),
+			      tal_hex(msg, msg));
 }
 
 /* Send a temporary `channel_announcement` and `channel_update`. These
--- a/channeld/channel.c
+++ b/channeld/channel.c
@@ -314,7 +314,12 @@ static void gossip_in(struct peer *peer, const u8 *msg)
 
 	if (is_msg_for_gossipd(gossip))
 		enqueue_peer_msg(peer, gossip);
-	else
+	else if (fromwire_peektype(gossip) == WIRE_ERROR) {
+		struct channel_id channel_id;
+		char *what = sanitize_error(msg, msg, &channel_id);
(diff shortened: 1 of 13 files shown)


channeld/channel.c
```diff
@@ -306,12 +306,17 @@ static void enqueue_peer_msg(struct peer *peer, const u8 *msg TAKES)
...
+	u64 gossip_index;
 
-	if (!fromwire_gossip_send_gossip(msg, msg, &peer->gossip_index, &gossip))
+	if (!fromwire_gossip_send_gossip(msg, msg, &gossip_index, &gossip))
 		status_failed(STATUS_FAIL_GOSSIP_IO,
 			      "Got bad message from gossipd: %s",
 			      tal_hex(msg, msg));
 
+	/* Zero is a special index meaning this is unindexed gossip. */
+	if (gossip_index != 0)
```
cdecker: I assume this is used for the range signalling for a later commit?

gossipd/gossip.c
```diff
@@ -291,6 +291,22 @@ static void reached_peer(struct daemon *daemon, const struct pubkey *id,
...
 }
 
+static void queue_peer_msg(struct peer *peer, const u8 *msg TAKES)
+{
+	const u8 *send;
+
+	if (peer->local) {
+		msg_enqueue(&peer->local->peer_out, msg);
+		return;
+	}
```
cdecker: I try to personally avoid tearing apart an if-else statement into if-return, especially when both branches are equally likely. It just helps with readability.

gossipd/routing.c
```diff
@@ -566,8 +611,12 @@ const struct short_channel_id *handle_channel_announcement(
 					   &pending->node_id_2,
 					   &pending->bitcoin_key_1,
 					   &pending->bitcoin_key_2)) {
+		err = towire_errorfmt(rstate, NULL,
+				      "Malformed channel_announcement %s",
+				      tal_hex(pending, pending->announce));
 		tal_free(pending);
-		return NULL;
+		*scid = NULL;
```
cdecker: Having these null-sets plastered all over the place seems dangerous, wouldn't it not be better to just set NULL once on top? I know you like `valgrind` telling you about unset things, but there is no extra information to be had here, other than "oops forgot to set".

cdecker: I'd even go as far and say the caller should set whatever special value he is expecting for unset.

gossipd/gossip.c
```diff
@@ -1369,7 +1371,11 @@ static void gossip_send_keepalive_update(struct routing_state *rstate,
 	status_trace("Sending keepalive channel_update for %s",
 		     type_to_string(tmpctx, struct short_channel_id, &scid));
 
-	handle_channel_update(rstate, update);
+	err = handle_channel_update(rstate, update);
```
cdecker: This should probably be fatal, it's an update we generate so if we don't accept it, that's on us.

gossipd/gossip.c
```diff
@@ -1927,7 +1934,11 @@ static struct io_plan *handle_disable_channel(struct io_conn *conn,
 			      strerror(errno));
 	}
 
-	handle_channel_update(daemon->rstate, msg);
+	err = handle_channel_update(daemon->rstate, msg);
```
cdecker: Same here, if we can't be trusted to disable tell ourselves about a channel we disable, then that's on us.

gossipd/routing.c
```diff
@@ -793,10 +805,11 @@ bool handle_pending_cannouncement(struct routing_state *rstate,
 		pubkey_eq(&pending->node_id_2, &rstate->local_id);
 
 	/* Did we have an update waiting?  If so, apply now. */
+	/* FIXME: We don't remember who sent us updates, so we can't error them */
 	if (pending->updates[0])
-		handle_channel_update(rstate, pending->updates[0]);
+		tal_free(handle_channel_update(rstate, pending->updates[0]));
```
cdecker: We should log them maybe?

cdecker:
@ZmnSCPxj `openingd` and `closingd` are very constrained as to what messages are allowed during the opening and closing flows, so they do not handle gossip messages since that'd interfere with the flows. They are shortlived daemons so there is little use in plumbing in the gossip machinery.

ZmnSCPxj:
@cdecker my poor knowledge is that nothing in particular prevents a peer from sending gossip messages during opening or closing.  For example if I make a custom client that performs `open_channel` but then forgets the flow, ignores `accept_channel`, and never sends `funding_created`, but starts sending gossip messages, then that is "valid" under the protocol, especially if my custom client does not particularly want an open channel anyway.  Or is there something in the BOLT spec which prevents gossip during opening and closing flows?

cdecker:
You are of course right, but we currently just drop them I think.

rustyrussell:
Both those daemons will relay gossip they receive from the peer, *but* never relay from gossipd to the peer.  ie. no outgoing gossip.

gossipd/gossip.c
rustyrussell: I'm generally trying to cut down on fatal errors, especially now we have status_broken().  But perhaps I'm premature: will change to fatal().

gossipd/routing.c
rustyrussell: I was aiming for minimal change, but I think you're right.  Making the caller set it is a bit too far: this is effectively a trinary return: not found, found, malformed.  My instincts are to try to split the malformed test off into a separate function, but that's awkward here, hence the weirdness.

gossipd/gossip.c
rustyrussell: Agreed.

gossipd/routing.c
rustyrussell: Actually, I turned them into gotos, which is even clearer IMHO.

gossipd/gossip.c
rustyrussell: Actually, it *is* fatal (when gossipd dies, main daemon dies):

```
	if (err)
			      "rejected keepalive channel_update: %s",
			      tal_hex(trc, err));
```

gossipd/routing.c
rustyrussell: Yes, good idea.

cdecker:
Excellent work :+1:

ACK f0fc165

PR #1195, opened 2018-03-10
rewrite contrib/lightning-pay in Python, support bolt11

### LINKED ISSUE #718: Update contrib/lightning-pay to accept bolt11s
Following #633:

1. Still use getroute, sendpay to prompt routing fees
2. Parse id, amount and payment hash from decodepay and pass to getroute/sendpay

I believe bolt11s have route hints so we can use that as well?

izgzhen:
fixes #718


contrib/pylightning/lightning-pay
```diff
@@ -0,0 +1,48 @@
...
+else:
+    amount = args.amount_in_milli_satoshi
+    id_ = args.destination_id
+    payment_hash = args.payment_hash
+
+route = l.getroute(id_, amount, 1)
+
+fee = route["route"][0]["msatoshi"] - amount
+
+reply = input("Paying fee %s on amount %s (%.3f). Send [Y/n]?" % (fee, amount, fee / amount * 100.0))
```
ZmnSCPxj: The `%.3f` is a percentage right? Should it not be `%.3f%%` to print the percent symbol also?

izgzhen: sure, fixed!

cdecker:
Sorry for the delay reviewing this. I was wondering why `flake8` isn't complaining about some things, and it turns out that it's due to the file-ending not being `*.py`. Anyway, here's some warnings:
```
contrib/pylightning/lightning-pay:4:1: F401 'random' imported but unused
contrib/pylightning/lightning-pay:16:1: E302 expected 2 blank lines, found 1
contrib/pylightning/lightning-pay:22:1: E305 expected 2 blank lines after class or function definition, found 1
contrib/pylightning/lightning-pay:23:1: E741 ambiguous variable name 'l'
```

I'm also unsure why this is now in the `pylightning` subdirectory. It makes use of it, but isn't really part of the library. So I suggest moving it up one level and modifying the `check-python` target to check this file as well.

izgzhen:
> I'm also unsure why this is now in the pylightning subdirectory. It makes use of it, but isn't really part of the library

If it is put in the current position, I can import `pylightning`'s module `lightning` as if it is installed in pip. I wonder how would user use this script in future? If as a separate tool, then we might consider installing this script to `bin` in PIP.

cdecker:
If we want to distribute the tools as part of pylightning then we need to mark it as an executable script in `setup.py`. That way we can use `pip install pylightning`, `python setup.py install` or `python setup.py develop` to make it available to users. It should not be sitting in the `pylightning` directory solely because of import paths.

izgzhen:
fixed, thanks! @cdecker

cdecker:
This is shaping up nicely, thanks for the move. One minor thing that would improve the ergonomics of this is simplifying the arguments. Currently almost everybody will need to specify `--bolt11` since that is the way we hope most people will use payments. How about we reorder the arguments to be `destination`, `amount`, `payment_hash`, drop the `--bolt11` flag and instead check whether destination starts with `ln`. If it does we have a bolt11 style request, otherwise we need the 3 args.

In its current form it doesn't check the presence of any arguments, which can result in a call to `getroute` and subsequent error that is rather surprising, I think making the first argument mandatory and showing the help text otherwise is better.

ZmnSCPxj:
Quick question, is this `lightning-pay` intended to be installed in `$(bindir)`?  If so, probably merits a manpage in `doc/`, but notice we use `man 7 lightning-pay` to refer to the RPC command `pay`, which could be confusing...

izgzhen:
> is this lightning-pay intended to be installed in $(bindir)?

I imagine that the python package is a separate thing from the C package. The user might install one thing but not another.

izgzhen:
@cdecker thanks for the comment! Let me try to revise it a bit

ZmnSCPxj:
>I imagine that the python package is a separate thing from the C package. The user might install one thing but not another.

Okay, I shall leave Pythonistas to their land then.  Although the Python package would have to depend on the C package.

cdecker:
ACK f41c9c2

PR #1196, opened 2018-03-10
help: listnodes and listchannels,have optional arg

isghe:
that permits to list one single node or channel.


ZmnSCPxj:
ACK 91e23df9a3f38a3ba9d108e769a4ce13db6ea93c

PR #1197, opened 2018-03-10
Append to crash log instead of creating new one

jBarz:
Multiple crashes accumulate in case you have some kind
of auto-restart for daemons.

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -510,7 +510,7 @@ static void log_crash(int sig)
 
 		/* We expect to be in config dir. */
 		logfile = "crash.log";
-		fd = open(logfile, O_WRONLY|O_CREAT, 0600);
+		fd = open(logfile, O_WRONLY|O_CREAT|O_APPEND, 0600);
 		if (fd < 0) {
 			logfile = "/tmp/lightning-crash.log";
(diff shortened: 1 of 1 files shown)


lightningd/log.c
```diff
@@ -557,6 +557,9 @@ void log_dump_to_file(int fd, const struct log_book *lr)
 	struct log_data data;
 	time_t start;
 
+	sprintf(buf, "%s\n", "Start of new crash log");
```
ZmnSCPxj: Why not `strcpy(buf, "Start of new crash log\n");`?  Or even `write_all(fd, "Start of new crash log\n", strlen("Start of new crash log\n"));`, a technique used a few lines below?

jBarz: Thanks, I changed it to the later.

jBarz: Also added another optimization below where we unnecessarily calculate strlen

ZmnSCPxj:
ACK 9c71ed6cee624328a10267b572bfbc4dc5c3f9c5

cdecker:
I'd eventually prefer to have a rotation in place to have numbered crashes and not needing to search for the end-of-crash markers, but this fixes the immediate problem, so:

ACK 9c71ed6

Issue #1201, opened 2018-03-11
ONCHAIN state, channel stuck for +300 confirmations


fixone:
"version": "v0.5.2-2016-11-21-2262-gd363a68",

I have two channels which seem to be stuck in ONCHAIN state, although the unilateral close tx has 382 confirmations (at this moment)
```
	{ "id" : "027e92b4fa1988ae5b0d44b87da80ab8875312fdae6631a8896851bc41857e66aa", "connected" : false, "channels" :
			[
				{ "state" : "ONCHAIN", "owner" : "lightning_onchaind", "short_channel_id" : "1284270:177:0", "funding_txid" : "c590a2cde116c74cd04facd2364ee6b8b9b667b92f56cc52bec5d1b5c724bb87", "msatoshi_to_us" : 999646994, "msatoshi_total" : 1000000000, "dust_limit_satoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minimum_msat" : 0, "to_self_delay" : 6, "max_accepted_htlcs" : 483, "status" :
					[ "ONCHAIN:Tracking our own unilateral close", "ONCHAIN:1 outputs unresolved: waiting confirmation that we spent DELAYED_OUTPUT_TO_US (e04b3f41191df92d800390d34f0717ae18bf2eeaa2ad606ca2ec60ee3eee7acc:0) using OUR_DELAYED_RETURN_TO_WALLET" ] } ] }
```
there's a second channel with 285confirms
```
	{ "id" : "03172377a939b21aadab64943ac82fc54a6600db48812dec9046c78abd78f5a0a2", "connected" : false, "channels" :
			[
				{ "state" : "ONCHAIN", "owner" : "lightning_onchaind", "short_channel_id" : "1287428:7:0", "funding_txid" : "73460608c906b1b0dcd0b969443e712073bba96d3930025a1ee6b5c227e3edb4", "msatoshi_to_us" : 980000000, "msatoshi_total" : 980000000, "dust_limit_satoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minimum_msat" : 0, "to_self_delay" : 6, "max_accepted_htlcs" : 483, "status" :
[...]

PR #1202, opened 2018-03-12
peer_control: Print both `to_self_delay` in listpeers.

ZmnSCPxj:
Confusingly we only printed the `to_self_delay` we imposed on other side, not what the other side imposed on us.  Print both, attributing them correctly (our configuration is `their_to_self_delay`, their configuration is `our_to_self_delay`).


ZmnSCPxj:
Should help debug #1201

cdecker:
I always get confused about who is imposing what to whom, but I think you are right, it matches my reading of [the spec](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#the-open_channel-message)

ACK 223025f

ZmnSCPxj:
I believe `to_self_delay` is the only setting set by one side that is imposed on the other side.

The naming "to self" is unfortunate and probably accounts for much of the confusion.

PR #1206, opened 2018-03-13
Gossipd more cleanups

rustyrussell:
Built on #1188 (new work is 'gossipd: use u64 for broadcast index.' and beyond).

Firstly this simplifies the broadcast system: we now always access by index so it can be radically simplified.   This also fixes a latent bug where we didn't clean up gossip for pruned nodes/channels.

It then moves on to the features system, importantly fixing the other spam bug where we weren't honoring the `initial_routing_sync` feature, even though it's trivial.

Other cleanups are prep for the work to support enhanced gossip querying as per pending spec suggestions.


ZmnSCPxj:
ACK 55963f05c795e80be1ff6f4cd179d90ec5c0f785

Will continue after that commit, please wait...

cdecker:
#1188 got merged, but I'll wait rebasing this until @ZmnSCPxj has finished reviewing, so we don't lose his progress :wink:

cdecker:
ACK a51e68d

Will rebase and merge once we get the ok from @ZmnSCPxj

ZmnSCPxj:
ACK a51e68d8790c3359f00d8da7438829899761e918

rustyrussell:
OK, will rebase and merge.  Thanks!

PR #1214, opened 2018-03-14
FEEDBACK WANTED: Global tmpctx and style

rustyrussell:
This is a bit of an experiment, but I think it worked.  I cleaned up gossipd a bit, but I leaked a tmpctx: I looked for other leaks and decided it was time to convert that into a global.

I also wrote a [Style guide](doc/STYLE.md) which I'd love additions to...


ZmnSCPxj:
Style guide link in original comment is broken.  Will check through commits....

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Prefer 80 Columns
+
+We have to stop somewhere.  The two tools here are extracting
+deeply-indented code into their own functions, and use of short-cuts
+using early returns or continues, eg:
+
+```C
+for (i = start; i != end; i++) {
+	if (i->something)
```
ZmnSCPxj: Indentation seems lacking here?

common/utils.h
```diff
@@ -16,12 +16,16 @@ char *tal_hex(const tal_t *ctx, const tal_t *data);
...
-tal_t *tal_tmpctx_(const tal_t *ctx, const char *file, unsigned int line);
-#define tal_tmpctx(ctx)							\
-	tal_tmpctx_((ctx), __FILE__, __LINE__)
+/* FIXME: Remove in favor of global */
+#define tal_tmpctx(ctx) tal((ctx), char)
 
-/* Return non-NULL if any tmpctx still allocated. */
-const char *tmpctx_any(void);
+/* Global temporary convenience context: freed in io loop core. */
+extern const tal_t *tmpctx;
```
ZmnSCPxj: In principle, I like this, despite my general horror of static-lifetime objects.

A concern to consider is that the amount of floating, allocated memory that is in-principle "dead" (and which will eventually get cleaned but is not cleaned *now*) increases.  This may be of concern to limited-memory devices such as ARM targets.

I wonder how much of garbage collection we end up implementing.  Currently we effectively make `tmpctx` a singular arena whose contained objects are periodically cleaned, almost like a garbage-collection nursery.  Allocating off `tmpctx` and then `tal_steal`ing them off into a more permanent object parent is almost like garbage-collection promotion.  And so on.

ZmnSCPxj:
Comment on the global `tmpctx` seems to have gotten covered, so:

In principle, I like this, despite my general horror of static-lifetime objects.

A concern to consider is that the amount of floating, allocated memory that is in-principle "dead" (and which will eventually get cleaned but is not cleaned *now*) increases.  This may be of concern to limited-memory devices such as ARM targets.

I wonder how much of garbage collection we end up implementing.  Currently we effectively make `tmpctx` a singular arena whose contained objects are periodically cleaned, almost like a garbage-collection nursery.  Allocating off `tmpctx` and then `tal_steal`ing them off into a more permanent object parent is almost like garbage-collection promotion.  And so on.

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+		continue;
+
+	do_something(i);
+}
+```
+
+## Prefer Simple Statements
+
+Notice the statement above uses separate tests, rather than combining
+them.  I prefer to only combine conditionals which are fundamentally
```
cdecker: "I" as in Rusty -> "we" the c-lightning devs?

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+There's a convenient temporary tal context which gets cleaned
+regularly: you should use this for throwaways rather than (as you'll
+see some of my older code do!) grabbing some passing object to hang
+your temporaries off!
+
+## Enums and Switch Statements
+
+If you handle various enumerated values in a `switch`, don't use
+`default:` but instead mention every enumeration case-by-case.  That
+way when a new enumeration case is added, `gcc` will warn that you
```
cdecker: The same goes for `clang` btw

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+	if (bar == foo)
+		is_foo = true;
+	else
+		is_foo = false;
+
+	...
+	if (is_foo)...
+```
+
+This way gcc will warn you if you have one path which doesn't set the
```
cdecker: gcc -> `gcc`

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+```C
+	msg = towire_shutdown(NULL, &peer->channel_id, peer->final_scriptpubkey);
+	enqueue_peer_msg(peer, take(msg));
+```
+
+## Use of `tmpctx`
+
+There's a convenient temporary tal context which gets cleaned
+regularly: you should use this for throwaways rather than (as you'll
+see some of my older code do!) grabbing some passing object to hang
```
practicalswift: Nit: "our" instead of my? :-)

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+initialize only the fields you expect to be used.
+
+Similarly, you can use `memcheck(mem, len)` to explicity assert that
+memory should have been initialized, rather than having valgrind
+trigger later.  We use this when placing things on queues, for example.
+
+## Use of static and const
+
+Everything should be declared static and const by default.  Note that
+`tal_free()` can free a const pointer (also, that it returns NULL, for
```
practicalswift: Nit: `NULL`?

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Use of FIXME
+
+There are two cases in which you should use a `/* FIXME: */` comment:
+one is where an optimization is possible but it's not clear that it's
+yet worthwhile, and the second one is to note an ugly corner case
+which could be improved (and may be in a following patch).
+
+There are always compromises in code: eventually it needs to ship.
+FIXME are grep-fodder for yourself and others, as well as useful
```
practicalswift: Nit: FIXMEs?

doc/STYLE.md
```diff
@@ -0,0 +1,160 @@
...
+perhaps consider your fancy structure, but don't implement a generic
+thing.  Use `/* FIXME: ...*/` to salve your conscience.
+
+## Keep Your Patches Reviewable
+
+Try to make a single change at a time.  It's tempting to do "drive-by"
+fixes as you see other things, and a minimal amount is unavoidable, but
+you can end up shaving infinite yaks.  This is a good time to drop a 
+`/* FIXME: ...*/` comment and move on.
+
```
practicalswift: Nit: Trailing newline.

doc/STYLE.md
rustyrussell: Didn't want to speak for everyone, but will change since I take comments here as broadly approving :)

doc/STYLE.md
rustyrussell: OK, will generalize.

doc/STYLE.md
rustyrussell: Yes, fixed.

doc/STYLE.md
rustyrussell: Indeed, fixed.

rustyrussell:
OK, pushed a new one with some fixes, the first of which makes memleak detection not crash on a leak.

I applied all the STYLE feedback.

doc/STYLE.md
```diff
@@ -0,0 +1,159 @@
+# Care And Feeding of Your Fellow Coders
```
ZmnSCPxj: Maybe link also from `HACKING.md` to `STYLE.md`?

ZmnSCPxj:
ACK 1e502a69ff9c3d30cda46f262cdcdfa009b3ff01

doc/STYLE.md
rustyrussell: Thanks, done.

ZmnSCPxj: `HACKING.md` gets installed `$(docdir)`, usually `/usr/local/share/doc/c-lightning`.  If `HACKING.md` refers to `STYLE.md` perhaps it should be added to `DOC_DATA` in the main `Makefile`

Issue #1088, opened 2018-02-24
Better feedback when pay takes long


Sjors:
The past few days I'm having difficulty making payments, with the `pay` command just hanging there. Aborting it and trying again results in "Payment is still in progress". Log entries aren't super clear (to me) either. It would be nice to get more details.

One approach could be that the pay command has a `verbose` argument that shows what it's doing.

Another would be a `paymentlog` command, which prints whatever log information is available for a given payment.

In particular I'd like to see some sort of table which lists the steps that are in progress, i.e. which nodes it's waiting on for what, etc.

ZmnSCPxj:
The way `lightning-cli` is implemented is as a simple JSON-RPC wrapper, meaning we cannot interact with the caller of `lightning-cli`.  The only way to print something is to return a definite failure or a definite success.

What I intend to do is to create `startpay` and `statuspay` and `waitpay`.  `startpay` just starts the payment and returns immediately.  `statuspay` polls what payment is doing.  `waitpay` stalls until the payment definitely succeeds or definitely fails.

>In particular I'd like to see some sort of table which lists the steps that are in progress, i.e. which nodes it's waiting on for what, etc.

The only node it can wait on is the first direct peer it initially sent to.  The details of where the payment has reached, if some peer other than the first direct peer died or disconnected, is not known by the current node (and might not be known by any of the other nodes on the route either).  In short: the only thing we can report is whether it succeeded, it failed, or we are still waiting for the payment to complete from the first direct peer, which is the only node it talks to.

ZmnSCPxj:
Fleshing things out:

    $ lightning-cli startpay ${BOLT11}
    { pay_id: 123456 }
    $ lightning-cli statuspay 123456
    { completed: false, faillog: [], status: "Getting Route", getroute_tries: 0, sendpay_tries: 0 }
    $ lightning-cli statuspay 123456
    { completed: false, faillog: [], status: "Attempting payment", getroute_tries: 1, sendpay_tries: 0, route: [{id: "0211111111111", channel: "111:11:1", msatoshi: 1000, delay: 42}] }
    $ lightning-cli statuspay 123456
    { completed: false, faillog: ["WIRE_TEMPORARY_CHANNEL_FAILURE at channel 111:11:1"], status: "Getting Route", getroute_tries: 1, sendpay_tries: 1 }
    $ lightning-cli statuspay 123456
[...]

PR #1217, opened 2018-03-14
Change `sendpay` result to be nearer to `listpayments`

ZmnSCPxj:
Part of fix for #1088.

A note on `json_add_payment_fields`: I intend to reuse this function for the modification for `pay` output.  My intent is that `pay` displays the fields for the payment, but also adds some fields --- `getroute_tries`, `sendpay_tries`, `route`, `failures` --- indicating detailed information about the payment. Thus, the reason why `json_add_payment_fields` provides the payment details but does not open and close an object.

### DIFF
--- a/doc/lightning-waitsendpay.7.txt
+++ b/doc/lightning-waitsendpay.7.txt
@@ -8,7 +8,7 @@ lightning-waitsendpay - Protocol for sending a payment via a route.
 
 SYNOPSIS
 --------
-*waitsendpay* 'hash' ['timeout']
+*waitsendpay* 'payment_hash' ['timeout']
 
 DESCRIPTION
 -----------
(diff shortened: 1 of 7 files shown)


doc/lightning-waitsendpay.7.txt
```diff
@@ -8,7 +8,7 @@ lightning-waitsendpay - Protocol for sending a payment via a route.
```
rustyrussell: Hmm, just noticed this title line is wrong: cut & paste bug?

ZmnSCPxj: Well, it is part of the protocol for paying via a route.  You perform `sendpay`, then when convenient you perform a `waitsendpay` to monitor the status of the payment.

PR #1218, opened 2018-03-15
Make `pay` return more information, also have a timeout

### LINKED ISSUE #1088: Better feedback when pay takes long

ZmnSCPxj:
Fixes #1088 

Ping @Sjors 

Builds on top of #1217


Sjors:
Timeout is slightly confusing term because it suggests the payment itself times out.

ZmnSCPxj:
`retry_until`? `stop_retrying_after`?

lightningd/payalgo.c
```diff
@@ -15,6 +16,107 @@
...
+	/* The raw onion reply, if TYPE_UNPARSEABLE_ONION, a
+	 * tal_arr whose parent is this struct */
+	const u8 *onionreply;
+	/* The routing failure, if TYPE_PAYMENT_REPLY, a tal
+	 * object whose parent is this struct */
+	struct routing_failure *routing_failure;
+};
+
+/* FIXME: move json_add_route_hop and json_add_route to
+ * common/json.c, share code with getroute */
```
rustyrussell: Use lightningd/jsonrpc.h where we put the other lightningd-only JSON helpers.  Or, better, refactor those into lightningd/json.[ch].

lightningd/payalgo.c
```diff
@@ -41,14 +143,73 @@ struct pay {
...
+	f->route = tal_steal(f, pay->route);
+	pay->route = NULL;
+}
+
 static void
-json_pay_success(struct command *cmd,
+json_pay_success(struct pay *pay,
 		 const struct preimage *payment_preimage,
 		 unsigned int getroute_tries,
 		 unsigned int sendpay_tries)
```
rustyrussell: I assume getroute_tries and sendpay_tries are the same as in the `struct pay` so these can be removed and maybe further unified with the json_pay_failure case?

lightningd/payalgo.c
```diff
@@ -15,6 +16,107 @@
...
 #include <sodium/randombytes.h>
 
+/* Record of failures. */
+enum pay_failure_type {
+	FAIL_UNPARSEABLE_ONION,
+	FAIL_PAYMENT_REPLY
+};
+/* A payment failure. */
+struct pay_failure {
+	/* Part of a list of failures */
```
rustyrussell: Generally prefer this comment to refer to the parent type, i.e. "Off struct pay_command pay_failures".

doc/lightning-pay.7.txt
```diff
@@ -30,6 +30,9 @@ The response will occur when the payment fails or succeeds.  Once a
 payment has succeeded, calls to *pay* with the same 'bolt11' will
 succeed immediately.
 
+If the command times out while an outgoing payment attempt is ongoing,
+you may call *pay* again to continue monitoring the payment attempt.
+
```
rustyrussell: timeout default value needs to be documented.  I'd probably choose 30 rather than 20: internet is pretty slow in Australia :)

rustyrussell:
The timeout is unlikely to be reached, but we can call it `retry_for` and perhaps increase it to 60 seconds?  I think most people would prefer slow to failure, and for the *really* slow cases this has no effect (payment stuck).

ZmnSCPxj:
Temporary channel bans (from `WIRE_TEMPORARY_CHANNEL_FAILURE` et al) are currently 20 seconds, which is why my initial timeout/"`retry_for`" default was only 20 seconds.  Should we increase the channel ban also?

lightningd/payalgo.c
ZmnSCPxj: You are correct, that should be fixed too.

lightningd/payalgo.c
ZmnSCPxj: okay.

ZmnSCPxj:
Rebaased, increased time limit before stopping auto-retry.

ZmnSCPxj:
@Sjors `retry_for` was now used.

Issue #1221, opened 2018-03-15
BINTOPKGLIBEXECDIR - `tools/rel.sh` uses GNUisms in `sed` script


jsarenik:
### Issue and Steps to Reproduce
This is a quick note that `make check` actually requires `make install` to run before it in order to do its work. Otherwise there is following error:

```
make check
...
make[2]: Leaving directory '/home/jasan/src/lightning'
testinstall/usr/local/bin/lightningd --test-daemons-only --lightning-dir=testinstall
lightningd: Could not run /usr/local/libexec/c-lightning/lightning_channeld: No such file or directory
make[1]: *** [Makefile:436: installcheck] Error 1
make[1]: Leaving directory '/home/jasan/src/lightning'
make: *** [Makefile:205: check] Error 2
```

Which IMHO points to file `lightningd/lightningd.c`, [line 183](https://github.com/ElementsProject/lightning/blob/master/lightningd/lightningd.c#L183) - the line with `BINTOPKGLIBEXECDIR`.

ZmnSCPxj:
That is very strange, since it passes in CI, which does not perform install before check.

Note that check itself performs an install check where it checks that even if installed in a nonstandard location, it can still run. Cam you provide log of `make check` run?

jsarenik:
I see your point and now I am investigating further. Will be back with full log if it makes sense, but so far I have two different results on two machines. Tracking down the differences between them...

ZmnSCPxj:
When compiling, what is the `-DBINTOPKGLIBEXECDIR=` value?  Also, can you execute this command in the lightningd tree?

    tools/rel.sh /usr/local/bin /usr/local/libexec/c-lightning

jsarenik:
Sorry, not answering your question as I want to show one intermediate result now:

The compilation procudure is completely the same on both systems. Both systems are musl-based and up-to-date Alpine Linuxes.

On the system where it is failing:

```
$ strings lightningd/lightningd | grep libexec
/../../..//usr/local/libexec/c-lightning
lightningd/lightningd.c:184:pkglibexecdir
pkglibexecdir
find_my_pkglibexec_path
find_my_pkglibexec_path
```

On the system which passes `make check`:

```
$ strings lightningd/lightningd | grep libexec
../libexec/c-lightning
lightningd/lightningd.c:184:pkglibexecdir
pkglibexecdir
find_my_pkglibexec_path
find_my_pkglibexec_path
```

jsarenik:
[...]

PR #1222, opened 2018-03-15
tools/rel.sh: Make POSIXLY compatible `sed` script.

ZmnSCPxj:
Fixes: #1221

We were using `\x0` to match NUL chars in the input (on the
assumption that NUL chars are "impossible" for decent LFS-compliant
systems).
However `\x0` is a GNUism.
Use the `\n` and the newline character, which is supported by (most)
POSIX sed.

### DIFF
--- a/tools/rel.sh
+++ b/tools/rel.sh
@@ -2,6 +2,6 @@
 
 from=${1}
 to=${2}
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
A spurious CI failure in code not touched by this PR led me to #1223. Specifically `test_blockchaintrack`. This is not reliably caught since block hashes are random and rerunning `test_blockchaintrack` would sometimes spuriously pass.  However fails are always rerun while passes move on to the next test, masking this problem.

Rerunning the spurious CI failure...

PR #1224, opened 2018-03-16
Fix for crash when onchaind hits a deleted payment.

rustyrussell:
I also note that we *never* delete HTLCs, though we can (and should!) once they're in the final revoked state.

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -431,6 +431,17 @@ void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
 					 &hout->payment_hash);
 
 #ifdef COMPAT_V052
+	/* Prior to "pay: delete HTLC when we delete payment." we would
+	 * delete a payment on retry, but leave the HTLC. */
+	if (!payment) {
+		log_unusual(hout->key.channel->log,
+			    "No payment for %s:"
+			    " was this an old database?",
+			    type_to_string(tmpctx, struct sha256,
+					   &hout->payment_hash));
+		return;
+	}
+
 	/* FIXME: Prior to 299b280f7, we didn't put route_nodes and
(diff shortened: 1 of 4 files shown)


lightningd/pay.c
```diff
@@ -411,6 +411,17 @@ void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
 					 &hout->payment_hash);
 
 #ifdef COMPAT_V052
+	/* Prior to "pay: delete HTLC when we delete payment." we would
+	 * delete a payment on retry, but leave the HTLC. */
+	if (!payment) {
+		log_unusual(hout->key.channel->log,
+			    "No payment for %s:"
+			    " was this an old database?",
+			    type_to_string(ltmp, struct sha256,
```
ZmnSCPxj: `s/ltmp/tmpctx/`, ha, obviously I did not read things closely enough.

rustyrussell: Yep, rebase without compile check was a bad idea!  Done now....

ZmnSCPxj:
ACK 006b586372360f3c551ec679cc44b3499838692d

wallet/wallet.c
```diff
@@ -1468,6 +1468,24 @@ struct htlc_stub *wallet_htlc_stubs(const tal_t *ctx, struct wallet *wallet,
...
+void wallet_local_htlc_out_delete(struct wallet *wallet,
+				  struct channel *chan,
+				  const struct sha256 *payment_hash)
+{
+	sqlite3_stmt *stmt;
+
+	stmt = db_prepare(wallet->db,
+			  "DELETE FROM channel_htlcs"
+			  " WHERE direction = ?"
+			  " AND origin_htlc_id = ?"
```
ZmnSCPxj: Should just be `origin_htlc`, not `origin_htlc_id`

Issue #1229, opened 2018-03-16
"Funding transaction spent" shouldn't be a connection-level error


pm47:
The `error` message for "Funding transaction spent" has `channel_id`=0 which means that all channels associated to the node will get closed. I think this should be a channel-level error.

Unfortunately I can't tell exactly what version of c-lightning does that or if it has been fixed since.

ZmnSCPxj:
It has not been fixed: there is code that always uses "all channels" regardless. https://github.com/ElementsProject/lightning/blob/4ffd5e213b9099a676cf417c2a00d940cedd7e80/lightningd/channel.c#L307-L317

Note that the above code already has an in-memory `struct channel` as argument it could have referred to in order to get the correct channel id, but it does not use it and instead always fails all.

Arguably the code above should be using `towire_errorfmt` which will use the channel id from the given `struct channel`

pm47:
> Arguably the code above should be using towire_errorfmt which will use the channel id from the given struct channel

Surely it doesn't really matter as long as c-lightning doesn't support channel multiplexing over a single connection, but I think the "spririt" of the BOLT 2 is that `error` messages should still be as specific as possible:

`The channel is referred to by channel_id, unless channel_id is 0 (i.e. all bytes are 0), in which case it refers to all channels.`  ([BOLT 2](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#requirements-1))

ZmnSCPxj:
No, but I believe there are vague plans to eventually allow multiple channels per peer eventually, as otherwise some changes (moving channel-related peer state into a `channels` array in JSON-RPC, for example) that would make no sense without such plans.  So better to make `error` specific to channel.

PR #1230, opened 2018-03-16
channel_fail_permanent: Use a channel-level error, not an all-channels error.

ZmnSCPxj:
Fixes: #1229

The principle is that `error` should really be as specific as possible.  We also have vague plans to maybe, might be, have multiple channels per peer, in some possibility that may or may not be 0.

### DIFF
--- a/lightningd/channel.c
+++ b/lightningd/channel.c
@@ -1,6 +1,7 @@
 #include <bitcoin/script.h>
 #include <ccan/crypto/hkdf_sha256/hkdf_sha256.h>
 #include <ccan/tal/str/str.h>
+#include <common/wire_error.h>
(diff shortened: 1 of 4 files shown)


cdecker:
Excellent fix, though we also need to change some of the tests that were looking for fail-all:

 - `test_permfail`
 - `test_reconnect_normal`

The former is just a string search failure, the latter needs some more digging.

ZmnSCPxj:
Have `fundchannel` and `listpeers` return the non-short channel ID, so that the test can find out what exact channel ID to match in the status log of the channel.  Please review again!

cdecker:
Great @ZmnSCPxj :+1: 

ACK 2b25a3a

PR #1240, opened 2018-03-18
Store gossip messages on disk

### LINKED ISSUE #247: Persist channel_announcement messages during restart
This was pointed out in #237: if we restart all nodes that knew about a channel, then the channels get re-established, but the `channel_announcement`s are lost and hence are inexistent in the view of all other nodes in the network. Notice that this does not include scenarios where a single node knowing about the channel since it'll re-broadcast upon re-establishment.

The simplest way of fixing this is to store our node's `channel_announcement` messages to disk and then feed them back into `gossipd` upon restart. Another variant is that we store just the announcement sigs and feed that back to `channeld` on reconnect. The first one is easier, the second stores less data (168 bytes / channel vs. 430 bytes / channel). Optionally we could also store other messages, but they are likely stale anyway, and we'll get a full dump from peers anyway.

I'd propose a simple file-based append-only store for the full `channel_announcement` messages. Alternatively we could also store these in the DB, alongside the channel table.

cdecker:
This simply stores incoming gossip messages, that were accepted into the network topology view, to disk. This way we can replay them on startup, and learn about the network without having to start gossiping with a peer.

It doesn't yet rewrite the store file, so outdated messages may be replayed only to be later replaced, but that should be easy in the next PR :wink: 

Fixes #247


gossipd/gossip_store.c
```diff
@@ -0,0 +1,48 @@
...
+
+struct gossip_store {
+	int read_fd, write_fd;
+};
+
+struct gossip_store *gossip_store_new(const tal_t *ctx)
+{
+	struct gossip_store *gs = tal(ctx, struct gossip_store);
+	gs->write_fd = open(GOSSIP_STORE_FILENAME, O_RDWR|O_APPEND|O_CREAT, 0600);
+	gs->read_fd = open(GOSSIP_STORE_FILENAME, O_RDONLY);
```
ZmnSCPxj: We `open` here, but never `close`. Granted this object has same duration as lifetime of daemon, but I think better for cleanliness.  Probably install a destructor on the `gossip_store` that closes the fd.

gossipd/gossip_store.c
```diff
@@ -24,6 +31,15 @@ void gossip_store_append(struct gossip_store *gs, const u8 *msg)
 	u16 msglen = tal_len(msg);
 	beint16_t belen = CPU_TO_BE16(msglen);
 
+	/* FIXME: this method of detecting replayed messages is best effort
```
ZmnSCPxj: This seems hackish to me.  Perhaps we could do instead:

1.  Split `handle_gossip_message` into `handle_gossip_message_and_store` and `handle_gossip_message`.
    - `handle_gossip_message_and_store` calls `handle_gossip_message`.
    - `handle_gossip_message` adds a new output argument, `bool *new_message`, which indicates whether the given gossip message is new/fresh.
    - In `handle_gossip_message_and_store`, only store the gossip message if `handle_gossip_message` did not return an error and returned a false `new_message` flag.  This is the only point at which gossip messages are stored.
2. peer and master message handlers call `handle_gossip_message_and_store` instead.
3. The individual `handle_cannouncement` etc do not store the message.  Instead they also have a `bool *new_message` output argument that they clear if the message has been received before or set if the message is not encountered before.
4. The gossip store replaying uses `handle_gossip_message`.

ZmnSCPxj:
Would it not be better, to use a database (possibly one separate from `lightningd.sqlite3`)?  The gossip store, as is, would record and keep old gossip messages for years. I believe all gossip has a lifetime of 2 weeks?  We could simply periodically clean up old gossip that has reached 2 weeks from the database.

gossipd/gossip_store.c
```diff
@@ -0,0 +1,48 @@
...
+	struct gossip_store *gs = tal(ctx, struct gossip_store);
+	gs->write_fd = open(GOSSIP_STORE_FILENAME, O_RDWR|O_APPEND|O_CREAT, 0600);
+	gs->read_fd = open(GOSSIP_STORE_FILENAME, O_RDONLY);
+	return gs;
+}
+
+void gossip_store_append(struct gossip_store *gs, const u8 *msg)
+{
+	u16 msglen = tal_len(msg);
+	beint16_t belen = CPU_TO_BE16(msglen);
```
rustyrussell: Use `cpu_to_be16()` instead of` CPU_TO_BE16` which is designed for constants and only exists so that we can get compile-time-constants if needed.

gossipd/gossip_store.c
```diff
@@ -0,0 +1,48 @@
...
+	if (!read_all(gs->read_fd, &belen, sizeof(belen)))
+		return NULL;
+
+	msglen = BE16_TO_CPU(belen);
+	msg = tal_arr(ctx, u8, msglen);
+
+	if (!read_all(gs->read_fd, msg, msglen))
+		status_failed(
+		    STATUS_FAIL_INTERNAL_ERROR,
+		    "Short read from gossip-store, expected lenght %d", msglen);
```
rustyrussell: We should use status_unusual() and truncate the file at this point, IMHO.  Since we're using this as best-effort (ie. no db syncing) we need to handle corruption somewhat gracefully.  In fact, we should probably deal with any bad messages the same way: either truncate at the last good message, or throw away altogether and restart.

gossipd/gossip_store.c
rustyrussell: Agreed, close in destructor is clenaest. But also, you don't need two fds here at all.

1. Keep an off_t for where you're up to in reading.
2. Use pread().
3. You don't need read_all or write_all, since these are normal files and read/write have those semantics.

gossipd/gossip_store.c
rustyrussell: Note that this works in well with the "use pread and keep an offset" advice/

gossipd/gossip.c
```diff
@@ -1522,6 +1522,27 @@ static void setup_listeners(struct daemon *daemon, u16 portnum)
...
+{
+	const u8 *msg, *err;
+	struct daemon *daemon = container_of(dc, struct daemon, master);
+	msg = gossip_store_read_next(tmpctx, daemon->rstate->store);
+
+	if (msg) {
+		err = handle_gossip_msg(daemon, msg);
+
+		/* Since we are replaying from store we should not get errors */
+		assert(!err);
```
rustyrussell: I think we don't want to do this like this:

1. We'll eventually need to take action when we hit the end of store or an error, so know what gossip to ask for.  For the moment this means using the feature bit, which means we want to know this before we connect to anyone.

2. This will continue to replay messages as we get them in, which is unnecessary.

gossipd/gossip_store.c
```diff
@@ -24,6 +31,15 @@ void gossip_store_append(struct gossip_store *gs, const u8 *msg)
...
 
+	/* FIXME: this method of detecting replayed messages is best effort
+	 * only. It should avoid doubling the store file size on every start,
+	 * but it'll allow the last few messages to be duplicated since replay
+	 * and write are async, and we'll think we are done replaying a bit too
+	 * early. */
+        /* Check if we are replaying the store */
+	if (lseek(gs->read_fd, 0, SEEK_CUR) < gs->replaysize)
+		return;
+
```
rustyrussell: Yech.  Definitely need to split this into a startup phase which reads all the gossip, and everything else.  In that phase the store won't write anything.

ZmnSCPxj: I think the main concern is that we want a quick startup, so we just replay gossip when gossipd is not busy doing anything else, which is why `gossip_store` tries to allow reading and writing to interleave.

I think my concern about years-old gossip is still not addressed, too.

gossipd/gossip_store.c
cdecker: I did that at first, with an additional `store` argument, but due to the way we weave the handling through async calls, we'd have to drag that value through everything as well. In the end I had to add the `store` value in 20+ places, which seemed absolute overkill. The reason we need to weave it through is that only at the end of the verification do we actually know whether this is a good message that should be stored or not.

The new output argument you mention does not work exactly because we may need to call out to `masterd` to verify a `txout` or we may have been queuing behind a `channel_announcement` being verified. This just seemed like the fix with the minimal impact.

gossipd/gossip_store.c
cdecker: That's exactly it. The startup phase would need to wait for txout resolution, i.e., the `io_loop` already needs to run, and we might be waiting for `bitcoind` to return blocks, which makes startup incredibly slow. I think we'll probably not get around the fact that we need to weave a `store` argument through the entire process to know whether this is an old or a new message.

gossipd/gossip_store.c
cdecker: I don't see the difference to doing an `lseek` for the `write_fd` when we find a corruption.

gossipd/gossip_store.c
cdecker: I'll add a destructor. But I don't see much of a difference between keeping offsets and using `pread`/`pwrite` instead of just keeping two fds, aside from saving one fd which is not really a bottleneck for `gossipd` anyway.

Good point about not needing `write_all` and `read_all` I'll use those instead.

cdecker:
I added `bool store` to signal whether or not we want the message to be added to the `gossip_store`. This now allows us to drop the simple heuristic we were using before and this should now be cleaner.

I also applied @rustyrussell's tip to use `pread` and `pwrite` instead of keeping two fds open, makes truncating a little bit easier.

ZmnSCPxj:
I still worry about gossip being persisted for years, possibly with very old information that is pointlessly obsolete.  Maybe OK as a first measure but I think the intent is that persisted information (including `gossip_store`) that we make in a 0.6 release should, at least, be upgradeable to a later release, meaning we would now commit ourselves to at least being able to upgrade this to something more pruneable later on.

cdecker:
No problem, I have a patch in the pipeline that rewrites the gossip store from the broadcast queue, which gets rid of duplicates. We can also extend the pruning rules to only allow `channel_update`s and `node_announcement`s that are above the 2 week pruning horizon, so if stuff doesn't get a keepalive it'll get dropped on the floor.

The rewrite-pruning being rather expensive should not happen too often (otherwise I would have just moved the old store aside, and while reading it created a new store), so I'm thinking either on a JSON-RPC trigger or a daily timer. What do you think would be reasonable?

ZmnSCPxj:
Automation is best, so a regular timer (one that is programmable/settable, so that we can test in `test_lightningd.py`).

cdecker:
Excellent, what do you think would be a good timeout for the timer by default?

ZmnSCPxj:
@cdecker Given a 2-week time window, once or twice a day seems reasonable.

@rustyrussell Agree, we need timestamp at least to record when the gossip was taken.  I suppose the SHA256 checks for message/file corruption and makes it safe to trust it.

cdecker:
I don't think the timestamps are necessary, the existing prune logic will get rid of things we want to forget anyway. This leaves us without information that is not already covered by signatures in the messages, so corruption is a non-issue. The regular rewriting will get rid of messages that were replaced with newer ones.

cdecker:
> Not sure how bitcoind tells us about spent channels, though that will be needed.

That'll be given by the `spend_height` updates for outpoints tracked in the `utxoset` (once I implement that feedback).

rustyrussell:
We should rewrite by churn, not time, IMHO.  We can easily keep track of the total size of messages and the number of them: from this we can deduce how "sparse" the file has become.  When it's out by a factor of 5 (?), write a new one and rename it over the top?

cdecker:
Excellent idea Rusty, I'll implement the rewriting right away. As for the additional information in the store and bypassing of the checks (and potentially a hash for consistency) is a nice feature, but I was postponing it as an optimization. Maybe I should tackle that now as well.

rustyrussell:
Ah, but I want timestamps for the gossip-queries branch I'm working on :)

rustyrussell:
And yes, we should store capacity with announcements.  And if we use sha, we can skip signature checks (though we can do them if we want): we want sha because we won't be checking the bitcoin signatures if we don't talk to bitcoind.

rustyrussell:
Oh, and please put a version number on the front of the store, in case we change fmt later!  And refuse to read it if version > current.

cdecker:
> Ah, but I want timestamps for the gossip-queries branch I'm working on :)

Ok, what timestamp should I put there for now? I presume it's not the timestamp included in the message itself.

cdecker:
Kind of went the long way around, but I finally got there. I think I addressed all the mentioned issue:

 - Added a version byte at the beginning of the store, and refuse to load if it doesn't match what we expect
 - Added gossip_wire messages based on the wire protocol that allow us to add additional data to raw gossip messages
 - Gossip store is loaded once after initializing and bypasses verification
 - Got rid of the ugly `store` parameter that differentiated messages coming from peers and those from the store
 - Gossip store uses a single fd and `pread`/`pwrite` as suggested by @rustyrussell 

I tried to make the PR easy to follow, but some of the changes were hard to fixup with reasonable amount of work, so I decided to keep it this way.

gossipd/gossip_wire.csv
```diff
@@ -224,3 +224,18 @@ gossipctl_peer_disconnect_reply,3123
...
+gossip_store_channel_announcement,,announcement,len*u8
+gossip_store_channel_announcement,,satoshis,u64
+
+gossip_store_channel_update,4097
+gossip_store_channel_update,,len,u16
+gossip_store_channel_update,,update,len*u8
+
+gossip_store_node_announcement,4098
+gossip_store_node_announcement,,len,u16
+gossip_store_node_announcement,,announcement,len*u8
```
rustyrussell: We should use a different file for this, but that's OK for now.

gossipd/gossip_store.h
```diff
@@ -29,11 +29,11 @@ void gossip_store_append(struct gossip_store *gs, const u8 *msg);
...
  * @param gs  The `gossip_store` to read from
- * @return The gossip message allocated from `ctx`, `NULL` if no more messages are
- *         available.
+ * @return whether a message was read from the store. False means that all
+ * messages have been processed.
  */
-const u8 *gossip_store_read_next(const tal_t *ctx, struct routing_state *rstate,
-				 struct gossip_store *gs);
+bool gossip_store_read_next(struct routing_state *rstate,
+			    struct gossip_store *gs);
```
rustyrussell: We could now call it gossip_store_init and have it of standard form:

    struct gossip_store *gossip_store_init(const tal_t *ctx, struct routing_state *rstate);

Issue #1194, opened 2018-03-09
channel funding tx fee too low


fixone:
"version" : "v0.5.2-2016-11-21-2262-gd363a68"

any channel I try to fund ends up with 66: min relay fee not met? and the log shows
```
2018-03-09T07:51:40.117Z lightningd(2550): Owning output 0 49543 (SEGWIT) txid ecce68e9f6f03882c8ea7ed389b1f9db7e3c54c6812d4a7ad10320064297aad4
2018-03-09T07:51:40.117Z lightningd(2550): 	 (tx ecce68e9f6f03882c8ea7ed389b1f9db7e3c54c6812d4a7ad10320064297aad4)
2018-03-09T07:51:40.117Z lightningd(2550): sendrawtransaction: 02000000000101158bf6c859e74dfb8b23b583944fb055ffbbfdbd5027f4df4f98168232909d320000000000ffffffff0287c10000000000001600145231c938e2213be703e3b3926e4fda4c1b68ce1ef07e0e000000000022002087d6c3b1e3b2bc251ee0c79d09ae8e85c1a7e05ec72314e281dc21877928ee3802473044022044a2a9a63a34d73bec8ec3873b1f31b8177b07a9a71c630df04ec679c63b1d2b02207ce83dcb73ae6dbcab8d9a4f0893f4439e5447c523b1a3745871e613cabeb215012102d263d904d10206f426099b6f4bd839b221ca2743f5e40b56871726aa7f15037900000000
2018-03-09T07:51:40.118Z lightningd(2550): Client: Received message 9 from client
2018-03-09T07:51:40.131Z lightningd(2550): lightning_channeld-03bf2b602bc3d9dabc6d9172808a0bbe970120ff89be2343ac251980ca8602ffe6 chan #92: pid 841, msgfd 54
2018-03-09T07:51:40.131Z lightningd(2550): 03bf2b602bc3d9dabc6d9172808a0bbe970120ff89be2343ac251980ca8602ffe6 chan #92: Waiting for funding confirmations
2018-03-09T07:51:40.131Z lightningd(2550): 03bf2b602bc3d9dabc6d9172808a0bbe970120ff89be2343ac251980ca8602ffe6 chan #92: Ignoring fee limits!
2018-03-09T07:51:40.136Z lightningd(2550):jcon fd 39: Success
2018-03-09T07:51:40.138Z lightningd(2550): sendrawtx exit 26, gave error code: -26?error message:?66: min relay fee not met?
[...]

PR #1247, opened 2018-03-19
Update onchain.c

roconnor-blockstream:
Round tx fees up.

Fixes https://github.com/ElementsProject/lightning/issues/1194#issuecomment-373333023 ... maybe.  I didn't even try compiling this.  I just directly edited it in github.

### DIFF
--- a/onchaind/onchain.c
+++ b/onchaind/onchain.c
@@ -242,8 +242,8 @@ static struct bitcoin_tx *tx_to_us(const tal_t *ctx,
 						   &our_wallet_pubkey);
 
 	/* Worst-case sig is 73 bytes */
-	fee = feerate_per_kw * (measure_tx_weight(tx)
-			 + 1 + 3 + 73 + 0 + tal_len(wscript))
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
ACK 806bc022dcc8a0f69c929de9130a077174ca3071

jb55:
this is begging for tests ;)

ZmnSCPxj:
Looking more deeply, maybe it is incomplete? `wallet/wallet.c` also contains fee calculation code.https://github.com/ElementsProject/lightning/blob/044705a547f853cb89fba8a6711ac5420f658620/wallet/wallet.c#L289

roconnor-blockstream:
Good catch in `wallet.c`

Sorry, this PR wasn't entirely serious.

roconnor-blockstream:
Actually, while I think these are all errors, Perhaps they are not causing the problems in #1194.  Since feerate_per_kw should be exactly 4000, the division should be exact here.

roconnor-blockstream:
Superseded  by https://github.com/ElementsProject/lightning/pull/1251

Issue #1210, opened 2018-03-13
Cannot connect to node


asocjal:
### Issue and Steps to Reproduce
When created channel between both c-lightning version v0.5.2-2016-11-21-2269-g249464c channels worked fine. After some time I realized that nodes are nort connected (listpeers). When trying to connect from cli, I have error:

```
{ "peers" : 
        [ 
                { "id" : "034fd0be2b82f609f30593d2c5b7f27eb33c5e84f660446e4bf1501c838ce052d4", "connected" : false, "channels" : 
                        [ 
                                { "state" : "ONCHAIN", "short_channel_id" : "513012:1879:0", "funding_txid" : "de62f6cae46f0545e46f4436ba645d697b17961d373b6d9743ecbbd50957a216", "msatoshi_to_us" : 2225000, "msatoshi_total" : 100000000, "dust_limit_satoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minimum_msat" : 0, "to_self_delay" : 144, "max_accepted_htlcs" : 483, "status" : 
                                        [ "Loaded from database" ] } ] } ] }
```

```
root@cln-02:~/cln/lightning/cli# ./lightning-cli connect 034fd0be2b82f609f30593d2c5b7f27eb33c5e84f660446e4bf1501c838ce052d4
{ "code" : -1, "message" : "Could not connect to 034fd0be2b82f609f30593d2c5b7f27eb33c5e84f660446e4bf1501c838ce052d4 after 45 seconds and 10 attempts" }
```

### `getinfo` output
node1:
`{ "id" : "03f8245976b7cbcf8872e9c328abca86f0f42e503b0532430608f11ad73c6bd4a0", "port" : 9735, "address" : 
        [ 
                { "type" : "ipv4", "address" : "80.211.252.104", "port" : 9735 } ], "version" : "v0.5.2-2016-11-21-2269-g249464c", "blockheight" : 513370, "network" : "$itcoin" }`

node2
`{ "id" : "034fd0be2b82f609f30593d2c5b7f27eb33c5e84f660446e4bf1501c838ce052d4", "port" : 9735, "address" : 
        [ 
                { "type" : "ipv4", "address" : "80.211.252.169", "port" : 9735 } ], "version" : "v0.5.2-2016-11-21-2269-g249464c", "blockheight" : 513352, "network" : "b
itcoin" }
`

[...]

PR #1251, opened 2018-03-20
chaintopology: set a fee floor to avoid creating unrelayable txs.

rustyrussell:
Naively, this would be 250 satoshi per sipa, but it's not since bitcoind's
fee calculation was not rewritten to deal with weight, but instead bolted
on using vbytes.

The resulting calculations made me cry; I dried my tears on the thorns
of BUILD_ASSERT (I know that makes no sense, but bear with me here as I'm
trying not to swear at my bitcoind colleagues right now).

Fixes: #1194

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -324,10 +324,15 @@ static void update_feerates(struct bitcoind *bitcoind,
 		if (feerate < feerate_floor())
 			feerate = feerate_floor();
 
-		if (feerate != topo->feerate[i])
+		if (feerate != topo->feerate[i]) {
 			log_debug(topo->log, "%s feerate %u (was %u)",
 				  feerate_name(i),
 				  feerate, topo->feerate[i]);
+			if (feerate != satoshi_per_kw[i])
+				log_debug(topo->log,
+					  "...feerate %u hit floor %u",
+					  satoshi_per_kw[i], feerate);
+		}
 		old_feerates[i] = topo->feerate[i];
 		topo->feerate[i] = feerate;
 	}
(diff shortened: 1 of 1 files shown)


rustyrussell:
I think this is the more general fix than #1247 in that it fixes all transactions we generate.

lightningd/chaintopology.c
```diff
@@ -277,12 +319,17 @@ static void update_feerates(struct bitcoind *bitcoind,
 	bool changed = false;
 
 	for (size_t i = 0; i < NUM_FEERATES; i++) {
-		if (satoshi_per_kw[i] != topo->feerate[i])
+		u32 feerate = satoshi_per_kw[i];
+
+		if (feerate < feerate_floor())
+			feerate = feerate_floor();
```
ZmnSCPxj: Suggest adding a `log_debug` here that we are bumping the feerate up because reasons.

ZmnSCPxj:
ACK c059b8d5688d518fba9741dac0ab6c719f696989

lightningd/chaintopology.c
rustyrussell: Done.

ZmnSCPxj:
ACK 1d3b05661df45ba214b6beca2ac2702f497c8454

ZmnSCPxj:
Amusing: 
@rustyrussell wants to merge 2 commits into `ElementsProject:master` from `rustyrussell:omg-bitcoind-vbytes-must-die-die-die`

cdecker:
Fixes #1210

cdecker:
ACK 1d3b056

Issue #1108, opened 2018-02-26
JSON RPC hangs on special characters


bitonic-cjp:
### Issue and Steps to Reproduce
I discovered that adding a newline to a command argument value in the Python RPC interface makes it hang. I first discovered this on the bolt11 argument of the decodepay command (not very important - it's not supposed to contain a newline), but it seems to be the case for all arguments of all commands. I tested the description argument of the invoice command, and the problem appears there too. This is a bigger issue: in my view, descriptions are likely to contain newline characters.

Looking at the logging, the Python RPC client says:

> 2018-02-26 15:36:06,958 DEBUG: Calling decodepay with payload {'description': None, 'bolt11': '\n'}

Note: it does not log anything about return data; apparently it hangs, waiting for a reply.
Meanwhile, lightningd logs:

> 2018-02-26T14:36:06.958Z lightningd(3239): Invalid token in json input: '{"method": "decodepay", "params": {"bolt11": "\n"}, "id": 0}'

Note that, while the RPC client hangs, lightningd does not hang.

Looking at the code, it seems the Python client side correctly escapes special characters, turning a newline into \n. I think lightningd rejects special characters because of the code added in commit 5502a19d1ecf7517535d995ccbc95d201e67d318. This includes the backspace added by the escaped newline character.

Instead of directly making a bug fix, I'd like to discuss what the desired behavior should be. Following BOLT-11, any valid UTF-8 string should be a valid description, right? On the RPC-level, it may be required to escape certain characters to turn any UTF-8 string into a JSON-encoded string. The Python-RPC client side already does this - I think the lightningd server side should also support it.

Is this the desired behavior for all command arguments?

### `getinfo`  output
getinfo
{ 'address': [],
  'blockheight': 1286431,
  'id': '03f3b0a454583f712318a0c9023423b3093ce534dff288907107b3ec337ccda10f',
  'network': 'testnet',
  'port': 9700,
  'version': 'v0.5.2-2016-11-21-2014-g46534935'}

ZmnSCPxj:
Duplicates: #607 

Duplicates: #387

Issue #387, opened 2017-11-28
JSON parse chokes on escaped quotes in strings


cdecker:
The JSON parser is apparently not doing a good job at escaping strings:
```
lightningd(17291): Invalid token in json input: '{"method":"invoice","params":["603777911","Um3AE6UKOh5dk3aGQEg64","Lightning Shop: <span class=\"woocommerce-Price-amount amount\"><span class=\"woocommerce-Price-currencySymbol\">&#36;</span>60.05</span> for 7 Pizza, 5 Guitar and 6 Cat",null],"id":10}'
```

Reported-by: @shesek

rustyrussell:
It's deliberately preventing strings which have weird chars in them, such as escapes.

Now, we *could* hoist the strange_chars() check into the callers, and add a '*' prefix (or something) to json_get_params to say a parameter *may* be escaped, then use that for the invoice description, then provide a santize_json_string() helper to un-escape it as required.

But in this case, it looks like we caught a bug in the caller?  Putting HTML fields in the description is making assumptions about the clients (wallets) that I don't think are reasonable.

shesek:
@rustyrussell you're right, the HTML was indeed a bug and shouldn't have been there (WooCommerce was returning the price in that format, I since fixed that), but this is still valid JSON. There might be a legitimate reason to have `"` or `\` inside the description field.

Can't c-lightning integrate a fully spec-compliant JSON parser library to handle that?

rustyrussell:
Nadav Ivgi <notifications@github.com> writes:
> @rustyrussell you're right, the HTML was indeed a bug and shouldn't have been there (WooCommerce was returning the price in that format, I since fixed that), but this is still valid JSON. There might be a legitimate reason to have `"` or `\` inside the description field.
>
> Can't c-lightning integrate a fully spec-compliant JSON parser library to handle that?

Well, our parser is fully spec compliant.  I chose not to do full spec
compiance in our output originally, but rather refuse to pass through
anything weird.

That seemed safer if *other* users have JSON escape issues.  Thus the
few places where we take strings, I forbade it.

However, I agree; we should simply re-escape it when we output strings.

Hence this PR is left open...

PR #1254, opened 2018-03-21
Cleanup jsonrpc stringhandling

rustyrussell:
Fixes: #1108 

Fixes: #387 

This introduces a new type for things we know are JSON escaped, to try to maintain a safety net.  I then audited all the uses, and found some places which needed patching.  Finally, this gets rid of the JSON input restriction we had, and does more intelligent escaping for output (particularly for logs).

(I have a more ambitious rework of json parameter parsing, but it's a big job that can wait)


common/json.c
```diff
@@ -538,6 +489,15 @@ void json_add_object(struct json_result *result, ...)
 	va_end(ap);
 }
 
+void json_add_escaped_string(struct json_result *result, const char *fieldname,
```
ZmnSCPxj: Function is defined in `common/json.c`, but is declared in header `common/json_escaped.h`.

lightningd/invoice.c
```diff
@@ -110,7 +111,7 @@ static void json_invoice(struct command *cmd,
 	struct invoice_details details;
 	jsmntok_t *msatoshi, *label, *desc, *exp, *fallback;
 	u64 *msatoshi_val;
-	const char *label_val;
+	const struct json_escaped *label_val;
 	const char *desc_val;
```
ZmnSCPxj: Why not description too?

lightningd/invoice.c
```diff
@@ -284,22 +281,34 @@ static void json_listinvoice_internal(struct command *cmd,
...
 	if (!json_get_params(cmd, buffer, params,
-			     "?label", &label,
+			     "?label", &labeltok,
 			     NULL)) {
 		return;
 	}
 
+	if (labeltok) {
+		label = json_tok_escaped_string(cmd, buffer, labeltok);
+		if (!label) {
```
ZmnSCPxj: Before this change, we could label an invoice as `null`, `true`, `false`, or a true number, and use it as-is.  This is interesting in the case where `lightning-cli` is used instead of a direct programmatic RPC interface: if you give a number on the command line it is passed as a number.  On previous versions giving `label=0` on the `lightning-cli` command line would succeed in much the same way that giving `label=a` would, but now will `label=0` would fail and would need to add explicit quotes (which would also need to be escaped at the system shell level).  Maybe this is OK, though.

wallet/db.c
```diff
@@ -754,3 +755,18 @@ bool sqlite3_bind_sha256_double(sqlite3_stmt *stmt, int col, const struct sha256
...
+{
+	return json_escaped_string_(ctx,
+				    sqlite3_column_blob(stmt, col),
+				    sqlite3_column_bytes(stmt, col));
+}
+
+bool sqlite3_bind_json_escaped(sqlite3_stmt *stmt, int col,
+			       const struct json_escaped *esc)
+{
+	sqlite3_bind_text(stmt, col, esc->s, strlen(esc->s), SQLITE_TRANSIENT);
```
ZmnSCPxj: This makes the string data on-database to be escaped as per JSON rules.  Probably OK and any unexpected string data passed to us is the fault of the client, but, we will be "stuck" with this encoding permanently.

ZmnSCPxj:
Fixes: #1108

common/json.c
rustyrussell: Oops. I moved it because it wanted access to internal routines in json.c. Will move declaration.

lightningd/invoice.c
rustyrussell: Hmm, numeric labels make sense, but we never *really* had them. Label 00 was different from label 0 for example, and fixing that is a bigger job.

wallet/db.c
rustyrussell: True.

ZmnSCPxj:
ACK fa2af56a35e5c8151990b53c8eef67a66f0a844d

Issue #1086, opened 2018-02-24
Add delay limit to `pay`


ZmnSCPxj:
In LN, there is a delay for each hop, increasing for each forwarding hop.  However, the delay for this affects the payer when a node on the route forwards the HTLC and then dies or disconnects before it can receive the response from the next node: the money is effectively locked for this time.

Inspired by #1065.

I would like to do this after #1012 gets merged in.

PR #1256, opened 2018-03-21
payalgo: Add maximum delay.

ZmnSCPxj:
Fixes: #1086

***NOTE*** Although I use `delay` here, I wonder if a better term would be  `locktime`.  This is because, in the worst case, this is the number of blocks that the money will get locked before it can be recovered.

### DIFF
--- a/doc/lightning-pay.7
+++ b/doc/lightning-pay.7
@@ -2,12 +2,12 @@
 .\"     Title: lightning-pay
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 03/16/2018
+.\"      Date: 03/21/2018
 .\"    Manual: \ \&
 .\"    Source: \ \&
 .\"  Language: English
 .\"
-.TH "LIGHTNING\-PAY" "7" "03/16/2018" "\ \&" "\ \&"
+.TH "LIGHTNING\-PAY" "7" "03/21/2018" "\ \&" "\ \&"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


rustyrussell:
My current stats for mainnet give the following channel delays:

| Number of channels  | Delay |
| ------------------- | ------------- |
 | 3 | 4 |
 | 5 | 6 |
 | 1629 | 14 |
 | 8 | 16 |
 | 1581 | 144 |
 | 12 | 1000 |

ZmnSCPxj:
So what is a good limit? 300, to support up to two 144-delay hops plus some margin for the final payee delay?

ZmnSCPxj:
ping!

cdecker:
Wow, those delays are crazy, I definitely don't feel comfortable with the 144 delays. I think 500 for the entire route should be sufficient.

cdecker:
RE: locktime vs. delay: we always use locktime for absolute values (expressed as blockheight), while we use delay for relative values (expressed as blockheight differences) :wink:

ZmnSCPxj:
@cdecker I wonder how to explain it to users, though.  The total maximum delay is the amount of time (measured in blocks) their money could get locked in the worst case (and cannot be used to pay somebody else, and will take that time to be recovered if dropped onchain), hence "lock time" or "lockup time". But I already implemented as "delay" so fine with me.

So 500 as default maximum delay? That is more than 3 days of lockup in the worst case.

cdecker:
Yeah, I think that's a really hard one to explain in simple terms, so the best we can do currently is to provide some reasonable default so that most people just don't have to bother to understand this.

What that reasonable default value is however, that's hard to say. I ran a [quick check][ipython-notebook] on the current mainnet, and it seems that 500 maximum delay is more than enough to reach any point from any other point in the network.

[ipython-notebook]: https://gist.github.com/cdecker/649f1230139dc25bb6b2ffa9569e3ba4

rustyrussell:
Ack 9dd98fda7c6f14e2ad47b3d37aad9885ae15486e

Issue #1089, opened 2018-02-24
Randomized overpaying (treat as routing fees) to reduce traceability


ZmnSCPxj:
BOLT#4 specifically allows overpayment of invoices of up to twice the invoice amount.  BOLT #11 explicitly mentions:

> the payee SHOULD accept up to twice the expected amount, so the payer can make payments harder to track by adding small variations.

My intent is the below:

1.  In combination with `fuzz` from #1012, also generate a random amount to add up to some fraction of the `maxfeepercent` setting to add to the destination.
2.  If the resulting route exceeds the actual desired payment by more than `maxfeepercent`, reduce the fraction for overpayment and/or `fuzz` (maybe even use the same variable for both?) and retry.  Only raise a `PAY_ROUTE_TOO_EXPENSIVE`) if both the fuzz and overpayment fraction are very close to 0.0

An idea is to pay exactly when the resulting route is a single hop (i.e. direct route to destination), since nobody else will receive our onion routing message and the destination already knows how much the exact payment is (it generated the invoice...), but if we want to hide our identity from the destination, we will want to be able to pretend to be somebody else further away, meaning we should still overpay in that case.

In combination #1086, we could have "plausible" shadow routes by also overdelaying payment timeouts.

PR #1257, opened 2018-03-21
payalgo: Be willing to overpay up to maxfeepercent, for privacy.

ZmnSCPxj:
This obscures how far an intermediate hop is from the ultimate
payee, and also obscures slightly the exact payment value.

Fixes: #1089

### DIFF
--- a/lightningd/payalgo.c
+++ b/lightningd/payalgo.c
@@ -5,6 +5,7 @@
 #include <ccan/tal/str/str.h>
 #include <ccan/time/time.h>
 #include <common/bolt11.h>
+#include <common/pseudorand.h>
 #include <common/timeout.h>
 #include <common/type_to_string.h>
 #include <gossipd/gen_gossip_wire.h>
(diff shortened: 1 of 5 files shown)


lightningd/payalgo.c
```diff
@@ -514,13 +516,29 @@ static bool json_pay_try(struct pay *pay)
...
+	 * we do not end up paying more than maxfeepercent
+	 * of the msatoshi we intend to pay. */
+	maxoverpayment = ((double) pay->msatoshi * pay->fuzz * pay->maxfeepercent)
+		/ 100.0;
+	if (maxoverpayment > 0) {
+		randombytes_buf(&overpayment, sizeof(overpayment));
+		/* We will never generate the maximum computed
+		 * overpayment this way. Maybe OK for most
+		 * purposes. */
+		overpayment = overpayment % maxoverpayment;
```
rustyrussell: We have pseudorand for this, which makes this code neater and also indicates we don't need cryptographic randomness.  We only use randombytes_buf() above because we don't have a routine to give us a pseudorandom siphash seed, but that would be trivial to implement too.

```
if (maxoverpayment > 0)
     overpayment = pseudorand(maxoverpayment);
else
    overpayment = 0;
```

ZmnSCPxj:
ping!

cdecker:
ACK 1c84a33

Issue #840, opened 2018-01-29
Fatal signal 11 received on Mac OS


bitstein:
### Issue and Steps to Reproduce
I am running Mac OS X 10.13.2. With the new libwally update, I was able to build c-lightning just fine this morning. However, after I got my bitcoind testnet daemon synced and booted lightningd, upon making a command, I received this output in the lightningd log:

```
lightningd(68725): Connected json input
lightningd(68725): FATAL SIGNAL 11 RECEIVED
```

### `getinfo` output

`getinfo` hangs, no output.

cdecker:
Can you run with a debugger? `gdb lightningd` might work, and once it breaks type in `bt` to see a backtrace. The copy paste it here :-)

yashbhutwala:
@cdecker I'm having the same issues as @bitstein on a Mac. I tried to run debug using lightningd, but for me both the daemon and the client hang upon any request, so I'm not getting a backtrace.

adam-hudspith:
I'm getting the same problem. I managed to get this:
`(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x00007fff73cf041a libsystem_kernel.dylib`read + 10
  * frame #1: 0x0000000104730b1a lightningd`check_with_child(batch=0x00007ffeeb540608, p=0x00000001046be000, size=8, is_write=false) at ptr_valid.c:258
    frame #2: 0x000000010473087b lightningd`ptr_valid_batch(batch=0x00007ffeeb540608, p=0x00000001046be000, alignment=1, size=8, write=false) at ptr_valid.c:289
    frame #3: 0x000000010471cf8f lightningd`autodata_make_table(example=0x00000001047d0d70, name="json_command", nump=0x00000001047d1b20) at autodata.c:33
    frame #4: 0x00000001046caedf lightningd`get_cmdlist at jsonrpc.c:281
    frame #5: 0x00000001046cc335 lightningd`find_cmd(buffer="{ \"method\" : \"help\", \"id\" : \"lightning-cli-68421\", \"params\" : [ ] }", tok=0x00007fc51d009978) at jsonrpc.c:312
    frame #6: 0x00000001046cc0de lightningd`parse_request(jcon=0x00007fc51d000760, tok=0x00007fc51d009950) at jsonrpc.c:574
[...]

Issue #1093, opened 2018-02-24
SIGABRT on macOS


Sjors:
Perhaps related to #840. Fresh macOS node compiled using #1090.

```
 lldb -- lightningd/lightningd --port=9736
(lldb) target create "lightningd/lightningd"
Current executable set to 'lightningd/lightningd' (x86_64).
(lldb) settings set -- target.run-args  "--port=9736"
(lldb) run
Process 8200 launched: '/Users/bitcoin/src/lightning/lightningd/lightningd' (x86_64)
2018-02-24T15:17:12.856Z lightningd(8200): Creating configuration directory /Users/bitcoin/.lightning
2018-02-24T15:17:12.910Z lightningd(8200): Creating database
2018-02-24T15:17:13.006Z lightningd(8200): Server started with public key 0316182b1c2e66d436f6f09e624b9d6a6b29d5461db1b604dc2a5a673a72e824fc, alias REDWALK--21-2093-g3ad078d2-dirty (color #031618) and lightningd v0.5.2-2016-11-21-2093-g3ad078d2-dirty
bitcoin-cli -testnet getblock 00000000000000ed3e35f4c19100865fc96e410f772b3d119779447171b3112c false Interrupted system call
2018-02-24T15:17:13.530Z lightningd(8200): bitcoin-cli -testnet getblock 00000000000000ed3e35f4c19100865fc96e410f772b3d119779447171b3112c false Interrupted system call
Log dumped in crash.log
Process 8200 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fff58a7fe3e libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff58a7fe3e <+10>: jae    0x7fff58a7fe48            ; <+20>
    0x7fff58a7fe40 <+12>: movq   %rax, %rdi
    0x7fff58a7fe43 <+15>: jmp    0x7fff58a770b8            ; cerror_nocancel
    0x7fff58a7fe48 <+20>: retq   
Target 0: (lightningd) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fff58a7fe3e libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff58bbe150 libsystem_pthread.dylib`pthread_kill + 333
[...]

PR #1258, opened 2018-03-22
Disable backtrace on mac

conanoc:
fix https://github.com/ElementsProject/lightning/issues/840

backtrace module does not support mac. So, disable crash log if BACKTRACE_SUPPORTED is 0.

The reason why this patch fix https://github.com/ElementsProject/lightning/issues/840 is not fully identified. On mac, SIGSEGV and SIGBUS signal happens after jsonrpc only when signal handlers are registered with ``sigaction()``.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.pb
 *.gcda
 *.gcno
+*.dSYM
 TAGS
 ccan/tools/configurator/configurator
 ccan/ccan/cdump/tools/cdump-enumstr
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,4 @@
 [submodule "external/libwally-core"]
 	path = external/libwally-core
 	url = https://github.com/ElementsProject/libwally-core.git
+	ignore = dirty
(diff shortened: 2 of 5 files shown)


isghe:
@lvaccaro @gabridome 
ACK 56ef03666135f78df04ab134d07361966b0cf76a

Sjors:
Awesome, this works for me on macOS. I'm now able to connect to peers and fund channels (master  crashes on launch once I have some channels open).

conanoc:
yeah. I saw the same issue with lldb. But, it's not related with this patch. You could file a new issue for that.

arowser:
Yes, it seems not related, and the issue have been addressed in #1093

laanwj:
> The reason why this patch fix #840 is not fully identified. On mac, SIGSEGV and SIGBUS signal happens after jsonrpc only when signal handlers are registered with sigaction().

What happens (see https://github.com/ElementsProject/lightning/issues/840#issuecomment-364400896) is that `autodata`  (the library used for constructing RPC and other tables) uses SIGSEGV/SIGBUS on MacOSX and other BSDs to probe memory maps (in a subprocess). It registers its own handler for this, but apparently backtrace's handler also still triggers. This results in a logged error.

So this fix, removing backtrace, is overkill but it works. Another option would have been to disable backtrace handling just for autodata's subprocesses.

PR #1260, opened 2018-03-22
Track confirmations for change and their_unilateral/to_us outputs

### LINKED ISSUE #1209: Confirmations for `their_unilateral/to_us` outputs are not tracked correctly
Due to `their_unilateral/to_us` being a special case, and the key corresponding to that address not being part of the HD key derivation, it also needs a bit of extra care when loading from DB and adding to the `owned_outpoints` filter.

cdecker:
We were not tracking confirmations for change outputs and their_unilateral/to_us outputs:

 - Scripts of change outputs were not being added to the txfilter, hence we didn't get a notification
 - their_unilateral/to_us were not added to the txfilter and since we learnt about them only _after_ they were confirmed through `onchaind` we were missing its confirmation.

This PR fixes both, and includes a small fix.

Fixes #1209

### DIFF
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -371,6 +371,12 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 	/* Extract the change output and add it to the DB */
 	wallet_extract_owned_outputs(ld->wallet, fundingtx, NULL, &change_satoshi);
 
+	/* Make sure we recognize our change output by its scriptpubkey in
+	 * future. This assumes that we have only two outputs, may not be true
+	 * if we add support for multifundchannel */
+	if (tal_count(fundingtx->output) == 2)
+		txfilter_add_scriptpubkey(ld->owned_txfilter, fundingtx->output[!funding_outnum].script);
+
 	/* Send it out and watch for confirms. */
 	broadcast_tx(ld->topology, channel, fundingtx, funding_broadcast_failed);
 
--- a/onchaind/onchain_wire.csv
+++ b/onchaind/onchain_wire.csv
@@ -87,3 +87,4 @@ onchain_add_utxo,,prev_out_tx,struct bitcoin_txid
 onchain_add_utxo,,prev_out_index,u32
 onchain_add_utxo,,per_commit_point,struct pubkey
 onchain_add_utxo,,value,u64
+onchain_add_utxo,,blockheight,u32
--- a/common/utxo.h
+++ b/common/utxo.h
@@ -28,10 +28,10 @@ struct utxo {
 	struct unilateral_close_info *close_info;
 
 	/* NULL if we haven't seen it in a block, otherwise the block it's in */
-	const int *blockheight;
+	const u32 *blockheight;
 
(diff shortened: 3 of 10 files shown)


lightningd/opening_control.c
```diff
@@ -371,6 +371,10 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 	/* Extract the change output and add it to the DB */
 	wallet_extract_owned_outputs(ld->wallet, fundingtx, NULL, &change_satoshi);
 
+	/* Make sure we recognize our change output by its scriptpubkey in future */
+	if (tal_count(fundingtx->output) == 2)
```
ZmnSCPxj: Would not work if ever we make a `multifundchannel`, but maybe cross that bridge when we get there.

onchaind/onchain_wire.csv
```diff
@@ -87,3 +87,4 @@ onchain_add_utxo,,prev_out_tx,struct bitcoin_txid
 onchain_add_utxo,,prev_out_index,u32
 onchain_add_utxo,,per_commit_point,struct pubkey
 onchain_add_utxo,,value,u64
+onchain_add_utxo,,blockheight,u32
```
ZmnSCPxj: No newline at end-of-file, this could cause issues with some editors that insist on putting newlines at eof.

lightningd/opening_control.c
cdecker: Good point, I'll add a `FIXME`, presumably we'd have to change things around anyway, since the code explicitly expects a single funding output in its current state.

cdecker:
Should be good now, decided not to add a FIXME because it doesn't require an immediate fix, but added a note should we eventually support `multifundchannel`.

ZmnSCPxj:
ACK 6c0b68fcf4379b61ae6799058f8b9be5ab13b87b

lightningd/opening_control.c
```diff
@@ -371,6 +371,12 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 	/* Extract the change output and add it to the DB */
 	wallet_extract_owned_outputs(ld->wallet, fundingtx, NULL, &change_satoshi);
 
+	/* Make sure we recognize our change output by its scriptpubkey in
+	 * future. This assumes that we have only two outputs, may not be true
+	 * if we add support for multifundchannel */
+	if (tal_count(fundingtx->output) == 2)
+		txfilter_add_scriptpubkey(ld->owned_txfilter, fundingtx->output[!funding_outnum].script);
+
```
rustyrussell: I'm a bit confused... since this spend is to our own wallet, don't we watch it anyway, just as if someone else send to that address?

rustyrussell: Shouldn't we be watching all known keyindexes all the time? Incrementing the key index should also mean watching for any outputs. Seems simplest and safest?

cdecker: Indeed we should, but is it better to push the scriptpubkey generation and registration in the filter down into the keyindex generation? Alternatively we could add a new function that takes care of that. It may need to tear `funding_tx` open to inject the scriptpubkey generated externally.

PR #1262, opened 2018-03-22
Add a basic static analysis pass as part of Travis check (via check-source using cppcheck)

practicalswift:
Add a basic static analysis pass as part of Travis check (via `check-source` using `cppcheck`).

This will increase the likelihood of catching null pointer dereferences, stack overflows and other issues before merge.

`make check-cppcheck` takes ~45 seconds so it should be worth its cost.

**Note to reviewers:** `cppcheck` doesn't generate a lot of false positives in general but it appears that it has some problems understanding how the second parameter to `list_for_each_safe` and `list_for_each_safe` are used. I worked around that by initializing to `NULL`. Please let me know if you find a smarter way to fix this.


ZmnSCPxj:
I think @rustyrussell would object to unnecessary initialization of variables/fields.  `valgrind` can probably do a more accurate job of reporting misuse of uninitialized data.  I guess we need to balance that with being able to catch some errors in the quicker source-only test, though.

practicalswift:
@ZmnSCPxj I can see that argument for initialization of fields, but initializing a pointer to struct to `NULL` shouldn't make any difference for `valgrind`, no?

ZmnSCPxj:
@practicalswift yes, although I think I saw some initializations for non-pointers `a`, `b`, `c`, `d` somewhere... https://github.com/practicalswift/lightning/blob/892bb999804e4ab6b9a5a82260ef2e67c843caa6/gossipd/test/run-find_route.c#L143

practicalswift:
@ZmnSCPxj Ah, you're right – the change to `run-find_route.c` is the exception. Hopefully that change can be accepted in order to get the benefit of automatic `cppcheck` checking before merge.

Technically only `a` needs to be initialized to make `cppcheck` happy. I initialized `b`, `c` and `d` for consistency but can revert that to keep the change to a minimum.

This is the error:

```
$ cppcheck -q --language=c --std=c11 --error-exitcode=1 gossipd/test/run-find_route.c
[gossipd/test/run-find_route.c:154]: (error) Uninitialized variable: a
```

Let me know if the initialization of `a` is not acceptable and I'll add a `cppcheck` suppression for that error instead :-)

ZmnSCPxj:
ACK d955aa05340c5f9befe87b1fc87fbf7812cf2d03

Okay by me.  Waiting for @rustyrussell to object or not.

practicalswift:
Oh, perhaps the reported issue in `new_routing_state(…, …, a, …)` in [`gossipd/test/run-find_route.c:154`](https://github.com/ElementsProject/lightning/blob/master/gossipd/test/run-find_route.c#L143-L154) isn't a false positive?

ZmnSCPxj:
You seem correct.  Probably the intent was to compute the public key from the `aaaaaa` private key.  Good catch!

practicalswift:
@ZmnSCPxj Like this?

```diff
diff --git a/gossipd/test/run-find_route.c b/gossipd/test/run-find_route.c
index b7ed6d4..8c3c763 100644
--- a/gossipd/test/run-find_route.c
+++ b/gossipd/test/run-find_route.c
@@ -151,9 +151,9 @@ int main(void)
        setup_tmpctx();

        memset(&tmp, 'a', sizeof(tmp));
+       pubkey_from_privkey(&tmp, &a);
        rstate = new_routing_state(tmpctx, &zerohash, &a, 0);

-       pubkey_from_privkey(&tmp, &a);
        new_node(rstate, &a);

        memset(&tmp, 'b', sizeof(tmp));
```

ZmnSCPxj:
Likely.

practicalswift:
Rebased! Please review :-)

ZmnSCPxj:
ACK e12be8420f8b8f0cc8181099608cb5f2349ac418

Although the docker image has no `cppcheck`....

rustyrussell:
That was a lot of work to find nothing of value, which is disappointing.

How about this?

https://github.com/rustyrussell/lightning/commit/289838887afe6ab231351ebfc57704b753bc9796

practicalswift:
@rustyrussell Nice auto-generation of suppressions for all the `list_for_each(_safe)`. Cherry-picked into this PR. Thanks!

@ZmnSCPxj Please re-review :-)

rustyrussell:
Rebased again (trivial) to kick off travis...

practicalswift:
@cdecker Would it be possible to get `cppcheck` installed on the Travis Docker images to be able to test this PR? :-)

cdecker:
Sorry for the delay. The docker images now contain the `cppcheck` utility, but it complains about an uninitialized variable:

```
[hsmd/hsm.c:629]: (error) Uninitialized variable: changekey
[openingd/opening.c:335]: (error) Uninitialized variable: changekey
Makefile:266: recipe for target 'check-cppcheck' failed
make: *** [check-cppcheck] Error 123
```

I'll file a PR for the updated `Dockerfile`s :wink:

practicalswift:
Now passing Travis. Please re-review :-)

ZmnSCPxj:
ACK 014cbddc9a39e86c95710096f99588b852531f05

hsmd/hsm.c
```diff
@@ -610,7 +610,7 @@ static void sign_funding_tx(struct daemon_conn *master, const u8 *msg)
 	struct bitcoin_tx *tx;
 	u16 outnum;
 	size_t i;
-	struct pubkey changekey;
+	struct pubkey changekey = {};
```
cdecker: Isn't this exactly the kind of initialization that @rustyrussell hates?

practicalswift: I'm afraid so! :-(

What makes `funding_tx`, `list_for_each` and `list_for_each_safe` special? Seems like only these three trigger uninit warnings.

If `cppcheck` had a general issue with uninit `struct`:s being passed as arguments we'd see false positives all over the code base I guess?

rustyrussell:
Does this work instead?  Pushed to get Travis to test for me....

hsmd/hsm.c
rustyrussell: Yes!

In this case, it's close to correct.  We *do* hand changekey to funding_tx, but it's only used if change != 0: the same condition which we use to initialize changekey.

But I don't get the warning here with cppcheck 1.80 in Ubuntu.  I'll try frobbing it and seeing if I can fix it...

cdecker:
ACK 70bebd8

practicalswift:
@ZmnSCPxj @rustyrussell Would you mind re-reviewing? :-)

ZmnSCPxj:
ACK 70bebd877d39797078e8e0c3eda4d1e8494ed9df

PR #1259, opened 2018-03-22
Fix redundant includes. Fix inconsistent LIGHTNING_DIR_FILE_H defines. Add regression test.

practicalswift:
* Fix redundant includes: `foo.c` includes `bar.h`, but `bar.h` is already included in `foo.h`
* Fix inconsistent `LIGHTNING_DIR_FILE_H` defines
* Add regression test

PR #1271, opened 2018-03-25
lightningd/subd.h: Add missing wire/wire.h.

ZmnSCPxj:
If not included, a source file containing only `#include<lightningd/subd.h>` will fail compilation.


practicalswift:
@ZmnSCPxj Makes sense, but please also remove `#include <wire/wire.h>` in `lightningd/subd.c`.

FWIW having #1259 merged would have catched this redundant include :-)

PR #1273, opened 2018-03-25
Add --bitcoin-rpcport option

icota:
It's useful when we are connecting to a bitcoind set to a port other than default 8332


practicalswift:
Concept ACK

cdecker:
ACK 4a9c159

PR #1275, opened 2018-03-25
Remove unused functions not covered by tests

practicalswift:
Remove unused functions not covered by tests.

**Note to reviewers:** I'm sure we want to keep a subset of these around for consistency (preferably with the addition of one test per unused function). Let me know which we want to keep and I'll revert the removal of these.

### DIFF
--- a/bitcoin/signature.c
+++ b/bitcoin/signature.c
@@ -246,16 +246,6 @@ bool signature_from_der(const u8 *der, size_t len, secp256k1_ecdsa_signature *si
 						   sig, der, len);
 }
 
-/* Signature must have low S value. */
-bool sig_valid(const secp256k1_ecdsa_signature *sig)
-{
-	secp256k1_ecdsa_signature tmp;
-
-	if (secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &tmp, sig) == 0)
-		return true;
-	return false;
-}
-
 static char *signature_to_hexstr(const tal_t *ctx,
 				 const secp256k1_ecdsa_signature *sig)
 {
--- a/common/json.c
+++ b/common/json.c
@@ -187,44 +187,6 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
 	return NULL;
 }
 
-/* Guide is a string with . for members, [] around indexes. */
-const jsmntok_t *json_delve(const char *buffer,
-			    const jsmntok_t *tok,
-			    const char *guide)
-{
-	while (*guide) {
-		const char *key;
-		size_t len = strcspn(guide+1, ".[]");
-
-		key = tal_strndup(NULL, guide+1, len);
-		switch (guide[0]) {
-		case '.':
-			if (tok->type != JSMN_OBJECT)
-				return tal_free(key);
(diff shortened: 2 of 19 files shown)


common/json.c
```diff
@@ -187,44 +187,6 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
 	return NULL;
 }
 
-/* Guide is a string with . for members, [] around indexes. */
-const jsmntok_t *json_delve(const char *buffer,
```
ZmnSCPxj: I intend to eventually use this function in the pilotd, possibly.

wallet/wallet.c
```diff
@@ -1653,26 +1629,6 @@ wallet_payment_by_hash(const tal_t *ctx, struct wallet *wallet,
 	return payment;
 }
 
-struct secret *wallet_payment_get_secrets(const tal_t *ctx,
```
ZmnSCPxj: Thanks for catching this, yes this function is no longer used as `path_secrets` is now part of the `struct payment`.

common/json.c
rustyrussell: I vote we drop it now and remember it exists.  You can resurrect it later :)

bitcoin/signature.c
```diff
@@ -246,16 +246,6 @@ bool signature_from_der(const u8 *der, size_t len, secp256k1_ecdsa_signature *si
...
-/* Signature must have low S value. */
-bool sig_valid(const secp256k1_ecdsa_signature *sig)
-{
-	secp256k1_ecdsa_signature tmp;
-
-	if (secp256k1_ecdsa_signature_normalize(secp256k1_ctx, &tmp, sig) == 0)
-		return true;
-	return false;
-}
-
```
rustyrussell: This made me double-check: in fact, yes, libsecp256k1 only accepts low-s signatures, so this is redundant.

wallet/txfilter.c
```diff
@@ -108,14 +108,6 @@ bool outpointfilter_matches(struct outpointfilter *of, const struct bitcoin_txid
...
 }
 
-void outpointfilter_remove(struct outpointfilter *of, const struct bitcoin_txid *txid, const u32 outnum)
-{
-	struct outpointfilter_entry op;
-	op.txid = *txid;
-	op.outnum = outnum;
-	outpointset_del(of->set, &op);
-}
-
```
rustyrussell: @cdecker I assume this *should* be called from somewhere?

practicalswift:
Rebased. Now removes `strange_chars(…)` too.

wallet/txfilter.c
cdecker: I wanted to call this directly when marking them as spent, but then had to revert since reorgs would un-spend them again and then we need to track them again. I'll hook it up in the pruning after being buried deep enough. Feel free to remove for now, I'll re-add it when appropriate :-)

cdecker: Since this was a rather quick fix I opened PR #1279 addressing this. So the function is no longer unused :-p

practicalswift:
Updated. `outpointfilter_remove(…)` no longer removed. Please re-review :-)

PR #1277, opened 2018-03-25
Use correct type field for signed integers in format strings

practicalswift:
Use correct type field for signed integers in format strings.


ZmnSCPxj:
ACK 083bbe47125474c1a5f0914c167b4dd0fad77447

PR #1279, opened 2018-03-26
wallet: Remove pruned outpoints from utxoset outpoint_filter

cdecker:
As noted by @rustyrussell [here](https://github.com/ElementsProject/lightning/pull/1275#discussion_r176962952) I never got around to clear pruned outpoints from the filter, so this fixes that.


ZmnSCPxj:
ACK 20c1d132ab27d3a726233ae808736a441601425c

PR #1284, opened 2018-03-27
Remove unused functions not covered by unit tests

practicalswift:
Remove unused functions not covered by tests (round 2).

**Note to reviewers:** It could be the case that we want to keep a subset of these around for consistency (preferably with the addition of one test per unused function). Let me know which we want to keep and I'll revert the removal of these.


cdecker:
ACK 3e9b401

PR #1286, opened 2018-03-27
Add funding_address to listfunds RPC response, fixes #1227

### LINKED ISSUE #1227: `listfunds` should also display the address that onchain funds are in
Inspired by #1226

izgzhen:


izgzhen:
@ZmnSCPxj thanks for the pointer! code updated

wallet/walletrpc.c
```diff
@@ -369,13 +369,22 @@ static void json_listfunds(struct command *cmd, const char *buffer UNUSED,
...
 	json_array_start(response, "outputs");
 	for (size_t i = 0; i < tal_count(utxos); i++) {
 		json_object_start(response, NULL);
 		json_add_txid(response, "txid", &utxos[i]->txid);
 		json_add_num(response, "output", utxos[i]->outnum);
 		json_add_u64(response, "value", utxos[i]->amount);
+		bip32_pubkey(cmd->ld->wallet->bip32_base, &funding_pubkey,
+			     utxos[i]->keyindex);
+		pubkey_to_hash160(&funding_pubkey, &h160);
+		out = p2sh_to_base58(cmd,
```
ZmnSCPxj: Are all outputs P2SH-P2WPKH? I think some outputs are P2WPKH now?

izgzhen: Oh sorry, https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#funding-transaction-output  said that the funding output is P2WSH. Do you mean this one?

ZmnSCPxj: This code is inside the `outputs` array loop describing onchain addresses, not the funding transaction of `channels`.  My understanding is that we recently added the ability to receive onchain via P2WPKH addresses.  Also I think change addresses are P2WPKH.

ZmnSCPxj: See `enum wallet_output_type` in `wallet/wallet.h`.  Various outputs have different types that change the effective onchain address they will have.

izgzhen: @ZmnSCPxj if so, then maybe it is worth to add such `type` to `utxo` struct?

ZmnSCPxj: Maybe `is_p2sh` and `close_info` are enough to derive information on exact type of address? Will need to trace what `is_p2sh` means at least (does it differentiate between P2SH-P2WPKH vs P2WPKH?)

ZmnSCPxj: Investigating `wallet_stmt2output` in `wallet/wallet.c`, yes, `is_p2sh` is enough to differentiate P2SH-P2WPKH from the rest.  Probably `close_info` is for outputs that are not yet claimable (they need onchaind to claim it I think...?), so suppressing output of address field may be OK for outputs with non-NULL `close_info`, but I have not investigated deeply.

izgzhen:
more refinement is added. if `close_info` is not `NULL`, the address output will be suppressed. And if `is_p2sh` is true, then it will be encoded in `base58` as P2SH, or else it will be encoded as segwit P2WPKH address.

wallet/walletrpc.c
```diff
@@ -369,14 +369,36 @@ static void json_listfunds(struct command *cmd, const char *buffer UNUSED,
...
+	struct pubkey funding_pubkey;
 	json_object_start(response, NULL);
 	json_array_start(response, "outputs");
 	for (size_t i = 0; i < tal_count(utxos); i++) {
 		json_object_start(response, NULL);
 		json_add_txid(response, "txid", &utxos[i]->txid);
 		json_add_num(response, "output", utxos[i]->outnum);
 		json_add_u64(response, "value", utxos[i]->amount);
-
+		// @closed_info is for outputs that are not yet claimable
```
ZmnSCPxj: `closed_info` misspelled, also, we use `/* */` comments.

cdecker: Fixed this quickly myself :-)

ZmnSCPxj:
ACK 694c6d8a42748f9a5014dbf73e04bc3e64757705

cdecker:
I'm a bit confused by the name `funding_address`, could you ELI5 that for me?

ZmnSCPxj:
Should be `address` for onchain funds.

Also the channel-level `funding_address` is wrong.  Maybe just remove from channels? We can locate the funding transaction and funding outpoint already.

cdecker:
Thanks @ZmnSCPxj, so it's just the address the funds are on, that's a useful addition indeed. Agreed on removing the "address" from the channels, since we want to prevent users from accidentally sending funds there believing that it's their address or that they can "top up", which would result in the funds being locked up in a multisig...

wallet/walletrpc.c
```diff
@@ -407,6 +429,11 @@ static void json_listfunds(struct command *cmd, const char *buffer UNUSED,
 				     c->funding_satoshi);
 			json_add_txid(response, "funding_txid",
 				      &c->funding_txid);
+			derive_basepoints(&c->seed, &funding_pubkey, NULL, NULL, NULL);
+			pubkey_to_hash160(&funding_pubkey, &h160);
+			out = p2sh_to_base58(cmd,
+				get_chainparams(cmd->ld)->testnet, &h160);
+			json_add_string(response, "funding_address", out);
```
cdecker: Please remove this since it may lead people into thinking that it's one of their addresses. Sending to this address will lock up funds.

cdecker:
LGTM

ACK 1609819

PR #1295, opened 2018-03-28
Tmpctx cleanups

rustyrussell:


rustyrussell:
True, we could do that.

common/io_debug.c
```diff
@@ -1,17 +0,0 @@
-#include <ccan/err/err.h>
```
ZmnSCPxj: This was removed but the `lightningd/Makefile` `channeld/Makefile` still seems to use it?

rustyrussell: Thanks!  Fixed and rebased...

ZmnSCPxj:
ACK 4bde63993e6ee94b44134943866672fdc07e2351

Looks fine to me, although, there is a valgrind warn in a CI run, but I feel it could be rerun?

ZmnSCPxj:
ACK 3c3fcf0bc0879c968af0d729c08876c82ffca91c

PR #1301, opened 2018-03-29
.gitignore: Ignore `gossip_store` file.

ZmnSCPxj:
This, is generated, by `make check`.
Probably better to put it in some location elsewhere, though?


cdecker:
ACK 7c79386

PR #1302, opened 2018-03-29
Drive-by fixes

cdecker:
Just two things I noticed while doing other stuff. One should get us slightly more robust results when testing  and the other was failing my CI builds.


ZmnSCPxj:
ACK 472554412975a84f6372293e2b1b93ad682ff310

cdecker:
Removed the invoice manpage update, since it was changed in a recent PR. Will merge the trivial fix once CI is happy.

PR #1303, opened 2018-03-29
gossip: Monitor blockchain for funding txs being spent and remove channels

### LINKED ISSUE #1250: Noisy logs due to announcements from fully-spent channels

The Blockstream Store node at `02f6725f9c1c40333b67faea92fd211c183050f28df32cac3f9d69685fe9665432@104.198.32.198:9735` continually announces / sends out advertisements for channels that have already been fully spent.

I'm running an LND node, and my logs are littered with lines like this:

```
2018-03-19 18:57:50.937 [ERR] DISC: Router rejected channel edge: unable to fetch utxo for chan_id=565220444960129024, chan_point=bacd9978f199f84224fd5a86d61a2c744f71e3a88752a3acc46c0c592eaabb31:0: 
target output has been spent
2018-03-19 18:57:51.289 [ERR] DISC: Router rejected channel edge: unable to fetch utxo for chan_id=565220444960129024, chan_point=bacd9978f199f84224fd5a86d61a2c744f71e3a88752a3acc46c0c592eaabb31:0: target output has been spent
2018-03-19 18:57:52.051 [ERR] DISC: Router rejected channel edge: unable to fetch utxo for chan_id=565220444960653312, chan_point=303d599f19022d8df4c464376e36895019c73badae11e5b6c2f21e7e0075bcc5:0: target output has been spent
```
Seems like a bug on your guys' side of things, but I'm not 100% as I'm pretty new to LN. Thanks for having a look.

cdecker:
Watch for potential channel closures and notify `gossipd` about them so we can remove the channels from our local network view. Since we store the announcements and replay them during a restart, we also need to remember that we deleted the channel. We do that by simply adding a tombstone with the scid and replay that as well during startup.

Fixes #1250


cdecker:
The first two commits are from #1302.

gossipd/gossip_store.c
```diff
@@ -135,6 +144,9 @@ bool gossip_store_read_next(struct routing_state *rstate,
 	} else if(type == WIRE_GOSSIP_STORE_NODE_ANNOUNCEMENT) {
 		fromwire_gossip_store_node_announcement(msg, msg, &gossip_msg);
 		routing_add_node_announcement(rstate, gossip_msg);
+	} else if(type == WIRE_GOSSIP_STORE_CHANNEL_DELETE) {
+		fromwire_gossip_store_channel_delete(msg, &scid);
+		tal_free(get_channel(rstate, &scid));
```
ZmnSCPxj: Can `get_channel` return `NULL`?

Consider the case below:

1.  On a previous run, we added a channel, then removed it by this mechanism.  Both events are stored in the `gossip_store`.
2.  We restart and add a channel by `gossip_store`.
3.  Chaintopology catches up and notices the channel is spent.  It causes gossipd to delete the channel.
4.  We delete the channel by `gossip_store`.  At this point, `get_channel` returns `NULL`.

`tal_free` works on `NULL` though (does nothing if `NULL`).  But is the above case handled correctly?

ZmnSCPxj:
ACK 0dc5d8cd254e347031da1d9bb74e89c29afb4df6

Some commits from #1302 seem included though

gossipd/gossip_store.c
cdecker: The situation in which we open a local channel, then spend it before it gets announced may end up in the situation you described. I'm indeed relying on the property that `tal_free(NULL)` will work in this case :wink:

For a non-local channel it is not possible that a channel is deleted without a matching announcement in the store. But I decided not to differentiate the two cases, since this is just simpler.

ZmnSCPxj:
I somewhat belatedly remembered, that, a routing failure with `PERM` bit set will also delete channels from the routemap. Should we store this in the `gossip_store`?

https://github.com/ElementsProject/lightning/blob/e41d5651031c7e23caf9db8c33b14208eb3c41da/gossipd/routing.c#L1286-L1305

What happens when:

1.  Somebody else has a channel it wants to close, because reasons.
2.  We route via that doomed channel, which triggers a `PERMANENT_CHANNEL_FAILURE` routing failure (which has `PERM` bit set, so the channel is deleted from routemap by `routing_failure_channel_out`).
3.  We see the funding tx output getting spent.  This triggers `handle_outpoint_spent`, which cannot find the channel in the routemap, and thus does NOT save the channel deletion in the store.
4.  We restart.  The channel announcement remains in the gossip store but is not deleted afterwards.
5.  Our eventual goal is not to restart chaintopology too early. Assuming that is successfully done, the previous `handle_outpoint_spent` will not trigger after the restart (chaintopology already triggered it pre-restart) and the channel remains in the routemap! (admittedly attempting to pay via that channel in a route will trigger `UNKNOWN_NEXT_PEER`, which also has the `PERM` bit set, so we can recover later...)

cdecker:
Good point, we can just add the tombstone in the store and it'll add and immediately delete the channel again. Since we also apply the delete to our in-memory view, it should also not get re-added to the store on compaction, which will read the in-memory queue.

PR #1304, opened 2018-03-30
Node statistics

ZmnSCPxj:
Add node-level statistics for tracking various events:

1.  Time a node was first seen.
2.  Time a node was last seen (including successful payment attempts and successful forwards, and `node_announcement`)
3.  Number of times a node was unable to receive a payment for forwarding or final payment (`forwarding_failures`).
4.  Number of times we were unable to reach a node for connecting (`connect_failures`).
5.  Number of times a channel with a node was `error`ed by either side (`channel_failures`).

Intent is usage with automated programmable systems for managing channels ("pilot").  Hopefully node statistics can be used to judge appropriateness of a node for channeling with.

----

I am somewhat unsatisfied with how I send information that a `node_announcement` was accepted from gossipd to masterd.  What I did, was, enqueue public keys to an array in `routing_state`, and every 1 second grab the current array and send them all to masterd.  Asking also, for suggestion of alternative for this system.

### DIFF
--- a/lightningd/nodestatsrpc.c
+++ b/lightningd/nodestatsrpc.c
@@ -1,0 +1,77 @@
+#include <ccan/autodata/autodata.h>
+#include <common/json.h>
+#include <lightningd/json.h>
+#include <lightningd/jsonrpc.h>
+#include <lightningd/lightningd.h>
+#include <wallet/wallet.h>
+
+static void
+json_add_nodestats(struct json_result *result, char const *field,
+		   const struct nodestats_detail *detail)
+{
+	json_object_start(result, field);
+	json_add_u64(result, "index", detail->index);
+	json_add_pubkey(result, "nodeid", &detail->nodeid);
+	json_add_u64(result, "time_first_seen", detail->time_first_seen);
+	json_add_u64(result, "time_last_seen", detail->time_last_seen);
+	json_add_num(result, "forwarding_failures", detail->forwarding_failures);
+	json_add_num(result, "connect_failures", detail->connect_failures);
+	json_add_num(result, "channel_failures", detail->channel_failures);
+	json_object_end(result);
+}
+
+static void
+json_listnodestats(struct command *cmd,
+		   const char *buffer,
+		   const jsmntok_t *params)
+{
+	struct wallet *wallet = cmd->ld->wallet;
+	jsmntok_t *idtok;
+	struct pubkey id;
+	struct json_result *result;
+	struct nodestats_detail detail;
+	u64 it;
+
+	if (!json_get_params(cmd, buffer, params,
+			     "?id", &idtok,
+			     NULL)) {
+		return;
+	}
+
+	if (idtok && !json_tok_pubkey(buffer, idtok, &id)) {
+		command_fail(cmd, "id is not a public key: '%.*s'",
+			     idtok->end - idtok->start,
+			     buffer + idtok->start);
+		return;
+	}
+
+	result = new_json_result(cmd);
+
+	json_object_start(result, NULL);
+	json_array_start(result, "nodestats");
+	if (idtok) {
+		if (wallet_nodestats_get_by_pubkey(wallet, &detail, &id))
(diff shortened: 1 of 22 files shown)


ZmnSCPxj:
Sorry, made wrong command while pushing a rebase...

wallet/db.c
```diff
@@ -270,6 +270,17 @@ char *dbmigrations[] = {
     /* Record the msatoshi actually sent in a payment. */
     "ALTER TABLE payments ADD msatoshi_sent INTEGER;",
     "UPDATE payments SET msatoshi_sent = msatoshi;",
+    /* -- Node statistics -- */
+    "CREATE TABLE node_statistics"
+    "     ( id INTEGER PRIMARY KEY AUTOINCREMENT"
+    "     , nodeid BLOB UNIQUE"
```
cdecker: I've always used the commas as the last element on the line.

lightningd/nodestatsrpc.c
```diff
@@ -0,0 +1,73 @@
...
+			     buffer + idtok->start);
+		return;
+	}
+
+	result = new_json_result(cmd);
+
+	json_object_start(result, NULL);
+	json_array_start(result, "nodestats");
+	if (idtok) {
+		if (wallet_nodestats_get_by_pubkey(wallet, &detail, &id))
```
cdecker: May return empty despite the caller asking for a specific ID, could return an error instead.

lightningd/pay.c
```diff
@@ -179,9 +179,36 @@ sendpay_result_in_progress(const tal_t *ctx,
 	return result;
 }
 
+/* Report to wallet nodestats that the specified nodes are
+ * alive. */
+static void
+report_nodestats_alive(struct wallet *wallet,
+		       const struct pubkey *nodes,
+		       int num_nodes)
```
cdecker: I think we're using a `tal_arr` here anyway, so we could use `tal_count`.

wallet/db.c
ZmnSCPxj: I prefer the commas as the first element, since the usual modification is to add new items at the end of sequence.

Consider a sequence like the below:

    FOO A,
        B

When we add C, we also modify the previous line for B even if in principle it should not need modification:

    FOO A,
        B,
        C

But with parentheses at the start of lines:

    FOO A
      , B

Then we add a new C, the line for B is unmodified:

    FOO A
      , B
      , C

lightningd/nodestatsrpc.c
ZmnSCPxj: ok

lightningd/pay.c
ZmnSCPxj: ok

ZmnSCPxj: On checking again, no.  On routing failure, we know that all nodes up to the erring node are alive (else we would not have received a routing failure onion reply), but not any of the succeeding nodes.  So we call this function with `num_nodes < tal_count(nodes)`.

We would not be able to receive a remote routing failure if for instance the first hop failed (we would be the node that raises the error to ourself, and would not call this code).  But actually receiving a remote routing failure means that every node on the route, up to the node that raised the error, is alive, so we should mark them as being seen as alive in our stats.

rustyrussell:
I'm afraid this PR is a case of typing first, designing later.  That's *particularly* problematic for stats, since you *never* have the statistics you want: you might want to time-decay the statistics, for example.

The failure stats are going to be useless unless you're making thousands of payments yourself, for example.  The "alive" stats are useless because you're only supposed to send an update once a week, etc.

With all statistics, your best option is to present the raw data and have the user create stats they want.  Where that isn't possible, you need to justify how each statistic would inform the user: that tells you whether you want a total, a count, a mean, a median, a min/max, a decaying average, etc.

IOW, this needs a user before we build it.  They need to be designed together.

Meanwhile, we should have a command showing all HTLCs we added/failed on a channel.  We should also have whole-node totals (succeeded, failed, total amount xferred, total fees collected), because channels will eventually get pruned.

PR #1311, opened 2018-04-02
Adjust loglevel for error message "Failed to get peername for incoming conn"

practicalswift:
Adjust loglevel for error message "Failed to get peername for incoming conn".

I've seen this triggered on testnet. Likely not of interest for an end-user running with the default loglevel.


cdecker:
ACk cd1eb52

Issue #1308, opened 2018-04-01
lightningd keeps crashing with FATAL SIGNAL 6 RECEIVED


robtex:
### Issue and Steps to Reproduce
```
2018-04-01T05:28:42.394Z lightningd(1159): lightning_channeld-027eae72368a57dc9d0c26de4fce5637117e7968f422bdca3d169704fadf8b436b chan #957: peer_out WIRE_SHUTDOWN             2018-04-01T05:28:42.394Z lightning_gossipd(1167): Received channel_update for channel 515454:1348:0(1) now ACTIVE                                                              2018-04-01T05:28:42.394Z lightningd(1159): lightning_channeld-027eae72368a57dc9d0c26de4fce5637117e7968f422bdca3d169704fadf8b436b chan #957: UPDATE WIRE_CHANNEL_SHUTDOWN_COMPLETE                                                                                                                                                                             2018-04-01T05:28:42.394Z lightning_gossipd(1167): Received channel_update for channel 515419:1416:0(0) now ACTIVE                                                              2018-04-01T05:28:42.394Z lightningd(1159): lightning_channeld-027eae72368a57dc9d0c26de4fce5637117e7968f422bdca3d169704fadf8b436b chan #957: UPDATE WIRE_CHANNEL_SHUTDOWN_COMPLETE                                                                                                                                                                             2018-04-01T05:28:42.410Z lightningd(1159): lightning_closingd-027eae72368a57dc9d0c26de4fce5637117e7968f422bdca3d169704fadf8b436b chan #957: pid 17724, msgfd 19                2018-04-01T05:28:42.426Z lightningd(1159): 027eae72368a57dc9d0c26de4fce5637117e7968f422bdca3d169704fadf8b436b chan #957: State changed from CHANNELD_AWAITING_LOCKIN to CLOSINGD_SIGEXCHANGE
2018-04-01T05:28:42.426Z lightningd(1159): channel state CHANNELD_AWAITING_LOCKIN should be CHANNELD_SHUTTING_DOWN
2018-04-01T05:28:42.426Z lightningd(1159): FATAL SIGNAL 6 RECEIVED
2018-04-01T05:28:42.470Z lightningd(1159): backtrace: lightningd/log.c:494 (log_crash) 0x415f89
2018-04-01T05:28:42.470Z lightningd(1159): backtrace: (null):0 ((null)) 0x7fdb982e24af
[...]

PR #1313, opened 2018-04-03
Shutdown while awaiting lockin

rustyrussell:


ZmnSCPxj:
ACK ed767c86647b78cb4a1e4dab15ded07569fc2b5a

Does this address #1308?

lightningd/channel_control.c
```diff
@@ -77,9 +77,9 @@ static void peer_got_shutdown(struct channel *channel, const u8 *msg)
 	}
 
 	/* If we weren't already shutting down, we are now */
-	if (channel->state == CHANNELD_NORMAL)
+	if (channel->state != CHANNELD_SHUTTING_DOWN)
```
cdecker: Might a bit too permissive now? We probably also don't want to accept in state sigexchange.

ZmnSCPxj:
I think the PR makes the state machine transition to `CHANNELD_SHUTTING_DOWN` from `CHANNELD_AWAITING_LOCKIN`, so that the code you quoted becomes valid.

lightningd/channel_control.c
rustyrussell: If we're in channeld (which we are), we have to be in AWAITING_LOCKIN or NORMAL or SHUTTING_DOWN.  So this is actually safe.

rustyrussell:
It's channeld which tells us that it received shutdown, so it can't die at the same time?

The order is:
1. channeld tells us it got the remote shutdown, we save the scriptpubkey for closing and update state to CHANNELD_SHUTTING_DOWN if not already done.
2. When channeld has cleared all the HTLCs, it tells us shutdown is complete, we update state to CLOSINGD_SIGEXCHANGE and start closingd (killing channeld in the process).

Note that if we reconnect, the peer can re-send shutdown so we need to handle the case where we're already in CHANNELD_SHUTTING_DOWN.

We could eliminate the CHANNELD_SHUTTING_DOWN state and use the existence of the remote closing scriptpubkey as a flag, but as shutdown can be extended, I think having the state is clearer to users.

rustyrussell:
Oh, and on test order: I've taken to leaving the test-then-fix order in my pushes, using unittest.expectedFail() and removing it in the fix.  That proves that I actually tested it before and after :)

PR #1187, opened 2018-03-07
Add usable dockerfile to clightning

NicolasDorier:
@cdecker, you are providing a dockerfile on https://hub.docker.com/r/cdecker/lightningd .
Sadly, this docker file is useless for developers as it is impossible to connect to the RPC unix domain socket.

This PR propose a `Dockerfile` which expose clightning on TCP with `socat`. This allow developers not using linux to still develop stuff on top of CLightning by dockerizing the app and connecting via TCP.

It will also simplify `lightning-charge` which currently is starting clightning and charge in the same container though a very hacky manner.

This port is not meant to be mapped to the host (except for testing) for obvious security reasons.

PR #1318, opened 2018-04-03
Usable Dockerfile for CLightning

NicolasDorier:
I already did a PR in the past with this, but decided to close it after there was complains about `socat` dependency.

This PR makes a second attempt, this time, `socat` is disabled by default. (it is used only for dev purposes)
It ships `litecoin-cli` and `bitcoin-cli` with it.

Environment variable:
* `EXPOSE_TCP` default to false, if true, use socat to expose clightning on port `9835`.
* `LIGHTNINGD_OPT` is the config file of the clightning instance.

A new parameter, not existing in clightning parameters called `chain=btc|ltc` has been added and is interpreted in the `entrypoint.sh`, it will transform the `network=` parameter with the following table if present:

chain + network = replaced network
-----------
btc + mainnet = bitcoin
btc + testnet = testnet
btc + regtest = regtest
ltc + mainnet = litecoin
ltc + testnet = litecoin-testnet

This has been done to uniformise the  way of identifying the chain to use across different docker images I am using.

Note that this image is used in production by BTCPay (https://hub.docker.com/r/nicolasdorier/clightning/), so I don't strictly need it in the lightning repository. But I think it has better its place here than in my fork. (I rebase this PR/docker image one time per week)

Using `EXPOSE_TCP` allow the contributors to debug BTCPay with clightning integration whatever the OS they are using.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,104 @@
+FROM alpine:3.7 as builder
+
+RUN apk add --no-cache \
+     ca-certificates \
+     autoconf \
+     automake \
+     build-base \
+     libressl \
+     libtool \
+     gmp-dev \
+     python \
+     python-dev \
+     python3 \
+     sqlite-dev \
+     wget \
+     git \
+     file \
+     gnupg \
+     swig
+
+WORKDIR /opt
+
+ENV BITCOIN_VERSION 0.16.0
+ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz
+ENV BITCOIN_SHA256 e6322c69bcc974a29e6a715e0ecb8799d2d21691d683eeb8fef65fc5f6a66477
+ENV BITCOIN_ASC_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/SHA256SUMS.asc
+ENV BITCOIN_PGP_KEY 01EA5486DE18A882D4C2684590C8019E36C2E964
+
+RUN mkdir /opt/bitcoin && cd /opt/bitcoin \
+    && wget -qO bitcoin.tar.gz "$BITCOIN_URL" \
+    && echo "$BITCOIN_SHA256  bitcoin.tar.gz" | sha256sum -c - \
+    && gpg --keyserver keyserver.ubuntu.com --recv-keys "$BITCOIN_PGP_KEY" \
+    && wget -qO bitcoin.asc "$BITCOIN_ASC_URL" \
+    && gpg --verify bitcoin.asc \
+    && BD=bitcoin-$BITCOIN_VERSION/bin \
+    && tar -xzvf bitcoin.tar.gz $BD/bitcoin-cli --strip-components=1 \
+    && rm bitcoin.tar.gz
+
+ENV LITECOIN_VERSION 0.14.2
+ENV LITECOIN_URL https://download.litecoin.org/litecoin-0.14.2/linux/litecoin-0.14.2-x86_64-linux-gnu.tar.gz
+ENV LITECOIN_SHA256 05f409ee57ce83124f2463a3277dc8d46fca18637052d1021130e4deaca07b3c
+ENV LITECOIN_ASC_URL https://download.litecoin.org/litecoin-0.14.2/linux/litecoin-0.14.2-linux-signatures.asc
+ENV LITECOIN_PGP_KEY FE3348877809386C
+
+# install litecoin binaries
+RUN mkdir /opt/litecoin && cd /opt/litecoin \
+    && wget -qO litecoin.tar.gz "$LITECOIN_URL" \
+    && echo "$LITECOIN_SHA256  litecoin.tar.gz" | sha256sum -c - \
+    && gpg --keyserver keyserver.ubuntu.com --recv-keys "$LITECOIN_PGP_KEY" \
+    && wget -qO litecoin.asc "$LITECOIN_ASC_URL" \
+    && gpg --verify litecoin.asc \
+    && BD=litecoin-$LITECOIN_VERSION/bin \
+    && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
+    && rm litecoin.tar.gz
+
+ENV LIGHTNINGD_VERSION=master
+
+WORKDIR /opt/lightningd
+COPY . .
+
+ARG DEVELOPER=0
+RUN make -j3 DEVELOPER=${DEVELOPER} && cp lightningd/lightning* cli/lightning-cli /usr/bin/
+
+FROM alpine:3.7
+
+RUN apk add --no-cache \
+     gmp-dev \
+     sqlite-dev \
+     inotify-tools \
+     socat \
+     bash
+
+ENV GLIBC_VERSION 2.27-r0
+ENV GLIBC_SHA256 938bceae3b83c53e7fa9cc4135ce45e04aae99256c5e74cf186c794b97473bc7
+ENV GLIBCBIN_SHA256 3a87874e57b9d92e223f3e90356aaea994af67fb76b71bb72abfb809e948d0d6
+# Download and install glibc (https://github.com/jeanblanchard/docker-alpine-glibc/blob/master/Dockerfile)
+RUN apk add --update curl && \
+  curl -Lo /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
+  curl -Lo glibc.apk "https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-${GLIBC_VERSION}.apk" && \
+  echo "$GLIBC_SHA256  glibc.apk" | sha256sum -c - && \
+  curl -Lo glibc-bin.apk "https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-bin-${GLIBC_VERSION}.apk" && \
+  echo "$GLIBCBIN_SHA256  glibc-bin.apk" | sha256sum -c - && \
+  apk add glibc-bin.apk glibc.apk && \
+  /usr/glibc-compat/sbin/ldconfig /lib /usr/glibc-compat/lib && \
+  echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
+  apk del curl && \
+  rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*
+
+ENV LIGHTNINGD_DATA=/root/.lightning
+ENV LIGHTNINGD_PORT=9835
+ARG DEVELOPER=0
+
+RUN mkdir $LIGHTNINGD_DATA && \
+    touch $LIGHTNINGD_DATA/config
+VOLUME [ "/root/.lightning" ]
+
+COPY --from=builder /opt/lightningd/cli/lightning-cli /usr/bin
+COPY --from=builder /opt/lightningd/lightningd/lightning* /usr/bin/
+COPY --from=builder /opt/bitcoin/bin /usr/bin
+COPY --from=builder /opt/litecoin/bin /usr/bin
+COPY tools/docker-entrypoint.sh entrypoint.sh
+
+EXPOSE 9735 9835
+ENTRYPOINT  [ "./entrypoint.sh" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,9 +70,16 @@ RUN apk add --no-cache \
      socat \
      bash
 
+ARG TRACE_TOOLS=false
+ENV TRACE_TOOLS=$TRACE_TOOLS
+
 ENV GLIBC_VERSION 2.27-r0
 ENV GLIBC_SHA256 938bceae3b83c53e7fa9cc4135ce45e04aae99256c5e74cf186c794b97473bc7
 ENV GLIBCBIN_SHA256 3a87874e57b9d92e223f3e90356aaea994af67fb76b71bb72abfb809e948d0d6
+
+ENV TRACE_LOCATION=/opt/traces
+VOLUME /opt/traces
+
 # Download and install glibc (https://github.com/jeanblanchard/docker-alpine-glibc/blob/master/Dockerfile)
 RUN apk add --update curl && \
   curl -Lo /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub && \
@@ -83,12 +90,23 @@ RUN apk add --update curl && \
   apk add glibc-bin.apk glibc.apk && \
   /usr/glibc-compat/sbin/ldconfig /lib /usr/glibc-compat/lib && \
   echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
+  ( ! $TRACE_TOOLS || \
+    ( \
+        sed -i -e 's/v[[:digit:]]\.[[:digit:]]/edge/g' /etc/apk/repositories && \
+        echo "http://nl.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories && \
+        apk add --no-cache perf perl && \
+        mkdir FlameGraph && cd FlameGraph && \
+        curl -Lo FlameGraph.tar.gz "https://github.com/brendangregg/FlameGraph/archive/v1.0.tar.gz" && \
+        tar -zxvf FlameGraph.tar.gz --strip-components=1 && rm FlameGraph.tar.gz && cd .. \
+    ) \
+  ) && \
   apk del curl && \
   rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*
 
 ENV LIGHTNINGD_DATA=/root/.lightning
 ENV LIGHTNINGD_PORT=9835
 ARG DEVELOPER=0
+ENV DEVELOPER=$DEVELOPER
 
 RUN mkdir $LIGHTNINGD_DATA && \
     touch $LIGHTNINGD_DATA/config
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,7 +106,6 @@ RUN apk add --update curl && \
   rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*
 
 ENV LIGHTNINGD_DATA=/root/.lightning
-ENV LIGHTNINGD_PORT=9835
 
 RUN mkdir $LIGHTNINGD_DATA && \
     touch $LIGHTNINGD_DATA/config
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,8 @@ RUN apk add --no-cache \
      git \
      file \
      gnupg \
-     swig
+     swig \
+     zlib-dev
 
 WORKDIR /opt
 
@@ -59,7 +60,7 @@ WORKDIR /opt/lightningd
 COPY . .
 
 ARG DEVELOPER=0
-RUN make -j3 DEVELOPER=${DEVELOPER} && cp lightningd/lightning* cli/lightning-cli /usr/bin/
+RUN ./configure && make -j3 && cp lightningd/lightning* cli/lightning-cli /usr/bin/
 
 FROM alpine:3.7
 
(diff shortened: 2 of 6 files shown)


.dockerignore
```diff
@@ -0,0 +1 @@
+Dockerfile
```
practicalswift: Nit: Missing `\n`

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+ENV LITECOIN_PGP_KEY FE3348877809386C
+
+# install litecoin binaries
+RUN mkdir /opt/litecoin && cd /opt/litecoin \
+	&& wget -qO litecoin.tar.gz "$LITECOIN_URL" \
+	&& echo "$LITECOIN_SHA256 litecoin.tar.gz" | sha256sum -c - \
+	&& gpg --keyserver keyserver.ubuntu.com --recv-keys "$LITECOIN_PGP_KEY" \
+	&& wget -qO litecoin.asc "$LITECOIN_ASC_URL" \
+	&& gpg --verify litecoin.asc \
+    && BD=litecoin-$LITECOIN_VERSION/bin \
```
practicalswift: Nit: Inconsistent indentation.

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+VOLUME [ "/root/.lightning" ]
+
+COPY --from=builder /opt/lightningd/cli/lightning-cli /usr/bin
+COPY --from=builder /opt/lightningd/lightningd/lightning* /usr/bin/
+COPY --from=builder /opt/bitcoin/bin /usr/bin
+COPY --from=builder /opt/litecoin/bin /usr/bin
+COPY tools/docker-entrypoint.sh entrypoint.sh
+
+EXPOSE 9112 9735 9835
+ENTRYPOINT  [ "./entrypoint.sh" ]
```
practicalswift: Nit: Missing `\n`

practicalswift:
Please fix the linting warnings reported by `shellcheck tools/docker-entrypoint.sh` :-)

practicalswift:
Concept ACK from me :-)

NicolasDorier:
Cool! Ping @cdecker does it change your mind since https://github.com/ElementsProject/lightning/pull/1187 ?

cdecker:
Thanks @NicolasDorier this is indeed much better :smiley: Making `socat` optional is much better I think.

I'd prefer to have these files in the `contrib` directory though, since that is where we stash most of the auxiliary files. That makes `COPY . .` harder, but then again I'd prefer having the builder image just checkout from `git` which should guarantee relative stability of the compiled binary. Alternatively we can write a `Makefile` that tars up the files from `git ls-files` and copies that tarball in.

NicolasDorier:
This docker is meant to be integrated with docker hub automated build. This mean that everytimes you commit, dockerhub pick it up, build from the sources and apply a tag to it.

Using `COPY . .` automated build is building exactly the right commit. It makes it also easy for someone to tweak clightning code then build themselves their own images and test again. (no need to git push / tweak the dockerfile before doing a `docker build`)

analogic:
Would not be better to use debian:stretch-slim? System itself it is same as ubuntu but resulting image will be almost 30MB smaller...

NicolasDorier:
I am not against it, just not a linux expert and I don't really have time to hunt about which package name / tooling change between distro. :(

analogic:
Technicaly Ubuntu extends Debian, tooling and packages are mostly same. This will do the trick (I've tested build):

```diff
1c1
< FROM ubuntu:16.04 as builder
---
> FROM debian:stretch-slim as builder
4c4
<   libsqlite3-dev python python3 wget
---
>   libsqlite3-dev python python3 wget gnupg dirmngr
42c42
< FROM ubuntu:16.04
---
> FROM debian:stretch-slim
```

NicolasDorier:
Thanks @analogic , trying and testing this now.

NicolasDorier:
Fixed spellcheck (travis was quite useful), added and tested with debian, it works fine.

practicalswift:
@NicolasDorier Do you mean `shellcheck`? :-)

NicolasDorier:
@cdecker if you concept ACK the `COPY . .` (which is unavoidable if you want to leverage automated build on docker hub), I'll work on adding a good documentation in the README about this.

cdecker:
Yeah, I see your point, just a bit hesitant to put stuff in the root of the project :-)

Concept ACK

NicolasDorier:
Just added documentation in the `README.md` to teach how to use this docker file.

NicolasDorier:
Note that it assume you have setup automated build on docker hub on elementsproject/lightning !

tools/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+    echo "C-Lightning starting"
+    while read i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
+    < <(inotifywait  -e create,open --format '%f' --quiet "$LIGHTNINGD_DATA" --monitor)
+    echo "C-Lightning started"
+
+
+    socat -d -d "TCP4-listen:$LIGHTNINGD_PORT,fork,reuseaddr" "UNIX-CONNECT:$LIGHTNINGD_DATA/lightning-rpc"
+else
+    lightningd
+fi
```
practicalswift: Nit: Missing `\n`

tools/docker-entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ "$EXPOSE_TCP" == "true" ]; then
+
+    lightningd &
+
+    echo "C-Lightning starting"
+    while read i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
+    < <(inotifywait  -e create,open --format '%f' --quiet "$LIGHTNINGD_DATA" --monitor)
+    echo "C-Lightning started"
+
+
```
practicalswift: Nit: Remove redundant `\n`

README.md
```diff
@@ -63,14 +63,78 @@ For the impatient here's the gist of it for Ubuntu and Debian:
...
+| ltc | mainnet | litecoin |
+| ltc | testnet | litecoin-testnet |
+
+Here is an example of a docker-compose file with litecoind and c-lightning on `testnet` which expose litecoin's rpc interface on port `32610` and c-lightning API on port `32611`:
+
+```
+version: "3"
+
+  litecoind:
+    image: nicolasdorier/docker-litecoin:0.15.1
```
cdecker: Why litecoin?

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    touch $LIGHTNINGD_DATA/config
+VOLUME [ "/root/.lightning" ]
+
+COPY --from=builder /opt/lightningd/cli/lightning-cli /usr/bin
+COPY --from=builder /opt/lightningd/lightningd/lightning* /usr/bin/
+COPY --from=builder /opt/bitcoin/bin /usr/bin
+COPY --from=builder /opt/litecoin/bin /usr/bin
+COPY tools/docker-entrypoint.sh entrypoint.sh
+
+EXPOSE 9112 9735 9835
```
shesek: `9112` is Lightning Charge's port, this shouldn't be here

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+ENV BITCOIN_PGP_KEY 01EA5486DE18A882D4C2684590C8019E36C2E964
+RUN mkdir /opt/bitcoin && cd /opt/bitcoin \
+    && wget -qO bitcoin.tar.gz "$BITCOIN_URL" \
+    && echo "$BITCOIN_SHA256 bitcoin.tar.gz" | sha256sum -c - \
+    && gpg --keyserver keyserver.ubuntu.com --recv-keys "$BITCOIN_PGP_KEY" \
+    && wget -qO bitcoin.asc "$BITCOIN_ASC_URL" \
+    && gpg --verify bitcoin.asc \
+    && BD=bitcoin-$BITCOIN_VERSION/bin \
+    && tar -xzvf bitcoin.tar.gz $BD/bitcoin-cli --strip-components=1 \
+    && rm bitcoin.tar.gz
```
shesek: Why is the `rm` (here and below for litecoin) needed when its using a multi-stage build?

NicolasDorier: Note this has no impact if you download the full image. Only if you build it yourself without asking to clean the intermediate layers.

tools/docker-entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+if [[ $REPLACEDNETWORK ]]; then
+    sed -i '/^network=/d' "$LIGHTNINGD_DATA/config"
+    echo "network=$REPLACEDNETWORK" >> "$LIGHTNINGD_DATA/config"
+    echo "Replaced network $NETWORK by $REPLACEDNETWORK in $LIGHTNINGD_DATA/config"
+fi
+
+if [ "$EXPOSE_TCP" == "true" ]; then
+
```
practicalswift: Nit: Remove blank line above to improve readability.

tools/docker-entrypoint.sh
```diff
@@ -0,0 +1,52 @@
...
+    lightningd &
+
+    echo "C-Lightning starting"
+    while read i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
```
practicalswift: Nit: `SC2162: read without -r will mangle backslashes.`

See https://github.com/koalaman/shellcheck/wiki/SC2162 for details.

NicolasDorier:
Add the build-arg DEVELOPER.
I intend next to develop a docker hub hook so docker hub build automatically a DEVELOPER version when a version is tagged with `-dev` at the end.

NicolasDorier:
Pushed a new commit so you can make docker hub build developer version of the image via by tagging.

```
git tag -a "v0.0.0.5" -m "v0.0.0.5"
git tag -a "v0.0.0.5-dev" -m "v0.0.0.5-dev"
git push --tags
```

Built automatically this: https://hub.docker.com/r/nicolasdorier/clightning/tags/

With this docker hub config

image

NicolasDorier:
argh! nit fixed. :p

NicolasDorier:
Travis finally happy!

cdecker:
Thanks @NicolasDorier, excellent work! I'm a bit hesitant to merge it before the 0.6 release though. Would you be ok with holding this PR in the queue until we have finalized the 0.6 release? I have gone ahead and registered the elementsproject org on the docker hub, so that should work.

NicolasDorier:
Sure no problem, I am using my own docker hub for now.

analogic:
Thanks @NicolasDorier awesome work! I've started using image with https://github.com/analogic/clightning-admin (see https://github.com/analogic/clightning-admin/blob/master/docker-compose.mainnet.yml). I will switch to official image once merged

NicolasDorier:
Trying to move to Alpine3.7

NicolasDorier:
Getting an error trying to compile on alpine

```
In file included from ./bitcoin/block.h:5:0,
                 from ./wire/wire.h:4,
                 from ./common/sphinx.h:12,
                 from devtools/onion.c:5:
ccan/ccan/endian/endian.h:116:0: error: "__BYTE_ORDER" redefined [-Werror]
 #define __BYTE_ORDER __LITTLE_ENDIAN

In file included from /usr/include/sys/types.h:70:0,
                 from external/libsodium/src/libsodium/include/sodium/randombytes.h:5,
                 from ./common/sphinx.h:11,
                 from devtools/onion.c:5:
/usr/include/endian.h:11:0: note: this is the location of the previous definition
 #define __BYTE_ORDER __BYTE_ORDER__

make: *** [<builtin>: devtools/onion.o] Error 1
```
@rustyrussell  is it issue with ccan?

NicolasDorier:
https://github.com/rustyrussell/ccan/issues/68

NicolasDorier:
Fixed by https://github.com/ElementsProject/lightning/pull/1484

NicolasDorier:
For the sport, I tried to build litecoin-cli and bitcoin-cli.
Because I was unable how to compile them statically, I needed to add

```
boost \
     boost-program_options \
     libevent \
```

to the runtime image... which would make it shoots to 38MB :(
If someone knows how to ask bitcoin core to statically link libevent and boost, let me know.

NicolasDorier:
My mistake. Compiling myself, with dynamic linking to boost and libevent save 2.7MB at the cost of longer build time.

cdecker:
Whoa, awesome progress @NicolasDorier, I definitely want to merge this. Hopefully we can pin down 0.6 soon.

NicolasDorier:
I rebased and removed the fix for musl which has been merged.

NicolasDorier:
I am thinking just squashing everything as a single commit. It does not make much sense to review as individual commit. Would it be fine?

cdecker:
That'd be perfectly fine, I can squash upon merge :+1:

NicolasDorier:
rebased and squashed everything into single commit

NicolasDorier:
Working on adding a build arg which add perftrace and flamegraph.

NicolasDorier:
Add a build-arg `TRACE_TOOLS`. If `true`, then perf and flamegraph tooling get installed, as well as an utility `sample.sh` which take a 60 seconds stacktrace and save it to `/opt/traces`.

Example to build:
```
docker build --build-arg TRACE_TOOLS=true --build-arg DEVELOPER=1 -t nicolasdorier/clightning:0.0.0.16-dev .
```

You just have to map `/opt/traces` to the host's directory to be able to extract the flamegraph.

NicolasDorier:
My config so the an image with benchmarking tools get generated when I tag the image in git.

image

NicolasDorier:
Added `sample-loop.sh`. A script taking flamegraph every 5 minutes. Useful to run while running perf intense test.

NicolasDorier:
Can someone run travis again? seems like unrelated failure.

Sjors:
Concept ACK. Is there a way to include only Bitcoin or only Litecoin binaries? Or for that matter to skip those binaries altogether if you already have a container for them?

NicolasDorier:
@Sjors one idea can be to add build time args which skip this, as I did for the TRACE_TOOLS. Seems reasonable?

Sjors:
@NicolasDorier it does. I suspect it's best if the resulting image name is different depending on those build time flags.

NicolasDorier:
So the build is broken for alpine, see https://github.com/ElementsProject/lightning/issues/1532

I temporarily revert https://github.com/ElementsProject/lightning/pull/1520/files on this branch so I can move on.

I am unsure if moving to alpine is a good idea now.

jsarenik:
@NicolasDorier Moving to Alpine for docker is generally a very good idea and it shows us the hardcoded stuff. Sooner or later it would kick so thank you for pointing it out by playing with Alpine.

NicolasDorier:
Note to myself: Use `./configure` once https://github.com/ElementsProject/lightning/pull/1538 get merged and remove the revert commit.

NicolasDorier:
@analogic ah that might explain some of my benchmarks. I did not managed to pass more than 22 invoice creation + payment per second.

Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+    && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
+    && rm litecoin.tar.gz
+
+ENV LIGHTNINGD_VERSION=master
+
+WORKDIR /opt/lightningd
+COPY . .
+
+ARG DEVELOPER=0
+RUN make -j3 DEVELOPER=${DEVELOPER} && cp lightningd/lightning* cli/lightning-cli /usr/bin/
```
shesek: This should `./configure` before `make` (required as of https://github.com/ElementsProject/lightning/commit/c49765553ffc352e678684703c513351d418969e)

Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+     gmp-dev \
+     python \
+     python-dev \
+     python3 \
+     sqlite-dev \
+     wget \
+     git \
+     file \
+     gnupg \
+     swig
```
shesek: This should also install zlib (see https://github.com/ElementsProject/lightning/commit/083a2cee0744e93c9f43ab553466b0c6fe01867f#diff-04c6e90faac2675aa89e2176d2eec7d8)

Dockerfile
```diff
@@ -0,0 +1,122 @@
...
+        curl -Lo FlameGraph.tar.gz "https://github.com/brendangregg/FlameGraph/archive/v1.0.tar.gz" && \
+        tar -zxvf FlameGraph.tar.gz --strip-components=1 && rm FlameGraph.tar.gz && cd .. \
+    ) \
+  ) && \
+  apk del curl && \
+  rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*
+
+ENV LIGHTNINGD_DATA=/root/.lightning
+ENV LIGHTNINGD_PORT=9835
+ARG DEVELOPER=0
```
shesek: `ARG DEVELOPER=0` is already set above (line 61), this seems duplicated

jsarenik:
It is better to set developer mode using configure script, just to keep it
DRY. There it's set once and doesn't need to be repeated on command
line/environment.

Dockerfile
NicolasDorier: No, this is a different image so I need to set the environment variable.

NicolasDorier:
mmh getting `external/libsodium is not a submodule!`

jsarenik:
@NicolasDorier just once and just during build. Either by exporting `DEVELOPER=1` in environment before running `./configure` OR running configure with `--enable-developer`.

cdecker:
I'm currently setting things up, but we might end up having to manually push them for the time being (security settings around this org is pretty strict).

Also I find myself overriding the `--entrypoint` in almost every run, so I wonder if the entrypoint script isn't doing too much for its own good. If we go down this route we'll have two different ways to invoke the daemon (through the config and through command line args) that do not match. In particular I find it annoying that command line args get ignored.

Can we trim the magic down to adding prefix and suffix to the executed command, but not completely overriding? I like the tracing instrumentation, but I don't like the forced path through a config file and the cli args being discarded.

jsarenik:
By the way I have good experience with separate set of scripts inside the git tree which are shared across all CI assistance (e.g. Travis config would point to a script that contains commands instead of defining them, the same for Docker, etc.)

See https://github.com/jsarenik/spf-tools/tree/master/misc

NicolasDorier:
Why are you overriding with `--entrypoint` ? It never happened once to me

shesek:
> Because I'd like to have access to the other command line flags (like `--log-level`, `--lightning-dir`, ...) without having to edit a config file.

Lightning Charge allows specifying an `LN_OPT` environment variable that gets passed as arguments to `lightningd`, which you can then specify with i.e. `docker run -e LN_OPT='--log-level debug' elements/lightning`.

Alternatively, since lightningd is the main thing being run here, you can forward `$@` instead then specify with i.e. `docker run elements/lightning --log-level debug`, which is even nicer.

NicolasDorier:
@cdecker why not doing

```bash
docker run -e LIGHTNINGD_OPT=" \
log-level=debug
blah=value
" clightning/lightningd 
```
But additionally, I can also pass the pass parameters so you can do this:

```bash
docker run clightning/lightningd  --log-level=debug blah=value
```

NicolasDorier:
@cdecker I pushed, try `docker run clightning/lightningd --log-level=debug` it should work as well.

shesek:
Note that `network` is still expected to be specified only via `LIGHTNINGD_OPT`, meaning that `docker run clightning/lightningd --network mainnet` won't work as expected, which might be somewhat confusing.

Perhaps it would be better to specify the `network` and `chain` using separate environment variables, instead of overloading `network` in the config file?

Also, mutating the `config` file, which might be mounted from outside, doesn't seem like the cleanest solution. Using environment variables and passing the resulting "REPLACEDNETWORK" as an argument to `lightningd` rather than via the config file seems better.

NicolasDorier:
> Also, mutating the config file, which might be mounted from outside, doesn't seem like the cleanest solution. 

I never had any reason to mount and modify the config file outside the docker-compose or command line parameters via mounted folders. But what we can do in this case is to not modify the config file if `LIGHTNINGD_OPT` is blank.

I think supporting the `REPLACEDNETWORK` feature should be done only in config file. Because if we don't it mean we need even more complex logic to handle whether chain and network comes from config/command line or both and also rewriting the command line arguments... Lot's of headaches not worth the additional code.

The best would be either for this logic either to be backed directly into clightning code, or that we completely remove it.
However, if you remove it, I'll still need to maintain my own fork which use this for my projects instead of using clightning's repo version. :(

shesek:
> Because if we plan to support via command line as well, we need even more complex logic to handle whether chain and network comes from config/command line or both and also rewriting the command line arguments

I would suggest that:

1. Don't do any special handling for the config file at all, let it simply pass through. (or even, just remove the file-related `LIGHTNING_OPT` handling entirely in favor of passing args)

2. Introduce separate environment variables configuration for chain/network, without calling them "network".

3. If that new config is passed, determine `network` based on them and add it to the arguments list. Otherwise, just pass the argument as-is.

That seems to be even simpler than the current implementation, no?

NicolasDorier:
Closing this one, supersede by https://github.com/ElementsProject/lightning/pull/1616

PR #1321, opened 2018-04-04
pay, payalgo: Show erring_node as compressed DER pubkey.

ZmnSCPxj:
For consistency with other node pubkeys.


cdecker:
Was just wondering the other day about this, thanks @ZmnSCPxj 

ACK f51d1a8

Issue #1281, opened 2018-03-26
Auto-reconnect to peers we have channels to


ZmnSCPxj:
#1274 brings up that our peers with whom we have channels to, often end up having `connected: false`.

One way to help this is to have a periodic auto-reconnect every 5 to 10 minutes, to all known peers with whom we have channels to.

1.  Modify `connect` so it can be invoked internally.  See what I did in `lightningd/pay.c` to `sendpay` command: I made a public `send_payment` function that is given a callback for completion or failure, as well as a `struct sendpay_result` for the result of payment.  This is needed so that `connect` can be invoked internally by lightningd.
2.  Set up a tweakable autoconnect system (can set autoonnect cycle time range by command line, default 5 to 10 minutes).  Make sure to randomize the time between autoconnect attempts to reduce network spike.  Pay attention to `addr_unknown` field from the gossip connection failure message: if this is true then we really cannot connect to the peer due to lack of knowledge of address.

CubicEarth:
I can see how invoking ```connect``` can make sense for now, in the name of simplicity. Longer term, it seems like the periodic check would be better if it happened more frequently (every minute or two), but with a smaller number of pings, perhaps 2, rather than the 10 tries that ```connect``` currently makes.

ZmnSCPxj:
A better solution would be to let `connect` specify how many tries max to make and how long between each try as optional extra args. Then the autoreconnect can give values reasonable for its operation.

PR #1322, opened 2018-04-04
Maintain permanent connections for active channels

rustyrussell:
Fixes: #1281


lightningd/closing_control.c
```diff
@@ -95,6 +95,7 @@ static void peer_received_closing_signature(struct channel *channel,
 
 static void peer_closing_complete(struct channel *channel, const u8 *msg)
 {
+	u8 *msg;
```
ZmnSCPxj: `msg` is parameter of this function and is also a local variable.

cdecker:
Other than the `msg` param being shadowed, this is good to go :-)

rustyrussell:
> Should we worry about the case:
> 
> 1.   An "important" peer with a channel disappears forever.
> 2.  gossipd goes on a loop attempting to retry that peer forever.
> 3.  User (or automated pilot) attempts a connect command.

User will get an error:

	if (find_reaching(daemon, id)) {
		/* FIXME: Perhaps kick timer in this case? */
		status_trace("try_reach_peer: already trying to reach %s",
			     type_to_string(tmpctx, struct pubkey, id));
		return false;
	}

Perhaps we should probably stop the multiple connect attempts now for manual connects, which would allow us to fail faster.

But generally, I think trying forever is correct in this case: if we have an HTLC we'll eventually force close, and when we implement the json_close timeout logic that would do the same thing.

rustyrussell:
Fixed shadowed variable, missing `#include`.

ZmnSCPxj:
The code quoted is called by `reach_peer`, which only sends a message to masterd if the quoted code returns true. If it returns false, nothing is done and the `connect` command stalls.

rustyrussell:
Good point.  I agree, let's make connect a one-shot, and have a separate system for important peers.

gossipd/gossip.c
```diff
@@ -1634,37 +1638,32 @@ static struct io_plan *connection_out(struct io_conn *conn,
...
+		     type_to_string(tmpctx, struct pubkey, &reach->id));
+
+	/* If we want to keep trying, do so. */
+	imp = important_peerid_map_get(&reach->daemon->important_peerids,
+				       &reach->id);
+	if (imp) {
+		/* FIXME: Exponential backoff! */
+		status_trace("...will try again in %u seconds", 5);
+		/* If important_id freed, this will be removed too */
+		new_reltimer(&reach->daemon->timers, imp,
```
ZmnSCPxj: Should we not keep the returned `struct oneshot` in the `struct important_peerid`?  And before making a new timer, we should free any existing timers?  Otherwise `connect` commands may cause multiple overlapping timers to exist (I think?).

    /* Kick the timer if present, or start a new one if not present. */
    tal_free(imp->timer);
    imp->timer = new_reltimer(&reach->daemon->timers, imp, time_from_sec(5), retry_important, imp);

The `retry_important` function also needs to clear this timer:

    imp->timer = tal_free(imp->timer);

rustyrussell: Good point, we will end up with one timer per connect for every important peer.  Will fix, though not here.

gossipd/gossip.c
```diff
@@ -1817,17 +1821,24 @@ static void try_reach_peer(struct daemon *daemon, const struct pubkey *id,
...
 	struct pubkey id;
+	struct important_peerid *imp;
 
 	if (!fromwire_gossipctl_connect_to_peer(msg, &id))
 		master_badmsg(WIRE_GOSSIPCTL_CONNECT_TO_PEER, msg);
 
+	/* If this is an important peer, free any outstanding timer */
+	imp = important_peerid_map_get(&daemon->important_peerids, &id);
+	if (imp)
+		imp->reconnect_timer = tal_free(imp->reconnect_timer);
```
ZmnSCPxj: At first glance `imp->reconnect_timer` seems not initialized when new `important_peerid` is created.  However `important_peerid` is created only one place, and, it calls `retry_important` immediately, which sets `imp->reconnect_timer` to `NULL`, so no issue here.

ZmnSCPxj:
ACK f9f38b1d46cdffdbce3be49936178efdaf216e4a

ZmnSCPxj:
It is failing in CI, but I am uncertain why; the CI error pertains to openingd, not gossipd...? It seems the funder openingd died because the fundee openingd disconnected, while the fundee can see its own openingd disconnected from itself but did not die.  Some edge case of connecting?  Perhaps a reconnect try is still spuriously triggering?

rustyrussell:
Fixing that turned out to be more surgery than I anticipated!  Added WIP since I'm pushing an update to check CI...

rustyrussell:
The fast reconnect seems to be triggering a heap of fun bugs.   Will spend more time diagnosing tomorrow.

NicolasDorier:
looking forward for this one!

cdecker:
Yep, definitely a huge usability improvement. @rustyrussell is currently hunting down an issue with 32bit builds, but last I heard he had reproduced it locally.

CubicEarth:
Yes, this will be very significant. I am not sure how channels across the network are counted, but I suspect once these changes are adopted there will be a definite improvement in the connectivity of the graph.

BTW, I am checking on the status of this PR far more often than the price :)

lightningd/channel.c
```diff
@@ -320,6 +321,10 @@ void channel_fail_permanent(struct channel *channel, const char *fmt, ...)
 		channel->error = towire_errorfmt(channel, &cid, "%s", why);
 	}
 
+	/* Tell gossipd we no longer need to keep connection to this peer */
+	msg = towire_gossipctl_peer_important(NULL, &channel->peer->id, false);
```
ZmnSCPxj: If we should consider the peer unimportant when we drop the channel to chain, would it not be better to put it in `drop_to_chain` instead of putting in two different places? That way, if we add some other condition where we drop to chain, we do not need to add this again?  The improved close of #1355 hooks into `drop_to_chain`, for example.

common/read_peer_msg.c
```diff
@@ -38,6 +40,38 @@ static void handle_ping(const u8 *msg,
...
+			      tal_hex(msg, msg));
+		io_error(arg);
+	}
+
+	/* Gossipd can send us gossip messages, OR errors */
+	if (is_msg_for_gossipd(gossip)) {
+		if (!send_msg(cs, peer_fd, gossip, arg))
+			io_error(arg);
+	} else if (fromwire_peektype(gossip) == WIRE_ERROR) {
+		status_debug("Gossipd old us to send error");
```
ZmnSCPxj: Gossipd ***told*** us to send error

closingd/Makefile
```diff
@@ -66,7 +66,10 @@ CLOSINGD_COMMON_OBJS :=				\
 	common/type_to_string.o			\
 	common/utils.o				\
 	common/version.o			\
-	common/wire_error.o
+	common/wire_error.o			\
```
ZmnSCPxj: The `read_peer_msg` handles this new feature in closingd and openingd?  So it still occurs even if closingd and openingd do not include new code?

gossipd/gossip.c
```diff
@@ -1594,37 +1599,32 @@ static struct io_plan *connection_out(struct io_conn *conn,
...
-			     secs);
-		new_reltimer(&reach->daemon->timers, reach, time_from_sec(secs),
-			     try_connect, reach);
+	u8 *msg;
+	struct important_peerid *imp;
+
+	/* Tell any connect commands what happened. */
+	msg = towire_gossipctl_connect_to_peer_result(reach, &reach->id,
+						      false, strerror(errno));
+	for (size_t i = 0; i < reach->num_master_responses; i++)
```
cdecker: This fan-out in a subdaemon is rather strange, shouldn't we be doing it in the master? At least that's what we've been doing for invoices. IMHO the subdaemon should just report all raw events and the master should map them to commands. No need to pull in the commands concept into the subdaemon.

gossipd/gossip.c
```diff
@@ -1667,31 +1667,47 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
...
 
-	/* Already succeeded somehow? */
-	if (find_peer(reach->daemon, &reach->id)) {
-		status_trace("Already reached %s, not retrying",
-			     type_to_string(tmpctx, struct pubkey, &reach->id));
-		tal_free(reach);
+	if (peer) {
+		status_debug("try_reach_peer: have peer %s",
+			     type_to_string(tmpctx, struct pubkey, id));
+		if (master_needs_response) {
```
cdecker: Just tell the master, no need to disambiguate here.

lightningd/connect_control.c
```diff
@@ -28,79 +28,51 @@ static struct connect *new_connect(struct lightningd *ld,
...
 
-/* Gossipd tells us peer was already connected. */
-void peer_already_connected(struct lightningd *ld, const u8 *msg)
-{
-	struct pubkey id;
 
-	if (!fromwire_gossip_peer_already_connected(msg, &id))
-		fatal("Gossip gave bad GOSSIP_PEER_ALREADY_CONNECTED message %s",
-		      tal_hex(msg, msg));
+	c = find_connect(ld, &id);
```
cdecker: We should really loop through all connects here and tell all of them the same result.

gossipd/gossip.c
```diff
@@ -951,11 +945,33 @@ struct returning_peer {
 	struct daemon *daemon;
 	struct pubkey id;
 	struct crypto_state cs;
-	u64 gossip_index;
 	u8 *inner_msg;
 	int peer_fd, gossip_fd;
 };
 
+static void drain_and_forward_gossip(struct peer *peer, int gossip_fd)
```
cdecker: I was briefly afraid that this may introduce an ordering issue: enqueuing a `channel_announcement` while the peer was remote, and the `channel_update` as soon as the peer returned, and then draining into the `msg_queue` would change the order. But I think this is safe, since we only set the flag to reroute gossip messages after draining.

ZmnSCPxj:
@cdecker, I had a similar reaction too, but my thought was instead "we should factor out all the common code into a new peerd which filters out the gossip messages (forwarding those to gossipd) and only forwards channel-specific messages to the channel-level subdaemons, and forget about passing peer fds and peer crypto state between subdaemons!!!111eleven"

Then I suddenly remembered the old story of "Once upon a time, there was a programmer with a problem.  Ey said to erself, "I know, I will use a regex!".  Then ey had two problems." and decided not to mention it instead.

(But, peerd idea has advantage of making it possible to support multiple channels per peer, but do we want that really?)

NicolasDorier:
` but do we want that really?` yes. So there is no need to have channel downtime for 6 blocks if the channel get empty. One could just create an additional channel before the previous one is exhausted, then once the additional channel is created, close the old one.

ZmnSCPxj:
> One could just create an additional channel before the previous one is exhausted, then once the additional channel is created, close the old one.

Better done by creating a channel to a different node rather than the same node.  It would also enable you to do routing, by receiving on the old channel (which has most funds on the other side) and forwarding on the new channel (which has most funds on your own side).

NicolasDorier:
> Better done by creating a channel to a different node rather than the same node.

It depends. Imagine I am using a micro betting service, I want to have direct connect with this particular node, not another one.

But sure it does not have to be done for 0.6.

lightningd/channel.c
rustyrussell: Agreed, fixed.

gossipd/gossip.c
rustyrussell: The master can ask to connect to the same peer multiple times, we need to respond to all of them.  That seems the simpler API model.

We could enforce that the master can only ask for one at a time, though?

gossipd/gossip.c
rustyrussell: Might be cleaner to split this into two functions.  One caller is because of an explicit connect, which needs a response, one for an implicit (important peer retry), which doesn't.

But there are several things which can go wrong, so the return from such a function would be complex:
1. Peer already connected.
2. Currently trying to reach peer.
3. Some kind of socket/connect error.

Hence the "tell master" flag, though it's ugly.

lightningd/connect_control.c
rustyrussell: If we invert the model and merge requests in the master, yes.  I'll append that as a separate patch, I think, rather than rebasing and making a mess...

closingd/Makefile
rustyrussell: Yes, it's all inside read_peer_msg, which is why this change was a no-brainer.

gossipd/gossip.c
rustyrussell: I will address this, but note that we have to handle the case where we are reconnecting (because important peer) and we get a connect command for the same peer.  So the counter becomes a boolean.

ZmnSCPxj: We could put the `connect` commands in a list on `struct lightningd`, then send a single message to the masterd, and then the handler of that message would iterate over the entire list and trigger each of the matching `connect` commands.

rustyrussell:
This time, fixed conflict properly...

rustyrussell:
Holiday with kids means less time to code :)

Or, put another way: parenting is my chosen career, which demanded a full week on-site.  Thus my coding hobby suffered a small setback.

Thanks for your patience, everyone!

PR #1323, opened 2018-04-04
Add support for local socket connections

icota:
This is a flexible way to connect two instances of c-lightning via means other than the official ones. Run one side with `--localsocket=lightning-listener` and the other simply connects to the full socket path.

In practice two c-lightning instances are run on separate machines with glue code connecting to sockets and forwarding data between them.

I use this to make a lightning connection (on an otherwise disconnected device) via NFC, but can be made to support bluetooth, infrared, sound or even smoke signals, anything really. Use case would be communicating with always-connected payment terminals with a mobile device that might not be.

This is way out of lightning spec but _it doesn't seem to hurt anything_. Let me know know what you think!

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -8,6 +8,7 @@
 #include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/types.h>
+#include <sys/un.h>
 #include <wire/wire.h>
 
 /* Returns false if we didn't parse it, and *cursor == NULL if malformed. */
@@ -22,6 +23,9 @@ bool fromwire_wireaddr(const u8 **cursor, size_t *max, struct wireaddr *addr)
 	case ADDR_TYPE_IPV6:
 		addr->addrlen = 16;
 		break;
+	case ADDR_TYPE_PADDING:
+		addr->addrlen = ((u8)*max) - 2;
+		break;
 	default:
 		return false;
 	}
@@ -33,7 +37,7 @@ bool fromwire_wireaddr(const u8 **cursor, size_t *max, struct wireaddr *addr)
 
 void towire_wireaddr(u8 **pptr, const struct wireaddr *addr)
 {
-	if (!addr || addr->type == ADDR_TYPE_PADDING) {
+	if (!addr) {
 		towire_u8(pptr, ADDR_TYPE_PADDING);
 		return;
 	}
@@ -180,6 +184,20 @@ bool parse_wireaddr(const char *arg, struct wireaddr *addr, u16 defport,
 		ip = "::1";
 
 	memset(&addr->addr, 0, sizeof(addr->addr));
+
+	/* Addresses starting with '/' are local socket paths */
+	if (ip[0] == '/') {
+		/* Check if the path is too long */
+		if (strlen(ip) > sizeof(addr->addr)) {
+			goto finish;
+		}
+		addr->type = ADDR_TYPE_PADDING;
+		addr->addrlen = strlen(ip);
+		addr->port = 0;
+		memcpy(&addr->addr, ip, addr->addrlen);
+		res = true;
+		goto finish;
+	}
 
 	if (inet_pton(AF_INET, ip, &v4) == 1) {
 		addr->type = ADDR_TYPE_IPV4;
--- a/common/wireaddr.h
+++ b/common/wireaddr.h
@@ -33,7 +33,7 @@ enum wire_addr_type {
 struct wireaddr {
 	enum wire_addr_type type;
 	u8 addrlen;
-	u8 addr[16];
+	u8 addr[108];
 	u16 port;
 };
 
--- a/gossipd/gossip.c
+++ b/gossipd/gossip.c
@@ -46,6 +46,7 @@
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <sys/un.h>
 #include <unistd.h>
 #include <wire/gen_peer_wire.h>
 #include <wire/wire_io.h>
@@ -1552,12 +1553,13 @@ static struct io_plan *connection_in(struct io_conn *conn, struct daemon *daemon
 				   init_new_peer, daemon);
 }
 
-static void setup_listeners(struct daemon *daemon, u16 portnum)
+static void setup_listeners(struct daemon *daemon, u16 portnum, const char *localsocket)
 {
 	struct sockaddr_in addr;
 	struct sockaddr_in6 addr6;
+	struct sockaddr_un sun;
 	socklen_t len;
-	int fd1, fd2;
+	int fd1, fd2, fd3;
 
 	if (!portnum) {
 		status_info("Zero portnum, not listening for incoming");
@@ -1615,6 +1617,27 @@ static void setup_listeners(struct daemon *daemon, u16 portnum)
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Could not bind to a network address on port %u",
 			      portnum);
+
+	/* Optional UNIX socket */
+	if (localsocket) {
+		memset(&sun, 0, sizeof(sun));
+		sun.sun_family = AF_UNIX;
+		strcpy(sun.sun_path, (char*)localsocket);
+
+		fd3 = make_listen_fd(AF_UNIX, &sun, sizeof(sun), false);
+		if (fd3 >= 0) {
+			len = sizeof(sun);
+			if (getsockname(fd3, (void *)&sun, &len) != 0) {
+				status_broken("Failed get UNIX sockname: %s",
+					      strerror(errno));
+				close_noerr(fd3);
+				fd3 = -1;
+			} else {
+				status_trace("Creating UNIX socket listener");
+				io_new_listener(daemon, fd3, connection_in, daemon);
+			}
+		}
+	}
 }
 
 /* Parse an incoming gossip init message and assign config variables
@@ -1653,11 +1676,12 @@ static struct io_plan *gossip_activate(struct daemon_conn *master,
 				       const u8 *msg)
 {
 	u16 port;
+	u8* localsocket;
 
-	if (!fromwire_gossipctl_activate(msg, &port))
+	if (!fromwire_gossipctl_activate(msg, msg, &port, &localsocket))
 		master_badmsg(WIRE_GOSSIPCTL_ACTIVATE, msg);
 
-	setup_listeners(daemon, port);
+	setup_listeners(daemon, port, (const char*)localsocket);
 
 	/* OK, we're ready! */
 	daemon_conn_send(&daemon->master,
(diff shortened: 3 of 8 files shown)


common/wireaddr.h
```diff
@@ -26,14 +26,15 @@ enum wire_addr_type {
 	ADDR_TYPE_PADDING = 0,
 	ADDR_TYPE_IPV4 = 1,
 	ADDR_TYPE_IPV6 = 2,
+	ADDR_TYPE_LOCAL_SOCKET = 5
```
ZmnSCPxj: I am hesitant to put this in `struct wireaddr` as this is also used in BOLT standard-compliant `node_announcement` that get broadcast as node gossip.

At minimum the numbering should not be directly after the BOLT-defined ones.

It might be better to bring up this possible use case in lightning-dev mailing list.

Alternatively we could skip this address type when sending gossip announcements.

lightningd/options.c
```diff
@@ -155,6 +155,19 @@ static char *opt_add_ipaddr(const char *arg, struct lightningd *ld)
 
 }
 
+static char *opt_add_localsocket(const char *arg, struct lightningd *ld)
```
ZmnSCPxj: `listconfigs` is crashing in CI, probably due to not handling this case properly in `add_config` function. https://github.com/ElementsProject/lightning/blob/1a5514787037039477b27c7762461c5f2dde34a1/lightningd/options.c#L874-L877

common/wireaddr.h
icota: Agreed, it makes no sense at all to broadcast these. It hadn't occurred to me. I will look into alternatives.

icota: I don't think there's much value in adding local sockets to the BOLTs but I settled on a MIME type (`application/lightning`) and an application id (`LIGHTNING`) for NFC payments. I'll bring this up on the ML when I get the time since I'm not sure how to proceed on standardising these.

ZmnSCPxj: If we never broadcast these there is no need to bring it up on the mailinglist, but, I am uncertain if we should be using `struct wireaddr` then, or some other mechanism.  Intent, as I understand it,  is that `struct wireaddr`, is for standard broadcastable messages.  @rustyrussell @cdecker opinions?

rustyrussell:
It's cute to see these in the local address listings, but the consequences of putting it in struct wireaddr are probably not worth dealing with.

Just add a '--local-socket' option perhaps?  And set addr->type to ADDR_TYPE_PADDING.

icota:
@rustyrussell There is a `--localsocket` when we start the daemon, to make it listen. Do you mean a JSON option when connecting?

icota:
New code avoids changes to `struct_wireaddr` opting to just use `ADDR_TYPE_PADDING`. It's less code and doesn't broadcast.

`check-source` fails at `shellcheck` and `check` fails at `test-blockchaintrack`. I don't think either of those are my fault. I guess I'm missing something to run these tests correctly.

ZmnSCPxj:
CI is fine so I suppose it is OK.  You can install `shellcheck`, most distros seem to have it.   `test_blockchaintrack` is known flaky and rerunning it should be OK.

Now reviewing code, please wait...

gossipd/gossip.c
```diff
@@ -1519,6 +1521,27 @@ static void setup_listeners(struct daemon *daemon, u16 portnum)
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Could not bind to a network address on port %u",
 			      portnum);
+
+	/* Optional UNIX socket */
+	if (localsocket) {
```
ZmnSCPxj: This comes from the `fromwire_gossipctl_init`.  In `towire_gossipctl_init` a `NULL` pointer is converted to a 0-length array on-the-wire, I am uncertain if a 0-length array on-the-wire will be converted by `fromwire_gossipctl_init` to a `NULL` pointer.  I  shall check later when I can get access to my hacking node.

gossipd/gossip.c
```diff
@@ -1765,12 +1765,62 @@ static void try_connect(struct reaching *reach)
 	io_new_conn(reach, fd, conn_init, reach);
 }
 
+static void try_connect_local_socket(struct daemon *daemon, const struct pubkey *id, const struct addrhint *a)
```
ZmnSCPxj: Might conflict with ongoing changes in #1322, that is higher priority than this I think, so please monitor that PR when it gets merged in and see if this gets conflicted (maybe not a text-level conflict but some of the functions this `try_connect_local_socket` calls might change their signatures and exact behavior).  The first thing I can think of is that `reach->succeeded` has been removed in #1322, but please see that PR more for other changes that might affect this code.

gossipd/gossip.c
ZmnSCPxj: No issue here: turns out 0-length arrays on-the-wire are converted to `NULL` by `fromwire_*` functions, so ok.

ZmnSCPxj:
ACK 61cece68a6dd78b7a3546be0ca4470b319dcc4a6

Note however that #1322 changes code that this PR depends on, so we should take care. I somewhat prefer that we get #1322 merged first, then adapt this PR to it, but anyway adapting is needed at some point.

icota:
Thanks @ZmnSCPxj! No big deal if you merge #1322 first, this one is smaller so easier to adapt

cdecker:
Holding this PR back a bit further since #1322 is rather high priority and I don't want to delay by introducing conflicts. Just wanted to acknowledge that we haven't forgotten about this excellent change @icota :wink:

ZmnSCPxj:
@icota could you please rebase? #1322 has been merged.

icota:
Rebased, waiting to see if CI is happy

gossipd/gossip.c
```diff
@@ -1552,12 +1553,13 @@ static struct io_plan *connection_in(struct io_conn *conn, struct daemon *daemon
 				   init_new_peer, daemon);
 }
 
-static void setup_listeners(struct daemon *daemon, u16 portnum)
+static void setup_listeners(struct daemon *daemon, u16 portnum, u8 *localsocket)
```
ZmnSCPxj: `localsocket` is only ever read in this function, and is effectively an input string.  Maybe better to write as `const char *` and move the type conversion from `u8*` to `char *` to caller of `setup_listeners`.  Or at least `const u8*` in the argslist.

gossipd/gossip_wire.csv
```diff
@@ -24,6 +24,9 @@ gossipctl_init,,no_reconnect,bool
 # Gossipd->master, I am ready.
 gossipctl_init_reply,3100
 
+gossipctl_init,,lslen,u16
```
ZmnSCPxj: Should this not be with the `gossipctl_init` fields *before* `gossipctl_init_reply`?  Clearer that way.

ZmnSCPxj:
Sorry @icota but a new commit in #1440 also conflicts with this: specifically the third commit by @rustyrussell moves setup of listeners to after the `activate` message, so the `localsocket` argument in `init` needs to be moved to the new `activate`.

icota:
No worries @ZmnSCPxj, local sockets are kind of a niche application so obviously important stuff needs to go in first. I've rebased yet again.

ZmnSCPxj:
ACK 4ae3aaa50619c4605f16f8c51b8b4850698e05c7

icota:
`make check`  fails at `test_pay_disconnect` which doesn't seem to be related at all:

```
  File "tests/test_lightningd.py", line 4280, in test_pay_disconnect
    l1.daemon.wait_for_log(r'Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: received ERROR channel .*: update_fee 150000 outside range 4250-75000')
```

ZmnSCPxj:
Kicked CI, hopefully it is a temporary thing.

common/wireaddr.c
```diff
@@ -181,6 +185,16 @@ bool parse_wireaddr(const char *arg, struct wireaddr *addr, u16 defport,
 
 	memset(&addr->addr, 0, sizeof(addr->addr));
 
+	/* Addresses starting with '/' are local socket paths */
+	if (ip[0] == '/') {
+		addr->type = ADDR_TYPE_PADDING;
+		addr->addrlen = strlen(ip);
+		addr->port = 9999;
+		memcpy(&addr->addr, ip, addr->addrlen);
```
cdecker: Is this missing an `addrlen < sizeof(addr->addr)` check?

gossipd/gossip.c
```diff
@@ -1542,6 +1542,12 @@ static struct io_plan *connection_in(struct io_conn *conn, struct daemon *daemon
 		BUILD_ASSERT(sizeof(s4->sin_addr) <= sizeof(addr.addr));
 		memcpy(addr.addr, &s4->sin_addr, addr.addrlen);
 		addr.port = ntohs(s4->sin_port);
+	} else if (s.ss_family == AF_UNIX) {
+		struct sockaddr_un *sun = (void *)&s;
+		addr.type = ADDR_TYPE_PADDING;
+		addr.addrlen = sizeof(sun->sun_path);
+		memcpy(addr.addr, &sun->sun_path, addr.addrlen);
+		addr.port = 9999;
```
cdecker: I think 0 as magic value might be better here, just in case we use the port somewhere.

cdecker:
Just two last nits, happy to merge this soon ^^

icota:
I've amended the code, I'm thinking this might be the day :crossed_fingers:

icota:
Foiled by CI `test_closing_different_fees` :cry:

rustyrussell:
Actually, after more thought, you can just introduce a new 'wireaddr_or_socket' type, and wire that up in the various places...

ZmnSCPxj:
Agree with the `wireaddr_or_socket` type idea, but maybe name to `contactaddr`?

    struct contactaddr {
        enum contactaddr_type type;
        union {
            struct wireaddr wireaddr;
            char *localsocket;
        } u;
    };

This would probably require special marshaling `towire_` and `fromwire_` code, though.

rustyrussell:
Yuck, this ended up quite invasive.  There are some nasty places to fix, too.

I've got a very rough cut here:
https://github.com/rustyrussell/lightning/tree/guilt/pr-1323

It's a bit weird since it's built on this PR then ends up reworking much of it: it might be cleaner to combine into one logical sequence.

ZmnSCPxj:
@rustyrussell your branch looks fine to me, although I mildly agree to combine to a better logical sequence.  I would have moved the "single `localsocket` -> multiple `wireaddr_or_sockname` and make local sockets in a loop" conversion to its own commit (i.e. leave the `localsocket` in `gossipctl_activate` in the big `wireaddr` to `wireaddr_or_sockname` move).

rustyrussell:
OK, This has been swallowed into the "crap, where did my time go!" PR #1450...

Issue #590, opened 2018-01-13
Must allow multiple bolt11 fallback addresses


rustyrussell:
And decodepay needs to make them an array in the JSON output.

PR #1325, opened 2018-04-05
API break: allow multiple fallback address

rustyrussell:
Fixes: #590 

This breaks JSON API, so we'd better do it now.  We use the deprecated_apis trick to make things work for now.


doc/lightning-invoice.7.txt
```diff
@@ -32,6 +32,10 @@ UTF-8, and cannot use '\u' JSON escape codes.
 The 'expiry' is optionally the number of seconds the invoice is valid for.
 If no value is provided the default of 3600 (1 Hour) is used.
 
+The 'fallbacks' array is one or more fallback addresses to include in
```
ZmnSCPxj: This paragraph is not up-to-date in the manpage file.

cdecker:
Concept ACK, pending manpage fix

rustyrussell:
Thanks, updated!

cdecker:
CI failure was unrelated, will look at it later :wink: 
ACK 8bc9488

PR #1326, opened 2018-04-05
pytest: Use a dict to pass options to lightningd

cdecker:
This allows us to have some default options that can then be overridden easily
on a per-test basis. Ran into this while writing a test that was trying to set
`--locktime-blocks`, but was overridden by the defaults due to CLI precedence.


practicalswift:
post merge utACK 7b19ff0

PR #1327, opened 2018-04-05
Support debugging with lldb

conanoc:
fix https://github.com/ElementsProject/lightning/issues/1093

Running with lldb cause SIGINT, which makes waitpid(), read()
returns error with errno as EINTR. This patch retry these calls
to ignore EINTR errors.

Tested only on macOS.

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -157,7 +157,7 @@ static void bcli_finished(struct io_conn *conn UNUSED, struct bitcoin_cli *bcli)
 	bool ok;
 
 	/* FIXME: If we waited for SIGCHILD, this could never hang! */
-	ret = waitpid(bcli->pid, &status, 0);
+	while ((ret = waitpid(bcli->pid, &status, 0)) < 0 && errno == EINTR);
 	if (ret != bcli->pid)
 		fatal("%s %s", bcli_args(bcli),
 		      ret == 0 ? "not exited?" : strerror(errno));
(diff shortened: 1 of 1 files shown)


ccan/ccan/ptr_valid/ptr_valid.c
```diff
@@ -161,7 +161,7 @@ static void finish_child(struct ptr_valid_batch *batch)
 {
 	close(batch->to_child);
 	close(batch->from_child);
-	waitpid(batch->child_pid, NULL, 0);
+	while(waitpid(batch->child_pid, NULL, 0) < 0 && errno == EINTR);
```
ZmnSCPxj: Unfortunately changes to CCAN could should preferably be pushed upstream to CCAN: https://github.com/rustyrussell/ccan

arowser:
Tested pass on macOS 10.12.4

rustyrussell:
Correct fix is to get rid of ptr_valid which is needed by autodata on non-ELF platforms (ie. MacOS).

Ideally someone who knows macos would come up with a clever way of doing autodata there?  Otherwise we could generate the type_to_string and json_command arrays using grep...

conanoc:
Removed other changes except bitcoind.c from the previous commit.

conanoc:
@ZmnSCPxj @rustyrussell I think this patch is ready. Please take a look.

ZmnSCPxj:
ACK 474cc1b24d726d40b422c3277e37aece560e3ddc

cdecker:
ACK 474cc1b

PR #1329, opened 2018-04-05
topology: Two off-by-one error when catching up with the blockchain

cdecker:
There are two very hard problems in software engineering:

 1. Off-by-one errors

In this case we were rolling back further than needed and we were starting the
catchup one block further than expected.


ZmnSCPxj:
ACK 8330f34e73041eb16076ce1fa3de5dde119ab601

PR #1330, opened 2018-04-05
payalgo: Support bolt11 'r' fields.

ZmnSCPxj:
Part of spec.

Not sure how to test, though.We need some way to generate `r` fields in bolt11 invoices.


lightningd/payalgo.c
```diff
@@ -87,6 +89,120 @@ json_add_failures(struct json_result *r, char const *n,
...
+	amount = try_msatoshi;
+	for (i = tal_count(cp->subroute) - 1; i >= 0; --i) {
+		(*route)[route_end + i].channel_id = cp->subroute[i].short_channel_id;
+		/* Get next hop, or dest if last on route. */
+		if (i != tal_count(cp->subroute) - 1)
+			(*route)[route_end + i].nodeid = cp->subroute[i + 1].pubkey;
+		else
+			(*route)[route_end + i].nodeid = *receiver_id;
+		(*route)[route_end + i].amount = amount;
+		(*route)[route_end + i].delay = (*route)[route_end + i - 1].delay - cp->subroute[i].cltv_expiry_delta;
```
ZmnSCPxj: This line is wrong.  It assumes the previous delay is loaded.  But it is not loaded because we are loading route hops backward.  Instead, we should load the delay as + from next delay (or final target delay if this is the last hop).

ZmnSCPxj:
Unfortunately it seems unannounced channels cannot be used for `r` bolt11 fields.  `r` fields include the `fee_base_msat` and `fee_proportional_millionths`.  There is no way to know how much the peer will want to charge for the route from that peer to us.

The BOLT spec mentions `fee_base_msat` and `fee_proportional_millionths` in only BOLT 7, which is about gossip (and hence means the channel is announced), and BOLT 11, the invoice spec.

My understanding of the invoice route is that for example if it indicates a route peer A going through channel C to us, then the fees should be the fees that A will charge for use of channel C.  How do we learn that? Nothing in the spec suggests that we can learn that from a peer except for `channel_update` which I think is not created unless the channel is announced.

ZmnSCPxj:
Even with non-public channels, how does the other side know how much we want to charge for use of our channel?  Seems, limitation of BOLT spec?

cdecker:
We had an alternative code path that'd allow `channel_update`s for unannounced channels, which was later replaced with one that would send unsigned `channel_announcement`s followed by signed `channel_update`s, but now we're left without any of these. We probably need to bring this up at the next spec meeting, and fix a canonical way of handling these. I think `channel_update`s should be exchangeable between endpoints, and the lack of `channel_announcement` ensures they are local only.

cdecker:
Just to add some more color to this: we might eventually want to introduce a `routing_state_view` to be able to temporarily add channels to our view, especially for the non-public channels.

ZmnSCPxj:
I would argue the opposite; we should retain even non-published channels in our routing map once we learn of them via invoice.  This makes it harder for route analyzers if we sometimes use non-published channels to pay unrelated nodes; the route analyzer might assume we terminate to the node after the non-published channel, when in fact we are using that node as a hop node, going into one non-published channel and out another non-published channel, and probably helps the privacy of that node too, since this is activity that naive route analysis might assume terminates at them but is actually hopped.

ZmnSCPxj:
Yet another wrinkle: it seems c-lightning will return `WIRE_UNKNOWN_NEXT_PEER` for channels in `CHANNELD_AWAITING_LOCKIN`, meaning they cannot be used for private routes, ha.

It seems to test this at integration test, we need to actually support non-published channels too, sigh.

In any case my attempts to make this work also revealed a new bug in my bolt11-private-route handling:

* If a channel or node failure occurs on the private part of the route, we inform gossipd, but for private routes, then gossipd does not know it, so it does nothing.  We then retry with the same "contactpoint" i.e. the private route start point, and gossipd returns the same public route.  So probably I need to keep track of the public route length, and if the erring node index exceeds the public part, go to the next contactpoint.

ZmnSCPxj:
@cdecker has there been some resolution regarding `channel_update` to get fees for unpublished channels?

cdecker:
@ZmnSCPxj not really, they only make sense for published nodes since otherwise we are one of the peers (and know what we'd charge) or the other end, in which case we most likely don't want to route in that direction (because we're the receiving endpoint).

We should probably just send a `channel_update` to our peer upon lockin. That way we could have `invoice` look up incoming not yet public channels and pass that in as `r`. Then we could do a 2 hop route through not yet public channels. Would that help?

ZmnSCPxj:
Yes, that is indeed what I was referring to when I asked about "`channel_update` for unpublished channels": if the channel is unpublished, we still need to get `channel_update` (or something equivalent anyway) from the other side, else we cannot really use the channel for receiving.  If we are receiving, it is the other side which determines what fees are for forwarding to us, so we need to get fees from them, meaning even for unpublished channels we should give fees to the other side too.

ZmnSCPxj:
I am getting occasional spurious errors in encoding and decoding `r` fields.

`Invalid bolt11: r: hop 0 truncated`

ZmnSCPxj:
Ready for review.

cdecker:
I have neglected this PR for way too long, sorry. One thing that jumps right at me is the introduction of contact points, rather than updating the topology view. Any reason we are building an entire new layer over the routing mechanism here? It feels a bit weird to be iterating over contact points that are either the destination itself or its peers and attempting to find routes to them, when we could just temporarily add the channels and route normally.

contrib/pylightning/lightning/lightning.py
```diff
@@ -161,7 +161,8 @@ def invoice(self, msatoshi, label, description, expiry=None, fallbacks=None, pre
             "description": description,
             "expiry": expiry,
             "fallbacks": fallbacks,
-            "preimage": preimage
+            "preimage": preimage,
+            "maxroutes": maxroutes
```
cdecker: `maxroutes` isn't really the most intuitive name. Can we make it something like `privchannels`?

lightningd/invoice.c
```diff
@@ -156,105 +158,153 @@ static bool parse_fallback(struct command *cmd,
...
 }
-/* Create 'r' routes for bolt11. */
-static struct route_info **
-create_bolt11_routes(const tal_t *ctx,
-		     struct lightningd *ld,
-		     unsigned int max_routes)
+
+/* Handle the reply from the gossip_get_private_routes request. */
+static void
+invoice_command_get_private_routes_reply(struct subd *_ UNUSED,
```
cdecker: Rather ugly that now even the simple `invoice` command has to become an async command stashing it's state aside...

cdecker: But if it gets the job done

ZmnSCPxj:
> Any reason we are building an entire new layer over the routing mechanism here?

Honestly? I wanted to not have to touch the gossipd code too much, as I have not fully studied the BFG algorithm fully.

But if we are going to switch over to something like `getallroutes` (which is arguably more intuitive for users, rather than our `getroute` which magically selects alternate routes if a `sendpay` fails), then the code here will get removed either way (and I will still end up not touching the BFG code, since BFG seems to be an inherently shortest-path algorithm, not a find-all-routes algorithm like `getallroutes` will be).

It could be done by extending the routemap, admittedly, but adding a newer layer was simpler than modifying the routemap, routing, then modifying it *back*.

lightningd/invoice.c
ZmnSCPxj: An alternative would be for gossipd to send `channel_update`s for local channels to lightningd, and lightningd would then use that information.  It might even be better for lightningd, since it can also directly check channel capacity.

contrib/pylightning/lightning/lightning.py
ZmnSCPxj: `maxprivchannels`?

cdecker:
Gotcha, I was just curious about adding the additional layer that basically wraps `n` separate path finding calls (and which is in turn wrapped by repeated path finding calls when an attempt fails). I'll add a few more points inline:

> Honestly? I wanted to not have to touch the gossipd code too much, as I have not fully studied the BFG algorithm fully.
> 
>But if we are going to switch over to something like getallroutes (which is arguably more intuitive for users, rather than our getroute which magically selects alternate routes if a sendpay fails), then the code here will get removed either way (and I will still end up not touching the BFG code, since BFG seems to be an inherently shortest-path algorithm, not a find-all-routes algorithm like getallroutes will be).

BFG is not that special actually, think of it as a negative-cost-loop tolerant Djikstra :wink: Adding a channel to the network view and then running BFG will do what you want it to do.

Once we switch to k-shortest-paths, like Yen's Algorithm, we still need a shortest-path algorithm that does the heavy lifting, and it could still be BFG (though we have eliminated negative fees, so we don't have to anymore).

> It could be done by extending the routemap, admittedly, but adding a newer layer was simpler than modifying the routemap, routing, then modifying it back. There is also the issue of "what happens if multiple pay commands run in parallel" --- do they keep their modifications of the routemap enabled at the same time, and is that OK? Or should we just not modify the routemap back to before adding the temporary private routes?

Since `gossipd` is single threaded, like all other daemons, we can ensure that there are no concurrent path requests (as long as we don't call out to other daemons or use some timers). So adding channels, computing a route and then deleting the channels is safe. Arguably even if we didn't remove the edges at all, I think it'd still be ok. After all we can only end up at the node that told us about that channel, i.e., we don't have any outgoing edges from it.

> Accepting invoice-contained private routes to the routemap on a permanent basis, on thinking about it, is also a little strange --- we go through all those checks (signatures of the channel_announcement, checking the txout exists and is unspent) before accepting channels to the routemap. But simply because a payee signs an invoice we want to pay, we now accept one or more channels as existing without even checking them? (or should we at least check them onchain?)

Excellent point, hadn't thought about this before. In my mind all the signatures and outpoint checks were to limit the spam that a node can produce, i.e., ensuring that a node actually has opened a channel, and that a node cannot send updates for remote channels. When adding channels from invoices our view can get polluted by people we accept invoices from, and we get added non-functional channels (that we'll probably drop when trying to use them anyway). Not sure what to make of it tbh.

ZmnSCPxj:
>So adding channels, computing a route and then deleting the channels is safe.

An issue (which came up in testing due to a bug in my test) is when the payment fails and the channel that failed is one of the private channels introduced.
Since `getroute` is magical and secretly marks the channel as unroutable, we have the situation below:

1. We add a private channel to the routemap.
2. We get a route that uses the private channel.
3.  We remove the private channel from the routemap.
4. We `sendpay`, and it fails at the private channel.
5.  `gossipd` marks the failed channel, but silently fails since it does not exist in the routemap.
6. We loop and start from 1 again.  Since the private channel was not marked unroutable, we get the exact same route as before.

So we either remove the private channels after the entire payment completely succeeds or completely fails, or we accept private channels to the routemap on a permanent basis.

The contactpoint technique just drops the current contactpoint we are using if the failure was on the private channel.  This can also be done with the above technique, but is slightly harder since if multiple private routes are provided, we need to scan all of them to find out if the failing channel or node is on that route, whereas the contactpoint system just drops the current contactpoint.

ZmnSCPxj:
Looking at Yen's algorithm, it seems suspiciously similar (if you squint) to what we already do currently (call shortest-path algo, if it fails mark a channel on that path as unroutable, call shortest-path algo, etc.).

rustyrussell:
For `r` we should, in general:

1. Try without it.  If that fails...
2. For each path in r:
    * Try routing to the first node in that 'r' path, then sew the path together.

The only change to the routing algo is it now needs an arg telling it the (reduced) max-hops.

It's far less important to generate 'r' invoices: I would drop that (generate them manually if need be, for testing). Later we might do it automagically if we have no public channels.

A larger routing overhaul is a separate work which need serious post-0.6 design and thought.  Meanwhile we should seek the minimum viable fix.

ZmnSCPxj:
The contactpoint system does what you describe already. We generate a contactpoint for each `r`, then prepend the destination as another contactpoint.

The contactpoint list is then iterated over. We compute the target payment and compute how much has to reach the contact point node. If we fail to find a route to the contactpoint, or we get a failure onion past the contactpoint node, we drop the current contactpoint and move on to the next.

I think the current PR is minimal enough as is.

ZmnSCPxj:
Yes.  The contactpoint created at the destination is added using `list_add`, which adds to the front of the list:

https://github.com/ZmnSCPxj/lightning/blob/31f4b259d89098ea5867374da154ebd6b7b36359/lightningd/payalgo.c#L267-L278

Contactpoints created for each `r` route are added using `list_add_tail`, which adds to the end of the list:

https://github.com/ZmnSCPxj/lightning/blob/31f4b259d89098ea5867374da154ebd6b7b36359/lightningd/payalgo.c#L279-L298

cdecker:
Great, I'd say this is good to go, and let's revisit the `contact_point`s layer later on, maybe.

ACK 34f73c2

cdecker:
Currently needs a rebase, and @rustyrussell's approval :-)

cdecker:
Seems we now have a case clash:

```
lightningd/gossip_control.c: In function ‘gossip_msg’:
lightningd/gossip_control.c:160:2: error: duplicate case value
  case WIRE_GOSSIP_LOCAL_CHANNEL_UPDATE:
  ^
lightningd/gossip_control.c:143:2: error: previously used here
  case WIRE_GOSSIP_GET_PRIVATE_ROUTES:
  ^
<builtin>: recipe for target 'lightningd/gossip_control.o' failed
```

ZmnSCPxj:
Ready for review.

Sadly my hacking node will be offline for some time, so I will have some difficulty updating this pull request in the close future, but I will try to respond to reviews anyway.

cdecker:
ACK 2ad7867

Code is ok, but I'll let @rustyrussell get the final say.

cdecker:
@ZmnSCPxj I'm sorry I botched the merge edit, and seem to be unable to push to your branch to to a real rebase. Could you rebase and I'll merge it :-)

ZmnSCPxj:
Sorry for the ridiculously long hiatus, I shall see what I can do (hacking computer is currently unstable due to lower-level issues, am currently stabilizing it).

cdecker:
This PR has been inactive for a long time (partly my fault), and is probably more work to rebase than it is to rewrite from scratch. Would anybody object if I close this? @ZmnSCPxj @rustyrussell

PR #1332, opened 2018-04-06
ccan: update to more recent version.

rustyrussell:
In particular, this gets some MacOS fixes from #1327.
It also includes a major intmap update which fixes corner cases in traversals,
and requires ccan/bitops.


ZmnSCPxj:
ACK 471464d6c6eb903cf8c27cfe42b2964935c2bfe9

Just a cursory check, I am mostly trusting that CCAN has proper checking itself, overall looks fine to me.

cdecker:
ACK 471464d

rustyrussell:
Oops, a bogus change snuck in:

```
check-shellcheck:
	true||git ls-files -- "*.sh" | xargs shellcheck
```

That should no longer be necessary!  I'll fix...

PR #1339, opened 2018-04-07
master: Move pid-file creation after the daemonization

### LINKED ISSUE #1338: Pid-file issue when creating `systemd` service

I'm trying to setup `lightningd` as a background service which I can run through `systemd`. I want to have my LN node running at all times on my VPS, and this seems like a logical way to do it. Please correct me if that's not the case.

This is how `/etc/systemd/system/c-lightning.service` looks:
```
[Unit]
Description=Lightning network daemon
After=network.target

[Service]
ExecStart=/usr/local/bin/lightningd --network bitcoin --log-file /home/torkel/.lightning/debug.log --daemon --pid-file /run/lightningd/lightningd.pid
RuntimeDirectory=lightningd
User=torkel
Type=forking
PIDFile=/run/lightningd/lightningd.pid
```

After `systemctl start c-lightning`, the command hangs without exiting. `lightnind` starts up without issue, and `getinfo` returns the usual JSON.

Checking the PID of `lightningd`, I get 8967. However, `/run/lightningd/lightning.pid` contains 8951.

I believe the issue here is that `lightningd` writes the wrong PID to the file, but I know very little about these sort of issues. Any help would be much appreciated.

EDIT: I realized I don't have to specify a PID-file for `systemd` to work. I'm unfamiliar with all of this, so I just assumed so since that was how my `bitcoind` service worked. In that case, this issue might not be that big of a deal, or even a issue at all.

cdecker:
Creating the pid-file before daemonizing results in the pid-file containing the
pid of the process that started the daemon, but is now dead.

Fixes #1338

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -290,9 +290,6 @@ int main(int argc, char *argv[])
 	/* Handle options and config; move to .lightningd */
 	newdir = handle_opts(ld, argc, argv);
 
-	/* Create PID file */
-	pidfile_create(ld);
-
 	/* Ignore SIGPIPE: we look at our write return values*/
 	signal(SIGPIPE, SIG_IGN);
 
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
ACK 0993315748bf488106fa06499b8de493106e9bf8

PR #1343, opened 2018-04-08
Some drive-by fixes

cdecker:
A number of fixes that are not really big enough to warrant their own PRs.

 - Update the builder images to include `shellcheck`
 - Some warnings
 - An attempt to fix the flaky `test_blockchaintrack` test by ensuring rollback is complete before proceeding
 - A less noisy `gossipd` to be less annoying during block rescan
 - Lowerbound rescan to the start of LN, otherwise repeated restarts will back off indefinitely
 - Don't copy newline when looking up a block for a `txout`

### DIFF
--- a/contrib/Dockerfile.builder
+++ b/contrib/Dockerfile.builder
@@ -29,6 +29,7 @@ RUN apt-get -qq update && \
 	python3-pip \
 	python3-setuptools \
 	python-pkg-resources \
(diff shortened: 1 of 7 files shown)


ZmnSCPxj:
ACK 8567e655716c84f1a4cc0c0e753d3b221ab928df

PR #1345, opened 2018-04-09
wallet, payalgo: Save detail of payment failures for later reporting.

ZmnSCPxj:
Pointless for remote failures as those are never sent by
the erring node, but for local failures we can give more
detail.


cdecker:
Had to merge the tiny conflict, hence the additional commit, if you sign off this fix I'll merge it today @ZmnSCPxj 

Sorry for the delay btw

ZmnSCPxj:
ack 9478f081737804dc080be60a838da59e7a1fe0f6 and thank you

Issue #1334, opened 2018-04-06
Duplicate channel_announcements cause gossipd to abort


fixone:
lightningd was running fine, but had some problems receiving payments (from and Eclair Wallet) so I did a ```git pull``` + ```make clean && make``` and after compilation lightningd won't start 
```
 lightningd: lightning_gossipd failed (exit status 242), exiting.
```

and log shows
```
2018-04-06T19:03:42.854Z lightningd(415): Adding block 1283896: 00000000000004471e6adcce501a50f780c1a52d6b4ed609cfd599d90e3cb603
2018-04-06T19:03:42.909Z lightning_gossipd(423): req: type WIRE_GOSSIPCTL_INIT len 126
2018-04-06T19:03:42.909Z lightning_gossipd(423): Creating IPv6 listener on port 9735
2018-04-06T19:03:42.909Z lightning_gossipd(423): STATUS_FAIL_INTERNAL_ERROR: Announcement 010083a97a21d92b7f4077b1bb1396e5c634f33127a6d4f6b392c0abf992565500f428f1a4b361c51e89cfacc8238f28d6c803b047d61669695181c106b11b6189f763af1927e17f475eadf5ae486244eb79eab05004f6b0da5ae288451c6445ca7602b23fc1a1c4ff26fe2b0437660d0668a2b172bb073a215bbee73600a9448847248cadc07b3799b097f2de43ce441d04954f65e18b7fd2e28c8d5e414b65725f5c87efa73a9e63fbece02b14e0b227a63b5c2edd2e37b0df8aa2210540126e5c973f043622c3817e80205d01b31d0bb3ecf605d77935a426d87264e2ee5ce8b56708299e74fe7ffb46cf0a4c1bfb501d15494290f167211403be928b1e3f1ad0000043497fd7f826957108f4a30fd9cec3aeba79972084e90ead01ea33090000000013a8de0000070000038f80b541c3a9c1bbd54b4b15a379db83221e3fb50d5ffc543cef6d62a7f7163103eca120d4d74b1d0cdbb4c8c66d36aceba0ed5a21c404cf672379f7bd43c61e460317fe38cfbbae93ee25f3b4d4bf3c00bc95a7632051ea2e9fbe611314712223180369bb6d303fb4c06379d6a2afd8b974f60404a4375936f59a1826eb6486c5a934 was replaced?
```

fixone:
[...]

PR #1348, opened 2018-04-09
Demote 'Announcement X was replaced' to status_unusual

Sjors:
Workaround for #1334.

Although my node has been running fine with it for days, apparently this isn't without risk, as @cdecker wrote:

>  Not making the error fatal results in null-pointers later in the execution, which is really strange

Removing `gossip_store` in `~/.lightning` as suggested by @ZmnSCPxj is probably safer.

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -621,9 +621,8 @@ void routing_add_channel_announcement(struct routing_state *rstate,
 
 	if (replace_broadcast(chan, rstate->broadcasts,
 			      &chan->channel_announce_msgidx, take(msg)))
-		status_failed(STATUS_FAIL_INTERNAL_ERROR,
-			      "Announcement %s was replaced?",
-			      tal_hex(tmpctx, msg));
+    status_unusual("Announcement %s was replaced?",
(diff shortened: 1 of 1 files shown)


gossipd/routing.c
```diff
@@ -621,9 +621,8 @@ void routing_add_channel_announcement(struct routing_state *rstate,
 
 	if (replace_broadcast(chan, rstate->broadcasts,
 			      &chan->channel_announce_msgidx, take(msg)))
-		status_failed(STATUS_FAIL_INTERNAL_ERROR,
-			      "Announcement %s was replaced?",
-			      tal_hex(tmpctx, msg));
+    status_unusual("Announcement %s was replaced?",
```
cdecker: Nit: Indentation

rustyrussell:
We should really be doing simple sanity checks on the store and truncating/removing it when they fail.

But this error is the *reason* we got a bug report, and we still haven't figured out *why* that's happening, so I dislike removing it...

rustyrussell:
See #1352

Sjors:
Agreed, and #1352 looks much more sophisticated...

Issue #1344, opened 2018-04-08
absurd fees


FeedTheWeb:
### Issue and Steps to Reproduce
Keep getting these absurd fees errors  when trying to pay more than what is currently my balance  

> {
>   "code": 206,
>   "message": "Fee 18446743931975630848 is 12932592384.904261% of payment 142637635077; max fee requested is 0.500000%.",
>   "data": {
>     "msatoshi": 142637635077,
>     "fee": 18446743931975630000,
>     "feepercent": 12932592384.90426,
>     "maxfeepercent": 0.5,
>     "delay": 6,
>     "maxdelay": 500,
>     "getroute_tries": 6,
>     "sendpay_tries": 0,
>     "route": [
>       {
>         "id": "03ec6c583cd7cadd7c03ecba4042ee4cca35d4f56cae49c2fd7abf71380ba04043",
>         "channel": "1291977:56:0",
>         "msatoshi": 903714309,
>         "delay": 6
>       }
>     ],
>     "failures": []
>   }
> }
>

FeedTheWeb:
nevermind... my seems to be because of something in my scripts

ZmnSCPxj:
Strange. Paying with too little money should have led to an eventual exhaustion of available routes (since channels with too little money will be marked as unuseable and thus not-routable-over), and a route-not-found 205 error.  Did you modify `lightningd`?

The reason for the ridiculous fee is an underflow of the computation `route[0].msatoshi - payment`.  `route[0].msatoshi` is 903714309, and `payment` is 142637635077.  The difference is -141733920768.  In unsigned 64-bit that is 18446743931975630848.

However the routing component will target exactly the `payment` you gave it, not the amount in the channel.  So it seems the `route[0].msatoshi` value was modified after the routing component returned it, which is not the programmed behavior and the reason I ask if you modified `lightningd`.  Can you give version and the BOLT11 invoice?

FeedTheWeb:
it is a direct connect between 2 CL nodes. 
I just created a new invoice with the same amount ($10k) , which is obviously too much

1.   v0.5.2-2016-11-21-2493-g723b640
```
lightning-cli invoice 142637635077 test test
{
  "payment_hash": "4d3db6135c5b0100dfbe656dd07007a6323a3257f287db0a3fa087210702a7ed", 
  "expiry_time": 1523264293, 
  "expires_at": 1523264293, 
[...]

PR #1350, opened 2018-04-09
routing: Use 64-bit msatoshi for messages to and from routing.

ZmnSCPxj:
Fixes: #1344

Internally both payment and routing use 64-bit, but the interface
between them used 32-bit.
Since both components already support 64-bit we should use that.


ZmnSCPxj:
We also need to impose a limit on invoices (`channeld/channel.c` already imposes the limit for payments) so that invoices greater than the LN BOLT limit for payments are not createable.  But maybe separate patchset as that is arguably a separate problem.

Issue #1341, opened 2018-04-07
Crashes every start with "STATUS_FAIL_INTERNAL_ERROR: Announcement ... was replaced?"


Ulmo:
### Issue and Steps to Reproduce

Every time lightningd starts, it says this:

2018-04-07T14:43:44.998Z lightning_gossipd(32734): STATUS_FAIL_INTERNAL_ERROR: Announcement 0100be25012839603159d89806034f4b4ac40bcd05bf70a8753887d513820099f0ed13d9c44083a8ed1999f79b7abaa279d099ade31c78ffefdac5495a0fc652cd87e052e7f31261023db8953cb030444ed7124eb9b66d512d544b37f8246c86f6897fee91d0be2b588a081e4334d1976d90ee1e14edd2443368760f076aa16f12a7b112d5fa7e93f37edf0e65b359a00ebb0b8d105a907e4ebfabd62c10ac21333a542832137c0b58b2afe3317c46d517f4aac9b23697fe357cc7a95e914ac10961f2c549fbd8564da62d1e282a3342855e95be99c2709c4a8bbd7c769cd9c55d070788cf8887e5fff4cc967bd0b9c8935644b79955f122b476443e87e95ae5b60f00006fe28c0ab6f1b372c1a6a246ae63f74f931e8365e15a089c68d619000000000007c70100029800000371190acfb2e92bd1faa6ce4d12ff248798515a92ac903ac14d31b5172d9b2917037b7cd33aada9dd91ccb95857d8f71c34fd31c426057016f28f48b0633ae844cd028ad3936869eb28a6296ee18611621937b895e4b696bfa62e430e3746ea06ae5b023cdaad34e871ab8f3dbbc5ad9a806ae56d27937dfa3903634312e84970ce7bd2 was replaced?

This happened first on 2018-04-06T09:12:50.115Z, with the following at the top of the git log file.  By the way, why is my name there?  And why does it make me edit a file every time I issue the command "git pull origin master"?  Who wrote GIT, a bunch of upside down people?  It is very confusing.  I understand version control perfectly, but the words GIT uses are backwards and confusing.

commit 35b0d1f28eaaa7c195cfd4e5aa637e1e1b5b907c
Merge: fc54a8e 9ab28d1
Author: Ulmo <Ulmo+GIT@Sonic.Net>
Date:   Fri Apr 6 09:07:49 2018 +0000

    Merge branch 'master' of https://github.com/ElementsProject/lightning

[...]

PR #1352, opened 2018-04-10
gossipd: improve diag for the weird announce duplicate, make store more robust

rustyrussell:
Technically this "fixes" #1341 but we still need the root cause.

### DIFF
--- a/gossipd/gossip_store.c
+++ b/gossipd/gossip_store.c
@@ -170,10 +170,7 @@ void gossip_store_load(struct routing_state *rstate, struct gossip_store *gs)
 		known_good += sizeof(belen) + msglen;
 		tal_free(msg);
 	}
-	status_trace("gossip_store: Read %zu/%zu/%zu/%zu cannounce/cupdate/nannounce/cdelete from store in %"PRIu64" bytes",
-		     stats[0], stats[1], stats[2], stats[3],
-		     (u64)known_good);
-	return;
+	goto out;
 
 truncate:
 	status_unusual("gossip_store: %s (%s) truncating to %"PRIu64,
@@ -182,4 +179,8 @@ truncate_nomsg:
 	if (ftruncate(gs->fd, known_good) != 0)
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Truncating store: %s", strerror(errno));
+out:
+	status_trace("gossip_store: Read %zu/%zu/%zu/%zu cannounce/cupdate/nannounce/cdelete from store in %"PRIu64" bytes",
+		     stats[0], stats[1], stats[2], stats[3],
+		     (u64)known_good);
 }
--- a/gossipd/gossip_store.c
+++ b/gossipd/gossip_store.c
@@ -174,9 +174,13 @@ void gossip_store_load(struct routing_state *rstate, struct gossip_store *gs)
 
 truncate:
 	status_unusual("gossip_store: %s (%s) truncating to %"PRIu64,
-		       bad, tal_hex(msg, msg), (u64)known_good);
+		       bad, tal_hex(msg, msg), (u64)1);
 truncate_nomsg:
-	if (ftruncate(gs->fd, known_good) != 0)
+	/* FIXME: We would like to truncate to known_good, except we would
+	 * miss channel_delete msgs.  If we put block numbers into the store
+	 * as we process them, we can know how far we need to roll back if we
+	 * truncate the store */
+	if (ftruncate(gs->fd, 1) != 0)
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Truncating store: %s", strerror(errno));
 out:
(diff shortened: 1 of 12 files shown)


ZmnSCPxj:
ACK 660e8524ba765d76be96917f15ad9551fbbe0076

rustyrussell:
Fixed valgrind-found issue where routing_add_channel_announcement() would take() the msg even if it wasn't supposed to.

gossipd/routing.c
```diff
@@ -1246,9 +1246,9 @@ u8 *handle_node_announcement(struct routing_state *rstate, const u8 *node_ann)
 	status_trace("Received node_announcement for node %s",
 		     type_to_string(tmpctx, struct pubkey, &node_id));
 
-	gossip_store_add_node_announcement(rstate->store, serialized);
 	applied = routing_add_node_announcement(rstate, serialized);
 	assert(applied);
+	gossip_store_add_node_announcement(rstate->store, serialized);
```
cdecker: Good catch!

gossipd/gossip_store.c
```diff
@@ -96,58 +97,69 @@ void gossip_store_add_channel_delete(struct gossip_store *gs,
...
+	while (read(gs->fd, &belen, sizeof(belen)) == sizeof(belen)) {
+		msglen = be32_to_cpu(belen);
+		msg = tal_arr(gs, u8, msglen);
+
+		if (read(gs->fd, msg, msglen) != msglen) {
+			status_unusual("gossip_store: truncated file?");
+			goto truncate_nomsg;
+		}
+
+		if (fromwire_gossip_store_channel_announcement(msg, msg,
```
cdecker: Any reason we are moving away from the `peektype` / `switch` kind of routing?

gossipd/gossip_store.c
```diff
@@ -96,58 +97,69 @@ void gossip_store_add_channel_delete(struct gossip_store *gs,
...
+			if (!c) {
+				bad = "Bad channel_delete";
+				goto truncate;
+			}
+			tal_free(c);
+		} else {
+			bad = "Unknown message";
+			goto truncate;
+		}
+		known_good += sizeof(belen) + msglen;
```
cdecker: We should probably just nuke the entire store. With this partial persistence we may end up adding channels that had a tombstone later in the store, giving us a myopic  view of the network as a whole.

gossipd/gossip_store.c
rustyrussell: It's done internally in those functions, so it was redundant.

gossipd/gossip_store.c
rustyrussell: The store can get pretty large, so I'd rather not; I'm thinking of the gossip enhancements where we can efficiently ask for information past some timestamp.

rustyrussell:
Hmm, cool store bugs, seem to be sometimes getting a node_announcement as the first entry in the store (which Shouldn't Happen of course).   Spinning now to reproduce, seems like about 1 in 10.

rustyrussell:
First two patches fix the node_announce issue (which was real, and independent).  Fixed some valgrind problems caused by moving store after message receipt, too.

Ready for re-review.

gossipd/gossip_store.c
rustyrussell: Ah,I see what you're saying.  We probably need to put block numbers into the gossip store, and if we truncate then we replay from that block?  For the moment, I'll truncate to zero.

gossipd/routing.c
```diff
@@ -1187,6 +1195,8 @@ u8 *handle_node_announcement(struct routing_state *rstate, const u8 *node_ann)
 	/* Beyond this point it's not malformed, so safe if we make it
 	 * pending and requeue later. */
 	node = get_node(rstate, &node_id);
+	if (!node_has_public_channels(node))
```
cdecker: `node_has_public` dereferences `node` which may be `NULL`, this needs a check.

gossipd/gossip_store.c
cdecker: Going back to that block is also bad, because it means we might be syncing several thousand blocks. Truncating to zero is correct(-ish) IMHO, since this is ephemeral information that can be recovered by connecting to a peer.

rustyrussell: Yes, there has to be a limit.  In practice I think recent corruption is more likely, so it might be worth some heuristic in future.

gossipd/routing.c
rustyrussell: That will teach me: I originally rewrote the code, then decided it was clearer as two patches: the  minimal fix, then the rework.  So I hand-hacked in the minimal version.

Fixed!

gossipd/gossip_store.c
cdecker: That sort of implies the order on the announcements matching the blockheight progression, which is not true, e.g., I could learn about a really old channel and then get a corruption, causing us to go back to that old block.

cdecker:
ACK a877608

cdecker:
I think this has resolved the root cause of #1341, would you agree @rustyrussell ?

In that case this also fixes #1363 and #1334

Issue #1207, opened 2018-03-13
Timeout on `close`, then force unilateral.


ZmnSCPxj:
I propose we add a timeout on `close` command, and, if not complete bilateral negotiation by timeout, force a unilateral close.

I think this also implicitly includes `dev-fail` functionality: set the timeout to 0 and you will immediately try a forced unilateral close.

(`dev-fail` seems too useful for getting things unstuck to be a dev-only command).

Issue #714, opened 2018-01-22
closing a channel should output more info


thestick613:
`lightning-cli close xyxyxyxyxyxyxyx...xxxxxxxxxxxx`
outputs only
`{  }`

Running it a second time outputs:
```
lightning-cli close xyxyxyxyxyxyxyx...xxxxxxxxxxxx
"Peer is in state CHANNELD_SHUTTING_DOWN"
```
I guess that the first run should output the closing transaction id, for more verbosity.

ZmnSCPxj:
It should also probably output whether it was able to do a bilateral close, or used a unilateral one.

alaniz3:
It seems to me like a more elaborate response should be added here, https://github.com/ElementsProject/lightning/blob/master/lightningd/peer_control.c#L2721, I'm not familiar enough with the codebase yet to submit a PR to show the closing txid but any guidance (and/or) a similar example would be interesting to me to further understand the codebase. As far as I understand each of these 'functions' has its own daemon but its not clear to me how peer_control.c communicates with say closingd in order to be able to use the response from closingd in the command_success response.

ZmnSCPxj:
It is `channeld` which actually initiates the closing.  I am unsure how `closingd` works in detail myself.

Look up files ending in `csv` --- these contain message definitions.  In general a `csv` file in a subdaemon directory is for messages that subdaemon uses.  By convention, a message pair, with the same prefix and with one ending in `request` and the other ending in `reply`, is a request-reply pair (obviously, although this is not a strict convention and there are a few violations).  In the code you pointed to, notice that the function `towire_channel_send_shutdown` is called a few lines above, to get the message sent by `subd_send_msg`; `towire_channel_send_shutdown` creates the serialization of the `channel_send_shutdown` message (whose definition you can find in `channeld/channel_wire.csv`).  Note that `subd_send_msg` is a one-sided send, meaning it does not wait for a reply; this means that `channeld` does not report the result of closing to `lightningd`.

You might occassionally see references to `masterd`, this is basically the main `lightningd` binary.

[...]

PR #1355, opened 2018-04-10
peer_control: Make close wait for complete closure, with timeout.

ZmnSCPxj:
Also report tx and txid, and whether we closed unilaterally or
bilaterally, if we could close the channel.

Also make a manpage.

Fixes: #1207
Fixes: #714
Fixes: #622

### DIFF
--- a/doc/lightning-close.7.txt
+++ b/doc/lightning-close.7.txt
@@ -1,0 +1,70 @@
+LIGHTNING-CLOSE(7)
+==================
+:doctype: manpage
+
+NAME
+----
+lightning-close - Protocol for closing channels with direct peers
+
+SYNOPSIS
+--------
+*close* 'id' ['force'] ['timeout']
+
+DESCRIPTION
+-----------
+
+The *close* RPC command attempts to close the channel cooperatively
+with the peer.
+It applies to the active channel of the direct peer corresponding to
+the given peer 'id'.
+
+The *close* command will time out and return with an error when the
+number of seconds specified in 'timeout' is reached.
+If unspecified, it times out in 30 seconds.
+
+The 'force' argument, if the JSON value 'true', will cause the
+channel to be unilaterally closed when the timeout is reached.
+If so, timeout will not cause an error, but instead cause the
+channel to be failed and put onchain unilaterally.
+Unilateral closes will lead to your funds getting locked according
+to the 'to_self_delay' parameter of the peer.
+
+Normally the peer needs to be live and connected in order to negotiate
+a mutual close.
+Forcing a unilateral close can be used if you suspect you can no longer
+contact the peer.
+
+RETURN VALUE
+------------
+
+On success, an object with fields 'tx' and 'txid' containing the
+closing transaction are returned.
+It will also have a field 'type' which is either the JSON string
+'mutual' or the JSON string 'unilateral'.
+A 'mutual' close means that we could negotiate a close with the
+peer, while a 'unilateral' close means that the 'force' flag was
+set and we had to close the channel without waiting for the
+counterparty.
+
+A unilateral close may still occur with 'force' set to 'false' if
+the peer did not behave correctly during the close negotiation.
+
+Unilateral closes will return your funds after a delay.
+The delay will vary based on the peer 'to_self_delay' setting, not
+your own setting.
+
+On failure, if *close* failed due to timing out with 'force'
+argument 'false', the channel will still eventually close once
+we have contacted the peer.
+
+AUTHOR
+------
+ZmnSCPxj <ZmnSCPxj@protonmail.com> is mainly responsible.
+
+SEE ALSO
+--------
+
+
+RESOURCES
+---------
+Main web site: https://github.com/ElementsProject/lightning
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -44,6 +44,17 @@
 #include <wally_bip32.h>
 #include <wire/gen_onion_wire.h>
 
+struct close_command {
+	/* Inside struct lightningd close_commands. */
+	struct list_node list;
+	/* Command structure. This is the parent of the close command. */
+	struct command *cmd;
+	/* Channel being closed. */
+	struct channel *channel;
+	/* Should we force the close on timeout? */
+	bool force;
+};
+
 static void destroy_peer(struct peer *peer)
 {
 	list_del_from(&peer->ld->peers, &peer->list);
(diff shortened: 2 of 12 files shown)


rustyrussell:
This closely mirrors my https://github.com/rustyrussell/lightning/tree/guilt/remove-dev-fail branch which I obviously forgot to put a PR in for!  But yours is better anyway...

I avoided putting in a timeout for simplicity, but OTOH you return more information from close, which is nice but makes the timeout a necessity. You're clearly not from Australia though, because 2 seconds is grossly insufficient as a default :)

You also force a connection initiation to the peer, which is nice.  And wrote the man page!

Can I ask you to cherry-pick commits  or steal stuff as appropriate from my branch?

ZmnSCPxj:
> You're clearly not from Australia though, because 2 seconds is grossly insufficient as a default :)

I am also obviously not an embedded semi-independent agent in an AI agent network sharing computational and network resources with other competing and cooperating knowledge agents, especially since 2 seconds would also be grossly insufficient as a default in that situation.

What would be a plausible default for a human being in Australia?

>Can I ask you to cherry-pick commits or steal stuff as appropriate from my branch?

Okay, I shall check.  There are no useful tests for the new closing functionality in particular, so if I see any tests I will steal them.

rustyrussell:
30 seconds seems to be the modern default.

ZmnSCPxj:
Rebased, changed timeout to 30 seconds, added some code from @rustyrussell , please review again.

doc/lightning-close.7.txt
```diff
@@ -0,0 +1,58 @@
...
+channel to be unilaterally closed when the timeout is reached.
+If so, timeout will not cause an error, but instead cause the
+channel to be failed and put onchain unilaterally.
+Unilateral closes will lead to your funds getting locked according
+to the 'to_self_delay' parameter of the peer.
+
+Normally the peer needs to be live and connected in order to negotiate
+a bilateral close.
+Forcing a unilateral close can be used if you suspect you can no longer
+access the peer.
```
cdecker: access -> contact

doc/lightning-close.7.txt
```diff
@@ -0,0 +1,58 @@
...
+
+RETURN VALUE
+------------
+
+On success, an object with fields 'tx' and 'txid' containing the
+closing transaction are returned.
+It will also have a field 'type' which is either the JSON string
+'bilateral' or the JSON string 'unilateral'.
```
cdecker: I usually use collaborative close to describe the bilateral close, which is a bit nicer. The spec talks about a "mutual close" or "cooperative close" which is also slightly nicer than bilateral (and you use cooperative in the code later).

lightningd/peer_control.c
```diff
@@ -44,6 +44,17 @@
 #include <wally_bip32.h>
 #include <wire/gen_onion_wire.h>
 
+struct close_command {
+	/* Off struct lightningd close_commands. */
```
cdecker: Off -> Of?

lightningd/peer_control.c
```diff
@@ -209,13 +220,125 @@ static void remove_sig(struct bitcoin_tx *signed_tx)
...
+					 struct close_command *cc)
+{
+	/* The cc has the command as parent, so resolving the
+	 * command destroys the cc and triggers destroy_close_command.
+	 * Clear the cc->channel first so that we will not try to
+	 * remove a destructor. */
+	cc->channel = NULL;
+	command_fail(cc->cmd, "Channel forgotten before proper close.");
+}
+/* Destroy the close command structure. */
```
cdecker: Nit: one empty line between functions.

tests/test_lightningd.py
```diff
@@ -4205,7 +4226,9 @@ def test_peerinfo(self):
         assert l1.rpc.getpeer(l2.info['id'])['color'] == l1.rpc.listnodes(l2.info['id'])['nodes'][0]['color']
 
         # Close the channel to forget the peer
-        l1.rpc.close(l2.info['id'])
+        self.assertRaisesRegex(ValueError,
```
cdecker: I don't understand why now all of these closes fail in the integration tests. Aren't these supposed to be collaborative closes? The 0 timeout everywhere is a bit excessive IMHO

lightningd/channel.h
```diff
@@ -64,6 +64,9 @@ struct channel {
 	/* Channel if locked locally. */
 	struct short_channel_id *scid;
 
+	/* Closing txid, for reporting to JSON */
+	struct bitcoin_txid *closing_txid;
```
cdecker: This field will not be repopulated when restarting / loading from DB, could that lead us into an inconsistent behavior? Before restart we show the `closing_txid` after the restart we lose that. If we broadcast the transaction we'll have it in the `transactions` table btw.

tests/test_lightningd.py
ZmnSCPxj: The tests assume that `close` does not wait for the close to complete.  This PR makes `close` wait either for the timeout or the close to complete; to emulate the old behavior we set timeout to 0 and let the command fail.  Otherwise I would have to check each instance of `.rpc.close` to see if I can remove the `wait_for_logs` afterwards or if they scan for something more substantial than bilateral close.

doc/lightning-close.7.txt
ZmnSCPxj: Hmm.  Grepping `lightning-rfc`:

* cooperative: 1 time, in 00-introduction.md
* bilateral: 0 times, ha
* mutual: many times, in 00-introduction.md, 02-peer-protocol.md, 05-onchain.md

Perhaps we should go with 'mutual' instead, as that seems to be the preferred term in the RFC.  Note that the BOLT spec uses 'mutual' and 'unilateral'.

tests/test_lightningd.py
cdecker: Well, ok, we should slowly make these closes normal again though (noting this for my cleanup queue).

ZmnSCPxj:
Rebased, fixed nits.  Removed `closing_txid` from `listpeers` command instead as I decided it was complicated to have to save it to the DB too, and the new `close` now reports the transaction that actually closes the channel.  Also did some minor cleanups of some calls to the `close` RPC in the test. @cdecker please review again!

cdecker:
Thanks @ZmnSCPxj, I oversaw the removal of dev-fail by @rustyrussell, which I think is a bit eager. Forced closes do have the side effect of potentially sending close messages to the peer, which might not be desired in the test cases, so maybe we should keep dev-fail for integration tests?

ZmnSCPxj:
Okay, I will remove that commit that removes `dev-fail`.

cdecker:
Just kicked Travis, let's see if the failure was a flake

ZmnSCPxj:
Well, I notice that the retry system does not seem to work well with this particular test, sigh.

ZmnSCPxj:
Can you teach me how to remove the rerunfailures? I tried, to run without rerunfailures, as I suspect that an earlier run will show the root cause, and rerunfailures does not completely clear the test environment, so that reruns will fail for spurious reason once the first run fails for the real reason.

I tried removing the rerunfailures here: https://github.com/ZmnSCPxj/lightning/commit/b8925ff5acbb95c42bc7ee0380fc11e228c589b3 but the CI still runs it with rerun.  Hmm....

ZmnSCPxj:
Increasing the timeouts for closing in the `test_closing_different_fees` seems to make it work even in `NO_VALGRIND=0` mode.  Probably just a load problem in the CI (we are running 7 lightningd in parallel and performing 6 closes in parallel).

cdecker:
And to answer your question: rerunfailures is enabled automagically on Travis using the following lines in the Makefile:

https://github.com/ElementsProject/lightning/blob/774af5f8176968be895e528586f5569da1daba93/Makefile#L43-L47

So if you set `TRAVIS=false` it'll not rerun.

cdecker:
I also missed this comment:

> Removed closing_txid from listpeers command instead as I decided it was complicated to have to save it to the DB too, and the new close now reports the transaction that actually closes the channel.

We are now storing all transactions that we broadcast or that we watched in the `transactions` table, so you could just add the txid into the table and then fetch it from there :wink: But we can add that later as well.

ZmnSCPxj:
We already store `last_tx` associated with the channel; the `transactions` table does not have an easy association between channel and transaction.  I just need to figure out when to show it as `closing_tx`/`closing_txid`.  Definitely when it is in `CLOSINGD_COMPLETE`, possibly `ONCHAIN` too (but I am uncertain about the case where we trigger an `error` and the other side gets its commitment onchain before we do; do we track the transaction in this case?).  Let us do it in another PR though, as arguably the important bit (reporting the closing transaction on return of `close` command) is here already and `closing_txid` is an enhancement.

Issue #1354, opened 2018-04-10
Strange loop termination in `closing_control.c`


ZmnSCPxj:
While reading through closing control while building a patchset for #1207, I came across this loop:

https://github.com/ElementsProject/lightning/blob/86290b54d49d183e49f905be6a18bfc65612580e/lightningd/closing_control.c#L34-L35

Seems like an incorrect termination condition, as there is expected to be only a single `last_tx` but it may have multiple outputs.

PR #1356, opened 2018-04-10
closing_control: Fix loop limit in better_closing_fee.

ZmnSCPxj:
Fixes: #1354


rustyrussell:
Holy crap.  Good catch!

Issue #1361, opened 2018-04-11
send `closing_signed` twice


nayuta-ueno:
commit: 957513666c494e35f71cb735eb82faed08f7b594

c-lightning send `closing_signed` twice(below `2` and `3a`).
Our node send `closing_signed` same as `2`'s `fee_satoshis`, and c-lightning disconnect because our `fee_satoshis` isn't "strictly between".
Is it necessary to wait for the second `closing_singed` ?

----
* funder: our node
* fundee: c-lightning

0. (`shutdown` exchanged)
1. our node: send `closing_signed` (fee_satoshis: feeA)
2. c-lightning: send `closing_signed` (fee_satoshis: feeB (<feeA) )
3. message cross
  3a. c-lightning: send `closing_signed`(fee_satoshis: feeC (>feeB) )
  3b. our node: send `closing_signed`(fee_satoshis: feeB)
4. c-lightning send `error` and disconnect( remote offer `feeB` not between `feeC` and `feeA` )
----

```
[Wed Apr 11 17:22:49.455 2018] 2018-04-11T08:22:48.869Z lightningd(123264): lightning_channeld-03d09bdb6da742dd1039e1ff40517c86e331011e86fa0dc26bf13e31dc193959cd chan #1: peer_in WIRE_SHUTDOWN
[Wed Apr 11 17:22:49.455 2018] 2018-04-11T08:22:48.870Z lightning_gossipd(123272): Received channel_update for channel 1292289:8:0(0) now DISABLED
[Wed Apr 11 17:22:49.455 2018] 2018-04-11T08:22:48.870Z lightningd(123264): lightning_channeld-03d09bdb6da742dd1039e1ff40517c86e331011e86fa0dc26bf13e31dc193959cd chan #1: UPDATE WIRE_CHANNEL_GOT_SHUTDOWN
[Wed Apr 11 17:22:49.455 2018] 2018-04-11T08:22:48.870Z lightningd(123264): 03d09bdb6da742dd1039e1ff40517c86e331011e86fa0dc26bf13e31dc193959cd chan #1: State changed from CHANNELD_NORMAL to CHANNELD_SHUTTING_DOWN
[Wed Apr 11 17:22:49.471 2018] 2018-04-11T08:22:48.885Z lightning_gossipd(123272): Received channel_update for channel 1292289:8:0(0) now DISABLED
[...]

PR #1362, opened 2018-04-11
closingd: Correct whose turn it is when moving from initial negotiation to continuous negotiation.

ZmnSCPxj:
Fixes: #1361 

Reported-by: @nayuta-ueno

### DIFF
--- a/closingd/closing.c
+++ b/closingd/closing.c
@@ -338,7 +338,8 @@ struct feerange {
 
 static void init_feerange(struct feerange *feerange,
 			  u64 commitment_fee,
-			  const u64 offer[NUM_SIDES])
+			  const u64 offer[NUM_SIDES],
+			  bool allow_mistakes)
 {
 	feerange->min = 0;
 
@@ -349,6 +350,7 @@ static void init_feerange(struct feerange *feerange,
          *    in [BOLT #3](03-transactions.md#fee-calculation).
 	 */
 	feerange->max = commitment_fee;
+	feerange->allow_mistakes = allow_mistakes;
 
 	if (offer[LOCAL] > offer[REMOTE])
 		feerange->higher_side = LOCAL;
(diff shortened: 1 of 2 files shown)


rustyrussell:
Yech!  Valgrind didn't spot the uninitialized variable because it was on the stack :(

Fixing that in a separate patch should reveal the error, I think.  Let me try...

ZmnSCPxj:
I have had a dozen different tries at the fix, they all fail for some reason; I will probably steal your code.

Also the uninitialized `allow_mistakes` field *was* caught in a Valgrind run of one of the various versions I was trying.  It is simply that the old version never managed to even enter that part of the code, since all our tests involve positive testing that both sides could come to an agreement.  Most of my various versions managed to diverge, entering the `feerange->allow_mistakes` check and triggering a Valgrind error.

rustyrussell:
Most failures are subtle: we restrict the range to "previous value +/- 1", so you can't get stuck offering the same value over and over.  The one exception to this, of course, is the termination condition where both agree, so this always has to be checked before calling adjust_feerange().

In fact, adjust_feerange() could detect this case and return true, which might simplify our logic.

closingd/closing.c
```diff
@@ -530,18 +530,12 @@ int main(int argc, char *argv[])
...
 	/* Now any extra rounds required. */
 	while (offer[LOCAL] != offer[REMOTE]) {
-		/* If they differ, adjust feerate. */
-		adjust_feerange(&cs, gossip_index, &channel_id, &feerange,
-				offer[whose_turn], whose_turn);
-
-		/* Now its the other side's turn. */
-		whose_turn = !whose_turn;
+		/* Still don't agree: adjust feerange based on previous offer */
+		adjust_feerange(&cs, gossip_index, &channel_id,
```
ZmnSCPxj: It seems to me the below situation will erroneously pass?

0. `commitment_fee` = 1000
1. funder: 800
2. fundee: 200
3. funder: 900

`init_feerange` sets min 0, max 1000.  On entering the `while` loop we load the fundee fee into the feerange and set min 201, max 1000, but we do not constrain the funder side simllarly?

rustyrussell: Agreed.  Fix coming...

rustyrussell: Try those...

cdecker:
ACK 7b66689

PR #1364, opened 2018-04-11
Persistence transaction tracking in the database

cdecker:
This is some preparatory work to get to no-rescan startups. We store transactions we own, watch or broadcast in a DB table. This allows us to track their depth and recover the transactions for debugging purposes. The transactions will also be used by the `onchaind` replay (next PR) to feed transactions in and restore its state.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2206,7 +2206,7 @@ u32 wallet_transaction_height(struct wallet *w, const struct bitcoin_txid *txid)
 {
 	u32 blockheight;
 	sqlite3_stmt *stmt = db_prepare(
-		w->db, "SELECT blockheight, txindex, rawtx FROM transactions WHERE id=?");
+		w->db, "SELECT blockheight FROM transactions WHERE id=?");
 	sqlite3_bind_sha256(stmt, 1, &txid->shad.sha);
 
 	if (sqlite3_step(stmt) != SQLITE_ROW) {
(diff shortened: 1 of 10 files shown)


wallet/wallet.c
```diff
@@ -2158,3 +2158,46 @@ struct outpoint *wallet_outpoint_for_scid(struct wallet *w, tal_t *ctx,
...
+			    const u32 blockheight, const u32 txindex)
+{
+	struct bitcoin_txid txid;
+	sqlite3_stmt *stmt = db_prepare(w->db, "SELECT blockheight FROM transactions WHERE id=?");
+	bool known;
+
+	bitcoin_txid(tx, &txid);
+	sqlite3_bind_sha256(stmt, 1, &txid.shad.sha);
+	known = sqlite3_step(stmt) == SQLITE_ROW;
+	sqlite3_finalize(stmt);
```
rustyrussell: This seems like a debugging check, in which case it'd be better to check that the other fields haven't changed?  Plus add a /*FIXME: change to INSERT OR UPDATE */ for later optimization?

wallet/wallet.c
```diff
@@ -2218,3 +2218,32 @@ u32 wallet_transaction_height(struct wallet *w, const struct bitcoin_txid *txid)
...
+
+	if (sqlite3_step(stmt) != SQLITE_ROW) {
+		goto fail;
+
+	}
+
+	if (!sqlite3_column_int(stmt, 0))
+		goto fail;
+
+	loc = talz(ctx, struct txlocator);
```
rustyrussell: Avoid talz, use tal.

wallet/wallet.c
```diff
@@ -2218,3 +2218,32 @@ u32 wallet_transaction_height(struct wallet *w, const struct bitcoin_txid *txid)
...
+	stmt = db_prepare(
+	    w->db, "SELECT blockheight, txindex FROM transactions WHERE id=?");
+	sqlite3_bind_sha256(stmt, 1, &txid->shad.sha);
+
+	if (sqlite3_step(stmt) != SQLITE_ROW) {
+		goto fail;
+
+	}
+
+	if (!sqlite3_column_int(stmt, 0))
```
rustyrussell: Maybe an explicit sqlite3_column_null ?

wallet/wallet.c
rustyrussell: I know, cut & paste, but still an antipattern.

wallet/wallet.c
cdecker: I'm not sure I get the question. I could change this to become `SELECT count(*) FROM transactions WHERE id=?` and then check that I get a 1 as a result, but the effect is the same, either I have a row or I don't, and depending on which it is I need to insert or update.

Or are you saying we should fetch all fields and check that they are identical?

PR #1376, opened 2018-04-14
Make deprecated out-of-range fee handling deterministic

practicalswift:
Make deprecated out-of-range fee handling deterministic.

Prior to this commit `feerange->allow_mistakes ` had an indeterminate value (an unspecified value or a trap representation).


ZmnSCPxj:
Duplicates: #1362 which also adds this fix among others.

practicalswift:
Oh, I missed that one. Good that we found it independently of each other, and also using different methods (I used static analysis and you used dynamic analysis) :-)

PR #1377, opened 2018-04-14
Assert our parse_wireaddr assumptions in run-ip_port_parsing.c

practicalswift:
Assert our `parse_wireaddr` assumptions in `run-ip_port_parsing.c`.

We assume that the `parse_wireaddr(...)` calls return `true`. Otherwise `addr.port` will be uninitialized.


ZmnSCPxj:
ACK 9f4e29bbdcd09818574cf1f4ba4298339967218a

PR #1379, opened 2018-04-14
Avoid static analyzer warnings about potential null pointer dereferences and uninitialized values

practicalswift:
Avoid static analyzer warnings about potential null pointer dereferences and uninitialized values.


gossipd/gossip.c
```diff
@@ -1426,7 +1426,7 @@ static void gossip_refresh_network(struct daemon *daemon)
 static struct io_plan *connection_in(struct io_conn *conn, struct daemon *daemon)
 {
 	struct wireaddr addr;
-	struct sockaddr_storage s;
+	struct sockaddr_storage s = {};
 	socklen_t len = sizeof(s);
 
 	if (getpeername(io_conn_fd(conn), (struct sockaddr *)&s, &len) != 0) {
```
ZmnSCPxj: Does not either `getpeername` initialize `s`, or we return immediately with a close of the socket?

gossipd/gossip.c
```diff
@@ -1485,7 +1485,7 @@ static void setup_listeners(struct daemon *daemon, u16 portnum)
 	/* IPv6, since on Linux that (usually) binds to IPv4 too. */
 	fd1 = make_listen_fd(AF_INET6, &addr6, sizeof(addr6), true);
 	if (fd1 >= 0) {
-		struct sockaddr_in6 in6;
+		struct sockaddr_in6 in6 = {};
 
 		len = sizeof(in6);
 		if (getsockname(fd1, (void *)&in6, &len) != 0) {
```
ZmnSCPxj: Does not `getsockname` either initialize `in6` or we do not use `in6` at all?

gossipd/test/run-find_route.c
```diff
@@ -189,6 +189,7 @@ int main(void)
 	add_connection(rstate, &a, &b, 1, 1, 1);
 
 	route = find_route(tmpctx, rstate, &a, &b, 1000, riskfactor, 0.0, NULL, &fee);
+	assert(route);
```
ZmnSCPxj: `tal_count(NULL)` returns 0, so I believe `assert(tal_count(route) == 1)` implicitly covers `assert(route)`.  Also applies for succeeding changes.

gossipd/gossip.c
practicalswift: But `getpeername(…)` doesn't necessarily fill the entire `struct sockaddr_storage` (note that `s` is a `struct sockaddr_storage`). That's why the static analyzer is complaining, right? :-)

gossipd/gossip.c
practicalswift: See previous comment :-)

gossipd/test/run-find_route.c
practicalswift: Yes, you're right! I'm afraid the static analyzer needs some guidance to understand that `assert(tal_count(route) == 1)` implies `assert(route)` :-)

cdecker:
Seems innocent enough, assuming the initializations in the tests are ok for now.

ACK 83c7cbe

Issue #1353, opened 2018-04-10
Feature Request: RPC endpoint for Watch Tower (Simple)


NicolasDorier:
Ideally, I would like clightning to have very simple watch tower capabilities.

The use case I have in mind has quite simple UX:

Merchant A knows Merchant B:

Merchant A share his watch tower endpoint (URI exposing a REST API made by BTCPAY) to Merchant B.

Merchant B can then configure his store to send watch information after every lightning payment received to the store. (Merchant B don't have to care to send watch information when he spends the money)

BTCPay would just relay this information to the underlying clightning instance.

Note that my case is quite simple: Merchant A knows Merchant B beforehand so no need of "discovery protocol". There is also no need of P2P level protocol at clightning level, as BTCPay is the one exposing the REST endpoints. Only things needed is a RPC method at clightning level to do this.

ZmnSCPxj:
Hi @NicolasDorier, sorry but I cannot visualize exactly what you want as interface. Do you need some serialization of a specific channel state and RPC calls to  get current channel state, and override that state? Or something else entirely?

ZmnSCPxj:
Okay, that seems a little more designable.

Note however that currently our blockchain monitoring seems a little underoptimized:  #1337.

Also ping @rustyrussell as I am unaware of status of plans for a "real" watchtower.  I understand how merchants would want to have some assurance of channel justice even though a "real" watchtower system has not completed at LN yet.

Now, first some things:

* In LN, a commitment transaction held by party A has two branches: a CSV+sigA branch, and a combined (sigA+sigB) branch which happens to have a single priv and pubkey pair (do not ask me how, I never got around to studying the math).  The party A revokes its old commitment by providing sigA to party B.  Hashes are not used in the justice branch.
* What we monitor is spending of the funding transaction. When we see the funding transaction spent when we know we have not broadcast our commitment transaction or a cooperative bilateral close, then we assume it is the other party commitment transaction and try to get the justice branch of the contract.
* We store revocation keys using a "shachain" structure (do not ask me how that looks like, I never got around to studying the math). The shachain is O(1) storage for an indefinite number of revocation keys.
* Every transfer through a channel involves 2 updates: 1st to transfer funds to an HTLC, 2nd to transfer funds from HTLC to the destination.  So we need to provide two revocation keys to the watchtower for each transfer through a channel.

A rough draft would be something like:

[...]

PR #1381, opened 2018-04-15
peer_control: Implement `listrevokeinfos` command.

ZmnSCPxj:
Fixes: #1353

Purpose of this command is to allow TRUSTED WatchTowers to steal
from you, or if they actually do their job, steal it back from
someone who steals from you.

Ping @NicolasDorier

### DIFF
--- a/doc/lightning-listrevokeinfos.7
+++ b/doc/lightning-listrevokeinfos.7
@@ -1,0 +1,152 @@
+'\" t
+.\"     Title: lightning-listrevokeinfos
+.\"    Author: [see the "AUTHOR" section]
+.\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
+.\"      Date: 04/15/2018
+.\"    Manual: \ \&
+.\"    Source: \ \&
+.\"  Language: English
+.\"
+.TH "LIGHTNING\-LISTREVOK" "7" "04/15/2018" "\ \&" "\ \&"
+.\" -----------------------------------------------------------------
+.\" * Define some portability stuff
+.\" -----------------------------------------------------------------
+.\" ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.\" http://bugs.debian.org/507673
+.\" http://lists.gnu.org/archive/html/groff/2009-02/msg00013.html
+.\" ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.ie \n(.g .ds Aq \(aq
+.el       .ds Aq '
+.\" -----------------------------------------------------------------
+.\" * set default formatting
+.\" -----------------------------------------------------------------
+.\" disable hyphenation
+.nh
+.\" disable justification (adjust text to left margin only)
+.ad l
+.\" -----------------------------------------------------------------
+.\" * MAIN CONTENT STARTS HERE *
+.\" -----------------------------------------------------------------
+.SH "NAME"
+lightning-listrevokeinfos \- List revocation info for channels, for giving to *TRUSTED* WatchTowers\&.
+.SH "SYNOPSIS"
+.sp
+\fBlistrevokeinfos\fR [\fIchannel_id\fR]
+.SH "DESCRIPTION"
+.sp
+The \fBlistrevokeinfos\fR RPC command returns revocation information for all open channels, or for a specific channel\&. The intent is to somehow pass this revocation information to some \fBTRUSTED\fR WatchTower\&.
+.sp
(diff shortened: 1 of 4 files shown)


doc/lightning-listrevokeinfos.7
```diff
@@ -0,0 +1,152 @@
...
+.SH "NAME"
+lightning-listrevokeinfos \- List revocation info for channels, for giving to *TRUSTED* WatchTowers\&.
+.SH "SYNOPSIS"
+.sp
+\fBlistrevokeinfos\fR [\fIchannel_id\fR]
+.SH "DESCRIPTION"
+.sp
+The \fBlistrevokeinfos\fR RPC command returns revocation information for all open channels, or for a specific channel\&. The intent is to somehow pass this revocation information to some \fBTRUSTED\fR WatchTower\&.
+.sp
+Note that this requires a \fBTRUSTED\fR WatchTower: it contains information that would let your couterparty steal funds outright while you are offline\&. You are trusting the WatchTower not to leak this information to channel counterparties\&.
```
NicolasDorier: > it contains information that would let your couterparty steal funds outright while you are offline

How is it different from not having watch tower? Without watch tower a counter party can do the same thing.

Also typo *couterparty*

ZmnSCPxj: Without this, the counterparty has to wait out the entire timelock before it can steal the funds.  If you manage to come online during the timelock, you can revoke the stolen funds.

With cooperation of a WatchTower that has access to `our_revocation_basepoint_secret`, the counterparty does not have to wait at all: it can steal the funds immediately.  Basically: counterparty publishes revoked transaction, WatchTower revokes it *to the counterparty and WatchTower splitting the profits 50/50*.  This is strictly worse and requires that the WatchTower be trusted.  You also cannot safely form a direct channel with the WatchTower.

NicolasDorier:
Having a trusted watch tower seems to defeat the point. My goal was that a web of cooperating merchant agree to watch for each other without trusting each other. 

I thought it was theorically possible to have untrusted watch tower thanks to segwit, am I missing something?

ZmnSCPxj:
Maybe if we immediately generate and keep a justice transaction for each commitment transaction we sign for the other side?  Then the WatchTower would not need to be given `our_revocation_basepoint_secret`.

Sorry, but I know too little of the math involved.  All I know is that Dryja was working on a truly trustless WatchTower system.  I cannot find any traces of discussion regarding a trustless WatchTower, although maybe pre-2017 lightning-dev archives may have some discussion.  Perhaps bring this up to the mailing list instead?

The fact that the WatchTower can generate the justice transaction completely, including its outputs, is the key vulnerability against trustlessness here, I think.  So maybe generating and keeping justice transactions for every signed commitment transaction would work (the WatchTower becomes a simple "if you see this transaction, broadcast that transaction") but that would require significant storage ondisk for every commitment update.

@rustyrussell  may know more but I believe he is on vacation for a week. @cdecker do you know about longer-term plans for WatchTower?

NicolasDorier:
@ZmnSCPxj requiring significant disk space is not too much and issue in my case. Worse come to worse merchants will start pruning randomly some justice transactions. If you have like 10 merchant friends you'll still have reasonable chance to not get frauded.
Also most merchant does not make sub second micro payment for now, the amount of data is negligible.

ZmnSCPxj:
Okay, we could consider instead, the idea to make the justice transaction immediately when an old commitment is revoked.  However we do need to figure out how to coordinate between the c-lightning node and the WatchTower (how c-lightning reports the justice transaction to the WatchTower; does c-lightning store those justice transactions and wait for a command to fetch them and delete them, or does c-lightning contact the WatchTower point somehow and send the justice transaction immediately?).

Also, HTLCs can be revoked in multiple ways depending on which branch is triggered (revoked before being resolved, resolved with preimage, resolved with timeout).  Easiest is to make one justice transaction for the main outputs, and sets of 3 justice transactions for each branch of each HTLC on the commitment transaction, but that increases the blockchain space needed for justice.  Currently we just process the revoked commitment outputs individually to form the correct single justice transaction, but that would require many more variations of the justice transaction if we pre-write it (currently we just react appropriately to how the transactions currently look like onchain).

This seems to require quite a bit of engineering and work, I will let @rustyrussell and @cdecker chip in here.  There are projects to build c-lightning on mobile devices, so we should probably make this recording of all possible justice transactions optional and disable that option on mobile.

NicolasDorier:
@ZmnSCPxj I think it is acceptable to returns the justice transaction in the `waitanyinvoice` call and not saving it on c-lightning side.

Sure, I might miss some of them, but an attacker can't really know which one.

NicolasDorier:
Another idea might be to save it inside the invoice, this would be cleared when the user want to clear his invoices on c-lightning.

ZmnSCPxj:
The issue is that payment forwards and failed payments (e.g. the payer sends a particular delay setting to payee, but suddenly two blocks are discovered in sequence, violating the delay setting and forcing the payee to fail the payment) also revoke commitment transactions.

In particular for failed payments we would not be sure which invoice to assign the justice transactions for the multiple commitment transactions involved, if the invoice completely fails (i.e. expires before it can be successfully paid).

We could set a new mode that rejects payment forwards (!!! but that is bad for the network as a whole and I would not push for it) but we cannot easily assign an invoice for a failed payment that terminates at our node.

So even if we just write and sign justice transactions and publish pairs of (`watched_txid`, `justice_tx`), we need to put those in a different queue that gets reported.  And we would make this a new mode also, since mobile devices will probably not want to keep prepared justice transactions.

ZmnSCPxj:
Closing, as this solution is unsatisfactory.  Now discussing on lightning-dev mailinglist:

https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-April/001188.html
https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-April/001194.html

PR #1385, opened 2018-04-16
lightning-autocleaninvoice.7: Properly generate instead of using an empty file.

ZmnSCPxj:
Strange behavior where `a2x` sometimes does not work at all.


cdecker:
Yeah, `a2x` will fail in Travis because installing it in the builder image would pull in ~2GB of layers (`a2x` depends on latex, which is huge).

ACK a984c79

PR #1388, opened 2018-04-17
Adapt relative weights of channel traversal costs

ZmnSCPxj:
Dynamically adapt relative weights of channel traversal costs, to more likely succeed to pay given the `pay` acceptable cost settings.

### DIFF
--- a/doc/lightning-getroute.7
+++ b/doc/lightning-getroute.7
@@ -2,12 +2,12 @@
 .\"     Title: lightning-getroute
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 02/19/2018
+.\"      Date: 04/17/2018
 .\"    Manual: \ \&
 .\"    Source: \ \&
 .\"  Language: English
 .\"
-.TH "LIGHTNING\-GETROUTE" "7" "02/19/2018" "\ \&" "\ \&"
+.TH "LIGHTNING\-GETROUTE" "7" "04/17/2018" "\ \&" "\ \&"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
@@ -31,7 +31,7 @@
 lightning-getroute \- Protocol for routing a payment (low\-level)\&.
 .SH "SYNOPSIS"
 .sp
-\fBgetroute\fR \fIid\fR \fImsatoshi\fR \fIriskfactor\fR [\fIcltv\fR] [\fIfuzzpercent\fR] [\fIseed\fR]
+\fBgetroute\fR \fIid\fR \fImsatoshi\fR \fIriskfactor\fR [\fIcltv\fR] [\fIfuzzpercent\fR] [\fIseed\fR] [\fIfeecostbias\fR]
 .SH "DESCRIPTION"
 .sp
 The \fBgetroute\fR RPC command attempts to find the best route for the payment of \fImsatoshi\fR to lightning node \fIid\fR, such that the payment will arrive at \fIid\fR with \fIcltv\fR\-blocks to spare (default 9)\&.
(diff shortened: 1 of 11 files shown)


cdecker:
Interesting take on the issue, however these RPC commands are getting really unwieldy, it's hard enough for me to understand what exactly they attempt to do, let alone explain that to someone else. There parameters should not be exposed to the end-users, because they will force us to maintain backwards compatibility even if we chose to go a completely different way.

I think it'd be much better to track delay, fees, and similar, that can result in an invalid route to be returned while computing the route and discarding the attempt in the bfg algo if we overstep any of them. That'd then mean that we effectively only ever return valid routes.

ZmnSCPxj:
It also makes it harder to give a diagnostic to the user.  Currently we can say "the fees were too big, boost maxfeepercent if you want" or "the delay was too big, boost maxdelay if you want".

Hmm.  Possibly we could extend the getroute reply message with "maxfeepercentviolated" and "maxdelayviolated" flags, and still return the "best" route found.  That way `pay` can still report the error in detail rather than just a generic "route not found".  Hmm.

rustyrussell:
Yeah, this is getting out of hand :(

Perhaps we need a getroutes RPC which returns a pruned subset of the graph from A->B so that external programs can make their own evaluations.   This may, of course, turn into something ridiculous (the entire graph!), so needs some careful thought.

ZmnSCPxj:
The issue is that we are targeting multiple criteria (delay, fees) whereas pathfinding algorithms are really designed to target a single criterion (cost).

I do not see `getroute` as a "user-level" interface, and mostly consider it an interface for another programmer to use.  My intent is that `pay` is the "user-level" interface, hence why it provides user-level concerns like `maxdelay` and `maxfeepercent` but not `feecostbias`.  I would argue that `riskfactor` should also be hidden by `pay`.

rustyrussell:
The idea behind getroute was that UI can prompt user to confirm with fees. Though I think that may be unnecessary in practice unless fees are large, and awkward UX if it has to retry another path.

We might deprecate it for a new getroutes which provides all options within some limit, so caller can decide.

ZmnSCPxj:
> We might deprecate it for a new getroutes which provides all options within some limit, so caller can decide.

I imagine the current BFG algorithm would not be useable with this, so we would have to rewrite the routing algorithm from scratch.

So my question is: do we do this pre-0.6 or post-0.6?

Since this is effectively a "generate all routes and filter" algorithm, we could consider the below:

1.  Start with a list of partial routes.  Each partial route is a singly-linked list of channels, the node at which the partial route starts, and a set of nodes that the route has passed through.  Also keep an initially empty list of completed routes.
2.  Initialize the list of partial routes from the destination: create a partial route from each channel of the destination.  We could also add the partial routes from BOLT11 `r` here.
3.  Start of loop.
4.  Filter out partial routes that have violated the `maxdelay` and `maxfee`.
5.  Filter the list of partial routes.  Routes that have reached the source of the payment are moved from the list of partial routes to the list of completed routes. (this filtering can probably be merged with the above filtering).
6.  If the list of partial routes is empty or we have reached `maxhops` iterations, return the list of completed routes.
7.  Perform `mappend` operation (i.e. generate a list for each partial route and append the lists to form a new list): for each partial route, create all extensions of that route that do not go to a node the route already went through (i.e. avoid loops).  This creates a new list of partial routes that replaces the current list of partial routes.
8.  Jump to start of loop.

The above algorithm is O( n^2 ) I believe; this is mitigated by the fact that `maxhops` is 20 or less, and that it returns all possible routes within the limit.

It has the advantage that if the Decker-Russell-Osuntokun channels are used, which would require us to track the highest `csv_delta` for the entire route, we could track that more easily than most standard O(n log n) routing algorithms can do (since this algorithm does not in fact search for the best route).

The the `pay` algorithm can sort the returned routes by some measure of fee and delta, and possibly do a few random swaps of the sorted list to implement route randomization (or just do a complete shuffle: after all, all returned routes match the required specs).  Then since it has all the routes that could be tried, it just goes through the list of routes until it runs out (it could do a delay-and-retry also, restarting with the list after some time, in the hope that

PR #1396, opened 2018-04-20
Clean up codes by adjusting wrong indents

conanoc:
There were 7 spaces in routing.c and 4 spaces in other files. Replaced them with tabs.
I searched them with this script and edited manually.
```
grep -r --exclude bech32.c --exclude signature.c --exclude chainparams.c --exclude options.c --exclude netaddress.c --exclude db.c --exclude wallet.c --exclude *wire.c '^    ' */*.c | grep -v '        '
```

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -474,12 +474,12 @@ static u8 *check_channel_update(const tal_t *ctx,
 }
 
 static u8 *check_channel_announcement(const tal_t *ctx,
-    const struct pubkey *node1_key, const struct pubkey *node2_key,
-    const struct pubkey *bitcoin1_key, const struct pubkey *bitcoin2_key,
-    const secp256k1_ecdsa_signature *node1_sig,
-    const secp256k1_ecdsa_signature *node2_sig,
-    const secp256k1_ecdsa_signature *bitcoin1_sig,
-    const secp256k1_ecdsa_signature *bitcoin2_sig, const u8 *announcement)
+	const struct pubkey *node1_key, const struct pubkey *node2_key,
+	const struct pubkey *bitcoin1_key, const struct pubkey *bitcoin2_key,
+	const secp256k1_ecdsa_signature *node1_sig,
+	const secp256k1_ecdsa_signature *node2_sig,
+	const secp256k1_ecdsa_signature *bitcoin1_sig,
+	const secp256k1_ecdsa_signature *bitcoin2_sig, const u8 *announcement)
 {
 	/* 2 byte msg type + 256 byte signatures */
 	int offset = 258;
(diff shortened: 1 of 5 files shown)


gossipd/routing.c
```diff
@@ -1086,39 +1086,39 @@ static struct wireaddr *read_addresses(const tal_t *ctx, const u8 *ser)
...
+	struct node *node;
+	secp256k1_ecdsa_signature signature;
+	u32 timestamp;
+	struct pubkey node_id;
+	u8 rgb_color[3];
+	u8 alias[32];
+	u8 *features, *addresses;
+	struct wireaddr *wireaddrs;
+	fromwire_node_announcement(tmpctx, msg,
+	                           &signature, &features, &timestamp,
```
ZmnSCPxj: Standard is to use tabs as much as possible, and spaces afterwards for finer control of indent.  So I think for this line and similar, it should have more tabs followed by 0 to 7 spaces.

gossipd/routing.c
```diff
@@ -1086,39 +1086,39 @@ static struct wireaddr *read_addresses(const tal_t *ctx, const u8 *ser)
...
+	tal_free(node->addresses);
+	node->addresses = tal_steal(node, wireaddrs);
+
+	node->last_timestamp = timestamp;
+	memcpy(node->rgb_color, rgb_color, 3);
+	tal_free(node->alias);
+	node->alias = tal_dup_arr(node, u8, alias, 32, 0);
+
+	replace_broadcast(node, rstate->broadcasts,
+	                  &node->node_announce_msgidx,
```
ZmnSCPxj: Similar to `fromwire_node_announcement` case above.

lightningd/gossip_msg.c
```diff
@@ -83,8 +85,8 @@ void fromwire_gossip_getchannels_entry(const u8 **pptr, size_t *max,
 	}
 }
 
-void towire_gossip_getchannels_entry(
-    u8 **pptr, const struct gossip_getchannels_entry *entry)
+void towire_gossip_getchannels_entry(u8 **pptr,
+	const struct gossip_getchannels_entry *entry)
```
ZmnSCPxj: We "should" align the `const` to the `u8` on the previous line, again with tabs followed by spaces.

wallet/walletrpc.c
```diff
@@ -83,7 +83,7 @@ static void wallet_withdrawal_broadcast(struct bitcoind *bitcoind UNUSED,
  * the HSM to generate the signatures.
  */
 static void json_withdraw(struct command *cmd,
-			      const char *buffer, const jsmntok_t *params)
+			  const char *buffer, const jsmntok_t *params)
```
ZmnSCPxj: `const` should align with `struct` on previous line, unless this is an artifact of the `+` in the diff, in which case, ignore this comment.

gossipd/routing.c
conanoc: Done. In fact, there are many files where 8 spaces are used instead of tabs. They could be changed later.

gossipd/routing.c
conanoc: done

lightningd/gossip_msg.c
conanoc: done

wallet/walletrpc.c
conanoc: Yes. This is an artifact of the + in the diff.

cdecker:
ACK 6701155

PR #1398, opened 2018-04-20
Store onchaind state in the database and reduce rescan on startup

### LINKED ISSUE #1168: Feature request:  sync faster using stored state from last time.

It takes only a fraction of a second to crash lightningd, but many hours to sync.  I'd prefer if it at least attempt to save some of the sync state so that when it recovers, it can resume syncing at a less drastically long sync size.

I understand that one of the things you don't want with Lightning Network is a node using old state on a channel.

The hardball way to get around that is to do periodic saves of state, and then when restarting, load the state from a fresh perspective and sync from that periodic save point.  The only items that need to be saved are those items that the node would recover upon a restart by scanning the blocks.  One idea is to simply save every relevant transaction in the blockchain as the node runs, and then know the last fully saved transaction block, and upon restart, load all of those transactions instead of parsing full blocks; that is slower than saving a state and resuming from that state, but it might just be faster than going through full blocks and the code would be less intense.  A hybrid approach of doing both would work, too.  As implementation, both could be programmed, with attention paid to the journaling of transactions first, and get it running first, then keep saving state but not using it to resume, slowly implementing a background "what if" state checker that checks to see if saved state equals journaled transaction state recovery, just to test the idea.  Ideas?

### `getinfo` output

{ "id" : "031c527494f295220d26a1814c18e7db3f4f03203ce1a53a0b74b52ed570a3972a", 
        [ 
                "version" : "v0.5.2-2016-11-21-2207-g31b9b6b", "blockheight" : 506218, "network" : "bitcoin" }

### LINKED ISSUE #1179: Reduce rescanning on startup
Now that we have blockchain tracking with #1117 it would be nice to reduce the need to rescan from a fixed blockheight on every startup. The main issue that I see is that we have `onchaind` that may be running, triggered by seeing a spend on-chain, and not getting triggered after the restart.

A simple proposal would be to have a relative rescan period that is larger than the maximum `onchaind` lifetime, e.g., 288 blocks. However, that period may need to be rather large, also due to HTLCs timeouts.

The more involved solution would be to remember the state of `onchaind` across restarts, though I'm not clear about how much we'd need to remember there, and what format we could use. Again, a simple solution would be to just store the messages we sent to `onchaind` in an append-only log in the DB and pruning once `onchaind` is happy and closes on its own.

What do you guys think makes the most sense in this situation? I have reports from users that take hours to catch up with the blockchain and I think it may also be causing a few of the `awaiting funding locking` issues that get reported (funding depth callback not being triggered on the remote end).

### LINKED ISSUE #1243: LN Mainnet - CHANNELED AWAITING LOCKIN...
Hello,
I have successfully installed and synced Bitcoind & C-Lightning node on Linux Ubuntu 17.10 VPS.
```
root@vmi168564:~/builds/lightning# bitcoin-cli getblockcount
514176
root@vmi168564:~/builds/lightning# cli/lightning-cli dev-blockheight
{ "blockheight" : 514176 }
```

I have funded the channel with US$10 i.e. 0.0013102

https://blockchain.info/address/3HeSsWiscn7wjxCdc6knKyfi2vaXmGWPzu

I have created channel with a node and funded it with 40000 Satoshis.

Its been now more than 15 hours and still the channel state is awaiting lockin only. On blockchain if you see above then it has already get 80+ Confirmations.
Here is what it says.
```
root@vmi168564:~/builds/lightning# cli/lightning-cli listpeers
{ "peers" :
        [
                { "id" : "024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd", "connected" : true, "netaddr" :
                        [ "165.227.69.151:9735" ], "alias" : "CL.rompert.com🔵 ", "color" : "0000ff", "channels" :
                        [
                                { "state" : "CHANNELD_AWAITING_LOCKIN", "owner" : "lightning_channeld", "short_channel_id" : "514097:798:0", "funding_txid" : "3b10a79a607bbfa37ff9367b494f40c215b1bc08ef626166b65695f573dec240", "msatoshi_to_us" : 40000000, "msatoshi_total" : 40000000, "dust_limit_satoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minimum_msat" : 0, "their_to_self_delay" : 144, "our_to_self_delay" : 144, "to_self_delay" : 144, "max_accepted_htlcs" : 483, "status" :
                                        [ "CHANNELD_AWAITING_LOCKIN:Reconnected, and reestablished.", "CHANNELD_AWAITING_LOCKIN:We've confirmed funding, they haven't yet." ] } ] } ] }
```

Kindly give your opinion.
Thank You

### LINKED ISSUE #1246: LN Mainnet - Headache CHANNELD_AWAITING_LOCKIN
I follow this article to make my LN node: 
https://medium.com/@fkayaist/kendi-lightning-network-nodeunuzu-nas%C4%B1l-kurars%C4%B1n%C4%B1z-10-ay-%C3%BCcretsiz-53b71c9f5f10

I use Google Cloud Server. Bitcoin and LN is OK. I sent 0.002BTC to my LN Bitcoin address:
https://blockchain.info/tx/858b36d8f123027ac51e069da8c8714ec4292822dd61352537dbaaaf02e5bb61

I am facing with CHANNELD_AWAITING_LOCKIN error. I tried to connect CL.rompert.com

**Here is the listfunds:**

{ "outputs" : 
        [ 
                { "txid" : "858b36d8f123027ac51e069da8c8714ec4292822dd61352537dbaaaf02e5bb61", "output" : 1, "value
" : 200000, "status" : "confirmed" } ], "channels" : 
        [ 
                { "peer_id" : "024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd", "channel_sat" :
 100000, "channel_total_sat" : 100000, "funding_txid" : "82f2a80d5a672d74f51f81be0975f8bf00b9e228ec3448b7b2867b5657
5fafc8" } ] }

**And listpeers:**

{ "peers" : 
        [ 
                { "id" : "024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd", "connected" : false,
 "channels" : 
                        [ 
                                { "state" : "CHANNELD_AWAITING_LOCKIN", "funding_txid" : "82f2a80d5a672d74f51f81be0
975f8bf00b9e228ec3448b7b2867b56575fafc8", "msatoshi_to_us" : 100000000, "msatoshi_total" : 100000000, "dust_limit_s
atoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minim
um_msat" : 0, "their_to_self_delay" : 144, "our_to_self_delay" : 144, "to_self_delay" : 144, "max_accepted_htlcs" :
 483, "status" : 
                                        [ "CHANNELD_AWAITING_LOCKIN:Received error from peer: channel c8af5f57567b8
6b2b74834ec28e2b900bff87509be811ff5742d675a0da8f283: lightning_channeld: received ERROR channel ALL: Internal error
: Failing due to dev-fail command", "CHANNELD_AWAITING_LOCKIN:Sent reestablish, waiting for theirs" ] } ] } ] }

**Listconfigs**

{ "# version" : "v0.5.2-2016-11-21-2290-g640ff4b", "port" : 9735, "rgb" : "d2691e", "alias" : "botbittrex.com", "pi
d-file" : "lightningd-bitcoin.pid", "channel-update-interval" : 604800, "log-level" : "DEBUG", "log-prefix" : "ligh
tningd(1):", "lightning-dir" : "/root/.lightning", "rpc-file" : "lightning-rpc", "daemon" : "false", "ignore-fee-li
mits" : false, "locktime-blocks" : 144, "max-locktime-blocks" : 432, "anchor-onchain" : 10, "anchor-confirms" : 3, 
"max-anchor-confirms" : 10, "commit-fee-min" : 200, "commit-fee-max" : 2000, "commit-fee" : 500, "default-fee-rate"
 : 40000, "cltv-delta" : 14, "cltv-final" : 8, "max-htlc-expiry" : 720, "bitcoind-poll" : "30s", "commit-time" : "1
0ms", "fee-base" : 1000, "fee-per-satoshi" : 10, "ipaddr" : "35.192.6.142:9735", "network" : "bitcoin", "allow-depr
ecated-apis" : true, "dev-no-reconnect" : "false", "dev-fail-on-subdaemon-fail" : "false", "dev-broadcast-interval"
 : 60000 }

**After reboot my server I can see following logs. I use dev-fail command after waited more than 15 hours. What I have to do? Where is my money? I don't believe LN is the future of bitcoin. IT is very hard to make setup proper node. It should be hard as much as like that.**

root@instance-1:~# docker start bitcoind_mainnet
bitcoind_mainnet
root@instance-1:~# docker run --rm --name lightning --network container:bitcoind_mainnet -v /scratch/bitcoin/mainnet/bitcoind:/root/.bitcoin -v /scratch/bitcoin/mainnet/clightning:/root/.lightning --entrypoint /usr/bin/lightningd cdecker/lightningd:latest --network=bitcoin --rgb=D2691E --alias=botbittrex.com --ipaddr=35.192.6.142 --log-level=debug
2018-03-19T08:56:17.887Z lightningd(1): Not guessing addresses: manually set
2018-03-19T08:56:17.962Z lightningd(1): testing /usr/bin/lightning_channeld
2018-03-19T08:56:18.015Z lightningd(1): testing /usr/bin/lightning_closingd
2018-03-19T08:56:18.111Z lightningd(1): testing /usr/bin/lightning_gossipd
2018-03-19T08:56:18.121Z lightningd(1): testing /usr/bin/lightning_hsmd
2018-03-19T08:56:18.142Z lightningd(1): testing /usr/bin/lightning_onchaind
2018-03-19T08:56:18.194Z lightningd(1): testing /usr/bin/lightning_openingd
2018-03-19T08:56:19.209Z lightningd(1): Client: Received message 11 from client
2018-03-19T08:56:19.225Z lightningd(1): Client: Received message 9 from client
2018-03-19T08:56:19.226Z lightning_gossipd(12): pid 12, msgfd 13
2018-03-19T08:56:19.226Z lightningd(1): Loaded 0 HTLC signatures from DB
2018-03-19T08:56:19.227Z lightningd(1): Loaded 1 channels from DB
2018-03-19T08:56:19.227Z lightningd(1): Loading HTLCs for channel 1
2018-03-19T08:56:19.227Z lightningd(1): Restored 0 incoming and 0 outgoing HTLCS
2018-03-19T08:56:19.631Z lightningd(1): Waiting for bitcoind to warm up...
2018-03-19T08:56:34.726Z lightning_gossipd(12): req: type WIRE_GOSSIPCTL_INIT len 126
2018-03-19T08:56:34.727Z lightning_gossipd(12): Creating IPv6 listener on port 9735
2018-03-19T08:56:34.917Z lightningd(1): Adding block 513978: 0000000000000000004c3270bc11b00779e2a9ce2cdbc67a26a339abec01d06a
2018-03-19T08:56:34.923Z lightningd(1): Listening on 'lightning-rpc'
2018-03-19T08:56:34.923Z lightningd(1): Server started with public key 02d447157dd105474ac5c22b5866d42daa5a92439064cda022b3399505dfe46aa8, alias botbittrex.com (color #d2691e) and lightningd v0.5.2-2016-11-21-2290-g640ff4b
2018-03-19T08:56:34.925Z lightning_gossipd(12): req: type WIRE_GOSSIPCTL_PEER_ADDRHINT len 42
2018-03-19T08:56:34.925Z lightning_gossipd(12): req: type WIRE_GOSSIPCTL_REACH_PEER len 35
2018-03-19T08:56:34.953Z lightning_gossipd(12): Connected out for 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd
2018-03-19T08:56:34.978Z lightningd(1): Adding block 513979: 000000000000000000376267d342878f869cb68192ff5d73f5f1953ae83e3e1e
2018-03-19T08:56:34.985Z lightning_gossipd(12): peer_out WIRE_INIT
2018-03-19T08:56:34.988Z lightningd(1): Immediate feerate 1263 (was 0)
2018-03-19T08:56:34.988Z lightningd(1): Normal feerate 250 (was 0)
2018-03-19T08:56:34.988Z lightningd(1): Slow feerate 250 (was 0)
2018-03-19T08:56:35.019Z lightning_gossipd(12): peer_in WIRE_INIT
2018-03-19T08:56:35.019Z lightning_gossipd(12): UPDATE WIRE_GOSSIP_PEER_CONNECTED
2018-03-19T08:56:35.020Z lightning_gossipd(12): UPDATE WIRE_GOSSIP_PEER_CONNECTED
2018-03-19T08:56:35.020Z lightningd(1): 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: Peer has reconnected, state CHANNELD_AWAITING_LOCKIN
2018-03-19T08:56:35.020Z lightningd(1): Client: Received message 1 from client
2018-03-19T08:56:35.020Z lightningd(1): Client: Received message 9 from client
2018-03-19T08:56:35.021Z lightningd(1): lightning_channeld-024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: pid 39, msgfd 34
2018-03-19T08:56:35.025Z lightningd(1): 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: Waiting for funding confirmations
2018-03-19T08:56:35.082Z lightningd(1): Adding block 513980: 0000000000000000001a0822fbaef92ef048967fa32c68f96e3d57d13183ef2b
2018-03-19T08:56:35.445Z lightningd(1): Adding block 513981: 0000000000000000000478d3bb59c997624a254b1deb716b599137c447ec14b8
2018-03-19T08:56:35.941Z lightningd(1): lightning_channeld-024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: init LOCAL: remote_per_commit = 0348d5a730dd008
93ec7038f01ca8ef559e4f09f9d090ca2e114528a01c43c8c4a, old_remote_per_commit = 0348d5a730dd00893ec7038f01ca8ef559e4f09f9d090ca2e114528a01c43c8c4a next_idx_local = 1 next_idx_remote = 1
 revocations_received = 0 feerates 250/250 (range 250-6315)
2018-03-19T08:56:35.942Z lightningd(1): lightning_channeld-024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: peer_out WIRE_CHANNEL_REESTABLISH
2018-03-19T08:56:36.059Z lightningd(1): Adding block 513982: 00000000000000000017ed644e63726329929f078235aa75e9e674f06cded647
2018-03-19T08:56:36.412Z lightningd(1): Adding block 513983: 0000000000000000001bc8ba079a981d20ea748bf9a6021d0a8aec0d7490df96
2018-03-19T08:56:36.790Z lightningd(1): Adding block 513984: 0000000000000000001baf257c6d338c4cee810def408297608f0fd204a01283
2018-03-19T08:56:37.125Z lightningd(1): Adding block 513985: 0000000000000000002d3ef22e1c31f61487fea0e1f93d05f0683a257c6ab8fd
2018-03-19T08:56:37.374Z lightningd(1): Adding block 513986: 00000000000000000029153311ede8c1964bc92eefb196e1d0256680bf650a89
2018-03-19T08:56:38.961Z lightningd(1): Adding block 513987: 000000000000000000257f168634e40919aed17e65d73ea9624f217a7fcca4c7
2018-03-19T08:56:39.246Z lightningd(1): Adding block 513988: 0000000000000000000fbb9d14b56d1c102a3f89d1efc0b97e413523c724766d
2018-03-19T08:56:39.418Z lightningd(1): Adding block 513989: 00000000000000000053fb70ff89bc4dcdd6a45ab8c4241090ee99066a01e58e
2018-03-19T08:56:39.647Z lightningd(1): Adding block 513990: 00000000000000000027a41ac2d767fbf11fe3ddf00261ff6539b12da290b998
2018-03-19T08:56:39.753Z lightningd(1): Adding block 513991: 000000000000000000379a7b3ae10e11580d61462b14e6d3ab2a6c21825517a3
2018-03-19T08:56:39.817Z lightningd(1): Adding block 513992: 0000000000000000000daf839d7c73b3211faa32a280ba019baa04afa1e14536
2018-03-19T08:56:39.860Z lightningd(1): lightning_channeld-024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: peer_in WIRE_ERROR
2018-03-19T08:56:39.860Z lightningd(1): 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd chan #1: Peer permanent failure in CHANNELD_AWAITING_LOCKIN: lightning_chan
neld: received ERROR channel c8af5f57567b86b2b74834ec28e2b900bff87509be811ff5742d675a0da8f283: lightning_channeld: received ERROR channel ALL: Internal error: Failing due to dev-fail
 command
2018-03-19T08:56:39.860Z lightningd(1):          (tx 527e905af0b0c24c710d5144e2cdd2b56c008cbc35ed5b4efc19c285df2708fa)
2018-03-19T08:56:39.861Z lightningd(1): sendrawtransaction: 02000000000101c8af5f57567b86b2b74834ec28e2b900bff87509be811ff5742d675a0da8f28201000000000547c18001eb85010000000000220020a4
59c74acc19400f76f190e224e6f253998af6a1d56b1834f931ec5405773d180400483045022100c542b8761a15cec179f7fa2ef380906f9f99cd0f63b3669cf1011b21e16f2a3d02205378d0a5f9440fcbb18a3c4b7cb934c56d7f
7cc7252023057deb42420955a7a1014730440220752de536c582ef14886802f1eef89a47828d0912dc1fd5830ba2294543f3fe230220596c1b5a075d5061c87d553fb59c30b226d569a04aad71e73c6d91c950198f350147522103
28e562d39ce141d815985f7f75199fc068815bb71ed8f458c22af9ea75475f3421039cb6b9456f1b8d36dd672862baad9c2647d25ab9594398d902faf55cbabda29f52ae640e0620
2018-03-19T08:56:39.862Z lightning_gossipd(12): req: type WIRE_GOSSIPCTL_HAND_BACK_PEER len 189
2018-03-19T08:56:39.862Z lightning_gossipd(12): hand_back_peer 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd: now local again
2018-03-19T08:56:39.862Z lightning_gossipd(12): Received channel_announcement for channel 504149:1124:0
2018-03-19T08:56:39.865Z lightning_gossipd(12): UPDATE WIRE_GOSSIP_GET_TXOUT
2018-03-19T08:56:39.865Z lightning_gossipd(12): peer_in WIRE_ERROR
2018-03-19T08:56:39.865Z lightning_gossipd(12): 024bd94f0425590434538fd21d4e58982f7e9cfd8f339205a73deb9c0e0341f5bd sent ERROR channel c8af5f57567b86b2b74834ec28e2b900bff87509be811ff5
742d675a0da8f283: lightning_channeld: received ERROR channel ALL: Internal error: Failing due to dev-fail command
2018-03-19T08:56:39.939Z lightningd(1): sendrawtx exit 25, gave error code: -25?error message:?Missing inputs?
2018-03-19T08:56:40.317Z lightningd(1): Adding block 513993: 00000000000000000051b34c9733219768cba744dab635864f2847670412326c
2018-03-19T08:56:40.610Z lightning_gossipd(12): req: type WIRE_GOSSIP_GET_TXOUT_REPLY len 54
2018-03-19T08:56:40.802Z lightningd(1): Adding block 513994: 0000000000000000003fc5c5600c720452a1e3b25d064bc5d96b22e4e8a59992
2018-03-19T08:56:41.084Z lightningd(1): Adding block 513995: 00000000000000000006a280456d7f52bc6e402127097c9966373d1c85a6bfe8
2018-03-19T08:56:41.338Z lightningd(1): Adding block 513996: 0000000000000000003cb3e6f32a0c470522d74c70ae5268839745527aeade52
2018-03-19T08:56:41.519Z lightningd(1): Adding block 513997: 00000000000000000050a83ee84efb0e196bfb9fe09dfd8a4005ef732a3fc0af
2018-03-19T08:56:41.692Z lightningd(1): Adding block 513998: 0000000000000000003b76912dab1b719950fb0da2be96ca9ae13aa33b7f71a6
2018-03-19T08:56:41.951Z lightningd(1): Adding block 513999: 0000000000000000000efc557a92c70811da6fa7ec4db56f42d57b97f9d9325d
2018-03-19T08:56:42.282Z lightningd(1): Adding block 514000: 0000000000000000003fc2067c2f17f0712caa4eedda294e5290e2fff278ed88
2018-03-19T08:56:42.341Z lightningd(1): Adding block 514001: 000000000000000000040c421005d950a4e667af5dcfee739e5acbdf34d88414
2018-03-19T08:56:42.397Z lightningd(1): Adding block 514002: 0000000000000000002adf6c3ec9dc179b8ec2f70f5f11fd7985c1fde38af919
2018-03-19T08:56:42.702Z lightningd(1): Adding block 514003: 00000000000000000017c12ec86497f1cc8d41c07b55f5023316eff798c1cba3
2018-03-19T08:56:42.952Z lightningd(1): Adding block 514004: 00000000000000000051f108c957ecab11cae2ba5d7347a81f272f103a9be4be
2018-03-19T08:56:43.043Z lightningd(1): Adding block 514005: 00000000000000000044ed0c59dc034ded314472c1ee9ee2ce29824af72ad790
2018-03-19T08:56:43.110Z lightningd(1): Adding block 514006: 00000000000000000043fd6f70d8776c601c0bc9e6586234b3969843889c40d7
2018-03-19T08:56:43.213Z lightningd(1): Adding block 514007: 0000000000000000001aee0fb6c6213b132e95519fd0b8a327aa84181c3f0dcf
2018-03-19T08:56:43.279Z lightningd(1): Adding block 514008: 000000000000000000482a0a225ab915d700a277f453941d38880674a0eacdbe
2018-03-19T08:56:43.368Z lightningd(1): Adding block 514009: 00000000000000000038c38c9802836e9fab1e9f38df5ddbf6ba1101db616e9d
2018-03-19T08:56:43.504Z lightningd(1): Adding block 514010: 000000000000000000018529ad32dff054460b18179d5658d761972fae3938d2
2018-03-19T08:56:43.717Z lightningd(1): Adding block 514011: 000000000000000000142c69a138eaafce0aad9e7864b730b58f5222ce435666
2018-03-19T08:56:44.073Z lightningd(1): Adding block 514012: 0000000000000000002697b773c77e98bb557d9b16c0eebe638f089f66cff346
2018-03-19T08:56:44.259Z lightningd(1): Adding block 514013: 000000000000000000219b686edc67bb78096674f7203f125ab1958b0ca1d15d

### `getinfo` output

{ "id" : "02d447157dd105474ac5c22b5866d42daa5a92439064cda022b3399505dfe46aa8", "port" : 9735, "address" : 
        [ 
                { "type" : "ipv4", "address" : "35.192.6.142", "port" : 9735 } ], "version" : "v0.5.2-2016-11-21-22
90-g640ff4b", "blockheight" : 514222, "network" : "bitcoin" }

### LINKED ISSUE #1283: Use pruneheight instead of hammering getblock
When starting a new lightning node that's connected to a pruned bitcoind node, I get a flood of log messages like:

```
2018-03-27T10:29:59.549Z lightningd(12118): bitcoin-cli getblock 000000000000000000075750b9d433983830f4560044e4977b10b100668e6509? exited with status 1
```

This seems to go on for quite a while. I can't tell if it's going in circles or fetching blocks in random order.

It would seem more efficient to call `getblockchaininfo`, check if `pruned` is `true` and then use `pruneheight` to know which blocks not to ask for.

### LINKED ISSUE #1331: CHANNELD_AWAITING_LOCKIN:We've confirmed funding, they haven't yet.
i've opened a channel but after 20+ blocks im still getting CHANNELD_AWAITING_LOCKIN
i am connected and synced.

is there anything i can do? or it's a problem on the other side?
### `getinfo` output

{ "id" : "03a8f92735a6e7aac231ec46d51f0b9e84aeed6ce8abd76431849b4e76e9745bfb", "port" : 9735, "address" :
        [
                { "type" : "ipv4", "address" : "88.99.164.229", "port" : 9735 },
                { "type" : "ipv6", "address" : "2a01:4f8:10a:3524::2", "port" : 9735 } ], "version" : "v0.5.2-2016-11-21-2473-g1a55147", "blockheight" : 516810, "network" : "bitcoin" }

### `listpeers` output
```
{ "peers" :
        [
                { "id" : "02f6725f9c1c40333b67faea92fd211c183050f28df32cac3f9d69685fe9665432", "connected" : true, "netaddr" :
                        [ "104.198.32.198:9735" ], "alias" : "SLEEPYARK-6-11-21-2360-g044705a5", "color" : "02f672", "channels" :
                        [
                                { "state" : "CHANNELD_AWAITING_LOCKIN", "owner" : "lightning_channeld", "short_channel_id" : "516778:1569:1", "channel_id" : "e52f963dec9b31aa28ca3c45337b290a5984ec8825e7ba1f736b5b1fd0322f11", "funding_txid" : "102f32d01f5b6b731fbae72588ec84590a297b33453cca28aa319bec3d962fe5", "msatoshi_to_us" : 4999000000, "msatoshi_total" : 4999000000, "dust_limit_satoshis" : 546, "max_htlc_value_in_flight_msat" : 18446744073709551615, "channel_reserve_satoshis" : 0, "htlc_minimum_msat" : 0, "their_to_self_delay" : 144, "our_to_self_delay" : 144, "to_self_delay" : 144, "max_accepted_htlcs" : 483, "status" :
                                        [ "CHANNELD_AWAITING_LOCKIN:We've confirmed funding, they haven't yet." ], "in_payments_offered" : 0, "in_msatoshi_offered" : 0, "in_payments_fulfilled" : 0, "in_msatoshi_fulfilled" : 0, "out_payments_offered" : 0, "out_msatoshi_offered" : 0, "out_payments_fulfilled" : 0, "out_msatoshi_fulfilled" : 0 } ] },
                { "state" : "GOSSIPING", "id" : "0327049d8d63f0c40193cdf3afc61817c8647808a4e482de0716fcef74e6d92ebf", "alias" : "FederalReserve", "color" : "42f45d", "netaddr" :
                        [ "35.184.46.254:9735" ], "connected" : true, "owner" : "lightning_gossipd" },
                { "state" : "GOSSIPING", "id" : "024a2e265cd66066b78a788ae615acdc84b5b0dec9efac36d7ac87513015eaf6ed", "alias" : "Bitrefill", "color" : "3399ff", "netaddr" :
                        [ "52.16.240.222:9735" ], "connected" : true, "owner" : "lightning_gossipd" } ] }
```

### LINKED ISSUE #1358: 'lightning-rpc': Connection refused
This started happening somewhere in the past few days on Ubuntu and is still happening with the latest master.

```sh
cli/lightning-cli help
lightning-cli: Connecting to 'lightning-rpc': Connection refused
```

The log is happily spitting out `lightningd` and `lightning_gossipd` messages, but I can't use RPC.

```sh
ps aux | grep lightningd
bitcoin  21805  0.0  0.1  26184  2980 ?        Ss   19:28   0:00 SCREEN -S lightningd
bitcoin  22008  0.2  0.5  26128 12220 pts/3    S+   19:28   0:00 lightningd/lightningd --network=bitcoin --ipaddr=... --log-level=debug --log-file=/home/bitcoin/.lightning/debug.log
bitcoin  22015  0.0  0.1  11112  2776 pts/3    S+   19:28   0:00 /home/bitcoin/lightning/lightningd/lightning_hsmd
bitcoin  22016  4.6  0.9  28264 20428 pts/3    S+   19:28   0:11 /home/bitcoin/lightning/lightningd/lightning_gossipd
````

I think it's still catching up so that might be related, but that didn't use to matter. I also noticed that when running with `-daemon` it doesn't daemonize, at least not in the first minute that I waited.

### PR DESCRIPTION
**Pull request by cdecker**
With these changes `onchaind` no longer relies on rescanning the blockchain from the wallet's birth height on each startup. Instead we remember the transactions that triggered some behavior in `onchaind` and replay them before syncing with the blockchain.

This is the continuation of #1364 which took care of storing transactions we are interested in and tracking their confirmations when syncing. The `onchaind` persistence is a small layer on top of those transactions remembering which transaction triggered what notification to `onchaind`.

In addition the sync offset calculation was simplified, and we can now use the `--rescan` command line option to specify either a relative or absolute height to continue syncing from. The default is 30 blocks for testnet and 15 for mainnet, to counter potential forks.

This may also partially address pruning nodes, as long as the sync horizon doesn't get pruned out from under our feet.

Fixes #1168 
Fixes #1179 
Fixes #1283 
Fixes #1331 
Fixes #1358
Fixes #1246 
Fixes #1243 

And maybe some more?

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -309,6 +309,10 @@ static void config_register_opts(struct lightningd *ld)
 	opt_register_arg("--fee-base", opt_set_u32, opt_show_u32,
 			 &ld->config.fee_base,
 			 "Millisatoshi minimum to charge for HTLC");
+	opt_register_arg("--rescan", opt_set_s32, opt_show_s32,
+			 &ld->config.rescan,
+			 "Number of blocks to rescan from the current head, or "
+			 "absolute blockheight if negative");
 	opt_register_arg("--fee-per-satoshi", opt_set_s32, opt_show_s32,
 			 &ld->config.fee_per_satoshi,
 			 "Microsatoshi fee for every satoshi in HTLC");
@@ -420,6 +424,9 @@ static const struct config testnet_config = {
 
 	/* Testnet sucks */
 	.ignore_fee_limits = true,
+
+	/* Rescan 5 hours of blocks on testnet, it's reorg happy */
+	.rescan = 30,
 };
 
 /* aka. "Dude, where's my coins?" */
@@ -481,6 +488,9 @@ static const struct config mainnet_config = {
 
 	/* Mainnet should have more stable fees */
 	.ignore_fee_limits = false,
+
+	/* Rescan 2.5 hours of blocks on startup, it's not so reorg happy */
+	.rescan = 15,
 };
 
 static void check_config(struct lightningd *ld)
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -4341,6 +4341,40 @@ class LightningDTests(BaseLightningDTests):
         self.assertRaisesRegex(ValueError, "Peer is not in gossip mode",
                                l1.rpc.disconnect, l3.info['id'])
 
+    def test_rescan(self):
+        """Test the rescan option
+        """
+        l1 = self.node_factory.get_node()
+        btc = l1.bitcoin
+
+        # The first start should start at current_height - 30 = 71, make sure
+        # it's not earlier
+        l1.daemon.wait_for_log(r'Adding block 101')
+        assert not l1.daemon.is_in_log(r'Adding block 70')
+
+        # Restarting with a higher rescan should go back further
+        l1.daemon.opts['rescan'] = 50
+        l1.restart()
+        l1.daemon.wait_for_log(r'Adding block 101')
+        assert l1.daemon.is_in_log(r'Adding block 51')
+        assert not l1.daemon.is_in_log(r'Adding block 50')
+
+        # Restarting with an absolute rescan should start from there
+        l1.daemon.opts['rescan'] = -31
+        l1.restart()
+        l1.daemon.wait_for_log(r'Adding block 101')
+        assert l1.daemon.is_in_log(r'Adding block 31')
+        assert not l1.daemon.is_in_log(r'Adding block 30')
+
+        # Restarting with a future absolute blockheight should just start with
+        # the current height
+        l1.daemon.opts['rescan'] = -500000
+        l1.stop()
+        btc.rpc.generate(4)
+        l1.daemon.start()
+        l1.daemon.wait_for_log(r'Adding block 105')
+        assert not l1.daemon.is_in_log(r'Adding block 102')
+
 
 if __name__ == '__main__':
     unittest.main(verbosity=2)
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -1448,12 +1448,14 @@ class LightningDTests(BaseLightningDTests):
 
         # 10 later, l1 should collect its to-self payment.
         bitcoind.generate_block(10)
-        l1.daemon.wait_for_log('Broadcasting OUR_DELAYED_RETURN_TO_WALLET .* to resolve OUR_UNILATERAL/DELAYED_OUTPUT_TO_US')
+        l1.daemon.wait_for_log('Broadcasting OUR_DELAYED_RETURN_TO_WALLET .* to resolve '
+                               'OUR_UNILATERAL/DELAYED_OUTPUT_TO_US')
         l1.daemon.wait_for_log('sendrawtx exit 0')
 
         # First time it sees it, onchaind cares.
         bitcoind.generate_block(1)
-        l1.daemon.wait_for_log('Resolved OUR_UNILATERAL/DELAYED_OUTPUT_TO_US by our proposal OUR_DELAYED_RETURN_TO_WALLET')
+        l1.daemon.wait_for_log('Resolved OUR_UNILATERAL/DELAYED_OUTPUT_TO_US by our proposal '
+                               'OUR_DELAYED_RETURN_TO_WALLET')
 
         # Now test unrelated onchain churn.
         # Daemon gets told about wallet; says it doesn't care.
@@ -1477,6 +1479,57 @@ class LightningDTests(BaseLightningDTests):
 
         # Note: for this test we leave onchaind running, so we can detect
         # any leaks!
+
+    @unittest.skipIf(not DEVELOPER, "needs DEVELOPER=1")
+    def test_onchaind_replay(self):
+        disconnects = ['+WIRE_REVOKE_AND_ACK', 'permfail']
+        options = {'locktime-blocks': 201, 'cltv-delta': 101}
+        l1 = self.node_factory.get_node(options=options, disconnect=disconnects)
+        l2 = self.node_factory.get_node(options=options)
+        btc = l1.bitcoin
+
+        l1.rpc.connect(l2.info['id'], 'localhost', l2.info['port'])
+        self.fund_channel(l1, l2, 10**6)
+
+        rhash = l2.rpc.invoice(10**8, 'onchaind_replay', 'desc')['payment_hash']
+        routestep = {
+            'msatoshi': 10**8 - 1,
+            'id': l2.info['id'],
+            'delay': 101,
+            'channel': '1:1:1'
+        }
+        l1.rpc.sendpay(to_json([routestep]), rhash)
+        l1.daemon.wait_for_log(r'Disabling channel')
+        btc.rpc.generate(1)
+
+        # Wait for nodes to notice the failure, this seach needle is after the
+        # DB commit so we're sure the tx entries in onchaindtxs have been added
+        l1.daemon.wait_for_log("Deleting channel .* due to the funding outpoint being spent")
+        l2.daemon.wait_for_log("Deleting channel .* due to the funding outpoint being spent")
+
+        # We should at least have the init tx now
+        assert len(l1.db_query("SELECT * FROM channeltxs;")) > 0
+        assert len(l2.db_query("SELECT * FROM channeltxs;")) > 0
+
+        # Generate some blocks so we restart the onchaind from DB (we rescan
+        # last_height - 100)
+        btc.rpc.generate(100)
+        sync_blockheight([l1, l2])
+
+        # l1 should still have a running onchaind
+        assert len(l1.db_query("SELECT * FROM channeltxs;")) > 0
+
+        l2.rpc.stop()
+        l1.restart()
+
+        # Can't wait for it, it's after the "Server started" wait in restart()
+        assert l1.daemon.is_in_log(r'Restarting onchaind for channel')
+
+        # l1 should still notice that the funding was spent and that we should react to it
+        l1.daemon.wait_for_log("Propose handling OUR_UNILATERAL/DELAYED_OUTPUT_TO_US by OUR_DELAYED_RETURN_TO_WALLET")
+        sync_blockheight([l1])
+        btc.rpc.generate(10)
+        sync_blockheight([l1])
 
     @unittest.skipIf(not DEVELOPER, "needs DEVELOPER=1")
     def test_onchain_dust_out(self):
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -3486,6 +3486,7 @@ class LightningDTests(BaseLightningDTests):
     def test_funding_while_offline(self):
         l1 = self.node_factory.get_node()
         addr = l1.rpc.newaddr()['address']
+        sync_blockheight([l1])
 
         # l1 goes down.
         l1.stop()
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -767,29 +767,6 @@ void wallet_channel_stats_load(struct wallet *w,
 	stats->out_msatoshi_fulfilled = sqlite3_column_int64(stmt, 7);
 }
 
-#ifdef COMPAT_V052
-/* Upgrade of db (or initial create): do we have anything to scan for? */
-static bool wallet_ever_used(struct wallet *w)
-{
-	sqlite3_stmt *stmt;
-	bool channel_utxos;
-
-	/* If we ever handed out an address. */
-	if (db_get_intvar(w->db, "bip32_max_index", 0) != 0)
-		return true;
-
-	/* Or if they do a unilateral close, the output to us provides a UTXO. */
-	stmt = db_query(__func__, w->db,
-			"SELECT COUNT(*) FROM outputs WHERE commitment_point IS NOT NULL;");
-	int ret = sqlite3_step(stmt);
-	assert(ret == SQLITE_ROW);
-	channel_utxos = (sqlite3_column_int(stmt, 0) != 0);
-	sqlite3_finalize(stmt);
-
-	return channel_utxos;
-}
-#endif
-
 /* We want the earlier of either:
  * 1. The first channel we're still watching (it might have closed),
  * 2. The last block we scanned for UTXO (might have new incoming payments)
@@ -799,43 +776,16 @@ static bool wallet_ever_used(struct wallet *w)
  */
 u32 wallet_first_blocknum(struct wallet *w, u32 first_possible)
 {
-	int err;
-	u32 first_channel, first_utxo;
-	sqlite3_stmt *stmt =
-	    db_query(__func__, w->db,
-		     "SELECT MIN(first_blocknum) FROM channels;");
+	u32 blockheight;
+	sqlite3_stmt *stmt = db_prepare(w->db, "SELECT MAX(height) FROM blocks;");
 
-	/* If we ever opened a channel, this will give us the first block. */
-	err = sqlite3_step(stmt);
-	if (err == SQLITE_ROW && sqlite3_column_type(stmt, 0) != SQLITE_NULL)
-		first_channel = sqlite3_column_int(stmt, 0);
-	else
-		first_channel = UINT32_MAX;
-	sqlite3_finalize(stmt);
-
-#ifdef COMPAT_V052
-	/* This field was missing in older databases. */
-	first_utxo = db_get_intvar(w->db, "last_processed_block", 0);
-	if (first_utxo == 0) {
-		/* Don't know?  New db, or upgraded. */
-		if (wallet_ever_used(w))
-			/* Be conservative */
-			first_utxo = first_possible;
-		else
-			first_utxo = UINT32_MAX;
-	}
-#else
-	first_utxo = db_get_intvar(w->db, "last_processed_block", UINT32_MAX);
-#endif
-
-	/* Never go below the start of the Lightning Network */
-	if (first_utxo < first_possible)
-		first_utxo = first_possible;
-
-	if (first_utxo < first_channel)
-		return first_utxo;
-	else
-		return first_channel;
+	/* If we ever processed a block we'll get the latest block in the chain */
+	if (sqlite3_step(stmt) == SQLITE_ROW && sqlite3_column_type(stmt, 0) != SQLITE_NULL) {
+		blockheight = sqlite3_column_int(stmt, 0);
+		sqlite3_finalize(stmt);
+		return blockheight;
+	} else
+		return first_possible;
 }
 
 static void wallet_channel_config_insert(struct wallet *w,
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -767,14 +767,7 @@ void wallet_channel_stats_load(struct wallet *w,
 	stats->out_msatoshi_fulfilled = sqlite3_column_int64(stmt, 7);
 }
 
-/* We want the earlier of either:
- * 1. The first channel we're still watching (it might have closed),
- * 2. The last block we scanned for UTXO (might have new incoming payments)
- *
- * chaintopology actually subtracts another 100 blocks to make sure we
- * catch chain forks.
- */
-u32 wallet_first_blocknum(struct wallet *w, u32 first_possible)
+u32 wallet_blocks_height(struct wallet *w, u32 def)
 {
 	u32 blockheight;
 	sqlite3_stmt *stmt = db_prepare(w->db, "SELECT MAX(height) FROM blocks;");
@@ -785,7 +778,7 @@ u32 wallet_first_blocknum(struct wallet *w, u32 first_possible)
 		sqlite3_finalize(stmt);
 		return blockheight;
 	} else
-		return first_possible;
+		return def;
 }
 
 static void wallet_channel_config_insert(struct wallet *w,
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -11,6 +11,7 @@
 #include <lightningd/log.h>
 #include <lightningd/peer_control.h>
 #include <lightningd/peer_htlcs.h>
+#include <onchaind/gen_onchain_wire.h>
 #include <string.h>
 
 #define SQLITE_MAX_UINT 0x7FFFFFFFFFFFFFFF
@@ -2277,3 +2278,77 @@ struct bitcoin_txid *wallet_transactions_by_height(const tal_t *ctx,
 	return txids;
 }
 
+void wallet_channeltxs_add(struct wallet *w, struct channel *chan,
+			   const int type, const struct bitcoin_txid *txid,
+			   const u32 input_num, const u32 blockheight)
+{
+	sqlite3_stmt *stmt;
+	stmt = db_prepare(w->db, "INSERT INTO channeltxs ("
+			  "  channel_id"
+			  ", type"
+			  ", transaction_id"
+			  ", input_num"
+			  ", blockheight"
+			  ") VALUES (?, ?, ?, ?, ?);");
+	sqlite3_bind_int(stmt, 1, chan->dbid);
+	sqlite3_bind_int(stmt, 2, type);
+	sqlite3_bind_sha256(stmt, 3, &txid->shad.sha);
+	sqlite3_bind_int(stmt, 4, input_num);
+	sqlite3_bind_int(stmt, 5, blockheight);
+
+	db_exec_prepared(w->db, stmt);
+}
+
+u32 *wallet_onchaind_channels(struct wallet *w,
+			      const tal_t *ctx)
+{
+	sqlite3_stmt *stmt;
+	size_t count = 0;
+	u32 *channel_ids = tal_arr(ctx, u32, 0);
+	stmt = db_prepare(w->db, "SELECT DISTINCT(channel_id) FROM channeltxs WHERE type = ?;");
+	sqlite3_bind_int(stmt, 1, WIRE_ONCHAIN_INIT);
+
+	while (sqlite3_step(stmt) == SQLITE_ROW) {
+		count++;
+		tal_resize(&channel_ids, count);
+			channel_ids[count-1] = sqlite3_column_int64(stmt, 0);
+	}
+
+	return channel_ids;
+}
+
+struct channeltx *wallet_channeltxs_get(struct wallet *w, const tal_t *ctx,
+					u32 channel_id)
+{
+	sqlite3_stmt *stmt;
+	size_t count = 0;
+	struct channeltx *res = tal_arr(ctx, struct channeltx, 0);
+	stmt = db_prepare(w->db,
+			  "SELECT"
+			  "  c.type"
+			  ", c.blockheight"
+			  ", t.rawtx"
+			  ", c.input_num"
+			  ", c.blockheight - t.blockheight + 1 AS depth"
+			  ", t.id as txid "
+			  "FROM channeltxs c "
+			  "JOIN transactions t ON t.id == c.transaction_id "
+			  "WHERE channel_id = ? "
+			  "ORDER BY c.id ASC;");
+	sqlite3_bind_int(stmt, 1, channel_id);
+
+	while (sqlite3_step(stmt) == SQLITE_ROW) {
+		count++;
+		tal_resize(&res, count);
+
+		res[count-1].channel_id = channel_id;
+		res[count-1].type = sqlite3_column_int(stmt, 0);
+		res[count-1].blockheight = sqlite3_column_int(stmt, 1);
+		res[count-1].tx = sqlite3_column_tx(ctx, stmt, 2);
+		res[count-1].input_num = sqlite3_column_int(stmt, 3);
+		res[count-1].depth = sqlite3_column_int(stmt, 4);
+		sqlite3_column_sha256(stmt, 5, &res[count-1].txid.shad.sha);
+	}
+
+	return res;
+}
--- a/wallet/wallet.h
+++ b/wallet/wallet.h
@@ -128,7 +128,7 @@ struct channel_stats {
 	u64 out_msatoshi_offered, out_msatoshi_fulfilled;
 };
 
-struct onchaindtx {
+struct channeltx {
 	u32 channel_id;
 	int type;
 	u32 blockheight;
@@ -845,5 +845,24 @@ struct txlocator *wallet_transaction_locate(const tal_t *ctx, struct wallet *w,
 struct bitcoin_txid *wallet_transactions_by_height(const tal_t *ctx,
 						   struct wallet *w,
 						   const u32 blockheight);
+
+/**
+ * Store transactions of interest in the database to replay on restart
+ */
+void wallet_channeltxs_add(struct wallet *w, struct channel *chan,
+			    const int type, const struct bitcoin_txid *txid,
+			   const u32 input_num, const u32 blockheight);
+
+/**
+ * List channels for which we had an onchaind running
+ */
+u32 *wallet_onchaind_channels(struct wallet *w,
+			      const tal_t *ctx);
+
+/**
+ * Get transactions that we'd like to replay for a channel.
+ */
+struct channeltx *wallet_channeltxs_get(struct wallet *w, const tal_t *ctx,
+					u32 channel_id);
 
 #endif /* LIGHTNING_WALLET_WALLET_H */
--- a/wallet/wallet.h
+++ b/wallet/wallet.h
@@ -314,12 +314,15 @@ void wallet_channel_stats_incr_out_fulfilled(struct wallet *w, u64 cdbid, u64 ms
 void wallet_channel_stats_load(struct wallet *w, u64 cdbid, struct channel_stats *stats);
 
 /**
- * wallet_first_blocknum - get first block we're interested in.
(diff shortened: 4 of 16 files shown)

### CODE REVIEW

**Comment by ZmnSCPxj** [comment]
Minor nits.

Ack ab7280d

File: wallet/wallet.c [hunk 1]
```diff
@@ -2277,3 +2278,77 @@ struct bitcoin_txid *wallet_transactions_by_height(const tal_t *ctx,
...
+			   const int type, const struct bitcoin_txid *txid,
+			   const u32 input_num, const u32 blockheight)
+{
+	sqlite3_stmt *stmt;
+	stmt = db_prepare(w->db, "INSERT INTO channeltxs ("
+			  " channel_id,"
+			  " type,"
+			  " transaction_id,"
+			  " input_num"
+			  ", blockheight"
```
ZmnSCPxj: Inconsistent comma usage

wallet/wallet.h
```diff
@@ -128,7 +128,7 @@ struct channel_stats {
 	u64 out_msatoshi_offered, out_msatoshi_fulfilled;
 };
 
-struct onchaindtx {
+struct channeltx {
```
ZmnSCPxj: Maybe this particular change can be put in the previous commit instead?

tests/test_lightningd.py
```diff
@@ -312,10 +312,10 @@ def fund_channel(self, l1, l2, amount):
         decoded2 = bitcoind.rpc.decoderawtransaction(tx)
         raise ValueError("Can't find {} payment in {} (1={} 2={})".format(amount, tx, decoded, decoded2))
 
-    def line_graph(self, n=2):
+    def line_graph(self, n=2, options=None):
```
ZmnSCPxj: An improvement certainly, but not used in the same commit that contains it. In principle, better to separate in a different commit.  Fine by me either way though.

lightningd/options.c
```diff
@@ -481,6 +488,9 @@ static const struct config mainnet_config = {
 
 	/* Mainnet should have more stable fees */
 	.ignore_fee_limits = false,
+
+	/* Do not rescan */
+	.rescan = 15,
```
ZmnSCPxj: This will rescan 15 blocks before the current block height? The comment seems not matching? (should it not be 0 if we just continue scanning at where we left off?)

Also, I remember we had an issue before that our block height would keep getting pushed back if we kept getting restarted, would that behavior occur again?

lightningd/options.c
```diff
@@ -309,6 +309,10 @@ static void config_register_opts(struct lightningd *ld)
 	opt_register_arg("--fee-base", opt_set_u32, opt_show_u32,
 			 &ld->config.fee_base,
 			 "Millisatoshi minimum to charge for HTLC");
+	opt_register_arg("--rescan", opt_set_s32, opt_show_s32,
+			 &ld->config.rescan,
+			 "Number of blocks to rescan from the current head, or "
+			 "absolute blockheight if negative");
```
ZmnSCPxj: Seems more sensible, to use negative to mean "current blockheight minus absolute value" and to use positive for "absolute blockheight".  And to use 0 to mean "current blockheight, no rescan".  Similar to Python arrays/strings, with the "end" being the current known blockheight and a negative index being relative from the end.

But this comment is not a blocking one and I think the current standard is reasonable.

lightningd/options.c
cdecker: Good catch, I had 0 in there initially and then I remembered that we don't handle reorgs below our in-memory chain all too well, so I made it 15 blocks. If we process less than 15 blocks before crashing, then yes, this could end up walking backwards, we could add the lower bound at the segwit activation back in, but that's also painful.

lightningd/options.c
cdecker: I preferred making the more complex case, i.e., choosing a fixed offset, the case with the added sign. Everybody should just use relative offsets from their current head, in particular if they run from a startup script, and only use the absolute offset (with negation) in a recovery scenario.

lightningd/options.c
cdecker: An alternative could be to use `last_processed_block` and not decrement that if it is above our current blockheight. That would make sure that our offset if computed from a fixed point, rather than risking sliding back in time.

tests/test_lightningd.py
cdecker: Removed changes, I initially used `line_graph` in the onchaind test but then needed to instruct one to fail, so I used the primitives instead.

cdecker:
Since this may break a thing or too, I'd like to wait for @rustyrussell to sign off as well, just as a sanity check :-)

wallet/wallet.c
ZmnSCPxj: Another wrinkle of comma placement, is that one notices that SQL consistently, uses longer command keywords, than the keywords for clauses: `SELECT` vs `WHERE` vs `ORDER` vs `FROM`; `INSERT` vs `VALUES`; `UPDATE` vs `WHERE` vs `SET`, etc.

    SELECT col1
         , col2
      FROM table
     WHERE col3 = 'whatever'
       AND col4 = 42
         ;

This has the advantage of never requiring spaces before the end of strings being concatenated, since everything after the initial command keyword gets an extra space before it:

    "SELECT col1"
    "     , col2"
    "  FROM table"
    " WHERE col3 = 'whatever'"
    "   AND col4 = 42"
    "     ;"

cdecker: That's a bit much imho, with the spacing that I added we actually get the same spacing as if we'd written it in the shell (`.schema` will show excessive spaces in the `CREATE` statement for example, which is hard to read).

robtex:
i want this nooow! 
thanks @cdecker, good work !

lightningd/options.c
ZmnSCPxj: Comment is still `/* Do not rescan */` even though default is to rescan by 15 blocks back?

wallet/wallet.h
```diff
@@ -128,6 +128,16 @@ struct channel_stats {
 	u64 out_msatoshi_offered, out_msatoshi_fulfilled;
 };
 
+struct onchaindtx {
+	u32 channel_id;
+	int type;
```
rustyrussell: `int type` always makes me wince.  I prefer an enum, if only for readability.

wallet/wallet.h
```diff
@@ -128,6 +128,16 @@ struct channel_stats {
...
+	u32 blockheight;
+	struct bitcoin_txid txid;
+	struct bitcoin_tx *tx;
+	u32 input_num;
+	u32 depth;
```
rustyrussell: depth and blockheight seem implicitly related; it seems odd to have this here.  Perhaps I'll revise that comment when reviewing a followon patch?

wallet/wallet.c
```diff
@@ -2277,3 +2278,77 @@ struct bitcoin_txid *wallet_transactions_by_height(const tal_t *ctx,
...
+{
+	sqlite3_stmt *stmt;
+	stmt = db_prepare(w->db, "INSERT INTO channeltxs ("
+			  "  channel_id"
+			  ", type"
+			  ", transaction_id"
+			  ", input_num"
+			  ", blockheight"
+			  ") VALUES (?, ?, ?, ?, ?);");
+	sqlite3_bind_int(stmt, 1, chan->dbid);
```
rustyrussell: Is the 'id' database field used at all?

wallet/wallet.c
```diff
@@ -799,43 +776,16 @@ static bool wallet_ever_used(struct wallet *w)
...
-	else
-		return first_channel;
+	/* If we ever processed a block we'll get the latest block in the chain */
+	if (sqlite3_step(stmt) == SQLITE_ROW && sqlite3_column_type(stmt, 0) != SQLITE_NULL) {
+		blockheight = sqlite3_column_int(stmt, 0);
+		sqlite3_finalize(stmt);
+		return blockheight;
+	} else
+		return first_possible;
 }
```
rustyrussell: If they are currently onchain, and upgrade, this will break them?  I think this logic needs to stay for a while :(

lightningd/options.c
rustyrussell: Yeah, nobody should be using this anyway, so I think -ve for relative is pretty common.  Otherwise you want --rescan-absolute or something...

ZmnSCPxj:
ACK ddd53f1f76a0bfad56ead695ab776d6fe6ace0b2

wallet/wallet.c
cdecker: It's an autoincrement field that is used in the `SELECT` below to make sure we replay the events in the same order they were created, i.e., seen originally. This is just for added consistency should we have some dependency on the order (multiple transactions being confirmed in the same block).

wallet/wallet.h
cdecker: They are different: the `blockheight` is the height at which the TX was confirmed, the `depth` is the difference between current blockheight and confirmation blockheight at the time the original notification was delivered (notice that this is not the current blockheight when we are replaying the events).

cdecker:
Thanks for the review @rustyrussell. Regarding backward compatibility we could insert a dummy block at a given height as a migration and start the scan from there (when lightning first became interesting seems like a good place to start).

cdecker:
@rustyrussell I added 3 migrations that'll insert a dummy block with `height=first_blocknum` (which then gets dropped upon rescanning immediately) if we have a channel, and do nothing otherwise :wink:

wallet/wallet.c
ZmnSCPxj: It is declared `INTEGER PRIMARY KEY`, but note, this implies it is equivalent to `ROWID` (an internal detail) under SQLITE.  In particular, if a row is deleted, it is possible for SQLITE to reuse that row when another row is inserted. We need an explicit `AUTOINCREMENT` in the declaration of `id` in `CREATE TABLE channeltxs`.

https://www.sqlite.org/autoinc.html

> If the largest ROWID is equal to the largest possible integer (9223372036854775807) then the database engine starts picking positive candidate ROWIDs at random until it finds one that is not previously used. If no unused ROWID can be found after a reasonable number of attempts, the insert operation fails with an SQLITE_FULL error.

Admittedly the condition of reaching 2^63 is extremely unlikely, and if we neglect that, we can generally think of `INTEGER PRIMARY KEY` as implicitly `AUTOINCREMENT`, but I think explicitness is a win here.

ZmnSCPxj:
ACK fc2b6c0fe3a83806bc71600e703d66968d1d60dd

Sjors:
I have one c-lightning node that's normally connected to a pruned node, though I temporarily connected it to an unpruned node. I updated to the latest master and it's currently catching up again using the full node. Once that's done, I'll switch it back to the pruned node and see what happens.

cdecker:
That'd be great @Sjors, though this is not really a permanent solution for pruned nodes: if the c-lightning node is not running for a long time and it's sync horizon drops below the pruning limit for `bitcoind` it'll be unable to fetch the pruned nodes. We might want to eventually attach to the P2P network directly and fetch blocks from there.

Issue #665, opened 2018-01-19
Feature Request: fund channel with all available funds, or a specific output


Sjors:
Avoids the need for a change address, especially if you want to put all available funds in a single channel.

gdistasi:
I agree, this is a very useful feature.

windsok:
Agree. I accidentally sent ~8000 satoshis to a change address when doing my first channel opening. My own fault, but an easier way to force spending all available funds and not producing a change UTXO would be useful.

Sjors:
With fees ticking up again, it's now spending ~€1.50 to open a channel, this becomes more relevant.

cdecker:
Marking as good first issue, shouldn't be too hard and it's a good exercise in getting acquainted with the source.

ZmnSCPxj:
Suggestion for those who want to consider taking this issue: `withdraw` command supports `all` input to withdraw all funds; it may be useful to unify the code with `withdraw`, then support selecting an output by "txid:outnum" form in the unified code (to fund using a specific output).  This implies adding a wrapper around `wallet_select_coins`/`wallet_select_all`.

wythe:
I added support to fundchannel to accept "all". For now, I just made `json_fund_channel` look as close to a copy of `json_withdraw` as possible.  Now, as @ZmnSCPxj  suggests, I will unify the code and also support selecting a specific output.

It worked on testnet for me, and I added a `test_funding_all` python test.  

Its all on my `fund_all` branch.  Should I make a pull request and mark as work-in-progress?

ZmnSCPxj:
Yes, please make pull request, thank you!  It could be marked WIP if you intend to also add the specific-output support in the same pull request but we could also do a separate pull request for the specific-output support.

wythe:
Ok sounds like a plan. Will do.

wythe:
I hear you @robtex.  Maybe another option, closest fit:
`fundchannel fit 10000 16000`

I also am experimenting with providing a list of utxos instead of just a single one.

cdecker:
I'd rather not make this RPC call any more complex than necessary. We can always add a new RPC command, and then we can externalize the coinselection completely, i.e., pass in a list of outputs that should be selected for the funding, and a value, similar to `createrawtransaction` in `bitcoind`.

wythe:
Yeah that makes sense.  Should I even add support for specifying the single utxo?

wythe:
Yes.

jb55:
[...]

PR #1399, opened 2018-04-20
Removed redundancies in withdraw and fundchannel.

wythe:
No new functionality, just a continuation of my work toward completing #665.

I removed the common members of `struct withdrawal` and `struct fund_channel`
and placed them in a new `struct wallet_tx`.  Then it was fairly straightforward
to reimplement the existing code in terms of `wallet_tx`.

Since I made some structural changes I wanted to get this approved before I
go any farther.


ZmnSCPxj:
ACK 2b79542

Please rebase and we should merge this too...

wythe:
I'm not really sure where I went wrong here.  The tests sometimes pass and sometimes don't on my home machine.  It might be better to discard this pull request and let me recreate another.

wythe:
Oh I didn't realize master was broken atm. Will wait.

common/wallet_tx.c
```diff
@@ -0,0 +1,66 @@
...
+	wtx->cmd = cmd;
+	wtx->amount = 0;
+	wtx->change_key_index = 0;
+	wtx->utxos = NULL;
+	wtx->all_funds = false;
+	wtx->desttok = NULL;
+}
+
+bool wtx_json_get_params(struct wallet_tx * tx, const char * buffer,
+	            const jsmntok_t *params, const char * dest)
```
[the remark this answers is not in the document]
wythe: Yeah I noticed that too.  They are different concepts but just happen to be used exactly the same.  I will get to work on the changes this afternoon.

lightningd/jsonrpc.h
```diff
@@ -101,5 +102,9 @@ json_tok_address_scriptpubkey(const tal_t *ctx,
 			      const char *buffer,
 			      const jsmntok_t *tok, const u8 **scriptpubkey);
 
+/* A specialized get_params used for both withdraw and fundchannel */
+bool json_wtx_get_params(struct wallet_tx * tx, const char * buffer,
```
ZmnSCPxj: I would suggest using a `json_tok_wtx` rather than replacing `json_get_params`: we may want to add different arguments to `fundchannel` vs `withdraw` in the future.  So you would retain `json_get_params` but replace the satoshi-parsing with `json_tok_wtx` instead.

wythe: Very good. I'm on it.

wythe:
Ready to go.

lightningd/jsonrpc.c
```diff
@@ -852,3 +853,15 @@ json_tok_address_scriptpubkey(const tal_t *cxt,
 
 	return ADDRESS_PARSE_UNRECOGNIZED;
 }
+
+bool json_tok_wtx(struct wallet_tx * tx, const char * buffer,
```
ZmnSCPxj: Other `json_tok_*` functions are usually of the form `bool json_tok_whatever(const char *buffer, const jsmntok_t *tok, struct whatever *output)` and do not call `command_fail` themselves.  This is acceptable, though.

ZmnSCPxj:
ACK 9c7a7bce140cecc7198826d640c62a822371b35b

Issue #1400, opened 2018-04-21
listfunds returns incorrect address


vwoo:
I successfully sent some testnet bitcoins onchain to my c_lightning node but when I use the listfunds command to verify it lists an incorrect address.

I sent the bitcoins from an lnd node and the transaction can be seen here: https://live.blockcypher.com/btc-testnet/tx/7cfa060c643a4724640ac90397f356b03d0124b1c4a5db5b961264dd79c9cef6/
```
# lncli sendcoins --addr 2NADQntPgVux67EwQ1MkC1DTpy3feie9KjP --amt 2222
{
    "txid": "7cfa060c643a4724640ac90397f356b03d0124b1c4a5db5b961264dd79c9cef6"
}
```

Here is the truncated listfunds response:
```
# lightning-cli listfunds
{
  "outputs": [
    {
      "txid": "7cfa060c643a4724640ac90397f356b03d0124b1c4a5db5b961264dd79c9cef6", 
      "output": 0, 
      "value": 2222, 
      "address": "2Msf3DaEtYk2D8tgMtixAxqLRX9jc4sQWDg", 
      "status": "confirmed"
    }
  ]
}
```

I have no idea where that 2Msf3DaEtYk2D8tgMtixAxqLRX9jc4sQWDg came from. Seens to be an unused address according to https://live.blockcypher.com/btc-testnet/address/2Msf3DaEtYk2D8tgMtixAxqLRX9jc4sQWDg/ and I can't find it in my list of previously generated addresses.
```
# returns nothing
# lightning-cli dev-listaddrs | grep 2Msf3DaEtYk2D8tgMtixAxqLRX9jc4sQWDg
```

The issue is repeatable. Here is a second transaction with the same problem.
```
# lncli sendcoins --addr 2NEPDNMRKER4zDNsWKP5skVJsXNPrSBrwzB --amt 4567
{
    "txid": "b4c0cc588744352214e7419a5eeea288a7c99b0167acb0ee4f1369f8c72cc102"
}

# lightning-cli listfunds
{
  "outputs": [
      {
        "txid": "b4c0cc588744352214e7419a5eeea288a7c99b0167acb0ee4f1369f8c72cc102", 
        "output": 1, 
        "value": 4567, 
        "address": "2MvUHv3QqUfhF7bgSxgPBG5QwGsUHTWbpYv", 
        "status": "confirmed"
      }
    ]
}
```

[...]

PR #1403, opened 2018-04-21
Fix listfunds output address

ZmnSCPxj:
Fixes: #1400 

Eventually. For now we have a failing test confirming the bug report.

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -371,6 +371,7 @@ static void json_listfunds(struct command *cmd, const char *buffer UNUSED,
 	struct utxo **utxos =
 	    wallet_get_utxos(cmd, cmd->ld->wallet, output_state_available);
 	char* out;
+	struct sha256 h;
 	struct ripemd160 h160;
 	struct pubkey funding_pubkey;
 	json_object_start(response, NULL);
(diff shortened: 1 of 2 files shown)


cdecker:
Added the Work in Progress tag for now. Ping me if this is ready to review :-)

wallet/walletrpc.c
```diff
@@ -198,13 +198,63 @@ static const struct json_command withdraw_command = {
 };
 AUTODATA(json_command, &withdraw_command);
 
+/* May return NULL if encoding error occurs. */
+static char *
+encode_pubkey_to_addr(const tal_t *ctx,
```
cdecker: This might actually be interesting to add to `pubkey.c` and make it globally available. In that case I'd rename it to `pubkey_to_addr` to match the rest of the functions there.

ZmnSCPxj: Unfortunately it pulls in `common/bech32` due to `segwit_addr_encode` if we put it in `bitcoin/pubkey`.

The issue is that `common/test/Makefile` assumes that `bitcoin/*` modules will not depend on any `common/*` modules other than the trivial `common/utils` (which does not have its own test).  What happens is that every `common/test` now mocks `segwit_addr_encode`, and also need to include `common/bech32.h`.

`Makefile`s of each subdaemon also get affected (they now need to include `common/bech32`) even if subdaemons never need to print an onchain address anyway.

The effort does not seem commensurate to theoretical benefits...

cdecker: Good point, maybe we can reorg later a bit to move the bech32 dependency into `bitcoin/`, but that's for another time.

ZmnSCPxj:
Rebased

PR #1412, opened 2018-04-23
Use the DNS Seed to locate nodes that we can't find otherwise

### LINKED ISSUE #246: Implement DNS based node lookup and bootstrapping
When losing connection to a peer, that may have restarted, we currently have no way of discovering it's IP/port. The DNS seed exposes functionality to query for a specific peer's contact information so we could use it to locate the peer and reconnect.

Furthermore we could use the DNS seed to bootstrap some random connections, and maybe even channels, similar to `lnd`s autopilot.

### LINKED ISSUE #547: Gossipd should not always need addrhint
gossipd/gossip.c:1428:		/* FIXME: now try node table, dns lookups... */

- [x] We need to search for peers in existing gossip
- [ ] We need to do DNS bootstrap (#246 )

This will also help #390 and #527

cdecker:
Fixes #547 

Fixes #246

### DIFF
--- a/gossipd/gossip.c
+++ b/gossipd/gossip.c
@@ -14,6 +14,8 @@
 #include <ccan/take/take.h>
 #include <ccan/tal/str/str.h>
 #include <ccan/timer/timer.h>
+#include <common/bech32.h>
+#include <common/bech32_util.h>
 #include <common/cryptomsg.h>
 #include <common/daemon_conn.h>
 #include <common/features.h>
@@ -1627,6 +1629,32 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
 	return io_connect(conn, &ai, connection_out, reach);
 }
 
+static struct addrhint *
+seed_resolve_addr(const tal_t *ctx, const struct pubkey *id, const u16 port)
+{
+	struct addrhint *a;
+	char bech32[100], *addr;
+	u8 der[PUBKEY_DER_LEN];
+	u5 *data = tal_arr(ctx, u5, 0);
+
+	pubkey_to_der(der, id);
+	bech32_push_bits(&data, der, PUBKEY_DER_LEN*8);
+	bech32_encode(bech32, "ln", data, tal_count(data), sizeof(bech32));
+	addr = tal_fmt(ctx, "%s.lseed.bitcoinstats.com", bech32);
+
+	status_trace("Resolving %s", addr);
+
+	a = tal(ctx, struct addrhint);
+	if (!wireaddr_from_hostname(&a->addr, addr, port, NULL)) {
+		status_trace("Could not resolve %s", addr);
+		return tal_free(a);
+	} else {
+		status_trace("Resolved %s to %s", addr,
+			     type_to_string(ctx, struct wireaddr, &a->addr));
+		return a;
+	}
+}
+
 static void try_connect(struct reaching *reach)
 {
 	struct addrhint *a;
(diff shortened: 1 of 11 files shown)


gossipd/gossip.c
```diff
@@ -1627,6 +1629,32 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
...
+{
+	struct addrhint *a;
+	char bech32[100], *addr;
+	u8 der[PUBKEY_DER_LEN];
+	u5 *data = tal_arr(ctx, u5, 0);
+
+	pubkey_to_der(der, id);
+	bech32_push_bits(&data, der, PUBKEY_DER_LEN*8);
+	bech32_encode(bech32, "ln", data, tal_count(data), sizeof(bech32));
+	addr = tal_fmt(ctx, "%s.lseed.bitcoinstats.com", bech32);
```
ZmnSCPxj: Probably better to put `lseed.bitcoinstats.com` in an option, but maybe OK as first cut, can always do later.

gossipd/gossip.c
```diff
@@ -1641,6 +1669,10 @@ static void try_connect(struct reaching *reach)
 	}
 
 	a = find_addrhint(reach->daemon, &reach->id);
+
+	if (!a)
+		a = seed_resolve_addr(reach, &reach->id, 9735);
+
 	if (!a) {
 		/* FIXME: now try node table, dns lookups... */
```
ZmnSCPxj: Update this FIXME?

gossipd/gossip.c
```diff
@@ -1627,6 +1629,32 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
...
+
+	pubkey_to_der(der, id);
+	bech32_push_bits(&data, der, PUBKEY_DER_LEN*8);
+	bech32_encode(bech32, "ln", data, tal_count(data), sizeof(bech32));
+	addr = tal_fmt(ctx, "%s.lseed.bitcoinstats.com", bech32);
+
+	status_trace("Resolving %s", addr);
+
+	a = tal(ctx, struct addrhint);
+	if (!wireaddr_from_hostname(&a->addr, addr, 9735, NULL)) {
```
ZmnSCPxj: This function `seed_resolve_addr` has `port` as argument, should it not propagate that from caller instead of hardcoding 9735 here?

practicalswift:
Concept ACK

@cdecker, very nice work!

Out of curiosity – how is the server that responds to A-queries for `ln[bech32].lseed.bitcoinstats.com` populated with data? Is it populated also with testnet data?

cdecker:
> Is it populated also with testnet data?

Currently not, but that can be changed, since there shouldn't be any overlap :-)

ZmnSCPxj:
ACK b05e08a1164ae7108c4cc16f960685562abedeb9

gossipd/gossip.c
```diff
@@ -1641,8 +1669,11 @@ static void try_connect(struct reaching *reach)
 	}
 
 	a = find_addrhint(reach->daemon, &reach->id);
+
+	if (!a)
+		a = seed_resolve_addr(reach, &reach->id, 9735);
+
 	if (!a) {
-		/* FIXME: now try node table, dns lookups... */
```
ZmnSCPxj: We now do the BOLT10 DNS lookup, but maybe node table is still not used? I think, "node table" here is the gossip-derived map of nodes, which contain `struct wireaddr` to contact them with?

PR #1413, opened 2018-04-24
WIP: distinguish network so we can run multiple lightningd easily in same .lightning dir

rustyrussell:
If we think this is a good idea, the final patch needs to be a man page for lightningd!

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -169,7 +169,7 @@ int main(int argc, char *argv[])
 {
 	int fd, i, off;
 	const char *method;
-	char *cmd, *resp, *idstr, *rpc_filename;
+	char *cmd, *resp, *idstr, *rpc_filename, *netname;
 	struct sockaddr_un addr;
 	jsmntok_t *toks;
 	const jsmntok_t *result, *error, *id;
@@ -184,7 +184,8 @@ int main(int argc, char *argv[])
 	jsmn_init(&parser);
 
 	opt_set_alloc(opt_allocfn, tal_reallocfn, tal_freefn);
-	configdir_register_opts(ctx, &lightning_dir, &rpc_filename);
+	netname = tal_strdup(ctx, "testnet");
+	configdir_register_opts(ctx, &lightning_dir, &rpc_filename, &netname);
 
 	opt_register_noarg("--help|-h", opt_usage_and_exit,
 			   "<command> [<params>...]", "Show this message. Use the command help (without hyphens -- \"lightning-cli help\") to get a list of all RPC commands");
@@ -217,6 +218,9 @@ int main(int argc, char *argv[])
 		else
 			format = JSON;
 	}
+
+	/* If they didn't set --rpc-file, do it now */
+	config_finalize_rpc_name(ctx, &rpc_filename, netname);
 
 	if (chdir(lightning_dir) != 0)
 		err(ERROR_TALKING_TO_LIGHTNINGD, "Moving into '%s'",
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -165,6 +165,37 @@ static void add_input(char **cmd, const char *input,
 		tal_append_fmt(cmd, ", ");
 }
 
+static void ignore_err(const char *fmt UNNEEDED, ...)
+{
+}
+
+/* They might specify default network in config, for example.
+ * We don't understand most options, so we ignore failure here.
+ *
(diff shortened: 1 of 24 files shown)


hsmd/hsm.c
```diff
@@ -460,36 +463,52 @@ static void bitcoin_keypair(struct privkey *privkey,
...
+				const char *network_name)
+{
+	return tal_fmt(ctx, "hsm_secret-%s", network_name);
+}
+
+static int open_hsm_secret(const char *filename, int flags)
+{
+#ifdef COMPAT_V052
+	/* Filename used not to include network name */
+	rename("hsm_secret", filename);
```
ZmnSCPxj: I suppose in practice it will not be likely, but in theory it would be possible for somebody running on testnet currently (or mainnet), stop, then update past this commit, then run on mainnet (or testnet), then fail.

In particular `wallet_network_check` comes after `hsm_init`, so the above mistake is possible to make `hsm_secret` of current network to be renamed incorrectly to other network.  It is recoverable by renaming the `hsm_secret` appropriately, but care must be done by those upgrading!

wallet/db.c
```diff
@@ -422,16 +420,22 @@ void db_commit_transaction(struct db *db)
...
+	const char *filename;
 
 	if (SQLITE_VERSION_NUMBER != sqlite3_libversion_number())
 		fatal("SQLITE version mismatch: compiled %u, now %u",
 		      SQLITE_VERSION_NUMBER, sqlite3_libversion_number());
 
+	filename = tal_fmt(tmpctx, "%s.sqlite3", network_name);
+#ifdef COMPAT_V052
+	/* Earlier versions didn't include network name */
+	rename("lightningd.sqlite3", filename);
```
ZmnSCPxj: Similar to previous comment regarding `hsm_secret`. We should take care to inform users before merging this PR as much as possible (maybe send an email on the mailing list? It is rare enough that hopefully people will notice?)

cli/lightning-cli.c
```diff
@@ -169,7 +169,7 @@ int main(int argc, char *argv[])
 {
 	int fd, i, off;
 	const char *method;
-	char *cmd, *resp, *idstr, *rpc_filename;
+	char *cmd, *resp, *idstr, *rpc_filename, *netname;
```
ZmnSCPxj: This will break existing setups which do not set the (new) `--network` argument for `lightning-cli`, although a later commit makes this mostly work.

common/configdir.c
```diff
@@ -44,6 +55,11 @@ void configdir_register_opts(const tal_t *ctx,
 			       netname,
 			       "Select the network parameters (bitcoin, testnet,"
 			       " regtest, litecoin or litecoin-testnet)");
+
+	opt_register_early_noarg("--testnet", opt_set_testnet, netname,
+				 "Alias for --network=testnet");
+	opt_register_early_noarg("--mainnet", opt_set_mainnet, netname,
+				 "Alias for --network=bigtcoin");
```
ZmnSCPxj: New altcoin "Big T Coin"?

lightningd/connect_control.c
```diff
@@ -177,7 +177,13 @@ static void json_connect(struct command *cmd,
...
 			addr.port = port;
 		} else {
-			addr.port = DEFAULT_PORT;
+			/* BOLT #1:
+			 *
+			 * The default TCP port is 9735. This corresponds to
+			 * hexadecimal `0x2607`, the Unicode code point for
+			 * LIGHTNING.
+			 */
+			addr.port = 9735;
```
ZmnSCPxj: Why this in particular for outgoing, and not use default?

rustyrussell: I deleted DEFAULT_PORT for clarity, and to catch anyone assuming it.

cdecker:
I'm a bit uneasy with this change since we originally intended `lightningd` not to be specific to one blockchain, with the hope to eventually have atomic cross-chain swaps and distributed exchange network and similar. This commit would mean that we distance us even more from that idea. Then again we already are pretty much one-chain only at the moment.

rustyrussell:
Well, I never believed that crossing chains makes high-level sense, so I didn't even think about that when I resurrected this patch series.  I wanted to push this now because it's a significant change which we'd want to do pre-0.6, if we want to do it at all...

rustyrussell:
I'm closing this; we may want elements of it later, but it's in the branch if needed.

PR #1414, opened 2018-04-24
build: Add linter checking for accidental introduction of locale dependence

practicalswift:
This linter will check for code accidentally introducing locale dependencies.

Unnecessary locale dependence can cause bugs that are very tricky to isolate and fix. We should avoid using locale dependent functions if possible.

**Note to reviewers 1:** What is the most appropriate `LOCALE_DEPENDENT_FUNCTIONS` function list? Is the current list too strict? What should be removed or added?

**Note to reviewers 2:** Could any of the `KNOWN_EXCEPTIONS` cause problems due to locale dependence?


ZmnSCPxj:
ACK 12ee4f1eeb057d2a37bdb8c5e66aa6bb18ccf29f

practicalswift:
Rebased!

@ZmnSCPxj would you mind re-reviewing? :-)

ZmnSCPxj:
Since #1416 has been merged, maybe this can be skipped? We now use a known-fixed locale now, so the behavior of all locale-dependent functions *should* not vary (and if they do vary still, that is already a bug in the system C library).

PR #1416, opened 2018-04-25
Force the use of the POSIX C locale for all our commands and in third-party commands we invoke via exec{l,lp,v,vp}(...)

practicalswift:
Force the use of the POSIX C locale for all commands and their subprocesses.

```
void setup_locale(void) {
    setlocale(LC_ALL, "C");
    putenv("LC_ALL=C"); /* For exec{l,lp,v,vp}(...) */
}
```

Rationale: To avoid localization related parsing bugs in our code or in localized third-party commands we invoke via `exec{l,lp,v,vp}(...)`.


ZmnSCPxj:
Mostly OK, but Travis says at least one test program does not include `common/utils.h` to get the `setup_locale` prototype.

practicalswift:
@ZmnSCPxj Updated version. Would you mind re-reviewing? :-)

ZmnSCPxj:
ACK f07749e152f052a01c6ab9e0d87e3c5b340bae55

PR #1407, opened 2018-04-22
invoice: Support providing preimage when making invoice.

ZmnSCPxj:
Fixes: #1402

PR #1417, opened 2018-04-25
invoice: Check duplicate preimage when explicitly sprcified.

ZmnSCPxj:
Reported-by: @mcudev


mcudev:
Tested a3a7b79f339495d28241e824c39508ebfbb59f44. It works as expected now. The "preimage already used" response is returned (and logged) and the node stays running. Seems reasonable.
Link to my report of this issue: https://github.com/ElementsProject/lightning/pull/1407#issuecomment-384086699

PR #1408, opened 2018-04-23
Misc test fixes

rustyrussell:
This contains some of the simpler fixes which crept into the ongoing permconnect (#1322) which is taking far too long.  Let's make sure these pass CI and get reviewed, while I work on that one.

PR #1422, opened 2018-04-26
test_lightningd.py: Make test_penalty_inhtlc not fail due to reconnect.

ZmnSCPxj:
The test disconnects at `=WIRE_COMMITMENT_SIGNED-nocommit`,
then reconnects to continue `REVOKE_AND_ACK`, so I think it
should indeed reconnect.

This was introduced in #1408.
I am uncertain why, the last commit in that PR passed, but an
intermediate commit failed on `test_penalty_inhtlc`:

https://travis-ci.org/ElementsProject/lightning/jobs/370024920


rustyrussell:
This was fixed properly in first commit of #1322 ...

PR #1425, opened 2018-04-26
wallet: free sql statement in wallet_onchaind_channels.

rustyrussell:
I have a more thorough fix which wraps sqlite3_stmt, so we get told about leaks in general, but this is the immediate fix.


ZmnSCPxj:
ACK 1bcfb3e09c6195d72f18e88a9e9ca10f3ae3db3c

cdecker:
ACK 1bcfb3e

Issue #1420, opened 2018-04-25
sqlite crash when `--daemon` flag used


redshark1802:
lightnind crashes after pulling the latest master (89ff46f1e68d6e5eb8a8bfafbb58ae2f367ed082).
The error is ``sqlite3_close: unable to close due to unfinalized statements or unfinished backups``.

This also happens with a completely new sqlite database: I made a backup of the current sqlite db and removed it, started lightningd again with ``--port 0``.

redshark1802:
Yes, this is the command I use to start lightningd:
``./lightningd/lightningd --network bitcoin --log-level debug --alias [hidden] --log-file /home/bitcoin/lightning.log --daemon``

There is no additional information in the log, the last line is just ``Listening on 'lightning-rpc'``

ZmnSCPxj:
It seems some of the recent code rearrangement around db close-fork-db reopen process for `--daemon` failed.

Without `--daemon` I suspect highly it will succeed to start, so if you need to keep your node up then try running without `--daemon` in the meantime.  We will need to investigate this further, please wait.

redshark1802:
Thanks, that works

Rav3nPL:
It still exists, if you start fresh with --daemon.
If you start w/o --daemon and close, then --daemon works fine.

9735dev:
To follow-up on @Rav3nPL's comment, if you start a new node with the daemon flag, like below, it fails with this error still. You have to start without `--daemon` one time. Then after the first successful run, you can use daemon.

Easy to reproduce with the following as of commit 69cdfba3c85a2ad82e5f587b2e3c9c3349d4a944:
```
/lightning/lightningd/lightningd --daemon --network=regtest --port=9735 --lightning-dir=/tmp/l1 --bitcoin-datadir=/tmp/n1 --log-file=server.log --log-level=debug --bitcoin-cli=/bitcoin/src/bitcoin-cli --bitcoin-rpcuser=n1 --bitcoin-rpcpassword=n1 --override-fee-rates=253/253/253
```
This first start problem appears to be fixed by #1430 commit 6308c1fee276b47b1231d3ecbe1588e6154bf64f.

PR #1430, opened 2018-04-27
Track db statements

rustyrussell:
This adds minimal tracking to our sqlite3_stmts, to avoid problems like the one we saw causing --daemon to fail in #1420 

Turns out there are more leaks, so this fixes them.

### DIFF
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -396,7 +396,7 @@ void db_stmt_done(sqlite3_stmt *stmt)
 	sqlite3_finalize(stmt);
 }
 
-sqlite3_stmt *db_prepare_(const char *caller, struct db *db, const char *query)
+sqlite3_stmt *db_prepare_(const char *location, struct db *db, const char *query)
 {
 	int err;
 	sqlite3_stmt *stmt;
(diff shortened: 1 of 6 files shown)


ZmnSCPxj:
ACK 6308c1fee276b47b1231d3ecbe1588e6154bf64f

9735dev:
Tested that 6308c1fee276b47b1231d3ecbe1588e6154bf64f appears to fix the daemon first start issue of #1420.

wallet/db.c
```diff
@@ -331,7 +331,72 @@ char *dbmigrations[] = {
...
+}
+
+static void dev_statement_end(sqlite3_stmt *stmt)
+{
+	struct db_statement *dbstat = find_statement(stmt);
+	list_del_from(&db_statements, &dbstat->list);
+	tal_free(dbstat);
+}
+#else
+static void dev_statement_start(sqlite3_stmt *stmt, const char *origin)
```
cdecker: Any particular reason we leave these stubs around? There are very few places we actually call them, so we could just add a `DEVELOPER=1` check in those places.

cdecker: Then again I guess they'll be optimized out by the compiler.

ZmnSCPxj: Yes, it reduces the number of `#ifdef`..`#endif` blocks in the code, and as you indeed mention, they will tend to get optimized out when empty, especially as they are `static` also.

PR #1432, opened 2018-04-27
lightningd: loosen feerate minimum.

### LINKED ISSUE #613: "update_fee outside range" failure for two nodes with identical --override-fee-rates
I have two nodes running with `--override-fee-rates=40000/40000/40000`. I was able to open a channel between them and send payments through it, but after awhile the channel failed and got automatically closed.

These are the logs on the receiver side:

```
lightning_channeld(23839): TRACE: peer_in WIRE_UPDATE_FEE
lightning_channeld(23839): STATUS_FAIL_PEER_BAD: update_fee 265 outside range 40000-200000
lightningd(13849): peer 0298446273407650985c4a3a286ea3a2a0236c3b74eef79a0fb6327472c23f9a5d: Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: update_fee 265 outside range 40000-200000
lightningd(13849):          (tx 4de73b4164d65f11b2d5dd647177c739c48e28bf4e031e3cb3f29d4a60b9e207)
lightningd(13849): sendrawtransaction: 020000000001016ef1a64298320e04aaa6f6709759d7658a1db6bf043e11cab8bb4470ef060a3701000000000541d3800242d8260000000000220020b4d72a553d503bfdbd90dde2405466a0164df0c928a30edec0e3fb5158ccd5878db6d800000000001600145eb81571272192056beee8430117521a29c0b6e40400483045022100cb9cc8ab6dbc6b0685a4b683e86f184c41871f9f87b0b198041761141daf382902201f7226a7dbdf0ccb435c6dec766655c43a015318b43d5bf0ea08e055b350251401483045022100ecb34904bd8c3c1bb14de57eaaabfe32069e6106a2a2b5fc917a1d8ac8e5919102205313f91f6f117d33aacdffc190b298df143e2bfbbf966162ebb6e21e062204900147522102116f068e43fc180a98d30cdedf22689f515f76dfdd346d289c5a74b48d3f90032102ddafa8d4524cd74bdde10d27f860f905db72f7e39b9c618cb1d18a88a449283352ae8ea5f120
lightning_gossipd(13857): TRACE: Channel 1258845:61:1(1) was updated.
lightning_gossipd(13857): TRACE: Received channel_update for channel 1258845:61:1(0)
lightning_gossipd(13857): TRACE: Channel 1258845:61:1(0) was updated.
```

The sender was running without `--log-level` and has no useful logs, sorry!

rustyrussell:
We're getting spurious closures, even on mainnet.  Using --ignore-fee-limits
is dangerous; it's slightly less so to lower the minimum (which is the
usual cause of problems).

So let's halve it.

This is a workaround, until we get independent feerates in the spec.

(Note: I've labeled this as fixes #613 but that should have been fixed by previous update_fee fixes; this addresses more the other comments on that issue which refer to spontaneous closures without --override-fee-rates).

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -162,12 +162,18 @@ u8 *p2wpkh_for_keyidx(const tal_t *ctx, struct lightningd *ld, u64 keyidx)
 
 u32 feerate_min(struct lightningd *ld)
 {
-	if (ld->config.ignore_fee_limits)
-		return 1;
+	u32 min;
 
-	/* Set this to average of slow and normal.*/
-	return (get_feerate(ld->topology, FEERATE_SLOW)
-		+ get_feerate(ld->topology, FEERATE_NORMAL)) / 2;
+	/* We can't allow less than feerate_floor, since that won't relay */
+	if (ld->config.ignore_fee_limits)
+		min = 1;
+	else
+		/* Set this to half of slow rate.*/
+		min = get_feerate(ld->topology, FEERATE_SLOW) / 2;
+
+	if (min < feerate_floor())
+		return feerate_floor();
+	return min;
 }
 
 /* BOLT #2:
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -253,7 +253,7 @@ static void next_updatefee_timer(struct chain_topology *topo);
  * This formula is satisfied by a feerate of 4030 (hand-search).
  */
 #define FEERATE_FLOOR 253
-static u32 feerate_floor(void)
+u32 feerate_floor(void)
 {
 	/* Assert that bitcoind will see this as above minRelayTxFee */
 	BUILD_ASSERT(FEERATE_BITCOIND_SEES(FEERATE_FLOOR, MINIMUM_TX_WEIGHT)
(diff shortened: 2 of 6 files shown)


lightningd/peer_control.c
```diff
@@ -165,9 +165,8 @@ u32 feerate_min(struct lightningd *ld)
 	if (ld->config.ignore_fee_limits)
 		return 1;
 
-	/* Set this to average of slow and normal.*/
-	return (get_feerate(ld->topology, FEERATE_SLOW)
-		+ get_feerate(ld->topology, FEERATE_NORMAL)) / 2;
+	/* Set this to half of slow rate.*/
+	return get_feerate(ld->topology, FEERATE_SLOW) / 2;
```
ZmnSCPxj: Could this potentially violate `feerate_floor()`? `get_feerate` gets its data from `topo->feerate`, which is loaded by `update_feerates`, which respects `feerate_floor()`.  The previous code averaged the slow and normal feerates.  Now we are getting half the slow feerate.  Could the result violate `feerate_floor` and is there some other point where we check against `feerate_floor()`? (it seems unlikely, as, `feerate_floor()` is `static` in `chaintopology`)

rustyrussell: Good point!  We should do that here, too.  I'll rework this logic.

practicalswift:
Concept ACK

cdecker:
This also now fails the CI tests, but that should be a minor thing to fix.

ZmnSCPxj:
ack 65db01c6d8fbeef6494ab300579dd1d0eacad463

PR #1433, opened 2018-04-27
RFC: Remove non-working litecoin support.

rustyrussell:
It's been unmaintained (and undocumented), so let's not encourage people
to cut themselves on it.


NicolasDorier:
It has been working for me (https://main2-btc-ltc.forkbitpay.ninja).

Don't throw the baby out with the bathwater. The reason it does not work now might be an unrelated bug.

ZmnSCPxj:
Ack  ffee842

But note that there is opposition to this change from BtcPay and their users.

NicolasDorier:
I think this is stupid, we can start experimenting cross atomic swap offchain thanks to this at BTCPay level.

cdecker:
NACK, I really like the multi-blockchain aspect of Lightning. I see the point of not wanting to allocate our, already strapped, resources to another target platform, but maybe @NicolasDorier can commit to running the integration tests and coordinate the Litecoin side of things?

practicalswift:
Concept ACK unless someone (like the excellent @NicolasDorier?) volunteers to take on responsibility for Lightning including testing/bug reporting

Concept NACK otherwise :-)

NicolasDorier:
well I am trying to replicate the bug right now.

rustyrussell:
What are the minrelayfee rules for litecoin?  What about the different dust rules?

There are many subtle bitcoin-specific things in lightning.  Litecoin users will lose funds and have bad experiences because we don't care.

Given that litecoin will receive no support from me, I will ignore bug reports and leave others to apply litecoin patches, and will nak them if they add complexity, I suggest you fork the project, instead, to avoid confusion.

NicolasDorier:
> Given that litecoin will receive no support from me, I will ignore bug reports and leave others to apply litecoin patches, and will nak them if they add complexity

Seems good enough to me.

cdecker:
Speaking of fees, we should probably also consolidate any magic fee values in a single place so we can reference them by name :grinning:

NicolasDorier:
Litecoin is currently working (between two clightning instance)

If anything, supporting multiple chain will on the contrary decrease complexity by modelling your code better. (Like @cdecker suggested, the magic values should be set at the chainparams level instead of const)

lepipele:
Really sad to see this. Especially the part about "ignoring litecoin and threatening to nak"... if you don't need some functionality @rustyrussell doesn't mean others don't need it. 

And if you are too lazy to maintain something in codebase, then please don't mask it as "caring about user experience".

rustyrussell:
I'm going to close this; I think I've made my position clear, but I respect Christian's opinion.  No doubt this will re-occur each time we have a bug report using litecoin, until eventually we drop it or decide to support it.

PR #1442, opened 2018-04-30
More cleanups

rustyrussell:
These are mainly pulled from #1413 which we might not be doing, plus one which @practicalswift might want to review (the last one).

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -344,17 +344,6 @@ static void config_register_opts(struct lightningd *ld)
 }
 
 #if DEVELOPER
-static char *opt_set_hsm_seed(const char *arg, struct lightningd *ld)
-{
-	assert(arg != NULL);
-
-	ld->dev_hsm_seed = tal_hexdata(ld, arg, strlen(arg));
-	if (ld->dev_hsm_seed)
-		return NULL;
-
-	return tal_fmt(NULL, "bad hex string '%s'", arg);
-}
-
 static void dev_register_opts(struct lightningd *ld)
 {
 	opt_register_noarg("--dev-no-reconnect", opt_set_bool,
@@ -369,8 +358,6 @@ static void dev_register_opts(struct lightningd *ld)
 			 "Time between gossip broadcasts in milliseconds");
 	opt_register_arg("--dev-disconnect=<filename>", opt_subd_dev_disconnect,
 			 NULL, ld, "File containing disconnection points");
-	opt_register_arg("--dev-hsm-seed=<seed>", opt_set_hsm_seed,
-			 NULL, ld, "Hex-encoded seed for HSM");
 }
 #endif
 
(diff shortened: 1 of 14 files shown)


lightningd/options.c
```diff
@@ -141,13 +153,14 @@ static char *opt_set_s32(const char *arg, s32 *u)
...
 	size_t n = tal_count(ld->wireaddrs);
+	u16 defport = get_portnum(ld);
 	char const *err_msg;
 
 	assert(arg != NULL);
 
 	tal_resize(&ld->wireaddrs, n+1);
 
-	if (!parse_wireaddr(arg, &ld->wireaddrs[n], ld->portnum, &err_msg)) {
+	if (!parse_wireaddr(arg, &ld->wireaddrs[n], defport, &err_msg)) {
```
ZmnSCPxj: It seems, if we do something like the below in our config file:

    ipaddr=127.0.0.1
    port=9999

The `wireaddr` will be `127.0.0.1:9735`?  Because `ipaddr` came before the `port`.  Perhaps we should make this clear in our documentation that `port` should be set before `ipaddr`, or explicitly provide the port in `ipaddr` setting.

tests/utils.py
```diff
@@ -263,10 +256,11 @@ def __init__(self, lightning_dir, bitcoin_dir, port=9735, random_hsm=False):
 
         # Last 32-bytes of final part of dir -> seed.
         seed = (bytes(re.search('([^/]+)/*$', lightning_dir).group(1), encoding='utf-8') + bytes(32))[:32]
+        if not random_hsm:
+            with open(os.path.join(lightning_dir, 'hsm_secret-regtest'), 'wb') as f:
```
ZmnSCPxj: CI complains due to this, since *this* PR does not separate filenames by network type.

practicalswift:
Concept ACK

ACK: `-with-local-modifications` → `-modded`

lightningd/options.c
rustyrussell: You're right: we need to use port there if it's set.

tests/utils.py
rustyrussell: Yes, overzealous cherry pick. Will fix.

lightningd/options.c
rustyrussell: After consideration, this whole option mess needs more work.  I've dropped this patch, and will revisit in another PR.

tests/utils.py
```diff
@@ -263,10 +256,11 @@ def __init__(self, lightning_dir, bitcoin_dir, port=9735, random_hsm=False):
 
         # Last 32-bytes of final part of dir -> seed.
         seed = (bytes(re.search('([^/]+)/*$', lightning_dir).group(1), encoding='utf-8') + bytes(32))[:32]
+        if not random_hsm:
+            with open(os.path.join(lightning_dir, 'hsm_secret'), 'wb') as f:
```
ZmnSCPxj: Failing in CI, probably because the `lightning_dir` might not exist yet at this point.

cdecker: Assuming this comment predates the latest version, since the call to `makedirs` is 5 lines above this :wink:

lightningd/options.c
```diff
@@ -184,6 +184,16 @@ static char *opt_set_network(const char *arg, struct lightningd *ld)
 	return NULL;
 }
 
+static char *opt_set_testnet(struct lightningd *ld)
```
cdecker: This overlaps with the `--network` names now.

lightningd/options.c
```diff
@@ -326,6 +336,10 @@ static void config_register_opts(struct lightningd *ld)
 			       ld,
 			       "Select the network parameters (bitcoin, testnet,"
 			       " regtest, litecoin or litecoin-testnet)");
+	opt_register_early_noarg("--testnet", opt_set_testnet, ld,
+				 "Alias for --network=testnet");
+	opt_register_early_noarg("--mainnet", opt_set_mainnet, ld,
+				 "Alias for --network=bigtcoin");
```
cdecker: Stop pumping your altcoin xD

ZmnSCPxj:
Mostly OK, would ACK except for the Big T Coin pump, will ACK once I get premine privileges on it.

cdecker:
The CI failure was a flake, I really need to take a closer look at the tests some time soon.

ZmnSCPxj:
ACK 225a07dfaadf32090ee5719db4de896089ddd2d5

Issue #1394, opened 2018-04-19
API change: `close` accept channel ID


ZmnSCPxj:
In the (far) future we probably want to support multiple channels per peer, despite this being suboptimal for network health.  Otherwise what is point of making `channels` an array in `listpeers`?

In line with that, we should probably have `close` accept a channel ID rather than peer ID.  And, we should probably transition to this earlier rather than later.

For transitioning:

1.  Support peer ID, OR channel ID.  We can differentiate by size: peer ID is 33 bytes, channel ID is 32 bytes (maybe even support short channel ID "xx:y:z")
2.  If peer ID specified, only valid if peer has one channel exactly.  If more than one channel, error and tell user that multiple channel ID exists that can be closed.

cdecker:
We are currently limited to a single active channel, i.e., not `onchaind` or `closingd`, per peer. Not sure if we want to lift this limitation any time soon since it is already possible to have channels open back-to-back, with only minor downtime.

But this improvement is a good idea either way.

ZmnSCPxj:
A wrinkle here is that we support closing uncommitted channels.  Uncommitted channels do not have a stable funding transaction yet, and so do not have a channel ID.  So I suppose the best thing to do is to retain support for peer id indefinitely, so that we can close uncommitted channels. Hmm.

jsarenik:
Just a summary as a preparation for me to understand what's going on here in order to make sure the documentation is ready for the release:

The related documentation is in file [`doc/lightning-close.7.txt`](https://github.com/ElementsProject/lightning/blob/52c243852ea36b2053bb08879e0293321b3a8380/doc/lightning-close.7.txt).

The related source is in [`lightningd/peer_control.c`](https://github.com/ElementsProject/lightning/blob/435e85a5b295076764f94959b7c65152658069eb/lightningd/peer_control.c). I suppose that `onchaind/onchain.c` is not related.

When the source is changed, the tests in `tests/test_lightning.py` need to be updated as well. Please correct me if I'm wrong or missed out something important.

ZmnSCPxj:
>The related documentation is in file [`doc/lightning-close.7.txt`](https://github.com/ElementsProject/lightning/blob/52c243852ea36b2053bb08879e0293321b3a8380/doc/lightning-close.7.txt).

Yes

[...]

PR #1444, opened 2018-05-01
Use channel id for `close` parameter

ZmnSCPxj:
Fixes: #1394

### DIFF
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -1263,10 +1263,25 @@ class LightningDTests(BaseLightningDTests):
 
         l2.daemon.wait_for_log('to_self_delay 100 larger than 99')
 
+    def test_closing_id(self):
+        """Test closing using peer ID and full channel ID"""
+        l1, l2 = self.connect()
+
+        # Close by full channel ID.
+        self.fund_channel(l1, l2, 10**6)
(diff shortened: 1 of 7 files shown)


cdecker:
ACK 8a583dd

tests/test_lightningd.py
```diff
@@ -1439,17 +1454,16 @@ def test_closing_different_fees(self):
...
         # when valgrind is enabled.
-        # (close timeout defaults to 30 as of this writing, more
-        # than double the default)
-        closes = [self.executor.submit(l1.rpc.close, p.info['id'], False, 72) for p in peers]
+        # (close timeout defaults to 30 as of this writing)
+        closes = [self.executor.submit(l1.rpc.close, p.channel, False, 90) for p in peers]
 
         for c in closes:
-            c.result(72)
+            c.result(90)
```
cdecker: Looking at the logs on Travis this doesn't seem to be a timeout issue at all, there's more than a minute of radio-silence, no progress on fee negotiation.

ZmnSCPxj: You may be right; there may be a heisenbug here when closing multiple channels at the same time. In one run I am looking at now, all 6 peers transitioned their channel to `CLOSINGD_COMPLETE`, but `l1` only transitioned 5 channels to `CLOSINGD_COMPLETE`, so something is broken somewhere.

PR #1446, opened 2018-05-01
make: Disable DEVELOPER by default

cdecker:
We had quite a few users running into issues because the `--dev-xyz` options and
`dev-xyz` RPC calls were available. Before a release we should make sure that
the default compilation flags are safe.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ VALGRIND=valgrind -q --error-exitcode=7
 VALGRIND_TEST_ARGS = --track-origins=yes --leak-check=full --show-reachable=yes --errors-for-leak-kinds=all
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ VALGRIND_TEST_ARGS = --track-origins=yes --leak-check=full --show-reachable=yes
 endif
 
 # By default, we are in DEVELOPER mode, use DEVELOPER= on cmdline to override.
```
ZmnSCPxj: Comment needs updating.

ZmnSCPxj:
ACK f2d9f1e06c29e89be1380513633fdb5fc6834ded

PR #1450, opened 2018-05-03
Network arg rework

rustyrussell:
Our network args were a mess, as Tor and local socket options have made clear.

What do people think of going in this direction?


lightningd/options.c
```diff
@@ -767,11 +768,11 @@ void handle_opts(struct lightningd *ld, int argc, char *argv[])
 
 	check_config(ld);
 
-	if (ld->portnum && tal_count(ld->wireaddrs) == 0)
+	if (ld->portnum && ld->listen && tal_count(ld->wireaddrs) == 0)
 		guess_addresses(ld);
 	else
 		log_debug(ld->log, "Not guessing addresses: %s",
-			  ld->portnum ? "manually set" : "port set to zero");
+			  (ld->portnum && ld->listen) ? "manually set" : "port set to zero");
```
ZmnSCPxj: Since `ld->listen` is now involved, it may seem strange if user sets `portnum` to nonzero and then set `--offline` and then get the log message "port set to zero".  Maybe better "listening disabled"?

common/wireaddr.h
```diff
@@ -37,6 +37,14 @@ struct wireaddr {
 	u16 port;
 };
 
+/* We use wireaddr to tell gossipd both what to listen on, and what to
+ * announce */
+enum addr_listen_announce {
+	ADDR_LISTEN = 1,
+	ADDR_ANNOUNCE = 2,
+	ADDR_LISTEN_AND_ANNOUNCE = 3
```
ZmnSCPxj: We take advantage that `ADDR_LISTEN_AND_ANNOUNCE` == `ADDR_LISTEN | ADDR_ANNOUNCE`.  Maybe it would be better to be explicit about that here?

    enum addr_listen_announce {
        ADDR_LISTEN = 1 << 0, // be explicit that we are using bits
        ADDR_ANNOUNCE = 1 << 1,
        ADDR_LISTEN_AND_ANNOUNCE = ADDR_LISTEN | ADDR_ANNOUNCE // be explicit that this is combination of bits.
    };

Otherwise it may naively seem that numbers are just incrementing.

gossipd/gossip_wire.csv
```diff
@@ -13,15 +13,16 @@ gossipctl_init,,lflen,u16
 gossipctl_init,,lfeatures,lflen*u8
 gossipctl_init,,num_wireaddrs,u16
 gossipctl_init,,wireaddrs,num_wireaddrs*struct wireaddr
+gossipctl_init,,announce_listen,num_wireaddrs*u32
```
ZmnSCPxj: While `u32` on most systems using gcc will match `enum`, C99 standard 6.7.2.2 is:

> Each enumerated type shall be compatible with `char`, a signed integer type, or an unsigned integer type. The choice of type is implementation-defined, but shall be capable of representing the values of all the members of the enumeration.

This may bite us later.  Under gcc `sizeof(enum foo) == sizeof(int)` and in most systems `sizeof(int) == sizeof(uint32_t)`.

tests/test_lightningd.py
```diff
@@ -656,7 +656,7 @@ def test_connect_by_gossip(self):
         l3port = self.node_factory.get_next_port()
         l3 = self.node_factory.get_node(options={"addr": "127.0.0.1:{}".format(l3port)})
```
ZmnSCPxj: This special casing can be removed now I think?  The code below was added in `tests/utils.py`:

    +            'addr': '127.0.0.1:{}'.format(port),

Saibato:
--addr can then easily also be used when the tor PR reemerges for tor addresses :+1: 

but  dev_hsm_seed I would keep in, its easy to guess when c-ln started first to create initial hsm which in fact is not only a seed it is in fact the key.
I prefer to provide the option to set our own bip32 start and not to mandate to trust PRNG

gossipd/gossip_wire.csv
rustyrussell: Yeah, that was kind of a hack (though you'd get a compiler warning if it was wrong).  We can actually provide proper mashalling for this type.

tests/test_lightningd.py
rustyrussell: Yes, good point.

rustyrussell:
Documentation is definitely required.  I was originally going for separate options, before I used prefixes. I think I like it better: it's then clear that you can only use --bind-add with local sockets, and only --announce-addr with Tor.

I will revise tomorrow based on the feedback here, thanks!

gossipd/gossip.c
```diff
@@ -1552,69 +1555,60 @@ static struct io_plan *connection_in(struct io_conn *conn, struct daemon *daemon
...
+			fd = make_listen_fd(AF_INET6, &addr6, sizeof(addr6),
+					    true);
+			if (fd >= 0) {
+				status_trace("Created IPv6 listener on port %u",
+					     daemon->wireaddrs[i].port);
+				io_new_listener(daemon, fd, connection_in,
+						daemon);
+			break;
+		case ADDR_TYPE_PADDING:
+			/* Shouldn't happen. */
```
cdecker: Could be `fatal`?

tests/test_lightningd.py
```diff
@@ -288,7 +288,7 @@ class LightningDTests(BaseLightningDTests):
     def connect(self, may_reconnect=False):
         l1 = self.node_factory.get_node(may_reconnect=may_reconnect)
         l2 = self.node_factory.get_node(may_reconnect=may_reconnect)
-        ret = l1.rpc.connect(l2.info['id'], 'localhost', l2.info['port'])
+        ret = l1.rpc.connect(l2.info['id'], 'localhost', l2.info['address'][0]['port'])
```
cdecker: We should just shortcut this and make the `port` an attribute of the `LightningNode`, no need to keep the structure of the `getinfo` call

tests/test_lightningd.py
cdecker: I just stumbled over this myself, while parallelizing node startups, would be nice to remove this.

gossipd/gossip_wire.csv
ZmnSCPxj: Also nit: you consistently use `listen_announce` elsewhere, and only this csv file uses `announce_listen`.

tests/test_lightningd.py
rustyrussell: We should actually use an ephemeral port for testing.

I thing we should enhance --addr to allow:
1. Hostname variants: 
    * `` (empty string) - All IPv4 & IPv6 interfaces.  This would be default unless of course --autolisten=0.  Announce if addresses seem routable.
    * `0.0.0.0`: All IPv4 interfaces.  Announce if address seems routable.
    * `::`: All IPv6 interfaces.  Announce if address seems routable.
2. Port variants:
    * ``: (empty string) default port, ie. 9735.
    * `:0`: an ephemeral port.

This is more of a rework though, since guess_addresses needs to move into gossipd, when then needs to feed the actual address results *back* to the master for getinfo.

lightningd/options.c
```diff
@@ -339,7 +339,7 @@ static void config_register_opts(struct lightningd *ld)
 	opt_register_early_noarg("--testnet", opt_set_testnet, ld,
 				 "Alias for --network=testnet");
 	opt_register_early_noarg("--mainnet", opt_set_mainnet, ld,
-				 "Alias for --network=bigtcoin");
+				 "Alias for --network=bitcoin");
```
ZmnSCPxj: But my premine!!

tests/test_lightningd.py
```diff
@@ -643,9 +654,9 @@ def test_connect_by_gossip(self):
         l2 = self.node_factory.get_node()
         # Force l3 to give its address.
         l3port = self.node_factory.get_next_port()
-        l3 = self.node_factory.get_node(options={"ipaddr": "127.0.0.1:{}".format(l3port)})
+        l3 = self.node_factory.get_node(options={"addr": "127.0.0.1:{}".format(l3port)})
```
ZmnSCPxj: Previous comment not yet covered?  This can be replaced outright now since we now use `addr` in the factory, which also adds the address in the announcement now; basically the `addr` specification here is now redundant.

    -        l3port = self.node_factory.get_next_port()
    -        l3 = self.node_factory.get_node(options={"addr": "127.0.0.1:{}".format(l3port)})
    +        l3 = self.node_factory.get_node()

We can also remove the `get_next_port` method in the factory too, since I added that only to be able to get at the port number for `l3`, and paving the way for ephemeral ports.

rustyrussell: Yes, decided to leave that for the ephemeral port changes.

cdecker: I have a patch that randomizes ports and allows parallel tests, I'll clean up and open a or tomorrow

ZmnSCPxj: @cdecker, I like the current system where if I have this:

    l1 = self.node_factory.get_node()
    l2 = self.node_factory.get_node()
    l3 = self.node_factory.get_node()

In the logs, I can filter:

    lightningd(16331) -> l1
    lightningd(16332) -> l2
    lightningd(16333) -> l3

So even with randomized ports, I think it would be nice if each test could show the index of the `lightningd` easily in the logs.

One possibility would be to have a node-factory-factory in the `self` object.  This allocates IDs in batches of 100.  Then a node-factory created for each individual test allocates an ID for each created `lightningd`.  So for example one test would have `l1` -> `lightningd(101)`, `l2` -> `lightningd(102)`... and another test would have `l1` -> `lightningd(201)`, `l2` -> `lightningd(202)`...

cdecker: Yep, I have added a `lightningd-{node_id}` prefix that'll allow to filter by a specific node. See commit 3b33a5c65945e0d5b82a81af466188350b3a5329 for details

rustyrussell:
OK, rebased, and more clean ups.  I included @icota 's unix socket support, but not yet @Saibato 's Tor support (that will be a separate PR).

ZmnSCPxj:
ACK 8089103b5dbf6ba4b60585926a61b0c5afb90016

PR #1452, opened 2018-05-03
I added "git submodule update --init" where it should be (i think)

Perlover:
Without it the lightning will not be builded, i think ;-)


ZmnSCPxj:
We already have this in a `Makefile`: https://github.com/ElementsProject/lightning/blob/333dcbf3736cec86080e5dc1e81fb067bd0d166f/external/Makefile#L24-L27

Issue #1454, opened 2018-05-03
CPPFLAGS & LDFLAGS from environment are not passed and replaced your ones


Perlover:
### Issue and Steps to Reproduce

**\[UPDATE\]** How to reproduce: https://github.com/ElementsProject/lightning/issues/1454#issuecomment-386396911

If user has `CPPFLAGS` & `LDFLAGS` environments as set and exported (these variables consist flags for gcc & linker) then your Makefile redefines its and `configure` libraries (`external/*`) gets incorrect `LDFLAGS` and `CPPFLAGS` variables.

#### Your Makefile:

https://github.com/ElementsProject/lightning/blob/master/Makefile#L163
https://github.com/ElementsProject/lightning/blob/master/Makefile#L171

*For example*: I cannot compile your c-lightning in CentOS 6.* from user `bitcoin` because i have my own gcc & autotools programs in $HOME. I have as defined these variables but your make's recipe commands get your ones not my ones. It's incorrect behaviour.

#### The [make docs](https://www.gnu.org/software/make/manual/make.html#Environment):

    When make runs a recipe, variables defined in the makefile are placed into
    the environment of each shell. This allows you to pass values to sub-make
    invocations (see Recursive Use of make). By default, only variables that
    came from the environment or the command line are passed to recursive
    invocations. You can use the export directive to pass other variables. See
    Communicating Variables to a Sub-make, for full details. 

So if i have `CPPFLAGS` & `LDFLAGS` are defined in my environment all **recipes** in your Makefiles and sub-Makefiles will get your CPPFLAGS & LDFLAGS variables not my ones. And i cannot compile your c-lightning (because `gcc` and other tools should look in directories which i defined in these variables).

#### Correct behaviour:

I'm sure you should consider the value of these variables from outside. I checked Makefiles created by third-party programs (bitcoin core for example) which i compilied successfully. The `configure` script consides these variables and to add own options not replace.

#### Conclusion

May be to use `+=` operation for these flags not `=`. I will try to make patch.

Perlover:
I made patch but it didn't help to me
Now i restore CPPFLAGS & LDFLAGS from user but i noticed  that it's not main problem

My patch was:

    diff --git a/Makefile b/Makefile
    index 7248739..6d8d624 100644
    --- a/Makefile
    +++ b/Makefile
    @@ -31,8 +31,8 @@ COVFLAGS = --coverage
     endif
     
     ifeq ($(PIE),1)
[...]

PR #1456, opened 2018-05-03
Makefile: Do not override CFLAGS, CPPFLAGS, and LDFLAGS.

ZmnSCPxj:
Fixes: #1454

Instead use CL_FLAGS, CL_CPPFLAGS, and CL_LDFLAGS for our own flags.
Leave CFLAGS, CPPFLAGS, and LDFLAGS to the builder of the program.

@Perlover please check this and confirm it can work for your use-case.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,15 +160,22 @@ WIRE_GEN := tools/generate-wire.py
 
 ALL_PROGRAMS =
 
-CPPFLAGS = -DBINTOPKGLIBEXECDIR='"'$(shell sh tools/rel.sh $(bindir) $(pkglibexecdir))'"'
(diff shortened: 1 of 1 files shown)


Perlover:
Now i even installed by this command :)

    make -j28 prefix=$HOME && make prefix=$HOME install

May be update docs about installation? :)

ZmnSCPxj:
Okay, thank you for confirming @Perlover.

Ready for review.

rustyrussell:
I deeply disagree. User should fix their make love for their special case.

cdecker:
> I deeply disagree. User should fix their make love for their special case.

So your preferred solution would for the user to make local modifications to the `Makefile`. That could work.

PR #1459, opened 2018-05-05
travis: Fix the locale in the builder images

cdecker:
These were spamming the logs and making the log view almost unusable.

Forgot I had this laying around, and it got dropped in the latest push to the docker hub, I'll regenerate them once this is merged in.

### DIFF
--- a/contrib/Dockerfile.builder
+++ b/contrib/Dockerfile.builder
@@ -3,6 +3,7 @@ MAINTAINER Christian Decker <decker.christian@gmail.com>
 
 ENV DEBIAN_FRONTEND noninteractive
 ENV BITCOIN_VERSION 0.16.0
+
 WORKDIR /build
 
 RUN apt-get -qq update && \
@@ -16,6 +17,7 @@ RUN apt-get -qq update && \
 	software-properties-common \
 	build-essential \
 	autoconf \
+	locales \
 	libtool \
 	libprotobuf-c-dev \
 	libsqlite3-dev \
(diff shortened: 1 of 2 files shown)


contrib/Dockerfile.builder
```diff
@@ -33,11 +35,16 @@ RUN apt-get -qq update && \
...
+
 RUN cd /tmp/ && \
     wget https://bitcoin.org/bin/bitcoin-core-$BITCOIN_VERSION/bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz -O bitcoin.tar.gz && \
     tar -xvzf bitcoin.tar.gz && \
     mv /tmp/bitcoin-$BITCOIN_VERSION/bin/bitcoin* /usr/local/bin/ && \
     rm -rf bitcoin.tar.gz /tmp/bitcoin-$BITCOIN_VERSION
 
 RUN pip3 install --upgrade pip && \
-    pip3 install python-bitcoinlib==0.7.0 pytest==3.0.5 setuptools==36.6.0 pytest-test-groups==1.0.3 flake8==3.5.0 pytest-rerunfailures==3.1
+    python3 -m pip install python-bitcoinlib==0.7.0 pytest==3.0.5 setuptools==36.6.0 pytest-test-groups==1.0.3 flake8==3.5.0 pytest-rerunfailures==3.1
```
ZmnSCPxj: You change `s/pip3/python -m pip/` on this line, but not on the previous line?  This is deliberate?

cdecker: There is an outstanding issue that breaks the `pip3` wrapper after the upgrade. Hence the `pip3 install --upgrade pip` works, but afterwards `pip3` will fail due to the upgrade itself :-S

ZmnSCPxj:
ACK 63557e90047505f8d301e3ae505e402a7ba3fb55

PR #1460, opened 2018-05-05
pytest: Enable parallel test execution and start migration to modern fixtures

cdecker:
I got a bit sidetracked while implementing the `local_channel_updates` patch (that's going to be my next PR). This PR modernizes the integration test environment using the pytest style fixtures, flattening the test structure (functions instead of class methods) and begins the migrating the tests into smaller, better structured test files. The goal is to split the huge `test_lightningd` into smaller, topic based test files that can be selectively executed for changes, e.g., executing `test_gossip.py` if we changed gossip and don't expect failures in other parts of the suite (CI will still catch those).

One major change is that we now have the tests isolated into their own directory and selecting random ports for all daemons, which means that we can run the tests in parallel with `pytest-xdist`:

```
$ NO_VALGRIND=1 DEVELOPER=1 TEST_DEBUG=0 py.test -v tests/ -p no:logging -n 25
...
=== 99 passed, 1 skipped in 137.54 seconds ===
```

Which should speed up the change-test-fix cycle alot. One downside is that we add a few required dependencies: `pytest` and `ephemeral-port-reserve`, and since we're at it we also add `flaky`. These can be easily installed via `pip3 install -r tests/requirements.txt` and I strongly suggest using virtualenvs for these anyway.

`flaky` now also works as a decorator, it'll just rerun the test once (and really produce an independent result thanks to the directory isolation). This is more of a stop-gap solution to be able to work with the CI, and should not free us from actually stabilizing the test though :wink:

### DIFF
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -1,5 +1,6 @@
 from concurrent import futures
 from decimal import Decimal
+from utils import wait_for
 
 import copy
 import json
@@ -64,14 +65,6 @@ def setupBitcoind(directory):
     elif bitcoind.rpc.getwalletinfo()['balance'] < 1:
         logging.debug("Insufficient balance, generating 1 block")
         bitcoind.generate_block(1)
-
-
-def wait_for(success, timeout=30, interval=0.1):
-    start_time = time.time()
-    while not success() and time.time() < start_time + timeout:
-        time.sleep(interval)
-    if time.time() > start_time + timeout:
-        raise ValueError("Error waiting for {}", success)
 
 
 def wait_forget_channels(node):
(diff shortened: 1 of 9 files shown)


cdecker:
Just as a side note: I did not enable parallel test execution on Travis yet, since it was exacerbating the flakiness of some of the tests, but the builder already has all the depencies and with `-n=2` we can cut the test time almost in half, once we fix the flakyness.

tests/test_lightningd.py
```diff
@@ -651,9 +654,7 @@ def test_connect_by_gossip(self):
         """
         l1 = self.node_factory.get_node()
         l2 = self.node_factory.get_node()
-        # Force l3 to give its address.
-        l3port = self.node_factory.get_next_port()
-        l3 = self.node_factory.get_node(options={"ipaddr": "127.0.0.1:{}".format(l3port)})
+        l3 = self.node_factory.get_node(options={"ipaddr": "127.0.0.1"})
```
ZmnSCPxj: Is port not needed here?  The reason why `ipaddr` is specified is to make `l3` gossip its own address and port correctly, else `l1` will be unable to look up `l3` via gossip.

#1450 makes `l3` gossip correctly (currently we only gossip `--ipaddr` or what is auto-detected as our address) so the option could be removed there.

tests/utils.py
```diff
@@ -284,7 +284,7 @@ def __init__(self, lightning_dir, bitcoin_dir, port=9735, random_hsm=False):
             self.opts['dev-broadcast-interval'] = 1000
             if not random_hsm:
                 self.opts['dev-hsm-seed'] = binascii.hexlify(seed).decode('ascii')
-        self.prefix = 'lightningd(%d)' % (port)
+        self.prefix = 'lightningd-%d' % (node_id)
```
ZmnSCPxj: If two tests run in parallel, would they get separate `node_factory`?  We instantiate a new factory at `setUp`.  Sorry, but I am unfamiliar with the Python test environment (and Python in general); can tests in a single test class run in parallel or do they have to be separate standalone functions like in `test_gossip.py` in order to run in parallel?

rustyrussell:
I have another patch which guesses ports and restarts if it guesses wrong. But we can put that in after, to remove ephemeral ports dependent.

rustyrussell:
Also, I hate this. Make already handles parallelism, so doing it again in pytest is redundant (and wrong, though it works in this case since make ends up just running pytest after everything else is finished). However, running individual tests via make gives v ugly output, so I had to redirect it all and that's also ugly.

End "Old man yells at cloud" :)

tests/test_lightningd.py
ZmnSCPxj: Hmm the CI seems to work, so I guess it does work?  Hmm.

rustyrussell:
Rebased, will apply once CI happy...

rustyrussell:
Had to restart a couple of tests, will submit a patch for one of them which I think I understand how it ran too slow...

rustyrussell:
Ack 07f2e7d3b7eea815e60b7a7958e2ffcc9aa1506d

PR #1461, opened 2018-05-05
Send a local channel_update upon funding_locked so that the incoming halfchan becomes announceable

cdecker:
This should make PR #1330 testable: we send a channel_update for non-public channel directions so that our peer can make use of the channel for incoming payments.

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -210,6 +210,9 @@ struct chan *new_chan(struct routing_state *rstate,
 	size_t n;
 	struct node *n1, *n2;
 
+	/* We should never add a channel twice */
+	assert(!uintmap_get(&rstate->chanmap, scid->u64));
+
 	/* Create nodes on demand */
 	n1 = get_node(rstate, id1);
 	if (!n1)
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -1047,9 +1047,11 @@ u8 *handle_channel_update(struct routing_state *rstate, const u8 *update)
 	if (!routing_add_channel_update(rstate, serialized))
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Failed adding channel_update");
-	/* Only store updates for public channels */
-	if (chan->public)
-		gossip_store_add_channel_update(rstate->store, serialized);
+
+	/* Store the channel_update for both public and non-public channels
+	 * (non-public ones may just be the incoming direction). We'd have
+	 * dropped invalid ones earlier. */
+	gossip_store_add_channel_update(rstate->store, serialized);
 
 	return NULL;
 }
--- a/tests/test_lightningd.py
+++ b/tests/test_lightningd.py
@@ -1135,11 +1135,11 @@ class LightningDTests(BaseLightningDTests):
 
     def test_pay(self):
         l1, l2 = self.connect()
-
         chanid = self.fund_channel(l1, l2, 10**6)
 
         # Wait for route propagation.
         self.wait_for_routes(l1, [chanid])
+        sync_blockheight([l1, l2])
 
         inv = l2.rpc.invoice(123000, 'test_pay', 'description')['bolt11']
         before = int(time.time())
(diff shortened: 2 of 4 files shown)


tests/test_lightningd.py
```diff
@@ -2335,6 +2337,7 @@ def test_permfail_htlc_out(self):
         bitcoind.generate_block(1)
         wait_forget_channels(l2)
 
+    @unittest.skipIf(not DEVELOPER, "needs DEVELOPER=1")
```
ZmnSCPxj: Why does this need `DEVELOPER=1` now?  I do not see the test now using a `dev` command or option, unless I missed it?

(actually, for these `skipIf(not DEVELOPER` deals, should we not give the reason why we need `DEVELOPER=1` instead of the obvious `"needs DEVELOPER=1"`? E.g. `"needs --dev-foo-option"` or `"needs dev-foo-command so we can foo the funds before barring them"`)

gossipd/routing.c
```diff
@@ -1047,9 +1047,10 @@ u8 *handle_channel_update(struct routing_state *rstate, const u8 *update)
 	if (!routing_add_channel_update(rstate, serialized))
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Failed adding channel_update");
-	/* Only store updates for public channels */
-	if (chan->public)
-		gossip_store_add_channel_update(rstate->store, serialized);
+	/* Store the channel_update for both public and non-public channels
+	 * (non-public ones may just be the incoming direction. We'd have
+	 * dropped invalid ones earlier. */
```
ZmnSCPxj: Missing close parentheses `(non-public ones .... earlier.)`

gossipd/routing.c
```diff
@@ -1047,9 +1047,10 @@ u8 *handle_channel_update(struct routing_state *rstate, const u8 *update)
...
+	gossip_store_add_channel_update(rstate->store, serialized);
```
ZmnSCPxj: The earlier commit, `create_channel_update` in `channeld/channel.c` creates a `channel_update` with all-0s signature.  And above this, there is a check that the signature matches (call to `check_channel_update`).  Will such a "local channel update" created by `channeld` reach this point?

gossipd/routing.c
cdecker: Fixed :-)

tests/test_lightningd.py
cdecker: Yeah, this is a bit hidden, but we set the `--dev-broadcast-interval` if we have `DEVELOPER=1` which speeds up the staggered broadcast needed in this test 30x, without it we'd just be sitting there for over a minute and wait for the gossip to sync. I changed the `needs DEVELOPER=1` to hint at this requirement.

ZmnSCPxj:
This comment seems to have gotten suppressed somehow: https://github.com/ElementsProject/lightning/pull/1461/commits/86d24093e92e6903c313fa02e7ace785e7ea0580#r186274912

My understanding is that the initial `channel_update` from `channeld` is unsigned, but `gossipd` requires properly-signed `channel_update` always.

gossipd/routing.c
cdecker: The comment on the `send_temporary_announcement` was outdated: we don't send a `channel_announcement`, instead just telling `gossipd` with a `local_add_channel`, and the `channel_update` is signed, in order to avoid having to special case this. It's also what we discussed during last week's call, and others are happy to use a fully signed update for this. It won't make much sense to nodes that haven't seen the `channel_announcement` anyway :wink: 

Changed the stale function comment for `send_temporary_announcement` to reflect these changes.

cdecker:
Yeah, rather annoying that GH folds comments away. I added an explanation inline. TL;DR: the updates are now fully signed, no need for a special case there :-)

ZmnSCPxj:
ACK ce34460517bc132aacf0ea9a82575bc88386af29

cdecker:
Rebased and squashed, no other changes, recycling @ZmnSCPxj's ACK:

ACK db7a49c

Issue #1360, opened 2018-04-11
manpages


ZmnSCPxj:
As we already have manpages for many commands, we should go whole-hog and make manpages for all commands.

We even have manpages for obscure commands (`autocleaninvoice`) but some important commands are missing description!

* [x] `fundchannel` #1463
* [x] `connect` #1427
* [x] `disconnect`
* [x] `listpeers`
* [x] `listchannels`
* [x] `close`  #1355 
* [x] `withdraw` #1427 
* [x] `newaddr` #1427
* [x] `listfunds` #1427
* [x] `lightningd` itself #2899

wythe:
Agreed, this is something I've had my eye on.  I've wound down some of my other activities and plan to spend a lot more time on this project.

fixone:
I've created PR #1406 that covers 2 of the 8 still needed manpages

wythe:
As far as I can tell, the actual JSON RPC interface is not documented anywhere (I figured it out using printf).  Maybe we could use a `lightning-rpc` manpage. Or, the yet to be written `lightningd` page will define it.

ZmnSCPxj:
We are supposedly following this spec for JSON-RPC: http://www.jsonrpc.org/specification

Note however that we do not support batch command processing as far as I can tell.

wythe:
That makes sense.  Thanks.  I knew I was missing something.

wythe:
So far we just point everyone to this issue instead of opening a new one for each.  I suggest you just write them one at a time, making a PR for each.  `disconnect` looks like a good one to start with.

And yes, `lightningd` is a bit daunting!

wythe:
For the initial lightningd manpage, my thoughts are that it might be best to just present a concise overview with pointers to other manpages for the details.  Take a look at the git manpage for inspiration. 

It also might be a good place to walk the reader through the steps of opening a channel and processing payments.  This would be similar to giteveryday(7), but I don't think requires a separate manpage yet.

I agree that lightningd-config could/should be reworked like @host and @hawkmauk suggests, but maybe that can be done in a later step.

My two cents!

hawkmauk:
Thanks for the feedback, I like the idea of including steps for actions such as opening channels and making payments.  I would expect the man page to include command line args though which would only serve to duplicate what is in lighningd-config which is my main concern.

ZmnSCPxj:
The `lightningd` could ahow the more likely of interest config options, then point to `lightningd-config` for details

[...]

PR #1463, opened 2018-05-05
doc: Added fundchannel man page.

wythe:
This describes `fundchannel` as it works today (as I understand it). #1360

It definitely could use better error codes instead of -1 for everything.

Is there a better way to tell if the channel is ready than the one I described?

### DIFF
--- a/doc/lightning-fundchannel.7.txt
+++ b/doc/lightning-fundchannel.7.txt
@@ -1,0 +1,56 @@
+LIGHTNING-FUNDCHANNEL(7)
+========================
+:doctype: manpage
+
+NAME
+----
+lightning-fundchannel - Command for establishing a lightning channel.
+
+SYNOPSIS
+--------
+*fundchannel* 'id' 'satoshi'
+
+DESCRIPTION
+-----------
+The *fundchannel* RPC command opens a payment channel with a peer by commiting
+a funding transaction to the blockchain as defined in BOLT #2.
+*fundchannel* by itself does not attempt to open a connection.
+A connection must first be established using *connect*.
+Once the transaction is confirmed, normal channel operations may begin.
+Readiness is indicated by *listpeers* reporting a 'state' of CHANNELD_NORMAL
+for the channel.
+
+'id' is the peer id obtained from *connect*.
+
+'satoshi' is the amount in satoshis taken from the internal wallet to fund the channel.
+The string 'all' can be used to specify all available funds.
+This value must be greater than the dust limit, currently set to 546.
+And it must be less than 1<<24 (approximately 0.16778 BTC).
+
+RETURN VALUE
+------------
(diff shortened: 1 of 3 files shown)


doc/lightning-fundchannel.7.txt
```diff
@@ -0,0 +1,54 @@
...
+
+SYNOPSIS
+--------
+*fundchannel* 'id' 'satoshi'
+
+DESCRIPTION
+-----------
+The *fundchannel* RPC command opens a payment channel with a peer by commiting
+a funding transaction to the blockchain as defined in BOLT #2.  Once the
+transaction is confirmed, normal channel operations may begin.  Readiness is
```
ZmnSCPxj: Suggest adding newlines at end of sentence.  That way, even if an earlier sentence is later amended, only that sentence needs to be reformatted and later sentences in the same paragraph need not be affected.  E.g.

    The *fundchannel* RPC command opens a payment channel with a peer by commiting
    a funding transaction to the blockchain as defined in BOLT #2.
    Once the transaction is confirmed, normal channel operations may begin.
    Readiness is indicated by *listfunds* reporting a `short_channel_id` for
    the channel.

Then later if we amend it:

    The *fundchannel* RPC is a totally radical command that opens a really cool payment
    channel with a peer by commiting a funding transaction to the blockchain as defined
    in the outrageously awesome BOLT #2.
    Once the transaction is confirmed, normal channel operations may begin.
    Readiness is indicated by *listfunds* reporting a `short_channel_id` for
    the channel.

doc/lightning-fundchannel.7.txt
```diff
@@ -0,0 +1,54 @@
...
+indicated by *listfunds* reporting a 'short_channel_id' for the channel.
+Channel status can then be monitored using *listchannels*.
```
ZmnSCPxj: `listchannels`, despite its name, lists all channels known in the public network, i.e. the entire routemap.  Local channels are listed in `listpeers`.

doc/lightning-fundchannel.7.txt
```diff
@@ -0,0 +1,54 @@
...
+'channel_id' of the newly created channel.  On failure, an error is reported
+and the channel is not funded.
+
+The following error codes may occur:
+
+* -1. Catchall nonspecific arror.
+
+The above error may include a descriptive message indicating:
+
+* The 'id' is invalid.
```
ZmnSCPxj: Yes, we should probably start an effort at some point to give unique error codes for some of the more complex commands. Since it is a JSON-RPC change this will probably have to be done pre-0.6 release too.  This is fine for now though; we can update the docs later when we finally have unique error codes.

cdecker:
Should probably also mention that an existing connection is required, and that `fundchannel` does not attempt to open a connection by itself.

rustyrussell:
Ack aa0d120

PR #1465, opened 2018-05-06
Properly close TCP sockets

ZmnSCPxj:
If not properly closed (as described in [this document](http://ia800504.us.archive.org/3/items/TheUltimateSo_lingerPageOrWhyIsMyTcpNotReliable/the-ultimate-so_linger-page-or-why-is-my-tcp-not-reliable.html)), the last message may not get sent when the socket is closed.  This affects `closingd` primarily, since we should send a final `closing_signed` before disconnecting.

The added `test_closing_torture` attempts to trigger the bug but even with 15 tries it often passes prior to the fix.  A higher value takes too long and may trigger the CI giving up on a VALGRIND run.

### DIFF
--- a/closingd/Makefile
+++ b/closingd/Makefile
@@ -60,6 +60,7 @@ CLOSINGD_COMMON_OBJS :=				\
 	common/permute_tx.o			\
 	common/ping.o				\
 	common/read_peer_msg.o			\
+	common/socket_close.o			\
 	common/status.o				\
 	common/status_wire.o			\
 	common/subdaemon.o			\
--- a/closingd/closing.c
+++ b/closingd/closing.c
@@ -9,6 +9,7 @@
 #include <common/peer_billboard.h>
 #include <common/peer_failed.h>
 #include <common/read_peer_msg.h>
+#include <common/socket_close.h>
 #include <common/status.h>
 #include <common/subdaemon.h>
 #include <common/type_to_string.h>
@@ -581,6 +582,11 @@ int main(int argc, char *argv[])
 		       offer[LOCAL]);
 
 	/* We're done! */
+	/* Properly close the channel first. */
+	if (!socket_close(PEER_FD))
+		status_unusual("Closing and draining peerfd gave error: %s",
+			       strerror(errno));
+	/* Sending the below will kill us! */
 	wire_sync_write(REQ_FD,	take(towire_closing_complete(NULL)));
 	tal_free(ctx);
 	daemon_shutdown();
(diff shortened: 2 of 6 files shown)


practicalswift:
Concept ACK!

Perhaps the `status_unusual`:s should be `status_trace` if they are unlikely to be of interest to the operator in the general case?

ZmnSCPxj:
Normal closing (none of the syscalls returned errors) reports `syslog_trace`.  Errors on the syscalls `shutdown`, `read`, and `close` are reported as `syslog_unusual`: it would be unusual for them to error.

1.  `shutdown` - errors if not a file descriptor (not gonna happen), the second argument is invalid (a fixed constant in the code and therefore pointing to a bug if this occurs), the socket was never connected (not gonna happen at this point, we were sending happily with the peer at this point in the code), or the fd was not a socket (even local UNIX sockets are sockets).
2.  `read` - errors if it is a non-blocking socket that would block (hmmm I should probably go check it, but it has not triggered during testing, and if it does, I should probably go make the socket blocking before doing this), not a file descriptor (not gonna happen), a signal was raised (we handle `EINTR` internally), the buffer is outside of the address space (it is allocated on the stack, so not gonna happen), or an I/O error happens (so we definitely should go report this event).
3. `close` - who even checks for errors from `close` these days? Well it errors if it is not a file descriptor (not gonna happen), a signal was raised (we handle `EINTR`), or an I/O error happens (which we definitely should go report).

practicalswift:
@ZmnSCPxj Makes sense! Let's keep it as-is :-)

ZmnSCPxj:
I wanted to be able to determine *which* syscall failed as there are three of them (`shutdown`, `read`, and `close`).  TCP code is unfamiliar to me so I decided to be err in the side of details. But if you judge we can safely throw away the details, I will modify.

ZmnSCPxj:
Rebase and simplified, ready for review @rustyrussell

practicalswift:
Will normal closing still report using `syslog_trace`?

ZmnSCPxj:
Not anymore, no.

rustyrussell:
Rebase conflict with the new port stuff, trivial to fix up, so I've rebased and force-pushed.  Will apply once travis happy.

rustyrussell:
Ack 75179e44996667dabe8c3a2b04e5f638be478f0f

ZmnSCPxj:
Rebased on top of cdecker recent PR, which should clean up the test regression.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/1465/commits/4426d9334709627e2414bc4ac54068ce7a1173b4

PR #1468, opened 2018-05-06
Forget channel if we are fundee and it has not confirmed yet for a long time

ZmnSCPxj:
As per BOLT spec:

> A non-funding node (fundee):
>   - SHOULD forget the channel if it does not see the
> funding transaction after a reasonable timeout.

We use blocks as unit of timeout.  We forget the channel, if we are the fundee, if it has been 2016 blocks (`--max-funding-unconfirmed-blocks`, two weeks) since the channel was opened, and we have not seen it locked in yet.

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -16,6 +16,7 @@
 #include <common/timeout.h>
 #include <common/utils.h>
 #include <inttypes.h>
+#include <lightningd/channel_control.h>
 #include <lightningd/gossip_control.h>
 
 /* Mutual recursion via timer. */
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -1,6 +1,10 @@
+#include <bitcoin/pubkey.h>
 #include <bitcoin/script.h>
 #include <ccan/fdpass/fdpass.h>
 #include <channeld/gen_channel_wire.h>
+#include <common/memleak.h>
+#include <common/timeout.h>
+#include <common/utils.h>
 #include <errno.h>
 #include <gossipd/gossip_constants.h>
 #include <hsmd/capabilities.h>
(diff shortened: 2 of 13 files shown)


cdecker:
We have a `fake-bitcoin-cli` that could be used to selectively fail a `sendrawtransaction`.

ZmnSCPxj:
Thank you, will go make the test.

ZmnSCPxj:
Forgetting the channel has the side effect that we disconnect from the peer; is that desirable?

cdecker:
Probably yes, since we don't really tell the other side about just having a lobotomy, and they'll expect us to know what they're talking about. By disconnecting (and potentially reconnecting) we can tell them to go away :wink:

ZmnSCPxj:
Rebased, should fix the test regressions. Ready for review.

lightningd/channel_control.c
```diff
@@ -322,3 +326,100 @@ bool channel_tell_funding_locked(struct lightningd *ld,
...
+		    "loss of funds.",
+		    block_height - channel->first_blocknum,
+		    type_to_string(tmpctx, struct bitcoin_txid,
+				   &channel->funding_txid));
+	/* Forget it later, to prevent loops from crashing.
+	 * Allocate timer from channel so if multiple deletions
+	 * by other checks that get added in the future, only
+	 * first timer triggers. */
+	notleak(new_reltimer(&ld->timers, channel, time_from_sec(0),
+			     &delete_channel, channel));
```
rustyrussell: Hanging off a timer like this is always a hack.   What happens without this?

If it needs to be outside loop, I would change this to a simple "is_fundee_timed_out", then do:

```
const tal_t *expired = tal(NULL, char);

list_for_each(peer) {
    list_for_each(channel) {
        if (is_fundee_timed_out(channel, blockheight))
            tal_steal(expired, channel);
    }
}

/* Now we free them all at once */
tal_free(expired);
```

lightningd/chaintopology.c
```diff
@@ -341,6 +342,8 @@ static void updates_complete(struct chain_topology *topo)
 	if (topo->tip != topo->prev_tip) {
 		/* Tell lightningd about new block. */
 		notify_new_block(topo->bitcoind->ld, topo->tip->height);
+		channel_notify_new_block(topo->bitcoind->ld,
+					 topo->tip->height);
```
rustyrussell: Minor cleanup:

I'd put the call to notify_new_block() in lightningd.c, rename the current one to (say) htlcs_notify_new_block() and have notify_new_block() call htlcs_notify_new_block and channel_notify_new_block.

it's a bit cleaner to have it outside chaintopology I think...

lightningd/options.c
```diff
@@ -398,6 +398,12 @@ static void config_register_opts(struct lightningd *ld)
 			 opt_set_u64, opt_show_u64,
 			 &ld->ini_autocleaninvoice_cycle,
 			 "If expired invoice autoclean enabled, invoices that have expired for at least this given seconds are cleaned");
+	opt_register_arg("--max-funding-unconfirmed-blocks",
+			 opt_set_u32, opt_show_u32,
+			 &ld->max_funding_unconfirmed,
+			 "Maximum number of blocks we wait for a channel "
+			 "funding transaction to confirm, if we are the "
+			 "fundee.");
 }
```
rustyrussell: I've been removing options like this in my documentation patch; I think this should just be a constant in a header somewhere.  I can't see why anyone would want to change it, unless we want to change it for everyone.

lightningd/channel_control.c
ZmnSCPxj: Ha, yes, that is indeed better.

ZmnSCPxj: Hmm, maybe not safe?  `delete_channel` does more than just `tal_free`, it also deletes the channel from the DB.  However a side effect of `delete_channel` is to also do a `tal_free` of the channel.

How about putting it in an array then having a separate loop to destroy the array?

    struct channel **expired = tal_arr(NULL, struct channel *, 0);

    list_for_each(peer) {
        list_for_each(channel) {
            if (is_fundee_should_forget(ld, channel, block_height)) {
                tal_resize(&expired, tal_count(expired) + 1);
                expired[tal_count(expired) - 1] = channel;
            }
        }
    }
    for (size_t i = 0; i < tal_count(expired); ++i)
        delete_channel(expired[i]);
    tal_free(expired);

lightningd/options.c
ZmnSCPxj: I kept it since it is useful in the test to lower the threshold.  Or should I make it into a `dev-` option instead?

lightningd/channel_control.c
```diff
@@ -322,3 +326,100 @@ bool channel_tell_funding_locked(struct lightningd *ld,
...
+	 *   - SHOULD forget the channel if it does not see the
+	 * funding transaction after a reasonable timeout.
+	 */
+
+	/* Only applies if we are fundee. */
+	if (channel->funder == LOCAL)
+		return false;
+
+	/* Does not apply if we already saw the funding tx. */
+	/* FIXME: This is only set when the funding tx is confirmed
```
cdecker: Don't think this is a huge problem, and I think seeing it in the mempool could still hold us up indefinitely due to low fees. Happy to keep this as is. `scid` is set after the first confirmation.

lightningd/options.c
cdecker: Making this a `dev-` option would probably be sensible, yes :-)

ZmnSCPxj:
`s/--max-funding-unconfirmed-blocks/--dev-max-funding-unconfirmed-blocks/`

cdecker:
ACK 6389477

PR #1472, opened 2018-05-07
payalgo: Report reason to delay before clearing try memory.

ZmnSCPxj:
We allocate the reason to delay, if any, from the `pay->try_parent`.
So we should not clear the `pay->try_parent` until after we print
the reason.


practicalswift:
utACK 65083bb66477f389ea8cd9c58e955c47e5ced2a8

Nice find

PR #1476, opened 2018-05-08
check `wireaddr_to_sockname` return value

isghe:
Accessory to PR https://github.com/ElementsProject/lightning/pull/1477


cdecker:
I have to agree with @ZmnSCPxj here, sorry @isghe :slightly_frowning_face:

PR #1481, opened 2018-05-09
lightning-cli: clean up -H a little more, don't suppress `help listpeeers` etc

rustyrussell:
No wonder nobody was filling these in!

    $ cli/lightning-cli help listpeers
    HELP! Please contribute a description for this command!

And:

    $ cli/lightning-cli help help
    help [command]
    Without [command]:
      Outputs an array of objects with 'command' and 'description'
    With [command]:
      Give a single object containing 'verbose', which completely describes
      the command inputs and outputs.

This is a subtle attempt to encourage @isghe as in #1476 :)

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -41,16 +41,17 @@ static void tal_freefn(void *ptr)
 	tal_free(ptr);
 }
 
-struct netaddr;
-
 /* Returns number of tokens digested */
-static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
+static size_t human_readable(const char *buffer,
+			     const jsmntok_t *t,
+			     const char *prefix)
 {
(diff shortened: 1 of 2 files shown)


wythe:
Unfortunately this also takes us back to having the basic `lightning-cli help`, intended for humans, to be more unreadable than `lightning-cli -J help`. and that just doesn't make sense.

Idea:  Remove the `-J` and `-H` options altogether; default to json for everything except `help`, and have `help` behave more like `git`:

`lightning-cli help` : brings up human readable overview of all commands
`lightning-cli help <cmd>` : brings up man page on systems that support it, otherwise a detailed description as currently defined.

rustyrussell:
I actually after agree with @wythe here. I'll rework this so help is still nicely formatted, and we'll plan for man integration sometime later.

PR #1488, opened 2018-05-10
Tor/socks cleanups and hostname support.

rustyrussell:
How's this @Saibato ?  Works for me; basically after all the cleanups the actual patch to implement it is pretty short.

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -270,7 +270,8 @@ static bool separate_address_and_port(const tal_t *ctx, const char *arg,
 }
 
 bool wireaddr_from_hostname(struct wireaddr *addr, const char *hostname,
-			    const u16 port, bool dns_ok, const char **err_msg)
+			    const u16 port, bool *no_dns,
+			    const char **err_msg)
 {
 	struct sockaddr_in6 *sa6;
 	struct sockaddr_in *sa4;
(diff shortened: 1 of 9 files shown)


gossipd/gossip.c
```diff
@@ -1939,10 +1939,27 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
...
+	switch (reach->addr.itype) {
+	case ADDR_INTERNAL_WIREADDR:
+		host = fmt_wireaddr_without_port(tmpctx,
+						 &reach->addr.u.wireaddr);
+		port = reach->addr.u.wireaddr.port;
+		break;
+	case ADDR_INTERNAL_SOCKNAME:
+	case ADDR_INTERNAL_ALLPROTO:
+	case ADDR_INTERNAL_AUTOTOR:
+		break;
```
ZmnSCPxj: `abort();` instead?

ZmnSCPxj:
ACK 8f39adaf6f2cf36e919eed299dab2c55d4f600b5

common/wireaddr.c
```diff
@@ -299,13 +303,19 @@ bool wireaddr_from_hostname(struct wireaddr *addr, const char *hostname,
...
 
+	/* Tell them we wanted DNS and fail. */
+	if (no_dns) {
+		if (err_msg)
+			*err_msg = "Needed DNS, but lookups suppressed";
+		*no_dns = true;
+		return false;
+	}
+
 	memset(&hints, 0, sizeof(hints));
```
Saibato: Thanks! for doing this, confidence in that we care will give us later the feedback ln needs

gossipd/gossip.c
rustyrussell: The next line did that and more, by catching any invalid values and calling status_failed...

PR #1490, opened 2018-05-10
Unannounced channel fix

rustyrussell:
... and a tweak I've had sitting around for ages which helps speed my debug cycles :)

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -195,7 +195,6 @@ static void init_half_chan(struct routing_state *rstate,
 
 	c->channel_update = NULL;
 	c->unroutable_until = 0;
-	c->active = false;
 	c->flags = idx;
 	/* We haven't seen channel_update: make it halfway to prune time,
 	 * which should be older than any update we'd see. */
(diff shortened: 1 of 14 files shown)


gossipd/routing.c
```diff
@@ -948,6 +960,15 @@ bool routing_add_channel_update(struct routing_state *rstate,
 			      (flags & ROUTING_FLAGS_DISABLED) == 0, timestamp,
 			      htlc_minimum_msat);
 
+	/* For private channels, we get updates without an announce: don't
```
ZmnSCPxj: A little undesirable for #1330: how will `invoice` query feerates for private channels?

ZmnSCPxj: It could be an advantage though. We could have a `gossip_get_private_routes` message, which just iterates over the local channels and checks for `private_update` field.

rustyrussell:
Actually, the broadcast owning the messages is annoying.  We should ahve the messages owning the broadcast tags.  Patch coming...

gossipd/routing.h
```diff
@@ -37,6 +37,9 @@ struct half_chan {
 	/* Cached `channel_update` we might forward to new peers (or 0) */
 	u64 channel_update_msgidx;
 
+	/* If it's a private update, it's not in the broadcast map. */
+	const u8 *private_update;
```
cdecker: Any reason we keep the `channel_update` around at all? It's not like we'll be forwarding it anyway

PR #1492, opened 2018-05-11
fee_proportional_millionths unit is `satoshi`.

nayuta-ueno:
change unit from `msat` to `sat`.

```
MUST set fee_proportional_millionths to the amount (in millionths of a satoshi) it will charge per transferred satoshi.
```


nayuta-ueno:
not need change unit.
close.

PR #1495, opened 2018-05-11
Separate counter variable declaration from for loop statements.

bitonic-cjp:
This makes the code conform to (an older version of?) the C standard.
The reason I propose this change is because I wanted to analyze the (rather large) code base with ncc, but it failed to recognize constructions like for(int i=0; ...).

### DIFF
--- a/bitcoin/chainparams.c
+++ b/bitcoin/chainparams.c
@@ -58,7 +58,9 @@ const struct chainparams networks[] = {
 
 const struct chainparams *chainparams_for_network(const char *network_name)
 {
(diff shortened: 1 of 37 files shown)


practicalswift:
Thanks for letting `c-lightning` developers use `ncc`!

I don't think we should merge this into `master`: generally we we want to minimize the scope of all variables. When a larger scope than necessary is used, code becomes less readable, harder to maintain, and more likely to reference unintended variables.

But please keep your fork/branch around so that `c-lightning` developers can use it in order to get `ncc` to run :-)

rustyrussell:
Yeah, I think this is a local mod.  local vars in for loops is one part of C99 I really like.

PR #1498, opened 2018-05-13
Do not use potentially uninitialized value needed_dns

practicalswift:
Do not use potentially uninitialized value `needed_dns`.

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -299,7 +299,7 @@ static char *opt_set_offline(struct lightningd *ld)
 
 static char *opt_add_proxy_addr(const char *arg, struct lightningd *ld)
 {
-	bool needed_dns;
+	bool needed_dns = false;
 	tal_free(ld->proxyaddr);
 
 	/* We use a tal_arr here, so we can marshal it to gossipd */


lightningd/options.c
```diff
@@ -308,8 +308,8 @@ static char *opt_add_proxy_addr(const char *arg, struct lightningd *ld)
 	if (!parse_wireaddr(arg, ld->proxyaddr, 9050,
 			    ld->use_proxy_always ? &needed_dns : NULL,
 			    NULL)) {
-		return tal_fmt(NULL, "Unable to parse Tor proxy address '%s' %s",
-			       arg, needed_dns ? " (needed dns)" : "");
+		return tal_fmt(NULL, "Unable to parse Tor proxy address '%s'",
+			       arg);
```
ZmnSCPxj: Maybe better: `(ld->use_proxy_always && needed_dns) ? " (needed dns)" : ""`?

rustyrussell: In this case, it's clearer to simply initialize it to 'false' I think, since it's ONLY for this msg.

ZmnSCPxj:
ACK 402c6e12088dd3530e1401749d4cc7c4597dc58e

PR #1505, opened 2018-05-15
features fixes

rustyrussell:
We had the endianness wrong, and it turns out writing even simple tests is useful :(


ZmnSCPxj:
ACK f1736db657d157e397a4e5009393ca57ab2f54a3

PR #1511, opened 2018-05-17
tests: fix to work with bitcoind master branch.

rustyrussell:
Now fixed to work with 0.16.1 and master as well.   Obviously the first change is needed before release, but I'd like the rest too, so we don't get weird test breakage when 0.16.2 is released.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -15,7 +15,6 @@ BITCOIND_CONFIG = {
     "regtest": 1,
     "rpcuser": "rpcuser",
     "rpcpassword": "rpcpass",
-    "rpcport": 18332,
 }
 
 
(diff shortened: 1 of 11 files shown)


rustyrussell:
This fix seems to break *old* bitcoinds!  Let me try harder...

ZmnSCPxj:
looks OK to me, but will wait for something which does not break *old* bitcoinds

tests/utils.py
```diff
@@ -236,7 +238,8 @@ def __init__(self, bitcoin_dir="/tmp/bitcoind-test", rpcport=None):
         ]
         BITCOIND_CONFIG['rpcport'] = rpcport
         btc_conf_file = os.path.join(regtestdir, 'bitcoin.conf')
-        write_config(os.path.join(bitcoin_dir, 'bitcoin.conf'), BITCOIND_CONFIG)
+        write_config(os.path.join(bitcoin_dir, 'bitcoin.conf'), BITCOIND_CONFIG,
+                     '[regtest]')
```
cdecker: You probably want to move the `[` `]` into `write_config`

jsarenik:
@ZmnSCPxj The solution is simple, but I don't know how to write it in Python.

It would be something like

```diff
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -38,9 +38,13 @@ def wait_for(success, timeout=TIMEOUT, interval=0.1):
         raise ValueError("Error waiting for {}", success)
 
 
 def write_config(filename, opts, section=None):
     with open(filename, 'w') as f:
-        if section:
+        if section_version and section:
```

Now just how to set `section_version` to True when `getnetworkinfo()` RPC call returns `version >= 169900`.

rustyrussell:
Set milestone 0.6, since this happens with bitcoind 0.16.1 :(

cdecker:
ACK a7a7cc9

PR #1513, opened 2018-05-17
Makefile: Propagate `NO_VALGRIND` to `py.test` invocation.

jsarenik:
Previously, by uncommenting NO_VALGRIND in Makefile, the variable
was not propagated and the children processes would still try to
use Valgrind.
    
Tested on Alpine Linux:

    make NO_VALGRIND=1 check

Also works after uncommenting both related lines in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ ifndef PYTEST
 	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt'"
 	exit 1
 else
-	PYTHONPATH=contrib/pylightning:$$PYTHONPATH TEST_DEBUG=1 DEVELOPER=$(DEVELOPER) $(PYTEST) tests/ $(PYTEST_OPTS)
+	PYTHONPATH=contrib/pylightning:$$PYTHONPATH TEST_DEBUG=1 DEVELOPER=$(DEVELOPER) NO_VALGRIND=$(NO_VALGRIND) $(PYTEST) tests/ $(PYTEST_OPTS)
 endif
 
 # Keep includes in alpha order.


ZmnSCPxj:
Requesting review from @rustyrussell : I am uncertain if `.EXPORT_ALL_VARIABLES` will have negative effects on the build, and there may be other ways to fix this.

jsarenik:
@ZmnSCPxj The change may be simplified in a way that it would just note in the comment in Makefile that setting `NO_VALGRIND` in Makefile would actually not help because the variable will not be exported and the Python tests will not get it (i.e. [line checking for `NO_VALGRIND`](https://github.com/ElementsProject/lightning/blob/90dc5173d5654eb605a573c98593f2ffe5b1505a/tests/test_lightningd.py#L31) environment variable in `tests/test_lightningd.py` would not do the expected thing).

ZmnSCPxj:
Perhaps better to change the invocation in pytest, then:

    -        PYTHONPATH=contrib/pylightning:$$PYTHONPATH TEST_DEBUG=1 DEVELOPER=$(DEVELOPER) $(PYTEST) tests/ $(PYTEST_OPTS)
    +        PYTHONPATH=contrib/pylightning:$$PYTHONPATH TEST_DEBUG=1 DEVELOPER=$(DEVELOPER) NO_VALGRIND=$(NO_VALGRIND) $(PYTEST) tests/ $(PYTEST_OPTS)

jsarenik:
@ZmnSCPxj  Thanks much! Great idea and it simplifies the problem. Patch changed accordingly.

ZmnSCPxj:
ACK 4a7f2a932964471da967bbb88a99a47039297408

rustyrussell:
Ok, this added to a number of issues about building, and prompted me to get moving on a real configure step. I'll post a WIP for review soon...

PR #1517, opened 2018-05-21
Fix channel_update timestamp races

rustyrussell:
Rather than having gossipd and channeld both generate signed updates, have gossipd always do it.


ZmnSCPxj:
ACK 79fd0815b40814b7b08d48a02aa444b146cb9767

PR #1518, opened 2018-05-21
Do not call strlen for every character in the log entry

NicolasDorier:
Low hanging fruit taking quite of CPU.

image

(was log level debug)

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -255,8 +255,10 @@ void logv(struct log *log, enum log_level level, const char *fmt, va_list ap)
 
 	l->log = tal_vfmt(l, fmt, ap);
 
+	size_t log_len = strlen(l->log);
+
 	/* Sanitize any non-printable characters, and replace with '?' */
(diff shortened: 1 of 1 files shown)


cdecker:
Shouldn't the compiler have optimized that away?

NicolasDorier:
Should be, but we are not currently compiling with the best optimized gcc settings I think.  Like `-g` for the symbol/stack traces might disable some optimizations I guess.

NicolasDorier:
will run a flamegraph before after.

NicolasDorier:
It does not optimize... After:

image

ZmnSCPxj:
ACK 5601a9ffbcb7fa93c892467dbc70d9c8250b30aa

Compiler "should" optimize it away but the code cost is reasonably low.

NicolasDorier:
@ZmnSCPxj well here it is clear it does not optimize it. (I guess there is not proper compiler flags for it yet)

PR #1519, opened 2018-05-22
doc/HACKING.md: remove note about DEVELOPER mode being default

jsarenik:
The DEVELOPER mode is not the default since 00e75fee,
change the documentation accordingly.


ZmnSCPxj:
ACK 071a8c27351b0eb8f959baba573821d9a73875ab

Issue #1231, opened 2018-03-16
FreeBSD (tcsh) build problem due to HOST environment variable


laanwj:
On master commit 4ffd5e21 (just checked out) I'm unable to build on FreeBSD ~~- with 73cda2f2 it was working~~. Output:
```
if git submodule status external/libwally-core | grep -q '^[-+]'; then rm -rf external/libwally-core; git submodule update --init external/libwally-core; fi
cd external/libwally-core && ./tools/autogen.sh && ./configure CC="cc" --enable-static=yes --host="medea" --build="" --enable-module-recovery --enable-shared=no --prefix=/ --libdir=/ && gmake
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
libtoolize: copying file 'build-aux/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'build-aux/m4'.
libtoolize: copying file 'build-aux/m4/libtool.m4'
libtoolize: copying file 'build-aux/m4/ltoptions.m4'
libtoolize: copying file 'build-aux/m4/ltsugar.m4'
libtoolize: copying file 'build-aux/m4/ltversion.m4'
libtoolize: copying file 'build-aux/m4/lt~obsolete.m4'
configure.ac:10: installing 'build-aux/compile'
configure.ac:9: installing 'build-aux/missing'
Makefile.am: installing 'build-aux/depcomp'
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'tools/build-aux'.
libtoolize: copying file 'tools/build-aux/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'tools/build-aux/m4'.
libtoolize: copying file 'tools/build-aux/m4/libtool.m4'
libtoolize: copying file 'tools/build-aux/m4/ltoptions.m4'
libtoolize: copying file 'tools/build-aux/m4/ltsugar.m4'
libtoolize: copying file 'tools/build-aux/m4/ltversion.m4'
libtoolize: copying file 'tools/build-aux/m4/lt~obsolete.m4'
configure.ac:28: installing 'tools/build-aux/compile'
configure.ac:27: installing 'tools/build-aux/missing'
src/Makefile.am: installing 'tools/build-aux/depcomp'
checking build system type... amd64-unknown-freebsd11.1
checking host system type... Invalid configuration `medea': machine `medea' not recognized
[...]

PR #1520, opened 2018-05-22
external/Makefile: Change HOST to MAKE_HOST

jsarenik:
Proposed fix for issue #1231 - FreeBSD (tcsh) build problem due
to HOST environment variable.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -233,7 +233,7 @@ This will allow you to properly configure
 the build for the target device environment.
 Build with:
 
-    BUILD=x86_64 HOST=arm-linux-androideabi \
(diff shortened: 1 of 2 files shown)


jsarenik:
@ZmnSCPxj thank you for feed-back! I changed it appropriately.

ZmnSCPxj:
ACK af804ef5846eccfbd3195c34f26133a851bc1870

jsarenik:
@ZmnSCPxj `MAKE_HOST` is an internal variable of GNU Make. It may work well for others who use autotools, but did not translate well for `lightning` which is Makefile-only + also submodules which use Autotools themselves

jsarenik:
Ref: #1533

PR #1523, opened 2018-05-24
Modify comments about the precision

conanoc:
Just a modification of comments.

Rationale
1. The loss of precision will be like feepercent of 3.0000..(some dots)..1 % vs 3.0 %.
2. The loss of 0.0000..(some dots)..1 above is not representable in double.

I mean, the loss of precision in msatoshi might be significant but the loss of precision in feepercent is not significant (Or there is no loss of precision in feepercent, I'm not sure).


ZmnSCPxj:
Please use the rationale in your above comment in the in-code comment instead?

conanoc:
How about this?

ZmnSCPxj:
ACK f851531

Issue #1464, opened 2018-05-06
Proper JSON-RPC error codes


ZmnSCPxj:
Currently almost all commands just give a `-1` error code in all errors.  However this is not really particularly useful programmatically: it would be better for progammatic access to be able to determine what the exact error is without having to parse the error message (which is in fact the intent of the error code).

So I want to propose that we change `command_fail` to accept a proper code argument and make an effort to provide error codes for all error types!

IF we decide to go this route (this issue is a `question`) then w should probably schedule this for 0.6 also, as we would like to present a reasonably stable JSON-RPC interface.

Inspired-by: @wythe #1463

wythe:
Other parts of the code use `command_failed_detailed()` to report the error with a code. We could just incrementally convert `command_fail` over to that.  I can get to work with `withdraw` and `fundchannel`  if we want to do this.

ZmnSCPxj:
`command_fail_detailed` includes a JSON `data` field, which is optional for JSON-RPC errors.  In many cases the `data` field is not needed.  However JSON-RPC has the error codes mandatory, so we really should require `command_fail` to accept the error code as argument instead of always using `-1`.

ZmnSCPxj:
A lot of errors are also related to parameters being wrong, which we should really have error code `-32602`, also `#define`d in `lightningd/jsonrpc_errors.h` as `JSONRPC2_INVALID_PARAMS`.  We should have a simple `command_fail_param` that accepts the message, gives `JSONRPC2_INVALID_PARAMS` standard error code, and outputs a message like `"msatoshi": Must be a positive integer: "string"`

wythe:
Yes, and in that case, maybe we could enhance `json_get_params` to take an additional "token_type" parameter for each token, AND a place to store the value upon success.  

Just looking through a few of the json handlers, here's some possible values for the enum:

    enum token_type {
        bool_type,
        msatoshi,
        positive_integer,
        sha256,
        json_array,
        bolt11,
        short_channel_id,
    };

And then json_get_params would call, for example, `json_tok_bool` in the case of bool_type for us
automatically.

Well that's just a thought.  Easier said than done.

ZmnSCPxj:
It would not be very general, though; we would have to keep extending this.  Consider `newaddr`, for example, which has an argument that is the address type, and is the only command which has this kind of argument.   `json_get_params` also returns the `JSONRPC2_INVALID_PARAMS` error already.

wythe:
[...]

PR #1524, opened 2018-05-24
Added error code parameter to command_fail

wythe:
Until now, `command_fail()` reported an error code of -1 for all uses.
This PR adds an `int code` parameter to `command_fail()`, requiring the
caller to explicitly include the error code.

This is part of #1464.

The majority of the calls are used during parameter validation and
their error code is now JSONRPC2_INVALID_PARAMS.

The rest of the calls report an error code of LIGHTNINGD, which I defined to
-1 in `jsonrpc_errors.h`.  The intention here is that as we improve our error
reporting all occurenaces of LIGHTNINGD will go away and we can eventually
remove it.

I also converted calls to `command_fail_detailed()` that took a `NULL` `data`
parameter to use the new `command_fail()`.

The only difference from an end user perspecive is that bad input errors that
used to be -1 will now be -32602 (JSONRPC2_INVALID_PARAMS).

### DIFF
--- a/common/wallet_tx.c
+++ b/common/wallet_tx.c
@@ -22,7 +22,8 @@ bool wtx_select_utxos(struct wallet_tx * tx, u32 fee_rate_per_kw,
 					      &tx->amount,
 					      &fee_estimate);
 		if (!tx->utxos || tx->amount < 546) {
-			command_fail(tx->cmd, "Cannot afford fee %"PRIu64,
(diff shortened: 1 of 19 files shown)


rustyrussell:
Ack 2daa80e951f2e713dc73a1d70f0a2d43f47245be

This is great work. I would have called LIGHTNINGD something else, but that's nitpicking since it's temporary.

ZmnSCPxj:
ACK 2daa80e

PR #1528, opened 2018-05-28
Add CRC32 checksums to gossip_store

### LINKED ISSUE #1436: Add sha256 to gossip store
Since it's a best-effort store, it'd be nice (and cheap) to add a SHA256 to each entry.  It's probably overkill, since the unmarshalling code will probably catch it, but it's easy.

cdecker:
Using CRC32 instead of SHA256 due to better performance and on-disk space.

Fixes #1436


ZmnSCPxj:
ACK e56f6edd0fcd1c98dfcb188782009ac0c53a4506

rustyrussell:
It's tempting to suggest writev and readv for the length and checksum (or just a header structure) but that's probably premature optimization. This is nice and simple!

cdecker:
I did not know about `readv` and `writev`, they look interesting though :-)

PR #1531, opened 2018-05-30
Stabilize a number of flaky tests and fix some gossip semantics while we're at it

### LINKED ISSUE #1515: Travis should test in parallel
Our tests run much faster in parallel, particularly due to polling delays, etc.  I run -n10 on my laptop (without valgrind) with no problems.

That would allow us to re-enable some non-developer tests, for example.

cdecker:
The flaky pruning tests were really annoying, so I took a closer look at them. It turns out that we were relying on a `channel_update` being sent before actually closing the connection, otherwise we were not disabling both directions.

This now explicitly disables both directions on close, and on disconnect. While not a big issue before, with the `r` fields being merged soon, we could end up sending a disabled channel as a `contact_point` if we didn't disable them.

Other changes include better handling of concurrent opens and watching to logs for the wrong signals, and filtering out some output (looking at you Fee Estimator)

Fixes #1515 since I now feel much better about the test stability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -367,7 +367,7 @@ update-mocks/%: %
 	@tools/update-mocks.sh "$*"
 
 unittest/%: %
-	$(VALGRIND) $(VALGRIND_TEST_ARGS) $*
+	$(VALGRIND) $(VALGRIND_TEST_ARGS) $* > /dev/null
 
 # Installation directories
 prefix = /usr/local
--- a/gossipd/gossip.c
+++ b/gossipd/gossip.c
@@ -2591,6 +2591,33 @@ static struct io_plan *handle_outpoint_spent(struct io_conn *conn,
 	return daemon_conn_read_next(conn, &daemon->master);
 }
 
+/**
+ * Disable both directions of a channel due to an imminent close.
+ *
+ * We'll leave it to handle_outpoint_spent to delete the channel from our view
+ * once the close gets confirmed. This avoids having strange states in which the
+ * channel is list in our peer list but won't be returned when listing public
+ * channels. This does not send out updates since that's triggered by the peer
+ * connection closing.
+ */
+static struct io_plan *handle_local_channel_close(struct io_conn *conn,
+						  struct daemon *daemon,
+						  const u8 *msg)
+{
+	struct short_channel_id scid;
+	struct chan *chan;
+	struct routing_state *rstate = daemon->rstate;
+	if (!fromwire_gossip_local_channel_close(msg, &scid))
+		master_badmsg(WIRE_GOSSIP_ROUTING_FAILURE, msg);
+
+	chan = get_channel(rstate, &scid);
+	if (chan) {
+		chan->half[0].flags |= ROUTING_FLAGS_DISABLED;
+		chan->half[1].flags |= ROUTING_FLAGS_DISABLED;
+	}
+	return daemon_conn_read_next(conn, &daemon->master);
+}
+
 static struct io_plan *recv_req(struct io_conn *conn, struct daemon_conn *master)
 {
 	struct daemon *daemon = container_of(master, struct daemon, master);
@@ -2656,6 +2683,9 @@ static struct io_plan *recv_req(struct io_conn *conn, struct daemon_conn *master
 
 	case WIRE_GOSSIP_OUTPOINT_SPENT:
 		return handle_outpoint_spent(conn, daemon, master->msg_in);
+
+	case WIRE_GOSSIP_LOCAL_CHANNEL_CLOSE:
+		return handle_local_channel_close(conn, daemon, master->msg_in);
 
 	/* We send these, we don't receive them */
 	case WIRE_GOSSIPCTL_ACTIVATE_REPLY:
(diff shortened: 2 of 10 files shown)


lightningd/closing_control.c
```diff
@@ -213,4 +213,11 @@ void peer_start_closingd(struct channel *channel,
 	/* We don't expect a response: it will give us feedback on
 	 * signatures sent and received, then closing_complete. */
 	subd_send_msg(channel->owner, take(initmsg));
+
+	/* Now tell gossipd that we're closing and that neither direction should
+	 * be used. */
+	if (channel->scid)
+		subd_send_msg(channel->peer->ld->gossip,
+			      take(towire_gossip_local_channel_close(
+				  tmpctx, channel->scid)));
```
rustyrussell: Minor nit: take() should be generally done from something off NULL if it's transitory; no need to use tmpctx here.  That means that if take() doesn't work for some reason we see the leak.

gossipd/gossip.c
```diff
@@ -2496,8 +2496,9 @@ static struct io_plan *handle_disable_channel(struct io_conn *conn,
 	if (timestamp <= hc->last_timestamp)
 		timestamp = hc->last_timestamp + 1;
 
-	/* Active is bit 1 << 1, mask and apply */
-	flags = (0xFFFD & flags) | (!active << 1);
+	flags = direction;
+	if (!active)
+		flags |= ROUTING_FLAGS_DISABLED;
```
rustyrussell: How does this "fix" it?  It's neater, sure, but it should give the same effect?

gossipd/gossip.c
```diff
@@ -2355,11 +2355,99 @@ static struct io_plan *peer_important(struct io_conn *conn,
...
+	}
+
+	err = handle_channel_update(rstate, msg, "disable_channel");
+	if (err)
+		status_failed(STATUS_FAIL_INTERNAL_ERROR,
+			      "rejected disabling channel_update: %s",
+			      tal_hex(tmpctx, err));
+}
+
+static void peer_disable_channels(struct routing_state *rstate, struct node *node)
```
rustyrussell:
~~~This function *actually* disables 0 or 1 channel,, between us and this peer.   It does so redundantly, since the callers already set the DISABLED flag.  The 'gossip_disable_channel' also assumes we're one end of the channel (and doesn't assert it).~~~

~~~Perhaps neater to call this "disable_channel" and have it set both disabled flags, then if one side is us, send an update.~~~ (Cleaned up in later commit, great work!)

I prefer not to touch the flags word if the halfchan is not defined, either, though valgrind won't complain.

gossipd/gossip.c
```diff
@@ -2375,6 +2463,11 @@ static struct io_plan *peer_disconnected(struct io_conn *conn,
 	status_trace("Forgetting remote peer %s",
 		     type_to_string(tmpctx, struct pubkey, &peer->id));
 
+	/* Disable any channels to and from this peer */
```
rustyrussell: And we should not do this if we this is caused by a reconnecting peer (below in this function).  Though I have a patch WIP which sets up a timer for every channel_update so we don't spam updates, which would also cover this.

gossipd/gossip.c
```diff
@@ -2394,6 +2394,11 @@ static void gossip_disable_channel(struct routing_state *rstate, struct chan *ch
 		    "Unable to parse previously accepted channel_update");
 	}
 
+	/* Avoid sending gratuitous disable messages, e.g., on close and
+	 * subsequent disconnect */
+	if (flags & ROUTING_FLAGS_DISABLED)
+		return;
+
```
rustyrussell: I have the same patch in my queue, though I genericised it (so any "no-change" update gets suppressed except keepalives), but this is still a good change.

gossipd/gossip.c
```diff
@@ -1900,6 +1900,26 @@ static void gossip_disable_channel(struct routing_state *rstate, struct chan *ch
...
+	    get_node(daemon->rstate, &daemon->rstate->local_id);
+	struct chan *c;
+	size_t i;
+
+	/* We don't have a local_node, so we don't have any channels yet
+	 * either */
+	if (!local_node)
+		return;
+
+	for (i=0; i<tal_count(local_node->chans); i++) {
```
rustyrussell: Mildly prefer `for (size_t i` these days; also `c` can go inside loop.

tests/utils.py
```diff
@@ -338,12 +338,22 @@ def __init__(self, daemon, rpc, btc, executor, may_fail=False, may_reconnect=Fal
...
+
+        if confirm or announce:
+            self.bitcoin.generate_block(1)
+
+        if announce:
+            self.bitcoin.generate_block(5)
+
+        if confirm or announce:
+            self.daemon.wait_for_log(
+                r'Funding tx {} depth'.format(fundingtx['txid']))
```
rustyrussell: But we don't actually wait for announce?  Should we do something like:

```python
if announce:
    blocks = 6
elif confirm:
    blocks = 1
else
    blocks = None

if blocks:
    bitcoind.generate_block(blocks)
    self.daemon.wait_for_log('Funding tx {} depth {}'.format(fundingtx['txid'], blocks)
```

Makefile
```diff
@@ -367,7 +367,7 @@ update-mocks/%: %
 	@tools/update-mocks.sh "$*"
 
 unittest/%: %
-	$(VALGRIND) $(VALGRIND_TEST_ARGS) $*
+	$(VALGRIND) $(VALGRIND_TEST_ARGS) $* > /dev/null
```
rustyrussell: Mildly prefer redirecting to tmpfile and dumping the output if it fails, but that's more code...

PR #1533, opened 2018-05-31
external/Makefile: Change MAKE_HOST and BUILD to CROSS_*

jsarenik:
MAKE_HOST is internally defined by GNU Make. See
http://git.savannah.gnu.org/cgit/make.git/commit/?id=ef11217

Checked on Alpine Linux locally.

I am sorry for not checking the presence of MAKE_HOST variable
in my first commit. Thanks to @NicolasDorier for pointing it out.

Ref: #1318 #1231

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -233,7 +233,7 @@ This will allow you to properly configure
 the build for the target device environment.
 Build with:
 
-    BUILD=x86_64 MAKE_HOST=arm-linux-androideabi \
+    CROSS_BUILD=x86_64 CROSS_HOST=arm-linux-androideabi \
       make PIE=1 DEVELOPER=0 \
       CONFIGURATOR_CC="arm-linux-androideabi-clang -static"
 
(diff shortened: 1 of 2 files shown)


jsarenik:
@ZmnSCPxj see this

NicolasDorier:
ACK, this fix my issue

ZmnSCPxj:
I would prefer autotools, but I think @rustyrussell is intending to create a `configure` script for this.

This is OK to me if it fixes a problem in our `Makefile`. Cannot really check in detail as of now due to lack of access to my hacking computer.

utACK 708edebb00cbd4a55ed7953778eb7064373ba77f

jsarenik:
@ZmnSCPxj I see, but we are not there yet and all I wanted was to do a minimal fix that makes sure both FreeBSD and Musl systems works.

The problem lies actually in the Autoconf-enabled submodules which have a very outdated `config.guess` and `config.sub`. These are static files that need to be updated in-source of `libbacktrace`, but can be updated in-place like this:

```shell
cd ~/src/lightning
git clean -xfd
git submodule deinit --all -f
git checkout 27a186be # the state with MAKE_HOST
git submodule update --init --recursive
sh /tmp/config-fix.sh
make -j4
```

And here is the `config-fix.sh` script:

```shell
#!/bin/sh

MYLOC=/tmp

myget() {
  wget -q --content-disposition \
    "http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;\
f=config.$1;hb=HEAD" && echo File config.$1 downloaded
}
cd /tmp; myget guess; myget sub; cd - >/dev/null 2>&1

find . -name 'config.guess' | while read target
do
  dir=${target%/*}
  cp -f $MYLOC/config.guess $MYLOC/config.sub $dir/
  echo Files in $dir updated
  cd - >/dev/null 2>&1
done
rm $MYLOC/config.sub $MYLOC/config.guess
```

jsarenik:
@NicolasDorier ^^ above in-place change of `config.{guess,sub}` files is a way to compile on Musl system without the patch in this pull-request.

jsarenik:
If the above pull-request to libbacktrace is accepted, the submodule should be updated and that update commited to our `.gitmodules` file. Then this patch is not needed and rather should not be merged.

Still, the fact that `./configure` script in `libbacktrace` is _always_ called with `--host=$MAKE_HOST` is rather non-standard I would say (in case of no cross-compilation). Luckily when the `config.guess` and `config.sub` scripts are up-to-date (and know musl) they can cope with it.

There is a way to add `--host` flag to `./configure` only when needed, but in that case `MAKE_HOST` can not be used because this is _always_ defined. It would be something like:

beginning of `external/Makefile`:

```
ifdef CROSS_HOST
HOST_ADD := --host=$(CROSS_HOST)
endif

...

... ./configure ... $(HOST_ADD) ...
...
```

NicolasDorier:
Not very familiar with build systems in C++, so I wonder: What is the difference between updating libbacktrace to allow `-musl` and keep as now, and not updating it and use `CROSS_HOST`.

jsarenik:
The difference is that CROSS_HOST it totally made-up, but the fact this
environment variable should not be used by others is what fixes the problem
(i.e. it will be empty instead of containing GNU Make pre-defined
architecture identifier).

On the other hand, updating the config dot sub and guess scripts in
libbacktrace makes the build system understand it is not asked to
cross-compile when  `--host` option is set and contains GNU Make
pre-defined MAKE_HOST
value of a musl libc based host (which is not known to the config.guess/sub
scripts in their 2012 version and it may require more updates in the future
when new architectures or libc implementations appear).

The best solution, using something like GNU Autoconf, would make use of
proper cross-compilation variables only when needed so this would not
happen until someone really tries to cross-compile (i.e. `--host` and
`--build` should not be set when calling `./configure` script when one does
not intend to cross-compile).

General RANT following - no offense please:

By the way, one paraphrased quote related to GNU Autoconf is that those who
cannot use it are condemned to reimplement it*. Poorly. With all respect. I
love simplicity myself, just that portability is not an easy thing in the
real world of corporate UNIXes and actually relies on a list of particular
bugs (implementation details in different systems or their versions) which
is very well provided by GNU Autotools.

* https://en.wikiquote.org/wiki/George_Santayana

jsarenik:
The difference is that CROSS_HOST it totally made-up, but the fact this
environment variable should not be used by others is what fixes the problem
(i.e. it will be empty instead of containing GNU Make pre-defined
architecture identifier).


On the other hand, updating the config dot sub and guess scripts in libbacktrace makes the build system understand it is not asked to cross-compile when  `--host` option is set and contains GNU Make pre-defined MAKE_HOST value of a musl libc based host (which is not known to the config.guess/sub scripts in their 2012 version and it may require more updates in the future when new architectures or libc implementations appear).

The best solution, using something like GNU Autoconf, would make use of proper cross-compilation variables only when needed so this would not happen until someone really tries to cross compile (i.e. `--host` and `--build` should not be set when calling `./configure` script when one does
not intend to cross-compile).

General RANT following - no offense please:

By the way, one paraphrased quote related to GNU Autoconf is that those who cannot use it are condemned to reimplement it*. Poorly. With all respect. I love simplicity myself, just that portability is not an easy thing in the real world of corporate UNIXes and actually relies on a list of particular
bugs (implementation details in different systems or their versions) which is very well provided by GNU Autotools.

* https://en.wikiquote.org/wiki/George_Santayana

jsarenik:
Closing this one. Build works well in #1538

PR #1534, opened 2018-05-31
Implement gossip_store compactions

cdecker:
Compacts the `gossip_store` if we have more than 100 messages and 10% stale.

### DIFF
--- a/gossipd/broadcast.c
+++ b/gossipd/broadcast.c
@@ -88,6 +88,7 @@ const u8 *next_broadcast(struct broadcast_state *bstate,
 	return NULL;
 }
 
+#ifdef PEDANTIC
 static const struct pubkey *
 pubkey_keyof(const struct pubkey *pk)
 {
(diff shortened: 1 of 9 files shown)


cdecker:
I have absolutely no idea why GH reorders the commits this way...

ZmnSCPxj:
github reorders commits according to the commit-time, so if you did some reordering of commits in e.g. `git rebase -i`, the ordering on github gets massively confused since they do not follow the `parent` links but instead just the commit times.

Sometimes I just do `git format-patch` and manually edit the times to be incrementing in the .patches, reset hard to origin, then `git am` the patches.

rustyrussell:
Actually, do we have to store local announcements in the gossip_store at all?  It seems to cause all kinds of issues; perhaps we should simply hang the gossip_store off the broadcast api?

cdecker:
Rebased on top of `master`, now has some regression that stores invalid `channel_updates` in the store...

cdecker:
Should be good to review. Rebased and added the last 3 commits :wink:

gossipd/broadcast.c
```diff
@@ -88,6 +88,7 @@ const u8 *next_broadcast(struct broadcast_state *bstate,
 	return NULL;
 }
 
+#ifdef PEDANTIC
 static const struct pubkey *
```
rustyrussell: OK, the way CCAN does this is that the xxx_check function is always defined, and a wrapper xxx_debug() calls in if XXX_DEBUG is defined, otherwise it's a noop.  I'll implement that as a separate patch (we should really have it for all primitives we use, and then have a RPC call to do an internal consistency check).

PR #1539, opened 2018-06-04
udpate broken link in deployable-lightning.lyx

arowser:
The schnorr signatures link have been update now.


cdecker:
@rustyrussell are we actively maintaining this document? If not I suggest removing it.

PR #1538, opened 2018-06-04
WIP: ./configure

rustyrussell:
[ On top of gossip changes, so ignore all patches up to and including 'gossipd: minor cleanups.' ]

This adds a ./configure step for those of us who hate remembering to 'make DEVELOPER=1'.  It's not autoconf, but if you can write shell it should be pretty simple to enhance.

Main user-visible change is that you need to use 'VALGRIND=0' instead of 'NO_VALGRIND=1' to override valgrind testing.

@jsarenik might want to comment, since he's doing exotic cross-builds!

PR #1542, opened 2018-06-04
external/Makefile: Simplify build and fix race condition

jsarenik:
Git magic should not happen from inside the Makefiles. It may
case pain when simultaneous make jobs are run. See #1538
where I first mentioned this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: c
 dist: trusty
 sudo: true
 
+before_install:
+  - git submodule update --init --recursive
+
 notifications:
   email: false
 
--- a/external/Makefile
+++ b/external/Makefile
@@ -21,20 +21,12 @@ EXTERNAL_INCLUDE_FLAGS :=					\
 
 EXTERNAL_LDLIBS := -Lexternal $(patsubst lib%.a,-l%,$(notdir $(EXTERNAL_LIBS)))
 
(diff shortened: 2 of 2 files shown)


jsarenik:
@cdecker Please re-run the tests. It was a random wait-fail. Is anyone looking at this? If not, I will. Randomly failing tests are a plague :-)

rustyrussell:
No. Not requiring people to run magic commands, esp. after every pull, is precisely what build systems are *for*.  I added the configure step to *reduce* the stuff I forget to do, this increases it again.

I will investigate the Alpine build race tomorrow, but we need make to deal with the git submodules insanity because nobody else can.

jsarenik:
I see your point. Yes, the build systems are very useful in simplifying (grouping related) things. The configure script is wonderful and linear, thank you for that.

jsarenik:
On Debian 18.04 the build success rate is 100% when the git submodules are initialized before running `make`. For reproduction steps of the race issue see #1543 

I think 08b2362ddc91c94721ec95088dd603025c27eb6a made it much probable for build (ran according to the README, i.e. calling `make` right after `git clone`) to not succeed.

jsarenik:
Closing as the https://github.com/rustyrussell/lightning/commit/5d77c6a9fb578aa38eec401ee042580ba7ee9a63 commit fixes the issue and will be merged in #1538 . Thanks!

PR #1544, opened 2018-06-05
In Debian at least, the Python 3 version of pytest is called pytest-3.

bitonic-cjp:
Use this executable name to make sure we don't try to use the Python 2 version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ ifneq ($(NO_COMPAT),1)
 COMPAT_CFLAGS=-DCOMPAT_V052=1
 endif
 
-PYTEST := $(shell command -v pytest 2> /dev/null)
+PYTEST := $(shell command -v pytest-3 2> /dev/null)
 PYTEST_OPTS := -v -x
 
 # This is where we add new features as bitcoin adds them.


cdecker:
NACK

This breaks the build for people that use `pip3` to install `pytest`. This is really debian/ubuntu's fault for renaming the scripts.

bitonic-cjp:
Is a different approach possible? Maybe make this name configurable? Maybe have pytest-3 as default, and automatically try pytest if pytest-3 doesn't exist? I'd prefer pytest-3 as default, since explicit (mentioning of Python version) is better than implicit.

How do other distributions deal with having both Python 2 and Python 3 versions of this program?

cdecker:
Given that this is probably highly platform specific I think it should be rolled into #1538 and select the correct executable from a list of known executables.

bitonic-cjp:
I agree with using #1538 as a solution.

cdecker:
@bitonic-cjp can you check that PR #1557 detects the pytest binary correctly? It'll go through a list of known aliases, the last ones being the original scripts names if installed from `pip`.

bitonic-cjp:
@cdecker it correctly detects pytest-3 on my system, even though (to make it difficult) the Python 2 pytest was also installed.

So, all OK; thanks.

PR #1545, opened 2018-06-05
Do not ask bitcoind for outpoints we should know

cdecker:
We now remember the block range that we have stored in the DB and don't go ask for outpoints that we would know if they'd exist.

This was a common problem before: we'd receive a `channel_announcement` for a spent outpoint (the remote node didn't prune) and we'd always call out to `bitcoind` even though we removed the outpoint earlier and it not existing should have been the signal to ignore the announcement.

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -745,14 +745,14 @@ struct chain_topology *new_topology(struct lightningd *ld, struct log *log)
 
 void setup_topology(struct chain_topology *topo,
 		    struct timers *timers,
-		    u32 first_blocknum)
+		    u32 min_blockheight, u32 max_blockheight)
 {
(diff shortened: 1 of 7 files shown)


lightningd/chaintopology.c
```diff
@@ -745,14 +745,14 @@ struct chain_topology *new_topology(struct lightningd *ld, struct log *log)
 
 void setup_topology(struct chain_topology *topo,
 		    struct timers *timers,
-		    u32 first_blocknum)
+		    u32 min_blockheight, u32 max_blockheight)
 {
 	memset(&topo->feerate, 0, sizeof(topo->feerate));
 	topo->timers = timers;
 
 	/* FIXME(cdecker) Actually load this from DB */
```
rustyrussell: Remove FIXME now?

PR #1551, opened 2018-06-08
Increase cltv_final

verretor:
I think it's better if the default value works with other implementations. If you want to be more aggressive, you'd start lightningd with `--cltv-final=6` for instance.

As explained in https://github.com/ACINQ/eclair/issues/526, Eclair won't accept payment requests with `cltv_final` too low. Error message from Eclair for any value below 10:
```
1 attempts:
- (remote) payment expiry is too close to the current block height for safe handling by the final node
```

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -492,7 +492,7 @@ static const struct config testnet_config = {
 
 	/* Be aggressive on testnet. */
 	.cltv_expiry_delta = 6,
-	.cltv_final = 6,
+	.cltv_final = 10,
 
 	/* Send commit 10msec after receiving; almost immediately. */
 	.commit_time_ms = 10,
@@ -544,7 +544,7 @@ static const struct config mainnet_config = {
 	 *
 	 * The minimum `cltv_expiry` we will accept for terminal payments: the
 	 * worst case for the terminal node C lower at `2R+G+S` blocks */
-	.cltv_final = 8,
+	.cltv_final = 10,
 
 	/* Send commit 10msec after receiving; almost immediately. */
 	.commit_time_ms = 10,


cdecker:
NACK

We discussed this during our meetings. The issue really is that Eclair is happy to create a payment route that uses a smaller final CLTV delta for the last hop, but then will fail when attempting to route that payment. There is absolutely no reason for them to enforce a remote CLTV delta at all, and even if they wanted to they could just bump that last CLTV delta when generating the route.

The current way is like taking a city map, planning a path through it and then complaining that the path they chose crosses one bridge too many, when they could just have selected a different path on their own.

verretor:
Another reason that I suggested this:
In BOLT11, it specifies: "`min_final_cltv_expiry` to use for the last HTLC in the route. Default is 9 if not specified."

Currently, min_final_cltv_expiry == cltv_final
https://github.com/ElementsProject/lightning/blob/294c76e749996bd5d035d0bcc72dad12d63113dd/lightningd/invoice.c#L332

I personally set cltv-final to 10 just to test with Eclair. Feel free to keep the current settings if it is the right thing to do.

cdecker:
Quoting my reply to ACINQ/eclair#526 since it is relevant for this discussion:

> @sstone I'm still not sure why eclair would enforce the final CLTV delta of 10 on the last hop for sending payments, even though we explicitly signaled that we're happy with 6. This is really confusing to me, since your path finding code will happily produce such a route, but then when issuing the payment using that route, the routing mechanism refuses to route it. I think rather than us having to bump the default param, Eclair's path finding should be amended to bump the CLTV delta if it's too small, since we will accept something that is higher than our default.
> 
> I remember we talked about this before, and I think I convinced you that Eclair was enforcing a value that it should not, and even if, it should be the sending side to ensure its lower bounds are met. Why should we as receiving party have to guess what the sending party's implementation is so that we can pass them an invoice that matches their enforced values?

verretor:
How would you like it to be phrased?

cdecker:
Let me give it a try:

> During a meeting earlier this week we agreed with Eclair to temporarily increase the final CLTV delta in our invoices to establish compatibility with the already deployed Eclair wallets. They in turn agreed to remove the enforcement of higher final CLTV deltas, or bump it locally should it not match their expectations as allowed by [BOLT 11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#tagged-fields). This has since been implemented in ACINQ/eclair#627.

verretor:
Alright. Will change the commit message when I come back home.

cdecker:
That would be perfect, I'll merge it right away :-)

verretor:
Commit message updated.

cdecker:
ACK 63f0d36

Issue #1549, opened 2018-06-07
In-memory order for gossip messages is incorrect


cdecker:
### Issue and Steps to Reproduce
While trying to get the compaction working, rebased on top of #1521, I encountered an issue that would result in the `gossip_store` file almost always being truncated, if the previous run performed a compaction. The compaction reads from the in-memory ordered gossip messages, which is supposed to always be kept in the correct order, but some `channel_update`s and `node_announcement`s seem to be queued in incorrect order.

This is a regression and was introduced by PR #1521.

cdecker:
Not surprisingly my short journey down git bisect land points the finger at 601a8555437b2adbc9c3011e98f7de5c7c53592d as being the first commit that breaks.

cdecker:
This hacky function can be used to find inconsistent message ordering:

```C
#ifdef DEVELOPER

static const struct pubkey *
pubkey_keyof(const struct pubkey *pk)
{
	return pk;
}

static size_t pubkey_hash(const struct pubkey *id)
{
	return siphash24(siphash_seed(), id, sizeof(*id));
}

HTABLE_DEFINE_TYPE(struct pubkey,
		   pubkey_keyof,
		   pubkey_hash,
		   pubkey_eq,
		   pubkey_set);

static void routing_check_consistency(struct routing_state *rstate)
{
	secp256k1_ecdsa_signature sig;
	const u8 *msg;
	u8 *features, *addresses, color[3], alias[32];
	struct bitcoin_blkid chain_hash;
	struct short_channel_id scid;
	struct pubkey node_id_1,  node_id_2, bitcoin_key;
	u32 timestamp, fees;
	u16 flags, expiry;
	u64 index = 0, htlc_minimum_msat;

	struct pubkey_set pubkeys;
	pubkey_set_init(&pubkeys);
	UINTMAP(u64 *) channels;

	while ((msg = next_broadcast(rstate->broadcasts, 0, UINT32_MAX,
				     &index)) != NULL) {
		if (fromwire_channel_announcement(
			tmpctx, msg, &sig, &sig, &sig, &sig, &features,
			&chain_hash, &scid, &node_id_1, &node_id_2,
			&bitcoin_key, &bitcoin_key)) {
			uintmap_add(&channels, scid.u64, &index);
			pubkey_set_add(&pubkeys, &node_id_1);
			pubkey_set_add(&pubkeys, &node_id_2);

		} else if (fromwire_channel_update(msg, &sig, &chain_hash,
[...]

PR #1552, opened 2018-06-08
Fix node announce order

rustyrussell:
Fixes: #1549


gossipd/broadcast.c
```diff
@@ -71,3 +82,88 @@ const u8 *next_broadcast(struct broadcast_state *bstate,
...
+		      scid ? type_to_string(tmpctx,
+					    struct short_channel_id,
+					    scid)
+		      : type_to_string(tmpctx, struct pubkey, node_id));
+	if (abortstr)
+		abort();
+	return NULL;
+}
+
+struct broadcast_state *broadcast_state_check(struct broadcast_state *b,
```
cdecker: We should definitely guard this with `DEVELOPER` since it's quadratic in the number of gossip messages.

cdecker: Oh yes, this is painful on mainnet :wink:

cdecker:
ACK 3ef3d0614ab80c7b2d6a8127f6fd9bd852203a31

PR #1556, opened 2018-06-08
gossip: Duplicate removes can result in null pointers in broadcast

cdecker:
The destructor not being the only place where we delete from the `uintmap` means that sometimes the delete in the destructor will end up not finding the item. Specifically this will happen here:

https://github.com/ElementsProject/lightning/blob/4429c6e7cd78c1eca39742ad57d5ff02ce5e6852/gossipd/routing.c#L230-L233

We delete from the map, don't reinsert, and then eventually we `tal_free` the message.


cdecker:
ACK 71aaf46

Acking this myself since it is rather urgent.

PR #1558, opened 2018-06-09
Armbian build instructions

Sjors:
There's more pies than the Raspberry.

Armbian uses Ubuntu 16.04 / 18.04 to build images that you can then put on an SD card. You can pass extra commands to that process which I've used to compile c-lightning. This works fine using just the regular Ubuntu instructions, but it's pretty slow. I'm not entirely sure how Armbian tricks the build process into creating the correct binary for the device architecture, but it does.

Based on experience cross-compiling Bitcoin Core, I suspect cross-compiling c-lightning will be much faster if done on the Ubuntu VM, rather than during the Armbian build process. However I wasn't able to figure out how to tweak the Raspberry Pi instructions to do this.

Perhaps these instructions are useful as-is, but otherwise we can leave this open until someone figures it out.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -7,7 +7,8 @@
 6. [macOS](#to-build-on-macos)
 7. [Android](#to-cross-compile-for-android)
 8. [Raspberry Pi](#to-cross-compile-for-raspberry-pi)
-9. [Additional steps](#additional-steps)
+9. [Armbian](#to-compile-for-armbian)
+10. [Additional steps](#additional-steps)
 
 Library Requirements
 --------------------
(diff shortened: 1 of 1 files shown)


cdecker:
Excellent addition, I wanted to create a custom image as well, and this is just what I was looking for. Shall I mark this as WIP while you try to figure out the `TODO` in there?

wythe:
Make sure you add a link from the table of contents.

Sjors:
WIP sounds sounds good. It's unlikely I'll be able to figure out how to do the cross-compile though.

cdecker:
That's also ok, but we should remove the TODO then :-)

Sjors:
I thought you meant keeping the PR open in case someone does have clues / ability how to make the cross compile work.

cdecker:
Sure that works as well, I just don't like having TODO markers in published documentation :wink:

cdecker:
ACK 1a474b7

Sjors:
It also needs a TOC entry, let me just push that and get rid of the todo for you...

Sjors:
Pushed, I also linked to the working example at https://github.com/Sjors/armbian-bitcoin-core

cdecker:
ACK c9f37ad

PR #1569, opened 2018-06-12
Improved JSON-RPC Parameter Processing

wythe:
This PR introduces a typesafe callback system for unmarshaling and validating
json token parameters.  It improves upon the current `json_get_params()` method.

It's a continuation of our efforts on #1464.

  Typical usage: (this is from a comment in params.h)

	uint64_t msatohi;
	unsigned cltv = 9;
	const jsmntok_t *seedtok;

	struct param_table *pt = new_param_table(cmd);
	param_add(pt, "msatoshi", json_tok_u64, &msatoshi);
	param_add(pt, "?cltv", json_tok_number, &cltv);
	param_add(pt, "?seed", json_tok_tok, &seedtok);
	if (!param_parse(pt, buffer, params))
		return;

  At this point in the code, you can be assured msatoshi, cltv, and seedtok are
  valid.  If not, param_parse() returned false, having already called
  command_fail() with a descriptive error message. The data section of the json
  result contains the offending parameter and its value.

  The '?' before cltv indicates it is optional.  You can see if cltv was
  specified by calling param_is_set(pt, &cltv).

  The seed parameter (also optional) uses a special callback, json_tok_tok(). It
  simply sets seedtok to the appropriate value and lets the handler do the
  validating. It has the added feature of setting seedtok to NULL if it is null
  or not specified.

  There are canned failure messages for common callbacks. An example:

	'msatoshi' should be an unsigned 64 bit integer, not '123z'

  Otherwise a generic message is provided.

I converted all existing commands over to this new system, but
json_get_params() is still available if you really want to use it.

NOTE: I don't know why the diff is showing differences for libbacktrace and libsodium, and I don't know how to fix it.


rustyrussell:
I had a tree like this that I abandoned. I'll dig it out today and we can compare notes: I think it's a good idea though!

cdecker:
You seem to have a bump of `libbacktrace` in there, that's unrelated I assume.

wythe:
Yeah I'm not sure what happened with libbacktrace and libsodium.  i did not intend to change those.  I mentioned that above. What's the proper way for me to undo that?

rustyrussell:
Good q!  I think you need to go into the submodules, check out the expected versions, then at the top dir do `git submodule update`.  Then `git commit --amend`.

Though this patch does a lot of things, and really should be split into parts for review anyway.

rustyrussell:
Here's my (uncommented, v. out-of-date) attempt at a similar thing: https://github.com/rustyrussell/lightning/tree/guilt/cleanup-jsonrpc-stringhandling

I used explicit macros for typesafety (ie. if you said something was a u32, you needed to supply a u32 pointer), and used a vararg function rather than building a structure, but it's the same idea.  I decided it was too much churn for 0.6, though.

Comments?

wythe:
Rusty are you sure that is the right link you provided?  It seems to just be about string handling.

However, my approach also uses type safe callbacks.  In fact, it was discovering `typesafe_cb.h` that convinced me this effort might be worth while! `param_add()` is a macro defined in `lightningd/params.h` that handles it.

I know this PR got pretty big. But I wanted to convert the handlers to make sure it made sense.  And for the most part I think it does.  I have no problem breaking it up though.

rustyrussell:
OIC!  Yeah, I'd prefer to just hand them as variable args to param_parse, though, eg:

```C
uint64_t msatohi;
unsigned cltv = 9;
const jsmntok_t *seedtok;

if (!param_parse(buffer, params,
                           COMPULSORY_PARAM("msatoshi", json_tok_u64, &msatoshi),
                           OPTIONAL_PARAM("cltv", json_tok_number, &cltv),
                           OPTIONAL_PARAM("seed", json_tok_tok, &seedtok),
                           NULL))
	return;
```

Or something... Then I'd split into separate patches:

1. Add new JSON error codes you need, and document.
2. Add param_parse and helpers as required.
3. Convert existing callers (or maybe just the trivial ones).
4. Optionally add new boutique (ie. local static) json_tok_ handlers to do the harder cases.
5. If you do them all, then remove the old parsing code.

Thanks!
Rusty.

wythe:
Sounds good.  I'm on it.

PR #1570, opened 2018-06-12
daemon: print version in crash backtrace.

rustyrussell:
Sometimes people only send us the crash, not the logs.


cdecker:
ACK 808b07d

Issue #486, opened 2017-12-28
Test suite might incorrectly run tests under Python 2 due to inconsistent pytest script naming


practicalswift:
The correct fix is probably to teach `configure` to identify the correct Python 3-based `pytest` script.

See discussion in #406.

jsarenik:
Can the Python2 Pytest be version 3 itself?

jsarenik:
Yes, it can. https://docs.pytest.org/en/latest/changelog.html

jsarenik:
@practicalswift Please have a look at #1574 . Thanks!

jsarenik:
I found it. https://github.com/bitcoin/bitcoin/issues/6100

i.e. `utACK` means untested ACK

cdecker:
Fixed as per PR #1574

PR #1574, opened 2018-06-13
configure: Check for version of pytest

jsarenik:
Addresses #486

### DIFF
--- a/configure
+++ b/configure
@@ -66,6 +66,7 @@ find_pytest()
     PYTEST_BINS="pytest-3 pytest3 pytest py.test"
     for p in $PYTEST_BINS; do
 	if [ "$(which $p)" != "" ] ; then
+	    "$p" --version 2>&1 | grep -q "python3" || continue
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -73,6 +73,7 @@ find_pytest()
 }
 
 PYTEST=${PYTEST:-`find_pytest`}
+$PYTEST --version 2>&1 | grep -q "version 3" || unset PYTEST
```
rustyrussell: We should NOT override if they set PYTEST manually, though.  I think this logic needs to be in find_pytest?

jsarenik:
I have tested with `PYTEST=/bin/true ./configure`

practicalswift:
utACK 59c54341fbf198d3475916eeb03054a1cb5bafbf

jsarenik:
@practicalswift what does `ut` in `utACK` mean?

cdecker:
ACK 59c5434

PR #1578, opened 2018-06-14
Gossip leak plugging/neatening and channeld crash fix.

rustyrussell:
We've had some weird crashes on sleepyark, so I audited gossipd.  Found some lifetime issues, and user-after-free in corner cases, but no smoking gun.

I also fixed an overzealous assert in lightning_channeld.

### DIFF
--- a/lightningd/bitcoind.h
+++ b/lightningd/bitcoind.h
@@ -38,7 +38,7 @@ struct bitcoind {
 	struct lightningd *ld;
 
 	/* Are we currently running a bitcoind request (it's ratelimited) */
-	bool req_running;
+	struct bitcoin_cli *current;
 
 	/* Pending requests. */
 	struct list_head pending;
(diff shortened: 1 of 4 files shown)


lightningd/bitcoind.h
```diff
@@ -38,7 +38,7 @@ struct bitcoind {
 	struct lightningd *ld;
 
 	/* Are we currently running a bitcoind request (it's ratelimited) */
-	bool req_running;
+	struct bitcoin_cli *req_running;
```
cdecker: The name `req_running` still suggests that it is boolean. Shall we change it to `current` or something similar?

rustyrussell: Ack, will rebase w that change...

cdecker:
ACK deecfcdc3ef1f5b5d162cf78cfa9fbd9459bf2f3

PR #1579, opened 2018-06-14
listinvoice: speed up single-invoice case.

rustyrussell:
satoshis.place was slowing to a crawl, c-lightning was unresponsive.
Logs revealed charged doing many, many listinvoice <label> RPCs.

We were iterating the entire db every time: stop that!

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -384,11 +384,20 @@ static void json_add_invoices(struct json_result *response,
 	struct invoice_iterator it;
 	struct invoice_details details;
 
+	/* Don't iterate entire db if we're just after one. */
+	if (label) {
+		struct invoice invoice;
+		if (wallet_invoice_find_by_label(wallet, &invoice, label)) {
+			wallet_invoice_details(response, wallet, invoice,
+					       &details);
+			json_add_invoice(response, &details, modern);
+		}
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


shesek:
> Logs revealed charged doing many, many listinvoice RPCs.

Interesting. The only thing Charge uses `listinvoices` for is to ensure expired invoices are really deleted from c-lightning's db before deleting them from charge's db. Other than that, everything is being queried directly from charge's db to minimize the load on c-lightning as much as possible.

https://github.com/ElementsProject/lightning-charge/blob/db0569ae9e7983b82880c28aaf240071345fb4f9/src/model.js#L56-L69

Any thoughts on an alternative way to implement this? This seems to be the main thing in Charge that puts a load on c-lightning. (other than that, the only things the c-lightning RPC is used for are invoice creation and a single `waitanyinvoice` listener)

shesek:
Also, I just realized that running c-lightning without setting up `autocleaninvoice` would cause Charge to try and delete the same invoices over and over again (but won't actually delete anything ever), with the list of expired invoices only growing over time and never shrinking.

This could be the reason Charge sent "many, many" `listinvoice` calls, which it normally shouldn't (with `autocleaninvoice` on, it should only fetch invoices that expired in the last 4 hours, which shouldn't really grow that much).

Issue #338, opened 2017-11-01
Require all SQL parameters to be bound and explicitly set null if we don't have a value


cdecker:
As pointed out by @rustyrussell this should help us detect an eventual off-by-one error in the parameter assignment.

jsarenik:
How to find a case which does not use bound parameters or does not explicitly set null? Is it still in the code?

jsarenik:
Ah, sorry, just realized the `v.0.6.1` tag means you will get to it after release. Sorry for bothering.

cdecker:
No worries @jsarenik, we can definitely start working on this and merge it if it's ready earlier :-)

I'm not actually sure `sqlite3` has a way to enforce null parameters to be explicitly set, but a lot of our code looks like this:

```
...
if (a) {
    sqlite3_bind_int(stmt, 0, 42);
}
...
```

it'd be more readable to make if 

```
...
if (a) {
    sqlite3_bind_int(stmt, 0, 42);
} else {
    sqlite3_bind_null(stmt, 0);
}
...
```
in order to make that implicit null explicit.

jsarenik:
```
~/src/lightning$ git grep -n -B1 "sqlite3_bind_" | grep "if (.*)\( {\)
*"
wallet/db.c-731-	if (!id) {
wallet/db.c-829-	if (!pks) {
wallet/invoices.c-301-	if (msatoshi)
wallet/wallet.c-90-	if (utxo->close_info) {
wallet/wallet.c-100-	if (utxo->blockheight) {
wallet/wallet.c-105-	if (utxo->spendheight)
wallet/wallet.c-889-	if (chan->scid)
wallet/wallet.c-908-	if (chan->remote_shutdown_scriptpubkey)
wallet/wallet.c-1121-	if (in->preimage)
wallet/wallet.c-1163-	if (out->in)
wallet/wallet.c-1171-	if (out->preimage)
wallet/wallet.c-1199-	if (payment_key)
wallet/wallet.c-1810-	if (failonionreply)
wallet/wallet.c-1819-	if (failnode)
wallet/wallet.c-1823-	if (failchannel) {
wallet/wallet.c-1831-	if (failupdate)
wallet/wallet.c-1973-	if (b->prev) {
wallet/wallet.c-2152-		if (blockheight) {
```

jsarenik:
[...]

PR #1588, opened 2018-06-15
Jasan/bound sql

jsarenik:
Address #338

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -1176,6 +1176,8 @@ void wallet_htlc_save_out(struct wallet *wallet,
 
 	if (out->preimage)
 		sqlite3_bind_preimage(stmt, 8,out->preimage);
+	else
+		sqlite3_bind_null(stmt, 8);
 	sqlite3_bind_int(stmt, 9, out->hstate);
 
 	sqlite3_bind_blob(stmt, 10, &out->onion_routing_packet,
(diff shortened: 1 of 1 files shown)


rustyrussell:
What I'd really like is far more ambitious: a set of db wrappers which were used like:

```
   for_db(wallet->db, "SELECT x,y from Y WHERE foo=? and bar=?",
              DB_ARG(u64, &x),
              DB_ARG(struct pubkey, &y),
              DB_CONST_ARG(struct pubkey, &foo),
              DB_CONST_ARG(u64, bar))
        ...
```

This would set x and y for each row returned, and of course do compile time checks on the arguments and runtime checking that they matched the statement (both in number, and constness, perhaps in DEVELOPER mode with the type as well, though that's approximate).

and 
```
   alter_db(wallet->db, "DELETE FROM blocks WHERE hash = ?",
                 DB_ARG(struct sha256_double, &hash));
```

which is a simpler variant.

But it's terribly complex, which is why I haven't done it!

jsarenik:
@rustyrussell I see your point, but why not adding these in the meantime in order to address the old issue? Or maybe rephrase #338 then.

The currently failing test on c75c5f1e is totally unrelated (`a2x` missing in one docker instance - new issue https://github.com/ElementsProject/lightning/issues/1642). So the tests are passing.

PR #1590, opened 2018-06-15
Added wallet related error codes

wythe:
New codes: FUND_MAX_EXCEEDED, FUND_CANNOT_AFFORD, FUND_DUST_LIMIT_UNMET.

The error message "Cannot afford fee" was not exactly correct because
it would also occur if the amount requested could not be afforded.  So
I changed it to the more generic "Cannot afford transaction".

Other things:

* Fixed off-by-one satoshi in fundchannel manpage.
* Changed 'arror' to 'error' because we are not pirates.

### DIFF
--- a/common/wallet_tx.c
+++ b/common/wallet_tx.c
@@ -12,6 +12,21 @@ void wtx_init(struct command *cmd, struct wallet_tx * wtx)
 	wtx->all_funds = false;
 }
 
+static bool check_amount(const struct wallet_tx *tx)
+{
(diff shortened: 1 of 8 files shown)


rustyrussell:
Minor change: I thought "dust limit unmet" a bit vague compared with "output was too small"...  @wythe please ack/nak!

wythe:
ACK

cdecker:
ACK 2341bd5

PR #1595, opened 2018-06-17
Bolt updates (no code changes)

rustyrussell:
We'd fallen well behind, this laboriously catches us up.

I'd really like to analyze how much of the requirements we cover, but at least now we quote the
spec correctly!


cdecker:
ACK e776453

PR #1600, opened 2018-06-20
pytest: make test_connect use an unknown id.

rustyrussell:
This test failed:

       AssertionError: "No address known" does not match "RPC call failed: {'code': -1, 'message': 'Connection establishment: Connection timed out'}, method: connect, payload: {'id': '032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e'}"

Because someone seems to have actually used this ID on testnet!

	DEBUG:root:lightningd-1: 2018-06-19T07:14:39.413Z lightning_gossipd(3467): Resolved ln1qvk0zhg6m8z2prfxa2cerrmn9k803lwx4wukgzlnmvt5xukyjycyugwvuk8.lseed.bitcoinstats.com to 91.194.91.220:9735

Change the last digit, creating an unknowable ID so this can't happen again.

Reported-by: @sturles
Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

PR #1601, opened 2018-06-20
Add cli option to disable DNS node ID lookup

cdecker:
This is a more durable fix of #1600 I think. It adds the option to disable DNS lookups of node IDs, uses that option while testing, and also tries to identify a misbehaving DNS ISP resolver that hijacks NXDOMAIN replies, like @gwillen pointed out might happen. It's best effort, but even if it fails, it'll just perform a failing connect attempt.

What do you think @rustyrussell, shall we keep that last commit in there?

### DIFF
--- a/doc/lightningd-config.5.txt
+++ b/doc/lightningd-config.5.txt
@@ -100,7 +100,7 @@ Lightning daemon options:
 
 *pid-file*='PATH'::
     Specify pid file to write to.
-  
+
 *log-level*='LEVEL'::
     What log level to print out: options are io, debug, info, unusual, broken.
 
@@ -278,10 +278,13 @@ or precisely control where to bind and what to announce with the
     can still connect to Unix domain sockets manually).  This also disables
     all DNS lookups, to avoid leaking information.
 
+*disable-dns*::
+    Disable the DNS bootstrapping mechanism to find a node by its node ID.
+
 *tor-service-password*='PASSWORD'::
     Set a Tor control password, which may be needed for 'autotor:' to
     authenticate to the Tor control port.
-    
+
 BUGS
 ----
 You should report bugs on our github issues page, and maybe submit a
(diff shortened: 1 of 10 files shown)


gossipd/gossip.c
```diff
@@ -360,6 +364,39 @@ new_local_peer_state(struct peer *peer, const struct crypto_state *cs)
...
+			      &hints, &addrinfo);
+
+	daemon->broken_resolver_response =
+	    tal_free(daemon->broken_resolver_response);
+
+	if (err == 0) {
+		daemon->broken_resolver_response = tal(daemon, struct sockaddr);
+		memcpy(daemon->broken_resolver_response, addrinfo->ai_addr,
+		       sizeof(struct sockaddr));
+		freeaddrinfo(addrinfo);
```
rustyrussell: `daemon->broken_resolver_response = tal_dup(daemon, struct sockaddr, addrinfo->ai_addr)` ?

doc/lightningd-config.5.txt
```diff
@@ -278,10 +278,13 @@ or precisely control where to bind and what to announce with the
     can still connect to Unix domain sockets manually).  This also disables
     all DNS lookups, to avoid leaking information.
 
+*disable-dns*::
+    Disable the DNS bootstrapping mechanism to find a node by its node ID.
+
```
rustyrussell: I generally prefer options to be xxx=<bool>, since you can override them (eg. suppress in the config file, enable on the cmdline). But re-enabling DNS seems weird, so it's OK.

lightningd/options.c
```diff
@@ -409,6 +409,9 @@ static void config_register_opts(struct lightningd *ld)
 			       opt_set_bool_arg, opt_show_bool,
 			       &ld->use_proxy_always, "Use the proxy always");
 
+	opt_register_noarg("--disable-dns", opt_set_bool, &ld->config.no_dns,
+			   "Disable DNS lookups of peers");
```
rustyrussell: I've been shying away from negative bools, so suggest we call it 'dns' and use opt_set_invbool here...

PR #1616, opened 2018-06-21
docker: Add a usable dockerfile

cdecker:
This is based on @NicolasDorier's excellent proposal for a Dockerfile in PR #1318, sans the
writing of a config file.


README.md
```diff
@@ -64,13 +64,58 @@ For the impatient here's the gist of it for Ubuntu and Debian:
...
+If you build the image yourself, you can use the build arg `DEVELOPER=1` to build c-lightning in developer mode.
 
-    sudo docker run \
-    	-v $HOME/.lightning:/root/.lightning \
-    	-v $HOME/.bitcoin:/root/.bitcoin \
-    	-p 9735:9735 \
-    	cdecker/lightningd:latest
+It has the following environment variable:
+
+* `EXPOSE_TCP` default to false, if true, use expose c-lightning on port 9835. (Use this only for testing)
```
rustyrussell: You mean 9735?

Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+    && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
+    && rm litecoin.tar.gz
+
+ENV LIGHTNINGD_VERSION=master
+
+WORKDIR /opt/lightningd
+COPY . .
+
+ARG DEVELOPER=0
+RUN ./configure && make -j3 DEVELOPER=${DEVELOPER} && cp lightningd/lightning* cli/lightning-cli /usr/bin/
```
NicolasDorier: `DEVELOPER=${DEVELOPER}` is needed ? I thought with `./configure` it should be ok.

Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+  curl -Lo glibc-bin.apk "https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-bin-${GLIBC_VERSION}.apk" && \
+  echo "$GLIBCBIN_SHA256  glibc-bin.apk" | sha256sum -c - && \
+  apk add glibc-bin.apk glibc.apk && \
+  /usr/glibc-compat/sbin/ldconfig /lib /usr/glibc-compat/lib && \
+  echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
+  apk del curl && \
+  rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*
+
+ENV LIGHTNINGD_DATA=/root/.lightning
+ENV LIGHTNINGD_PORT=9735
```
NicolasDorier: Actually uneeded and can lead to confusion because this depends on the parameters passed to docker run.

Dockerfile
```diff
@@ -0,0 +1,103 @@
...
+    && wget -qO litecoin.tar.gz "$LITECOIN_URL" \
+    && echo "$LITECOIN_SHA256  litecoin.tar.gz" | sha256sum -c - \
+    && gpg --keyserver keyserver.ubuntu.com --recv-keys "$LITECOIN_PGP_KEY" \
+    && wget -qO litecoin.asc "$LITECOIN_ASC_URL" \
+    && gpg --verify litecoin.asc \
+    && BD=litecoin-$LITECOIN_VERSION/bin \
+    && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
+    && rm litecoin.tar.gz
+
+ENV LIGHTNINGD_VERSION=master
```
shesek: `LIGHTNINGD_VERSION` is not used for anything

shesek:
Should the entry point also start bitcoind if an external node isn't configured? Why are the bitcoin binaries included if not?

NicolasDorier:
@shesek because clightning depends on bitcoin-cli.

shesek:
Ah, I see, it's just the `-cli` binaries that are there, not that daemon itself. Would it make sense to bundle the daemon also?

NicolasDorier:
@shesek I don't think so. It makes the image two time bigger, also running two processes in one image is not recommended.

Dockerfile
NicolasDorier: EDIT: It is actually used for `socat`.

Dockerfile
jsarenik: There's backward compatibility in case someone would set environment variable and then expect things to work.

Dockerfile
shesek: Still somewhat confusing, maybe rename it to something like `RPC_TCP_PORT` to make it distinct from the P2P network port?

cdecker: Yeah, sorry, my bad for mixing those two up :-(

PR #1617, opened 2018-06-22
Add configure and .gitmodules to .gitattributes

NicolasDorier:
Without this, building the `dockerfile` on windows will likely fail.


cdecker:
ACK 55b100d

CI failure is unrelated

PR #1618, opened 2018-06-22
listpeers enhancements and suppress duplicates

rustyrussell:
I had these lying around; I'm cleaning up my branches so I figured it was time to push them.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -702,27 +702,18 @@ struct getpeers_args {
 };
 
 static void json_add_node_decoration(struct json_result *response,
-				     struct gossip_getnodes_entry **nodes,
-				     const struct pubkey *id)
+				     struct gossip_getnodes_entry *node)
 {
-	for (size_t i = 0; i < tal_count(nodes); i++) {
-		struct json_escaped *esc;
+	struct json_escaped *esc;
 
-		/* If no addresses, then this node announcement hasn't been
-		 * received yet So no alias information either.
-		 */
-		if (nodes[i]->addresses == NULL)
-			continue;
+	/* If node announcement hasn't been received yet, no alias information.
+	 */
+	if (node->last_timestamp < 0)
+		return;
 
-		if (!pubkey_eq(&nodes[i]->nodeid, id))
-			continue;
-
-		esc = json_escape(NULL, (const char *)nodes[i]->alias);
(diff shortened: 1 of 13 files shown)


cdecker:
Looks quite ok, except maybe the last commit. I don't get the logic behind deferring a list of updates when we can combine multiple updates into a single one. If we change the `base_fee` in update 1 and the `proportional_fee` in update 2, we can just create a new update that combines both changes, no need to actually queue two of them.

rustyrussell:
@cdecker I don't understand.  Updates already override, and don't combine.  So update 3 will always replace update 2.  (Not that we have a mechanism to change the fees we advertise yet). 

The logic in the last patch is simple: first, suppress obviously redundant updates.  Second, wait a while in case we see a redundant update, eg. disabling then re-enabling channel.

cdecker:
I don't really see the point of making it a list. Shouldn't we only ever have 1 pending update. If we already have an update waiting, and want to queue a new one, then we should amend the existing one, instead of queuing a new one.

rustyrussell:
We delete the previous pending one, and append a new one.  It does mean we restart the 15 second timer, though?

rustyrussell:
OK, this is simpler, and nicer IMHO.

jb55:
bisect shows compile error on f9b8237d500c0a2f8e8de171ba287760facda724 due to structeq update conflict?

```
gossipd/gossip.c: In function ‘find_local_update’:
gossipd/gossip.c:1702:7: error: implicit declaration of function ‘structeq’; did you mean ‘strtoq’? [-Werror=implicit-function-declaration]
   if (structeq(scid, &i->scid))
       ^~~~~~~~
       strtoq
```

cdecker:
Do we really need to re-enable the rebase check on github?

lightningd/peer_control.c
```diff
@@ -776,10 +777,18 @@ static void gossipd_getpeers_complete(struct subd *gossip, const u8 *msg,
 			json_array_end(response);
 		}
 
-		json_add_node_decoration(response, nodes, &p->id);
+		/* Search gossip reply for this ID, to add extra info. */
+		for (size_t i = 0; i < tal_len(nodes); i++) {
+			if (pubkey_eq(&nodes[i]->nodeid, &p->id)) {
```
jb55: I'm getting an invalid pointer here, so I crash in listpeers now. will poke around to try and figure out why...

PR #1629, opened 2018-06-26
[Docker] Fix: Bind socat to the right port

NicolasDorier:
A bug in the dockerfile has been introduced because of unclear `LIGHTNINGD_PORT` environment variable.

`socat` was trying to bind on the same port as lightning, while it should have been on a different one exposing the API by TCP. 

As such I renamed `LIGHTNINGD_PORT` to `LIGHTNINGD_RPC_PORT`, and made sure it is properly used.


NicolasDorier:
Renamed to `LIGHTNINGD_RPC_PORT` to be more explicit, also changed the documentation to precise RPC.

PR #1631, opened 2018-06-26
lightningd: don't send uninialized malformed fields to channeld.

rustyrussell:
==1224== Uninitialised byte(s) found during client check request
==1224==    at 0x152CAD: memcheck_ (mem.h:247)
==1224==    by 0x152D18: towire (towire.c:17)
==1224==    by 0x152DA1: towire_u16 (towire.c:28)
==1224==    by 0x142189: towire_failed_htlc (htlc_wire.c:29)
==1224==    by 0x16343F: towire_channel_init (gen_channel_wire.c:596)
==1224==    by 0x115C2C: peer_start_channeld (channel_control.c:249)
==1224==    by 0x131701: peer_connected (peer_control.c:503)
==1224==    by 0x117820: gossip_msg (gossip_control.c:182)
==1224==    by 0x139D97: sd_msg_read (subd.c:500)
==1224==    by 0x139676: read_fds (subd.c:327)
==1224==    by 0x179D52: next_plan (io.c:59)
==1224==    by 0x17A84F: do_plan (io.c:387)
==1224==  Address 0x1ffefffabe is on thread 1's stack
==1224==  in frame #2, created by towire_u16 (towire.c:26)

Followed by:

2018-06-18T21:53:04.129Z lightningd(1224): 03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134 chan #1: Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: received ERROR channel d0101486543e1a8b6871556a4fe1fba4ad4d83ce7f6f92919fd17bd1545d2fd5: UpdateFailMalformedHtlc message doesn't have BADONION bit set

### DIFF
--- a/common/htlc_wire.c
+++ b/common/htlc_wire.c
@@ -25,8 +25,11 @@ void towire_fulfilled_htlc(u8 **pptr, const struct fulfilled_htlc *fulfilled)
 
 void towire_failed_htlc(u8 **pptr, const struct failed_htlc *failed)
 {
+	/* Only one can be set. */
+	assert(failed->failcode || tal_len(failed->failreason));
+	assert(!failed->failcode || !tal_len(failed->failreason));
 	towire_u64(pptr, failed->id);
-	towire_u16(pptr, failed->malformed);
+	towire_u16(pptr, failed->failcode);
 	towire_u16(pptr, tal_count(failed->failreason));
 	towire_u8_array(pptr, failed->failreason, tal_count(failed->failreason));
 }
@@ -84,9 +87,12 @@ struct failed_htlc *fromwire_failed_htlc(const tal_t *ctx, const u8 **cursor, si
 	struct failed_htlc *failed = tal(ctx, struct failed_htlc);
 
 	failed->id = fromwire_u64(cursor, max);
-	failed->malformed = fromwire_u16(cursor, max);
+	failed->failcode = fromwire_u16(cursor, max);
 	failreason_len = fromwire_u16(cursor, max);
-	failed->failreason = tal_arr(failed, u8, failreason_len);
+	if (failreason_len)
+		failed->failreason = tal_arr(failed, u8, failreason_len);
+	else
+		failed->failreason = NULL;
 	fromwire_u8_array(cursor, max, failed->failreason, failreason_len);
 
 	return failed;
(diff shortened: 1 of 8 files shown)


lightningd/peer_htlcs.c
```diff
@@ -1384,6 +1384,7 @@ static void add_fulfill(u64 id, enum side side,
 }
 
 static void add_fail(u64 id, enum side side,
+		     enum onion_type malformed,
```
cdecker: I found the renaming from `failcode` to `malformed` rather hard to follow, could you add a comment to this?

rustyrussell: Actually, better: we should use the same terminology everywhere.  Let me create a patch...

rustyrussell:
Thanks to @cdecker I did a more thorough review, and so I have another bugfix as well as a much-needed cleanup!

common/htlc_wire.c
```diff
@@ -84,9 +87,12 @@ struct failed_htlc *fromwire_failed_htlc(const tal_t *ctx, const u8 **cursor, si
...
 
 	failed->id = fromwire_u64(cursor, max);
-	failed->malformed = fromwire_u16(cursor, max);
+	failed->failcode = fromwire_u16(cursor, max);
 	failreason_len = fromwire_u16(cursor, max);
-	failed->failreason = tal_arr(failed, u8, failreason_len);
+	if (failreason_len)
+		failed->failreason = tal_arr(failed, u8, failreason_len);
+	else
+		failed->failreason = NULL;
```
cdecker: Isn't `fromwire_u8_array` technically allowed to dereference `failed->failreason`, even though the length is 0 in this case? Can we more the call to `fromwire_u8_array` into the `if`-branch?

rustyrussell: Not quite: the argument would be that the compiler could assume the latter branch never happens because you can't hand "NULL" to memcpy (as done in fromwire).

But this fights with Kernighan and Plauger's "Make sure special cases are truly special".  We could, at best, hide it inside fromwire() and expicitly test for 0-length there, but TBH compilers don't currently actually assume the NULL-to-memcpy-is-impossible because, well, I am right :)

cdecker:
ACK ff03f14

PR #1636, opened 2018-06-27
Typesafe callback system for parsing json

wythe:
This is part of #1464 and incorporates Rusty's suggested updates from #1569.

See comment in param.h for description, here's the basics:

	unsigned cltv;
	const jsmntok_t *note;
	u64 msatoshi;
	struct param * mp;

	if (!param_parse(cmd, buffer, tokens,
			 param_req("cltv", json_tok_number, &cltv),
			 param_opt("note", json_tok_tok, &note),
			 mp = param_opt("msatoshi", json_tok_u64, &msatoshi),
			 NULL))
		return;

	if (param_is_set(mp))
		do_something()

There is a lot of developer mode code to make sure we don't make mistakes,
like trying to unmarshal into the same variable twice or adding a required param
after optional.

During testing, I found a bug (of sorts) in the current system.  It allows you
to provide two named parameters with the same name without error; e.g.:

	# cli/lightning-cli -k newaddr addresstype=p2sh-segwit addresstype=bech32
	{
		  "address": "2N3r6fT65PhfhE1mcMS6TtcdaEurud6M7pA"
	}

It just takes the first and ignores the second.  The new system reports this as an
error for now.  We can always change this later.


rustyrussell:
OK, here are some suggestions on top of this, not sure what you think:

https://github.com/rustyrussell/lightning/tree/guilt/pr-1636

wythe:
Good stuff.  And now that we don't need the clunky `param_is_set` anymore, we can `tal_free` the `param *` array at the end of `param_parse` and just return bool.

To your previous point, it would be nice for `json_tok` functions to return a struct containing the string and the error code. I think we agree here the goal is to get rid of the `fail_format` array.  However it does work well for now.

rustyrussell:
Good point!  OK, I'm having way to much fun bikeshedding this :)

Please cherry-pick the bits you want, and I'll merge.  Then the mass-conversion can begin!

wythe:
This is looking great! 

Double check what I just did and then merge at will.

rustyrussell:
Nice work!  Yes, the previous prototype issue with json_tok_tok is gone, which makes it much nider.

I think we finally got there.  Phew!

rustyrussell:
@wythe Are you planning the mass conversion, or should I start it?  Seems redundant to double-up.

wythe:
@rustyrussell I was planning on doing it.  I'll start tomorrow (about 8 hours from now).  Just shoot me a note if you start in on it though so we don't clobber each other.

wythe:
@rustyrussell We lost type safety in callbacks!

I didn't notice this until I made a mistake while migrating.  This error was introduced in c091ecccabe64625748aeb580f61a4f4ad45c076. 

I didn't notice anything obviously wrong. Hopefully you can.

You can easily reproduce this by just changing `u64 ival` to a `double ival` in the `bad_programmer()` unit test  and recompile. The compilation doesn't fail like it should.

PR #1639, opened 2018-06-28
Improve the output of decodemsg.

rustyrussell:
In particular, we need further decoding of some fields: alias, addresses and encoded_short_ids.


cdecker:
ACK 8883c2b

PR #1641, opened 2018-06-28
Documentation fixup

hannemennah:
Fixes for typos and minor oversights.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ It has the following environment variable:
 
 * `EXPOSE_TCP` default to false, if true, use expose c-lightning RPC on port 9835. (Use this only for testing)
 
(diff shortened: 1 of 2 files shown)


hannemennah:
The bumping down of the test-specific dependencies is to ensure that the instructions can be followed directly and in-order. The `sudo pip3 install -r tests/requirements.txt` command presupposes the existence of the tests/requirements.txt folder and file, which will only exist locally after the `git clone https://github.com/ElementsProject/lightning.git` (and `cd lightning`) commands have been run. 

As it stands, it's initially (and unnecessarily) unclear what "tests/requirements.txt" refers to. Relocating this line below the `git clone` section makes it unambiguous and straightforward.

PR #1644, opened 2018-06-29
fix #1638 - make ValueError exception a JSON encoded string

### LINKED ISSUE #1638: ValueError exception thrown by LIghtningRpc is extremely difficult to parse
When a ValueError is thrown, for instance when you're paying an invoice but the locktime exceeds the max delay, the resulting error is extremely difficult to parse.

Here's an example of one (I stripped it down to be easier to read):

`RPC call failed: {u'message': u'Delay (locktime) is 734 blocks; max delay requested is 500.', u'getroute_tries': 8}, method: pay, payload: {'bolt11': u'lnbc1...x'}`

It's essentially a single string with keys like "RPC call failed", "method" and "payload", but each of the values associated with those keys are string representations of a dict.

In order to use one of those dicts in a meaningful way, you'd have to extract that portion of the string out, then use something like the ast.literal_eval class.

I've tried to implement something to parse it [here](https://github.com/marzig76/fulmo/blob/13f748a82586e9555f945260725b5d07fc41c1a4/fulmo.py#L214), but it doesn't always work because the depth of nested dicts seems to change from case to case, making it hard to extract the exact portion of the string that is a dict.

If I'm going about this the wrong way, please let me know.  But I feel it would be much easier to use if the whole exception message was just a JSON object or even a single dict, which could then easily be used.

marzig76:
This will make LightningRpc exceptions much easier to work with

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import socket
-
+import json
 
 class UnixDomainSocketRpc(object):
     def __init__(self, socket_path, executor=None, logger=logging):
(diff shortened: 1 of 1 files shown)


marzig76:
Working with these exceptions today, I have to isolate the portion of the string that I'm looking for  (either the error, method or payload) with `str.find()`, then use `ast.literal_eval()` to turn that substring into a dict, then finally get access to the error with the dict key.  With this json solution, I can just do:
`json.loads(e)[error]`, which is a lot easier and cleaner.

But I do agree that your suggestion to subclass ValueError would be an ever better solution.

PR #1651, opened 2018-07-01
cli: add manpage hint at the end of human help output

jb55:
It may not immediately obvious that this is available


jb55:
alternatively, we can have `lightning-cli help <command>` launch the manpage, similar to what git does

cdecker:
ACK 621bc8a

PR #1657, opened 2018-07-03
cli: launch a manpage when help is passed an argument

jb55:
Lift it to the start of main, so that it can short-circuit the lightningd
connection. It does not need to be running for this to work.

example:

  lightning-cli help pay
  lightning-cli help cli

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -104,7 +104,7 @@ static void human_help(const char *buffer, const jsmntok_t *result) {
 		curr++;
 	}
 
-	printf("---\nrun `man lightning-{command}` for more information on each command\n");
+	printf("---\nrun `lightning-cli help <command>` for more information on a specific command\n");
 }
 
 enum format {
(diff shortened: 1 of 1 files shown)


jb55:
ah CI caught a bug, fixing

cli/lightning-cli.c
```diff
@@ -168,6 +168,15 @@ static void add_input(char **cmd, const char *input,
 		tal_append_fmt(cmd, ", ");
 }
 
+static void
+exec_man (const char *page) {
+	int res = execlp("man", "man", page, (char *)NULL);
+	bool ok = res != -1;
```
rustyrussell: Um.. execlp doesn't return if it succeds :)

You really want:

```C
        execlp("man", "man", page, (char *)NULL);
        err(1, "Running man command");
```

jb55:
> Um.. execlp doesn't return if it succeds :)

oops I read the manpage wrong. fixed.

PR #1661, opened 2018-07-04
structeq update/rewrite

rustyrussell:
@jb55 asked about padding: indeed, structeq() is unreliable if structures have padding, and while we take care not to have padding, it always makes me nervous.

Those comments, however, convinced me to rewrite ccan/structeq to be more robust, if a little harder to use.  It now generates a foo_eq() inline function, and handles padding correctly: in fact if padding exists in any of our structures compilation will fail now.


cdecker:
ACK f384205

CI failure was unrelated (see #1662).

ccan/ccan/structeq/structeq.h
```diff
@@ -1,17 +1,45 @@
...
+		return memcmp(_a, _b, sizeof(*_a)) == 0;		\
+	else								\
+		return CPPMAGIC_JOIN(&&,				\
+				     CPPMAGIC_MAP(STRUCTEQ_MEMBER_CMP_, \
+						  __VA_ARGS__));	\
+}
+
+/* Helpers */
+#define STRUCTEQ_MEMBER_SIZE_(m) sizeof((_a)->m)
+#define STRUCTEQ_MEMBER_CMP_(m) memcmp(&_a->m, &_b->m, sizeof(_a->m)) == 0
```
jb55: this fails on member structs with padding, correct? no need to spin your wheels on edge cases, just making sure I understand what's happening here.

ccan/ccan/structeq/structeq.h
```diff
@@ -1,17 +1,45 @@
...
- * different types!
+ * This generates a single memcmp() call in the common case where the
+ * structure contains no padding.  Since it can't tell the difference between
+ * padding and a missing member, @padbytes can be used to assert that
+ * there isn't any, or how many we expect.  -1 means "expect some", since
+ * it can be platform dependent.
  */
-#define structeq(a, b) \
-	(memcmp((a), (b), sizeof(*(a)) + 0 * sizeof((a) == (b))) == 0)
+#define STRUCTEQ_DEF(sname, padbytes, ...)				\
```
jb55: extraordinary.

wire/test/run-peer-wire.c
```diff
@@ -704,21 +703,22 @@ static bool channel_announcement_eq(const struct msg_channel_announcement *a,
...
+		&& short_channel_id_eq(&a->short_channel_id,
+				       &b->short_channel_id)
 		&& eq_between(a, b, node_id_1, bitcoin_key_2);
 }
 
 static bool funding_locked_eq(const struct msg_funding_locked *a,
 			      const struct msg_funding_locked *b)
 {
-	return structeq(a, b);
+	return memcmp(a, b, sizeof(*a)) == 0;
```
jb55: just curious why these aren't using structeq?

ccan/ccan/structeq/structeq.h
rustyrussell: Yes, that's why in that case we do a member-by-member comparison.

Of course, if the member itself is a struct with padding, we'll lose (you can actually lie in the caller in that case and refer to submembers, like 'foo.a, foo.b').

ccan/ccan/structeq/structeq.h
rustyrussell: That's why we don't look behind the curtain... :)

wire/test/run-peer-wire.c
rustyrussell: Theory of minimal change, and laziness due to this being test code.  Otherwise, yes, we could have just STRUCTEQ_DEF() these into existence.

PR #1669, opened 2018-07-06
Restore typechecking to param parsing

wythe:
While migrating to our new typesafe callback system for parameter parsing I
realized we no longer had typesafe callbacks!

It stopped working at c091ecccabe64625748aeb580f61a4f4ad45c076.  I'm guessing it
was a combination of the macros within va_list, but I really have no idea.

Anyway, I restored the feature by reintroducing the indirection of a function that
returns a temporary `param *` and then copying its contents into the internal array.

This also revealed that we *did* need the special `param_opt_tok` macro so
I readded it.

drive-by: renamed `last` to `p` in `param_add`.

### DIFF
--- a/lightningd/params.c
+++ b/lightningd/params.c
@@ -15,28 +15,24 @@ struct param {
 	size_t argsize;
 };
 
-static void param_add(struct param **params,
-		      const char *name, param_cb cb, void *arg,
-		      size_t argsize)
+struct param *param_add(const char *name, param_cb cb, void *arg,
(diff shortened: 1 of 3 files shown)


rustyrussell:
You were very polite in not pointing out that I'm an idiot :+1: 

But we can simply use sizeof() in this case, see #1673

wythe:
I think its just funny we both missed it and I didn't catch it until I was 3 files into the migration.

PR #1673, opened 2018-07-07
lightningd/params: fix typesafe check.

rustyrussell:
typesafe_cb isn't suitable here, as it is simply a conditional cast,
and the result is passed through '...' and doesn't matter.

Reported-by: @wythe

### DIFF
--- a/lightningd/params.h
+++ b/lightningd/params.h
@@ -1,7 +1,6 @@
 #ifndef LIGHTNING_LIGHTNINGD_PARAMS_H
 #define LIGHTNING_LIGHTNINGD_PARAMS_H
 #include "config.h"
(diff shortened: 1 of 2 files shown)


wythe:
tested ACK https://github.com/ElementsProject/lightning/pull/1673/commits/5fa62b240f460ade037a4c3f8eed6fb2e2cc1746

Excellent!

rustyrussell:
Ack 5fa62b240f460ade037a4c3f8eed6fb2e2cc1746

But weird to ack myself, but @wythe's ack is good enough for this fairly small change IMHO.

Plus, I've made @cdecker do too much review recently!!

cdecker:
At least I can still review code while at conferences, should be back productive in a few days :wink:

PR #1675, opened 2018-07-08
README.md: 18332 is default bitcoind rpc port for testnet.

yzernik:
The README currently says the docker-compose file will "expose litecoin's rpc interface on default ports 18332". This is confusing because I don't see litecoin being used anywhere.

It looks like 18332 is the default RPC port for bitcoind on testnet.


cdecker:
Good catch, it's a leftover from the old dockerfile that was showing the litecoind example

ACK 4a0ebb6

PR #1682, opened 2018-07-10
Migrating to param_parse

wythe:
This converts about 50% of the `json_get_params` calls over.

Please note my handling of default values for missing optional parameters.
For most cases I used the terniary operator, and that seemed to work well.
But for when the value was used more than once, I wrote a `param_set_if_null`
macro, and that worked ok.  I'm open to any better ideas of course.


lightningd/params.h
```diff
@@ -89,4 +90,16 @@ typedef bool(*param_cb)(const char *buffer, const jsmntok_t *tok, void *arg);
 		      (arg) + 0*sizeof(*(arg) == (jsmntok_t *)NULL),	\
 		      sizeof(const jsmntok_t *)
 
+/*
+ * Convenient way to set an optional @param to a @value if it wasn't set during
+ * parsing.
+ */
+#define param_set_if_null(ctx, param, value)              \
```
rustyrussell: Hmm, can we do something like param_opt_default() and get this into one line?

Seems to work?

     https://github.com/rustyrussell/lightning/tree/param-with-defaults

wythe: Yes it works great.  I have it running locally and here's what it looks like:
```C
if (!param_parse(cmd, buffer, params,
                 param_req("msatoshi", json_tok_tok, &msatoshi),
                 param_req("label", json_tok_tok, &label),
                 param_req("description", json_tok_tok, &desctok),
                 param_opt_default("expiry", json_tok_u64, &expiry, 3600),
                 param_opt_tok("fallback", &fallback),
                 param_opt_tok("fallbacks", &fallbacks),
                 param_opt_tok("preimage", &preimagetok), NULL))
        return;
```
What do you think of shortening the macro names? Something like:
```C
if (!param_parse(cmd, buffer, params,
                 p_req("msatoshi", json_tok_tok, &msatoshi),
                 p_req("label", json_tok_tok, &label),
                 p_req("description", json_tok_tok, &desctok),
                 p_opt_def("expiry", json_tok_u64, &expiry, 3600),
                 p_opt_tok("fallback", &fallback),
                 p_opt_tok("fallbacks", &fallbacks),
                 p_opt_tok("preimage", &preimagetok), NULL))
        return;
```

rustyrussell: Agreed, good idea. Rename param_parse to params() too?

wythe: Yes.

wythe: I can take care of all this if you like.

rustyrussell:
Hmm, this became impossible to review :(  But it's small enough now that you can pretty much rebase it into a single commit, like so:

https://github.com/rustyrussell/lightning/commit/e3f356d47d8d2813e701c5352590e221d6bd04c8

Issue #1526, opened 2018-05-26
Custom path for config file


Sjors:
Currently the configuration is assumed to be in `~/.lightning/config`. I'd like to be able to put a config file in `/etc/lightning/config` and have that file owned by a different user, while keeping the datadir in `~/.lightning`.

I think this is different from what's being discussed in #1204, which seems to be about access to Bitcoin's config.

rustyrussell:
Sjors Provoost <notifications@github.com> writes:
> Currently the configuration is assumed to be in `~/.lightning/config`. I'd like to be able to put a config file in `/etc/lightning/config` and have that file owned by a different user.
>
> I think this is different from what's being discussed in #1204, which seems to be about access to Bitcoin's config.

I'm tempted to add an include directive, and then make the default
<lightningdir>/config-<netname> (with fallback to just "config" if that DNE).

I think that's more flexible, and allows you to include some global
config file like this.

Thoughts?

Sjors:
That would still allow the lightning user to just point the include somewhere else though.

benthecarman:
This issue should be closed

PR #1683, opened 2018-07-11
 add --conf parameter

arowser:
#1526 Custom path for config file

### DIFF
--- a/lightningd/lightningd.h
+++ b/lightningd/lightningd.h
@@ -80,6 +80,8 @@ struct lightningd {
 	char *config_dir;
 	char *rpc_filename;
 
+	/* Configuration file name */
+	char *config_filename;
 	/* Configuration settings. */
 	struct config config;
 
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -339,6 +339,9 @@ static char *opt_set_locktime(const char *arg, u32 *u)
 
 static void config_register_opts(struct lightningd *ld)
 {
+	opt_register_early_arg("--conf=<file>", opt_set_talstr, NULL,
+		&ld->config_filename,
+		"Specify configuration file. Relative paths will be prefixed by lightning-dir location. (default: config)");
 	opt_register_noarg("--daemon", opt_set_bool, &ld->daemon,
 			 "Run in the background, suppress stdout/stderr");
 	opt_register_arg("--ignore-fee-limits", opt_set_bool_arg, opt_show_bool,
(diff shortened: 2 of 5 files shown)


lightningd/options.c
```diff
@@ -697,7 +700,12 @@ static void opt_parse_from_config(struct lightningd *ld)
 	char *argv[3];
 	int i, argc;
 
-	contents = grab_file(ld, "config");
+	if (ld->config_filename!= NULL) {
+		if (!path_is_file(ld->config_filename))
+			fatal("Configuration file: %s not exists.", ld->config_filename);
```
rustyrussell: This test is unnecessary?

lightningd/options.c
```diff
@@ -697,7 +700,12 @@ static void opt_parse_from_config(struct lightningd *ld)
...
+		contents = grab_file(ld, ld->config_filename);
```
rustyrussell: Do this by itself, then below...

lightningd/options.c
```diff
@@ -697,7 +700,12 @@ static void opt_parse_from_config(struct lightningd *ld)
...
+	} else
+		contents = grab_file(ld, "config");
 	/* Doesn't have to exist. */
 	if (!contents) {
 		if (errno != ENOENT)
```
rustyrussell: `if (errno != ENOENT || ld->config_filename != NULL)`

lightningd/lightningd.h
```diff
@@ -80,6 +80,8 @@ struct lightningd {
 	char *config_dir;
 	char *rpc_filename;
 
+	/* Configuration file name */
+	char *config_filename;
```
rustyrussell: You need to initialize this to NULL in lightningd.c new_lightningd(), too.

arowser: Thanks, fixed all

lightningd/options.c
```diff
@@ -697,10 +700,13 @@ static void opt_parse_from_config(struct lightningd *ld)
 	char *argv[3];
 	int i, argc;
 
-	contents = grab_file(ld, "config");
+	if (ld->config_filename!= NULL) {
```
cdecker: Nit: whitespace

lightningd/options.c
```diff
@@ -697,10 +700,13 @@ static void opt_parse_from_config(struct lightningd *ld)
...
+		contents = grab_file(ld, ld->config_filename);
+	} else
+		contents = grab_file(ld, "config");
 	/* Doesn't have to exist. */
 	if (!contents) {
-		if (errno != ENOENT)
+		if ((errno != ENOENT) || ld->config_filename != NULL)
```
cdecker: A comment here would be much appreciated: in the default config case we allow it not to exist, but if the config was specified on the command line it has to exist.

cdecker: Also inconsistent parentheses.

cdecker:
Squashing this into one commit would be nice :wink:

cdecker:
ACK 57e2c6d

PR #1686, opened 2018-07-12
Application connection

bitonic-cjp:
This is the first step for implementing the application-defined transaction handling feature. It's not quite finished yet, but, the perfect being the enemy of the good, I decided to make a pull request anyway, to have a place of discussion. Any constructive criticism is welcome (code style, bugs, noncompliances, design choices etc.).

# What does it do?
It opens up a socket in the lightningd directory, to which other applications can connect.

If lightningd receives an incoming transaction for which it doesn't know how to handle it, and if an application is connected to this socket, lightningd does a JSON-RPC function call over the socket. As input, the function accepts information about the incoming transaction, and as output, it returns the BOLT 4 error code (or 0 if the transaction is to be accepted). If no application is connected to the socket, lightningd refuses the incoming transaction with an appropriate error code.

The sendpay RPC command has been extended, so that elements in the "route" parameter can optionally contain a "realm" and/or "data" element. When provided, these override the data that would normally be included as the onion  message to a node. This is useful to trigger the above functionality in another node.

# Why?
This allows applications to atomically attach application-defined actions to a Lightning payment, in the same way that individual hops in Lightning are atomically attached to each other. Note that it is the application's responsibility to have a way to get the preimage for a payment. The preimage may or may not be generated/stored as an invoice in lightningd; it may even be the case that, initially, only some external party knows the preimage.

# Code overview
The number of commits may be a bit intimidating, but much of it is minor tweaks and rewriting when design decisions evolved. It's probably easier to read the total code change than to read its full history.

Changes in sphinx.* and onion.* are related to the need to include customized data in the onion blob: the assumption in the onion-making code that the standard "realm 0" format is always used is no longer (always) true. So, (de)serialization of the "realm 0" format is separated out into its own functions.

app_connection.* contains the socket-handling and -protocol code. The socket listener is handled with the ccan/io library, but when an application connects, the fd for that connection is stolen from the io library. The reason for this is that the current implementation is using synchronous I/O to do the RPC calls. Socket close by the application is detected whenever a write/read fails; additionally, when a new application tries to connect to the socket, we check whether the old connection was closed.

# Known issues
## The name
I don't like the name "app connection", which I used just about anywhere. It's simply not descriptive. Before changing it, I'd like to ask you for suggestions. An e-mail from @rustyrussell suggested to see it as an "approver", but in reality it's much more than that. Maybe it's more like a "custom router"?

## Code coverage
I think code coverage of the test is still pretty bad. Is there a way to get a code coverage report?

## Failing tests
Sometimes, tests fail on me, complaining about "Bad gossip order from ...". This is especially the case for the "test_forward" test, since I modified it (note: ~~the modified test should also pass without any of my other changes~~ edit: it does need the sendpay changes). Unfortunately I don't understand this problem, I don't know if it's my fault and how I could solve it.

## FIXMEs
See the code.

# Missing features

* The data passed to the application in the RPC function call currently only consists of the realm number. In practice, this is way too limited for useful applications. In the future, this is to be extended with the transaction hash, the incoming HTLC time-out, the incoming amount, the custom routing message and the remaining sphinx data.
* Currently, the interface is only activated on an unsupported (non-zero) realm number. In the future, this is to be extended to certain other cases, for instance a routing attempt to a non-existing channel.
* If the application accepts the incoming transaction (that is, its return value is 0 == no BOLT 4 errors), it has to either cancel it later, or provide the pre-image later. Currently, there is no way to do either; this currently stops the interface from being useful in practice. In the future, the plan is to extend the JSON RPC of lightningd to include functions for doing this.
* If the application blocks in an RPC call, lightningd blocks. In the future, it might be better to rewrite this as asynchronous I/O. It is TBD what the impact of this will be on the rest of the code. Note that the entry point for the new code is added in peer_accepted_htlc in peer_htlcs.c, which ultimately gets called when we get notified by a channeld about the receival of a commitsig on a new HTLC.
* Maybe on the wishlist: the ability to have multiple applications connected to us at once? But what should be the behavior in that case? Seems to be a very low-priority feature to me: use cases can also be met by a user-defined dispatcher script.


jb55:
Corné Plooy <notifications@github.com> writes:
> Any constructive criticism is welcome (code style, bugs, noncompliances, design choices etc.).

one thing that would make this easier to review is to have properly
formatted commit messages, they currently go right off my screen:

https://jb55.com/s/cfd7e93464352fc5.png
https://jb55.com/s/bf023e705ccf48eb.png

This should be a quick fix with a rebase -i, and mark everything with
reword (r).

jb55:
Corné Plooy <notifications@github.com> writes:

> The number of commits may be a bit intimidating, but much of it is
> minor tweaks and rewriting when design decisions evolved. It's
> probably easier to read the total code change than to read its full
> history.

Hmm would it be too much to ask to tidy up the commits with rebase -i
and fixup/squashing out of date commits? This looks a bit daunting to
review in its current state.

bitonic-cjp:
> one thing that would make this easier to review is to have properly formatted commit messages, they currently go right off my screen:

What would be the recommended maximum length?

> fixup/squashing out of date commits?

I could do that. Do we prefer commits that are optimized for understandability, even if that means rewriting history? From what I know, different projects have different policies on this.

In case of a rewrite, I could also do some other things:
* change the name of the feature
* maybe split into a sendpay pull request and a pull request for the socket functionality

Rewriting shouldn't be too much effort: after all, I can copy-paste most of it; I don't need to re-do the thinking. But first, I'd like to know if there are other comments: then I might process those as well.

bitonic-cjp:
Self-criticism: wouldn't it be better to have the application open a socket, and let lightningd connect to it, instead of the other way around? That would make more sense, since it's the application that is offering an RPC. Also, it would remove the listener from lightningd, and all the complexity of checking whether a previous connection was closed before accepting another one. Lightningd would inherently only ever have zero or one connection.

jb55:
> What would be the recommended maximum length?

Standard convention is that subject lines are typically max 72 chars 

This is usually followed by a body that describes the change in more
detail. 

> I could do that. Do we prefer commits that are optimized for
> understandability, even if that means rewriting history? From what I
> know, different projects have different policies on this.

yes, I was referring to the fact that you mentioned things have evolved
over time, and that older commits may be out of date. One of the things
that can be jarring during a review is if there are older patches that
have been re-written in later commits.

Also non-semantic changes such as whitespace can be squashed into their
respective commits.

Thanks! I'll try to test this out sometime this week.

bitonic-cjp:
Based on the discussion so far, I propose that this version of the code is not pulled. Instead, I will create two new pull requests, with the following changes:

* Separate pull requests for the sendpay change and for the socket connection (the second depends on the first)
* Rename from "app connection" to "custom router"
* Have the app create the socket, and lightningd connect to it
* Commit messages have a max 72 char subject line, optionally followed by a body (from what I understand, the two are separated by an empty line)
* Make sure there aren't any commits that aren't necessary for understanding the code (such as obsolete code or typo/whitespace fixes)

In the mean time, this pull request can be kept as a place for discussion. The new code will be heavily inspired / copy-pasted from the code here, so any comments here will probably continue to be relevant.

PR #1689, opened 2018-07-12
pylightning: add missing arguments to getroute

gallizoltan:
All getroute parameter could be used from python.


cdecker:
ACK f58ed1f

PR #1695, opened 2018-07-13
params: shortened names

wythe:
This is a cosmetic change only. No functional changes.

I shortened the names of macros and changed param_parse() to param().

Also went through params.h with a fine-toothed comb and updated the comments
to reflect the current API.

I wanted to change the files:

	params.c -> param.c
	params.h -> param.h
	run-params.c -> run->param.c

but that confused `git diff` for params.h so its best left for another PR.

I'm keeping #1682 updated locally with all these changes.

### DIFF
--- a/lightningd/params.c
+++ b/lightningd/params.c
@@ -269,10 +269,9 @@ static void check_params(struct param *params)
 }
 #endif
 
-static bool param_parse_arr(struct command *cmd,
-			    const char *buffer,
-			    const jsmntok_t tokens[],
-			    struct param *params)
+static bool param_arr(struct command *cmd, const char *buffer,
+		      const jsmntok_t tokens[],
+		      struct param *params)
 {
 #if DEVELOPER
 	check_params(params);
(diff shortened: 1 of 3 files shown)


lightningd/params.c
```diff
@@ -269,10 +269,9 @@ static void check_params(struct param *params)
 }
 #endif
 
-static bool param_parse_arr(struct command *cmd,
-			    const char *buffer,
-			    const jsmntok_t tokens[],
-			    struct param *params)
+static bool param_arr(struct command *cmd, const char *buffer,
```
rustyrussell: Renaming a static was probably overkill :)

rustyrussell:
Put the rename at the front of #1682 maybe: the 24-hour turnaround on my review must really suck for your momentum!!

lightningd/params.c
wythe: Haha I know but it was a search and replace rename, and then I decided to keep it.

PR #1697, opened 2018-07-14
pylightning: Don't attempt to deserialize after every read

### LINKED ISSUE #1681: [pylightning] listchannels() takes 4 minutes with 10000 channels

- call listchannels() with 10000 known channels
- wait for return

Observed:
- It takes about 4 minutes to return the data

Expected:
- equivalent listchannels CLI command takes less than 1 sec. It should take far less than 4 minutes for the listchannels() method to complete, even with son parsing involved.

Suggestions:
I see 2 potentials performances issues in `UnixDomainSocketRpc._readobj`:
- recv only read 1024 bytes per call while there is more than 4.8 MB to read from the socket. changing this value to 10240 reduces the duration to about 30 seconds. Bigger chunk sizes don't improve things much.
- after each recv, _readobj tries to parse the  growing buffer. Not sure about this but maybe we could improve performance by detecting the right {} balance before trying to parse json.

### `getinfo` output

```
{
  "id": "035a533b73cc430c29b4f823b5d83d504ba8cf3afcbaafbfde8bff90a2af8be0f2", 
  "port": 9735, 
  "address": [
    {
      "type": "ipv4", 
      "address": "176.186.114.42", 
      "port": 9735
    }
  ], 
  "binding": [
    {
      "type": "ipv4", 
      "address": "192.168.1.33", 
      "port": 9735
    }
  ], 
  "version": "v0.6", 
  "blockheight": 531293, 
  "network": "bitcoin"
}
```

cdecker:
We make use of the structure of the final read to decide when to deserialize.

Fixes #1681


rustyrussell:
Seems this tripped a timing issue in tests, fixed in #1704

cdecker:
Yep, we're tripping quite a few lately, I'll do a new pass over the tests and see if I can fix some more.

cdecker:
With #1704 merged the tests pass :-)

Issue #1698, opened 2018-07-14
At node start, first fee estimate always fails and defaults to default fee rate


SimonVrouwe:
### Issue and Steps to Reproduce
Every time I start the node, the log shows `No fee estimate for Immediate: basing on default fee rate` Was wondering if that is normal and is there a way to prevent this. I am working on exponential smoothing of fee estimates and because of this initial defaulting of the feerate to 40000 sat/kw it takes a while to smooth back to the _real_ fee estimates.

It only happens at startup.

```
2018-07-14T09:08:52.769Z lightningd(8395): No fee estimate for Immediate: basing on default fee rate
2018-07-14T09:08:52.601Z lightningd(8395): No fee estimate for Immediate: basing on default fee rate
2018-07-14T09:08:52.601Z lightningd(8395): No fee estimate for Slow: basing on default fee rate
2018-07-14T09:08:52.618Z lightning_gossipd(8404): Connected out for 02fd98ebd4cbedd1317d629d62d192cf943f0134c61464de48f1e232d861de4ef9
2018-07-14T09:08:52.677Z lightningd(8395): Adding block 531848: 0000000000000000002296ac6ae8df440d19039abbfeca61c3adf903d8b9523b
2018-07-14T09:08:52.767Z lightning_gossipd(8404): peer_out WIRE_INIT
2018-07-14T09:08:52.767Z lightningd(8395): lightning_channeld-02977901c53b5299c7641acf11d49a46f0957e9e3f8191e6171d1222486d872317 chan #5: init LOCAL: remote_per_commit = 03df1a4c265ae43e9c1a828b
be2dba5db5f9f857ed45e01c8b4ffceb866349ace8, old_remote_per_commit = 021dfd2e14f52a4a6eff41d89b6c41beffa44733c95dcd1d279930722eccbdcc47 next_idx_local = 118 next_idx_remote = 118 revocations_rece
ived = 117 feerates 1569/1569 (range 5000-400000)
2018-07-14T09:08:52.767Z lightningd(8395): lightning_channeld-02977901c53b5299c7641acf11d49a46f0957e9e3f8191e6171d1222486d872317 chan #5: peer_out WIRE_CHANNEL_REESTABLISH
2018-07-14T09:08:52.769Z lightningd(8395): No fee estimate for Immediate: basing on default fee rate
2018-07-14T09:08:52.769Z lightningd(8395): Immediate feerate 8486 (was 0)
[...]

PR #1699, opened 2018-07-14
improves exponential smoothing of feerate estimates

SimonVrouwe:
- fixes problem with polling interval > 150 * 0.9
- fixes log message `feerate hit floor` at every feerate change
- smoothed fee now reaches 90% of (exp weighted) fee estimates polled in last
120s, independent of polling interval
- only apply smoothing when effect > 10 percent so it doesn't correct forever
- fix indentation

NOTE: Currently when starting a node, the first fee estimate always is `default_fee_rate = 40000`. So it takes about 4min for smoothed fee to reach _normal_ values, see #1698. Not sure if that is a problem.


cdecker:
ACK 247b825

PR #1708, opened 2018-07-16
Move all secrets into the HSM

rustyrussell:
Based on the previous #1679.

Note that the HSM still doesn't do any real sanity-checking on what it's signing, but the interfaces are now there for it to do so.


hsmd/hsm_client_wire_csv
```diff
@@ -90,3 +91,16 @@ hsm_cupdate_sig_req,,cu,culen*u8
 hsm_cupdate_sig_reply,103
 hsm_cupdate_sig_reply,,culen,u16
 hsm_cupdate_sig_reply,,cu,culen*u8
+
+# Master asks HSM to sign a commitment transaction.
+hsm_sign_commitment_tx,5
+hsm_sign_commitment_tx,,peer_id,struct pubkey
+hsm_sign_commitment_tx,,channel_dbid,u64
+hsm_sign_commitment_tx,,tx,struct bitcoin_tx
```
cdecker: Shouldn't we be generating the tx on the `hsmd` side?

cdecker: Thinking about this some more, we should probably just pass in additional information to allow `hsmd` to sanity check the tx. Same goes for some of the other messages as well.

PR #1715, opened 2018-07-17
Sendpay custom onion

bitonic-cjp:
For more explanation and discussion, see pull request #1686 .

This is the first part of a replacement of #1686 : this contains the additional functionality in sendpay. The rest will come in a later pull request, and will depend on this functionality.

### DIFF
--- a/common/sphinx.h
+++ b/common/sphinx.h
@@ -94,7 +94,7 @@ struct route_step {
  */
 struct onionpacket *create_onionpacket(
 	const tal_t * ctx,
-	struct pubkey path[],
+	const struct pubkey path[],
 	struct hop_data hops_data[],
 	const u8 * sessionkey,
 	const u8 *assocdata,
--- a/common/sphinx.h
+++ b/common/sphinx.h
@@ -65,10 +65,7 @@ enum route_next_case {
  */
 struct hop_data {
 	u8 realm;
-	struct short_channel_id channel_id;
-	u64 amt_forward;
-	u32 outgoing_cltv;
-	/* Padding omitted, will be zeroed */
+	u8 per_hop_data[PAYLOAD_SIZE];
 	u8 hmac[SECURITY_PARAMETER];
 };
 
@@ -121,7 +118,6 @@ bool onion_shared_secret(
  * @ctx: tal context to allocate from
  * @packet: incoming packet being processed
  * @shared_secret: the result of onion_shared_secret.
- * @hoppayload: the per-hop payload destined for the processing node.
  * @assocdata: associated data to commit to in HMACs
  * @assocdatalen: length of the assocdata
  */
@@ -131,6 +127,36 @@ struct route_step *process_onionpacket(
 	const u8 *shared_secret,
 	const u8 *assocdata,
 	const size_t assocdatalen
+	);
+
+/**
+ * serialize_per_hop_data - Serialize the per_hop data (for realm 0)
+ *
+ * @channel_id: the short channel ID
+ * @amt_forward: the amount to forward
+ * @outgoing_cltv: the CLTV value of the outgoing HTLC
+ * @per_hop_data: the serialized data buffer
+ */
+void serialize_per_hop_data(
+	const struct short_channel_id *channel_id,
+	u64 amt_forward,
+	u32 outgoing_cltv,
+	u8 *per_hop_data
+	);
+
+/**
+ * deserialize_per_hop_data - Deserialize the per_hop data (for realm 0)
+ *
+ * @per_hop_data: the serialized data buffer
+ * @channel_id: the short channel ID
+ * @amt_forward: the amount to forward
+ * @outgoing_cltv: the CLTV value of the outgoing HTLC
+ */
+void deserialize_per_hop_data(
+	const u8 *per_hop_data,
+	struct short_channel_id *channel_id,
+	u64 *amt_forward,
+	u32 *outgoing_cltv
 	);
 
 /**
--- a/common/sphinx.h
+++ b/common/sphinx.h
@@ -81,9 +81,8 @@ struct route_step {
  *
  * @ctx: tal context to allocate from
  * @path: public keys of nodes along the path.
- * @hoppayloads: payloads destined for individual hosts (limited to
- *    HOP_PAYLOAD_SIZE bytes)
- * @num_hops: path length in nodes
+ * @hops_data: data destined for individual hosts.
+ *    hops_data[..].hmac will be filled in by this function.
  * @sessionkey: 32 byte random session key to derive secrets from
  * @assocdata: associated data to commit to in HMACs
  * @assocdatalen: length of the assocdata
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -998,6 +998,7 @@ static void json_sendpay(struct command *cmd,
 	size_t n_hops;
 	struct sha256 *rhash;
 	struct route_hop *route;
+	struct payment_route_data payment_route_data;
 	u64 *msatoshi;
 	const char *description;
 
@@ -1036,6 +1037,8 @@ static void json_sendpay(struct command *cmd,
 		n_hops++;
 	}
 
+	route_to_payment_route_data(cmd->ld, route, &payment_route_data);
+
 	if (n_hops == 0) {
 		command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Empty route");
 		return;
@@ -1056,7 +1059,8 @@ static void json_sendpay(struct command *cmd,
 		}
 	}
 
-	if (send_payment(cmd, cmd->ld, rhash, route,
+	if (send_payment_custom_route_data(cmd, cmd->ld, rhash,
+			payment_route_data, route,
 			 msatoshi ? *msatoshi : route[n_hops-1].amount,
 			 description,
 			 &json_sendpay_on_resolve, cmd))
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -17,6 +17,7 @@
 #include <lightningd/peer_htlcs.h>
 #include <lightningd/subd.h>
 #include <sodium/randombytes.h>
+#include <string.h>
 
 /*-----------------------------------------------------------------------------
 Internal sendpay interface
@@ -995,7 +996,7 @@ static void json_sendpay(struct command *cmd,
 {
 	const jsmntok_t *routetok;
 	const jsmntok_t *t, *end;
-	size_t n_hops;
+	size_t i, n_hops;
 	struct sha256 *rhash;
 	struct route_hop *route;
 	struct payment_route_data payment_route_data;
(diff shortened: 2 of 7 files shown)


lightningd/pay.c
```diff
@@ -624,54 +634,76 @@ bool wait_payment(const tal_t *cxt,
 	return cb_not_called;
 }
 
+static struct custom_route_data route_to_custom_route_data(
```
jb55: it's fairly unconventional to return structs as values from my experience, usually you would just pass in a pointer.

lightningd/pay.c
```diff
@@ -624,54 +634,76 @@ bool wait_payment(const tal_t *cxt,
...
+	struct lightningd* ld,
```
jb55: μ-nit: struct lightning *ld

lightningd/pay.c
```diff
@@ -624,54 +634,76 @@ bool wait_payment(const tal_t *cxt,
...
+
 /* Returns false if cb was called, true if cb not yet called. */
-bool
-send_payment(const tal_t *ctx,
+static bool
+send_payment_custom_route_data(const tal_t *ctx,
 	     struct lightningd* ld,
 	     const struct sha256 *rhash,
-	     const struct route_hop *route,
+	     struct custom_route_data route_data,
```
jb55: pass a pointer instead of a value here

lightningd/pay.c
```diff
@@ -30,6 +31,15 @@ struct sendpay_command {
 	void *cbarg;
 };
 
+/* data consumed by send_payment_custom_route_data */
+struct custom_route_data
```
jb55: nit: perhaps `payment_route_data`, `custom` is a bit ambiguous.

lightningd/pay.c
```diff
@@ -1006,6 +1054,46 @@ static void json_sendpay(struct command *cmd,
 		return;
 	}
 
+	/* Default (realm = 0) realm / hop_data, and other route data */
+	custom_route_data = route_to_custom_route_data(cmd->ld, route);
```
jb55: pass custom_route_data as a pointer to route_to_custom_route_data here instead of returning.

bitonic-cjp:
Yes, the first couple of commits is just refactoring, in preparation for the "feaure" commits:

aa216e4 adds the new feature
9d72d63 fixes an issue that would otherwise partially nullify the feature addition
c233849 adds a test that uses the new feature

bitonic-cjp:
I added a couple of commits that address @jb55 's suggestions. Cherry-pick what you like, or just merge it all.

jb55:
> Should I merge master into this pull request and resolve conflicts in the merge commit, rebase & force-push the pull request

yup!

lightningd/pay.c
cdecker: Well, we do pass back quite a few pointers to structs if we pass in a context we can allocate off.

lightningd/pay.c
```diff
@@ -625,62 +634,76 @@ bool wait_payment(const tal_t *cxt,
...
+	/* Copy channels used along the route. */
+	for (i = 0; i < n_hops; ++i)
+		ret.channels[i] = route[i].channel_id;
+
+	return ret;
+}
+
+/* Returns false if cb was called, true if cb not yet called. */
+static bool
+send_payment_custom_route_data(const tal_t *ctx,
```
cdecker: The name is a bit misleading imho, it should be the common entrypoint for injected and locally originating payments.

cdecker: Notice also that this function does quite a lot of book-keeping that should really only be done for payments that originated here. If we later inject an onion from outside we likely don't want it to show up in the `listpayments` output. These should be extracted in a wrapper instead, that is bypassed when injecting.

cdecker:
Seems ok to me, I'll let @rustyrussell do a code-style check.

lightningd/pay.c
jb55: right that would work too

lightningd/pay.c
bitonic-cjp: I think, for now, we can see this as a locally originating payment. The extended functionality is not related to an app-specific -> standard-lightning bridge on this node, but to enable a standard-lightning -> app-specific bridge on another node, downstream the route.

The app-specific -> standard-lightning bridge idea is an interesting line of thought. Maybe it should have its own RPC command, or maybe an option in sendpay to consider this routing of someone else's transaction instead of considering this our own transaction. For simplicity, let's not include that in this pull request.

bitonic-cjp:
I added a merge commit that resolves a conflict with recent changes in lightningd/pay.c in the master branch.

cdecker:
> I added a merge commit that resolves a conflict with recent changes in lightningd/pay.c in the master branch.

Generally speaking rebases are to be preferred since they avoid pulling in unrelated stuff from `master`

common/sphinx.h
```diff
@@ -133,6 +128,36 @@ struct route_step *process_onionpacket(
...
+ * @amt_forward: the amount to forward
+ * @outgoing_cltv: the CLTV value of the outgoing HTLC
+ */
+void deserialize_per_hop_data(
+	const u8 *per_hop_data,
+	struct short_channel_id *channel_id,
+	u64 *amt_forward,
+	u32 *outgoing_cltv
+	);
+
```
wythe: Here and in other places the style should be a little more like:
```C
void deserialize_per_hop_data(const u8 * per_hop_data,
                              struct short_channel_id * channel_id,
                              u64 * amt_forward,
                              u32 * outgoing_cltv);
```

jb55: I wonder if we should just have a clang-format linting step in travis...

cdecker: I'm running clang-format on patches that I create, but it's slow progress and having travis check it is really painful if the code mostly does its own thing.

bitonic-cjp: I was about to create a new commit to fix this, but then I saw the surrounding, old code was inconsistent w.r.t. this coding style. For instance, the process_onionpacket declaration does not conform to your style suggestion.

Instead of only fixing my new code, it might be better to make one commit that fixes all code. Including such a commit in this pull request would mess up the readability of this pull request, since you'd have a mix of functional and non-functional changes, obscuring the view on where the functional changes are.

So I agree it should be done, but I don't think it should be done right here right now. Let's make a separate pull request that makes the coding style consistent across the entire code base.

cdecker: Shall we maybe commit a `.clang-format` config to format all future code?

wythe: We could just copy what the linux kernel does:

https://www.kernel.org/doc/html/v4.17/process/clang-format.html

And here's their config file:

https://github.com/torvalds/linux/blob/master/.clang-format

I currently just use `indent -kr -i8 -l80` and then just touch-up the result by hand.

lightningd/pay.c
```diff
@@ -1048,13 +1049,51 @@ static void json_sendpay(struct command *cmd,
...
+					custom_route_data.hop_data[i].per_hop_data,
+					sizeof(custom_route_data.hop_data[i].per_hop_data))) {
+				command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				     "Route %zu invalid data",
+				     i);
+				return;
+			}
+		}
+	}
+
```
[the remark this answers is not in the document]
bitonic-cjp: If you use a union, then how does the code that reads the data know which of the two to use? If it always uses the generic one, then you implicitly depend on the way how the C compiler structures the route_hop_bitcoin struct for its serialization. I guess you *might* get away with that if you are very careful with struct alignment, endianness and so on, but do you really want to go that way?

In the rest of the code, I see all serialization being done explicitly, without depending on how the C compiler deals with structs/unions. To me, that seems like the more sensible / portable way of doing serialization.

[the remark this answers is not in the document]
bitonic-cjp: I'd like to add that I allow the caller to change realm number and data independently of each other, so the user might change the realm number to signal something, but still want to use the the realm 0 format, or keep realm value 0, but use a different data format. The second might not be very useful, except maybe for our own testing purposes (it allows us to inject completely invalid realm 0 data, and see how a node responds).

lightningd/pay.c
bitonic-cjp: I see the point you make about the layering violation, and it still isn't resolved in my later commits. Another thing I don't like about my approach is that this data gets tal_stolen when it gets passed to send_payment_custom_route_data. So I'm open for suggestions on how to improve this.

However, I'm not sure I fully understand what you suggest. Maybe it's easiest if you just write the code improvement yourself?

Side note: I was wondering how well tal combines with data allocated on the stack. I guess that when a stack-allocated object is used as a context, the child objects are not automatically going to be deallocated when the context object on the stack goes out of scope, right? This isn't C++ after all.

lightningd/pay.c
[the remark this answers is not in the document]
rustyrussell: I'm not suggesting using layout assumptions: the realm byte is the descriminator.  You check that.  You can call 'generic' 'unknown' if you prefer...

bitonic-cjp:
Here you have the long-awaited rebase. These should be the only changes compared to before the rebase:

* Adjusted to the new code in the master branch
* Merged some "fix" commits into the commits they fix, to simplify the code history

lightningd/pay.c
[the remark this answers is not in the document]
ZmnSCPxj: Sending a realm byte of 0 while having a hop with a different format would be confusing to BOLT-conforming implementations, so I think `realm` as discriminator, is quite fine.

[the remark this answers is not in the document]
bitonic-cjp: @ZmnSCPxj : If the caller passes BOLT-conforming realm 0 data, it still fails. You are really eliminating possibilities from the user; even if you don't think those possibilities will ever be useful, someone else will think they are useful to him, and try to use the API that way. It will cause unexpected failures, and maybe even a security issue for the user doing this.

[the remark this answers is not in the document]
ZmnSCPxj: If it is BOLT-conforming realm 0 data, then the `struct route_hop_bitcoin` will work and is perfectly appropriate.  Probably if realm is 0 then `sendpay` should check the route data in detail and see that it conforms, then build the `struct route_hop_bitcoin` correctly.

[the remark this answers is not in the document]
bitonic-cjp: That adds way too much complexity. Why do that? To be able to use the onion? Why not just do what the user asks for? The user explicitly overrides the default data formation, so apparently the user knows what he wants, and it's different from what we'd normally do.

Let the receiving node complain if it does not understand the message.

Besides: sending "wrong" realm 0 data is useful in security / bug testing. Having this as a feature adds to "swiss army knife" usability of C-Lightning.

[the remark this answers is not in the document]
ZmnSCPxj: Ah, I understand.

bitonic-cjp:
Yet another rebase.

bitonic-cjp:
@cdecker maybe you can remove the needs-rebase label? At least for now, it seems it doesn't need a rebase. We'll keep playing the add-label rebase remove-label game until this gets merged.

Trust me: this costs me a lot more effort than you. Some of these rebase changes are really non-trivial, and require me to understand how the surrounding code has been refactored.

PR #1735, opened 2018-07-23
tools: fix bash path in check-manpage.sh

jb55:
The command fails on my system, since I do not have a bash under /bin


jb55:
going to update the others as well...

PR #1736, opened 2018-07-23
doc: update python test requirements in HACKING

jb55:
Looks like this was out of date

### DIFF
--- a/doc/HACKING.md
+++ b/doc/HACKING.md
@@ -160,11 +160,11 @@ lightning funds.
 
 Testing
 -------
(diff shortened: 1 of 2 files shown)


jb55:
not sure this is right, sorry for the noise...

cdecker:
I quite liked the idea, there is no harm in installing more dependencies than necessary.

jb55:
yeah I changed my mind. I think it makes sense to have pytest in requirements.txt as well? or do you need to install it separately? I can't test because pip doesn't work on nixos.

PR #1740, opened 2018-07-24
Add descriptions to invoices and payments

### LINKED ISSUE #1696: Feature request: store descriptions for invoices/payments and expose them over RPC
Useful for wallet applications that want to make this information accessible to the user.

Note that descriptions are technically already saved for invoices as part of the bolt11 string and could be extracted from that in app-land, but it would be more convenient to get them directly from the rpc.

For outgoing payments there's no way to get them at all, short of the application maintaining a separate database with them. (for an application I'm currently working on, this means going from being stateless and solely depending on the RPC, to the application having its own db just for that.)

cdecker:
Fixes #1696


wallet/db.c
```diff
@@ -328,6 +328,8 @@ char *dbmigrations[] = {
     "INSERT OR IGNORE INTO blocks (height) VALUES ((SELECT MIN(first_blocknum) FROM channels));",
     "DELETE FROM blocks WHERE height IS NULL;",
     /* -- End of  PR #1398 -- */
+    "ALTER TABLE invoices ADD description BLOB;",
+    "ALTER TABLE payments ADD description BLOB;",
```
rustyrussell: TEXT?  We really can't handle anything but UTF-8 here.  To be SQL compliant we could use VARCHAR(639), but we're not SQL compliant elsewhere anyway...

wallet/wallet.h
```diff
@@ -108,6 +108,9 @@ struct wallet_payment {
 	struct secret *path_secrets;
 	struct pubkey *route_nodes;
 	struct short_channel_id *route_channels;
+
+	/* The description of the payment. Must support `tal_len` */
+	char *description;
```
rustyrussell: This may be const?

This patch could have been folded into where the changes are used, but that's not worth rebasing for.

wallet/wallet.h
```diff
@@ -442,6 +445,9 @@ struct invoice_details {
 	u64 paid_timestamp;
 	/* BOLT11 encoding for this invoice */
 	const char *bolt11;
+
+	/* The description of the payment. Must support `tal_len` */
+	char* description;
```
rustyrussell: Counter-intuitivvely, we put the space before the `*` here, like above.

wallet/wallet.c
```diff
@@ -1656,6 +1664,13 @@ static struct wallet_payment *wallet_stmt2payment(const tal_t *ctx,
 
 	payment->msatoshi_sent = sqlite3_column_int64(stmt, 10);
 
+	if (sqlite3_column_type(stmt, 11) != SQLITE_NULL) {
+		len = sqlite3_column_bytes(stmt, 11);
+		payment->description = tal_arr(payment, char, len);
+		memcpy(payment->description, sqlite3_column_blob(stmt, 11), len);
```
rustyrussell: payment->description = tal_strdup(payment, sqlite3_column_text(stmt, 11);

wallet/wallet.h
```diff
@@ -442,6 +445,9 @@ struct invoice_details {
 	u64 paid_timestamp;
 	/* BOLT11 encoding for this invoice */
 	const char *bolt11;
+
+	/* The description of the payment. Must support `tal_len` */
```
rustyrussell: Just make it a normal nul-terminated string.  It's less confusing, IMHO.

lightningd/pay.c
```diff
@@ -792,7 +793,11 @@ send_payment(const tal_t *ctx,
 	payment->path_secrets = tal_steal(payment, path_secrets);
 	payment->route_nodes = tal_steal(payment, ids);
 	payment->route_channels = tal_steal(payment, channels);
-	payment->description = NULL;
+	if (description != NULL)
+		payment->description = tal_dup_arr(payment, char, description,
+						   tal_len(description), 0);
```
rustyrussell: tal_strdup.

lightningd/payalgo.c
```diff
@@ -128,6 +128,9 @@ struct pay {
 
 	/* Whether we are attempting payment or not. */
 	bool in_sendpay;
+
+	/* The description from the bolt11 string */
+	char *description;
```
rustyrussell: const?

lightningd/payalgo.c
```diff
@@ -683,6 +687,8 @@ static void json_pay(struct command *cmd,
 	/* Start with no failures */
 	list_head_init(&pay->pay_failures);
 	pay->in_sendpay = false;
+	pay->description = tal_dup_arr(pay, char, b11->description,
+				       tal_len(b11->description), 0);
```
rustyrussell: `pay->description = b11->description;`

b11 is a tal child of pay, so there are no lifetime issues with this,and it handled NULL correctly.

wallet/invoices.c
```diff
@@ -115,6 +115,12 @@ static void wallet_stmt2invoice_details(const tal_t *ctx,
 	}
 
 	dtl->bolt11 = tal_strndup(ctx, sqlite3_column_blob(stmt, 9), sqlite3_column_bytes(stmt, 9));
+	if (sqlite3_column_type(stmt, 10) != SQLITE_NULL)
+		dtl->description =
+		    tal_strndup(ctx, sqlite3_column_blob(stmt, 10),
+				sqlite3_column_bytes(stmt, 10));
```
rustyrussell: `tal_strdup(ctx, sqlite3_column_text(stmt, 10))`

Side note: Really, this function and its callers are weird.  They should allocate and return dtl, and allocate all members with that as parent.

lightningd/pay.c
```diff
@@ -913,17 +919,20 @@ static void json_sendpay_on_resolve(const struct sendpay_result* r,
...
 	struct route_hop *route;
 	u64 *msatoshi;
+	const struct json_escaped *desc;
+	const char *description;
 
 	if (!param(cmd, buffer, params,
 		   p_req("route", json_tok_tok, &routetok),
 		   p_req("payment_hash", json_tok_tok, &rhashtok),
 		   p_opt("msatoshi", json_tok_u64, &msatoshi),
+		   p_opt_tok("description", &desctok),
```
wythe: Just a side note: Despite all our efforts with the new parameter parsing system, we still can't directly capture a simple string.  I'm open to any ideas.

wallet/db.c
cdecker: Good point, `VARCHAR(639)` would also not be supported by MySQL xD

wallet/invoices.c
cdecker: Agreed, the function is rather strange, but I kept it as is to keep changes minimal. Might be a good time to fix this up though.

lightningd/pay.c
cdecker: There is `json_tok_escaped_string` which does an ok job imho

cdecker:
Ready to re-review @rustyrussell  :-)

cdecker:
Investigating the crash failure in `channeld`.

cdecker:
Seems failures where related to #1756, removing WIP again.

wallet/wallet.c
```diff
@@ -1609,6 +1610,12 @@ void wallet_payment_store(struct wallet *wallet,
 					    payment->route_channels);
 	sqlite3_bind_int64(stmt, 9, payment->msatoshi_sent);
 
+	if (payment->description != NULL)
+		sqlite3_bind_blob(stmt, 10, payment->description,
+				  tal_len(payment->description), SQLITE_TRANSIENT);
```
rustyrussell: Don't use tal_len() here, as it's not well-defined for tal_strdup().  This works only because in developer mode we define CCAN_TAL_DEBUG and that *always* assigns a length when it allocates.  I will fix this today so it correctly asserts if you ask for tal_len/tal_count on a unlengthed object...

rustyrussell: ie. sqlite3_bind_text()

wallet/wallet.c
```diff
@@ -1662,6 +1669,11 @@ static struct wallet_payment *wallet_stmt2payment(const tal_t *ctx,
 
 	payment->msatoshi_sent = sqlite3_column_int64(stmt, 10);
 
+	if (sqlite3_column_type(stmt, 11) != SQLITE_NULL) {
+		payment->description = tal_strdup(payment, sqlite3_column_blob(stmt, 11));
```
rustyrussell: sqlite3_column_text()?

wallet/invoices.c
```diff
@@ -115,6 +115,11 @@ static void wallet_stmt2invoice_details(const tal_t *ctx,
 	}
 
 	dtl->bolt11 = tal_strndup(ctx, sqlite3_column_blob(stmt, 9), sqlite3_column_bytes(stmt, 9));
+	if (sqlite3_column_type(stmt, 10) != SQLITE_NULL)
+		dtl->description =
+		    tal_strdup(ctx, sqlite3_column_blob(stmt, 10));
```
rustyrussell: sqlite3_column_text()?

wallet/wallet.c
cdecker: I was not aware that there is this quirk, I was always under the impression that things allocated with `tal_*` or `tal` would be tagged with their length and would therefor be acceptable for `tal_len`.

rustyrussell: Yeah :( That's why I fixed it to be that way in my other PR. "If the guy with the PhD can't use your interface, maybe your interface sucks" :)

lightningd/pay.c
rustyrussell: Still needs to be tal_strdup!

lightningd/payalgo.c
```diff
@@ -683,6 +687,7 @@ static void json_pay(struct command *cmd,
 	/* Start with no failures */
 	list_head_init(&pay->pay_failures);
 	pay->in_sendpay = false;
+	pay->description = b11->description;
```
rustyrussell: We should set pay->description to either `desc` or `b11->description`, one of which must not be NULL.  Unfortunately we'll still have NULL descriptions when we load from the DB, but we can wrap that in COMPAT_V06:

```diff
diff --git a/Makefile b/Makefile
index 6c83a780..57a3131c 100644
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@ PIE_LDFLAGS=-pie
 endif
 
 ifeq ($(COMPAT),1)
-# We support compatibility with pre-0.6.
-COMPAT_CFLAGS=-DCOMPAT_V052=1
+# We support compatibility with 0.6, and even before.
+COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V06=1
 endif
 
 PYTEST_OPTS := -v -x
```

contrib/pylightning/lightning/lightning.py
```diff
@@ -275,13 +275,14 @@ def getinfo(self):
         """
         return self.call("getinfo")
 
-    def sendpay(self, route, payment_hash):
+    def sendpay(self, route, payment_hash, description=None):
```
rustyrussell: Maybe default to "" here instead of empty, since it really shouldn't be NULL.  Or we could insist on it and just break backwards compat :(

lightningd/pay.c
rustyrussell: Hmm, we could add a ctx arg to all json_tok_() and then pass cmd in, then make json_tok_escaped_string of form `bool (const tal_t *ctx, const char *buffer, const jsmntok_t *tok, char **str)` to match?

lightningd/pay.c
```diff
@@ -1026,10 +1029,23 @@ static void json_sendpay(struct command *cmd,
...
+	if (desctok) {
+		desc = json_tok_escaped_string(cmd, buffer, desctok);
+		if (!desc) {
+			command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				     "description '%.*s' not a string",
+				     desctok->end - desctok->start,
+				     buffer + desctok->start);
+			return;
+		}
+		description = json_escaped_unescape(cmd, desc);
```
rustyrussell: We need to fail if this returns NULL, which it absolutely can.

wallet/invoices.c
```diff
@@ -87,10 +87,12 @@ trigger_invoice_waiter_expire_or_delete(struct invoices *invoices,
...
-					struct invoice_details *dtl)
+static struct invoice_details *wallet_stmt2invoice_details(const tal_t *ctx,
+							   sqlite3_stmt *stmt)
 {
+	struct invoice_details *dtl = tal(ctx, struct invoice_details);
+	assert(stmt);
+	assert(dtl);
 	dtl->state = sqlite3_column_int(stmt, 0);
 
 	sqlite3_column_preimage(stmt, 1, &dtl->r);
```
rustyrussell: Need to replace every other occurrence of `ctx` with `dtl` in this function (annoyingly, github doesn't let me annotate them, since they're not in this patch).

lightningd/payalgo.c
cdecker: The null description is always possible because we can also insert a payment through the `sendpay` command, so I'm happy to keep the conditional display in the json result returned by `listpayments`.

cdecker: Actually not so sure we want to treat the description in the bolt11 the same way we treat the description the `description_hash` commits to. For one we seem to be forcing users to provide a matching description to the hash in order to process a payment at all, this might be a bit cumbersome. Furthermore we currently assume that the description is a UTF-8 string, which might not be the case, e.g., if it commits to a PDF invoice. The spec specifies that the description `d` must be a UTF-8 string, so that is safe, however it does not say that the description hash `h` also commits to a string (which is functionality that I'd like to keep btw).

So simply using `desc` is probably not what we want, since it might not actually be a UTF-8 string, would require hex encoding to be usable in JSON, and wouldn't make much sense to a user in `listpayments` anyway.

lightningd/pay.c
wythe: Yes that would be helpful in many places, including `json_tok_label`. It would go a long way in helping us remove usage of `json_tok_tok`.

lightningd/payalgo.c
rustyrussell: OK, I'm persuaded.  Deeply tempted to drop this functionality altogether, but not 100% sure we won't want to back, so let's keep.

rustyrussell:
Rebased, mainly due to conflicts with deprecated removal.

Ack bd4c5b80eaedaf59ce4f7853ffcdb387cc94ba28

shesek:
Works for me. Thanks!

shesek:
Actually, it appears like descriptions are returned in reply to `listpayments`, `pay` and `waitanyinvoice`, but not in `listinvoices`. @cdecker

cdecker:
We don't actually backfill old invoices, but new ones should show a description. Backfilling would have required hooking into the migrations to call a conversion function.

shesek:
It looks like the description is available when fetching a specific invoice (`listinvoices <label>`), but not when fetching a list of them.

With a fresh node running on current master:

```
$ lightning-cli getinfo | jq -r .version
v0.6-218-ga039630

$ lightning-cli invoice 5000 a a
{
  "payment_hash": "cbca0a37cad184707a289cb97cad07b52981a1790a6c3c635eaf73c604a8330a", 
  "expires_at": 1533248517, 
  "bolt11": "lnbcrt50n1pdkxa04pp5e09q5d726xz8q73gnjuhetg8k55crgtepfkrcc674aeuvp9gxv9qdqzvycqp2ezd9d3xryfv9588qu7y5znl2xjzd8rmdhwv6japsuepv6xe47jyjmguupr4kzwgn4mmwunckm5dsj9ctavz03yznmwtvgd43mgefd9gqr3u6gk"
}

$ lightning-cli listinvoices
{
  "invoices": [
    {
      "label": "a", 
      "bolt11": "lnbcrt50n1pdkxa04pp5e09q5d726xz8q73gnjuhetg8k55crgtepfkrcc674aeuvp9gxv9qdqzvycqp2ezd9d3xryfv9588qu7y5znl2xjzd8rmdhwv6japsuepv6xe47jyjmguupr4kzwgn4mmwunckm5dsj9ctavz03yznmwtvgd43mgefd9gqr3u6gk", 
      "payment_hash": "cbca0a37cad184707a289cb97cad07b52981a1790a6c3c635eaf73c604a8330a", 
      "msatoshi": 5000, 
      "status": "unpaid", 
      "expires_at": 1533248517
    }
  ]
}

$ lightning-cli listinvoices a
{
  "invoices": [
    {
      "label": "a", 
      "bolt11": "lnbcrt50n1pdkxa04pp5e09q5d726xz8q73gnjuhetg8k55crgtepfkrcc674aeuvp9gxv9qdqzvycqp2ezd9d3xryfv9588qu7y5znl2xjzd8rmdhwv6japsuepv6xe47jyjmguupr4kzwgn4mmwunckm5dsj9ctavz03yznmwtvgd43mgefd9gqr3u6gk", 
      "payment_hash": "cbca0a37cad184707a289cb97cad07b52981a1790a6c3c635eaf73c604a8330a", 
      "msatoshi": 5000, 
      "status": "unpaid", 
      "description": "a", 
      "expires_at": 1533248517
    }
  ]
}
```

cdecker:
Got it, the SQL query was wrong. I'll file a PR with the fix tomorrow.

Issue #1730, opened 2018-07-22
temporary_channel_failure wrong format together with lnd


robtex:
getting garbled error messages when routing thru clightning.
lnd devs claims it is a c-lightning bug.
can you have a look please? 

https://github.com/lightningnetwork/lnd/issues/1599

cdecker:
We disagree on whether to include the message type or not. Continuing discussion on lightningnetwork/lnd#1599 and maybe moving to the spec repo for resolution.

PR #1742, opened 2018-07-25
Fix channel_update messages, especially inside errors.

rustyrussell:
#1730 lead me to start investigating (and testing!) channel_update.  Indeed, we get it completely wrong (though not malformed).

~~~Note that first patch is #1756 which should make Travis happier.~~~ (Now rebased)


rustyrussell:
Failed check-python due to extra whitespace; trivially rebased and fixed that now.

PR #1752, opened 2018-07-25
at startup initialize smoothed feerate to polled feerate

SimonVrouwe:
Fixes problem of unrealistic feerate smoothing at startup when a call to `get_feerate` happened before any _actual_ feerate poll.
Fixes most of #1698
Uses variable `topo->startup` because it seemed available

Will test on my `mainnet` node when it is finished syncing.


SimonVrouwe:
Perhaps the log messages can be improved to better see what is going on with the feerates, I will push something later. On my system all tests pass, so I don't understand why not on Travis. Even ran a docker container with the image used by Travis's failing test and that passed here also.

Testing on my mainnet node now.

PR #1758, opened 2018-07-26
exchanged code containing goto with a do while loop

renepickhardt:
did not find anything in the coding style which would require the use of goto therefor I thought it can be exchanged with a `do {...} while()` statement

### DIFF
--- a/lightningd/watch.c
+++ b/lightningd/watch.c
@@ -270,19 +270,17 @@ void watch_topology_changed(struct chain_topology *topo)
 	struct txwatch_hash_iter i;
 	struct txwatch *w;
 	bool needs_rerun;
+	do{
+		/* Iterating a htable during deletes is safe, but might skip entries. */
+		needs_rerun = false;
+		for (w = txwatch_hash_first(&topo->txwatches, &i);
+	    	w;
+	    	w = txwatch_hash_next(&topo->txwatches, &i)) {
+				u32 depth;
 
-again:
-	/* Iterating a htable during deletes is safe, but might skip entries. */
-	needs_rerun = false;
-	for (w = txwatch_hash_first(&topo->txwatches, &i);
-	     w;
-	     w = txwatch_hash_next(&topo->txwatches, &i)) {
-		u32 depth;
-
-		depth = get_tx_depth(topo, &w->txid);
-		if (depth)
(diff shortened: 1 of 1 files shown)


lightningd/watch.c
```diff
@@ -270,19 +270,17 @@ void watch_topology_changed(struct chain_topology *topo)
...
+		for (w = txwatch_hash_first(&topo->txwatches, &i);
+	    	w;
+	    	w = txwatch_hash_next(&topo->txwatches, &i)) {
+				u32 depth;
+
+				depth = get_tx_depth(topo, &w->txid);
+				if (depth)
+					needs_rerun |= txw_fire(w, &w->txid, depth);
+		}
+	}while (needs_rerun);
```
cdecker: Nit: whitespace befor the `while` is missing, and after the `do`.

lightningd/watch.c
```diff
@@ -270,19 +270,17 @@ void watch_topology_changed(struct chain_topology *topo)
...
-			needs_rerun |= txw_fire(w, &w->txid, depth);
-	}
-	if (needs_rerun)
-		goto again;
+	do {
+		/* Iterating a htable during deletes is safe, but might skip entries. */
+		needs_rerun = false;
+		for (w = txwatch_hash_first(&topo->txwatches, &i);
+	    	w;
+	    	w = txwatch_hash_next(&topo->txwatches, &i)) {
```
wythe: nit: Above two lines should be tabbed over.  They are part of the for loop.

wythe:
Yes this is a nice improvement.

lightningd/watch.c
cdecker: Looks like they mix tabs and spaces

renepickhardt: I can quickly fix this again but in my editor it looked pretty good

cdecker: I know, but it doesn't in other editors that display tabs as 2, 4 or 8 spaces. Please configure your editor to use tabs instead :-)

cdecker:
While we don't have a style-guide of our own we tend to rely heavily on the linux kernel guide (literally the first piece of information I got from @rustyrussell when I joined the project btw :wink:): https://www.kernel.org/doc/html/v4.10/process/coding-style.html

Issue #1738, opened 2018-07-24
Lost UTXO (P2WKH) after (concurrent?) unilateral close and not claimed by c-lightning


renepickhardt:
my channel partner used eclair mobile and created a channel with me and paid an invoice which I have created. After a couple of days and a peer connection loss both sides unilaterally closed the channel. My channel partner seemed to be successful. However I cannot redeem my output (P2WKH) from his confirmed commitment transaction (though it should not be time locked) also even stranger my node directly was able to claim his funds (P2WSH). 

Question: How can I look at the sqlite db (or my logs) to see which derivation of my key should be used in order to spend the "lost" UTOX. 

My lightning node: 
```
 036f464b54416ea583dcfae3872d28516dbe85414ed838513b1c34fb3a4aee4e7a@144.76.235.20:9735 
```

I created an invoice which was paid (json): 
```
{
      "label": "bitcoitechweekly2nd", 
      "bolt11": "lnbc4543210n1pdnlydqpp5u69x06u7d09slyjg69kdnx4xgm346ghgmmect8y0z5z4g5aeuedsdp4va6k2um5g9e8g6trd3j57mjvd9nksarwd9hxwnn9w3mk7unt235hqxqy2ljqcqpgsu9n53xcfsl95t2yxkh00np8rasxn6cfzkjf7lljnw9drwkxj47pzl9ml5uk7ypav20n8y8p8r26x38zmdewdrvc7zjdr3yazklksyqp64p4sq", 
      "payment_hash": "e68a67eb9e6bcb0f9248d16cd99aa646e35d22e8def3859c8f15055453b9e65b", 
      "msatoshi": 454321000, 
      "status": "paid", 
      "pay_index": 9, 
      "msatoshi_received": 454321000, 
      "paid_timestamp": 1530902807, 
      "paid_at": 1530902807, 
      "expiry_time": 1531252704, 
      "expires_at": 1531252704
    }, 
```

The channel partner created a channel with a founding of 500000 satoshi and paid the invoice. Before the channel closed the channel balance according to c-lighting looked like this: 

```
  {
      "peer_id": "038c04b4c0dcaa2d315a452309f3251d271c5c7d7b08ec7070de355a0fc8dae4fa",
      "short_channel_id": "530765:2414:0",
      "channel_sat": 454321,
      "channel_total_sat": 500000,
      "funding_txid": "3a5cb1352889e405f27074151dd9641a6d6c5944d21546cf19f805e0420dffae"
    }
```

[...]

PR #1764, opened 2018-07-27
added a test to fix the enum wallet_payment_status 

renepickhardt:
according to the comment at: 
https://github.com/ElementsProject/lightning/blob/master/wallet/wallet.h#L83

I understand that the values for the constants of this enum are critical in the communication with the data storage and should thus be fixed by a unit test.

### DIFF
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -9,6 +9,11 @@
 
 #define DB_FILE "lightningd.sqlite3"
 
+/* For testing, we want to catch fatal messages. */
+#ifndef db_fatal
+#define db_fatal fatal
+#endif
+
 /* Do not reorder or remove elements from this array, it is used to
  * migrate existing databases from a previous state, based on the
  * string indices */
@@ -361,7 +366,7 @@ void db_assert_no_outstanding_statements(void)
 
 	dbstat = list_top(&db_statements, struct db_statement, list);
 	if (dbstat)
-		fatal("Unfinalized statement %s", dbstat->origin);
+		db_fatal("Unfinalized statement %s", dbstat->origin);
 }
 
 static void dev_statement_start(sqlite3_stmt *stmt, const char *origin)
@@ -408,7 +413,7 @@ sqlite3_stmt *db_prepare_(const char *location, struct db *db, const char *query
 	err = sqlite3_prepare_v2(db->sql, query, -1, &stmt, NULL);
 
 	if (err != SQLITE_OK)
-		fatal("%s: %s: %s", location, query, sqlite3_errmsg(db->sql));
+		db_fatal("%s: %s: %s", location, query, sqlite3_errmsg(db->sql));
 
 	dev_statement_start(stmt, location);
 	return stmt;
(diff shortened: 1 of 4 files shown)


rustyrussell:
We should probably make these BUILD_ASSERT() statements where they're used, in db/wallet.c.

There should be matching ones for the others, too.

renepickhardt:
not sure if I get your right: 
1. matching ones for others means that you suggest to also fix the api of other constants / enums via test cases? 
2. I just read about BUILD_ASSERT() in the ccan package. What is the advantage of using them and move the asssertion the wallet.c where the enum is being used? I imagin a scenario in which a coder uses the enums at another position and later (for what ever reason) the original use of the API gets deleted together with those BUILD_ASSERT statments. Then the API would become unstable again. What I am trying to say: isn't the entire idea of using seperate test cases to prevent situations like the one described above?

rustyrussell:
Good point, and it turns out that's far from the only enum we put into the db.

I had to do some test fixups, but I think this covers almost all the bases:

https://github.com/rustyrussell/lightning/tree/db-enum-asserts

Thoughts?

renepickhardt:
As mentioned I am new to the codebase and would need more time to study your patch in order to give more meaningful feedback. I am pretty busy this week so I might not be able to do this quickly. 

On a high level I think I see what you did there and where you are going. I don't think there is a problem with having a test case like I did and your tests (since as far as I currently understand you test on the database side that it interprets the values correctly). If you wish I can merge your changes to my PR and we can merge this PR and close it for now and I would try to provide more feedback at a later point in time.

rustyrussell:
Sure, this is one case where I'm happy for duplication.  Ideally we'd simply annotate these enums and have a magic tool ensure they don't change across commits...

renepickhardt:
Done with merging both patches. You can merge to master and close if you wish.

renepickhardt:
I was looking closer at wallet.h and wallet.c to recover my funds from #1738 what I don't get: how do you (@rustyrussell ) make sure coders will use `static inline enum output_status output_status_in_db(enum output_status s)` instead of addressing the constants from the enum right away?

rustyrussell:
You can't *make* people use it, unless we wrap these in a struct type (which is a lot of overhead for marginal gain).  But even if they don't use it for runtime checking, its mere existence means those BUILD_ASSERT()s are making sure they don't change the enum values.

rustyrussell:
Hmm, you should have probably just cherry-picked my two changes: github is fairly confused right now :)

I don't think there were any significant changes, so I just rebased the 4 commits onto master.

rustyrussell:
Travis got stuck somehow, doing trivial rebase to try to unstick it...

cdecker:
Nope, travis is complaining about "Abuse detected", I reached out to their support to get more details about this.

rustyrussell:
OK, testing locally, will override and merge if all OK.

@renepickhardt BTW, I fixed this, too:

```C
static bool test_wallet_payment_status_enum()
{
```

Needs to be `(void)` otherwise it's considered a pre-ANSI C prototype...

rustyrussell:
Sorry, my commits were completely broken (valgrind noticed that test/run-wallet no longer cleaned up).

I've fixed it a different way, and will open a new PR to fix travis.

cdecker:
Duplicate of #1814

PR #1765, opened 2018-07-27
pay: Exempt a fee from the maxfeepercent rule it is still tiny

### LINKED ISSUE #1746: maxfeepercent should be ignored for small payments e.g. < 10 sat
Can't send payments of 1 satoshi to some expermental online games
https://github.com/ElementsProject/lightning/blob/f85084948678177199cf44bdb9f991cfba3edf2a/lightningd/payalgo.c#L438
Change that to something like
`fee_too_high = (feepercent > pay->maxfeepercent && fee > minfee);`

Where `minfee` default to 2000 millisatoshi. Maybe even add minfee in pay RPC call.

Would help on sites requiring 1 sat for access to a game when everyone is defaulting to a min of 1 satoshi fee for HTLC even C-Lightning

Thanks

cdecker:
Several users have noticed that they cannot pay satoshis.place or similar places
that have tiny payment amounts if they are not directly connected. This is due
to the forwarding fee dominating the transferred amount.

This commit adds a new option, exempting tiny fees (up to 5 satoshis by default)
from having to pass the maxfeepercent flag. While we could have told users to
tweak maxfeepercent I think it is usefull to have a default exemption.

Not sure about the name of the option just yet, the current naming is based on
exempting it from the check, which feels clunky. Maybe someone else has a
good idea?

Fixes #1746 

Ping @SerafinTech, @renepickhardt, @jonathancross

### DIFF
--- a/lightningd/payalgo.c
+++ b/lightningd/payalgo.c
@@ -711,6 +711,7 @@ static const struct json_command pay_command = {
 	"{description} (required if {bolt11} uses description hash), "
 	"{riskfactor} (default 1.0), "
 	"{maxfeepercent} (default 0.5) the maximum acceptable fee as a percentage (e.g. 0.5 => 0.5%), "
+	"{exemptfee} (default 5000 msat) disables the maxfeepercent check for fees below the threshold, "
 	"{retry_for} (default 60) the integer number of seconds before we stop retrying, and "
 	"{maxdelay} (default 500) the maximum number of blocks we allow the funds to possibly get locked"
 };
--- a/lightningd/payalgo.c
+++ b/lightningd/payalgo.c
@@ -128,6 +128,11 @@ struct pay {
 
 	/* Whether we are attempting payment or not. */
 	bool in_sendpay;
+
+	/* Maximum fee that is exempted from the maxfeepercent computation. This
+	 * is mainly useful for tiny transfers for which the leveraged fee would
+	 * be dominated by the forwarding fee. */
+	u64 exemptfee;
 };
 
 static struct routing_failure *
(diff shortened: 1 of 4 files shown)


lightningd/payalgo.c
```diff
@@ -612,6 +618,7 @@ static void json_pay(struct command *cmd,
 		   p_opt_def("retry_for", json_tok_number, &retryfor, 60),
 		   p_opt_def("maxdelay", json_tok_number, &maxdelay,
 			     cmd->ld->config.locktime_max),
+		   p_opt_def("exemptfee", json_tok_number, &exemptfee, 5000),
```
renepickhardt: does that also make the argument available as an option in the config file?

renepickhardt:
thanks for pinging me! that really helped (especially after I already started looking at `pay.c` and `payalgo.c`.  besides not being able to judge for side effects (which I find highly unlikely this particular change) this patch looks valid to me.

lightningd/payalgo.c
```diff
@@ -435,7 +440,7 @@ static void json_pay_getroute_reply(struct subd *gossip UNUSED,
 	 * That loss will not be representable in double. So, it's Okay to
 	 * cast u64 to double for feepercent calculation. */
 	feepercent = ((double) fee) * 100.0 / ((double) pay->msatoshi);
-	fee_too_high = (feepercent > pay->maxfeepercent);
+	fee_too_high = (fee <= pay->exemptfee && feepercent > pay->maxfeepercent);
```
SerafinTech: I think it should be `fee_too_high = (fee > pay->exemptfee && feepercent > pay->maxfeepercent);`

`fee_too_high` should only be true if it is more than the `pay->exemptfee`

cdecker: Doh :+1:

lightningd/payalgo.c
wythe: No. This only deals with json parameters. `config` and command line parameters are handled in `lightningd/options.c`.

wythe:
Also add the description to the ` static const struct json_command pay_command ` help.

rustyrussell:
Ack dddf5caf9278b7af49ecf2964ae2a1327a49ea5d

PR #1766, opened 2018-07-27
added (or remotely) to README

JBaczuk:
I was confused at first as to whether or not you can connect to a remote Bitcoin full node. This clarifies that you can in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ mailing list [lightning-dev@lists.linuxfoundation.org][ml2].
 ## Getting Started
 
 c-lightning currently only works on Linux (and possibly Mac OS with some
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ mailing list [lightning-dev@lists.linuxfoundation.org][ml2].
 ## Getting Started
 
 c-lightning currently only works on Linux (and possibly Mac OS with some
-tweaking), and requires a locally running `bitcoind` (version 0.15 or
+tweaking), and requires a locally (or remotely) running `bitcoind` (version 0.15 or
```
wythe: nit: Parentheses are not needed here. But then they are not needed in the above "and possibly Mac OS..." line either. This is a very good addition though.

PR #1770, opened 2018-07-28
Remove sigaction from run-param test

### LINKED ISSUE #1693: tests: run-params Segmentation fault on musl

On a musl-libc system run following:

```
git clone https://github.com/jsarenik/lightning-auto-test
cd lightning-auto-test
./run.sh https://github.com/wythe/lightning fix-param-arg-size
```

See `log`. Should be similar to following:

  * [x86_64 Alpine Linux](https://gist.github.com/jsarenik/d7d6ae798c38c1b7809274cf1eff9322#file-auto-test-alpine-x86_64-txt-L1740), [strace of run-params](https://gist.github.com/jsarenik/d7d6ae798c38c1b7809274cf1eff9322#file-strace-log-x86_64-musl-txt)

The same happens also on other architectures (tested aarch64 but I do not have the log handy).

wythe:
This should fix #1693. The `sigaction` code was not working for musl devices for some reason. So I just removed all the DEVELOPER mode `asserts` from `param.c` and instead just returned an error.

Then I removed all the `sigaction/setjump/longjmp` shenanigans from `run-param.c`.

It's less code and easier to read.  @jsarenik can test when he gets a chance. But I think it is a worthwhile
patch regardless.

### DIFF
--- a/lightningd/jsonrpc_errors.h
+++ b/lightningd/jsonrpc_errors.h
@@ -15,6 +15,7 @@
  * with a specific error code, and then removed.
  */
 #define LIGHTNINGD                      -1
+#define LIGHTNINGD_INTERNAL             -2
 
 /* Errors from `pay`, `sendpay`, or `waitsendpay` commands */
 #define PAY_IN_PROGRESS			200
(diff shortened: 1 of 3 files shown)


jsarenik:
[Full log, x86_64](https://gist.github.com/jsarenik/3fc5808a71d82012bb7314116a587c7e)

PR #1775, opened 2018-07-30
onion: Make sure we understand the channel_update in the onionreply

### LINKED ISSUE #1730: temporary_channel_failure wrong format together with lnd

cdecker:
As was pointed out by @robtex we have underspecified the format of the nested
`channel_update` in the onionreply: lnd and eclair inserted the raw
channel_update without the type prefix, while we went for the full wire format,
including the type prefix. While we agreed that with the type it is more
flexible, and consistent, we decided to adapt to the majority and at least be
compatibly broken.

Fixes #1730


lightningd/pay.c
```diff
@@ -397,12 +397,27 @@ static void report_routing_failure(struct log *log,
...
+		towire(&channel_update, channel_update+2, tal_bytelen(channel_update) - 2);
+	} else {
+		channel_update = fail->channel_update;
+	}
+
 	gossip_msg = towire_gossip_routing_failure(tmpctx,
 						   &fail->erring_node,
 						   &fail->erring_channel,
 						   (u16) fail->failcode,
 						   fail->channel_update);
```
rustyrussell: You still actually send 'fail->channel_update' though!

You probably want to add a status_debug() on the receiving end and enhance the test to look for it?

cdecker: Ehm, true :disappointed:

PR #1777, opened 2018-07-30
Consider channel parameters before selecting for a route

cdecker:
So far we were ignoring the channel capacity and the self-declared `htlc_minimum_msat`. This PR changes that, returning fewer results when looking for a route, and reducing the number of failed attempts.

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -389,15 +389,21 @@ static void bfg_one_edge(struct node *node,
 		/* FIXME: Bias against smaller channels. */
 		u64 fee;
 		u64 risk;
+		u64 requiredcap;
 
 		if (node->bfg[h].total == INFINITE)
 			continue;
 
 		fee = connection_fee(c, node->bfg[h].total) * fee_scale;
-		risk = node->bfg[h].risk + risk_fee(node->bfg[h].total + fee,
-						    c->delay, riskfactor);
+		requiredcap = node->bfg[h].total + fee;
+		risk = node->bfg[h].risk +
+		       risk_fee(requiredcap, c->delay, riskfactor);
 
-		if (node->bfg[h].total + fee + risk >= MAX_MSATOSHI) {
+		if (requiredcap > chan->satoshis * 1000) {
+			/* Skip this edge if the channel has insufficient
+			 * capacity to route the required amount */
+			continue;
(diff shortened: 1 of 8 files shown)


cdecker:
Ok, finally got the tests to pass. It turns out we didn't set the capacity on locally added channels that haven't received a `channel_announcement` yet.

cdecker:
Ready to review @rustyrussell :wink:

gossipd/routing.c
```diff
@@ -389,15 +389,25 @@ static void bfg_one_edge(struct node *node,
...
-		if (node->bfg[h].total + fee + risk >= MAX_MSATOSHI) {
+		if (requiredcap > chan->satoshis * 1000) {
+			/* Skip this edge if the channel has insufficient
+			 * capacity to route the required amount */
+			continue;
+		} else if (requiredcap < c->htlc_minimum_msat) {
+			/* Skip a channels if it indicated that it won't route
+			 * the requeuested amount. */
+			continue;
+		} else if (requiredcap + risk >= MAX_MSATOSHI) {
```
jb55: can you help me understand why risk is added to requiredcap here and not the previous ones?

gossipd/routing.c
```diff
@@ -1685,5 +1689,6 @@ void handle_local_add_channel(struct routing_state *rstate, const u8 *msg)
 		     type_to_string(tmpctx, struct short_channel_id, &scid));
 
 	/* Create new (unannounced) channel */
-	new_chan(rstate, &scid, &rstate->local_id, &remote_node_id);
+	chan = new_chan(rstate, &scid, &rstate->local_id, &remote_node_id);
+	chan->satoshis = satoshis;
```
rustyrussell: We should move the satoshis setting into new_chan, since it's always set immediately now...

gossipd/routing.c
rustyrussell: `risk` is our internal "added cost to take into account riskiness of using this channel".  It's not seen externally, so it doesn't effect whether they'll reject the payment as too small/big.

jb55: it's all good rusty explained this to me

jb55: `man lightning-getroute` got me sorted. thanks !

cdecker:
Recycling @rustyrussell's ack after addressing the two points he raised.

ACK 5d99214

PR #1783, opened 2018-07-31
Make sure we're not using uninitialized values

practicalswift:
Make sure we're not using uninitialized values.

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -431,12 +431,17 @@ bool parse_wireaddr_internal(const char *arg, struct wireaddr_internal *addr,
 		return true;
 	}
 
+	splitport = port;
+	if (!separate_address_and_port(tmpctx, arg, &ip, &splitport)) {
+		if (err_msg) {
+			*err_msg = "Error parsing hostname";
+		}
+		return false;
+	}
+
 	/* An empty string means IPv4 and IPv6 (which under Linux by default
 	 * means just IPv6, and IPv4 gets autobound). */
-	splitport = port;
-	if (wildcard_ok
-	    && separate_address_and_port(tmpctx, arg, &ip, &splitport)
-	    && streq(ip, "")) {
+	if (wildcard_ok && streq(ip, "")) {
 		addr->itype = ADDR_INTERNAL_ALLPROTO;
 		addr->u.port = splitport;
 		return true;


common/wireaddr.c
```diff
@@ -461,7 +461,7 @@ bool parse_wireaddr_internal(const char *arg, struct wireaddr_internal *addr,
 		return false;
 
 	/* We can't do DNS, so keep unresolved. */
-	if (!wireaddr_from_unresolved(addr, ip, splitport)) {
+	if (!ip || !wireaddr_from_unresolved(addr, ip, splitport)) {
```
rustyrussell: Good catch!  This is a bug,  but this is the wrong fix AFAICT.

We should unconditionally call `separate_address_and_port` (and fail if that returns false, meaning the address is completely malformed).  Putting it under the wildcard_ok condition is a bug.

practicalswift:
@rustyrussell Thanks for reviewing! Updated. Please re-review :-)

channeld/channel.c
```diff
@@ -169,7 +169,7 @@ static void start_commit_timer(struct peer *peer);
 
 static void billboard_update(const struct peer *peer)
 {
-	const char *funding_status, *announce_status, *shutdown_status;
+	const char *funding_status = NULL, *announce_status = NULL, *shutdown_status = NULL;
 
 	if (peer->funding_locked[LOCAL] && peer->funding_locked[REMOTE])
 		funding_status = "Funding transaction locked.";
```
rustyrussell: It is this very initialization which suppresses any -O1 compile warning about them being unset (if they were, in some future path).  So there's no great answer here, but either we do the initialization *and* assert, or we do neither.

common/wireaddr.c
```diff
@@ -431,11 +431,17 @@ bool parse_wireaddr_internal(const char *arg, struct wireaddr_internal *addr,
...
+		return false;
+	}
+
 	/* An empty string means IPv4 and IPv6 (which under Linux by default
 	 * means just IPv6, and IPv4 gets autobound). */
-	splitport = port;
 	if (wildcard_ok
-	    && separate_address_and_port(tmpctx, arg, &ip, &splitport)
 	    && streq(ip, "")) {
 		addr->itype = ADDR_INTERNAL_ALLPROTO;
```
rustyrussell: Great! We could now collapse the if () conditional onto a single line if you wanted...

practicalswift:
@rustyrussell Ah, updated! Please re-review :-)

Should I open a PR for building with `-O1`, or how should we tackle `channel.c`? If we should, that is :-)

rustyrussell:
Maybe we just add a -O1 build to travis?  If we built non-dev with -O1, backtraces might get harder to read...

rustyrussell:
Ack 6efe60546f942fbc6509add3439819f472a7e331

PR #1785, opened 2018-07-31
Remove unused function parameters

practicalswift:
Remove unused function parameters.

### DIFF
--- a/gossipd/gossip.c
+++ b/gossipd/gossip.c
@@ -197,7 +197,7 @@ static struct peer *find_peer(struct daemon *daemon, const struct pubkey *id)
 
 static u8 *encode_short_channel_ids_start(const tal_t *ctx)
 {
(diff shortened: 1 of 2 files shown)


gossipd/gossip.c
```diff
@@ -195,7 +195,7 @@ static struct peer *find_peer(struct daemon *daemon, const struct pubkey *id)
 	return NULL;
 }
 
-static u8 *encode_short_channel_ids_start(const tal_t *ctx)
+static u8 *encode_short_channel_ids_start(void)
```
rustyrussell: General preference is not to return things allocated off tmpctx, but pass it in directly.  So I think the tmpctx at line 200 should be `ctx`?

practicalswift:
@rustyrussell Thanks for reviewing! Updated. Please re-review :-)

PR #1788, opened 2018-07-31
Remove redundant code

practicalswift:
Remove redundant code.

### DIFF
--- a/connectd/connect.c
+++ b/connectd/connect.c
@@ -1393,7 +1393,6 @@ static void connect_failed(struct io_conn *conn, struct reaching *reach)
 				       retry_important, imp);
 	}
 	tal_free(reach);
-	return;
 }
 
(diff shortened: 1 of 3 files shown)


lightningd/peer_control.c
```diff
@@ -1016,7 +1016,8 @@ static void json_close(struct command *cmd,
 			return;
 	}
 
-	if (!channel && peer) {
+	if (!channel) {
+		assert(peer);
```
rustyrussell: Yech, this points out how confusing this function is.

I think this whole thing needs to be hoist into the `if (peer)` case above.  perhaps as a separate cleanup patch?

practicalswift:
@rustyrussell Agreed, the logic is hard to follow. Will look into `json_close` and submit a separate cleanup patch if I come up with something better.

Removed the `peer_control.c` change from this PR. Please re-review :-)

PR #1789, opened 2018-07-31
lightningd: add status flags, show when we're syncing

### LINKED ISSUE #1652: Provide more obvious status message when syncing blocks?
I loaded up an old lightning node from ~6 months ago and opened a new channel. I was trying to figure out why my node was not confirming: 

`CHANNELD_AWAITING_LOCKIN:They've confirmed funding, we haven't yet.`

I turned on debug logging and only then I noticed it was still syncing blocks from bitcoin (I assume to catch up with the latest utxo state).

with bitcoin-cli it's very obvious when the blockchain is syncing, as cli commands return an error/warning message. Perhaps we could have some type of general "status" or "state" that lightningd is in, such as "syncing", so that the cli can provide helpful feedback?

jb55:
This patchset adds status flags to lightningd, which is available in `getinfo`.

See #1652 for motivation.

Fixes #1652

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 
 - Documentation: Added CHANGELOG.md
-- JSON API: `getinfo` has new fields `alias` and `color`.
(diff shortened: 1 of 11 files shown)


jb55:
going to switch to an enum instead of flags after a chat with rusty, will rewrite and open anew

PR #1793, opened 2018-08-01
Use snprintf(...) instead of sprintf(...)

practicalswift:
Use `snprintf(...)` instead of `sprintf(...)`.

Better safe than sorry.

Before:

```
$ git grep [^a-z/]sprintf -- "*.h" "*.c" | grep -v ccan/ | wc -l
11
$ git grep [^a-z/]snprintf -- "*.h" "*.c" | grep -v ccan/ | wc -l
5
```

After:

```
$ git grep [^a-z/]sprintf -- "*.h" "*.c" | grep -v ccan/ | wc -l
0
$ git grep [^a-z/]snprintf -- "*.h" "*.c" | grep -v ccan/ | wc -l
16
```

**Note to reviewers**: The `snprintf` call in `common/json_escaped.c` is the only not using `sizeof(…)`.


rustyrussell:
Nice; particularly like the new checks.

PR #1796, opened 2018-08-02
doc: update HACKING to suggest ramdisk pytests

jb55:
gotta go fast

Suggested-by: @cdecker


jb55:
oops I should have kept reading.. apparently it doesn't help much

PR #1799, opened 2018-08-03
Openingd does babysitting, connectd just connects and trades INIT msgs

rustyrussell:
This is a significant simplification; every peer we care about has a separate daemon, so we don't need to ask connectd about peers, and connectd only deals with peers until we've got the INIT msg, which is a  very linear path.  All other communication with peers is isolated and synchronous.

The result is a lot of internal API cleanup, and I'm sure there's more we can now do where APIs only have a single user, for example.


openingd/opening.c
```diff
@@ -73,35 +73,38 @@ struct state {
...
 	va_list ap;
 	const char *errmsg;
 	u8 *msg;
 
 	va_start(ap, fmt);
-	errmsg = tal_vfmt(state, fmt, ap);
+	errmsg = tal_vfmt(tmpctx, fmt, ap);
 	va_end(ap);
 
+	status_debug("%s", errmsg);
```
cdecker: Prefixing this with some context would be nice.

connectd/connect_wire.csv
```diff
@@ -36,21 +36,18 @@ connectctl_activate_reply,2125
...
 connectctl_connect_to_peer,,id,struct pubkey
 connectctl_connect_to_peer,,seconds_waited,u32
+connectctl_connect_to_peer,,addrhint,?struct wireaddr_internal
 
 # Connectd->master: connect failed.
 connectctl_connect_failed,2020
 connectctl_connect_failed,,id,struct pubkey
 connectctl_connect_failed,,failreason,wirestring
 connectctl_connect_failed,,seconds_to_delay,u32
+connectctl_connect_failed,,addrhint,?struct wireaddr_internal
```
cdecker: I think we should actually be asking `gossipd` for its address hint instead of just blindly handing the cached one back in, would involve yet another message to `gossipd` inbetween, but it'd allow us to react to IP changes better.

cdecker:
Rebased to solve conflict.

PR #1800, opened 2018-08-03
lightningd: add topology and lightningd statuses

### LINKED ISSUE #1652: Provide more obvious status message when syncing blocks?

jb55:
This patchset adds a `status` field to getinfo. This field tracks various statuses such as the new `toplogy_status` in `chain_toplogy`:

  - TOPOLOGY_INITIALIZING
  - TOPOLOGY_SYNCING
  - TOPOLOGY_AT_TIP

These states are mapped to the "high-level" status of lightningd. Right now there's a 1-to-1 mapping between chain_topology statuses and lightningd statuses:

```
TOPOLOGY_INITIALIZING -> LIGHTNINGD_STATUS_INITIALIZING
TOPOLOGY_SYNCING      -> LIGHTNINGD_STATUS_SYNCING
TOPOLOGY_AT_TIP       -> LIGHTNINGD_STATUS_READY
```

but in the future we may want to map other subsystem statuses to lightningd's status.

See #1652 for motivation.

Fixes #1652

This is a rework of #1789 to use enums instead of bit flags for the status, and a rework of #1792 to keep track of topology's status separately as recommended by @rustyrussell.


rustyrussell:
OK, I apologize for spinning my wheels on this for so long.  It's not your code (which can be changed anyway), it's that the JSON API is so hard to change, we need to make sure it's (1) correct, and (2) extensible.

There are two things here:
1. Should we fail (some/most) JSON requests until we're synced?  That actually fixes the "I opened a channel but them I'm waiting for sync and I can't see it" issue which motivated this.

2. Imagine if we indicated that our bitcoind is not on the current tip?  How would we add such a status without breaking things which assumed what the status string was?

So, I think 'status: ready' is a bad API.  We almost want a 'not-ready-because' field, which is clearly non-programmatic.  So, maybe 'startup' and a string field?  Which vanishes once we're ready?  It would be a more global version of the peer billboard; later topology would post to it saying how many blocks to go, for example?  And if we did the 'is bitcoind synced' test we could similarly guesstimate how we're going there.  We could similarly have a gossip startup with messages, or whatever; it would all fit in the same API...

jb55:
sounds good 👍 I'll take a new stab at this the next chance I get

PR #1804, opened 2018-08-06
Remove unnecessary 'find_peer_by_dbid'

goudbor:
find_peer_by_dbid loops through ld->peers but up until
this point ld->peers is never touched, so this will never
return a peer

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -555,7 +555,6 @@ static struct channel *wallet_stmt2channel(const tal_t *ctx, struct wallet *w, s
 	struct channel_info channel_info;
 	struct short_channel_id *scid;
 	struct channel *chan;
-	u64 peer_dbid;
 	struct peer *peer;
 	struct wallet_shachain wshachain;
 	struct channel_config our_config;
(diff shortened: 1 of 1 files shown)


rustyrussell:
Except `wallet_peer_load` calls `new_peer` which places it in the list?

polydin:
`wallet_peer_load` does load a peer from the db into the wallet, 
but prior to calling `wallet_peer_load`, `find_peer_by_dbid` is called 
to check `ld->peers` and load the local variable `peers` using
`ld`.  

So the logic seems to be, get channels from the db, get
the `peer_dbid` from each channel, check `ld->peers` to find
a peer for the corresponding `peer_dbid`, if no such peer found
in `ld->peers` loop through the db and load the peer from there.

But up until this point in the code (before `wallet_peer_load`), 
`ld->peers` has no chance of being populated, so the first if clause 
will never be satisfied (again before `wallet_peer_load` is called).

polydin:
In any event I need to recheck the tests because it looks like some wallet tests aren't passing.

PR #1805, opened 2018-08-06
Fix address issues when seeking both IPv4 and IPv6 on same port.

rustyrussell:


rustyrussell:
Good point.  I've rebased, and added the comments as a fixup! commit for second review.

cdecker:
Squashed, and ready to be merged

ACK 2f86ead

PR #1807, opened 2018-08-06
lightning-cli: consider empty inputs to be strings

bitonic-cjp:
I got a user report with a strange error like this:

    lightning-cli invoice msatoshi="any" label="" description=""
    { "code" : -32602, "message" : "Missing 'description' parameter" }

It turns out that

* The shell translates foo="" into an argv element "foo="
* lightning-cli correctly sees a parameter with name "foo" and value ""
* lightning-cli then incorrectly considers "" to be a literal value, because it consists entirely of decimal characters
* lightning-cli then does *not* put quotes around the empty string, which messes up the JSON-formatted command

This patch fixes this behavior by always considering empty values to be strings (which are then quoted in the JSON command).

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -145,6 +145,9 @@ static char *opt_set_ordered(enum input *input)
 static bool is_literal(const char *arg)
 {
 	size_t arglen = strlen(arg);
+	if (arglen == 0) {
+		return false;
+	}
 	return strspn(arg, "0123456789") == arglen
 		|| streq(arg, "true")
 		|| streq(arg, "false")


cli/lightning-cli.c
```diff
@@ -145,6 +145,7 @@ static char *opt_set_ordered(enum input *input)
 static bool is_literal(const char *arg)
 {
 	size_t arglen = strlen(arg);
+	if(!arglen) return false;
```
rustyrussell: Trivial style things:
1. space after if.
2. it's genuinely a number, so I prefer (arglen == 0).
3. I dislike single-line ifs, especially with control flow.

Thanks!

bitonic-cjp:
Done.

Side note: I used this pull request to experiment with rebase + force-push, to see how Github responds to that. It doesn't look as bad as I feared, though the disadvantage that earlier versions are lost persists. Maybe I should also do that in my bigger pull request.

cdecker:
I quite like the workflow of opening a branch, creating a PR on top of it, then addressing any requests as `fixup!`s, that then get squashed after all the discussion is done (`git rebase -i --autosquash master`). This is quite usable for smallish PRs, that only get rebased at the end and which have an easily squashable fixup series, but can lead to some conflicts during rebase, and gets painful for long-lived PRs.

cdecker:
ACK 812605d

No idea why travis is ignoring your PRs @bitonic-cjp, I'll try to see why.

cdecker:
Looks like your PRs are flagged as "Abuse detected" by Travis. I contacted support to see what's the reason.

PR #1809, opened 2018-08-06
Enhanced param parsing

wythe:
This adds and tests two new macros; `p_req_tal()` and `p_opt_tal()`. They support
callbacks that take a ~~`tal_t *`~~ `struct command *`  context.  Example:

	static char *json_tok_label_x(struct command *cmd,
				      const char *buffer,
				      const jsmntok_t *tok,
				      struct json_escaped **label)

The above is taken from the run-param unit test (near the bottom of the diff).
On success, `*label` is allocated and initialized, and NULL is returned. On failure, an error string
is returned.

`param()` will take care of calling `command_fail` with the error message.

We can pretty much remove all remaining usage of `json_tok_tok` in the codebase
with this type of callback.

The question remains: Should we convert all existing `json_tok_X()` functions to
this new form or leave them alone?

### DIFF
--- a/lightningd/param.c
+++ b/lightningd/param.c
@@ -112,13 +112,7 @@ static bool make_callback(struct command *cmd,
 			return false;
 		}
 	} else {
-		char *msg = def->cbx(cmd, def->name, buffer, tok, def->arg);
-		if (msg) {
-			command_fail_detailed(cmd, JSONRPC2_INVALID_PARAMS,
-					      make_result(cmd, def->name, buffer, tok),
-					      "%s: %s", def->name, msg);
-			return false;
-		}
+		return def->cbx(cmd, def->name, buffer, tok, def->arg);
 	}
 	return true;
 }
--- a/lightningd/param.c
+++ b/lightningd/param.c
@@ -12,16 +12,18 @@ struct param {
 	bool is_set;
 	bool required;
 	param_cb cb;
+	param_cbx cbx;
 	void *arg;
 	size_t argsize;
 };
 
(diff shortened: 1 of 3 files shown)


wythe:
Funny thing is I was actually using `cmd` but then converted it at the last minute.  I'll change it back.

So, I can migrate everything toward this new form over several commits, and in the end we will have:
 ```C
p_req()
p_opt()
p_opt_def()
```
And they all expect to have an address of pointer as their arg, and will be responsible for allocating.
No more need for `argsize` in `struct param`. (Currently `p_req` expects a stack allocated variable).

**Comment by wythe** [comment]
Note I added `p_opt_def_tal`.  They sure let you do a lot in a macro these days!

**Comment by wythe** [comment]
If you look at `json_tok_msat` in the `run-param` test, you see that it calls `json_tok_u64`, and it provides a more verbose error message than just `not an unsigned integer`.  This current method lets us chain handlers together. 

I did think about returning a struct such as:

```C
struct param_result {
        int error;
        char *msg;
};
```
And we could do that, but like you said the payoff is marginal.  I think there may be just 1 or 2 places where the error wouldn't be the default.

Also, `param.c` adds the bad parameter to the data section of the result.

rustyrussell:
"Hard cases make bad law."

if chaining is actually pretty rare, so it's easy to create a 'bool json_to_number(const char *buffer, const jsmntok_t *tok, uint64_t *num)` helper and make json_tok_u64 (and json_tok_num) call that.

If it's more common than my rough reading suggests, yeah, we stick with a char * return...

PR #1810, opened 2018-08-06
Keep test directories of failed tests, and migrate all tests to new fixtures

### LINKED ISSUE #1599: Tests which fail delete their directories.
I had a `wait_for_logs` fail in `test_onchain_all_dust`, and the directory was missing in /tmp afterwards.

Seems we do cleanup even if the tests fail; can we determine that somehow?

cdecker:
This was supposed to be a simple fix of #1599, but I didn't want to re-implement that logic for legacy tests, so I bit the bullet and migrated all existing tests to the new framework.

Most of the changes are move-only, with the adaptation from `self.node_factory` -> `node_factory` and similar. In addition I added some smaller fixes, in separate commits to make it easier to spot them.

Fixes #1599


CHANGELOG.md
```diff
@@ -59,6 +59,8 @@ changes.
   to be compatible with eclair and lnd (#1730).
 - JSON API: `listnodes`: now correctly prints `addresses` if more than
   one is advertised.
+- Failing tests no longer delete the test directory, to allow easier debugging
+  (Issue: #1599)
```
rustyrussell: I've not been putting dev changes into Changelog.  Thoughts?

tests/utils.py
```diff
@@ -507,6 +507,28 @@ def subd_pid(self, subd):
...
+        """Return the state of the channel to the other node.
+
+        Returns None if there is no such peer, or a channel hasn't been funded
+        yet.
+
+        """
+        peers = self.rpc.listpeers(other.info['id'])['peers']
+        if not peers or 'channels' not in peers[0]:
+            return None
+        channel = peers[0]['channels'][0]
```
rustyrussell: There can be multiple channels (ie. onchain ones, too).  Use only_one(peers[0]['channels']) here, since we're assuming there's only one?

tests/utils.py
```diff
@@ -507,6 +507,28 @@ def subd_pid(self, subd):
...
+        return channel['state']
+
+    def get_channel_scid(self, other):
+        """Get the short_channel_id for the channel to the other node.
+        """
+        peers = self.rpc.listpeers(other.info['id'])['peers']
+        if not peers or 'channels' not in peers[0]:
+            return None
+        channel = peers[0]['channels'][0]
```
rustyrussell: Here too...

rustyrussell:
Ack c3a1f6bdc896d669a6858b53ff03d5fbb3ed4137

PR #1811, opened 2018-08-08
Networking enhancements.

### LINKED ISSUE #1487: c-lightning does not send ping
c-lightning handles ping message and responds with pong. But it does not send ping itself.

rustyrussell:
A collection of random network things to make us faster and more reliable, basically.  And the compulsory random code driveby cleanup :)

It's on top of #1799 

Fixes #1487

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Documentation: Added CHANGELOG.md
 - JSON API: `getinfo` has new fields `alias` and `color`.
 - JSON API: `listpeers` has new fields `global_features` and `local_features`.
-- JSON API:`listnodes` has new field `global_features`.
+- JSON API: `listnodes` has new field `global_features`.
+- JSON API: `ping` command to send a ping to a connected peer.
 - Protocol: gossipd now deliberately delays spamming with `channel_update`.
 - Config: `--conf` option to set config file.
 - JSON API: Added description to invoices and payments (#1740).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - JSON API: `listnodes` has new field `global_features`.
 - JSON API: `ping` command to send a ping to a connected peer.
 - Protocol: gossipd now deliberately delays spamming with `channel_update`.
+- Protocol: liveness ping when we commit changes but peer is idle: speeds up
+  failures and reduces forced closures.
 - Config: `--conf` option to set config file.
 - JSON API: Added description to invoices and payments (#1740).
 - pylightning: RpcError now has `method` and `payload` fields.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Logging: JSON connections no longer spam debug logs.
 - Routing: We no longer consider channels that are not usable either because of
   their capacity or their `htlc_minimum_msat` parameter (#1777)
+- We now try to connect to all known addresses for a peer, not just
+  the one given or the first one announced.
 
 ### Deprecated
 
--- a/channeld/channel.c
+++ b/channeld/channel.c
@@ -155,6 +155,9 @@ struct peer {
 
 	/* Make sure timestamps move forward. */
 	u32 last_update_timestamp;
+
+	/* Make sure peer is live. */
+	struct timeabs last_recv;
 };
 
 static u8 *create_channel_announcement(const tal_t *ctx, struct peer *peer);
@@ -928,12 +931,23 @@ static struct commit_sigs *calc_commitsigs(const tal_t *ctx,
 	return commit_sigs;
 }
 
+/* Have we received something from peer recently? */
+static bool peer_recently_active(struct peer *peer)
+{
+	return time_less(time_between(time_now(), peer->last_recv),
+			 time_from_sec(30));
+}
+
 static void maybe_send_ping(struct peer *peer)
 {
 	/* Already have a ping in flight? */
 	if (peer->expecting_pong)
 		return;
 
+	if (peer_recently_active(peer))
+		return;
+
+	/* Send a ping to try to elicit a receive. */
 	sync_crypto_write_no_delay(&peer->cs, PEER_FD,
 				   take(make_ping(NULL, 1, 0)));
 	peer->expecting_pong = true;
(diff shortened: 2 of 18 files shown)


cdecker:
Rebased on top of `master` after merging #1799

connectd/connect.c
```diff
@@ -882,7 +889,7 @@ struct io_plan *connection_out(struct io_conn *conn, struct reaching *reach)
 
 	reach->connstate = "Cryptographic handshake";
 	return initiator_handshake(conn, &reach->daemon->id, &reach->id,
-				   &reach->addr,
+				   &reach->addrs[reach->addrnum],
```
cdecker: Are we sure this cannot overrun? `% tal_count(reach->addrs)` seems to be a good idea here.

connectd/connect.c
```diff
@@ -961,18 +972,18 @@ static struct io_plan *conn_init(struct io_conn *conn, struct reaching *reach)
 static struct io_plan *conn_proxy_init(struct io_conn *conn,
 				       struct reaching *reach)
 {
-	char *host = NULL;
+	const char *host = NULL;
 	u16 port;
+	const struct wireaddr_internal *addr = &reach->addrs[reach->addrnum];
```
cdecker: Same `% tal_count` concern as above.

connectd/connect.c
```diff
@@ -1038,68 +1055,36 @@ gossip_resolve_addr(const tal_t *ctx, const struct pubkey *id)
...
+ 	int fd, af;
+	bool use_proxy = reach->daemon->use_proxy_always;
 
-	/* Already done?  May happen with timer. */
-	if (pubkey_set_get(&daemon->peers, id))
-		return;
-
-	/* If we're trying to reach it right now, that's OK. */
-	if (find_reaching(daemon, id))
+	if (reach->addrnum == tal_count(reach->addrs)) {
```
cdecker: Ok, I see how this clears up my previous concerns, and since there is no way to forget about an `addr` it's safe :+1:

connectd/connect.c
```diff
@@ -1038,68 +1055,36 @@ gossip_resolve_addr(const tal_t *ctx, const struct pubkey *id)
...
-		return;
-	}
+ 	/* Might not even be able to create eg. IPv6 sockets */
+ 	af = -1;
 
-	/* Might not even be able to create eg. IPv6 sockets */
-	af = -1;
-
-	switch (a->itype) {
+	switch (reach->addrs[reach->addrnum].itype) {
```
cdecker: Nit: Pull out into an `addr` pointer instead of indexing everywhere.

CHANGELOG.md
```diff
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Documentation: Added CHANGELOG.md
 - JSON API: `getinfo` has new fields `alias` and `color`.
 - JSON API: `listpeers` has new fields `global_features` and `local_features`.
-- JSON API:`listnodes` has new field `global_features`.
+- JSON API: `listnodes` has new field `global_features`.
+- JSON API: `ping` command to send a ping to a connected peer.
```
cdecker: Pretty sure that was supposed to reference #1487 instead :wink: 

Fixes #1487

tests/test_misc.py
```diff
@@ -778,6 +778,59 @@ def test_reserve_enforcement(node_factory, executor):
...
+    # Make sure channels get announced.
+    bitcoind.generate_block(5)
+
+    # Last thing each node receives is the three NODE_ANNOUNCEMENTS.
+    l1.daemon.wait_for_logs(['\[IN\] 0101'] * 3)
+    l2.daemon.wait_for_logs(['\[IN\] 0101'] * 3)
+
+    # Now need to wait have 30 seconds with no traffic to l1, so it
+    # tries to ping before sending WIRE_COMMITMENT_SIGNED.
+    time.sleep(30)
```
cdecker: Might be worth having a test mode that just scales all timers down by a factor of 10?

tests/test_misc.py
```diff
@@ -778,6 +778,59 @@ def test_reserve_enforcement(node_factory, executor):
...
+    inv = l3.rpc.invoice(123000, 'test_htlc_send_timeout', 'description')
+    try:
```
cdecker: The pytest way of doing this is to use the `pytest.raises` context and stash the exception info (https://docs.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions). That'd also make sure that the `RpcError` gets raised in the first place. This currently fails to check that we fail at all.

lightningd/peer_htlcs.c
```diff
@@ -431,6 +445,12 @@ enum onion_type send_htlc_out(struct channel *out, u64 amount, u32 cltv,
 			    payment_hash, onion_routing_packet, in);
 	tal_add_destructor(hout, destroy_hout_subd_died);
 
+	/* We give it 30 seconds to commit htlc. */
+	if (!out->htlc_timeout)
```
cdecker: How can it happen that we ever attempt to add it twice?

rustyrussell: I'm pretty sure you're right: I will switch this to an assert().

connectd/connect.c
rustyrussell: Yeah, agreed.

lightningd/peer_htlcs.c
rustyrussell: (I started fixing this then realized that the timer is on the channel, not the HTLC).

I replaced comment with:

```
	/* Give channel 30 seconds to commit (first) htlc. */
```

cdecker: Thanks for clarifying

connectd/connect.c
```diff
@@ -1073,6 +1073,7 @@ static void try_reach_one_addr(struct reaching *reach)
 {
  	int fd, af;
 	bool use_proxy = reach->daemon->use_proxy_always;
+	const struct wireaddr_internal *addr = &reach->addrs[reach->addrnum];
```
cdecker: This assignment needs to move down a bit, otherwise the `reach->addrnum == tal_count(reach->addrs)` blows up in our face.

channeld/channel.c
```diff
@@ -971,6 +996,14 @@ static void send_commit(struct peer *peer)
 		return;
 	}
 
+	/* If we haven't received a packet for > 30 seconds, delay. */
+	if (!peer_recently_active(peer)) {
+		/* Mark this as done and try again. */
+		peer->commit_timer = NULL;
+		start_commit_timer(peer);
```
conanoc: This will send PING every 10ms until we receive PONG. It violates BOLT#1.

>A node sending a ping message:
>- SHOULD NOT send ping messages more often than once every 30 seconds.

conanoc: Ah... I didn't notice that ``peer->expecting_pong`` is checked before sending PING.

PR #1812, opened 2018-08-08
pylightning: add method and payload as proper fields.

rustyrussell:
They're much more useful being programatically-accessible, AFAICT.

The string stays the same so they're backwards compatible.


cdecker:
ACK bcdbbbe

PR #1817, opened 2018-08-09
travis: Actually build with clang if we want to

cdecker:
Somewhere along the way we renamed `CC` to `COMPILER`, which effectively disabled the `clang` config on travis when we switched to `./configure`. This just re-enables `clang`


cdecker:
Noticed this while manually checking #1814 :wink:

Issue #1606, opened 2018-06-21
Packaging support for RPM-based systems


rustyrussell:

cdecker:
Is the openSuse build server a common way to distribute RPM packages? I still have an account there, and I could probably try to set up an automated build for various distros, but I'm not sure how commonly they are used.

whitslack:
For whatever it's worth, I have been maintaining a Gentoo ebuild for C-Lightning [here](https://gitlab.com/bitcoin/gentoo/tree/master/net-p2p/c-lightning). It's pretty straightforward, so I don't know if anything in there would be helpful to packagers on other distros.

cdecker:
It's mainly that we need a scalable solution for building the packages for a number of distros x a number of architectures. My current goal is to look at the OpenSuse Build Service to do that for us.

PR #1824, opened 2018-08-11
contrib: add init file for systemd

dougsland:
Similar to init/bitcoind.service, this patch
includes an initial lightningd.service.

### DIFF
--- a/contrib/init/lightningd.service
+++ b/contrib/init/lightningd.service
@@ -1,0 +1,48 @@
+# It is not recommended to modify this file in-place, because it will
+# be overwritten during package upgrades. If you want to add further
+# options or overwrite existing ones then use
+# $ systemctl edit lightningd.service
+# See "man systemd.service" for details.
+
+# Note that almost all daemon options could be specified in
+# /etc/lightningd/bitcoin.conf
+
+[Unit]
+Description=C-Lightning daemon
+Requires=bitcoind.service
+After=bitcoind.service
+
+[Service]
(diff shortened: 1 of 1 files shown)


dougsland:
This init will be useful for packaging schema. Tested against Fedora 28.

Reference: https://github.com/bitcoin/bitcoin/blob/master/contrib/init/bitcoind.service
Related: https://github.com/ElementsProject/lightning/issues/1606#issuecomment-412167373
              https://github.com/ElementsProject/lightning/blob/master/doc/INSTALL.md#to-build-on-fedora

Additional information for testing (On Fedora 28):

0. Create user bitcoin
1. Make sure bitcoind and lightningd are available in /usr/bin
2. Copy https://github.com/bitcoin/bitcoin/blob/master/contrib/init/bitcoind.service to /usr/lib/systemd/system
3. Set bitcoin.conf
     cat /etc/bitcoin/bitcoin.conf 
     testnet=1
4.Copy the lightningd.service from this PR to /usr/lib/systemd/system
5. systemctl daemon-reload
6. systemctl start bitcoind
7. systemctl start lightningd
8 . systemctl status lightningd (should be all green)
9. Reboot the system and see if the service is started automatically.

dougsland:
Before the reboot: 
6.1
systemctl enable bitcoind.service
systemctl enable lightningd.service

Otherwise, won't start the services automatically.

dougsland:
Additionally:
# cat /etc/lightningd/lightningd.conf 
network=testnet

rustyrussell:
Thanks!  Since it works for one user, I'll apply it and if others improve it, that's all good.

Great work!

PR #1825, opened 2018-08-11
do not start fee estimation with option: --dev-override-fee-rates

### LINKED ISSUE #1806: Request: don't log fee-related messages in regtest mode
They fill up the logs and make it hard to find anything else, and regtest doesn't have fee estimation anyway.

SimonVrouwe:
To remove  _fee estimation_ log messages in certain cases:
- Disable fee estimation (loop) completely when `--dev-override-fee-rates` is used
- Because fee estimates is not needed when overridden
- Fixes #1806

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -727,8 +727,17 @@ void setup_topology(struct chain_topology *topo,
 
 void begin_topology(struct chain_topology *topo)
 {
-	/* Begin fee estimation. */
+#if DEVELOPER
+    if (topo->dev_override_fee_rate) {
+        log_info(topo->log, "Fee estimation disabled because: "
+                 "--dev-override-fee-rates");
+    } else {
+        /* Begin fee estimation. */
+        start_fee_estimate(topo);
+    }
+#else
 	start_fee_estimate(topo);
+#endif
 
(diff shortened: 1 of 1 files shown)


SimonVrouwe:
squashed.
To keep setup as general as possible, we could further restrict to `regtest` also, but for now it seems to work.

SimonVrouwe:
@rustyrussell Don't understand why travis fails, does it need a kick or something? On my system all tests pass.

lightningd/chaintopology.c
```diff
@@ -727,8 +727,17 @@ void setup_topology(struct chain_topology *topo,
...
+    if (topo->dev_override_fee_rate) {
+        log_info(topo->log, "Fee estimation disabled because: "
+                 "--dev-override-fee-rates");
+    } else {
+        /* Begin fee estimation. */
+        start_fee_estimate(topo);
+    }
+#else
 	start_fee_estimate(topo);
+#endif
```
rustyrussell: I generally prefer to make the `#if DEVELOPER` clause as small as possible, but that's awkward here, so this is as good a solution as any.

rustyrussell:
Travis fails are unrelated test flakiness, kicking now...

PR #1830, opened 2018-08-12
Export hsm secret to mnemonic seed

### LINKED ISSUE #1762: Feature to export bip32 root key as mnemonic or xprv?
I'm trying to figure out fund backup/recovery strategies, in particular offline.   Typically in HD wallets one receives a bip39 mnemonic and also has access to the root xprv key.  One can write these down, store in safe location.  Also, one can watch or restore funds by importing into 3rd party wallets/tools if need be.

In c-lightning, I haven't yet found a way to do that.   I do see that:

* there is the hsm_secret file, which is a binary format.
* one can list addresses via lightning-cli dev-liastaddrs. Help text mentions bip32 derivation.

So I'm hoping for something that enables me to turn hsm_secret into an xprv.  It could take the form of:

* written instructions.
* a contrib tool/utility
* an rpc call such as export-priv-key

renepickhardt:
This is a suggestion to fix #1762
I am not sure if that patch is desirable since it basically writes the private key to the logfile. On the other hand the logfile will be stored in `./lightning` next to `hsm_secret` so from a security perspective it does not change anything. 
Also I am not to happy with the message in status. knowing `hsm_secret` does not mean that one can recover channel states...

edit: of course the function `display_mnemonic_word_list` does not have to be called in `maybe_create_new_hsm `. It could rather be wrapped to some API call for `lightning-cli`

### DIFF
--- a/hsmd/hsm.c
+++ b/hsmd/hsm.c
@@ -1022,7 +1022,7 @@ static void bitcoin_keypair(struct privkey *privkey,
 }
 
 static void display_mnemonic_word_list(void){
-	char **mnemonic_secret=NULL;
+	char *mnemonic_secret[24];
 	int ret = bip39_mnemonic_from_bytes(NULL,secretstuff.hsm_secret.data,32,mnemonic_secret);
     if (ret!=WALLY_OK){
     	status_info("HSM: convert hsm_secret to mnemonic seed failed. error code: %d",ret);
--- a/hsmd/hsm.c
+++ b/hsmd/hsm.c
@@ -39,6 +39,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 #include <wally_bip32.h>
+#include <wally_bip39.h>
 #include <wire/gen_peer_wire.h>
 #include <wire/wire_io.h>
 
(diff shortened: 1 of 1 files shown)


jb55:
I don't think this would work because we use a different key derivation function than the one in bip39?

Basically: we take the `hsm_secret` and pass it through `hkdf_sha256` to generate the bip39 seed, whereas bip39 specifies PBKDF2+sha512+2048 iterations. So if you were to generate the mnemonic directly from `hsm_secret`, wallets would generate a different seed, unless I'm missing something...

renepickhardt:
I guess you are right. @gabridome said the same in #1762 were I still thought it would be two different problems

rustyrussell:
Hmm, I'm sure I commented on this before, but I can't see it now.  Sorry.

I think this is the place for a standalone tool.  Mainly because logs are a bad place for secret information, because the HSM is deliberately isolated from the rest, and because restoring will require much more than just restoring the HSM seed, for example.

cdecker:
I totally agree that this is a bit backwards, we should give users the possibility to initialize the `hsm_secret` file with a separate tool if they want to, and have a flag in `lightningd` that exits if the `hsm_secret` file was not found and we didn't allow it to be autogenerated. This way we can tell people to initialize, save the mnemonic and then run as normal, or they can tell us they don't care about a mnemonic and we just generate it. It's way more in line with the one-way nature of the mnemonic-to-seed generation process.

I think the final verdict on this is to close this PR and try again?

PR #1833, opened 2018-08-12
replace use of non standard mv -n flag

Mischi:
Hi,

i built c-lighning on OpenBSD-current today and run into a problem with the use of the non-standard `mv -n` flag in `refresh-submodules.sh`.

The FreeBSD man page says that this flag should be avoided in scripts if possible.
See https://man.openbsd.org/FreeBSD-11.1/mv#COMPATIBILITY.

This PR replaces the use of `mv -n`. Now we first test if the file is there. If it is, we exit. If it is not we write the current PID to it.

Cheers,
Fabian

### DIFF
--- a/tools/refresh-submodules.sh
+++ b/tools/refresh-submodules.sh
@@ -6,12 +6,11 @@ if [ $# = 0 ]; then
 fi
 
 # git submodule can't run in parallel.  Really.
-echo $$ > .refresh-submodules.$$
-if ! mv -n .refresh-submodules.$$ .refresh-submodules; then
-    rm -f .refresh-submodules.$$
+if ! mkdir .refresh-submodules 2>/dev/null ; then
     exit 0
 fi
-trap "rm -f .refresh-submodules" EXIT
+
+trap "rmdir .refresh-submodules" EXIT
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
I was hoping that mv -n would be atomic here, but that may have been naive anyway!

I may have to come up with something cleverer (mv of a directory is atomic, for example, but if the second arg is a directory it tries to move into it, instead of over it, so you need to handle that case too).

Mischi:
Good point! Did not thought of this. After a little bit of googling it seems to me that there is no trivial solution which handles all corner cases. My current favorit ist the following snippet I found [here](https://stackoverflow.com/questions/185451/quick-and-dirty-way-to-ensure-only-one-instance-of-a-shell-script-is-running-at/327991#answer-731634):
```sh
if ! mkdir /tmp/myscript.lock 2>/dev/null; then
    echo "Myscript is already running." >&2
    exit 1
fi
```
The comments states that this will not work on every filesystem (e.g. NFS) though ....

rustyrussell:
yes, let's 'mkdir .refresh-submodules' and change the 'rm -rf' to 'rmdir'?

Mischi:
Thanks!

PR #1834, opened 2018-08-13
Fix confirmation of to-us outputs on reorganization or restart.

rustyrussell:
Came across this bug while completing the implementation of option_dataloss_protect.

I think this shows fairly nicely that we should be using BIP32 for the to-remote outputs in a future version of the spec.  We'd still need a different base for the output, but it'd be far simpler.


SimonVrouwe:
This is nice, was just about to submit an issue for an output from `unilateral close` that showed up `unconfirmed` _forever_ when calling `listfunds`. :+1:

PR #1838, opened 2018-08-14
pytest: fix flakiness in test_channel_reenable.

rustyrussell:
In one case, the channel_update which we expected to activate the channel
from l2 was suppressed as redundant.  This is certainly valid, so just
check the results.


rustyrussell:
I'm applying this myself because it's now failing most (unrelated) PRs.

PR #1839, opened 2018-08-14
Update HACKING.md

conanoc:
Added some info in debugging section.
Fix me if I'm wrong.

### DIFF
--- a/doc/HACKING.md
+++ b/doc/HACKING.md
@@ -96,7 +96,7 @@ Here's a list of parts, with notes:
 Debugging
 ---------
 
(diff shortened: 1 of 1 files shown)


wythe:
Good addition.  It might be better to mention you can run `./configure` with `--enable-developer`.  Then we don't have to always add `DEVELOPER=` to the`make` command line.

doc/HACKING.md
```diff
@@ -96,6 +96,8 @@ Here's a list of parts, with notes:
 Debugging
 ---------
 
+You can build c-lightning with DEVELOPER=1 to use dev commands listed in ``cli/lightning-cli help``. Edit ccan/config.h to change DEVELOPER macro. You can log console messages with log_info() in lightningd and status_trace() in other subdaemons.
+
```
rustyrussell: You should use `./configure --enable-developer` instead...

rustyrussell:
Thanks!

PR #1841, opened 2018-08-14
hsmd: log master extended private key when built in developer mode and logging at debug log level

9735dev:
when built for developers and while logging at debug log level, output the base58 check encoded master private key when it is loaded by the hsm at startup.

this makes tinkering easier because otherwise developers have to use/make another utility to get the master key (the hsm_secret is stretched into the bip32 seed by the hkdf and then the master key is derived from the bip32 seed).

build time:
`./configure --enable-developer`

run time:
`--log-level debug`

here's an example log message (with most of the key redacted):
`2018-08-14T14:06:48.155Z lightning_hsmd(12345): HSM: master key tprv...`

### DIFF
--- a/hsmd/hsm.c
+++ b/hsmd/hsm.c
@@ -976,6 +976,20 @@ static void populate_secretstuff(void)
 				  &secretstuff.bip32) != WALLY_OK)
 		status_failed(STATUS_FAIL_INTERNAL_ERROR,
 			      "Can't derive private bip32 key");
+
+#if DEVELOPER
+	unsigned char master_ext_prv_key_serialized[BIP32_SERIALIZED_LEN];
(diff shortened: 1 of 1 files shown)


SimonVrouwe:
So if I understand correctly, the master private key will be logged in plain text in all log files?
Then all the reckless testers on `mainnet` should be very careful with their logs, for example when posting logs reporting issues etc. So in my opinion, at the very least, this should be restricted to coins that don't have any value.

9735dev:
@SimonVrouwe when they compile with the developer options and log at debug level, yes. Maybe it's preferable to make a utility script and put it under the tools dir. I figured that even if this does not get merged, the PR will provide some documentation for other people that might want to get their key.

jb55:
NACK, I would prefer an explicit export command. I would not want my priv key showing up in a log file during development.

PR #1842, opened 2018-08-14
Move max funding and payment values to chainparams

bisoge:
For Litecoin use 60x BTC values to stay consistent with lnd.
Also enforces chain hash values are known to receiver.


channeld/channel.c
```diff
@@ -2060,7 +2061,8 @@ static void handle_offer_htlc(struct peer *peer, const u8 *inmsg)
 					 onion_routing_packet))
 		master_badmsg(WIRE_CHANNEL_OFFER_HTLC, inmsg);
 
-	e = channel_add_htlc(peer->channel, LOCAL, peer->htlc_id,
+	e = channel_add_htlc(peer->channel, &peer->chain_hash,
```
cdecker: See @rustyrussell's comment above

channeld/channel.c
```diff
@@ -2110,6 +2112,10 @@ static void handle_offer_htlc(struct peer *peer, const u8 *inmsg)
 		failcode = WIRE_TEMPORARY_CHANNEL_FAILURE;
 		failmsg = tal_fmt(inmsg, "Too many HTLCs");
 		goto failed;
+	case CHANNEL_ERR_UNKNOWN_CHAIN_HASH:
+		failcode = WIRE_REQUIRED_CHANNEL_FEATURE_MISSING;
```
cdecker: Pretty sure this is not the correct `failcode` for an unsupported blockchain.

channeld/test/run-full_channel.c
```diff
@@ -235,8 +237,10 @@ static void send_and_fulfill_htlc(struct channel *channel,
 	memset(&r, 0, sizeof(r));
 	sha256(&rhash, &r, sizeof(r));
 
-	assert(channel_add_htlc(channel, sender, 1337, msatoshi, 900, &rhash,
-				dummy_routing, NULL) == CHANNEL_ERR_ADD_OK);
+	assert(channel_add_htlc(channel,
+				&chainparams_by_index(0)->genesis_blockhash,
```
cdecker: This index needs to be pulled out into a `#define` or a named `const`.

channeld/full_channel.c
```diff
@@ -299,14 +299,6 @@ static enum channel_add_err add_htlc(struct channel *channel,
...
 	const struct channel_view *view;
 	size_t i;
-
-	/* BOLT #2:
-	 *
-	 * A receiving node MUST fail the channel if:
-	 *...
-	 * - the chain_hash value is set to a hash of a chain that is unknown
-	 *   to the receiver.
-     */
```
cdecker: Why was this removed? This is perfectly reasonable and blockchain agnostic.

bisoge:
Thanks for the reviews, I believe all the requested changes are addressed.

I ended up adding the chain params to the peer struct, not totally necessary but makes the code simpler. If you would prefer to keep them out of the struct they can always be looked up on demand based on genesis blockhash.

For the test I just switched to a lookup by name instead of using the magic number. If that is not clear enough I can still add a define somewhere.

Finally I was able to restore the BOLT quote and pass the source check. I am not sure why based on the text in the RFC (possibly a bug with check-bolt?), but it wants to match:
"The receiver: - if the `chain_hash` value, within the `open_channel`, message is set to a hash of a chain that is unknown to the receiver: - MUST reject the channel."

channeld/full_channel.c
```diff
@@ -298,6 +300,18 @@ static enum channel_add_err add_htlc(struct channel *channel,
...
 	size_t i;
 
+	/* BOLT #2:
+	 *
+	 * A receiving node MUST fail the channel if:
+	 *...
+	 * - the chain_hash value is set to a hash of a chain that is unknown
+	 *   to the receiver.
+     */
+	const struct chainparams *chain_params = chainparams_by_hash(chain_hash);
```
cdecker: I don't understand why this is being checked here at all. The `chain_hash` is channel-specific, and the chain is implicitly communicated through the channel that is being updated, so we should not be passing this down here at all.

cdecker:
I proposed #1913 as a replacement for this change.

bisoge:
@cdecker , agree that adding chainparams to channel struct as done in #1913 is a simpler approach.
Okay with closing this, thanks.

Issue #1843, opened 2018-08-14
Crash after process_getblockhash "Block height out of range"


jb55:
@rustyrussell 

I noticed in 0204c442432733b546df4c21cd07c3430d2b801b it mentions that this is expected, but in the following callback, `blockid` is null, and then it crashes:

```
#0  bitcoin_blkid_to_hex (blockid=0x0, hexstr=hexstr@entry=0x7fff3f35b1a0 "", hexstr_len=hexstr_len@entry=65) at bitcoin/block.c:57
57              fake_txid.shad = blockid->shad;
```

bt:
```
#0  bitcoin_blkid_to_hex (blockid=0x0, hexstr=hexstr@entry=0x7fff3f35b1a0 "", hexstr_len=hexstr_len@entry=65) at bitcoin/block.c:57
#1  0x0000000000405eb6 in bitcoind_getrawblock_ (bitcoind=0xf87f08, blockid=<optimized out>, cb=0x4068c0 <init_topo>, arg=0xf87d58) at lightningd/bitcoind.c:430
#2  0x00000000004056ae in process_getblockhash (bcli=0x13be928) at lightningd/bitcoind.c:662
#3  0x0000000000405530 in bcli_finished (conn=<optimized out>, bcli=0x13be928) at lightningd/bitcoind.c:189
#4  0x000000000045c2fd in notify (ctx=ctx@entry=0x13cd0d0, type=type@entry=TAL_NOTIFY_FREE, info=info@entry=0x13cd0f8, saved_errno=saved_errno@entry=9)
    at ccan/ccan/tal/tal.c:235
#5  0x000000000045c334 in del_tree (t=0x13cd0d0, orig=0x13cd0f8, saved_errno=9) at ccan/ccan/tal/tal.c:395
#6  0x000000000045ccc4 in tal_free (ctx=ctx@entry=0x13cd0f8) at ccan/ccan/tal/tal.c:479
#7  0x0000000000453629 in io_close (conn=conn@entry=0x13cd0f8) at ccan/ccan/io/io.c:443
#8  0x0000000000454c4b in io_loop (timers=timers@entry=0x0, expired=expired@entry=0x0) at ccan/ccan/io/poll.c:314
#9  0x0000000000407bf1 in setup_topology (topo=<optimized out>, timers=timers@entry=0xf87758, min_blockheight=<optimized out>, max_blockheight=<optimized out>)
    at lightningd/chaintopology.c:725
[...]

PR #1845, opened 2018-08-15
chaintopology: always cap max block to bitcoind's block height.

rustyrussell:
We only did this when we were first creating a wallet, or when we
asked for a relative rescan, not in the normal case!

Fixes: #1843


jb55:
Thanks I'll try to reproduce this again. I got the crash when I was testing raciness: I was starting bitcoind after being offline for awhile, and then immediately starting lightningd as well.

Issue #1836, opened 2018-08-13
Don't enforce fees on the peer if we're using default fees


rustyrussell:
molz on IRC reported:

          "status": [
            "CHANNELD_NORMAL:update_fee 829 outside range 5000-400000",
            "ONCHAIN:Tracking our own unilateral close",
            "ONCHAIN:All outputs resolved: waiting 48 more blocks before forgetting channel"
          ],

          "status": [
            "CHANNELD_NORMAL:update_fee 253 outside range 5000-400000",
            "ONCHAIN:Tracking our own unilateral close",
            "ONCHAIN:3 outputs unresolved: in 90 blocks will spend DELAYED_OUTPUT_TO_US (9fd822a85fbc64d916536c9178b837ef30b9ff9d6ac9b48183eeddc20dfb7e61:0) using IGNORING_TINY_PAYMENT"
          ],

          "status": [
            "CHANNELD_NORMAL:update_fee 501 outside range 5000-400000",
            "ONCHAIN:Tracking our own unilateral close",
            "ONCHAIN:All outputs resolved: waiting 46 more blocks before forgetting channel"
          ],

          "status": [
            "CHANNELD_NORMAL:update_fee 720 outside range 5000-400000",
            "ONCHAIN:Tracking their unilateral close",
            "ONCHAIN:All outputs resolved: waiting 45 more blocks before forgetting channel"
          ],

          "status": [
            "CHANNELD_NORMAL:Received error from peer: channel 993fea0d48375990bad016139ec72fbe3533ae83138b52a83f04d7a18c9f18f2: update_fee 22856 outside range 253-8290",
            "ONCHAIN:Tracking their unilateral close",
            "ONCHAIN:All outputs resolved: waiting 50 more blocks before forgetting channel"
          ],

molxyz:
@rustyrussell 
[crash.log.tail.zip](https://github.com/ElementsProject/lightning/files/2285237/crash.log.tail.zip)

PR #1846, opened 2018-08-15
All the feerate you can eat...

rustyrussell:
Lots of things in here, mainly weaning us off manual fee manipulation where possible.

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -20,6 +20,7 @@
 #include <lightningd/channel_control.h>
 #include <lightningd/gossip_control.h>
 #include <lightningd/json.h>
+#include <lightningd/jsonrpc_errors.h>
 #include <lightningd/param.h>
 
 /* Mutual recursion via timer. */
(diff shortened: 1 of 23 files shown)


lightningd/chaintopology.c
```diff
@@ -444,6 +459,20 @@ static void json_feerates(struct command *cmd,
 	if (missing)
 		json_add_string(response, "warning",
 				"Some fee estimates unavailable: bitcoind startup?");
+	else {
+		json_object_start(response, "onchain_estimates");
```
SimonVrouwe: change `onchain_estimates` to `onchain_fee_estimates` to make it more descriptive

rustyrussell: Agreed!

SimonVrouwe:
Like the idea of `feehistory`, but it should take into account the time constant `poll_second`, which can change using `--dev-bitcoind-poll`.

It looks `feehistory` is currently only used to determine the `min` and `max` feerate estimates of the recent past.
https://github.com/ElementsProject/lightning/blob/f5e023d012b2ffc82fff1b6aee5ec1a0ec8ab974/lightningd/chaintopology.c#L744-L748
https://github.com/ElementsProject/lightning/blob/f5e023d012b2ffc82fff1b6aee5ec1a0ec8ab974/lightningd/chaintopology.c#L785-L788

Just some thoughts: 
- Perhaps `feehistory` should consist of _just_ the `min`/`max` outliers of past (lets say) 90 seconds. 
- Not sure how to code that, maybe by adding timestamp or a (`poll_counter`) to these outliers to replace them when too old.

SimonVrouwe:
Or forget about timestamping, set a floor to `poll_seconds` of (lets say) 5 seconds? Currently it can be set to any `u32` using `--dev-bitcoind-poll`.

Then give `fee_history` array 18 elements so it can store 90 seconds of history. And limit the search for the outliers to the first `floor(90 / poll_seconds)` elements of that array.

rustyrussell:
It's tempting to get more sophisticated, but remember that dev options are only for testing: here having only 3 is a feature.

The long term averaging is The Right Thing, and it's already in place. But older nodes don't do that, plus there may be other reasons why they're using what we consider outliers, eg. just starting up.

For min-max range checks, we really want to know "is this reasonable": I think if it's a value that estimatefee returns, the answer has to be "yes".

rustyrussell:
Rebased to fix conflicts with new param sweep...

rustyrussell:
Travis uncovered a real issue, will address tomorrow, meanwhile this is now a WIP

SimonVrouwe:
Testing this branch and the `feerates` command showed some weird results concerning unit `sipa<->bitcoind` conversions.

I did some digging and found, which looks like an old bug:
https://github.com/ElementsProject/lightning/blob/4e0e1492f56d6febf8d62526a67561b8bba177d4/lightningd/bitcoind.c#L335

should be **multiply** by 4 because `feerate` **[sat/kilobyte]** is `feerate` **[sat /( kw/4)]** is `4*feerate` **[sat/kw]**

same here:
https://github.com/ElementsProject/lightning/blob/4e0e1492f56d6febf8d62526a67561b8bba177d4/lightningd/chaintopology.c#L414-L416

`/ 4` should be `* 4` and `mulfactor` should `1/4` but maybe better change `mulfactor` to `u32 divfactor` ...etc

https://github.com/ElementsProject/lightning/blob/4e0e1492f56d6febf8d62526a67561b8bba177d4/lightningd/chaintopology.c#L458

needs `/ 1000` factor

lightningd/chaintopology.c
```diff
@@ -25,6 +25,17 @@
...
 
+/* get_init_blockhash sets topo->root, start_fee_estimate clears
+ * feerate_uninitialized (even if unsuccessful) */
+static void maybe_completed_init(struct chain_topology *topo)
+{
+	if (topo->feerate_uninitialized)
+		return;
+	if (!topo->root)
+		return;
+	io_break(topo);
```
cdecker: Looks a bit like we're trying to reinvent the countdown latch here :-)

lightningd/chaintopology.c
```diff
@@ -288,9 +299,13 @@ static void update_feerates(struct bitcoind *bitcoind,
 		}
 		topo->feerate[i] = feerate;
 	}
-	/* Moving this forward in time is ok, but feerate smoothing is effectively
-	 * disabled until topo->startup is set to false */
-	topo->startup = false;
+
+	if (topo->feerate_uninitialized) {
+		/* Moving this forward in time is ok, but feerate smoothing is effectively
+		 * disabled until topo->startup is set to false */
```
cdecker: `topo->startup` was removed, comment needs adjusting.

tests/utils.py
```diff
@@ -601,15 +601,23 @@ def wait_pay():
...
         self.rpc.waitsendpay(rhash)
 
-    def fake_bitcoind_fail(self, exitcode):
+    def fake_bitcoind_fail(self, failscript='exit 1', cmd=None):
         # Create and rename, for atomicity.
         f = os.path.join(self.daemon.lightning_dir, "bitcoin-cli-fail.tmp")
         with open(f, "w") as text_file:
-            text_file.write("%d" % exitcode)
-        os.rename(f, os.path.join(self.daemon.lightning_dir, "bitcoin-cli-fail"))
+            text_file.write(failscript)
```
cdecker: We'll definitely need to change this, a python script writing a python script that is then run by a subprocess, that's an antipattern if I ever saw one :smile:

rustyrussell:
This PR is hard to split.  About half the changes are test changes required to test the code changes.

The initial motivation was #1836 but when I started digging I found more cases of problemsl hence the branch name :(  I'll split it on 'channel: update fees after lockin.' which does that then.

rustyrussell:
@SimonVrouwe I don't think that's right (we had it that way in the past though, so you're not alone!)

If the rate is 1 satoshi per byte, and we have a 250 byte tx, it costs 1 * 250 = 250 satoshi.

That tx is 1000 sipa, so the rate per *kw* is 0.25: 0.25 * 1000 = 250 satoshi.  So to convert bitcoind-style feerates to ours, we *divide* by 4.

Your point about the gross feerate estimate is correct however.  I've invited you to be a r/o collaborator, which means I can ping you for reviews in future :)

SimonVrouwe:
@rustyrussell Thanks for clearing that out! I found myself a bit lost in the woods there :cold_sweat:
What looks like a _unit conversion_ of a certain quantity (size in bytes), but it is not. Weight and size are really separate quantities/dimensions which happen to have a (numerical) relation between them.

PR #1850, opened 2018-08-15
Add support for litecoin p2sh-segwit and bech32 address types

bisoge:


bitcoin/base58.c
```diff
@@ -92,9 +92,10 @@ bool p2sh_from_base58(bool *test_net,
 	if (!from_base58(&version, p2sh, base58, len))
 		return false;
 
-	if (version == 196)
+	/* 196 BTC testnet, 58 LTC testnet, 5 BTC mainnet, 50 LTC mainnet */
```
cdecker: These values could also be moved into chainparams, and here we'd just iterate over all known chainparams and set `*test_net = cp->testnet` if `cp->p2sh_version` matches.

bisoge: Agreed, done in next commit.

bisoge:
Rebased and squashed into single commit due to conflicts with #1913. Also added legacy address support and fixed invoice and p2sh-segwit address generation.

PR #1852, opened 2018-08-16
Do periodic ping

conanoc:
Send ping message every 60 sec. If the peer was recently active,
wait until next period. If we are waiting for pong, do not send
ping until pong message is received.
Also handles ping flood. There should be less than 10 pings in 30 sec.

### DIFF
--- a/channeld/channel.c
+++ b/channeld/channel.c
@@ -108,6 +108,7 @@ struct peer {
 	struct msg_queue from_master, from_gossipd;
 
 	struct timers timers;
+	struct oneshot *ping_timer;
 	struct oneshot *commit_timer;
 	u64 commit_timer_attempts;
 	u32 commit_msec;
@@ -162,6 +163,7 @@ struct peer {
 
 static u8 *create_channel_announcement(const tal_t *ctx, struct peer *peer);
 static void start_commit_timer(struct peer *peer);
+static void start_ping_timer(struct peer *peer);
 
 static void billboard_update(const struct peer *peer)
 {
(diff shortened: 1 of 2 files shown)


rustyrussell:
(I thought I commented on this before... weird)

For the first commit; we already ping as required, so why ping every 60 seconds?

I don't think a ping flood is a worry, TBH.  They're not allowed to send more than one at a time, so they're latency limited already.  There are other ways peers can waste our bandwidth, such as appending junk to messages, or doing lots of gossip queries, or sending odd-numbered unknown messages, or just spamming us with repeated gossip.  To deal with this properly we might need to sum all the "useless" bytes sent and received, and if it passes some ratio we either throttle or give up on the peer?  (Note: this is a much bigger job!)

The simple option is to enable TCP keepalives; here are the default settings for Linux systems:

       tcp_keepalive_intvl (integer; default: 75; since Linux 2.4)
              The number of seconds between TCP keep-alive probes.

       tcp_keepalive_probes (integer; default: 9; since Linux 2.2)
              The maximum number of TCP keep-alive probes to send before  giv‐
              ing  up  and  killing  the connection if no response is obtained
              from the other end.

       tcp_keepalive_time (integer; default: 7200; since Linux 2.2)
              The number of seconds a connection needs to be idle  before  TCP
              begins sending out keep-alive probes.  Keep-alives are sent only
              when the SO_KEEPALIVE socket option  is  enabled.   The  default
              value  is  7200 seconds (2 hours).  An idle connection is termi‐
              nated after approximately an additional 11 minutes (9 probes  an
              interval of 75 seconds apart) when keep-alive is enabled.

              Note that underlying connection tracking mechanisms and applica‐
              tion timeouts may be much shorter.

conanoc:
I agree. Anyway, this is what the specification(BOLT#1) is recommending.

rustyrussell:
I'm going to close this, as I think the existing ping-before-use method is sufficient.  Please reopen if you disagree.

PR #1854, opened 2018-08-17
Support option_dataloss_protect

rustyrussell:
Based on #1834 (first six commits).

It warms up with 5 bigfixes/cleanups.

The next four implement *sending* option_dataloss_protect fields.

The next two add a test that we're sending it correctly.

The next three add HSM support for checking a future secret, without adding a side-channel attack.

The next three implement the check and make sure we don't broadcast.

The last two let onchaind use the info to collect the to-us output, even if it's in the future and can't do anything else, and then make sure we save it to do so we're good after restart.


cdecker:
Rebased on top of `master` after merging #1863

channeld/channel.c
```diff
@@ -1091,30 +1091,42 @@ static void start_commit_timer(struct peer *peer)
...
+							 &s))
 		status_failed(STATUS_FAIL_HSM_IO,
 			      "Bad per_commitment_point reply %s",
 			      tal_hex(tmpctx, msg));
 
-	if (!old_commit_secret)
-		status_failed(STATUS_FAIL_HSM_IO,
-			      "No secret in per_commitment_point_reply %"PRIu64,
-			      revoke_index+2);
+	if (old_secret) {
```
cdecker: We might want to move this check up, and avoid the roundtrip to the HSM if we're going to discard its result anyway.

channeld/channel.c
```diff
@@ -1849,10 +1854,19 @@ static void peer_reconnect(struct peer *peer)
...
+	 *       - MUST set `your_last_per_commitment_secret` to all zeroes
+	 *     - otherwise:
+	 *       - MUST set `your_last_per_commitment_secret` to the last
+	 *         `per_commitment_secret` it received
 	 */
-	msg = towire_channel_reestablish(NULL, &peer->channel_id,
-					 peer->next_index[LOCAL],
-					 peer->revocations_received);
+	msg = towire_channel_reestablish_option_data_loss_protect
+		(NULL, &peer->channel_id,
```
cdecker: Nit: code style seems off here, the `(` should be part of the previous line shouldn't it?

lightningd/channel_control.c
```diff
@@ -101,6 +102,32 @@ static void peer_got_shutdown(struct channel *channel, const u8 *msg)
...
+		= tal_dup(channel, struct pubkey, &per_commitment_point);
+
+	/* We don't fail yet, since we want daemon to send them an error
+	 * to trigger rebroadcasting.  But make sure we set error now in
+	 * case something else goes wrong! */
+	derive_channel_id(&cid,
+			  &channel->funding_txid,
+			  channel->funding_outnum);
+	channel->error = towire_errorfmt(channel, &cid,
+					 "Catastrophic failure: please close channel");
```
cdecker: Is it really a good idea to tell the peer that we're basically at their mercy?

lightningd/peer_control.c
```diff
@@ -370,15 +370,28 @@ register_close_command(struct lightningd *ld,
...
-	sign_last_tx(channel);
+	/* BOLT #2:
+	 *
+	 * - if `next_remote_revocation_number` is greater than expected
+	 *   above, AND `your_last_per_commitment_secret` is correct for that
+	 *   `next_remote_revocation_number` minus 1:
+	 *      - MUST NOT broadcast its commitment transaction.
+	 */
+	if (channel->future_per_commitment_point && !cooperative) {
+		log_broken(channel->log,
```
cdecker: Probably should also change the state of the channel, to make this case evident in the `listpeers` output. It's not like we can proceed beyond this without the other side collaborating anyway.

onchaind/onchain.c
```diff
@@ -2117,6 +2117,94 @@ static void handle_their_unilateral(const struct bitcoin_tx *tx,
...
+
+			/* Tell the master that it will want to add
+			 * this UTXO to its outputs */
+			wire_sync_write(REQ_FD, towire_onchain_add_utxo(
+						    tmpctx, txid, i,
+						    possible_remote_per_commitment_point,
+						    tx->output[i].amount,
+						    tx_blockheight));
+			to_us_output = i;
+			continue;
```
cdecker: This is pretty much a no-op, can be removed?

tests/test_connection.py
```diff
@@ -1194,4 +1194,14 @@ def test_dataloss_protection(node_factory):
     l2.daemon.wait_for_log("Cannot broadcast our commitment tx: they have a future one")
     assert not l2.daemon.is_in_log('sendrawtx exit 0')
 
-    # FIXME: l2 should still be able to collect onchain.
+    closetxid = only_one(bitcoind.rpc.getrawmempool(False))
```
cdecker: On slow systems this may be flaky (since Travis is slow), better to wait for `len(bitcoind.rpc.getrawmempool(False)) == 1`.

lightningd/peer_control.c
cdecker: Fixed in the very next commit :+1:

cdecker:
Just some minor comments, and one sanity check that I couldn't clearly find a solution to: since the last commit obliviously stores that last commitment point is it possible that a peer can misuse it to make us forget it? Consider the following:

 - Peer reconnects, but we forgot a commitment, which we discover since they sent us a commitment point we didn't know. We stall and wait for them to drop to chain and send them an error
 - The peer notices that we're behind and they just told us the info to grab our funds in case of a unilateral
 - They reconnect, this time sending an old commitment point, and we overwrite the latest one with an earlier one
 - They drop, and we can't grab our funds anymore.

We should either refuse to communicate with the peer completely once we discover that we had amnesia, or we should remember all of the commitment points.

cdecker:
If the above concerns are addressed I'm happy to

ACK c3b181b

:wink:

channeld/channel.c
rustyrussell: It's hard with names this long; I have no strong opinion either way since it's so ugly already.

But usually I prefer operators on the line below, as it makes it visually obvious that it's a continuation, eg:

```
    x = really_long thing
         + something else;
```

lightningd/channel_control.c
rustyrussell: We *need* to send an error, so they close the channel.  If we pretend it's some other error, we better make sure it's exactly the same, and an error that actually happens.  Suggestions welcome!

onchaind/onchain.c
rustyrussell: Agreed.  Cut & paste from the normal unilateral case, which has other branches....

tests/test_connection.py
rustyrussell: I've yet to see that actually happen; we had some false positives where it seemed bitcoind accepted a tx which wasn't in the mempool, but it turned out to be a bug in our tests.  If it happens, I'll fix...

rustyrussell:
Yes, this is possible because we don't store channel->error in the database; if we restart, we'll happily talk with the peer again (since channel->error is cleared).

I've decided to introduce a new state, AWAITING_UNILATERAL.  We set that both when we drop to chain, and when we can't because of this option, and send the generic "Awaiting unilateral close" in both cases.

cdecker:
ACK f5c0814

ZmnSCPxj:
A quick question to get up to speed: do we *utilize* `option_data_loss_protect` ourselves if the peer has it, or do we only support it for the peer and not our own side?

cdecker:
We will actually drop to chain and make use of the commitment point if we see a discrepancy.

ZmnSCPxj:
Ok, thank you, that is good to hear..

ZmnSCPxj:
Wait, maybe my question, was not clear....

If we asked our peer to save our data with `option_data_loss_protect`, and our data does not match with the peer data, then the peer should drop to chain, since it knows a later state than we do.

And if our peer saved its data with us, and later it comes up and we realize our data does not match, we should drop to chain since we know a later state than the peer does.

I am asking the first scenario above (we save our data with peer, and if the peer finds out we are using obsolete data, the peer drops to chain with its later state).

cdecker:
Yes, we are indeed sending an error, and disconnecting if we discover we're behind, waiting for the other side to drop to chain, and then using the commitment point they just told us to grab our funds. As pointed out in the PR, the last part is really hoping there weren't any intermediate updates and the peer doesn't use one of those gambling on us not knowing the punishment for it.

ZmnSCPxj:
Okay, thanks, that seems quite good.

Issue #1607, opened 2018-06-21
Expand memleak detection to subdaemons


rustyrussell:
We currently only do memory leak detection on the main daemon: but some subdaemons are longlived too, so they should be monitored as well.

cdecker:
Can this be closed following #1856 being merged?

PR #1856, opened 2018-08-20
Leak detection in subdaemons.

rustyrussell:
This is #1607 

I've had some issues with the reliability of leak detection code in subdaemons which needs more work, so this just contains the infrastructure cleanups (first 8 patches) and the actual leak (or false positive) fixes so far for all the daemons, which is the important part for -rc1.


hsmd/hsm.c
```diff
@@ -61,6 +63,12 @@ struct client {
 	u64 capabilities;
 };
 
+/* We keep a map of nonzero dbid -> clients */
+static UINTMAP(struct client *) clients;
+/* We get three zero-dbid clients: master, gossipd and connnectd. */
+static struct client *dbid_zero_clients[3];
```
cdecker: Nit: `global_daemon_clients` sounds a bit more intuitive to me

hsmd/hsm.c
```diff
@@ -61,6 +63,12 @@ struct client {
...
+static size_t num_dbid_zero_clients;
```
cdecker: This seems not to get initialized to 0.

cdecker:
Regarding the reordering inside of structs to have `list_node`s as first element: how stable is that? Will we get some warnings or even memleak errors if we forget that in the future or will it be silent?

ACK 951c854

PR #1857, opened 2018-08-20
docker: Make the bitcoind version configurable at build time

cdecker:
This was suggested via mail: the SHA256 sums should be extracted from the
sha256sums file we are checking against, which also allows us to switch bitcoind
version at build time.

Suggested-by: Giles Hall <@vishnubob>
Signed-off-by: Christian Decker <@cdecker>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,21 +21,22 @@ RUN apk add --no-cache \
 
 WORKDIR /opt
 
-ENV BITCOIN_VERSION 0.16.0
-ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz
-ENV BITCOIN_SHA256 e6322c69bcc974a29e6a715e0ecb8799d2d21691d683eeb8fef65fc5f6a66477
+ARG BITCOIN_VERSION=0.16.0
+ENV BITCOIN_TARBALL bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz
+ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/$BITCOIN_TARBALL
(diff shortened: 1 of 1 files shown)


NicolasDorier:
This should not be done. If somebody change bitcoin-core binaries, he can as well change as well the hash in the file.

cdecker:
> NACK, if somebody change bitcoin-core binaries, he can as well change as well the hash in SHA256SUMS.asc.

I'd contest that, since the SHA256SUMS.asc file is signed by @laanwj's release key, giving us a chain of trust to him, and we actually verify the signature before comparing the checksums.

NicolasDorier:
Yes, but if laanwj goes rogue or loose his key... maybe I am paranoiac but I just like to be sure the binary did not changed, regardless whether I trust him or not.

At least I would make the hash a  `ARG`, which, if empty, do what you are doing, and if not empty, also check against the BITCOIN_SHA passed as build parameter.

laanwj:
> Yes, but if laanwj goes rogue or loose his key... maybe I am paranoiac but I just like to be sure the binary did not changed, regardless whether I trust him or not.

:+1:

not entirely sure what is the context here, but what you ideally would want to verify are gitian signatures of multiple people (from https://github.com/bitcoin-core/gitian.sigs), as that is the way to audit that the source code matches the distribution binary

cdecker:
Ok, I'll open a separate issue for further hardening :+1:

vishnubob:
Prior to this change, the downloading and verification of the ASC file would only break the build if the ASC file was /itself/ tampered with.  This file only lists the hashes of the various archives, and is therefore not a verification of the tarball, but a catalog of SHA values, one for each archive.  In other words, the verification of the ASC file was essentially a NOP.  I guess if the ASC file changed, this could serve as a canary to break the build process, but the combination of the embedded hash combined with the NOP-like verification gave me the creepy-crawlies.  My proposed solution was to align the code with I understand to be its actual intent.  Since Dockerfiles can provide a systematic attack vector at build time, clarity seems like a worthy cause.

PR #1864, opened 2018-08-22
Never guess feerate

rustyrussell:
This is based on #1863 and #1861 (merged)

It's the basically the middle part of #1846 .

Things this fixes:

1.  There was a window on startup where we'd use default feerate. Instead, we wait.
2.  We did not correctly update fees when restarting, or if they changed before funding tx locked in.
3.  If we couldn't get a fee estimate on startup, we put zeroes into the averager, which meant when bitcoind started giving estimates, we'd be low for quite a while.
4.  Since we hid the default feerate in the estimator, callers couldn't know when it was being used.
     In practice, they either have other sources to estimate feerate (existing tx), or should refuse
     (opening channels, withdrawing funds).

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -189,16 +189,24 @@ u8 *p2wpkh_for_keyidx(const tal_t *ctx, struct lightningd *ld, u64 keyidx)
 	return scriptpubkey_p2wpkh(ctx, &shutdownkey);
 }
 
-u32 feerate_min(struct lightningd *ld)
+u32 feerate_min(struct lightningd *ld, bool *unknown)
 {
 	u32 min;
+
+	if (unknown)
+		*unknown = false;
 
 	/* We can't allow less than feerate_floor, since that won't relay */
 	if (ld->config.ignore_fee_limits)
 		min = 1;
-	else
-		/* Set this to half of slow rate.*/
-		min = get_feerate(ld->topology, FEERATE_SLOW) / 2;
+	else {
+		u32 feerate = try_get_feerate(ld->topology, FEERATE_SLOW);
+		if (!feerate && unknown)
+			*unknown = true;
+
+		/* Set this to half of slow rate (if unknown, will be floor) */
+		min = feerate / 2;
+	}
 
 	if (min < feerate_floor())
 		return feerate_floor();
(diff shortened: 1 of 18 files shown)


rustyrussell:
Clearly I broke something trying to rebase this to separate it :(

WIll have to investigate travis failures tomorrow...

cdecker:
Rebased on top of #1863 and #1861

tests/test_connection.py
```diff
@@ -848,62 +848,36 @@ def test_update_fee(node_factory, bitcoind):
...
     # FIXME: Test case where opening denied.
-    l1, l2 = node_factory.line_graph(2, opts={'dev-max-fee-multiplier': 5}, fundchannel=True)
+    l1, l2 = node_factory.line_graph(2, opts={'dev-max-fee-multiplier': 5, 'may_reconnect': True}, fundchannel=True)
 
-    # L1 asks for stupid low fees
-    l1.rpc.dev_setfees(15)
+    # L1 asks for stupid low fee (will actually hit the floor of 253)
+    l1.stop()
+    l1.set_feerates((15, 15, 15), False)
+    l1.start()
```
cdecker: Seems like we'll be doing this triple of operations a few times, we could pull this into its own method to document why we restart here.

lightningd/peer_control.c
```diff
@@ -179,16 +179,25 @@ u8 *p2wpkh_for_keyidx(const tal_t *ctx, struct lightningd *ld, u64 keyidx)
...
 	/* We can't allow less than feerate_floor, since that won't relay */
 	if (ld->config.ignore_fee_limits)
 		min = 1;
-	else
-		/* Set this to half of slow rate.*/
-		min = get_feerate(ld->topology, FEERATE_SLOW) / 2;
+	else {
+		u32 feerate = try_get_feerate(ld->topology, FEERATE_SLOW);
+		if (!feerate) {
+			if (unknown)
```
cdecker: These two `if`s can be joined

SimonVrouwe:
https://github.com/ElementsProject/lightning/blob/72cd0fc83e6ece075d6465702aa0e61199e02da9/lightningd/chaintopology.c#L347-L348
 This comment can be removed. Moving this forward in time will post-pone `setup_topology()`.

rustyrussell:
@SimonVrouwe Thanks, I changed the comment to something sensible...

cdecker:
Waiting for #1854 to be ready to merge before I rebase this on top and then we can merge (if no rebase failures emerge)

cdecker:
The rebase took a bit longer than expected due to some merge conflicts with #1854:

The fee-change to upset the peer needed to be made more acute:

https://github.com/ElementsProject/lightning/blob/80724e4eb4af931e525d92b52f9a71fad20ad23e/tests/test_pay.py#L108-L112

but it should still have the desired semantics.

The new commit after reconnect causing the `test_dataloss_protection` to become really brittle since it may now happen with either commitment 3 or 4. I decided to weaken the check, but left it in a separate `fixup!` so we can decide what to do.

Other than those two things the PR looks good, so feel free to squash and merge if you're ok with my changes.

ACK 3329524

rustyrussell:
Ack f0bef2ea09027ac3f4425b7fd94bd59c0effe05f

(Self-ack since it's just a rebase with --autosquash).

Issue #1759, opened 2018-07-26
node doesn't start


robtex:
### Issue and Steps to Reproduce

updated to latest master `v0.6-155-gd6048de`

node doesn't seem to start. last things in logs is:

```
2018-07-26T18:28:44.439Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_channeld
2018-07-26T18:28:44.449Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_closingd
2018-07-26T18:28:44.454Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_connectd
2018-07-26T18:28:44.460Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_gossipd
2018-07-26T18:28:44.468Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_hsmd
2018-07-26T18:28:44.471Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_onchaind
2018-07-26T18:28:44.474Z lightningd(25339): testing /usr/local/libexec/c-lightning/lightning_openingd
2018-07-26T18:28:44.848Z lightning_hsmd(25347): pid 25347, msgfd 14
2018-07-26T18:28:44.948Z lightning_connectd(25348): pid 25348, msgfd 17
2018-07-26T18:28:44.948Z lightning_hsmd(25347): Client: Received message 11 from client
2018-07-26T18:28:44.949Z lightning_hsmd(25347): Client: Received message 9 from client
2018-07-26T18:28:45.070Z lightning_connectd(25348): Created IPv4 listener on port 9735
2018-07-26T18:28:45.070Z lightning_connectd(25348): REPLY WIRE_CONNECTCTL_INIT_REPLY with 0 fds
2018-07-26T18:28:45.077Z lightning_gossipd(25349): pid 25349, msgfd 16
2018-07-26T18:28:46.646Z lightningd(25339): Loaded 0 HTLC signatures from DB
2018-07-26T18:28:46.657Z lightningd(25339): Loaded 0 HTLC signatures from DB
2018-07-26T18:28:46.668Z lightningd(25339): Loaded 0 HTLC signatures from DB
2018-07-26T18:28:46.673Z lightningd(25339): Loaded 0 HTLC signatures from DB
2018-07-26T18:28:46.688Z lightningd(25339): Loaded 0 HTLC signatures from DB
.
.
.
[...]

PR #1868, opened 2018-08-23
hsmd: use async for status reporting.

rustyrussell:
We can otherwise deadlock against lightningd which also talks to us sync.

Fixes: #1759


cdecker:
ACk 8ced62b

PR #1870, opened 2018-08-24
Feerate enhancements: cleanups, `feerate` command and wider min-max tolerances.

rustyrussell:
This is the final (and less critical) part of the original mega-feerate PR #1846 

It was entitled i-dream-of-feerate after I woke up thinking about feerates and how we should be handling it.  I had no idea how much work there was to do on this neglected corner of the code until the bug report which triggered #1836 ...

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -20,6 +20,7 @@
 #include <lightningd/channel_control.h>
 #include <lightningd/gossip_control.h>
 #include <lightningd/json.h>
+#include <lightningd/jsonrpc_errors.h>
 #include <lightningd/param.h>
 
 /* Mutual recursion via timer. */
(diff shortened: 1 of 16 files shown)


lightningd/chaintopology.c
```diff
@@ -388,6 +389,84 @@ static void start_fee_estimate(struct chain_topology *topo)
...
+				"Some fee estimates unavailable: bitcoind startup?");
+	json_object_end(response);
+
+	command_success(cmd, response);
+}
+
+static const struct json_command feerates_command = {
+	"feerates",
+	json_feerates,
+	"Add/query feerate estimates, either satoshi-per-kw ({style} sipa) or satoshi-per-kb ({style} bitcoind) for {urgent}, {normal} and {slow}."
```
cdecker: `sipa` is bound to be confusing for users, I'd suggest `spkw` and `spkb` as units instead, defaulting to `spkb`

cdecker:
Rebased on top of `master` btw :-)

lightningd/chaintopology.c
rustyrussell: Changed to perkw / perkb.  I think I'll remember that better...

SimonVrouwe:
Shouldn't the `feerate` _inject_ command be a developer option? 

As of now, after injecting fee estimates using the `feerate` command, within 30s it is overridden by the `estimatesmartfee` poll, so currently not really much use of the _inject_ part unless you also use a `fake-bitcoin-cli` that spews feerates 0.

SimonVrouwe:
Thinking about it bit longer, the `feerate` command calls `update_feerate` which calls `next_updatefee_timer` that adds new poll (timer) loop in _addition_ to loop(s) already running. This is not what you want I think. If you now call `feerates` command a few times you get faster and irregular fee updates.

Perhaps a solution is when a `feerates` _inject_ command is called, set `topo->poll_seconds = 0` and have `next_updatefee_timer` do nothing so it doesn't start a new poll loop. In addition, instead of having to edit a `fake-bitcoin-cli` maybe we should have a `stop_fee_estimate` function.

Then finally add a 3th `feerates` command option, for example call it `auto`, that restarts `start_fee_estimate` or just checks it is running.

I am happy to give it a try, but I am still a beginner.

PR #1880, opened 2018-08-26
WIP: Have feerates command disable/enable fee estimate polling

SimonVrouwe:
This is a proof of concept of:
- `feerate` command used to add/injects fee rates now first _disables_ the  polling loop and smoothing
- add `feerate` command option `restart-polling` to restart the polling loop
- fixes a bug in #1870  where add/inject `feerates` starts new polling loop, resulting in irregular fee updates
- don't need to rely on `fake-bitcoin-cli` (scripts) to override fee rates

Still has many bugs and needs improvements. Perhaps I am missing something essential why this is NOT a good idea? Would love to get some feedback.

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -317,6 +317,10 @@ static void update_feerates(struct bitcoind *bitcoind,
 	 * 2 minutes. The following will do that in a polling interval
 	 * independent manner. */
 	double alpha = 1 - pow(0.1,(double)topo->poll_seconds / 120);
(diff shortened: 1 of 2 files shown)


rustyrussell:
Usually, there are no fee estimates because bitcoind hasn't been running long enough.  So the idea was you may inject one so you can open a channel / withdraw funds.

But I think that we're better off having a feerate parameter for those explicitly, and use feerate the way you're implying here?

SimonVrouwe:
Thanks for the feedback!

PR #1860, opened 2018-08-21
Replace the fake-bitcoin-cli with a proxy that can be instrumented from tests

cdecker:
Fixes #1858

PR #1887, opened 2018-08-27
Manual feerate options for fundchannel and withdraw

rustyrussell:
Inspired by #1880


SimonVrouwe:
Thanks, nice to see this is being taken care of :+1: . So if I understand correctly:
- _users_  are  not allowed to directly change fee rates related to channel operation
- for test and development, use `fake-bitcoin-cli` or (soon to come) `bitcoind proxy` #1860 

If that is true, then the `feerates` command should only query fee rates and not also add/inject any.
Currently it still does.
https://github.com/ElementsProject/lightning/blob/38ba497bc94fcd8f7be801830939a45794f5b240/lightningd/chaintopology.c#L510
A well I see the code is not finished ... :grin:

SimonVrouwe:
Maybe a comments could be placed with `update_fee` and `next_updatefee_timer` functions that they shouldn't be called directly because they start new recursive timer loops. The only way to get to them is via `start_fee_estimate` which is only called once.

cdecker:
I quickly fixed up the man-page generation and the include lint check. Now `test_feerates` fails with an assertion:

```
assert feerates['perkb']['max_acceptable'] == (2**32 - 1) * 4
assert 4294967292 == (((2 ** 32) - 1) * 4)
```

Might be operator precedence here?

SimonVrouwe:
https://github.com/ElementsProject/lightning/blob/fb0f83a28dc281c957b2c2a8650dd582057b36fa/lightningd/chaintopology.c#L437-L443

`UINT_MAX * 4` doesn't fit in `u32` ?
`feerate_to_style` should return `u64` and needs cast `(u64)feerate_perkw * 4`, that passes the test. But then it fails on the next ones trying to _set_ feerates, but these will be taken out as I understand.

contrib/pylightning/lightning/lightning.py
```diff
@@ -331,13 +331,15 @@ def listpeers(self, peerid=None, level=None):
         }
         return self.call("listpeers", payload)
 
-    def fundchannel(self, channel_id, satoshi):
+    def fundchannel(self, channel_id, satoshi, feerate=None, feeratestyle=None):
```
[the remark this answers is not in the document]
rustyrussell: Ok, agreed. Will add optional extension to numbers.

rustyrussell:
Ok, I think we have consensus that this should be used instead of feerate being adjustable via feerate cmd. (Maybe in future we'll support non-bitcoind fee estimation, then we can revisit).

I will also allow "urgent", "normal" and "slow"as feerates.

rustyrussell:
OK, I've appended more patches for review: I'll want to fold them after that, to avoid API whiplash :)

rustyrussell:
Simply rebase on new master to resolve conflict.

lightningd/chaintopology.c
```diff
@@ -515,7 +520,7 @@ static void json_feerates(struct command *cmd,
 static const struct json_command feerates_command = {
 	"feerates",
 	json_feerates,
-	"Add/query feerate estimates, either satoshi-per-kw ({style} sipa) or satoshi-per-kb ({style} bitcoind) for {urgent}, {normal} and {slow}."
+	"Query feerate estimates, either satoshi-per-kw ({style} sipa) or satoshi-per-kb ({style} bitcoind) for {urgent}, {normal} and {slow}."
```
SimonVrouwe: {urgent}, {normal} and {slow} here are not parameters any more
and the `withdraw_command` help (below) misses the {feerate} parameter

wallet/walletrpc.c
```diff
@@ -101,15 +99,16 @@ static void json_withdraw(struct command *cmd,
...
 		   NULL))
 		return;
 
 	if (!json_tok_wtx(&withdraw->wtx, buffer, sattok, -1ULL))
 		return;
 
 	if (!feerate_per_kw) {
-		command_fail(cmd, LIGHTNINGD, "Cannot estimate fees");
-		return;
+		if (!json_feerate_estimate(cmd, &feerate_per_kw, FEERATE_NORMAL))
```
SimonVrouwe: Was wondering why a seperate `json_feerate_estimate` here and not use a `withdraw_feerate`, similar to `opening_feerate`, `unilateral_feerate`, `mutual_close_feerate` etc.

SimonVrouwe:
A general remark / side note  about `test_no_fee_estimate`.

https://github.com/ElementsProject/lightning/blob/9525cc9029fc9bf6eeaa66ce5fc94fa03a3e0ed2/tests/test_connection.py#L1109-L1112

Test the case where a node has no functioning fee estimate _from the start_. Not sure what the odds are, but if (bitcoind's) fee estimation fails later (after at least one successful fee estimate) `witdraw` and `fundchannel` just use the last successful fee estimate, however old that may be.

For the rest looks pretty cool! :+1:

lightningd/chaintopology.c
rustyrussell: Hmm, those fields are still *returned* though.  I'll clean it up to clarify (as a separate PR, since this has been acked and I'd like to merge).  Similar for the other two...

wallet/walletrpc.c
rustyrussell: That would be good, too.  Will address.

rustyrussell:
@SimonVrouwe Yes, we don't gracefully handle the out-of-date case.  This is unlikely, fortunately: they would have to restart bitcoind underneath us without it reloading the mempool.  Not sure how long we should go before treating our estimates as invalid though; I'll file a separate issue for it however.

PR #1892, opened 2018-08-28
cleanup: derive_basepoints is no longer needed in channel.c

stephaniestroka:


stephaniestroka:
I was looking through the recent changes in regards to separation of the channel and hsmd. Nice progress! Just noticed that you do not need the common/derive_basepoint.h include any longer so here's a tiny contribution :-D

PR #1896, opened 2018-08-30
devtools: dump-gossipstore.

rustyrussell:
Not very useful by itself, but when combined with decodemsg it can tell us quite a bit.

I also noticed that my mainnet node is compacting the store every hour.  I think we should increase the compaction threshold significantly; say from 1.25 to 10?  Or maybe this needs to be combined with something more clever, like deleting by flipping the type of the gossip_store_channel_announce in-place to something else, so we can ignore it on read and don't do extra work setting it up then tearing it down?

### DIFF
--- a/devtools/dump-gossipstore.c
+++ b/devtools/dump-gossipstore.c
@@ -1,0 +1,82 @@
+#include <ccan/crc/crc.h>
+#include <ccan/err/err.h>
+#include <common/type_to_string.h>
+#include <common/utils.h>
+#include <fcntl.h>
+#include <gossipd/gen_gossip_store.h>
+#include <gossipd/gossip_store.h>
+#include <inttypes.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+int main(int argc, char *argv[])
+{
+	int fd;
+	u8 version;
+	beint32_t belen, becsum;
+
+	setup_locale();
+
+	if (argc > 2)
+		errx(1, "Need the filename of a gossip store, or stdin");
+
+	if (argc == 2) {
+		fd = open(argv[1], O_RDONLY);
+		if (fd < 0)
+			err(1, "Opening %s", argv[1]);
+	} else
+		fd = STDIN_FILENO;
+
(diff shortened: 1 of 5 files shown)


devtools/dump-gossipstore.c
```diff
@@ -0,0 +1,76 @@
...
+	int fd;
+	u8 version;
+	beint32_t belen, becsum;
+
+	setup_locale();
+
+	if (argc > 2)
+		errx(1, "Need the filename of a gossip store, or stdin");
+
+	if (argv[1]) {
```
cdecker: Check `argc == 2` instead here.

devtools/dump-gossipstore.c
```diff
@@ -0,0 +1,76 @@
...
+		fd = open(argv[1], O_RDONLY);
+		if (fd < 0)
+			err(1, "Opening %s", argv[1]);
```
cdecker: Adding the error message might be useful here.

devtools/dump-gossipstore.c
```diff
@@ -0,0 +1,76 @@
...
+	} else
+		fd = STDIN_FILENO;
+
+	if (read(fd, &version, sizeof(version)) != sizeof(version))
```
cdecker: We should also abort if the `version` doesn't match any of the versions we support.

devtools/dump-gossipstore.c
rustyrussell: err() does that, errx() is the one which doesn't.  Yeah, not my invention :)

devtools/dump-gossipstore.c
rustyrussell: I'd prefer to warn in that case?

devtools/dump-gossipstore.c
rustyrussell: It's technically equivalent, but OK...

devtools/dump-gossipstore.c
cdecker: Not sure we'd be reading anything meaningful if we upgrade the format later. For example the checksum addition would have resulted in us reading the checksum as the next packet's length. But we can address that once we intro

PR #1899, opened 2018-08-30
The Great Code Documentation Effort, Part I

rustyrussell:
Obviously, I couldn't resist some cleanups along the way.  Part II is finished already, but I wanted feedback on this one.

The idea is to walk through the code and along the way point out the techniques and style we use; as it goes on, it will concentrate more on the specifics of the particular daemon.  It should be read in order, however.


lightningd/lightningd.c
```diff
@@ -101,36 +179,51 @@ static const char *daemons[] = {
...
 	size_t i;
-	for (i = 0; i < ARRAY_SIZE(daemons); i++) {
+
+	/*~ CCAN's ARRAY_SIZE() should always be used on defined arrays: it will
+	 * fail to build if the argument is actually a pointer, not an array! */
+	for (i = 0; i < ARRAY_SIZE(subdaemons); i++) {
 		int outfd;
-		const char *dpath = path_join(tmpctx, ld->daemon_dir, daemons[i]);
+		/*~ CCAN's path module uses tal, so wants a context to allocate
+		 * from.  We have a magic context 'tmpctx' which is freed in
```
renepickhardt: this is very helpful for me especially knowing that it is freed in the eventloop. but I think it needs more elaboration. Also You mention the eventloop here for the first time. I guess the eventloop is a crucial design decission that would be important to understand?

lightningd/lightningd.c
```diff
@@ -101,36 +179,51 @@ static const char *daemons[] = {
...
+		 * the event loop for transient allocations like this. */
+		const char *dpath = path_join(tmpctx, ld->daemon_dir, subdaemons[i]);
 		const char *verstring;
+		/*~ CCAN's pipecmd module is like popen for grownups. */
```
renepickhardt: don't know popen. from the variable name I guess a process id file is created here but I don't know for sure

lightningd/lightningd.c
```diff
@@ -101,36 +179,51 @@ static const char *daemons[] = {
...
 		pid_t pid = pipecmd(&outfd, NULL, &outfd,
 				    dpath, "--version", NULL);
 
+		/*~ Our logging system: spam goes in at log_debug level */
```
renepickhardt: do we have a convtion what stuff is logged at what leve or is that implicit developer knowelege?

lightningd/lightningd.c
```diff
@@ -139,10 +232,16 @@ static bool has_all_daemons(const char* daemon_dir)
 	return !missing_daemon;
 }
 
+/* This routine tries to determine what path the lightningd binary is in.
+ * It's not actually that simple! */
```
renepickhardt: what does "It's not actually that simple!" mean?

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
 			errx(1, "Cannot find %s in $PATH", argv0);
 	}
 
+	/*~ The caller just wants the directory we're in.
+	 *
+	 * Note the magic "take()" macro here: it annotates a pointer as "to
```
renepickhardt: The needs to be moved else where. See the above comment. When I want to study the lightning network and you write that knowing which directory we are in is not trivial I believe you but don't want to learn at this moment how it is being done. However you document a situation where the tal take() takes place (which I did not understand until right now) and I am not even sure if I really understood it correctly. As uggested before I would suggest a MD file which explains some examples and it could point to this function and call and say here you see a non BOLT related usecase of a standard task where it comes into action...

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
...
+	 *
+	 * Many CCAN and our own routines support this, but if you hand a take()
+	 * to a non-take routine unfortunately you don't get a compile error.
+	 */
 	return path_dirname(ctx, take(me));
 }
+
+/*~ This returns the PKGLIBEXEC path which is where binaries get installed.
+ * Note the 'TAKES' annotation which is merely documentation that it will
+ * take ownership of 'my_path' if the caller hands take() there.
```
renepickhardt: ownership means that it is a node of the parent tal pointer and that my_path can now be freed withour resulting in seg fault ?

renepickhardt: and: how are take(...) variable and argument TAKES connected? It must be possible to call take more than once within one function?

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
...
-	if (has_all_daemons(my_path))
+	/* If we're running in-tree, all the subdaemons are with lightningd. */
+	if (has_all_subdaemons(my_path))
 		return my_path;
+
+	/* Otherwise we assume they're in the installed dir. */
 	return find_my_pkglibexec_path(ctx, take(my_path));
 }
 
+/*~ We like to free everything on exit, so valgrind doesn't complain.  In some
```
renepickhardt: it is not the first time you mention valgrind. What is valgrind?

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
...
+ * ways it would be neater not to do this, but it turns out some transient
+ * objects still need cleaning. */
 static void shutdown_subdaemons(struct lightningd *ld)
 {
 	struct peer *p;
 
+	/*~ Because tal objects can be free indirectly, by freeing their parents
+	 * it turns out to be vital to be able to add *destructors* to objects.
```
renepickhardt: you are so loosing me... Where and how are the destructors defined, implemented and even more important registered. I thought the entire philosophy of tal_pointers was that I don't have to worry so much about what is happening when freeing memory. or do you actually mean that `shutdown_subdeamons()` is already the destructor? which is called before freeing `*ld`

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
...
+	 * As a result, freeing them may cause callbacks; in this case, some
+	 * objects freed here can cause database writes, which must be inside
+	 * a transaction */
 	db_begin_transaction(ld->wallet->db);
+
 	/* Let everyone shutdown cleanly. */
 	close(ld->hsm_fd);
+	/*~ The three "global" daemons, which we shutdown explicitly. */
```
renepickhardt: I know I could look in the signature but what does 10 mean?

lightningd/lightningd.c
```diff
@@ -180,47 +280,85 @@ static const char *find_my_path(const tal_t *ctx, const char *argv0)
...
 	subd_shutdown(ld->connectd, 10);
 	subd_shutdown(ld->gossip, 10);
 	subd_shutdown(ld->hsm, 10);
 
+	/* Now we free all the HTLCs */
 	free_htlcs(ld, NULL);
 
+	/*~ For every peer, we free every channel.  Note that the peer has a
+	 * destructor (by convention, called destroy_peer) which removes it
+	 * from the list.  Thus we use list_top() not list_pop() here. */
```
renepickhardt: but `destroy_peer()` is not called in this code block... as above I am confused with those destructors

lightningd/lightningd.c
```diff
@@ -232,20 +370,37 @@ static void shutdown_subdaemons(struct lightningd *ld)
...
+	 * single-threaded, so commits never fail and we don't need
+	 * spin-and-retry logic everywhere. */
 	db_commit_transaction(ld->wallet->db);
 }
 
+/*~ Chainparams are the parameters for eg. testnet vs mainnet.  This wrapper
+ * saves lots of struggles with our 80-column guideline! */
 const struct chainparams *get_chainparams(const struct lightningd *ld)
 {
+	/* "The lightningd is connected to the chain topology."
```
renepickhardt: what exactly is the topology?

lightningd/lightningd.c
```diff
@@ -232,20 +370,37 @@ static void shutdown_subdaemons(struct lightningd *ld)
...
+	 * "The chain topology is connected to the bitcoind API."
+	 * "The bitcoind API is connected chain parameters."
+	 * -- Worst childhood song ever. */
 	return ld->topology->bitcoind->chainparams;
 }
 
+/*~ Our wallet logic needs to know what outputs we might be interested in: we
+ * keep the maximum-ever-used key index in the db, and add them all to the
+ * filter here. */
```
renepickhardt: maybe add a reference to BOLT 2 with the key derivation? For someone new to lightning it might not be clear yet that all keys are derived from a HD wallet

lightningd/lightningd.c
```diff
@@ -254,46 +409,74 @@ static void init_txfilter(struct wallet *w, struct txfilter *filter)
...
+	/*~ Note that tal_fmt() is what asprintf() dreams of being. */
 	pid = tal_fmt(tmpctx, "%d\n", getpid());
+	/*~ CCAN's write_all writes to a file descriptor, looping if necessary
+	 * (which, on a file unlike a socket, is never, for historical UNIX
+	 * reasons).  It also isn't declared with GCC's warn_unused_result
+	 * which write() is when FORTIFY_SOURCE is defined, so we're allowed
+	 * to ignore the result without jumping through hoops. */
 	write_all(pid_fd, pid, strlen(pid));
+
+	/* Leave file open: we close it implicitly when we exit */
```
renepickhardt: because the os does or parent is closed or how do I have to understand this?

lightningd/lightningd.c
```diff
@@ -254,46 +409,74 @@ static void init_txfilter(struct wallet *w, struct txfilter *filter)
...
 }
 
-/* Yuck, we need globals here. */
+/*~ Yuck, we need a global here.
+ *
+ * ccan/io allows overriding the poll() function for special effects: for
+ * lightningd, we make sure we haven't left a db transaction open.  All
+ * daemons which use ccan/io add sanity checks in this loop, so we chain
+ * that after our own override.
+ */
```
renepickhardt: even with the comment don't get what is happening here

lightningd/lightningd.c
```diff
@@ -316,92 +503,94 @@ int main(int argc, char *argv[])
...
-	/* Ignore SIGPIPE: we look at our write return values*/
-	signal(SIGPIPE, SIG_IGN);
-
-	/* Make sure we can reach other daemons, and versions match. */
-	test_daemons(ld);
+	/*~ Make sure we can reach the subdaemons, and versions match. */
+	test_subdaemons(ld);
 
-	/* Initialize wallet, now that we are in the correct directory */
+	/*~ Our "wallet" code really wraps the db, which is more than a simple
```
renepickhardt: what more is it?

lightningd/lightningd.c
```diff
@@ -316,92 +503,94 @@ int main(int argc, char *argv[])
...
+	/*~ We keep a filter of scriptpubkeys we're interested in. */
 	ld->owned_txfilter = txfilter_new(ld);
-	ld->topology->wallet = ld->wallet;
 
-	/* We do extra checks in io_loop. */
+	/*~ This is the ccan/io central poll override from above. */
 	io_poll_debug = io_poll_override(io_poll_lightningd);
 
-	/* Set up HSM. */
+	/*~ Set up HSM: it knows our node secret key, so tells us who we are. */
```
renepickhardt: as the hsmd was mentioned to be the one daemon that needs to be responsive we should explain this a little bit more here?

lightningd/lightningd.c
```diff
@@ -411,51 +600,80 @@ int main(int argc, char *argv[])
...
 
-	/* Initialize block topology (does its own transaction) */
-	setup_topology(ld->topology, &ld->timers, min_blockheight, max_blockheight);
+	/*~ Initialize block topology.  This does its own io_loop to
+	 * talk to bitcoind, so does its own db transactions. */
+	setup_topology(ld->topology, &ld->timers,
+		       min_blockheight, max_blockheight);
 
-	/* Create RPC socket (if any) */
+	/*~ Create RPC socket (if any): now we can talk to clients. */
```
renepickhardt: I believe this one is ncessary  to have lightning-cli. Maybe it is standard stuff for people familar with rpc but how is such an API exposed (sure over unix domain socket) but where in the code are api calls provided? I guess this doesnot happen in jsonrpc so a hint in this comment would be nice.

lightningd/lightningd.c
```diff
@@ -411,51 +600,80 @@ int main(int argc, char *argv[])
...
+	/*~ Now that all the notifications for transactions are in place, we
+	 *  can start the poll loop which queries bitcoind for new blocks. */
 	begin_topology(ld->topology);
 
-	/* Activate crash log now we're not reporting startup failures. */
+	/*~ Setting this (global) activates the crash log: we don't usually need
+	 * a backtrace if we fail during startup. */
 	crashlog = ld->log;
 
+	/*~ The root of every backtrace (almost). */
```
renepickhardt: is this the event loop?

renepickhardt: if yes where does io_loop come from? would elaborate here a little bit

lightningd/lightningd.c
```diff
@@ -482,5 +699,7 @@ int main(int argc, char *argv[])
 	memleak_cleanup();
 #endif
 	daemon_shutdown();
+
+	/*~ Farewell.  Next stop: hsmd/hsm.c. */
```
renepickhardt: why is it actually `lightningd.c` with a d but `hsm.c` without d ?

rustyrussell:
@renepickhardt Thank you*so* much for the detailed review. And sorry I didn't write this s month ago :(

My initial reservations were that I was using too much space with these comments, but I see that my problem was too little, not too much! I'll go through and create a fix-up using your feedback as a guide...

doc/HACKING.md
```diff
@@ -14,24 +14,10 @@ Getting Started
 It's in C, to encourage alternate implementations.  Patches are welcome!
 You should read our [Style Guide](STYLE.md).
 
-To read the code, you'll probably need to understand `ccan/tal`: it's a
```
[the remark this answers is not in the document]
rustyrussell: OK, I'll move this to the first occurrence in the code then.

lightningd/lightningd.c
```diff
@@ -18,11 +48,14 @@
 #include <ccan/tal/grab_file/grab_file.h>
 #include <ccan/tal/path/path.h>
 #include <ccan/tal/str/str.h>
+
+/*~ This is common code: routines shared by one or more programs. */
```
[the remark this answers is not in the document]
rustyrussell: lightning-cli in at least one case... I've expanded.

lightningd/lightningd.c
```diff
@@ -39,43 +72,86 @@
...
+	/*~ Behaving differently depending on environment variables is a hack,
+	 * *but* hacks are allowed for dev-mode stuff.  In this case, there's
+	 * a significant overhead to the memory leak detection stuff, and
+	 * we can't use it under valgrind, so the test harness uses this var
+	 * to disable it in that case. */
 	if (getenv("LIGHTNINGD_DEV_MEMLEAK"))
 		memleak_init();
 #endif
 
+	/*~ These are CCAN lists: an embedded double-linked list.  It's not
```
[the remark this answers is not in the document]
rustyrussell: I'd prefer to simple put an example inline if we need it.  Having a single go-to place to read is always better than  some out-of-the-way file nobody will read or maintain...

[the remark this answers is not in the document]
rustyrussell: I ended up just listing some helpers and fleshing it out a bit more. I'd rather reference them as we see their use.  I tried to create a minimal example, but it was contrived and still 20 lines :(

lightningd/lightningd.c
```diff
@@ -39,43 +72,86 @@
...
+	 * really typesafe, but relies on convention to access the contents.
+	 * It's inspired by the closely-related Linux kernel list.h. */
 	list_head_init(&ld->peers);
+
+	/*~ These are hash tables of incoming and outgoing HTLCs (contracts) */
 	htlc_in_map_init(&ld->htlcs_in);
 	htlc_out_map_init(&ld->htlcs_out);
+
+	/*~ We have a log-book infrastructure: we define a 20MB log book and
+	 * point our log objects into it. */
```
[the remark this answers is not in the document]
rustyrussell: Expanded.

lightningd/lightningd.c
```diff
@@ -39,43 +72,86 @@
...
 	ld->log_book = new_log_book(20*1024*1024, LOG_INFORM);
+	/*~ Note the tal context arg (by convention, the first argument to any
```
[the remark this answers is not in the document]
rustyrussell: We'll get to tmpctx in hsmd, actually.

lightningd/lightningd.c
rustyrussell: Changed to: "It's more complex than you might expect."

lightningd/lightningd.c
rustyrussell: Agreed.  However, I will resolve this differently: by moving the unnecessary bits into subd.c, and so this is just a simple example, with take().

lightningd/lightningd.c
rustyrussell: Enhanced, I hope I covered that now.

lightningd/lightningd.c
rustyrussell: OK, I added "(an awesome runtime memory usage detector for C and C++ programs)" to the first reference.

lightningd/lightningd.c
rustyrussell: OK, how's this:

	/*~ tal supports *destructors* using `tal_add_destructor()`; the most
	 * common use is for an object to delete itself from a linked list
	 * when it's freed.
	 *
	 * As a result, freeing an object (which frees any tal objects
	 * allocated off it, and any allocated off them, etc) may cause
	 * callbacks; in this case, some objects freed here can cause database
	 * writes, which must be inside a transaction. */

lightningd/lightningd.c
rustyrussell: "blockchain" :)

lightningd/lightningd.c
rustyrussell: Ack.

lightningd/lightningd.c
rustyrussell: OS.  fixed.

lightningd/lightningd.c
rustyrussell: Good point.  That's bugged me for a while too, so I renamed them all...

lightningd/lightningd.c
```diff
@@ -39,43 +78,123 @@
...
+	 * The most common operations are list_head_init(), list_add(),
+	 * list_del() and list_for_each().
+	 *
+	 * This method of manually declaring the list hooks avoids dynamic
+	 * allocations to put things into a list. */
 	list_head_init(&ld->peers);
+
+	/*~ These are hash tables of incoming and outgoing HTLCs (contracts),
+	 * defined as `struct htlc_in` and `struct htlc_out`in htlc_end.h.
+	 * The hash tables are declared ther using the very ugly
```
practicalswift: Typo: "ther"

lightningd/lightningd.c
```diff
@@ -39,43 +72,86 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-char *bitcoin_datadir;
-
-int pid_fd;
-
+/*~ The core lightning object: it's passed everywhere, and is basically a
+ * global variable.  This new_xxx pattern is something we'll see often:
+ * it allocates and initializes a new structure, using *tal*, the heirarchitcal
```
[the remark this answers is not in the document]
cdecker: Shall we call `lightningd` the context object, or maybe root of the entire program?

lightningd/lightningd.c
```diff
@@ -39,39 +72,86 @@
...
+ * allocator. */
 static struct lightningd *new_lightningd(const tal_t *ctx)
 {
+	/*~ tal: each allocation is a child of an existing object (or NULL,
+	 * the top-level object).  When an object is freed, all the objects
+	 * 'tallocated' off it are also freed.  In this case, freeing 'ctx'
+	 * will free 'ld'.
+	 *
+	 * It's incredibly useful for grouping object lifetimes, as we'll see.
```
cdecker: "For the technically inclined: `tal` allocations usually build a tree, and `tal_free`ing any node in the tree will result in the entire subtree rooted at that node to be freed.

Be careful though that a any cycle in the allocation graph may end up causing an infinite loop."

Also for my personal understanding: the allocations are freed in post-order right? (first children, then the parent).

lightningd/lightningd.c
```diff
@@ -39,43 +72,86 @@
...
 	if (getenv("LIGHTNINGD_DEV_MEMLEAK"))
 		memleak_init();
 #endif
 
+	/*~ These are CCAN lists: an embedded double-linked list.  It's not
+	 * really typesafe, but relies on convention to access the contents.
+	 * It's inspired by the closely-related Linux kernel list.h. */
 	list_head_init(&ld->peers);
+
+	/*~ These are hash tables of incoming and outgoing HTLCs (contracts) */
```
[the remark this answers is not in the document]
cdecker: Maybe a reference to https://github.com/ElementsProject/lightning/blob/50efe34e3b247def991cd6bdd03deee1ce7d1b85/lightningd/htlc_end.h#L104-L108 helps?

lightningd/lightningd.c
```diff
@@ -251,26 +409,40 @@ static void init_txfilter(struct wallet *w, struct txfilter *filter)
...
-	/* Move back: important, since lightning dir may be relative! */
+	/*~ Move back: important, since lightning dir may be relative! */
 	if (chdir(cwd) != 0)
 		fatal("Could not return to directory %s: %s",
 		      cwd, strerror(errno));
 
 	db_reopen_after_fork(ld->wallet->db);
+
+	/*~ Why not allocate cwd off tmpctx?  Probably because this code predates
+	 * tmpctx.  So we free manually here. */
```
cdecker: I thought the it's more because the global `tmpctx` is an event loop concept and this is outside of the loop.

lightningd/lightningd.c
rustyrussell: It won't loop on cycles.  Freeing is done in two phases: object is marked destroyed, destructor called, then recurse into children, then finally free(object).  So destructors are pre-order, deallocation is post-order.

It's common to want to access parents in a destructor (eg. delete from its linked list), so deallocation has to be deferred.

lightningd/lightningd.c
rustyrussell: Yeah, but it's before the loop; we tend to allow that elsewhere...

rustyrussell:
Despite lack of (concrete) acks, I'm going to merge now; it's going to cause rebase hell to delay further, and I'm already itching to post parts II and III :)

lightningd/lightningd.c
cdecker: Currently the JSON-RPC autoregisters callbacks using the `ccan/autodata` module, however that poses some issues if we'd like to use `clang` based sanitizers since it does weird things with uninitialized data, so maybe we'd want to ween off that anyway at some point.

jb55:
Knuth would be proud. Thanks rusty!

PR #1908, opened 2018-09-04
Fix gossip store loading on real nodes.

rustyrussell:


common/subdaemon.c
```diff
@@ -45,10 +41,11 @@ void subdaemon_setup(int argc, char *argv[])
 	/* From debugger, set debugger_spin to 0. */
 	for (int i = 1; i < argc; i++) {
 		if (streq(argv[i], "--debugger")) {
-			fprintf(stderr, "gdb -ex 'attach %u' -ex 'p debugger_connected=1' %s\n",
-				getpid(), argv[0]);
-			while (!debugger_connected)
-				usleep(10000);
+			char *cmd = tal_fmt(NULL, "${DEBUG_TERM:-gnome-terminal --} gdb -ex 'attach %u' %s &", getpid(), argv[0]);
```
cdecker: Nice one :clap: I was hoping to autostart a debugger at some point, but I was looking into parsing the output. This is much nicer, and still allows me to integrate with `tmux`!

PR #1909, opened 2018-09-04
Fix last saved commit if multiple htlcs

rustyrussell:


Makefile
```diff
@@ -35,7 +35,7 @@ endif
 
 ifeq ($(COMPAT),1)
 # We support compatibility with pre-0.6.
```
ZmnSCPxj: Version in comment may need updating to pre-0.6.1?

PR #1910, opened 2018-09-04
pytest: Fix configure to find pytest when installed using pip3.

lucash-dev:
Running `make pytest` when pytest is not installed, results in a message instructing the user to install it using `pip3`.
However, after installing it using `pip3`, `configure` still can't find it.

This PR addresses this issue by letting `configure` find if the module is accessible to `python3` and setting the command as `python3 -m pytest`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,7 +216,7 @@ check:
 
 pytest: $(ALL_PROGRAMS)
 ifndef PYTEST
-	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt'"
+	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt', and rerun 'configure'."
 	exit 1
 else
 # Explicitly hand DEVELOPER and VALGRIND so you can override on make cmd line.
--- a/configure
+++ b/configure
@@ -72,10 +72,16 @@ find_pytest()
 	fi
     done
 
-    if python3 -c "import pytest" ; then
-        echo "python3 -m pytest"
-        return
-    fi
+    PYTHON_BINS="python python3"
+    for p in $PYTHON_BINS; do
+        if [ "$(which $p)" != "" ] ; then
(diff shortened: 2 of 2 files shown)


configure
```diff
@@ -71,6 +71,11 @@ find_pytest()
 	    return
 	fi
     done
+
+    if python3 -c "import pytest" ; then
+        echo "python3 -m pytest"
```
ZmnSCPxj: I notice that the before loop iterates over multiple possible namings for `pytest`.  I wonder, might `python3` similarly be named differently depending on distribution? Should we not also iterate among possible `PYTHON_BINS`?

Makefile
```diff
@@ -216,7 +216,7 @@ check:
 
 pytest: $(ALL_PROGRAMS)
 ifndef PYTEST
-	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt'"
+	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt', and rerun 'config'."
```
ZmnSCPxj: The script is `configure` and not `config`, is it not?

cdecker:
The test failure seems to be a flake btw.

configure
cdecker: If we are, we might need to also check that `python` actually is a version 3 and not a version 2 interpreter, so that adds a bit more work.

ZmnSCPxj: I would suggest doing so, although I am uncertain how many distributions would have a version 3 python that is in `python` but not in `python3`.  This is acceptable as-is, although perhaps with a `# FIXME: find out if a Python 3 could be in python but not in python3`.

Makefile
lucash-dev: :man_facepalming: of course you're right. will fix it.

lucash-dev:
thanks for the prompt and thorough feedback! I think I addressed all comments.

rustyrussell:
Great commit, thanks!!

PR #1913, opened 2018-09-06
Make chainparams available in channeld

cdecker:
This is a replacement of #1842, which passes a reference to the `chainparams` to the `channel` so that we can check various constraints during runtime, instead of having global constants and magic values in the code.

The fact that these become configurable is really just a side-effect.


jb55:
utACK cca47fd11681b5bc26c3a2e5277a82b23910400e

rustyrussell:
Why don't we use the genesis block hash elsewhere, rather than having the arbitrary id?

common/initial_channel.c
```diff
@@ -58,6 +60,8 @@ struct channel *new_initial_channel(const tal_t *ctx,
 	channel->commitment_number_obscurer
 		= commit_number_obscurer(&channel->basepoints[funder].payment,
 					 &channel->basepoints[!funder].payment);
+	channel->chainparams = chainparams_by_chainhash(chain_hash);
+	assert(channel->chainparams != NULL);
```
bisoge: Per Bolt #2 the receiver must reject the channel if the chain hash is unknown. Do we really just want to assert here?

cdecker: Excellent point, I'm making this an `if (!channel->chainparams) return tal_free(channel);` :+1:

channeld/full_channel.h
```diff
@@ -81,6 +82,7 @@ u32 actual_feerate(const struct channel *channel,
 /**
  * channel_add_htlc: append an HTLC to channel if it can afford it
  * @channel: The channel
+ * @chain_hash: genesis hash of the target blockchain
```
bisoge: This comment can be removed, leftover from #1842

cdecker:
Rebased and squashed one fixup that fixed a doc comment.

cdecker:
Re-applying @rustyrussell's ACK after rebase

ACK 7365165

PR #1922, opened 2018-09-09
Weekly cleanups

cdecker:
- Moves JSON-RPC setup below the pid-file
 - Fixes the builder with the missing dependencies for a2x which was failing on travis some times.

### DIFF
--- a/contrib/Dockerfile.builder
+++ b/contrib/Dockerfile.builder
@@ -13,6 +13,7 @@ RUN apt-get -qq update && \
 	automake \
 	clang \
 	cppcheck \
+	docbook-xml \
 	shellcheck \
 	eatmydata \
 	software-properties-common \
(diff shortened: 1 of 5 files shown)


cdecker:
Added one more commit that @rustyrussell requested.

ZmnSCPxj:
ACK af98e7d5dee522375b8620559ebaddc4dc9cdd73

rustyrussell:
Suggest splitting pidfile handling instead: better to do everything which could fail *before* we daemonize and can't print to stderr any more...

cdecker:
> Suggest splitting pidfile handling instead: better to do everything which could fail before we daemonize and can't print to stderr any more...

Roger :+1:

cdecker:
ACK eadc570

PR #1924, opened 2018-09-10
ping: First parameter is called 'id' not 'peerid', fix help

laanwj:
The first argument of 'ping' was documented as 'peerid', however internally it is expected to be just 'id'.

To avoid breaking the API, opt to fix the documentation.


wythe:
ACK

niftynei:
ACK 53d5e28

PR #1930, opened 2018-09-11
Wallet makefile improvements

rustyrussell:


lightningd/Makefile
```diff
@@ -100,6 +100,8 @@ LIGHTNINGD_HEADERS_GEN =			\
 
 ALL_GEN_HEADERS += $(LIGHTNINGD_HEADERS_GEN)
 
+include wallet/Makefile
```
ZmnSCPxj: Maybe a comment for why this is here (whereas all other top-level dir/Makefile are in top-level Makefile).

PR #1931, opened 2018-09-11
Fixup reference to peer-id in JSON rpc

niftynei:
Fix-up peerid name references in all JSON rpc.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/1931/commits/f6993692bbede4c938ee71125f97c564a599ba4f

PR #1944, opened 2018-09-17
Openingd: prioritize incoming peer traffic over sending out gossip

### LINKED ISSUE #1943: deadlock in openingd caused by gossip traffic
Two nodes (both) 36eab5d on my lan connected to each other, the `ping` command hangs in both directions.
Issue seems related to `gossip` traffic because deleting `gossip_store` file solves it. Connecting a debugger to `openingd` shows that the daemon blocks on a `write` call to the peer socket.
```
#0  0x00007f5d2518c730 in __write_nocancel () at ../sysdeps/unix/syscall-template.S:84                                                                                                                               
#1  0x0000564bcc8c4817 in write_all (fd=3, data=0x564bcd2127f8, size=164) at ccan/ccan/read_write_all/read_write_all.c:11                                                                                            
#2  0x0000564bcc89518d in sync_crypto_write (cs=0x564bcd212038, fd=3, msg=0x564bcd212668) at common/crypto_sync.c:44                                                                                                 
#3  0x0000564bcc89c704 in handle_gossip_msg (peer_fd=3, cs=0x564bcd212038, msg=0x564bcd2125a8 "\v\310") at common/read_peer_msg.c:95                                                                                 
#4  0x0000564bcc8931ef in handle_gossip_in (state=0x564bcd212038) at openingd/openingd.c:1023                                                                                                                        
#5  0x0000564bcc893823 in main (argc=2, argv=0x7ffc58acc7f8) at openingd/openingd.c:1161 
```
Checking the socket buffers shows Recv-Q and Send-Q are not really empty.
```
ss -4 -t -p | egrep "lightning|State"
State      Recv-Q Send-Q Local Address:Port                 Peer Address:Port                
ESTAB      354111 386232 192.168.1.147:60100                192.168.1.184:9738                  users:(("gdb",pid=8673,fd=3),("konsole",pid=8669,fd=3),("lightning_openi",pid=8667,fd=3))
```
This is the case on both nodes, so it looks like some kind of deadlock: both sides block on `write` because the send buffer is full and therefore cannot receive. Related to and likely the cause for #1928, #1920 as these where solved by deleting `gossip_store` files.

Not sure what causes the simultaneous sending of so much gossip that it fills up the Send-Q (at least on one side), but this has only happened between nodes on my LAN (520 Mbit/s wifi) and with certain `gossip_store` files.

SimonVrouwe:
openingd: prioritize incoming peer traffic over handling (and sendind out) gossip

- reduces probability for a deadlock where we block on sending data because
  the other peer cannot receive because it blocks on sending data etc.
- when either side sends so much data that it fills up the kernel/network buffer
- however sending out gossip can still block when (malicious) peer never receives

Fixes #1943 
thanks to the advice of @rustyrussell

### DIFF
--- a/openingd/openingd.c
+++ b/openingd/openingd.c
@@ -1157,10 +1157,11 @@ int main(int argc, char *argv[])
 		 * don't try to service more than one fd per loop. */
 		if (pollfd[0].revents & POLLIN)
 			msg = handle_master_in(state);
-		else if (pollfd[1].revents & POLLIN)
-			handle_gossip_in(state);
 		else if (pollfd[2].revents & POLLIN)
 			msg = handle_peer_in(state);
+		else if (pollfd[1].revents & POLLIN)
+			handle_gossip_in(state);
+
 		clean_tmpctx();
 	}
 


SimonVrouwe:
Thanks for the feedback. 
It's not a perfect solution and a bid ugly, maybe a better option is to make the peer socket `O_NONBLOCK` or `SOCK_NONBLOCK`, but I think it requires significant more rewriting of code.

> You can still block writing to the peer, since we will write the entire packet.

Yes. To solve that `PEER_FD` should be a non-blocking socket, see above.

> This will make the code spin consuming 100% CPU.

Fixed that with a flag `canwritepeer`.

> I think you might be wanting to prioritize handling peer traffic over handling gossipd traffic ...

Reordered it, but wasn't this (`handle_peer_in`) placed at the bottom as some form DOS protection?

SimonVrouwe:
> I think you might be wanting to prioritize handling peer traffic over handling gossipd traffic

You are right, just tested this and it works as effective and is much simpler. Will push it.

rustyrussell:
Sorry about my initial comment; the tone was bad :(  I'll blame jetlag but you still deserve better.

Yeah, the real fix is more complex; we're playing fast and loose with buffering assumptions to make the current simple approach work (in particular, we assume a single packet we write will be less than the socket buffer size, which is always true currently).

But if we're going to do non-blocking read, I'm thinking of optimizing and reading as much as we can at once, rather than reading header then body separately.  That means the entire buffer has to get handed to the next daemon on transition.  All possible, but more code than your simple fix!

rustyrussell:
So the remaining cases (we block and don't read gossip) are correct: the gossipd is written not to infinitely buffer in such cases, and so the peer is just punishing themselves.

SimonVrouwe:
No need to apologize! I find your feedback concise and you where right, please keep them coming. My initial commit was sloppy and somewhat hasted, I was just excited that I found the problem :smile:

PR #1945, opened 2018-09-18
Don't accept announcements for recently-closed channels.

rustyrussell:
Dragging in `gossipfrom` (now `gossipwith`) is a hack, but it's hard to test this without writing significant new infrastructure...

### DIFF
--- a/tests/test_gossip.py
+++ b/tests/test_gossip.py
@@ -823,10 +823,6 @@ def test_gossip_store_load(node_factory):
     l1 = node_factory.get_node(start=False)
     with open(os.path.join(l1.daemon.lightning_dir, 'gossip_store'), 'wb') as f:
         f.write(bytearray.fromhex("03"  # GOSSIP_VERSION
-                                  "00000099"  # len
-                                  "12abbbba"  # csum
-                                  "1002"  # WIRE_GOSSIP_STORE_NODE_ANNOUNCEMENT
-                                  "00950101cf5d870bc7ecabcb7cd16898ef66891e5f0c6c5851bd85b670f03d325bc44d7544d367cd852e18ec03f7f4ff369b06860a3b12b07b29f36fb318ca11348bf8ec00005aab817c03f113414ebdc6c1fb0f33c99cd5a1d09dd79e7fdf2468cf1fe1af6674361695d23974b250757a7a6c6549544300000000000000000000000000000000000000000000000007010566933e2607"
                                   "000001bc"  # len
                                   "521ef598"  # csum
                                   "1000"  # WIRE_GOSSIP_STORE_CHANNEL_ANNOUNCEMENT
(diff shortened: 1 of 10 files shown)


tests/test_gossip.py
```diff
@@ -876,13 +882,34 @@ def test_gossip_notices_close(node_factory, bitcoind):
     wait_for(lambda: len(l1.rpc.listnodes()['nodes']) == 2)
     l1.rpc.disconnect(l2.info['id'])
 
+    # Grab channel_announcement from io logs (ends in ')
+    channel_announcement = l1.daemon.is_in_log('\[IN\] 0100').split(' ')[-1][:-1]
+    channel_update = l1.daemon.is_in_log('\[IN\] 0102').split(' ')[-1][:-1]
+    node_announcement = l1.daemon.is_in_log('\[IN\] 0101').split(' ')[-1][:-1]
```
cdecker: Are those `[:-1]` to strip a newline? If yes, please use `.strip()` instead, that's more explicit.

lightningd/gossip_control.c
```diff
@@ -71,13 +71,13 @@ static void get_txout(struct subd *gossip, const u8 *msg)
 
 	op = wallet_outpoint_for_scid(gossip->ld->wallet, scid, scid);
 
-	if (op) {
+	if (op && !op->spendheight) {
```
cdecker: We could also just push that down to the sqlite query by adding an argument `includespent`. That'd allow us to skip materializing the result in memory just to drop it again immediately.

cdecker: Also `op` is allocated off of `scid` which could be a long-lived object (it currently isn't).

cdecker:
Just some minor comments, I'd like to push the spentness check into the query, since it makes the checks easier and avoids allocating memory if we don't need it.

rustyrussell:
Agreed.  Rewrite coming...

wallet/wallet.c
```diff
@@ -2139,7 +2139,8 @@ struct outpoint *wallet_outpoint_for_scid(struct wallet *w, tal_t *ctx,
 			  "FROM utxoset "
 			  "WHERE blockheight = ?"
 			  " AND txindex = ?"
-			  " AND outnum = ?");
+			  " AND outnum = ?"
+			  " AND spendheight IS NOT NULL");
```
cdecker: Isn't this inverted? We're interested in the ones that aren't spent yet.

rustyrussell: That would explain the test breakage :)

rustyrussell:
1. Fixed the inverted case.  Thanks!
2. Folded the two patches together, so there's no 'add code in caller, then remove it again to do it in db'.
3. Rebased on master to avoid spurious unrelated travis error.

cdecker:
Nice work @rustyrussell :rocket: 

ACK 64d70e182c4e8bdb719c42c69bb96b4dd69b02e0

PR #1948, opened 2018-09-19
Fix flaky  test_no_fee_estimate

rustyrussell:


tests/utils.py
```diff
@@ -560,6 +560,10 @@ def is_channel_active(self, chanid):
         active = [(c['short_channel_id'], c['flags']) for c in channels if c['active']]
         return (chanid, 0) in active and (chanid, 1) in active
 
+    def wait_for_channel_onchain(self, peerid):
+        txid = only_one(only_one(self.rpc.listpeers(peerid)['peers'])['channels'])['scratch_txid']
+        wait_for(lambda: txid in self.bitcoin.rpc.getrawmempool())
```
cdecker: This assumes we haven't mined it yet. Could extend this later to include UTXO checks as well.

cdecker: Should be good for now though.

PR #1949, opened 2018-09-19
sendpay should fail when the payment fails.

rustyrussell:
Not return a fake "payment in progress".  That seems weird.

It's also clearer to do the callbacks in the same place as the general wait callbacks, not where we write into the db.

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -72,23 +72,32 @@ add_waitsendpay_waiter(const tal_t *cxt,
 	tal_add_destructor(pc, destroy_sendpay_command);
 }
 
-/* Caller responsible for freeing ctx. */
-static void waitsendpay_resolve(const tal_t *ctx,
-				struct lightningd *ld,
-				const struct sha256 *payment_hash,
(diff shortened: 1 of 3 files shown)


ZmnSCPxj:
If I remember correctly, the current system, `sendpay` dispatches only, and does not wait for a response from the counterparty.
You are supposed to use `waitsendpay` to actually wait for the payment to resolve into either an error or a success.

Before, we would wait for a response from the first hop node before responding to `sendpay`.
In current code, as soon as we send the commitment signature, we trigger `sendpay`.
I believe this patch returns the old behavior.
This is not ideal for wallet software that uses `sendpay` in a single-threaded environment, as I believe we already promised that `sendpay` is asynchronous.

ZmnSCPxj:
Here: https://github.com/ElementsProject/lightning/issues/1088#issuecomment-369904740

rustyrussell:
Agreed. I'd like immediate failure in the case where the first channel is completely unavailable, but that's a corner case.

PR #1950, opened 2018-09-19
json: accept > 100% for both getroute fuzzpercent and pay maxfeepercent.

rustyrussell:
There's no reason for these to be limited to 100%.  I can't remember who mentioned this to me :(


rustyrussell:
@wythe counterintuitively, I think this change is correct...

cdecker:
Damn, you beat me to the punch (#1951 does the same). But you have a unit-test failure: the last test returns 1000 not 0.

PR #1951, opened 2018-09-19
Remove 100% upper limit for percentage arguments

### LINKED ISSUE #1914: Set maxfeepercent to an arbitrary value
With the purpose of testing some extremely tiny payments, I tried to set the `maxfeepercent` value above 100%, and obtained the following output:
```
{ "code" : -32602, "message" : "'maxfeepercent' should be a double in range [0.0, 100.0], not '1000'", "data" : 
{
  "maxfeepercent": "1000"
} }
```
Is there a way with `lightning-cli` to avoid the 100% limit and set an arbitrary `maxfeepercent` value?

cdecker:
The `json_tok_percentage` parser is used for the `fuzzpercent` in `getroute` and
`maxfeepercent` in `pay`. In both cases it seems reasonable to allow values
larger than 100%. This has bitten users in the past when they transferred single
satoshis to things like satoshis.place over a route longer than 2 hops.

Fixes #1914

### DIFF
--- a/doc/lightning-getroute.7
+++ b/doc/lightning-getroute.7
@@ -2,12 +2,12 @@
 .\"     Title: lightning-getroute
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 04/26/2018
+.\"      Date: 09/17/2018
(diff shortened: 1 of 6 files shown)


wythe:
ACK https://github.com/ElementsProject/lightning/pull/1951/commits/abb97e9e8fed92ceb283e19cb16e8481a7d8462b

Issue #1938, opened 2018-09-14
addr=autotor:localhost doesn't parse


rustyrussell:
```
$ /home/rusty/lightning/lightningd/lightningd 
lightning config file line 5: addr: Unable to parse address 'autotor:localhost': Error parsing hostname
...
$ /home/rusty/lightning/lightningd/lightningd 
lightning config file line 5: addr: Unable to parse address 'autotor:127.0.0.1': Error parsing hostname
```

Yet man page has:
```
*addr*='[IPADDRESS[:PORT]]|autotor:TORIPADDRESS[:TORPORT]'::
```
Implying TORPORT is optional, and it also says:

```
If necessary, and 'always-use-proxy' is not specified, a DNS lookup may be done to resolve 'IPADDRESS' or 'TORIPADDRESS'.
```

PR #1952, opened 2018-09-19
parse autotor: address before separate_address_and_port

Saibato:
this enables addr like --addr=autotor:127.0.0.1 or
--addr=autotor:localhost to just use the default tor service port
fixes: #1938

Signed-off-by: Saibato <Saibato.naga@pm.me>


cdecker:
ACK 35482dd

PR #1957, opened 2018-09-20
pytest: wait until mock is called for set_feerates.

rustyrussell:
Got a spurious failure in test_no_fee_estimate; we fired too soon from the logs (presumably
we raced in on the first response, but estimatesmartfee gets called 3 times).


cdecker:
ACK 805fefe

PR #1959, opened 2018-09-20
pytest: more reliable onchain tests, ie. YA Travis Failure

rustyrussell:
We extract the tx from the logs, and then we wait until that hits
the mempool.  This is more reliable than 'sendrawtx' in the logs,
which might catch a previous sendrawtx; it's also more explicit
that we expect that tx exactly.


cdecker:
> Agreed; if we get more cases, we'll want a wait_for_onchaind_broadcasts I think. I guess I got sick of fixing tests ;)

Totally understandable, I'll spend some cycles on the test infrastructure to get them up to speed again.

PR #1963, opened 2018-09-20
Help usage

wythe:
The help command now adds command usage to its output.

Instead of seeing, for example:

    decodepay
        Decode {bolt11}, using {description} if necessary

we see:

    decodepay bolt11 [description]
        Decode {bolt11}, using {description} if necessary

Note: Removed discussion of `check` command.  It was removed and saved for later patch.

This patch is a big step toward allowing us to write tests that:
1. Verify the manpage SYNOPSIS is correct.
2. Exhaustively test the API.

The first three commits adds the infrastructure.  The approach I took
is to introduce different command modes.  And the behavior of `param()` changes based on the mode.

Inspired-By: discussions with @rustyrussell

### DIFF
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -78,6 +78,9 @@ static void json_stop(struct command *cmd,
 {
 	struct json_result *response = new_json_result(cmd);
 
+	if (!param(cmd, buffer, params, NULL))
+		return;
+
 	/* This can't have closed yet! */
 	cmd->jcon->stop = true;
 	json_add_string(response, NULL, "Shutting down");
@@ -121,6 +124,9 @@ AUTODATA(json_command, &dev_rhash_command);
 static void json_crash(struct command *cmd UNUSED,
 		       const char *buffer UNUSED, const jsmntok_t *params UNUSED)
 {
+	if (!param(cmd, buffer, params, NULL))
+		return;
+
 	fatal("Crash at user request");
 }
 
@@ -136,6 +142,9 @@ static void json_getinfo(struct command *cmd,
 			 const char *buffer UNUSED, const jsmntok_t *params UNUSED)
 {
 	struct json_result *response = new_json_result(cmd);
+
+	if (!param(cmd, buffer, params, NULL))
+		return;
 
 	json_object_start(response, NULL);
 	json_add_pubkey(response, "id", &cmd->ld->id);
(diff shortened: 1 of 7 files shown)


lightningd/jsonrpc.c
```diff
@@ -231,11 +234,24 @@ static void json_help(struct command *cmd,
...
 	json_array_start(response, "help");
 	for (i = 0; i < num_cmdlist; i++) {
+		char *usage = tal_strdup(cmd, cmdlist[i]->name);
+		/* We have to skip all the ones that don't call param() in
+		 * their handler. */
+		if (strcmp(cmdlist[i]->name, "help")) {
+			ok = true;
+			cmdlist[i]->dispatch(cmd, NULL, NULL);
+			if (!ok)
+				return;
```
rustyrussell: This looks really weird.  I'm not sure how we ever set cmd->ok to false.  Actually, seems we don't?  Maybe remove the field altogether...

rustyrussell: Ah, I see it later.  Still not sure when help() would fail in a command though?

cli/lightning-cli.c
```diff
@@ -282,10 +311,10 @@ int main(int argc, char *argv[])
...
 		tal_append_fmt(&cmd, "} }");
 	} else {
-		tal_append_fmt(&cmd, "[ ");
-		for (i = 2; i < argc; i++)
-			add_input(&cmd, argv[i], i, argc);
-		tal_append_fmt(&cmd, "] }");
+		if (streq(method, "check"))
+			default_check_input(&cmd, argv, argc);
+		else
+			default_input(&cmd, argv, argc);
```
rustyrussell: This is a bit awkward, and doesn't allow the keyword case.

Perhaps we're better off doing this work in the server; hoist 'cmd' out of params manually there?  The main reason for that is because there are other clients, like the python library.

lightningd/jsonrpc.c
wythe: You're right. I over thought this.

cli/lightning-cli.c
wythe: Agreed!  This was the easiest way to get the feature in and see if you  liked it.  I'll get to work on manually constructing the command in the check handler next.

wythe:
Good review.  No need to debate anything!  Fixups are in except for the last one.

rustyrussell:
Ack.  Can we drop the 'check' commit for later, and roll the rest of the fixes in to the specific patches, so it'll be a clean history for me to apply?  Thanks!

wythe:
Done.

PR #1964, opened 2018-09-21
Makefile: don't rebuild all the time

rustyrussell:
The code to regenerate the local BOLT copy was causing eternal rebuild.

So only build if it's wrong.


cdecker:
I was wondering whether me adding a new file was causing this, but glad you found & fixed the issue.

ACK 6ab2a8c

PR #1966, opened 2018-09-22
json: Support streaming JSON messages

cdecker:
It turns out we were heavily relying on the fact that after each message from
the client there'd be a flush, and that there would not be anything after the
JSON object we read. This will no longer be the case once we start streaming
things or we are very quick in issuing the JSON-RPC requests.

This just takes one of the error paths (incomplete read) and makes it into a
successful path if we have indeed read a full root element.

### DIFF
--- a/common/json.c
+++ b/common/json.c
@@ -187,7 +187,10 @@ jsmntok_t *json_parse_input(const char *input, int len, bool *valid)
 	jsmntok_t *toks;
 	int ret;
 
-	toks = tal_arr(input, jsmntok_t, 10);
+	/* Zero out so we can count elements correctly even on incomplete reads
+	 * (when jsmn_parse returns -1). This results in all toks being of type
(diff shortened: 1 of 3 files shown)


rustyrussell:
OK, so the problem is that we get JSMN_ERROR_PART for `{}{` rather than having it stop parsing on the first top-level complete object and returning 2.   In practice we'd eventually get a whole object, so we could make the caller loop on all the tokens in case we get multiple commands at once, but that's (1) horrible, and (2) interferes with a patch I have pending to ensure we only process one JSON RPC command at once.

Better would be to change the semantics of jsmn_parse() or have a jsmn_parse_single() which did the Right Thing.  The JSMN code is ... dense... but I've appended to an existing issue here: https://github.com/zserge/jsmn/issues/80#issuecomment-423793934

PR #1968, opened 2018-09-24
Gossipd channel improvements

rustyrussell:
This was in my part IV Doc, but it's too good not to push this simplification now.  It will significantly reduce network update spam (and would have been a better way in the first place than the delay code).  Part of my brain says @cdecker suggested this, but I could be wrong...


gossipd/gossipd.c
```diff
@@ -1043,201 +953,191 @@ static u8 *create_channel_update(const tal_t *ctx,
...
+			queue_peer_msg(peer, update);
+	}
+
+	msg = handle_channel_update(daemon->rstate, take(update), caller);
+	if (msg)
+		status_failed(STATUS_FAIL_INTERNAL_ERROR,
+			      "%s: rejected local channel update %s: %s",
+			      caller,
+			      /* This works because handle_channel_update
+			       * only steals onto tmpctx */
```
niftynei: what do you mean by 'steals onto'?

rustyrussell: The line above says take(update), so we should not generally touch it (handle_channel_update owns it, is allowed to free it).  In fact, this is a "shouldnt happen" path already, and handle_channel_update actually just tal_steal(tmpctx) the update in the error path, rather than actually freeing it...

niftynei: Sweet. That makes sense. You could roughly translate it as 'steal onto tmpctx' -> `tal_steal(tmpctx)`

PR #1969, opened 2018-09-24
Invoiceapidoc patch

renepickhardt:
previously the API doc for the invoice command did not mention that one can optionally set a fallback address. If one wanted to set custome preimage there was however the error asking for a fallback address. also the python interfaces asks for the fallback address. 
Therefor I added the optional fallback address to the API doc


lightningd/invoice.c
```diff
@@ -267,7 +267,8 @@ static void json_invoice(struct command *cmd,
 static const struct json_command invoice_command = {
     "invoice", json_invoice, "Create an invoice for {msatoshi} with {label} "
 			     "and {description} with optional {expiry} seconds "
-			     "(default 1 hour) and optional {preimage} "
+			     "(default 1 hour) and optional {fallback addresses}"
```
rustyrussell: The name of the parameter is `fallbacks` (and it's in the lightning-invoice manual page, I checked!).  So perhaps
`optional {fallbacks} addresses (default empty list)`?

renepickhardt:
didn't know that the help was not meant to be complete.  I interpreted 
```
``` 
as in info about what values can be taken for the arguments

Anyway I changed the name and used `{fallbacks} address list` since I didn't like the double plural of `{fallbacks} addresses`

PR #1970, opened 2018-09-24
WIP: A first taste of optimization: stream lightningd JSON output

rustyrussell:
Christian did `lightning-cli listchannels` on my mainnet node (which runs under valgrind) and it pegged the CPU for minutes until I rebooted it :(

With only 11,000 channel pairs or so, it shouldn't take that long even under valgrind.  This patch series uses various techniques to speed it up from 1.4 seconds to 0.35 seconds (non-valgrind).  We could do more, but this shows how much work could be done; we've never really optimized before.

(Valgrind version runs in 11 seconds afterwards, I don't know how long before, as it's still running after 18 minutes...)


rustyrussell:
OK, this removes the overflow, in favor of ' jsonrpc: helper to move an existing io_write.'.  Which is simpler, if a layering violation...

Rest is unchanged, modulo rebase.

rustyrussell:
Trivial rebase onto master; added individual patch which fixed mutliple commands at once:

https://github.com/ElementsProject/lightning/pull/1970/commits/f23841ee97a80165c698ddd9765d8318529230bf

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/1970/commits/2fa96fd38f1b5e54876ac0f9fe14f2e4a563b140

PR #1974, opened 2018-09-25
git: ignore all current and future unit tests

wythe:
I wanted to add a couple recently created unit tests to the
.gitignore file, but instead I replaced all the names of tests with a simple
regex.

This works because git doesn't ignore files that are already in git. So
source and header file are safe.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,32 +21,11 @@ __pycache__
 config.vars
 
 # Ignore some generated binaries
-lightningd/test/run-channel
-lightningd/test/run-cryptomsg
-lightningd/test/run-commit_tx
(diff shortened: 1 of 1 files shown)


rustyrussell:
Yeah, but if you add a test, now it's ignored and you won't see you didn't commit it :(

I think you can append '!*/test/run-*.c' to fix this, however?

wythe:
Yes, seems to work. Updated.

niftynei:
ACK 187cf2f92756c68969ab814b401c1666b552f745

niftynei:
This really cleans up nice, thanks @wythe !

PR #1980, opened 2018-09-26
common: Added a small locking mechanism to simplify how we lock IO

cdecker:
We've done this a number of times already where we're getting
exclusive access to either the out direction of a connection, or we
try to lock out the read side while we are responding to a previous
request. They usually are really cumbersome because we reach around to
the other direction to stop it from proceeding, or we flag our
exclusive access somewhere, and we always need to know whom to notify.

PR ElementsProject/lightning#1970 adds two new instances of this:

 - Streaming a JSON response requires that nothing else should write
   while the stream is active.
 - We also want to stop reading new requests while we are responding
   to one.

To remove the complexity of having to know whom to stop and notify
when we're done, this adds a simple `io_lock` primitive that can be
used to get exclusive access to a connection. This inverts the
requirement for notifications, since everybody registers interest in
the lock and they get notified if the lock holder releases it.


common/io_lock.c
```diff
@@ -0,0 +1,74 @@
+#include "io_lock.h"
+#include <assert.h>
+#include <ccan/io/io_plan.h>
+
+struct io_lock {
+	bool free;
```
rustyrussell: nitpick mode: enabled!

I would `bool locked` this, as free has other connotations which are unwelcome here....

common/io_lock.c
```diff
@@ -0,0 +1,74 @@
...
+	lock->free = true;
+	return lock;
+}
+
+static struct io_plan *io_lock_try_acquire(struct io_conn *conn,
+					   struct io_lock_waiter *waiter)
+{
+	if (waiter->lock->free) {
+		waiter->lock->free = false;
+		return io_out_always(conn, waiter->next, waiter->arg);
```
rustyrussell: Can we just `return waiter->next(conn, waiter->arg)`?  Hmm, that's a memleak, so we need:

```
    struct io_plan *(*next)(struct io_conn *conn, void *next_arg) = waiter->next;
    void *next_arg = waiter->arg;

    tal_free(waiter);
    return next(conn, next_arg);
}
```

common/io_lock.c
```diff
@@ -0,0 +1,74 @@
...
+	} else {
+		return io_out_wait(conn, waiter->lock, io_lock_try_acquire, waiter);
```
rustyrussell: So, internally io knows whether we're dealing with in or out.  Maybe I should simplify the API to avoid this?  I haven't looked hard though, might be a good reason not to...

common/io_lock.c
```diff
@@ -0,0 +1,74 @@
...
+    struct io_conn *conn, struct io_lock *lock, enum io_direction dir,
+    struct io_plan *(*next)(struct io_conn *, void *), void *arg)
+{
+	if (lock->free) {
+		lock->free = false;
+		if (dir == IO_OUT)
+			return io_out_always(conn, next, arg);
+		else
+			return next(conn, arg);
+	}
```
rustyrussell: Is this fast-path a premature optimization?  It avoids the allocation, but I'd write it simple and put a FIXME suggestion in, so we look clever if profiling later reveals a bottleneck.

common/io_lock.c
cdecker: My mistake, I initially had it only for `IO_OUT`, I fixed it up now to check the direction and then calling the correct version. `io_wait` automatically locks `IO_IN` which I think is not correct if we are in `io_duplex` and want to lock the `IO_OUT` explicitly, or am I missing something?

common/io_lock.c
cdecker: Sounds good :+1:

cdecker:
> This is pretty nice! My comments are from a POV that this might eventually go upstream into ccan/io or a new ccan/io/lock, if it proves a generic enough primitive.

I was hoping you'd be suggesting that :-)

If we extend the lock to include a timer that can grab a lock (or wait for the lock to be released) it'd actually cover all trigger points that I can think of that could cause unwanted concurrency across multiple reads or writes.

common/io_lock.c
```diff
@@ -0,0 +1,75 @@
...
+struct io_plan *
+io_lock_acquire_in_(struct io_conn *conn, struct io_lock *lock,
+		    struct io_plan *(*next)(struct io_conn *, void *), void *arg)
+{
+	return io_lock_acquire_dir(conn, lock, IO_IN, next, arg);
+}
+
+void io_lock_release(struct io_lock *lock)
+{
+	assert(lock->locked);
```
niftynei: do we want ignorant clients to crash?

cdecker: Yes :+1:

I think we should enforce that people actually lock when they think they locked. This is after the fact but still gives a clear message that something the user expected (being locked) went wrong.

common/io_lock.c
```diff
@@ -0,0 +1,75 @@
...
+#include <assert.h>
+#include <ccan/io/io_plan.h>
+
+struct io_lock {
+	bool locked;
+};
+
+/* Struct to hold information while we wait for the lock to be freed */
+struct io_lock_waiter {
+	struct io_plan *(*next)(struct io_conn *conn, void *next_arg);
```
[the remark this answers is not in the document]
cdecker: Right, we haven't done this so far, but it might increase legibility. @rustyrussell would that be Rusty-Style C? :wink:

common/io_lock.c
```diff
@@ -0,0 +1,75 @@
...
+static struct io_plan *io_lock_try_acquire(struct io_conn *conn,
+					   struct io_lock_waiter *waiter)
+{
+	/* Destructure waiter, since we might be freeing it below */
+	struct io_plan *(*next)(struct io_conn *, void *) = waiter->next;
+	void *next_arg = waiter->arg;
+
+	if (!waiter->lock->locked) {
+		waiter->lock->locked = true;
+		tal_free(waiter);
```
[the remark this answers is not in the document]
jb55: I'm guessing because it's not necessarily freed in all branches?

cdecker:
Forgot to add the test I wrote for the locking. It serializes multiple concurrent multi-step writers.

common/io_lock.c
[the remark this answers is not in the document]
rustyrussell: No, I don't typedef structs, unless I really don't want you to treat them differently (eg. they're structs on some platforms, and ints in others, or int on some, long on others, etc.).

When I see 'struct foo' I know:
1. I can assign it, but not compare it naively.
2. I should pass it by pointer, rather than copy.
3. It can be predeclared with 'struct foo' to avoid pulling in the entire definition.

common/io_lock.c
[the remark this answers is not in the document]
rustyrussell: There are four kinds of *pointer) function parameters on this spectrum:
1. ones where ownership stays with the caller.  This is most common.
2. ones where the callee always takes ownership, eg `free(ptr)` obviously takes over ptr.
3. ones which sometimes take ownership, depending on internal conditions, like this one, or a reference-counting release function.  These are always dangerous (eg. realloc!), but sometimes necessary, esp. for internal functions handling internal objects.
4. ones which always take ownership if the caller tells them to, otherwise never take ownership.  This is what TAKES annotations mean; it takes the pointer IF, ONLY IF, and ALWAYS IF the caller hands it take(ptr).

common/io_lock.c
```diff
@@ -0,0 +1,75 @@
...
+		return next(conn, next_arg);
+	} else {
+		if (waiter->dir == IO_IN)
+			return io_wait(conn, waiter->lock, io_lock_try_acquire, waiter);
+		else
```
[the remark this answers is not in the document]
rustyrussell: Yep.

common/io_lock.c
rustyrussell: Yes.  It's fast (both in programmer time and in runtime performance), and it's really hard to add more robustness like this *after* something is in use, since you never know what would break.

cdecker:
Hm, I like the idea of using `switch` for completeness checking at compile time. However the `fixup` I just posted will fail because of this:

```
In file included from common/test/run-lock.c:1:0:
common/test/../io_lock.c: In function ‘io_lock_try_acquire’:
common/test/../io_lock.c:44:1: error: control reaches end of non-void function [-Werror=return-type]
 }
 ^
```

There are two ways of removing this error that I see (both defeat the purpose of the `switch` statement in the first place): use a `default` clause in the `switch` (gets rid of completeness check that we wanted), or have an additional "default return" after the `switch` (can't be reached and is there solely to silence the compiler). Is there a nice solution I'm missing @rustyrussell ?

cdecker:
Ok, fixed up, squashed and rebased. Ready for review :wink: 

Ping @rustyrussell and/or @niftynei

PR #1981, opened 2018-09-26
pytest: Fix benchmarks after the fixture migration

cdecker:
Back when we migrated to pure `py.test` based tests we broke some of the benchmarks.


niftynei:
ACK 34135c85fc3332b8320c99c9455bd1d479835040

PR #1982, opened 2018-09-27
Routeboost support

rustyrussell:
~~~First commit is the tal_arr_expand() helper which is also in several other PRs; I'll rebase on whatever gets merged first :)~~~
First 4 commits are basically #1983 

Note that these are our first warnings in our API.  My thoughts on these are:

1. Warnings will be in the top-level object, if at all, and the key will start with 'warning'.
2. The key for warnings will be unique, and are part of the JSON API (thus should not change).  This key is for machine digest (eg. do something if it appears).
3. The value of the warning is human-readable, so can be passed through to user, even if the caller doesn't understand the warning.


cdecker:
Rebased.

ACK https://github.com/ElementsProject/lightning/pull/1982/commits/701b1dc35713ecff8e25af6d0f07d53b3c4fdfc0

PR #1986, opened 2018-09-28
OMG even more gossip fixes

rustyrussell:
More reading, more fixing...

We really should test the `riskfactor` parameter, but at least this is better than "1"!


rustyrussell:
Rebased... eliminated the two commits which were already applied...

PR #1987, opened 2018-09-28
pytest: Stabilize test_feerates

cdecker:
Seems we simply were too quick to check for effects.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/1987/commits/ca60deec4f74be01426e7fc5f9b95b628e7fd7d9

PR #1992, opened 2018-10-04
BOLT 7 update: optional `htlc_max_capacity` field in `channel_update`

niftynei:
Update to now parse and obey the `htlc_maximum_msat` field that was added to BOLT 11 in the `channel_update`.  Nonsense values just mean you don't get routed through.

Not entirely sure if these will get persisted to the database correctly.

Changeset to allow a node to set this value for its `channel_update`s up next. :soon:


gossipd/routing.c
```diff
@@ -1064,20 +1064,30 @@ bool routing_add_channel_update(struct routing_state *rstate,
...
 				     &short_channel_id, &timestamp,
 				     &message_flags, &channel_flags,
 				     &expiry, &htlc_minimum_msat, &fee_base_msat,
-				     &fee_proportional_millionths,
-				     &htlc_maximum_msat))
+				     &fee_proportional_millionths))
+		return false;
+	/* If it's flagged as containing the optional field, reparse for
+	 * the optional field */
+	if (message_flags & ROUTING_OPT_HTLC_MAX_MSAT &&
```
niftynei: is there a nicer way to do this, other than re-parsing once we find out the flag is set for it?

rustyrussell:
I think we should reject any update claiming a capacity of > chan->satoshis, too. Clearly that's impossible, so we shouldn't forward them either.

Only minor quibbles with the rest; nice work!

gossipd/routing.c
```diff
@@ -1048,9 +1064,10 @@ bool routing_add_channel_update(struct routing_state *rstate,
 	u32 timestamp;
 	u8 message_flags, channel_flags;
 	u16 expiry;
-	u64 htlc_minimum_msat;
+	u64 htlc_minimum_msat = 0; // init as 0
```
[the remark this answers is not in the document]
niftynei: in the case where an incoming `channel_update` message doesn't have the optional field set, it won't get initialized before being passed to `set_connection_values`.

niftynei:
@rustyrussell this should be a fixedup version of the things you noticed. thanks for the pointer on using the `channel->satoshi` value as a default!

gossipd/routing.c
```diff
@@ -1051,15 +1058,17 @@ bool routing_add_channel_update(struct routing_state *rstate,
 	u64 htlc_minimum_msat;
 	u32 fee_base_msat;
 	u32 fee_proportional_millionths;
+	u64 htlc_maximum_msat;
 	struct bitcoin_blkid chain_hash;
 	struct chan *chan;
 	u8 direction;
 
-	if (!fromwire_channel_update(update, &signature, &chain_hash,
+	if (!fromwire_channel_update_option_channel_htlc_max(update, &signature, &chain_hash,
```
rustyrussell: How does this work?  Won't we fail to parse old-style channel_updates now?  (Hmm, maybe this is fixed in another commit, since I review one commit at a time...)

gossipd/routing.h
```diff
@@ -27,7 +27,7 @@ struct half_chan {
 	s64 last_timestamp;
 
 	/* Minimum number of msatoshi in an HTLC */
-	u32 htlc_minimum_msat;
+	u64 htlc_minimum_msat;
```
rustyrussell: Good catch!

gossipd/test/run-find_route-specific.c
```diff
@@ -126,8 +129,9 @@ get_or_make_connection(struct routing_state *rstate,
 	/* Make sure it's seen as initialized (update non-NULL). */
 	chan->half[idx].channel_update = (void *)chan;
 	chan->half[idx].htlc_minimum_msat = 0;
+	chan->half[idx].htlc_maximum_msat = satoshis * 1000;
 
-	return &chan->half[pubkey_idx(from_id, to_id)];
+	return &chan->half[idx];
```
cdecker: Good catch :+1:

niftynei:
Bisectability seems like something worth preserving. @rustyrussell what do you think about squashing them before merging this time?

PR #1994, opened 2018-10-05
Make it more obvious that you need to install cppcheck + fix warning

niftynei:
I finally figured out why `cppcheck` and `shellcheck` weren't working for me -- I didn't have them installed. :man_facepalming: 

This updates the notes in HACKING.md to include them as a part of the setup instructions for the test infrastructure.

Once they were running, `cppcheck` was failing because of a warning. This also adds a null pointer check to quiet down the warning so that the check will pass.


rustyrussell:
OK, but I really dislike the 'if (cmd != NULL &&' solution: cmd *cannot* be NULL here, and that kind of implies it can be, just to keep cppcheck quiet...  Hence I suggested assert().

cdecker:
Ok, would `cppcheck` stop complaining if we do the `assert(cmd)`?

niftynei:
`assert(cmd)` does indeed kill the warning. Updated changeset to use `assert()` instead of the `NULL`.

PR #1998, opened 2018-10-09
cli: help improvements

### LINKED ISSUE #1882: cli: help with a specific command should fallback to api if there's no manpage
Right now help fails on a single command if it can't find the manpage. We might still have verbose help from the api (see `help -J command`), so we should fall back to that?

jb55:
* Fallback to help rpc when manpage is not found
* Simplify json help response

Fixes #1882

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -89,22 +89,24 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
 	abort();
 }
 
-static void human_help(const char *buffer, const jsmntok_t *result) {
+static void human_help(const char *buffer, const jsmntok_t *result, bool has_command) {
 	int i;
 	const jsmntok_t * help_array = result + 2;
 	/* the first command object */
 	const jsmntok_t * curr = help_array + 1;
 	/* iterate through all commands, printing the name and description */
-	for (i = 0; i<help_array->size; i++) {
+	for (i = 0; i < help_array->size; i++) {
 		curr += 2;
 		printf("%.*s\n", curr->end - curr->start, buffer + curr->start);
 		curr += 2;
 		printf("    %.*s\n\n", curr->end - curr->start, buffer + curr->start);
+		curr += 2;
 		/* advance to next command */
 		curr++;
 	}
 
-	printf("---\nrun `lightning-cli help <command>` for more information on a specific command\n");
+	if (!has_command)
+		printf("---\nrun `lightning-cli help <command>` for more information on a specific command\n");
 }
 
 enum format {
@@ -363,9 +365,12 @@ int main(int argc, char *argv[])
 		     json_tok_len(id), json_tok_contents(resp, id));
 
 	if (!error || json_tok_is_null(resp, error)) {
+		// if we have specific help command
 		if (format == HUMAN)
-			if (streq(method, "help")) human_help(resp, result);
-			else human_readable(resp, result, '\n');
+			if (streq(method, "help") && command == NULL)
+				human_help(resp, result, false);
+			else
+				human_readable(resp, result, '\n');
 		else
 			printf("%.*s\n",
 			       json_tok_len(result),
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -14,6 +14,7 @@
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <sys/un.h>
+#include <sys/wait.h>
 #include <unistd.h>
 
 #define NO_ERROR 0
@@ -171,9 +172,23 @@ static void add_input(char **cmd, const char *input,
 }
 
 static void
-exec_man (const char *page) {
-	execlp("man", "man", page, (char *)NULL);
-	err(1, "Running man command");
+try_exec_man (const char *page) {
+	int status;
+
+	switch (fork()) {
+	case -1:
+		err(1, "Forking man command");
+	case 0:
+		/* child, run man command. */
+		execlp("man", "man", page, (char *)NULL);
+		err(1, "Running man command");
+	default:
+		break;
+	}
+
+	wait(&status);
+	if (WIFEXITED(status) && WEXITSTATUS(status) == 0)
+		exit(0);
 }
 
 int main(int argc, char *argv[])
@@ -192,6 +207,7 @@ int main(int argc, char *argv[])
 	int parserr;
 	enum format format = DEFAULT_FORMAT;
 	enum input input = DEFAULT_INPUT;
+	char *command = NULL;
 
 	err_set_progname(argv[0]);
 	jsmn_init(&parser);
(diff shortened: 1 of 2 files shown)


cli/lightning-cli.c
```diff
@@ -170,10 +170,11 @@ static void add_input(char **cmd, const char *input,
...
 
-static void
-exec_man (const char *page) {
-	execlp("man", "man", page, (char *)NULL);
-	err(1, "Running man command");
+static int
+spawn_man (const char *page) {
+	char command[sizeof("man") + strlen(page)];
+	snprintf(command, sizeof(command), "man %s", page);
+	return system(command);
```
rustyrussell: `lightning-cli help '; echo Hello world'`

It's nasty, but I think we need do this this properly (*UNTESTED CODE AHEAD!*):

```
/* If man command succeeds, exit silently.  Otherwise we return. */
static void spawn_man(const char *page) {
int status;

switch (fork()) {
case -1:
    err(1, "Forking man command");
case 0:
    /* child, run man command. */
    execlp("man", "man", page, (char *)NULL);
    err(1, "Running man command");
default:
    break;
}

wait(&status);
if (WIFEXITED(status) && WEXITSTATUS(status) == 0)
    exit(0);
}
```

cli/lightning-cli.c
```diff
@@ -78,8 +78,10 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
 			return human_readable(buffer, t + 2, '\n') + 3;
 		n = 1;
 		for (i = 0; i < t->size; i++) {
-			n += human_readable(buffer, t + n, '=');
+			n += human_readable(buffer, t + n, '\t');
```
rustyrussell: A nice change, but it was done this way so 'lightning-cli -H listconfigs' gave a valid config.

While you're here giving the cli some much-needed love, can you make listconfigs output usable by default?  Thanks!

cli/lightning-cli.c
jb55: good call

cli/lightning-cli.c
jb55: I didn't realize it was used for this. I think I'll just revert this for now instead of trying to factor out two different human/config renderers.

jb55:
pushed v2

v1 -> v2

- backed out config rendering changes for now
- use fork+exec instead of system
- bracket style fix in jsonrpc

```
$ git range-diff b7b0a1fb~3..b7b0a1fb b708ff4e~3..b708ff4e
```


```diff
7cb864fa ! 1:  04158a34 rpc: add json_add_help_command for help command objects
@@ -34,8 +34,7 @@
 +				"HELP! Please contribute"
 +				" a description for this"
 +				" json_command!");
-+	}
-+	else {
++	} else {
 +		struct json_escaped *esc;
 +
 +		esc = json_escape(NULL, json_command->verbose);
e4b27047 ! 2:  b8511082 cli: handle missing manpages gracefully
@@ -11,18 +11,37 @@
 --- a/cli/lightning-cli.c
 +++ b/cli/lightning-cli.c
 @@
- 		tal_append_fmt(cmd, ", ");
+ #include <sys/socket.h>
+ #include <sys/types.h>
+ #include <sys/un.h>
++#include <sys/wait.h>
+ #include <unistd.h>
+ 
+ #define NO_ERROR 0
+@@
  }
  
--static void
+ static void
 -exec_man (const char *page) {
 -	execlp("man", "man", page, (char *)NULL);
 -	err(1, "Running man command");
-+static int
-+spawn_man (const char *page) {
-+	char command[sizeof("man") + strlen(page)];
-+	snprintf(command, sizeof(command), "man %s", page);
-+	return system(command);
++try_exec_man (const char *page) {
++	int status;
++
++	switch (fork()) {
++	case -1:
++		err(1, "Forking man command");
++	case 0:
++		/* child, run man command. */
++		execlp("man", "man", page, (char *)NULL);
++		err(1, "Running man command");
++	default:
++		break;
++	}
++
++	wait(&status);
++	if (WIFEXITED(status) && WEXITSTATUS(status) == 0)
++		exit(0);
  }
  
  int main(int argc, char *argv[])
@@ -42,10 +61,9 @@
 -		snprintf(command, sizeof(command), "lightning-%s", argv[2]);
 -		exec_man(command);
 +		command = argv[2];
-+		char mancommand[strlen(command) + sizeof("lightning-")];
-+		snprintf(mancommand, sizeof(mancommand), "lightning-%s", command);
-+		if (spawn_man(mancommand) == 0)
-+			exit(0);
++		char page[strlen(command) + sizeof("lightning-")];
++		snprintf(page, sizeof(page), "lightning-%s", command);
++		try_exec_man(page);
  	}
  
  	if (chdir(lightning_dir) != 0)
b7b0a1fb ! 3:  b708ff4e cli: return more human friendly command help
@@ -23,18 +23,6 @@
 diff --git a/cli/lightning-cli.c b/cli/lightning-cli.c
 --- a/cli/lightning-cli.c
 +++ b/cli/lightning-cli.c
-@@
- 			return human_readable(buffer, t + 2, '\n') + 3;
- 		n = 1;
- 		for (i = 0; i < t->size; i++) {
--			n += human_readable(buffer, t + n, '=');
-+			n += human_readable(buffer, t + n, '\t');
-+			printf("\n\t");
- 			n += human_readable(buffer, t + n, '\n');
-+			printf("\n");
- 		}
- 		return n;
- 	case JSMN_UNDEFINED:
 @@
  	abort();
  }
@@ -68,12 +56,11 @@
  
  	if (!error || json_tok_is_null(resp, error)) {
 +		// if we have specific help command
-+		bool has_command = command != NULL;
  		if (format == HUMAN)
 -			if (streq(method, "help")) human_help(resp, result);
 -			else human_readable(resp, result, '\n');
-+			if (streq(method, "help") && !has_command)
-+				human_help(resp, result, has_command);
++			if (streq(method, "help") && command == NULL)
++				human_help(resp, result, false);
 +			else
 +				human_readable(resp, result, '\n');
  		else
```

cdecker:
CI failure seems to be unrelated. `git range-diff` seems be a really useful find, thanks for pointing that out @jb55, I'll try to add that to ackbot.

cli/lightning-cli.c
```diff
@@ -89,22 +89,24 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
 	abort();
 }
 
-static void human_help(const char *buffer, const jsmntok_t *result) {
+static void human_help(const char *buffer, const jsmntok_t *result, int has_command) {
```
rustyrussell: Modern style would make this 'bool has_command'?

rustyrussell:
You're quick!  I added a changelog entry, too.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Bitcoind: more parallelism in requests, for very slow nodes.
 - Testing: fixed logging, cleaner interception of bitcoind, minor fixes.
 - JSON API: `invoice` now adds route hint to invoices for incoming capacity (RouteBoost), and warns if insufficient capacity.
+- CLI: `help` prettified, and falls back to internal help if manual page missing.
```
jb55: oh it's not prettified yet, it's just spitting out that config-style output for now. I should update that commit description...

jb55: I'm going to prettify the fallback help in a future PR since I need to refactor a few things first.

jb55:
William Casarin <jb55@jb55.com> writes:
> I noticed man is no longer paginated after the fork. Is there any
> workaround for this? I'm not super familiar with fork+exec to know why
> this would happen or how to fix it. Kind of a minor issue but it's a bit
> annoying.

Hmm, it Works For Me!  Both in gnome-terminal, on console and via ssh.
These are Ubuntu machines though.

By default, man runs 'pager' which (on most systems) is actually 'less'.
Either man isn't running the pager, or the pager thinks it shouldn't
paginate.

Any chance you have $PAGER $MANPAGER or $MANOPT set?  You can try
setting them to see what's going on...

Hmm,
Rusty.

PR #1999, opened 2018-10-09
make: fix broken error message in bolt-check

jb55:
It looks like a printf was missing for the bolt-check error message

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -243,7 +243,7 @@ check-makefile:
 
 # Any mention of BOLT# must be followed by an exact quote, modulo whitespace.
 bolt-check/%: % bolt-precheck tools/check-bolt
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -243,7 +243,7 @@ check-makefile:
 
 # Any mention of BOLT# must be followed by an exact quote, modulo whitespace.
 bolt-check/%: % bolt-precheck tools/check-bolt
-	@if [ -d .tmp.lightningrfc ]; then tools/check-bolt .tmp.lightningrfc $<; else Not checking BOLTs: BOLTDIR $(BOLTDIR) does not exist >&2; fi
+	@if [ -d .tmp.lightningrfc ]; then tools/check-bolt .tmp.lightningrfc $<; else printf "Not checking BOLTs: BOLTDIR $(BOLTDIR) does not exist\n" >&2; fi
```
rustyrussell: Though "echo" is more normal here!

jb55: switched to echo

PR #2002, opened 2018-10-09
options: Parse config file before opening working directory

trueserve:
# Bug:
lightningd does not respect `lightning-dir` configuration option; always uses `$HOME/.lighting` instead

# Steps to reproduce:
Set `lightning-dir` in user's `config` file
`lightningd/lightningd --conf=config`

# Expected results:
Working directory is path set by `lightning-dir` in user's `config` file

# Actual results:
`$HOME/.lightning` is always used

# Description:
Right now, the `config` file is read *after* the configuration working directory is moved to in the software. However one configuration option `lightning-dir` settable in the `config` file sets this working directory. As the directory is already opened (which defaults to `$HOME/.lightning`) before the configuration is read, the configured directory will not be used.

This patch parses the configuration file before opening the working directory, fixing this bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Protocol: `channel_update` sent to disable channel only if we reject an HTLC.
 - Protocol: we don't send redundant `node_announcement` on every new channel.
+- Config: config file can override `lightning-dir` (makes sense with `--conf`).
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Protocol: `channel_update` sent to disable channel only if we reject an HTLC.
 - Protocol: we don't send redundant `node_announcement` on every new channel.
 - Config: config file can override `lightning-dir` (makes sense with `--conf`).
+- Config: `--conf` option is now relative to current directory, not `lightning-dir`.
 
 ### Deprecated
 
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -2,12 +2,12 @@
 .\"     Title: lightningd-config
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 09/07/2018
+.\"      Date: 10/11/2018
 .\"    Manual: \ \&
 .\"    Source: \ \&
 .\"  Language: English
 .\"
-.TH "LIGHTNINGD\-CONFIG" "5" "09/07/2018" "\ \&" "\ \&"
+.TH "LIGHTNINGD\-CONFIG" "5" "10/11/2018" "\ \&" "\ \&"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


rustyrussell:
Hmm, until we added `--conf`, this was expected behavior.  I agree it needs to be fixed, though.  We currently don't use any paths specified in the config file until after parsing, so this is safe (they're *defined* to be lightning-dir-relative).

I'll append a CHANGELOG.md entry...

rustyrussell:
OK, this actually doesn't work any more, since opt_parse_from_config() tries to just read 'config' (assuming it's in the current directory!).  I'll append a fix for your review...

trueserve:
The current patch requires an absolute path to the config file, which doesn't make sense. I can't think of other tools that work like this, they generally don't care if a config file is relative or absolute. I can understand `lightningd` wanting to work from a working directory but the config is read only at startup, it doesn't need to live there. (though looking for it by default in `lightning-dir` might make sense, see below.)

Let me try to clarify things, let me know if something is wrong...

- Previously, if not specifying a config file, the file named `config` in the current path would be used. With the latest patch, this file is looked for in the `lightning-dir`. I think this makes sense for the software to look in some defined path for configuration if not otherwise specified, rather than cwd.
- With the latest patch, specifying something like `--conf=lightning.conf` will look in the `lightning-dir` (so something like `$HOME/.lightning/lightning.conf`) and if not found, return an opaque error. But shouldn't `--conf` look relative to where `lightningd` was invoked, like any other tool? Because otherwise, this works unlike any other tool, and makes `lightning-dir` in the config meaningless. Put another way, is there any reason to force use of the `lightning-dir` in this case, especially since the config can contain `lightning-dir`, rather than common practice where this will be read relative to the current working path?
- An idea: if specifying a configuration file via --conf, should the file:
  - Only be looked at from the cwd? (most consistent / sane)
  - Only be looked at from the `lightning-dir` unless absolute? (not consistent or standard)
  - The cwd is checked first, and if the file is not found, then `lightning-dir` is checked before giving up? (is there any use case for multiple configs in the same working dir?)

I have a patch ready for the last case but I don't really see much use of specifying different configs from within `lightning-dir` by short hand; if someone wants to use them, they can specify them fully or launch `lightningd` from the working directory. I see this method only causing confusion. So the attached patch is for the first case, which is similar to how things were originally. Let me know if you would prefer the third case, or if there is any argument for the second.

The documentation is mostly okay but it also adds some confusion regarding `lightning-dir` and the config file - especially as this can be specified within the config file itself. This extraneous information was removed and hopefully things are clear.

rustyrussell:
Right, but the current documented behavior of --conf is that it's lightning-dir relative.  That means you can copy config to testnet.config and tweak some things then specify --conf=testnet.config and that Just Works.

rustyrussell:
I also added the path we were trying to open to the error msg, thanks!

trueserve:
> No, the default config file was 'config' in lightning-dir. That seems like the Right Thing?

Before these patches, it was not. I was using `config` in the shell's current working directory, not in `lightning-dir`, and I was actually not specifying `--conf`. After your patch, it explicitly is looking `config` in `lightning-dir`.

But yes, it does make sense for the **default config and path** to be in `lightning-dir`.

Your latest patch matches my latest changes almost exactly (even the docs), I think this way is the most clear for anyone using this feature. Thanks.

rustyrussell:
> Before these patches, it was not. I was using config in the shell's current working directory, not in lightning-dir, and I was actually not specifying --conf. After your patch, it explicitly is looking config in lightning-dir.

I'm confused: it's never been that way.  My nodes run with config in the .lightning directory, and none on the directory they run in, and they read it.  I restarted with strace (they're on git master as of a few days ago), to be sure (and checked there was no `config` in the current directory):

```
8950  chdir("/home/rusty/.lightning")   = 0
8950  openat(AT_FDCWD, "config", O_RDONLY) = 3
```

Did you have some customizations?

trueserve:
Yes, I run a fork for a different coin, but I did not see any changes to configuration handling in that fork. Ultimately it doesn't matter, this patch makes everything clear. 👍

PR #2005, opened 2018-10-10
lightningd: fail HTLCs which are in-transit as we shut down.

rustyrussell:
This is the source of failure in the test_restart_many_payments stress
test: we don't commit the outgoing HTLC immediately, instead waiting for
gossip to tell us the peer for the outgoing channel, then waiting for
that channeld to tell is it's committed.  The result was incoming HTLCs
with no outgoing.

I initially pushed the HTLCs through that same path, but of course
(since peers are not connected yet!) the only result was that we failed
these HTLCs immediately.  So I chose the far simpler course of just
failing them directly.

To reproduce this, I had to increase the test_restart_many_payments
num to 10, and run it with nice -20 taskset -c 0.

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -1742,6 +1742,20 @@ void htlcs_reconnect(struct lightningd *ld,
 	struct htlc_out_map_iter outi;
 	struct htlc_in *hin;
 	struct htlc_out *hout;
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2005/commits/6824ada7c9632a95cd7e63d65e73eb95d754bd6a

PR #2006, opened 2018-10-10
python: add cheroot to requirements.txt

jb55:
It looks like this is a new dependency.

Sort the list while we're at it.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2006/commits/97f92bc1c8e721b7420112ff10d09d36bdb611ac

PR #2007, opened 2018-10-10
Clarify when htlc_maximum_msat is used in routing_add_channel_update(...)

practicalswift:
Clarify when htlc_maximum_msat is used in routing_add_channel_update(...).

~~Avoid theoretical read of uninitialized variable `htlc_maximum_msat` in `routing_add_channel_update(...)` if `message_flags` changes between checks.~~

~~More specifically if `message_flags` changes between line 1074 and 1087 due to the call to `fromwire_channel_update_option_channel_htlc_max`.~~

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -1071,7 +1071,8 @@ bool routing_add_channel_update(struct routing_state *rstate,
 		return false;
 	/* If it's flagged as containing the optional field, reparse for
 	 * the optional field */
-	if ((message_flags & ROUTING_OPT_HTLC_MAX_MSAT) &&
+	bool contains_htlc_max_msat = message_flags & ROUTING_OPT_HTLC_MAX_MSAT;
+	if (contains_htlc_max_msat &&
 			!fromwire_channel_update_option_channel_htlc_max(
 				update, &signature, &chain_hash,
 				&short_channel_id, &timestamp,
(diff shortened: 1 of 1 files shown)


practicalswift:
@cdecker Yes, after re-reading the code I think you're right. The fact that `message_flags` is guaranteed to be unchanged after the call to `fromwire_channel_update_option_channel_htlc_max` was not entirely clear for me the first time I read the code, so I think it might be good to introduce the boolean as a clarification. I'll update the PR title, description and commit message :-)

practicalswift:
@cdecker Updated – please re-review :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2007/commits/d20e1082a144470a98cf9b83e8582273200ac7f5

rustyrussell:
This is really up to @niftynei who wrote that code, I think...

niftynei:
If you read the comment, it's pretty clear what we're doing with the flag check in the first instance.

```
	/* If it's flagged as containing the optional field, reparse for
	 * the optional field */
```

imo this change, switching to a named variable, makes it a bit harder to tell how/where we're deciding to check the `htlc_maximum_msat` field at line 1088. In the existing version, it's much easier to map back the underlying if check to a field in the channel_update message. 

Additionally, grepping for usages of the flag is much easier in the existing version, as the flag + usage are in the same place.

PR #2013, opened 2018-10-12
gossipd: update node announcement even if we change within a second.

rustyrussell:
Usually Travis triggers corner cases because it's so slow, but this
time the moons aligned, and it managed to fail test_node_reannounce
because it generated the updated node_announcement with the same
timestamp as the old one.

This is because we only updated "last_announce_timestamp" when
we generated the announcement, not when we got it off the wire or
loaded it from the gossip store.

The fix is to ask the routing code what the latest timestamp is;
we could still generate a clashing timestamp if (1) the gossip store
is lost, and (2) we restart within one second.  Hard to care.


gossipd/gossipd.c
```diff
@@ -310,14 +307,20 @@ static u8 *create_node_announcement(const tal_t *ctx, struct daemon *daemon,
 
 static void send_node_announcement(struct daemon *daemon)
 {
-	u32 timestamp = time_now().ts.tv_sec;
+	u32 timestamp = time_now().ts.tv_sec & ~0x3;
```
jb55: perhaps a comment for this line

cdecker: Had the same issue, this is just rounding down to `// 3` isn't it? Is this to make room for the +1 and still be lower than now? We don't prohibit future timestamps, so why is this necessary?

rustyrussell: Oops, this was just to trigger the problem more easily.  Was **NOT** supposed to be in the patch!  Will remove with !fixup

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2013/commits/f473fdf5a617dc9c1bd579901dec6737a6d4ec3c

rustyrussell:
Trivial rebase, waiting for tests to pass...

PR #2018, opened 2018-10-12
Track fees earned by forwarding

### LINKED ISSUE #1786: Add forwarding tracking to channel state
We currently don't keep any stats on the number of payments we forwarded, or how many fees we accumulated over time.

## `channel_metrics` table
I propose we add a new table `channel_metrics` that keeps simple counters, gauges and other numeric data that is associated with a channel. The goal is to keep the `channels` table simple while still allowing us to extend the metrics in the future. Since we have been doing this for other tables we can just use an `enum` in the source to differentiate types of metrics, and keep the footprint small:

```
channel_id: foreign_key u64 | metric_type: u32 | value: u64
```

## Updating forwarding stats
The update to these values is rather simple, since we just bump the respective number and save it alongside the channel.

cdecker:
This adds simple tracking of fees when forwarding payments. The fee is counted both on the upstream as well as the downstream channel, so to get the total fee earned we need to sum up `in_msatoshi_fee` OR `out_satoshi_fee` not both, since that'd double the amount, but that's a minor quirk. If desired I can add the total fee earned into some other RPC call (`getinfo`?) that computes it correctly.

The first three commits are from #2017.

Fixes #1786

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ changes.
 - JSON RPC: `ping` now works even after one peer fails to respond.
 - JSON RPC: `getroute` `fuzzpercent` and `pay` `maxfeepercent` can now be > 100.
 - JSON RPC: `riskfactor` in `pay` and `getroute` no longer always treated as 1.
+- JSON-RPC: `listpeers` was always reporting 0 for all stats.
 - Protocol: fix occasional deadlock when both peers flood with gossip.
 - Protocol: fix occasional long delay on sending `reply_short_channel_ids_end`.
 - Protocol: re-send `node_announcement` when address/alias/color etc change.
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -647,6 +647,56 @@ static void json_add_htlcs(struct lightningd *ld,
 	json_array_end(response);
 }
 
+static void json_channel_add_stats(struct lightningd *ld,
+				struct json_result *response, u64 channel_id)
+{
+	struct channel_stats channel_stats;
+	/* Provide channel statistics */
+	wallet_channel_stats_load(ld->wallet, channel_id, &channel_stats);
+
+	if (deprecated_apis) {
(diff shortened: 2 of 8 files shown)


niftynei:
ACK 301630d

rustyrussell:
There are no in and out fees. It doesn't even make sense; there are "through" fees. 

This should be a single db field, and yes, expose it in getinfo.

cdecker:
I would have like to have a single fee field as well, however it is hard to associate the fee with a single channel. The fees are by definition the difference in value of an incoming HTLC and an outgoing HTLC, which are associated with different channels. We could make the fee counter a field of the node itself (the lowest common denominator of those two channels), but that's not really useful either if you want to learn which is your best earning / network supporting channel and which ones are just stale.

One alternative to this would be to move all of these stats to a counterset that is assigned to forwardings. That way we can drop the `*_payments_*` counters since we have more granular information, and the fees can be tracked in a single field, not sure if it's really worth it though. The table would look something like this:

```
channel_in_id | channel_out_id | htlc_in_id | htlc_out_id | in_msat | out_msat
```

That'd allow us to keep the dangling countersets even after HTLCs and channels are removed (avoiding awkward jumps in the statistics).

cdecker:
If we add in a timestamp we can create pretty graphs as well xD

rustyrussell:
OK, so just to confirm the plan:

1. A new db table for fees we collect.
2. A new `listfees` (listfeescollected? not sure) command to show them?
3. A new field in `getinfo` to show the total?

Do we do this for 0.6.2, or defer?  I really like the idea of having it...

cdecker:
Will work on this as fast as possible. The idea is to have the totals added to the `getinfo` output, not create a new `listfees`, but have a `listforwards` command that details the incoming and outgoing HTLCs, as well as their current status and amounts (including fees). That ought to give a better insight into what is happening on your node as far as forwardings go.

cdecker:
Added Work in Progress so this doesn't get merged accidentally

Issue #546, opened 2018-01-10
HTLC expiry etc. must be checked against config.


rustyrussell:
`channeld/full_channel.c:484:	/* FIXME: check expiry etc. against config. */`

We need to check that the HTLC we or they are offering is acceptable according to the config of the other side (eg. amount too small).

jBarz:
I wanted to deep dive into how HTLCs work so I was wondering if I could tackle this if no one else is :-)

ZmnSCPxj:
Go for it.

rustyrussell:
We actually seem to do all this now.  Closing.

PR #2019, opened 2018-10-13
Send max msat field in `channel_update`

niftynei:
Have c-lightning nodes advertise their `htlc_maximum_msat` value, to help other nodes make intelligent routing decisions. As written, this potentially leaks the joint channel reserve amount. Is this ok?

Also, the function definition for `htlc_maximum_msat()` is in a sort of weird place. It doesn't really pertain to the `initial_channel` flow, but it *is* right next to the function for `htlc_minimum_msat()`...

Todo: still needs tests. There didn't seem to be much ready infrastructure for manipulating the `max_inflight_htlc` (ie it's not currently configurable)

### DIFF
--- a/common/initial_channel.h
+++ b/common/initial_channel.h
@@ -109,7 +109,7 @@ static inline u64 channel_reserve_msat(const struct channel *channel,
  *
  * `htlc_minimum_msat` indicates the smallest value HTLC this node will accept.
  */
-static inline u32 htlc_minimum_msat(const struct channel *channel,
+static inline u64 htlc_minimum_msat(const struct channel *channel,
 				    enum side recipient)
 {
 	return channel->config[recipient]->htlc_minimum_msat;
--- a/gossipd/gossipd.c
+++ b/gossipd/gossipd.c
@@ -942,16 +942,13 @@ static void update_local_channel(struct daemon *daemon,
 				 u64 htlc_minimum_msat,
 				 u32 fee_base_msat,
 				 u32 fee_proportional_millionths,
+				 u64 htlc_maximum_msat,
 				 const char *caller)
 {
 	secp256k1_ecdsa_signature dummy_sig;
 	u8 *update, *msg;
 	u32 timestamp = time_now().ts.tv_sec;
 	u8 message_flags, channel_flags;
-
-	/* `message_flags` are optional.
-	 * Currently, not set by c-lightning */
-	message_flags = 0;
 
 	/* So valgrind doesn't complain */
 	memset(&dummy_sig, 0, sizeof(dummy_sig));
@@ -965,7 +962,10 @@ static void update_local_channel(struct daemon *daemon,
 	if (disable)
 		channel_flags |= ROUTING_FLAGS_DISABLED;
 
-	update = towire_channel_update(tmpctx, &dummy_sig,
+	// We set the htlc_maximum_msat value
+	message_flags = 0 | ROUTING_OPT_HTLC_MAX_MSAT;
+
+	update = towire_channel_update_option_channel_htlc_max(tmpctx, &dummy_sig,
 				       &daemon->rstate->chain_hash,
 				       &chan->scid,
 				       timestamp,
@@ -973,7 +973,8 @@ static void update_local_channel(struct daemon *daemon,
 				       cltv_expiry_delta,
 				       htlc_minimum_msat,
 				       fee_base_msat,
-				       fee_proportional_millionths);
+				       fee_proportional_millionths,
+				       htlc_maximum_msat);
 
 	if (!wire_sync_write(HSM_FD,
 			     towire_hsm_cupdate_sig_req(tmpctx, update))) {
@@ -1026,6 +1027,7 @@ static void maybe_update_local_channel(struct daemon *daemon,
 			     hc->htlc_minimum_msat,
 			     hc->base_fee,
 			     hc->proportional_fee,
+			     hc->htlc_maximum_msat,
 			     __func__);
 }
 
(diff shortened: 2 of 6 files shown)


common/initial_channel.h
```diff
@@ -109,11 +111,32 @@ static inline u64 channel_reserve_msat(const struct channel *channel,
...
 {
 	return channel->config[recipient]->htlc_minimum_msat;
 }
+
+/*
+ * The maximum msat that this node will accept for an htlc.
+ * It's flagged as an optional field in `channel_update`.
+ *
+ * We advertise the maximum value possible,
+ * defined as the smaller of either the remote's indicated
```
niftynei: this comment is out of date. fixing.

common/initial_channel.h
```diff
@@ -114,6 +116,27 @@ static inline u32 htlc_minimum_msat(const struct channel *channel,
...
+ * or the total channel capacity minus the cumulative reserve.
+ * FIXME: does this need fuzz?
+ * FIXME: where to put this?
+ */
+static inline u64 htlc_maximum_msat(const u64 funding_msat,
+		const struct channel_config *our_config,
+		const struct channel_config *remote_config)
+{
+	u64 min_htlc_msat = min(our_config->max_htlc_value_in_flight_msat,
+			remote_config->max_htlc_value_in_flight_msat);
```
rustyrussell: Interestingly, our_config->max_htlc_value_in_flight_msat doesn't matter; that effects only incoming HTLCs, not outgoing ones (which is what this channel_update is about).

channeld/full_channel.c
```diff
@@ -351,6 +351,11 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	if (htlc->msatoshi < htlc_minimum_msat(channel, recipient)) {
 		return CHANNEL_ERR_HTLC_BELOW_MINIMUM;
 	}
+	if (htlc->msatoshi > htlc_maximum_msat(channel->funding_msat,
```
rustyrussell: This is definitely a good thought; in fact, there's a FIXME and an issue about this (#546).  We don't check that we're not trying to send more than their max_htlc_value_in_flight_msat, or exceeding their max_accepted_htlcs, which we absolutely should do.  Those checks would be a great PR!

But that's orthogonal to this; we already check there's no reserve violation in this code, we should NOT check for additional things.  For example, we *can* send more than max-2*reserve for our very first HTLC.

common/initial_channel.h
rustyrussell: I think this belongs in channeld.c, perhaps called 'advertized_htlc_max()'?

gossipd/gossipd.c
```diff
@@ -965,15 +962,20 @@ static void update_local_channel(struct daemon *daemon,
 	if (disable)
 		channel_flags |= ROUTING_FLAGS_DISABLED;
 
-	update = towire_channel_update(tmpctx, &dummy_sig,
+	message_flags = 0;
+	if (htlc_maximum_msat > 0)
+		message_flags |= ROUTING_OPT_HTLC_MAX_MSAT;
```
rustyrussell: I don't *think* this can be zero?  Even if we end up refreshing old pre-this-patch channels (from the store), we'll just set htlc_maximum_msat to the channel capacity (which is what hc->htlc_maximum_msat defaults to.

rustyrussell: Moreover, if we don't set the flag, we probably shouldn't append a value.  We might assume that any other future appendings would have their own message_flag, but in theory they wouldn't have to...

niftynei: This is true, as written this will never be zero. I will update to remove it. I mostly added the if switch because it serves as an 'explainer' for why we're setting the message flag to that value (ie concretely links the htlc_maximum_msat to the ROUTING_OPT_HTLC_MAX_MSAT flag), but i guess that's what commit histories and inline comments are for!

tests/test_pay.py
```diff
@@ -150,7 +150,7 @@ def test_pay_get_error_with_update(node_factory):
     # channel_update, and it should patch it to include a type prefix. The
     # prefix 0x0102 should be in the channel_update, but not in the
     # onionreply (negation of 0x0102 in the RE)
-    l1.daemon.wait_for_log(r'Extracted channel_update 0102.*from onionreply 10070080(?!.*0102)')
+    l1.daemon.wait_for_log(r'Extracted channel_update 0102.*from onionreply 10070088[0-9a-fA-F]{88}')
```
niftynei: By setting the `message_flags` to 1 for the optional htlc_maximum_msat, the flags portion of the `channel_update` causes this to fail. Additionally, the message is now 88 bytes long.  Instead, we just check that the onionreply has the correct prefix and actual length.

rustyrussell: For future reference: I think this note belongs in the commit message; I consider github transitory.  Plus, it's a fun story, and such tales are the intermittent reward for reading the git log :)

gossipd/gossipd.c
```diff
@@ -1093,15 +1095,17 @@ static void handle_get_update(struct peer *peer, const u8 *msg)
 /* Return true if the information has changed. */
 static bool halfchan_new_info(const struct half_chan *hc,
 			      u16 cltv_delta, u64 htlc_minimum_msat,
-			      u32 fee_base_msat, u32 fee_proportional_millionths)
+			      u32 fee_base_msat, u32 fee_proportional_millionths,
+			      u64 htlc_maximum_msat)
```
rustyrussell: Ooh, I would have missed this.  Nice catch!

tests/test_pay.py
niftynei: :+1:

PR #2021, opened 2018-10-14
docker: Update docker images and docs to non-CVE-2018-17144 bitcoind

### LINKED ISSUE #2020: Update to at least bitcoind 0.16.3 in README and Dockerfiles
Bitcoin Core 0.16.0 is vulnerable to CVE-2018-17144. It should be replaced with version 0.16.3 or 0.17.0 in

* https://github.com/ElementsProject/lightning/blob/master/README.md#installation
* https://github.com/ElementsProject/lightning/blob/master/contrib/Dockerfile.builder
* https://github.com/ElementsProject/lightning/blob/master/contrib/Dockerfile.builder.fedora
* https://github.com/ElementsProject/lightning/blob/master/contrib/Dockerfile.builder.i386
* https://github.com/ElementsProject/lightning/blob/master/Dockerfile

cdecker:
While not strictly necessary it's certainly a good idea to test
against the latest one and not encourage users to use old versions.

Doesn't seem big enough of a change to note in the changelog.

Fixes #2020

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN apk add --no-cache \
 
 WORKDIR /opt
 
-ARG BITCOIN_VERSION=0.16.0
+ARG BITCOIN_VERSION=0.17.0
 ENV BITCOIN_TARBALL bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz
 ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/$BITCOIN_TARBALL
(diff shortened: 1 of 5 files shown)


cdecker:
Currently testing the CI docker images before pushing them to the hub.

NicolasDorier:
It is actually necessary, the link to 0.16.0 are now broken, so you can't build the image anymore.

NicolasDorier:
Be careful by upgrade to 0.17.0 though, the configuration file have breaking changes for testnet and regtest.

cdecker:
The tests pass correctly so we don't seem to be affected by any config changes, but thanks @NicolasDorier for pointing that out :+1:

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/9d763bf14bd379ef41fe8882eae3d025885f11d6

PR #2024, opened 2018-10-15
gossipd: minor fixes.

rustyrussell:
More side-effects of writing documentation and re-reading the code:

1. Don't leak private channels in routeboost.
2. Don't accept non-local short-channel-ids when forwarding.
3. Clarify the gossip timer expiry setting.

### DIFF
--- a/gossipd/gossipd.c
+++ b/gossipd/gossipd.c
@@ -1846,32 +1846,32 @@ static struct io_plan *gossip_init(struct daemon_conn *master,
 	return daemon_conn_read_next(master->conn, master);
 }
 
-static struct io_plan *resolve_channel_req(struct io_conn *conn,
-					   struct daemon *daemon, const u8 *msg)
+static struct io_plan *get_channel_peer(struct io_conn *conn,
+					struct daemon *daemon, const u8 *msg)
 {
 	struct short_channel_id scid;
 	struct chan *chan;
-	struct pubkey *keys;
+	const struct pubkey *key;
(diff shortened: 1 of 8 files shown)


gossipd/gossipd.c
```diff
@@ -1538,6 +1538,10 @@ static struct io_plan *get_incoming_channels(struct io_conn *conn,
 			const struct half_chan *hc;
 			struct route_info *ri;
 
+			/* Don't leak private channels. */
+			if (!is_chan_public(c))
```
cdecker: This will need to change if we ever decide to not announce channels ourselves.

lightningd/peer_htlcs.c
```diff
@@ -569,30 +569,18 @@ struct gossip_resolve {
...
+	if (!fromwire_gossip_get_channel_peer_reply(msg, msg, &peer_id)) {
 		log_broken(gossip->log,
-			   "bad fromwire_gossip_resolve_channel_reply %s",
+			   "bad fromwire_gossip_get_channel_peer_reply %s",
 			   tal_hex(msg, msg));
 		return;
 	}
 
-	if (tal_count(nodes) == 0) {
+	if (!peer_id) {
```
cdecker: Nice simplification :+1:

gossipd/gossipd.c
rustyrussell: Good thought. Not sure what the right thing is here though: why did we want it private?

PR #2025, opened 2018-10-15
Update Bitcoin and Litecoin binaries to fix CVE-2018-17144

romanornr:
https://bitcoincore.org/en/2018/09/20/notice/

Might be worth to don't use the versions that are vulnerable.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ RUN apk add --no-cache \
 
 WORKDIR /opt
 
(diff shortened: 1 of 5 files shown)


romanornr:
Oeps, I see this is duplicate https://github.com/ElementsProject/lightning/pull/2021

Well except with a Litecoin fix.

cdecker:
If you reduce this to only touch the litecoin part and not downgrade bitcoind I'm ok with this :-)

PR #2036, opened 2018-10-18
WIP: Update for latest bolt version: bca814e270dcbee2fea51c0a26ca99efef261f2b

rustyrussell:
The only change is that the final_incorrect_htlc_amount field is now 64
bit.  Since no implementation yet parses that field, we just updated it
quietly in the spec.


rustyrussell:
OK, now based on #2054

rustyrussell:
Ack 35d63cf7a8f0c8f01d356a61a3acd8059b7d6214 (dup of @niftynei ack, post-rebase).

PR #2045, opened 2018-10-22
bitcoin: If we fail to estimate the fee in testnet use the minfee

cdecker:
When developing in regtest or testnet it is really inconvenient to
have to fake traffic and generate blocks just to get estimatesmartfee
to return a valid estimate. This just sets the minfee if bitcoind
doesn't return a valid estimate.


lightningd/bitcoind.c
```diff
@@ -352,7 +352,15 @@ static bool process_estimatefee(struct bitcoin_cli *bcli)
...
 		log_unusual(bcli->bitcoind->log, "Unable to estimate %s/%u fee",
 			    efee->estmode[efee->i], efee->blocks[efee->i]);
-		efee->satoshi_per_kw[efee->i] = 0;
+
+		/* If we are in testnet mode we want to allow payments
+		 * with the minimal fee even if the estimate didn't
+		 * work out. This is less disruptive than erring out
+		 * all the time. */
+		if (get_chainparams(bcli->bitcoind->ld)->testnet)
+			efee->satoshi_per_kw[efee->i] = 253;
```
rustyrussell: FEERATE_FLOOR?

cdecker:
> Maybe make it !DEVELOPER only?

Won't really help, since we removed `--dev-override-fees` and `dev-setfees`, and are now manipulating the fees through the RPC proxy instead. There is no difference between dev mode anymore. I think the only option is add a dev option to set to 0 on failure instead, but that's not really nice either.

rustyrussell:
No, I'm saying make the "fallback to FEERATE_FLOOR on testnet" iff !DEVELOPER.  That means we can still test what happens when we can't estimate fee...

cdecker:
Ok, this now works. Rebased and squashed, ready for review @rustyrussell

PR #2049, opened 2018-10-24
make daemon_conn and msg_queue tal objects

rustyrussell:
This is more tal-idiomatic IMHO.  Though it means an extra allocation at startup time, I think it adds clarity.

### DIFF
--- a/common/daemon_conn.c
+++ b/common/daemon_conn.c
@@ -69,27 +69,21 @@ bool daemon_conn_sync_flush(struct daemon_conn *dc)
 static struct io_plan *daemon_conn_start(struct io_conn *conn,
 					 struct daemon_conn *dc)
 {
-	dc->conn = conn;
 	return io_duplex(conn, daemon_conn_read_next(conn, dc),
 			 daemon_conn_write_next(conn, dc));
 }
 
 void daemon_conn_init(tal_t *ctx, struct daemon_conn *dc, int fd,
 		      struct io_plan *(*daemon_conn_recv)(struct io_conn *,
-							  struct daemon_conn *),
-		      void (*finish)(struct io_conn *, struct daemon_conn *dc))
+							  struct daemon_conn *))
 {
-	struct io_conn *conn;
-
 	dc->daemon_conn_recv = daemon_conn_recv;
 
 	dc->ctx = ctx;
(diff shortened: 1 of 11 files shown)


common/daemon_conn.c
```diff
@@ -73,23 +101,30 @@ static struct io_plan *daemon_conn_start(struct io_conn *conn,
...
-	dc->msg_in = NULL;
-	msg_queue_init(&dc->out, dc->ctx);
-	dc->msg_queue_cleared_cb = NULL;
-	dc->conn = io_new_conn(ctx, fd, daemon_conn_start, dc);
+	/* Harmless free loop if conn is being destroyed because dc freed */
+	tal_free(dc);
 }
 
-void daemon_conn_clear(struct daemon_conn *dc)
+struct daemon_conn *new_daemon_conn_(const tal_t *ctx, int fd,
```
cdecker: Keeping with the `ccan` naming convention this should be `daemon_conn_new_`.

common/msg_queue.c
```diff
@@ -1,18 +1,22 @@
...
 #include <common/msg_queue.h>
 #include <common/utils.h>
 #include <wire/wire.h>
 
-void msg_queue_init(struct msg_queue *q, const tal_t *ctx)
+struct msg_queue {
+	const u8 **q;
+};
+
+struct msg_queue *new_msg_queue(const tal_t *ctx)
```
cdecker: Nit; naming (`msg_queue_new`)

common/daemon_conn.c
rustyrussell: Yes.  Good point!

common/msg_queue.c
rustyrussell: Ack...

cdecker:
Rebased on top of `master` and squashed fixup. Will merge once Travis passes (and my plane lands) :-)

cdecker:
No idea why `test_onchain_multihtlc_their_unilateral` seems to stall and not complete with `VALGRIND=0`, `DEVELOPER=1` on `ARCH=32`

rustyrussell:
Looks like a rebase error ATM: the first commit does not belogn?

PR #2052, opened 2018-10-25
leak detection enabled (iff DEVELOPER) for all subdaemons.

rustyrussell:
Looks clean now, please review.


channeld/channeld.c
```diff
@@ -2656,6 +2656,14 @@ static void init_channel(struct peer *peer)
 	 * create error packet if necessary. */
 	init_shared_secrets(peer->channel, htlcs, hstates);
 
+	/* We don't need these any more, so free them. */
+	tal_free(htlcs);
+	tal_free(hstates);
+	tal_free(fulfilled);
+	tal_free(fulfilled_sides);
+	tal_free(failed);
+	tal_free(failed_sides);
```
cdecker: Any chance we can group them into a common parent?

rustyrussell:
> b90b313 says it is also annotating with notleak but then doesn't

Oops, will fix.  I think that's because I decided to separate the patch later.

> In general I'm not sure if bisectability is maintained since it seems to enable checks and patch the subdaemons up afterwards.

I think I got the order correct: <fixup x>, <wire up dev-memleak for x>.  But I'll do a thorough test today.

> onchaind has a bit too much magic with the "everything before this is ok"-rule, could we group these into a singleton that acts as a root?

I'll take a look; generally I was trying *not* to alter the logic to add in memleak detection, but this is a good argument.

> We might as well byte the bullet and have closingd also talk to lightningd it might come in handy when a close is stuck to introspect (some users have complained that close is a bit of a black box)

Agreed.

channeld/channeld.c
rustyrussell: Almost.  They're allocated by `fromwire_channel_init`, off `peer`.  But it seems we want to discard most of them so we should allocate off tmpctx and explicitly steal the ones we actually want to keep...

niftynei:
looks like the tests are failing to compile in some cases, but provided that the tests are passing 

ACK https://github.com/ElementsProject/lightning/pull/2052/commits/4222e28b43b963e69f17904291efa2eb12f350e2

cdecker:
Needs one more rebase (the last one, I promise)

rustyrussell:
Nope, one *more* rebase!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2052/commits/efb262da6038bad7be5ad007a6a2cf198a46ed8a

PR #2053, opened 2018-10-26
Making releases notes and helper script.

rustyrussell:
Please give feedback on anything I missed, particularly docker commands and the checklist...


rustyrussell:
I'm just trying to automate what @cdecker did previously.  Doesn't help for arm though, which would be nice if course!

Technically these can all be cross-compiled, but they can't be tested them...

contrib/build-release.sh
```diff
@@ -0,0 +1,49 @@
...
+	    DOCKERFILE=contrib/Dockerfile.builder.i386
+	    TAG=ubuntu-i386
+	    ;;
+	*)
+	    echo "No Dockerfile for $platform" >&2
+	    exit 1
+    esac
+
+    docker build -f $DOCKERFILE -t $TAG .
+    docker run --rm=true -v "$(pwd)":/build -w /build $TAG ./configure
```
cdecker: We could save on the make distclean (and make really sure we don't have stale things around by `git clone`ing in the docker container, merge the steps below and just delete the container afterwards. That way we don't rely on `distclean` getting absolutely everything.

cdecker:
Looking good :+1:

The PPA instructions are a bit more involved because we need to materialize the external dependencies, i.e., remove the submodule, checkout, remove `.git` dirs and add them in the tree (which usually makes the clone unusable since submodule conflicts are completely mental). Maybe we can move away from external deps altogether, at some point?

cdecker:
Flaky test now pass again

PR #2058, opened 2018-10-29
closingd: retransmit `shutdown` on reconnect.

rustyrussell:
The spec says so, and it's right: with the right pattern of packet loss
(thanks Travis!) the other end can still be in channeld, waiting for our
`shutdown` message.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2058/commits/6142044fc06af4466c8dee478663f166d5cfcbf7

PR #2064, opened 2018-10-31
Add support for aarch64 in Dockerfile

78bash:

PR #2066, opened 2018-11-01
Litecoin versioning fixup + support for aarch64

78bash:


78bash:
Did travis-ci checks fail because I made 2 pull requests for the same file?

cdecker:
Nope, seems we have a test that hangs somewhere. Could you also parametrize the bitcoin arch with `uname`? There's no point in having it download the amd64 version if we run on a different arch.

Speaking of which, I'm not sure about this, but wouldn't the base `alpine` image always be an amd64 image? Or is docker smart enough to use the host arch and pull the matching images if they exist?

cdecker:
Just saw your other PR #2064, so no need to amend this here, but this PR will need a rebase once #2064 is merged, since there is a merge conflict.

78bash:
RUN will treat its arguments as cli commands. So `uname -m` will return the arch available on the server. This is the same arch portion for amd64 and aarch64 used in the bitcoin and litecoin tarballs.

78bash:
Looks like there's also an aarch64 version of Alpine

cdecker:
Superseded by #2071 and #2064

PR #2070, opened 2018-11-02
Help special needs gcc versions

jb55:
This fixes a compilation error on gcc 7.3.0

### DIFF
--- a/onchaind/onchaind.c
+++ b/onchaind/onchaind.c
@@ -1293,7 +1293,7 @@ static size_t resolve_our_htlc_ourcommit(struct tracked_output *out,
 					 const struct htlc_stub *htlcs,
 					 u8 **htlc_scripts)
 {
-	struct bitcoin_tx *tx;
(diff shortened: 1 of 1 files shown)


cdecker:
Seems like a false positive that is not covering up a real issue:

This line asserts that we run through the loop which does the assignment at least once:

https://github.com/ElementsProject/lightning/blob/d5bb536ef0c08a813f767b3fb016eb20292de4dd/onchaind/onchaind.c#L1300

And this is the loop assigning:

https://github.com/ElementsProject/lightning/blob/d5bb536ef0c08a813f767b3fb016eb20292de4dd/onchaind/onchaind.c#L1304-L1322

While this goes a bit against the "no dummy assignments"-rule, I think this is ok for once.

ACK https://github.com/ElementsProject/lightning/pull/2070/commits/fc7db8da47b7bd5b2211d51e7438c9dd107c6a9c

Issue #934, opened 2018-02-06
Two null-param'd listpeers calls in a row crash daemon


icota:
### Issue and Steps to Reproduce

Send this via the RPC interface:
`{"id":1,"jsonrpc":"2.0","method":"listpeers","params":[null]}`
Returns as expected. Send another:
`{"id":2,"jsonrpc":"2.0","method":"listpeers","params":[null]}`
lightning steps on `assert(!jcon->current)` in `parse_request`. 

Reason being the second call never gets to `gossipd_getpeers_complete` and from there to its destructor. I'll be so bold and suggest this is a bug in `gossipd`

icota:
I'm not sure but I think my description and steps might be misleading. I see now the assertion failure also occurs on any call whose id is not 1 (or is generally not the first call?), whether params array is null or just plain empty:

`{"id":1,"jsonrpc":"2.0","method":"listpeers","params":[]}` ---> A-OK

`{"id":2,"jsonrpc":"2.0","method":"listpeers","params":[]}` ---> CRASH

cdecker:
Thanks for reporting this @icota. I'm having a bit of trouble reproducing this. How are you calling the RPC (python RPC client, `lightning-cli`, JS RPC client, ...). Calling `lightning-cli listpeers` in a loop (which internally then passes an empty params array) doesn't crash. Are the two calls concurrent?

icota:
It's a C++ RPC client. The calls are not concurrent they are in sequence but I can't reproduce with `lightning-cli` either. I'm guessing that's cause my client doesn't close the connection after every call.

icota:
Here's the output when it occurs:
```
2018-02-07T11:28:28.272Z lightningd(24586): Connected json input
2018-02-07T11:28:31.857Z lightning_gossipd(24594): TRACE: req: type WIRE_GOSSIP_GETPEERS_REQUEST len 4
2018-02-07T11:28:31.857Z lightning_gossipd(24594): REPLY WIRE_GOSSIP_GETPEERS_REPLY with 0 fds
2018-02-07T11:28:31.857Z lightningd(24586):jcon fd 14: Success
lightningd: lightningd/jsonrpc.c:562: parse_request: Assertion `!jcon->current' failed.
2018-02-07T11:28:31.857Z lightningd(24586): FATAL SIGNAL 6 RECEIVED
2018-02-07T11:28:31.865Z lightningd(24586): backtrace: lightningd/log.c:473 (log_crash) 0x7471114d41
[...]

PR #2081, opened 2018-11-13
Fix multirpc for real

rustyrussell:
It started with travis hanging on this occasionally, and it ended with a thing of beauty!*

\* May be in the eye of the beholden...

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -25,25 +25,22 @@ class UnixDomainSocketRpc(object):
         s = json.dumps(obj)
         sock.sendall(bytearray(s, 'UTF-8'))
 
-    def _readobj(self, sock):
-        buff = b''
+    def _readobj(self, sock, buff=b''):
+        """Read a JSON object, starting with buff; returns object and any buffer left over"""
         while True:
             try:
-                b = sock.recv(1024)
-                buff += b
-                if len(b) == 0:
-                    return {'error': 'Connection to RPC server lost.'}
-
-                if buff[-3:] != b' }\n':
-                    continue
-
                 # Convert late to UTF-8 so glyphs split across recvs do not
                 # impact us
-                objs, _ = self.decoder.raw_decode(buff.decode("UTF-8"))
-                return objs
+                objs, len_used = self.decoder.raw_decode(buff.decode("UTF-8"))
+                return objs, buff[len_used:].lstrip()
             except ValueError:
                 # Probably didn't read enough
                 pass
+
+            b = sock.recv(1024)
+            buff += b
+            if len(b) == 0:
+                return {'error': 'Connection to RPC server lost.'}, buff.lstrip()
 
     def __getattr__(self, name):
         """Intercept any call that is not explicitly defined and call @call
@@ -65,6 +62,7 @@ class UnixDomainSocketRpc(object):
         # Filter out arguments that are None
         payload = {k: v for k, v in payload.items() if v is not None}
 
+        # FIXME: we open a new socket for every readobj call...
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         sock.connect(self.socket_path)
         self._writeobj(sock, {
@@ -72,7 +70,7 @@ class UnixDomainSocketRpc(object):
             "params": payload,
             "id": 0
         })
-        resp = self._readobj(sock)
+        resp, _ = self._readobj(sock)
         sock.close()
 
         self.logger.debug("Received response for %s call: %r", method, resp)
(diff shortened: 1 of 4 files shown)


tests/test_misc.py
```diff
@@ -605,13 +605,13 @@ def test_multirpc(node_factory):
         b'{"id":5,"jsonrpc":"2.0","method":"listpeers","params":[]}',
         b'{"id":6,"jsonrpc":"2.0","method":"listpeers","params":[]}',
         b'{"method": "invoice", "params": [100, "foo", "foo"], "jsonrpc": "2.0", "id": 7 }',
-        b'{"method": "waitinvoice", "params": ["foo"], "jsonrpc" : "2.0", "id": 8 }',
```
ZmnSCPxj: What?? This will return once `delinvoice` executes.  In fact, this is a key part of the test: we have different paths for commands that complete immediately, and commands that will defer, and we should test both paths. Both `invoice` and `delinvoice` exercise the path that completes immediately, while `waitinvoice` exercises the path where it defers.  It is important to check both.

rustyrussell: Ah, good catch!  I wonder why it didn't work, will debug!

ZmnSCPxj: If I remember correctly, there was an issue once that a user complained that parallel JSON-RPC connections could run commands in parallel, but a single JSON-RPC connection could not, which is why we moved from processing commands linearly, to processing them in parallel. This test was intended to check precisely this change, I think. I will try to see if if I can dig up this issue.

ZmnSCPxj: Here: https://github.com/ElementsProject/lightning/issues/934#issuecomment-364274318

The user was complaining about parallel `listpeers`, and not the pattern in that comment (which, it turns out, was mine, and is precisely what is tested in `test_multirpc`: `invoice` `waitinvoice` `delinvoice`).

However it seems a regression to me to remove our ability to run commands in parallel within a single connection.  If this feature is removed, then `multirpc` should also be removed outright, not "fixed" by not testing what it is intended to test in the first place.

rustyrussell: Hmm, I've thought some more about this, and I agree.  We should change the limit such that it only stop processing input if there's still stuff in the output buffer.

That's actually quite a large change, as it means resolving the case where two commands interleave their returns.  We need that change for notifications, but now it's important for commands as well.

I'll do this as a separate PR; for this one I'll simply note the FIXME.

contrib/pylightning/lightning/lightning.py
```diff
@@ -25,26 +25,23 @@ def _writeobj(sock, obj):
         s = json.dumps(obj)
         sock.sendall(bytearray(s, 'UTF-8'))
 
-    def _readobj(self, sock):
-        buff = b''
+    def _readobj(self, sock, buff=b''):
```
cdecker: Please just make `buff` a member of `UnixDomainSocketRpc`, passing it out and back in is not idiomatic and will cause misuse.

cdecker:
Wait, how can we ever end up with more than one response if every call gets its own socket? Are we getting wires crossed internally somehow?

contrib/pylightning/lightning/lightning.py
rustyrussell: No, that *doesnt' work* (and took me a few hours debugging!).  It's simple to change if you want to see it yourself.

When the daemon restarts, for example, we can read half a reply, then fail.  Similarly if we use timeouts in our testing.  That half-reply is stored in self.buff.  Now we try to use the same UnixDomainSocketRPC for another command, and *append* the result to self.buff and get garbage.

And "passing back" is AFAICT the idiomatic way to do pass-by-reference in Python?

(We really want to make it a member of 'sock', so it gets cleared correctly when sock is recreated.  Though even better would be to fix pylightning not to reopen unless necessary?)

contrib/pylightning/lightning/lightning.py
```diff
@@ -25,25 +25,20 @@ def _writeobj(sock, obj):
...
-                # impact us
-                objs, _ = self.decoder.raw_decode(buff.decode("UTF-8"))
-                return objs
-            except ValueError:
-                # Probably didn't read enough
-                pass
+                    return {'error': 'Connection to RPC server lost.'}, buff
+            else:
+                buff = parts[1]
+                obj, _ = self.decoder.raw_decode(parts[0].decode("UTF-8"))
```
conscott: Nit: Being paranoid, you could still keep the try / `except ValueError` on `raw_decode` call to catch if there is ever un-decodable garbage in the buffer

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2081/commits/c35c23d7a24f738ebb0aa67be38d10adc116a1a8

We'll revisit this later again

PR #2083, opened 2018-11-13
[doc] Add lightning-pay script as example for pylightning

conscott:
I found the script useful since it uses the [default socket](https://github.com/ElementsProject/lightning/blob/master/contrib/pylightning/lightning-pay#L15-L23) to setup the LightningRPC object, which I imagine other people will want to do when playing with the library.


jb55:
Does that script still work? I found it hangs the last few times I've tried to use it.

conscott:
You are right, fixed in #2088

PR #2085, opened 2018-11-14
[doc] Remove old GOSSIPING state from README

conscott:
The `GOSSIPING` state was removed some time ago: see [CHANGELOG](https://github.com/ElementsProject/lightning/blame/master/CHANGELOG.md#L133-L136)

I think it's confusing to keep it here since it's not an actual  internal state or text that can be returned from the RPC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,8 +195,6 @@ that `state` is `CHANNELD_NORMAL`; after 6 confirmations you can use
 `true`.
 
 ### Different states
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2085/commits/8d0997e92f14728aa6e240f4512ad5ee3f9fa8a3

PR #2096, opened 2018-11-20
bolt11: accept lightning: prefix.

rustyrussell:
The Blockstream store produces these, for example, so let's ignore them.

### DIFF
--- a/common/bolt11.c
+++ b/common/bolt11.c
@@ -467,6 +467,14 @@ struct bolt11 *bolt11_decode(const tal_t *ctx, const char *str,
 
         b11->routes = tal_arr(b11, struct route_info *, 0);
 
+	/* BOLT #11:
+	 *
+	 * If a URI scheme is desired, the current recommendation is to either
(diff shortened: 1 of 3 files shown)


common/bolt11.c
```diff
@@ -467,6 +467,14 @@ struct bolt11 *bolt11_decode(const tal_t *ctx, const char *str,
 
         b11->routes = tal_arr(b11, struct route_info *, 0);
 
+	/* BOLT #11:
+	 *
+	 * If a URI scheme is desired, the current recommendation is to either
+	 * use 'lightning:' as a prefix before the BOLT-11 encoding
+	 */
+	if (strstarts(str, "lightning:") || strstarts(str, "LIGHTNING:"))
```
ZmnSCPxj: Why we do not accept `lIghtnIng:`?  Would it be better to apply some case-conversion?

(bech32 is case-insensitive in any case....)

cdecker: bech32 does not allow for mixed case.

Issue #2051, opened 2018-10-24
[Feature Request] Stop depending on bitcoin-cli 


NicolasDorier:
It would be nice if there was no dependency on bitcoin-cli.

It makes a multi arch docker image harder to create as you need to pull out the right binaries
Because of 0.17.0 having breaking change in configuration, a clightning image can't target at the same time a 0.16.3 node and a 0.17.0 node on something else than mainnet
Because setting up some configuration value for regtest and testnet require a section in 0.17.0, but such section can't be read from 0.16.3 bitcoin-cli tooling. And not using a section on regtest and testnet on 0.17.0 will result in some config parameter which not being properly applied by bitcoind, while it would be applied by bitcoin-cli 0.16.3 tooling.

I really blame core to have messed up so badly with breaking changes in configuration files... It makes bitcoin-cli dependency an even bigger burden.

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: About deprecated feature; Very old (7.3 years), no comments, inactive 2667 days

If you believe this issue should remain open, please comment and we can reopen it.

PR #2111, opened 2018-11-26
Add native bitcoin rpc client and remove dependency on bitcoin-cli

arowser:
implement #2051

Remove dependency on bitcoin-cli,  the new rpc api is based libevent library,  the implementation  reference bitcoin-cli.

Configure rpc paramters is same with old 
```
--bitcoin-rpcconnect=127.0.0.1
--bitcoin-rpcpassword=<rpcpass>
--bitcoin-rpcport=8332 
--bitcoin-rpcuser=rpcuser 
```
or
```
--bitcoin-rpccookiefile=<cookiefile path>
```

Still not support read rpc config from bitcoin.conf.


arowser:
The fail reason is the docker image no libevent library, so add another pr #2112 to update Dockerfile.

rustyrussell:
Mixing event loops like this is always kinda ugly: in this case we simply spin polling the http socket which is pretty awful.

I know Christian intended to push bitcoin-cli usage out to a plugin, for which this would be an awesome replacement!  So I'll leave this one to him to review...

arowser:
@rustyrussell  I agree mixing event loop is ugly, since I could't find another way to make libevnt work, the lightningd is a single thread process, and with ccan io_loop,  that's why I polling the libevent in io_loop,  if we can keep this, and move it  to a  plugin after plugin subsystem finished?

cdecker:
Oh, I misjudged the aim of this PR, I thought it was a separate binary. I think this is much better suited as a plugin as well.

cdecker:
Any chance this can be ported on top of `io_loop` in a minimal fashion? We know quite well what our counterparty (`bitcoind`) expects so we wouldn't have to handle too many nits of HTTP.

arowser:
@cdecker I can't figure out how to do that, the lightningd is a single thread, but every rpc request need wait bitcoind response, so its need async call rpc for performance, event if use bitcoin-cli or make it a plugin, so a event queue need to handle async messages, if not use libevent  and create a raw socket rpc client,  it will make that more complication, and still need do same things to handle async messages.

ZmnSCPxj:
> I can't figure out how to do that, the lightningd is a single thread, but every rpc request need wait bitcoind response, so its need async call rpc for performance

You will essentially "roll your own" async call system on top of `io_loop`.

Basically, you will need to transform every query of `bitcoind` into a continuation-passing-style.

You will have a separate "main loop" to monitor for requests within `lightningd` to call out to `bitcoind`.  This main loop will be several C functions that do not directly call each other, but instead pass around each other as the "next function" to `io_loop` primitives.
The `io_loop` primitives can be viewed as continuation-passing style.

Through this, you can have a "green thread" that monitors for requests from `lightningd` to the `bitcoind` communication system.
You will need to implement something that coordinates requests (a queue) from the rest of `lightningd` to the `bitcoind` communication system, and if that queue is empty, simply idle/do nothing in the corresponding `io_loop`.  And if the queue-push operation moves from empty to non-empty, retrigger the `io_loop` section at that point.

Our command handling roughly follows this approach and might be a good study for this.

Higher-level languages like Scheme often implement green threads using approximately this approach.

arowser:
@ZmnSCPx Thanks for you detailed explanation.

I think the  code is working like your descripted, I try to explain that more clear in follow:

The current master rpc send and receive:
```
io_loop 
    |   
    |
    V
rpc rquest-->new process "bitcoin-cli"-->send rpc request with libevent-->bitcoind 
    |                          |                                            |
    |                          +<---------return rpc response --------------+
    |                          |
    |                          V
    |                    write result to stdout
    V
polling stdout if bitcoin-cli finished---yes-->call process_rpc
    |
    no
    |
    V
  next event
```

The new rpc send and receive:

```
io_loop event
    |   
    |
    V
rpc rquest --> send a rpc request with libevent
    |   
    |
    V
polling libevent queue-->if rpc return --> call porcess_rpc
    |   
    |
    V
   next 
```
So I think the new code not increase complication, its just move libevent part from
bitcoin-cli to lightningd, and remove the other useless code like process create, exit.

But this pr take libevent queue into io_loop, its not good, do you have any 
better solution for that?

ZmnSCPxj:
Why not send the RPC comman via `ccan/io`?  That way it integrates directly into `io_loop`.

arowser:
@ZmnSCPxj  If use ccan/io,  Its seems need add tcp and http client related code, its will make rpc client code more complicated.

@cdecker  @rustyrussell  If use the ccan/io,  I think need create a raw socket client,  and use socket descriptor as file descriptor in ccan/io,  then add implment tcp and http in read/write io. could you please confim if these are right steps?  And do you think this a better solutions for rpc client? I can try to do that if you agree.

ZmnSCPxj:
I believe only http is needed, as TCP should be done by OS.

I think more important to reduce dependency on external code. External code we use is fixed to particular github revision. This prevents cases where somebody injects Bitcoin thief into a lib that a Bitcoin project uses.

cdecker:
The one downside that using `ccan/io` has is that we don't (yet) have an HTTP library that parses requests. Given the many idiosyncrasies and pitfalls that HTTP brings with it that's a big task. However in this case we know exactly who we're talking to (`bitcoind`) and we can tailor the parsing to whatever minimal headers it spits out.

As for the parts that need implementing: TCP streams are implemented, you just need to buffer and then parse the headers, and afterwards buffer the payload which you can pass 1-to-1 back to the caller.

arowser:
replaced by #2295

PR #2113, opened 2018-11-26
enable add_steal_notifiers only support backtrace library

arowser:
FIx build error on Mac OS:
```
common/daemon.c:101:13: error: function 'add_steal_notifiers' is not needed and will not be emitted
      [-Werror,-Wunneeded-internal-declaration]
```


rustyrussell:
Thanks, good catch!

Issue #2090, opened 2018-11-18
macOS (10.14.1): Implicit declaration of function 'kill' is invalid in c99


brakmic:
### Issue and Steps to Reproduce

Today I have tried to compile the last code I could get from your master.
My configure command: `./configure --enable-developer --disable-valgrind`

The compilation error is:

`lightningd/plugin.c:
Implicit declaration of function 'kill' is invalid in c99`

My `cc -v`:

```shell
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
````

My `clang -v`:

```shell
clang version 5.0.0 (http://root.cern.ch/git/clang.git 7fd3024be56d751958d68ea3abeca4ab2f89dd91) (http://root.cern.ch/git/llvm.git e0b472e46eb5861570497c2b9efabf96f2d4a485)
Target: x86_64-apple-darwin18.2.0
Thread model: posix
```

After having searched around a bit I was able to compile it by adding these includes into `lightningd/plugin.c`

```c
#include <signal.h>
#include <unistd.h>
#include <sys/types.h>
#include <sys/wait.h>
```

The `unistd.h`was already there but others were missing.

This is actually not an issue anymore but more a hint to other macOS users. There is also another problem I faced during compilation which is not related to `plugin.c` but to the `developer mode` instead. So I will open another *issue*.

I am not sure if your project has a Wiki or similar platform, therefore *sorry* for having misused the **Issues** for posting the above info.

Kind regards,

PR #2114, opened 2018-11-26
add needed include file

arowser:
Fix build error on MacOS:

```
lightningd/plugin.c:111:2: error: implicit declaration of function 'kill' is invalid in C99
      [-Werror,-Wimplicit-function-declaration]
        kill(plugin->pid, SIGKILL);
        ^
```

PR #2115, opened 2018-11-26
fixes make error in the plugin system for mac users 

renepickhardt:
I don't know why @brakmic only opened an issue #2090 instead of providing this patch. I believe this enhancement will be useful for any mac user as I just experienced exactly the same problem.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -9,6 +9,12 @@
 #include <lightningd/json.h>
 #include <unistd.h>
 
+#if defined(__APPLE__) && defined(__MACH__)
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#endif
+
 struct plugin {
 	pid_t pid;
 	char *cmd;


lightningd/plugin.c
```diff
@@ -9,6 +9,12 @@
 #include <lightningd/json.h>
 #include <unistd.h>
 
+#if defined(__APPLE__) && defined(__MACH__)
```
ZmnSCPxj: Why not just add all these includes unconditionally?

renepickhardt: as far as I understood the issue this problem only occurred if macos is used. that is why I added them solely for macos

brakmic:
Hi @renepickhardt 

I simply wasn't sure whether I should provide a patch for macOS that is not officially supported. I saw in their README that macOS *could work* after having applied some extras. And indeed I had to change a few things to make it work on my mac. 

So I thought I'd better avoid opening a Pull Request and let others, more experienced people decide.
But in any case, many thanks for having "escalated" it 😺 

Best,
Harris

cdecker:
For what it's worth the CI failures seem to be only complaining about the order of the header includes, which is expected to be in lexicographic order (except some small exceptions).

renepickhardt:
this means that we have to have several #ifdefs ? or just put the headers in lexecographic order within the #ifdef?

conscott:
Also seems to be a duplicate with #2114  (which includes lexographically, but doesn't ifdef)

ZmnSCPxj:
If it works even on non Mac systems, I suggest to do without `#ifdef` and add in lexicographical order.

renepickhardt:
@arowser sorry for the duplicate PR! I could make a PR on your side include the #if !defined() and then we ask to merge your request.

cdecker:
So with `<signal.h>` included, does this now work on Mac OS?

cdecker:
AFAIK the latest `master` now compiles correctly. Is that true @renepickhardt ?

renepickhardt:
Yes

cdecker:
Thanks for confirming @renepickhardt :+1:

PR #2117, opened 2018-11-26
Small intermezzo in the endless plugin saga (Part 1.5, fixes)

### LINKED ISSUE #2116: Lightningd does not start with --plugin= command line argument
Maybe this is a mac issue related to #2090 but I cannot start `lightningd` when passing an argument. 
first of all the documentation contradicts itself:
I find that the plugins need to be called with `--plugin=PATH` and with `--plugin PATH` the first version stops lightningd to start where as the second version will have lightningd starting up but ignoring the code from the hello world example plugin. 

as far as I understood the file `docs/plugin.md` I should have been able to make a call like `lightning-cli hello` or `lightning-cli --hello` or with `greeting` instead of `hello`. Also the `plugin.md` states that it adds command line arguments to lightningd but I guess that does not make sense as lightningd will already be started if everything goes successfully.

Would be great if you could help out quickly as I am about to finish the next youtube video and wanted to provide the code also as a c-lightning plugin

### `getinfo` output (if started without plugin argument)
```
ln1 getinfo
{
  "id": "02eea4a069fe8298b63ce0cfee67a68f85f594b0385b3b0f062ff92fcbeb0fa446", 
  "alias": "YELLOWSPAWN", 
  "color": "02eea4", 
  "num_peers": 0, 
  "num_pending_channels": 0, 
  "num_active_channels": 0, 
  "num_inactive_channels": 0, 
  "address": [
  ], 
  "binding": [
    {
      "type": "ipv4", 
      "address": "127.0.0.1", 
      "port": 9600
    }
  ], 
  "version": "v0.6.2-112-gb8bf887", 
  "blockheight": 0, 
  "network": "regtest", 
  "msatoshi_fees_collected": 0
}
```

cdecker:
Just a bunch of fixes that came up while working on the JSON-RPC passthrough. These include:

 - Adding the `pytest-timeout` plugin with a config that supports `pytest-xdist`, and hopefully gives us a bit more context before Travis-CI kills a job for inactivity.
 - Upgraded the `pytest` dependencies in the builder
 - Move plugins into a `contrib/plugins` instead of having a sub-directory for each plugin.
 - Fix a naming issue after we renamed `init` -> `getmanifest` and `configure` -> `init`, now the plugin example actually works and we test it.

Fixes #2116


renepickhardt:
maybe already include https://github.com/ElementsProject/lightning/pull/2115 to this pull request to enable building for macos. I had to fix it for this branch again.
I believe  `lightningd --network regtest --lightning-dir /tmp/spark-env/ln1 --bitcoin-datadir /tmp/spark-env/btc --addr 127.0.0.1:9600 --log-level=debug --plugin=/PATH/TO/lightning/contrib/plugins/helloworld.py` works now however I can't see the effect of the plugin as I already ran lightningd with the plugin to be activated so how would I interact with lightningd?

contrib/Dockerfile.builder
```diff
@@ -51,4 +51,19 @@ RUN cd /tmp/ && \
     rm -rf bitcoin.tar.gz /tmp/bitcoin-$BITCOIN_VERSION
 
 RUN pip3 install --upgrade pip && \
-    python3 -m pip install python-bitcoinlib==0.7.0 pytest==3.0.5 setuptools==36.6.0 pytest-test-groups==1.0.3 flake8==3.5.0 pytest-rerunfailures==3.1 ephemeral-port-reserve==1.1.0 pytest-xdist==1.22.2 flaky==3.4.0 CherryPy==17.3.0 Flask==1.0.2
+    python3 -m pip install \
```
conscott: Is there a reason these requirements are specifically listed out for each Dockerfile rather than just sharing 
a requirements.txt and installing like:
```
COPY requirements.txt /tmp
...
python3 -m pip install -r /tmp/requirements.txt
```

cdecker: If we want to have files that are copied in we need to send the docker build context to the docker daemon. By making the `Dockerfile`s self-contained we can use the following:

```
docker build - < contrib/Dockerfile.builder
```

Which saves us from sending the entire repo, and whatever files you have in the same directory to the daemon (2GB in my case...). Also we'd need to keep a separate `requirements.txt` since we want to add more dependencies to the builder than what we require from developers to have on their box.

conscott: Got it :+1:  thanks for the explanation.

PR #2121, opened 2018-11-28
JSON-RPC passthrough for plugin (Endless plugin saga Part 2)

cdecker:
This is the second major plugin feature: plugins can register RPC methods with `lightningd`'s JSON-RPC interface, and calls to those methods will be transparently routed to the plugin.

The preparatory work requires wrapping the information related to the JSON-RPC in a new struct. This also gives us a clean root object for all related objects as well, which was so far done with the listener object. Furthermore the JSON-RPC instantiation and initialization (listening) was split so we can instantiate and add new methods before starting to listen, and the example plugin now handles exceptions and array-based as well as dict-based arguments correctly (I plan on formalizing this in a utility Plugin class for python plugins once we have plugin -> lightningd notifications for logging).

### DIFF
--- a/doc/plugins.md
+++ b/doc/plugins.md
@@ -82,7 +82,15 @@ currently only string options are supported.*
 The `rpcmethods` are methods that will be exposed via `lightningd`'s
 JSON-RPC over Unix-Socket interface, just like the builtin
 commands. Any parameters given to the JSON-RPC calls will be passed
-through verbatim.
+through verbatim. Notice that the `name` and the `description` fields
+are mandatory, while the `long_description` can be omitted (it'll be
+set to `description` if it was not provided).
+
+Plugins are free to register any `name` for their `rpcmethod` as long
+as the name was not previously registered. This includes both built-in
+methods, such as `help` and `getinfo`, as well as methods registered
+by other plugins. If there is a conflict then `lightningd` will report
+an error and exit.
 
 ### The `init` method
 
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -77,6 +77,18 @@ struct json_connection {
 	struct json_stream **js_arr;
 };
 
+/**
+ * `jsonrpc` encapsulates the entire state of the JSON-RPC interface,
+ * including a list of methods that the interface supports (can be
+ * appended dynamically, e.g., for plugins, and logs. It also serves
+ * as a convenient `tal`-parent for all JSON-RPC related allocations.
+ */
+struct jsonrpc {
+	struct io_listener *rpc_listener;
+	struct json_command **commands;
+	struct log *log;
+};
+
 /* The command itself usually owns the stream, because jcon may get closed.
  * The command transfers ownership once it's done though. */
 static struct json_stream *jcon_new_json_stream(const tal_t *ctx,
@@ -292,10 +304,9 @@ static void json_add_help_command(struct command *cmd,
 static void json_help(struct command *cmd,
 		      const char *buffer, const jsmntok_t *params)
 {
-	unsigned int i;
 	struct json_stream *response;
-	struct json_command **cmdlist = get_cmdlist();
 	const jsmntok_t *cmdtok;
+	struct json_command **commands = cmd->ld->jsonrpc->commands;
 
 	if (!param(cmd, buffer, params,
 		   p_opt("command", json_tok_tok, &cmdtok),
@@ -303,10 +314,10 @@ static void json_help(struct command *cmd,
 		return;
 
 	if (cmdtok) {
-		for (i = 0; i < num_cmdlist; i++) {
-			if (json_tok_streq(buffer, cmdtok, cmdlist[i]->name)) {
+		for (size_t i = 0; i < tal_count(commands); i++) {
+			if (json_tok_streq(buffer, cmdtok, commands[i]->name)) {
 				response = json_stream_success(cmd);
-				json_add_help_command(cmd, response, cmdlist[i]);
+				json_add_help_command(cmd, response, commands[i]);
 				goto done;
 			}
 		}
@@ -320,8 +331,8 @@ static void json_help(struct command *cmd,
 	response = json_stream_success(cmd);
 	json_object_start(response, NULL);
 	json_array_start(response, "help");
-	for (i = 0; i < num_cmdlist; i++) {
-		json_add_help_command(cmd, response, cmdlist[i]);
+	for (size_t i=0; i<tal_count(commands); i++) {
+		json_add_help_command(cmd, response, commands[i]);
 	}
 	json_array_end(response);
 	json_object_end(response);
@@ -330,17 +341,17 @@ done:
 	command_success(cmd, response);
 }
 
-static const struct json_command *find_cmd(const char *buffer,
+static const struct json_command *find_cmd(const struct jsonrpc *rpc,
+					   const char *buffer,
 					   const jsmntok_t *tok)
 {
-	unsigned int i;
-	struct json_command **cmdlist = get_cmdlist();
+	struct json_command **commands = rpc->commands;
 
-	/* cmdlist[i]->name can be NULL in test code. */
-	for (i = 0; i < num_cmdlist; i++)
-		if (cmdlist[i]->name
-		    && json_tok_streq(buffer, tok, cmdlist[i]->name))
-			return cmdlist[i];
+	/* commands[i]->name can be NULL in test code. */
+	for (size_t i=0; i<tal_count(commands); i++)
+          if (commands[i]->name &&
+              json_tok_streq(buffer, tok, commands[i]->name))
+            return commands[i];
 	return NULL;
 }
 
@@ -516,9 +527,9 @@ static void parse_request(struct json_connection *jcon, const jsmntok_t tok[])
 		return;
 	}
 
-	/* This is a convenient tal parent for duration of command
-	 * (which may outlive the conn!). */
-	c = tal(jcon->ld->rpc_listener, struct command);
+	/* Allocate the command off of the `jsonrpc` object and not
+	 * the connection since the command may outlive `conn`. */
+	c = tal(jcon->ld->jsonrpc, struct command);
 	c->jcon = jcon;
 	c->ld = jcon->ld;
 	c->pending = false;
@@ -543,8 +554,8 @@ static void parse_request(struct json_connection *jcon, const jsmntok_t tok[])
 		return;
 	}
 
-	c->json_cmd = find_cmd(jcon->buffer, method);
-	if (!c->json_cmd) {
+        c->json_cmd = find_cmd(jcon->ld->jsonrpc, jcon->buffer, method);
+        if (!c->json_cmd) {
 		command_fail(c, JSONRPC2_METHOD_NOT_FOUND,
 			     "Unknown command '%.*s'",
 			     method->end - method->start,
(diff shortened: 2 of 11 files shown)


lightningd/jsonrpc.c
```diff
@@ -749,9 +806,9 @@ void setup_jsonrpc(struct lightningd *ld, const char *rpc_filename)
 
 	if (listen(fd, 1) != 0)
 		err(1, "Listening on '%s'", rpc_filename);
-
-	log_debug(ld->log, "Listening on '%s'", rpc_filename);
-	ld->rpc_listener = io_new_listener(ld->rpc_filename, fd, incoming_jcon_connected, ld);
+	jsonrpc->rpc_listener = io_new_listener(
+		ld->rpc_filename, fd, incoming_jcon_connected, ld);
+	log_debug(jsonrpc->log, "Listening on '%s'", ld->rpc_filename);
```
niftynei: nice fix.

lightningd/plugin.c
```diff
@@ -373,6 +407,182 @@ static bool plugin_opts_add(const struct plugin_request *req)
...
+		plugin_kill(
+		    plugin,
+		    tal_fmt(plugin,
+			    "rpcmethod does not have a string \"name\": %.*s",
+			    meth->end - meth->start, buffer + meth->start));
+		return false;
+	}
+
+	if (!desctok || desctok->type != JSMN_STRING) {
+		plugin_kill(plugin, tal_fmt(plugin,
```
niftynei: meta comment that we should make sure we make clear in the docs that a description is mandatory

lightningd/plugin.c
```diff
@@ -373,6 +407,182 @@ static bool plugin_opts_add(const struct plugin_request *req)
...
+	request->id = tal_strndup(request, buffer + idtok->start,
+				      idtok->end - idtok->start);
+	request->params = tal_strndup(request, buffer + params->start,
+				      params->end - params->start);
+	request->plugin = NULL;
+	request->cmd = cmd;
+
+	/* Find the plugin that registered this RPC call */
+	list_for_each(&plugins->plugins, plugin, list) {
+		for (size_t i=0; i<tal_count(plugin->methods); i++) {
```
niftynei: If I'm reading this correctly, this means that every plugin will need to register independent method names, ie they should namespace their methods so there's no conflict with another plugin, otherwise you might call the wrong plugin! (probably the first one that registered that method name)

renepickhardt: That is a very nice catch. However I believe forcing users of an API to do something like name spacing correctly will probably be hard to enforce.
Therefor I believe that c-lightning should handle the name spacing and not the plugin developer. In that sense within clightning there would be a namespace given by something like `plugin_name::plugin->method`

tests/test_plugin.py
```diff
@@ -27,3 +29,25 @@ def test_option_passthrough(node_factory):
...
+    plugin_path = 'contrib/plugins/helloworld.py'
+    n = node_factory.get_node(options={'plugin': plugin_path, 'greeting': 'Mars'})
+
+    # Make sure that the 'hello' command that the helloworld.py plugin
+    # has registered is available.
+    cmd = [hlp for hlp in n.rpc.help()['help'] if 'hello' in hlp['command']]
+    assert(len(cmd) == 1)
+
+    # Now try to call it and see what it returns:
+    greet = n.rpc.hello(name='Sun')
```
renepickhardt: meta: how will developers that use the python RPC interface know that there is a new RPC-call available? Even if plugin developers would inharite from the LightningRPC class this would not help as this derived class is not shipped via pip3. 

I believe the best way to handle this is to change the pylightning.py file in a way so that the available RPC calls provide methods dynamically (maybe via reflection?)

lightningd/plugin.c
```diff
@@ -397,12 +412,67 @@ static void plugin_rpcmethod_destroy(struct json_command *cmd,
...
+			if (streq(request->method, plugin->methods[i])) {
+				request->plugin = plugin;
+				break;
```
rustyrussell: This break is a bit suboptimal, since it doesn't break out of the outer loop.

Cleanest is extract this code into 'find_plugin_for_cmd()', and have this `return plugin;`.

lightningd/plugin.c
```diff
@@ -397,12 +412,67 @@ static void plugin_rpcmethod_destroy(struct json_command *cmd,
...
+			}
+		}
+	}
+
+	/* This should never happen, it'd mean that a plugin didn't
+	 * cleanup after dying */
+	assert(request->plugin);
+
+	tal_steal(request->plugin, request);
```
rustyrussell: I wonder what happens if plugin dies while this request is still pending?

lightningd/plugin.c
```diff
@@ -415,10 +416,33 @@ static void plugin_rpcmethod_destroy(struct json_command *cmd,
...
+
+	/* If this is a raw string result we'd be cropping the quotes from the
+	 * result, so add them back */
+	if (req->resulttok->type == JSMN_STRING)
+		json_add_member(response, NULL, "\"%.*s\"",
+				res->end - res->start,
+				req->response + res->start);
+	else
+		json_add_member(response, NULL, "%.*s", res->end - res->start,
+				req->response + res->start);
```
rustyrussell: BTW, there are `json_tok_contents` and `json_tok_len` for exactly this "strings are special" case:

```
     json_add_member(response, NULL, "%.*s", json_tok_len(t), json_tok_contents(t));
```

tests/test_plugin.py
rustyrussell: Unknown rpc calls are handled in pylightning, but it's a bit obscure IMHO. You need to call it using a dict of options, not inline, IIRC....

lightningd/plugin.c
cdecker: The plugin should get freed, cleaning up the connections and all pending requests. I don't currently plan on restarting failed plugins.

lightningd/jsonrpc.c
```diff
@@ -330,17 +341,17 @@ static void json_help(struct command *cmd,
...
-	struct json_command **cmdlist = get_cmdlist();
+	struct json_command **commands = rpc->commands;
 
-	/* cmdlist[i]->name can be NULL in test code. */
-	for (i = 0; i < num_cmdlist; i++)
-		if (cmdlist[i]->name
-		    && json_tok_streq(buffer, tok, cmdlist[i]->name))
-			return cmdlist[i];
+	/* commands[i]->name can be NULL in test code. */
+	for (size_t i=0; i<tal_count(commands); i++)
```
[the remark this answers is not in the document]
cdecker: Well they can now be null because we destroyed a plugin :-) So maybe we need to amend the comment.

lightningd/plugin.c
cdecker: I think it's fair to push the "multiplexing" of multiple plugins wanting to register the same method out to a "dispatch plugin", i.e., a plugin that knows how to route them, and not handle them internally.

tests/test_plugin.py
cdecker: We can certainly make pylightning a bit smarter to accept either all list arguments or all keyword arguments. I'll note that for the next cleanup PR.

lightningd/jsonrpc.c
```diff
@@ -713,18 +724,50 @@ static struct io_plan *incoming_jcon_connected(struct io_conn *conn,
...
+	size_t count = tal_count(rpc->commands);
+
+	/* Check that we don't clobber a method */
+	for (size_t i = 0; i < count; i++)
+		if (rpc->commands[i] != NULL &&
+		    streq(rpc->commands[i]->name, command->name))
+			return false;
+
+	tal_resize(&rpc->commands, count + 1);
+	rpc->commands[count] = command;
```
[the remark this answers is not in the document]
cdecker: That is one weird syntax imho :wink: 

Thanks for pointing that out, I was wondering how to call it. Fixup incoming.

lightningd/jsonrpc.c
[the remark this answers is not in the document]
cdecker: The comment has been amended in a later commit to reflect this.

lightningd/plugin.c
```diff
@@ -382,6 +383,101 @@ static bool plugin_opts_add(const struct plugin_request *req)
...
+	nametok = json_get_member(buffer, meth, "name");
+	desctok = json_get_member(buffer, meth, "description");
+	longdesctok = json_get_member(buffer, meth, "long_description");
+
+	if (!nametok || nametok->type != JSMN_STRING) {
+		plugin_kill(
+		    plugin,
+		    tal_fmt(plugin,
+			    "rpcmethod does not have a string \"name\": %.*s",
+			    meth->end - meth->start, buffer + meth->start));
```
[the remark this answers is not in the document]
cdecker: Added a new commit to the end.

lightningd/plugin.c
```diff
@@ -382,6 +383,101 @@ static bool plugin_opts_add(const struct plugin_request *req)
...
+
+	if (methods->type != JSMN_ARRAY) {
+		plugin_kill(req->plugin, "\"result.rpcmethods\" is not an array");
+		return false;
+	}
+
+	for (cur = methods + 1; cur->parent == methpos; cur = json_next(cur))
+		ok &= plugin_rpcmethod_add(req->plugin, buffer, cur);
+
+	return ok;
```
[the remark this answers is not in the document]
cdecker: Nice catch, I add a fixup that exits the loop on the first error, logs a message about the cause and kills the plugin instead of keeping it limping along.

lightningd/plugin.c
cdecker: Added fixup that uses a `goto` instead.

lightningd/plugin.c
cdecker: Thanks, will do a pass and replace where necessary.

cdecker: Turns out this is the only case this is necessary, so I added a fixup :+1:

cdecker:
I think I addressed all the current feedback. With the `notleak` fix for `json_command`s registered by the plugins postponed to a cleanup PR later since that'd be a new commit on top anyway.

The fixups don't apply cleanly (there are 2 that touch code that is adjacent to a later commit which git can't handle automatically), so I'm happy to rebase & squash once everybody has given the all-clear on the fixups.

doc/plugins.md
```diff
@@ -82,7 +82,15 @@ currently only string options are supported.*
...
-through verbatim.
+through verbatim. Notice that the `name` and the `description` fields
+are mandatory, while the `long_description` can be omitted (it'll be
+set to `description` if it was not provided).
+
+Plugins are free to register any `name` for their `rpcmethod` as long
+as the name was not previously registered. This includes both built-in
+methods, such as `help` and `getinfo`, as well as methods registered
+by other plugins. If there is a conflict then `lightningd` will report
+an error and exit.
```
niftynei: :+1:

lightningd/plugin.c
```diff
@@ -492,11 +492,12 @@ static void plugin_rpcmethod_dispatch(struct command *cmd, const char *buffer,
...
-				break;
+				goto found;
 			}
 		}
 	}
 
+found:
 	/* This should never happen, it'd mean that a plugin didn't
 	 * cleanup after dying */
 	assert(request->plugin);
```
rustyrussell: Hmm, prefer found: below this, and make the assert() and abort().

rustyrussell:
GH decided to "hide" two of my comments (maybe I commented on code which was replaced in a later patch?).  Quoting here: in particular, this means there's no dangling NULLs for plugins:

Let's add some code to utils.h for this case... ***UNTESTED CODE FOLLOWS***

```
#define tal_arr_remove(p, n) tal_arr_remove_((p), sizeof(**p), (n))

void tal_arr_remove_(void *p, size_t elemsize, size_t n)
{
    // p is a pointer-to-pointer for tal_resize.
    char *objp = *(char **)p;
    size_t len = tal_bytelen(objp);
    assert(len % elemsize == 0);
    assert((n + 1) * elemsize <= len);
    memmove(objp + elemsize * n, objp + elemsize * (n+1),
                      len - (elemsize * (n+1)));
    tal_resize((char **)p, len - elemsize);
}
```

cdecker:
I had seen that comment, and then forgot about it :worried: 

Since we now have a conflict and the series doesn't squash cleanly I'll rebase and squash, before adding new fixups on top.

rustyrussell:
Ack 01e555a9a360c36ad40770c606859aadda8da677

PR #2123, opened 2018-11-29
rpc: add check command

wythe:
The `check` command will check the parameters of a command without running it; e.g.:

```
#cli/lightning-cli check invoice ant hello description
{ "code" : -32602, "message" : "'msatoshi' should be a positive number or 'any', not 'ant'" }

# cli/lightning-cli check invoice any hello description
{
  "command": "invoice", 
  "parameters": "ok"
}
```
And yes, you can check the `check` command:
```
# cli/lightning-cli check check
{ "code" : -32602, "message" : "missing required parameter: 'command_to_check'" }
```

The first commit might be a bit overkill but it made for easy implementation.

While playing with this I thought it would be nice if the error case would present the usage string for 
the command also.  But that can be saved for later.


common/json.c
```diff
@@ -216,3 +216,54 @@ jsmntok_t *json_parse_input(const char *input, int len, bool *valid)
 
 	return toks;
 }
+
+void json_tok_print(const char *buffer, const jsmntok_t *params)
+{
+	const jsmntok_t *first = params;
+	const jsmntok_t *last = json_next(params);
+	printf("size: %d, count: %ld\n", params->size, last - first);
```
rustyrussell: "count: %td" (ptrdiff_t) BTW.

common/json.c
```diff
@@ -216,3 +216,54 @@ jsmntok_t *json_parse_input(const char *input, int len, bool *valid)
...
+	while (first != last) {
+		printf("%ld. %.*s, %d\n", first - params,
+		        first->end - first->start, buffer + first->start,
```
rustyrussell: "%td" again I think for first param.  We should add a `jsmntype_to_string` for the type printing too, which will be prettier :)

common/json.c
```diff
@@ -216,3 +216,54 @@ jsmntok_t *json_parse_input(const char *input, int len, bool *valid)
...
+
+	if (!tok)
+		return NULL;
+	jsmntok_t *arr = tal_arr(ctx, jsmntok_t, last - first);
+	jsmntok_t *dest = arr;
+	while (first != last)
+		*dest++ = *first++;
+
+	return arr;
+}
```
rustyrussell: I think this is *almost* a 1-liner?

```
   if (!tok)
        return NULL;
   return tal_dup_arr(ctx, jsmntok_t, tok, json_next(tok) - tok, 0);
```

common/json.c
```diff
@@ -216,3 +216,54 @@ jsmntok_t *json_parse_input(const char *input, int len, bool *valid)
...
+
+void json_tok_remove(jsmntok_t **tokens, jsmntok_t *tok, size_t num)
+{
+#if DEVELOPER
+	assert(*tokens);
+	assert((*tokens)->type == JSMN_ARRAY || (*tokens)->type == JSMN_OBJECT);
+#endif
```
rustyrussell: i would assert these unconditionally, TBH.  We're in for worse (and more confusing!) hurt if they're not true...

common/test/run-json_remove.c
```diff
@@ -0,0 +1,155 @@
...
+};
+
+static void convert_quotes(char *first)
+{
+	while (*first != '\0') {
+		if (*first == '\'')
+			*first = '"';
+		first++;
+	}
+}
```
rustyrussell: Cute trick :)

lightningd/jsonrpc.c
```diff
@@ -884,3 +885,71 @@ bool json_tok_wtx(struct wallet_tx * tx, const char * buffer,
...
+
+	if (cmd->mode == CMD_USAGE) {
+		mod_params = NULL;
+	} else {
+		mod_params = json_tok_copy(cmd, params);
+		cmd->allow_unused = true;
+	}
+
+	if (!param(cmd, buffer, mod_params,
+		   p_req("command_to_check", json_tok_command, &name_tok),
```
rustyrussell: I would remove "_to_check" here, since it makes the JSON API neater (the context is pretty clear).

lightningd/param.c
```diff
@@ -281,5 +281,5 @@ bool param(struct command *cmd, const char *buffer,
 		return false;
 	}
 
-	return param_arr(cmd, buffer, tokens, params);
+	return param_arr(cmd, buffer, tokens, params) && cmd->mode == CMD_NORMAL;
```
rustyrussell: This probably deserves a comment, like "Make caller return immediately if CMD_CHECK"?

lightningd/jsonrpc.c
rustyrussell: ... or just call it 'check'?

lightningd/param.c
```diff
@@ -79,7 +79,7 @@ static bool parse_by_position(struct command *cmd,
 	}
 
 	/* check for unexpected trailing params */
-	if (tok != end) {
+	if (!cmd->allow_unused && tok != end) {
 		command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 			     "too many parameters:"
 			     " got %u, expected %zu",
```
rustyrussell: Can we modify parse_by_name similarly?

Currently:
```
$ ./cli/lightning-cli check command_to_check=listpeers id=xxx
{ "code" : -32602, "message" : "unknown parameter: 'id'" }
```

lightningd/jsonrpc.c
ZmnSCPxj: `command` seems better to me.

wythe: I discussed my reason for choosing such an awkward name in the commit message for https://github.com/ElementsProject/lightning/pull/2123/commits/e45dd7f7c8170526c177a247700979442c599cd6. Basically, it has to be unique to all other parameter names for all commands. But I'm fine with changing it to whatever you are comfortable with. I can also add a developer mode check to make sure it is not used again. Note @ZmnSCPxj, we already are using `command` for the help command.

common/json.c
wythe: I took this approach at first, but it turns out the `tok` is not a tal allocated array.  It's an offset into a tal allocated array. Its setup in  `parse_request` (jsonrpc.c:506).

lightningd/jsonrpc.c
rustyrussell: Note that you *can't* make it unique, since 'check check' is supposed to work :)  But clashing with some other command would be painful, so happy to stop bikeshedding now :)

lightningd/param.c
wythe: oops.

rustyrussell:
OK, so I've got a few of additions if you want them in https://github.com/rustyrussell/lightning/commits/pr-2123 but they can be a separate PR and I don't want to hold this one up any longer!

common/json.c
wythe: Ok you're right.  I tried a getting the tal_length of tok and it didn't work, so I bailed out. My bad.

cdecker:
Rebased on top of `master`, checked locally.

Will merge once CI is happy.

ACK https://github.com/ElementsProject/lightning/pull/2123/commits/8a361b47a64c59ada4ba565d02a03b91ed32ad2c

rustyrussell:
> Your additions look great. I will make a new PR for them (unless you want to). I also started on the man page.

You steal them :)

Thanks!

PR #2126, opened 2018-12-02
cleanups and preparation work for option_simplified_commitment

rustyrussell:
It's going to be a while before we actually add support (we need some significant wallet logic rework it seems), but these cleanups are all standalone.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2126/commits/617413729d3b5b7cce6b97425b51d2273861ac51

PR #2137, opened 2018-12-04
pylightning: Bump version number to 0.0.5 to make Pypi work again

### LINKED ISSUE #2135: JSON rpc interface broke after upgrading to the plugin system
after upgrading to the plugin system which supports JSON API calls the pylightning json rpc interface breaks with a call like `rpc_interface.getinfo()`
the main reason seems that the response is shorter than 1024 Bytes and the socket does not close anymore. 

c.f.: 
```
  File "test.py", line 9, in <module>
    res = rpc_interface.getinfo()
  File "/Users/rpickhardt/anaconda3/lib/python3.6/site-packages/lightning/lightning.py", line 275, in getinfo
    return self.call("getinfo")
  File "/Users/rpickhardt/anaconda3/lib/python3.6/site-packages/lightning/lightning.py", line 75, in call
    resp = self._readobj(sock)
  File "/Users/rpickhardt/anaconda3/lib/python3.6/site-packages/lightning/lightning.py", line 32, in _readobj
    b = sock.recv(1024)
```

I was able to quickfix this by changing https://github.com/ElementsProject/lightning/blob/master/contrib/pylightning/lightning/lightning.py#L34
to `b = sock.recv(1024)``

I do not provide this patch as a pull request because I want to look at the changes to the server side. I realized that https://github.com/ElementsProject/lightning/blob/master/lightningd/lightningd.c#L705 is new. Before there was `setup_jsonrpc(ld, ld->rpc_filename);` instead of `jsonrpc_listen(ld->jsonrpc, ld);` so I guess the way the socket is handled within c-lightning has changed and breaks the RPC interface.

on a sidenote:
Shouldn't it be possible from a plugin to send json messages directly to `lightningd` without the necessity to specify the rpc-interface? Generally the plugin does not know where the data directory of the lightning node that calls the plugin is placed and would not know where the RPC interface is placed. since the plugin already talks to lightningd over stdout and stdin I believe there should be a way of making regular rpc-calls

cdecker:
Fixes #2135

### DIFF
--- a/contrib/pylightning/setup.py
+++ b/contrib/pylightning/setup.py
@@ -1,7 +1,7 @@
 from setuptools import setup
 
 setup(name='pylightning',
-      version='0.0.3',
+      version='0.0.5',
       description='Client library for lightningd',
       url='http://github.com/ElementsProject/lightning',
       author='Christian Decker',


cdecker:
Turns out I bumped the version some time on August to 0.0.4 without committing that bump to the repo, this skips 0.0.4 and goes straight for 0.0.5

rustyrussell:
We should also revert b9b7411d88073fc5a20103af882de3d96242d3e9 until 0.6.4, so everyone has time to upgrade.

PR #2138, opened 2018-12-05
pylightning compatibility mode for older versions

### LINKED ISSUE #2135: JSON rpc interface broke after upgrading to the plugin system

cdecker:
A bit more work to add a backward compatible mode and upgrade to the new `\n\n`-delimiter only if a request is read that indicates support. This still uses the slow method for the first request (which hopefully is something small like `getinfo` anyway) and upgrades if that request has the new delimiter.

This has been tested to be compatible with v0.6 as well, but I'd like a confirmation from @renepickhardt and @conscott as well :wink: 

Supersedes #2137 
Fixes #2135


cdecker:
Notice that I didn't push this to pypi just yet, so you'll have to clone the repo and do `python setup.py develop` or `python setup.py install` to get this version tested in your project @renepickhardt

renepickhardt:
I am currently working on creating a plugin not sure if I manage to test today. latest by tomorrow

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ changes.
 - JSON API: commands are once again read even if one hasn't responded yet (broken in 0.6.2).
 - Protocol: allow lnd to send `update_fee` before `funding_locked`.
 - pylightning: handle multiple simultanous RPC replies reliably.
+- pylightning: fixed an incompatibility when parsing responses from older clients (#2135).
```
rustyrussell: No, this was never broken in a release!  b2378654d70aea2a5b456f948eb67104dd50a5fe introduced it *after* 0.6.2.

cdecker: Ok, will amend to say it supports both :-)

cdecker:
`CHANGELOG.md` amended to spell out the double-newline separator and hint at a future incompatibility if the library or `lightningd` don't get updated.

conscott:
ACK 0d639343b8f71c2cc6c219297b20295269a81f7c

Thanks for the quick response

rustyrussell:
Moved the line about deprecation to the "Deprecated" section of the CHANGELOG.md...

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2138/commits/4638fe559ae588e7b4a0d814c6f6b3e21821ecb3

PR #2131, opened 2018-12-03
plugin: Some cleanups that didn't make it into part 2 of the plugin saga

cdecker:
- Changelog entry
 - `tal_arr_remove`
 - Timeout for `getmanifest` call
 - Double free issue when killing a plugin.

Issue #2104, opened 2018-11-23
listfunds does not contains unconfirmed transactions


asocjal:
### Issue and Steps to Reproduce
1. Create new address by "newaddr" command
2. Send some funds to the newest created address
3. listfunds does not contains this transaction as long, as it is unconfirmed

### `getinfo` output
`{
  "id": "035a71a8d2e36f5ff894fd6f97225a493e0cdf013c60c3b1c6992be19bd901a9ca", 
  "alias": "cln-tst-01", 
  "color": "035a71", 
  "address": [
    {
      "type": "ipv4", 
      "address": "176.107.130.62", 
      "port": 9735
    }
  ], 
  "binding": [
    {
      "type": "ipv6", 
      "address": "::", 
      "port": 9735
    }, 
    {
      "type": "ipv4", 
      "address": "0.0.0.0", 
      "port": 9735
    }
  ], 
  "version": "v0.6.1", 
  "blockheight": 551232, 
  "network": "bitcoin"
}`

kristapsk:
I would also like to have `listfunds` to output unconfirmed incoming blockchain transactions too.

Probably some additional parameter to specify minimum number of confirmations for listing would be nice to have (like Bitcoin's `listunspent` have). It could default to 1 (not 0), so the current default behaviour would not change, also 1 is default for minconf for `listunspent` too.

kristapsk:
Ahh, but `listfunds` only shows `"status": "confirmed"`, so either that proposed parameter in my comment above should be confirmed/unconfirmed boolean of `listfunds` should also show the number of confirmations, for the sake of consistency.

asocjal:
I pray for give high priority for this feature, as this is the only obstacle that prevent us to finish clightning-based wallet.

conscott:
After investigating, this is my understanding... 

It looks like, at the moment c-lightning only polls for new block searching for transactions relevant to the addresses it cares about, which explains why you would _normally_ only ever see a "confirmed" state (it's already in a block). Note that in case of a re-org it can be 'undone' and the txid can roll back to "unconfirmed" 

[...]

PR #1888, opened 2018-08-27
first version of lib-autopilot.py

renepickhardt:
As mentioned in the c-lightning list: https://lists.ozlabs.org/pipermail/c-lightning/2018-August/000065.html you wellcome plugins which are not written in c.

 So I took the liberty to provide a first sketch for an autopilot for c-lightning written an python. In the beginning of the year I have stated on lnd's autopilot that the Barabasi Albert model was probably not the best choice (c.f.: https://github.com/lightningnetwork/lnd/issues/677 ) This is why my lib can currently generate candidates for an autopilot to open channels with according to four strategies (and it combines those currently):

* Random: following the Erdoes Renyi model nodes are drawn from a uniform distribution
* Central: nodes are sampled from a uniform distribution of the top most central nodes (betweenness)
* Network_Improovement: nodes are sampled from a uniform distribution of the nodes which are badly connected
* richness: nodes with high liquidity are taken and it is sampled from a uniform distribution of those

this library is work in progress. there is no logic yet implemented on how many funds to allocate to which channel and actually opening the channels (though I think that this should be the easy part) Also due to the randomness the lib might sometimes find less channels than the user asked for. Also many conner cases will let the code break.

Therefor the PR is more for code review and suggestions.

Running the code produces this output: 
```
2018-08-27 16:06:19,285 - lib-autopilot - INFO - Try to load graph from file system at: data/networkx_graph
2018-08-27 16:06:19,306 - lib-autopilot - INFO - Successfully restored the lightning network graph from data/networkx_graph
2018-08-27 16:06:19,306 - lib-autopilot - INFO - running the autopilot on a graph with 1528 nodes and 7997 edges.
2018-08-27 16:06:19,306 - lib-autopilot - INFO - GENERATE CANDIDATES: Try to generate up to {} nodes with 4 strategies: (random, central, network Improvement, liquidity)
2018-08-27 16:06:19,306 - lib-autopilot - INFO - RANDOM_NODES: try to generate a set of 6 nodes sampled with uniform distribution
2018-08-27 16:06:19,306 - lib-autopilot - INFO - CENTRAL_NODES: Try to seek 6 nodes which are currently central
2018-08-27 16:06:32,083 - lib-autopilot - INFO - CENTRAL_NODES: Generated top 24 central nodes (according to betweeness centrality)
2018-08-27 16:06:32,083 - lib-autopilot - INFO - CENTRAL_NODES: Sample 6 items from the candidates as was requested
[...]

PR #2139, opened 2018-12-05
[plugin] Plugin that depicts the off- and onchain funds in a nicer overview

renepickhardt:
the plugins directory doesn't seem to work yet. I had to call the plugin explicitly with 
`lightningd --plugin=/Users/rpickhardt/hacken/lightning/plugins/funds.py` not sure though weather this plugin should be a standard plugin anyway. 

you can test the plugin with: 
`lightning-cli funds` and the output should be similar to this: 
`{"total_sat": 200000, "onchain_sat": 100000, "offchain_sat": 100000}`
if you call the plugin with the `m` argument the ammounts are depicted in millibitoins (other units are available via help) eg:
```
lightning-cli funds m
{"total_mBTC": 2, "onchain_mBTC": 1, "offchain_mBTC": 1}
```

I will probably use this plugin as a tutorial for next Mondays youtube video, so that people can learn how easy it is to create plugins in c-lightning (chapeau for @cdecker )

I have prepared a couple more plugins (for example for channel balance overview, an invoice server for donations, maybe even the spontaneous payments one) but I would like to have feedback on this "quicky" first instead of pushing several plugins.

Once the easy ones are there I will transfer my autopilot code to the plugin system. 

I have also some feedback where the current system could be improved to ease up the life of developers. Should I do this as a pull request? 

Also is there a better way to talk to the RPC interface (for example directly via stdout) so that the plugin does not need to know where the rpc-interface is located?

### DIFF
--- a/contrib/plugins/funds/funds.py
+++ b/contrib/plugins/funds/funds.py
@@ -10,6 +10,12 @@ Activate the plugin with:
 Call the plugin with: 
 `lightning-cli funds`
 
+The standard unit to depict the funds is set to satoshis. 
+The unit can be changed by and arguments after `lightning-cli funds` 
+for each call. It is also possible to change the standard unit when 
+starting lightningd just pass `--funds_display_unit={unit}` where
+unit can be s for satoshi, b for bits, m for milliBitcoin and B for BTC.
+
 
 Author: Rene Pickhardt (https://ln.rene-pickhardt.de)
 Development of the plugin was sponsored by fulmo.org
@@ -24,48 +30,52 @@ from os.path import join
 rpc_interface = None
 plugin = Plugin(autopatch=True)
 
+unit_aliases = {
+    "bitcoin": "BTC",
+    "btc": "BTC",
+    "satoshi": "sat",
+    "satoshis": "sat",
+    "bit": "bit",
+    "bits": "bit",
+    "milli": "mBTC",
+    "mbtc": "mBTC",
+    "millibtc": "mBTC",
+    "B": "BTC",
+    "s": "sat",
+    "m": "mBTC",
+    "b": "bit",
+}
+
+unit_divisor = {
+    "sat": 1,
+    "bit": 100,
+    "mBTC": 100*1000,
+    "BTC": 100*1000*1000,
+}
+
 
 @plugin.method("funds")
-def funds(plugin, unit="s"):
+def funds(unit=None, plugin=None):
     """Lists the total funds the lightning node owns off- and onchain in {unit}.
 
-{unit} can take the following values:
-s, satoshi, satoshis to depict satoshis
-b, bit, bits to depict bits
-m, milli, btc to depict milliBitcoin
-B, bitcoin, btc to depict Bitcoins
+    {unit} can take the following values:
+    s, satoshi, satoshis to depict satoshis
+    b, bit, bits to depict bits
+    m, milli, btc to depict milliBitcoin
+    B, bitcoin, btc to depict Bitcoins
 
-When not using Satoshis (default) the comma values are rounded off."""
+    When not using Satoshis (default) the comma values are rounded off."""
 
-    switcher = {
-        "bitcoin": "BTC",
-        "btc": "BTC",
-        "satoshi": "sat",
-        "satoshis": "sat",
-        "bit": "bit",
-        "bits": "bit",
-        "milli": "mBTC",
-        "mbtc": "mBTC",
-        "millibtc": "mBTC",
-        "B": "BTC",
-        "s": "sat",
-        "m": "mBTC",
-        "b": "bit",
-    }
+    plugin.log("call with unit: {}".format(unit), level="debug")
+    if unit is None:
+        unit = plugin.get_option("funds_display_unit")
 
     if unit != "B":
-        unit = switcher.get(unit.lower(), "sat")
+        unit = unit_aliases.get(unit.lower(), "sat")
     else:
         unit = "BTC"
 
-    switcher = {
-        "sat": 1,
-        "bit": 100,
-        "mBTC": 100*1000,
-        "BTC": 100*1000*1000,
-    }
-
-    div = switcher.get(unit, 1)
+    div = unit_divisor.get(unit, 1)
 
     funds = rpc_interface.listfunds()
 
@@ -84,17 +94,18 @@ When not using Satoshis (default) the comma values are rounded off."""
 @plugin.method("init")
 def init(options, configuration, plugin):
     global rpc_interface
-    plugin.log("start initialization of the funds plugin")
-    print(configuration)
+    plugin.log("start initialization of the funds plugin", level="debug")
     basedir = configuration['lightning-dir']
     rpc_filename = configuration['rpc-file']
     path = join(basedir, rpc_filename)
     plugin.log("rpc interface located at {}".format(path))
     rpc_interface = LightningRpc(path)
-
     plugin.log("Funds Plugin successfully initialezed")
+    plugin.log("standard unit is set to {}".format(
+        plugin.get_option("funds_display_unit")), level="debug")
 
 
-plugin.add_option('funds', 'funds',
-                  'Lists the total funds the lightning node owns off- and onchina in {units}')
+# set the standard display unit to satoshis
+plugin.add_option('funds_display_unit', 's',
+                  'pass the unit which should be used by default for the simple funds overview plugin')
 plugin.run()
--- a/contrib/plugins/funds/funds.py
+++ b/contrib/plugins/funds/funds.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+""" This plugin gives you a nicer overview of the funds that you own.
+
+Instead of calling listfunds and adding all outputs and channels
+this plugin does that for you.
+
+Author: Rene Pickhardt (https://ln.rene-pickhardt.de)
+
+"""
+
+import json
+import sys
+
+from lightning.lightning import LightningRpc
+from os.path import join
+
+rpc_interface = None
+
+
+def json_get_funds(request, unit="s"):
+
+    switcher = {
+        "bitcoin": "BTC",
+        "btc": "BTC",
+        "satoshi": "sat",
+        "satoshis": "sat",
+        "bit": "bit",
+        "bits": "bit",
+        "milli": "mBTC",
+        "mbtc": "mBTC",
+        "millibtc": "mBTC",
+        "B": "BTC",
+        "s": "sat",
+        "m": "mBTC",
+        "b": "bit",
+    }
+
+    if unit != "B":
+        unit = switcher.get(unit.lower(), "sat")
+    else:
+        unit = "BTC"
+
+    switcher = {
+        "sat": 1,
+        "bit": 100,
+        "mBTC": 100*1000,
+        "BTC": 100*1000*1000,
+    }
+
+    div = switcher.get(unit, 1)
+
+    funds = rpc_interface.listfunds()
+
+    onchain_value = sum([int(x["value"]) for x in funds["outputs"]])
+    offchain_value = sum([int(x["channel_sat"]) for x in funds["channels"]])
+
+    total_funds = onchain_value + offchain_value
+
+    return {
+        'total_'+unit: total_funds//div,
+        'onchain_'+unit: onchain_value//div,
+        'offchain_'+unit: offchain_value//div,
+    }
+
+
+def json_getmanifest(request):
+
+    verbose = """Lists the total funds the lightning node owns off- and onchain in {unit}.
+
+{unit} can take the following values:
+s, satoshi, satoshis to depict satoshis
+b, bit, bits to depict bits
+m, milli, btc to depict milliBitcoin
+B, bitcoin, btc to depict Bitcoins
+
+When not using Satoshis (default) the comma values are rounded off."""
+
+    return {
+        "options": [
+        ],
+        "rpcmethods": [
+            {
+                "name": "funds",
+                "description": "Lists the total funds the lightning node owns off- and onchain in {unit}",
+                "long_description": verbose
+            },
+        ]
+    }
+
+
+def json_init(request, options, configuration):
+    """The main daemon is telling us the relevant cli options
+    """
+    global rpc_interface
+
+    basedir = request['params']['configuration']['lightning-dir']
+    rpc_filename = request['params']['configuration']['rpc-filename']
+    path = join(basedir, rpc_filename)
+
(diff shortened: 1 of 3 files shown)


plugins/funds.py
```diff
@@ -0,0 +1,139 @@
...
+
+"""
+
+import json
+import sys
+
+from lightning.lightning import LightningRpc
+from os.path import expanduser
+
+path = expanduser("~/.lightning/lightning-rpc")
```
renepickhardt: I was talking explicitly about this line. My feeling is that I should be able to send RPC calls via json directly to the lightningdaemon. However I don't know the format of the json and it is not documented yet.

cdecker: #2131 will contain a `configuration` field in the `configure` call that'll explicitly tell you were to connect :wink:

plugins/funds.py
```diff
@@ -0,0 +1,139 @@
...
+
+rpc_interface = LightningRpc(path)
+
+
+def json_get_funds(request, unit="s"):
+    if unit.lower() == "bitcoin" or unit.lower() == "btc":
```
cdecker: How about creating a dict that maps the long form to the short form? If you do the reverse step, you can also create the inverse map with a simple dict comprehension :-)

plugins/funds.py
```diff
@@ -0,0 +1,139 @@
...
+
+    verbose = """Lists the total funds the lightning node owns off- and onchain in {unit}.
+
+{unit} can take the following values: 
+s, satoshi, satoshis to depict satoshis
+b, bit, bits to depict bits
+m, milli, btc to depict milliBitcoin
+B, bitcoin, btc to depict Bitcoins
+
+When not using Satshis (default) the comma values are rounded off."""
```
cdecker: Typo: Satshis -> Satoshis

plugins/funds.py
renepickhardt: Yeah dict will be more python style

renepickhardt: Thx

plugins/funds.py
renepickhardt: Thx

plugins/funds.py
rustyrussell: #2131 didn't?  I have a patch which adds a 'rpc_filename' to the init msg however, will push (I needed this too, of course)!

renepickhardt: So I will be able to find this in def json_init() in the options argument?

How can I look at these variables? Since stdout is sent to lightningd I can't print variables to inspect them which is annoying

cdecker: Wait, where did that patch end up?

plugins/funds.py
```diff
@@ -0,0 +1,139 @@
...
+        "B": "BTC",
+    }
+    unit = switcher.get(unit, "sat")
+
+    info = rpc_interface.getinfo()
+    funds = rpc_interface.listfunds()
+    outputs = funds["outputs"]
+    onchain_value = 0
+    for output in outputs:
+        onchain_value += int(output["value"])
```
conscott: Nothing wrong with this code, but fyi, you can also use list comprehensions to shorten this down a bit to something like:
```
onchain_value = sum([int(output['value']) for output in funds['outputs']])
offchain_value = sum([int(channel["channel_sat"]) for channel in funds['channels']])
```

conscott:
Cool! There should be lots of nice simple plugins that can be added, and it's nice to have a reference for a 'standard' plugin 

Would also be nice to add unconfirmed balance later if #2104 is implemented, but I understand this may just be a sample. 

As a general question to @cdecker and @rustyrussell .... Is there a general idea or documentation in mind for some 'standard' plugins to be released with c-lightning? 

Something like `closedchannels` (close all channels) command in `lnd` seems better implemented as a simple plugin command in c-lightning rather than base rpc, just based on how cmd life-cycles are implemented in lightningd. Just didn't know if there was any guidance on what would or would not be considered for inclusion into a standard set. 

Sorry to hijack the comments @renepickhardt - Looking forward to seeing the autopilot revived :)

plugins/funds.py
renepickhardt: yeah I will do that since I have to touch the code once more anyway (:

renepickhardt:
I rebased the master and have made the requested changes and also merged #2141 in this branch as this plugin relies on the fact that the deamon is giving us the location of the RPC interface in the init message. I thought this is the best way to do so otherwise we would have to touch this plugin again once #2141 is merged

renepickhardt:
@cdecker @rustyrussell I have force pushed the code once more to have the plugin in the required location and I have also taken back my early merge of #2141 . I guess this plugin should now be ready to be merged (: On the way I have created a more complex (in the sense of more third party technologies that have been used) plugin over at #2152 . 

Once these two are merged I guess I will be able to transfer the code of #1888 to the plugin architecture. Should the autopilot be placed in `/plugins` or in `/contrib/plugins/autopilot` ? My understanding is that the autopilot is a rather standard / close feature (in particular it should help users who are less technical so activating it as a sperate plugin seems unncessary burden)

renepickhardt:
I have yet updated the plugin to the new python bindings which makes use of the decorators so that there is no more boilerplate code left in the plugin code. 

If everything is fine now and the plugin can be merged I would love to create the educational video about the plugin system soon (:

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+    rpc_filename = configuration['rpc-file']
+    path = join(basedir, rpc_filename)
+    plugin.log("rpc interface located at {}".format(path))
+    rpc_interface = LightningRpc(path)
+
+    plugin.log("Funds Plugin successfully initialezed")
+
+
+plugin.add_option('funds', 'funds',
+                  'Lists the total funds the lightning node owns off- and onchina in {units}')
```
cdecker: `onchina` -> `onchain`

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+    basedir = configuration['lightning-dir']
+    rpc_filename = configuration['rpc-file']
+    path = join(basedir, rpc_filename)
+    plugin.log("rpc interface located at {}".format(path))
+    rpc_interface = LightningRpc(path)
+
+    plugin.log("Funds Plugin successfully initialezed")
+
+
+plugin.add_option('funds', 'funds',
```
cdecker: This option doesn't really seem to do anything. From the description I take it that it's supposed to be the `@plugin.method` call above.

cdecker: What you could do instead is provide the default unit here, so that on startup we can chose whether we'd like to display in satoshis, bits, mBTC or BTC by default.

Another option that would make sense is the number of decimal places to show.

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+        'onchain_'+unit: onchain_value//div,
+        'offchain_'+unit: offchain_value//div,
+    }
+
+
+@plugin.method("init")
+def init(options, configuration, plugin):
+    global rpc_interface
+    plugin.log("start initialization of the funds plugin")
+    print(configuration)
```
cdecker: This looks like a debug statement.

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+
+    onchain_value = sum([int(x["value"]) for x in funds["outputs"]])
+    offchain_value = sum([int(x["channel_sat"]) for x in funds["channels"]])
+
+    total_funds = onchain_value + offchain_value
+
+    return {
+        'total_'+unit: total_funds//div,
+        'onchain_'+unit: onchain_value//div,
+        'offchain_'+unit: offchain_value//div,
```
cdecker: Using [`round`](https://docs.python.org/2/library/functions.html#round) or [`ceil`](https://docs.python.org/2/library/math.html#math.ceil) is almost always preferable here.

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+def funds(plugin, unit="s"):
+    """Lists the total funds the lightning node owns off- and onchain in {unit}.
+
+{unit} can take the following values:
+s, satoshi, satoshis to depict satoshis
+b, bit, bits to depict bits
+m, milli, btc to depict milliBitcoin
+B, bitcoin, btc to depict Bitcoins
+
+When not using Satoshis (default) the comma values are rounded off."""
```
cdecker: I tend to indent multi-line comments as well, otherwise they get hard to read. I'm surprised `flake8` isn't complaining about these actually.

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+
+    switcher = {
```
cdecker: Minor optimization and readability improvement: these `dict`s can be instantiated globally, since they don't change at all, and it'd safe us from instantiating them on every call.

Furthermore since they serve different purposes they ought to have different names as well. Maybe call this `unit_aliases` and the one below `unit_divisors`?

contrib/plugins/funds/funds.py
```diff
@@ -0,0 +1,100 @@
...
+        "milli": "mBTC",
+        "mbtc": "mBTC",
+        "millibtc": "mBTC",
+        "B": "BTC",
+        "s": "sat",
+        "m": "mBTC",
+        "b": "bit",
+    }
+
+    if unit != "B":
```
cdecker: Why is `B` being special-cased here? imho it'd still work using the above mapping.

contrib/plugins/funds/funds.py
renepickhardt: Maybe I did not understand the `add_option()` call correctly. I had the feeling this was similar to the entry in the `get_manifest` that I had to be used previously. I guess I got that wrong. Will take a closer look at the implementation tomorrow

contrib/plugins/funds/funds.py
renepickhardt: true. However I kept in in out purpose. to show that stuff like this is possible in the video. I guess I could also set the log level to debug? I think I saw this somewhere in the code. I will search for it tomorrow unless you provide me with the method to do it

contrib/plugins/funds/funds.py
renepickhardt: why / what's the difference?

contrib/plugins/funds/funds.py
renepickhardt: will fix

contrib/plugins/funds/funds.py
renepickhardt: I had a to lower of the argument before checking the switch statement. however `B` is for BTC and `b` for bits. I fully agree it is ugly but I did not have any better idea and I thought about this several times for more than a couple of minutes. Totally open to a better suggestion here

renepickhardt:
Will fix those changes. I have uploaded the current version (before I saw your feedback) and with regard to the conversation in the irc to https://github.com/renepickhardt/c-lightning-plugin-collection (where I would also publish the donation plugin and the autopilot) 
(though I would recommend to put the autopilot to the default plugin dir of plugins which are loaded automatically)

contrib/plugins/funds/funds.py
cdecker: Because with round you can specify the number of decimal places that you'd like to display, which is more flexible imho

contrib/plugins/funds/funds.py
cdecker: `add_option` will add command line options to `lightningd` when starting the daemon and the option values will be passed to the plugin on `init`. So this declaration would result in a command line option `funds` with default value `funds` for the `lightningd`. So this would be perfectly fine:

```
lightningd --funds=bla
```

But then you don't extract and store the value of that option, so it's not being used at all :wink:

contrib/plugins/funds/funds.py
cdecker: I think you are referring to `stderr` vs `stdout`, I usually give `stderr` output a higher priority, yes :-)

contrib/plugins/funds/funds.py
cdecker: Oh I see, sorry for missing that. I can't think of an elegant way, maybe something will pop up :-)

cdecker:
> (though I would recommend to put the autopilot to the default plugin dir of plugins which are loaded automatically)

Absolutely, my plan is to externally curate a list somewhere and pull plugins in as they grow in popularity. Maybe some sort of plugin manager might be in order. But I'll definitely bundle things like the autopilot at least during release.

contrib/plugins/funds/funds.py
renepickhardt: Damn it I already knew that. But with the frequent api changes I oversaw that. Well certainly fix that (:

contrib/plugins/funds/funds.py
renepickhardt: don't want to doubt you. But I doubt it. I was very unhappy about this and thought about it several times :( but maybe it is just ok (:

contrib/plugins/funds/funds.py
renepickhardt: also I am just happy that I made this mistake. I will try to change the code so that I will add an argument -`--fundsunit=s` to lightningd which would set the standard value for the command line tool. this is also good for educational purpose to see the difference and both methods

contrib/plugins/funds/funds.py
renepickhardt: fixed the bigger issue though is that the line breaks are removed when the doc string is used for the plugin help. will look if I will be able to fix this and make a PR. Will not open an issue for this.

renepickhardt:
I have changed the requested changes (besides replacing `//` with `round`) and extended the API of `lightningd` to be able to set a default unit.
 
As a side effect I have opened #2217 which still seems like a pitty to me.

conscott:
The following has just occurred in playing with another balance plugin:

`listfunds` output will show channels in `ONCHAIN` state (closed) in the `channels` list (I guess until the channel is forgotten) while ALSO showing the closing tx transaction in the `outputs` list. 

For this reason, I think the plugin will 'count' a closed channel into the "offchain" (channel) balance, as well as the "onchain" balance at the same time. The correct behavior would be to not count a closed channel as part of the channel balance. 

If calculating from `listpeers` you can know the channel state and can avoid this condition, but it also brings up the question... 

Should `listfunds` even list closed channels, given that the closing txid is already listed in the `outputs` field? @cdecker 

If it's still desired to list it, perhaps it makes sense to add the channel state so people have to the context to know its closed / closing.

cdecker:
Since we now have quite a few plugins that will address this I think I'll close this :wink:

PR #2140, opened 2018-12-05
Manpage for lightning-listpeers created

runcrypto-michael:
Continuing work on https://github.com/ElementsProject/lightning/issues/1360
lightning-listpeers.7.txt created in asciidoc format and lightning-listpeers.7 generated using a2x.

Appears to be duplicated fields returned for peers (global_features/globalfeatures and local_features/localfeatures), not sure of the difference of these, if there is any?


ZmnSCPxj:
The fields with underscores `local_features` and `global_features` were used in older versions, and are retained for back-compatibility; they need not be mentioned in current manpage since they will disappear "soon".

doc/lightning-listpeers.7.txt
```diff
@@ -0,0 +1,63 @@
...
+peer objects.
+
+Each peer object contains the following data:
+
+- 'id' : The unique id of the peer
+- 'connected' : A boolean value showing the connection status
+- 'netaddr' : An array of network addresses the node is listening on
+- 'globalfeatures' : Bit flags showing supported global features (BOLT #9)
+- 'localfeatures' : Bit flags showing supported local features (BOLT #9) 
+- 'global_features' : Duplicate of globalfeatures
```
ZmnSCPxj: This field is deprecated; no need to mention it in the manpage.  Same for `local_features`.

rustyrussell:
Yes: you can run lightningd with '--allow-deprecated-apis=false' and they won't appear at all!  That's how you should always run it for testing stuff, since deprecated APIs will be removed in future...

runcrypto-michael:
Thanks for the info, let me know if there's anything else I can do to improve the commits as this is the first project I've contributed to!

runcrypto-michael:
I think there should also be something about when peers are returned but "connected" is false.  I've only just come across this but from what I understand, this happens when a channel has been created and a node goes offline without closing the channel first?
Is this resolvable and should handling it be covered in the man page?

doc/lightning-listpeers.7.txt
```diff
@@ -0,0 +1,63 @@
...
+'id' that was used with the *connect* command.
+
+Supplying 'id' will return only data on a connected lightning node with a
+matching 'id', if one exists.
+
+If no 'id' is supplied, then data on all connected lightning nodes is returned.
+
+RETURN VALUE
+------------
+On success, a "peers" object is returned, containing an array of 0 or more
```
cdecker: Technically it's an object with a `peers` key that is a  list of objects.

doc/lightning-listpeers.7.txt
```diff
@@ -0,0 +1,63 @@
...
+----
+lightning-listpeers - Command for returning data on connected lightning nodes.
+
+SYNOPSIS
+--------
+*listpeers* ['id']
+
+DESCRIPTION
+-----------
+The *listpeers* RPC command returns data on nodes in the Lightning Network
```
cdecker: "Lightning Network" seems a bit redundant here

doc/lightning-listpeers.7.txt
runcrypto-michael: Happy to change this.  I've checked through and made other references consistent (e.g. using list instead of array)

doc/lightning-listpeers.7.txt
runcrypto-michael: Thanks for the clarification, they've been removed

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2140/commits/b87f6562d0ef7410cc7b391faec7b4955f19f852

PR #2141, opened 2018-12-06
plugin: Added .params.configuration to init call

cdecker:
This tells the plugin both the `lightning-dir` as well as the
`rpc-filename` to use to talk to `lightningd`. Prior to this they'd
had to guess.

This was supposed to be in #2131, but I had it stashed and forgot about it :slightly_frowning_face:

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -90,6 +90,7 @@ struct plugins {
 	struct jsonrpc *rpc;
 
 	struct timers timers;
+	struct lightningd *ld;
 };
 
 /* Represents a pending JSON-RPC request that was forwarded to a
@@ -116,7 +117,7 @@ struct plugin_opt {
 };
 
 struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
-			    struct jsonrpc *rpc)
+			    struct jsonrpc *rpc, struct lightningd *ld)
 {
 	struct plugins *p;
 	p = tal(ctx, struct plugins);
@@ -125,6 +126,7 @@ struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
 	p->log = new_log(p, log_book, "plugin-manager");
 	p->rpc = rpc;
 	timers_init(&p->timers, time_mono());
+	p->ld = ld;
 	return p;
 }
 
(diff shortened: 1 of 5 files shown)


renepickhardt:
tested your code with my updated plugin and it worked like a charm. how should I push my code? just the plugin your PR merged in my PR?

renepickhardt:
I have merged this PR into #2139 because the proposed plugin depends on these changes and they turned out to work well in practice. While I am happy with this PR I suggest to close this PR here and merge #2139 in order to avoid merging conflicts on master. Otherwise I can also change my PR with the plugin by just adding the python file.

renepickhardt:
I have merged this PR into #2139 because the proposed plugin depends on these changes and they turned out to work well in practice. While I am happy with this PR I suggest to close this PR here and merge #2139 in order to avoid merging conflicts on master. Otherwise I can also change my PR with the plugin by just adding the python file.

renepickhardt:
I have merged this PR into #2139 because the proposed plugin depends on these changes and they turned out to work well in practice. While I am happy with this PR I suggest to close this PR here and merge #2139 in order to avoid merging conflicts on master. Otherwise I can also change my PR with the plugin by just adding the python file.

lightningd/plugin.c
```diff
@@ -839,6 +842,12 @@ static void plugin_config(struct plugin *plugin)
 	}
 	json_object_end(req->stream); /* end of .params.options */
 
+	/* Add .params.configuration */
+	json_object_start(req->stream, "configuration");
```
rustyrussell: "dir", "params", "configuration".... One of these is not like the other :)

lightningd/plugin.c
```diff
@@ -116,7 +117,7 @@ struct plugin_opt {
 };
 
 struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
-			    struct jsonrpc *rpc)
+			    struct jsonrpc *rpc, struct lightningd *ld)
```
arowser: Maybe should remove log_book and rpc here, and replace by ld->log_book and log->rpc in follow.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2141/commits/091811e4c36e9911b77c9ec77e366d2b89fea745

PR #2144, opened 2018-12-06
Prefer direct routes when paying.

ZmnSCPxj:
Closes: #2168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - JSON API: `pay` and `decodepay` accept and ignore `lightning:` prefixes.
 - pylightning: Allow either keyword arguments or positional arguments.
 - JSON-RPC: messages are now separated by 2 consecutive newlines.
+- JSON API: `pay` now tries a direct channel first if one exists.
 
 ### Deprecated
 
--- a/lightningd/channel.c
+++ b/lightningd/channel.c
@@ -297,6 +297,16 @@ struct channel *active_channel_by_id(struct lightningd *ld,
 	return peer_active_channel(peer);
 }
 
+struct channel *normal_channel_by_id(struct lightningd *ld,
+				     const struct pubkey *id)
+{
+	struct peer *peer = peer_by_id(ld, id);
+	if (!peer)
+		return NULL;
+
+	return peer_normal_channel(peer);
+}
+
 struct channel *channel_by_dbid(struct lightningd *ld, const u64 dbid)
 {
 	struct peer *p;
--- a/lightningd/channel.h
+++ b/lightningd/channel.h
@@ -185,6 +185,9 @@ struct channel *peer_normal_channel(struct peer *peer);
 struct channel *active_channel_by_id(struct lightningd *ld,
 				     const struct pubkey *id,
 				     struct uncommitted_channel **uc);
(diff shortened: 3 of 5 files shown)


renepickhardt:
Oh sorry had this tab open in my browser and oversaw the comment by @rustyrussell

ZmnSCPxj:
> Perhaps our fuzz is too large or default risk is too low and it's overriding the natural bias towards direct channels?

The intent is to try direct channels first because:

1.  They do not leak our payment information to anyone other than the payee (who has to know our payment information anyway).
2.  The direct route is more likely to not stall, especially as we already have stall mitigation, but non-recent-c-lightning intermediate nodes may or may not.
3.  Users are weirded out that we do not use direct channel.

Fuzz parameter is used to give widely divergent random routes to prevent the same third parties (those trying to bias us towards them by greatly reducing their fees  and connecting to as many nodes as they can) from being used over and over.

Thus I think we should have a separate preference for a direct channel if possible, but retain high fuzz.

An alternative is:  Provide local channel balance information and etc to gossipd, and have `getroute` return the direct route if it is viable.  This ties in to other intended future changes to routefinding I describe here for bass amplifier payments: https://lists.ozlabs.org/pipermail/c-lightning/2018-December/000095.html ; routing will have to keep track of assumed channel balance (for non-local channels) and actual channel balance (for local channels) in order to be able to give a decision about how to much a route can transport, to support bass amplifier splitting.  Then we would also have to implement "usedness" tracking (i.e. avoid using a channel we used recently in a path), which is needed to get multiple different routes for bass amplifier splitting.

ZmnSCPxj:
>Perhaps our fuzz is too large or default risk is too low and it's overriding the natural bias towards direct channels?

The bias seems to be very tiny:

https://github.com/ElementsProject/lightning/blob/d7e233e47da383c06a5087f1614efaf1013b1ffb/gossipd/routing.c#L359-L364

I note also that `connection_fee` is the below:

https://github.com/ElementsProject/lightning/blob/d7e233e47da383c06a5087f1614efaf1013b1ffb/gossipd/routing.c#L347-L357

The key point here is that `connection_fee` still applies the fee even if this half-channel is directly from the source.

But the first hop from the source is always 0-fee.

So I propose to modify it so that if the half-chan is from the source of the route, `connecftion_fee` returns 0.

Thus, the direct route will have a very strong total cost compared to other routes, and `fuzz` will not affect direct routes (since it is a multiplier on the fee, and anything times 0 is 0).

I shall try to find time to make this alternate solution.

cdecker:
Since @rustyrussell is planning to write the `pay` plugin, and this is essentially that logic I'll let him decide whether to apply or not :wink:

ZmnSCPxj:
If the modification to `getroute` I described is taken instead, that should simplify the future `pay` plugin --- it would be a simple `getroute`-`sendpay` loop.

Regarding plugins, can plugins receive multiple command requests even while an existing command response has not been returned yet?
I worry, the case where multiple `pay` command is performed at same time.
In current `payalgo.c`, multiple `pay` command is perfectly fine and can be performed at same time (`payalgo.c` allocates multiple objects to handle the paymetn).
I worry if `pay` plugin is written in Python or similar, that only a single `pay` command can occur at one time.
It may make the `pay` plugin inappropriate for some applications, such as evil trust-requiring centralized custodial Lightning wallets.

ZmnSCPxj:
Closes #2168

cdecker:
> Closes #2168

Github seems to only auto-close if these statements are in the first message in the thread.

PR #2147, opened 2018-12-07
Private Channels

niftynei:
Allows users to create private channels, solves #2125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - JSON API: `getinfo` now returns `num_peers` `num_pending_channels`,
   `num_active_channels` and `num_inactive_channels` fields.
 - JSON API: use `\n\n` to terminate responses, for simplified parsing (pylightning now relies on this)
+- JSON API: `fundchannel` now includes an `announce` option, when false it will keep channel private. Defaults to true.
 - Plugins: Added plugins to `lightningd`, including option passthrough and JSON-RPC passthrough.
 
 ### Changed
--- a/doc/lightning-fundchannel.7.txt
+++ b/doc/lightning-fundchannel.7.txt
@@ -8,7 +8,7 @@ lightning-fundchannel - Command for establishing a lightning channel.
 
 SYNOPSIS
 --------
-*fundchannel* 'id' 'satoshi' ['feerate']
+*fundchannel* 'id' 'satoshi' ['feerate' 'announce']
 
 DESCRIPTION
 -----------
@@ -30,6 +30,10 @@ than 16777215 satoshi.
 'feerate' is an optional feerate to use.  It can be one of the strings
 'urgent', 'normal' or 'slow' to use lightningd's internal estimates:
 'normal' is the default.
+
+'announce' is an optional flag that triggers whether to announce
+this channel or not. Defaults to `true`. An unannounced channel
+is considered private.
 
 Otherwise, 'feerate' is a number, with an optional suffix:
 'perkw' means the number is interpreted as satoshi-per-kilosipa
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -770,6 +770,7 @@ static void json_fund_channel(struct command *cmd,
 	struct peer *peer;
 	struct channel *channel;
 	u32 *feerate_per_kw;
+	bool *announce_channel;
 	u8 *msg;
 	u64 max_funding_satoshi = get_chainparams(cmd->ld)->max_funding_satoshi;
 
@@ -780,6 +781,7 @@ static void json_fund_channel(struct command *cmd,
 		   p_req("id", json_tok_pubkey, &id),
 		   p_req("satoshi", json_tok_tok, &sattok),
 		   p_opt("feerate", json_tok_feerate, &feerate_per_kw),
+		   p_opt_def("announce", json_tok_bool, &announce_channel, true),
 		   NULL))
 		return;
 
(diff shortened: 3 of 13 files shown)


lightningd/opening_control.c
```diff
@@ -780,6 +781,7 @@ static void json_fund_channel(struct command *cmd,
 		   p_req("id", json_tok_pubkey, &id),
 		   p_req("satoshi", json_tok_tok, &sattok),
 		   p_opt("feerate", json_tok_feerate, &feerate_per_kw),
+		   p_opt_def("private", json_tok_bool, &private_channel, false),
```
rustyrussell: The spec calls this bit `announce_channel`; should we invert this and call it "announce", default=true?

lightningd/peer_control.c
```diff
@@ -542,21 +542,6 @@ static enum watch_result funding_lockin_cb(struct lightningd *ld,
...
-	 *     - MUST send the `announcement_signatures` message.
-	 *       - MUST NOT send `announcement_signatures` messages until
-	 *         `funding_locked` has been sent AND the funding transaction has
-	 *         at least six confirmations.
-	 *   - otherwise:
-	 *     - MUST NOT send the `announcement_signatures` message.
-	 */
-	if (!(channel->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL))
-		return DELETE_WATCH;
-
```
rustyrussell: This change is probably OK, but I'd like something in the commit msg justifying its deletion; it seems orthogonal?

tests/test_connection.py
```diff
@@ -791,6 +791,21 @@ def test_channel_persistence(node_factory, bitcoind, executor):
     l1.daemon.wait_for_log(' to ONCHAIN')
 
 
+def test_private_channel(node_factory):
+    l1, l2 = node_factory.line_graph(2, fundchannel=True, announce=True, privatechannels=True)
+    l3, l4 = node_factory.line_graph(2, fundchannel=True, announce=True, privatechannels=False)
```
rustyrussell: fundchannel defaults True already AFAICT?  We should probably rename the 'announce' parameter to 'wait_for_announce'?

tests/utils.py
```diff
@@ -839,9 +839,10 @@ def both_dirs_ready(n, scid):
         wait_for(lambda: both_dirs_ready(nodes[-1], scids[0]))
 
         # Make sure we have all node announcements, too (just check ends)
-        for n in nodes:
-            for end in (nodes[0], nodes[-1]):
-                wait_for(lambda: 'alias' in only_one(end.rpc.listnodes(n.info['id'])['nodes']))
+        if not privatechannels:
```
rustyrussell: Hmm announce=True and privatechannels=True is an invalid combination (which kinda reinforces my preference to call them 'wait_for_announce' and 'announce' BTW)  Just assert() above that they're not asking for it?

doc/lightning-fundchannel.7.txt
```diff
@@ -31,6 +31,9 @@ than 16777215 satoshi.
 'urgent', 'normal' or 'slow' to use lightningd's internal estimates:
 'normal' is the default.
 
+'private' is an optional flag to keep this channel private, i.e. not
+broadcast a `channel_announcement` message.
```
rustyrussell: s/don't offer to create a `channel_announcement` message.  Note that we won't create a channel_announcement message unless the peer agrees to as well./?

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - JSON API: `getinfo` now returns `num_peers` `num_pending_channels`,
   `num_active_channels` and `num_inactive_channels` fields.
 - JSON API: use `\n\n` to terminate responses, for simplified parsing (pylightning now relies on this)
+- JSON API: `fundchannel` now includes option to keep channel private
```
rustyrussell: Name of option, and note default is public?

lightningd/opening_control.c
niftynei: works for me. :+1:

lightningd/peer_control.c
niftynei: reverted. I was under the mistaken assumption that this would prevent private channels from being locked. Turns out that check is the line above...

tests/utils.py
niftynei: Great point. Updated.

CHANGELOG.md
niftynei: Updated.

doc/lightning-fundchannel.7.txt
rustyrussell: Damn, you're right!  Oops, I don't know how I got that idea then.  Thanks!

tests/utils.py
```diff
@@ -839,9 +840,10 @@ def both_dirs_ready(n, scid):
         wait_for(lambda: both_dirs_ready(nodes[-1], scids[0]))
 
         # Make sure we have all node announcements, too (just check ends)
-        for n in nodes:
-            for end in (nodes[0], nodes[-1]):
-                wait_for(lambda: 'alias' in only_one(end.rpc.listnodes(n.info['id'])['nodes']))
+        if not announce_channels:
+            for n in nodes:
+                for end in (nodes[0], nodes[-1]):
+                    wait_for(lambda: 'alias' in only_one(end.rpc.listnodes(n.info['id'])['nodes']))
```
rustyrussell: Can't happen?  We have `if not wait_for_announce: return` above

niftynei: ack.

niftynei:
self ACK https://github.com/ElementsProject/lightning/pull/2147/commits/8a1fc7b834a0a222921ed9197e27dd2b1fb25164

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2147/commits/7129e99584e76e40ac271a1a690d3ba29f122f6c

shesek:
Is there a way to tell if an existing channel is private or not? I can't seem to find anything indicating this in `listpeers`.

niftynei:
`listchannels` has a `public` boolean which should tell you if a channel is private or not. it'd make sense to add this to the channel obj in `listpeers` as well.

PR #2152, opened 2018-12-07
[plugin] 0.0.1 of donation plugin and invoice query service

renepickhardt:
I have created a plugin that allows c-lightning nodes to spin up a flask webserver which wraps the invoice RPC command. In this way owners of lightning nodes have a "donation"-web address for spontaneous payments.

I plan to activate this plugin tomorrow on my website and do my next youtube video coming Monday about plugin development for c-lightning. Therefor it would be cool if this plugin and #2139 would be available by than on the c-lightning master branch as I can point to them easily during the video that I am about to create. 

Otherwise I will just point to the PR. 

Other than that feedback is highly appriciated. In particular for the java script (somehow eventstream notifications and flask don't run with me) and security of the plugin. My feeling is that this is not the most secure web application.

### DIFF
--- a/contrib/plugins/donations/LICENSE
+++ b/contrib/plugins/donations/LICENSE
@@ -1,0 +1,22 @@
+Note: the modules in the ccan/ directory have their own licenses, but
+the rest of the code is covered by the following (BSD-MIT) license:
+
+	Copyright Rene Pickhardt 2018.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


contrib/plugins/donations/LICENSE
```diff
@@ -0,0 +1,22 @@
+Note: the modules in the ccan/ directory have their own licenses, but
```
ZmnSCPxj: I am uncertain why this particular note is here, when the code is in Python and does not (directly) use ccan, or have a `ccan/` subdir within this dir.

renepickhardt: Cause I copied the file from rusty and forgot to remove this stament. Will fix this later

ZmnSCPxj: Ping

contrib/plugins/donations/README.md
```diff
@@ -0,0 +1,39 @@
...
+A Simple HTTP Server is created that can serve a donation webpage and allow to issue invoices.
+The plugin takes one of the following three commands {start/stop/restart} as the first agument
+By default the plugin starts the server on port 8088. This can however be changed with the
+port argument.
+
+this means after starting `lightningd` together with the plugin you can run: `lightning-cli donationserver start` and access the server at http://localhost:8088/donation (in case you run your lightning node at localhost)
+
+## About the plugin
+You can see a demo of the plugin on the authors website at:
+https://ln.rene-pickhardt.de/donation
```
conscott: Looks like this is giving a 500 error right now

conscott:
I was unable to get the donation server running locally. After calling 
```
lightning-cli donationserver start
```
It seems the python instance crashes, but I haven't been able to dig into why just yet

I think once this is rebased off of #2188 it may be easier to debug. Ill keep trying when I have more time!

cdecker:
I added the plugin to the plugins repo, to give it a new home where it can be found and improved.

https://github.com/lightningd/plugins/tree/master/donations

Closing this PR :wink:

PR #2153, opened 2018-12-07
rpc: check improvements and manpage

wythe:
I stole (with permission) a few commits from @rustyrussell and then added a couple of my own, including the `check` manpage.


common/json.c
```diff
@@ -250,17 +250,9 @@ void json_tok_print(const char *buffer, const jsmntok_t *tok)
...
-
 	if (!tok)
 		return NULL;
-	jsmntok_t *arr = tal_arr(ctx, jsmntok_t, last - first);
-	jsmntok_t *dest = arr;
-	while (first != last)
-		*dest++ = *first++;
-
-	return arr;
+	return tal_dup_arr(ctx, jsmntok_t, tok, json_next(tok) - tok, 0);
```
wythe: Maybe we should remove the NULL check and expect the caller to make sure `tok` is non-null.  This would only be needed in the call made by `json_check`.

doc/lightning-check.7.txt
```diff
@@ -0,0 +1,37 @@
...
+-----------
+The *check* RPC command verifies another command's parameters without running it.
+
+The 'command_to_check' is the name of the relevant command.
+
+'parameters' is the command's parameters.
+
+This does not guarantee successful execution of the command in all cases.
+For example, a call to lightning-getroute(7) may still fail to find a route even if
+checking the parameters succeeds.
```
wythe: This makes me wonder if we should change the command's name to `verify` and not `check`.

common/json.c
rustyrussell: Good point: json_check() can never have NULL params, too.

My general rule is to give a nice crash on NULL unless there's a good reason to handle it.

doc/lightning-check.7.txt
rustyrussell: Meh, "check" is OK.  It would be nicer if we removed more json_tok_tok calls and did proper parsing inline, however.  For example, 'check connect asdfsaf' passes because the parsing of the 'id' parameter happens afterwards.

common/json.c
rustyrussell: Since it's trivial, I'll fold and apply...

PR #2154, opened 2018-12-08
Make plugins more talkative (log notifications plugin->lightningd)

cdecker:
Adds notification handling to `lightningd` so that we can start injecting logs into the main daemon output. Should make it easier to follow what happens in the plugins.

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -45,12 +45,27 @@ def json_getmanifest(request):
     }
 
 
+def plugin_log(message, level="info"):
+    payload = {
+        "jsonrpc": "2.0",
+        "method": "log",
+        "params": {
+            "level": level,
+            "message": message,
+        }
+    }
+    json.dump(payload, fp=sys.stdout)
+    sys.stdout.write('\n\n')
+    sys.stdout.flush()
+
+
 def json_init(request, options, configuration):
     """The main daemon is telling us the relevant cli options
     """
     global greeting
-
     greeting = request['params']['options']['greeting']
+    plugin_log("Plugin helloworld.py initialized with greeting \"{}\"".format(greeting), "debug")
+
     return "ok"
 
 
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -234,6 +234,66 @@ static void plugin_request_queue(struct plugin *plugin,
 	io_wake(plugin);
 }
 
+static void plugin_notification_handle(struct plugin *plugin,
+				       const jsmntok_t *toks)
+{
+	const jsmntok_t *methtok, *paramstok;
+
+	methtok = json_get_member(plugin->buffer, toks, "method");
+	paramstok = json_get_member(plugin->buffer, toks, "params");
+
+	if (!methtok || !paramstok) {
+		plugin_kill(plugin,
+			    "Malformed JSON-RPC notification missing "
+			    "\"method\" or \"params\": %.*s",
+			    toks->end - toks->start,
+			    plugin->buffer + toks->start);
+		return;
+	}
+
+	/* Dispatch incoming notifications. This is currently limited
+	 * to just a few method types, should this ever become
+	 * unwieldy we can switch to the AUTODATA construction to
+	 * register notification handlers in a variety of places. */
+	if (json_tok_streq(plugin->buffer, methtok, "log")) {
+		plugin_log_handle(plugin, paramstok);
+	} else {
+		plugin_kill(plugin, "Unknown notification method %.*s",
+			    json_tok_len(methtok),
+			    json_tok_contents(plugin->buffer, methtok));
+	}
+}
+
+static void plugin_response_handle(struct plugin *plugin,
+				   const jsmntok_t *toks,
+				   const jsmntok_t *idtok)
+{
+	struct plugin_request *request;
+	u64 id;
+	/* We only send u64 ids, so if this fails it's a critical error (note
+	 * that this also works if id is inside a JSON string!). */
+	if (!json_to_u64(plugin->buffer, idtok, &id)) {
+		plugin_kill(plugin,
+			    "JSON-RPC response \"id\"-field is not a u64");
+		return;
+	}
+
+	request = uintmap_get(&plugin->plugins->pending_requests, id);
+
+	if (!request) {
+		plugin_kill(
+		    plugin,
+		    "Received a JSON-RPC response for non-existent request");
+		return;
+	}
+
+	/* We expect the request->cb to copy if needed */
+	request->cb(request, plugin->buffer, toks, idtok, request->arg);
+
+	uintmap_del(&plugin->plugins->pending_requests, id);
+	tal_free(request);
+}
+
 /**
  * Try to parse a complete message from the plugin's buffer.
  *
@@ -242,11 +302,8 @@ static void plugin_request_queue(struct plugin *plugin,
  */
 static bool plugin_read_json_one(struct plugin *plugin)
 {
-	jsmntok_t *toks;
 	bool valid;
-	u64 id;
-	const jsmntok_t *idtok;
-	struct plugin_request *request;
+	const jsmntok_t *toks, *jrtok, *idtok;
 
 	/* FIXME: This could be done more efficiently by storing the
 	 * toks and doing an incremental parse, like lightning-cli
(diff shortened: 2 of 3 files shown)


contrib/plugins/helloworld.py
```diff
@@ -45,12 +45,27 @@ def json_getmanifest(request):
...
+    sys.stdout.flush()
+
+
 def json_init(request, options, configuration):
     """The main daemon is telling us the relevant cli options
     """
     global greeting
-
     greeting = request['params']['options']['greeting']
+    plugin_log("Plugin helloworld.py initialized with greeting \"{}\"".format(greeting), "debug")
```
renepickhardt: This allows the developer to depict a message via calling b plugin_log. This is very helpful and nice! For exceptions I will be able to catch them and push the error string to this method. 

However if my third party libs use std.out I will still run into problems.

lightningd/plugin.c
```diff
@@ -256,6 +256,91 @@ static void plugin_request_queue(struct plugin_request *req)
...
+
+	if (!request) {
+		plugin_kill(
+		    plugin,
+		    "Received a JSON-RPC response for non-existent request");
+		return;
+	}
+
+	/* We expect the request->cb to copy if needed */
+	request->response = plugin->buffer;
```
renepickhardt: So the request message body is replied back in the response? This will not make the plugin display the responds or an I mistaking? I saw in line 290 you push the message to the clightning logfiles as I had expected it

lightningd/plugin.c
```diff
@@ -256,6 +256,91 @@ static void plugin_request_queue(struct plugin_request *req)
...
+		plugin_kill(plugin,
+			    "Unknown log-level %.*s, valid values are "
+			    "\"debug\", \"info\", \"warn\", or \"error\".",
+			    json_tok_len(leveltok),
+			    json_tok_contents(plugin->buffer, leveltok));
+		return;
+	}
+
+	/* We strip the \" otherwise they'd be printed here. */
+	log_(plugin->log, level, "log: %.*s", json_tok_len(msgtok) - 2,
```
renepickhardt: Would be great if there would not only be the message but also the name of the plugin written down to the log. Not sure what's the best way to achieve this. 

I guess having the plugin send their name in the log message. However it would be great to fix the name and map it against registered plugins.

contrib/plugins/helloworld.py
renepickhardt: well that might not be to bad. for exaple in flask I could do app.run in this case. 
also I could regularly do plugin_log(f.getvalue()) to send the message to c-lightning) but yes I guess this is really just temporary.

lightningd/plugin.c
cdecker: `plugin->log` is a logger that is prefixed with the plugin's basename, so the complete line here reads 

```
2018-12-09T12:42:59.072Z plugin-helloworld.py log: Plugin helloworld.py initialized with greeting \"World\"
```

Which I think is what you want.

lightningd/plugin.c
cdecker: This was a refactor and not part of the notifications handling. This is when we receive a response to a call that we sent ourselves (and not a notification which doesn't have an ID). See the split in lines 389 - 432. The `request` struct contains all data that is related to this request-response cycle and will be used to dispatch the response to the caller (which might be internal or an external JSON-RPC call that we forwarded).

This is just moved into its own function from here: https://github.com/ElementsProject/lightning/blob/c497bad2e92a687e181d32adf02af05b8a96bfd9/lightningd/plugin.c#L317-L322

contrib/plugins/helloworld.py
cdecker: Working on a plugin class that monkey-patches stdout and stderr so we can start using these streams like we normally do, but `print`s result in log notifications instead.

lightningd/plugin.c
renepickhardt: When I Previously used plugin->log I did not see the plugin name. (maybe I was just blind) But maybe I set up my plugin incorrectly or did some other mistake. I will try to see if I can reconstruct this behavior. But it's great that it is already implemented!

contrib/plugins/helloworld.py
rustyrussell: Note that stderr now goes to parent stderr: this is nice for simple debugging.  Like when my plugins backtrace :)

lightningd/plugin.c
rustyrussell: Was changed recently, in commit 111d6df442d68a7f05fc7c54a2345752ed221cc3

cdecker: I was hoping to differentiate the entries `lightningd` writes about the plugin from the ones that the plugins inject themselves. But you're probably right, it's unlikely that a plugin would inject a log-line pretending to kill itself :-)

cdecker:
Rebased on top of recent changes to the plugins.

contrib/plugins/helloworld.py
cdecker: My small python Plugin library grabs `stderr` as well and turns things into log notifications. This has the advantage that it provides more context than just spurious writes to `stderr`. See the `plugin-6` PR that I'll open in a few minutes.

lightningd/plugin.c
```diff
@@ -256,6 +256,91 @@ static void plugin_request_queue(struct plugin_request *req)
...
+	const jsmntok_t *msgtok, *leveltok;
+	enum log_level level;
+	msgtok = json_get_member(plugin->buffer, paramstok, "message");
+	leveltok = json_get_member(plugin->buffer, paramstok, "level");
+
+	if (!msgtok) {
+		plugin_kill(plugin, "Log notification from plugin doesn't have "
+				    "a \"message\" field");
+		return;
+	}
```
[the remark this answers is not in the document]
cdecker: Check added as a fixup.

lightningd/plugin.c
```diff
@@ -256,6 +256,91 @@ static void plugin_request_queue(struct plugin_request *req)
...
+			    "\"debug\", \"info\", \"warn\", or \"error\".",
+			    json_tok_len(leveltok),
+			    json_tok_contents(plugin->buffer, leveltok));
+		return;
+	}
+
+	/* We strip the \" otherwise they'd be printed here. */
+	log_(plugin->log, level, "log: %.*s", json_tok_len(msgtok) - 2,
+	     json_tok_contents(plugin->buffer, msgtok) + 1);
+}
```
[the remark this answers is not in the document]
cdecker: It just felt a bit safer to use the accessor, but you're right, if we use the accessor and then work around its quirks, it's probably best to just use the raw pointer access.

lightningd/plugin.c
```diff
@@ -256,6 +256,91 @@ static void plugin_request_queue(struct plugin_request *req)
...
+
+	if (!leveltok || json_tok_streq(plugin->buffer, leveltok, "info"))
+		level = LOG_INFORM;
+	else if (json_tok_streq(plugin->buffer, leveltok, "debug"))
+		level = LOG_DBG;
+	else if (json_tok_streq(plugin->buffer, leveltok, "warn"))
+		level = LOG_UNUSUAL;
+	else if (json_tok_streq(plugin->buffer, leveltok, "error"))
+		level = LOG_BROKEN;
+	else {
```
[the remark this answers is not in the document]
cdecker: Notice that this is just mapping incoming log notifications from the plugin. It does not handle `LOG_IO` which would be produced internally by `lightningd` and be passed directly to the logger that way.

cdecker:
Ready for round 2 :-)

Ping @renepickhardt and @rustyrussell

lightningd/plugin.c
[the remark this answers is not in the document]
rustyrussell: We should probably rename those helpers, they're damn misleading...

json_tok_fulltoken() and json_tok_fulltoken_len()?

lightningd/plugin.c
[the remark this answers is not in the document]
rustyrussell: But the plugin may want to do IO logging for other unrelated things...

lightningd/plugin.c
```diff
@@ -269,32 +296,54 @@ static bool plugin_read_json_one(struct plugin *plugin)
...
-		return false;
-	}
-
-	request = uintmap_get(&plugin->plugins->pending_requests, id);
-
-	if (!request) {
-		plugin_kill(plugin, "Received a JSON-RPC response for non-existent request");
-		return false;
+	if (idtok && (errortok || resulttok)) {
+		/* When a rpc call is made, the Server MUST reply with
```
rustyrussell: Remove all these checks, IMHO.  It's none of our business what they send back and forth, as long as it's an object with matching `id`.  That's minimal.

In fact, this patch seems entirely extraneous; the next patch should just touch the "!idtok" case....

lightningd/plugin.c
```diff
@@ -314,7 +363,7 @@ static struct io_plan *plugin_read_json(struct io_conn *conn UNUSED,
 	/* Read and process all messages from the connection */
 	do {
 		success = plugin_read_json_one(plugin);
-	} while (success);
+	} while (success && ! plugin->stop);
 
 	if (plugin->stop)
 		return io_close(plugin->stdout_conn);
```
rustyrussell: Good (unrelated) catch; but just check plugin->stop inside the loop, rather than doing so twice?

lightningd/plugin.c
cdecker: That might be true for passthrough RPC calls, but not all are passthrough, so we better make sure that they adhere to the JSON-RPC standard. We have `getmanifest` and `init` which are handled by us internally, and we can check either here, or in the callbacks, but this looked like a good place to do so.

Also we'll add more of the internally handled RPC calls, once we start moving the bitcoin connection to a plugin (`bitcoind`, neutrino, ...) so I'd like to keep them here.

rustyrussell: It's not just a layering violation, it's also slow for large objects; at least with 'id' it's likely to be at the front.  Especially once you add looking for "params" and "method" in the next patch.

We really should be doing this only once we've identified that it's something we need to deal with.  That's both minimal and clearer.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2154/commits/909ac60761143c09dcb4041d78ce24e6b690640e

The JSON-RPC.org quotes are a bit verbose, but it's nice to have them somewhere people might read!

PR #2155, opened 2018-12-10
Disable option_dataloss_protect

rustyrussell:
My test node has started getting disconnects from lnd nodes on reconnection.  I think it's our fault, but I had a great deal of trouble with this code the first time, so I'm disabling it now to avoid going down that rabbit hole again.

The option is subsumed by option_simplified_commitment anyway, which is coming in Two Weeks(TM)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ changes.
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


cdecker:
This will effectively make us incompatible with BLW again since they require the option with all their peers.

I'm ok with that, if fixing this takes more effort than to just roll over to the next commitment.

PR #2159, opened 2018-12-10
plugin: Fix an issue with string IDs ending up quoted twice

cdecker:
The transparent passthrough that was recently introduced would end up
causing phantom quotes to appear around IDs when one of them was a
string. This happened for example when using `lightning-cli`, the code
would copy the quotes from the original request, insert our u64 ID,
and then re-add them on the way back as well.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -484,13 +484,18 @@ static void json_stream_forward_change_id(struct json_stream *stream,
 					  const jsmntok_t *idtok,
 					  const char *new_id)
 {
-	/* We copy everything, but replace the id (maybe inside a string, we
-	 * don't care) */
+	/* We copy everything, but replace the id. Special care has to
+	 * be taken when the id that is being replaced is a string. If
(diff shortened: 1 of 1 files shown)


lightningd/plugin.c
```diff
@@ -484,13 +484,18 @@ static void json_stream_forward_change_id(struct json_stream *stream,
...
 					  const char *new_id)
 {
-	/* We copy everything, but replace the id (maybe inside a string, we
-	 * don't care) */
+	/* We copy everything, but replace the id. Special care has to
+	 * be taken when the id that is being replaced is a string. If
+	 * we don't crop the quotes off we'll transform a numeric
+	 * new_id into a string, or even worse, quote a string id
+	 * twice. */
+	size_t offset = idtok->type==JSMN_STRING?1:0;
```
niftynei: nit: spaces? ie. `..ING ? 1 : 0;`

rustyrussell:
Ack a5da3c81a126ad91c5978ea13069389be2da4f81

PR #2161, opened 2018-12-10
pylightning: Add a plugin framework

cdecker:
This framework is very much inspired by flask, in that I wanted to make it as trivial as possible to get started. I had to jump through some hoops but I'm rather happy with the end result. A plugin with an option and a RPC passthrough is as simple as this:

```python
from lightning import Plugin, monkey_patch
plugin = Plugin(autopatch=True)

@plugin.method("hello")
def hello(name, plugin):
    """Say hello"""
    greeting = plugin.get_option('greeting')
    plugin.log('{} {}'.format(greeting, name))
    return s

plugin.add_option('greeting', 'Hello', 'The greeting I should use.')
plugin.run()
````

It has support for monkey-patching `stdout` and `stderr` by default so that if run by `lightningd` it'll automatically wrap anything written to those streams into log notifications that get displayed nicely.

Depends-On: #2154

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -1,0 +1,231 @@
+import sys
+import os
+import json
+import inspect
+import traceback
+
+
+class Plugin(object):
+    """Controls interactions with lightningd, and bundles functionality.
+
+    The Plugin class serves two purposes: it collects RPC methods and
+    options, and offers a control loop that dispatches incoming RPC
+    calls and hooks.
+
+    """
+
+    def __init__(self, stdout=None, stdin=None, autopatch=True):
+        self.methods = {}
+        self.options = {}
+
+        if not stdout:
+            self.stdout = sys.stdout
+        if not stdin:
+            self.stdin = sys.stdin
+
+        if os.getenv('LIGHTNINGD_PLUGIN') and autopatch:
+            monkey_patch(self, stdout=True, stderr=True)
+
+        self.add_method("getmanifest", self._getmanifest)
+        self.rpc_filename = None
+        self.lightning_dir = None
+        self.init = None
+
+    def add_method(self, name, func):
+        """Add a plugin method to the dispatch table.
+
+        The function will be expected at call time (see `_dispatch`)
+        and the parameters from the JSON-RPC call will be mapped to
+        the function arguments. In addition to the parameters passed
+        from the JSON-RPC call we add a few that may be useful:
+
+         - `plugin`: gets a reference to this plugin.
+
+         - `request`: gets a reference to the raw request as a
+           dict. This corresponds to the JSON-RPC message that is
+           being dispatched.
+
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: contrib/pylightning/lightning/plugin.py [hunk 1]
```diff
@@ -0,0 +1,227 @@
...
+
+        Returns the last partial message that was not complete yet.
+        """
+        for payload in msgs[:-1]:
+            request = json.loads(payload)
+
+            try:
+                result = {
+                    "jsonrpc": "2.0",
+                    "result": self._dispatch(request),
```
ZmnSCPxj: It looks to me that this will result in "linear" dispatching, i.e. commands are handled one-at-a-time.

Would it be possible to have this handle dispatching in parallel, i.e. if a command stalls processing, another command can be processed?

Or is some Python limitation exists to prevent this, or it becomes complicated?

(maybe there can be a separate decoration for methods that might block)

cdecker: Good point, hadn't thought about it yet. I wanted to keep away from multi-threading as much as possible, but I'll try to see what I can come up with.

cdecker: I opened issue #2167 to track progress on this. Would love to hear feedback for the initial idea, which is future-inspired, but since the python futures are tightly bound to multi-threading I'm thinking of implementing my own `AsyncResult` instead.

ZmnSCPxj: This seems sensible.

contrib/pylightning/lightning/plugin.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def monkey_patch(plugin, stdout=True, stderr=False):
+    """Monkey patch stderr and stdout so we use notifications instead.
+
+    A plugin commonly communicates with lightningd over its stdout
+    and stdin filedescriptor, so if we use them in some other way
+    (printing, logging, ...) we're breaking our communication
+    channel. This function
+    """
```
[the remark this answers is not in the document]
ZmnSCPxj: It is a common problem that occurs when monkey patching system internals. The comment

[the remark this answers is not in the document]
cdecker: xD

contrib/pylightning/lightning/plugin.py
```diff
@@ -17,6 +17,7 @@ class Plugin(object):
     def __init__(self, stdout=None, stdin=None, autopatch=True):
         self.methods = {}
         self.options = {}
+        self.option_values = {}
```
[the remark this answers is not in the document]
cdecker: `options` are the names of options that we registered with all their metadata, whereas `option_values` are the ones that were set by the `init` call from the plugin. We can merge them if desired of course.

cdecker:
Merged `options` and `option_values` as suggested by @rustyrussell and fixed the comm

PR #2158, opened 2018-12-10
static link sqlite library

arowser:
Static link sqlite3 library to avoid sqlite3 library version mismatch error: #1375 #1497
```
./lightningd
SQLITE version mismatch: compiled 3011000, now 3022000
```

Before:
```
ldd ./lightningd   
	linux-vdso.so.1 (0x00007fff8d357000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f4d75946000)
	libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007f4d7563d000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f4d75425000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f4d75034000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f4d75ce4000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f4d74e15000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f4d74c11000)

```

After:
```
 ldd lightningd/lightningd
	linux-vdso.so.1 (0x00007ffe968e7000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007ff2dc5ce000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007ff2dc3af000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007ff2dc1ab000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007ff2dbf93000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007ff2dbba2000)
	/lib64/ld-linux-x86-64.so.2 (0x00007ff2dcda7000)

```

PR #2164, opened 2018-12-11
Check system-provided library versions

rustyrussell:
@arowser 's #2158 PR drew my attention to this problem.  We have two other libraries for which this could be an issue, and generally static linking is a bad idea, as security updates are then omitted.

Unfortunately I can't find a statement on sqlite3's versioning which guarantees ABI compatibility across minor versions, so I'm reluctant to loosen the ABI check.  Though ubuntu obviously thought it ABI compatible...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,8 +334,16 @@ gen_version.h: FORCE
 	@(echo "#define VERSION \"$(VERSION)\"" && echo "#define BUILD_FEATURES \"$(FEATURES)\"") > $@.new
 	@if cmp $@.new $@ >/dev/null 2>&2; then rm -f $@.new; else mv $@.new $@; echo Version updated; fi
 
-# All binaries require the external libs, ccan
-$(ALL_PROGRAMS) $(ALL_TEST_PROGRAMS): $(EXTERNAL_LIBS) $(CCAN_OBJS)
+# We force make to relink this every time, to detect version changes.
+tools/headerversions: FORCE tools/headerversions.o $(CCAN_OBJS)
+	@$(LINK.o) tools/headerversions.o $(CCAN_OBJS) $(LOADLIBES) $(LDLIBS) -o $@
+
+# That forces this rule to be run every time, too.
+gen_header_versions.h: tools/headerversions
+	@tools/headerversions $@
+
+# All binaries require the external libs, ccan and external library versions.
+$(ALL_PROGRAMS) $(ALL_TEST_PROGRAMS): $(EXTERNAL_LIBS) $(CCAN_OBJS) gen_header_versions.h
 
 # Each test program depends on its own object.
 $(ALL_TEST_PROGRAMS): %: %.o
--- a/Makefile
+++ b/Makefile
@@ -342,8 +342,11 @@ tools/headerversions: FORCE tools/headerversions.o $(CCAN_OBJS)
 gen_header_versions.h: tools/headerversions
 	@tools/headerversions $@
 
-# All binaries require the external libs, ccan and external library versions.
-$(ALL_PROGRAMS) $(ALL_TEST_PROGRAMS): $(EXTERNAL_LIBS) $(CCAN_OBJS) gen_header_versions.h
+# Rebuild the world if this changes.
+ALL_GEN_HEADERS += gen_header_versions.h
+
+# All binaries require the external libs, ccan and system library versions.
+$(ALL_PROGRAMS) $(ALL_TEST_PROGRAMS): $(EXTERNAL_LIBS) $(CCAN_OBJS)
 
 # Each test program depends on its own object.
 $(ALL_TEST_PROGRAMS): %: %.o
--- a/tools/headerversions.c
+++ b/tools/headerversions.c
@@ -12,10 +12,42 @@
 #include <sys/types.h>
 #include <zlib.h>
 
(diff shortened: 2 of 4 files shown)


tools/headerversions.c
```diff
@@ -0,0 +1,80 @@
...
+	"/* Generated file by tools/headerversions, do not edit! */\n"
+	"/* GMP version: %s */\n"
+	"/* SQLITE3 version: %u */\n"
+	"/* ZLIB version: %s */\n"
+	"#include <ccan/err/err.h>\n"
+	"#include <gmp.h>\n"
+	"#include <sqlite3.h>\n"
+	"#include <zlib.h>\n"
+	"\n"
+	"static void check_linked_library_versions(void)\n"
```
niftynei: this is marked `static` which i thought made methods 'private' to the file. how does `lightningd` have access to it?

Makefile
```diff
@@ -334,7 +334,18 @@ gen_version.h: FORCE
...
+# We force make to relink this every time, to detect version changes.
+tools/headerversions: FORCE tools/headerversions.o $(CCAN_OBJS)
+	@$(LINK.o) tools/headerversions.o $(CCAN_OBJS) $(LOADLIBES) $(LDLIBS) -o $@
+
+# That forces this rule to be run every time, too.
+gen_header_versions.h: tools/headerversions
+	@tools/headerversions $@
+
+# Rebuild the world if this changes.
+ALL_GEN_HEADERS += gen_header_versions.h
```
niftynei: much magic. very cool. :sparkles:

tools/headerversions.c
rustyrussell: Because lightningd literally `#include`s this file.  The norm would be to make this 'static inline' for a header file; I've fixed that now.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2164/commits/e3e8a3c8e3f9bd4447077a298f2ef433ac9ad5ad

arowser:
@rustyrussell 
Sorry for comments after merge, I think this pr will add more dependcy,  and make lightningd wok only with some special sqlite3/zlib/gm library version.

If the user download the release binary, they must reinstall these special version library to make the c-lightning work. if we static link these library, it will make the release binary working on more linux system then dynamic link.

Before this PR:
```
ldd ../lightningd_old 
	linux-vdso.so.1 (0x00007fffc41a8000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f16240ce000)
	libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007f1623dc5000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f1623bad000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f16237bc000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f16247a1000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f162359d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f1623399000)
```
After this PR:
```
ldd lightningd/lightningd
	linux-vdso.so.1 (0x00007ffcd17f5000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fade29de000)
	libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007fade275d000)
	libsqlite3.so.0 => /usr/lib/x86_64-linux-gnu/libsqlite3.so.0 (0x00007fade2454000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fade2237000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fade201f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fade1c2e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fade30b1000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fade1a0f000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fade180b000)
```

arowser:
The sqlite have good compatibility in different version.
From https://sqlite.org/formatchng.html found:
```
Since 2004, there have been enhancements to SQLite such that newer database files are unreadable by older versions of the SQLite library. But the most recent versions of the SQLite library should be able to read and write any older SQLite database file without any problems
```
D. Richard Hipp( the architect and primary author of SQLite ) said in:
http://sqlite.1065341.n5.nabble.com/sqlite-versions-binary-compatibility-td27401.html
```
The database file for all versions of SQLite are 
backwards compatible for all versions of SQLite going 
bach through version 3.0.0.  Any newer version of 
SQLite can read and write database written by older 
versions of SQLite.  For the most part this works 
in the opposite direction too - older versions of 
SQLite can read and write databases created by 
newer versions of SQLite.  The exception is if the 
newer version of SQLite created database that uses 
some feature that was added after the older version 
of SQLite was released. 
```

If we can just ignore sqlite version check If don't use static link the sqlite3 library?

Issue #2119, opened 2018-11-27
Direct channel should be preferred


ZmnSCPxj:
Currently, route randomization means that sometimes we do not take the direct route.

However, it should be remembered that route randomization is intended to prevent third parties from learning all of our payments.

And if a one-channel route exists that is viable, we should instead consider paying it so that none of our payment information to the payee node is even leaked to intermediaries.
Thus, for this particular case (a route exists that is direct between payer to payee, and is viable for paying), route randomization is hurtful.

--

One might argue, that perhaps we might not want to let the payee node know that we are in fact directly connected, so using an alternate route will at least obscure this.

One can also argue that the payee probably already knows where we live in order to  deliver their goods to us, but at least we can prevent the payee from having an idea of where our node is on LN.

In addition, this is viable only if we regularly transact with the payee anyway.  For one-off purchases, being directly connected to the payee is largely immaterial.

--

I can implement this now with current payalgo.

However, we will need to change the payalgo almost completely in order to support ZKCP-preserving multipath payments anyway, so I wonder if I should instead defer this particular subfeature to the new payalgo.

kristapsk:
I would also prefer use of a direct channel, if it is available. If others may not, probably this could be made configurable?

kristapsk:
Today tried to pay another yalls.org invoice and it failed (still have direct channel with yours node open). Not able to pay at all when I have direct channel open sounds really like a bug to me. At least, if routing fails, it should fall back to direct route, if it exists.
```
{ "code" : 200, "message" : "Stopped retrying during payment attempt; continue monitoring with pay or listpayments", "data" : 
{
  "getroute_tries": 1, 
  "sendpay_tries": 1, 
  "id": 11, 
  "payment_hash": "f98639a3901b6dec7ebebf213dd29f3f9fe1134a07175b73087da64068b5c953", 
  "destination": "03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56", 
  "msatoshi": 150000, 
  "msatoshi_sent": 151396, 
  "created_at": 1543532561, 
  "status": "pending", 
  "description": "Read: Hi guys, New here just testing o", 
  "failures": [
  ]
} }
```
Related to #2108.

That `lightningd` process had crashed before (most likely my fault, disk partition got full) and probably I tried payment too early, before it restored connections to other nodes or something? (but I'm not sure that is the case)

ZmnSCPxj:
[...]

PR #2168, opened 2018-12-11
Prefer direct routes when paying - alternate solution

ZmnSCPxj:
Fixes: #2119 

Closes: #2144 

Alternate solution that tweaks `connection_fee` so it gives 0 fees for direct channels.

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -344,12 +344,32 @@ static void clear_bfg(struct node_map *nodes)
 	}
 }
 
-static u64 connection_fee(const struct half_chan *c, u64 msatoshi)
+/**
+ * @brief connection_fee
+ *        Compute the fee for the connection.
+ *
+ * @param from - the node that is the ultimate source of the
+ *        payment.
+ * @param chan - the channel to consider.
+ * @param idx - which direction (half_chan) of the channel to consider. 0 or 1.
+ * @param msatoshi - the value to transport over this channel.
+ */
+static u64 connection_fee(const struct node *from,
+			  const struct chan *chan, int idx, u64 msatoshi)
 {
 	u64 fee;
+	const struct half_chan *c;
 
+	assert(idx == 0 || idx == 1);
 	assert(msatoshi < MAX_MSATOSHI);
+
+	c = &chan->half[idx];
+
 	assert(c->proportional_fee < MAX_PROPORTIONAL_FEE);
+
+	/* The ultimate source never charges fees. */
+	if (chan->nodes[idx] == from)
+		return 0;
 
 	fee = (c->proportional_fee * msatoshi) / 1000000;
 	/* This can't overflow: c->base_fee is a u32 */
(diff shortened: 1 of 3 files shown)


gossipd/routing.c
```diff
@@ -344,13 +344,33 @@ static void clear_bfg(struct node_map *nodes)
...
 
-static u64 connection_fee(const struct half_chan *c, u64 msatoshi)
+/**
+ * @brief connection_fee
+ *        Compute the fee for the connection.
+ *
+ * @param from - the node that is the ultimate source of the
+ *        payment.
+ * @param chan - the channel to consider.
+ * @param idx - which half of the channel to consider. 0 or 1.
```
cdecker: We commonly call this direction.

gossipd/routing.c
```diff
@@ -1529,10 +1556,11 @@ struct route_hop *get_route(const tal_t *ctx, struct routing_state *rstate,
 		hops[i].nodeid = n->id;
 		hops[i].amount = total_amount;
 		hops[i].delay = total_delay;
-		total_amount += connection_fee(c, total_amount);
+		total_amount += connection_fee(sourcenode, route[i], idx, total_amount);
```
cdecker: I'm wondering if we should simply skip this addition if `i==0` since that'd be the hop that we are in control over anyway. That'd avoid having to thread an additional argument in and do the check in another function.

ZmnSCPxj:
> wouldn't just skipping the last fee computation have the same effect?

No, as that is done, once a route has been chosen.
We need to change, the way of choosing a route, so we need to change either `connection_fee` or `bfg_one_edge`.

gossipd/routing.c
ZmnSCPxj: The data structure is called `half_chan` and the field is `half[2]`, but I suppose `direction (half_chan)` would be okay.

gossipd/routing.c
ZmnSCPxj: `connection_fee` is used in both this and `bfg_one_edge`.
We could argue for moving the check to `bfg_one_edge`, but properly, the source node charges no fee to itself for forwarding, so I thought, more appropriate at `connection_fee`.

rustyrussell:
In the initial implementation I decided to keep the weighting because I figured it reflected our own desires for usage (e.g. we dropped fee on one channel because we wanted more traffic there).

What I don't understand is why this would change anything?  I suppose now the cost comparison is X vs 0, whereas it was 2X vs X?

ZmnSCPxj:
> I suppose now the cost comparison is X vs 0, whereas it was 2X vs X?

Yes. Further, route randomization fuzz is a random multiplier, so 0 removes the effect of random fuzz.

cdecker:
> Yes. Further, route randomization fuzz is a random multiplier, so 0 removes the effect of random fuzz.

Interesting. I forgot to ask about randomization, but a 0 here seems to solve this issue :-)

Any major objections @rustyrussell ?

NicolasDorier:
I wish this was merged. I created direct channel in the expectation of having higher latency for payments to a specific party. I was bit sad to see it was not working that way. :(

NicolasDorier:
would be nice to get in 0.6.3

Issue #2074, opened 2018-11-05
[Feature Request] Respond to SIGTERM


NicolasDorier:
When clightning receive a SIGTERM, it does not cleanly shutdown. It would be nice if it could shutdown properly.

PR #2172, opened 2018-12-12
Handle SIGTERM and SIGINT

wythe:
Addresses #2074.

I made`ld` a file global variable in order to do this. @rustyrussell already admitted it was "basically a global variable" in the comment below :-)

It seems the code really wants to exit via `io_break` so that's the approach I took.

I tried to just set a `bool` and exit the `main()` `for(;;)` loop by checking it.  But it was not very responsive and I got an error from `hsmd` with SIGINT.

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -574,6 +574,38 @@ void notify_new_block(struct lightningd *ld, u32 block_height)
 	channel_notify_new_block(ld, block_height);
 }
 
+static void on_sigint(int _ UNUSED)
+{
+        static const char *msg = "lightningd: SIGINT caught, exiting.\n";
+        write(STDERR_FILENO, msg, strlen(msg));
+        _exit(1);
+}
+
+static void on_sigterm(int _ UNUSED)
+{
+        static const char *msg = "lightningd: SIGTERM caught, exiting.\n";
+        write(STDERR_FILENO, msg, strlen(msg));
+        _exit(1);
+}
+
+/*~ We only need to handle SIGTERM and SIGINT for the case we are PID 1 of
+ * docker container since Linux makes special this PID and requires that
+ * some handler exist. */
+static void setup_sig_handlers(void)
+{
+	struct sigaction sigint, sigterm;
+	memset(&sigint, 0, sizeof(struct sigaction));
+	memset(&sigterm, 0, sizeof(struct sigaction));
+
+	sigint.sa_handler = on_sigint;
+	sigterm.sa_handler = on_sigterm;
+
+	if (1 == getpid()) {
+		sigaction(SIGINT, &sigint, NULL);
+		sigaction(SIGTERM, &sigterm, NULL);
+	}
+}
+
 int main(int argc, char *argv[])
 {
 	struct lightningd *ld;
@@ -582,6 +614,8 @@ int main(int argc, char *argv[])
 
 	/*~ What happens in strange locales should stay there. */
 	setup_locale();
+
+	setup_sig_handlers();
 
 	/*~ This checks that the system-installed libraries (usually
 	 * dynamically linked) actually are compatible with the ones we


lightningd/lightningd.c
```diff
@@ -573,11 +575,18 @@ void notify_new_block(struct lightningd *ld, u32 block_height)
 	channel_notify_new_block(ld, block_height);
 }
 
+/*~ Signal handler. Shuts down by breaking out of io loop. */
+static void shutdown_lightningd(int signum)
+{
+	log_info(ld->log, "Received signal %d, shutting down.", signum);
```
ZmnSCPxj: Is `log_info` reentrant? It is theoretically possible that `log_info` (or a function manipulating same data structures) is executing when a signal is delivered.  Calling it again while it is already executing may leave its data structures in an invalid state, leading to problems as we try to clean it up while exiting.

I suggest using a Posix `pipe` as I mentioned.  This is known as the self-pipe trick:

* https://www.sitepoint.com/the-self-pipe-trick-explained/
* https://lwn.net/Articles/177897/
* http://man7.org/tlpi/code/online/diff/altio/self_pipe.c.html

wythe: I appreciate the feedback.  I'll get to work on the pipe approach.

rustyrussell:
I'm not sure why we want to handle these?  Our system should be entirely robust against such failures; if it's simply to log them that's OK, but we should use the pipe trick to do it...

NicolasDorier:
@rustyrussell I use clightning in BTCPay through docker.

If there is any restart, or update that require clightning to restart, by default, docker-compose will hang for 10 sec before killing the service. Making the downtime during update to 10 sec, by default.
Knowing there is like 10 services in the docker-compose, if everybody did the same it would take 100 sec of downtime.

Now, because I don't want to kill bitcoind to not have a corrupt state, I don't use a default timeout of 10 sec, but of 300 sec!
So it means docker-compose will wait 300 sec for clightning to close before killing it.

Luckily, you can ask docker-compose to use SIGKILL instead of SIGTERM for specific containers. But this is just a bandaid that bring more burden on the person integrating clightning. When the only thing to do is to handle SIGTERM like does virtually 99% of the apps which are not written in C.

Also, are you really sure that SQLite (this is the DB you are using right?) will be happy to be killing in the middle of a write operation?

ZmnSCPxj:
I am almost sure sqlite3 would be perfectly happy with being killed; that is not much different from literal pull-the-plug. https://www.sqlite.org/howtocorrupt.html

That said, cleanly closing and flushing our own logs and prints would be a nice feature especially to capture possible problems that might induce a human operator to send `SIGTERM` et al.

NicolasDorier:
> You misunderstand. We currently "handle" SIGTERM now by exiting immediately. This makes no difference; we just don't get a chance to log that we got SIGTERM.

I don't think you do, else this PR would not be very useful. Also I would be at loss explaining why the clightning container does not close cleanly without me configuring docker compose to use SIGKILL instead of SIGTERM.

cdecker:
This works perfectly fine:

```
pkill -TERM lightningd
```

and the corresponding process exits without noticeable delay:

```
[2]    20623 terminated  lightningd/lightningd --lightning-dir=/tmp/l2 --network=bitcoin
```

`docker` is known to mishandle signals in a few cases (I always use `-t` to have it allocate a `tty`, which seems to fix any issues I have with signals not working as expected).

rustyrussell:
I wonder if docker is doing something strange: our code doesn't handle SIGTERM and thus should exit as per default signal handling.

We could just `signal(SIGTERM, SIG_DFL);` to be absolutely sure, but i am confused...

NicolasDorier:
Strange... I gave a try with https://github.com/btcpayserver/btcpayserver/blob/master/BTCPayServer.Tests/docker-compose.yml

```bash
docker-compose up customer_lightningd -d
docker-compose down -t 100
# Result: Shutdown immediately
```

Now I remove the SIGKILL

```diff
diff --git a/BTCPayServer.Tests/docker-compose.yml b/BTCPayServer.Tests/docker-compose.yml
index 7aba34f..c90d9a4 100644
--- a/BTCPayServer.Tests/docker-compose.yml
+++ b/BTCPayServer.Tests/docker-compose.yml
@@ -119,7 +119,6 @@ services:

   customer_lightningd:
     image: nicolasdorier/clightning:v0.6.2-3-dev
-    stop_signal: SIGKILL
     restart: unless-stopped
     environment:
       EXPOSE_TCP: "true"
```

```bash
docker-compose up customer_lightningd -d
docker-compose down -t 100
# Result: Hangs for 100 seconds.
```

I will take a look if by any chance this signal is not trapped by something in the `docker-entrypoint`.

NicolasDorier:
Ok, so this is a mistery for me at the moment...
As @cdecker, I tried `pkill -TERM lightningd` which work as it should...
But `docker-compose down` does not seem to send the TERM signal to clightning. Which is strange, first case I see. I am searching why.

NicolasDorier:
I am reading https://medium.freecodecamp.org/docker-entrypoint-cmd-dockerfile-best-practices-abc591c30e21 it seems it is a problem unrelated to this PR.

NicolasDorier:
Ok so here is why I seems fine in other images I use:

For some images, I don't execute a entrypoint script, meaning the process receive directly SIGTERM.

For clightning, I execute an entrypoint script, which execute bash with PID 1, which for some reason, block SIGTERM to reach clightning.

For other images (like [bitcoind](https://github.com/btcpayserver/dockerfile-deps/blob/master/Bitcoin/0.17.0/docker-entrypoint.sh)), I use `exec` which replace the bash process by the bitcoind process, making it receive the signal.

Proposed fix: using `exec` in the dockerfile, but I need to test if this will work even with `EXPOSE_TCP=true`. Ping @shesek I think your image on spark has same issue.

It seems that we don't need to handle `SIGTERM` explicitely in clightning after all.

NicolasDorier:
Nevermind @shesek handle it with a trap: https://github.com/shesek/spark-wallet/blob/master/scripts/docker-entrypoint.sh

NicolasDorier:
Ok @shesek hack of trapping TERM does not work, because TERM does not even reach the bash script because bash run with PID 1.

NicolasDorier:
I am reading this blog post which explain the problem quite well. https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

Playing with this now. Will make a PR once I find something that work.

NicolasDorier:
There is some linux black magic I don't quite understand yet.

I tried to use `exec lightningd` in the `docker-entrypoint.sh`, so `ligthningd` ends up with PID 1.

But then `pkill -TERM lightningd` does not work anymore!

NicolasDorier:
Surprisingly, this PR fix it!

| This PR merged | use `exec lightningd` | `pkill -TERM lightningd` works | `docker-compose down -t 100` terminate quick |
|:-:|:-:|:-:|:-:|
| ✔️ | ✔️ | ✔️ | ✔️ | 
| ✔️ | :x: | ✔️ | :x: |
| :x: | ✔️  | :x: | :x: | 
| :x: | :x:  | ✔️ | :x: | 


> A process running as PID 1 inside a container is treated specially by Linux: it ignores any signal with the default action. So, the process will not terminate on SIGINT or SIGTERM unless it is coded to do so.

Upon investigation, I think this PR should be merged. tACK.

NicolasDorier:
I was thinking about fixing the problem more cleanly via [dumb-init](https://github.com/Yelp/dumb-init). Because having this PR + `exec ligthningd` will work correctly if `EXPOSE_TCP` is `false`. `dumb-init` would solve for everything.

However, `EXPOSE_TCP` is only useful for dev time, where we don't really care about having short timeout for docker-compose. I don't like adding one dependency just for this corner case.

ZmnSCPxj:
Looking at `dumb-init/README.md`:

> However, if the process receiving the signal is PID 1, it gets special treatment by the kernel; if it hasn't registered a handler for the signal, the kernel won't fall back to default behavior, and nothing happens.

So it looks like all that is needed is to explicitly suicide when receiving SIGTERM and SIGINT, but only if we are PID 1.

If self-pipe trick is more than what we want to chew, we could do something simple like:

    int on_sigint(int _ UNUSED)
    {
            static const char *msg = "lightningd: SIGINT caught, exiting.\n";
            write(STDERR_FILENO, msg, strlen(msg));
            _exit(1);
    }
    int on_sigterm(int _ UNUSED)
    {
            static const char *msg = "lightningd: SIGTERM caught, exiting.\n";
            write(STDERR_FILENO, msg, strlen(msg));
            _exit(1);
    }

    ...

             if (1 == getpid()) {
                     signal(SIGINT, &on_sigint);
                     signal(SIGTERM, &on_sigterm);
             }

wythe:
I've been studying `ccan/io` along with your above links to create a proper solution.  It's definitely been a nice exercise to understand `ccan/io` and the clightning architecture better.

However I've been a bit distracted lately in RL.  I can implement you're above solution later today and then pursue the self-pipe solution on my own time.  I appreciate the help.

ZmnSCPxj:
We do not really need self-pipe trick; outside of docker we just not handle `SIGTERM` and `SIGINT` and depend on the fact that sqlite3 is 100% reliable and will never corrupt data file, because it is made of magic and not code.


Only need to handle `SIGTERM` and `SIGINT` for case that we are PID 1 of docker container, since, Linux makes special this PID and requires that some handler exist.

lightningd/lightningd.c
```diff
@@ -574,15 +574,49 @@ void notify_new_block(struct lightningd *ld, u32 block_height)
...
+{
+        static const char *msg = "lightningd: SIGTERM caught, exiting.\n";
+        write(STDERR_FILENO, msg, strlen(msg));
+        _exit(1);
+}
+
+/*~ Signal handlers for SIGINT and SIGTERM to print message to
+ * STDERR and exit. We only do this if we are PID 1. */
+static void setup_sig_handlers(struct sigaction *sigint,
+			       struct sigaction *sigterm)
```
ZmnSCPxj: Why not just declare `sigint` and `sigterm` structures as local here? Caller does not use them anyway.

wythe:
I also took your wording from above and used it for the code comment :-)

ZmnSCPxj:
@NicolasDorier if you can confirm this new version of PR is OK, I can ACK this.

@rustyrussell @cdecker @niftynei Should we be more ambitious and consider also proper signalling to child processes?
Currently, our child processes are "only" our own daemons, none of which maintain any persistent state and can be `SIGKILL`ed arbitrarily, but, in the future, more complex plugins may exist, which expect "correct" cleanup behavior when c-lightning docker is stopped.
I am almost tempted to just say "please put a minimal `init` system in Docker container".

NicolasDorier:
> I am almost tempted to just say "please put a minimal init system in Docker container". EDIT: Docker >=1.25 has --init option in run command which includes a minimal init system.

Yeah, or just use `SIGKILL`. Sadly, init has been supported on docker-compose version 3 since not too long (first supported in 2.2, then dropped, and added again), so I can't suppose the users running BTCPay support it.

Adding stop_signal SIGKILL is not too much of a burden, but I think it is better to make it harder for the docker image user to fuck up.

NicolasDorier:
tACK  9005a45

ZmnSCPxj:
ACK 9005a45

Still need some thought about taking on more `init` responsibilities (propagating `SIGINT`/`SIGTERM` to children, adopting orphans), or maybe just adding some minimal `init` implementation in the code base, or as a `git submodule`.

PR #2176, opened 2018-12-14
channeld: fix fee calculation.

rustyrussell:
Funder can't spend the fee it needs to pay for the commitment transaction:
we were not converting to millisatoshis, however!

This breaks our routeboost test, which no longer has sufficient funds to make payment.

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -414,7 +414,7 @@ static enum channel_add_err add_htlc(struct channel *channel,
 			- commit_tx_num_untrimmed(removing, feerate, dust,
 						  recipient);
 
-		fee_msat = commit_tx_base_fee(feerate, untrimmed);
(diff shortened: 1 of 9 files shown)


channeld/full_channel.c
```diff
@@ -414,7 +414,7 @@ static enum channel_add_err add_htlc(struct channel *channel,
 			- commit_tx_num_untrimmed(removing, feerate, dust,
 						  recipient);
 
-		fee_msat = commit_tx_base_fee(feerate, untrimmed);
+		fee_msat = commit_tx_base_fee(feerate, untrimmed) * 1000;
```
niftynei: should we annotate this method to denote the units it returns, ie `commit_tx_base_fee_sat`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2176/commits/13992dbd99bcb76884d5cafa69d1b04304bd1251

PR #2177, opened 2018-12-14
WIP: JSON internal cleanups

rustyrussell:
This ends with a major work to try to ensure we use command_fail successfully, by returning immediately.  Maybe it's too ugly to live though?


lightningd/payalgo.c
```diff
@@ -688,11 +684,12 @@ static void json_pay(struct command *cmd,
 	if (json_pay_try(pay))
 		command_still_pending(cmd);
 	else
-		return;
+		return command_its_complicated();
```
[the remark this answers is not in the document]
niftynei: you cleaned this up later. :+1:

common/test/run-param.c
```diff
@@ -457,8 +461,8 @@ static void advanced(void)
 		struct json *j = json_parse(cmd, "[ 3 'foo' ]");
```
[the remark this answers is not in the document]
rustyrussell: Interesting: that code was unchanged...

lightningd/invoice.c
```diff
@@ -570,6 +568,8 @@ static void json_waitanyinvoice(struct command *cmd,
 	/* Find next paid invoice. */
 	wallet_invoice_waitany(cmd, wallet, *pay_index,
 			       &wait_on_invoice, (void*) cmd);
+
+	return command_its_complicated();
```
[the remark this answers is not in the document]
rustyrussell: I was aiming for Facebook-style coyness. Guess I failed: will add extra relationship details here :)

lightningd/json.c
```diff
@@ -143,7 +143,7 @@ bool json_tok_short_channel_id(struct command *cmd, const char *name,
 
 	command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 		     "'%s' should be a short channel id, not '%.*s'",
```
wythe:
```
lightning-cli check listchannels foo
{ "code" : -32602, "message" : "'short_channel_id' should be a short channel id, not '\"foo\"'" }
```
Since we are now properly escaping double quotes, then surrounding them with single quotes may not be necessary anymore (as in `'\"foo\"'` above).  And then maybe change `'short_channel_id'` to `\"short_channel_id\"` also?

Of course this would affect many places in the code.

rustyrussell: Generally I find ' quotes clearer for errors like this, TBH, so I think we leave it.  I like the quotes in case spaces sneak in, too.

wythe: 👍

lightningd/jsonrpc.h
```diff
@@ -127,7 +127,7 @@ static inline void fixme_ignore(const struct command_result *res)
 }
 
 /* FIXME: For the few cases where return value is indeterminate */
```
niftynei: is the FIXME still valid, if you've annotated them?

common/test/run-param.c
```diff
@@ -458,7 +458,7 @@ static void advanced(void)
 		assert(msat_opt2 == NULL);
 	}
 	{
-		struct json *j = json_parse(cmd, "[ 3 'foo' ]");
+		struct json *j = json_parse(cmd, "[ 3, 'foo' ]");
```
niftynei: :grin:

lightningd/jsonrpc.h
rustyrussell: It's still a bad idea, though.  Just because I don't know how to fix it right now doesn't make it less of a FIXME :)

PR #2178, opened 2018-12-15
mocks: Fix the mocks generation fix

cdecker:
Turns out that I should have tested these with a new dependency
instead of just submitting. `sed` was missing the s command :slightly_frowning_face:


rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/d7062cc07d5ee3a72536355a75150c47a2b0f890

PR #2185, opened 2018-12-17
ccan: update to latest version, which includes more take() checks.

rustyrussell:
And add the missing unit test from my previous update.


cdecker:
ACK 925464e280fc583ed71e4b4af91f251b30950824

PR #2187, opened 2018-12-18
update pytest check when configure

arowser:
If no pytest in system:
1. The configure show importError, its better to hide this message.
```
 ./configure
  File "<string>", line 1, in <module>
ImportError: No module named 'pytest'
```
2. configure print PYTEST is null.
```
Setting COMPAT... 1
Setting PYTEST... 
Setting STATIC... 0
```

So this PR hide the pytest check message, and set PYTEST var to 0 if not find pytest.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -230,7 +230,7 @@ check:
 	$(MAKE) pytest
 
 pytest: $(ALL_PROGRAMS)
-ifndef PYTEST
+ifeq ($(PYTEST),)
 	@echo "py.test is required to run the integration tests, please install using 'pip3 install -r tests/requirements.txt', and rerun 'configure'."
 	exit 1
 else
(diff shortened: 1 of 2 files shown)


rustyrussell:
Hmm, '0' is weird here.  Maybe leave it empty, print "not found" in add_var in this case, and fix the Makefile from "ifndef PYTEST" to "ifeq ($(PYTEST),)"?

arowser:
Its makes sense, updated

rustyrussell:
Doesn't quite work, this will set PYTEST to "not found".  Remove that, and add a check in
add_var:

    if [ -n "$2" ]; then
        echo "Setting $1... $2"
    else
        echo "$1 not found"
    fi

?

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2187/commits/dfb2934523f3a6ea87aedc8531f4c55f4201b15c

PR #2188, opened 2018-12-18
plugin: Add notifications from lightningd to plugins (Plugin Saga, 3rd reprise)

cdecker:
We're nearing the end of the endless plugin saga, with this new feature. It allows plugins to subscribe to a number of notifications/events that will then be streamed from `lightningd` to the plugin when something happens.

Notice that, while this provides the infrastructure, I've only implemented two very simple notification topics (`connect` and `disconnect`). We can (and will) add more as we come up with new use-cases.

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -24,5 +24,15 @@ def init(options, configuration, plugin):
     plugin.log("Plugin helloworld.py initialized")
 
 
+@plugin.subscribe("connect")
+def on_connect(id, address, plugin):
+    plugin.log("Received connect event for peer {}".format(id))
+
+
+@plugin.subscribe("disconnect")
+def on_disconnect(id, plugin):
+    plugin.log("Received disconnect event for peer {}".format(id))
+
+
 plugin.add_option('greeting', 'Hello', 'The greeting I should use.')
 plugin.run()
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -1,0 +1,13 @@
+#include "lightningd/notification.h"
+#include <ccan/array_size/array_size.h>
+
+const char *notification_topics[] = {
+};
+
+bool notifications_have_topic(const char *topic)
+{
+	for (size_t i=0; ARRAY_SIZE(notification_topics); i++)
+		if (streq(topic, notification_topics[i]))
+			return true;
+	return false;
+}
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -2,6 +2,8 @@
 #include <ccan/array_size/array_size.h>
 
 const char *notification_topics[] = {
+	"connect",
+	"disconnect",
 };
 
 bool notifications_have_topic(const char *topic)
@@ -10,4 +12,24 @@ bool notifications_have_topic(const char *topic)
 		if (streq(topic, notification_topics[i]))
 			return true;
 	return false;
+}
+
+void notify_connect(struct lightningd *ld, struct pubkey *nodeid,
+		    struct wireaddr_internal *addr)
+{
+	struct jsonrpc_notification *n =
+	    jsonrpc_notification_start(NULL, notification_topics[0]);
+	json_add_pubkey(n->stream, "id", nodeid);
+	json_add_address_internal(n->stream, "address", addr);
+	jsonrpc_notification_end(n);
+	plugins_notify(ld->plugins, take(n));
+}
+
+void notify_disconnect(struct lightningd *ld, struct pubkey *nodeid)
+{
+	struct jsonrpc_notification *n =
+	    jsonrpc_notification_start(NULL, notification_topics[1]);
+	json_add_pubkey(n->stream, "id", nodeid);
+	jsonrpc_notification_end(n);
+	plugins_notify(ld->plugins, take(n));
 }
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -18,6 +18,7 @@
 #include <errno.h>
 #include <lightningd/json.h>
 #include <lightningd/lightningd.h>
+#include <lightningd/notification.h>
 #include <signal.h>
 #include <sys/stat.h>
 #include <sys/types.h>
@@ -60,6 +61,9 @@ struct plugin {
 	/* Timer to add a timeout to some plugin RPC calls. Used to
 	 * guarantee that `getmanifest` doesn't block indefinitely. */
 	const struct oneshot *timeout_timer;
+
+	/* An array of subscribed topics */
+	char **subscriptions;
 };
 
 struct plugin_request {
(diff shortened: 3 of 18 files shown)


contrib/plugins/helloworld.py
```diff
@@ -24,5 +24,15 @@ def init(options, configuration, plugin):
     plugin.log("Plugin helloworld.py initialized")
 
 
+@plugin.subscribe("connect")
```
renepickhardt: this is an awesome API!

lightningd/plugin.c
```diff
@@ -1007,3 +1057,27 @@ void json_add_opt_plugins(struct json_stream *response,
...
+					  const char *method)
+{
+	for (size_t i = 0; i < tal_count(plugin->subscriptions); i++)
+		if (streq(method, plugin->subscriptions[i]))
+			return true;
+
+	return false;
+}
+
+void plugins_notify(struct plugins *plugins,
```
renepickhardt: LGTM ( I was searching if subscriptions to the same events would work if several plugins subscribed to the same event

lightningd/notification.c
```diff
@@ -0,0 +1,35 @@
...
+
+bool notifications_have_topic(const char *topic)
+{
+	for (size_t i=0; ARRAY_SIZE(notification_topics); i++)
+		if (streq(topic, notification_topics[i]))
+			return true;
+	return false;
+}
+
+void notify_connect(struct lightningd *ld, struct pubkey *nodeid,
```
renepickhardt: so just for the gist I could create a method `void notify_incoming_htlc(struct lightningd *ld, struct pubkey *nodeid, struct channel *chan, struct htlc *h)` (not sure if `struct channel` and `struct htlc` exist) and create a json string and send it via plugin_notify to the plugin.  I would obviously have to call that function at the place where incoming htlc's are being handled. But other than that I would not have to extend other parts of c-lightning or am I missing something?

cdecker: Yes, the idea is that you'd create a `notify_xyz` function that gathers all the information and formats it into a JSON message and then passes that to `plugins_notify` which will take care of dispatching notifications to all subscribers.

lightningd/plugin.c
cdecker: That is indeed supported, since unlike jsonrpc calls and hooks there is no problem with notification semantics.

contrib/plugins/helloworld.py
cdecker: Thanks, Flask was a huge inspiration.

cdecker:
I will push the new version of the pylightning library once we have the `@plugin.hook` functionality in there. I don't like churning through versions too quickly, especially when this is functionality that is as of now unreleased.

Should get a 0.0.7 out by the end of the week though.

lightningd/plugin.c
```diff
@@ -770,6 +774,47 @@ static bool plugin_rpcmethods_add(struct plugin *plugin,
...
+		const jsmntok_t *s = json_get_arr(subscriptions, i);
+		if (s->type != JSMN_STRING) {
+			plugin_kill(
+			    plugin,
+			    "result.subscriptions[%d] is not a string: %s", i,
+			    plugin->buffer);
+			return false;
+		}
+		topic = tal_strndup(
+		    plugin, plugin->buffer + s->start, s->end - s->start);
```
rustyrussell: Note: after rebase onto master, you'll be able to use `json_strdup(plugin, plugin->buffer, s);`

lightningd/plugin.c
```diff
@@ -770,6 +774,47 @@ static bool plugin_rpcmethods_add(struct plugin *plugin,
...
+
+		if (!notifications_have_topic(topic)) {
+			plugin_kill(
+			    plugin,
+			    "topic %s is not a know notification topic", topic);
```
rustyrussell: s/know/known/

rustyrussell: Slight simplification, I would skip the 's->type != JSMN_STRING' check above, and rely on it here (in which case, please put  ' around the %s).

lightningd/plugin.c
```diff
@@ -229,12 +229,11 @@ plugin_request_new_(struct plugin *plugin,
...
 /**
- * Given a request, send it to the plugin.
+ * Send a JSON-RPC message (request or notification) to the plugin.
  */
-static void plugin_request_queue(struct plugin *plugin,
-				 struct plugin_request *req)
+static void plugin_send(struct plugin *plugin, struct json_stream *stream TAKES)
 {
-	*tal_arr_expand(&plugin->js_arr) = req->stream;
+	*tal_arr_expand(&plugin->js_arr) = tal_steal(plugin->js_arr, stream);
```
rustyrussell: Make this two lines: we got bitten by this evaluation order indeterminism before.

lightningd/plugin.c
```diff
@@ -229,12 +229,11 @@ plugin_request_new_(struct plugin *plugin,
 	    (arg))
 
 /**
- * Given a request, send it to the plugin.
+ * Send a JSON-RPC message (request or notification) to the plugin.
  */
-static void plugin_request_queue(struct plugin *plugin,
-				 struct plugin_request *req)
+static void plugin_send(struct plugin *plugin, struct json_stream *stream TAKES)
```
rustyrussell: TAKES isn't quite right here: that means it takes ownership iff caller calls take().  Perhaps call it something more obviously "takes-ownership"-is, like "plugin_attach_stream()"?

lightningd/plugin.c
```diff
@@ -1057,3 +1057,27 @@ void json_add_opt_plugins(struct json_stream *response,
...
+
+void plugins_notify(struct plugins *plugins,
+		    const struct jsonrpc_notification *n TAKES)
+{
+	struct plugin *p;
+	list_for_each(&plugins->plugins, p, list) {
+		if (plugin_subscriptions_contains(p, n->method))
+			plugin_send(p, json_stream_dup(p, n->stream));
+	}
+	tal_free(n);
```
rustyrussell:
```
if (taken(n))
    tal_free(n);
```

We could of course optimize this; when we optimize to a per-topic queue of plugins, we'd not have to copy the last one if it's taken().

lightningd/notification.c
```diff
@@ -2,6 +2,8 @@
 #include <ccan/array_size/array_size.h>
 
 const char *notification_topics[] = {
+	"connect",
```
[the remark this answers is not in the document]
cdecker: I bikeshedded a couple of variants for this, with broader topics, and naming variants, but in the end I just picked the simplest and shortest variant I could think of, i.e., just the actions that are performed. Happy to discuss the naming here further, but then I'd split the addition of this topic out and just get the infrastructure merged, so I can continue working on top of this :wink:

lightningd/notification.c
```diff
@@ -11,3 +13,23 @@ bool notifications_have_topic(const char *topic)
...
+	json_add_pubkey(n->stream, "id", nodeid);
+	json_add_address_internal(n->stream, "address", addr);
+	jsonrpc_notification_end(n);
+	plugins_notify(ld->plugins, n);
+}
+
+void notify_disconnect(struct lightningd *ld, struct pubkey *nodeid)
+{
+	struct jsonrpc_notification *n =
+	    jsonrpc_notification_start(NULL, notification_topics[1]);
```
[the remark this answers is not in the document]
cdecker: There is a facility to generate names from `enum`s, maybe we want to use that instead? I had an `enum` that represented the indices into this array, but it got really repetitive, so I dropped it for now.

cdecker:
Rebased and addressed the feedback in fixups. @rustyrussell and @niftynei thanks for the reviews, if you're happy with the fixups I'll squash and merge :-)

P.S.: I know it looks like a lot of fixups, but they are all single-lines and address a single comment each to make reviewing easier :wink:

lightningd/notification.c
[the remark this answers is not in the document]
ZmnSCPxj: We use `AUTODATA` to generate an array of conmands, cannot we use same to generate array of notifications?

lightningd/notification.c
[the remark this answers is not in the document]
niftynei: sounds good. :+1: 

fwiw, i like how the 'actor_action' pattern (ie 'node_connected') gives a bit more insight at a glance as to what this refers to, but being as i'm more familiar with the codebase now, `connect` is pretty easy to map back to RPC commands, which is also nice :)

lightningd/notification.c
```diff
@@ -11,3 +13,23 @@ bool notifications_have_topic(const char *topic)
 			return true;
 	return false;
 }
+
+void notify_connect(struct lightningd *ld, struct pubkey *nodeid,
+		    struct wireaddr_internal *addr)
+{
+	struct jsonrpc_notification *n =
+	    jsonrpc_notification_start(NULL, notification_topics[0]);
+	json_add_pubkey(n->stream, "id", nodeid);
```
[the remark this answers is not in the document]
niftynei: Ack. I totally misread some python code; I redact this comment.

conscott:
Tested ACK 91c9ce760dfe19530d5b2981467607a31f830035

Just left a minor nit above.

cdecker:
Rebased, dropped the `id` -> `peer_id` fixups, added a docstring squash fixup as suggested by @rustyrussell :wink:

cdecker:
Rebased and squashed on top of `master`. Reapplying @rustyrussell's ACK (@bitcoin-bot) is still having issues recognizing rebases and squashes)

ACK https://github.com/ElementsProject/lightning/pull/2188/commits/5a8c2c6eb6c4487a82b2f637a3d1fa12c3bba74e

PR #2193, opened 2018-12-20
[Docker] Make sure lightningd receives SIGTERM

NicolasDorier:
In conjunction to https://github.com/ElementsProject/lightning/pull/2172 this fix https://github.com/ElementsProject/lightning/issues/2074 if EXPOSE_TCP is false (which is the case in production)

More info https://github.com/ElementsProject/lightning/pull/2172#issuecomment-447727668


ZmnSCPxj:
ACK 122addd

PR #2194, opened 2018-12-20
Do not produce bad onions on malformed

rustyrussell:
I was writing a pay test when I got a non-decodable onion.  Turns out our handling of "update_fail_malformed_htlc" (where the next peer can't decrypt the onion) is completely wrong, and results in us packaging up incorrect data.  This is easy to duplicate by simply lying to 'sendpay' about what id the next hop is.

The resulting cleanups and fixes solve this.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -529,7 +529,8 @@ static void handle_peer_announcement_signatures(struct peer *peer, const u8 *msg
 
 static struct secret *get_shared_secret(const tal_t *ctx,
 					const struct htlc *htlc,
-					enum onion_type *why_bad)
+					enum onion_type *why_bad,
+					struct sha256 *next_onion_sha)
 {
 	struct pubkey ephemeral;
 	struct onionpacket *op;
@@ -558,6 +559,10 @@ static struct secret *get_shared_secret(const tal_t *ctx,
 		*why_bad = WIRE_INVALID_ONION_HMAC;
 		return tal_free(secret);
 	}
+
+	/* Calculate sha256 we'll hand to next peer, in case they complain. */
+	msg = serialize_onionpacket(tmpctx, rs->next);
+	sha256(next_onion_sha, msg, tal_bytelen(msg));
 
 	return secret;
 }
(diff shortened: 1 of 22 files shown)


lightningd/peer_htlcs.c
```diff
@@ -640,35 +640,36 @@ static bool peer_accepted_htlc(struct channel *channel,
...
+	if (!hin->shared_secret) {
+		*failcode = WIRE_INVALID_ONION_KEY;
 		goto out;
 	}
 
-	/* Channeld sets this to zero if HSM won't ecdh it */
-	if (memeqzero(&hin->shared_secret, sizeof(hin->shared_secret))) {
-		*failcode = WIRE_INVALID_ONION_KEY;
-		goto out;
+	/* FIXME: Have channeld hand through the just the route_step! */
```
cdecker:
```suggestion
 	/* FIXME: Have channeld hand through just the route_step! */
```

channeld/channeld.c
```diff
@@ -1602,12 +1601,16 @@ static void handle_peer_fail_malformed_htlc(struct peer *peer, const u8 *msg)
 	 *    `update_fail_malformed_htlc`:
 	 *      - MUST fail the channel.
 	 */
-	if (!(failure_code & BADONION)) {
+	/* We only handle these cases. */
```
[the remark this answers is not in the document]
rustyrussell: OK, I converted unknown ones to temporary_failure.

There are two channelds involved here.  This path is the outgoing one: it sets failcode to one of these three values.  The incoming channeld turns that failcode into a message.  It only knows how to handle these three types: for any new type it would hit this code in make_failmsg:

		      "Asked to create failmsg %u (%s)",
		      failcode, onion_type_name(failcode));

which would cause failure of the *incoming* channel.

lightningd/peer_htlcs.c
rustyrussell: Fixed, thanks.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2194/commits/73822309cad0180a86f4e0351d4a45371331cf41

PR #2197, opened 2018-12-21
Manpage for lightning-listchannels created

runcrypto-michael:
Created lightning-listchannel manpage in asciidoc and generated manpage using a2x.  Updated doc/Makefile.


runcrypto-michael:
This work is a continuation of that on issue https://github.com/ElementsProject/lightning/issues/1360

cdecker:
I took the liberty of rebasing this onto `master` removing the merge commits.

doc/lightning-listchannels.7.txt
```diff
@@ -0,0 +1,65 @@
+LIGHTNING-LISTCHANNELS(7)
+=======================
+:doctype: manpage
+
+NAME
+----
+lightning-listchannels - Command for returning data on active lightning
+channels..
```
cdecker: nit: double `.` at the end.

cdecker: I also prefer "Command to query active lightning channels in the entire network." since it is for all known, not just our own channels.

doc/lightning-listchannels.7.txt
```diff
@@ -0,0 +1,65 @@
...
+DESCRIPTION
+-----------
+The *listchannels* RPC command returns data on channels that are known to the
+node.
+
+If no 'short_channel_id' is supplied, then data on all lightning channels that
+are known to this node, are returned.
+
+Supplying 'short_channel_id' will filter the results to only return data on a
+channel with a matching 'short_channel_id', if one exists.
```
cdecker: We could add a mention that channel directions are considered independently, hence even specifying a `short_channel_id` might results in 0, 1 or 2 resulting channel directions being returned.

doc/lightning-listchannels.7.txt
```diff
@@ -0,0 +1,65 @@
...
+- 'satoshis' - 
+- 'message_flags' - 
+- 'channel_flags' - 
+- 'active' - 					
+- 'last_update' -
+- 'base_fee_millisatoshi' -
+- 'fee_per_millionth' -
+- 'delay' -
+
+If 'shcort_channel_id' is supplied and no matching channels are found, a
```
cdecker: `shcort_channel_id` -> `short_channel_id`

doc/lightning-listchannels.7.txt
```diff
@@ -28,29 +28,39 @@ On success, an object with a "channels" key is returned containing a list of 0
...
-- 'channel_flags' - 
-- 'active' - 					
-- 'last_update' -
-- 'base_fee_millisatoshi' -
-- 'fee_per_millionth' -
-- 'delay' -
-
-If 'shcort_channel_id' is supplied and no matching channels are found, a
+
+- 'source' : The node that initiated the channel creation
```
cdecker: This is not really true, the source and destination are defined by the direction in which we'd want to traverse the channel. We can't determine who the funder is. The source is the node that specified the fees and CLTV deltas if we want to traverse the channel in this direction.

This is also why we could get 2 results for the 2 different directions when specifying a `short_channel_id`.

doc/lightning-listchannels.7.txt
```diff
@@ -28,29 +28,39 @@ On success, an object with a "channels" key is returned containing a list of 0
...
+- 'destination' : The node that was the target of the channel creation
+- 'short_channel_id' : The channel identifier
+- 'public' : Boolean value, is publicly available
```
cdecker: This will only ever be non-true on local channels and distinguishes whether the channel was received through a `channel_announcement` or whether we side-loaded it by being one of the endpoints. Might be interesting information.

doc/lightning-listchannels.7.txt
```diff
@@ -28,29 +28,39 @@ On success, an object with a "channels" key is returned containing a list of 0
...
+- 'satoshis' : The amount of funds available in the channel
+- 'message_flags' : Bitfield showing the presence of optional fields in the
+'channel_update' message (BOLT #7)
+- 'channel_flags' : Bitfield showing the end of the channel the node is on
```
cdecker: Well, that's exactly the direction bit and the disable bit. Would it maybe be better to be less verbose and just referring them to BOLT 07?

doc/lightning-listchannels.7.txt
```diff
@@ -28,29 +28,39 @@ On success, an object with a "channels" key is returned containing a list of 0
...
 "channels" object with an empty list is returned.
 
 ERRORS
 ------
 If 'short_channel_id' is not a valid short_channel_id, an error message will be
 returned:
 
-	{ "code" : -32602, "message" : "'short_channel_id' should be a short
-	channel id, not '...'" }
+{ "code" : -32602, "message" : "'short_channel_id' should be a short channel
```
cdecker: Is there maybe a way to markup this as a code excerpt instead of free text?

runcrypto-michael:
Thanks for the prompt feedback, I'll look to get the changes implemented over the next week or so.   I'm still a bit green when it comes to submitting pull requests so thanks too for your patience on merging misdemeanors 👍 
Merry Christmas!

runcrypto-michael:
Messed up my remote branch trying to remove commit history so will have to submit this again

PR #2198, opened 2018-12-21
fix crash on check with plugin command

rsbondi:
When issuing command `check PLUGIN_COMMAND` from the CLI or through RPC, the lightningd daemon crashes, using the helloworld example, PLUGIN_COMMAND would be "hello"


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2198/commits/a9f8d16e37c02c2604390c4ab33a78945f60b4ea

PR #2205, opened 2018-12-28
pylightning: Increase buffer size for big JSON-RPC responses

cdecker:
This was causing `listchannels` to be incredibly slow. The response is
several megabyte in size, and we were only buffering 1Kb on each
iteration.

We can be at most a factor of 2 away from what was actually needed, and it considerably reduces parse time.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2205/commits/c72ddcb8611f353c571a54b939244949a3bf0d38

PR #2207, opened 2018-12-30
Add funder to listpeers command

conscott:
This change adds a field to the `listpeers` command called `funder`, which is a string set to either `"LOCAL"` or `"REMOTE"` based on who initiated the channel open. 

I having been playing with some plugins to show more advanced channel stats and it seems useful to know who opened the channel. If there is another way to know this from current info in listpeers, please let me know.


cdecker:
Since we're working on the spec for dual-funding, this field might become obsolete rather quickly. That is undesirable since it is exposed to users of the API and therefore people might start building on top of this.

Maybe we can make this a map that maps the endpoint to contributed amounts or initial settlement amounts? That way we can make this `{"<funder_id>": <total_msatoshi>, "<fundee_id>": 0}` for now and switch it to actual values once we have dual-funding.

PR #2209, opened 2018-12-30
build: fix a few more compile errors

jb55:
I have a few patches that may be of interest.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -901,6 +901,7 @@ static struct command_result *json_close(struct command *cmd,
 {
 	const jsmntok_t *idtok;
 	struct peer *peer;
+	/* FIXME: gcc 7.3.0 thinks this might not be initialized. */
 	struct channel *channel = NULL;
 	unsigned int *timeout;
 	bool *force;


lightningd/lightningd.c
```diff
@@ -574,17 +574,17 @@ void notify_new_block(struct lightningd *ld, u32 block_height)
 	channel_notify_new_block(ld, block_height);
 }
 
-static void on_sigint(int _ UNUSED)
+static void on_sigint(int _)
 {
         static const char *msg = "lightningd: SIGINT caught, exiting.\n";
-        write(STDERR_FILENO, msg, strlen(msg));
+        _ = write(STDERR_FILENO, msg, strlen(msg));
```
rustyrussell: I prefer to fix this using "write_all" which specifically doesn't have this problem.

Or at least put in a snarky reference to glibc+gcc braindamage a-la https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425 :)

lightningd/peer_control.c
```diff
@@ -901,7 +901,7 @@ static struct command_result *json_close(struct command *cmd,
 {
 	const jsmntok_t *idtok;
 	struct peer *peer;
-	struct channel *channel;
+	struct channel *channel = NULL;
```
rustyrussell: Not sure what compiler you're using, but does putting "assert(channel);" after:
```
		res = command_find_channel(cmd, buffer, idtok, &channel);
		if (res)
			return res;
```

also suppress the warning?  That seems clearer to me...

jb55:
Rusty Russell <notifications@github.com> writes:

> rustyrussell commented on this pull request.
>
> Thanks for catching these!  What compiler are you using?

gcc 7.3.0

> Not sure what compiler you're using, but does putting "assert(channel);" after:
> ```
> 		res = command_find_channel(cmd, buffer, idtok, &channel);
> 		if (res)
> 			return res;
> ```
>
> also suppress the warning?  That seems clearer to me...

sure, but if channel had uninitialized/garbage data wouldn't it pass the assert?

I tried with the assert in a few places, it still gives me the compile error.

rustyrussell:
> sure, but if channel had uninitialized/garbage data wouldn't it pass the assert? I tried with the assert in a few places, it still gives me the compile error.

gcc 7.3 isn't smart enough to know that channel is always initialized if `res` is NULL. (8.2 gets this right, hence no warning).  I'll apply this with a comment, thanks.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2209/commits/5d6f7f14ba8c72753c3e64086fa752504f1693e4

PR #2210, opened 2019-01-02
Removed duplicate error message in build

### LINKED ISSUE #2130: Configure error printed twice
Trying to make without using `./configure` prints out a funny error / suggestion:
```
make
The 1990s are calling: use ./configure!
The 1990s are calling: use ./configure!
```
I guess this string should be printed only once?

lucash-dev:
Changed `Makefile` to check only once for `config.vars`, instead of twice.
Also changed the funny message about the 1990s to something more informative (the new message is quite boring though).

I think it fixes #2130.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,9 +188,12 @@ endif
 
 default: all-programs all-test-programs
 
-config.vars ccan/config.h: configure ccan/tools/configurator/configurator.c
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2210/commits/eb4c1006e1eec98c7fa48ac8c1e8176b124b4847

PR #2211, opened 2019-01-02
changed import statement for plugins

renepickhardt:
example script will not run if the `plugin.py` file from the `lightning` directory is not loaded.

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from lightning import Plugin
+from lightning.plugin import Plugin
 
 
 plugin = Plugin(autopatch=True)


cdecker:
This line should theoretically fix this issue by exposing the `Plugin` class in the `lightning` module directly:

https://github.com/ElementsProject/lightning/blob/e6c68b606d02505bed6adfbbf8225edb84479723/contrib/pylightning/lightning/__init__.py#L2

The version on pipy is not up to date, and should not be used for plugin development until we release c-lightning. If you are running c-lighthning from the repo, then you should also use the pylightning version that comes bundled with it, since we have no compatibility guarantees on unreleased versions.

renepickhardt:
I have used the version that comes with c-lightning by running `python3 setup.py  install` and I had the problem on macos with my anaconda python distribution. after fixing it ran smoothly

cdecker:
`python setup.py install` will take the current version and freeze it. `python setup.py develop` will take the library and create symlinks instead, which means that (unless you create new files) it'll always be up to date.

Personally I just set `PYTHONPATH=contrib/pylightning` in the directory when working there, that avoids having to touch the python environment altogether.

renepickhardt:
thanks for this explanation. In any case it does not explain why I run into problems without this fix. I will try to reproduce this and create an issue if I am able to

PR #2212, opened 2019-01-02
[plugin] Improved the example code to encourage usage of optional arguments

renepickhardt:
The example code had the `plugin` argument as the last argument. this disallows arguments that have a standard value. As far as I understand the dispatching code the order of arguments does not matter since it is the name `plugin` that is relevant. Therefor I changed the order so that newbe's don't have to read the entire code and can easily add optional arguments

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -6,7 +6,7 @@ plugin = Plugin(autopatch=True)
 
 
 @plugin.method("hello")
-def hello(name, plugin):
+def hello(plugin,name = "world"):
     """This is the documentation string for the hello-function.
 
     It gets reported as the description when registering the function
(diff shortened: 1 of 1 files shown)


contrib/plugins/helloworld.py
```diff
@@ -6,7 +6,7 @@
 
 
 @plugin.method("hello")
-def hello(name, plugin):
+def hello(plugin,name = "world"):
```
cdecker:
```suggestion
def hello(plugin, name="world"):
```

cdecker: I think you can apply these suggestions somehow, never tried them before :wink:

renepickhardt:
I have accepted your in git addition. thus it should be ready to be merged.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2212/commits/4db3f476afb20523ca00733579cd272dedda1043

PR #2214, opened 2019-01-03
lightningd: don't allow zero cltv HTLCs.

rustyrussell:
Seems like someone is sending out zero-cltv HTLCs, which is we weren't catching correctly, thus failing on a sanity check.

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -198,7 +198,7 @@ static bool check_amount(struct htlc_in *hin,
 static bool check_cltv(struct htlc_in *hin,
 		       u32 cltv_expiry, u32 outgoing_cltv_value, u32 delta)
 {
-	if (cltv_expiry - delta >= outgoing_cltv_value)
+	if (delta < cltv_expiry && cltv_expiry - delta >= outgoing_cltv_value)
 		return true;
 	log_debug(hin->key.channel->log, "HTLC %"PRIu64" incorrect CLTV:"
 		  " %u in, %u out, delta reqd %u",


cdecker:
Shouldn't this check have caught that exact problem?

https://github.com/ElementsProject/lightning/blob/0321d79540e4d2ef5296c4717cae390321b3b7d5/lightningd/peer_htlcs.c#L524-L531

I don't see how a 0 CLTV could get past that check.

cdecker:
CI failure was a flake btw, re-running.

rustyrussell:
> Shouldn't this check have caught that exact problem?
> 
> [lightning/lightningd/peer_htlcs.c](https://github.com/ElementsProject/lightning/blob/0321d79540e4d2ef5296c4717cae390321b3b7d5/lightningd/peer_htlcs.c#L524-L531)
> 
> Lines 524 to 531 in [0321d79](/ElementsProject/lightning/commit/0321d79540e4d2ef5296c4717cae390321b3b7d5)
>  if (get_block_height(ld->topology) >= outgoing_cltv_value) { 
>  	log_debug(hin->key.channel->log, 
>  		  "Expiry cltv %u too close to current %u", 
>  		  outgoing_cltv_value, 
>  		  get_block_height(ld->topology)); 
>  	failcode = WIRE_EXPIRY_TOO_SOON; 
>  	goto fail; 
>  } 
> 
> I don't see how a 0 CLTV could get past that check.

That's examining the CLTV inside the onion, not the incoming CLTV value.

What seems to be happening is that the onion is valid, but the previous peer decided to offer an HTLC with ctlv of 0.

cdecker:
Oh, that makes sense, my head was hurting a bit with the deltas and the various CLTVs :-)

ACK https://github.com/ElementsProject/lightning/pull/2214/commits/de2569200d43169caebc0d285e1277816dffd2c5

PR #2218, opened 2019-01-03
[plugin] Made rpc_interface a public member of Plugin class and sorted imports

renepickhardt:
after this code change people can use `plugin.rpc_interface()` from anywhere in their plugin code this is much nicer than going this way within the init function (or somewhere else):
```
@plugin.method("init")
    global rpc_interface
    basedir = plugin.lightning_dir 
    rpc_filename = plugin.rpc_filename 
    path = join(basedir, rpc_filename)
    rpc_interface = LightningRpc(path)
```

or similarly that way: 
```
@plugin.method("init")
    global rpc_interface
    basedir = configuration['lightning-dir']
    rpc_filename = configuration['rpc-file']
    path = join(basedir, rpc_filename)
    rpc_interface = LightningRpc(path)
```

also the imports have been sorted alphabetically

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -1,10 +1,11 @@
 from collections import OrderedDict
+from lightning import LightningRpc
 
-import sys
-import os
-import json
 import inspect
+import json
+import os
 import re
+import sys
 import traceback
 
 
@@ -35,6 +36,7 @@ class Plugin(object):
         self.add_method("getmanifest", self._getmanifest)
         self.rpc_filename = None
         self.lightning_dir = None
+        self.rpc = None
         self.init = None
 
     def add_method(self, name, func):
@@ -283,6 +285,8 @@ class Plugin(object):
     def _init(self, options, configuration, request):
         self.rpc_filename = configuration['rpc-file']
         self.lightning_dir = configuration['lightning-dir']
+        path = os.path.join(self.lightning_dir, self.rpc_filename)
+        self.rpc = LightningRpc(path)
         for name, value in options.items():
             self.options[name]['value'] = value
 


cdecker:
Great point, I'm happy to merge this, however I'd prefer calling it `rpc` instead of `rpc_interface` which is way too verbose for my taste. In addition we could make the `rpc`-member of the plugin a `@property` which would allow us to raise an exception if it gets called before the `init` call comes in. That could be more helpful than just returning `None` imho.

cdecker:
Oh, and it needs squashing :wink:

contrib/pylightning/lightning/plugin.py
```diff
@@ -1,10 +1,13 @@
...
-import json
 import inspect
+import json
+import os
 import re
+import sys
 import traceback
 
+from .lightning import LightningRpc
+from os.path import join
```
cdecker: I prefer to import `os` or `os.path` since just importing `join` is a bit ambiguous as to what join operation this actually is. Also we already have `os` imported so why not just use that import?

contrib/pylightning/lightning/plugin.py
```diff
@@ -33,6 +36,7 @@ def __init__(self, stdout=None, stdin=None, autopatch=True):
         self.add_method("getmanifest", self._getmanifest)
         self.rpc_filename = None
         self.lightning_dir = None
+        self.rpc_interface = None
```
cdecker: I'd prefer calling this `rpc`

cdecker:
```suggestion
        self.rpc = None
```

contrib/pylightning/lightning/plugin.py
```diff
@@ -263,6 +267,8 @@ def _getmanifest(self):
     def _init(self, options, configuration, request):
         self.rpc_filename = configuration['rpc-file']
         self.lightning_dir = configuration['lightning-dir']
+        path = join(self.lightning_dir, self.rpc_filename)
+        self.rpc_interface = LightningRpc(path)
```
cdecker:
```suggestion
        self.rpc = LightningRpc(path)
```

contrib/pylightning/lightning/plugin.py
```diff
@@ -263,6 +267,8 @@ def _getmanifest(self):
     def _init(self, options, configuration, request):
         self.rpc_filename = configuration['rpc-file']
         self.lightning_dir = configuration['lightning-dir']
+        path = join(self.lightning_dir, self.rpc_filename)
```
cdecker:
```suggestion
        path = os.path.join(self.lightning_dir, self.rpc_filename)
```

contrib/pylightning/lightning/plugin.py
```diff
@@ -1,10 +1,13 @@
...
-import os
-import json
 import inspect
+import json
+import os
 import re
+import sys
 import traceback
 
+from .lightning import LightningRpc
```
cdecker:
```suggestion
from lightning import LightningRpc
```

renepickhardt:
yeah just have to finnish the changed requests from cdecker

renepickhardt:
thank for feedback @conscott I think I should have applied all changes from @cdecker so it should be ready to merge. Though it is still no `@Property`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2218/commits/3c38a04fe9de4625a0ebd8c308e66fc666c0428b

PR #2223, opened 2019-01-05
First basic draft version of a java client library. need feedback 

renepickhardt:
I have started to create a Java Client Library for c-lightning called `JLightning` which is basically following `pylightning` code. 

This PR is work in progress. I would like to have some feedback before I continue to finish the work. 

1. As Java is a typed language I wonder if the API calls should really give back JSON or just Strings?
2. Naming methods and arguments should follow a certain standard. I started out taking over the argument names from c-lightning but java has a different naming conventions. 
3. DocStrings. I started to copy the doc strings from pylightning but again Java doc strings are usually formatted differently producing a particular line for each parameter. 
4. The Client Library handles the communication with the Unix Domain Socket differently than pylightning (less complex but it is still working) am I missing something? I guess I will not be backward compatible with older c-lightning nodes.
5. Java does not support default values for methods so calls without arguments would get a lot of `null` values. 
6. The check which fields in the payload are null is quite annoying in java I wonder if I should check this within the lib when adding the fields to the HashMap

### DIFF
--- a/contrib/JLightning/README.md
+++ b/contrib/JLightning/README.md
@@ -1,0 +1,14 @@
+Draft version of the JLightning rpc interface for c-lightning.
+
+Using this client library is is as simple as the following code:
+```
+public static void main(String[] args) {
+	JLightningRpc rpc_interface = new JLightningRpc("/tmp/spark-env/ln1/lightning-rpc");
+	String res = rpc_interface.listInvoices(null);
+	System.out.println(res);
+	res = rpc_interface.listFunds();
+	System.out.println(res);
+}
+```
+
+This client library is provided and maintained by Rene Pickhardt
--- a/contrib/JLightning/src/de/renepickhardt/ln/JLightning.java
+++ b/contrib/JLightning/src/de/renepickhardt/ln/JLightning.java
@@ -1,0 +1,36 @@
+package de.renepickhardt.ln;
+
+/**
+ * JLightning a small test / example class to demonstrate how to use JLightningRpc
+ * 
+ * This file is basically a port of the pylightning python client library
+ * that comes with c-lightning.
+ * 
+ * The Author of this Java Client library is Rene Pickhardt. 
+ * He also holds the copyright of this file. The library is licensed with
+ * a BSD-style license. Have a look at the LICENSE file. 
+ * 
+ * If you like this library consider a donation via bitcoin or the lightning
+ * network at http://ln.rene-pickhardt.de
+ * 
+ * @author Rene Pickhardt
+
+ */
+
+public class JLightning {
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+		
+		JLightningRpc rpc_interface = new JLightningRpc("/tmp/spark-env/ln1/lightning-rpc");
+		String res = rpc_interface.listInvoices(null);
+		System.out.println(res);
+		res = rpc_interface.listFunds();
+		System.out.println(res);
+		res = rpc_interface.listInvoices(null);
+		System.out.println(res);
+
+	}
+
+
+}
--- a/contrib/JLightning/src/de/renepickhardt/ln/JLightningRpc.java
+++ b/contrib/JLightning/src/de/renepickhardt/ln/JLightningRpc.java
@@ -1,0 +1,147 @@
+package de.renepickhardt.ln;
+/**
+ * JLightningRpc extends the UnixDomainSocketRpc and exposes the specific 
+ * API that is provided by c-lightning. It  is a java client library for the 
+ * c-lightning node. It connects to c-lightning via a Unix Domain Socket over
+ * JsonRPC v2.0 
+ * 
+ * This file is basically a port of the pylightning python client library
+ * that comes with c-lightning.
+ * 
+ * The Author of this Java Client library is Rene Pickhardt. 
+ * He also holds the copyright of this file. The library is licensed with
+ * a BSD-style license. Have a look at the LICENSE file. 
+ * 
+ * If you like this library consider a donation via bitcoin or the lightning
+ * network at http://ln.rene-pickhardt.de
+ * 
+ * @author Rene Pickhardt
+ */
+
+import java.util.HashMap;
+
+
+
+public class JLightningRpc extends UnixDomainSocketRpc {
+
+	public JLightningRpc(String socket_path) {
+		super(socket_path);
+		
+	}
+	
+	/**
+	 * Delete unpaid invoice {label} with {status}
+	 * @param label of the invoice
+	 * @param status status of the invoice
+	 * @return
+	 */
+	public String delInvoice(String label, Status status) {
+		HashMap<String, String> payload = new HashMap<String,String> ();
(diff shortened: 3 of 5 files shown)


contrib/JLightning/src/de/renepickhardt/ln/UnixDomainSocketRpc.java
```diff
@@ -0,0 +1,119 @@
...
+			
+			// FIXME: Using a StringBuilder would be preferable but not so easy
+			String response = "";
+			int buffSize = 1024;
+			byte[] buf = new byte[buffSize];
+
+			while(true) { 
+				int read = this.is.read(buf);
+				response = response + new String(buf, 0, read, "UTF-8");
+				if(read <=0 || response.contains("\n\n" )) {
```
jb55: in my Haskell implementation I count open and closed curly brackets (ensuring they sum to 0) to determine when to stop reading.

This could fail in an unlikely event that both `\n`'s appear at the end and start of two separate reads. This would result in the entire program blocking indefinitely on the next read call.

renepickhardt: Response is the string that has been appended with the buffer so I see no problem with split \n

On the other side the sum of open and closed brakets might fail as those could be part of the strings in the json data. So those need to be ignored

jeffrade:
@renepickhardt I'm personally not a fan of using [`Optional`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html), but if using Java 1.8, you can use it [to handle](https://www.oracle.com/technetwork/articles/java/java8-optional-2175753.html) `null`.

contrib/JLightning/src/de/renepickhardt/ln/JLightning.java
```diff
@@ -0,0 +1,36 @@
...
+ * 
+ * The Author of this Java Client library is Rene Pickhardt. 
+ * He also holds the copyright of this file. The library is licensed with
+ * a BSD-style license. Have a look at the LICENSE file. 
+ * 
+ * If you like this library consider a donation via bitcoin or the lightning
+ * network at http://ln.rene-pickhardt.de
+ * 
+ * @author Rene Pickhardt
+
```
jeffrade: nit - extra line break.

contrib/JLightning/src/de/renepickhardt/ln/JLightning.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+public class JLightning {
+
+	public static void main(String[] args) {
+		// TODO Auto-generated method stub
+		
+		JLightningRpc rpc_interface = new JLightningRpc("/tmp/spark-env/ln1/lightning-rpc");
```
jeffrade: I would inject this `"/tmp/spark-env/ln1/lightning-rpc"` via command line. Will be available in `args`.

contrib/JLightning/src/de/renepickhardt/ln/JLightning.java
```diff
@@ -0,0 +1,36 @@
+package de.renepickhardt.ln;
+
+/**
+ * JLightning a small test / example class to demonstrate how to use JLightningRpc
```
jeffrade: Maybe move this into a `README.md` and add the necessary code snippets that you provide here in the `main` method.

contrib/JLightning/README.md
```diff
@@ -0,0 +1,14 @@
+Draft version of the JLightning rpc interface for c-lightning.
+
+Using this client library is is as simple as the following code:
```
jeffrade: Even if someone is familiar with java, would be nice to have the actual command to build and launch: e.g.
```
$ javac <details-to-build>
$ java <details-to-run>
```

jeffrade: On that note, have you though of using Gradle or Maven to handle building and dependency management? Preferably [Gradle](https://docs.gradle.org/current/userguide/getting_started.html)

contrib/JLightning/src/de/renepickhardt/ln/JLightningRpc.java
```diff
@@ -0,0 +1,147 @@
...
+	/**
+	 * Delete unpaid invoice {label} with {status}
+	 * @param label of the invoice
+	 * @param status status of the invoice
+	 * @return
+	 */
+	public String delInvoice(String label, Status status) {
+		HashMap<String, String> payload = new HashMap<String,String> ();
+		payload.put("label", label);
+		payload.put("status", status.toString());
```
jeffrade: To handle `NullPointerException` here on `toString()`, you can use [`String.valueOf(status)`](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#valueOf-java.lang.Object-). In case of `null`, it will return the `String` `"null"`.

contrib/JLightning/src/de/renepickhardt/ln/UnixDomainSocketRpc.java
```diff
@@ -0,0 +1,119 @@
...
+		}
+	}
+	
+	public String call(String method, HashMap<String, String> payload) {
+		// if no payload is given make empty one
+		if(payload == null) {
+			payload = new HashMap<String, String>();
+		}
+		
+		//remove null items from payload
```
jeffrade: There's no need to add the keys to a new `HashSet` here. You can simply iterate through the `payload.keySet()` and [and remove an entry if value is `null`](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html#remove-java.lang.Object-java.lang.Object-).

Here's some sample code:
```
Map<String, String> m = new HashMap<String, String>();
    m.put("1", "One");
    m.put("2", null);
    System.out.println(m);
    for(String k:m.keySet()) {
      m.remove(k, null);
    }
    System.out.println(m);
```

And output is:
```
{1=One, 2=null}
{1=One}
```

contrib/JLightning/src/de/renepickhardt/ln/UnixDomainSocketRpc.java
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}
+		
+		JSONObject json = new JSONObject();
+		
+		try {
+			json.put("method", method);
+			json.put("params", new JSONObject(payload));
+			// FIXME: Use id field to dispatch requests
+			json.put("id", Integer.toString(this.id++));
```
jeffrade: `this.id` isn't thread-safe since it is `static`. Use [AtomicInteger](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicInteger.html) instead and then call `getAndIncrement()`.

greenaddress:
Not sure how relevant but in abcore I needed json rpc to talk to Bitcoin core daemon running in the background and I used https://github.com/Polve/bitcoin-rpc-client , maybe it can be adapted?

cdecker:
@renepickhardt are you planning on moving this library into its own repository? Otherwise I could give it a new home in the `lightningd` org and add you there :-)

PR #2230, opened 2019-01-07
listpeers: include `private` field in `channels` output

niftynei:
Reveal channel's 'privacy' in `listpeers` output

Suggested-By: @shesek


tests/test_connection.py
```diff
@@ -805,6 +805,12 @@ def test_private_channel(node_factory):
     assert not l1.daemon.is_in_log('Received node_announcement for node {}'.format(l2.info['id']))
     assert not l2.daemon.is_in_log('Received node_announcement for node {}'.format(l1.info['id']))
 
+    # test for 'private' flag in rpc output
+    assert len(l1.rpc.listpeers()) == 1
```
rustyrussell: I don't think this does what you want?  This passes because the output only has one field 4, "peers".

The next line checks there's only one entry in the peers array: only_one asserts that

PR #2232, opened 2019-01-08
[rpc] Add funding allocation map to listpeers command

conscott:
A redo of #2207 

The point is the know the initial state of funding allocated to the channel by the funder / fundee. At the moment there are only one-sided channel opens, but dual-funded channels will be available in the near future :crossed_fingers: 

This change also implicitly let's you know who initiated the channel open, which was the original intent of #2207

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -720,6 +720,21 @@ static void json_add_peer(struct lightningd *ld,
 			      &channel->funding_txid);
 		json_add_bool(response, "private",
 				!(channel->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL));
+
+		// FIXME @conscott : Modify this when dual-funded channels
+		// are implemented
+		json_object_start(response, "funding_allocation_msat");
+		if (channel->funder == LOCAL) {
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &p->id), 0);
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &ld->id),
+					channel->funding_satoshi * 1000);
+		} else {
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &ld->id), 0);
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &p->id),
+					channel->funding_satoshi * 1000);
+		}
+		json_object_end(response);
+
 		json_add_u64(response, "msatoshi_to_us",
 			     channel->our_msatoshi);
 		json_add_u64(response, "msatoshi_to_us_min",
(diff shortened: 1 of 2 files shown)


lightningd/peer_control.c
```diff
@@ -718,6 +718,21 @@ static void json_add_peer(struct lightningd *ld,
...
 		json_add_bool(response, "private",
 				!(channel->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL));
+
+		// FIXME @conscott : Modify this when dual-funded channels
+		// are implemented
+		json_object_start(response, "funding_allocation_msat");
+		if (channel->funder == LOCAL) {
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &p->id), 0);
+			json_add_u64(response, pubkey_to_hexstr(tmpctx, &ld->id),
+					channel->funding_satoshi * 1000);
```
niftynei: What's the rationale for multiplying this by 1000? `funding_satoshi` should be a satoshi value already.

conscott:
Sure @niftynei !

The original intent of #2207 was to allow you to see if the the channel open was inbound or outbound, basically to know who funded the channel. `lnd` has a similar field in `listpeers`. 

Christian then mentioned that dual funded channels are coming soon, so why not just have a map of channels id's -> initial channel state funding (what I have done here). From this information we can implicitly know who opened the channel, but also in the future know if it was dual-funded and what the initial state was.

I think this useful because it can allow you to see more rich stats on the 'progress' of your channel usage since it has been opened, along with `msatoshi_to_us_min` and `msatoshi_to_us_max`

niftynei:
Cool yeah I can see how that would be useful for metric tracking. 

Could you add an entry to the CHANGELOG.md file and update the `listpeers` [manpage](//github.com/ElementsProject/lightning/tree/master/doc/lightning-listpeers.7.txt) to include a short blurb about this added fieldset?

conscott:
Rebased on top of #2248 to get the CHANGELOG.md for next release (should wait to merge this until that one is through). 

@niftynei : I added note to the CHANGELOG, however for the `listpeers` manpage, we currently [don't describe anything in the channel field](https://github.com/ElementsProject/lightning/blob/master/doc/lightning-listpeers.7.txt#L51), so I am thinking we could open a separate issue and make a note to have a PR to fully describe all these fields, instead of the one I just added.

What do you think?

niftynei:
i'll take it lol

ACK https://github.com/ElementsProject/lightning/pull/2232/commits/2442e9502646c3ac0867f47d93ff601fff97cd62

niftynei:
fwiw https://github.com/ElementsProject/lightning/issues/1360 has a todo for the `listpeers` manpage.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2232/commits/aa7a0514b59233deb571ceceee30a9eec4203eec

PR #2237, opened 2019-01-09
The final chapter in the plugin saga (hooks)

cdecker:
This PR implements the hooks infrastructure for plugins and adds the `htlc_accepted` hook as a sample of how to implement hooks and how they can be used.

The `REGISTER_PLUGIN_HOOK` macro is the glue that binds everything together, defining the full flow of `call -> serialize_request -> send_request -> deserialize_response -> callback`.

This is not yet complete, missing one commit to actually react to the response to the `htlc_accepted` hook, and I'm planning to add another commit that tests the various hook return values.

I'm adding this now since I'd like to get the reviews going.


cdecker:
Notice that in `htlc_accepted_hook_deserialize` all parsing failures are `fatal`. I'm not exactly sure if this is what we want, but I expect that we can't really recover from an error in this case, since going back to default behavior might not be what we want (if the plugin hook checks some policies, and would reject it, the default action might be to accept and thus to violate the policy). Therefore I made it as pedantic as possible in its current form.

Another nit in this case is that at this point we have not committed to the DB yet, which means that crashing will cause the DB transaction to rollback, potentially dropping the commitment we just got. This could however be solved by a variant of `fatal` that commits in case we have a DB transaction open.

What do you guys think @rustyrussell and @niftynei?

lightningd/peer_htlcs.c
```diff
@@ -590,16 +592,169 @@ static void channel_resolve_reply(struct subd *gossip, const u8 *msg,
...
+			    "Plugin did not specify a 'payment_key' in return "
+			    "value to the htlc_accepted hook: %s",
+			    json_tok_full(buffer, resulttok));
+
+		if (!json_to_preimage(buffer, paykeytok,
+				      &response->payment_key))
+			fatal("Plugin specified an invalid 'payment_key': %s",
+			      json_tok_full(buffer, resulttok));
+	} else {
+		fatal("Plugin responded with an unknowk result to the "
```
niftynei:
```suggestion
		fatal("Plugin responded with an unknown result to the "
```

lightningd/plugin.c
```diff
@@ -1057,3 +1027,15 @@ void plugins_notify(struct plugins *plugins,
 	if (taken(n))
 		tal_free(n);
 }
+
+void plugin_request_send(struct plugin *plugin,
+			 struct jsonrpc_request *req TAKES)
+{
+	/* Add to map so we can find it later when routing the response */
+	tal_steal(plugin, req);
```
rustyrussell: This leaks take() if caller does it as implied by the TAKES above.  You can just document that this function consumed req, and remove the TAKES annotation?

lightningd/peer_htlcs.c
```diff
@@ -673,31 +747,23 @@ static bool peer_accepted_htlc(struct channel *channel,
...
+	 * hook so plugins can interject if they want */
+	hook_payload = tal(hin, struct htlc_accepted_hook_payload);
+	hook_payload->route_step = tal_steal(hook_payload, rs);
+	hook_payload->ld = ld;
+	hook_payload->hin = hin;
+	hook_payload->channel = channel;
+	plugin_hook_call_htlc_accepted(ld, hook_payload, hook_payload);
 
+	/* Falling through here is ok, after all the HTLC locked */
 	*failcode = 0;
```
rustyrussell: This needs careful design.  For example, I'm pretty sure we want to separate hooks for terminating onions and non-terminating, since most hooks will want one or the other.  For terminating we'll also want a hook after we've entered it into the db, I think, and maybe an API to query the db on restart.

In short, someone needs to use the APIs in anger before they can be finalized.

common/json.c
```diff
@@ -155,6 +155,30 @@ bool json_to_bitcoin_amount(const char *buffer, const jsmntok_t *tok,
...
+		return tal_free(result);
+
+	return result;
+}
+
+bool json_to_preimage(const char *buffer, const jsmntok_t *tok, struct preimage *preimage)
+{
+	size_t hexlen = tok->end - tok->start;
+
+	if (hexlen != sizeof(preimage->r))
```
rustyrussell: Pretty sure hex_decode already does this check?

cdecker:
> Another nit in this case is that at this point we have not committed to the DB yet, which means that crashing will cause the DB transaction to rollback, potentially dropping the commitment we just got. This could however be solved by a variant of fatal that commits in case we have a DB transaction open.

My mistake, the last part is not an issue, since calling out to the plugin will result in the DB transaction that triggered the event to be committed. So rather than us forgetting stuff, this means that actions spanning across a plugin hook are no longer in the same DB transaction, which makes them not atomic anymore. Shouldn't be an issue for now since we never rollback a DB transaction currently and only failures can result in aborted transactions. Just something to keep in mind when working with hooks in the future.

lightningd/plugin_hook.h
```diff
@@ -0,0 +1,118 @@
...
+/* Create a small facade in from of `plugin_hook_call_` to make sure
+ * arguments are of the correct type before downcasting them to `void
+ * *`. Not really necessary, but nice since it also makes sure that
+ * the method-name is correct for the call.
+ */
+/* FIXME: Find a way to avoid back-to-back declaration and definition */
+#define PLUGIN_HOOK_CALL_DEF(name, payload_type, response_cb_arg_type)         \
+	void plugin_hook_call_##name(struct lightningd *ld,                    \
+				     payload_type payload,                     \
+				     response_cb_arg_type cb_arg);             \
```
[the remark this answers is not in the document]
cdecker: Yes, like mentioned above this one is the declaration of the function (no function body) and the lower one is the function definition (which contains the body). It's a bit of an antipattern since the definition should be in the matching `.c` file, but since we'd like to define this in a macro it has to be in the same place, and since we may want to call this from somewhere other than the module defining it (hooks should be defined together with the data they act on, and called from the entrypoints of the hooks), this is the best I could come up with.

Maybe @rustyrussell has a better option? An alternative that I haven't tried yet is to use a `#define` instead of this tiny function, avoiding all the linker issues altogether.

lightningd/plugin.c
cdecker: Hm, seems I'm missing the nuances of `TAKES` here, I thought the annotation means that it "takes ownership" and that the caller should not assume that the argument is accessible after the call anymore.

lightningd/peer_htlcs.c
cdecker: I don't think there is much difference, immediately failing the HTLC after it was accepted (calling `local_fail_htlc` -> `fail_in_htlc`) or passing through the hook and then calling `fail_in_htlc` from the hook callback doesn't really do much. Things would be different if we could prevent an HTLC being committted to in the first place (a real immediate failure, instead of commit-then-fail), but as it stands the only difference is that the two updates to the HTLC (committed then failed) are no longer in the same DB transaction.

cdecker: As for terminating vs non-terminating onions (i.e., I'm the intended recipient vs just an intermediate hop), I did not differentiate on purpose in order to allow things like short-circuiting a payment, i.e., I know the `payment_key` because I'm in an ensemble of load-balancing nodes so I can claim the funds directly.

common/json.c
cdecker: :+1: will drop duplicate check :-)

lightningd/plugin_hook.c
```diff
@@ -1,5 +1,35 @@
...
+}
+
+bool plugin_hook_register(struct plugin *plugin, const char *method)
+{
+	struct plugin_hook *hook = plugin_hook_by_name(method);
+	if (!hook) {
+		/* No such hook name registered */
+		return false;
+	} else if (hook->plugin != NULL) {
+		/* Another plugin already registered for this name */
```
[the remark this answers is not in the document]
cdecker: Yeah, that might become a problem once people start using a myriad of different plugins, but I see no clean to fan-out to multiple plugins (serial vs parallel), and more importantly how to do fan-in, i.e., given two or more results which one to pick (serial call, and first non-default is picked vs. some sort of priority vs. ...).

Further complicating this is that hooks might have different ways of handling this, i.e., a hook that is just used to synchronously save information somewhere is trivially safe to use in parallel and have multiple ones, while a hook that actually modifies the continuation is not.

So the solution I'm eyeing currently is that we just punt this sort of multiplexing to a multiplexer plugin, that will then run the child-plugins instead of registering them with `lightningd` directly, and that would then know how to handle things.

We should probably start out like this and if there is a need to, we can start moving this kind of functionality into the plugin_hooks as necessary (I personally don't think there will be a huge competition for hooks, since they cater really specific needs, and most users will be happy with the default implemented in `lightningd` anyway).

lightningd/plugin.c
```diff
@@ -1013,7 +983,7 @@ static void plugin_config(struct plugin *plugin)
 	json_object_end(req->stream);
 
 	jsonrpc_request_end(req);
-	plugin_send(plugin, req->stream);
+	plugin_request_send(plugin, req);
```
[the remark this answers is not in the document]
cdecker: See @rustyrussell's comment below, I seem to be missing some detail about the `TAKES` annotation, but I'd generally agree that that's what I needed.

lightningd/plugin_hook.h
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * - `serialize_payload` which takes a payload of type `payload_type`
+ *   and serializes it into the given `json_stream`. `
+ *
+ * - `deserialize_response` takes a `json_stream` and parses it into a
+ *   new struct of type `response_type`,
+ *
+ * - `response_cb` is called once the plugin has responded and the
+ *   response has been parsed by `deserialize_response`. In addition
+ *   an arrbitrary additional argument of type `cb_arg_type` can be
```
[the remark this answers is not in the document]
cdecker: Isn't it talk like a pirate day? My bad :wink:

lightningd/plugin_hook.c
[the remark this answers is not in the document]
niftynei: Right I think this makes sense at the outset.  It seems worth noting somewhere doc wise to cut down on potential for confusion later on.

lightningd/plugin_hook.h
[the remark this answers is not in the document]
rustyrussell: This Just Worked for me?

```diff
diff --git a/lightningd/plugin_hook.h b/lightningd/plugin_hook.h
index 48dffe23..74be1e74 100644
--- a/lightningd/plugin_hook.h
+++ b/lightningd/plugin_hook.h
@@ -71,12 +71,8 @@ void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
  * *`. Not really necessary, but nice since it also makes sure that
  * the method-name is correct for the call.
  */
-/* FIXME: Find a way to avoid back-to-back declaration and definition */
 #define PLUGIN_HOOK_CALL_DEF(name, payload_type, response_cb_arg_type)         \
-       void plugin_hook_call_##name(struct lightningd *ld,                    \
-                                    payload_type payload,                     \
-                                    response_cb_arg_type cb_arg);             \
-       void plugin_hook_call_##name(struct lightningd *ld,                    \
+       static inline void plugin_hook_call_##name(struct lightningd *ld,      \
                                     payload_type payload,                     \
                                     response_cb_arg_type cb_arg)              \
        {                                                                      \
```

lightningd/plugin.c
rustyrussell: It means "IF you call with take(x), it will take ownership".

Maybe we need an annotation for "consumes"?

lightningd/peer_htlcs.c
rustyrussell: We need a real user of this htlc_accepted API; I'm not smart enough to design it without a concrete one :(

Consider a just-in-time invoicing system.  The current hook is in the wrong place: any of the checks in handle_localpay could still fail.

And I can't see how the plugin's db can stay in sync with lightningd's: a restart or crash can lead to one committed and one not.  For example, after plugin commits to its db, and says continue, we crash or simply shutdown.  Now htlcs_reconnect() fails the HTLC on restart.

Adding an external dependency like this is going to be painstaking work; I think we need to pick the smallest subset so we can make mistakes with limited consequences.

lightningd/peer_htlcs.c
```diff
@@ -590,16 +592,194 @@ static void channel_resolve_reply(struct subd *gossip, const u8 *msg,
 	tal_free(gr);
 }
 
+/**
```
conscott: Nit : Perhaps all the hook structures / functions can live in another file? It seems there are many options for future hooks and I can imagine them bloating some files that are already quite large, so it might set the precedent to keep those in their own space.

lightningd/peer_htlcs.c
conscott: IF you were interested in starting with a simpler case with potentially less downstream implications, I could imagine various hooks for when channels change state. For instance, a simple hook for when the funding transaction is confirmed, or if a channel is asked to closed. 

On the plugin side, something simple like an email/sms alert when a channel closes could be nice. Perhaps this meets the balance between some utility and less consequential.

lightningd/plugin_hook.h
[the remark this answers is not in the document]
cdecker: Interesting, I thought `static inline` would only be callable in the same compilation unit, but you're right, since we are `#include`ing the file, this probably means the client header as well. Will fixup :+1:

lightningd/peer_htlcs.c
cdecker: This is not meant to interfere with the built-in invoicing system, it is rather a complement to it, and is meant to be used at a much lower level. With this you can pretty much create an external invoicing system, that'll get queried for open invoices.

> Consider a just-in-time invoicing system. The current hook is in the wrong place: any of the checks in handle_localpay could still fail.

I added a small test that short-circuits a payment going through `l1 -> l2 -> l3` at the middle node `l2` and just claims the HTLC, leaving the invoice on `l3` unpaid. Notice that we don't use `handle_localpay` or `forward_htlc` at all if we get told to fail or resolve.

> And I can't see how the plugin's db can stay in sync with lightningd's: a restart or crash can lead to one committed and one not. For example, after plugin commits to its db, and says continue, we crash or simply shutdown. Now htlcs_reconnect() fails the HTLC on restart.

As for the problems that may emerge from `htlcs_reconnect`, notice that I avoided having a forward-to-this channel call on purpose, so that we cannot end up with an outgoing HTLC that cannot be associated with an incoming one. The incoming HTLC on the other hand immediately fails, or gets told the preimage if the plugin returns, hence we end up always in a case that we already handle correctly.

I think as it stands this is the simplest hook I can come up with: we get an HTLC and decide whether to continue, fail or resolve. Continue is the old, and proven, logic. Fail is about as simple as it gets, and resolve is just resolving an HTLC with a `payment_key` that did not come from an invoice, but somewhere else :wink:

lightningd/peer_htlcs.c
cdecker: I considered something similar, but most of the logic is actually either taken from the `peer_htlcs.c` file (in the callback) or wraps and unwraps data that may not be available elsewhere, which is why I ended up putting it here. I'd prefer to keep the logic at least in the same place.

lightningd/peer_htlcs.c
cdecker: After discussing this in person with @rustyrussell I now understand the issue a bit better. We may end up imposing that the hook calls must be idempotent so that we can call them again in case we have an HTLC that was accepted and wasn't resolved or failed or forwarded.

cdecker:
Since the `htlc_accepted` hook proves to be a bit more controversial than I thought (and rightfully so), I split the PR into two branches, with this one retaining only the infrastructure changes to enable hooks, and moving the `htlc_accepted` hook into #2267.

Since this part was more or less accepted, this should be much easier to review. I just added a single fixup to address the unused-function issue for our dummy hook.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2237/commits/fdc66cf303c53a23f86ac3db1dfbe8832fc25389

PR #2240, opened 2019-01-10
make: fail build if git isn't present

### LINKED ISSUE #2238: VERSION is empty when compiled without git in path
I just upgraded to 0.6.3rc1, when I type lightning-cli --version I get an empty string. getinfo also has an empty version. configure flags are `./configure --prefix=$out --disable-developer --disable-valgrind`

jb55:
git is needed to generate version information

Add a sanity check so that the build doesn't continue with an empty VERSION. This
is useful for sandboxed build where we might have forgot to include git.

Fixes #2238

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,11 @@
 #! /usr/bin/make
 VERSION_NAME=Principled Opposition to SegWit
 VERSION=$(shell git describe --always --dirty=-modded --abbrev=7)
+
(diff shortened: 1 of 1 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2240/commits/785e71122f363efd6da7c7c292722e926bda2eaa

PR #2243, opened 2019-01-10
misc: Update short_channel_id representation to use 'x' separators

### LINKED ISSUE #2239: Short channel ids should use x as a separator instead of :
When working with human-readable short channel ids, such as `1440743:38:0` the colons should be replaced with `x` characters, like `1440743x38x0`

cdecker:
Only the changes in `bitcoin/short_channel_id.c` are important, everything else is just test compat cleanup.

Fixes #2239

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
+- The `short_channel_id` separator has been changed to be `x` to match the specification.
+
 ### Deprecated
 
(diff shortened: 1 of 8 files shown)


rustyrussell:
Needs a BOLT update and quote, but I suspect that will require other changes.  Also needs a CHANGELOG.md entry.

cdecker:
Thanks for pointing this out, I updated the bolts and quotes in #2256 and added a new commit to add the quote.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2243/commits/dbbb8c0056f3a42d78e3cae9377961bdc383b274

PR #2244, opened 2019-01-10
plugin: Gate the plugin subsystem with the --enable-plugins configure flag

rustyrussell:
[ Actual work done by Christian, I just made it a new configure flag and tweaked docs --RR]

Since we are planning to release a bug fix release, and the plugin
subsystem is not yet complete, it is better to make plugin support
opt-in while we continue testing.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2244/commits/050c6357bad94a09a2823ca2e821a4f68c842aed

PR #2247, opened 2019-01-11
Fix release zipfile

rustyrussell:


CHANGELOG.md
```diff
@@ -230,7 +232,7 @@ There predate the BOLT specifications, and are only of vague historic interest:
 7. [0.5.2] - 2016-11-21: "Bitcoin Savings & Trust Daily Interest II"
 
 [Unreleased]: https://github.com/ElementsProject/lightning/compare/v0.6.2...HEAD
-[0.6.3-rc1]: https://github.com/ElementsProject/lightning/compare/v0.6.3-rc1...HEAD
+[0.6.3]: https://github.com/ElementsProject/lightning/compare/v0.6.3...HEAD
```
jb55: should this line be:  `https://github.com/ElementsProject/lightning/releases/tag/v0.6.3` ?

jb55: and `[Unreleased]: https://github.com/ElementsProject/lightning/compare/v0.6.3...HEAD`

rustyrussell: Yeah, I fuxked up that should not have been here at all.  Will force push

jb55:
if you have a fresh zip file generated from this I would be interested in testing it. for now I'm still using recursive clone in https://github.com/NixOS/nixpkgs/pull/53732

cdecker:
Small fixup to fix the `shellcheck` checks about unquoted values. `$FILES` could not be quoted since we actually want that var to be split, so I opted to silence the error instead.

rustyrussell:
Rebased onto master and folded @cdecker fix in.

PR #2255, opened 2019-01-11
added manpage for listforwards command (needs change)

renepickhardt:
I don't understand how manpages are created automatically via the a2x tool. That is why id wrote both files manually. I guess I need only the .7 or the .7.txt file and the other one is automatically created. my machine always creates an xml file and then make produces an error. 
So I need some help and will then redo the PR

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -18,6 +18,7 @@ MANPAGES := doc/lightning-cli.1 \
 	doc/lightning-getroute.7 \
 	doc/lightning-invoice.7 \
 	doc/lightning-listchannels.7 \
+	doc/lightning-listforwards.7 \
 	doc/lightning-listfunds.7 \
 	doc/lightning-listinvoices.7 \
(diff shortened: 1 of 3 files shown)


conscott:
I just experienced this. You _should_ only have to modify the `.txt` and it should build with `make`, cleanup ingthe temporary xml files, but I was also erroring out running the a2x command manually.

I had to install `xsltproc`, and then it started working. It seems on some systems this is a dependency of asciidoc in the package manager, but not on others.

If that also fixes your issue, I suppose we can make a note in the doc/INSTALL.md to install that as well if changing manpages.

renepickhardt:
thanks @conscott will check that later... I am in a hurry right now

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2255/commits/9da3ff3080732b1d9f9c9adec40275f02fd9128b

PR #2256, opened 2019-01-11
bolt: Updated the BOLT specification to the latest version

cdecker:
This is mainly just copying over the copy-editing from the
lightning-rfc repository. The only code changes are due to
`unknown_payment_hash` being renamed to
`incorrect_or_unknown_payment_details`.

Reported-by: Rusty Russell <@rustyrussell>


cdecker:
Thanks for the cleanup @rustyrussell, I must've missed this in the avalanche of changes :-)

Is this still bisectable? That was the main reason I didn't split it into multiple commits.

rustyrussell:
Yes, is bisectable.  We're falling behind with our PRs, so I'm going to take that as an Ack :)

PR #2272, opened 2019-01-17
Json tok speedup

rustyrussell:
Built on top of #2215 simply because that's where @cdecker pointed out we weren't using tok->size for iteration and thus doing twice the work we needed to.

So just review the last 4 ("common/test/run-json: check tok->size is as expected" onwards)

### DIFF
--- a/common/json.h
+++ b/common/json.h
@@ -51,7 +51,7 @@ bool json_tok_is_num(const char *buffer, const jsmntok_t *tok);
 /* Is this the null primitive? */
 bool json_tok_is_null(const char *buffer, const jsmntok_t *tok);
 
-/* Returns next token with same parent. */
+/* Returns next token with same parent (WARNING: slow!). */
 const jsmntok_t *json_next(const jsmntok_t *tok);
 
 /* Get top-level member. */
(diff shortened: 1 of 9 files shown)


cdecker:
Rebased, reviewing today.

common/test/run-param.c
```diff
@@ -183,7 +183,7 @@ static void tok_tok(void)
 	{
 		unsigned int n;
 		const jsmntok_t *tok = NULL;
-		struct json *j = json_parse(cmd, "{ 'satoshi', '546' }");
+		struct json *j = json_parse(cmd, "{ 'satoshi': '546' }");
```
cdecker: Interesting issue, I was not expecting `jsmn` to catch this correctly, given the overall quality of the library we've seen so far.

Also kudos to @rustyrussell for hunting this down :+1:

common/json.h
```diff
@@ -85,4 +85,13 @@ const jsmntok_t *json_delve(const char *buffer,
...
 			    const char *guide);
 
+/* Iterator macro for array: i is counter, t is token ptr, arr is JSMN_ARRAY */
+#define json_for_each_arr(i, t, arr) \
+	for (i = 0, t = (arr) + 1; i < (arr)->size; t = json_next(t), i++)
+
+/* Iterator macro for object: i is counter, t is token ptr (t+1 is
+ * contents of obj member), obj is JSMN_OBJECT */
+#define json_for_each_obj(i, t, obj) \
+	for (i = 0, t = (obj) + 1; i < (obj)->size; t = json_next(t+1), i++)
```
cdecker: Awesome macro-magic :-)

Issue #2171, opened 2018-12-12
Channel "AWAITING_UNILATERAL" since days. Attempted force closing, CPU 100%


gabridome:
I found one of my channel was in "AWAITING_UNILATERAL" so I waited but days after it was still in that state.

Relevant log with grep *peerID*:
```
2018-11-25T13:41:40.283Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: 0: 10648 satoshi (SEGWIT) 695254bb08add33af6f907960913869c25903c967541b2b8765be96431d2cb59?                                                                               [100/260]
2018-11-25T13:41:40.283Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: 1: 2005802 satoshi (SEGWIT) bd729c73e2cd2647f5ce6a2ebd24419720486f720124d9e0c66649b0b3d2835d?                                                                                      
2018-11-25T13:41:40.283Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: 2: 252848 satoshi (SEGWIT) bbe3783d4c7fde8eb6433403fc1c34fa0098e4c738cedeac087243a1655a0b81?                                                                                       
2018-11-25T13:41:40.300Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: Getting HSM to sign funding tx                                                                                                                                                     
2018-11-25T13:41:40.397Z lightningd(17827): lightning_channeld-03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: pid 31244, msgfd 33                                                                                                                                             
2018-11-25T13:41:40.397Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: Waiting for funding confirmations                                                                                                                                                  
2018-11-25T13:41:41.001Z lightningd(17827): 03e50492eab4107a773141bb419e107bda3de3d55652e6e1a41225f06a0bbf2d56 chan #10: Peer internal error CHANNELD_AWAITING_LOCKIN: Funding broadcast exited with 25: error code: -25?error message:?Missing inputs?                                                     
[...]

PR #2274, opened 2019-01-17
fundchannel command: return an error when broadcast fails

SimonVrouwe:
openingd/json_fund_channel:
    - result fundchannel command now depends on successful or failed broadcast of the funding tx
    - failure returns error code FUNDING_BROADCAST_FAIL
    - don't fail the channel when broadcast failed, but keep in CHANNELD_AWAITING_LOCKIN
    - after fixing the initial broadcast failure, the user could manually rebroadcast the tx and
      keep the channel
    
openingd/opening_funder_finished:
    - broadcast_tx callback function now handles both success and failure
    
jsonrpc: added error code FUNDING_BROADCAST_FAIL
    manpage: added error code returned by fundchannel command
    
This makes the user more aware of broadcast failure, so it hopefully doesn't
try to broadcast new tx's that depend on its change_outputs. Some users have reported (see
issue #2171) a whole sequence of fundings failing, because each funding was using the change
output of the previous one, which would not confirm.

### DIFF
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -66,15 +66,20 @@ struct uncommitted_channel {
 
 
 struct funding_channel {
-	struct command *cmd; /* Which also owns us. */
+	struct command *cmd; /* Which initially owns us until openingd request */
+
 	struct wallet_tx wtx;
 	u64 push_msat;
 	u8 channel_flags;
 
+	/* Variables we need to compose fields in cmd's response */
+	u8 *linear;
+	struct channel_id cid;
+
 	/* Peer we're trying to reach. */
 	struct pubkey peerid;
 
-	/* Channel. */
+	/* Channel, subsequent owner of us */
 	struct uncommitted_channel *uc;
 };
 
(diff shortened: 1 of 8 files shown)


lightningd/opening_control.c
```diff
@@ -216,18 +221,57 @@ wallet_commit_channel(struct lightningd *ld,
...
+			"Error broadcasting funding transaction: %s", output));
+
+	/* Frees fc too */
+	tal_free(fc->uc);
+
+	/* Keep in state CHANNELD_AWAITING_LOCKIN until (manual) broadcast */
+}
+
+static void funding_broadcast_success(struct channel *channel,
+				     int exitstatus, const char *msg)
```
rustyrussell: exitstatus isn't needed here....

tests/test_connection.py
```diff
@@ -1115,7 +1115,7 @@ def test_fundee_forget_funding_tx_unconfirmed(node_factory, bitcoind):
     time.sleep(1)
 
     def mock_sendrawtransaction(r):
-        return {'error': 'sendrawtransaction disabled'}
+        return {'42': 'sendrawtransaction disabled'}
```
rustyrussell: This seems wrong...

rustyrussell: This turns out to be a completely different problem; I've written a fix for it.

But it does make the fundchannel *fail*, which is what your patch wants to do!  So I fixed that too...

lightningd/opening_control.c
```diff
@@ -216,18 +221,57 @@ wallet_commit_channel(struct lightningd *ld,
...
-	channel_internal_error(channel,
-			       "Funding broadcast exited with %i: %s",
-			       exitstatus, err);
+	struct funding_channel *fc = channel->peer->uncommitted_channel->fc;
+	struct command *cmd = fc->cmd;
+
+	/* Massage output into shape so it doesn't kill the JSON serialization */
+	char *output = tal_strjoin(cmd, tal_strsplit(cmd, msg, "\n", STR_NO_EMPTY), " ", STR_NO_TRAIL);
+	was_pending(command_fail(cmd, FUNDING_BROADCAST_FAIL,
+			"Error broadcasting funding transaction: %s", output));
```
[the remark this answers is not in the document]
SimonVrouwe: Don't we want an unescaped string here? I just copied copied this from [wallet_withdrawal_broadcast](https://github.com/ElementsProject/lightning/blob/018a3f1d58e38e464e32cbee26826a0bd63aab59/wallet/walletrpc.c#L54)

tests/test_connection.py
SimonVrouwe: Ah yes I see now, in that test we need to catch the exception, triggered by `fundchannel`'s fail, to complete the test. Will wait for your fix.

cdecker:
Really like the way this is shaping up. It gives more control to the user. We might want to go a step further and move the following things into `funding_broadcast_success` as well:

https://github.com/ElementsProject/lightning/blob/f96cdd975a9932b60c137ef8ffee5c46533e81da/lightningd/opening_control.c#L408-L409

https://github.com/ElementsProject/lightning/blob/f96cdd975a9932b60c137ef8ffee5c46533e81da/lightningd/opening_control.c#L425

The rationale behind this is that some users have reported a whole sequence of fundings failing, because each funding was using the change output of the previous one, which would not confirm. If we move those two parts into the success callback we avoid adding outputs that are unlikely to confirm, and we also delay marking the reserved outputs as spent (which is something that we might want to move to the output confirmation path anyway, instead of the funding path), making it easier to reset them should the funding fail indefinitely.

rustyrussell:
> Should the funding eventually confirm `wallet_extract_owned_outputs` will be called anyway when the tx gets confirmed, we just call it to avoid people asking where their funds went.

Let's make sure that this works if we die between telling bitcoind to add the tx to the mempool and getting the response.  As far as I can tell, we'll then try to doublespend the output, which will fail: do we catch that case correctly if it's a wallet withdraw?

SimonVrouwe:
Hmmm, hope I did this correctly ... rebased my local branch on https://github.com/ElementsProject/lightning/commit/37c276daaa7d31d6641d6b996bd79e9b78372f5c then --force pushed it back

cdecker:
> Let's make sure that this works if we die between telling bitcoind to add the tx to the mempool and getting the response. As far as I can tell, we'll then try to doublespend the output, which will fail: do we catch that case correctly if it's a wallet withdraw?

The UTXO is set to `reserved` after selecting it, and before passing it to `openingd`, and the confirm call just goes from `reserved` to `spent` (which is a useless state anyway since we should really be tracking confirmations). Selection ignores both `reserved` and `spent` outputs, so that shouldn't cause us to double spend.

If we don't manually add the funding TX, and therefore the change output, we will grab it from the block once the funding confirms, so this really is just an optimization.

I'm happy to move this to another PR, if we want to discuss it further, and unblock this PR :wink:

cdecker:
> Hmmm, hope I did this correctly ... rebased my local branch on 37c276d then --force pushed it back

Seems that you rebased on something that was not `master` which can pick up some additional commits when comparing to `master` afterwards. No problem though since it'll automatically be solved once the extra commits are also merged into `master`.

SimonVrouwe:
> Really like the way this is shaping up. It gives more control to the user. We might want to go a step further and move the following things into `funding_broadcast_success` as well:

We could indeed make it further fail/fool-proof, but I don't see a way to do that without deeper modifications and new infrastructure.
> Let's make sure that this works if we die between telling bitcoind to add the tx to the mempool and getting the response.

Indeed something for a new PR.

PR #2275, opened 2019-01-18
option_data_loss_protect: fixup commitment point check

niftynei:
Spurious errors were occuring around checking the provided
current commitment point from the peer on reconnect when
option_data_loss_protect is enabled. The problem was that
we were using an inaccurate measure to screen for which
commitment point to compare the peer's provided one to.

This fixes the problem with screening, plus makes our
data_loss test a teensy bit more robust.


cdecker:
Looking good :+1:

rustyrussell:
Please submit another commit to remove it from experimental features?

PR #2290, opened 2019-01-23
docs: fix getroute manpage rendering

wythe:
The curly braces are causing missing text in the resulting man page.

I switched them to quotes and it works fine. This is somewhat more consistent
with our other man pages anyway.


niftynei:
Nice catch. ACK https://github.com/ElementsProject/lightning/pull/2290/commits/2ea04de95032938f02bfb72c11b21027ad3997c5

PR #2295, opened 2019-01-25
Add native Bitcoin rpc client

arowser:
implement #2051, Add native Bitcoin rpc client, so the c-lightning will not depend bitcoin-cli.

@rustyrussell @cdecker 
This new implement used raw socket to remove libevent depend in #2111


rustyrussell:
OK, I chatted with @cdecker about this.  Our plan was to make the bitcoin interface a plugin, so you we could ship multiple backends; you've kind of beaten us to it!

Speaking unencrypted HTTP over localhost is OK, but obv. not a great idea.  But neither is linking lightningd with libssl, so hence this seems like a great place for a plugin.

The question is, who will do the plumbing to make it a plugin?  I expect this would be post-0.7... @cdecker ?

lightningd/bitcoind.c
```diff
@@ -141,189 +55,239 @@ static void bcli_failure(struct bitcoind *bitcoind,
...
-		bcli->stopper = NULL;
+		brpc->stopper = NULL;
+
+	brpc->cmd = tal_fmt(brpc, "%s ", cmd);
+	brpc->request = tal_fmt(brpc,
+				"{\"jsonrpc\": \"1.0\", \"id\":\"lightningd\", "
+				"\"method\": \"%s\", \"params\":",
+				cmd);
+
+	tal_append_fmt(&brpc->request, "[ ");
```
cdecker: We have a library built on top of the `json_request` struct that should facilitate the creation of JSON requests. That makes this kind of operations easier.

lightningd/bitcoind.c
```diff
@@ -757,110 +692,125 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
...
-	int from, status, ret;
-	pid_t child;
-	const char **cmd = cmdarr(bitcoind, bitcoind, "echo", NULL);
-	bool printed = false;
+	struct bitcoin_rpc *brpc = tal(NULL, struct bitcoin_rpc);
+
+	if ((bitcoind->rpccookiefile == NULL) &&
+	    ((bitcoind->rpcuser == NULL) || (bitcoind->rpcpass == NULL) ||
+	     (bitcoind->rpcconnect == NULL) || (bitcoind->rpcport == 0)))
+		fatal("RPC server is not config,  See:\n"
```
cdecker: `config` -> `configured`

CHANGELOG.md
```diff
@@ -31,6 +32,7 @@ changes.
 ### Removed
 
 - JSON API: the `waitsendpay` command error return no longer includes `channel_update`
+- lightningd: remove bitcoin-cli dependcy
```
cdecker: `dependcy` -> `dependency`

cdecker:
> The question is, who will do the plumbing to make it a plugin? I expect this would be post-0.7... @cdecker ?

We can certainly move this functionality, but it'd be post-0.7 since we need to define a fine-grained interface that we ask from the plugins (instead of asking for the block, parse the transaction list, then ask for the UTXO entry, we should just ask for the scid-status).

cdecker:
Sorry for the repeated nagging @arowser, it was not my intention to send you off yet again. It's an awesome RPC CLI implementation, and I'm really looking forward to merging it.

Thanks for the time you are investing, it's very much appreciated :slightly_smiling_face:

arowser:
@cdecker No worry about that, I'm glad to learn more from your suggestion, please keep more strict with me.
In fact I try with btcd server too , its only need change several rpc command, the change in https://github.com/arowser/lightning/tree/bitcoin_rpc_btcd
And also hope we can  add neutrino support in future, so the c-lighning could be running on more embed devices.

Issue #2301, opened 2019-01-30
C-lightning can create wumbo invoices but can't pay


molxyz:
### Issue and Steps to Reproduce
I'm not sure if this is a feature intentionally or a bug that hasn't been noticed. However it seems annoying that invoices over the payment limit are allowed to create.

Here's a screenshot:  https://i.imgur.com/1293Did.png

rustyrussell:
Definitely a bug.  Will fix.

PR #2306, opened 2019-01-31
invoice: don't allow creation of unpayable invoices.

rustyrussell:
Fixes: #2301


conscott:
ACK a8f80ace1b5495cd4e51ad89cc2ba7c71a6fa7e5

If we are getting specific, you can't _receive_ an invoice for more than your incoming capacity, which itself is always changing, however, you could set a confident lower bound as the sum capacity of all active channels. 

Do you see any benefit to taking it a step further and disabling invoice creation (to yourself) for more than you could possibly receive?

PR #2307, opened 2019-02-01
Various fixes to fix the "we don't choose direct routes" problem.

rustyrussell:
This was something of a saga:

1. The test looped 8 times, but because our RNG is deliberately fixed, that didn't change our route.
2. `pay` tries routehints *first*, so sometimes we went indirect because we were following the routehint
    (which *did* change around the loop).
3. The test case only covered the simple case; a more sophisticated test reveals it's a route evaluation error.
4. We completely mishandled riskfactor in multiple(!) ways, and even when fixed it's in the noise.

Fixes: #2119
Closes: #2168
Closes: #2144


doc/lightning-getroute.7.txt
```diff
@@ -21,11 +21,10 @@ There are two considerations for how good a route is: how low the
...
 during the process.  The 'riskfactor' floating-point field controls
 this tradeoff; it is the annual cost of your funds being stuck (as a
-percentage), multiplied by the percentage chance of each node failing.
+percentage).
 
-For example, if you thought there was a 1% chance that a node would
-fail, and it would cost you 20% per annum if that happened,
-'riskfactor' would be 20.
+For example, if you thought the inconvenience of having funds stuck was
+worth 20% per annum interest, 'riskfactor' would be 20.
```
SimonVrouwe: do you mean:
For example, if you thought the convenience of not having funds stuck was
worth 20% per annum interest, 'riskfactor' would be 20.

doc/lightning-getroute.7.txt
```diff
@@ -21,11 +21,10 @@ There are two considerations for how good a route is: how low the
 fees are, and how long your payment will get stuck if a node goes down
 during the process.  The 'riskfactor' floating-point field controls
 this tradeoff; it is the annual cost of your funds being stuck (as a
-percentage), multiplied by the percentage chance of each node failing.
+percentage).
```
SimonVrouwe: percentage of what? channel capacity?

cdecker:
ACK bc1c0af16a15ba9a872e0a4051fea3c66aae1a6d

Though I seem to have gotten lost in the maxfeepercent vs fuzz factor vs riskfactor logic: a 5% fuzz allows us to use routes that are up to 5% worse than what exactly?

doc/lightning-getroute.7.txt
niftynei: Restating this to confirm I'm understanding: `riskfactor` is a way of dynamically valuing shorter routes, by accounting for the potential loss of liquidity in the case that a payment gets stuck along the route. i.e. the more hops in a route (and payment value) the higher the `riskfactor`. In some sense, this makes higher fee yet shorter routes more attractive as the possibility for getting stuck is lowered due to less chances of a link failure en route.

Issue #2300, opened 2019-01-28
Funding Tx not broadcasted and unable to withdraw


mloop1:
### Issue and Steps to Reproduce
Running the following sql command, finds 6 funding tx not broadcasted:
sqlite3 $HOME/.lightning/lightningd.sqlite3 "SELECT count(1) FROM transactions WHERE blockheight is null;"

When trying to broadcast them again using the hex(rawtx) I get the error "missing inputs" on all of them.
Same when trying to withdraw.
Example of a funding tx not broadcasted: 9c420a9ad6f3633aa3f72fafcd1f5f5d99b7c5f2d281c8faa57dd61d21a21c60

Unfortunately I also have a mixed situation in the dabase with both testnet and mainnet tx.
I'm now taking into account only mainnet tx. 

Not sure if possible to recover from such situation... any help is appreciated.
### `listpeers` output
```json
{
  "peers": [
    {
      "id": "02d28c3aac4b4f36746052a735831afbe65bc5698a7be5bd41b42fd1ddf2a1a358", 
      "connected": false, 
      "channels": [
        {
          "state": "AWAITING_UNILATERAL", 
          "scratch_txid": "1e0e5efc53b28295331adf476473dea7ab3f69469063dd766ea8ee30742e06f4", 
          "short_channel_id": "1261498:30:0", 
          "channel_id": "0d64321a1c926f51ab91834b3bb3d1a42a5ea234722b36da28c796dab6479ebd", 
          "funding_txid": "bd9e47b6da96c728da362b7234a25e2aa4d1b33b4b8391ab516f921c1a32640d", 
          "private": false, 
          "msatoshi_to_us": 9900000, 
          "msatoshi_to_us_min": 9900000, 
          "msatoshi_to_us_max": 9900000, 
          "msatoshi_total": 10000000, 
          "dust_limit_satoshis": 546, 
          "max_htlc_value_in_flight_msat": 18446744073709551615, 
          "their_channel_reserve_satoshis": 0, 
          "our_channel_reserve_satoshis": 100, 
          "spendable_msatoshi": 9800000, 
          "htlc_minimum_msat": 0, 
          "their_to_self_delay": 6, 
          "our_to_self_delay": 4, 
          "max_accepted_htlcs": 483, 
          "status": [
            "AWAITING_UNILATERAL:Attempting to reconnect"
          ], 
          "in_payments_offered": 0, 
          "in_msatoshi_offered": 0, 
[...]

PR #2310, opened 2019-02-01
json_withdraw: fix change_outputs showing forever as unconfirmed in listfunds

SimonVrouwe:
Add the change output to owned_txfilter so its entry in db will
get a confirmation_height when detected in a block by filter_block_txs

before this commit, after a 'withdraw' command, 'listfunds' would
not show our change outputs as confirmed. This was mentioned in #2300 

Modified the log message in wallet_extract_owned_outputs to
append 'CONFIRMED' when it is called with a blockheight arg.
To make distinction between (1st call) when adding owned output to the
db and (2th call) when confirmed in block.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -1132,11 +1132,11 @@ int wallet_extract_owned_outputs(struct wallet *w, const struct bitcoin_tx *tx,
 		utxo->blockheight = blockheight?blockheight:NULL;
 		utxo->spendheight = NULL;
 
-		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s",
+		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s %s",
 			  output, tx->output[output].amount,
 			  is_p2sh ? "P2SH" : "SEGWIT",
 			  type_to_string(tmpctx, struct bitcoin_txid,
-					 &utxo->txid));
+					 &utxo->txid), blockheight ? "CONFIRMED" : "");
 
 		if (!wallet_add_utxo(w, utxo, is_p2sh ? p2sh_wpkh : our_change)) {
 			/* In case we already know the output, make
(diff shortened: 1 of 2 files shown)


SimonVrouwe:
To comment to myself: I am working/thinking about the semantics of commands like `withdraw` and `fundchannel` that try to broadcast and how to make it more user friendly in case of broadcast failure.

The `withdraw` command currently only adds/updates outputs in db when broadcast succeeds, but manual broadcasts are not detected.

My idea now is to handle `withdraw` and `fundchannel` in a similar way to take into account that users may manually rebroadcast failed tx and, in the meanwhile, prevent reusing outputs or perhaps even _lock_  the wallet until `bitcoin-cli` problem is fixed. So we don't end up with stuck channels etc. that try to spend change outputs of failed broadcasts.

Will likely create separate (combined) PR for that and further discussion.

wallet/wallet.c
```diff
@@ -1132,11 +1132,11 @@ int wallet_extract_owned_outputs(struct wallet *w, const struct bitcoin_tx *tx,
 		utxo->blockheight = blockheight?blockheight:NULL;
 		utxo->spendheight = NULL;
 
-		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s",
+		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s %s",
 			  output, tx->output[output].amount,
 			  is_p2sh ? "P2SH" : "SEGWIT",
 			  type_to_string(tmpctx, struct bitcoin_txid,
-					 &utxo->txid));
+					 &utxo->txid), blockheight ? "CONFIRMED" : "");
```
cdecker:
```suggestion
					 &utxo->txid), blockheight ? " CONFIRMED" : "");
```

wallet/wallet.c
```diff
@@ -1132,11 +1132,11 @@ int wallet_extract_owned_outputs(struct wallet *w, const struct bitcoin_tx *tx,
 		utxo->blockheight = blockheight?blockheight:NULL;
 		utxo->spendheight = NULL;
 
-		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s",
+		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s %s",
```
cdecker:
```suggestion
		log_debug(w->log, "Owning output %zu %"PRIu64" (%s) txid %s%s",
```

Moving the space between txid and state into the latter avoids trailing spaces :wink:

cdecker:
I'll add the space issue into a cleanup commit :wink:

wallet/wallet.c
SimonVrouwe: :+1:

PR #2311, opened 2019-02-01
Fix docker image build from fresh clone

### LINKED ISSUE #2309: Alpine docker images not building anymore on Docker Hub
It appears that we somehow broke the alpine docker images on Docker hub used to serve `elementsproject / lightningd`.

The build complains about `git` not being present, despite it being installed just a few lines above:

```
Makefile:7: *** "ERROR: git is required for generating version information". Stop.
```

This is during the call to `make`, and happens during the version extraction in the `Makefile`:

https://github.com/ElementsProject/lightning/blob/0da40540453712b836b2a2a5383c8625137964c7/Makefile#L3-L4

I have tried to build the image at that point, but was unable to reproduce the issue, and it builds correctly on my machine.

Maybe @NicolasDorier has an idea on what might have gone wrong?

cdecker:
This was failing the docker hub builds, since the git-config retains
an absolute path to the worktree location when cloning. Copying it
over from the host system means that this path now points to a
non-existent location, which then interfered with the submodule
initialization.

This fixes it by not using the copy directly, but rather it creates a
clean clone from the copied location, including a submodule init.

Fixes #2309

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,7 +56,9 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
 ENV LIGHTNINGD_VERSION=master
 
 WORKDIR /opt/lightningd
-COPY . .
+COPY . /tmp/lightning
+RUN git clone --recursive /tmp/lightning . && \
+    git checkout $(git --work-tree=/tmp/lightning --git-dir=/tmp/lightning/.git rev-parse HEAD)
 
 ARG DEVELOPER=0
 RUN ./configure && make -j3 DEVELOPER=${DEVELOPER} && cp lightningd/lightning* cli/lightning-cli /usr/bin/
--- a/contrib/linuxarm32v7.Dockerfile
+++ b/contrib/linuxarm32v7.Dockerfile
@@ -82,7 +82,10 @@ RUN wget -q https://gmplib.org/download/gmp/gmp-6.1.2.tar.xz \
 
 COPY --from=downloader /usr/bin/qemu-arm-static /usr/bin/qemu-arm-static
 WORKDIR /opt/lightningd
-COPY . .
+COPY . /tmp/lightning
+RUN git clone --recursive /tmp/lightning . && \
+    git checkout $(git --work-tree=/tmp/lightning --git-dir=/tmp/lightning/.git rev-parse HEAD)
+
 ARG DEVELOPER=0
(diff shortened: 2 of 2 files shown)


NicolasDorier:
I tried to replicate the issue but it works on my machine, which is quite strange.... :(

Anyway I tried this PR and it does not break my build, so ACK. I would just add a small comment so that somebody reading this has an idea of why is it useful.

NicolasDorier:
Ah you might need to do this on the arm32v7 build as well.

cdecker:
Good catch, I pushed a fixup.

NicolasDorier:
Will try later today. I need to test https://github.com/ElementsProject/lightning/pull/2312 anyway. Will rebasing it on top of this PR, and see if it works as expected.

NicolasDorier:
Ok so this PR actually break the build https://circleci.com/gh/NicolasDorier/lightning/20 checking

Dockerfile
```diff
@@ -56,7 +56,10 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
 ENV LIGHTNINGD_VERSION=master
 
 WORKDIR /opt/lightningd
-COPY . .
+COPY . /tmp/lightning
+RUN git clone --recursive /tmp/lightning . && \
+    git checkout $(git --work-tree=/tmp/lightning --git-dir=/tmp/lightning/.git rev-parse HEAD) && \
+    rm -rf /tmp/lightning
```
NicolasDorier: `rm -rf /tmp/lightning/.git && rm -rf /tmp/lightning` works. `rm -rf` does not delete hidden folders maybe? Strange.

cdecker: Shouldn't change the outcome though, since this is just in the builder image which gets discarded anyway.

NicolasDorier: It does, because `rm -rf /tmp/lightning` is returning an error. (see https://circleci.com/gh/NicolasDorier/lightning/20 which fail the build)

cdecker: Could it be some sort of concurrent access? iirc `rm -rf` still recursively deletes things, so if it might be that something new was created inside those directories it can fail.

NicolasDorier:
I am unsure what happen, actually rm does not work always, I tried `sleep 1` but it does not work.

I think you should just drop `rm -rf /tmp/lightning`, it uses multi stage build so anyway this is dropped in the final image.

cdecker:
Ok, removed the `rm -rf` and squashed into a single commit. @NicolasDorier can you confirm this works as expected?

NicolasDorier:
yep, all good tACK

PR #2308, opened 2019-02-01
WIP - Add plugin RPC command args to manifest and help text

conscott:
The aim here is to allow the `help` output show expected arguments to plugin rpc commands. At the moment everything [just hardcoded](https://github.com/ElementsProject/lightning/blob/master/lightningd/plugin.c#L622-L627) to `[params]`

For example, the output of `lightning-cli help` is...
```
...
pay [params]
    Send payment specified by {bolt11} with {msatoshi}
...
```

This change adds an optional field `args` to the `rpcmethods` fields for the plugin manifest. It has been implemented for the pylightning library, also detecting when arguments are optional, and surrounding them with square brackets like lightning-cli command args. 

The result is `lightning-cli help` showing the arguments list. For example, adding  [this plugins-dir](https://github.com/conscott/c-lightning-plugins), gives help output
```
balance [unit] 
    List detailed onchain and channel balance information     

closeall [force] [timeout] 
    Close all active channels     

getnetworkinfo
    Get statistical information about the current state of the network
```

## Problem
This is a bit hacky, in particular commit b415725ea9b48c14a4eb80e6fada4a4d15d37eb5 contains the turd that swaps pointers :hankey: 

Feedback appreciated. 

If accepted I can also update `pay` plugin to provide args

PR #2315, opened 2019-02-04
Plugin params passthru

rustyrussell:
This gathered some fixes along the way too...

See also: #2308

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -19,7 +19,7 @@ def hello(plugin, name="world"):
     return s
 
 
-@plugin.method("init")
+@plugin.init()
 def init(options, configuration, plugin):
     plugin.log("Plugin helloworld.py initialized")
 
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -25,7 +25,7 @@ class Plugin(object):
     """
 
     def __init__(self, stdout=None, stdin=None, autopatch=True):
-        self.methods = {}
+        self.methods = {'init': (self._init, MethodType.RPCMETHOD)}
         self.options = {}
 
         # A dict from topics to handler functions
@@ -43,7 +43,7 @@ class Plugin(object):
         self.rpc_filename = None
         self.lightning_dir = None
         self.rpc = None
-        self.init = None
+        self.child_init = None
 
     def add_method(self, name, func):
         """Add a plugin method to the dispatch table.
@@ -158,6 +158,16 @@ class Plugin(object):
             return f
         return decorator
 
+    def init(self, *args, **kwargs):
+        """Decorator to add a function called after plugin initialization
+        """
+        def decorator(f):
+            if self.child_init is not None:
+                raise ValueError('The @plugin.init decorator should only be used once')
+            self.child_init = f
+            return f
+        return decorator
+
     def _exec_func(self, func, request):
         params = request['params']
         sig = inspect.signature(func)
@@ -265,12 +275,6 @@ class Plugin(object):
         return msgs[-1]
 
     def run(self):
-        # Stash the init method handler, we'll handle opts first and
-        # then unstash this and call it.
-        if 'init' in self.methods:
-            self.init = self.methods['init']
-        self.methods['init'] = (self._init, MethodType.RPCMETHOD)
-
         partial = ""
         for l in self.stdin:
             partial += l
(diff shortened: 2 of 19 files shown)


conscott:
I am happy to have this PR takeover, however, one of the differences between #2308 is that I made `usage` in pylightning implicit, just based on arg names. See d14e41f9b9798b186d01ff0469e69d39b83b08b5

So if you had an rpc method declared like
```
@plugin.method("hello")
def hello(plugin, first_name, last_name, end_with="!"):
    ...
```
It would inspect the method and generate usage string
```
hello first_name last_name [ends_with]
```

Maybe if you don't explicitly pass usage, you can try to do it implicitly?

Regardless the PLUGINS.md probably needs to be updated to show how `usage` should be used in the manifest

contrib/plugins/helloworld.py
```diff
@@ -19,7 +19,7 @@ def hello(plugin, name="world"):
     return s
 
 
-@plugin.method("init")
+@plugin.init()
```
conscott: This change seems to throw an error now. 

```

**Comment by rustyrussell** [comment]
Yes, your pylightning is way way way way nicer than mine! I'll steal that...

File: contrib/pylightning/lightning/plugin.py [hunk 2]
```diff
@@ -306,9 +306,20 @@ def _getmanifest(self):
                 doc = "Undocumented RPC method from a plugin."
             doc = re.sub('\n+', ' ', doc)
 
+            argspec = inspect.getargspec(func)
+            args = argspec.args[1:]
```
conscott: My assumption here was that `plugin` had to be the first arg, but @cdecker pointed out that `request` or `plugin` can be _any_ positional arg.

This need to be updated to
```
args = [arg for arg in argspec.args if arg not in ('plugin', 'request')]
```

One confusion here is that [this comment](https://github.com/ElementsProject/lightning/blob/master/contrib/pylightning/lightning/plugin.py#L64-L66)  seems to suggest `request` and `plugin` should be the last args, but that does not seem to actually matter. Maybe I am reading this wrong.

contrib/pylightning/lightning/plugin.py
```diff
@@ -306,9 +306,20 @@ def _getmanifest(self):
...
+            defaults = argspec.defaults
+
+            # Make optional args be surrounded by square brackets
+            # list regular lightning-cli commands args
+            if defaults:
```
conscott: Detecting default args below should still be correct unless someone tried to set a default for `plugin` or `request` like
```python
def hello(plugin, request=None, name="world')
```
In which case, you will know this for sure `if len(defaults) > len(args)`, in which case you could ditch method registration and print an error like "plugin and request args do not require default".

contrib/plugins/helloworld.py
conscott: This is actually a breaking change for existing plugin libs (i.e. mine :smile: ) 

Is that sufficient to require a version bump of pylightning ?

contrib/pylightning/lightning/plugin.py
conscott: Apologies for turning this commit into a :hankey: This is what you _really_ want
```python
            argspec = inspect.getargspec(func)
            defaults = argspec.defaults
            num_defaults = len(defaults) if defaults else 0
            args = []
            for idx, arg in enumerate(argspec.args):
                if arg in ('plugin', 'request'):
                    continue
                # Positional arg
                if idx < (num_defaults):
                    args.append("%s" % arg)
                # Keyword arg
                else:
                    args.append("[%s]" % arg)
```

contrib/pylightning/lightning/plugin.py
conscott: This is fixed by my latest comment above.

contrib/plugins/helloworld.py
rustyrussell: Well, plugins aren't released yet! Which is why I washed to get this right before they are...

contrib/pylightning/lightning/plugin.py
rustyrussell: Applied!

rustyrussell:
OK, that change applied, and documentation is now updated too.

conscott:
ACK 0eff8c4d1f48c3bc1baeb1ebae0e2d7a35e6bd18 :tada:

lightningd/memdump.c
```diff
@@ -154,6 +154,7 @@ static void scan_mem(struct command *cmd,
 	memleak_remove_htable(memtable, &ld->topology->txowatches.raw);
 	memleak_remove_htable(memtable, &ld->htlcs_in.raw);
 	memleak_remove_htable(memtable, &ld->htlcs_out.raw);
+	jsonrpc_remove_memleak(memtable, ld->jsonrpc);
```
cdecker: How is this not a leak?

rustyrussell: These functions are kind of misnamed :(

That just removes the pointers from the table of all pointers ("memtable"), so they don't count as leaks.

cdecker: But isn't `ld` supposed to be the root of allocations (and the only thing that is marked as not being a leak)? This is all reminiscent of mark and sweep garbage collection :wink:

PR #2326, opened 2019-02-07
pylightning: Asynchronous plugin rpc-methods and hooks

### LINKED ISSUE #1701: async keyword for functions in the Python API Client
looking at the python API client I saw that the blocking functions are supposed to have an async keyword in order to return a future. c.f.: 
https://github.com/ElementsProject/lightning/blob/cf12130627599cb76ed64f26e5c7b81bc9dc99bb/contrib/pylightning/lightning/lightning.py#L89

However I could not find any function that has this. so either delete the comment or add this to those functions

### LINKED ISSUE #2167: Implement an asynchronous dispatch to the plugin framework in pylightning
As pointed out by @ZmnSCPxj plugins currently only support a single JSON-RPC in flight at any point in time. This works fine for most interactions but long-polling cannot be done if other requests need to be served as well.

The idea is to introduce an additional keyword to the `@plugin.method` decorator called `async`. If this keyword is provided, the method must also have an argument called `request` which will be handed an instance of a new `AsyncResult` class that can be passed around and replied to eventually.

This does not introduce any multi-threaded or async primitives itself, leaving the choice to the users whether they want to use multi-threading, executors, greenlets, or whatever.

cdecker:
This PR allows RPC methods and hook-callbacks to defer returning a result by adding the `sync=False` argument to the `@plugin.hook` or `@plugin.method` annotations. If the annotation includes the `sync=False` argument it'll disregard the return value of the callback function. The function can then use the `set_result` or `set_exception` methods on the `request` object that they get as a parameter, when it suits them.

This allows passing a pending request off to another thread, stashing them in the local state, or any other operation, and still return a result at a later point in time. `set_exception` and `set_result` are threadsafe and can be called from anywhere.

I'm still a bit unsure if `sync` is the best name for the argument, I wanted to use `async` but the linter yells at me that `async` will become a keyword in python 3.8, so no luck there. Maybe @renepickhardt or our namer in chief @niftynei has a good idea? :wink: 

Fixes #2167


tests/plugins/asynctest.py
```diff
@@ -0,0 +1,28 @@
...
+
+The plugin registers a method `callme` with an argument. All calls are
+stashed away, and are only resolved on the fifth invocation. All calls
+will then return the argument of the fifth call.
+
+"""
+from lightning import Plugin
+
+plugin = Plugin()
+requests = []
```
conscott: you can just dynamically attach this to object instance
```
plugin.requests = []
```
Rather that using globals below.

I suspect in the future people may also just subclass Plugin and add their local state there

contrib/pylightning/lightning/plugin.py
```diff
@@ -39,13 +113,15 @@ def __init__(self, stdout=None, stdin=None, autopatch=True):
         if os.getenv('LIGHTNINGD_PLUGIN') and autopatch:
             monkey_patch(self, stdout=True, stderr=True)
 
-        self.add_method("getmanifest", self._getmanifest)
+        self.add_method("getmanifest", self._getmanifest, sync=True)
```
conscott: I don't have an whimsical ideas, but if `async` is taken, how about `is_async` ?

cdecker: Here's a few variants that I thought of that might also work: `direct=False`, `background=True`, `_async=True`.

I tend to like `background` since it sort of describes that we want the thing to continue in the background.

tests/plugins/asynctest.py
cdecker: Right, will fixup

cdecker: Decided to initialize with a  `@plugin.init()` decorator to make sure we have the new field whenever it is used.

contrib/pylightning/lightning/plugin.py
niftynei: +1 for `background`. i also like `wait=False`

rustyrussell:
It seems weird to me to use a variant of the same annotation for a completely different beast.

After all, these functions will look quite different, won't they? Maybe @method_async ?

cdecker:
I went with the `background` proposal, since that is the best description I could come up with. It's best not to have too many decorators since the interface gets too terse imho.

conscott:
Concept ACK - tested this before, but it needs rebase

As for having this is a new decorator vs decorator arg, I don't have a super strong option, but I see rusty's case. A decorator should do _one_ predictable thing and the `background` significantly changes the behavior, and expected way to handle the result. 

You guys have better visibility on how many foreseeable decorators may be apart of the plugin lib, so I can't make a call on that part, which may be a valid concern. 

Look there, I managed to agree with you both and say nothing :)

cdecker:
> As for having this is a new decorator vs decorator arg, I don't have a super strong option, but I see rusty's case. A decorator should do _one_ predictable thing and the `background` significantly changes the behavior, and expected way to handle the result.

Ok, I consider myself overruled, and will create a separate decorator instead. tbh I didn't have a super strong reason to go for the single decorator, but you convinced me :-)

PR #2328, opened 2019-02-07
short_channel_id: don't accept :-separated in JSON if --allow-deprecated-apis=false

rustyrussell:
We need to still accept it when parsing the database, but this flag
should allow upgrade testing for devs building on top.

### DIFF
--- a/bitcoin/short_channel_id.c
+++ b/bitcoin/short_channel_id.c
@@ -36,7 +36,8 @@ bool mk_short_channel_id(struct short_channel_id *scid,
 }
 
 bool short_channel_id_from_str(const char *str, size_t strlen,
-			       struct short_channel_id *dst)
+			       struct short_channel_id *dst,
+			       bool may_be_deprecated_form)
 {
 	u32 blocknum, txnum;
 	u16 outnum;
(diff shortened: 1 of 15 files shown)


cdecker:
Wow that's ugly, can't we use a global for this? It's really weird having to wire these things through the entire call-stack only to later remove them once we have really deprecated them.

bitcoin/short_channel_id.c
```diff
@@ -48,7 +49,7 @@ bool short_channel_id_from_str(const char *str, size_t strlen,
 
 #ifdef COMPAT_V062
 	/* Pre-adelaide format vs. post-adelaide format */
-	if (strchr(buf, ':'))
+	if (may_be_deprecated_form && strchr(buf, ':'))
```
wythe: My only thought: It *might* be cleaner if we communicated `deprecated_apis` (which is a global) to the subdaemons and then use that  here directly instead of `may_be_deprecated`. But it might not.

rustyrussell:
No, because the DB needs this compat mode, and the JSON API depends on the flag :(

wythe:
Ok.  
Ack https://github.com/ElementsProject/lightning/pull/2328/commits/7f402e91075d34a50da860dd2214e9e6a3ba0119

PR #2329, opened 2019-02-08
doc/REPRODUCIBLE.md: how to make a reproducible zipfile and Ubuntu build.

rustyrussell:
The final commit contains instructions, but to reproduce the results listed there you need to use the ~PREVIOUS~ rc1 commit!

You can find my results and working at https://ozlabs.org/~rusty/clightning-repro


cdecker:
The linked image is no longer available :frowning_face:

PR #2335, opened 2019-02-10
fix: add inline exception for recent cppcheck false positive

m-schmoock:
Recent cppcheck version (at least 1.86) will have a false positive in `gossipd/routing.c` line 903. The cppcheck misunderstands ccan's external memory management by assigning it to a local struct pointer:

```C
struct pending_cannouncement *pending;
pending = tal(rstate, struct pending_cannouncement); 
// ... 
*scid = &pending->short_channel_id;
```

It complains with `[gossipd/routing.c:903]: (error) Address of local auto-variable assigned to a function parameter.` 

When the diff is applied, test are able to run normally, as cppcheck is a required precondition. I didnt find a better way than to add and enable a exception.


rustyrussell:
Hmm, that's a bit nasty.  Of course, cppcheck 1.84 here works fine... But agreed, we should suppress it.

Thanks!

PR #2336, opened 2019-02-10
fix: change type of config->fee_per_satoshi to uint32

m-schmoock:
fix: change type of config->fee_per_satoshi to uint32

This patch will properly set fee_per_satoshi to _unsigned_ integer,
as support for negative fees was removed from overall design.

This change does not break any tests, so I assume its better this way.


rustyrussell:
Great catch, this is never used in arithmetic (it's handed to subdaemons as a a u32 anyway), it's a relic as you expected.

PR #2341, opened 2019-02-11
plugin: Connected hook

cdecker:
This is a simpler example than the `htlc_accepted` hook that I'm still working on, but it already allows us to do quite some interesting things. For example it'd allow us to implement a drain-plugin that allows connections to and from peers we are still negotiating a close with, but reject all connections that could result in a new channel.

It also serves as an example on how hooks are implemented.

### DIFF
--- a/lightningd/json.c
+++ b/lightningd/json.c
@@ -299,6 +299,11 @@ void json_add_bool(struct json_stream *result, const char *fieldname, bool value
 	json_add_member(result, fieldname, value ? "true" : "false");
 }
 
+void json_add_null(struct json_stream *stream, const char *fieldname)
+{
+	json_add_member(stream, fieldname, "null");
+}
+
 void json_add_hex(struct json_stream *result, const char *fieldname,
 		  const void *data, size_t len)
 {
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -432,6 +432,187 @@ struct peer_connected_hook_response {
 	enum peer_connected_hook_result result;
 };
 
+static void json_add_htlcs(struct lightningd *ld,
+			   struct json_stream *response,
+			   const struct channel *channel)
+{
+	/* FIXME: make per-channel htlc maps! */
+	const struct htlc_in *hin;
+	struct htlc_in_map_iter ini;
+	const struct htlc_out *hout;
+	struct htlc_out_map_iter outi;
+
+	/* FIXME: Add more fields. */
+	json_array_start(response, "htlcs");
+	for (hin = htlc_in_map_first(&ld->htlcs_in, &ini);
+	     hin;
+	     hin = htlc_in_map_next(&ld->htlcs_in, &ini)) {
+		if (hin->key.channel != channel)
+			continue;
+
+		json_object_start(response, NULL);
+		json_add_string(response, "direction", "in");
+		json_add_u64(response, "id", hin->key.id);
+		json_add_u64(response, "msatoshi", hin->msatoshi);
+		json_add_u64(response, "expiry", hin->cltv_expiry);
+		json_add_hex(response, "payment_hash",
+			     &hin->payment_hash, sizeof(hin->payment_hash));
+		json_add_string(response, "state",
+				htlc_state_name(hin->hstate));
+		json_object_end(response);
+	}
+
+	for (hout = htlc_out_map_first(&ld->htlcs_out, &outi);
+	     hout;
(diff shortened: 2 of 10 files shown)


cdecker:
Need to debug the CI failure tomorrow, hence the WIP tag :wink:

cdecker:
Should be ready to be reciewed, turns out it was just the switch from `@plugin.method('init')` to `@plugin.init()` decorators.

Ping @rustyrussell

lightningd/peer_control.c
```diff
@@ -411,39 +412,51 @@ void channel_errmsg(struct channel *channel,
...
-{
-	struct pubkey id;
-	struct crypto_state cs;
-	u8 *globalfeatures, *localfeatures;
-	u8 *error;
+/* Possible outcomes returned by the `connected` hook. */
+enum peer_connected_hook_result {
+	peer_connected_continue,
+	peer_connected_disconnect,
+};
```
rustyrussell: enum constants should be CAPS...  I know it seems shouty but there it is

wire/fromwire.c
```diff
@@ -253,7 +253,7 @@ REGISTER_TYPE_TO_HEXSTR(channel_id);
  * (i.e. `funding_output_index` alters the last 2 bytes).
  */
 void derive_channel_id(struct channel_id *channel_id,
-		       struct bitcoin_txid *txid, u16 txout)
+		       const struct bitcoin_txid *txid, u16 txout)
```
rustyrussell: :+1:

lightningd/peer_control.c
```diff
@@ -617,6 +617,21 @@ static void
...
+	    stream, "addr",
+	    type_to_string(stream, struct wireaddr_internal, &payload->addr));
+	json_add_hex_talarr(stream, "globalfeatures", p->globalfeatures);
+
+	if (channel != NULL) {
+		json_add_channel(payload->ld, stream, "channel", channel);
+	} else {
+		json_add_null(stream, "channel");
+	}
+	json_object_end(stream); /* .peer */
```
rustyrussell: I think this hook may be far too helpful.  Just hand it the peer id, address and the global and local features from this connection message; the rest it can get via an RPC call if it wants?

lightningd/json.c
```diff
@@ -299,6 +299,11 @@ void json_add_bool(struct json_stream *result, const char *fieldname, bool value
 	json_add_member(result, fieldname, value ? "true" : "false");
 }
 
+void json_add_null(struct json_stream *stream, const char *fieldname)
+{
+	json_add_member(stream, fieldname, "null");
+}
```
rustyrussell: This commit should precede the prior one, but moreover: I think the usage here is wrong.  If a field doesn't exist it shouldn't exist, not be set to 'null'.  That's how we handle every other optional field.

lightningd/peer_control.c
```diff
@@ -638,7 +638,20 @@ static struct peer_connected_hook_response *
...
+	resulttok = json_get_member(buffer, toks, "result");
+	if (!resulttok) {
+		return NULL;
+	}
+
+	resp = tal(ctx, struct peer_connected_hook_response);
+	if (json_tok_streq(buffer, resulttok, "continue"))
+		resp->result = peer_connected_continue;
+	else
+		resp->result = peer_connected_disconnect;
```
rustyrussell: Probably best to be explicit about the alternative to continue, eg "disconnect"?  Otherwise typos can have bad effects, and its hard to add other return types in the future (eg. an error message to send or something?).

cdecker:
Thanks @rustyrussell for the feedback, I rebased and amended as follows:

 - Enums are now extra shouty (https://github.com/ElementsProject/lightning/pull/2341/commits/a1f6656e86b705e9e8947b92a4ff3ebfe7d90bea and https://github.com/ElementsProject/lightning/pull/2341/commits/4a71ec075902142b987dee7e380a9a28a8c436d0) :wink: 
 - The `connected_hook` is now less helpful, but contains 100% more local features (https://github.com/ElementsProject/lightning/pull/2341/commits/a09dc4941f56c1bbba4eb528e7490da3f0741b49)
 - Adjusted the docs to mention where to get additional info if the hook isn't helpful enough (https://github.com/ElementsProject/lightning/pull/2341/commits/82263617b1ddd3f17be36077dd25d6099c9e4963)

I totally agree that the hook call was too heavy :+1:

cdecker:
Interpreting the approval above as an ACK:

ACK 968e9127e255f75875e54330205cd0b354ca9d72

Issue #2316, opened 2019-02-04
Allow setting feerate on per-channel basis


rustyrussell:
As requested by ctrlbreak on IRC: documenting here.

- [ ] Add db fields for per-channel feebase + ppm
- [ ] Put these fields into struct channel
- [ ] Use the fields in forward_htlc instead of global values
- [ ] Show fields in listchannels
- [ ] Plumb db fields through to gossipd for advertising.
- [ ] Provide API eg. `setchannelfee <short-chanid-or-peerid> <base> <ppm>`
- [ ] Create tests, CHANGELOG.md, man page.
- [ ] Dance a happy dance.

renepickhardt:
also another question: might it be reasonable to implement this as a plugin. In this way more hooks and events that people would eventually want to expose would have to be implemented in c-lightning

m-schmoock:
Hey, I think I can have a look at this. But it would be, besides isolating issue https://github.com/ElementsProject/lightning/pull/2321 , my first contribution here, so please expect me being exeptionally fast :D

m-schmoock:
Started working on it: 7c5cdee682ff4a4bdc2d6ce0f590ae32e5cf18aa

m-schmoock:
@rustyrussell can you explain what "ppm" means in this context?

niftynei:
@m-schmoock i believe it means 'part per millionth' -- the total fee is a 'percent' of the amount routed, in millionths plus the base

m-schmoock:
Can we have a branch for this where a nobody like me is able to push?

m-schmoock:
Here is my workspace: https://github.com/ElementsProject/lightning/compare/master...m-schmoock:feerate-per-channel

I will add the API function `setchannelfee` tomorrow.

m-schmoock:
Hey, two quick questions: 

- [lightningd/lightningd.h line 43](https://github.com/ElementsProject/lightning/blob/master/lightningd/lightningd.h#L43) Is there a reason why the config.fee_per_satoshi value is a signed integer? 
- [lightningd/lightningd.h line 495ff](https://github.com/ElementsProject/lightning/blob/master/lightningd/peer_htlcs.c#L495) the comment tells that the fee is calculated using `fee_proportional_millionths / 1000000` but later in the code its `fee_per_satoshi / 1000000`. Is  ppm == fee_per_satoshi or is the comment misleading?

cdecker:
[...]

PR #2342, opened 2019-02-11
Allow setting feerate on per-channel basis

m-schmoock:
# PR for issue #2316  
Progress:

> * [x]  Add db fields for per-channel feebase + ppm
> * [x]  Put these fields into struct channel
> * [x]  Use the fields in forward_htlc instead of global values
> * [x]  Show fields in listchannels
> * [x]  Plumb db fields through to gossipd for advertising.
> * [x]  Provide API eg. `setchannelfee id <base> <ppm>`
> * [x]  Create tests, CHANGELOG.md, man page.
> * [x]  Dance a happy dance.

Currently workin on: **FINALIZATIONS**

Maybes:
- Maybe constraint checks / warning when trying to set too high fees? If yes, which exactly?
- Maybe allow wildcard '*' or 'all' as `channel id` in json command to set fees for all channels. Note: will be separate PR and is already done.

Issues:
- None

Notes:
- JSON: Leaving out values for ppm/base values will re-apply global config values.
- COMMON: I added some amount_msat function in order to check and transform them to u32, if not wanted, I need other access method, or pointer magic, or use old style u32 number.
- DB: the fix #2444 to the issue #2442 now raises a valgrind complaint which cause the travis to be red. See #2449 for details.


m-schmoock:
I did good progress. only test finilazations, changelog, manpage and review missing

m-schmoock:
@cdecker I'm almost done here. Need a bit input on the remaining issues:

* are zero values allowed for <base> or <ppm>?
* do there have to be any min/max checks for the values, if so, which exactly?
* in what channel states am I allowed to send an update for the fees other than state NORMAL?
* suggestion: shoud I use a -1 in the databse to disable the specific fees or an additional boolean flag value? Or maybe 0 (only applies if zero is disallowed)

lightningd/opening_control.c
```diff
@@ -212,7 +212,9 @@ wallet_commit_channel(struct lightningd *ld,
 			      true,
 			      &uc->local_basepoints,
 			      &uc->local_funding_pubkey,
-			      NULL);
+			      NULL,
+			      feerate,
```
cdecker: This is most likely the wrong feerate. The `feerate` variable concerns on-chain fees for the funding transaction, not the base fee leveraged off-chain.

lightningd/peer_htlcs.c
```diff
@@ -478,28 +478,35 @@ static void forward_htlc(struct htlc_in *hin,
...
+
 	/* BOLT #7:
 	 *
 	 * The origin node:
 	 *   - SHOULD accept HTLCs that pay a fee equal to or greater than:
 	 *     - fee_base_msat + ( amount_to_forward * fee_proportional_millionths / 1000000 )
 	 */
-	if (mul_overflows_u64(amt_to_forward,
-			      ld->config.fee_per_satoshi)) {
+	if (mul_overflows_u64(amt_to_forward, ld->config.fee_per_satoshi)) {
```
cdecker: Please avoid whitespace changes.

lightningd/peer_htlcs.c
```diff
@@ -478,28 +478,35 @@ static void forward_htlc(struct htlc_in *hin,
...
 	if (!next || !next->scid) {
 		local_fail_htlc(hin, WIRE_UNKNOWN_NEXT_PEER, NULL);
 		return;
 	}
 
+	/* Fee base and ppm can also be set per channel */
+	if (next->feerate_base > 0) {
+		fee_base = next->feerate_base;
+	}
+	// TODO: next->feerate_ppm
```
cdecker: The commit in which you address this should be squashed with this one, to keep changes logically close together.

lightningd/peer_htlcs.c
```diff
@@ -491,23 +493,22 @@ static void forward_htlc(struct htlc_in *hin,
...
+		fee_ppm = next->feerate_ppm;
 
 	/* BOLT #7:
 	 *
 	 * The origin node:
 	 *   - SHOULD accept HTLCs that pay a fee equal to or greater than:
 	 *     - fee_base_msat + ( amount_to_forward * fee_proportional_millionths / 1000000 )
 	 */
-	if (mul_overflows_u64(amt_to_forward, ld->config.fee_per_satoshi)) {
+	if (mul_overflows_u64(amt_to_forward, fee_ppm)) { // TODO: why is /= 1000000 missing here?
```
cdecker: Most likely because the `/ 10**6` is applied after the multiplication would already have overflown.

lightningd/peer_htlcs.c
```diff
@@ -504,7 +504,7 @@ static void forward_htlc(struct htlc_in *hin,
 	 *   - SHOULD accept HTLCs that pay a fee equal to or greater than:
 	 *     - fee_base_msat + ( amount_to_forward * fee_proportional_millionths / 1000000 )
 	 */
-	if (mul_overflows_u64(amt_to_forward, fee_ppm)) { // TODO: why is /= 1000000 missing here?
+	if (mul_overflows_u64(amt_to_forward, fee_ppm)) {
```
cdecker: This commit needs to be squashed

lightningd/opening_control.c
```diff
@@ -174,6 +174,10 @@ wallet_commit_channel(struct lightningd *ld,
 	/* old_remote_per_commit not valid yet, copy valid one. */
 	channel_info->old_remote_per_commit = channel_info->remote_per_commit;
 
+    // init new channel with global fee settings per default
+    u32 feerate_base = ld->config.fee_base;
```
cdecker: This fixes the feerate to the configured value at the time the channel was opened, which may not always be the desired effect. A user may restart its node with different fees, but all its channels are pinned to the config value before the change. Since you are checking for `0` you might pass in `0` here and default to the config value instead. Otherwise this needs documenting.

cdecker: Also if a channel defaults to `0` you should probably remove the arguments to `new_channel` again, since they're now fixed again.

lightningd/peer_control.c
```diff
@@ -1192,6 +1192,9 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 	const jsmntok_t *idtok;
 	unsigned int *base;
 	unsigned int *ppm;
+	struct peer *peer;
+	/* FIXME: gcc 7.3.0 thinks this might not be initialized. */
```
cdecker: This should be easy to fix by returning in the `else` branch on line 1213 always.

lightningd/peer_control.c
```diff
@@ -1200,10 +1203,39 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 		   NULL))
 		return command_param_failed();
 
-	// TODO: write this stuff
+	// find the channel
```
cdecker: We don't use `//`-style comments.

lightningd/peer_control.c
```diff
@@ -1200,10 +1203,39 @@ static struct command_result *json_setchannelfee(struct command *cmd,
...
+	}
+	if (!channel)
+		return command_fail(cmd, LIGHTNINGD, "Could not find peer with that id");
+
+	// set actual values
+	channel->feerate_base = *base;
+	channel->feerate_ppm = *ppm;
+	// TODO: call channel save manually?
+
+	struct channel_id cid;
```
cdecker: Variable declarations should be at the top of the function.

lightningd/peer_control.c
```diff
@@ -1200,10 +1203,39 @@ static struct command_result *json_setchannelfee(struct command *cmd,
...
-	// TODO: write this stuff
+	// find the channel
+	peer = peer_from_json(cmd->ld, buffer, idtok);
+	if (peer) {
+		channel = peer_active_channel(peer);
+		if (!channel)
+			return command_fail(cmd, LIGHTNINGD,
+				    "Could not find active channel of peer with that id");
+	} else {
+		struct command_result *res;
```
cdecker: Variable declarations should be at the top of the function.

lightningd/peer_control.c
```diff
@@ -1200,10 +1203,39 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 		   NULL))
 		return command_param_failed();
 
-	// TODO: write this stuff
+	// find the channel
+	peer = peer_from_json(cmd->ld, buffer, idtok);
+	if (peer) {
+		channel = peer_active_channel(peer);
+		if (!channel)
```
cdecker: You can un-nest the `if`-statements if you move the `!peer` case up, and always return in it. Then the `if (!channel)` case can be unindented.

lightningd/opening_control.c
```diff
@@ -217,8 +213,8 @@ wallet_commit_channel(struct lightningd *ld,
 			      &uc->local_basepoints,
 			      &uc->local_funding_pubkey,
 			      NULL,
-			      feerate_base,
-			      feerate_ppm);
+			      /* channel specific feerates are disabled with 0 */
```
cdecker: Commit ought to be squashed.

tests/test_pay.py
```diff
@@ -1407,7 +1407,7 @@ def test_setchannelfees(node_factory, bitcoind):
     bitcoind.generate_block(6)
 
     # get short channel id
-    cid = l1.db_query('SELECT short_channel_id from channels;')[0]['short_channel_id'].decode("utf-8")
+    cid = l1.rpc.listfunds()['channels'][0]['short_channel_id']
```
cdecker: Squash this with the previous commit.

lightningd/peer_control.c
```diff
@@ -1219,11 +1219,12 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 	if (!channel)
 		return command_fail(cmd, LIGHTNINGD, "Could not find peer with that id");
 
-	// set actual values
+	// set and save new values
 	channel->feerate_base = *base;
 	channel->feerate_ppm = *ppm;
-	// TODO: call channel save manually?
+	wallet_channel_save(cmd->ld->wallet, channel);
```
cdecker: Squash

tests/test_pay.py
```diff
@@ -1425,3 +1425,17 @@ def test_setchannelfees(node_factory, bitcoind):
     feerate_ppm_specific = fees_specific[0]['feerate_ppm']
     assert(feerate_base_specific == 1337)
     assert(feerate_ppm_specific == 137)
+
+    # check if negative fees raise error and DB keeps values
+    raised = None
+    try:
+        l1.rpc.setchannelfee(cid, -1, -1)
+    except RpcError as err:
```
cdecker: You can do this with the `with pytest.raises(RpcError)` context much easier. It'll fail if the code within doesn't raise the expected exception and allows matchers on the exceptions.

channeld/channeld.c
```diff
@@ -263,6 +263,15 @@ static void send_channel_update(struct peer *peer, int disable_flag)
 
 	assert(peer->short_channel_ids[LOCAL].u64);
 
+	// TODO: set channel specific fees for own outgoing channels
```
cdecker: This commit ought to be squashed as well.

lightningd/channel_control.c
```diff
@@ -352,8 +360,8 @@ void peer_start_channeld(struct channel *channel,
 				      &channel->channel_info.remote_per_commit,
 				      &channel->channel_info.old_remote_per_commit,
 				      channel->funder,
-				      cfg->fee_base,
-				      cfg->fee_per_satoshi,
+				      fee_base,
```
cdecker: Were the previous commits failing the CI tests without this commit? If yes we need to re-arrange the commits so that no commit fails. This is important for bisectability, i.e., being able to test for a specific failure on a range of changes to find where a bug was introduced. If the commits in the commit history do not pass the regular tests, this is made a lot harder.

m-schmoock:
Thx for the initial feedback, I just wasnt expecting you to go to each commit seperately, but discuss the open issues and do a diff with master and squash :D

cdecker:
> Thx for the initial feedback, I just wasnt expecting you to go to each commit seperately, but discuss the open issues and do a diff with master and squash :D

I find it easier to walk through the commits since it allows me to follow the rationale, rather than trying to split the final diff into logical chunks :wink:

lightningd/opening_control.c
m-schmoock: not in final version, redoing branch

lightningd/peer_control.c
m-schmoock: this was actually a copyNpaste from same file which contains same logic for function:
```C
static struct command_result *json_close(struct command *cmd, ...)
```

Always returning in else is imho not an option, as `command_find_channel` sets the pointer as fourth argument: `&channel`.  Or maybe I misunderstood your suggestion. I can remove the comment though, as initializing with NULL seems fine to me :)

lightningd/peer_control.c
m-schmoock: same here, in-function declaration done same file function:
```C
static struct command_result *json_close(struct command *cmd, ...)
```

as i see stuff like this, is it not my task to correct other code? if so, who does the cleanup sometime?

lightningd/peer_control.c
m-schmoock: okay, thx

lightningd/peer_control.c
m-schmoock: No, we cant do this, same reason because the function returns the error and sets a pointer (the one gcc may complain unset): 

```C
res = command_find_channel(cmd, buffer, idtok, &channel);
```

we could rename `res` to `err` to make it more clear. If it returns NULL it means it did set the `&channel`. Im just using your code here :]

tests/test_pay.py
m-schmoock: thx, helpful indeed

lightningd/channel_control.c
m-schmoock: thx, will do. I will completely redo the branch, force push here and we can hopefully do a cleaner second round :D

lightningd/peer_control.c
m-schmoock: sounds legit, will be fixed

lightningd/opening_control.c
m-schmoock: Both issues fixed by later commit.

m-schmoock:
@cdecker can we do another try for the review? I reorganized the branch, rebased and tested each subrevision on the current master.

Now I also understand why we are doing it this way, by isolating each commit I found out that there were inconsistencies in old my changes.

Note: As I wrote at the top of the PR, I have open questions regarding min/max constraint checks and zero values...

lightningd/opening_control.c
```diff
@@ -223,7 +223,9 @@ wallet_commit_channel(struct lightningd *ld,
 			      true,
 			      &uc->local_basepoints,
 			      &uc->local_funding_pubkey,
-			      NULL);
+			      NULL,
+			      /* Channel specific feerates are disabled with 0 */
+			      0, 0);
```
rustyrussell: But zero is a valid value, which is a problem!  We should simply initialize the values to the globals if not set.

wallet/db.c
```diff
@@ -359,6 +359,9 @@ char *dbmigrations[] = {
     /* Fix dangling peers with no channels. */
     "DELETE FROM peers WHERE id NOT IN (SELECT peer_id FROM channels);",
     "ALTER TABLE outputs ADD scriptpubkey BLOB;",
+    /* #2316 feerate per channel */
+    "ALTER TABLE channels ADD feerate_base INTEGER;",
+    "ALTER TABLE channels ADD feerate_ppm INTEGER;",
```
rustyrussell: Yeah, this lacks sophistication.  We really want to set these to the current global values, but our db upgrade code is dumb.  @cdecker might have ideas, but perhaps each upgrade should be a pair: a string and a function, either of which can be NULL.  The function in this case would initialize those feerate_ vars correctly?

lightningd/peer_htlcs.c
```diff
@@ -500,22 +500,28 @@ static void forward_htlc(struct htlc_in *hin,
...
 		local_fail_htlc(hin, WIRE_UNKNOWN_NEXT_PEER, NULL);
 		return;
 	}
 
+	/* Fee base and ppm can also be set per channel */
+	if (next->feerate_base > 0)
+		fee_base = next->feerate_base;
+	if (next->feerate_ppm > 0)
+		fee_ppm = next->feerate_ppm;
+
```
rustyrussell: I would always initialize it, and node do this here.

channeld/channel_wire.csv
```diff
@@ -23,6 +23,8 @@ channel_init,,old_remote_per_commit,struct pubkey
 channel_init,,funder,enum side
 channel_init,,fee_base,u32
 channel_init,,fee_proportional,u32
+channel_init,,fee_base_global,u32
+channel_init,,fee_proportional_global,u32
 channel_init,,local_msatoshi,struct amount_msat
```
rustyrussell: I don't think this is channeld's business: it just needs to know fee_base and fee_proportional.  Not sure this commit is needed at all?

channeld/channel_wire.csv
```diff
@@ -180,3 +180,7 @@ channel_dev_memleak_reply,,leak,bool
 channel_fail_fallen_behind,1028
 channel_fail_fallen_behind,,remote_per_commitment_point,struct pubkey
 
+# Handle a channel specific feerate base ppm configuration
+channel_json_feerates,1029
+channel_json_feerates,,feerate_base,u32
+channel_json_feerates,,feerate_ppm,u32
```
rustyrussell: The word 'json' is anomolous here.  Sure, the feerate change is probably driven by a JSON command somewhere, but that's not anything to do with channeld; 'channel_set_feerates' would be a better name?

lightningd/peer_control.c
```diff
@@ -1343,6 +1343,89 @@ static const struct json_command getinfo_command = {
...
+				    "Could not find active channel of peer with that id");
+	} else {
+		res = command_find_channel(cmd, buffer, idtok, &channel);
+		if (res)
+			return res;
+	}
+
+	/* check channel is found and in valid state */
+	if (!channel)
+		return command_fail(cmd, LIGHTNINGD, "Could not find peer with that id");
```
rustyrussell: Move this check up before '}' and change message to "Could not find channel with that id"?

lightningd/peer_control.c
```diff
@@ -1343,6 +1343,89 @@ static const struct json_command getinfo_command = {
...
+	/* FIXME: gcc 7.3.0 thinks this might not be initialized. */
+	struct channel *channel = NULL;
+	struct channel_id cid;
+	struct command_result *res;
+
+	if (!param(cmd, buffer, params,
+		   p_req("id", param_tok, &idtok),
+		   p_opt_def("base", param_number, &base, 0),
+		   p_opt_def("ppm", param_number, &ppm, 0),
+		   NULL))
```
rustyrussell: Hmm, should not providing one of these leave it unchanged, or return it to the global value?

I would be tempted to side-step the debate by insisting on both values for now, and as suggested before, removing the 0 == global idea.

lightningd/peer_control.c
```diff
@@ -1343,6 +1343,89 @@ static const struct json_command getinfo_command = {
...
+	return command_success(cmd, response);
+}
+
+static const struct json_command setchannelfee_command = {
+	"setchannelfee",
+	json_setchannelfee,
+	"Sets specific routing fees for channel with {id} "
+	"(either peer ID, or short channel ID). "
+	"The routing fees are calclulated by a fixed {base} "
+	"and a proportional {ppm}. Params are in MSAT, "
```
rustyrussell: parameters are not in msat (well, base is).  You *could* use param_msat for that, in fact (which lets them specify it in sat if they want, for example), but if you do make sure you check it doesn't exceed 32 bits!

doc/lightning-setchannelfee.7.txt
```diff
@@ -0,0 +1,64 @@
+LIGHTNING-SETCHANFEE(7)                                LIGHTNING-SETCHANFEE(7)
```
rustyrussell: This doesn't look like asciidoc?  You should copy one of the existing .txt files and add it to the doc/Makefile, then run 'make doc-all' to generate lightning-setchannelfee.7.

lightningd/peer_control.c
m-schmoock: I also would like to return the global value if unchanged or not set, but I need to have access to these values within the context of peer_control, which currently does not have access to global `ln->config` values.

Should I introduce these changes?

m-schmoock:
@cdecker @rustyrussell Thanks for all your your input. I will come back with an update.

**No Magic Zero**
In order to support '0' fees without disabling it (which btw I also find very uncool), I think I need one of the following databse changes:

1. Allow -1 in the database to indicate that channel specific fee is turned of, which is kinda iffy, but probably the least invasive solution.
2. **OR** Add one or two bool (database Integers) that indicate the activation of channel specific fees. Two are needed if one of the args base/ppm can be optional, which I find good.
3. **OR** Add one Integer field that is interpreted bitwise to check if channel specific values ought to be applied or not.

**Return global values if unset/unchanged/defaulted**
As written above, I need to add code to access global `ln->config` within the `lightningd/peer_control.c`. Should I add it?

**Min/Max constraint checks**
Should I add protection for the user against himself, trying to set too low/high values? If so, what values and should it just return a warning or prevent usage? 

Can you please give me an indication which of the solutions you find appropriate?

lightningd/opening_control.c
m-schmoock: Thanks, will find a way to allow zero then, this was an open question to me that was not addressed yet.

lightningd/peer_control.c
rustyrussell: See https://github.com/ElementsProject/lightning/pull/2394 :)

channeld/channel_wire.csv
m-schmoock: `channel_set_feerates` is ambiguous, as there already exists a `channe_feerates (1027)` message which is meant to set the onchain rates. This was my reason for calling it `channel_json_feerates`. 

So make it *a bit* clearer im going to call it now: `channel_specific_feerates`

wallet/db.c
m-schmoock: hmmm..  I see what you are up to. "DB: more advanced upgrade" #2394  would allow me to read out the current global values and set all channels specific feerate values to the current global on DB update.

I kinda dislike the idea, because if on a existing clightning node someone changes the gobal feerate config, it will not affect **any** of exiting channels, which is probably not what the user expects here. Is this really what you want?

My suggestion is maybe to have a third DB value indicating the usage of specific or global fees or maybe allow -1 for that matter. This way specific rates will only apply when set and not disabled yet, global values are normal operation 'fallback'.

lightningd/peer_control.c
m-schmoock: same here, if I set all channels to current global values, they will always be applied even if global values change. is this intended/expected?

m-schmoock: other than that, good change on the db migrations :D

lightningd/peer_control.c
m-schmoock: okay did so, but now its actually a "cannot happen exception", since `comman_find_channel` will always have channel set if no error was returned.

doc/lightning-setchannelfee.7.txt
m-schmoock: thx for the hint

lightningd/peer_control.c
m-schmoock: yes, thx. I now do the 32bit overflow check and use `base->millisatoshis`, works but complains about code style: `[Makefile:332: check-amount-access] Error 1`

How am I supposed to access a checked value?

m-schmoock: update: probably not like this :D
```C
u32 value = *(u32*)base;
```

channeld/channel_wire.csv
m-schmoock: thanks, I found a better way

lightningd/peer_htlcs.c
m-schmoock: fixed

wallet/db.c
m-schmoock: My suggestion since 2394 is not merged yet and we can have a global fallback anytime:

https://github.com/ElementsProject/lightning/pull/2342/commits/40af7d1910d892d2b7d40349fd52b8efa463c00c

m-schmoock:
@cdecker @rustyrussell 
I need another review and final feedback. I integrated your inputs and remarks. The last commit is my suggestion on how to handle DB update and keep global fallback option. If you rather want it in #2394 style without fallback, I can do that when #2394 is merged.

wallet/db.c
rustyrussell: No, I'm happy for the change in sematics: now we have per-channel control, the global simply becomes a default-at-creation.  It's unusual for people to modify them at the moment anyway.

m-schmoock: OK, thanks for the clarification

m-schmoock:
@cdecker @rustyrussell 
Thanks, Rusty clearified that we will freeze the current global config fee/ppm values as default for each existing channel. **Thus a change of global values does not have to affect existing channels but only new channels to be opened in the future**. I will also update the doc `doc/lightningd-config.5.txt` to point that out and come back with another update without hacky commit https://github.com/ElementsProject/lightning/commit/40af7d1910d892d2b7d40349fd52b8efa463c00c . It would help if #2394 was meregd.

m-schmoock:
@cdecker @rustyrussell I updated the branch.

- Manpage on config updated to point out that fee settings wil only apply for new channels not existing
- Until #2394 is merged, I will set the channel base/ppm values to UNT32_MAX
- When channels are restored from DB, apply global config values

Question: do we want a `channel id` wildcard '*' feature in the JSON command to update all channels at once? This will help when the user wanted to change the config values, but this is not possible anymore because fee values frozen in DB when channel was created.

m-schmoock:
Thanks for merging #2394, will do a proper DB update routine now...

m-schmoock:
@rustyrussell I removed the DB `-1` hack and added the new database context migration method. Works nice, except for the hooks being called with `*ld := NULL` for some wallet/db unit tests, which forces us to have an early return in all migration hooks.

I also stabilized the tests and finished TODOs in a straight forward way, where I was missing input. For details see top description of this thread. Feel free to audit the PR...

> [*] Dance a happy dance.

common/amount.h
```diff
@@ -98,6 +98,13 @@ bool amount_msat_less_sat(struct amount_msat msat, struct amount_sat sat);
 /* Is msat <= sat? */
 bool amount_msat_less_eq_sat(struct amount_msat msat, struct amount_sat sat);
 
+/* Does value overflow u32 bit */
+bool amount_msat_overflow(struct amount_msat msat);
```
cdecker: This should probably have a `u32` in its name.

lightningd/peer_control.c
```diff
@@ -1339,6 +1339,102 @@ static const struct json_command getinfo_command = {
...
+	derive_channel_id(&cid, &channel->funding_txid, channel->funding_outnum);
+
+	/* write response */
+	response = json_stream_success(cmd);
+	json_object_start(response, NULL);
+	json_add_num(response, "base", amount_msat_to_u32(*base));
+	json_add_num(response, "ppm", *ppm);
+	json_add_string(response, "channel_id",
+			type_to_string(tmpctx, struct channel_id, &cid));
+	if (peer)
```
cdecker: Can we actually ever get here without having a `peer`?

lightningd/peer_control.c
```diff
@@ -1339,6 +1339,102 @@ static const struct json_command getinfo_command = {
...
+		json_add_pubkey(response, "id", &peer->id);
+	if (channel->scid)
```
cdecker: Same here, if we have `channel->state == CHANNELD_NORMAL` (checked above) we also need to have a `channel->scid`.

cdecker:
One minor `valgrind` issue CI found:

```
==11779== Use of uninitialised value of size 8
==11779==    at 0x567776B: _itoa_word (_itoa.c:179)
==11779==    by 0x567B12C: vfprintf (vfprintf.c:1631)
==11779==    by 0x56A3A48: vsnprintf (vsnprintf.c:114)
==11779==    by 0x425D1A: do_vfmt (str.c:66)
==11779==    by 0x425E00: tal_vfmt_ (str.c:92)
==11779==    by 0x42AC11: db_exec (db.c:492)
==11779==    by 0x42C08D: migrate_pr2342_feerate_per_channel (db.c:991)
==11779==    by 0x42B19A: db_migrate (db.c:640)
==11779==    by 0x42C311: test_empty_db_migrate (run-db.c:61)
==11779==    by 0x42C7D4: main (run-db.c:126)
==11779==  Uninitialised value was created by a heap allocation
==11779==    at 0x4C2DB8F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==11779==    by 0x426F01: allocate (tal.c:245)
==11779==    by 0x42748E: tal_alloc_ (tal.c:423)
==11779==    by 0x42C7C4: main (run-db.c:124)
```

m-schmoock:
@cdecker thanks, will correct the minor stuff. About wildcard `setchannelfee "all" <base> <ppm>`, im working on it, do we want to have it in this PR or as seperate so we can close this sooner?

cdecker:
> @cdecker thanks, will correct the minor stuff. About wildcard setchannelfee "all" <base> <ppm>, im working on it, do we want to have it in this PR or as seperate so we can close this sooner?

Let's have that support in a new PR :-)

common/amount.h
m-schmoock: addressed

lightningd/peer_control.c
m-schmoock: addressed

lightningd/peer_control.c
m-schmoock: addressed

m-schmoock:
I cherrypicked 

fix: a pr2444 introduced valgrind complaint b546a85
fix: also set testtub config for wallet test 255daeb

To get around travis CI valgrind check. Probably not final. But rest is final.

m-schmoock:
Looks like valgrind and all the tests are happy now, are you as well?
I can update the PR without the cherry picks (if merged via #2449) ...

lightningd/peer_control.c
```diff
@@ -1339,6 +1343,101 @@ static const struct json_command getinfo_command = {
...
+			return res;
+		/* check channel is found and in valid state */
+		if (!channel)
+			return command_fail(cmd, LIGHTNINGD,
+					"Could not find channel with that id");
+	}
+
+	if (channel->state != CHANNELD_NORMAL)
+		return command_fail(cmd, LIGHTNINGD,
+				"Channel is not in state CHANNELD_NORMAL");
```
niftynei: as a nicety, i'd include the channel id and current state in the fail message.

lightningd/peer_control.c
```diff
@@ -1339,6 +1343,101 @@ static const struct json_command getinfo_command = {
...
+
+	/* set, notify and save new values */
+	channel->feerate_base = amount_msat_to_u32(*base);
+	channel->feerate_ppm = *ppm;
+
+	/* tell channeld to make a send_channel_update */
+	if (channel->owner && streq(channel->owner->name, "lightning_channeld"))
```
niftynei: out of curiosity, in what cases will the channel->owner not be "lightning_channeld"? i.e. is there a mitigation or message value we should return if this check fails?

tests/test_pay.py
```diff
@@ -1425,3 +1426,276 @@ def test_pay_direct(node_factory, bitcoind):
...
+        'SELECT feerate_base, feerate_ppm FROM channels '
+        'WHERE hex(short_channel_id)="' + scid_hex + '";')
+    assert(db_fees[0]['feerate_base'] == 0)
+    assert(db_fees[0]['feerate_ppm'] == 0)
+
+    # disable and check for global values to be returned
+    result = l1.rpc.setchannelfee(scid)
+    assert(result['base'] == DEF_BASE)
+    assert(result['ppm'] == DEF_PPM)
+    # check zero values in DB
```
niftynei:
```suggestion
    # check default values in DB
```

tests/test_pay.py
```diff
@@ -1425,3 +1426,276 @@ def test_pay_direct(node_factory, bitcoind):
...
+    assert(db_fees[0]['feerate_base'] == 42)
+    assert(db_fees[0]['feerate_ppm'] == 43)
+
+    # check if invalid scid raises proper error
+    with pytest.raises(RpcError, match=r'-1.*Could not find active channel of peer with that id'):
+        result = l1.rpc.setchannelfee(l3.info['id'], 42, 43)
+    with pytest.raises(RpcError, match=r'-32602.*Given id is not a channel ID or short channel ID'):
+        result = l1.rpc.setchannelfee('f42' + scid[3:], 42, 43)
+
+    # check if 'base' unit can modified to satoshi
```
niftynei:
```suggestion
    # check if 'base' unit can be modified to satoshi
```

tests/test_pay.py
```diff
@@ -1425,3 +1426,276 @@ def test_pay_direct(node_factory, bitcoind):
         # have changed.
         l1l2msat = only_one(l1.rpc.getpeer(l2.info['id'])['channels'])['msatoshi_to_us']
         assert l1l2msat == l1l2msatreference
+
```
niftynei: tests for channels in non-`CHANNEL_NORMAL` state?

m-schmoock:
@niftynei thanks for your input, I will push an update this weekend

lightningd/peer_control.c
m-schmoock: @niftynei  This is a security measure because I don't know the insights of clightning in detail enough. It may be the case that if state := `CHANNELD_NORMAL` it is always owned by channeld. Now that you ask, this is maybe not necessary in this scenario, or we may want a check that enforces it just like the `CHANNELD_NORMAL` a few lines earlier.

I had a ton of open detail questions that were never addressed, one of them being the question if we should allow setting feerates no channel in states other than `CHANNELD_NORMAL`.  In this case a statement like this makes sense, because we only wire the new fees if the channel is online.

So maybe my question to you: In which channel states are we supposed to allow the feerate settings? Because the DB update can maybe be made earlier and the channel will apply correct fees when it is in state `CHANNELD_NORMAL` eventually.

tests/test_pay.py
m-schmoock: addressed by 06883dd

tests/test_pay.py
m-schmoock: addressed

tests/test_pay.py
m-schmoock: addressed

lightningd/peer_control.c
m-schmoock: addressed by 
773bc94

tests/test_pay.py
m-schmoock: (test_pay `test_setchannelfee_state`)

lightningd/peer_control.c
niftynei: well. let's see. the possible channel states are (from `lightningd/channel_state.h):

```lightningd/channel_state.h
       /* In channeld, still waiting for lockin. */
        CHANNELD_AWAITING_LOCKIN = 2,

        /* Normal operating state. */
        CHANNELD_NORMAL,

        /* We are closing, pending HTLC resolution. */
        CHANNELD_SHUTTING_DOWN,

        /* Exchanging signatures on closing tx. */
        CLOSINGD_SIGEXCHANGE,

        /* Waiting for onchain event. */
        CLOSINGD_COMPLETE,

        /* Waiting for unilateral close to hit blockchain. */
        AWAITING_UNILATERAL,

        /* We've seen the funding spent, we're waiting for onchaind. */
        FUNDING_SPEND_SEEN,

        /* On chain */
        ONCHAIN
```

of these, `CHANNELD_NORMAL` and `CHANNELD_AWAITING_LOCKIN` appear to be reasonable candidates for permitting updates to the configured channel. in theory* there's no reason that you _couldn't_ update the channel fees in any of the other states, except that they would have zero effect since the channel is no longer accepting new htlc's (ie those fee updates will never be used).

re: the restriction to only channels that are owned by `channeld`, imo this is coding too tightly to the implementation. the simple heuristic of using well defined channel states to restrict the window of updates is much more flexible, imo. in other words, if we've checked the channel is in the correct state, we should assume that the update is valid.

*ignoring some implementation reality that i'm unaware of that makes this infeasible

m-schmoock: Thanks for some input ;)
Will include waiting channels without wiring the fees to an otherwise non existent channeld. Write test to see if it works as expected and correct fees will be applied when the channel is in state NORMAL eventually.

lightningd/peer_control.c
```diff
@@ -1034,7 +1034,7 @@ AUTODATA(json_command, &listpeers_command);
 static struct command_result *
 command_find_channel(struct command *cmd,
 		     const char *buffer, const jsmntok_t *tok,
-		     struct channel **channel)
+		     struct channel **channel, struct peer **peer_out)
```
cdecker: Doesn't `channel` have a pointer to the associated `peer`?

m-schmoock: yep, addressed by https://github.com/ElementsProject/lightning/pull/2342/commits/a9e84acf67b75923450c94dd738922f0b1df18a6

common/amount.c
```diff
@@ -362,6 +362,21 @@ bool amount_msat_less_eq_sat(struct amount_msat msat, struct amount_sat sat)
...
+
+u32 amount_msat_to_u32(struct amount_msat msat)
+{
+	return (u32)msat.millisatoshis;
+}
+
+void amount_msat_from_u32(struct amount_msat *msat, u32 value)
+{
+	msat->millisatoshis = (u64)value;
+}
```
rustyrussell: We don't create these primitives for very good reasons; there *are* times to use direct access, but they should be carefully checked, hence the requirement to annotate them with an excuse.  Otherwise people can start making mistakes with amounts (eg. doing arith on them!).

But I'm going to propose this change as a separate fixup, since I don't want to delay approving this excellent work further!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/c8b8f809773d34a3bf1b1575b4f56167f0a8f210

Ackbot, are you there? :)

common/amount.c
m-schmoock: Thanks. Yes, I was also not happy with that direct access.

PR #2345, opened 2019-02-11
WIP: wireshark dissector logformat revamp

niftynei:
Improvements to the beta version of the wireshark plugin integration. 

Wireshark plugin needs to be revamped to be able to take advantage of this new key output format, which is a work in progress. I'll update this PR once that is done.


common/cryptomsg.c
```diff
@@ -16,22 +21,88 @@
...
-#endif
+
+void dissector_add_keys(const char *host, const char *sk, const char *rk)
+{
+	char host_out[256], host_in[256];
+	snprintf(host_out, sizeof host_out, "->%s", host);
+	snprintf(host_in, sizeof host_in, "<-%s", host);
+	strmap_add(&keymap, strdup(sk), strdup(host_in));
+	strmap_add(&keymap, strdup(rk), strdup(host_out));
+	update_file();
```
niftynei: @rustyrussell this works as intended.. if i connect to another peer, the logfile is updated as i'd expect

common/cryptomsg.c
```diff
@@ -16,22 +21,88 @@
...
+/*
+ * When any key is updated, we update the map and then reprint the
+ * file that wireshark reads to update keys.
+ */
+void dissector_update_key(const char *old_key, const char *new_key)
+{
+	const char *host;
+	const char *key = strdup(new_key);
+	/* Delete the old key's entry + retrieve host */
+  	if (!strmap_del(&keymap, old_key, &host)) {
```
niftynei: @rustyrussell this does not work as expected. the `&keymap` appears to be empty? this fails with a message that the old key cannot be found in the map.

niftynei: if i print the `keymap` to file at this point, the file will be empty.

common/dissector.c
```diff
@@ -0,0 +1,69 @@
...
+	if (fp == NULL)
+		return;
+
+	char *keys = tal_fmt(NULL, "sk:%s\ns_ck:%s\nrk:%s\nr_ck:%s\n",
+		tal_hexstr(NULL, &cs->sk, sizeof(cs->sk)),
+		tal_hexstr(NULL, &cs->s_ck, sizeof(cs->s_ck)),
+		tal_hexstr(NULL, &cs->rk, sizeof(cs->rk)),
+		tal_hexstr(NULL, &cs->r_ck, sizeof(cs->r_ck)));
+
+	if (fputs(keys, fp) == -1) {
```
niftynei: @rustyrussell this is failing a `check-discouraged-functions` -- what would you recommend we replace it with?

common/dissector.c
```diff
@@ -0,0 +1,65 @@
...
+		status_debug("Error printing keys to %s , exiting early", filename);
+		return;
+	}
+
+	fclose(fp);
+}
+
+void dissector_remove_connection(const char *our_addr, const char *peer_addr)
```
niftynei: currently unused...

PR #2358, opened 2019-02-16
fix crash with lightning charge and plugin opts

rsbondi:
**Problem:** 
when running the daemon that loads a plugin with options, then starting lightning-charge the daemon crashes

**Resolution**
in `options.c`, `add_config` is called with a `opt->cb_arg` of type `plugin_opt_set`, I added a check for this

I am not sure why this is called in this manner so this may not be the desired solution, but I thought it best to show my findings in code rather than open an issue.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2358/commits/61eaa9e8a2d6a2bece561efdb994a55969accc51

PR #2359, opened 2019-02-17
pytest: fix race in test_pay_direct.

rustyrussell:
I got a spurious failure because the final node gave a CLTV error and
so it decided to use a different channel.  It should probably handle
this corner case better, but meanwhile make the test robust.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2359/commits/e58123b6748fe904cc3ab1d9f2163be608b21a9c

PR #2360, opened 2019-02-17
Amounts Part II: Amount API change

rustyrussell:
This changes all the JSON APIs to reduce satoshi/millisatoshi confusion.  It's built on the groundwork from Part I in #2353. Part III converts all the internal fields to 'struct amount_sat' and 'amount_msat;.

1. Amount parameters now accept raw integers as before, or msat/sat/btc-suffixed strings of restricted forms, see documentation.
2. Every output amount field has new _msat ~~or _sat~~ equivalent which are strings with the value with 'msat' ~~or 'sat'~~ appended.
3. The python library turns these into ~~Satoshi and~~ Millisatoshi classes, which you can also use to hand to all those amount parameters.

After much flip-flopping:
1. I didn't deprecate any old fields.  That could come later.
2. I didn't deprecate non-suffixed integers.  That could also come later.
3. I left _msat ~~and _sat~~ redundantly appended to field names so that users don't have to write generic parsers: they know 'amount_msat' will always have value "NNNmsat".


lightningd/invoice.c
```diff
@@ -806,8 +802,9 @@ static struct command_result *json_decodepay(struct command *cmd,
 	json_add_u64(response, "created_at", b11->timestamp);
 	json_add_u64(response, "expiry", b11->expiry);
 	json_add_pubkey(response, "payee", &b11->receiver_id);
-        if (b11->msatoshi)
-                json_add_u64(response, "msatoshi", *b11->msatoshi);
+        if (b11->msat)
+                json_add_amount_msat(response, *b11->msat,
+				     "msatoshi", "amount_msat");
```
niftynei: petition to rename "amount_msat" as "formatted_msat" or "formatted_value" or something to denote the 'formatted' i.e. string nature of the field value -- amount imo connotes an int or float type field.

common/wallet_tx.c
```diff
@@ -4,26 +4,53 @@
...
-		return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+	if (tal_count(wtx->utxos) == 0) {
+		return command_fail(wtx->cmd, FUND_CANNOT_AFFORD,
 				    "Cannot afford transaction");
 	}
-	if (amount < 546) {
-		return command_fail(tx->cmd, FUND_OUTPUT_IS_DUST,
-				    "Output %"PRIu64" satoshis would be dust",
-				    amount);
+	if (amount.satoshis < get_chainparams(wtx->cmd->ld)->dust_limit) {
```
niftynei: :+1:

common/wallet_tx.c
```diff
@@ -34,18 +61,21 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+					      &amount.satoshis,
 					      &fee_estimate);
 		res = check_amount(tx, amount);
 		if (res)
 			return res;
 
-		if (amount <= tx->amount) {
-			tx->change = 0;
+		/* tx->amount is max permissible */
+		if (amount_sat_less(amount, tx->amount)) {
```
niftynei: i think you want `less_eq` here?

lightningd/json.c
```diff
@@ -32,7 +32,7 @@ json_add_route_hop(struct json_stream *r, char const *n,
 	json_add_short_channel_id(r, "channel",
 				  &h->channel_id);
 	json_add_num(r, "direction", h->direction);
-	json_add_u64(r, "msatoshi", h->amount);
+	json_add_amount_msat(r, h->amount, "msatoshi", "amount_msat");
```
niftynei: petition to rename "amount_msat" ... same as other comment.

plugins/pay.c
```diff
@@ -634,10 +634,11 @@ static struct command_result *add_shadow_route(struct command *cmd,
...
 
 	json_for_each_arr(i, chan, channels) {
-		u64 sats, v;
+		struct amount_sat sats;
+		u64 v;
 
-		json_to_u64(buf, json_get_member(buf, chan, "satoshis"), &sats);
-		if (sats * 1000 < pc->msatoshi)
+		json_to_sat(buf, json_get_member(buf, chan, "satoshis"), &sats);
+		if (sats.satoshis * 1000 < pc->msatoshi)
```
niftynei: should we use a conversion or comparison functions rather than multiplying by 1000?

lightningd/invoice.c
rustyrussell: The problem is that we may eventually deprecate the old raw-numeric fields.  So it does need some semantic :(

common/wallet_tx.c
rustyrussell: Good catch!  Effect should be the same, but it was *not* a deliberate change!

plugins/pay.c
rustyrussell: Yes; see part III.  This is the minimum shims to get the JSON API converted.

rustyrussell:
Rebase to fix conflicts with master (fairly minor), and correct error @niftynei found; didn't change amount_msat but better names welcome...

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2360/commits/9589a6f2bf935b616e8c3af14099fde040e601f0

PR #2363, opened 2019-02-18
Multi-frame onion format proposal

cdecker:
Quoting the description from https://github.com/lightningnetwork/lightning-onion/pull/31:

 - Use the 4 MSB in the first realm byte as a counter of how many 65 byte payload-chunks are destined for the processing node. This clearly distinguishes them from the current payload and directly gives an indication on how much it has to read/allocate.
 - Separate payload parsing from the payload extraction, i.e., the onion processing code just hands back a slice of bytes and a realm that are then passed to external parsers that don't interact with the onion processing at all. Mirroring this, when creating a new onion we just get passed the 4 LSB of the realm, and an opaque slices of bytes that may or may not be TLV encoded, we don't care.
 - Given the payload length we can infer how many hops we need to read/write. The format looks like this:

For a payload of length <= 32:
```
|-------+-------------------+------|
| realm | payload + padding | HMAC |
|-------+-------------------+------|
```

For payloads > 32 and <= 162 for example:
```
|-------+--------------+-----------------+-------------------------+------|
| realm | payload[:64] | payload[64:129] | payload[129:] + padding | HMAC |
|-------+--------------+-----------------+-------------------------+------|
```

In other words we use the realm byte from the first hop to determine the number of hops to read, the first hop still has room for 64 bytes of payload (per-hop-payload + HMAC size - realm byte). Any intermediate hop has the full 65 bytes available. The last hop has 33 bytes of payload available, since we will use its HMAC to pass it on to the next node in the path. Notice that we do not decrypt the payloads after the first since processing the first hop already decrypted all the following hops, including the ones we'll be processing. In addition we get a better use of the limited space that we have available and the entire payload is contiguous in memory and can be passed out to the parser as is, without having to stitch it together.

The implementation is also rather trivial, all we need to do is to pass the payload as a byte slice out during processing, and to get the next onion instead of shifting by 65 bytes and padding with 0s, we shift by `65*nhops` and pad that with 0s. So the only thing we really need to do is to have the [`rightShift`](https://github.com/lightningnetwork/lightning-onion/blob/725fcf830eac43d111b85c4d529d6f555eebb83d/sphinx.go#L362), [`headerWithPadding`](https://github.com/lightningnetwork/lightning-onion/blob/725fcf830eac43d111b85c4d529d6f555eebb83d/sphinx.go#L748-L749), and [`copy`](https://github.com/lightningnetwork/lightning-onion/blob/725fcf830eac43d111b85c4d529d6f555eebb83d/sphinx.go#L770) not take 65, but 65*nhops as arguments.

Closes #2689


cdecker:
> Unit tests don't need this exposed though?

Depends on whether `devtools/onion` is considered a unit test (and should therefore `#include <sphinx.c>`) or it is a user-facing tool (that should use the exposed API in `sphinx.h`). I'm on the fence for this one, since we need the ability to set the `session_key` for reproduceability.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2363/commits/af033d72ecbf9dc6367d77848316ba39f4c7afe2

Want to apply this rework for an independent PR, so I'm going to take those cleanups now :)

common/sphinx.c
```diff
@@ -89,10 +89,14 @@ struct sphinx_path *sphinx_path_new_with_key(const tal_t *ctx,
 
 static size_t sphinx_hop_count_frames(const struct sphinx_hop *hop)
 {
+#if EXPERIMENTAL_FEATURES
```
cdecker: This is likely redundant with the check on line 426, since this just clamps the number of frames to 1, never triggering the failure, and probably corrupting the frame if we end up using more anyway. I suggest we just keep the check below, and restore the normal behavior.

cdecker:
During my rebase I accidentally removed @rustyrussell's experimental guards. Rebased and cherry-picked it back onto the PR :-)

cdecker:
Updated the PR to reflect the change in realm byte split. It now uses a 5bit + 3bit split like the recent change to lightningnetwork/lightning-rfc#593. Once that is merged we can go ahead as well.

common/sphinx.c
```diff
@@ -166,7 +165,7 @@ static bool generate_header_padding(
 	)
 {
 	int i;
-	u8 cipher_stream[(NUM_MAX_HOPS + 1) * hopsize];
+	u8 cipher_stream[(NUM_MAX_FRAMES + 1) * FRAME_SIZE];
```
rustyrussell: Either this change is wrong, or hopsize should not be a parameter to this function?

common/sphinx.h
```diff
@@ -12,17 +12,19 @@
...
-#define ROUTING_INFO_SIZE (HOP_DATA_SIZE * NUM_MAX_HOPS)
-#define TOTAL_PACKET_SIZE (1 + 33 + SECURITY_PARAMETER + ROUTING_INFO_SIZE)
+#define VERSION_SIZE 1
+#define REALM_SIZE 1
+#define HMAC_SIZE 32
+#define PUBKEY_SIZE 33
+#define FRAME_SIZE 65
+#define NUM_MAX_FRAMES 20
+#define ROUTING_INFO_SIZE (FRAME_SIZE * NUM_MAX_FRAMES)
+#define TOTAL_PACKET_SIZE (VERSION_SIZE + PUBKEY_SIZE + HMAC_SIZE + ROUTING_INFO_SIZE)
```
rustyrussell: A properly marked `/* BOLT #4 ` quote here would be illiminating and useful :)

common/sphinx.c
```diff
@@ -34,6 +34,57 @@ struct keyset {
...
+{
+	struct sphinx_path *sp = tal(ctx, struct sphinx_path);
+	sp->associated_data = tal_dup_arr(sp, u8, associated_data,
+					  tal_bytelen(associated_data), 0);
+	sp->session_key = NULL;
+	sp->hops = tal_arr(sp, struct sphinx_hop, 0);
+	return sp;
+}
+
+struct sphinx_path *sphinx_path_new_with_key(const tal_t *ctx,
```
rustyrussell: Maybe wrap this declaratio (and defn) in `#if DEVELOPER` to emphasize it?  Means devtools/onion is now DEVELOPER-only, but that's probably a Good Thing?

common/sphinx.c
```diff
@@ -378,25 +405,35 @@ static void deserialize_hop_data(struct hop_data *data, const u8 *src)
 	fromwire(&cursor, &max, &data->hmac, HMAC_SIZE);
 }
 
+static void sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
+{
+		memset(dest, 0, FRAME_SIZE);
+		dest[0] = hop->realm;
+		memcpy(dest + 1, hop->payload, tal_bytelen(hop->payload));
+		memcpy(dest + FRAME_SIZE - HMAC_SIZE, hop->hmac, HMAC_SIZE);
+}
```
rustyrussell: Weird indent?

lightningd/pay.c
```diff
@@ -596,37 +596,36 @@ send_payment(struct lightningd *ld,
...
 		ids[i] = route[i].nodeid;
-		/* JSON parsing checked these were valid, so Shouldn't Happen */
-		if (!pubkey_from_node_id(&path[i], &ids[i])) {
-			return command_fail(cmd, PAY_UNSPECIFIED_ERROR,
-					    "Invalid nodeid %s",
-					    type_to_string(tmpctx,
-							   struct node_id,
-							   &ids[i]));
-		}
-	}
```
rustyrussell: Removing this sanity check makes me uncomfortable.  But why don't we simply make `ids` a struct pubkey array directly, and do the `pubkey_from_node_id(&ids[i], &route[i].nodeid))` here in the loop and then fail immediately like this did?

lightningd/pay.c
```diff
@@ -596,37 +596,36 @@ send_payment(struct lightningd *ld,
...
 
 	/* Copy hop_data[n] from route[n+1] (ie. where it goes next) */
 	for (i = 0; i < n_hops - 1; i++) {
+		ret = pubkey_from_node_id(&pubkey, &ids[i]);
+		assert(ret);
```
rustyrussell: ... That would remove this code, and the one below.

lightningd/pay.c
```diff
@@ -627,14 +628,10 @@ send_payment(struct lightningd *ld,
 				  route[i + 1].amount,
 				  base_expiry + route[i + 1].delay);
 	}
+	for (i = 0; i < n_hops - 1; i++)
```
rustyrussell: Huh? :confused:

common/sphinx.c
```diff
@@ -248,27 +250,40 @@ static bool generate_key(void *k, const char *t, u8 tlen, const u8 *s)
...
-	int i;
-	u8 cipher_stream[(NUM_MAX_FRAMES + 1) * FRAME_SIZE];
+	u8 stream[2 * NUM_MAX_FRAMES * FRAME_SIZE];
 	u8 key[KEY_LEN];
+	size_t fillerStart, fillerEnd, fillerFrames;
 
 	memset(dst, 0, dstlen);
-	for (i = 1; i < numhops; i++) {
-		if (!generate_key(&key, keytype, keytypelen, params[i - 1].secret))
+	for (int i = 0; i < tal_count(path->hops) - 1; i++) {
```
rustyrussell: `size_t i` preferred here BTW.

common/sphinx.h
```diff
@@ -230,6 +230,12 @@ struct sphinx_path *sphinx_path_new_with_key(const tal_t *ctx,
  * Add a V0 (Realm 0) single frame hop to the path.
  */
 void sphinx_add_v0_hop(struct sphinx_path *path, const struct pubkey *pubkey,
-		       const struct short_channel_id *scid,
-		       struct amount_msat forward, u32 outgoing_cltv);
+		       const struct short_channel_id *scid, struct amount_msat forward,
+		       u32 outgoing_cltv);
+/**
+ * Add a raw payload hop to the path.
```
rustyrussell: Maybe note it's expose for testing...

common/sphinx.c
rustyrussell: Weird indent fixed :)

common/sphinx.c
```diff
@@ -419,10 +419,12 @@ static void deserialize_hop_data(struct hop_data *data, const u8 *src)
 
 static void sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
 {
-		memset(dest, 0, FRAME_SIZE);
-		dest[0] = hop->realm;
-		memcpy(dest + 1, hop->payload, tal_bytelen(hop->payload));
-		memcpy(dest + FRAME_SIZE - HMAC_SIZE, hop->hmac, HMAC_SIZE);
+	u8 num_frames = sphinx_hop_count_frames(hop);
+	memset(dest, 0, num_frames * FRAME_SIZE);
+	dest[0] = hop->realm | (num_frames-1) << 4;
```
rustyrussell: `|` binds less than `<<` but I still want more brackets here please :)  Also, this is now 3 is it not?

Maybe a heap of assertions here, like `assert(hop->realm < (1 << 4)` and `assert(tal_bytelen(hop->payload) < sphinx_hop_count_frames(hop) * FRAME_SIZE) - HMAC_SIZE - 1)`?

common/sphinx.h
```diff
@@ -83,14 +83,24 @@ struct hop_data {
 	struct short_channel_id channel_id;
 	struct amount_msat amt_forward;
 	u32 outgoing_cltv;
-	/* Padding omitted, will be zeroed */
-	u8 hmac[HMAC_SIZE];
+};
+
+enum sphinx_payload_type {
+	SPHINX_V0_PAYLOAD = 0,
+	SPHINX_RAW_PAYLOAD = 255,
```
rustyrussell: Whenever we overload values for internal usage, I always worry about leakage (someone sends us a realm 255 onion in future).

Let's make `route_step.realm` and int, and use negative values for internal ones maybe?  Or 256?

common/test/onion-test-v0.json
```diff
@@ -0,0 +1,42 @@
+{
```
rustyrussell: Would prefer to see this tested in a unit test rather than a devtool, though.

common/sphinx.c
rustyrussell: OK.

cdecker:
Re-added the Work in Progress tag since we need an upstream decision about which proposal should be merged.

rustyrussell:
Looks like this should be closed in favor of the other PR; removed milestone tag anyway.

PR #2368, opened 2019-02-19
[pylightning] Corrected the size to read from the buffer

darosior:
To stop reading when `}\n` caused to fail to read output of call like `listfunds` where we stop to read 1024 bytes after the end of the "outputs" (since it is closed with `}\n`) instead of reading the whole response. It would later cause a JSON error (at `raw_decode(buff)`).

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -42,7 +42,7 @@ class UnixDomainSocketRpc(object):
 
                 if len(b) == 0:
                     return {'error': 'Connection to RPC server lost.'}
-                if b' }\n' not in buff:
+                if b' }\n' not in buff[-4:]:
                     continue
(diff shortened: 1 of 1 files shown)


cdecker:
Sorry, we can't currently rely on `\n\n` being there since we need to support prior versions that didn't adhere to this convention. The erroneous decode attempts are the price we pay for that backwards compatibility. We'll eventually end up doing this, just not now :wink:

darosior:
I changed the commit to search only in the last 4 "chars" of the buffer, that way it would pass the test with ` }\n` or ` }\n\n` and not fail it there is ` }\n` (like in `listfunds`) in the middle of the RPC answer.

cdecker:
That's not entirely correct, since we aren't guaranteed that read returns between two messages. So if two messages are sent back to back we might not detect the end of the first and wait for the second to finish. Every time this happens we keep one more message in the read buffer and will not return a message that was queued for delivery.

Issue #2219, opened 2019-01-03
BLW sometimes sends funding_locked before re-establish


rustyrussell:
### Issue and Steps to Reproduce
{
      "type": "UNUSUAL", 
      "time": "582.646173141", 
      "source": "lightningd(18552):", 
      "log": "033627ce012a83bb2fa87754ab2f35963d353b347d2b39a601599eb8d1f8c28576 chan #7: Peer permanent failure in CHANNELD_AWAITING_LOCKIN: lightning_channeld: sent ERROR bad reestablish msg: WIRE_FUNDING_LOCKED 0024a05b28d89ab50c4a7facd04d6c64ca7bde2835f57590eda6bc782063071d0b2102cabca945a66e3ae275d590e81bfbbc555ab157f907f5465522feb519a7738740 (tx 02e780ffa339db8b03994f13cc7871c1a6351ce7d2d202c952c4cac885d5ec1b)"
    }

btcontract:
Could it possibly be anything else? Can't reproduce this, also BLW sends `ChannelReestablish` in response to `Init` and then `FundingLocked` in response to peer's `ChannelReestablish` within a single thread so it's always in that order.

rustyrussell:
Someone else reported it with LND, so perhaps that's it.  Either way, will workaround for 0.7.

PR #2373, opened 2019-02-20
channeld: handle funding_locked message before reestablish.

rustyrussell:
LND seems to do this occasionally, though fixed in new versions.  Workaround
in the meantime.

I tested this by hacking our code to send it prematurely, and this worked.

Fixes: #2219


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2373/commits/96aae09e1d8291375d21b834a80713caa6b7fced

TheBlueMatt:
Note that this is *still* not "fixed in new versions" :sob:

PR #2374, opened 2019-02-20
Pylightning: Raise error if required argument is missing

conscott:
At the moment if you have the rpc method
```python
@plugin.method("hello")
def hello(plugin, name, punctuation="!"):
```
And call it from the cli, with
```
lightning-cli hello
```

Currently, the arg for `name` will default to `inspect._empty`, which is a bit clumsy. 

This change instead returns the error message:
```
"Error while processing hello: TypeError('Missing required parameter: name',)"
```

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -194,6 +194,9 @@ class Plugin(object):
                 if pos < len(params):
                     # Apply positional args if we have them
                     arguments[k] = params[pos]
+                elif sig.parameters[k].default is inspect.Signature.empty:
+                    # This is a positional arg with no value passed
+                    raise TypeError("Missing required parameter: %s" % sig.parameters[k])
                 else:
                     # For the remainder apply default args
                     arguments[k] = sig.parameters[k].default


cdecker:
It really should return an exception that an argument is missing imho

conscott:
Updated to error msg.

contrib/pylightning/lightning/plugin.py
```diff
@@ -194,6 +194,9 @@ def _exec_func(self, func, request):
                 if pos < len(params):
                     # Apply positional args if we have them
                     arguments[k] = params[pos]
+                elif sig.parameters[k].default is inspect.Signature.empty:
+                    # This is a positional arg with no value passed
+                    return "Error - Missing required parameter: %s" % sig.parameters[k]
```
cdecker:
```suggestion
                    raise TypeError("Error - Missing required parameter: %s" % sig.parameters[k])
```

cdecker: It should really raise an exception since otherwise we might end up using the `"Error - Missing required parameter: %s" % sig.parameters[k]` as a valid result from the call instead of detecting it as a failure.

conscott:
Sorry, should have seen the exception handler. Updated!

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2374/commits/00b9997b90c6a9787f6fae546ef3c4367b1d3ad6

PR #2319, opened 2019-02-05
Use external libraries and neaten release

rustyrussell:
I am working on a simple reproducible build script, at least for ubuntu 18.04 LTS.  These cleanups make things simpler.

PR #1462, opened 2018-05-05
doc: Remove libsodium-dev dependency, it's in-tree

cdecker:
Fixes #1451 

Signed-off-by: Christian Decker <decker.christian@gmail.com>
Reported-by: Perlover <@Perlover>

PR #2375, opened 2019-02-21
Switch to libsodium-dev for ubuntu install instructions

conscott:
`libsodium-dev` was removed in #1462 then added backed in #2319 - BUT as `libsodium`, which is not found. This updates to `libsodium-dev`


rustyrussell:
Ack f82605e1eac6fed1d3891e8368c4c0b906599178

PR #2380, opened 2019-02-21
Reduce the risk of change chaining transactions

cdecker:
We had quite a few cases in which one failed transaction (potentially due to insufficient fees) was causing all following transactions to remain in limbo as well.

This PR adds a `minconf` argument inspired by `bitcoind` that allows us to determine how many confirmations an output must have before being considered for coin selection. The default of 1 confirmation mitigates the risk of chainge chaining, with the drawback of failing a few more transactions at selection than strictly necessary.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -254,6 +254,7 @@ static const struct utxo **wallet_select(const tal_t *ctx, struct wallet *w,
 					 const u32 feerate_per_kw,
 					 size_t outscriptlen,
 					 bool may_have_change,
+					 u32 maxheight,
 					 struct amount_sat *satoshi_in,
 					 struct amount_sat *fee_estimate)
 {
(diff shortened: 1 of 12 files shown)


wallet/wallet.c
```diff
@@ -283,6 +284,13 @@ static const struct utxo **wallet_select(const tal_t *ctx, struct wallet *w,
 		struct amount_sat needed;
 		struct utxo *u = tal_steal(utxos, available[i]);
 
+		/* If we require confirmations check that we have a
+		 * confirmation height and that it is below the required
+		 * maxheight (current_height - minconf */
```
niftynei: ubernit: missing `)`

wallet/wallet.c
```diff
@@ -283,6 +284,13 @@ static const struct utxo **wallet_select(const tal_t *ctx, struct wallet *w,
...
+		if (maxheight != 0 &&
+		    (!u->blockheight || *u->blockheight > maxheight))
```
niftynei: tbh i find `maxheight` not intuitive wrt to what you're checking for here. maybe something like `gated_confirmation_height` or `req_conf_height` ... neither of these seems exactly right.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2380/commits/0defb3383fda8b7935607fe333608ecaf34b062a

wallet/wallet.c
cdecker: Yeah, I've been going back and forth with the name, it's the maximum acceptable `confirmation_height` really, and being maybe the special value `0` is what makes it confusing.

PR #2381, opened 2019-02-21
feat: block of newlines when attaching a logfile

m-schmoock:
This will make the logger write 4 newlines to re-attached logfiles.
The newlines wont appear on logfiles that are just created.
Additionally the server prints 50 '-' dashes before printing his
startup message, which also help increase readability on logfile.

This was inspired by the way Bitcoin Core handles logfiles.


rustyrussell:
Thanks; little usability touches like this make everyone's lives better!

PR #2382, opened 2019-02-22
listpays

rustyrussell:
Also cleans up our badly-misunderstood 'description' parameter for sendpay/pay to do what people (including me!) thought it did.

@cdecker Q on what you want to do with compat in pylightning: I didn't, but we can.


contrib/pylightning/lightning/lightning.py
```diff
@@ -464,18 +464,16 @@ def waitsendpay(self, payment_hash, timeout=None):
         }
         return self.call("waitsendpay", payload)
 
-    def pay(self, bolt11, msatoshi=None, description=None, riskfactor=None):
+    def pay(self, bolt11, msatoshi=None, label=None, riskfactor=None):
```
cdecker: I'd propose to just add both `label` and `description` and then check that only one is set when constructing the payload (assigning that one to `label`). We can also warn about `description` being phased out and remove it once we remove the first wave of deprecated APIs.

plugins/pay.c
```diff
@@ -1093,6 +1094,90 @@ static struct command_result *handle_paystatus(struct command *cmd,
...
+			copy_member(&ret, buf, t, "payment_preimage", ",");
+		copy_member(&ret, buf, t, "label", ",");
+		copy_member(&ret, buf, t, "amount_sent_msat", "");
+		tal_append_fmt(&ret, "}");
+	}
+	tal_append_fmt(&ret, "] }");
+	return command_success(cmd, ret);
+}
+
+static struct command_result *handle_listpays(struct command *cmd,
```
cdecker: So far we had the convention of calling these `json_<command_name>` which I found very useful for grepping them.

plugins/pay.c
```diff
@@ -1093,6 +1094,90 @@ static struct command_result *handle_paystatus(struct command *cmd,
...
+	char *ret;
+	bool some = false;
+
+	arr = json_get_member(buf, result, "payments");
+	if (!arr || arr->type != JSMN_ARRAY)
+		return command_fail(cmd, LIGHTNINGD,
+				    "Unexpected non-array result from listpayments");
+
+	ret = tal_fmt(cmd, "{ 'pays': [");
+	json_for_each_arr(i, t, arr) {
```
cdecker: Am I correct in thinking that this loop will eventually coalesce all parts of a multi-part payment?

contrib/pylightning/lightning/lightning.py
rustyrussell: Well, if we accept both, lightningd already does that check?

plugins/pay.c
rustyrussell: Yeah, good idea.

plugins/pay.c
rustyrussell: Yes.

contrib/pylightning/lightning/lightning.py
cdecker: Right :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2382/commits/a72782cc7d72c982f49a0214af5cc85101d58a26

PR #2386, opened 2019-02-22
pylightning: Fix missing encode if first pass fails

cdecker:
Without this the RPC will fail to continue buffering if the response does not
fit in the first read, and if we don't switch over to the non-compat
mode. This was introduced by our mitigation of the UTF-8 misalignment (#2355 ), but I
missed this path.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2386/commits/55d59afe812e7ca45531da6de5b2c1353051f106

PR #2387, opened 2019-02-22
Show addresses for their_unilateral/to_us outputs as well

### LINKED ISSUE #2080: Display the address on their_unilateral/to_us outputs
`listfunds` currently does not show the address in case of a unilateral close from the remote end. This is because the generation is dependent on the `per_commitment_point` that we store in the `close_info`, which we currently just skip. The address should be generated and shown there.

Reported-by: alan8325 on IRC

cdecker:
This was the only type of outputs we didn't show addresses until now, since it was a bit more involved, being derived from a combination of the outpoint and the private key. I decided to add the `scriptPubKey` to the outputs in these cases since otherwise we'd have to go talk to `hsmd` just to derive the pubkey and then hash it in order to generate the address, which is basically already what is in the `scriptPubKey`.

However I haven't found a good and clean way to backfill the missing ones for existing wallets since we don't track the utxos, and we can't call out to the `hsmd` in migrations. The number of affected users is minimal and it falls back gracefully to just not showing the address in those cases, so I decided not to add more code just to cover that corner case that'll quickly disappear :wink: 

Fixes #2080


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2387/commits/548a19ce5bf834c1059cd48936d1c0b971558a3d

Issue #2150, opened 2018-12-07
listpeers should show how many blocks until confirmation in status


rustyrussell:
### Issue and Steps to Reproduce

Currently:
            "CHANNELD_AWAITING_LOCKIN:Funding needs more confirmations."

### `getinfo` output
v0.6.2

darosior:
Hi,
I was looking for making a PR to solve this issue but I have some questions. 
How many confirmations are needed ? If it is more than one, do we start the count when the tx is included in a block or when it is broadcasted ?

trueptolemy:
hi, @rustyrussell i pulled the PR#2334 and try to fix this. what do u think of this solution? thanks

rustyrussell:
Fixed in #2405

Issue #1780, opened 2018-07-30
suggestion to fix the missing number of confirmations in static void billboard_update(const struct peer *peer)


renepickhardt:
looking at #1778 I found this FIXME tag: 
https://github.com/ElementsProject/lightning/blob/d3edfc802888b2df49a1db8b24a32ccc6ca81237/channeld/channel.c#L177

it asks to make the message for the user more specific by stating how many more confirmations of the funding tx have to happen so that the channel can be used. Afaik the funding tx has to have 6 confirmations. Since I was interacting a lot with the `lightningd.sqlite3` I wonder if something along the following line of pseudocode would actually fix this issue: 

```
		// struct bitcoin_txid * funding_tx =&(peer->channel->funding_txid);
		// funding_block_height = select blockheight from transaction where id == funding_tx->shad;
		// current_block_height = select select height from blocks order by height desc Limit 1;
		// confirmations_to_wait = 6 - (current_block_height - funding_block_height);
```

It seems ugly to me to resolve this via the database and db queries. On the other hand - as a newbe - this is currently the only way I am aware of  to gather the necessary information. Hence before providing the patch based on the sql solution I wanted to ask if that would be ok or if there is some preferable way of gathering this information?

cdecker:
The 6 confirmations is only relevant for the channel announcement and serves as a lower bound on the confirmations needed before others will accept a `channel_announcement` message. The actual number of confirmations needed before the channel becomes operational, i.e., moves into state `CHANNELD_NORMAL` depends on the `minimum_depth` in the `accept_channel` message (see [spec](https://github.com/lightningnetwork/lightning-rfc/blob/4b62d26af93a07166d6a9de2cb5eff80849d9c19/02-peer-protocol.md#the-accept_channel-message)). So this is a bit more complicated, but easily cached as well. As a matter of fact, since we have a reference to the `struct peer` we can easily just go an look at the `channel->scid` and extract the confirmation height from it (if set) and subtract it from `peer->ld->topology->tip->height` and compare it with `peer->channel->minimum_depth` :wink:

cdecker:
I took the liberty of assigning you :-)

renepickhardt:
I tried to fix this. However two problems:

1.) the code looks ugly and I am sure this can be done in a better way? 
[...]

PR #2405, opened 2019-02-26
fix: Add information about how many blocks to go until funding is confirmed

trueptolemy:
This is another try for issue #2150 and issue #1780 as @rustyrussell [suggested](https://github.com/ElementsProject/lightning/pull/2339#issuecomment-464540102):

1. Add `minimum_depth` in [`struct channel`](https://github.com/ElementsProject/lightning/blob/1dcc4823507df177bf11ca60ab7da988205139b1/common/initial_channel.h#L30) in common/initial_channel.h and change corresponding init function: [`new_initial_channel()`](https://github.com/ElementsProject/lightning/blob/1dcc4823507df177bf11ca60ab7da988205139b1/common/initial_channel.h#L86).

2. Add `confirmation_needed in` in [`struct peer`](https://github.com/ElementsProject/lightning/blob/1dcc4823507df177bf11ca60ab7da988205139b1/channeld/channeld.c#L67) in channeld/channeld.c.

3. Rename `channel_funding_locked` to `channel_funding_depth` in
channeld/channel_wire.csv.

4. Rename [`channel_tell_funding_locked`](https://github.com/ElementsProject/lightning/blob/1dcc4823507df177bf11ca60ab7da988205139b1/lightningd/channel_control.h#L18) to `channel_tell_depth`.

5. Rename [`funding_lockin_cb`](https://github.com/ElementsProject/lightning/blob/1dcc4823507df177bf11ca60ab7da988205139b1/lightningd/peer_control.c#L859) to `funding_depth_cb` in lightningd/peer_cntrol.c.

6. When funding depth changes, Master calls `channel_tell_depth` even if `depth` < `minimum_depth` and tell channeld the funding depth. But in `channel_tell_depth`, `lockin_complete` will be called iff `depth` > `minimum_depth`.

7. Channeld ignores the funding_depth unless `depth` >= `minimum_depth`(except to update billboard, and set `peer->confirmation_needed ` = `minimum_depth - depth`). Note `confirmation_needed` in `struct peer` must be 0 when funding hit `minimum_depth` in locally topology but it doesn't correspond to remote locking state because of network delay.

Thank you all for your comments and suggestions!


channeld/channel_wire.csv
```diff
@@ -62,10 +63,10 @@ channel_init,,last_remote_secret,struct secret
 channel_init,,lflen,u16
 channel_init,,localfeatures,lflen*u8
 
-# master->channeld funding hit new depth >= lock depth
-channel_funding_locked,1002
-channel_funding_locked,,short_channel_id,struct short_channel_id
-channel_funding_locked,,depth,u32
+# master->channeld funding hit new depth(funding locked if >= lock depth)
+channel_funding_depth,1002
+channel_funding_depth,,short_channel_id,?struct short_channel_id
```
SimonVrouwe: Ok, I believe normally cdecker does the first review for new contributors, but I will give already some comments (hopefully I give better advice this time).

That is a nice trick to get around first call where channel->scid=NULL !

channeld/channeld.c
```diff
@@ -164,9 +170,15 @@ static void billboard_update(const struct peer *peer)
 
 	if (peer->funding_locked[LOCAL] && peer->funding_locked[REMOTE])
 		funding_status = "Funding transaction locked.";
-	else if (!peer->funding_locked[LOCAL] && !peer->funding_locked[REMOTE])
-		/* FIXME: Say how many blocks to go! */
-		funding_status = "Funding needs more confirmations.";
+	else if (!peer->funding_locked[LOCAL] && !peer->funding_locked[REMOTE]) {
+		if(peer->funding_locked[LOCAL])
```
SimonVrouwe: This if() will never be true

channeld/channeld.c
```diff
@@ -153,6 +153,12 @@ struct peer {
 
 	/* Make sure peer is live. */
 	struct timeabs last_recv;
+
+	/* Must be 0 when funding locked locally. In fact, it keeps
+	 * watch over the changes on local topology.
+	 * Because of network delay, it doesn't correspond to remote
+	 * locking information. */
+	u32 confirmation_needed;
```
SimonVrouwe: A nitpit, how about a shorter name (depth_togo) ?

channeld/channeld.c
```diff
@@ -2381,38 +2393,57 @@ static void peer_reconnect(struct peer *peer,
...
+		if (!scid)
+			status_failed(STATUS_FAIL_INTERNAL_ERROR,
+						"funding channel id should be drived");
+		peer->short_channel_ids[LOCAL] = *scid;
+		if (!peer->funding_locked[LOCAL]) {
+			if (peer->confirmation_needed == 0)
+				status_failed(STATUS_FAIL_INTERNAL_ERROR,
+							"funding should be locked");
+			status_trace("funding_locked: sending commit index %"PRIu64": %s",
+				    	peer->next_index[LOCAL],
```
SimonVrouwe: There are spaces in these two lines. They usually show up when doing a `git show`as red blocks and probably caused by your editor doing some kind auto indent.

channeld/channeld.c
```diff
@@ -2720,6 +2751,7 @@ static void init_channel(struct peer *peer)
 	u8 *funding_signed;
 	const u8 *msg;
 	u32 feerate_per_kw[NUM_SIDES];
+	u32 minimum_depth;
```
SimonVrouwe: Now I may be wrong here, but isn't this the only place where `minimum_depth` field needs be added to channel.
And not to the other ones, i.e. `new_full_channel` and `new_initial_channel`? The `minimum_depth` is only relevant to a committed channel.

channeld/channeld.c
```diff
@@ -2381,38 +2393,57 @@ static void peer_reconnect(struct peer *peer,
...
-			     peer->next_index[LOCAL],
-			     type_to_string(tmpctx, struct pubkey,
-					    &peer->next_local_per_commit));
-		msg = towire_funding_locked(NULL,
-					    &peer->channel_id,
-					    &peer->next_local_per_commit);
-		sync_crypto_write(&peer->cs, PEER_FD, take(msg));
-		peer->funding_locked[LOCAL] = true;
-	}
+		billboard_update(peer);
```
SimonVrouwe: This `billboard_update` and the one below can be merged into a single one outside the if-else block.

channeld/channeld.c
```diff
@@ -2381,38 +2393,57 @@ static void peer_reconnect(struct peer *peer,
...
+	if (!fromwire_channel_funding_depth(tmpctx,
+					    msg,
+					    &scid,
+					    &depth))
+		master_badmsg(WIRE_CHANNEL_FUNDING_DEPTH, msg);
 
 	/* Too late, we're shutting down! */
 	if (peer->shutdown_sent[LOCAL])
 		return;
+	if (depth < peer->channel->minimum_depth) {
```
SimonVrouwe: another nitpick, a space above this lines makes it easier to read :)

SimonVrouwe: Maybe make the `unsigned int depth` also an `u32` ?

channeld/channeld.c
```diff
@@ -2381,38 +2393,57 @@ static void peer_reconnect(struct peer *peer,
...
-		msg = towire_funding_locked(NULL,
-					    &peer->channel_id,
-					    &peer->next_local_per_commit);
-		sync_crypto_write(&peer->cs, PEER_FD, take(msg));
-		peer->funding_locked[LOCAL] = true;
-	}
+		billboard_update(peer);
+	} else {
+		if (!scid)
+			status_failed(STATUS_FAIL_INTERNAL_ERROR,
```
SimonVrouwe: Just an suggestion, replace `status_failed` with an `assert(scid)`. Removing the `else` makes it slightly easier to read in my opinion.

SimonVrouwe:
Ok, I believe normally cdecker does the first review for new contributors, but I have already given some comments (hopefully better advice then the first one). Looks good :+1:

channeld/channel_wire.csv
trueptolemy: Thank you Simon, I always benefit from your comments and advice!

trueptolemy: > Ok, I believe normally cdecker does the first review for new contributors, but I will give already some comments (hopefully I give better advice this time).
> 
> That is a nice trick to get around first call where channel->scid=NULL !

Thank you Simon, I always benefit from your comments and advice!

channeld/channeld.c
trueptolemy: you are right. It's a redundant if() which duplicate with the next else if()

channeld/channeld.c
trueptolemy: good name:) !

channeld/channeld.c
trueptolemy: Thank you! I will adjust this and be careful.

channeld/channeld.c
trueptolemy: Yes, good idea!

channeld/channeld.c
trueptolemy: Thank you, your advice is great:)

channeld/channeld.c
trueptolemy: Great! It makes code slight!

trueptolemy:
> Ok, I believe normally cdecker does the first review for new contributors, but I have already given some comments (hopefully better advice then the first one). Looks good +1

Thank you again! Your advice are great :)

channeld/channeld.c
```diff
@@ -2381,36 +2390,58 @@ static void peer_reconnect(struct peer *peer,
...
+					"funding shouldn't be locked");
 
-	peer->announce_depth_reached = (depth >= ANNOUNCE_MIN_DEPTH);
+		peer->depth_togo = peer->channel->minimum_depth - depth;
 
-	/* Send temporary or final announcements */
-	channel_announcement_negotiate(peer);
+	} else {
+
+		assert(scid);
```
trueptolemy: @SimonVrouwe  hi, Simon. what about this? I didn't remove else() and just placed bilboard_update() outside the block?

SimonVrouwe: Sorry I mixed up a few things (didn't see there was no return in the first  `if`), so please ignore my previous comment. Looks good now!

trueptolemy: Thank you:)

rustyrussell:
OK, this looks great!  I'd usually say you should rebase and add a CHANGELOG.md message, but I'll do it in this case to avoid further delays :)

Thanks for your patience!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2405/commits/7a02553a309ed80c5a666d9f303c6054f4ad6399

trueptolemy:
@rustyrussell  Thanks for your suggestion:) 
Sorry, I will add the CHANGELOG.md next time!

channeld/channeld.c
```diff
@@ -2381,36 +2390,58 @@ static void peer_reconnect(struct peer *peer,
...
-					    &peer->next_local_per_commit));
-		msg = towire_funding_locked(NULL,
-					    &peer->channel_id,
-					    &peer->next_local_per_commit);
-		sync_crypto_write(&peer->cs, PEER_FD, take(msg));
-		peer->funding_locked[LOCAL] = true;
-	}
+	if (depth < peer->channel->minimum_depth) {
+		if (peer->funding_locked[LOCAL] || peer->depth_togo == 0)
+			status_failed(STATUS_FAIL_INTERNAL_ERROR,
```
cdecker: This error made me stumble a bit. Are we sure this can't happen during restarts with rescans?

trueptolemy: Hi, @cdecker I'm not sure if what I think is correct. Hope you don't mind my pool English.
 
In the original code, the channel will be saved in the DB in this [code block](https://github.com/ElementsProject/lightning/blob/0d35a7111a10af070bb6f9aaf53b6a27ca38b264/lightningd/peer_control.c#L875).
But now, to prevent saving channel with unlocked funding, I add the condition `local_locked && !channel->scid`. So, these channel with unlocked funding won't be saved until depth >= minimum_depth.

And the struct peer in Channeld always inits from [there](https://github.com/ElementsProject/lightning/blob/0d35a7111a10af070bb6f9aaf53b6a27ca38b264/channeld/channeld.c#L2902). The peer in Channeld won't be saved in DB and it is only dependent on the channel information in the Lightningd.

I suppose when we restart, we need rebuild this peer struct(in Channeld) according to the channel(in Lightningd). If the funding locked before, there may be 4 situations:
1. the depth >= minimum_depth(we can get the topo from DB) , and the channel has been saved in DB before: we can init the peer with `peer->funding_locked[LOCAL] == true`([there](https://github.com/ElementsProject/lightning/blob/0d35a7111a10af070bb6f9aaf53b6a27ca38b264/channeld/channeld.c#L2703) we use if `channel->scid` exist to init `funding_locked[LOCAL]`), and `peer->depth_togo == minimum_depth`(I directly set this value to peer->depth_togo). But after some minutes, Lightningd will tell us the depth >= minimum_depth, we won't meet this error.
2. the depth >= minimum_depth, but we didn't save the channel in time: we can init the peer with `peer->funding_locked[LOCAL] = false`, and we alse init the `peer->depth_togo = minimum_depth` directly. 
3. the depth < minimum_depth, and we saved the channel in the DB: it is impossible! we only save the channel under the condition that funding is locked.
4. the depth < minimum_depth, and we didn't save the channel: we can init peer with `funding_locked[LOCAL] = false` and `peer->depth_togo = minimum_depth`.

So I think we won't meet this error when restart. What do you think about this?

trueptolemy: Hi!@cdecker I think this error won't happen unless the topo reorganize with setting too samll minimum_depth.
The peer we used in Channeld is really initialed according the channel_init message from Lightning. And we never save the peer in the DB directly. In other words, during every start, we need initial the peer->funding_locked[LOCAL] from Lightningd(peer->funding_locked[LOCAL] corresponds if channel->scid exist). So the 4 situations I listed above are reasonable.

I think the only situation that this error happens is:
Lightningd find the funding depth changes and don't hit minimum_depth, then Lightningd tell us the depth now. But before that, funding has hit minimum_depth once and we has drived the channel->scid. The reason of this situation maybe the minimum_depth is too small and topo reorganized.
We don't limit the smallist minimum_depth(BOLT2 just ask mimimum_depth can't be too big, and clightning ask it must be smaller than 10). But I think we don't need to set the smallest minimum_depth. This condition here can be a kind of alert that reorg has happend.
what do you think about?

SimonVrouwe: I know this goes off-topic, but currently we [allow](https://github.com/ElementsProject/lightning/blob/e843e138fe43c48c6165d860ea99a627945a91a4/lightningd/options.c#L606) small `minimum_depth = 1` (the [default](https://github.com/ElementsProject/lightning/blob/e843e138fe43c48c6165d860ea99a627945a91a4/lightningd/options.c#L543) is 3), which is easily reorged and can change `scid`. I actually think that `scid` should not be calculated before `ANNOUNCE_MIN_DEPTH` (=6), do we really need the `scid` before that?

SimonVrouwe: So IMHO the above check can be removed.

channeld/channeld.c
```diff
@@ -2381,36 +2390,58 @@ static void peer_reconnect(struct peer *peer,
...
+				    	peer->next_index[LOCAL],
+				    	type_to_string(tmpctx, struct pubkey,
+					&peer->next_local_per_commit));
+
+			msg = towire_funding_locked(NULL,
+						    &peer->channel_id,
+						    &peer->next_local_per_commit);
+			sync_crypto_write(&peer->cs, PEER_FD, take(msg));
+			peer->funding_locked[LOCAL] = true;
+			peer->depth_togo = 0;
```
SimonVrouwe: I suggest move `peer->depth_togo = 0;` to line 2418, just after `else`. That makes it independent of `peer->funding_locked[LOCAL]`, so we (could) use it in case funding tx is reorged out _after_ lockin (with small `minimum_depth`) to update `depth_togo`.

channeld/channeld.c
```diff
@@ -2381,36 +2390,58 @@ static void peer_reconnect(struct peer *peer,
...
 
-	/* Send temporary or final announcements */
-	channel_announcement_negotiate(peer);
+	} else {
+
+		assert(scid);
+		peer->short_channel_ids[LOCAL] = *scid;
+
+		if (!peer->funding_locked[LOCAL]) {
+			if (peer->depth_togo == 0)
```
SimonVrouwe: IMHO this check can be removed.

channeld/channeld.c
SimonVrouwe: One other point d3aaa42 , we normally don't use merge commits, but instead `rebase` (on top of latest master).

You can do this using: 
```
git checkout <topic_branch>
git rebase master
```

channeld/channeld.c
```diff
@@ -153,6 +153,14 @@ struct peer {
 
 	/* Make sure peer is live. */
 	struct timeabs last_recv;
+
+	/* It means the number of confirmations needed before LOCKIN.
```
SimonVrouwe: How about: `Additional confirmations need for local lockin` ?

channeld/channeld.c
```diff
@@ -2849,6 +2884,9 @@ static void init_channel(struct peer *peer)
 	if (funding_signed)
 		sync_crypto_write(&peer->cs, PEER_FD, take(funding_signed));
 
+	/* from now we need keep watch over WIRE_CHANNEL_FUNDING_DEPTH */
+	peer->depth_togo = minimum_depth;
```
SimonVrouwe: Maybe this line should be moved to before `if (reconnected)` line 2880?
because `peer_reconnect` calls `billboard_update` indirectly, maybe it is safer to have `depth_togo` initialized?

SimonVrouwe:
> I think this error won't happen unless the topo reorganize with setting too samll minimum_depth.

I have been running some tests with reorgs. Because we [allow](https://github.com/ElementsProject/lightning/blob/e843e138fe43c48c6165d860ea99a627945a91a4/lightningd/options.c#L606) small `minimum_depth = 1`, so we can operate the channel already after 1 confirm (for direct payments only, without announcing), the funding tx may be _reorged_ out and then mined in another block _after_ `lockin`. 

When that happens, the `short_channel_id` should be recalculated and stored in DB. Or perhaps we should wait with calculating `scid` until  `ANNOUNCE_MIN_DEPTH`? Not sure if that belongs in this PR or a new one.

When funding tx is reorged out, the `funding_depth_cb` is [called](https://github.com/ElementsProject/lightning/blob/e843e138fe43c48c6165d860ea99a627945a91a4/lightningd/chaintopology.c#L663) with depth=0.

Maybe we should add a python test for these kind of reorgs? I can give it a shot.

I made a [branch](https://github.com/SimonVrouwe/lightning/tree/trueptolemy_pr2405_fixed2) based on your PR which includes suggested modifications.

channeld/channeld.c
```diff
@@ -165,8 +173,9 @@ static void billboard_update(const struct peer *peer)
 	if (peer->funding_locked[LOCAL] && peer->funding_locked[REMOTE])
 		funding_status = "Funding transaction locked.";
 	else if (!peer->funding_locked[LOCAL] && !peer->funding_locked[REMOTE])
-		/* FIXME: Say how many blocks to go! */
-		funding_status = "Funding needs more confirmations.";
+		funding_status = tal_fmt(tmpctx,
+					"Funding needs %d confirmations.",
```
SimonVrouwe: Finally, can I suggest a subtle change:
`Funding needs %d confirmations.` -> `Funding needs %d confirmations before lockin.`

Hopefully this makes it also more clear that a lockin is not reversed with a reorg.

trueptolemy:
@SimonVrouwe Thank you and you help me sooooo much:) ! Let me have a look!

channeld/channeld.c
trueptolemy: @SimonVrouwe You'are right, and I should delete these checks here.
And I also notice a extreme case: 
Suppose we set a private channel (don't set announcement flag), and set a too small minimum locking depth (as you say, 1). When we notice the funding tx locked and we drive `short_channel_id` and store into the DB, but unfortunately, our node crash. When we restart, the reorg happens.
Now we initial our channel from DB with old `short_channel_id`, and we may can't be told that reorg happened because we delete the corresponding [`watchtx`](https://github.com/ElementsProject/lightning/blob/c7ab5102ba68ef90209bb4cfc0ff1e63fa68598a/lightningd/peer_control.c#L907). 
The next step the short_channel_id will be use is update_htlc_add (there need a check?).

channeld/channeld.c
trueptolemy: good description!

channeld/channeld.c
SimonVrouwe: > ... and we may can't be told that reorg happened because we delete the corresponding watchtx.

Good catch!
Wallet transactions are never deleted from the db, but at reorg their `height` field is [set](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/wallet/db.c#L311) to NULL. When a block is reorged-out, all tx's with that height are [selected](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/lightningd/chaintopology.c#L658), and their `cb` is [fired](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/lightningd/chaintopology.c#L663) with depth 0. 

But if the `watch` was [deleted](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/lightningd/peer_control.c#L908), after only 1 confirm when `minimum_depth=1` and ANNOUNCE_FLAG unset, the [watch](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/lightningd/watch.c#L140) is gone and [nothing](https://github.com/ElementsProject/lightning/blob/2af3996bea21c91945e621aa49fe461fe3e39a0e/lightningd/watch.c#L247) happens!

@cdecker So this looks like a bug? Although it is a _very_ rare case (how often do tx get reorged to another height?), I will make a new issue about this.

channeld/channeld.c
trueptolemy: Yes!
`peer_reconnect` calls `peer_bilboard`(it has some difference with `bilboard_update`), but it's better to initial `depth_togo` ASAP!

channeld/channeld.c
trueptolemy: Thank you!
Now I'll rebase it!

trueptolemy: > I suggest move `peer->depth_togo = 0;` to line 2418, just after `else`. 

> EDIT3: on second thought, updating `depth_togo` after `lockin` doesn't make sense, `lockin` stays `lockin` also after reorg.

you are right! It's more logical!

channeld/channeld.c
trueptolemy: Nice descripthion! :)

channeld/channeld.c
trueptolemy: It's a really rare case. The main problem I think is too samll minimum_depth. 
Maybe we can set limit of the minimum_depth, or any other operations to handle it(fail the channel, or generate warning...).

trueptolemy:
@cdecker Thank you, and I've rebased it :)
I delete the check as you marked and do some change as @SimonVrouwe  suggested.
But there is a reorg case (maybe a bug we don't handle now) as Simon and I descripted in [here](https://github.com/ElementsProject/lightning/pull/2405#discussion_r266197632) and [here](https://github.com/ElementsProject/lightning/pull/2405#discussion_r265867935).
The original check is for this case, but failing the program directly is not approprite. The way to handle this case maybe need a another issue and PR.

rustyrussell:
Ack 7f55203e05fc6d1354c0db2367dd1d720dd0c7a9

rustyrussell:
I am *so sorry* that this merge took so long!  Great work though!

PR #2410, opened 2019-02-27
Don't count uncommitted channels towards num_pending in getinfo

conscott:
I orginally added the `num_x_channels` fields to `getinfo`, and wasn't quite sure how to handle the case of `peer->uncommitted_channel`, so I lumped it into `pending_channels`.

I now think this is a bit wrong, as it can stay in this state indefinitely if no channel is made, and it might be misleading to show it as pending.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1291,10 +1291,6 @@ static struct command_result *json_getinfo(struct command *cmd,
     /* Add some peer and channel stats */
     list_for_each(&cmd->ld->peers, peer, list) {
         num_peers++;
-        /* Count towards pending? */
-        if (peer->uncommitted_channel) {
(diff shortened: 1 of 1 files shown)


cdecker:
What's the rationale for adding this info to `getinfo`? I'm generally skeptical when adding redundant information (this could be grabbed much better from `listpeers`).

conscott:
@cdecker - agreed, i just gave it a shot since the other aggregates where there already. 

I updated the PR so that uncommitted channels are just not counted towards pending count, which is probably more appropriate.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2410/commits/80649dcabc8b80104748c75e0ddda185af3c550b

Issue #2377, opened 2019-02-21
Unable to pay invoice with docker deployment


sidaudhi:
### Issue and Steps to Reproduce
The issue seems to be related to #2282 but specific to the docker image elementsproject/lightningd
Using docker-compose i re-installed the container after pulling the latest image and when I execute:
```
docker exec lightning-cli pay <bolt1>
```
I get the error:
```
{ "code" : -32601, "message" : "Unknown command 'pay'" }
```
### `getinfo` output
```
{
  "id": "0230d70b0506aa447a1db407cc04e41448ab92bbd373f5c9a1029a40f357fcdcad",
  "alias": "LN.pay",
  "color": "20b0cf",
  "num_peers": 11,
  "num_pending_channels": 2,
  "num_active_channels": 9,
  "num_inactive_channels": 0,
  "address": [
    {
      "type": "ipv4",
      "address": "138.68.231.196",
      "port": 9735
    }
  ],
  "binding": [
    {
      "type": "ipv4",
      "address": "0.0.0.0",
      "port": 9735
    }
  ],
  "version": "v0.6.3rc1-118-g0da4054",
  "blockheight": 564007,
  "network": "bitcoin",
  "msatoshi_fees_collected": 0
}
```

wailo:
I am experiencing the same. Using both the provided docker image and building the sourcecode manually.

wailo:
Managed to get it work. Had to pass `--plugin-dir=PATH-TO-PAY-PLUGIN` to lightningd process

sidaudhi:
How would this look like in a docker deployment? I tried this:

```
clightning_bitcoin:
    image: elementsproject/lightningd
    container_name: lightning
    command:
      - --bitcoin-rpcconnect=bitcoind
      - --bitcoin-rpcuser=rpcuser
      - --bitcoin-rpcpassword=rpcpass
      - --network=bitcoin
      - --alias=LN.pay
      - --rgb=20B0CF
      - --log-level=debug
      - --bind-addr=0.0.0.0:9735
      - --announce-addr=138.68.231.196:9735
      - --fee-base=0
      - --fee-per-satoshi=6000
      - --plugin-dir=/root/plugins
    environment:
      EXPOSE_TCP: "true"
    expose:
      - "9735"
    ports:
      - target: 9735
        published: 9735
        protocol: tcp
        mode: host
    volumes:
      - type: bind
[...]

PR #2418, opened 2019-02-28
Installing the Pay plugin in Dockerfile

wailo:
This is a proposed fix for #2377.

**Current behavior**
When following instructions in readme.md with using docker image, running `lightning-cli` Pay will generate:  ` { "code" : -32601, "message" : "Unknown command 'pay'" }`

**Proposed fix**
Copy the Pay plugin to `/usr/libexec/c-lightning/plugins/` and use --plugin-dir argument to point to the plugins directory.

**Note to the maintainers**
After applying this fix, the docker image `elementsproject/lightningd` should be updated and uploaded to Dockerhub

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,6 +96,7 @@ VOLUME [ "/root/.lightning" ]
 
 COPY --from=builder /opt/lightningd/cli/lightning-cli /usr/bin
 COPY --from=builder /opt/lightningd/lightningd/lightning* /usr/bin/
+COPY --from=builder /opt/lightningd/plugins/pay /usr/libexec/c-lightning/plugins/
 COPY --from=builder /opt/bitcoin/bin /usr/bin
 COPY --from=builder /opt/litecoin/bin /usr/bin
 COPY tools/docker-entrypoint.sh entrypoint.sh
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ services:
       - --bitcoin-rpcconnect=bitcoind
       - --bitcoin-rpcuser=rpcuser
       - --bitcoin-rpcpassword=rpcpass
+      - --plugin-dir=/usr/libexec/c-lightning/plugins
       - --network=testnet
       - --alias=myawesomenode
       - --log-level=debug


m-schmoock:
ack

:D

m-schmoock:
ack

m-schmoock:
Seems good to me

m-schmoock:
Maybe wildcard for further plugins:

```
COPY --from=builder /opt/lightningd/plugins/* /usr/libexec/c-lightning/plugins/
```

rustyrussell:
Wildcard would be nice, but this is tested, so it wins for now given we're about to release!

rustyrussell:
Ack 7c3cea343694bf2fad6421f3e6ec9b7b13e66319

Thanks for this!

wailo:
Thank you. Want to add that using a wildcard will copy all files including non-executables, like pay.c, pay.h ..etc

samvid25:
Thanks for the fix. I was running into the same issue as #2377. But it seems that Dockerhub has not been updated with the modified Dockerfile yet.

NicolasDorier:
Does not seem to work on my side... diving into it. (the arm dockerfile has not been updated as well)

wailo:
Correct, the image on dockerhub is outdated. I left a note in the PR for the maintainers to update it.
@NicolasDorier I have posted a temporary workaround here 
https://github.com/ElementsProject/lightning/issues/2377#issuecomment-468447829

wailo:
I just noticed that the image is now updated on dockerhub and it works.

NicolasDorier:
@wailo yeah, needed this PR as well though https://github.com/ElementsProject/lightning/pull/2428

wailo:
Depends on how you start the container. The docker-compose file in the readme file does already passes the —plugin-dir argument

NicolasDorier:
@wailo  let's move it https://github.com/ElementsProject/lightning/pull/2428

I think that plugin-dir should be here by default, because this is what the user most likely expect.

If not the case, it is possible to override this through additional command args  in docker-compose or through configuration anyway.

Issue #2400, opened 2019-02-25
Backup and other feature requests


Elmit2015:
I feel increasingly uncomfortable if there is:

- no reliable backup
- parallel channels or both side funded channels
- channel top up payments 

Maybe we can setup a voting system to get features ranked by importance, like the video voting of Rene.

I know there are hundreds of things to do.

Coinomatron:
As a test, I'm mirroring the .lightning/ folder via a bash script:

while true; do inotifywait -r -e modify,create,delete /home/user/.lightning/; rsync -avz /home/user/.lightning/ /home/user/LN_folder_backup/; done

Seems to work, just restarted the client using the backup dir and made a payment. Maybe someone else can verify if this is a sufficient method for backing up.

rustyrussell:
The Right Way is to have a plugin with hooks at the db level, so you back up before c-lightning continues.  That plugin could back up anywhere.  I'm looking at it now, for 0.7.1; agreed it's vital!

MrManPew:
Do that and I'm switching from lnd to c-lightning :)

ZmnSCPxj:
> The Right Way is to have a plugin with hooks at the db level

Does this mean the plugin needs an implementation of SQLite3?

What exactly are the db-level hooks we want? Do we ask SQLite to provide a correctly-escaped query for every db query we make, then pass that sanitized query to the hook before invoking SQLite on the sanitized query?

trueptolemy:
> What exactly are the db-level hooks we want? 

I think it's db_write hook #2421 , and it passes database changes to hook.
I also think about the plugin to backup, and aslo not sure if plugin should implemente a SQLite directly.
It seems indeed need.

trueptolemy:
> Does this mean the plugin needs an implementation of SQLite3?

Another idea is:
Based the database, we can extract the channel event and implement the compatibility with SCB for Lnd.

ZmnSCPxj:
> Based the database, we can extract the channel event and implement the compatibility with SCB for Lnd.

If I understand correctly, recovering an SCB in lnd simply marks the channel such that on next reconnect to the peer, lnd falsely accuses it of not following the protocol, causing the peer to unilaterally close in despair.

If the peer suspects your node is lnd, and that you were using an SCB, it could replay old state and the lnd node will not be able to revoke it.

The only safe backup in Poon-Dryja is replicating the database.

trueptolemy:
[...]

PR #2421, opened 2019-03-01
db-write hook

rustyrussell:
It's a bit limited (see final commits for caveats) but it works.

See #2400

### DIFF
--- a/ccan/ccan/io/io.h
+++ b/ccan/ccan/io/io.h
@@ -723,6 +723,35 @@ bool io_plan_out_started(const struct io_conn *conn);
 bool io_flush_sync(struct io_conn *conn);
 
 /**
+ * io_conn_exclusive - set/unset an io_conn to exclusively serviced
+ * @conn: the connection
+ * @exclusive: whether to be exclusive or not
+ *
+ * If any io_conn is set exclusive, then no non-exclusive io_conn (or
+ * io_listener) will be serviced by io_loop().  If it's a io_duplex io_conn(),
+ * then io_conn_exclusive() makes the read-side exclusive; io_conn_out_exclusive()
+ * makes the write-side exclusive.
+ *
+ * This allows you to temporarily service only one (or several) fds.
+ * For example, you might want to flush out one io_conn and not
+ * receive any new connections or read any other input.
+ *
+ * Returns true if any exclusive io_conn remain, otherwise false.
+ * (This is useful for checking your own logic: dangling exclusive io_conn
+ * are dangerous!).
+ */
+bool io_conn_exclusive(struct io_conn *conn, bool exclusive);
+
+/**
+ * io_conn_out_exclusive - set/unset exclusive on the write-side of a duplex
+ * @conn: the connection, post io_duplex
+ * @exclusive: whether to be exclusive or not
+ *
+ * See io_conn_exclusive() above.
+ */
+bool io_conn_out_exclusive(struct io_conn *conn, bool exclusive);
+
+/**
  * io_fd_block - helper to set an fd blocking/nonblocking.
  * @fd: the file descriptor
  * @block: true to set blocking, false to set non-blocking.
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -16,8 +16,6 @@ variety of ways:
    internal events in `lightningd` and alter its behavior or inject
    custom behaviors.
    
-*Note: Hooks are not yet implemented, and the API is under active development.*
-
 A plugin may be written in any language, and communicates with
 `lightningd` through the plugin's `stdin` and `stdout`. JSON-RPCv2 is
 used as protocol on top of the two streams, with the plugin acting as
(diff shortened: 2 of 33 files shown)


rustyrussell:
Hmm, needs sqlite3 v3.14 or above.  Not sure what that would break if we include it, might have to make hook optional?  Yech...

cdecker:
I was expecting to see an implementation of `sqlite3_expanded_sql` took me while to figure out that it is actually a native `sqlite3` function. Maybe we can shim this? This hook seems far too useful not to exist. Notice though that it makes switching DBs even harder since plugins now may depend on the `sqlite3` dialect being used.

wallet/db.h
```diff
@@ -75,6 +78,35 @@ void db_set_intvar(struct db *db, char *varname, s64 val);
...
+ */
+#define db_select_prepare(db, query) \
+	db_select_prepare_(__FILE__ ":" stringify(__LINE__), db, query)
+sqlite3_stmt *db_select_prepare_(const char *location,
+				 struct db *db, const char *query);
+
+/**
+ * db_select_step -- iterate through db results.
+ *
+ * Returns false and frees stmt if we've we've reached end, otherwise
```
niftynei:
```suggestion
 * Returns false and frees stmt if we've reached end, otherwise
```

doc/PLUGINS.md
```diff
@@ -270,5 +268,25 @@ gossiped list of known addresses. In particular this means that the port for
...
 reconnections.
 
+#### `db_write`
+
+This hook is called whenever a change is about to be committed to the database.
+It is currently extremely restricted:
+
+1. a plugin registering for this hook should not perform anything that may cause
+   a db operation in response (pretty much, anything but logging).
+2. a plugin registering for this hook should not register for other hooks or
```
niftynei: should we canonize this by returning an error if this is broken?

ccan/ccan/io/io.h
```diff
@@ -722,6 +722,35 @@ bool io_plan_out_started(const struct io_conn *conn);
...
+ * @exclusive: whether to be exclusive or not
+ *
+ * If any io_conn is set exclusive, then no non-exclusive io_conn (or
+ * io_listener) will be serviced by io_loop().  If it's a io_duplex io_conn(),
+ * then io_conn_exclusive() makes the read-side exclusive; io_conn_out_exclusive()
+ * makes the write-side exclusive.
+ *
+ * This allows you to temporarily service only one (or several) fds.
+ * For example, you might want to flush out one io_conn and not
+ * receive any new connections or read any otherninput.
```
niftynei:
```suggestion
 * receive any new connections or read any other input.
```

ccan/ccan/io/io.h
```diff
@@ -722,6 +722,35 @@ bool io_plan_out_started(const struct io_conn *conn);
...
+ *
+ * Returns true of there any exclusive io_conn remain, otherwise false.
```
niftynei:
```suggestion
 * Returns true if any exclusive io_conn remain, otherwise false.
```

rustyrussell:
> I was expecting to see an implementation of `sqlite3_expanded_sql` took me while to figure out that it is actually a native `sqlite3` function. Maybe we can shim this? This hook seems far too useful not to exist. Notice though that it makes switching DBs even harder since plugins now may depend on the `sqlite3` dialect being used.

I think we can try to use the (now obsolete) sqlite3 tracing feature to simulate this in older sqlite3.  Let me try.

rustyrussell:
> really like the `db_select_step` improvements.
> 
> may be worth noting that in theory, your 'backup' copy may be one command set ahead of the actual on disk database if it crashes/fails during write... tho i guess that's actually the point eh. :)

Yes, but in this case both dbs are "valid".  You could naively copy the whole db in the plugin every startup, but better would be a separate "db_init" hook, which would use https://www.sqlite.org/c3ref/c_fcntl_begin_atomic_write.html#sqlitefcntldataversion, and give plugin the version startup time.

If the plugin also has an sqlite3 db, that's a trivial check, and if not, it can fall back to copy.

wallet/db.h
rustyrussell: :+1:

rustyrussell:
OK, put in backwards compat shim, rebased.

rustyrussell:
Ack 1d776eb4c1e264913715dbd1bcbb6ba534686bb7

Let's get it in so we can make progress :)

Coinomatron:
Updated to the latest commits, it's not clear to me how this backup function works. Can someone give a brief explanation?

cdecker:
> Updated to the latest commits, it's not clear to me how this backup function works. Can someone give a brief explanation?

It's not a backup mechanism on its own, it is the infrastructure necessary to write backup plugins on top.

Coinomatron:
Ah, thanks.

cdecker:
No problem, we'll get a move on and get some example implementations soon :-)

Coinomatron:
Any updates on this?

ZmnSCPxj:
My last information is that @jb55 was working on some backup system that implements this.

jb55:
oh yeah I should start that

jb55:
hmm now that you reminded me, I didn't go that route as I was going to see how the postgresql backend played out. if you could use a postgresql backend that replicates offsite, that perhaps could solve the problem as well? cc @cdecker 

although that's not very user friendly, so perhaps a plugin is still in order...

that makes me wonder how this interacts with the multi-db abstraction now... :thinking: I haven't looked at the code in either case yet

PR #2428, opened 2019-03-02
Use plugin directory in docker-entrypoint

NicolasDorier:
Even if the plugin directory exist on the container, by default `lightningd` do not use it.

### DIFF
--- a/tools/docker-entrypoint.sh
+++ b/tools/docker-entrypoint.sh
@@ -4,7 +4,7 @@
 
 if [ "$EXPOSE_TCP" == "true" ]; then
     set -m
-    lightningd "$@" &
+    lightningd --plugin-dir=/usr/libexec/c-lightning/plugins "$@" &
 
     echo "C-Lightning starting"
     while read -r i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
@@ -15,5 +15,5 @@ if [ "$EXPOSE_TCP" == "true" ]; then
     socat "TCP4-listen:$LIGHTNINGD_RPC_PORT,fork,reuseaddr" "UNIX-CONNECT:$LIGHTNINGD_DATA/lightning-rpc" &
     fg %-
 else
-    exec lightningd "$@"
+    exec lightningd --plugin-dir=/usr/libexec/c-lightning/plugins "$@"
 fi


wailo:
The essential issue is that in docker, lightnignd can not find plugins in the default directory. I think main issue is: 
1. In the dockerfile, the build is configured with the defult prefix path `/usr/local`
2. In the dockerfile, exectuables are installed in `/usr` folder,
3. At runtime, lightnignd is expecting the plugins to be at `../libexec/c-lightning/plugins` Where that path is relative to `/user/local/bin`
4. As a result, plugins will not be found.

A potential solution is to add `—prefix=${PREFIX}` to configure in the Dockerfile, where PREFIX is the install path, then make sure to copy all executable to the same path. After that, `—plugin-dir` Argument is not needed at all. This solution remove the need to add `—plugin-dir` everywhere and might also make it possible to do a single copy statement for all executable using wildcard, if the prefix is not a system path, i.e not /usr. 
I haven’t tested this solution yet.

NicolasDorier:
Or what about copying the plugins to `/usr/local/libexec/c-lightning/plugins` instead of `/usr/libexec/c-lightning/plugins`? This would work right?

wailo:
Yes, and do the same to all executables and for every new plugin that will be added in the future.
A better solution imo is to use --prefix option. Below is a working implementation that I have tested. It is cleaner when it comes to copying binaries, and there is no need to use `--plugin-dir` in docker-compose.yaml and entrypoint.sh anymore 

Key changes:
- ARG PREFIX=/tmp/lightning_install
- RUN ./configure **--prefix=$PREFIX** && make -j3 DEVELOPER=${DEVELOPER} && **make install**
- COPY --from=builder $PREFIX/ /usr/local/
- Removed lightning* COPY commands


```
ARG PREFIX=/tmp/lightning_install

FROM alpine:3.7 as builder

RUN apk add --no-cache \
     ca-certificates \
     autoconf \
     automake \
     build-base \
     libressl \
     libtool \
     gmp-dev \
     python \
     python-dev \
     python3 \
     sqlite-dev \
     wget \
     git \
     file \
     gnupg \
     swig \
     zlib-dev

WORKDIR /opt
ARG PREFIX
ARG BITCOIN_VERSION=0.17.0
ENV BITCOIN_TARBALL bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/$BITCOIN_TARBALL
ENV BITCOIN_ASC_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/SHA256SUMS.asc
ENV BITCOIN_PGP_KEY 01EA5486DE18A882D4C2684590C8019E36C2E964

RUN mkdir /opt/bitcoin && cd /opt/bitcoin \
    && wget -qO $BITCOIN_TARBALL "$BITCOIN_URL" \
    && gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "$BITCOIN_PGP_KEY" \
    && wget -qO bitcoin.asc "$BITCOIN_ASC_URL" \
    && gpg --verify bitcoin.asc \
    && grep $BITCOIN_TARBALL bitcoin.asc | tee SHA256SUMS.asc \
    && sha256sum -c SHA256SUMS.asc \
    && BD=bitcoin-$BITCOIN_VERSION/bin \
    && tar -xzvf $BITCOIN_TARBALL $BD/bitcoin-cli --strip-components=1 \
    && rm $BITCOIN_TARBALL

ENV LITECOIN_VERSION 0.16.3
ENV LITECOIN_PGP_KEY FE3348877809386C
ENV LITECOIN_URL https://download.litecoin.org/litecoin-${LITECOIN_VERSION}/linux/litecoin-${LITECOIN_VERSION}-x86_64-linux-gnu.tar.gz
ENV LITECOIN_ASC_URL https://download.litecoin.org/litecoin-${LITECOIN_VERSION}/linux/litecoin-${LITECOIN_VERSION}-linux-signatures.asc

# install litecoin binaries
RUN mkdir /opt/litecoin && cd /opt/litecoin \
    && wget -qO litecoin.tar.gz "$LITECOIN_URL" \
    && gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "$LITECOIN_PGP_KEY" \
    && wget -qO litecoin.asc "$LITECOIN_ASC_URL" \
    && gpg --verify litecoin.asc \
    && BD=litecoin-$LITECOIN_VERSION/bin \
    && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
    && rm litecoin.tar.gz

ENV LIGHTNINGD_VERSION=master

WORKDIR /opt/lightningd
COPY . /tmp/lightning
RUN git clone --recursive /tmp/lightning . && \
    git checkout $(git --work-tree=/tmp/lightning --git-dir=/tmp/lightning/.git rev-parse HEAD)

ARG DEVELOPER=0
RUN ./configure --prefix=$PREFIX && make -j3 DEVELOPER=${DEVELOPER} && make install

FROM alpine:3.7

RUN apk add --no-cache \
     gmp-dev \
     sqlite-dev \
     inotify-tools \
     socat \
     bash \
     zlib-dev

ENV GLIBC_VERSION 2.27-r0
ENV GLIBC_SHA256 938bceae3b83c53e7fa9cc4135ce45e04aae99256c5e74cf186c794b97473bc7
ENV GLIBCBIN_SHA256 3a87874e57b9d92e223f3e90356aaea994af67fb76b71bb72abfb809e948d0d6
# Download and install glibc (https://github.com/jeanblanchard/docker-alpine-glibc/blob/master/Dockerfile)
RUN apk add --update curl && \
  curl -Lo /etc/apk/keys/sgerrand.rsa.pub https://github.com/sgerrand/alpine-pkg-glibc/releases/download/$GLIBC_VERSION/sgerrand.rsa.pub && \
  curl -Lo glibc.apk "https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-${GLIBC_VERSION}.apk" && \
  echo "$GLIBC_SHA256  glibc.apk" | sha256sum -c - && \
  curl -Lo glibc-bin.apk "https://github.com/sgerrand/alpine-pkg-glibc/releases/download/${GLIBC_VERSION}/glibc-bin-${GLIBC_VERSION}.apk" && \
  echo "$GLIBCBIN_SHA256  glibc-bin.apk" | sha256sum -c - && \
  apk add glibc-bin.apk glibc.apk && \
  /usr/glibc-compat/sbin/ldconfig /lib /usr/glibc-compat/lib && \
  echo 'hosts: files mdns4_minimal [NOTFOUND=return] dns mdns4' >> /etc/nsswitch.conf && \
  apk del curl && \
  rm -rf glibc.apk glibc-bin.apk /var/cache/apk/*

ENV LIGHTNINGD_DATA=/root/.lightning
ENV LIGHTNINGD_RPC_PORT=9835

VOLUME [ "/root/.lightning" ]
ARG PREFIX
COPY --from=builder $PREFIX/ /usr/local/
COPY --from=builder /opt/bitcoin/bin /usr/bin
COPY --from=builder /opt/litecoin/bin /usr/bin
COPY tools/docker-entrypoint.sh entrypoint.sh

EXPOSE 9735 9835
ENTRYPOINT  [ "./entrypoint.sh" ]
```

NicolasDorier:
Both way are fine for me, though I will prefer moving plugings on `/usr/local/libexec/c-lightning/plugins` which is just a one line change on `COPY`.

I think it should not `ARG`, this should not be customized by the builder. Instead it should be ENV.

If you are motivated to do a PR for your solution, I can give it a try.

wailo:
> though I will prefer moving plugings on /usr/local/libexec/c-lightning/plugins which is just a one line change on COPY.

I tried that before using a wildcard, but other unneeded files were copied as well, such as source code and build artifacts. If you know a way to copy execuatables only using a wildcard with docker COPY, that will be ideal.
I have created a fork with my changes on it, I still have to test with using ENV instead or ARG.
I am happy to do a PR and also happy if you incorporate changes in this PR to keep this dialogue.

NicolasDorier:
I don't understand what you say here

```diff
- COPY --from=builder /opt/lightningd/plugins/pay /usr/libexec/c-lightning/plugins/
COPY --from=builder /opt/lightningd/plugins/pay/usr/local/libexec/c-lightning/plugins/
```
Should work as is right?

wailo:
Oh I interpreted plugins (with plural s) as a collection of plugins, that’s why I brought up wildcard which is preferred instead of a copy line per each new plugin.

I am thinking about how to deal with addition of new plugins or binaries into the project. The line you posted works for now, but as soon as a new plugins is created, the Dockerfile will be outdated.
PR created: https://github.com/ElementsProject/lightning/pull/2451

NicolasDorier:
I prefer https://github.com/ElementsProject/lightning/pull/2451

PR #2429, opened 2019-03-02
build: libsodium switched case in constants, add check for compat

### LINKED ISSUE #2425: libsodium compile error
I'm using Ubuntu 16.04 LTS. I've tried to compile c-lightning v0.7.0, but it failed with the following errors:
```
common/cryptomsg.c: In function ‘le64_nonce’:
common/cryptomsg.c:87:25: error: ‘crypto_aead_chacha20poly1305_ietf_NPUBBYTES’ undeclared (first use in this function)
  const size_t zerolen = crypto_aead_chacha20poly1305_ietf_NPUBBYTES - sizeof(le_nonce);
                         ^
common/cryptomsg.c:87:25: note: each undeclared identifier is reported only once for each function it appears in
common/cryptomsg.c: In function ‘cryptomsg_decrypt_body’:
common/cryptomsg.c:98:21: error: ‘crypto_aead_chacha20poly1305_ietf_NPUBBYTES’ undeclared (first use in this function)
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
                     ^
common/cryptomsg.c:98:16: error: unused variable ‘npub’ [-Werror=unused-variable]
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
                ^
common/cryptomsg.c: In function ‘cryptomsg_decrypt_header’:
common/cryptomsg.c:132:21: error: ‘crypto_aead_chacha20poly1305_ietf_NPUBBYTES’ undeclared (first use in this function)
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
                     ^
common/cryptomsg.c:132:16: error: unused variable ‘npub’ [-Werror=unused-variable]
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
                ^
common/cryptomsg.c: In function ‘cryptomsg_encrypt_msg’:
common/cryptomsg.c:164:21: error: ‘crypto_aead_chacha20poly1305_ietf_NPUBBYTES’ undeclared (first use in this function)
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
                     ^
common/cryptomsg.c:164:16: error: unused variable ‘npub’ [-Werror=unused-variable]
  unsigned char npub[crypto_aead_chacha20poly1305_ietf_NPUBBYTES];
```
I thought this is related to libsodium, so I checked it is surely installed:
```
$ sudo apt install libsodium-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libsodium-dev is already the newest version (1.0.8-5).
```
The version number seems to be old.
I assume libsodium related includes may come from external/libsodium. Instead in my case they look like coming from /usr/include/sodium/. And there are some minor changes in some variables' name between libsodium versions.

It may be worth mentioning when I run `./configure` I see this line: `checking for libsodium with IETF chacha20 variants... yes`.

### A workaround
I don't know what went wrong, but only a few lines of code modification needed in the lightning source to fix it. I attach what I did to manage to compile.
[libsodium.patch.txt](https://github.com/ElementsProject/lightning/files/2921972/libsodium.patch.txt)
After that everything seems to work fine!

### `getinfo` output
[Everything Is AWESOME!!!](https://www.youtube.com/watch?v=StTqXEQ2l-Y)

cdecker:
It seems libsodium broke their public interface at some point by switching
from upper-case IETF in constants to lower-case. This adds a check for the
lower-case version which seems to be what they are going with.

Fixes #2425 

Reported-by: Zoltán Gálli <@gallizoltan>

### DIFF
--- a/configure
+++ b/configure
@@ -168,6 +168,7 @@ code=
 int main(void)
 {
 	printf("%p\n", crypto_aead_chacha20poly1305_ietf_encrypt);
+	printf("%d\n", crypto_aead_chacha20poly1305_ietf_NPUBBYTES);
 	exit(0);
 }
 /*END*/


rustyrussell:
This seems to check for upper case? I think we could work around it with a strategic #define if we wanted...

gallizoltan:
This solves the problem!

I've tried to compile c-lightning with this patch, and in my case the `./configure` said `checking for libsodium with IETF chacha20 variants... no`.
After that `make` successfully run.

Thank you very much!

cdecker:
> This seems to check for upper case? I think we could work around it with a strategic #define if we wanted...

We could add `#define`s to even out this change, and they seem to have tried to do that themselves (https://github.com/jedisct1/libsodium/commit/8c0b91672949dbcbd8223323a6528d4fdafa0597#diff-2418b954b1f2d53d5ddd8aed9e66f39fR97). Are we sure we want to allow building on an old version that might have any number of other changes though?

rustyrussell:
I'll replace this commit msg with more precision, since I didn't understand initially:

The v1.0.9 release of libsodium added `crypto_aead_chacha20poly1305_ietf_NPUBBYTES` which we use; before that it was `...IETF_NPUBBYTES`.  Since that release was in April 2016, it seems fair to simply check for ancient versions and use the internal one if found.  The alternative would be to use the older names (which are still in the header), but given we've never tested with such old versions, this seems safer.

PR #2433, opened 2019-03-02
build: fix uninitialized variable error on gcc 7.4.0

jb55:
`*best` is checked for null before the comparison against the uninitialized
variable ever happens, so this isn't a real issue.

Initialize it to zero so that we don't fail to compile on certain gcc versions.

    plugins/pay.c: In function ‘add_shadow_route’:
    plugins/pay.c:644:18: error: ‘sample’ may be used uninitialized in this function
      if (!best || v > sample) {
                    ~~^~~~~~~~


m-schmoock:
yup, should be initialized with zero

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2433/commits/ffc03d2bc84dc42f745959fbb6c8007cf0a6f701

PR #2436, opened 2019-03-03
change the version of bip32 private_key to BIP32_VER_MAIN_PRIVATE

trueptolemy:
In libwally-core, the version BIP32_VER_TEST_PRIVATE is for testnet/regtest, and BIP32_VER_MAIN_PRIVATE is for mainnet.
Though the version of bip32 key don't affect the key-generation now, BIP32_VER_MAIN_PRIVATE is more accurate.

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -360,7 +360,7 @@ static void populate_secretstuff(void)
 			    "bip32 seed", strlen("bip32 seed"));
 		salt++;
 	} while (bip32_key_from_seed(bip32_seed, sizeof(bip32_seed),
-				     BIP32_VER_TEST_PRIVATE,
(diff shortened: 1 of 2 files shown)


rustyrussell:
Good catch!  But in this case, we could be either.  We need to tell hsmd whether we're mainnet or not.  And of course I have no idea what BIP32 constants are used for non-bitcoin (and I don't care).

I suggest adding a 'testnet' arg to hsmd/hsm_wire.csv's hsm_init format, and using that?

trueptolemy:
@rustyrussell Sorry, I forgot we had both mainnet and testnet two choice. You gave me good suggestion. Thanks:)
I will reopen this PR after I fix it,

PR #2438, opened 2019-03-04
UTF-8 Passthough

rustyrussell:
I enhanced my summry plugin to use UTF-8 for drawing bars, but it gets mangled into \uxxxx.  Fixing this took all morning, and you won't like the final patch, but I'm pretty sick of Python right now...

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -378,9 +378,11 @@ class Plugin(object):
             self.log(traceback.format_exc())
 
     def _write_locked(self, obj):
-        s = json.dumps(obj, cls=LightningRpc.LightningJSONEncoder) + "\n\n"
+        # ensure_ascii turns UTF-8 into \uXXXX so we need to suppress that,
+        # then utf8 ourselves.
+        s = bytes(json.dumps(obj, cls=LightningRpc.LightningJSONEncoder, ensure_ascii=False) + "\n\n", encoding='utf-8')
         with self.write_lock:
-            self.stdout.write(s)
+            self.stdout.buffer.write(s)
             self.stdout.flush()
 
     def notify(self, method, params):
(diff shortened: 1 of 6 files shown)


contrib/pylightning/tests/test_version.py
```diff
@@ -0,0 +1,7 @@
+import lightning
+
+
+def test_version():
```
cdecker: We can actually ensure this programmatically: `setup.py` can run arbitrary code to detect stuff, so we could just import the `lightning` module and use `__version__` in the call to `setup()`

contrib/pylightning/lightning/plugin.py
```diff
@@ -378,9 +378,11 @@ def _dispatch_notification(self, request):
             self.log(traceback.format_exc())
 
     def _write_locked(self, obj):
-        s = json.dumps(obj, cls=LightningRpc.LightningJSONEncoder) + "\n\n"
+        # ensure_ascii turns UTF-8 into \uXXXX so we need to suppress that,
+        # then utf8 ourselves.
+        s = bytes(json.dumps(obj, cls=LightningRpc.LightningJSONEncoder, ensure_ascii=False) + "\n\n", encoding='utf-8')
```
cdecker: Good catch, I always get confused with `decode` and `encode` as well.

contrib/pylightning/tests/test_version.py
rustyrussell: OK, great!  Wasn't sure if that would work if the dependencies for lightning module aren't installed yet, but maybe that's not what setup.py is for.

rustyrussell:
Rebased to fold the fix, will merge post-travis.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2438/commits/cdc4c8b61c2249678553467264316a0094a8bca4

Issue #2442, opened 2019-03-04
tests: new DB migration hooks called with NULL pointer


m-schmoock:
### Issue and Steps to Reproduce
Using the new (smarter) DB migration update routine currently requires a early return in all hooks as certain unit tests call the hooks with a `*ld` NULL pointer.

### Workaround
```C
void migrate_hook(struct lightningd *ld, struct db *db)
{
       /* Note: Some unit tests call migrate without *ld context:
        * - wallet/test/run-wallet
        * - wallet/test/run-db
        * We need to return so we dont get nullpointer access violation. */
       if (!ld) return;

       // your update here
```

### Solution
I don't know how to solve this in a clean way

m-schmoock:
@rustyrussell 

For the `wallet/test/run-wallet.c` its easy to add the pointer, but not for the `wallet/test/run-db.c` and probably future test cases. We could change the `static void db_migrate(struct lightningd *ld, struct db *db, struct log *log)` in `wallet/db.c` to only call the migration hooks if *ln was set.

Not elegant, but maybe sufficient:

### Workaround:
Change this:
```C
		if (dbmigrations[current].func)
			dbmigrations[current].func(ld, db);
```

To this:
```C

		if (dbmigrations[current].func && ld)
			dbmigrations[current].func(ld, db);
```

### Benefits:
No need to early return on every migration hook.

### Drawbacks: 
A test that is not supplying *ld cannot rely on the migrations to be applied correctly.

cdecker:
Closed by #2444

PR #2444, opened 2019-03-04
fix: only call DB migration hook if *ld was defined

m-schmoock:
Will prevent calling a DB migrate hook without the `*ld` context.
Unit tests not giving `*ld` cant rely on correct migration.

Don't know if its a elegant solution, but does its job.

### DIFF
--- a/wallet/test/run-db.c
+++ b/wallet/test/run-db.c
@@ -51,14 +51,14 @@ static struct db *create_test_db(void)
 	return db;
 }
 
-static bool test_empty_db_migrate(void)
+static bool test_empty_db_migrate(struct lightningd *ld)
 {
 	struct db *db = create_test_db();
 	CHECK(db);
 	db_begin_transaction(db);
 	CHECK(db_get_version(db) == -1);
 	db_commit_transaction(db);
-	db_migrate(NULL, db, NULL);
+	db_migrate(ld, db, NULL);
 	db_begin_transaction(db);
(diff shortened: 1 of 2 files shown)


m-schmoock:
Fixed https://github.com/ElementsProject/lightning/issues/2442

m-schmoock:
@rustyrussell thanks, we should consider removing my one line hack, since we might want to see this bug again if someone misses to pass a valid `*ld` in a future tests.

Alternatively we can `assert(ld)` in the migrate iterate function so this fails fast. What do you think?

m-schmoock:
ACK 4854ba8

@rustyrussell just saw the one liner is not there anymore, so you can merge it unless you want an additional assert() in the migrate

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2444/commits/4854ba85b210b85c98d3c484b6be6db8e9b880e1

m-schmoock:
@cdecker @rustyrussell 
Unfortunately we seem to have a problem here. the `wallet/test/run-db.c` simply allocates and frees the space for `struct lightningd`. This leads to the situation where Valgrind complains when the space is being used with no further initialization, i.e. https://travis-ci.org/ElementsProject/lightning/jobs/502030002

For `wallet/test/run-db.c` test I suggest following test stub zero allocation code:
```C
struct lightningd *ld = talz(NULL, struct lightningd);
```

This will work with valgrind, and unless the DB migration hooks do not do more magic than reading config, this is feasible for unittests.

PR #2449, opened 2019-03-05
fix: fix a pr2444 introduced valgrind complaint

m-schmoock:
PR #2444 introduced a problem where valgrind complained ab usage of otherwise uninitialized space in the `*ld` unittest stub.

We will reuse the testnet config and put it in testuitls.c ...

### DIFF
--- a/wallet/test/run-db.c
+++ b/wallet/test/run-db.c
@@ -122,6 +122,7 @@ int main(void)
 	bool ok = true;
 	/* Dummy for migration hooks */
 	struct lightningd *ld = tal(NULL, struct lightningd);
+	ld->config = test_config;
 
 	ok &= test_empty_db_migrate(ld);
 	ok &= test_vars(ld);
--- a/wallet/test/test_utils.h
+++ b/wallet/test/test_utils.h
@@ -1,6 +1,8 @@
 #ifndef LIGHTNING_WALLET_TEST_TEST_UTILS_H
 #define LIGHTNING_WALLET_TEST_TEST_UTILS_H
 
+#include "lightningd/lightningd.h"
+
 /* Definitions "inspired" by libsecp256k1 */
 #define TEST_FAILURE(msg) do { \
     fprintf(stderr, "%s:%d: %s\n", __FILE__, __LINE__, msg); \
@@ -20,5 +22,7 @@
 } while(0)
 
 #define CHECK(cond) CHECK_MSG(cond,"test condition failed");
+
+const struct config test_config;
 
 #endif /* LIGHTNING_WALLET_TEST_TEST_UTILS_H */
--- a/wallet/test/run-wallet.c
+++ b/wallet/test/run-wallet.c
@@ -1161,6 +1161,7 @@ int main(void)
 	wally_init(0);
 	secp256k1_ctx = wally_get_secp_context();
 	ld = tal(tmpctx, struct lightningd);
+	ld->config = test_config;
 
 	/* Only elements in ld we should access */
 	list_head_init(&ld->peers);
--- a/wallet/test/test_utils.c
+++ b/wallet/test/test_utils.c
@@ -1,0 +1,25 @@
+#include "test_utils.h"
+
+/*
+ * Copy of lightnind/options.c testnet_config
+ * To reduce code duplication move testnet_config from options.c to options.h.
+ */
+const struct config test_config = {
+	.locktime_blocks = 6,
+	.locktime_max = 14 * 24 * 6,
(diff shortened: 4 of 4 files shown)


m-schmoock:
@cdecker @rustyrussell 

We could also imagine a test stub being creation function like this in `lightningd/options.c`
```C
void setup_test_stub(struct lightningd *ld)
{
       ld->config = testnet_config;
}
```

This could be initiated by `wallet/db.c` on the fly when `*ld` was NULL (implicit test mode), so neither test developer, nor migration developer need to know anything (except that test stubs have very limited function).

Note: I had troubles with the Makefiles and the linker settings, so instead of committing a PR, this is  just food thought...

rustyrussell:
No, you need to populate the fields accessed. We don't zero allocated memory, exactly so valgrind will warn us if we access fields we didn't expect

m-schmoock:
@cdecker @rustyrussell what about this 2e0e3064a590ffe9623014e7146587c179d00515 ?
Will move the default configs to header so they can be used by the test stub...

niftynei:
I think I'm missing a lot of the context for this, but it seems preferable to just instantiate or define a 'test' config with with dummy values that can be easily manipulated as part of the test infrastructure, rather than expose/reuse the prod/staging values in test.

wallet/test/run-db.c
```diff
@@ -122,6 +123,7 @@ int main(void)
 	bool ok = true;
 	/* Dummy for migration hooks */
 	struct lightningd *ld = tal(NULL, struct lightningd);
+	ld->config = testnet_config;
```
niftynei:
```suggestion
	ld->config = {
	.locktime_blocks = 6 * 24,
	.locktime_max = 14 * 24 * 6,
	...
	};
```

niftynei: example of inline config definition

m-schmoock: @niftynei Yes I know. I tried to reduce code duplication an need for changes at three locations (main, test and now also stub) if config values are added.

To the context: the new database migration routines allow a `*ld` pointer to have context sensitive migrations. They are needed i.e. for the setchannelfee PR. Currently they need to access at least a config section (maybe more for future migrations). They caused a valgrind complaint as the wallet db tests tried to migrate with an uninitialized config section.

I thought it would be good to have reusable test stubs, thats why I did this...

m-schmoock: Will push an update tomorrow...

niftynei: cool, sg. right, instead of inlining them, the more sensible/DRY'er approach would be to have a single 'canonical' test config that's defined and reused within the context of the tests, but not totally sure that we've got that type of dependency infrastructure defined for tests at the moment....

m-schmoock:
Okay, will include the testconfig inlining and future setup of ld* stub pointer in separate file than. Still 3 config definitions where it could be 2. In the end its not really important, we just need a solution to use the wallet tests + new migrations + valgrind

wallet/test/run-db.c
m-schmoock: addressed a8d5189f76384c00797cfb67934825d1dadec8bd

m-schmoock:
@niftynei  I updated based on your input....

wallet/test/test_utils.h
```diff
@@ -21,4 +21,7 @@
 
 #define CHECK(cond) CHECK_MSG(cond,"test condition failed");
 
+#include "lightningd/lightningd.h"
```
cdecker: This include likely needs to be moved to the top.

wallet/test/test_utils.h
```diff
@@ -21,4 +21,7 @@
...
+static const struct config test_config;
```
cdecker: Not sure what `static` is used for in this case. It'll work without it as well :-)

m-schmoock: addressed a22b11093bee9d7a9e0aec0a4703c8f9db116d66

The `options.c` testnet and mainnet configs are also static, maybe because of that, removed it anyway.

wallet/test/test_utils.h
m-schmoock: addressed af1c99e6d704b8ead29a9d802721b399f35c0838

m-schmoock:
@cdecker addressed your input. it is on top of master... at least it looks like that for me.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2449/commits/af1c99e6d704b8ead29a9d802721b399f35c0838

PR #2450, opened 2019-03-05
fix: missing closedir in plugin.c

m-schmoock:
As pointed out by `bitonic-cjp` in the IRC log, there was a closedir missing.

**Note:** I do not check the return code of this operation by intention, as there would not be any meaningful error correction or notification at this point.


rustyrussell:
Yay, I found this too, but you beat me to a PR!

PR #2451, opened 2019-03-06
Fix installing plugins in dockerfile

wailo:
**Current state**
In docker, the lightning installation is achieved by copying binaries one by one ‘manually’.  As soon as a new binary is added, the dockerfile become outdated. See #2377 

**What has been done**
Binaries are installed into a temporary directory using `configure --prefix=` and `make install`, then the installation directory is copied ‘as-it-is’ to the final image in the second phase of the build. This will guarantee that any new lib/binary/plugin generated from `make install` will be reflected in the docker images, and more important, the docker image configuration is closer to the normal build.

**Note to the maintainers**
Update the official docker image after the merge of this PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,3 @@
-ARG PREFIX=/tmp/lightning_install
-
 FROM alpine:3.7 as builder
 
 RUN apk add --no-cache \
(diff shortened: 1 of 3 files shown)


NicolasDorier:
Concept ACK. Will test it.

NicolasDorier:
`lightningd_1       | lightningd: SQLITE version mismatch: compiled 3025003, now 3021000`

NicolasDorier:
This might be because alpine updated package but not stretch?

If so maybe the solution would be to build from Alpine instead of from stretch. (And we can get ride of glibc dependency!) However, there might be good reasons why I did not do this already when I did this dockerfile.

NicolasDorier:
The error above seems unrelated to this PR.

NicolasDorier:
For me to test this PR we need to fix first the version mismatch issue.

NicolasDorier:
So the problem came from my docker cache being old. That said, if I do enable-static, I can fix it for all... resuming my test for this PR.

NicolasDorier:
Tested ACK.

wailo:
Thanks!

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2451/commits/24a5065337bab8b4294dfdc35b51a1a01ec223dc

PR #2453, opened 2019-03-06
Remove missing config option "port"

dlogemann:
During installation I noticed that the README suggests using a config option "port=9735". Unfortunately lightningd rejects this option with error "unrecognized option". As this is the default port anyway, I simple removed this option.


rustyrussell:
Great catch! Yes, this was removed in favor of much more comprehensive address configuration, and I clearly missed this one!

Issue #2409, opened 2019-02-27
Store the `announcement_signature` in the database to work around some impls not sending it on reconnect


cdecker:
Proposed by @SimonVrouwe [here](https://github.com/ElementsProject/lightning/issues/1778#issuecomment-467807053).

This should reduce the number of "Waiting for their announcement signature" channels dramatically.

trueptolemy:
If you don't hurry, I want to have a try.

cdecker:
Absolutely, go ahead. I can't add you as assignee until you accept my collaborator invitation, but consider yourself the owner of this issue :wink:

trueptolemy:
> Absolutely, go ahead. I can't add you as assignee until you accept my collaborator invitation, but consider yourself the owner of this issue

thank u :) !

jb55:
yikers, I haven't looked at my lightning node in awhile and saw this:

```
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Channel announced."
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[]
[
  "CHANNELD_NORMAL:Attempting to reconnect"
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[
  "CHANNELD_NORMAL:Attempting to reconnect"
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
  "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
]
[
  "CHANNELD_NORMAL:Attempting to reconnect"
]
[
  "CHANNELD_NORMAL:Reconnected, and reestablished.",
[...]

Issue #1778, opened 2018-07-30
Clarification: Waiting for their announcement signatures.  What does it actually mean?


CluelessFool:
This is more of an information request because I don't understand something.

Most of my channels have the state & status of:

```
          "state": "CHANNELD_NORMAL",
          "status": [
            "CHANNELD_NORMAL:Reconnected, and reestablished.",
            "CHANNELD_NORMAL:Funding transaction locked. Waiting for their announcement signatures."
          ],
```

What I am expecting to see is that the channel has been announced, something like:

```
          "state": "CHANNELD_NORMAL",
          "status": [
            "CHANNELD_NORMAL:Reconnected, and reestablished.",
            "CHANNELD_NORMAL:Funding transaction locked. Channel announced."
          ],
```

From other issues like  https://github.com/ElementsProject/lightning/issues/1174 I gather that the channel, while usable by my node, is unusable for routing others because, as it literally says, lightningd does not yet have the other nodes signatures yet.

I also see in a lot of other issues where people have posted a channel they have channels in the same state.

So, my questions are:

* What is the root cause of this?

* Does the message "state": "CHANNELD_NORMAL" mean they are operating correctly despite the status message?

* What if anything I should do about these channels?  

* If I want them to be used for routing should I close them because they are not functioning in that capacity?

As a general comment, having multiple status fiels is a bit confusing, it might be better to have a single status message that progresses all the way from "INITALIZING" all the way to "FULLY_OPERATIONAL".

### Issue and Steps to Reproduce

Open a channel, *most* channels in my experience end up with this message.

### `getinfo` output

```
{
  "id": "------------------------------------------------------------------------------------",
  "alias": "-----------",
  "color": "---------------------",
  "address": [
    {
      "type": "ipv4",
      "address": "----------------",
      "port": 9735
    }
  ],
  "binding": [
    {
      "type": "ipv4",
      "address": "----------------",
      "port": 9735
    }
  ],
  "version": "v0.6-203-g5230302",
  "blockheight": 534445,
  "network": "bitcoin"
}
```

P.S.  Thanks for all the brilliant stuff you guys are doing for us (for free) I for one really appreciate it.

cdecker:
[...]

PR #2466, opened 2019-03-12
Fix: Store remote channel announcement information into wallet DB

trueptolemy:
This PR draft tries to fix issue #2409.

We don't store the announcement signatures before, so when restart we may waste much time on waiting for remote announcement reply. So this PR tries to store the remote announcement signatures into DB like what @SimonVrouwe [proposed](https://github.com/ElementsProject/lightning/issues/1778#issuecomment-467807053) (we don't store local announcement signatures because we can drive sigs locally).

Instead of adding a new wire type between Channeld and MASTER, I change the `wire_channel_got_funding_locked` and ask it not only tell MASTER remote funding locked, but also send remote announcement sigs to MASTER and store sigs.

There are 4 changes:

**1. Rename `wire_channel_got_funding_locked` to `wire_channel_got_locked_and_announced`, and add announcement data into the message of this type.** 

Now the message will be like this:
`WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT` + `got_remote_announcement(bool)`  + `next_per_commit_point(pubkey)` + `remote_announcement_node_sigs(signature)` + `remote_announcement_bitcoin_sigs(signature)`
The bool type above means if the message is to tell MASTER announcement or not(instead, to tell MASTER remote funding locked).

There 2 situations that this message will be sent:
-  When Channeld receive remote peer funding_locked message, it will tell MASTER funding got locked remote. We will set the message like this:
   msg: `WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT` + `false` + `next_per_commit_point(pubkey)` + `NULL` + `NULL`
  In this case, the msg equals the original `wire_channel_got_funding_locked`.
- When Channeld receive remote peer announcement message, it will tell MASTER the announcement signatures. We will the set message like this:
   msg: `WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT` + `true`  + `NULL` + `remote_announcement_node_sigs` + `remote_announcement_bitcoin_sigs`
Channeld will check if we received the same announcement before and ignore the same announcement.
Channeld only send `wire_channel_got_locked_and_announced` for announcement when it receives announcement for the first time.

**2. Add announcement struct in channel struct (in lightningd/channeld.h).**
```
struct announcement {
	secp256k1_ecdsa_signature announcement_node_sigs;
	secp256k1_ecdsa_signature announcement_bitcoin_sigs;
};
```
  When MASTER receive announcement signatures, we will put signatures in this struct and store it into DB.

**3. Add announcement table in DB, and add wallet_announcement_save() and
   wallet_announcement_load() function.**
- `announcement table`:
The announcement table store data with the type like: `(channelid, remote_node_sigs, remote_bitcoin_sigs)`.
When we create a new channel and insert channel into DB, the announcement data will be set. If channel doesn't set ANNOUNCEMENT flag or hasn't received remote announcement message, we will set `(channelid, NULL, NULL)` for this channel.
- `wallet_announcement_save()`:
This function will only be called when MASTER receive `wire_channel_got_locked_and_announced` with flag setting true and resolve announcement signatures into announcement struct.
- `wallet_announcement_load()`:
This function will only be called when we initial a channel from DB (this will happen when restart).

**4. Add announcement save and load test in `wallet/test/run_wallet.c`.**

Compared to adding a new announcement wire type between Channeld and Lightningd, this way isn't so direct.
Anyway, it's just a try. What do you think about?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - JSON API: `newaddr` outputs `bech32` or `p2sh-segwit`, or both with new `all` parameter (#2390)
 - JSON API: `listpeers` status now shows how many confirmations until channel is open (#2405)
 - Config: Adds parameter `min-capacity-sat` to reject tiny channels.
+- Lightningd: Store channel_announcement information of remote peer into DB (#2409)
 
 ### Changed
 
--- a/channeld/channel_wire.csv
+++ b/channeld/channel_wire.csv
@@ -9,6 +9,8 @@ channel_init,,funding_txid,struct bitcoin_txid
 channel_init,,funding_txout,u16
 channel_init,,funding_satoshi,struct amount_sat
 channel_init,,minimum_depth,u32
+channel_init,,remote_announcement_node_sigs,?secp256k1_ecdsa_signature
+channel_init,,remote_announcement_bitcoin_sigs,?secp256k1_ecdsa_signature
 channel_init,,our_config,struct channel_config
 channel_init,,their_config,struct channel_config
 # FIXME: Fix generate-wire.py to allow NUM_SIDES*u32 here.
@@ -92,9 +94,13 @@ channel_fulfill_htlc,,fulfilled_htlc,struct fulfilled_htlc
 channel_fail_htlc,1006
 channel_fail_htlc,,failed_htlc,struct failed_htlc
 
-# When we receive funding_locked.
-channel_got_funding_locked,1019
-channel_got_funding_locked,,next_per_commit_point,struct pubkey
+# We will send it when we receive funding_locked and announcement message
+channel_got_locked_and_announced,1019
+# we set ture iff when we receive remote channel announcement
+channel_got_locked_and_announced,,got_remote_announcement,bool
+channel_got_locked_and_announced,,next_per_commit_point,?struct pubkey
+channel_got_locked_and_announced,,remote_announcement_node_sigs,?secp256k1_ecdsa_signature
+channel_got_locked_and_announced,,remote_announcement_bitcoin_sigs,?secp256k1_ecdsa_signature
 
 # When we send a commitment_signed message, tell master.
 channel_sending_commitsig,1020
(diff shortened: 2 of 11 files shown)


rustyrussell:
Note that this is stuck waiting for #2405 to be rebased?

trueptolemy:
@rustyrussell Thanks for your reply! This PR has nothing to do with the rebase of #2405 .
My main idea here (24eb994) is to modify` wire_channel_got_funding_locked`, and let this wire message not only tell MASTER remote funding locked, but also send remote announcement sigs to MASTER and store sigs:

> 1. Rename wire_channel_got_funding_locked to wire_channel_got_locked_and_announced, and add announcement data into the message of this type.
> 
> Now the message will be like this:
> WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT + got_remote_announcement(bool) + next_per_commit_point(pubkey) + remote_announcement_node_sigs(signature) + remote_announcement_bitcoin_sigs(signature)
> The bool type above means if the message is to tell MASTER announcement or not(instead, to tell MASTER remote funding locked).
> 
> There 2 situations that this message will be sent:
> 
> When Channeld receive remote peer funding_locked message, it will tell MASTER funding got locked remote. We will set the message like this:
> msg: WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT + false + next_per_commit_point(pubkey) + NULL + NULL
> In this case, the msg equals the original wire_channel_got_funding_locked.
> When Channeld receive remote peer announcement message, it will tell MASTER the announcement signatures. We will the set message like this:
> msg: WIRE_CHANNEL_GOT_LOCKED_AND_ANNOUNCEMENT + true + NULL + remote_announcement_node_sigs + remote_announcement_bitcoin_sigs
> Channeld will check if we received the same announcement before and ignore the same announcement.
> Channeld only send wire_channel_got_locked_and_announced for announcement when it receives announcement for the first time.

In fact, I'm not sure if this trick is appropriate?

cdecker:
I would have actually liked a new message rather than piggy-backing this onto an existing message. Having messages that serve multiple purposes is really weird, and likely better served in separate handlers.

It would have also been better to just add the two fields to the channels table since there is a 1-to-1 relationship to that.

trueptolemy:
@cdecker Thank you for your very useful advice!! I'll modify it these days!

PR #2469, opened 2019-03-14
Use tini as docker entrypoint

NicolasDorier:
All Tini does is spawn a single child (Tini is meant to be run in a container), and wait for it to exit all the while reaping zombies and performing signal forwarding. (See [github](https://github.com/krallin/tini))

This is especially useful as by default Bash running in PID 1 does not forward SIGTERM to child processes.
Luckliy, we use `exec` which mean the signal is directly reaching `lightningd`. However, if the PID is 1, the program need to explicitly handle SIGTERM. (which we did recently, and it works great)

However, lightning might not properly reap zombie processes.
Also, if `EXPOSE_TCP` is used, we are spawning `socat` and running lightningd at the same time without using `exec`, meaning that lightningd will not receive SIGTERM.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,7 +71,8 @@ RUN apk add --no-cache \
      inotify-tools \
      socat \
      bash \
-     zlib-dev
+     zlib-dev \
+     tini
 
 ENV GLIBC_VERSION 2.27-r0
 ENV GLIBC_SHA256 938bceae3b83c53e7fa9cc4135ce45e04aae99256c5e74cf186c794b97473bc7
@@ -99,4 +100,4 @@ COPY --from=builder /opt/litecoin/bin /usr/bin
 COPY tools/docker-entrypoint.sh entrypoint.sh
 
 EXPOSE 9735 9835
-ENTRYPOINT  [ "./entrypoint.sh" ]
+ENTRYPOINT  [ "/sbin/tini", "-g", "--", "./entrypoint.sh" ]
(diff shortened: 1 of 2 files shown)


NicolasDorier:
Ping @wailo (since you use those dockerfiles as well)

leshik:
@NicolasDorier why not to use `init: true` in `docker-compose.yml` instead? It uses `tini` under the hood AFAIK and works even in `swarm` mode in recent Docker versions.

wailo:
@NicolasDorier Thanks for sharing this. Is there a difference in behaviour that I can observe with and without running the container with tiny? I will be interested test.

NicolasDorier:
@leshik yes, `init:` on docker-compose is still not widely deployed on 3.0 of docker-compose (It was on versus 2.2 of docker-compose files, but got removed on 3.0 because of swarm, then added again)

So right now, it is better to just not rely on it at all.

@wailo so on my side, I have a fork of this repo, and I do a loop in entrypoint until I can see if a file has been created. ([Here](https://github.com/btcpayserver/lightning/blob/btcpaymaster/tools/docker-entrypoint.sh#L75)) without init, if you do `docker-compose down -t 1000`  you will see docker-compose hanging for 1000s as it wait the container for shutting down with SIGTERM.
If you use `tini`, the container shut down immediately as bash properly handle SIGTERM. (not PID 1)

You can test is with this repo by using clightnign with `EXPOSE_TCP` to expose c-lightning as a TCP socket. If you do `docker-compose down -t 1000`, it will hang. But not with `tini`.

PR #2470, opened 2019-03-14
setchannelfee for 'all' channels

m-schmoock:
based ontop of PR #2342 

Will give an option to use the `setchannelfee` command to update `all` channels at once.
Adds feature, doc and test.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1492,12 +1492,12 @@ static const struct json_command setchannelfee_command = {
 	"setchannelfee",
 	json_setchannelfee,
 	"Sets specific routing fees for channel with {id} "
-	"(either peer ID, channel ID or short channel ID). "
-	"The routing fees are defined by a fixed {base} (msat) "
-	"and a {ppm} value (proportional per millionth)."
-	"To disable channel specific fees and restore global values, "
-	"just leave out the optional {base} and {ppm} values. "
-	"{base} can also be supplied in other units, i.e. '1sat'."
+	"(either peer ID, channel ID, short channel ID or 'all'). "
+	"Routing fees are defined by a fixed {base} (msat) "
+	"and a {ppm} (proportional per millionth) value. "
+	"If values for {base} or {ppm} are left out, defaults will be used. "
+	"{base} can also be defined in other units, for example '1sat'. "
+	"If {id} is 'all', the fees will be applied for all channels. "
 };
 AUTODATA(json_command, &setchannelfee_command);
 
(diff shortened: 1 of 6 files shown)


m-schmoock:
**Note:** We should merge https://github.com/ElementsProject/lightning/pull/2471 before this one

lightningd/peer_control.c
```diff
@@ -1390,6 +1409,32 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 		   NULL))
 		return command_param_failed();
 
+	// If the users requested 'all' channels we need to iterate
```
rustyrussell: Ok, I think the better way to do this is inside a new boutique (ie static local) param_peerid_or_all: this would do the id lookup, and return pointer to peer, or NULL for all.

In general, using param_tok is an antipattern: it means that when someone runs check on this command, it accepts anything.

lightningd/peer_control.c
```diff
@@ -1390,6 +1409,32 @@ static struct command_result *json_setchannelfee(struct command *cmd,
...
+		}
+		/* write response */
+		response = json_stream_success(cmd);
+		json_object_start(response, NULL);
+		json_add_num(response, "base", *base);
+		json_add_num(response, "ppm", *ppm);
+		json_add_string(response, "id", "all");
+		json_add_num(response, "count", count);
+		json_object_end(response);
+		return command_success(cmd, response);
```
rustyrussell: Should this return an array of results, one for each peer we changed? That implies the normal single-peer version should return an array too (we don't change form of output based on input parameters).

We can do this because API are not stable until release...

m-schmoock: Thanks, will do

m-schmoock: addressed by 287a0ed

lightningd/peer_control.c
m-schmoock: Okay, good idea to get rid of the iffy stuff in the command handler.
I am still working on this one...

m-schmoock: Okay, also addressed 
85449b2

lightningd/peer_control.c
```diff
@@ -1367,78 +1404,89 @@ static struct command_result *param_msat_u32(struct command *cmd,
 	return NULL;
 }
 
+void set_channel_fees(
```
rustyrussell: I think this should be static?  Plus, minor style variation: we don't usually end line  on '('

m-schmoock: thx, addressed 
96be8d6

m-schmoock:
@rustyrussell well that was an easy fix ;)

Issue #2473, opened 2019-03-15
lnd->lightning open channel failed if amount <57201 satoshi


arowser:
### Issue and Steps to Reproduce

lnd->lightning open channel failed if amount <57201 satoshi,  LND show:
```
[lncli] rpc error: code = Code(189) desc = You gave bad parameters:Our channel reserve 30000 would be below their dust 54600
```

if fund amount <57201 sat , the lnd->lightningd openchannel will failed, but lightningd->lnd will success.

The reason is  when open channel,   if local reserve < dust_limit,  clightning will set local_reserve = dust_limit,  local reserve= (state->funding.satoshis + 99) / 100,
but clightning dust limit is 546, but lnd is 573:
https://github.com/ElementsProject/lightning/blob/master/openingd/openingd.c#L306

It could be reproduced with https://github.com/arowser/lightning-integration/tree/dust_limit
```
TEST_DEBUG=1 py.test -v test.py  -s -k 'test_lnd_clightning_dust_open_channel[lnd_lightning]'
```

arowser:
Ref  https://github.com/lightningnetwork/lnd/issues/1142

rustyrussell:
AI cleanup of old issues.  If wrong, please open a new issue.


Triage decision summary:
Issue from 2019 about channel opening failures between lnd and Core Lightning due to dust limit calculation differences (546 vs 573 satoshi). The issue is old with last activity in 2025 but no recent updates, suggesting it may have been resolved in subsequent versions or the incompatibility persists without active development.

PR #2475, opened 2019-03-15
add dust_limit config options

arowser:
Add support for set dust_limit from command or cnfigure file.

And its a workround for lnd -> lightning open channel fail when funding amount <57201 sat, see issue #2473


lightningd/options.c
```diff
@@ -528,6 +548,8 @@ static const struct config testnet_config = {
 	.max_fee_multiplier = 10,
 
 	.use_dns = true,
+
+	.dust_limit = { 546 },
```
m-schmoock: Just noticed, you missed to add the `dust_limit`  value to the wallet teststub config in `wallet/test/test_utils.c`.
It compiles without it, but Valgrind will be unhappy when a DB migration will be using this field in the future...

m-schmoock: @niftynei    See? It will be missed   ;)

arowser:
@m-schmoock  add dust_limit to test_utils.c, missed that, I check out from a old master.
I can't find a easy way to add test dust_limit config now, I'll add a testcase to lightning-integration after this pr merged,  and the issue must triger from lnd https://github.com/arowser/lightning-integration/tree/dust_limit

And lnd keep the value to 573 for support litecoin,  the developer known that, see https://github.com/lightningnetwork/lnd/issues/1142

cdecker:
I don't quite get why the dust_limit needs to be configurable. Shouldn't this just be a simple change in `chainparams.c` to reflect the dust limit that is proper for that chain? Dust limits don't change all that often, and giving users more rope to hang themselves seems the wrong thing to do here.

arowser:
Its make sense, I think this should be fixed in lnd side, its should set
different dust_limit for litecoin and bitcoin.

PR #2477, opened 2019-03-15
Compatible posix sh

arowser:
make check-source error in my ubuntu 18.04

```
In tools/repro-build.sh line 33:
elif type lsb_release >/dev/null 2>&1; then
     ^-- SC2039: In POSIX sh, 'type' is undefined.
```

change `type` to `which` follow  https://wiki.ubuntu.com/DashAsBinSh#type suggestion.

### DIFF
--- a/tools/repro-build.sh
+++ b/tools/repro-build.sh
@@ -30,7 +30,7 @@ if [ -f /etc/os-release ]; then
     . /etc/os-release
     OS=$NAME
     VER=$VERSION_ID
-elif type lsb_release >/dev/null 2>&1; then
+elif command -v lsb_release >/dev/null 2>&1; then
     # linuxbase.org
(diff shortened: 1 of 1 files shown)


cdecker:
Currently fails with this:

```
In tools/repro-build.sh line 33:
elif which lsb_release >/dev/null 2>&1; then
     ^---^ SC2230: which is non-standard. Use builtin 'command -v' instead.
For more information:
  https://www.shellcheck.net/wiki/SC2230 -- which is non-standard. Use builti...
Makefile:318: recipe for target 'check-shellcheck' failed
make: *** [check-shellcheck] Error 123
```

arowser:
Wired, lets me try `command -v` .

cdecker:
Seems `test_setchannelfee_restart` is a bit flaky.

PR #2480, opened 2019-03-18
pytest: fix flaky test_setchannel_fee_restart test.

rustyrussell:
gossipd in l1 might not have registered l2 reconnecting, thus considering
the channel local-disabled.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2480/commits/a628469523302a0925bff7359fd5eef091a8703b

cdecker:
@m-schmoock that is likely a slow configuration due to `VALGRIND=1`. When running in Travis we also split the tests into groups to run more jobs in parallel. `test_setchannel_fee_restart` happens to be in group 6 in this case, hence that's the only configuration with `valgrind` that actually runs that particular test.

PR #2494, opened 2019-03-24
fix: openingd memory leaks on failures

m-schmoock:
Add a testcase that tries to open tiny channels. Currently this raises unpredictive Valgrind issues.

1. Try to establish a channel that is exactly 1 Satoshi to small when accounting for 2 times dust_limit as reservers. This raises a exeption correctly and the test asserts this correctly.
2. Try to establish a channel that should be exactly the minimal amount +-0 Satoshi. This test also raises a exception `openingd/openingd.c +668` where I do think the the `goto fail` handler performs correctly as it leads to raise the valgrind issues. Also I am not sure if the exception should be possible at all, because a code comment quotes that this is a "should not happen exception" (that actually does happen).

### DIFF
--- a/openingd/openingd.c
+++ b/openingd/openingd.c
@@ -442,8 +442,8 @@ static u8 *funder_channel(struct state *state,
 	struct pubkey *changekey;
 	struct bitcoin_signature sig;
 	u32 minimum_depth;
-	const u8 *wscript;
-	struct bitcoin_tx *funding;
+	struct bitcoin_tx *funding = 0;
+	const u8 *wscript = 0;
 	struct amount_msat local_msat;
 
 	/*~ For symmetry, we calculate our own reserve even though lightningd
@@ -825,6 +825,10 @@ static u8 *funder_channel(struct state *state,
 fail:
 	if (taken(utxos))
 		tal_free(utxos);
+	if (funding)
+		tal_free(funding);
+	if (wscript)
+		tal_free(wscript);
 	return NULL;
 }
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -128,6 +128,35 @@ def test_bad_opening(node_factory):
     l2.daemon.wait_for_log('to_self_delay 100 larger than 99')
 
 
+def test_opening_tiny_channel(node_factory):
+    # Try to establish channels with smaller than 1000000 msat (default)
+    #
+    # [l1] ---> [l2]
+    #
+    # BOLT2
+    # The receiving node MAY fail the channel if:
+    #  - funding_satoshis is too small
+    #
+    # Expects: opening is rejected
+    l1 = node_factory.get_node()
+    l2 = node_factory.get_node()
+
+    l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
+
+    min_capacity_sat = 1000
+    dustlimit_sat = 546
+    reserves = 2 * dustlimit_sat
+
+    # Open channel with less than 1000 sats (default) should be rejected
+    with pytest.raises(RpcError, match=r'channel capacity is 999sat, which is below 1000000msat'):
+        l1.fund_channel(l2, min_capacity_sat + reserves - 1)
+
+    # When the capacity is precisely met for the fundee
+    # NOTE: The next line will (currently) raise an exception
+    with pytest.raises(RpcError, match=r'Could not meet their fees and reserve'):
+        l1.fund_channel(l2, min_capacity_sat + reserves)
+
+
 def test_second_channel(node_factory):
     l1 = node_factory.get_node()
     l2 = node_factory.get_node()


m-schmoock:
I have the following memleaks in the logs of the failed test:

```
 peer_out WIRE_OPEN_CHANNEL'                                                    
 peer_in WIRE_ACCEPT_CHANNEL'                                                   
 Funder cannot afford fee on initial commitment transaction'                    
 peer_out WIRE_ERROR'                                                           
 aborted opening negotiation: Could not happen meet OUR fees and reserve'       
 UPDATE WIRE_OPENING_FUNDER_FAILED'                                             
K_REPLY with 0 fds'                                                             
 MEMLEAK: 0x55e0c1c7dde8'                                                       
   label=bitcoin/tx.c:334:struct bitcoin_tx'                                    
   backtrace:'                                                                  
     ccan/ccan/tal/tal.c:437 (tal_alloc_)'                                      
     bitcoin/tx.c:334 (bitcoin_tx)'                                             
     common/utxo.c:56 (tx_spending_utxos)'                                      
     common/funding_tx.c:26 (funding_tx)'                                       
     openingd/openingd.c:620 (funder_channel)'                                  
     openingd/openingd.c:1316 (handle_master_in)'                               
     openingd/openingd.c:1429 (main)'                                           
   parents:'                                                                    
     openingd/openingd.c:1351:struct state'                                     
 MEMLEAK: 0x55e0c1c7ebf8'                                                       
   label=bitcoin/script.c:166:u8[]'                                             
   backtrace:'                                                                  
     ccan/ccan/tal/tal.c:437 (tal_alloc_)'                                      
     ccan/ccan/tal/tal.c:466 (tal_alloc_arr_)'                                  
     bitcoin/script.c:166 (bitcoin_redeem_2of2)'                                
     common/initial_channel.c:87 (initial_channel_tx)'                          
     openingd/openingd.c:668 (funder_channel)'                                  
     openingd/openingd.c:1316 (handle_master_in)'                               
     openingd/openingd.c:1429 (main)'                                           
   parents:'                                                                    
     openingd/openingd.c:1351:struct state'
```

m-schmoock:
Travis seems to have the same issue with memleaks:
https://travis-ci.org/ElementsProject/lightning/jobs/510722257#L1250

m-schmoock:
@cdecker for the stated memoryleaks, it looks to me like the `tal_t *ctx`  (used as `struct state *state`) is not correctly initialized. This happens somewhere in `openingd/openingd.c` ->  `static u8 *funder_channel(struct state *state, ...)`. Why that only happens in my specific test case is not know to me (yet).

m-schmoock:
@cdecker I think the memory leak is fixed by 6b3b34f which was 'simply' a missing tal_free of `*state` in the `fail` handler. It does not seem to break any tests on my machine, lets wait what Travis thinks about it...

openingd/openingd.c
```diff
@@ -825,6 +825,7 @@ static u8 *funder_channel(struct state *state,
 fail:
 	if (taken(utxos))
 		tal_free(utxos);
+	tal_free(state);
```
cdecker: I'm not sure this is a good idea: `state` is allocated/owned by `main()`, where it'll also be freed on exit. If we `tal_free` it here we create a double-free bug. In addition this is specular to `fundee_channel` which does not free `state`. Maybe it's better to just mark `state` as a root element using `notleak`.

openingd/openingd.c
```diff
@@ -671,8 +671,7 @@ static u8 *funder_channel(struct state *state,
 	if (!tx) {
 		/* This should not happen: we should never create channels we
 		 * can't afford the fees for after reserve. */
-		negotiation_failed(state, true,
-				   "Could not meet their fees and reserve");
+		negotiation_failed(state, true, "Could not meet our fees and reserve");
```
cdecker: I'm confused, the comment above the creation of `tx` indicates that this is "their" commitment, which would indicate that they can't meet the agreed upon fees (fees are non-owned) and they can't meet the reserve we impose on them. Or is the comment above also switched around?

m-schmoock: @cdecker okay, first thanks for your attention here.

Thing is the code behaves unpredictably unstable. Sadly its not my machine alone as we can see at the Travis output. Sometimes I run in Valgrind issues, then I recompile and they disappear without code changes, both times clean build. :(

1. I will remove the non working commit, so you can see the same Valgrind issues on Travis again that I linked in IRC yesterday  1079737 fix: openingd memory leak *state on failures    
(so this PR will again be 'please have a look at my broken test')

2. I did the renaming of the exception, because in my testcase `test_opening_tiny_channel` this was raised by `[l1]` which is the founder of the channel and not `[l2]`. Maybe I misinterpreted the code, but I thought the message relates to the situation that the founders own fees (l1) cannot be met and not the responders (l2).

> /* we should never create channels we can't afford the fees for after reserve.

I left this comment untouched, it tells me that there is something wrong with the founders commitment tx (meaning not **theirs**)

m-schmoock:
As we can see following Travis Valgrind complains are raise with unmodified `master` sourcecode just running my added testcase...

- https://travis-ci.org/ElementsProject/lightning/jobs/511489072#L1252   
- https://travis-ci.org/ElementsProject/lightning/jobs/511489071#L1252
- https://travis-ci.org/ElementsProject/lightning/jobs/511489066#L1255

```
 aborted opening negotiation: Could not happen meet OUR fees and reserve'       
 UPDATE WIRE_OPENING_FUNDER_FAILED'                                             
K_REPLY with 0 fds'                                                             
 MEMLEAK: 0x55e0c1c7dde8'                                                       
   label=bitcoin/tx.c:334:struct bitcoin_tx'                                    
   backtrace:'                                                                  
     ccan/ccan/tal/tal.c:437 (tal_alloc_)'                                      
     bitcoin/tx.c:334 (bitcoin_tx)'                                             
     common/utxo.c:56 (tx_spending_utxos)'                                      
     common/funding_tx.c:26 (funding_tx)'                                       
     openingd/openingd.c:620 (funder_channel)'                                  
     openingd/openingd.c:1316 (handle_master_in)'                               
     openingd/openingd.c:1429 (main)'                                           
   parents:'                                                                    
     openingd/openingd.c:1351:struct state'                                     
 MEMLEAK: 0x55e0c1c7ebf8'                                                       
   label=bitcoin/script.c:166:u8[]'                                             
   backtrace:'                                                                  
     ccan/ccan/tal/tal.c:437 (tal_alloc_)'                                      
     ccan/ccan/tal/tal.c:466 (tal_alloc_arr_)'                                  
     bitcoin/script.c:166 (bitcoin_redeem_2of2)'                                
     common/initial_channel.c:87 (initial_channel_tx)'                          
     openingd/openingd.c:668 (funder_channel)'                                  
     openingd/openingd.c:1316 (handle_master_in)'                               
     openingd/openingd.c:1429 (main)'                                           
   parents:'                                                                    
     openingd/openingd.c:1351:struct state'
```

cdecker:
Ok, let's quickly recap the exception that travis sees:

```
: MEMLEAK: 0x1b684e8
:   label=bitcoin/tx.c:334:struct bitcoin_tx
:   backtrace:
:     ccan/ccan/tal/tal.c:437 (tal_alloc_)
:     bitcoin/tx.c:334 (bitcoin_tx)
:     common/utxo.c:56 (tx_spending_utxos)
:     common/funding_tx.c:26 (funding_tx)
:     openingd/openingd.c:620 (funder_channel)
:     openingd/openingd.c:1316 (handle_master_in)
:     openingd/openingd.c:1429 (main)
:   parents:
:     openingd/openingd.c:1351:struct state
: MEMLEAK: 0x1b69e28
:   label=bitcoin/script.c:166:u8[]
:   backtrace:
:     ccan/ccan/tal/tal.c:437 (tal_alloc_)
:     ccan/ccan/tal/tal.c:466 (tal_alloc_arr_)
:     bitcoin/script.c:166 (bitcoin_redeem_2of2)
:     common/initial_channel.c:87 (initial_channel_tx)
:     openingd/openingd.c:668 (funder_channel)
:     openingd/openingd.c:1316 (handle_master_in)
:     openingd/openingd.c:1429 (main)
:   parents:
:     openingd/openingd.c:1351:struct state
```

(I cropped it to exclude the prefix, makes it easier to read).

There are two potential leaks here: one array of `u8` allocated here:

https://github.com/ElementsProject/lightning/blob/b17b05b8ae271d097d39ff24840d17f1f47e88be/bitcoin/script.c#L166

and one `bitcoin_tx` allocated here:

https://github.com/ElementsProject/lightning/blob/b17b05b8ae271d097d39ff24840d17f1f47e88be/bitcoin/tx.c#L334

In both cases the `state` is the parent, and `tal_free` on `state` will cascade to free both leaks, but that's the sledgehammer approach :-)

Let's see for the first leak of the `u8[]` which turns out to be `wscript` in `funder_channel()`. That should likely be freed once we exit from `funder_channel()`, otherwise there is a period between calling `funder_channel` and `tal_free` in `main()` where we don't have a reference to `wscript` which is detected correctly as a leak (though a temporary one).

On to the second leak, the `bitcoin_tx`, this has the same issue as above, so calling `tal_free` on `tx` in `funder_channel` will resolve this. We cannot just `tal_free(state)` here because we already free it in `main()` which would then be a double free (and `valgrind` would start complaining afaik).

Also notice that this memleak tracking is implemented in `tal` itself, and doesn't require `valgrind`, but `valgrind` will increase the likelihood of triggering this issue since it has a 5x - 20x runtime overhead and increasing chances of `state` still being un-freed  when the memleak call comes in from the main daemon.

m-schmoock:
@cdecker gravis seems happy now...

Remaining Question if the tested 'should not happen' exception should happen or be handled otherwise...

tests/test_connection.py
```diff
@@ -128,6 +128,35 @@ def test_bad_opening(node_factory):
...
+    min_capacity_sat = 1000
+    dustlimit_sat = 546
+    reserves = 2 * dustlimit_sat
+
+    # Open channel with less than 1000 sats (default) should be rejected
+    with pytest.raises(RpcError, match=r'channel capacity is 999sat, which is below 1000000msat'):
+        l1.fund_channel(l2, min_capacity_sat + reserves - 1)
+
+    # When the capacity is precisely met for the fundee
+    # NOTE: The next line will (currently) raise an exception
```
niftynei: what needs to change to make this pass?

niftynei: typically if you've got a test that you want to make go from 'not passing' to 'passing' in a commit set, you can annotate the test with `@pytest.mark.xfail(strict=True)` e.g. https://github.com/ElementsProject/lightning/pull/2490/commits/a4ec45f5d6a5e43ba5dc2d5d47c45f54dc27657d#diff-af265d267530ebfb09faeaf242e16936R1036

m-schmoock: @niftynei Well, I'm not sure, as you can see in the changeset of the complete PR #2506 , the actual minimum funding amount for the funder is around 6000sat. Smaller channels will run into these funder side caused related exception.

If someone with deeper knowledge can tell me why, I can document the 'testcase' in a better helpful way.

niftynei: i'm pretty sure this is due to not providing enough to cover the funding tx fee. see https://github.com/ElementsProject/lightning/blob/master/common/initial_commit_tx.c#L62 and also

m-schmoock: @niftynei Thanks, I will resolve that now in the other PR #2506 so I do not have two diverging test cases... I will also close this PR as the memory leak is also included in the other...

PR #2497, opened 2019-03-25
Fix for reorgs changing short_channel_id after lockin

SimonVrouwe:
Before this PR we calculate `short_channel_id` only once, when the funding tx reaches `minimum_depth` confirmations, the channel will then _lockin_ and can be used locally. For private channels we then stop watching the funding depth.

When `minimum_depth=1`, a chain reorganization of 1 block deep can change funding tx height and/or txindex after lockin. The probability for this to happen is small,  just a fraction of the _orphan rate_ (for bitcoin [0.5%](https://medium.com/@jb395official/orphan-blocks-june-16-2018-a8f4799dcc2c)) and I couldn't find any case of this ever happening.

But if we don't agree with our peer about `scid` a public channel will fail at announce depth or `channel_updates` are not recognized properly, making the channel partially unusable (not routable). If peers agree on the _incorrect_ `scid`, the `channel_announcement` is ignored by other nodes, effectively keeping the channel private.
 
This PR addresses this by always keep watching the funding depth and keep updating `scid` until `ANNOUNCE_MIN_DEPTH`(=6). If `scid` changes, we kill `channeld` so it can restart with the updated `scid`, which adds it to the routing table. This leaves the channel (at least locally) usable with both `scid`s. If the remote node (impl.) still doesn't see the reorg'd `scid` at announce depth, the (public) channel will probably still fail. It is a bit tricky but, after some experimentation, I think this is the most straight forward solution.

A pytest method `simple_reorg` is added to create chain reorgs. ~Three~ Two pytests are included to test different scenarios. Not sure if all these are needed for such unlikely scenarios, please advice. Other commits include small fixes to tighten-up some tests and to give (at least me) insight in the blockchain syncing machinery (`try_extend_tip`) and when it encounters a stale block.

TODO:

- [x] remove commented-out debug code
- [x] test  `short_channel_id` and `depth_togo` (depends on #2405)
~- [ ] move `test_funding_reorgs` to proper place or include/mix it in another test~
- [x] remove (too much) verbosity and keep it concise

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -848,21 +848,22 @@ static enum watch_result funding_depth_cb(struct lightningd *ld,
 					   unsigned int depth)
 {
 	const char *txidstr;
+	struct short_channel_id scid;
 
 	txidstr = type_to_string(tmpctx, struct bitcoin_txid, txid);
 	log_debug(channel->log, "Funding tx %s depth %u of %u",
 		  txidstr, depth, channel->minimum_depth);
 	tal_free(txidstr);
 
-	bool local_locked = depth >= channel->minimum_depth;
+	bool min_depth_reached = depth >= channel->minimum_depth;
 
-	/* If we restart, we could already have peer->scid from database */
-	if (local_locked && !channel->scid) {
+	/* Reorg can change scid, so always update/save scid when possible (depth=0
+	 * means the stale block with our funding tx was removed) */
+	if ((min_depth_reached && !channel->scid) || (depth && channel->scid)) {
 		struct txlocator *loc;
 
 		loc = wallet_transaction_locate(tmpctx, ld->wallet, txid);
-		channel->scid = tal(channel, struct short_channel_id);
-		if (!mk_short_channel_id(channel->scid,
+		if (!mk_short_channel_id(&scid,
 					 loc->blkheight, loc->index,
 					 channel->funding_outnum)) {
 			channel_fail_permanent(channel, "Invalid funding scid %u:%u:%u",
@@ -871,30 +872,34 @@ static enum watch_result funding_depth_cb(struct lightningd *ld,
 			return DELETE_WATCH;
 		}
 
-		/* We've added scid, update */
-		wallet_channel_save(ld->wallet, channel);
+		/* If we restart, we could already have peer->scid from database */
+		if (!channel->scid) {
+			channel->scid = tal(channel, struct short_channel_id);
+			*channel->scid = scid;
+			wallet_channel_save(ld->wallet, channel);
+
+		} else if (!short_channel_id_eq(channel->scid, &scid)) {
+			/* This normally restarts channeld, initialized with updated scid
+			 * and also adds it (at least our halve_chan) to rtable. */
+			channel_fail_transient(channel,
+					        "short_channel_id changed to %s (was %s)",
+					        short_channel_id_to_str(tmpctx, &scid),
+					        short_channel_id_to_str(tmpctx, channel->scid));
+
+			*channel->scid = scid;
+			wallet_channel_save(ld->wallet, channel);
+			return KEEP_WATCHING;
+		}
 	}
 
 	/* Try to tell subdaemon */
 	if (!channel_tell_depth(ld, channel, txid, depth))
 		return KEEP_WATCHING;
 
-	if (!local_locked)
+	if (!min_depth_reached)
 		return KEEP_WATCHING;
-	/* BOLT #7:
-	 *
-	 * A node:
-	 *   - if the `open_channel` message has the `announce_channel` bit set AND a `shutdown` message has not been sent:
-	 *     - MUST send the `announcement_signatures` message.
-	 *       - MUST NOT send `announcement_signatures` messages until `funding_locked`
-	 *       has been sent and received AND the funding transaction has at least six confirmations.
-	 *   - otherwise:
-	 *     - MUST NOT send the `announcement_signatures` message.
-	 */
-	if (!(channel->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL))
-		return DELETE_WATCH;
 
-	/* We keep telling it depth until we get to announce depth. */
+	/* We keep telling it depth/scid until we get to announce depth. */
 	if (depth < ANNOUNCE_MIN_DEPTH)
 		return KEEP_WATCHING;
 
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -868,7 +868,7 @@ def test_blockchaintrack(node_factory, bitcoind):
     """Check that we track the blockchain correctly across reorgs
     """
     l1 = node_factory.get_node(random_hsm=True)
-    addr = l1.rpc.newaddr()['bech32']
+    addr = l1.rpc.newaddr(addresstype='all')['p2sh-segwit']
 
     ######################################################################
     # First failure scenario: rollback on startup doesn't work,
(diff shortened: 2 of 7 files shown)


tests/utils.py
```diff
@@ -322,6 +322,49 @@ def generate_block(self, numblocks=1):
...
 
+    def simple_reorg(self, height, shift=0):
+        """
+        Reorganize chain by creating a fork at height=[height] and re-mine all mempool
+        transactions into [height + shift], where shift >= 0. Returns hashes of generated
+        blocks.
+
+        Note that tx's that become invalid at [height] (because coin maturity, locktime
+        etc.) are removed from mempool. The length of the new chain will be one longer
+        then original OR original length + [shift], whichever is larger.
```
cdecker: `then` -> `than`

tests/test_misc.py
```diff
@@ -872,27 +872,24 @@ def test_funding_reorg_between_restart(node_factory, bitcoind):
     q1 = l1.db_query('SELECT * FROM transactions WHERE hex(id)="{}"'.format(txid_bo))
     assert q1[0]['blockheight'] == 106
 
-    # stop node before bitcoind restart to prevent memleak error
     l1.stop()
-
-#     assert hashes[3] == bitcoind.rpc.getblockhash(106)
```
cdecker: This commit needs to be squashed with the previous one.

SimonVrouwe:
Squashed last two commits and simplified the test a little by removing the 2nd reorg and instead just look at the other node's log (which was running all the time).

Not sure if all these added log lines is too much verbosity (first three), but IMHO they clarify what is happening.

The test does not yet include a real corner case where `announce=false` and `minimum-confirms=1`, because then it fails. I will probably create a separate PR with a fix proposal.

SimonVrouwe:
Sorry, accidentally pushed my (premature) fix onto this branch, undone that.

SimonVrouwe:
Removed the 3th pytest (moved it to draft #2623), which was only used to learn more about the meaning of `scid` in routes and route-hints.

What I understand so far:
- a route only works when the routing node has the route `scid` in its `rtable`
- channel's state (_active_) in `rtable` is just an indication, a running `channeld` means it is _active_

Please consider this PR for review.
BTW I don't understand why Travis CI fails, failures seem to be unrelated to changes from this PR.

lightningd/peer_control.c
```diff
@@ -871,30 +870,30 @@ static enum watch_result funding_depth_cb(struct lightningd *ld,
...
+		else if (!short_channel_id_eq(channel->scid, &scid)) {
+			/* Will restart channeld, initialized with updated scid, which
+			 * should add it (at least our halve of the channel) to rtable. */
+			channel_fail_transient(channel, "short_channel_id changed to %s (was %s)",
+					        short_channel_id_to_str(tmpctx, &scid),
+					        short_channel_id_to_str(tmpctx, channel->scid));
+		}
+
 		/* We've added scid, update */
+		*channel->scid = scid;
```
trueptolemy: I guess you want to ask Channeld to die after setting `*channel->scid = scid;` and saving into DB?
what about return immediately in` else if` block?
```
if (!channel->scid) {
	channel->scid = tal(channel, struct short_channel_id);
	*channel->scid = scid;
	wallet_channel_save(ld->wallet, channel);
} else if (!short_channel_id_eq(channel->scid, &scid)) {
	/* Will restart channeld, initialized with updated scid, which
	 * should add it (at least our halve of the channel) to rtable. */
	log_info(channel->log, "short_channel_id change %s -> %s",
			        short_channel_id_to_str(tmpctx, &scid),
			        short_channel_id_to_str(tmpctx, channel->scid));
	*channel->scid = scid;
	channel_fail_transient(channel, "Different scid, and reorg may happen. Restart channeld.");
	wallet_channel_save(ld->wallet, channel);
	return KEEP_WATCHING;
}
```

lightningd/peer_control.c
```diff
@@ -848,21 +848,20 @@ static enum watch_result funding_depth_cb(struct lightningd *ld,
...
 		  txidstr, depth, channel->minimum_depth);
 	tal_free(txidstr);
 
-	bool local_locked = depth >= channel->minimum_depth;
-
-	/* If we restart, we could already have peer->scid from database */
-	if (local_locked && !channel->scid) {
+	/* Reorg can change scid, so update scid when possible (depth=0 means
+	 * the stale block with our funding tx was removed) */
+	if (depth >= channel->minimum_depth || (channel->scid && depth)) {
```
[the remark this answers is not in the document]
SimonVrouwe: Ok, but to be honest I find the term `local_locked` a little misleading here (in `peer_control`), because it can be confused with `peer->funding_locked[LOCAL]` in `channeld`.

How about:
rename `local_locked` to `min_depth_reached`
and
`if ((min_depth_reached && !channel->scid) || (depth && channel->scid))`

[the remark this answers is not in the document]
SimonVrouwe: > Comparing with restarting Channeld, what do you think of failing the channel directly when we monitor the reorg?

One of the benefits of this PR is we don't need to fail the channel if both sides see the reorg. A public channel will fail anyway when either side doesn't see the reorg when reaching announce_depth (because then `scid`s don't match).

lightningd/peer_control.c
SimonVrouwe: Yes maybe that is more straight-forward (instead of having to go through `channel_tell_depth` first). Will fix that.

lightningd/peer_control.c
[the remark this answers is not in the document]
trueptolemy: Thank you, now I understand :)
And your name is good!

lightningd/peer_control.c
SimonVrouwe: With respect to `log_info`, I try to keep it concise and a `channel_fail_transient` is already logged at `log-level=info`.

SimonVrouwe:
> in light of the recent broader re-org conversations

Well I wasn't aware of any, was it on the lightning-dev mailing list?

trueptolemy:
> Well I wasn't aware of any, was it on the lightning-dev mailing list?
@SimonVrouwe  Well, in fact it's on twitter, the CEO of Binance had plans to re-org with some pools to retrieve coins stolen by hackers.

SimonVrouwe:
I know the conditions addresses by this PR border the ridiculous and the worst case scenario seems just: "failure of a _public_ channel at announce depth".

So maybe the added complexity/tests is not worth adding to the code base. For me this was more of an an educational exercise.

jb55:
Concept ACK

tests/test_misc.py
```diff
@@ -903,9 +903,50 @@ def test_blockchaintrack(node_factory, bitcoind):
...
+    wait_for(lambda: only_one(l1.rpc.listpeers()['peers'][0]['channels'])['status']
+             == ['CHANNELD_AWAITING_LOCKIN:Funding needs 1 more confirmations for lockin.'])
+    bitcoind.generate_block(1)                      # height 107
+    l1.wait_channel_active('106x1x0')
+    l1.stop()
+
+    # Create a fork that changes short_channel_id from 106x1x0 to 108x1x0
+    bitcoind.simple_reorg(106, 2)                   # heights 106-108
+    bitcoind.generate_block(1)                      # height 109 (to reach minimum_depth=2 again)
+    l1.daemon.rpcproxy = bitcoind.get_proxy()       # otherwise complains `address already in use`
```
cdecker: Recently addressed by #2675 :wink:

SimonVrouwe: Thanks, will add it to my fixup list.

PR #2501, opened 2019-03-26
TLV parsing

niftynei:
Have the wire generators be TLV savvy. Adds parsers for nested structs etc.

~Addition of a new TLV type requires adding a new file `gen_{tlv_name}_csv` to the wires. The parsers will identify TLV fields from the `_csv` file, and then find the appropriate additional file to parse.~

Also generates the structs necessary to figure out whether or not these fields have been included. In general, you can check for a TLV type's inclusion via whether or not the returned tlv struct's message field is null.

A few notes:
  - the size of any TLV individual message is ~capped at u8~ a varint.
  - we fail if an even numbered TLV type message is not understood by the parser


wire/gen_peer_wire_csv
```diff
@@ -63,6 +63,28 @@ funding_signed,32,signature,64
 funding_locked,36
 funding_locked,0,channel_id,32
 funding_locked,32,next_per_commitment_point,33
+open_channel2,56
```
rustyrussell: This is OK for now, but I think we're going to have to move to a system where we have experimental CSV files (which get compiled in if EXPERIMENTAL defined).

tools/generate-wire.py
```diff
@@ -126,6 +126,16 @@ def __init__(self, message, name, size, comments, prevname):
         self.lenvar = None
         self.num_elems = 1
         self.optional = False
+        self.is_tlv = False
+
+        # field name appended with '+' means this field contains a tlv
+        if name.endswith('+'):
```
rustyrussell: This isn't going to work well in the spec; let's use name.endswith('tlv')?

tools/generate-wire.py
```diff
@@ -384,7 +399,72 @@ def print_fromwire_array(self, ctx, subcalls, basetype, f, name, num_elems):
                 subcalls.append('fromwire_{}(&cursor, &plen, {} + i);'
                                 .format(basetype, name))
 
-    def print_fromwire(self, is_header):
+    def print_tlv_fromwire(self, tlv_name):
+        """ prints fromwire function definition for a TLV message.
+        these are significantly different in that they take in a struct
+        to populate, instead of fields, as well as a length to read in
+        """
```
rustyrussell: This is an interesting choice, and I'm not sure it's the correct one (we changed from structs to individual parameters early on in our wire parsing).

We could parse TLVs just like we do a formatted field, except everything has to be a ptr-to-ptr so we can set it to NULL if it doesn't appear.

tools/generate-wire.py
```diff
@@ -832,28 +832,28 @@ def print_struct(self):
 
 tlv__type_impl_towire_fields = """\tif ({tlv_name}->{name}) {{
 \t\ttlv_msg = tal_arr(ctx, u8, 0);
-\t\ttowire_u16(p, {enum});
+\t\ttowire_u8(p, {enum});
```
rustyrussell: Except I was wrong and it (and type) should be varint :)

tools/generate-wire.py
```diff
@@ -873,7 +874,7 @@ def print_struct(self):
 \t\t\t*plen -= msg_len;
 \t\t}}
 \t}}
-\tif (!*p) {{
+\tif (!*p || start_len - *plen != *len) {{
```
rustyrussell: It's not entirely clear to me that this is correct.  For onion, at least, we should on first type=0 and ignore any remaining.  Seems consistent to do that elsewhere, too.

wire/gen_peer_wire_csv
niftynei: I went back and forth on whether or not to include these. Really they're only here because I needed something to test with :cactus:

tools/generate-wire.py
niftynei: updated!!

tools/generate-wire.py
niftynei: updated!!

tools/generate-wire.py
niftynei: hmm you're right. there's also the 'ordering' aspect of the proposed [TLV spec](https://github.com/lightningnetwork/lightning-rfc/pull/572) that this implementation is missing at the moment. will fix in a second PR!

tools/generate-wire.py
niftynei: right, i think either of these is a valid way to do this. i had the impression that most TLV messages would be rather involved field wise, but it's also possible they're mostly one-off fields...

niftynei:
@rustyrussell fixed up to exclude any and all references to 'dual funding' :)

if you're happy with this, i think we're good to merge for now. here's a short list of things that need to be improved in a follow up PR:
- 0x0 type stops parsing
- ordering of messages as per spec
- onion types need to be zero padded (?)
- the print_towire/fromwire's for the devtool/decodemsg utility need to be amended

rustyrussell:
> @rustyrussell fixed up to exclude any and all references to 'dual funding' :)
> 
> if you're happy with this, i think we're good to merge for now. here's a short list of things that need to be improved in a follow up PR:
> 
>     * 0x0 type stops parsing

I think at the moment it will break parsing, being considered even :)

> 
>     * ordering of messages as per spec

I think this Just Works if the spec lists them in ascending order   But probably good to check!

>     * onion types need to be zero padded (?)

Naah, I think what will happen is we'll memcpy it in, so remainder will be zero.

>     * the print_towire/fromwire's for the devtool/decodemsg utility need to be amended

Def. nice to have.

wire/fromwire.c
```diff
@@ -101,6 +101,27 @@ bool fromwire_bool(const u8 **cursor, size_t *max)
 	return ret;
 }
 
+u64 fromwire_var_int(const u8 **cursor, size_t *max)
+{
+	if (*max < 1) {
+		fromwire_fail(cursor, max);
+		return 0;
+	}
```
rustyrussell: Note that this test is redundant!

Issue #2483, opened 2019-03-18
There should be a hook for accepting channel opens


rustyrussell:
In particular, people want to reject tiny channels.

m-schmoock:
@rustyrussell I just read the IRC logs of the last meeting... 
> 20:42 < sstone> yes. I think pretty soon people will start refusing tiny incoming channels, and be more aggressive with unreliable nodes
> ...
> 20:58 < rusty> Yeah, I had someone complain c-lgihtning didn't implement minimum channel amount for accepting channels.  I'd never thought of it, TBH.

Only plugin hook or also config value? If the latter, what would be a reasonable default for a config value?

m-schmoock:
I started working on this, will add config option for auto reject and a hook for plugins... First PR will follow soon

m-schmoock:
@cdecker @rustyrussell 
I will submit two separate PRs one for the hook and one for a config value...

m-schmoock:
**Update:** After starting working on this, I will set the `min_effective_htlc_capacity` via config. Default will then be `1000sat` which was the old internal magic number.

m-schmoock:
This is the working branch for the config part: https://github.com/m-schmoock/lightning/tree/issue-2483-config-value

I think discovered some memory leak and remote crash problems which I need to track down first...

m-schmoock:
@cdecker @rustyrussell For the plugin hook, is it feasible to implement the hook synchronously?

cdecker:
Config value was merged :+1:

PR #2506, opened 2019-03-26
Issue #2483 part1: min_capacity_sat config for rejecting tiny channels

m-schmoock:
**SUMMARY**
In regards to the IRC meeting logs, as the first part of Issue #2483 I see the need to introduce a config value and switch to automatically reject channel opening requests for tiny channels. The second part of adding a channel accept plugin hook will be in a separate PR that I am already working on. This PR will:

- Remove magic value `min_effective_htlc_capacity = AMOUNT_MSAT(1000000)`
- Introduce `min_capacity_sat` config value and `--min-capacity-sat` config switch.
- Add testcase
- Add doc
- Fix memory leaks discussed in PR #2494.

### DIFF
--- a/common/amount.c
+++ b/common/amount.c
@@ -370,6 +370,16 @@ bool amount_msat_to_u32(struct amount_msat msat, u32 *millisatoshis)
 	return true;
 }
 
+void amount_msat_from_u64(struct amount_msat *msat, u64 millisatoshis)
+{
+	msat->millisatoshis = millisatoshis;
+}
+
+void amount_msat_from_sat_u64(struct amount_msat *msat, u64 satoshis)
+{
+	msat->millisatoshis = satoshis * 1000;
+}
+
 bool amount_msat_fee(struct amount_msat *fee,
 		     struct amount_msat amt,
 		     u32 fee_base_msat,
--- a/common/amount.h
+++ b/common/amount.h
@@ -102,6 +102,10 @@ bool amount_msat_less_eq_sat(struct amount_msat msat, struct amount_sat sat);
 WARN_UNUSED_RESULT bool amount_msat_to_u32(struct amount_msat msat,
 					   u32 *millisatoshis);
 
+/* Programatically initialize from various types */
+void amount_msat_from_u64(struct amount_msat *msat, u64 millisatoshis);
+void amount_msat_from_sat_u64(struct amount_msat *msat, u64 satoshis);
+
 /* Common operation: what is the HTLC fee for given feerate?  Can overflow! */
 WARN_UNUSED_RESULT bool amount_msat_fee(struct amount_msat *fee,
 					struct amount_msat amt,
--- a/lightningd/lightningd.h
+++ b/lightningd/lightningd.h
@@ -67,6 +67,9 @@ struct config {
 
 	/* Are we allowed to use DNS lookup for peers. */
 	bool use_dns;
+
+	/* Minimal amount of effective funding_satoshis for accepting channels */
+	u64 min_capacity_sat;
 };
 
 struct lightningd {
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -662,8 +662,10 @@ static void channel_config(struct lightningd *ld,
 {
 	/* FIXME: depend on feerate. */
 	*max_to_self_delay = ld->config.locktime_max;
-	/* This is 1c at $1000/BTC */
-	*min_effective_htlc_capacity = AMOUNT_MSAT(1000000);
+
+	/* Take minimal effective capacity from config min_capacity_sat */
+	amount_msat_from_sat_u64(min_effective_htlc_capacity,
+		ld->config.min_capacity_sat);
 
 	/* BOLT #2:
 	 *
--- a/openingd/openingd.c
+++ b/openingd/openingd.c
@@ -442,8 +442,8 @@ static u8 *funder_channel(struct state *state,
 	struct pubkey *changekey;
 	struct bitcoin_signature sig;
 	u32 minimum_depth;
-	const u8 *wscript;
 	struct bitcoin_tx *funding;
+	const u8 *wscript;
 	struct amount_msat local_msat;
 
 	/*~ For symmetry, we calculate our own reserve even though lightningd
@@ -674,7 +674,7 @@ static u8 *funder_channel(struct state *state,
 		 * can't afford the fees for after reserve. */
 		negotiation_failed(state, true,
 				   "Could not meet their fees and reserve");
-		goto fail;
+		goto fail_2;
 	}
 
 	/* We ask the HSM to sign their commitment transaction for us: it knows
@@ -724,7 +724,7 @@ static u8 *funder_channel(struct state *state,
 	 * transaction. */
 	msg = opening_negotiate_msg(tmpctx, state, true);
 	if (!msg)
-		goto fail;
+		goto fail_2;
 
 	sig.sighash_type = SIGHASH_ALL;
 	if (!fromwire_funding_signed(msg, &id_in, &sig.s))
@@ -779,7 +779,7 @@ static u8 *funder_channel(struct state *state,
 	if (!tx) {
 		negotiation_failed(state, true,
 				   "Could not meet our fees and reserve");
-		goto fail;
+		goto fail_2;
 	}
 
 	if (!check_tx_sig(tx, 0, NULL, wscript, &their_funding_pubkey, &sig)) {
@@ -823,6 +823,9 @@ static u8 *funder_channel(struct state *state,
 					   state->feerate_per_kw,
 					   state->localconf.channel_reserve);
 
+fail_2:
+	tal_free(wscript);
+	tal_free(funding);
 fail:
 	if (taken(utxos))
 		tal_free(utxos);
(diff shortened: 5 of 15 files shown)


m-schmoock:
**Note:** Just noticed that I uploaded an unresolved TODO

openingd/openingd.c
```diff
@@ -442,8 +442,8 @@ static u8 *funder_channel(struct state *state,
 	struct pubkey *changekey;
 	struct bitcoin_signature sig;
 	u32 minimum_depth;
-	const u8 *wscript;
-	struct bitcoin_tx *funding;
+	struct bitcoin_tx *funding = 0;
+	const u8 *wscript = 0;
```
cdecker: We generally don't use dummy initializations since that disables `valgrind`'s check of uninitialized fields. It's better to have a separate `goto` fail target that we jump to after these fields are initialized so that the `tal_free` below is only invoked once the are initialized.

common/amount.h
```diff
@@ -102,6 +102,9 @@ bool amount_msat_less_eq_sat(struct amount_msat msat, struct amount_sat sat);
 WARN_UNUSED_RESULT bool amount_msat_to_u32(struct amount_msat msat,
 					   u32 *millisatoshis);
 
+/* Programatically initialize an msat from an u64. */
+void amount_msat_from_u64(struct amount_msat *msat, u64 millisatoshis);
```
cdecker: If I'm not mistaken the previous commit will fail to compile since it depends on these definitions. Please swap their order so we keep the history bisectable (each commit should compile and pass tests on its own).

m-schmoock: This is just a Github GUI glitch after I `git rebase -i` the commit order. As you can see on my local branch https://github.com/m-schmoock/lightning/commits/issue-2483-config-value the order is in fact correct.

```
* 5bb6756e (HEAD -> issue-2483-config-value, mschmoock/issue-2483-config-value) doc: commandline switch min-capacity-sat
* e7b650a1 test: opening_tiny_channel with min_capacity_sat
* a6df7938 feat: add min_capacity_sat config value and switch
* be0e5c97 common: add programatic initialization of msat by u64
* e1e10cce fix: openingd memory leaks when failing
```

Can you confirm this?

openingd/openingd.c
m-schmoock: > Doesn't the default value of 1000sat trigger this failure anyway?
Not sure if I understand this question...

I will fix the dummy initialization with the next force push

common/amount.h
m-schmoock: @cdecker  There is this GitHub bug: https://help.github.com/en/articles/why-are-my-commits-in-the-wrong-order

m-schmoock: I will do this, so this can be closed: https://stackoverflow.com/a/44705243/884474

m-schmoock:
> Doesn't the default value of 1000sat trigger this failure anyway?
> 
> [lightning/openingd/openingd.c](https://github.com/ElementsProject/lightning/blob/426b22fdcb360aa6ee8264991c316d858acf89f6/openingd/openingd.c#L963-L984)
> 
> Lines 963 to 984 in [426b22f](/ElementsProject/lightning/commit/426b22fdcb360aa6ee8264991c316d858acf89f6)
> 
>  if (amount_sat_greater(state->remoteconf.dust_limit, 
>  		       state->localconf.channel_reserve)) { 
>  	negotiation_failed(state, false, 
>  			   "Our channel reserve %s" 
>  			   " would be below their dust %s", 
>  			   type_to_string(tmpctx, struct amount_sat, 
>  					  &state->localconf.channel_reserve), 
>  			   type_to_string(tmpctx, struct amount_sat, 
>  					  &state->remoteconf.dust_limit)); 
>  	return NULL; 
>  } 
>  if (amount_sat_greater(state->localconf.dust_limit, 
>  		       state->remoteconf.channel_reserve)) { 
>  	negotiation_failed(state, false, 
>  			   "Our dust limit %s" 
>  			   " would be above their reserve %s", 
>  			   type_to_string(tmpctx, struct amount_sat, 
>  					  &state->localconf.dust_limit), 
>  			   type_to_string(tmpctx, struct amount_sat, 
>  					  &state->remoteconf.channel_reserve)); 
>  	return NULL; 
>  }

m-schmoock:
@cdecker I updated the PR, I didnt understand this question of you:

> Doesn't the default value of 1000sat trigger this failure anyway?

Can you explain more detailed?

lightningd/opening_control.c
```diff
@@ -660,8 +660,10 @@ static void channel_config(struct lightningd *ld,
 {
 	/* FIXME: depend on feerate. */
 	*max_to_self_delay = ld->config.locktime_max;
-	/* This is 1c at $1000/BTC */
-	*min_effective_htlc_capacity = AMOUNT_MSAT(1000000);
+
+	/* Take minimal effective capacity from config min_capacity_sat */
+	u64 min_capacity_msat = ld->config.min_capacity_sat * 1000;
```
niftynei: As a goal, we're moving away from having `* 1000` inline for sat/msat transformations. A better way to handle this would be to have a helper, i.e. `amount_msat_from_satval(...)`

tests/test_connection.py
```diff
@@ -128,6 +128,69 @@ def test_bad_opening(node_factory):
...
+    l1.rpc.connect(l3.info['id'], 'localhost', l3.port)
+    l1.rpc.connect(l4.info['id'], 'localhost', l4.port)
+
+    # Open channel with less than default 1000 sats should be rejected
+    with pytest.raises(RpcError, match=r'channel capacity is 999sat, which is below 1000000msat'):
+        l1.fund_channel(l2, l2_min_capacity + reserves - 1)
+    # Open a channel with exactly the minimal amount for the fundee.
+    # NOTE 1: This will raise an exception at l1, as the funder cannot afford the own fees for the initial_commit_tx.
+    # NOTE 2: m-schmoock thinks the current exception text for this case is misleading :)
+    #         How about the internal: "Funder cannot afford fee on initial commitment transaction"
```
niftynei: +1 to fixing this message.

lightningd/opening_control.c
m-schmoock: added `amount_msat_from_sat_u32`  for this purpose

tests/test_connection.py
m-schmoock: @niftynei what do you think of this 3b78165 solution?

As there are three different abort `return NULL` code paths below `initial_channel_tx` and  `initial_commit_tx`. A good way would be to pass a char** _output pointer so the code below can set the correct reasons to be used in the exception passed to the user.

Now the test will raise this exception: `Could not meet their fees and reserve: Funder cannot afford fee on initial commitment transaction`. Which means the funder is not able to cover the commitment tx fees on their side. If I understand it correct, this is because the own side commit tx fees are actually deducted from the wallet separately?

Note: The non-modifed master code raises this ambigious exception at: https://github.com/ElementsProject/lightning/blob/901bd7c1c30aacaa80e39706fa9a65fd8adf13e3/openingd/openingd.c#L674

m-schmoock:
@niftynei 
> the leading `_` is unusual for c-lightning code style. better to just say `err_reason`

Okay, I was used to use this style on output pointers. I corrected to 'err_reason' on the latest force push 5bf9be1b

For the wording of the other exceptions, I am not sure if we should change them (make them more simple) in this PR since they are not my code.

If you still want to have your new wording in this PR, please just tell me quickly and I do another update.

m-schmoock:
@niftynei  Or should I forward a different wording to `err_reason` that that is internally used and logged?

common/amount.c
```diff
@@ -370,6 +370,16 @@ bool amount_msat_to_u32(struct amount_msat msat, u32 *millisatoshis)
 	return true;
 }
 
+void amount_msat_from_u64(struct amount_msat *msat, u64 millisatoshis)
+{
+	msat->millisatoshis = millisatoshis;
+}
+
+void amount_msat_from_sat_u32(struct amount_msat *msat, u32 satoshis)
```
cdecker: Any particular reason you chose `u32` here? I think in most cases we use `u64` for satoshis as well.

lightningd/lightningd.h
```diff
@@ -67,6 +67,9 @@ struct config {
 
 	/* Are we allowed to use DNS lookup for peers. */
 	bool use_dns;
+
+	/* Minimal amount of effective funding_satoshis for accepting channels */
+	u32 min_capacity_sat;
```
cdecker: Shouldn't this be `u64` as well?

openingd/openingd.c
```diff
@@ -822,6 +822,11 @@ static u8 *funder_channel(struct state *state,
 					   state->feerate_per_kw,
 					   state->localconf.channel_reserve);
 
+fail_2:
```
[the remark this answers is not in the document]
m-schmoock: Yes, the `if`s are not necessary within `fail_2`.

- Either you initialize with 0/NULL and can do an `if(ptr)`  within just one `fail` handler, because you wont get valgrind `jump or move depends on uninitialized value` errors for the `if(ptr)` access. 
- Or you have two `fail_2` gotos, that 'know' at which point the pointers have to be initialized. In this case we can skip the `if(ptr)`s before the `free(ptr)`.

I will push the next update to remove the unnecessary `if(ptr)` checks in fail_2.
If you prefer just one `fail` handler, we need to initialize with NULL for valgrind mem access check.

common/amount.c
m-schmoock: Because more than 4BTC as a min-value for channels seems crazy, but who am I to judge :D

... Changed it to 64bit.

PR #2508, opened 2019-03-27
allow plugin opts in config

rsbondi:
This change should allow for plugin options to be included in the config file, currently it crashes, this patch ignores options not known by lightningd on the "early" pass, allowing plugins to be registered with options and use them on the second pass.  It will still crash on unknown options that are not part of lightningd known options or plugin registered options.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2508/commits/4805b16c5658c0d98603ff420f07110566bad091

PR #2511, opened 2019-03-29
Subtype Parsing

niftynei:
Adds parsing capabilities for 'subtypes' in csv structs.


tools/generate-wire.py
```diff
@@ -1077,17 +1089,20 @@ def parse_tlv_file(tlv_field_name):
         continue
 
     if len(parts) == 2:
-        # eg commit_sig,132
-        messages.append(Message(parts[0], Enumtype("WIRE_" + parts[0].upper(), parts[1]), comments))
+        if parts[1] == '$':  # this is a subtype
+            subtypes.append(Subtype(parts[0], comments))
+        else:
```
rustyrussell: $?  Can we just have a single entry be a subtype?

Note that there's a companion to these changes, which is that the spec itself needs to decide on a format, and lightning-rfc/tools/extract-formats.py needs to be taught to extract it into the form you want here.

wire/gen_peer_wire_csv
```diff
@@ -102,6 +102,25 @@ accept_channel2,204,htlc_basepoint,33
 accept_channel2,237,first_per_commitment_point,33
 accept_channel2,270,opening_tlv_len,2
 accept_channel2,272,opening_tlv+,opening_tlv_len
+funding_compose,58
+funding_compose,0,temporary_channel_id,32
+funding_compose,32,num_inputs,2
+funding_compose,34,input_info,$num_inputs
+funding_compose,34+$,num_outputs,2
+funding_compose,36+$,output_info,$num_outputs
```
rustyrussell: Maybe insist that subtypes are defined *before* use, and then you can just recognize the  typename here, no need for $?

niftynei: That makes the RFC's a bit harder to read, imo

tools/generate-wire.py
niftynei: updated the RFC parser to output desired format here: https://github.com/lightningnetwork/lightning-rfc/pull/597

niftynei:
> We should see if we can use it on various internal structures; all those places where we open-code fromwire_ and towire_. Eg. fromwire_route_info, fromwire_channel_config, fromwire_crypto_state, fromwire_basepoints, fromwire_added_htlc, etc. Some of them should be amenable I think?

definitely! most of this piggy-backs on that infrastructure for building those already. i seem to remember one of them have some custom parsing logic, but otherwise that'd be a quick fix up.

wire/gen_peer_wire_csv
niftynei: thinking about this more, we could swap out the `$` with the type name in place, but to truly accurately represent the 'size' we need to know both the count field (if it exists) + the subtype

tools/generate-wire.py
rustyrussell: OK, I moved my complaint to that PR :)

Otherwise, excellent.  Be nice to try this out on some internal types, too: autogen FTW.

wire/gen_peer_wire_csv
niftynei: this has been updated!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2511/commits/4c27a051a7328c897c9ea5fbdca58e73b9c0f30f

PR #2513, opened 2019-03-29
gossip: Cache txout query failures

### LINKED ISSUE #2503: Cache the `txout` replies in `gossipd` for failed lookups
When receiving a `channel_announcement` that we don't have a corresponding channel, `gossipd` will always reach out to `lightningd` and ask it to resolve the blockhash, the txid and check the UTXO set in `bitcoind`. A positive result (when the outpoint is still unspent) will end up adding the channel to the local view and future announcements for the channel will be skipped.

However, if the UTXO entry results in a negative result, i.e., the channel does not exist or was closed, we will forget everything about that channel right away, and a new announcement will cause us to query `lightningd` and `bitcoind` again. Since some implementations seem not to be actively monitoring the blockchain for channel closes, this ends up causing us a lot of failed lookups whenever we connect to one such node.

Caching negative results in an `intmap` inside of `gossipd` and querying that cache before reaching out to `lightningd` would cut down considerably on those failed lookups which are rather expensive.

cdecker:
If we asked `bitcoind` for a txout and it failed we were not storing that
information anywhere, meaning that when we see the channel announcement the
next time we'd be reaching out to `lightningd` and `bitcoind` again, just to
see it fail again. This adds an in-memory cache for these failures so we can
just ignore these the next time around.

Fixes #2503

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -94,6 +94,7 @@ struct routing_state *new_routing_state(const tal_t *ctx,
 	rstate->local_channel_announced = false;
 	list_head_init(&rstate->pending_cannouncement);
 	uintmap_init(&rstate->chanmap);
+	uintmap_init(&rstate->txout_failures);
 
 	rstate->pending_node_map = tal(ctx, struct pending_node_map);
 	pending_node_map_init(rstate->pending_node_map);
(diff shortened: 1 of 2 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2513/commits/7ba16fe644318d2922dc8c88f0f0a1a72a2f35d2

niftynei:
what's the lifetime of entries in the cache? should we worry them growing unbounded?

m-schmoock:
Is gossip spam with invalid tx ids announcements technically possible? Or is there another protection in place somewhere in the code or protocol? If there is no protection, this might be an issue, that might be solved by a peer/IP ban and not a limitation in this map.

If we limit this map, the function of the map in the current simple version can be 'disarmed' by flooding invalid channel announcements by an attacker to increase load on the bitcoind RPC interface.

cdecker:
> Is gossip spam with invalid tx ids announcements technically possible? Or is there another protection in place somewhere in the code or protocol? If there is no protection, this might be an issue, that might be solved by a peer/IP ban and not a limitation in this map.

It is currently possible, yes. With this PR we're trading off a very high impact operation (several calls to `bitcoind` and piping all that info through `lightningd`) for a bit more memory consumption. We should eventually add a DoS prevention mechanism, but it's still early :wink:

m-schmoock:
@cdecker Thanks for the clarification, maybe we should open an Issue for:
- DoS attack vector analysis
- Gossip DoS protection
- Temp ban / Blacklisting

cdecker:
> @cdecker Thanks for the clarification, maybe we should open an Issue for:

Maybe a Wiki page is better for such a thing? After all it should serve as a reference and is not that well suited for a discussion format.

PR #2528, opened 2019-04-03
Add the time we received and resolved a forwarding to `listforwards`

### LINKED ISSUE #2491: Feature Request: Include timestamp in `channel_htlcs` or `forwarded_payments` table
As far as I can tell, there's no way to know when your node forwarded a payment, aside from parsing the log data.

It would be very helpful for routing node operators to be able to know exactly when they have forwarded payments, so they could have a better understanding of how changing various aspects of their setup affects their routing traffic.

Currently, `forwarded_payments` contains the in and out HTLC IDs only. By joining them to `channel_htlcs`, we are able to get `cltv_expiry`, which doesn't tell us much about when we actually forwarded the payment. The only other sense of time is HTLC ID, which is still far from ideal.

It would be nice if we had an absolute timestamp, for example in `forwarded_payments` when we receive the forward request, or in `channel_htlcs` when we know the payment has either failed/succeeded.

As an example, it would be great to be able to create a simple time series plot showing cumulative fees through time. This would allow us to see the effects of changing fees, opening/closing channels, etc.

cdecker:
This adds a `received_time` timestamp to incoming HTLCs and adds both a `received_time` and resolved_time` timestamp to the channel `forwarded_payments`. This allows us to see how long an HTLC was pending before it was resolved (with a granularity of seconds).

If desired we could probably change the granularity to milliseconds, but I wanted to try this first since #2491 only talks about an absolute time.

Fixes #2491

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ endif
 
 ifeq ($(COMPAT),1)
 # We support compatibility with pre-0.6.
-COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1
+COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1 -DCOMPAT_V070=1
 endif
 
 # Timeout shortly before the 600 second travis silence timeout
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -1902,6 +1902,19 @@ static void listforwardings_add_forwardings(struct json_stream *response, struct
 				     cur->fee,
 				     "fee", "fee_msat");
 		json_add_string(response, "status", forward_status_name(cur->status));
+#ifdef COMPAT_V070
+		/* If a forwarding doesn't have received_time it was created
+		 * before we added the tracking, do not include it here. */
+		if (cur->received_time.ts.tv_sec) {
+			json_add_timeabs(response, "received_time", cur->received_time);
+			if (cur->resolved_time)
+				json_add_timeabs(response, "resolved_time", *cur->resolved_time);
+		}
+#else
+		json_add_timeabs(response, "received_time", cur->received_time);
+		if (cur->resolved_time)
+			json_add_timeabs(response, "resolved_time", *cur->resolved_time);
+#endif
 		json_object_end(response);
 	}
 	json_array_end(response);
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2547,7 +2547,9 @@ const struct forwarding *wallet_forwarded_payments_get(struct wallet *w,
 			  ", out_msatoshi"
 			  ", hin.payment_hash as payment_hash"
 			  ", in_channel_scid"
-			  ", out_channel_scid "
+			  ", out_channel_scid"
+			  ", f.received_time"
+			  ", f.resolved_time "
 			  "FROM forwarded_payments f "
 			  "LEFT JOIN channel_htlcs hin ON (f.in_htlc_id == hin.id)");
 
@@ -2575,6 +2577,15 @@ const struct forwarding *wallet_forwarded_payments_get(struct wallet *w,
 
 		cur->channel_in.u64 = sqlite3_column_int64(stmt, 4);
 		cur->channel_out.u64 = sqlite3_column_int64(stmt, 5);
+
+		cur->received_time = sqlite3_column_timeabs(stmt, 6);
+		if (sqlite3_column_type(stmt, 7) != SQLITE_NULL) {
+			cur->resolved_time = tal(ctx, struct timeabs);
+			*cur->resolved_time = sqlite3_column_timeabs(stmt, 7);
+		} else {
+			cur->resolved_time = NULL;
+		}
+
 	}
 
 	return results;
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2492,7 +2492,10 @@ void wallet_forwarded_payment_add(struct wallet *w, const struct htlc_in *in,
 		", out_channel_scid"
 		", in_msatoshi"
 		", out_msatoshi"
-		", state) VALUES (?, ?, ?, ?, ?, ?, ?);");
+		", state"
+		", received_time"
+		", resolved_time"
+		") VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?);");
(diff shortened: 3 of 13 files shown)


wallet/wallet.c
```diff
@@ -2594,6 +2611,8 @@ const struct forwarding *wallet_forwarded_payments_get(struct wallet *w,
 
 		cur->channel_in.u64 = sqlite3_column_int64(stmt, 4);
 		cur->channel_out.u64 = sqlite3_column_int64(stmt, 5);
+		cur->received_time = sqlite3_column_int64(stmt, 6);
+		cur->resolved_time = sqlite3_column_int64(stmt, 7);
```
trueptolemy: Hi! In the `FORWARD_OFFERED` case, we store `resolved_time` with NULL. There should handle this case.
BTW, may I ask you a question: what does `FORWARD_OFFERED` mean?

I'm trying to add local failcode #2524  , maybe I should add a new forward status: `FORWARD_FAILED_LOCAL` and also set `resolved_time` with NULL.

Makefile
```diff
@@ -47,7 +47,7 @@ endif
 
 ifeq ($(COMPAT),1)
 # We support compatibility with pre-0.6.
-COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1
+COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1 -DCOMPAT_V070=1
```
m-schmoock: Maybe I'm captain obvious, but the Makefile only sets these preprocessor variables if `COMPAT` is also `1`. Currently, this causes a problem with Travis runner 3, where DCOMPAT_V070 is not defined at all: https://travis-ci.org/ElementsProject/lightning/jobs/515380339
Proposed solution (without being in the know of Travis internals and why it didnt fail for other COMPAT vairables):

```
ifeq ($(COMPAT),1)                                                                 
# We support compatibility with pre-0.6.                                           
COMPAT_CFLAGS=-DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1 -DCOMPAT_V070=1
else
COMPAT_CFLAGS=-DCOMPAT_V052=0 -DCOMPAT_V060=0 -DCOMPAT_V061=0 -DCOMPAT_V062=0 -DCOMPAT_V070=0
endif
```

wallet/wallet.c
cdecker: `FORWARD_OFFERED` means that the forward was received, and potentially forwarded, but wasn't resolved yet. The state machine looks like this:

```
FORWARD_OFFERED --------> FORWARD_SETTLED
                                           \----> FORWARD_FAILED
```

`FORWARD_OFFERED` therefore is the initial state for all forwards, while `FORWARD_SETTLED` and `FORWARD_FAILED` are the terminal states for those. Hence we only set the `resolved_time` once we reach a terminal state.

lightningd/peer_htlcs.c
```diff
@@ -1902,6 +1902,17 @@ static void listforwardings_add_forwardings(struct json_stream *response, struct
 				     cur->fee,
 				     "fee", "fee_msat");
 		json_add_string(response, "status", forward_status_name(cur->status));
+#if COMPAT_V070
```
m-schmoock: Or change to, and do not set to `0` :D

```
#ifdef COMPAT_V070
```

Makefile
cdecker: Correct, I added a fixup using `#ifdef` :-)

cdecker:
> On an intentionally failed payment somewhere along the route received_time is increasing but 'resolved_time' is epxceted to be (question:) not a number ? (see @trueptolemy 's comment)

I added an `if`-statement that only adds `resolved_time` once it was actually resolved, not `0` :-)

m-schmoock:
ACK 10d749a   (but squash the fixups)

wallet/wallet.c
trueptolemy: Thanks for your explain :)

rustyrussell:
nsec granularity please!

m-schmoock:
> nsec granularity please!

Yes, higher resolution would be better if we want to trace stuff in local test cases etc.
@rustyrussell   Typo? **m**illiseconds or **n**anoseconds?

rustyrussell:
It's fairly easy to have sub-microsecond network latency between peers, so I'd think that micro or nanoseconds are necessary.  If we store in a 64 bit value as nanoseconds since 1970, we don't get in any trouble for another 200+ years anyway (500+ if we allow unsigned).  But it's up to the implementer.

cdecker:
Nanoseconds sounds excessive, but can do :-)

cdecker:
>  If we store in a 64 bit value as nanoseconds since 1970, we don't get in any trouble for another 200+ years anyway (500+ if we allow unsigned). But it's up to the implementer.

We are storing as `u64` anyway, so that's not an issue. Just felt trivial to use `ts.tv_sec` directly. I'll change this to use `nsec` then ^^

cdecker:
Ok, rebased and squashed.

I also switched to storing the nsec timestamps in the DB (with some helpers) and display the timestamp with 3 significant digits in the JSON output.

Ping @rustyrussell and @m-schmoock :wink:

kristapsk:
This allows also calculate fees earned in a specific time period, like `lncli feereport` which has `day_fee_sum`, `week_fee_sum`, `month_fee_sum`, if I understand it correctly (didn't test myself).

cdecker:
> This allows also calculate fees earned in a specific time period, like lncli feereport which has day_fee_sum, week_fee_sum, month_fee_sum, if I understand it correctly (didn't test myself).

It should indeed, yes. Would you like to implement something like this? A neat little table `channel x date_bucket` seems like it could be nice.

kristapsk:
Yes, I could. You mean like a built-in feature or as a plug-in?

> A neat little table `channel x date_bucket` seems like it could be nice.

You mean like not only total fees collected but also fees collected per channel?

Btw I already [have a shell script](https://github.com/kristapsk/lightning-scripts), that calculates summaries for fees collected per channel. But so far I only could get total fees, not the fees in a specific time period.

rustyrussell:
BTW, our general philosophy should be to expose all the data, and let plugins sort it out.  We have a few cases where we public summary data, and it's never *quite* the right thing, but we're committed to it for backwards compat reasons.

kristapsk:
Doing it as a plugin was my first idea too and I will go that way unless somebody here tells me to do otherwise and integrate that functionality directly into C-lightning code.

Issue #2314, opened 2019-02-03
"warning_capacity": "No channels have sufficient incoming capacity"


Coinomatron:
Error when making an invoice. I'm probably not technical enough to see the solution, I have no idea how to fix this. I have 1 channel open, from which I've made payments.

rustyrussell:
It means there isn't enough incoming capacity in your 1 channel for someone to pay the invoice.  listchannels will show the channel balance.

Coinomatron:
It's an invoice for 1400 sats, the channel has 59712 left out of 60000 sats. Does it mean that the difference has to be equal or bigger than 1400 sats ?

conscott:
@Coinomatron  - yes if you run `lightning-cli listpeers`, your incoming channel balance is:
`msatoshi_total` - `msatoshi_to_us`, while your outgoing balance is simply `msatoshi_to_us`

Your can increase your incoming capacity by having someone open a channel with your node, or by simply spending more money on your existing channel

Coinomatron:
Is there a way to move funds from the channel that is too 'full' to another channel I have opened myself? I couldn't find something like that in the help commands.

cdecker:
> Is there a way to move funds from the channel that is too 'full' to another channel I have opened myself? I couldn't find something like that in the help commands.

For that you need to create a circular route. There is currently no tool for that. But as long as you have one channel with sufficient capacity you should not get that warning anymore.

Coinomatron:
Ok, so basically I need to spend an amount to receive that equal amount if I have 1 channel that's too 'full'?

cdecker:
Yep, that's basically it, you send out one way, and receive on the other.

Coinomatron:
Ok, closing, but would be nice to have some other solution to this than spending funds not to ones own channel(s).

Coinomatron:
One more thing, if people want to use lightning only to receive payments (for instance twitter bot), how would this issue be resolved?

conscott:
Anyone can still send you payments by opening (and funding) a channel with your node, therefore creating incoming capacity the rest of the network can route through. Your site can give the nodeid/ip/port information necessary to do this, or you can simply ask other well connected nodes (like people on irc/reddit/twitter) to open a channel with you.

Coinomatron:
Ah, ok. Thanks!

billygarrison:
> For that you need to create a circular route. There is currently no tool for that.

[...]

PR #2532, opened 2019-04-05
lightningd: simplify incoming channel selection for invoice's route-hint

SimonVrouwe:
lightningd: invoice route-hint: replace Weighted-Reservoir-Sampling algo with a simpler one

fixes issue #2314
The incoming channel in the route-hint to be included in an new invoice, now
simply favors the channel with largest incoming excess capacity.

The fact that ~pytest::test_invoice_routeboost~ `test_invoice_routeboost_private` succeeded before this commit was because the `fundamount` was small enough to favor it in the algo (weird?).
For example when using max `fundamount=16777215` instead, that test failed.

IMHO for small n.o. incoming channels (<10 ?), there is not really a use case for a complicated algo. But I haven't looked into the reasons why it did, privacy?

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -149,14 +149,9 @@ static struct route_info **select_inchan(const tal_t *ctx,
 {
 	const struct route_info *r = NULL;
 	struct route_info **ret;
-
+	struct amount_msat excess_max = AMOUNT_MSAT(0);
 	*any_offline = false;
 
-	/* Weighted reservoir sampling.
-	 * Based on https://en.wikipedia.org/wiki/Reservoir_sampling
-	 *	 Algorithm A-Chao
-	 */
-	u64 wsum = 0;
 	for (size_t i = 0; i < tal_count(inchans); i++) {
 		struct peer *peer;
 		struct channel *c;
@@ -197,11 +192,11 @@ static struct route_info **select_inchan(const tal_t *ctx,
 			continue;
 		}
 
-		/* Avoid divide-by-zero corner case. */
-		wsum += excess.millisatoshis + 1; /* Raw: rand select */
-		if (pseudorand(1ULL << 32)
-		    <= ((excess.millisatoshis + 1) << 32) / wsum) /* Raw: rand select */
+		/* Pick the channel with largest excess (incoming) capacity */
+		if (amount_msat_greater_eq(excess,excess_max)) {
+			excess_max = excess;
 			r = &inchans[i];
+		}
 	}
 
 	if (!r)
--- a/lightningd/test/run-invoice-select-inchan.c
+++ b/lightningd/test/run-invoice-select-inchan.c
@@ -611,7 +611,6 @@ int main(void)
 	bool any_offline;
 	struct route_info *inchans;
 	struct route_info **ret;
-	size_t n;
 
 	setup_locale();
 	secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY
@@ -631,7 +630,7 @@ int main(void)
 	assert(select_inchan(tmpctx, ld, AMOUNT_MSAT(0), inchans, &any_offline) == NULL);
 	assert(any_offline == false);
 
-	/* connected peer but no inchan -> NULL result. */
+	/* inchan but peer not connected -> NULL result. */
 	add_peer(ld, 1, CHANNELD_NORMAL, false);
 	assert(select_inchan(tmpctx, ld, AMOUNT_MSAT(0), inchans, &any_offline) == NULL);
 	assert(any_offline == false);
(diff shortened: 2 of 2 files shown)


m-schmoock:
Hm, but for larger number of channels this is not so good. How about a threshold like if  n<10 select best channel, else random set?

SimonVrouwe:
Isn't the purpose of bolt11 r field to provide the payer with a _reliable_ (but perhaps more expensive) forward route (hint)?

From [bolt11](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#requirements-3)

```
if there is NOT a public channel associated with its public key:

    MUST include at least one r field.
        r field MUST contain one or more ordered entries, indicating the forward route from a public node to the final destination. 
```
and
```
The r field allows limited routing assistance: as specified, it only allows minimum information to use private channels, however, it could also assist in future partial-knowledge routing.
```

> Hm, but for larger number of channels this is not so good.

@m-schmoock Could you elaborate?

m-schmoock:
@SimonVrouwe 

Im not sure, we should ask @rustyrussell for the rationale about the random selection.

An ideal algorithm would not just select the most excess channel, but try to balance out the channels, meaning trying to select the channel that will result in the best 'rebalancing score' (percentage change in capacity from current to optimal 50/50) after the payment has been made.

Additionally, always selecting the most excess channel will cause problems when a lot of invoices are made at the same time, leading to not reliable hints, as the first ones will drive down liquidity until the route hint actually becomes a false positive ;)

So there has to be database foo about what invoices with what hints we gave out in the near past (which sounds like overkill) or some random selection of a kind...

SimonVrouwe:
thanks @m-schmoock , I see what you mean

> trying to select the channel that will result in the best 'rebalancing score' (percentage change in capacity from current to optimal 50/50)

That doesn't sound too hard to implement in the current function `select_inchan`. But we need a way to operate on `u64 millisatoshis` member of the  `struct amount_sat` without `make check-amount-access` complaining. Any idea how to do that?

This is what I have so far:

```
                /* Optimal balance is at 50 percent capacity plus reserve */
		optim_balance_sat = c->funding.satoshis * 0.5
							+ c->our_config.channel_reserve.satoshis;

		/* Excess relative to optimal can be negative */
		excess_optim_sat = excess.millisatoshis * 1000 - optim_balance_sat;

		/* Pick the channel that benefits the most or suffers the least */
		if (excess_optim_sat >= excess_optim_max_sat) {
			excess_optim_max_sat = excess_optim_sat;
			r = &inchans[i];
		}
```
> So there has to be database foo about what invoices with what hints we gave out in the near past

Maybe the more advanced stuff can be build as a plugin? BTW, in developer mode the `invoice` command accepts a `dev-routes` array of route-hints.

m-schmoock:
> But we need a way to operate on `u64 millisatoshis` member of the `struct amount_sat` without `make check-amount-access` complaining. Any idea how to do that?

Well for testing before commiting to public repo, you can always do things like this to disarm `make check-amount-access`

```C
// setting
*(u64*) &your_msat_struct_here = your_u64_here;
// getting
u64 value = *(u64*) &your_msat_struct_here;
```
Note: With u32 you have to zero clear (`your_msat_struct_here = {0};`) before setting.

In the end we can add functions to `common/amount.c` when needed.

m-schmoock:
@SimonVrouwe As mentioned, I really would like the opinion of @cdecker or @rustyrussell when changing the random selection. They would refuse to merge anyway ;)

m-schmoock:
Two things:
- Document the algos rational in code ;)
- Is optimizing for channel balance here a good idea? Or does using plain excess any advantage?

SimonVrouwe:
Ok, the reasoning starts to make sense to me. But there is one great point of confusion in this line of code:
https://github.com/ElementsProject/lightning/blob/fbb494fba31e1e8ff5eb866de7addf1787b0bb1e/lightningd/invoice.c#L186-L188
In above I assume `avail` means the amount that can be payed-to-us before hitting a reserve, shouldn't then the _peer_'s reserve be subtracted instead of ours?

If I try to visualize the balance as (what I think) seen from our node:
```
                                 |<------avail-to-pay-us----------->|<--their_reserve->|
                                 |                                  |                  |
|----------|---------------------|----------------------------------|------------------|
0          ^                     ^                                  ^                funding
       our_reserve            our_msat                   (funding - their_reserve)
```
Perhaps I am wrong about the perspectives of `our_msat`, `our_reserve` and `our_config`?

Then about the selection algo, from the [wiki](https://en.wikipedia.org/wiki/Reservoir_sampling#Algorithm_A-Chao) page, Algorithm A-Chao:
```
(*
  S has items to sample, R will contain the result
  S[i].Weight contains weight for each item
 *)
WeightedReservoir-Chao(S[1..n], R[1..k])
  WSum = 0
  // fill the reservoir array
  for i = 1 to k
      R[i] := S[i]
      WSum = WSum + S[i].Weight
  for i = k+1 to n
    WSum = WSum + S[i].Weight
    p = S[i].Weight / WSum // probability for this item
    j := random(0, 1);     // important: inclusive range
    if j <= p              // select item according to probability
        R[random(1,k)] := S[i]  //uniform selection in reservoir for replacement
```

Looking at the current code (in master):
https://github.com/ElementsProject/lightning/blob/fbb494fba31e1e8ff5eb866de7addf1787b0bb1e/lightningd/invoice.c#L198-L204

1. We never seem to populate R in the first loop (which explains empty result)
2. There is a potential overflow when (u64) `(excess.millisatoshis + 1) << 32` (mentioned by @rustyrussell)
3. For each sample (channel) we use weight = excess capacity

> Is optimizing for channel balance here a good idea?

@m-schmoock  in point 3, maybe we could use a weight factor that favors channels which are brought closer into their optimal balance, which I think lies exactly between `our_reserve` and `(funding - their_reserve)`.

rustyrussell:
> Ok, the reasoning starts to make sense to me. But there is one great point of confusion in this line of code:
> 
> [lightning/lightningd/invoice.c](https://github.com/ElementsProject/lightning/blob/fbb494fba31e1e8ff5eb866de7addf1787b0bb1e/lightningd/invoice.c#L186-L188)
> 
> Lines 186 to 188 in [fbb494f](/ElementsProject/lightning/commit/fbb494fba31e1e8ff5eb866de7addf1787b0bb1e)
>  /* Even after reserve taken into account */ 
>  if (!amount_msat_sub_sat(&avail, 
>  			 avail, c->our_config.channel_reserve)) 
> 
> In above I assume `avail` means the amount that can be payed-to-us before hitting a reserve, shouldn't then the _peer_'s reserve be subtracted instead of ours?

We specify their reserve, and vice-versa.  I know, it's a bit confusing...

> If I try to visualize the balance as (what I think) seen from our node:

Yep, you are correct.

>     1. We never seem to populate R in the first loop (which explains empty result)
> 
>     2. There is a potential overflow when (u64) `(excess.millisatoshis + 1) << 32` (mentioned by @rustyrussell)
> 
>     3. For each sample (channel) we use weight = excess capacity

Yeah, code is too tricky.  Seriously, let's just (1) write a simple `excess_capacity()` helper, (2) do it in two passes, where we get the total excess capacity and pick a point in that at random, then iterate again to figure out which channel contains that point.

That is something everyone can understand!

Cheers,
Rusty,

SimonVrouwe:
Thanks for all the feedback, I will create a new PR and close this one.

PR #2535, opened 2019-04-07
hsmd: Add the check of node announcement context before sig generation

trueptolemy:
> /* FIXME(cdecker) Check the node announcement's content */

>  it needs to check that this is actually a node_announcement, but it doesn't need to check anything else. In fact, it could just check fromwire_peektype() and that it is at least 66 bytes long.

So now I add the check for `fromwire_peektype()` before sig generation!

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -1555,7 +1555,10 @@ static struct io_plan *handle_sign_node_announcement(struct io_conn *conn,
 		return bad_req_fmt(conn, c, msg_in,
 				   "Node announcement too short");
 
-	/* FIXME(cdecker) Check the node announcement's content */
+	if (fromwire_peektype(ann) != WIRE_NODE_ANNOUNCEMENT)
+		return bad_req_fmt(conn, c, msg_in,
+				   "Invalid announcement");
+
 	node_key(&node_pkey, NULL);
 	sha256_double(&hash, ann + offset, tal_count(ann) - offset);
 


hsmd/hsmd.c
```diff
@@ -1500,6 +1502,110 @@ static struct io_plan *handle_sign_invoice(struct io_conn *conn,
...
+	char *err = NULL;
+
+	/* check length of announcement */
+	if (len < offset) {
+		err = tal_fmt(tmpctx,
+			      "Node announcement too short");
+		goto done;
+	}
+
+	serialized = tal_dup_arr(tmpctx, u8, ann, len, 0);
```
cdecker: No need to copy it here, we're not modifying it below, so just use `ann` for it.

hsmd/hsmd.c
```diff
@@ -1500,6 +1502,110 @@ static struct io_plan *handle_sign_invoice(struct io_conn *conn,
...
+			      "Malformed wireaddrs %s in %s.",
+			      tal_hex(tmpctx, wireaddrs),
+			      tal_hex(tmpctx, serialized));
+		goto done;
+	}
+
+	/* check node id */
+	if (pubkey_cmp(&node_id, &c->id)) {
+		err = tal_fmt(tmpctx,
+			      "Client %s received announcement with wrong id %s.",
```
cdecker:
```suggestion
			      "HSMd(%s) asked to sign node announcement with wrong ID %s.",
```

hsmd/hsmd.c
trueptolemy: Thank you! I've modifide it! :)

hsmd/hsmd.c
trueptolemy: Corrected! Thanks for your review!

trueptolemy:
Fix these nits and fix node_id struct, now I've complete~~ @cdecker

hsmd/hsmd.c
```diff
@@ -1525,6 +1527,107 @@ static struct io_plan *handle_sign_invoice(struct io_conn *conn,
...
+		}
+
+		tal_arr_expand(&wireaddrs, wireaddr);
+	}
+	return wireaddrs;
+}
+
+static char *check_node_announcement_content(struct client *c, const u8 *ann)
+{
+	size_t offset = 66;
```
niftynei: where does this constant come from?

hsmd/hsmd.c
```diff
@@ -1525,6 +1527,107 @@ static struct io_plan *handle_sign_invoice(struct io_conn *conn,
...
+	u32 timestamp;
+	u8 rgb_color[3];
+	u8 alias[32];
+	u8 *features, *addresses;
+	struct wireaddr *wireaddrs;
+	struct node_id node_id;
+	char *err = NULL;
+
+	/* check length of announcement */
+	if (len < offset) {
```
niftynei: is the length not implicitly checked by `fromwire_node_annoucement`?

hsmd/hsmd.c
```diff
@@ -1525,6 +1527,107 @@ static struct io_plan *handle_sign_invoice(struct io_conn *conn,
...
+	}
+
+	if (!fromwire_node_announcement(tmpctx, ann,
+					&signature, &features, &timestamp,
+					&node_id, rgb_color, alias,
+					&addresses)) {
+		err = tal_fmt(tmpctx,
+			      "Malformed node_announcement %s",
+			      tal_hex(tmpctx, ann));
+		goto done;
```
niftynei: why not just return the err instead of jumping?   e.g.:

```
    return tal_fmt(...)
```

hsmd/hsmd.c
trueptolemy: Thanks for your review :)
 I noticed [here](https://github.com/ElementsProject/lightning/blob/ba41238df926e2a0e45f503c13674987e914faa2/hsmd/hsmd.c#L1544): 

> 	/* 2 bytes msg type + 64 bytes signature */
	size_t offset = 66;

hsmd/hsmd.c
trueptolemy: Thanks, it's superfluous!

hsmd/hsmd.c
trueptolemy: Yes! It's more direct

trueptolemy:
@rustyrussell Thank you! Now I know more about  the role of `hsmd`.
The origin code has checked the length, so I only need to check `fromwire_peektype()`

PR #2537, opened 2019-04-08
closingd: send option_dataloss_protect fields when reestablishing.

rustyrussell:
Travis caught an error where this happened: when closingd reconnects it
was sending the reestablish message without the option_dataloss_protect
fields.  That causes the peer to fail the channel!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ changes.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2537/commits/7979fc15e2fb991e232fdc99c050671af052552c

PR #2539, opened 2019-04-08
fix: format specifies type 'long *' but the argument has type 'u64 *'

brakmic:
Hi,

This PR fixes the format error in *devtools/create-gossipstore.c*.

```shell
devtools/create-gossipstore.c:27:58: error: format specifies type 'long *' but the argument has type 'u64 *' (aka 'unsigned long long *') [-Werror,-Wformat]
        while(fscanf(scidfd, "%s ,%ld\n", scids[i].scid, &scids[i].sat.satoshis) == 2 ) { /* Raw: read from file */
```

My environment:

macOS 10.14.3
clang 8.0.0
XCode 10.2 (10E125)

### DIFF
--- a/devtools/create-gossipstore.c
+++ b/devtools/create-gossipstore.c
@@ -24,7 +24,7 @@ struct scidsat * load_scid_file(FILE * scidfd)
         fscanf(scidfd, "%s\n", title);	
 	struct scidsat * scids = calloc(n, sizeof(scidsat));
 	int i = 0;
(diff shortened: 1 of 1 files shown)


cdecker:
Thanks @brakmic for this, I was just about to merge it, when I saw #2547 which addresses the same issue, but doesn't downcast the pointer to an architecture specific size, rather uses SCNu64 which is always defined as being 64 bits on all architectures. Hope you don't mind if I merge #2547 instead.

brakmic:
@cdecker No problem. :)

PR #2542, opened 2019-04-08
Update 'withdraw' and 'fundchannel' calls documentation

darosior:
This PR adds the doc for the `minconf` parameter that was not in the detailed help.


darosior:
Force pushed to also add the `minconf` parameter doc for the `fundchannel` call.

PR #2545, opened 2019-04-09
Gossip store optimizations II: the gossipd parts

rustyrussell:
Based on #2541 

At the end of this, we leave all the gossip on disk (except as needed for routing).  There are more optimizations available, but vitally this cuts gossipd mem usage from 2.6G to 577M for the million channels project.

Three other statistically significant changes:
1. Store rewrite goes from 36 to 13 seconds.
2. listnodes goes from 0.6 to 1.3 seconds.
3. Sending all gossip to peers increases from 50 to 67 seconds.

Across this entire PR:
```
-store_load_msec:32825-36365(34615.6+/-1.1e+03)
+store_load_msec:36034-37853(37109.8+/-5.9e+02)
-vsz_kb:2637488
+vsz_kb:577456
-store_rewrite_sec:35.150000-36.200000(35.59+/-0.4)
+store_rewrite_sec:12.490000-13.250000(12.862+/-0.27)
-listnodes_sec:0.590000-0.710000(0.682+/-0.046)
+listnodes_sec:1.250000-1.480000(1.364+/-0.09)
-listchannels_sec:28.650000-31.080000(29.966+/-0.81)
+listchannels_sec:30.820000-31.480000(31.068+/-0.24)
-routing_sec:29.980000-33.100000(31.284+/-1.5)
+routing_sec:26.940000-27.990000(27.616+/-0.39)
-peer_write_all_sec:49.020000-52.890000(50.376+/-1.5)
+peer_write_all_sec:65.690000-68.600000(66.698+/-0.99)
```

PR #2546, opened 2019-04-09
Gossip Store Optimization III: Now i can spell Dijkstra

rustyrussell:
Based on #2545 

This cuts down the size of 'struct node' since Dijkstra needs 1/20 of the room that Bellman Ford Gibson did, but with a bit of algorithmic violence we can make it do what we want.

~~~I seem to have hit underflow when measuring route times on a million channels though:~~~

There was a bug in the test.  We go from about 60 seconds to 12 though, not bad.


rustyrussell:
Rebased with benchmark fix prepended

cdecker:
Excellent PR, had a lot of fun reading through and refreshing my memory about Dijkstra as well :wink:

PR #2549, opened 2019-04-09
gossipd: change the pending_cannouncement list to htable

trueptolemy:
fix `/* FIXME: Make this a htable! */` !

Change the pending_cannouncement (in `struct node`) from `struct list_head pending_cannouncement;` to `struct pending_cannouncement_map *pending_cannouncements;`

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -25,30 +25,6 @@
 /* 365.25 * 24 * 60 / 10 */
 #define BLOCKS_PER_YEAR 52596
 
-/* We've unpacked and checked its signatures, now we wait for master to tell
- * us the txout to check */
-struct pending_cannouncement {
-	/* Off routing_state->pending_cannouncement */
-	struct list_node list;
-
-	/* Unpacked fields here */
-	struct short_channel_id short_channel_id;
-	struct node_id node_id_1;
-	struct node_id node_id_2;
-	struct pubkey bitcoin_key_1;
-	struct pubkey bitcoin_key_2;
-
-	/* The raw bits */
-	const u8 *announce;
-
-	/* Deferred updates, if we received them while waiting for
-	 * this (one for each direction) */
-	const u8 *updates[2];
-
-	/* Only ever replace with newer updates */
-	u32 update_timestamps[2];
-};
-
 struct pending_node_announce {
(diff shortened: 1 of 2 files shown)


gossipd/routing.h
```diff
@@ -198,9 +199,8 @@ struct routing_state {
 	/* node_announcements which are waiting on pending_cannouncement */
 	struct pending_node_map *pending_node_map;
 
-	/* FIXME: Make this a htable! */
 	/* channel_announcement which are pending short_channel_id lookup */
-	struct list_head pending_cannouncement;
+	struct pending_cannouncement_map *pending_cannouncements;
```
rustyrussell: Nice!  Only one trivial change: we don't need a pointer here, saving an allocation.

trueptolemy: Thank you! I've done it :)

BTW, when shouldn't I use the pointer for a htable?
There `nodes` and `pending_node_map` are both the pointer to htable. Should I also save them as allocations?
```
struct routing_state {
...
	struct node_map *nodes;

	struct pending_node_map *pending_node_map;

	struct pending_cannouncement_map pending_cannouncements;
...
};
```

rustyrussell: Yes!  Unless you try to convert it and realize there's some subtle reason I'm wrong :)

gossipd/routing.c
```diff
@@ -832,19 +810,23 @@ static struct pending_cannouncement *
...
+	struct pending_cannouncement *pann;
 
-	list_for_each(&rstate->pending_cannouncement, i, list) {
-		if (short_channel_id_eq(scid, &i->short_channel_id))
-			return i;
+	for (pann = pending_cannouncement_map_first(&rstate->pending_cannouncements, &i);
+	     pann;
+	     pann = pending_cannouncement_map_next(&rstate->pending_cannouncements, &i)) {
+		if (short_channel_id_eq(scid, &pann->short_channel_id))
+			return pann;
```
rustyrussell: Oops, this iterates through the hash table!  You want:

```
pann = pending_cannouncement_map_get(&rstate->pending_cannouncements, scid);
```

trueptolemy: I'm so Sorry...I corrected it!

PR #2553, opened 2019-04-10
Experimental: include experimental_csv in build

niftynei:
this changeset solves for allowing a dev to check-in experimental feature csv changes. downstream users/builders can then optionally include these features by turning on --enable-experimental-features.

there's 3 roles for this:

1) protocol dev adding a diff for CSV changes. they need a command to add a patchfile for the csv's from their local/branched version of the spec repo. their updates end up as an appended patchset to wire/gen_experimental_[peer|onion]_csv.*
2) user who wants to run experimental features, has EXPERIMENTAL_FEATURES turned on. they'll have the checked-in experimental_csv files applied to their csv files.
3) user who wants to run default c-lightning without experimental features. all experimental features and csv patches are left out.


devtools/Makefile
```diff
@@ -22,11 +22,19 @@ DEVTOOLS_COMMON_OBJS :=				\
 
 devtools-all: $(DEVTOOLS)
 
+ifeq ($(EXPERIMENTAL_FEATURES),1)
+devtools/gen_print_wire.h: $(WIRE_GEN) wire/gen_peer_wire_csv $(wildcard wire/experimental_csv)
+	cat wire/gen_peer_wire_csv wire/experimental_csv | grep -v "cat: " | $(WIRE_GEN) --bolt --printwire --header $@ wire_type > $@
```
rustyrussell: `grep -v "cat: "` is weird: what cat prints cat?

This also won't work in the case where experimental_csv doesn't exist (and you don't include a rule to build it).  You worked around the dependency with $(wildcard) but I think you should actually insist it exists, and add a rule to build it.

So, commit 1 should change/remove the rule which generates wire/gen_peer_wire_csv (and the onion one) in favor of an explicit target which regenerates it (eg. `make spec-csv`?).  commit 2 should add a similar rule to make the experimental ones (eg. make spec-csv-experimental) perhaps by doing a diff or something?  You should also check in an empty wire/experimental_csv.

Then this, minus $(wildcard) logic, becomes commit 3.

wire/Makefile
```diff
@@ -26,27 +26,62 @@ ALL_OBJS += $(WIRE_OBJS) $(WIRE_ONION_OBJS)
...
+EXP_BOLT_EXTRACT=$(EXP_BOLTDIR)$(EXTRACT_PATH)
+
+wire/gen_peer_experimental_csv: wire/gen_peer_wire_csv
+	@if [ -d $(EXP_BOLTDIR) ]; then for f in $(EXP_BOLTDIR)/0[127]*.md $(EXP_BOLT_EXTRACT); do if [ $$f -nt $@ -o ! -f $@ ]; then $(EXP_BOLT_EXTRACT) --message-fields --message-types $(EXP_BOLTDIR)/0[127]*.md | diff -u wire/gen_peer_wire_csv - >>$@; break; fi; done; fi;
+
+wire/gen_onion_experimental_csv: wire/gen_onion_wire_csv
+	@if [ -d $(EXP_BOLTDIR) ]; then for f in $(EXP_BOLTDIR)/04*.md $(EXP_BOLT_EXTRACT); do if [ $$f -nt $@ -o ! -f $@ ]; then { echo '#include <wire/onion_defs.h>'; $(EXP_BOLT_EXTRACT) --message-fields --message-types $(EXP_BOLTDIR)/04*.md; } | diff -u $< - >>$@; break; fi; done; fi;
+
+spec-exp-peer: wire/gen_peer_experimental_csv
+spec-exp-onion: wire/gen_onion_experimental_csv
```
niftynei: these don't really work the way i want them to (regenerate the dependency file) but i didn't have much luck finding an alternative way to have Make re-run a 'named file target'. suggestions on how to fix this would be v welcome.

rustyrussell:
Deep in bikeshedding here.  Let me try!

rustyrussell:
OK, look at https://github.com/rustyrussell/lightning/tree/better-bolt-extraction

You can rebase on top of this, then add your empty `wire/extracted_experimental_peer_csv` and wire/extracted_experimental_onion_csv` (or whatever) and add a extract-experimental-bolt-csv target which updates them (using the same BOLTVERSION and BOLTDIR vars, since you can set them when you run the make cmd).

niftynei:
this changeset solves for allowing a dev to check-in experimental feature csv changes. downstream users/builders can then optionally include these features by turning on `--enable-experimental-features`. 

there's 3 roles for this:
1) protocol dev adding a diff for CSV changes. they need a command to add a patchfile for the csv's from their local/branched version of the spec repo. their updates end up as an appended patchset to `wire/gen_experimental_[peer|onion]_csv`.*
2) user who wants to run experimental features, has EXPERIMENTAL_FEATURES turned on. they'll have the checked-in experimental_csv files applied to their csv files.
3) user who wants to run default c-lightning without experimental features. all experimental features and csv patches are left out.

\* going to use your suggestion https://github.com/rustyrussell/lightning/commit/ac4e80f087ea8903e2a308416100e89c09ca6d9a to turn this into a command, rather than a file requirement.

thoughts on this? too ambitious? i like that it makes it easier to share spec changesets in a fairly controlled manner. once the proposed changes are checked into master on the RFC repo, you'll have to manually remove the patch from the experimental_csv file.

niftynei:
I should have included this write up in the original description. whoops.

wire/Makefile
niftynei: ok updated, thanks for the suggestion @rustyrussell

PR #2554, opened 2019-04-10
Minor small UX fixes

rustyrussell:
Really should be separate PRs, but I was being lazy.


cdecker:
One of the mocks needs to be implemented now, it seems it's actually being called now. Otherwise this is looking good.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2554/commits/ec6719d892a38cd1a8c3de08c9b23c5b1d787cba

PR #2557, opened 2019-04-10
closing: add message to billboard when closing txn is broadcast

### LINKED ISSUE #2552: add info about closing tx to listpeers' status
when you're closing a channel, it's hard to tell what the status of it is if the tx is in the mempool, since we don't track them til they're onchain.

it'd be nice to add a msg to listpeer's status with info about the broadcast status of a closing tx and the txid.

niftynei:
make it a bit easier to track mutual channel closures by
adding broadcast txid to the listpeers billboard

fixes #2552

### DIFF
--- a/lightningd/closing_control.c
+++ b/lightningd/closing_control.c
@@ -83,6 +83,7 @@ static void peer_received_closing_signature(struct channel *channel,
 {
 	struct bitcoin_signature sig;
 	struct bitcoin_tx *tx;
+	struct bitcoin_txid tx_id;
 	struct lightningd *ld = channel->peer->ld;
 
 	if (!fromwire_closing_received_signature(msg, msg, &sig, &tx)) {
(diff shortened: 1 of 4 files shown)


lightningd/closing_control.c
```diff
@@ -120,6 +124,12 @@ static void peer_closing_complete(struct channel *channel, const u8 *msg)
 	if (channel->state == CLOSINGD_COMPLETE)
 		return;
 
+	bitcoin_txid(channel->last_tx, &closing_txid);
+	billboard_msg = tal_fmt(channel,
```
niftynei: not sure if `channel` is the right level to allocate onto?

lightningd/closing_control.c
```diff
@@ -120,6 +124,12 @@ static void peer_closing_complete(struct channel *channel, const u8 *msg)
...
+			"Broadcasting closing transaction. txid:%s",
+			type_to_string(channel->last_tx, struct bitcoin_txid,
+				&closing_txid));
+	channel_set_billboard(channel, false, billboard_msg);
```
niftynei: i set this as transient; it disappears once the tx lands on chain

lightningd/closing_control.c
rustyrussell: No, that would be a (technical) memory leak. (Technical because this can only happen once per channel, so it's ultra constrained).  Unless you use "take()" below to hand it to channel_set_billboard, in whihc case the pattern is to allocate off NULL to make it clear it's going to be taken().

niftynei: cool, updated!

niftynei:
ok, rewrote it to pass back the last_tx's id from `lightningd` so that `closingd` can do the honors of updating the billboard with the txid when it's found an agreeable feerate.

PR #2560, opened 2019-04-11
More lnd compat fixes

rustyrussell:


channeld/channeld.c
```diff
@@ -2114,6 +2114,24 @@ static void check_current_dataloss_fields(struct peer *peer,
 	status_trace("option_data_loss_protect: fields are correct");
 }
 
+/* Older LND sometimes sends funding_locked before reestablish! */
+/* ... or announcement_signatures.  Sigh, let's handle whatever they send. */
+static bool capture_premature_msg(const u8 ***shit_lnd_says, const u8 *msg)
```
cdecker: Love the variable naming xD

closingd/closingd.c
```diff
@@ -171,8 +171,17 @@ static void do_reconnect(struct crypto_state *cs,
...
-		channel_reestablish = closing_read_peer_msg(tmpctx, cs, channel_id);
+	if (!channel_reestablish) {
+		do {
+			tal_free(channel_reestablish);
+			channel_reestablish = closing_read_peer_msg(tmpctx, cs,
+								    channel_id);
+			/* They *should* send reestablish first, but lnd
+			 * sends other messages, which we can ignore since
+			 * we're closing anyway... */
+		} while (fromwire_peektype(channel_reestablish)
```
niftynei: is it possible to get stuck in a loop here if `channel_reestablish` type is never sent?

rustyrussell: Yes, but we can do that anyway if they never send anything, so I don't think it's an issue...

PR #2562, opened 2019-04-11
Increase min capacity sat default

m-schmoock:
The old value of 1000 sat was too small to cover the dust reserves and
    initial commitment fees. This lead to the situation when trying to open
    a channel with minimal amount, the channels got refused because
    they were not able cover the commitment fees.
    
    For this reason the minimal capacity should be increased to i.e. 10k
    satoshi, as the technical minimum that also accounts for fees and
    reserves is somewhere around 6k sat.

    additionally we substract the dust reserves on the fly from that value,
    so that a `cli fundchannel VALUE` is possible with the same VALUE.

### DIFF
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -664,8 +664,9 @@ static void channel_config(struct lightningd *ld,
 	*max_to_self_delay = ld->config.locktime_max;
 
 	/* Take minimal effective capacity from config min_capacity_sat */
-	amount_msat_from_sat_u64(min_effective_htlc_capacity,
-		ld->config.min_capacity_sat);
+	if (!amount_msat_from_sat_u64(min_effective_htlc_capacity,
+				ld->config.min_capacity_sat))
+		fatal("amount_msat overflow for config.min_capacity_sat");
 
 	/* BOLT #2:
 	 *
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -660,6 +660,8 @@ static void channel_config(struct lightningd *ld,
 			   u32 *max_to_self_delay,
 			   struct amount_msat *min_effective_htlc_capacity)
 {
+	struct amount_msat dust_limit;
+
 	/* FIXME: depend on feerate. */
 	*max_to_self_delay = ld->config.locktime_max;
 
@@ -667,6 +669,17 @@ static void channel_config(struct lightningd *ld,
 	if (!amount_msat_from_sat_u64(min_effective_htlc_capacity,
 				ld->config.min_capacity_sat))
 		fatal("amount_msat overflow for config.min_capacity_sat");
+	/* Substract 2 * dust_limit, so fundchannel with min value is possible */
+	if (!amount_sat_to_msat(&dust_limit, get_chainparams(ld)->dust_limit))
+		fatal("amount_msat overflow for dustlimit");
+	if (!amount_msat_sub(min_effective_htlc_capacity,
+				*min_effective_htlc_capacity,
+				dust_limit))
+		*min_effective_htlc_capacity = AMOUNT_MSAT(0);
+	if (!amount_msat_sub(min_effective_htlc_capacity,
+				*min_effective_htlc_capacity,
+				dust_limit))
+		*min_effective_htlc_capacity = AMOUNT_MSAT(0);
 
 	/* BOLT #2:
 	 *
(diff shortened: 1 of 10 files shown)


m-schmoock:
@cdecker yes, thanks. just noted.

Also I post my IRC questions here, to make them persistent (not for bugging):

1. Do you think 10k sat is a good enough value, or may the value be soon too small when fees rise just a bit?
2. I noted that the limit also applies to channels of the node that funded the channel, as there will be a channel in both directions. The 'back-channel' will have the same limitation. So I cannot open a channel (by my intention) below my own limit. Is this okay, or do we need an additional if-statement that figures this out and allows it?

m-schmoock:
> Looks like there are a few more tests failing due to insufficient capacity now.

Addressed by ee42042 ,  the test used 10k sats for channel funding, which is (after fees) effective ~9000 sat , below our new minimum. Simplest fix to increase test channel funding to 100k, which didnt break the context of the test.

m-schmoock:
@cdecker what do you think about @niftynei comment https://github.com/ElementsProject/lightning/pull/2562#pullrequestreview-225867077   ?

(adding 2 x dust reserves, so fundchannel with that limit works)

m-schmoock:
@cdecker @niftynei  I updated this branch and added on-the-fly substraction of the dustlimit reserves as the last commit. Now you can fund using the exact minimum amount as for the limit as long as it is above the initial commitment fees.

lightningd/opening_control.c
```diff
@@ -660,12 +660,16 @@ static void channel_config(struct lightningd *ld,
...
+
 	/* FIXME: depend on feerate. */
 	*max_to_self_delay = ld->config.locktime_max;
 
 	/* Take minimal effective capacity from config min_capacity_sat */
-	amount_msat_from_sat_u64(min_effective_htlc_capacity,
+	overflows = amount_msat_from_sat_u64(min_effective_htlc_capacity,
 		ld->config.min_capacity_sat);
+	if (!overflows)
+		fatal("amount_msat overflow for config.min_capacity_sat");
```
rustyrussell: This var name is very very confusing!  Perhaps best to just:

```
if (!amount_msat_from_sat_u64(min_effective_htlc_capacity, ld->config.min_capacity_sat))
    fatal().
```

Though I wonder about making min_effective... an msat might be simpler, that's a larger change.

tests/test_invoices.py
```diff
@@ -181,7 +181,7 @@ def test_invoice_routeboost_private(node_factory, bitcoind):
     # Attach public channel to l1 so it doesn't look like a dead-end.
     l0 = node_factory.get_node()
     l0.rpc.connect(l1.info['id'], 'localhost', l1.port)
-    scid = l0.fund_channel(l1, 2 * (10**4))
+    scid = l0.fund_channel(l1, 2 * (10**5))
```
rustyrussell: For future reference: commit message should say *why* this is done; in this case "we're about to increase the minimum channel capacity".

lightningd/opening_control.c
```diff
@@ -670,6 +671,22 @@ static void channel_config(struct lightningd *ld,
...
+	/* Substract 2 * dust_limit, so fundchannel with min value is possible */
+	overflows = amount_sat_to_msat(&dust_limit, get_chainparams(ld)->dust_limit);
+	if (!overflows)
+		fatal("amount_msat overflow for dustlimit");
+	overflows = amount_msat_sub(
+			min_effective_htlc_capacity,
+			*min_effective_htlc_capacity,
+			dust_limit);
+	if (!overflows)
+		amount_msat_from_u64(min_effective_htlc_capacity, (u64)0);
```
rustyrussell:
```
min_effective_htlc_capacity = AMOUNT_MSAT(0);
```

lightningd/opening_control.c
```diff
@@ -670,6 +671,22 @@ static void channel_config(struct lightningd *ld,
...
+	overflows = amount_msat_sub(
+			min_effective_htlc_capacity,
+			*min_effective_htlc_capacity,
+			dust_limit);
+	if (!overflows)
+		amount_msat_from_u64(min_effective_htlc_capacity, (u64)0);
```
rustyrussell: `min_effective_htlc_capacity = AMOUNT_MSAT(0);` as above.

tests/test_invoices.py
m-schmoock: Thanks, I added this to the commits log message, as the standard commit message is a bit too short reasoning.

lightningd/opening_control.c
m-schmoock: Thanks for the hint, all this amount_msat handling seemed to turn off my sense for nice code :D

m-schmoock:
Okay, thanks for all your input, we are done here if Travis is happy...

rustyrussell:
Nice work!

Ack 4fa89863b97c2c50e6569d4e71fa9a291bb5537b

PR #2564, opened 2019-04-11
Lightningd: Refactor and improve incoming channel selection for invoice's route-hint

### LINKED ISSUE #2314: "warning_capacity": "No channels have sufficient incoming capacity"

SimonVrouwe:
This is a follow up of #2532

Fix and refactor `select_inchan` for invoice route-hint, now use fractional excess as weight.

Refactored the weighted-reservoir-sampling algo to make it more straightforward.
It now uses the excess as percentage of capacity as weight. This favors channels that
are more _relatively_ unbalanced to be used for incoming payment.

Now passes test_invoice_routeboost_private() when using max fundamount=16777215.
Hope that this fixes some issues.

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -313,27 +313,45 @@ static struct command_result *parse_fallback(struct command *cmd,
 	return NULL;
 }
 
-/* BOLT11 struct wants an array of arrays (can provide multiple routes) */
+/*
+ * From array of incoming channels [inchan], find suitable ones for
+ * a payment-to-us of [amount_needed], using criteria:
+ * 1. Channel's peer is known, in state CHANNELD_NORMAL and is online.
+ * 2. Channel's peer capacity to pay us is sufficient.
+ *
+ * Then use weighted reservoir sampling, which makes probing channel balances
+ * harder, to choose one channel from the set of suitable channels. It favors
+ * channels that have less balance on our side as fraction of their capacity.
+ *
+ * [any_offline] is set if the peer of any suitable channel appears offline.
+ */
 static struct route_info **select_inchan(const tal_t *ctx,
 					 struct lightningd *ld,
-					 struct amount_msat capacity_needed,
+					 struct amount_msat amount_needed,
 					 const struct route_info *inchans,
 					 bool *any_offline)
 {
-	const struct route_info *r = NULL;
-	struct route_info **ret;
+	/* BOLT11 struct wants an array of arrays (can provide multiple routes) */
+	struct route_info **R;
+	double wsum, p;
+
+	struct sample {
+		const struct route_info *route;
+		double weight;
+	};
+
+	struct sample *S = tal_arr(tmpctx, struct sample, 0);
 
 	*any_offline = false;
 
-	/* Weighted reservoir sampling.
-	 * Based on https://en.wikipedia.org/wiki/Reservoir_sampling
-	 *	 Algorithm A-Chao
-	 */
-	u64 wsum = 0;
+	/* Collect suitable channels and assign each a weight.  */
 	for (size_t i = 0; i < tal_count(inchans); i++) {
 		struct peer *peer;
 		struct channel *c;
-		struct amount_msat avail, excess;
+		struct sample sample;
+		struct amount_msat their_msat, capacity_to_pay_us, excess, capacity;
+		struct amount_sat cumulative_reserve;
+		double excess_frac;
 
 		/* Do we know about this peer? */
 		peer = peer_by_id(ld, &inchans[i].pubkey);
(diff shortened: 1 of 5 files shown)


SimonVrouwe:
Squashed and some minor re-wording. To be honest, I still don't understand the current implementation with the left shifts.

@m-schmoock Would love to hear your opinion on using excess percentage of capacity as weight factor. Note that incoming payments can only shift the balance towards our side, so here we can only optimize one-way.

If the comments are too verbose, please tell.

m-schmoock:
> @m-schmoock Would love to hear your opinion on using excess percentage of capacity as weight factor. Note that incoming payments can only shift the balance towards our side, so here we can only optimize one-way.

Yes, I think that is (kinda?) what I meant. Two remarks: 

- What is the `0.1` supposed to do? :
```C
excess_perc = (double)excess.millisatoshis * 0.1 /               /* Raw:excess_perc */
```
- 'excess_perc' is (in your proposal) being measured as a total against remaining capacity, right? I thought `excess_perc` should boost channels that are farther away from their 50/50 balance state, this will take total capacity into account, as a small channel is easier distorted than a big channel.

> If the comments are too verbose, please tell.

There is no such thing as being to verbose

m-schmoock:
It could treat channels that are already past their balance as 'neutral' as kind of a lambda function.

SimonVrouwe:
> What is the 0.1 supposed to do?

we go from `millisatoshis` to `satoshis` x 100% (0.001 * 100% = 0.1)

> excess_perc' is (in your proposal) being measured as a total against remaining capacity, right? I thought excess_perc should boost channels that are farther away from their 50/50 balance state,

Please have a look at this graph
https://github.com/ElementsProject/lightning/blob/2fbabbac21fe6b84d165a12238b6a000b7433abc/lightningd/invoice.c#L365-L378

`excess_perc = 100 % * excess / capacity` is same as `capacity_to_pay_us` _after_ this payment.

We cannot optimize `excess_perc` to 50% with only incoming payments, unless you want to shutdown incoming payment entirely when we pass that? (I don't think).

But we can select the channel that needs it the most or suffers the least. So even if all incoming channels `excess_perc`'s are below 50%, the one with highest `excess_perc` is still the _least bad_ option. To bring them back toward 50% (or higher) should be done by the _outgoing_ payment route selection.

lightningd/invoice.c
```diff
@@ -356,33 +388,51 @@ static struct route_info **select_inchan(const tal_t *ctx,
...
-		if (!amount_msat_sub(&excess, avail, capacity_needed))
+		if (!amount_msat_sub(&excess, capacity_to_pay_us, amount_needed))
 			continue;
 
+		/* Excess as percentage of capacity, assumes check_config_bounds */
+		excess_perc = (double)excess.millisatoshis * 0.1 /               /* Raw:excess_perc */
+			  (c->funding.satoshis                                       /* Raw:excess_perc */
+			  - c->our_config.channel_reserve.satoshis                   /* Raw:excess_perc */
+			  - c->channel_info.their_config.channel_reserve.satoshis);  /* Raw:excess_perc */
+
```
rustyrussell: OK, with that many /* Raw */ overrides is a big red flag.  We do overflow checking to ensure robustness, and we should keep that:

How about:
```
   struct amount_msat total_capacity;

   if (!amount_sat_to_msat(&total_capacity, c->funding)
       || !amount_msat_sub_sat(&total_capacity, c->our_config.channel_reserve)
       || !amount_msat_sub_sat(&total_capacity, c->channel_info.their_config.channel_reserve)) {
           log_broken(lg->log, "Channel %s capacity %s overflow (reserves %s + %s)!",
                          type_to_string(tmpctx, struct short_channel_id, &c->scid),
                          type_to_string(tmpctx, struct amount_sat, &c->funding),
...;
           continue;
    }

    excess_perc = (double)excess.millisatoshis / total_capacity.millisatoshis; /* Raw: double fraction */
```

I suggest changing excess_perc to excess_fraction; the 100 multiplier is just a distraction AFAICT?

lightningd/invoice.c
```diff
@@ -356,33 +388,51 @@ static struct route_info **select_inchan(const tal_t *ctx,
...
+	 * https://en.wikipedia.org/wiki/Reservoir_sampling#Algorithm_A-Chao
+	 * But (currently) the result will consist of only one sample (k=1) */
+	R = tal_arr(ctx, struct route_info *, 1);
+	R[0] = tal_dup(R, struct route_info, S[0].route);
+	wsum = S[0].weight;
+
+	for (size_t i = 1; i < tal_count(S); i++) {
+		wsum += S[i].weight;
+		p = S[i].weight / wsum;
+		double random_1 = (double)pseudorand(UINT64_MAX) / UINT64_MAX;
```
rustyrussell: We should expose (separate commit I'm afraid!) pseudorand_double() which calls isaac64_next_double() underneath.  Much nicer than this...

lightningd/invoice.c
SimonVrouwe: Agreed, I see what you mean. Then a small point: in the code base I couldn't find consistent use of the term `capacity`, [sometimes](https://github.com/ElementsProject/lightning/blob/77236caa91bc548d24d5e886bcaf659cc1fd4c77/gossipd/routing.c#L65) it's synonym for `funding`, and other [times](https://github.com/ElementsProject/lightning/blob/77236caa91bc548d24d5e886bcaf659cc1fd4c77/openingd/openingd.c#L201) for `funding` minus `reserves`.

The later seems more appropriate to me, which would then make `total_capacity` synonym to `funding` (or just call it `funding`).

SimonVrouwe:
fixups:
- use pseudorand_double()
- better overflow checking and use fraction instead of percentage
- tiny reordering

lightningd/invoice.c
SimonVrouwe: Hope that is ok.

Issue #2563, opened 2019-04-11
'bad block' crashing node


conscott:
### Issue and Steps to Reproduce

I am trying to resync my testnet node, and it seems to fumble on block `000000000000096f7e93d68aa1806c39823b659c8b1c438c8bbce94fd0cbdfb3` with the [bad block message](https://github.com/ElementsProject/lightning/blob/master/lightningd/bitcoind.c#L462)

Stack trace is
```
2019-04-11T13:54:09.885Z lightningd(6384): bitcoin-cli -testnet -rpcuser=... -rpcpassword=... getblock 000000000000096f7e93d68aa1806c39823b659c8b1c438c8bbce94fd0cbdfb3 false: bad block 

[...]

PR #2565, opened 2019-04-12
libwally: update to 0.6.8.

rustyrussell:
This fixes block parsing on testnet; specifically, non-standard tx versions.

We hit a type bug in libwally (wallt_get_secp_context()) which I had to
work around for the moment, and the updated libsecp adds an optional hash
function arg to the ECDH function.

Fixes: #2563


rustyrussell:
@cdecker Any idea why Travis isn't (AFAICT) updating the git submodule?  Works locally on a fresh clone...

cdecker:
@rustyrussell need to nuke the build caches. We are caching the `external` directory to avoid recompiling it every time.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2565/commits/8998772d459aaa1b96b3d802090de39ef84dfff4

PR #2567, opened 2019-04-13
rebalance plugin in python

gallizoltan:
- this plugin helps to move some liquidity between your channels using circular payments
- based on my previous work published here: https://medium.com/coinmonks/redistribute-lightning-channels-balances-9ba3265584ee

### DIFF
--- a/contrib/plugins/rebalance.py
+++ b/contrib/plugins/rebalance.py
@@ -1,12 +1,13 @@
 #!/usr/bin/env python3
 from lightning import Plugin, RpcError
 import time
+import uuid
 
-plugin = Plugin(autopatch=True)
+plugin = Plugin()
 
 
 def setup_routing_fees(plugin, route, msatoshi):
-    delay = 9
+    delay = int(plugin.get_option('cltv-final'))
     for r in reversed(route):
         r['msatoshi'] = msatoshi
         r['amount_msat'] = str(msatoshi) + "msat"
@@ -15,17 +16,43 @@ def setup_routing_fees(plugin, route, msatoshi):
         for ch in channels.get('channels'):
             if ch['destination'] == r['id']:
                 fee = ch['base_fee_millisatoshi']
-                fee += msatoshi * ch['fee_per_millionth'] / 1000000
-                msatoshi += round(fee)
+                fee += msatoshi * ch['fee_per_millionth'] // 1000000
+                msatoshi += fee
                 delay += ch['delay']
 
 
-def peer2channel(plugin, channel_id, my_node_id):
+def peer2channel(plugin, channel_id, my_node_id, payload):
     channels = plugin.rpc.listchannels(channel_id).get('channels')
     for ch in channels:
         if ch['source'] == my_node_id:
             return ch['destination']
-    raise ValueError('Cannot find peer for channel: ' + channel_id)
+    raise RpcError("rebalance", payload, {'message': 'Cannot find peer for channel: ' + channel_id})
+
+
+def find_worst_channel(route):
+    if len(route) < 4:
+        return None
+    start_id = 2
+    worst = route[start_id]['channel']
+    worst_val = route[start_id - 1]['msatoshi'] - route[start_id]['msatoshi']
+    for i in range(start_id + 1, len(route) - 1):
+        val = route[i - 1]['msatoshi'] - route[i]['msatoshi']
+        if val > worst_val:
+            worst = route[i]['channel']
+            worst_val = val
+    return worst
+
+
+def rebalance_fail(plugin, label, payload, success_msg, error=None):
+    try:
+        plugin.rpc.delinvoice(label, 'unpaid')
+    except RpcError as e:
+        # race condition: waitsendpay timed out, but invoice get paid
+        if 'status is paid' in e.error.get('message', ""):
+            return success_msg
+    if error is None:
+        error = RpcError("rebalance", payload, {'message': 'Rebalance failed'})
+    raise error
 
 
 @plugin.method("rebalance")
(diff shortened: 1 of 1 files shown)


gallizoltan:
I would be glad for a thorough review. I'm not sure my routing fee calculation method is correct. Also, I have some doubt about handling the payments pending for a long time.

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
...
+    label = "Rebalance" + str(start_ts)
+    invoice = plugin.rpc.invoice(msatoshi, label, "Rebalance", int(retry_for) + 60)
+    payment_hash = invoice['payment_hash']
+    plugin.log("Invoice payment_hash: %s" % payment_hash)
+    try:
+        error_logs = ""
+        erring_channel_msg = "\'erring_channel\': \'%s\'"
+        while int(time.time()) - start_ts < int(retry_for):
+            try:
+                r = plugin.rpc.getroute(incoming_node_id, msatoshi, riskfactor=1, cltv=9, fromid=outgoing_node_id)
```
rustyrussell: OK, this loop probably doesn't do very much?

1. You should exclude all your own channels; you don't want it to route back through you.  Using a single direction is sufficient to prevent that.
2. If you get an erring channel, you should either (a) exclude it, or (b) try for a lesser amount.

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
...
+                route_mid = r['route']
+            except RpcError:
+                # sometimes there is a route, but getroute raises 'Could not find a route' exception
+                continue
```
rustyrussell: This seems weird, and shouldn't happen in 0.7?  In 0.6.3, the routing code automagically excluded failed channels for 60 seconds, now it doesn't.

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
...
+            if any(r['id'] == my_node_id for r in route_mid):
+                continue
+            if any(erring_channel_msg % r['channel'] in error_logs for r in route_mid):
+                continue
+            route = [route_out] + route_mid + [route_in]
+            setup_routing_fees(plugin, route, msatoshi)
+            fees = route[0]['msatoshi'] - route[-1]['msatoshi']
+            if fees > int(exemptfee) and fees > msatoshi * float(maxfeepercent) / 100:
+                continue
```
rustyrussell: This will hard-loop until timeout.  You need to exclude something (most expensive channel?).  See how the pay plugin does it, perhaps?

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
...
+                if erring_channel_msg % incoming_channel_id in str(e):
+                    return "Error with incoming channel"
+                if erring_channel_msg % outgoing_channel_id in str(e):
+                    return "Error with outgoing channel"
+    except Exception as e:
+        plugin.log("Exception: " + str(e))
+    plugin.rpc.delinvoice(label, "unpaid")
+    msg = "Rebalance failed"
+    plugin.log(msg)
+    return msg
```
rustyrussell: You should raise an exception on failure.  That seems to be the only way to make the python library give an error to the `rebalance` call, which is what I'd expect in this case.

contrib/plugins/rebalance.py
gallizoltan: Maybe I misinterpreted its behavior. I saw a successful rebalance once, but a few hours later `getroute` raised an exception for the same node ids, same parameters. I assumed this is `getroute`'s failure. But other explanations are possible too.

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+from lightning import Plugin, RpcError
+import time
+
+plugin = Plugin(autopatch=True)
+
+
+def setup_routing_fees(plugin, route, msatoshi):
+    delay = 9
```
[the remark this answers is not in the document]
gallizoltan: I wrote a `plugin.add_option` for this with a default value, and I load it from `listconfigs` at init. But not sure I use `plugin.options` properly.

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
...
+    my_node_id = plugin.rpc.getinfo().get('id')
+    outgoing_node_id = peer2channel(plugin, outgoing_channel_id, my_node_id)
+    incoming_node_id = peer2channel(plugin, incoming_channel_id, my_node_id)
+    plugin.log("Outgoing node: %s, channel: %s" % (outgoing_node_id, outgoing_channel_id))
+    plugin.log("Incoming node: %s, channel: %s" % (incoming_node_id, incoming_channel_id))
+
+    route_out = {'id': outgoing_node_id, 'channel': outgoing_channel_id}
+    route_in = {'id': my_node_id, 'channel': incoming_channel_id}
+    start_ts = int(time.time())
+    label = "Rebalance" + str(start_ts)
```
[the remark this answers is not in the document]
gallizoltan: uuid.uuid4() would be okay?

gallizoltan:
@rustyrussell Thank you for the review! I tried to address all the mentioned issues. Please take a look at the newest version!

contrib/plugins/rebalance.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+from lightning import Plugin, RpcError
+import time
+
+plugin = Plugin(autopatch=True)
```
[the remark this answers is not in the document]
cdecker: Added that to my cleanups :+1:

gallizoltan:
Good idea, I've just opened a new PR here: https://github.com/lightningd/plugins/pull/6

Should I close this one as obsolete?

cdecker:
Just merged  lightningd/plugins#6, thanks @gallizoltan :wink:

PR #2573, opened 2019-04-15
Plugin hook: pass through errors

rustyrussell:
This is how I think it should work: return a JSON RPC error if you want to fail the hook.

I'm not happy with then pylightning changes though: exceptions should be for exceptional events, and I'm not sure that rejecting these hooks counts as an exceptional event (since that's their main purpose!).

But doing otherwise would mean returning an explicit 'result' or 'error' dict, which would be best done by providing helpers to ensure they don't return malformed JSON responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Note: You should always set `allow-deprecated-apis=false` to test for
 changes.
 
 - JSON API: `newaddr` output field `address`: use `bech32` or `p2sh-segwit` instead.
-- Plugins: the `peer_connected` hook should return `result` directly as a string: `disconnect` or `continue`, not a nested in a `result` sub-object.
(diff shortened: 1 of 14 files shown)


cdecker:
> But doing otherwise would mean returning an explicit 'result' or 'error' dict, which would be best done by providing helpers to ensure they don't return malformed JSON responses.

I don't quite understand the problem here. Yes, we have two nested dicts going over the wire: the JSON-RPC response, and the inner result. That is desired and IIRC the outer dict gets stripped before calling the deserialization callback and the hook callback, so we don't see it anyway.

rustyrussell:
Thanks!

There are two issues here: how it should look at the JSON-RPC layer, and how in Python. I wasn't delighted with it either.

I will close this and try again :)

PR #2575, opened 2019-04-16
Plugin API refinement

rustyrussell:
This adds some minor functionality, and cleans up my mistake about the invoice API.

I this the simplification of removing the deserialization step is worthwhile, and I included it because my new fundchannel hook assumes it, but I remove that part if you really dislike it.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2575/commits/c58cbd6e68309d5c08503fc2cd441b564c11d3f2

PR #2578, opened 2019-04-16
lightningd: make sure we correctly clear connections from connectd on error

rustyrussell:
Without this, the connect command hangs in one of my branches.  This logic
is from the old days when gossipd handled connections, and we wanted
to make sure it didn't hang up on this client due to the error.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2578/commits/99050e44e1737f82da0554e66f7b4d46c5688f1d

PR #2577, opened 2019-04-16
lightningd: don't abort if funding locks in during shutdown.

rustyrussell:
For me this happened only under valgrind with
test_option_upfront_shutdown_script (in a pending branch):

==5063==    by 0x51FC076: raise (raise.c:48)
==5063==    by 0x51DD534: abort (abort.c:79)
==5063==    by 0x1292D2: fatal (log.c:647)
==5063==    by 0x116570: channel_set_state (channel.c:340)
==5063==    by 0x116E04: lockin_complete (channel_control.c:73)
==5063==    by 0x116F15: peer_got_funding_locked (channel_control.c:108)
==5063==    by 0x117354: channel_msg (channel_control.c:208)

No CHANGELOG: this was introduced in a recent refactor.

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

PR #2579, opened 2019-04-16
option_upfront_shutdown_script support

rustyrussell:
First commit is the bugfix from #2578 without which the test hangs


rustyrussell:
This also triggers the bug fixed by #2577 so will re-run tests after that is merged...

cdecker:
Rebased to address changelog conflicts

rustyrussell:
Nice catch, turns out an unrelated race though.  Will fix separately...

rustyrussell:
Since Christian Acked before, only to remove it because of separate now-fixed issue, I'm going to self-ack:

Ack https://github.com/ElementsProject/lightning/pull/2579/commits/13272fe09b7323f1f39028362f30cd87dbbf484c

PR #2580, opened 2019-04-16
channeld: give some tolerance for empty commitments.

rustyrussell:
The spec says not to send a commitment_signed without any changes, but LND does this.  To understand why, you have to understand how LND works.  I haven't read the code, but I'm pretty sure it works like this:

1. lnd slows down to do garbage collection, because it's in Go.
2. When an alert timer goes off, noticing it's not making process, it sends a twitter message to @roasbeef.
3. @roasbeef sshs into the user's machine and binary patches lnd to send a commitment_signed message.
4. Unfortunately he works so fast that various laws of causality are broken, meaning sometimes the commitment_signed is sent before any of these other things happen.

I'm fairly sure that this will stop as @roasbeef ages, or lnd introduces some kind of causality enforcement fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ changes.
 - `--bind-addr=<path>` fixed for nodes using local sockets (eg. testing).
 - Unannounced local channels were forgotten for routing on restart until reconnection occurred.
 - lightning-cli: arguments containing `"` now succeed, rather than causing JSON errors.
-- protocol: handle lnd sending more messages before `reestablish`; don't fail channel.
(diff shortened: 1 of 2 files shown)


rustyrussell:
I haven't seen it in recent lnds; pinging @niftynei who saw it last IIRC...

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2580/commits/ef5fdf83906ff3002dac91d282a55702bc37acf7

niftynei:
not off the top of my head, i'll keep an eye out tho.

PR #2581, opened 2019-04-16
pylightning: millisatoshis type is int

gallizoltan:
Millisatoshi's inner representation expected to be an int, otherwise unwanted exceptions could occur.
The following example raises: TypeError: __int__ returned non-int (type decimal.Decimal)

from lightning import Millisatoshi
one_sat = Millisatoshi("1sat")
two_sats = one_sat * 2


rustyrussell:
Good catch!

PR #2582, opened 2019-04-16
pytest: fix old code in test, which can cause spurious failures.

rustyrussell:
We generated blocks to announce the channel, but it can also expire
the HTLC if the timing is wrong.  We don't need to anyway, since we
fixed the FIXME; we store local unannounced channels for restoration


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2582/commits/13a73a97c36ee7c789693a65b261f9c34ac56c5a

PR #2585, opened 2019-04-17
JSON: Add a 'getbalance' command

darosior:
This PR adds the `getbalance` command to the JSON-RPC interface (along with a manpage and a test).

Why ?
- There is not any concise command to get the total funds managed by C-Lightning
- The only possibility is `listfunds` which you have to parse (or to `less` to at least see utxos)
- It is convenient for bitcoin-core users (same name)

### DIFF
--- a/doc/lightning-getbalance.7.txt
+++ b/doc/lightning-getbalance.7.txt
@@ -1,0 +1,43 @@
+LIGHTNING-GETBALANCE(7)
+======================
+:doctype: manpage
+
+NAME
+----
+lightning-getbalance - Command showing the total wallet balance.
+
+SYNOPSIS
+--------
+*getbalance*
+
+DESCRIPTION
+-----------
+The *getbalance* RPC command displays the total funds either on chain or locked on channels.
+
+RETURN VALUE
+------------
+On success an object will be returned with six entries :
+
+- 'on_chain' - The sum of all the wallet utxos, in satoshis.
+
+- 'on_chain_msat' - The sum of all the wallet utxos, in milli satoshis.
+
+- 'on_channels' - The sum of all our satoshis locked in channels, in satoshis.
+
+- 'on_channels_msat' - The sum of all our satoshis locked in channels, in milli satoshis (rounded).
+
+- 'total' - The sum of on chain and on channels funds, in satoshis.
+
+- 'total_msat' - The sum of on chain and on channels funds, in milli satoshis.
+
+AUTHOR
+------
+Darosior <darosior@protonmail.com> is mainly responsible.
+
+SEE ALSO
+--------
+lightning-listfunds(7), lightning-newaddr(7), lightning-fundchannel(7), lightning-withdraw(7)
+
+RESOURCES
+---------
+Main web site: https://github.com/ElementsProject/lightning
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -531,6 +531,68 @@ static const struct json_command listfunds_command = {
 };
 AUTODATA(json_command, &listfunds_command);
 
+static struct command_result *json_getbalance(struct command *cmd,
+					     const char *buffer,
+					     const jsmntok_t *obj UNNEEDED,
+					     const jsmntok_t *params)
+{
+    struct json_stream *response;
+    struct peer *p;
+    struct channel *c;
+    struct utxo **utxos;
+    struct amount_sat on_chain, on_channels, total;
+
+    if (!param(cmd, buffer, params, NULL))
+        return command_param_failed();
+
+    response = json_stream_success(cmd);
+    json_object_start(response, NULL);
+
+    /* Add funds from utxos */
+    on_chain = AMOUNT_SAT(0);
+    utxos = wallet_get_utxos(cmd, cmd->ld->wallet, output_state_available);
+    for (size_t i = 0; i < tal_count(utxos); i++) {
+        if (utxos[i]->blockheight && !utxos[i]->spendheight) {
+            if (!amount_sat_add(&on_chain, on_chain, utxos[i]->amount))
+                return command_fail(cmd, LIGHTNINGD,
+                        "utxos addition failure.");
+        }
+    }
+    json_add_amount_sat(response, on_chain, "on_chain", "on_chain_msat");
+
+    /* Add funds that are allocated to channels */
+    on_channels = AMOUNT_SAT(0);
+    list_for_each(&cmd->ld->peers, p, list) {
+        list_for_each(&p->channels, c, list) {
+            if(!amount_sat_add(&on_channels, on_channels,
+                    amount_msat_to_sat_round_down(c->our_msat)))
+                return command_fail(cmd, LIGHTNINGD,
+                        "channels allocated funds addition failure.");
+        }
+    }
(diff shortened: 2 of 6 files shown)


m-schmoock:
I **really** like the idea and will also make quick review.
Until then, how about adding `total` and `total_msat` to the response for the lazy math dudes ?

wallet/walletrpc.c
```diff
@@ -531,6 +531,58 @@ static const struct json_command listfunds_command = {
...
+            if (!amount_sat_add(&on_chain, on_chain, utxos[i]->amount))
+                return command_fail(cmd, LIGHTNINGD,
+                        "utxos addition failure.");
+        }
+    }
+    json_add_amount_sat(response, on_chain, "on_chain", "on_chain_msat");
+
+    /* Add funds that are allocated to channels */
+    list_for_each(&cmd->ld->peers, p, list) {
+        struct channel *c;
```
m-schmoock: you should move the declaration of `struct channel *c;` to the beginning of the function.

wallet/walletrpc.c
```diff
@@ -531,6 +531,58 @@ static const struct json_command listfunds_command = {
...
 
+static struct command_result *json_getbalance(struct command *cmd,
+					     const char *buffer,
+					     const jsmntok_t *obj UNNEEDED,
+					     const jsmntok_t *params)
+{
+    struct json_stream *response;
+    struct peer *p;
+    struct utxo **utxos;
+    struct amount_sat on_chain = AMOUNT_SAT(0), on_channels = AMOUNT_SAT(0);
```
m-schmoock: I would declare `struct amount_sat on_chain` and `struct amount_sat on_channels` without initializing them and init them not sooner than when required (line 551 and 561). This can help detect unwanted variable access by valgrind when someone else changes the code in the future.

doc/lightning-getbalance.7.txt
```diff
@@ -0,0 +1,39 @@
...
+======================
+:doctype: manpage
+
+NAME
+----
+lightning-getbalance - Command showing the total wallet balance.
+
+SYNOPSIS
+--------
+*getbalance*
```
m-schmoock: Trailing whitespace :D

doc/lightning-getbalance.7.txt
```diff
@@ -0,0 +1,39 @@
...
+
+- 'on_channels_msat' - The sum of all our satoshis locked in channels, in milli satoshis (rounded).
+
+AUTHOR
+------
+Darosior <darosior@protonmail.com> is mainly responsible.
+
+SEE ALSO
+--------
+lightning-listfunds(7), lightning-newaddr(7), lightning-fundchannel(7), lightning-withdraw(7)
```
m-schmoock: ... also trailing whitespace.

I really hate those when I see a git diff, maybe I am too picky.

darosior:
I first thought that it would be better to not mix up on chain funds and lightning funds, but now I can't find any good reason and it might be useful in some cases. I'm going to add it.

doc/lightning-getbalance.7.txt
```diff
@@ -0,0 +1,39 @@
...
+------------
+On success an object will be returned with four entries :
+
+- 'on_chain' - The sum of all the wallet utxos, in satoshis.
+
+- 'on_chain_msat' - The sum of all the wallet utxos, in milli satoshis.
+
+- 'on_channels' - The sum of all our satoshis locked in channels, in satoshis.
+
+- 'on_channels_msat' - The sum of all our satoshis locked in channels, in milli satoshis (rounded).
```
m-schmoock: As mentioned in the comments, we could sum it up for the user: `total` and `total_msat`

darosior:
Added it.

doc/lightning-getbalance.7.txt
darosior: No you are not, I just did not `git diff̀` the man page :/

darosior:
Addressed the changes, thank you for your review :smiley:

rustyrussell:
I want this!

We're generally steering away from more high-level information, in favor of plugins, which are much more flexible and customizable.

For example, my summary plugin does a similar thing: https://github.com/lightningd/plugins/blob/master/summary/README.md

m-schmoock:
Yes, Rusty is right, we should add this as a plugin, as it uses just info that is available via RPC anyway. Sorry for your time on the C code, it was in good shape.

Similarily to the example summary plugin, you can output in BTC and Fiat value as well instead of msat (hi@UX)

conscott:
I also wrote detailed balance plugin here: https://github.com/conscott/c-lightning-plugins/blob/master/balance.py

m-schmoock:
where should he commit such a basic plugin: `contrib/plugins/` ? This is currently just for examples, and the plugins are not loaded by default.

I think we should organize and ship some basic plugins, so potential contributors quicker get a clue about this.

darosior:
> I want this!
> 
> We're generally steering away from more high-level information, in favor of plugins, which are much more flexible and customizable.



> Yes, Rusty is right, we should add this as a plugin, as it uses just info that is available via RPC anyway. Sorry for your time on the C code, it was in good shape.

Ok, so I'll PR it as a plugin. It might have been useful to RPC users too but they can always compute it.

> Similarily to the example summary plugin, you can output in BTC and Fiat value as well instead of msat (hi@UX)

On which third party should we rely to get the price then ?

> where should he commit such a basic plugin: `contrib/plugins/` ? This is currently just for examples, and the plugins are not loaded by default.
> 
> I think we should organize and ship some basic plugins, so potential contributors quicker get a clue about this.

Yes. Do you think of other basic plugins ?

niftynei:
> where should he commit such a basic plugin: contrib/plugins/ ? This is currently just for examples, and the plugins are not loaded by default.

We're building a repo of c-lightning plugins here, PR's for new plugins welcome! https://github.com/lightningd/plugins

darosior:
Ok so I have to admit I'm a little bit confused.

> I want this!

@rustyrussell where should I PR this ? Here, in ElementsProject/plugins/ as `pay`, or in lightningd/plugins ?

m-schmoock:
@darosior since there is nobody answering by now, I'd suggest you add a python getbalance plugin PR to the plugin repo: https://github.com/lightningd/plugins

We can discuss shipping example plugins or restructuring stuff any time later.

I would like to see onchain BTC, offchain BTC, total and also Fiat value. Fiat Exchange rate (Eur/$) and currency unit (BTC, Sat, .. ) can be parameterized and should default to BTC and USD I think. I would use the Summary plugin as a codebase ;)

m-schmoock:
Perhaps also max_can_receive ln and max_can_send ln

darosior:
I think this would be overkill to add a getbalance python plugin since [summary](https://github.com/lightningd/plugins/tree/master/summary) already does it :
```
num_utxos=5
utxo_amount=1.20119332000btc (USD $4473.84)
num_channels=29
num_connected=2
num_gossipers=1
avail_out=0.27095103btc (USD $1009.16)
avail_in=2.05851379btc (USD $7666.93)
```
But I think it would be useful as a C plugin loaded by default like `pay`.

rustyrussell:
pay is kind of special since it was a built-in before it was a plugin. Other plugins may graduate to "must haves" over time, and we'd move them into the main repo.

But having them separate means they can evolve fast and don't have to wait for us to release a new c-lightning.

PR #2588, opened 2019-04-18
pylightning: Fixing incompatibility issue with Python 2.7

hsjoberg:
Hi!
I didn't understand if this is the home for pylightning, but here's a quick fix for making pylightning work with Python 2.7. It seems to be a regression from 0.0.7.

Attempting to install pylightning==0.0.7 with Python 2.7 will end up with the following error:

```
  File "setup.py", line 5, in <module>
    with open('README.md', encoding='utf-8') as f:
TypeError: 'encoding' is an invalid keyword argument for this function
```

The reason for the error is that the global scope open doesn't have the `encoding` argument in Python 2.7. The solution is to import the `open` function from the `io` package, which makes the script compatible with both Python 2 and 3.

This fix makes the installation of https://github.com/marzig76/fulmo less of a hassle.

Cheers

### DIFF
--- a/contrib/pylightning/setup.py
+++ b/contrib/pylightning/setup.py
@@ -1,8 +1,9 @@
 from setuptools import setup
 import lightning
+import io
 
 
-with open('README.md', encoding='utf-8') as f:
+with io.open('README.md', encoding='utf-8') as f:
     long_description = f.read()
 
 setup(name='pylightning',


cdecker:
Not sure that clobbering `open` this way is desirable. It'd be better to do it this way:

```
import io

with io.open('README.md', encoding='utf-8') as f:
```

hsjoberg:
@cdecker Done

cdecker:
ACK b7485f54620bc3508c62d0014554978dbf143369

PR #2592, opened 2019-04-24
check: Run `check-manpages` in /tmp/ instead of the user config

cdecker:
If we have plugins configured in the default config location, then
`check-manpages` may fail due to plugin cli options being added to the
`--help` output.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -71,7 +71,7 @@ check: check-manpages
 
 check-manpages: cli/lightning-cli lightningd/lightningd
 	@tools/check-manpage.sh cli/lightning-cli doc/lightning-cli.1.txt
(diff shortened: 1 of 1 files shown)


cdecker:
> is this the right level for this fix? would it be better to fixup check-manpages and our help output so that it's easier to weed out the plugin commands, like we do for the --dev ones?

The main problem is that, unlike the `--dev` options, the plugin options may assume any name that isn't already assigned, which makes it really hard to recognize them as false positives for the detection. This change just makes sure(-ish) that we start with a pristine config instead of analyzing the developer's config :wink:

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2592/commits/9a5f4cddea9c76996a06cf9daa36bef80f1ba3d1

PR #2593, opened 2019-04-25
Fix bockheight underflow if `minconf` > `blockheight`

### LINKED ISSUE #2518: withdraw doesn't respect minconf parameter
`lightning-cli withdraw <address> 10000 normal 99999999` broadcasts a withdrawal tx, when no output is at that confirmation depth.

cdecker:
We are computing a `maxheight` that coins must have to be considered for coin selection by subtracting the current `minconf` parameter from the current `blockheight`. If the `minconf` parameter is larger than `blockheight` we were wrapping around and suddenly accepting all transactions with at least one confirmation.

We now clamp the `maxheight` requirement to `1` (since `0` is used to indicate no confirmations required), which is safe since we can't end up having an output at that height, resulting in us not selecting any coins which is desired in this case.

Fixes #2518

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -482,7 +482,7 @@ def test_withdraw(node_factory, bitcoind):
     with pytest.raises(RpcError, match=r'Cannot afford transaction'):
         l1.rpc.withdraw(waddr, 'all')
 
-@pytest.mark.xfail(strict=True)
+
 def test_minconf_withdraw(node_factory, bitcoind):
     """Issue 2518: ensure that ridiculous confirmation levels don't overflow
 
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -482,6 +482,31 @@ def test_withdraw(node_factory, bitcoind):
     with pytest.raises(RpcError, match=r'Cannot afford transaction'):
         l1.rpc.withdraw(waddr, 'all')
 
+@pytest.mark.xfail(strict=True)
+def test_minconf_withdraw(node_factory, bitcoind):
+    """Issue 2518: ensure that ridiculous confirmation levels don't overflow
+
+    The number of confirmations is used to compute a maximum height that is to
+    be accepted. If the current height is smaller than the number of
+    confirmations we wrap around and just select everything. The fix is to
+    clamp the maxheight parameter to a positive small number.
+
+    """
+    amount = 1000000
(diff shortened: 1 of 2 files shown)


tests/test_misc.py
```diff
@@ -483,6 +483,32 @@ def test_withdraw(node_factory, bitcoind):
...
+    # Add some funds to withdraw later
+    for i in range(10):
+        l1.bitcoin.rpc.sendtoaddress(addr, amount / 10**8 + 0.01)
+
+    bitcoind.generate_block(1)
+
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == 10)
+    with pytest.raises(RpcError):
+        l1.rpc.withdraw(destination=addr, satoshi=10000, feerate='normal', minconf=9999999)
+    bitcoind.generate_block(1)
```
niftynei: why generate a block here if we've already finished the test script?

practicalswift:
Good find!

Travis testing with `-fsanitize=integer` could perhaps help us catch _unintended_ wraparounds like this?

cdecker:
> Travis testing with -fsanitize=integer could perhaps help us catch unintended wraparounds like this?

As far as I know that would require us to run yet another CI configuration, since it is unlikely we can run this with the address sanitizer, or are the two compatible?

practicalswift:
@cdecker They are compatible. We could run `-fsanitize=address,integer,undefined` in one job as we do for Bitcoin Core: https://github.com/bitcoin/bitcoin/blob/master/.travis.yml#L165

Please note that `-fsanitize=integer` will require suppressions to cover the *intentional* unsigned integer wraparounds (typically found in crypto/hashing code) but I think that is a fairly low price to pay to get immediate notification in case of introduction of *unintentional* unsigned integer wraparounds.

See the `unsigned-integer-overflow` suppressions in the Bitcoin Core project: https://github.com/bitcoin/bitcoin/blob/master/test/sanitizer_suppressions/ubsan

cdecker:
Thanks for the hint @practicalswift I'll look into enabling the sanitizers on Travis :-)

Updated the PR to fix the extra block @niftynei

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2593/commits/3222d7c50e59793374f316b42df82a432f7935e5

PR #2594, opened 2019-04-25
remove libbase58, use base58 from libwally

greenaddress:
This removes libbase58 and uses libwally instead.

I also removed some functions from the base58 header that weren't implemented.

It allocates and then frees some memory, we may want to
add a function in wally that assumes caller allocates or override
wally_operations to use tal.

Should the above be a FIXME for later and if not which path forward?


bitcoin/base58.c
```diff
@@ -53,16 +51,22 @@ static bool from_base58(u8 *version,
...
-	b58_sha256_impl = my_sha256;
-
 	size_t buflen = sizeof(buf);
-	b58tobin(buf, &buflen, base58, base58_len);
-
-	int r = b58check(buf, buflen, base58, base58_len);
+	memset(buf, 0, buflen);
+	char *terminated_base58 = malloc(base58_len + 1);
+	memcpy(terminated_base58, base58, base58_len);
+	terminated_base58[base58_len] = '\0';
```
cdecker: This might be a candidate for `tal_arr_dup(NULL, char, base58_len, 1)` which allocates `base58_len + 1` and copies. You still need to set the terminating `\0`.

bitcoin/base58.c
```diff
@@ -12,26 +12,24 @@
...
 
-	b58_sha256_impl = my_sha256;
-	if (!b58check_enc(out, &outlen, version, rmd, sizeof(*rmd))) {
+	if (wally_base58_from_bytes((const unsigned char *) buf, total_length, BASE58_FLAG_CHECKSUM, &out) != WALLY_OK) {
 		return NULL;
 	}else{
-		return tal_strdup(ctx, out);
+		char *res;
+		res = tal_strdup(ctx, out);
+		return wally_free_string(out) == WALLY_OK ? res : NULL;
```
cdecker: This may be considered a leak by the leak detector since we return `NULL` without freeing `out`.

bitcoin/base58.c
```diff
@@ -53,16 +51,22 @@ static bool from_base58(u8 *version,
...
 	/* Avoid memcheck complaining if decoding resulted in a short value */
-	memset(buf, 0, sizeof(buf));
-	b58_sha256_impl = my_sha256;
-
 	size_t buflen = sizeof(buf);
-	b58tobin(buf, &buflen, base58, base58_len);
-
-	int r = b58check(buf, buflen, base58, base58_len);
+	memset(buf, 0, buflen);
+	char *terminated_base58 = malloc(base58_len + 1);
```
cdecker: Same as above, I'd use `tal_dup_arr` here as well.

bitcoin/base58.c
```diff
@@ -53,16 +51,22 @@ static bool from_base58(u8 *version,
...
+	memcpy(terminated_base58, base58, base58_len);
+	terminated_base58[base58_len] = '\0';
+
+	size_t written = 0;
+	int r = wally_base58_to_bytes(terminated_base58, BASE58_FLAG_CHECKSUM, buf, buflen, &written);
+	wally_bzero(terminated_base58, base58_len + 1);
+	free(terminated_base58);
```
cdecker: `tal_free`

bitcoin/base58.c
```diff
@@ -53,16 +51,22 @@ static bool from_base58(u8 *version,
...
-
-	int r = b58check(buf, buflen, base58, base58_len);
+	memset(buf, 0, buflen);
+	char *terminated_base58 = malloc(base58_len + 1);
+	memcpy(terminated_base58, base58, base58_len);
+	terminated_base58[base58_len] = '\0';
+
+	size_t written = 0;
+	int r = wally_base58_to_bytes(terminated_base58, BASE58_FLAG_CHECKSUM, buf, buflen, &written);
+	wally_bzero(terminated_base58, base58_len + 1);
```
cdecker: Is this necessary? Seems like we don't really store any secret in here.

greenaddress: Fair enough - I removed it - it seems to me at the moment this is only used for addresses.

bitcoin/base58.c
greenaddress: Done

bitcoin/base58.c
greenaddress: Done

bitcoin/base58.c
greenaddress: Is this ok now?

bitcoin/base58.c
greenaddress: done

bitcoin/base58.c
```diff
@@ -27,9 +27,10 @@ static char *to_base58(const tal_t *ctx, u8 version,
 	if (wally_base58_from_bytes((const unsigned char *) buf, total_length, BASE58_FLAG_CHECKSUM, &out) != WALLY_OK) {
 		return NULL;
 	}else{
-		char *res;
+		char *res = NULL;
```
cdecker: Rusty won't like this since it interferes with `valgrind`s ability to tell us about accesses to uninitialized fields.

greenaddress: @cdecker fair enough! updated in https://github.com/ElementsProject/lightning/pull/2594/commits/dc413ba6f043254115e7b8a84f4e1aed1ed5bcaa - let me know when you want me to rebase or squash

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2594/commits/dc413ba6f043254115e7b8a84f4e1aed1ed5bcaa

PR #2601, opened 2019-04-29
fix: pylightning msat round multiply and division

m-schmoock:
Currently, when a multiplication operator is invoked that
does not result in an even integer result but a floating result,
the pylightning code will raise an exception:

Millisatoshi must be string with msat/sat/btc suffix or int

This is because the internal float result will be used as
contructor argument like this:  return Millisatoshi(10000.5)

This happens especially on fee calculations where small uneven amounts
are calculated.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2601/commits/7c27b3f1d110ec0e89c906a87e31eccc763e5254

PR #2608, opened 2019-04-30
Simplify address parsing

cdecker:
This PR picks up some work by @bisoge in PR #1850 which I found useful. It now has a clearer separation between address parsing and deciding whether an address matches our current chain. I took the liberty of cherry-picking a bit and inverting the way we check for chain compatibility, but the inspiration came from PR #1850.

Closes #1850


bitcoin/base58.c
```diff
@@ -67,40 +67,24 @@ static bool from_base58(u8 *version,
 	return true;
 }
 
-bool bitcoin_from_base58(bool *test_net,
-			 struct bitcoin_address *addr,
+bool bitcoin_from_base58(u8 *version, struct bitcoin_address *addr,
```
niftynei: is it worth preserving this method (and the ones below) if they're just a wrapper for another?

niftynei:
you've got a build error tho: `external/libbase58 is not a submodule!`

rustyrussell:
> you've got a build error tho: `external/libbase58 is not a submodule!`

Some weird Travis caching thing.  Blowing away the cache seems to have fixed (hit this on another PR)

bitcoin/base58.c
cdecker: I was tempted to remove them, however the type-safety that this method gives (`struct bitcoin_address` wraps `struct ripemd160`) is worth it I think.

PR #2610, opened 2019-05-01
gossipd: don't spam with update messages.

rustyrussell:
On my node, this queues over 1M messages to lightningd, and that's with only 10 peers,
nailing the CPU for quite a while.

That could be fixed by making our queue scale, but we shouldn't be hitting this!


cdecker:
Need to also fix up the tests that were waiting on the message appearing in the log. I'd suggest polling for the channel using `listchannels` instead, might be a bit slower but also makes the order of activation less of an issue.

PR #2614, opened 2019-05-02
status: suppress status_trace/status_debug messages if queue too long.

rustyrussell:
We can be spammy, which is good for tests, but bad for our simple message queue.
This avoids breaking our tests but also avoid the worst case (1M entries and counting!)
for gossip status messages in the case where we're syncing a large peer.

[ This is less invasive than #2610 but in some ways less elegant.  Fixing those tests was non-trivial, and would have involved doing IO tracing for all of them... ]

Closes: #2610

### DIFF
--- a/common/daemon_conn.c
+++ b/common/daemon_conn.c
@@ -146,3 +146,8 @@ void daemon_conn_wake(struct daemon_conn *dc)
 {
 	msg_wake(dc->out);
 }
+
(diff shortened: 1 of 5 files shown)


cdecker:
I had a partial fix for the tests that was polling for channel state changes rather than waiting for the log lines, but yeah, for some tests it proved to be more difficult than I expected.

PR #2615, opened 2019-05-02
pytest::test_pay_direct fix the check for route randomization

SimonVrouwe:
Travis failed on this [test](https://travis-ci.org/ElementsProject/lightning/jobs/527005543) by (what appears to be) chance while testing #2497.
Fixes what appears to be a minor misinterpretation of route randomization.

### DIFF
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -1411,21 +1411,23 @@ def test_pay_direct(node_factory, bitcoind):
     # and gossip 2 minutes to propagate
     l0.wait_for_channel_updates([c0, c1, c2, c3])
 
-    # Find out how much msatoshi l1 owns on l1->l2 channel.
-    l1l2msatreference = only_one(l1.rpc.getpeer(l2.info['id'])['channels'])['msatoshi_to_us']
+    # Try multiple times because route randomization
+    # can override our preference for direct route
+    n_direct = 0
+    for i in range(10):
+        # Find out how much msatoshi l1 owns on l1->l2 channel.
+        l1l2msatreference = only_one(l1.rpc.getpeer(l2.info['id'])['channels'])['msatoshi_to_us']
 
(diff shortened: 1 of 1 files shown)


cdecker:
Well, that test is explicitly to make sure that we don't randomize if we have a direct channel that can be used for the transfer. If we are hitting the randomization then this is a true-positive and we should address it root cause.

SimonVrouwe:
Sorry, I got confused with the invoice's route-hint randomization. Maybe I broke it myself that put the direct channel unusable.

However when I try to reproduce the failure on master e3bac6c by changing in the test this:
```
    for i in range(8):
        inv = l3.rpc.invoice(20000000, 'pay{}'.format(i), 'desc')['bolt11']
```
to this:
```
    for i in range(80):
        inv = l3.rpc.invoice(2000000, 'pay{}'.format(i), 'desc')['bolt11']
```
then the test fails

SimonVrouwe:
https://github.com/ElementsProject/lightning/blob/47b5f2e83756e24ef4018c020b2abee42ecff788/tests/test_pay.py#L1667-L1680

@cdecker Just to be clear, should the outcome of this test be independent of the payment `amount`?
Because it is not, when amount is reduced by factor 100 (to 200000msat) it fails.

SimonVrouwe:
Sorry, I shouldn't reopen this PR, its just a question.

ZmnSCPxj:
>Just to be clear, should the outcome of this test be independent of the payment `amount`?

Yes.

Users expect that direct channel is reliable, so taking any nondirect route when direct channel exists risks unreliability and delay (due to involving more nodes).
Thus the we *must* strongly prefer direct route in order to maintain UX.
Direct route also has better privacy, and main purpose of route randomization is privacy (nodes with lower forwarding fees are likely to be spy nodes trying to get you to route through them, so we should discount fees, hence route randomization).

PR #2617, opened 2019-05-03
More gossipd optimizations and cleanups.

rustyrussell:
Feel free to argue with these; there are several other things I tried which didn't make a significant performance difference.  The next thing to optimize is really the towire_ routines which reallocate all the time, but that's quite invasive.

### DIFF
--- a/common/read_peer_msg.h
+++ b/common/read_peer_msg.h
@@ -71,7 +71,9 @@ bool handle_peer_gossip_or_error(int peer_fd, int gossip_fd, int gossip_store_fd
 				 const u8 *msg TAKES);
 
 /* We got this message from gossipd: forward/quit as it asks. */
-void handle_gossip_msg(int peer_fd, struct crypto_state *cs,
+void handle_gossip_msg(int peer_fd,
+		       int gossip_store_fd,
+		       struct crypto_state *cs,
 		       const u8 *msg TAKES);
 
 /**
--- a/common/read_peer_msg.h
+++ b/common/read_peer_msg.h
@@ -57,7 +57,7 @@ bool is_wrong_channel(const u8 *msg, const struct channel_id *expected,
 
 /**
  * handle_peer_gossip_or_error - simple handler for all the above cases.
- * @peer_fd, @gossip_fd: peer and gossip fd.
+ * @peer_fd, @gossip_fd, @gossip_store_fd: peer, gossip and gossip_store fds.
  * @cs: the cryptostate (updated)
  * @msg: the peer message (only taken if returns true).
  *
@@ -65,7 +65,7 @@ bool is_wrong_channel(const u8 *msg, const struct channel_id *expected,
  * to gossipd), an error packet (causes peer_failed_received_errmsg or
  * ignored), or a message about the wrong channel (sends sync error reply).
  */
-bool handle_peer_gossip_or_error(int peer_fd, int gossip_fd,
+bool handle_peer_gossip_or_error(int peer_fd, int gossip_fd, int gossip_store_fd,
 				 struct crypto_state *cs,
 				 const struct channel_id *channel_id,
 				 const u8 *msg TAKES);
--- a/common/read_peer_msg.h
+++ b/common/read_peer_msg.h
@@ -74,4 +74,10 @@ bool handle_peer_gossip_or_error(int peer_fd, int gossip_fd, int gossip_store_fd
 void handle_gossip_msg(int peer_fd, struct crypto_state *cs,
 		       const u8 *msg TAKES);
 
+/**
+ * new_gossip_store - handle replacement gossip_store_fd.
+ * @gossip_store_fd: our fixed fd we expect to use to read gossip_store.
+ * @new_gossip_store_fd: fd received from gossipd.
+ */
+void new_gossip_store(int gossip_store_fd, int new_gossip_store_fd);
 #endif /* LIGHTNING_COMMON_READ_PEER_MSG_H */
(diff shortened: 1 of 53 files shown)


common/read_peer_msg.h
```diff
@@ -57,21 +57,29 @@ bool is_wrong_channel(const u8 *msg, const struct channel_id *expected,
 
 /**
  * handle_peer_gossip_or_error - simple handler for all the above cases.
- * @peer_fd, @gossip_fd: peer and gossip fd.
+ * @peer_fd, @gossip_fd, @gossip_store_fd: peer, gossip and gossip_store fds.
```
trueptolemy: `channel_id` is missed here...a small corner that originally existed..

lightningd/peer_comms.c
```diff
@@ -0,0 +1,20 @@
...
+{
+	if (pcomms->peer_fd != -1)
+		close(pcomms->peer_fd);
+	if (pcomms->gossip_fd != -1)
+		close(pcomms->gossip_fd);
+	if (pcomms->gossip_store_fd != -1)
+		close(pcomms->gossip_store_fd);
+}
+
+struct peer_comms *new_peer_comms(const tal_t *ctx)
```
trueptolemy: here should we initialize `fd`s with -1 for safty?

common/read_peer_msg.h
rustyrussell: Yep, indeed.

lightningd/peer_comms.c
rustyrussell: I'd be tempted to, but we're literally freeing the memory at the end of this call.

gossipd/gossipd.c
```diff
@@ -2105,45 +2132,73 @@ static struct io_plan *getchannels_req(struct io_conn *conn,
...
+						  c, source);
 			}
 		}
 	} else {
 		u64 idx;
 
+		/* Starting guess at 4 channels per node. */
+		entries = tal_arr(tmpctx,
+				  struct gossip_getchannels_entry,
+				  (daemon->rstate->nodes->raw.elems + 1) * 4);
```
cdecker: This now has enough magic in it that it might be worth it just to iterate through the map twice, wouldn't it? That way we'd get an exact count, could directly allocate the correct size and wouldn't need to overallocate and to re-allocate if our over-allocation was too small anyway.

gossipd/broadcast.h
```diff
@@ -37,9 +37,13 @@ struct broadcast_state *new_broadcast_state(struct routing_state *rstate,
...
 
 /* Append a queued message for broadcast.  Must be explicitly deleted.
- * Also adds it to the gossip store. */
+ * Also adds it to the gossip store.
+ *
+ * If it's a channel_announcement, channel_announce_sat must be set.
+ */
 void insert_broadcast(struct broadcast_state **bstate,
 		      const u8 *msg,
+		      const struct amount_sat *channel_announce_sat,
```
cdecker: I have a bit of an aversion against this kind of argument whose existence/value depend on other arguments.

cdecker: I do agree that freeing `gossip_store` from having to look up external information is good here, but since we're driving this from outside anyway, why not just have the caller build the channel amount msg and store that?

rustyrussell:
> I really like the reduction in allocations that is achieved in the first commits, and the inversion of control when it comes to the channel amount for the `gossip_store` (though I'd like it taken one step further and drive the channel amount msg being written from outside). You will have guessed that I'm a bit skeptical when it comes to handing yet more file descriptors around and add more complexity by pushing the responsibility of reading fetching gossip messages to the peer-daemons.

We could have them simply open the gossip_store themselves, but that affects seccomp, so I handed an fd.

There's some internal gossipd pollution in the gossip_store, which is unfortunate, but that could be moved out to a separate gossipd db (ie. local channel information).

>     * [a1d0043](https://github.com/ElementsProject/lightning/commit/a1d00439fb154a8b0a14e085f2a56a2fd2b1a388) unwrapping the gossip store
>       messages into raw gossip messages pushes logic into the gossip store
>       itself, which is really ugly. Dependencies between messages are
>       messy. That's what got us the whole gossip message ordering mess in the
>       first place.

Well, we already comingle raw gossip and gossipd messages on the peer<->gossipd wire; the wrapping seemed spurious.  And this commit doesn't affect ordering, it just leaves them raw to avoid a level of indirection.

>     * [ea62450](https://github.com/ElementsProject/lightning/commit/ea6245050577207b5eefb4967dac5ea2d03eb85a) is a nice cleanup +1
> 
>     * [5154057](https://github.com/ElementsProject/lightning/commit/51540573735940e1b0be1616bdfc16a139065938) seems like it's sacrificing a lot
>       of the architecture at the temple of performance. It adds quite a bit of
>       complexity (compaction, tighter coupling, more fds being passed around,
>       ...) for performance. I feel this is premature optimization, for little
>       gain at the moment, and the gossip is one part where I definitely think we
>       don't need more complexity. We should rather be focusing on making our own
>       sync algorithm smarter rather than concentrating on single peer full sync.

If it were merely an optimization, I would agree.  But it's also what gossipd spends many cycles doing, much of that that's now moved out to the peer, which means we can limit it appropriately.  I have a follow-up series which completes this move, and so gossipd doesn't do any streaming or filtering *at all* for normal gossip, which is even better (and saves 130MB of broadcast map).

gossipd/gossipd.c
rustyrussell: I actually needed to do more work here; by jamming the entire response into lightningd at once, it ends up trying to allocate 1GB for the JSON, which does not work on the rPi I have, crashing lightningd.

The patch to make it batch is fairly simple, has nice performance properties (we can process in parallel while gossipd is generating replies) and that supercedes this anyway.  I'll drop it.

gossipd/broadcast.h
rustyrussell: I have done this, but it's not as trivially neat as I'd like.  Now routing.c creates a gossip_store msg, so I don't think we won.

The alternative was to append the channel size to the channel_announcement itself, and have everyone "know" to remove that.  Neither is perfect.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/c6d983a5e4838179bac89311156f6e492e07773b

PR #2619, opened 2019-05-05
DB: Store the signatures of channel announcement sent from remote peer into DB

### LINKED ISSUE #2409: Store the `announcement_signature` in the database to work around some impls not sending it on reconnect

trueptolemy:
Fixes #2409 
Here I add a new wire type:
**channel_got_announcement (1017)**

When Channeld receives the remote channel announcement signatures, it will generate a msg1 like:
**WIRE_CHANNEL_GOT_ANNOUNCEMENT + remote_ann_node_sig + remote_ann_bitcoin_sig**
and send this msg to Master. (After sending signatures, Channeld will be holding until receive the reply from Master.)
Master resolves msg and stores the signatures into DB. 

Like issue #2409 , the adventage is when we reenable our channel, you can init it with the remote announcement signatures from DB directly other than wasting your time to wait for remote peer's announce reply.

### DIFF
--- a/channeld/channel_wire.csv
+++ b/channeld/channel_wire.csv
@@ -64,6 +64,8 @@ channel_init,,lflen,u16
 channel_init,,localfeatures,lflen*u8
 channel_init,,upfront_shutdown_script_len,u16
 channel_init,,upfront_shutdown_script,upfront_shutdown_script_len*u8
+channel_init,,remote_ann_node_sig,?secp256k1_ecdsa_signature
+channel_init,,remote_ann_bitcoin_sig,?secp256k1_ecdsa_signature
 
 # master->channeld funding hit new depth(funding locked if >= lock depth)
 channel_funding_depth,1002
--- a/channeld/channel_wire.csv
+++ b/channeld/channel_wire.csv
@@ -185,3 +185,8 @@ channel_fail_fallen_behind,,remote_per_commitment_point,struct pubkey
 channel_specific_feerates,1029
 channel_specific_feerates,,feerate_base,u32
 channel_specific_feerates,,feerate_ppm,u32
+
+# When we receive announcement_signatures for channel announce
+channel_got_announcement,1017
+channel_got_announcement,,remote_ann_node_sig,secp256k1_ecdsa_signature
+channel_got_announcement,,remote_ann_bitcoin_sig,secp256k1_ecdsa_signature
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -2842,6 +2842,8 @@ static void init_channel(struct peer *peer)
 	u32 feerate_per_kw[NUM_SIDES];
 	u32 minimum_depth;
 	struct secret last_remote_per_commit_secret;
+	secp256k1_ecdsa_signature *remote_ann_node_sig;
+	secp256k1_ecdsa_signature *remote_ann_bitcoin_sig;
 
 	assert(!(fcntl(MASTER_FD, F_GETFL) & O_NONBLOCK));
 
@@ -2896,7 +2898,9 @@ static void init_channel(struct peer *peer)
 				   &peer->announce_depth_reached,
 				   &last_remote_per_commit_secret,
 				   &peer->localfeatures,
-				   &peer->remote_upfront_shutdown_script)) {
+				   &peer->remote_upfront_shutdown_script,
+				   &remote_ann_node_sig,
+				   &remote_ann_bitcoin_sig)) {
 					   master_badmsg(WIRE_CHANNEL_INIT, msg);
 	}
 
@@ -2914,6 +2918,18 @@ static void init_channel(struct peer *peer)
 		     peer->revocations_received,
 		     feerate_per_kw[LOCAL], feerate_per_kw[REMOTE],
 		     peer->feerate_min, peer->feerate_max);
+
+	if(remote_ann_node_sig && remote_ann_bitcoin_sig) {
+		peer->announcement_node_sigs[REMOTE] = *remote_ann_node_sig;
+		peer->announcement_bitcoin_sigs[REMOTE] = *remote_ann_bitcoin_sig;
+		peer->have_sigs[REMOTE] = true;
+
+		/* Before we store announcement into DB, we have made sure
+		 * remote short_channel_id matched the local. Now we initial
+		 * it directly!
+		 */
+		peer->short_channel_ids[REMOTE] = peer->short_channel_ids[LOCAL];
+	}
 
 	/* First commit is used for opening: if we've sent 0, we're on
 	 * index 1. */
@@ -2959,6 +2975,8 @@ static void init_channel(struct peer *peer)
 	tal_free(fulfilled_sides);
 	tal_free(failed);
 	tal_free(failed_sides);
+	tal_free(remote_ann_node_sig);
+	tal_free(remote_ann_bitcoin_sig);
 
 	peer->channel_direction = node_id_idx(&peer->node_ids[LOCAL],
 					      &peer->node_ids[REMOTE]);
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -451,8 +451,6 @@ static void announce_channel(struct peer *peer)
 {
 	u8 *cannounce;
 
-	check_short_ids_match(peer);
-
 	cannounce = create_channel_announcement(tmpctx, peer);
 
 	wire_sync_write(GOSSIP_FD, cannounce);
@@ -493,6 +491,10 @@ static void channel_announcement_negotiate(struct peer *peer)
 	 *      has been sent and received AND the funding transaction has at least six confirmations.
  	 */
 	if (peer->announce_depth_reached && !peer->have_sigs[LOCAL]) {
+		/* When we reenable the channel, we will also send the announcement to remote peer, and
+		 * receive the remote announcement reply. But we will rebuild the channel with announcement
+		 * from the DB directly, other than waiting for the remote announcement reply.
+		 */
 		send_announcement_signatures(peer);
 		peer->have_sigs[LOCAL] = true;
 		billboard_update(peer);
@@ -500,8 +502,18 @@ static void channel_announcement_negotiate(struct peer *peer)
 
 	/* If we've completed the signature exchange, we can send a real
 	 * announcement, otherwise we send a temporary one */
-	if (peer->have_sigs[LOCAL] && peer->have_sigs[REMOTE])
+	if (peer->have_sigs[LOCAL] && peer->have_sigs[REMOTE]) {
+		check_short_ids_match(peer);
+
+		/* After making sure short_channel_ids match, we can send remote
+		 * announcement to MASTER. */
+		wire_sync_write(MASTER_FD,
+			        take(towire_channel_got_announcement(NULL,
+			        &peer->announcement_node_sigs[REMOTE],
+			        &peer->announcement_bitcoin_sigs[REMOTE])));
+
 		announce_channel(peer);
+	}
 }
 
 static void handle_peer_funding_locked(struct peer *peer, const u8 *msg)
@@ -2778,6 +2790,7 @@ static void req_in(struct peer *peer, const u8 *msg)
 	case WIRE_CHANNEL_GOT_COMMITSIG_REPLY:
 	case WIRE_CHANNEL_GOT_REVOKE_REPLY:
 	case WIRE_CHANNEL_GOT_FUNDING_LOCKED:
+	case WIRE_CHANNEL_GOT_ANNOUNCEMENT:
 	case WIRE_CHANNEL_GOT_SHUTDOWN:
 	case WIRE_CHANNEL_SHUTDOWN_COMPLETE:
 	case WIRE_CHANNEL_DEV_REENABLE_COMMIT_REPLY:
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -117,6 +117,25 @@ static void peer_got_funding_locked(struct channel *channel, const u8 *msg)
 		lockin_complete(channel);
 }
 
+static void peer_got_announcement(struct channel *channel, const u8 *msg)
+{
+	secp256k1_ecdsa_signature remote_ann_node_sig;
+	secp256k1_ecdsa_signature remote_ann_bitcoin_sig;
+
+	if (!fromwire_channel_got_announcement(msg,
+					       &remote_ann_node_sig,
+					       &remote_ann_bitcoin_sig)) {
+		channel_internal_error(channel,
+				       "bad channel_got_funding_locked %s",
+				       tal_hex(tmpctx, msg));
+		return;
+	}
+
+	wallet_announcement_save(channel->peer->ld->wallet, channel->dbid,
+				 &remote_ann_node_sig,
+				 &remote_ann_bitcoin_sig);
+}
+
 static void peer_got_shutdown(struct channel *channel, const u8 *msg)
 {
 	u8 *scriptpubkey;
@@ -218,6 +237,9 @@ static unsigned channel_msg(struct subd *sd, const u8 *msg, const int *fds)
 		break;
 	case WIRE_CHANNEL_GOT_FUNDING_LOCKED:
 		peer_got_funding_locked(sd->channel, msg);
+		break;
+	case WIRE_CHANNEL_GOT_ANNOUNCEMENT:
+		peer_got_announcement(sd->channel, msg);
 		break;
 	case WIRE_CHANNEL_GOT_SHUTDOWN:
 		peer_got_shutdown(sd->channel, msg);
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -297,6 +297,7 @@ void peer_start_channeld(struct channel *channel,
 	const struct config *cfg = &ld->config;
 	bool reached_announce_depth;
 	struct secret last_remote_per_commit_secret;
+	secp256k1_ecdsa_signature *remote_ann_node_sig, *remote_ann_bitcoin_sig;
 
 	hsmfd = hsm_get_client_fd(ld, &channel->peer->id,
 				  channel->dbid,
@@ -370,6 +371,13 @@ void peer_start_channeld(struct channel *channel,
 	if (ld->config.ignore_fee_limits)
 		log_debug(channel->log, "Ignoring fee limits!");
 
+	if(!wallet_remote_ann_sigs_load(tmpctx, channel->peer->ld->wallet, channel->dbid,
+				       &remote_ann_node_sig, &remote_ann_bitcoin_sig)) {
+		channel_internal_error(channel,
+				       "Could not load remote announcement signatures");
+		return;
+	}
+
 	initmsg = towire_channel_init(tmpctx,
 				      &get_chainparams(ld)->genesis_blockhash,
 				      &channel->funding_txid,
@@ -419,7 +427,9 @@ void peer_start_channeld(struct channel *channel,
 				      reached_announce_depth,
 				      &last_remote_per_commit_secret,
 				      channel->peer->localfeatures,
-				      channel->remote_upfront_shutdown_script);
+				      channel->remote_upfront_shutdown_script,
+				      remote_ann_node_sig,
+				      remote_ann_bitcoin_sig);
 
 	/* We don't expect a response: we are triggered by funding_depth_cb. */
 	subd_send_msg(channel->owner, take(initmsg));
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -382,6 +382,9 @@ static struct migration dbmigrations[] = {
 	{ "ALTER TABLE channels ADD remote_upfront_shutdown_script BLOB;", NULL },
 	/* PR #2524: Add failcode into forward_payment */
 	{ "ALTER TABLE forwarded_payments ADD failcode INTEGER;", NULL },
+	/* remote signatures for channel announcement */
+	{ "ALTER TABLE channels ADD remote_ann_node_sig BLOB;", NULL },
+	{ "ALTER TABLE channels ADD remote_ann_bitcoin_sig BLOB;", NULL },
 };
 
 /* Leak tracking. */
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -595,6 +595,50 @@ wallet_htlc_sigs_load(const tal_t *ctx, struct wallet *w, u64 channelid)
 	return htlc_sigs;
 }
 
+bool wallet_remote_ann_sigs_load(const tal_t *ctx, struct wallet *w, u64 id,
+				 secp256k1_ecdsa_signature **remote_ann_node_sig,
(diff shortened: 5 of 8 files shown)


channeld/channeld.c
```diff
@@ -569,8 +650,54 @@ static void handle_peer_announcement_signatures(struct peer *peer, const u8 *msg
...
+				    &peer->channel_id,
+			    	    "Wrong announcement: "
+				    "first received short_channel_ids: %s"
+			    	    " , now second get %s",
+			    	    type_to_string(peer, struct short_channel_id,
+						   &peer->short_channel_ids[REMOTE]),
+			    	    type_to_string(peer, struct short_channel_id,
+						   remote_scid));
+
+		/* make sure new node signature and bitcoin signature equal
```
SimonVrouwe: Shouldn't we just check that the signature is _valid_? Signatures change when re-signing (using different k [value](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm)).

SimonVrouwe:
I think it would be nice to have some sort of (py)test to test the functionality. Also to get more clarity on what kind of behavior we actually want/expect.

Related questions: 
- Does anybody know which LND version(s) cause this issue?
- Could we use [lightning-integration](https://github.com/cdecker/lightning-integration) to test?

channeld/channeld.c
trueptolemy: Thank you! You're right, and there shouldn't compare with the old sigs if the remote generate new sigs. Gossipd will verify the announcement, so I think we'd better update the new sigs here.

trueptolemy:
> Also to get more clarity on what kind of behavior we actually want/expect.

I think the point is we want to get into the phase of announce from waiting for remote signatures more quickly.
We care how to get the **valid** signature faster, and we have 2 choice: wait for remote reply or use the old signatures(rejection for the new sigs is a bad idea:( ). In the origin code, we just check the validity of signatures and don't think about if signatures are different with the ones received before.
In this PR, local peer still send local sigs to the remote and receives the remote reply. It commucates with others like original. 
But the difference is : it use the old signatures directly to announce although it will also receive the remote reply with remote signatures(the old or the new). In my understanding, the announcement with different signatures is ok iff the signatures are all valid.

trueptolemy:
@SimonVrouwe Sorry. I find another change for the current PR: when we enable the channel, we directly announce with the old sigs meanwhile sending remote peer our sigs. So we will announce once again when we receive the announcement reply.
Sending our own sigs to the remote peer is necessary. This leads directly to the issue that how to handle the reply from remote peer. Now I handle this special reply as usual, and also ask Channeld to announce.
Another alternative way is to ignore this reply. 
I think the formor way(which was achived in this PR) seems more reasonable, but now we meet two announcement almost at the same time.
lightning-integration is needed though it doesn't matter for clightning.
Thank you very mush :)

trueptolemy:
> Sending our own sigs to the remote peer is necessary. This leads directly to the issue that how to handle the reply from remote peer. Now I handle this special reply as usual, and also ask Channeld to announce.
> Another alternative way is to ignore this reply.

@cdecker hi, what do you think of these two ways?

lightningd/channel.c
```diff
@@ -174,7 +174,9 @@ struct channel *new_channel(struct peer *peer, u64 dbid,
 			    const struct pubkey *future_per_commitment_point,
 			    u32 feerate_base,
 			    u32 feerate_ppm,
-			    const u8 *remote_upfront_shutdown_script)
+			    const u8 *remote_upfront_shutdown_script,
+				secp256k1_ecdsa_signature *remote_ann_node_sig,
+				secp256k1_ecdsa_signature *remote_ann_bitcoin_sig)
```
cdecker: @rustyrussell should these be marked as `TAKES` if we `tal_steal` them below? I can never remember the semantics of `TAKES` :wink:

wallet/db.c
```diff
@@ -376,12 +376,15 @@ static struct migration dbmigrations[] = {
 	{ "ALTER TABLE channels ADD feerate_base INTEGER;", NULL },
 	{ "ALTER TABLE channels ADD feerate_ppm INTEGER;", NULL },
 	{ NULL, migrate_pr2342_feerate_per_channel },
-	{ "ALTER TABLE channel_htlcs ADD received_time INTEGER", NULL },
-	{ "ALTER TABLE forwarded_payments ADD received_time INTEGER", NULL },
-	{ "ALTER TABLE forwarded_payments ADD resolved_time INTEGER", NULL },
+	{ "ALTER TABLE channel_htlcs ADD received_time INTEGER;", NULL },
+	{ "ALTER TABLE forwarded_payments ADD received_time INTEGER;", NULL },
+	{ "ALTER TABLE forwarded_payments ADD resolved_time INTEGER;", NULL },
```
cdecker: Please don't modify existing migrations. While this change may be harmless cosmetics, changing the migrations may end up corrupting deployed databases.

wallet/wallet.c
```diff
@@ -951,6 +974,29 @@ u64 wallet_get_channel_dbid(struct wallet *wallet)
...
+			  " WHERE id=?");
+
+	if ((chan->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL) &&
+				   chan->remote_ann_node_sig &&
+				   chan->remote_ann_bitcoin_sig) {
+		sqlite3_bind_signature(stmt, 1, chan->remote_ann_node_sig);
+		sqlite3_bind_signature(stmt, 2, chan->remote_ann_bitcoin_sig);
+	} else {
+		sqlite3_bind_null(stmt, 1);
+		sqlite3_bind_null(stmt, 2);
```
cdecker: Arguably we shouldn't be calling `wallet_announcement_save` at all when these are `NULL`.

channeld/channel_wire.csv
```diff
@@ -50,6 +50,8 @@ channel_init,,failed_sides,num_failed*enum side
 channel_init,,local_funding_locked,bool
 channel_init,,remote_funding_locked,bool
 channel_init,,funding_short_id,struct short_channel_id
+channel_init,,remote_ann_node_sig,?secp256k1_ecdsa_signature
+channel_init,,remote_ann_bitcoin_sig,?secp256k1_ecdsa_signature
```
cdecker: Please add optional fields only to the end of a message.

channeld/channeld.c
```diff
@@ -163,8 +163,9 @@ struct peer {
 
 	/* Empty commitments.  Spec violation, but a minor one. */
 	u64 last_empty_commitment;
-	/* It means if we are in reconnection process*/
-	bool reconnected;
+
+	/* It means if we have stored announcement signatures so far*/
+	bool remote_ann_stored;
```
cdecker: Like mentioned before we don't need to track this. If the remote end wants to give us a new (valid) signature we can certainly just overwrite any existing one.

channeld/channeld.c
```diff
@@ -510,7 +511,7 @@ static void channel_announcement_negotiate(struct peer *peer)
 		 * announcement to MASTER.
 		 * But we won't do this when restart, because we load announcement
 		 * signatures form DB when we reenable Channeld! */
-		if(peer->reconnected == false){
+		if(peer->remote_ann_stored == false){
```
cdecker: See above, just always send the signatures to the master.

lightningd/channel_control.c
```diff
@@ -116,6 +116,38 @@ static void peer_got_funding_locked(struct channel *channel, const u8 *msg)
 		lockin_complete(channel);
 }
 
+static void peer_got_announcement(struct channel *channel, const u8 *msg)
+{
+	secp256k1_ecdsa_signature *remote_ann_node_sig;
+	secp256k1_ecdsa_signature *remote_ann_bitcoin_sig;
+
+	remote_ann_node_sig = tal(tmpctx, secp256k1_ecdsa_signature);
+	remote_ann_bitcoin_sig = tal(tmpctx, secp256k1_ecdsa_signature);
```
cdecker: You can allocate them directly off of `channel` if you `tal_free` them on error as well (which should be rare anyway).

lightningd/channel.h
```diff
@@ -62,6 +62,11 @@ struct channel {
 	struct amount_sat funding;
 	struct amount_msat push;
 	bool remote_funding_locked;
+
+	/* channel announcement sigs info from remote peer*/
+	secp256k1_ecdsa_signature *remote_ann_node_sig;
+	secp256k1_ecdsa_signature *remote_ann_bitcoin_sig;
```
cdecker: Why are we keeping these in memory at all? We could just as well load them on demand when a new `channeld` gets started.

channeld/channel_wire.csv
```diff
@@ -101,6 +101,9 @@ channel_got_announcement,1017
 channel_got_announcement,,remote_ann_node_sig,secp256k1_ecdsa_signature
 channel_got_announcement,,remote_ann_bitcoin_sig,secp256k1_ecdsa_signature
 
+# Wait for reply, to make sure it's on disk.(We store the announcement only once!)
+channel_got_announcement_reply,1117
```
cdecker: I don't think we need to wait for the master in this case. Either it stored it and we're happy, or it crashed and we're going to die in a few milliseconds anyway :-)

cdecker: Especially because we suddenly need to buffer other things while waiting for this reply which really is just an optimization.

lightningd/channel_control.c
```diff
@@ -139,13 +139,21 @@ static void peer_got_announcement(struct channel *channel, const u8 *msg)
 		return;
 	}
 
+	/* To handle the case that Master was shut down before sends reply.
+	 * In this case, master may receive announcement more than once!
+	 */
+	tal_free(channel->remote_ann_node_sig);
+	tal_free(channel->remote_ann_bitcoin_sig);
```
cdecker: If the master was shut down we're dead as well. No need to handle that case here.

trueptolemy:
@cdecker Sincerely thank you! I'll correct these days. Many points, detials to learn and I hope I can do better next time :)

channeld/channel_wire.csv
trueptolemy: Sorry, I'll take care.

channeld/channel_wire.csv
trueptolemy: Thank you! I'll remove it

wallet/wallet.c
trueptolemy: another "overcheck"

wallet/db.c
trueptolemy: Sorry, I'll take care, too

lightningd/channel.h
trueptolemy: Now I replace it by a new load function, which is only called when we generate a channel_init message.

channeld/channeld.c
```diff
@@ -569,8 +596,33 @@ static void handle_peer_announcement_signatures(struct peer *peer, const u8 *msg
...
 
-	peer->have_sigs[REMOTE] = true;
-	billboard_update(peer);
+	/* If we received remote announcement before, now we need to make
+	 * sure the new announcement is same as the old and then ignore it.
+	 */
+	if(peer->have_sigs[REMOTE] == true) {
+
+		/* make sure new short_channel_id must equal the old one */
+		if (!short_channel_id_eq(remote_scid,
```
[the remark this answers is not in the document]
trueptolemy: I made a childish mistake (and now I learn a lot ).  Thank you!

channeld/channeld.c
trueptolemy: Yes, "Do The Simplest Thing"

channeld/channeld.c
trueptolemy: Yes, I simplied it in the new commit :)

lightningd/channel_control.c
trueptolemy: Yes, I'll remove it.

channeld/channeld.c
```diff
@@ -163,6 +163,8 @@ struct peer {
 
 	/* Empty commitments.  Spec violation, but a minor one. */
 	u64 last_empty_commitment;
+	/* It means if we are in reconnection process*/
+	bool reconnected;
```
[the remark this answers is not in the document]
trueptolemy: Thank you, I've removed it in the new commit.

lightningd/channel.c
trueptolemy: I find we marked as `TAKES` if we `tal_dup`, and there's some similar cases in this function:
we didn't marked `scid`, `last_tx`, `last_sent_commit` as `TAKES` and we tal_steal them below,
but we marked `transient_billboard` as `TAKES` because we `tal_strdup` it below.

channeld/channeld.c
```diff
@@ -492,15 +492,36 @@ static void channel_announcement_negotiate(struct peer *peer)
...
 	 * announcement, otherwise we send a temporary one */
-	if (peer->have_sigs[LOCAL] && peer->have_sigs[REMOTE])
+	if (peer->have_sigs[LOCAL] && peer->have_sigs[REMOTE]) {
+		check_short_ids_match(peer);
+
+		/* After making sure short_channel_ids match, we can send remote
+		 * announcement to MASTER.
+		 * But we won't do this when restart, because we load announcement
+		 * signatures form DB when we reenable Channeld! */
+		if(peer->reconnected == false){
```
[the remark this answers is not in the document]
trueptolemy: Yes, it need to be cleaned up.

lightningd/channel_control.c
trueptolemy: Yes, now I don't allocate them directly.

trueptolemy:
Hi, @cdecker I've correct the code. What do you think about it now? :-)

lightningd/channel.c
trueptolemy: Rusty replied me that `tal_steal` doesn't need to be marked as `TAKES`, and `tal_dup` can be marked  :)

wallet/wallet.c
```diff
@@ -595,6 +595,42 @@ wallet_htlc_sigs_load(const tal_t *ctx, struct wallet *w, u64 channelid)
 	return htlc_sigs;
 }
 
+bool wallet_remote_ann_sigs_load(struct wallet *w, u64 id,
+				 secp256k1_ecdsa_signature **remote_ann_node_sig,
+				 secp256k1_ecdsa_signature **remote_ann_bitcoin_sig)
```
cdecker: If we are passing a pointer-to-pointer-of-sugnature we should also be in charge of allocating the signature, otherwise we need to allocate outside of the call, keep an alias, call the function and then free the alias in case of a failure. If however we allocate internally then the pointer will get populated correctly.

wallet/wallet.c
```diff
@@ -595,6 +595,42 @@ wallet_htlc_sigs_load(const tal_t *ctx, struct wallet *w, u64 channelid)
...
+	sqlite3_bind_int64(stmt, 1, id);
+
+	res = sqlite3_step(stmt);
+
+	/* This must succeed, since we know the channel exists */
+	assert(res == SQLITE_ROW);
+
+	if (sqlite3_column_type(stmt, 0) != SQLITE_NULL) {
+		if (!sqlite3_column_signature(stmt, 0, *remote_ann_node_sig)) {
+			db_stmt_done(stmt);
```
cdecker: In other parts of the code we use a `fail:` label to clear and return if we have multiple similar failure paths in a single function.

wallet/wallet.c
```diff
@@ -741,6 +777,7 @@ static struct channel *wallet_stmt2channel(struct wallet *w, sqlite3_stmt *stmt)
 			   sqlite3_column_int(stmt, 42),
 			   sqlite3_column_int(stmt, 43),
 			   sqlite3_column_arr(tmpctx, stmt, 44, u8));
+
```
cdecker: Spurious newline.

wallet/wallet.c
```diff
@@ -1086,6 +1141,16 @@ void wallet_channel_save(struct wallet *w, struct channel *chan)
 		sqlite3_bind_null(stmt, 1);
 	sqlite3_bind_int64(stmt, 2, chan->dbid);
 	db_exec_prepared(w->db, stmt);
+
+	/* save announcement sigs only when we ge announcement*/
```
cdecker: `ge` -> `get`

cdecker: Also these fields ought to be initialized to `null` anyway in the DB, so no need to do this additional initialization.

wallet/wallet.h
```diff
@@ -1052,4 +1067,16 @@ struct amount_msat wallet_total_forward_fees(struct wallet *w);
...
 const struct forwarding *wallet_forwarded_payments_get(struct wallet *w,
 						       const tal_t *ctx);
+
+/**
+ * Load remote_ann_node_sig and remote_ann_bitcoin_sig
+ *
+ * @w: wallet containing the channel
+ * @id: channel database id
+ * @remote_ann_node_sig: location to load remote_ann_node_sig to
+ * @remote_ann_bitcoin_sig: location to load remote_ann_bitcoin_sig to
```
cdecker: Like mentioned above, with this signature I would have expected the function to allocate the signatures on the heap. But in that case we're missing a `ctx` to allocate from.

lightningd/channel_control.c
```diff
@@ -117,6 +117,31 @@ static void peer_got_funding_locked(struct channel *channel, const u8 *msg)
...
 
+static void peer_got_announcement(struct channel *channel, const u8 *msg)
+{
+	secp256k1_ecdsa_signature remote_ann_node_sig;
+	secp256k1_ecdsa_signature remote_ann_bitcoin_sig;
+
+	if(!(channel->channel_flags & CHANNEL_FLAGS_ANNOUNCE_CHANNEL)) {
+		channel_internal_error(channel,
+				       "No ANNOUNCE flag, no announcemnt signatures! %s",
+				       tal_hex(tmpctx, msg));
```
cdecker: This check, while correct, may be a bit aggressive in the current network and might result in a lot of failures (nodes usually just spray you with the signatures without checking the announcement flags). We can still enforce that we don't announce by simply not sending sigs ourselves.

Please don't fail the channel over such a minor offence, but log it and we can tighten this later.

channeld/channeld.c
```diff
@@ -2915,6 +2919,20 @@ static void init_channel(struct peer *peer)
 		     feerate_per_kw[LOCAL], feerate_per_kw[REMOTE],
 		     peer->feerate_min, peer->feerate_max);
 
+	if(remote_ann_node_sig && remote_ann_bitcoin_sig) {
+		peer->announcement_node_sigs[REMOTE] = *remote_ann_node_sig;
+		peer->announcement_bitcoin_sigs[REMOTE] = *remote_ann_bitcoin_sig;
+		peer->have_sigs[REMOTE] = true;
+
+		tal_steal(tmpctx, remote_ann_node_sig);
+		tal_steal(tmpctx, remote_ann_bitcoin_sig);
```
cdecker: Why are you stealing these onto `tmpctx`? You could also pass the pointers to `peer->announcement_node_sigs[REMOTE] ` and `peer->announcement_bitcoin_sigs[REMOTE]` into the `fromwire_channel_init` call directly, which avoids having these temp variables and having to deal with intermediate allocations and ownership.

lightningd/channel_control.c
```diff
@@ -303,6 +303,10 @@ void peer_start_channeld(struct channel *channel,
 	const struct config *cfg = &ld->config;
 	bool reached_announce_depth;
 	struct secret last_remote_per_commit_secret;
+	secp256k1_ecdsa_signature *remote_ann_node_sig, *remote_ann_bitcoin_sig;
+
+	remote_ann_node_sig = tal(tmpctx, secp256k1_ecdsa_signature);
+	remote_ann_bitcoin_sig = tal(tmpctx, secp256k1_ecdsa_signature);
```
cdecker: This is the weird ownership I was talking about before. Please have `wallet_remote_ann_sigs_load` manage allocations. The only reason we aren't producing a memory leak here is that `tmpctx` and all child allocations is cleared in the `io_loop`.

wallet/wallet.c
trueptolemy: Thank you! Now I manage the allocation in this func!

wallet/wallet.c
trueptolemy: Yes, now I use `fail:`

wallet/wallet.c
trueptolemy: fixed

lightningd/channel_control.c
trueptolemy: Thank you! Now I remove this check

channeld/channeld.c
trueptolemy: Yes, I find these tal_steal are extra. But I think `fromwire_channel_init` needs these 2 pointer, because they can handle `NULL` sigs:

> `channel_init,,remote_ann_node_sig,?secp256k1_ecdsa_signature` `channel_init,,remote_ann_bitcoin_sig,?secp256k1_ecdsa_signature`

lightningd/channel_control.c
trueptolemy: Thank you so much! Now I pass `ctx `to `wallet_remote_ann_sigs_load` and let it manage allocations.

channeld/channeld.c
trueptolemy: emm, after deleting tal_steal ,Travis CI tell me memory leak. So I tal_free these directly.(I find other variables used in fromwire_channel_init are freed directly, so I also do it.)

wallet/wallet.h
trueptolemy: Thank you! Now I add it.

wallet/wallet.c
```diff
@@ -595,6 +595,46 @@ wallet_htlc_sigs_load(const tal_t *ctx, struct wallet *w, u64 channelid)
...
+		if (!sqlite3_column_signature(stmt, 1, *remote_ann_bitcoin_sig))
+			goto fail;
+	}
+
+	db_stmt_done(stmt);
+	return true;
+
+fail:
+	*remote_ann_node_sig = tal_free(*remote_ann_node_sig);
+	*remote_ann_bitcoin_sig = tal_free(*remote_ann_bitcoin_sig);
```
cdecker: This can easily lead to a double-free or freeing a random memory location since we do not allocate in all cases. It'd be better to check both fields for `SQLITE_NULL` at the top, and set the out-args to `NULL` and return immediately if either is `SQLITE_NULL`. If both are not `SQLITE_NULL` initialized both using `tal` and then `goto fail;` if either of the `sqlite3_column_signature` calls fails. That way we either initialize all or none, and free only when both are actually initialized.

trueptolemy: Thank you! What you said is better, kinds of cases make sense easily.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2619/commits/e884bc687aa655c1bcdb1f2bbea82c350629eddb

PR #2624, opened 2019-05-07
fix: pylightning to_btc_str precision

m-schmoock:
The old codes if % 1000 statement logic was simply inverted
and produced the opposite output of the intention behind it.

Before Fix:
 - `Millisatoshi('42sat').to_btc_str()` => `0.00000042000btc`
 - `Millisatoshi('42001msat').to_btc_str()` => `0.00000042btc`

After  Fix:
 - `Millisatoshi('42sat').to_btc_str()` => `0.00000042btc`
 - `Millisatoshi('42001msat').to_btc_str()` => `0.00000042001btc`


rustyrussell:
How embarrassing! Thanks!

PR #2625, opened 2019-05-09
Update 'listfunds' doc and manpage

darosior:
Some fields were returning values as msat while being described as returning values in sat. Also added a short description to some utxo fields and ordered them.


rustyrussell:
Thanks!

Ack 3f382a5291533996be6c43005658384441b61382

PR #2628, opened 2019-05-10
Some driveby cleanups

cdecker:
Just the usual cleanup collection. The one stands out is the pylightning fix where I removed the hand-woven binding and now just fallback on python's own binding logic. This should make optional argument handling work again.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/3e997673c410aa20fc86ae62118aae46365576fc

PR #2630, opened 2019-05-12
Feat: pylightning extended bitcoin units

m-schmoock:
This adds an extended `Millisatoshi` unit parsing engine to `pylightning`, supporting:

 - btc, mbtc, µbtc, sat, msat, ...
 - undestands plural notations:  `100sats`
 - also accepts upperase: `100BTC`
 - understands alternative names: `100finney`, `100millibit`, ...
 - allows whitespaces trailing and between amount and unit: ` 10 btc `

Adds extensive testcases for all notations to the `pylightning` test modules.
See Bitcoin Wiki for details: https://en.bitcoin.it/wiki/Units

The PR also adds a `Millisatoshi.to_short_str(<digits:3>)` digits function that is useful to output rounded amounts to the user when there is less character space.

Note: This does not support `megabitcoin` and `kilobitcoin`/`kbtc` by intention for security reasons.
Disclaimer: If this is too much bikeshedding, I have no problem in removing the stuff like fuzzy whitespacing, plurals and unusual notations. When implementing `to_short_str` (https://github.com/lightningd/plugins/pull/25) I just found that also having `mbtc` and `µbtc` is a useful feature.

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -1,9 +1,13 @@
 from decimal import Decimal
 import json
 import logging
+import re
 import socket
 
-__version__ = "0.0.7.1"
+__version__ = "0.0.7.2"
+
+unit_regex = re.compile(r'^\s*([\d.]+)\s*([\w]+)\s*$')
+unit_error = TypeError("Millisatoshi must be numeric string with msat/sat/btc/mBTC/uBTC/µBTC/bits/finney... suffix or int")
 
 
 class RpcError(ValueError):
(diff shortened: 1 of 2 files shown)


contrib/pylightning/lightning/lightning.py
```diff
@@ -22,20 +27,48 @@ class Millisatoshi:
 
     Many JSON API fields are expressed in millisatoshis: these automatically get
     turned into Millisatoshi types.  Converts to and from int.
+
+    For common Bitcoin currency units see: https://en.bitcoin.it/wiki/Units
     """
     def __init__(self, v):
         """
-        Takes either a string ending in 'msat', 'sat', 'btc' or an integer.
+        Takes either a string ending in 'msat', 'sat', 'btc', "mbtc", "finney", ... or an integer.
```
darosior: nit: maybe choose between `"` or `'` for consistency ?

m-schmoock: :D will do

darosior:
ACK `ce461d9`

m-schmoock:
Ackbot either offline or he doesnt like us :D

rustyrussell:
The problem with adding these is that *removing* them would be an API break, so we're stuck with them.

Neither mbtc or ubtc is a common unit, and the others are worse.  By providing a "to_short_str()" which uses them you're now placing the burden on other consumers to support those units.

And adding these is more UI issue, so I'm not sure it belongs in pylightning at all?  But it's up to @cdecker

m-schmoock:
I see. In theory, `to_short_str()` can work with just the old base units (btc, sat, msat) since it does significant digit rounding and skips trailing zeros.

I can see how it would look in the summary plugin and drop the extended units parts completely.

m-schmoock:
Note: I added a reduced version of this PR (no units, just `to_short_str`) in PR #2636

m-schmoock:
I decided to close this for now, and maybe add a only 'mbtc' PR that will include the other consumers well.

PR #2635, opened 2019-05-14
Benchmark memory reductions

rustyrussell:
This should allow us to run MCP on the pi.

1. The first one is a cleanup suggested by @cdecker for previous PR.
2. The next two are bechmark cleanups.
3. The next two make us more robust should we OOM creating JSON replies.
4. The final two reduce our memory usage.


cli/lightning-cli.c
```diff
@@ -265,11 +265,42 @@ static void print_json(const char *str, const jsmntok_t *tok, const char *indent
 	abort();
 }
 
+/* Always returns a positive number < len.  len must be > 0! */
+static size_t read_nofail(int fd, void *buf, size_t len)
+{
+	ssize_t i = read(fd, buf, len);
```
ZmnSCPxj: Given the comment "len must be > 0" maybe add `assert(len > 0);`?

gossipd/gossip_wire.csv
```diff
@@ -45,8 +45,10 @@ gossip_getroute_reply,,hops,num_hops*struct route_hop
 gossip_getchannels_request,3007
 gossip_getchannels_request,,short_channel_id,?struct short_channel_id
 gossip_getchannels_request,,source,?struct node_id
+gossip_getchannels_request,,prev,?struct short_channel_id
```
ZmnSCPxj: This is a very good change and I strongly approve.

gossipd/test/run-bench-find_route.c
```diff
@@ -277,6 +278,13 @@ int main(int argc, char *argv[])
 		if (route_lengths[i])
 			printf(" Length %zu: %zu\n", i, route_lengths[i]);
 
+	/* Since we omitted destructors on these, clean up manually */
```
ZmnSCPxj: Maybe we can afford 40 bytes to write a destructor for the entire `rstate` that does `free_chan` on each `chan` it owns, instead of duplicating this code at each test?

cli/lightning-cli.c
rustyrussell: Agreed, done.

gossipd/test/run-bench-find_route.c
rustyrussell: I feel like an idiot; that is much neater!  Thanks!

rustyrussell:
Added assert(), and appended a commit which puts a destructor on rstate as @ZmnSCPxj suggested.

gossipd/routing.c
```diff
@@ -154,6 +154,16 @@ struct chan *next_chan(const struct node *node, struct chan_map_iter *i)
 	return chan_map_next(&node->chans.map, i);
 }
 
+static void destroy_routing_state(struct routing_state *rstate)
+{
+	/* Since we omitted destructors on these, clean up manually */
```
ZmnSCPxj: Comment could probably be better.  "Since we omitted destructors on the channels, clean them up here when we clean up the routing state".  It seem, inside a destructor itself, so word "manually" feels wrong.

PR #2636, opened 2019-05-15
Feat: msat to_approx_str

m-schmoock:
I reduced PR #2630 to just contain a reduced version of `to_approx_str`  and its testcase.
So you can quickly make a decision :D

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -1,6 +1,7 @@
 from decimal import Decimal
 import json
 import logging
+from math import floor, log10
 import socket
 
 __version__ = "0.0.7.1"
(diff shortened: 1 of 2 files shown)


contrib/pylightning/tests/test_units.py
```diff
@@ -0,0 +1,31 @@
+from lightning import Millisatoshi
+import pytest
```
rustyrussell: flake8 is upset:

    contrib/pylightning/tests/test_units.py:2:1: F401 'pytest' imported but unused

m-schmoock: Whoops. Fixed

rustyrussell:
This surprised me, that it lost precision.  I think this needs to be called "to_approx_str" maybe?

m-schmoock:
Good point, renamed it

m-schmoock:
Note: I added the PR that adds this and `mbtc` unit in https://github.com/ElementsProject/lightning/pull/2657

I leave this open for discussion.

rustyrussell:
I like it, but I'm reluctant to apply with @cdecker since he's the pylightning owner....

m-schmoock:
No changes, I just rebased to master, just so #2657 can build on this, as it had merge conflicts.

m-schmoock:
ACK 7edfb60  (since no change, just rebase since rusty's ack)

m-schmoock:
ACK 7edfb60

contrib/pylightning/lightning/lightning.py
```diff
@@ -85,6 +86,23 @@ def to_btc_str(self):
         else:
             return '{:.8f}btc'.format(self.to_btc())
 
+    def to_approx_str(self, digits: int = 3):
+        """
```
cdecker: Comments usually have the format

```python
"""Short one-line description

Longer, potentially multi-line description
"""
```

contrib/pylightning/tests/test_units.py
```diff
@@ -0,0 +1,30 @@
...
+    amount = Millisatoshi('1000sat')
+    assert amount.to_approx_str() == "1000sat"
+    amount = Millisatoshi('100msat')
+    assert amount.to_approx_str() == "0.1sat"
+
+    # also test significant rounding
+    amount = Millisatoshi('10001234sat')
+    assert amount.to_approx_str() == "0.1btc"
+    amount = Millisatoshi('1234sat')
+    assert amount.to_approx_str() == "1230sat"
```
cdecker: This is kind of weird, we don't gain a length advantage by rounding the last digit, but we do lose precision.

cdecker: I think the `digit` parameter should allow rounding, but if it doesn't give an advantage length-wise it should force it.

contrib/pylightning/lightning/lightning.py
m-schmoock: thx for pointing out

contrib/pylightning/tests/test_units.py
m-schmoock: fixed

cdecker:
ACK 3c35f79ec16bc28bf61bd6f919d64d5a656087f7

PR #2641, opened 2019-05-15
openingd: fix hangup when gossipd compacts.

rustyrussell:
My raspberry pi node hung up on my other node:
   lightning_openingd-... chan #1: Got bad message from gossipd: 0db1

This is because we didn't handle that message in one path.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2641/commits/c2a5f8dd9c9284199a24fe3983944cb4da1e30f8

PR #2645, opened 2019-05-17
Fix bash completion

darosior:
The regex `^command=` is no longer valid to get the command list from the output of `lightning-cli -H help 2>/dev/null`

### DIFF
--- a/contrib/lightning-cli.bash-completion
+++ b/contrib/lightning-cli.bash-completion
@@ -31,7 +31,7 @@ _lightning_cli() {
 
             # get the regular commands
             if [[ -z "$cur" || "$cur" =~ ^[a-z] ]]; then
-		helpopts=$($lightning_cli -H help 2>/dev/null | sed -n 's/^command=//p')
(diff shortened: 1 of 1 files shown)


m-schmoock:
Looks good, but I can't test, as ArchLinux bash_completions (`/usr/share/bash-completion/completions/`) seem to work different and I could not test it.

rustyrussell:
Your version still didn't work right for me.  I tried to fix it, how's that?

darosior:
Yours works too for me.
Just a question: if I change the help output (human readable from cli), can I use `jq` for bash completion (I'll have to do it again anyway since we take `lightning-cli -H`) to use raw output instead, or adding a dependency is not an option ?

rustyrussell:
> Just a question: if I change the help output (human readable from cli), can I use `jq` for bash completion (I'll have to do it again anyway since we take `lightning-cli -H`) to use raw output instead, or adding a dependency is not an option ?

Depending on jq is kind of ugly, IMHO, esp. since it's unnecessary.

PR #2646, opened 2019-05-17
Rpc: sort help output by command name

darosior:
The command list is easier to read when sorted.

### DIFF
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -13,11 +13,13 @@
  * that point, the `json_connection` becomes the owner (or it's simply freed).
  */
 /* eg: { "jsonrpc":"2.0", "method" : "dev-echo", "params" : [ "hello", "Arabella!" ], "id" : "1" } */
+#include "ccan/config.h"
 #include <arpa/inet.h>
 #include <bitcoin/address.h>
 #include <bitcoin/base58.h>
 #include <bitcoin/script.h>
 #include <ccan/array_size/array_size.h>
+#include <ccan/asort/asort.h>
 #include <ccan/err/err.h>
 #include <ccan/io/io.h>
 #include <ccan/str/hex/hex.h>
@@ -346,6 +348,12 @@ static const struct json_command *find_command(struct json_command **commands,
 	return NULL;
 }
 
+static int compare_commands_name(struct json_command *const *a,
+					struct json_command *const *b, void *unused)
+{
+	return strcmp((*a)->name, (*b)->name);
+}
+
 static struct command_result *json_help(struct command *cmd,
 					const char *buffer,
 					const jsmntok_t *obj UNNEEDED,
@@ -371,6 +379,8 @@ static struct command_result *json_help(struct command *cmd,
 					    buffer + cmdtok->start);
 	} else
 		one_cmd = NULL;
+
+	asort(commands, tal_count(commands), compare_commands_name, NULL);
 
 	response = json_stream_success(cmd);
 	json_object_start(response, NULL);


lightningd/jsonrpc.c
```diff
@@ -372,6 +379,8 @@ static struct command_result *json_help(struct command *cmd,
 	} else
 		one_cmd = NULL;
 
+	qsort(commands, tal_count(commands), sizeof(struct json_command*), compare_commands_name);
```
rustyrussell: Nice idea!  But ccan/asort will avoid all the casts:

    static int compare_commands_name(struct json_command *a, struct json_command *b, void *unused)...

    asort(commands, tal_count(commands), compare_commands_name, NULL);

darosior: Thank you, it is far more readable this way but I get an error with `qsort_r` :
```
error: implicit declaration of function ‘qsort_r’; did you mean ‘qsort’?
```

m-schmoock: @darosior Maybe related to https://stackoverflow.com/a/20685199/884474

As a tests, does `#define _GNU_SOURCE` work for you?

m-schmoock:
@darosior @rustyrussell I think it would be even greater to have a separation of the standard commands and the plugin extensions. So two sorted outputs, first one for standard commands then for plugin extension commands (if possible) i.e. like this:

```
--- STANDARD COMMANDS
...
waitinvoice label
    Wait for an incoming payment matching the invoice with {label}, or if the invoice expires

waitsendpay payment_hash [timeout]
    Wait for payment attempt on {payment_hash} to succeed or fail, but only up to {timeout} seconds.

withdraw destination satoshi [feerate] [minconf]
    Send to {destination} address {satoshi} (or 'all') amount via Bitcoin transaction, at optional {feerate}

--- PLUGIN COMMANDS

summary 
    Gets summary information about this node.

rebalance outgoing_scid incoming_scid [msatoshi] [maxfeepercent] [retry_for] [exemptfee]
    Rebalancing channel liquidity with circular payments. This tool helps to move some msatoshis between your channels.

---
```

What do you think?

darosior:
@m-schmoock I think it's a good idea, I can make it in another PR. I dont know if it should be handled on the cli or rpc side though.

lightningd/jsonrpc.c
darosior: @m-schmoock nop i already tested ^^

cdecker:
Added the Work in Progress tag until `asort` is added :wink:

lightningd/jsonrpc.c
darosior: @m-schmoock It finally worked by adding it at the top of the file.

darosior: @rustyrussell ccan/asort doesn't work:
I can't declare `compare_commands_name` as :
```C
static int compare_commands_name(struct json_command *a, struct json_command *b, void *unused)
```
without a typesafe error ("passing argument 4 of ‘qsort_r’ from incompatible pointer type").
And even if I then stick to the following (which doesn't avoid casts) :
```C
static int compare_commands_name(const void *a, const void *b, void *v)
```
sorting won't work.

rustyrussell:
OK, so this is the right answer (took some fiddling).  You also need to #include "config.h" at the top, as otherwise arpa.h pulls in qsort.h before we define _GNU_SOURCE.

```
static int compare_commands_name(struct json_command *const *a,
				 struct json_command *const *b, void *unused)
{
	return strcmp((*a)->name, (*b)->name);
}
```

Modifying this to put plugins last is fairly easy, too (but a separate commit).  We should put a plugin arg to json_command_add and sort appropriately (maybe add a [<pluginname>] to start of description?).

Note that we should also use opt_register_table() for plugin options, as well, rather than opt_register_arg: this will group them nicely by plugin :)

rustyrussell:
Oh, also, if you're going to sort here, make a copy:

```
	commands = tal_dup_arr(tmpctx, struct json_command *,
			       cmd->ld->jsonrpc->commands,
			       tal_count(cmd->ld->jsonrpc->commands), 0);
```

darosior:
@rustyrussell thank you for your help, I amended the changes.

> Modifying this to put plugins last is fairly easy, too (but a separate commit). We should put a plugin arg to json_command_add and sort appropriately (maybe add a [] to start of description?).
> 
> Note that we should also use opt_register_table() for plugin options, as well, rather than opt_register_arg: this will group them nicely by plugin :)

I can take care of that, while I'm at it should I sort it by category (maybe only for the `-H` output) ? i.e like in bitcoin-core : 
```
==== networking ====
connect 
ping
...

==== channels management =====
fundchannel
close
listchannels
...

==== payment ====
pay
invoice
waitinvoice
....

==== Bitcoin ====
newaddr
withdraw
....

etc..
```

PR #2647, opened 2019-05-18
Plugins: add a default 'plugins' directory in the config dir

darosior:
Having a `plugins` directory in the configuration path can be very handy instead of manually adding `--plugin-dir=` or `--plugin=`, and it's easier to have it in the configuration directory than in the installation one.

This PR adds a function that will, if it exists, iterate through `<path_to_config_dir>/plugins` and add, without failing on error, each directory it contains as a new `plugin-dir`.

It permits to, for instance :
```shell
git clone https://github.com/lightningd/plugins ~/.lightning/plugins
```
and have (once requirements installed) all the community curated list of plugins available.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -900,6 +900,20 @@ void clear_plugins(struct plugins *plugins)
 		tal_free(p);
 }
 
+void plugins_add_default_dir(struct plugins *plugins, const char *default_dir)
+{
+	DIR *d = opendir(default_dir);
+	if (d) {
+		struct dirent *di;
+		while ((di = readdir(d)) != NULL) {
+			if (streq(di->d_name, ".") || streq(di->d_name, ".."))
+				continue;
+			add_plugin_dir(plugins, path_join(tmpctx, default_dir, di->d_name), true);
+		}
+	}
+	closedir(d);
+}
+
 void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 {
 	struct plugin *p;
@@ -909,6 +923,8 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 	struct jsonrpc_request *req;
 	plugins->pending_manifests = 0;
 	uintmap_init(&plugins->pending_requests);
+
+	plugins_add_default_dir(plugins, path_join(tmpctx, plugins->ld->config_dir, "plugins"));
 
 	setenv("LIGHTNINGD_PLUGIN", "1", 1);
 	/* Spawn the plugin processes before entering the io_loop */
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - plugin: the `connected` hook can now send an `error_message` to the rejected peer.
 - Protocol: we now enforce `option_upfront_shutdown_script` if a peer negotiates it.
 - JSON API: `listforwards` now includes the local_failed forwards with failcode (Issue [#2435](https://github.com/ElementsProject/lightning/issues/2435), PR [#2524](https://github.com/ElementsProject/lightning/pull/2524))
+- Config: Added a default plugin directory : `lightning_dir/plugins`. Each plugin directory it contains will be added to lightningd on startup.
 
 ### Changed
 
--- a/lightningd/plugin.h
+++ b/lightningd/plugin.h
@@ -31,6 +31,12 @@ struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
 			    struct lightningd *ld);
 
 /**
+ * Search for `default_dir`, and if it exists add every directory it
+ * contains as a plugin dir.
+ */
+void plugins_add_default_dir(struct plugins *plugins, const char *default_dir);
+
+/**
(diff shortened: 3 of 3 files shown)


cdecker:
> It permits to, for instance :
> 
> ```shell
> git clone https://github.com/lightningd/plugins ~/.lightning/plugins
> ```

That likely is not what you want to do anyway, since at least that repository contains quite a few highly experimental plugins (and some that will kill the daemon if not configured correctly, see `autoreload`). So selectively adding plugins might be good, but advocating for entire collections to be cloned into the plugin dir seems like a bad idea.

cdecker:
Also you only have a single commit that sorts commands?

darosior:
Ooops no that's a bad one

darosior:
> > It permits to, for instance :
> > ```shell
> > git clone https://github.com/lightningd/plugins ~/.lightning/plugins
> > ```
> 
> That likely is not what you want to do anyway, since at least that repository contains quite a few highly experimental plugins (and some that will kill the daemon if not configured correctly, see `autoreload`). So selectively adding plugins might be good, but advocating for entire collections to be cloned into the plugin dir seems like a bad idea.

Yeah it is for now (as I experienced when testing ^^) but it was more to demonstrate that plugin setup would be easier.

*I messed up with my commit, rewriting it i'll push it in a second*

darosior:
Ok I rewrote it but adding the `default_dir` as a parameter of the function (you always have better ideas when you wrote a code twice ^^)

darosior:
Test which are failing on Travis pass on my side.

lightningd/plugin.c
```diff
@@ -900,6 +900,21 @@ void clear_plugins(struct plugins *plugins)
...
+	if (d) {
+		struct dirent *di;
+		while ((di = readdir(d)) != NULL) {
+			if (streq(di->d_name, ".") || streq(di->d_name, ".."))
+				continue;
+			add_plugin_dir(plugins, path_join(NULL, default_dir, di->d_name), true);
+		}
+	}
+	closedir(d);
+    tal_free(default_dir);
```
rustyrussell: Indent.  Also, it's an anti-pattern to take ownership of a parameter's lifetime like this without explicit indication to the caller.  I would just use tmpctx instead of NULL here, and in the caller.

darosior: I added that `tal_free` in a hurry to make travis pass, I wasn't aware of : 
```
* allocate from.  We have a magic convenience context
* `tmpctx` for temporary allocations like this.
```
it's useful ^^.

darosior:
@rustyrussell amended the changes, thanks for the review.

m-schmoock:
Nice idea. Just tried.

Two issues:
- We need documentation about this.
- It does not work for me
     - I tried copying the plugins directories into the config dir, so that the path was like this: `~/.lightning/summary/summary.py`
     - I tried to symlink the directories into it

darosior:
> Nice idea. Just tried.
> 
> Two issues:
> 
> * We need documentation about this. `doc/PLUGINS.md` and in the plugin repo.
> * It does not work for me

I'll add some.

>   * I tried copying the plugins directories into the config dir, so that the path was like this: `~/.lightning/summary/summary.py`
>   * I tried to symlink the directories into it
> 
> In both cases the daemon did not recognize the plugins

The path is `~/.lightning/plugins/<your_plugin_dir>`, it is so to separate plugins files (so it doesn't get messy if a plugin has multiple files).

m-schmoock:
Ah, thanks for the clarificaion :D

m-schmoock:
@darosior 
For the plugin repo I did this doc PR: https://github.com/lightningd/plugins/pull/32

darosior:
> @darosior
> For the plugin repo I did this doc PR: [lightningd/plugins#32](https://github.com/lightningd/plugins/pull/32)

Great ! :D

fiatjaf:
What if the plugin is just one executable? It should be possible to just put it in `~/.lightning/plugins/`, why not?

darosior:
> What if the plugin is just one executable? It should be possible to just put it in `~/.lightning/plugins/`,

Why not just create a directory for it ?

> why not?

I think it can become messy with many plugins (especially if plugins comport multiple files).

darwin:
Just FYI. My dockerized lightningd started segfaulting[1] on startup without changes on my side. Git bisect pinpointed it to a497295f92b9b15d718aba9998f8da3f140f55c7. 

I'm planning to investigate it further during this week.

[1] https://travis-ci.org/darwin/simverse/builds/535175428#L2948-L2955

darwin:
I was unable to get better stack trace on core dump, so I just bisected the commit line by line.

The culprit is `closedir` call with NULL `d` ("plugins" default_dir does not exist in my case).
https://github.com/ElementsProject/lightning/pull/2647/files#diff-d7945e0ea0fc3a128a17dec5e0ff98ddR914

I guess it works for most environments because it depends on libc. In my docker container I have Alpine Linux and this crashes it.

I did a bit of googling and this discussion might be relevant:
https://lists.freebsd.org/pipermail/freebsd-standards/2014-January/002704.html

TL;DR; passing NULL into closedir is not safu

darosior:
> TL;DR; passing NULL into closedir is not safu

My bad, thank you for the investigation.

PR #2648, opened 2019-05-18
Pylightning: add some missing RPC methods and correct a docstring

darosior:
This PR adds `listconfigs()` and `listsendpays()` methods to `LightningRpc` class. It also changes default value of `expiry_time` to 1 week in the `get_route()` method.


darosior:
Now this also add `paystatus()` method.

PR #2649, opened 2019-05-19
Doc: update 'waitsendpay' manpage

### LINKED ISSUE #2536: getroute doesn't change route on sendpay failure
From the man page for `waitsendpay`:
```
 On error, if the error occurred from a node other than the final destination, the route
       table will be updated so that getroute(7) should return an alternate route (if any).
```
However, after a `sendpay` and a `waitsendpay` that returns a failure with error from a channel that is not the final destination, calling `getroute` again returns the same route. No amount of `getroute -> sendpay -> waitsendpay` causes `getroute` to return an alternate route. Using the `pay` command, however, will successfully complete the payment. So there are definitely alternate routes available. This makes the `sendpay` command effectively useless, as it will fail on the first route and we are unable to find a new route.

darosior:
Closes #2536 .

### DIFF
--- a/doc/lightning-waitsendpay.7
+++ b/doc/lightning-waitsendpay.7
@@ -2,12 +2,12 @@
 .\"     Title: lightning-waitsendpay
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2649/commits/c68a4386a819f5899f8ec5dae0ad863680dad02b

Issue #2445, opened 2019-03-05
Invoice cleaning belongs in a plugin


rustyrussell:
The current invoice autoclean stuff clearly belongs in a plugin, now we have them.

PR #2652, opened 2019-05-20
Autoclean is now a plugin

rustyrussell:
First change was because I was really surprised at the empty string for missing options; I don't think it will break anyone to change that.

Closes: #2445 

We could move delinvoices out too, but one thing at a time; this is purely cosmetic, really.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -984,9 +984,8 @@ static void plugin_config(struct plugin *plugin)
 	list_for_each(&plugin->plugin_opts, opt, list) {
 		/* Trim the `--` that we added before */
 		name = opt->name + 2;
-		if (!opt->value)
-			opt->value = "";
-		json_add_string(req->stream, name, opt->value);
+		if (opt->value)
+			json_add_string(req->stream, name, opt->value);
 	}
 	json_object_end(req->stream); /* end of .params.options */
 
--- a/plugins/libplugin.c
+++ b/plugins/libplugin.c
@@ -33,10 +33,13 @@ struct plugin_conn {
 	MEMBUF(char) mb;
 };
 
+/* Connection to make RPC requests. */
+static struct plugin_conn rpc_conn;
+
 struct command {
 	u64 id;
 	const char *methodname;
-	struct plugin_conn *rpc;
+	bool usage_only;
 };
 
 struct out_req {
@@ -144,8 +147,7 @@ static struct command *read_json_request(const tal_t *ctx,
 		plugin_err("JSON id '%*.s' is not a number",
 			   id->end - id->start,
 			   membuf_elems(&conn->mb) + id->start);
-	/* Putting this in cmd avoids a global, or direct exposure to users */
-	cmd->rpc = rpc;
+	cmd->usage_only = false;
 	cmd->methodname = json_strdup(cmd, membuf_elems(&conn->mb), method);
 
 	return cmd;
@@ -271,7 +273,7 @@ struct command_result *command_fail(struct command *cmd,
 /* We invoke param for usage at registration time. */
 bool command_usage_only(const struct command *cmd)
 {
-	return cmd->rpc == NULL;
+	return cmd->usage_only;
 }
 
 /* FIXME: would be good to support this! */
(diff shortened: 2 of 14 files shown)


lightningd/plugin.c
```diff
@@ -984,9 +984,8 @@ static void plugin_config(struct plugin *plugin)
 	list_for_each(&plugin->plugin_opts, opt, list) {
 		/* Trim the `--` that we added before */
 		name = opt->name + 2;
-		if (!opt->value)
-			opt->value = "";
-		json_add_string(req->stream, name, opt->value);
+		if (opt->value)
```
ZmnSCPxj: Could this be because there is some difference between the existence of `--option` vs. its nonexistence?

That is, could it be that `--option` implicitly means `--option=""` and the non-existence of `--option` means it is not placed?

@cdecker seems to be related to nearby code, so maybe should check.

plugins/libplugin.c
```diff
@@ -480,12 +497,49 @@ static struct command_result *handle_init(struct command *init_cmd,
...
+		tal_free(opt);
+	}
+
 	if (init)
 		init(init_cmd->rpc);
 
 	return command_done_ok(init_cmd, "");
 }
 
+char *u64_option(const char *arg, u64 *i)
```
ZmnSCPxj: Almost same code (with same "This is how the manpage says to do it.  Yech.") exists in `lightningd/options.c` under function `opt_set_s32`, perhaps unifying this code in some `common/` library is better?

plugins/libplugin.c
```diff
@@ -33,10 +33,13 @@ struct plugin_conn {
 	MEMBUF(char) mb;
 };
 
+/* Connection to make RPC requests. */
+static struct plugin_conn rpc_conn;
```
ZmnSCPxj: Whaaaaat?  I suggest retaining in `command`, maybe require that objects keep a reference to this.  This commit by itself shows no reason to make this global.

plugins/libplugin.c
```diff
@@ -582,6 +599,31 @@ static void setup_command_usage(const struct plugin_command *commands,
 	}
 }
 
+static void call_plugin_timer(struct plugin_conn *rpc, struct timer *timer)
+{
+	struct plugin_timer *t = container_of(timer, struct plugin_timer, timer);
+
+	in_timer = true;
+	t->cb();
+	tal_free(t);
```
ZmnSCPxj: Instead of `tal_free` here, could it be possible to just re-parent the `t` to some `tmpctx` that is then freed?

See `trigger_invoice_waiter_resolve` in `wallet/invoices.c` for example of this technique.

That way it is OK for a timer callback to call code that destroys the  `struct plugin_timer` (which may be better if it unifies code that has to do cleanup in both "normal" and "timeout" paths of some command).

     tal_t *tmpctx = tal(NULL, int);
     struct plugin_timer *t = container_of(timer, struct plugin_timer, timer);
     tal_steal(tmpctx, t);
     in_timer = true;
     t->cb();
     tal_free(tmpctx);

plugins/libplugin.h
```diff
@@ -96,6 +98,17 @@ struct command_result *forward_result(struct command *cmd,
 				      const jsmntok_t *result,
 				      void *arg);
 
+/* Callback for timer where we expect a 'command_result'. */
+struct command_result *timer_complete(void);
+
+/* Access timer infrastructure to add a timer.
+ *
+ * Freeing this releases the timer (don't free it in timer cb though)
```
ZmnSCPxj: See other comment re: using `tal_steal` on a `tmpctx`.

plugins/libplugin.h
```diff
@@ -96,6 +98,17 @@ struct command_result *forward_result(struct command *cmd,
 				      const jsmntok_t *result,
 				      void *arg);
 
+/* Callback for timer where we expect a 'command_result'. */
+struct command_result *timer_complete(void);
```
ZmnSCPxj: Needs more clarification: should this *always* be called (possibly with the result destroyed by the caller if the timer is being used for non-command reason)?

If timer callback does not call this, `in_timer` remains `true`, should we check after return of callback that `in_timer` is not false?

cdecker:
Nice simplification :+1:

lightningd/plugin.c
rustyrussell: No, this is called when --option isn't specified at all, so that can't be it.

plugins/libplugin.c
rustyrussell: I should probably move the u32/u64 etc options into ccan, since it has all the int/uint/long/ulong values already.

plugins/libplugin.c
rustyrussell: The next commit needs it, however.

plugins/libplugin.c
rustyrussell: Yes, that's my usual technique where the caller may free it.  But there's no reason for the caller to free it, at least in the current examples.

plugins/libplugin.h
rustyrussell: I had a check, unfortunately this is not always called immediately; in particular, the timer cb for autoclean calls send_outreq().

I'll clarify the documentation, though I suspect the API will evolve as we get more users.

plugins/libplugin.c
ZmnSCPxj: Still suggest to keep it so caller is free to free it.  Rather than have a counterexample in the future.

PR #2655, opened 2019-05-21
Pay api fixes

rustyrussell:
Still have to fix the bad JSON problems, but that's a bigger effort.

### DIFF
--- a/doc/lightning-pay.7.txt
+++ b/doc/lightning-pay.7.txt
@@ -37,11 +37,10 @@ The response will occur when the payment fails or succeeds.  Once a
 payment has succeeded, calls to *pay* with the same 'bolt11' will
 succeed immediately.
 
-The command will keep finding routes and retrying the payment until
-it succeeds, or the given 'retry_for' seconds passes.
-Note that the command may stop retrying while a pending payment is
-ongoing, which you need to monitor with *listpays*.
-'retry_for' defaults to 60 seconds and can only be an integer.
+Until 'retry_for' seconds passes (default: 60), the command will keep
+finding routes and retrying the payment.  However, a payment may be
+delayed for up to `maxdelay` blocks by another node; clients should be
+prepared for this worst case.
 
 When using 'lightning-cli', you may skip optional parameters by using
 'null'.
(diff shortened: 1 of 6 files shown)


plugins/pay.c
```diff
@@ -273,7 +273,7 @@ static struct command_result *sendpay_done(struct command *cmd,
 {
 	return send_outreq(cmd, "waitsendpay",
 			   waitsendpay_done, waitsendpay_error, pc,
-			   "'payment_hash': '%s', 'timeout': 60",
+			   "'payment_hash': '%s'",
```
ZmnSCPxj: Also modify `doc/lightning-pay.7.txt` to remove the text about error 200, then.

ZmnSCPxj: For the manpage, we should probably warn:

> It is possible for the `pay` command to not return for `maxdelay` blocks *after* `retry_for` seconds, in the worst case.
> Clients should be prepared for this eventuality, which can occur in circumstances, beyond the control of the local node.

rustyrussell:
> Recommend some linebreaks...
> 
> ```
> Until 'retry_for' seconds passes (default: 60),
> the command will keep finding routes
> and retrying the payment.
> However, a payment may be delayed
> for up to `maxdelay` blocks by another node;
> clients should be prepared for this worst case.
> ```

Sure; the man page generator itself ignores linebreaks, but formatting does make it more readable.

doc/lightning-pay.7.txt
```diff
@@ -37,11 +37,10 @@ The response will occur when the payment fails or succeeds.  Once a
 payment has succeeded, calls to *pay* with the same 'bolt11' will
 succeed immediately.
 
-The command will keep finding routes and retrying the payment until
-it succeeds, or the given 'retry_for' seconds passes.
-Note that the command may stop retrying while a pending payment is
-ongoing, which you need to monitor with *listpays*.
-'retry_for' defaults to 60 seconds and can only be an integer.
+Until 'retry_for' seconds passes (default: 60), the command will keep
+finding routes and retrying the payment.  However, a payment may be
```
ZmnSCPxj: Suggest breaking at each sentence (so that rewording an earlier sentence in a paragraph will not affect succeeding sentences).

PR #2658, opened 2019-05-22
CLI : Split help output into categories

darosior:
## What 

This PR adds a category (respectively "Bitcoin", "Channels", "Network", "Payment", "Plugins", "Utility", "Developer") to each command. Here are the list of commands by category (which is also the output of `lightning-cli help`:
<details><summary>show `lightning-cli help` output</summary>

```
=== BITCOIN ===

feerates style
    Return feerate estimates, either satoshi-per-kw ({style} perkw) or satoshi-per-kb ({style} perkb).

newaddr [addresstype]
    Get a new {bech32, p2sh-segwit} (or all) address to fund a channel (default is bech32)

withdraw destination satoshi [feerate] [minconf]
    Send to {destination} address {satoshi} (or 'all') amount via Bitcoin transaction, at optional {feerate}



=== CHANNELS ===

close id [force] [timeout]
    Close the channel with {id} (either peer ID, channel ID, or short channel ID). If {force} (default false) is true, force a unilateral close after {timeout} seconds (default 30), otherwise just schedule a mutual close later and fail after timing out.

fundchannel id satoshi [feerate] [announce] [minconf]
    Fund channel with {id} using {satoshi} (or 'all') satoshis, at optional {feerate}. Only use outputs that have {minconf} confirmations.

getroute id msatoshi riskfactor [cltv] [fromid] [fuzzpercent] [exclude] [maxhops]
    Show route to {id} for {msatoshi}, using {riskfactor} and optional {cltv} (default 9). If specified search from {fromid} otherwise use this node as source. Randomize the route with up to {fuzzpercent} (default 5.0). {exclude} an array of short-channel-id/direction (e.g. [ '564334x877x1/0', '564195x1292x0/1' ]) from consideration. Set the {maxhops} the route can take (default 20).

listchannels [short_channel_id] [source]
    Show channel {short_channel_id} or {source} (or all known channels, if not specified)

listforwards 
    List all forwarded payments and their information

setchannelfee id [base] [ppm]
    Sets specific routing fees for channel with {id} (either peer ID, channel ID, short channel ID or 'all'). Routing fees are defined by a fixed {base} (msat) and a {ppm} (proportional per millionth) value. If values for {base} or {ppm} are left out, defaults will be used. {base} can also be defined in other units, for example '1sat'. If {id} is 'all', the fees will be applied for all channels. 



=== NETWORK ===

connect id [host] [port]
    Connect to {id} at {host} (which can end in ':port' if not default). {id} can also be of the form id@host

disconnect id [force]
    Disconnect from {id} that has previously been connected to using connect; with {force} set, even if it has a current channel

listnodes [id]
    Show node {id} (or all, if no {id}), in our local network view

listpeers [id] [level]
    Show current peers, if {level} is set, include logs for {id}

ping id [len] [pongbytes]
    Send peer {id} a ping of length {len} (default 128) asking for {pongbytes} (default 128)



=== PAYMENT ===

autocleaninvoice [cycle_seconds] [expired_by]
    Set up autoclean of expired invoices. 

decodepay bolt11 [description]
    Decode {bolt11}, using {description} if necessary

delexpiredinvoice [maxexpirytime]
    Delete all expired invoices that expired as of given {maxexpirytime} (a UNIX epoch time), or all expired invoices if not specified

delinvoice label status
    Delete unpaid invoice {label} with {status}

invoice msatoshi label description [expiry] [fallbacks] [preimage] [exposeprivatechannels] [dev-routes]
    Create an invoice for {msatoshi} with {label} and {description} with optional {expiry} seconds (default 1 hour), optional {fallbacks} address list(default empty list) and optional {preimage} (default autogenerated)

listinvoices [label]
    Show invoice {label} (or all, if no {label})

listpayments [bolt11] [payment_hash]
    Show outgoing payments

listpays [bolt11]
    List result of payment {bolt11}, or all

listsendpays [bolt11] [payment_hash]
    Show sendpay, old and current, optionally limiting to {bolt11} or {payment_hash}.

pay bolt11 [msatoshi] [label] [riskfactor] [maxfeepercent] [retry_for] [maxdelay] [exemptfee]
    Send payment specified by {bolt11} with {amount}

paystatus [bolt11]
    Detail status of attempts to pay {bolt11}, or all

sendpay route payment_hash [label] [msatoshi] [bolt11]
    Send along {route} in return for preimage of {payment_hash}

waitanyinvoice [lastpay_index]
    Wait for the next invoice to be paid, after {lastpay_index} (if supplied)

waitinvoice label
    Wait for an incoming payment matching the invoice with {label}, or if the invoice expires

waitsendpay payment_hash [timeout]
    Wait for payment attempt on {payment_hash} to succeed or fail, but only up to {timeout} seconds.



=== PLUGINS ===

gui 
    Launches the Qt GUI



=== UTILITY ===

check command_to_check
    Don't run {command_to_check}, just verify parameters.

getinfo 
    Show information about this node

getlog [level]
    Show logs, with optional log {level} (info|unusual|debug|io)

help [command]
    List available commands, or give verbose help on one {command}.

listconfigs [config]
    List all configuration options, or with [config], just that one.

listfunds 
    Show available funds from the internal wallet

stop 
    Shut down the lightningd process



=== DEVELOPER ===

dev-compact-gossip-store 
    Ask gossipd to rewrite the gossip store.

dev-crash 
    Crash lightningd by calling fatal()

dev-fail id
    Fail with peer {id}

dev-forget-channel id [short_channel_id] [force]
    Forget the channel with peer {id}, ignore UTXO check with {force}='true'.

dev-ignore-htlcs id ignore
    Set ignoring incoming HTLCs for peer {id} to {ignore}

dev-listaddrs [bip32_max_index]
    Show addresses list up to derivation {index} (default is the last bip32 index)

dev-memdump 
    Show memory objects currently in use

dev-memleak 
    Show unreferenced memory objects

dev-query-channel-range id first num
    Query peer {id} for short_channel_ids for {first} block + {num} blocks

dev-query-scids id scids
    Query peer {id} for [scids]

dev-reenable-commit id
    Re-enable the commit timer on peer {id}

dev-rescan-outputs 
    Synchronize the state of our funds with bitcoind

dev-rhash secret
    Show SHA256 of {secret}

dev-send-timestamp-filter id first range
    Send peer {id} the timestamp filter {first} {range}

dev-set-max-scids-encode-size max
    Set {max} bytes of short_channel_ids per reply_channel_range

dev-sign-last-tx id
    Sign and show the last commitment transaction with peer {id}

dev-slowcmd [msec]
    Torture test for slow commands, optional {msec}

dev-suppress-gossip 
    Stop this node from sending any more gossip.

---
```
</details>

The categories are sorted alphabetically, apart for the developer commands category which is appended at the end. Commands are *(will, not rebased yet)* also sorted alphabetically (#2646).
The raw json is not sorted in categories because it would break the API and it is quite useless, the sorting can be done on the other side, thanks to the new json member `category` contained in the commands returned by `help` (see below).  
  
## How 
A category member is added to the `json_command` structure, I've added it for all existing native commands and should now be specified if a new (native) command is added (there is no "unsorted" category).
The category `PLUGIN` is added to commands created with `plugin_rpcmethod_add()` in [`lightningd/plugins.c`](https://github.com/darosior/lightning/blob/123c7ab55b21022aadedf7cf41a8aa8fc83042f2/lightningd/plugin.c#L675).
A category identifier is added to the json created by [`json_add_help_command`](https://github.com/darosior/lightning/blob/cc08306d75ee5a69c45c565e6ea42036cc193985/lightningd/jsonrpc.c#L324).
On the CLI side, the [`human_help`](https://github.com/darosior/lightning/blob/390ab250452020a4f035e4e82aa75910b4adca0a/cli/lightning-cli.c#L96) function has been heavily modified to not anymore just `printf` commands on the fly, but to instead populate an array of pointers to commands tokens for each category. We then iterate through these (seven) arrays to `printf` commands under their respective category.

## Some thoughts
I arbitrarily picked the categories and filled them with the commands, there might some mistakes or some better category names choice.
I arbitrarily chose the way to render the category : 
```
=== CATEGORY ===
\n
command a
\n
command b
\n
\n
\n
=== CATEGORY ===
\n
...
```
There might be some prettier ways to do it ^^.


lightningd/plugin.c
```diff
@@ -672,6 +672,7 @@ static bool plugin_rpcmethod_add(struct plugin *plugin,
 	cmd = notleak(tal(plugin, struct json_command));
 	cmd->name = json_strdup(cmd, buffer, nametok);
 	cmd->description = json_strdup(cmd, buffer, desctok);
+	cmd->category = CMD_PLUGIN;
```
ZmnSCPxj: Should not plugins indicate the category of their commands directly, via some optional interface?

darosior: I think it is useful to tag plugin commands differently than native commands. However we can always do both.

ZmnSCPxj: I suggest that plugins should be "seamless".  We are moving functionality from "core" of lightningd to plugins, see `pay` and `autocleaninvoice`.

darosior: I agree for not externally installed plugins like `pay` and `waitinvoice`, but I suggest that externally installed plugins (i.e. most of the time little Python scripts) stay tagged as "add-ons".

ZmnSCPxj: That's why you should be able to indicate the command category using some optional itnerface, not just mark a command as "plugin" just because it comes from a plugin.

m-schmoock: @ZmnSCPxj I think plugin command categories can easily be done in an separate/additional PR

darosior:
Rebased, added the support to set the command category by plugins as requested by @ZmnSCPxj . Added the `Payment` category to commands created in the `pay` and `autocleaninvoice` plugins. I will add it to pylightning too, and add notes to CHANGELOG.
Edited the `lightning-cli help` output above : commands are now sorted both alphabetically and by category \o/ .

lightningd/plugin.c
```diff
@@ -671,6 +672,12 @@ static bool plugin_rpcmethod_add(struct plugin *plugin,
 
 	cmd = notleak(tal(plugin, struct json_command));
 	cmd->name = json_strdup(cmd, buffer, nametok);
+	// If the command category has been specified plugin side, set it
+	if (categorytok)
+		cmd->category = atoi(buffer + categorytok->start); // category is a number
```
ZmnSCPxj: Mildly suggest using a string in the JSON interface rather than a number. Possible to use a string in pylightning but that means every plugin library in every language needs to have the string->enum conversion, when we could put the string->enum conversion inside lightningd.  But no big deal.

lightningd/plugin.c
```diff
@@ -631,11 +631,12 @@ static bool plugin_rpcmethod_add(struct plugin *plugin,
 				 const char *buffer,
 				 const jsmntok_t *meth)
 {
-	const jsmntok_t *nametok, *desctok, *longdesctok, *usagetok;
+	const jsmntok_t *nametok, *categorytok, *desctok, *longdesctok, *usagetok;
 	struct json_command *cmd;
 	const char *usage;
 
 	nametok = json_get_member(buffer, meth, "name");
+	categorytok = json_get_member(buffer, meth, "category");
```
cdecker: I'm not sure we actually want to allow plugins to change their category at all. It would be nice if we could group their RPC methods by the plugin that exposes them, but that doesn't play nicely with the numeric category used in the rest of the PR.

Then again, having `pay` and `autoclean` be categorized in plugins seems weird as well.

darosior: > I'm not sure we actually want to allow plugins to change their category at all. It would be nice if we could group their RPC methods by the plugin that exposes them, but that doesn't play nicely with the numeric category used in the rest of the PR.

Yes, I did the code for Pylightning similarly to what I added to libplugin.c and did not push it because I remarked the same. Maybe we can assume that C plugins are internal and that others external so leave the possibility to chose the category for internal plugins in libplugin.c and not in Pylightning ? Then we can leave this PR as a base with unsorted plugin commands and i'll add another one that pass from : 
```
=== PLUGIN ===
"command a" from plugin 1

"command b" from plugin 2

"command c" from plugin 2
```
to
```
=== PLUGIN ===
plugin 1
    command a

plugin 2
    command b
    command c
```

cli/lightning-cli.c
```diff
@@ -95,20 +96,167 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+			curr += 2;
+			printf("%.*s\n", curr->end - curr->start, buffer + curr->start);
+			// Then to commmand description, 4 tokens forward
+			curr += 4;
+			printf("    %.*s\n\n", curr->end - curr->start, buffer + curr->start);
+		}
+	}
+	tal_free(bitcoin_commands);
+
+	printf("\n\n=== CHANNELS ===\n\n");
```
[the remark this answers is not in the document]
darosior: I could have done that if I sorted commands directly from `lightningd` and output them as :
```
help : {
    "bitcoin": {
        {command},
        {command},
        ...
    },
    "channels": {
        {command},
        {command},
        ...
    },
    ...
}
```
But I did not want to deeply change the JSON output just to sort the in the CLI.

cli/lightning-cli.c
```diff
@@ -95,20 +96,167 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+	 *     "verbose": "The command's detailed description"
+	 * }
+	 *
+	 * There are seven categories ("Bitcoin"(0), "Channels"(1), "Network"(2), "Payment"(3),
+	 * "Plugins"(4), "Utilities"(5), "Developer"(6)). We will iterate through all commands,
+	 * adding a pointer to the command token to the corresponding array. We'll then just
+	 * have to iterate through each category's array to printf the commands.
+	 */
+	/* A token array by category */
+	const jsmntok_t **bitcoin_commands = tal_arrz(tmpctx, const jsmntok_t*, 1);
```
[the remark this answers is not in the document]
darosior: I tried to
```C
const jsmntok_t **bitcoin_commands = tal_arr(tmpctx, const jsmntok_t*, 0);
```
but this didn't work and I assumed I could not instanciate an array with a 0 size with tal, as in classic C.

cli/lightning-cli.c
```diff
@@ -95,20 +96,167 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
 	for (i = 0; i < help_array->size; i++) {
-		curr += 2;
-		printf("%.*s\n", curr->end - curr->start, buffer + curr->start);
-		curr += 2;
-		printf("    %.*s\n\n", curr->end - curr->start, buffer + curr->start);
-		curr += 2;
-		/* advance to next command */
-		curr++;
+		// Go to the category field, 4 tokens forwards
+		curr += 4;
```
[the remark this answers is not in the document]
darosior: I'm new to jsmn and that was what was used, I assumed this was how it should be done : what are the approppriate JSON accessors ?

lightningd/plugin.c
ZmnSCPxj: > Maybe we can assume that C plugins are internal and that others external so leave the possibility to chose the category for internal plugins in libplugin.c and not in Pylightning ?

The language a plugin is written in should not matter -- we may someday want to add a sufficiently-useful plugin to the "stock" C-lightning release, and if it be written in Brainfuck, as long as it is useful, it should be acceptable.

Perhaps category should be a string and the provenance of the command (internal or plugin) be an enumerated field.

darosior:
Rebased to squash the two commits which changed libplugin, pay and autoclean. Ammended the first commit to remove
```C
json_add_num(response, "category", json_command->category);
```
and added it to the next commit when all `category` fields have been initialized.

cli/lightning-cli.c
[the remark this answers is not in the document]
ZmnSCPxj: https://github.com/ElementsProject/lightning/blob/82ff580a6649fc8751f6cb32b6ea169d224c0809/common/json.h#L83-L86

So something like:

   jsmntok_t const *categorytok = json_delve(buffer, curr, ".category");

cli/lightning-cli.c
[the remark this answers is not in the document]
cdecker: No I mean just the way it is being processed here is a bit repetitive, and could be simplified.

cli/lightning-cli.c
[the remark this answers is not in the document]
cdecker: It's quite a common pattern in the code-base, what was the compiler error you got?

cli/lightning-cli.c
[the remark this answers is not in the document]
cdecker: Oh, I missed that we did the offset thing before. Then we should probably split that out into a separate PR, for stabilization :wink:

cli/lightning-cli.c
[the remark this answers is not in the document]
ZmnSCPxj: How did it not work?  Counterexamples:

https://github.com/ElementsProject/lightning/blob/4811024518201dfea57437a0eeef4c62820906c9/bitcoin/script.c#L166

https://github.com/ElementsProject/lightning/blob/a00c357854bffd8add6f2a501110df0e56db6ce8/common/decode_short_channel_ids.c#L49-L54

https://github.com/ElementsProject/lightning/blob/b4e7b198e6ef8db2456dbc12ae6edc8b76368087/connectd/connectd.c#L982-L984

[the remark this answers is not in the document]
darosior: Ok I must have made a mistake then ... :smile: 
I'll correct it and push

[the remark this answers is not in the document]
darosior: Thanks for the counterexamples

cli/lightning-cli.c
```diff
@@ -95,20 +96,167 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+		category = atoi(buffer + curr->start);
+		// Get back to the command, which will be stored
+		curr -= 4;
+		switch(category) {
```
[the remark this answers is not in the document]
darosior: Should I had a test for backward compatibility then ? I assumed lightning-cli to be compiled at the same time and used at the same version..

lightningd/plugin.c
darosior: I don't say that the language matters but that for now internal plugins are written with libcplugin and that I think it's ok to let the possibility of specifying the command category in it (and only this one).

darosior:
According to the review, I initialized the commands arrays with a 0 size and used [json accessors](https://github.com/ElementsProject/lightning/blob/82ff580a6649fc8751f6cb32b6ea169d224c0809/common/json.h) instead of increasing the cursor with constant values in the `help_array` iteration. Thanks @cdecker and @ZmnSCPxj for the hints :D 

The iteration through each commands array is still a bit repetitive but I can't find a more effective way to do it.

@ZmnSCPxj proposed to make the `category` field a string, what do others think about that ?

cli/lightning-cli.c
```diff
@@ -95,20 +96,139 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+			tal_arr_expand(&utility_commands, curr);
+			break;
+		case CMD_DEVELOPER:
+			tal_arr_expand(&developer_commands, curr);
+			break;
+		}
+	}
+
+	/* Loop through each array of *jsmntok_t, sorted by category. Categories are sorted
+	 * alphabetically (apart the developer one which is appended at the end).
```
niftynei: nit: `apart the developer` -> `apart from the ..`

cli/lightning-cli.c
```diff
@@ -95,20 +96,139 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+	 */
+
+	printf("=== BITCOIN ===\n\n");
+	for (i = 0; i < tal_count(bitcoin_commands); i++) {
+		// Reuse the cursor to iterate through the 8 tokens of our command
```
niftynei: same note about `//` vs `/*` style comments.

cli/lightning-cli.c
```diff
@@ -95,20 +96,139 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+			tal_arr_expand(&developer_commands, curr);
+			break;
+		}
+	}
+
+	/* Loop through each array of *jsmntok_t, sorted by category. Categories are sorted
+	 * alphabetically (apart the developer one which is appended at the end).
+	 */
+
+	printf("=== BITCOIN ===\n\n");
```
niftynei: If you had a string mapping for categories, you could condense this code down a lot by using that as the heading for each sub-genre.

cli/lightning-cli.c
darosior: thanks !

cli/lightning-cli.c
darosior: Thank you, besides the hint you comment made me understand what @cdecker meant by "array of arrays" :D

darosior:
@niftynei thank you for the review.
I rebased to : 
- make the `category` field in the object created by `json_add_help()` a string ("bitcoin", "channels", "network", "payment", "plugin", "utility", "developer"
- change lightningd/plugin.c, libplugin.c, pay, and autoclean accordingly
- in the CLI : use an array of arrays of json tokens and a a string mapping between categories and their string representation, to merge the 7 for loops into 2 (the only downside is that now category headings are lowercase).

darosior:
> You could really do away with the enums entirely and just use strings; any new string additions would create new categories -- gives plugin devs more flexibility in either fitting into an existing category or creating new ones where appropriate.

Merge conflict with @cdecker ^^ :
> I'm not sure we actually want to allow plugins to change their category at all.

I for one think that categories are vague enough to not make possible for plugin devs to create anymore, but that it can be useful for them to specify it at command creation.

cli/lightning-cli.c
```diff
@@ -95,20 +96,88 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+			command_category = json_get_member(buffer, curr, "category");
+			printf("Invalid category : \"%.*s\"\n", command_category->end - command_category->start,
+					buffer + command_category->start);
+		}
+	}
+
+	/* Iterate through all categories and printf commands */
+	for (i = 0; i < tal_count(commands); ++i) {
+		for (j = 0; j < tal_count(commands[i]); ++j) {
+			if (j == 0)
```
cdecker: This if statement can be omitted by moving the `printf` outside of the loop.

cli/lightning-cli.c
```diff
@@ -95,20 +96,88 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+	 * all contained arrays
+	 */
+	const jsmntok_t ***commands = tal_arr(tmpctx, const jsmntok_t**, 7);
+	for (i = 0; i < tal_count(commands); ++i) {
+		commands[i] = tal_arr(tmpctx, const jsmntok_t*, 0);
 	}
+	/* A convenient mapping to compare categories in command objects (string) with our
+	 * categories enum, and to display category names when iterating through the arrays
+	 */
+	const char **category_names = tal_arr(tmpctx, const char*, 7);
```
cdecker: These can likely be statically initialized and do not need to be allocated on the heap. It'd also allow us to use `ARRAY_SIZE` instead of the constant `7` in the `tal_arr` initialization above.

cli/lightning-cli.c
```diff
@@ -95,20 +96,88 @@ static size_t human_readable(const char *buffer, const jsmntok_t *t, char term)
...
+				printf("=== %s ===\n\n", category_names[i]);
+			curr = commands[i][j];
+			/* Go to command name, 2 tokens forward */
+			curr += 2;
+			printf("%.*s\n", curr->end - curr->start, buffer + curr->start);
+			/* Go to command description, 4 tokens forward */
+			curr += 4;
+			printf("    %.*s\n\n", curr->end - curr->start, buffer + curr->start);
+			if (j == tal_count(commands[i]))
+				printf("\n\n");
```
cdecker: Just like the `if` above this can be moved outside of the loop and simplified.

darosior:
I force pushed to : 
- statically intialize `category_names` and use `ARRAY_SIZE` after instead of `7`
- move the `printf`s (I surely wasnt awake when I made that ^^ )

ZmnSCPxj:
@rustyrussell agree.

ACK 3a228af505aa64a0ea38960c64134db0c82619ab

Re: testing "incorrect" category names, maybe just fine to leave as-is.  People who care about it will notice new category and fix in plugin, and there may be future plugins with completely new features that require commands that do not fit into current categories: I believe the extra flexibility of letting the command author declare any string as category is superior despite typos.

rustyrussell:
Sure, but our tests will only test our code...

darosior:
@rustyrussell I ammended the change, thank you for having writing the whole CLI code so I just had to `curl` it from your branch.
Since we want to allow plugins to set any category they want at command creation, I'll commit a change for Pylightning accordingly.

darosior:
Finally, this PR adds:
- A new `category` field (string) in the command structure
- Adds it to each native commands
- Allows plugins to specify it and even to create new ones at command creation (and adds it to libplugin.c and pylightning)
- Uses this `category` field to parse commands for the CLI's human readable help

Here is an output of the help command :
<details>
<p>
```=== bitcoin ===

feerates style
    Return feerate estimates, either satoshi-per-kw ({style} perkw) or satoshi-per-kb ({style} perkb).

newaddr [addresstype]
    Get a new {bech32, p2sh-segwit} (or all) address to fund a channel (default is bech32)

withdraw destination satoshi [feerate] [minconf]
    Send to {destination} address {satoshi} (or 'all') amount via Bitcoin transaction, at optional {feerate}

=== channels ===

close id [force] [timeout]
    Close the channel with {id} (either peer ID, channel ID, or short channel ID). If {force} (default false) is true, force a unilateral close after {timeout} seconds (default 30), otherwise just schedule a mutual close later and fail after timing out.

fundchannel id satoshi [feerate] [announce] [minconf]
    Fund channel with {id} using {satoshi} (or 'all') satoshis, at optional {feerate}. Only use outputs that have {minconf} confirmations.

getroute id msatoshi riskfactor [cltv] [fromid] [fuzzpercent] [exclude] [maxhops]
    Show route to {id} for {msatoshi}, using {riskfactor} and optional {cltv} (default 9). If specified search from {fromid} otherwise use this node as source. Randomize the route with up to {fuzzpercent} (default 5.0). {exclude} an array of short-channel-id/direction (e.g. [ '564334x877x1/0', '564195x1292x0/1' ]) from consideration. Set the {maxhops} the route can take (default 20).

listchannels [short_channel_id] [source]
    Show channel {short_channel_id} or {source} (or all known channels, if not specified)

listforwards 
    List all forwarded payments and their information

setchannelfee id [base] [ppm]
    Sets specific routing fees for channel with {id} (either peer ID, channel ID, short channel ID or 'all'). Routing fees are defined by a fixed {base} (msat) and a {ppm} (proportional per millionth) value. If values for {base} or {ppm} are left out, defaults will be used. {base} can also be defined in other units, for example '1sat'. If {id} is 'all', the fees will be applied for all channels. 

=== network ===

connect id [host] [port]
    Connect to {id} at {host} (which can end in ':port' if not default). {id} can also be of the form id@host

disconnect id [force]
    Disconnect from {id} that has previously been connected to using connect; with {force} set, even if it has a current channel

listnodes [id]
    Show node {id} (or all, if no {id}), in our local network view

listpeers [id] [level]
    Show current peers, if {level} is set, include logs for {id}

ping id [len] [pongbytes]
    Send peer {id} a ping of length {len} (default 128) asking for {pongbytes} (default 128)

=== payment ===

autocleaninvoice [cycle_seconds] [expired_by]
    Set up autoclean of expired invoices. 

decodepay bolt11 [description]
    Decode {bolt11}, using {description} if necessary

delexpiredinvoice [maxexpirytime]
    Delete all expired invoices that expired as of given {maxexpirytime} (a UNIX epoch time), or all expired invoices if not specified

delinvoice label status
    Delete unpaid invoice {label} with {status}

invoice msatoshi label description [expiry] [fallbacks] [preimage] [exposeprivatechannels] [dev-routes]
    Create an invoice for {msatoshi} with {label} and {description} with optional {expiry} seconds (default 1 hour), optional {fallbacks} address list(default empty list) and optional {preimage} (default autogenerated)

listinvoices [label]
    Show invoice {label} (or all, if no {label})

listpayments [bolt11] [payment_hash]
    Show outgoing payments

listpays [bolt11]
    List result of payment {bolt11}, or all

listsendpays [bolt11] [payment_hash]
    Show sendpay, old and current, optionally limiting to {bolt11} or {payment_hash}.

pay bolt11 [msatoshi] [label] [riskfactor] [maxfeepercent] [retry_for] [maxdelay] [exemptfee]
    Send payment specified by {bolt11} with {amount}

paystatus [bolt11]
    Detail status of attempts to pay {bolt11}, or all

sendpay route payment_hash [label] [msatoshi] [bolt11]
    Send along {route} in return for preimage of {payment_hash}

waitanyinvoice [lastpay_index]
    Wait for the next invoice to be paid, after {lastpay_index} (if supplied)

waitinvoice label
    Wait for an incoming payment matching the invoice with {label}, or if the invoice expires

waitsendpay payment_hash [timeout]
    Wait for payment attempt on {payment_hash} to succeed or fail, but only up to {timeout} seconds.

=== plugin ===

gui 
    Launches the Qt GUI

=== utility ===

check command_to_check
    Don't run {command_to_check}, just verify parameters.

getinfo 
    Show information about this node

getlog [level]
    Show logs, with optional log {level} (info|unusual|debug|io)

help [command]
    List available commands, or give verbose help on one {command}.

listconfigs [config]
    List all configuration options, or with [config], just that one.

listfunds 
    Show available funds from the internal wallet

stop 
    Shut down the lightningd process

=== developer ===

dev-compact-gossip-store 
    Ask gossipd to rewrite the gossip store.

dev-crash 
    Crash lightningd by calling fatal()

dev-fail id
    Fail with peer {id}

dev-forget-channel id [short_channel_id] [force]
    Forget the channel with peer {id}, ignore UTXO check with {force}='true'.

dev-ignore-htlcs id ignore
    Set ignoring incoming HTLCs for peer {id} to {ignore}

dev-listaddrs [bip32_max_index]
    Show addresses list up to derivation {index} (default is the last bip32 index)

dev-memdump 
    Show memory objects currently in use

dev-memleak 
    Show unreferenced memory objects

dev-query-channel-range id first num
    Query peer {id} for short_channel_ids for {first} block + {num} blocks

dev-query-scids id scids
    Query peer {id} for [scids]

dev-reenable-commit id
    Re-enable the commit timer on peer {id}

dev-rescan-outputs 
    Synchronize the state of our funds with bitcoind

dev-rhash secret
    Show SHA256 of {secret}

dev-send-timestamp-filter id first range
    Send peer {id} the timestamp filter {first} {range}

dev-set-max-scids-encode-size max
    Set {max} bytes of short_channel_ids per reply_channel_range

dev-sign-last-tx id
    Sign and show the last commitment transaction with peer {id}

dev-slowcmd [msec]
    Torture test for slow commands, optional {msec}

dev-suppress-gossip 
    Stop this node from sending any more gossip.

---
run `lightning-cli help <command>` for more information on a specific command```
</p>
</details>

rustyrussell:
Thanks!  I think this wins a prize for the most commented PR in recent history, too!

Issue #2659, opened 2019-05-22
pay plugin regression: no longer does value randomization


ZmnSCPxj:
Review of code shows we do not actually randomize the value that reaches the payee anymore.

Solutions are:

1.  Just remove this from `doc/lightning-pay.7.txt`.  Is value randomization even useful?? Also update that we now use shadow routes (though the shadow route logic seems to completely ignore `maxdelay` and does not seem to have fallback when `maxdelay` is reached).
2.  Reimplement value randomization by computing fees backwards (i.e. from payer to payee) and getting a max acceptable value randomization target, getting a random number between 0 and the randomization target, then recompute the values forwards (from payee to payer).

PR #2669, opened 2019-05-23
doc/lightning-pay.7.txt: Remove references to value randomization, since it is no longer implemented, and describe shadow routes.

ZmnSCPxj:
Fixes: #2659

Value randomization is not implemented since `pay` was made into a plugin.
Nobody has been asking for it, so do not implement it and just remove
from documentation.

Shadow routes is implemented, and not describe, so describe it now.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2669/commits/e1a6650e1bdc6e2993cabf36e923c0a9071e111e

Issue #1936, opened 2018-09-13
Batch Funding Transaction for opening several channels (feature request)


renepickhardt:
When new nodes come to the network and they run the autopilot (c.f.: https://github.com/ElementsProject/lightning/pull/1888 ) they might want to open several channels. I think there is no need to do it (as currently implemented) sequentially but we could adapt the process so that it is just transaction with 1 output for each channel to be opened. This would save space and allow for more channels to be opened per block. 

I believe the API of c-lightning could be extended to support a batch channel open (and consequently also a batch channel closing command. I believe there is quite some protocol overhead as communication with all channel partners has to be established and commitment txs have to be signed before the batch funding tx could be published. However I think this would not interfere with the current protocol specification for channel establishment. 

I think this should not be an urgent feature but I believe it will be very useful on the long term as it saves precious blockchain space

I ping @rustyrussell as he asked me to open this as an issue here.

leishman:
This would be an awesome feature. 💯

ZmnSCPxj:
* [ ] split `fundchannel` #2672
* [ ] split `withdraw` #2713
* [ ] implement `multifundchannel`

ZmnSCPxj:
People are trying to implement multi-channel funding txes and getting badly bitten.

First: You really really really need to implement #2740 first.

Secondly:

`fundchannel_complete` is arguably a bad name.  It completes the `openingd` process but does not, in fact, complete the entire funding process: the completion of the funding process is the transition from `CHANNELD_AWAITING_LOCKIN` to `CHANNELD_NORMAL`.

So, let us clarify the proper process of funding (1 or more) channels:

1.  Start `fundchannel_start` on the node(s) you want to open channels to.
  * Note down the addresses specified.
2.  Create ***but do not sign or broadcast*** a transaction which pays the ***exact*** amount to the above address from `fundchannel_start`.
   * If multiple nodes, make a single transaction that pays out all of those.
   * Determine the txid of this ***unsigned, unbroadcast*** transaction.
3.  Invoke `fundchannel_complete` on the nodes.
   * If *any* fail, abort: trigger errors on all of the ***other*** channels that succeeded `fundchannel_complete`.  `dev-fail` might work, though maybe it will try to unilaterally close instead of just forgetting the channel, it is a `dev-` command so you need DEVELOPER option during compile.  See #2740 .
4.  If all `fundchannel_complete` succeeded, sign and broadcast the funding transaction.
[...]

PR #2672, opened 2019-05-24
External wallet funding 

### LINKED ISSUE #644: Feature Request: open channel directly from bitcoind wallet funds
If the user already has a `bitcoind` wallet, this would reduce cost and friction for the initial funding step (`cli/lightning-cli newaddr`). A generic approach would allow funding a channel from any wallet software that has the prerequisite RPC commands.

niftynei:
Adds RPC commands for funding a channel via an external wallet.

There's a lot of caveats around the expected behavior of the wallet (uses SegWit ie non-malleable inputs, doesn't broadcast until `continue` has completed, provides the _correct_ funding amount to `fundchannel_start`) that this doesn't have any way of enforcing. As such, it's behind an experimental flag -- a good plugin could probably paper over the majority of these issues. 

RPCs added are as follows:
 -  `fundchannel_start`. Returns a bech32 address that is the funding tx destination. 
 - `fundchannel_continue`. Takes a funding txid and vout, returns the channel_id and confirmation that we've secured the commitment transactions for this channel.
 - `fundchannel_cancel`. Instead of `continue`, halts the fund channel flow.

Fixes #644

### DIFF
--- a/common/addr.c
+++ b/common/addr.c
@@ -1,0 +1,22 @@
+#include "addr.h"
+#include <bitcoin/script.h>
+#include <common/bech32.h>
+
+/* Returns NULL if the script is not a P2WPKH or P2WSH */
+char *encode_scriptpubkey_to_addr(const tal_t *ctx,
+				   const char *hrp,
+				   const u8 *scriptPubkey)
+{
+	char *out;
+	size_t scriptLen = tal_bytelen(scriptPubkey);
+
+        /* Check that scriptPubkey is P2WSH or P2WPKH */
+	if (!is_p2wsh(scriptPubkey, NULL) && !is_p2wpkh(scriptPubkey, NULL))
+                return NULL;
+
+	out = tal_arr(ctx, char, 73 + strlen(hrp));
+	if (!segwit_addr_encode(out, hrp, 0, scriptPubkey + 2, scriptLen - 2))
+		return tal_free(out);
+
+	return out;
+}
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -259,6 +259,21 @@ static void funding_broadcast_failed(struct channel *channel,
 	/* Keep in state CHANNELD_AWAITING_LOCKIN until (manual) broadcast */
 }
 
+static void funding_success(struct channel *channel)
+{
+	struct json_stream *response;
+	struct funding_channel *fc = channel->peer->uncommitted_channel->fc;
+	struct command *cmd = fc->cmd;
+
+	response = json_stream_success(cmd);
+	json_object_start(response, NULL);
+	json_add_string(response, "channel_id",
+			type_to_string(tmpctx, struct channel_id, &fc->cid));
+	json_add_bool(response, "commitments_secured", true);
+	json_object_end(response);
+	was_pending(command_success(cmd, response));
+}
+
 static void funding_broadcast_success(struct channel *channel)
 {
 	struct json_stream *response;
@@ -336,58 +351,20 @@ failed:
 	tal_free(fc->uc);
 }
 
-static void opening_funder_finished(struct subd *openingd, const u8 *resp,
-				    const int *fds,
-				    struct funding_channel *fc)
+static bool compose_and_broadcast_tx(struct lightningd *ld,
+				     const u8 *resp,
+				     struct funding_channel *fc,
+				     struct channel_info *channel_info,
+				     struct channel *channel,
+				     struct bitcoin_txid *expected_txid,
+				     u32 feerate)
 {
 	u8 *msg;
-	struct channel_info channel_info;
-	struct bitcoin_tx *fundingtx;
-	struct bitcoin_txid funding_txid, expected_txid;
 	struct pubkey changekey;
-	struct bitcoin_signature remote_commit_sig;
-	struct bitcoin_tx *remote_commit;
 	u16 funding_outnum;
-	u32 feerate;
+	struct bitcoin_tx *fundingtx;
 	struct amount_sat change;
-	struct channel *channel;
-	struct lightningd *ld = openingd->ld;
-	u8 *remote_upfront_shutdown_script;
-	struct per_peer_state *pps;
-
-	/* This is a new channel_info.their_config so set its ID to 0 */
-	channel_info.their_config.id = 0;
-
-	if (!fromwire_opening_funder_reply(resp, resp,
-					   &channel_info.their_config,
-					   &remote_commit,
-					   &remote_commit_sig,
-					   &pps,
-					   &channel_info.theirbase.revocation,
-					   &channel_info.theirbase.payment,
-					   &channel_info.theirbase.htlc,
-					   &channel_info.theirbase.delayed_payment,
-					   &channel_info.remote_per_commit,
-					   &fc->uc->minimum_depth,
-					   &channel_info.remote_fundingkey,
-					   &expected_txid,
-					   &funding_outnum,
-					   &feerate,
-					   &fc->uc->our_config.channel_reserve,
-					   &remote_upfront_shutdown_script)) {
-		log_broken(fc->uc->log,
-			   "bad OPENING_FUNDER_REPLY %s",
-			   tal_hex(resp, resp));
-		was_pending(command_fail(fc->cmd, LIGHTNINGD,
-					 "bad OPENING_FUNDER_REPLY %s",
-					 tal_hex(fc->cmd, resp)));
-		goto failed;
-	}
-	per_peer_state_set_fds_arr(pps, fds);
-
-	log_debug(ld->log,
-		  "%s", type_to_string(tmpctx, struct pubkey,
-				       &channel_info.remote_per_commit));
+	struct bitcoin_txid funding_txid;
 
 	/* Generate the funding tx. */
 	if (!amount_sat_eq(fc->wtx->change, AMOUNT_SAT(0))
@@ -398,7 +375,7 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 	fundingtx = funding_tx(tmpctx, &funding_outnum,
 			       fc->wtx->utxos, fc->wtx->amount,
 			       &fc->uc->local_funding_pubkey,
-			       &channel_info.remote_fundingkey,
+			       &channel_info->remote_fundingkey,
 			       fc->wtx->change, &changekey,
 			       ld->wallet->bip32_base);
 
@@ -420,7 +397,7 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 
 	bitcoin_txid(fundingtx, &funding_txid);
 
-	if (!bitcoin_txid_eq(&funding_txid, &expected_txid)) {
+	if (!bitcoin_txid_eq(&funding_txid, expected_txid)) {
 		log_broken(fc->uc->log,
 			   "Funding txid mismatch:"
 			   " amount %s change %s"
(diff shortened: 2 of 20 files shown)


niftynei:
Wanted to point out that we have no way of confirming that the amount someone says they're going to fund a channel for is actually the amount that they fund the channel for.  We can correct/fix for this once the transaction is broadcast.  

We have to be given a funding amount, however, as the channel establishment protocol calls for a funding amount in `open_channel`.

ZmnSCPxj:
This needs to be designed carefully. In particular the external wallet must be capable of providing a signed tx without broadcasting it (and clients need to be very aware that it is `lightningd` that must broadcast the funding tx).

More concretely, it seems a good amount of duplication of code, thiugh I have not checked all details.

I believe the same messages can be leveraged to implement `multifundchannel`. Given sufficient hooks in the normal `lightningd` wallet (a flag to `withdraw` that says "give me the signed tx but do not broadcast it", the ability to specify multiple targets to `withdraw`, and a new command to unreserve UTXOs of an unbroadcast tx (in case the succeeding funding attempt fails)), plus the ability to tell `fundchannel_continue` to not broadcast the funding tx once the commitment is signed,, plus a `fundchannel_cancel`, plus a `broadcasttx` command. Then it is possible to implement multifundchannel as a plugin. #1936

ZmnSCPxj:
> Wanted to point out that we have no way of confirming that the amount someone says they're going to fund a channel for is actually the amount that they fund the channel for. We can correct/fix for this once the transaction is broadcast.

***We cannot fix anything once tx is briadcast***. The initial commitment tx signed commits to a specific value for the funding txout. If the external wallet pays a different amount to the same address, the initial commitment is useless.

The broadcast of the funding tx needs to be done after the other side signs initial commitment. So the external wallet must sign *but not broadcast*. If the funding tx is broadcast immediately then the counterparty can refuse to sign the initial commitment unless you `push_msat` at least half the value to them. Or consider what happens if the peer goes down permanently suddenly after `accept_channel`.

ZmnSCPxj:
Xref. https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-May/015925.html

The above link describes CoinSwap but note that the same funding transaction pattern underlies that protocol. It is unsafe to broadcast the funding tx until the backout tx is signed;if you want to to support wallets that do "sign and broadcast" as a single atomic step (e.g. every custodial wallet) you need `SIGHASH_ANYPREVOUT`. On LN the initial commitment serves as the backout (on LN unilateral close *is* our backout).

niftynei:
> More concretely, it seems a good amount of duplication of code, though I have not checked all details.

The duplication is by design, for a few reasons, namely that it's easier to DRY things out than to shoehorn too much divergence into a singular flow, especially at the work in progress stage. But agreed, it does need some consolidation to preserve some maintainability, especially between this and the ongoing dual funding work.

>> Wanted to point out that we have no way of confirming that the amount someone says they're going to fund a channel for is actually the amount that they fund the channel for. We can correct/fix for this once the transaction is broadcast.

>We cannot fix anything once funding tx is broadcast. The initial commitment tx signed commits to a specific value for the funding txout. If the external wallet pays a different amount to the same address, the initial commitment is useless.

Ah right, my mistake I forgot about the commitment transaction commitments to amounts, I was thinking more from an internal representation standpoint. You're right, that would be unfortunate. A user of this RPC should supply the correct funding amount then, to avoid any of these issues.  One way to mitigate this from an interaction standpoint would be to have the user supply a PSBT with the inputs, such that the `funding_amount` could be calculated internally by c-lightning, but that's really a convenience/sanity check rather than a deterrent.

ZmnSCPxj:
Fixes: #644 

xref. https://github.com/ElementsProject/lightning/issues/644#issuecomment-358818742

rustyrussell:
> This needs to be designed carefully. In particular the external wallet must be capable of providing a signed tx without broadcasting it (and clients need to be very aware that it is `lightningd` that must broadcast the funding tx).

Is that really necessary? I was thinking of a flow like:
1. Wallet tells lightningd how much it will fund.
2. Lightningd starts open channel, then tells wallet what address to send to.
3. Wallet tells lightningd the txid and output number.
4. Lightningd tells wallet to release tx.

I imagine a fund-a-friend feature on a mobile Lightning, where Bob tells his phone to open a channel for 0.01btc, it starts negotiation and presents a qr code with the address. Alice scans that, produces tx, presents another QR code with txid and outnum. Bob scans that, and if negotiation completed, Alice tells her wallet to release tx.

ZmnSCPxj:
Perhaps a reword to "(and clients need to be very aware that it is `lightningd` that must *initiate* broadcast of the funding tx)" more clearly expresses my meaning.

common/addr.c
```diff
@@ -0,0 +1,28 @@
...
+
+        /* Check that scriptPubkey is P2WSH or P2WPKH */
+	if (scriptPubkey[0] != 0x00)
+		return NULL;
+	if (scriptLen != 22 && scriptLen != 34)
+		return NULL;
+        if (scriptLen == 22 && scriptPubkey[1] != 0x14)
+		return NULL;
+        if (scriptLen == 34 && scriptPubkey[1] != 0x20)
+                return NULL;
```
rustyrussell: We already have these helpers in bitcoin/script.c:
```
if (!is_p2wsh(scriptPubkey, NULL) && !is_p2wpkh(scriptPubkey, NULL))
    return NULL;
```

common/addr.c
```diff
@@ -0,0 +1,28 @@
+#include "addr.h"
+#include <common/bech32.h>
+
+/* Returns NULL if the script is not a P2WPKH or P2WSH */
+char * encode_scriptpubkey_to_addr(const tal_t *ctx,
```
rustyrussell: Style nit: `char *encode_` with no space after the `*`

common/addr.c
```diff
@@ -0,0 +1,28 @@
...
+	if (scriptLen != 22 && scriptLen != 34)
+		return NULL;
+        if (scriptLen == 22 && scriptPubkey[1] != 0x14)
+		return NULL;
+        if (scriptLen == 34 && scriptPubkey[1] != 0x20)
+                return NULL;
+
+	out = tal_arr(ctx, char, 73 + strlen(hrp));
+	ok = segwit_addr_encode(out, hrp, 0, scriptPubkey + 2, scriptLen - 2);
+	if (!ok)
```
rustyrussell: Trivial: I think we can skip the 'ok' var here?

lightningd/opening_control.c
```diff
@@ -1052,13 +1052,20 @@ static struct command_result *json_fund_channel_continue(struct command *cmd,
...
 
 	if (!param(cmd, buffer, params,
 		   p_req("id", param_node_id, &id),
 		   p_req("txid", param_txid, &funding_txid),
+		   p_req("txout", param_number, &funding_txout_num),
 		   NULL))
 		return command_param_failed();
 
+	if (*funding_txout_num > UINT16_MAX)
+		return command_param_failed();
```
rustyrussell: command_param_failed is kinda magical: you need `command_fail()` here with a message.

ZmnSCPxj:
> I think funding_continue should be called funding_complete. Maybe?

In case of multifundchannel, it is not necessarily so that the fundchannel process is complete on completion of the second step --- if some other counterparty aborts on the second step, then we should not broadcast the funding transaction.  Instead we should support a fundchannel_abort that `error`s the channel and makes the counterparties that completed the second step forget the channel.

rustyrussell:
> > I think funding_continue should be called funding_complete. Maybe?
> 
> In case of multifundchannel, it is not necessarily so that the fundchannel process is complete on completion of the second step --- if some other counterparty aborts on the second step, then we should not broadcast the funding transaction. Instead we should support a fundchannel_abort that `error`s the channel and makes the counterparties that completed the second step forget the channel.
> 
> So I would argue that funding_continue is more appropriate name.

But if funding_continue returns successfully, the channel is funded.  It has to be, since we're relying on the external party to broadcast the funding tx.

So the `funding_complete` name works well.

rustyrussell:
Minor patch needed, to plug memleak:

```
diff --git a/openingd/openingd.c b/openingd/openingd.c
index 937598a0b..d89dd627f 100644
--- a/openingd/openingd.c
+++ b/openingd/openingd.c
@@ -616,7 +616,7 @@ static u8 *funder_channel_start(struct state *state,
 		return NULL;
 
 	funding_output_script =
-		scriptpubkey_p2wsh(state,
+		scriptpubkey_p2wsh(tmpctx,
 				   bitcoin_redeem_2of2(state,
 						       &state->our_funding_pubkey,
 						       &state->their_funding_pubkey));
```

ZmnSCPxj:
> But if funding_continue returns successfully, the channel is funded. It has to be, since we're relying on the external party to broadcast the funding tx.

That depends on when`fundchannel_continue` returns.

If it returns before `funding_locked` , then we can use it for `multifundchannel` and cancel it later if we decide to not  broadcast the fubding tx after all.

If it returns after `funding_locked` then it cannot be used for `multifundchannel`.

I strongly suggest splitting into smaller steps and just monitor the channel state moving to `CHANNELD` to determine if `funding_locked` occurred or not yet.

We can always create a plugin that combines multiple steps later, we cannot create a plugin to split up operations.

niftynei:
> If it returns before `funding_locked`, then we can use it for `multifundchannel` and cancel it later if we decide to not broadcast the funding tx after all.

Here's one way that `multifundchannel` might work, using the RPC commands that this PR 

```
#  Let's say a user wants to open 3 channels, one with 10 0000sats, 
# one with 5 0000sats, and one with 2 0000sats, with a single funding transaction 
# funded from an external wallet.

# The first step would be to get funding addresses for all the proposed channels.
$ lightning-cli fundchannel_start <node1_id> 100000
   {"funding_address":"bcrt1q6p9umlh85swksnnzymg3pm09cjhqyd8nd6m2ep"}
$ lightning-cli fundchannel_start <node2_id> 50000
      {"funding_address":"bcrt1qe0dmajvx920zwnkdqkwvke5hapsrvxf8hfaffd"}
$ lightning-cli fundchannel_start <node3_id> 20000
      {"funding_address":"bcrt1qkqp5dcgpm55k3tl6matr943whtmpmzy64mdanl"}

# User uses external wallet to compose desired transaction, and derives the
# funding txid for this transaction, then 'completes' the 3 outstanding channel
# opens.

$ lightning-cli fundchannel_complete <node1_id> <funding_txid> <funding_txout1>
$ lightning-cli fundchannel_complete <node2_id> <funding_txid> <funding_txout2>
$ lightning-cli fundchannel_complete <node3_id> <funding_txid> <funding_txout3>

# User broadcasts their transaction, all three channels complete.
```

Is this an accurate portrayal of a workable `multifundchannel` workflow?

ZmnSCPxj:
What happens if `funding_conplete` fails for `node2_id` because node 2 disconnected? Is node1 now stuck waiting for a funding tx that will now never be broadcast? (Until funding timeout but that is a long time)

That is why we need `funding_cancel` to do an `error`.

Xref this thread: https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-January/000905.html

In the end I concluded that `error` had the semantics needed. But it is still necessary.

rustyrussell:
I somewhat agree: it might be nice to be able to cancel even though lightningd thinks we're ready.  However, that can be added later and we're too close to release.

I also want to remove the fundchannel path from openingd, and always use the other path.  But that's easiest done by moving `fundchannel` to a simple plugin, then ripping it out.  This will have to wait for 0.7.2.

rustyrussell:
Still has a memleak, BTW...

ZmnSCPxj:
Maybe this? https://github.com/ElementsProject/lightning/pull/2672/commits/623385532d2fc4e38d7b5593fe5a37ddb3f2e74f#diff-1130d9d3f990f966967fbc527c10f275R1206 Previous value of `peer->uncommitted_channel->fc->cmd` is not freed?

Also, is not modifying `if (!peer->uncommitted_channel->fc || !peer->uncommitted_channel->fc->inflight)` to `if (!peer->uncommitted_channel->fc)` sufficient to make `fundchannel_cancel` work after `fundchannel_continue` is executed?

rustyrussell:
Fixed "%ls" which should have been "%u"...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2672/commits/dcab321aae208fb1f0b752a11dfa1d8ba5863d55

PR #2673, opened 2019-05-24
Fix build in distros with older GCCs

jb55:
Here are the latest line of patches that fixes the build on older GCCs. This is the case on NixOS which still hasn't migrated to a newer version yet.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -171,7 +171,7 @@ static void start_commit_timer(struct peer *peer);
 
 static void billboard_update(const struct peer *peer)
 {
-	const char *funding_status, *announce_status, *shutdown_status;
+	const char *funding_status = "", *announce_status = "", *shutdown_status = "";
 
 	if (peer->funding_locked[LOCAL] && peer->funding_locked[REMOTE])
 		funding_status = "Funding transaction locked.";
--- a/channeld/test/run-full_channel.c
+++ b/channeld/test/run-full_channel.c
@@ -120,7 +120,7 @@ static const struct htlc **include_htlcs(struct channel *channel, enum side side
 		struct preimage preimage;
 		struct sha256 hash;
 		enum channel_add_err e;
-		enum side sender;
+		enum side sender = 0;
 		struct amount_msat msatoshi = AMOUNT_MSAT(0);
 
 		switch (i) {
--- a/devtools/create-gossipstore.c
+++ b/devtools/create-gossipstore.c
@@ -69,7 +69,7 @@ int main(int argc, char *argv[])
 	struct amount_sat sat;
 	bool verbose = false;
 	char *infile = NULL, *outfile = NULL, *csvfile = NULL, *csat = NULL;
-	int infd, outfd, scidi = 0, channels = 0, nodes = 0, updates = 0;
+	int infd = 0, outfd, scidi = 0, channels = 0, nodes = 0, updates = 0;
 	struct scidsat *scidsats = NULL;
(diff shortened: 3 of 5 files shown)


jb55:
should we turn of maybe uninitialized warnings for older gcc versions? One downside of these patches is that they mask potential future *real* uninitialized errors during refactorings.

devtools/create-gossipstore.c
```diff
@@ -69,7 +69,7 @@ int main(int argc, char *argv[])
 	struct amount_sat sat;
 	bool verbose = false;
 	char *infile = NULL, *outfile = NULL, *csvfile = NULL, *csat = NULL;
-	int infd, outfd, scidi = 0, channels = 0, nodes = 0, updates = 0;
+	int infd = 0, outfd, scidi = 0, channels = 0, nodes = 0, updates = 0;
```
ZmnSCPxj: Is the intent that if this program is not given `--input`, it acts as a filter program?

If so, we should probably use `infd = STDIN_FILENO` instead.

Or better yet change this:

    if (infile) {
        infd = open(infile, O_RDONLY);
        if (infd < 0)
            err(1, "opening %s", infile);
    } else
        infd = STDIN_FILENO;


For symmetry with `outfd`.

channeld/test/run-full_channel.c
```diff
@@ -120,7 +120,7 @@ static const struct htlc **include_htlcs(struct channel *channel, enum side side
 		struct preimage preimage;
 		struct sha256 hash;
 		enum channel_add_err e;
-		enum side sender;
+		enum side sender = 0;
```
ZmnSCPxj: We should use the proper `enum side` label.

    /* Invalid value that will be written in the `switch` later -- only used to shut up the lint.  */
    enum side sender = NUM_SIDES;

The later `switch` statement properly initializes `sender`, this is just an egregious mistake by the compiler.

gossipd/routing.c
```diff
@@ -988,7 +988,7 @@ find_shorter_route(const tal_t *ctx, struct routing_state *rstate,
 		   struct amount_msat *fee)
 {
 	struct unvisited *unvisited;
-	struct chan **short_route;
+	struct chan **short_route = NULL;
```
ZmnSCPxj: Compiler is right in this case --- We could `goto bad_total` which goes through the `tal_free(short_route)` before the `assert(short_route)`.

channeld/channeld.c
```diff
@@ -171,7 +171,7 @@ static void start_commit_timer(struct peer *peer);
 
 static void billboard_update(const struct peer *peer)
 {
-	const char *funding_status, *announce_status, *shutdown_status;
+	const char *funding_status = "", *announce_status = "", *shutdown_status = "";
```
ZmnSCPxj: Compiler is wrong here but I suppose it cannot figure out that we have gotten all possible valid states, this is fine.

lightningd/jsonrpc.c
```diff
@@ -977,7 +977,7 @@ json_tok_address_scriptpubkey(const tal_t *cxt,
 	const char *bip173;
 
 	bool parsed;
-	bool right_network;
+	bool right_network = false;
```
ZmnSCPxj: Compiler is wrong here, probably needs a comment explaining that we initialize this solely because the lint cannot figure it out.

jb55:
I'm going to fix this for good in a new PR soon, the idea is to turn on `-Wno-error=maybe-uninitialized` for gcc 7

PR #2674, opened 2019-05-24
build: set -Wno-error=maybe-uninitialized for gcc 7

jb55:
Instead of silencing every individual maybe-uninitialized false positive, let's disable that warning error for older gcc versions.

Also revert all of my old build fixup commits. They shouldn't be needed anymore, and they potentially mask real uninitialized errors that could be caught in newer gcc versions.

### DIFF
--- a/configure
+++ b/configure
@@ -14,6 +14,19 @@ STATIC=${STATIC:-0}
 CONFIGURATOR_CC=${CONFIGURATOR_CC:-$CC}
 ASAN=${ASAN:-0}
 
+[ "$CC" = "gcc" ] && IS_GCC=1 || IS_GCC=0
+
+# older version of gcc give warnings which break the build. Let's create
+# a few helper flags that our build can use to fix these problems.
+if [ $IS_GCC -eq 1 ]; then
+    GCC_VERSION=$(gcc --version | grep ^gcc | sed 's/^.* //g')
+    GCC_MAJOR=$(printf "%s\n" "$GCC_VERSION" | sed -E -n 's/([[:digit:]]+).*/\1/p')
+    if [ $GCC_MAJOR -eq 7 ]; then
+        printf "older gcc detected, setting UNRELIABLE_MAYBE_UNINIT_WARNINGS=1\n"
+        UNRELIABLE_MAYBE_UNINIT_WARNINGS=1
+    fi;
+fi
+
 CONFIGURATOR=ccan/tools/configurator/configurator
 CONFIG_VAR_FILE=config.vars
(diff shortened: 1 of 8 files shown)


jb55:
@ZmnSCPxj so somehow an old gcc is catching valid uninitialized errors and the newer ones are missing them? that's a yikes if true.

jb55:
@ZmnSCPxj are you sure about 0ff3b6da23742e70790546004a7c86d2925dd1a0? It looks like it always gets assigned by build_route?

ZmnSCPxj:
Possibly.  Such things can only judge heuristically in many cases, unless you are doing some serious static analysis that only advanced AI can perform (disclaimer: I am not an advanced AI).

e59389f : if `infile` not specified, `infd` is not initialized.  It is probably better to emulate the `outfd` code except with `STDIN_FILENO`.

0ff3b6d : if the first `goto bad_total` executes, then that path has uninitialized `short_route` but `bad_total` passes through to `out` whose first call is `tal_free(short_route)`.

jb55:
@ZmnSCPxj oh I missed that goto. good call, I'll re-add those ones

ZmnSCPxj:
>@ZmnSCPxj are you sure about 0ff3b6d? It looks like it always gets assigned by build_route?

It is assigned at line 1022, but notice the `goto bad_total` at line 999.

jb55:
@ZmnSCPxj I've pushed those commits with your suggestions

rustyrussell:
Crap, Travis was supposed to be testing this, but it's not :(

There are some fixes needed for gcc-8, too  And we should use `-Wno-error=maybe-uninitialized` instead of removing -Werror.  Unfortunately, clang has '-Wno-error=uninitialized' instead, which seems to be accepted by gcc but does something different.

And your test is also in the wrong place in configure, as CC is still the default; better to test via configurator.  Let me do some prep work.

rustyrussell:
I ended up going down the rabbit hole a bit on this, sorry. I already had one patch to rework configure and I've applied some of these on top.

In particular, I want GCC 7 and above to build with -Werror with -O3 or no optimization, which is where gcc understands most of the code paths. I will pull your work into that PR when I get a break from family duties to submit it!

jb55:
@rustyrussell ok just to be sure we're on the same page, I haven't touched or removed -Werror for GCC7, or touched the existing build logic, I just added `-Wno-error=maybe-uninitialized` for gcc 7 specifically.

jb55:
after testing again I realize `-Werror` is only set when `--enable-developer` is on, so this PR only affects dev mode at the moment.

PR #2675, opened 2019-05-24
cdecker's weekly cleanups

cdecker:
Things that don't really fit in anywhere else:

 - Avoid trying to restart the `bitcoind` proxy along with `lightingd`, removes the `unable to bind` warning
 - A small fix to wait for a connection to be recognized before trying to use it
 - Marking some flaky tests as `@flaky` (will do a stabilization pass later, but they were really annoying and were addressed only by restarting instead of fixing them).
 - Remove the `UPDATE <message_type>` spam from `subd` which is really annoying in production.

### DIFF
--- a/lightningd/subd.c
+++ b/lightningd/subd.c
@@ -469,8 +469,6 @@ static struct io_plan *sd_msg_read(struct io_conn *conn, struct subd *sd)
 		}
 	}
 
-	log_debug(sd->log, "UPDATE %s", sd->msgname(type));
-
 	/* Might free sd (if returns negative); save/restore sd->conn */
 	sd->conn = NULL;
 	tal_add_destructor2(sd, mark_freed, &freed);
--- a/tests/test_closing.py
+++ b/tests/test_closing.py
@@ -308,7 +308,7 @@ def test_penalty_inhtlc(node_factory, bitcoind, executor):
     l2.daemon.wait_for_log('=WIRE_COMMITMENT_SIGNED-nocommit')
 
     # Make sure l1 got l2's commitment to the HTLC, and sent to master.
-    l1.daemon.wait_for_log('UPDATE WIRE_CHANNEL_GOT_COMMITSIG')
+    l1.daemon.wait_for_log('got commitsig')
 
     # Take our snapshot.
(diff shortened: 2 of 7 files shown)


cdecker:
Just found another simple to fix test, so I just added it:

 - upfront shutdown script was reconnecting manually but disallowing reconnections...

lightningd/subd.c
```diff
@@ -469,8 +469,6 @@ static struct io_plan *sd_msg_read(struct io_conn *conn, struct subd *sd)
 		}
 	}
 
-	log_debug(sd->log, "UPDATE %s", sd->msgname(type));
```
niftynei: would it be more appropriate to move this to log at the IO level?

cdecker: The IO level logs only raw messages (hex encoded of course) so far afaik

rustyrussell: It *can* log messages, but we logged this in lieu of full IO.  We should probably log internal IO messages, but that's a separate change.

rustyrussell:
> Just found another simple to fix test, so I just added it:
> 
>     * upfront shutdown script was reconnecting manually but disallowing reconnections...

Can't reproduce this here.  may_reconnect=False does two things: disable automatic reconnect, and check that we don't get reconnections manually.

A "reconnection" is only when we have a live channel, and this test does check l1 is onchain. Easiest fix is to change that to check l2 is onchain as well.

lightningd/subd.c
trueptolemy: Does the _internal IO messages_ mean the message between process?

cdecker:
> > Just found another simple to fix test, so I just added it:
> > ```
> > * upfront shutdown script was reconnecting manually but disallowing reconnections...
> > ```
> 
> Can't reproduce this here. may_reconnect=False does two things: disable automatic reconnect, and check that we don't get reconnections manually.
> 
> A "reconnection" is only when we have a live channel, and this test does check l1 is onchain. Easiest fix is to change that to check l2 is onchain as well.

Decided to drop this commit, in order not to hold up. Will resubmit a better, less driveby, fix in the next round :-)

PR #2678, opened 2019-05-25
Plugin: accept different option types (str, int, bool)

darosior:
This PR adds the possibility to register plugin options of type `int`, `bool` in addition to `string`.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -81,11 +81,22 @@ struct plugins {
 	struct lightningd *ld;
 };
 
+/* The value of a plugin option, which can have different types.
+ * The presence of the integer and boolean values will depend of
+ * the option type, but the string value will always be filled.
+ */
+struct plugin_opt_value {
+	char *as_str;
+	int *as_int;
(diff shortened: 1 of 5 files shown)


lightningd/plugin.c
```diff
@@ -486,31 +502,51 @@ static bool plugin_opt_add(struct plugin *plugin, const char *buffer,
...
-	if (!json_tok_streq(buffer, typetok, "string")) {
-		plugin_kill(plugin,
-			    "Only \"string\" options currently supported");
-		return false;
-	}
-
 	popt = tal(plugin, struct plugin_opt);
+	popt->value = talz(popt, struct plugin_opt_value);
+	popt->value->as_str = talz(popt->value, char);
+	popt->value->as_int = talz(popt->value, int);
```
ZmnSCPxj: Possibly better to leave `as_int` and `as_bool` as `NULL` if they are not of the appropriate type, and modify the earlier code that assigns them to only allocate at those points.

darosior:
This is not a WIP anymore.

PR #2680, opened 2019-05-26
Fix warnings with different gcc versions, -O3.

rustyrussell:
I thought Travis was testing -O3, but SOURCE_CHECK_ONLY wasn't actually building the source!

A while ago I removed -Werror when we have optimization but not -O3, as we were getting spurious "may be uninitialized" warnings.  @jb55 pointed out we should only suppress those errors, not all errors, so this test is now more sophisticated: suppress if gcc < 7, or we're using optimization other than -O3.

Meanwhile, I tested with gcc 7  and gcc 9 as well as my normal gcc 8, and found both real and spurious errors, which I fixed.


ZmnSCPxj:
ACK 6886b31

ZmnSCPxj:
Hmm

    [0;1;31merror: [0munknown warning option '-Werror=maybe-uninitialized'; did you mean
          '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option][0m
    make: *** [ccan/ccan/cdump/tools/cdump-enumstr.o] Error 1
    [0;1;31merror: [0munknown warning option '-Werror=maybe-uninitialized'; did you mean
          '-Werror=uninitialized'? [-Werror,-Wunknown-warning-option][0m
    make: *** [ccan-cdump.o] Error 1
    make: *** wait: No child processes.  Stop.

rustyrussell:
Hmm, it should not have set that flag for clang. Will check here...

jb55:
looks ok, from what I can tell this still has maybe-uninitialized errors on gcc7? As long as those are getting caught and fixed now that works for me :+1: :)

rustyrussell:
Those are all still need for gcc 8 and -O3. We should create a gcc_warn_init macro to clearly mark them though

connectd/tor.c
```diff
@@ -59,11 +59,12 @@ static struct io_plan *connect_finish(struct io_conn *conn,
...
-				       (connect->buffer + SIZE_OF_RESPONSE -
-					SIZE_OF_IPV4_RESPONSE),
+				       connect->buffer + SIZE_OF_RESPONSE +
+				       SIZE_OF_IPV4_RESPONSE,
 				       SIZE_OF_IPV6_RESPONSE -
-				       SIZE_OF_RESPONSE - SIZE_OF_IPV4_RESPONSE,
+				       SIZE_OF_IPV4_RESPONSE,
 				       &connect_finish2, connect);
 
 		} else if ( connect->buffer[3] == SOCKS_TYP_IPV4) {
```
Saibato: tACK  09caab5eef76b6599c262f565b15ece6b07378d

It's not easy to test since even if you call on ipv6 the Tor circuit will almost always return an empty IP4 return code.
Since we call on DOMAIN and never use the empty ipv4/6 return anyway.

As ipv6 often leaks mac, ipv6 and Tor are not the best match, but anyway   
Luckily the buffer[-2] went in the void

Good Catch!  :+1::

configure
```diff
@@ -32,18 +32,42 @@ usage_with_default()
     echo "  $1 (default $DEF)"
 }
 
+# Given DEVELOPER, what COPTFLAGS do we default to.
+default_coptflags()
+{
+    if [ "$1" = 0 ]; then
+	echo "-Og"
+    else
+	echo
```
niftynei: is this empty `echo` necessary?

configure
```diff
@@ -42,16 +42,19 @@ default_coptflags()
...
 {
-    BASE="-Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition"
-    # We can use -Werror if no optimization, or if -O3.
-    if [ -z "${1##*-O3*}" ]; then
-	echo "$BASE -Werror"
-    else
-	echo "$BASE"
+    F=$BASE_WARNFLAGS
+    # Clang doesn't like -Wno-error=maybe-uninitialized, but doesn't seem
+    # tp give spurious warnings, either.
```
niftynei: `tp` -> `to`

rustyrussell: Ph np, damn typps!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2680/commits/9476d6baadfb4b3f1302e853b9bc62ac696eb55a

(Assuming that @niftynei is OK with the two trivial fixes).

PR #2685, opened 2019-05-28
Gossip store fixes

rustyrussell:
As part of my final gossip optimizations PR, I tightened and added testing.  This revealed existing bugs: in particular, the "bad node_announcement" message we sometimes get loading store.

Fixing this revealed another nasty issue, where we tried to remove from the gossip_store as it was being loaded.  This is worked around for now: it will be fixed properly by rebasing #2644 on top of this....

### DIFF
--- a/gossipd/gossip_store.c
+++ b/gossipd/gossip_store.c
@@ -644,3 +644,9 @@ out:
 		     gs->len);
 	gs->writable = true;
 }
+
+/* FIXME: Remove */
+bool gossip_store_loading(const struct gossip_store *gs)
+{
+	return !gs->writable;
+}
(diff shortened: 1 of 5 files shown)


darosior:
Tested it on my testnet node, works well (few Python nits on Travis but tests should pass, they passed on my side)

rustyrussell:
Trivial rebase to remove whitespace to make flake8 happy.

SimonVrouwe:
Tested on a few bad `gossip_store` files that would truncate/crash before this PR. Now none of them truncate anymore and all `dev-compact-gossip-store` succeed. Nice!

PR #2689, opened 2019-05-29
Variable hop payload size for sphinx

cdecker:
This is the competing proposal to #2363. It implements the variable size `hop_payloads` discussed duriing the last IRC spec meeting.

It is implemented as fixups based on that PR, so you should be able review just the `fixup!` commits and get a good picture of the differences. I also cleaned up all the common mentions of frames in the final commit, but left it in other places to facilitate reviews.

Closes #2363


common/sphinx.c
```diff
@@ -135,10 +137,14 @@ void sphinx_add_v0_hop(struct sphinx_path *path, const struct pubkey *pubkey,
 		       const struct short_channel_id *scid,
 		       struct amount_msat forward, u32 outgoing_cltv)
 {
+	u8 padding[] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+			0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
 	u8 *buf = tal_arr(path, u8, 0);
 	towire_short_channel_id(&buf, scid);
 	towire_u64(&buf, forward.millisatoshis); /* Raw: low-level serializer */
 	towire_u32(&buf, outgoing_cltv);
+	towire(&buf, padding, ARRAY_SIZE(padding));
```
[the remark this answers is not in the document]
rustyrussell: ARRAY_SIZE is basically always preferable when you have an array.  Unlike sizeof(), it will break compile if the argument is a pointer instead of an array, due to refactoring.

common/sphinx.c
```diff
@@ -563,9 +590,17 @@ struct route_step *process_onionpacket(
...
-	/* Extract how many frames we need to shift away */
-	shift_size = ((paddedheader[0] >> 4) + 1) * FRAME_SIZE;
+	/* Extract how many bytes we need to shift away */
+	if (paddedheader[0] == 0x00) {
+		shift_size = FRAME_SIZE;
+	} else {
+		/* In addition to the raw payload we need to also shift the
+		 * length encoding itself and the HMAC away. */
+		vsize = varint_get(paddedheader, 3, &shift_size);
+		shift_size += vsize + HMAC_SIZE;
```
rustyrussell: This is wrong on several levels.

1. I worry what happens if we don't have 3 bytes here, can we be tricked into reading off the end?
2. This uses bitcoin's varint and we are now using BigSize which is big-endian.
3. shift_size needs to be sanity checked carefully!

lightningd/peer_htlcs.c
```diff
@@ -747,11 +747,32 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
...
+			  type_to_string(tmpctx, struct short_channel_id,
+					 &gr->next_channel));
+		subd_req(hin, ld->gossip, req, -1, 0,
+			 channel_resolve_reply, gr);
+	} else
+		handle_localpay(hin, hin->cltv_expiry, &hin->payment_hash,
+				hop_data->amt_forward,
+				hop_data->outgoing_cltv);
+/*		if (rs->nextcase == ONION_FORWARD) {
 			struct gossip_resolve *gr =
```
rustyrussell: Erk, please delete don't comment out.

devtools/onion.c
```diff
@@ -1,10 +1,16 @@
+#include <ccan/mem/mem.h>
```
rustyrussell: Erk, I just repurposed onion.c as a generic onion builder, which will clash horribly with this.

My approach with the TLV tests (which are also JSON) was to quote the spec in the test case (so if it's wrong or changes, check-source points it out) then pulls the JSON out of itself.  We should generalize that, IMHO.

common/sphinx.h
```diff
@@ -93,20 +86,19 @@ struct hop_data {
 
 enum sphinx_payload_type {
 	SPHINX_V0_PAYLOAD = 0,
+	SPHINX_TLV_PAYLOAD = 1,
 	SPHINX_INVALID_PAYLOAD = 254,
 	SPHINX_RAW_PAYLOAD = 255,
```
rustyrussell: What's SPHINX_RAW_PAYLOAD used for?

common/sphinx.c
```diff
@@ -435,10 +439,20 @@ static void deserialize_hop_data(struct hop_data *data, const u8 *src)
...
+
+	/* Backwards compatibility for the legacy hop_data format. */
+	if (hop->realm == 0x00)
+		dest[pos++] = 0x00;
+	else
+		pos += varint_put(dest+pos, raw_size);
+
+	memcpy(dest + pos, hop->payload, raw_size);
+	pos += raw_size;
+	memcpy(dest + hop_size - HMAC_SIZE, hop->hmac, HMAC_SIZE);
```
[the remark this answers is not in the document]
cdecker: Made more explicit:

```c
	memcpy(dest + pos, hop->payload, raw_size);
	pos += raw_size;

	padding_size = hop_size - pos - HMAC_SIZE;
	memset(dest + pos, 0, padding_size);
	pos += padding_size;

	memcpy(dest + pos, hop->hmac, HMAC_SIZE);
```

:wink:

devtools/onion.c
cdecker: Yep, ran into quite a bit of trouble rebasing, but it's working fine now :-)

lightningd/peer_htlcs.c
cdecker: My bad, must've slipped in

common/sphinx.c
cdecker: > 1. I worry what happens if we don't have 3 bytes here, can we be tricked into reading off the end?
No, we have `paddedheader[2*ROUTING_INFO_SIZE];`, i.e., 2600 bytes here.

> 2. This uses bitcoin's varint and we are now using BigSize which is big-endian.
Yep, fixing up

> 3. shift_size needs to be sanity checked carefully!
I went and just abort when `shift_size >= ROUTING_INFO_SIZE` since that'd shift out all non-filler infos and leave us with garbage at the end anyway :-)

cdecker:
Rebased and fixed up to reflect feedback :wink:

jb55:
fixup ACK on my comments!

common/sphinx.c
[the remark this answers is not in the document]
rustyrussell: I still like to see an assert(pos == hop_size); at the end?

bitcoin/varint.h
```diff
@@ -14,4 +14,13 @@ size_t varint_put(u8 buf[VARINT_MAX_LEN], varint_t v);
 
 /* Returns bytes used: 0 if max_len too small. */
 size_t varint_get(const u8 *p, size_t max_len, varint_t *val);
+
+
+/* Big-endian variant of varint_put, used in lightning */
+size_t bigsize_put(u8 buf[VARINT_MAX_LEN], varint_t v);
+
+/* Big-endian variant of varint_get, used in lightning */
+size_t bigsize_get(const u8 *p, size_t max, varint_t *val);
```
rustyrussell: These routines are in the wrong place, and they're redundant.

You want `fromwire_bigsize()` and `towire_bigsize()`.

common/sphinx.c
```diff
@@ -435,11 +435,17 @@ static void deserialize_hop_data(struct hop_data *data, const u8 *src)
...
 	size_t raw_size = tal_bytelen(hop->payload);
 	size_t hop_size = sphinx_hop_size(hop);
 	int pos = 0;
+
+#if !EXPERIMENTAL_FEATURES
+	if (hop->realm != 0x00)
+		return false;
+#endif
+
 	memset(dest, 0, hop_size);
```
rustyrussell: Still need to delete this...

devtools/onion.c
```diff
@@ -12,32 +13,55 @@
...
 
 	for (int i = 0; i < num_hops; i++) {
 		size_t klen = strcspn(argv[1 + i], "/");
-		if (!hex_decode(argv[1 + i], klen, privkeys[i], 32))
-			errx(1, "Invalid private key hex '%s'", argv[1 + i]);
-
-		if (secp256k1_ec_pubkey_create(secp256k1_ctx, &path[i].pubkey,
-					       privkeys[i]) != 1)
-			errx(1, "Could not decode pubkey");
+		assert(klen == 2 * PUBKEY_LEN || klen == 2 * PRIVKEY_LEN);
```
rustyrussell: Remove this assert, since you give a proper message below?

devtools/onion.c
```diff
@@ -12,32 +13,55 @@
...
+		if (klen == 2 * PRIVKEY_LEN) {
+			if (!hex_decode(argv[1 + i], klen, rawprivkey, PRIVKEY_LEN))
```
rustyrussell: I would combine these two, for simplicity.  hex_decode() will fail if the wrong length, and it avoids magic `2 *`.  The other option would be `if (hex_data_size(klen) == PRIVKEY_LEN)` but it still seems redundant?

devtools/onion.c
```diff
@@ -12,32 +13,55 @@
...
+
+			if (secp256k1_ec_pubkey_parse(secp256k1_ctx,
+						      &path[i].pubkey,
+						      rawpubkey, PUBKEY_LEN) != 1)
+				errx(1, "Could not decode pubkey");
+		} else {
+			fprintf(stderr,
+				"Provided key is neither a pubkey nor a "
+				"privkey: %s\n",
+				argv[1 + i]);
```
rustyrussell: errx()

devtools/onion.c
```diff
@@ -12,32 +13,55 @@
 #include <unistd.h>
 
 #define ASSOC_DATA_SIZE 32
+#define PUBKEY_LEN 33
+#define PRIVKEY_LEN 32
```
rustyrussell: `PUBKEY_CMPR_LEN` from bitcoin/pubkey.h?  Hmm we don't have a convenient PRIVKEY_LEN though...

If you're bored you could add one to bitcoin/privkey.h :)

cdecker:
Clean rebase to get code back on top of `master`, now doing a cleanup sweep to address @rustyrussell's feedback.

bitcoin/varint.h
cdecker: Agreed that they don't belong here, but they are useful since `fromwire_*` /  `towire_*` require `tal_arr` which they can extend, whereas the sphinx operations require in-place operations. I tried using the `fromwire` and `towire` variants, but it turned out to involve way more work than just implementing it again.

I can move the functions in a later cleanup PR.

common/sphinx.c
cdecker: It get's remove in a later commit :-)

devtools/onion.c
cdecker: Switched to use `hex_data_size` and `pubkey_from_hexstr` to do the conversion in one go. Would a counterpart for privkeys be desirable as well (`privkey_from_hexstr`)?

cdecker:
@rustyrussell I added the last commit to address your feedback. I think the last open question is the use of `bigsize_put` and `bigsize_get` which need to move into `common/` and whether we want to keep them at all, or whether we prefer going the `tal_arr` + `towire_bigsize` + `memcpy` + `tal_free` route. Alternatively we could have `towire_bigsize` and `fromwire_bigsize` use them internally, but I'm not sure how much logic is then left to the wrapper.

devtools/onion.c
rustyrussell: Hmm, hex_decode *does* this test internally, so it's redundant.  But not harmful, and I really want this merged :)

Issue #2687, opened 2019-05-28
lightningd hanging, not opening RPC port, not syncing, reducing latest block by 15


svewa:
I have c-lightning running for over a year now, but some time maybe two month ago I noticed it stopped and would not properly start again. I tried updating to 0.7.0 and now to latest git, does not help.

config file:
```
log-level=io
network=bitcoin
```

starting lightningd:
```
ln@c-lightning:~$ lightning/lightningd/lightningd 
2019-05-28T22:08:01.236Z plugin-autoclean[OUT] 7b226a736f6e727063223a22322e30222c226964223a312c226d6574686f64223a226765746d616e6966657374222c22706172616d73223a7b7d7d0a0a
2019-05-28T22:08:01.236Z plugin-pay[OUT] 7b226a736f6e727063223a22322e30222c226964223a302c226d6574686f64223a226765746d616e6966657374222c22706172616d73223a7b7d7d0a0a
2019-05-28T22:08:01.297Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_channeld
2019-05-28T22:08:01.301Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_closingd
2019-05-28T22:08:01.305Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_connectd
2019-05-28T22:08:01.311Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_gossipd
2019-05-28T22:08:01.315Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_hsmd
2019-05-28T22:08:01.319Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_onchaind
2019-05-28T22:08:01.323Z lightningd(3740): testing /home/ln/lightning/lightningd/lightning_openingd
2019-05-28T22:08:01.475Z lightning_hsmd(3750): pid 3750, msgfd 17
2019-05-28T22:08:01.508Z lightning_connectd(3751): pid 3751, msgfd 20
2019-05-28T22:08:01.509Z lightning_hsmd(3750): Client: Received message 11 from client
2019-05-28T22:08:01.509Z lightning_hsmd(3750): Client: Received message 9 from client
2019-05-28T22:08:01.509Z lightning_hsmd(3750): new_client: 0
2019-05-28T22:08:01.725Z lightning_connectd(3751): Created IPv6 listener on port 9735
[...]

PR #2690, opened 2019-05-30
lightningd/io_loop_with_timers.c: Move mainloop to its own source file, have chaintopology use it

ZmnSCPxj:
Fixes: #2687

### DIFF
--- a/lightningd/Makefile
+++ b/lightningd/Makefile
@@ -69,6 +69,7 @@ LIGHTNINGD_SRC :=				\
 	lightningd/hsm_control.c		\
 	lightningd/htlc_end.c			\
 	lightningd/invoice.c			\
+	lightningd/io_loop_with_timers.c	\
 	lightningd/json.c			\
 	lightningd/json_stream.c		\
 	lightningd/jsonrpc.c			\
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -22,6 +22,7 @@
 #include <inttypes.h>
 #include <lightningd/channel_control.h>
 #include <lightningd/gossip_control.h>
+#include <lightningd/io_loop_with_timers.h>
 
 /* Mutual recursion via timer. */
 static void try_extend_tip(struct chain_topology *topo);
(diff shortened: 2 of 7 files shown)


ZmnSCPxj:
Alternate solution for #2687 is to just fudge the fee estimation at the start.

ZmnSCPxj:
Missing a `#include"../io_loop_with_timers.c"` in `lightningd/test/run-find_my_abspath.c`, will fix after several hours.

ZmnSCPxj:
For that matter, why does `struct plugins` have its own `timers`?  Why not just unify it to a single `struct timers`?

cdecker:
> For that matter, why does `struct plugins` have its own `timers`? Why not just unify it to a single `struct timers`?

I did that to better encapsulate the plugins and make sure we have full control over the timers we create (in the same allocation hierarchy).

cdecker:
Am I correct in assuming the issue is that we don't process timers when starting the topology, therefore a failed fee estimate or a failed first block fetch will result in us hanging indefinitely?

ZmnSCPxj:
> Am I correct in assuming the issue is that we don't process timers when starting the topology, therefore a failed fee estimate or a failed first block fetch will result in us hanging indefinitely?

Yes, I believe so.

cdecker:
I took the liberty of add a test that reproduces #2687 first, this way we make sure the fix works and we don't get a regression later :-)

cdecker:
I'll ack @ZmnSCPxj's part and let someone else ack mine :-)

ACK https://github.com/ElementsProject/lightning/pull/2690/commits/cb0b9f7c45760e87099368a4c1928071e44f4774

m-schmoock:
ACK 592db38

Your test 592db38 without the fix properly xfails and demonstates the issue after a timeout of 70 seconds or so. It is also properly fixed by cb0b9f7 .

m-schmoock:
ACK 592db38

Your test 592db38 without the fix properly xfails and demonstates the issue after a timeout of 70 seconds or so. It is also properly fixed by cb0b9f7 . I have some unrelated flakes sometimes okay sometimes not when running locally, but behaves same on master with or without this fix:

 - tests/test_connection.py::test_no_fee_estimate (flaky)
 - tests/test_plugin.py::test_openchannel_hook  (always timeout wait_for ONCHAIN)
 - tests/test_misc.py::test_funding_reorg_private (always timeout wait_for_log "Deleting channel")

PR #2691, opened 2019-05-30
Pass htlc amount exceeded to exception

m-schmoock:
This will  pass HTLC fee amounts into the "Capacity exception" so they can be catched and parsed by a upper layer plugin (i.e. https://github.com/lightningd/plugins/pull/22 ). This change is useful, because the required HTLC fees cannot be computed/checked in advance.

## Rationale

currently, if a payment is made, the required HTLC onchain fees will be substracted from the `spendable` amount. This is a temporary hedge against the channel balance, if anything goes wrong. This is done in an opaque way, so the upper layers (like plugins) can't know the correct value that needs to be subtracted from the `spendable` amount, in order for the node to make the payment. For example the `drain` plugin has a ugly `while`   `try/except` loop to increase gradually until the `Capacity exceeded` exception is not thrown any more.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -644,7 +644,7 @@ static void handle_peer_add_htlc(struct peer *peer, const u8 *msg)
 
 	add_err = channel_add_htlc(peer->channel, REMOTE, id, amount,
 				   cltv_expiry, &payment_hash,
-				   onion_routing_packet, &htlc);
+				   onion_routing_packet, &htlc, NULL);
 	if (add_err != CHANNEL_ERR_ADD_OK)
 		peer_failed(peer->pps,
 			    &peer->channel_id,
@@ -2532,6 +2532,7 @@ static void handle_offer_htlc(struct peer *peer, const u8 *inmsg)
 	enum onion_type failcode;
 	/* Subtle: must be tal object since we marshal using tal_bytelen() */
 	const char *failmsg;
+	struct amount_sat htlc_fee;
 
 	if (!peer->funding_locked[LOCAL] || !peer->funding_locked[REMOTE])
 		status_failed(STATUS_FAIL_MASTER_IO,
@@ -2544,7 +2545,7 @@ static void handle_offer_htlc(struct peer *peer, const u8 *inmsg)
 
 	e = channel_add_htlc(peer->channel, LOCAL, peer->htlc_id,
 			     amount, cltv_expiry, &payment_hash,
-			     onion_routing_packet, NULL);
+			     onion_routing_packet, NULL, &htlc_fee);
 	status_trace("Adding HTLC %"PRIu64" amount=%s cltv=%u gave %s",
 		     peer->htlc_id,
 		     type_to_string(tmpctx, struct amount_msat, &amount),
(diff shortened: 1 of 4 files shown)


channeld/channeld.c
```diff
@@ -2576,7 +2577,7 @@ static void handle_offer_htlc(struct peer *peer, const u8 *inmsg)
 		goto failed;
 	case CHANNEL_ERR_CHANNEL_CAPACITY_EXCEEDED:
 		failcode = WIRE_TEMPORARY_CHANNEL_FAILURE;
-		failmsg = tal_fmt(inmsg, "Capacity exceeded");
+		failmsg = tal_fmt(inmsg, "Capacity exceeded by %s", fmt_amount_msat(inmsg, &exceed));
```
ZmnSCPxj: This will only be propagated to a local `sendpay` failure, incidentally.  If the failing HTLC is from a forward, we specifically do not propagate `failmsg` back to the payer, as we do not want to leak information to the payer.

ZmnSCPxj: Xref.

https://github.com/ElementsProject/lightning/blob/11412435ce0cae92205178db9a9f4b64d0d78830/lightningd/peer_htlcs.c#L360-L389

Notice how `failurestr` is only used in the path where `hout->am_origin` (i.e."I am the origin" of this outgoing HTLC, i.e. a local `pay`/`sendpay` attempt).  In the path where there is an incoming HTLC attached to the outgoing HTLC (i.e. a forward) `failurestr` is specifically not used.

m-schmoock: @ZmnSCPxj 
Having the exception just locally actually the intention behind.

The current channel amounts are, from a p2p protocol perspective, not know by intention.
However, hiding the exceeded amounts make it hard for local stuff (like plugins) to use a channel in its full volume (See ugly loop to determine correct HTLC fee in `drain` https://github.com/lightningd/plugins/pull/22/files#diff-04c04e1a674a88f0db43dd254913a601R332 )

If we expose the exceeded amounts (that depend on dynamic stuff like onchain fees) such plugin do only need to catch once to have the correct value and not loop in order to make an approximation.

ZmnSCPxj:
>This change is useful, because the required HTLC fees cannot be computed/checked in advance.

Do you mean the onchain fees? But the onchain fees are the responsibility of the forwarding node.  Do you mean the Lightning forwarding fees? But the fees are computed based on route (`pay` simply rejects routes that exceeds the user-imposed `maxfeepercent` and `exemptfee`)

m-schmoock:
@ZmnSCPxj 
> Do you mean the onchain fees? 
Yes and no. Its correct the LN fees can be calculated preciesly. But, the initiating node of a payment checks if the HTLC fees can be covered and dynamically and intransparently hides this from the user. The HTLC fee is just required as a temporary hedge against the channel balance. Normally, if the payment does not fail hard, the (properly accounted for) fees are left as 'our' balance on the channel.

ZmnSCPxj:
Ah, I understand a little better, then.

Even so, would it not be better for plugin to use `feerates` command?

https://github.com/ElementsProject/lightning/blob/11412435ce0cae92205178db9a9f4b64d0d78830/lightningd/chaintopology.c#L527-L531

Plugin would need to do some retries possibly, as feerate may change, though.

m-schmoock:
> Even so, would it not be better for plugin to use `feerates` command?

Yes, but the problem is the plugin does not know the Bitcoin transaction nor its size that is computed by the daemon. Actually, the daemon itself does not know, as is the case on the 'second' drain the transaction is smaller, thus less fee, thus the second drain, thus the whole mess...

m-schmoock:
Maybe this can be closed when this is polished: https://github.com/ElementsProject/lightning/pull/2688

m-schmoock:
@rustyrussell @cdecker @ZmnSCPxj  I changed it slightly to directly tell the client about the htlc_fee in satoshi that was used. I think its finished and it can be reviewed.

I also tested this successfully in https://github.com/lightningd/plugins/pull/22

cdecker:
Is there a point in avoiding having to parse the error message to extract that information? It just feels strange that we have all the information needed, smash it into a string, and then have to parse it on the caller side again.

m-schmoock:
@cdecker there is no API for all the logic that is in `channeld/full_channel.c` regarding calculation of TX fee for untrimmed outputs etc. Do you know how to get and calculate this information at a plugin level? I would be glad to do so... Only thing we have is `rpc.feerates` but that does not help, as I cant create the correct HTLC transaction.

PR #2695, opened 2019-05-31
Fix flakes funding_reorg_private openchannel_hook and no_fee_estimate

m-schmoock:
We had two tests that had a race condition when closing a channel and then doing `bitcoind.generate_block(1)` that the bitcoind did not have the close tx in the mempool yet.

This PR adds missing `wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)` before calling `generate_block`

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1266,6 +1266,7 @@ def test_no_fee_estimate(node_factory, bitcoind, executor):
 
     # Can do mutual close.
     l1.rpc.close(l2.info['id'])
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)
     bitcoind.generate_block(100)
 
     # Can do unilateral close.
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -942,6 +942,7 @@ def test_funding_reorg_private(node_factory, bitcoind):
     wait_for(lambda: [c['active'] for c in l2.rpc.listchannels('108x1x0')['channels']] == [True, True])
 
     l1.rpc.close(l2.info['id'])                     # to ignore `Bad gossip order` error in killall
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)
     bitcoind.generate_block(1)
     l1.daemon.wait_for_log(r'Deleting channel')
     l2.daemon.wait_for_log(r'Deleting channel')
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -303,6 +303,7 @@ def test_openchannel_hook(node_factory, bitcoind):
 
     # Close it.
     l1.rpc.close(l2.info['id'])
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)
     bitcoind.generate_block(1)
     wait_for(lambda: [c['state'] for c in only_one(l1.rpc.listpeers(l2.info['id'])['peers'])['channels']] == ['ONCHAIN'])
 


m-schmoock:
same can be done tests/test_connection.py::test_no_fee_estimate   l1.wait_for_channel_onchain(l2.info['id'])  needs a loop callback and we can 'fix' the bitcoind.generate_block(100) few lines above

tests/test_connection.py
```diff
@@ -1266,7 +1266,7 @@ def test_no_fee_estimate(node_factory, bitcoind, executor):
 
     # Can do mutual close.
     l1.rpc.close(l2.info['id'])
-    bitcoind.generate_block(100)
+    bitcoind.wait_for_log(r'txs: 1', None, lambda: bitcoind.generate_block(1))
```
trueptolemy: there may be bitcoind.wait_for_log(r'txs: 1', None, lambda: bitcoind.generate_block(100)) ?

tests/test_connection.py
```diff
@@ -1277,7 +1277,7 @@ def test_no_fee_estimate(node_factory, bitcoind, executor):
     l1.wait_for_channel_onchain(l2.info['id'])
     bitcoind.generate_block(5)
     wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)
-    bitcoind.generate_block(100)
+    bitcoind.wait_for_log(r'txs: 1', None, lambda: bitcoind.generate_block(1))
```
trueptolemy: and here, like above

tests/test_connection.py
ZmnSCPxj: As I understand it the only reason it was `generate_block(100)` before is that it was hedging that the single transaction it is interested in is present in one of those 100 blocks.

This is a more precise and more reliable wait.

ZmnSCPxj: As I understand it the only reason it was `generate_block(100)` before is that it was hedging that the single transaction it is interested in is present in one of those 100 blocks.

This is a more precise and more reliable wait.

m-schmoock: Nope, this was intentional as we just need one block with the TX. the 100 was overkill to make an indirect 'sleep' effect where eventually the TX got included and mined.

tests/test_connection.py
m-schmoock: see above

m-schmoock:
@cdecker for the record (since IRC logs are volatile) using/waiting for `getrawmempool` is not good, because the tests are still flaky. I will try something else...

m-schmoock:
@cdecker unfortunately, `getblocktemplate` currently does not work in regtest mode. we need to change the testsuite and add another regtest peer to get around that. Do you have other ideas or can we stick to a callback lambda?

If we get an agreement, we should also fix all the other `generate_block(100)` unit tests.

m-schmoock:
Okay,

for the record, our two other options here:

- (cdecker) how about we create a new function `confirm_txs` that'll tentatively generate a block and discard it with `invalidateblock` if it doesn't contain all the transactions we expect, then try again?
- adding an additional 'miner' to regtest suite, so getblocktemplate is allowed to work

rustyrussell:
> @cdecker for the record (since IRC logs are volatile) using/waiting for `getrawmempool` is not good, because the tests are still flaky. I will try something else...

In theory, getrawmempool could be async to the block generation, but every time I thought I caught bitcoind doing that I was wrong.  Are you sure?  Frankly, this is a case where sleep(1) might be the right answer.

Using invalidateblock is too much perturbation tests, since sometimes they'll see an additional block height.

m-schmoock:
@cdecker @rustyrussell 
Okay, it looks like there was just missing `wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)` for these three tests. They do are now stable on my local machine. Lets see what Travis thinks.

We can think about adding `wait_for(lambda: len(bitcoind.rpc.getrawmempool()) > 0)` into `generate_block(1, True)` , where the second boolean parameter tells to wait for mempool before generating.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2695/commits/1379cd386352ed1efea92174ce0b0e46bb7b8b35

PR #2696, opened 2019-05-31
json-rpc: Annotate transactions of interest and add the `listtransactions` RPC method

cdecker:
This PR extends the capabilities of our internal transaction tracking by annotating transactions of interest with a transaction type and a `channel_id`. This allows us to group transactions by the channel they affected, trace the on-chain footprint of a channel (improving our debuggability) and showing on-chain transaction history in wallet UIs.

 - The `channel_id` in the `transactions` table is not a foreign key into the `channels` table since we delete entries from that table, which would reset that grouping key. I found it better to keep them set, resulting in them being more of a grouping key rather than a real reference to a channel.
 - Transactions are inserted and then annotated with their type aposteriori since we add transactions in the main daemon based on `tx_watches` or `txo_watches` but only subdaemons (`onchaind` in particular) know their type. It's further complicated by the fact that some watch callbacks actually require the transaction to be in the `transactions` table before being called. The two alternatives I have tried were:
   - Have the `tx_watch` and `txo_watch` entries be marked with `type` and `channel_id` so that we can insert them with the correct values right away. This failed since for `onchaind` we automatically track descendants of a tracked transaction, and we'd have to ask `onchaind` for its type anyway.
    - Have the callbacks for the watches insert the transaction and annotate them at the same time. This also fails for `onchaind`, but in some places where we call `wallet_transaction_add` directly followed by `wallet_transaction_annotate` this worked out fine. I found the overhead small enough to warrant not combining the two into a single call.
 - I had a teardown test during testing that'd check that no `TX_UNKNOWN` transactions are left in the DB in the end, but decided against including it since it produced some false positives for tests where the node gets killed (mainly cheat attempts) and it was triggering on remote sweep transactions as well, which we just unwatch right away again. If desired I can include that teardown test again, but it'll need instrumentation to avoid false positives.
 - I need to rework the way we confirm transactions a bit, since recording a `tx_watch` for each of the transactions is a bit messy, and I'd like to merge the `transactions` and `channeltxs` tables in a future PR.

Open questions:
 - There are no migrations attempting to infer the type of existing transactions in the table yet. I'm not sure we want to spend extra cycles on that since we are now tracking more transactions that will not be present anyway (withdrawals were not tracked at all), and it is bound to reverse-engineer in an automated fashion what type a transaction had. What do the others think? Is keeping them `TX_UNKNOWN` good enough or how much work should we invest here?
 - Like mentioned above I'd like to revamp the way we track confirmations. That'd be for another PR, or would you prefer we include it here?

### DIFF
--- a/onchaind/onchain_wire.csv
+++ b/onchaind/onchain_wire.csv
@@ -99,3 +99,10 @@ onchain_dev_memleak,5033
 
 onchain_dev_memleak_reply,5133
 onchain_dev_memleak_reply,,leak,bool
+
+# Tell the main daemon what we've been watching, mainly used for transactions
+# that we tracked automatically but only onchaind knows how to classify their
+# transactions.
+onchain_transaction_annotate,5034
+onchain_transaction_annotate,,txid,struct bitcoin_txid
+onchain_transaction_annotate,,type,u16
--- a/onchaind/onchain_wire.csv
+++ b/onchaind/onchain_wire.csv
@@ -48,6 +48,7 @@ onchain_init_reply,5101
 # onchaind->master: Send out a tx.
 onchain_broadcast_tx,5003
 onchain_broadcast_tx,,tx,struct bitcoin_tx
+onchain_broadcast_tx,,type,u16
 
 # master->onchaind: Notifier that an output has been spent by input_num of tx.
 onchain_spent,5004
--- a/onchaind/onchain_wire.csv
+++ b/onchaind/onchain_wire.csv
@@ -1,5 +1,6 @@
 #include <common/derive_basepoints.h>
 #include <common/htlc_wire.h>
+#include <common/wallet.h>
 
 # Begin!  Here's the onchain tx which spends funding tx, followed by all HTLCs.
 onchain_init,5001
@@ -48,7 +49,7 @@ onchain_init_reply,5101
 # onchaind->master: Send out a tx.
 onchain_broadcast_tx,5003
 onchain_broadcast_tx,,tx,struct bitcoin_tx
-onchain_broadcast_tx,,type,u16
+onchain_broadcast_tx,,type,enum wallet_tx_type
 
 # master->onchaind: Notifier that an output has been spent by input_num of tx.
 onchain_spent,5004
@@ -105,4 +106,4 @@ onchain_dev_memleak_reply,,leak,bool
 # transactions.
 onchain_transaction_annotate,5034
 onchain_transaction_annotate,,txid,struct bitcoin_txid
-onchain_transaction_annotate,,type,u16
+onchain_transaction_annotate,,type,enum wallet_tx_type
--- a/onchaind/onchaind.c
+++ b/onchaind/onchaind.c
@@ -15,6 +15,7 @@
 #include <common/type_to_string.h>
 #include <common/utils.h>
 #include <common/version.h>
+#include <common/wallet.h>
 #include <errno.h>
 #include <hsmd/gen_hsm_wire.h>
 #include <inttypes.h>
@@ -457,6 +458,41 @@ static void ignore_output(struct tracked_output *out)
 	out->resolved->tx_type = SELF;
 }
 
+static txtypes onchain_txtype_to_wallet_txtype(enum tx_type t)
+{
+	switch (t) {
+	case FUNDING_TRANSACTION:
+		return TX_CHANNEL_FUNDING;
+	case MUTUAL_CLOSE:
+		return TX_CHANNEL_CLOSE;
+	case OUR_UNILATERAL:
+		return TX_CHANNEL_UNILATERAL;
+	case THEIR_HTLC_FULFILL_TO_US:
+	case OUR_HTLC_SUCCESS_TX:
+		return TX_CHANNEL_HTLC_SUCCESS;
+	case OUR_HTLC_TIMEOUT_TO_US:
+	case OUR_HTLC_TIMEOUT_TX:
+		return TX_CHANNEL_HTLC_TIMEOUT;
+	case OUR_DELAYED_RETURN_TO_WALLET:
+	case SELF:
+		return TX_CHANNEL_SWEEP;
+	case OUR_PENALTY_TX:
+		return TX_CHANNEL_PENALTY;
+	case THEIR_UNILATERAL:
+	case UNKNOWN_UNILATERAL:
+	case THEIR_REVOKED_UNILATERAL:
+		return TX_CHANNEL_UNILATERAL | TX_THEIRS;
+	case THEIR_HTLC_TIMEOUT_TO_THEM:
+		return TX_CHANNEL_HTLC_TIMEOUT | TX_THEIRS;
+	case OUR_HTLC_FULFILL_TO_THEM:
+		return TX_CHANNEL_HTLC_SUCCESS | TX_THEIRS;
+	case IGNORING_TINY_PAYMENT:
+	case UNKNOWN_TXTYPE:
+		return TX_UNKNOWN;
+	}
+	abort();
+}
+
 static void proposal_meets_depth(struct tracked_output *out)
 {
 	/* If we simply wanted to ignore it after some depth */
@@ -471,9 +507,11 @@ static void proposal_meets_depth(struct tracked_output *out)
 		     tx_type_name(out->tx_type),
 		     output_type_name(out->output_type));
 
-	wire_sync_write(REQ_FD,
-			take(towire_onchain_broadcast_tx(NULL,
-							 out->proposal->tx)));
+	wire_sync_write(
+	    REQ_FD,
+	    take(towire_onchain_broadcast_tx(
+		NULL, out->proposal->tx,
+		onchain_txtype_to_wallet_txtype(out->proposal->tx_type))));
 
 	/* Don't wait for this if we're ignoring the tiny payment. */
 	if (out->proposal->tx_type == IGNORING_TINY_PAYMENT) {
--- a/onchaind/onchaind.c
+++ b/onchaind/onchaind.c
@@ -458,7 +458,7 @@ static void ignore_output(struct tracked_output *out)
 	out->resolved->tx_type = SELF;
 }
 
-static txtypes onchain_txtype_to_wallet_txtype(enum tx_type t)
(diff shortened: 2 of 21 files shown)


wallet/wallet.h
```diff
@@ -270,6 +289,15 @@ struct channeltx {
 	u32 depth;
 };
 
+struct wallet_transaction {
```
darosior: Isn't having two "wallet transaction" structures a little bit ambiguous ? https://github.com/ElementsProject/lightning/blob/0c189fe3a751d56777d084c1f5ccbcbb1959c837/common/wallet_tx.h#L13

wallet/walletrpc.c
```diff
@@ -610,3 +618,76 @@ static const struct json_command dev_rescan_output_command = {
...
+		}
+		json_object_end(response);
+	}
+	json_array_end(response);
+	json_object_end(response);
+	return command_success(cmd, response);
+}
+
+static const struct json_command listtransactions_command = {
+    "listtransactions", json_listtransactions,
```
darosior: Maybe add a command category if #2658 gets merged ?

rustyrussell:
This PR is probably sufficient, though I have yet to review.  I'm itching to get 0.7.1 out.

My main concern is that Bring-Your-Own-Fees means we will combine/chain txs.  This means someone needs to be able to re-sign or re-create txs when one gets replaced, eg. unilateral close race.  Even more ambitious, we might want to RBF our own txs if we get another thing we want to spend.

cdecker:
> My main concern is that Bring-Your-Own-Fees means we will combine/chain txs. This means someone needs to be able to re-sign or re-create txs when one gets replaced, eg. unilateral close race. Even more ambitious, we might want to RBF our own txs if we get another thing we want to spend.

What we do here is basically just track transactions we sent, and transactions concerning our channels that are sent by the counterparty. Tracking in this case means adding a type, associating them with a channel and tracking confirmation status. If we have things like splicing or combine sweeps, we'll just add those types to the transactions, which is already supported. For RBF the confirmation status becomes more interesting since we want to filter by confirmation status there.

There is one limiting factor in the current PR, namely that we have a many-to-one relationship from transaction to channel, so for splicing with changing `channel_id` or sweep across multiple channels (unlikely due to CLTV incompatibilies) we lose expressiveness. Is that something we should address now by creating a `transaction_channel` associative table that make this a many-to-many relationship just for safety?

wallet/wallet.h
```diff
@@ -215,6 +215,25 @@ static inline enum wallet_payment_status wallet_payment_status_in_db(enum wallet
...
+/* Types of transactions we store in the `transactions` table. Mainly used for
+ * display purposes later. */
+enum wallet_tx_type {
+       TX_UNKNOWN = 0,
+       TX_THEIRS = 1,  /* This only affects their funds in the channel */
+       TX_WALLET_DEPOSIT = 2,
+       TX_WALLET_WITHDRAWAL = 4,
+       TX_CHANNEL_FUNDING = 8,
+       TX_CHANNEL_CLOSE = 16,
+       TX_CHANNEL_UNILATERAL = 32,
```
rustyrussell: We don't distinguish their unilateral from ours?

wallet/wallet.h
```diff
@@ -215,6 +215,25 @@ static inline enum wallet_payment_status wallet_payment_status_in_db(enum wallet
...
+       TX_CHANNEL_SWEEP = 64,
+       TX_CHANNEL_HTLC_SUCCESS = 128,
+       TX_CHANNEL_HTLC_TIMEOUT = 256,
+       TX_CHANNEL_PENALTY = 512,
+       TX_CHANNEL_CHEAT = 1024,
+};
+/* Any combination of the above wallet_tx_types */
+typedef unsigned short txtypes;
```
rustyrussell: In C (unlike C++!) enums can have any integer value for exactly this reason.  So this typedef is unnecessary; enum wallet_tx_type can be used directly.

onchaind/onchain_wire.csv
```diff
@@ -48,6 +48,7 @@ onchain_init_reply,5101
 # onchaind->master: Send out a tx.
 onchain_broadcast_tx,5003
 onchain_broadcast_tx,,tx,struct bitcoin_tx
+onchain_broadcast_tx,,type,u16
```
rustyrussell: Use enum here, and define towire_wallet_txtype and fromwire_ somewhere?

This means moving wallet_txtypes out to common; maybe just call it 'txtype' then?

wallet/wallet.h
rustyrussell: Ah, I see, THEIRS is a bit, the rest are distinct.  Maybe we should just make those three cases distinct?

TX_CHANNEL_OUR_UNILATERAL / TX_CHANNEL_THEIR_UNILATERAL, TX_CHANNEL_OUR_HTLC_SUCCESS ... TX_CHANNEL_OUR_HTLC_TIMEOUT...?

That gets us closer to the onchain types, too.

onchaind/onchaind.c
```diff
@@ -457,6 +458,41 @@ static void ignore_output(struct tracked_output *out)
...
+	switch (t) {
+	case FUNDING_TRANSACTION:
+		return TX_CHANNEL_FUNDING;
+	case MUTUAL_CLOSE:
+		return TX_CHANNEL_CLOSE;
+	case OUR_UNILATERAL:
+		return TX_CHANNEL_UNILATERAL;
+	case THEIR_HTLC_FULFILL_TO_US:
+	case OUR_HTLC_SUCCESS_TX:
+		return TX_CHANNEL_HTLC_SUCCESS;
```
rustyrussell: I had to re-read the code to see the difference between these two tx types!  `THEIR_HTLC_FULFILL_TO_US` is their unilateral close (simple case, no timeouts) and `OUR_HTLC_SUCCESS_TX` is our own unilateral close (using the sig they gave us, with timeout).

Does it make sense to distinguish these at the wallet level?  Maybe, because the second has a timeout.  Similarly OUR_HTLC_TIMEOUT_TO_US and OUR_HTLC_TIMEOUT_TX.

onchaind/onchaind.c
```diff
@@ -23,6 +23,7 @@
 #include <onchaind/onchain_types.h>
 #include <stdio.h>
 #include <unistd.h>
+#include <wallet/wallet.h>
```
rustyrussell: Definitely belongs in common...

onchaind/onchaind.c
```diff
@@ -1093,6 +1093,7 @@ static void output_spent(struct tracked_output ***outs,
 	status_trace("Notified about tx %s output %u spend, but we don't care",
 		     type_to_string(tmpctx, struct bitcoin_txid, &txid),
 		     tx->wtx->inputs[input_num].index);
+
```
rustyrussell: ENODIFF?

wallet/walletrpc.c
```diff
@@ -617,3 +618,76 @@ static const struct json_command dev_rescan_output_command = {
...
+    {TX_WALLET_DEPOSIT, "deposit"},
+    {TX_WALLET_WITHDRAWAL, "withdraw"},
+    {TX_CHANNEL_FUNDING, "channel_funding"},
+    {TX_CHANNEL_CLOSE, "channel_mutual_close"},
+    {TX_CHANNEL_UNILATERAL, "channel_unilateral_close"},
+    {TX_CHANNEL_SWEEP, "channel_sweep"},
+    {TX_CHANNEL_HTLC_SUCCESS, "channel_htlc_success"},
+    {TX_CHANNEL_HTLC_TIMEOUT, "channel_htlc_timeout"},
+    {TX_CHANNEL_PENALTY, "channel_penalty"},
+    {TX_CHANNEL_CHEAT, "channel_unilateral_cheat"},
```
rustyrussell: Another reason to make the enums distinct, put them in a common header file, and autogenerate this?

wallet/wallet.h
cdecker: Hm, thought I'd make it explicit, not that people start using `switch` statements over the `enum` and then wonder why they don't catch all the variants. Will remove.

wallet/wallet.h
cdecker: Not really, a close transaction is also a deposit to us if we had funds in them, but annotate at different places, creating all possible combinations is likely not what we want.

But I agree that `TX_THEIRS` is a bit strange since it indicates both who originated the transaction as well as filtering out transactions we don't have a stake in (e.g., their sweeps).

wallet/walletrpc.c
cdecker: I coded them manually as I wanted them to be nicely readable since they end up in the JSON output. These basically drop the `TX_` prefix, and have a LESS SHOUTY lowercase :wink:

cdecker:
> OK, reading this it's clear to me that annotating transactions is not actually what we want. We want to annotate outputs, and imply transactions from that.

Interesting idea, I like it. The reason I annotated transactions and not outputs is because the original issue I tried to address was having a way to list our on-chain transactions, and centralizing all of our full transaction tracking in the existing `transactions` table sounded like a good idea.

Tracking outputs requires us to have a separate table `transaction_outputs` with the following format:

```sql
CREATE TABLE transaction_outputs (
    transaction_id BLOB REFERENCES transactions(id) ON DELETE CASCADE,
    outnum INTEGER,
    type INTEGER,
    channel_id
);
```

In addition we also need to annotate inputs in some cases (such as withdrawals and channel closes), requiring yet another table. Then representing that in the JSON result requires additional logic (JOINs duplicate the shared entries so we iterate over entries until the `transactions.id` changes, and a double join with inputs and outputs just gets really annoying).

I'm happy to give a shot though, if the extra effort is warranted :-)

rustyrussell:
OK, after offline discussion with @cdecker we're going to defer a per-input & output tracking for post 0.7.1, and apply this as an obvious win now.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2696/commits/4b262e8faaea67c8af232478e4bc68ec25fc109b

rustyrussell:
autosquashed...

Ack https://github.com/ElementsProject/lightning/commit/6f25317a33fc74b6d93ddfb81f6b428fb66d83d4

PR #2699, opened 2019-06-02
fundchannel: add the possibility to fund a channel with specific utxos

### LINKED ISSUE #2682: Feature: open channel with specific UTXO
In addition to `all` it would be nice if the `satoshi` argument for `fundchannel` can be a specific UTXO.

darosior:
Fixes #2682 .

This PR adds the possibility to specify a list of outpoints (`[\"txid:vout\", \"txid:vout\"]`) as the second parameter of the `fundchannel` call in addition to `'all'` or an amount in satoshis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - JSON API: the command objects that `help` outputs now contain a new string field : `category` (can be "bitcoin", "channels", "network", "payment", "plugins", "utility", "developer" for native commands, or any other new category set by a plugin).
 - Plugins: a plugin can now set the category of a newly created RPC command. This possibility has been added to libplugin.c and pylightning.
 - CLI: the human readable help is now more human and more readable : commands are sorted alphabetically and ordered by categories.
+- JSON API: A new parameter is added to `fundchannel`, which now accepts an utxo array to use to fund the channel.
 
 ### Deprecated
 
--- a/common/wallet_tx.c
+++ b/common/wallet_tx.c
@@ -1,4 +1,7 @@
+#include <ccan/ccan/opt/opt.h>
+#include <ccan/tal/str/str.h>
 #include <common/json_command.h>
+#include <common/json_helpers.h>
 #include <common/jsonrpc_errors.h>
 #include <common/wallet_tx.h>
 #include <inttypes.h>
@@ -23,13 +26,14 @@ struct command_result *param_wtx(struct command *cmd,
 		return NULL;
 	}
 	wtx->all_funds = false;
+
 	if (!parse_amount_sat(&wtx->amount,
 			      buffer + tok->start, tok->end - tok->start))
 		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
-				    "'%s' should be satoshis or 'all', not '%.*s'",
-				    name,
-				    tok->end - tok->start,
-				    buffer + tok->start);
+					"'%s' should be an amount in satoshis or all, not '%.*s'",
+					name,
+					tok->end - tok->start,
+					buffer + tok->start);
 
 	if (amount_sat_greater(wtx->amount, max))
                 return command_fail(wtx->cmd, FUND_MAX_EXCEEDED,
@@ -37,6 +41,66 @@ struct command_result *param_wtx(struct command *cmd,
 				    type_to_string(tmpctx, struct amount_sat,
 						   &max));
         return NULL;
+}
+
+struct command_result *param_utxos(struct command *cmd,
+				 const char *name,
+				 const char *buffer,
+				 const jsmntok_t *tok,
+				 const struct utxo ***utxos)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	struct bitcoin_txid **txids = tal_arr(cmd, struct bitcoin_txid*, 0);
+	unsigned int **outnums = tal_arr(cmd, unsigned int*, 0);
+
+	json_for_each_arr(i, curr, tok) {
+		jsmntok_t txid_tok, outnum_tok;
+		if (!split_tok(buffer, curr, ':', &txid_tok, &outnum_tok))
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+								"Could not decode the outpoint from \"%s\""
+								" The utxos should be specified as"
+								" 'txid:output_index'.",
+								json_strdup(tmpctx, buffer, curr));
+
+		struct bitcoin_txid *txid = tal(txids, struct bitcoin_txid);
+		unsigned int *outnum = tal(txids, unsigned int);
+		if (!json_to_txid(buffer, (const jsmntok_t*)&txid_tok, txid)) {
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+								"Could not get a txid out of \"%s\"",
+								json_strdup(tmpctx, buffer, &txid_tok));
+		}
+		if(!json_to_number(buffer, (const jsmntok_t*)&outnum_tok, outnum))
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+								"Could not get a vout out of \"%s\"",
+								json_strdup(tmpctx, buffer, &outnum_tok));
+
+		tal_arr_expand(&txids, txid);
+		tal_arr_expand(&outnums, outnum);
+	}
+
+	if (!tal_count(txids) || !tal_count(outnums))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+							"Please specify an array of 'txid:output_index',"
+							" not \"%.*s\"",
+							tok->end - tok->start,
+							buffer + tok->start);
+
+	*utxos = wallet_select_specific(cmd, cmd->ld->wallet, txids, outnums);
+	tal_free(txids);
+	tal_free(outnums);
+
+	if (!*utxos)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+							"Could not decode all of the outpoints. The utxos"
+							" should be specified as an array of "
+							" 'txid:output_index'.");
+	if (tal_count(*utxos) == 0)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+							"No matching utxo was found from the wallet."
+							"You can get a list of the wallet utxos with"
+							" the `listfunds` RPC call.");
+	return NULL;
 }
 
 static struct command_result *check_amount(const struct wallet_tx *wtx,
@@ -103,4 +167,65 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
 		tx->change_key_index = wallet_get_newindex(tx->cmd->ld);
 	}
 	return NULL;
+}
+
+struct command_result *wtx_from_utxos(struct wallet_tx *tx,
+					u32 fee_rate_per_kw,
+					size_t out_len,
+					u32 maxheight,
+					const struct utxo **utxos)
+{
+	size_t weight;
+	struct amount_sat total_amount, fee_estimate;
+
+	tx->change = AMOUNT_SAT(0);
+	tx->change_key_index = 0;
+	total_amount = AMOUNT_SAT(0);
+
+	/* The transaction has `tal_count(tx.utxos)` inputs and one output */
+	/* (version + in count + out count + locktime)  (index + value + script length) */
+	weight = 4 * (4 + 1 + 1 + 4) + 4 * (8 + 1 + out_len);
+	for (size_t i = 0; i < tal_count(utxos); i++) {
+		if (!*utxos[i]->blockheight || *utxos[i]->blockheight > maxheight) {
+			tal_arr_remove(&utxos, i);
+			continue;
+		}
+		/* txid + index + sequence + script_len */
+		weight += (32 + 4 + 4 + 1) * 4;
+		/* P2SH variants include push of <0 <20-byte-key-hash>> */
+		if (utxos[i]->is_p2sh)
+			weight += 23 * 4;
+		/* Account for witness (1 byte count + sig + key) */
+		weight += 1 + (1 + 73 + 1 + 33);
+		if (!amount_sat_add(&total_amount, total_amount, utxos[i]->amount))
+			fatal("Overflow when computing input amount");
+	}
+	tx->utxos = utxos;
+
+	if (!tx->all_funds && amount_sat_less(tx->amount, total_amount)
+			&& !amount_sat_sub(&tx->change, total_amount, tx->amount))
+		fatal("Overflow when computing change");
+	if (amount_sat_greater_eq(tx->change, get_chainparams(tx->cmd->ld)->dust_limit)) {
+		tx->change_key_index = wallet_get_newindex(tx->cmd->ld);
+		/* Add the change output's weight */
+		weight += (8 + 1 + out_len) * 4;
+	}
+
+	fee_estimate = amount_tx_fee(fee_rate_per_kw, weight);
+
+	if (tx->all_funds || amount_sat_eq(tx->change, AMOUNT_SAT(0))) {
+		tx->amount = total_amount;
+		if (!amount_sat_sub(&tx->amount, tx->amount, fee_estimate))
+			return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction with %s sats of fees",
+								type_to_string(tmpctx, struct amount_sat,
+									&fee_estimate));
+	} else {
+		if (!amount_sat_sub(&tx->change, tx->change, fee_estimate))
+			return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction with %s sats of fees",
+								type_to_string(tmpctx, struct amount_sat,
+									&fee_estimate));
+	}
+	return check_amount(tx, tx->amount);
 }
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -996,6 +996,7 @@ static struct command_result *json_fund_channel(struct command *cmd,
 	bool *announce_channel;
 	u8 *msg;
 	struct amount_sat max_funding_satoshi;
+	const struct utxo **chosen_utxos;
 
 	max_funding_satoshi = get_chainparams(cmd->ld)->max_funding;
 
@@ -1009,6 +1010,7 @@ static struct command_result *json_fund_channel(struct command *cmd,
 		   p_opt("feerate", param_feerate, &feerate_per_kw),
 		   p_opt_def("announce", param_bool, &announce_channel, true),
 		   p_opt_def("minconf", param_number, &minconf, 1),
+		   p_opt("utxos", param_utxos, &chosen_utxos),
 		   NULL))
 		return command_param_failed();
 
@@ -1055,8 +1057,12 @@ static struct command_result *json_fund_channel(struct command *cmd,
 	}
 
 	maxheight = minconf_to_maxheight(*minconf, cmd->ld);
-	res = wtx_select_utxos(fc->wtx, *feerate_per_kw,
-			       BITCOIN_SCRIPTPUBKEY_P2WSH_LEN, maxheight);
+	if (chosen_utxos)
+		res = wtx_from_utxos(fc->wtx, *feerate_per_kw,
+				BITCOIN_SCRIPTPUBKEY_P2WSH_LEN, maxheight, chosen_utxos);
+	else
+		res = wtx_select_utxos(fc->wtx, *feerate_per_kw,
+				BITCOIN_SCRIPTPUBKEY_P2WSH_LEN, maxheight);
 	if (res)
 		return res;
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -786,6 +786,38 @@ def test_funding_toolarge(node_factory, bitcoind):
     l1.rpc.fundchannel(l2.info['id'], amount)
 
 
+def test_funding_by_utxos(node_factory, bitcoind):
+    """Fund a channel with specific utxos"""
+    l1, l2, l3 = node_factory.line_graph(3, fundchannel=False)
+
+    # Get 3 differents utxo
+    l1.fundwallet(0.01 * 10**8)
+    l1.fundwallet(0.01 * 10**8)
+    l1.fundwallet(0.01 * 10**8)
+    wait_for(lambda: len(l1.rpc.listfunds()["outputs"]) == 3)
+
+    utxos = [utxo["txid"] + ":" + str(utxo["output"]) for utxo in l1.rpc.listfunds()["outputs"]]
+
+    # Fund with utxos we don't own
+    with pytest.raises(RpcError, match=r"No matching utxo was found from the wallet"):
+        l3.rpc.fundchannel(l2.info["id"], int(0.01 * 10**8), utxos=utxos)
+
+    # Fund with an empty array
+    with pytest.raises(RpcError, match=r"Please specify an array of \\'txid:output_index\\', not \"*\""):
+        l1.rpc.fundchannel(l2.info["id"], int(0.01 * 10**8), utxos=[])
+
+    # Fund a channel from some of the utxos, without change
+    l1.rpc.fundchannel(l2.info["id"], "all", utxos=utxos[0:2])
+
+    # Fund a channel from the rest of utxos, with change
+    l1.rpc.connect(l3.info["id"], "localhost", l3.port)
+    l1.rpc.fundchannel(l3.info["id"], int(0.007 * 10**8), utxos=[utxos[2]])
+
+    # Fund another channel with already spent utxos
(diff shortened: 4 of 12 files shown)


wallet/wallet.c
```diff
@@ -377,6 +378,43 @@ const struct utxo **wallet_select_coins(const tal_t *ctx, struct wallet *w,
...
+
+		for (j = 0; j < tal_count(available); j++) {
+			struct utxo *u = tal_steal(utxos, available[j]);
+
+			if (bitcoin_txid_eq(&u->txid, txid) && u->outnum == output_index) {
+				    tal_arr_expand(&utxos, u);
+			    if (!wallet_update_output_status(
+				    w, &available[j]->txid, available[j]->outnum,
+				    output_state_available, output_state_reserved))
+				    fatal("Unable to reserve output");
```
rustyrussell: Um, does this mean I crash the daemon if I try to fund two channels at once from the same UTXO?

common/wallet_tx.c
```diff
@@ -39,6 +46,37 @@ struct command_result *param_wtx(struct command *cmd,
...
+				 const char *buffer,
+				 const jsmntok_t *tok,
+				 struct wallet_tx *wtx)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	const char **outpoints = tal_arr(cmd, const char*, 0);
+
+	json_for_each_arr(i, curr, tok)
+		tal_arr_expand(&outpoints, json_strdup(tmpctx, buffer, curr));
```
rustyrussell: You want to use `outputs` and the context for json_strdup here: it's a nice convention that a container owns all its members.

common/wallet_tx.c
```diff
@@ -39,6 +46,37 @@ struct command_result *param_wtx(struct command *cmd,
...
+	if (!tal_count(outpoints))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+			"Could not decode any of the outpoints. The utxos should be"
+			" specified as an array of 'txid:output_index'.");
```
rustyrussell: Empty array check should probably be left to later.

common/wallet_tx.c
rustyrussell: In `listfunds` we use `txid` and `output` fields.  Since we have JSON anyway, I suggest sticking with that, rather than using a boutique encoding for outpoints?

common/wallet_tx.c
```diff
@@ -39,6 +46,37 @@ struct command_result *param_wtx(struct command *cmd,
...
+
+	if (!wtx->utxos)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+			"Could not decode all of the outpoints. The utxos should be"
+			" specified as an array of 'txid:output_index'.");
+	if (tal_count(wtx->utxos) == 0)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				"No matching utxo was found from the wallet."
+				"You can get a list of the wallet utxos with"
+				" the `listfunds` RPC call.");
```
rustyrussell: Parsing here, before asking wallet_select_specific(), makes this code path simpler too.  If you can't decode txid or outnum, you give a meaningful error here.

Just hand wallet_select_specific an array of txids and an array of outnums: a failure means some outpoint wasn't available.

wallet/wallet.c
```diff
@@ -377,6 +378,43 @@ const struct utxo **wallet_select_coins(const tal_t *ctx, struct wallet *w,
...
+	size_t i, j;
+	char **outpoint;
+	unsigned int output_index;
+	struct bitcoin_txid *txid;
+	struct utxo **available = wallet_get_utxos(ctx, w, output_state_available);
+	const struct utxo **utxos = tal_arr(ctx, const struct utxo*, 0);
+	tal_add_destructor2(utxos, destroy_utxos, w);
+
+	for (i = 0; i < tal_count(outpoints); i++)
+	{
```
rustyrussell: style nit: `{` belongs on line above.

wallet/wallet.c
```diff
@@ -377,6 +378,43 @@ const struct utxo **wallet_select_coins(const tal_t *ctx, struct wallet *w,
...
 }
 
+const struct utxo **wallet_select_specific(const tal_t *ctx, struct wallet *w,
+					const char **outpoints)
+{
+	size_t i, j;
+	char **outpoint;
+	unsigned int output_index;
+	struct bitcoin_txid *txid;
+	struct utxo **available = wallet_get_utxos(ctx, w, output_state_available);
```
rustyrussell: available should either be allocated off NULL (which by convention, indicates you're explicitly going to take/free it, as you do below), or just off tmpctx and forget about it.

wallet/wallet.c
```diff
@@ -377,6 +378,43 @@ const struct utxo **wallet_select_coins(const tal_t *ctx, struct wallet *w,
...
+		if (!bitcoin_txid_from_hex(outpoint[0], strlen(outpoint[0]), txid)
+				|| opt_set_uintval(outpoint[1], &output_index))
+			return NULL;
+
+		for (j = 0; j < tal_count(available); j++) {
+			struct utxo *u = tal_steal(utxos, available[j]);
+
+			if (bitcoin_txid_eq(&u->txid, txid) && u->outnum == output_index) {
+				    tal_arr_expand(&utxos, u);
+			    if (!wallet_update_output_status(
```
rustyrussell: Weird indent?

tests/test_connection.py
```diff
@@ -753,6 +753,20 @@ def test_funding_toolarge(node_factory, bitcoind):
...
+    l1, l2 = node_factory.line_graph(2, fundchannel=False)
+
+    # Get 2 differents utxo
+    l1.fundwallet(0.01 * 10**8)
+    l1.fundwallet(0.01 * 10**8)
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == 2)
+
+    # Then fund a channel with them
+    utxos = [utxo["txid"] + ":" + str(utxo["output"]) for utxo in l1.rpc.listfunds()["outputs"]]
+    l1.rpc.fundchannel(l2.info["id"], utxos)
```
rustyrussell: OK, test that you can't grab some bogus UTXO too?  And maybe an already-consumed UTXO?

wallet/wallet.c
darosior: I think so, as in the already used function to select outputs : https://github.com/ElementsProject/lightning/blob/35827111ee837fc0118e0190b9e12539adae81b3/wallet/wallet.c#L315-L318
Anyway we consume the whole utxos if they are directly passed to `fundchannel` so one could not fund 2 channels, or am I missing something ?

common/wallet_tx.c
darosior: > In listfunds we use txid and output fields. Since we have JSON anyway, I suggest sticking with that, rather than using a boutique encoding for outpoints?

Does it mean that the user should pass an array of objects ? `[{"txid":"....", "output":0}, ...]`

tests/test_connection.py
darosior: Ok, how could I get a bogus UTXO ?

wallet/wallet.c
ZmnSCPxj: I agree with @darosior, line 388 of the new code is `wallet_get_utxos` which ensures that we can only change update status of UTXOs with `output_state_available`.  Will mark as resolved.

tests/test_connection.py
ZmnSCPxj: Dump 32 bytes from `/dev/random` for txid plus 2 bytes for outnum?

darosior:
@rustyrussell thank you for the review. I rebased to apply most of the requested changes : 
- Parse outpoints before calling the utxo selection function
- Add some test for failing cases

Added a CHANGELOG brag line :)

tests/test_connection.py
darosior: I didn't see your comment and finally added a test that makes a node funding a channel with utxos it doesn't own, is it OK ? It takes the same code path.

ZmnSCPxj: It is fine.

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+				 const char *name,
+				 const char *buffer,
+				 const jsmntok_t *tok,
+				 struct wallet_tx *wtx)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	struct bitcoin_txid **txids;
+	unsigned int **outnums;
+	char **outpoint = tal_arr(tmpctx, char*, 0);
```
niftynei: nit: it's weird to me that you allocate this one inline with the declaration, but the others (`txids` and `outnums`) below.

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+
+	txids = tal_arr(cmd, struct bitcoin_txid*, 0);
+	outnums = tal_arr(cmd, unsigned int*, 0);
+	json_for_each_arr(i, curr, tok) {
+		outpoint = tal_strsplit(cmd, json_strdup(cmd, buffer, curr), ":",
+				STR_NO_EMPTY);
```
niftynei: nit: you should line up the beginning of this with the end of the `(` above

```
tal_strsplit(cmd ..
             STR_NO_EMPTY
```

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+		if (tal_count(outpoint) < 3 || strlen(outpoint[0]) != 64)
```
niftynei: Why are you expecting 'outpoint' to be at least 3 here? Why not check that it's exactly 3? Moving 3 and 64 to constants might be a better way of expressing what you're checking as well.

This would be easier to review if you had included an example of the expected rpc call in the commit notes.

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+			break;
+
+		struct bitcoin_txid *txid = tal(tmpctx, struct bitcoin_txid);
+		unsigned int *outnum = tal(tmpctx, unsigned int);
```
niftynei: It's weird to see `unsigned int`s allocated, but maybe I'm just new to this.

You probably want to allocate these off of the same thing as the array that you're putting them in, either `cmd` or `outnums`.

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+		if (!bitcoin_txid_from_hex(outpoint[0], strlen(outpoint[0]), txid)
+				|| opt_set_uintval(outpoint[1], outnum))
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				"Could not decode the outpoint with txid \"%s\" and "
```
niftynei: nit: should be aligned with `(` in previous line.

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+		tal_arr_expand(&outnums, outnum);
+	}
+
+	if (!tal_count(txids) || !tal_count(outnums))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+			"Please specify an amount in satoshis, 'all' or an array"
+			" of 'txid:output_index'.");
+
+	wtx->utxos = wallet_select_specific(cmd, cmd->ld->wallet, txids, outnums);
+	tal_free(txids);
```
niftynei: if the `command_fail`s at any point previous to this, do we need to call `tal_free` for each of those branches as well?

common/wallet_tx.c
```diff
@@ -39,6 +48,59 @@ struct command_result *param_wtx(struct command *cmd,
...
+	tal_free(outnums);
+
+	if (!wtx->utxos)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+			"Could not decode all of the outpoints. The utxos should be"
+			" specified as an array of 'txid:output_index'.");
+	if (tal_count(wtx->utxos) == 0)
```
niftynei: nit: you can also do `!tal_count(wtx->utxos)`

common/wallet_tx.c
```diff
@@ -104,3 +166,34 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+					u32 fee_rate_per_kw,
+					size_t out_len)
+{
+	struct command_result *res;
+	size_t weight, i;
+
+	tx->amount = AMOUNT_SAT(0);
+	/* The transaction has `tal_count(tx.utxos)` inputs and one output */
+	/* version + in count + out count + locktime    index + value + script length */
+	weight = 4 * (4 + 1 + 1 + 4) + 4 * (8 + 1 + out_len);
```
niftynei: Technically the `out_len` might be longer than `1` but this is probably fine.

```
(8 + 1 + out_len)
```

lightningd/opening_control.c
```diff
@@ -1059,8 +1059,15 @@ static struct command_result *json_fund_channel(struct command *cmd,
...
-			       BITCOIN_SCRIPTPUBKEY_P2WSH_LEN, maxheight);
+	if (tal_count(fc->wtx.utxos)) {
+		/* We consume the whole specified utxo(s) */
+		fc->wtx.change = AMOUNT_SAT(0);
+		fc->wtx.change_key_index = 0;
+		res = wtx_fees_from_utxos(&fc->wtx, *feerate_per_kw,
+				BITCOIN_SCRIPTPUBKEY_P2WSH_LEN);
+	} else
+		res = wtx_select_utxos(&fc->wtx, *feerate_per_kw,
+				BITCOIN_SCRIPTPUBKEY_P2WSH_LEN, maxheight);
```
niftynei: nit: line up with `(`

common/wallet_tx.c
```diff
@@ -104,3 +166,34 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+	for (i = 0; i < tal_count(tx->utxos); ++i) {
```
niftynei: nit: use postfix, not prefix `++i` -> `i++`

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Config: Increased default amount for minimal channel capacity from 1k sat to 10k sat.
 - Build: Non-developer builds are now done with "-Og" optimization.
 - JSON API: `pay` will no longer return failure until it is no longer retrying; previously it could "timeout" but still make the payment.
+- JSON API: `fundchannel` now accepts an utxo array to use to fund the channel as the second parameter (in addition to 'all' or an amount in sats)
```
niftynei: nit: instead of saying `in addition to`, I'd reword this to `instead of` as you can only provide one of the three ('all', satoshis, or a utxo set)

common/wallet_tx.c
rustyrussell: ... I prefer == 0 here, since it's literally a count.

rustyrussell:
Agree with all @niftynei feedback.  In particular, I think adding a separate optional 'utxos' param is nicer: if not specified, utxos are selected from current wallet utxos as now.  That does allow us to have a change output from UTXOs if we want to (or 'all' if we don't).

common/wallet_tx.c
darosior: It seemed weird to me too : I was not aware of the convention @rustyrussell pointed out in his review (to not directly allocate a pointer I will free). Since I dont free this one (`tmpctx`) I thought it was ok to allocate it directly.
I have to admit I'm a little bit confused with which style to use each time I declare a variable.

common/wallet_tx.c
darosior: I test the size of the outpoint array (which always ends with NULL and must have 2elements). I put this condition while testing and should have changed it.

common/wallet_tx.c
darosior: I allocated them off `tmpctx` because they are temo variable, however I'm also new to this and am not sure it's the correct way.

common/wallet_tx.c
darosior: No because tal allocated variables passed to this function are off `cmd`, which is desallocated on `comand_error` and then so they are

common/wallet_tx.c
darosior: That's why it's a variable

CHANGELOG.md
darosior: Ok. I appreciate that you review my english too, since I'm not a native speaker.

darosior:
> Lots of little things, nice work overall!
> 
> I think the biggest question is whether or not we want to have this in addition to the 'amount' field. It seems like it'd be nice to be able to specify a utxo set and a total amount to fund as well, which isn't possible with the current design.
> 
> How you do you see people using this @darosior ?

Thank you for the review.
I see people using it in a few corner cases : a traditional user doing `newaddr`, `connect`, and `fundchannel` won't use it at all. Some users having a lot of tiny inputs wanting more control over them could use it. Actually I think most people who'll use it will consume the whole utxo, that's why I made this `command_result` of `command_result` in case the parameter is an array : I thought making another parameter will just make them copy paste the utxo amount.

common/wallet_tx.c
niftynei: Ok cool. Moving to JSON format like rusty recommended above will probably change this too :)

rustyrussell:
Ideally, we'd allow them to specify any combination, but that's a bit ambitious.  I really think a "utxos" arg is best here.

This will combine well with #2713 too

darosior:
Ok, let's do this

common/wallet_tx.c
darosior: I don't think, since I need to decode the txid and the outpoint out of the jmn token

darosior: Ok, finally made a json_helper for json token to transaction id :D

darosior: Oh, I just saw that you did the same in your external wallet funding https://github.com/ElementsProject/lightning/pull/2672/commits/d8dcbce4485b91a9b4e6f9037270fedf18b19002

rustyrussell:
Sorry, you really can't do the API this way.  As @niftynei points out, you may well want to specify both utxos and the amount to spend, so it needs to be a separate 'utxos' argument.

And while txid:outnum is neat on the commandline, it's nasty JSON.  You can always create a plugin which does converts from this syntax :)

darosior:
> Sorry, you really can't do the API this way. As @niftynei points out, you may well want to specify both utxos and the amount to spend, so it needs to be a separate 'utxos' argument.

When saying "let's do this" I meant "ok I'll work on the separate 'utxos' argument version (I was not pushing the initial implementation without argument ^^), it a translated sentence from french and has apparently not the same signification in english.

> And while txid:outnum is neat on the commandline, it's nasty JSON. You can always create a plugin which does converts from this syntax :)

I just saw your comment now that I finished to work on the new implementation. I finally kept using JSON as you requested (I added a bitcoin_txid json helper), relying on the fact that a txid would always be 64 hex characters long.

common/wallet_tx.c
```diff
@@ -39,6 +43,63 @@ struct command_result *param_wtx(struct command *cmd,
...
+	struct bitcoin_txid **txids = tal_arr(cmd, struct bitcoin_txid*, 0);
+	unsigned int **outnums = tal_arr(cmd, unsigned int*, 0);
+
+	json_for_each_arr(i, curr, tok) {
+		/* A txid is always 32 bytes (64 hex characters) long */
+        const jsmntok_t txid_tok = {JSMN_STRING, curr->start,
+									curr->start + 64, curr->size};
+		/* Pass the `:` separator */
+		const jsmntok_t outnum_tok = {JSMN_STRING, txid_tok.end + 1,
+									curr->end, curr->size};
```
rustyrussell: This code can read past the end of malformed input. 

Prefer something like this:

```
static bool split_tok(const char *buffer, const jsmntok_t *t, char split, jsmntok_t *a, jsmntok_t *b)
{
        const char *p = memchr(buffer + t->start, split, buffer + t->end);

        if (!p)
                 return false;

        *a = *b = *t;
        a->end = p - 1 - buffer;
        b->start = p + 1 - buffer;
        return true;
}
```

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+
+	tx->change = AMOUNT_SAT(0);
+	tx->change_key_index = 0;
+	total_amount = AMOUNT_SAT(0);
+
+	/* The transaction has `tal_count(tx.utxos)` inputs and one output */
+	/* (version + in count + out count + locktime)  (index + value + script length) */
+	weight = 4 * (4 + 1 + 1 + 4) + 4 * (8 + 1 + out_len);
+	for (i = 0; i < tal_count(utxos); i++) {
+        if (!*utxos[i]->blockheight || *utxos[i]->blockheight > maxheight) {
```
rustyrussell: Weird indent?

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+
+	fee_estimate = amount_tx_fee(fee_rate_per_kw, weight);
+
+	if (tx->all_funds || amount_sat_eq(tx->change, AMOUNT_SAT(0))) {
+		tx->amount = total_amount;
+        if (!amount_sat_sub(&tx->amount, tx->amount, fee_estimate))
+			return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction");
+	} else {
+        if (!amount_sat_sub(&tx->change, tx->change, fee_estimate))
```
rustyrussell: Weird indent again...

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+		weight += (8 + 1 + out_len) * 4;
+	}
+
+	fee_estimate = amount_tx_fee(fee_rate_per_kw, weight);
+
+	if (tx->all_funds || amount_sat_eq(tx->change, AMOUNT_SAT(0))) {
+		tx->amount = total_amount;
+        if (!amount_sat_sub(&tx->amount, tx->amount, fee_estimate))
+			return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction");
```
rustyrussell: Maybe give a clue here as to what the fee estimate was?

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+	} else {
+        if (!amount_sat_sub(&tx->change, tx->change, fee_estimate))
+            return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction");
```
rustyrussell: Fee estimate printing here too?

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+	struct amount_sat total_amount, fee_estimate;
+
+	tx->change = AMOUNT_SAT(0);
+	tx->change_key_index = 0;
+	total_amount = AMOUNT_SAT(0);
+
+	/* The transaction has `tal_count(tx.utxos)` inputs and one output */
+	/* (version + in count + out count + locktime)  (index + value + script length) */
+	weight = 4 * (4 + 1 + 1 + 4) + 4 * (8 + 1 + out_len);
+	for (i = 0; i < tal_count(utxos); i++) {
```
rustyrussell: mildly prefer C-99-style `(for size_t i = 0;...` here.

common/wallet_tx.c
```diff
@@ -104,3 +165,62 @@ struct command_result *wtx_select_utxos(struct wallet_tx *tx,
...
+        if (!amount_sat_sub(&tx->amount, tx->amount, fee_estimate))
+			return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction");
+	} else {
+        if (!amount_sat_sub(&tx->change, tx->change, fee_estimate))
+            return command_fail(tx->cmd, FUND_CANNOT_AFFORD,
+								"Cannot afford transaction");
+	}
+	res = check_amount(tx, tx->amount);
+    return res;
```
rustyrussell: indent is weird, but just `return check_amount...` instead of a temporary variable?

rustyrussell:
OK, minor fixes only.  I'm trying to tag rc1, but there are a couple of other merges pending review, so I'll wait (was tempted to tweak it myself, but that's a bad habit!).

common/wallet_tx.c
darosior: That was what I had at first, but I need `const` tokens and didnt want to do multiple cast. Is it preferable to cast ?

darosior:
@rustyrussell I corrected the minor fixes, added the split function (as a helper) and feerate estimate in errors, I also split the first check about parsing the txid and vout in two parts to have more accurate errors.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/035ea5be213c235c2e17ca8b5da867adf789447a

GitHub ordered the commits weirdly, but I checked locally and they're in the expected order...

PR #2707, opened 2019-06-04
Test generate block wait mempool

m-schmoock:
This will add an optional wait_for_mempool parametr to the `generate_block` function in the BitcoinD testsuite. Can help fix flaky tests and reduces code cuplications for all the `wait_for(lambda: len(self.rpc.getrawmempool()) > 0`

The new `wait_for_mempool` parameter can be:

- `True` to wait for 1 transaction in mempool
- `int > 0` to wait for N transactions in the mempool
- `"tx_id_str"` to wait for only one specific transaction ID
- `["tx_id_str_1", "tx_id_str_2", ...]` to wait for N specific transaction IDs

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -320,7 +320,19 @@ class BitcoinD(TailableProc):
         proxy.start()
         return proxy
 
-    def generate_block(self, numblocks=1):
+    # wait_for_mempool can be used to wait for the mempool before generating blocks:
+    # True := wait for at least 1 transation
+    # int > 0 := wait for at least N transactions
+    # 'tx_id' := wait for one transaction id given as a string
+    # ['tx_id1', 'tx_id2'] := wait until all of the specified transaction IDs
+    def generate_block(self, numblocks=1, wait_for_mempool=0):
+        if wait_for_mempool:
+            if isinstance(wait_for_mempool, str):
+                wait_for_mempool = [wait_for_mempool]
+            if isinstance(wait_for_mempool, list):
+                wait_for(lambda: all(txid in self.rpc.getrawmempool() for txid in wait_for_mempool))
+            else:
+                wait_for(lambda: len(self.rpc.getrawmempool()) >= wait_for_mempool)
         # As of 0.16, generate() is removed; use generatetoaddress.
         return self.rpc.generatetoaddress(numblocks, self.rpc.getnewaddress())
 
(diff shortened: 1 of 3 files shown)


tests/utils.py
```diff
@@ -320,7 +320,11 @@ def get_proxy(self):
         proxy.start()
         return proxy
 
-    def generate_block(self, numblocks=1):
+    # wait_for_mempool can be True or a positive integer to wait for the
+    # amount of transactions in the mempool before generating blocks.
+    def generate_block(self, numblocks=1, wait_for_mempool=0):
```
cdecker: We might want to go a step further and not just check for the number of transactions, but also the specific transaction IDs we'd like to include. So this could become 
```suggestion
    def generate_block(self, numblocks=1, must_include=None):
```

And we call it with a list of transaction IDs that we check for.

m-schmoock: okay, will do :D

m-schmoock:
@cdecker like this?

tests/utils.py
```diff
@@ -320,7 +320,19 @@ def get_proxy(self):
         proxy.start()
         return proxy
 
-    def generate_block(self, numblocks=1):
+    # wait_for_mempool can be used to wait for the mempool before generating blocks:
+    # True := wait for at leasst 1 transations
```
niftynei:
```suggestion
    # True := wait for at least 1 transaction
```

tests/utils.py
```diff
@@ -320,7 +320,19 @@ def get_proxy(self):
...
+    # int > 0 := wait for at least N transactions
+    # 'tx_id' := wait for one transaction id given as a string
+    # ['tx_id1', 'tx_id2'] := wait until all of the specified transaction IDs
+    def generate_block(self, numblocks=1, wait_for_mempool=0):
+        if wait_for_mempool:
+            if isinstance(wait_for_mempool, str):
```
SimonVrouwe: Is there any difference between `wait_for_mempool=True` and `wait_for_mempool=1`?

darosior: > Is there any difference between `wait_for_mempool=True` and `wait_for_mempool=1`?

According to what criteria ? If this is about how they will evaluate line 329, they will both pass the condition. Otherwise there is a big difference : it changes the type of the `wait_for_mempool` variable (/object) :
```python
<class 'bool'>
>>> type(1)
<class 'int'>
```

m-schmoock: nope, both will wait for at least one TX

tests/utils.py
m-schmoock: thx :D

rustyrussell:
API seems ridiculously wide (vs Python's "Only one way to do it"), but it's a minor thing we can fix later.

Ack https://github.com/ElementsProject/lightning/commit/6c70001e49c534ad5828cdeb9bd8d028f08cc02e

PR #2712, opened 2019-06-04
newaddr: fix return value for deprecated "address" field.

rustyrussell:
It would always return bech32; fix that, and don't bother printing
it if they use the (new) 'all' parameter.

This API was introduced in 3e67c09d5e6022300d785fbf5be76aeaa756c822,
which means it wasn't in a release so no CHANGELOG entry necessary.

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -339,8 +339,9 @@ static struct command_result *json_newaddr(struct command *cmd,
 
 	response = json_stream_success(cmd);
 	json_object_start(response, NULL);
-	if (deprecated_apis)
-		json_add_string(response, "address", bech32 ? bech32 : p2sh);
+	if (deprecated_apis && *addrtype != ADDR_ALL)
+		json_add_string(response, "address",
+				*addrtype & ADDR_BECH32 ? bech32 : p2sh);
 	if (*addrtype & ADDR_BECH32)
 		json_add_string(response, "bech32", bech32);
 	if (*addrtype & ADDR_P2SH_SEGWIT)
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1402,6 +1402,14 @@ def test_newaddr(node_factory):
     assert both['bech32'].startswith('bcrt1')
 
 
+def test_newaddr_deprecated(node_factory):
+    l1 = node_factory.get_node(options={'allow-deprecated-apis': True})
+    p2sh = l1.rpc.newaddr('p2sh-segwit')
+    assert p2sh['address'].startswith('2')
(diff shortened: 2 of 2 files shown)


wallet/walletrpc.c
```diff
@@ -339,8 +339,9 @@ static struct command_result *json_newaddr(struct command *cmd,
 
 	response = json_stream_success(cmd);
 	json_object_start(response, NULL);
-	if (deprecated_apis)
-		json_add_string(response, "address", bech32 ? bech32 : p2sh);
+	if (deprecated_apis && *addrtype != ADDR_ALL)
+		json_add_string(response, "address",
+				*addrtype & ADDR_BECH32 ? bech32 : p2sh);
```
jb55: nit: for future compatibility with segwit v1 addresses we might want to call this ADDR_P2WPKH instead of ADDR_BECH32 which would become ambiguous. happy to defer that to a future PR though.

rustyrussell: True, though the caller should handle any bech32 address return, really.  If we end up with some weird addresses in future, that shouldn't stop the caller from paying to them, as long as they're bech32 encoded segwit addresses.

ZmnSCPxj: Our own withdraw code rejects future versions of SegWit (currently only v0 is supported). Should we change that? https://github.com/ElementsProject/lightning/blob/fba050e8918ad630ceba3c4e69b2dac3b6bb1fc1/lightningd/jsonrpc.c#L1023

niftynei: +1 for ensuring future compatibility of address formats. I haven't had a chance to look at the taproot/tapscript proposals -- do they include a new address type as well?

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2712/commits/0a6a5d686324899a191d1b43a0225e95b62740c3

wallet/walletrpc.c
ZmnSCPxj: It uses a new SegWit version, encoded as the first quintet of the bech32. SegWit v0 starts `bc1q`, SegWit v1 starts `bc1p`.

jb55: my point was bech32 is ambiguous, it's like saying "base58 addresses". which one does our api give you? It might not be obvious. bech32 could mean p2wsh, p2wpkh, or p2pk segwit v1 outputs. Like @rustyrussell said the caller might not care, just give me any bech32 address, so if that's the api we want to expose then that's fine.

I personally might care, since I'm not sure I'm comfortable with p2pk addresses yet, but perhaps most people would not.

PR #2713, opened 2019-06-05
Withdraw split

rustyrussell:
(Built on top of #2712 which touches similar files)

With @niftynei working on split funding in #2672 I realized our own wallet can't do this.  I stole one of her patches, and made split payments work.

This will let us rebase #2672 and then turn fundchannel into a plugin which does the two-step to fund the channel, removing much redundant code.

@cdecker Please review final patch!


ZmnSCPxj:
> turn fundchannel into a plugin which does the two-step to fund the channel, removing much redundant code.

Should we consider how a "real" hardware signing module would work?

I imagine "real" hsm would be similar to a Trezor or Ledger, where most tx sends would cause the device to open its own display and require  user confirmation on device hardware.

However, for channel opens, we can provide the initial commitment transaction and signature, with the hsm validating that the commitment pays to an address it controls, without requiring independent confirmation.

I suppose it depends on how vital it would be to habe a "real" hardware signer.

rustyrussell:
I think we'd allow destination to be an array of address & amount objects (one of which could be "all").

The other extension is to add the utxo arg for utxo selection, then finally to allow foreign utxos, with the signatures provided in the txsend step.

Or we could go full pbst and miniscript eventually...

ZmnSCPxj:
I have a wip branch containing commits you could use for multi destination withdraw  https://github.com/ZmnSCPxj/lightning/tree/wip/multiwithdraw?files=1

niftynei:
Test errored out because it ran out of time. Moving towards PSBT for multi anything is the right move imo.

wallet/walletrpc.c
```diff
@@ -146,40 +187,168 @@ static struct command_result *json_withdraw(struct command *cmd,
...
 
-	u8 *msg = towire_hsm_sign_withdrawal(cmd,
-					     withdraw->wtx->amount,
-					     withdraw->wtx->change,
-					     withdraw->wtx->change_key_index,
-					     withdraw->destination,
-					     withdraw->wtx->utxos);
+	(*utx)->tx = withdraw_tx(*utx, (*utx)->wtx->utxos,
+				 (*utx)->destination, (*utx)->wtx->amount,
+				 changekey, (*utx)->wtx->change,
```
jb55: looks like changekey could be uninitialized here if wtx->change is 0?

```
wallet/walletrpc.c: In function ‘json_prepare_tx’:
wallet/walletrpc.c:203:15: error: ‘changekey’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  (*utx)->tx = withdraw_tx(*utx, (*utx)->wtx->utxos,
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      (*utx)->destination, (*utx)->wtx->amount,
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      changekey, (*utx)->wtx->change,
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      cmd->ld->wallet->bip32_base,
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      &(*utx)->change_outnum);
      ~~~~~~~~~~~~~~~~~~~~~~~
```

cdecker: That's quite alright, it's not actually used unless we have a non-zero change amount. But still it's a rather annoying warning that we should address.

Issue #2305, opened 2019-01-30
Create plugin to opt-in report unusual events


rustyrussell:

trueptolemy:
If there need a notification for unusual events?

ZmnSCPxj:
As we already have a notification for unusual messages, should we still write an actual plugin to report this elsewhere? Or is the notification enough?

rustyrussell:
AI cleanup of old issues.  If wrong, please open a new issue.


Triage decision: close_resolved (confidence 0.92)

Evidence:
- PR #2715 was merged on 2019-06-07, implementing a 'warning' notification type that addresses the core request
- The PR implements notifications for LOG_BROKEN and LOG_UNUSUAL level logs, which directly fulfills the issue's goal of reporting unusual events
- The notification includes level (warn/error), timestamp, source, and log context - providing comprehensive event reporting
- The implementation is plugin-based and opt-in as requested in the issue title
- Issue was created 2019-01-30 and PR #2715 was merged 2019-06-07, showing direct response to the feature request
- No updates to the issue since 2021-01-08, suggesting the implementation has been stable and sufficient

PR #2715, opened 2019-06-05
plugin: A new notification type 'warning'

trueptolemy:
This notification bases on `LOG_BROKEN` and `LOG_UNUSUAL` level log. Hope it can help to deal with #2305 .

## Introduction
A notification for topic `warning` is sent every time a new `BROKEN`/`UNUSUAL` level(in plugins, we use `error`/`warn`) log generated, which means an unusual/borken thing happens, such as channel failed,
message resolving failed...

```json
{
	"warning": {
	"level": "warn",
	"time": "1559743608.565342521",
	"source": "lightningd(17652): 0821f80652fb840239df8dc99205792bba2e559a05469915804c08420230e23c7c chan #7854:",
	"log": "Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: sent ERROR bad reestablish dataloss msg"
  }
}
```
1. `level` is `warn` or `error`:
`warn` means something seems bad happened and it's under control, but we'd better check it;
`error` means something extremely bad is out of control, and it may lead to crash;
2. `time` is the second since epoch;
3. `source`, in fact, is the `prefix` of the log_entry. It means where the event happened, it may have the following forms:  
`<node_id> chan #<db_id_of_channel>:`, `lightningd(<lightningd_pid>):`, `plugin-<plugin_name>:`, `<daemon_name>(<daemon_pid>):`, `jsonrpc:`, `jcon fd <error_fd_to_jsonrpc>:`, `plugin-manager`;
4. `log` is the context of the original log entry.

### Note:
1. The main code uses `UNUSUAL`/`BROKEN`, and plugin module uses `warn`/`error`, considering the consistency with plugin, warning choose `warn`/`error`. But users who use c-lightning with plugins may want to `getlog` with specified level when receive warning. It's the duty for plugin dev to turn `warn`/`error` into `UNUSUAL`/`BROKEN` and present it to the users, or pass it directly to `getlog`.
2. About time, `json_log()` in `log` module uses the Relative Time, from the time when `log_book` inited to the time when this event happend. But I consider the `UNUSUAL`/`BROKEN` event is rare, and it is very likely to happen after running for a long time, so for users, they will pay more attention to Absolute Time.

## Related Changes
1. Remove the definitions of `log`, `log_book`, `log_entry` from `log.c` to `log.h`, then they can be used in warning declaration and definition.
2. Add a pointer to `struct lightningd` in `struct log_book`. This may affect the independence of the `log` module, but storing a pointer to `ld` is more direct;
3. Add a `bool` type parameter in `log_()` and `lov()`, this `bool` flag indicates if now we are in copy process.
The process of copying `log_book` of every peer to the `log_book` of `ld` is usually included in `log_()` and `lov()`, and it may lead to repeated `warning` notification. So a `bool`, which explicitly indicates if the `warning` notification is disabled during this call, is necessary.

I named this notification as `unusual_event`, but after many hesitations, I finally choose `warning`, because it looks more visually impactful.

Thanks for cdecker, jb55 and m-schmoock, the discussion of this notification on IRC helpd me :)

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -4,7 +4,6 @@
 #include <ccan/array_size/array_size.h>
 #include <ccan/err/err.h>
 #include <ccan/io/io.h>
-#include <ccan/list/list.h>
 #include <ccan/opt/opt.h>
 #include <ccan/read_write_all/read_write_all.h>
 #include <ccan/str/hex/hex.h>
@@ -19,6 +18,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
+#include <lightningd/json.h>
 #include <lightningd/jsonrpc.h>
 #include <lightningd/lightningd.h>
 #include <lightningd/options.h>
@@ -30,39 +30,6 @@
 
 /* Once we're up and running, this is set up. */
 struct log *crashlog;
-
-struct log_entry {
-	struct list_node list;
-	struct timeabs time;
-	enum log_level level;
-	unsigned int skipped;
-	const char *prefix;
-	char *log;
-	/* Iff LOG_IO */
-	const u8 *io;
-};
-
-struct log_book {
-	size_t mem_used;
-	size_t max_mem;
-	void (*print)(const char *prefix,
-		      enum log_level level,
-		      bool continued,
-		      const struct timeabs *time,
-		      const char *str,
-		      const u8 *io, size_t io_len,
-		      void *arg);
-	void *print_arg;
-	enum log_level print_level;
-	struct timeabs init_time;
-
-	struct list_head log;
-};
-
-struct log {
-	struct log_book *lr;
-	const char *prefix;
-};
 
 static void log_to_file(const char *prefix,
 			enum log_level level,
@@ -662,17 +629,6 @@ static void add_skipped(struct log_info *info)
 		json_object_end(info->response);
 		info->num_skipped = 0;
 	}
-}
-
-static void json_add_time(struct json_stream *result, const char *fieldname,
-			  struct timespec ts)
-{
-	char timebuf[100];
-
-	snprintf(timebuf, sizeof(timebuf), "%lu.%09u",
-		(unsigned long)ts.tv_sec,
-		(unsigned)ts.tv_nsec);
-	json_add_string(result, fieldname, timebuf);
 }
 
 static void log_to_json(unsigned int skipped,
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -21,6 +21,7 @@
 #include <lightningd/json.h>
 #include <lightningd/jsonrpc.h>
 #include <lightningd/lightningd.h>
+#include <lightningd/notification.h>
 #include <lightningd/options.h>
 #include <signal.h>
 #include <stdio.h>
@@ -102,7 +103,7 @@ static size_t prune_log(struct log_book *log)
 	return deleted;
 }
 
-struct log_book *new_log_book(size_t max_mem,
+struct log_book *new_log_book(struct lightningd *ld, size_t max_mem,
 			      enum log_level printlevel)
 {
 	struct log_book *lr = tal_linkable(tal(NULL, struct log_book));
@@ -114,6 +115,7 @@ struct log_book *new_log_book(size_t max_mem,
 	lr->print = log_to_stdout;
 	lr->print_level = printlevel;
 	lr->init_time = time_now();
+	lr->ld = ld;
 	list_head_init(&lr->log);
 
 	return lr;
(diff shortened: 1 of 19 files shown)


wallet/test/run-db.c
```diff
@@ -3,7 +3,7 @@
 static void db_test_fatal(const char *fmt, ...);
 #define db_fatal db_test_fatal
 
-static void db_log_(struct log *log UNUSED, enum log_level level UNUSED, const char *fmt UNUSED, ...)
+static void db_log_(struct log *log UNUSED, enum log_level level UNUSED, bool notifer UNUSED, const char *fmt UNUSED, ...)
```
rustyrussell: This belongs in the next commit I think?

wallet/test/run-wallet.c
```diff
@@ -4,7 +4,7 @@ static void wallet_test_fatal(const char *fmt, ...);
 #define db_fatal wallet_test_fatal
 #include "test_utils.h"
 
-static void db_log_(struct log *log UNUSED, enum log_level level UNUSED, const char *fmt UNUSED, ...)
+static void db_log_(struct log *log UNUSED, enum log_level level UNUSED, bool notifer UNUSED, const char *fmt UNUSED, ...)
```
rustyrussell: This belongs in the next commit I think?

lightningd/log.c
```diff
@@ -243,6 +246,12 @@ void logv(struct log *log, enum log_level level, const char *fmt, va_list ap)
 	maybe_print(log, l, 0);
 
 	add_entry(log, l);
+
+	if(notifer) {
```
rustyrussell: Nit: space after `if` before `(`

lightningd/log.c
```diff
@@ -226,7 +228,8 @@ static void maybe_print(const struct log *log, const struct log_entry *l,
 			       l->io, tal_bytelen(l->io), log->lr->print_arg);
 }
 
-void logv(struct log *log, enum log_level level, const char *fmt, va_list ap)
+void logv(struct log *log, enum log_level level, bool notifer,
```
rustyrussell: `tell_notifier` is a better name for the bool I think?

lightningd/log.c
```diff
@@ -243,6 +246,12 @@ void logv(struct log *log, enum log_level level, const char *fmt, va_list ap)
 	maybe_print(log, l, 0);
 
 	add_entry(log, l);
+
+	if(notifer) {
+		if(level >= LOG_UNUSUAL)
+			notify_warning(log->lr->ld, l);
```
rustyrussell: Why not put this logic in the caller?   ie. have the log_unusual and log_broken macros set `call_notifier` and the others not?

tests/test_plugin.py
```diff
@@ -107,27 +107,9 @@ def test_plugin_disable(node_factory):
         n.rpc.hello(name='Sun')
 
 
-def test_plugin_hook(node_factory, executor):
```
rustyrussell: This patch is messed up?  Seems to remove existing plugin tests....

rustyrussell:
I'm going to try to sneak this into 0.7.1, which I was hoping to do an rc for today...

lightningd/log.c
trueptolemy: Thank you! I'll take care!

lightningd/log.c
trueptolemy: nice name!

tests/test_plugin.py
trueptolemy: This is my last rebase I made a mistake...

lightningd/log.c
trueptolemy: Thank you! Good idea.

trueptolemy:
> I'm going to try to sneak this into 0.7.1, which I was hoping to do an rc for today...

Sorry...I am late...now I've packed up my accident scene.

trueptolemy:
1. Rebased and organized the commit content;
2. Added back the wrong deletion of tests;
3. Rename `notifer` to `call_notifier`(correct spelling), and directly set `call_notifier` as `TRUE` for `LOG_UNUSUAL`/`LOG_BROKEN`, as `FALSE` for `LOG_DEBUG`/`LOG_INFO`.

trueptolemy:
test_pay::test_pay_direct always fail because of timeout, now I wait for its log...

BTW, @rustyrussell do you think if we need other notifications now?

Like the notification for forward_payment fulfilled or failed(I think it can work well with htlc_accept_hook), 
and the notification for pay_fail or oay_succeed(after we have waitsendpay, if we need this notification type), 
and the notification for channel_lock/channel_announced, 
or other type.

rustyrussell:
Please wait for post 0.7.1: these awesome patches are delaying the release 😆

PR #2716, opened 2019-06-06
configure: fix --reconfigure, improve style, set defaults in one place.

rustyrussell:
I noticed that DEVELOPER was being reset to 0 by ./configure --reconfigure;
that's because we set the defaults first, then --reconfigure would only
override *unset* vars.  We now set up defaults last.

We unify all our "find the default" functions, to neaten them; in
particular find_pytest and our open-coded valgrind testing function.

We can figure out whether valgrind works using /bin/true instead of waiting
until configurator is built.

We're also more careful with ${FOO-default} vs ${FOO:-default}; the former
does not replace if FOO is set to the empty string, which is possible for
some of our settings (COPTFLAGS, CDEBUGFLAGS in particular).

### DIFF
--- a/configure
+++ b/configure
@@ -3,16 +3,6 @@
 
 set -e
 
-# Default values, loaded from environment or canned.
-PREFIX=${PREFIX:-/usr/local}
-CC=${CC:-cc}
-CDEBUGFLAGS=${CDEBUGFLAGS:--std=gnu11 -g -fstack-protector}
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2716/commits/84bc58c54e606393d0d8e11c141762604cb8c156

Issue #2701, opened 2019-06-03
c-lightning rejects its own no-changes commitment


pm47:
### Issue and Steps to Reproduce

c-lightning sends the same `update_fee` several times within a minute (is that normal?), with a `commit_sig` in-between. After the first `update_fee`, there are effectively no changes. ~~Eclair now accepts `commit_sig` with no changes for compatibility reasons with lnd, so it accepts the identical `commit_sig`.~~ But c-lightning apparently only accepts one single identical `commit_sig`, so it ends up rejecting eclair's `commit_sig` in response to its own identical `commit_sig`. 🙃

In this scenario all the `update_fee` were identical and all the `commit_sig` have no changes:

```
 eclair           c-lightning
   |                   |
   |  (idle, no htlc)  |
   |                   |
   |<------ fee -------|
   |<------ sig -------|
   |------ rev ------->|
   |------- sig ------>|
   |<------ fee -------| <- fee2 == fee1
   |<------ sig -------|
   |------ rev ------->|
   |<----- rev --------|
   |------- sig ------>|
   |<------ fee  ------| <- fee3 == fee2 == fee1
   |<------ sig -------|
   |------ rev ------->|
   |<----- rev --------|
   |------- sig ------>|
   |<------ err -------| <- commit_sig with no changes (again!)
   |        ...        |
```

I think we should probably forbid sending several times the same `update_fee` in a row ~~, but then again, we already are supposed to forbid no-changes commitments.🙃🙃 This makes me want to remove the tolerance that we added just for lnd.~~

```
2019-05-31 14:25:58,194 INFO IN msg=UpdateFee(96c6954710749862a630cf1d1567131f97df190af5225ac80a3331e29d07298c,52000)
2019-05-31 14:26:00,419 INFO IN msg=CommitSig(96c6954710749862a630cf1d1567131f97df190af5225ac80a3331e29d07298c,7ae4cbf0a0da7518c178b39d854da341a02fbe52f125303a15750bf0f05924a57a0e3be761e711290c710f057fb3990936ee6959e94fd30134dce4d38913d1f5,List())
[...]

PR #2718, opened 2019-06-06
channeld: don't send feerate spam if we can't set it as high as we want.

rustyrussell:
@pm47 gave a great bug report showing c-lightning sending the same UPDATE_FEE over and over, with the final surprise result being that we blamed the peer for sending us multiple empty commits!

The spam is caused by us checking "are we at the desired feerate?" but then if we can't afford the desired feerate, setting the feerate we can afford, even though it's a duplicate.  Doing the feerate cap before we test if it's what we have already eliminates this.

But the empty commits was harder to find: it's caused by a heuristic in channel_rcvd_revoke_and_ack:

```
	/* For funder, ack also means time to apply new feerate locally. */
	if (channel->funder == LOCAL &&
	    (channel->view[LOCAL].feerate_per_kw
	     != channel->view[REMOTE].feerate_per_kw)) {
		status_trace("Applying feerate %u to LOCAL (was %u)",
			     channel->view[REMOTE].feerate_per_kw,
			     channel->view[LOCAL].feerate_per_kw);
		channel->view[LOCAL].feerate_per_kw
			= channel->view[REMOTE].feerate_per_kw;
		channel->changes_pending[LOCAL] = true;
	}
```

We assume we never send duplicates, so we detect an otherwise-empty change using the difference in feerates.  If we don't set this flag, we will get upset if we receive a commitment_signed since we consider there to be no changes to commit.

This is actually hard to test: the previous commit adds a test which spams update_fee and doesn't trigger this bug, because both sides use the same "there's nothing outstanding" logic.

Fixes: #2701

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -1133,9 +1133,7 @@ static void send_commit(struct peer *peer)
 	}
 
 	/* If we wanted to update fees, do it now. */
-	if (peer->channel->funder == LOCAL
-	    && peer->desired_feerate != channel_feerate(peer->channel, REMOTE)) {
-		u8 *msg;
+	if (peer->channel->funder == LOCAL) {
 		u32 feerate, max = approx_max_feerate(peer->channel);
 
 		feerate = peer->desired_feerate;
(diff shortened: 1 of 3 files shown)


channeld/channeld.c
```diff
@@ -1145,14 +1143,19 @@ static void send_commit(struct peer *peer)
...
-			status_failed(STATUS_FAIL_INTERNAL_ERROR,
-				      "Could not afford feerate %u"
-				      " (vs max %u)",
-				      feerate, max);
+		if (feerate != channel_feerate(peer->channel, REMOTE)) {
+			u8 *msg;
 
-		msg = towire_update_fee(NULL, &peer->channel_id, feerate);
-		sync_crypto_write(peer->pps, take(msg));
+			if (!channel_update_feerate(peer->channel, feerate))
```
ZmnSCPxj: Is my reading right, that if this condition fails, we are unable to afford the feerate but we still send `update_fee`?

rustyrussell: If that fails, we abort because something is badly wrong.

Patch is hard to read, resulting code is easier:


```
		u32 feerate, max = approx_max_feerate(peer->channel);

		feerate = peer->desired_feerate;

		/* FIXME: We should avoid adding HTLCs until we can meet this
		 * feerate! */
		if (feerate > max)
			feerate = max;

		if (feerate != channel_feerate(peer->channel, REMOTE)) {
			u8 *msg;

			if (!channel_update_feerate(peer->channel, feerate))
					      "Could not afford feerate %u"
					      " (vs max %u)",
					      feerate, max);

			msg = towire_update_fee(NULL, &peer->channel_id,
						feerate);
			sync_crypto_write(peer->pps, take(msg));
		}
```

cdecker:
Rebased on top of `master` to get rid of CHANGELOG conflict.

ACK https://github.com/ElementsProject/lightning/pull/2718/commits/afd1c792307d210de0892faa0c7e518e122287a3

PR #2721, opened 2019-06-06
plugin: Fix how we send the shared_secret to the plugin in htlc_accepted.

cdecker:
~Found this while trying to do some magic on top of `shared_secret`, turns out we were not keeping the secret around long enough, so now we just stash it in the `htlc_accepted_payload`.~

I had confused pointer and pointee :frowning_face:

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -721,7 +721,7 @@ static void htlc_accepted_hook_serialize(struct htlc_accepted_hook_payload *p,
 	}
 
 	json_add_hex_talarr(s, "next_onion", p->next_onion);
-	json_add_hex(s, "shared_secret", &hin->shared_secret, sizeof(hin->shared_secret));
+	json_add_secret(s, "shared_secret", hin->shared_secret);
 	json_object_end(s);
 
 	json_object_start(s, "htlc");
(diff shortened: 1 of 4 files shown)


rustyrussell:
Why did valgrind not catch this BTW?

cdecker:
Sorry, my bad, the usual confusion with pointer and pointee :frowning_face:

lightningd/peer_htlcs.c
```diff
@@ -721,7 +721,7 @@ static void htlc_accepted_hook_serialize(struct htlc_accepted_hook_payload *p,
 	}
 
 	json_add_hex_talarr(s, "next_onion", p->next_onion);
-	json_add_hex(s, "shared_secret", &hin->shared_secret, sizeof(hin->shared_secret));
+	json_add_hex(s, "shared_secret", hin->shared_secret, sizeof(struct secret));
```
rustyrussell: Gah, this is why writing helpers is important!

But yes, as @ZmnSCPxj points out, the commit msg is now nonsensical.

ZmnSCPxj: So new `json_add_secret`?

cdecker:
Right, I shouldn't rush these things, my bad.
 - Added a wrapper to add `struct secret` to a JSON stream
 - Using that wrapper now
 - Fixed commit messages.

Sorry for the confusion :slightly_frowning_face:

PR #2722, opened 2019-06-07
txfilter: Migrate the list of owned scriptPubKeys into a hashset

cdecker:
I was bumping against some blocksync performance issues with 12k+ keys, 24k+
scriptpubkeys being checked against, and migrating that list to a hashset is
an easy fix to shave off 99% of the time to process a block.


ZmnSCPxj:
Ack  979c4dd

Issue #2719, opened 2019-06-06
Failure to parse a pending_node_announcement when loading the gossip_store


thestick613:
I've been having trouble with this node. I have two nodes, they had a channel between them. I was able to invoice from A and pay from B, but every invoice generated from B failed to get payed from A. They were small invoices, like 2000 sat each.
I investigated the channel status using the output i learned from #2709, and the healthy node changed the values of msatoshi_to_us, their_channel_reserve_satoshis and our_channel_reserve_satoshis after every payment, but the 'sick' one didn't. I decided to close all the channels of the bad node, because i was running on mainnet funds. After i closed the last channel, the node crashed:

```
lightning_gossipd: pending node_announcement 01014db000711e286148c285a8565ecafcf24a8955ad12d96d338f9cf67fb708fa3225fee739a37f1eddecaf041b79e60f1f7790630672e921be58e3eaf71749fc4900005cc1a7d9028b6dc171f6034
67690f799b6a0d52b1f5a5b8977b1453440016ea7917f53925f028b6d686f73743463727970746f2e636f6d000000000000000000000000000000000000070150f1d8cc2607 malformed? (version v0.7.0-490-gbbedd38)
0x410fc4 send_backtrace
        common/daemon.c:40
0x413a54 status_failed
        common/status.c:192
0x40970f process_pending_node_announcement
        gossipd/routing.c:1315
0x409b32 routing_add_channel_update
        gossipd/routing.c:1891
0x406652 gossip_store_load
        gossipd/gossip_store.c:582
0x40e36d gossip_init
        gossipd/gossipd.c:1862
0x40f2da recv_req
        gossipd/gossipd.c:2702
0x41111a handle_read
        common/daemon_conn.c:31
0x427521 next_plan
        ccan/ccan/io/io.c:59
0x4279e1 do_plan
        ccan/ccan/io/io.c:407
0x427a0c io_ready
        ccan/ccan/io/io.c:417
0x4290b5 io_loop
        ccan/ccan/io/poll.c:445
0x40f515 main
        gossipd/gossipd.c:2827
0x7faae2d2b82f ???
        ???:0
0x402b18 ???
        ???:0
0xffffffffffffffff ???
        ???:0
2019-06-06T07:46:23.887Z lightning_gossipd(15217): gossip_store node_announcement 2480389 replaces 2480389!
[...]

PR #2723, opened 2019-06-08
gossip: bump version to remove lingering issues with master.

rustyrussell:
There were several gossip breakages in master; bumping version means
upgrades get a clean store (not just those upgrading from stable version).

Fixes: #2719


cdecker:
The test vector in `test_gossip_store_load` needs to be bumped as well.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2723/commits/26625d869f8824b11583d30e4363a224c5281fec

PR #2727, opened 2019-06-10
Make human help human friendly

abitfan:
HELLO FELLOW HUMANS!

THE FOLLOWING DISPLAY IS MORE READABLE FOR US HUMANS
PERHAPS DUE TO THE RESSEMBLENCE TO bitcoin-cli help OUTPUT.
CHECK IT OUT!

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -188,9 +188,9 @@ static void human_help(char *buffer, const jsmntok_t *result, bool has_command) 
 
 		command = json_get_member(buffer, help[i], "command");
 		desc = json_get_member(buffer, help[i], "description");
-		printf("%.*s\n",
+		printf("%.*s",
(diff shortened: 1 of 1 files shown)


m-schmoock:
Capslock is totally uncool. Makes things less humanly readable ;)

darosior:
What is the rationale ? Without line break the help is far less readable..

abitfan:
Generally when one issues help is looking to locate a command (and only after that
the description is of interest).
As it is it wastes a lot of screen space making even more difficult to navigate.
It may be just a matter of personal taste but I even think listing just the 
commands the same way as bitcoin-cli help is preferred for efficiency.

rustyrussell:
Well, if we just printed the synopsis and not a description, it might be OK, but this is a pig's salad.

Though now I look at it, we should put the extra '\n' *before* the section headings, not after!

PR #2728, opened 2019-06-11
ccan: update to faster and correct crc32c implementation.

rustyrussell:
I decided to try a faster implementation, only to find our crc32c was
not correct!  Ouch.

I removed the crc32c functions from ccan/crc, and added a new crc32c
module which has the Mark Adler x86-64-optimized variants.

We bump gossip_store version again, since csums have changed.


cdecker:
ACK e368300f19e7b9216dcbb0219979e1895aab2f4f

PR #2729, opened 2019-06-12
ASAN fixes

rustyrussell:
Reported-by: @niftynei 

None of these are unconstrained memory leaks, but it's nice to clean up on exit.

We could override the htable alloc routines to use tal off the htable, but that actually means each htable would have to be tal'ocated itself, which they currently are not (they're inside structures).  So we do this instead.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -16,6 +16,7 @@
 #include <common/timeout.h>
 #include <dirent.h>
 #include <errno.h>
+#include <lightningd/io_loop_with_timers.h>
 #include <lightningd/json.h>
 #include <lightningd/lightningd.h>
 #include <lightningd/notification.h>
@@ -77,7 +78,6 @@ struct plugins {
 	struct log *log;
 	struct log_book *log_book;
 
-	struct timers timers;
 	struct lightningd *ld;
 };
 
@@ -107,7 +107,6 @@ struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
 	list_head_init(&p->plugins);
 	p->log_book = log_book;
 	p->log = new_log(p, log_book, "plugin-manager");
-	timers_init(&p->timers, time_mono());
 	p->ld = ld;
 	return p;
 }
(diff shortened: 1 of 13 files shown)


ZmnSCPxj:
ACK  75f41a2

rustyrussell:
> Why not unify with `timers` in `struct lightningd`? And use `io_loop_with_timers`?

I believe we could, TBH.  And since this isn't going to go in 0.7.1, I will fix it.  Thanks!

lightningd/plugin.c
```diff
@@ -1009,18 +1007,15 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
...
 		else {
 			p->timeout_timer
-				= new_reltimer(&plugins->timers, p,
+				= new_reltimer(&plugins->ld->timers, p,
 					       time_from_sec(PLUGIN_MANIFEST_TIMEOUT),
 					       plugin_manifest_timeout, p);
 		}
 		tal_free(cmd);
 	}
 	while (plugins->pending_manifests > 0) {
```
ZmnSCPxj: The `while` loop can be omitted entirely, we should probably `assert(plugins->pending_manifests == 0);` after the `io_loop_with_timers`.

niftynei:
tACK https://github.com/ElementsProject/lightning/pull/2729/commits/d3a41bfc6a61bedebe1f788b064aa284a55482de

spot checked that this does indeed fix ASAN errors on `test_funding_change`.

lightningd/plugin.c
ZmnSCPxj: Or replace the `while` with an `if`.

rustyrussell:
Reworked timer cleanup (first patch), since my naive code made everything use-after-free...

lightningd/plugin.c
rustyrussell: Ack, good catch.

cdecker:
Not merging since we are in the 0.7.1 release window, but will do once we have tagged `v0.7.1` :-)

PR #2734, opened 2019-06-12
openingd: in handle_peer_in fix missing `default` case in switch

SimonVrouwe:
Unless we are just _not interested_ in incoming gossip or errors from peers in _openingd_ stadium, I think this  should be correct.

Found this when doing tests with `devtools/gossipwith` and It explains the many lines like
`chan #1172: Peer connection lost` in my logs.

**edit**: Not sure, but I suspect `gossipwith` did work so far because it is only used to receive gossip?


SimonVrouwe:
Hmm sorry, was false alarm. I thought my debugger couldn't reach that default case, must have been something else. But a `default:` there wouldn't harm the clarity does it?

Issue #2733, opened 2019-06-12
crash: nullpointer on iteration setchannelfee 'all' 


m-schmoock:
### Issue
I discovered that the following loop iteration over all active channels may cause some nullpointer if some channels are in a certain state. Since we are close to `v0.7.1` release, I wanted to bring that up quickly. I have not looked into it yet, I believe that maybe the closed channel with `03ce648061614b971a3a990e62cb5f782ac94e9576f2311256b2985bc05b5033e0` which only has a `"status" : ["Loaded from database"]` for unknown reasons may cause this or maybe the peer withouth channels `02059220fa2153fb6d7768d7d67c6b86f441f83b7046aea492f2018f7b800edfa9`. See `listpeers` output further down below.

`lightningd/peer_control.c +1451`
```C
	/* If the users requested 'all' channels we need to iterate */
	if (channel == NULL) {
		list_for_each(&cmd->ld->peers, peer, list) {
			list_for_each(&peer->channels, channel, list) {
				channel = peer_active_channel(peer);
				if (!channel)
					continue;
				if (channel->state != CHANNELD_NORMAL &&
					channel->state != CHANNELD_AWAITING_LOCKIN)
					continue;
				set_channel_fees(cmd, channel, *base, *ppm, response);
			}
		}
```

In particular the loop generator of this line causes it, which is why the stack contains `(null):0 ((null)) 0x7f9bdb11d8af`:
```C
list_for_each(&peer->channels, channel, list) {
 ...
}
```
### Stacktrace
```
2019-06-10T11:17:35.048Z lightningd(3430): FATAL SIGNAL 11 (version v0.7.0-427-g7961a81)
2019-06-10T11:17:35.048Z lightningd(3430): backtrace: common/daemon.c:45 (send_backtrace) 0x558eb71ad15c
2019-06-10T11:17:35.048Z lightningd(3430): backtrace: common/daemon.c:53 (crashdump) 0x558eb71ad1ac
2019-06-10T11:17:35.048Z lightningd(3430): backtrace: (null):0 ((null)) 0x7f9bdb11d8af
2019-06-10T11:17:35.048Z lightningd(3430): backtrace: lightningd/peer_control.c:1451 (json_setchannelfee) 0x558eb719bbfa
2019-06-10T11:17:35.048Z lightningd(3430): backtrace: lightningd/jsonrpc.c:641 (parse_request) 0x558eb7187dca
[...]

PR #2736, opened 2019-06-12
lightningd: fix bogus channel iteration on setchannelfee

rustyrussell:
Setting channel to NULL then iterating doesn't work!

Fixes: #2733

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1537,15 +1537,13 @@ static struct command_result *json_setchannelfee(struct command *cmd,
 	/* If the users requested 'all' channels we need to iterate */
 	if (channel == NULL) {
 		list_for_each(&cmd->ld->peers, peer, list) {
-			list_for_each(&peer->channels, channel, list) {
-				channel = peer_active_channel(peer);
-				if (!channel)
-					continue;
-				if (channel->state != CHANNELD_NORMAL &&
-					channel->state != CHANNELD_AWAITING_LOCKIN)
-					continue;
-				set_channel_fees(cmd, channel, *base, *ppm, response);
-			}
+			channel = peer_active_channel(peer);
+			if (!channel)
+				continue;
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
That is not the behavior I would ecpect from `list_for_each`...

m-schmoock:
@rustyrussell 
Hm, yes sorry for my broken code, in retrospect it does look wrong :D

I think the nullpointer was caused as we did set the iterator pointer
in the loop itself: (https://github.com/ElementsProject/lightning/pull/2736/files#diff-b34700033c5cd47ee522323329a7a789L1541) `channel = peer_active_channel(peer);`.

This also explains why it does work most of the times and also `tests/test_pay.py::test_setchannelfee_all` works, there is no test with a peer with more than one channel.

@ZmnSCPxj 
> That is not the behavior I would expect from list_for_each...

Correct. `list_for_each` should and does work with an initial `NULL` pointer, as noted above, I think rustys diagnose was not precise, the problem was cause by the fact that we set the pointer again in the (useless) loop.

rustyrussell:
If you open code a single linked list it's clearer:

```C
for (p = head; p; p = p->next) {
    p = find_something();
}
```

You're messing with the iterator, of course it will crash!

ZmnSCPxj:
Then I think the "correct" solution should have been:

    list_for_each (&peer->channels, channel, list) {
       if (channel->state != CHANNELD_NORMAL && channel->state != CHANNELD_AWAITING_LOCKIN)
           continue;
       set_channel_fees(cmd, channel, *base, *ppm, response);
    }

This is resilient against the possible future that we will have more than one normal channel per peer (assuming we plan to have such a future...).

Other parts of the code already use iteration over `&peer->channels`.

m-schmoock:
@ZmnSCPxj yes, thats correct, but `peer_active_channel` will have to be removed then, so this and other occurrences will not be missed.

PR #2738, opened 2019-06-13
common/gossip_store: handle short reads.

rustyrussell:
This happened on Travis, and the gossip_store was a suspicious 4096
bytes long.  This implies they're using some non-atomic filesystem
(gossipd always does atomic writes to gossip_store), but if they are,
others surely are too.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2738/commits/c706bdafc0fdebc814af409bc23fda4bec2f250e

PR #2739, opened 2019-06-13
pytest: fail tests if we get any LOG_BROKEN level messages, unless flagged

rustyrussell:
And clean up some dev ones which actually happen (mainly by calling
channel_fail_permanent which logs UNUSUAL, rather than
channel_internal_error which logs BROKEN).

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -32,6 +32,24 @@
 /* Once we're up and running, this is set up. */
 struct log *crashlog;
 
+static const char *level_prefix(enum log_level level)
+{
+	switch (level) {
+	case LOG_IO_OUT:
+	case LOG_IO_IN:
+		return "IO";
+	case LOG_DBG:
+		return "DEBUG";
+	case LOG_INFORM:
+		return "INFO";
+	case LOG_UNUSUAL:
+		return "UNUSUAL";
+	case LOG_BROKEN:
+		return "**BROKEN**";
+	}
+	abort();
+}
+
 static void log_to_file(const char *prefix,
 			enum log_level level,
 			bool continued,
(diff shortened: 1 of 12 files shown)


lightningd/log.c
```diff
@@ -53,7 +53,7 @@ static void log_to_file(const char *prefix,
 			iso8601_s, prefix, str, dir, hex);
 		tal_free(hex);
 	} else 	if (!continued) {
-		fprintf(logf, "%s %s %s\n", iso8601_s, prefix, str);
+		fprintf(logf, "%s %s %s%s\n", iso8601_s, prefix, level == LOG_BROKEN ? "**BROKEN** " : "", str);
```
cdecker: I'd prefer actually just prefixing all log lines with their respective log level rather than making this special case for `BROKEN`.

tests/test_misc.py
```diff
@@ -1044,7 +1044,7 @@ def check_new_log():
 @unittest.skipIf(VALGRIND,
                  "Valgrind sometimes fails assert on injected SEGV")
 def test_crashlog(node_factory):
-    l1 = node_factory.get_node(may_fail=True)
+    l1 = node_factory.get_node(may_fail=True, may_log_broken=True)
```
cdecker: `allow_broken_log`?

lightningd/log.c
ZmnSCPxj: I concur.  For `LOG_BROKEN` we can special case it with highlighting asterisks.

rustyrussell:
Changes since first review:

1. Split into two commits.
2. Label every level, not just `**BROKEN**`, but do it before prefix, eg
   `2019-06-21T01:20:48.064Z DEBUG lightningd(734): testing /home/rusty/devel/cvs/lightning/lightningd/lightning_channeld`
3. Change pytest `may_log_broken` to `allow_log_broken`.

cdecker:
The `test_warning_notification` needs a fixup, other than that this is good to go :-)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/98f2cfbe771d17f3607bff051c3c276cd8761403

PR #2744, opened 2019-06-14
Gossip start fail

rustyrussell:


ZmnSCPxj:
ACK  f767f92

jb55:
I think this fixed my first crash, but I'm hitting another one. Will open up a new issue.

PR #2746, opened 2019-06-14
Truncated gossip store fixes

rustyrussell:
After the previous gossip_store bugs, @jb55 produced another one.  It's a corner case where the store is in a bad state due to truncation, but good to fix.  I also added extra tests for other similar cases!

### DIFF
--- a/gossipd/gossip_store.c
+++ b/gossipd/gossip_store.c
@@ -508,6 +508,48 @@ int gossip_store_readonly_fd(struct gossip_store *gs)
 	return fd;
 }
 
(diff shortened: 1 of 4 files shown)


jb55:
Somehow the crashing stopped before this PR, perhaps it was a transitory thing? I'm running on this now and it appears to be working.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2746/commits/6db60cbf50a6e382096a4d361cb0af5090ca9113

Issue #2749, opened 2019-06-15
RPC in master is incompatible with pip upstream pylightning (hangs)


laanwj:
It looks like recent changes to clightning upstream have made it incompatible with the current version of pylightning from pip. This was tested with:

- git commit e9ff8e4
- pylightning-0.0.7 (latest from pip)

strace output:
```
socket(AF_UNIX, SOCK_STREAM|SOCK_CLOEXEC, 0) = 3
connect(3, {sa_family=AF_UNIX, sun_path="/home/.../.lightning/lightning-rpc"}, 41) = 0
sendto(3, "{\"method\": \"getinfo\", \"params\": "..., 44, 0, NULL, 0) = 44
recvfrom(3, "{\"jsonrpc\":\"2.0\",\"id\":0,\"result\""..., 1024, 0, NULL, NULL) = 508
recvfrom(3, … (hangs forever)
```

It looks like it doesn't detect the new message separator so keeps waiting before returning the response. Using the exact version of pylightning from the repository works fine.

cdecker:
Good catch, it was broken in commit https://github.com/ElementsProject/lightning/pull/2705/commits/aaaaf62b9cdfdb95059f5d8a8dd23e0025770d68. I'll release the new version which handles the cleaned up JSON formatting correctly.

cdecker:
Updated:

```
pip install pylightning==0.0.7.3
```

@laanwj can you confirm that it worked?

rustyrussell:
Yuck, I hadn't realized we would break that. Let's hack the extra space back in the JSON for another release.

PR #2752, opened 2019-06-17
lightningd: restore ' ' before '}' for older pylightning.

rustyrussell:
It needs this in compat mode to detect old (pre-0.6.3) end of JSON.
But it always does the first command in compat mode.

This was never really reliable, since the first command could be to
a plugin for which we simply pass through the JSON (though, carefully
appending the expected '\n\n' if not already there).

Reported-by: @laanwj
Fixes: #2749 

(Tested by reverting adc4bf181735c943727d96842216806fe9a6ad16)

### DIFF
--- a/lightningd/json_stream.c
+++ b/lightningd/json_stream.c
@@ -160,6 +160,15 @@ void json_object_end(struct json_stream *js)
 		js_oom(js);
 }
 
+void json_object_compat_end(struct json_stream *js)
+{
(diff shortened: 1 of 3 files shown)


laanwj:
Thanks !

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2752/commits/7c1b2a79420883fa0754bbae5a6160d8ff4c9bc8

PR #2757, opened 2019-06-18
Simplified gossip_store recovery

rustyrussell:
I cannot reproduce #2750 on my nodes.  However, they seem to mainly be caused by us attempting to recover a gossip_store which we've truncated.  Instead, this discards all gossip when that happens.

It should also make any corruption easier to diagnose, as the last patch means we save the corrupt gossip_store to `gossip_store.corrupt` for later diagnosis.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2757/commits/a74977a853631833975ba9b04fc84526add1f9f5

PR #2758, opened 2019-06-20
gossipd: truncate gossip_store.tmp for compaction.

rustyrussell:
If something went wrong and there was an old one, we were appending to it!

Reported-by: @SimonVrouwe


ZmnSCPxj:
ACK e038b17

Can we have a test for it though?

PR #2764, opened 2019-06-21
Offline compact: avoid automatic compaction

rustyrussell:
This is based on #2757 (first 5 commits)

The last commit adds a simple offline compaction step, which is much simpler than the online compaction which seems to be causing problems for #2750

### DIFF
--- a/gossipd/gossip_store.c
+++ b/gossipd/gossip_store.c
@@ -77,12 +77,95 @@ static bool append_msg(int fd, const u8 *msg, u32 timestamp, u64 *len)
 	return writev(fd, iov, ARRAY_SIZE(iov)) == sizeof(hdr) + msglen;
 }
 
+/* Read gossip store entries, copy non-deleted ones.  This code is written
+ * as simply and robustly as possible! */
+static void gossip_store_compact_offline(void)
+{
+	size_t count = 0, deleted = 0;
+	int old_fd, new_fd;
+	struct gossip_hdr hdr;
+	u8 version;
+
+	old_fd = open(GOSSIP_STORE_FILENAME, O_RDONLY);
+	if (old_fd == -1)
+		return;
+	new_fd = open(GOSSIP_STORE_TEMP_FILENAME, O_RDWR|O_TRUNC|O_CREAT, 0600);
+	if (new_fd < 0) {
+		status_broken(
+		    "Could not open file for gossip_store compaction");
+		goto close_old;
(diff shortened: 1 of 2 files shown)


ZmnSCPxj:
I find strange, that we have a `gossip_store` fd in our per-peer state: https://github.com/ElementsProject/lightning/blob/adcb6641a7d1df9c17639d5b9ff2794148fa1844/common/per_peer_state.h#L28

Could it be, that other processes are manipulating gossip store, while compaction is being done?

Even if not the root cause, stylistically it seems to me only `gossipd` should manipulate `gossip_store`.

gossipd/gossip_store.c
```diff
@@ -77,12 +77,95 @@ static bool append_msg(int fd, const u8 *msg, u32 timestamp, u64 *len)
...
+		return;
+	new_fd = open(GOSSIP_STORE_TEMP_FILENAME, O_RDWR|O_TRUNC|O_CREAT, 0600);
+	if (new_fd < 0) {
+		status_broken(
+		    "Could not open file for gossip_store compaction");
+		goto close_old;
+	}
+
+	if (!read_all(old_fd, &version, sizeof(version))
+	    || version != GOSSIP_STORE_VERSION) {
```
niftynei: is there an upgrade path for old versions?

rustyrussell: No.  This simply upgrades the current version.  And the next step deletes unknown version, but that's already-existing code

rustyrussell:
> I find strange, that we have a `gossip_store` fd in our per-peer state:
> 
> https://github.com/ElementsProject/lightning/blob/adcb6641a7d1df9c17639d5b9ff2794148fa1844/common/per_peer_state.h#L28
> 
> Could it be, that other processes are manipulating gossip store, while compaction is being done?
> 
> Even if not the root cause, stylistically it seems to me only `gossipd` should manipulate `gossip_store`.

No, that fd is specifically read-only.

rustyrussell:
Thanks, fixed that test.

gossipd/gossip_store.c
niftynei: oh nice. I should have looked a bit further perhaps. sgtm :)

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2764/commits/13c90cd46e2b3c4ae24c260a0c663d26ab69b1e9

PR #2765, opened 2019-06-22
 bugfix: Proper json for listtransactions' response

jtimon:
Also add the method to pylightning for convenience and consistency.


rustyrussell:
Nice catch! This was caused by the change in semantics at the same time list transactions was submitted.

Of course, a single test would have caught this :(

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/1d2ad59bace04e357f71e7261d3b3ca1d9ac2dd7

rustyrussell:
Triival fix for use of uninitialized var.  Which, again, would have been found with even the most trivial test :(
@cdecker

PR #2766, opened 2019-06-22
listtransactions cleanup

rustyrussell:
It had neither tests (hence it was broken), nor man pages, nor does it follow the field names used elsewhere.

I wrote a simple test and it fails.  Suggest we remove it for release.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2837,11 +2837,34 @@ void wallet_clean_utxos(struct wallet *w, struct bitcoind *bitcoind)
 		tal_free(utxos);
 }
 
+static bool scid_for_channel(struct wallet *w,
+			     u64 channel_id,
+			     struct short_channel_id *scid)
+{
+	sqlite3_stmt *stmt;
+
+	stmt = db_select_prepare(w->db,
+				 "short_channel_id "
+				 "FROM channels "
+				 "WHERE id = ?;");
+
+	sqlite3_bind_int64(stmt, 1, channel_id);
+	if (!db_select_step(w->db, stmt))
+		return false;
+
+	if (!sqlite3_column_short_channel_id(stmt, 0, scid))
+		fatal("Cannot convert short_channel_id for channel id %"PRIu64,
+		      channel_id);
+	db_stmt_done(stmt);
+	return true;
+}
+
 struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t *ctx)
 {
 	sqlite3_stmt *stmt;
 	size_t count;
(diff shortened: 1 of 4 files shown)


wallet/wallet.c
```diff
@@ -2855,7 +2878,23 @@ struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t
 		cur->blockheight = sqlite3_column_int(stmt, 3);
 		cur->txindex = sqlite3_column_int(stmt, 4);
 		cur->type = sqlite3_column_int(stmt, 5);
-		cur->channel_id = sqlite3_column_int(stmt, 6);
+		tal_arr_expand(&channel_ids, sqlite3_column_int(stmt, 6));
+	}
+
+	/* FIXME: This could be done with a complex join. */
```
cdecker: It's called a left outer join :-)

wallet/walletrpc.c
```diff
@@ -801,11 +801,10 @@ static struct command_result *json_listtransactions(struct command *cmd,
 		json_add_u64(response, "blockheight", txs[i].blockheight);
 		json_add_num(response, "txindex", txs[i].txindex);
 		json_add_txtypes(response, "type", txs[i].type);
-		if (txs[i].channel_id != 0) {
-			json_add_num(response, "channel_id", txs[i].channel_id);
-		} else {
-			json_add_null(response, "channel_id");
-		}
+		if (txs[i].scid)
```
cdecker: imho we should keep the `channel_id` in there, maybe call it a `group` key. Alternatively we could also stop deleting channels from the `channels` table ensuring that the `short_channel_id` will always resolve correctly. If a wallet makes use of the grouping it would suddenly not group them anymore once we delete the channel from the `channels` table.

cdecker: I've added a new PR that keeps the `channels` entries around, so that these references will keep working after the channel is fully settled.

wallet/walletrpc.c
```diff
@@ -797,9 +797,9 @@ static struct command_result *json_listtransactions(struct command *cmd,
 	for (size_t i=0; i<tal_count(txs); i++) {
 		json_object_start(response, NULL);
 		json_add_txid(response, "txid", &txs[i].id);
+		json_add_num(response, "output", txs[i].txindex);
```
cdecker: Not really, this is not the transaction output, rather it's the index of the transaction in the block (second element of the short channel ID).

rustyrussell: I originally thought so too, but then my tests failed, so I guessed again :(

PR #2767, opened 2019-06-22
json-rpc: Disable `listtransactions` for the 0.7.1 release

cdecker:
Given the number of open issues with the `listtransactions` RPC method I think it is probably best if we disable the method for the 0.7.1 release, and then fix it up for the next release.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2767/commits/cc976426e13f12cf50e26e06d994cd8f8b230757

PR #2768, opened 2019-06-22
Don't delete channels once they are resolved, just mark them closed.

### LINKED ISSUE #2028: Simply mark channels as closed instead of deleting them from the DB
This is an idea that I've been playing with for some time now, and in the context of the stats tables dangling off of the `channels` table it seems to be a good time to propose this again.

Currently whenever we close a channel and see it confirm past 100 confirmations we actually go and delete the channel from the `channels` table, basically losing all information that was associated with it since it cascades through the database to all other tables. This can result in a couple of strange behaviors that are not immediately apparent to the casual observer, due to the delay.

So my proposal here is to, instead of deleting the entry from the database, simply mark it as `SETTLED` and which would also not load it on restart. This has the same semantics as deleting the entries, but allows us to retrieve information about long settled channels after the fact, and not have some infos, such as channel stats and htlc stats, have dangling references.

I went ahead and measured how much more data a really busy channel would cost if we left it lying around and the end-result is negligible, as long as we clear the `channel_htlcs` table when marking the channel as settled. For 100k payments, the sending side has a 219 MB database, of which 201 MB is `channel_htlcs`, and 17 MB is the `payments` table. For the receiving side things look similar, with the `invoices` table being slightly bigger than the `payments` table on the sending side.

What do you think @rustyrussell @ZmnSCPxj and @niftynei?

cdecker:
Since we are starting to have some tables that reference the `channels` table using foreign keys and these survive the channel I think it's a good idea to keep a minimal version of the channels around. This ensures that foreign key relationships aren't broken once we fully resolve a channel, including access to channel-related fields, and it can allow us to recover some information about old channels for debugging or display purposes.

Fixes #2028

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -850,8 +850,7 @@ bool wallet_channels_load_active(struct wallet *w)
 	sqlite3_stmt *stmt;
 
 	/* We load all channels */
-	stmt = db_select(w->db, "%s FROM channels;", channel_fields);
-
+	stmt = db_select(w->db, "%s FROM channels WHERE state < %d;", channel_fields, CLOSED);
 	w->max_channel_dbid = 0;
 
 	int count = 0;
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -1213,12 +1213,56 @@ void wallet_channel_insert(struct wallet *w, struct channel *chan)
 	wallet_channel_save(w, chan);
 }
 
-void wallet_channel_delete(struct wallet *w, u64 wallet_id)
+void wallet_channel_close(struct wallet *w, u64 wallet_id)
 {
+	/* We keep a couple of dependent tables around as well, such as the
+	 * channel_configs table, since that might help us debug some issues,
+	 * and it is rather limited in size. Tables that can grow quite
+	 * considerably and that are of limited use after channel closure will
+	 * be pruned as well. */
+
 	sqlite3_stmt *stmt;
+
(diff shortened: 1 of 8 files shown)


cdecker:
For reference #2766 was dealing with some channel-related fields becoming unavailable.

wallet/wallet.c
```diff
@@ -1213,13 +1213,57 @@ void wallet_channel_insert(struct wallet *w, struct channel *chan)
...
+	sqlite3_bind_int64(stmt, 1, wallet_id);
+	db_exec_prepared(w->db, stmt);
+
+	/* Set the channel to closed and disassociate with peer */
+	stmt = db_prepare(w->db,
+			  "UPDATE channels "
+			  "SET state=?, peer_id=?"
+			  "WHERE channels.id=?");
+	sqlite3_bind_int64(stmt, 1, CLOSED);
+	sqlite3_bind_null(stmt, 2);
```
rustyrussell: Might be nice to keep peer_id around, but that's a separate change.  Then we'd have some historic information about who our peer was.

cdecker: `peer_id` is a foreign key into the `peers` table, so we'd need to keep the peer around in that table, which then conflicts with the unique constraint when inserting on channel open. We could either stub the peer out (dropping unique fields or replacing them with uniquified versions), or we could query for an existing entry first and insert only if it doesn't exist yet.

PR #2771, opened 2019-06-24
RPC plugin management

darosior:
~~This PR adds a new RPC command : `reloadplugins`, which will initialize and configure each plugin from the default directory (`~/.lightning/plugins/`) *-or from another passed as parameter-* that were not already loaded at lightningd startup.~~

This PR has mutated to now add a `plugin` RPC command : https://github.com/ElementsProject/lightning/pull/2771#issuecomment-505914017.

This permits to manage plugins without having to restart lightningd.

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -10,6 +10,7 @@ export COMPAT=${COMPAT:-1}
 export PATH=$CWD/dependencies/bin:"$HOME"/.local/bin:"$PATH"
 export TIMEOUT=180
 export PYTEST_PAR=2
+export PYTHONPATH=$PWD/contrib/pylightning:$PYTHONPATH
 # If we're not in developer mode, tests spend a lot of time waiting for gossip!
 # But if we're under valgrind, we can run out of memory!
 if [ "$DEVELOPER" = 0 ] && [ "$VALGRIND" = 0 ]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - build: now requires `python3-mako` to be installed, i.e. `sudo apt-get install python3-mako`
 - plugins: a new notification type `invoice_payment` (sent when an invoice is paid) has been added
 - plugins: a new 'channel_opened' notification type is added, which is emitted when a peer succesfully funds a channel to us
+- rpc: a new rpc command is added, `plugin`. It allows one to manage plugins without restarting `lightningd`.
+- plugins: a new boolean field is added to the `init`'s `configuration`, `startup`. It allows a plugin to know if it has been started on `lightningd` startup.
+- plugins: a new boolean field can be added to a plugin manifest, `dynamic`. It allows a plugin to tell if it can be started or stopped "on-the-fly".
 
 ### Deprecated
 
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -714,6 +714,51 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("ping", payload)
 
+    def plugin_start(self, plugin):
+        """
+        Adds a plugin to lightningd.
+        """
+        payload = {
+            "subcommand": "start",
+            "plugin": plugin
+        }
+        return self.call("plugin", payload)
+
+    def plugin_startdir(self, directory):
+        """
+        Adds all plugins from a directory to lightningd.
+        """
+        payload = {
+            "subcommand": "startdir",
+            "directory": directory
+        }
+        return self.call("plugin", payload)
+
+    def plugin_stop(self, plugin):
+        """
+        Stops a lightningd plugin, will fail if plugin is not dynamic.
+        """
+        payload = {
+            "subcommand": "stop",
+            "plugin": plugin
+        }
+        return self.call("plugin", payload)
+
+    def plugin_list(self):
+        """
+        Lists all plugins lightningd knows about.
+        """
+        payload = {
+            "subcommand": "list"
+        }
+        return self.call("plugin", payload)
+
+    def plugin_rescan(self):
+        payload = {
+            "subcommand": "rescan"
+        }
+        return self.call("plugin", payload)
+
     def sendpay(self, route, payment_hash, description=None, msatoshi=None):
         """
         Send along {route} in return for preimage of {payment_hash}
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -99,7 +99,7 @@ class Plugin(object):
 
     """
 
-    def __init__(self, stdout=None, stdin=None, autopatch=True):
+    def __init__(self, stdout=None, stdin=None, autopatch=True, dynamic=True):
         self.methods = {'init': Method('init', self._init, MethodType.RPCMETHOD)}
         self.options = {}
 
@@ -118,6 +118,8 @@ class Plugin(object):
         self.rpc_filename = None
         self.lightning_dir = None
         self.rpc = None
+        self.startup = True
+        self.dynamic = dynamic
         self.child_init = None
 
         self.write_lock = RLock()
@@ -496,6 +498,7 @@ class Plugin(object):
             'rpcmethods': methods,
             'subscriptions': list(self.subscriptions.keys()),
             'hooks': hooks,
+            'dynamic': self.dynamic
         }
 
     def _init(self, options, configuration, request):
@@ -503,6 +506,7 @@ class Plugin(object):
         self.lightning_dir = configuration['lightning-dir']
         path = os.path.join(self.lightning_dir, self.rpc_filename)
         self.rpc = LightningRpc(path)
+        self.startup = configuration['startup']
         for name, value in options.items():
             self.options[name]['value'] = value
 
--- a/doc/lightning-plugin.7.txt
+++ b/doc/lightning-plugin.7.txt
@@ -1,0 +1,48 @@
+LIGHTNING-PLUGIN(7)
+===================
+:doctype: manpage
+
+NAME
+----
+lightning-plugin - Manage plugins with RPC
+
+SYNOPSIS
+--------
+*plugin* command [parameter] [second_parameter]
+
+DESCRIPTION
+-----------
+
+The *plugin* RPC command allows to manage plugins without having to restart lightningd.
+It takes 1 to 3 parameters: a command (start/stop/startdir/rescan/list) which describes the
+action to take and optionally one or two parameters which describes the plugin on which
+the action has to be taken.
+
+The 'start' command takes a path as parameter and will load the plugin available from this
+path.
+
+The 'stop' command takes a plugin name as parameter and will kill and unload the specified
+plugin.
+
+The 'startdir' command takes a directory path as parameter and will load all plugins this
+directory contains.
+
+The 'rescan' command starts all not-already-loaded plugins from the default plugins directory
+(by default '~/.lightning/plugins').
+
+The 'list' command will return all the active plugins.
+
+RETURN VALUE
+------------
+
+On success, this returns an array 'plugins' of objects, one by plugin. Each object contains
+the name of the plugin ('name' field) and its status ('active' boolean field).
+Since plugins are configured asynchronously, a freshly started plugin may not appear immediately.
+
+AUTHOR
+------
+Antoine Poinsot <darosior@protonmail.com> is mainly responsible.
+
+RESOURCES
+---------
+Main web site: https://github.com/ElementsProject/lightning
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1,8 +1,4 @@
-#include "lightningd/plugin.h"
-
 #include <ccan/array_size/array_size.h>
-#include <ccan/intmap/intmap.h>
-#include <ccan/io/io.h>
 #include <ccan/list/list.h>
 #include <ccan/opt/opt.h>
 #include <ccan/pipecmd/pipecmd.h>
@@ -21,6 +17,7 @@
 #include <lightningd/lightningd.h>
 #include <lightningd/notification.h>
 #include <lightningd/options.h>
+#include <lightningd/plugin.h>
 #include <lightningd/plugin_hook.h>
 #include <signal.h>
 #include <sys/stat.h>
@@ -34,70 +31,6 @@
  * willing to wait. Plugins shouldn't do any initialization in the
  * `getmanifest` call anyway, that's what `init `is for. */
 #define PLUGIN_MANIFEST_TIMEOUT 60
-
-struct plugin {
-	struct list_node list;
-
-	pid_t pid;
-	char *cmd;
-	struct io_conn *stdin_conn, *stdout_conn;
-	bool stop;
-	struct plugins *plugins;
-	const char **plugin_path;
-
-	/* Stuff we read */
-	char *buffer;
-	size_t used, len_read;
-
-	/* Our json_streams. Since multiple streams could start
-	 * returning data at once, we always service these in order,
-	 * freeing once empty. */
-	struct json_stream **js_arr;
-
-	struct log *log;
-
-	/* List of options that this plugin registered */
-	struct list_head plugin_opts;
-
-	const char **methods;
-
-	/* Timer to add a timeout to some plugin RPC calls. Used to
-	 * guarantee that `getmanifest` doesn't block indefinitely. */
-	const struct oneshot *timeout_timer;
-
-	/* An array of subscribed topics */
-	char **subscriptions;
-};
-
-struct plugins {
-	struct list_head plugins;
-	size_t pending_manifests;
-
-	/* Currently pending requests by their request ID */
-	UINTMAP(struct jsonrpc_request *) pending_requests;
-	struct log *log;
-	struct log_book *log_book;
-
-	struct lightningd *ld;
-};
-
-/* The value of a plugin option, which can have different types.
- * The presence of the integer and boolean values will depend of
- * the option type, but the string value will always be filled.
- */
-struct plugin_opt_value {
-	char *as_str;
-	int *as_int;
-	bool *as_bool;
-};
-
-struct plugin_opt {
-	struct list_node list;
-	const char *name;
-	const char *type;
-	const char *description;
-	struct plugin_opt_value *value;
-};
 
 struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
 			    struct lightningd *ld)
@@ -133,7 +66,7 @@ void plugin_register(struct plugins *plugins, const char* path TAKES)
 	tal_add_destructor(p, destroy_plugin);
 }
 
-static bool paths_match(const char *cmd, const char *name)
+bool plugin_paths_match(const char *cmd, const char *name)
 {
 	if (strchr(name, PATH_SEP)) {
 		const char *cmd_canon, *name_canon;
@@ -159,7 +92,7 @@ bool plugin_remove(struct plugins *plugins, const char *name)
 	bool removed = false;
 
 	list_for_each_safe(&plugins->plugins, p, next, list) {
-		if (paths_match(p->cmd, name)) {
+		if (plugin_paths_match(p->cmd, name)) {
 			list_del_from(&plugins->plugins, &p->list);
 			tal_free(p);
 			removed = true;
@@ -168,10 +101,7 @@ bool plugin_remove(struct plugins *plugins, const char *name)
 	return removed;
 }
 
-/**
- * Kill a plugin process, with an error message.
- */
-static void PRINTF_FMT(2,3) plugin_kill(struct plugin *plugin, char *fmt, ...)
+void PRINTF_FMT(2,3) plugin_kill(struct plugin *plugin, char *fmt, ...)
 {
 	char *msg;
 	va_list ap;
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -121,7 +121,7 @@ void PRINTF_FMT(2,3) plugin_kill(struct plugin *plugin, char *fmt, ...)
 	msg = tal_vfmt(plugin, fmt, ap);
 	va_end(ap);
 
-	log_broken(plugin->log, "Killing plugin: %s", msg);
+	log_info(plugin->log, "Killing plugin: %s", msg);
 	plugin->stop = true;
 	io_wake(plugin);
 	kill(plugin->pid, SIGKILL);
@@ -813,6 +813,12 @@ static void plugin_manifest_cb(const char *buffer,
 		plugin_kill(
 		    plugin,
 		    "Failed to register options, methods, hooks, or subscriptions.");
+
+	/* If all plugins have replied to getmanifest and this is not
+	 * the startup init, configure them */
+	if (!plugin->plugins->startup && plugin->plugins->pending_manifests == 0)
+		plugins_config(plugin->plugins);
+
 	/* Reset timer, it'd kill us otherwise. */
 	tal_free(plugin->timeout_timer);
 }
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -69,6 +69,7 @@ void plugin_register(struct plugins *plugins, const char* path TAKES)
 	p->configured = false;
 	p->js_arr = tal_arr(p, struct json_stream *, 0);
 	p->used = 0;
+	p->signal_startup = false;
 
 	p->log = new_log(p, plugins->log_book, "plugin-%s",
 			 path_basename(tmpctx, p->cmd));
@@ -808,8 +809,10 @@ static void plugin_manifest_cb(const char *buffer,
 	}
 
 	dynamictok = json_get_member(buffer, resulttok, "dynamic");
-	if (dynamictok && json_to_bool(buffer, dynamictok, &dynamic_plugin))
+	if (dynamictok && json_to_bool(buffer, dynamictok, &dynamic_plugin)) {
+		plugin->signal_startup = true;
 		plugin->dynamic = dynamic_plugin;
+	}
 
 	if (!plugin_opts_add(plugin, buffer, resulttok) ||
 	    !plugin_rpcmethods_add(plugin, buffer, resulttok) ||
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -51,11 +51,22 @@ static void destroy_plugin(struct plugin *p)
 
 void plugin_register(struct plugins *plugins, const char* path TAKES)
 {
-	struct plugin *p;
+	struct plugin *p, *p_temp;
+
+	/* Don't register an already registered plugin */
+	list_for_each(&plugins->plugins, p_temp, list) {
+		if (streq(path, p_temp->cmd)) {
+			if (taken(path))
+				tal_free(path);
+			return;
+		}
+	}
+
 	p = tal(plugins, struct plugin);
 	list_add_tail(&plugins->plugins, &p->list);
 	p->plugins = plugins;
 	p->cmd = tal_strdup(p, path);
+	p->configured = false;
 	p->js_arr = tal_arr(p, struct json_stream *, 0);
 	p->used = 0;
 
@@ -781,9 +792,9 @@ static void plugin_manifest_cb(const char *buffer,
 	const jsmntok_t *resulttok;
 
 	/* Check if all plugins have replied to getmanifest, and break
-	 * if they are */
+	 * if they have and this is the startup init */
 	plugin->plugins->pending_manifests--;
-	if (plugin->plugins->pending_manifests == 0)
+	if (plugin->plugins->startup && plugin->plugins->pending_manifests == 0)
 		io_break(plugin->plugins);
 
 	resulttok = json_get_member(buffer, toks, "result");
@@ -893,20 +904,17 @@ void plugins_add_default_dir(struct plugins *plugins, const char *default_dir)
 	}
 }
 
-void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
+void plugins_start(struct plugins *plugins, const char *dev_plugin_debug)
 {
 	struct plugin *p;
 	char **cmd;
 	int stdin, stdout;
 	struct jsonrpc_request *req;
-	plugins->pending_manifests = 0;
-	uintmap_init(&plugins->pending_requests);
 
-	plugins_add_default_dir(plugins, path_join(tmpctx, plugins->ld->config_dir, "plugins"));
-
-	setenv("LIGHTNINGD_PLUGIN", "1", 1);
-	/* Spawn the plugin processes before entering the io_loop */
 	list_for_each(&plugins->plugins, p, list) {
+		if (p->configured)
+			continue;
+
 		bool debug;
 
 		debug = dev_plugin_debug && strends(p->cmd, dev_plugin_debug);
@@ -945,9 +953,24 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 		}
 		tal_free(cmd);
 	}
+}
+
+void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
+{
+	plugins->pending_manifests = 0;
+	uintmap_init(&plugins->pending_requests);
+
+	plugins_add_default_dir(plugins,
+				path_join(tmpctx, plugins->ld->config_dir, "plugins"));
+
+	setenv("LIGHTNINGD_PLUGIN", "1", 1);
+	/* Spawn the plugin processes before entering the io_loop */
+	plugins_start(plugins, dev_plugin_debug);
 
 	if (plugins->pending_manifests > 0)
 		io_loop_with_timers(plugins->ld);
+	// There won't be io_loop anymore to wait for plugins
+	plugins->startup = false;
 }
 
 static void plugin_config_cb(const char *buffer,
@@ -955,7 +978,7 @@ static void plugin_config_cb(const char *buffer,
 			     const jsmntok_t *idtok,
 			     struct plugin *plugin)
 {
-	/* Nothing to be done here, this is just a report */
+	plugin->configured = true;
 }
 
 /* FIXME(cdecker) This just builds a string for the request because
@@ -995,7 +1018,8 @@ void plugins_config(struct plugins *plugins)
 {
 	struct plugin *p;
 	list_for_each(&plugins->plugins, p, list) {
-		plugin_config(p);
+		if (!p->configured)
+			plugin_config(p);
 	}
 }
 
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -789,7 +789,8 @@ static void plugin_manifest_cb(const char *buffer,
 			       const jsmntok_t *idtok,
 			       struct plugin *plugin)
 {
-	const jsmntok_t *resulttok;
+	const jsmntok_t *resulttok, *dynamictok;
+	bool dynamic_plugin;
 
 	/* Check if all plugins have replied to getmanifest, and break
 	 * if they have and this is the startup init */
@@ -805,6 +806,10 @@ static void plugin_manifest_cb(const char *buffer,
 			    buffer + toks[0].start);
 		return;
 	}
+
+	dynamictok = json_get_member(buffer, resulttok, "dynamic");
+	if (dynamictok && json_to_bool(buffer, dynamictok, &dynamic_plugin))
+		plugin->dynamic = dynamic_plugin;
 
 	if (!plugin_opts_add(plugin, buffer, resulttok) ||
 	    !plugin_rpcmethods_add(plugin, buffer, resulttok) ||
@@ -1014,6 +1019,7 @@ static void plugin_config(struct plugin *plugin)
 	json_object_start(req->stream, "configuration");
 	json_add_string(req->stream, "lightning-dir", ld->config_dir);
 	json_add_string(req->stream, "rpc-file", ld->rpc_filename);
+	json_add_bool(req->stream, "startup", plugin->plugins->startup);
 	json_object_end(req->stream);
 
 	jsonrpc_request_end(req);
--- a/lightningd/plugin.h
+++ b/lightningd/plugin.h
@@ -21,6 +21,8 @@ struct plugin {
 	struct plugins *plugins;
 	const char **plugin_path;
 	bool configured;
+	/* If this plugin can be restarted without restarting lightningd */
+	bool dynamic;
 
 	/* Stuff we read */
 	char *buffer;
--- a/lightningd/plugin.h
+++ b/lightningd/plugin.h
@@ -23,6 +23,7 @@ struct plugin {
 	bool configured;
 	/* If this plugin can be restarted without restarting lightningd */
 	bool dynamic;
+	bool signal_startup;
 
 	/* Stuff we read */
 	char *buffer;
--- a/lightningd/plugin.h
+++ b/lightningd/plugin.h
@@ -1,28 +1,88 @@
 #ifndef LIGHTNING_LIGHTNINGD_PLUGIN_H
 #define LIGHTNING_LIGHTNINGD_PLUGIN_H
 #include "config.h"
+#include <ccan/intmap/intmap.h>
+#include <ccan/io/io.h>
 #include <ccan/take/take.h>
 #include <ccan/tal/tal.h>
 #include <lightningd/jsonrpc.h>
 #include <lightningd/log.h>
 
 /**
+ * A plugin, exposed as a stub so we can pass it as an argument.
+ */
+struct plugin {
+	struct list_node list;
+
+	pid_t pid;
+	char *cmd;
+	struct io_conn *stdin_conn, *stdout_conn;
+	bool stop;
+	struct plugins *plugins;
+	const char **plugin_path;
+
+	/* Stuff we read */
+	char *buffer;
+	size_t used, len_read;
+
+	/* Our json_streams. Since multiple streams could start
+	 * returning data at once, we always service these in order,
+	 * freeing once empty. */
+	struct json_stream **js_arr;
+
+	struct log *log;
+
+	/* List of options that this plugin registered */
+	struct list_head plugin_opts;
+
+	const char **methods;
+
+	/* Timer to add a timeout to some plugin RPC calls. Used to
+	 * guarantee that `getmanifest` doesn't block indefinitely. */
+	const struct oneshot *timeout_timer;
+
+	/* An array of subscribed topics */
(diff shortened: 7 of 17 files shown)


rustyrussell:
Can we have a top-level "plugin" command, with "stop", "start", "adddir"? Then we can stop and start them individually or all?

lightningd/plugin.c
```diff
@@ -974,11 +986,11 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 	plugins->pending_manifests = 0;
 	uintmap_init(&plugins->pending_requests);
 
-	plugins_add_default_dir(plugins, path_join(tmpctx, plugins->ld->config_dir, "plugins"));
-
-	setenv("LIGHTNINGD_PLUGIN", "1", 1);
```
ZmnSCPxj: Why is env removed?  In theory we do this so that a plugin could be run outside of `lightningd` (it could do something as simple as print "this is a lightningd plugin, not intended to be run directly as a command").

lightningd/plugin.c
```diff
@@ -1146,3 +1161,51 @@ struct log *plugin_get_log(struct plugin *plugin)
...
+		DIR *d = opendir(plugin_path);
+		if (!d)
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					"Could not open %s", plugin_path);
+		closedir(d);
+		add_plugin_dir(cmd->ld->plugins, plugin_path, true);
+	} else
+		plugins_add_default_dir(cmd->ld->plugins,
+					  path_join(tmpctx, cmd->ld->config_dir, "plugins"));
+	plugins_init(cmd->ld->plugins, cmd->ld->dev_debug_subprocess);
```
ZmnSCPxj: `plugin_init` clears the `pending_requests` field, which is used to route plugin responses back to the command (I think).  So if we execute `reload_plugin` while a plugin command is ongoing, this will prevent those command responses from being recognized one the plugin finishes the command, and also potentially leak the memory involved.

Should `reload_plugin` actively *kill* running plugins (and possibly raise an `error` on existing commands that are being handled by the plugin)?

lightningd/plugin.c
darosior: This was not intended

lightningd/plugin.c
darosior: > plugin_init clears the pending_requests field, which is used to route plugin responses back to the command (I think). So if we execute reload_plugin while a plugin command is ongoing, this will prevent those command responses from being recognized once the plugin finishes the command, and also potentially leak the memory involved.

>

darosior:
> Can we have a top-level "plugin" command, with "stop", "start", "adddir"? Then we can stop and start them individually or all?
> 
> And list of course...

Yes, I'm not sure but I think I can do that.
@rustyrussell by the way, do you know why calling another function from [`lightningd/plugin.c`](https://github.com/darosior/lightning/blob/bf41958f90b2ff262d61ac2ebccba7c7a19974ab/lightningd/lightningd.c#L669) when having configured with `--enable-dev` causes a link error ?

@ZmnSCPxj thank you for the review

> You also need to consider what happens to ongoing plugin-based commands. Would not it be better if we enter a "shutting down" mode where existing plugin-provided commands are allowed to complete, new plugin-provided commands are delayed until after reload, and once current plugin commands have completed, only then kill and reload plugins?

I may add a check, even if it seems weird to me that someone would `reload_plugins` if a `withdraw` command is ongoing.

lightningd/plugin.c
ZmnSCPxj: > I thought that nobody would reload plugins while a command while pending..

Nobody might do so deliberately, but there exist such things as bugs.

darosior: > I cannot see yet any code that kills plugins for reload_plugins, is this deliberate or did I miss?

Actually it only loads plugin which have not been loaded at launch. The name might not be accurate..

ZmnSCPxj: `load_new_plugins`?

If so then running plugin-provided commands should be unaffected. Suggest refactoring the clearing of the fields away from starting new plugins then.

darosior:
According to the above comments by @rustyrussell and @ZmnSCPxj : 
- it is now a `plugin` command which takes a `command` and an optional `parameter`. The command returns in any case a list `active_plugins` of plugins loaded after the command execution.
    - The `start` command adds a plugin given a path as `parameter`
    - The `stop` command stops (kill and unload) a plugin given a plugin name as `parameter`
    - The `adddir` command adds a plugin directory given a directory path as `parameter`
    - The `reload` command adds all non-already-loaded plugins from the default plugins directory
    - The `list` command just output the active plugins
- it has been moved to a new file : `plugin_control.c`

darosior:
Rebased to move `plugins_add_directory()` back from `lightningd.c`'s `main()` to `plugins_init()` in `plugin.c` in order to make the `--enable-developer` build pass.

Since the load from the default plugin directory is now done each time we call a `plugin` command there is no more need for a dedicated `reload` command, so I removed it along with the above movement.

darosior:
The travis failure is due to `lightningd` crashing if the `plugin` command is used just after having started the daemon.
I tried (the above force pushs) to increase the `time.sleep(n)` in tests to make Travis tests pass (they pass locally) but I need to find the root cause of this crash in order to make a condition in the command so that `plugins_init()` is not called too early.
I've been investigating today and did not find it : the backtrace gets back to the `plugins_init()`'s `io_loop` to wait for manifests, but it appears that the loop is broken very early and that it's not the cause of the crash. Does anyone could explain me why we can't make another `io_loop` just after startup without waiting about 5 seconds ? Knowing that passed that delay, we can call the `plugin` command as many times in a row as we want ? Is disabling the command for 5s after startup (just to avoid the crash for the user) something that should be avoided ?

<details>
<summary> Backtrace to avoid pull and compilation </summary>
<p>

The backtrace after having quiet the warning notifications (`plugin_notify()`) which were part of the backtrace (memleak due to crash) without being the root cause.

```bash
2019-06-27T20:52:39.201Z lightningd(23821): --------------------------------------------------
2019-06-27T20:52:39.201Z lightningd(23821): Server started with public key 023bafa606486c3066f71d98f385dcfb24742191589922bcbea49373c21c4e7457, alias PEEVEDSOURCE-623-g726d6b1-modded (color #023baf) and lightningd v0.7.0-623-g726d6b1-modded
Already in transaction from lightningd/jsonrpc.c:653
2019-06-27T20:52:40.251Z lightningd(23821): Already in transaction from lightningd/jsonrpc.c:653
lightningd: FATAL SIGNAL 6 (version v0.7.0-623-g726d6b1-modded)
0x55af55bb0aa8 send_backtrace
	common/daemon.c:40
0x55af55bb0b4e crashdump
	common/daemon.c:53
0x7fa88ffdc83f ???
	???:0
0x7fa88ffdc7bb ???
	???:0
0x7fa88ffc7534 ???
	???:0
0x55af55b8be73 fatal
	lightningd/log.c:622
0x55af55bdd0ae db_begin_transaction_
	wallet/db.c:637
0x55af55ba990e sd_msg_read
	lightningd/subd.c:409
0x55af55bf64c9 next_plan
	ccan/ccan/io/io.c:59
0x55af55bf7046 do_plan
	ccan/ccan/io/io.c:407
0x55af55bf7084 io_ready
	ccan/ccan/io/io.c:417
0x55af55bf90d9 io_loop
	ccan/ccan/io/poll.c:445
0x55af55ba778d plugins_init
	lightningd/plugin.c:969
0x55af55ba8042 json_plugin_control
	lightningd/plugin_control.c:65
0x55af55b8729b parse_request
	lightningd/jsonrpc.c:654
0x55af55b876d5 read_json
	lightningd/jsonrpc.c:752
0x55af55bf64c9 next_plan
	ccan/ccan/io/io.c:59
0x55af55bf7046 do_plan
	ccan/ccan/io/io.c:407
0x55af55bf7084 io_ready
	ccan/ccan/io/io.c:417
0x55af55bf90d9 io_loop
	ccan/ccan/io/poll.c:445
0x55af55b83b0d io_loop_with_timers
	lightningd/io_loop_with_timers.c:24
0x55af55b89d43 main
	lightningd/lightningd.c:827
0x7fa88ffc909a ???
	???:0
0x55af55b720d9 ???
	???:0
0xffffffffffffffff ???
	???:0
2019-06-27T20:52:40.258Z lightningd(23821): FATAL SIGNAL 6 (version v0.7.0-623-g726d6b1-modded)
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: common/daemon.c:45 (send_backtrace) 0x55af55bb0afe
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: common/daemon.c:53 (crashdump) 0x55af55bb0b4e
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0x7fa88ffdc83f
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0x7fa88ffdc7bb
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0x7fa88ffc7534
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/log.c:622 (fatal) 0x55af55b8be73
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: wallet/db.c:637 (db_begin_transaction_) 0x55af55bdd0ae
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/subd.c:409 (sd_msg_read) 0x55af55ba990e
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:59 (next_plan) 0x55af55bf64c9
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:407 (do_plan) 0x55af55bf7046
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:417 (io_ready) 0x55af55bf7084
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/poll.c:445 (io_loop) 0x55af55bf90d9
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/plugin.c:969 (plugins_init) 0x55af55ba778d
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/plugin_control.c:65 (json_plugin_control) 0x55af55ba8042
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/jsonrpc.c:654 (parse_request) 0x55af55b8729b
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/jsonrpc.c:752 (read_json) 0x55af55b876d5
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:59 (next_plan) 0x55af55bf64c9
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:407 (do_plan) 0x55af55bf7046
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/io.c:417 (io_ready) 0x55af55bf7084
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: ccan/ccan/io/poll.c:445 (io_loop) 0x55af55bf90d9
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/io_loop_with_timers.c:24 (io_loop_with_timers) 0x55af55b83b0d
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: lightningd/lightningd.c:827 (main) 0x55af55b89d43
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0x7fa88ffc909a
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0x55af55b720d9
2019-06-27T20:52:40.258Z lightningd(23821): backtrace: (null):0 ((null)) 0xffffffffffffffff
Log dumped in crash.log.20190627205240
Aborted
```

</p>

ZmnSCPxj:
Like I suggested, you should split `plugins_init`.  Initiating the startup of new plugins should be a separate function, which `plugins_init` calls, e.g. `plugins_start_plugins`.

In particular, you should not do an `io_loop` within an `io_loop`.

So `plugins_start_plugins` would be an async interface which calls the callback only when `pending_manifests` drops to 0.  Both `plugins_init` and `json_plugin_control` should call into it.  And `json_plugin_control` ***MUST NOT*** call `plugins_init`.

For `plugins_init`, it provides a callback which executes `io_break`.  `plugins_init` will also include its own `io_loop`.

For `json_plugin_control`, it provides a callback which provides the actual result object of the `plugin` command.  In particular it would ***not*** have its own `io_loop`.

darosior:
@ZmnSCPxj thank you the support : I had to modify `plugin.c` some more but it now works well.

I added the `reload` control back since `json_plugin_control` does not call `plugin_init` anymore. I also added a `unregister_hook` control that might be useful.

darosior:
@ZmnSCPxj do the changes seem good to you ?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,127 @@
...
+	"	adds a new plugin to c-lightning\n"
+	"cli plugin stop plugin_name\n"
+	"	stops an already registered plugin\n"
+	"cli plugin adddir /path/to/a/plugin_dir/\n"
+	"	adds a new plugin directory\n"
+	"cli plugin reload\n"
+	"	loads not-already-loaded plugins from the default plugins dir\n"
+	"cli plugin list\n"
+	"	lists all active plugins\n"
+	"\n"
```
ZmnSCPxj: `unregister_hook` is not exist here?

lightningd/plugin_control.h
```diff
@@ -0,0 +1,14 @@
+#ifndef LIGHTNING_LIGHTNINGD_PLUGIN_CONTROL_H
+#define LIGHTNING_LIGHTNINGD_PLUGIN_CONTROL_H
+#include "config.h"
+#include <lightningd/lightningd.h>
+
+/**
+ * A plugin command which permits to control (start, stop, ..) plugins
+ */
+struct command_result *json_plugin_control(struct command *cmd,
```
ZmnSCPxj: Not need to make this public; can make `static` in the `plugin_control.c` file.

doc/lightning-plugin.7.txt
```diff
@@ -0,0 +1,45 @@
...
+
+DESCRIPTION
+-----------
+
+The *plugin* RPC command allows to manage plugins without having to restart lightningd.
+It takes 1 to 3 parameters: a command (start/stop/adddir/reload/list) which describes the
+action to take and optionally one or two parameters which describes the plugin on which
+the action has to be taken.
+
+The 'start' command takes a path as parameter and will load the plugin available from this
```
ZmnSCPxj: The help description contains example, but manpage does not?

tests/test_plugin.py
```diff
@@ -88,6 +88,48 @@ def test_plugin_dir(node_factory):
     node_factory.get_node(options={'plugin-dir': plugin_dir, 'greeting': 'Mars'})
 
 
+def test_plugin_command(node_factory):
+    """Tests the 'reloadplugins' RPC command"""
```
ZmnSCPxj: Inaccurate comment.

lightningd/plugin_control.h
darosior: Ok, is an emty header file OK ?

lightningd/plugin_control.c
darosior: Forgot to add it to the verbose help.

doc/lightning-plugin.7.txt
darosior: I made a `cli command parameter` in verbose help to make it concise. I can replicate it in the manpage too, but it seems already more detailed.

lightningd/plugin_control.h
ZmnSCPxj: Should be okay.

doc/lightning-plugin.7.txt
ZmnSCPxj: Generally manpage is more detailed and includes things like possible errors.  Maybe okay for now.

darosior:
Thank you for the review @ZmnSCPxj. I added the forgotten `unregister_hook` verbose help, corrected the test comment and made `json_plugin_control()` static.

contrib/pylightning/lightning/plugin.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, stdout=None, stdin=None, autopatch=True):
         if not stdin:
             self.stdin = sys.stdin
 
-        if os.getenv('LIGHTNINGD_PLUGIN') and autopatch:
+        if autopatch:
```
rustyrussell: This patch is wrong, AFAICT.

lightningd/plugin.c
```diff
@@ -781,9 +792,9 @@ static void plugin_manifest_cb(const char *buffer,
 	const jsmntok_t *resulttok;
 
 	/* Check if all plugins have replied to getmanifest, and break
-	 * if they are */
+	 * if they have and this is the startup init */
 	plugin->plugins->pending_manifests--;
-	if (plugin->plugins->pending_manifests == 0)
+	if (plugin->plugins->pending_manifests == 0 && !getenv("LIGHTNINGD_PLUGIN"))
```
rustyrussell: Abusing the env var to see if we're starting up is pretty horrible!

Better to have a 'startup' bool, or simply set pending_manifests to (size_t)-1 every time you start a plugin after startup, so this never happens?

lightningd/plugin.c
```diff
@@ -943,17 +951,31 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
...
+	plugins_add_default_dir(plugins,
+				path_join(tmpctx, plugins->ld->config_dir, "plugins"));
+
+	/* Spawn the plugin processes before entering the io_loop */
+	plugins_start(plugins, dev_plugin_debug);
 
 	if (plugins->pending_manifests > 0)
 		io_loop_with_timers(plugins->ld);
+
+	setenv("LIGHTNINGD_PLUGIN", "1", 1);
```
rustyrussell: You need to move this line back to before the io_loop, BTW.

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	const char *control_command, *control_param, *control_second_param;
+	struct plugin *p;
+	struct json_stream *response;
+	bool plugin_found;
+
+	if (!param(cmd, buffer, params,
+		   p_req("command", param_string, &control_command),
+		   p_opt("parameter", param_string, &control_param),
+		   p_opt("second_parameter", param_string, &control_second_param),
+		   NULL))
```
rustyrussell: We probably need more powerful param handling for cases like this, TBH.  This will do for now, however.

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	if (!control_param && !streq(control_command, "list")
+			&& !streq(control_command, "reload"))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					"Missing a required parameter to the"
+					" plugin command \"%s\"", control_command);
+
+	if (streq(control_command, "stop")) {
+		plugin_found = false;
+		list_for_each(&cmd->ld->plugins->plugins, p, list) {
+			if (paths_match(p->cmd, control_param)) {
```
rustyrussell: Rather than export paths_match, we should export 'find_plugin' which does this loop for us?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+				plugin_kill(p, "%s stopped by lightningd via RPC",
+						control_param);
+				break;
+			}
+		}
+		if (!plugin_found)
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					   "Could not find plugin %s", control_param);
+	}
+	else if (streq(control_command, "start")) {
```
rustyrussell: Nit: `else if` belongs on same line as `}`

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+		if (access(control_param, X_OK) == 0)
+			plugin_register(cmd->ld->plugins, control_param);
+		else
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						   "Could not access %s with execution permission",
+						   control_param);
```
rustyrussell: How about: `"%s is not executable: %s", control_param, strerror(errno))` which is more informative if the problem is something weird?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					   "Could not find plugin %s", control_param);
+	}
+	else if (streq(control_command, "list")) {
+		/* Don't do anything as we return the plugin list anyway */
+	}
+	else {
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					"Invalid control : %s should be one of"
+					" stop, start, adddir, reload",
```
rustyrussell: ... or unregister_hook?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	 * in 'plugin_manifest_cb'.*/
+	plugins_start(cmd->ld->plugins, cmd->ld->dev_debug_subprocess);
+
+	response = json_stream_success(cmd);
+	json_array_start(response, "active_plugins");
+	list_for_each(&cmd->ld->plugins->plugins, p, list) {
+		if (p->configured)
+			json_add_string(response, NULL, p->cmd);
+	}
+	json_array_end(response);
```
rustyrussell: Might you want inactive plugins too?  How about a simple "plugins" with a 'active' boolean?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	return command_success(cmd, response);
+}
+
+static const struct json_command plugin_control_command = {
+	"plugin",
+	"plugin",
+	json_plugin_control,
+	"Control plugins (start, stop, adddir, reload, list)",
+	.verbose = "Usage :\n"
+	"cli plugin start /path/to/a/plugin\n"
```
rustyrussell: skip the 'cli' here, since it doesn't apply for non-lightning-cli users?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	"	adds a new plugin to c-lightning\n"
+	"cli plugin stop plugin_name\n"
+	"	stops an already registered plugin\n"
+	"cli plugin adddir /path/to/a/plugin_dir/\n"
+	"	adds a new plugin directory\n"
```
rustyrussell: `start` vs `adddir` is weird.  `add` vs `adddir` or `start` vs `startdir`?

rustyrussell: Or maybe just `start` can take a filename or a directory?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,135 @@
...
+	"cli plugin reload\n"
+	"	loads not-already-loaded plugins from the default plugins dir\n"
```
rustyrussell: That is not what I expected to happen when I say `reload`; I expected it to take a plugin name to restart, or maybe `all` or nothing for everything?

lightningd/plugin.c
darosior: > Abusing the env var to see if we're starting up is pretty horrible!
Better to have a 'startup' bool, or simply set pending_manifests to (size_t)-1 every time you start a plugin after startup, so this never happens?

I did that because it was already used and to avoid yet another member, I'll amend it if it's preferable.

lightningd/plugin.c
darosior: This was because of the above, to stop `io_break`ing after this loop.

lightningd/plugin_control.c
darosior: `find_plugin` would abort on find failure IIRC

lightningd/plugin_control.c
darosior: Because of the `else if` repetitions I took the liberty to add some space for readability. Will amend.

lightningd/plugin_control.c
darosior: A plugin could be reloaded by `stop`ing and `start`ing it again, however I think it's a good design to involve the `plugins` directory and to make it possible to load all new plugins from this directory. Maybe the command name is not well chosen (as @ZmnSCPxj already pointed out)

contrib/pylightning/lightning/plugin.py
darosior: Since the `LIGHTNINGD_PLUGIN` was moved after the `io_loop` for startup plugins initialization, this would prevent some plugins to be initiated at startup. I'll amend this anyway.

lightningd/plugin_control.c
darosior: I chose `start` and `startdir` to avoid `ddd` which is error prone :)

lightningd/plugin_control.c
darosior: Done

darosior:
Thank you for the review @rustyrussell 
I rebased to : 
- remove the use of the (already present :) `LIGHTNINGD_PLUGIN` environment variable in favor of a `plugins` struct member to check for startup.
- add the inactive plugins to the command return. It knows output an object with a plugin name as key and its state (active/inactive) as value.
- some style/log/help nits.

rustyrussell:
With respect to @cdecker's concerns, I only somewhat agree.  Some plugins are fine with stop, start and reload, others are not.  I think it has to be an opt-in mechanism at getmanifest time: a new `"dynamic": true` field in the reply?

In addition, we should probably tell plugins, with `"startup": true` in the init call, that it's being called as part of the initial startup or not.  Separately, we might add a shutdown phase for plugins, but thats a different PR IMHO.

darosior:
> I'm not really convinced we want this at all:
> 
> * One of the guarantees that we currently have is that a plugin cannot miss a notification or hook, because they are configured and initialized before any can trigger. If we allow plugins to be started after `lightnignd` has been running for a while we lose that guarantee.

Is this guarantee required ? If one starts a plugin after `lightningd` has been running for a while he expects that the plugin won't receive notifications until it is fully initialized. 

> * Adding plugins really is a configuration option, not a runtime operation, so restarting the daemon is the correct way of adding a plugin in my mind.

I think this is powerful to be able to start plugins "on-the-fly" : some users might don't want to restart `lightningd` every time they want to add or remove a plugin.

> This is important since some plugins may rely on them being present after a restart (if we are managing invoices externally, or anything that uses a `hook` really). If we allow runtime loading of plugins we are breaking with that assumption.

This was why this PR initially permitted to only reload plugins from the plugins directory. If using the default plugins directory (or the `lightning/plugins` directory) to store plugins becomes standard then plugins added while `lightningd` was running will be loaded at the next startup. 

> * Being able to unregister a hook may also break a load of things: image having a backup plugin using the `db_write`, that gets unregistered and then re-registered after a while. It'd result in an incomplete backup that could be dangerous. Plugins registering hooks really expect the registration to be sticky. Let's rather add hook-result-combinators like @ZmnSCPxj proposed in #2796

I can remove this one. I understand that it potentially permits to break a load of things.
 
> That being said, you might also achieve the same goals by using the [autoreload](https://github.com/lightningd/plugins/tree/master/autoreload) plugin.

The [autoreload](https://github.com/lightningd/plugins/tree/master/autoreload) plugin seems to not allow to start new plugins after `lightningd` startup.




> With respect to @cdecker's concerns, I only somewhat agree. Some plugins are fine with stop, start and reload, others are not. I think it has to be an opt-in mechanism at getmanifest time: a new `"dynamic": true` field in the reply?
> 
> In addition, we should probably tell plugins, with `"startup": true` in the init call, that it's being called as part of the initial startup or not. Separately, we might add a shutdown phase for plugins, but thats a different PR IMHO.

I'll add this two fields, along with modifying the params to use #2814 if it worths that I continue updating this.

darosior:
Rebased, changed params to use #2814's subcommands, and removed the `unregister_hook` subcommand.

darosior:
Added the `getmanifest` and `init` fields as @rustyrussell [proposed](https://github.com/ElementsProject/lightning/pull/2771#issuecomment-511612602).

darosior:
I added a flag as adding a new field to `init` would break Python plugins using previous pylightning versions.

contrib/pylightning/lightning/lightning.py
```diff
@@ -714,6 +714,19 @@ def ping(self, peer_id, length=128, pongbytes=128):
...
+        Manage plugins.
+        """
+        payload = {
+            "subcommand": subcommand
+        }
+        if subcommand in {"stop", "start"}:
+            payload["plugin"] = parameter
+        elif subcommand == "startdir":
+            payload["directory"] = parameter
+        return self.call("plugin", payload)
```
rustyrussell: Hmm, discarding parameters if we don't know the subcommand seems wrong?  I think we need to whitelist the known parameters?

lightningd/plugin.c
```diff
@@ -1055,6 +1022,9 @@ static void plugin_config(struct plugin *plugin)
 	json_add_string(req->stream, "rpc-file", ld->rpc_filename);
 	json_object_end(req->stream);
 
+	if (plugin->signal_startup)
+		json_add_bool(req->stream, "startup", plugin->plugins->startup);
```
rustyrussell: I don't understand why having this set unconditionally breaks older pylightning?  We should be able to add init fields without repercussions; if we can't, can we fix that now so this won't be a problem in future?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,132 @@
...
+ * lightningd. It takes a subcommand, and an optional subcommand parameter.
+ */
+static struct command_result *json_plugin_control(struct command *cmd,
+						const char *buffer,
+						const jsmntok_t *obj UNNEEDED,
+						const jsmntok_t *params)
+{
+	const char *subcmd;
+	subcmd = param_subcommand(cmd, buffer, params,
+							  "start", "stop", "startdir", "reload", "list", NULL);
```
rustyrussell: Weird indent here: do you have tabs set to 4 spaces?

lightningd/plugin_control.c
rustyrussell: "rescandir"?

lightningd/plugin_control.c
```diff
@@ -0,0 +1,132 @@
...
+
+	/* The config function is called once we got the manifest,
+	 * in 'plugin_manifest_cb'.*/
+	plugins_start(cmd->ld->plugins, cmd->ld->dev_debug_subprocess);
+
+	response = json_stream_success(cmd);
+	json_object_start(response, "plugins");
+	list_for_each(&cmd->ld->plugins->plugins, p, list)
+		json_add_string(response, p->cmd, p->configured ? "active" : "inactive");
+	json_object_end(response);
```
rustyrussell: We can't add fields to this, so it needs to be object-wrapped, even if we don't have other fields today, so it needs to be:
```
response = json_stream_success(cmd);
json_array_start(response, "plugins");
list_for_each(&cmd->ld->plugins->plugins, p, list) {
        json_add_bool(response, "active", p->configured);
        json_object_end(response);
}
```

In particular, in future we'll want to know what ones are dynamic, and in future what hooks they are registered for...

contrib/pylightning/lightning/plugin.py
```diff
@@ -99,7 +99,7 @@ class Plugin(object):
 
     """
 
-    def __init__(self, stdout=None, stdin=None, autopatch=True):
+    def __init__(self, stdout=None, stdin=None, autopatch=True, dynamic=True):
```
rustyrussell: dynamic needs to default to False, I think.  Though perhaps there are only a few plugins which can simply be fixed.  @cdecker ?

lightningd/plugin.c
darosior: This is not a problem with plugins in general, but with how Pylightning (handfully) unpack fields into function. To avoid that now and in the future, we could either wrap the init response in an `init` object as in #2823 bit this would break all existing plugins.

lightningd/plugin_control.c
darosior: No, but when aligning I'm tricked by tabs

contrib/pylightning/lightning/plugin.py
darosior: All plugins from the repo seems ok (even with `dynamic = True` they are not restarted.

darosior:
> and a question about adding init fields?

*(answering here too)*The problem is not with adding fields in the init response object but with how Pylightning conveniently unpack it to function parameters (cf #2823) : it has the side effect that a change in the object received will cause a break of the plugin if the callback Python function isn't amended to add a new parameter. We could wrap it into a top object as I propose for hooks and notifications in #2826 but this is a backward incompatible change.
However I think that's outside of the scope of this PR : we could handle this along with plugin shutdown (I have questions about this too but i'll make an issue to discuss this).

> 
> Oh, and this definitely deserves a CHANGELOG line :)

I'll add it :-)

contrib/pylightning/lightning/lightning.py
darosior: I don't understand, do you mean whitelist on the C side ? Like accepting parameters for `reload` (soon `rescandir` :) and `list` ?
https://github.com/darosior/lightning/blob/09a2dea65292e6713aadeb60e253da8d0bd4f92c/lightningd/plugin_control.c#L81-L95

darosior:
Rebased to : 
- correct indentation
- rename `reload` subcommand to `rescan`
- return an array of objects (`{ "name": "my_plugin", "active": true }`)

lightningd/plugin.h
```diff
@@ -1,28 +1,88 @@
...
+#include <ccan/io/io.h>
 #include <ccan/take/take.h>
 #include <ccan/tal/tal.h>
 #include <lightningd/jsonrpc.h>
 #include <lightningd/log.h>
 
+/**
+ * A plugin, exposed as a stub so we can pass it as an argument.
+ */
+struct plugin {
```
cdecker: Why is it necessary to make all these things public? Wouldn't it be sufficient to make `plugin_kill` public?

lightningd/plugin.c
```diff
@@ -813,6 +813,12 @@ static void plugin_manifest_cb(const char *buffer,
 		plugin_kill(
 		    plugin,
 		    "Failed to register options, methods, hooks, or subscriptions.");
+
+	/* If all plugins have replied to getmanifest and this is not
+	 * the startup init, configure them */
+	if (!plugin->plugins->startup && plugin->plugins->pending_manifests == 0)
+		plugins_config(plugin->plugins);
```
cdecker: Doesn't this result in us re-configuring all plugins every time we add a new plugin?

doc/lightning-plugin.7.txt
```diff
@@ -0,0 +1,48 @@
...
+
+SYNOPSIS
+--------
+*plugin* command [parameter] [second_parameter]
+
+DESCRIPTION
+-----------
+
+The *plugin* RPC command allows to manage plugins without having to restart lightningd.
+It takes 1 to 3 parameters: a command (start/stop/startdir/reload/list) which describes the
```
cdecker: This still calls it `reload` whereas the code below calls it `rescan`

contrib/pylightning/lightning/lightning.py
cdecker: Maybe we should rewrite this into multiple calls that map to the various `plugin` subcommands?

 - plugin_start
 - plugin_stop
 - plugin_rescan
 - etc

That way we have something to attach the specific help messages to instead of handling opaque sets of parameters?

.travis/build.sh
```diff
@@ -10,6 +10,7 @@ export COMPAT=${COMPAT:-1}
 export PATH=$CWD/dependencies/bin:"$HOME"/.local/bin:"$PATH"
 export TIMEOUT=180
 export PYTEST_PAR=2
+export PYTHONPATH=$PWD/contrib/pylightning:$PYTHONPATH
```
cdecker: This has always been working, what is the error you are getting without this?

contrib/pylightning/lightning/plugin.py
cdecker: Should be ok for now, I'll fixup the various plugins in `lightningd/plugins` once this gets merged.

lightningd/plugin.c
```diff
@@ -1014,6 +1019,8 @@ static void plugin_config(struct plugin *plugin)
 	json_add_string(req->stream, "rpc-file", ld->rpc_filename);
 	json_object_end(req->stream);
 
+	json_add_bool(req->stream, "startup", plugin->plugins->startup);
```
cdecker: Adding startup like this will make all existing plugins incompatible since pylightning unpacks the arguments, and adding a new one will cause it to get an unknown argument. Can we make this part of the `configuration`?

lightningd/plugin.c
cdecker: Yeah, that argument unpacking was a bit of a brainfart on my end, but it has been working nicely so far. We should just define a minimal set of top-level objects that need to be passed in and then add new stuff to those. That also makes this semantically clean.

lightningd/plugin.c
darosior: No, 2ba5145a3e460684e5f65e5a09523238cf65dee5 introduces 
```C
	if (p->configured)
		continue;
```

doc/lightning-plugin.7.txt
darosior: Didnt see this one, thanks !

.travis/build.sh
darosior: Dynamically-started plugins don't get the non-exported environment variable passed at the command line

lightningd/plugin.c
darosior: Yes sure, anyway I made a fixup afterwards :-). I can remove the fixup commit and make this field part of the `configuration`.

darosior:
> Yeah, that argument unpacking was a bit of a brainfart on my end, but it has been working nicely so far. We should just define a minimal set of top-level objects that need to be passed in and then add new stuff to those. That also makes this semantically clean.

Should we add this to `init` too ?

lightningd/plugin.c
darosior: Ok, i'll add `startup`to `configuration` then

lightningd/plugin.h
darosior: I think I need it for https://github.com/darosior/lightning/blob/99b4686406d5c79761cd03fe111b1140016afbe7/lightningd/plugin_control.c#L26

darosior:
Thank you for the review @cdecker , I rebased to (mainly) :
- Break Pylightning `pluging` method into one method by subcommand (which is much nicer ! :)
- Move the `init`'s `startup` field to its `configuration` object and thus remove the now useless fixup commit.

contrib/pylightning/lightning/plugin.py
SimonVrouwe: This was commented about earlier, but is it not more conservative to default to `dynamic=False` ?

Or is it now expected that old plugins need to update anyway to this new behavior?
Most of the plugins don't even have tests!

darosior: > This was commented about earlier, but is it not more conservative to default to `dynamic=False` ?
> 
> Or is it now expected that old plugins need to update anyway to this new behavior?
> Most of the plugins don't even have tests!

This was discussed above. This does not change the plugin behavior at all, it just makes it possible for lightningd to restart it : this is completely transparent plugin side (hence the new init field).

darosior: That said, a plugin that is not intended to be stopped at runtime can set dynamic to `False` to ensure that an user won't *inadvertently* stop it via RPC.

SimonVrouwe: > This does not change the plugin behavior at all

Sorry to be blunt, but if the behavior is not changed, then why is there an extra parameter?
What is wrong with defaulting to `dynamic=False` ?

Maybe to you its trivial, but how can we know for sure that current (known or unknown) plugins work properly when re-starting with a running c-lightning?

darosior: > Sorry to be blunt, but if the behavior is not changed, then why is there an extra parameter?
What is wrong with defaulting to dynamic=False ?

This is a new field passed in the `getmanifest` call, [used to give informations to lightningd about plugins](https://lightning.readthedocs.io/PLUGINS.html#the-getmanifest-method). This allows a plugin creator to pass whatever he wants.

> What is wrong with defaulting to dynamic=False ?

Pylightning users not setting it.

> Maybe to you its trivial, but how can we know for sure that current (known or unknown) plugins work properly when re-starting with a running c-lightning?

 Restarting a plugin means stopping and starting it : on the `stop` command, the plugin will be killed. On the `start` command an interface similar (but asynchronous) to which is setup on startup is setup by `lightningd` : a plugin must behave properly at `start` or it means it wouldn't even at startup.
The only case this could lead to an issue is a plugin which has to be loaded (for whatever reason, i.e. backup the first db calls) at startup *only* : that was `myplugin.dynamic = False` is for :-)

SimonVrouwe: Ok thanks for the clarification, now I also read @cdecker's comment:

> Should be ok for now, I'll fixup the various plugins in lightningd/plugins once this gets merged.

darosior:
Rebased

rustyrussell:
> The only really important thing is the new arguments for pylightning which I need to find a fix for.

Would rpc.plugin.start() etc be more pythony?  ie. make plugin like a command namespace... somehow?

darosior:
> > The only really important thing is the new arguments for pylightning which I need to find a fix for.
> 
> Would rpc.plugin.start() etc be more pythony? ie. make plugin like a command namespace... somehow?

I think this was about dynamic json fields unpacking to function arguments which breaks plugins made with Pylightning if we add a new field (in `init` for example :-))

darosior:
Rebased on top of master and resolve CHANGELOG conflicts. Added new brag lines for `init` and `getmanifest` new fields.

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - build: now requires `python3-mako` to be installed, i.e. `sudo apt-get install python3-mako`
 - plugins: a new notification type `invoice_payment` (sent when an invoice is paid) has been added
+- rpc: a new rpc command is added, `plugin`. It allows one to manage plugins without restarting `lightningd`.
+- plugins: a new boolean field is added to the `init`'s `configuration`, `startup`. It allows a plugin to know if it has been started on `lightningd` startup.
+- plugins: a new boolean field can be added to a plugin manifest, `dynamic`. It allows a plugin to tell if it can be started or stopped "on-the-fly".
```
rustyrussell: I'll probably collapse these three lines into "plugins: dynamic plugin support", which is the better summary for this level IMHO.

rustyrussell:
Trivial rebase to resolve CHANGELOG.md conflict...

SimonVrouwe:
Can someone add the `plugin` label to this threat?

Issue #2772, opened 2019-06-25
Signal 6 crash when running listpeers


hsjoberg:
### Issue and Steps to Reproduce

When running `listpeers`, c-lightning crashes with the log showing the following:

```
jun 25 10:05:03 raspberrypi lightningd[11067]: lightningd: bitcoin/tx.c:217: push_tx: Assertion `res == WALLY_OK' failed.
jun 25 10:05:03 raspberrypi lightningd[11067]: lightningd: FATAL SIGNAL 6 (version v0.7.1rc3)
jun 25 10:05:03 raspberrypi lightningd[11067]: 2019-06-25T10:05:03.370Z lightningd(11067): FATAL SIGNAL 6 (version v0.7.1rc3)
jun 25 10:05:07 raspberrypi systemd[1]: lightningd.service: Main process exited, code=killed, status=6/ABRT
jun 25 10:05:07 raspberrypi systemd[1]: lightningd.service: Unit entered failed state.
jun 25 10:05:07 raspberrypi systemd[1]: lightningd.service: Failed with result 'signal'.
```

Apparently an assertion in bitcoin/tx.c fails for some reason:
https://github.com/ElementsProject/lightning/blob/c1328b3fb78aa4fb43ca9dc27106aabe882f50d9/bitcoin/tx.c#L217

### `getinfo` output
```
{
   "id" : "03939ff69d65a13c4bb2585042e7eb7e75a7c77289ab5794d1b973721d86c6839c",
   "alias" : "CocoPi⚡",
   "color" : "f2a900",
   "num_peers" : 14,
   "num_pending_channels" : 0,
   "num_active_channels" : 14,
   "num_inactive_channels" : 0,
   "address" : [
      {
         "type" : "ipv4",
         "address" : "213.114.84.246",
         "port" : 9735
      },
      {
         "type" : "torv2",
         "address" : "hphtfkdy4zpefsnz.onion",
         "port" : 9735
      }
   ],
   "binding" : [
      {
         "type" : "ipv4",
         "address" : "192.168.1.106",
         "port" : 9735
      },
      {
         "type" : "ipv4",
         "address" : "127.0.0.1",
         "port" : 9735
      }
   ],
   "version" : "v0.7.1rc3",
   "blockheight" : 582357,
   "network" : "bitcoin",
   "msatoshi_fees_collected" : 19814,
   "fees_collected_msat" : "19814msat"
}
```

ZmnSCPxj:
[...]

Issue #2759, opened 2019-06-20
listpeers crash lightning node


chek2fire:
### Issue and Steps to Reproduce

listpeers crash my node.
I have reproduce this to three of my lightning nodes.
All of nodes are up and running for one year now with upgrades to newest version time2time

> lightningd: bitcoin/tx.c:215: push_tx: Assertion `res == WALLY_OK && len == written' failed.
lightningd: FATAL SIGNAL 6 (version v0.7.1rc2-10-g92bd3a3-modded)
0x429611 send_backtrace
        common/daemon.c:40
0x42969b crashdump
        common/daemon.c:53
0x7f127918d4af ???
        ???:0
0x7f127918d428 ???
        ???:0
0x7f127918f029 ???
        ???:0
0x7f1279185bd6 ???
        ???:0
0x7f1279185c81 ???
        ???:0
0x436b20 push_tx
        bitcoin/tx.c:215
0x4372c4 bitcoin_txid
        bitcoin/tx.c:266
0x41c52e json_add_channel
        lightningd/peer_control.c:573
0x41cc05 json_add_peer
        lightningd/peer_control.c:1061
0x41d87a json_listpeers
        lightningd/peer_control.c:1093
0x410abb parse_request
        lightningd/jsonrpc.c:654
0x410c9d read_json
        lightningd/jsonrpc.c:752
0x45658f next_plan
        ccan/ccan/io/io.c:59
0x456a4f do_plan
        ccan/ccan/io/io.c:407
0x456a7a io_ready
        ccan/ccan/io/io.c:417
0x458123 io_loop
        ccan/ccan/io/poll.c:445
0x40e82d io_loop_with_timers
        lightningd/io_loop_with_timers.c:24
0x4122c5 main
        lightningd/lightningd.c:827
0x7f127917882f ???
        ???:0
0x403da8 ???
        ???:0
0xffffffffffffffff ???
        ???:0
2019-06-20T14:41:19.602Z lightningd(26528): FATAL SIGNAL 6 (version v0.7.1rc2-10-g92bd3a3-modded)
2019-06-20T14:41:19.602Z lightningd(26528): backtrace: common/daemon.c:45 (send_backtrace) 0x429655
2019-06-20T14:41:19.602Z lightningd(26528): backtrace: common/daemon.c:53 (crashdump) 0x42969b
2019-06-20T14:41:19.602Z lightningd(26528): backtrace: (null):0 ((null)) 0x7f127918d4af
[...]

PR #2773, opened 2019-06-27
Handle zero output commitments

rustyrussell:
Looks like old deployments have some of these, so add a workaround and a check.

Fixes: #2772 
Fixes: #2759

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -340,6 +340,18 @@ register_close_command(struct lightningd *ld,
 		     &close_command_timeout, cc);
 }
 
+static bool invalid_last_tx(const struct bitcoin_tx *tx)
+{
+	/* This problem goes back further, but was discovered just before the
+	 * 0.7.1 release. */
+#ifdef COMPAT_V070
+	/* Old bug had commitment txs with no outputs; bitcoin_txid asserts. */
+	return tx->wtx->num_outputs == 0;
+#else
+	return false;
+#endif
+}
+
 void drop_to_chain(struct lightningd *ld, struct channel *channel,
 		   bool cooperative)
 {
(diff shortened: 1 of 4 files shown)


lightningd/peer_htlcs.c
```diff
@@ -1224,6 +1224,21 @@ static bool changed_htlc(struct channel *channel,
 		return update_in_htlc(channel, changed->id, changed->newstate);
 }
 
+/* FIXME: This should be a complete check, not just a sanity check.
+ * Perhaps that means we need a cookie from the HSM? */
```
ZmnSCPxj: Seems to me we can add the previous commitment TX and see what changed.  HTLCs that have been fulfilled should also include the preimage, new HTLCs we offer should be backed by some proof of `sendpay`-command being executed or an incoming HTLC on a different channel (check same hash and lower locktime of outgoing than incoming).  New HTLCs offered  by the peer should be fine, and failed HTLCs should be fine without further validation.  Fee increases... dunno, we just trust whoever we get our fees from currently, do we need some proof basis for fees?

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2773/commits/a52303f14c99bf5efa96c040dbded38f15b56232

lightningd/peer_control.c
```diff
@@ -355,6 +367,10 @@ void drop_to_chain(struct lightningd *ld, struct channel *channel,
 		log_broken(channel->log,
 			   "Cannot broadcast our commitment tx:"
 			   " they have a future one");
+	} else if (invalid_last_tx(channel->last_tx)) {
+		log_broken(channel->log,
```
ZmnSCPxj: Now that I think more about it.... Should we not at least make an effort to send an `error` message to the peer, so that hopefully the peer can drop to chain for us?  Then maybe forget the channel?

rustyrussell: Yes. We've already sent an error to the peer at this point, which is why we're trying to broadcast the last state, fortunately.

lightningd/peer_htlcs.c
rustyrussell: Yes, this is the realm of the HSM though.  If it has to do these checks anyway, no reason to double-up here.

Issue #2352, opened 2019-02-13
Channel stuck: CHANNELD_AWAITING_LOCKIN


billygarrison:
### Issue and Steps to Reproduce
I had two channel opening transactions stuck in the mempool for the past couple days. When the mempool cleared, both transactions got confirmed in the same block. Only one of the channels actually opened though, the other says the state is "CHANNELD_AWAITING_LOCKIN".

The funding_txid for the channel has [32 confirmations](https://blockstream.info/tx/adb0e33d88aa2c5cb6168ea16cd46f492ee5387f3b4d295bcb2ec09588e6650e) at the time of this post.

`connect` and `ping` both work with the peer_id. And using 1ml.com I can see the peer has had one channel opened since my funding_tx was confirmed.

### `listpeers` output for that peer
```
      "channels": [
        {
          "state": "CHANNELD_AWAITING_LOCKIN",
          "scratch_txid": "04f5e9c4dcd919ac3bfd1571cba6901f5f61c1e5190543c215cc013884899509",
          "owner": "lightning_channeld",
          "short_channel_id": "562892:912:1",
          "channel_id": "0e65e68895c02ecb5b294d3b7f38e52e496fd46ca18e16b65c2caa883de3b0ac",
          "funding_txid": "adb0e33d88aa2c5cb6168ea16cd46f492ee5387f3b4d295bcb2ec09588e6650e",
          "private": false,
          "msatoshi_to_us": 840000000,
          "msatoshi_to_us_min": 840000000,
          "msatoshi_to_us_max": 840000000,
          "msatoshi_total": 840000000,
          "dust_limit_satoshis": 546,
          "max_htlc_value_in_flight_msat": 18446744073709551615,
          "their_channel_reserve_satoshis": 8400,
          "our_channel_reserve_satoshis": 8400,
          "spendable_msatoshi": 831600000,
          "htlc_minimum_msat": 0,
          "their_to_self_delay": 144,
          "our_to_self_delay": 144,
          "max_accepted_htlcs": 483,
          "status": [
            "CHANNELD_AWAITING_LOCKIN:We've confirmed funding, they haven't yet."
          ],
          "in_payments_offered": 0,
          "in_msatoshi_offered": 0,
          "in_payments_fulfilled": 0,
          "in_msatoshi_fulfilled": 0,
          "out_payments_offered": 0,
          "out_msatoshi_offered": 0,
          "out_payments_fulfilled": 0,
[...]

PR #2776, opened 2019-06-28
lightningd: fix occasional missing txid detection.

rustyrussell:
I was working on rewriting our (somewhat chaotic) tx watching code
for 0.7.2, when I found this bug: we don't always notice the funding
tx in corner cases where more than one block is detected at
once.

This is just the one commit needed to fix the problem: it has some
unnecessary changes, but I'd prefer not to diverge too far from my
cleanup-txwatch branch.

Fixes: #2352

### DIFF
--- a/lightningd/watch.c
+++ b/lightningd/watch.c
@@ -63,12 +63,17 @@ struct txwatch {
 
 	/* Transaction to watch. */
 	struct bitcoin_txid txid;
+
+	/* May be NULL if we haven't seen it yet. */
+	const struct bitcoin_tx *tx;
+
 	unsigned int depth;
 
 	/* A new depth (0 if kicked out, otherwise 1 = tip, etc.) */
 	enum watch_result (*cb)(struct lightningd *ld,
 				struct channel *channel,
 				const struct bitcoin_txid *txid,
+				const struct bitcoin_tx *tx,
 				unsigned int depth);
 };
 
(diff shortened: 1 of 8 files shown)


lightningd/watch.c
```diff
@@ -295,3 +305,15 @@ void watch_topology_changed(struct chain_topology *topo)
...
+void txwatch_inform(const struct chain_topology *topo,
+		    const struct bitcoin_txid *txid,
+		    const struct bitcoin_tx *tx_may_steal)
+{
+	struct txwatch *txw;
+
+	txw = txwatch_hash_get(&topo->txwatches, txid);
+
+	if (txw && !txw->tx)
+		txw->tx = tal_steal(txw, tx_may_steal);
```
ZmnSCPxj: In case multiple watchers are watching the same txid (channel factories in the far future, where multiple peer-daemons want to see the confirmation depth of the same tx holding the channel factory), would not this stealing be troublesome? Would not be better to `tal_dup` instead?

rustyrussell: We need a deep copy, hence the FIXME.  libwally doesn't provide a copy function, which would make it easy to implement bitcoin_tx_dup().

ZmnSCPxj: A lousy copy would be to serialize, then deserialize, i.e. the "JSON.parse(JSON.stringify())" trick.

Issue #2779, opened 2019-06-29
Fatal signal 6 after getblock fails.


dan-da:
I'm still having trouble getting lightningd to run reliably against spruned.  Both operating over tor.  Maybe slow tor/network is a root cause, but I don't think lightningd should crash because of it.

What I notice is that:

1. there will be some calls to getblock that fail exiting with status 1.  These seems to be handled ok.

2. The crash occurs after a different log message is printed containing the string "(after xx other errors)".

```
2019-06-29T11:32:16.241Z lightningd(15089): bitcoin-cli -datadir=/home/c-lightning/.bitcoin getblock 000000000000000000153dc4fc7ec2aa952cbcaeecbdb15cf1b5526e1924ca17 false exited with status 1
2019-06-29T11:32:18.330Z lightningd(15089): bitcoin-cli -datadir=/home/c-lightning/.bitcoin getblock 000000000000000000153dc4fc7ec2aa952cbcaeecbdb15cf1b5526e1924ca17 false exited with status 1
2019-06-29T11:32:20.436Z lightningd(15089): bitcoin-cli -datadir=/home/c-lightning/.bitcoin getblock 000000000000000000153dc4fc7ec2aa952cbcaeecbdb15cf1b5526e1924ca17 false exited with status 1
2019-06-29T11:32:22.330Z lightningd(15089): bitcoin-cli -datadir=/home/c-lightning/.bitcoin getblock 000000000000000000153dc4fc7ec2aa952cbcaeecbdb15cf1b5526e1924ca17 false exited 1 (after 16 other errors) 'error code: -1?error message:?Server error?'
2019-06-29T11:32:23.099Z lightningd(15089): FATAL SIGNAL 6 (version v0.7.1rc4)
2019-06-29T11:32:23.099Z lightningd(15089): backtrace: common/daemon.c:45 (send_backtrace) 0x4296b5
2019-06-29T11:32:23.099Z lightningd(15089): backtrace: common/daemon.c:53 (crashdump) 0x4296fb
2019-06-29T11:32:23.099Z lightningd(15089): backtrace: (null):0 ((null)) 0x7f24217fccaf
2019-06-29T11:32:23.099Z lightningd(15089): backtrace: (null):0 ((null)) 0x7f24217fcc37
2019-06-29T11:32:23.099Z lightningd(15089): backtrace: (null):0 ((null)) 0x7f2421800027
[...]

PR #2780, opened 2019-06-29
lightningd/options.c: Add option for setting how long to keep trying bitcoin-cli command.

ZmnSCPxj:
Requested in: #2779

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -793,6 +793,12 @@ void register_opts(struct lightningd *ld)
 	opt_register_arg("--bitcoin-rpcport", opt_set_talstr, NULL,
 			 &ld->topology->bitcoind->rpcport,
 			 "bitcoind RPC port");
+	opt_register_arg("--bitcoin-retry-timeout",
+			 opt_set_u64, opt_show_u64,
+			 &ld->topology->bitcoind->retry_timeout,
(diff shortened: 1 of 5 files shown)


lightningd/options.c
```diff
@@ -793,6 +793,11 @@ void register_opts(struct lightningd *ld)
 	opt_register_arg("--bitcoin-rpcport", opt_set_talstr, NULL,
 			 &ld->topology->bitcoind->rpcport,
 			 "bitcoind RPC port");
+	opt_register_arg("--bitcoin-try-timeout", opt_set_u64, NULL,
```
cdecker: :warning: bikeshedding here :warning: 

I'd prefer `--bitcoin-retry-timeout` since it's not a single attempt but actually counts the time between the first failure and the first success.

ZmnSCPxj:
Previous failures were caused by #2782, now that #2783 has been merged, should be fixed.

ZmnSCPxj:
PING

ZmnSCPxj:
> I like it, but I'd like to see us also hand that argument through to '-rpcclienttimeout' as well, as @cdecker points out. That can obv be a separate commit.

Should we consider the time that has already passed, then subtract that from the configured value, before passing to `-rpcclienttimeout`?

PR #2784, opened 2019-06-30
bitcoin-cli: check chain from bitcoin-cli

### LINKED ISSUE #1074: Double-check chain from bitcoin-cli
Related to: #678 

We should abort early if our net settings do not match the result of `bitcoin-cli getblockchaininfo` . https://bitcoin.org/en/developer-reference#getblockchaininfo

We should check if the resulting `chain` is `"main"` for mainnet, `"test"` for testnet, `"regtest"` for regtest.

trueptolemy:
fixes #1074 
At the beginning of the lightningd, we use `echo` command to check if bitcoin-cli is running.

Now I raplace `echo` with `getblockchaininfo` for this check, and also check whether the `chain` field in response is same as the blockchain that  lightningd is on.

`getblockchaininfo` is valid for litecoin-cli.
Beginning bitcoin-cli with `getblockchaininfo` is also helpful in solving "caught up with network".


lightningd/bitcoind.c
```diff
@@ -111,19 +111,19 @@ static struct io_plan *output_init(struct io_conn *conn, struct bitcoin_cli *bcl
 static void next_bcli(struct bitcoind *bitcoind, enum bitcoind_prio prio);
 
 /* For printing: simple string of args. */
-static char *bcli_args(const tal_t *ctx, struct bitcoin_cli *bcli)
+static char *bcli_args(const tal_t *ctx, const char **args)
```
ZmnSCPxj: I would have instead renamed this to `bcli_args_direct`, then created a new `bcli_args`:

    static char *bcli_args(const tal_t *ctx, struct bitcoin_cli *bcli)
    {
        return bcli_args_direct(ctx, &bcli->args);
    }

This would have reduced the large number of changes later in the file.

A generally good rule is that we should avoid changing the interface of a function, instead change function name when we change interface and create a wrapper that retains previous interface.

ZmnSCPxj:
Is this not, more properly "fixes #1074"?

lightningd/bitcoind.c
trueptolemy: Thank you! Nice idea :)

trueptolemy:
Like @ZmnSCPxj suggested, rename `bcli_args()` to `bcli_args_direct()`, and also add a warper on it(the new `bcli_args()`), to avoid the large number of changes later in the file;
Also edit the description for this PR, "fixes #1074".

lightningd/bitcoind.c
```diff
@@ -814,12 +814,73 @@ static void fatal_bitcoind_failure(struct bitcoind *bitcoind, const char *error_
 	exit(1);
 }
 
+static char* get_blockchainname_for_cli(struct bitcoind *bitcoind)
+{
+	if (!strcmp(bitcoind->chainparams->network_name, "bitcoin")
```
ZmnSCPxj: I believe we have a `streq` macro in `ccan/str/str.h` which matches this better.

lightningd/bitcoind.c
```diff
@@ -814,12 +814,73 @@ static void fatal_bitcoind_failure(struct bitcoind *bitcoind, const char *error_
...
+		      bcli_args_direct(tmpctx, cmd), strerror(errno));
+
+	output_bytes = tal_count(output);
+
+	tokens = json_parse_input(cmd, output, output_bytes,
+			  &valid);
+
+	if (!tokens)
+		return tal_fmt(ctx, "%s: %s response",
+	      bcli_args_direct(tmpctx, cmd),
```
ZmnSCPxj: Strange indent.

lightningd/bitcoind.c
```diff
@@ -814,12 +814,73 @@ static void fatal_bitcoind_failure(struct bitcoind *bitcoind, const char *error_
...
+				char* output, const char **cmd)
+{
+	size_t output_bytes;
+	const jsmntok_t *tokens, *valuetok;
+	char *blockchain;
+	bool valid;
+
+	if (!output)
+		return tal_fmt(ctx, "Reading from %s failed: %s",
+		      bcli_args_direct(tmpctx, cmd), strerror(errno));
```
ZmnSCPxj: Mildly strange indent.

lightningd/bitcoind.c
trueptolemy: Yes, I fixed it.

lightningd/bitcoind.c
trueptolemy: Yes, I corrected it.

lightningd/bitcoind.c
trueptolemy: Corrected.

lightningd/bitcoind.c
```diff
@@ -814,12 +814,73 @@ static void fatal_bitcoind_failure(struct bitcoind *bitcoind, const char *error_
 	exit(1);
 }
 
+static char* get_blockchainname_for_cli(struct bitcoind *bitcoind)
```
ZmnSCPxj: Actually, sorry for the churn -- would it not be better to instead have a `getblockchaininfo_chain` field in `struct chainparams`?

That way if new altcoins want to commit to maintaining a branch of C-lightning with support for their altcoin, they just need to fill in that field according to the `chain` param of their `getblockchaininfo` command.

trueptolemy: Thank you :+1: 
What about `bip70_name`?

ZmnSCPxj: I am uncertain how [BIP-70 Payment Protocol](https://github.com/bitcoin/bips/blob/master/bip-0070.mediawiki) relates?

lightningd/bitcoind.c
```diff
@@ -814,12 +814,58 @@ static void fatal_bitcoind_failure(struct bitcoind *bitcoind, const char *error_
...
+			       bcli_args_direct(tmpctx, cmd),
+			       (int)output_bytes, output);
+
+	valuetok = json_get_member(output, tokens, "chain");
+	if (!valuetok)
+		return tal_fmt(ctx, "%s: had no chain member (%.*s)?",
+			       bcli_args_direct(tmpctx, cmd),
+			       (int)output_bytes, output);
+
+	blockchain = tal_strdup(tmpctx, bitcoind->chainparams->bip70_name);
```
ZmnSCPxj: This `tal_strdup` is not needed.  We could do:

    if (!json_tok_streq(output, valuetok, bitcoind->chainparams->bip70_name))

trueptolemy: Thank you, I corrected.

ZmnSCPxj:
ACK 10b090e

PR #2785, opened 2019-06-30
CHANGELOG.md: Reset for next release.

ZmnSCPxj:


rustyrussell:
Oops!  I just did the same thing in #2787 .  Let's use yours instead, though I'll add the [Unreleased] link...

PR #2786, opened 2019-06-30
Pylightning: allow to set description and long description manually

darosior:
The command description taken from the docstring is pretty handy as a default, but one might want to use it otherwise. This also adds the possibility to specify a long description.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2786/commits/0517f2182ba1957c1552555e4b73928e1811487f

PR #2792, opened 2019-07-03
Log_info on broadcasting error

### LINKED ISSUE #2790: Shouldn't log_unusual for failed broadcast on unilateral close

```
2019-07-03T09:26:52.806Z [UNUSUAL] lightningd(26941): Broadcasting tx 02000...000000: 25 error code: -25?error message:?Missing inputs?
```

Especially since we now tag [UNUSUAL] it's a bit scary-looking!

Reported-by: @JavierRSobrino

tasoshi:
Fix #2790

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -134,13 +134,11 @@ static void broadcast_remainder(struct bitcoind *bitcoind,
 {
 	/* These are expected. */
 	if (strstr(msg, "txn-mempool-conflict")
-	    || strstr(msg, "transaction already in block chain"))
+	    || strstr(msg, "transaction already in block chain")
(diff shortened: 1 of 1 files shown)


rustyrussell:
Actually, I think this could be log_debug for every failure case, avoiding the branch?

I thought about plumbing a flag from the initial broadcast_tx path to this, but this is only the broadcast-retry path: we get a message if we expect success initially and we don't get it, so no point spamming.

tasoshi:
Sorry for the delay !

rustyrussell:
Thanks!

Ack https://github.com/ElementsProject/lightning/commit/9063dc793ea2172553038b73c2ea80f9b5e1783d

PR #2799, opened 2019-07-07
Plugin: Another new notification type, forward_event

trueptolemy:
# Introduction
A notification for topic `forward_event` is sent every time the status of a forward payment is set. The json format is same as the API `listforwards`.
Now when the forwad payment status changes, we'll store forward payments with the new status into DB. This notification will notify subscribers after storing operation.

```json
{
  "forward_event": {
  "payment_hash": "f5a6a059a25d1e329d9b094aeeec8c2191ca037d3f5b0662e21ae850debe8ea2",
  "in_channel": "103x2x1",
  "out_channel": "103x1x1",
  "in_msatoshi": 100001001,
  "in_msat": "100001001msat",
  "out_msatoshi": 100000000,
  "out_msat": "100000000msat",
  "fee": 1001,
  "fee_msat": "1001msat",
  "status": "settled",
  "received_time": 1560696342.368,
  "resolved_time": 1560696342.556
  }
}
```
or

```json
{
  "forward_event": {
  "payment_hash": "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
  "in_channel": "103x2x1",
  "out_channel": "110x1x0",
  "in_msatoshi": 100001001,
  "in_msat": "100001001msat",
  "out_msatoshi": 100000000,
  "out_msat": "100000000msat",
  "fee": 1001,
  "fee_msat": "1001msat",
  "status": "local_failed",
  "failcode": 16392,
  "failreason": "WIRE_PERMANENT_CHANNEL_FAILURE",
  "received_time": 1560696343.052
  }
}
```
#### ~~EDIT: change the "forward_event" fieldname in json to "forward_payment".~~

# Related Changes
- Set the fee to 0 when forward doesn't have out channel with `LOCAL_FAILED` status;
- API: `listforwards` include `payment_hash` field;
- JSON: Warp the process of forward payment json object and use it in notification.

### A Related Question
I am confused about the difference of `struct sha256` and `struct sha256_double`: these two struct are both used to store `payment_hash`, maybe we need a cleanup to unify their usage for `payment_hash`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Lightningd: add support for `signet` networks using the `--network=signet` or `--signet` startup option
 - JSON API: `listfunds` now returns also `funding_output` for `channels`
 - plugins: plugins can now suggest `lightning-cli` default to -H for responses.
+- Plugin: new notification `forward_event` offered/settled/failed/local_failed.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - plugins: a new boolean field can be added to a plugin manifest, `dynamic`. It allows a plugin to tell if it can be started or stopped "on-the-fly".
 - lightningd: check bitcoind version when setup topology and confirm the version not older than v0.15.0.
 - startup: space out reconnections on startup if we have more than 5 peers.
+- JSON API: `listforwards` includes the 'payment_hash' field.
 
 ### Deprecated
 
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -295,6 +295,89 @@ forms:
 `jcon fd <error_fd_to_jsonrpc>:`, `plugin-manager`;
 4. `log` is the context of the original log entry.
 
+#### `forward_event`
+
+A notification for topic `forward_event` is sent every time the status
+of a forward payment is set. The json format is same as the API
+`listforwards`.
+
+```json
+{
+  "forward_event": {
+  "payment_hash": "f5a6a059a25d1e329d9b094aeeec8c2191ca037d3f5b0662e21ae850debe8ea2",
+  "in_channel": "103x2x1",
+  "out_channel": "103x1x1",
+  "in_msatoshi": 100001001,
+  "in_msat": "100001001msat",
+  "out_msatoshi": 100000000,
+  "out_msat": "100000000msat",
+  "fee": 1001,
+  "fee_msat": "1001msat",
+  "status": "settled",
+  "received_time": 1560696342.368,
+  "resolved_time": 1560696342.556
+  }
+}
+```
+or
+
+```json
+{
+  "forward_event": {
+  "payment_hash": "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+  "in_channel": "103x2x1",
+  "out_channel": "110x1x0",
+  "in_msatoshi": 100001001,
+  "in_msat": "100001001msat",
+  "out_msatoshi": 100000000,
+  "out_msat": "100000000msat",
+  "fee": 1001,
+  "fee_msat": "1001msat",
+  "status": "local_failed",
+  "failcode": 16392,
+  "failreason": "WIRE_PERMANENT_CHANNEL_FAILURE",
+  "received_time": 1560696343.052
+  }
+}
+                  
+```
+ - The status includes `offered`, `settled`, `failed` and `local_failed`,
+   and they are all string type in json.
+   - When the forward payment is valid for us, we'll set `offered`
+     and send the forward payment to next hop to resolve;
+   - When the payment forwarded by us gets paid eventually, the forward
+     payment will change the status from `offered` to `settled`;
+   - If payment fails locally(like failing to resolve locally) or the
+     corresponding htlc with next hop fails(like htlc timeout), we will
+     set the status as `local_failed`. `local_failed` may be set before
+     setting `offered` or after setting `offered`. In fact, from the
+     time we receive the htlc of the previous hop, all we can know the
+     cause of the failure is treated as `local_failed`. `local_failed`
+     only occuors locally or happens in the htlc between us and next hop;
+     - If `local_failed` is set before `offered`, this
+       means we just received htlc from the previous hop and haven't
+       generate htlc for next hop. In this case, the json of `forward_event`
+       sets the fields of `out_msatoshi`, `out_msat`,`fee` and `out_channel`
+       as 0;
+       - Note: In fact, for this case we may be not sure if this incoming
+         htlc represents a pay to us or a payment we need to forward.
+         We just simply treat all incoming failed to resolve as 
+         `local_failed`.
+     - Only in `local_failed` case, json includes `failcode` and
+       `failreason` fields;
+   - `failed` means the payment forwarded by us fails in the
+     latter hops, and the failure isn't related to us, so we aren't
+     accessed to the fail reason. `failed` must be set after
+     `offered`.
+     - `failed` case doesn't include `failcode` and `failreason`
+       fields;
+ - `received_time` means when we received the htlc of this payment from
+   the previous peer. It will be contained into all status case;
+ - `resolved_time` means when the htlc of this payment between us and the
+   next peer was resolved. The resolved result may success or fail, so
+   only `settled` and `failed` case contain `resolved_time`;
+ - The `failcode` and `failreason` are defined in [BOLT 4][bolt4-failure-codes].
+
 
 ## Hooks
 
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -514,7 +514,7 @@ static void gossipd_incoming_channels_reply(struct subd *gossipd,
 	details = wallet_invoice_details(info, wallet, invoice);
 
 	response = json_stream_success(info->cmd);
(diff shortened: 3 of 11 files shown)


ZmnSCPxj:
`sha256` is intended to specify the result of `SHA256(x)` while `sha256_double` is intended to specify the result of `SHA256(SHA256(x))`.  Both are used since onchain we use the `SHA256(SHA256(x))` but offchain we use `SHA256(x)` only (if I remember my BOLT specs correctly).

lightningd/peer_htlcs.c
```diff
@@ -2109,32 +2109,29 @@ AUTODATA(json_command, &dev_ignore_htlcs);
...
-		json_add_string(response, "status", forward_status_name(cur->status));
-
-		if(cur->failcode != 0) {
-			json_add_num(response, "failcode", cur->failcode);
-			json_add_string(response, "failreason",
-					 onion_type_name(cur->failcode));
-		}
+	json_object_start(response, fieldname);
+
+	json_add_hex(response, "payment_hash",
```
ZmnSCPxj: Why did you remove the `for` loop in **this** commit?

As-is if we bisect to this commit, compiling will result in failure as `cur` variable does not exist. 

You should probably remove the `for` loop on the *next* commit, and *only* add the `payment_hash` on *this* commit.

trueptolemy: Sorry, I squash the wrong commit.

trueptolemy:
Rebased to fix the wrong content in 'API: 'listforwards' now include 'payment_hash' field' commit.
But now the order of commits present here is [disrupted](https://help.github.com/en/articles/why-are-my-commits-in-the-wrong-order).
The commit 0a5b216 should be the third one(after cdceda3).

lightningd/invoice.c
```diff
@@ -510,7 +510,7 @@ static void gossipd_incoming_channels_reply(struct subd *gossipd,
 	details = wallet_invoice_details(info, wallet, invoice);
 
 	response = json_stream_success(info->cmd);
-	json_add_hex(response, "payment_hash", details->rhash.u.u8,
+	json_add_hex(response, "payment_hash", &details->rhash,
```
ZmnSCPxj: If this is common enough, maybe a botique `json_add_sha256` could be added?

doc/PLUGINS.md
```diff
@@ -247,6 +247,92 @@ forms:
...
+  "payment_hash": "f5a6a059a25d1e329d9b094aeeec8c2191ca037d3f5b0662e21ae850debe8ea2",
+  "in_channel": "103x2x1",
+  "out_channel": "103x1x1",
+  "in_msatoshi": 100001001,
+  "in_msat": "100001001msat",
+  "out_msatoshi": 100000000,
+  "out_msat": "100000000msat",
+  "fee": 1001,
+  "fee_msat": "1001msat",
+  "status": "settled",
```
ZmnSCPxj: The rest of your document refers to `FORWARD_OFFERED`, `FORWARD_SETTLED`, etc.  Which one is actually used on the actual plugin interface?  Please use what is actually visible to plugins, as this is `PLUGINS.md`, and avoid reference to internal defines that plugins need not know about.

lightningd/invoice.c
trueptolemy: Sounds good! What about adding it in anothor PR?
It seems need change many places.

lightningd/notification.c
```diff
@@ -59,4 +62,40 @@ void notify_warning(struct lightningd *ld, struct log_entry *l)
...
+		cur->channel_out = *out->key.channel->scid;
+		cur->msat_out = out->msat;
+		assert(amount_msat_sub(&cur->fee, in->msat, out->msat));
+	} else {
+		cur->channel_out.u64 = 0;
+		cur->msat_out = AMOUNT_MSAT(0);
+		cur->fee = AMOUNT_MSAT(0);
+	}
+	cur->payment_hash = tal(cur, struct sha256_double);
+	cur->payment_hash->sha = in->payment_hash;
```
trueptolemy: @ZmnSCPxj  My thoughts about `struct sha256` and struct `sha256_double` are mainly reflected here.
The same payment hash data, the different structures we used. 
Before storing forward payment information into DB, we used `struct sha256`. And after storing, we used `struct sha256_double`, which is involved in `struct forwarding`:
https://github.com/ElementsProject/lightning/blob/2945b25b57db37dca31c10da34a4b1ca33d697c3/wallet/wallet.h#L189-L198
For another example, we used `struct sha256` for `payment_hash` in `struct wallet_payment`:
https://github.com/ElementsProject/lightning/blob/2945b25b57db37dca31c10da34a4b1ca33d697c3/wallet/wallet.h#L241-L262

Like [BOLT#11](https://github.com/lightningnetwork/lightning-rfc/blob/90faaeda89e34cb5257317981e793b47d6182f6d/11-payment-encoding.md#tagged-fields) said,

> set `payment_hash` to the `SHA2` 256-bit hash of the `payment_preimage` that will be given in return for payment.

Should we change the structure of `payment_hash` in `struct forwarding` to `struct sha256`?

doc/PLUGINS.md
trueptolemy: Thank you, I'll change `FORWARD_SETTLED` to `"settled"`, etc.

lightningd/notification.c
ZmnSCPxj: It does look quite wrong... yes, we should probably fix this in a new PR.

doc/PLUGINS.md
```diff
@@ -251,6 +251,90 @@ forms:
...
 
+#### `forward_event`
+
+A notification for topic `forward_event` is sent every time the status
+of a forward payment is set. The json format is same as the API
+`listforwards`.
+
+```json
+{
+  "forward_event": {
```
darosior: Isn't adding a grouping object of the same name of the notification type a little bit redundant ? I would suggest to prefer
```
notification_name {
    "useful_field_a": "1",
    "useful_field_b": "2"
}
```
over
```
notification_name {
    notification_name {
        "useful_field_a": "1",
        "useful_field_b": "2"
    }
}
```

trueptolemy: Thank you, I'll remove it!

trueptolemy: Sorry, @darosior 
For the notification handle method, we need pass in a lot of parameters, If we remove the inner layer of object structure. This is inconvenient, especially there are some opptional  fields(like `resolved_time`).
For the above consideration, I think removing inner object structure may not be the best solution.
In order to avoid redundant , I‘d better change inner name, like "forward_payment".(I notice `invoice_payment hook` and `openchannel hook` also have the inner object structure).

darosior: Yes and moreover as cdecker pointed out in #2823, adding a field in notifications would break plugins relying on it : we might be better off wrapping all notifications into objects.
About the name redundancy I think that if we keep the wrapping, we should keep the same field name as the notification name for all notifications.

trueptolemy: Sounds good! Reduce confusion.

trueptolemy:
Rebased and handled conflicts.

lightningd/notification.c
```diff
@@ -1,13 +1,16 @@
...
+#include <lightningd/peer_htlcs.h>
 
 const char *notification_topics[] = {
 	"connect",
 	"disconnect",
 	"warning",
 	"invoice_payment",
-	"channel_opened"
+	"channel_opened",
+	"forward_event"
```
rustyrussell: Nit: For future reference, I prefer , on the trailing element, to avoid the extra line change like this.  It's legal since C99 and it's been allowed by compilers for even longer, too.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Lightningd: add support for `signet` networks using the `--network=signet` or `--signet` startup option
 - JSON API: `listfunds` now returns also `funding_output` for `channels`
+- Plugin: new notification `forward_event` to report `FORWARD_OFFERED`, `FORWARD_SETTLED`, `FORWARD_FAILED` and `FORWARD_LOCAL_FAILED` events for forward payments.
```
rustyrussell: Uses old names? But really, CHANGELOG should aim for  one line, so "Plugin: new notification `forward_event` offered/settled/failed/local_failed" is sufficient.

lightningd/notification.c
trueptolemy: I guess you prefer `autodata` to register notification. :)

CHANGELOG.md
trueptolemy: Thank you, it's better description. Corrected.

trueptolemy:
Rebased and corrected the CHANGELOG.
Need another rebase to correct conflict when #2859 merged.

trueptolemy:
Rebased and Resolved conflict.

lightningd/notification.c
```diff
@@ -93,3 +96,39 @@ void notify_channel_opened(struct lightningd *ld, struct node_id *node_id,
...
+
+void notify_forward_event(struct lightningd *ld,
+			  const struct htlc_in *in,
+			  const struct htlc_out *out,
+			  enum forward_status state,
+			  enum onion_type failcode,
+			  struct timeabs *resolved_time)
+{
+	struct jsonrpc_notification *n =
+		jsonrpc_notification_start(NULL, notification_topics[5]);
```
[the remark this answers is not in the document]
trueptolemy: Sorry, I coorected.

ZmnSCPxj:
ACK 52f552f

Fixed merge conflict.

ZmnSCPxj:
ACK d5edafa 

Sorry, temporarily messed up the branch, but should be ok state now.

trueptolemy:
Thank you very much!

PR #2803, opened 2019-07-09
pylightning: Implement the lightning handshake and wire protocol

cdecker:
Simple transcript of the specification in python :-)

Sadly I had to drag in two more dependencies (cryptography and coincurve), I will consolidate on top of a simple secp256k1 FFI wrapper at a later point, but these were the minimal ones I could use atm.


cdecker:
```
----------- coverage: platform linux, python 3.6.7-final-0 -----------
Name                     Stmts   Miss  Cover
--------------------------------------------
lightning/__init__.py        2      0   100%
lightning/lightning.py     316    200    37%
lightning/plugin.py        283    131    54%
lightning/wire.py          259     24    91%
--------------------------------------------
```

`test_listen_connect` contains a minimal example on how to use the library:

```python
from wire import LightningServerSocket, PrivateKey
from binascii import unhexlify
ls_privkey = PrivateKey(unhexlify('1111111111111111111111111111111111111111111111111111111111111111'))
l = LightningServerSocket(ls_privkey)
l.bind(('0.0.0.0', 19735))
l.listen()
c, a = l.accept()
print(c.read_message())
c.send_message(b'\x00\x10\x00\x00\x00\x01\xaa')
```

```python
from wire import connect, PrivateKey, PublicKey
from binascii import unhexlify

ls_privkey = PrivateKey(unhexlify(b'1111111111111111111111111111111111111111111111111111111111111111'))
remote_pubkey = PublicKey(unhexlify(b'03b31e5bbf2cdbe115b485a2b480e70a1ef3951a0dc6df4b1232e0e56f3dce18d6'))

lc = connect(ls_privkey, remote_pubkey, '127.0.0.1', 19375)
lc.send_message(b'\x00\x10\x00\x00\x00\x01\xaa')
print(lc.read_message())
```
(these were running against my own c-lightning daemon, hence the port and node ID mismatch in the scripts).

rustyrussell:
TBH I think this belongs in a separate "protocol" library, rather than pylightning.  

For example, if we were to extend it to generating commitment txs, etc.  It's all outside the scope of talking to c-lightning.  Of course, I'm tempted to steal this as-is for my lightning-rfc test branch.

cdecker:
> TBH I think this belongs in a separate "protocol" library, rather than pylightning.

Good point, we might be better off just creating a separate library and add that to pypi (`pylightning-wire` / `pylightning-proto` maybe?)

darosior:
This is great ! Actually I had planned to do something like this myself to dive deeper into the protocol but you did all the work : tweaking with it will for sure be helpful for a better understanding of the specs :).



> > TBH I think this belongs in a separate "protocol" library, rather than pylightning.
> 
> Good point, we might be better off just creating a separate library and add that to pypi (`pylightning-wire` / `pylightning-proto` maybe?)

Why not extend `pylightning` to be the reference Lightning Network Python library ? We could use just part of it for C-lightning (`from pylightning import Plugin` or `LightningRpc`) and some other Lightning Network applications (or some plugins!) could use other parts (`from pylightning import wire`).

cdecker:
> Why not extend `pylightning` to be the reference Lightning Network Python library ? We could use just part of it for C-lightning (`from pylightning import Plugin` or `LightningRpc`) and some other Lightning Network applications (or some plugins!) could use other parts (`from pylightning import wire`).

I looked into how we can have multiple packages that use the same namespace (`lightning` in our case), but apparently having namespace-top-level code prevents us from having a second module that then also uses the same prefix. So either we break  the `pylightning` into `lightning.rpc` and `lightning.plugin`, which would then allow us to have `lightning.network` for the wire implementation, or we create a new namespace and map things in there, or we create a completely new prefix.

darosior:
> or we create a new namespace and map things in there

Maybe `lightning` and `clightning` ? Otherwise the first solution is good if you prefer to not put C-lightning aside.

contrib/pyln-client/README.md
```diff
@@ -0,0 +1,101 @@
+# pylightning: A python client library for lightningd
```
darosior: `s/pylightning/pyln-client/` ?

contrib/pyln-proto/examples/connect.py
```diff
@@ -0,0 +1,23 @@
...
+from binascii import unhexlify, hexlify
+
+ls_privkey = PrivateKey(unhexlify(
+    b'1111111111111111111111111111111111111111111111111111111111111111'
+))
+remote_pubkey = PublicKey(unhexlify(
+    b'03b31e5bbf2cdbe115b485a2b480e70a1ef3951a0dc6df4b1232e0e56f3dce18d6'
+))
+
+lc = connect(ls_privkey, remote_pubkey, '127.0.0.1', 19375)
```
darosior: `9735` ? Or is it intended ?

cdecker: I was using a non-standard port for testing. Will change.

cdecker:
- Added an alias for `pylightning` in `pyln-client` with the correct module prefix (`pyln.client`). `pylightning` remains fully functional for now, in order to maintain backward compatibility
 - Added the wire package in `pyln-proto` (import path `pyln.proto.wire`) as proposed by @rustyrussell 

I haven't published the modules on pypi yet (`pip install` will work once we do), but better be sure that this is what we want. Also I didn't migrate the testing-related things into `pyln-testing` yet, since I need to figure out where to make the cut between library and c-lightning testing specific things.

Ping @rustyrussell @niftynei

contrib/pyln-proto/tests/test_wire.py
```diff
@@ -0,0 +1,189 @@
+from binascii import hexlify, unhexlify
+from lightning.wire import PrivateKey, PublicKey, LightningConnection
```
ysangkok: shouldn't this be `pyln.proto.wire`?

contrib/pylightning/lightning/wire.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+class LightningServerSocket(socket.socket):
+    def __init__(self, local_privkey):
+        socket.socket.__init__(self)
+        self.local_privkey = local_privkey
+        self.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+
+    def accept(self):
+        print("Accepting a new socket")
```
ysangkok: shouldn't this use a logging framework?

contrib/pyln-client/requirements.txt
```diff
@@ -0,0 +1 @@
+pylightning==0.0.7.3
```
ysangkok: should also require `cryptography`

contrib/pyln-proto/tests/test_wire.py
```diff
@@ -0,0 +1,189 @@
...
+    msg = unhexlify(b'68656c6c6f')
+    lc.send_message(msg)
+    m = c2.recv(18 + 21)
+    assert(hexlify(m) == b'cf2b30ddf0cf3f80e7c35a6e6730b59fe802473180f396d88a8fb0db8cbcf25d2f214cf9ea1d95')
+
+    # Send 498 more messages, to get just below the switch threshold
+    for i in range(0, 498):
+        lc.send_message(msg)
+        m = c2.recv(18 + 21)
+    # Check the last message against the test vector
```
ysangkok: this is not actually checking a message, but the `sk` key.

contrib/pyln-proto/pyln/proto/wire.py
```diff
@@ -0,0 +1,395 @@
...
+            mc = encryptWithAD(self.sk, self.nonce(self.sn + 1), b'', m)
+            self.sn += 2
+            self.connection.send(lc)
+            self.connection.send(mc)
+            assert(self.sn % 2 == 0)
+            self._maybe_rotate_keys()
+
+    def _maybe_rotate_keys(self):
+        if self.sn == 1000:
+            self.sck, self.sk = hkdf_two_keys(salt=self.sck, ikm=self.sk)
```
ysangkok: only the tests `rotation` and `shake` set `self.sck`. if i expand `test_handshake` to send more, i get an exception as `self.sck` is not set.

contrib/pylightning/setup.py
```diff
@@ -6,6 +6,9 @@
 with io.open('README.md', encoding='utf-8') as f:
     long_description = f.read()
 
+with io.open('requirements.txt', encoding='utf-8') as f:
+    requirements = [r for r in f.read().split('\n') if len(r)]
+
 setup(name='pylightning',
       version=lightning.__version__,
```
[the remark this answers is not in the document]
cdecker: Seems correct to me, this is still the `pylightning` package, which contains the `lightning` module, which in turn is where the version is stored.

contrib/pyln-proto/pyln/proto/wire.py
cdecker: I didn't intend for `test_handshake` to leave `lc1` and `lc2` in a usable state, since they just step through the test vector to pin-point an eventual regression. I added the initialization of `sck` and `rck` anyway :-)

cdecker:
I think I addressed all the things that @ysangkok's thorough review unearthed. Thanks for the time you invested @ysangkok.

Ping @rustyrussell and @darosior for final compliance tests :wink:

cdecker:
Thanks @darosior, I have no immediate plan to re-implement bolts 02 and 03, so go for it :-)

I added a comment in the example and fixed a spurious empty line, making `make check-python` happy again.

PR #2804, opened 2019-07-12
pytest: Strengthen the htlc_accepted tests

cdecker:
We were having a few issues with malformed data in the past, so this time we
really check that stuff.

### DIFF
--- a/tests/plugins/hold_htlcs.py
+++ b/tests/plugins/hold_htlcs.py
@@ -8,16 +8,26 @@ settled/forwarded/
 
 
 from lightning import Plugin
+import json
+import os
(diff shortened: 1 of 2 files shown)


cdecker:
> Should it not match `Onion written to \(.*\)$` instead? The details of tempfile, as well as various OSs, might change.

Sure we can fix that up :+1:

cdecker:
Changed the regex as suggested by @ZmnSCPxj, re-applying prior ACK

ACK https://github.com/ElementsProject/lightning/pull/2804/commits/3d595eb60ab97f78550f461b59de5faab011dc08

PR #2811, opened 2019-07-14
Fix typos and options

jooray:
--proxy= instead of --proxy:
--always-use-proxy needs value (true)


rustyrussell:
I'm really surprised that --always-use-proxy takes an argument, but good catch!

PR #2814, opened 2019-07-16
param: implement helpers for multiplex commands.

rustyrussell:
Our previous param support was a bit limited in this case, and we want it for #2771 

We replace some dev- commands with a multiplexer, so we can exercise it

### DIFF
--- a/common/param.c
+++ b/common/param.c
@@ -263,6 +263,50 @@ static struct command_result *param_arr(struct command *cmd, const char *buffer,
 			    "Expected array or object for params");
 }
 
+#include <stdio.h>
+
+const char *param_subcommand(struct command *cmd, const char *buffer,
+			     const jsmntok_t tokens[],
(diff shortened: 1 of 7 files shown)


common/param.c
```diff
@@ -263,6 +263,50 @@ static struct command_result *param_arr(struct command *cmd, const char *buffer,
...
 
+#include <stdio.h>
+
+const char *param_subcommand(struct command *cmd, const char *buffer,
+			     const jsmntok_t tokens[],
+			     const char *name, ...)
+{
+	va_list ap;
+	struct param *params = tal_arr(cmd, struct param, 0);
+	const char *arg, **names = tal_arr(tmpctx, const char *, 1);
```
ZmnSCPxj: Mildly disagree with allocating a `names` array but that is just me being too much of a premature optimizer.

PR #2815, opened 2019-07-17
bugfix: fix uninitialized variable compiler errors

benthecarman:
Was unable to build the project without doing this

### DIFF
--- a/ccan/ccan/tal/path/path.c
+++ b/ccan/ccan/tal/path/path.c
@@ -302,7 +302,7 @@ fail_take_to:
 
  char *path_readlink(const tal_t *ctx, const char *linkname)
  {
-	ssize_t len, maxlen = 64; /* good first guess. */
+	ssize_t len = 0, maxlen = 64; /* good first guess. */
 	char *ret = NULL;
 
 	if (unlikely(!linkname) && is_taken(linkname))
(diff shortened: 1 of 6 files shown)


benthecarman:
I am on the most recent version of gcc and still unable to build.  I get errors saying this
```
ccan/ccan/tal/path/path.c: In function ‘path_readlink’:
ccan/ccan/tal/path/path.c:325:6: error: ‘len’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
   ret[len] = '\0';
Makefile:586: recipe for target 'ccan-tal-path.o' failed

```

rustyrussell:
What gcc version, what optimization flags, are you using --enable-developer? Was this a clean build?

We are supposed to ignore those warnings when using anything other than -O3 or no optimization at all...

benthecarman:
`gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0`

I am just pulling the repo and then using `make`

rustyrussell:
> `gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0`
> 
> I am just pulling the repo and then using `make`

Well, you must have run `configure` at some stage, so my guess is you have old warning flags in there.

Try running './configure' again?

benthecarman:
> Try running './configure' again?

Oh wow I am just an idiot, that fixed it. Normally I don't need to do that.

PR #2816, opened 2019-07-18
add signet support

kallewoof:
This PR adds support for signet networks.

Blocked by: https://github.com/bitcoin/bitcoin/pull/16411

See also: https://en.bitcoin.it/wiki/Signet

### DIFF
--- a/bitcoin/chainparams.c
+++ b/bitcoin/chainparams.c
@@ -9,6 +9,8 @@
 #define BIP32_VER_MAIN_PRIVATE 0x0488ADE4
 #define BIP32_VER_TEST_PUBLIC  0x043587CF
 #define BIP32_VER_TEST_PRIVATE 0x04358394
+#define BIP32_VER_SIGT_PUBLIC  0x043587CF
+#define BIP32_VER_SIGT_PRIVATE 0x04358394
 
 const struct chainparams networks[] = {
     {.network_name = "bitcoin",
@@ -48,6 +50,21 @@ const struct chainparams networks[] = {
      .p2sh_version = 196,
      .testnet = true,
      .bip32_key_version = {.bip32_pubkey_version = BIP32_VER_TEST_PUBLIC, .bip32_privkey_version = BIP32_VER_TEST_PRIVATE}},
+    {.network_name = "signet",
+     .bip173_name = "sb",
+     .bip70_name = "signet",
+     .genesis_blockhash = {{{.u.u8 = {0x2d, 0x09, 0x8b, 0x08, 0x55, 0x1b, 0xf3, 0xda, 0xb1, 0xa2, 0x72, 0xa8, 0x71, 0x4a, 0x12, 0x9d, 0x06, 0x5c, 0xfb, 0x32, 0xfa, 0x7e, 0x18, 0xdf, 0x00, 0x99, 0x09, 0x5f, 0xbc, 0xd8, 0x53, 0x62}}}},
+     .rpc_port = 38332,
+     .cli = "bitcoin-cli",
+     .cli_args = "-signet",
+     .dust_limit = { 546 },
+     .max_funding = AMOUNT_SAT_INIT((1 << 24) - 1),
+     .max_payment = AMOUNT_MSAT_INIT(0xFFFFFFFFULL),
+     .when_lightning_became_cool = 1,
+     .p2pkh_version = 125,
+     .p2sh_version = 87,
+     .testnet = true,
+     .bip32_key_version = {.bip32_pubkey_version = BIP32_VER_SIGT_PUBLIC, .bip32_privkey_version = BIP32_VER_SIGT_PRIVATE}},
     {.network_name = "testnet",
      .bip173_name = "tb",
      .bip70_name = "test",
--- a/bitcoin/chainparams.c
+++ b/bitcoin/chainparams.c
@@ -123,3 +123,11 @@ const struct chainparams *chainparams_by_bip173(const char *bip173_name)
 	}
 	return NULL;
 }
+
+const char *chainparams_get_network_names(const tal_t *ctx)
+{
+    char *networks_string = tal_strdup(ctx, networks[0].network_name);
+    for (size_t i = 1; i < ARRAY_SIZE(networks); ++i)
+        tal_append_fmt(&networks_string, ", %s", networks[i].network_name);
+    return networks_string;
+}
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2239,12 +2239,12 @@ bool wallet_network_check(struct wallet *w,
 					   "match network blockchain hash: %s "
 					   "!= %s. "
 					   "Are you on the right network? "
-					   "(--network={bitcoin, testnet, regtest, "
-					   "litecoin or litecoin-testnet})",
+					   "(--network={one of %s})",
 				   type_to_string(w, struct bitcoin_blkid,
 						  &chainhash),
 				   type_to_string(w, struct bitcoin_blkid,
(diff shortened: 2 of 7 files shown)


cdecker:
Awesome, thanks for the PR @kallewoof. I think this is a good addition once bitcoin/bitcoin#16411 is merged ^^

wallet/wallet.c
```diff
@@ -2239,7 +2239,7 @@ bool wallet_network_check(struct wallet *w,
 					   "match network blockchain hash: %s "
 					   "!= %s. "
 					   "Are you on the right network? "
-					   "(--network={bitcoin, testnet, regtest, "
+					   "(--network={bitcoin, testnet, signet, regtest, "
```
ZmnSCPxj: Maybe we should expose a new function in `bitcoind/chainparams.c` that returns a comma-separated string of networks we support, derived from the array of supported networks? That lets us not have to keep updating this as we add new networks or remove networks, just the chainparams table.

kallewoof: I did this, but I am not sure how to trigger that function easily. See 0d8d59243b3df51564f25a0748a67f78fdd02e42

bitcoin/chainparams.c
```diff
@@ -123,3 +124,20 @@ const struct chainparams *chainparams_by_bip173(const char *bip173_name)
...
+    static char* nwstr = NULL;
+    if (!nwstr) {
+        size_t len = (ARRAY_SIZE(networks) << 1) - 1; // 1 too many ", " counted -> 1 byte extra for final \0
+        for (size_t i = 0; i < ARRAY_SIZE(networks); ++i) {
+            len += strlen(networks[i].network_name);
+        }
+        nwstr = malloc(len);
+        char* ptr = nwstr + sprintf(nwstr, "%s", networks[0].network_name);
+        for (size_t i = 1; i < ARRAY_SIZE(networks); ++i) {
+            ptr += sprintf(ptr, ", %s", networks[i].network_name);
```
ZmnSCPxj: Use `tal_fmt` from https://github.com/ElementsProject/lightning/blob/master/ccan/ccan/tal/str/str.h instead.  `#include<ccan/tal/str/str.h>`.

https://github.com/ElementsProject/lightning/blob/5179025977b1179ba78f5c1c27cc95c99eb17543/ccan/ccan/tal/str/str.h#L35-L45

Have `chainparams_networks` accept a `const tal_t *ctx` argument, which is what you pass as first argument to `tal_fmt`.

Use `tal_append_fmt` to append.

So something like:

    networks = tal_fmt(ctx, "%s", networks[0].network_name);
    for (i = 0; i < ARRAY_SIZE(networks); ++i)
        networks = tal_append_fmt(ctx, take(networks), ", %s", networks[i].network_name);
    return networks;

Don't bother with keeping a `static` var, just always allocate a new string, this is not performance-critical code like signing or routefinding.

ZmnSCPxj: Nitpick: `s/chainparams_networks/chainparams_get_network_names/` maybe?

kallewoof: Updated the code to use `tal_t`. (Neat library!)

bitcoin/chainparams.c
```diff
@@ -123,3 +123,12 @@ const struct chainparams *chainparams_by_bip173(const char *bip173_name)
 	}
 	return NULL;
 }
+
+const char *chainparams_get_network_names(const tal_t *ctx)
+{
+    char* networks_string = tal_fmt(ctx, "%s", networks[0].network_name);
```
ZmnSCPxj: Sorry for the flip-flop, but I realized just now that this could be simpler done as `tal_strdup(ctx, networks[0].network_name)`.

`tal_append_fmt` for the loop is fine.

Also, nitpick, it would be nice to `STATIC_ASSERT(ARRAY_SIZE(networks) != 0);` before this, but then we would not be supporting any networks in that case and be useless, so maybe pointless to do that, you can probably ignore this bit.

Nitpick:  we use `char *networks_string` not `char* networks_string` in this project (I agree with the latter personally, but this project is pretty consistent with binding asterisks to the variable name).

bitcoin/chainparams.c
```diff
@@ -123,3 +123,12 @@ const struct chainparams *chainparams_by_bip173(const char *bip173_name)
...
+    for (size_t i = 1; i < ARRAY_SIZE(networks); ++i) {
+        tal_append_fmt(&networks_string, ", %s", networks[i].network_name);
+    }
```
ZmnSCPxj: For single-statement constructs we consistently omit the braces, so:

    for (...)
       statement;
    statement;

wallet/wallet.c
```diff
@@ -2235,16 +2235,18 @@ bool wallet_network_check(struct wallet *w,
...
 					   "Are you on the right network? "
-					   "(--network={bitcoin, testnet, regtest, "
-					   "litecoin or litecoin-testnet})",
+					   "(--network={one of %s})",
 				   type_to_string(w, struct bitcoin_blkid,
 						  &chainhash),
 				   type_to_string(w, struct bitcoin_blkid,
-						  &chainparams->genesis_blockhash));
+						  &chainparams->genesis_blockhash),
+				   chainparams_get_network_names(ctx));
```
ZmnSCPxj: Use `tmpctx` instead here, it should be in scope if you `#include<common/utils.h>` (which is probably already included in this file somewhere, basically just check if `tmpctx` is used anywhere else).

This is a temporary context that is automatically freed in the mainloop, rather than you having to manually create and free one every time you need it.

kallewoof:
@ZmnSCPxj Thanks for the review, I believe I addressed everything you pointed at.

bitcoin/chainparams.c
```diff
@@ -48,6 +50,21 @@ const struct chainparams networks[] = {
...
+     .rpc_port = 38332,
+     .cli = "bitcoin-cli",
+     .cli_args = "-signet",
+     .dust_limit = { 546 },
+     .max_funding = AMOUNT_SAT_INIT((1 << 24) - 1),
+     .max_payment = AMOUNT_MSAT_INIT(0xFFFFFFFFULL),
+     .when_lightning_became_cool = 1,
+     .p2pkh_version = 125,
+     .p2sh_version = 87,
+     .testnet = false,
```
niftynei: I'd argue that we should set this to true. see `lightningd/bitcoin.c:380-387` and `lightningd/options.c:685--688` for rationale.

kallewoof:
@niftynei Thanks! I switched `.testnet` to `true` and updated comment in `jsonrpc.c`.

ZmnSCPxj:
ACK d12c329

ZmnSCPxj:
Were we not supposed to hold off on this until https://github.com/bitcoin/bitcoin/pull/16411 was merged?

kallewoof:
Ideally, but it's not harmful or causing weird dependency issues. Only thing is people will have to run a custom build of bitcoin core to use this, at the moment.

ZmnSCPxj:
I am more concerned that changes to the bitcoin PR might require changes to this PR, but I suppose the upstream PR is now mostly settled.

kallewoof:
Ahh, that is a good point.

niftynei:
Yeah ideally this would have been held til the bitcoind PR was merged; I
wasn’t aware of the status of the upstream dependency til @cdecker pointed
it out (after it was merged).

It’s also missing a CHANGELOG entry 🙃

kallewoof:
Sad, but if you wish, you can revert this PR and I will make a new one once the bitcoin core PR has been merged. If not, I will make sure that any changes done on bitcoin side are duplicated here, in a follow-up PR.

PR #2821, opened 2019-07-21
Wire in TLV's

niftynei:
Wire in @rustyrussell's TLV parsing code to the bolt generators, so we can generate bolts in real-time:tm:


devtools/Makefile
```diff
@@ -38,32 +38,30 @@ devtools/gen_print_onion_wire.h: $(DEVTOOL_BOLT_DEPS) wire/gen_onion_wire_csv
...
+devtools/onion: $(DEVTOOLS_OBJS) $(DEVTOOLS_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS) $(BITCOIN_OBJS) wire/fromwire.o wire/towire.o wire/tlvstream.o devtools/onion.o common/sphinx.o
 
-devtools/gossipwith: $(DEVTOOLS_OBJS) $(DEVTOOLS_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS) $(BITCOIN_OBJS) wire/fromwire.o wire/towire.o wire/gen_peer_wire.o devtools/gossipwith.o common/cryptomsg.o common/cryptomsg.o common/crypto_sync.o
+devtools/gossipwith: $(DEVTOOLS_OBJS) $(DEVTOOLS_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS) $(BITCOIN_OBJS) wire/fromwire.o wire/towire.o wire/tlvstream.o wire/gen_peer_wire.o devtools/gossipwith.o common/cryptomsg.o common/cryptomsg.o common/crypto_sync.o
 
 $(DEVTOOLS_OBJS) $(DEVTOOLS_TOOL_OBJS): wire/wire.h devtools/gen_print_wire.h devtools/gen_print_onion_wire.h
 devtools/gen_print_wire.o: devtools/gen_print_wire.h wire/gen_peer_wire.h devtools/print_wire.h
 devtools/gen_print_onion_wire.o: devtools/gen_print_onion_wire.h devtools/print_wire.h
 
-devtools/bolt11-cli: $(DEVTOOLS_OBJS) $(DEVTOOLS_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS) $(BITCOIN_OBJS) wire/fromwire.o wire/towire.o devtools/bolt11-cli.o
```
niftynei: Note: this was a duplicate of the same entry above. removed this one.

tools/gen/impl_template
```diff
@@ -105,12 +112,15 @@ ${static}bool fromwire_${subtype.name}(${'const tal_t *ctx, ' if subtype.has_len
 	%endif
 	% endfor
 
-	return cursor != NULL;
+	return ${subtype.name};
```
cdecker: This seems strange: isn't this supposed to return a `bool`?

tools/generate-bolts.py
```diff
@@ -102,12 +105,12 @@ def __init__(self):
         self.optional_fields = False
         self.len_fields = {}
 
-    def add_data_field(self, field_name, type_obj, count=1, is_optional=[]):
+    def add_data_field(self, field_name, type_obj, count=1, is_optional=[], comments=[]):
```
cdecker: Having arrays as default parameters in python functions is a Bad Idea :tm:: they are allocated statically and so anything that modifies them inside the function itself will cause them to be modified for all calls:

```python

def test(arg=[]):
    arg.append(1)
    print(arg)

test()
test()
```

will result in the following output:

```sh
[1]
[1, 1]
```

It's not a big issue since we use them here in a read-only way, but something to keep in mind.

tools/generate-bolts.py
```diff
@@ -197,9 +200,10 @@ def __init__(self, name):
         self.depends_on = {}
         # FIXME: internal msgs can be enums
         self.is_enum = False
+        self.type_comments = []
 
-    def add_data_field(self, field_name, type_obj, count=1, is_optional=[]):
-        FieldSet.add_data_field(self, field_name, type_obj, count, is_optional)
+    def add_data_field(self, field_name, type_obj, count=2, is_optional=[], comments=[]):
```
cdecker: Just for my own understanding: what is `count` used for, and why is it being bumped here?

tools/gen/impl_template
niftynei: there's two different 'types' of subtype method declarations: 
  - a 'varsize'd subtype that allocates the subtype itself and returns it. this belongs to that 'type' of subtype fromwire method. a type that has variable length fields is considered 'varsized'. This mimics how the `varlen_structs` in `generate-wire.py` originally work.
 - a 'normal' subtype (no internal allocations). the method signature includes a pointer to the already allocated struct for that subtype. it returns nothing. This mimics the way that other, manually defined subtypes work.

Ostensibly, I could have used the same format for both of them, but chose to copy the existing handwritten parsing format.

tools/generate-bolts.py
niftynei: good to know. i think i handled this correctly then for the comment sets (passing a copy of the list down instead of the list itself).

tools/generate-bolts.py
niftynei: This is a mistake, it should be 1. It's used to mark whether or not a field is variable-length.

niftynei: > I got lost a couple of times trying to reverse engineer the code (python code in particular and some of the accessors whose use I couldn't understand). It'd be nice to have more comments explaining the rationale behind the individual parts

Agreed that more comments is probably useful. I'd also recommend checking out the generated test case code; it's fairly well documented wrt the scope of what all the cases that the generators cover. 

    make check-tools

will generate four files (and sort of check that they compile): `tools/test/gen_wire.[c,h]` and `tools/test/gen_print.[c,h]`

niftynei:
> I got lost a couple of times trying to reverse engineer the code (python code in particular and some of the accessors whose use I couldn't understand). It'd be nice to have more comments explaining the rationale behind the individual parts

Agreed that more comments is probably useful. I'd also recommend checking out the generated test case code; it's fairly well documented wrt the scope of what all the cases that the generators cover. 

    make check-tools

will generate four files (and sort of check that they compile): `tools/test/gen_wire.[c,h]` and `tools/test/gen_print.[c,h]`

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2821/commits/ded3cf073dec1f75cb787e09265cf30b5e355f07

PR #2822, opened 2019-07-21
Plugins: a notification for invoice payments

darosior:
As discussed on IRC, a plugin might want to be notified on invoice payment without needing `lightningd` to wait for his response.
This PR adds an `invoice_payment` notification type similar to the same named hook.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -223,6 +223,21 @@ to a peer was lost.
 }
 ```
 
+#### `invoice_payment`
+
+A notification for topic `invoice_payment` is sent every time an invoie is paid.
+
+```json
+{
+	"invoice_payment": {
+		"label": "unique-label-for-invoice",
+		"preimage": "0000000000000000000000000000000000000000000000000000000000000000",
+		"msat": "10000msat"
+	}
+}
+```
+
+
 #### `warning`
 
 A notification for topic `warning` is sent every time a new `BROKEN`
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -1,7 +1,4 @@
 #include "invoice.h"
-#include "json.h"
-#include "jsonrpc.h"
-#include "lightningd.h"
 #include <bitcoin/address.h>
 #include <bitcoin/base58.h>
 #include <bitcoin/script.h>
@@ -25,7 +22,11 @@
 #include <inttypes.h>
 #include <lightningd/channel.h>
 #include <lightningd/hsm_control.h>
+#include <lightningd/json.h>
+#include <lightningd/jsonrpc.h>
+#include <lightningd/lightningd.h>
 #include <lightningd/log.h>
+#include <lightningd/notification.h>
 #include <lightningd/options.h>
 #include <lightningd/peer_control.h>
 #include <lightningd/peer_htlcs.h>
@@ -172,6 +173,10 @@ invoice_payment_hook_cb(struct invoice_payment_hook_payload *payload,
 	struct lightningd *ld = payload->ld;
 	struct invoice invoice;
 	enum onion_type failcode;
+
+	/* We notify here to benefit from the payload and because the hook callback is
+	 * called even if the hook is not registered. */
+	notify_invoice_payment(ld, payload->msat, payload->preimage, payload->label);
 
 	tal_del_destructor2(payload->hin, invoice_payload_remove_hin, payload);
 	/* We want to free this, whatever happens. */
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -1,11 +1,12 @@
-#include "lightningd/notification.h"
 #include <ccan/array_size/array_size.h>
 #include <lightningd/json.h>
+#include <lightningd/notification.h>
 
 const char *notification_topics[] = {
 	"connect",
 	"disconnect",
-	"warning"
+	"warning",
+	"invoice_payment"
 };
 
 bool notifications_have_topic(const char *topic)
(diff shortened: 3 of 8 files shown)


lightningd/notification.c
```diff
@@ -59,4 +60,17 @@ void notify_warning(struct lightningd *ld, struct log_entry *l)
 	json_object_end(n->stream); /* .warning */
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
-}
\ No newline at end of file
+}
+
+void notify_invoice_payment(struct lightningd *ld, struct amount_msat amount,
+						    struct preimage preimage, const struct json_escape *label)
```
trueptolemy: It looks like a messy format here.

lightningd/notification.h
```diff
@@ -14,4 +16,7 @@ void notify_disconnect(struct lightningd *ld, struct node_id *nodeid);
 
 void notify_warning(struct lightningd *ld, struct log_entry *l);
 
+void notify_invoice_payment(struct lightningd *ld, struct amount_msat amount,
+						    struct preimage preimage, const struct json_escape *label);
```
trueptolemy: Also here.

lightningd/test/run-invoice-select-inchan.c
```diff
@@ -287,6 +287,10 @@ void notify_connect(struct lightningd *ld UNNEEDED, struct node_id *nodeid UNNEE
 /* Generated stub for notify_disconnect */
 void notify_disconnect(struct lightningd *ld UNNEEDED, struct node_id *nodeid UNNEEDED)
 { fprintf(stderr, "notify_disconnect called!\n"); abort(); }
+/* Generated stub for notify_invoice_payment */
+void notify_invoice_payment(struct lightningd *ld UNNEEDED, struct amount_msat amount UNNEEDED,
+						    struct preimage preimage UNNEEDED, const struct json_escape *label UNNEEDED)
```
trueptolemy: And here.

darosior: As indicated by the comment above, I did not write that.

lightningd/notification.c
darosior: indeed, it's hard to align with tabs..

darosior:
Thank you for the review @trueptolemy

darosior:
Thanks for the review @trueptolemy , I've corrected the indentation

lightningd/invoice.c
```diff
@@ -267,6 +268,7 @@ void invoice_try_pay(struct lightningd *ld,
 
 	log_debug(ld->log, "Calling hook for invoice '%s'", details->label->s);
 	plugin_hook_call_invoice_payment(ld, payload, payload);
+	notify_invoice_payment(ld, payload->msat, payload->preimage, payload->label);
```
cdecker: At this point we no longer have ownership over the payload, so maybe it's better to switch the two calls, or to create a copy.

lightningd/notification.c
```diff
@@ -59,4 +60,17 @@ void notify_warning(struct lightningd *ld, struct log_entry *l)
...
-}
\ No newline at end of file
+}
+
+void notify_invoice_payment(struct lightningd *ld, struct amount_msat amount,
+			    struct preimage preimage, const struct json_escape *label)
+{
+	struct jsonrpc_notification *n =
+		jsonrpc_notification_start(NULL, notification_topics[3]);
+	json_add_string(n->stream, "msat",
```
cdecker: Might be better to wrap this in an object, otherwise the pylightning code will need to update in sync with the changes to the arguments (it unpacks the arguments into a function call).

darosior:
Thanks @cdecker, I updated to wrap the notification in a object of the same name as the notification and switched the notification and hook call.

lightningd/invoice.c
```diff
@@ -265,6 +266,8 @@ void invoice_try_pay(struct lightningd *ld,
 	payload->hin = hin;
 	tal_add_destructor2(hin, invoice_payload_remove_hin, payload);
 
+	notify_invoice_payment(ld, payload->msat, payload->preimage, payload->label);
+
 	log_debug(ld->log, "Calling hook for invoice '%s'", details->label->s);
 	plugin_hook_call_invoice_payment(ld, payload, payload);
```
ZmnSCPxj: Should we perform the notify * before* the plugin-hook?

The plugin-hook may (in theory) tell the `lightningd` to reject the payment, so we would end up notifying an invoice payment that does not push through.  Would not it be better to notify *after* the plugin-hook returns with an acceptance?

darosior: Actually that was exactly why I put the hook call before in the first place.

Since, as pointed out by cdecker above, we don't have ownership on `payload` when passed to notify after the hook. To avoid a copy I assumed a plugin which has subsribed to the `invoice_payment` hook won't rely on the (or even subscribe to)  `invoice_payment` notification to know if a payment arrived.

cdecker: Maybe we can put the notification call into the hook callback? That should get back a copy of the payload as well.

darosior: > Maybe we can put the notification call into the hook callback? That should get back a copy of the payload as well.

Wouldn't that condition the notification to be sent only if the hook is registered too ?

ZmnSCPxj: No, the hook callback is always called.  If there is no plugin registered, it gets a `buffer` of `NULL`, which it should always check for.

darosior:
Rebased on top of master and moved the `notify` to the hook callback

lightningd/invoice.c
```diff
@@ -173,6 +174,10 @@ invoice_payment_hook_cb(struct invoice_payment_hook_payload *payload,
 	struct invoice invoice;
 	enum onion_type failcode;
 
+	/* We notify here to benefit from the payload and because the hook callback is
+	 * called even if the hook is not register. */
```
ZmnSCPxj: Grammar nazi nit: even if the hook is not register**ed**.

doc/PLUGINS.md
```diff
@@ -223,6 +223,19 @@ to a peer was lost.
 }
 ```
 
+#### `invoice_payment`
+
+A notification for topic `invoice_payment` is sent every time an invoie is paid.
+
+```json
+{
```
ZmnSCPxj: Is this not wrapped in an `invoice_payment` object?  From my reading of the code it should be similar to the `warning` notification which has a `warning` object wrapping it.

lightningd/invoice.c
darosior: Haha thanks

doc/PLUGINS.md
darosior: I forgot to updated the docs, thanks !

darosior:
Corrected the typo and updated the doc to add the top-wrapping object

PR #2823, opened 2019-07-21
Plugins: remove the 'warning' object from the 'warning' notification

darosior:
It's more a proposition than a merge request, but the changes were trivial so I made it a PR.

The `warning` object from the `warning` notification type seems redundant, and all other notifications types don't have a "framing" object.
It changes the API but since this type was not added long ago..

What do you think @trueptolemy ?

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -232,12 +232,10 @@ message resolving failed...
 
 ```json
 {
-	"warning": {
 	"level": "warn",
 	"time": "1559743608.565342521",
 	"source": "lightningd(17652): 0821f80652fb840239df8dc99205792bba2e559a05469915804c08420230e23c7c chan #7854:",
 	"log": "Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: sent ERROR bad reestablish dataloss msg"
-  }
 }
 ```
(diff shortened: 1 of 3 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2823/commits/3ebeba81aba699b2be57bcb073b1db24f29f2dc8

cdecker:
Come to think of it, we shouldn't unpack this, since it gets unpacked into a function call by the pylightning call, so if we were to add another field we'd break any plugin using this.

darosior:
> Come to think of it, we shouldn't unpack this, since it gets unpacked into a function call by the pylightning call, so if we were to add another field we'd break any plugin using this.

So we should either pack all other notifications, or remove the unpacking for notifications. Which one do you think is better ?

darosior:
Hmm that's the case for hooks too : some of them have a wrapping object of the same name, some does not. Some even have a wrapping object but with the field name being not the same as the hook name.. Maybe we should normalize this by adding (and requiring for future ones) a wrapping object for all hooks/notifications which field must be the name of the hook/notification ?

rustyrussell:
Yeah, you always have to wrap top-level responses in an object: you never know when you'll have to return a warning in future!

darosior:
Closed in favor of #2826

PR #2825, opened 2019-07-22
Make check fixes

rustyrussell:
This means we don't check BOLTS when EXPERIMENTAL_FEATURES is enabled, except for those tagged with this specific version.  This allows you to check BOLT quotes without getting tangled in old BOLTs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -282,9 +282,16 @@ check-bolt-dependency:
 
 BOLT_DEPS += check-bolt-dependency
 
+# Experimental quotes quote the exact version.
+ifeq ($(EXPERIMENTAL_FEATURES),1)
+CHECK_BOLT_PREFIX=--prefix="BOLT-$(BOLTVERSION)"
+else
+CHECK_BOLT_PREFIX=
+endif
+
 # Any mention of BOLT# must be followed by an exact quote, modulo whitespace.
 bolt-check/%: % bolt-precheck tools/check-bolt
-	@if [ -d .tmp.lightningrfc ]; then tools/check-bolt .tmp.lightningrfc $<; else echo "Not checking BOLTs: BOLTDIR $(BOLTDIR) does not exist" >&2; fi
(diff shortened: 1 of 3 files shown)


tools/generate-bolts.py
```diff
@@ -529,31 +529,34 @@ def main(options, args=None, output=sys.stdout, lines=None):
...
+                    # if this is an 'extension' field*, we want to add a new 'message' type
+                    # in the future, extensions will be handled as TLV's
+                    #
+                    # *(in the spec they're called 'optional', but that term is overloaded
+                    #   in that internal wire messages have 'optional' fields that are treated
+                    #   differently. for the sake of clarity here, for bolt-wire messages,
+                    #   we'll refer to 'optional' message fields as 'extensions')
+                    #
+                    orig_msg = msg
+                    for extension in tokens[5:]:
```
niftynei: i think i'm missing something about the expected tokens value here. in my understanding, the input data that we'd expect for these fields is only ever of size 1, eg:

    msgdata,channel_update,htlc_maximum_msat,u64,,option_channel_htlc_max

niftynei: i should really add one of these to `test_cases`

rustyrussell: The option_static_remotekey PR added a single (and probably our only ever!) case of this:

```diff
@@ -1117,7 +1129,7 @@ messages are), they are independent of requirements here.
    * [`channel_id`:`channel_id`]
    * [`u64`:`next_local_commitment_number`]
    * [`u64`:`next_remote_revocation_number`]
-   * [`32*byte`:`your_last_per_commitment_secret`] (option_data_loss_protect)
+   * [`32*byte`:`your_last_per_commitment_secret`] (option_data_loss_protect,option_static_remotekey)
    * [`point`:`my_current_per_commitment_point`] (option_data_loss_protect)
 
 `next_local_commitment_number`: A commitment number is a 48-bit
```

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/824fe1bc5df16e0ca43a402e8f629320cac73872

After rebase...

PR #2826, opened 2019-07-22
Plugins: standard object wrapper for hooks and notifications

darosior:
This is a follow-up to #2823 in which I wanted to remove the wrapping object. It ended up that it was actually more convenient to have one : this PR either adds or rename the top object of each notification and hook so that each one has a standard form : 
```
{
    "notification_name": {
        "a": 1,
        "b": 2
    }
}
```

### DIFF
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -23,8 +23,10 @@ void notify_connect(struct lightningd *ld, struct node_id *nodeid,
 {
 	struct jsonrpc_notification *n =
 	    jsonrpc_notification_start(NULL, "connect");
+	json_object_start(n->stream, "connect");
 	json_add_node_id(n->stream, "id", nodeid);
 	json_add_address_internal(n->stream, "address", addr);
+	json_object_end(n->stream);
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
 }
@@ -33,7 +35,9 @@ void notify_disconnect(struct lightningd *ld, struct node_id *nodeid)
 {
 	struct jsonrpc_notification *n =
 	    jsonrpc_notification_start(NULL, "disconnect");
+	json_object_start(n->stream, "disconnect");
 	json_add_node_id(n->stream, "id", nodeid);
+	json_object_end(n->stream);
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
 }
@@ -53,7 +57,7 @@ void notify_warning(struct lightningd *ld, struct log_entry *l)
 	json_add_string(n->stream, "level",
 			l->level == LOG_BROKEN ? "error"
 			: "warn");
-	/* unsuaul/broken event is rare, plugin pay more attentions on
+	/* unusual/broken event is rare, plugin pay more attentions on
 	 * the absolute time, like when channels failed. */
 	json_add_time(n->stream, "time", l->time.ts);
 	json_add_string(n->stream, "source", l->prefix);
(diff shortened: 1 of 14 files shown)


lightningd/notification.c
```diff
@@ -42,7 +46,7 @@ void notify_warning(struct lightningd *ld, struct log_entry *l)
 {
 	struct jsonrpc_notification *n =
 	    jsonrpc_notification_start(NULL, notification_topics[2]);
-	json_object_start(n->stream, "warning");
+	json_object_start(n->stream, notification_topics[2]);
```
cdecker: Using `notification_topics` with magic indices is very brittle, we should use explicit strings like we do elsewhere. Otherwise changes in the order of `notification_topics` can have unintended consequences.

darosior:
> While I agree with the general idea, I think it's likely too late to change these just like this, since this will break all existing plugins.

```
darosior@darosior:~/projects/plugins$ grep -ri "subscribe" *
darosior@darosior:~/projects/plugins$ grep -ri "hook" *
autoreload/README.md: - Since the registration of options, subscriptions, methods and hooks happens
autoreload/autoreload.py:    for h in manifest.get("hooks", []):
autoreload/autoreload.py:        plugin.add_hook(h, plugin.child.proxy_method, background=True)
conftest.py:@pytest.hookimpl(tryfirst=True, hookwrapper=True)
conftest.py:    # execute all other hooks to obtain the report object
```
I did not find any plugin which has registered either for a notification or a hook in [lightningd/plugins](https://github.com/lightningd/plugins) (though there is the db backup plugin in the PRs but it can be trivially amended before merged), and all from this repo have been corrected in this PR. Moreover it only affects Python plugins.
I'm not particularly attached to get this merged, but I think it's better to modify (and standardize for new types) it now before more plugins register for hooks/notifications.

darosior:
Rebased on top of master to comply with recent notifications changes

ZmnSCPxj:
@niftynei 

Adding a wrapper was judged to be necessary due to quirks of pylightning: https://github.com/ElementsProject/lightning/pull/2823#issuecomment-513575544

Also, better to add a wrapper for some reason: https://github.com/ElementsProject/lightning/pull/2823#issuecomment-513610782

darosior:
> While standardization is nice, if we're going to do that my inclination would be to collapse them down, i.e. remove the upper level that some of them have

That was what I first thought of and implemented in #2823. But @cdecker pointed out that we could not modify (add/remove) a field without risking to break the API, it's better to add a top-wrapping object so that fields can be added/removed from it (this problem was mitigated, cf below).

> Also note: since you're changing the structure of the json returned from c-lightning here, this will break the Golang plugin framework that I maintain glightning. Long term, I think this is fine as the current goal I'm working towards with g-lightning is to be tightly pegged to c-lightning release versions.

Yeah that might be the case for other plugin libraries too. On the last break @rsbondi was pretty reactive though :-).

Globally I think it's ok to break notifications until no one use them (I implemented them in lightning-qt yesterday but I can trivially amend the change).

> Also, better to add a wrapper for some reason: #2823 (comment)

This problem was mitigated by adding `**kwargs` to unpacked functions in Pylightning. https://github.com/ElementsProject/lightning/commit/9d4148ce683223ec286d14c3d5bc737cc7ba66ed

cdecker:
I think we should take care to wrap the calls in a top-level object going forward, but due to our backwards compatibility requirements we should likely not just change all existing ones.

So let's keep this rule in mind going forward, and slowly switch the existing ones out as we change them.

I'll close this PR for now since it'd be a major break in our backward compatibility.

PR #2827, opened 2019-07-22
Plugins: correct 'hold_htlc' testing plugin

darosior:
We `print()` then wait for logs :/

### DIFF
--- a/tests/plugins/hold_htlcs.py
+++ b/tests/plugins/hold_htlcs.py
@@ -27,7 +27,7 @@ def on_htlc_accepted(htlc, onion, plugin):
 
     time.sleep(10)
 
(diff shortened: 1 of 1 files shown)


cdecker:
NACK, calling `print` inside a plugin is identical to calling `plugin.log` since we autopatch stdout to wrap lines we print to be valid JSON-RPC notifications:

https://github.com/ElementsProject/lightning/blob/bb301040e43a5ba4cb6b1423b4869183147b6655/contrib/pylightning/lightning/plugin.py#L542-L557

darosior:
Hmm I had an error and thought this was caused by that but it does not make much sense actually, closing it

Issue #2830, opened 2019-07-23
building on ubuntu 16.04 since commit 5c07afac7d082ee4d878c6fe20c2910bbc08458d 


mocacinno:
### Issue and Steps to Reproduce

Since commit 5c07afac7d082ee4d878c6fe20c2910bbc08458d i can no longer build on ubuntu 16.04.6 LTS. Building on ubuntu 18.04.1 LTS works fine.

I've doublechecked the prereqs, but i always end up with following error messages:

```
cc -DBINTOPKGLIBEXECDIR="\"../libexec/c-lightning\"" -Wall -Wundef -Wmissing-prototypes -Wmissing-declarations -Wstrict-prototypes -Wold-style-definition -Werror -Wno-error=maybe-uninitialized             -std=gnu11 -g -fstack-protector -Og -I ccan -I external/libwally-core/include/ -I external/libwally-core/src/secp256k1/include/ -I external/jsmn/ -I external/libbacktrace/ -I external/libbacktr            ace-build -I . -I/usr/local/include    -DSHACHAIN_BITS=48 -DJSMN_PARENT_LINKS  -DCOMPAT_V052=1 -DCOMPAT_V060=1 -DCOMPAT_V061=1 -DCOMPAT_V062=1 -DCOMPAT_V070=1 -DBINTOPKGLIBEXECDIR="\"../libexec            /c-lightning\""  -c -o hsmd/hsmd.o hsmd/hsmd.c
hsmd/hsmd.c: In function 'handle_channel_update_sig':
hsmd/hsmd.c:696:59: error: passing argument 2 of 'fromwire_channel_update_option_channel_htlc_max' from incompatible pointer type [-Werror=incompatible-pointer-types]
  if (!fromwire_channel_update_option_channel_htlc_max(cu, &sig,
                                                           ^
In file included from hsmd/hsmd.c:54:0:
./wire/gen_peer_wire.h:166:6: note: expected 'u8 * {aka unsigned char *}' but argument is of type 'secp256k1_ecdsa_signature * {aka struct <anonymous> *}'
 bool fromwire_channel_update_option_channel_htlc_max(const void *p, u8 *message_flags, u32 *timestamp, struct bitcoin_blkid *chain_hash, u16 *cltv_expiry_delta, struct amount_msat *htlc_maximu
      ^
[...]

PR #2832, opened 2019-07-23
bolt-gen: use ordered-dict for fields

niftynei:
Use ordered dict for fields, whose order matters when iterated
through (especially for argument lists).  see #2830 

Reported-By: @mocacinno

### DIFF
--- a/tools/generate-bolts.py
+++ b/tools/generate-bolts.py
@@ -18,6 +18,7 @@
 #   subtypedata,<subtypename>,<fieldname>,<typename>,[<count>]
 
 from argparse import ArgumentParser, REMAINDER
+from collections import OrderedDict
 import copy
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2832/commits/0f92f88e7d88c4be7625c6ea66f036174db6455c

jdabs:
Isn't `self.fields` var supposed to be an array?

niftynei:
> Isn't self.fields var supposed to be an array?

i'm not sure i understand what you mean by 'supposed to' @jdabs. it's written as a dict as that allows for faster field retrieval in the `find` function, https://github.com/ElementsProject/lightning/blob/master/tools/generate-bolts.py#L124

PR #2833, opened 2019-07-24
Final wire-gen cleanups

niftynei:
Final pieces of the wire-generator move. This adds a few features that the internal wire specification format require (namely enum handling) and fixes up an issue with how we were generating code for optional fields that are of an 'assignable' type.

Also moves all of the internal wire formats over to the new one. Huzzah!?

based on #2821 !!


Makefile
```diff
@@ -184,8 +184,8 @@ ALL_GEN_HEADERS += gen_version.h
 
 CDUMP_OBJS := ccan-cdump.o ccan-strmap.o
 
-WIRE_GEN := tools/generate-wire.py
-BOLT_GEN := tools/generate-bolts.py
+BOLT_GEN := tools/generate-wire.py
+WIRE_GEN := $(BOLT_GEN)
```
niftynei: call me weird, but i kind of like how we've now clearly 'labeled' which csv's are 'bolts' and which are 'wires'. it's really an unnecessary distinction since the formatter handles them the same.

tools/generate-bolts.py
```diff
@@ -238,13 +238,16 @@ def true_type(type_name, field_name=None):
...
+        self.name, self.is_enum = self.parse_name(name)
         self.depends_on = {}
-        # FIXME: internal msgs can be enums
-        self.is_enum = False
         self.type_comments = []
         self.tlv = False
 
+    def parse_name(self, name):
+        if name.startswith('e:'):
+            return name[2:], True
```
rustyrussell: I prefer 'enum ' here over 'e:'.  But it's a trivial change, so can happen afterwards.

niftynei: ok, i'm not partial to either. put in a patch commit at the end that makes this change!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2833/commits/8c240129fbb8ff23a1ee1d332916028276def36d

PR #2836, opened 2019-07-24
plugins: log PID of starting plugins, and some other small fixes

SimonVrouwe:
A few things I had in my TODO list:

~openingd: Fix the _fall through_ in `handle_peer_in` switch. It now errs when receiving msg type that is a protocol violation, which happens in `test_fee_limits`. So fixed that too and, while at it, also solved a FIXUP.~

~openingd: in  `handle_peer_in`, tidy up  the switch statement. Now easier to follow and more redundant.~

**EDIT2**: Dropped the above

pytest: solved a FIXUP in `test_fee_limits` 

Improve the description of `dev-no-reconnect`, it took me a while to figure out its behavior.

Log PID's of plugins, I needed this for the dbbackup plugin, but maybe it is also useful for debugging

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -508,7 +508,7 @@ static void dev_register_opts(struct lightningd *ld)
 {
 	opt_register_noarg("--dev-no-reconnect", opt_set_invbool,
 			   &ld->reconnect,
-			   "Disable automatic reconnect attempts");
+			   "Disable automatic reconnect-attempts by this node, but accept incoming");
 	opt_register_noarg("--dev-fail-on-subdaemon-fail", opt_set_bool,
 			   &ld->dev_subdaemon_fail, opt_hidden);
 	opt_register_early_arg("--dev-debugger=<subprocess>", opt_subprocess_debug, NULL,
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -989,6 +989,8 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 		if (p->pid == -1)
 			fatal("error starting plugin '%s': %s", p->cmd,
 			      strerror(errno));
+		else
+			log_debug(plugins->log, "started(%u) %s", p->pid, p->cmd);
 		p->buffer = tal_arr(p, char, 64);
 		p->stop = false;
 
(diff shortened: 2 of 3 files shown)


openingd/openingd.c
```diff
@@ -1537,8 +1537,13 @@ static u8 *handle_peer_in(struct state *state)
...
 	case WIRE_GOSSIP_TIMESTAMP_FILTER:
 	case WIRE_ERROR:
-	case WIRE_CHANNEL_REESTABLISH:
+		if (handle_peer_gossip_or_error(state->pps,
+						&state->channel_id, msg))
+			return NULL;
+		break;
+
 	/* These are all protocol violations at this stage. */
+	case WIRE_CHANNEL_REESTABLISH:
```
rustyrussell: Hmm, I had to read the code, but this is wrong. handle_peer_gossip_or_error also handles messages (esp. reestablish) which are talking about the wrong channel, and sends a more useful error message.

SimonVrouwe: Thanks my mistake, I should've read the function description more clearly.

SimonVrouwe:
Modified the first two commits.

WIRE_CHANNEL_REESTABLISH in `openingd` is not a protocol violation, as it could belong to another channel (for same peer), but we don't support multi-channels yet.

rustyrussell:
> WIRE_CHANNEL_REESTABLISH in `openingd` is not a protocol violation, as it could belong to another channel (for same peer), but we don't support multi-channels yet.

But what about other messages they might send about the incorrect channel?  Best to leave it to handle_peer_gossip_or_error().  Which, true, doesn't handle all the cases yet, but it could.

BTW: the only reason we don't use default: here is because we want a compile warning when we add new protocol fields, to force an audit of how we handle the new msg at this point.

SimonVrouwe:
> But what about other messages they might send about the incorrect channel? Best to leave it to handle_peer_gossip_or_error(). Which, true, doesn't handle all the cases yet, but it could.

Aha now I (finally) see the pattern. Thanks for the clarification (and your patience :)).

Dropped the first commit.

darosior:
ACK 80e1e8d7ae71950c5c451267761f69c616ec8d94

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/80e1e8d7ae71950c5c451267761f69c616ec8d94

PR #2846, opened 2019-07-25
RPC: Return output idx for a funding tx in a channel information in listfunds

kristapsk:
I am [working on a plug-in](https://github.com/kristapsk/lightning-feereport) that emulates `lncli feereport` command of LND for c-lightning. It needs to return `chan_point` for each channel, so I need not only channel funding txid, but also output index. This information is present anyways, why not to return, and, besides, it is already returned for onchain UTXOs.

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -657,6 +657,8 @@ static struct command_result *json_listfunds(struct command *cmd,
 						   "amount_msat");
 			json_add_txid(response, "funding_txid",
 				      &c->funding_txid);
+			json_add_num(response, "funding_output",
+				      c->funding_outnum);
 			json_object_end(response);
 		}
 	}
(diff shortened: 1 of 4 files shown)


wallet/walletrpc.c
```diff
@@ -657,6 +657,8 @@ static struct command_result *json_listfunds(struct command *cmd,
 						   "amount_msat");
 			json_add_txid(response, "funding_txid",
 				      &c->funding_txid);
+			json_add_num(response, "funding_output",
```
darosior: nit: `funding_outnum` ?

kristapsk: It already uses `txid` and `output` (for `outnum`) in `outputs`, so I think it's better to have consitency here with `channels` too.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2846/commits/8e544505773c888b9ddc60c6d7633abb69ce9dec

kristapsk:
Added also changelog line.

ZmnSCPxj:
ACK 470391d

Issue #2796, opened 2019-07-06
Allow plugin hook chaining


ZmnSCPxj:
Currently, only one plugin at a time can register for each hook.  If two plugins want to register the same hook, then they have to be merged into a single plugin.  This gets worse, if multiple plugins want to register the same hook.

I propose to add to each hook, the ability to chain hooks.

This is useful for many things.

# Motivating Examples

## Multiple backup strategies.

Currently, it is possible to connect only a single "backup" plugin to your node.

However, consider that you might want to have not just a single backup policy, but multiple:

* A plugin that replays the db queries on an sqlite database on the same machine, but on a different physical persistent storage.
* A plugin that sends the db queries unencrypted over an encrypted connection, to a receiver on a remote computer you control,
* A plugin that encrypts the db queries, then saves it on some cloud-rented storage.

We could write a single plugin that does all the above and maintain a complicated plugin whose many functionalities are controlled by multiple options.

Or we could instead write one plugin for each backup policy and retain simplicity, and simply connect multiple plugins to the `db_hook`.

In principle `db_hook` is a "broadcast"-type hook, with the added wrinkle that all backup plugins should be in lockstep.

## Multiple uses of `htlc_accepted`

Currently we could use `htlc_accepted` for many usecases:

* Cross-chain swaps as per @bitonic-cjp 
* Replacement for paycodes #2666 
* Implementing JIT routing as per @renepickhardt , when we implement deferring the hook.

Combining the disparate intended uses into a single plugin would be ridiculously unwieldy.

# Proposal

When we define a hook in C-Lightning, we should add to the hook data also the below:

* a `combiner_function`
* a `default_result`

Then, hook callbacks no longer have to check if there is a result or not --- semantically, there is always a result.

Each `plugin_hook` then maintains a list of plugins connected to that hook.  When a plugin is killed, it is removed from the list of plugins (care must be taken in the implementation of the list and list traversal in that a plugin may die either before, during, or after it is called on a hook).

If there is no plugin connected to the hook, the `default_result` is parsed as JSON and hook callbacks respond to that.

If there is at least one plugin connected, then after one plugin successfully returns in some result, we always invoke the `combiner_function`. It is an async function, and is give a callback (`combined_result_cb`) that will be invoked if it can provide the result now.  Generally its choices are:

[...]

PR #2851, opened 2019-07-26
Plugin: Make `lightningd` requests and notifications registrable.

trueptolemy:
# Background
The last days, when I thought about `bitcoin` plugin, I find `lightningd` need ask `bitcoin` plugin for some extern information or ask the plugin to execute some external commands.
At the beginning, I made many hooks, like `estimate_fee`, `sendrawtx`, `getrawblock`, `getblockcount`, `getblockhash`, `getoutput` and `gettxout`... But then I found "ask for extern information or ask to execute some external commands" does not match the role of `hook`. 

## Existing Disadvantage
1. In my understanding, `hook` type usually return a command to tell `lightningd` what to do. Replying lots of information doesn't match the task of `hook`.
2. `hook` is synchronous, which limits the `clightning` performance if `lightningd` send `hook` out usually.

# Summary
In this PR, I implement two plugin types:
- `request`
  - Based on `bitcoin-cli` request management.
    - Extraly add a timer, which can trigger `stopper`.
  - Used by `lightningd` to ask plugins for extern information or ask plugins to execute some external commands.
  - Can be called before daemons(except `hsmd`) initial.
    - Move pid and jsonrpc initial processes before network setup.
    - Add a new plugin method `request_init` called here to per-initial plugins which subscribe `request`. 
  - Implement a `request` type: `getclientversion` as the example.
- `notification_in`
  - Used by plugins to send notifications to `lightningd`.
  - Plugin don't need to subscribe, and send `lightningd` notifications under given topic and with correct json format.
  - Must be send after `init` method.
    - Make `log` as a `notification_in` example, but it is a little different. It can be send after `manifest`, and plugins should use `log` APIs other than send `log` notification with payload directly.

I also implement a new plugin method:
 -  `request_init`
    - Move pid and jsonrpc initial processes before network setup.
    - Call `request_init` immediately after pid and jsonrpc initial processes to activate the requests interface of plugins.
    - The `request_init` method is only used for `request` type. Only the plugin that subscribes any `request` must handle `request_init`, and all the plugins still need handle `init`.(In order not to destroy the running of existing plugins).

I am not sure if this PR is the proper solution. So post it here and welcome feedback.


ZmnSCPxj:
Have not checked at all, but I have question: How would this interact with #2796 ?

trueptolemy:
> Have not checked at all, but I have question: How would this interact with #2796 ?

I think it doesn't interact `hook` type.
#2796 is the management for `hook`, and `request` is another type with its own management(based on `bitcoin-cli` request design).
The diffience between `hook` and `request` is like this:
- `hook`:
`openchannel - default_result = {'result': 'continue'}` 
The default result is a command from plugins.
And `openchannel` itself is more like a ~~notification~~ other than a command.
- `request`:
`getclientversion - result = {'client': 'bitcoind', 'version': 180000}`
The `getclientversion` itself is a command from `lightningd`, and the plugin must do something and return the result.(no default result).

ZmnSCPxj:
Since plugins can provide commands, how is this so different from, say, having a set of commands e.g. `backend_getfeeestimate` `backend_broadcasttx` which a plugin can define?  Then move the current bitcoind interface to an optional plugin?

trueptolemy:
> Since plugins can provide commands, 

Do you mean the  `rpcmethods`  that plugins register? But `lightningd` couldn't call these `rpcmethods` directly, and plugins still need to send back these extern information to `lightningd` through `hook` or other ways.
The case I supposed is when `lightningd` need to know some extern information(eg, know external bitcoin client version) or need to do some external command(eg, send transactions), it tell plugins what it need and what plugin should do. Then plugins do these and return result.

This process is really like what `hook` does now, and the difference I can think of is `hook` maybe not be as the interface (though which `lightningd` ask for plugins lots of information), and `hook` don't have flexible asynchronous management.

And another thought about `hook` and `request` is:
-  `hook` extend `lightningd` functionality through plugins.
-  But `request` cut the original essential `lightningd` features, and instead let the plugin replace the these implementation.

trueptolemy:
I need continue cleanup the conflict when merge and fix valgrind error...

ZmnSCPxj:
> But `lightningd` couldn't call these `rpcmethods` directly

Maybe not now, but it can be modified to do so, right?  What I am mostly thinking is to fold your `requests` into `rpcmethods` and provide a way to internally invoke `rpcmethods` (it should be doable: after all, the JSON-RPC interface is able to forward messages to them).  Though maybe we should also ask opinion of @rustyrussell @cdecker  @niftynei

trueptolemy:
> What I am mostly thinking is to fold your requests into rpcmethods and provide a way to internally invoke rpcmethods

Thank you. Yes, these requests indeed can be folded into `rpcmethod`. And I think maybe they still need deal with some points:
1. Need given topics(`lightningd` can register and plugins implement detail)
Plugins will change but the requirement of `lightningd` is changeless.
2. Need to `init` before network setup.
Then these `rpcmethods`s can supply necessary information.
3. Asynchronous.

In fact, at the first I'm not sure how to change current `init` and `rpcmethod` interface.(Changes may affect some plugin existed)

This PR is more like to raise 3 points above and incidentally present a way(not the most proper way), which avoids to make more changes for current API and simultaneously to fix these 3 points above: 1. Add new type(`request`) to avoid change `hook`/`rpcmethod`, and supply `autodata` to register any topic; 2. Add new method(`request_init` and only used for `request` type) to avoid change `init`; 3. Reuse code(`bitcoin-cli`).

cdecker:
I do agree that the synchronous nature of hooks is not optimal for this use-case. I had intended to use them, but you have convinced me that might not be a good idea :wink: 

I think all of your points are already addressed by the JSON-RPC passthrough: if you have a plugin that registers some RPC methods `lighntingd` can make use of them just like a CLI caller might, we just need to tell `lightningd` to use them instead of the call to `bitcoin-cli` :slightly_smiling_face: 

So I'd propose using the existing JSON-RPC infrastructure and instead add a flag in the `getmanifest` result to mark some calls as internal-only calls (calls that shouldn't be re-exposed through `lightningd`s RPC).

Furthermore I'd propose using more coarse-grained calls instead of going through the motions with `bitcoin-cli`:

 - `getblockbyheight [height]` combines `getblockhash [height]` and `getrawblock [hash]` in a single call
 - `estimatefee` which internally calls `estimatesmartfee` for a range of blockheights
 - `gettxoutbyscid` which combines the tedious work of looking up a txout by the short-channel-id (`getblockhash [height]`, `getblock [hash], `extract txid, `gettxout [txid] [outnum]` to get the current output state)

I think by combining these we can implement them in a variety of ways without sacrificing flexibility.

ZmnSCPxj:
> `getblockbyheight [height]` combines `getblockhash [height]` and `getrawblock [hash]` in a single call

How does this let us handle the case of orphaned chain tips?

trueptolemy:
Thank you : ) @cdecker ! Now I make sense yours and @ZmnSCPxj  idea.
I really agree with 

> add a flag in the getmanifest result to mark some calls as internal-only calls

 and 

> using more coarse-grained calls instead of going through the motions with bitcoin-cli


Could I  continue to ask a few questions about details?

#### 1. About internal "rpcmethod". 

**What about letting these special `rpcmethod`s have stable names?**
(In other words, should we selected fixed names in advance, like `getblockbyheight`, `estimatefee` and `gettxoutbyscid`, and plugins must use these names?)

For `hook`s and `request`s,  `lightningd` can register names in advance, then it have call interface and callback. But for `rpcmethod`, their name are selected by plugins.

Thus for internal `rpcmethod`, a fixed name(**proper noun**) seems necessary!
If so, we even need not set the flag in `getmanifest`, just use the fixed name(**proper noun**) like what you proposed.

#### 2. About `init` before network setup. 

**Should we move `init` method call before network steup or let internal `rpcmethod` work after `getmanifest`?**
I'm not sure the effect for existing plugins if we move `init` method call before network steup.

cdecker:
> > `getblockbyheight [height]` combines `getblockhash [height]` and `getrawblock [hash]` in a single call
> 
> How does this let us handle the case of orphaned chain tips?

The way we currently handle this is by checking if the `prev_blockhash` matches when receiving a new block, and then removing the tip until we match with our expectations again, followed by querying for the next `getblockhash` and `getrawblock`. By combining the two calls into one we may end up with a bit of overhead (if we don't remember blocks during the unrolling. In the normal case we'd just save a roundtrip to the bitcoin backend (important for IBD) and it'd allow us to have a single 
call to be handled instead of two.

For reference this is the code that currently does the check and unrolling:

https://github.com/ElementsProject/lightning/blob/1af8d9ebdb8711fa6309196be8e56443361f159a/lightningd/chaintopology.c#L702-L714

> Thank you : ) @cdecker ! Now I make sense yours and @ZmnSCPxj idea.
> I really agree with
> 
> > add a flag in the getmanifest result to mark some calls as internal-only calls
> 
> and
> 
> > using more coarse-grained calls instead of going through the motions with bitcoin-cli
> 
> Could I continue to ask a few questions about details?
> 
> #### 1. About internal "rpcmethod".
> **What about letting these special `rpcmethod`s have stable names?**
> (In other words, should we selected fixed names in advance, like `getblockbyheight`, `estimatefee` and `gettxoutbyscid`, and plugins must use these names?)

That is definitely the intention, they need stable names otherwise we cannot call them ourselves :-) The internal-only flag is mostly intended for future extensibility: we might not want CLI users to be able to call the internal-only methods (or maybe we do?), and if a plugin is built for a later version of c-lightning that has added a new internal method name, but is run on an older version that doesn't recognize the method's name as internal-only, it'd be exposed through the JSON-RPC passthrough. Adding a flag for internal-only guards against this since a `lightningd` that gets a method registration which is marked as internal-only, but not understood by c-lightning, it would still not re-expose it through its JSON-RPC.

> For `hook`s and `request`s, `lightningd` can register names in advance, then it have call interface and callback. But for `rpcmethod`, their name are selected by plugins.
> 
> Thus for internal `rpcmethod`, a fixed name(**proper noun**) seems necessary!
> If so, we even need not set the flag in `getmanifest`, just use the fixed name(**proper noun**) like what you proposed.

Absolutely agree, the names need to be specified in the docs.

> #### 2. About `init` before network setup.
> **Should we move `init` method call before network steup or let internal `rpcmethod` work after `getmanifest`?**
> I'm not sure the effect for existing plugins if we move `init` method call before network steup.

That is a bit of a strange one indeed, we inadvertently already added such a case with the `db_write` hook, which may send migration SQL statements to the plugin before its `init` is called. Luckily we can just cache those early statements and flush them once we get `init`ed. This might be not so easy for a bitcoin backend since we may require those configuration parameters to know how to act.

The reason we have the `init` call where it is right now is that we also use `init` to signal the availability of the JSON-RPC which the plugin may want to use from that point onward.

There are three options here:
 - Return dummy results until we get the `init`
 - Start the JSON-RPC interface earlier so that it precedes any call to the bitcoin backend
 - Split init into two parts and have an early `configure` call, and a later `init` call that signals RPC readiness.

The last option is probable not a good idea since we break existing plugins. The second to last option is also kind of ugly since we may end up performing some startup checks (fee estimation, blockchain head and chain check) after the plugin was started, which also a bit weird.

ZmnSCPxj:
> we might not want CLI users to be able to call the internal-only methods (or maybe we do?)

I believe we do?  For example, consider a plugin that uses `fundchannel_start` and `fundchannel_complete` and a connection to the Wasabi Chaumian CoinJoin server to fund channels directly from a mix.  Such a plugin might want to use our access to the `bitcoind` for its own transaction.

cdecker:
> > we might not want CLI users to be able to call the internal-only methods (or maybe we do?)
> 
> I believe we do? For example, consider a plugin that uses `fundchannel_start` and `fundchannel_complete` and a connection to the Wasabi Chaumian CoinJoin server to fund channels directly from a mix. Such a plugin might want to use our access to the `bitcoind` for its own transaction.

Agreed, having a way for users to use `lightningd` as their bitcoin backend might be interesting. So we could also make the `internal` flag just a filter for `help` so that they don't get displayed by default in order not to pollute the help output?

If not, I'm also happy to expose them by default.

So the path forward on this would be to just have plugins register methods, and some methods are special in that they adhere to a method signature and can be used by `lightningd` internally as a bitcoin backend. How does that sound @trueptolemy and @ZmnSCPxj ? This would reduce the number of duplicate things we need to implement, making better use of what we already have, and could potentially expose the bitcoin backend to more users.

trueptolemy:
> So the path forward on this would be to just have plugins register methods, and some methods are special in that they adhere to a method signature and can be used by lightningd internally as a bitcoin backend. 

It is very clear and reasonable :+1: . I'll submit another PR ASAP.

ZmnSCPxj:
Seems okay to me.

PR #2852, opened 2019-07-26
Handle relative paths for lightningd startup options

darosior:
This joins the current working directory to the options argument if the option argument needs to be a path and if the provided argument is not an absolute path.


rustyrussell:
So, except for plugins and the config file paths are relative to the lightning-dir. I wanted those, too, but others disagreed.

That's also what the documentation says. What problem are you seeing sure relative paths?

darosior:
Thank you for the clarifications.

PR #2853, opened 2019-07-26
wire gen fixups

niftynei:
- Small fix for the way 'variable-length, variable-sized' subtype objects are allocated. Discovered while working on wire-tests for the dual-funding proposal that nests witness_elements into witness_stacks.

- Allow `towire_`'s to pass NULL in for the TLV field, if they want to.


rustyrussell:
I like the NULL addition, will simplify some paths for me!

Ack https://github.com/ElementsProject/lightning/pull/2853/commits/1181909d52b7c4d1f2d40d9eb2367dab2ec229c7

PR #2854, opened 2019-07-27
Clean-up some warnings. Initialized some uninitialized variables.

Chirimen-Jako:
### Changed

- Clean-up some warnings.
- Initialize some uninitialized auto variables which value was unstable.

### DIFF
--- a/ccan/ccan/tal/path/path.c
+++ b/ccan/ccan/tal/path/path.c
@@ -302,7 +302,7 @@ fail_take_to:
 
  char *path_readlink(const tal_t *ctx, const char *linkname)
  {
-	ssize_t len, maxlen = 64; /* good first guess. */
+	ssize_t len = 0, maxlen = 64; /* good first guess. */
 	char *ret = NULL;
 
 	if (unlikely(!linkname) && is_taken(linkname))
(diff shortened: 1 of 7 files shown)


Chirimen-Jako:
Hi,
Thanks you very much for your reviewing. Sorry to take your precious time.
> initializing things which should not be initialized can cover up real bugs!
I agreed with it. However, I think disable warning is more risky.
I'm not sure there is a variable which should not be initialized in the C/C++ world.
Each of variables should be reviewed one by one and initialized by obviously invalid value in the future in my opinion.

Thanks!

ZmnSCPxj:
> I agreed with it. However, I think disable warning is more risky.
I'm not sure there is a variable which should not be initialized in the C/C++ world.
Off course, I understand performance problem. No need to initialize unused memory, or if it's overwritten by valid value exactly. But some of them are referred in the code context in this case.
Each of variables should be reviewed one by one and initialized by obviously invalid value in the future in my opinion.

We have found `valgrind` to be more reliable in finding *actual* uninitialized memory on both stack and heap.  Indeed, initializing variables to "invalid" values tends to *hide* such uninitialized memory from `valgrind`.

As we run `valgrind` in our CI we prefer to keep variables uninitialized if possible, so that `valgrind` can find such bugs accurately (unlike the compilers, which are often hit-or-miss).

Chirimen-Jako:
Thank you for your kind explanation.
I understood completely. Because I was using `BoundsChecker` (maybe, similar to `valgrind`) decades ago.

cdecker:
This seems to be addressed now by #2884

PR #2857, opened 2019-07-28
gossipd: fix crash when we prune old, un-updated channel announcements.

rustyrussell:
Hey, turns out this @m-schmoock was *not* due to some ancient curse!


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2857/commits/cb88c3157f63e74303cc4382e1328a556375412d

PR #2858, opened 2019-07-28
Added possibility to configure max_concurrent_htlcs value 

renepickhardt:
Eclair has a default value of 30 and I thought why not going with their value and while doing so make it configureable.

@cdecker suggested to call the option  `max_concurrent_htlcs` instead of the name in the BOLTs `max_concurrent_htlcs`

Here is the relevant line in eclair:

https://github.com/ACINQ/eclair/blob/9afb26e09c69dd5d6a14732baf5dcdf2b7a9142b/eclair-core/src/main/resources/reference.conf#L62

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -367,6 +367,7 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	enum side sender = htlc_state_owner(state), recipient = !sender;
 	const struct htlc **committed, **adding, **removing;
 	const struct channel_view *view;
+	u32 min_concurrent_htlcs;
 
 	htlc = tal(tmpctx, struct htlc);
 
@@ -443,8 +444,16 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	 *     HTLCs to its local commitment transaction...
 	 *     - SHOULD fail the channel.
 	 */
+	/* Also we should not add more htlc's than sender or recipient
+	 * configured.  This mitigates attacks in which a peer can force the
+	 * funder of the channel to pay unnecessary onchain fees during a fee
+	 * spike with large commitment transactions.
+	 */
+	min_concurrent_htlcs = channel->config[recipient].max_accepted_htlcs;
+	if (min_concurrent_htlcs > channel->config[sender].max_accepted_htlcs)
+		min_concurrent_htlcs = channel->config[sender].max_accepted_htlcs;
 	if (tal_count(committed) - tal_count(removing) + tal_count(adding)
-	    > channel->config[recipient].max_accepted_htlcs) {
+	    > min_concurrent_htlcs) {
 		return CHANNEL_ERR_TOO_MANY_HTLCS;
 	}
 
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -2,12 +2,12 @@
 .\"     Title: lightningd-config
 .\"    Author: [see the "AUTHOR" section]
 .\" Generator: DocBook XSL Stylesheets v1.79.1 <http://docbook.sf.net/>
-.\"      Date: 08/04/2019
+.\"      Date: 08/09/2019
 .\"    Manual: \ \&
 .\"    Source: \ \&
 .\"  Language: English
 .\"
-.TH "LIGHTNINGD\-CONFIG" "5" "08/04/2019" "\ \&" "\ \&"
+.TH "LIGHTNINGD\-CONFIG" "5" "08/09/2019" "\ \&" "\ \&"
 .\" -----------------------------------------------------------------
 .\" * Define some portability stuff
 .\" -----------------------------------------------------------------
@@ -255,6 +255,11 @@ Limits on what onchain fee range we\(cqll allow when a node opens a channel with
 \fIestimatesmartfee 2\fR\&. If they\(cqre outside this range, we abort their opening attempt\&. Note that
 \fBcommit\-fee\-max\fR
 can (should!) be greater than 100\&.
+.RE
+.PP
+\fBmax\-concurrent\-htlcs\fR=\fIINTEGER\fR
+.RS 4
+Number of HTLCs one channel can handle concurrently in each direction\&. Should be between 1 and 483 (default 30)\&.
 .RE
 .PP
 \fBcltv\-delta\fR=\fIBLOCKS\fR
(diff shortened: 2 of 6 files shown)


renepickhardt:
I added an entry to `doc/lightningd` however I have proplems (like last time I tried to add a manpage) to run a2x even after installing asciidoc

```
a2x --format=manpage doc/lightningd-config.5.txt
a2x: ERROR: "xmllint" --nonet --noout --valid "/Users/rpickhardt/hacken/tmp/lightning/doc/lightningd-config.5.xml" returned non-zero exit status 4
``` 
web search was of no help I guess I am missing some dependencies.

rustyrussell:
This doesn't solve your problem though, as this doesn't restrict how many HTLCs we'll send to a peer, just how many we'll receive.

I would suggest a follow-up patch which caps the sending number to max(what-peer-set, what-our-max-is)?

renepickhardt:
While the first paragraph of your review makes sense to me I am confused by your suggestion for a follow up patch. 

Shouldn't it be `min(what-peer-set, what-our-max-is)` instead of `max`? Or did you mean `what-peer-seNt` (in the sense of how many did we received so far)?

In my understanding concurrent htlcs was meant to be offered plus received. So the max would not help here as we would have to check against the current sum in both cases or is there a mistake in my thinking?

renepickhardt:
can't I just modify this line: 
https://github.com/ElementsProject/lightning/blob/54ce4ed1cf4007ac8754c5bb640c4afdbdbc3934/channeld/full_channel.c#L451

to test `channel->config[eachside].max_accepted_htlcs`

with `LOCAL` and `REMOTE` as values for each side and take the min of both values or did I miss something in the code base?

rustyrussell:
> can't I just modify this line:
> 
> https://github.com/ElementsProject/lightning/blob/54ce4ed1cf4007ac8754c5bb640c4afdbdbc3934/channeld/full_channel.c#L451
> 
> to test `channel->config[each_side].max_accepted_htlcs`
> 
> with `LOCAL` and `REMOTE` as values for `each_side` and take the min of both values or did I miss something in the code base?

Yes, exactly, that would have the same effect.

renepickhardt:
I added the additional patch as @rustyrussell requested and rebased to current master. Think everything should be working now. fyi @cdecker

rustyrussell:
> I added an entry to `doc/lightningd` however I have proplems (like last time I tried to add a manpage) to run a2x even after installing asciidoc
> 
> ```
> a2x --format=manpage doc/lightningd-config.5.txt
> a2x: ERROR: "xmllint" --nonet --noout --valid "/Users/rpickhardt/hacken/tmp/lightning/doc/lightningd-config.5.xml" returned non-zero exit status 4
> ```
> 
> web search was of no help I guess I am missing some dependencies.

I got this too recently.  But Works Now, so I'm adding a simple fixup.

rustyrussell:
check-source got upset because it thought your comment was part of the BOLT quote.  Split.

renepickhardt:
I can autosquash that patch to be one commit in about 1 hour and rebase again if necessary so that it is ready to be merged.

I did not know you test for bolt quotes. That is beautiful!

practicalswift:
Nit: Typo in PR title and commit message. Should be "possibility" :-)

renepickhardt:
am I supposed to [change the commit message](https://help.github.com/en/articles/changing-a-commit-message) now?

ZmnSCPxj:
Yes. :)

renepickhardt:
@ZmnSCPxj @practicalswift I can't believe I actually did that (and while so I also fixed configer to configure)

rustyrussell:
Rebased to squash fixup and fixed your commit msg

rustyrussell:
... and again to fix merge conflict with master...

channeld/full_channel.c
```diff
@@ -443,8 +444,16 @@ static enum channel_add_err add_htlc(struct channel *channel,
...
+	 * funder of the channel to pay unnecessary onchain fees during a fee
+	 * spike with large commitment transactions.
+	 */
+	min_concurrent_htlcs = channel->config[recipient].max_accepted_htlcs;
+	if (min_concurrent_htlcs > channel->config[sender].max_accepted_htlcs)
+		min_concurrent_htlcs = channel->config[sender].max_accepted_htlcs;
 	if (tal_count(committed) - tal_count(removing) + tal_count(adding)
-	    > channel->config[recipient].max_accepted_htlcs) {
+	    > min_concurrent_htlcs) {
 		return CHANNEL_ERR_TOO_MANY_HTLCS;
```
SomberNight: Seems to me that the spec specifies only counting htlc in one direction (not both) against this limit, further it only specifies comparing against the recipient's constraint.
Obviously you can enforce arbitrary restrictions for outgoing HTLCs, but if you enforce stricter restrictions than the spec says for incoming ones and then you force-close when (not if) they are violated by the remote then that's not ideal?

See what eclair does (as above) here:
https://github.com/ACINQ/eclair/blob/eae113f0988069044dc268d8a7f16728563baa5f/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala#L216
https://github.com/ACINQ/eclair/blob/eae113f0988069044dc268d8a7f16728563baa5f/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala#L259
(check only one direction of HTLCs against only one of the constraints)

SomberNight: Oh, I've just realised that the channel does not actually get force-closed, only the peer get disconnected from (and then the htlc is not added as it was not committed).
BOLT-02 specifically says "fail the channel" - I guess I had a different interpretation of that...
I mean BOLT-02 sometimes says "fail the connection" and sometimes "fail the channel", but then that means c-lightning considers those both to be identical?
Re the original remark, I guess it's less bad then, but still not ideal.

SomberNight: Note: my comments have been addressed in https://github.com/ElementsProject/lightning/pull/4432 which reverted the strict checks for remote behaviour

SomberNight: Actually, only one of the two concerns have been addressed by #4432.
AFAICT c-lightning is still checking the total number of htlcs (sum of both directions) against `max_accepted_htlcs`, while I think BOLT-02 only specifies to check the number of htlcs the sender offered (so one direction only).
This is means c-lightning is still stricter than e.g. eclair.
https://github.com/ElementsProject/lightning/blob/9470ea3032a00438504e3769a65212fe22165717/channeld/full_channel.c#L575-L578
Or am I misunderstanding the code here? It seems to me `committed`, `removing`, and `adding` include htlcs in both directions.
@rustyrussell

PR #2859, opened 2019-07-28
lightningd: Check bitcoind version when `setup_topology`

### LINKED ISSUE #2697: Check bitcoind version atstartup
Root cause of #2687 was that user had unsupported bitcoind version.  So we should probably determine what bitcoind versions we need and check for those specifically.

trueptolemy:
Fix #2697 .
Corresponding to the [comment](https://github.com/ElementsProject/lightning/issues/2697#issuecomment-515762045), this PR add the 
check that the bitcoind version must be at least v0.15.0, and any older version will let `lightningd` down.
Related changes include:
- Add the `cli_min_supported_version` field in `chainparams`:
it's numeric version for `bitcoind`, now set all `cli_min_supported_version` as 150000;
- Call `getnetworkinfor` (call `getinfor` for the version older than v0.16.0) to get the numeric version and check if the version is supported.


lightningd/bitcoind.c
```diff
@@ -783,6 +783,85 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
...
+static bool process_getclientversion(struct bitcoin_cli *bcli)
+{
+	u64 version;
+	u64 min_version = bcli->bitcoind->chainparams->cli_min_supported_version;
+
+	if (*bcli->exitstatus != 0) {
+		/* bitcoin/src/rpc/protocol.h:
+		 * 	RPC_METHOD_NOT_FOUND = -32601
+		 */
+		if (*bcli->exitstatus != 32601)
```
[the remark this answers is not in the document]
trueptolemy: emm, it should be `89`. We use `WEXITSTATUS` to return low-order 8 bits of the exit status value. For `32601`, the corresponding exitstatus is `89`.(I not sure why it is a positive number).
And I confirmed again, from v0.14.0, `getnetworkinfo` was available, and `getinfo` was deprecated and was removed in v0.16.0. So I should use `getnetworkinfo` here.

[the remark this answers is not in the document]
trueptolemy: What about `fatal` for the case if `exitstatus==89`(this means the version older than v0.14), and call `bcli_failure` for other non-zero exitstatus?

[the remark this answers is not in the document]
ZmnSCPxj: Would it work to just try `getinfo` and if that fails (regardless of failure reason) then `getnetworkinfo`?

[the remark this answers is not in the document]
trueptolemy: Yes, it will work. But there are much more people runs the newer(than v0.14.0) version, and the `getinfo` has the very high probability to failure. Maybe we call `getnetworkinfo` first is better?
If `89` is ugly, we can remove  this `exitstatus==89` code block, and just (use `getnetworkinfo`) call `bcli_failure` for all non-zero exitstatus.
Thus, we just set `nonzero_exit_ok` parameter as `false` when we call `start_bitcoin_cli`, and any non-zero exitstatus will trigger `bcli_failure`.
What do you think about the code now?

cdecker:
Looks good to me, I think we can just fail with versions <0.14.0, since we don't support them anyway. The error message makes sense so that should work like it is.

ACK https://github.com/ElementsProject/lightning/pull/2859/commits/df8770c2bb3926496d5709db9bc5b0f3b7e1ffe1

lightningd/bitcoind.c
```diff
@@ -783,6 +783,66 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
...
+				     bcli->output_bytes,
+				     &version)) {
+		fatal("%s: Unable to getclientversion (%.*s)",
+		      bcli_args(tmpctx, bcli),
+		      (int)bcli->output_bytes,
+		      bcli->output);
+	}
+
+	if (version < min_version)
+		fatal("Unsupported client version? client version: %"PRIu64","
```
rustyrussell: The word "bitcoind" needs to appear in this message message somewhere?  Otherwise I'd have no idea what client you're referring to!

lightningd/bitcoind.c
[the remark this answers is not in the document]
ZmnSCPxj: Looks ok.

lightningd/bitcoind.c
trueptolemy: Yes! Corrected.

bitcoin/chainparams.c
```diff
@@ -86,6 +90,7 @@ const struct chainparams networks[] = {
      .rpc_port = 9332,
      .cli = "litecoin-cli",
      .cli_args = NULL,
+     .cli_min_supported_version = 150000,
```
arowser: litecoin seems not release 0.15 version.

lightningd/bitcoind.c
```diff
@@ -783,6 +783,66 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
...
+
+	tokens = json_parse_input(output, output, output_bytes, &valid);
+	if (!tokens)
+		fatal("%s: %s response",
+		      bcli_args(tmpctx, bcli),
+		      valid ? "partial" : "invalid");
+
+	if (tokens[0].type != JSMN_OBJECT) {
+		log_unusual(bcli->bitcoind->log,
+			    "%s: gave non-object (%.*s)?",
```
arowser: "(%.*s)?"   wrong format?

ZmnSCPxj: Correct format.  `%.*s` means get two arguments: first is length of string, second is string (which might not be null terminated).

arowser: Thanks, glad to known new tricks.

PR #2860, opened 2019-07-28
Add optional start_time and end_time filtering parameters to listforwards

kristapsk:
It's useful especially in case of a lot of forwarding happening through node. In my [feereport plugin](https://github.com/kristapsk/lightning-feereport) I'm only interested in last 30 days of forwards, not whole history.

Also adds generic support for time parameters. Currently accepts integer timestamps, but can be modified to support decimals and even human readable time strings in the future.

Couldn't get `tools/update-mocks.sh` working, modified `wallet/test/run-wallet.c` by hand.

Tested with my mainnet node.

### DIFF
--- a/common/json.c
+++ b/common/json.c
@@ -128,6 +128,17 @@ bool json_to_bool(const char *buffer, const jsmntok_t *tok, bool *b)
 	return false;
 }
 
+bool json_to_timeabs(const char *buffer, const jsmntok_t *tok, struct timeabs *t)
+{
+	uint64_t u64;
+
+	if (!json_to_u64(buffer, tok, &u64))
+		return false;
+	*t = timeval_to_timeabs(timerel_to_timeval(time_from_sec(u64)));
+
+	return true;
+}
+
 u8 *json_tok_bin_from_hex(const tal_t *ctx, const char *buffer, const jsmntok_t *tok)
 {
 	u8 *result;
--- a/common/json.h
+++ b/common/json.h
@@ -3,6 +3,7 @@
 #include "config.h"
 #include <bitcoin/preimage.h>
 #include <ccan/tal/tal.h>
+#include <ccan/time/time.h>
 #include <stdbool.h>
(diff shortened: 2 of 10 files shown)


rustyrussell:
Sorry, in general we're steering away from this kind of call filtering (except in the case where lists are really huge), in favor of plugins which do the filtering.

For example, most people would want to listforwards by channel source, or destination.  We end up with an explosion of selection options.

The way I want to go with these APIs is:

1. All objects we can list contain (at least one) unique, stable index.  For listforwards it's assigned at db insertion time.
2. All list* RPCs take optional fromindex= and count=, so they can reliably walk these lists.
3. We add a general `waitindex` RPC which waits until a given index is reached, so you don't have to poll.
4. A plugin can now easily maintain the list you want, and provide access to it.

Note that this is how waitanyindex works, but I think we should generalize it?

Thoughts?

The correct way of handling this is actually to provide a stable index for all forwards, then provide a

kristapsk:
> except in the case where lists are really huge

My idea was that this list could get huge for old long running nodes doing a lot of forwards. Don't know how accurate is that but [LNBIG claims doing 200-300 forwards per day](https://www.reddit.com/r/btc/comments/bxxiul/lnbig_the_lns_largest_source_of_liquidity/). Before working on this I looked at the [LND gRPC API Reference](https://api.lightning.community/) and their [`ForwardingHistory`](https://api.lightning.community/#forwardinghistory) RPC is the only one that have similar time based filtering as this PR, likely because of the possibility of list growing too big.

> All list* RPCs take optional fromindex= and count=, so they can reliably walk these lists.

Idea of having some universal filtering and limiting for all RPCs that return lists of course looks better. But I would suggest adding toindex= too, there may be a lot of cases, including my feereport plugin, where you just need from..to filtering and paging would just complicate things. But, of course, there could be gazillions of forwards in a single month for a single node in the future, then that would be necessary anyway, most likely. :)

Issue #2850, opened 2019-07-26
build fail on freebsd11


arowser:
### Issue and Steps to Reproduce

Tested with 6c7a45623e6c2367315f8db58f8faf0c47592943 on freebsd 11.2
./configure
gmake

### error log
```
tools/generate-wire.py --page impl tools/test/gen_test.h test_type < tools/test/test_cases > tools/test/gen_test.c
tools/test/gen_test.c:
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 6: Missing dependency operator
make[1]: Unknown modifier ' '
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 7: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 8: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 22: Missing dependency operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 25: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 27: Missing dependency operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 29: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 31: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 33: Missing dependency operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 35: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 37: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 39: Missing dependency operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 41: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 43: Missing dependency operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 46: Need an operator
make[1]: "/home/vagrant/workspace/c-lightning-x86-freebsd11/Makefile" line 48: Missing dependency operator
[...]

PR #2861, opened 2019-07-29
FreeBSD: fix build.

rustyrussell:
This makes it build for me on FreeBSD 11:
1. $(MAKE) has to passed through into update-mocks.
2. FreeBSD sed doesn't turn \n into a newline on RHS.
3. Bash and mako dependencies were missing from INSTALL.md

Fixes: #2850


ZmnSCPxj:
ACK a604444

PR #2862, opened 2019-07-29
lightningd: fix two crashes in plugin controller

rustyrussell:
Several corner cases I hit in practice.

### DIFF
--- a/tests/plugins/slow_init.py
+++ b/tests/plugins/slow_init.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+from lightning import Plugin
+import time
+
+plugin = Plugin()
+
+
+@plugin.init()
+def init(options, configuration, plugin):
+    plugin.log("slow_init.py initializing")
+    time.sleep(1)
+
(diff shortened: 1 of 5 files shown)


tests/plugins/slow_init.py
```diff
@@ -0,0 +1,14 @@
...
+from lightning import Plugin
+import time
+
+plugin = Plugin()
+
+
+@plugin.init()
+def init(options, configuration, plugin):
+    plugin.log("slow_init.py initializing")
+    time.sleep(1)
```
ZmnSCPxj: Would the test be reliable with just a 1-second sleep here?

darosior:
ACK 63bc5f9124d80a1802b4222bb4a8c9f04d263ac9

tests/plugins/slow_init.py
cdecker: Should be ok, maybe `valgrind` might be a bit too slow on CI, but 1 second is usually plenty in these cases.

PR #2866, opened 2019-07-29
Pylightning: use an absolute path for RPC socket

darosior:
I've hit some issues when dynamically `start`ing plugins which could not talk to `lightningd` on init

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -504,8 +504,7 @@ class Plugin(object):
     def _init(self, options, configuration, request):
         self.rpc_filename = configuration['rpc-file']
         self.lightning_dir = configuration['lightning-dir']
-        path = os.path.join(self.lightning_dir, self.rpc_filename)
-        self.rpc = LightningRpc(path)
+        self.rpc = LightningRpc(os.path.abspath(self.rpc_filename))
         self.startup = configuration['startup']
         for name, value in options.items():
             self.options[name]['value'] = value


cdecker:
Could you expand on the issue that this is addressing? iirc this works because we `chdir` into the `--lightning-dir` before loading plugins, but shouldn't prepending the `--lightning-dir` also work? That is as long as we don't use shorthands or environment variables to specify the `lightning-dir` on the command line, in which case expanding those would be more appropriate I think.

darosior:
To reproduce:
- Modify `helloworld.py` to make a `plugin.rpc.getinfo()` on init
- Start `lightningd` with a custom directory
- `plugin start helloworld.py`

The plugin could not talk to `lightningd` because the `os.path.join(lndir, rpc-file)` did not return the right socket path. Maybe the `chdir` is not done on dynamic start ? In this case would it be better to pass an absolute path to the `lightning-dir` entry in the `configuration` object of init ?

cdecker:
Sounds strange, is the `lightning-dir` entry in `configuration` being set correctly? Because the `rpc-file` is supposed to be always in the `lightning-dir` and that's what the `os.path.join` does.

darosior:
Sorry, I deleted the comment because I did not test.

darosior:
iirc the `lightning_dir` was relative. We might just dont `chdir` in `plugins_start` but only `plugins_init`. I'll check once I cab acces my laptop.

darosior:
I'm not even able to reproduce it, and I don't remember why I did that. Sorry for the false positive, closing it.

PR #2867, opened 2019-07-29
libplugin: add new 'getmanifest' field and pass configuration to init callback

darosior:
This adds the management of the two fields added in #2771 to libplugin and defaults the `dynamic` field to `false` for `pay` and `autoclean`.

For the `startup` field in `init` I added the configuration JSON token to the plugin's callback : so that if we add new fields they can be accessed too.

### DIFF
--- a/plugins/autoclean.c
+++ b/plugins/autoclean.c
@@ -90,7 +90,7 @@ static const struct plugin_command commands[] = { {
 int main(int argc, char *argv[])
 {
 	setup_locale();
-	plugin_main(argv, init, commands, ARRAY_SIZE(commands),
+	plugin_main(argv, init, PLUGIN_RESTARTABLE, commands, ARRAY_SIZE(commands),
 		    plugin_option("autocleaninvoice-cycle",
 				  "string",
 				  "Perform cleanup of expired invoices every"
--- a/plugins/autoclean.c
+++ b/plugins/autoclean.c
@@ -65,7 +65,8 @@ static struct command_result *json_autocleaninvoice(struct command *cmd,
 					   expired_by, cycle_seconds));
 }
 
-static void init(struct plugin_conn *prpc)
+static void init(struct plugin_conn *prpc,
+		  const char *buf UNUSED, const jsmntok_t *config UNUSED)
 {
 	rpc = prpc;
 
--- a/plugins/libplugin.c
+++ b/plugins/libplugin.c
@@ -513,21 +513,24 @@ static struct command_result *handle_init(struct command *init_cmd,
 					  const char *buf,
 					  const jsmntok_t *params,
 					  const struct plugin_option *opts,
-					  void (*init)(struct plugin_conn *))
+					  void (*init)(struct plugin_conn *,
+						       const char *buf, const jsmntok_t *))
 {
-	const jsmntok_t *rpctok, *dirtok, *opttok, *t;
+	const jsmntok_t *configtok, *rpctok, *dirtok, *opttok, *t;
 	struct sockaddr_un addr;
 	size_t i;
 	char *dir;
 	struct json_out *param_obj;
 
+	configtok = json_delve(buf, params, ".configuration");
+
 	/* Move into lightning directory: other files are relative */
-	dirtok = json_delve(buf, params, ".configuration.lightning-dir");
+	dirtok = json_delve(buf, configtok, ".lightning-dir");
 	dir = json_strdup(tmpctx, buf, dirtok);
 	if (chdir(dir) != 0)
 		plugin_err("chdir to %s: %s", dir, strerror(errno));
 
-	rpctok = json_delve(buf, params, ".configuration.rpc-file");
+	rpctok = json_delve(buf, configtok, ".rpc-file");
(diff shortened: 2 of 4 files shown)


cdecker:
Another excellent addition @darosior :+1:

ACK https://github.com/ElementsProject/lightning/pull/2867/commits/3638a300c18354bf3aac1c52e9d15b3688525d42

plugins/libplugin.c
```diff
@@ -501,6 +502,7 @@ handle_getmanifest(struct command *getmanifest_cmd,
 		json_out_end(params, '}');
 	}
 	json_out_end(params, ']');
+	json_out_addstr(params, "dynamic", dynamic ? "true" : "false");
```
rustyrussell: json_add_bool?

plugins/autoclean.c
```diff
@@ -90,7 +90,7 @@ static const struct plugin_command commands[] = { {
 int main(int argc, char *argv[])
 {
 	setup_locale();
-	plugin_main(argv, init, commands, ARRAY_SIZE(commands),
+	plugin_main(argv, init, false, commands, ARRAY_SIZE(commands),
```
rustyrussell: This can be dynamic, BTW.  In fact, it makes most sense that way.

plugins/pay.c
```diff
@@ -1311,5 +1311,5 @@ static const struct plugin_command commands[] = { {
 int main(int argc, char *argv[])
 {
 	setup_locale();
-	plugin_main(argv, init, commands, ARRAY_SIZE(commands), NULL);
+	plugin_main(argv, init, false, commands, ARRAY_SIZE(commands), NULL);
```
rustyrussell: This can be dynamic too, though restarting it clears pay history.

plugins/libplugin.h
```diff
@@ -147,7 +147,8 @@ char *charp_option(const char *arg, char **p);
 
 /* The main plugin runner: append with 0 or more plugin_option(), then NULL. */
 void NORETURN LAST_ARG_NULL plugin_main(char *argv[],
-					void (*init)(struct plugin_conn *rpc),
+					void (*init)(struct plugin_conn *rpc, const jsmntok_t *),
+					const bool dynamic,
```
ZmnSCPxj: Mildly suggest defining an `enum plugin_restartability` type with `PLUGIN_STATIC` `PLUGIN_RESTARTABLE` rather than an otherwise  featureless Boolean.

plugins/libplugin.c
darosior: I would have to include a lightningd/ header, is it ok or libplugin should not depend on lightningd/ files ?

darosior: Anyway, the use of an enum instead of a bool as proposed below settled this

darosior:
Made `pay` and `autoclean` dynamic. Used an `enum` instead of booleans (thus did not change to `json_add_bool`).

plugins/autoclean.c
```diff
@@ -65,7 +65,7 @@ static struct command_result *json_autocleaninvoice(struct command *cmd,
 					   expired_by, cycle_seconds));
 }
 
-static void init(struct plugin_conn *prpc)
+static void init(struct plugin_conn *prpc, const jsmntok_t *config UNUSED)
```
ZmnSCPxj: The `jsmntok_t` is not useable without the `const char *buf` that contains the actual JSON data.  Notice how `json_delve` et al. accept a `buffer` argument.

plugins/libplugin.c
```diff
@@ -547,7 +549,7 @@ static struct command_result *handle_init(struct command *init_cmd,
 					  take(param_obj),
 					  &rpc_conn,
 					  ".allow-deprecated-apis"),
-				"true");
+					  "true");
```
ZmnSCPxj: Also, the original, indent of this line is correct; it is not a continuation of the argument list of the previous line.

plugins/autoclean.c
darosior: Thank you. Yes I noticed, I just did not think about using it enough before comitting..

darosior:
Added the buffer argument to the init callback, corrected the indent.

cdecker:
@darosior could you please squash the `fixup!` and then I think we're good to merge :-)

darosior:
Sure, did I use `fixup!` the right way ?

darosior:
Squashed

darosior:
Squashed, and rebased on top of master btw

plugins/libplugin.c
```diff
@@ -698,7 +700,9 @@ void plugin_log(enum log_level l, const char *fmt, ...)
 }
 
 void plugin_main(char *argv[],
-		 void (*init)(struct plugin_conn *rpc),
+		 void (*init)(struct plugin_conn *rpc,
```
ZmnSCPxj: This change makes the commit break, hurting bisectability.  Presumably *this* particular change should be in the next commit.

Or I could just squash it into a single large commit...

darosior:
Sorry I messed up a rebase, should be fixed now.

PR #2870, opened 2019-07-29
readme: First pass at homogenizing the readme a bit

cdecker:
We haven't touched the readme for quite some time, just randomly added to it,
and it's starting to show. This is my attempt at cleaning it up a bit (more to
come):

 - No longer discourage users from running on mainnet, we're way beyond that
   point.
 - No longer instruct users to build from source, when we have real binary
   releases, on the PPA, the releases page and the docker images.
 - Cut down on the docker specific instructions, they are taking a lot of room
   when only a minority will likely run them that way
 - Generally make the README more of a dispatch for more in-depth
   documentation rather than trying to address everything right on the
   front-page.
 - Add a bit of context about running on top of a pruned node

This is by no means a final version, but I needed to get started somewhere, and I welcome any suggestion on how we can make this shorter and better at getting new users to the information they need.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,279 +1,74 @@
 # c-lightning: A specification compliant Lightning Network implementation in C
 
-c-lightning is a [standard compliant][std] implementation of the Lightning
-Network protocol.
-The Lightning Network is a scalability solution for Bitcoin, enabling
-secure and instant transfer of funds between any two parties for any
-amount.
+c-lightning is a lighweight, highly customizable and [standard compliant][std] implementation of the Lightning Network protocol.
 
-[std]: https://github.com/lightningnetwork/lightning-rfc
-
-For more information about the Lightning Network please refer to
-http://lightning.network.
 
 ## Project Status
 
 [![Build Status][travis-ci]][travis-ci-link]
 [![Pull Requests Welcome][prs]][prs-link]
 [![Irc][IRC]][IRC-link]
-[![Documentation Status](https://readthedocs.org/projects/lightning/badge/?version=docs)](https://lightning.readthedocs.io/)
+[![Documentation Status](https://readthedocs.org/projects/lightning/badge/?version=docs)][docs]
 
-[travis-ci]: https://travis-ci.org/ElementsProject/lightning.svg?branch=master
-[travis-ci-link]: https://travis-ci.org/ElementsProject/lightning
-[prs]: https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat
-[prs-link]: http://makeapullrequest.com
-[IRC]: https://img.shields.io/badge/chat-on%20freenode-brightgreen.svg
-[IRC-link]: https://webchat.freenode.net/?channels=c-lightning
+This implementation has been in production use on the Bitcoin mainnet since early 2018, with the launch of the [Blockstream Store][blockstream-store-blog].
+We recommend getting started by experimenting on `testnet`, but the implementation is considered stable and can be safely used on mainnet.
 
-This implementation is still very much a work in progress.
-It can be used for testing, but __it should not be used for real funds__.
-We do our best to identify and fix problems, and implement missing
-features.
-
-Any help testing the implementation, reporting bugs, or helping with
-outstanding issues is very welcome.
-Don't hesitate to reach out to us on IRC at
-[#lightning-dev @ freenode.net][irc1], [#c-lightning @
-freenode.net][irc2], or on the implementation-specific mailing list
-[c-lightning@lists.ozlabs.org][ml1], or on the Lightning Network-wide
-mailing list [lightning-dev@lists.linuxfoundation.org][ml2].
-
-[irc1]: http://webchat.freenode.net/?channels=%23lightning-dev
-[irc2]: http://webchat.freenode.net/?channels=%23c-lightning
-[ml1]: https://lists.ozlabs.org/listinfo/c-lightning
-[ml2]: https://lists.linuxfoundation.org/mailman/listinfo/lightning-dev
+Any help testing the implementation, reporting bugs, or helping with outstanding issues is very welcome.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,220 +1,130 @@
...
-[IRC]: https://img.shields.io/badge/chat-on%20freenode-brightgreen.svg
-[IRC-link]: https://webchat.freenode.net/?channels=c-lightning
+[![Documentation Status](https://readthedocs.org/projects/lightning/badge/?version=docs)][docs]
 
-This implementation is still very much a work in progress.
-It can be used for testing, but __it should not be used for real funds__.
-We do our best to identify and fix problems, and implement missing
-features.
+This implementation has been in production use on the Bitcoin mainnet since early 2018, with the launch of the [Blockstream Store][blockstream-store-blog].
+We recommend getting started by experimenting on `testnet`, but the implementation is considered stable and can be safely used on mainnet.
```
renepickhardt: I guess the testnet link is only helpful for developers and not for endusers. When I was new to the project it took me almost 2 months to figure out there is regtest so we might want to add a link to this file https://github.com/shesek/spark-wallet/blob/master/doc/dev-regtest-env.md or at least copy those instructions to our doc respository.

README.md
```diff
@@ -1,220 +1,130 @@
...
-      libsqlite3-dev python python3 net-tools zlib1g-dev libsodium-dev \
-      python3-mako
-    git clone https://github.com/ElementsProject/lightning.git
-    cd lightning
-    ./configure
-    make
+ - Installation from the [Ubuntu PPA][ppa]
+ - Installation of a pre-compiled binary from the [release page][releases] on Github
+ - Using one of the [provided docker images][dockerhub] on the Docker Hub
+ - Compiling the source code yourself (suggested mainly for developers or if you need one of the still [unreleased features][changelog-unreleased])
```
renepickhardt: do we have those lines 
```
sudo apt-get update
    sudo apt-get install -y \
      autoconf automake build-essential git libtool libgmp-dev \
      libsqlite3-dev python python3 net-tools zlib1g-dev libsodium-dev \
      python3-mako
    git clone https://github.com/ElementsProject/lightning.git
    cd lightning
    ./configure
    make
``` 
somewhere else in the `doc` folder? if yes we should link to them (instead of the changelog) otherwise we should create such a file. This again was and is extremely helpful for people and something the look at. I might even suggest to leave this inside the readme

README.md
```diff
@@ -1,220 +1,130 @@
...
+Notice that running `lightningd` against a pruned node may cause some issues if not managed carefully, see [below](#pruning) for more information.
 
 You can start `lightningd` with the following command:
 
-    lightningd/lightningd --network=testnet --log-level=debug
+```bash
+lightningd --network=testnet --log-level=debug
+```
+
+Please refer to `lightningd --help` for all other command line options.
```
renepickhardt: we might want to explain the config-dir variable or at least mention where `~/.lightning`-dir will be placed by default.

README.md
```diff
@@ -1,220 +1,130 @@
...
+
+### JSON-RPC Interface
 
-### Listing all commands:
-`cli/lightning-cli help` will print a table of the API and lists the
-following commands
+c-lightning exposes a [JSON-RPC 2.0][jsonrpcspec] interface over a Unix Domain socket located in its home directory (default: `$HOME/.lightning`).
+The Unix Domain Socket has the advantage of not being exposed over the network by default, allowing users to add their own authentication and authorization mechanism, while still providing a fully functional RPC interface out of the box.
+
+You can use `lightning-cli help` to print a table of the available RPC methods that can be called.
```
renepickhardt: we might want to state that there is a difference between `lightning-cli help` and `lightning-cli --help` something that I completely did not get in the beginning.

README.md
```diff
@@ -1,220 +1,130 @@
...
+The JSON-RPC interface is also documented in the following manual pages:
```
renepickhardt: While I like the fact that a couple of manpages are linked I think it is more interesting to tell the people that they can access the manpage of a command with `lightning-cli help COMMAND_NAME` and that `lightning-cli help` only has a short overview for the commands in comparison to the real manpage

README.md
```diff
@@ -1,220 +1,130 @@
...
+* [delinvoice](doc/lightning-delinvoice.7.txt)
+* [getroute](doc/lightning-getroute.7.txt)
+* [sendpay](doc/lightning-sendpay.7.txt)
+* [pay](doc/lightning-pay.7.txt)
+* [listpays](doc/lightning-listpays.7.txt)
+* [decodepay](doc/lightning-decodepay.7.txt)
+
+For simple access to the JSON-RPC interface you can use the `lightning-cli` tool, or the [python API client](contrib/pylightning).
 
 ### Opening a channel on the Bitcoin testnet
```
renepickhardt: see above comment I would strongly suggest to make this for regtestnet.

darosior:
>  For example I believe `configure --enable-developer` might be something we should mention and not hide it in the contribution guide.

Even if the audience are developers, they might not be familiar with Lightning Network, `dev-*` commands are *potentially* dangerous, mostly used for testing and corner cases such as crashes. I don't think this is required to figure in a get started documentation such as a README.

cdecker:
Ok, maybe we should do a more fundamental change here: have the `README.md` just be a landing page with some general information about scope and goals of the project, and directing users to more in-depth documents, such as a getting started guide which includes the installation and start procedures, and a document that helps developers getting started with a developer setup.

I agree that having the manpages linked here is not as useful as showing users how to access them (teach to fish and all that), especially because the list is incomplete here.

README.md
rustyrussell: Yes, regtest is usually what you want for testing anyway.  There's also contrib/startup_regtest.sh.

README.md
rustyrussell: doc/INSTALL.md, yes!  A link would help here, for sure.

renepickhardt:
Also I just remembered that the lightningd/plugins repository should be mentioned on the readme. I think clightning is a platform for highly customizable lightning nodes and the plugin architecture is an important part for it. Thus we should emphasize on their existence (:

rustyrussell:
Folded and rebased.


Ack https://github.com/ElementsProject/lightning/commit/0f2e7eecbfb5ed2e2b1345c512271ea0ba92fb7c

We can add more post -rc1, see #2932

PR #2871, opened 2019-07-29
pylightning: Warn users of plugins that may break due to extra args

cdecker:
We recently noticed that the way we unpack the call arguments for hooks and
notifications in pylightning breaks pretty quickly once you start changing the
hook and notification params. If you add params they will not get mapped
correctly causing the plugin to error out.

This can be fixed by adding a `VAR_KEYWORD` argument to the calbacks, i.e., by
adding a single `**kwargs` argument at the end of the signature. This commit
adds a check that such a catch-all argument exists, and emits a warning if it
doesn't.

It also fixes up the plugins that we ship ourselves.

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -183,6 +183,19 @@ class Plugin(object):
             raise ValueError(
                 "Topic {} already has a handler".format(topic)
             )
+
+        # Make sure the hook callback has a **kwargs argument so that it
+        # doesn't break if we add more arguments to the call later on. Issue a
+        # warning if it does not.
+        s = inspect.signature(func)
+        kinds = [p.kind for p in s.parameters.values()]
+        if inspect.Parameter.VAR_KEYWORD not in kinds:
+            self.log(
+                "Notification handler {} for hook {} does not have a variable "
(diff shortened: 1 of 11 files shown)


contrib/pylightning/lightning/plugin.py
```diff
@@ -181,6 +181,19 @@ def add_subscription(self, topic, func):
             raise ValueError(
                 "Topic {} already has a handler".format(topic)
             )
+
+        # Make sure the hook callback has a **kwargs argument so that it
```
darosior: nit: `s/hook/notification` ?

rustyrussell:
Aside: BTW, we need to bump pylightning version!  We've added quite a lot since last release, and if you want portable plugins I think you'll need to detect some of that.

cdecker:
> Aside: BTW, we need to bump pylightning version! We've added quite a lot since last release, and if you want portable plugins I think you'll need to detect some of that.

Absolutely, I will bump it to 0.8 during next week's release :-)

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2871/commits/8b8d173c7d88656e2074f8bc6d70d9da443198ee

PR #2872, opened 2019-07-30
gossipd: don't use O_APPEND on the gossip_store.

rustyrussell:
We always know the length, so we don't need it.  It causes much extra work
when we want to delete a record, which I suspect may cause issues amongst
some users who've been seeing gossip_store corruption.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2872/commits/d0cc245a56be8c2f7bb67832bd8674c079282d33

PR #2876, opened 2019-07-30
devtools: fix credit script.

rustyrussell:
Lisa's git name is lower-case, whereas CHANGELOG.md uses upper case,
so it doesn't realize she's named a commit already.


niftynei:
ack https://github.com/ElementsProject/lightning/pull/2876/commits/88a05fdc26d160cfd7469ae99497eb7a65e7a611

PR #2877, opened 2019-07-30
bigsize cleanup, with real onion bug!

rustyrussell:
This separates bigsize from varint completely, and as a result I found a bug in sphinx :(

I also assumed that we want the same rules inside the onion as elsewhere: bigsize must be minimal representation or we'll reject it.


ZmnSCPxj:
ACK ddae991

PR #2882, opened 2019-07-30
Split the channels of listfunds into operational and non-operational channels

renepickhardt:
As pointed out [at the end of my most recent video](https://www.youtube.com/watch?v=Gyt4nxRHy04) (describing unilateral channel closes by playing around with the command line and log files) it seems to be really confusing for end users that channels which are not operational anymore are still depicted in the `listfunds` command. (For example to wait 100 blocks before channel data was forgotten in case a reorg happens)

I suggest a fix of this by splitting the channels into `channels` (which now only include channels with `channel_state NORMAL` and which are connected to a peer) and another list of `non operational channels` (which includes all other channels)

The only thing that is still confusing that an `ONCHAIN` fund is still counted twice. This API change should not be breaking as the `channel` field has not been touched but might contain less entries. The channels in the `non operational channels` will however contain two more entries which are the 
`connected` field and the `channel_state`. 

I could drop the `ONCHAIN` Channels from the output as the UTXO is already in the `outputs` field

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -18,6 +18,8 @@
 #include <inttypes.h>
 #include <lightningd/bitcoind.h>
 #include <lightningd/chaintopology.h>
+#include <lightningd/channel.h>
+#include <lightningd/channel_state.h>
 #include <lightningd/hsm_control.h>
 #include <lightningd/json.h>
 #include <lightningd/jsonrpc.h>
@@ -580,6 +582,41 @@ static const struct json_command listaddrs_command = {
 };
 AUTODATA(json_command, &listaddrs_command);
 
+static void json_add_channel_state(struct json_stream *response,
+			       const char *fieldname,
+			       const enum channel_state state)
+{
+	json_add_member(response, fieldname, true, "%s", channel_state_str(state));
+}
+
+static void json_add_channel_meta_data(struct json_stream *response, const struct channel *c, const struct peer *p)
+{
+	json_object_start(response, NULL);
+	json_add_node_id(response, "peer_id", &p->id);
+	if (c->state != CHANNELD_NORMAL || !c->connected){
+		json_add_channel_state(response, "state",c->state);
+		json_add_bool(response, "connected", false);
+	}
+
+	if (c->scid)
+		json_add_short_channel_id(response,
+					  "short_channel_id",
+					  c->scid);
+
+	json_add_amount_sat_compat(response,
+				   amount_msat_to_sat_round_down(c->our_msat),
+				   "channel_sat",
+				   "our_amount_msat");
+	json_add_amount_sat_compat(response, c->funding,
+				   "channel_total_sat",
+				   "amount_msat");
+	json_add_txid(response, "funding_txid",
+		      &c->funding_txid);
+	json_add_num(response, "funding_output",
+		     c->funding_outnum);
+	json_object_end(response);
+}
+
 static struct command_result *json_listfunds(struct command *cmd,
 					     const char *buffer,
 					     const jsmntok_t *obj UNNEEDED,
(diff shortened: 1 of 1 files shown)


wallet/walletrpc.c
```diff
@@ -580,6 +582,41 @@ static const struct json_command listaddrs_command = {
 };
 AUTODATA(json_command, &listaddrs_command);
 
+static void json_add_channel_state(struct json_stream *response,
+			       const char *fieldname,
+			       const enum channel_state state)
+{
+	json_add_member(response, fieldname, true, "%s", channel_state_str(state));
```
rustyrussell: json_add_string() is better here: add_member is low-level.

renepickhardt:
> OK, this change is problematic, as it's not backward compatible. 
 
I know I was breaking the API but isn't my change / suggestion as backwards compatible as your proposal of extending the entries in the `channel` field with a `usable` boolean flag? Or why is it less backwards compatible? The `outputs ` and `channel` fields are not touched. just some entries are moved away from the `channel` field which IMHO should never have been there to begin with.

I can change my proposed API but my feeling was that also for users it is very confusing to have non operational channels within the list funds command.  Thus I would like to get a response to my concern before doing so. 

Also it is not the first time that the API of list-funds breaks. I just updated to lightning-0.7.1 and [this simple shell script](https://github.com/renepickhardt/lightning-helpers/blob/master/listfunds/listfunds.sh) does not work anymore (have not looked why). To be fair in todays world a plugin would be better to replace this plugin, also I realized that the output of this script have been wrong as funds could appear in the `channel` field as well as the `outputs` field.

rustyrussell:
> Also it is not the first time that the API of list-funds breaks. I just updated to lightning-0.7.1 and [this simple shell script](https://github.com/renepickhardt/lightning-helpers/blob/master/listfunds/listfunds.sh) does not work anymore (have not looked why). 

We added a space in our JSON formatting.  (`channel_sat" : 40565` not `channel_sat": 40565`).  Still, it's a regression, and I'll fix it, thanks for the report!

rustyrussell:
> I know I was breaking the API but isn't my change / suggestion as backwards compatible as your proposal of extending the entries in the `channel` field with a `usable` boolean flag? Or why is it less backwards compatible? The `outputs ` and `channel` fields are not touched. just some entries are moved away from the `channel` field which IMHO should never have been there to begin with.

You're always allowed to add new fields.  Removing them is an API break.

Your change removed entries from `channels`, which is a breaking change.  If we wanted to do this, we'd have to leave them there until we deprecate them, and have an additional array.  During the transition, non-operation channels would appear in both.

Hence the simple flag.  BTW, your script should use `jq`: it's a complex tool, but easily allows shell scripts to deal with JSON like this.

renepickhardt:
> You're always allowed to add new fields. Removing them is an API break.

This is how I intuitively would have understood it. Which is why I added a new field "non operational channels" instead of splitting the channel field into two lists "operational" and "non operational" ones. 

Sorry just learning that stuff as I am a mathematician and no software engineer. That is why I have to ask again. I did not remove a field I just removed the data / entries (potentially falsely?) listed in that field. That is already considered not to be backwards compatible? I made that choice because I thought explicitly this is backwards compatible. For example my shell script should not break with that change (which obviously is not an ultimate backwards compatibility test) but this is what I had in mind. 

Will create the boolean field though it might make sense to just add the channel status instead of my interpretation whether the channel is usable or not. What do you think? 

> BTW, your script should use jq: it's a complex tool, but easily allows shell scripts to deal with JSON like this.

As for my script and jq. I usually use jq but I wanted this script to work for newbies and be able to run without installing any dependencies and they might not have jq installed

ZmnSCPxj:
>  I did not remove a field I just removed the data / entries (potentially falsely?) listed in that field. That is already considered not to be backwards compatible?

Possibly yes: there may exist scripts that expect to find non-operational channels in the `channels` list, for some reason, and would have to be modified after this.  Not a big backwards-compat issue, but still a potential backwards-compat break.

PR #2883, opened 2019-07-30
Disabled uninitialized warnings with GCC

Chirimen-Jako:
### Changed
- Hide some false positive warnings by #pragma directive.
- Compatible both GCC and Clang. Fixed #2881

### DIFF
--- a/ccan/ccan/tal/path/path.c
+++ b/ccan/ccan/tal/path/path.c
@@ -321,8 +321,17 @@ fail_take_to:
 			goto fail;
 	}
 
+#pragma GCC diagnostic push
+#if defined(__has_warning) /* Clang */
+#if __has_warning("-Wmaybe-uninitialized")
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+#else /* GCC */
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
 	if (ret)
 		ret[len] = '\0';
+#pragma GCC diagnostic pop
 
 out:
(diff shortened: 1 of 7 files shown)


ccan/ccan/tal/path/path.c
```diff
@@ -321,8 +321,17 @@ char *path_rel(const tal_t *ctx, const char *from, const char *to)
 			goto fail;
 	}
 
+#pragma GCC diagnostic push
```
ZmnSCPxj: ccan is an upstream library: http://ccodearchive.net/

So we should push any updates to ccan https://github.com/rustyrussell/ccan/

Chirimen-Jako: Hi,
Thank you giving me ccan info. I'll try to update ccan too!
I don't like cluttered code too! But the root cause is sharing code both GCC and Clang.
And, I just want to reduce false positive warnings to raise more important warnings.
I don't mind the actual method in this case (initialize variable, add many preprocessor directives, or, change Makefile...). Anyway, Lightning users don't see it finally.
I hope the reviewers' policies are in line.

Chirimen-Jako: Hi ZmmSCPxj,
I pushed an updated to cpan https://github.com/rustyrussell/ccan  [Suppress false positive warning "-Wmaybe-uninitialized" with GCC -O3]: https://github.com/rustyrussell/ccan/pull/83
I'm in the status waiting for any reply. 
BR,

cdecker:
Was this addressed by PR #2884?

PR #2884, opened 2019-07-31
Travis upgrade and uninitialized warning fixes

rustyrussell:


rustyrussell:
Still futzing with travis bionic to get build working...

rustyrussell:
It works!  WIP removed, please review.

.travis.yml
```diff
@@ -1,20 +1,11 @@
...
-    packages:
-      - python3-pip
-      - libsqlite3-dev
-      - shellcheck
-      - cppcheck
-      - valgrind
-      - libomp-dev
-      - gcc-4.8
+before_install:
+  sudo apt-get install -y libsqlite3-dev cppcheck valgrind gcc-4.8
```
cdecker: It seems Travis-CI has completely phased out it's container based build environment which was the main reason we weren't using `sudo` (containers used to run much faster than the virtualized environment, but it no longer seems to be an option as of octobre[1]). So it seems we're stuck with a slower CI environment.

[1] https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures

Issue #2847, opened 2019-07-25
Node suddenly force-closed 40 channels - Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: received ERROR channel


pavlenex:
### Issue and Steps to Reproduce

I've noticed that the majority of the channels are now closed, pending or offline. For now, funds have not been retrieved, but probably because of 144 limbo.

### Logs
```
The chain is fully synched
Waiting /var/lib/tor/hidden_services/c-lightning/hostname to be created by tor...
announce-addr=fj2b35ny3kaxessk6su3ydbeh3ppqmxykqzbshesf43kzwgmmrjp2sid.onion:9735 added to /root/.lightning/config
C-Lightning starting, listening on port 9735
2019-07-25T15:01:35.198Z lightningd(6): --------------------------------------------------
2019-07-25T15:01:35.198Z lightningd(6): Server started with public key 02d35ad6428b083d63575e3b0ce33e06c507910c131970b07dc194953b43a9794c, alias BitcoinShirt (color #02d35a) and lightningd basedon-v0.7.1
2019-07-25T15:02:23.094Z lightningd(6): 03d37fca0656558de4fd86bbe490a38d84a46228e7ec1361801f54f9437a18d618 chan #9225: Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: received ERROR channel 47ac844a9b91f4365a4334164a3bf4f1cf688a24d1d0ff049bfecfcdb389ecc8: sync error
2019-07-25T15:02:23.094Z lightningd(6): 	 (tx edeaec247dbba4e1399b07130349cede852c40dfecb012940b9da329dd330765)
2019-07-25T15:02:23.148Z lightningd(6): 03d37fca0656558de4fd86bbe490a38d84a46228e7ec1361801f54f9437a18d618 chan #9225: State changed from CHANNELD_NORMAL to AWAITING_UNILATERAL
2019-07-25T15:02:23.643Z lightningd(6): 030995c0c0217d763c2274aa6ed69a0bb85fa2f7d118f93631550f3b6219a577f5 chan #4513: Peer permanent failure in CHANNELD_NORMAL: lightning_channeld: received ERROR channel 8bbf90cb5993fc3fc58949513a86664fe7ea90897e7d7fed9ff3de84d7540914: sync error
[...]

PR #2885, opened 2019-07-31
lightningd: avoid thundering herd on restart.

rustyrussell:
The reason lnd was sending sync error was that we were taking more than
30 seconds to send the channel_reestablish after connect.  That's
understandable on my test node under valgrind, but shouldn't happen normally.

However, it seems it has at least once,
(see https://github.com/ElementsProject/lightning/issues/2847)
: space out startup so it's less likely to happen.

Suggested-by: @cfromknecht

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1303,13 +1303,11 @@ static const struct json_command close_command = {
 };
 AUTODATA(json_command, &close_command);
 
-static void activate_peer(struct peer *peer)
+static void activate_peer(struct peer *peer, u32 delay)
 {
 	u8 *msg;
 	struct channel *channel;
 	struct lightningd *ld = peer->ld;
-	/* Avoid thundering herd: after first five, delay by 1 second. */
-	int delay = -5;
 
 	/* We can only have one active channel: make sure connectd
 	 * knows to try reconnecting. */
@@ -1330,7 +1328,6 @@ static void activate_peer(struct peer *peer)
 			channel_set_billboard(channel, false,
 					      "Attempting to reconnect");
 		}
-		delay++;
 	}
 
 	list_for_each(&peer->channels, channel, list) {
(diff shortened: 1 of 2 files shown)


cfromknecht:
@rustyrussell glad we figured it out, thanks for taking time to debug with me :)

ZmnSCPxj:
> Regarding the original bug do I understand it correctly that if I have a network error during reestablishment (e. G. Somebody pulls the cable) that the channel will end up force closed as or the peer will fail the channel / connection?

Not particularly, as, we only force-close if we receive an actual `error`, which can only occur if we are actually connected.  So the original bug seems to be more tied to trying to do too many things at the same time.

lightningd/peer_control.c
```diff
@@ -1308,15 +1308,29 @@ static void activate_peer(struct peer *peer)
 	u8 *msg;
 	struct channel *channel;
 	struct lightningd *ld = peer->ld;
+	/* Avoid thundering herd: after first five, delay by 1 second. */
+	int delay = -5;
```
ZmnSCPxj: Did you intend to make this variable `static`, or define it in `activate_peers` and pass it in by pointer to this function?

rustyrussell:
> (though I might have made it more symmetric. If the first 5 reconnects happened directly the next five should have happened in the next second.)

This is an approximation.  In practice there's a spread in response times, so the first 5 won't actually arrive all at once.  An ideal implementation would stop reconnecting until we're idle, and prioritize reconnections based on where HTLCs want to go.  But in practice it's just a few seconds and nobody cares, really.  But it makes it more reasonable to run 100 channels on an rPi.

The real bottleneck is still gossipd: after 0.7.2, I'll revisit our gossip catchup heuristics so we do selective probing rather than asking for a complete gossip dump when we're missing something.

lightningd/peer_control.c
rustyrussell: Err... yes... I was just checking that review was working? :laughing: 

Thanks, you, um, passed the test!  Fixed...

PR #2889, opened 2019-08-01
--daemonize fix

rustyrussell:
"This time for sure!"

I run with 'daemon' on my test node, and I noticed it very slow to shut down.


lightningd/lightningd.c
```diff
@@ -496,30 +496,37 @@ static void init_txfilter(struct wallet *w, struct txfilter *filter)
...
+	int ok_status = 0;
+
+	/* Don't hold files open. */
+	close(STDIN_FILENO);
+	close(STDOUT_FILENO);
+	close(STDERR_FILENO);
+
+	/* Many routines write to stderr; that can cause chaos if used
+	 * for something else, so set it here. */
+	if (open("/dev/null", O_WRONLY) != 0)
```
ZmnSCPxj: It might be OS specific that file descriptors are allocated in order.  Does POSIX assure us that closing `STDIN_FILENO` means the next `open` will return `STDIN_FILENO`?

rustyrussell: Absolutely, but we do check it's true.  Moreover, this is cut & paste from the previous ccan module we imported, so it's unchanged.

cdecker:
I really like that it gets rid of the DB reopening after fork, that was always a bit weird.

CI complains about `--help` calls not working:

```
lightningd: Failed to open PID file: No such file or directory
```

Probably a rather easy fix :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2889/commits/b94703d10d9cb3149d15b4bbb112dfa40178e277

PR #2890, opened 2019-08-01
Permuteroute

ZmnSCPxj:
This is a new command/routing heuristic, inspired  by @renepickhardt JIT-Routing.

It turns out to be an actual technique used in real-time strategy games (whose pathfinding issues actually surprisingly match that of LN: low acceptable latency for UX, dynamically-changing world, incomplete information (at least the really good games will not let your units know the "real" state of the world, only what your fog-of-war knows), large maps).

I will be posting a very lengthy article on lightning-dev mailinglist regarding permuteroute and other heuristics inspired by real-time strategy game techniques.

This is incomplete.  Help wanted.

* [x] Test for `permuteroute`, especially for route smoothing (i.e. backing out of a node when all alternatives are excluded).
* [x] Modify `pay` plugin to use `permuteroute` and fall back on `getroute` if failed / too expensive.

### DIFF
--- a/tests/test_gossip.py
+++ b/tests/test_gossip.py
@@ -1130,6 +1130,52 @@ def test_getroute_exclude(node_factory, bitcoind):
         l1.rpc.getroute(l4.info['id'], 1, 1, exclude=[chan_l2l3, chan_l2l4])
 
 
+@unittest.skipIf(not DEVELOPER, "gossip propagation is slow without DEVELOPER=1")
+def test_permuteroute(node_factory, bitcoind):
+    """Test permuteroute"""
+    l1, l2, l3, l4 = node_factory.line_graph(4, wait_for_announce=True)
+
+    # Initial route
+    route = l1.rpc.getroute(l4.info['id'], 1, 1)['route']
+
+    chan_l1l2 = route[0]['channel']
+
+    # Not sure what l3 and l4 id relationships are; just exclude both directions.
+    chan_l3l4_0 = route[2]['channel'] + '/0'
+    chan_l3l4_1 = route[2]['channel'] + '/1'
+
+    # Create a direct l2->l4 route.
+    l2.rpc.connect(l4.info['id'], 'localhost', l4.port)
+    scid = l2.fund_channel(l4, 1000000, wait_for_active=False)
+    bitcoind.generate_block(5)
+
+    # We don't wait above, because we care about it hitting l1.
+    l1.daemon.wait_for_logs([r'update for channel {}/0 now ACTIVE'
+                             .format(scid),
+                             r'update for channel {}/1 now ACTIVE'
+                             .format(scid)])
+
+    # l4 is > l2
+    chan_l2l4 = scid + '/0'
+
+    # Simulate a failure at l3->l4.
+    route2 = l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1])['route']
+    # This should be the path l1->l2->l4
+    assert len(route2) == 2
+    assert route2[0]['channel'] == chan_l1l2
+    assert route2[1]['channel'] == scid
+
+    # Simulate a failure at l2->l4 for the route2, but with
+    # l3->l4 path still valid.
+    route3 = l1.rpc.permuteroute(route, 1, exclude=[chan_l2l4])['route']
+    # This should be same as original path l1->l2->l3->l4
+    assert route == route3
+
+    # Simulate impossible case.
+    with pytest.raises(RpcError):
+        l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1, chan_l2l4])['route']
+
+
 @unittest.skipIf(not DEVELOPER, "need dev-compact-gossip-store")
 def test_gossip_store_local_channels(node_factory, bitcoind):
     l1, l2 = node_factory.line_graph(2, wait_for_announce=False)
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -566,6 +566,24 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("getroute", payload)
 
+    def permuteroute(self, route, erring_index, exclude=[], fromid=None, maxhops=20):
+        """
+        Modify an existing failing {route}, with a known failing
+        channel after the {erring_index} hop.
+        {exclude} is an optional array of scid/direction to exclude.
+        If specified the route starts from {fromid} otherwise use
+        this node as source.
+        Limit the number of hops in the route to {maxhops}.
+        """
+        payload = {
+            "route": route,
+            "erring_index": erring_index,
+            "exclude": exclude,
+            "fromid": fromid,
+            "maxhops": maxhops
+        }
+        return self.call("permuteroute", payload)
+
     def help(self, command=None):
         """
         Show available commands, or just {command} if supplied.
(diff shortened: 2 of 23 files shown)


ZmnSCPxj:
Article on `permuteroute` and other routing techniques on lightning-dev: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-August/002095.html

tests/test_gossip.py
```diff
@@ -1106,6 +1106,52 @@ def test_getroute_exclude(node_factory, bitcoind):
...
+    l1.daemon.wait_for_logs([r'update for channel {}/0 now ACTIVE'
+                             .format(scid),
+                             r'update for channel {}/1 now ACTIVE'
+                             .format(scid)])
+
+    # l4 is > l2
+    chan_l2l4 = scid + '/0'
+
+    # Simulate a failure at l3->l4.
+    route2 = l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1])
```
darosior:
```suggestion
    route2 = l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1])['route']
```

tests/test_gossip.py
```diff
@@ -1106,6 +1106,52 @@ def test_getroute_exclude(node_factory, bitcoind):
...
+    # This should be the path l1->l2->l4
+    assert len(route2) == 2
+    assert route2[0]['channel'] == chan_l1l2
+    assert route2[1]['channel'] == scid
+
+    # Simulate a failure at l2->l4 for the route2, but with
+    # l3->l4 path still valid.
+    route3 = l1.rpc.permuteroute(route, 1, exclude=[chan_l2l4])
```
darosior:
```suggestion
    route3 = l1.rpc.permuteroute(route, 1, exclude=[chan_l2l4])['route']
```

tests/test_gossip.py
```diff
@@ -1106,6 +1106,52 @@ def test_getroute_exclude(node_factory, bitcoind):
...
+    # This should be same as original path l1->l2->l3->l4
+    assert route == route3
+
+    # Simulate impossible case.
+    with pytest.raises(RpcError):
+        l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1, chan_l2l4])
```
darosior:
```suggestion
        l1.rpc.permuteroute(route, 2, exclude=[chan_l3l4_0, chan_l3l4_1, chan_l2l4])['route']
```

darosior:
I gave it a look and especially to integrate it in `pay`, so should we use `getroute`, use `permaroute` on error, and then fall back to `getroute` if `permaroute`'s route is not acceptable ?

ZmnSCPxj:
Yes, that is approximately what is needed.

Some subtleties:

1.  If we keep repeating `permuteroute` on the same pivot (`erring_index` remains same) then the node might have just incoming capacity and we might be better off  backing off and using `getroute`.
2.  In case of a `NODE` (bit `0x2000` in `failure_code`) error, we should really add the *node* to the `exclude` list.  Unfortunately we do not support this yet (and we need a `node_is_routable` function to determine that a node is in the exclude list; alternately we just exclude every channel the node has).  So for initial cut of `pay` using `permuteroute` just falling back to `getroute` for `NODE` failures would be doable.  Also, for the purpose of `permuteroute` we should back off by one `erring_index` i.e. use `erring_index - 1` for `NODE` errors; also, a local `NODE` error (`erring_index == 0`) means our own node is breaking down and we should probably fail all `pay` anyway.

ZmnSCPxj:
Test passing whee!

ZmnSCPxj:
Re-marking as WIP.  I am reconceptualizing the search algorithm, as the current naive depth-first search algorithm is likely to rescan the same nodes over and over when at nodes with many channels.

Sketch:

* Make the `struct node` `djikstra` field a `union` among different search scratch spaces. `union { struct { struct amount_msat total_fee; struct amount_msat total_risk; } djikstra; } s;`
* Define a new `struct { struct chan *from_chan; int depth; } permute;` in the above union (since we cannot do `getroute` and `permuteroute` in parallel anyway).
* Have a fixed-length FIFO queue of `struct node *`.  When full and we want to add nodes we just drop the nodes (and probably break out of the scanning loop).
* Put the pivot as the first node in the queue, with `depth = 1` for that pivot.
* While the queue is not empty:
  * Get a node.
  * If it is a goal node, extract the channels to the pivot and exit the loop: we have found a route.
  * If node has `depth == PERMUTE_DEPTH + 1` then skip back to the top of the loop.
  * For each channel of the node:
     * If channel is not routable, skip to next channel.
     * Get the other node of the channel.
     * If the other node has `depth != 0` skip to the next channel.
     * Set the other node  `depth` to this node `depth` plus 1, and its `from_chan` to the channel.
     * Put the node into the queue.
     * If queue is full, exit this loop (continue the outer loop).

ZmnSCPxj:
Tweaked algorithm seems good.

I need help with someone who can run a mainnet node to compare `getroute` vs. `permuteroute` times.

rustyrussell:
In general, I prefer to adapt new low-level commands and do fancy things in plugins.

route permutation can be done by asking to route from the node before the erroring channel to a node afterwards.  That's currently no more efficient than doing a complete new route, since we don't limit our Dijkstra (we normally want to know if there's a possible route).  Here's a patch which adds that (untested!) though:

https://gist.github.com/rustyrussell/3c752686991fb73ce256eefabee2ee2a

ZmnSCPxj:
True, but the currently-exposed `getroute` only allows one destination.  Preferentially we want to have multiple points after the  break to connect to, as multiple possible destinations allows for earlier return (and our main point is to reduce the time it takes to get an alternate route).  Current `getroute` also runs from the destination to the source, so I am uncertain how to handle multiple potential destinations, hmm.  It probably can be made to work.

We also need to somehow expose `smoothen_route`.  Or if not that, expose a `recalcroute`, where we recompute the fees of a route.  The latter is preferred, because `recalcroute` is needed to handle `UPDATE` failure codes i.e. we should not be excluding channels on `UPDATE` failure codes, we should instead `recalcroute` and see if the recalculated route has different fees, and fail the attempt (and exclude the channel)  if the fee has not changed (meaning the `UPDATE` did not, in fact, change the costs).

`smoothen_route` can be done simply by performing the smoothening (i.e. if a node is duplicated in the route, cut from the first point to the duplicate point) and then doing a `recalcroute`.  Smoothening is needed when splicing paths together, because loops are just useless, but could be formed by a naive pathfinding algo.

In any case I am now working on `getroutequick`, so what is our plan for this now?

ZmnSCPxj:
> https://gist.github.com/rustyrussell/3c752686991fb73ce256eefabee2ee2a

As I noted there, this adds at least 4 bytes to each `struct node`, and if we are not careful about alignment issues on 64-bit may end up adding 8 bytes, for a feature of `getroute` that only really benefits `permuteroute`.

Now, the reason to limit the number of hops to scan is really to limit the number of nodes to scan, because O(n log n) on number of nodes scanned.  So a different thing we can do is to instead a `max_nodes_scanned`.  This requires a single counter on the stack rather than having an extra counter per node.  We can backward-compatibly default this to `U32_MAX`.  We can also report number of nodes scanned back over the RPC interface (we are maintaining the variable anyway during the Dijkstra run, might as well report it back).

Then we can do some test `getroute` runs to find out how many nodes we typically scan during a typical `getroute` run, then divide by 10 or something and use that as the `max_nodes_scanned` for `permuteroute`.

Then `permuteroute` can just `getroute` from the before-break node to the after-break node.  If this results in a route that passes through a later node before reaching the after-break node, route smoothening should be able to shorten the path.

PR #2891, opened 2019-08-01
Update to latest BOLT versions

rustyrussell:
This is required to get parsing of new onion TLV format, though it doesn't actually implement that.

It required some tool mods to expose otherwise-unused definitions: @niftynei ?


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2891/commits/0474dd3911d8d845f5034e978b5532ba74ae250d

niftynei:
late to the punch but lgtm!

PR #2892, opened 2019-08-02
Plugin cwd cleanup

rustyrussell:
Now we have dynamic plugins, the difference in cwd becomes an issue.  Make it uniform, and bandaid for now.


darosior:
Besides the above-mentioned fixs, looks good to me too.

PR #2893, opened 2019-08-02
channeld: delay sending channel_announcement by 60 seconds.

rustyrussell:
We currently send channel_announcement as soon as we and our
peer agree it's 6 blocks deep.  In theory, our other peers might
not have seen that block yet though, so delay a little.

This is mitigated by two factors:
1. lnd will stash any "not ready yet" channel_announcements anyway.
2. c-lightning doesn't enforce the 6 depth minimum at all.

We should not rely on other nodes' generosity or laxity, however!

Next release, we can start enforcing the depth limit, and maybe stashing
ones which don't quite make it (or simply enforce depth 5, not 6).


ZmnSCPxj:
I mildly suggest having a new option for this, but fine as-is.

ACK  f624ccb

cdecker:
This doesn't really help at all if the remote end is catching up with the blockchain, or catching up takes longer than the delay, but hopefully that's a rare thing nowadays.

ACK https://github.com/ElementsProject/lightning/pull/2893/commits/f624ccb7028d5e63bf9fcf93970e366a2732574e

PR #2897, opened 2019-08-03
json-rpc: Add size and cumulative size to dev-memdump

cdecker:
Since we are walking the entire allocation tree anyway, and access the tal
metadata anyway, we can just as well also track the size of the memory
allocations to simplify debugging of memory use.


ZmnSCPxj:
ACK  ad71256

PR #2899, opened 2019-08-04
doc/lightningd.8.txt: Initial manpage for lightningd itself.

ZmnSCPxj:
Closes: #1360

### DIFF
--- a/doc/lightningd.8.txt
+++ b/doc/lightningd.8.txt
@@ -1,0 +1,185 @@
+LIGHTNINGD(8)
+=============
+:doctype: manpage
+
+NAME
+----
+lightningd - Daemon for running a Lightning Network node
+
+SYNOPSIS
+--------
+lightningd [--conf=<config-file>] ['OPTIONS']...
+
+DESCRIPTION
+-----------
+*lightningd* starts the C-Lightning daemon, which implements a
+standards-compliant Lightning Network node.
+
+CONFIGURATION OPTIONS
+---------------------
+*--conf*='FILE'::
+  Specify configuration file.
+  If not an absolute path, will be relative from the lightning-dir
+  location.
+  Defaults to 'config'.
+*--lightning-dir*='DIR'::
+  Set the directory for the C-Lightning daemon.
+  Defaults to '$HOME/.lightning'.
+
+MORE OPTIONS
+------------
+Command line options are mirrored as configuration options in the
+configuration file, so 'foo' in the configuration file simply
+becomes '--foo' on the command line, and 'foo=bar' becomes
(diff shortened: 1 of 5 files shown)


cdecker:
> ACK [4d2c474](https://github.com/ElementsProject/lightning/commit/4d2c474eb96c91d1a66ba761fb5e27fe58c276f9)
> 
> It's great to have that ! I think I'll add a new page to the C-lightning's readthedocs too.

Would be great if we could find a way to automatically convert manpages to the markdown / restructured format used by readthedocs, and avoid duplication or documentation getting out of sync.

Do you have a good idea on how to get that working @darosior ?

darosior:
Seems possible with `pandoc`, we could convert all manpages to a readthedocs page, and then add it at build time. I can take care of that.
However I think a how-to page on readthedoc might be useful for newcomers, a little bit like Rusty's `helpme` plugin and this manpage. (I can take care of that too :-) )

doc/lightningd.8.txt
```diff
@@ -0,0 +1,185 @@
...
+
+The *getinfo* command in particular will return a 'blockheight'
+field, which indicates the block height to which *lightningd*
+has been synchronized to (this is separate from the block height
+that your *bitcoind* has been synchronized to, and will always
+lag behind *bitcoind*).
+You will have to wait until the 'blockheight' has reached the
+actual blockheight of the Bitcoin network.
+
+Before you can get funds offchain, you need to have some funds
```
rustyrussell: I wonder if we should recommend the `helpme` plugin at this point.  It does things like connect to nodes on the network for you, and suggest things.

But that might have to wait until it's more mature?

ZmnSCPxj: If it does not come standard with a C-Lightning release, it is hard to recommend in the C-Lightning manpage.  I am already wary about suggesting `dig` and `https://1ml.com` but could not find an easier way to  get to *some* node, but at least those are common tools that can be easily acquired.

rustyrussell: Agreed.  Let's add a bootstrap script.

rustyrussell:
I think we could convert to reStructured text and make manpages from that.  Anyone want to try?

darosior:
> I think we could convert to reStructured text and make manpages from that. Anyone want to try?

Yes I'v looked at this, but do we want to make doc from the manpages (as proposed above), or manpages from the doc (as you proposed) ?

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2899/commits/9b24fd88a3d5dd7423fb0c749c934b01943b7627

PR #2904, opened 2019-08-05
Plugins nit: Remove the 'signal_startup' member of the plugin struct

darosior:
It's here, but we don't use it.

I think I forgot to remove it while moving the `dynamic` field to the `configuration` object in the `init` call in #2771.


SimonVrouwe:
@darosior Maybe also remove it from `plugin.h` :-)

https://github.com/ElementsProject/lightning/blob/9dd314226dfaabaab6456b634deae499e9da07d8/lightningd/plugin.h#L34

PR #2907, opened 2019-08-06
Add docker ARM64 support

NicolasDorier:
This is boring code, mostly repeating what is already in arm32 build.

The notable difference is that I needed to bump sqlite version because their autoconf on older version was not supporting arm64.

### DIFF
--- a/contrib/linuxarm64v8.Dockerfile
+++ b/contrib/linuxarm64v8.Dockerfile
@@ -1,0 +1,114 @@
+# This dockerfile is meant to cross compile with a x64 machine for a arm64v8 host
+# It is using multi stage build: 
+# * downloader: Download litecoin/bitcoin and qemu binaries needed for c-lightning
+# * builder: Cross compile c-lightning dependencies, then c-lightning itself with static linking
+# * final: Copy the binaries required at runtime
+# The resulting image uploaded to dockerhub will only contain what is needed for runtime.
+# From the root of the repository, run "docker build -t yourimage:yourtag -f contrib/linuxarm64v8.Dockerfile ."
+FROM debian:stretch-slim as downloader
+
+RUN set -ex \
+	&& apt-get update \
+	&& apt-get install -qq --no-install-recommends ca-certificates dirmngr wget \
+     qemu qemu-user-static qemu-user binfmt-support
+
+WORKDIR /opt
+
+RUN wget -qO /opt/tini "https://github.com/krallin/tini/releases/download/v0.18.0/tini-arm64" \
+    && echo "7c5463f55393985ee22357d976758aaaecd08defb3c5294d353732018169b019 /opt/tini" | sha256sum -c - \
+    && chmod +x /opt/tini
+
+ARG BITCOIN_VERSION=0.17.0
+ENV BITCOIN_TARBALL bitcoin-$BITCOIN_VERSION-aarch64-linux-gnu.tar.gz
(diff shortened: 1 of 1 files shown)


NicolasDorier:
@darosior if you want to run this on x64, you need QEmu

This is what I use in my CI
```
sudo docker run --rm --privileged multiarch/qemu-user-static:register --reset
sed -i -e 's/#EnableQEMU //g' "contrib/linuxarm64v8.Dockerfile"
```

contrib/linuxarm64v8.Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+	&& apt-get install -qq --no-install-recommends ca-certificates dirmngr wget \
+     qemu qemu-user-static qemu-user binfmt-support
+
+WORKDIR /opt
+
+RUN wget -qO /opt/tini "https://github.com/krallin/tini/releases/download/v0.18.0/tini-arm64" \
+    && echo "7c5463f55393985ee22357d976758aaaecd08defb3c5294d353732018169b019 /opt/tini" | sha256sum -c - \
+    && chmod +x /opt/tini
+
+ARG BITCOIN_VERSION=0.17.0
```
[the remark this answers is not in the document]
NicolasDorier: Because that's what the other image use. I think this should all be changed at once in a single PR.
Anyway, the only reason for this dependency is to get `bitcoin-cli`, so the version does not matter too much.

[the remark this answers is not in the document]
darosior: Ok, thanks

contrib/linuxarm64v8.Dockerfile
```diff
@@ -0,0 +1,114 @@
...
+# * builder: Cross compile c-lightning dependencies, then c-lightning itself with static linking
+# * final: Copy the binaries required at runtime
+# The resulting image uploaded to dockerhub will only contain what is needed for runtime.
+# From the root of the repository, run "docker build -t yourimage:yourtag -f contrib/linuxarm64v8.Dockerfile ."
+FROM debian:stretch-slim as downloader
+
+RUN set -ex \
+	&& apt-get update \
+	&& apt-get install -qq --no-install-recommends ca-certificates dirmngr wget \
+     qemu qemu-user-static qemu-user binfmt-support
```
darosior: You need to install `python3-mako` too

NicolasDorier: why? I am able to build without it.

darosior: I was not and it's a new dependency which has been added for `generate_wire`. I may be wrong, I'm retesting.

darosior: At least I was wrong on the place I commented.. It's below at the next `apt install`.

darosior: Re-tested: I could not build without including `python3-mako` L52.

NicolasDorier:
Thanks for the review @darosior  I just added it. I will make a separate PR to fix the two other dockerfile which likely miss it.

NicolasDorier:
I was building on top of latest release, not master that may be why I was building without python3-mako

NicolasDorier:
Tested it, it build on my machine. Let me know.

darosior:
ACK 128ca72237062f32a823edb6ac5744a58ff77777

PR #2908, opened 2019-08-06
pytest: Make test_funding_cancel_race SLOW_MACHINE compatible

cdecker:
This test is spawning 100 nodes concurrently, which is a lot even when not
running with `valgrind`, especially when executing tests in parallel.


cdecker:
Had to fixup the missing wildcard import and rebased on top of `master`.

PR #2910, opened 2019-08-06
Backfill blocks in the database as needed for gossip

cdecker:
As noted in #2756 we were not closing old channels, i.e., channels below our
wallet birth height.The cause was that the outpoints where not re-initialized
upon restart, since we didn't store the result of the txout lookup in the
database at all, hence not looking for spends.

This PR addresses this by extending the way we backfill the blocks in the
database on demand as we get `channel_announcements`. Instead of fetching the
block header, the list of transaction IDs and then checking the single txout,
we now fetch the raw block, filter out any uninteresting outpoint, i.e., not
outpoints that are not P2WSH, and then look up all of the remaining outpoints
in `bitcoind`s UTXO.

This is much more involved, since we process the entire blocks and look up all
of the outpoints, instead of just the one we suspect to be present (which we
were told about by the `channel_announcement`), but it results in a more
complete view of the current UTXO that might be of interest.

So what we do here is basically implement our own version of the dreaded IBD
:wink:, but there are a number of advantages:

 - We minimize the number of `getblockhash` and `getblock` calls, whereas
   these were called for each potential channel in the legacy version we now
   just call them once, minimizing the impact of someone flooding us with
   bogus `channel_announcement`s.
 - Should we ever truncate the `gossip_store` then sync is incredibly fast.
 - Should we get a trickle of `channel_announcement`s they don't result in a
   call to `bitcoind` at all if we've seen a prior announcement for the same
   blockheight.
 - This table can easily be seeded and bundled with the download if we ever
   chose to.

To put a measure on how slow it really is I ran a number of experiments:

 - `ex1` is the baseline experiment with the previous method of
   `getblockhash` + `getblock true` + `gettxout` only for the specific
   outpoint. This obviously is fast, since we don't parse the entire block,
   and just fetch a single `txout`.
 - `ex2` is the implementation of `getfilteredblock` but without writing to
   the database, and therefore without looking up outpoints in the DB
   either. This has to be worse since we do a lot more work for each
   announcement but don't get any of the benefit.
 - `ex3` is the implementation with the lookup and storing the block and
   outpoints in the DB. Should be slightly better, but most requests are
   already queued before we get the results and could make use of the previous
   results.
 - `ex4` adds queuing of requests. The main goal is to reduce the pressure on
   `bitcoind`, but it also clears all pending requests for the same block once
   we get a result.
 - `ex5` gets up to 22k channels and then restarts to test how well resuming
   works if we nuke the `gossip_store` inbetween restarts. It catches right up
   in <30 seconds and continues from there.

The plot below shows the various times to catch up with the 32k channels in
the network starting from a fresh `--lightning-dir`:

getfiltered-benchmarks

([plot source][gist])

We could potentiall parallelize the `gettxout` calls again, keeping the
overall `getfilteredblock` concurrency limited, to avoid duplicate work. One
final consideration that is in favor of limiting the concurrency of
`getfilteredblock` is the fact that the two initial calls (`getblockhash` and
`getblock`) have a high call priority and may crowd out other, more important,
calls such as our blockchain head tracking. I didn't want to expose the
priority in those calls, but that might also be an option for later, such that
the head tailing can skip the backfill queue.

Closes #2756 

[gist]: https://gist.github.com/cdecker/7747f9f42dd3e4ffa9621d6a54cf7644

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -786,6 +786,7 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
 /* Context for the getfilteredblock call. Wraps the actual arguments while we
  * process the various steps. */
 struct filteredblock_call {
+	struct list_node list;
 	void (*cb)(struct bitcoind *bitcoind, struct filteredblock *fb,
 		   void *arg);
 	void *arg;
@@ -795,6 +796,11 @@ struct filteredblock_call {
 	size_t current_outpoint;
 	struct timeabs start_time;
 };
+
+/* Declaration for recursion in process_getfilteredblock_step1 */
+static void
+process_getfiltered_block_final(struct bitcoind *bitcoind,
+				const struct filteredblock_call *call);
 
 static void
 process_getfilteredblock_step3(struct bitcoind *bitcoind,
@@ -815,8 +821,7 @@ process_getfilteredblock_step3(struct bitcoind *bitcoind,
 				  process_getfilteredblock_step3, call);
 	} else {
 		/* If there were no more outpoints to check, we call the callback. */
-		call->cb(bitcoind, call->result, call->arg);
-		tal_free(call);
+		process_getfiltered_block_final(bitcoind, call);
 	}
 }
 
@@ -855,8 +860,7 @@ static void process_getfilteredblock_step2(struct bitcoind *bitcoind,
 	if (tal_count(call->outpoints) == 0) {
 		/* If there were no outpoints to check, we can short-circuit
 		 * and just call the callback. */
-		call->cb(bitcoind, call->result, call->arg);
-		tal_free(call);
+		process_getfiltered_block_final(bitcoind, call);
 	} else {
 
 		/* Otherwise we start iterating through call->outpoints and
@@ -880,6 +884,34 @@ static void process_getfilteredblock_step1(struct bitcoind *bitcoind,
 	bitcoind_getrawblock(bitcoind, blkid, process_getfilteredblock_step2, call);
 }
 
+/* Takes a call, dispatches it to all queued requests that match the same
+ * height, and then kicks off the next call. */
+static void
+process_getfiltered_block_final(struct bitcoind *bitcoind,
+				const struct filteredblock_call *call)
+{
+	struct filteredblock_call *c, *next;
+	u32 height = call->result->height;
+	/* Need to steal so we don't accidentally free it while iterating through the list below. */
+	struct filteredblock *fb = tal_steal(NULL, call->result);
+	list_for_each_safe(&bitcoind->pending_getfilteredblock, c, next, list) {
+		if (c->result->height == height) {
+			c->cb(bitcoind, fb, c->arg);
+			list_del(&c->list);
+			tal_free(c);
+		}
+	}
+	tal_free(fb);
+
+	/* Nothing to free here, since `*call` was already deleted during the
+	 * iteration above. It was also removed from the list, so no need to
+	 * pop here. */
+	if (!list_empty(&bitcoind->pending_getfilteredblock)) {
+		c = list_top(&bitcoind->pending_getfilteredblock, struct filteredblock_call, list);
+		bitcoind_getblockhash(bitcoind, c->result->height, process_getfilteredblock_step1, c);
+	}
+}
+
 void bitcoind_getfilteredblock_(struct bitcoind *bitcoind, u32 height,
 				void (*cb)(struct bitcoind *bitcoind,
 					   struct filteredblock *fb,
@@ -889,6 +921,8 @@ void bitcoind_getfilteredblock_(struct bitcoind *bitcoind, u32 height,
 	/* Stash the call context for when we need to call the callback after
 	 * all the bitcoind calls we need to perform. */
 	struct filteredblock_call *call = tal(bitcoind, struct filteredblock_call);
+	/* If this is the first request, we should start processing it. */
+	bool start = list_empty(&bitcoind->pending_getfilteredblock);
 	call->cb = cb;
 	call->arg = arg;
 	call->result = tal(call, struct filteredblock);
@@ -898,7 +932,9 @@ void bitcoind_getfilteredblock_(struct bitcoind *bitcoind, u32 height,
 	call->result->outpoints = tal_arr(call->result, struct filteredblock_outpoint *, 0);
 	call->current_outpoint = 0;
 
-	bitcoind_getblockhash(bitcoind, height, process_getfilteredblock_step1, call);
+	list_add_tail(&bitcoind->pending_getfilteredblock, &call->list);
+	if (start)
+		bitcoind_getblockhash(bitcoind, height, process_getfilteredblock_step1, call);
 }
 
 static bool extract_numeric_version(struct bitcoin_cli *bcli,
@@ -1112,6 +1148,7 @@ struct bitcoind *new_bitcoind(const tal_t *ctx,
 		bitcoind->num_requests[i] = 0;
 		list_head_init(&bitcoind->pending[i]);
 	}
+	list_head_init(&bitcoind->pending_getfilteredblock);
 	bitcoind->shutdown = false;
 	bitcoind->error_count = 0;
 	bitcoind->retry_timeout = 60;
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -787,7 +787,7 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
  * process the various steps. */
 struct filteredblock_call {
 	struct list_node list;
-	void (*cb)(struct bitcoind *bitcoind, struct filteredblock *fb,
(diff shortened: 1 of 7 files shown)


tests/test_db.py
```diff
@@ -15,3 +19,57 @@ def test_db_dangling_peer_fix(node_factory):
...
+    # tx in the same block (`line_graph` with `fundchannel=True` generates
+    # blocks).
+    l1, l2 = node_factory.line_graph(2, fundchannel=False)
+
+    # Get some funds to l1
+    addr = l1.rpc.newaddr()['bech32']
+    bitcoind.rpc.sendtoaddress(addr, 1)
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 1)
+    bitcoind.generate_block(1)
+    l1.daemon.wait_for_log(r'Owning')
```
rustyrussell: Where feasible, I prefer avoiding wait_for_log (of course, it's not always).  Waiting on listfunds should work here, no?

tests/test_db.py
```diff
@@ -15,3 +19,57 @@ def test_db_dangling_peer_fix(node_factory):
...
+    """
+    # Need to manually open the channels later since otherwise we can't have a
+    # tx in the same block (`line_graph` with `fundchannel=True` generates
+    # blocks).
+    l1, l2 = node_factory.line_graph(2, fundchannel=False)
+
+    # Get some funds to l1
+    addr = l1.rpc.newaddr()['bech32']
+    bitcoind.rpc.sendtoaddress(addr, 1)
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 1)
```
rustyrussell: Every time we thought we caught bitcoind being async, we were wrong AFAIK.  So this is probably redundant.

tests/test_db.py
```diff
@@ -15,3 +19,57 @@ def test_db_dangling_peer_fix(node_factory):
...
+    l1.daemon.wait_for_log(r'Owning')
+
+    # Now send the needle we will go looking for later:
+    bitcoind.rpc.sendtoaddress('bcrt1qtwxd8wg5eanumk86vfeujvp48hfkgannf77evggzct048wggsrxsum2pmm', 1)
+    l1.rpc.fundchannel(l2.info['id'], 10**6, announce=True)
+    wait_for(lambda: len(bitcoind.rpc.getrawmempool()) == 2)
+
+    # Confirm and get some distance between the funding and the l3 wallet birth date
+    bitcoind.generate_block(100)
+    wait_for(lambda: len(l1.rpc.listnodes()['nodes']) == 2)
```
rustyrussell: What does this do?  Seems like this will always be true, even if we haven't seen node_announcement?

tests/test_db.py
```diff
@@ -15,3 +19,57 @@ def test_db_dangling_peer_fix(node_factory):
...
+    # and call out to `bitcoind` to backfill the block.
+    l3 = node_factory.get_node()
+    heights = [r['height'] for r in l3.db_query("SELECT height FROM blocks")]
+    assert(103 not in heights)
+
+    l3.rpc.connect(l1.info['id'], 'localhost', l1.port)
+
+    wait_for(lambda: len(l3.rpc.listnodes()['nodes']) == 2)
+    heights = [r['height'] for r in l3.db_query("SELECT height FROM blocks")]
+    assert(103 in heights)
```
rustyrussell: Hmm, I would prefer an *explicit* test for the symptoms we're seeing: get unrelated nodes to make a channel, announce it.  Then they close channel, see if we still think we see it.

But we can add that test later.

lightningd/bitcoind.c
```diff
@@ -784,6 +785,124 @@ void bitcoind_gettxout(struct bitcoind *bitcoind,
...
+static void process_getfilteredblock_step1(struct bitcoind *bitcoind,
+					   const struct bitcoin_blkid *blkid,
+					   struct filteredblock_call *call)
+{
+	/* So we have the first piece of the puzzle, the block hash */
+	call->result->id = *blkid;
+
+	/* Now get the raw block to get all outpoints that were created in
+	 * this block. */
+	bitcoind_getrawblock(bitcoind, blkid, process_getfilteredblock_step2, call);
```
rustyrussell: How does this interact with pruning BTW?

lightningd/gossip_control.c
```diff
@@ -59,6 +59,34 @@ static void got_txout(struct bitcoind *bitcoind,
 	tal_free(scid);
 }
 
+static void got_filteredblock(struct bitcoind *bitcoind,
+		      struct filteredblock *fb,
+		      void *arg)
+{
+	struct short_channel_id *scid = (struct short_channel_id *)arg;
```
rustyrussell: Seems like it's actually a typesafe cb, so you can make scid an arg?

tests/test_db.py
cdecker: It will return true once we've seen and verified the `channel_announcement` this means that at that time we've definitely exchanged gossip for the channels and fetched the block if missing. I could also mock `getblock` on the nodes, but since we are more interested in the gossip being present when `l3` joins this shorthand works just as well.

lightningd/bitcoind.c
cdecker: It doesn't change the requirement for the blocks actually being present in `bitcoind`. Before we were asking `bitcoin-cli getblock xyz true` which parses the block in `bitcoind` and returns just a list of transaction IDs, now we call `getblock xyz false` and parse in `lightningd`. If the block we were looking for was pruned away we'd fail the lookup and discard the announcement. Now we do the same. Hence the warning that pruned nodes are not supported :-)

The reason I built `getfilteredblock` as a pseudo `bitcoin-cli` command is that this should make it easier to replace in a plugin with a custom method of retrieving blocks (web service, cache, ...) :wink:

lightningd/gossip_control.c
cdecker: That's what I was going for as well, but if I have the correct type in there I get this:

```
lightningd/gossip_control.c: In function ‘get_txout’:
ccan/ccan/typesafe_cb/typesafe_cb.h:33:2: error: passing argument 3 of ‘bitcoind_getfilteredblock_’ from incompatible pointer type [-Werror=incompatible-pointer-types]
  __builtin_choose_expr(      \
  ^
ccan/ccan/typesafe_cb/typesafe_cb.h:109:2: note: in expansion of macro ‘typesafe_cb_cast’
  typesafe_cb_cast(rtype (*)(__VA_ARGS__, atype),   \
  ^~~~~~~~~~~~~~~~
lightningd/bitcoind.h:162:8: note: in expansion of macro ‘typesafe_cb_preargs’
        typesafe_cb_preargs(void, void *, \
        ^~~~~~~~~~~~~~~~~~~
lightningd/gossip_control.c:122:3: note: in expansion of macro ‘bitcoind_getfilteredblock’
   bitcoind_getfilteredblock(topo->bitcoind, short_channel_id_blocknum(scid), got_filteredblock, scid);
   ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from lightningd/gossip_control.c:1:0:
lightningd/bitcoind.h:154:6: note: expected ‘void (*)(struct bitcoind *, struct filteredblock *, void *)’ but argument is of type ‘void (*)(struct bitcoind *, struct filteredblock *, struct short_channel_id *)’
 void bitcoind_getfilteredblock_(struct bitcoind *bitcoind, u32 height,
      ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

and I got stuck debugging macro expansions. Maybe you have a good idea what is happening?

tests/test_db.py
cdecker: Removed in https://github.com/ElementsProject/lightning/pull/2910/commits/5bab815d8158d5bd2df81468079dfd919dbfe3b5

tests/test_db.py
cdecker: Made the test more precise in https://github.com/ElementsProject/lightning/pull/2910/commits/5bab815d8158d5bd2df81468079dfd919dbfe3b5. This now includes looking for other P2WSH outputs in the same block and checking that the channel below `l3`s start height is also cleaned up correctly.

lightningd/gossip_control.c
rustyrussell:
```
void bitcoind_getfilteredblock_(struct bitcoind *bitcoind, u32 height,
				void (*cb)(struct bitcoind *bitcoind,
					   struct filteredblock *fb,
					   void *arg),
				void *arg);
#define bitcoind_getfilteredblock(bitcoind_, height, cb, arg)		\
	bitcoind_getfilteredblock_((bitcoind_),				\
				   (height),				\
				   typesafe_cb_preargs(void, void *,	\
						       (cb), (arg),	\
						       struct bitcoind *, \
						       const struct filteredblock *), \
				   (arg))
```

You convert the fn if it takes a `const struct filteredblock *`, but it doesn't.  Fixed it to take a const everywhere, because we can.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2910/commits/b4b0207df2dd82b8d6fa2467a9e091c194f030da

PR #2911, opened 2019-08-07
contrib: simple script to bootstrap node.

rustyrussell:
The helpme plugin is more comprensive, but this at least connects to a
few random nodes, and doesn't require python libraries in path or anything.

I selected the nodes from helpme.py, eliminating ones I couldn't reach.

Inspired by #2899


contrib/bootstrap-node.sh
```diff
@@ -0,0 +1,55 @@
...
+    exit 1
+fi
+
+if ! $LCLI "$@" -H getinfo | grep 'network=bitcoin'; then
+    echo "lightningd not running, or not on mainnet?" >&2
+    exit 1
+fi
+
+# Pick two random peers from this list, and connect to them.
+# IPV4: 024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605@128.199.202.168
```
ZmnSCPxj: Is this not arguably centralizing?  Making this script into a kingmaker?

I would consider instead:

* Use `dig lseed.bitcoinstats.com SRV`.  Maybe put a list of DNS seeds other than @cdecker seed (are there even any?).
* Write a small program to parse the bech32-encoding returned by SRV and extract the public key.
  * Or update `connect` so that bech32-encoded addresses like those returned by SRV can be determined directly.
* Get the public key via the program.  Let `connectd` query `lseed.bitcoinstats.com` for the IP address.

contrib/bootstrap-node.sh
```diff
@@ -0,0 +1,55 @@
...
+if type lightning-cli >/dev/null 2>&1; then
+    LCLI=lightning-cli
+elif [ -x cli/lightning-cli ]; then
+    LCLI=cli/lightning-cli
+else
+    echo "Cannot find lightning-cli!" >&2
+    exit 1
+fi
+
+if ! $LCLI "$@" -H getinfo | grep 'network=bitcoin'; then
```
ZmnSCPxj: :+1:

contrib/bootstrap-node.sh
rustyrussell: Not really.  Once you have your gossip, you can connect to anyone.  We just don't have that much influence.  And bitcoind also has a static fallback list, so we're at least in good company.

And connectd *will* query the seed for addresses, BTW.  But you need the pubkeys, which are harder to get.

ZmnSCPxj: Okay, then.  Suggest adding several more, to distribute the load.

ZmnSCPxj:
Hmmmm belated thought: do `contrib` scripts get installed at `make install`?

Issue #2791, opened 2019-07-03
close should force by default, increase timeout to 48 hours?


rustyrussell:
Suggested by: @JavierRSobrino

The current behavior of returning, but still trying to close, it not very useful.  It should probably, after a long default delay, do a unilateral close.

2 days is long enough to reconnect to anything likely to come back anytime soon, IMHO.

What do you think, @ZmnSCPxj ?

JavierRSobrino:
I can confirm this issue. External wallets like Spark Wallet use the normal close, so if you close an offline channel (which is the usual reason) it remains for ever in the state of awaiting connection with the default options. I opened a related issue in Spark Wallet here https://github.com/shesek/spark-wallet/issues/97

ZmnSCPxj:
Current `force=True` behavior is to unilaterally fail without even trying a mutual close. Maybe what we want is to have the timeout force a unilateral instead?

The issue is that if you wait 48 hours *before* initiating unilateral, you have to wait xx *more* blocks before you get funds.

Do we need more smarts? I would prefer keeping a dumb `close` (`directclose`?) that does just initiate a mutual/unilateral close, then have some smart `close` with timeouts and heuristics (default to a mutual close if connected, or wait and try `connect` if disconnected, then have some hour-long timeout to unilateral) in a `porcelain` builtin plugin.

JavierRSobrino:
I think it can be as simple is to force by default with a timeout of 48h, and inform the user that he has to wait for both the timeout to initiate the unilateral and  for the blocks.
Also an option to remove the close operation if it hasn't be committed yet anytime in the waiting period.

ZmnSCPxj:
I am uncertain what you imagine the `close` operation to be.  "Force" means drop the channel unilaterally, meaning posting our unilateral close tx to the Bitcoin mempool.  We cannot "remove" a close operation; once the peer has seen that the unilateral close tx has been broadcast and is now in its own mempool, it will presume that the close has been done and will no longer talk with us about the channel (i.e. it becomes unuseable already).  Timeout means little if you drop unilaterally --- your *real* timeout is started when it gets confirmed onchain, and ends when the locktime indicated by your peer ends.

If you `force` by default, then there is no point to doing a timeout --- it will be closed, and the ***real*** timeout is the locktime indicated by your peer.

JavierRSobrino:
[...]

PR #2914, opened 2019-08-07
close: change parameters, defaults

rustyrussell:
Fixes: #2791 

Painful transition, but I think a long-term win for sane defaults.

### DIFF
--- a/doc/lightning-close.7.txt
+++ b/doc/lightning-close.7.txt
@@ -8,13 +8,14 @@ lightning-close - Command for closing channels with direct peers
 
 SYNOPSIS
 --------
-*close* 'id' ['force'] ['timeout']
+*close* 'id' ['unilateraltimeout']
 
 DESCRIPTION
 -----------
 
 The *close* RPC command attempts to close the channel cooperatively
-with the peer.
+with the peer, or unilaterally after 'unilateraltimeout'.
+
 If the given 'id' is a peer ID (66 hex digits as a string), then
 it applies to the active channel of the direct peer corresponding to
 the given peer ID.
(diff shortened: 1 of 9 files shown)


doc/lightning-close.7.txt
```diff
@@ -8,36 +8,38 @@ lightning-close - Command for closing channels with direct peers
...
 If the given 'id' is a channel ID (64 hex digits as a string, or
 the short channel ID 'blockheight:txindex:outindex' form), then it
 applies to that channel.
 
-The *close* command will time out and return with an error when the
-number of seconds specified in 'timeout' is reached.
-If unspecified, it times out in 30 seconds.
+The *close* command will unilaterally close the channel when the
+number of seconds specified by 'unilateraltimeout' is reached,
+unless that is 0.  The default is 2 days (172800 seconds).
```
ZmnSCPxj: Suggest describing in more detail what happens if 'unilateraltimeout' is 0.

    If 'unilateraltimeout' is 0, then the *close* command will wait
    indefinitely until the peer is online and can negotiate a
    mutual close.

tests/test_closing.py
```diff
@@ -241,7 +240,7 @@ def test_closing_different_fees(node_factory, bitcoind, executor):
     # so increase the timeout so that this test will pass
     # when valgrind is enabled.
     # (close timeout defaults to 30 as of this writing)
```
ZmnSCPxj: Maybe change the comment as well?

tests/test_closing.py
```diff
@@ -105,11 +103,12 @@ def test_closing_while_disconnected(node_factory, bitcoind):
     l2.stop()
 
     # The close should still be triggered afterwards.
-    with pytest.raises(RpcError, match=r'Channel close negotiation not finished'):
-        l1.rpc.close(chan, False, 0)
+    fut = executor.submit(l1.rpc.close, chan, 0)
```
ZmnSCPxj: `check-python` is failing this:

>      tests/test_closing.py:106:11: F405 'executor' may be undefined, or defined from star imports: fixtures

Looks like need to add `executor` to the arguments of `test_closing_while_disconnected`?

tests/test_closing.py
rustyrussell: Actually, changed the arg to 0, so it's infinite and no unilateral.

ZmnSCPxj: Latest commit-set seems to have skipped this use of `close` entirely?

rustyrussell: No, it eliminates it?

```diff
-    # Now close all channels
-    # All closes occur in parallel, and on Travis,
-    # ALL those lightningd are running on a single core,
-    # so increase the timeout so that this test will pass
-    # when valgrind is enabled.
-    # (close timeout defaults to 30 as of this writing)
-    closes = [executor.submit(l1.rpc.close, p.channel, False, 90) for p in peers]
+    # Now close all channels (not unilaterally!)
+    closes = [executor.submit(l1.rpc.close, p.channel, 0) for p in peers]
```

ZmnSCPxj: Hmm, must have been some weirdness in github then.  Okay.

PR #2924, opened 2019-08-08
Look ma, multiple DB support (Part I of the DB saga)

cdecker:
Data integrity and data safety is paramount for nodes in the Lighning Network
since loss of data, or use of outdated data can directly lead to loss of
funds. c-lightning currently stores all its information in a `sqlite3`
database, and offers the opportunity to synchronously replicate any changes to
other storages through the use of the `db_write` hook.

To further bolster our data resilience, and allow for a variety of more
enterprisey deployments it is desirable to allownode operators to switch out
the `sqlite3` backend for something else. In this PR implement a DB
abstraction framework that allows to configure other DBMSs as backends, and
future PRs will add concrete implementations of those drivers.

This PR consists of a couple of distinct things, but I wanted to make sure it
actually works before submitting a partial solution.

 - Compile time SQL statement extraction, rewriting for each supported DBMS,
   and runtime lookup of the rewritten queries.
 - Creation of a minimal common API that all DBMSs that I looked at can
   support, enabling the abstraction over the specific APIs. The API is very
   much inspired by `sqlite3` but I made sure that it can be easily mapped to
   MySQL and Postgresql. In addition we have number of helper functionality to
   bolt down some avenues for errors:
   - Both bindings and column access use 0-based indexing, not like `sqlite3`s
     1-based bindings (which I found out were inspired by `printf`). This
     caused a couple of misunderstandings in the past.
   - Runtime statement checking such as bind checking on execution (statement
     placeholders without a binding will result in failing the execution).
   - Compile-time instrumentation (placeholders are counted during extraction
     and allow minimizing allocations at runtime).
   - Allow numbered placeholders to avoid repetitive bindings (not implemented
     yet).
   - 
 - Implementation of the `sqlite3` driver, now using the uniform API, and
   keeping all `sqlite3`-specific code in a separate file that is easily
   configured out if the library is missing or incompatible.

Since the "S" in SQL is a marketing lie, and SQL statements are not portable
across DBMSs I had to come up with a method to write portable code. The
solution I went for consists of marking the SQL statements that are to be
extracted with two macros (`SQL` and `NAMED_SQL`), which then allows a tool
(`devtools/sql-rewrite.py`) to extract them and rewrite them into variants
that work for the specific DBMS that we are rewriting for. The queries are
then stored in a large array inside of `wallet/gen_db_${dbms}.c` and looked up
using a unique name (autogenerated in the case of the `SQL` macro).

When preparing a DB statement we look up the rewritten statement in the array
and initialize a `struct db_stmt`, to which we can attach bindings. When
executing we perform a number checks and run the query against the selected
DBMS.

The DB-specific implementation of the common API is in `wallet/db_${dbms}.c`
and adheres to a common set of functions defined in
`wallet/db_common.h`. `wallet/db.c` itself only dispatches calls to the
DB-specific implementations through the use of a `struct db_config` instance
(yes, it is a virtual dispatch table, I implemented dynamic dispatch in a
non-OOP language...), but none of this should ever leak outside of the files
mentioned here, and it'll just look like a normal DB from outside.

Still to do for this PR:
 - [x] Migrate SQL statements in `db.c`.
 - [x] Migrate SQL statements in `wallet.c`.
 - [x] Remove traces of `sqlite3` in the non-sqlite3-specific files.

Once I have these missing pieces I will remove the draft status, so don't feel
forced to review this until I do, but feedback is always welcome :wink:

Still to do for future PRs:
 - Implement MySQL driver
 - Implement Postgresql driver
 - Change hard to rewrite queries into simpler ones (e.g., `INSERT OR REPLACE INTO`)


jb55:
very cool

ZmnSCPxj:
> Since the "S" in SQL is a marketing lie, and SQL statements are not portable
across DBMSs I had to come up with a method to write portable code. The
solution I went for consists of marking the SQL statements that are to be
extracted with two macros (`SQL` and `NAMED_SQL`), which then allows a tool
(`devtools/sql-rewrite.py`) to extract them and rewrite them into variants
that work for the specific DBMS that we are rewriting for.

So we now have our own SQL variant?

Which is not to say it is a bad solution, mostly a gripe.  I encountered this kind of issue before (GLSL shaders), and sometimes this is the only solution you can come up with without going whole hog and creating your own language.

>yes, it is a virtual dispatch table, I implemented dynamic dispatch in a
non-OOP language...

C is an OOP language (for some definition of "is").  It is just that you have to implement OO yourself.  See the shenanigans that gtk gets up to in its C code, for example.

jb55:
> Which is not to say it is a bad solution, mostly a gripe. I encountered this kind of issue before (GLSL shaders), and sometimes this is the only solution you can come up with without going whole hog and creating your own language.

it's an interesting approach, I don't feel that strongly about it yet. the alternative being programming to some ORM which is never much fun. As long as it doesn't require that much hacking to get postgresql working, then replication will be much easier.

wallet/db.c
```diff
@@ -771,6 +782,20 @@ static struct db *db_open(const tal_t *ctx, char *filename)
 	db = tal(ctx, struct db);
 	db->filename = tal_strdup(db, filename);
 	db->sql = sql;
+
+	for (size_t i=0; i<num_configs; i++)
```
niftynei: spacing here is very Go-like >.<

wallet/db_common.h
```diff
@@ -33,13 +34,11 @@ struct db_query {
...
-	   size_t placeholders;
-};
+	size_t placeholders;
 
-struct db_config {
-	const char *name;
-	struct db_query *queries;
-	size_t num_queries;
+	/* Is this a read-only query? If it is there's no need to tell plugins
+	 * about it. */
```
niftynei: nice :nail_care:

wallet/db_sqlite3.c
```diff
@@ -1,11 +1,141 @@
...
+		db_fatal("Failed to begin a transaction: %s", errmsg);
+}
+
+static void db_sqlite3_commit_tx(struct db *db)
+{
+	int err;
+	char *errmsg;
+	err = sqlite3_exec(db->conn, "COMMIT;", NULL, NULL, &errmsg);
+	if (err != SQLITE_OK)
+		fatal("Failed to begin a transaction: %s", errmsg);
```
niftynei: begin -> commit

wallet/db_sqlite3.c
```diff
@@ -1,11 +1,141 @@
...
+			sqlite3_bind_text(s, pos, b->v.text, b->len,
+					  SQLITE_TRANSIENT);
+			break;
+		case DB_BINDING_NULL:
+			sqlite3_bind_null(s, pos);
+			break;
+		}
+	}
+
+	if (err != SQLITE_OK) {
```
niftynei: it's a bit weird semantically to be checking if the 'err' is OK. Consider renaming to `result`?

wallet/db_sqlite3.c
```diff
@@ -1,11 +1,141 @@
...
+static bool db_sqlite3_exec(struct db_stmt *stmt)
+{
+	int err;
+	if (!db_sqlite3_query(stmt)) {
+		/* If the prepare step caused an error we hand it up. */
+		return false;
+	}
+
+	err = sqlite3_step(stmt->inner_stmt);
+	if (err != SQLITE_DONE) {
```
niftynei: same as previous comment.

cdecker: I pretty much copied this from the old code, and there I had gotten it from the sqlite3 example code :-)

cdecker:
Ok, I think this is now good to go. I went ahead and migrated all the places we used the legacy implementation with the new, indirect, implementation and everything seems to be working ok.

I simplified the interface for querying a bit, removed a lot of redundancy and implemented the binding and column accessor functions to use the abstraction layer as well.

I left the migration commits unsquashed for now so they are a bit less daunting, but once the PR has the required ACKs I'll squash them into a single commit.

cdecker:
PING @rustyrussell @niftynei @ZmnSCPxj

jb55:
@cdecker I assume you squash the fixups now that it's ready to review?

cdecker:
> @cdecker I assume you squash the fixups now that it's ready to review?

Keeping the migration fixups separate for until the review is done in order to make the review less daunting, but I'll squash before merging. @bitcoin-bot should be able to infer that the squashed version and the unsquashed version are the same (same changes) and re-apply ACKs automatically :-)

cdecker:
CI failures are caused by `mrkd` not working currently. This is addressed by #3014

wallet/invoices.c
```diff
@@ -329,109 +326,131 @@ bool invoices_find_by_label(struct invoices *invoices,
...
+						       " FROM invoices;"));
+		db_query_prepared(stmt);
 		it->p = stmt;
 	} else
 		stmt = it->p;
 
-	if (db_select_step(invoices->db, stmt))
+
+	if (db_step(stmt))
 		return true;
```
rustyrussell: Is this a memory leak?

wallet/invoices.c
```diff
@@ -539,29 +559,29 @@ void invoices_waitany(const tal_t *ctx,
...
+	if (db_step(stmt)) {
+		invoice.id = db_column_u64(stmt, 0);
 
 		cb(&invoice, cbarg);
-		return;
-	}
-
-	/* None found. */
-	add_invoice_waiter(ctx, &invoices->waiters,
+	}else {
```
rustyrussell: Missing space between `}` and `else`

wallet/invoices.c
cdecker: It is :-)

cdecker:
Most nits are addressed, but I'm currently hunting down a bug introduced by a (clean) rebase on top of `master`, should have a new version up soon.

cdecker:
Ok, fixed the rebase issue I was having problems with yesterday.

Ready for review @rustyrussell :wink:

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/969f4699a26b059ed54b2be3f3780620904902c5

PR #2925, opened 2019-08-08
One step toward wallet locking

darosior:
*[updated 09/27]*
This adds a new `sensitive_command` hook, which allows a plugin to take over any RPC command execution. 
This was at first aimed to only restrict the use of sensitive commands, but was made a general purpose RPC hook as suggested by @ZmnSCPxj (https://github.com/ElementsProject/lightning/pull/2925#discussion_r312381909), it now : 
- allows a plugin (who registered this hook) to return a custom result to a chosen command, hence effectively allowing a plugin to reimplement some RPC
- allows a plugin (who registered this hook) to return a custom error to a chosen command, hence effectively allowing a plugin to restrict the use of some (sensitive?) RPC

Going further, and off topic of the wallet locking, a plugin registering for the `rpc_command` hook could take over the `plugin start` (and cie.) command and extend the plugins possibilities, such as permitting a workaround for hook chaining (#2796).
For example,  the plugin registering for `rpc_command` could register for `htlc_accepted` and override `plugin start`, then ask to every new plugin started with `plugin start` and registering this __same__ hook the permission to accept the HTLC before returning the response to `lightningd`.
*[/updated 09/27]*

### Next step forward
I plan to make a (static) plugin (to be added here) which would allow users to set up a passphrase and to lock / unlock the sensitive RPC calls (~~a little bit like~~ completely like bitcoin-core's `walletpassphrase` mechanism).
I prefered to separate this into multiple PRs.

### Another step forward
If we encrypt `hsm_secret` then even is `lightningd` is stopped somehow (likely killed) it cannot be restarted without the locking plugin. I don't know exactly how to achieve this final step, or if it is even a good scheme ?


lightningd/plugin_hook.c
```diff
@@ -102,13 +121,22 @@ void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
...
 		ph_req->db = ld->wallet->db;
 		hook->serialize_payload(payload, req->stream);
 		jsonrpc_request_end(req);
 		plugin_request_send(hook->plugin, req);
+		if (sync) {
+			ret = plugin_exclusive_loop(hook->plugin);
+			if (ret != ph_req) {
+				void *ret2 = plugin_exclusive_loop(hook->plugin);
+				assert(ret2 == ph_req);
+				io_break(ret);
```
rustyrussell: I... definitely do not understand what this code is doing!

lightningd/plugin_hook.c
```diff
@@ -183,3 +184,41 @@ REGISTER_SYNC_PLUGIN_HOOK(db_write, db_write_hook_callback,
...
+
+/* `sensitive_command` hook */
+
+static void sensitive_command_hook_serialize(struct command *cmd,
+                                             struct json_stream *s)
+{
+	json_object_start(s, "sensitive_command");
+	json_add_string(s, "command", cmd->json_cmd->name);
+	json_object_end(s);
+}
```
rustyrussell: This is, IMHO, way too narrow.  Just create a command hook, and call it.  Have a parameter as to whether it's sensitive maybe (?), but you should definitely include the entire json command parameters.

It should also be able to *alter* the commands, IMHO. This lets a plugin do other things, like logging all commands, or supporting new units (e.g. fiat amounts) or whatever.

lightningd/invoice.c
```diff
@@ -270,7 +270,7 @@ void invoice_try_pay(struct lightningd *ld,
 	tal_add_destructor2(hin, invoice_payload_remove_hin, payload);
 
 	log_debug(ld->log, "Calling hook for invoice '%s'", details->label->s);
-	plugin_hook_call_invoice_payment(ld, payload, payload);
+	plugin_hook_call_invoice_payment(ld->wallet->db, payload, payload);
```
rustyrussell: Changing this parameter seems strange: the two seem unrelated.  We can get at ld in other ways, but handing it explicitly here is the normal pattern, and plugin_hook isn't a db function.

lightningd/plugin_hook.h
```diff
@@ -132,8 +132,10 @@ bool plugin_hook_unregister(struct plugin *plugin, const char *method);
 /* Unregister all hooks a plugin has registered for */
 void plugin_hook_unregister_all(struct plugin *plugin);
 
-/* Special sync plugin hook for db: changes[] are SQL statements, with optional
+/* A sync plugin hook for db: changes[] are SQL statements, with optional
  * final command appended. */
-void plugin_hook_db_sync(struct db *db, const char **changes, const char *final);
+UNNEEDED void plugin_hook_call_db_write(struct db *db,
```
rustyrussell: Is UNNEEDED necessary on this declaration?  Seems odd...

lightningd/invoice.c
darosior: I had to do this to standardize `db_write`. I thought it could have been acceptable since we did not use the `lightningd` struct, only the db.

lightningd/plugin_hook.c
darosior: I took it from the `db_write` hook but the comment made it out to the next commit.

lightningd/plugin_hook.h
darosior: Yes I might not need `UNNEEDED`. I dumbly copied the definition's signature from the macro.

lightningd/plugin_hook.c
ZmnSCPxj: A sufficiently-well-designed hook, in combination with #2796 , would allow the command to be replaced completely, i.e. the hook might allow: `{ "passthrough": true }` to mean no change in the command, `{ "replace-command": { "method" : "foo", "params":{ "arg1" : "arg1" } } }` to replace the command with another command (or just its params), or `{ "return" : { "result" : 42 } }` or `{ "return" : { "error" : { "code" : 42, "message" : "life, the universe, and everything", "data" : "Earth" }}}` to not execute the command but return the specified result/error instead.

lightningd/plugin_hook.h
ZmnSCPxj: > `UNNEEDED` seems unneeded

Recursion depth exceeded.

darosior:
> I don't think command filtering should be done at this level: I prefer a proxy, perhaps offered by a plugin, which can do more serious enforcement.

What more could be done ? *(To get ideas)*

> But the idea of a command hook is fine, and offers some nice capabilities, however it should be plumbed in at the jsonrpc.c parse_request() level, before command dispatch. A security plugin should whitelist, rather than blacklist: this is safe if unknown commands are added (eg. plugins).

I'll amend this to call the hook at `parse_request()` and to include ZmnSCPxj's suggestions.

lightningd/plugin_hook.c
cdecker: The option that @ZmnSCPxj proposes may go a bit too far, but I'd also advocate for the entire payload at least being passed to the hook in order for the plugin to make a more informed decision.

I'd also drop the arbitrary distinction between sensitive and non-sensitive and just call the hook for all JSON-RPC calls that way we don't have two separate control flows (we can add the hook call directly in the JSON-RPC dispatch logic) and the plugin can decide what is sensitive and what's not (with the nice side-effect that we don't have to come up with a distinction that is going to be wrong anyway :wink:).

lightningd/invoice.c
cdecker: I agree with @rustyrussell here, let's keep it `ld` since hooks are not related to the DB at all.

lightningd/gossip_control.c
```diff
@@ -510,13 +511,22 @@ static struct command_result *json_dev_query_scids(struct command *cmd,
...
 	size_t i;
+	struct sensitive_command_hook_payload *sensitive_hook;
 
 	if (!param(cmd, buffer, params,
 		   p_req("id", param_node_id, &id),
 		   p_req("scids", param_array, &scidstok),
 		   NULL))
 		return command_param_failed();
 
+	sensitive_hook = tal(cmd, struct sensitive_command_hook_payload);
```
cdecker: If you call the hook in [parse_request](https://github.com/ElementsProject/lightning/blob/4274b9f0afbb919d004be86943f7675aaa1a26cd/lightningd/jsonrpc.c#L586) and don't distinguish between sensitive and non-sensitive you can avoid all this copy-paste in this commit.

darosior: Yeah, I've almost finished the modifications, indeed this commit was awkward to rebase ^^

lightningd/plugin_hook.c
darosior: > The option that @ZmnSCPxj proposes may go a bit too far, but I'd also advocate for the entire payload at least being passed to the hook in order for the plugin to make a more informed decision.

What I did (not already pushed) is passing the method name and params to the plugin, what do you think ?

> I'd also drop the arbitrary distinction between sensitive and non-sensitive and just call the hook for all JSON-RPC calls that way we don't have two separate control flows (we can add the hook call directly in the JSON-RPC dispatch logic) and the plugin can decide what is sensitive and what's not (with the nice side-effect that we don't have to come up with a distinction that is going to be wrong anyway wink).

That's part of the change for the next push :-)

lightningd/invoice.c
darosior: In that case I cannot call the `db_write` hook without modifying the `db` struct.

darosior:
> One thing that I don't quite get is why this has to be synchronous in the first place. Since the internal dispatch will not be called until we get the result of the hook we can just call the hook asynchronously (we might even be able to call it before we start a new DB transaction, hence removing the split transaction across DB transaction issue). Don't get me wrong, I love the generalization of the sync hook call, but I'm wondering whether we should make use of it here.

To be really honest, I first [expected hooks to be asynchronous](https://lightning.readthedocs.io/PLUGINS.html#hooks) and had a hard time debugging. I then understood that I would (in the "command is restricted" case) call `command_fail(cmd, ...` when there was no `cmd` anymore. I haven't checked how other hooks handle this though.
Why aren't all hooks synchronous by default ?

> If we do it asynchronously, we can also get some of the features @ZmnSCPxj wants: upon receiving a call that we'd like to rewrite, we will just call the rewritten RPC ourselves (us being the plugin) and then directly return the result as the result of the original call. No need to intercept the original call, rewrite, reinject into the dispatch, and then have the dispatch return the result for the rewritten call.

In that case, what would happen to the ongoing request that called us ? Would we return a `command_fail()` but actually making a new call with different parameters after ? I like the synchronous way, because it either success, fail, or be altered (I've implemented this one even if I don't see a case were we would take over an user's decision for now).

cdecker:
> To be really honest, I first expected hooks to be asynchronous and had a hard time debugging. I then understood that I would (in the "command is restricted" case) call command_fail(cmd, ... when there was no cmd anymore. I haven't checked how other hooks handle this though.
> Why aren't all hooks synchronous by default ?

They were, but somewhere along the way they were made asynchronous, which is really strange and contradicts the documentation, but may be a blessing since it actually allows us to make RPC calls which would otherwise not be possible.

I am almost convinced now that only making hooks synchronous (read: `lightningd` makes absolutely no progress while the plugins ruminates over the hook call) if the use-case requires it (`db_write` for example would require the DB transaction to be split between statements and commit, which is really an edgecase).

> In that case, what would happen to the ongoing request that called us ? Would we return a command_fail() but actually making a new call with different parameters after ? I like the synchronous way, because it either success, fail, or be altered (I've implemented this one even if I don't see a case were we would take over an user's decision for now).

The dispatch code in `parse_request` parses the JSON-RPC call, stashes it in a hook call payload, calls the hook and exits. Upon receiving a response from the plugin, the callback gets the stashed call as well as the plugin's response and can perform the actual dispatch to the registered `dispatch` function for that RPC method. While the plugin is thinking about what to do with the JSON-RPC call, the connection just sits idle, no intermediate result is called (hence no `command_fail` or `command_success`), and we just wait for the plugin to make up its mind.

We have a couple of similar situations (basically everywhere a JSON-RPC call results in further queries to other daemons such as `listnodes` below:

https://github.com/ElementsProject/lightning/blob/69b7ef1508eefdf9affaefc20f449679e34d8026/lightningd/gossip_control.c#L258-L274

We start processing the call, then issue the followup query that will eventually return the results (this is equivalent to the hook-call) and signal that the RPC call is pending using the call to `command_still_pending` (this just marks the call as intentionally suspended in order not to upset our sanity checks).

darosior:
> The dispatch code in parse_request parses the JSON-RPC call, stashes it in a hook call payload, calls the hook and exits. Upon receiving a response from the plugin, the callback gets the stashed call as well as the plugin's response and can perform the actual dispatch to the registered dispatch function for that RPC method. While the plugin is thinking about what to do with the JSON-RPC call, the connection just sits idle, no intermediate result is called (hence no command_fail or command_success), and we just wait for the plugin to make up its mind.

Thank you for the detailed explanation, it makes more sense, I'll try it that way. Just to be sure, this means we'd return `command_still_pending` before the dispatch https://github.com/ElementsProject/lightning/blob/f1e84b3d994f9faa099f28a16016649cc1e76aad/lightningd/jsonrpc.c#L648-L652 ?
So we should dispatch inside the hook ?

cdecker:
Added the state::shelved tag to signal that we may be trying another solution, but don't want to close this PR right away :wink:

darosior:
> Added the state::shelved tag to signal that we may be trying another solution, but don't want to close this PR right away wink

I'm still working on this, I've just started something else (almost finished) that I wanted to finish before coming back to this.

cdecker:
> I'm still working on this, I've just started something else (almost finished) that I wanted to finish before coming back to this.

No problem, I'm just tagging these so I don't get the feeling that our PR stack is growing and I can stop checking this one every day :wink:

darosior:
*(messed up my test branch and the PR branch because of the `rework` suffix, still WIP)*

darosior:
Finally got the time to fix this. Updated the OP.
(Sorry @rustyrussell I again use a timer as a workaround for a `command_still_pending`)

lightningd/jsonrpc.c
```diff
@@ -575,14 +578,140 @@ struct json_stream *json_stream_fail(struct command *cmd,
...
+	params = json_get_member(p->buffer, p->request, "params");
+
+	json_object_start(s, "rpc_command");
+	json_add_string(s, "method", tal_fmt(tmpctx, "%.*s",
+	                                     method->end - method->start,
+	                                     p->buffer + method->start));
+	json_add_string(s, "params", tal_fmt(tmpctx, "%.*s",
+	                                     params->end - params->start,
+	                                     p->buffer + params->start));
+	json_object_end(s);
```
cdecker: This is effectively filtering out anything that isn't a `method` or `params`. As we discussed in Berlin it might be advantageous to deviate slightly from the JSON-RPC spec, and allow additional top-level keys in order to authenticate the caller and authorize an RPC call. In order to do this we could add the entire incoming JSON object as the payload for `rpc_command` instead of selectively copying things over.

cdecker: This could easily be achieved by using `json_add_tok` from the previous commit.

lightningd/jsonrpc.c
```diff
@@ -575,14 +578,140 @@ struct json_stream *json_stream_fail(struct command *cmd,
...
+		custom_return = json_get_member(buffer, tok, "result");
+		if (custom_return) {
+			response = json_start(p->cmd);
+			json_add_tok(response, "result", custom_return, buffer);
+			json_object_compat_end(response);
+			return was_pending(command_raw_complete(p->cmd, response));
+		}
+
+		custom_return = json_get_member(buffer, tok, "error");
+		if (custom_return) {
```
cdecker: This is a bit strange since it allows us to return both an `error` as well as a `result`. Maybe we shouldn't perform any validation at all and just return the payload of `return` to the caller?

The only issue I see is the need to rewrite the ID back

lightningd/jsonrpc.c
```diff
@@ -575,14 +578,140 @@ struct json_stream *json_stream_fail(struct command *cmd,
...
+	const jsmntok_t *method, *params;
+
+	method = json_get_member(p->buffer, p->request, "method");
+	params = json_get_member(p->buffer, p->request, "params");
+
+	json_object_start(s, "rpc_command");
+	json_add_string(s, "method", tal_fmt(tmpctx, "%.*s",
+	                                     method->end - method->start,
+	                                     p->buffer + method->start));
+	json_add_string(s, "params", tal_fmt(tmpctx, "%.*s",
```
cdecker: Converting the params to a string, so they need to be unpacked on the plugin side is a bit of a strange choice. Why not use `json_add_tok` to create a direct copy instead?

tests/plugins/rpc_command.py
```diff
@@ -0,0 +1,27 @@
...
+from lightning import Plugin
+
+plugin = Plugin(dynamic=False)
+
+
+@plugin.hook("rpc_command")
+def on_rpc_command(plugin, rpc_command, **kwargs):
+    if rpc_command["method"] == "invoice":
+        # Replace part of this command
+        rpc_command["params"] = json.loads(rpc_command["params"])
```
cdecker: See comment on previous comment regarding serializing the `params` as a string instead of a raw JSON object.

lightningd/jsonrpc.c
darosior: Hmm yes, I was starting to work on a basic RPC locker plugin "à la" bitcoin-core (`lockrpc`, `unlockrpc`, `setrpcpassphrase`) but maybe we then want to have the plugin registering for the `rpc_command` hook to parse an additional field like a cookie auth (once again as in core) ? We could call this plugin `macaroon` !

lightningd/jsonrpc.c
darosior: > This is a bit strange since it allows us to return both an error as well as a result

If there is a `result` field in the response we return ? Line 670

> Maybe we shouldn't perform any validation at all and just return the payload of return to the caller?
> The only issue I see is the need to rewrite the ID back

I think that was to be able to replace the call, as https://github.com/ElementsProject/lightning/pull/2925#discussion_r312381909

lightningd/jsonrpc.c
darosior: > is a bit of a strange choice

Hmmm it is ^^ I'll give `json_add_tok` a shot.

lightningd/jsonrpc.c
```diff
@@ -39,11 +39,14 @@
 #include <lightningd/log.h>
 #include <lightningd/memdump.h>
 #include <lightningd/options.h>
+#include <lightningd/plugin_hook.h>
 #include <stdio.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/un.h>
+#include <wallet/db.h>
```
darosior: This include is weird too..

lightningd/jsonrpc.c
darosior: Ok that's far better using `json_add_tok` ! An excuse is that i wrote this helper after writing the serialization function..

darosior:
Thanks for the review ! Fixed the `method` and `params` serialization.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2925/commits/350b2941289e7084534dbda4618e64474c96d1df

cdecker:
@rustyrussell is your request for changes still valid? If not I'll merge this :wink:

darosior:
Forgot to add documentation for the hook, documented it and rebased on top of master.

By the way @cdecker about the caller authentication, is there any drawback at making the authentication field part of the `params` instead of the TL object ? That way the flow of the locker plugin would be:
- If locked
    - If there is an auth field in the params
        - Check the cookie auth
    - Else
        - Send a custom error
- Else
    - Just response with `continue`

cdecker:
> By the way @cdecker about the caller authentication, is there any drawback in making the authentication field part of the `params` instead of the TL object ?

The downside of that would be that we are mixing call parameters and authentication mechanism, whereas they really are orthogonal. It's certainly doable, by having the client inject authentication parameters and the auth plugin filtering them out again, but it feels really awkward.

darosior:
Ok, I think I'll go for a violation of the JSONRPC specs then :D

rustyrussell:
> Ok, I think I'll go for a violation of the JSONRPC specs then :D

No, clearly you just wrap the user-supplied params inside a sub-object?

i.e 
```json
"params": { "params": ...user params here..., "auth" ... }
```

lightningd/json.c
```diff
@@ -598,3 +598,45 @@ struct command_result *param_bitcoin_address(struct command *cmd,
...
+		json_array_end(result);
+		break;
+
+	case JSMN_OBJECT:
+		json_object_start(result, fieldname);
+		json_for_each_obj(i, t, tok)
+			json_add_tok(result, json_strdup(tmpctx, buffer, t), t, buffer);
+		json_object_end(result);
+		break;
+
```
rustyrussell: OK, since tok->type should always be one of these types, change these breaks to returns, except the last JSMN_UNDEFINED.  Then abort() at the bottom of the function.

lightningd/jsonrpc.c
rustyrussell: Yes, copy entire thing please, rather than cherrypicking.  This allows users to get creative.  I may regret that, of course!

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+Plugin: A new plugin hook, `rpc_command` allows a plugin to take over `lightningd` for any RPC command.
```
rustyrussell: Hey, can we use the new commit line, in anticipation of Christian's Changelog tool being merged?

i.e. in commit msg, put
Changelog-added: Plugin: A new plugin hook, `rpc_command` allows a plugin to take over `lightningd` for any RPC command.

rustyrussell:
> Ack [2d7c4d6](https://github.com/ElementsProject/lightning/commit/2d7c4d660d52dbf6b683ca31ab58db4d1ec15db2)
> 
> This is really neat! I'll work on getting rid of that annoying timer ;)

```patch
diff --git a/lightningd/jsonrpc.c b/lightningd/jsonrpc.c
index 2562d7627..a68ffde2c 100644
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -775,11 +775,12 @@ parse_request(struct json_connection *jcon, const jsmntok_t tok[])
        /* Duplicate since we might outlive the connection */
        rpc_hook->buffer = tal_dup_arr(rpc_hook, char, jcon->buffer, tal_count(jcon->buffer), 0);
        rpc_hook->request = tal_dup_arr(rpc_hook, const jsmntok_t, tok, tal_count(tok), 0);
-       /* Prevent a race between was_pending and still_pending */
-       new_reltimer(c->ld->timers, rpc_hook, time_from_msec(1),
-                    call_rpc_command_hook, rpc_hook);
+       call_rpc_command_hook(rpc_hook);
 
-       return command_still_pending(c);
+       /* Could have already completed the request, or it could be pending.
+        * Hard to plumb through the layers of callbacks, and caller doesn't
+        * care. */
+       return NULL;
 }
 
 /* Mutual recursion */
```

CHANGELOG.md
darosior: Will try !

lightningd/json.c
darosior: Ok

darosior: So, a `JSMN_UNDEFINED` tok basically means "bad json" ?

darosior:
Thanks for the patch !

Replaced the cherry-picking of fields by a `json_add_tok()`.
Made @bitcoin-bot /changelog happy :-)

darosior:
Ok so I removed your patch because otherwise valgrind tests would not pass if there is no plugin registered for the `rpc_command` hook.

Hope the timer isn't just hiding the stuff to both me and valgrind ?...

doc/PLUGINS.md
```diff
@@ -683,6 +683,69 @@ processed before the HTLC was forwarded, failed, or resolved, then the plugin
...
+    }
+}
+```
+
+which can in turn:
+
+Let `lightningd` execute the command with
+
+```json
+{"continue"}
```
cdecker: Pretty sure this isn't valid JSON

darosior: Will change to result: "continue"..

darosior: EDIT: "continue": true seems more natural

darosior:
Changed the bad json in the "continue" case. Here is the patch diff
```diff
diff --git a/lightningd/jsonrpc.c b/lightningd/jsonrpc.c
index 4e534d242..82c445f3e 100644
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -623,16 +623,23 @@ static void
 rpc_command_hook_callback(struct rpc_command_hook_payload *p,
                           const char *buffer, const jsmntok_t *resulttok)
 {
-       const jsmntok_t *tok, *method, *params, *custom_return;
+       const jsmntok_t *tok, *method, *params, *custom_return, *tok_continue;
        struct json_stream *response;
+       bool exec;
 
        params = json_get_member(p->buffer, p->request, "params");
 
        /* If no plugin registered, just continue command execution. Same if
         * the registered plugin tells us to do so. */
-       if (buffer == NULL || json_tok_streq(buffer, resulttok, "continue"))
-               return was_pending(command_exec(p->cmd->jcon, p->cmd, p->buffer,
+       if (buffer == NULL)
+           return was_pending(command_exec(p->cmd->jcon, p->cmd, p->buffer,
                                                p->request, params));
+       else {
+               tok_continue = json_get_member(buffer, resulttok, "continue");
+               if (tok_continue && json_to_bool(buffer, tok_continue, &exec) && exec)
+                       return was_pending(command_exec(p->cmd->jcon, p->cmd, p->buffer,
+                                                       p->request, params));
+       }
 
        /* If the registered plugin did not respond with continue,
         * it wants either to replace the request... */
```

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2925/commits/640f9498468490d6827a767b3023db26be138339

PR #2927, opened 2019-08-09
Add manpages to readthedoc

### LINKED ISSUE #2918: Integrate the man-pages into the generated docs webpage
We have quite a wealth of man-pages that we make extensive use when calling `lightning-cli help [command]` on the command-line, but it'd be nice to serve that info also in the [generated doc](https://lightning.readthedocs.org).

I could not find a way to convert our current `asciidoc` format to a format understood by readthedocs / sphinx, which internally uses sphinx, but I found a way to generate man-pages from restructuredtext, so maybe we should just port the sources of the man-pages over?

rst seems like a more supported format (and one whose compiler doesn't attempt to drag in the entire latex distribution on CI).

darosior:
As discussed in https://github.com/ElementsProject/lightning/pull/2899#issuecomment-518252597, closes #2918.

This adds a script to the tools to generate a markdown file understood by readthedoc by manpage, then a new section to readthedoc: "Manpages", then a page under that section by markdown-manpage. This is at the expense of having `pandoc` as a dependency for this script.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -22,3 +22,44 @@ c-lightning Documentation
    Coding Style Guideline <STYLE.md>
    MAKING-RELEASES.md
    CHANGELOG.md
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Manpages
+
+   doc/lightning-autocleaninvoice.7.md
+   doc/lightning-check.7.md
+   doc/lightning-cli.1.md
+   doc/lightning-close.7.md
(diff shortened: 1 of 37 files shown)


rustyrussell:
This definitely needs to be a 'make' target, invoked by `make doc-all`.  I'll take a shot at converting it...

darosior:
Closed in favor of #2928.

PR #2928, opened 2019-08-09
Readthedocs via makefile

rustyrussell:
I stole @darosior #2927 and Makefilized it.  Now 'make doc-all' Does The Right Thing.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -22,3 +22,44 @@ c-lightning Documentation
    Coding Style Guideline <STYLE.md>
    MAKING-RELEASES.md
    CHANGELOG.md
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Manpages
+
+   doc/lightning-cli.1.md
+   doc/lightningd.8.md
+   doc/lightningd-config.5.md
+   doc/lightning-autocleaninvoice.7.md
+   doc/lightning-check.7.md
+   doc/lightning-close.7.md
+   doc/lightning-connect.7.md
+   doc/lightning-decodepay.7.md
+   doc/lightning-delexpiredinvoice.7.md
+   doc/lightning-delinvoice.7.md
+   doc/lightning-disconnect.7.md
+   doc/lightning-fundchannel.7.md
+   doc/lightning-fundchannel_start.7.md
+   doc/lightning-fundchannel_complete.7.md
+   doc/lightning-fundchannel_cancel.7.md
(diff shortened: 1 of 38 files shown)


doc/index.rst
```diff
@@ -22,3 +22,44 @@ c-lightning Documentation
    Coding Style Guideline <STYLE.md>
    MAKING-RELEASES.md
    CHANGELOG.md
+
+.. toctree::
+   :maxdepth: 1
+   :caption: Manpages
+
+   doc/lightning-cli.1.md
```
cdecker: These paths are relative to the `index.rst` file, therefore should not have the `doc/` path.

ZmnSCPxj:
I would mildly suggest instead:

* Find some tool to convert Markdown to manpage.
* Hand-convert (or auto-convert but with a sentient intelligence postprocessing pass) our current a2x files to Markdown.

Also suggest doing this after 0.7.2 instead.

darosior:
> I would mildly suggest instead:
> 
> * Find some tool to convert Markdown to manpage.
> * Hand-convert (or auto-convert but with a sentient intelligence postprocessing pass) our current a2x files to Markdown.
> 
> Also suggest doing this after 0.7.2 instead.

https://pypi.org/project/mrkd/ seems able to do that. I'll try (if we end up doing it this way) it since I think it's better to use pip dep that `pandoc` (which seems able to do that, too).

jb55:
> it's better to use pip dep than pandoc 

why's that? I've always had the best experience using pandoc for doc conversions.

darosior:
> > it's better to use pip dep than pandoc
> 
> why's that? I've always had the best experience using pandoc for doc conversions.

Less dependencies ?

cdecker:
> > it's better to use pip dep than pandoc
> 
> why's that? I've always had the best experience using pandoc for doc conversions.

Sorry, that's easy to mistake: I meant we shouldn't use Ubuntu packages to install python dependencies, but rather have a top-level `requirements.txt` file listing python dependencies and then using `pip` to install them into the python environment. That way we are independent of the OS package manager and can use more up to date dependencies.

rustyrussell:
Closed in favor of #2936

ZmnSCPxj:
Closing as per https://github.com/ElementsProject/lightning/pull/2928#issuecomment-523284962 it seems Rusty clicked wrong button with human fingers.

Issue #2831, opened 2019-07-23
`fundchannel_complete` crashes on peer_error?


niftynei:
### Issue and Steps to Reproduce

Crash when calling `fundchannel_complete`.

```
+59249.972233100 lightningd(25401):DEBUG: lightning_openingd-02...4a chan #40: peer_in WIRE_ERROR
+59249.972659597 lightningd(25401):INFO: 02...4a chan #40: channel 5c...f8: funding failed due to internal error                           
+59250.101029186 lightningd(25401):BROKEN: FATAL SIGNAL 11 (version v0.7.1-64-gbb30104)
+59250.101053244 lightningd(25401):BROKEN: backtrace: common/daemon.c:45 (send_backtrace) 0x56450653b7ae
+59250.101067785 lightningd(25401):BROKEN: backtrace: common/daemon.c:53 (crashdump) 0x56450653b7fe
+59250.101077871 lightningd(25401):BROKEN: backtrace: (null):0 ((null)) 0x7fb50c7bef1f
+59250.101090137 lightningd(25401):BROKEN: backtrace: lightningd/jsonrpc.c:539 (json_stream_raw_for_cmd) 0x564506513b46 
+59250.101102902 lightningd(25401):BROKEN: backtrace: lightningd/jsonrpc.c:561 (json_start) 0x564506513c76
+59250.101115643 lightningd(25401):BROKEN: backtrace: lightningd/jsonrpc.c:580 (json_stream_fail_nodata) 0x564506513d2c 
+59250.101127982 lightningd(25401):BROKEN: backtrace: lightningd/jsonrpc.c:498 (command_fail) 0x5645065139c8
+59250.101137717 lightningd(25401):BROKEN: backtrace: lightningd/opening_control.c:103 (uncommitted_channel_disconnect) 0x56450651b110
+59250.101149916 lightningd(25401):BROKEN: backtrace: lightningd/opening_control.c:719 (opening_channel_errmsg) 0x56450651cda1
+59250.101160025 lightningd(25401):BROKEN: backtrace: lightningd/subd.c:383 (handle_peer_error) 0x564506534618
+59250.101171846 lightningd(25401):BROKEN: backtrace: lightningd/subd.c:464 (sd_msg_read) 0x564506534939
+59250.101183525 lightningd(25401):BROKEN: backtrace: lightningd/subd.c:303 (read_fds) 0x56450653434e
[...]

PR #2931, opened 2019-08-09
lightningd: fix crash when peer disconnects after fundchannel_start, before cancel/complete

rustyrussell:
Fixes: #2831


darosior:
I made one here https://github.com/rustyrussell/lightning/pull/6 but was not quick enough

PR #2935, opened 2019-08-09
rpc: `lightningd` can call plugin's rpcmethods internally.

trueptolemy:
This PR tries to implement `internal` rpcmethod interface. And the `internal` rpcmethod means the `rpcmethod` supplied by plugins can be called by `lightningd`. It's based on @cdecker and @ZmnSCPxj  's [idea](https://github.com/ElementsProject/lightning/pull/2851#issuecomment-515675462).

This time, my idea is like this:
- First, `lightningd` can register kinds of topics, and it will specify json format and parameters in document. During registration, we need set `serialize` interface.

- Then, plugins can supply internal `rpcmethod` by set `internal` field as true in the response of `getmanifest`. These special `rpcmethod` must use the given `name` rigistered by `lightningd`.

- When `lightningd` want to call plugin's rpcmethod, it need to set `response_callback` interface. Setting callback when calling is more flexible than specifying callback when registeration.

In addition, this PR present `getclientversion` internal rpcmethod interface as the example.
- If `getclientversion` is called before `lightningd` has sent all plugins `init`, a timer will be set and after a littel time(eg, 1 second) `getclientversion` will be retried.

- [ ] Unfinished: related document. This PR has only one internal rpcmethod: `getclientversion` just for test, it's useless so far... (If this idea is ok, I'll add the doc.)

Welcome feedback!


trueptolemy:
Add more comment.

lightningd/jsonrpc.c
```diff
@@ -633,22 +700,27 @@ parse_request(struct json_connection *jcon, const jsmntok_t tok[])
...
-	res = c->json_cmd->dispatch(c, jcon->buffer, tok, params);
-	db_commit_transaction(jcon->ld->wallet->db);
+	/* Especially for internal call, if we set a timer, and the timer loop
+	 * of `lightningd` will make the timer callback in db transaction. */
+	if (!db_already_in_transaction(ld->wallet->db)) {
+		db_begin_transaction(ld->wallet->db);
+		res = c->json_cmd->dispatch(c, buffer, tok, params);
+		db_commit_transaction(ld->wallet->db);
+	} else
+		res = c->json_cmd->dispatch(c, buffer, tok, params);
```
rustyrussell: I don't think this can work, at least in this form.

Nesting transactions is always a bad idea.  Various callbacks we have assume that the transaction will be successfully completed before anything else happens: e.g. we'll commit to disk, then send something to a subdaemon.  This implicit handling keeps our data consistency model simple (and, frankly, managable!).

I'm still reviewing the patches, and am having trouble seeing the forest through the trees so far...

trueptolemy:
> OK, I think I see what you're doing: implementing call-outs. I think though that this needs to be done lower-level, at the plugin level, not the jsonrpc level.

Do you mean the better way is to choose another way to directly dispatch rpcmethod (called by `lightningd` itselt) to plugins or use some way like `hook`?

> even if the plugin thinks of it as a magic internal function (using `__name` or `lightningd.name` would probably work as well as an explicit flag, TBH).

My wish is to make `lightningd` can call plugin rpcmethods. This call is from `lightningd`, not users or plugins.
The origin idea is to implement a way that `lightningd` can get external information through plugins, like `lightningd` asks plugins to supply the block info from any kind of bitcoin clients(the bitcoin plugin need to implement this). 

We have `hook`, but I think the `hook` is used to supply command/simple information, and its synchronous limits its usage, like discussed [here](https://github.com/ElementsProject/lightning/pull/2851#issuecomment-515675462). And another aternative way is to make `lightningd` could call plugin rpcmethod...

SimonVrouwe:
> My wish is to make lightningd can call plugin rpcmethods. 

@trueptolemy Have not read the code, but what exactly do you mean with above?

I am probably misreading, but it sound as if you want to make `lightningd` code depend on rpc methods that are added by _external_ plugins. Wouldn't that break the design principle of the plugin-system as an _add-on_ system, i.e. where features/function are added without modifying CL's code base?

ZmnSCPxj:
The desire as I understand it is to allow one or more plugins to handle some internal queries.

For example, consider the query "what is the block at height 42?"

In current C-Lightning, this is handled by internal `lightningd/bitcoind.c`.  This uses `bitcoin-cli` to get the block data.

But in principle this can be done differently by e.g. using an Electrum backend.

The desire is to be able to swap the blockchain backend for others.

Since we already have a mechanism by which code can be selectively added to C-Lightning (plugins) it make some sense to reuse the mechanism.

Hooks might work, actually.  Only `db_write` is synchronous, and most of the queries you seem to imagine needing this would be asynchronous (querying block data, broadcasting txes...), and every other hook is asynchronous.

trueptolemy:
Thank you @ZmnSCPxj , your words are clear.

> Wouldn't that break the design principle of the plugin-system as an add-on system, i.e. where features/function are added without modifying CL's code base?

@SimonVrouwe The normal rpcmethod call will create a `struct json_connection`, which includes the `struct io_conn *conn` with the caller. And based on this structure, we dispatch the command(in `parse_request()`) to plugins or other part to handle and we will send the response to the caller through `io_conn`.

The internal rpcmethod is used for `lightningd` to call plugin rpcmethods. For internal call, I use a struct `struct internal_json_connection` to replace `struct json_connection`. And this structure doesn't have the connection with caller, but it has `response_cb` member. The internal calls will also be dispatched (in `parse_request()`) like the normal rpcmethod, but they will be dispatched only to plugins. When we receive the response from plugins, `response_cb` will be called.

Now I only make only one internal rpcmethod topic: `getclientversion`, and we can continue to implement `gettxoutbyscid`, `getblockbyheight`, `sendrawtx`, `estimatefee` and `getclientsync`. If the plugin supports these rpcmethods and mark these as internal, this plugin can theoretically replace the `bitcoin-cli` functions of `lightningd`(of course, may need other changes, but these are core parts).

In fact , like @ZmnSCPxj  said, `hook` can also achive the same goal. And he's right, synchronous is not the problem.
The another view I conserned is `hook`s always tell `lightningd` a simple command or simply information, like `continue`, `disconnect` or `reject`. And now we need plugins to tell `lightningd` about the data of `block` or `transaction`, even we need plugins to send transactions. In this point, these are all ok if we accept `hook` can do these things.

The internal rpcmethods need given topics, and the calling process and the registration I implement is a bit similar to the `hook`. If it need to be "at the plugin level, not the jsonrpc level", for my understanding, we may need use `hook` directly, or directly dispatch to plugins not use `parse_request()`? @rustyrussell

lightningd/jsonrpc.c
trueptolemy: `getclientversion` will be called before rpc initials and I want to use timer to retry `getclientversion`, but `io_loop_with_timers` always use the transaction which will cause the conflict with here.
I think the alternative way is to implement the waiter like `struct sync_waiter`. What about implementing the waiter to pending all request?

rustyrussell:
Yes, I think modifying the hook infra is the Right Way.  Of course, as always, devil is in the details and it might not seem so ideal once implemented!

Hooks are currently simple, but that's simple an artifact of how they're used, not fundamental.

To be clear: we've long wanted this ability (esp. to move bitcoind querying out to a plugin): @cdecker has talked about implementing it in the past, in fact, but never got sufficient Round Tuits.

trueptolemy:
@rustyrussell  Thank you very much for your clear explanation :)

> Yes, I think modifying the hook infra is the Right Way. Of course, as always, devil is in the details and it might not seem so ideal once implemented!

> Hooks are currently simple, but that's simple an artifact of how they're used, not fundamental.

In the future, if I have time I'll continue to think about it.

Issue #2923, opened 2019-08-08
htlc hook not working on startup


renepickhardt:
it seems like the `@plugin.hook("htlc_accepted")` hook is not working as excpected when starting a node which has an unresolved htlc. 

from the logfile I copy
```
2019-08-08T11:02:01.889Z plugin-htlchook.py on_htlc_accepted called
2019-08-08T11:02:01.889Z lightningd(12424): Plugin for htlc_accepted returned non-result response {"jsonrpc": "2.0", "id": 3, "error": "Error while processing htlc_accepted: AttributeError(\"'NoneType' object has no attribute 'getinfo'\",)"}
```

this error is invoked because I called `plugin.rpc.getinf()` it seems like on startup `plugin.rpc` is still `None` but the hook is already being invoked. 

### Issue and Steps to Reproduce
create a payment channel between two nodes. Initiate the payment process and let the other node crash during the payment process. then restart the node and experience this behavior. 

### `getinfo` output
``` 
ln1 getinfo
{
   "id" : "03d1bba2403b9901096f565cec8f274805adfba9fcf80ea693360d54cf711b7bb6",
   "alias" : "alice",
   "color" : "03d1bb",
   "num_peers" : 1,
   "num_pending_channels" : 0,
   "num_active_channels" : 0,
   "num_inactive_channels" : 1,
   "address" : [],
   "binding" : [
      {
         "type" : "ipv4",
         "address" : "127.0.0.1",
         "port" : 1337
      }
   ],
   "version" : "v0.7.1",
   "blockheight" : 649,
   "network" : "regtest",
   "msatoshi_fees_collected" : 0,
   "fees_collected_msat" : "0msat"
}
```

rustyrussell:
Yuck.  @cdecker changed htlcs to be retried, rather than failed, on startup.  But plugins aren't ready yet.

Completing the HTLCs later is the obvious solution, but I'm not sure what the effect of having "half-finished" HTLCs lying around in the interim would be (which is *why* we failed on startup!).

Assigning to @cdecker for this thoughts.

cdecker:
I'm surprised calling any RPC call while handling a hook is actually working at all. Remember that the hooks are supposed to be synchronous and no other action is allowed to make any progress, including RPC calls. This is the main motivation when I felt ok moving the call to `htlc_accepted` hook calls before `init` calls in the first place (you're not supposed to interact with the RPC when handling a hook, though I could have made that more clear in the documentation).

[...]

PR #2938, opened 2019-08-10
Call htlc accepted on startup after plugin init

rustyrussell:
Fixes: #2923

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -632,6 +632,7 @@ int main(int argc, char *argv[])
 	int stop_fd;
 	struct timers *timers;
 	const char *stop_response;
+	struct htlc_in_map *unprocessed_htlcs;
 
 	/*~ What happens in strange locales should stay there. */
 	setup_locale();
@@ -769,7 +770,7 @@ int main(int argc, char *argv[])
 	 *  topology is initialized since some decisions rely on being able to
 	 *  know the blockheight. */
 	db_begin_transaction(ld->wallet->db);
-	load_channels_from_wallet(ld);
+	unprocessed_htlcs = load_channels_from_wallet(ld);
 	db_commit_transaction(ld->wallet->db);
 
 	/*~ Create RPC socket: now lightning-cli can send us JSON RPC commands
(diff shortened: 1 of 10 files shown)


cdecker:
While postponing the replay of HTLCs seems to address this, it leaves HTLCs temporarily in a weird state. I'd propose adding a new state for the HTLC (`hold`) and deferring the resolution until we the plugin has access to the RPC again (see https://github.com/ElementsProject/lightning/issues/2923#issuecomment-520164117 for rationale).

cdecker:
I'll revisit this, adding the `hold` state, after the 0.7.2 release.

rustyrussell:
Actually, should we just go back to failing these HTLCs on restart?

AFAICT we will anyway, since no peers can be connected at this point?

cdecker:
> Actually, should we just go back to failing these HTLCs on restart?
> 
> AFAICT we will anyway, since no peers can be connected at this point?

No,  we can't since an incomplete hook call may have triggered an external event (if we are the recipient me might have marked a shopping cart as paid), and the triggering a failure would mean we end up in an inconsistent state (shopping cart will be shipped, but payment was returned).

The same scenario may happen if the hook short-circuits the payment or switches transport (cross-chain atomic swaps).

cdecker:
I think your code does exactly what we'd like to do, sans the explicit `HOLD` state, i.e., defer resolution until startup is complete. I'd still like to add the `HOLD` state anyway, just to make it more explicit.

(I'm no longer sure we'd need the `HOLD` state for cross-chain atomic swaps since hooks are no longer always synchronous btw)

ZmnSCPxj:
> (I'm no longer sure we'd need the `HOLD` state for cross-chain atomic swaps since hooks are no longer always synchronous btw)

Does that mean it would be unnecessary for this `HOLD` state also if we want to implement JIT-Routing?

@renepickhardt for interest.

cdecker:
> > (I'm no longer sure we'd need the `HOLD` state for cross-chain atomic swaps since hooks are no longer always synchronous btw)
> 
> Does that mean it would be unnecessary for this `HOLD` state also if we want to implement JIT-Routing?
> 
> @renepickhardt for interest.

We would need to try, but it seems we could do without. Since the `htlc_accepted` hook is not called synchronously, we can do whatever intbetween getting the call and resolving it. The non-final state basically becomes an alias for `HOLD` so I think we can do without.

cdecker:
Rebased, and I'm happy with the change as is :wink: 

Sorry for the delay @rustyrussell, you were right after all :-)

ACK https://github.com/ElementsProject/lightning/pull/2938/commits/10e3c7f210925e8afdf453f0ab5dc7099e8424c5

PR #2940, opened 2019-08-10
pytest: fix flaky "peer reconnected" message in test_option_upfront_s…

rustyrussell:
…hutdown_script

If l2 doesn't think we're onchain yet, it treats the new connection from l1
as a reconnection, triggering 'ValueError: 1 nodes had unexpected reconnections'


ZmnSCPxj:
ACK 7abafa9

cdecker:
Nice find, I was hitting this but couldn't figure out the solution.

PR #2944, opened 2019-08-11
Plugin: Use `autodata` to register notifications

trueptolemy:
I remember @rustyrussell  [said](https://github.com/ElementsProject/lightning/pull/2799#discussion_r308490161),  we should avoid the extra line change in `const char *notification_topics[]` when we plan to add new notifications(in fact, I want to add `sendpay_success` and `sendpay_failure` notifications).

I guess using `autodata` to register notifications is an available way and I rewrite notification interfaces.

### DIFF
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -5,7 +5,6 @@
 #include <lightningd/peer_htlcs.h>
 
 const char *notification_topics[] = {
-	"connect",
 	"disconnect",
 	"warning",
 	"invoice_payment",
@@ -39,13 +38,27 @@ bool notifications_have_topic(const char *topic)
 	return false;
 }
 
+static void connect_notification_serialize(struct json_stream *stream,
+					   struct node_id *nodeid,
+					   struct wireaddr_internal *addr)
+{
+	json_add_node_id(stream, "id", nodeid);
+	json_add_address_internal(stream, "address", addr);
+}
+
+REGISTER_NOTIFICATION(connect,
+		      connect_notification_serialize);
+
 void notify_connect(struct lightningd *ld, struct node_id *nodeid,
 		    struct wireaddr_internal *addr)
 {
-	struct jsonrpc_notification *n =
-	    jsonrpc_notification_start(NULL, "connect");
-	json_add_node_id(n->stream, "id", nodeid);
-	json_add_address_internal(n->stream, "address", addr);
+	void (*serialize)(struct json_stream *,
+			  struct node_id *,
+			  struct wireaddr_internal *) = connect_notification_gen.serialize;
+
+	struct jsonrpc_notification *n
+		= jsonrpc_notification_start(NULL, connect_notification_gen.topic);
+	serialize(n->stream, nodeid, addr);
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
 }
@@ -149,7 +162,3 @@ void notify_forward_event(struct lightningd *ld,
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
 }
-
-/* TODO: It's a dummy notification. Will be removed when we have a 'real' one
- * in this file. */
-REGISTER_JSON_INTERNAL_COMMAND(hello_notification, NULL, void *);
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -4,10 +4,6 @@
 #include <lightningd/notification.h>
 #include <lightningd/peer_htlcs.h>
 
-const char *notification_topics[] = {
-	"forward_event"
-};
-
 static struct notification *find_notification_by_topic(const char* topic)
 {
 	static struct notification **notilist = NULL;
@@ -27,10 +23,6 @@ bool notifications_have_topic(const char *topic)
 	if (noti)
 		return true;
 
-	/* TODO: Remove this block after making all notifications registered. */
-	for (size_t i=0; i<ARRAY_SIZE(notification_topics); i++)
-		if (streq(topic, notification_topics[i]))
-			return true;
 	return false;
 }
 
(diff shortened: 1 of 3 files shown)


lightningd/notification.c
```diff
@@ -13,8 +13,26 @@ const char *notification_topics[] = {
...
+	return NULL;
+}
+
 bool notifications_have_topic(const char *topic)
 {
+	struct notification *noti = find_notification_by_topic(topic);
+	if (noti)
+		return true;
+
+	/* Remove this block after making all notifications registered. */
```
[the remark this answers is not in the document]
trueptolemy: I'm sorry about my unclear comment.

trueptolemy:
Correct the comment and rebased to resolve the conflict.

lightningd/notification.c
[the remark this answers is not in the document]
cdecker: No problem, just thought I could point it out quickly :-)

lightningd/notification.c
```diff
@@ -131,3 +149,19 @@ void notify_forward_event(struct lightningd *ld,
 	jsonrpc_notification_end(n);
 	plugins_notify(ld->plugins, take(n));
 }
+
+void notification_call(struct lightningd *ld, const char* topic,
+		       void *payload)
+{
+	struct notification *noti = find_notification_by_topic(topic);
+	assert(noti);
```
cdecker: Nit: It's likely better to changes this into `if (noti == NULL) { fatal("Could not find notification topic %s", topic); }` just to make it more obvious later.

lightningd/notification.h
```diff
@@ -37,4 +38,24 @@ void notify_forward_event(struct lightningd *ld,
...
 
+struct notification {
+	const char *topic;
+	void (*serialize_payload)(void *src, struct json_stream *dest);
+};
+
+AUTODATA_TYPE(notifications, struct notification);
+
+void notification_call(struct lightningd *ld, const char* topic,
+		       void *payload);
```
cdecker: We can likely also generate a type-safe wrapper for each notification that wraps the association `(topic, payload)` and internally calls `notification_call`.

I'd suggest making this `notification_##topic##_call` or `notify_##topic` and then we can just pass in the payload, instead of having to match topic and payload type.

lightningd/peer_control.c
```diff
@@ -894,7 +895,10 @@ peer_connected_hook_cb(struct peer_connected_hook_payload *payload,
 		abort();
 	}
 
-	notify_connect(ld, &peer->id, &addr);
+	noti_payload = tal(tmpctx, struct connect_notification_payload);
```
cdecker: Not sure this needs to be allocated on the heap at all, if we immediately call the notification-serialization and pass that to the subscribers (if there are any), we retain control of the struct and can leave it on the stack.

cdecker: Would save us one allocation :-)

lightningd/notification.c
trueptolemy: Thank you, it's better!

lightningd/notification.h
trueptolemy: Like the `hook`. Ok, I'll do it :-)

lightningd/peer_control.c
trueptolemy: Thank you! I'll try to take all struct fields as arguments for function calls. :-)

trueptolemy:
@cdecker 
If "take all struct fields as arguments for function calls", it seems difficult to generate the unified format for `notification_##topic##_call` or `notify_##topic`, because we need use different parameters to replace a `payload_type` parameter.
From this point, it may be necessary to declare each `notification_##topic##_call` separately here.

lightningd/notification.c
trueptolemy: Fixed.

trueptolemy:
@cdecker 
Now I declare each notification call functions separately (doesn't choose to generate a wrapper for internally calls notification_call), and take all struct fields as arguments for function calls

cdecker:
I see the problem. How about creating a struct for each notification that holds the payload of the notification itself?

The struct for a forward would look like this:

```c
struct forward_event {
    const struct htlc_in *in,
    const struct htlc_out *out,
    enum forward_status state,
    enum onion_type failcode,
    struct timeabs *resolved_time
};
```

We'd then just declare a function that knows how to serialize a single instance to the json-rpc notification:

```c
void forward_event_serialize(struct jsonrpc_notification *n, struct forward_event *e) {...}
```

And then in the registration we ensure that the serialize's function signature matches the event type we declared:

```c 
#define REGISTER_NOTIFICATION(topic, payload_type, serialize)                  \
	struct notification topic##_notification_gen = {                       \
	    stringify(topic),                                                  \
	    typesafe_cb_cast(                                                  \
		void (*)(void *, struct jsonrpc_notification *),               \
		void (*)(struct node_id *, struct jsonrpc_notification *),     \
		serialize),                                                    \
	};                                                                     \
	AUTODATA(notifications, &topic##_notification_gen);
```

(this is free-hand code, and I'm not sure it works as is)

And then we can also autogenerate the dispatch. The whole thing then looks a bit like this:

```c
struct notification_type {
	const char *topic;
	/* the serialization interface */
	void *serialize;
};

/* This actually dispatches the notification, using the configured settings from type */
void notify_(struct lightningd *ld, struct notification_type *type, void *event);

AUTODATA_TYPE(notifications, struct notification);

#define REGISTER_NOTIFICATION(topic, payload_type, serialize)                  \
	struct notification topic##_notification_gen = {                       \
	    stringify(topic),                                                  \
	    typesafe_cb_cast(                                                  \
		void (*)(void *, struct jsonrpc_notification *),               \
		void (*)(struct node_id *, struct jsonrpc_notification *),     \
		serialize),                                                    \
	};                                                                     \
	void notify_##topic(struct lightningd *ld, payload_type payload)       \
	{                                                                      \
		notify(ld, &topic##_notification_gen, payload);                \
	}                                                                      \
	AUTODATA(notifications, &topic##_notification_gen);
```

`notify_` then just needs to call `type.serialize(stream, payload)` to get the serialized payload, and can then issue the notification.

The use of `typesafe_cb_cast` ensures the serialization matches the payload type, and we save the manual casting, converting and dispatching for each event type.

What do you think?

trueptolemy:
@cdecker  Thank you so much!!  :-) 

Your idea is like the original version(the original code for this PR is here [37e249](https://github.com/trueptolemy/lightning/commit/37e249149ca3be4a78d3fb248a0a26682306148c), I backuped before change :-) ), and generates a wrapper (`notification_##topic##_call`) for each notification based on internally calls notification_call.

The original version implemented warpper, but it can't well implement "take all struct fields as arguments for function calls", and it also need to [allocate the payload](https://github.com/ElementsProject/lightning/pull/2944#discussion_r318007770) structure every time before notification call.

But concerned the [comment](https://github.com/ElementsProject/lightning/pull/2944#pullrequestreview-280097469), I present a new version now. This new version takes all struct fields as arguments, but can't well warp (because it discarded the payload structure wapper).

This is a contradiction. In my understanding, we need to choose between "takes all struct fields as arguments for notification call" and "well warp, like payload wapper". Do you prefer the old version or the new version?

Because of the backup, it is very convenient to switch between two versions for this PR.

cdecker:
@trueptolemy you're right, we are contradicting ourselves there a bit (I personally prefer the wrap-everything-in-structs approach, while Rusty prefers wide function signatures that explicitly pass in the arguments forcing you to purposefully pass in `NULL`).

I think your PR is good as is, and we can bikeshed the details later a bit more :wink: 

ACK https://github.com/ElementsProject/lightning/pull/2944/commits/9119c2a58f288e804d32246ee55ff6e1f4901616

PR #2945, opened 2019-08-11
plugin: `sendpay_success` and `sendpay_failure` notifications

trueptolemy:
Long long ago, @ZmnSCPxj [mentioned](https://github.com/ElementsProject/lightning/pull/2188#issuecomment-448423897) `sendpay_success` and `sendpay_failure` notifications.

My first thought was to combine these two case into one notification type, `sendpay_result`.
But I found the `json` fields are two different between `sendpay` success and failure, so finally I divided `sendpay_result` into `sendpay_success` and `sendpay_failure` again.

For these two notification type, I still use the original notification interface, because I'm not sure whether #2944 is valid.

~I pre-present these ideas here though I'll wait for 0.7.3 to add `CHANGELOG`,~

### DIFF
--- a/tests/plugins/sendpay_notifications.py
+++ b/tests/plugins/sendpay_notifications.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python3
+"""This plugin is used to check that sendpay_success and sendpay_failure calls are working correctly.
+"""
+from lightning import Plugin
+
+plugin = Plugin()
+
+
+@plugin.init()
+def init(configuration, options, plugin):
+    plugin.success_list = []
+    plugin.failure_list = []
+
+
+@plugin.subscribe("sendpay_success")
+def notify_sendpay_success(plugin, sendpay_success):
+    plugin.log("receive a sendpay_success recored, id: {}, payment_hash: {}".format(sendpay_success['id'], sendpay_success['payment_hash']))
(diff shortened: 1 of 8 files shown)


tests/plugins/sendpay_notifications.py
```diff
@@ -0,0 +1,53 @@
...
+            plugin.log("success record exists")
+            check_result = check(success, response)
+            return check_result
+    for fail in plugin.fail_list:
+        if fail['data']['payment_hash'] == payment_hash:
+            plugin.log("fail record exists")
+            check_result = check(fail, response)
+            return check_result
+    plugin.log("no record")
+    return False
```
rustyrussell: FWIW, I just undid similar logic in plugins/forward_payment_status.py.

Turns out it's better to have the plugin just save up the notifications and spew them back on command, then have the comparison logic in the test itself.  That way, if something goes wrong, it's much easier to debug (and also to enhance the tests in future, since they only need to be changed in one place).

trueptolemy: Thank you :-)
I'll learn from plugins/forward_payment_status.py!

trueptolemy:
Modified the test and the test plugin. @rustyrussell What about now?

trueptolemy:
Thank you! :-)
Rebased. Wait for Travis-CI to pass.

PR #2949, opened 2019-08-13
Grind fee bug

rustyrussell:
Fixes: #2820

PR #2951, opened 2019-08-14
ccan: update to fix shutdown slowness after plugin load.

rustyrussell:
Dynamic plugins were keeping fds open; they should not have these
at all anyway, but worse, they interfere with operation because
we don't notice they're closed.

The symptom was that shutdown of the test_plugin_slowinit and
test_plugin_command was 30 seconds (10 seconds grace to kill each daemon).

### DIFF
--- a/ccan/ccan/htable/htable.h
+++ b/ccan/ccan/htable/htable.h
@@ -76,6 +76,15 @@ bool htable_init_sized(struct htable *ht,
 		       void *priv, size_t size);
 
 /**
+ * htable_count - count number of entries in a hash table.
+ * @ht: the hash table
+ */
+static inline size_t htable_count(const struct htable *ht)
+{
+	return ht->elems;
+}
+
+/**
  * htable_clear - empty a hash table.
  * @ht: the hash table to clear
  *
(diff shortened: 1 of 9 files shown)


ccan/ccan/htable/htable.h
```diff
@@ -75,6 +75,15 @@ bool htable_init_sized(struct htable *ht,
 		       size_t (*rehash)(const void *elem, void *priv),
 		       void *priv, size_t size);
 
+/**
```
ZmnSCPxj: Just to be clear, the `htable` changes are not part of the fix, only the `pipecmd` changes are?  Basically `htable` just happened to get pulled in due to the CCAN update?

rustyrussell: Yep!  They're also a useful cleanup (we access .raw.elems in some places) but I'll do that as part of another tree I have waiting for 0.7.3...

rustyrussell:
Lots of timeouts, will rebase on master once #2949 merged which cuts them down...

rustyrussell:
Trivial rebase on master...

rustyrussell:
> ACK [108722e](https://github.com/ElementsProject/lightning/commit/108722ec44e74bba17f21a26cffc6a7041299b25)
> 
> Though the slow tests seem to have been covered in #2953, I suppose you will now remove them here?

No, I missed those two in #2953 and the tests are still timing out.

PR #2952, opened 2019-08-14
json: speed up shutdown.

rustyrussell:
We currently end up sleeping for 1 second for channeld and gossipd:
better to use a normal blocking waitpid and an alarm to wake us in
case they don't exit.

This speeds up `lightning-cli stop` on my machine from 2.008s to 0.008s:
a 286 times speedup!

### DIFF
--- a/lightningd/subd.c
+++ b/lightningd/subd.c
@@ -749,8 +749,16 @@ void subd_req_(const tal_t *ctx,
 	add_req(ctx, sd, type, num_fds_in, replycb, replycb_data);
 }
 
+/* SIGALRM terminates by default: we just want it to interrupt waitpid(),
+ * which is implied by "handling" it. */
+static void discard_alarm(int sig UNNEEDED)
+{
+}
+
 struct subd *subd_shutdown(struct subd *sd, unsigned int seconds)
 {
+	struct sigaction sa, old;
+
 	log_debug(sd->log, "Shutting down");
 
 	tal_del_destructor(sd, destroy_subd);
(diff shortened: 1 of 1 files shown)


lightningd/subd.c
```diff
@@ -749,8 +749,14 @@ void subd_req_(const tal_t *ctx,
 	add_req(ctx, sd, type, num_fds_in, replycb, replycb_data);
 }
 
+static void discard_alarm(int sig UNNEEDED)
```
ZmnSCPxj: So we are discarding the alarm itself, we are simply depending on signals causing syscalls to return `EINTR`?  That seems quite clever.  Maybe a comment or two about this?

cdecker:
Might need a rebase in order to include the Travis CI "fix" :-)

lightningd/subd.c
```diff
@@ -759,19 +767,23 @@ struct subd *subd_shutdown(struct subd *sd, unsigned int seconds)
...
-	while (seconds) {
-		if (waitpid(sd->pid, NULL, WNOHANG) > 0) {
-			return (struct subd*) tal_free(sd);
-		}
-		sleep(1);
-		seconds--;
+	/* Set up alarm to wake us up if child doesn't exit. */
+	sa.sa_handler = discard_alarm;
+	sigemptyset(&sa.sa_mask);
+	sigaction(SIGALRM, &sa, &old);
```
ZmnSCPxj: `valgrind` seems to be complaining that `sa.sa_flags` is uninitialized.

PR #2958, opened 2019-08-15
Prevent Travis timeouts by brute force and a few fixes

rustyrussell:


.travis/build.sh
```diff
@@ -45,7 +45,7 @@ if [ "$SOURCE_CHECK_ONLY" == "false" ]; then
     echo -en 'travis_fold:end:script.2\\r'
 
     echo -en 'travis_fold:start:script.3\\r'
-    make check
+    make -j$PYTEST_PAR check
```
cdecker: ~This is unlikely to be what you want here: `make check` includes things that are more benchmark-like, e.g., `bitcoin/test/run-secret_eq_consttime`, which may now fail due to the increased parallelism, or more precisely the non-constant load cause by running `make pytest` in parallel with those timing sensitive tests would cause them to flake a lot.~

PR #2961, opened 2019-08-15
14 aug plugin check version

SimonVrouwe:
Plugin [dbbackup](https://github.com/lightningd/plugins/pull/40) would like to check `lightningd` version during init, but `plugin.rpc.getinfo()` call puts it in a deadlock with `lightningd`.

@cdecker already mentioned similar, plugins that registered `db_write` hook shouldn't make any rpc calls at all other then `stop`?


contrib/pylightning/lightning/plugin.py
```diff
@@ -530,6 +531,7 @@ def _getmanifest(self, **kwargs):
     def _init(self, options, configuration, request):
         self.rpc_filename = configuration['rpc-file']
         self.lightning_dir = configuration['lightning-dir']
+        self.lightning_version = configuration['lightning-version']
```
cdecker: Notice that this will cause a plugin written against this API crash if it is run against an older `lightningd`. This would remove that crash:
```suggestion
        self.lightning_version = configuration.get('lightning-version', None)
```

It'll check to see if the key exists, and sets `self.lightning_version` to `None` if it doesn't.

SimonVrouwe: Thanks for that  catch, on second thought: do we really want `Plugin` to have the `lightning_version` member? The user can easily pull it from `configuration` himself.

cdecker: That might also be an option. Maybe we can add a `@property` that just looks into the configuration? That way the users have simple access to it :-)

darosior: Or just have the configuration dict as a member ? That would make plugin themselves take care of backward compatibility : if the version entry is part of the dict, < 0.7.2, otherwise we can have it from the entry's value.

SimonVrouwe: Thanks, if we make plugins aware of `lightning-version`, then maybe pass it even earlier with `getmanifest` or as a command line arg when starting the plugin?

Would it be safe to pass a command line arg such as:
`--lightning_version v0.7.2rc1-32-g596c3cf-modded` ?
If we stick to its position, `Plugin` can simply grab sys.argv[2].

darosior: > Thanks, if we make plugins aware of `lightning-version`, then maybe pass it even earlier with `getmanifest` or as a command line arg when starting the plugin?

I think this is cleaner with `init` as it's its actual purpose.

> Would it be safe to pass a command line arg such as: --lightning_version v0.7.2rc1-32-g596c3cf-modded ? If we stick to its position, Plugin can simply grab sys.argv[2]*

It seems ugly to me to rely on starting plugins with command line arguments. Why prefer this over using the `init` message ?

SimonVrouwe: > Why prefer this over using the init message

@darosior Maybe some future fancy plugin wants to make its command line options, or the `dynamic` parameter, dependent on `lightning_version`. If we want plugins to have that version-control-like flexibility (which I think we should) it sounds logic to me to share the version as early as possible. But passing it with `getmanifest` is probably early enough for plugins to make up their mind.

> It seems ugly to me to rely on starting plugins with command line arguments.

Plugins are not required to use it, and there is already the very _uggly_ [`--debugger`](https://github.com/ElementsProject/lightning/blob/804aee766e968d27fa117a3b21dca6269e78aa7d/lightningd/plugin.c#L972) developer arg :)

SimonVrouwe:
~Force pushed, it now just adds an argument such as `--version=v0.7.2.1-13-g9c5f31c` to the plugin command.~

~Skipped adding the `lightning_version` member to the `plugin`, because it would require some more logic to pass certain `pylighting` tests. It also seems a bit redundant as users can parse it from `sys.argv[1]` themselves.~

Then it just occurred to me there is also an `env` variable `LIGHTNINGD_PLUGIN`, set in `plugins_init`. Can we set that to version string? I really don't know what is customary here.

SimonVrouwe:
Done! Now uses `LIGHTNINGD_PLUGIN` env variable, which seems appropriate.

from [archlinux wiki](https://wiki.archlinux.org/index.php/Environment_variables)
"environment variables provide a simple way to share configuration settings between multiple applications and processes in Linux"

lightningd/plugin.c
```diff
@@ -1011,7 +1012,7 @@ void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 	plugins_add_default_dir(plugins,
 				path_join(tmpctx, plugins->ld->config_dir, "plugins"));
 
-	setenv("LIGHTNINGD_PLUGIN", "1", 1);
```
niftynei: Love the approach, but it seems like it'd be better to add a second env called `LIGHTNINGD_VERSION`. As written, `LIGHTNINGD_PLUGIN` is a bool that flags whether or not you're running as a plugin.

SimonVrouwe: Fixed it, thanks

PR #2962, opened 2019-08-16
Makefile: fix occasional race where we fail to compile tools/tests.

rustyrussell:
This simplifies the dependencies:
1. Objs depend on headers, not other objs.
2. Programs depend on objs.
3. A .o file will generally implicitly depend on the .c file it's built from.
4. If a file has a build line, it's often better to list all deps there.
5. I spotted some missing 'make clean' files.

The particular problem in this case seems to be that make would use
tools/test/gen_test.c before it was ready.  It's probably confused by
the use of recursive make via update-mocks, so explicitly split that
into two stages.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2962/commits/bd2c8608f6ec10ab4c5bc3a6c01bc0140e8a56ad

PR #2964, opened 2019-08-16
API: `txprepare` for multiple outputs

### LINKED ISSUE #2679: Feature: withdraw to multiple addresses
Because batching is a good thing, especially if we want to support c*stodial services that might have multiple withdrawal requests from its victims.

It is also a good preparation for eventually implementing #1936 , which requires building transactions that pay to multiple outputs, by combining a multi-output withdraw with #2672 (by using our own wallet as the "external" wallet).

trueptolemy:
Fix #2679 
- Change the paramaters of ~`withdraw` and~ `txprepare` :

~*withdraw* 'output' ['feerate'] ['minconf']~
*txprepare* 'output' ['feerate'] ['minconf']

The 'output' is the array of outputs that must specify 'destination' and 'satoshi'.
Its format is like:
[{'destination': address1, 'satoshi': satoshi1}, {'destination': address2, 'satoshi': satoshi2}, ...]
or
[{'destination': address, 'satoshi': "all"}]).

~There's also 2 related questions:~
~1. Do you think this `output` array is suitable?~
~2. About the test `test_mutiple_withdraw`, it looks a bit duplicative with `test_withdraw`. Should I merge these two tests?~

**EDIT:** Remove the change of *withdraw*, only make *txprepare* support mutiple outputs.

---
- [x] Add CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- JSON API: `txprepare` now uses `outputs` as parameter other than `destination` and `satoshi`
+
 ### Deprecated
 
 Note: You should always set `allow-deprecated-apis=false` to test for
--- a/bitcoin/tx.c
+++ b/bitcoin/tx.c
@@ -30,6 +30,16 @@ int bitcoin_tx_add_output(struct bitcoin_tx *tx, const u8 *script,
 	return i;
 }
 
+int bitcoin_tx_add_multi_outputs(struct bitcoin_tx *tx,
+				 struct bitcoin_tx_output **outputs)
+{
+	for (size_t j = 0; j < tal_count(outputs); j++)
+		bitcoin_tx_add_output(tx, outputs[j]->script,
+				      outputs[j]->amount);
+
+	return tx->wtx->num_outputs;
+}
+
 int bitcoin_tx_add_input(struct bitcoin_tx *tx, const struct bitcoin_txid *txid,
 			 u32 outnum, u32 sequence,
 			 struct amount_sat amount, u8 *script)
--- a/doc/lightning-txprepare.7.md
+++ b/doc/lightning-txprepare.7.md
@@ -6,17 +6,34 @@ internal wallet.
 SYNOPSIS
 --------
 
-**txprepare** *destination* *satoshi* \[*feerate*\] \[*minconf*\]
+**txprepare** *outputs* \[*feerate*\] \[*minconf*\]
 
 DESCRIPTION
 -----------
 
 The **txprepare** RPC command creates an unsigned transaction which
-spends funds from c-lightning’s internal wallet to the address specified
-in *destination*.
+spends funds from c-lightning’s internal wallet to the outputs specified
+in *outputs*.
 
-Effectively, it is the first part of a **withdraw** command, and uses
-the same parameters. The second part is provided by **txsend**.
+The *outputs* is the array of output that include *destination*
+and *amount*(\{*destination*: *amount*\}). Its format is like:
+\[\{address1: amount1\}, \{address2: amount2\}\]
+or
+\[\{address: *all*\}\].
+It supports the any number of outputs.
+
+The *destination* of output is the address which can be of any Bitcoin accepted
+type, including bech32.
+
+The *amount* of output is the amount to be sent from the internal wallet
+(expressed, as name suggests, in amount). The string *all* can be used to specify
+all available funds. Otherwise, it is in amount precision; it can be a whole
+number, a whole number ending in *sat*, a whole number ending in *000msat*,
+or a number with 1 to 8 decimal places ending in *btc*.
+
+**txprepare** is similar to the first part of a **withdraw** command, but
+supports multiple outputs and uses *outputs* as parameter. The second part
+is provided by **txsend**.
 
 RETURN VALUE
 ------------
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -1,6 +1,7 @@
 #include <bitcoin/address.h>
 #include <bitcoin/base58.h>
 #include <bitcoin/script.h>
+#include <ccan/cast/cast.h>
 #include <ccan/tal/str/str.h>
 #include <common/addr.h>
 #include <common/bech32.h>
@@ -163,10 +164,12 @@ static struct command_result *json_prepare_tx(struct command *cmd,
 	struct command_result *res;
 	u32 *minconf, maxheight;
 	struct pubkey *changekey;
+	struct bitcoin_tx_output **outputs;
 
 	*utx = tal(cmd, struct unreleased_tx);
 	(*utx)->wtx = tal(*utx, struct wallet_tx);
 	wtx_init(cmd, (*utx)->wtx, AMOUNT_SAT(-1ULL));
+	outputs = tal_arr(tmpctx, struct bitcoin_tx_output *, 0);
 
 	if (!param(cmd, buffer, params,
 		   p_req("destination", param_bitcoin_address,
@@ -201,8 +204,14 @@ static struct command_result *json_prepare_tx(struct command *cmd,
 	} else
 		changekey = NULL;
 
-	(*utx)->tx = withdraw_tx(*utx, get_chainparams(cmd->ld), (*utx)->wtx->utxos,
-				 (*utx)->destination, (*utx)->wtx->amount,
+	struct bitcoin_tx_output *output = tal(outputs,
+					       struct bitcoin_tx_output);
+	output->script = tal_dup_arr(output, u8, (*utx)->destination,
+				     tal_count((*utx)->destination), 0);
+	output->amount = (*utx)->wtx->amount;
+	tal_arr_expand(&outputs, output);
+	(*utx)->tx = withdraw_tx(*utx, get_chainparams(cmd->ld),
+				 (*utx)->wtx->utxos, outputs,
 				 changekey, (*utx)->wtx->change,
 				 cmd->ld->wallet->bip32_base,
 				 &(*utx)->change_outnum);
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -31,13 +31,6 @@
 #include <wally_bip32.h>
 #include <wire/wire_sync.h>
 
-struct withdrawal {
-	struct command *cmd;
-	struct wallet_tx *wtx;
-	u8 *destination;
-	const char *hextx;
-};
-
 /**
  * wallet_withdrawal_broadcast - The tx has been broadcast (or it failed)
  *
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -93,7 +93,9 @@ static struct command_result *param_bitcoin_address(struct command *cmd,
 					      scriptpubkey)) {
 	case ADDRESS_PARSE_UNRECOGNIZED:
 		return command_fail(cmd, LIGHTNINGD,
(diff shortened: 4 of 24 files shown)


ZmnSCPxj:
`withdraw` is an old, user-facing command, whereas `txprepare` is a relatively new and arguably low-level command.  I would be fine with changing `txprepare` interface but not `withdraw` interface.  Others might have different opinions.

trueptolemy:
> I would be fine with changing `txprepare` interface but not `withdraw` interface. Others might have different opinions.

Thank you, I'll restore `withdraw` to its original.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- JSON API: `txprepare` now uses `output` as parameter other than `destination` and `satoshi`
```
niftynei: this should be called 'outputs', to denote the fact that the expected input is an array

doc/lightning-txprepare.7.md
```diff
@@ -6,17 +6,35 @@ internal wallet.
...
-in *destination*.
-
-Effectively, it is the first part of a **withdraw** command, and uses
-the same parameters. The second part is provided by **txsend**.
+spends funds from c-lightning’s internal wallet to the output specified
+in *output*.
+
+The *output* is the array of outputs that must specify *destination*
+and *satoshi*. Its format is like:
+\[\{*destination*: address1, *satoshi*: satoshi1\},
```
niftynei: Couple of gripes with the format of this

- The bitcoin-cli interface for outputs is [{"address":"amount"}]. Ex: `[{"bcrt1qc80w6na6rhvh9jtjgv9wwjlm8crxfkcfg87gxq":"20000sat"}]`. It seems preferable to mimic this, if possible.
- If not, I think we should rename `satoshi` to `amount`. It's a bit confusing that you can specify the `satoshi` field as `2.999btc`.

bitcoin/tx.c
```diff
@@ -30,6 +30,16 @@ int bitcoin_tx_add_output(struct bitcoin_tx *tx, const u8 *script,
 	return i;
 }
 
+int bitcoin_tx_add_multi_outputs(struct bitcoin_tx *tx,
+				 struct bitcoin_tx_output **outputs)
```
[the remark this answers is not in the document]
trueptolemy: Thank you very much for your review :)
I will align it.

doc/lightning-txprepare.7.md
```diff
@@ -6,17 +6,35 @@ internal wallet.
...
+*destination* is the address which can be of any Bitcoin accepted type,
+including bech32.
+
+*satoshi* is the amount to be withdrawn from the internal wallet (expressed,
+as name suggests, in satoshi). The string *all* can be used to specify
+withdrawal of all available funds. Otherwise, it is in satoshi precision; it
+can be a whole number, a whole number ending in *sat*, a whole number ending
+in *000msat*, or a number with 1 to 8 decimal places ending in *btc*.
+
+**txprepare** is samilar to the first part of a **withdraw** command, but
```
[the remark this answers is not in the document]
trueptolemy: Thank you!

CHANGELOG.md
trueptolemy: Yes, `outputs` is better.

doc/lightning-txprepare.7.md
trueptolemy: > The bitcoin-cli interface for outputs is [{"address":"amount"}]. Ex: [{"bcrt1qc80w6na6rhvh9jtjgv9wwjlm8crxfkcfg87gxq":"20000sat"}]. It seems preferable to mimic this, if possible.

Thank you, I'll try the `[{"address":"amount"}]`.
> If not, I think we should rename satoshi to amount. It's a bit confusing that you can specify the satoshi field as 2.999btc.

Yes, `amount` is better.
BTW, should `fundchannel` and `fundchannel_start` also use `amount` to replace `satoshi`?

bitcoin/tx.c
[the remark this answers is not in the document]
trueptolemy: It's strange that in fact these codes are actually aligned.

CHANGELOG.md
trueptolemy: Fixed.

doc/lightning-txprepare.7.md
trueptolemy: Sorry for blocking so long time...now I changes the format like [{address: amount}]. What do you think about it now?

doc/lightning-txprepare.7.md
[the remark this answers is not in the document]
trueptolemy: Fixed.

wallet/walletrpc.c
```diff
@@ -188,21 +216,96 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+	outputs = tal_arr(tmpctx, struct bitcoin_tx_output *, outputstok->size);
+	out_len = 0;
+	(*utx)->wtx->all_funds = false;
+	json_for_each_arr(i, t, outputstok) {
+		struct amount_sat *amount;
+		const u8 *destination;
+		struct command_result *res;
+
+		/* output format: {destination: amount} */
+		res = param_bitcoin_address(cmd, NULL, buffer, &t[1], &destination);
```
trueptolemy: Here I'd better use a `json_to_bitcoin_address` interface.

wallet/walletrpc.c
```diff
@@ -188,21 +216,96 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+		if (res)
+			return res;
+
+		res = param_sat_or_all(cmd, NULL, buffer, &t[2], &amount);
```
trueptolemy: Also here, the interface about param shouldn't be called directly.

trueptolemy: Fixed.

wallet/walletrpc.c
trueptolemy: Fixed.

bitcoin/tx.c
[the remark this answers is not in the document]
ZmnSCPxj: Not strange at all.  It has to do with tabs being used.  So the extra `+ ` (two characters) inserted by the `diff` is skipped over by the succeeding tab character to align the print to 8-char boundary, however the previous line does not use tab character.

Thus, there is illusion of misalignment, but only in diff.  When applied, this illusion disappears.

Of note, is that by convention we use tab characters in our C code, and tabs are 8-aligned.

wallet/walletrpc.c
```diff
@@ -191,27 +223,106 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+				return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						    "outputs[%zi]: this destination wants"
+						    " all satoshi. The amount of outputs"
+						    " can't be more than 1. ", i);
+			(*utx)->wtx->all_funds = true;
+			break;
+		}
+
+		if (i == 0)
+			(*utx)->wtx->amount = AMOUNT_SAT(0);
```
niftynei: we should do this before the loop, right after we initialize `all_funds = false`

wallet/walletrpc.c
```diff
@@ -191,27 +223,106 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+		/* In fact, the maximum amount of bitcoin satoshi is 2.1e15.
+		 * It can't be equal to/bigger than 2^64.
+		 * On the hand, the maximum amount of litoshi is 8.4e15,
+		 * which also can't overflow. */
+		/* This means this destination need "all" satoshi we have. */
+		if (amount_sat_eq(*amount, AMOUNT_SAT(-1ULL))) {
+			if (outputstok->size > 1)
+				return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						    "outputs[%zi]: this destination wants"
+						    " all satoshi. The amount of outputs"
```
niftynei: amount -> count.  It's potentially unclear what 'amount' you're referring to.

wallet/walletrpc.c
```diff
@@ -191,27 +223,106 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+			(*utx)->wtx->all_funds = true;
+			break;
+		}
+
+		if (i == 0)
+			(*utx)->wtx->amount = AMOUNT_SAT(0);
+
+		if (!amount_sat_add(&(*utx)->wtx->amount, (*utx)->wtx->amount, *amount))
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					    "outputs: The sum of first %zi satoshi"
```
niftynei: satoshi -> outputs

CHANGELOG.md
niftynei: Should we fix this CHANGELOG line too?

niftynei:
Few minor nits, otherwise lgtm

CHANGELOG.md
trueptolemy: I forgot it ;-(. Now I fixed.

wallet/walletrpc.c
trueptolemy: Good catch! Fixed.

wallet/walletrpc.c
trueptolemy: Corrected.

wallet/walletrpc.c
trueptolemy: Yes, thank you. Corrected.

niftynei:
Travis shows that `test_txprepare`  and `test_txprepare_restart` in test_wallet.py are failing with this error

```
E           lightning.lightning.RpcError: RPC call failed: method: txprepare, payload: {'outputs': [{'bcrt1qeyyk6sl5pr49ycpqyckvmttus5ttj25pd0zpvg': 'all'}]}, error: {'code': 302, 'message': 'Output 0sat would be dust'}
```

wallet/walletrpc.c
trueptolemy: Here shouldn't be changed:
At the beginning, we initial the `(*utx)->wtx->amount` as `AMOUNT_SAT(-1ULL)`.
This is the max permissible for `wallet_select_all`.
```
	if (amount_sat_eq(*amount, AMOUNT_SAT(-1ULL))) {
		if (outputstok->size > 1)
			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
					    "outputs[%zi]: this destination wants"
					    " all satoshi. The count of outputs"
					    " can't be more than 1. ", i);
		(*utx)->wtx->all_funds = true;
		break;
	}

	if (i == 0)
		(*utx)->wtx->amount = AMOUNT_SAT(0);
```

After handling the `all` case, we can initial `(*utx)->wtx->amount` again.

trueptolemy:
Rebased and solved the bug.(The reason is [here](https://github.com/ElementsProject/lightning/pull/2964#discussion_r320839901) ).
Wait for Travis-CI to pass.

wallet/walletrpc.c
niftynei: then we're in a split mode, because you've initialized `all_funds` to be false, but left the wtx->amount pegged at what basically signifies `all_funds`. this is not ideal.

the wtx amount should be set to zero at the loop start along with the `all_funds` flag. in the case that the `*amount` is the max, then set (utx)->wtx->amount to be *amount before breaking. this keeps the state fields of the wtx in sync.

trueptolemy: Yes, I got your idea. It's better :+1:

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2964/commits/08882c7cf31edc7abd6aa9ee775dec873e967eae

Issue #2968, opened 2019-08-17
[Bug] 0.7.2: UNIQUE constraint failed: `block.height`


NicolasDorier:
When I am running the tests for btcpay on the new version of clightning, it crashes.

```
2019-08-17T06:52:16.254Z DEBUG lightningd(7): 03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: Balance 1204000msat -> 1205000msat
2019-08-17T06:52:16.258Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: ... , awaiting 1121
2019-08-17T06:52:16.259Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: Got it!
2019-08-17T06:52:16.259Z DEBUG lightning_hsmd(19): Client: Received message 18 from client
2019-08-17T06:52:16.259Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: Sending revoke_and_ack
2019-08-17T06:52:16.259Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: htlc 10: RCVD_REMOVE_ACK_COMMIT->SENT_REMOVE_ACK_REVOCATION
2019-08-17T06:52:16.259Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: sending_revoke_and_ack: HTLC REMOTE 10 = SENT_REMOVE_ACK_REVOCATION/RCVD_REMOVE_ACK_REVOCATION FULFILLED
2019-08-17T06:52:16.261Z DEBUG lightningd(7): lightning_channeld-03b02f952b85ee87bc53235102a393e6d218712f6dfaba2aa327d5b0cf66d89236 chan #1: peer_out WIRE_REVOKE_AND_ACK
[...]

PR #2966, opened 2019-08-17
[Docker] Fixing the environment variable for arm32

NicolasDorier:
This just match the amd64 and arm64 docker

PR #2971, opened 2019-08-19
v0.7.2.1: Nakamoto's Pre-approval by US Congress

rustyrussell:
(Last-minute build fixes, all Rusty's fault!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.7.2] - 2019-08-17: "Nakamoto's Pre-approval by US Congress"
+## [0.7.2.1] - 2019-08-19: "Nakamoto's Pre-approval by US Congress"
 
 This release named by Antoine Poinsot @darosior.
+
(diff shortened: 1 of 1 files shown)


NicolasDorier:
please fix https://github.com/ElementsProject/lightning/issues/2968 before

NicolasDorier:
Also merging this one https://github.com/ElementsProject/lightning/pull/2966 would be nice. I think arm32 is broken without it.

rustyrussell:
> Also merging this one #2966 would be nice. I think docker arm32 is broken without it.

Done, as it's unlikely to make it worse, and we're delayed anyway.

rustyrussell:
> please fix #2968 before. I can't roll out the new version, as it break my test suite consistently.

Thanks!  I've delayed release.

Erk, if we'd released on time, we'd have missed this.  "Fortunately" I committed a build issue in rc2, and so we delayed release to fix that.

Just FYI: we now release -rc1 on the 10th of every second month.  That's a great time to test the release for upstream project like yours.  If we get no new bug reports in 3 working days, we feel free to tag the final one...

NicolasDorier:
@rustyrussell thanks will keep in mind. It is the first time I see c-lightning failing my tests after an update, so I took for granted that non-rc are bug free. :p

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2971/commits/545d0c640a7478277eed334ac6e84da41489a5d7

NicolasDorier:
Tested after 8b85380 and it seems the problem I reported is fixed.

PR #2972, opened 2019-08-19
signet: update genesis hash to use new POW-valid version

kallewoof:
After review, it was concluded that signet genesis blocks should be POW-valid, i.e. they should be "mined" just like the other genesis blocks.

The default signet genesis block changes its nonce from 0 to 621297 and now fulfills difficulty target requirements.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/2972/commits/623e975360c542df53d0ddc117aeeea053010b71

PR #2643, opened 2019-05-16
systemd: wait until network is up

Sjors:
Prevents error when launching with bind-addr.

PR #2975, opened 2019-08-19
systemd: After=network-online.target

Sjors:
I added `Wants=network-online.target` in #2643, but it looks like that's not enough. Adding `After` seems to help, at least on Ubuntu 18.04.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/2975/commits/a68f0ba03499ae09c8eaad7f3ee7b675c2e06e73

Issue #2916, opened 2019-08-07
current master (cd91c0...) does not compile


renepickhardt:
There seems to be an issue with `peer_billboard` in `channeld/channel.c`:

```
git log
commit cd91c06ce9e869f8143a3993729bcf975dd120e9 (HEAD -> master, origin/master, origin/HEAD)
Author: darosior <darosior@protonmail.com>
Date:   Tue Aug 6 17:49:26 2019 +0200

    lightningd/notification: Add missing includes for 'forward_event'
    
    And update test mocks
```

when running make (after running make clean) I get: 
```
channeld/channeld.c: In function ‘billboard_update’:
channeld/channeld.c:214:2: error: ‘shutdown_status’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  peer_billboard(false, "%s%s%s", funding_status,
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          announce_status, shutdown_status);
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
channeld/channeld.c:214:2: error: ‘announce_status’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
channeld/channeld.c:214:2: error: ‘funding_status’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
cc1: all warnings being treated as errors
<builtin>: recipe for target 'channeld/channeld.o' failed
make: *** [channeld/channeld.o] Error 1
```

renepickhardt:
might have been my problem. Cloning to a new / clean git repo worked. I ran `make clean` before but still there seemed to be some leftovers. So we might have to check if everything is getting cleaned successfully.

darosior:
You need to `./configure` / `make update-mocks`. When making the above commit, I wondered to add the `update-mocks` rule to the `default` one in the Makefile.

rustyrussell:
update-mocks uses the linker output to work(!).  So it's fragile, and supposed to be developer-only.  However, a test was added which used it, so now everyone needs it.  However, I think I'll fix that instead, so it no longer is built by default or uses update-mocks, not sure which.

PR #2980, opened 2019-08-21
Makefile cleanup

rustyrussell:
Fixes: #2916


tools/test/Makefile
```diff
@@ -41,8 +41,10 @@ tools/test/gen_print.c: $(TOOLS_WIRE_DEPS)
 	echo '#include "gen_test.h"' > $@
 	$(BOLT_GEN) -P --page impl ${@:.c=.h} test_type < tools/test/test_cases >> $@
 
-
+# Don't require update-mocks hack unless developer mode.
+ifeq ($(DEVELOPER),1)
```
niftynei: ideally this only gets built when make-check is called?

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/2980/commits/e2d4a40e635c5aedf712ae66b6646453df5d758f

tools/test/Makefile
rustyrussell: Up until now we've build tests by default, so I didn't change this.

PR #2988, opened 2019-08-25
WIP lightningd/plugin: add sync'd init for static plugins at startup and handle init errors

SimonVrouwe:
Some static  (`dynamic=false`) [plugins](https://github.com/lightningd/plugins/pull/40) would like to make assumptions about the startup stage `lightningd` is in while the plugin is initializing. For example to assure certain db transactions haven't happened yet.

This PR makes the `init` of all **static** plugins synchronous, i.e. `lightningd's` startup waits until it has received the `init` response of all static plugins. That response can now contain an `error`, which is handled as follow:
- if the plugin is static and `lightning` is starting up, then abort the startup by killing `lightningd`
- if not starting up, the plugin is killed

As I understand, the `dynamic` member currently only determines the behavior of `plugin` stop/start command. Note however that an unloaded (fresh) static plugins can still be started using that command. In that case, `init` can check the `startup` field and now return an error to have itself cleanly killed.

This PR add a rule to static plugins: they cannot make **any** RPC calls from their `init`. ~Which seems to make sense, as an RPC call would just wait for `lightningd` to be up-and-running, which contradicts the static requirement? But haven't checked this with other plugins.~

**edit**: This last part of the story is not so solid and I also misunderstood the call-back system. Maybe I shouldn't try to redefine (hijack) the _static_ property. And perhaps there are other tricks I can use for this particular [plugin](https://github.com/lightningd/plugins/pull/40), like using the `db_write` hook itself to block `lightningd`'s startup in combination with threading. But that needs some further study.

TODO:
- [ ] complete the `json` error handling
- [ ] add tests

Please give feedback!

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -804,7 +804,7 @@ static void plugin_manifest_timeout(struct plugin *plugin)
 
 static void plugin_sync_init_timeout(struct plugin *plugin)
 {
-	log_broken(plugin->log, "The non-dynamic plugin failed to respond to \"init\" in time, terminating.");
+	log_broken(plugin->log, "The static plugin failed to respond to \"init\" in time, terminating.");
 	fatal("Can't recover from plugin failure, terminating.");
 }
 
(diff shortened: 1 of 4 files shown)


SimonVrouwe:
Will close this PR and make a new one that adds a new plugin property `early_conf` which allows plugins to (sync) `init` before db is touched.

PR #3005, opened 2019-08-28
Adding TRACE loglevel

althalusian:
Hi! Just checking if you might be interested in this direction regarding the logging. I made a small addition to the code to try to enable TRACE loglevel so that trace messages wouldn't fill up the log on the DEBUG level (for example gosspid is too active for prod logging in my opinion when trace messages go into debug).

I modified what I found at first, but there could still be some places missing that my grep shotgun didn't initially find, so I hope someone more knowledgeable would take a good look before even considering accepting anything into master. 

I've been running 0.7.1 with these changes for a week without any noticeable problems so at least in normal conditions it seems to work ok. Haven't done any extensive testing yet though.


althalusian:
Ok, travis already found some missing things that passed normal make.

Might need to try to do some real testing before PR. But comments are welcome, especially the info is this is something you might be interested in merging when it's ready.

Issue #2741, opened 2019-06-13
`notify_disconnect` when remote peer sends `error`?


ZmnSCPxj:
While studying to implement #2740, I tried to trace what `channeld` does when it receives an `error`.
As https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-January/000958.html suggests, instead of a new `funding_cancelled` message to free up a `multifundchannel` attempt in case one of the peers fails the funding process while the others have completed the initial commitment signing, we can use `error` instead to free up an attempt to open a channel, that fails for other reasons.

Briefly, a `multifundchannel` requires not only that we split up the funding process, but also that, if *some* of the peers succeed but *others* fail, we ***MUST*** fail the entire `multifundchannel` and tell the succeeded peers to stop wasting their time waiting for the channel to lock in, which will now never happen since we cannot broadcast the transaction.

I proposed the `funding_cancelled` message before, but it got bogged down in stupid arguments about resources and attacks on nodes rather than on the merits of the message itself, so the next best thing is to use the `error` message to tell the peer to forget a channel it accepted and signed commitments for, but which we are now very sure will never get confirmed because of other reasons (such as other peers failing the `multifundchannel`).

This is particularly bad for C-Lightning since we currently restrict each peer to have only one channel.  If our peer makes a `multifundchannel` attempt to us and we complete it faithfully, but another of *its* peers fails the same `multifundchannel`, then we will still hope for the funding to push through even if our peer wants to cancel it (unless we implement `funding_cancelled` or treat `error` during `CHANNELD_AWAITING_LOCKIN` as a proxy for `funding_cancelled`).

So I start the below analysis:

https://github.com/ElementsProject/lightning/blob/ea1c52f3806e64e4bcbf7dc644aab7787760cf61/channeld/channeld.c#L1858-L1870

Okay, so we find `handle_peer_gossip_or_error`: https://github.com/ElementsProject/lightning/blob/ea1c52f3806e64e4bcbf7dc644aab7787760cf61/common/read_peer_msg.c#L149-L151

And see this code: https://github.com/ElementsProject/lightning/blob/ea1c52f3806e64e4bcbf7dc644aab7787760cf61/common/read_peer_msg.c#L165-L173

`is_peer_error` clears `err` if the channel ID is unrecognized and is not all-0s.  If channel ID is unrecognized, and it is not all-0s (signifying "all channels failed") then the `error` should be ignored.

[...]

PR #3006, opened 2019-08-28
Fundchannel: extend `fundchannel_cancel` to work before funding broadcast

### LINKED ISSUE #2740: Feature: Cancel a `fundchannel_start`ed channel after `fundchannel_complete`
Needed to fully implement `multfundchannel` #1936 

I would prefer extending `fundchannel_cancel` to also work after `fundchannel_complete` (as this simplifies a `multifundchannel` implementation: we can just install destructors that `fundchannel_cancel` everything that has been `fundchannel_start`ed, and delete the destructors once all `fundchannel_complete`s finish correctly, i.e. the "kickable RAII pattern"),

Alternately we can add a **non-`dev`** `forgetfundchannel` which `error`s the peer and then forgets the channel, but can only be called after `fundchannel_complete`.

+1 if we can implement a simple check to ensure that the funding tx has 0 confirmations still, as a basic "sanity check" of correct usage.

Either way we should immediately forget the channel if we receive `error` during `CHANNELD_AWAITING_LOCKIN` if we are fundee.

-----

This is needed to handle error cases in `mutlifundchannel`.

Suppose we want to fund to node `$A` and `$B`:

    ADDRA=$(fundchannel_start $A $AVAL)
    ADDRB=$(fundchannel_start $B $BVAL)

    TX=$(txpreparemultidest "[{\"destination\" : $ADDRA, \"amount\" : $AVAL}, {\"destination\": $ADDRB, \"amount\" : $BVAL}]")

    fundchannel_complete $A $TX
    if !fundchannel_complete $B $TX; then
        # Note: cancel $A after complete fundchannel of $A, because
        # we can never safely broadcast the tx now.
        fundchannel_cancel $A || true # ignore errors.
        txcancel $TX
    else
        txbroadcast $TX
    fi

trueptolemy:
Fix #2740 
This PR try to make `fundchannel_cancel` can work after `fundchannel_complete` and before funding broadcast.

### Process
The extended process is:
For funder:
- At the first, funder used `funderchannel_complete`; 
- Then funder calls `funderchannel_cancel`:
  - funder check if the `fundchannel` process completed ==>
  - If completed, funder check if its clightning has broadcast the funding and check if the funding has been onchain ==>
  - If these 2 check pass, funder ask its `channeld` send a error message to fundee ==> 
  - After sending error message, the funder delete(forget) the channel(also delete the peer, it's the limit of `delete_channel()`)

For fundee:
- After funder's `funderchannel_complete`, the channel state is `AWAITING_LOCKIN`. And the fundee begin to wait for funding locked both locally and remotely
- During `AWAITING_LOCKIN`, when fundee receives the error message from funder, fundee will delete(forget) the channel(also delete the peer, because of the limit of #2741 )

### Limits
But here's also some limits:
- **The check for funding broadcast can't completely ensure that the funding transaction isn't broadcast.** We can't know the case if the funding is broadcast by external wallet but the transaction hasn't been onchain.
This also ask caller to use `fundchannel_cancel` correctly: must be before funding broadcast.
- If `fundchannel_cancel` is called after `fundchannel_complete` and we cancel the channel successfully. **Then we must reconnect with the remote peer before `fundchannel_start` again.**
Like #2741 said, when the peer receive error message or when the peer deletes the channels, the peer will disconnect with the remote peer.
The main reason for this is `channel_set_owner(channel, NULL, false)` moves responsibility for the connection to connectd not openingd. If we want to keep connection, rolling back from `channeld` to `openingd` is necessary.

### Open Question
In f77d6c7 (Fix: Store the transaction(broadcast by `txsend`) into DB), I add the label `TX_UNKNOWN`
for the transaction broadcast by `txsend`.
I think maybe we can add optional string field for `txsend` command to specific the label. What do you think about it?

### Unfinished
- [x] Update document

- [x] Add CHANGELOG

~**Note**:  9ebb521 is duplicated with a commit of #2964 , just for making travis-ci happy~

Thanks for @ZmnSCPxj and @niftynei's explanation, direction and help!

### DIFF
--- a/doc/lightning-fundchannel_cancel.7.md
+++ b/doc/lightning-fundchannel_cancel.7.md
@@ -9,10 +9,18 @@ SYNOPSIS
 DESCRIPTION
 -----------
 
-`fundchannel_cancel` is a lower level RPC command. It allows a user to
-cancel an initiated channel establishment with a connected peer.
+`fundchannel_cancel` is a lower level RPC command. It allows channel funder
+to cancel a channel before funding broadcast with a connected peer.
 
-*id* is the node id of the remote peer with which to cancel the
+*id* is the node id of the remote peer with which to cancel.
+
+Note that the funding transaction MUST NOT be broadcast before
+`fundchannel_cancel`. Broadcasting transaction before `fundchannel_cancel`
+WILL lead to unrecoverable loss of funds.
+
+If `fundchannel_cancel` is called after `fundchannel_complete`, the remote
+peer may disconnect when command succeeds. In this case, user need to connect
+to remote peer again before opening channel.
 
 RETURN VALUE
 ------------
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1022,6 +1022,11 @@ def test_funding_external_wallet(node_factory, bitcoind):
 
     l1.daemon.wait_for_log(r'Funding tx {} depth 1 of 1'.format(txid))
 
+    # Check that tx is broadcast by a third party can be catched.
+    # Only when the transaction (broadcast by a third pary) is onchain, we can catch it.
+    with pytest.raises(RpcError, match=r'.* been broadcast.*'):
+        l1.rpc.fundchannel_cancel(l2.info['id'])
+
     for node in [l1, l2]:
         node.daemon.wait_for_log(r'State changed from CHANNELD_AWAITING_LOCKIN to CHANNELD_NORMAL')
         channel = node.rpc.listpeers()['peers'][0]['channels'][0]
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from decimal import Decimal
 from fixtures import *  # noqa: F401,F403
 from flaky import flaky  # noqa: F401
 from lightning import RpcError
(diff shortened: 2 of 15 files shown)


trueptolemy:
@ZmnSCPxj  Thank you! Now I complete the document.

doc/lightning-fundchannel_cancel.7.md
```diff
@@ -11,10 +11,18 @@ SYNOPSIS
...
-`fundchannel_cancel` is a lower level RPC command. It allows a user to
-cancel an initiated channel establishment with a connected peer.
+`fundchannel_cancel` is a lower level RPC command. It allows channel funder
+to cancel a channel before funding broadcast with a connected peer.
 
-*id* is the node id of the remote peer with which to cancel the
+*id* is the node id of the remote peer with which to cancel.
+
+Note that the funding transaction MUST NOT be broadcast before
+`fundechannel_cancel`. Broadcasting transaction before `fundechannel_cancel`
```
ZmnSCPxj: Extraneous `e` in `fundechannel_cancel`.

doc/lightning-fundchannel_cancel.7.md
```diff
@@ -11,10 +11,18 @@ SYNOPSIS
...
+can lead to unrecoverable loss of funds.
+
+If `fundechannel_cancel` is called after `fundchannel_complete`, the remote
```
ZmnSCPxj: Extraneous `e` in `fundechannel_cancel`.

doc/lightning-fundchannel_cancel.7.md
ZmnSCPxj: For that matter "Broadcasting transaction before `fundchannel_cancel` WILL lead to unrecoverable loss of funds", "can" is too weak.

trueptolemy: Thank you! Corrected.

doc/lightning-fundchannel_cancel.7.md
trueptolemy: Corrected.

tests/test_connection.py
```diff
@@ -848,7 +849,37 @@ def test_funding_external_wallet_corners(node_factory, bitcoind):
...
+
+    # Be sure fundchannel_complete is successful
+    assert l1.rpc.fundchannel_complete(l2.info['id'], prep['txid'], txout)['commitments_secured']
+    # Canceld channel after fundchannel_complete
+    assert l1.rpc.fundchannel_cancel(l2.info['id'])['cancelled']
+
+    l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
+    l1.rpc.fundchannel_start(l2.info['id'], amount)['funding_address']
+    assert l1.rpc.fundchannel_complete(l2.info['id'], prep['txid'], txout)['commitments_secured']
+    l1.rpc.txsend(prep['txid'])
```
niftynei: should we also check that it catches that the tx was broadcast by a third party, like bitcoind?

ZmnSCPxj: This is a belt and suspenders check. I  theory the caller MUST NOT call `fundchannel_cancel` after any possibility of broadcast.

trueptolemy: @ZmnSCPxj  Maybe the transaction is blocked in the txmempool of the miner node?

For now, we can't fully ensure that the broadcast was catched from a third pary. Only when the transaction (broadcast by a third pary) is onchain, we can catch it.

On the other hand, because of f77d6c7 , the transaction broadcast by `lightningd` will have a entry in DB so we can catch it.

trueptolemy: in lightningd/channel_control.c (line 669-677):
```
	enum wallet_tx_type type;
	if(wallet_transaction_type(cmd->ld->wallet,
				   &cancel_channel->funding_txid,
				   &type))
		return command_fail(cmd, LIGHTNINGD,
				    "Has the funding transaction been broadcast? "
				    "Please use `close` or `dev-fail` instead.");
```
Here we try to catch the tx broadcast by `lightningd`. We can make sure if the tx is broadcast by `lightningd`.

and in line690-698:
```
	/* Check if the transaction is onchain. */
	/* Note: The above check and this check can't completely ensure that
	 * the funding transaction isn't broadcast. We can't know if the funding
	 * is broadcast by external wallet and the transaction hasn't been onchain. */
	bitcoind_gettxout(cmd->ld->topology->bitcoind,
			  &cancel_channel->funding_txid,
			  cancel_channel->funding_outnum,
			  process_check_funding_broadcast,
			  cancel_channel);
```
Here we check the tx broadcast by a third party. But it's a `incomplete` check because we can't catch the case if the tx hasn't been onchain.

niftynei: > and in line690-698:

Yeah, I saw that. it's not foolproof, but not a bad check. i was suggesting that if we're going to include the bitcoind check we should include a test to verify it works as intended.

trueptolemy: Ok! I'll add the check for bitcoind ;-)

trueptolemy:
Add the check that the tx broadcast by bitcoind can be catched.

trueptolemy:
Rebased and resolved the conflict.

niftynei:
ack https://github.com/ElementsProject/lightning/pull/3006/commits/67c15a57e9116bc28fe0c1d529a4af67c342cc73

PR #3009, opened 2019-08-29
Fix experimental build, make Travis test it.

rustyrussell:
master currently doesn't build with --enable-experimental-features.


.travis.yml
```diff
@@ -9,6 +9,9 @@ before_install:
 
 env:
   - ARCH=64 SOURCE_CHECK_ONLY=true COPTFLAGS="-O3"
+  - ARCH=32 SOURCE_CHECK_ONLY=true DEVELOPER=0
```
cdecker: Given we no longer run the tests in docker we currently don't have a way to test any architecture other than amd64, so maybe we should drop the `ARCH` param altogether?

rustyrussell: Well, at least we check it *compiles* now!

Issue #2977, opened 2019-08-20
Anormal CPU consumption


NicolasDorier:
When I develop for BTCPayServer, I use a docker-compose which run all the dependencies needed for a comfortable dev environment on regtest. (on 0.7.2)

In this docker-compose I have two clightning: A and B where A open channel to B and send payment to B. (never the reverse)

I noticed that `A` was taking 100% of my CPU while being idle. Everything seemed to works fine, just that the CPU was strangely high.

I have not yet tried to update to 0.7.2 in production, so I don't know yet if this is a general problem, if this is specific to 0.7.2, or specific to regtest and my own setup.

I thought reporting it in case somebody could reproduce.

NicolasDorier:
Ok it did not happened to me only. 

https://github.com/btcpayserver/btcpayserver/issues/978

I have good  tracing tool, will eventually figure out the issue, but there is a bug somewhere in 0.7.2

NicolasDorier:
@rustyrussell @cdecker here are some flamegraph I took:

https://aois.blob.core.windows.net/public/1.svg

https://aois.blob.core.windows.net/public/2.svg

NicolasDorier:
I tested again with 0.7.1, no problem.

ZmnSCPxj:
Can you also make flamegraph for 0.7.1?  Thanks.

The flamegraphs for 0.7.2 seem to blame something called `_arch_local_irq_enable`, I am uncertain if this is relevant or not.

NicolasDorier:
@ZmnSCPxj it appears relevant https://aois.blob.core.windows.net/public/3.svg

ZmnSCPxj:
Ouch.  I have no idea what `_arch_local_irq_enable` is, it is nowhere in our codebase.  Looks like digging through the libc and/or kernel is in order ....

NicolasDorier:
@ZmnSCPxj unsure if it can help you our dockerfile is the same as yours https://github.com/btcpayserver/lightning/blob/master/Dockerfile

NicolasDorier:
Might it be the fact we run on alpine... but strange the problem was not here before. And no idea about `__clone` ?

ZmnSCPxj:
`__clone` either means `fork` or `pthread_create`, and since we mostly use `fork` (and we use it *a lot*) it is likely just `fork`.  Unless some library we use actually uses `pthread_create` underneath.

I think there was a recent thing where we had to close fds when starting plugins, lemme try to dig that...

[...]

PR #3011, opened 2019-08-29
[Docker] Use stretch-slim instead of Alpine

NicolasDorier:
This one is not strictly needed, but I figured out that it is weird that arm32 and arm64 are on stretch, but not amd64.

It is also weird to download some third party binaries for having `glibc` available on alpine.

It is also one less moving part when trying to find a bug (like this one https://github.com/ElementsProject/lightning/issues/2977).


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3011/commits/7e57a00763d4ca9b66a42966f6ee60636db07b05

Issue #3007, opened 2019-08-28
plugin: dbbackup "Failed to write to backup: UNIQUE constraint failed: utxoset.txid, utxoset.outnum"


SimonVrouwe:
Was running version `v0.7.2rc2-modded` (no idea why it shows modded) together with the [dbbackup](https://github.com/lightningd/plugins/pull/40/commits/de14f62db7e2eb0310754c9391dbd89c329461ce).

**edit**: maybe plugin issues don't belong here, but maybe it is relevant 

There was a chain reorg 26 aug, which triggered a write failure in the plugin when writing to backup:
```
2019-08-26T15:37:28.388Z DEBUG lightningd(18203): Adding block 591848: 0000000000000000001a190874bde466b6443be941e8add217eadce44bd4799f
2019-08-26T15:37:28.691Z **BROKEN** plugin-dbbackup.py Failed to write to backup: UNIQUE constraint failed: utxoset.txid, utxoset.outnum, SQL-statement: INSERT INTO utxoset ( txid, outnum, blockheight, spendheight, txindex, scriptpubkey, satoshis) VALUES(x'8695bd92b9d1230708afd3ab32a4e8a81c036d1f8b80b85da706f3377e0f5bc6', 1, 591848, NULL, 4, x'0020701a8d401c84fb13e6baf169d59684e17abd9fa216c8cc5b9fc63d622ff8c58d', 1129870000);
```
and crashed the node, see full crash log below. Looks similar to #1280, or maybe #2968?

What I found so far using sqlite3 `.dbinfo` on original db and backup:
- sqlite3 version of c-lightning original db: `software version:    3016002`
- sqlite3 version of the backup: `software version:    3022000`

The plugin runs on python3 (3.6.4 Anaconda Inc.) and its `sqlite3` module shows indeed v3.22.0.

I forgot to make a copy of the original right after the crash, but I do have the backup, which correctly failed the compare-check when restarting the node.

```
2019-08-26T15:37:18.826Z DEBUG lightningd(18203): Removing stale block 591848: 000000000000000000181b1a2354620f66868a723c0c4d5b24e4be8bdfc35a7f
2019-08-26T15:37:25.536Z DEBUG lightningd(18203): Feerate estimate for urgent set to 8651 (was 9421)
2019-08-26T15:37:25.536Z DEBUG lightningd(18203): ... feerate estimate for slow hit floor 253
[...]

PR #3013, opened 2019-08-30
WIP: lightningd/plugin: add field `early_conf`, allow plugin early (sync) config or abort

SimonVrouwe:
Another proposal to address issues raised in #2988, but now add a new plugins category `early_conf`. Plugins that set `early_conf = True` in `getmanifest` will initialize early in lightningd's startup, i.e. before the database is touched.

Useful to make a backup [plugin](https://github.com/lightningd/plugins/pull/40) more bulletproof. The  plugin now doesn't need to worry about db writes that happen before `init` or have to think about what their importance may be when crashing. It can now also safely copy the db during `init` without missing any `PRAGMA` statements (see #3007) or other that doesn't persist in the db (file).

Additionally this PR allows plugin `init` to return an `error`, which is handled as follow:
- At startup (meaning: not started via rpc `plugin start`) an init-error from `early_conf` plugin kills lightningd and aborts lightningd startup early.
- Outside startup (when started via `plugin start` cmd) an init-error from the plugin (any type) makes lightningd kill the plugin.

So `early_conf` plugins can now shutdown `lightningd` before anything happened, this wasn't possible before without hacks.

~Plugins that don't like to be started via rpc `plugin start` can check the configuration['startup'] field and return an init-error and be killed. However, plugins that register the (sync) `db_write` hook (which happens _before_ `init` is called) should then just _pass_ the hook calls when not fully initialized, or maybe it's `manifest` can check `startup = False` field and then not register the hook.~
**edit:** Plugins that are _freshly_ started via rpc `plugin start` and have set `dynamic=False` or `early_conf=True` in their manifest, should probably be killed by lightningd ASAP (in `plugin_manifest_cb`), so that no hook etc. is registered.

TODO:
- [ ] complete json error handling
- [ ] add tests
- [ ] update docs

### DIFF
--- a/contrib/pylightning/lightning/plugin.py
+++ b/contrib/pylightning/lightning/plugin.py
@@ -99,7 +99,7 @@ class Plugin(object):
 
     """
 
-    def __init__(self, stdout=None, stdin=None, autopatch=True, dynamic=True):
+    def __init__(self, stdout=None, stdin=None, autopatch=True, dynamic=True, early_conf=False):
         self.methods = {'init': Method('init', self._init, MethodType.RPCMETHOD)}
         self.options = {}
 
@@ -121,8 +121,9 @@ class Plugin(object):
         self.rpc_filename = None
         self.lightning_dir = None
         self.rpc = None
-        self.startup = True
+        self.startup = None
         self.dynamic = dynamic
+        self.early_conf = early_conf
         self.child_init = None
 
         self.write_lock = RLock()
@@ -527,7 +528,8 @@ class Plugin(object):
             'rpcmethods': methods,
             'subscriptions': list(self.subscriptions.keys()),
             'hooks': hooks,
-            'dynamic': self.dynamic
+            'dynamic': self.dynamic,
+            'early_conf': self.early_conf
         }
 
     def _init(self, options, configuration, request):
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -686,6 +686,9 @@ int main(int argc, char *argv[])
 	/*~ Make sure we can reach the subdaemons, and versions match. */
 	test_subdaemons(ld);
 
+	/* Some *restricted* plugins may init (or abort) early before db touched */
+	plugins_early_config(ld->plugins);
+
 	/*~ Our "wallet" code really wraps the db, which is more than a simple
 	 * bitcoin wallet (though it's that too).  It also stores channel
 	 * states, invoices, payments, blocks and bitcoin transactions. */
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1021,6 +1021,7 @@ void plugins_start(struct plugins *plugins, const char *dev_plugin_debug)
 void plugins_init(struct plugins *plugins, const char *dev_plugin_debug)
 {
 	plugins->pending_manifests = 0;
+	plugins->pending_early_configs = 0;
 	uintmap_init(&plugins->pending_requests);
 
 	plugins_add_default_dir(plugins,
@@ -1040,16 +1041,45 @@ static void plugin_config_cb(const char *buffer,
 			     const jsmntok_t *idtok,
 			     struct plugin *plugin)
 {
-	plugin->plugin_state = CONFIGURED;
+	const jsmntok_t *errortok;
+	/* An init-error during startup from an `early_conf` plugin will crash
+	 * lightningd, outside startup it will kill the plugin (any type). */
+	errortok = json_get_member(buffer, toks, "error");
+	/* FIXME: shouldn't this be a JSMN_OBJECT ? */
+	if (errortok && errortok->type != JSMN_STRING) {
+		/* FIXME: should `static` plugins be treated same way?*/
+		if (plugin->plugins->startup && plugin->early_config)
+			fatal("init error at early config of plugin '%s':"
+				  " \"error\" result is not an object: %.*s", plugin->cmd,
+						    toks[0].end - toks[0].start,
+						    buffer + toks[0].start);
 
+		/* Plugins that don't like to be started at run-time, can return an
+		 * init error ... and be killed cleanly */
+		plugin_kill(plugin,
+			    "\"error\" result is not an object: %.*s",
+			    toks[0].end - toks[0].start,
+			    buffer + toks[0].start);
+	}
+
+	/* TODO: add error code + message
+	 * TODO: check json `id` */
 	if (plugin->plugins->startup && plugin->early_config) {
+		if (errortok)
+			fatal("init error at early config of plugin '%s'", plugin->cmd);
 		plugin->plugins->pending_early_configs--;
 		tal_free(plugin->timeout_timer);
+
 		/* This will break the io_loop in plugins_early_config */
 		if (plugin->plugins->pending_early_configs == 0)
 			io_break(plugin->plugins);
 	}
 
+	if (errortok) {
+		plugin_kill(plugin, "error initializing plugin");
+		return;
+	}
+	plugin->plugin_state = CONFIGURED;
 }
 
 /* FIXME(cdecker) This just builds a string for the request because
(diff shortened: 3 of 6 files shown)


darosior:
> Plugins that are freshly started via rpc plugin start and have set dynamic=False or early_conf=True in their manifest, should probably be killed by lightningd ASAP

Why ? This is a common case where a plugin doesnt want to be stopped before `lightningd` is stopped but can be started at any time.

SimonVrouwe:
> This is a common case

@darosier Do  you have an example of that?  Perhaps I am confused about the (intended) meaning of  `dynamic` here, as trying to stop a static plugin returns "plugin cannot be _managed_ when lightningd is up", well _managed_ sounds as that it includes _starting_, I guess?

darosior:
Actually I originally wanted to show this error for both `start` and `stop` subcommands : but it finally seemed useful to start plugins after `lightningd` startup. Maybe we should change the error. Cf https://github.com/ElementsProject/lightning/blob/49496ab3bac15e2ad00f181e92b8986ada5d9a89/lightningd/plugin_control.c#L29

SimonVrouwe:
> originally wanted to show this error for both start and stop subcommands

But then you first need to start it and wait for  `getmanifest` to return its `dynamic` variable, which would change the design of `plugin` command, i.e. it should then probably return a `command_still_pending`.

@darosior BTW you still haven't told me your case of a plugin that for some reason does not want to be stopped via RPC (`dynamic=False`) but has no problem being started via RPC.

I was also thinking, why are these decisions (start/stop policies etc.) not delegated to the plugin itself by adding a `stop` call in addition to `getmanifest` and `init`? And then have the plugin decide what to do, which (again) would require your `plugin` command return a `command_still_pending`.

Then finally, if the only motivation for this PR is for a plugin that uses the `db_write` hook to create  a backup, then maybe it shouldn't be a _plugin_ because (apparently) it just doesn't fit into that category well enough.

darosior:
> > originally wanted to show this error for both start and stop subcommands
> 
> But then you first need to start it and wait for `getmanifest` to return its `dynamic` variable, which would change the design of `plugin` command, i.e. it should then probably return a `command_still_pending`.

This could indeed have been a pending command until the plugin is either started or fails, but simplicity was prefered.

> @darosior BTW you still haven't told me your case of a plugin that for some reason does not want to be stopped via RPC (`dynamic=False`) but has no problem being started via RPC.

Actually, I answered you just above....

> I was also thinking, why are these decisions (start/stop policies etc.) not delegated to the plugin itself by adding a `stop` call in addition to `getmanifest` and `init`? And then have the plugin decide what to do

I think that's something we want.

> , which (again) would require your `plugin` command return a `command_still_pending`.

As mentioned above, simplicity was prefered when integrated, but feel free to improve and make the `plugin` command (which is not mine..) a `command_still_pending` : I'd be happy to review.

darosior:
> Then finally, if the only motivation for this PR is for a plugin that uses the db_write hook to create a backup, then maybe it shouldn't be a plugin because (apparently) it just doesn't fit into that category well enough.

> Or maybe a plugin is just not the right choice for this or even having a sync db_write hook for plugins was a mistake because it just doesn't fit in well enough.
*(Quote from https://github.com/lightningd/plugins/pull/40)*

Maybe we should just introduce a notification about the DB writes since a backup plugin doesnt need to take control over this writes ? I think we should wait for #2924 if we do this.

cdecker:
Really like the idea of giving the plugin an early warning that it is about to receive notifications and hook calls, but maybe we should make it a `preinit` instread of moving the `init` call?

My main fear is that we are introducing more states into the process, i.e., some plugins get inited early, some late. imho a plugin should be able to signal in the `getmanifest` that it requires a pre-init with some info (lightning-dir) before any other notification or hook call.

I would just make it a `preinit` notification that is delivered just after we `cd` into the lightning-dir and before even opening the DB or starting the `io_loop`. We can make it a notification since it is optional (we decided to make `init` an RPC call in order not to force plugin writers to have to implement notification handling if they didn't need to. Since this is about notifications and hook calls anyway, this rationale is not applicable here). Alternatively we can even make it a hook call, that'd allow us to even kill the main daemon if something goes terribly wrong during startup.

How does that sound? imho it minimizes code movements, plugin state changes, and is nicely backwards compatible, while still accomplishing the goal.

> Maybe we should just introduce a notification about the DB writes since a backup plugin doesnt need to take control over this writes ? I think we should wait for #2924 if we do this.

The key with the `db_write` hook is that it is called synchronously and therefore we can ensure the write was backed up before continuing to make progress. If we were to just use notifications, the plugin may not have stored the write in the backup when it gets killed, resulting in the backup and the DB being out of sync.

SimonVrouwe:
> I would just make it a preinit notification that is delivered just after we cd into the lightning-dir and before even opening the DB

If the notification includes the `configuration` field (same as in `init`) then this looks useful. So it should be send after [handle_opts](https://github.com/ElementsProject/lightning/blob/0e2ec275769b2bb170bb81a39bb6f8fd3171d015/lightningd/lightningd.c#L681) and before [wallet_new](https://github.com/ElementsProject/lightning/blob/0e2ec275769b2bb170bb81a39bb6f8fd3171d015/lightningd/lightningd.c#L693).

If I understand correctly, the preinit notification is a sort of fake `init` signal for the plugin to prepare (check options, db files etc.) and be ready for the first `db_write` hook call. Assumed the preinit notification is handled before the first `db_write` hook call, the plugin can block `lightningd`'s [startup](https://github.com/ElementsProject/lightning/blob/0e2ec275769b2bb170bb81a39bb6f8fd3171d015/lightningd/plugin.c#L1130), until the plugin has completed pre-init. The pre-init could set some flag to tell the `db_write` hook handler to return False and kill the daemon if it wants to abort the startup.

Sounds good. And what about plugins that shouldn't be started at run-time via RPC? Can they return an error in `getmanifest` (when `startup=False`) and be killed before registering their sync hooks?

cdecker:
> Assumed the preinit notification is handled before the first `db_write` hook call, the plugin can block `lightningd`'s [startup](https://github.com/ElementsProject/lightning/blob/0e2ec275769b2bb170bb81a39bb6f8fd3171d015/lightningd/plugin.c#L1130), until the plugin has completed pre-init. The pre-init could set some flag to tell the `db_write` hook handler to return False and kill the daemon if it wants to abort the startup.

Just one caveat: notifications are non-blocking, but they will be delivered in-order. So it is not possible for a notification to hold up startup (and I don't think it should), but it can process the notification and is assured to receive the signals in the correct order.

> Sounds good. And what about plugins that shouldn't be started at run-time via RPC? Can they return an error in `getmanifest` (when `startup=False`) and be killed before registering their sync hooks?

I thought that was the `dynamic` flag: if `dynamic` is set they can be controlled by the `plugin` command, otherwise not. If a plugin requires guarantees on notification delivery and getting all hook calls it should not be marked as `dynamic`, but we can leave that to plugin authors.

SimonVrouwe:
> Assumed the preinit notification is handled before the first db_write hook call, the plugin can block lightningd's startup, until the plugin has completed pre-init.

Meaning it can _effectively_ hold up the `lightningd` startup because, while the (single thread) plugin handles the notification, it cannot handle the hook for which `lightningd` is waiting. If the plugin wants to abort the startup, it can make the hook handler return false. Which is also a bit hack-ish.

**TLDR**; It should probably be a **preinit hook instead of notification**, that gives plugins direct means to hold-up the startup.

> if dynamic is set they can be controlled by the plugin command, otherwise not

Yes, but a plugin needs to be started first to _know_ its preference from its manifest :-)

SimonVrouwe:
I am having second thoughts while trying to code-up a `preinit_hook`. What should be done with the `plugin->plugin_state` around this? Using a hook or notification to (fake) `init` a plugin actually seems to mess up the plugin's state.

> My main fear is that we are introducing more states into the process, i.e., some plugins get inited early, some late.

What exactly do you mean and what is wrong with plugins completing their init in an early stage instead of _somewhere_ in the main loop?

PR #2936, opened 2019-08-10
Readthedocs and markdown manpages

darosior:
As per https://github.com/ElementsProject/lightning/pull/2928#issuecomment-519828918, fixes #2918.

This completely moves on from asciidoc to Markdown for manpages thanks to a [tweaked version of a Python Markdown->Roff converter](https://github.com/darosior/mrkd). That way, manpages can be completely written in Markdown, displayed in readthedocs, and converted to Roff.

There are still some little things to be fixed with final Roff manpages (WIP) but I wanted to get some feedbacks (for example, is the huge number of insertions/deletions due to file creation/deletion/regeneration a problem ?). Here is a sample of a Markdown manpage : https://gist.github.com/darosior/48bfdf65017ad4a10811be5f00adc32d .

Closes #2928

PR #3014, opened 2019-08-30
Cleanups

cdecker:
Just some cleanups regarding the python tests.


doc/Makefile
```diff
@@ -44,7 +44,7 @@ MANPAGES := doc/lightning-cli.1 \
 doc-all: $(MANPAGES) doc/index.rst
 
 $(MANPAGES): doc/%: doc/%.md
-	@if $(CHANGED_FROM_GIT); then echo mrkd $<; mrkd $<; else touch $@; fi
+	if $(CHANGED_FROM_GIT); then mrkd $< $@; else touch $@; fi
```
darosior: You should either `mrkd $< -output $@` or remove the `$@` otherwise you got a `mrkd: error: unrecognized arguments: [the_manpage_name.manpage_number]` when running `make doc-all`. It's weird that travis doesn't fail on this.

cdecker: I think we may have different versions of `mrkd`. My help shows the following:

```bash
$ mrkd --help
usage: mrkd [-h] [-name NAME] [-section SECTION] [-template TEMPLATE]
            [-index INDEX] [-format roff] [-colors friendly]
            source output

positional arguments:
  source              The source man page
  output              The output file

optional arguments:
  -h, --help          show this help message and exit
  -name NAME          The name to use for the man page
  -section SECTION    The section to use for the man page
  -template TEMPLATE  The HTML template file to use
  -index INDEX        An index file to use for HTML links
  -format roff        The output format
  -colors friendly    The Pygments style to use for HTML syntax highlighting
```

Notice that `output` is a positional argument, not a flag argument. This is btw also the `mrkd` that is deployed on travis, and is the one installed via `pip install mrkd`

cdecker:
> I'm not sure if you encounter this error too ? For example :
> 
> ```shell
> rm doc/lightning-autocleaninvoice.7
> make doc-all
> ```

That's exactly the reason for the first commit :-)

doc/Makefile
darosior: I have the tweaked version introduced in #2936. I think you had problem because you used the original version : actually I think most of the modification I made in the tweaked version (https://github.com/darosior/mrkd/commits/master) are necessary to have a clean output. That's why I added this version in `doc/requirements.txt`

darosior: > This is btw also the mrkd that is deployed on travis

Ah I thought Travis would install the one in `doc/requirements.txt`

cdecker: Ah, I had not realized that you customized the `mrkd` utility.

All but the [last commit](https://github.com/refi64/mrkd/commit/79a7365dc6c81a8458ff33ad69ef469edacc6daa) seem to be also addressed by https://github.com/ElementsProject/lightning/pull/3014/commits/14f4e3c61eb2599d417587278e8abc9730c283e7. What's the purpose of that last commit? And would it be ok if we used the pypi version, in order to have less custom code?

darosior: > All but the last commit seem to be also addressed by 14f4e3c. What's the purpose of that last commit?

The Markdown list were badly rendered in Roff, this commit fixes this.

> And would it be ok if we used the pypi version, in order to have less custom code?

I think specifically c36db8c7227d832e655e0d7d8f5a16f2b3e21170 and 79a7365dc6c81a8458ff33ad69ef469edacc6daa are important and cannot be done on our side in a clean way. The first one permits to not have a header imposed by `mrkd` and the second one corrects list rendering.

I wondered to add the tweaked version to the `requirements` in the first place, but I can post it to PyPi if needed (the licence permits this I think) so it looks less hacky.

darosior: I check the failing build from #2924 and it seems that Travis has the same `mrkd` version as you even if it should install the one from `doc/requirements.txt`

cdecker: I think https://github.com/darosior/mrkd/commit/c36db8c7227d832e655e0d7d8f5a16f2b3e21170 is redundant if we format the markdown files correctly, i.e., we provide a well-formed title like https://github.com/ElementsProject/lightning/pull/3014/commits/14f4e3c61eb2599d417587278e8abc9730c283e7 does (just setting it to "text" doesn't seem like an improvement tbh).

As for the indentation of lists, I'm not sure it is important enough to warrant its own. We may want to contribute it upstream and wait for it to trickle through to pypi.

cdecker: Ok, just checked the upstream seems to be inactive (last commit is over a year old). So I think our only option is to keep your branch for the indentation.

Could you make `[output]` a positional argument instead? That way both versions of `mrkd` will still work, the one from pypi will just happen not to have the indentation fix.

darosior: > I think [darosior/mrkd@c36db8c](https://github.com/darosior/mrkd/commit/c36db8c7227d832e655e0d7d8f5a16f2b3e21170) is redundant if we format the markdown files correctly, i.e., we provide a well-formed title like [14f4e3c](https://github.com/ElementsProject/lightning/commit/14f4e3c61eb2599d417587278e8abc9730c283e7) does (just setting it to "text" doesn't seem like an improvement tbh).

This was not about adding "text", but removing
```python
if '--' not in raw:
    sys.exit(f'Invalid header: {raw}')
```
Which imposed the burden on us to duplicate the manpage description when writing it in Markdown.

> As for the indentation of lists, I'm not sure it is important enough to warrant its own. We may want to contribute it upstream and wait for it to trickle through to pypi.

OK

> Could you make [output] a positional argument instead? That way both versions of mrkd will still work, the one from pypi will just happen not to have the indentation fix.

I can drop https://github.com/darosior/mrkd/commit/dfbc8c342b3c634117e9c117f38b39f45c8570e9 (we would have to update the requirements), or we can also use `-output` explicitly. Anyway are you sure about imposing the duplication of the description ?

cdecker: Currently we aren't duplicating the description, we are making it the title :-)

`command-subcommand - Short description` was more of a title anyway imho

doc/lightning-autocleaninvoice.7.md
```diff
@@ -1,5 +1,5 @@
-LIGHTNING-AUTOCLEANINVOICE(7) Manual Page
-=========================================
+lightning-autocleaninvoice -- Set up auto-delete of expired invoice
+===================================================================
 lightning-autocleaninvoice - Set up auto-delete of expired invoice
```
darosior: Trailing description :-)

doc/Makefile
cdecker: fwiw I found `LIGHTNING-CHECK(7) Manual Page` (for the `<h1/>` title and the html `<title/>`) way too shouty and not really informative. With the switch to what formerly was the description we now get an `<h1/>` and `<title/>` similar to this `lightning-fundchannel – Command for establishing a lightning channel` which I find less shouty and more descriptive :wink:

darosior: Ok, I actually looked at the __only__ diff which had a duplicated description x)

> command-subcommand - Short description was more of a title anyway imho

I agree

doc/lightning-autocleaninvoice.7.md
cdecker: Ouch, good catch.

doc/Makefile
darosior: darosior/mrkd@dfbc8c3 will be dropped in a second on my branch

darosior: Fwiw I gave you write access to the repo, since it is used as a dependency here.

trueptolemy: > darosior/mrkd@dfbc8c3 has been dropped on my branch, so doc/requirements.txt needs to be updated

What about adding a tag?

darosior:
Tagged `mrkd` and opened https://github.com/cdecker/lightning/pull/4 which fixes Travis errors.

PR #3015, opened 2019-09-01
JSON-API: `getroute` supports excluding `node`s

### LINKED ISSUE #2894: `getroute` should support `exclude`-ing nodes too
Routing may result in a `NODE`-level error in the `failure_code` (i.e. the bit masked by `0x2000` is set).  While such is (probably?) very rare, they may still exist and we should be robust against these.

Arguably, the current `pay` behavior is not correct if the intermediate hop signalled a `NODE` error.  What it currently does now is to only add the outgoing channel from that node to the `exclude` list.  However, the node itself has the problem, not the outgoing channel.  Future `getroute` can still find a route going through that same node, using an alternative channel, and forwarding payments through it is still likely to be frustrated due to the node continuing to have a problem.

My proposal is to extend the `exclude` argument of `getroute` such that we support either channel-directions or entire nodes.

Implementation thoughts:

* Create a new `struct exclude_entry` that includes `enum exclude_entry_type type;` and `union { struct short_channel_id_dir channel; struct node_id node; } u;`.
* Create boutique `towire`/`fromwire` for `exclude_entry` so we can pass arrays of it to/from `gossipd`.
* When excluding a node, we can treat this as if we excluded every incoming channel to the node.  This prevents the node from being visited, without having to add space for a flag or somesuch on `struct node` (as we are on a strict diet with the `struct node` and `struct channel` types).

trueptolemy:
Fixed #2894 

Following @ZmnSCPxj 's nice and clear proposal in #2894 , now `getroute` supports excluding `node`s by excluding all incoming channels of nodes.

- [x] Add the modification to `pay` to also exclude nodes if the failcode has the NODE bit set.

### DIFF
--- a/doc/lightning-getroute.7.md
+++ b/doc/lightning-getroute.7.md
@@ -40,8 +40,10 @@ route generated. 0.0 means the exact fee of that channel is used, while
 default is 5.0, or up to 5% fee distortion.
 
 *exclude* is a JSON array of short-channel-id/direction (e.g. \[
-"564334x877x1/0", "564195x1292x0/1" \]) which should be excluded from
-consideration for routing. The default is not to exclude any channels.
+"564334x877x1/0", "564195x1292x0/1" \]) or node-id which should be excluded
+from consideration for routing. The default is not to exclude any channels
+or nodes. Note if the source or destination is excluded, the command result
+is undefined.
 
 *maxhops* is the maximum number of channels to return; default is 20.
 
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -2336,7 +2336,7 @@ struct route_hop *get_route(const tal_t *ctx, struct routing_state *rstate,
 			    struct amount_msat msat, double riskfactor,
 			    u32 final_cltv,
 			    double fuzz, u64 seed,
-			    const struct short_channel_id_dir *excluded,
+			    struct exclude_entry **excluded,
 			    size_t max_hops)
 {
 	struct chan **route;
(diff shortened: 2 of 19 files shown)


gossipd/routing.c
```diff
@@ -2310,22 +2310,50 @@ struct route_hop *get_route(const tal_t *ctx, struct routing_state *rstate,
...
+			if (!node)
+				continue;
+
+			struct chan_map_iter i;
+			struct chan *chan;
+			for (chan = first_chan(node, &i); chan; chan = next_chan(node, &i)) {
+				int dir = half_chan_to(node, chan);
+				tal_arr_expand(&saved_capacity, chan->half[dir].htlc_maximum);
+
+				struct short_channel_id_dir *id= tal(tmpctx, struct short_channel_id_dir);
```
ZmnSCPxj: No need to `tal` allocate this.  `tal_arr_expand` does not require `tal` allocated object for second argument.

    struct short_channel_id_dir id;
    id.scid = chan->scid;
    id.dir = dir;
    tal_arr_expand(&excluded_chan, id);

trueptolemy: Thank you! Fixed.

doc/lightning-getroute.7.md
```diff
@@ -40,8 +40,10 @@ route generated. 0.0 means the exact fee of that channel is used, while
 default is 5.0, or up to 5% fee distortion.
 
 *exclude* is a JSON array of short-channel-id/direction (e.g. \[
-"564334x877x1/0", "564195x1292x0/1" \]) which should be excluded from
-consideration for routing. The default is not to exclude any channels.
+"564334x877x1/0", "564195x1292x0/1" \]) or node-id which should be excluded
+from consideration for routing. If a node is specified, all the incoming
```
ZmnSCPxj: No need to specify exact mechanism by which we exclude nodes: as long as the node gets excluded, it is sufficient to just say "the node should be excluded from consideration for routing."

ZmnSCPxj:
How about the modification to `pay` to also exclude nodes if the `failcode` has the `NODE` bit set?

doc/lightning-getroute.7.md
rustyrussell: Well, it matters in the corner case where we're trying to route from the excluded node I guess.

ZmnSCPxj: Possibly, but in theory ideally that "should" fail.  Hmm...

trueptolemy:
@ZmnSCPxj OK! I'll add related modification of `pay` in this PR.

doc/lightning-getroute.7.md
trueptolemy: Now I delete the sentence of "If a node is specified, all the incoming ...".
I also add "Note if the source or destination is excluded, the command result is undefined". :-)

trueptolemy:
Add the modification to pay to also exclude nodes if the failcode has the NODE bit set.

trueptolemy:
> This adds athe erring _node_, not channel?

@ZmnSCPxj  The original code is to add the error channel. Now if the error message sets the NODE bit, I add the error node other than channel. If there's not NODE bit, I just add the error channel.
Here's the code:
```
	if (failcode & NODE) {
		nodeidtok = json_delve(buf, error, ".data.erring_node");
		if (!nodeidtok)
			plugin_err("waitsendpay error no erring_node '%.*s'",
				   error->end - error->start, buf + error->start);
		node_err = true;
	} else {
		scidtok = json_delve(buf, error, ".data.erring_channel");
		if (!scidtok)
			plugin_err("waitsendpay error no erring_channel '%.*s'",
				   error->end - error->start, buf + error->start);
		dirtok = json_delve(buf, error, ".data.erring_direction");
		if (!dirtok)
			plugin_err("waitsendpay error no erring_direction '%.*s'",
				   error->end - error->start, buf + error->start);
	}
```

trueptolemy:
If the message set the NODE bit, excluding the error node should also exlcude the channel that is specified in the message.

ZmnSCPxj:
No objections for 4 days, will merge in.

Issue #2990, opened 2019-08-26
gossip: don't relay back to sending peers


rustyrussell:
We should filter received gossip to avoid sending it back at the next gossip broadcast.

PR #3016, opened 2019-09-02
Dont re-xmit received gossip

rustyrussell:
Closes: #2990


common/per_peer_state.c
```diff
@@ -118,6 +120,7 @@ bool time_to_next_gossip(const struct per_peer_state *pps,
 		*t = time_from_sec(0);
 	else
 		*t = timemono_between(pps->gs->next_gossip, now);
+
```
cdecker: Spurious whitespace change.

common/per_peer_state.c
```diff
@@ -107,8 +110,7 @@ static inline bool timemono_after(struct timemono a, struct timemono b)
 	return time_greater_(a.ts, b.ts);
 }
 
-bool time_to_next_gossip(const struct per_peer_state *pps,
-			 struct timerel *t)
+bool time_to_next_gossip(const struct per_peer_state *pps, struct timerel *t)
```
cdecker: Whitespace only change.

rustyrussell:
It's 10,000 allocs, so that's closer to 10000 x 48 bytes.  We could make it much more efficient with a simple bump allocator, but in practice if you send us 10k updates within 60 seconds, either we're initially syncing, or you're spamming.

rustyrussell:
Trivial rebase

rustyrussell:
OK, since Travis found a (not really but) memleak, I had to do some memleak infra work.  But the world is a nicer place now...

cdecker:
It seems `test_node_reannounce` now is really flaky or broken.

rustyrussell:
Hmm, will investigate today.

PR #3017, opened 2019-09-02
Fix: Update the mrkd dependency

trueptolemy:


trueptolemy:
Just updating can't solve the problem. It seems must be updated in #3014.

Issue #2753, opened 2019-06-17
pylightning: "'description' should be a string, not '#  Frog' (note, we don't allow \\u)"


jarret:
In my Python3 application which uses `pylightning` I am seeing this error:

```Traceback (most recent call last):
  File "/home/ubuntu/btc-rust/git/mart/lib/daemon/daemon.py", line 266, in invoice_c_lightning
    expiry=expiry, preimage=preimage)
  File "/home/ubuntu/.local/lib/python3.6/site-packages/lightning/lightning.py", line 347, in invoice
    return self.call("invoice", payload)
  File "/home/ubuntu/.local/lib/python3.6/site-packages/lightning/lightning.py", line 220, in call
    raise RpcError(method, payload, resp['error'])
lightning.lightning.RpcError: RPC call failed: method: invoice, payload: {'msatoshi': 7000, 'label': 'pOUskOQfR0GyAhSyug5DRg==', 'description': '#  Frog', 'expiry': 600, 'preimage': '9c610ec7b3e3e4c78dc8abdc242da1656402cdc95d73569d86ae7a3ae43f6032'}, error: {'code': -32602, 'message': "'description' should be a string, not '#  Frog' (note, we don't allow \\u)"
```

I was trying to include a frog emoji `🐸` (in the place of the `#` in the error string) in the description which I am feeding it in via the current `pylightning` version:

```$ pip3 show pylightningName: pylightning
Version: 0.0.7
Summary: Client library for lightningd
Home-page: http://github.com/ElementsProject/lightning
Author: Christian Decker
Author-email: decker.christian@gmail.com
License: MIT
Location: /home/ubuntu/.local/lib/python3.6/site-packages
Requires: 
```
Adding an emoji to the description field works fine via `lighning-cli` when it is give the string on the bash console like so:
` ./lightning/usr/bin/lightning-cli --lightning-dir=./lightning-dir invoice 1000 abc123label "🐸  Frog"`

I have determined that the issue is with the `json.dumps()` call in the `_writeobj()` call of `UnixDomainSocketRpc`. The output produced has escaped non-ascii characters with `\u` as the error message is indicating.

I have tried the `ensure_ascii=False` setting and it does work for me, however I am unsure about possible side-effects.

```diff --git a/contrib/pylightning/lightning/lightning.py b/contrib/pylightning/lightning/lightning.py
index e2e46f12..983bdcd2 100644
--- a/contrib/pylightning/lightning/lightning.py
[...]

PR #3018, opened 2019-09-02
Pylightning: pass through non-ascii characters in the JSON input formatting

jarret:
Addresses issue #2753, which describes the problem and exception that was being seen.

This change enables non-ascii characters such as emojis to be included in the description field via pylightning:

Have done a bit of testing by passing in ascii strings:  `\ud83d\udc7d` `\\\\udc74`  and  `\\udc7d` seem to work as being interpreted as literal ascii rather than some unicode character, which is what is intended. I suspect the JSON encoding/decoding is handling this correctly in this mode.

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -172,7 +172,7 @@ class UnixDomainSocketRpc(object):
         self.next_id = 0
 
     def _writeobj(self, sock, obj):
-        s = json.dumps(obj, cls=self.encoder_cls)
+        s = json.dumps(obj, ensure_ascii=False, cls=self.encoder_cls)
         sock.sendall(bytearray(s, 'UTF-8'))
 
     def _readobj(self, sock, buff=b''):
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -2000,7 +2000,6 @@ def test_new_node_is_mainnet(node_factory):
     assert os.path.isfile(os.path.join(basedir, "lightningd-bitcoin.pid"))
 
 
-@pytest.mark.xfail(strict=True)
 def test_unicode_rpc(node_factory):
     node = node_factory.get_node()
     desc = "Some candy 🍬 and a nice glass of milk 🥛."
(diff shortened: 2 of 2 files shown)


rustyrussell:
Definitely needs a test!  We had an issue in the other direction, where we didn't handle RPC output properly.  

But this doesn't seem to work when I tried it:

```
    def test_unicode_input(node_factory):
        l1 = node_factory.get_node()
    
        for s in [ '\ud83d\udc7d', '\\\\udc74', '\\udc7d' ]:
>           inv = l1.rpc.invoice(123000, s, s)

tests/test_invoices.py:498: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
contrib/pylightning/lightning/lightning.py:634: in invoice
    return self.call("invoice", payload)
contrib/pylightning/lightning/lightning.py:222: in call
    "id": self.next_id,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <lightning.lightning.LightningRpc object at 0x7fa38c025e10>
sock = <socket.socket fd=12, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=lightning-rpc>
obj = {'id': 3, 'method': 'invoice', 'params': {'description': '\ud83d\udc7d', 'label': '\ud83d\udc7d', 'msatoshi': 123000}}

    def _writeobj(self, sock, obj):
        s = json.dumps(obj, ensure_ascii=False, cls=self.encoder_cls)
>       sock.sendall(bytearray(s, 'UTF-8'))
E       UnicodeEncodeError: 'utf-8' codec can't encode characters in position 63-64: surrogates not allowed
```

cdecker:
@rustyrussell I added a small test with some emojis and it seems to me that this addresses the issue correctly.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3018/commits/1379435c8b62f29421edaea801f043157b2b3371

PR #3022, opened 2019-09-03
pylightning: adds Millisatoshi __radd__ method

m-schmoock:
This is useful when using stuff like `sum()` on an array of Millisatoshis etc. that would otherwise throw an TypeError, i.e.:

```
$ lightning-cli helpme
"Error while processing helpme: TypeError(\"unsupported operand type(s) for +: 'int' and 'Millisatoshi'\")"
```

I tested this RP with a version of the `helpme` plugin that was iterating on a `Millisatoshi` list.
Also a unittest is in-place that can/should be enriched with further `Millisatoshi` related tests.

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -157,6 +157,9 @@ class Millisatoshi:
     def __mod__(self, other):
         return Millisatoshi(int(self) % other)
 
+    def __radd__(self, other):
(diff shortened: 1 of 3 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3022/commits/942620a707fc59cdea51353f6fbd8870c7c0368e

PR #3024, opened 2019-09-03
Track withdrawal change

niftynei:
Change outputs from withdrawn (and tx_prepare'd) transactions aren't being marked as confirmed since they aren't being added to our txfilter; this fixes this.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3024/commits/a60213ecb72694fcd2858d4fc1280d4f35fa2f43

PR #3026, opened 2019-09-04
lightningd/options: Make a specified lightning_dir absolute

### LINKED ISSUE #3023: Startup fails when specifying a relative `--lightning-dir`
It seems we had a little regression when changing the order of some of the startup operations. Specifying a relative `--lightning-dir` will now result in an error:

```
$ lightningd/lightningd --lightning-dir=r1
lightningd: Failed to open PID file: No such file or directory
```

This is caused by changing working directory and then attempting to open the relative path from inside the directory, which'd result in opening the pid-file in `r1/r1` instead of just `r1`.

We need to either transform relative `--lightning-dir`s into absolute ones, or drop the relative path from the PID-file.

darosior:
This adds a custom `option_set_` function for the `config_dir` in order to handle relative paths for the `--lightning-dir` option.
Fixes #3023.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ changes.
 
 ### Fixed
 
+- Relative `--lightning_dir` is now working again.
+
 ### Security
 
 ## [0.7.2.1] - 2019-08-19: "Nakamoto's Pre-approval by US Congress"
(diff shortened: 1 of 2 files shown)


rustyrussell:
Also, please add to the Fixes line in Changelog, that relative lightning-dir is now working again.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3026/commits/5e411dc1f5b3bdff9ccf4daf90ebbb775fd733c8

PR #3033, opened 2019-09-06
pytest: Collect errors in teardowns and allow running in ramdisk

cdecker:
My machine would accumulate a number of zombie lightningd and bitcoind
processes over time while testing. Investigating this showed that if a fixture
raised an exception during fixture teardown then other fixtures that have not
been torn down would linger around. The issue is that pytest treats exceptions
in fixtures as non-recoverable and therefore will not catch them and call the
remaining ones.

The first commit adds a new fixture, that is there just to collect eventual errors
from other fixtures and ensure that anything that needs to clean up something,
e.g., processes started by the fixture, are cleaned up before we raise an
eventual exception. This is achieved by making any fixture that needs cleaning
up dependent on the teardown_checks fixture, which also serves as central
point to collect errors and printer of eventual errors.

The last commit also adds the ability to select which directory to use as a root
directory for all tests. If you set this to `/dev/shm` you can get quite a speed
increase (roughly halving the test time in my benchmarks on a beefy test
machine).

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -161,38 +161,19 @@ def node_factory(request, directory, test_name, bitcoind, executor, teardown_che
     for e in errs:
         teardown_checks.add_error(e)
 
-    if VALGRIND:
-        for node in nf.nodes:
-            if printValgrindErrors(node):
-                teardown_checks.add_node_error(node, "reported valgrind errors")
+    def map_node_error(nodes, f, msg):
+        for n in nodes:
+            if n and f(n):
+                teardown_checks.add_node_error(n, msg)
 
-    for node in nf.nodes:
-        if printCrashLog(node):
-            teardown_checks.add_node_error(node, "had crash.log files")
-
-    for node in [n for n in nf.nodes if not n.allow_broken_log]:
-        if checkBroken(node):
-            teardown_checks.add_node_error(node, "had BROKEN messages")
-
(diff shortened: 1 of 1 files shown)


tests/fixtures.py
```diff
@@ -29,7 +29,9 @@
 
 @pytest.fixture(scope="session")
 def test_base_dir():
-    directory = tempfile.mkdtemp(prefix='ltests-')
+    d = os.getenv("TEST_DIR", "/tmp")
+
+    directory = tempfile.mkdtemp(prefix='ltests-', dir=d)
```
niftynei: :ok_hand:

niftynei:
`check-python` has some compliants 

``` 
tests/fixtures.py:150:1: E302 expected 2 blank lines, found 1
532tests/fixtures.py:160:5: F841 local variable 'err_count' is assigned to but never used
533Makefile:306: recipe for target 'check-python' failed
```

rustyrussell:
We might find eatmydata gives the same benefits, FWIW.  But this works...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/d1c7df8a708b9eb06bf9b85be889a931fe2a51e3

cdecker:
Thanks for fixing the `check-python` tests @rustyrussell :+1:

PR #3038, opened 2019-09-09
bitcoind: Remove unused type `bitcoind_mode`

trueptolemy:
This type is unused. It seems can be removed?


rustyrussell:
Removing lines: my favorite kind of patch!

Issue #2994, opened 2019-08-26
gossip: ignore redundant updates, filter spammers


rustyrussell:
If an update doesn't change anything, ignore it (unless it's >= 1 week since last update).
Rate-limit nodes which produce updates < 120 seconds apart.

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: Very old (6.5 years), no comments, inactive 2361 days

If you believe this issue should remain open, please comment and we can reopen it.

Issue #2993, opened 2019-08-26
gossip: refuse to generate new updates/announce unless 300 seconds have passed


rustyrussell:
And make sure the bottom timestamp bit alternates.

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: Very old (6.5 years), no comments, inactive 2361 days

If you believe this issue should remain open, please comment and we can reopen it.

PR #3040, opened 2019-09-09
Gossip ratelimiting

rustyrussell:
We more aggressively suppress our own gossip, and we have a ratelimit on others' gossip.  We also enforce timestamp sanity on gossip we receive (I'm surprised we didn't already!).

Closes: #2993 
Closes: #2994


rustyrussell:
Closing to break it up into more reviewable pieces.

PR #3043, opened 2019-09-10
wire: update how experimental csv patches are managed

niftynei:
We used to append new patches to a single file. This caused
some problems and is a lot harder to cleanup later.

This patch moves the experimental patches to their own, individual
patch files, that are named for the current commit of the BOLTDIR.
(Excepting the existing patchfile, which is from the gossip queries
branch).

Also moves from using LOCAL_BOLTDIR to BOLTDIR, since that's likely
to be the directory that the updated RFC files are in. I blew
my rfc repo away by accident a few weeks ago because of this.


wire/Makefile
```diff
@@ -38,10 +38,11 @@ extract-bolt-csv: bolt-precheck
 	@echo '#include <wire/onion_defs.h>' > wire/extracted_onion_wire_csv
 	@$(BOLT_EXTRACT) $(LOCAL_BOLTDIR)/04*.md >> wire/extracted_onion_wire_csv
 
-# Explicit command to add diff of BOLT CSV's to an experimental_csv file
+# Explicit command to add patchfile for BOLT CSV's
 extract-experimental-bolt-csv: bolt-precheck
-	@$(BOLT_EXTRACT) $(LOCAL_BOLTDIR)/0[127]*.md | diff -u wire/extracted_peer_wire_csv - >>wire/extracted_peer_experimental_csv | if [ $$? -lt 0 ];then exit 1;fi
-	@{ echo '#include <wire/onion_defs.h>'; $(BOLT_EXTRACT) $(LOCAL_BOLTDIR)/04*.md; } | diff -u wire/extracted_onion_wire_csv - >>wire/extracted_onion_experimental_csv | if [ $$? -lt 0 ];then exit 1;fi
+	@$(eval BOLT_GIT_SHA := $(shell git -C $(BOLTDIR) log | head -n1 | cut -c 8-16))
```
niftynei: uses the git sha of the bolt-directory

wire/Makefile
```diff
@@ -50,11 +51,15 @@ wire/extracted_onion_experimental_csv:
 	@touch $@
 
 ifeq ($(EXPERIMENTAL_FEATURES),1)
-wire/gen_peer_wire_csv: wire/extracted_peer_experimental_csv wire/extracted_peer_wire_csv
-	@patch --silent -o $@ wire/extracted_peer_wire_csv $<
+wire/gen_peer_wire_csv: wire/extracted_peer_wire_csv
+	@cp $< $@
+	@$(eval EXP_FILES := $(wildcard wire/extracted_peer_experimental_*))
```
niftynei: uses this instead of an inline shell glob because empty globs return themselves, which errors.

wire/Makefile
rustyrussell: `git rev-parse HEAD` ?  But elsewhere we use BOLTVERSION rather than whatever happens to be checked out in $BOLTDIR.  I can see argumetns for both.

The mix of Make and shell here is really doing my head in, too! :boom:

wire/Makefile
rustyrussell: True, and shopt -s nullglob is a bashism :(

Still, using a Makefile var instead of using $(wildcard...) directly below is kinda weird.

Also, please consider GitHub to be transient: anyhing worth commenting like this needs to be in the source itself...

wire/Makefile
```diff
@@ -50,11 +50,16 @@ wire/extracted_onion_experimental_csv:
...
-wire/gen_peer_wire_csv: wire/extracted_peer_experimental_csv wire/extracted_peer_wire_csv
-	@patch --silent -o $@ wire/extracted_peer_wire_csv $<
+EXPERIMENTAL_PEER_PATCHES := $(wildcard wire/extracted_peer_experimental_*)
+EXPERIMENTAL_ONION_PATCHES := $(wildcard wire/extracted_onion_experimental_*)
 
-wire/gen_onion_wire_csv: wire/extracted_onion_experimental_csv wire/extracted_onion_wire_csv
-	@patch --silent -o $@ wire/extracted_onion_wire_csv $<
+wire/gen_peer_wire_csv: wire/extracted_peer_wire_csv
+	@cp $< $@
+	@for exp in $(EXPERIMENTAL_PEER_PATCHES); do patch --silent $@ $$exp; done
```
rustyrussell: Let's do this via a temporary in case things go wrong?  Plus set -e to abort?

```
@set -e; trap "rm -f $@.$$$$" 0; cp $< $@.$$$$; for exp in $(EXPERIMENTAL_PEER_PATCHES); do patch --silent $@.$$$$ $$exp; done; mv $@.$$$$ $@
```

niftynei:
re-upping the ACK https://github.com/ElementsProject/lightning/pull/3043/commits/ecce60e19441382a1fe7b5a9f5f0153b7da58070

PR #3045, opened 2019-09-10
WIP: Allow creating invoices with `description_hash`.

fiatjaf:
I'm creating a PR instead of an issue because it seemed less trashy; Sorry if I'm doing everything wrong. Also, my experience with C is almost none so my commits will all look like crap -- feel free to tell me to shut up and go home (but if you do that please implement `description_hash` for me).

Anyway, I'm not sure if I should create a new `description_hash` option in the `invoice` API or if I should just capture a large description and turn it into a`description_hash`. Then I don't know if I should discard the full description entirely or if I should store it in the `invoices` table. If I don't store the full description in the invoices table then I should store the description hash, in which case I should create that column in the table, right?

Last thing, I think, is that the `decodepay` API should be modified to accept a description so it can check against the description_hash -- but I don't think it should fail if a description isn't provided, or should it? Also a description shouldn't be required when `pay`ing an invoice with description_hash. I'm not sure if something has to be changed somewhere to allow that.


lightningd/invoice.c
```diff
@@ -700,7 +700,6 @@ static struct command_result *json_invoice(struct command *cmd,
 		   p_opt("fallbacks", param_array, &fallbacks),
 		   p_opt("preimage", param_tok, &preimagetok),
 		   p_opt("exposeprivatechannels", param_bool, &exposeprivate),
-		   p_opt("description_hash", param_sha256, &description_hash),
```
darosior: Why removing this parameter ? It might be useful. Anyway if you want to remove it you should drop the first commit instead.

lightningd/invoice.c
```diff
@@ -682,7 +682,7 @@ static struct command_result *json_invoice(struct command *cmd,
 	const u8 **fallback_scripts = NULL;
 	u64 *expiry;
 	struct sha256 rhash;
-	struct sha256 *description_hash;
```
darosior: You might want to keep it as a pointer tu support both parameter and long descriptions

fiatjaf:
This is not ready. I put this PR here instead of an issue to ask questions about how should I proceed.
Yes, my second commit reverts the first, but I can rebase/squash everything in the end.

lightningd/invoice.c
fiatjaf: According to BOLT11 if there's a description_hash in the invoice there shouldn't be a description, and vice-versa. The current API requires a description parameter, so I'm not sure how could we allow a description_hash to be passed.

rustyrussell:
I ripped that functionality out a few versions ago.  There's currently no transport which handles an external description, so the recommendation is to avoid description_hash for now.

fiatjaf:
I wanted this for `lnurl-pay`. That would be one of the transports which would handle the external description. I didn't think that transport should be something standardized and specified in the BOLTs. But now if we're going to have "offers" that could be it.

PR #3053, opened 2019-09-12
fix macOS build & update docs

jrapoport:
Makefile wasn't respecting adding things to LDFLAGS or CPPFLAGS. Besides general goodness, this is necessary to add the proper include paths to avoid a sqlite mismatch error when running the build with a newer version of sqlite installed (e.g. via brew).

Removed -dn & -dy from LDLIBS because clang doesn't understand them and they will cause --enable-static to fail on macOS. Since these are just debugging flags, and can still be passed in if someone really needs them, seems fine.

v0.7.2 wants Python 3.x + mako and make wants xgettext. I updated the docs to reflect  instructions for these additional dependencies to fix those breaks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,15 +188,15 @@ BOLT_DEPS := $(BOLT_GEN)
 
 ALL_PROGRAMS =
 
-CPPFLAGS = -DBINTOPKGLIBEXECDIR="\"$(shell sh tools/rel.sh $(bindir) $(pkglibexecdir))\""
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3053/commits/4b309c178bf45ec4b419721cf523ee09278d36b9

Would like @rustyrussell's feedback on `-dy` and -dn` though.

PR #3057, opened 2019-09-13
Postgresql driver for the multi-database abstraction

cdecker:
This is the counterpart to #2924. It implements the postgresql driver for `lightningd`, the changes to the query rewriting script and the changes to the testing framework to select the DB to test against.

The tests can be run, if you have postgresql installed, using the `TEST_DB_PROVIDER=postgres` environment variable. The tests pass, with one exception `test_reserve_enforcement`, which I am still investigating, and the tests are a bit flaky when running in parallel.

Sadly I had to modify the migrations in order to add the information that `sqlite3` just glossed over (field size, foreign key enforcement in a transaction, autoincrement primary keys). Changing the migrations is something that I warned against back when I created the migration system, however it is safe to do here since we only reorder one single migration pair (since postgres requires the dependency table `peers` to be created before the dependent table `channels`, safe because it was part of the same git commit back in the day and there is no way someone has a DB with one table but not the other), and some changes in types.

The type changes are limited to the `INTEGER` changing to `BIGSERIAL` if it is the primary key, or `BIGINT` if it is backing a `u64`. There are one or two occurrences where I had to change from `TEXT` to `BLOB` (which then gets translated to `BYTEA` for postgres), since `TEXT` requires the contents to be valid UTF-8 strings, which public keys are not.

Finally, I was planning on using `sqlparse` to make the statement rewriting context aware, and safer than simple query-replace, but I found it rather cumbersome to work with, and we are just replacing a couple of types and placeholders anyway, so I ripped that dependency out again in this PR.

Some things that I will try to address before removing the draft status:

 - [x] Auto-detect the location of the `initdb` and the `postgres` binaries instead of hardcoding the path
 - [x] Fix `test_reserve_enforcement`
 - [x] Run `valgrind` with postgres to ensure that everything is still memory-safe.

Looking forward to the feedback :wink:

### DIFF
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -149,6 +149,12 @@ is a relative path, it is relative to the starting directory, not
 readable (we allow missing files in the default case)\. Using this inside
 a configuration file is meaningless\.
 
+
+ \fBwallet\fR=\fIDSN\fR
+Identify the location of the wallet\. This is a fully qualified data source
+name, including a scheme such as \fBsqlite3\fR or \fBpostgres\fR followed by the
+connection parameters\.
+
 .SH Lightning node customization options
 
  \fBalias\fR=\fIRRGGBB\fR
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -644,13 +644,26 @@ void db_commit_transaction(struct db *db)
 	db->in_transaction = NULL;
 }
 
-static struct db_config *db_config_find(const char *driver_name)
+static struct db_config *db_config_find(const char *dsn)
 {
 	size_t num_configs;
 	struct db_config **configs = autodata_get(db_backends, &num_configs);
-	for (size_t i=0; i<num_configs; i++)
-		if (streq(driver_name, configs[i]->name))
+	const char *sep, *driver_name;
+	sep = strstr(dsn, "://");
+
+	if (!sep)
+		db_fatal("%s doesn't look like a valid data-source name (missing \"://\" separator.", dsn);
+
+	driver_name = tal_strndup(tmpctx, dsn, sep - dsn);
+
+	for (size_t i=0; i<num_configs; i++) {
+		if (streq(driver_name, configs[i]->name)) {
+			tal_free(driver_name);
 			return configs[i];
+		}
+	}
+
+	tal_free(driver_name);
 	return NULL;
 }
 
@@ -660,15 +673,16 @@ static struct db_config *db_config_find(const char *driver_name)
 static struct db *db_open(const tal_t *ctx, char *filename)
 {
 	struct db *db;
-	const char *driver_name = "sqlite3";
 
 	db = tal(ctx, struct db);
 	db->filename = tal_strdup(db, filename);
 	list_head_init(&db->pending_statements);
+	if (!strstr(db->filename, "://"))
+		db_fatal("Could not extract driver name from \"%s\"", db->filename);
 
-	db->config = db_config_find(driver_name);
+	db->config = db_config_find(db->filename);
 	if (!db->config)
-		db_fatal("Unable to find DB driver for %s", driver_name);
+		db_fatal("Unable to find DB driver for %s", db->filename);
 
 	tal_add_destructor(db, destroy_db);
 	db->in_transaction = NULL;
@@ -678,8 +692,8 @@ static struct db *db_open(const tal_t *ctx, char *filename)
 	assert(!db->in_transaction);
 
 	db_prepare_for_changes(db);
-	if (db->config->setup_fn)
-		db->config->setup_fn(db);
+	if (db->config->setup_fn && !db->config->setup_fn(db))
+		fatal("Error calling DB setup: %s", db->error);
 	db_report_changes(db, NULL, 0);
 
 	return db;
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -502,7 +502,7 @@ struct db_stmt *db_prepare_v2_(const char *location, struct db *db,
 
 	/* Look up the query by its ID */
 	for (size_t i = 0; i < db->config->num_queries; i++) {
-		if (streq(query_id, db->config->queries[i].query)) {
+		if (streq(query_id, db->config->queries[i].name)) {
 			stmt->query = &db->config->queries[i];
 			break;
 		}
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -55,6 +55,13 @@ static struct migration dbmigrations[] = {
 	 ", PRIMARY KEY (shachain_id, pos)"
 	 ");"),
      NULL},
+    {SQL("CREATE TABLE peers ("
+	 "  id INTEGER"
+	 ", node_id BLOB UNIQUE" /* pubkey */
+	 ", address TEXT"
+	 ", PRIMARY KEY (id)"
+	 ");"),
+     NULL},
     {SQL("CREATE TABLE channels ("
 	 "  id INTEGER," /* chan->id */
 	 "  peer_id INTEGER REFERENCES peers(id) ON DELETE CASCADE,"
@@ -95,13 +102,6 @@ static struct migration dbmigrations[] = {
 	 "  closing_fee_received INTEGER,"
 	 "  closing_sig_received BLOB,"
 	 "  PRIMARY KEY (id)"
-	 ");"),
-     NULL},
-    {SQL("CREATE TABLE peers ("
-	 "  id INTEGER"
-	 ", node_id BLOB UNIQUE" /* pubkey */
-	 ", address TEXT"
-	 ", PRIMARY KEY (id)"
 	 ");"),
      NULL},
     {SQL("CREATE TABLE channel_configs ("
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -459,6 +459,10 @@ static struct migration dbmigrations[] = {
     /* option_static_remotekey is nailed at creation time. */
     {SQL("ALTER TABLE channels ADD COLUMN option_static_remotekey"
 	 " DEFAULT FALSE;"), NULL },
+    {SQL("ALTER TABLE vars ADD COLUMN intval INTEGER"), NULL},
+    {SQL("ALTER TABLE vars ADD COLUMN blobval BLOB"), NULL},
+    {SQL("UPDATE vars SET intval = CAST(val AS INTEGER) WHERE name IN (\"bip32_max_index\", \"last_processed_block\", \"next_pay_index\")"), NULL},
+    {SQL("UPDATE vars SET blobval = CAST(val AS BLOB) WHERE name = 'genesis_hash'"), NULL},
 };
 
 /* Leak tracking. */
@@ -798,13 +802,13 @@ s64 db_get_intvar(struct db *db, char *varname, s64 defval)
 {
 	s64 res = defval;
 	struct db_stmt *stmt = db_prepare_v2(
-	    db, SQL("SELECT val FROM vars WHERE name= ? LIMIT 1"));
+	    db, SQL("SELECT intval FROM vars WHERE name= ? LIMIT 1"));
 	db_bind_text(stmt, 0, varname);
 	if (!db_query_prepared(stmt))
 		goto done;
 
 	if (db_step(stmt))
-		res = atol((const char*)db_column_text(stmt, 0));
+		res = db_column_int(stmt, 0);
 
 done:
 	tal_free(stmt);
@@ -813,10 +817,9 @@ done:
 
 void db_set_intvar(struct db *db, char *varname, s64 val)
 {
-	char *v = tal_fmt(NULL, "%"PRIi64, val);
 	size_t changes;
-	struct db_stmt *stmt = db_prepare_v2(db, SQL("UPDATE vars SET val=? WHERE name=?;"));
-	db_bind_text(stmt, 0, v);
+	struct db_stmt *stmt = db_prepare_v2(db, SQL("UPDATE vars SET intval=? WHERE name=?;"));
+	db_bind_int(stmt, 0, val);
 	db_bind_text(stmt, 1, varname);
 	if (!db_exec_prepared_v2(stmt))
 		db_fatal("Error executing update: %s", stmt->error);
@@ -824,14 +827,13 @@ void db_set_intvar(struct db *db, char *varname, s64 val)
 	tal_free(stmt);
 
 	if (changes == 0) {
-		stmt = db_prepare_v2(db, SQL("INSERT INTO vars (name, val) VALUES (?, ?);"));
+		stmt = db_prepare_v2(db, SQL("INSERT INTO vars (name, intval) VALUES (?, ?);"));
 		db_bind_text(stmt, 0, varname);
-		db_bind_text(stmt, 1, v);
+		db_bind_int(stmt, 1, val);
 		if (!db_exec_prepared_v2(stmt))
 			db_fatal("Error executing insert: %s", stmt->error);
 		tal_free(stmt);
 	}
-	tal_free(v);
 }
 
 /* Will apply the current config fee settings to all channels */
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -41,40 +41,40 @@ static struct migration dbmigrations[] = {
 	 ");"),
      NULL},
     {SQL("CREATE TABLE shachains ("
-	 "  id INTEGER"
-	 ", min_index INTEGER"
-	 ", num_valid INTEGER"
+	 "  id BIGSERIAL"
+	 ", min_index BIGINT"
+	 ", num_valid BIGINT"
 	 ", PRIMARY KEY (id)"
 	 ");"),
      NULL},
     {SQL("CREATE TABLE shachain_known ("
-	 "  shachain_id INTEGER REFERENCES shachains(id) ON DELETE CASCADE"
+	 "  shachain_id BIGINT REFERENCES shachains(id) ON DELETE CASCADE"
 	 ", pos INTEGER"
-	 ", idx INTEGER"
+	 ", idx BIGINT"
 	 ", hash BLOB"
 	 ", PRIMARY KEY (shachain_id, pos)"
 	 ");"),
      NULL},
     {SQL("CREATE TABLE peers ("
-	 "  id INTEGER"
+	 "  id BIGSERIAL"
 	 ", node_id BLOB UNIQUE" /* pubkey */
 	 ", address TEXT"
 	 ", PRIMARY KEY (id)"
 	 ");"),
      NULL},
     {SQL("CREATE TABLE channels ("
-	 "  id INTEGER," /* chan->id */
-	 "  peer_id INTEGER REFERENCES peers(id) ON DELETE CASCADE,"
+	 "  id BIGSERIAL," /* chan->id */
+	 "  peer_id BIGINT REFERENCES peers(id) ON DELETE CASCADE,"
 	 "  short_channel_id TEXT,"
-	 "  channel_config_local INTEGER,"
-	 "  channel_config_remote INTEGER,"
+	 "  channel_config_local BIGINT,"
+	 "  channel_config_remote BIGINT,"
 	 "  state INTEGER,"
 	 "  funder INTEGER,"
 	 "  channel_flags INTEGER,"
 	 "  minimum_depth INTEGER,"
-	 "  next_index_local INTEGER,"
-	 "  next_index_remote INTEGER,"
-	 "  next_htlc_id INTEGER, "
+	 "  next_index_local BIGINT,"
+	 "  next_index_remote BIGINT,"
+	 "  next_htlc_id BIGINT,"
 	 "  funding_tx_id BLOB,"
 	 "  funding_tx_outnum INTEGER,"
 	 "  funding_satoshi BIGINT,"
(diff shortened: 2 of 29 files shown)


cdecker:
Turns out `test_reserve_enforcement` was broken because the `db.execute()` call wasn't committing the DB transaction. So when we restarted, it'd still use the non-zero reserve and failing locally instead of trying and having the remote end slam the door in our face :man_facepalming:

cdecker:
This should be ready for review. The tests now pass correctly and I checked with `valgrind`. I didn't add a Travis-CI config for it, but I'm planning to add a quick check in our homegrown CI for it, since that's cheaper.

Ping @rustyrussell and @ZmnSCPxj :wink:

doc/lightningd-config.5
```diff
@@ -418,7 +424,7 @@ to gain our eternal gratitude!
 
 .SH AUTHOR
 
-Rusty Russell &lt;\fIrusty@rustcorp.com.au\fR&gt; wrote this man page, and
+Rusty Russell &lt;\fBNone\fR (\fIrusty@rustcorp.com.au\fR)&gt; wrote this man page, and
```
rustyrussell: This kind of gratuitous format change seems to keep happening: do we need to fixup the md files somehow?

wallet/db_sqlite3.c
```diff
@@ -24,8 +24,25 @@ static const char *db_sqlite3_fmt_error(struct db_stmt *stmt)
...
 	sqlite3_stmt *stmt;
-	int err;
+	sqlite3 *sql;
+	int err, flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE;
+
+	if (!strstarts(db->filename, "sqlite3://") || strlen(db->filename) < 10)
+		db_fatal("Could not parse the wallet DSN: %s", db->filename);
+
+	/* Strip the scheme from the dsn. */
+	filename = db->filename + strlen("sqlite3://");
```
rustyrussell: Wonder if we could use a `strafter` helper which did this dance for us?

wallet/db.h
```diff
@@ -40,7 +40,7 @@ struct db;
  * devtools/sql-rewrite.py needs to change as well, since they need to
  * generate identical names to work correctly.
  */
-#define SQL(x) NAMED_SQL( __FILE__ ":" stringify(__LINE__) ":" stringify(__COUNTER__), x)
+#define SQL(x) NAMED_SQL( __FILE__ ":" stringify(__COUNTER__), x)
```
rustyrussell: Hmm, I guess we can assume all compilers support __COUNTER__.  Should be fine.

wallet/db.c
```diff
@@ -393,7 +392,7 @@ static struct migration dbmigrations[] = {
     /* Now make sure we have the lower bound block with the first_blocknum
      * height. This may introduce a block with NULL height if we didn't have any
      * blocks, remove that in the next. */
-    {SQL("INSERT OR IGNORE INTO blocks (height) VALUES ((SELECT "
+    {SQL("INSERT INTO blocks (height) VALUES ((SELECT "
```
rustyrussell: Both should be happy with INSERT .... ON CONFLICT IGNORE

wallet/db.c
```diff
@@ -215,7 +214,7 @@ static struct migration dbmigrations[] = {
      * invoices to current time. */
     {SQL("ALTER TABLE invoices ADD paid_timestamp INTEGER;"), NULL},
     {SQL("UPDATE invoices"
-	 "   SET paid_timestamp = strftime('%s', 'now')"
+	 "   SET paid_timestamp = 1"
```
rustyrussell: Fair enough, though we could argue this belongs in the postgres translation, this is good enough since it will Never Happen.

wallet/db_postgres.c
```diff
@@ -1,34 +1,264 @@
...
+
+/* Cherry-picked from here: libpq/src/interfaces/ecpg/ecpglib/pg_type.h */
+#define BYTEAOID		17
+#define INT8OID			20
+#define INT4OID			23
+#define TEXTOID			25
+
+static bool db_postgres_setup(struct db *db)
+{
+	printf("Connecting to %s\n", db->filename);
```
rustyrussell: Probably needs removal?

wallet/db_postgres.c
```diff
@@ -1,34 +1,264 @@
...
-	.column_bytes_fn = NULL,
-	.column_blob_fn = NULL,
-	.column_text_fn = NULL,
-
-	.last_insert_id_fn = NULL,
-	.count_changes_fn = NULL,
-	.setup_fn = NULL,
-	.teardown_fn = NULL,
+    .name = "postgres",
+    .queries = db_postgres_queries,
```
rustyrussell: tabs to spaces gratuitous change?

wallet/db.c
```diff
@@ -684,6 +684,24 @@ static struct db *db_open(const tal_t *ctx, char *filename)
...
+ * that doesn't exist yet, so we need to terminate and restart the DB
+ * transaction.
+ */
+static bool db_query_prepared_may_fail(struct db_stmt *stmt)
+{
+	bool ok = db_query_prepared(stmt);
+	if (!ok) {
+		db_commit_transaction(stmt->db);
+		db_begin_transaction(stmt->db);
+	}
```
rustyrussell: This function is horrifying!  It's safe only as used here, since there's nothing else in the transaction.  Please just put this into db_get_version itself so nobody else thinks they can use it...

wallet/db.c
```diff
@@ -659,15 +659,16 @@ static struct db_config *db_config_find(const char *driver_name)
...
-	const char *driver_name = "sqlite3";
 
 	db = tal(ctx, struct db);
 	db->filename = tal_strdup(db, filename);
 	list_head_init(&db->pending_statements);
+	if (!strstr(db->filename, "://"))
+		db_fatal("Could not extract driver name from \"%s\"", db->filename);
 
-	db->config = db_config_find(driver_name);
+	db->config = db_config_find(db->filename);
```
rustyrussell: This will do weird things if we ever introduce "sqlite-new://" or something.

I think we need to make a temp copy here and continue doing exact matches:

```
const char *sep;
...
sep = strstr(db->filename, "://");
if (!sep)
        db_fatal(...);
driver_name = tal_strndup(tmpctx, db->filename, sep - db->filename);
```

wallet/db.c
```diff
@@ -458,6 +458,10 @@ static struct migration dbmigrations[] = {
     /* option_static_remotekey is nailed at creation time. */
     {SQL("ALTER TABLE channels ADD COLUMN option_static_remotekey"
 	 " DEFAULT FALSE;"), NULL },
+    {SQL("ALTER TABLE vars ADD COLUMN intval INTEGER"), NULL},
```
rustyrussell: New table entirely perhaps (`variables`?), with columns "bip32_max_index INTEGER, last_processed_block INTEGER, next_pay_index INTEGER, genesishash BLOB"?

wallet/wallet.c
```diff
@@ -3094,7 +3096,13 @@ struct amount_msat wallet_total_forward_fees(struct wallet *w)
 	res = db_step(stmt);
 	assert(res);
 
-	db_column_amount_msat(stmt, 0, &total);
+	if (db_column_is_null(stmt, 0))
+		/* Some databases return a NULL result if there are no
+		 * operands for the SUM(). This is the case for example for
+		 * postgres. */
+		total = AMOUNT_MSAT(0);
```
rustyrussell: `COALESCE(SUM(in_msatoshi - out_msatoshi), 0)` is Sequelly I think?

wallet/wallet.c
```diff
@@ -2684,7 +2699,8 @@ void wallet_filteredblock_add(struct wallet *w, const struct filteredblock *fb)
 	struct db_stmt *stmt;
 	if (wallet_have_block(w, fb->height))
 		return;
-	stmt = db_prepare_v2(w->db, SQL("INSERT OR IGNORE INTO blocks "
+
+	stmt = db_prepare_v2(w->db, SQL("INSERT INTO blocks "
 					"(height, hash, prev_hash) "
 					"VALUES (?, ?, ?);"));
```
rustyrussell: ON CONFLICT IGNORE?  Or was the "OR IGNORE" unnecessary?

wallet/wallet.c
```diff
@@ -2883,7 +2902,10 @@ u32 wallet_transaction_height(struct wallet *w, const struct bitcoin_txid *txid)
 		return 0;
 	}
 
-	blockheight = db_column_int(stmt, 0);
+	if (!db_column_is_null(stmt, 0))
+		blockheight = db_column_int(stmt, 0);
+	else
+		blockheight = 0;
```
rustyrussell: Maybe a db_column_int_or_null() helper for this pattern?

wallet/wallet.c
```diff
@@ -3017,15 +3039,79 @@ struct channeltx *wallet_channeltxs_get(struct wallet *w, const tal_t *ctx,
...
+		resolved_time = NULL;
+	}
+
+	if (wallet_forwarded_payment_update(w, in, out, state, failcode, resolved_time))
+		goto notify;
+
 	stmt = db_prepare_v2(w->db,
-			     SQL("INSERT OR REPLACE INTO forwarded_payments ("
+			     SQL("INSERT INTO forwarded_payments ("
 				 "  in_htlc_id"
```
rustyrussell: Upsert went into postgres 9.5, and we seem to be insisting on postgres 10 anyway?

wallet/wallet.c
```diff
@@ -3087,7 +3170,7 @@ struct amount_msat wallet_total_forward_fees(struct wallet *w)
 	bool res;
 
 	stmt = db_prepare_v2(w->db, SQL("SELECT"
-					" SUM(in_msatoshi - out_msatoshi) "
+					" CAST(SUM(in_msatoshi - out_msatoshi) AS BIGINT)"
```
rustyrussell: Hard to find exaclty, but I think CAST(NULL) gives 0 as expected.  Might deserve a comment though?

wallet/db.h
cdecker: Yeah, the `__LINE__` macro didn't work across multiple compilers. I'm now simply using the raw string of the original query, so we could drop this again (or reinstate the `__LINE__` as a debugging hint when working on the rewriter).

I was hoping to shortcut the string comparison a bit by having shorter names, rather than comparing the full statements which are longer, but generating identifiers proved to be more difficult.

wallet/db.c
cdecker: Yep, that was my reason for just setting a dummy value, nobody should have a DB this old, and nothing bad happens if it goes wrong.

wallet/db_sqlite3.c
cdecker: Sounds good, would you like 2 versions: one returning a pointer into the string and the other copying into a separate tallocated string?

I'll fix this up in my cleanups PR at the end of the week.

wallet/db.c
cdecker: Nope, postgres [doesn't like `IGNORE`, it calls it `DO NOTHING`](https://www.postgresql.org/docs/9.5/sql-insert.html). How I love these little nuggets :wink: I'll add a rewrite rule for it

wallet/db_postgres.c
cdecker: Hm, seems a strange switch in my clang-format configuration. I don't remember, what's the canonical indentation mechanism for c-lightning? Not to get into a tabs-vs-spaces war here :stuck_out_tongue_winking_eye:

wallet/db.c
cdecker: Inlined :+1:

wallet/db.c
cdecker: No, that's transposing the table contents, and usually results in huge tables with a single row, that is horrible to load due to its size. Might not be a big deal now, but depending on the DB an update results in a rewrite of the entire row.

I strongly prefer the multiple-rows, typed-columns, approach since we have very few fundamental types (`int`, `u64`, `blob`) that we are storing here.

wallet/db_postgres.c
ZmnSCPxj: One tab character. Indents to align text on previous lines use a mix of tabs followed by 0 to 7 spaces.

wallet/db.c
cdecker: Nope, turns out `sqlite3` doesn't support `ON CONFLICT` statements up until version 3.24 so this is not portable in any case, unless we want to make for a really weird rewrite-rule that transforms a `INSERT OR IGNORE` into a `INSERT INTO ...loads of other stuff... ON CONFLICT DO NOTHING`.

In this case I don't think it's worth it, since it predates v0.6 :wink:

wallet/db.c
cdecker: Just to make sure I went and looked up where the migration was introduced: https://github.com/ElementsProject/lightning/commit/a88c73a41b9b4f5c43b8348e5247bd22578d0042

This predates v0.6 as well.

wallet/wallet.c
cdecker: Yeah, problem is `COALESCE` doesn't result in a `BIGINT`, so a `CAST(COALESCE(field, 0) AS BIGINT)` is required.

wallet/wallet.c
cdecker: Turns out that given the above check for `wallet_have_block` and being in a DB transaction this should be unnecessary.

wallet/wallet.c
cdecker: I had a branch with `db_column_type_or_default()` at some point which I may dig up again if this happens alot.

wallet/wallet.c
cdecker: Yep, but the `ON CONFLICT(field) DO UPDATE` syntax is not in `sqlite3` below 3.24, so we can't use it. Splitting it into two statements was the only really portable way I found.

wallet/wallet.c
cdecker: The combination of `CAST(COALESCE(field, 0) AS BIGINT)` should take care of this.

doc/lightningd-config.5
cdecker: Yep, it seems to be an issue with `mrkd` which assumes we have link text even on bare links. We can fix this in `mrkd` since @darosior has a fork, either by ignoring non-marked up links of default them to the link URL.

cdecker: PR up: https://github.com/darosior/mrkd/pull/1

Will switch the requirements.txt file to use that branch once the PR gets merged.

darosior: I've also PRed it upstream : https://github.com/refi64/mrkd/pull/2

wallet/wallet.c
cdecker: Just noticed that the `COALESCE` call comes in in a later commit so not touching it here.

wallet/db_postgres.c
cdecker: Ok, will need to adjust my `.clang-format` file accordingly.

doc/lightningd-config.5
cdecker: Nice, hadn' noticed that refi64 is still merging PRs :-)

cdecker:
I addressed most of the things that @rustyrussell commented on

 - I didn't add the upsert variant for really old migrations (all pre-dating v0.6) and I had to split the upsert that is still active because sqlite3 lower than 3.24 doesn't understand the `ON CONFLICT` syntax...

For easier review I created a [range-diff](https://gist.github.com/cdecker/d2c0ade8ae5dd69938274ae459b6d8a6) that just contains the changes between 7c5d5de to a997888 :wink:

wallet/db.c
rustyrussell: We should make this explicit then, I think (separate patch at the end).  Remove early migrations, complain if db is before that, and replace them with nicer db statements which are only used on initializing the db.

wallet/wallet.c
rustyrussell: Yech.  Nobody cares about compatibility, do they?!?

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/a9978886c22c86255fc32ef71e6d483f6543d0ab

I would like to see that db migration cleanup though.  With a message for old dbs that they have to migrate to 0.7+ first.

wallet/db.c
cdecker: Yech, that's admitting defeat! I'll give it another go.

cdecker:
Ok, I looked a little bit harder and it turns out rewriting `INSERT OR REPLACE INTO` into the standard SQL upsert form is rather easy, so I did that instead of dumbing down the statements. This way we don't have to break backward compatibility and everybody is happy :-)

I rebased and squashed right away, but here's the range-diff for between the reviewed version and the current head:

```
 1:  9d7560e63 =  1:  52a2a1e27 make: Add configuration detection and linking of libpq
 2:  71953a205 =  2:  4f2c8a27f pytest: Add db_provider and db instances for configurable backends
 3:  f0816fb50 =  3:  42118f221 cli: Add command line option to specify the wallet location
 4:  25d29a16d =  4:  5d1ac31c7 postgres: Add postgres statement rewriting support
 5:  7f92a07ad !  5:  30af21620 db: Move remainder of the sqlite3 into the apropriate file
    @@ -137,24 +137,6 @@
      	err = sqlite3_step(stmt);
      	sqlite3_finalize(stmt);
     @@
    - #if !HAVE_SQLITE3_EXPANDED_SQL
    - 	/* Register the tracing function if we don't have an explicit way of
    - 	 * expanding the statement. */
    --	sqlite3_trace(db->sql, trace_sqlite3, stmt);
    -+	sqlite3_trace(db->conn, trace_sqlite3, stmt);
    - #endif
    - 
    - 	if (!db_sqlite3_query(stmt)) {
    -@@
    - #else
    - 	/* Unregister the trace callback to avoid it accessing the potentially
    - 	 * stale pointer to stmt */
    --	sqlite3_trace(db->sql, NULL, NULL);
    -+	sqlite3_trace(db->conn, NULL, NULL);
    - #endif
    - 
    - 	return true;
    -@@
      
      static void db_sqlite3_close(struct db *db)
      {
 6:  524c8a010 =  6:  bdcda5573 db: Reorder migrations to reflect their relationship
 7:  8cb65ba36 =  7:  6d0172dc1 db: Switch statement lookup to use the original query instead
 8:  c8635a29e !  8:  3bea36c80 db: Adjust some db migrations to be compatible with postgres
    @@ -61,7 +61,7 @@
          {SQL("ALTER TABLE invoices ADD paid_timestamp INTEGER;"), NULL},
          {SQL("UPDATE invoices"
     -	 "   SET paid_timestamp = strftime('%s', 'now')"
    -+	 "   SET paid_timestamp = 1"
    ++	 "   SET paid_timestamp = CURRENT_TIMESTAMP()"
      	 " WHERE state = 1;"),
           NULL},
          /* We need to keep the route node pubkeys and short channel ids to
 9:  55427326c =  9:  ebe1e6ff3 db: Implement postgres driver primitives
10:  7dde498c9 = 10:  628a4a1ca db: Allow some internal queries to fail
11:  c02078fdf = 11:  d1a6f2013 db: Select driver by dsn prefix
12:  be77ae6f1 = 12:  2ccb97226 db: Split the vars table to have type-specific columns
13:  47bc85bc0 ! 13:  1c17a9cc7 db: Implement SQL statement rewriting
    @@ -63,9 +63,11 @@
     +class Sqlite3Rewriter(Rewriter):
     +    def rewrite_single(self, query):
     +        typemapping = {
    -+            'BIGINT': 'INTEGER',
    -+            'BIGINTEGER': 'INTEGER',
    -+            'BIGSERIAL': 'INTEGER',
    ++            r'BIGINT': 'INTEGER',
    ++            r'BIGINTEGER': 'INTEGER',
    ++            r'BIGSERIAL': 'INTEGER',
    ++            r'CURRENT_TIMESTAMP\(\)': "strftime('%s', 'now')",
    ++            r'INSERT INTO[ \t]+(.*)[ \t]+ON CONFLICT.*DO NOTHING;': 'INSERT OR IGNORE INTO \\1;',
     +        }
     +        return self.rewrite_types(query, typemapping)
     +
    @@ -84,6 +86,7 @@
     +
     +        typemapping = {
     +            r'BLOB': 'BYTEA',
    ++            r'CURRENT_TIMESTAMP\(\)': "EXTRACT(epoch FROM now())",
     +        }
     +
     +        query = self.rewrite_types(query, typemapping)
14:  c12fd9fe6 = 14:  977c535b5 db: Change migrations to use types of the correct cardinality
15:  5d183b02f = 15:  2c6bcb7b6 db: Strengthen some null-checks on queries
16:  83bd8ff4d = 16:  6f4cc8b9d db: Check execution when accessing the result of a statement
17:  51b8fa0e4 ! 17:  e489527fc db: Change table field types to be more specific
    @@ -164,7 +164,7 @@
     -    {SQL("ALTER TABLE invoices ADD paid_timestamp INTEGER;"), NULL},
     +    {SQL("ALTER TABLE invoices ADD paid_timestamp BIGINT;"), NULL},
          {SQL("UPDATE invoices"
    - 	 "   SET paid_timestamp = 1"
    + 	 "   SET paid_timestamp = CURRENT_TIMESTAMP()"
      	 " WHERE state = 1;"),
     @@
           * because we cannot safely save them as blobs due to byteorder
18:  b9b28f71b = 18:  9b8c9accc db: Adjust queries to work with postgres
19:  12e1a6448 = 19:  d8bd6e9aa pytest: Skip some tests that assume we have a sqlite3 db on postgres
20:  799b1c9e5 = 20:  4c2a7b482 pytest: Consolidate fee-fetching in test_setchannelfee_usage
21:  d66481ea3 = 21:  514471754 pytest: Have the DB provider search for the postgres binary
22:  a9978886c = 22:  bfce0ba58 pytest: Stabilize test_no_fee_estimate against UTXO selection issues
 -:  --------- > 23:  ccdd5995b fixup! wallet: Annotate migrations using the SQL macro

```

Notice that whenever possible I chose the standardized syntax over the sqlite3 specific, and then rewrote the sqlite3 version, hopefully that reduces the need for many rewrite rules for future DBs.

Issue #3054, opened 2019-09-13
The segfault related to "get multiple addresses from hostname" is still present


darwin:
I'm sorry but #3046 is not resolved for me. I believe you fixed a different segfault.

I'm still able to reproduce it both on my dev machine and in our travis build[2] with [recent master](https://github.com/ElementsProject/lightning/commit/a988ded3fabdf70271d68f04feba5c43c726b0d2)[1] which should contain a fix for #3046.

[1] https://travis-ci.org/darwin/simverse/builds/584484170#L318
[2] https://travis-ci.org/darwin/simverse/builds/584484170#L3134-L3140

darosior:
I'm not able to reproduce this, but if it is still because of the same commit as in #3046 this might be because we don't use an initialized array in `parse_wireaddr`. Could you try this patch ?
```
diff --git a/common/wireaddr.c b/common/wireaddr.c
index 5e87f3b5..170671ef 100644
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -405,8 +405,12 @@ bool parse_wireaddr(const char *arg, struct wireaddr *addr, u16 defport,
        }
 
        /* Resolve with getaddrinfo */
-       if (!res)
-               res = wireaddr_from_hostname(&addr, ip, port, no_dns, NULL, err_msg);
+       if (!res) {
+               struct wireaddr *addresses = tal_arr(NULL, struct wireaddr, 1);
+               res = wireaddr_from_hostname(&addresses, ip, port, no_dns, NULL, err_msg);
+               *addr = addresses[0];
+               tal_free(addresses);
+       }
```

darwin:
With your patch it does not crash (good news!), but still fails to resolve:
```
> alice connect 021c36f49c6271ae7469b5eb93224002c27dd5ad50821717d795df2cead62c93e5@bob
{
  "code": -1,
  "message": "Unknown type 184 :0: opening -1 socket gave Protocol not supported. "
}
```

If you wanted to play with it, here is a repro case:
```
> git clone --recursive https://github.com/darwin/simverse.git
> cd simverse
> ./sv repos init
> export SIMVERSE_TEST_REPOS=_repos
> echo -e "#!/usr/bin/env bash\ncd \"\$(dirname \"\${BASH_SOURCE[0]}\")\"\n./../cases/01-pay-directly/test.sh a1k2" > tests/suites/_x.sh
```

You can tweak `_repos/lighting` and changes should be reflected in each subsequent test run:
[...]

PR #3058, opened 2019-09-14
Some more fixes in wireaddr

darosior:
Fixes: #3054

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -308,19 +308,21 @@ bool wireaddr_from_hostname(struct wireaddr **addrs, const char *hostname,
 	if (strends(hostname, ".onion")) {
 		u8 *dec = b32_decode(tmpctx, hostname,
 				     strlen(hostname) - strlen(".onion"));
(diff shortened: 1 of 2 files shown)


common/wireaddr.c
```diff
@@ -405,8 +405,12 @@ bool parse_wireaddr(const char *arg, struct wireaddr *addr, u16 defport,
 	}
 
 	/* Resolve with getaddrinfo */
-	if (!res)
-		res = wireaddr_from_hostname(&addr, ip, port, no_dns, NULL, err_msg);
+	if (!res) {
+		struct wireaddr *addresses = tal_arr(NULL, struct wireaddr, 0);
+		res = wireaddr_from_hostname(&addresses, ip, port, no_dns, NULL, err_msg);
+		*addr = addresses[0];
+		tal_free(addresses);
```
rustyrussell: Hmm, it'd be nice if parse_wireaddr handled an array, too, and connectd took that array so we could try connecting to multiple addresses if they resolve that way.  But that's a bigger task...

Issue #2665, opened 2019-05-23
Support for hacking...


ZmnSCPxj:
The old #614 was already hopelessly buried in page 8 of issues.

So, I remake this thread here.

For those contributing now, if you have any questions about the architecture of `lightningd` and so on, please feel free to post them here.

@m-schmoock @darosior @trueptolemy @niftynei @bisoge

darosior:
Thanks for that; is there a way to quiet the deprecation warning when running tests ?

ZmnSCPxj:
>is there a way to quiet the deprecation warnings when running tests ?

Did not #2681 fix?

Sorry, I am not python expert, maybe @cdecker can field this question.

darosior:
Some of them but some are from the dependencies (`cheroot`, `jinja2`), they might be resolved soon but until then I wanted to know if there is a way to avoid the endless scroll on pytest error.

jdabs:
Is there an official API reference for lightning-cli ? Or is the README on this repo the main one?

The README was helpful and I was able to set up my node easily with it. The main draw back is that it is not great as a full reference doc because it switches between different command lines tools, such as bitcoind, bitcoin-cli, and lightningd

ZmnSCPxj:
They are in the `docs` directory; every `docs/lightning-*.txt` file describes a command.  If you `make install` they should be available as `man` pages e.g. `man 7 lightning-pay`

trueptolemy:
@ZmnSCPxj Hi, when I study the route, I meet a question....
In this test:
https://github.com/ElementsProject/lightning/blob/04a780c7158655ddc7c391c726c4839b55aa8481/tests/test_pay.py#L82-L104

For my understanding, this pay attempt will generate a piece of `routehint`, that includes the channel from `l2` to `l3`?
So why not 3 attempt here?
One for normal attempt, one excluding the channel between l1 and l2, and one with the `routehint`?

For this test, I find there 2 attempts, one is the normal attempt, and another one is exluding the channel from l2 to l3.

So...my question is if there's an error?
Because we should exclude channel l1->l2 other than l2->l3?

I find the `find_worst_channel` interface in `pay` plugin may be wrong, but I'm not sure...

ZmnSCPxj:
I am uncertain myself --- I have not looked deeply into the `pay` plugin very much yet.  I will try to check some time "soon", possibly, for some definition of "soon".

darosior:
Hi (/good morning?),

[...]

PR #3061, opened 2019-09-15
Pay plugin: Fix the traversal order in `find_worst_channel`

trueptolemy:
Related with my ['question' comment](https://github.com/ElementsProject/lightning/issues/2665#issuecomment-531479284)

I guess the original idea is to use the reverse order to traverse the entire route collection to find the worst channel/hop. I think it's the reason of the 'abnormal performance' in `test_pay_limits` if I'm right...

But the reverse order is hard to implement for our json interface, so I fix the logic with `json_for_each_arr`.

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -451,22 +451,31 @@ static struct command_result *sendpay_error(struct command *cmd,
 
 static const jsmntok_t *find_worst_channel(const char *buf,
 					   const jsmntok_t *route,
-					   const char *fieldname,
-					   u64 final)
+					   const char *fieldname)
 {
-	u64 prev = final, worstval = 0;
-	const jsmntok_t *worst = NULL, *t;
+	u64 prev, worstval = 0;
+	const jsmntok_t *worst = NULL, *t, *t_prev = NULL;
 	size_t i;
 
 	json_for_each_arr(i, t, route) {
 		u64 val;
 
 		json_to_u64(buf, json_get_member(buf, t, fieldname), &val);
-		if (worst == NULL || val - prev > worstval) {
-			worst = t;
-			worstval = val - prev;
+
+		/* For the first hop, now we can't know if it's the worst.
+		 * Just store the info and continue. */
+		if (!i) {
+			prev = val;
+			t_prev = t;
+			continue;
+		}
+
+		if (worst == NULL || prev - val > worstval) {
+			worst = t_prev;
+			worstval = prev - val;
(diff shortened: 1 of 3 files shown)


plugins/pay.c
```diff
@@ -454,19 +454,38 @@ static const jsmntok_t *find_worst_channel(const char *buf,
...
+	const jsmntok_t *worst = NULL, *t, *t_prev = NULL;
 	size_t i;
 
 	json_for_each_arr(i, t, route) {
 		u64 val;
+		/* If there's only 2 hops, the worst channel must be the first one. */
+		if (route->size <= 2) {
+			worst = t;
+			break;
+		}
```
rustyrussell: Don't special-case this, I don't think it needs it?

plugins/pay.c
```diff
@@ -454,19 +454,38 @@ static const jsmntok_t *find_worst_channel(const char *buf,
...
+		 * Just store the info and continue. */
+		if (!i) {
+			prev = val;
+			t_prev = t;
+			continue;
+		}
+
+		/* the final hop */
+		if (i + 1 == route->size)
+			assert(val == final);
```
rustyrussell: We should just drop the final arg, it's now redundant.

plugins/pay.c
trueptolemy: Yes...there's no need.

plugins/pay.c
trueptolemy: Yes, I remove this, and also remove the `final` parameter in this interface.

trueptolemy:
@rustyrussell I'm also not sure about the sencond commit https://github.com/ElementsProject/lightning/pull/3061/commits/07f506f17dea7a987c696c45ae77772eacd6ecc3 :
Is it ok if I add more than one `destructor` for a `tal` point?
Here `master` as the `struct daemon_conn` already has a `destructor`, and I add another special `destructor` on it.

trueptolemy:
Thank you! Very soon! :-)

rustyrussell:
> @rustyrussell I'm also not sure about the sencond commit [07f506f](https://github.com/ElementsProject/lightning/commit/07f506f17dea7a987c696c45ae77772eacd6ecc3) :
> Is it ok if I add more than one `destructor` for a `tal` pointer?
> Here `master` as the `struct daemon_conn` already has a `destructor`, and I add another special `destructor` on it.

Yep, completely legal.

Issue #3001, opened 2019-08-27
Proposal: `routetricks` plugin


ZmnSCPxj:
# Introduction

In https://github.com/ElementsProject/lightning/pull/2890#issuecomment-519380842 @rustyrussell expresses the opinion to move `permuteroute` to a plugin.

I am currently busy on `getroutequick`, and additionally, also in non-Bitcoin-related tasks, thus cannot do this yet.

So I add this sketch here.

`routetricks` is a "built=in" plugin, as it will provide some functionality that will (eventually) be done by `pay`, which itself is a built-in plugin.

This provides the commands below:

1.  `recalcroute` - Recompute the fees of a route.
2.  `smoothenroute` - Remove cycles in a route.
3.  `permuteroute` - Modify an existing route to avoid known-failing channels / nodes.

# `recalcroute`

Usage:

    recalcroute route [msatoshi] [cltv] [fromid]

Recompute the fees and delays of the given route, or fail if any channel or node is not existing in our local routemap.

This simply does `listchannels` on each channel along the `route` to get the `delay`, `base_fee_millisatoshi`, and `fee_per_millionth`, starting from the end.

In addition, the `route` is validated so that each hop connects to the previous via the given channel, and that the first hop is directly connected to `fromid` (if specified), or our own node (if `fromid` not specified).

If `msatoshi` and `cltv` are unspecified, they will be extracted from the last hop of the input route.

## Usefulness

* Needed for `smoothenroute`, which is needed for `permuteroute`.
* In the future, for multipath payments, we can `recalcroute` a route which fails a given amount to try to use a smaller amount, hence why `msatoshi` can be specified.
* In the future, if `pay` receives a `failure_code` with the `UPDATE` bit set, rather than add the specified channel to the current `excludes` set, it should attempt a `recalcroute` first: if `recalcroute` succeeds and gives a different value at the first hop than the input route, then it should re-attempt the same route with the new fee setup as long as the new setup is still within the `maxfeepercent` and `maxdelay`.
  * This is vital if in the future we want to widely deploy a "passive-rebalance" plugin, where we adjust our channel fees via `setchannelfees` depending on how much we own of a channel (set fee to 0 if e.g. >75% is owned by us, double the normal fee if e.g. <25% is owned by us, etc.).  If `common `pay` implementations cannot handle `UPDATE` failures, then it would be negative utility to install such a plugin.
      * This would help us make "fee-free rebalance" needed for widespread JIT-Routing support @renepickhardt 

[...]

PR #3070, opened 2019-09-18
JSON-API: Add the `maxscannodes` parameter in `getroute`

trueptolemy:
This is proposed by @ZmnSCPxj . The idea is based on his [comment](https://github.com/ElementsProject/lightning/pull/2890#issuecomment-519499783).
Limitting the number of scanning node can get the partial result of `getroute` quickly. This partial route will be used in `permuteroute`(#2890 and #3001 ).

It's better to post this change before `routetricks` plugin.

### DIFF
--- a/doc/lightning-getroute.7.md
+++ b/doc/lightning-getroute.7.md
@@ -5,7 +5,7 @@ SYNOPSIS
 --------
 
 **getroute** *id* *msatoshi* *riskfactor* \[*cltv*\] \[*fromid*\]
-\[*fuzzpercent*\] \[*exclude*\] \[*maxhops*\]
+\[*fuzzpercent*\] \[*exclude*\] \[*maxhops*\] \[*maxscannodes*\]
 
 DESCRIPTION
 -----------
@@ -46,6 +46,10 @@ or nodes. Note if the source or destination is excluded, the command result
 is undefined.
 
 *maxhops* is the maximum number of channels to return; default is 20.
+
+*maxscannodes* sets an upper bound the number of nodes examined in route
+traversal, to limit time spend. The default is no limit. It excludes the
+source node.
 
 RISKFACTOR EFFECT ON ROUTING
 ----------------------------
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -877,16 +877,32 @@ static void dijkstra(struct routing_state *rstate,
 		     u64 riskbias,
 		     double fuzz, const struct siphash_seed *base_seed,
 		     struct unvisited *unvisited,
+		     size_t *max_nodes,
 		     costfn_t *costfn)
 {
 	struct node *cur;
+	size_t nodes_limit = *max_nodes;
 
+	if (!(*max_nodes))
+		return;
+
+	/* Include the source node. */
+	(*max_nodes)++;
 	while ((cur = first_unvisited(unvisited)) != NULL) {
 		update_unvisited_neighbors(rstate, cur, me,
 					   riskfactor, riskbias,
 					   fuzz, base_seed, unvisited, costfn);
+
 		remove_unvisited(cur, unvisited, costfn);
-		if (cur == dst)
+
+		if (cur == dst) {
+			status_debug("dijkstra: scan %zu nodes",
+				     nodes_limit + 1 - *max_nodes);
+			return;
+		}
+
+		/* Exclude the source node. */
+		if (!(--(*max_nodes)))
 			return;
 	}
 }
@@ -1043,6 +1059,7 @@ find_shorter_route(const tal_t *ctx, struct routing_state *rstate,
 		   const struct node *me,
 		   struct amount_msat msat,
 		   size_t max_hops,
+		   size_t max_nodes,
 		   double fuzz, const struct siphash_seed *base_seed,
 		   struct chan **long_route,
 		   struct amount_msat *fee)
@@ -1052,6 +1069,7 @@ find_shorter_route(const tal_t *ctx, struct routing_state *rstate,
 	struct amount_msat long_cost, short_cost, cost_diff;
 	u64 min_bias, max_bias;
 	double riskfactor;
+	size_t max_limit = max_nodes;
 
 	/* We traverse backwards, so dst has largest total */
 	if (!amount_msat_sub(&long_cost,
@@ -1075,9 +1093,12 @@ find_shorter_route(const tal_t *ctx, struct routing_state *rstate,
 		     type_to_string(tmpctx, struct node_id, &dst->id),
 		     type_to_string(tmpctx, struct node_id, &src->id));
 	dijkstra(rstate, dst, NULL, riskfactor, 1, fuzz, base_seed,
-		 unvisited, shortest_cost_function);
+		 unvisited, &max_limit, shortest_cost_function);
 	dijkstra_cleanup(unvisited);
 
+	/* Before find_shorter_route, we already find the route under
+	 * the max_scan_nodes limitation. */
+	assert(max_limit);
 	/* This must succeed, since we found a route before */
 	short_route = build_route(ctx, rstate, dst, src, me, riskfactor, 1,
 				  fuzz, base_seed, fee);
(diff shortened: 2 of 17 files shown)


gossipd/routing.c
```diff
@@ -877,17 +877,39 @@ static void dijkstra(struct routing_state *rstate,
 		     u64 riskbias,
 		     double fuzz, const struct siphash_seed *base_seed,
 		     struct unvisited *unvisited,
+		     size_t max_nodes,
+		     bool *nodes_limit,
```
rustyrussell: You're generally best-off expressing this as a "budget".  When it's zero, you stop.  Simplifies the logic somewhat.

gossipd/routing.c
```diff
@@ -877,17 +877,39 @@ static void dijkstra(struct routing_state *rstate,
...
 					   fuzz, base_seed, unvisited, costfn);
+
+		if (*nodes_limit)
+			return;
+
 		remove_unvisited(cur, unvisited, costfn);
-		if (cur == dst)
+
+		if (cur == dst) {
+			status_info("dijkstra: scan %zu nodes", scan_nodes);
```
rustyrussell: Probably needs to go, or at least be a debug!

doc/lightning-getroute.7.md
```diff
@@ -47,6 +47,9 @@ is undefined.
 
 *maxhops* is the maximum number of channels to return; default is 20.
 
+*maxscannodes* is the maximum number of nodes to scan for finding route;
```
rustyrussell: I would express this as `*maxscannodes* sets an upper bound the number of nodes examined in route traversal, to limit time spend.  The default is no limit`.

ZmnSCPxj:
> In other words, maybe gossipd should get out of the routing business? If we can communicate the information fast enough, it might make sense.

You still need *some* entity that manages a view of the network channel graph anyway: the question is *what* is the interface we have with this entity.

Does it only provide channel costs and capacities? Then an external routing plugin will have to sip this data through the JSON-RPC straw, which slows down routing and makes it very bad UX when running on slower RaspPi devices, even with nicer algorithms like Greedy Best First.

Does it have routefinding abilities? Then it has to provide a very generic algorithm with lots of nifty toggles and options, like what we have to add *now*.

Is every routefinding plugin supposed to maintain its own view?  Then the data gets duplicated multiple times in each plugin, increasing total memory costs and hurting lower-capacity RaspPi devices.

Do we just say "sorry we will stop supporting RaspPis"?  Then people are going to use older versions that *do* run on RaspPis and lose out on future security improvements.

I think going for a kitchen-sink `getroute` gives us the best tradeoffs when considering our speed and memory constraints on RaspPis, however much it hurts us to do so.

ZmnSCPxj:
In any case: I fully intend to implement a `getroutequick`, which uses an additional 8 bytes per node as a double-buffered "distance from self" data (one `u32` distance-from-self for each double-buffer).  This is then used in a differential-heuristic guide for A\* and Greedy Best First, which would be quicker than Dijkstra (and hence the reason to call it `getroutequick`).

This "distance from self" data will be updated by a background task, hence the need for double-buffering (`getroutequick` queries use one copy while the background task updates the other copy of the u32 distance-from-self).

Unfortunately it is *not* a drop-in replacement for `getroute`.  In particular the `riskfactor` has to be fixed while the background task is running, and cannot be varied by each `getroutequick` call.  Further, the exact value being sent is mostly ignored by `getroutequick`, as the distance-from-self has to use *some* fixed value-being-sent as guide for the distance-from-self, and we cannot easily rescale it: the value being sent can only be used to determine if a channel is too small to fit the value that needs to be transmitted.

Finally, just after startup, while the distance-from-self data has not been precomputed yet, `getroutequick` can only fail: it has no data to guide it.  We would have to fall back on Dijkstra anyway.

The main advantage of `getroutequick` is that it potentially reduces routing requests by an order of magnitude at the cost of probably getting nonoptimal routes, especially if the "distance from self" data is relatively stale.

You can merge it into `getroute` but again you would add even more options into `getroute` (specifically, A\* is actually a spectrum of algorithms from Greedy Best First on one end to Dijkstra on the other end, with "formal" A\* being the "center" between Greedy Best First and Dijkstra: you need at least one additional option to `getroutequick` to transition between Greedy Best First (g multiplier = 0) and A\* (g multipler = 1)).

Moving it into a plugin separate from the `getroute` implementation would either double your memory utilization (need to also support both `getroutequick` and `getroute`, at least until we can deprecate `getroute` safely) or have `getroutequick` drink through a JSON-RPC straw, losing its speed advantage.

(Alternately have the `getroutequick` plugin have direct access to the routemap, then have `getroute` drink through the JSON-RPC straw --- but note that we have to fall back to `getroute` after startup while the distance-from-self data is not yet filled in, so the increased slowness at this point is worsened: finally, if the current payment amount is very far from the reference amount used by the background to precompute the distance-from-self, then we should really fall back on `getroute` for a more accurate reckoning of routing costs).

gossipd/routing.c
trueptolemy: Good idea!

gossipd/routing.c
trueptolemy: Fixed!

doc/lightning-getroute.7.md
trueptolemy: Good description! :-)

trueptolemy:
It seems now we have 3 views about how to improve the `route` function:
**1. Seperate routing bussiness from `gossipd`, and make a plugin.**
The condition is as @rustyrussell you said:
> If we can communicate the information fast enough, it might make sense.

Will this plugin will "monitoring gossip_store" through the JSON-RPC straw? There may be much frequent read and write. Considering this, I have the same warry as @ZmnSCPxj .

**2. The new command `getroutequick`**
I think this is a bit like what @ZmnSCPxj did in #2980 , and it does increase a lot of complexity. And the performance of this implementation need to be verified.

**3. The plugin implementing `permuteroute` based on `getroute`**
This seems as a compromise between 1 and 2. I also have the warry whether the JSON-RPC straw affects performance.

I think I should put this patch into the `routetricks` plugin patch and look at the performance of that plugin(close to completion now). And then we can do more experiments and decide if it should be accepted.

ZmnSCPxj:
`getroutequick` is different from `permuteroute`.  I describe them, and some other techniques, in a series of lightning-dev posts here: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-August/002095.html

The `getroutequick` description in that initial post is not the final `getroutequick` I am currently **very slowly** making.  I had to retrace many of the well-worn steps that RTS game devs have taken.

In detail:

1.  Precompute a map of each node distance from an arbitrary landmark.
     Interestingly, for a single landmark, it is best to use the own node as the landmark. (for multiple landmarks, it is best to use "corners" of the map far from most routes).
2.  Derive a "differential heuristic" using the above distance-from-landmark.
     This is simply the realization that if we have two nodes A and B, and a landmark L, the shortest possible distance between A and B exists if the path A-to-L has B in it (or the path B-to-L has A in it).  Then the estimated distance from A to B is the difference (hence "differential" heuristic) of the distance between A and L, and the distance between B and L.
3.  Use the differential heuristic above for A\* and Greedy Best First algorithms.
     Greedy Best First only uses the heuristic, and selects the node with the shortest distance to the goal node.
     This leads to very fast routefinding, but often yields suboptimal routes (which is fine, `pay` should check `maxfeepercent` and `maxdelay` anyway, and switch algorithms from Greedy Best First if violated).
     A\* mixes Dijkstra and Greedy Best First; it uses the sum of the heuristic (Greedy Best First judgment point) plus the sum of costs to reach the node (Dijkstra judgment point).
     In fact, A\* is really a spectrum of algorithms, with Greedy Best First at one end and Dijkstra at the other end.

trueptolemy:
Thank you for your explaination! I confused the related concepts.

> Use the differential heuristic above for A* and Greedy Best First algorithms.

I heard that there is another algorithm, [JPS+ with Goal Bounding](https://github.com/SteveRabin/JPSPlusWithGoalBounding). The author of this algorithm said it's over 100x Faster than A*. I have not carefully considered whether it can be applied, but I also recommend it here.

ZmnSCPxj:
I encountered Jump Point Search too, but could not figure out how to translate it in Lightning terms.

JPS works because in many RTS games, the world is a regular grid, and there are large areas with the same traversal costs (i.e. wide open plains).  Lightning network is not a regular grid.  Though there *are* large areas with similar traversal costs.  But the common explanation of JPS involves "go to this direction continuously until you find a blockage", and a Lightning Network has no sense of "direction" (though maybe we could synthesize one?).

JPS need can be avoided in RTS games too, if you use a navigation mesh instead of a grid for each location, with similar speedup (and still using A\*), though again I could not figure out how to translate navigation meshes to Lightning.

trueptolemy:
In favor of #3075 , close this one.

PR #3071, opened 2019-09-19
Update to latest bolts, make gossip_query_ex non-experimental!

rustyrussell:
Done in a few steps, since there have been a number of spec changes / cleanups since we last updated our quotes.

This uses https://github.com/lightningnetwork/lightning-rfc/pull/673 so we can parse the spec properly, but it's not dependent on that.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3071/commits/b553da3910daab41d5ef62281d92bb3499a146b2

PR #3073, opened 2019-09-19
JSON-API: `fundchannel` and `fundchannel_start` use `amount` as fieldname to replace `satoshi`

trueptolemy:
Based to the [comment](https://github.com/ElementsProject/lightning/pull/2964#discussion_r317824584) with @niftynei .
As she said, it's better to rename `satoshi` to `amount`. `satoshi` is a bit confusing that you can specify the `satoshi` field as `2.999btc`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 Note: You should always set `allow-deprecated-apis=false` to test for
 changes.
+- JSON-API: `fundchannel` now uses `amount` as the parameter name to replace `satoshi`
+- JSON-API: `fundchannel_start` now uses `amount` as the parameter name to replace `satoshi`
 
 ### Removed
 
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -499,15 +499,11 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("feerates", payload)
 
-    def fundchannel(self, node_id, satoshi, feerate=None, announce=True, minconf=None, utxos=None):
-        """
-        Fund channel with {id} using {satoshi} satoshis with feerate
-        of {feerate} (uses default feerate if unset).
-        If {announce} is False, don't send channel announcements.
-        Only select outputs with {minconf} confirmations.
-        If {utxos} is specified (as a list of 'txid:vout' strings),
-        fund a channel from these specifics utxos.
-        """
+    def _deprecated_fundchannel(self, node_id, satoshi, feerate=None, announce=True, minconf=None, utxos=None):
+        warnings.warn("fundchannel: the 'satoshi' field is renamed 'amount' : expect removal"
+                      " in Mid-2020",
+                      DeprecationWarning)
+
         payload = {
             "id": node_id,
             "satoshi": satoshi,
@@ -517,6 +513,32 @@ class LightningRpc(UnixDomainSocketRpc):
             "utxos": utxos
         }
         return self.call("fundchannel", payload)
+
+    def fundchannel(self, node_id, *args, **kwargs):
+        """
+        Fund channel with {id} using {amount} satoshis with feerate
+        of {feerate} (uses default feerate if unset).
+        If {announce} is False, don't send channel announcements.
+        Only select outputs with {minconf} confirmations.
+        If {utxos} is specified (as a list of 'txid:vout' strings),
+        fund a channel from these specifics utxos.
+        """
+
+        if 'satoshi' in kwargs:
+            return self._deprecated_fundchannel(node_id, *args, **kwargs)
+
+        def _fundchannel(node_id, amount, feerate=None, announce=True, minconf=None, utxos=None):
+            payload = {
+                "id": node_id,
+                "amount": amount,
+                "feerate": feerate,
+                "announce": announce,
+                "minconf": minconf,
+                "utxos": utxos
+            }
+            return self.call("fundchannel", payload)
+
+        return _fundchannel(node_id, *args, **kwargs)
 
     def fundchannel_start(self, node_id, satoshi, feerate=None, announce=True):
         """
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -540,16 +540,11 @@ class LightningRpc(UnixDomainSocketRpc):
 
         return _fundchannel(node_id, *args, **kwargs)
 
-    def fundchannel_start(self, node_id, satoshi, feerate=None, announce=True):
-        """
-        Start channel funding with {id} for {satoshi} satoshis
-        with feerate of {feerate} (uses default feerate if unset).
-        If {announce} is False, don't send channel announcements.
-        Returns a Bech32 {funding_address} for an external wallet
-        to create a funding transaction for. Requires a call to
-        'fundchannel_complete' to complete channel establishment
-        with peer.
-        """
+    def _deprecated_fundchannel_start(self, node_id, satoshi, feerate=None, announce=True):
+        warnings.warn("fundchannel_start: the 'satoshi' field is renamed 'amount' : expect removal"
+                      " in Mid-2020",
+                      DeprecationWarning)
+
         payload = {
             "id": node_id,
             "satoshi": satoshi,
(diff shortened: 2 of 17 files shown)


contrib/pylightning/lightning/lightning.py
```diff
@@ -507,9 +507,9 @@ def feerates(self, style, urgent=None, normal=None, slow=None):
         }
         return self.call("feerates", payload)
 
-    def fundchannel(self, node_id, satoshi, feerate=None, announce=True, minconf=None, utxos=None):
+    def fundchannel(self, node_id, amount, feerate=None, announce=True, minconf=None, utxos=None):
```
ZmnSCPxj: I am uncertain about python, but would this not break existing Python code that uses this library but explicitly uses `satoshi=` syntax to call `fundchannel` as well?

plugins/fundchannel.c
```diff
@@ -376,17 +376,29 @@ static struct command_result *json_fundchannel(struct command *cmd,
...
+		   p_opt("satoshi", param_string, &satoshi_str),
 		   p_opt("feerate", param_string, &fr->feerate_str),
 		   p_opt_def("announce", param_bool, &fr->announce_channel, true),
 		   p_opt_def("minconf", param_number, &fr->minconf, 1),
 		   p_opt("utxos", param_string, &fr->utxo_str),
 		   NULL))
 		return command_param_failed();
 
+	if (!amount_str) {
+		if (deprecated_apis || !satoshi_str)
```
ZmnSCPxj: I think this logic is wrong?  If amount is not specified, *and* we support deprecated APIs, we fail?  Maybe it should be `!deprecated_apis`?

trueptolemy: Too embarrassed, I confused the meaning of `deprecated_apis`...

trueptolemy: I also find this is not compatible with the case if `params->type == JSMN_ARRAY`.

contrib/pylightning/lightning/lightning.py
trueptolemy: Thank you. I find an example,  `close`:
https://github.com/ElementsProject/lightning/blob/9578fb4035d14690bd2ae8d9788cbb05486dbeaa/contrib/pylightning/lightning/lightning.py#L316-L354

trueptolemy: Rename it directly is reckless.

plugins/fundchannel.c
ZmnSCPxj: I believe a technique we used before would be to use both as `p_opt` if `params->type` is an object and to `p_req` if an array instead.

trueptolemy: I can confirm `p_req(name, cbx, arg)` is used to `Add a required parameter`, and `p_opt(name, cbx, arg)` is used to `Add an optional parameter.  *arg is set to NULL if it isn't found.` (in `common/param.h`). The "req" in `p_req` represents "require"? And "opt" represents "optional"? These make sense. :-)


> I also find this is not compatible with the case if params->type == JSMN_ARRAY.

This is because I find here, also for `close`:
https://github.com/ElementsProject/lightning/blob/9578fb4035d14690bd2ae8d9788cbb05486dbeaa/lightningd/peer_control.c#L1240-L1313

This `else if` block shows if `params->type == JSMN_ARRAY`, `param` will prase the parameters by position.

So for my original code:
```

	if (!param(cmd, buf, params,
		   p_req("id", param_node_id, &fr->id),
		   p_opt("amount", param_string, &amount_str),
                   p_opt("satoshi", param_string, &satoshi_str),
		   p_opt("feerate", param_string, &fr->feerate_str),
		   p_opt_def("announce", param_bool, &fr->announce_channel, true),
		   p_opt_def("minconf", param_number, &fr->minconf, 1),
		   p_opt("utxos", param_string, &fr->utxo_str),
		   NULL))
```

Parsing both `amount` and `satoshi` by position at the same time will lead to the failure.
Now I fix it.

cdecker:
Sorry for all the delayed comments, I had them queued up but forgot to submit. I'll take a closer look tomorrow :-)

trueptolemy:
Thank you! :-) @cdecker

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - JSON API: `fundchannel_cancel` is extended to work before funding broadcast.
 - JSON API: The parameter `exclude` of `getroute` now also support node-id.
 - JSON-API: `pay` can exclude error nodes if the failcode of `sendpay` has the NODE bit set
+- JSON-API: `fundchannel` now uses `amount` as the parameter name to replace `satoshi`
+- JSON-API: `fundchannel_start` now uses `amount` as the parameter name to replace `satoshi`
```
rustyrussell: Maybe move these lines under `## Deprecated`?  I scan those so I know what to remove in future releases...

lightningd/opening_control.c
```diff
@@ -1070,14 +1071,23 @@ static struct command_result *json_fund_channel_start(struct command *cmd,
...
 		return command_param_failed();
 
+	if (!amount) {
+		if (deprecated_apis || !satoshi)
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					    "fundchannel_satrt: Must set 'amount' field");
+		else
+			amount = satoshi;
+	}
+
```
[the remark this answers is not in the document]
rustyrussell: Pretty sure this doesn't work: once param() fails, fc->cmd is freed.  Returning the magic command_param_failed() is just a formality.  We could change that but it's larger surgery.  I think you need to do the horrible hack we did previously in cases like this:

```
if (params->type == JSMN_ARRAY || !deprecated_apis) {
 ... modern code...
} else {
... param(... p_opt("satoshi", param_sat, &amount), p_opt("amount", param_sat, &amount)...
if (!amount) { command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Need amount"); return; }
}
```

Sure, they could specify both and get random results, but we're not here to babysit them entirely!

plugins/fundchannel.c
```diff
@@ -310,7 +310,7 @@ static struct command_result *fundchannel_start(struct command *cmd,
 
 	json_out_start(ret, NULL, '{');
 	json_out_addstr(ret, "id", node_id_to_hexstr(tmpctx, fr->id));
-	json_out_addstr(ret, "satoshi", fr->funding_str);
```
rustyrussell: This still needs to be here, just `if (deprecated_apis) json_out_addstr(ret, "satoshi", fr->funding_str);`

plugins/fundchannel.c
```diff
@@ -379,13 +379,26 @@ static struct command_result *json_fundchannel(struct command *cmd,
 
 	if (!param(cmd, buf, params,
 		   p_req("id", param_node_id, &fr->id),
-		   p_req("satoshi", param_string, &fr->funding_str),
```
rustyrussell: Here too.

plugins/fundchannel.c
```diff
@@ -379,13 +379,26 @@ static struct command_result *json_fundchannel(struct command *cmd,
...
+
+		if (!param(cmd, buf, params,
+			   p_req("id", param_node_id, &fr->id),
+			   p_req("satoshi", param_string, &fr->funding_str),
+			   p_opt("feerate", param_string, &fr->feerate_str),
+			   p_opt_def("announce", param_bool, &fr->announce_channel, true),
+			   p_opt_def("minconf", param_number, &fr->minconf, 1),
+			   p_opt("utxos", param_string, &fr->utxo_str),
+			   NULL))
+			return command_param_failed();
```
rustyrussell: Needs same param trick as above :(

rustyrussell:
You're brave!  Backwards compat is *hard*...

trueptolemy:
When I just thought about changing the name, I didn’t expect to have so much work to do. :frowning:  
Thanks to @rustyrussell 's reminder, I realized I made a big mistake for the parameter prase of `txprepare`. 
And this bug is related with `fundchannel`, so I add the fix in this PR. I'm not sure if @rustyrussell  like the fix of `txprepare`. TBH, I use a direct way to solve it but the code seems not beautiful. :pensive: And there's a disadvantage: `txprepare` can't support `check` command completely.

In addition, I add some test for compact of old-style apis, including `json object` type and `json array` type.

trueptolemy:
Rebased and resolve the conflict.

trueptolemy:
- [ ] Add the test for "leave out optional parameters" (reminded by @niftynei in #3116  )

niftynei:
> And this bug is related with fundchannel, so I add the fix in this PR. I'm not sure if @rustyrussell like the fix of txprepare. TBH, I use a direct way to solve it but the code seems not beautiful. pensive And there's a disadvantage: txprepare can't support check command completely.

In the future, it's better to make a new PR for an 'unrelated' fix :)

Agree that the code isn't exactly beautiful; I think it's safe to leave making it nicer as a future exercise for the next contributor :) keen to get this in so we get closer to the deprecation mark that much sooner (and it can all be deleted anyway)

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3073/commits/7fc49b449812175b4dedce8b7dedaf307eab4bb8

PR #3075, opened 2019-09-20
Mapping gossip store

rustyrussell:
[ Note: finally got back to this after 11 months, below is the rewrite ]

This is a complete implementation of mapping the gossip store for routing, including an optimized implementation of Dijkstra.

It's not *used* anywhere yet; that's next!

Changelog-None


ZmnSCPxj:
Is this basically your response to https://github.com/ElementsProject/lightning/pull/3070#issuecomment-533068023 ?

I believe this brings in the possibility of race conditions, where nodes and channels that are currently being referred to by pointers are subsequently deleted from the store by the `gossipd` thread?  It seems to me that `struct mapped_chan` is a pointer to `mmap`ed data?  What happens in case of a use-free-realloc-use?

Can we at least have some locking to tell `gossipd` "please do not remove any nodes/channels yet"?

At least with `mmap` with read-only shared, the backing real memory is common among all plugins.

It seems this will be primarily of benefit to C plugins, with a much larger effort to get this useable in higher-level languages.

common/map_store.c
```diff
@@ -0,0 +1,504 @@
...
+		= hc->proportional_fee
+		= hc->delay
+		= 0;
+}
+
+static void map_catchup(struct gossip_map *map)
+{
+	size_t reclen;
+
+	/* You can handle unaligned accesses, right? */
```
ZmnSCPxj: The "correct" fix for unaligned access is to have a local variable that you `memcpy` to:

    struct gossip_hdr ghdr_stack;
    memcpy(&ghdr_stack, (void *)(map->map + map->map_end), sizeof(struct gossip_hdr));
    struct gossip_hdr *ghdr = &ghdr_stack;

Of course, determining if unaligned access is allowed is hard.  My usual habit is to always use the above `memcpy` trick.

common/map_store.c
```diff
@@ -0,0 +1,504 @@
...
+	struct htable nodes;
+
+	/* Map of short_channel_id id -> channel */
+	struct chan_htable channels;
+
+	/* Array of nodes, so we can use simple index. */
+	struct mapped_node *node_arr;
+
+	/* Array of chans, so we can use simple index */
+	struct mapped_chan *chan_arr;
```
ZmnSCPxj: Hmmm.  So if I have two plugins accessing the routemap, they will have separate arrays?  Will this not increase the memory footprint?

common/map_store.h
```diff
@@ -0,0 +1,69 @@
...
+
+struct mapped_chan {
+	struct short_channel_id scid;
+	/* two nodes we connect (lesser idx first) */
+	struct half_chan {
+		unsigned int nodeidx;
+		fixedp_t htlc_min, htlc_max;
+
+		/* millisatoshi. */
+		u64 base_fee : 24;
```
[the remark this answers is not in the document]
ZmnSCPxj: It indicates to allocate 24 bits to this field, not a "true" `u64`.  This exists in C99, not certain if it existed in C89 or K&R.

common/map_store.c
rustyrussell: I was thinking they might want completely different representations. The current one are designed for routing as we have now, but even then it needs to be extended.

Plus the overhead is only 10.5 mb, which is pretty good.

common/map_store.h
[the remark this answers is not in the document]
rustyrussell: Definitely C89, but I would have to check K&R. (I'm 90% certain though).

common/map_store.h
```diff
@@ -0,0 +1,69 @@
...
+		/* millionths */
+		u64 proportional_fee : 20;
+		/* Delay for HTLC in blocks. */
+		u64 delay : 20;
+	} half[2];
+};
+
+static inline u64 fixedp_to_u64(fixedp_t val)
+{
+	return (val & ((1 << 11)-1)) << (val >> 11);
```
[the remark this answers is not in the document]
rustyrussell: Note: only because we never add them, simply compare. That's cheap (comparison is literally a normal comparison).

rustyrussell:
> I believe this brings in the possibility of race conditions, where nodes and channels that are currently being referred to by pointers are subsequently deleted from the store by the gossipd thread? It seems to me that struct mapped_chan is a pointer to mmaped data? What happens in case of a use-free-realloc-use?

There is no race. The gossip store is append-only except when we set the bit to delete fields. Even online compaction writes to a new file.

common/map_store.h
```diff
@@ -0,0 +1,69 @@
...
+struct node_id;
+
+/* 5 bit exponent, 11 bit mantissa approximations of min/max */
+typedef u16 fixedp_t;
+
+struct mapped_node {
+	/* Offset in memory map for node details */
+	u32 node_off;
+	u32 num_chans;
+	unsigned int *chan_idxs;
```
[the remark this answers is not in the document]
rustyrussell: Yes, better to use u32 everywhere!

common/map_store.c
rustyrussell: We already have an HAVE_UNALIGNED_ACCESS macro. But there's also the case where we can't mmap and need to read(). The trick I've used in the past is to have a function which returns a direct pointer if it can, or does a copy and returns a pointer to that. Just need to plumb it in...

ZmnSCPxj:
> There is no race. The gossip store is append-only except when we set the bit to delete fields. Even online compaction writes to a new file.

Yes, I misunderstood this at first.
I thought you meant that the routemap is somehow shared across processes by abusing `mmap` on the `gossip_store`, since "everything is a file descriptor" simply means that memory is just a representation of a file, which may or may not be replicated on persistent storage devices.

Some routing algorithms (`getroutequick` in particular) will need a mapping from nodes to data or channels to data.  So `check_gossip_store` needs to  also signal the caller when nodes or channels are deleted, so that those mappings can be freed as well, otherwise the plugin may very well indefinitely retain data for nodes or channels that no longer exist.

ZmnSCPxj:
> Some routing algorithms (`getroutequick` in particular) will need a mapping from nodes to data or channels to data.

We could add a `void *annotation;` to `struct mapped_node` and `struct mapped_chan`.  When a node/channel is constructed, clear this to `NULL`.  When a node/channel is destroyed (put back in the freelist), or the entire map is destroyed, if the field is non-`NULL`, call `tal_free` on it.  The plugin can then `tal` (using a `NULL` parent) any extra annotations it needs per node or channel.

common/map_store.h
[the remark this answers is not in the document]
ZmnSCPxj: Technically this is a floating point, as the mantissa (the lower 11 bits) is "floated" according to an exponent.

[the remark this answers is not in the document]
rustyrussell: Yes, I realized it was misnamed after I hit the PR.  It is indeed 16 bit floating point.

rustyrussell:
> Some routing algorithms (`getroutequick` in particular) will need a mapping from nodes to data or channels to data. So `check_gossip_map` needs to also signal the caller when nodes or channels are deleted, so that those mappings can be freed as well, otherwise the plugin may very well indefinitely retain data for nodes or channels that no longer exist.

Indeed.  See commit 4 "gossipd: add tombstone when we remove a channel."

rustyrussell:
> > Some routing algorithms (`getroutequick` in particular) will need a mapping from nodes to data or channels to data.
> 
> We could add a `void *annotation;` to `struct mapped_node` and `struct mapped_chan`. When a node/channel is constructed, clear this to `NULL`. When a node/channel is destroyed (put back in the freelist), or the entire map is destroyed, if the field is non-`NULL`, call `tal_free` on it. The plugin can then `tal` (using a `NULL` parent) any extra annotations it needs per node or channel.

The reason I didn't do this is that it's 40 bytes of tal overhead per allocation, plus 8 bytes for the pointer (halve these numbers for 32 bit).  It's better for an extension to use the same array approach for their metadata as we use internally; though they can always use an array of void * and simulate this approach if space isn't a concern.

ZmnSCPxj:
> The reason I didn't do this is that it's 40 bytes of tal overhead per allocation, plus 8 bytes for the pointer (halve these numbers for 32 bit). It's better for an extension to use the same array approach for their metadata as we use internally; though they can always use an array of void * and simulate this approach if space isn't a concern.

Using this technique is fine, but what happens if a node / channel is removed and then its array index is reused for a different node / channel?

Also, the public interface in the header uses `struct mapped_node *`/`struct mapped_chan *`, which is not an index into the array and does not contain an index to the array.  So if I have a parallel array of my annotations, then I use `find_node_in_map`, how do I get to my annotation?  There is a `nodeidx` in `mapped_chan` but that is pretty indirect is it not?

`struct gossip_map` is also private inside `map_store.c`, so presumably I cannot rely on this array structure if I am the user of this library.

Using a parallel array seems to move us closer to an entity-component pattern, maybe we should reorganize the map as such instead?

common/gossmap.c
```diff
@@ -0,0 +1,831 @@
...
+			  & ~(GOSSIP_STORE_LEN_DELETED_BIT|
+			      GOSSIP_STORE_LEN_PUSH_BIT))
+			+ sizeof(ghdr);
+
+		if (be32_to_cpu(ghdr.len) & GOSSIP_STORE_LEN_DELETED_BIT)
+			continue;
+
+		/* Partial write, this can happen. */
+		if (map->map_end + reclen > map->map_size)
+			break;
```
cdecker, line 586: This should only ever happen in non-mmap mode though, right? With `mmap` I'd expect the write of `map->map_size` to be atomic and happen after the update succeeded.

common/route.h
```diff
@@ -0,0 +1,49 @@
...
+
+struct dijkstra;
+struct gossmap;
+
+struct route {
+	int dir;
+	struct gossmap_chan *c;
+};
+
+/* Can c carrt amount in dir? */
```
cdecker, line 15:
```suggestion
/* Can c carry amount in dir? */
```

common/route.c
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/* Generally only one side gets marked disabled, but it's disabled. */
+bool route_can_carry(const struct gossmap *map,
+		       const struct gossmap_chan *c,
+		       int dir,
+		       struct amount_msat amount,
+		       void *arg)
+{
+	if (!c->half[dir].enabled || !c->half[!dir].enabled)
```
cdecker, line 35: Still not sure if this isn't too aggressive. I'm thinking in particular of my lnbig channel where the remote end refused to mark my incoming direction as active, despite it having sufficient capacity. It also goes against our previous advice of disabling channel directions if their capacity is unreasonably low and unlikely to carry any non-tiny HTLC. If we include the last ditch attempt that ignores disabled flags if no other path is available we can likely see how well this heuristic works.

I guess we'll have to see how well one heuristic compares to the other.

devtools/clean_topo.h
```diff
@@ -0,0 +1,13 @@
+#ifndef LIGHTNING_DEVTOOLS_CLEAN_TOPO_H
+#define LIGHTNING_DEVTOOLS_CLEAN_TOPO_H
+#include "config.h"
+
+struct gossmap;
+
+/* Cleans topology:
+ * 1. Removes channels not enabled in both dirs.
+ * 2. (if remove_singles) remove nodes with only one connection.
+ * 3. Remove isolated nodes (we assume first node is well-connected!).
```
cdecker, line 10: This is generally called a strongly connected component of the graph (or connected component in an undirected graph) :-)

cdecker (approved):
Very nice optimization, I think I learnt a lot about memory allocators through this. Would it make sense to extract the commonalities between the `chan_arr` and the `node_arr` to deduplicate the code, or are they different enough to make that not suitable?

I'm a bit unclear as to the semantics / intended API: from my understanding `gossipd` would be the only one that is writing to the file, and nodes would regularly refresh their `gossmap` which effectively is an index on top of the file. Consumers are however not allowed to update the file (append or delete) as far as I understand, otherwise we risk having multiple writers race each other.

 - So this is single-writer / multiple-consumer, with `gossipd` as the writer, correct?
 - Since in-place updates are limited to marking things as deleted (and these are tiny and presumably atomic), it also can't happen that we read a partial write, right?

ACK https://github.com/ElementsProject/lightning/pull/3075/commits/4de348d2e94aadf6201b8aed456412eff4695d41

gossipd/routing.c
```diff
@@ -2257,6 +2257,7 @@ void remove_channel_from_store(struct routing_state *rstate,
 	if (is_chan_public(chan)) {
 		update_type = WIRE_CHANNEL_UPDATE;
 		announcment_type = WIRE_CHANNEL_ANNOUNCEMENT;
+		gossip_store_mark_channel_deleted(rstate->gs, &chan->scid);
```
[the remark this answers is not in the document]
rustyrussell: Non-public channels are not in the gossip_store, as such (they have their own non-standard messages, but there's no channel_announcement).

common/route.c
rustyrussell: Yes, my follow-on code has a "ignore disabled channels" fallback mode.

rustyrussell:
> Very nice optimization, I think I learnt a lot about memory allocators through this. Would it make sense to extract the commonalities between the `chan_arr` and the `node_arr` to deduplicate the code, or are they different enough to make that not suitable?

They're different enough, and they're definitely distinct.  The main overlap is when they're used as a linked freelist, and then they're using different fields.

> I'm a bit unclear as to the semantics / intended API: from my understanding `gossipd` would be the only one that is writing to the file, and nodes would regularly refresh their `gossmap` which effectively is an index on top of the file. Consumers are however not allowed to update the file (append or delete) as far as I understand, otherwise we risk having multiple writers race each other.
> 
>     * So this is single-writer / multiple-consumer, with `gossipd` as the writer, correct?

Yes!

>     * Since in-place updates are limited to marking things as deleted (and these are tiny and presumably atomic), it also can't happen that we read a partial write, right?

Over some filesystems we've seen partial writes (which breaks POSIX, but seems to happen).  So we actually handle this in the read code.

> ACK [4de348d](https://github.com/ElementsProject/lightning/commit/4de348d2e94aadf6201b8aed456412eff4695d41)

lightningd/invoice.c
```diff
@@ -1414,6 +1396,7 @@ static struct command_result *json_waitanyinvoice(struct command *cmd,
 				       " is non-trivial.");
 }
 
+
```
rustyrussell, line 1399: It was to avoid line number clash with our autogen code; might be unnecessary at end of series though.

common/gossmap.c
rustyrussell: mmap or no, it can happen in practice :(

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3075/commits/d236f0184a2990b195b835eb70b9f930100d10cc

after @cdecker typo fixes.

PR #3076, opened 2019-09-20
More gossip cleanups

rustyrussell:
A weird valgrind fix, some file movements.

Gossipd now will refuse to even look at channel announcements until they're 6 deep.  As a side-effect, gossipd now knows when we're synced, and won't even ask for gossip before then (to avoid rejecting channels).

Finally, we no longer ask for a flood if the user doesn't support gossip_queries (but they all do anyway).


lightningd/channel_control.c
```diff
@@ -376,9 +376,9 @@ void peer_start_channeld(struct channel *channel,
 
 	if (channel->scid) {
 		scid = *channel->scid;
-		/* Subtle: depth=1 at funding height. */
-		reached_announce_depth = get_block_height(ld->topology) + 1 >=
-				       short_channel_id_blocknum(&scid) + ANNOUNCE_MIN_DEPTH;
+		reached_announce_depth
+			= is_scid_depth_announcable(&scid,
```
darosior: Why not using directly the above pointer instead of a temporary variable ?

rustyrussell: Minimal change, really: the replaced code used &scid too.

rustyrussell:
Rebase onto master...

bitcoin/short_channel_id.h
```diff
@@ -47,6 +48,16 @@ static inline u16 short_channel_id_outnum(const struct short_channel_id *scid)
 	return scid->u64 & 0xFFFF;
 }
 
+/* Subtly, at block N, depth is 1, hence the -1 here. eg. 103x1x0 is announcable
+ * when height is 108. */
+static inline bool
+is_scid_depth_announcable(const struct short_channel_id *scid,
```
cdecker: Did you mean: announceable :-)

common/gossip_constants.h
```diff
@@ -29,15 +25,16 @@
  * The `message_flags` bitfield is used to indicate the presence of optional
  * fields in the `channel_update` message:
  * | Bit Position  | Name                      | Field                |
- * | ------------- | ------------------------- | -------------------- |
+ * | ------------- | ------------------------- | ----------------...- |
```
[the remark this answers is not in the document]
rustyrussell: Actually, there are more `---` in the original. `...` is matched with "anything" when we check the specs.

rustyrussell:
Trivial rebase, @cdecker spelling fix, and added a CHANGELOG.md line.

PR #3089, opened 2019-09-24
pylightning: Fix the parameter `label` of `sendpay` 

trueptolemy:
`description` is renamed as `label` in 0.7.0, so now use `label` as the field name in pylightning.

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -804,17 +804,36 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("plugin", payload)
 
-    def sendpay(self, route, payment_hash, description=None, msatoshi=None):
-        """
-        Send along {route} in return for preimage of {payment_hash}
-        """
+    def _deprecated_sendpay(self, route, payment_hash, description, msatoshi=None):
+        warnings.warn("sendpay: the 'description' field is renamed 'label' : expect removal"
+                      " in early-2020",
+                      DeprecationWarning)
         payload = {
             "route": route,
             "payment_hash": payment_hash,
-            "description": description,
+            "label": description,
             "msatoshi": msatoshi,
         }
         return self.call("sendpay", payload)
+
+    def sendpay(self, route, payment_hash, *args, **kwargs):
+        """
+        Send along {route} in return for preimage of {payment_hash}
+        """
+
+        if 'description' in kwargs:
+            return self._deprecated_sendpay(route, payment_hash, *args, **kwargs)
+
+        def _sendpay(route, payment_hash, label=None, msatoshi=None):
+            payload = {
+                "route": route,
+                "payment_hash": payment_hash,
+                "label": label,
+                "msatoshi": msatoshi,
+            }
+            return self.call("sendpay", payload)
+
+        return _sendpay(route, payment_hash, *args, **kwargs)
 
     def setchannelfee(self, id, base=None, ppm=None):
         """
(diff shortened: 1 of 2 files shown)


contrib/pylightning/lightning/lightning.py
```diff
@@ -804,18 +804,37 @@ def plugin_rescan(self):
...
 
+    def sendpay(self, route, payment_hash, *args, **kwargs):
+        """
+        Send along {route} in return for preimage of {payment_hash}
+        """
+
+        if 'description' in kwargs:
+            return self._deprecated_sendpay(route, payment_hash, *args, **kwargs)
+
+        def _sendpay(route, payment_hash, label=None, msatoshi=None):
```
ZmnSCPxj: I find it strange that this function is a sub-function here, but `_deprecated_sendpay` is a method of the class.  Suggest making `_deprecated_sendpay` into a sub-function here as well.

trueptolemy: Good idea! Now I make it as a sub-function. I also make similar changes in #3073  .

lightningd/pay.c
```diff
@@ -804,17 +804,43 @@ static struct command_result *json_sendpay(struct command *cmd,
...
+		const char *desc = NULL;
+		if (!param(cmd, buffer, params,
+			   p_req("route", param_array, &routetok),
+			   p_req("payment_hash", param_sha256, &rhash),
+			   p_opt("label", param_escaped_string, &label),
+			   p_opt("description", param_escaped_string, &desc),
+			   p_opt("msatoshi", param_msat, &msat),
+			   p_opt("bolt11", param_string, &b11str),
+			   NULL))
+		return command_param_failed();
```
cdecker: Is the indentation of the `return`s off? They seem aligned with the `if` above them. Made me check twice how the 2 lines below aren't unreachable.

contrib/pylightning/lightning/lightning.py
cdecker: I think wrapping this in sub-functions is a bit strange. It's just a single `warning` and an assignment that differs.

contrib/pylightning/lightning/lightning.py
```diff
@@ -804,17 +804,36 @@ def plugin_rescan(self):
...
         """
-        payload = {
-            "route": route,
-            "payment_hash": payment_hash,
-            "description": description,
-            "msatoshi": msatoshi,
-        }
-        return self.call("sendpay", payload)
+
+        def _deprecated_sendpay(route, payment_hash, description=None, msatoshi=None):
```
cdecker: Giving `description` a default value seems strange here, after all we make sure we call it with a description below.

lightningd/pay.c
trueptolemy: This is an alignment error, my carelessness. :-(

contrib/pylightning/lightning/lightning.py
trueptolemy: Yes! I should remove this default.

contrib/pylightning/lightning/lightning.py
trueptolemy: Yeah. Let this function be independent。

lightningd/pay.c
cdecker: No problem, was just really confused :wink:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3089/commits/dbf1d25463601565a19f71f0668bce763cdd45f1

PR #3095, opened 2019-09-24
JAON-API: Add a `txtype` parameter for `txprepare`

trueptolemy:
As I mentioned in #3006 :

> Open Question
> In f77d6c7 (Fix: Store the transaction(broadcast by txsend) into DB), I add the label TX_UNKNOWN
> for the transaction broadcast by txsend.
> I think maybe we can add optional string field for txsend command to specific the label. What do you think about it?

But then I thought we should add the optional string field `txtype` in `txprepare` other than `txsend`.

For now, `txprepare` is used for withdrawal, funding channel or other unknown things. 
So `txtype` should only support these 3 types: "withdrawal", "funding" and "unknown".

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -146,6 +146,32 @@ static struct command_result *broadcast_and_wait(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
+/* Now we only support using `txprepare` for "withdrawal",
+ * "funding" and "unknown". */
+static struct command_result *param_txtype(struct command *cmd,
+					   const char *name,
+					   const char *buffer,
+					   const jsmntok_t *tok,
+					   enum wallet_tx_type **txtype)
+{
+	*txtype = tal(cmd, enum wallet_tx_type);
+	if (json_tok_streq(buffer, tok, "unknown"))
+		**txtype = TX_UNKNOWN;
+	else if (json_tok_streq(buffer, tok, "withdrawal"))
+		**txtype = TX_WALLET_WITHDRAWAL;
+	else if (json_tok_streq(buffer, tok, "funding"))
+		**txtype = TX_CHANNEL_FUNDING;
+	else {
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "'%s' should be 'unknown', 'withdrawal', or "
+				    "'funding', not '%.*s'",
+				    name,
+				    json_tok_full_len(tok),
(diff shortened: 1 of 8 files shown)


wallet/walletrpc.c
```diff
@@ -146,6 +146,32 @@ static struct command_result *broadcast_and_wait(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
+/* Now we only support using `txprepare` for "withdrawal",
+ * "funding" and "unknown". */
+static struct command_result *param_txtype(struct command *cmd,
+					   const char *name,
```
trueptolemy: In fact, the code here is aligned. Maybe it's the display error?

ZmnSCPxj: It has to do with tabs.  Tabs are aligned to 8 positions.  Try typing a line with tab first then nonwhitespace character, then on next line type a nonwhitespace character, then tab, then nonwhitespace character.

You can ignore this for the most part.

ZmnSCPxj: Other than that... I am tempted to just make `txtype` an arbitrary string, just change the db schema instead.  Though I am still not quite up-to-date to the latest db changes.

trueptolemy: Like the `label` for `invoice`?

ZmnSCPxj: not quite, `label` for `invoice` has to be unique per invoice.

trueptolemy: Yeah, it need to replace `enum wallet_tx_type` with a string field.
Should this change deserve another PR?

ZmnSCPxj: Maybe same PR? Depends on what others think though.

trueptolemy: This change seems be related with `listtransactions`(#2696 ). I will continue to study it.

cdecker:
NACK https://github.com/ElementsProject/lightning/pull/3095/commits/15f7066b5bb8ae712057c4ae6db16c9100d57307

This is in direct conflict with #2766 which will remove the transaction-type, and instead use input and output annotations. That would then allow us to have transactions that are multi-purpose and have them annotated accordingly. For example a splice operation consists of a close and a reopen in the same transaction, which cannot be represented with the current schema.

Having to specify a type for each input and output when using `txprepare` and friends is cumbersome, so I'd leave them unannotated.

trueptolemy:
Thank you!

PR #3096, opened 2019-09-24
pylightning: Responses may not be dict

jtimon:
Sometimes when calling getinfo before the node has started I get an int (concretely 2019) instead of a dict in resp, resulting in a TypeError when ```"error" in resp``` is checked, for ints are not iterable in that way.

Not a big deal, but I think a ValueError like when no "result" field is present would be more appropriate and clear.

### DIFF
--- a/contrib/pylightning/lightning/lightning.py
+++ b/contrib/pylightning/lightning/lightning.py
@@ -229,7 +229,9 @@ class UnixDomainSocketRpc(object):
         sock.close()
 
         self.logger.debug("Received response for %s call: %r", method, resp)
-        if "error" in resp:
+        if not isinstance(resp, dict):
+            raise ValueError("Malformed response, response is not a dictionary %s." % resp)
+        elif "error" in resp:
             raise RpcError(method, payload, resp['error'])
         elif "result" not in resp:
             raise ValueError("Malformed response, \"result\" missing.")


rustyrussell:
Any idea *why* you're getting "2019"?  Used-by date? :)

jtimon:
No, not sure why I'm getting the 2019.
Can anyone reproduce this?

cdecker:
NACK

This seems more like papering over a real issue rather than addressing it. Need to look into why we return the best-by-date :wink:

jtimon:
> Address the cause, not the symptoms.

Sure, I mean, why not both?
Even if the cause is fixed, this should make things clearer for next time something similar happens.
But yeah, we should probably look into what's exactly happening in those cases.

cdecker:
> > Address the cause, not the symptoms.
> 
> Sure, I mean, why not both?
> Even if the cause is fixed, this should make things clearer for next time something similar happens.
> But yeah, we should probably look into what's exactly happening in those cases.

Hm, you're probably right, we should make pylightning stable in case it gets paired with an old daemon that still behaves badly :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/commit/0177e4de78e9f35b9704dc7c6c3125bf90b20398

PR #3098, opened 2019-09-25
Three more gossip fixes

rustyrussell:
More prep work for my "gossip seeker" series.


gossipd/routing.c
```diff
@@ -233,16 +234,55 @@ static void memleak_help_routing_tables(struct htable *memtable,
...
+
+static bool in_txout_failures(struct routing_state *rstate,
+			      const struct short_channel_id *scid)
+{
+	if (uintmap_get(&rstate->txout_failures, scid->u64))
+		return true;
+
+	/* If we were going to expire it, we no longer are. */
+	if (uintmap_get(&rstate->txout_failures_old, scid->u64)) {
+		add_to_txout_failures(rstate, scid);
```
cdecker: Great little detail :+1:

Should we also remove it from the `txout_failures_old` uintmap? Otherwise we could end up with 20k entries (10k + 10k) in the worst scenario, since the `num_txout_failures` doesn't get incremented.

rustyrussell: num_txout_failures is incremented in add_to_txout_failures, so we end up with 10k each.  I think deleting from the old map is probably more work than it's worth though.

Note that glibc overhead is 1 ptr per allocation, and the intmap is a binary tree which works out to about one internal node and one leaf node per value, and each node is two pointers.  Thus 10,000 allocations costs us about 480k (240k for 32-bit).  1MB seems a reasonable cache size in these modern times.

Note also that a htable removes many of these inefficiencies: it we init a htable for 10k we get 16k entries, costing 128k (64k on 32-bit).  Two of these is still 256k,  and far more cache-friendly to boot.  Main cost is probably siphash24, we could use a weaker hash TBH since scids are hard to generate.

rustyrussell:
Trivial rebase onto master...

PR #3101, opened 2019-09-25
Set the rlimit for NOFILE to 1024 to reduce the subprocess cleanup

### LINKED ISSUE #2977: Anormal CPU consumption

### LINKED ISSUE #3074: Missing libraries in the Docker image
I tried to build a docker image of the repository locally on my mac but at runtime it looks like some libraries are missing (or have not linked properly during building).

Namely: 
1. libgmp.so.10
2. sqlite

## Local environment information
macOS 10.14.16 (18G95)
Docker version 19.03.1, build 74b1e89

Steps to 
```bash
git clone https://github.com/ElementsProject/lightning.git
cd lightning
docker build -t lightning .
docker run -ti lightning
```
### Output

```
lightningd: error while loading shared libraries: libgmp.so.10: cannot open shared object file: No such file or directory
```

cdecker:
This should limit the overhead introduced by 25ddb8082369451024f471bc2d77fa7e7207cc47, specifically the loop calling syscalls here:

https://github.com/ElementsProject/lightning/blob/25ddb8082369451024f471bc2d77fa7e7207cc47/ccan/ccan/pipecmd/pipecmd.c#L141-L145

In some environments this would be way larger (e.g., running as root in docker), resulting in way more `close()` calls than intended. This PR limits the number of concurrent open fds, so we have to check and close a smaller range.

Ping @NicolasDorier 

Fixes #2977 
Fixes #3074

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -76,6 +76,7 @@
 #include <lightningd/options.h>
 #include <onchaind/onchain_wire.h>
 #include <signal.h>
+#include <sys/resource.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
@@ -629,6 +630,11 @@ int main(int argc, char *argv[])
 	struct timers *timers;
 	const char *stop_response;
 	struct htlc_in_map *unprocessed_htlcs;
+	struct rlimit nofile = {1024, 1024};
+
+	/*~ Make sure that we limit ourselves to something reasonable. Modesty
+	 *  is a virtue. */
+	setrlimit(RLIMIT_NOFILE, &nofile);
 
 	/*~ What happens in strange locales should stay there. */
 	setup_locale();
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,29 +48,26 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
 FROM debian:stretch-slim as builder
 
 ENV LIGHTNINGD_VERSION=master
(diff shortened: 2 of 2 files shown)


NicolasDorier:
cool thanks, will release a new version of current release with this patch today,

lightningd/lightningd.c
```diff
@@ -629,6 +630,11 @@ int main(int argc, char *argv[])
 	struct timers *timers;
 	const char *stop_response;
 	struct htlc_in_map *unprocessed_htlcs;
+	struct rlimit nofile = {1024, 1024};
+
+	/*~ Make sure that we limit ourselves to something reasonable. Modesty
+	 *  is a virtue. */
+	setrlimit(RLIMIT_NOFILE, &nofile);
```
ZmnSCPxj: Would it be moderately useful to have this as a configurable option?  Or do we need to do this `setrlimit` ASAP before any command-line processing?

cdecker: No we can make it configurable, but for now this limits us to ~1000 peers, which should be an ok guess for now.

For non-root users we're unlikely to even have a higher limit since it requires system configuration changes. If this becomes an issue we can definitely introduce the option.

PR #3102, opened 2019-09-26
Revert "docker: Use system libraries, don't bring your own"

NicolasDorier:
This was working up to the latest release because it was statically linked to the c-lightning binary.
Also, this does not fix arm32 and arm64.

The reason why I was "building my own" was because of docker cache layers, you might end up in a situation where you compiled on the build image on a different version of dependency that what end up installed in the runtime image. This happened to me already.

The cause of https://github.com/ElementsProject/lightning/issues/3074 comes from somewhere else.
Probably a change in the build system which does not build statically anymore?

Changelog-None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,26 +48,29 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
 FROM debian:stretch-slim as builder
 
 ENV LIGHTNINGD_VERSION=master
-RUN apt-get update && apt-get install -y --no-install-recommends \
-    ca-certificates \
-    autoconf \
-    automake \
-    build-essential \
-    git \
-    libtool \
-    python \
-    python3 \
-    python3-mako \
-    wget \
-    gnupg \
-    dirmngr \
-    git \
-    gettext \
-    unzip \
-    tclsh \
-    libsqlite3-dev \
-    libgmp-dev \
-    zlib1g-dev
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates autoconf automake build-essential git libtool python python3 python3-mako wget gnupg dirmngr git gettext
+
+RUN wget -q https://zlib.net/zlib-1.2.11.tar.gz \
+&& tar xvf zlib-1.2.11.tar.gz \
+&& cd zlib-1.2.11 \
+&& ./configure \
+&& make \
+&& make install && cd .. && rm zlib-1.2.11.tar.gz && rm -rf zlib-1.2.11
+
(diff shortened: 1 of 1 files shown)


cdecker:
Well the docker image was broken when I tested it in amd64 and i386 from my ubuntu machine, so we needed a fix for that.

I do not think that caching layers is a good reason to overcomplicate the `Dockerfile`, especially since all automated docker repositories will skip the cache.

NACK https://github.com/ElementsProject/lightning/pull/3102/commits/a03102f1c64dd2ae436403286847a361893e9f60

NicolasDorier:
@cdecker the amd64 was working during the last release. So I think something broke in the build scripts. Regardless whether we merge this or not, things were statically linked before (last release) but not after (master), any idea what happened?

NicolasDorier:
Notice I use ` --enable-static`, that you should have removed with the other commit if you want to use system libs.

NicolasDorier:
Just another reason to revert the commit: arm32 and arm64 still using static.

NicolasDorier:
I don't think this should be closed.

NicolasDorier:
It does not have to be in 0.7.3 though.

cdecker:
@NicolasDorier maybe we should instead open an issue to track a workable solution, rather than discussing on this specific solution?

NicolasDorier:
So basically I understand why you passed from static lib to dynamic lib, but I think this should not be done.

* You want build to be reproducible so we don't waste time searching bugs caused by the same dockerfile but different library version.
* You can clearly document which version of deps for c-ligthning
* Other images are also statically linked

I don't think the complexity in dockerfile is too bad. This is almost never updated code. The reason you updated it was because static linking build was broken, which mean that this was a pretty effective "unit test".

rustyrussell:
So, @NicolasDorier you're saying that static builds should *really* be fully static?  I must admit I was a bit surprised that they're not (they just build our own sub-library statically).

I'm happy to make --enable-static do that, and have --enable-static-bundled do what we have now.

NicolasDorier:
`--enable-static` works fine now, this is why I want to revert changes to use it in the dockerfile for amd64. The change back to dynamic was because `--enable-static` was broken before you fixed it.

`--enable-static-bundled` I don't understand how different it would be from `--enable-static`?

cdecker:
Rebased and tested on top of `master`. @NicolasDorier was right that the static build was borked.

ACK https://github.com/ElementsProject/lightning/pull/3102/commits/7c78daa91444129aae99459fba82b9ec1c86ffb6

PR #3103, opened 2019-09-26
Dedicated flag to decrease prune timeout

rustyrussell:
I tried increasing the prune interval, but we still had issues with --dev-fast-gossip reducing the gossip timeout causing unexpected "bad gossip" messages.

We can't make it too big, or our prune test will simply time out, so return to the original version where we had a setting for this test specifically.

### DIFF
--- a/common/gossip_constants.h
+++ b/common/gossip_constants.h
@@ -63,16 +63,16 @@
  *     - MAY prune the channel.
  *     - MAY ignore the channel.
  */
-#define GOSSIP_PRUNE_INTERVAL(dev_fast_gossip_flag) \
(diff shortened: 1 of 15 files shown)


cdecker:
Probably needs a squash commit, since otherwise rebaseability is broken. Otherwise all good :-)

ACK https://github.com/ElementsProject/lightning/pull/3103/commits/b727f7dc4ae246063c5db4664d853c61a2408c59

rustyrussell:
Actually, valgrind issue was pre-existing, and I only noticed it when I tried making the default prune timout v. v. long, so it's safe.

PR #3104, opened 2019-09-26
Make option_static_remotekey non-EXPERIMENTAL now it's in spec.

rustyrussell:
https://github.com/lightningnetwork/lightning-rfc/pull/642


cdecker:
Seems the bolt quote is not matching the bolts. Other than that LGTM :-)

PR #3111, opened 2019-09-28
Allow --announce-addr to work also with autotor: prefix

Saibato:
Make --announce-addr with autotor: also
a meaningful use case.
The option  --announce-addr=autotor: is more
intuitive than to use the --addr=autotor: option

Signed-off-by: Saibato <saibato.naga@pm.me>

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -44,6 +44,11 @@
 bool deprecated_apis = true;
 static bool opt_table_alloced = false;
 
+/* Declare opt_add_addr here, because we we call opt_add_addr
+ * and opt_announce_addr vice versa
+*/
+static char *opt_add_addr(const char *arg, struct lightningd *ld);
+
 /* Tal wrappers for opt. */
 static void *opt_allocfn(size_t size)
 {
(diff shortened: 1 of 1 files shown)


Saibato:
> Make --announce-addr with autotor: also
> a meaningful use case.

This is for sure just an opinion, this PR has some nack potential as that is my biased view and we can certain live without it :smile: .

lightningd/options.c
```diff
@@ -163,7 +163,14 @@ static char *opt_add_announce_addr(const char *arg, struct lightningd *ld)
 {
 	const struct wireaddr *wn;
 	size_t n = tal_count(ld->proposed_wireaddr);
-	char *err = opt_add_addr_withtype(arg, ld, ADDR_ANNOUNCE, false);
+	char *err;
+
+	/* Check for autotor and reroute the call to --addr  */
+	if (strstarts(arg, "autotor:"))	{
```
ZmnSCPxj: Nitpick, for `if` with one statement we omit `{ }`.

Saibato: Done :-) and rebased 

The rebase to current master needs the opt-add-addr declaration to be on top of source so 
please review again.

ZmnSCPxj: Looks OK, suggest squashing into one commit.

Saibato: Done :-)

ZmnSCPxj:
ACK 8696738

PR #3115, opened 2019-09-29
Add devtools/mkpoint

Sjors:
Tool to dump a channel commitment point, useful for recovering funds.

It takes the following arguments:
* node id (the public key of the other peer)
* channel database id (see `chan #` entries in log)
* commit_num: 0 for a freshly opened channel

For some reason hsmd.c doesn't have a header file, so I'm confused how to properly call its functions.

Background: https://bitcoin.stackexchange.com/questions/90196/how-to-recover-funds-from-a-force-closed-lnd-lightning-channel

### DIFF
--- a/devtools/.gitignore
+++ b/devtools/.gitignore
@@ -8,3 +8,4 @@ mkfunding
 mkclose
 mkgossip
 mkencoded
+mkpoint
--- a/devtools/Makefile
+++ b/devtools/Makefile
@@ -1,6 +1,6 @@
 DEVTOOLS_SRC := devtools/gen_print_wire.c devtools/gen_print_onion_wire.c devtools/print_wire.c
 DEVTOOLS_OBJS := $(DEVTOOLS_SRC:.c=.o)
-DEVTOOLS := devtools/bolt11-cli devtools/decodemsg devtools/onion devtools/dump-gossipstore devtools/gossipwith devtools/create-gossipstore devtools/mkcommit devtools/mkfunding devtools/mkclose devtools/mkgossip devtools/mkencoded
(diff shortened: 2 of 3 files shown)


Sjors:
From re-reading IRC I suspect I also need to mix in hsm_secret somewhere in this process.

darosior:
> For some reason hsmd.c doesn't have a header file, so I'm confused how to properly call its functions.

You can use it as a subdaemon (as `lightningd` does) and use the autogenerated wire functions to request sigs. See `lightningd/hsm_control.h`.



> From re-reading IRC I suspect I also need to mix in hsm_secret somewhere in this process.

Yep, it contains the bip32 master key, but it's handled by hsmd.

Sjors:
So what would be the right approach to dump a per-commitment point? Should I make `devtools/mkpoint` spin up an `hsmd` daemon and then talk to it? Or is it better to add a developer RPC call? Or a plugin?

darosior:
I'd go for spinning up hsmd as a subdaemon, however that's the path I know the less about so I might be wrong.
A plugin would need a hook or notification, I don't think that fits your need here.
A dev RPC call might be a good solution too if the channel is still active, to use when `lightningd` is running. However, as I think I read on IRC, this was to recover the points directly from the database file ? In this case an independent tool seems more appropriate.

Sjors:
The channel I need this for is long closed and pruned from listfunds.

darosior:
@Sjors I wonder for the same reasons I wonder moving @rustyrussell `signmessage` and `checkmessage` to `hsmtools`..
I'm now anyway giving a look to cleanup this ;-).

darosior:
@Sjors I wonder for the same reasons I wonder moving @rustyrussell `signmessage` and `checkmessage` to `hsmtools` (cf #3177)..

Sjors:
Closing in favor of #3186 which introduces the equivalent: `tools/hsmtools dumpcommitments NODE_PUBKEY DB_INDEX COMMITMENT_INDEX`

PR #3117, opened 2019-09-30
Much smarter gossip probing.

rustyrussell:
This probably needs more test tweaking (our gossip is less predictable than it was), but also some real-life testing: we want to see how it handles actual peers on the network, and I'll be doing that this week.


gossipd/routing.c
```diff
@@ -45,19 +45,24 @@ struct pending_node_announce {
 #define TOKENS_PER_MSG 24
 #define TOKEN_MAX (24 * 4)
 
-static bool ratelimit(const struct routing_state *rstate,
-		      u8 *tokens, u32 prev_timestamp, u32 new_timestamp)
+static u8 update_tokens(const struct routing_state *rstate,
+			u8 tokens, u32 prev_timestamp, u32 new_timestamp)
 {
-	u64 num_tokens;
+	u64 num_tokens = tokens;
```
niftynei: what is a token here?

gossipd/seeker.c
```diff
@@ -0,0 +1,991 @@
...
+
+/* Get a random peer, but try our preferred peer first, if any.  This
+ * biasses us to the peer that told us of unexpected gossip. */
+static struct peer *random_seeker(struct seeker *seeker,
+				  bool (*check_peer)(const struct peer *peer))
+{
+	struct peer *peer = seeker->preferred_peer_softref;
+
+	/* 80% chance of just choosing a peer who reported the missing stuff */
+	if (peer && check_peer(peer) && pseudorand(5) != 0) {
```
niftynei: where do these 80% odds come from? not sure i follow the connection with the preferred peer selection.

if there's 3 peers, wouldn't the odds be closer to 66%, or are we weighting more heavily towards it since there's a good chance it was inn operation at the time?

rustyrussell:
> nice change set. seems like there's a decent number of assumptions made throughout, such as 10m for gossip to propagate. it'd be nice to get some harder data for those limits.

Yes, I tried to make sure that if I'm wrong, they just waste a bit of resources or we end up recovering by probing.  When it was too short, we found that we sometimes missed gossip when rotating gossiping peers.

> how's the real-world testing looking?

I am making minor fixes now to address review comments and that valgrind error (I mistakenly thought ts was initialized to 0, it's not).

Then I'll load this on my nodes and start more thorough RW testing...

common/utils.c
```diff
@@ -15,6 +15,60 @@ bool is_elements(const struct chainparams *chainparams)
...
+	assert((char *)ptr >= (char *)outer);
+	assert((char *)ptr < (char *)outer + outersize);
+
+	/* This is harmless if there was no prior, otherwise constrains the
+	 * leak: we don't have enough information in softref_nullify to
+	 * clear softref_cleanup */
+	tal_del_destructor2(outer, softref_cleanup, ptr);
+
+	if (obj) {
+		tal_add_destructor2(outer, softref_cleanup, ptr);
```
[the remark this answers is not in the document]
rustyrussell: No, del vs add....

common/utils.c
```diff
@@ -15,6 +15,60 @@ bool is_elements(const struct chainparams *chainparams)
...
+		tal_add_destructor2(obj, softref_nullify, ptr);
+#if DEVELOPER
+		tal_add_notifier(outer, TAL_NOTIFY_MOVE, dont_move_softref);
```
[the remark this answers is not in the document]
rustyrussell: Hmm, we also probably want to remove these rather than have them build up!  Will fix.

gossipd/routing.c
rustyrussell: It's an exciting new way to raise capital!

Wait, no, it's https://en.wikipedia.org/wiki/Token_bucket

gossipd/seeker.c
rustyrussell: Added comment:

```C

	/* 80% chance of immediately choosing a peer who reported the missing
	 * stuff: they presumably can tell us more about it.  We don't
	 * *always* choose it because it could be simply spamming us with
	 * invalid announcements to get chosen, and we don't handle that case
	 * well yet. */
```

rustyrussell:
Some issues in RL testing.  Analyzing now.

rustyrussell:
That was a longer diversion than I'd hoped!  But now working well...

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3117/commits/57eccc97f3b6133dc0e3e042e775ff1e1c1922fe

PR #3123, opened 2019-10-01
query_messages: fail the connection if the peer is not on the same network

darosior:
This PR : 
- MUST make `handle_query_short_channel_ids` respond with `complete` flag set to `0` when we are queried for an unknown `chain_hash`.
- MAY make `handle_reply_channel_range` and `handle_reply_short_channel_ids_end` fail the connection if the peer says it does not have up to date informations about our network. (related to https://github.com/lightningnetwork/lightning-rfc/pull/678)

### DIFF
--- a/gossipd/queries.c
+++ b/gossipd/queries.c
@@ -678,6 +678,10 @@ const u8 *handle_reply_channel_range(struct peer *peer, const u8 *msg)
 
 	scids = decode_short_ids(tmpctx, encoded);
 	if (!scids) {
+		if (complete == 0)
+			return towire_errorfmt(peer, NULL,
+			                       "No up to date infos about this network: %s",
+			                       tal_hex(tmpctx, msg));
 		return towire_errorfmt(peer, NULL,
 				       "Bad reply_channel_range encoding %s",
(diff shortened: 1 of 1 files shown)


gossipd/queries.c
```diff
@@ -666,6 +678,10 @@ const u8 *handle_reply_channel_range(struct peer *peer, const u8 *msg)
 
 	scids = decode_short_ids(tmpctx, encoded);
 	if (!scids) {
+		if (complete == 0)
+			return towire_errorfmt(peer, NULL,
+			                       "No up to date infos about this network: %s",
+			                       tal_hex(tmpctx, msg));
```
cdecker: Should we maybe add the chain hash instead of the entire message they sent us?

darosior: We send the whole message for all errors, I can make it `chain_hash` instead but that's anyway not very useful to the peer (he kind of errored first)

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3123/commits/d78d888c2d284660f05134c572f6251785e866f2

PR #3124, opened 2019-10-02
[dev-suppress-gossip] - Set new peers to GOSSIP_NONE with flag enabled

willcl-ark:
After the dev-suppress-gossip flag has been set via RPC this PR forces gossipd to set gossip level to GOSSIP_NONE for new peers we connect to.

Currently setting this flag does not appear to suppress gossip.

Signed-off-by: willcl-ark <will8clark@gmail.com>

### DIFF
--- a/gossipd/gossipd.c
+++ b/gossipd/gossipd.c
@@ -561,6 +561,12 @@ static enum gossip_level peer_gossip_level(const struct daemon *daemon,
 	if (!gossip_queries_feature)
 		return GOSSIP_HIGH;
 
+#if DEVELOPER
+	/* Don't ask new peers for new gossip is dev-suppress-gossip has been set*/
+	if (suppress_gossip)
+	    return GOSSIP_NONE;
+#endif
+
 	/* Figure out how many we have at each level. */
 	memset(gossip_levels, 0, sizeof(gossip_levels));
 	list_for_each(&daemon->peers, peer, list)


rustyrussell:
Can we just remove the option? What are you actually trying to do?

willcl-ark:
@rustyrussell

I would like it if, when dev-suppress-gossip is activated, the node neither requested new gossip updates, nor responded to requests for gossip updates from other nodes.

Rationale: I am trying to stabilise lightning payments over mesh net (using goTenna). Our bandwidth is approx. 1KB/minute, so gossip updates of even a few MB is a killer for me. Current startup is about 2MB of updates upon channel re-activation...

My current model is: someone has already (with a fully online node) synced the network graph, and already opened a channel to a reliable peer. When they are then offline, with their partially outdated graph, they can completely disable gossip and attempt payments using their current view of the network. 

Does my patch achieve this?

rustyrussell:
OK, I'm (separately) reworking gossiping entirely in another PR, which (while much more efficient!) still won't be enough for this use case.

Do you need this control static or dynamic?  We could have a 'gossip-only' command which specifies 0 or more peers to gossip with, and it will never gossip with any other peer.  But you might also want a 'suppress-gossip' startup option?

I'll apply this for now, but obviously you'd like a longer-term solution which works out of dev mode?

willcl-ark:
Hi @rustyrussell, sorry in advance for the long response!

To answer your questions, I think that all I would be looking for in the near future, and to keep things as simple as possible, would be a command line flag that lightningd could be started with (`suppress-gossip`?) which would disable gossip for _all_ peers until next restart.

This would give our users the workflow:

1) Run `lightningd` as normal, open a channel, gossip with peers, form a network graph etc.
1) Move to a mesh-only  environment: no internet connection, blockstream blocksat downlink.
1) Run `lightningd --suppress-gossip`. lightningd will then not request gossip from any peers, nor respond to gossip requests.
1) Perhaps polite to respond to requests for gossip from peers with the "I don't have gossip for that timestamp_filter" message, if simply ignoring them would be considered rude.

In the longer term, I think it might be interesting to store, per peer, if they have previously requested to *not* receive gossip, then upon channel re-establishment we could skip sending gossip requests and timestamp_filters between those peers altogether, until one peer again requests gossip. However this (I imagine) would be a fair bit more work than the simple flag...

Finally, this commit is working really well for us right now, and will be great to have in 0.7.3 proper, but I think I have not gone far enough with it: unless I enable it on _both_ nodes I don't get the behaviour I would like, which is OK. But, I would love to not need to rely on the channel partner to also be running a version of C-Lightning compiled with the --enable-developer flag, and then to have called `Rpc.dev_suppress_gossip`. Do you have any pointers on how I can stop replying to peer gossip requests? now that bits have moved into queries.c I can't read the (suppress_gossip) flag...

PR #3129, opened 2019-10-03
hsm_secret encryption

darosior:
This adds a new startup option which allows one to encrypt its `hsm_secret`, bearer of all its secrets. This uses the already-included (and amazing!) [`libsodium`](https://download.libsodium.org/doc/) with defaults (i.e. [Argon2id](https://github.com/p-h-c/phc-winner-argon2) for deriving a key out of the provided password, and [XChacha20_Poly1305](https://tools.ietf.org/html/draft-arciszewski-xchacha-03) for stream encryption).

This and https://github.com/ElementsProject/lightning/pull/2925 put, I think, the basis to have wallet locking. A plugin using the `rpc_command` hook to restrict the use of some RPC commands against a password, which can easily be bypassed if [it / `lightningd`] is killed, but not with an encrypted `hsm_secret`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ You can start `lightningd` with the following command:
 lightningd --network=bitcoin --log-level=debug
 ```
 
-This creates a `.lightning/` subdirectory in your home directory: see `man -l doc/lightningd.8`.
+This creates a `.lightning/` subdirectory in your home directory: see `man -l doc/lightningd.8` (or https://lightning.readthedocs.io/) for more runtime options.
 
 ### Using The JSON-RPC Interface
 
@@ -101,6 +101,8 @@ Useful commands:
 Once you've started for the first time, there's a script called
 `contrib/bootstrap-node.sh` which will connect you to other nodes on
 the lightning network.
+
+You can encrypt the BIP32 root seed (what is stored in `hsm_secret`) by passing the `--encrypted-hsm` startup argument. You can start `lightningd` with `--encrypted-hsm` on an already existing `lightning-dir` (with a not encrypted `hsm_secret`). If you pass that option, you __will not__ be able to start `lightningd` (with the same wallet) again without the password, so please beware with your password management. Also beware of not feeling too safe with an encrypted `hsm_secret`: unlike for `bitcoind` where the wallet encryption can restrict the usage of some RPC command, `lightningd` always need to access keys from the wallet which is thus __not locked__ (yet), even with an encrypted BIP32 master seed.
 
 There are also numerous plugins available for c-lightning which add
 capabilities: in particular there's a collection at:
--- a/common/daemon.c
+++ b/common/daemon.c
@@ -11,6 +11,7 @@
 #include <common/utils.h>
 #include <common/version.h>
 #include <signal.h>
+#include <sodium.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/types.h>
@@ -145,6 +146,12 @@ void daemon_setup(const char *argv0,
 	if (getenv("LIGHTNINGD_DEV_MEMLEAK"))
 		memleak_init();
 #endif
+
+	/* We rely on libsodium for some of the crypto stuff, so we'd better
+	 * not start if it cannot do its job correctly. */
+	if (sodium_init() == -1)
+		errx(1, "Could not initialize libsodium. Maybe not enough entropy"
+		         " available ?");
 
 	/* We handle write returning errors! */
 	signal(SIGPIPE, SIG_IGN);
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -580,6 +580,7 @@ static struct io_plan *init_hsm(struct io_conn *conn,
 	struct secret *seed;
 	struct secrets *secrets;
 	struct sha256 *shaseed;
+	struct secret *hsm_encryption_key;
 
 	/* This must be lightningd. */
 	assert(is_lightningd(c));
@@ -589,7 +590,7 @@ static struct io_plan *init_hsm(struct io_conn *conn,
 	 * an extension of the simple comma-separated format output by the
 	 * BOLT tools/extract-formats.py tool. */
 	if (!fromwire_hsm_init(NULL, msg_in, &bip32_key_version, &chainparams,
-			       &privkey, &seed, &secrets, &shaseed))
+	                       &hsm_encryption_key, &privkey, &seed, &secrets, &shaseed))
 		return bad_req(conn, c, msg_in);
 
 #if DEVELOPER
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -45,7 +45,7 @@
 #include <hsmd/gen_hsm_wire.h>
 #include <inttypes.h>
 #include <secp256k1_ecdh.h>
-#include <sodium/randombytes.h>
+#include <sodium.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
 #include <sys/types.h>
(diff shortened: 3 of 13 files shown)


darosior:
Rebased and added handling for non-encrypted to encrypted hsm. Still need to polish the tests and add the doc.

darosior:
Rebased, added tests and doc, edited OP. Removing WIP.

darosior:
Rebased, increase `sleep()`ing time for the test to pass on Travis.

hsmd/hsmd.c
```diff
@@ -554,15 +598,66 @@ static void maybe_create_new_hsm(void)
...
+		if (crypto_secretstream_xchacha20poly1305_pull(&crypto_state,
+		                                               secretstuff.hsm_secret.data,
+		                                               NULL, 0, cipher, sizeof(cipher),
+		                                               NULL, 0) != 0) {
+			/* Exit but don't throw a backtrace when the user made a mistake in typing
+			 * its password. Instead exit and `lightningd` will be able to give
+			 * an error message. */
+			exit(1);
+		}
+	}
```
rustyrussell: ... else?

If it's encrypted, and they don't provide the key?  [Edit: ah, fixed in next commit!] 

Or if it's not, and they do?  Or if it's been truncated to 31 bytes?

hsmd/hsmd.c
```diff
@@ -603,8 +707,15 @@ static struct io_plan *init_hsm(struct io_conn *conn,
...
-	maybe_create_new_hsm();
-	load_hsm();
+	maybe_create_new_hsm(hsm_encryption_key, true);
+	load_hsm(hsm_encryption_key);
+
+	/*~ We don't need the hsm_secret encryption key anymore.
+	 * Note that sodium_munlock() also zeroes the memory. */
+	if (hsm_encryption_key) {
+		tal_free(hsm_encryption_key);
+		sodium_munlock(hsm_encryption_key, sizeof(hsm_encryption_key));
```
rustyrussell: sizeof(*hsm_encryption_key)?

lightningd/lightningd.c
```diff
@@ -657,6 +658,12 @@ int main(int argc, char *argv[])
 	 * backtraces when we crash (if supported on this platform). */
 	daemon_setup(argv[0], log_backtrace_print, log_backtrace_exit);
 
+	/*~ We rely on libsodium for some of the crypto stuff, so we'd better
+	 * not start if it cannot do its job correctly. */
+	if (sodium_init() == -1)
```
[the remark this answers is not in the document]
darosior: We can even copy it, "It is safe to call this function more than once and from different threads -- subsequent calls won't have any effects." from https://download.libsodium.org/doc/usage

hsmd/hsmd.c
darosior: > Or if it's not, and they do?

Then we load the seed, delete the `hsm_secret` and call `create_encrypted_hsm_secret()`.

> Or if it's been truncated to 31 bytes?

Could it be ? Anyway the encrypted seed is always > 32 so I can make the condition a `<=`

lightningd/lightningd.c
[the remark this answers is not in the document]
darosior: Done.

hsmd/hsmd.c
darosior: Corrected

hsmd/hsmd.c
darosior: > If it's encrypted, and they don't provide the key? [Edit: ah, fixed in next commit!]

As a side not I handle this in `hsm_control` because I was not able to get `satus_failed` (with a new failure type) to send an error before replying to `init`. The msg type would always be zero. Hence the hacky "wrong password" case handling.

darosior:
- Amended 8debd2fc69dfe206e2b3169229337cc229d54f61 to init libsodium in `setup_subdaemons()`
- Correct the `sizeof` argument when unlocking memory (and correct a wrong freeing)
- Use `<=` for the condition on the `hsm_secret` size
- Add a skipping condition if VALGRIND to the test: the key generation consumes too much, and the prompt doesn't play well with it.

Thanks for the review!

niftynei:
> Hmm looks like there is a travis cache problem: the check-source is obviously not up to date

afaict there's no cache for branch 3129 on Travis

darosior:
> > Hmm looks like there is a travis cache problem: the check-source is obviously not up to date
> 
> afaict there's no cache for branch 3129 on Travis

Ok, thanks for checking.
That's weird because the file version is not the same on Travis. I'm trying something else now ^^

common/daemon.c
```diff
@@ -146,6 +147,12 @@ void daemon_setup(const char *argv0,
 		memleak_init();
 #endif
 
+	/* We rely on libsodium for some of the crypto stuff, so we'd better
+	 * not start if it cannot do its job correctly. */
+	if (sodium_init() == -1)
+		errx(1, "Could not initialize libsodium. Maybe not enough entropy"
+		         " available ?");
```
trueptolemy: Only a question: What does this("not enough entropy") mean? It means the machine can't run `clightning` forever?

lightningd/options.c
```diff
@@ -366,6 +368,58 @@ static char *opt_clear_plugins(struct lightningd *ld)
...
+		return "Could not disable password echoing.";
+	printf("Enter hsm_secret password : ");
+	if (getline(&passwd, &passwd_size, stdin) < 0)
+		return "Could not read password from stdin.";
+	if (tcsetattr(fileno(stdin), TCSAFLUSH, &current_term) != 0)
+		return "Could not restore terminal options.";
+	printf("\n");
+
+	/* Derive the key from the password. */
+	if (strlen(passwd) < crypto_pwhash_argon2id_PASSWD_MIN)
```
trueptolemy: Will the length of the password and the complexity of the password affect the encryption?
Should we limit minimum length?

lightningd/options.c
```diff
@@ -366,6 +368,58 @@ static char *opt_clear_plugins(struct lightningd *ld)
...
+	/* Get the password from stdin, but don't echo it. */
+	if (tcgetattr(fileno(stdin), &current_term) != 0)
+		return "Could not get current terminal options.";
+	temp_term = current_term;
+	temp_term.c_lflag &= ~ECHO;
+	if (tcsetattr(fileno(stdin), TCSAFLUSH, &temp_term) != 0)
+		return "Could not disable password echoing.";
+	printf("Enter hsm_secret password : ");
+	if (getline(&passwd, &passwd_size, stdin) < 0)
+		return "Could not read password from stdin.";
```
trueptolemy: I think we'd better ask users to enter the password twice and check the consistency.

README.md
```diff
@@ -102,6 +102,8 @@ Once you've started for the first time, there's a script called
 `contrib/bootstrap-node.sh` which will connect you to other nodes on
 the lightning network.
 
+You can encrypt the BIP32 root seed (what is stored in `hsm_secret`) by passing the `--encrypted-hsm` startup argument. You can start `lightningd` with `--encrypted-hsm` on an already existing `lightning-dir` (with a not encrypted `hsm_secret`). If you pass that option, you __will not__ be able to start `lightningd` (with the same wallet) again without the password, so please beware with your password management. Also beware of not feeling too safe with an encrypted `hsm_secret`: unlike for `bitcoind` where the wallet encryption can restrict the usage of some RPC command, `lightningd` always need to access keys from the wallet which is thus __not locked__ (yet), even with an encrypted BIP32 master seed.
+
```
trueptolemy: I think here we can also remind users like the sentence "Note if the password or hsm_secret losts, you WILL lost all your funds forever. MUST take good care of your `hsm_secret` file and its password."

lightningd/options.c
trueptolemy: If two passwords are different, users must repeat this setting process.

trueptolemy: And if user enter `enter` or `null`, this process will be cancelled?

common/daemon.c
darosior: No it means that, in this rare case (maybe just after booting), the machine does not have enough source of randomness. https://en.wikipedia.org/wiki/Entropy_(computing)

lightningd/options.c
darosior: It is limited actually in `lightningd/options.c`, but the limit is set to 0 by the library.

lightningd/options.c
darosior: The decryption is done in `hsmd`, and I think it's better to keep it that way so that only one daemon interacts with `hsm_secret`. In case of a wrong password, an user can still `arrow_up` then `enter` :-).

> And if user enter enter or null, this process will be cancelled?

No, actually entering `null` is a backdoor to crack any C-lightning node's master seed :wink:

darosior:
>     1. Change the passwaord:
>        Ask user to enter the old password, if it is valid then ask user to enter a new password.
>
>     2. Unencrypt:
>        Ask user to enter the old password, if it is valid then decrypt the secret and don't encrypt it again.

I thought about making a tool to unencrypt `hsm_secret`, I'm not sure of making a JSONRPC request to `hsmd` (that's also why I made this a startup option in the first place).

rustyrussell:
Trivial rebase.

rustyrussell:
Ack e1af450062685045eeae609e66dffee3eaeec62c

Once Travis is happy, I'll merge.

PR #3132, opened 2019-10-04
Make `listtransactions` non-experimental but keep annotations hidden for now

cdecker:
We've been promising `listtransactions` for a while now, but it was kept back due to the annotations not being too useful since they don't have any association with inputs and outputs, and channel references could get clobbered if a transaction is associated with multiple transactions (splice anyone?).

This PR attempts to fix both of these by associating the annotations with the inputs and outputs that really give a transaction that facette (no longer clobbering channel references). Since the migration is not complete we still hide the annotation in `listtransactions` so people don't rely on them just yet, but we make `listtransactions` itself available.

`listtransactions` has details attached to each transaction, in the same format as `bitcoin-cli`'s `decoderawtransaction`, but I decided not to use the internal names (`vin`, `vout`, ...) instead opting for slightly longer but more descriptive names.

There is only one source of transactions that hasn't been migrated due to time, and that's `onchain_broadcast_transaction` which needs a bit more work, but is also mostly covered by explicit annotations created by `onchaind` while tracking the close.

In a followup PR I will:

 - Make annotations not experimental
 - Migrate all annotations from tx-based to input- or output-based (unless we want to keep transaction annotations for `txprepare` and `txsend`)
 -  Extend the `test_transaction_annotations` test to test for more annotations

Since this PR touches SQL queries I tested it against both DB backends :wink: 

Closes #2766 

Not assigning milestone since this is rather short-notice before the feature freeze.

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -3,7 +3,7 @@ from fixtures import *  # noqa: F401,F403
 from fixtures import TEST_NETWORK
 from flaky import flaky  # noqa: F401
 from lightning import RpcError, Millisatoshi
-from utils import only_one, wait_for, sync_blockheight
+from utils import only_one, wait_for, sync_blockheight, EXPERIMENTAL_FEATURES
 
 import pytest
 import time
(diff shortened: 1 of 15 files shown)


tests/test_wallet.py
```diff
@@ -451,6 +451,7 @@ def test_txprepare_restart(node_factory, bitcoind, chainparams):
 
 
 @unittest.skipIf(TEST_NETWORK != 'regtest', "Fee outputs throw off our output matching logic")
+@unittest.skipIf(not EXPERIMENTAL_FEATURES, "Tests annotations which are compiled only with experimental features")
```
darosior: You should include `EXPERIMENTAL_FEATURES` from utils

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+
+		if (tx->annotation.type != 0)
+			json_add_txtypes(response, "type", tx->annotation.type);
+		else
+			json_add_null(response, "type");
+
+		if (tx->annotation.channel.u64 != 0)
+			json_add_short_channel_id(response, "channel", &tx->annotation.channel);
+		else
+			json_add_null(response, "channel");
```
rustyrussell: I far prefer simply omit fields which don't make sense for a field, rather than set them to null.

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+			const char *txtype = txtype_to_string(ann->type);
+			json_object_start(response, NULL);
+			json_add_hex(response, "txid", in->txhash, sizeof(in->txhash));
+			json_add_u32(response, "index", in->index);
+			json_add_u32(response, "sequence", in->sequence);
+
+			if (txtype != NULL)
+				json_add_string(response, "type", txtype);
+			else
+				json_add_null(response, "type");
```
rustyrussell: Here too

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+			if (ann->channel.u64 != 0)
+				json_add_short_channel_id(response, "channel", &ann->channel);
+			else
+				json_add_null(response, "channel");
```
rustyrussell: Here too

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+			struct tx_annotation *ann = &tx->output_annotations[i];
+			const char *txtype = txtype_to_string(ann->type);
+			struct amount_asset amt = bitcoin_tx_output_get_amount(tx->tx, i);
+			struct amount_sat sat;
+
+			/* TODO We should eventually handle non-bitcoin assets as well. */
+			if (amount_asset_is_main(&amt))
+				sat = amount_asset_to_sat(&amt);
+			else
+				sat = AMOUNT_SAT(0);
```
rustyrussell: Hmm, I think you should just omit the "satoshis" field instead.

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+
+			json_object_start(response, NULL);
+
+			json_add_u32(response, "index", i);
+			json_add_amount_sat_only(response, "satoshis", sat);
+
+			if (txtype != NULL)
+				json_add_string(response, "type", txtype);
+			else
+				json_add_null(response, "type");
```
rustyrussell: Here too.

wallet/walletrpc.c
```diff
@@ -915,6 +923,92 @@ static void json_add_txtypes(struct json_stream *result, const char *fieldname,
...
+
+			if (ann->channel.u64 != 0)
+				json_add_short_channel_id(response, "channel", &ann->channel);
+			else
+				json_add_null(response, "channel");
```
rustyrussell: Here too.

wallet/wallet.c
```diff
@@ -3375,24 +3414,88 @@ struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t
...
+			if (!db_column_is_null(stmt, 4))
+				cur->annotation.type = db_column_u64(stmt, 4);
+			else
+				cur->annotation.type = 0;
+			if (!db_column_is_null(stmt, 5))
+				db_column_short_channel_id(stmt, 5, &cur->annotation.channel);
+			else
+				cur->annotation.channel.u64 = 0;
+
+			cur->output_annotations = tal_arrz(txs, struct tx_annotation, cur->tx->wtx->num_outputs);
```
niftynei: we should allocate these off of the lowest leaf that's still a parent, so `cur` in this case.

wallet/wallet.c
```diff
@@ -3375,24 +3414,88 @@ struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t
...
+			cur->input_annotations = tal_arrz(txs, struct tx_annotation, cur->tx->wtx->num_inputs);
```
niftynei: same as above

wallet/wallet.c
```diff
@@ -3375,24 +3414,88 @@ struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t
...
+
+		/* If this is a new entry, allocate it in the array and set
+		 * the common fields (all fields from the transactions
+		 * table. */
+		if (!bitcoin_txid_eq(&last, &curtxid)) {
+			last = curtxid;
+			tal_resize(&txs, count + 1);
+			cur = &txs[count];
+			db_column_txid(stmt, 0, &cur->id);
+			cur->tx = db_column_tx(txs, stmt, 1);
```
niftynei: these should be allocated off of `cur`, not `txs`

niftynei:
i will say one downside to omitting fields is that it makes them harder to discover for API users, but all in all i think it's better design to leave them out.

wallet/wallet.c
rustyrussell: Ah, no: cur is not a tal object, it's inside an array.  So this is actually correct.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3132/commits/0d0a6b5d8932f137702ed8a7a63ff595298dba70

PR #3135, opened 2019-10-05
lightningd: initialize the chainparams global in main.

### LINKED ISSUE #3134: Segfault when not explicitly specifying a network
ef7a63d8f87224be137a24672a1c10e3dd3e42b2 relies on `opt_set_network()` to be called in order to set the `chainparams` global.
```
diff --git a/lightningd/options.c b/lightningd/options.c
index 0472d9ab..480392e6 100644
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -241,6 +241,7 @@ static char *opt_set_network(const char *arg, struct lightningd *ld)
 {
        assert(arg != NULL);
 
+       printf("\n\n%s\n\n", arg);
        /* Set the global chainparams instance */
        chainparams = chainparams_for_network(arg);
```

```
$ make -j10 && ./lightningd/lightningd --lightning-dir $PWD/test_dir --testnet


testnet

Could not connect to bitcoind using bitcoin-cli. Is bitcoind running?

Make sure you have bitcoind running and that bitcoin-cli is able to connect to bitcoind.

You can verify that your Bitcoin Core installation is ready for use by running:

    $ bitcoin-cli -testnet echo 'hello world'
```

```
$ make -j10 && ./lightningd/lightningd --lightning-dir $PWD/test_dir
lightningd: FATAL SIGNAL 11 (version v0.6.1-2203-gef7a63d-modded)
0x55faf64ea1aa send_backtrace
	common/daemon.c:40
0x55faf64ea234 crashdump
	common/daemon.c:53
0x7f17553cc83f ???
	???:0
0x7f17554f16c3 ???
	???:0
0x55faf64fc7ab towire
	wire/towire.c:21
0x55faf64fcc57 towire_sha256
	wire/towire.c:172
0x55faf64fcc65 towire_sha256_double
	wire/towire.c:177
0x55faf64fccd9 towire_bitcoin_blkid
	wire/towire.c:194
0x55faf64fce5e towire_chainparams
	wire/towire.c:262
0x55faf64f39e3 towire_hsm_init
	hsmd/gen_hsm_wire.c:103
0x55faf64c8f13 hsm_init
	lightningd/hsm_control.c:96
0x55faf64cf975 main
	lightningd/lightningd.c:712
0x7f17553b909a ???
	???:0
0x55faf64bf109 ???
	???:0
0xffffffffffffffff ???
	???:0
Segmentation fault
```

darosior:
Fix #3134. What is fun is that I noticed it while locally testing https://github.com/ElementsProject/lightning/pull/3132, but that all Travis tests pass (we always pass a `--network` at daemon startup).

This makes it `testnet` by default but maybe we want to default to mainnet ? :-)

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -665,6 +665,10 @@ int main(int argc, char *argv[])
 	 * variables. */
 	ld = new_lightningd(NULL);
 
+	/*~ The global chainparams is needed to init subdaemons, and defaults
+	 * to testnet. */
+	chainparams = chainparams_for_network("testnet");
+
 	/* Figure out where our daemons are first. */
 	ld->daemon_dir = find_daemon_dir(ld, argv[0]);
 	if (!ld->daemon_dir)


darosior:
I finally changed my mind : I think we should default to mainnet. There are now more (public) nodes on the mainnet than on the testnet, and we claim in the README that C-lightning can be safely used on the mainnet.

darosior:
So there are now two commits, one which makes `bitcoind` the default network, and one which corrects the segfault.

rustyrussell:
> Fix #3134. What is fun is that I noticed it while locally testing #3132, but that all Travis tests pass (we always pass a `--network` at daemon startup).
> 
> This makes it `testnet` by default but maybe we want to default to mainnet ? :-)

I think for 0.8.0 we should switch the defaults, with some care (ie. if the db is already initialized, that should set the default, so it will only apply for new installations).

PR #3139, opened 2019-10-06
check for return code in the else block

yashbhutwala:


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3139/commits/9014878b4048ddbe637ea605029cd376300c8664

Issue #3118, opened 2019-09-30
Feature: Asymmetric decryption using `hsmd` secret (encrypted using `hsmd` public key)


ZmnSCPxj:
Propose addition of new command:

## `decryptblob`

    decryptblob blob ephemeralPubkey

Decrypts a hex-encoded blob using an ephemeral public key and the node private key.

To encrypt:

* Allocate some non-swappable memory.
* Generate a new ephemeral private key in the non-swappable memory from a high-entropy random source.
* Multiply the ephemeral private key by G and store the point in normal memory.  This is the ephemeral public key and should be stored with the encrypted message.
* Multiply the ephemeral private key by the node public key and store the point in non-swappable memory.
* SHA256 the point (product of ephemeral privkey and node pubkey), put hash in non-swappable memory.
* Create a Chacha20 state in non-swappable memory, copy the hash above as the key, set the nonce and counter to 0.
* XOR the Chacha20 output with the message (in normal memory).
* Clear all non-swappable memory to 0 and release.

To decrypt, `hsmd` will recover the key state by multiplying its node private key with the ephemeral public key, generating the same point which when hashed generates the same Chacha20 state.

(exact details subject to change, we just need *some* method of asymmetric encryption, which is easiest done using symmetric encryption and ECDH, and which is difficult to hack with known-plaintext attack.)

----------

Use case is for `db_write` hooks that need to store the database queries in a server that is only trusted to store the data (but not trusted to leak it to everybody else).  The `db_write` hook could encrypt on-the-fly before responding to the `db_write` hook.

On recovery, we only need:

* The original `hsm_secret`
  * We could switch to generating this from standard words.
* The encrypted database queries file, containing point-ciphertext pairs.

We replace the `hsm_secret` with the recovered `hsm_secret`, then restart `lightningd` with `--offline`, then `decodeblob` each point-ciphertext pair and extract each database query on a fresh database.  Then stop the `lightningd` and replace its database.

We may need to have a better way to push binaries through the RPC though.

Thoughts?

ZmnSCPxj:
Some more thoughts.

[...]

PR #3150, opened 2019-10-11
signmessage support

rustyrussell:
Special thanks to Twitter for supplying testcases:

https://twitter.com/rusty_twit/status/1182524014440222720


tests/test_misc.py
```diff
@@ -1646,3 +1646,39 @@ def test_relative_config_dir(node_factory):
     assert os.path.isabs(l1.rpc.listconfigs()["lightning-dir"])
     l1.stop()
     os.chdir(initial_dir)
+
+
+def test_signmessage(node_factory):
+    l1 = node_factory.get_node()
+
+    corpus = [ [ None,
+                 "this is a test!",
```
trueptolemy: These `whitespace`s make travis-ci unhappy :-)

rustyrussell:
> BTW, can lightning-checkmessage upgrade to json command from a tool in the future?

I thought about making it a command, but it really doesn't need a running node at all. A plugin would be trivial, however.

niftynei:
strongly disagree with not providing a 'symmetric' decryption command with this. from the average user's perspective, the tool is not discoverable.

a plugin is fine, but we shouldn't ship this without an equally obvious, integrated lightning-cli command for decryption.

darosior:
> a plugin is fine, but we shouldn't ship this without an equally obvious, integrated lightning-cli command for decryption.

On the same subject I began to make a `hsmtools` plugin to which I wanted to include `hsm_secret` decryption, and these two functionalities (so both the encryption and the decryption could be part of the same interface). We could also add https://github.com/ElementsProject/lightning/issues/3118 and even wallet locking (`wallettools`?).

trueptolemy:
> On the same subject I began to make a `hsmtools` plugin to which I wanted to include `hsm_secret` decryption, and these two functionalities (so both the encryption and the decryption could be part of the same interface). We could also add #3118 and even wallet locking (`wallettools`?).

`hsmtools` and `wallettools` can be merged into one plugin to provides general encryption/decryption related functions?

darosior:
> hsmtools and wallettools can be merged into one plugin to provides general encryption/decryption related functions?

Yeah that's what I propose

rustyrussell:
Yes, a hsmtools combo plugin makes sense.

I'll move the standalone tool to devtools/ and move implementation into the core.

rustyrussell:
Trivial rebase (CHANGELOG.md conflict).  All feedback addressed.

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - JSON API: `listfunds` now lists a blockheight for confirmed transactions, and has `connected` and `state` fields for channels, like `listpeers`.
 - JSON API: `fundchannel_start` now includes field `scriptpubkey`
 - JSON API: New method `listtransactions`
-- JSON API: `signmessage` will now create a signature from your node on a message.
+- JSON API: `signmessage` will now create a signature from your node on a message; `checkmessage` will verify it.
```
niftynei: :+1: thanks!!

doc/lightning-checkmessage.7.md
```diff
@@ -0,0 +1,51 @@
...
+
+The **checkmessage** RPC command is the counterpart to
+**signmessage**: given a node id (*pubkey*), signature (*zbase*) and a
+*message*, it verifies that the signature was generated by that node
+for that message (more technically: by someone who knows that node's
+secret).
+
+As a special case, if *pubkey* is not specified, we will try every
+known node key (as per *listnodes*), and verification succeeds if it
+matches for any one of them.  Note: this is implemented far more
```
niftynei: @niftynei is laughing at 'implemented more efficiently'

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3150/commits/4a73962e8aa0ac2f34b957ad36308a60975d6603

PR #3152, opened 2019-10-11
gossipd: push our own gossip messages harder.

rustyrussell:
I had a report of a 0.7.2 user whose node hadn't appeared on 1ml.  Their
node_announcement wasn't visible to my node, either.

I suspect this is a consequence of recent version reducing the amount of
gossip they send, as well as large nodes increasingly turning off gossip
altogether from some peers (as we do).  We should ignore timestamp filters
for our own channels: the easiest way to do this is to push them out
directly from gossipd (other messages are sent via the store).


gossipd/gossipd.c
```diff
@@ -402,6 +402,49 @@ static u8 *handle_node_announce(struct peer *peer, const u8 *msg)
...
+/*~ Large peers often turn off gossip msgs (using timestamp_filter) from most
+ * of their peers, however if the gossip is about us, we should spray it to
+ * everyone whether they've set the filter or not, otherwise it might not
+ * propagate! */
+void push_gossip(struct daemon *daemon, const u8 *msg TAKES)
+{
+	struct peer *peer;
+
+	list_for_each(&daemon->peers, peer, list)
+		queue_peer_msg(peer, msg);
```
trueptolemy: Sorry for my "too early" ACK.
If we call `push_gossip(daemon, take(msg))`, here `msg` is regarded as `TAKEN` not only in `push_gossip()`, but also in `queue_peer_msg()`.
In this loop, if the parameter `msg` is marked as `TAKEN`, then the first `queue_peer_msg` would `free` it.
So, maybe `msg` shouldn't be declared as "TAKES" in `void push_gossip(struct daemon *daemon, const u8 *msg TAKES)`.
Is this reason accurate?

rustyrussell: Yep, good catch. Will fix...

rustyrussell:
Rebased and fixed thinko, as diagnosed by @trueptolemy

trueptolemy:
Indeed LGTM now, but always TIMEOUT on travis-ci.
I restart the failed jobs again.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3152/commits/80e50baf34fb170bc372eba3b04341c12fd8ca27

PR #3153, opened 2019-10-12
Add `close_to` param for specifying upfront shutdown script

niftynei:
Allow users to specify the `upfront_shutdown_script` via a new parameter on `fundchannel_start`.

### DIFF
--- a/doc/lightning-fundchannel_start.7
+++ b/doc/lightning-fundchannel_start.7
@@ -3,7 +3,7 @@
 lightning-fundchannel_start - Command for initiating channel establishment for a lightning channel
 .SH SYNOPSIS
 
-\fBfundchannel_start\fR \fIid\fR \fIamount\fR [\fIfeerate\fR \fIannounce\fR]
+\fBfundchannel_start\fR \fIid\fR \fIamount\fR [\fIfeerate\fR \fIannounce\fR \fIclose_to\fR]
 
 .SH DESCRIPTION
 
@@ -26,6 +26,11 @@ commitment transactions\.
 \fIannounce\fR whether or not to announce this channel\.
 
 
+\fIclose_to\fR is a Bitcoin address to which the channel funds should be sent to
+on close\. Only valid if both peers have negotiated \fBoption_upfront_shutdown_script\fR\.
+Returns \fBclose_to\fR set to closing script iff is negotiated\.
+
+
 Note that the funding transaction MUST NOT be broadcast until after
 channel establishment has been successfully completed by running
 \fBfundchannel_complete\fR, as the commitment transactions for this channel
@@ -35,6 +40,8 @@ transaction before that can lead to unrecoverable loss of funds\.
 .SH RETURN VALUE
 
 On success, returns the \fIfunding_address\fR and the \fIscriptpubkey\fR for the channel funding output\.
+If a \fBclose_to\fR address was provided, will close to this address iff the \fBclose_to\fR address is
+returned in the response\. Otherwise, the peer does not support \fBoption_upfront_shutdownscript\fR\.
 
 
 On failure, returns an error\.
@@ -51,8 +58,4 @@ lightning-fundchannel_\fBcomplete\fR(7), lightning-fundchannel_\fBcancel\fR(7)
 .SH RESOURCES
 
 Main web site: \fIhttps://github.com/ElementsProject/lightning\fR
-
-.HL
-
-Last updated 2019-06-12 11:16:20 CEST
 
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -77,6 +77,7 @@ struct funding_channel {
 	struct amount_msat push;
 	struct amount_sat funding;
 	u8 channel_flags;
+	const u8 *our_upfront_shutdown_script;
 
 	/* Variables we need to compose fields in cmd's response */
 	const char *hextx;
@@ -165,6 +166,7 @@ wallet_commit_channel(struct lightningd *ld,
 		      u8 channel_flags,
 		      struct channel_info *channel_info,
 		      u32 feerate,
+		      const u8 *our_upfront_shutdown_script,
 		      const u8 *remote_upfront_shutdown_script)
 {
 	struct channel *channel;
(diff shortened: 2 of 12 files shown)


lightningd/opening_control.c
```diff
@@ -1078,6 +1092,7 @@ static struct command_result *json_fund_channel_start(struct command *cmd,
 			   p_req("amount", param_sat, &amount),
 			   p_opt("feerate", param_feerate, &feerate_per_kw),
 			   p_opt_def("announce", param_bool, &announce_channel, true),
+			   p_opt("close_to", param_bitcoin_address, &fc->our_upfront_shutdown_script),
```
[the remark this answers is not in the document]
darosior: I think it would made VALGRIND tests pass

justinmoon:
Concept ACK. This would be really useful.

trueptolemy:
https://github.com/ElementsProject/lightning/blob/a91b454749f43e5eea414ad0a3cb5f66adc9b666/lightningd/peer_control.c#L1459-L1475
Here's a `FIXME` that should be considered. If we set `close_to` in `fundchannel_start`, we must fail the `close` command if the `destination` doesn't equal to `close_to` address.

And here's should be another check:
https://github.com/ElementsProject/lightning/blob/bd55f6d9408d7928852637e9581a3d9d5107111f/channeld/channeld.c#L2791-L2794

rsbondi:
> If we set close_to in fundchannel_start, we must fail the close command if the destination doesn't equal to close_to address.

I ran into an issue with this when testing, got an unknown random P2WSH address receiving the funds rather than `close_to` or `destination`, so needs a fix, but maybe `destination` should override the `close_to`?   But failing would be acceptable.

trueptolemy:
> random P2WSH address

Maybe it is the internal wallet address? Have you executed only one `close` command? 

> but maybe destination should override the close_to?

I think if we set `close_to`(and it was accepted by the remote), it has the highest priority(because **in spec, `close` must accept upfront shutdown script**). Or if we set the `destination` of `close`, it has the second priority. Finally, the internal wallet address has the lowest priority.

rsbondi:
> Maybe it is the internal wallet address

those are P2WPKH or P2SH

> Have you executed only one close command and specified the destination?

`close` with `destination` works as expected when I open without `close_to`

spec rules, so fail seems best.

trueptolemy:
> those are P2WPKH or P2SH

Yeah, they're p2wpkh

lightningd/opening_control.c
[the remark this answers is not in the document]
niftynei: `p_opt_def` sets the variable to `NULL` by default; the problem was in the deprecated codepath. :)

niftynei:
> Needs to go into db, I think, and we make sure it's the same before trying shutdown.

There's already a database field for it in `channels`: local_shutdown_script. It plugs directly into that.

> I ran into an issue with this when testing, got an unknown random P2WSH address receiving the funds rather than close_to or destination, so needs a fix, but maybe destination should override the close_to? But failing would be acceptable.

The `P2SH` script that you got should be the `unilateral` closing address. In the version of this that you tested, attempting to issue a 'close' command to a destination different than what was specified upfront results in the peer throwing an error and the channel closing to a unilateral output script.

I changed this so that if you attempt to called `close [destination]` with a destination that is not the same as  the upfront shutdown script you get an RPC error. One side effect of this change is that you'll no longer be able to call `close` mutliple times with different destinations, but this is a reasonable tradeoff.

> Here's a FIXME that should be considered. If we set close_to in fundchannel_start, we must fail the close command if the destination doesn't equal to close_to address.
> And here's should be another check:

regarding this second check, we check before we call into this; checking here would be redundant.

thanks for the reviews :+1:

doc/lightning-fundchannel_start.7
```diff
@@ -1,9 +1,9 @@
 .TH "LIGHTNING-FUNDCHANNEL_START" "7" "" "" "lightning-fundchannel_start"
 .SH NAME
-lightning-fundchannel_start - Command for initiating channel establishment for a lightning channel
+
```
rustyrussell: How did this get corrupted?  Do you need to update your mrkd?

doc/lightning-fundchannel_start.7
```diff
@@ -26,6 +26,11 @@ commitment transactions\.
 \fIannounce\fR whether or not to announce this channel\.
 
 
+\fIclose_to\fR is a Bitcoin address to which the channel funds should be sent to
+on close\. Only valid if both peers have negotiated \fBoption_upfront_shutdown_script\fR\.
+Returns \fBusing_close_to\fR set to true if it will\.
```
rustyrussell: Prefer just returning the actual shutdown script as "close_to" iff it's used?

doc/lightning-fundchannel_start.7
niftynei: I did in fact need to update `mrkd`

rustyrussell:
I've collapsed these fixes back into two commits, and am testing locally now:

```
E            - Node /tmp/ltests-msvtc38j/test_closing_specified_destination_1/lightning-1/ has memory leaks: [
E               {
E                       "ccan/ccan/tal/tal.c:437 (tal_alloc_)",
E                       "ccan/ccan/tal/tal.c:466 (tal_alloc_arr_)",
E                       "bitcoin/script.c:294 (scriptpubkey_p2wpkh)",
E                       "lightningd/peer_control.c:190 (p2wpkh_for_keyidx)",
E                       "lightningd/peer_control.c:1458 (json_close)",
E                       "lightningd/jsonrpc.c:645 (parse_request)",
E                       "lightningd/jsonrpc.c:743 (read_json)",
E                       "ccan/ccan/io/io.c:417 (io_ready)",
E                       "ccan/ccan/io/poll.c:445 (io_loop)",
E                       "lightningd/io_loop_with_timers.c:24 (io_loop_with_timers)",
E                       "lightningd/lightningd.c:859 (main)"
E                   ],
E                   "label": "bitcoin/script.c:294:u8[]",
E                   "parents": [
E                       "lightningd/channel.c:190:struct channel"
E                   ],
E                   "value": "0x5645bf695b88"
```

rustyrussell:
Rebased fixups back into two patches (no changes), then added a trivial fix on top, found by leak detection.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3153/commits/9542e13d4988cabb83b3553a43a7b9cd505d61e5

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3153/commits/3a90f33e9faf7a408cd8595da7114d0ab7695c43

PR #3158, opened 2019-10-14
gossip seeker fixes

rustyrussell:
1. We were tending to give up on streaming gossip too soon when we have no gossip_store,
   since we didn't count channel_updates on new channels correctly.
2. Our node_announcement probe was a bit useless: make it sample randomly instead.
3. Choose 5, not 3, peers to stream us gossip in steady-state.  This is less aggressive until we
    have evidence that 3 is sufficient (if it is!), or we move to a more dynamic basis.

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -1955,7 +1955,8 @@ bool handle_pending_cannouncement(struct daemon *daemon,
 
 static void update_pending(struct pending_cannouncement *pending,
 			   u32 timestamp, const u8 *update,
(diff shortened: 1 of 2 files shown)


rustyrussell:
I've been torturing my node and watching it try to recover gossip.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3158/commits/e894f662c1085145718c233eac5b4c8c96a5f68e

Issue #3119, opened 2019-09-30
Make Millisatoshi JSON serializable


s-tikhomirov:
### Issue and Steps to Reproduce

I'm writing some Python script that interact with c-lightning instances via pylightning. Many commands such as `listpeers` return a long JSON string, which is hard to make sense of in the terminal. I tried to pretty-print it using JSON package:
```
print(json.dumps(l1.listpeers()))
```
but got an error:
```
TypeError: Object of type 'Millisatoshi' is not JSON serializable
```
Would be nice to make Millisatoshi JSON-serializable to enable pretty-printing.

### `getinfo` output
```
{
   "id": "020bb1df60102c24c4b28612c23fb34356559a001924fa6997ba0ea5ee19bbed41",
   "alias": "PEEVEDWATCH",
   "color": "020bb1",
   "num_peers": 1,
   "num_pending_channels": 0,
   "num_active_channels": 1,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv6",
         "address": "<IP address>",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv6",
         "address": "::",
         "port": 9735
      },
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "v0.7.2.1-291-g14b55d0",
   "blockheight": 378,
   "network": "regtest",
   "msatoshi_fees_collected": 0,
   "fees_collected_msat": "0msat"
}
```

whitslack:
This deficiency also impacts the `lightning-pay` script that ships with C-Lightning.

```
Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.6/lightning-pay", line 34, in <module>
    print("Bolt11 decoded:\n%s" % json.dumps(bolt11, indent=4))
  File "/usr/lib64/python3.6/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python3.6/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib64/python3.6/json/encoder.py", line 430, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib64/python3.6/json/encoder.py", line 404, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.6/json/encoder.py", line 437, in _iterencode
    o = _default(o)
[...]

PR #3163, opened 2019-10-14
Pylightning cleanups

darosior:
I went ahead to try to fix https://github.com/ElementsProject/lightning/issues/3119, but actually I don't think it should be fixed on our end. Here are anyway some drive-by cleanups.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3163/commits/ce001a0f3fd63ed9f734e491e4c96d98ae95a768

PR #3168, opened 2019-10-15
Cleanup chainparams duplication

darosior:
Prefer using the constant everywhere instead of sometimes using it and some other times the `bitcoind` member.

Still to do:
- [] ~Remove the `chainparams` member in other structures (eg `struct tx`)~
- [] ~Remove the `chainparams` arguments in utility functions~

Changelog-None


darosior:
Removing the `chainparams` member of structures like [channels / peers / `hsmd`'s client] states I wonder if:
- we really want this ? [Need concept ack]
- we really need this constant, and if I should rather remove the constant than the local structs (ping @cdecker) ?

I can't add a tag so I'll dirtyly rename the PR title to ask  for conceptual reviews ^^

cdecker:
I have been struggling with this for a while myself. The general consensus is to avoid globals whenever possible, and have them be passed in through context arguments when needed. In our case the top-most context is the `struct lightningd` (often referred to as `ld`) instance. Currently (and for the foreseeable future) we have a single chain that we serve by a daemon at a time, so stuffing the `chainparams` into the `ld` context is basically equivalent with a global variable, except the former requires explicit argument passing to get access to the `chainparams` instance.

I think the one downside that a global `chainparams` instance has is that we end up indirectly depending on it being initialized, without a clear indication of this (tests need to know that they are supposed to initialize the variable for instance). This is already the case in the current `master` version, so this PR doesn't worsen this situation, it just makes it more difficult if we decide to clean this up some time in the future.

So no real opposition from my side, but I wonder if we could somehow clean this up and avoid the global instead :wink:

darosior:
> this PR doesn't worsen this situation, it just makes it more difficult if we decide to clean this up some time in the future.

Don't you think it's better to have only the global variable rather than both the global and contextual variables ?

> So no real opposition from my side, but I wonder if we could somehow clean this up and avoid the global instead wink 

Yes, that's why I pinged you :wink: : to know if, since you introduced it, there was a blocker to avoid the global so that I could inspect on that side instead of making the structural change of removing contextual `chainparams`s.

rustyrussell:
Do it, just make sure it always crashes if it's not initialized.

As Linus Torvalds once said (misquoting Voltaire) "The perfect is the enemy of the good".

If we need to make it contextual in some future, that's simple (if not *easy*).

cdecker:
I rebased the PR and would be happy to merge it unless @darosior wants to add some more changes on top of what we already have :wink:

darosior:
@cdecker It's unfortunately not finished at all (cf todos) :/ I'm on something else at this time but will continue this asap (anyway it's up for grab ^^)

cdecker:
I think we should merge it as is, and then revisit it later.

ACK https://github.com/ElementsProject/lightning/pull/3168/commits/b7d25730623f6ee398bc4ef7351ea071cca583fd

darosior:
As you want, ACK then :-)

PR #3172, opened 2019-10-16
hsm encryption fixups

darosior:
~~Noticed it while working on decryption with an `hsmtools` plugin.. This can be workarounded but since the encryption functionality was not merged long time ago and we haven't released yet, probably better to just correct it now.~~

Some `hsm_secret` encryption corrections : 
- don't take the newline character into account when deriving the encryption key
- correct the salt length (...)

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -398,6 +398,8 @@ static char *opt_set_hsm_password(struct lightningd *ld)
 	printf("Enter hsm_secret password : ");
 	if (getline(&passwd, &passwd_size, stdin) < 0)
 		return "Could not read password from stdin.";
+	if(passwd[strlen(passwd) - 1] == '\n')
+		passwd[strlen(passwd) - 1] = '\0';
 	if (tcsetattr(fileno(stdin), TCSAFLUSH, &current_term) != 0)
 		return "Could not restore terminal options.";
 	printf("\n");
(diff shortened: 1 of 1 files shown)


lightningd/options.c
```diff
@@ -379,7 +379,7 @@ static char *opt_set_hsm_password(struct lightningd *ld)
 	struct termios current_term, temp_term;
 	char *passwd = NULL;
 	size_t passwd_size = 0;
-	u8 salt[11] = "c-lightning";
+	u8 salt[16] = "c-lightning\0\0\0\0\0";
```
trueptolemy: Only a question :-) : why add these `\0` ?

lightningd/options.c
```diff
@@ -398,6 +398,8 @@ static char *opt_set_hsm_password(struct lightningd *ld)
 	printf("Enter hsm_secret password : ");
 	if (getline(&passwd, &passwd_size, stdin) < 0)
 		return "Could not read password from stdin.";
+	if(passwd[strlen(passwd) - 1] == '\n')
+		passwd[strlen(passwd) - 1] = '\0';
```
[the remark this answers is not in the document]
darosior: Yes, that's why I said this on IRC (though I don't think anybody has tested this as there is a high probability they ran into https://github.com/ElementsProject/lightning/pull/3174). Sorry about that...

lightningd/options.c
darosior: > Did valgrind catch this at all?

No..

> For reference the salt length is defined here

Yes, and I knew it : I don't know how I missed that in the first place..

PR #3177, opened 2019-10-17
hsmtools: a plugin for all things hsm_secret

darosior:
This not really all things since I wanted to also move `signmessage` and `checkmessage` to the plugin, but then if the `hsm_secret` is encrypted the user would need to also pass its encryption password. What do you think ? (They can also be added later)

So this comports only two calls ATM : `decrypthsm` and `encrypthsm`. This also moves the encryption doc in the README to the "Further information" part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ c-lightning is a lighweight, highly customizable and [standard compliant][std] i
 	* [Configuration File](#configuration-file)
 * [Further Information](#further-information)
     * [Pruning](#pruning)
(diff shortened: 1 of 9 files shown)


darosior:
After talking with @cdecker and @ZmnSCPxj it seems that having another process which deals with `hsm_secret` is not a good design. I'm now making this a tool and will try to integrate a generalization of @Sjors 's commitment dumper.

darosior:
Hmm I wonder if it makes sense to include it in [lightningd/plugins](https://github.com/lightningd/plugins/) instead ?

PR #3181, opened 2019-10-18
Fix elements after the libwally upgrade

cdecker:
`libwally` was getting upset when we were trying to be helpful, and we weren't too happy to see blinded value or asset commitments.

### DIFF
--- a/bitcoin/tx.c
+++ b/bitcoin/tx.c
@@ -204,6 +204,8 @@ const u8 *bitcoin_tx_output_get_script(const tal_t *ctx,
 	return res;
 }
 
+/* FIXME(cdecker) Make the caller pass in a reference to amount_asset, and
+ * return false if unintelligible/encrypted. (WARN UNUSED). */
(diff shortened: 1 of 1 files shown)


bitcoin/tx.c
```diff
@@ -326,7 +329,7 @@ static void push_tx(const struct bitcoin_tx *tx,
 	assert(res == WALLY_OK);
 	serialized = tal_arr(tmpctx, u8, len);
 
-	res = wally_tx_to_bytes(tx->wtx, flag, serialized, len, &written);
+	res = wally_tx_to_bytes(tx->wtx, flag & WALLY_TX_FLAG_USE_WITNESS, serialized, len, &written);
```
niftynei: at this point you really could just initialize the flag to `WALLY_TX_FLAG_USE_WITNESS`, no?

cdecker: Yep.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3181/commits/b6cf4311dc3e3f06a259da93fb0a54c61285b22d

PR #3182, opened 2019-10-18
add gettext to build release script

niftynei:
new dependency for this release means we need it for the release builds. I couldn't figure out how to add it for the fedora builds, but this seems to fix the breaks.


jb55:
ACK 365efeaad0d4ac25805dbc22e71e6bcbed026bfb

I also ran into this when packaging the new version

PR #3186, opened 2019-10-19
Non-plugin version of hsmtool

darosior:
This closes #3177 as it is a concurrent version of the `hsmtools` functionalities.

Changelog-None

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -567,7 +567,7 @@ def test_transaction_annotations(node_factory, bitcoind):
 
 
 @unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
-def test_hsm_secret_encryption(node_factory, executor):
+def test_hsm_secret_encryption(node_factory):
     l1 = node_factory.get_node()
     password = "reckful\n"
     # We need to simulate a terminal to use termios in `lightningd`.
@@ -588,7 +588,6 @@ def test_hsm_secret_encryption(node_factory, executor):
     l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
                     wait_for_initialized=False)
     time.sleep(3 if SLOW_MACHINE else 1)
-    os.write(master_fd, password[2:].encode("utf-8"))
     err = "hsm_secret is encrypted, you need to pass the --encrypted-hsm startup option."
     assert l1.daemon.is_in_log(err)
 
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -605,3 +605,61 @@ def test_hsm_secret_encryption(node_factory):
     os.write(master_fd, password.encode("utf-8"))
     l1.daemon.wait_for_log("Server started with public key")
     assert id == l1.rpc.getinfo()["id"]
+
+
+@unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
+def test_hsmtool_secret_decryption(node_factory):
+    l1 = node_factory.get_node()
+    password = "reckless\n"
+    hsm_path = os.path.join(l1.daemon.lightning_dir, "hsm_secret")
+    # We need to simulate a terminal to use termios in `lightningd`.
+    master_fd, slave_fd = os.openpty()
+
+    # Encrypt the master seed
+    l1.rpc.stop()
+    l1.daemon.opts.update({"encrypted-hsm": None})
+    l1.daemon.start(stdin=slave_fd, wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    os.write(master_fd, password.encode("utf-8"))
+    l1.daemon.wait_for_log("Server started with public key")
+    node_id = l1.rpc.getinfo()["id"]
+    l1.stop()
+
+    # We can't use a wrong password !
+    cmd_line = ["tools/hsmtool", "decrypt", hsm_path, "A wrong pass"]
+    assert subprocess.Popen(cmd_line).wait() != 0
+
+    # Decrypt it with hsmtool
+    cmd_line[3] = password[:-1]
+    assert subprocess.Popen(cmd_line).wait() == 0
+    # Then test we can now start it without password
+    l1.daemon.opts.pop("encrypted-hsm")
+    l1.daemon.start(stdin=slave_fd, wait_for_initialized=True)
+    assert node_id == l1.rpc.getinfo()["id"]
+    l1.stop()
+
+    # Test we can encrypt it offline
+    cmd_line[1] = "encrypt"
+    assert subprocess.Popen(cmd_line).wait() == 0
+    # Now we need to pass the encrypted-hsm startup option
+    l1.stop()
+    l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
+                    wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    err = "hsm_secret is encrypted, you need to pass the --encrypted-hsm startup option."
+    assert l1.daemon.is_in_log(err)
+    l1.daemon.opts.update({"encrypted-hsm": None})
+    l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
+                    wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    os.write(master_fd, password.encode("utf-8"))
+    l1.daemon.wait_for_log("Server started with public key")
+    assert node_id == l1.rpc.getinfo()["id"]
+    l1.stop()
+
+    # And finally test that we can also decrypt if encrypted with hsmtool
+    cmd_line[1] = "decrypt"
+    assert subprocess.Popen(cmd_line).wait() == 0
+    l1.daemon.opts.pop("encrypted-hsm")
+    l1.daemon.start(stdin=slave_fd, wait_for_initialized=True)
+    assert node_id == l1.rpc.getinfo()["id"]
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@ c-lightning is a lighweight, highly customizable and [standard compliant][std] i
 	* [Configuration File](#configuration-file)
 * [Further Information](#further-information)
     * [Pruning](#pruning)
+    * [HD wallet encryption](#hd-wallet-encryption)
 	* [Developers](#developers)
 
 ## Project Status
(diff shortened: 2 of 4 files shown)


tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
...
+    l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
+                    wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    err = "hsm_secret is encrypted, you need to pass the --encrypted-hsm startup option."
+    assert l1.daemon.is_in_log(err)
+    l1.daemon.opts.update({"encrypted-hsm": None})
+    l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
+                    wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    os.write(master_fd, "{}".format(password).encode("utf-8"))
```
niftynei:
```suggestion
    os.write(master_fd, password.encode("utf-8"))
```

tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
     os.write(master_fd, password.encode("utf-8"))
     l1.daemon.wait_for_log("Server started with public key")
     assert id == l1.rpc.getinfo()["id"]
+
+
+@unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
+def test_hsmtools_secret_decryption(node_factory):
+    l1 = node_factory.get_node()
+    password = "reckless\n"
```
niftynei:
```suggestion
    password = "reckless\n".encode("utf-8") 
```

?? then you can use it inline elsewhere.

tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
...
+    assert node_id == l1.rpc.getinfo()["id"]
+    l1.stop()
+
+    # And finally test that we can also decrypt if encrypted with hsmtools
+    cmd_line = cmd_line.replace("encrypthsm", "decrypthsm")
+    assert subprocess.Popen(cmd_line.split(" ")).wait() == 0
+    l1.daemon.opts.pop("encrypted-hsm")
+    l1.daemon.start(stdin=slave_fd, wait_for_initialized=True)
+    assert node_id == l1.rpc.getinfo()["id"]
+    l1.stop()
```
niftynei: should we also verify that an incorrect password fails to decrypt?

tools/hsmtools.c
```diff
@@ -0,0 +1,214 @@
...
+	                                                    key.data) != 0)
+		errx(ERROR_LIBSODIUM, "Could not initialize the crypto state");
+	if (crypto_secretstream_xchacha20poly1305_pull(&crypto_state, hsm_secret.data,
+	                                               NULL, 0, cipher, sizeof(cipher),
+	                                               NULL, 0) != 0)
+		errx(ERROR_LIBSODIUM, "Could not retrieve the seed. Wrong password ?");
+	close(fd);
+
+	/* Create a backup file, "just in case". */
+	rename(hsm_secret_path, "hsm_secret.backup");
```
niftynei: does `rename` update just the file name, or the entire path?

tools/hsmtools.c
```diff
@@ -0,0 +1,214 @@
...
+
+	/* Write the encrypted hsm_secret. */
+	if (!write_all(fd, header, sizeof(header))
+		|| !write_all(fd, cipher, sizeof(cipher))) {
+		unlink_noerr(hsm_secret_path);
+		close(fd);
+		rename("hsm_secret.backup", hsm_secret_path);
+		errx(ERROR_HSM_FILE, "Failure writing cipher to hsm_secret.");
+	}
+	if (!write_all(fd, cipher, sizeof(cipher))) {
```
niftynei: i don't think you need this, you're writing the cipher in the previous if statement as well.

tests/test_wallet.py
darosior: Yeah but then I would have to decode it for Pylightning `call()` method, which (since we also need to remove the trailing `\n`) is arguably dirtier ?

tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
...
+    l1.rpc.stop()
+    l1.daemon.opts.update({"encrypted-hsm": None})
+    l1.daemon.start(stdin=slave_fd, wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
+    os.write(master_fd, password.encode("utf-8"))
+    l1.daemon.wait_for_log("Server started with public key")
+    node_id = l1.rpc.getinfo()["id"]
+    l1.stop()
+
+    # Decrypt it via RPC...
```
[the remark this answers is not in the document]
darosior: Oops that's a trailing comment from the plugin version :-)

tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
...
+    cmd_line = "tools/hsmtools decrypthsm {} {}".format(hsm_path, password[:-1])
```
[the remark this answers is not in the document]
darosior: I find it more readable this way, but I can make it a list

tests/test_wallet.py
darosior: Hmmm I might have wrote this while there was an interesting talk ongoing :joy:

tests/test_wallet.py
darosior: Absolutely ! Thanks

tools/hsmtools.c
darosior: entire path

tools/hsmtools.c
darosior: I need to write both the header and the body of the cipher : the variable name is maybe not very explicit..

darosior:
Thanks for the review, I am about to push the polish and the `dumpcommitments` method. I'll integrate your comments.

tools/hsmtools.c
niftynei: yeah you write the cipher already in L161

tools/hsmtools.c
niftynei: a bit of a stretch goal, but it'd be nice if it's in the same directory as the original hsm file, so that it's easy to find/recover in case of a problem

tools/hsmtools.c
darosior: Oh yes, I forgot to remove it here thanks

tools/hsmtools.c
darosior: Actually it was in plugin version (we chdir in libplugin.c at `init`) but I kept it although we don"t run it from the same directory.
In case of a not-already handled problem, the backup is still in the running directory.. And it would be quite hacky, just for a tool  : recovering the path to the lightning-dir from the hsm_secret path, then prefixing the rename..

tests/test_wallet.py
```diff
@@ -606,3 +605,61 @@ def test_hsm_secret_encryption(node_factory, executor):
...
+    assert subprocess.Popen(cmd_line.split(" ")).wait() != 0
+
+    # Test we can encrypt it offline
+    cmd_line = cmd_line.replace("decrypthsm", "encrypthsm")
+    assert subprocess.Popen(cmd_line.split(" ")).wait() == 0
+    # Now we need to pass the encrypted-hsm startup option
+    l1.stop()
+    l1.daemon.start(stdin=slave_fd, stderr=subprocess.STDOUT,
+                    wait_for_initialized=False)
+    time.sleep(3 if SLOW_MACHINE else 1)
```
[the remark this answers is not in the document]
darosior: I can't as with `wait_for_log` the log messages is already in the log when we start watching. I at first used an executor but `is_in_logs` is much neater

Sjors:
Nice! I prefer `Co-Authored-By: Sjors Provoost <sjors@sprovoost.nl>` over an @ mention in the commit, because the latter can lead to lots of altcoin spam... (when they get around to cloning this)

darosior:
Great, Github even takes it into account.

darosior:
> Can you add a (optional) node public key argument to dumpcommitments? I don't know if this gets purged from the database 30 (?) days after channel closure, but regardless it's nice to be able to dump a commitment without having access to the full database.

You don't need access to the database to dump the commitment, just the `hsm_secret` from which we derive the node pubkey : that's why I removed it, it's redundant if you already have the hsm

darosior:
> Can you add a (optional) node public key argument to dumpcommitments? I don't know if this gets purged from the database 30 (?) days after channel closure, but regardless it's nice to be able to dump a commitment without having access to the full database.

Yes, you right this should be the default I think.

Sjors:
I'm able to produce the same commitment point with this PR as with my original tool.

darosior:
Polished, corrected, and rebased on master.

tools/hsmtools.c
```diff
@@ -0,0 +1,135 @@
...
+static bool ensure_hsm_secret_exists(int fd)
+{
+	if (fsync(fd) != 0) {
+		close(fd);
+		return false;
+	}
+	if (close(fd) != 0)
+		return false;
+
+	fd = open(".", O_RDONLY);
```
rustyrussell: This is fsyncing the wrong dir?  You want dirname(hsm_secret_path) here.

tools/hsmtools.c
```diff
@@ -19,6 +19,7 @@ static void show_usage(void)
 	printf("./hsmtools <method> [arguments]\n");
 	printf("methods:\n");
 	printf("	- decrypthsm <path/to/hsm_secret> <password>\n");
+	printf("	- encrypthsm <path/to/hsm_secret> <password>\n");
```
rustyrussell: Since the tool is call htmtools, we could just call these commands "decrypt" and "encrypt" perhaps?  And maybe "hsmtool" is better than "hsmtools" though that's marginal.

tools/hsmtools.c
```diff
@@ -44,25 +53,33 @@ static bool ensure_hsm_secret_exists(int fd)
...
+static void get_hsm_secret(struct secret *hsm_secret,
+                           const char *hsm_secret_path)
 {
 	int fd;
-	struct stat st;
-	struct secret key, hsm_secret;
+
+	fd = open(hsm_secret_path, O_RDONLY);
+	if (fd < 0)
+		errx(ERROR_HSM_FILE, "Could not open hsm_secret");
```
rustyrussell: Use err() not errx() here, which will tell them what the error was?

tools/hsmtools.c
```diff
@@ -44,25 +53,33 @@ static bool ensure_hsm_secret_exists(int fd)
...
+	if (!read_all(fd, hsm_secret, sizeof(*hsm_secret)))
+		errx(ERROR_HSM_FILE, "Could not read hsm_secret");
```
rustyrussell: Here too...

tools/hsmtools.c
darosior: Bad copy/paste!

tools/hsmtools.c
darosior: Yeah I wanted to add more tool*s* (https://github.com/ElementsProject/lightning/pull/3217) but I can make it singular ^^

tools/hsmtools.c
darosior: err.. I confused the err functions

darosior:
Corrected the `fsync` and the confusion between `err` functions, renamed encryption methods to `encrypt` and `decrypt`, and renamed the tool to `hsmtool`. Thanks for the review!

PR #3191, opened 2019-10-20
Make pedantic checks less fatal, fix an upgrade related null field access

### LINKED ISSUE #3189: consistent crash in listforwards
on 78c9d691117b6c27dfeffcfc575e2748cc25bebf

`lightning-cli listforwards` crashes my node every time after updating to master:

log:
https://jb55.com/s/842a84c59c6bea65.txt

cdecker:
Reverts the `assert` into a warning that we can watch for in the tests and logs, so we can fix them as they come up and then re-arm the asserts once we're happy that we have caught all of them. On postgres we'll still catch them given the field length check,

We also fix an upgrade related null field (which was the real cause for #3189), where we'd have a null field if a payment was forwarded inbetween commits 66a47d2 (which started tracking forwardings) and d901304 (which added the received_time column). Since this is a rare thing, and we can't really just remove a compat check later, I opted to just set these cases to 0.

Fixes #3189


jb55:
thanks, this fixed it

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3191/commits/037aad23fc84f24a56dc392cc20fa9d29091b95e

PR #2312, opened 2019-02-01
Add circleci support for building images and pushing to docker when t…

NicolasDorier:
The following environment variable which need to be configured in CircleCI
`DOCKERHUB_REPO` (for example `elementsproject/lightning`)
`DOCKERHUB_USER` (for example `nico`)
`DOCKERHUB_PASS` (for example `hello`)

Every time there is a new tagged release with the format like `v1.2.3`, this pull request should automatically do the following on CircleCI:
* Build a docker image for x64, tag it and push (`elementsproject/lightning:1.2.3-amd64`)
* Build a docker image for arm32v7, tag it and push (`elementsproject/lightning:1.2.3-arm32v7`)
* Build a docker image with `DEVELOPER=1`, tag it and push (`elementsproject/lightning:1.2.3-dev`)
* Build a manifest image, tag it and push (`elementsproject/lightning:1.2.3`)

On top of this, every times there is a commit on any branch, it will attempt to build without publishing. Ensuring you did not break the build process. 

It uses the credentials `nico` and `hello` to connect to dockerhub in this example.

As a result, you can run the image `elementsproject/lightning:1.2.3` on both arm7v32 and amd64.

This is a modified version of what I use for BTCPay, still need to test it to be sure it works properly on this repo.

The use of `docker_layer_caching` make builds way faster, especially when debugging. However this seems to be a [premium feature](https://circleci.com/docs/2.0/docker-layer-caching/). However if my memory is correct, this is still enabled for open source projects. (Though I think that for BTCPay, @kukks contacted them directly... my memory is failing me)

To test it:
* Fork your own (ie `nicolasdorier/lightning`)
* Create an account on circleci and attach it to your project
* In the project settings, configure the environment variable DOCKERHUB_REPO, DOCKERHUB_USER, DOCKERHUB_PASS . (I advise you to create a special account for this)
* Pull this PR
* Create a test tag (ie `v9.9.9`)
* Push to your repo
* Circle CI must run the build.

If you have a bug or need to fix something

* Remove the tag, (`git tag -d v9.9.9`)
* Make your change and commit
* Re-tag (`git tag -a v9.9.9 -m v9.9.9.9`)
* Re-push the tags (`git push nicolasdorier --force --tags`)

# Note to maintainers of this repository

After merging this:
* Create an account on circleci
* Attach to this repo (this will give some rights to [CircleCI to this repo](https://circleci.com/docs/2.0/gh-bb-integration/#overview))
[...]

PR #3197, opened 2019-10-22
add arm32v7/arm64v8 unit test to travis-ci

arowser:
Add arm32v7/arm64v8 cross build and unit test to travis-ci

Changelog-None

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -43,8 +43,51 @@ cat > pytest.ini << EOF
 addopts=-p no:logging --color=no --force-flaky
 EOF
 
+if [ "$TARGET_HOST" == "arm-linux-gnueabihf" ] || [ "$TARGET_HOST" == "arm-linux-gnueabihf" ]
+then
+    export QEMU_LD_PREFIX=/usr/"$TARGET_HOST"/
+    export MAKE_HOST="$TARGET_HOST"
+    export AR="$TARGET_HOST"-ar
+    export AS="$TARGET_HOST"-as
+    export CC="$TARGET_HOST"-gcc
+    export CXX="$TARGET_HOST"-g++
+    export LD="$TARGET_HOST"-ld
+    export STRIP="$TARGET_HOST"-strip
 
-if [ "$SOURCE_CHECK_ONLY" == "false" ]; then
+    wget -q https://zlib.net/zlib-1.2.11.tar.gz \
+    && tar xvf zlib-1.2.11.tar.gz \
+    && cd zlib-1.2.11 \
+    && ./configure --prefix="$QEMU_LD_PREFIX" \
+    && make \
+    && sudo make install 
+    cd .. && rm zlib-1.2.11.tar.gz && rm -rf zlib-1.2.11
+
+    wget -q https://www.sqlite.org/2018/sqlite-src-3260000.zip \
+    && unzip sqlite-src-3260000.zip \
+    && cd sqlite-src-3260000 \
+    && ./configure --disable-tcl --enable-static --disable-readline --disable-threadsafe --disable-load-extension --host="$TARGET_HOST" --prefix="$QEMU_LD_PREFIX" \
+    && make \
+    && sudo make install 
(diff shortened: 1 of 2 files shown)


NicolasDorier:
@arowser I would suggest instead to just try to build the docker arm32 and arm64 images. This test the same thing, without code duplication.

NicolasDorier:
if you want to run the test suite, I think you should add a build arg to the docker image for this.

An alternative is to isolate the build script for arm32 and 64 in a script file and call this from docker as well.

arowser:
@NicolasDorier  I agree with you, but I don't known if its a good idea that add a dependency to a docker image that maintain by me, I guess that the docker image should be make by @cdecker or @rustyrussell

NicolasDorier:
@arowser but they are already being maintained in https://github.com/ElementsProject/lightning/tree/master/contrib

NicolasDorier:
If it can help, I tried to do it in CI using Circle on https://github.com/ElementsProject/lightning/pull/2312/files but we closed the PR because CircleCI is too greedy on github permission.

.travis/build.sh
```diff
@@ -43,8 +43,51 @@ cat > pytest.ini << EOF
...
+    export STRIP="$TARGET_HOST"-strip
 
-if [ "$SOURCE_CHECK_ONLY" == "false" ]; then
+    wget -q https://zlib.net/zlib-1.2.11.tar.gz \
+    && tar xvf zlib-1.2.11.tar.gz \
+    && cd zlib-1.2.11 \
+    && ./configure --prefix="$QEMU_LD_PREFIX" \
+    && make \
+    && sudo make install 
+    cd .. && rm zlib-1.2.11.tar.gz && rm -rf zlib-1.2.11
```
cdecker: This sort of chaining is not necessary since we are using `set -e` in the script header:

```
      -e  Exit immediately if a command exits with a non-zero status.
```

.travis/build.sh
```diff
@@ -43,8 +43,51 @@ cat > pytest.ini << EOF
 addopts=-p no:logging --color=no --force-flaky
 EOF
 
+if [ "$TARGET_HOST" == "arm-linux-gnueabihf" ] || [ "$TARGET_HOST" == "arm-linux-gnueabihf" ]
```
cdecker: Why are we ORing two identical checks here? My guess is the second should be `aarch64`?

.travis/build.sh
```diff
@@ -43,8 +43,51 @@ cat > pytest.ini << EOF
...
+    && sudo make install 
+    cd .. && rm sqlite-src-3260000.zip && rm -rf sqlite-src-3260000
+
+    wget -q https://gmplib.org/download/gmp/gmp-6.1.2.tar.xz \
+    && tar xvf gmp-6.1.2.tar.xz \
+    && cd gmp-6.1.2 \
+    && ./configure --disable-assembly --prefix="$QEMU_LD_PREFIX" --host="$TARGET_HOST" \
+    && make \
+    && sudo make install 
+    cd .. && rm gmp-6.1.2.tar.xz && rm -rf gmp-6.1.2
```
cdecker: If you checkout the dependencies in an architecture specific directory in the `dependencies` subdirectory we can cache them across builds, speeding things up and saving us a whole lot of extra CPU cycles :wink:

NicolasDorier:
> @NicolasDorier I agree with you, but I don't known if its a good idea that add a dependency to a docker image that maintain by me, I guess that the docker image should be make by @cdecker or @rustyrussell

I don't see what you mean. This would not be a dependency on an image maintained by you.
You just build the whole thing inside docker, no dependency involve.

> In the past Travis has caused us a couple of issues, e.g., as timeouts pulling images, so the less we rely on Docker's / Travis' infrastructure the better it is IMHO.

This PR is actually adding code that only the travis system run.

With the alternative, using docker build, we eliminate the need to maintain this travis specific code, and you increase test code coverage. Docker images are well used, so it makes sense to make sure their build are not broken.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3197/commits/c836c3e44703cf1894381792e600e2493fb3fab7

PR #3202, opened 2019-10-24
config file: tolerate trailing whitespaces and comments. Fix #3201

### LINKED ISSUE #3201: Config file should tolerate pending whitespace and comments
The following configs cause lightningd to fail to launch:

```
# Trailing space after 5000
fee-base=5000 
```

```
fee-base=5000 # 5 satoshi
```

gorazdko:
## Status
WIP

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -719,6 +719,49 @@ static void config_log_stderr_exit(const char *fmt, ...)
 }
 
 /**
+ * Remove comments and white-spaces from the i-th line.
+ * But do not allow having # in bitcoin-rpcpassword.
+*/
+static void opt_clean_line_of_comments_and_spaces(char **lines, u16 i)
+{
+	bool ispassword = (strstr(lines[i], "bitcoin-rpcpassword") != NULL);
+
+	char *pos = strchr(lines[i], '#');
+	if (pos != NULL) {
+		/* There's a comment in this line. */
+
+		if (ispassword == true)
+			err(1, "error reading configuration file: using # in bitcoin-rpcpassword can be ambiguous and should be avoided");
+
+		u16 index = (u16)(pos - lines[i]);
+		/* Remove the comment. */
+		lines[i][index] = 0;
+	}
+
+	/* Remove trailing white-spaces. */
+	for (u16 j = strlen(lines[i]) - 1; j >= 0; --j) {
+		if (isspace(lines[i][j]) != 0) {
+		lines[i][j] = 0;
+		}
+		else
+			break;
+	}
+
+	/* Remove initial white-spaces. */
+	u16 k = 0;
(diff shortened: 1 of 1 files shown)


lightningd/options.c
```diff
@@ -759,6 +759,23 @@ static void opt_parse_from_config(struct lightningd *ld, bool early)
 		else {
 			/* Only valid forms are "foo" and "foo=bar" */
 			all_args[i] = tal_fmt(all_args, "--%s", lines[i]);
+
+			char *pos = strchr(all_args[i], '#');
```
arowser: Error in follow config file:
```
   bitcoin-rpcpassword=xxx#pass  # xxx
```
Maybe the fix should add to ccan/ccan/opt/parse.c

darosior: ccan fixes should be integrated into ccan directly https://github.com/rustyrussell/ccan/

gorazdko: tnx for suggestions, i'll find a cleaner approach.

Core doesnt allow such passwords, so neither should we:

> bitcoin-rpcpassword=xxx#pass  # xxx

(16115)gorazd@gorazd-MS-7C37:~/Projects/bitcoin$ ./src/bitcoind
Error: Error reading configuration file: parse error on line 5, using # in rpcpassword can be ambiguous and should be avoided

gorazdko: tnx ill find a cleaner approach

i see now how Core doesnt allow neither # in passwords nor trailing comments after passwords:

> Error: Error reading configuration file: parse error on line 6, using # in rpcpassword can be ambiguous and should be avoided

rustyrussell:
If you start down this path, how do you use `#` or ` ` in a name, etc?

No, I think we have to stick with exact usage; we just need to make sure that the messages are clear if it's accidental.

gorazdko:
I wanted to make the parser compatible with Core (remove trailing comments, initial and trailing white-spaces which this PR does), but yes, i see this brings more bad things because Core too has unhandled corner cases:

* they allow trailing comments everywhere except for the rpcpassword where they throw an error, everywhere else # is considered a comment and cannot be part of a name (no warnings thrown)
  * they allow trailing comments at torpassword without a warning though
* they trim the trailing spaces which could be a part of the password



* therefore the only thing that could be done here is to throw:
  * error message:    "config file line X: '#' can be used only at the beginning of the line to denote a comment"
  * error message:    "config file line X: line must not begin with a space "
  * warning message:  "config file line X: trailing spaces are not recommended"
  * also fix the line counting:

  ```
  fee-base=5000


  line4

  ```
  * lightning config file line 2: line4: unrecognized option


I can force push these changes here if they are reasonable, otherwise the issue can be closed.

cdecker:
I'm afraid it's going to be a NACK from me as well. Starting with these exceptions is a never-ending rabbit hole.

PR #3212, opened 2019-10-26
Restore payment amount fuzzing

darosior:
[It got stuck into the red part of a diff](https://github.com/ElementsProject/lightning/pull/2215/files#diff-fc86e59a6aa3bf284fbaf6be349102d7L536)

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -820,8 +820,6 @@ static struct command_result *add_shadow_route(struct command *cmd,
 	const jsmntok_t *chan, *best = NULL;
 	size_t i;
 	u64 sample = 0;
-	/* FIXME: Use route_info's cltv_expiry_delta member instead */
-	u32 cltv_fuzz;
 	struct route_info *route = tal_arr(NULL, struct route_info, 1);
 
 	json_for_each_arr(i, chan, channels) {
@@ -835,9 +833,9 @@ static struct command_result *add_shadow_route(struct command *cmd,
 				continue;
 
 			/* Don't use if total would exceed 1/4 of our time allowance. */
-			json_to_number(buf, json_get_member(buf, chan, "delay"),
-			               &cltv_fuzz);
-			if ((pc->final_cltv + cltv_fuzz) * 4 > pc->maxdelay)
+			json_to_u16(buf, json_get_member(buf, chan, "delay"),
+			            &route[0].cltv_expiry_delta);
+			if ((pc->final_cltv + route[0].cltv_expiry_delta) * 4 > pc->maxdelay)
 				continue;
 
 			json_to_number(buf, json_get_member(buf, chan, "base_fee_millisatoshi"),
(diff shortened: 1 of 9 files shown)


plugins/pay.c
```diff
@@ -1061,6 +1067,11 @@ static struct command_result *json_pay(struct command *cmd,
 		pc->msat = *msat;
 	}
 
+	if (*fuzzpercent > 0.0) {
```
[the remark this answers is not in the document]
niftynei: though there's a comment on the old version of fuzzing that mentions that the total amount doesn't exceed the allowed limit for a payment. i wonder if this is still implemented also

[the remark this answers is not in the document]
darosior: Hmm yeah I wondered to make it a `do {} while (overpayment > maxfeepercent * amount)`. But the default `fuzzpercent <= maxpercent` and since `fuzzpercent` and `maxpercent` are optional, I think someone providing them knows what they doing ?

rustyrussell:
I wonder if we should, instead, use the shadow route to fuzz?  We use that to fuzz the cltv, should we use that to fuzz the amount?

I'd like to generally make our shadow route stronger, and this would be a good start...

darosior:
> I wonder if we should, instead, use the shadow route to fuzz? We use that to fuzz the cltv, should we use that to fuzz the amount?

Thinking about it, I said myself it would be harder to test : this means it might be harder to analyze as well, I'll give it a try!

darosior:
Did it through shadow routing, added a FIXME in `pay` because the `cltv` delta in `route_info` is an `u16` (why?) and I didn't want to go down the rabbit hole to change it in this PR. Added an other one in the shadow routing test as testing the cltv delta needs the `pay` command to return it and it seems outside the scope of this PR.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- JSON API: The `pay` command now accepts a new parameter, `fuzzpercent` used to fuzz the fees for route calculation.
```
rustyrussell: I've been looking at this metric, and it doesn't really make sense.  Fees are so tiny, that 5% is hardly a change at all, and seems to contribute almost nothing to route diversity :(   But we don't want to increase it so much that we breach 'maxfeepercent' which people really care about!

I'd be happier with moving the routing algo towards randomizing the route so that it spends up to min(maxfeepercent / 2, best-fee * 2).  i.e. if we're well under our `maxfeepercent`, then you can spend twice as much as the best case on route obfuscation.

Claudia Diaz made some pointed remarks about our route selection in Lightning, and I'd really like to improve it.

Anyway, this is an aside: I'd prefer to avoid exposing this for now.

plugins/pay.c
```diff
@@ -824,25 +824,32 @@ static struct command_result *add_shadow_route(struct command *cmd,
 	const jsmntok_t *chan, *best = NULL;
 	size_t i;
 	u64 sample = 0;
-	u32 cltv, best_cltv;
+	/* FIXME: Use route_info's cltv_expiry_delta member instead */
```
rustyrussell: The gossip protocol restricts that to a u16 btw...

plugins/pay.c
```diff
@@ -103,6 +103,9 @@ struct pay_command {
 	/* Any remaining routehints to try. */
 	struct route_info **routehints;
 
+	/* Disable the use of shadow route ? */
+	double use_shadow;
+
```
rustyrussell: Can we make this dev-only please?  I don't want normal people to disable shadow routing, it's a terrible idea.

plugins/pay.c
darosior: Didn't notice that subtlety, thanks

plugins/pay.c
darosior: Will do

CHANGELOG.md
darosior: Ok, will remove this commit then.

darosior: > I'd be happier with moving the routing algo towards randomizing the route so that it spends up to min(maxfeepercent / 2, best-fee * 2).

Is this part of the reason why you proposed higher fees, to have more chances to fuzz route computation ?

darosior:
- Dropped the first commit (`fuzzpercent` parameter).
- Added a json helper (basically a copy of the above one, `json_to_number`) for `u16` and used it for cltv (that fixed the previously introduced FIXME).
- Made `use_shadow` dev-only, which introduced a some changes since it's a parameter of a non-dev command. Users are now shadow-safe !
(and rebased on master)

Thanks for the review !

darosior:
Ah I forgot the CHANGELOG trick, will do now

Issue #3080, opened 2019-09-21
Failed to make install on FreeBSD


reablaz:
Failed to install on FreeBSD 12.0-RELEASE. I was running c-lightning  v0.7.1-30-ga6516f4 and decided to upgrade to the latest version (v0.7.2.1) from github yesterday. with no luck. 
To ./configure this time, there were new additional dependencies required: py36-mako and gettext, so i've installed them. After that, ./configure & gmake executed fine, but failed on gmake install.

```
_/usr/bin/ld: error: undefined symbol: __start_xautodata_db_backends                       
 >>> referenced by db.c:662 (wallet/db.c:662)                                                
 >>>               wallet/db.o:(db_config_find)                                                                                                                                                                                                
                                                                                                                                                                                                                                              
/usr/bin/ld: error: undefined symbol: __stop_xautodata_db_backends                                                                                                                                          
 >>> referenced by db.c:662 (wallet/db.c:662)                                                                                                                                                                                                  
 >>>               wallet/db.o:(db_config_find)                                                                                                                                                                                                
cc: error: linker command failed with exit code 1 (use -v to see invocation)                                                                                                                                                                  
gmake: *** [Makefile:390: lightningd/lightningd] Error 1_ 
```

![image](https://user-images.githubusercontent.com/3844596/65368383-1e43e480-dc49-11e9-8fef-e32cf74255a8.png)

reablaz:
gmake actually failed as well with error:

/usr/bin/ld: error: too many errors emitted, stopping now (use -error-limit=0 to see all errors)
cc: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[1]: *** [<builtin>: tools/test/gen_test.c.tmp] Error 1            
gmake: *** [tools/test/Makefile:40: tools/test/gen_test.c] Error 1
rm external/libwally-core/src/secp256k1/libsecp256k1.la tools/test/gen_test.c.tmp.c

fiatjaf:
I see these same errors on Ubuntu at `v0.7.2`. `master` compiles fine.

laanwj:
Same problem here on FreeBSD 12.0 with master (as of a4204226b46a31b3eb637967fb5c18d87d1cd9bb).

I bisected the problem to the commit:
```
acedcc593f52b17936843496336eee6c04bf8a9e is the first bad commit
commit acedcc593f52b17936843496336eee6c04bf8a9e
Author: Christian Decker <decker.christian@gmail.com>
[...]

PR #3214, opened 2019-10-28
configure: fix FreeBSD which has sqlite3 in /usr/local

rustyrussell:
Fixes: #3080


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3214/commits/80e05d43442131fde2267c0cc08210a52d968980

Issue #2661, opened 2019-05-22
Invalid signature on reconnection eclair/c-lightning


pm47:
### Issue and Steps to Reproduce

Following an unrelated memory issue with eclair causing it to grind to a halt, I had to forcekill the node, and on restart I saw a few `Bad commit_sig` from several c-lightning nodes. After investigating I think this may have revealed a corner case problem on c-lightning's side related to `update_fee`.

Here is what happened (taken from the logs below):

```
 eclair           c-lightning
   |                   |
   |  (idle, no htlc)  |
   |                   |
   |<------ fee1 ------|
   |<------ sig -------|
   |------ rev ------->|
   |------- sig ------>|
   |<------ fee2 ------|
   |<------ sig -------|
   |<----- rev --------|
   |-rev-> ?           |
   |                   |
   |   (reconnection)  |
   |                   |
   |------- rev ------>|
   |------- sig ------>|
   |<------ err -------| <- Bad commit_sig feerate=fee1
   |        ...        |
```

I think eclair is right to have been using feerate=fee2 after reconnection.

```
2019-05-21 18:55:39,565 INFO IN msg=UpdateFee(edc7ef107af445a433be4b2b992ca251b64df6bcb43c8c1306104cac783fa277,41186)
2019-05-21 18:55:39,565 INFO IN msg=CommitSig(edc7ef107af445a433be4b2b992ca251b64df6bcb43c8c1306104cac783fa277,3bdbd03cccbf8bf1f14909ec9bc2de9e4e7d36ccc1a14d643ff0e84e5111c9fe314c2b270241044708077b3c02029ea7589f2f24734b286811afbc140667bd46,List())
2019-05-21 18:56:15,680 INFO built local commit number=1868 htlc_in= htlc_out= feeratePerKw=41186 txid=b1155d1d24358f1351cf32dabf04dcb2ff583ff8515d819991916d183167a251 tx=0200000001edc7ef107af445a433be4b2b992ca251b64df6bcb43c8c1306104cac783fa2770000000000bafd188001466de400000000001600140fbea428c4ff928c2a0090ba0842499b8471b5f515231220
[...]

PR #3215, opened 2019-10-28
Fix "Bad commit_sig signature" error.

rustyrussell:
Thanks to DarthCoin and Javier for reporting this; simple to diagnose once I reproduced.

@cdecker had reports of similar issues with LND under stress, which we assumed was their end.  Oops, it was us!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ Note: You should always set `allow-deprecated-apis=false` to test for changes.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


darosior:
I think this fixes https://github.com/ElementsProject/lightning/issues/2661

darosior:
ACK 907364bf2d79b0f106594ae7c7a2344ce72b6274

JavierRSobrino:
I am one of the affected, if you require more information that I can provide, just ask.

PR #3217, opened 2019-10-28
hsmtools: add a tool to recover a `to_remote` output from peer unilateral close

darosior:
This builds upon https://github.com/ElementsProject/lightning/pull/3186 (starts at https://github.com/ElementsProject/lightning/commit/02a1934267bbee8bd09721e566e827906f4473a5).

I suffered a power outage some weeks ago and lost the `dbid` of many channels from the database, the remote peer has then unilaterally closed the channel : I was not able to recover the private key from the `to_remote` output of its commitment transaction without the `dbid`. Since I've seen others in the same situation (this has been in the news, too...), here is a tool which derives `n` payment keys for every `dbid` and compares it to the pubkeyhash output in the witness program (we just take a bech32 P2WPKH address and derive the goal hash from it).

### DIFF
--- a/tools/hsmtools.c
+++ b/tools/hsmtools.c
@@ -1,0 +1,135 @@
+#include <bitcoin/privkey.h>
+#include <ccan/str/str.h>
+#include <ccan/err/err.h>
+#include <ccan/noerr/noerr.h>
+#include <ccan/read_write_all/read_write_all.h>
+#include <common/utils.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <sodium.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#define ERROR_HSM_FILE errno
+#define ERROR_USAGE 2
+#define ERROR_LIBSODIUM 3
+
+static void show_usage(void)
+{
+	printf("./hsmtools <method> [arguments]\n");
+	printf("methods:\n");
+	printf("	- decrypthsm <path/to/hsm_secret> <password>\n");
+	exit(0);
+}
+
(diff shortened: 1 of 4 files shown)


tools/hsmtools.c
```diff
@@ -282,6 +286,119 @@ static int dump_commitments_infos(struct node_id *node_id, u64 channel_id,
...
+			return 0;
+		}
+
+		/* Otherwise derive payment keys up to <depth> to check */
+		if (!derive_shaseed(&channel_seed, &shaseed))
+			errx(ERROR_KEYDERIV, "Could not derive shaseed for dbid %"PRIu64
+			                     " and channel seed %s.", dbid,
+			                     type_to_string(tmpctx,
+			                                    struct secret, &channel_seed));
+		for (u64 i = 0; i < depth; i++) {
```
darosior: Hmm I think I'm wrong on this branch since I can't manage to recover the payment key without `option_static_remotekey`....

darosior: Can't find what's wrong though...

darosior: I'm wrong on this branch since I can't manage to recover the payment key without `option_static_remotekey`.... Can't manage to find what's wrong though..

darosior: Ok after hours of testing I realized that we actually derive the `to_remote`'s pubkey from OUR commitment point, which means we can't recover the `to_remote` outputs to us without knowing the remote peer's `per_commit_point`..................

darosior:
Closing this, since it doesn't make sense anymore...

PR #3218, opened 2019-10-28
pyln: Migrate testing utilities and fixtures into pyln-testing

cdecker:
I've been planning to do this for a while now, but I didn't know how clever to get, so this is the blunt version that just moves all the pieces over while keeping the surface stable.

The goal is to split out the testing framework, including fixtures and utilities to setup and manage nodes, into its own package, so we can reuse them in other related projects. Currently we have copy-paste versions of this code in https://github.com/cdecker/lnet and https://github.com/cdecker/lightning-integration, and some people have expressed interest in setting up tests for their plugins in https://github.com/lightningd/plugins as well.

The copy in https://github.com/cdecker/lightning-integration is likely the hardest to retro-fit to use pyln-testing, since it manages multiple parametrizable implementations, but I plan to add runners for those other projects eventually as well. One step at a time :-)


contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -5,7 +5,7 @@
 import socket
 import warnings
 
-__version__ = "0.0.7.4"
+__version__ = "0.7.4"
```
darosior: Why not `0.7.3` to be in sync with the release ? `0.7.3 > 0.0.7.4` anyway

darosior:
Running
```bash
PYTHONPATH=`pwd`/contrib/pyln-client:`pwd`/contrib/pyln-testing:$PYTHONPATH TEST_DEBUG=1 DEVELOPER=1 VALGRIND=1 pytest tests/ -v --timeout=550 --timeout_method=thread -p no:logging -n=24
```
Gives me an error on the first test (`test_deprecated_closing_compat`) because
```python
with pytest.raises(RpcError, match=r"......."):
     the_call()
```
Does not seem to catch the error.

Looking forward to use `pyln-testing` for plugins ! :-)

cdecker:
Yep, should have marked this as WIP since it clearly has some edge-cases. For example outside of this repo we want to use the `lightningd` from the `$PATH` while here we explicitly want the one in `lightningd/lightningd`, so I'll need to create a way to customize the nodes we run.

cdecker:
> Running
> 
> ```shell
> PYTHONPATH=`pwd`/contrib/pyln-client:`pwd`/contrib/pyln-testing:$PYTHONPATH TEST_DEBUG=1 DEVELOPER=1 VALGRIND=1 pytest tests/ -v --timeout=550 --timeout_method=thread -p no:logging -n=24
> ```
> 
> Gives me an error on the first test (`test_deprecated_closing_compat`) because
> 
> ```python
> with pytest.raises(RpcError, match=r"......."):
>      the_call()
> ```
> 
> Does not seem to catch the error.
> 
> Looking forward to use `pyln-testing` for plugins ! :-)

I'm afraid I was not able to reproduce this issue in any of my runs. Is there anything special about your setup? Does `make pytest` work as expected (you can instrument `make pytest` using a number of env vars such as `PYTEST_PAR` and `TEST_DB_PROVIDER` if that's why your running the tests manually)?

PR #3223, opened 2019-10-29
Display `close_to` address in `listpeers`

niftynei:
Make close-to address more visible by printing in `listpeers` channels.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -14,6 +14,7 @@
 #include <ccan/take/take.h>
 #include <ccan/tal/str/str.h>
 #include <channeld/gen_channel_wire.h>
+#include <common/addr.h>
 #include <common/dev_disconnect.h>
 #include <common/features.h>
 #include <common/htlc_trim.h>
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
...
+### Deprecated
+
+Note: You should always set `allow-deprecated-apis=false` to test for
+changes.
+
+### Removed
+
+### Fixed
+
+
```
rustyrussell: Needs rebase to use Changelog-Added: instead.

lightningd/peer_control.c
```diff
@@ -619,6 +620,16 @@ static void json_add_channel(struct lightningd *ld,
 	json_add_string(response, "channel_id",
 			type_to_string(tmpctx, struct channel_id, &cid));
 	json_add_txid(response, "funding_txid", &channel->funding_txid);
+
+	if (channel->shutdown_scriptpubkey[LOCAL]) {
+		json_add_string(response, "close_to_addr",
+				encode_scriptpubkey_to_addr(tmpctx,
+					get_chainparams(ld),
+					channel->shutdown_scriptpubkey[LOCAL]));
```
rustyrussell: Should we handle encode_scriptpubkey_to_addr returning NULL?  The classic would be not to add the field if it's NULL...

niftynei: good idea. updated.

PR #3224, opened 2019-10-29
Warnings: if behind blockchain, don't show cannot afford

niftynei:
If you're replaying or syncing with the blockchain, show that error
instead of 'cannot afford', in the case of not having enough utxos
to pay for a transaction. This is the 'more correct' error to show, as
there's a chance that the funds you're expecting to spend are in the
portion of the blockchain that hasn't been synced yet.

### DIFF
--- a/common/wallet_tx.c
+++ b/common/wallet_tx.c
@@ -107,9 +107,16 @@ static struct command_result *check_amount(const struct wallet_tx *wtx,
 					   struct amount_sat amount)
 {
 	if (tal_count(wtx->utxos) == 0) {
+		/* Since it's possible the lack of utxos is because we haven't finished
+		 * syncing yet, report a sync timing error first */
+		if (!topology_synced(wtx->cmd->ld->topology))
+			return command_fail(wtx->cmd, FUNDING_STILL_SYNCING_BITCOIN,
+					    "Still syncing with bitcoin network");
+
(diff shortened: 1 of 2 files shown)


rustyrussell:
Don't understand the test change?

cdecker:
> Don't understand the test change?

The change to the test is pretty much unrelated if I'm not mistaken. `line_graph` overfunds the nodes, so they should have sufficient funds for the `fundchannel` further down in the test. The removed lines were just adding even more funds and confirming them.

ACK https://github.com/ElementsProject/lightning/pull/3224/commits/66146229da75cbf6be3228c39d3fef7da4c0e0d6

niftynei:
> Don't understand the test change?

Breaking out `fundchannel` into a plugin uncovered a race condition on start between replaying the last 100 blocks where we'd hit a "Cannot fund channel" error instead of "Not synced yet", 304 error occasionally.  This PR fixes the race by rejecting any failed attempts to fund a wallet while a sync is happening (because we assume that there might be funds in the un-played blocks), so we consistently will get a `304` error on a `fundchannel` attempt now (either at the funding check stage or at the 'fundchannel' stage).

The better fixup would be to check both that `txprepare` throws 304 when there's no funds in the wallet while we're behind and that `fundchannel_start` also throws a 304 when we're behind (because we don't allow fundchannels while behind). Will fix.

PR #3228, opened 2019-10-31
pytest: fix flaky 'Bad gossip' errors

rustyrussell:
One real test fix, and a rework of our slightly-broken "push this msg through despite timestamp_filter" option which was causing many more "Bad gossip" order messages.


cdecker:
The reduction in surface by passing in the `push` flag is a nice change, allows us to spread the logic out far less.

You're missing a Changelog entry in one of your commit messages, or opt-out using `Changelog-None` in one of the commit messages. The bot is currently set to `pending` rather than `failure` for that check, but it'd be nice to get used to adding those :-)

ACK https://github.com/ElementsProject/lightning/pull/3228/commits/0dcc5f3e13c46bfc5ba07de3dbbf9d95ce62c77a

PR #3233, opened 2019-11-01
Fix bad htlc add during reconnect

rustyrussell:
As found by random breakage on Travis on #3221 !

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -1178,22 +1178,23 @@ bool channel_force_htlcs(struct channel *channel,
 	struct htlc_map_iter it;
 
 	if (tal_count(hstates) != tal_count(htlcs)) {
-		status_debug("#hstates %zu != #htlcs %zu",
+		status_broken("#hstates %zu != #htlcs %zu",
 			     tal_count(hstates), tal_count(htlcs));
 		return false;
 	}
 
 	if (tal_count(fulfilled) != tal_count(fulfilled_sides)) {
-		status_debug("#fulfilled sides %zu != #fulfilled %zu",
+		status_broken("#fulfilled sides %zu != #fulfilled %zu",
 			     tal_count(fulfilled_sides), tal_count(fulfilled));
 		return false;
 	}
 
 	if (tal_count(failed) != tal_count(failed_sides)) {
-		status_debug("#failed sides %zu != #failed %zu",
+		status_broken("#failed sides %zu != #failed %zu",
 			     tal_count(failed_sides), tal_count(failed));
 		return false;
 	}
+
 	for (i = 0; i < tal_count(htlcs); i++) {
 		enum channel_add_err e;
 		struct htlc *htlc;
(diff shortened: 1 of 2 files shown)


tests/test_connection.py
```diff
@@ -2196,3 +2196,32 @@ def test_feerate_stress(node_factory, executor):
...
+                                                                  '-WIRE_COMMITMENT_SIGNED']}])
+
+        scid12 = l1.get_channel_scid(l2)
+        routel2l1 = [{'msatoshi': '10000msat', 'id': l1.info['id'], 'delay': 5, 'channel': scid12}]
+
+        # Get invoice from l1 to pay.
+        payhash1 = l1.rpc.invoice(10000, "invoice", "invoice")['payment_hash']
+    
+        # Start balancing payment.
+        fut = executor.submit(l1.pay, l2, 10**9 // 2)
```
niftynei: `check-python` is complaining about `fut` being unused here

tests/test_connection.py
```diff
@@ -2196,3 +2196,32 @@ def test_feerate_stress(node_factory, executor):
...
+                                                  {'may_reconnect': True,
+                                                   'disconnect': ['=WIRE_UPDATE_ADD_HTLC',
+                                                                  '-WIRE_COMMITMENT_SIGNED']}])
+
+        scid12 = l1.get_channel_scid(l2)
+        routel2l1 = [{'msatoshi': '10000msat', 'id': l1.info['id'], 'delay': 5, 'channel': scid12}]
+
+        # Get invoice from l1 to pay.
+        payhash1 = l1.rpc.invoice(10000, "invoice", "invoice")['payment_hash']
+
```
niftynei: `check-python` is complaining about there being whitespace in this line

channeld/full_channel.c
```diff
@@ -1316,16 +1351,21 @@ bool channel_force_htlcs(struct channel *channel,
 			htlc->failed_scid = NULL;
 	}
 
-	/* Now adjust balances.  The balance never goes negative, because
-	 * we do them in id order. */
+	/* Add giant offset so we never go negative here. */
+	if (!offset_balances(channel))
```
niftynei: i assume we check an HTLC's validity (in terms of amount) before this point?

rustyrussell: This code is only used when we're restoring HTLCs lightningd told us about.  So presumably they've been fully checked.

rustyrussell:
To be clear: htlc restore amount checking is simply a sanity check.

To do the sanity check properly, we'd have to replay each HTLC state *in order*: once one has been fulfilled, for example, another can be spent.  It's not even enough to add them in creation order, we need every balance adjustment in order.  We don't have enough information to do that.

rustyrussell:
flake8 now happy, trivial rebase.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3233/commits/eb012c036e144ac744ae0c501087aac15c09d266

PR #3236, opened 2019-11-01
dev: add option flag for specifying temporary channel id

niftynei:
--dev-force-tmp-channel-id flag takes a 64-character hex string
to use as the temporary channel id. Useful for spec tests where we want
to test the 'opener's side of things.

### DIFF
--- a/openingd/openingd.c
+++ b/openingd/openingd.c
@@ -15,6 +15,7 @@
 #include <ccan/breakpoint/breakpoint.h>
 #include <ccan/cast/cast.h>
 #include <ccan/fdpass/fdpass.h>
+#include <ccan/mem/mem.h>
 #include <ccan/tal/str/str.h>
 #include <common/crypto_sync.h>
 #include <common/derive_basepoints.h>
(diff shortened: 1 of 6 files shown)


rustyrussell:
Ah.  I think we should use deterministic channel ids.  I went to get the URL to the archive only to find I sent it to the lightning-dev-bounces email by accident!

I've fwd it to you, please consider.

niftynei:
Oh rad. Just saw your email, thanks.

That works well for v2 opens but any node-initiated open tests of the
existing protocol need something

openingd/openingd.c
```diff
@@ -463,6 +469,12 @@ static bool setup_channel_funder(struct state *state)
 	 * until we know their funding_pubkey) */
 	temporary_channel_id(&state->channel_id);
 
+#if DEVELOPER
+	/* --dev-force-tmp-channel-id specified */
+	if (dev_force_tmp_channel_id)
+		memcpy(&state->channel_id, dev_force_tmp_channel_id,
+		       sizeof(*dev_force_tmp_channel_id));
```
rustyrussell: `state->channel_id = *dev_force_tmp_channel_id;`

People often forget that struct assignment works (whereas comparison doesn't) in C.

niftynei: oh. good to know.

rustyrussell:
Trivial fix for NULL-deref when !DEVELOPER.

PR #3237, opened 2019-11-02
Fix python 3.4 syntax error in sql-rewrite.py

sgeisler:
On python 3.4.2 having kwargs after unpacking a dict in a function call seems to be a syntax error. I know it's an old python version but if it doesn't break anything else it might be worth changing.

2019-11-02-125824_5760x1080_scrot

### DIFF
--- a/devtools/sql-rewrite.py
+++ b/devtools/sql-rewrite.py
@@ -12,7 +12,7 @@ DEBUG = False
 def eprint(*args, **kwargs):
     if not DEBUG:
         return
-    print(*args, **kwargs, file=sys.stderr)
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3237/commits/75c2a038f6e8154308f16d4921dde18e4273a86c

sgeisler:
Sorry, had to force push because of a trailing white space :hides:

sgeisler:
What should I do about the failed test? Seems a bit random ... maybe just rerun?

```
DEBUG:root:Found 'Bad gossip order from (?!error)' in logs
```

PR #3238, opened 2019-11-02
changelog: Add a tool to extract changelog entries from the commits

cdecker:
Since the `CHANGELOG.md` file is a major source for merge conflicts I decided
to build a tiny tool that generates the entries for the changelog
automatically from the commit messages.

Changelog-None

### DIFF
--- a/devtools/changelog.py
+++ b/devtools/changelog.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+from collections import namedtuple
+import re
+import sys
+import shlex
+import subprocess
+import requests
+from mako.template import Template
+import argparse
+from datetime import datetime
+
+# What sections do we support in the changelog:
+sections = [
+    'added',
+    'changed',
+    'deprecated',
+    'fixed',
+    'removed',
+    'security',
+]
+
+repo = 'ElementsProject/lightning'
+
+Entry = namedtuple("Entry", ["commit", "pullreq", "content", "section"])
+
+
+def git(cmd):
+    cmd = shlex.split(cmd)
+    out = subprocess.check_output(['git'] + cmd)
+    return out.decode('UTF-8')
+
+
+def get_commit_range():
+    """Find a commit range that we should collect the CHANGELOG for.
+    """
+    description = git("describe")
+    version = description.split('-')[0]
+    return "{version}..master".format(version=version)
+
+
+def get_log_entries(commitrange):
+    commit = None
+    logs = git("log {commitrange}".format(commitrange=commitrange))
+    entries = []
+
+    for l in logs.split('\n'):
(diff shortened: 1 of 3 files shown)


doc/STYLE.md
```diff
@@ -160,3 +160,40 @@ Try to make a single change at a time.  It's tempting to do "drive-by"
...
 `/* FIXME: ...*/` comment and move on.
+
+## Github Workflows
+
+We have adopted a number of workflows to facilitate the development of
+c-lightning, and to make things more pleasant for contributors.
+
+### Changelog Entries in Commit Messages
+
+We are maintaining a chanelog in the top-level directory of thi
```
darosior: this

devtools/changelog.py
```diff
@@ -0,0 +1,143 @@
...
+    entries = []
+
+    for l in logs.split('\n'):
+        m = re.match(r'^commit ([A-Fa-f0-9]{40})$', l)
+        if m:
+            commit = m.group(1)
+
+        m = re.match(
+            r'^\s+Changelog-({}): (.*)$'.format("|".join(sections)), l)
+        if not m:
```
rustyrussell: Hmm, would be nice to have a warning if I misspell the section name after Changelog-?

cdecker:
> Couldn't we just opt-out by default instead of adding `Changelog-None` in a commit ?

With the bot currently setting the check status to pending it isn't blocking the merge, so if we keep it this way it'll just be a tiny little reminder. If we were to set it to failure later it'll require admin privileges to merge anyway. So currently it's not really all that binding, however there is value in making the requirement by default, otherwise this mechanism of tracking changes will not get used.

devtools/changelog.py
cdecker: Would it make more sense to have that check in the bot? At the time we are generating the changelog using this script it's likely too late to actually go and fix it since the commit was merged into master and presumably buried under all the other changes that came afterwards.

rustyrussell: Ack!

rustyrussell:
> Couldn't we just opt-out by default instead of adding `Changelog-None` in a commit ?

Really prefer it in PR, instead, and just have it suppress the bot.  "There's no changelog for this" does not belong with a specific commit; in the long term it's simply noise.

rustyrussell:
Is this ready to be made non-draft?  Would also be a good time to make the bot recognize the Changelog-none: line in the PR description :)

cdecker:
> Is this ready to be made non-draft? Would also be a good time to make the bot recognize the Changelog-none: line in the PR description :)

I'll implement this today and mark this as non-draft asap.

cdecker:
Ok, the bot now checks the PR description (first post on the top) for `Changelog-None` opt-out (case-insensitive), in addition to the commit logs.

This should be ready to go. Ping @rustyrussell

PR #3246, opened 2019-11-06
sphinx: fix potential data leak.

rustyrussell:
[ Based on #3245 to keep Travis happy: just look at last commit ]

https://github.com/lightningnetwork/lightning-rfc/pull/697
https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002288.html

Reported-by: @roasbeef

Changelog-None

### DIFF
--- a/common/sphinx.c
+++ b/common/sphinx.c
@@ -511,6 +511,7 @@ struct onionpacket *create_onionpacket(
 			      sphinx_hop_size(&sp->hops[num_hops - 1]);
 	u8 filler[fillerSize];
 	struct keyset keys;
+	u8 padkey[KEY_LEN];
 	u8 nexthmac[HMAC_SIZE];
 	u8 stream[ROUTING_INFO_SIZE];
 	struct hop_params *params;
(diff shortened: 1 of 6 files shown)


niftynei:
ACK 5a24039

cdecker:
I'd prefer we have a way to turn the random bytes off, or at least produce them in a reproducible way, e.g., usin chacha20 which is already in use elsewhere in the onion.

Otherwise we'll end up without test-vectors.

rustyrussell:
> I'd prefer we have a way to turn the random bytes off, or at least produce them in a reproducible way, e.g., usin chacha20 which is already in use elsewhere in the onion.
> 
> Otherwise we'll end up without test-vectors.

Well, we already have a backdoor to set the session_key for testing, so we'd just need another one.

But you're right: generating it this way is Just Good Practice anyway, so let's do that.  Take II coming.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3246/commits/d8c386a141beccc67fc9d30de70cf027331de089

Issue #3194, opened 2019-10-21
Some aliases no longer visible in listnodes


sumBTC:
I automatically find aliases via: `lightning-cli listnodes`

Since I updated to `v0.7.3`some aliases are no longer part of `listnodes`. The IDs are there but without aliases. I suspect it happens with c-lightning versions `<= 0.7.2`. LND and Eclair aliases seem the same (see my [website]( http://Bitkoins.nl) and look for alias_unknown)

For example `ln.rene-pickhardt.de`. The node ID is in `listnodes` (I even have a channel with that node) but without the alias.

### `getinfo` output
Alias: Bitkoins.nl
Version: v0.7.3rc3"

darosior:
> Since I updated to v0.7.3

Which version ? There has been some gossip fixups between the rcs.

sumBTC:
Version: v0.7.3rc3 (that was written under getinfo output.) but the problem started already with rc2.

niftynei:
missing aliases is a side-effect of your node not having node_announcement gossip messages for that node.

prior RC's of c-lightning weren't being very aggressive with pushing out their own node_announcements, see #3152

sumBTC:
@niftynei So it's only a problem with c-lightning versions 0.7.3.rc1 and/or 0.7.3.rc2 or also with older versions of c-lightning? If only with the two rcs, I will close the issue.

whitslack:
It was a problem in 0.7.2 too. See #2969.

I am still experiencing the issue on 0.7.3rc3. Even after my node has been running for three days on 0.7.3rc3, I am still missing node announcements for a sizable fraction of my connected peers.

niftynei:
There are two issues to discuss here, which have the same net result— your node appears to be missing aliases for other nodes.

In the case of “my node is missing alias information from recent c-lightning nodes”, this is likely due to them not sending them very aggressively. #3152 addresses this

In the case of “Im missing large swaths of aliases from nodes, not just from clightning peers”, this is typically a result of not asking our peers for a wide enough range to net the missing information. It should be fixed in RC3 (we now look for a wide range of missing node announcements to ask our peers about). If you’re still having trouble, you can give it a kick by deleting* your gossip store, this will resync all the gossip en masse. Note that if the reason they’re missing is because the above issue, this won’t do much to help as it’s unlikely your peers have the missing information either.

*i.e. stop your node, rename the gossip store to `gossip_store.bak`, restart node

whitslack:
[...]

PR #3247, opened 2019-11-06
gossipd: allow node_announcements and channel_announcements with unsu…

rustyrussell:
…pported features.

The flat feature PR changes the rules so these are OK to propagate.
That makes sense: the unsupported features means there's something
unsupported about the *node* or *channel*, not the msg itself
(for that we'd use a different message type).

Changelog-none

### DIFF
--- a/gossipd/routing.c
+++ b/gossipd/routing.c
@@ -1763,22 +1763,6 @@ u8 *handle_channel_announcement(struct routing_state *rstate,
 	/* FIXME: Handle duplicates as per BOLT #7 */
 
 	/* BOLT #7:
-	 *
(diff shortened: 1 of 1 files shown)


darosior:
Hope that will help with current [gossip problems](https://github.com/ElementsProject/lightning/issues/3194)..

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3247/commits/e1ce255220bbcfe245cca2cd516d9f872e8ab357

darosior:
ACK e1ce255220bbcfe245cca2cd516d9f872e8ab357

PR #3249, opened 2019-11-08
display a warning if the user use --proxy and hidden service and not …

Saibato:
…disables dns

As @bitcoin-software reported the default behavior is in UX therms
not what the users expects if --proxy is configured.  Users tent to associate
with the term proxy, that all traffic is routed trough the proxy.

We should warn the user also in the logs and on startup, if he not selects also --use-always-proxy=true
and --disable-dns, when having defined and used also a hidden service.

The default behavior is that ipv4/6 FQDN will be resolved locally,
since we use tor socks5 proxy for onion address with the socks4a improvement,
the user could and should let Tor do the dns.

Changelog-Changed: Warn the users that he had configed a possible tor dns/ip leak

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -672,6 +672,16 @@ static void check_config(struct lightningd *ld)
 
 	if (ld->use_proxy_always && !ld->proxyaddr)
 		fatal("--always-use-proxy needs --proxy");
+
+	if (ld->proxyaddr && (!ld->use_proxy_always || ld->config.use_dns)) {
+		if (tal_count(ld->proposed_wireaddr) != 0 && some_tor_addresses(ld->proposed_wireaddr)) {
+			log_unusual(ld->log, "WARNING: Your Tor setup use a hidden service address."
+						" --always-use-proxy=%s and --disable-dns %s"
+						" is not recommended, please set --always-use-proxy=true and use --disable-dns, we hope you know what you are doing!"
+						" c-lightning try to help you and detected that you have configured a possible dns leak of your"
+						" hidden service",ld->use_proxy_always?"true":"false", ld->config.use_dns?"not set":"set");
+		}
+	}
 }
 
 static void setup_default_config(struct lightningd *ld)
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -606,3 +606,30 @@ bool all_tor_addresses(const struct wireaddr_internal *wireaddr)
 	}
 	return true;
 }
+
+bool some_tor_addresses(const struct wireaddr_internal *wireaddr)
+{
+	for (int i = 0; i < tal_count(wireaddr); i++) {
+		switch (wireaddr[i].itype) {
+		case ADDR_INTERNAL_SOCKNAME:
+			continue;
+		case ADDR_INTERNAL_FORPROXY:
+			continue;
+		case ADDR_INTERNAL_ALLPROTO:
+			continue;
+		case ADDR_INTERNAL_AUTOTOR:
+			return true;
+		case ADDR_INTERNAL_WIREADDR:
+			switch (wireaddr[i].u.wireaddr.type) {
+			case ADDR_TYPE_IPV4:
+			case ADDR_TYPE_IPV6:
+				continue;
+			case ADDR_TYPE_TOR_V2:
+			case ADDR_TYPE_TOR_V3:
+				return true;
+			}
+		}
+		abort();
+	}
+	return false;
+}
--- a/common/wireaddr.h
+++ b/common/wireaddr.h
@@ -157,4 +157,6 @@ struct addrinfo *wireaddr_internal_to_addrinfo(const tal_t *ctx,
 
 bool all_tor_addresses(const struct wireaddr_internal *wireaddr);
 
+bool some_tor_addresses(const struct wireaddr_internal *wireaddr);
+
 #endif /* LIGHTNING_COMMON_WIREADDR_H */


darosior:
> --use-always-proxy=true and --disable-dns

What do you think of disabling DNS if `--use-always-proxy=true` ?

lightningd/options.c
```diff
@@ -672,7 +672,19 @@ static void check_config(struct lightningd *ld)
 
 	if (ld->use_proxy_always && !ld->proxyaddr)
 		fatal("--always-use-proxy needs --proxy");
-}
+
+	if (ld->proxyaddr && (!ld->use_proxy_always || ld->config.use_dns)) {
+		if (tal_count(ld->proposed_wireaddr) != 0
+	    && some_tor_addresses(ld->proposed_wireaddr)) {
```
darosior: Indentation :-)

Saibato: > What do you think of disabling DNS if --use-always-proxy=true ?


aktualy doing the follow PR right now 

`if (!daemon->use_proxy_always) 
      if (broken_resolver(daemon)) {
or so ...
		`

lightningd/options.c
```diff
@@ -672,6 +672,16 @@ static void check_config(struct lightningd *ld)
 
 	if (ld->use_proxy_always && !ld->proxyaddr)
 		fatal("--always-use-proxy needs --proxy");
+
+	if (ld->proxyaddr && (!ld->use_proxy_always || ld->config.use_dns)) {
+		if (tal_count(ld->proposed_wireaddr) != 0 && some_tor_addresses(ld->proposed_wireaddr)) {
+				log_unusual(ld->log, "WARNING: Your Tor proxy setup with use of hidden service address"
```
darosior:
```suggestion
			log_unusual(ld->log, "WARNING: Your Tor proxy setup with use of hidden service address"
```

Saibato: :-)

lightningd/options.c
```diff
@@ -672,6 +672,16 @@ static void check_config(struct lightningd *ld)
 
 	if (ld->use_proxy_always && !ld->proxyaddr)
 		fatal("--always-use-proxy needs --proxy");
+
+	if (ld->proxyaddr && (!ld->use_proxy_always || ld->config.use_dns)) {
+		if (tal_count(ld->proposed_wireaddr) != 0 && some_tor_addresses(ld->proposed_wireaddr)) {
+			log_unusual(ld->log, "WARNING: Your Tor proxy setup with use of hidden service address"
+						" and --allways-use-proxy=%s and --disable-dns=%s"
```
gorazdko: spelling: --always-use-proxy, --disable-dns (without argument)

Saibato: WARNING: Your Tor setup use a hidden service address --always-use-proxy=**false** and --disable-dns **not set** is not recommended, please set --always-use-proxy=true and use --disable-dns, we hope you know what you are doing!
 c-lightning try to help you and detected that you have configured a possible dns leak of your hidden service

gorazdko: Thanks, much more understandable.

WARNING: Your Tor setup use**s** a hidden service address **with** --always-use-proxy=false and **without** --disable-dns **which** is not recommended, please set --always-use-proxy=true and use --disable-dns, we hope you know what you are doing!
c-lightning **is trying** to help you **because it** detected that you have configured a possible dns leak of your hidden service

Saibato: Thank you too, you caught one, but now its enough for me, since I fear that ends in tautologies.
Some say I am as polite and hansom as a satoshi, but I clearly differ when there is no real progress, then I tent to get rude and bossy.

Here is the bossy unfair mean and I feel a shame for exscript of the original draft of this PR and some hints for the next party, where some caught the front row  tick tick ...

> Cybermen have tagged your system for upgrade by low r value and the doctor is out of town ...

gorazdko: There was no ill intent in my posts. We can leave it as it is.

Saibato: keep CALM,  You are clearly welcome and even now there is one mistype, if you insist, I will change the p to P to make you happy

gorazdko: i dont insist on anything - its your call. And this PR does not affect my feelings whatsoever.

rustyrussell:
Hmm, if this change is desired, we should go further:

1. Deprecate `--always-use-proxy`.
2. Add `--proxy-tor-only`.
3. Change default to be always-use-proxy.

That's pretty clear, I think?  Personally, I want my node to be able to reach nodes on Tor, but I don't need most traffic over it.

Saibato:
> That's pretty clear, I think? Personally, I want my node to be able to reach nodes on Tor, but I don't need most traffic over it.

Me too, AFAICS the new logo of Tor is "Take back the internet with Tor" , c-ln's Tor implementation is unique in LN space, since we have this true quiet mode ( if we want ). If Tor is the default and all have a this semi unique Tor address bound to node id fine, most traffic from node to node will be intern relayed and never leave Tor. 
Exitspoints are on the other hand a single point of failure.

I agree to reach Ip address directly should never be removed, we could add a random jitter between the hopps and the first exit to ip that should be enough, to not called out as doxy app.

In so far, this PR must not be applied, if we are crystal clear in documentation and defaults, what we try to do here. 

We are in early stages and most users up till now don't need this warning, since they are mostly deep techie and know.
C-ln will probably be a major transport layer and at some point default will be the norm, so ...

For historical purpose, I personally would like the discussion here and not in semi hidden telegram or signal channels, since such channeled unopen discussion modes are often the begin of control structures applied.   

I like your suggestion of --proxy-tor-only and to change the defaults of proxy, how about  --add --tor= soly for tor socks ?.  And have --proxy for general socks4/5 proxy ?

I push a WIP  --proxy-tor-only and to change the defaults,  today that could trigger this discussion further.
And one for --random-jitter for all packets might be better work than use of dandelon , but that almost yields in research space. Sigh..
Please feel free to speed me out and take the lead in this, since that matter is also your home turf.

Saibato:
With #3251 :+1:  merged and  #3254 on it"s way I close this now, to defer priority to other issue. This is now low prio and I don't want to be part of pr inflation and review lag, since attention defers can also be a kind of vector programing ... thx for review,  I close this now

cdecker:
> With #3251  merged and #3254 on it"s way I close this now, to defer priority to other issue. This is now low prio and I don't want to be part of pr inflation and review lag, since attention defers can also be a kind of vector programing ... thx for review, I close this now

Not at all, thanks for the PR, and feel free to reopen later or file an issue to track progress :+1:

PR #3250, opened 2019-11-08
devtools/onion: fixed argument passing for generate and decode methods

remyers:
Fixed 'generate' and 'decode' methods to properly parse argv arguments.
Fixed so methods correctly use the associated data set by the --assoc-data flag.

I also added some examples to -h to better show how these methods work with real onion data.

Introduced a behavioral change to read the onion from a file instead of stdin; this makes it easier to debug in some IDEs (eg. vscode).

### DIFF
--- a/devtools/onion.c
+++ b/devtools/onion.c
@@ -31,10 +31,11 @@ static void do_generate(int argc, char **argv,
 	struct sphinx_path *sp;
 	struct hop_data hops_data[num_hops];
 
-	assocdata = tal_arr(ctx, u8, ASSOC_DATA_SIZE);
+	const u8* tmp_assocdata =tal_dup_arr(ctx, u8, assocdata,
+					  ASSOC_DATA_SIZE, 0);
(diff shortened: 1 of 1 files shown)


devtools/onion.c
```diff
@@ -109,7 +109,7 @@ static void do_generate(int argc, char **argv,
 	u8 *serialized = serialize_onionpacket(ctx, res);
 	if (!serialized)
 		errx(1, "Error serializing message.");
-	printf("%s\n", tal_hex(ctx, serialized));
+	printf("%s", tal_hex(ctx, serialized));
```
rustyrussell: Hmm, this is a bit awkward if you run it on the cmdline though?

Can we instead fix it up on the input side?

remyers: Good point; I've rebased to leave the trailing newline but instead trim whitespace with 6a078de . Is this better?

rustyrussell:
Nice work!

remyers:
Thanks! and ht to @willcl-ark for the testing that lead to these fixes.

PR #3251, opened 2019-11-08
We state not to do 'any local DNS' if --always-use-proxy flag is set,…

Saibato:
… so we should do this

Even if it is on startup only once ...
Like @bitcoin-software indicated the expected UX should be in
line with what a user expects the software will do
so we should not dns if we say so with a flag that suggest that.

Signed-off-by: Saibato <saibato.naga@pm.me>

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -200,8 +200,8 @@ static bool broken_resolver(struct daemon *daemon)
 	const char *hostname = "nxdomain-test.doesntexist";
 	int err;
 
-	/* If they told us to never do DNS queries, don't even do this one */
-	if (!daemon->use_dns) {
+	/* If they told us to never do DNS queries, don't even do this one and also not if we just say that we don't */
+	if (!daemon->use_dns || daemon->use_proxy_always) {
 		daemon->broken_resolver_response = NULL;
 		return false;
 	}


darosior:
Shouldn't we go beyond and set `use_dns` to `false` so that `connectd` does not try to resolve unknown ids through bitcoinstats ? It would also cover the same feature as your present commit : 
https://github.com/ElementsProject/lightning/blob/4bf2bbc8616c9a50311058abf5ff9e7b56775400/connectd/connectd.c#L203-L207

Saibato:
> Shouldn't we go beyond and set `use_dns` to `false` so that `connectd` does not try to resolve unknown ids through bitcoinstats ? It would also cover the same feature as your present commit :
> 
> https://github.com/ElementsProject/lightning/blob/4bf2bbc8616c9a50311058abf5ff9e7b56775400/connectd/connectd.c#L203-L207

--always_use_proxy=true even in active master prevents the bootstap resolve call even with out changing the value of use_dns.
The sad part here in this function,  is that the comment and expected outcome is also false here. stated, i guess that was your point.
use_dns disables only the bootstrap call if default always_use_proxy false, but not the resolve for bogus dns servers.
Suppose the cmd line would have say --addr=nowhere:1234 even with **use_dns** false and use_proxy_always default , we would resolve nowhere, only use_proxy_always will disable all local dns and route the resolve to tor , even the seednames from id, that seams to me the right way to do since when we have semi fixed tor address you could learn your tor addr by node id ( I am not sure if tor by now on release resolve also srv entry ..hmm...) , all will be in line with documentation, if we also prevent the call to resolve bogus dns servers here,

Since we get the value for use_dns as bootstrap flag from towire, we should keep it and not change it here.
What we can do here, is do the if (!daemon.always_use_proxy) local in
`static bool broken_resolver(struct daemon *daemon)` since we have daemon in the function, side effect we could set the  broken_resolver_response to NULL.since we accept the proxy we choose even tor clearly as not broken  :smile:  
I push the change now . please review.

PR #3254, opened 2019-11-12
Add handling and define ip-socks5-proxy and route by default ip and tor traffic over Tor

Saibato:
Keep the changes  minimal and backward compatible

TODO

- [x] documnetation

- [x] tests


Saibato:
This is for now better handled in  #3290 so I close this now.

PR #3260, opened 2019-11-14
Add `createonion` and `sendonion` RPC commands

### LINKED ISSUE #3227: Revive "sendpay custom onion" feature
With the htlc_accepted feature in place, I tried to see if my plugin would work with C-Lightning. Conclusion: I'm still missing some functionality on the interface.

The main thing I'm missing is the ability to send customized data as onion payload - something along the lines of pull request #1715 .  I believe that pull request was closed without being merged, and no equivalent functionality currently exists.

I'd like to revive it; hopefully I'll now have less trouble fighting merge conflicts with an ever-changing master branch. **But before I do the hard work, I'd like to ask for consensus on this feature.** This was the RPC interface design of pull request #1715 :

Each element of the "route" argument has optional "realm" and "data" elements. When given, "realm" overrides the default realm number (0) for that hop. When given, "data" overrides the default onion payload data for that hop.

This was designed before the Adelaide summit - it may need to be updated for the new realm number semantics.

Note that this design prefers to be feature-complete to such a degree that it allows the user to violate the Lightning protocol. The rationale is that it's good to be feature-complete, that it's really others' responsibility to make sure you're following the protocol, that sensible app developers will probably want to follow the protocol anyway, and that the ability to break the protocol is useful for testing our own code for robustness against others' protocol violations.

cdecker:
This PR introduces two new RPC methods: `createonion` and
`sendonion`. `createonion` is basically an RPC version of the `devtools/onion`
CLI tool allowing the caller to create a custom onion. A custom onion can be
used to send arbitrary payloads to hops along a route, and thus allows
experimenting with protocol extensions. The counterpart is the `htlc_accepted`
hook which allows a plugin implementing the handling of a custom payload
without having to modify the c-lightning source itself.

The `sendonion` RPC method can be used to initiate a payment using a custom
onion, independently from how or where it was constructed. This was achieved
by splitting `sendpay` into the onion creation part and the HTLC addition
part, and injecting the onion into the latter. We therefore reuse the entire
payments machinery and can use `listsendpays` to inspect in-flight and
previous payments initiated with `sendonion`.

Depending on whether `sendonion` was given the shared secrets it can decrypt
the onion it receives, and thus also decrypt an eventual error message
returned from a node along the route. If the shared secrets have not been
provided the daemon cannot decrypt the error and any error will result in the
ecrypted error being stored and exposed in `listsendpays` so the caller, which
knows the shared secrets, can decrypt it externally. This allows _oblivious
sending_ of payments in which the node only learns what would be revealed to
intermediate hops, but not the length or the nodes in the route.

## Use cases

 - Cross-Chain Atomic Swaps: this is the missing piece @bitonic-cjp was
   waiting for. It allows an incoming payment to be intercepted by a plugin on
   one blockchain, and then reinjected into a node running on the other
   blockchain, without having to generate an onion from scratch. `createonion`
   also allows the inclusion of whatever metadata is necessary to signal to
   the swap node what to do (which chain to swap to, what exchange rate,
   etc.).
 - Experimental protocol extensions: a number of protocol extensions and
   experiments can be based on `createonion` and `sendonion`. Here are a few
   examples:
   - Rendezvous routing: the rendez vous draft I wrote requires the ephemeral
     key to be swapped out at the rendez-vous node. The rendez-vous point can
     intercept using `htlc_accepted` mangle the onion packet as required and
     then re-inject the modified onion into the node. Forwarding is then
     simply handled by the plugin.
   - Trampoline routing: a trampoline waits for incoming trampoline requests
     and then takes over by searching routes to the destination / next
     trampoline and driving retries until the payment succeeds or must fail.
   - Chat: inspired by @joostjager's WhatSat I implemented a chat plugin that
     should be mostly compatible. It uses `createonion` and `sendonion` to
     stuff the chat message in the onion payload for the destination.
 - Oblivious payments: Since we can now create an onion externally, i.e., in a
   wallet interface, and inject it, we can hide the route and destination from
   the node itself, reducing the information that a compromised node may learn
   from the node alone.

## Open Questions

 - Should `createonion` and `sendonion` be developer-only (at least initially)?

Depends https://github.com/ElementsProject/lightning/pull/3278
Closes #3227 
Fixes #1715


common/sphinx.c
```diff
@@ -518,11 +518,6 @@ static bool sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
 	size_t padding_size;
 	int pos = 0;
 
-#if !EXPERIMENTAL_FEATURES
-	if (hop->type != SPHINX_V0_PAYLOAD)
-		return false;
-#endif
-
```
rustyrussell: Changelog should actually announce TLV!  (Changelog-Added: We handle modern TLV-style payloads).  This is important because we never announce EXPERIMENTAL features.

lightningd/pay.c
```diff
@@ -1070,3 +1071,103 @@ static const struct json_command listsendpays_command = {
...
+		   NULL)) {
+		return command_param_failed();
+	}
+
+	if (session_key == NULL)
+		sp = sphinx_path_new(cmd, assocdata);
+	else
+		sp = sphinx_path_new_with_key(cmd, assocdata, session_key);
+
+	json_for_each_arr(i, hop, hopstok) {
```
rustyrussell: It's better, BTW, to do the parsing in a new param_hops_array.  That way you get proper checking if you use the 'check' rpc command.  Might be too awkward though?

lightningd/pay.c
```diff
@@ -1070,3 +1071,103 @@ static const struct json_command listsendpays_command = {
...
+
+		if (!payload)
+			return command_fail(
+			    cmd, JSONRPC2_INVALID_PARAMS,
+			    "'payload' should be a hex encoded binary, not '%.*s'",
+			    pubkeytok->end - pubkeytok->start,
+			    buffer + pubkeytok->start);
+
+		
+		if (!typetok || !json_tok_streq(buffer, typetok, "legacy")) {
```
rustyrussell: Prefer an explicit test here?  Either "legacy" or "raw" (the default), rather than a random string.  Particularly if we add "tlv" later.

lightningd/pay.c
```diff
@@ -1070,3 +1071,103 @@ static const struct json_command listsendpays_command = {
...
+			type = SPHINX_V0_PAYLOAD;
+		}
+
+		sphinx_add_raw_hop(sp, &pubkey, type, payload);
+		hop_count++;
+	}
+
+	if (hop_count == 0)
+		return command_fail(cmd, JSONRPC2_INVALID_REQUEST,
+				    "Cannot create an onion without hops.");
```
rustyrussell: "Onion but no hops?  The beer would be terrible!"

CHANGELOG.md
```diff
@@ -529,3 +529,4 @@ There predate the BOLT specifications, and are only of vague historic interest:
 [0.3]: https://github.com/ElementsProject/lightning/releases/tag/v0.3-2016-05-26
 [0.2]: https://github.com/ElementsProject/lightning/releases/tag/v0.2-2016-01-22
 [0.1]: https://github.com/ElementsProject/lightning/releases/tag/v0.1-2015-08-08
+
```
rustyrussell: Drop this?

lightningd/pay.c
```diff
@@ -630,6 +630,24 @@ static bool should_use_tlv(enum route_hop_style style)
...
+				   enum onion_type *failcode)
+{
+	const u8 *onion;
+	unsigned int base_expiry;
+	struct htlc_out *hout;
+	base_expiry = get_block_height(ld->topology) + 1;
+	onion = serialize_onionpacket(tmpctx, packet);
+	*failcode = send_htlc_out(channel, first_hop->amount,
+				  base_expiry + first_hop->delay,
+				  payment_hash, onion, NULL, &hout);
```
rustyrussell: The inversion of failcode and hout here gives me whiplash.

"Always return the error"!

lightningd/pay.c
```diff
@@ -890,6 +890,98 @@ param_route_hop(struct command *cmd, const char *name, const char *buffer,
...
+	enum onion_type failcode;
+	struct htlc_out *hout;
+	struct route_hop *first_hop;
+	struct sha256 *payment_hash;
+	struct channel *channel;
+	struct lightningd *ld = cmd->ld;
+	struct wallet_payment *payment;
+	const char *label;
+
+	if (!params || !deprecated_apis) {
```
rustyrussell: This test should be eliminated?  Looks like cut & paste from a function which had to handle weird deprecated options?

lightningd/pay.c
```diff
@@ -890,6 +890,98 @@ param_route_hop(struct command *cmd, const char *name, const char *buffer,
...
+	}
+
+	hout = send_onion(cmd->ld, packet, first_hop, payment_hash, channel,
+			  &failcode);
+
+	payment = tal(hout, struct wallet_payment);
+	payment->id = 0;
+	payment->payment_hash = *payment_hash;
+		payment->status = PAYMENT_PENDING;
+	payment->msatoshi = first_hop->amount;
```
rustyrussell: This should be the last_hop amount I think.

lightningd/pay.c
```diff
@@ -814,6 +814,82 @@ send_payment(struct lightningd *ld,
 	return NULL;
 }
 
+static struct command_result *
+param_route_hop(struct command *cmd, const char *name, const char *buffer,
+		const jsmntok_t *tok, struct route_hop **hop)
+{
```
rustyrussell: Need to mark UNNEEDED here, and eliminate in next patch.  Otherwise compile bisection is broken :(

lightningd/pay.c
```diff
@@ -453,30 +453,25 @@ void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
 					   &hout->payment_hash));
 		return;
 	}
-
-	/* FIXME: Prior to 299b280f7, we didn't put route_nodes and
-	 * route_channels in db.  If this happens, it's an old payment,
-	 * so we can simply mark it failed in db and return. */
```
rustyrussell: It's OK to delete this (released in v0.6, so over 12 months ago: a payment is unlikely to still be pending since then), but you must place a justification for it in the commit msg.  Like this one :)

common/json.c
```diff
@@ -143,6 +143,14 @@ bool json_to_bool(const char *buffer, const jsmntok_t *tok, bool *b)
 	return false;
 }
 
+bool json_to_secret(const char *buffer, const jsmntok_t *tok, struct secret *dest)
+{
+	size_t seclen = sizeof(struct secret), hexlen = tok->end - tok->start;
+	if (hexlen != 2 * seclen)
+		return false;
```
rustyrussell: hex_decode does this check for you, BTW!

i.e. this function is a one-liner.

lightningd/pay.c
```diff
@@ -946,6 +969,20 @@ static struct command_result *json_sendonion(struct command *cmd,
...
+			if (!json_to_secret(buffer, cur, &path_secrets[i]))
+				return command_fail(
+				    cmd, JSONRPC2_INVALID_PARAMS,
+				    "shared_secret[%zu] isn't a valid "
+				    "hex-encoded 32 byte secret",
+				    i);
+		}
+	} else {
+		path_secrets = NULL;
+	}
```
rustyrussell: Again, it's be nice to implement this as param_secrets_array so 'check' rpc works better.

doc/lightning-createonion.7.md
```diff
@@ -0,0 +1,126 @@
...
+	"shared_secrets": [
+		"88ce98c73e4d9293ab1797b0a913fe9bca0213a566252047d01b8af6da871f3e",
+		"4474d296810e57bd460ef8b83d2e7d288321f8a99ff7686f87384699747bcfc4",
+		"2a862e4123e01799a732be487fbce297f7dc7cc1467e410f18369cfee476adc2"
+	]
+}
+```
+
+The `onion` corresponds to 1366 hex-encoded bytes. Each shared secret consists
+of 23 hex-encoded bytes. Both arguments can be passed on to **sendonion**.
```
rustyrussell: 32 hex-encoded bytes?

doc/lightning-createonion.7.md
```diff
@@ -0,0 +1,126 @@
...
+	{
+		"type": "legacy",
+		"pubkey": "022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59",
+		"payload": "000067000001000100000000000003e90000007b000000000000000000000000"
+	}, {
+		"type": "legacy",
+		"pubkey": "035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d",
+		"payload": "000067000003000100000000000003e800000075000000000000000000000000"
+	}, {
+		"type": "legacy",
```
rustyrussell: We should change "type" to "style" since we've used that elsewhere.  Consistency in APIs is really important.

I'd also be in favor of changing "pubkey" here to "id", though the former is more correct.

lightningd/pay.c
cdecker: We don't have access to that though, since we're given a fully wrapped onion only. We don't know what happens to the funds, apart from how much is sent initially.

common/sphinx.c
cdecker: I was hoping you'd add a changelog entry to #3167, since you did all the heavy lifting there, but you're right, that we don't note experimental features in the changelog.

I'd propose the following ling:

```
Changelog-Added: Plugins may now handle modern TLV-style payloads via the `htlc_accepted` hook
```

That's a less broad claim than full TLV support for sending and receiving without a plugin, which as you noted is still experimental.

lightningd/pay.c
cdecker: There is actually no difference between raw and tlv:

https://github.com/ElementsProject/lightning/blob/21555b277d2891c9b306e677f77441c81b068549/common/sphinx.c#L526-L530

Raw really is just confusing and we should remove it.

lightningd/pay.c
cdecker: I see, I guess `send_onion` should follow that convention as well :-)

bitonic-cjp:
Nice.

Out of curiosity: will this allow sendpay to be re-implemented as a plugin?

cdecker:
> Nice.
> 
> Out of curiosity: will this allow sendpay to be re-implemented as a plugin?

Indeed that's a possibility. An example that does this can be seen here https://github.com/ElementsProject/lightning/pull/3260/commits/05c5146a49e21395a8df0cee3bfef5dea7a3188e

lightningd/pay.c
```diff
@@ -314,9 +319,17 @@ local_routing_failure(const tal_t *ctx,
...
-						   &payment->route_nodes[0]);
+
+	if (payment->route_nodes != NULL && payment->route_channels != NULL) {
+		routing_failure->erring_channel =
+		    tal_dup(routing_failure, struct short_channel_id,
+			    &payment->route_channels[0]);
+		routing_failure->channel_dir =
+		    node_id_idx(&ld->id, &payment->route_nodes[0]);
+	} else {
+		routing_failure->erring_channel = NULL;
```
niftynei: does the `channel_dir` not need to be set either?

lightningd/pay.c
```diff
@@ -361,6 +361,25 @@ remote_routing_failure(const tal_t *ctx,
...
+	if (route_nodes == NULL) {
+		/* This means we have the `shared_secrets`, but cannot infer
+		 * the erring channel and node since we don't have them. This
+		 * can happen if the payment was initialized using `sendonion`
+		 * and the `shared_secrets` where specified. */
+		dir = 0;
+		erring_channel = NULL;
+		erring_node = NULL;
+
+		/* We don't know if there's another route, that'd depend in
```
niftynei: nit: depend in -> depend on

wallet/wallet.c
```diff
@@ -2273,6 +2280,7 @@ wallet_payment_by_hash(const tal_t *ctx, struct wallet *wallet,
 					     ", msatoshi_sent"
 					     ", description"
 					     ", bolt11"
+					     ", failonionreply"
```
niftynei: does this column already exist in the database? i don't see a migration statement in this commit :eyes:

doc/lightning-createonion.7.md
```diff
@@ -0,0 +1,126 @@
...
+RETURN VALUE
+------------
+
+On success, an object containing the onion and the shared secrets will be
+returned. Otherwise an error will be reported. The following example is the
+result of calling *createonion* with the above hops parameter:
+
+```json
+{
+	"onion": "0003f3f80d2142b953319336d2fe4097[..snip...]6af33fcf4fb113bce01f56dd62248a9e5fcbbfba35c",
```
niftynei: ....:scissors:.... :D

tests/test_pay.py
```diff
@@ -2434,3 +2434,36 @@ def test_shadow_routing(node_factory):
     assert total_amount.millisatoshis < (n_payments * amount) * (1 + 0.01)
 
     # FIXME: Test cltv delta too ?
+
+
+def test_createonion_rpc(node_factory):
+    l1 = node_factory.get_node()
+
+    hops = [{
+        "type": "legacy",
```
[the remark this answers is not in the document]
rustyrussell: The problem is that the spec reserves one value for further extensions: that is neither legac

lightningd/pay.c
cdecker: We already have an enum for that (`enum sphinx_payload_type`) so I went ahead and just added an explicit check for `legacy` or `tlv`, and added an error message if neither is used,

lightningd/pay.c
cdecker: `NULL` in the DB and 0 in-memory sounds like the correct answer here, since we have no idea how much will arrive at the destination and we still have the `msatoshi_sent` amount here. Does that sound ok to you @rustyrussell and @niftynei?

cdecker: An improvement would be to make `msatoshi` in `struct wallet_payment` a pointer since it can be `NULL`, that would come at the cost of an extra allocation, but be more correct?

cdecker:
> I think you inadvertently removed the randomization initialization for the onion padding.

This branch predates that fix. I have rebased it locally, and checked that the fix is still in there. Will push as soon as I have addressed your feedback :+1:

lightningd/pay.c
```diff
@@ -1070,3 +1071,103 @@ static const struct json_command listsendpays_command = {
...
+
+	if (hop_count == 0)
+		return command_fail(cmd, JSONRPC2_INVALID_REQUEST,
+				    "Cannot create an onion without hops.");
+
+
+	packet = create_onionpacket(cmd, sp, &shared_secrets);
+	if (!packet)
+		return command_fail(cmd, LIGHTNINGD,
+				    "Could not create onion packet");
```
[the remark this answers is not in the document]
cdecker: Not a whole lot we can do actually, it either failed to generate the keys in `generate_hop_params` which might happen due to invalid keys, or ecdh failing, or we failed to write an onion frame, e.g., payload was too long.

lightningd/pay.c
```diff
@@ -1070,3 +1071,103 @@ static const struct json_command listsendpays_command = {
...
+
+	serialized = serialize_onionpacket(cmd, packet);
+	if (!serialized)
+		return command_fail(cmd, LIGHTNINGD,
+				    "Could not serialize onion packet");
```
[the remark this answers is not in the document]
cdecker: Just checked, and `serialize_onionpacket` does not fail. I removed the check since it'll never fail.

tests/test_pay.py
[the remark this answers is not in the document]
cdecker: I made sure that either the `style` is not provided (defaulting to TLV) or it is one that we know how to handle (`legacy` or `tlv` for now).

lightningd/pay.c
```diff
@@ -734,19 +750,12 @@ send_payment(struct lightningd *ld,
 		return command_failed(cmd, data);
 	}
 
-	randombytes_buf(&sessionkey, sizeof(sessionkey));
```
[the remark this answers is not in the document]
cdecker: That seems to be a left-over from a prior migration that I'm just cleaning up here. `sphinx_path_new` internally generates a sessionkey and derives the `path_secrets`, so there's no need to do that here anymore.

lightningd/pay.c
```diff
@@ -890,6 +890,98 @@ param_route_hop(struct command *cmd, const char *name, const char *buffer,
...
+		if (!param(cmd, buffer, params,
+			   p_req("onion", param_bin_from_hex, &onion),
+			   p_req("first_hop", param_route_hop, &first_hop),
+			   p_req("payment_hash", param_sha256, &payment_hash),
+			   p_opt("label", param_escaped_string, &label),
+			   NULL))
+			return command_param_failed();
+	}
+
+	packet = parse_onionpacket(cmd, onion, tal_bytelen(onion), &failcode);
```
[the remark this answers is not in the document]
cdecker: I'll fix this up in a cleanup PR since it is not new code of this PR.

lightningd/pay.c
```diff
@@ -494,7 +489,6 @@ void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
 				 hout->key.id,
 				 tal_hex(tmpctx, hout->failuremsg));
 			/* Cannot record failure. */
-			fail = NULL;
```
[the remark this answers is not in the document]
cdecker: It's redundant with the one [here](https://github.com/ElementsProject/lightning/pull/3260/files/05c5146a49e21395a8df0cee3bfef5dea7a3188e..028be006c4cdcc158c7e142551a93cb316f2c81b#diff-e45966e073c4ce4ae4f03161e5a93440L438)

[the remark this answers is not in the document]
cdecker: But I re-added it in a fixup to minimize the diff :-)

tests/test_pay.py
```diff
@@ -2512,3 +2512,27 @@ def serialize_payload(n):
...
+    # And now for a failing payment, using a payment_hash that doesn't match an
+    # invoice
+    payment_hash = "00"*32
+    onion = l1.rpc.createonion(hops=hops, assocdata=payment_hash)
+    l1.rpc.sendonion(onion=onion['onion'], first_hop=first_hop,
+                     payment_hash=payment_hash)
+
+    try:
+        l1.rpc.waitsendpay(payment_hash=payment_hash)
+        raise ValueError()
```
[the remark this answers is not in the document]
cdecker: The `try`-block catches `RpcError`s, so we need to use something else to report an error we didn't expect. This is a bit weird here, since we expect this to fail, and not failing is a test-failure. Really anything other than an `RpcError` (or a subclass) would be good here, and `ValueError` is just my goto "don't know what I should use"-error :-)

lightningd/pay.c
```diff
@@ -343,10 +347,7 @@ remote_routing_failure(const tal_t *ctx,
 	route_channels = payment->route_channels;
 	origin_index = failure->origin_index;
 
-	assert(origin_index < tal_count(route_nodes));
-
-	/* Check if at destination. */
-	if (origin_index == tal_count(route_nodes) - 1) {
+	assert(route_nodes == NULL || origin_index < tal_count(route_nodes));
 		/* If any channel is to blame, it's the last one. */
```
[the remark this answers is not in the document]
cdecker: Whoops, that's what I get for trying to disect a change into separate commits. Fixed in a fixup.

lightningd/pay.c
cdecker: Leaving it unset on purpose so valgrind can tell us if we touch it inadvertently. We use the `erring_channel` as sentinel for the presence of both, and if we touch the direction we know it's an error (it's used in the JSON-RPC only anyway).

wallet/wallet.c
cdecker: It does indeed :+1: We just didn't display it so far.

cdecker:
@niftynei: GH just decided to forget all the replies I had pending addressing your excellent poitns. I'll try to recreate the important ones from memory.

Pushing the fixups and rebased stack now.

tests/test_pay.py
```diff
@@ -2434,3 +2434,36 @@ def test_shadow_routing(node_factory):
...
+
+    res = l1.rpc.createonion(hops=hops, assocdata="BB" * 32)
+    assert(len(res['onion']) == 2 * 1366)
+    assert(len(res['shared_secrets']) == len(hops))
+
+    res = l1.rpc.createonion(hops=hops, assocdata="42" * 32,
+                             session_key="41" * 32)
+    # The trailer is generated using the filler and can be ued as a
+    # checksum. This trailer is from the test-vector in the specs.
+    assert(res['onion'].endswith('be89e4701eb870f8ed64fafa446c78df3ea'))
```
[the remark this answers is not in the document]
cdecker: I'll add a few in a cleanup PR. Anything in particular I should test for? I'm thinking the following:

 - shared_secrets length mismatching the actual onion secrets in `sendonion`
 - Some broken hex strings

cdecker:
> @niftynei: GH just decided to forget all the replies I had pending addressing your excellent poitns. I'll try to recreate the important ones from memory.

Nevermind, GH decided to have a reverse lobotomy when I submitted the comment :-)

cdecker:
Thanks @niftynei and @rustyrussell for the extensive feedback. I tried to address everything, and the resulting fixup stack is a bit big. Here's the range-diff for an easier comparison: [eefd97ff9671b99714305e296e951d4c3075539c..124773bfb14c56e0ea6d474a5872bf3b7581cd72](http://46.101.246.115:8888/ElementsProject/lightning/pull/3260/range_diff/eefd97ff9671b99714305e296e951d4c3075539c..124773bfb14c56e0ea6d474a5872bf3b7581cd72)

cdecker:
As always, happy to squash in the `fixup!`s once the reviewers have acknowledged the changes (ACKs should carry over automatically unless I managed to break @bitcoin-bot again).

tests/test_pay.py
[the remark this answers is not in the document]
darosior: Or the `else:` block can be used but it's strictly the same behaviour.

cdecker:
I rebased and squashed the fixups. The fixups can still be seen [here in the form of a range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3260/range_diff/aa1ca5c9ea5577e3a09fb56e5dfd4b3c02a6b0b9..f256bd808326ced4e5b406e247315c51b63c5bd5)

Ping @rustyrussell and @niftynei :wink:

rustyrussell:
Ack f256bd808326ced4e5b406e247315c51b63c5bd5

PR #3261, opened 2019-11-14
plugin: Pass the full raw_payload including realm to htlc_accepted

cdecker:
So far we've only handled legacy payloads, which meant we could drop the realm
byte since it was always 0x00. Once we start handling TLV payloads the first
byte, i.e., the former realm byte, is important since it gives us the length
of the payload. This is a breaking change, however I don't think there's
anyone using the `raw_payload` as of yet.

Changelog-Changed: JSON-RPC: the `raw_payload` now includes the first byte, i.e., the realm byte, of the payload as well. This allows correct decoding of a TLV payload in the plugins.

### DIFF
--- a/common/sphinx.c
+++ b/common/sphinx.c
@@ -580,7 +580,7 @@ static void sphinx_parse_payload(struct route_step *step, const u8 *src)
 	}
 
 	/* Copy common pieces over */
-	step->raw_payload = tal_dup_arr(step, u8, src + vsize, raw_size, 0);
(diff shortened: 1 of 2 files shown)


rustyrussell:
Ack, but needs test fixup too?

cdecker:
Tests fixed ^^

cdecker:
Re-applying Rusty's ACK

ACK https://github.com/ElementsProject/lightning/pull/3261/commits/b96b537a685863287ccd0a67bc222d4b3f87cfb6

PR #3264, opened 2019-11-15
Handle reply_channel_range from LND (and enhance gossipwith)

rustyrussell:
See https://github.com/lightningnetwork/lnd/issues/3728

Tested and verified on my node, under valgrind.

### DIFF
--- a/gossipd/queries.c
+++ b/gossipd/queries.c
@@ -721,23 +721,34 @@ const u8 *handle_reply_channel_range(struct peer *peer, const u8 *msg)
 	if (end > peer->range_end_blocknum)
 		end = peer->range_end_blocknum;
 
-	/* We keep a bitmap of what blocks have been covered by replies: bit 0
-	 * represents block peer->range_first_blocknum */
-	b = bitmap_ffs(peer->query_channel_blocks,
-		       start - peer->range_first_blocknum,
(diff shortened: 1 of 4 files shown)


gossipd/queries.c
```diff
@@ -721,23 +721,34 @@ const u8 *handle_reply_channel_range(struct peer *peer, const u8 *msg)
...
-				       peer->range_first_blocknum + b);
-	}
+	/* LND mis-implemented the spec.  If they have multiple replies, set
+	 * each one to the *whole* range, with complete=0 except the last.
+	 * Try to accomodate that (pretend we make no progress until the
+	 * end)! */
+	if (first_blocknum == peer->range_first_blocknum
+	    && first_blocknum + number_of_blocks == peer->range_end_blocknum
+	    && !complete
+	    && tal_bytelen(msg) == 64046) {
```
cdecker: Where does this constant come from?

rustyrussell: That's the size of the message LND sends (8000 uncompressed scids).  I wanted this workaround to be as narrow as possible to avoid catching real cases.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3264/commits/ced01b3366c80a5d6f31151c96d7aa7aa339a6e0

Issue #3265, opened 2019-11-15
db: txindex known without blockheight 


m-schmoock:
### Issue and Steps to Reproduce
An data consistency assert added by commit ed238758a2c0d733240e8f2eb6b20629943c1f89 (merged yesterday into master) causes problems on my testnet and mainnet nodes. The added assert checks that if a transaction index is known a blockheight must also be known or wise versa.

If your node is affected it can be reproduced by:
```
lightning-cli listtransactions
```

If the above command asserts, its very likely your node has corrupt data in `sqlite` db.

### SQL query to verify
```
> sqlite3 ~/.lightning/lightningd.sqlite3

SELECT  HEX(t.id), t.blockheight, t.txindex FROM transactions t WHERE t.txindex > 0 and t.blockheight IS NULL;

(for my mainnet node:)
CB785ABCC3971F94607A270009D161460D1D63C8E915CFEC933320DF37ED372F||1663
5E35A558D3061E6EC6D91A89346639DEDAAA704281A4F7B597D63EB6591307B8||855
```

### Notes
This cannot be 'fixed' by using `dev-rescan-outputs`. We could add a `dev-rescan-transactions`  or add functionality to `dev-rescan-outputs` and call it on startup if we detect this inconsistency in the database.

m-schmoock:
It may very well be that current code does not reproduce this, as the code that lead to it has already been fixed, but the databases with corruptions are still out in the wild...

cdecker:
I see, the assertion might be a bit overzealous. The `blockheight` might be under on a reorg, leaving the txindex set. We should just remove the assertion and add the index only if we also have a block height.

m-schmoock:
I also thought about reorgs on code that didnt handle DB update correctly. But I doubt this was the reason since I have 4 such TX on two nodes. Too many to be a coincidence.

In any case, the txindex should be erased from the DB if the blockheight is unknown for whatever reason. If we check this we could keep the assert...

It would also be good to know what causes this...

cdecker:
We'll likely not unset the `txindex` on reorg at least since we rely heavily on the `ON DELETE` foreign keys in the DB which cannot really be applied to non-foreign keys. I'd prefer to gate the `txindex` being shown based on whether we have a blockheight.

m-schmoock:
@cdecker addressed by https://github.com/ElementsProject/lightning/pull/3267

PR #3267, opened 2019-11-18
fix: listtx set txindex only if blockheight is known

m-schmoock:
This will patch the `listtransactions` command in a way that the hard assert regarding `txindex` and blockheight is removed (https://github.com/ElementsProject/lightning/issues/3265). We will now set the `txindex` only if a blockheight is known.

**Also:** I do not believe the missing blockheights are reorg artifacts for two reasons:
- The `txindex` we have in the DB are the correct onchain `txindex` when looking up the proper block. Incase of a reorg artifact they are expected to be different, as they would have been replaced by a block with different index. If this is still reorg related, the reorg code does not write the correct blockheight BUT the txindex into the DB, which would be bad.
- they are too many to be a coincident., in my case 4 transactions on two nodes: mainnet and testnet.

We might investigate this further. SQL command to check this:
```
> sqlite3 ~/.lightning/lightningd.sqlite3

SELECT  HEX(t.id), t.blockheight, t.txindex FROM transactions t WHERE t.txindex > 0 and t.blockheight IS NULL;

(my mainnet node)
CB785ABCC3971F94607A270009D161460D1D63C8E915CFEC933320DF37ED372F||1663
5E35A558D3061E6EC6D91A89346639DEDAAA704281A4F7B597D63EB6591307B8||855
```

**Note:** Additionally, I replaced the TABs with BLANKs in this block that got messed up last time, sorry I used a new IDE.

---------------------

Alternatively we could set the `blockheight` and `txindex` only if both are known, since having a `blockheight` without `txindex` is also not very meaningful:

```C
               if (!db_column_is_null(stmt, 2) && !db_column_is_null(stmt, 3)) {
                       cur->blockheight = db_column_int(stmt, 2);
                       cur->txindex = db_column_int(stmt, 3);
               } else {
                       cur->blockheight = 0;
                       cur->txindex = 0;
               }
```

Changelog-None

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -3461,16 +3461,18 @@ struct wallet_transaction *wallet_transactions_get(struct wallet *w, const tal_t
 			cur->tx = db_column_tx(txs, stmt, 1);
 			cur->rawtx = tal_dup_arr(txs, u8, db_column_blob(stmt, 1),
 						 db_column_bytes(stmt, 1), 0);
-            /* TX may be unconfirmed. */
-            if (!db_column_is_null(stmt, 2) || !db_column_is_null(stmt, 3)) {
-                /* assert incomplete information */
-                assert(!db_column_is_null(stmt, 2) && !db_column_is_null(stmt, 3));
-                cur->blockheight = db_column_int(stmt, 2);
-                cur->txindex = db_column_int(stmt, 3);
-            } else {
-                cur->blockheight = 0;
-                cur->txindex = 0;
-            }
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3267/commits/851fbcd0ac7d5dc9fc008056f8770ce5d9e5a9e2

cdecker:
You're right, it seems one of our txwatches sets the txindex but not the blockheight.

PR #3268, opened 2019-11-18
Config changes

rustyrussell:
* Files now moved into bitcoin/testnet/regtest/etc subdir.
* Config file still parsed in old location, but also network/config is parsed
* Default for new installs in bitcoin, not testnet!

### DIFF
--- a/common/configdir.c
+++ b/common/configdir.c
@@ -84,7 +84,8 @@ static void config_log_stderr_exit(const char *fmt, ...)
 	errx(1, "%s", msg);
 }
 
-static void parse_include(const char *filename, bool must_exist, bool early)
+static void parse_include(const char *filename, bool must_exist, bool early,
+			  size_t depth)
 {
 	char *contents, **lines;
 	char **all_args; /*For each line: either `--`argument, include file, or NULL*/
@@ -136,7 +137,11 @@ static void parse_include(const char *filename, bool must_exist, bool early)
 			continue;
 
 		if (!strstarts(all_args[i], "--")) {
-			parse_include(all_args[i], true, early);
+			/* There could be more, but this gives a hint. */
+			if (depth > 100)
+				errx(1, "Include loop with %s and %s",
+				     filename, all_args[i]);
+			parse_include(all_args[i], true, early, ++depth);
 			continue;
 		}
 
@@ -260,7 +265,7 @@ static void parse_implied_config_file(const char *config_basedir,
 		dir = path_join(NULL, take(dir), network);
 
 	filename = path_join(NULL, take(dir), "config");
-	parse_include(filename, false, early);
+	parse_include(filename, false, early, 0);
 	tal_free(filename);
 }
 
@@ -273,7 +278,7 @@ void parse_config_files(const char *config_filename,
 {
 	if (config_filename) {
 		parse_state = FORCED_CONFIG;
-		parse_include(config_filename, true, early);
+		parse_include(config_filename, true, early, 0);
 		parse_state = CMDLINE;
 		return;
 	}
@@ -387,7 +392,7 @@ void initial_config_opts(const tal_t *ctx,
 
 	/* Read config file first, since cmdline must override */
 	if (*config_filename)
-		parse_include(*config_filename, true, true);
+		parse_include(*config_filename, true, true, 0);
 	else
 		parse_implied_config_file(*config_basedir, NULL, true);
 	opt_early_parse_incomplete(argc, argv, opt_log_stderr_exit);
--- a/common/configdir.c
+++ b/common/configdir.c
@@ -79,7 +79,7 @@ static void config_log_stderr_exit(const char *fmt, ...)
 	errx(1, "%s", msg);
 }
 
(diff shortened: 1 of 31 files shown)


jb55:
this breaks my current system clightning system service with: `network: not permitted in configuration files`. since I had `network=bitcoin` in my config file, so now I will need to add it to the systemd service line. Not too big a deal but just giving a heads up.

jb55:
Another thing that is a bit confusing is lightning-cli now shows this message when clightning isn't running: 
`lightning-cli: WARNING: default network changing in 2020: please set network=testnet in config!`

rustyrussell:
Oops, this is a bug. Network should only be disallowed in the <network>/config file. Should still be valid on the root config.

rustyrussell:
Thanks @jb55 ... and this time with tests!

jb55:
hmm I was able to test it and it worked without doing your `mv` and `rmdir` command, but then again my folder is `~/.lightning-bitcoin` which I explicitly set as an argument to lightningd

common/configdir.c
```diff
@@ -241,6 +244,33 @@ void parse_config_files(const char *config_filename,
...
+static bool smells_like_old_testnet(const char *config_basedir)
+{
+	struct stat st;
+	const char *base = default_base_configdir(NULL);
+
+	if (!config_basedir)
+		config_basedir = base;
+
+	/* If it doesn't exist, it's not testnet. */
+	if (stat(base, &st) != 0) {
```
niftynei: i think you might mean to be testing `config_basedir` here?

common/configdir.c
```diff
@@ -241,6 +244,33 @@ void parse_config_files(const char *config_filename,
...
+		tal_free(base);
+		return false;
+	}
+
+	/* Does it have a bitcoin/ subdir and no testnet/ subdir? */
+	if (stat(path_join(base, config_basedir, "bitcoin"), &st) == 0
```
niftynei: ok i'll admit i'm a bit confused here. at L255 we might have set `config_basedir` to be `base`; in that case we'd be joining the same path twice.

rustyrussell:
> This might be a very painful upgrade, but honestly I don't see how we could make it less painful. Some general comments below:
> 
>     * We have no recursion depth, so circular includes can result in endless
>       loops

Yes, this I can fix quite easily.

>     * Starting a node without a `--network` parameter may end up moving `testnet`
>       files into the `bitcoin` directory and treating them like mainnet.

That should not happen (unless they specify --allow-deprecated-apis=false): if nothing is specified, they get testnet on existing installs.  I will add a test for this, however.

> 
> 
> The latter is really what bothers me, we are moving things around without really knowing what network the config is for, but I also don't see how we could avoid that without reading the DB chain and then comparing with the chainparams.

rustyrussell:
> As @cdecker mentioned, if we want to be super safe for the transition/upgrade path we could check the database for the chainhash. Not sure what our risk profile is though :)

Doing so is amazingly ugly, since we need to plumb though a *lot* of code to get the db open early.  So I chose the simpler approach of leaving existing installs as testnet by default.

common/configdir.c
```diff
@@ -1,17 +1,155 @@
...
+	}
+
+	lines = tal_strsplit(contents, contents, "\r\n", STR_EMPTY_OK);
+
+	/* We have to keep all_args around, since opt will point into it: use
+	 * magic tal name to tell memleak this isn't one. */
+	all_args = tal_arr_label(options_ctx, char *, tal_count(lines) - 1,
+				 TAL_LABEL(options_array_notleak, ""));
+
+	for (i = 0; i < tal_count(lines) - 1; i++) {
```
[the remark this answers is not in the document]
rustyrussell: It NULL-terminates the array, yes.

common/configdir.c
rustyrussell: First arg is context: we're not relying on tmpctx here, so I'm abusing base as a per-function ctx.  Will comment :)

lightningd/options.c
```diff
@@ -972,7 +974,7 @@ void handle_early_opts(struct lightningd *ld, int argc, char *argv[])
 				 ld->config_netdir);
 
 	/* Set default PID file name to be per-network */
-	ld->pidfile = tal_fmt(ld, "lightningd-%s.pid",
+	ld->pidfile = tal_fmt(ld, "../lightningd-%s.pid",
```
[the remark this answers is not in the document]
rustyrussell: No, because they get parsed next.

However, this "../" is wrong (network/ may be a symlink): this code is left over from before I gave in and made basedir and netdir explicit.  Will fix:

```C
	/* Set default PID file name to be per-network (in base dir) */
	ld->pidfile = path_join(ld, ld->config_basedir,
				tal_fmt(tmpctx, "lightningd-%s.pid",
					chainparams->network_name));
```

rustyrussell:
Trivial rebase to fold fixes.

rustyrussell:
Yes, I can remove doubling in a separate patch, is a side effect of when we finally set up the log. Previously it was only for broken messages, which usually meant we exited before log setup anyway...

PR #3271, opened 2019-11-18
devtools: Print hashes + privkeys for sigs

niftynei:
Recent updates to the protocol tests mean that we can now use a privkey + hash to verify a signature; this updates the relevant `mkcommit/mkgossip` devtools to now provide this info for signatures.

### DIFF
--- a/devtools/mkcommit.c
+++ b/devtools/mkcommit.c
@@ -162,16 +162,6 @@ static int parse_config(char *argv[],
 	return argnum;
 }
 
-static char *sig_as_hex(const struct bitcoin_signature *sig)
-{
-	u8 compact_sig[64];
-
-	secp256k1_ecdsa_signature_serialize_compact(secp256k1_ctx,
-						    compact_sig,
-						    &sig->s);
-	return tal_hexstr(NULL, compact_sig, sizeof(compact_sig));
-}
-
 static int parse_htlc(char *argv[],
 		      struct added_htlc **htlcs,
 		      enum htlc_state **htlc_states,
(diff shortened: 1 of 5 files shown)


devtools/mkcommit.c
```diff
@@ -224,6 +224,21 @@ static const struct preimage *preimage_of(const struct sha256 *hash,
...
+
+	printf("privkey: ");
+	for (i = 0; i < 32; i++)
+		printf("%02x", privkey->secret.data[i]);
+	printf("\n");
+	printf("tx_hash: ");
+	for (i = 0; i < sizeof(hash->sha.u.u8); i++)
+		printf("%02x", hash->sha.u.u8[i]);
+	printf("\n");
+}
```
rustyrussell: I generally prefer to use tal_hexstr() here, as I think it's simpler.

rustyrussell: In fact, just print this in SIG() notation so I can cut & paste?

devtools/mkcommit.c
```diff
@@ -388,26 +404,31 @@ int main(int argc, char *argv[])
...
+	/* Get the hash out, for printing */
+	bitcoin_tx_hash_for_sig(local_txs[0], 0, funding_wscript,
+				SIGHASH_ALL, &hash);
 	sign_tx_input(local_txs[0], 0, NULL, funding_wscript,
 		      &local.funding_privkey,
 		      &funding_localkey,
 		      SIGHASH_ALL,
 		      &local_sig);
 	printf("localsig_on_local: %s\n", sig_as_hex(&local_sig));
+	print_privkey_and_hash(&hash, &local.funding_privkey);
```
rustyrussell: I would get rid of the sig_as_hex and just use SIG() notation for that?  Maybe just:
```
printf("localsig_on_local: %s\n", sig_notation(&hash, local.funding_privkey));
```

devtools/mkgossip.c
```diff
@@ -298,9 +311,13 @@ int main(int argc, char *argv[])
 
 	printf("type=channel_announcement\n");
 	printf("   node_signature_1=%s\n", sig_as_hex(&nodesig[lesser_key]));
+	print_hash_and_privkey(&node_privkey[0], &hash);
 	printf("   node_signature_2=%s\n", sig_as_hex(&nodesig[!lesser_key]));
+	print_hash_and_privkey(&node_privkey[1], &hash);
 	printf("   bitcoin_signature_1=%s\n", sig_as_hex(&bitcoinsig[lesser_key]));
+	print_hash_and_privkey(&funding_privkey[0], &hash);
 	printf("   bitcoin_signature_2=%s\n", sig_as_hex(&bitcoinsig[!lesser_key]));
+	print_hash_and_privkey(&funding_privkey[1], &hash);
```
rustyrussell: Same as above, so it can be cut & paste.  Maybe print a real signature in hex with --verbose or something?

niftynei:
rewrote the sig printing commit, 5c04b08

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/49c59d0645503f7ac97dccf1bddcc576a271ac2e

Issue #3269, opened 2019-11-18
Version v0.7.3-113-gdda792c crashes all the time


sumBTC:
### Issue and Steps to Reproduce

See this [file](http://82.217.214.215/out_fail.txt). Rebuilding the gossip_store didn't help

### `getinfo` output
`{
   "id": "026bc066924bca8a22e4baa751533bf2dda4161544d08dab544a001a5385ab0d0c",
   "alias": "myshtery",
   "color": "23375f",
   "num_peers": 5,
   "num_pending_channels": 1,
   "num_active_channels": 0,
   "num_inactive_channels": 4,
   "address": [],
   "binding": [
      {
         "type": "ipv4",
         "address": "192.168.178.20",
         "port": 9835
      }
   ],
   "version": "v0.7.3-113-gdda792c",
   "blockheight": 604360,
   "network": "bitcoin",
   "msatoshi_fees_collected": 0,
   "fees_collected_msat": "0msat",
   "lightning-dir": "/home/user/.lightning_test"
}
`

rustyrussell:
Good catch, and great report.  Reproducing now, then will write fix.

Thanks for the report!

sumBTC:
I've seen the fix and understand why you're so glad I reported this: a c programmer's worst nightmare :-)

PR #3273, opened 2019-11-19
gossipd: don't crash if we have > 7000 stale short_channel_ids.

rustyrussell:
Fixes: #3269


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3273/commits/ce421fa092af468fd6c4a341f1037dedc63b2cbd

niftynei:
does this merit a changelog entry?

cdecker:
> does this merit a changelog entry?

I think so, it fixes something that was part of the prior release (0.7.3) and it's not experimental.

Issue #3243, opened 2019-11-05
Very slow block sync


JavierRSobrino:
I had to start a resync of several months after a blockchain corruption. Using the current RPC mechanism, it is extremely slow even in a fast machine.
My calculations are from 1 second up to 3 second per block, so if you are several months syncing, it can take 30 to 40 hours to sync in my example, which is just too much.

My proposition is to change to something faster, like ZMQ like LND uses, Neutrino, or something that allows fast traversal of blocks.

Version 0.7.2.

cdecker:
Are you running in docker or as root? If yes that's a known regression that was fixed in 0.7.3 with this pull request https://github.com/ElementsProject/lightning/pull/3101.

JavierRSobrino:
I'm running as docker. I'm still waiting for the docker image to update.

cdecker:
Right, I updated the docker images (`elementsproject/lightningd:latest` and `elementsproject/v0.7.3`) today. Sorry for the delay :-)

JavierRSobrino:
Thank you for the update. I have now 0.7.3 in my docker.
Unfortunately, it is still slow to sync, around 3s per block.

Saibato:
@cdecker  can confirm this on a no docker pi that remote rpc's the bitcoin core full node up to 3s per block.

JavierRSobrino:
Here is the requested information:

1. bip32_max_index|280|292|
2. 13
3. 139
4. 65811

As you can see, yes, they are very high in some cases.

No, I'm runing in a good performance server, an i7 with 4 cores and 8 threads. The lightning db is in an SSD, but the blockchain is in an HDD. It's a decent speed SATA HDD anyway.

Anyway, my node hasn't been the biggest or even close of the nodes out there. I'm sure there are hundreds of nodes which had many more channels.

Saibato:
Thanks @cdecker for the hint, tfw you think slow is so goooooooood. :smile: 

I think now, this must somehow be reducable from potential worse than quadratic to N log*N,  in case of sync to blockheight  could we use db speedup magic and check once at the end against the final state or am I thinking to short here ?

whitslack:
[...]

PR #3274, opened 2019-11-19
Never follow bitcoind backwards

rustyrussell:
This solves the underlying reason why #3243 was such a problem!

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1135,17 +1135,24 @@ def test_rescan(node_factory, bitcoind):
     assert l1.daemon.is_in_log(r'Adding block 31')
     assert not l1.daemon.is_in_log(r'Adding block 30')
 
-    # Restarting with a future absolute blockheight should just start with
-    # the current height
+    # Restarting with a future absolute blockheight should *fail* if we
+    # can't find that height
     l1.daemon.opts['rescan'] = -500000
     l1.stop()
     bitcoind.generate_block(4)
+    with pytest.raises(ValueError):
+        l1.start()
+
+    # Restarting with future absolute blockheight is fine if we can find it.
+    l1.daemon.opts['rescan'] = -105
+    oldneedle = l1.daemon.logsearch_start
     l1.start()
+    # This could occur before pubkey msg, so move search needle back.
+    l1.daemon.logsearch_start = oldneedle
     l1.daemon.wait_for_log(r'Adding block 105')
     assert not l1.daemon.is_in_log(r'Adding block 102')
 
 
-@pytest.mark.xfail(strict=True)
 def test_bitcoind_goes_backwards(node_factory, bitcoind):
     """Check that we refuse to acknowledge bitcoind giving a shorter chain without explicit rescan"""
     l1 = node_factory.get_node(may_fail=True, allow_broken_log=True)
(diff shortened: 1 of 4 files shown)


tests/test_misc.py
```diff
@@ -1145,6 +1145,57 @@ def test_rescan(node_factory, bitcoind):
...
+@pytest.mark.xfail(strict=True)
+def test_bitcoind_goes_backwards(node_factory, bitcoind):
+    """Check that we refuse to acknowledge bitcoind giving a shorter chain without explicit rescan"""
+    l1 = node_factory.get_node(may_fail=True, allow_broken_log=True)
+
+    bitcoind.generate_block(10)
+    sync_blockheight(bitcoind, [l1])
+    l1.stop()
+
+    # Now shrink chain (invalidateblock leaves 'headers' field)
```
cdecker: Just tried it myself, and indeed the `headers` remains unchanged after `invalidateblock`, however if `invalidateblock` is followed by a restart `headers` will reset to the expected value. Maybe that's a more stable way of going backwards than blowing away the underlying directory?

rustyrussell: Ah, that would have been easier!

rustyrussell:
Rebased to use the Decker technique for creating reorgs...

tests/test_misc.py
```diff
@@ -1145,6 +1145,60 @@ def test_rescan(node_factory, bitcoind):
...
+
+    bitcoind.rpc.invalidateblock(bitcoind.rpc.getblockhash(105))
+    bitcoind.rpc.stop()
+    TailableProc.stop(bitcoind)
+    bitcoind.start()
+    bitcoind.generate_block(5)
+
+    # It will ignore bitcoind and keep asking for block 110.
+    time.sleep(5)
+    assert l1.rpc.getinfo()['blockheight'] == 110
```
cdecker: This could be made into a `wait_for`. Happy to fix it in a cleanup though.

cdecker:
Seems we have a bit of collateral damage:

 - `test_db_dangling_peer_fix` uses a DB that was synced up to 205 while we migrate on a chain that has 101 blocks, need to generate 104 more blocks before starting the node, or specify a rescan.
 - `test_rescan` seems to believe it was once synced to a 500k chain, and is being started with a 101 block chain.

Other than that LGTM :+1:

tests/test_misc.py
rustyrussell: No, it can't: we want to make sure it's *still* at 110.

rustyrussell:
> Seems we have a bit of collateral damage:
> 
>     * `test_db_dangling_peer_fix` uses a DB that was synced up to 205 while we migrate on a chain that has 101 blocks, need to generate 104 more blocks before starting the node, or specify a rescan.
> 
>     * `test_rescan` seems to believe it was once synced to a 500k chain, and is being started with a 101 block chain.
> 
> 
> Other than that LGTM +1

Thanks, fixed.

I actually like the change in behavior (refusing to start) if you ask it to rescan from a not-yet-present block, so I've kept and tested that.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3274/commits/cdafcf35f8ccf88e789759c775f6d870a404144c

Fixed all the test collateral damage.

PR #3275, opened 2019-11-19
Add cross-compilation options only when requested

jsarenik:
Requesting them is done by setting BUILD=<target_arch>

Otherwise autotools (used by external dependencies like
libsodium) is not happy with setting cross-compilation
variables and may possibly lead to unexpected results.

Changelog-None

### DIFF
--- a/external/Makefile
+++ b/external/Makefile
@@ -4,6 +4,10 @@ SUBMODULES =					\
 	external/jsmn				\
 	external/libbacktrace
 
+ifdef BUILD
+MAKE_HOST_ADD := --host="$(MAKE_HOST)" --build="$(BUILD)"
+endif
+
 LIBSODIUM_HEADERS := external/libsodium/src/libsodium/include/sodium.h
 LIBWALLY_HEADERS := external/libwally-core/include/wally_bip32.h		\
 			external/libwally-core/include/wally_core.h	\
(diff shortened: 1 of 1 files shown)


jsarenik:
Sorry about this, I held it for too long. It was fresh long time ago, when I first started poking into c-lightning. Happy to see it unresolved yet, I decided to send this PR :)

cdecker:
Thanks @jsarenik for the PR. Seems quite all right but I'd like to get @rustyrussel's sign-off since he's our resident `Makefile` expert.

One thing that seems a bit strange is the naming of the variables. I think `TARGET` might be a better name for the `BUILD` variable, and the `MAKE_HOST_ADD` might be better called `CROSSCOMPILEOPTS` since we usually have names like `XYZOPTS` for things that get added to the command lines.

jsarenik:
> Thanks @jsarenik for the PR. Seems quite all right
> 
> One thing that seems a bit strange is the naming of the variables. I think `TARGET` might be a better name for the `BUILD` variable, and the `MAKE_HOST_ADD` might be better called `CROSSCOMPILEOPTS` since we usually have names like `XYZOPTS` for things that get added to the command lines.

Sure. Thank you for feed-back. I agree with `CROSSCOMPILEOPTS`, but the `TARGET` is a bit tricky. For one thing, it was `BUILD` variable used in c-lightning since some time ago so such a change could be considered as a regression for some. And `BUILD` also seems correct from other point of view:

See [Autoconf documentation on this topic](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Hosts-and-Cross_002dCompilation.html) which nicely explains how it is meant to be. Generally, I would change even more to not use the `$MAKE_HOST` variable by not setting `--host` at all. Just `--build` as the documentation clearly explains:

> They all default to the result of running config.guess, unless you specify either --build or --host. In this case, the default becomes the system type you specified. If you specify both, and they're different, configure enters cross compilation mode, so it doesn't run any tests that require execution.

So now I will change the patch from `MAKE_HOST_ADD` to `CROSSCOMPILEOPTS` and wait for @rustyrussell to decide on whether to leave `--host` together with `${MAKE_HOST}` or whether to drop it. I am all for dropping it and leaving only `--build` and will happily do all the needed testing to (dis)prove it.

Just for reference: [Another solid source from GNU Automake docs](https://www.gnu.org/software/automake/manual/html_node/Cross_002dCompilation.html)

jsarenik:
Just a note: doc/INSTALL.md might need to be changed, I am going to test it all now. First unchanged, just following the docs.

cdecker:
Great, thanks @jsarenik, I hadn't considered that `BUILD` is already part of the external interface, thus changing would be breaking things :+1:

jsarenik:
> ...I hadn't considered that `BUILD` is already part of the external interface, thus changing would be breaking things

Maybe it is not really. It is referenced only once in the `doc/INSTALL.md`.

jsarenik:
To refresh memories, I went through https://github.com/ElementsProject/lightning/pull/1533

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3275/commits/635a9d792df9c72d26c47350ef42b0bbc71c6b58

PR #3277, opened 2019-11-19
TLV Onion Fixes

niftynei:
Found these while manually verifying #3259

### DIFF
--- a/common/sphinx.c
+++ b/common/sphinx.c
@@ -147,7 +147,7 @@ static void sphinx_add_v0_hop(struct sphinx_path *path,
 	towire_u32(&buf, outgoing_cltv);
 	towire(&buf, padding, ARRAY_SIZE(padding));
 	assert(tal_bytelen(buf) == 32);
-	sphinx_add_raw_hop(path, pubkey, 0, buf);
+	sphinx_add_raw_hop(path, pubkey, SPHINX_V0_PAYLOAD, buf);
 }
 
 static void sphinx_add_tlv_hop(struct sphinx_path *path,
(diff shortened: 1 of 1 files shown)


common/sphinx.c
```diff
@@ -580,7 +580,7 @@ static void sphinx_parse_payload(struct route_step *step, const u8 *src)
 	}
 
 	/* Copy common pieces over */
-	step->raw_payload = tal_dup_arr(step, u8, src, raw_size + vsize, 0);
+	step->raw_payload = tal_dup_arr(step, u8, src + vsize, raw_size, 0);
```
cdecker: I just undid that, since we were stripping only the former realm byte, instead of the entire length prefix for TLV. I have a slight preference for keeping the length prefix in since it makes parsing on the plugin side uniform (no need to special case the plugin use from the proto based use-cases) and it matches up with serializing if we ever end up creating TLVs on the plugin side.

cdecker: See my change here: https://github.com/ElementsProject/lightning/commit/dda792c766aaa6db98208019356d8ac38a41b0fc

niftynei: > I just undid that, 

Right, I remembered seeing your change but couldn't remember where :D

I'm a bit confused as to the purpose of the length byte; previously it was the realm, now it's the length. It seems that the length of the raw packet is meta data at this point, so stripping it makes sense (i.e. you only need it to know how many bytes to pull into the `raw_payload`). 

> (no need to special case the plugin use from the proto based use-cases)
Stripping the realm/vsize bytes would be equally sane though; you're still going to need to special case legacy from 'tlv' in the plugin parsing (either with a flag or something etc). In both cases, it's not informative and can easily be inferred from context (i.e. flagged as legacy or lifted from measuring the length of the payload returned)

niftynei:
removed the controversial length trimming commit; just the enum not len fix now.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/091e530d332f1e12cf19b3f888eebf2b39f087d5

PR #3278, opened 2019-11-19
Split the TLV payload processing into parsing and validation

cdecker:
In order to be useful #3260 requires that we allow TLV payloads with types that we may not know how to handle. In this case the plugin needs to terminate the onion and handle it by either forwarding manually using `sendonion` or terminate it either successfully or fail it.

This PR splits the parsing of the TLV payload from the validation. Parsing just stashes the entire payload in a `tlv_payload` instance, with typed pointers to fields that we know how to parse and handle. Validation then goes through and checks that there are no TLV fields that we don't know how to handle, that the types are in the correct order and that we have all the fields that we need in order to forward. The validity is then only used if `lightningd` isn't explicitly told how to handle the HTLC and to forward or terminate with an internal invoice.


tools/gen/impl_template
```diff
@@ -211,6 +217,94 @@ ${static}const struct tlv_record_type tlvs_${tlv.name}[] = {
...
+				goto fail;
+			}
+		} else {
+			/* We didn't read from *cursor through a fromwire, so
+			 * update manually. */
+			*cursor += field.length;
+		}
+		/* We've read bytes in ->fromwire, so update max */
+		*max -= field.length;
+		tal_expand(&record->fields, &field, 1);
```
rustyrussell: tal_arr_expand() in utils.h does this a bit more compactly, BTW.

tools/gen/impl_template
```diff
@@ -211,6 +217,94 @@ ${static}const struct tlv_record_type tlvs_${tlv.name}[] = {
...
+	}
+	assert(tal_count(record->fields) > 0);
```
rustyrussell: If we ever hand a zero-length field in here, this will assert, right?  *max == 0?  That seems wrong.

tools/gen/impl_template
```diff
@@ -305,6 +305,39 @@ fail:
 	fromwire_fail(cursor, max);
 	return false;
 }
+bool ${tlv.name}_is_valid(const struct ${tlv.struct_name()} *record)
```
rustyrussell: Please have this return the problematic offset within the TLV? There's a FIXME for this requirement in the onion parsing code, where we're supposed to return the offset which was the problem in the error msg.

tools/gen/impl_template
```diff
@@ -305,6 +305,39 @@ fail:
...
+{
+	size_t numfields = tal_count(record->fields);
+	bool first = true;
+	u64 prev_type = 0;
+	for (size_t i=0; i<numfields; i++) {
+		struct tlv_field *f = &record->fields[i];
+		if (f->numtype % 2 == 0 && f->meta == 0) {
```
rustyrussell: f->meta == NULL please.

tools/gen/impl_template
```diff
@@ -305,6 +305,39 @@ fail:
...
+			 *  - if decoded `type`s are not monotonically-increasing:
+			 *    - MUST fail to parse the `tlv_stream`.
+			 */
+			if (f->numtype == prev_type)
+				SUPERVERBOSE("duplicate tlv type");
+			else
+				SUPERVERBOSE("invalid ordering");
+			return false;
+		}
+		first = false;
```
rustyrussell: This code doesn't actually work; we don't update prev_type!

We had tests for the generic code which I think we should replicate here which should have caught this?

common/sphinx.c
```diff
@@ -678,6 +678,48 @@ struct onionpacket *create_onionpacket(
 	return packet;
 }
 
+/**
+ * Helper to extract fields from the legacy or tlv payload into the top-level
+ * struct.
+ */
+static void route_step_decode(struct route_step *rs)
+{
+	if (rs->type == SPHINX_V0_PAYLOAD) {
```
rustyrussell: Make this a switch() to catch if we ever add another type?

lightningd/peer_htlcs.c
```diff
@@ -779,14 +779,28 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
 	enum htlc_accepted_result result;
 	enum onion_type failure_code;
 	u8 *channel_update;
+	bool valid;
 	result = htlc_accepted_hook_deserialize(buffer, toks, &payment_preimage, &failure_code, &channel_update);
 
+	/* In order to handle it internally the payload needs to have at least
+	 * the `outgoing_cltv_value`, the `amt_to_forward` and the
+	 * `short_channel_id` if we are forwarding. */
```
rustyrussell: Please quote the BOLT directly instead, so we know when this requirement changes (which it will!).

tools/gen/impl_template
cdecker: Ouch, sorry. Copy pasted it without checking.

tools/gen/impl_template
cdecker: This one slipped in, it checks whether we had any TLV fields,not whether one of the fields had zero-length. However the check is wrong since an empty TLV is valid per spec, even though not very sensible.

tools/gen/impl_template
cdecker: Ah, was looking for that but couldn't find it so I thought I was misremembering the name xD

tools/gen/impl_template
cdecker: We can return the exact offset of the offending field in `fromwire_tlv_payload` which has the locations, while here we would have to track the offset by accumulating the `varint` lengths and field lengths while verifying. I can add the offset in the `fromwire` code and just add the field index that we get upset about if you'd like :-)

cdecker: Might make more sense anyway, since we can then say things like "Unknown even type x in field y" instead of saying I got upset somewhere at byte `z`.

tools/gen/impl_template
cdecker: I'll migrate the `run-tlvstream` tests to use the typesafe wrapper and remove the old ones. :+1:

lightningd/peer_htlcs.c
```diff
@@ -779,14 +779,28 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
...
+	valid = rs->amt_to_forward != NULL && rs->outgoing_cltv != NULL &&
+		(rs->nextcase == ONION_END ||
+		 (rs->nextcase == ONION_FORWARD && rs->forward_channel != NULL));
+
+	/* In addition we also enforce the TLV validity rules:
+	 *  - No unknown even types
+	 *  - Types in monotonical non-repeating order
+	 */
+	valid = valid && (rs->type == SPHINX_V0_PAYLOAD || tlv_payload_is_valid(rs->payload.tlv));
```
niftynei: seems like it'd be nicer to have a separate method for `validity check`, instead of making it all inline here.

lightningd/peer_htlcs.c
```diff
@@ -779,14 +779,28 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
...
+
 	switch (result) {
 	case htlc_accepted_continue:
-		if (rs->type == SPHINX_TLV_PAYLOAD && !tlv_payload_is_valid(rs->payload.tlv)) {
-			log_debug(channel->log, "Failing HTLC because of an invalid TLV payload");
+		if (!valid) {
```
niftynei: you could just call the `valid()` method here

lightningd/peer_htlcs.c
cdecker: Added a `htlc_accepted_can_continue` function which tells us whether we can continue with the route-step if we're told to by the hook.

cdecker:
Updated the PR to address the feedback:

 - @rustyrussell  here's the [range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3278/range_diff/78ca90ee6664bbdf8695d5fff81dee16c34e1991..df4c187953fdabe70934528dbb530c59792c7425) since your review
 - @niftynei  here's the [range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3278/range_diff/08fc431d9b114cf5ddda9f8b9a3b13ef2706cbde..df4c187953fdabe70934528dbb530c59792c7425) since your review

Notice that there are two new commit that will not be shown in the range-diff: 02fb929c3 and df4c18795 :wink:

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/df4c187953fdabe70934528dbb530c59792c7425

Note: Travis picked up an unrelated race, causing #3286 ...

PR #3279, opened 2019-11-20
More local chainparams removal

rustyrussell:
This leaves bitcoin/ using explicit chainparams, and some common/ files; subdaemons now all use the global.

Changelog-None


cdecker:
AC a451854986660e1aa4f177869a1ff628f86fc54c

PR #3280, opened 2019-11-21
openchannel hook: add new `close_to` field

niftynei:
I wanted to add this to dual_funding, but decided that it's better to not cram too much more into that than need be, so instead did this off of master.  This is more of a 'nice to have' now, but will be much more important come dual funded channels.

Rounds out the application of `upfront_shutdown_script`, allowing
an accepting node to specify a close_to address.

Prior to this, only the opening node could specify one.

Changelog-Added: Plugins: Allow the 'accepter' to specify an upfront_shutdown_script for
a channel via a `close_to` field in the result.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -577,6 +577,21 @@ the string `reject` or `continue`.  If `reject` and
 there's a member `error_message`, that member is sent to the peer
 before disconnection.
 
+For a 'continue'd result, you can also include a `close_to` address,
+which will be used as the output address for a mutual close transaction.
+
+e.g.
+
+```json
+{
+    "result": "continue",
+    "close_to": "bc1qlq8srqnz64wgklmqvurv7qnr4rvtq2u96hhfg2"
+}
+```
+
+Note that `close_to` must be a valid address for the current chain; an invalid address will cause the node to exit with an error.
+
+
 #### `htlc_accepted`
 
 The `htlc_accepted` hook is called whenever an incoming HTLC is accepted, and
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -455,7 +455,7 @@ static void opening_fundee_finished(struct subd *openingd,
 	u32 feerate;
 	u8 channel_flags;
 	struct channel *channel;
-	u8 *remote_upfront_shutdown_script;
+	u8 *remote_upfront_shutdown_script, *local_upfront_shutdown_script;
 	struct per_peer_state *pps;
 
 	log_debug(uc->log, "Got opening_fundee_finish_response");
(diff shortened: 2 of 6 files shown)


lightningd/opening_control.c
```diff
@@ -779,17 +781,43 @@ static void openchannel_hook_cb(struct openchannel_hook_payload *payload,
...
+			if (t) {
+				switch (json_to_address_scriptpubkey(tmpctx, chainparams,
+								     buffer, t,
+								     &our_upfront_shutdown_script)) {
+					case ADDRESS_PARSE_UNRECOGNIZED:
+						fatal("Plugin returned an invalid response to the"
+						      " openchannel.close_to hook: %.*s",
+						      t->end - t->start, buffer + t->start);
+					case ADDRESS_PARSE_WRONG_NETWORK:
+						fatal("Plugin returned invalid respnse to the"
```
rustyrussell: s/respnse/response/

lightningd/opening_control.c
```diff
@@ -808,7 +836,8 @@ static void opening_got_offer(struct subd *openingd,
 	if (peer_active_channel(uc->peer)) {
 		subd_send_msg(openingd,
 			      take(towire_opening_got_offer_reply(NULL,
-					  "Already have active channel")));
+					  "Already have active channel",
+					  tal_arr(tmpctx, u8, 0))));
```
rustyrussell: NULL works here.

cdecker:
Your Changelog entry is likely not working as expected: the script (currently) doesn't know how to handle multiple lines:

```
Changelog-Added: Plugins: Allow the 'accepter' to specify an upfront_shutdown_script for
a channel via a `close_to` field in the openchannel hook result
```

The entry would end at `for`. If you have a good idea on how to handle this correctly I'm open for suggestions. We could either mandate that the entries must be the last in the commit (not very portable, and would not allow for multiple entries in a single commit), or we could require continuations to be indented with some whitespace (doesn't really work for editors that wrap automatically).

doc/PLUGINS.md
```diff
@@ -577,6 +577,21 @@ the string `reject` or `continue`.  If `reject` and
...
+e.g.
+
+```json
+{
+    "result": "continue",
+    "close_to": "bc1qlq8srqnz64wgklmqvurv7qnr4rvtq2u96hhfg2"
+}
+```
+
+Note that `close_to` must be a valid address for the current chain; an invalid address will cause the node to crash.
```
cdecker: Shall we call it "exit with an error" instead of "crash" it is after all a safety abort :-)

cdecker:
Very nice change, yet another step towards full external on-chain wallet support :+1:

niftynei:
> If you have a good idea on how to handle this correctly I'm open for suggestions. 

The 'traditional' way to signal that "i'm continuing onto the next line" is to use a `\` to 'escape' the newline character. This would be my suggestion.

rustyrussell:
> Your Changelog entry is likely not working as expected: the script (currently) doesn't know how to handle multiple lines:

Yech, stick with a single line for now.  Release Captain has to neaten them anyway...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3280/commits/d22b1fd05684cf12efed1343792c10d7bd8b9d1d

Issue #3281, opened 2019-11-21
Duplicated keys in JSON rpc interface.


m-schmoock:
### Issue and Steps to Reproduce

I saw that the command outputs duplicated config entries (i.e. the `"plugin"` switch) as duplicated keys in the JSON output. This is only more or less valid, while the formal specification allows this, all the implementations complain about it and most will filter out duplicated entries by picking the first or last one. We can see this by pushing the output of `listconfigs` through the neat command-line tool `jq`:

```bash
lightning-cli listconfigs
```

```JSON
 lightning-cli listconfigs
{
   "# version": "v0.7.3",
   "lightning-dir": "/home/btc/.lightning",
   "wallet": "sqlite3:///home/btc/.lightning/lightningd.sqlite3",
   "plugin": "/home/btc/lightning.git/lightningd/../plugins/pay",
   "plugin": "/home/btc/lightning.git/lightningd/../plugins/autoclean",
   "plugin": "/home/btc/lightning.git/lightningd/../plugins/fundchannel",
   "..." : "..."
}
```
---------------

```bash
lightning-cli listconfigs | jq
```

```JSON
{
  "# version": "v0.7.3",
  "lightning-dir": "/home/btc/.lightning",
  "wallet": "sqlite3:///home/btc/.lightning/lightningd.sqlite3",
  "plugin": "/home/btc/.lightning/plugins/summary/summary.py",
  "network": "bitcoin",
  "..." : "..."
}
```

### Impact

External applications that are using the RPC interface (such as web interfaces, remote wallets, ...) will have a hard time to access the duplicated entries when their JSON library do not support this.

### Suggestion

We could replace duplicated keys by just one key having an array of the the values that were duplicated, example:

```JSON
{
   "# version": "v0.7.3",
   "lightning-dir": "/home/btc/.lightning",
   "wallet": "sqlite3:///home/btc/.lightning/lightningd.sqlite3",
   "plugin": [
      "/home/btc/lightning.git/lightningd/../plugins/pay",
      "/home/btc/lightning.git/lightningd/../plugins/autoclean",
      "/home/btc/lightning.git/lightningd/../plugins/fundchannel",
      "..."
   ],
   "network": "bitcoin",
   "..." : "..."
}
```

### Considerations

- Backwards compatibility may be affected, though a lot of applications were missing out the duplicates anyway already.
[...]

Issue #3206, opened 2019-10-25
Show plugin configs in `listconfigs`


niftynei:
For maximum correctness, `listconfigs` should show all configs, including plugins.

m-schmoock:
@niftynei how do you want to have the plugin configurations be formatted in the JSON output?
Note that there is aready a duplicated key `"plugin"` per plugin in the listconfigs output:

```JSON
ln_cli listconfigs
{
   "# version": "v0.7.3-126-g654faa6-modded",
   "lightning-dir": "/home/will/.lightning",
   "wallet": "sqlite3:///home/will/.lightning/lightningd.sqlite3",
   "plugin": "/home/will/projects/lightning.git/lightningd/../plugins/autoclean",
   "plugin": "/home/will/projects/lightning.git/lightningd/../plugins/fundchannel",
   "plugin": "/home/will/projects/lightning.git/lightningd/../plugins/pay",
   "plugin": "/home/will/.lightning/plugins/sendinvoiceless/sendinvoiceless.py",
```

For plugins, we could change the output to something like this:
```JSON
{
   "# version": "v0.7.3-126-g654faa6-modded",
   "lightning-dir": "/home/will/.lightning",
   "wallet": "sqlite3:///home/will/.lightning/lightningd.sqlite3",
   "plugins": {
      "/home/will/projects/lightning.git/lightningd/../plugins/autoclean" : { },
      "/home/will/projects/lightning.git/lightningd/../plugins/fundchannel" : { },
      "/home/will/projects/lightning.git/lightningd/../plugins/pay" : { },
      "/home/will/.lightning/plugins/sendinvoiceless/sendinvoiceless.py" : {
         "cltv-final": "10",
         "fee-base": "null",
         "fee-per-satoshi": "null"
      }
   }
}

m-schmoock:
Can be closed

PR #3283, opened 2019-11-21
feat: restructure plugin and options in listconfigs #3206

m-schmoock:
Addresses https://github.com/ElementsProject/lightning/issues/3206 and https://github.com/ElementsProject/lightning/issues/3281

This will restructure the `lightning-cli listconfigs` output in a way that plugins are not duplicated and have their options within a substructure like this:

```JSON
{
   "# version": "v0.7.3-141-gabdc4af",
   "lightning-dir": "/home/will/.lightning",
   "wallet": "sqlite3:///home/will/.lightning/lightningd.sqlite3",
   "plugins": [
      {
         "name": "autoclean",
         "path": "/home/will/projects/lightning.git/lightningd/../plugins/autoclean",
         "options": {
            "autocleaninvoice-cycle": null,
            "autocleaninvoice-expired-by": null
         }
      },
      {
         "name": "fundchannel",
         "path": "/home/will/projects/lightning.git/lightningd/../plugins/fundchannel"
      },
      {
         "name": "pay",
         "path": "/home/will/projects/lightning.git/lightningd/../plugins/pay"
      },
      {
         "name": "sendinvoiceless.py",
         "path": "/home/will/.lightning/plugins/sendinvoiceless/sendinvoiceless.py",
         "options": {
            "cltv-final": "10",
            "fee-base": "null",
            "fee-per-satoshi": "null"
         }
      },
      {
         "name": "rebalance.py",
         "path": "/home/will/.lightning/plugins/rebalance/rebalance.py",
         "options": {
            "cltv-final": "10"
         }
      },
      {
         "name": "summary.py",
         "path": "/home/will/.lightning/plugins/summary/summary.py",
         "options": {
            "summary-currency": "USD",
            "summary-currency-prefix": "USD $"
         }
      },
      {
         "name": "drain.py",
         "path": "/home/will/.lightning/plugins/drain/drain.py",
         "options": {
            "cltv-final": "10"
         }
      }
   ],
   "network": "testnet",
   "allow-deprecated-apis": false,
   "..." : "..."
}
```

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1086,9 +1086,49 @@ void json_add_opt_plugins(struct json_stream *response,
 			  const struct plugins *plugins)
 {
 	struct plugin *p;
-	list_for_each(&plugins->plugins, p, list) {
-		json_add_string(response, "plugin", p->cmd);
+	struct plugin_opt *opt;
+	const char *plugin_name;
+	const char *opt_name;
+
+	/* DEPRECATED: duplicated JSON "plugin" entries */
+	if (deprecated_apis) {
+		list_for_each(&plugins->plugins, p, list) {
+			json_add_string(response, "plugin", p->cmd);
+		}
 	}
+
+	/* we output 'plugins' and their options as an array of substructures */
+	json_array_start(response, "plugins");
+	list_for_each(&plugins->plugins, p, list) {
+		json_object_start(response, NULL);
+		json_add_string(response, "path", p->cmd);
+
+		/* FIXME: use executables basename until plugins can define their names */
+		plugin_name = path_basename(NULL, p->cmd);
+		json_add_string(response, "name", plugin_name); // basename(p->cmd));
+		tal_free(plugin_name);
+
+		if (!list_empty(&p->plugin_opts)) {
+			json_object_start(response, "options");
+			list_for_each(&p->plugin_opts, opt, list)
+			{
+				/* Trim the `--` that we added before */
+				opt_name = opt->name + 2;
+				if (opt->value->as_bool) {
+					json_add_bool(response, opt_name, opt->value->as_bool);
+				} else if (opt->value->as_int) {
+					json_add_bool(response, opt_name, opt->value->as_int);
+				} else if (opt->value->as_str) {
+					json_add_string(response, opt_name, opt->value->as_str);
+				} else {
+					json_add_null(response, opt_name);
+				}
+			}
+			json_object_end(response);
+		}
+		json_object_end(response);
+	}
+	json_array_end(response);
 }
 
 /**
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -742,10 +742,23 @@ def test_listconfigs(node_factory, bitcoind, chainparams):
 
     # Test one at a time.
     for c in configs.keys():
-        if c.startswith('#'):
+        if c.startswith('#') or c.startswith('plugins'):
             continue
         oneconfig = l1.rpc.listconfigs(config=c)
         assert(oneconfig[c] == configs[c])
+
+
+def test_listconfigs_plugins(node_factory, bitcoind, chainparams):
+    l1 = node_factory.get_node()
+
+    # assert that we have pay plugin and that plugins have a name and path
+    configs = l1.rpc.listconfigs()
+    assert configs['plugins']
+    assert len([p for p in configs['plugins'] if p['name'] == "pay"]) == 1
+    for p in configs['plugins']:
+        assert p['name'] and len(p['name']) > 0
+        assert p['path'] and len(p['path']) > 0
+        assert os.path.isfile(p['path']) and os.access(p['path'], os.X_OK)
 
 
 def test_multirpc(node_factory):


m-schmoock:
We could think about add a "name" feature to a plugin so we dont have to use the long command path-name as key...

m-schmoock:
@cdecker how about adding a "name" to the plugin interface. For all the old plugins we guess the name by inspecting the commands paths lasts elements:

- if directory, take submost directory name
- if filename, take this name and remove extensions like '.py' and such

If we do this we can make this nicely without update all the plugins first...

lightningd/plugin.c
```diff
@@ -1086,9 +1086,24 @@ void json_add_opt_plugins(struct json_stream *response,
 			  const struct plugins *plugins)
 {
 	struct plugin *p;
+	struct plugin_opt *opt;
+	const char *name;
+
+	// we output 'plugins' and their config as substructure:
+	// "plugins" : { "name" -> { "param" : "value" }, ... }
```
cdecker: We usually use `/* ... */` style comments.

cdecker:
> @cdecker how about adding a "name" to the plugin interface. For all the old plugins we guess the name by inspecting the commands paths lasts elements:
> 
> * if directory, take submost directory name
> * if filename, take this name and remove extensions like '.py' and such
> 
> If we do this we can make this nicely without update all the plugins first...

I think this is a great idea, though we should also still display the path somewhere so we have a mapping between the file being executed and the (self-assigned) alias.

lightningd/plugin.c
m-schmoock: Yes sorry, I almost ever miss that!

m-schmoock:
> > @cdecker how about adding a "name" to the plugin interface. For all the old plugins we guess the name by inspecting the commands paths lasts elements:
> > 
> > * if directory, take submost directory name
> > * if filename, take this name and remove extensions like '.py' and such
> > 
> > If we do this we can make this nicely without update all the plugins first...
> 
> I think this is a great idea, though we should also still display the path somewhere so we have a mapping between the file being executed and the (self-assigned) alias.

Okay, I added the self assigned alias as key in the `plugins` structure. I furthermore added a string `_cmd`  to a plugins substructure that holds the full path of the plugins executable.

m-schmoock:
The output now looks like this:

```JSON
{
   "# version": "v0.7.3-127-g3237119-modded",
   "lightning-dir": "/home/will/.lightning",
   "wallet": "sqlite3:///home/will/.lightning/lightningd.sqlite3",
   "plugins": {
      "autoclean": {
         "_cmd": "/home/will/projects/lightning.git/lightningd/../plugins/autoclean"
      },
      "fundchannel": {
         "_cmd": "/home/will/projects/lightning.git/lightningd/../plugins/fundchannel"
      },
      "pay": {
         "_cmd": "/home/will/projects/lightning.git/lightningd/../plugins/pay"
      },
      "sendinvoiceless": {
         "_cmd": "/home/will/.lightning/plugins/sendinvoiceless/sendinvoiceless.py",
         "cltv-final": "10",
         "fee-base": "null",
         "fee-per-satoshi": "null"
      },
      "rebalance": {
         "_cmd": "/home/will/.lightning/plugins/rebalance/rebalance.py",
         "cltv-final": "10"
      },
      "summary": {
         "_cmd": "/home/will/.lightning/plugins/summary/summary.py",
         "summary-currency": "USD",
         "summary-currency-prefix": "USD $"
      },
      "drain": {
         "_cmd": "/home/will/.lightning/plugins/drain/drain.py",
         "cltv-final": "10"
      }
   },
   "network": "testnet",
   "allow-deprecated-apis": true,
   "..." : "..."
}
```

m-schmoock:
We can address the dedicated plugin naming to a separate pull request...

lightningd/plugin.c
```diff
@@ -1086,9 +1087,35 @@ void json_add_opt_plugins(struct json_stream *response,
 			  const struct plugins *plugins)
 {
 	struct plugin *p;
+	struct plugin_opt *opt;
+	const char *opt_name;
+	const char *plugin_name;
+	char *plugin_ext;
+
+	/* we output 'plugins' and their config as substructure:
+	 * "plugins" : { "name" -> { "param" : "value" }, ... }  */
```
niftynei: i almost think it'd be nicer to put the name inside the object, so instead of a list of keyed objects we get a set of uniform objects. (this is easier on static parsers like in Go, i think). actually, it's probably even easier if we bucket the configs into their own dict then..

e.g.

```
plugins: [
    {
        "name": "hello",
        "path": "../hello/path../path",
        "options": {
            "option1": "value",
            "option2": "value"
        }
        },{
    ...
    }
]
```

lightningd/plugin.c
```diff
@@ -1086,9 +1087,35 @@ void json_add_opt_plugins(struct json_stream *response,
...
+
+		/* guess a suitable plugin name: basename without extension */
+		plugin_name = tal_strdup(p, basename(p->cmd));
+		plugin_ext = strrchr(plugin_name, '.');
+		if (plugin_ext != NULL)
+			plugin_ext[0] = 0;
+		json_object_start(response, plugin_name);
+		tal_free(plugin_name);
+
+		json_add_string(response, "_cmd", p->cmd);
```
niftynei: suggestion to change this to `path` or `binary_path`

lightningd/plugin.c
```diff
@@ -1086,9 +1087,35 @@ void json_add_opt_plugins(struct json_stream *response,
...
+		list_for_each(&p->plugin_opts, opt, list) {
+			/* Trim the `--` that we added before */
+			opt_name = opt->name + 2;
+			if (opt->value->as_str) {
```
niftynei: we should probably still include it if it's not a string? or at least include the option name?

niftynei: or at least add a FIXME to change this when/if we allow non-string values for plugin options.

lightningd/plugin.c
```diff
@@ -1086,9 +1087,35 @@ void json_add_opt_plugins(struct json_stream *response,
...
+	struct plugin_opt *opt;
+	const char *opt_name;
+	const char *plugin_name;
+	char *plugin_ext;
+
+	/* we output 'plugins' and their config as substructure:
+	 * "plugins" : { "name" -> { "param" : "value" }, ... }  */
+	json_object_start(response, "plugins");
 	list_for_each(&plugins->plugins, p, list) {
-		json_add_string(response, "plugin", p->cmd);
```
niftynei: if `deprecated_apis` is flagged on, we should still print this old style.

note that you'll want to add a `Changelog-Deprecated` note to the commit message for this.

lightningd/plugin.c
niftynei: as an additional addendum, calling them 'options' is nice because it matches the language we use in the plugins documentation.

m-schmoock: ok. thanks for the feedback. done in latest force push

lightningd/plugin.c
m-schmoock: Okay, bool and ints are now also returned. Anything other than string, bool and int will be mapped to a `null` so we have at least the key.

m-schmoock:
```
        "options": {
            "autocleaninvoice-cycle": null,
            "autocleaninvoice-expired-by": null
         }
```

m-schmoock:
I wonder if we can improve on the untyped values, what are those, date/times?:
```
            "autocleaninvoice-cycle": null,
            "autocleaninvoice-expired-by": null
```

lightningd/plugin.c
```diff
@@ -1086,9 +1087,53 @@ void json_add_opt_plugins(struct json_stream *response,
...
+	}
+
+	/* we output 'plugins' and their options as an array of substructures */
+	json_array_start(response, "plugins");
 	list_for_each(&plugins->plugins, p, list) {
-		json_add_string(response, "plugin", p->cmd);
+		json_object_start(response, NULL);
+
+		/* guess a suitable plugin name: basename without extension */
+		plugin_name = tal_strdup(p, basename(p->cmd));
```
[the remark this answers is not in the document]
darosior: And use `NULL` as parent if you `free` it just after

lightningd/plugin.c
```diff
@@ -1086,9 +1087,53 @@ void json_add_opt_plugins(struct json_stream *response,
...
+		plugin_ext = strrchr(plugin_name, '.');
+		if (plugin_ext != NULL)
+			plugin_ext[0] = 0;
```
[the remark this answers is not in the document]
m-schmoock: Okay, lets take the filename with extension and replace it optionally if/when plagins can define their name otherwise than via the executable name.

m-schmoock:
Okay, lets use the basename including extension for now until plugin can set their name properly.

darosior:
ACK 02059ee3f8120651c44a70182c2fd8b633505720

darosior:
@m-schmoock Regarding
> > How do we handle python plugin parameters more correctly that are None or non-string types?

> By setting them to null Pylightning side ?..

I gave a quick look to `pyln` and they seem to actually be set to `null` : https://github.com/ElementsProject/lightning/blob/6defc69482d5c88796ba76c5ed7f6dfee0c02672/contrib/pyln-client/pyln/client/plugin.py#L416-L419 ...

PR #3284, opened 2019-11-21
Cleanups

cdecker:
Just a couple of cleanups that didn't fit anywhere else:

 - Fix for the TLV check in `EXPERIMENTAL_FEATURES` mode
 - Memory leak check works only in developer mode
 - A test was not prepared to handle PGSQLs out-of-order results
 - Teaching github to leave generated man-pages out of the diff (we have the source for that anyway)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,7 @@
 Makefile text eol=lf
 configure text eol=lf
 .gitmodules text eol=lf
+# The following files are generated and should not be shown in the
+# diffs by default on Github.
+doc/lightning*.7 linguist-generated=true
+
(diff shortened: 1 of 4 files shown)


.gitattributes
```diff
@@ -4,3 +4,7 @@
 Makefile text eol=lf
 configure text eol=lf
 .gitmodules text eol=lf
+# The following files are generated and should not be shown in the
+# diffs by default on Github.
+doc/lightning*.7 linguist-generated=true
```
niftynei: Agree with this in principle but this was helpful once when my generator code was out of date and @rustyrussell noticed.

cdecker: Good point, we might want to hold off on this one.

rustyrussell: What's the chance it'll happen twice?  :)

I'm OK with it, TBH.  Should probably include doc/lightning-cli.1 though.

PR #3285, opened 2019-11-21
 Let canonicalize correctly handle a NULL pointer 

sumBTC:
This is my first pull request ever. Hope it makes sense and I'm doing this right.

### DIFF
--- a/tools/check-bolt.c
+++ b/tools/check-bolt.c
@@ -24,6 +24,9 @@ static char *canonicalize(char *str)
 	char *to = str, *from = str;
 	bool have_space = true;
 
+	if (!str)
+		return(str);
+
 	while (*from) {
 		if (cisspace(*from)) {
 			if (!have_space)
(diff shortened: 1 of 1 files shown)


rustyrussell:
Hi!  The patch itself seems fine, but I'm not sure what it solves?  If it can't open the bolt file, I guess it just ignores it now?

sumBTC:
It doesn't really solve anything in how `canonicalize` is  being used, I just thought it's more complete as a function by itself, for example, if someone wants to copy-past the code and use it somewhere else.

rustyrussell:
Right; by C convention though, string functions don't accept NULL, but instead give a nice diagnostic from the operating system by crashing :)

But if you're touching check-bolt, something which would be cool would be changing it so "..." refused to match across `#` which would stop it from matching a completely different section.  Might break too much stuff though!

sumBTC:
I get it. I'll close the pull request.

PR #3286, opened 2019-11-22
channeld: allow transient negative balance.

rustyrussell:
Travis randomly picked up an error in test_feerate_stress:
**BROKEN** 0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518-channeld-chan#1: Cannot add htlc #0 10000msat to LOCAL (version a2541b9-modded)

This is because it hit an unlikely corner case involving applying multiple HTLCs
(similar to the previous c96cee9b8d0333d90edd04e309dfc229b5ee1a49).

In this case, the test sends a 500,000,000 "balancing" setup payment L1->L2.
It waits for L2 to get the preimage (which is the when pay() helper returns),
but crucially, it starts spamming with HTLCs before that HTLC is completely
removed.

From L2's point of view, the setup HTLC is in state RCVD_REMOVE_REVOCATION;
gone from L1's commitment tx, but still waiting for the commitment_signed
from L1 to remove it from L2's.

Note that each side keeps a local and remove view of both sides' current
balances: at this point, L2's view is REMOTE: "500,000,000 to L1, 499,900,000
to L2", LOCAL: "500,000,000 to L1, 0 to L2".

L2 sends a 10,000 msat HTLC to L1: legal, since L1 will allow it,
then the commitment_signed.  L1 sends the revoke-and-ack for this,
*then* belatedly follows with the commitment_signed which both completes the
removal of the setup HTLC and adds the new one.

But L2 processes the HTLCs in hashtable (i.e. random) order: so if it
tries to apply its own HTLC first, it freaks out because it doesn't have
funds in its local view.

Changelog-Fixed: Unlikely corner case is simultanous HTLCs near balance limits fixed.

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -32,6 +32,59 @@ static void memleak_help_htlcmap(struct htable *memtable,
 }
 #endif /* DEVELOPER */
 
+/* This is a dangerous thing!  Because we apply HTLCs in many places
+ * in bulk, we can temporarily go negative.  You must check balance_ok()
+ * at the end! */
+struct balance {
+	s64 msat;
+};
+
+static void to_balance(struct balance *balance,
+		       const struct amount_msat msat)
+{
+	balance->msat = msat.millisatoshis; /* Raw: balance */
+	assert(balance->msat >= 0);
+}
+
+/* What does adding the HTLC do to the balance for this side (subtracts) */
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3286/commits/90a9bd2887e1b1fc5afda01f34ba51eed4356b10

PR #3288, opened 2019-11-24
don't enforce description_hash checking

fiatjaf:
Although BOLT11 says "a reader MUST check" I'm not sure "reader" refers specifically to the Lightning server.

Sometimes the reader may be an application and the application may want to decode the invoice first before looking for the full description somewhere else and then proceeding to check either on its own or by calling `decodepay` again.

Sometimes the reader is a final payer, but there are some circumstances, like a custodial wallet provider or some other [odd situations](https://devpost.com/software/txawaitbot) in which the Lightning server is not the _final payer_ and thus not the "reader" that should care about this.

Sometimes the "full description" is not a UTF-8 string. In these cases they may still be checked against the `description_hash` but it will be impossible to pass it to lightningd.

I don't know, just think that failing when a `description` to be checked is not given is too brutal.

Changelog-None

### DIFF
--- a/common/bolt11.c
+++ b/common/bolt11.c
@@ -704,7 +704,7 @@ struct bolt11 *bolt11_decode(const tal_t *ctx, const char *str,
         if (!have_p)
                 return decode_fail(b11, fail, "No valid 'p' field found");
 
-        if (have_h) {
+        if (have_h && description) {
                 struct sha256 sha;
 
                 /* BOLT #11:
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3288/commits/0ed0c1394a2be21b30ad7e8ad42ec2e6a746638c

PR #3290, opened 2019-11-24
deprecate --always-use-proxy 

Saibato:
The proxy can be assumed in a later version to be a Tor proxy by default.

Therefore we can then by default disable all local dns lookup and resolve all names over Tor, if the proxy option is used.
Mark therefore now the  --always-use-proxy deprecated and define a new option  --tor-proxy-only-tor
The user can override this later default behavior with the new option --tor-proxy-only-tor

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -780,7 +780,7 @@ static void try_connect_one_addr(struct connecting *connect)
 
 	/* This creates the new connection using our fd, with the initialization
 	 * function one of the above. */
-	if (use_proxy)
+	if (use_proxy || connect->daemon->use_proxy_always)
 		notleak(io_new_conn(connect, fd, conn_proxy_init, connect));
 	else
 		notleak(io_new_conn(connect, fd, conn_init, connect));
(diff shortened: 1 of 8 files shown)


cdecker:
I don't quite follow, why not make the `--always-use-proxy` just default to `true` instead, allowing users to use `--always-use-proxy=false` to explicitly opt out? Adding a new option and deprecating the old one is way more disruptive.

Saibato:
> why not make the `--always-use-proxy` just default to `true` instead,

Whatever we do in essence that is the inner code change to the bound use_proxy_always var.  Some disruption or failure reports of unresolved names will occur.
If we change it the hard way, without deprecate, my view is that this will even be more disruptive, since  --always-use-proxy=true disables all local dns.

The effective code change in lines is fewer if we do it the hard way and just set it true.
3 lines AFAICS
But I fear we will have a lot of lamers if we do it that way.
My prefered path is deprecate, show a warning about what always-use-proxy not true means in terms of leak, change the docu to be more explicit about add an extra ip proxy and then make the cut.

PR #3292, opened 2019-11-25
tool/hsmtool: add a 'guesstoremote' command

darosior:
This is the `static_remotekey`-only version of #3217 (since I noticed `to_remote` keys are (/were) actually derived from the *local* `per_commitment_point`, by the way if someone [could explains me why](https://github.com/ElementsProject/lightning/issues/2665#issuecomment-547409859) I'd be interested ^^).

I added a note to the README about it since all major implementations [will](https://github.com/ACINQ/eclair/pull/1141) implement it, it's likely that most of the newly created channels will enable `option_static_remotekey` in the near future and that this might be useful.

### DIFF
--- a/tools/hsmtool.c
+++ b/tools/hsmtool.c
@@ -5,6 +5,7 @@
 #include <ccan/read_write_all/read_write_all.h>
 #include <ccan/tal/path/path.h>
 #include <ccan/str/str.h>
+#include <common/bech32.h>
 #include <common/derive_basepoints.h>
 #include <common/node_id.h>
 #include <common/type_to_string.h>
@@ -29,6 +30,8 @@ static void show_usage(void)
 	printf("	- decrypt <path/to/hsm_secret> <password>\n");
 	printf("	- encrypt <path/to/hsm_secret> <password>\n");
 	printf("	- dumpcommitments <node id> <channel dbid> <depth> "
+	       "<path/to/hsm_secret> [hsm_secret password]\n");
+	printf("	- guesstoremote <P2WPKH address> <node id> <tries> "
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
 	exit(0);
 }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -210,6 +211,10 @@ You can encrypt the `hsm_secret` content (which is used to derive the HD wallet'
 
 If you encrypt your `hsm_secret`, you will have to pass the `--encrypted-hsm` startup option to `lightningd`. Once your `hsm_secret` is encrypted, you __will not__ be able to access your funds without your password, so please beware with your password management. Also beware of not feeling too safe with an encrypted `hsm_secret`: unlike for `bitcoind` where the wallet encryption can restrict the usage of some RPC command, `lightningd` always need to access keys from the wallet which is thus __not locked__ (yet), even with an encrypted BIP32 master seed.
 
+### Data loss
+
+If you lose data (likely corrupted `lightningd.sqlite3`) about a channel __with `option_static_remotekey` enabled__, you can, after your peer unlateraly closed the channel, use `tools/hsmtool` with the `guesstoremote` command to (try to) recover your output from your peer's published commitment transaction.
+
```
cdecker: Not sure this belongs on the front-page. It's not like a corruption of `lightningd.sqlite3` is an everyday occurrence and this would be far better suited for a faq in the doc directory.

tools/hsmtool.c
```diff
@@ -319,5 +392,16 @@ int main(int argc, char *argv[])
 		                              argv[5], argv[6]);
 	}
 
+	if (streq(method, "guesstoremote")) {
+		/*   address    node_id    depth    hsm_secret  ?password? */
+		if (!(argv[2] && argv[3] && argv[4] && argv[5]))
+			show_usage();
```
cdecker: This will likely upset valgrind. Better to check `argc` before attempting to access the array elements.

tools/hsmtool.c
```diff
@@ -285,6 +288,76 @@ static int dump_commitments_infos(struct node_id *node_id, u64 channel_id,
...
+
+	secp256k1_ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY
+	                                         | SECP256K1_CONTEXT_SIGN);
+
+	if (passwd)
+		get_encrypted_hsm_secret(&hsm_secret, hsm_secret_path, passwd);
+	else
+		get_hsm_secret(&hsm_secret, hsm_secret_path);
+
+	for (u64 dbid = 0; dbid < tries ; dbid++) {
```
cdecker: `dbid` starts with 1

cdecker:
Nice tool, my main concern is that we're starting to crowd the `README.md` again, putting niche things in there and distracting from the simple getting started part. We should probably start working on an faq that collects these things instead.

darosior:
> Nice tool, my main concern is that we're starting to crowd the README.md  again, putting niche things in there and distracting from the simple  getting started part. We should probably start working on an faq that  collects these things instead.

Yeah I was not sure about the README addtion.
About the FAQ that's my feeling too, maybe we could add an intro (we would repeat things but I think it's OK) and a FAC to https://lightning.readthedocs.io/, maybe both on the homepage until we have too many questions ?

README.md
darosior: Seemed to be frequently asked ... Hence the FAQ I guess ^^

tools/hsmtool.c
darosior: oops, thanks!

darosior:
Thanks for the review !
Actually the premature `argv[]` access was used in all commands and so dirty that I added a drive-by commit to fix it.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3292/commits/e0988a2b82bcbb650ea295034cae6d1a2beb4a96

cdecker:
I think we should earmark these things for later collection into an faq in the docs :wink:

PR #3295, opened 2019-11-26
'invoice' command returning preimage.

fiatjaf:
I'm sorry for this silly PR, but this has been bugging me for a long time. Being able to see the preimages of invoices issued is good for overall understanding of the protocol and for easy checking of payment proofs from others, not to mention other advanced uses (which may be covered by passing your own preimage to `invoice` as the docs imply, but still).

The point is that this doesn't hurt and makes everything better.

---

### Some useless context

On https://t.me/lntxbot I like to show a lot of debugging information (because I'm crazy and I think users will want to understand what is happening): 2019-11-25-231518_445x806_scrot

However, because I can't get the preimages my invoices are using from the `invoice` command to show these preimages I have to generate them on my own using a probably-not-so-safe generator and store them in not-so-optimal ways instead of having `lightningd` generate them for me.

**edit**: just realized the above image is from the sender side, not the receiver, so completely unrelated.

### DIFF
--- a/doc/lightning-invoice.7
+++ b/doc/lightning-invoice.7
@@ -84,6 +84,7 @@ not\.
 
 The following error codes may occur:
 
+.RS
 .IP \[bu]
 -1: Catchall nonspecific error\.
 .IP \[bu]
@@ -91,9 +92,11 @@ The following error codes may occur:
 .IP \[bu]
 901: An invoice with the given \fIpreimage\fR already exists\.
 
+.RE
 
 One of the following warnings may occur (on success):
 
+.RS
 .IP \[bu]
 \fIwarning_offline\fR if no channel with a currently connected peer has
   the incoming capacity to pay this invoice
@@ -102,6 +105,7 @@ One of the following warnings may occur (on success):
   incoming capacity and has a peer that is publicly connected (i\.e\.
   not a dead end)
 
+.RE
 .SH AUTHOR
 
 Rusty Russell \fI<rusty@rustcorp.com.au\fR> is mainly responsible\.
--- a/doc/lightning-listinvoices.7
+++ b/doc/lightning-listinvoices.7
@@ -12,17 +12,16 @@ if it exists, or the status of all invoices if given no argument\.
 
 .SH RETURN VALUE
 
-On success, an array \fIinvoices\fR of objects is returned\. Each object
-contains \fIlabel\fR, \fIpayment_hash\fR, \fIstatus\fR (one of \fIunpaid\fR, \fIpaid\fR or
-\fIexpired\fR), and \fIexpiry_time\fR (a UNIX timestamp)\. If the \fImsatoshi\fR
-argument to \fBlightning-invoice\fR(7) was not "any", there will be an
-\fImsatoshi\fR field as a number, and \fIamount_msat\fR as the same number
-ending in \fImsat\fR\. If the invoice \fIstatus\fR is \fIpaid\fR, there will be a
-\fIpay_index\fR field and an \fImsatoshi_received\fR field (which may be
-slightly greater than \fImsatoshi\fR as some overpaying is permitted to
-allow clients to obscure payment paths); there will also be an
-\fIamount_received_msat\fR field with the same number as
-\fImsatoshi_received\fR but ending in \fImsat\fR\.
+On success, an array \fIinvoices\fR of objects is returned\. Each object contains
+\fIlabel\fR, \fIpayment_hash\fR, \fIstatus\fR (one of \fIunpaid\fR, \fIpaid\fR or \fIexpired\fR),
+\fIpayment_preimage\fR (for paid invoices), and \fIexpiry_time\fR (a UNIX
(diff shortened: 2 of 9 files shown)


cdecker:
For context: we aren't displaying the `preimage` mainly because it currently serves also as a proof-of-payment ("I have the preimage so I have paid you, how else could I have gotten it?"). If we were to show the `preimage` in the cli output, then the user could be tricked into sharing that with the sender, thus giving him the proof.

Worse, if any node on the path between sender and recipient learns about the preimage before the payment is attempted it can grab the transfer, short-circuiting it using the preimage. This would then be a really hard to debug situation (sender sends payment successfully, but the payment never arrives at the recipient, but sender still has proof that he paid).

A middle ground could be to show the preimage after the payment has completed and the invoice was paid, at which point users could compare the preimage and just see that it matches on both ends.

fiatjaf:
Well, but doesn't this seem like a big user failure from someone who doesn't know what they're doing? A user like that can do a lot of harm to himself already in other ways. Like being tricked into issuing `pay` on an invoice he didn't inspect first and end up paying 1000000 sat instead of 1000 (actually I'm very afraid of doing that in a moment of distraction).

Anyway, ignore the above paragraph. The middle ground you suggested is a big improvement, I think. And it even solves my specific issue now that I'm thinking about it. I'll change this PR to do that.

darosior:
About the middle group: `waitsendpay`, `listpays` and `listsendpays` already contains the preimage. https://lightning.readthedocs.io/lightning-waitsendpay.7.html, https://lightning.readthedocs.io/lightning-listpays.7.html, https://lightning.readthedocs.io/lightning-listsendpays.7.html. There are also a hook and a notification https://lightning.readthedocs.io/PLUGINS.html#invoice-payment

cdecker:
> About the middle ground: `waitsendpay`, `listpays` and `listsendpays` already contains the preimage. https://lightning.readthedocs.io/lightning-waitsendpay.7.html, https://lightning.readthedocs.io/lightning-listpays.7.html, https://lightning.readthedocs.io/lightning-listsendpays.7.html. There are also a hook and a notification https://lightning.readthedocs.io/PLUGINS.html#invoice-payment sent when an invoice was paid and containing the payment preimage

All of those RPC calls however are on the sender side, and I think @fiatjaf is talking about the recipient side.

I'll implement the middle solution ASAP :-)

cdecker:
I went ahead and added a couple of fixup commits that I'll squash before merging, but I wanted to get @fiatjaf's sign-off first:

 - Good work on adding `json_add_preimage` and consolidating it's use
 - I removed the `payment_preimage` in the return of `invoice` again as discussed before
 - I added a test for the preimage when listing paid invoices

The commit message for the first commit doesn't really match anymore, if you'd like I can reword that before squashing ^^

fiatjaf:
I think this is great, thanks!

Please reword the commit while squashing, yes.

Also thank you for the compliment, it means a lot to my insufficient C skills.

rustyrussell:
I share @cdecker's reluctance to share the preimage, but this could be made a parameter to listinvoices (default off).

cdecker:
> I share @cdecker's reluctance to share the preimage, but this could be made a parameter to listinvoices (default off).

I had thought about this, however if we assume the attacker can trick the user into sharing the output of `listinvoices` it'd be easy for him to also trick him to use the extra parameter ("My payment isn't working? That's strange! Can you share the output of `lightning-cli listinvoices [payment_hash] true`?"). By gating the reveal of the preimage on the status of the invoice we at least make sure it was paid first, which is in line with the "don't reuse invoices" practice.

cdecker:
Rebased and squashed, should be good to go, unless @rustyrussell wants to guard the `payment_preimage` for paid invoices using a flag.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3295/commits/f309dc04bab43d67cafaf5e88b8dcb710d0c9666

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3295/commits/d0905dfa5149bb3e3e9468a0e57cf1356ec9f7d0

rustyrussell:
Ack d0905dfa5149bb3e3e9468a0e57cf1356ec9f7d0

rustyrussell:
> Ack [d0905df](https://github.com/ElementsProject/lightning/commit/d0905dfa5149bb3e3e9468a0e57cf1356ec9f7d0)
> 
> [ Changed ACK to Ack to see if ackbot will recognize it? --RR ]

That didn't work, but once I acked it recognized @cdecker's ack too.  Weird.

cdecker:
> > Ack [d0905df](https://github.com/ElementsProject/lightning/commit/d0905dfa5149bb3e3e9468a0e57cf1356ec9f7d0)
> > [ Changed ACK to Ack to see if ackbot will recognize it? --RR ]
> 
> That didn't work, but once I acked it recognized @cdecker's ack too. Weird.

Yeah, the bot had caught it, but calling the status RPC on Github failed, so changes to the ack message were bumping against the uniqueness constraint and not retrying the Github RPC call. Will fix :wink:

PR #3297, opened 2019-11-26
TLV generation cleanup and testing

cdecker:
As promised in #3278 I went back and modified the `run-tlvstream` tests to use the typesafe TLV parser. Those last uses removed we could also remove the non-typesafe variant `fromwire_tlvs`. I'm planning to do the same for the `towire_tlvs` helper and make that typesafe (hence the draft status).

Changelog-None


cdecker:
Ready for review :-)

Ping @niftynei

cdecker:
Rebased on top of `master` and fixed the type error pointed out by @rustyrussell. Re-applying ack.

ACK https://github.com/ElementsProject/lightning/pull/3297/commits/b346c5b57c7c6c158856083502fb196dbae3b0ce

tools/gen/impl_template
```diff
@@ -321,8 +321,10 @@ int ${tlv.name}_is_valid(const struct ${tlv.struct_name()} *record)
 			 *   - otherwise, if `type` is odd:
 			 *     - MUST discard the next `length` bytes.
 			 */
-			SUPERVERBOSE("Unknown even type in TLV");
-			return i;
+			SUPERVERBOSE("unknown even");
+			if (err_index != NULL)
```
niftynei: is this better than `if (err_index)` ?

cdecker: I slightly prefer it since it makes it clearer that we aren't checking a boolean but rather a pointer. I'm pretty sure it results in the same assembly.

cdecker: Indeed: https://godbolt.org/z/tcXEAc

niftynei: personally, i prefer the more concise `if (err_index)`; the fact that `err_index`'s type is a pointer should be clue enough that we're checking for pointer presence

PR #3299, opened 2019-11-28
doc: adds pyln contrib dependencies to the command for running blackbox tests

m-schmoock:
This one got missing when introducing `pyln` modules.

Changelog-None


m-schmoock:
I also removed the `.` dots after the monospaced commands, so they can be copy and pasted more easily, and it also looks better...

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3299/commits/a98b47cb77b8c3659cb14f5d68a090c75f23c553

PR #3300, opened 2019-11-28
Add network to init message

darosior:
This implements https://github.com/lightningnetwork/lightning-rfc/pull/682


connectd/connectd.c
```diff
@@ -151,6 +151,9 @@ struct daemon {
 
 	/* Allow to define the default behavior of tot services calls*/
 	bool use_v3_autotor;
+
+	/* What network are we running on ? */
+	const struct chainparams *chainparams;
```
cdecker: In other daemons we have a global instead, which might cause some discrepancies if we reuse code meant for other daemons in `connectd`. We should probably make this a global as well just in case.

connectd/peer_exchange_initmsg.c
```diff
@@ -51,13 +65,35 @@ static struct io_plan *peer_init_received(struct io_conn *conn,
...
+	 *  - upon receiving `networks` containing no common chains
+	 *    - MAY fail the connection.
+	 */
+	if (tlvs->networks) {
+		if (!tlvs->networks->chains) {
+			status_peer_debug(&peer->id,
+			                  "bad networks TLV in init '%s', closing",
+			                  tal_hex(tmpctx, msg));
+			return io_close(conn);
+		}
```
cdecker: This would be a parsing error, wouldn't it? It'd fail the `fromwire_init` return value test above, right?

devtools/gossipwith.c
```diff
@@ -137,11 +156,21 @@ static struct io_plan *handshake_success(struct io_conn *conn,
...
-		msg = towire_init(NULL, NULL, features, NULL);
+		struct tlv_init_tlvs *tlvs = NULL;
+		if (blockchainparams) {
+			tlvs = tlv_init_tlvs_new(NULL);
+			tlvs->networks = tal(tlvs, struct tlv_init_tlvs_networks);
+			tlvs->networks->chains = tal_arr(tlvs->networks, struct bitcoin_blkid, 1);
+			tlvs->networks->chains[0] = blockchainparams->genesis_blockhash;
+			msg = towire_init(NULL, NULL, features, tlvs);
+		} else
+			msg = towire_init(NULL, NULL, features, NULL);
```
cdecker: These lines can be pulled out of the `if`-statement and combined.

devtools/gossipwith.c
```diff
@@ -137,11 +156,21 @@ static struct io_plan *handshake_success(struct io_conn *conn,
...
 
 		sync_crypto_write(pps, take(msg));
 		/* Ignore their init message. */
 		tal_free(sync_crypto_read(NULL, pps));
+		if (tlvs)
+			tal_free(tlvs);
```
cdecker: Might be an option to make `towire_init` `take()` the `tlvs` and therefore hand off ownership.

devtools/gossipwith.c
```diff
@@ -254,6 +283,10 @@ int main(int argc, char *argv[])
 			   "Print out messages in hex");
 	opt_register_arg("--features=<hex>", opt_set_features, NULL,
 			 &features, "Send these features in init");
+	opt_register_arg("--network", opt_set_network, opt_show_network,
+	                 NULL,
+	                 "Select the network parameters (bitcoin, testnet,"
+	                 " regtest, litecoin or litecoin-testnet)");
```
cdecker: `testnet`, `liquid` and `liquid-regtest` are also valid networks.

connectd/connectd.c
darosior: I also tried using the constant in `common/utils.h`, hmm reading other daemons maybe I could just populate this one in `fromwire_connectctl_init` insead.. Will try

connectd/peer_exchange_initmsg.c
darosior: It would be set to `NULL` by `fromwire_tlv_init_tlvs_network`. I'm not sure but I don't think it would be a parsing error.

devtools/gossipwith.c
darosior: How could we `blockchainparams->genesis_blockhash;` if `blockchainparams` isn't defined ?

devtools/gossipwith.c
darosior: Yes, but should be an other PR I think ?

darosior:
Rebased to correct the use of `chainparams` (just initializing it at `connectd` init was enough) and add minor fixups to `gossipwith` (and add a once-again-forgotten `Changelog-Added`!).

devtools/gossipwith.c
cdecker: No I mean we can make this:

```c
			msg = towire_init(NULL, NULL, features, tlvs);
		} else
			msg = towire_init(NULL, NULL, features, NULL);
```
into this
```c
		}
		msg = towire_init(NULL, NULL, features, tlvs);
```

Since `tlvs` is initialized to `NULL` anyway.

darosior: Ah, yeah absolutely, I..

devtools/gossipwith.c
rustyrussell: And a big one!  There are a few places we may hand pointers *assuming* generated routines don't take()...

devtools/gossipwith.c
```diff
@@ -137,11 +155,20 @@ static struct io_plan *handshake_success(struct io_conn *conn,
...
+			tlvs->networks->chains = tal_arr(tlvs->networks, struct bitcoin_blkid, 1);
+			tlvs->networks->chains[0] = chainparams->genesis_blockhash;
+		}
+			msg = towire_init(NULL, NULL, features, tlvs);
 
 		sync_crypto_write(pps, take(msg));
 		/* Ignore their init message. */
 		tal_free(sync_crypto_read(NULL, pps));
+		if (tlvs)
+			tal_free(tlvs);
```
rustyrussell: Don't need to check for NULL here, BTW.

darosior: Removed, thanks

cdecker:
Re-applying @rustyrussell's ACK, and merging:

ACK https://github.com/ElementsProject/lightning/pull/3300/commits/bcc7e708f108ee8b6d16d685233fd934d3520aa7

connectd/peer_exchange_initmsg.c
```diff
@@ -51,13 +61,34 @@ static struct io_plan *peer_init_received(struct io_conn *conn,
...
+	 * ...
+	 *  - upon receiving `networks` containing no common chains
+	 *    - MAY fail the connection.
+	 */
+	if (tlvs->networks) {
+		if (!tlvs->networks->chains) {
+			status_peer_debug(&peer->id,
+			                  "bad networks TLV in init '%s', closing",
+			                  tal_hex(tmpctx, msg));
+			return io_close(conn);
```
rustyrussell: Note this is technically redundant, since it's covered elegently by the next part anyway.

PR #3307, opened 2019-12-01
plugins/fundchannel: try to connect to peer before funding

darosior:
Just ya callback before `fundchannel_start` :-)

### DIFF
--- a/plugins/fundchannel.c
+++ b/plugins/fundchannel.c
@@ -306,7 +306,9 @@ static struct command_result *fundchannel_start_done(struct command *cmd,
 }
 
 static struct command_result *fundchannel_start(struct command *cmd,
-						struct funding_req *fr)
+                                                const char *buf UNUSED,
+                                                const jsmntok_t *result UNUSED,
+                                                struct funding_req *fr)
 {
 	struct json_out *ret = json_out_new(NULL);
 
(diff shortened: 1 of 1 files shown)


plugins/fundchannel.c
```diff
@@ -326,14 +328,25 @@ static struct command_result *fundchannel_start(struct command *cmd,
...
+                                              struct funding_req *fr)
+{
+	struct json_out *ret = json_out_new(NULL);
+
+	json_out_start(ret, NULL, '{');
+	json_out_addstr(ret, "id", node_id_to_hexstr(tmpctx, fr->id));
+	json_out_end(ret, '}');
+	json_out_finished(ret);
+
+	return send_outreq(cmd, "connect", fundchannel_start, tx_abort,
```
niftynei: there's no reason to call `tx_abort` here in the failure case, we can just pipe the failure from connect back to the client.

plugins/fundchannel.c
```diff
@@ -326,14 +328,25 @@ static struct command_result *fundchannel_start(struct command *cmd,
...
 			   fr, take(ret));
 }
 
+static struct command_result *connect_to_peer(struct command *cmd,
+                                              struct funding_req *fr)
+{
+	struct json_out *ret = json_out_new(NULL);
+
+	json_out_start(ret, NULL, '{');
+	json_out_addstr(ret, "id", node_id_to_hexstr(tmpctx, fr->id));
```
niftynei: does this work without the IP address of the client?

darosior: Yes, it can be seen as a sanity check if the peer is connected. If it is not, this will at least try using the seed.

plugins/fundchannel.c
darosior: That's what `tx_abort` will do, it's useless but does not seem to hurt ? This avoids another failure callback which does essentially the same

niftynei: that's not true, `tx_abort` makes an extra rpc call to lightningd; `forward_error` (plugins/libplugin.c#L283) is what you want

plugins/fundchannel.c
niftynei: yeah i think the net result of this is that it will mostly paper over how openingd disconnects when opens fail

niftynei: we should *really* fix `openingd` to not disconnect on certain classes of failure, as that cuts off the gossip with that peer. that's a much bigger fix though

plugins/fundchannel.c
darosior: Thanks

darosior:
Nevermind I should have spent more time reading the code around the fixme :grimacing:

PR #3308, opened 2019-12-02
lightningd: fix reference to out-of-scope var.

rustyrussell:
cppcheck found this:

	[lightningd/options.c:1137] -> [lightningd/options.c:1120] -> [lightningd/options.c:1193]: (error) Using pointer to local variable 'buf' that is out of scope.

Indeed, answer can point into buf, which is no longer in scope at the end.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3308/commits/b0985b4b98edefc011c9bc4293766715fe6bf5fe

PR #3309, opened 2019-12-02
EXPERIMENTAL: MPP send and receive support (lowlevel)

rustyrussell:
Not added to the pay plugin, but it's in sendpay etc.

There's no documentation changes since this only adds optional params, which are only useful with EXPERIMENTAL_FEATURES=1.


wallet/db.c
```diff
@@ -481,6 +481,84 @@ static struct migration dbmigrations[] = {
     {SQL("UPDATE forwarded_payments SET received_time=0 WHERE received_time IS NULL;"),
 	 NULL},
     {SQL("ALTER TABLE invoices ADD COLUMN features BLOB DEFAULT '';"), NULL},
+   /* We can now have multiple payments in progress for a single hash, so
+    * add two fields; combination of payment_hash & parallel_id is unique. */
+    {SQL("ALTER TABLE payments RENAME TO temp_payments;"), NULL},
+    {SQL("CREATE TABLE payments ("
```
cdecker: Should we maybe go the extra mile and call this `payment_parts` instead? We could then pull the common fields into a separate `payments` table. Would be a nice cleanups and fix that naming nit about pays being the combination of multiple payments.

wallet/wallet.h
```diff
@@ -249,13 +249,18 @@ struct wallet_payment {
 	struct list_node list;
 	u64 id;
 	u32 timestamp;
+
+	/* The combination of these two fields is unique: */
 	struct sha256 payment_hash;
+	u64 parallel_id;
```
cdecker: So the MPP acronym is for "multi-parallel payment", and not like I thought "multi-part payment"? :wink:

wallet/wallet.c
```diff
@@ -1770,8 +1769,11 @@ static bool wallet_stmt2htlc_in(struct channel *channel,
 	return ok;
 }
 
-static bool wallet_stmt2htlc_out(struct channel *channel,
-				struct db_stmt *stmt, struct htlc_out *out)
+/* Removes matching htlc from htlcs_in */
```
cdecker: Nit: this probably means `unconnected_htlcs_in`.

cdecker: I'm a bit skeptical about these conversion functions having side-effects.

lightningd/pay.c
```diff
@@ -1013,74 +1042,12 @@ static struct command_result *json_sendonion(struct command *cmd,
...
-	else
-		payment->label = NULL;
-
-	/* We write this into db when HTLC is actually sent. */
-	wallet_payment_setup(ld->wallet, payment);
-
-	add_sendpay_waiter(ld, cmd, payment_hash);
-	return command_still_pending(cmd);
+	return send_payment_core(ld, cmd, payment_hash, /* FIXME: Set parallel_id! */0,
+				 first_hop, AMOUNT_MSAT(0), AMOUNT_MSAT(0),
```
cdecker: Pretty sure setting `msat_total = AMOUNT_MSAT(0)` on all onion payments will cause issues further down the line. We might need to have an exemption in the check in the next commit to allow payments with different onions to exceed the `msatoshi_total` amount.

tests/test_pay.py
```diff
@@ -2564,11 +2564,11 @@ def serialize_payload(n):
         assert(e.error['data']['raw_message'] == "400f00000000000003e80000006c")
 
 
-@unittest.skipIf(not DEVELOPER, "needs DEVELOPER=1")
+@unittest.skipIf(not DEVELOPER, "needs dev-disconnect, dev-no-htlc-timeout")
 @unittest.skipIf(not EXPERIMENTAL_FEATURES, "needs parallel_id support")
 def test_partial_payment(node_factory, bitcoind, executor):
     # We want to test two payments at the same time, before we send commit
-    l1, l2, l3, l4 = node_factory.get_nodes(4, [{}] + [{'disconnect': ['=WIRE_UPDATE_ADD_HTLC-nocommit']}] * 2 + [{}])
+    l1, l2, l3, l4 = node_factory.get_nodes(4, [{}] + [{'disconnect': ['=WIRE_UPDATE_ADD_HTLC-nocommit'], 'dev-no-htlc-timeout': None}] * 2 + [{}])
```
cdecker: The holding of HTLCs is necessary so the `waitsendpay` isn't just resolved internally below if the response comes back in time?

tests/test_pay.py
```diff
@@ -2630,3 +2630,15 @@ def test_partial_payment(node_factory, bitcoind, executor):
...
+    assert "'type': 'tlv'" in line
+    assert "'forward_amount': '500msat'" in line
+    assert "'total_msat': '1000msat'" in line
+    assert "'payment_secret': '{}'".format(paysecret) in line
+
+    line = l4.daemon.wait_for_log('print_htlc_onion.py: Got onion')
+    assert "'type': 'tlv'" in line
+    assert "'forward_amount': '500msat'" in line
+    assert "'total_msat': '1000msat'" in line
+    assert "'payment_secret': '{}'".format(paysecret) in line
```
cdecker: Maybe use a loop to emphasize that these are identical and the order is not important.

lightningd/htlc_set.c
```diff
@@ -0,0 +1,193 @@
...
+	}
+	abort();
+#endif
+}
+
+void fulfill_htlc_set(struct htlc_set *set, const struct preimage *preimage)
+{
+	for (size_t i = 0; i < tal_count(set->htlcs); i++) {
+		/* Don't remove from set */
+		tal_del_destructor2(set->htlcs[i], htlc_set_hin_destroyed, set);
```
cdecker: This is done independently of whether it was set as a destructor, hence the function sig above being outside of the compile guards, correct? Just checked, `tal_del_destructor2` just returns `false` if the function was never registered. Why not make this call and the function definition above experimental only?

lightningd/htlc_set.c
```diff
@@ -0,0 +1,193 @@
...
+
+static struct htlc_set *new_htlc_set(struct lightningd *ld,
+				     struct htlc_in *hin,
+				     struct amount_msat total_msat)
+{
+	struct htlc_set *set;
+
+	set = tal(ld, struct htlc_set);
+	set->total_msat = set->so_far = total_msat;
+	set->payment_hash = hin->payment_hash;
```
cdecker: This is only true for the basic AMP approach though. If we get hop decorrelation in the spec we'd end up removing this again. Maybe we should not even start relying on it.

lightningd/htlc_set.c
```diff
@@ -0,0 +1,193 @@
...
+	 */
+	notleak(new_reltimer(ld->timers, set, time_from_sec(70),
+			     timeout_htlc_set, set));
+	htlc_set_map_add(&ld->htlc_sets, set);
+	tal_add_destructor2(set, destroy_htlc_set, &ld->htlc_sets);
+#endif
+	return set;
+}
+
+void htlc_add_set(struct lightningd *ld,
```
cdecker: Nit: `htlc_add_set` -> `htlc_set_add`

lightningd/htlc_set.c
```diff
@@ -0,0 +1,193 @@
...
+		  struct htlc_in *hin,
+		  struct amount_msat total_msat,
```
cdecker: Nit: `total_msat` -> `msatoshi_total` to match other occurrences.

lightningd/htlc_set.c
```diff
@@ -0,0 +1,193 @@
...
+#endif /* EXPERIMENTAL_FEATURES */
+
+static struct htlc_set *new_htlc_set(struct lightningd *ld,
+				     struct htlc_in *hin,
+				     struct amount_msat total_msat)
+{
+	struct htlc_set *set;
+
+	set = tal(ld, struct htlc_set);
+	set->total_msat = set->so_far = total_msat;
```
cdecker: Why is `set->so_far` being set here, when it is also being set to `AMOUNT_MSAT(0)` two lines below?

lightningd/htlc_set.c
```diff
@@ -181,8 +181,7 @@ void htlc_add_set(struct lightningd *ld,
 	}
 
 	if (amount_msat_eq(set->so_far, total_msat)) {
-		/* FIXME: hand to invoice_try_pay! */
-		tal_free(set);
+		invoice_try_pay(ld, set, details);
```
cdecker: Should this be `take(set)` for self-documentation?

tests/test_pay.py
```diff
@@ -2620,16 +2620,13 @@ def test_partial_payment(node_factory, bitcoind, executor):
...
 
-    # See FIXME in peer_htlcs: WIRE_INVALID_ONION_PAYLOAD would be better.
-    with pytest.raises(RpcError, match=r'WIRE_FINAL_INCORRECT_HTLC_AMOUNT'):
-        l1.rpc.call('waitsendpay', [inv['payment_hash'], None, 1])
-    with pytest.raises(RpcError, match=r'WIRE_FINAL_INCORRECT_HTLC_AMOUNT'):
-        l1.rpc.call('waitsendpay', {'payment_hash': inv['payment_hash'],
-                                    'parallel_id': 2})
+    l1.rpc.call('waitsendpay', [inv['payment_hash'], None, 1])
+    l1.rpc.call('waitsendpay', {'payment_hash': inv['payment_hash'],
+                                'parallel_id': 2})
```
cdecker: These need checks that we actually get the parts in the correct call, and that there isn't a place we cross wires :-)

tests/test_pay.py
```diff
@@ -2645,3 +2645,76 @@ def test_partial_payment(node_factory, bitcoind, executor):
...
+
+@unittest.skipIf(not EXPERIMENTAL_FEATURES, "needs parallel_id support")
+def test_partial_payment_timeout(node_factory, bitcoind):
+    l1, l2 = node_factory.line_graph(2)
+
+    inv = l2.rpc.invoice(1000, 'inv', 'inv')
+    paysecret = l2.rpc.decodepay(inv['bolt11'])['payment_secret']
+
+    route = l1.rpc.getroute(l2.info['id'], 500, 1)['route']
+    l1.rpc.call('sendpay', [route, inv['payment_hash'], None, 1000, inv['bolt11'], paysecret, 1])
```
cdecker: The `None` to skip an unassigned argument is hard to read. Could use the `kwargs` dict instead which is also nice self-documenting.

tests/test_pay.py
```diff
@@ -2645,3 +2645,76 @@ def test_partial_payment(node_factory, bitcoind, executor):
...
+    l1.rpc.call('sendpay', [route, inv['payment_hash'], None, 1000, inv['bolt11'], paysecret, 2])
+
+    l1.rpc.call('waitsendpay', [inv['payment_hash'], TIMEOUT, 1])
+    l1.rpc.call('waitsendpay', [inv['payment_hash'], TIMEOUT, 2])
+
+
+@unittest.skipIf(not DEVELOPER, "needs dev-fail")
+@unittest.skipIf(not EXPERIMENTAL_FEATURES, "needs parallel_id support")
+def test_partial_payment_htlc_loss(node_factory, bitcoind):
+    """Test that we discard a set when the HTLC is lost"""
```
cdecker: Comment seems a copy-pasta

cdecker:
Since this is touching the DB in some places I'm running the tests against postgres now.

cdecker:
I assume the changes to `pyln-client` are deferred until the feature is non-experimental, corret?

cdecker:
One of the newly added fields seems to have the wrong type:

```
u64 field doesn't match size: expected 8, actual 4
```

My guess would be this:

```c
{SQL("ALTER TABLE channel_htlcs ADD parallel_id INTEGER;"), NULL},
```
Since it declares the field as a 4 byte INTEGER, while trying to read it as a 8 byte BIGINT.

cdecker:
> One of the newly added fields seems to have the wrong type:
> 
> ```
> u64 field doesn't match size: expected 8, actual 4
> ```
> 
> My guess would be this:
> 
> ```c
> {SQL("ALTER TABLE channel_htlcs ADD parallel_id INTEGER;"), NULL},
> ```
> 
> Since it declares the field as a 4 byte INTEGER, while trying to read it as a 8 byte BIGINT.

Just verified that that fixes the issue. Now all I'm left with is some spurious `MPP_TIMEOUT`s that seem a bit odd.

tests/test_pay.py
rustyrussell: Yes, and because this test was implemented before mpp, so I wanted to test the *send* side allowed two payments in parallel.

lightningd/htlc_set.c
rustyrussell: Oops, good catch.  I originally set it to the first payment, then realized the code was neater to start at 0.

tests/test_pay.py
rustyrussell: Well, it'll go away once sendpay arg is no longer experimental anyway, so hard to really mind.

wallet/db.c
rustyrussell: I'm not sure that should be done at the db level; it may be premature optimization?

wallet/wallet.h
rustyrussell: Yeah, naming predated the MPP name.  I like 'partid'.

wallet/wallet.c
rustyrussell: I agree.  It's something of a necessary evil though; the previous load-and-join split was pretty nasty, involving a temporary field which was bad too.

lightningd/pay.c
```diff
@@ -1090,13 +1090,15 @@ static struct command_result *json_sendonion(struct command *cmd,
...
 	struct secret *path_secrets;
+	u64 *parallel_id;
 
 	if (!param(cmd, buffer, params,
 		   p_req("onion", param_bin_from_hex, &onion),
 		   p_req("first_hop", param_route_hop, &first_hop),
 		   p_req("payment_hash", param_sha256, &payment_hash),
 		   p_opt("label", param_escaped_string, &label),
 		   p_opt("shared_secrets", param_secrets_array, &path_secrets),
+		   p_opt_def("parallel_id", param_u64, &parallel_id, 0),
```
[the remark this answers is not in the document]
rustyrussell: Generally prefer to avoid `_` in names unless they're long, so `partid` is what I went for in the end.

lightningd/htlc_set.c
rustyrussell: Yes, that's probably even clearer.

lightningd/htlc_set.c
rustyrussell: Hmm, good thought.

But we'd do this by switching the keying from payment_hash to secret: that's going to be easier once secret is a requirement, which it should be before then?

lightningd/htlc_set.c
rustyrussell: Also new, fail, fulfull...

lightningd/htlc_set.c
rustyrussell: total_msat is from spec, rest should be changed instead?

lightningd/htlc_set.c
rustyrussell: Hmm, no, it *has* to dispose of it.  take() is for things where it may or may not take ownership.

We've generally used naming to try to indicate which functions are a sink.  Should we rename this to 'invoice_resolve_htlc_in'?

wallet/db.c
cdecker: Yes, you're right, that'd be a bigger refactor that we should probably be done during a bit of downtime. It'd just be nice to have an actual `payment` concept that collects all the `attempt`s that are performed trying to complete it. It may be my OOP/ORM past that is catching up with me :wink:

wallet/wallet.c
cdecker: Right, well at least we have documentation for it :-)

lightningd/htlc_set.c
cdecker: You're right, the `htlc_set` being affected by any decorrelation technique is only relevant at the endpoints which do have access to the secret (I was confused for a moment since intermediate nodes don't have access to that, but then again they'd treat parts as separate payments).

lightningd/htlc_set.c
cdecker: That sounds like a good idea. I keep getting confused by `take()` :-)

I'm tempted to propose an `OWNS` annotation and an `own()` nop-helper that just serves as annotation to say: I'll take ownership no matter what and you MUST take ownership. Purely documentation but we'd know which side of the call is wrong if we have a leak without having to dig through both the caller and callee to see what the semantics should have been. Just fantasizing here of course :smiley:

PR #3310, opened 2019-12-03
Dont crash on bad sphinx

rustyrussell:
Found the missing validity check by inspection :(

Clearly, we need to call the validity *by default*, and have a special "don't check validity" version for the one case where that is wanted.

### DIFF
--- a/common/sphinx.c
+++ b/common/sphinx.c
@@ -91,6 +91,9 @@ static size_t sphinx_hop_size(const struct sphinx_hop *hop)
 	if (hop->type == SPHINX_V0_PAYLOAD)
 		return 65;
 
+	if (hop->type == SPHINX_RAW_PAYLOAD)
+		return size + HMAC_SIZE;
+
 	/* Since this uses the bigsize serialization format for variable
 	 * length integer encodings we need to allocate enough space for
 	 * it. Values >= 0xfd are used to signal multi-byte serializations. */
@@ -531,11 +534,17 @@ static bool sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
 	size_t padding_size;
 	int pos = 0;
 
-	/* Backwards compatibility for the legacy hop_data format. */
-	if (hop->type == SPHINX_V0_PAYLOAD)
+	switch (hop->type) {
+	case SPHINX_V0_PAYLOAD:
+		/* Backwards compatibility for the legacy hop_data format. */
 		dest[pos++] = 0x00;
-	else
+		break;
+	case SPHINX_TLV_PAYLOAD:
 		pos += bigsize_put(dest+pos, raw_size);
+		break;
+	case SPHINX_RAW_PAYLOAD:
+		break;
+	}
 
 	memcpy(dest + pos, hop->payload, raw_size);
 	pos += raw_size;
--- a/common/sphinx.c
+++ b/common/sphinx.c
@@ -603,7 +603,11 @@ static bool sphinx_parse_payload(struct route_step *step, const u8 *src)
 		fromwire_varint(&tlv, &max);
 
 		if (!fromwire_tlv_payload(&tlv, &max, step->payload.tlv)) {
-			/* FIXME: record offset of violation for error! */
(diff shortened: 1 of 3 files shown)


cdecker:
I generally agree with the PR, however adding the call to `tlv_payload_is_valid` in https://github.com/ElementsProject/lightning/pull/3310/commits/a72200177e7c4aa11b071adc182787e0af3c8cc7 destroys the whole point of the previous split. The split is necessary since we want to allow TLV payloads that contain TLV types that we ourselves do not understand, but our plugins may be able to handle. By calling `tlv_payload_is_valid` we prevent that case.

I'll create a PR that moves the well-formedness test into the `fromwire_tlv_payload` function, and `tlv_payload_is_valid` will be reduced to just check whether we know how to handle all the even types in the TLV (this should only get called after the hook returns, never before). The semantic check and the check that in case of `continue` we have all the necessary information is also located after the hook return, which is where we can make a final decision on what to do with it.

Removing the `INVALID_ONION_PAYLOAD` in favor of a deserialization error is a good thing, and we should have done that earlier.

cdecker:
My counter-proposal can be found here https://github.com/ElementsProject/lightning/compare/master...cdecker:pr-3310-take2

It removes the call to `tlv_payload_is_valid` in https://github.com/ElementsProject/lightning/commit/a72200177e7c4aa11b071adc182787e0af3c8cc7, moving the well-formedness check into the `fromwire_tlv_payload` function, deferring the semantic validation in `tlv_payload_is_valid` until after the hook returns.

IMHO the actual bug was already fully addressed in your first commit https://github.com/ElementsProject/lightning/pull/3310/commits/6890b692fd43139742e872a5f296bbac2b987ff9 removing the invalid state ever leaving `sphinx_parse_payload`.

rustyrussell:
> IMHO the actual bug was already fully addressed in your first commit [6890b69](https://github.com/ElementsProject/lightning/commit/6890b692fd43139742e872a5f296bbac2b987ff9) removing the invalid state ever leaving `sphinx_parse_payload`.

*That* bug was.

But since we never called tlv_is_valid(), we were not correctly erroring on TLVs with unknown even fields, which is a distinct bug.

Even then, the whole idea of this code is wrong.  Even now, what you want to send something in the onion which is not a valid TLV at all, and receive it in the hook?

I think the real fix here is not to have the sphinx code do the parsing *at all*.  Do it just before the accepted_hook (so we can offer those fields if we can decode, otherwise we only offer raw).  I'll prepare a patch.

cdecker:
> > IMHO the actual bug was already fully addressed in your first commit [6890b69](https://github.com/ElementsProject/lightning/commit/6890b692fd43139742e872a5f296bbac2b987ff9) removing the invalid state ever leaving `sphinx_parse_payload`.
> 
> _That_ bug was.
> 
> But since we never called tlv_is_valid(), we were not correctly erroring on TLVs with unknown even fields, which is a distinct bug.

But we _are_ calling is_valid where it is appropriate, namely after the hook was able to decide what to do with it:

https://github.com/ElementsProject/lightning/blob/db92c2ac5ec5cf46bad6023f7560a3e06c7a9e39/lightningd/peer_htlcs.c#L794-L800

This is why I split the parsing from the validity in the first place, let unknown even fields through until we get told to continue at which point it's up to us to enforce our internally implemented logic :-)

> Even then, the whole idea of this code is wrong. Even now, what you want to send something in the onion which is not a valid TLV at all, and receive it in the hook?

Good point, that's why I propose we move the well-formedness checks for TLV back into the parsing. That was indeed an error on my side.

> I think the real fix here is not to have the sphinx code do the parsing at all. Do it just before the accepted_hook (so we can offer those fields if we can decode, otherwise we only offer raw). I'll prepare a patch.

Sounds like a good idea. As it is right now the parsing in `route_step_decode` is little more than a helper putting some commonly accessed pieces of information in an easier to access top-level pointer. It should never `abort()`. We could also remove that entirely, but it still makes access easier in the plugin hook serialization and the `continue` checks after the hook returns.

PR #3314, opened 2019-12-04
fundchannel connect

niftynei:
supercedes #3307 

still unclear to me what cases you can reconnect to a peer without their IP address, however.


darosior:
> still unclear to me what cases you can reconnect to a peer without their IP address, however.

@cdecker's seed ;-)

PR #3315, opened 2019-12-04
Add a `sendcustommsg` RPC call and a `custommsg` plugin hook for experimental protocol extensions

cdecker:
This pull request implements the `custommsg` facility that allows users to
implement custom protocol extensions on top of what c-lightning offers. The
implementation aims to maximize the flexibility given to the users while at
the same time minimizing the dangers of sending custom messages (see the
limitations section).

## Limitations
 - Custom messages are not allowed to have the same type as an internally
   handled message type. This is necessary to ensure that the custom messages
   injected at one end do not throw off the internal state handling code at
   the other end. This means that it's ok to implement things that are in the
   spec (e.g., experimental extensions that are not yet fully specified), but
   once c-lightning implements them internally they may no longer be used.
 - Custom message types are limited to odd-numbered types. This is necessary
   since even-typed messages are required to drop the connection if they are
   unknown. If we were to defer that decision to a plugin we would need to
   lock the owning daemon, and `lightningd`, until the plugin has returned the
   verdict on whether this is known or unknown. Odd-typed messages on the
   other hand can be dropped silently meaning we can easily continue
   processing while the plugin deliberates on whether or not it knows how to
   handle the message.
 - Custom messages can only be sent if a peer is owned by `openingd`, i.e.,
   before opening a channel and currently just gossiping, or by `channeld`,
   i.e., we currently have a channel open and are not closing it. The only
   other daemon that talks to a peer directly is `closingd` which doesn't
   handle unexpected messages well (hence it also doesn't handle gossip),
   however in most cases that state is very transitory, so it is unlikely to
   be necessary that we need to send custommsgs while owned by `closingd`.

## Examples

The limitations mentioned above aside, this PR opens a wide variety of
possible extensions and opportunity for experimental implementations of
protocols. This is just a minimal list of things that came to my mind while
working on this:

 - @renepickhardt's idea of sharing a bit more information about capacities in
   the friend-of-a-friend network in order to collaborate on finding
   rebalancings.
 - Add a new broadcast for on-chain data, e.g., also transport block headers
   on top of the encryptedoverlay network to provide hints to peers that there
   is a new block or that they might be lagging behind.
 - Add a new broadcast for liquidity announcements or exchange rate offers for
   cross-chain atomic swaps.
 - More advanced gossip sync mechanisms such as set-reconciliation. These
   might negotiate the set differences in `custommsg`s and then send
   non-`custommsg`s for the actual exchange of information.
 - Many more that others will come up with...

## Testing

The `test_custommsg` test aims to cover all the limitations and also test the
happy path, in either of the two supported subdaemons. Currently it does not
test when a peer is owned by `closingd`, however that is identical to not
being owned at all.

## Open Questions

 - [x] We currently don't do anything with the result from the hook, we could
       as well make it a notification. However we might later implement the
       hook-chaining and we'd likely want to exit the chain as soon as a
       plugin has handled the message instead of continuing to send it to the
       other plugins. There the result could tell us whether it is handled or
       not.
 - [x] Should we make this experimental only for now?
 - [x] Should we make this generally available (non-developer)?

Changelog-None


lightningd/peer_control.c
```diff
@@ -2244,5 +2245,132 @@ void peer_dev_memleak(struct command *cmd)
...
+
+	return command_success(cmd, response);
+}
+
+static const struct json_command sendcustommsg_command = {
+    "dev-sendcustommsg",
+    "utility",
+    json_sendcustommsg,
+    "Send a custom message to the peer with the given {node_id}",
+    .verbose = "sendcustommsg node_id hexcustommsg",
```
darosior: nit: `dev-sendcustommsg node_id hexcustommsg`

tests/test_misc.py
```diff
@@ -2009,3 +2009,70 @@ def test_unicode_rpc(node_factory):
...
+    msg = r'ff' * 32
+    serialized = r'04070020' + msg
+
+    # This address doesn't exist so we should get an error when we try sending
+    # a message to it.
+    node_id = '02df5ffe895c778e10f7742a6c5b8a0cefbe9465df58b92fadeb883752c8107c8f'
+    with pytest.raises(RpcError, match=r'No such peer'):
+        l1.rpc.dev_sendcustommsg(node_id, msg)
+
+    # `l4` is disconnected and we can't send messages to it
```
darosior:
```suggestion
    # `l3` is disconnected and we can't send messages to it
```

cdecker:
> Looks good, but answering the open question I'd argue for `sendcustommsg` to be non-dev as it does not seem to be a shootgun for a regular user as could be other `dev-only` commands (such as `dev-forget-channel`). It also allows a plugin using this command to not require a compilation with `enable-developer`. But removing some restrictions could be developer-only  ..
> 
> > Custom message types are limited to odd-numbered types.

Good point, however I get the feeling we might end up confusing people, and they might think this is to send some sort of chat message to an arbitrary node. It's more of a "don't show a normal mortal what they aren't supposed to use" kind of deal :-)

ZmnSCPxj:
> Good point, however I get the feeling we might end up confusing people, and they might think this is to send some sort of chat message to an arbitrary node. It's more of a "don't show a normal mortal what they aren't supposed to use" kind of deal :-)

Maybe rename to `sendcustomprotocolmsg` and `customprotocolmsg` to deter such?

cdecker:
> > Good point, however I get the feeling we might end up confusing people, and they might think this is to send some sort of chat message to an arbitrary node. It's more of a "don't show a normal mortal what they aren't supposed to use" kind of deal :-)
> 
> Maybe rename to `sendcustomprotocolmsg` and `customprotocolmsg` to deter such?

Let's keep it a developer-only option for now, and see if my fears of causing confusion is justified. If it is we can always rename while making it non-dev-only :-)

wire/Makefile
```diff
@@ -85,6 +85,13 @@ wire/gen_onion_wire.h: wire/gen_onion_wire_csv $(WIRE_BOLT_DEPS) wire/Makefile
 wire/gen_onion_wire.c: wire/gen_onion_wire_csv $(WIRE_BOLT_DEPS) wire/Makefile
 	$(BOLT_GEN) -s --expose-tlv-type=tlv_payload --page impl ${@:.c=.h} onion_type < $< > $@
 
+# Some messages that are common among all daemons
+wire/gen_common_wire.h: wire/common_wire_csv $(WIRE_BOLT_DEPS) wire/Makefile
+	$(BOLT_GEN) -s --page header $@ common_wire_type < $< > $@
+
+wire/gen_common_wire.c: wire/common_wire_csv $(WIRE_BOLT_DEPS) wire/Makefile
+	$(BOLT_GEN) -s --page impl ${@:.c=.h} common_wire_type < $< > $@
```
rustyrussell: TODO: enhance devtools/decodemsg to decode internal messages.  That is not only useful, if it's constructed as a giant switch() statement, it also ensures we don't have any duplicate message numbers anywhere (otherwise it won't compile).

Meanwhile, we should probably have such a switch statement somewhere in the source (even if it's just manually maintained) to force this check?

tools/gen/impl_template
```diff
@@ -31,6 +31,18 @@ const char *${enum_set['name']}_name(int e)
...
+{
+	switch ((enum ${enum_set['name']})type) {
+    % for msg in enum_set['set']:
+	case ${msg.enum_name()}:;
+    % endfor
+	      return true;
+	}
+	return false;
+}
+
```
rustyrussell: can we use thse functions instead of the hacky "name contains INVALID" checks now?

cdecker:
> Am in two minds about dev-. Should we drop it?

Let's keep it for now and see how crazy people go with `custommsg`, and we can promote it in a later release.

cdecker:
Had to rebase on top of `master` and I'll address the nits in a cleanup PR. Merging to keep the conflict surface low :-)

ACK https://github.com/ElementsProject/lightning/pull/3315/commits/09c72dddf9f5d872e5692cc2d005e6c860f16e40

PR #3316, opened 2019-12-05
Rework sphinx error handling

rustyrussell:
Closes: #3310 

I think this approach is better, and allows random experimentation with invalid onions!  We should probably write those tests now to make sure we handle them correctly...

Changelog-none


common/onion.c
```diff
@@ -0,0 +1,305 @@
...
+			*type = ONION_V0_PAYLOAD;
+		assert(*cursor - start == 1);
+		*len = 1 + 32;
+		return true;
+	}
+
+#if !EXPERIMENTAL_FEATURES
+	/* Only handle legacy format */
+	return false;
+#else
```
cdecker: TLV payloads support is not experimental, using it by default is IIRC.

cdecker: And payment_secret support is experimental only.

common/onion.c
```diff
@@ -0,0 +1,305 @@
...
+
+	return false;
+#endif /* EXPERIMENTAL_FEATURES */
+}
+
+size_t onion_payload_length(const u8 *raw_payload,
+			    bool *valid,
+			    enum onion_payload_type *type)
+{
+	size_t max = ROUTING_INFO_SIZE, len;
```
cdecker: We are mixing the assumption that `raw_payload` has a specific size and using `tal_bytelen` in a couple of places. Might be nice to consolidate that.

cfr [line 216](https://github.com/ElementsProject/lightning/pull/3316/commits/ad9d96821a3c2e920e82e36fd6fbfa999d9290a4#diff-fdb43cd4844e8dd1f19779e44c6c49f9R216) below where we use `tal_bytelen`.

common/onion.h
```diff
@@ -0,0 +1,65 @@
...
+#include <common/amount.h>
+
+struct route_step;
+
+enum onion_payload_type {
+	ONION_V0_PAYLOAD = 0,
+	ONION_TLV_PAYLOAD = 1,
+};
+
+struct onion_contents {
```
cdecker: Why not call it `onion_payload`?

common/sphinx.c
```diff
@@ -663,11 +441,8 @@ struct onionpacket *create_onionpacket(
 		size_t shiftSize = sphinx_hop_size(&sp->hops[i]);
 		memmove(packet->routinginfo + shiftSize, packet->routinginfo,
 			ROUTING_INFO_SIZE-shiftSize);
-		if (!sphinx_write_frame(packet->routinginfo, &sp->hops[i])) {
-			tal_free(packet);
-			tal_free(secrets);
-			return NULL;
-		}
+		fprintf(stderr, "Writing hop %i\n", i);
```
cdecker: Debug statement slipped through? :wink:

common/sphinx.c
```diff
@@ -799,37 +507,35 @@ struct route_step *process_onionpacket(
 	if (!blind_group_element(&step->next->ephemeralkey, &msg->ephemeralkey, blind))
 		return tal_free(step);
 
-	if (!sphinx_parse_payload(step, paddedheader))
+	payload_size = onion_payload_length(paddedheader, &valid, NULL);
+#if !EXPERIMENTAL_FEATURES
+	/* We don't even attempt to handle non-legacy or malformed headers */
```
cdecker: `headers` -> `payloads`

lightningd/peer_htlcs.c
```diff
@@ -867,21 +841,30 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
...
+			if (rs->nextcase == ONION_END) {
+				log_broken(channel->log,
+					   "htlc_acccepted hook: Can't return failure %u on last hop!",
+					   failure_code);
+				failure_code = WIRE_TEMPORARY_NODE_FAILURE;
+			} else if (!request->onion) {
+				log_broken(channel->log,
+					   "htlc_acccepted hook: Can't return failure %u on undecodable onion!",
+					   failure_code);
+				failure_code = WIRE_TEMPORARY_NODE_FAILURE;
```
cdecker: How come we can't fail if we can't decode?

common/onion.c
```diff
@@ -0,0 +1,305 @@
...
+			c->total_msat->millisatoshis /* Raw: tu64 on wire */
+				= tlv->payment_data->total_msat;
+		}
+#endif
+		tal_free(tlv);
+		return c;
+	}
+
+	/* You said it was a valid type! */
+	abort();
```
cdecker: Did we? I don't see a check anywhere close to the [call on line 975](https://github.com/ElementsProject/lightning/pull/3316/commits/ad9d96821a3c2e920e82e36fd6fbfa999d9290a4#diff-4cff6342b488c21bd84d66ba632d1d4eR975), unless it is covered by a prior call to `pull_payload_length` that I'm missing. Would be good to annotate that on line 975 if that's the case.

common/onion.c
```diff
@@ -0,0 +1,305 @@
...
+ * for backward compatibility.  Its payload is defined as:
+ *
+ * 1. type: `hop_data` (for `realm` 0)
+ * 2. data:
+ *    * [`short_channel_id`:`short_channel_id`]
+ *    * [`u64`:`amt_to_forward`]
+ *    * [`u32`:`outgoing_cltv_value`]
+ *    * [`12*byte`:`padding`]
+ */
+static u8 *v0_hop(const tal_t *ctx,
```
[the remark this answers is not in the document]
rustyrussell: Interesting: I usually prefer short names on static functions. Especially those only called in one place.

common/onion.c
rustyrussell: Ok, will fix, but I thought I copied that first #if from the old code...

rustyrussell: Yes, I did: sphinx_parse_payload used to bail if not legacy...

common/onion.c
```diff
@@ -0,0 +1,305 @@
...
+	u8 *buf = tal_arrz(ctx, u8, 1);
+	towire_short_channel_id(&buf, scid);
+	towire_u64(&buf, forward.millisatoshis); /* Raw: low-level serializer */
+	towire_u32(&buf, outgoing_cltv);
+	towire(&buf, padding, ARRAY_SIZE(padding));
+	assert(tal_bytelen(buf) == 1 + 32);
+	return buf;
+}
+
+static u8 *tlv_hop(const tal_t *ctx,
```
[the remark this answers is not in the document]
rustyrussell: Renamed these to "make_v0_hop" and "make_tlv_hop".

common/onion.c
rustyrussell: Yes, this is carefully documented in the header:

```
 * @raw_payload is must be at least ROUTING_INFO_SIZE, or the value
 * returned from a previous call to onion_payload_length.
```

But agreed it's clearer if I make "length" an explicit parameter.

common/onion.h
rustyrussell: Yep, that's better.

lightningd/peer_htlcs.c
rustyrussell: We can't send an UPDATE, since that requires us to know next hop...

common/onion.c
rustyrussell: Yes.  pull_payload_length() is called at the top of this function, and we bail if it returns false.

cdecker:
Thanks @rustyrussell for the fixups. Let me see if @bitcoin-bot is smart enough to recognize this as a clean squash :wink: 

ACK https://github.com/ElementsProject/lightning/pull/3316/commits/fb039c2979389f50b506a336769074b297806e96

common/onion.c
```diff
@@ -0,0 +1,304 @@
...
+						&p->amt_to_forward);
+		}
+		return p;
+
+	case ONION_TLV_PAYLOAD:
+		tlv = tal(p, struct tlv_tlv_payload);
+		if (!fromwire_tlv_payload(&cursor, &len, tlv))
+			return tal_free(p);
+		if (!tlv_payload_is_valid(tlv, NULL))
+			return tal_free(p);
```
cdecker: This needs to go, it's still pre-hook call, and too early to apply our checks. These should later done in `htlc_accepted_hook_callback`, otherwise we never call the hook.

common/onion.c
```diff
@@ -0,0 +1,304 @@
...
+
+		/* BOLT #4:
+		 *
+		 * The reader:
+		 *   - MUST return an error if `amt_to_forward` or
+		 *     `outgoing_cltv_value` are not present.
+		 */
+		if (!tlv->amt_to_forward || !tlv->outgoing_cltv_value)
+			return tal_free(p);
```
cdecker: The same is true for these checks.

common/onion.c
```diff
@@ -0,0 +1,304 @@
...
+
+		/* BOLT #4:
+		 *
+		 * The writer:
+		 *...
+		 * - MUST include `short_channel_id` for every non-final node.
+		 */
+		if (rs->nextcase == ONION_FORWARD) {
+			if (!tlv->short_channel_id)
+				return tal_free(p);
```
cdecker: This one needs to go as well.

cdecker:
Just to be clear we _want_ plugins to be able to receive TLV payloads that may _not_ contain an `amt_to_forward`, `outgoing_cltv_value` and `short_channel_id`. While these are incomplete from our point of view, they may make sense to a plugin which can then accept or reject the payment. `continue` without these fields may fail, so that's the only place we should perform the enforcement of these fields.

cdecker:
Added a fixup (ef8cda0b1832cb0feb6bec5d34e399efa42b760e) commit that fixes two issues:

 - We need to use the `tlv_tlv_payload_new` constructor since otherwise the `fields` array is not initialized correctly.
 - We need to use `max` instead of `len` when decoding the tlv payload since `len` also counts the bytes used for the length prefix, whereas `max` was counted down in the call to `pull_payload_length`.

No idea how this ever worked without these fixes....

rustyrussell:
OK, expanded the @cdecker fix, and added an extra sanity check that our size was sufficient in the legacy case (I think that's possible now).

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3316/commits/a49509b9ba144ae14561ce07182863eee1f3f802

PR #3319, opened 2019-12-06
improve pyln a bit

m-schmoock:
This adds a `announce_channel` to `fund_channel` and slightly optimizes RPC usage in `wait_channel_active`.

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -765,7 +765,7 @@ class LightningNode(object):
         return channel['short_channel_id']
 
     def is_channel_active(self, chanid):
-        channels = self.rpc.listchannels()['channels']
+        channels = self.rpc.listchannels(chanid)['channels']
         active = [(c['short_channel_id'], c['channel_flags']) for c in channels if c['active']]
         return (chanid, 0) in active and (chanid, 1) in active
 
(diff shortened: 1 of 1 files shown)


m-schmoock:
Travis seems fine except a unrelated hsm flake.

@cdecker question:
I tried to use `wait_channel_active` for channels not directly connected like this:

```
def test_stuff(node_factory):
    l1, l2, l3, l4 = node_factory.line_graph(4, announce_channels=True)

    # connect all and wait for each other
    l4.rpc.connect(l1.info['id'], 'localhost', l1.port)
    scid4 = l4.fund_channel(l1, 10**5)
    scid1 = l1.get_channel_scid(l2)
    scid2 = l2.get_channel_scid(l3)
    scid3 = l3.get_channel_scid(l4)

    l1.wait_channel_active(scid1)
    l1.wait_channel_active(scid2)
    l1.wait_channel_active(scid3)
    l1.wait_channel_active(scid4)
```

It works for local channels (`scid1`) but not for remote channels (`scid2-4`), it gets timeout for remote ones.
What I didn't expect is, that when debugging, it fails at `listchannels(REMOTE_SCID)`
always keeps returning an empty result. But when I do manually `lightning-cli listchannels SOME_REMOTE_SCID` it does return a non-empty result on main and testnet (which is the correct behaviour).

```
    def is_channel_active(self, chanid):
        channels = self.rpc.listchannels()['channels']
        channels = self.rpc.listchannels(chanid)['channels']
```

Any clues? Is it just the usual `@unittest.skipIf(not DEVELOPER)` slow gossip thing? If yes, how am I supposed to run such a test? I cant even run it locally..

darosior:
> Travis seems fine except a unrelated hsm flake.

Restarted. Happens now regularly :/.

PR #3320, opened 2019-12-06
Minor test cleanups

cdecker:
My weekly cleanup PR, focused on pytest and elements compatibility. Some tests accidentally broke, so I had to parametrize them. Also the "Some node exited with a non-zero return code" was annoying since it gave absolutely no indication as to which node and what the return code was.

Changelog-None

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -595,7 +595,8 @@ def test_hsm_secret_encryption(node_factory):
                     wait_for_initialized=False)
     l1.daemon.wait_for_log(r'The hsm_secret is encrypted')
     os.write(master_fd, password[2:].encode("utf-8"))
-    l1.daemon.wait_for_log("Wrong password for encrypted hsm_secret.")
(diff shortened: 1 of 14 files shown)


tests/test_wallet.py
```diff
@@ -568,7 +568,7 @@ def test_transaction_annotations(node_factory, bitcoind):
 
 @unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
 def test_hsm_secret_encryption(node_factory):
-    l1 = node_factory.get_node()
+    l1 = node_factory.get_node(may_fail=True)  # May fail when started without key
```
darosior: Hmm it shouldn't, should it ? At this point it doesn't need a key?

cdecker: The `may_fail` argument basically is just there for the teardown checks which look into the node and check if it failed the last time in the test.

You are right that with the test as it is right now (last variant exits correctly) we should not need this, let me try if it passes correctly without.

PR #3321, opened 2019-12-06
test: adds a wait_channel_active testcase for remote channels

m-schmoock:
Currently, waiting for announced **remote** channels in a simple setup fails.
For **local** channels there is no such problem, but for **remote** channels the test runs into timeouts as gossip is not able to see channels eventually.

It works however if adding `generate_block(3)` blocks if the channel is connected one hop away.
For further away channels we need an additional `generate_block(2)` in order to make the test run.

**Update:** this is actually expected behaviour, as the channel can be locally active once it reaches the agreed upon funding depth (that's negotiated between the pees and is usually 2 blocks), but in order for it to be valid from the PoV of other nodes in the network it needs 6 confirmations

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -2009,3 +2009,55 @@ def test_unicode_rpc(node_factory):
     assert(len(invoices) == 1)
     assert(invoices[0]['description'] == desc)
     assert(invoices[0]['label'] == desc)
+
+
+@unittest.skipIf(not DEVELOPER, "needs DEVELOPER=1")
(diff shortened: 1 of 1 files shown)


m-schmoock:
As we can see it also fails on travis:

- https://travis-ci.org/ElementsProject/lightning/jobs/621619683#L1285
- https://travis-ci.org/ElementsProject/lightning/jobs/621619685#L1019
- https://travis-ci.org/ElementsProject/lightning/jobs/621619686#L1266
- ...

PR #3325, opened 2019-12-09
channeld: don't send multiple update_fee until previous settled.

rustyrussell:
Our feerate logic is overly simplistic (ha!) and we seem to get confused
on back-to-back feerate updates.  Fixing it seems to require much more
complex feerate logic which I am deeply reluctant to implement, as it's
terribly fragile and will be discarded eventually once we have
simplified commitments.

Fixes: #2661
Changelog-Fixed: Occasional 'Bad commit_sig signature' channel failures with fee updates (#2661)


rustyrussell:
This doesn't work.

PR #3326, opened 2019-12-10
lightningd: don't always defer commitment_signed if we're not synced.

rustyrussell:
Because my node runs under valgrind, it can take quite a while to
sync; nodes tend to disconnect and reconnect if you block too long.

This is particularly problematic since we often update fees: when the
other side sends its commitment_signed we block.

In particular, this triggers the corner case we have where we
update_fee twice, disconnecting each time, and our state machine gets
confused (which is why we never saw this exact corner case before this
change in 0.7.3!).

Changelog-None

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -1568,24 +1568,6 @@ void peer_got_commitsig(struct channel *channel, const u8 *msg)
 	size_t i;
 	struct lightningd *ld = channel->peer->ld;
 
-	/* If we're not synced with bitcoin network, we can't accept
-	 * any HTLCs.  We stall at this point, in the hope that it
-	 * won't take long! */
-	if (!topology_synced(ld->topology)) {
-		struct deferred_commitsig *d;
-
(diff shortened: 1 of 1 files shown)


lightningd/peer_htlcs.c
```diff
@@ -1602,6 +1584,25 @@ void peer_got_commitsig(struct channel *channel, const u8 *msg)
 				    tal_hex(channel, msg));
 		return;
 	}
+
+	/* If we're not synced with bitcoin network, we can't accept
+	 * any new HTLCs.  We stall at this point, in the hope that it
+	 * won't take long! */
+	if (added && !topology_synced(ld->topology)) {
```
cdecker: Is it possible then to process failed and fulfilled HTLCs?

It at least seems safe, unless the CLTV delta is such that the downstream (i.e., the failed) HTLC isn't fully settled (claimed) before the upstream timeout expires. In that case we'd fail the upstream HTLC but would still pay out the downstream one. Then again, if that were the case we would have chosen our parameters poorly.

PR #3328, opened 2019-12-11
doc/lightningd-config: misc corrections

darosior:
And `make doc-all` also updated `addr` doc.

Changelog-None

### DIFF
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -215,15 +215,15 @@ Note that once you encrypt the \fBhsm_secret\fR this option will be mandatory fo
 
 .SH Lightning node customization options
 
- \fBalias\fR=\fIRRGGBB\fR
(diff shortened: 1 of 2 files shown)


doc/lightningd-config.5
```diff
@@ -216,16 +216,16 @@ Note that once you encrypt the \fBhsm_secret\fR this option will be mandatory fo
 .SH Lightning node customization options
 
  \fBalias\fR=\fIRRGGBB\fR
```
vasild: `alias` is not RRGGBB, maybe NAME suits here?

darosior: Good catch

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3328/commits/4177a28bb8c579be9a730b93490121180e2f66e4

Issue #3185, opened 2019-10-19
Peer connections flapping with "Failed reading header: Success"


whitslack:
### Issue and Steps to Reproduce
I have two peers that are reconnecting and failing repeatedly every 1-2 seconds. I do not know how to reproduce this failure mode. The two peers are:

* `0265fae305778b7cb157365f70cf3a2047d2cad5c1ccc5f550c6d8a033084a8ea5` ([ln.taborsky.cz](https://1ml.com/node/0265fae305778b7cb157365f70cf3a2047d2cad5c1ccc5f550c6d8a033084a8ea5))
* `0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f` ([BitMEXResearch](https://1ml.com/node/0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f))

Both of these peers are repeatedly reconnecting and failing, with a cycle in the log of this form:
```
{
   "type": "DEBUG",
   "time": "1344.732314406",
   "source": "0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f chan #14989:",
   "log": "Peer has reconnected, state CHANNELD_NORMAL"
},
{
   "type": "DEBUG",
   "time": "1344.738348636",
   "source": "lightning_channeld-0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f chan #14989:",
   "log": "pid 25681, msgfd 266"
},
{
   "type": "DEBUG",
   "time": "1344.738381691",
   "source": "0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f chan #14989:",
   "log": "Already have funding locked in (and ready to announce)"
},
{
   "type": "DEBUG",
   "time": "1344.768072142",
   "source": "lightning_channeld-0395033b252c6f40e3756984162d68174e2bd8060a129c0d3462a9370471c6d28f chan #14989:",
   "log": "init LOCAL: remote_per_commit = 02ccf2cc7376a54e4221c1166a3c634f3e57b1e533d307ba5cf1525035a72e4f2d, old_remote_per_commit = 0328794d1ed852dc0c229a302e0659dd1ea64c9ebb0eee050430b0f0e09f320bb9 next_idx_local = 4883 next_idx_remote = 4895 revocations_received = 4894 feerates 4554/4554 (range 253-45540)"
},
{
   "type": "DEBUG",
[...]

PR #3330, opened 2019-12-11
ccan: fix read_all() to always set errno on failure

vasild:
read(2) will return 0 to designate end-of-file and in this case it will
not set errno.

This is not considered as an error condition by read(2) itself, but all
callers of read_all() expect that exactly the requested number of bytes
will be read into the buffer, not less, so set errno to EBADMSG
"Bad message" in this case.

Related to https://github.com/ElementsProject/lightning/issues/3185

### DIFF
--- a/ccan/ccan/read_write_all/read_write_all.c
+++ b/ccan/ccan/read_write_all/read_write_all.c
@@ -26,12 +26,20 @@ bool read_all(int fd, void *data, size_t size)
 		ssize_t done;
 
 		done = read(fd, data, size);
-		if (done < 0 && errno == EINTR)
-			continue;
-		if (done <= 0)
+
+		switch (done) {
+		case -1:
+			if (errno == EINTR) {
+				continue;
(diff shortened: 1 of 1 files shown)


vasild:
I chose `EBADMSG` arbitrarily:
```
     89 EBADMSG Bad message. A corrupted message was detected.
```
maybe `EPROTO` is also a good choice:
```
     92 EPROTO Protocol error. A device or socket encountered an unrecoverable
             protocol error.
```

cdecker:
Thanks for the PR @vasild. This is part of ccan which is tracked externally at https://github.com/rustyrussell/ccan. Could you open a PR there? If accepted we'd update the in-tree version from the upstream version in a second step.

vasild:
@cdecker yes, sorry for the confusion.

continues at: https://github.com/rustyrussell/ccan/pull/86

cdecker:
No problem ^^

PR #3336, opened 2019-12-12
Fixup fundchannel_cancel

niftynei:
There's a few small, subtle bugs currently in `fundchannel_cancel`:
-  one: the remote peer can call `fundchannel_cancel` while waiting for lockin. it's debatable whether or not this is desirable. for now it seems prudent to remove the capability altogether.
- two: if you attempt to cancel a channel while the peer is offline, your node will crash

this PR addresses both of these issues. note that `forget_channel` is currently exposed because in a later commit (in the dual funding branch) i'm using it elsewhere.

### DIFF
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -698,14 +698,10 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
 		return;
 	}
 
-	const char *error_reason = "Cancel channel by our RPC "
-				   "command before funding "
-				   "transaction broadcast.";
-	/* Set error so we don't try to reconnect. */
-	cancel->error = towire_errorfmt(cancel, NULL, "%s", error_reason);
-
-	subd_send_msg(cancel->owner,
-		      take(towire_channel_send_error(NULL, error_reason)));
+	char *error_reason = "Cancel channel by our RPC "
+			     "command before funding "
+			     "transaction broadcast.";
+	forget_channel(cancel, error_reason);
 }
 
 struct command_result *cancel_channel_before_broadcast(struct command *cmd,
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -726,6 +726,11 @@ struct command_result *cancel_channel_before_broadcast(struct command *cmd,
 					    buffer + cidtok->start);
 	}
 
+	if (cancel_channel->funder == REMOTE)
+		return command_fail(cmd, LIGHTNINGD,
+				    "Cannot cancel channel that was "
+				    "initiated by peer");
+
 	/* Check if we broadcast the transaction. (We store the transaction type into DB
 	 * before broadcast). */
 	enum wallet_tx_type type;
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -235,17 +235,10 @@ static void peer_start_closingd_after_shutdown(struct channel *channel,
 	channel_set_state(channel, CHANNELD_SHUTTING_DOWN, CLOSINGD_SIGEXCHANGE);
 }
 
-static void handle_error_channel(struct channel *channel,
-				 const u8 *msg)
+static void forget(struct channel *channel)
 {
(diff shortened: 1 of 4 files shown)


lightningd/channel_control.c
```diff
@@ -261,6 +254,35 @@ static void handle_error_channel(struct channel *channel,
...
+	if (!fromwire_channel_send_error_reply(msg)) {
+		channel_internal_error(channel, "bad send_error_reply: %s",
+				       tal_hex(tmpctx, msg));
+		return;
+	}
+
+	forget(channel);
+}
+
+void forget_channel(struct channel *channel, char *why)
```
cdecker: This could be `const char *why`, saving the const-ness change in the last patch.

tests/test_connection.py
```diff
@@ -1005,6 +1005,27 @@ def test_funding_external_wallet_corners(node_factory, bitcoind):
     # We can cancel channel after fundchannel_complete
     assert l1.rpc.fundchannel_cancel(l2.info['id'])['cancelled']
 
+    l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
+    l1.rpc.fundchannel_start(l2.info['id'], amount)['funding_address']
+    assert l1.rpc.fundchannel_complete(l2.info['id'], prep['txid'], txout)['commitments_secured']
```
cdecker: This could do with a short description of what it tests. I guess i checks that we can still start and complete a fundchannel after cancelling?

cdecker: The `assert` also seems a bit strange, are you trying to check that there is a `commitments_secured` element in the result? In that case an `assert 'commitments_secured' in l1.rpc.fundchannel_complete...` seems more appropriate.

ZmnSCPxj:
> the remote peer can call `fundchannel_cancel` while waiting for lockin. it's debatable whether or not this is desirable. for now it seems prudent to remove the capability altogether.

Can you clarify this for me? By "remote peer" do you mean the peer that did *not* initiate the `open_channel` message?  Or do you mean that the remote peer initiated `open_channel` all the way to giving `funding_signed` but then sent an `error` because (for example) a different peer it is opening a *different* channel with on the same funding transaction timed out?

tests/test_connection.py
trueptolemy: Yeah, `'commitments_secured'` is always `true`. Checking if `'commitments_secured'` field exists is more reasonable.

lightningd/channel_control.c
```diff
@@ -661,14 +683,10 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
 		return;
 	}
 
-	const char *error_reason = "Cancel channel by our RPC "
+	char *error_reason = "Cancel channel by our RPC "
 				   "command before funding "
 				   "transaction broadcast.";
```
trueptolemy: only nit: aligned with the next line?

tests/test_connection.py
niftynei: this is a bit of an API assertion, in that it checks that the field `commitments_secured` is returned and is set to true. if this API changes, this test will fail.

niftynei: but if the `commitments_secured` is ever returned as `false` then this would not have been a successful `fundchannel_complete`.

lightningd/channel_control.c
niftynei: updated, thanks

lightningd/channel_control.c
niftynei: updated, thanks

tests/test_connection.py
ZmnSCPxj: I suppose you could instead just add a comment before this like `# A successful funding_complete will always have a commitments_secured that is true, otherwise it would have failed.`?

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/8d1d36af2167cdd4bfb7bdcd2d84860e6d4b4556

PR #3340, opened 2019-12-13
channeld: treat all incoming errors as "soft", so we retry.

rustyrussell:
We still close the channel if we *send* an error, but we seem to have hit
another case where LND sends an error which seems transient, so this will
make a best-effort attempt to preserve our channel in that case.

Some test have to be modified, since they don't terminate as they did
previously :(

### DIFF
--- a/tests/test_closing.py
+++ b/tests/test_closing.py
@@ -1608,7 +1608,7 @@ def test_shutdown(node_factory):
 
 @flaky
 @unittest.skipIf(not DEVELOPER, "needs to set upfront_shutdown_script")
-def test_option_upfront_shutdown_script(node_factory, bitcoind):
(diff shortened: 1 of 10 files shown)


tests/test_closing.py
```diff
@@ -1618,14 +1618,16 @@ def test_option_upfront_shutdown_script(node_factory, bitcoind):
     l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
     l1.fund_channel(l2, 1000000, False)
 
-    l1.rpc.close(l2.info['id'])
+    # This will block, as l12 will send an erorr but l2 will retry.
```
cdecker:
```suggestion
    # This will block, as l12 will send an error but l2 will retry.
```

PR #3344, opened 2019-12-13
build: fix developer mode compilation on FreeBSD

vasild:
Recent FreeBSD versions use LLVM's lld as a linker, not the GNU ld.
Their behavior slightly differs, so adapt the build system to handle
either one.

* The LLVM's linker prints "undefined symbol:" instead of
  "undefined reference to". Tweak tools/mockup.sh to also look for that
  message.

* The LLVM's linker may only print the first dozen errors (omitting
  the rest to avoid flooding the screen). tools/update-mocks.sh relies
  on getting all errors as it extracts the missing symbols' names from
  the error output and creates mocks for them. Detect if errors were
  omitted and re-run, telling the linker to not omit any. The GNU linker
  does not support -error-limit=0, so unfortunately we can't just run
  with that option unconditionally from the first attempt.

* Nit: FreeBSD's sed(1) prints "t" for "\t" instead of a horizontal tab.
  Use a verbatim tab in the command, instead of "\t" which works on
  both.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3344/commits/0092d4b6311752cf8f1e71adc930ed1cc1069527

PR #3345, opened 2019-12-14
connectd: translate socks5 errors to human strings

vasild:
For example:
"Tor connect out for host example.com error: 4 "
becomes:
"Error connecting to example.com: Tor server reply: host unreachable"

Also clarify the error:
"Connected out for example.com error"
to:
"Connected out for example.com error: authentication required"
because the only reason it could have failed is that the Tor socks5
server did not like any of our proposed auth methods and we only
proposed "no auth".

### DIFF
--- a/connectd/tor.c
+++ b/connectd/tor.c
@@ -41,6 +41,32 @@ struct connecting_socks {
 	struct connecting *connect;
 };
 
+static const char* socks5strerror(const tal_t *ctx, u8 code)
+{
+	// Error codes defined in https://tools.ietf.org/html/rfc1928#section-6
+	switch (code) {
+	case 0:
+		return tal_strdup(ctx, "success");
+	case 1:
+		return tal_strdup(ctx, "general SOCKS server failure");
+	case 2:
+		return tal_strdup(ctx, "connection not allowed by ruleset");
+	case 3:
+		return tal_strdup(ctx, "network unreachable");
+	case 4:
+		return tal_strdup(ctx, "host unreachable");
+	case 5:
+		return tal_strdup(ctx, "connection refused");
+	case 6:
+		return tal_strdup(ctx, "TTL expired");
+	case 7:
+		return tal_strdup(ctx, "command not supported");
+	case 8:
+		return tal_strdup(ctx, "address type not supported");
+	}
+	return tal_fmt(ctx, "unknown error: %d", code);
+}
+
 static struct io_plan *connect_finish2(struct io_conn *conn,
 				       struct connecting_socks *connect)
 {
@@ -57,6 +83,9 @@ static struct io_plan *connect_finish(struct io_conn *conn,
 	status_io(LOG_IO_IN, NULL, "proxy",
 		  connect->buffer, SIZE_OF_IPV4_RESPONSE + SIZE_OF_RESPONSE);
 
+	/* buffer[1] contains the reply status code and 0 means "success",
+	 * see https://tools.ietf.org/html/rfc1928#section-6
+	 */
 	if ( connect->buffer[1] == '\0') {
 		if ( connect->buffer[3] == SOCKS_TYP_IPV6) {
 			/* Read rest of response */
@@ -73,13 +102,15 @@ static struct io_plan *connect_finish(struct io_conn *conn,
 			return connection_out(conn, connect->connect);
 		} else {
 			status_debug
-			    ("Tor connect out for host %s error invalid type return ",
-			     connect->host);
+			    ("Tor connect out for host %s error invalid "
+			     "type return: %0x", connect->host,
+			     connect->buffer[3]);
 			return io_close(conn);
 		}
 	} else {
-		status_debug("Tor connect out for host %s error: %x ",
-			     connect->host, connect->buffer[1]);
+		status_debug("Error connecting to %s: Tor server reply: %s",
+			     connect->host,
+			     socks5strerror(tmpctx, connect->buffer[1]));
 		return io_close(conn);
 	}
 }
@@ -103,7 +134,10 @@ static struct io_plan *io_tor_connect_after_resp_to_connect(struct io_conn
 	status_io(LOG_IO_IN, NULL, "proxy", connect->buffer, 2);
 
 	if (connect->buffer[1] == SOCKS_ERROR) {
-		status_debug("Connected out for %s error", connect->host);
+		// The Tor socks5 server did not like any of our authentication
+		// methods and we provided only "no auth".
+		status_debug("Connected out for %s error: authentication required",
+			     connect->host);
 		return io_close(conn);
 	}
 	/* make the V5 request */


connectd/tor.c
```diff
@@ -57,29 +98,30 @@ static struct io_plan *connect_finish(struct io_conn *conn,
...
-			return io_close(conn);
-		}
+		return connection_out(conn, connect->connect);
 	} else {
-		status_debug("Tor connect out for host %s error: %x ",
-			     connect->host, connect->buffer[1]);
+		status_debug("Tor connect out for host %s error invalid type "
+			     "return: %0x", connect->host, connect->buffer[3]);
 		return io_close(conn);
 	}
```
vasild: What about address type "DOMAINNAME (03)" here? It is not implemented and I left it as is, but
https://tools.ietf.org/html/rfc1928#section-6
says `connect->buffer[3]` could also be `03`:
>             o  IP V4 address: X'01'
>             o  DOMAINNAME: X'03'
>             o  IP V6 address: X'04'

I think the code should just consume it an ignore it (like it does with IPv4 and IPv4 types) to be fully compliant with the Socks5 spec.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3345/commits/ac3aa1c294a192f46d211a60c485dc17c35fc0a2

Saibato:
semi NACK https://github.com/ElementsProject/lightning/commit/ac3aa1c294a192f46d211a60c485dc17c35fc0a2

I don't like **const** promises we keep anyway, but what if that triggers a bogus compiler ?, also i refuse unused int enums and compares to not (!) of value when we only can be sure what we want, so @vasild  maybe you like to look at a diff that does the same human readable error return adaption different with less change and comments on the fact that we still do by drop consume remaining bytes like you noticed if i.e. state 0x3

I think we should leave it close conn which drops consumes the remaining bytes ( if any ).
Such an unexpected answer even tor state in there code " can this ever happen? " is not of any use for us and could if handled other than drop and log say as domain tlv string create an infinite loop.

 
```
diff --git a/connectd/tor.c b/connectd/tor.c
index 6a3951b3..0e9c73a9 100644
--- a/connectd/tor.c
+++ b/connectd/tor.c
@@ -33,6 +33,33 @@
 /* some crufts can not forward ipv6*/
 #undef BIND_FIRST_TO_IPV6
 
+static char *for_most_humans_readable_tor_socks5_response(u8 return_code)
+{
+	switch (return_code) {
+		case 0x00:
+			return "Connection is not accepted";
+		case 0x01:
+			return "General SOCKS server failure";
+		case 0x02:
+			return "Connection not allowed by ruleset";
+		case 0x03:
+			return "Network is unreachable";
+		case 0x04:
+			return "Host is unreachable";
+		case 0x05:
+			return "Connection is refused";
+		case 0x06:
+			return "TTL has expired";
+		case 0x07:
+			return "Command is not supported";
+		case 0x08:
+			return "Address type is not supported";
+		default:
+		/* opps we wish to never read this in the logs */
+			return "Unknown error";
+	}
+}
+
 struct connecting_socks {
 	u8 buffer[MAX_SIZE_OF_SOCKS5_REQ_OR_RESP];
 	size_t hlen;
@@ -72,14 +99,23 @@ static struct io_plan *connect_finish(struct io_conn *conn,
 				     connect->host);
 			return connection_out(conn, connect->connect);
 		} else {
+			/* Even tor socks code states can this ever happen?
+			 * at least dump the return code to logs
+			*/
 			status_debug
-			    ("Tor connect out for host %s error invalid type return ",
-			     connect->host);
+			    ("Tor connect out for host %s error ignored response invalid response type %0x returned",
+			     connect->host,
+			     connect->buffer[1]);
+			/* Since we do not write any more, we do not need to sync_flush to consume bytes
+			 * we just close
+			*/
 			return io_close(conn);
 		}
 	} else {
-		status_debug("Tor connect out for host %s error: %x ",
-			     connect->host, connect->buffer[1]);
+		status_debug("Tor connect out for host %s error: %0x %s",
+			     connect->host,
+			     connect->buffer[1],
+			     for_most_humans_readable_tor_socks5_response(connect->buffer[1]));
 		return io_close(conn);
 	}
 }
@@ -103,7 +139,10 @@ static struct io_plan *io_tor_connect_after_resp_to_connect(struct io_conn
 	status_io(LOG_IO_IN, NULL, "proxy", connect->buffer, 2);
 
 	if (connect->buffer[1] == SOCKS_ERROR) {
-		status_debug("Connected out for %s error", connect->host);
+		// The Tor socks5 server did not like any of our authentication
+		// methods and we provided only "no auth".
+		status_debug("Connected out for %s error: authentication required",
+			     connect->host);
 		return io_close(conn);
 	}
 	/*
```

connectd/tor.c
```diff
@@ -41,6 +57,31 @@ struct connecting_socks {
...
+	case SOCKS5_REPLY_ERR_CONN_REFUSED:
+		return "connection refused";
+	case SOCKS5_REPLY_ERR_TTL_EXPIRED:
+		return "TTL expired";
+	case SOCKS5_REPLY_ERR_CMD_NOT_SUPP:
+		return "command not supported";
+	case SOCKS5_REPLY_ERR_ADDR_NOT_SUPP:
+		return "address type not supported";
+	}
+	return "other error";
```
niftynei: i'd also return the error number in this case.

ZmnSCPxj: That would suggest passing in a `const tal_t *ctx` and using `tal_fmt` / `tal_strdup`, which I suppose ties well with the thinking of @Saibato .  Caller could pass in `tmpctx` for context.

vasild:
@Saibato There is more than one way to do it, I am fine with either way as long as we get human readable messages instead of error codes.

In general a code with less nesting is more readable.

connectd/tor.c
vasild: Done.

Saibato:
> we get human readable messages instead of error codes

@vasild valid point, and that's what we try together, pls feel free to take from my diff.
In general what i write and do is CC0  

> in general a code with less nesting is more readable.

That should not be a primary focus in critical d dos prone financial software.
The shorter by fewer jumps or calls the less bugs in general.

Our followup digital overlords can read it anyway and for now i try to get error resilient byte code on most platforms.

What we write in a language and what a compiler does with that is somewhat dependent, but as you know not always what we expect. :game_die:  :game_die:  :game_die: 
Since i come from asm and low level coding, i often see the asm or physical output in my inner visual even if i write c  or program quantum gates .  So i had some inno(cent) objections.

Sadly i am a multiple personality and the other part of me is evil, that part of me thinks always how can i break this or that ?.
And i must tell you, the code you would like to merge would trigger in my view some nice facts where i can live out my dark side. :imp:

But anyway i guess we will progress from here and get a version that works stable and dumps also human readable error codes,

Even if I suspect our digital overlords are not distracted in any way by what we write as long as we not stall the ship (chip) and it's a bit funny.
But  pls don't  `const char reply_status_code = connect->buffer[1];` and let the main nesting as is. and compare to == '\0' and pls no enum here,
I saw you changed a bit like @niftynei and @ZmnSCPxj  suggested in your fixup  PR https://github.com/ElementsProject/lightning/pull/3345/commits/79ed7c75dd0b796241535b49ac39bd04ba019cb5
But It's still a NACK from me. 

BTW: i welcome your effort to port to freeBSD.  thx!
If you ever want to replace the "sucking" pkg manager of freeBSD with something like done in nix or better, i guess also @ZmnSCPxj like me would love to help out with some ideas and more ...

vasild:
@Saibato updated

Saibato:
> @Saibato updated

@vasild perfect response !
:+1:  LGTM maybe the char should be u8 i will review later today but lets wait what master of wording @niftynei  and @ZmnSCPxj and @cdecker think 

BTW I think there is an not handled edge case in the orig code 0.8.0 see fix https://github.com/ElementsProject/lightning/commit/2a6c962ff0a3fca137122af094981b9501cec648
if you like you can add this from there and make a combined commit, as you know i like to have a low profile so pls feel free to include this in your pr

vasild:
> maybe the char should be u8

Yes, I changed it to `u8` because `connect->buffer[]` is `u8`.

> and travis lamers about `reply_status_code`

Oh, that tool Travis is not smart enough to guess that I meant to say `connect->buffer[1]` instead of `reply_status_code`. The trouble with computers is that they do what you tell them, not what you want.

connectd/tor.c
```diff
@@ -41,6 +41,32 @@ struct connecting_socks {
...
+	case 5:
+		return tal_strdup(ctx, "connection refused");
+	case 6:
+		return tal_strdup(ctx, "TTL expired");
+	case 7:
+		return tal_strdup(ctx, "command not supported");
+	case 8:
+		return tal_strdup(ctx, "address type not supported");
+	}
+	return tal_fmt(ctx, "unknown error: %d", code);
```
Saibato: just some minor nit, it's OK to int from IS0/IEC 9899:1999 perspective and its just a questions, but why not add an empty` default:`
for readability? :smile:

ZmnSCPxj: >`%"PRIu8` might be  ̶c̶o̶r̶r̶e̶c̶t̶ pefered here instead of `%d` ? 

Good point.

cdecker: > just some minor nit, it's OK to int from IS0/IEC 9899:1999 perspective and its just a questions, but why not add an empty` default:`
> for readability?

It'd definitely oppose that, since it suppresses the compiler warning about not exhaustively matching all cases. If we were for example to add a case in the `enum` (not likely in this case, but still), the compiler would warn us that there is an unhandled case in the `switch` statement. Without that warning it is likely that we'd wait for a user to complain about `unknown error: xyz` and only then add `case xyz` to the switch statement.

cdecker:
IMHO the version with the `enum` was a bit more readable, and `tal_strdup`ing all the strings seems excessive, resulting in extra allocations, but for now it's ok I think.

ACK https://github.com/ElementsProject/lightning/pull/3345/commits/9bca7316329ac99f04ee7e332032c816fc20cf95

Cheers :wink:

connectd/tor.c
ZmnSCPxj: But the latest code does not use an `enum` anymore, rather a direct `u8`,so that specific objection does not apply.

vasild:
Thanks all for the reviews, it feels good to know that somebody is actually looking at the submitted code!

A followup answers:

@Saibato https://github.com/ElementsProject/lightning/commit/2a6c962ff0a3fca137122af094981b9501cec648 that commit looks out of scope for this PR. Probably a new PR should be submitted.

There was a reason to deliberately omit the `default:` when the value was an enum, exactly as @cdecker explained above. Now with `u8` it does not matter, but anyway an *empty* one would be just an unnecessary clutter.

I have changed the `%d` to `PRIu8` in a local branch here where I accumulate minor changes and will submit a PR eventually. So it will not be forgotten.

PR #3353, opened 2019-12-17
cli: allow to specify only the --rpc-file option

### LINKED ISSUE #3352: Issuing RPC commands to a specified rpc socket file without knowing its network
I'm trying to issue RPC commands using `lightning-cli` against an rpc socket file at a known location, specified with `--rpc-file`. It seems like even then, c-lightning still expects `lightning-dir` and `network` to be provided, and fails if they're invalid (when trying to move into them).

Is there a reason they're necessary? Are they used for anything other than determining the path of the socket file?

darosior:
fixes  #3352

I at first made the `--rpc-file` option take over the `--lightning_dir` one, but we don't seem to be using the config dir in the cli anyway, so ..

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -9,6 +9,7 @@
 #include <ccan/opt/opt.h>
 #include <ccan/read_write_all/read_write_all.h>
 #include <ccan/str/str.h>
+#include <ccan/tal/path/path.h>
 #include <ccan/tal/str/str.h>
 #include <common/configdir.h>
 #include <common/json.h>
@@ -490,6 +491,13 @@ int main(int argc, char *argv[])
 		try_exec_man(page, dirname(argv[0]));
 
 		tal_free(page);
+	}
+
+	/* If an absolute path to the RPC socket is given, it takes over other
+	 * configuration options. */
+	if (path_is_abs(rpc_filename)) {
+		net_dir = path_dirname(ctx, rpc_filename);
+		rpc_filename = path_basename(ctx, rpc_filename);
 	}
 
 	if (chdir(net_dir) != 0)


shesek:
This did resolve #3352, but it looks like the `chdir` was needed to resolve the path of `lightning-rpc` as relative to the data dir. Without it, c-lightning attempts to connect to `lightning-rpc` relative to the CWD and not to `<lightning-dir>/<network>/lightning-rpc`.

cdecker:
Notice also that the `chdir` papered over another quirk of AF_UNIX sockets: the path length is limited to ~103 characters. I ran into this with the testing framework which can't `chdir` and the paths are generated from the test names (sometimes resulting in longer paths). So if `chdir` is removed here, we might expose ourselves to the same issue.

darosior:
Yeah the default is relative but I tested using a relative path, will correct to handle both cases.

darosior:
So this is now the `--rpc-file` option taking over `--lightning-dir` and `--network` (and thus don't need them) if specified as an absolute path.

shesek:
Tested ACK 1b5cb2faaa4dbde909d133a292f5fa48fb66c2d2, works as expected for both absolute and relative `--rpc-file` locations.

cdecker:
The CI failures seem unrelated, working on getting those fixed first. Other than that this seems good to go :+1:

ACK https://github.com/ElementsProject/lightning/pull/3353/commits/1b5cb2faaa4dbde909d133a292f5fa48fb66c2d2

jb55:
nit:

Changelog-Changed: Usage of `lightning-cli` by specifying only `--rpc-file` has been restore.

restore -> restored

darosior:
Fixed, thanks !

ZmnSCPxj:
ACK  1a82e5a

PR #3356, opened 2019-12-18
travis: Make python environment selection less picky

cdecker:
Travis seems to have upgraded to a newer minor version of python3, so let's be
way less picky about which exact version we want to run against.

Changelog-None


ZmnSCPxj:
Seems a small change, and unblocks all our other PRs, merging as-is.

Hopefully python does not change drastically within a minor revision.

PR #3357, opened 2019-12-18
Some minor cleanups for tests and pyln

cdecker:
I bumped into a couple of issue with long path names for the lightning socket. This PR adds a wrapper for UNIX sockets that uses a short alias in `/proc/self/fd` if possible.

It also fixes a small issue with accessing `__annotations__` without checking whether it exists first (Reported by @jarret here https://github.com/lightningd/plugins/pull/70).

And finally the torture test is just really unreliable under `valgrind`, often timing out on Travis. The logic is tested in countless other tests, so there shouldn't be anything that `valgrind` can add here.

Changelog-None

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -1,9 +1,10 @@
-import json
-import logging
-import socket
-import warnings
 from decimal import Decimal
 from math import floor, log10
+import json
+import logging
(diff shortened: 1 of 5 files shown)


cdecker:
> the socket workaround is pretty cool

Yea, I got lucky that someone on the OVS project figured that one out. I couldn't have come up with it myself :-)

PR #3358, opened 2019-12-18
Add a `data_version` to the `db_write` hook an implement optimistic locking

cdecker:
Currently the `db_write` hook is a simple sink to which we write, but a plugin
has no way of telling whether there were any gaps or whether the `lightningd`
database lost some updates. This PR introduces a numeric `data_version` that
is incremented on any dirty transaction. A ditry transaction is a DB
transaction that performed some changes on the underlying database.

The `data_version` is guaranteed to change between successive changes, however
it does not guarantee monotonicity (i.e., it can wrap around at 2**32). Its
goal is to assert to a plugin that no changes to the DB have been performed by
c-lightning between successive calls to `db_write`. The `data_version`
persists across restarts so a plugin can check whether it matches a backup
after startup.

In addition this PR implements an optimistic locking strategy. The
`data_version` is incremented only if it matches what we expect from
startup. If it doesn't match we `abort()` since there might be another
instance writing to the database. This is not really important for `sqlite3`
which is already protected through the use of the pid file, but in `postgres`
we may not have such alternative protection mechanism against concurrent
instances.

It is not meant as a graceful shutdown, but rather a last resort protection. I
expect that some other locking mechanism is implemented in a wrapper for
c-lightning that'd start the instance only after acquiring ownership. As a
standalone though it should be sufficient to protect against concurrent
instances by serializing all write operations, and `abort()`ing the loser of
the write race.

As a minor detail the order in which the increment and the reporting is
currently done the plugin gets the version number after the transaction would
be applied, but that doesn't really matter, as long as the values differ
between calls.

The lock provides linearizability of DB modifications: if a database is
changed under the feet of a running process that process will `abort()`, which
from a global point of view is as if it had crashed right after the last
successful commit. Any process that also changed the DB must've started
between the last successful commit and the unsuccessful one since otherwise
its counters would not have matched (which would also have aborted that
transaction). So this reduces all the possible timelines to an equivalent
where the first process died, and the second process recovered from the DB.

### DIFF
--- a/wallet/db_common.h
+++ b/wallet/db_common.h
@@ -34,6 +34,10 @@ struct db {
 	/* Were there any modifying statements in the current transaction?
 	 * Used to bump the data_version in the DB.*/
 	bool dirty;
+
+	/* The current DB version we expect to update if changes are
+	 * committed. */
+	u32 data_version;
 };
 
 struct db_query {
--- a/wallet/db_common.h
+++ b/wallet/db_common.h
@@ -30,6 +30,10 @@ struct db {
 	char *error;
 
 	struct log *log;
+
+	/* Were there any modifying statements in the current transaction?
+	 * Used to bump the data_version in the DB.*/
+	bool dirty;
 };
 
 struct db_query {
--- a/lightningd/plugin_hook.c
+++ b/lightningd/plugin_hook.c
@@ -156,13 +156,14 @@ static void db_hook_response(const char *buffer, const jsmntok_t *toks,
 	io_break(ph_req);
 }
 
-void plugin_hook_db_sync(struct db *db, const char **changes, const char *final)
+void plugin_hook_db_sync(struct db *db)
 {
 	const struct plugin_hook *hook = &db_write_hook;
 	struct jsonrpc_request *req;
 	struct plugin_hook_request *ph_req;
 	void *ret;
 
+	const char **changes = db_changes(db);
 	if (!hook->plugin)
 		return;
 
@@ -177,8 +178,6 @@ void plugin_hook_db_sync(struct db *db, const char **changes, const char *final)
 	json_array_start(req->stream, "writes");
 	for (size_t i = 0; i < tal_count(changes); i++)
 		json_add_string(req->stream, NULL, changes[i]);
-	if (final)
-		json_add_string(req->stream, NULL, final);
 	json_array_end(req->stream);
 	jsonrpc_request_end(req);
 
--- a/lightningd/plugin_hook.c
+++ b/lightningd/plugin_hook.c
@@ -175,6 +175,8 @@ void plugin_hook_db_sync(struct db *db)
 	ph_req->hook = hook;
 	ph_req->db = db;
 
+	json_add_num(req->stream, "data_version", db_data_version_get(db));
+
 	json_array_start(req->stream, "writes");
 	for (size_t i = 0; i < tal_count(changes); i++)
 		json_add_string(req->stream, NULL, changes[i]);
--- a/lightningd/plugin_hook.h
+++ b/lightningd/plugin_hook.h
@@ -108,8 +108,7 @@ bool plugin_hook_unregister(struct plugin *plugin, const char *method);
 /* Unregister all hooks a plugin has registered for */
 void plugin_hook_unregister_all(struct plugin *plugin);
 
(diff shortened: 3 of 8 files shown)


cdecker:
Still checking rebaseability, will un-draft asap :-)

ZmnSCPxj:
When considering as well the possibillity of chaining `db_write` hooks, perhaps we should consider a two-stage commit as well:

* `db_write` hook, where plugins save the queries in some log, and return with `"result" : "continue"` if they are able to save the queries or `"result" : "abort"` if they fail.
* `db_commit` notification, where if all the plugins hooked on `db_write` returned `"result" : "continue"`, this notification is broadcast so that plugins can advance the state.

We may not have hook chaining yet, but it seems good to prepare the interface at this point as well, so that plugin authors for remote backup in particular can start adapting to the new interface, then later we can support hook chaining and thereby allow multiple backup plugins to hook the `db_write` hook and stop the daemon if any backup is having trouble.

cdecker:
> When considering as well the possibillity of chaining `db_write` hooks, perhaps we should consider a two-stage commit as well:
> 
> * `db_write` hook, where plugins save the queries in some log, and return with `"result" : "continue"` if they are able to save the queries or `"result" : "abort"` if they fail.
> * `db_commit` notification, where if all the plugins hooked on `db_write` returned `"result" : "continue"`, this notification is broadcast so that plugins can advance the state.
> 
> We may not have hook chaining yet, but it seems good to prepare the interface at this point as well, so that plugin authors for remote backup in particular can start adapting to the new interface, then later we can support hook chaining and thereby allow multiple backup plugins to hook the `db_write` hook and stop the daemon if any backup is having trouble.

Totally with you on the chainability of hooks, standardizing on `{"result": "xyz"}` allowing the plugin to abort or continue is important. It is possible today by having the hook return an error, but that looks a lot scarier to the user when he gets a backtrace with an inscrutable error message. So cleaning that up would be great.

I'm not convinced that a `db_commit` notification is something we should do. Plugins have to manage cases with and without that notification anyway (we might exit inbetween calling the hook and committing) so adding the notification might lead to more complex logic given now two different entrypoints into the logic are possible. Worse is that plugin developers may start relying on the notification, causing issues for committed but not notified transactions. These two arguments lead me to be cautious with this kind of apparent optimization.

darosior:
> Worse is that plugin developers may start relying on the notification, causing issues for committed but not notified transactions.

FWIW we could send a notification at `begin_transaction` and one at `commit_transaction`, so that the plugin knows what's committed or not ?

cdecker:
> > Worse is that plugin developers may start relying on the notification, causing issues for committed but not notified transactions.
> 
> FWIW we could send a notification at `begin_transaction` and one at `commit_transaction`, so that the plugin knows what's committed or not ?

But the plugin doesn't know as a matter of fact, since we might terminate inbetween calling the hook and issuing the notification, so the plugin _must_ handle the case where it doesn't get a notification but the DB transaction is still committed. Once you already have to handle that, why add yet another possible entrypoint for the happy case, and not just always assume the bad case? No matter how many notifications we add the plugin will have to have the ability to handle the case that a notification never comes :-)

wallet/db_common.h
```diff
@@ -34,6 +34,10 @@ struct db {
 	/* Were there any modifying statements in the current transaction?
 	 * Used to bump the data_version in the DB.*/
 	bool dirty;
+
+	/* The current DB version we expect to update if changes are
+	 * committed. */
+	u32 data_version;
```
niftynei: u64 not u32?

cdecker: Yes, `intval` in the `vars` table is defined as an `INTEGER` which gets translated into a `u32` in postgres and the size has to match in non-sqlite3 databases. So I thought I shouldn't be too picky about this value.

Alternatives are 1) creation of a new table, or 2) adding a new field `bigintval` to the `vars` table. If this is a problem I'd prefer the second. But really all we're guaranteeing is that there is enough variety to identify if we skipped writes or have lost some. In particular we can't do much if we're off by more than 2: if the plugin missed an update we must take a snapshot, if the DB moved back to an earlier point in time we must abort. We just need to have the probability of accidentally using the same version again small enough (1/2**32 seems good enough for me 😉).

niftynei: sweet, ok. that's a good thing to know about int -> u32 in postgres.

PR #3361, opened 2019-12-19
fix: normalize pip requirements

m-schmoock:
This will change some `requirements.txt` of pyln-testing in a way that
it does not require different package version i.e. to `tests/requirements.txt`.

The reason for this is that users are not forced to hassle with pyenv
or virtualenv and could just use `--user`.

```bash
pip install --user -r tests/requirements.txt -r contrib/pyln-testing/requirements.txt  # ...
```

Changelog-None

### DIFF
--- a/contrib/pyln-testing/requirements.txt
+++ b/contrib/pyln-testing/requirements.txt
@@ -1,6 +1,6 @@
-pytest==5.0.1
+pytest==5.3.1
 Flask==1.1.1
-cheroot==6.5.5
+cheroot==8.2.1
 ephemeral-port-reserve==1.1.1
-python-bitcoinlib==0.10.1
+python-bitcoinlib==0.10.2
 psycopg2-binary==2.8.4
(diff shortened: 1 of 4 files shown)


contrib/pyln-testing/requirements.txt
```diff
@@ -1,6 +1,7 @@
-pytest==5.0.1
+pytest==5.3.1
 Flask==1.1.1
-cheroot==6.5.5
+cheroot==8.2.1
 ephemeral-port-reserve==1.1.1
 python-bitcoinlib==0.10.1
```
darosior: Trailing old requirement?

m-schmoock: good question, I think this was just left over when reorganizing from pylightning into pyln.

m-schmoock: we could/should update `contrib/Dockerfile.builder` and `contrib/Dockerfile.builder.i386` as well they also point to that old version.

darosior: I was talking about `python-bitcoinlib`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3361/commits/ea03f3cbcb77a6e21585d30d84202aaa974b0086

PR #3363, opened 2019-12-19
hsmd: Added fields to hsm_sign_remote_commitment_tx to allow complete validation.

ksedgwic:
Added output_witscripts, remote_per_commit and option_static_remotekey  to hsm_sign_remote_commitment_tx to allow complete validation.

### DIFF
--- a/bitcoin/tx.c
+++ b/bitcoin/tx.c
@@ -398,6 +398,7 @@ struct bitcoin_tx *bitcoin_tx(const tal_t *ctx,
 	tx->input_amounts = tal_arrz(tx, struct amount_sat*, input_count);
 	tx->wtx->locktime = 0;
 	tx->wtx->version = 2;
+	tx->output_witscripts = tal_arrz(tx, struct witscript*, output_count);
 	tx->chainparams = chainparams;
 	return tx;
 }
--- a/bitcoin/tx.h
+++ b/bitcoin/tx.h
@@ -12,6 +12,10 @@
 
 #define BITCOIN_TX_DEFAULT_SEQUENCE 0xFFFFFFFF
 
+struct witscript {
+    u8 *ptr;
+};
+
 struct bitcoin_txid {
 	struct sha256_double shad;
 };
@@ -23,6 +27,9 @@ struct bitcoin_tx {
 	 * unknown) */
 	struct amount_sat **input_amounts;
 	struct wally_tx *wtx;
+
+	/* Need the output wscripts in the HSM to validate transaction */
+	struct witscript **output_witscripts;
 
 	/* Keep a reference to the ruleset we have to abide by */
 	const struct chainparams *chainparams;
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -994,7 +994,10 @@ static secp256k1_ecdsa_signature *calc_commitsigs(const tal_t *ctx,
 
 	msg = towire_hsm_sign_remote_commitment_tx(NULL, txs[0],
 						   &peer->channel->funding_pubkey[REMOTE],
-						   *txs[0]->input_amounts[0]);
+						   *txs[0]->input_amounts[0],
+						   (const struct witscript **) txs[0]->output_witscripts,
+						   &peer->remote_per_commit,
+						   peer->channel->option_static_remotekey);
 
 	msg = hsm_req(tmpctx, take(msg));
 	if (!fromwire_hsm_sign_tx_reply(msg, commit_sig))
--- a/channeld/commit_tx.c
+++ b/channeld/commit_tx.c
@@ -37,7 +37,7 @@ size_t commit_tx_num_untrimmed(const struct htlc **htlcs,
 static void add_offered_htlc_out(struct bitcoin_tx *tx, size_t n,
 				 const struct htlc *htlc,
 				 const struct keyset *keyset,
-				 struct witscript * o_wscript)
+				 struct witscript *o_wscript)
 {
 	struct ripemd160 ripemd;
 	u8 *wscript, *p2wsh;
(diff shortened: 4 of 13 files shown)


cdecker:
@bitcoin-bot was offline for a couple of hours, he should be back now :-)

bitcoin/tx.c
```diff
@@ -398,6 +398,7 @@ struct bitcoin_tx *bitcoin_tx(const tal_t *ctx,
 	tx->input_amounts = tal_arrz(tx, struct amount_sat*, input_count);
 	tx->wtx->locktime = 0;
 	tx->wtx->version = 2;
+	tx->output_witscripts = tal_arrz(tx, struct witscript*, output_count);
```
cdecker: We usually don't zero our allocations right away (prefer `tal_arr`) since that'd mask accesses to uninitialized memory that can be caught with `valgrind`. In this case it's ok since you want to have a NULL-pointer out of the gate.

bitcoin/tx.h
```diff
@@ -12,6 +12,10 @@
 
 #define BITCOIN_TX_DEFAULT_SEQUENCE 0xFFFFFFFF
 
+struct witscript {
+    u8 *ptr;
+};
+
```
cdecker: In many other locations we just hand a `u8 *` around for this. Is your goal to have more type-checking here?

channeld/channeld.c
```diff
@@ -993,7 +993,10 @@ static secp256k1_ecdsa_signature *calc_commitsigs(const tal_t *ctx,
 
 	msg = towire_hsm_sign_remote_commitment_tx(NULL, txs[0],
 						   &peer->channel->funding_pubkey[REMOTE],
-						   *txs[0]->input_amounts[0]);
+						   *txs[0]->input_amounts[0],
+						   (const struct witscript **) txs[0]->output_witscripts,
```
cdecker: These fields could be folded into the `towire_bitcoin_tx` and `fromwire_bitcoin_tx` helpers allowing you to just pass in `txs[0]` here and serialize/deserialize automatically.

cdecker: Might be a slightly bigger change though, so maybe we can do it in a second step :wink:

channeld/commit_tx.c
```diff
@@ -209,6 +221,11 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 		SUPERVERBOSE("# to-local amount %s wscript %s\n",
 			     type_to_string(tmpctx, struct amount_sat, &amount),
 			     tal_hex(tmpctx, wscript));
+		tx->output_witscripts[n] =
+			tal(tx->output_witscripts, struct witscript);
+		tx->output_witscripts[n]->ptr =
+			tal_dup_arr(tx->output_witscripts[n], u8,
+				    wscript, tal_count(wscript), 0);
```
cdecker: If we just hand `u8 *` around we can save these double-allocations.

cdecker:
Some minor things that could be simplified, but overall I'm all for giving the `hsmd` the information it needs to verify independently :+1:

bitcoin/tx.c
ksedgwic: Yes, want them NULL in the cases where we don't set a witscript.

bitcoin/tx.h
ksedgwic: I'm trying to marshal a vector of witness scripts in the wire message:
https://github.com/ksedgwic/lightning-1/blob/33d756bf6a65c5fa8b1d11b5233029bd2f49a950/hsmd/hsm_wire.csv#L161-L162
I don't see how to specify an array of length-prefixed u8* should be serialized without the wrapping struct.
Is there a better way?

ksedgwic: ping?

wire/fromwire.c
```diff
@@ -392,6 +392,19 @@ struct bitcoin_tx_output *fromwire_bitcoin_tx_output(const tal_t *ctx,
...
+{
+	struct witscript *retval = tal(ctx, struct witscript);
+	u16 len = fromwire_u16(cursor, max);
+	if (len == 0) {
+		retval->ptr = NULL;
+	} else {
+		retval->ptr = tal_arr(retval, u8, len);
+		fromwire_u8_array(cursor, max, retval->ptr, len);
+	}
+	return retval;
```
rustyrussell: For consistency, I think this should return NULL (just like you allocate witscripts in a tx as NULL), rather than a struct containing a NULL ptr.

Something like:

```
struct witscript *retval;
u16 len = fromwire_u16(cursor, max);

if (!len)
    return NULL;

retval = tal(ctx, struct witscript);
retval->ptr = tal_arr(retval, u8, len);
fromwire_u8_array(cursor, max, retval->ptr, len);
return retval;
```

wire/towire.c
```diff
@@ -257,6 +257,16 @@ void towire_bitcoin_tx_output(u8 **pptr, const struct bitcoin_tx_output *output)
 	towire_u8_array(pptr, output->script, tal_count(output->script));
 }
 
+void towire_witscript(u8 **pptr, const struct witscript *script)
+{
+	if (script == NULL || script->ptr == NULL) {
+		towire_u16(pptr, 0);
+	} else {
```
rustyrussell: Subtly, the script->ptr check here is redundant (as `tal_count(NULL)` == 0), though I'd be tempted to assert it's set, once the previous change is done so it's never NULL.

ZmnSCPxj:
:+1: type safety

wire/fromwire.c
ksedgwic: Fixed.

wire/towire.c
ksedgwic: Fixed.

cdecker:
I think we need a minor cleanup, but otherwise LGTM. 8c593e2ce5f7e3094e11543f8333cad0617a8870 and c1e1634403c8a44eef3605285ec82327e5adfea6 need to be squashed into 5a47254820ac4a2d15e7baebf1627eb73a045e68.

PR #3372, opened 2019-12-23
lightningd: Added --alt-subdaemon command to allow alternate subdaemons.

ksedgwic:
This PR adds --alt-subdaemon to facilitate substituting a standard subdaemon with an alternate.

Examples:

    # Use remote_hsmd instead of lightning_hsmd for signing:
    lightningd --alt-subdaemon=lightning_hsmd:remote_hsmd ...

    # Same as previous but daemon is located w/ absolute path:
    lightningd --alt-subdaemon=lightning_hsmd:/some/where/remote_hsmd ...

### DIFF
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -107,6 +107,16 @@ Sets the working directory. All files (except *--conf* and
 is only valid on the command-line, or in a configuration file specified
 by *--conf*.
 
+ **alt-subdaemon**=*SUBDAEMON*:*PATH*
+Specifies an alternate subdaemon binary.  If the supplied path
+is relative the subdaemon binary is found in the working directory.
+This option may be specified multiple times, but only once for
+each subdaemon.
+
+ So, **alt-subdaemon=lightning_hsmd:remote_signer** would use a
+hypothetical remote signing proxy instead of the standard *lightning_hsmd*
+binary.
+
  **pid-file**=*PATH*
 Specify pid file to write to.
 
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -110,8 +110,7 @@ by *--conf*.
  **alt-subdaemon**=*SUBDAEMON*:*PATH*
 Specifies an alternate subdaemon binary.  If the supplied path
 is relative the subdaemon binary is found in the working directory.
-This option may be specified multiple times, but only once for
-each subdaemon.
+This option may be specified multiple times.
 
  So, **alt-subdaemon=lightning_hsmd:remote_signer** would use a
 hypothetical remote signing proxy instead of the standard *lightning_hsmd*
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -109,12 +109,12 @@ by *--conf*.
 
  **subdaemon**=*SUBDAEMON*:*PATH*
 Specifies an alternate subdaemon binary.
-Current subdaemons are *channel*, *closing*,
-*connect*, *gossip*, *hsm*, *onchain*, and *opening*.
+Current subdaemons are *channeld*, *closingd*,
+*connectd*, *gossipd*, *hsmd*, *onchaind*, and *openingd*.
 If the supplied path is relative the subdaemon binary is found in the
 working directory. This option may be specified multiple times.
 
- So, **subdaemon=lightning_hsmd:remote_signer** would use a
+ So, **subdaemon=hsmd:remote_signer** would use a
 hypothetical remote signing proxy instead of the standard *lightning_hsmd*
 binary.
 
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -107,12 +107,14 @@ Sets the working directory. All files (except *--conf* and
 is only valid on the command-line, or in a configuration file specified
 by *--conf*.
 
- **alt-subdaemon**=*SUBDAEMON*:*PATH*
-Specifies an alternate subdaemon binary.  If the supplied path
-is relative the subdaemon binary is found in the working directory.
-This option may be specified multiple times.
+ **subdaemon**=*SUBDAEMON*:*PATH*
+Specifies an alternate subdaemon binary.
+Current subdaemons are *channel*, *closing*,
+*connect*, *gossip*, *hsm*, *onchain*, and *opening*.
+If the supplied path is relative the subdaemon binary is found in the
+working directory. This option may be specified multiple times.
 
- So, **alt-subdaemon=lightning_hsmd:remote_signer** would use a
+ So, **subdaemon=lightning_hsmd:remote_signer** would use a
 hypothetical remote signing proxy instead of the standard *lightning_hsmd*
 binary.
 
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -81,6 +81,12 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+static void destroy_alt_subdaemons(struct lightningd *ld);
+#if DEVELOPER
+static void memleak_help_alt_subdaemons(struct htable *memtable,
+					struct lightningd *ld);
+#endif /* DEVELOPER */
+
 /*~ The core lightning object: it's passed everywhere, and is basically a
  * global variable.  This new_xxx pattern is something we'll see often:
  * it allocates and initializes a new structure, using *tal*, the hierarchical
@@ -248,6 +254,11 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
 	 */
 	ld->encrypted_hsm = false;
 
+	/* This is used to override subdaemons */
+	strmap_init(&ld->alt_subdaemons);
+	tal_add_destructor(ld, destroy_alt_subdaemons);
+	memleak_add_helper(ld, memleak_help_alt_subdaemons);
+
 	/*~ We change umask if we daemonize, but not if we don't. Initialize the
 	 * initial_umask anyway as we might rely on it later (`plugin start`). */
 	ld->initial_umask = umask(0);
@@ -278,6 +289,57 @@ static const char *subdaemons[] = {
 	"lightning_openingd"
 };
 
+/* Return true if called with a recognized subdaemon */
+bool is_subdaemon(const char *sdname)
+{
+	for (size_t i = 0; i < ARRAY_SIZE(subdaemons); i++)
+		if (strcmp(sdname, subdaemons[i]) == 0)
+			return true;
+	return false;
+}
+
+static void destroy_alt_subdaemons(struct lightningd *ld)
+{
+	strmap_clear(&ld->alt_subdaemons);
+}
+
+#if DEVELOPER
+static void memleak_help_alt_subdaemons(struct htable *memtable,
+					struct lightningd *ld)
+{
+	memleak_remove_strmap(memtable, &ld->alt_subdaemons);
+}
+#endif /* DEVELOPER */
+
+const char *subdaemon_path(const struct lightningd *ld, const char *name)
+{
+	const char *dpath;
+
+	/*~ CCAN's path module uses tal, so wants a context to
+	 * allocate from.  We have a magic convenience context
+	 * `tmpctx` for temporary allocations like this.
+	 *
+	 * Because all our daemons at their core are of form `while
+	 * (!stopped) handle_events();` (an event loop pattern), we
+	 * can free `tmpctx` in that top-level loop after each event
+	 * is handled.
+	 */
+
+	/* Is there an alternate path for this subdaemon? */
+	const char *alt = strmap_get(&ld->alt_subdaemons, name);
+	if (alt) {
+		/* Is the alternate path absolute? */
+		if (alt[0] == '/')
+			dpath = tal_strdup(tmpctx, alt);
+		else
+			dpath = path_join(tmpctx, ld->daemon_dir, alt);
+	} else {
+		/* This subdaemon is found in the standard place. */
+		dpath = path_join(tmpctx, ld->daemon_dir, name);
+	}
+	return dpath;
+}
+
 /*~ Check we can run them, and check their versions */
 void test_subdaemons(const struct lightningd *ld)
 {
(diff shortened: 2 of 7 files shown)


lightningd/lightningd.c
```diff
@@ -263,6 +274,57 @@ static const char *subdaemons[] = {
...
+	 * (!stopped) handle_events();` (an event loop pattern), we
+	 * can free `tmpctx` in that top-level loop after each event
+	 * is handled.
+	 */
+
+	/* Is there an alternate path for this subdaemon? */
+	const char *alt = strmap_get(&ld->alt_subdaemons, name);
+	if (alt) {
+		/* Is the alternate path absolute? */
+		if (alt[0] == '/')
```
darosior: `path_is_abs(alt)` is already defined in ccan ;-)

lightningd/options.c
```diff
@@ -191,6 +191,38 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
 	return opt_add_addr_withtype(arg, ld, ADDR_LISTEN_AND_ANNOUNCE, true);
 }
 
+static char *opt_alt_subdaemon(const char *arg, struct lightningd *ld)
+{
+	// example arg: "hsmd:rmthsmd"
```
darosior: style nit: We use `/* */`

lightningd/lightningd.h
```diff
@@ -249,12 +250,20 @@ struct lightningd {
...
 	bool encrypted_hsm;
+
+	STRMAP(const char *) alt_subdaemons;
 };
 
 /* Turning this on allows a tal allocation to return NULL, rather than aborting.
  * Use only on carefully tested code! */
 extern bool tal_oom_ok;
 
+/* Return true if called with a recognized subdaemon, eg: "lightning_hsmd" */
```
darosior: nit: `Returns`

lightningd/options.c
```diff
@@ -191,6 +191,38 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+	char *argbuf = tal_strdup(ld, arg);
+	char *colon = strchr(argbuf, ':');
+	if (!colon)
+	{
```
darosior: (here and below) We use
```
if (!colon) {
```

lightningd/options.c
```diff
@@ -191,6 +191,38 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+		tal_free(argbuf);
```
darosior: We'll free its parent immediately anyway?

lightningd/options.c
```diff
@@ -191,6 +191,38 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+	}
+	*colon = '\0';
+
+	if (!is_subdaemon(argbuf))
+	{
+		char *ret = tal_fmt(NULL, "%s is not a subdaemon", argbuf);
+		tal_free(argbuf);
+		return ret;
+	}
+	char *subdaemon = tal_strdup(ld, argbuf);
```
darosior: It's preferable to declare it at the top of the function ?

lightningd/lightningd.c
```diff
@@ -263,6 +274,57 @@ static const char *subdaemons[] = {
...
+static void destroy_alt_subdaemons(struct lightningd *ld)
+{
+	strmap_clear(&ld->alt_subdaemons);
+}
+
+#if DEVELOPER
+static void memleak_help_alt_subdaemons(struct htable *memtable,
+					struct lightningd *ld)
+{
+	memleak_remove_strmap(memtable, &ld->alt_subdaemons);
```
darosior: You forgot to `#include <lightningd/memdump.h>`

lightningd/options.c
devrandom: also, shouldn't it be `lightning_hsmd:rmthsmd`?

lightningd/lightningd.c
ksedgwic: fixed

lightningd/options.c
ksedgwic: fixed both

lightningd/options.c
ksedgwic: fixed

lightningd/options.c
ksedgwic: Not sure about this.  It needs to be after argbuf is modified by replacing the ':' w/ '/0'

Do you mean declared at the top and then initialized later?

lightningd/lightningd.c
ksedgwic: fixed.  Thanks for the feedback!

lightningd/lightningd.h
ksedgwic: fixed

lightningd/options.c
ksedgwic: fixed

lightningd/options.c
ZmnSCPxj: Generally it is declared at the start of a scope (not necessarily start of a function).

lightningd/lightningd.c
```diff
@@ -263,6 +276,57 @@ static const char *subdaemons[] = {
...
+}
+#endif /* DEVELOPER */
+
+const char *subdaemon_path(const struct lightningd *ld, const char *name)
+{
```
rustyrussell: We usually pass a ctx arg in here: caller will use tmpctx but then it knows that the return value is allocated off that.

lightningd/lightningd.c
```diff
@@ -263,6 +276,57 @@ static const char *subdaemons[] = {
...
+	 */
+
+	/* Is there an alternate path for this subdaemon? */
+	const char *alt = strmap_get(&ld->alt_subdaemons, name);
+	if (alt) {
+		/* Is the alternate path absolute? */
+		if (path_is_abs(alt))
+			dpath = tal_strdup(tmpctx, alt);
+		else
+			dpath = path_join(tmpctx, ld->daemon_dir, alt);
```
rustyrussell: path_join is a noop if second arg is absolute, so this can be a one-liner?

lightningd/options.c
```diff
@@ -191,6 +191,27 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+	if (!is_subdaemon(argbuf))
+		return tal_fmt(NULL, "%s is not a subdaemon", argbuf);
+
+	char *subdaemon = tal_strdup(ld, argbuf);
+	char *sdpath = tal_strdup(ld, colon+1);
+	if (!strmap_add(&ld->alt_subdaemons, subdaemon, sdpath))
+		return tal_fmt(NULL, "%s already exists", argbuf);
+
+	tal_free(argbuf);
+	return NULL;
```
rustyrussell: I think tal_strsplit will do this slightly more neatly.

And in general, later args should replace earlier ones. So free the old one if any, rather than complain?

lightningd/options.c
ksedgwic: fixed

lightningd/lightningd.c
ksedgwic: fixed

lightningd/lightningd.c
ksedgwic: fixed

lightningd/options.c
ksedgwic: done

devrandom:
seems like there was an intermittent failure in the build for `6c727d6`

arowser:
Nice pr, @rustyrussell @cdecker  another idea: if change the message channels between sub daemons from stdio to socket , then the subdaemon could run on different machines, even run remote-hsmd on a real remote HSM.

ZmnSCPxj:
@arowser how would we start the processes remotely? I suggest rather that such setups can be done with a bridge daemon that replaces `--alt-subdaemon` locally, then just `splice` (on Linux; otherwise just copy-in-userspace) from the stdio to a local pipe and to a remote socket (on Linux, pipes are "just" representations of kernel buffers, so the local pipe is only needed to properly adapt to any combination of endpoint devices), with the bridge daemon knowing how to properly initiate starting the remote process.

lightningd/options.c
```diff
@@ -191,6 +191,40 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+	char *sdpath;
+	char *prevname;
+	char *prevval;
+
+	/* example arg: "lightning_hsmd:remote_hsmd" */
+	char *argbuf = tal_strdup(ld, arg);
+	char *colon = strchr(argbuf, ':');
+	if (!colon)
+		return tal_fmt(NULL, "argument must contain \':\'");
+	*colon = '\0';
```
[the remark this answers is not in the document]
ksedgwic: Fixed

lightningd/options.c
```diff
@@ -191,6 +191,40 @@ static char *opt_add_addr(const char *arg, struct lightningd *ld)
...
+	/* Remove any preexisting alt subdaemon mapping. */
+	prevname = strmap_del(&ld->alt_subdaemons, subdaemon,
+			      (const char **) &prevval);
+	if (prevname) {
+		tal_free(prevname);
+		tal_free(prevval);
+	}
+
+	sdpath = tal_strdup(ld, colon+1);
+	strmap_add(&ld->alt_subdaemons, subdaemon, sdpath);
```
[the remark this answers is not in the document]
ksedgwic: Fixed

lightningd/options.c
```diff
@@ -1080,6 +1124,35 @@ static void json_add_opt_addrs(struct json_stream *response,
...
+
+static void json_add_opt_alt_subdaemons(struct json_stream *response,
+					const char *name0,
+					alt_subdaemon_map *alt_subdaemons)
+{
+	struct json_add_opt_alt_subdaemon_args args;
+	args.name0 = name0;
+	args.response = response;
+	strmap_iterate(alt_subdaemons, json_add_opt_alt_subdaemon, &args);
+}
```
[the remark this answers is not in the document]
ksedgwic: Fixed.

doc/lightningd-config.5.md
```diff
@@ -107,6 +107,16 @@ Sets the working directory. All files (except *--conf* and
 is only valid on the command-line, or in a configuration file specified
 by *--conf*.
 
+ **alt-subdaemon**=*SUBDAEMON*:*PATH*
```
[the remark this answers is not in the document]
cdecker: Wrote this a couple of days ago, and forgot to submit :smirk:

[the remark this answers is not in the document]
ksedgwic: Are you proposing just changing the name of the flag?  Or should the behavior change too?
In a sense it already defaults to the current subdaemon binaries ...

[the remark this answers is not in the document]
rustyrussell: Yeah, that makes sense:

1. remove the 'lightningd_' prefix from subdaemon names used here; they're kinda redundant.
2. make the argument name '--subdaemon', so you'd use it like '--subdaemon=hsm:/tmp/myhsm'.
3. the default now prepends "lightning_" and the daemon path.

But it's looking good!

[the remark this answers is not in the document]
ksedgwic: Done.

rustyrussell:
@ksedgwic Thanks for your patience!

Almost there: let's not skip the final `d` (my typo, sorry!):

```patch
diff --git a/doc/lightningd-config.5.md b/doc/lightningd-config.5.md
index c72121bff..9d6548505 100644
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -109,12 +109,12 @@ by *--conf*.
 
  **subdaemon**=*SUBDAEMON*:*PATH*
 Specifies an alternate subdaemon binary.
-Current subdaemons are *channel*, *closing*,
-*connect*, *gossip*, *hsm*, *onchain*, and *opening*.
+Current subdaemons are *channeld*, *closingd,
+*connectd*, *gossipd*, *hsmd*, *onchaind*, and *openingd*.
 If the supplied path is relative the subdaemon binary is found in the
 working directory. This option may be specified multiple times.
 
- So, **subdaemon=lightning_hsmd:remote_signer** would use a
+ So, **subdaemon=hsmd:remote_signer** would use a
 hypothetical remote signing proxy instead of the standard *lightning_hsmd*
 binary.
 
index 2f658affc..a43f3f07a 100644
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -316,11 +316,9 @@ static void memleak_help_alt_subdaemons(struct htable *memtable,
 
 const char *subdaemon_path(const tal_t *ctx, const struct lightningd *ld, const char *name)
 {
-	/* Strip the leading "lightning_" and following 'd' before looking
-	 * in alt_subdaemons
-	 */
-	assert(strlen(name) > 10 + 1);	/* len("lightning_") + len("d") */
-	const char *short_name = tal_strndup(ctx, name + 10, strlen(name) - 11);
+	/* Strip the leading "lightning_" before looking in alt_subdaemons */
+	assert(strlen(name) > strlen("lightning_"));
+	const char *short_name = tal_strdup(ctx, name + strlen("lightning_"));
 
 	/* Is there an alternate path for this subdaemon? */
 	const char *dpath;
```

And `if (strncmp(sdname, subdaemons[i] + 10, strlen(sdname)) == 0)`
should probably be strcmp, e.g.: `if (streq(sdname, subdaemons[i] + strlen("lightning_"))`
otherwise you can do `--subdaemon=c:mydaemon` and replace all subdaemons starting with c, which is not what we want?

Cheers!

lightningd/lightningd.c
```diff
@@ -316,11 +316,12 @@ static void memleak_help_alt_subdaemons(struct htable *memtable,
...
-	/* Strip the leading "lightning_" and following 'd' before looking
-	 * in alt_subdaemons
+	/* Strip the leading "lightning_" before looking in alt_subdaemons.
 	 */
-	assert(strlen(name) > 10 + 1);	/* len("lightning_") + len("d") */
-	const char *short_name = tal_strndup(ctx, name + 10, strlen(name) - 11);
+	size_t pfxlen = strlen("lightning_");
+	assert(strlen(name) > pfxlen);
+	const char *short_name =
+		tal_strndup(ctx, name + pfxlen, strlen(name) - pfxlen);
```
rustyrussell: tal_strdup() perhaps?

ksedgwic: Fixed.

ksedgwic:
I think that was a spurious CI failure (timeout), force pushed to try again.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3372/commits/95b5bdb92ebb23f8d32cd4c91f6dd33d3d546214

PR #3374, opened 2019-12-24
plugin_control: more descriptive key for 'plugin stop' result

darosior:
I found the json output
```
{
   "": "Successfully stopped reckless.py."
}
```
a bit weird.


cdecker:
Might need a changelog entry given it changes the format of the JSON-RPC which is user-facing. It'd be better if we add the new key-value pair to the result and guard the old one with the `deprecated_apis` flag.

darosior:
Added both

cdecker:
ACK https://github.com/ElementsProject/lightning/commit/51a2607aa871a89892d54a4d0150d27d23d2ca04

PR #3375, opened 2019-12-24
"plugin start": spawn plugin processes with a non-0 umask

darosior:
Since they are currently spawned with a umask of 0 (set when daemonizing), plugins will create files with 666 and directory with 777 permissions. This might not be a big deal but `22` may be a more expected default for plugin developers ?

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -244,9 +244,14 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
 	ld->stop_conn = NULL;
 
 	/*~ This is used to signal that `hsm_secret` is encrypted, and will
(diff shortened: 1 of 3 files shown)


rustyrussell:
We should leave umask unchanged, from whatever it was at start?

darosior:
Ok, will make it a `lightningd` member then.

cdecker:
Needs a changelog entry, otherwise looks good to me 👍

darosior:
Modified to use initial umask instead, and added the Changelog

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3375/commits/ab28258fcd7886d35c2a14a96fbbf3a3678cbe5d

Issue #3378, opened 2019-12-27
[Bug] rpc-file moved upon migration


NicolasDorier:
So this one is weird.
I ran my tests on 0.8 => no problem.
But when I try deploying on mainnet, it seems the `rpc-file` parameter is completely ignored.

The only difference between my tests and the mainnet is that in mainnet, the data folder has been migrated from /root/.lightning/ to /root/.lightning/bitcoin, on regtest, there was no migration.

On mainnet my `/root/.lightning/config` is

```
bitcoin-datadir=/etc/bitcoin
bitcoin-rpcconnect=bitcoind

proxy=tor:9050

bind-addr=0.0.0.0:9735
network=bitcoin
announce-addr=nico2.temp.btcpayserver.org:9735
announce-addr=uewlcjmuuwcgs43lpcc7d64k2pf4xs7yyer7ztsq6cblyjxue3bc35qd.onion:9735
rpc-file=/root/.lightning/lightning-rpc
```

On regtest (where rpc-file works) it is

```
bitcoin-datadir=/etc/bitcoin
bitcoin-rpcconnect=bitcoind
network=regtest
announce-addr=lightningd
log-level=debug

bind-addr=0.0.0.0:9735
network=regtest
rpc-file=/root/.lightning/lightning-rpc
```

But on mainnet the rpc-file still show up in the bitcoin folder.

Do you see anything I would have missed?

Edit: in the log I saw:

```
2019-12-27T13:55:42.579Z UNUSUAL lightningd: Moving lightning-rpc into /root/.lightning/bitcoin/
```

During the migration. It does not appear again if I restart. But it should not have been moved anyway.

darosior:
> on regtest, there was no migration.

Weird, should be `.lightning/regtest/`.

> Do you see anything I would have missed?
> Edit: in the log I see:

In 0.8 the `lightning_dir` was separated into networks, you shouldn't be able to access the RPC socket at the root of the `lightning_dir` anymore.

NicolasDorier:
@darosior on regtest there is no problem. The lightning-rpc file is created in .lightning and the data is in /regtest. 

But on mainnet, the data is in /bitcoin as well as the lightning-rpc.

darosior:
You should prefer using `lightning-dir=` (superfluous as you seem to be using the default) and `network=`, I think.

darosior:
> But on mainnet, the data is in /bitcoin as well as the lightning-rpc.

[...]

PR #3379, opened 2019-12-27
Do not move the socket file during migration

NicolasDorier:
While no side effect, it confused me a lot on #3378

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -881,6 +881,7 @@ static void promote_missing_files(struct lightningd *ld)
 		if (streq(d->d_name, ".")
 		    || streq(d->d_name, "..")
 		    || streq(d->d_name, "config")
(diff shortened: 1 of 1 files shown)


rustyrussell:
Hmm, if we leave it there, unused, it will be very confusing for users, too.

NicolasDorier:
@rustyrussell so as far as I have tested right now here is what happen during migration if `rpc-file=.lightning/lightning-rpc` is set:
1 `.lightning/lightning-rpc` get moved to `bitcoin/lightning-rpc`
2 `.lightning/lightning-rpc` file get created, the socket listening to this one.

The file in `.lightning/bitcoin/lightning-rpc` is not responding, but exists.
The file on `.lightning/lightning-rpc` is the real socket.

rustyrussell:
You're right. We should delete the RPC file instead of moving it.

PR #3380, opened 2019-12-27
pytest: test lightning-rpc path after migration

darosior:
This is a formalisation of the behavior observed by @NicolasDorier in #3378. The `lightning-rpc` socket is not created in the same path for each network: hence one of this tests pass (`regtest`), not the other one (`testnet`).


darosior:
Ok so funily, it passes on Travis but not locally

NicolasDorier:
Hey I opened https://github.com/ElementsProject/lightning/pull/3379 as well.
Maybe it has been fixed on master already?

PR #3383, opened 2019-12-30
build: fix version extraction from non-git directory

vasild:
https://github.com/ElementsProject/lightning/releases/download/v0.8.0/clightning-v0.8.0.zip
extracts into "clightning-v0.8.0".

https://github.com/ElementsProject/lightning/archive/v0.8.0.tar.gz
extracts into "lightning-0.8.0".

Tweak the regular expression in the Makefile to support both (extract
0.8.0 from either one). Also, downloading an arbitrary version, based on
git commit id:

https://codeload.github.com/ElementsProject/lightning/tar.gz/eb2b225c?dummy=/filename_this_is_irrelevant.tar.gz
extracts into "lightning-eb2b225c".

so, support extracting "eb2b225c" from that.

Changelog-None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 #! /usr/bin/make
 
 # Extract version from git, or if we're from a zipfile, use dirname
-VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's,.*/clightning-\(v[0-9.rc]*\)$$,\1,p')
+VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c\{0,1\}lightning-v\{0,1\}\([0-9a-f.rc]*\)$$|\1|gp')
 
 ifeq ($(VERSION),)
 $(error "ERROR: git is required for generating version information")


Makefile
```diff
@@ -1,7 +1,7 @@
 #! /usr/bin/make
 
 # Extract version from git, or if we're from a zipfile, use dirname
-VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's,.*/clightning-\(v[0-9.rc]*\)$$,\1,p')
+VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c\{0,1\}lightning-v\{0,1\}\([0-9a-f.rc]*\)$$|\1|gp')
```
cdecker:
```suggestion
VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c?lightning-v?\([0-9a-f.rc]*\)$$|\1|gp')
```
`{0,1}` and `?` should be functionally identical, but `?` is shorter and easier to recognize imho

vasild: Yes, `?` is shorter and more readable, but it is only supported in "extended" regular expressions, not in "basic" ones. I chose to stick with "basic" REs because the switch to enable "extended" REs is different on Linux (`sed -r`) and FreeBSD (`sed -E`).

However, now I see that there has been [development](https://unix.stackexchange.com/questions/310446/are-there-standard-versions-of-sed-on-which-e-is-unsupported?noredirect=1) on this front recently (in 2013 actually) and GNU sed now supports `-E`, woho! The "extended" REs variant would look like:
```
sed -En 's|.*/c?lightning-v?([0-9a-f.rc]*)$$|\1|gp'
```
(double `$$` to escape the `$` in the Makefile).

Should I replace the basic RE or leave it as is? I guess some old Linux distros that ship with ancient sed will break with `-E`.

ZmnSCPxj: Suggest keeping with basic regular expressions.

cdecker: I wasn't aware of the distinction, thanks for clarifying :+1: Let's keep it as is :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3383/commits/846323f1264317f5c29384f2cefbab4f9dc3cd03

PR #3384, opened 2019-12-30
txprepare fixes

niftynei:
Fixes two crashes/bugs I found in `txprepare`:
 1) using `txprepare` in keywords mode in `pyln-client` would crash (fails an check for what version of the API you're using) 
 2) crash when attempting to specify more than two outputs


cdecker:
ACK https://github.com/ElementsProject/lightning/commit/7f59865e40f9f229fe120915b2473b406148ae5d

PR #3386, opened 2020-01-01
wallet: remove edgecase around transaction checks

niftynei:
we don't populate the tx item when we're running a transaction check from deep chain (prior to a chain replay). this change allows us to remove the null check on `tx` where we do verification on the funding tx.

I'm not sure this is a desirable change; these transactions have been seen and processed by c-lightning already, so the check is redundant. On the other hand, I think removing the possibility of skipping the check is worth the added lookup as it decreases the risk of new code changes inadvertently disabling it by failing to populate `tx`.

Changelog-None

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -990,7 +990,7 @@ static enum watch_result funding_depth_cb(struct lightningd *ld,
 	struct short_channel_id scid;
 
 	/* Sanity check */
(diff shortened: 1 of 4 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3386/commits/9cedf91cec4538f13244b87e3dec35a2164e8a71

Issue #3270, opened 2019-11-18
Setting manual channel closure fee (question)


bitcoin-software:
Reffering to this question: https://bitcoin.stackexchange.com/questions/91755/private-agreement-between-two-c-lightning-nodes-on-channel-closure-fee

> When i was at LNConf, one of the devs of c-lightning (IIRC) told me that there is a custom, manual way to agree between two nodes on channel closing fees.
> Can anyone elaborate how this is possible? I am operator of two cln nodes, i would like to close channel between these nodes with minimum possible fee, because i am not in hurry.
> Are there some side-effects of doing so?

>  I dont want to wait until mempool clears to perform tx. i just want to broadcast it and lay back, so it will be included, when mempool clears. so i dont need to monitor state of mempool

Would like to hear your thoughts.

rustyrussell:
When nodes mutual close, they negotiate iteratively until they agree.  By default we split the difference between what we want and what they want.  The `feerates perkb` gives you an indication of what fees we'll aim for in each case.

There's currently no way to force a lowball fee, though it would be easy to add one to the close command.  It's a negotiation though, so the other side needs to agree or we can't close.

whitslack:
> It's a negotiation though, so the other side needs to agree or we can't close.

@rustyrussell: Can the negotiation safely be abandoned (leaving the channel open) if the peer wants too high a fee? I can see a use case like "mutually close this channel if it can be done for less than 500 sats; otherwise, leave it open."

ZmnSCPxj:
It may require a spec change, as once the closing negotiation is initiated, we refuse to forward over that channel.  We would have to include something in the closing negotiation that says "never mind, cancel the closing negotiation" which would be sendable only by the peer that initiated the closing negotiation.  Code-wise we need infrastructure to transfer from `closingd` back to `channeld`.

It *still* does not quite do what you want, which is basically to ensure that a particular feerate is used.

ZmnSCPxj:
> There's currently no way to force a lowball fee, though it would be easy to add one to the close command. It's a negotiation though, so the other side needs to agree or we can't close.

More specifically

The spec requires that we move the fee we propose by at least 1 unit at each round of the negotiation.

What we currently do now is that we start with our fee then when the other node counter-proposes we average our last proposal and the other node last proposal, until roundoff makes us send the same feerate as the other node.

[...]

PR #3390, opened 2020-01-03
closingd: configurable closing fee negotiation step

vasild:
When negotiating the transaction fee for closing a channel [1], we used
to always pick the middle of the range between our proposal and the
peer's proposal.

Introduce a new option `fee_negotiation_step` to the close command, so
the peer who initiates the close can choose his back off step.

Partially resolves https://github.com/ElementsProject/lightning/issues/3270

[1] https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#closing-negotiation-closing_signed

Changelog-Added: New optional parameter to the `close` command to control the closing transaction fee negotiation back off step


vasild:
To complete this PR:

- [ ] Make this behavior configurable (replace the `if (true)`). Maybe add a global config option `--stubborn-closing-fee-negotiation`? What about setting this behavior per channel?
- [ ] Add tests - how much of this code is covered by existent tests? Maybe this change warrants adding more tests.
- [ ] Add a way to manually specify the starting fee for the negotiation. Maybe extend the close command:
```
close id [unilateraltimeout] [destination]
    Close the channel with {id} (either peer ID, channel ID, or short channel ID). Force a unilateral close after {unilateraltimeout} seconds (default 48h). If {destination} address is provided, will be used as output address.
```
with one more parameter `[fee]`, but that must be "less than or equal to the base fee of the final commitment transaction" so if the user requests some bigger fee we should reject his request without initiating a channel close. Also, what about specifying the fee if we are not the side that initiated the channel close?

ZmnSCPxj:
> Maybe add a global config option `--stubborn-closing-fee-negotiation`? What about setting this behavior per channel?

Would not an option to the `close` command be better?

> Also, what about specifying the fee if we are not the side that initiated the channel close?

Given that the closing fee is paid for by the side which opened the channel ("initiator pays" principle which closes some active griefing attacks), perhaps it should only matter for such channels where I opened but the other side wants to close?  In such a case perhaps it should be stubborn as well.

Detecting the stubborn closing behavior from the closing peer might also be an option to trigger stubborn closing behavior on our side when we are not the one initiating the close.

vasild:
@ZmnSCPxj yes, an option to the `close` command would be better, but it would be best if both sides can specify the negotiation strategy and the starting fee.

We have two new config options (per channel): 1. closing negotiation strategy (bisect or stubborn) and 2. closing negotiation starting fee.

What about this:
* When opening a channel one can specify 1. and 2. They are saved somewhere in the database and used during channel close. If the saved fee turns out to be too high during close, then it is lowered to "the base fee of the final commitment transaction". The default values, if not specified would be "strategy=bisect" and "fee=derive from bitcoind during close".
* During channel lifetime 1. and 2. can be updated by the user - in case he changed his mind or forgot to specify them during channel open or he was not the one who created the channel.
* It is possible to also specify 1. and 2. to the `close` command in which case the provided values override the ones from the database.

vasild:
@ZmnSCPxj, What would be the best way to pass an option of the close command to the fee negotiating code?

That is - adding a new option to the close command would mean that [json_close()](https://github.com/ElementsProject/lightning/blob/faac4b2/lightningd/peer_control.c#L1281) will receive it and then it needs to pass it somehow down the call chain, through the daemons towards [closingd](https://github.com/ElementsProject/lightning/blob/faac4b2/closingd/closingd.c#L585).

ZmnSCPxj:
Not sure, maybe add a field to `struct channel` in `lightningd` (which you would initialize from options, then override if the `close` command provides it)?  Then send it over to `closingd` in `peer_start_closingd`.  Note there are multiple different `struct channel` because hysterical raisins, make sure to add to the one in `lightning/channel.h`. @rustyrussell @cdecker @niftynei opinions?

vasild:
Yes, I did exactly that after some staring at the code! Need to clean up the patch and will push it here.

I am not yet sure what would happen if lightningd gets restarted during the close fee negotiation. I guess the newly added option telling it how to negotiate would be lost.

vasild:
> Need to clean up the patch and will push it here.

Done, ready for review.

The first commit (mocks re-record) belongs to a separate PR, but then this PR would depend on it. Should I create a separate PR for it?

contrib/pyln-client/pyln/client/__init__.py
```diff
@@ -2,7 +2,7 @@
 from .plugin import Plugin, monkey_patch
 
 
-__version__ = "0.8.0"
+__version__ = "0.9.0"
```
darosior: Hmm why did you bump the version ?

vasild: Hmm, good question. Because my understanding was that the tests used `pylightning` which used `pyln-client` which I extended with a new argument to `close()` and so `pylightning` and tests would not work with an earlier version of `pyln-client`, so I bumped the version with the intent to require `pyln-client>=0.9.0` in `pylightning`.

But now I double checked the relationship and I see that tests use `pyln-client` [directly](https://github.com/ElementsProject/lightning/blob/15ca3b6/Makefile#L275) and that `pylightning` is an independent package seemingly not used by anybody.

Reverted the above change and the change to `contrib/pylightning/requirements.txt` which was supposed to require the new version with the extended `close()`.

darosior: Yeah we explicitly set the `PYTHONPATH` to test against the newest `pyln`, and bump the version at each release.

vasild:
> I find it really cool to be able to greed on closing fees, could it be made to directly set the step for even more customability ?

Yes, this would be even more flexible.

So we have an interval, lets say `[x, y]` and our next closing fee proposal must be a number from that interval. We also know if we started higher than the peer or lower - that is, if we would prefer to be closer to `y` or `x`.

The current code chooses `(x + y) / 2`.

This PR adds an option to pick `x` (if we started lower, ie prefer lower fee than the peer) or `y` (if we started higher, ie prefer higher fee than the peer).

What about instead of adding a close parameter `fee_negotiation_strategy=(bisect|stubborn)` to add `fee_negotiation_step=(N|N%)` which would give up "N satoshi" or "N% from the interval" on each negotiation step? Then
* `fee_negotiation_step=50%` would be the default and equivalent to the current behavior (bisect)
* `fee_negotiation_step=1` would be equivalent to "stubborn" negotiation strategy where we give up just 1 satoshi at a time
* `fee_negotiation_step=100%` would be to just accept peer's proposal and finish with negotiation as quickly as possible.

closingd/closingd.c
```diff
@@ -519,18 +527,35 @@ static struct amount_sat adjust_offer(struct per_peer_state *pps,
...
-	/* Bisect between our minimum and max. */
-	if (amount_sat_greater(feerange->min, min_fee_to_accept))
-		min_fee_to_accept = feerange->min;
+	if (fee_negotiation_step_unit ==
+	    CLOSING_FEE_NEGOTIATION_STEP_UNIT_SATOSHI) {
+		/* -1 because the range boundary has already been adjusted with
+		 * one from our previous proposal. So, if the user requested a
+		 * step of 1 satoshi at a time we should just return our end of
+		 * the range from this function. */
+		step = (struct amount_sat){fee_negotiation_step - 1};
```
darosior: `amount_sat_sub()` ? :)

vasild: Do you mean to add a new function `amount_sat_from_u64()`? Are there other places in the code that would benefit from it?

darosior: Ok I confused with `msat_from_u64()`... https://github.com/ElementsProject/lightning/blob/e99ab03f79bf71ab313c5b3e1d8446aa3fa45044/common/amount.c#L375

tests/test_closing.py
```diff
@@ -451,26 +435,91 @@ def get_fee_from_status(node, peer_id, i):
...
-        'fundee_feerate_per_kw': 27625,
-        'close_initiated_by': 'funder',
-        'expected_close_fee': 33533 if chainparams['elements'] else 20333
+        'fundee_feerate_per_kw': 27625
     }
 
+    opts['fee_negotiation_step'] = '50%'
+
+    opts['close_initiated_by'] = 'funder'
+    opts['expected_close_fee'] = 101 if chainparams['elements'] else 20334
```
vasild: Here and below the fees for "elements" need to be checked and recorded. But how to run the test for that chain?

closingd/closingd.c
vasild: That can still be used, but I think would be too cumbersome:

```c
struct amount_msat step_msat;
...
amount_msat_from_u64(&step_msat, (fee_negotiation_step - 1) * MSAT_PER_SAT);
step = amount_msat_to_sat_round_down(step_msat);
```

vasild: Done that way with `amount_msat_from_u64()` because the typecast was upsetting `make check-amount-access`.

darosior: I think you are still better off adding the helper for `sat`, but it can be done in a follow up cleanup, I think?

tests/test_closing.py
darosior: You need to set `TEST_NETWORK=liquid-regtest` (and obv have the elements daemon available in your path).

tests/test_closing.py
```diff
@@ -432,45 +454,82 @@ def get_mempool_when_size_1():
...
+    assert opts['expected_close_fee'] == fee_mempool
 
-    status_agreed_regex = re.compile("agreed on a closing fee of ([0-9]+) satoshi")
 
-    # [fee_from_funder_status, fee_from_fundee_status]
-    fees_from_status = [None, None]
+opts = {
+    'funder_feerate_per_kw': 29006,  # closing fee negotiation starts at 21000
+    'fundee_feerate_per_kw': 27625   # closing fee negotiation starts at 20000
+}
```
darosior: Using a global isn't really nice (especially since we make an extensive use of the `opts` name in the functional tests). Note that it could be made a function which doesn't begin with `test_` (like `closing_fee_options(initiator, expected_fees)`) so that it doesn't get treated as a test by `pytest`.

tests/test_closing.py
vasild: Thanks, updated.

tests/test_closing.py
vasild: I agree the global was bad. Moved into the generic test function that's called from the other `test_*()` functions.

vasild:
Rebased due to conflicts with master.

closingd/closingd.c
```diff
@@ -609,13 +649,21 @@ int main(int argc, char *argv[])
 	/* stdin == requests, 3 == peer, 4 = gossip, 5 = gossip_store, 6 = hsmd */
 	per_peer_state_set_fds(pps, 3, 4, 5);
 
+	snprintf(fee_negotiation_step_str, sizeof(fee_negotiation_step_str),
+		 "%" PRIu64 "%s", fee_negotiation_step,
+		 fee_negotiation_step_unit ==
+			 CLOSING_FEE_NEGOTIATION_STEP_UNIT_PERCENTAGE
+		     ? "%"
+		     : "sat");
+
```
rustyrussell, line 658: tal_fmt() is generally safer than snprintf, BTW.

doc/lightning-close.7.md
```diff
@@ -28,6 +28,22 @@ The default is 2 days (172800 seconds).
 The *destination* can be of any Bitcoin accepted type, including bech32.
 If it isn't specified, the default is a c-lightning wallet address.
 
+The *fee_negotiation_step* parameter controls how closing fee
+negotiation is performed assuming the peer proposes a fee that is
+different than our estimate. On every negotiation step we must give up
+some amount from our proposal towards the peer's proposal. This parameter
+can be an integer in which case it is interpreted as number of satoshis
+to step at a time. Or it can be an integer followed by "%s" to designate
```
rustyrussell, line 36: "%s"?  Not sure where the 's' came from?

lightningd/peer_control.c
```diff
@@ -1285,12 +1288,16 @@ static struct command_result *json_close(struct command *cmd,
...
+	const char *fee_negotiation_step_str;
+	char* end;
 
 	if (!param(cmd, buffer, params,
 		   p_req("id", param_tok, &idtok),
 		   p_opt_def("unilateraltimeout", param_number, &timeout,
 			     48 * 3600),
 		   p_opt("destination", param_bitcoin_address, &close_to_script),
+		   p_opt("fee_negotiation_step", param_string,
+			 &fee_negotiation_step_str),
```
rustyrussell, line 1300: Generally better to write a param handler for this: that way "check" calls it and does more checking for the caller.

closingd/closingd.c
vasild: It was using `tal_fmt()` in a [previous incarnation](https://github.com/ElementsProject/lightning/commit/662c351a826fe01d91c92df1e4bbc86665871d50#diff-9d466ee4dfc90b4d2dbf7958605b7d4aR642) of this PR. However, that ended up with an obscure use-after-free:

* create the string with `tal_fmt()` from `tmpctx`
* print the string with `status_debug()`
* call `do_reconnect()` which calls `closing_read_peer_msg()` which calls `clean_tmpctx()` which frees `tmpctx`
* try to print the string with `peer_billboard()`, but it is already freed

doc/lightning-close.7.md
vasild: [Fixed in a followup PR](https://github.com/ElementsProject/lightning/pull/3629/commits/6502e643b344ea0dda4a8bc2c8e4b3f4bc422ec9). I must have been in C-mode-autopilot when writing the .md file.

lightningd/peer_control.c
vasild: Something like this:

<details>
<summary>patch</summary>

```diff
commit ef31b2b4
Parent: 48fa053f
Author:     Vasil Dimov <vd@FreeBSD.org>
AuthorDate: Tue Apr 7 12:03:24 2020 +0200
Commit:     Vasil Dimov <vd@FreeBSD.org>
CommitDate: Tue Apr 7 12:04:43 2020 +0200
gpg: Signature made Tue Apr  7 12:04:43 2020 CEST
gpg:                using RSA key E64D8D45614DB07545D9CCC154DF06F64B55CBBF
gpg: Good signature from "Vasil Dimov <vd@myforest.net>" [ultimate]
gpg:                 aka "Vasil Dimov <vd@FreeBSD.org>" [ultimate]
gpg:                 aka "Vasil Dimov <vasild@gmail.com>" [ultimate]


    json: add JSON handler for num or num%
    
    Add a generic handler that accepts either a bare number or a bare number
    followed by "%". Use that one when taking the "close fee negotiation
    step" argument of the close command.
    
    Changelog-None

diff --git a/common/json.c b/common/json.c
index b5af5e06..e1671622 100644
--- a/common/json.c
+++ b/common/json.c
@@ -76,12 +76,27 @@ bool json_to_u64(const char *buffer, const jsmntok_t *tok,
 	if (*num != l)
 		return false;
 
 	return true;
 }
 
+bool json_to_u64_or_u64pct(const char *buffer, const jsmntok_t *tok,
+			   struct u64_or_u64pct *result)
+{
+	jsmntok_t tok_copy = *tok;
+
+	if (buffer[tok_copy.end - 1] == '%') {
+		tok_copy.end--;
+		result->is_pct = true;
+	} else {
+		result->is_pct = false;
+	}
+
+	return json_to_u64(buffer, &tok_copy, &result->n);
+}
+
 bool json_to_s64(const char *buffer, const jsmntok_t *tok, s64 *num)
 {
 	char *end;
 	long long l;
 
 	l = strtoll(buffer + tok->start, &end, 0);
diff --git a/common/json.h b/common/json.h
index f92a634d..10c33e51 100644
--- a/common/json.h
+++ b/common/json.h
@@ -1,12 +1,13 @@
 #ifndef LIGHTNING_COMMON_JSON_H
 #define LIGHTNING_COMMON_JSON_H
 #include "config.h"
 #include <ccan/short_types/short_types.h>
 #include <ccan/tal/tal.h>
 #include <common/errcode.h>
+#include <common/utils.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdlib.h>
 
 #define JSMN_STRICT 1
 # include <external/jsmn/jsmn.h>
@@ -55,12 +56,17 @@ bool json_to_number(const char *buffer, const jsmntok_t *tok,
 		    unsigned int *num);
 
 /* Extract number from this (may be a string, or a number literal) */
 bool json_to_u64(const char *buffer, const jsmntok_t *tok,
 		 uint64_t *num);
 
+/* Extract a non-negative integer or an integer percentage (an integer in
+ * [0, 100] followed by "%"). */
+bool json_to_u64_or_u64pct(const char *buffer, const jsmntok_t *tok,
+			   struct u64_or_u64pct *result);
+
 /* Extract signed 64 bit integer from this (may be a string, or a number literal) */
 bool json_to_s64(const char *buffer, const jsmntok_t *tok, s64 *num);
 
 /* Extract number from this (may be a string, or a number literal) */
 bool json_to_u32(const char *buffer, const jsmntok_t *tok,
 		 uint32_t *num);
diff --git a/common/json_tok.c b/common/json_tok.c
index a08de9e0..e33f933d 100644
--- a/common/json_tok.c
+++ b/common/json_tok.c
@@ -5,12 +5,13 @@
 #include <common/amount.h>
 #include <common/json_command.h>
 #include <common/json_helpers.h>
 #include <common/json_tok.h>
 #include <common/jsonrpc_errors.h>
 #include <common/param.h>
+#include <common/utils.h>
 
 struct command_result *param_array(struct command *cmd, const char *name,
 				   const char *buffer, const jsmntok_t *tok,
 				   const jsmntok_t **arr)
 {
 	if (tok->type == JSMN_ARRAY) {
@@ -139,12 +140,28 @@ struct command_result *param_u64(struct command *cmd, const char *name,
 
 	return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 			    "'%s' should be an unsigned 64 bit integer, not '%.*s'",
 			    name, tok->end - tok->start, buffer + tok->start);
 }
 
+struct command_result *param_u64_or_u64pct(struct command *cmd,
+					   const char *name, const char *buffer,
+					   const jsmntok_t *tok,
+					   struct u64_or_u64pct **result)
+{
+	*result = tal(cmd, struct u64_or_u64pct);
+
+	if (json_to_u64_or_u64pct(buffer, tok, *result))
+		return NULL;
+
+	return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+			    "'%s' should be an unsigned 64 bit integer "
+			    "optionally followed by %%, not '%.*s'",
+			    name, tok->end - tok->start, buffer + tok->start);
+}
+
 struct command_result *param_tok(struct command *cmd, const char *name,
 				 const char *buffer, const jsmntok_t * tok,
 				 const jsmntok_t **out)
 {
 	*out = tok;
 	return NULL;
diff --git a/common/json_tok.h b/common/json_tok.h
index 038015a8..ef811b37 100644
--- a/common/json_tok.h
+++ b/common/json_tok.h
@@ -3,12 +3,13 @@
 #define LIGHTNING_COMMON_JSON_TOK_H
 #include "config.h"
 #include <ccan/short_types/short_types.h>
 #include <common/json.h>
 #include <common/node_id.h>
 #include <common/sphinx.h>
+#include <common/utils.h>
 #include <wire/wire.h>
 
 struct amount_msat;
 struct amount_sat;
 struct command;
 struct command_result;
@@ -63,12 +64,19 @@ struct command_result *param_sha256(struct command *cmd, const char *name,
 
 /* Extract number from this (may be a string, or a number literal) */
 struct command_result *param_u64(struct command *cmd, const char *name,
 				 const char *buffer, const jsmntok_t *tok,
 				 uint64_t **num);
 
+/* Extract a non-negative integer or an integer percentage (an integer in
+ * [0, 100] followed by "%"). */
+struct command_result *param_u64_or_u64pct(struct command *cmd,
+					   const char *name, const char *buffer,
+					   const jsmntok_t *tok,
+					   struct u64_or_u64pct **result);
+
 /* Extract msatoshi amount from this string */
 struct command_result *param_msat(struct command *cmd, const char *name,
 				  const char *buffer, const jsmntok_t *tok,
 				  struct amount_msat **msat);
 
 /* Extract satoshi amount from this string */
diff --git a/common/utils.h b/common/utils.h
index 484da8ea..0510c952 100644
--- a/common/utils.h
+++ b/common/utils.h
@@ -5,12 +5,17 @@
 #include <ccan/crypto/sha256/sha256.h>
 #include <ccan/short_types/short_types.h>
 #include <ccan/structeq/structeq.h>
 #include <ccan/tal/tal.h>
 #include <secp256k1.h>
 
+struct u64_or_u64pct {
+	u64 n;
+	bool is_pct;
+};
+
 extern secp256k1_context *secp256k1_ctx;
 
 extern const struct chainparams *chainparams;
 
 /* Simple accessor function for our own dependencies to use, in order to avoid
  * circular dependencies (should only be used in `bitcoin/y`). */
diff --git a/lightningd/peer_control.c b/lightningd/peer_control.c
index ae7d4ae5..619b93a0 100644
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1285,22 +1285,24 @@ static struct command_result *json_close(struct command *cmd,
 	struct channel *channel COMPILER_WANTS_INIT("gcc 7.3.0 fails, 8.3 OK");
 	unsigned int *timeout = NULL;
 	bool force = true;
 	bool do_timeout;
 	const u8 *close_to_script = NULL;
 	bool close_script_set;
-	const char *fee_negotiation_step_str;
-	char* end;
+	struct u64_or_u64pct *fee_negotiation_step;
+	struct u64_or_u64pct fee_negotiation_step_default = {.n = 50,
+							     .is_pct = true};
 
 	if (!param(cmd, buffer, params,
 		   p_req("id", param_tok, &idtok),
 		   p_opt_def("unilateraltimeout", param_number, &timeout,
 			     48 * 3600),
 		   p_opt("destination", param_bitcoin_address, &close_to_script),
-		   p_opt("fee_negotiation_step", param_string,
-			 &fee_negotiation_step_str),
+		   p_opt_def("fee_negotiation_step", param_u64_or_u64pct,
+			     &fee_negotiation_step,
+			     fee_negotiation_step_default),
 		   NULL))
 		return command_param_failed();
 
 	do_timeout = (*timeout != 0);
 
 	peer = peer_from_json(cmd->ld, buffer, idtok);
@@ -1360,48 +1362,34 @@ static struct command_result *json_close(struct command *cmd,
 			= p2wpkh_for_keyidx(channel, cmd->ld, channel->final_key_idx);
 		/* We don't save the default to disk */
 		close_script_set = false;
 	} else
 		close_script_set = false;
 
-	if (fee_negotiation_step_str == NULL) {
-		channel->closing_fee_negotiation_step = 50;
-		channel->closing_fee_negotiation_step_unit =
-		    CLOSING_FEE_NEGOTIATION_STEP_UNIT_PERCENTAGE;
-	} else {
-		channel->closing_fee_negotiation_step =
-		    strtoull(fee_negotiation_step_str, &end, 10);
-
-		if (channel->closing_fee_negotiation_step == 0)
-			return command_fail(
-			    cmd, JSONRPC2_INVALID_PARAMS,
-			    "Wrong value given for fee_negotiation_step: "
-			    "\"%s\", must be positive",
-			    fee_negotiation_step_str);
-		else if (*end == '%') {
-			if (channel->closing_fee_negotiation_step > 100)
-				return command_fail(
-				    cmd, JSONRPC2_INVALID_PARAMS,
-				    "Wrong value given for "
-				    "fee_negotiation_step: \"%s\", the "
-				    "percentage should be between 1 and 100",
-				    fee_negotiation_step_str);
-			channel->closing_fee_negotiation_step_unit =
-			    CLOSING_FEE_NEGOTIATION_STEP_UNIT_PERCENTAGE;
-		} else if (*end == '\0')
-			channel->closing_fee_negotiation_step_unit =
-			    CLOSING_FEE_NEGOTIATION_STEP_UNIT_SATOSHI;
-		else
-			return command_fail(
-			    cmd, JSONRPC2_INVALID_PARAMS,
-			    "Wrong value given for fee_negotiation_step: "
-			    "\"%s\", should be an integer or an integer "
-			    "followed by %%",
-			    fee_negotiation_step_str);
+	if (fee_negotiation_step->n == 0) {
+		return command_fail(
+		    cmd, JSONRPC2_INVALID_PARAMS,
+		    "Wrong value given for fee_negotiation_step: \"%" PRIu64
+		    "\", it should be a positive number",
+		    fee_negotiation_step->n);
+	}
+
+	if (fee_negotiation_step->is_pct && fee_negotiation_step->n > 100) {
+		return command_fail(
+		    cmd, JSONRPC2_INVALID_PARAMS,
+		    "Wrong value given for fee_negotiation_step: \"%" PRIu64
+		    "\", the percentage should be between 1 and 100",
+		    fee_negotiation_step->n);
 	}
 
+	channel->closing_fee_negotiation_step = fee_negotiation_step->n;
+	channel->closing_fee_negotiation_step_unit =
+	    fee_negotiation_step->is_pct
+		? CLOSING_FEE_NEGOTIATION_STEP_UNIT_PERCENTAGE
+		: CLOSING_FEE_NEGOTIATION_STEP_UNIT_SATOSHI;
+
 	/* Normal case.
 	 * We allow states shutting down and sigexchange; a previous
 	 * close command may have timed out, and this current command
 	 * will continue waiting for the effects of the previous
 	 * close command. */
 
diff --git a/common/test/run-sphinx.c b/common/test/run-sphinx.c
index f73bc74e..a75dae8e 100644
--- a/common/test/run-sphinx.c
+++ b/common/test/run-sphinx.c
@@ -44,15 +44,12 @@ bool amount_sat_eq(struct amount_sat a UNNEEDED, struct amount_sat b UNNEEDED)
 /* Generated stub for bigsize_get */
 size_t bigsize_get(const u8 *p UNNEEDED, size_t max UNNEEDED, bigsize_t *val UNNEEDED)
 { fprintf(stderr, "bigsize_get called!\n"); abort(); }
 /* Generated stub for bigsize_put */
 size_t bigsize_put(u8 buf[BIGSIZE_MAX_LEN] UNNEEDED, bigsize_t v UNNEEDED)
 { fprintf(stderr, "bigsize_put called!\n"); abort(); }
-/* Generated stub for ecdh */
-void ecdh(const struct pubkey *point UNNEEDED, struct secret *ss UNNEEDED)
-{ fprintf(stderr, "ecdh called!\n"); abort(); }
 /* Generated stub for pubkey_from_node_id */
 bool pubkey_from_node_id(struct pubkey *key UNNEEDED, const struct node_id *id UNNEEDED)
 { fprintf(stderr, "pubkey_from_node_id called!\n"); abort(); }
 /* AUTOGENERATED MOCKS END */
 
 extern secp256k1_context *secp256k1_ctx;
diff --git a/connectd/test/run-responder-success.c b/connectd/test/run-responder-success.c
index cd9410cb..5657d178 100644
--- a/connectd/test/run-responder-success.c
+++ b/connectd/test/run-responder-success.c
@@ -37,15 +37,12 @@ const void *fromwire_fail(const u8 **cursor UNNEEDED, size_t *max UNNEEDED)
 u8 *fromwire_tal_arrn(const tal_t *ctx UNNEEDED,
 		       const u8 **cursor UNNEEDED, size_t *max UNNEEDED, size_t num UNNEEDED)
 { fprintf(stderr, "fromwire_tal_arrn called!\n"); abort(); }
 /* Generated stub for fromwire_u16 */
 u16 fromwire_u16(const u8 **cursor UNNEEDED, size_t *max UNNEEDED)
 { fprintf(stderr, "fromwire_u16 called!\n"); abort(); }
-/* Generated stub for notleak_ */
-void *notleak_(const void *ptr UNNEEDED, bool plus_children UNNEEDED)
-{ fprintf(stderr, "notleak_ called!\n"); abort(); }
 /* Generated stub for towire_u16 */
 void towire_u16(u8 **pptr UNNEEDED, u16 v UNNEEDED)
 { fprintf(stderr, "towire_u16 called!\n"); abort(); }
 /* Generated stub for towire_u8_array */
 void towire_u8_array(u8 **pptr UNNEEDED, const u8 *arr UNNEEDED, size_t num UNNEEDED)
 { fprintf(stderr, "towire_u8_array called!\n"); abort(); }
diff --git a/lightningd/test/run-invoice-select-inchan.c b/lightningd/test/run-invoice-select-inchan.c
index 2392f76f..6f054ae9 100644
--- a/lightningd/test/run-invoice-select-inchan.c
+++ b/lightningd/test/run-invoice-select-inchan.c
@@ -20,13 +20,13 @@ void bitcoind_getutxout_(struct bitcoind *bitcoind UNNEEDED,
 				    const struct bitcoin_tx_output *txout UNNEEDED,
 				    void *arg) UNNEEDED,
 			 void *arg UNNEEDED)
 { fprintf(stderr, "bitcoind_getutxout_ called!\n"); abort(); }
 /* Generated stub for bolt11_decode */
 struct bolt11 *bolt11_decode(const tal_t *ctx UNNEEDED, const char *str UNNEEDED,
-			     const struct feature_set *fset UNNEEDED,
+			     const struct feature_set *our_features UNNEEDED,
 			     const char *description UNNEEDED, char **fail UNNEEDED)
 { fprintf(stderr, "bolt11_decode called!\n"); abort(); }
 /* Generated stub for bolt11_encode_ */
 char *bolt11_encode_(const tal_t *ctx UNNEEDED,
 		     const struct bolt11 *b11 UNNEEDED, bool n_field UNNEEDED,
 		     bool (*sign)(const u5 *u5bytes UNNEEDED,
@@ -45,15 +45,12 @@ void broadcast_tx(struct chain_topology *topo UNNEEDED,
 /* Generated stub for channel_tell_depth */
 bool channel_tell_depth(struct lightningd *ld UNNEEDED,
 				 struct channel *channel UNNEEDED,
 				 const struct bitcoin_txid *txid UNNEEDED,
 				 u32 depth UNNEEDED)
 { fprintf(stderr, "channel_tell_depth called!\n"); abort(); }
-/* Generated stub for command_check_only */
-bool command_check_only(const struct command *cmd UNNEEDED)
-{ fprintf(stderr, "command_check_only called!\n"); abort(); }
 /* Generated stub for command_fail */
 struct command_result *command_fail(struct command *cmd UNNEEDED, errcode_t code UNNEEDED,
 				    const char *fmt UNNEEDED, ...)
 
 { fprintf(stderr, "command_fail called!\n"); abort(); }
 /* Generated stub for command_failed */
@@ -100,18 +97,12 @@ const u8 *failmsg_incorrect_or_unknown(const tal_t *ctx UNNEEDED,
 /* Generated stub for fatal */
 void   fatal(const char *fmt UNNEEDED, ...)
 { fprintf(stderr, "fatal called!\n"); abort(); }
 /* Generated stub for feature_is_set */
 bool feature_is_set(const u8 *features UNNEEDED, size_t bit UNNEEDED)
 { fprintf(stderr, "feature_is_set called!\n"); abort(); }
-/* Generated stub for featurebits_or */
-u8 *featurebits_or(const tal_t *ctx UNNEEDED, const u8 *f1 TAKES UNNEEDED, const u8 *f2 TAKES UNNEEDED)
-{ fprintf(stderr, "featurebits_or called!\n"); abort(); }
-/* Generated stub for fixup_htlcs_out */
-void fixup_htlcs_out(struct lightningd *ld UNNEEDED)
-{ fprintf(stderr, "fixup_htlcs_out called!\n"); abort(); }
 /* Generated stub for fmt_wireaddr_without_port */
 char *fmt_wireaddr_without_port(const tal_t *ctx UNNEEDED, const struct wireaddr *a UNNEEDED)
 { fprintf(stderr, "fmt_wireaddr_without_port called!\n"); abort(); }
 /* Generated stub for fromwire_channel_dev_memleak_reply */
 bool fromwire_channel_dev_memleak_reply(const void *p UNNEEDED, bool *leak UNNEEDED)
 { fprintf(stderr, "fromwire_channel_dev_memleak_reply called!\n"); abort(); }
@@ -348,12 +339,18 @@ struct command_result *param_tok(struct command *cmd UNNEEDED, const char *name
 { fprintf(stderr, "param_tok called!\n"); abort(); }
 /* Generated stub for param_u64 */
 struct command_result *param_u64(struct command *cmd UNNEEDED, const char *name UNNEEDED,
 				 const char *buffer UNNEEDED, const jsmntok_t *tok UNNEEDED,
 				 uint64_t **num UNNEEDED)
 { fprintf(stderr, "param_u64 called!\n"); abort(); }
+/* Generated stub for param_u64_or_u64pct */
+struct command_result *param_u64_or_u64pct(struct command *cmd UNNEEDED,
+					   const char *name UNNEEDED, const char *buffer UNNEEDED,
+					   const jsmntok_t *tok UNNEEDED,
+					   struct u64_or_u64pct **result UNNEEDED)
+{ fprintf(stderr, "param_u64_or_u64pct called!\n"); abort(); }
 /* Generated stub for peer_get_owning_subd */
 struct subd *peer_get_owning_subd(struct peer *peer UNNEEDED)
 { fprintf(stderr, "peer_get_owning_subd called!\n"); abort(); }
 /* Generated stub for peer_memleak_done */
 void peer_memleak_done(struct command *cmd UNNEEDED, struct subd *leaker UNNEEDED)
 { fprintf(stderr, "peer_memleak_done called!\n"); abort(); }
diff --git a/wallet/test/run-wallet.c b/wallet/test/run-wallet.c
index 5d71b7be..62d838b8 100644
--- a/wallet/test/run-wallet.c
+++ b/wallet/test/run-wallet.c
@@ -55,15 +55,12 @@ void broadcast_tx(struct chain_topology *topo UNNEEDED,
 /* Generated stub for channel_tell_depth */
 bool channel_tell_depth(struct lightningd *ld UNNEEDED,
 				 struct channel *channel UNNEEDED,
 				 const struct bitcoin_txid *txid UNNEEDED,
 				 u32 depth UNNEEDED)
 { fprintf(stderr, "channel_tell_depth called!\n"); abort(); }
-/* Generated stub for command_check_only */
-bool command_check_only(const struct command *cmd UNNEEDED)
-{ fprintf(stderr, "command_check_only called!\n"); abort(); }
 /* Generated stub for command_fail */
 struct command_result *command_fail(struct command *cmd UNNEEDED, errcode_t code UNNEEDED,
 				    const char *fmt UNNEEDED, ...)
 
 { fprintf(stderr, "command_fail called!\n"); abort(); }
 /* Generated stub for command_param_failed */
@@ -339,21 +336,12 @@ void json_object_start(struct json_stream *ks UNNEEDED, const char *fieldname UN
 /* Generated stub for json_strdup */
 char *json_strdup(const tal_t *ctx UNNEEDED, const char *buffer UNNEEDED, const jsmntok_t *tok UNNEEDED)
 { fprintf(stderr, "json_strdup called!\n"); abort(); }
 /* Generated stub for json_stream_success */
 struct json_stream *json_stream_success(struct command *cmd UNNEEDED)
 { fprintf(stderr, "json_stream_success called!\n"); abort(); }
-/* Generated stub for json_to_address_scriptpubkey */
-enum address_parse_result json_to_address_scriptpubkey(const tal_t *ctx UNNEEDED,
-			     const struct chainparams *chainparams UNNEEDED,
-			     const char *buffer UNNEEDED,
-			     const jsmntok_t *tok UNNEEDED, const u8 **scriptpubkey UNNEEDED)
-{ fprintf(stderr, "json_to_address_scriptpubkey called!\n"); abort(); }
-/* Generated stub for json_to_bool */
-bool json_to_bool(const char *buffer UNNEEDED, const jsmntok_t *tok UNNEEDED, bool *b UNNEEDED)
-{ fprintf(stderr, "json_to_bool called!\n"); abort(); }
 /* Generated stub for json_to_node_id */
 bool json_to_node_id(const char *buffer UNNEEDED, const jsmntok_t *tok UNNEEDED,
 			       struct node_id *id UNNEEDED)
 { fprintf(stderr, "json_to_node_id called!\n"); abort(); }
 /* Generated stub for json_to_number */
 bool json_to_number(const char *buffer UNNEEDED, const jsmntok_t *tok UNNEEDED,
@@ -489,22 +477,23 @@ struct command_result *param_number(struct command *cmd UNNEEDED, const char *na
 struct command_result *param_short_channel_id(struct command *cmd UNNEEDED,
 					      const char *name UNNEEDED,
 					      const char *buffer UNNEEDED,
 					      const jsmntok_t *tok UNNEEDED,
 					      struct short_channel_id **scid UNNEEDED)
 { fprintf(stderr, "param_short_channel_id called!\n"); abort(); }
-/* Generated stub for param_string */
-struct command_result *param_string(struct command *cmd UNNEEDED, const char *name UNNEEDED,
-				    const char * buffer UNNEEDED, const jsmntok_t *tok UNNEEDED,
-				    const char **str UNNEEDED)
-{ fprintf(stderr, "param_string called!\n"); abort(); }
 /* Generated stub for param_tok */
 struct command_result *param_tok(struct command *cmd UNNEEDED, const char *name UNNEEDED,
 				 const char *buffer UNNEEDED, const jsmntok_t * tok UNNEEDED,
 				 const jsmntok_t **out UNNEEDED)
 { fprintf(stderr, "param_tok called!\n"); abort(); }
+/* Generated stub for param_u64_or_u64pct */
+struct command_result *param_u64_or_u64pct(struct command *cmd UNNEEDED,
+					   const char *name UNNEEDED, const char *buffer UNNEEDED,
+					   const jsmntok_t *tok UNNEEDED,
+					   struct u64_or_u64pct **result UNNEEDED)
+{ fprintf(stderr, "param_u64_or_u64pct called!\n"); abort(); }
 /* Generated stub for parse_onionpacket */
 enum onion_type parse_onionpacket(const u8 *src UNNEEDED,
 				  const size_t srclen UNNEEDED,
 				  struct onionpacket *dest UNNEEDED)
 { fprintf(stderr, "parse_onionpacket called!\n"); abort(); }
 /* Generated stub for payment_failed */
```
</details>

If that does not look like too much code, then I can append it to https://github.com/ElementsProject/lightning/pull/3629.

Issue #3366, opened 2019-12-20
[Feature request] Proper error code when peer is not connected


NicolasDorier:
### Issue and Steps to Reproduce

At every release, lightningd is breaking clients due to not using proper error code when funding a channel to a peer that is not connected.

### Workaround

Testing on several possible different error message thrown at each version of clightning.

https://github.com/btcpayserver/BTCPayServer.Lightning/commit/f78bd5b6d5a598646d28503be8a363bde985acd3#diff-3aa3aa916b8d25b5d374f2953506b989L357-R362

cdecker:
Would checking the error code not be more sensible? We can definitely assign a
more unique numeric code to "Not connected", "No address" and "Unknown peer",
and you could better match those and take the appropriate actions (looking up
addresses, reporting the details up to the user, etc).

The errors really are distinct as they report various different causes that
led the `fundchannel` to fail, and we can't just merge them into a "something
went wrong" error.

fwiw the "Unable to connect" error in response to `fundchannel` is new only
because we introduced an autoconnect feature in v0.8.0, since users reported
that it was strange to have a peer in the list, know its address, and don't
just connect in case we aren't.

NicolasDorier:
@cdecker anything I can compare other than an error message would make sense.

vasild:
This should help https://github.com/ElementsProject/lightning/pull/3395?

The error "Unable to connect, no address known for peer" (https://github.com/ElementsProject/lightning/blob/f6ff5e5/connectd/connectd.c#L1431) is not returned during funding, but during connect. How to add a code for it? `connect_failed()` is only sending a textual error message to the master daemon.

ZmnSCPxj:
> How to add a code for it? 

You will need to modify the message from connectd to lightningd, in file `connectd/connect_wire.csv`, to add an error code from connectd for the `connect_failed` message.  This changes all `fromwire_connectctl_connect_failed` and `towire_connectctl_connect_failed` calls (there should just be one of each, `fromwire` in `lightningd/connect_control.c` and `towire` somewhere in `connectd/`).

The type to use for the error code is a bit iffy though.  You could cast to a `u64` for the message but that is bad style; unfortunately it seems there is not yet any use for an ordinary `int` for messaging yet.  You could probably just create some kind of `fromwire_int` and `towire_int` in `wire/fromwire.c` and `wire/towire.c`, and do some magic in `tools/generate-wire.py` (add it to `assignables` array I think).

[...]

PR #3395, opened 2020-01-04
common: add 2 specific error codes wrt funding

vasild:
Add "peer not connected" and "unknown peer" as error codes, so that
users can check against numeric error codes instead of textual error
messages.

Will ease https://github.com/ElementsProject/lightning/issues/3366

Changelog-None

### DIFF
--- a/doc/lightning-fundchannel_cancel.7.md
+++ b/doc/lightning-fundchannel_cancel.7.md
@@ -28,7 +28,12 @@ RETURN VALUE
 On success, returns confirmation that the channel establishment has been
 canceled.
 
-On failure, returns an error.
+On error the returned object will contain `code` and `message` properties,
+with `code` being one of the following:
+
+- -32602: If the given parameters are wrong.
+- -1: Catchall nonspecific error.
+- 306: Unknown peer id.
 
 AUTHOR
 ------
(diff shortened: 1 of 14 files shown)


vasild:
I added the errors to the man pages in 10826f0b. This is not very maintainable because every time the error message is edited in the source code the man page will become outdated. I guess this is acceptable for now.

ZmnSCPxj:
See `doc/lightning-pay.7.md` for reference.  We do not document the exact error string, we just document the numeric error code.

cdecker:
Looks good to me, just needs the change in the docs @ZmnSCPxj suggested. We can be a bit more descriptive in the docs about what the error codes mean and how to resolve them, instead of just showing the terse error messages from the CLI again :-)

vasild:
> See `doc/lightning-pay.7.md` for reference.

I mimicked `doc/lightning-listpeers.7.md` instead :-(

Now I tweaked my changes to look like `doc/lightning-pay.7.md` and also, in another commit ef7b0c73 "converted" `doc/lightning-listpeers.7.md` and others to the `doc/lightning-pay.7.md` style so all are consistent now.

doc/lightning-fundchannel_cancel.7.md
```diff
@@ -28,7 +28,12 @@ RETURN VALUE
 On success, returns confirmation that the channel establishment has been
 canceled.
 
-On failure, returns an error.
+On error the returned object will contain `code` and `message` properties,
+with code being one of the following:
```
ZmnSCPxj: Inconsistent: You indicate "`code`" earlier but only "code" on this line.  Also applies to most of the rest as well.

vasild: Sharp eyes, amended, thanks!

PR #3398, opened 2020-01-05
doc: update FreeBSD installation instructions

vasild:
Now that c-lightning is in the FreeBSD ports system (\o/)
update the installation instructions, which basically boil down
to just `pkg install c-lightning`.

Changelog-None

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -128,45 +128,42 @@ $ lightningd --network=testnet
 ```
 
 To Build on FreeBSD
----------------------
+-------------------
 
 OS version: FreeBSD 11.1-RELEASE or above
 
-Get dependencies:
+c-lightning is in the FreeBSD ports, so install it as any other port
+(dependencies are handled automatically):
 
-    # pkg install -y \
-      autoconf automake bash gettext git gmp gmake libtool \
-      python python3 sqlite3 libsodium py36-mako py36-pip
-    # pip install mrkd
+    # pkg install c-lightning
 
-If you don't have Bitcoin installed locally you'll need to install that
-as well:
+for a binary, pre-compiled package. If you want to compile locally and
+fiddle with compile time options:
 
-    # pkg install -y bitcoin-daemon bitcoin-utils
+    # cd /usr/ports/net-p2p/c-lightning && make install
 
-Clone lightning:
+mrkd is required to build man pages from markdown files (not done by the port):
 
-    $ git clone https://github.com/ElementsProject/lightning.git
-    $ cd lightning
+    # cd /usr/ports/devel/py-pip && make install
+    $ pip install --user mrkd
 
-Build lightning:
+See `/usr/ports/net-p2p/c-lightning/Makefile` for instructions on how to
+build from an arbitrary git commit, instead of the latest release tag.
 
 **Note**: Make sure you've set an utf-8 locale, e.g. 
 `export LC_CTYPE=en_US.UTF-8`, otherwise manpage installation may fail.
 
-    $ ./configure
-    $ gmake
-    $ gmake install
-
 Running lightning:
 
-**Note**: Edit your `/usr/local/etc/bitcoin.conf` to include
-`rpcuser=<foo>` and `rpcpassword=<bar>` first, you may also need to
-include `testnet=1`
+Configure bitcoind, if not already: add `rpcuser=<foo>` and `rpcpassword=<bar>`
+to `/usr/local/etc/bitcoin.conf`, maybe also `testnet=1`.
+
+Configure lightningd: copy `/usr/local/etc/lightningd-bitcoin.conf.sample` to
+`/usr/local/etc/lightningd-bitcoin.conf` and edit according to your needs.
 
     # service bitcoind start
-    $ ./lightningd/lightningd &
-    $ ./cli/lightning-cli help
+    # service lightningd start
+    # lightning-cli --rpc-file /var/db/c-lightning/bitcoin/lightning-rpc --lightning-dir=/var/db/c-lightning help
 
 To Build on NixOS
 --------------------


bitcoin-software:
WOW! mate, thank you so much for contributing to this. is the package with working init scripts as well? was going to write init script

bitcoin-software:
well.. currently i failed to get working with remote btc node cln package from pkg.

should i address this to ports maintainer? (never did this)

my observations:

1) after installing package everything seems ok (except python dependencies are not installed. can this cause any issue?) 
2) the user `c-lightning` is created in the system. and its homedir is `/var/db/c-lightning`. previously i was using such trick: created `.bitcoin` dir in homedir, with specifying remote node rpc credentials. in tihs way it is easy to debug from CLI and same user as cln is run on. now ~/.bitcoin dir is ignored it seems
3) i was not able to connect to remote node:

```
# service lightningd restart
lightningd not running? (check /var/db/c-lightning/lightningd-bitcoin.pid).
Starting lightningd.
Could not connect to bitcoind using bitcoin-cli. Is bitcoind running?
Make sure you have bitcoind running and that bitcoin-cli is able to connect to bitcoind.
You can verify that your Bitcoin Core installation is ready for use by running:
    $ bitcoin-cli -rpcconnect=my-remote-node.tld -rpcport=8332 -rpcuser=... -rpcpassword=... echo 'hello world'
/usr/local/etc/rc.d/lightningd: WARNING: failed to start lightningd
cln2:/var/db/c-lightning@[21:34] # bitcoin-cli -rpcconnect=my-remote-node.tld -rpcport=8332 -rpcuser=usr -rpcpassword=pwd echo 'hello world'
[
  "hello world"
]

```

so with dumb cli command everything is ok, but when starting daemon - nope

ZmnSCPxj:
> i cant find python dependencies in that port... weird.. shouldn't they be there?

`lightningd` by itself should not require python as a dependency.  External plugins might, and our test suite definitely has it, and we provide python bindings, but `lightningd` by itself should not require python.  Unless things have changed since I last checked.

bitcoin-software:
@ZmnSCPxj there was no python dep before at least 0.7.3. Later python was added to build instructions as well. May it be that some of default plugins (are there?!) requires python now?

ZmnSCPxj:
Builtin plugins are all written in C.  Hmm.  Maybe needed for build?  But installation and operation should not require Python.... possibly (cough) our build system is a mess.

bitcoin-software:
Actually, IIRC py36-mako requirement appeared very recently, before 0.8.x (~last weeks of 0.7.x)

"To ./configure this time, there were new additional dependencies required: py36-mako and gettext,"

I found (probably mine) a [question regarding this](https://gitmemory.com/issue/ElementsProject/lightning/3080/533772760). That was actualy 0.7.3 requirement.

ZmnSCPxj:
`mako` is used during build, specifically `tools/generate-wire.py` and `devtools/sql-rewrite.py`.  It should not get used during installation and operation.  It also happens to be used in release process (`devtools/changelog.py`).

bebebe-cmd:
@ZmnSCPxj Thank you very much for clarification! Good to know this. You both guys are rockstars!

bebebe-cmd:
But still, issue with remote node persists. Waiting response from @vasild on that.. am i doing something wrong or there's something broken in package initscript..

bitcoin-software:
@ZmnSCPxj Thank you very much for clarification! Good to know this. You both guys are rockstars! But still, issue with remote node persists.

 Waiting response from @vasild on that.. am i doing something wrong or there's something broken in package initscript..

vasild:
> P.S. thought you are port maintainer, aren't you?

[Yes](https://github.com/freebsd/freebsd-ports/blob/accac85/net-p2p/c-lightning/Makefile#L13)

> i cant find python dependencies in that port... weird.. shouldn't they be there?

The port has build-time-only dependencies to [python](https://github.com/freebsd/freebsd-ports/blob/accac85/net-p2p/c-lightning/Makefile#L24) and [py-mako](https://github.com/freebsd/freebsd-ports/blob/accac85/net-p2p/c-lightning/Makefile#L18). So if you build from source from the port those will be installed automatically during the build. If you use the binary "pkg install ...", you don't need to have Python at all.

> well.. currently i failed to get working with remote btc node cln package from pkg.
> should i address this to ports maintainer? (never did this)

Yes, usually you just send an email to the maintainer or open a bug at https://bugs.freebsd.org/. But see below:

> after installing package everything seems ok (except python dependencies are not installed. can this cause any issue?)

Python is not needed at runtime, all plugins are written in C and installed as binary executable files, no Python is involved.

> i was not able to connect to remote node, even i specify rpc creds in cln config:

A few things to check:

* Double check that your cln config contains all of the correct `bitcoin-rpcconnect=...`, `bitcoin-rpcpassword=...`, `bitcoin-rpcport=...`, `bitcoin-rpcuser=...`.

* The cln config should by default be in `/usr/local/etc/lightningd-bitcoin.conf`. One is not installed by default. An example config is installed under `/usr/local/etc/lightningd-bitcoin.conf.sample`. You did not edit the sample config and leave it as `.sample`, right?

* If the config's location and contents are ok, then check when lightningd is being started it is pointed to it: `sh -x /usr/local/etc/rc.d/lightningd start`. The exact command of lightningd to be executed by the rc script should be `/usr/local/bin/lightningd --lightning-dir=/var/db/c-lightning --network=bitcoin --daemon --conf=/usr/local/etc/lightningd-bitcoin.conf`. Is the script doing that?

* Try to run lightningd manually as the `c-lightning` user instead of via the rc script: `su -m c-lightning -c /usr/local/bin/lightningd --lightning-dir=/var/db/c-lightning --network=bitcoin --daemon --conf=/usr/local/etc/lightningd-bitcoin.conf`. Does that work?

* If even that does not work, then next would be to `ktrace(1)` lightningd and see if it is indeed executing the proper `bitcoin-cli -rpcconnect=... -rpcport=... -rpcuser=... -rpcpassword=...`. But I guess we will figure it out before having to do that :)

Let me know how it goes.

vasild:
@minidisk1147 you touched the FreeBSD installation instructions recently, maybe you have some feedback on this PR?

Btw, a global `pip install foo` as root is [discouraged](https://github.com/freebsd/freebsd-ports/blob/d564de4/devel/py-pip/pkg-message#L4).

minidisk1147:
> @minidisk1147 you touched the FreeBSD installation instructions recently, maybe you have some feedback on this PR?

`make install` looks into git logs to find out whether the last git message of **doc/*.[0-9].md** differs from the git message of **doc/*.[0-9]**. If so, it builds the manpage using the python program "mrkd". (Look for CHANGED_FROM_GIT in the makefiles)

That's why I put the `pip install mrkd` into installation instructions.

> 
> Btw, a global `pip install foo` as root is [discouraged](https://github.com/freebsd/freebsd-ports/blob/d564de4/devel/py-pip/pkg-message#L4).

Thank you for the information, I'm not deep into python and wasn't aware of this. But more important:  Thanks for building a FreeBSD-Package :-)

ZmnSCPxj:
> `make install` looks into git logs to find out whether the last git message of **doc/*.[0-9].md** differs from the git message of **doc/*.[0-9]**

.....

This should be a release process (or a build process, if we just not store the manpages in the repo, nobody reads the manpages on git anyway), not an install process, and must not be done in `make install`.  Grrrrrrr.....

....

bitcoin-software:
@vasild thank you for your replies and instructions.

So, when i do ` sh -x /usr/local/etc/rc.d/lightningd start` daemon is started, but if i do `service c-lightning start` i got error described before. 

Just in case you may want reproduce this in my env:

`$ ssh root@37.59.8.98 -p22878`

`Password for root@cln2.my.domain: pa7eesaoZ9baeT6`

`tmux a`

image

vasild:
> `make install` looks into git logs to find out whether the last git message of **doc/*.[0-9].md** differs from the git message of **doc/*.[0-9]**. If so, it builds the manpage using the python program "mrkd". (Look for CHANGED_FROM_GIT in the makefiles)
> 
> That's why I put the `pip install mrkd` into installation instructions.

Yes. The port downloads an archive from github that does not contain a `.git` directory (is not a git repository), so that `CHANGED_FROM_GIT` fails in an excellent way and `mrkd` is never executed :)

vasild:
@bitcoin-software should be fixed in https://github.com/freebsd/freebsd-ports/commit/cd0fda3eeca06dd1d3d419a8f75eeaa1a14db65d
`bitcoin-cli(1)` tries to create its datadir when it starts even though we are using it solely as a "proxy" to talk to `bitcoind`:
```
# HOME=/ bitcoin-cli -rpcconnect=... echo abcd
************************
EXCEPTION: N5boost10filesystem16filesystem_errorE       
boost::filesystem::create_directory: Read-only file system: "/.bitcoin"       
bitcoin in AppInitRPC()       
```
That is another reason to prefer talking to `bitcoind` directly via TCP instead of executing `bitcoin-cli`.

bitcoin-software:
@vasild thank you so much for your contribution and rapid support! tomorrow will check it out :)

vasild:
This should be good to go. Btw, I like these installation instructions: https://mosh.org/#getting

PR #3399, opened 2020-01-05
fix build instructions for freebsd

bitcoin-software:
current instructions are bit outdated so can bring confusion to users

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -130,13 +130,13 @@ $ lightningd --network=testnet
 To Build on FreeBSD
 ---------------------
 
-OS version: FreeBSD 11.1-RELEASE or above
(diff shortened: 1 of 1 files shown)


bitcoin-software:
ouh... looks like me made a naming of PR/commits not that best.. sorry for that.

anyway the story is that i cannot compile cln using current instructions because latest FreeBSD releases defaults to python37 and moreover there is no py36-mako package anymore (stil can be installed via pip).

So to follow new OS releases and make cln compatible with current this PR is created

bitcoin-software:
https://github.com/ElementsProject/lightning/pull/3398/commits/083c5a836b8bf2145a58c833867f97cfbcb3fa6c not anymore required on that!

Issue #3354, opened 2019-12-18
Bitcoin backend plugins planning


cdecker:
I've been planning to work on a plugin-based bitcoin backend for a while but
haven't gotten around to actually working on it. Since others are trying to
get started on this I thought I'd share some of my thoughts around how this
could be accomplished.

## Current situation

Our current interactions with the bitcoin network is nicely encapsulated in
`lightningd/bitcoin.c`. From there we can see that the RPC methods we call are
the following:

 - `estimatesmartfee` used for fee estimation on a couple of different
   horizons (1, 3, 6 blocks).
 - `getblock` (raw and hex encoded) used to retrieve a block after calling
   `getblockhash` to get its hash given the height.
 - `getblockhash` to translate a height into a blockhash
 - `gettxout` to check whether a tx output is still unspent, i.e., in the UTXO.
 - `getnetworkinfo` used to check that `bitcoind` is a supported version only.
 - `getblockchaininfo` used to check if `bitcoind` is in IBD.
 - `sendrawtransaction` used to send a transaction to the network.
 
## Backend Interface

I'd like to make the interface for backend a bit more coarse-grained, in order
to give the plugins more freedom as to how they accomplish their task. My go
to example is `getblockhash` + `getblock` which is always called in this order
and requires multiple roundtrips to `bitcoind` for a single logical operation.

Some of the methods above can be translated 1-to-1 into the plugin:

 - `sendrawtransaction` can become a simple hook, if only for the reason that
   we don't want to expose it to the JSON-RPC users. In addition with
   prioritized and chained hooks we can have a failover chain that tries one
   plugin after the other in case of failure.
   
Other methods can be combined into more abstract methods:

 - `getblockbyheight` is a combination of `getblockhash` and `getblock` which
   either returns the hash and the block at a given height or an error if
   there is no block (yet). In addition we could give a tiny hint as to the
   context in which the query is done (`scan`) to tell the plugin whether we
   are in a linear scan to catch up with the blockchain head. That'd allow the
   plugin to pre-fetch the next few blocks, effectively pipelining the sync.
 - `getchaininfo` can subsume the calls to `getnetworkinfo` and
   `getblockchaininfo` because they are mostly done on startup, and
   effectively contain the same information. A version number for the set of
   capabilities exposed by the backend plugin would still be in order to guard
   against using an outdated backend with a newer version that expects some
   things to be available.
 - `getoutputbyscid` is a combination of `getblockhash`, `getblock`
[...]

PR #3400, opened 2020-01-06
RPC methods versioning for plugins

darosior:
Rationale described here: https://github.com/ElementsProject/lightning/issues/3354#issuecomment-569715461.

Summary: We may be going to rely on plugin-exposed RPC methods for crucial operations (such as Bitcoin backend), a versioning is useful to avoid nasty corner cases.

Open questions:
- should we allow to register multiple RPC methods with the same name, but different versions ?


darosior:
Marking WIP as I shoudl also expose the version in `list` or `get*info` calls

cdecker:
> Marking WIP as I shoudl also expose the version in `list` or `get*info` calls

You can make a PR unmergeable if you use the ["Create draft Pull request"](https://github.blog/2019-02-14-introducing-draft-pull-requests/) button instead of the "Create Pull Request" button :-)

darosior:
Changed my mind just after :/ I should make use of the "think before clicking the green button" button, too

cdecker:
> Changed my mind just after :/ I should make use of the "think before clicking the green button" button, too

No problem, I thought I might as well point it out for the future :-)

darosior:
I think this is ready for review

cdecker:
After mulling this over for a couple of days I am no longer convinced that
just adding a version number is what we want. Specifically there is no
connection between the version and the signature of the method/hook, which
kind of defeats the point. I've been playing with the idea of having a
meta-description of the signature (name + arguments + return structure) and
hash that to a version number, but it seems a bit too much at the moment.

It could also get rather confusing for plugin devs to distinguish which
version of the method/hook they need to implement. Maybe we should just
version in the method name (e.g. `getblockbyheight1`, `getblockbyheight2`,
...) and document them thoroughly in the docs instead of trying to paper over
the differences with code.

As long as we only add arguments or add more things to the returned results we
wouldn't need to change the method name either. This matches the current
semantics which allows plugins to return richer-than-expected results, and
suggests they should allow more arguments than strictly expected (everybody is
adding `**kwargs` at the end of the signature of pyln plugin methods and
hooks, right?).

Changing the name allows us to skip plugins that weren't updated with the
current signature in mind, and automatically fall back to the built-in
`bitcoin-cli` mechanism.

TL;DR: Maybe we should wait for the versioning to really become an issue, and when we have more information about how things might break.

I'm sorry I sent you down this rabbit hole by suggesting the versions in the
first place, but maybe simpler is also better?

darosior:
> everybody is adding **kwargs at the end of the signature of pyln plugin methods and
hooks, right?

Of course ! *quickly adds a TODO line*
More seriously (and a bit off-subject of this PR) maybe it's worth making YA round of updates for lightningd/plugins as we did with `pyln` replacement and tests ?

> Changing the name allows us to skip plugins that weren't updated with the current signature in mind, and automatically fall back to the built-in bitcoin-cli mechanism.

That's what version numbers aimed to do, but I'm fine with names too.

> Maybe we should wait for the versioning to really become an issue, and when we have more information about how things might break.

I agree, and I'm also in favor of adding less things as the diff for this is becoming really big on my side and it would just make review more complicated for a (somewhat?) premature optimisation.

> I'm sorry I sent you down this rabbit hole by suggesting the versions in the
first place, but maybe simpler is also better?

Absolutely.

I'll PR the drive-by fixups  commits and close this.

ZmnSCPxj:
> everybody is
adding `**kwargs` at the end of the signature of pyln plugin methods and
hooks, right?

/me looks guiltily around, then pretends nothing happened.

PR #3405, opened 2020-01-08
Always notify if a `sendpay` payment attempt is resolved

### LINKED ISSUE #3403: Sendpay notifications not sent unless we have at least one waiter
It appears that the `sendpay_success` and `sendpay_failure` notifications are only sent if we also have at least one JSON connection actively waiting for the result. `notify_sendpay_failure` for example is only called from `sendpay_fail` which terminates a single `sendpay` command or a `waitsendpay` command. If there is no active command we don't call it, and therefore don't get a notification either:

https://github.com/ElementsProject/lightning/blob/899f5dee53203fb0fe49a0fcfc19c77454226261/lightningd/pay.c#L245-L266

The same is true for the success case (no waiter -> no notification). Furthermore if we have multiple waiters we may end up getting the notification multiple times... We also can't piggy back on `tell_waiters_failed` since we sometimes skip that and call `sendpay_fail` directly in a couple of places, replying only to a single `sendpay` or `waitsendpay` command.

We need to have a cleaner separation here: a command that retrieves a prior result, should be terminated directly and not result in a notification, while a notification should be issued whenever a payment attempt terminates independently of whether there is a waiter or not.

cdecker:
Sendpay notifications would get sent only if we have an active watcher, and multiple times if we have more than one watcher. This kind of defeats the point since we can't just fire and forget sendpay and wait for the notification if that notification is bound to the watchers.

This PR fixes that by pulling the notification out of the watcher loop, concentrating the error message allocation in a single place, and inverting ownership of `wallet_payment_store`. This also reduces a couple of useless allocations, since we don't store, free, then reload the `wallet_payment` anymore.

Fixes #3403

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -487,26 +487,26 @@ remote_routing_failure(const tal_t *ctx,
 	return routing_failure;
 }
 
-void payment_store(struct lightningd *ld,
-		   const struct sha256 *payment_hash, u64 partid)
+void payment_store(struct lightningd *ld, struct wallet_payment *payment TAKES)
 {
 	struct sendpay_command *pc;
 	struct sendpay_command *next;
-	const struct wallet_payment *payment;
+	/* Need to remember here otherwise wallet_payment_store will free us. */
+	bool ptaken = taken(payment);
 
-	wallet_payment_store(ld->wallet, payment_hash, partid);
-	payment = wallet_payment_by_hash(tmpctx, ld->wallet,
-					 payment_hash, partid);
-	assert(payment);
+	wallet_payment_store(ld->wallet, payment);
 
 	/* Trigger any sendpay commands waiting for the store to occur. */
 	list_for_each_safe(&ld->sendpay_commands, pc, next, list) {
-		if (!sha256_eq(payment_hash, &pc->payment_hash))
+		if (!sha256_eq(&payment->payment_hash, &pc->payment_hash))
 			continue;
 
 		/* Deletes from list, frees pc */
 		json_sendpay_in_progress(pc->cmd, payment);
 	}
+
+	if (ptaken)
+		tal_free(payment);
 }
 
 void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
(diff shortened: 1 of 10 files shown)


lightningd/pay.c
```diff
@@ -487,26 +487,26 @@ remote_routing_failure(const tal_t *ctx,
...
 
-void payment_store(struct lightningd *ld,
-		   const struct sha256 *payment_hash, u64 partid)
+void payment_store(struct lightningd *ld, struct wallet_payment *payment TAKES)
 {
 	struct sendpay_command *pc;
 	struct sendpay_command *next;
-	const struct wallet_payment *payment;
+	/* Need to remember here otherwise wallet_payment_store will free us. */
+	bool ptaken = taken(payment);
```
ZmnSCPxj: Would `if (taken(payment)) tal_steal(tmpctx, payment);` work as well? `/* keep payment alive across the loop, but schedule it for freeing later. */`

cdecker: Interesting idea, I prefer freeing as soon as possible, so we don't have lingering things, and `valgrind` can get to work, and point out potential use-after-free issues that are masked by us deferring free to the `io_loop`. But I'm sure it'd work perfectly fine to steal onto `tmpctx`.

lightningd/pay.c
```diff
@@ -191,6 +191,21 @@ void json_sendpay_fail_fields(struct json_stream *js,
...
+static const char *sendpay_errmsg_fmt(const tal_t *ctx, int pay_errcode,
+				      const struct routing_failure *fail,
+				      const char *details)
+{
+	char *errmsg;
+	if (pay_errcode == PAY_UNPARSEABLE_ONION)
+		errmsg = "Malformed error reply";
+	else {
+		assert(fail);
+		errmsg = tal_fmt(tmpctx, "failed: %s (%s)",
```
ZmnSCPxj: You passed in a `ctx` to `sendpay_errmsg_fmt`, but you never use it, instead using `tmpctx` here.

cdecker: Thanks @ZmnSCPxj, I've fixed it up, and you can see the change [here](http://46.101.246.115:8888/ElementsProject/lightning/pull/3405/range_diff/dd925d896befd0deceb0bb34fd83b0cb90373fd3..e1264f7f5485ddf21fac52fd7cd6fcc0305e94f2) :+1:

PR #3408, opened 2020-01-10
Tor: Fix an edge case where the connect answer is not zero and not error.

Saibato:
The tor proxy might want auth 0x2 or answer what ever not defined
in https://tools.ietf.org/html/rfc1928.

Changelog-Fixed: We don't send any further request if the return code of connect is not zero or error.

### DIFF
--- a/connectd/tor.c
+++ b/connectd/tor.c
@@ -30,7 +30,7 @@
 #define SOCK_REQ_V5_LEN			5
 #define SOCK_REQ_V5_HEADER_LEN	7
 
-/* some crufts can not forward ipv6*/
+/* some crufts can not forward ipv6 */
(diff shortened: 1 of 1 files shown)


cdecker:
Rebased on top of `master` and made the changelog entry a bit more specific by adding the `TOR: ` prefix. Changes can be inspected [here](http://46.101.246.115:8888/ElementsProject/lightning/pull/3408/range_diff/4c23977c090657e828b1b54bf4ed126a06b0c2aa..56488439244e8610ebdfedebbb86035dd982a6b4). :wink:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3408/commits/56488439244e8610ebdfedebbb86035dd982a6b4

PR #3414, opened 2020-01-14
doc: also build docs when building everything else

vasild:
Before this change `make` would build everything but the docs.
They would be built during `make install` which is unexpected.

Add a dependency to `doc-all` to the `default` target so that docs are
build together with the rest of the programs. Leave the `install-data`
as is, to still build the docs if they are not built yet during install.

Changelog-None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,7 @@ ifeq ($(HAVE_POSTGRES),1)
 LDLIBS += -lpq
 endif
 
-default: all-programs all-test-programs
+default: all-programs all-test-programs doc-all
 
 ccan/config.h: config.vars configure ccan/tools/configurator/configurator.c
 	./configure --reconfigure


vasild:
This change was provoked by:
https://github.com/ElementsProject/lightning/pull/3398#issuecomment-571221211
> This should be a release process (or a build process, if we just not store the manpages in the repo, nobody reads the manpages on git anyway), not an install process, and must not be done in `make install`. Grrrrrrr.....

Maybe reconsider whether to store man pages (generated files) in the git repository?

ZmnSCPxj:
>  Maybe reconsider whether to store man pages (generated files) in the git repository?

I am guessing the reason we do *not* do this is that we need some kind of ` mrkd` tool to build the manpages from the Markdown documents. At least if the manpages are in the repo, somebody can clone the repo and install it without having installed `mrkd`.

GNU has a separation between "maintainer" and "installer", the repo is used only by the maintainer and does not contain any generated files, and then the maintainer installs all the necessary tools and then releases source tarballs (which *do* include generated files) that installers can use with only a minimal system (`sh`, `cc`, `ld`).

PR #3421, opened 2020-01-17
remove `--silent` flag from patch in wire/Makefile

niftynei:
It's been brought to my attention that BusyBox's `patch` binary doesn't support the `--silent` option. (nor `quiet`, as of patch `020abc8856f94d6e355f4daa972ac75fb05ae113` on the busybox master branch).

This patch removes the silent flag from the patch command used to build the wire csv files. Only affects experimental builds.

Changelog-None


cdecker:
ACK aada399

ZmnSCPxj:
I start to wonder if we should do a `busybox` test in travis.  #1222 is similarly triggered by `busybox` `sed` complaining about a GNUism.

PR #3427, opened 2020-01-21
sphinx: Actually use the pad stream to generate the packet

cdecker:
Turns out https://github.com/ElementsProject/lightning/pull/3246 did not actually do anything. We flipped two buffers and were not actually using the chacha20 stream. The buffer we were initializing with the chacha20 stream was simply being overwritten in the next iteration. We need to initialize the `packet->routinginfo` instead.

Keeping this as a draft until we've resolved the broken test-vectors in lightningnetwork/lightning-rfc#697

Changelog-None

### DIFF
--- a/bitcoin/block.c
+++ b/bitcoin/block.c
@@ -13,6 +13,7 @@ bitcoin_block_from_hex(const tal_t *ctx, const struct chainparams *chainparams,
 	u8 *linear_tx;
 	const u8 *p;
 	size_t len, i, num;
+	struct sha256_ctx shactx;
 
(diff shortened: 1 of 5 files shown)


cdecker:
This is now ready for review.

Ping @ZmnSCPxj and @rustyrussell

cdecker:
Re-applying @ZmnSCPxj's ack:

ACK https://github.com/ElementsProject/lightning/pull/3427/commits/2368e8a6055d473183b4ef0d9bd74e8ad64b7958

Issue #3430, opened 2020-01-22
`doc/TOR.md` does not mention the `statictor:` option for `--addr`


ZmnSCPxj:
We already have the ability to make a persistent Tor address without having to modify the `torrc`, but `doc/TOR.md` does not mention it.  `lightningd-config.5` does, though.

If my understanding is correct, if the desire is to have a persistent hidden service and not announce any other method of contact, then assuming a standard Tor installation (9050 SOCKS5, 9051 control) on the same hardware:

    --proxy=127.0.0.1:9050 --bind-addr=127.0.0.1:9735 --addr=statictor:127.0.0.1:9051

Would that be correct @Saibato @cdecker ?

Saibato:
Yup and BTW i would use --always-use-proxy=true , i left some work TODO so others can take part in the process of evolving c/ln and contribute from a user perspective.

PR #3433, opened 2020-01-23
doc/TOR.md: Mention about `statictor`.

ZmnSCPxj:
Fixes: #3430

### DIFF
--- a/doc/TOR.md
+++ b/doc/TOR.md
@@ -31,6 +31,97 @@ To provide the node with a .onion address you can:
 
 * create a **persistent** address with a hidden service.
 
+### Quick Start On Linux
+
+It is easy to create a single persistent Tor address and not announce a public IP.
+This is ideal for most setups where you have an ISP-provided router connecting your
+Internet to your local network and computer, as it does not require a stable
+public IP from your ISP (which might not give one to you for free), nor port
+forwarding (which can be hard to set up for random cheap router models).
+Tor provides NAT-traversal for free, so even if you or your ISP has a complex
+network between you and the Internet, as long as you can use Tor you can
+be connected to.
+
+On most Linux distributions, making a standard installation of `tor` will
+automatically set it up to have a SOCKS5 proxy at port 9050.
+As well, you have to set up the Tor Control Port.
+On most Linux distributions there will be commented-out settings below in the
+`/etc/tor/torrc`:
+
+```
+ControlPort 9051
+CookieAuthentication 1
+CookieAuthFileGroupReadable 1
+```
+
+Uncomment those in, then restart `tor` (usually `systemctl restart tor`  or
+`sudo systemctl restart tor` on most SystemD-based systems, including recent
+Debian and Ubuntu, or just restart the entire computer if you cannot figure
+it out).
+
+Then add these to your `${LIGHTNING_DIR}/config` or other C-Lightning configuration
+(or prepend `--` to each of them and add them to your `lightningd` invocation
+command line):
+
+```
+proxy=127.0.0.1:9050
+bind-addr=127.0.0.1:9735
+addr=statictor:127.0.0.1:9051
+always-use-proxy=true
+```
+
+1.  `proxy` informs C-Lightning that you have a SOCKS5 proxy at port 9050.
+    C-Lightning will assume that this is a Tor proxy, port 9050 is the
+    default in most Linux distributions; you can double-check `/etc/tor/torrc`
+    for a `SocksPort` entry to confirm the port number.
+2.  `bind-addr` informs C-Lightning to bind itself to port 9735.
+    This is needed for the subsequent `statictor` to work.
+    9735 is the normal Lightning Network port.
+    This has to appear before any `statictor:` setting.
+3.  `addr=statictor:` informs C-Lightning that you want to create a persistent
+    hidden service that is based on your node private key.
+    This informs C-Lightning as well that the Tor Control Port is 9051.
+    You can also use `bind-addr=statictor:` instead to not announce the
+    persistent hidden service, but if anyone wants to make a channel with
+    you, you either have to connect to them, or you have to reveal your
+    address to them explicitly (i.e. autopilots and the like will likely
+    never connect to you).
+4.  `always-use-proxy` informs C-Lightning to always use Tor even when
+    connecting to nodes with public IPs.
+    You can set this to `false` or remove it,
+    if you are not privacy-conscious **and** find Tor is too slow for you.
+
+### Tor Browser and Orbot
+
+It is possible to not install Tor on your computer, and rely on just
+Tor Browser.
+Tor Browser will run a built-in Tor instance, but with the proxy at port
+9150 and the control port at 9151
+(the normal Tor has, by default, the proxy at port 9050 and the control
+port at 9051).
+The mobile Orbot uses the same defaults as Tor Browser (9150 and 9151).
+
+You can then use these settings for C-Lightning:
+
+```
+proxy=127.0.0.1:9150
+bind-addr=127.0.0.1:9735
+addr=statictor:127.0.0.1:9151
+always-use-proxy=true
+```
+
+You will have to run C-Lightning after launching Tor Browser or Orbot,
+and keep Tor Browser or Orbot open as long as C-Lightning is running,
+but this is a setup which allows others to connect and fund channels
+to you, anywhere (no port forwarding! works wherever Tor works!), and
+you do not have to do anything more complicated than download and
+install Tor Browser.
+This may be useful for operating system distributions that do not have
+Tor in their repositories, assuming we can ever get C-Lightning running
+on those.
+
+### Detailed Discussion
+
 #### Creation of an auto service for non-persistent .onion addresses
 
 To provide the node a non-persistent .onion address it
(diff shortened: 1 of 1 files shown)


Saibato:
> Fixes: #3430

:+1:  One suggestion, maybe you like to add a line or two mentioning, that when just starting the tor browers app on desktops or phones and tablets, useful for those who run c/ln  i.e. in termux with some litle changes regarding endianess and make ( Reminder: to myself i should post the patch ) the opened Tor browser app or Orbot will automatic open as long as run 9150 and 9151, since statictor is persistent on restart, you have a complete torrc worry free setup that vanish on close and with some magic creates the same onion address on every restart.  You can even prove with that address worry free easy your ln discrete log to delegate in future i.e. a PSBT or generic remote signing service.

doc/TOR.md
```diff
@@ -100,6 +153,41 @@ is necessary to set up a [Tor Hidden Service].
...
+You can also have multiple announced persistent .onion addresses
+by adding `/torblob=BLOB`, where `BLOB` is up to 64 bytes of
+text.
+You can indicate `--addr=statictor:127.0.0.1:9051/torblob=1
+--addr=statictor:127.0.0.1/torblob=2` to create two different
+.onion addresses.
+
+Finally, the default port number for the autogenerated persistent
+.onion address will be 9735, but you can change this by adding
+`/torport=9999` to change the external port for the .onion address.
```
Saibato: Finally, the default **outbound port** number, i.e if your node is behind a vpn or nat forwarder 
the local bound port will be what you define in your generic  --addr///bind-addr=IP[:PORT] line 
siggh should have used `outbound `instead of` torport `for the label
In essence one should leave that as is, since you get a bit more visible if you are the only one who uses a specific port other than 9735, but we support the option and footguns have often dual use. ;-)

doc/TOR.md
```diff
@@ -31,6 +31,59 @@ To provide the node with a .onion address you can:
 
 * create a **persistent** address with a hidden service.
 
+### Quick Start On Linux
+
+It is easy tto create a single persistent Tor address and not announce a public IP.
```
Saibato: nit: tto

doc/TOR.md
ZmnSCPxj: Or "external" instead?  Outgoing suggests it is only going out, but incoming connections go into it as well.

ZmnSCPxj:
Updated as per @Saibato feedback.

doc/TOR.md
```diff
@@ -100,6 +185,42 @@ is necessary to set up a [Tor Hidden Service].
...
+
+Note that you have to specify a `--bind-addr` first before using
+`--addr=statictor:`.
+Generally `--bind-addr=127.0.0.1:9735` should work fine.
+
+You can also have multiple announced persistent .onion addresses
+by adding `/torblob=BLOB`, where `BLOB` is up to 64 bytes of
+text.
+You can indicate `--addr=statictor:127.0.0.1:9051/torblob=1
+--addr=statictor:127.0.0.1/torblob=2` to create two different
```
Saibato: at least 32 char should be used to evade collisions

ZmnSCPxj: Given that you already mix the node privkey and pubkey, is that not enough to evade collisions in practice?  We already assume no two nodes can ever have the same pubkeys.

Saibato: That full magic is only when used without ` /torblob=`  the `torblob `option gives you the opportunity to define in essence your own unique portable tor address with the 32Byte prefix that results after the ED25519-V3 pub key generation by the algo inside tor the human readable onion address part, the second 32byte define the PRF secret, if you define only 32 char the PRF will be from getrandom of sodium. So that the ED25519-V3 algo can  generate unique pub/priv key's.

ZmnSCPxj: Ah, so you mean that it is not mixed with the ECDH secret?  i.e. the ECDH secret is only used if `/torblob` is not specified?

Saibato: yup that was the intention.

Saibato: in essence an ED25519 key is just some random bits.

ZmnSCPxj: Okay, will fix.  Unfortunately underlying hardware has to perform regular maintenance shutdown for a while, so perhaps tomorrow.

Saibato: Hope your meat puppet is well and not effected by things i tried to outsource into silicon in the 80ties till then ...

ZmnSCPxj:
Updated per @Saibato feedback once again.

ZmnSCPxj:
Added `Changelog-none`, rebased on master.

doc/TOR.md
```diff
@@ -100,6 +185,44 @@ is necessary to set up a [Tor Hidden Service].
...
+Then you can use `--addr=statictor:127.0.0.1:9051` instead of
+`--announce-addr=.onionAddressV3`.
+By default V3 onion addresses are generated.
+
+Note that you have to specify a `--bind-addr` first before using
+`--addr=statictor:`.
+Generally `--bind-addr=127.0.0.1:9735` should work fine.
+
+You can also have multiple announced persistent .onion addresses
+by adding `/torblob=BLOB`, where `BLOB` is up to 32 to 64 bytes of
```
Saibato: Wow that protocol change i really missed,
I always wanted to announce multiple address of the same type, but was overruled like multiple regtest in early dev.
Although i guess it was a typo and should word like.
Sugessted change:

> You can defined multiple persistent .onion addresses and Tor will route them from and to your node,
> but lightning in it's BOLTS is a bit strict here and only allows one address of a type, so
> only the the first address of v3 onion type you define will be announced over LN gosssp

ZmnSCPxj: Oh.... that is sad.  Will fix.

Saibato:
ACK https://github.com/ElementsProject/lightning/pull/3433/commits/323ca04fd66290e3591d9d3dd3dcd13f58df464e

I had `/torblob=` in its first version as plugin to define short lived on the fly tor onions so its
a bit different in handling and security assumptions than the worry free pure statictor version.
Maybe i will remove `torblob` option from cmd line and make an update PR, then we have to change that file again, but for now :+1:

doc/TOR.md
```diff
@@ -100,6 +185,51 @@ is necessary to set up a [Tor Hidden Service].
...
+Note that you have to specify a `--bind-addr` first before using
+`--addr=statictor:`.
+Generally `--bind-addr=127.0.0.1:9735` should work fine.
+
+You can also have multiple persistent .onion addresses
+by adding `/torblob=BLOB`, where `BLOB` is 32 to 64 ***random***
+bytes of text.
+Note that this blob will be used to derive the secret key behind
+the .onion address and you should keep the blob secret otherwise
+anyone who steals it can spoof your node.
```
cdecker:
```suggestion
anyone who steals it can spoof your address.
```
The node identity is still bound to the `hsm_secret` not the onion blob. A squatter can only self-assign the onion address preventing you from getting incoming connections through that address.

doc/TOR.md
```diff
@@ -31,6 +31,91 @@ To provide the node with a .onion address you can:
...
 * create a **persistent** address with a hidden service.
 
+### Quick Start On Linux
+
+It is easy to create a single persistent Tor address and not announce a public IP.
+This is ideal for most setups where you have an ISP-provided router connecting your
+Internet to your local network and computer, as it does not require a stable
+public IP from your ISP (which might not give one to you for free), nor port
+forwarding (which can be hard to set up for random cheap router models).
+Tor provides NAT-punching for free, so even if you or your ISP has a complex
```
cdecker:
```suggestion
Tor provides NAT-traversal for free, so even if you or your ISP has a complex
```

NAT-hole-punching is a different technique (send outgoing packets matching the incoming parameters to allocate routing table entries on the router). Traversal is the generic concept.

doc/TOR.md
```diff
@@ -31,6 +31,91 @@ To provide the node with a .onion address you can:
...
+
+On most Linux distributions, making a standard installation of `tor` will
+automatically set it up to have a SOCKS5 proxy at port 9050.
+As well, you have to set up the Tor Control Port.
+On most Linux distributions there will be commented-out settings below in the
+`/etc/tor/torrc`:
+
+    ControlPort 9051
+    CookieAuthentication 1
+    CookieAuthFileGroupReadable 1
```
cdecker: I prefer wrapping in [multiline code-markup](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code), since it is more explicit and allows annotating with the highlight language.

doc/TOR.md
```diff
@@ -31,6 +31,91 @@ To provide the node with a .onion address you can:
...
+it out).
+
+Then add these to your `${LIGHTNING_DIR}/config` or other C-Lightning configuration
+(or prepend `--` to each of them and add them to your `lightningd` invocation
+command line):
+
+    proxy=127.0.0.1:9050
+    bind-addr=127.0.0.1:9735
+    addr=statictor:127.0.0.1:9051
+    always-use-proxy=true
```
cdecker: Code markup.

doc/TOR.md
Saibato: In general one could say this blob cmd is not intended and designed to be worry free, 
Its a mighty cmd and **the blob is the private key of this tor onion address**.
If the user is not careful in the randomness and lose that information like you say, that inbound tor connection could easly be blocked or MITM.   
Its use is quite questionable and maybe we should change or remove it? , but it is i,e,an easy way to get inbound reached behind a nat or fw when ppl only know say the id  or name of the node and have no address and the user defined i.e. the first 16 hex from the node id as blob and the rest 32 char with some randomness.
If the node id is known to the public this allows like you point out quite easy blocking, if not, its a quite private way to get first hidden inbound and then escalate hiddeness or services i,e, signing or new address on the fly by paying an invoice that triggers a script, since that traffic is encoded as you point out by our hsm we should be insofar on every step quite safe.

so maybe
> anyone who know that secret can spoof your address and block income traffic from that address

cdecker: I might have formulated this wrongly, I don't think that generating the blob from which we generate the onion key-pair shouldn't be securely generated, I'm just saying that the way it is formulated now (spoof your node) is not correct, since the Noise_XK handshake authenticates both endpoints. The risk of the tor blob leaking is limited to address spoofing, while the node cannot be impersonated and the funds are not at risk :wink:

Saibato: Yes correct, i also feel strong safe on the c/ln side even if that blob done wrong might dox us or leak on and in tor, so your point is absolutely correct the node is not touched by what the transport does, that is a strong side of ln we should promote to users at every level.
And that node wording should be changed.
siggh documentation is not my turf  ... :smile:

Saibato: if PR #3439 gets merged, even pealing the Tor onion should be no longer possible if blob data say less than 48 char get lost and if greater 16 collisions are not likely.

ZmnSCPxj:
Changed as per feedback from @cdecker

PR #3434, opened 2020-01-23
More malformed cleanups

rustyrussell:
While cleaning up our handling of HTLC failures, I found a few bugs in the code.

Changelog-None

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -179,7 +179,7 @@ void json_sendpay_fail_fields(struct json_stream *js,
 	/* "immediate_routing_failure" is before payment creation. */
 	if (payment)
 		json_add_payment_fields(js, payment);
-	if (pay_errcode == PAY_UNPARSEABLE_ONION)
+	if (pay_errcode == PAY_UNPARSEABLE_ONION && onionreply)
 		json_add_hex_talarr(js, "onionreply", onionreply->contents);
 	else
 		json_add_routefail_info(js,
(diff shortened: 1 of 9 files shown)


lightningd/pay.c
```diff
@@ -551,6 +575,11 @@ void payment_failed(struct lightningd *ld, const struct htlc_out *hout,
 		pay_errcode = PAY_UNPARSEABLE_ONION;
 		fail = NULL;
 		failmsg = NULL;
+	} else if (hout->failcode) {
```
ZmnSCPxj: Whut?  How did we get a `failcode` here if the onion was malformed?  If we got a `failcode` from the direct peer, should not we consider it as a local failure and the `localfail` was set?

cdecker: The erring node was not able to parse the onion, but it was able to return an error (it knows it has the correct shared-secret since the HMAC test succeeds, but the parsing of the payload failed). This is `PERM|22 (invalid_onion_payload)` if I'm not mistaken, not one of the `BADONION|xyz` failcodes.

lightningd/peer_htlcs.c
```diff
@@ -1682,9 +1682,8 @@ void peer_got_revoke(struct channel *channel, const u8 *msg)
 	for (i = 0; i < tal_count(changed); i++) {
 		/* If we're doing final accept, we need to forward */
 		if (changed[i].newstate == RCVD_ADD_ACK_REVOCATION) {
-			if (!peer_accepted_htlc(channel, changed[i].id, false,
-						&failcodes[i]))
-				return;
+			peer_accepted_htlc(channel, changed[i].id, false,
+					   &failcodes[i]);
```
cdecker: I don't quite follow the logic why this check was removed. @rustyrussell could you help me?

rustyrussell: Just because an htlc failed does *not* mean we should stop processing the rest!

If it's due to an internal error, sure, [eer_accepted_htlc calls channel_internal_error() and we clean up the subdaemon.  But there are some paths which return false and don't call this: in particular, the case where we can't process the onionpacket.

In these cases, we need to respond to channeld, and save things to the db.  Otherwise it just sits there waiting for a reply we never send.

lightningd/pay.c
ZmnSCPxj: If so, then the succeeding "Must be remote fail." case should cover it.  So it is this entire case which confuses me.

rustyrussell: localfail is a bit of a misnomer.  It means *lightningd* gave an error (eg. can't find next hop).  *channeld* returning an error goes down the "remote fail" path.

Let me add a comment: see fixup patch at end.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3434/commits/aabcb3ecfcda96fa618fb3110aaaace89fd739b4

cdecker:
Rebased and squashed :-)

Issue #1366, opened 2018-04-12
Umask for creating socket should be available as cmd param


analogic:
Right now there is no way to set up rights on rpc socket, everytime lightningd is restarted new rpc socket is created with rw only for running user and thus making very hard to communicate between different users reliably.

Sjors:
Just ran into that as well. Although there's currently no way to constrain RPC permissions. I'd like to be able to install RPC consumers as a separate users. This would be especially useful if there's a way to constrain RPC permissions (perhaps per socket?). E.g. a limited ability to create and check the status of invoices.

ZmnSCPxj:
@Sjors, it seems you are proposing that we impose some kind of capabilities system to particular RPC sockets? I think that would be useful, but is complex enough to deserve a separate issue/feature request here.

Sjors:
@ZmnSCPxj done in #1535

PR #3437, opened 2020-01-24
lightningd/jsonrpc.c: Set JSON-RPC socket permissions by command line.

### LINKED ISSUE #3394: Persistent permissions for 'lightning-rpc'
In order to access the `lightning-rpc` socket with a different user, I would like to use group permissions `g+rw` set specifically on the socket (and nothing else in the lightning-dir).

As lightningd resets the permissions to `600` on every restart (probably by recreating the socket), I currently set these permissions with a script run by systemd after starting lightningd. It just waits 10s and then sets the permissions. Waiting a specific time is unreliable, though, as sometimes even 10s is not enough for the `lightning-rpc` socket to be recreated (e.g. if `bitcoind` is still warming up). And if there are external dependencies, maybe even waiting a few minutes might be not enough?

Is there a more elegant way to permanently set the `lightning-rpc` socket permissions, in my case to `660`?

ZmnSCPxj:
Fixes: #1366

Arguably also works for #3394.  It is probably more sensible to set permissions via this patch and `--rpc-file-mode`, since this will work in a new install or if you place the RPC file in a `tmpfs` (which might make more sense as well, so that at startup the rpc file does not exist yet, so you can wait on the file existing to determine if you can now connect to it and the correct lightningd already is listening).

### DIFF
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -188,6 +188,14 @@ cause it to reopen this file (useful for log rotation)\.
 Set JSON-RPC socket (or /dev/tty), such as for \fBlightning-cli\fR(1)\.
 
 
+ \fBrpc-file-mode\fR=\fIMODE\fR
+Set JSON-RPC socket file mode, as a 4-digit octal number\.
+Default is 0600, meaning only the user that launched lightningd
+can command it\.
+Set to 0660 to allow users with the same group to access the RPC
+as well\.
+
+
  \fBdaemon\fR
 Run in the background, suppress stdout and stderr\.
 
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -149,6 +149,13 @@ cause it to reopen this file (useful for log rotation).
  **rpc-file**=*PATH*
 Set JSON-RPC socket (or /dev/tty), such as for lightning-cli(1).
 
+ **rpc-file-mode**=*MODE*
+Set JSON-RPC socket file mode, as a 4-digit octal number.
+Default is 0600, meaning only the user that launched lightningd
+can command it.
+Set to 0660 to allow users with the same group to access the RPC
+as well.
+
  **daemon**
 Run in the background, suppress stdout and stderr.
 
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -1074,7 +1074,7 @@ bool command_check_only(const struct command *cmd)
 void jsonrpc_listen(struct jsonrpc *jsonrpc, struct lightningd *ld)
 {
 	struct sockaddr_un addr;
-	int fd, old_umask;
+	int fd, old_umask, new_umask;
 	const char *rpc_filename = ld->rpc_filename;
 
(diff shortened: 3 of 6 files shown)


ZmnSCPxj:
A comment: JSON, technically speaking, disallows encoding octal in the JSON stream, and the only numbers that can start with 0 are 0 and floating-point numbers that have *only* 0 before the `.` sign.  Of course, traditionally we express file modes in octal, because tradition.  So in `listconfigs`, the `rpc-file-mode` setting is a string that is the octal representation of the mode.

cdecker:
The implementation looks good to me, however the reason I was reluctant to add a CLI flag was that it'd also get us to `--rpc-file-owner` and `-rpc-file-group`. Would it make sense to rather not recreate the socket file on each startup, keeping any manual changes in place? That'd be less unexpected than the file reverting any changes an operator might have done on the file, and free us from having to wire all possible file-related changes through `lightningd` options. It'd also match the `systemd`-way-of-doing-things, where a separate socket unit can pre-populate the socket file with the desired ACLs before starting the node.

Then again, I see the point of restoring a known good state on each startup, which is what I think @rustyrussell aimed to do by recreating every time. Just wanted to discuss this briefly before merging. What do you think @rustyrussell @ZmnSCPxj?

ZmnSCPxj:
I intend to set up my own rpc-file in a `tmpfs`, for the same reason: a clean startup.  As well, it allows me to determine if `lightningd` is ready to listen, by checking if the file exists (since it is in a `tmpfs`, at system boot the file does not exist yet, then when `lightningd` is ready to listen, it exists tthen).

Ownership and groupship are generally handled by running daemons in their own unique UID and GID, e.g. Tor is typically run in `tor:tor` or `debian-tor:debian-tor` and if you want to let particular users access the Tor control port you do `CookieAuthFileGroupReadable 1` and add the `tor`/`debian-tor` group to the user ancillary group list.  That is how I intend to set up `lightningd` myself.

cdecker:
> Ownership and groupship are generally handled by running daemons in their own unique UID and GID, e.g. Tor is typically run in `tor:tor` or `debian-tor:debian-tor` and if you want to let particular users access the Tor control port you do `CookieAuthFileGroupReadable 1` and add the `tor`/`debian-tor` group to the user ancillary group list. That is how I intend to set up `lightningd` myself.

Sounds compelling. Would you suggest we change the file mask to allow groups then? It's something that several projects have bumped against, and I like your "add them to the `lightningd` group" proposal :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3437/commits/5c28d1f8d13291024884060f4060a4653cfcf2a6

ZmnSCPxj:
>Would you suggest we change the file mask to allow groups then?

This command-line argument allows you to set this...?  Just pass in `0660` instead of the default `0600`.  Or `0666` if you are insane and want to get your node funds stolen.  In retrospect a `--rpc-file-allow-group` might have made better sense, since the only sensible settings are `0600` (no group) and `0660` (allow group).

PR #3438, opened 2020-01-24
Implement `immediate` flag for `waitanyinvoice`.

ZmnSCPxj:
An alternative to #3449. Closes: #3449

This adds the `immediate` flag, equivalent to `timeout=0` on #3449


contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -987,14 +987,18 @@ def stop(self):
         """
         return self.call("stop")
 
-    def waitanyinvoice(self, lastpay_index=None):
+    def waitanyinvoice(self, lastpay_index=None, immediate=False, **kwargs):
         """
         Wait for the next invoice to be paid, after {lastpay_index}
-        (if supplied)
+        (if supplied).
+        Faill immediately if no invoice paid yet, instead of waiting,
```
cdecker:
```suggestion
        Fail immediately if no invoice paid yet, instead of waiting,
```

ZmnSCPxj:
> Just as a discussion topic, would a `timeout` parameter, that may be `0` and return immediately, be a more flexible way to implement this? The downside obviously is that it'd require a timer and a timeout callback similar to the success callback.

I considered this as well --- `wallet/invoices.c` can certainly handle having the callback context tal-deleted under it --- but then I thought what would be the point of using anything other than 0 or `UINT32_MAX` for that.

Or maybe it is just better to have a consistent interface, and a good number of our commands do have `timeout`.

ZmnSCPxj:
Changed to use a timeout instead of threading an `immediate` parameter through to `wallet/invoices.c`  The `wallet/invoices.c` sub-module is perfectly capable of having its callback cancelled, so this is perfectly fine and thus `immediate` need not be passed in to the invoices sub-module.

PR #3439, opened 2020-01-25
insert by default entropy in the torblob string. 

Saibato:
There might be the case, that a user wish to define less than 32 or 64 char of blob string.
We allow that, as there might be uses cases for that, but we init now the upper half of the key bytes for the Tor V3 key by default with random data from libsodium.

That insures that the .onion generated with the /torblob= option is still persistent over reboots or backups and system changes,
but will now by default have a minimum 256 bit entropy, whatever the tor key gen from blob algorithm might be.
While insuring that the blob data on the cmd line will be way less or non tor toxic if lost.
 
One nice side effect will be, the same way like your node id bound tor onion address works with the plain statictor:port:ip command without blob, also the torblob generated key material for the tor onion will be new, if less than 64 char torblob used, on every restart of the node, while the address stays always backup worry free the same.    

This PR is backwards compatible, all statictor blob defined onion address stay as are if the node is updated.
  
A follow up PR might have the option /genfrom_short_entropy and mix in static node id key material to have say a bunch of safe tor address just from short one char strings like |1| , |2| , |3| .... but i have to figure out a bit, if that idea is not to toxic for or our hsm key.

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -511,6 +511,10 @@ bool parse_wireaddr_internal(const char *arg, struct wireaddr_internal *addr,
 		addr->itype = ADDR_INTERNAL_STATICTOR;
 		addr->u.torservice.port = DEFAULT_PORT;
 		memset(&(addr->u.torservice.blob[0]), 0, sizeof(addr->u.torservice.blob));
+		/* Init the upper part of blob with random bytes so that a user can gen
+		 * an persistent .onion from less than 32 char blobstring with at least 128bit entropy
+		 */
+		randombytes_buf((void * const)&(addr->u.torservice.blob[32]), 32);
 
 		/* Format is separated by slash. */
 		char **parts = tal_strsplit(tmpctx, arg, "/", STR_EMPTY_OK);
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
Sorry, it is not clear to me, but it looks to me like the generated tor .onion address will differ at each run?  Given that this is ***`statictor`*** should not it be the same address each time (i.e. "static")?

I would suggest instead initializing the buffer with the key we get from ecdh with the `hsmd`,

Saibato:
> Sorry, it is not clear to me, but it looks to me like the generated tor .onion address will differ at each run? Given that this is _**`statictor`**_ should not it be the same address each time (i.e. "static")?
> 
> I would suggest instead initializing the buffer with the key we get from ecdh with the `hsmd`,

Yes good idea i stole from you out of the comment in the docu pr, i have done an ugly QAD draft  that address your idea to derive them with short seed entropy from ecdh and hsm. pls see https://github.com/ElementsProject/lightning/commit/c35c14648e47bd7a18c72535ec5d3e8f96e4d1f1 and https://github.com/ElementsProject/lightning/commit/64951c28b5887ce2d01fc9503dacce4e972b803b
And maybe to you wonder, but in fact tor generates the pub onion key from the first 32 byte scalar the scalar+prf ( upper 32 byte random ) are used for the private key of the onion.   

I general one can say to use less the 64 char with the blob is not recommended, but foot guns have often dual use, for instance plausible deniability.  Did you btw ask yourself not about the AAAAAAAAAAA.... memes?

PR #3440, opened 2020-01-25
Implement dynafed block header support for elements >= 0.18.1

### LINKED ISSUE #3362: Implement dynafed support for Elements 0.18+
This is a tracking issue for `dynafed` support necessary to be able to run
against liquid with `elementsd` v0.18+.

Since v0.18 `elementsd` supports dynamic federations (`dynafed`), however our
block header parsing does not support that format yet, resulting in a crash of
the node when attempting to parse a dynafed header. Notice that starting with
v0.18 the default header is dynafed.

Here are examples of the parsing logic in other languages:

 - [python3](https://github.com/ElementsProject/elements/blob/master/test/functional/test_framework/messages.py#L939)
 - [rust](https://github.com/ElementsProject/rust-elements/blob/master/src/block.rs#L284)
 
I'm planning implement this soon :wink:

cdecker:
This implements support for the upcoming `elementsd` dynafed block header
format, which is slightly different from either the bitcoin block header
format as well as the `elementsd` <= 0.17.0 block header format. A couple of
fields were added, 3 different serialization formats for the information
(`null`, `compact` and `full`) and some of those fields are not hashed into
the block hash, since they'd be self-referential (like the pre-0.18.0 solution
field).

In order to implement this I chose to also simplify the block hash generation,
to happen in parallel with the deserialization. This has a couple of
side-effects:

 - The block hash is computed only once, instead of multiple times when we
   call `bitcoin_block_blkid`. For compatibility I left that function, but it
   now just copies over the cached hash instead of re-serializing it.
 - We do not need any serialization code for the block header, since we are
   just consumers of blocks, and we don't have to generate new ones from its
   constituent parts.
 - We don't need to carry block header fields around that are otherwise of no
   use to us. It always felt weird having to remember constituent parts only
   to be able to re-serialize the header. Now we can just feed it into the
   hashing algorithm and skip the actual fields. This also ends up saving us a
   couple of allocations.
   
There are a couple of drive-by fixes, such as weening off of `elementsd`'s
deprecated `generate` RPC in favor of `generatetoaddress`, but they should be
minor.

Fixes #3362

### DIFF
--- a/bitcoin/block.c
+++ b/bitcoin/block.c
@@ -91,27 +91,7 @@ bitcoin_block_from_hex(const tal_t *ctx, const struct chainparams *chainparams,
 void bitcoin_block_blkid(const struct bitcoin_block *b,
 			 struct bitcoin_blkid *out)
 {
-	struct sha256_ctx shactx;
-
-	sha256_init(&shactx);
-	sha256_le32(&shactx, b->hdr.version);
-	sha256_update(&shactx, &b->hdr.prev_hash, sizeof(b->hdr.prev_hash));
-	sha256_update(&shactx, &b->hdr.merkle_hash, sizeof(b->hdr.merkle_hash));
-	sha256_le32(&shactx, b->hdr.timestamp);
-
-        if (is_elements(chainparams)) {
-		size_t clen = tal_bytelen(b->elements_hdr->proof.challenge);
-		sha256_le32(&shactx, b->elements_hdr->block_height);
-
-		sha256_varint(&shactx, clen);
-		sha256_update(&shactx, b->elements_hdr->proof.challenge, clen);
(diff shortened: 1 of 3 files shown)


cdecker:
As a minor caveat, there are intermediate versions (v0.18.0, v0.18.1.1,
v0.18.1.2) that predate the introduction of the `elided_root` in the block
header params when serialized in the `compact` format. However, there is no
programmatic way to identify whether the block header will contain the field
or not, since the version field was not updated. In this PR I chose to
implement the parsing that includes the `elided_root`, therefore running any
of the versions above results in an incompatibility and will crash
`lightningd`. You've been warned :warning:

More details on the rationale for the change can be found here: https://github.com/ElementsProject/elements/pull/773

PR #3441, opened 2020-01-26
Use dedicated type for error codes

vasild:
Before this patch we used `int` for error codes. The problem with
`int` is that we try to pass it to/from wire and the size of `int` is
not defined by the standard. So a sender with 4-byte `int` would write
4 bytes to the wire and a receiver with 2-byte `int` (for example) would
read just 2 bytes from the wire.

To resolve this:

* Introduce an error code type with a known size:
  `typedef int32_t errcode_t`.

* Change all error code macros to constants of type `errcode_t`.
  Constants also play better with gdb - it would visualize the name of
  the constant instead of the numeric value.

* Change all functions that take error codes to take the new type
  `errcode_t` instead of `int`.

* Introduce towire / fromwire functions to send / receive the newly added
  type `errcode_t` and use it instead of `towire_int()`.

In addition:

* Remove the now unneeded `towire_int()`.

* Replace a hardcoded error code `-2` with a new constant
  `INVOICE_EXPIRED_DURING_WAIT` (903).

Changelog-Changed: The waitinvoice command would now return error code 903 to designate that the invoice expired during wait, instead of the previous -2

### DIFF
--- a/common/errcode.h
+++ b/common/errcode.h
@@ -1,0 +1,13 @@
+#ifndef LIGHTNING_COMMON_ERRCODE_H
+#define LIGHTNING_COMMON_ERRCODE_H
+
+#include "config.h"
+
+#include <ccan/short_types/short_types.h>
+#include <inttypes.h>
+
+typedef s32 errcode_t;
+
+#define PRIerrcode PRId32
+
+#endif /* LIGHTNING_COMMON_ERRCODE_H */
--- a/common/json.c
+++ b/common/json.c
@@ -65,6 +65,30 @@ bool json_to_u64(const char *buffer, const jsmntok_t *tok,
 	return true;
 }
 
+bool json_to_s64(const char *buffer, const jsmntok_t *tok, s64 *num)
+{
+	char *end;
+	long long l;
+
+	l = strtoll(buffer + tok->start, &end, 0);
+	if (end != buffer + tok->end)
+		return false;
+
+	BUILD_ASSERT(sizeof(l) >= sizeof(*num));
+	*num = l;
+
+	/* Check for overflow/underflow */
+	if ((l == LONG_MAX || l == LONG_MIN) && errno == ERANGE)
+		return false;
+
+	/* Check if the number did not fit in `s64` (in case `long long`
+	is a bigger type). */
+	if (*num != l)
+		return false;
+
+	return true;
+}
+
 bool json_to_double(const char *buffer, const jsmntok_t *tok, double *num)
 {
 	char *end;
@@ -122,22 +146,29 @@ bool json_to_u32(const char *buffer, const jsmntok_t *tok,
 
 bool json_to_int(const char *buffer, const jsmntok_t *tok, int *num)
 {
-	char *end;
-	long l;
+	s64 tmp;
 
-	l = strtol(buffer + tok->start, &end, 0);
-	if (end != buffer + tok->end)
+	if (!json_to_s64(buffer, tok, &tmp))
+		return false;
+	*num = tmp;
+
+	/* Just in case it doesn't fit. */
+	if (*num != tmp)
 		return false;
 
-	BUILD_ASSERT(sizeof(l) >= sizeof(*num));
-	*num = l;
+	return true;
+}
 
-	/* Check for overflow/underflow */
-	if ((l == LONG_MAX || l == LONG_MIN) && errno == ERANGE)
+bool json_to_errcode(const char *buffer, const jsmntok_t *tok, errcode_t *errcode)
+{
+	s64 tmp;
+
+	if (!json_to_s64(buffer, tok, &tmp))
 		return false;
+	*errcode = tmp;
 
-	/* Check for truncation */
-	if (*num != l)
+	/* Just in case it doesn't fit. */
+	if (*errcode != tmp)
 		return false;
 
 	return true;
--- a/common/jsonrpc_errors.h
+++ b/common/jsonrpc_errors.h
@@ -3,55 +3,59 @@
  */
 #ifndef LIGHTNING_COMMON_JSONRPC_ERRORS_H
 #define LIGHTNING_COMMON_JSONRPC_ERRORS_H
+
 #include "config.h"
 
+#include <common/errcode.h>
+
 /* Standard errors defined by JSON-RPC 2.0 standard */
-#define JSONRPC2_INVALID_REQUEST	-32600
-#define JSONRPC2_METHOD_NOT_FOUND	-32601
-#define JSONRPC2_INVALID_PARAMS		-32602
+static const errcode_t JSONRPC2_INVALID_REQUEST = -32600;
+static const errcode_t JSONRPC2_METHOD_NOT_FOUND = -32601;
+static const errcode_t JSONRPC2_INVALID_PARAMS = -32602;
 
 /* Uncategorized error.
  * FIXME: This should be replaced in all places
  * with a specific error code, and then removed.
  */
-#define LIGHTNINGD                      -1
+static const errcode_t LIGHTNINGD = -1;
 
 /* Developer error in the parameters to param() call */
-#define PARAM_DEV_ERROR                 -2
+static const errcode_t PARAM_DEV_ERROR = -2;
 
 /* Plugin returned an error */
-#define PLUGIN_ERROR                    -3
+static const errcode_t PLUGIN_ERROR = -3;
 
 /* Errors from `pay`, `sendpay`, or `waitsendpay` commands */
-#define PAY_IN_PROGRESS			200
-#define PAY_RHASH_ALREADY_USED		201
-#define PAY_UNPARSEABLE_ONION		202
-#define PAY_DESTINATION_PERM_FAIL	203
-#define PAY_TRY_OTHER_ROUTE		204
-#define PAY_ROUTE_NOT_FOUND		205
-#define PAY_ROUTE_TOO_EXPENSIVE		206
-#define PAY_INVOICE_EXPIRED             207
-#define PAY_NO_SUCH_PAYMENT		208
-#define PAY_UNSPECIFIED_ERROR		209
-#define PAY_STOPPED_RETRYING		210
+static const errcode_t PAY_IN_PROGRESS = 200;
+static const errcode_t PAY_RHASH_ALREADY_USED = 201;
+static const errcode_t PAY_UNPARSEABLE_ONION = 202;
+static const errcode_t PAY_DESTINATION_PERM_FAIL = 203;
+static const errcode_t PAY_TRY_OTHER_ROUTE = 204;
+static const errcode_t PAY_ROUTE_NOT_FOUND = 205;
+static const errcode_t PAY_ROUTE_TOO_EXPENSIVE = 206;
+static const errcode_t PAY_INVOICE_EXPIRED = 207;
+static const errcode_t PAY_NO_SUCH_PAYMENT = 208;
+static const errcode_t PAY_UNSPECIFIED_ERROR = 209;
+static const errcode_t PAY_STOPPED_RETRYING = 210;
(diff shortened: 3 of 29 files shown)


common/errcode.h
```diff
@@ -0,0 +1,11 @@
+#ifndef LIGHTNING_COMMON_ERRCODE_H
+#define LIGHTNING_COMMON_ERRCODE_H
+
+#include <inttypes.h>
+#include <stdint.h>
+
+typedef int32_t errcode_t;
```
ZmnSCPxj: `#include<ccan/short_types/short_types.h>` instead, and `typedef s32 errcode_t;`, for consistency with the rest of the codebase.

common/json.c
```diff
@@ -65,6 +66,30 @@ bool json_to_u64(const char *buffer, const jsmntok_t *tok,
 	return true;
 }
 
+bool json_to_s64(const char *buffer, const jsmntok_t *tok, int64_t *num)
```
ZmnSCPxj: s/`int64_t`/`s64`/ .  Any code file that uses `u32` has included `ccan/short_types/short_types.h`

common/json.c
```diff
@@ -122,22 +147,29 @@ bool json_to_u32(const char *buffer, const jsmntok_t *tok,
 
 bool json_to_int(const char *buffer, const jsmntok_t *tok, int *num)
 {
-	char *end;
-	long l;
+	int64_t s64;
```
ZmnSCPxj: s/`int64_t`/`s64`/ as well.  Also the variable name probably needs to be changed because `s64 s64` is bleah.

common/jsonrpc_errors.h
```diff
@@ -3,54 +3,58 @@
...
+
 #include "config.h"
 
+#include <common/errcode.h>
+
 /* Standard errors defined by JSON-RPC 2.0 standard */
-#define JSONRPC2_INVALID_REQUEST	-32600
-#define JSONRPC2_METHOD_NOT_FOUND	-32601
-#define JSONRPC2_INVALID_PARAMS		-32602
+extern const errcode_t JSONRPC2_INVALID_REQUEST;
```
ZmnSCPxj: This.... just adds some space to our .const sections without reducing the actual .text sections in practice (because the code in .text will need an address to read this variable from).  It also increases the effort to make a new error code, now we have to modify the `.h` *and* a `.c`, and we already know that developers do not make the effort to make a new error code in a single `.h` file already, everybody just keeps reusing `LIGHTNINGD` (which is sad, you are saying `lightningd` itself is an error....).

Suggest instead `static const errcode_t JSONRPC2_INVALID_REQUEST = -32600` which decent compilers should optimize away (i.e. no actual storage is allocated for these variables) if not used by the compilation unit\.  `gdb` will still be able to grok these variables as well.

common/errcode.h
vasild: Done in fb8ab0aa.

common/json.c
vasild: Done in fb8ab0aa.

common/json.c
vasild: Done in fb8ab0aa.

common/jsonrpc_errors.h
vasild: I deliberately avoided `static` variables in the `.h` file because that will define the symbols in every `.o` file (that uses them).

Currently there are no `static` variables defined in header files, but there are 15 `extern` ones:

```
common/bech32.h:119:extern const char bech32_charset[32];
common/bech32.h:120:extern const int8_t bech32_charset_rev[128];
common/configdir.h:9:extern bool deprecated_apis;
common/derive_basepoints.h:160:extern struct secrets *dev_force_channel_secrets;
common/derive_basepoints.h:161:extern struct sha256 *dev_force_channel_secrets_shaseed;
common/dev_disconnect.h:33:extern bool dev_suppress_commit;
common/per_peer_state.h:61:extern bool dev_fast_gossip;
common/status.h:33:extern volatile bool logging_io;
common/utils.h:11:extern secp256k1_context *secp256k1_ctx;
common/utils.h:13:extern const struct chainparams *chainparams;
common/utils.h:63:extern const tal_t *tmpctx;
gossipd/seeker.h:27:extern bool dev_suppress_gossip;
lightningd/lightningd.h:261:extern bool tal_oom_ok;
lightningd/log.h:59:extern struct log *crashlog;
plugins/libplugin.h:19:extern bool deprecated_apis;
```


Done in fb8ab0aa.

ZmnSCPxj: Seems to be optimized out with `-O2`:

```
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ cat main.c

static const int FOO = 42;
static const int BAR = 43;

int main() {
        return FOO;
}

zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ gcc -O2 -g main.c
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ gdb ./a.out
GNU gdb (Ubuntu 7.11.1-0ubuntu1~16.5) 7.11.1
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./a.out...done.
(gdb) p &FOO
Can't take address of "FOO" which isn't an lvalue.
(gdb) p &BAR
Can't take address of "BAR" which isn't an lvalue.
(gdb) p FOO
$1 = 42
(gdb) p BAR
$2 = 43
(gdb) quit
```

ZmnSCPxj: Also optimized away in `gcc -O1`.  At `gcc -O0` the variables are instantiated and it is possible to get their address from `gdb`.  This is `gcc --version` `gcc (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609`, which seems somewhat stable; latest GCC is 9.2, it seems unlikely to me that this will be reverted in future GCC.

ZmnSCPxj: Looking at the assembly from `gcc -S`, well it looks like the `.debug_info` section is populated with duplicates from each compilation unit, so you have a point there, but debug information is not loaded during normal execution and the generated binary code is shorter with `static` than with `extern` (because no need for an externally-resolved address).  Installer can always `strip` debug information from a binary later if size of the debugging info is an issue.

vasild: I see `static` wins :)

rustyrussell:
Hmm, this is a readability improvement (though I generally stick with #define for constants), but note that the compiler won't see a difference between errcode_t and int, because typedefs are completely fungible.

I couldn't understand how we ever got to/fromwire_int until I traced the git history and saw you introduced it yourself in an earlier patch! :)

Ack, pending check fixes.

ZmnSCPxj:
> but note that the compiler won't see a difference between errcode_t and int, because typedefs are completely fungible.

inb4 `typedef struct { s32 value; } errcode_t;`

ZmnSCPxj:
```
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ cat tmp.c
typedef struct { int value; } errcode_t;

static const errcode_t FOO = {42};
static const errcode_t BAR = {0};

int main() {
        return FOO;
}
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ gcc -O1 -g tmp.c
tmp.c: In function ‘main’:
tmp.c:7:9: error: incompatible types when returning type ‘errcode_t’ {aka ‘const struct <anonymous>’} but ‘int’ was expected
    7 |  return FOO;
      |         ^~~
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ ed tmp.c
140
/return FOO/
        return FOO;
s/FOO/FOO.value/
wq
146
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ cat tmp.c
typedef struct { int value; } errcode_t;

static const errcode_t FOO = {42};
static const errcode_t BAR = {0};

int main() {
        return FOO.value;
}
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ gcc -O1 -g tmp.c
zmnscpxj@zmnscpxj-server-2020:~/dvpt/tmp$ gdb ./a.out 
GNU gdb (Ubuntu 8.3-0ubuntu1) 8.3
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from ./a.out...
(gdb) p FOO
$1 = {value = 42}
(gdb) p &FOO
Can't take address of "FOO" which isn't an lvalue.
(gdb) p BAR
$2 = {value = 0}
(gdb) p &BAR
Can't take address of "BAR" which isn't an lvalue.
(gdb) quit
```

In *some* modern ABIs as well, a `struct` than can fit in a machine word can be passed as-is inside a register, just like an `int` would, if passed by value rather than via pointer.  Though not all modern ABIs allow this.

vasild:
> Hmm, this is a readability improvement (though I generally stick with #define for constants), but note that the compiler won't see a difference between errcode_t and int, because typedefs are completely fungible.

The main reason I used constant variables over macros is to ensure that they are of type `errcode_t` and not the default `int`. Now I realize this can also be achieved by `#define FOO (errcode_t)123` (yuck!). Plus some minor benefits for `gdb`.

 
> I couldn't understand how we ever got to/fromwire_int until I traced the git history and saw you introduced it yourself in an earlier patch! :)

I couldn't understand either how did I end up authoring `towire_int()` which is contradictory by itself - `send_in_a_portable_way__a_type_that_cannot_be_sent_in_portable_way()`. I guess everybody got fooled by the idea [initially](https://github.com/ElementsProject/lightning/pull/3397#issuecomment-572572068). Then I [realized](https://github.com/ElementsProject/lightning/pull/3397#issuecomment-573424035) it has some problems, but it was the way with least resistance and was ok, given the current PR was coming.

Btw there is also `towire_double()` which should be ditched. 

> Ack, pending check fixes.

Thanks!

lightningd/invoice.c
```diff
@@ -82,8 +82,7 @@ static struct command_result *tell_waiter(struct command *cmd,
 		json_add_invoice(response, details);
 		return command_success(cmd, response);
 	} else {
-		/* FIXME: -2 should be a constant in jsonrpc_errors.h.  */
-		response = json_stream_fail(cmd, -2,
+		response = json_stream_fail(cmd, INVOICE_EXPIRED_DURING_WAIT,
```
ZmnSCPxj: Probably also need to update: https://github.com/ElementsProject/lightning/blame/602b81fef2f47834aa815d723d0f210af80062ce/doc/lightning-waitinvoice.7.md#L24-L26

vasild: Done in e3879d93.

vasild:
> It is the `waitinvoice` command...

Fixed in e3879d9.

PR #3460, opened 2020-01-29
lightningd: fix reorg bug where we don't fire watches.

rustyrussell:
Note to the reader: this is a classic bug, and yes, it was me who made it (and realized what it was once I finally tracked it down, since I've made it before!).  Learn from my mistakes: comparing with a freed pointer is always dangerous, even if you think it's OK!

If the same memory gets reallocated, our "has the tip changed?" test
gets a false negative.  This happened for me about one time in 10,
causing tests/test_misc.py::test_funding_reorg_remote_lags to fail.

Changelog-None

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -585,7 +585,7 @@ void topology_add_sync_waiter_(const tal_t *ctx,
 /* Once we're run out of new blocks to add, call this. */
 static void updates_complete(struct chain_topology *topo)
 {
-	if (topo->tip != topo->prev_tip) {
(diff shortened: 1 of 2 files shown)


cdecker:
> Okay, but why was it not caught by valgrind?

Good question, I'd suspect that re-allocating and initializing masks the access to the memory region?

rustyrussell:
Because comparing pointers is perfectly valid, and even noticing that it's a pointer comparison requires type information.

Also, since valgrind tries to avoid recycling memory for as long as possible, this bug wouldn't have manifested.

PR #3461, opened 2020-01-29
doc/PLUGINS.md: Mention `data_version` field of `db_write`, sketch ho…

ZmnSCPxj:
…w the backup plugin should work.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -514,16 +514,66 @@ It is currently extremely restricted:
    commands, as these may become intermingled and break rule #1.
 3. the hook will be called before your plugin is initialized!
 
+This hook, unlike all the other hooks, is also strongly synchronous:
+`lightningd` will stop almost all the other processing until this
+hook responds.
+
 ```json
 {
+  "data_version": 42,
   "writes": [
     "PRAGMA foreign_keys = ON"
   ]
 }
 ```
 
+This hook is intended for creating continuous backups.
+The intent is that your backup plugin maintains three
+pieces of information (possibly in separate files):
+(1) a snapshot of the database, (2) a log of database queries
+that will bring that snapshot up-to-date, and (3) the previous
+`data_version`.
+
+`data_version` is an unsigned 32-bit number that will always
+increment by 1 each time `db_write` is called.
+Note that this will wrap around on the limit of 32-bit numbers.
+
+`writes` is an array of strings, each string being a database query
+that modifies the database.
+If the `data_version` above is validated correctly, then you can
+simply append this to the log of database queries.
+
+Your plugin **MUST** validate the `data_version`.
+It **MUST** keep track of the previous `data_version` it got,
+and:
+
+1. If the new `data_version` is ***exactly*** one higher than
+   the previous, then this is the ideal case and nothing bad
+   happened and we should save this and continue.
+2. If the new `data_version` is ***exactly*** the same value
+   as the previous, then the previous set of queries was not
+   committed.
+   Your plugin **MAY** overwrite the previous set of queries with
+   the current set, or it **MAY** overwrite its entire backup
+   with a new snapshot of the database and the current `writes`
+   array (treating this case as if `data_version` were two or
+   more higher than the previous).
+3. If the new `data_version` is ***less than*** the previous,
+   your plugin **MUST** halt and catch fire, and have the
+   operator inspect what exactly happend here.
+4. Otherwise, some queries were lost and your plugin **SHOULD**
+   recover by creating a new snapshot of the database: copy the
+   database file, back up the given `writes` array, then delete
+   (or atomically `rename` if in a POSIX filesystem) the previous
+   backups of the database and SQL statements, or you **MAY**
+   fail the hook to abort `lightningd`.
+
+The "rolling up" of the database could be done periodically as well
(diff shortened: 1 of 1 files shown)


doc/PLUGINS.md
```diff
@@ -514,14 +514,49 @@ It is currently extremely restricted:
...
+  "data_version": 42,
   "writes": [
     "PRAGMA foreign_keys = ON"
   ]
 }
 ```
 
+This hook is intended for creating continuous backups.
+The intent is that your backup plugin maintains two files:
+(1) an (obsolete) copy of the database, and (2) a file of
```
cdecker:
```suggestion
(1) a snapshot of the database, and (2) a file of
```

doc/PLUGINS.md
```diff
@@ -514,14 +514,49 @@ It is currently extremely restricted:
...
+database queries that will bring that database up-to-date.
+
+`data_version` is an unsigned 32-bit number that will always
+increment by 1 each time `db_write` is called.
```
cdecker: This is technically not true, we will wrap around at 2**32-1, but for most people this is sufficient I guess.

doc/PLUGINS.md
```diff
@@ -514,14 +514,49 @@ It is currently extremely restricted:
...
+Your plugin **MUST** validate that the `data_version` is exactly 1
+higher than the previous `data_version` it was called.
+If it is not exactly that, some database queries might have gotten
+lost.
```
cdecker: We need to check for `n-1` (all good, we are in sync) and `n-2` (last transaction was not committed). If we are any other number we need to abort and kill `lightningd` (see [this logic](https://github.com/lightningd/plugins/blob/3ed5d935feafc842c702b4fa18c725d7f9311dab/backup/backup.py#L124-L155) for comparison).

doc/PLUGINS.md
```diff
@@ -514,14 +514,49 @@ It is currently extremely restricted:
...
+simply append this to the log of database queries.
+
+If `data_version` is not exactly 1 higher than the previous, but
+is still higher than the previous, then your plugin **SHOULD**
+back up a full copy of the current database, then also back up the
+`writes` array, then delete its previous backups of both the
+database and SQL statements, then change its sense of the
+`data_version` with that.
+If `data_version` is less than or equal to the previous
+`data_version`, you might do better to halt and catch fire,
```
cdecker:
```suggestion
`data_version`, you **MUST** halt and catch fire,
```

doc/PLUGINS.md
ZmnSCPxj: That seems to point to some extra behavior that should be documented as well (otherwise your backup plugin is depending on undocumented behavior...).  Do you mean that for every update made by "normal" `lightningd` operation, we increment twice, once for the update, another for its transaction being committed?

doc/PLUGINS.md
ZmnSCPxj: Added line about wraparound for unsigned 32-bit ints.

ZmnSCPxj:
Modified as per @cdecker feedback, but still need clarification regarding the "n-2" rule, since it implies particular behavior of `db_write` which we should really document otherwise future maintainers may change it due to ignorance of this behavior.

doc/PLUGINS.md
cdecker: No, each DB transaction bumps the number only once, but we can recover automatically from two cases:

 - Backup is at `n-1` and `db_write` tells us about `n`: this is the best case, we are in sync, the last transaction we wrote to the backup was successfully committed to the database, and we are continuing normally.
 - Backup is at `n-1` and `db_write` tells us about `n-1`: we were told about a DB transaction (`n-1`) during a prior run, but the DB transaction was not committed (remember we get told after the last statement, but before the tx gets committed). This can happen due to a crash, the DB not being ready, or a user killing `lightningd`. In this case we should roll back our `n-1`th transaction and apply the one we were just told about.

So really all we care about is that `n-1`, n-2`, and `n` are distinct from each other, and that there is no prior version that could be mistaken for any of these.

If we wanted to exaggerate we could also create a hash-chain: each DB transaction gets hashed, and hashes get accumulated over the entire lifetime of the DB. Thus each hash represents a unique state in the DB, and by remembering the last couple of hashes we can uniquely identify where we are in the backup. The numeric value of `data_version` is nothing more than a short-hand for that hash :wink:

The important thing remains that we can only recover automatically in one of the two situations above.

cdecker:
> Modified as per @cdecker feedback, but still need clarification regarding the "n-2" rule, since it implies particular behavior of `db_write` which we should really document otherwise future maintainers may change it due to ignorance of this behavior.

My `n-2` comment might have been more confusing than helpful, see my comment above for a case-by-case analysis. I also found that my rollback logic on the backup plugin was broken, so I'll have to fix that, and hope it didn't confuse anybody.

doc/PLUGINS.md
ZmnSCPxj: Ah, thanks for the clarification, will add this text as well.

ZmnSCPxj:
Modified and itemized exactly what to do for various conditions around `data_version`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3461/commits/f63707ec9ba9e627275381049543e7bfe23993c1

PR #3464, opened 2020-01-29
Add CONFIGURATOR_WRAPPER

NickeZ:
This was necessary for cross-compiling. Otherwise the configurator will be compiled for the host architecture.

Here are my scripts for cross compiling with docker, if you want to reproduce:

https://github.com/NickeZ/bitbox-base/commit/cf7c4e851638d547891351bf57dca8197310239e

Thanks

### DIFF
--- a/ccan/README
+++ b/ccan/README
@@ -1,3 +1,3 @@
 CCAN imported from http://ccodearchive.net.
 
-CCAN version: init-2488-g7aac849a
+CCAN version: init-2491-ga1f9c169
--- a/ccan/ccan/bitmap/bitmap.h
+++ b/ccan/ccan/bitmap/bitmap.h
@@ -21,7 +21,7 @@ typedef unsigned long bitmap_word;
 /*
  * We wrap each word in a structure for type checking.
  */
-typedef struct {
+typedef struct bitmap {
 	bitmap_word w;
 } bitmap;
 
--- a/ccan/ccan/json_out/json_out.h
+++ b/ccan/ccan/json_out/json_out.h
@@ -87,11 +87,11 @@ bool json_out_end(struct json_out *jout, char type);
  * If the resulting string requires escaping, and @quote is true, we
  * call json_escape().
  */
+PRINTF_FMT(4,5)
 bool json_out_add(struct json_out *jout,
 		  const char *fieldname,
 		  bool quote,
-		  const char *fmt, ...)
-	PRINTF_FMT(4,5);
+		  const char *fmt, ...);
 
 /**
  * json_out_addv - add a formatted member (vararg variant)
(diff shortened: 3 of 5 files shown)


NickeZ:
After reading the code and some old PRs some more I realize that I might have misunderstood things. Am I supposed to cross compile `configurator.c` and then run it in QEMU?

cdecker:
> After reading the code and some old PRs some more I realize that I might have misunderstood things. Am I supposed to cross compile `configurator.c` and then run it in QEMU?

I'm not sure tbh, I think the idea behind the configurator is to determine the target architecture capabilities. Running it on the host machine is unlikely to return the correct configuration (though most archs don't differ too much afaik). I think running it with the target configuration (natively or through QEMU) is the intended use, but @rustyrussell can likely say more to this :-)

NickeZ:
> I'm not sure tbh, I think the idea behind the configurator is to determine the target architecture capabilities. Running it on the host machine is unlikely to return the correct configuration (though most archs don't differ too much afaik). I think running it with the target configuration (natively or through QEMU) is the intended use, but @rustyrussell can likely say more to this :-)

Right, other config generators typically never run machine code for that reason. You have to use hacks, like generate compiler warnings, to figure out the capabilities of the host system.

But then there needs to be some variable that specifies a wrapper with which the configuration tool should be called so that it can be run in a virtual machine.

NickeZ:
I updated this PR by adding an environment variable for wrapping the calls to `configurator` and also compile the `configurator` and all small programs it creates with `-static` to avoid the dynamic linker. This allows us to run `configurator` with `qemu-aarch64-static`.

NickeZ:
this didn't help much I think. Doing it like this requires a native aarch64 compiler for the configurator.

NickeZ:
new update: Now I build the configurator like before but I added a feature to wrap every call to the test binaries that the configurator compiles. This actually successfully compiles for me and I hope this was how the configurator was meant to be used.

NickeZ:
Actually, this might not even be necessary. On modern debian (and I guess most distributions) it is actually possible to transparently run binaries for the wrong architecture. Somehow you can register with the kernel that it should launch binaries for certain architectures with qemu-static.

@rustyrussell Maybe for cross-compiling on windows/osx you would still need to run the binaries with some wrapper. WDYT?

remyers:
> Maybe for cross-compiling on windows/osx you would still need to run the binaries with some wrapper. WDYT?

From what I can tell, running qemu-user-static on osx is not possible. I assume qemu-user-static would work in the Windows Subsystem for Linux (WSL) on windows, but it's unclear. A docker based alternative is really useful, thanks for that. I'll give it a try soon on osx.

rustyrussell:
I like this change, but I applied it to upstream (ccan/) instead, and then pulled that back in and used your ./configure patch.

Will apply once Travis acks, thanks!

Ack: https://github.com/ElementsProject/lightning/commit/65a72d61c045b242e78eb37470339919602d34e0

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3464/commits/65a72d61c045b242e78eb37470339919602d34e0

PR #3465, opened 2020-01-29
wallet: set nLockTime to the tip for withdrawal transactions

darosior:
This sets the nLockTime to the tip (and accordingly each input's nSequence to
0xfffffffe) for withdrawal transactions.

Even if the anti fee-sniping argument might not be valid until some time yet,
this makes our regular wallet transactions far less distinguishable from
bitcoind's ones since it now defaults to using native Segwit transactions
(like us). Moreover other wallets are likely to implement this (if they
haven't already).

Changelog-Added: wallet: withdrawal transactions now sets nlocktime to the current tip.

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -698,3 +698,44 @@ def test_fundchannel_listtransaction(node_factory, bitcoind):
 
     tx = [t for t in txs if t['hash'] == txid][0]
     assert tx['blockheight'] == 0
+
+
+def test_withdraw_nlocktime(node_factory):
+    """
+    Test that we don't set the nLockTime to 0 for withdrawal transactions.
+    """
+    l1 = node_factory.get_node(1)
+    l1.fundwallet(10**4)
+
+    addr = l1.rpc.newaddr()["bech32"]
+    tx = l1.rpc.withdraw(addr, 10**3)["tx"]
+    nlocktime = node_factory.bitcoind.rpc.decoderawtransaction(tx)["locktime"]
+    tip = node_factory.bitcoind.rpc.getblockcount()
+
+    assert nlocktime > 0 and nlocktime <= tip
+
+
+@flaky
+@unittest.skipIf(VALGRIND, "A big loop is used to check fuzz.")
+def test_withdraw_nlocktime_fuzz(node_factory, bitcoind):
+    """
+    Test that we eventually fuzz nLockTime for withdrawal transactions.
+    Marked flaky "just in case" as we fuzz from 0 to 100 with a 10%
+    probability.
+    """
+    l1 = node_factory.get_node(1)
+    l1.fundwallet(10**8)
+
+    for i in range(100):
+        addr = l1.rpc.newaddr()["bech32"]
+        withdraw = l1.rpc.withdraw(addr, 10**3)
+        bitcoind.generate_block(1)
+        l1.daemon.wait_for_log('Owning output .* txid {} CONFIRMED'.
+                               format(withdraw["txid"]))
+        decoded = bitcoind.rpc.decoderawtransaction(withdraw["tx"])
+        tip = node_factory.bitcoind.rpc.getblockcount()
+        assert decoded["locktime"] > 0
+        if decoded["locktime"] < tip:
+            return
+    else:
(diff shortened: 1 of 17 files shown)


wallet/walletrpc.c
```diff
@@ -278,6 +280,17 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
+		 * benefits:
+		 * - anti fee-snipping (even if not yet likely)
+		 * - less distinguishable transactions (with this we create
+		 *   general-purpose transactions which looks like bitcoind:
+		 *   native segwit, nlocktime set to tip, and sequence set to
+		 *   0xFFFFFFFE by default. Other wallets are likely to implement
+		 *   this too).
+		 * FIXME: Do we want to also fuzz this like bitcoind does ?
+		 */
+		locktime = cmd->ld->topology->tip->height;
```
kristapsk: Both Bitcoin Core and Electrum does not always set it to current block height, sometimes it randomly sets it to older one. I also recently implemented the same policy for non-CJ direct send transaction in JoinMarket. I think it would be better for c-lightning to do the same. Formula is to with 10% probablity do "locktime = max(0, locktime - random(0, 99))", otherwise just use current height.

Apart from this ACK and it's in any case better than setting it always to 0, even without random factor.

rustyrussell: Yep, agreed, let's do the whole thing.

rustyrussell: Also, a test would be good (though I note that the randomness makes testing a bit harder!).

darosior: Will add

wallet/walletrpc.c
```diff
@@ -288,9 +289,11 @@ static struct command_result *json_prepare_tx(struct command *cmd,
 		 *   native segwit, nlocktime set to tip, and sequence set to
 		 *   0xFFFFFFFE by default. Other wallets are likely to implement
 		 *   this too).
-		 * FIXME: Do we want to also fuzz this like bitcoind does ?
 		 */
 		locktime = cmd->ld->topology->tip->height;
+		/* Eventually fuzz it too. */
+		if (pseudorand(10) == 0)
+			locktime -= (u32)pseudorand(100);
```
ZmnSCPxj: No range check for negative block heights?  Though that may be overly paranoid at this point for Bitcoin, imagine a new blockchain (bleah!) that supports Lightning from genesis...  Naaaa keep this.

tests/test_wallet.py
```diff
@@ -698,3 +698,44 @@ def test_fundchannel_listtransaction(node_factory, bitcoind):
...
+def test_withdraw_nlocktime_fuzz(node_factory, bitcoind):
+    """
+    Test that we eventually fuzz nLockTime for withdrawal transactions.
+    Marked flaky "just in case" as we fuzz from 0 to 100 with a 10%
+    probability.
+    """
+    l1 = node_factory.get_node(1)
+    l1.fundwallet(10**8)
+
+    for i in range(20):
```
ZmnSCPxj: Suggest bumping to 100, and mark this as `!VALGRIND` to avoid slowness.

darosior: Done.

tests/test_wallet.py
```diff
@@ -698,3 +698,44 @@ def test_fundchannel_listtransaction(node_factory, bitcoind):
...
+        withdraw = l1.rpc.withdraw(addr, 10**3)
+        bitcoind.generate_block(1)
+        l1.daemon.wait_for_log('Owning output .* txid {} CONFIRMED'.
+                               format(withdraw["txid"]))
+        decoded = bitcoind.rpc.decoderawtransaction(withdraw["tx"])
+        tip = node_factory.bitcoind.rpc.getblockcount()
+        assert decoded["locktime"] > 0
+        if decoded["locktime"] < tip:
+            return
+    else:
```
ZmnSCPxj: strange indent (why did not Python catch it though...?).

darosior: It's an else clause for the `for` ;)

darosior: If I got through the 100 cases then I get into the `else` clause

ZmnSCPxj: Ah, how strange.

PR #3467, opened 2020-01-29
travis: Actually build for arm

### LINKED ISSUE #3466: potential typo in CI script
Unless I'm blind you are checking the same condition twice here: https://github.com/ElementsProject/lightning/blob/038ee288a62dc225c7ed60bb670af7e474946838/.travis/build.sh#L46

NickeZ:
Fixes #3466

Changelog-None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,9 @@ env:
   - VALGRIND=0 ARCH=arm64v8 DEVELOPER=1 TARGET_HOST=aarch64-linux-gnu
 cache:
   directories:
-    - dependencies
+    - external/x86_64-linux-gnu
+    - external/aarch64-linux-gnu
+    - external/arm-linux-gnueabihf
 
 script:
   .travis/build.sh
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -51,39 +51,42 @@ cat > pytest.ini << EOF
 addopts=-p no:logging --color=no --force-flaky
 EOF
 
-if [ "$TARGET_HOST" == "arm-linux-gnueabihf" ] || [ "$TARGET_HOST" == "arm-linux-gnueabihf" ]
+if [ "$TARGET_HOST" == "arm-linux-gnueabihf" ] || [ "$TARGET_HOST" == "aarch64-linux-gnu" ]
 then
     export QEMU_LD_PREFIX=/usr/"$TARGET_HOST"/
     export MAKE_HOST="$TARGET_HOST"
+    export BUILD=x86_64-pc-linux-gnu
     export AR="$TARGET_HOST"-ar
     export AS="$TARGET_HOST"-as
     export CC="$TARGET_HOST"-gcc
     export CXX="$TARGET_HOST"-g++
     export LD="$TARGET_HOST"-ld
     export STRIP="$TARGET_HOST"-strip
+    export CONFIGURATION_WRAPPER=qemu-"${TARGET_HOST%%-*}"-static
 
     wget -q https://zlib.net/zlib-1.2.11.tar.gz \
-    && tar xvf zlib-1.2.11.tar.gz \
+    && tar xf zlib-1.2.11.tar.gz \
     && cd zlib-1.2.11 \
     && ./configure --prefix="$QEMU_LD_PREFIX" \
     && make \
-    && sudo make install 
+    && sudo make install
     cd .. && rm zlib-1.2.11.tar.gz && rm -rf zlib-1.2.11
 
     wget -q https://www.sqlite.org/2018/sqlite-src-3260000.zip \
-    && unzip sqlite-src-3260000.zip \
+    && unzip -q sqlite-src-3260000.zip \
     && cd sqlite-src-3260000 \
+    && automake --add-missing --force-missing --copy || true \
     && ./configure --disable-tcl --enable-static --disable-readline --disable-threadsafe --disable-load-extension --host="$TARGET_HOST" --prefix="$QEMU_LD_PREFIX" \
     && make \
-    && sudo make install 
+    && sudo make install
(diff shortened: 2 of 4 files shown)


cdecker:
Wow, that's what I call a fast turnaround from reporting to fixing this, great work @NickeZ :+1:

ACK https://github.com/ElementsProject/lightning/commit/df76ddecf232434ff511275fb6b52d75d28489e1

NickeZ:
@cdecker do you prefer that I force push or create `--fixup`s?

cdecker:
> @cdecker do you prefer that I force push or create `--fixup`s?

Both work fine, we track changes to PRs [here](http://46.101.246.115:8888/ElementsProject/lightning) when force pushing. I personally use fixups on larger PRs to make it easier on reviewers :-)

rustyrussell:
Seems like there's more work to do on building the libraries?

```
/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: skipping incompatible external/libwallycore.a when searching for -lwallycore

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: cannot find -lwallycore

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: skipping incompatible external/libsecp256k1.a when searching for -lsecp256k1

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: cannot find -lsecp256k1

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: skipping incompatible external/libbacktrace.a when searching for -lbacktrace

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: skipping incompatible external/libsodium.a when searching for -lsodium

/usr/lib/gcc-cross/aarch64-linux-gnu/7/../../../../aarch64-linux-gnu/bin/ld: cannot find -lsodium
```

NickeZ:
This will have to be rebased again when https://github.com/ElementsProject/lightning/pull/3464 is merged.

rustyrussell:
Rebased on master...

rustyrussell:
Sigh.  Still doesn't build :(

NickeZ:
I'll take a look at it. I'm a bit surprised though since it built locally for me.

NickeZ:
It seems like we were over-caching a bit when the whole directory `external` was cached. That directory actually contains source code as well which definitely shouldn't be cached.

I had to start building `x86_64` things in its own directory so that I could cache those files.

Now it seems like I dug up a bunch of issues since CI probably has used old versions of external dependencies for a while.

NickeZ:
At least armhf and aarch64 builds now :tada:. However there are some failures which I don't know how to deal with:

Could this be that libwally doesn't support gcc-4.8 or have we introduced this breaking change some other time? Pretty sure this PR doesn't introduce this regression.
```
../../../libwally-core/src/elements.c: In function ‘wally_asset_pak_whitelistproof’:
../../../libwally-core/src/elements.c:629:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (size_t i = 0; i < num_keys; ++i) {
     ^
../../../libwally-core/src/elements.c:629:5: note: use option -std=c99 or -std=gnu99 to compile your code
```

I have no idea:

```
==================================== ERRORS ====================================
_________________ ERROR at teardown of test_onchain_middleman __________________
[gw0] linux -- Python 3.7.5 /opt/python/3.7.5/bin/python3.7
request = <SubRequest 'teardown_checks' for <Function test_onchain_middleman>>
    @pytest.fixture
    def teardown_checks(request):
        """A simple fixture to collect errors during teardown.
    
        We need to collect the errors and raise them as the very last step in the
        fixture tree, otherwise some fixtures may not be cleaned up
        correctly. Require this fixture in all other fixtures that need to either
        cleanup before reporting an error or want to add an error that is to be
        reported.
    
        """
        errors = TeardownErrors()
        yield errors
    
        if errors.has_errors():
            # Format a nice list of everything that went wrong and raise an exception
            request.node.has_errors = True
>           raise ValueError(str(errors))
E           ValueError: 
E           Node errors:
E            - lightningd-3: Node exited with return code 1
E           Global errors:
```

rustyrussell:
gcc 4.8 is *ancient*.  Still, let me see if we can wrangle it using more configure hacks...

Ah, libwally.  I've filed a bug https://github.com/ElementsProject/libwally-core/issues/176

NickeZ:
It seems a fix has been merged in in libwally upstream. Will wait until new release is out and then bump the dependency.

cdecker:
Since we updated libwally (#3559) I rebased this on top of `master`, let's see if it build.

NickeZ:
Looks better, but still some test that fails...

rustyrussell:
I hacked it to not assume gcc, too.  Let's see how this goes...

NickeZ:
I think it should be possible to reproduce the CI failure locally since now it's part of the tests that fail and not just compilation problems. I don't have much knowledge about lightning-c so it would take too much time for me to dig into it.

rustyrussell:
Trivial rebase

rustyrussell:
This is a complete mess, I'm assigning to @cdecker.  Perhaps it's something to do with caching, since there are complaints that "lightningd/lightningd" DNE.

cdecker:
After rebasing it appears that the build passes correctly. @NickeZ can you confirm that this is what you'd expect? If it is working I'm happy to merge it asap :-)

NickeZ:
Yeah sure, I think this PR still is valid.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3467/commits/2a838252d343a79145bab5b84d28362164057fc7

PR #3469, opened 2020-01-29
Separate external by host

NickeZ:
Ideally we would support out-of-source-tree builds. But until then we can at least separate the external dependencies per architecture.

The problem I ran into is that we want to cache external dependencies in CI. But this means that the cached files are compiled for the wrong host when we cross-compile.

This PR will compile all dependencies outside their soure trees. The "native" files will stay in `external`. cross compiled files will go into `external/<host-triplet>` like `external/aarch64-linux-gnu`.

The first time you build this PR you might need to wipe your dependencies. Since currently we are building "in-source-tree" automake will complain loudly if you try to build "out-of-source-tree" without cleaning up first.

Changelog-None

### DIFF
--- a/external/Makefile
+++ b/external/Makefile
@@ -6,6 +6,11 @@ SUBMODULES =					\
 
 ifdef BUILD
 CROSSCOMPILE_OPTS := --host="$(MAKE_HOST)" --build="$(BUILD)"
+TARGET_DIR := external/"$(MAKE_HOST)"
+TOP := ../..
+else
+TARGET_DIR := external
+TOP := ..
 endif
 
 LIBSODIUM_HEADERS := external/libsodium/src/libsodium/include/sodium.h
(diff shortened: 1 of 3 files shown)


NickeZ:
I had to bump libwally-core as well since they had a bug regarding out-of-tree builds.

external/Makefile
```diff
@@ -64,32 +75,33 @@ external/jsmn/jsmn.h: submodcheck
...
 
 external-clean:
-	$(RM) $(EXTERNAL_LIBS) external/*.la external/*.o
-	if [ -f external/libsodium/Makefile ]; then make -C external/libsodium clean; fi
-	if [ -f external/libwally-core/Makefile ]; then make -C external/libwally-core clean; fi
-	if [ -f external/ibwally-core/src/Makefile ]; then make -C external/libwally-core/src clean; fi
+	$(RM) $(EXTERNAL_LIBS) $(TARGET_DIR)/*.la $(TARGET_DIR)/*.o
+	if [ -f ${TARGET_DIR}/libsodium-build/Makefile ]; then make -C ${TARGET_DIR}/libsodium-build clean; fi
+	if [ -f ${TARGET_DIR}/libwally-core-build/Makefile ]; then make -C ${TARGET_DIR}/libwally-core-build clean; fi
+	if [ -f ${TARGET_DIR}/ibwally-core-build/src/Makefile ]; then make -C ${TARGET_DIR}/libwally-core-build/src clean; fi
```
rustyrussell: Hmm, note the (preexisting!) "ibwally-core" typo!

niftynei:
> I had to bump libwally-core as well since they had a bug regarding out-of-tree builds.

cc'ing @cdecker re: libwally update

NickeZ:
Fixed the pre-existing bug and improved the commit message. Thanks for the reviews.

cdecker:
I allowed myself to rephrase the commit message for the wally bump.

ACK https://github.com/ElementsProject/lightning/pull/3469/commits/db97a44e2fe0d09a27dd71f88394cec4c30302d5

PR #3450, opened 2020-01-28
json-rpc: Check that sendpay with mpp has all the details it needs

cdecker:
We were implicitly assuming a multi-part payment if the `msatoshi` argument
was specified, but then didn't check that we have all the pieces in place for
mpp. This adds a couple of additional checks to the arguments and makes it
more explicit what it means to do an mpp.

Fixes #3431

Reported-by: Sergei Tikhomirov <@s-tikhomirov>

PR #3470, opened 2020-01-30
lightningd: disallow `msatoshi` arg to sendpay unless exact when non-…

rustyrussell:
…MPP.

Using it with a different value to the amount sent causes a crash in 0.8.0,
which is effectively deprecating it, so let's disallow it now.

Changelog-Changed: If the optional `msatoshi` param to sendpay for non-MPP is set, it must be the exact amount sent to the final recipient.
Fixes: #3431 
Closes: #3450

### DIFF
--- a/doc/lightning-sendpay.7.md
+++ b/doc/lightning-sendpay.7.md
@@ -27,8 +27,8 @@ definite failure.
 The *label* and *bolt11* parameters, if provided, will be returned in
 *waitsendpay* and *listsendpays* results.
 
-The *msatoshi* amount, if provided, is the amount that will be recorded
-as the target payment value. If not specified, it will be the final
+The *msatoshi* amount must be provided if *partid* is non-zero, otherwise
+it must be equal to the final
 amount to the destination. By default it is in millisatoshi precision; it can be a whole number, or a whole number
 ending in *msat* or *sat*, or a number with three decimal places ending
 in *sat*, or a number with 1 to 11 decimal places ending in *btc*.
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -2601,7 +2601,7 @@ def test_partial_payment(node_factory, bitcoind, executor):
     with pytest.raises(RpcError, match=r'Already have parallel payment in progress'):
         l1.rpc.sendpay(route=r124,
                        payment_hash=inv['payment_hash'],
-                       msatoshi=1000,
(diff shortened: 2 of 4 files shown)


doc/lightning-sendpay.7.md
```diff
@@ -27,8 +27,8 @@ definite failure.
 The *label* and *bolt11* parameters, if provided, will be returned in
 *waitsendpay* and *listsendpays* results.
 
-The *msatoshi* amount, if provided, is the amount that will be recorded
-as the target payment value. If not specified, it will be the final
+The *msatoshi* amount must be provided if *partid* is non-zero, otherwise
```
ZmnSCPxj: This parameter used to be used for value randomization, where we overpay the payee, but record the exact amount indicated in the invoice in our database rather than the overpaid amount (i.e. treating the overpayment as a fee).  We have since replaced value randomization with shadow routing, also overpaying the payee, but it seems this was no longer used to override the database record.  I think it is better if our permanent records did not reveal the use of overpayment to obscure the location of the payee, but well.  In any case changing `msatoshi` to make it signal multipath is a breaking change in the previous interface.

rustyrussell: Yes, the original intent with mpp was to only override the semantic for msatoshi for this case, but since it's broken in 0.8.0 anyway :(

The original use has been largely superceded by the fact that we now save the bolt11 string with the payment.

tests/test_pay.py
```diff
@@ -2800,3 +2800,35 @@ def pay(l1, inv):
 
     # pay command should complete without error
     fut.result()
+
+
+def test_sendpay_msatoshi_arg(node_factory):
+    """sendpay msatoshi arg was used for non-MPP to indicate the amount
+they asked for.  But using it with anything other than the final amount
+caused a crash in 0.8.0, so we then disallowed it.
+    """
```
cdecker: Nit: comments should be a short single line description (<80 chars), followed by a blank line, and then a multiline description (if any) that is indented to match the `"""`, finally a `"""` on it's own on a last line. You comment would become this:

```python3
    """Verify that the msatoshi arg is checked for `sendpay`.

    The `sendpay` `msatoshi` argument was used for non-MPP to indicate the amount
    they asked for.  But using it with anything other than the final amount caused a crash
    in 0.8.0, so we then disallowed it.
    """
```

PR #3476, opened 2020-01-31
Minor fixups

vasild:
Minor changes combined in one PR to reduce the amount of github noise,
better reviewed commit by commit:

* Simplify `doc/index.rst` generation
* Use `LC_ALL=C sort` everywhere
* Don't use non-standard `\s` in regexp
* Re-record `make update-mocks`
* Use the proper format specifier for printing `u8`
* `s/listpayments/listpays/` in response message


lightningd/pay.c
```diff
@@ -235,7 +235,7 @@ json_sendpay_in_progress(struct command *cmd,
 {
 	struct json_stream *response = json_stream_success(cmd);
 	json_add_string(response, "message",
-			"Monitor status with listpayments or waitsendpay");
+			"Monitor status with listsendpays or waitsendpay");
```
darosior: Good catch !

doc/index.rst
```diff
@@ -69,3 +67,5 @@ c-lightning Documentation
    lightning-waitinvoice <lightning-waitinvoice.7.md>
    lightning-waitsendpay <lightning-waitsendpay.7.md>
    lightning-withdraw <lightning-withdraw.7.md>
+   lightningd <lightningd.8.md>
```
darosior: But `lightningd` is before `lightningd-*` in alphabetic order ?

vasild: Yes, this is why `lightningd` appears before `lightningd-config`. Notice that the earlier entries are for `lightning-...` (no trailing `d`).

darosior: Oops, tricked by the trailing `d`.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3476/commits/a9cf7d1b00d01aad6b1f4b9126abb92df40cc3d4

lightningd/pay.c
ZmnSCPxj: No, `listsendpays` is wrong in this context.  Replace with `listpays` instead!  `listsendpays` will sometimes waffle a payment as "failed", because `pay` works by brute-forcing `sendpay`s.  `listpays` filters that out and sets a payment as "ongoing" as long as `pay` is still brute-forcing it.

ZmnSCPxj: No, `listsendpays` is wrong in this context.  Replace with `listpays` instead!  `listsendpays` will sometimes waffle a payment as "failed", because `pay` works by brute-forcing `sendpay`s.  `listpays` filters that out and sets a payment as "ongoing" as long as `pay` is still brute-forcing it.

ZmnSCPxj: People have lost funds running services which use `pay` to pay to clients but monitor the success using `listsendpays`, so we should really ***not*** point to `listsendpays`, just `listpays`.

vasild: Updated in 8c88e2cf.

vasild: Updated in 8c88e2cf.

PR #3478, opened 2020-01-31
Add missing `extern` qualifiers for gcc 10

laanwj:
GCC 10 defaults to `-fno-common`. no longer automatically sharing global variable definitions, which makes it important to define them in only one place (otherwise there will be duplicate definition errors). Add `extern` qualifiers where (I think) is the best place for them.

Fixes the build on gcc 10.
(ref: see also [this kind](https://www.spinics.net/lists/devicetree-compiler/msg02959.html) of Linux kernel patches)

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3478/commits/1697c5d58147ea70737c47e06776dd0ca2b0f5d3

laanwj:
The travis failure is:
```
common/memleak.o: In function `add_backtrace':
/home/travis/build/ElementsProject/lightning/common/memleak.c:207: undefined reference to `backtrace_state'
common/memleak.o: In function `memleak_init':
/home/travis/build/ElementsProject/lightning/common/memleak.c:271: undefined reference to `backtrace_state'
collect2: error: ld returned 1 exit status
make: *** [cli/lightning-cli] Error 1
```
Strange, I'm not getting this in builds locally. But it likely means that the variable needs to be defined somewhere else (currently the main definition is in `common/daemon.c`, which seems to make sense as the declaration is in `common/daemon.h`).

laanwj:
Narrowed down the problem to the `lightning-cli` build directly depending on `common/memleak.o`, but not `common/daemon.o` which defines `backtrace_state`. Ostensibly it shouldn't because it's not a daemon.

But then, what would be the most sensible place to define `backtrace_state`?

I worked around it by moving the definition to  `memleak.c` for now, this seems to fix the issue locally.

rustyrussell:
Didn't realize we weren't building with -fno-common: that's my usual for exactly these reasons.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/5b3d2362bf24119400a8fb8491364e912b5b7c06

PR #3480, opened 2020-02-01
libplugin: use json stream helpers and add sanity tests

darosior:
This is a follow-up to #3443 (based on it and implements the first of the left TODOs).

This moves all plugins from "raw" `jout`s to using higher-level `json_stream` helpers.
This also adds sanity functional tests for libplugin plugins, which uncovered a bug.

### DIFF
--- a/ccan/ccan/opt/opt.h
+++ b/ccan/ccan/opt/opt.h
@@ -229,6 +229,15 @@ void opt_register_table(const struct opt_table *table, const char *desc);
 		      (arg), (desc))
 
 /**
+ * opt_unregister - unregister an option.
+ * @names: the names it was registered with.
+ *
+ * This undoes opt_register[_early]_[no]arg.  Returns true if the option was
+ * found, otherwise false.
+ */
+bool opt_unregister(const char *names);
+
+/**
  * opt_parse - parse arguments.
  * @argc: pointer to argc
  * @argv: argv array.
--- a/plugins/libplugin.c
+++ b/plugins/libplugin.c
@@ -591,8 +591,7 @@ handle_getmanifest(struct command *getmanifest_cmd)
 		json_add_string(params, NULL, p->hook_subs[i].name);
 	json_array_end(params);
 
-	json_add_string(params, "dynamic",
-			p->restartability == PLUGIN_RESTARTABLE ? "true" : "false");
+	json_add_bool(params, "dynamic", p->restartability == PLUGIN_RESTARTABLE);
 
 	return command_finished(getmanifest_cmd, params);
 }
--- a/plugins/libplugin.c
+++ b/plugins/libplugin.c
@@ -28,24 +28,6 @@ struct plugin_timer {
 	struct command_result *(*cb)(struct plugin *p);
 };
 
-struct out_req {
-	/* The unique id of this request. */
-	u64 id;
-	/* The command which is why we're calling this rpc. */
-	struct command *cmd;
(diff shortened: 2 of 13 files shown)


darosior:
Rebased after #3443 rebase.

darosior:
(Another bug found by sanity checks)

darosior:
Re-rebased :)

plugins/libplugin.c
```diff
@@ -533,8 +533,7 @@ handle_getmanifest(struct command *getmanifest_cmd)
 		json_add_string(params, NULL, p->hook_subs[i].name);
 	json_array_end(params);
 
-	json_add_string(params, "dynamic",
-			p->restartability == PLUGIN_RESTARTABLE ? "true" : "false");
+	json_add_bool(params, "dynamic", p->restartability);
```
ZmnSCPxj: `restartability` is of type `enum plugin_restartability`, not `bool`.  Suggest doing:

    switch (p->restartability) {
    case PLUGIN_STATIC:
        json_add_bool(params, "dynamic", false);
        break;
    case PLUGIN_RESTARTABLE:
        json_add_bool(params, "dynamic", true);
        break;
    }

darosior: oops thanks !

darosior: I'll go with adding `p->restartability == PLUGIN_RESTARTABLE`

darosior:
I'm not really happy with read after free fix..

rustyrussell:
> I'm not really happy with the read after free fix, if there are suggestions..

The "technically correct" answer is to create the infrastructure to unregister options in ccan/opt.  To do anything else risks duplicate options in the option table as we remove then re-add a plugin.

Let me create that now...

plugins/libplugin.c
ZmnSCPxj: `switch` means we can add more `enum`s later and we will get warned about this code, but I suppose nobody will add more `enum` entries to `enum plugin_restartability`.

ccan/ccan/opt/opt.h
```diff
@@ -228,6 +228,15 @@ void opt_register_table(const struct opt_table *table, const char *desc);
...
 		      (arg), (desc))
 
+/**
+ * opt_unregister - unregister an option.
+ * @names: the names it was registered with.
+ *
+ * This undoes opt_register[_early]_[no]arg.  Returns true if the option was
+ * found, otherwise false.
+ */
+bool opt_unregister(const char *names);
```
rustyrussell: BTW, CCAN updates should be done via "make update-ccan" (optional: CCAN_NEW="<new modules>".  Assumes that ccan you want is in ../ccan.

Easy to fix this post, however.

tests/test_plugin.py
```diff
@@ -791,6 +791,7 @@ def test_rpc_command_hook(node_factory):
     l1.rpc.plugin_stop('rpc_command.py')
 
 
+@flaky
```
rustyrussell: Why is this now @flaky?

ccan/ccan/opt/opt.h
darosior: I did it this way then withdrawn it because of the huge diff, will fix!

tests/test_plugin.py
darosior: Because it started failing randomly on Travis but I was not able to reproduce locally..

Btw, i can no longer restart Travis jobs.. Don't know if it's a perm error or anything

PR #3486, opened 2020-02-06
Implement `getecdh`

ZmnSCPxj:
This will eventually be a basis for a full #3118 implementation that can be placed in a plugin.

@devrandom @ksedgwic please notice change in `hsmd` interface.  I took the liberty of changing the exact message number for the `ecdh` request and response, so that you can write an alternate `hsmd` that can work with both old and new `ecdh` request types (a specific `lightningd` version will use one or the other, but you might want an alternate `hsmd` to work with older versions as well).  The difference is that the old `ecdh` request used the BOLT standard specifically (convert to DER, hash with SHA256) in the `hsmd` side, the new `ecdh` request just does convert to DER on sending back.  This is because the ECIES standard uses the X coord of the point, not the hash of the DER-compressed format, so to support ECIES, we give the point from `hsmd`.

Closes #3490

I think this is better than #3490.

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -750,8 +750,8 @@ static struct io_plan *handle_ecdh(struct io_conn *conn,
 	if (!fromwire_hsm_ecdh_req(msg_in, &point))
 		return bad_req(conn, c, msg_in);
 
-	/*~ We simply use the secp256k1_ecdh function: if ss.data is invalid,
-	 * we kill them for bad randomness (~1 in 2^127 if ss.data is random) */
+	/*~ We simply use the secp256k1_ecdh function: if privkey.secret.data is invalid,
+	 * we kill them for bad randomness (~1 in 2^127 if privkey.secret.data is random) */
 	node_key(&privkey, NULL);
 	if (secp256k1_ecdh(secp256k1_ctx, ss.data, &point.pubkey,
 			   privkey.secret.data, NULL, NULL) != 1) {
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -2079,7 +2079,8 @@ int main(int argc, char *argv[])
 	status_setup_async(status_conn);
 	uintmap_init(&clients);
 
-	master = new_client(NULL, NULL, NULL, 0, HSM_CAP_MASTER | HSM_CAP_SIGN_GOSSIP,
+	master = new_client(NULL, NULL, NULL, 0,
+			    HSM_CAP_MASTER | HSM_CAP_SIGN_GOSSIP | HSM_CAP_ECDH,
 			    REQ_FD);
 
 	/* First client == lightningd. */
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -735,6 +735,30 @@ static struct io_plan *init_hsm(struct io_conn *conn,
 						    &secretstuff.bip32)));
 }
 
+/*~ The libsecp256k1 EC Diffie Hellman function provides X and Y coordinates
+ * of the point.
+ * We want the point itself, but the official constructors for struct pubkey
+ * use DER encoding, so we use a temporary DER encoding in-memory and
+ * convert from that.
+ */
+static int ecdh_load_point(u8 *uoutput,
+			   const u8 *x, const u8 *y,
+			   void *data UNUSED)
+{
+	struct pubkey *output = (struct pubkey *) uoutput;
+	u8 der_encoding[PUBKEY_CMPR_LEN];
+	/* Copied from libsecp256k1 code, no idea why this is.  */
+	u8 version = (y[31] & 0x01) | 0x02;
+
+	der_encoding[0] = version;
+	memcpy(&der_encoding[1], x, PUBKEY_CMPR_LEN - 1);
+
+	if (pubkey_from_der(der_encoding, sizeof(der_encoding), output))
+		return 1;
+	else
+		return 0;
+}
+
 /*~ The client has asked us to extract the shared secret from an EC Diffie
  * Hellman token.  This doesn't leak any information, but requires the private
  * key, so the hsmd performs it.  It's used to set up an encryption key for the
@@ -745,7 +769,7 @@ static struct io_plan *handle_ecdh(struct io_conn *conn,
 {
 	struct privkey privkey;
 	struct pubkey point;
-	struct secret ss;
+	struct pubkey ss;
 
 	if (!fromwire_hsm_ecdh_req(msg_in, &point))
 		return bad_req(conn, c, msg_in);
@@ -753,8 +777,8 @@ static struct io_plan *handle_ecdh(struct io_conn *conn,
 	/*~ We simply use the secp256k1_ecdh function: if privkey.secret.data is invalid,
 	 * we kill them for bad randomness (~1 in 2^127 if privkey.secret.data is random) */
 	node_key(&privkey, NULL);
-	if (secp256k1_ecdh(secp256k1_ctx, ss.data, &point.pubkey,
-			   privkey.secret.data, NULL, NULL) != 1) {
+	if (secp256k1_ecdh(secp256k1_ctx, (u8 *) &ss, &point.pubkey,
+			   privkey.secret.data, &ecdh_load_point, NULL) != 1) {
 		return bad_req_fmt(conn, c, msg_in, "secp256k1_ecdh fail");
 	}
 
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -11,6 +11,8 @@ from pyln.testing.utils import (
 )
 from ephemeral_port_reserve import reserve
 
+import binascii
+import hashlib
 import json
 import os
 import pytest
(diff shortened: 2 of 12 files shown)


tests/test_misc.py
```diff
@@ -2148,3 +2148,18 @@ def test_sendcustommsg(node_factory):
...
+
+def test_getecdh(node_factory):
+    """
+    Test getecdh command.
+    """
+    l1 = node_factory.get_node()
+
+    # FIXME: Actually set the node to a specific privkey and
+    # compute the actual result with a specific point.
+    shared_secret = l1.rpc.getecdh("0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518")['shared_secret']
```
darosior: Here is a test with BOLT8 test vectors..
```patch
diff --git a/tests/test_misc.py b/tests/test_misc.py
index 851a4d749..253bcc21b 100644
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -11,6 +11,8 @@ from pyln.testing.utils import (
 )
 from ephemeral_port_reserve import reserve
 
+import binascii
+import hashlib
 import json
 import os
 import pytest
@@ -2154,12 +2156,22 @@ def test_getecdh(node_factory):
     """
     Test getecdh command.
     """
-    l1 = node_factory.get_node()
+    # Cf BOLT8 test vectors for Act one.
+    options = [{
+        "dev-force-privkey":
+        "1212121212121212121212121212121212121212121212121212121212121212",
+    }, {}]
+    l1, l2 = node_factory.get_nodes(2, opts=options)
+
+    assert (l1.info["id"] ==
+            "036360e856310ce5d294e8be33fc807077dc56ac80d95d9cd4ddbd21325eff73f7")
 
-    # FIXME: Actually set the node to a specific privkey and
-    # compute the actual result with a specific point.
-    shared_secret = l1.rpc.getecdh("0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518")['shared_secret']
+    ss = l1.rpc.getecdh("028d7500dd4c12685d1f568b4c2b5048e8534b873319f3a8daa612b469132ec7f7")
+    assert (hashlib.sha256(binascii.unhexlify(ss["shared_secret"])).hexdigest() ==
+            "1e2fb3c8fe8fb9f262f649f64d26ecf0f2c0a805a767cf02dc2d77a6ef1fdcc3")
+
+    del l1.daemon.opts["dev-force-privkey"]
+    l1.restart()
 
-    # Results in a DER-compressed point, 33 bytes (66 hex chars).
-    assert isinstance(shared_secret, str)
-    assert len(shared_secret) == 66
+    assert (l1.rpc.getecdh(l2.info["id"])["shared_secret"] ==
+            l2.rpc.getecdh(l1.info["id"])["shared_secret"])
```

ZmnSCPxj: I understand giving `l1` a fixed privkey, but maybe `l2` should be allowed a randomly-generated privkey as well?  I think it should be possible to make multiple calls to `node_factory.get_node()`.

ZmnSCPxj: Ah, I missed that you remove the forced privkey in `l1` then restart.  But it might make more sense to remove from both as well, or to have just `l1` get a fixed one then remove it and restart `l1` while `l2` gets a random one.

darosior:
l2 doesn't have a forced privkey:
```
# Privkey for l1, nothing special for l2
options = [{"dev-force-privkey": "xxx"}, {}]
```

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
Le jeudi, février 6, 2020 12:39 PM, ZmnSCPxj, ZmnSCPxj jxPCSmnZ <notifications@github.com> a écrit :

ZmnSCPxj:
Yes, saw that later.  Anyway changed to your test code, thank you very much, also some syntax tweaks to make our linter happy.

cdecker:
Are you absolutely sure it is safe to return the raw ECDH result, without hashing it? If that is not the case we could effectively be leaking our private key into a lower security domain (`lightningd`). I don't quite get why returning the raw result is even needed, since we can hash it, return the hash and then use that to seed our key generation mechanism.

Regarding naming it seems a bit strange to expose a purely internal concept (ecdh) in the name. Wouldn't `getencryptionsecret` and `getdecryptionsecret` be more descriptive?

The combination of both these issues (to much internals exposed, needless change to the key-derivation scheme) should be changed.

NACK https://github.com/ElementsProject/lightning/pull/3486/commits/00d7e780a04c138934df2147a0cf5c0b18528119

ZmnSCPxj:
> Are you absolutely sure it is safe to return the raw ECDH result, without hashing it?

ECIES uses the X coordinate of the point, which is almost the entirety of the point data (the rest is just the sign of Y, a 50/50 easily-grindable bit of information).  The interface was changed this way solely to support ECIES.  If this is considered insecure, we would have to either drop support for ECIES outright, or implement one or more ECIES key derivation functions in `hsmd`.  The latter is also undesirable, as `lightningd` is not intended to be a full cryptosuite.  So if you maintain this NACK, we will probably just have to drop support for ECIES.

ZmnSCPxj:
For that matter: I can compute the ECDH of any node by just generating my own scalar and multiplying by the node ID.  That is mathematically the same as me handing the public key of that scalar to `getecdh`.  If *that* result can somehow let me learn the private key behind the node ID of the node executing `getecdh`, then I can also do the same by generating my own scalar and multiplying by your node ID somehow, meaning I can derive your node privkey just by knowing your node ID, which we assume is impossible.

As to naming it `getecdh`: this is not a complete cryptosuite, this only gives a single primitive, the ECDH.  Calling it `getencryptionsecret` would be misleading.  We could expose the existing `hsm_ecdh_req`, but that eliminates support for ECIES, and ECIES *is* a standard.

Saibato:
@cdecker depends on how often you sha and derive from that sha\s new secrets from even more secret secrets.
 
In general one could say would you prefer a sequence of messages  s(n) = enc(sk,m)  where sk = my_precious_secret  ,  some known k and primes < 2 ttpo 256 where p(1)*..*(p(i) < 2 ttpo 256
and msg(n) == (k(n) ttpo s(n)) mod (p1 .. pi) . in essence sha/

would you recommend to use such a sequence? and would known moduli with fast_gcd or groover  reveal something about sk or s(n) ? 

I see your point and it seams in line with how c/ln sec model is viewed by you and hsm should be as closed as possible.
But would you not also say, that sha is a cryptographic primitive with some unknown 
inner property.  And also in general in sight of groover any 128 assumption must be viewed as less than 64.  So sha/3 ? Maybe, Maybe not i am no cryptographer.     

It's a bit like a mine field and math theater, but often it was a good advice to not roll out your own crypto.       

@ZmnSCPxj 
> (Just to be clear, I am not God and I have no aspiration to uplift my cognition until I reach a level approaching such a being)

Just to be clear, if you ever feel such ambitions you have at least my ACK to use such powers wisely and hopefully share the results with us.
For now, some polymers and a laser like tuneable  non linear wave guide is enough on the training ranch.

jonasnick:
Drive by comment, but what seems to be missing in this discussion is that there's a reason why libsecp discourages you from doing this and why the notion of Diffie-Hellman usually includes the hashing step (and every sane standard does). The X-coordinate is not uniformly distributed because only half of the 256 bit strings are a valid X-coordinates on secp256k1. Therefore, it can't be used as a key in general and calling something ECDH that doesn't produce a key is misleading. Is there a reason for not having an RPC that does ECIES directly?

cdecker:
I'd like to echo @jonasnick's comment and add that it is preferable to have an API that cannot be used in an insecure manner than to maximize the potential use-cases, but also exposing user to potential misuse.

If we were to find additional use-cases we should rather add them as separate entry-points, it's not like we can get away with an `hsmd` that doesn't have a hash function anyway :-)

Saibato:
As always the magic t\s, timing timing timing . :sweat_smile: 
Suppose hash is something like a(trunc(log n/log p)) ottp n mod p, lets hope nobody is God like and know f(a) and p  or could do QFT

I doubt the three letters have ever figured out or when then lately, why the bumpers at the front door GCHQ look like dalek and  the door like the talis?  ooops
What works is POW, so whatever we do in signing or crypto must follow that path, to salt has also its limits and if Mallory is a Timelord, hold my beer.

hsmd/hsmd.c
```diff
@@ -735,6 +735,30 @@ static struct io_plan *init_hsm(struct io_conn *conn,
...
+{
+	struct pubkey *output = (struct pubkey *) uoutput;
+	u8 der_encoding[PUBKEY_CMPR_LEN];
+	/* Copied from libsecp256k1 code, no idea why this is.  */
+	u8 version = (y[31] & 0x01) | 0x02;
+
+	der_encoding[0] = version;
+	memcpy(&der_encoding[1], x, PUBKEY_CMPR_LEN - 1);
+
+	if (pubkey_from_der(der_encoding, sizeof(der_encoding), output))
```
rustyrussell: This copies 64 bytes (in implementation-defined order!) into the "hash function" output.

I'm pretty sure that if we do point pass-through like this, ecdh can be made to reveal the secret key by passing an x point of "1".

IIRC this was the entire rationale behind the hashing before ECDH.  I am not comfortable discarding it.

ZmnSCPxj:
@jonasnick: if we followed that principle, then we would have disabled access to `fundchannel_start` and `listsendpays`, both of which we know have caused our users to ***lose money***.  Low entropy on the X coordinate is relatively banal compared to ***financial loss***.  We cannot disable access to `fundchannel_start` and `listsendpays` because plugins (which provide safe interfaces that wrap those interfaces, like `fundchannel` and `listpays`) need to access those lower-level commands. `getecdh` is intended to be a similar low-level command that will be later be wrapped by a plugin that provides safer key-stretching and encryption on top of `getecdh`.  We would have to revamp our command structure so that some commands are accessible only to some plugins if we were to follow that principle.

hsmd/hsmd.c
rustyrussell: > This copies 64 bytes (in implementation-defined order!) into the "hash function" output.

This is true, but it really is a struct pubkey, just passed through secp256k1_ecdh in a u8 *.  It's fine.

> I'm pretty sure that if we do point pass-through like this, ecdh can be made to reveal the secret key by passing an x point of "1".
> 
> IIRC this was the entire rationale behind the hashing before ECDH. I am not comfortable discarding it.

Nope, this hashing is on the output, not the input.  It's fine, though a bit weird, yes.

real-or-random:
Putting the specific use case here aside, if there's need for an ECIES implementation in the Bitcoin ecosytem, then we should think about providing one in libsecp256k1. I can't speak for the other contributors but personally I think this is realistic given that we already expose ECDH. This would avoid that everybody needs to write their own ECIES code.

Issue #3484, opened 2020-02-04
Androids dreaming of stormy weather


cdecker:
There have been a couple of attempts to get c-lightning running on Android devices, and I thought it might be a good idea to pull experiences together to see what can be done to make Android support a possibility and make the process of getting it running on Android simpler.

I know that the abcore team (@greenaddress and @lvaccaro) has done some preliminary work to get c-lightning to compile on Android (https://github.com/greenaddress/bitcoin_ndk/pull/21). The approach they followed is:

 - Patch the makefile, the endiannes library and the JSON-RPC
 - Generate the `config.vars` and `gen_header_versions.h` on device, since the `configurator` runs on the compiler host
 - Finally compile the dependencies and `lightningd` using `clang` for the target arch

Ideally we'd backport the patches they needed to apply in order to internalize that part. @NickeZ on the other hand has been tackling the need to pre-generate the target `config.vars` and `gen_header_versions.h` files in https://github.com/ElementsProject/lightning/pull/3464. That'd allow us to compile the `configurator` for the target arch and run it using the QEMU wrapper. Hopefully that eliminates the need for pre-canned "generated files".

Finally @remyers is also working on getting Android support working, with a host `configurator` but otherwise cross-compiling for ARM: https://github.com/ElementsProject/lightning/issues/3473

Ideally we'd pool our collective experiences in getting the cross-compiled flow working, and document it for others to find in future.

What do you guys think? Are there any stumbling stones that I am missing? Could this allow us to cross-compile reproducible builds as well?

Saibato:
Nice, i.e. i did once run c/ln compiled in termux on the phone from old source source that i patched step by step to reach master side by side with a full node over orbot tor proxy, stopped this effort a year or so.
The pruned full node i synced fast on a laptop and then insert the sd card in the phone.
Then i did some patches on the core node and c/ln to not fall behind and get block info even on prunded node, i.e. if you know funding unspent that's not perfect but better than nothing.
Since the power consumption is high and the old phone had a bad battery pack i stopped this and lets be honest i am sure the custodial's apps incl. BlockS. efforts would not react welcome like,  if i begin to crush there BM on there own platform before moon, so i stopped work on this.
We  should use there VC bugs improve ln and crush them later.
Hopefully others have time and can improve, and can in general reduce power consumption of c/ln !!!

darosior:
> What do you guys think?

[...]

PR #3488, opened 2020-02-06
Bitcoin backend generalization

darosior:
This standardizes our requests for Bitcoin data as a set of commands plugins can register, and makes our `bitcoind` interaction through `bitcoin-cli` the default plugin, which registers all of these commands.

This makes it possible to not rely on a bitcoin-core backend, which can help for instance for mobile integration (cf #3484).

See #3354 for some design decisions, but here is a summary:
- Separated plugins can register different calls, which allows plugins specialization (credits to @cdecker)
- Uses JSON-RPC commands instead of hooks (see #3354).
- Needs 5 commands to be registered by our plugin(s):
  -  `getchaininfo` - called at startup to get infos about our network, and about the block chain.
  -  `getrawblockatheight` - get the block at that height in hex format, along with its id.
  -  `sendrawtransaction` - self explanatory
  -  `gettxout` - get the amount and the scriptPubkey of this txo.
  -  `getfeerate` - get the feerate in **btc/kVB**, because we *love* vbytes for fee-sensitive Bitcoin applications. Params are `bitcoind`-style.

This is based on #3480 as it uses its higher-level JSONRPC helpers for plugins.

### TODO
- [x] bcli-specific functional tests
- ~~interface documentation, for helping Bitcoin L1 developers to develop a plugin to serve as a C-lightning Bitcoin backend :)~~

### Next
- Add the possibility for the Bitcoin plugin to send notifications to `lightningd` (like "got a new block")
- Add a Bitcoin backend plugin for [insert Bitcoin data source]

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -495,33 +495,6 @@ void bitcoind_estimate_fees_(struct bitcoind *bitcoind,
 	do_one_estimatefee(bitcoind, efee);
 }
 
-static bool process_sendrawtx(struct bitcoin_cli *bcli)
-{
-	void (*cb)(struct bitcoind *bitcoind,
-		   int, const char *msg, void *) = bcli->cb;
-	const char *msg = tal_strndup(bcli, bcli->output,
-				      bcli->output_bytes);
-
-	log_debug(bcli->bitcoind->log, "sendrawtx exit %u, gave %s",
-		  *bcli->exitstatus, msg);
-
-	cb(bcli->bitcoind, *bcli->exitstatus, msg, bcli->cb_arg);
-	return true;
-}
-
-void bitcoind_sendrawtx_(struct bitcoind *bitcoind,
-			 const char *hextx,
-			 void (*cb)(struct bitcoind *bitcoind,
-				    int exitstatus, const char *msg, void *),
-			 void *arg)
-{
-	log_debug(bitcoind->log, "sendrawtransaction: %s", hextx);
-	start_bitcoin_cli(bitcoind, NULL, process_sendrawtx, true,
-			  BITCOIND_HIGH_PRIO,
-			  cb, arg,
-			  "sendrawtransaction", hextx, NULL);
-}
-
 static bool process_rawblock(struct bitcoin_cli *bcli)
 {
 	struct bitcoin_block *blk;
@@ -565,6 +538,85 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
 	fatal("%s error: bad response to %s (%s), response was %.*s",
 	      p->cmd, method, reason,
 	      toks->end - toks->start, buf + toks->start);
+}
+
+/* `sendrawtransaction`
+ *
+ * Send a transaction to the Bitcoin backend plugin. If the broadcast was
+ * not successful on its end, the plugin will populate the `errmsg` with
+ * the reason.
+ *
+ * Plugin response:
+ * {
+ *	"success": <true|false>,
+ *	"errmsg": "<not empty if !success>"
+ * }
+ */
+
+struct sendrawtx_call {
+	struct bitcoind *bitcoind;
+	void (*cb)(struct bitcoind *bitcoind,
+		   bool success,
+		   const char *err_msg,
+		   void *);
+	void *cb_arg;
+};
+
+static void sendrawtx_callback(const char *buf, const jsmntok_t *toks,
+			       const jsmntok_t *idtok,
+			       struct sendrawtx_call *call)
+{
+	const jsmntok_t *resulttok, *successtok, *errtok;
+	bool success = false;
+
+	resulttok = json_get_member(buf, toks, "result");
+	if (!resulttok)
+		bitcoin_plugin_error(call->bitcoind, buf, toks,
+				     "sendrawtransaction",
+				     "bad 'result' field");
+
+	successtok = json_get_member(buf, resulttok, "success");
+	if (!successtok || !json_to_bool(buf, successtok, &success))
+		bitcoin_plugin_error(call->bitcoind, buf, toks,
+				     "sendrawtransaction",
+				     "bad 'success' field");
+
+	errtok = json_get_member(buf, resulttok, "errmsg");
+	if (!success && !errtok)
+		bitcoin_plugin_error(call->bitcoind, buf, toks,
+				     "sendrawtransaction",
+				     "bad 'errmsg' field");
+
+	db_begin_transaction(call->bitcoind->ld->wallet->db);
+	call->cb(call->bitcoind, success,
+		 errtok ? json_strdup(tmpctx, buf, errtok) : NULL,
+		 call->cb_arg);
+	db_commit_transaction(call->bitcoind->ld->wallet->db);
+
+	tal_free(call);
+}
+
+void bitcoind_sendrawtx_(struct bitcoind *bitcoind,
+			 const char *hextx,
+			 void (*cb)(struct bitcoind *bitcoind,
+				    bool success, const char *err_msg, void *),
+			 void *cb_arg)
+{
+	struct jsonrpc_request *req;
+	struct sendrawtx_call *call = tal(bitcoind, struct sendrawtx_call);
+
+	call->bitcoind = bitcoind;
+	call->cb = cb;
+	call->cb_arg = cb_arg;
+	log_debug(bitcoind->log, "sendrawtransaction: %s", hextx);
+
+	req = jsonrpc_request_start(bitcoind, "sendrawtransaction",
+				    bitcoind->log, sendrawtx_callback,
+				    call);
+	json_add_string(req->stream, "tx", hextx);
+	jsonrpc_request_end(req);
+	plugin_request_send(strmap_get(&bitcoind->pluginsmap,
+				       "sendrawtransaction"), req);
 }
 
 /* `getrawblockbyheight`
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -781,89 +781,88 @@ void bitcoind_getchaininfo_(struct bitcoind *bitcoind,
 			    req);
 }
 
-struct get_output {
+/* `getutxout`
+ *
+ * Get informations about an UTXO. If the TXO is spent, the plugin will set
+ * all fields to `null`.
+ * {
+ *	"amount": <The output's amount in *sats*>,
+ *	"script": "The output's scriptPubKey",
+ * }
+ */
+
+struct getutxout_call {
+	struct bitcoind *bitcoind;
 	unsigned int blocknum, txnum, outnum;
 
 	/* The real callback */
-	void (*cb)(struct bitcoind *bitcoind, const struct bitcoin_tx_output *txout, void *arg);
-
+	void (*cb)(struct bitcoind *bitcoind,
+		   const struct bitcoin_tx_output *txout, void *arg);
 	/* The real callback arg */
-	void *cbarg;
+	void *cb_arg;
 };
 
-static bool process_gettxout(struct bitcoin_cli *bcli)
+static void getutxout_callback(const char *buf, const jsmntok_t *toks,
+			      const jsmntok_t *idtok,
+			      struct getutxout_call *call)
 {
-	void (*cb)(struct bitcoind *bitcoind,
-		   const struct bitcoin_tx_output *output,
-		   void *arg) = bcli->cb;
-	const jsmntok_t *tokens, *valuetok, *scriptpubkeytok, *hextok;
-	struct bitcoin_tx_output out;
-	bool valid;
+	const jsmntok_t *resulttok, *amounttok, *scripttok;
+	struct bitcoin_tx_output txout;
 
-	/* As of at least v0.15.1.0, bitcoind returns "success" but an empty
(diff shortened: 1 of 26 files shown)


darosior:
A minimal plugin which uses an esplora instance instead of `bitcoind` to serve Bitcoin data to C-lightning can be found here https://github.com/lightningd/plugins/pull/89.

ZmnSCPxj:
Spurious failure:

>      lightningd/lightningd: --bitcoin-rpcuser=rpcuser: unrecognized option

https://travis-ci.org/ElementsProject/lightning/jobs/647097656#L1083

That seems a strange error to report.

darosior:
> That seems a strange error to report

This is just that I remove the default plugin in one of the tests, to test the Bitcoin plugin registration behaviour.
But I'll give it a look anyway.

darosior:
Still one unrelated spurious Travis error, and two jobs timeout..

darosior:
I've removed the doc from the TODOs after discussion with @cdecker on IRC so that it doesn't lock up with that interface too early. I had already written it so if reviewers find it useful, you can find it [here](https://gist.github.com/darosior/92287246d26b5accb34d3381577126f0).

lightningd/bitcoind.c
```diff
@@ -38,6 +38,43 @@ static const char *methods[] = {"getchaininfo", "getrawblockbyheight",
                                 "sendrawtransaction", "gettxout",
                                 "getfeerate"};
 
+static void plugin_config_cb(const char *buffer,
```
rustyrussell: "We need our Bitcoin backend to be initialized, but the plugins have not yet been
started at this point."

(GH does not allow comments on commit msgs).

This seems like a can of worms.

OTOH, reordering to create the rpc before topology is ready is also problematic: there could innocently be an RPC call requiring topology information (plugin init, or anyone else) and the only sane semantic is to block at that point until we have it.

Can we create the rpc file, but not service it until we have topology?  That puts a burden on any `bitcoind` plugin not to deadlock, but it's simple?

lightningd/bitcoind.c
```diff
@@ -571,6 +544,84 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
+				     "sendrawtx",
+				     "bad 'result' field");
+
+	successtok = json_get_member(buf, resulttok, "success");
+	if (!successtok || !json_to_bool(buf, successtok, &success))
+		bitcoin_plugin_error(call->bitcoind, buf, toks,
+				     "sendrawtx",
+				     "bad 'success' field");
+
+	errtok = json_get_member(buf, resulttok, "errmsg");
```
rustyrussell: Prefer not to require errtok unless !success.

lightningd/bitcoind.c
```diff
@@ -784,89 +784,88 @@ void bitcoind_getchaininfo_(struct bitcoind *bitcoind,
 			    req);
 }
 
-struct get_output {
+/* `gettxout`
```
rustyrussell: Can we call this `getutxout` to emphasize that it's only UTXOs?

lightningd/bitcoind.c
```diff
@@ -1149,6 +1152,7 @@ struct bitcoind *new_bitcoind(const tal_t *ctx,
 {
 	struct bitcoind *bitcoind = tal(ctx, struct bitcoind);
 
+	strmap_init(&bitcoind->pluginsmap);
```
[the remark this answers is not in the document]
darosior: We want to be able to make distinct plugins register different commands.

lightningd/bitcoind.c
darosior: Ok, I had some edge cases when I tried but now it's working fine.

I'm still not sure about requiring the RPC socket to be initialized in `libplugin`, especially for plugins ""under"" `lightningd`.

plugins/libplugin.h
```diff
@@ -212,7 +212,8 @@ struct command_result *timer_complete(struct plugin *p);
  */
 struct plugin_timer *plugin_timer(struct plugin *p,
 				  struct timerel t,
-				  struct command_result *(*cb)(struct plugin *p));
+				  void (*cb)(void *cb_arg),
+				  void *cb_arg);
```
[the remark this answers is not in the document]
darosior: I'm happy to do it, but I wasnt sure which one to pick since the two callers don't use the same type ?

lightningd/bitcoind.c
darosior: Done.

lightningd/bitcoind.c
darosior: Hmmm I'll have to make the plugins initialization synchronous then

darosior: > but not service it until we have topology

Ok I missed this part.

darosior:
> Only one significant change req, really: split rpc-file creation from rpc servicing, and do creation before starting plugins.

Ok so I tempted but there is a circular dependency which causes a deadlock.
If we want the JSONRPC interface to be set up before topology but not serve requests until topology is setup, we can simply:
```patch
diff --git a/lightningd/jsonrpc.c b/lightningd/jsonrpc.c
index d2f896904..15a2df66a 100644
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -984,6 +984,13 @@ static struct io_plan *jcon_connected(struct io_conn *conn,
 static struct io_plan *incoming_jcon_connected(struct io_conn *conn,
                                               struct lightningd *ld)
 {
+       /* The topology is required to be ready by most of the JSONRPC calls,
+        * even the simplest ones (`getinfo` queries the tip). However we
+        * want to serve the RPC socket before we setup the topology (mostly
+        * for plugins), so accept incoming requests but don't treat them yet. */
+       if (!ld->topology->ready)
+               return io_wait(conn, ld->jsonrpc, incoming_jcon_connected, ld);
+
        /* Lifetime of JSON conn is limited to fd connect time. */
        return jcon_connected(notleak(conn), ld);
 }
index 0874bee54..594383fff 100644
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -855,6 +855,11 @@ int main(int argc, char *argv[])
        setup_topology(ld->topology, ld->timers,
                       min_blockheight, max_blockheight);
 
+       /* Now that the topology is set up, we can treat incoming JSONRPC
+        * requests ! */
+       assert(ld->topology->ready);
+       io_wake(ld->jsonrpc);
+
```
But we need to start plugins in order to setup the topology.... And plugins use `libplugin` which requires the JSONRPC to answer a request (`listconfigs` in order to set the `allow_deprecated_apis` global) before responding to `init`.

So either we get rid of this call in `handle_init`, which I believe is not acceptable, or we setup the plugins needed to setup the topology before the RPC socket (and other current plugins).

I think the right thing would be to make a distinction between plugins which rely upon `lightningd` (and thus need the RPC socket, are setup after the topology, etc..), and plugins which `lightningd` relies upon.
That said, with this PR the only plugin(s) `lightningd` relies on are `bitcoind`-ones, so it's special-cased into `lightningd/bitcoind`(and this is cleaner than all my attempts to create the socket before starting these plugins).

darosior:
Rebased, renamed `gettxout` to `getutxout`, removed the requirement for `errmsg` to be present if `sendrawtransaction` succeeds, and generalized `find_plugin_for_command()`.

The doc is ready but I wanted @cdecker feedback before as he had a reserve about documenting too early.

rustyrussell:
Yeah, OK, let's do this for now then.  Nasty stuff though :(

rustyrussell:
Trivial rebase.  Still needs a Changelog entry though!

darosior:
Rebased and changeloged

lightningd/bitcoind.c
```diff
@@ -420,59 +119,74 @@ struct estimatefee {
...
+	struct jsonrpc_request *req;
+
+	req = jsonrpc_request_start(bitcoind, "getfeerate",
+				    bitcoind->log, getfeerate_callback,
+				    call);
+	json_add_num(req->stream, "blocks", call->blocks[call->i]);
+	json_add_string(req->stream, "mode", call->estmode[call->i]);
+	jsonrpc_request_end(req);
+	plugin_request_send(strmap_get(&bitcoind->pluginsmap,
+				       "getfeerate"), req);
```
cdecker: Would be nice if we could bundle these and send all the block/mode combinations to the plugin at once. That'd allow us to have a single all-or-nothing handler and not having to loop in `lightningd`.

lightningd/bitcoind.c
```diff
@@ -80,6 +80,9 @@ void bitcoind_check_commands(struct bitcoind *bitcoind)
 	for (i = 0; i < ARRAY_SIZE(methods); i++) {
 		p = find_plugin_for_command(bitcoind->ld, methods[i]);
 		if (p == NULL) {
+			/* For testing .. */
+			log_debug(bitcoind->ld->log, "Missing a Bitcoin plugin"
+						     " command");
```
cdecker: Should probably include the method name that can't be resolved.

tests/test_plugin.py
```diff
@@ -932,3 +932,84 @@ def test_hook_chaining(node_factory):
...
+        l1.daemon.opts["plugin-dir"] = os.path.join(os.getcwd(),
+                                                    "tests/plugins/bitcoin/")
+        try:
+            l1.daemon.start()
+        except ValueError:
+            msg = "All Bitcoin plugin commands registered"
+            assert l1.daemon.is_in_log(msg)
+        else:
+            raise Exception("We registered all commands but couldn't start!")
+    else:
```
cdecker: This nesting of `try`-`except` seems strange to me.

tests/test_plugin.py
```diff
@@ -932,3 +932,84 @@ def test_hook_chaining(node_factory):
...
+    # We cant stop it dynamically
+    with pytest.raises(RpcError):
+        l1.rpc.plugin_stop("bcli")
+
+    # Failure case of feerate is tested in test_misc.py
+    assert "feerate" in l1.rpc.call("getfeerate", {"blocks": 3,
+                                                   "mode": "CONSERVATIVE"})
+
+    resp = l1.rpc.call("getchaininfo")
+    assert resp["chain"] == "regtest"
```
cdecker: This needs mocking as well since we may be running on `liquid-regtest`.

plugins/libplugin.h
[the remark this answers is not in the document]
darosior: > make sure the cb_arg matches the last argument of the cb function

Ah ok ! I'll put up a PR :-)

THanks for the issue

lightningd/bitcoind.c
darosior: This is just used in functional tests, the `fatal` below includes it

tests/test_plugin.py
darosior: I tried to test for errors, if you have a more elegant idea ? I could still put them one after the other but it's the same

tests/test_plugin.py
darosior: Ah forgot this one thanks!

lightningd/bitcoind.c
darosior: Indeed. As a cleanup or before the release ?

lightningd/bitcoind.c
```diff
@@ -556,34 +556,211 @@ void bitcoind_getrawblock_(struct bitcoind *bitcoind,
...
+	if (!blockhashtok)
+		bitcoin_plugin_error(call->bitcoind, buf, toks,
+				     "getrawblockbyheight",
+				     "bad 'blockhash' field");
+	/* If block hash is `null`, this means not found! Call the callback
+	 * with NULL values. */
+	if (json_tok_is_null(buf, blockhashtok)) {
+		db_begin_transaction(call->bitcoind->ld->wallet->db);
+		call->cb(call->bitcoind, NULL, NULL, call->cb_arg);
+		db_commit_transaction(call->bitcoind->ld->wallet->db);
```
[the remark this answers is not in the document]
darosior: I don't see why it should: `lightningd/io_loop_with_timers` only begins a db_transaction is a timer expires (out of the `io_loop`).

Before this `bcli_finished` would make a db transaction for each callback https://github.com/ElementsProject/lightning/blob/86c28b2272b65fb0926b919ea0c82327ba3051de/lightningd/bitcoind.c#L224-L226

darosior:
Spurious Travis error

lightningd/bitcoind.c
rustyrussell: Post.  Too late for release now.

cdecker:
ACK https://github.com/ElementsProject/lightning/commit/93596e36cda393aea7df9e35a240a3f34c563e7f

rustyrussell:
Apparently not :(  Mind you, GH doesn't give anything for that 'force-push' link either...

rustyrussell:
Squashed fixup:

Ack f7b48ed0879c1943a7afd7076cd9e753472895e0

cdecker:
Interestingly [the web frontend](http://46.101.246.115:8888/ElementsProject/lightning/pull/3488) seems to recognize them as identical (green background for the comparison between the last two versions). I'll look into it

lightningd/bitcoind.c
darosior: Opened https://github.com/ElementsProject/lightning/issues/3508 to track this.

PR #3489, opened 2020-02-06
Allow hook chaining for the `htlc_accepted` hook

cdecker:
This pull request creates the infrastructure for registering multiple plugins
for a given hook and showcases this new capability with the `htlc_accepted`
hook. The `htlc_accepted` hook was chosen due to its relative simple semantics
and because it currently is the hook with the most plugins making.

Before you could only use a single plugin that registers the `htlc_accepted`
hook at a time, which was severely limiting its uses. With this you can
finally run multiple at the same time.

The `htlc_accepted` hook returns one of the following three outcomes:

 - `{"result": "continue"}`: the plugin signals that it could not handle the
   HTLC internally, so the chain continues. If there are more plugins that
   have registered for this hook they get called, otherwise we handle
   internally.
 - `{"result": "fail", ...}`: the plugin decided that the HTLC should be
   rejected, with the given parameters. This exits the call chain and fails
   the HTLC. If there are more plugins they will not get called.
 - `{"result": "resolve", ...}`: similar to the previous case, the plugin
   knows the preimage and choses to resolve this HTLC. The call chain is
   exited, and no further plugins are called.

## Future steps

The following details are as of yet not addressed in this pull request, but I
plan to continue working on these in new PRs:

 - Allow plugins to provide a numeric priority which determines the order in
   which the plugins should be called
 - Allow more hooks to be chained. We currently only enable `htlc_accepted` to
   register multiple times, and we will enable more as the need arises and the
   semantics are clarified.
 - Potentially we could add a `fanout` type that calls all hooks in parallel,
   and collected the results again. One potential candidate is the `db_write`
   hook whose results can be combined into all `continue`s or at least one
   `fail`.

---

## Related issues

 - Allow plugin hook chaining (#2796)
 - Bitcoin backend plugins planning (#3354)
 - Add a `sendcustommsg` RPC call and a `custommsg` plugin hook for
   experimental protocol extensions (#3315)

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -873,7 +873,8 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
 	tal_free(request);
 }
 
-REGISTER_PLUGIN_HOOK(htlc_accepted, htlc_accepted_hook_callback,
+REGISTER_PLUGIN_HOOK(htlc_accepted, PLUGIN_HOOK_SINGLE,
+		     htlc_accepted_hook_callback,
 		     struct htlc_accepted_hook_payload *,
 		     htlc_accepted_hook_serialize,
 		     struct htlc_accepted_hook_payload *);
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -873,7 +873,7 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
 	tal_free(request);
 }
 
-REGISTER_PLUGIN_HOOK(htlc_accepted, PLUGIN_HOOK_SINGLE,
+REGISTER_PLUGIN_HOOK(htlc_accepted, PLUGIN_HOOK_CHAIN,
 		     htlc_accepted_hook_callback,
 		     struct htlc_accepted_hook_payload *,
 		     htlc_accepted_hook_serialize,
--- a/lightningd/plugin_hook.c
+++ b/lightningd/plugin_hook.c
@@ -115,10 +115,26 @@ static void plugin_hook_callback(const char *buffer, const jsmntok_t *toks,
 		tal_free(r);
 }
 
+static void plugin_hook_call_next(struct plugin_hook_request *ph_req)
+{
+	struct jsonrpc_request *req;
+	const struct plugin_hook *hook = ph_req->hook;
+	ph_req->current_plugin++;
+	assert(ph_req->current_plugin < tal_count(hook->plugins));
+	ph_req->plugin = ph_req->hook->plugins[ph_req->current_plugin];
+
+	req = jsonrpc_request_start(NULL, hook->name,
+				    plugin_get_log(ph_req->plugin),
+				    plugin_hook_callback, ph_req);
+
+	hook->serialize_payload(ph_req->payload, req->stream);
+	jsonrpc_request_end(req);
+	plugin_request_send(ph_req->plugin, req);
+}
+
 void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
 		       void *payload, void *cb_arg)
 {
-	struct jsonrpc_request *req;
 	struct plugin_hook_request *ph_req;
 	if (tal_count(hook->plugins)) {
 		/* If we have a plugin that has registered for this
@@ -130,17 +146,9 @@ void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
 		ph_req->hook = hook;
 		ph_req->cb_arg = cb_arg;
 		ph_req->db = ld->wallet->db;
-		ph_req->payload = payload;
-		ph_req->current_plugin = 0;
-		ph_req->plugin = hook->plugins[ph_req->current_plugin];
-
-		req = jsonrpc_request_start(NULL, hook->name,
-					    plugin_get_log(ph_req->plugin),
-					    plugin_hook_callback, ph_req);
-
-		hook->serialize_payload(payload, req->stream);
-		jsonrpc_request_end(req);
-		plugin_request_send(ph_req->plugin, req);
+		ph_req->payload = tal_steal(ph_req, payload);
+		ph_req->current_plugin = -1;
+		plugin_hook_call_next(ph_req);
 	} else {
 		/* If no plugin has registered for this hook, just
 		 * call the callback with a NULL result. Saves us the
--- a/lightningd/plugin_hook.c
+++ b/lightningd/plugin_hook.c
@@ -13,6 +13,7 @@ struct plugin_hook_request {
 	int current_plugin;
 	const struct plugin_hook *hook;
 	void *cb_arg;
+	void *payload;
 	struct db *db;
 };
 
@@ -129,6 +130,7 @@ void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
 		ph_req->hook = hook;
 		ph_req->cb_arg = cb_arg;
 		ph_req->db = ld->wallet->db;
+		ph_req->payload = payload;
 		ph_req->current_plugin = 0;
 		ph_req->plugin = hook->plugins[ph_req->current_plugin];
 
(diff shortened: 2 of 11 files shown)


lightningd/plugin_hook.h
```diff
@@ -44,13 +44,24 @@
...
+	/* Which type of plugin is this? It'll determine how many plugins can
+	 * register this hook, and how the hooks are called. */
+	enum plugin_hook_type type;
+
 	void (*response_cb)(void *arg, const char *buffer, const jsmntok_t *toks);
 	void (*serialize_payload)(void *src, struct json_stream *dest);
 
-	/* Which plugin has registered this hook? */
-	struct plugin *plugin;
+	/* Which plugins hav registered this hook? This is a `tal_arr`
```
darosior: typo `have`

lightningd/plugin_hook.c
```diff
@@ -100,22 +122,26 @@ void plugin_hook_call_(struct lightningd *ld, const struct plugin_hook *hook,
 {
 	struct jsonrpc_request *req;
 	struct plugin_hook_request *ph_req;
-	if (hook->plugin) {
+	if (hook->plugins != NULL && tal_count(hook->plugins) > 0) {
```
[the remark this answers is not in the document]
cdecker: It does, but better be explicit? 🤔

lightningd/plugin_hook.c
```diff
@@ -33,25 +35,46 @@ bool plugin_hook_register(struct plugin *plugin, const char *method)
...
 		return false;
-	} else if (hook->plugin != NULL) {
-		/* Another plugin already registered for this name */
-		return false;
 	}
-	hook->plugin = plugin;
+
+	/* Make sure the plugins array is initialized. */
+	if (hook->plugins == NULL)
+		hook->plugins = notleak(tal_arr(NULL, struct plugin *, 0));
```
rustyrussell: Why is this notleak() instead of allocated off hook?

lightningd/plugin_hook.c
```diff
@@ -33,25 +35,46 @@ bool plugin_hook_register(struct plugin *plugin, const char *method)
...
+	count = tal_count(hook->plugins);
+
+	for (n = 0; n < count; n++)
+		if (hook->plugins[n] == plugin)
+			break;
+
+	if (n < count)
+		tal_arr_remove(&hook->plugins, n);
+
 	return true;
```
rustyrussell: This no longer ever returns false?  I think it should do:

```
for (size_t i = 0; i < tal_count(hook->plugins); i++) {
    if (hook->plugins[i] == plugin) {
        tal_arr_remove(&hook->plugins, i);
        return true;
    }
}
return false;
```

lightningd/plugin_hook.c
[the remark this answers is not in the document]
rustyrussell: Well, you seem to assume tal_count() is always valid below....

lightningd/plugin_hook.c
```diff
@@ -193,18 +219,21 @@ void plugin_hook_db_sync(struct db *db)
...
 	void *ret;
+	struct plugin *plugin;
 
 	const char **changes = db_changes(db);
-	if (!hook->plugin)
+	if (tal_count(hook->plugins) == 0)
 		return;
 
-	ph_req = notleak(tal(hook->plugin, struct plugin_hook_request));
+	ph_req = notleak(tal(hook->plugins, struct plugin_hook_request));
```
rustyrussell: This is a subtle change: logically it would become `tal(hook->plugins[0]`. If plugin was freed, the ph_req would go with it.

lightningd/plugin_hook.c
rustyrussell: Ah, I see: hook is not a tal obj.  Fair enough.

lightningd/plugin_hook.c
```diff
@@ -118,10 +118,26 @@ static void plugin_hook_callback(const char *buffer, const jsmntok_t *toks,
 		tal_free(r);
 }
 
+static void plugin_hook_call_next(struct plugin_hook_request *ph_req)
+{
+	struct jsonrpc_request *req;
+	const struct plugin_hook *hook = ph_req->hook;
+	ph_req->current_plugin++;
+	assert(ph_req->current_plugin < tal_count(hook->plugins));
```
rustyrussell: If a plugin is freed during traversal, this assert will fail.  In general, this problem is hard: even without this, you can skip plugins due to removal at the wrong time.

lightningd/peer_htlcs.c
```diff
@@ -873,7 +873,7 @@ htlc_accepted_hook_callback(struct htlc_accepted_hook_payload *request,
 	tal_free(request);
 }
 
-REGISTER_PLUGIN_HOOK(htlc_accepted, PLUGIN_HOOK_SINGLE,
+REGISTER_PLUGIN_HOOK(htlc_accepted, PLUGIN_HOOK_CHAIN,
```
rustyrussell: From commit msg:

> Changelog-Changed: plugin: The `htlc_accepted` hook is not a chained hook. Multiple plugins can now register to handle incoming HTLCs and they will be called in order until one returns a result that isn't `{"result": "continue"}`, at which point the HTLC is considered handled.

s/not/now/.

Changelog is not a place for documentation!  How about simply:

Changelog-Changed: plugin: Multiple plugins can now register for the `htlc_accepted` hook.

ZmnSCPxj:
> Think we might now be buggy if we remove a plugin while it's supposed to be answering a hook? Request used to be freed previously, what now?

What I was planning for this was that invocation of a hook would create a new list of plugins currently registered on that hook.  Each list element would have a pointer to the plugin and a pointer to the hook invocation; the list element would be `tal`located off the hook invocation object, but would have a `tal_destructor2` triggered by detsruction of the plugin object.  If the plugin object is destroyed, we check if the hook invocation was currently calling into the plugin, and schedule the hook invocation to call into the next plugin; regardless we would then remove and `tal_free` the list element from the hook invocation list.  When hook chaining completes --- if a plugin returns non-`{'result':'continue'}` or all plugins have been invoked --- then we unregister the detructors on the plugins and destroy the hook invocation object (could be a `tal_destructor` on the hook invocation object).  This allows multiple parallel invocations of the hook (each invocation has its own list which it traverses independently of the other lists) and removal of plugins while our hook is calling into it as well.

rustyrussell:
> > Think we might now be buggy if we remove a plugin while it's supposed to be answering a hook? Request used to be freed previously, what now?
> 
...
> If the plugin object is destroyed, we check if the hook invocation was currently calling into the plugin, and schedule the hook invocation to call into the next plugin`

Even with the current scheme, we could walk all plugin_hook_request (we'd need a list somewhere), and if we're about to remove the `hook->plugins[req->current_plugin]` then we do `req->current_plugin--`.

We still need to check that we don't crap ourselves if the plugin gets removed while a hook is pending, though.  Let me try that...

rustyrussell:
OK, testing answered my question.  If a plugin is freed, the requests are abandoned.

This mean an HTLC is stuck, which is kinda nasty, but OK for now.  I've filed a separate issue #3496 to track that.

lightningd/plugin_hook.c
rustyrussell: ... but I see how this works, never mind.

cdecker:
Yes, I had noticed that as well, and was planning to address this in a separate PR (using `current_plugin--` as @rustyrussell suggested). I probably should have mentioned that in the OP to avoid having reviewers need to check.

I'll address the feedback asap and update the PR :+1: Thanks for the review @rustyrussell and @ZmnSCPxj

lightningd/plugin_hook.c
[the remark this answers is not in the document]
cdecker: Removed the extra check :-)

cdecker:
Added `fixup!`s to address the review, rebased and squashed.

You can see the changes between the version reviewed and the current version of the PR [here](http://46.101.246.115:8888/ElementsProject/lightning/pull/3489/range_diff/7367597ca7fb2a0c82441bdfd92f99e61db1f37a..d0ae53be1b2ee5614dd8d2bb2edd86d2a646054f)

cdecker:
Now having a failing assertion:

```
lightningd: lightningd/plugin_hook.c:135: plugin_hook_call_next: Assertion `ph_req->current_plugin < tal_count(hook->plugins)' failed.
```

Need to look into this.

rustyrussell:
Trivial rebase...

PR #3501, opened 2020-02-11
[DRAFT] channeld: allow fundee dip into funder reserves, once.

rustyrussell:
This is an alternative to #3500, but it requires acks from the other implementations before I'd consider releasing it into the wild.

The spec does NOT prohibit the fundee from adding an HTLC if the funder can't afford the fee, (due to races, it cannot!) though implementations try to avoid it.  We do too, but this code skips that check if this is the only HTLC we're adding, so we don't get stuck.

Not sure how other implementations will respond to this, so waiting for discussion around https://github.com/lightningnetwork/lightning-rfc/issues/728

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -363,8 +363,19 @@ static bool get_room_above_reserve(const struct channel *channel,
 	return true;
 }
 
+static size_t num_untrimmed_htlcs(enum side side,
+				  struct amount_sat dust_limit,
+				  u32 feerate,
+				  const struct htlc **committed,
+				  const struct htlc **adding,
+				  const struct htlc **removing)
+{
+	return commit_tx_num_untrimmed(committed, feerate, dust_limit, side)
+		+ commit_tx_num_untrimmed(adding, feerate, dust_limit, side)
+		- commit_tx_num_untrimmed(removing, feerate, dust_limit, side);
+}
+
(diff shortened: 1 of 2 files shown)


m-schmoock:
I think will be a much cleaner solution, as it does not add iffy fee-based magic number margins that are not 100% reliable.

ZmnSCPxj:
Let us consider the case where the channel is small enough that the fee for a single HTLC is a substantial fraction of the reserve.

In that case, the funds of the funder after the HTLC is instantiated might end up below the dust threshold, and the funder, in the latest commitment transaction(s),  has 0 funds.

The funder might now costlessly steal by publishing old state; even if punished, it has no fees in the channel anyway.

Now, we might consider that the funder, if it were to fulfill or fail the HTLC, would be able to recover the fee in a future state anyway.  So it is not accurate to say it has absolutely 0 funds --- it is simply that the funds it has are now reserved for onchain fees *in case the latest commitment transaction is dropped onchain*.

The problem is when the fundee that offered the HTLC then goes offline with the HTLC not resolved either way.  In that case, the funder now has a temptation-to-cheat.  The fact that the fundee went offline is as well evidence that increases the funder estimate that the fundee might not *ever* come online again.

So I am wary of dipping into the funder reserves for this, unless we have a check that it does not in fact cause the funder output to go below the dust threshold.

rustyrussell:
Spec consensus was for the other solution, so closing this.

PR #3512, opened 2020-02-13
db: put scids in forwards even if we didn't actually send.

rustyrussell:
If the peer is not connected, or other error which means we don't actually create an outgoing HTLC, we don't record the short_channel_id.  This is unhelpful!

Pass the scid down to the wallet code, and explicitly hand the scid and amount down to the notification code rather than handing it the htlc_out (which it doesn't need).

### DIFF
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -176,7 +176,8 @@ void notify_channel_opened(struct lightningd *ld, struct node_id *node_id,
 
 static void forward_event_notification_serialize(struct json_stream *stream,
 						 const struct htlc_in *in,
-						 const struct htlc_out *out,
+						 const struct short_channel_id *scid_out,
+						 const struct amount_msat *amount_out,
 						 enum forward_status state,
 						 enum onion_type failcode,
 						 struct timeabs *resolved_time)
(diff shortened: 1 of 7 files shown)


lightningd/notification.c
```diff
@@ -186,10 +187,16 @@ static void forward_event_notification_serialize(struct json_stream *stream,
...
 	cur->msat_in = in->msat;
-	if (out) {
-		cur->channel_out = *out->key.channel->scid;
-		cur->msat_out = out->msat;
-		assert(amount_msat_sub(&cur->fee, in->msat, out->msat));
+	if (scid_out) {
+		cur->channel_out = *scid_out;
+		if (amount_out) {
+			cur->msat_out = *amount_out;
+			assert(amount_msat_sub(&cur->fee,
```
ZmnSCPxj: In `NDEBUG` builds `assert` can remove the expression it has as argument, prefer instead `int result = foo(); assert(result);`.

rustyrussell: Yes, I didn't change this code, I only moved it.

I'll do a separate fix for post-0.8.1.

Issue #3517, opened 2020-02-13
Testnet channel 1451585x82x0 crashes routing


s-tikhomirov:
### Issue and Steps to Reproduce

I understand that this quite an edge case, still I'd like my lightningd not to crash.

I'm calling `getroute` from a plugin to find a route between `src` and `dst`. For my purposes, I put many restrictions on `getroute` for it to return the specific channel between the two nodes. Among other calls, I call `getroute` where the very channel between `src` and `dst` is excluded, and `maxhops = 1`. I expect it to either return with "No route found", or return a parallel channel between `src` and `dst` if one exists.

For one particular channel on Bitcoin testnet, I run into problems:
```
2020-02-13T12:28:22.131Z **BROKEN** gossipd: FATAL SIGNAL 6 (version v0.8.0-40-g899f5de)
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: common/daemon.c:46 (send_backtrace) 0x55fdc9118576
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: common/daemon.c:54 (crashdump) 0x55fdc91185c1
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7f82920a5f1f
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7f82920a5e97
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7f82920a7800
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7f8292097399
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7f8292097411
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: gossipd/routing.c:1223 (find_shorter_route) 0x55fdc911166c
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: gossipd/routing.c:1353 (find_route) 0x55fdc911207e
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: gossipd/routing.c:2687 (get_route) 0x55fdc91134b8
2020-02-13T12:28:22.132Z **BROKEN** gossipd: backtrace: gossipd/gossipd.c:919 (getroute_req) 0x55fdc910957f
[...]

PR #3519, opened 2020-02-14
Find shorter route assertion

rustyrussell:
Thanks to @s-tikhomirov  for the great bug report!

Fixes: #3517
Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3519/commits/992dce85f3b2369ac4e979b48efe37447bbb2de0

rustyrussell:
It would be tricky, but possible, to have a situation where we find a route with low riskfactor, and fail with high. That's what happens here.

 Imagine A with channels to B & C, both with channels to D, which goes to E. A-B is low fee, high delay. A-C is higher fee, lower delay. 

With low riskfactor we choose A-B, with high, A-C. Now, when we get to D we have different amounts, and the higher amount can't do D-E due to the max.

PR #3531, opened 2020-02-17
ccan: update to latest version which allows destructors to self-delete.

rustyrussell:
Also, a fix for compiling with NDEBUG.


ZmnSCPxj:
I am not seeing anything related to `NDEBUG` here?

rustyrussell:
If you compile with NDEBUG, the assertions in list.c get removed, and then it complains about the static function.  Making it inline removes the warning.

PR #3536, opened 2020-02-18
Removing duplicated wscript generation for channel txs

dr-orlovsky:
`struct bitcoin_tx` has a field to keep a witness script for each of its outputs (`output_witscripts`); and it is used throughout the code. However, in `channel_txs` a special array for keeping the witness scripts for all HTLC outputs is added and passed through the call stack. It appeared to me that this functionality is duplicated and contradicts the other parts of the code. After brief discussion with @cdecker today I tried to do my best in order to remove unnecessary script generation code and duplicated script objects passed around channel functions (we still have to pass the funding output witness script, since we don't pass around `bitcoin_tx` for the funding tx). 

This is my first PR, so pls don't judge if I'm done smth in a wrong way :)

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -810,14 +810,14 @@ static secp256k1_ecdsa_signature *calc_commitsigs(const tal_t *ctx,
 {
 	size_t i;
 	struct bitcoin_tx **txs;
-	const u8 **wscripts;
+	const u8 *funding_wscript;
 	const struct htlc **htlc_map;
 	struct pubkey local_htlckey;
 	const u8 *msg;
 	secp256k1_ecdsa_signature *htlc_sigs;
 
 	txs = channel_txs(tmpctx, &htlc_map,
-			  &wscripts, peer->channel, &peer->remote_per_commit,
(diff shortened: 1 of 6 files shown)


rustyrussell:
Sorry for slow review; this can be subtle and I wanted to check it carefully.  I'm on this today!

rustyrussell:
One fix: the tal() in bitcoin_redeem_2of2 was using an uninitialized pointer.  And I rebased on master (trivially).

rustyrussell:
Ack other than that...

dr-orlovsky:
Now I am sorry, lost the track of this issue. I have applied the patch and tried to do a rebase on master, but it seems that the stuff was already fixed by some other PR...

dr-orlovsky:
My bad, it was just an other change in the same lines in `mkcommit.c`. Rebases successfully and force-pushed.

PS. GitHub repo push access permissions are very complex, I was not able to give you push rights w/o inviting you to the org.

niftynei:
What do the libwally updates included in this patch encompass?

dr-orlovsky:
@niftynei hm, I have no actual idea how they have happened... Will try to figure

PR #3539, opened 2020-02-19
Handle plugins being terminated correctly

cdecker:
Due to a bug in our plugin cleanup logic we'd be waiting for both the plugin's
`stdout` and `stdin` to get closed before cleaning it up. However, since we
only poll `stdout` for incoming messages from the plugin, and not poll its
`stdin`, we'd defer cleaning up indefinitely, until we either send something
to the plugin or we shut down the node entirely. This means that we'd also not
detect crashes in a reasonable time, and a plugin crashing while handling a
hook event could hang forever.

This PR first demonstrates this issue using a plugin that exits as soon as it
`htlc_accepted` hook is called, and then proceeds to fix the issue. We take
the closing of a plugin's `stdout` as the sole signal that the plugin is
exiting and trigger the cleanup immediately. This then surfaced a number of
issue where we had memory either lingering around or the `tal_free` order
being incorrect, resulting in a number of use-after-free issues. So I had to
dive in and clean things up a bit.

In order to facilitate skipping a crashed plugin I changed the call chain for
hook events to be a list instead of an array. Each hook event now has a list
of plugins that are still to call, and we pop off elements as we receive
responses or plugins exit.

Another issue was that we'd now be falsely detecting a node shutdown during
hook calls as the plugin exiting spontaneously. To facilitate these back out
cases in future I added a state variable that indicates whether we are
operational or we are shutting down. This used to be detected through a number
of side-effects that weren't well documented (variables being set to `NULL`
etc), so making this explicit should make this clearer.

### DIFF
--- a/lightningd/plugin_control.c
+++ b/lightningd/plugin_control.c
@@ -32,6 +32,9 @@ static struct command_result *plugin_dynamic_list_plugins(struct command *cmd)
 	return command_success(cmd, response);
 }
 
+/* Mutual recursion. */
+static void plugin_dynamic_crash(struct plugin *plugin, struct dynamic_plugin *dp);
+
 /**
  * Returned by all subcommands on error.
  */
@@ -42,6 +45,8 @@ plugin_dynamic_error(struct dynamic_plugin *dp, const char *error)
 		plugin_kill(dp->plugin, "%s", error);
 	else
 		log_info(dp->cmd->ld->log, "%s", error);
+
+	tal_del_destructor2(dp->plugin, plugin_dynamic_crash, dp);
 	return command_fail(dp->cmd, JSONRPC2_INVALID_PARAMS,
 	                    "%s: %s", dp->plugin ? dp->plugin->cmd : "unknown plugin",
 	                    error);
@@ -50,6 +55,11 @@ plugin_dynamic_error(struct dynamic_plugin *dp, const char *error)
 static void plugin_dynamic_timeout(struct dynamic_plugin *dp)
 {
 	plugin_dynamic_error(dp, "Timed out while waiting for plugin response");
+}
+
+static void plugin_dynamic_crash(struct plugin *p, struct dynamic_plugin *dp)
+{
+	plugin_dynamic_error(dp, "Plugin exited before completing handshake.");
 }
 
 static void plugin_dynamic_config_callback(const char *buffer,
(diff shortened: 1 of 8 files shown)


cdecker:
I have one more cleanup, removing the plugin array from the hook call altogether and rely solely on the `call_chain` list. Working on it now.

cdecker:
Caught one more memory-leak :-)

lightningd/plugin_control.c
```diff
@@ -133,9 +144,21 @@ static struct command_result *plugin_start(struct dynamic_plugin *dp)
 	/* Give the plugin 20 seconds to respond to `getmanifest`, so we don't hang
 	 * too long on the RPC caller. */
 	p->timeout_timer = new_reltimer(dp->cmd->ld->timers, dp,
-	                                time_from_sec((10)),
+	                                time_from_sec((20)),
```
darosior: Wasn't 10 seconds long enough ?

cdecker: The documentation says 20 seconds, and this was not matching, so I adapted the value to the documented one. We can also go the other way around :-)

darosior:
I think I just forgot to update it last time i lowered the timeout ^^ But that's really a nit !

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
Le vendredi, février 21, 2020 4:46 PM, Christian Decker <notifications@github.com> a écrit :

lightningd/plugin_hook.c
```diff
@@ -89,6 +103,42 @@ void plugin_hook_unregister_all(struct plugin *plugin)
...
+	 * the plugin has not commmitted any changes internally. This is the
+	 * weakest assumption we can make short of restarting the plugin and
+	 * calling the hook again (potentially crashing the plugin the same
+	 * way again.
+	 */
+	if (link == head) {
+		/* Call next will unlink, so we don't need to. This is treated
+		 * equivalent to the plugin returning a continue-result.
+		 */
+		plugin_hook_callback(NULL, NULL, NULL, link->req);
```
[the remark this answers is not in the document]
cdecker: We get called due to `plugin_killed` being called or because we `tal_free(plugin)` which itself prints a warning to the logs and cleans up / unregisters the plugin from all hooks, hook-calls, and other things they might have registered for.

cdecker:
Rebased on top of `master`, ACK was automatically re-applied (so @bitcoin-bot  works... sometimes...)

PR #3541, opened 2020-02-20
EXPERIMENTAL_FEATURE: directed messages

rustyrussell:
This gives a minimal implementation of https://github.com/lightningnetwork/lightning-rfc/pull/748.  We simply reply to all messages with a dummy, for testing.  Obviously we'd use a plugin hook for this (and probably assign a feature bit so we know who's likely to route for us).

This is a required precursor to offers, which will use this to request invoices.


rustyrussell:
Closing: there's a better solution using an enclosed onion for the reply, which makes this stateless. Requires @cdecker's zeroed-onion trick, so worth a new PR.

PR #3554, opened 2020-02-28
fix: check source

m-schmoock:
- adds BCLI to check-source targets
- fixes two false positive check-source errors on uninitialized variable usage

`make check-source` fails on `Cppcheck 1.90` with two false positives that ignore the `NORETURN` directive on `fatal()`: 

```C
$> cppcheck wallet/wallet.c lightningd/peer_htlcs.c

Checking lightningd/peer_htlcs.c ...
lightningd/peer_htlcs.c:900:9: error: Uninitialized variable: result [uninitvar]
 return result;
        ^
Checking wallet/wallet.c ...
wallet/wallet.c:3744:4: error: Uninitialized variable: ann [uninitvar]
   ann->type = db_column_int(stmt, 8);
   ^
```

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -897,6 +897,7 @@ htlc_accepted_hook_deserialize(const tal_t *ctx,
 		      json_strdup(tmpctx, buffer, resulttok));
 	}
 
+	/* cppcheck-suppress uninitvar - false positive on fatal() above */
 	return result;
 }
 
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -62,9 +62,9 @@ $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_
 ALL_PROGRAMS += plugins/pay plugins/autoclean plugins/fundchannel plugins/bcli
 ALL_OBJS += $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS)
 
-check-source: $(PLUGIN_PAY_SRC:%=check-src-include-order/%) $(PLUGIN_AUTOCLEAN_SRC:%=check-src-include-order/%) $(PLUGIN_FUNDCHANNEL_SRC:%=check-src-include-order/%)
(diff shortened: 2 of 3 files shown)


m-schmoock:
Note: there is an unrelated test/check buid failure for:

https://travis-ci.org/ElementsProject/lightning/jobs/656193247?utm_medium=notification&utm_source=github_status
```
../../libwally-core/src/address.c: In function ‘wally_bip32_key_to_address’:
../../libwally-core/src/address.c:25:5: warning: C++ style comments are not allowed in ISO C90 [enabled by default]
     // Catch known incorrect combinations of address type and version:
     ^
../../libwally-core/src/address.c:25:5: warning: (this will be reported only once per input file) [enabled by default]
../../libwally-core/src/elements.c: In function ‘wally_asset_pak_whitelistproof’:
../../libwally-core/src/elements.c:629:5: error: ‘for’ loop initial declarations are only allowed in C99 mode
     for (size_t i = 0; i < num_keys; ++i) {
     ^
../../libwally-core/src/elements.c:629:5: note: use option -std=c99 or -std=gnu99 to compile your code
```

lightningd/peer_htlcs.c
```diff
@@ -852,6 +852,7 @@ htlc_accepted_hook_deserialize(const tal_t *ctx,
 		fatal("Plugin responded with an unknown result to the "
 		      "htlc_accepted hook: %s",
 		      json_strdup(tmpctx, buffer, resulttok));
+		abort();
```
rustyrussell: This makes the code much uglier, and I can't tell *why* you need to do this since you didn't quote the error in the commit message (and, obviously, it doesn't happen here!).

fatal() is correctly labeled NORETURN; what is failing?

m-schmoock: Hi @rustyrussell , the reason I put in these `abort()`'s was that when I currently do `make check-source` on master d9b248241 it fails with:

```C
...
lightningd/peer_htlcs.c:900:9: error: Uninitialized variable: result [uninitvar]
 return result;
        ^
wallet/wallet.c:3744:4: error: Uninitialized variable: ann [uninitvar]
   ann->type = db_column_int(stmt, 8);
   ^
make: *** [Makefile:339: check-cppcheck] Error 123
```

When we look at the source `lightningd/peer_htlcs.c:900` or `wallet/wallet.c:3744`, we see that the code checker fails recognize that the previous command `fatal("Plugin responded ...")` already aborts or terminates in any circumstance, so that the uninitialized variable access to `result` or `ann` (wallet.c) are actual false positives. I decided to put in `abort()`'s , in order to not change the mechanics in how the variables are initialized or accessed so that the code-checks remain inplace. Maybe you know a nicer way for this.

m-schmoock: Maybe this is useful to you as well:
```
$> cppcheck --version 
Cppcheck 1.90
```

m-schmoock: I will change it to a cppcheck ignore comment like this: 
`// cppcheck-suppress autoVariables`

rustyrussell: Yes, please do.

PR #3557, opened 2020-02-28
Compressible onions for rendez-vous routing

cdecker:
This is a draft of the compressible onion being discussed on the mailing list at the moment. It includes generation, compression, serialization and decompression of the compressible onions in the `devtools/onion` tool.

There are still a number of rough edges that I'll address before un-drafting this:

 - [x] Move some more of the shared secret logic into `sphinx.c`. With this I also intend to introduce a new type `compressed_onion` that can be derived from a full onion, and that can be decompressed into a full onion. Currently this is handed around as a `u8*`.
 - [x] There seems to be a stack-smashing bug somewhere, that I need to trace down.
 - [x] Some of the commits need to be rephrased.


common/sphinx.c
```diff
@@ -78,6 +87,18 @@ struct sphinx_path *sphinx_path_new_with_key(const tal_t *ctx,
...
 }
 
+bool sphinx_path_set_rendezvous(struct sphinx_path *sp,
+				const struct node_id *rendezvous_id)
+{
+	if (rendezvous_id == NULL) {
+		sp->rendezvous_id = tal_free(sp->rendezvous_id);
+		return true;
+	} else {
+		sp->rendezvous_id = tal(sp, struct pubkey);
```
rustyrussell: Potential leak here if set twice?

I think start with sp->rendezvous_id = tal_free(...) before checking if the new value is NULL?

common/sphinx.c
```diff
@@ -379,6 +454,41 @@ static void sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
...
 }
 
+static void sphinx_prefill_stream_xor(u8 *dst, size_t dstlen,
+				      struct secret *shared_secret)
+{
+	u8 padkey[KEY_LEN];
+	u8 prefill[dstlen];
+	generate_key(padkey, "prefill", 7, shared_secret);
+	generate_cipher_stream(prefill, padkey, dstlen);
+	xorbytes(dst, dst, prefill, dstlen);
```
rustyrussell: Note the new xor_cypher_stream() helper!

devtools/onion.c
```diff
@@ -172,6 +180,19 @@ static void opt_show_ad(char buf[OPT_SHOW_LEN], const u8 *assocdata)
...
+	node_id_from_hexstr(arg, strlen(arg), node_id);
+	return NULL;
+}
+
+static void opt_show_node_id(char buf[OPT_SHOW_LEN], const struct node_id *node_id)
+{
+	char *hex = node_id_to_hexstr(NULL, node_id);
+	assert(strlen(hex) < OPT_SHOW_LEN);
+	strncpy(buf, hex, OPT_SHOW_LEN);
+}
```
rustyrussell: Maybe drop the opt_show_node_id: it's used to print the default value in --help, but that's not very useful here (the key 00000000... is not very explanatory!)

common/sphinx.c
```diff
@@ -669,3 +786,24 @@ u8 *unwrap_onionreply(const tal_t *ctx,
 		return tal_free(final);
 	return final;
 }
+
+u8 *sphinx_decompress(const tal_t *ctx, const u8 *compressed,
+		      struct secret *shared_secret)
+{
+	size_t compressedlen = tal_bytelen(compressed);
+	size_t prefill_size = TOTAL_PACKET_SIZE - compressedlen;
```
rustyrussell: Beware underflow here!  Must return NULL if compressedlen > TOTAL_PACKET_SIZE!

rustyrussell: Also, if compressedlen < HMAC_SIZE.

common/sphinx.c
cdecker: Had to rebase to get it, but will use that from now on :+1:

common/sphinx.c
cdecker: Added asserts to check upper and lower size limits

cdecker:
Reworked the API for handling compressed sphinx onions by adding the following functions:

 - `serialize_compressed_onion`
 - `sphinx_decompress`
 - `sphinx_compress`
 - `sphinx_compressed_onion_deserialize`
 - `sphinx_compressed_onion_serialize`

This should allow transforming the compressed onion from/to both full onions and serialized formats.

common/sphinx.c
```diff
@@ -386,6 +462,39 @@ static void sphinx_write_frame(u8 *dest, const struct sphinx_hop *hop)
 	memcpy(dest + tal_bytelen(hop->raw_payload), hop->hmac, HMAC_SIZE);
 }
 
+static void sphinx_prefill_stream_xor(u8 *dst, size_t dstlen,
+				      struct secret *shared_secret)
```
rustyrussell: const struct secret *?

common/sphinx.c
```diff
@@ -669,3 +785,98 @@ u8 *unwrap_onionreply(const tal_t *ctx,
 		return tal_free(final);
 	return final;
 }
+
+struct onionpacket *
+sphinx_decompress(const tal_t *ctx,
+		     const struct sphinx_compressed_onion *src,
+		     struct secret *shared_secret)
```
rustyrussell: const struct secret *?

common/sphinx.c
```diff
@@ -669,3 +785,98 @@ u8 *unwrap_onionreply(const tal_t *ctx,
...
+struct sphinx_compressed_onion *
+sphinx_compress(const tal_t *ctx, const struct onionpacket *packet,
+		struct sphinx_path *path)
+{
+	struct sphinx_compressed_onion *res;
+	size_t payloads_size = sphinx_path_payloads_size(path);
+
+	/* We can't compress an onion that doesn't have a rendez-vous node. */
+	if (path->rendezvous_id)
+		return NULL;
```
rustyrussell: This is clearly untested? :(

common/sphinx.c
```diff
@@ -669,3 +785,98 @@ u8 *unwrap_onionreply(const tal_t *ctx,
...
+	memcpy(res->routinginfo, src->routinginfo, srclen);
+	sphinx_prefill_stream_xor(res->routinginfo + srclen, prefill_size,
+				  shared_secret);
+
+	return res;
+}
+
+struct sphinx_compressed_onion *
+sphinx_compress(const tal_t *ctx, const struct onionpacket *packet,
+		struct sphinx_path *path)
```
rustyrussell: const for path?

common/sphinx.c
```diff
@@ -669,3 +785,98 @@ u8 *unwrap_onionreply(const tal_t *ctx,
...
+	write_buffer(dst, der, PUBKEY_SIZE, &p);
+	write_buffer(dst, onion->routinginfo, routelen, &p);
+	write_buffer(dst, onion->mac, HMAC_SIZE, &p);
+
+	assert(p == len);
+	return dst;
+}
+
+struct sphinx_compressed_onion *
+sphinx_compressed_onion_deserialize(const tal_t *ctx, u8 *src)
```
rustyrussell: const for src?

common/sphinx.c
cdecker: Right, I replaced the `serialize_compressed_onion` method with the compress + serialize primitives I added. I also added a check to make sure non-compressible onions return `NULL`.

common/sphinx.c
```diff
@@ -669,3 +754,94 @@ u8 *unwrap_onionreply(const tal_t *ctx,
...
+	if (dst->version != 0x00)
+		return tal_free(dst);
+
+	fromwire_pubkey(&cursor, &max, &dst->ephemeralkey);
+
+	dst->routinginfo = tal_arr(dst, u8, routelen);
+	fromwire_u8_array(&cursor, &max, dst->routinginfo, routelen);
+	fromwire_u8_array(&cursor, &max, dst->mac, HMAC_SIZE);
+
+	assert(max == 0);
```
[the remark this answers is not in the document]
cdecker: Good point, I like `fromwire_tal_bytes`, however it doesn't address the full issue, since that check isn't just there to avoid allocating a huge array, but also to enforce the logic that an onion that is larger than an uncompressed onion can't possibly be a compressed onion.

I added an initial check that the total compressed onion length is smaller than the total length of an uncompressed one, no longer computing `routelen`

rustyrussell:
Ack: https://github.com/ElementsProject/lightning/commit/ebe9f9c66dec0de2f2e772a6abec06e2301192e0

Rebase onto master, rename -devtool/onion option.

rustyrussell:
Removed final commit which broke tests; we can do a separate PR for that.

PR #3560, opened 2020-03-03
Remove excessive wrapping in the `rpc_command` hook call

### LINKED ISSUE #3558: rpc_command hook passes wrong payload
According to the [documentation](https://lightning.readthedocs.io/PLUGINS.html#rpc-command) (and the standard from other hooks) `rpc_command` should pass the following payload to the plugin:

```json
{
    "rpc_command": {
        "id": 3,
        "method": "method_name",
        "params": {
            "param_1": [],
            "param_2": {},
            "param_n": "",
        }
    }
}
```

However what they do send is

```json
{
    "rpc_command": {
        "rpc_command": {
            "id": 3,
            "method": "method_name",
            "params": {
                "param_1": [],
                "param_2": {},
                "param_n": "",
            }
        }
    }
}
```

cdecker:
We were nesting like the following:

```json
{"params": {
  "rpc_command": {
    "rpc_command": {
    }
  }
}
```

This is really excessive, so we unwrap once, and now have the following:

```json
{"params": {
  "rpc_command": {
  }
}
```

Still more wrapping than necessary (the method is repeated in the `params`
object), but it's getting closer.

Suggested-by: @fiatjaf
Signed-off-by: Christian Decker <@cdecker>

Fixes #3558


lightningd/jsonrpc.c
```diff
@@ -611,8 +611,21 @@ struct rpc_command_hook_payload {
 static void rpc_command_hook_serialize(struct rpc_command_hook_payload *p,
```
darosior: I think this commit needs a Changelog-Deprecated line ?

lightningd/jsonrpc.c
```diff
@@ -611,8 +611,22 @@ struct rpc_command_hook_payload {
...
 {
+	const jsmntok_t *tok;
+	size_t i;
+	char *key;
 	json_object_start(s, "rpc_command");
-	json_add_tok(s, "rpc_command", p->request, p->buffer);
+
+#ifdef COMPAT_V081
+	if (deprecated_apis)
+		json_add_tok(s, "rpc_command", p->request, p->buffer);
```
niftynei: heh i came across this yesterday. can you remove the outer object and instead do this line? it should achieve the same result without needing to iterate through the object.

cdecker: I tried to do that, however if I use `json_add_tok` to fill in `{"rpc_command": <copy>}` I cannot re-enter the `rpc_command` key to then add the compat version, i.e., `{"rpc_command": {"rpc_command": <copy>, <copied-fields}}`. Once we remove the `COMPAT_V081` flag we can actually do this and avoid iterating through, but until then I'm afraid we need to keep this.

niftynei: ack. i understand what this outputs now. thanks.

cdecker:
Rebased and squashed :-)

PR #3563, opened 2020-03-04
wire: add fromwire_tal_bytes() helper.

rustyrussell:
Does the allocation and copying; this is useful because we can avoid being fooled into doing giant allocations.

This will be useful for #3557

Changelog-None

### DIFF
--- a/common/onionreply.c
+++ b/common/onionreply.c
@@ -12,8 +12,8 @@ struct onionreply *fromwire_onionreply(const tal_t *ctx,
 				       const u8 **cursor, size_t *max)
 {
 	struct onionreply *r = tal(ctx, struct onionreply);
-	r->contents = tal_arr(r, u8, fromwire_u16(cursor, max));
-	fromwire_u8_array(cursor, max, r->contents, tal_count(r->contents));
+	r->contents = fromwire_tal_arrn(r, cursor, max,
+					fromwire_u16(cursor, max));
 	if (!*cursor)
 		return tal_free(r);
 	return r;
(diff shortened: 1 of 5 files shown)


cdecker:
Good idea, shall we maybe call it `tal_arrn` in the same vein as `tal_arrz` and it'd match `strnlen` and `strncmp` variants in the stdlib.

niftynei:
going to wade cautiously into this naming debate and say that since it conforms to the `fromwire_` API family (takes a context, cursor, and 'max/plen') the `fromwire`_` prefix is probably preferrable?

> we can avoid being fooled into doing giant allocations.
out of curiosity, what situations do giant allocations occur?

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3563/commits/6107d74e0106b8c3d1036506ed2ef7c9d4fe4ce7

rustyrussell:
> going to wade cautiously into this naming debate and say that since it conforms to the `fromwire_` API family (takes a context, cursor, and 'max/plen') the `fromwire`_` prefix is probably preferrable?

I was assuming fromwire_tal_arrn(), yes.

> > we can avoid being fooled into doing giant allocations.
> 
> out of curiosity, what situations do giant allocations occur?

If we ever had a bigsize or u64 length, or a derived length where the calculation could underflow, we'd risk an out-of-memory crash in tal with code like this:

```
mem = tal_arr(ctx, u8, len);
fromwire_u8(mem, p, len);
```

Having a helper check that the length is < the message length before allocation sidesteps this nicely.

PR #3567, opened 2020-03-04
Return the correct error when the TLV payload was not valid

cdecker:
Fixes a minor memory leak, return the necessary information when checking the TLV payload for validity and do not return using `local_fail_in_htlc_badonion` when the real issue is `INVALID_ONION_PAYLOAD`.

Changelog-None


niftynei:
ACK 21576fcf6717247416a085c048f10373e909f4ed

PR #3561, opened 2020-03-03
Fix build

vasild:

Issue #3542, opened 2020-02-21
feature request: config option to change default fundchannel feerate


instagibbs:

instagibbs:
I get livid when I see 20x+ the satoshis going out the window than required to set up cheap payments for later ;)

darosior:
Why not using the argument directly ? Like:
```
lightning-cli fundchannel <id> <amount> 253perkw
```
Since the feerate changes quickly, I'm not sure about a static configuration option ?

instagibbs:
Picking a default block target and/or mode is what I'd want. "why don't you
just type out all arguments to avoid defaults you don't like" isn't very
persuasive.

darosior:
I'm not talking about a default block target and/or mode but a precise fee in sats perKW...

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐
Le samedi, février 22, 2020 12:59 PM, Gregory Sanders <notifications@github.com> a écrit :

> Picking a default block target and/or mode is what I'd want. "why don't you
> just type out all arguments to avoid defaults you don't like" isn't very
> persuasive.
>

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: Very old issue (6.0 years old) no recent activity

If you believe this issue should remain open, please comment and we can reopen it.

PR #3570, opened 2020-03-05
Bitcoin backend: batch fee estimation requests

### LINKED ISSUE #3508: Batching feerate requests for our plugin backend
See https://github.com/ElementsProject/lightning/pull/3488#discussion_r377740150.

We have a callback logic in `lightningd/bitcoind` to ask our backend for all targets and modes, it would be great to extend the plugin and just make all the modes and targets a JSON object `lightningd`-side.

darosior:
EDIT: the scope changed from just batching feerate requests, cf https://github.com/ElementsProject/lightning/pull/3570#issuecomment-597382006

fixes #3508


lightningd/bitcoind.c
```diff
@@ -102,98 +102,75 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
+				    "Unable to estimate %s fees",
+				    targets[i]);
 
 #if DEVELOPER
-		/* This is needed to test for failed feerate estimates
-		 * in DEVELOPER mode */
-		call->satoshi_per_kw[call->i] = 0;
+			/* This is needed to test for failed feerate estimates
+			* in DEVELOPER mode */
+			feerates[i] = 0;
```
niftynei: the better way to do this is to use a config-flag to trigger this behavior if you need it for tests. 

actually the better option would be to use a config flag for the floor for all cases, which the test can then set to zero if desired. (which removes the switch for `testnet` below, also

plugins/bcli.c
```diff
@@ -478,23 +492,89 @@ static struct command_result *process_estimatefee(struct bitcoin_cli *bcli)
...
+/* We got the response for the normal feerate, now treat the slow one. */
+static struct command_result *estimatefees_third_step(struct bitcoin_cli *bcli)
+{
+	struct command_result *err;
+	struct estimatefees_stash *stash = bcli->stash;
+	const char **params = tal_arr(bcli->cmd, const char *, 2);
+
+	/* bitcoind could theoretically fail to estimate for a higher target. */
+	if (*bcli->exitstatus != 0)
+		return estimatefees_null_response(bcli);
```
niftynei: why not return what you've gotten so far and log the anomaly?

lightningd/bitcoind.c
```diff
@@ -102,98 +102,75 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
-				     "bad 'feerate' field");
+	for (size_t i = 0; i < tal_count(feerates); i++) {
+		feeratetok = json_get_member(buf, resulttok, targets[i]);
+		if (!feeratetok)
+			bitcoin_plugin_error(call->bitcoind, buf, toks,
+					     "estimatefees",
+					     "bad 'urgent', 'normal' or 'slow'"
+					     " field");
 
-	/* FIXME: We could trawl recent blocks for median fee... */
```
niftynei: you're leaving out this fixme?? :P

lightningd/bitcoind.c
```diff
@@ -102,98 +102,75 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
-				     "getfeerate",
+				     "estimatefees",
 				     "bad 'result' field");
 
-	feeratetok = json_get_member(buf, resulttok, "feerate");
-	if (!feeratetok)
-		bitcoin_plugin_error(call->bitcoind, buf, toks,
-				     "getfeerate",
-				     "bad 'feerate' field");
+	for (size_t i = 0; i < tal_count(feerates); i++) {
```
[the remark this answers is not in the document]
niftynei: update this is fixed in #3561

lightningd/bitcoind.c
darosior: I did not modify this

plugins/bcli.c
darosior: Hmm I could log in `estimatefees_null_response()`, but the fee estimation failing is really common, moreover this one really is only a sanity check as `bitcoin-cli` would return `0` on fee estimation failure..

lightningd/bitcoind.c
darosior: Oops :)

darosior: Restored

lightningd/bitcoind.c
```diff
@@ -102,98 +102,75 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
-				const jsmntok_t *idtok,
-				struct estimatefee_call *call)
+static void estimatefees_callback(const char *buf, const jsmntok_t *toks,
+				  const jsmntok_t *idtok,
+				  struct estimatefee_call *call)
 {
 	const jsmntok_t *resulttok, *feeratetok;
-	u64 feerate;
+	const char *targets[] = {"urgent", "normal", "slow"};
+	u32 *feerates = tal_arr(call, u32, 3);
```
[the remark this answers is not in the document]
darosior: > can you just straight declare this as an array of size 3? not sure how this interacts with the cb below. it seems like it'd be fine since you clean up call at the end of this routine anyway.

I'm almost sure that it has to be allocated on the heap to be passed to the cb.

> if you use a constant for the size, you can re-use it for the for loop count below (instead of tal_count)
e.g. #define FEERATE_COUNT 3; u32 feerates[FEERATE_COUNT];

Ok, I'll use chaintopology's one then.

darosior:
> something to think about: this PR locks us into the 3 definitions of "urgent, normal, slow" in a harder to change way than the previous method of calling getfeerate 3 times. i can see how this is preferable for a plugin API.

imho it's an improvement, this classification is clearer, not backend-specific, and closer to what we use internally.

> further it canonicalizes 'urgent, normal and slow', while removing the bitcoind standard that we adhered to previously of CONSERVATIVE/2 , ECONOMICAL/100 etc. what guarantees do these new standards require?

That's not a complete removal, `bcli` does use the same standards as before. But it gives more flexibility to the plugin if it tries to be smart.

> if a plugin reports back an urgent feerate, how fast does that mean it'll be mined on average?

We completely trust the plugin for providing an urgent feerate when asked, and it doesn't change default behaviour at all as we still use CONSERVATIVE/2 for urgent feerate in `bcli`.

> by moving away from them we're opening up some possibility for interpretation. we should add some strong guidelines around the meaning of them in any bitcoin-plugin documentation

Yep, the doc is ready and yet to come once we have the final API.
Of course I shall add that if one wants to write an exotic fee-estimation plugin, it shouldn't take the `urgent` feerate too lightly as it's crucial...

lightningd/bitcoind.c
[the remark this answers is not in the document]
darosior: Done

rustyrussell:
Agree with @niftynei that these differentiations suck.  We should expose them by usage.

Currently we use the following:
* Mutual close (NORMAL)
* Opening (NORMAL)
* Unilateral (URGENT)
* Minimum acceptable (SLOW / 2)
* Maximum acceptable (URGENT * config.max_fee_multiplier)
* onchaind: HTLC to wallet, penalty txs, returning our unilateral spend to us (NORMAL).

Perhaps we should expose these directly via the API?  Frankly, returning HTLC to wallet (if timed out) should probably be more urgent, whereas the other onchaind cases should be slower.  

An even better API would allow an amount and a timeout (in number of blocks), where the amount says how much we'd lose, and the timeout says how many blocks we have.

darosior:
Ok. I think too that a more fine-grained API is preferable.

What's great is that we could even register some `dev` commands in `bcli` for setting each feerate which I guess could help for `onchaind` functional tests ?
Hmm this would also allow to set the mutual close feerate, which IIRC was a feature request.. I'm on it!

I've looked into it and propose the following feerates :
- `opening`
- `mutual_close`
- `unilateral_close` (also for the return to wallet)
- `minimum`
- `maximum`
- `htlc`
- `penalty`

I propose to also bump the `penalty` and `htlc` feerates to URGENT `bcli-side`.

darosior:
Ok so now the patchset is much bigger and breaks some APIs but it's really better this way, I think :).

So, this now changes our feerate tracking from `urgent`, `normal` and `slow` feerates to a "type-based" feerate tracking.
This also batch the fee estimation requests to the Bitcoin backend which will answer with the same estimates as before, so the estimations themselves aren't actually changed.

> I propose to also bump the penalty and htlc feerates to URGENT bcli-side.

I'll propose this as an other PR, what do you think ?
I'll also propose a PR to be able to tweak the mutual close transaction fees since it's part of the types, iirc it was a feature request ?

lightningd/chaintopology.c
```diff
@@ -534,6 +526,23 @@ static struct command_result *json_feerates(struct command *cmd,
...
+			     feerate_to_style(unilateral_feerate(cmd->ld->topology), *style));
+		/* normal feerate was ECONOMICAL/4, i.e. what bcli gives us
+		 * now for opening feerate */
+		json_add_u64(response, "normal",
+			     feerate_to_style(opening_feerate(cmd->ld->topology), *style));
+		/* slow feerate was ECONOMICAL/100, i.e. what bcli gives us
+		 * now for min feerate, but doubled (the min was slow/2 but now
+		 * the Bitcoin plugin directly gives the real min acceptable). */
+		json_add_u64(response, "slow",
+			     feerate_to_style(feerate_min(cmd->ld, NULL) * 2, *style));
```
darosior: We can even not double here ? As done in param_feerate()

lightningd/json.c
```diff
@@ -115,6 +115,15 @@ struct command_result *param_feerate(struct command *cmd, const char *name,
 		if (json_tok_streq(buffer, tok, feerate_name(i)))
 			return param_feerate_estimate(cmd, feerate, i);
 	}
+	/* We used SLOW, NORMAL, and URGENT as feerate targets previously,
+	 * and many commands rely on this syntax now.
+	 * It's also really more natural for an user interface. */
+	if (json_tok_streq(buffer, tok, "slow"))
+		return param_feerate_estimate(cmd, feerate, FEERATE_MIN);
```
darosior: We don't double it, but ECONOMICAL/100 is that far in the future and beyond that it doesn't make a big difference : we'll likely hit floor anyway.

plugins/bcli.c
```diff
@@ -449,17 +449,36 @@ static struct command_result *process_getblockchaininfo(struct bitcoin_cli *bcli
 	return command_finished(bcli->cmd, response);
 }
 
-static struct command_result *process_estimatefee(struct bitcoin_cli *bcli)
+
+struct estimatefees_stash {
+	/* FIXME: We use u64 but lightningd will store them as u32. */
+	u64 urgent, normal, slow;
+};
```
vasild: Why not have these as `u32`?

plugins/bcli.c
```diff
@@ -623,25 +713,28 @@ static struct command_result *getchaininfo(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
-/* Get current feerate.
- * Calls `estimatesmartfee` and returns the feerate as btc/k*VBYTE*.
+/* Get the current feerates. We us an urgent feerate for unilateral_close and max,
```
vasild: nit: `s/We us/We use/`

tests/test_misc.py
```diff
@@ -1354,56 +1354,75 @@ def test_feerates(node_factory):
...
+        assert t not in feerates
 
     wait_for(lambda: len(l1.rpc.feerates('perkb')['perkb']) == 2)
     feerates = l1.rpc.feerates('perkb')
     assert feerates['warning'] == 'Some fee estimates unavailable: bitcoind startup?'
     assert 'perkw' not in feerates
     assert feerates['perkb']['max_acceptable'] == (2**32 - 1)
     assert feerates['perkb']['min_acceptable'] == 253 * 4
+    for t in types:
+        assert t not in feerates
```
vasild: Shouldn't this be `assert t not in feerates['perkb']`?

lightningd/bitcoind.c
```diff
@@ -121,98 +121,94 @@ static void bitcoin_plugin_error(struct bitcoind *bitcoind, const char *buf,
...
-		bitcoin_plugin_error(call->bitcoind, buf, toks,
-				     "getfeerate",
-				     "bad 'feerate' field");
+	for (size_t i = 0; i < NUM_FEERATES; i++) {
+		feeratetok = json_get_member(buf, resulttok, estimates[i]);
+		if (!feeratetok)
+			bitcoin_plugin_error(call->bitcoind, buf, toks,
+					     "estimatefees",
+					     "bad 'urgent', 'normal' or 'slow'"
+					     " field");
```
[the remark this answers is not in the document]
darosior: Thx it also talks about legacy urgent normal, slow..

lightningd/chaintopology.c
```diff
@@ -405,21 +411,8 @@ static void update_feerates(struct bitcoind *bitcoind,
...
-			if (topo->feerate[j] < topo->feerate[i]) {
-				log_unusual(topo->log,
-					  "Feerate estimate for %s (%u) above %s (%u)",
-					  feerate_name(i), topo->feerate[i],
-					  feerate_name(j), topo->feerate[j]);
-				topo->feerate[j] = topo->feerate[i];
-			}
-		}
 		if (try_get_feerate(topo, i) != old_feerates[i])
 			changed = true;
```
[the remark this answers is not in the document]
darosior: Minimal change, otherwise i'd make this a function

lightningd/chaintopology.c
```diff
@@ -432,30 +425,43 @@ static void update_feerates(struct bitcoind *bitcoind,
...
 {
-	/* FEERATE_IMMEDIATE, FEERATE_NORMAL, FEERATE_SLOW */
 	const char *estmodes[] = { "CONSERVATIVE", "ECONOMICAL", "ECONOMICAL" };
 	const u32 blocks[] = { 2, 4, 100 };
 
-	BUILD_ASSERT(ARRAY_SIZE(blocks) == NUM_FEERATES);
 
 	/* Once per new block head, update fee estimates. */
 	bitcoind_estimate_fees(topo->bitcoind, blocks, estmodes, NUM_FEERATES,
 			       update_feerates, topo);
```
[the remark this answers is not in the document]
darosior: Thanks... i didnt self reviewed before pushing..

lightningd/onchain_control.c
```diff
@@ -502,33 +502,41 @@ enum watch_result onchaind_funding_spent(struct channel *channel,
...
-				return KEEP_WATCHING;
+	/* We try to get the feerate for each transaction type, 0 if estimation
+	 * failed. */
+	feerates[0] = delayed_to_us_feerate(ld->topology);
+	feerates[1] = htlc_resolution_feerate(ld->topology);
+	feerates[2] = penalty_feerate(ld->topology);
+	/* We check them separately but there is a high chance that if estimation
+	 * failed for one, it failed for all.. */
+	for (size_t i = 0; i < NUM_FEERATES; i++) {
+		if (!feerates[i]) {
```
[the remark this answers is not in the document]
darosior: ouch. Shouldnt have pushed that early !

lightningd/chaintopology.c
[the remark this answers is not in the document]
darosior: Ok i'll make it a function, too bad for minimal change :p

lightningd/bitcoind.c
```diff
@@ -222,18 +218,20 @@ void bitcoind_estimate_fees_(struct bitcoind *bitcoind,
 					const u32 satoshi_per_kw[], void *),
 			     void *arg)
```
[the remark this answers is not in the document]
darosior: absolutely

plugins/bcli.c
darosior: Because of json_to_bitcoin_amount()

tests/test_misc.py
darosior: Good catch thanks !

darosior:
Thanks for the review @vasild ! Sorry for the obvious fails I was too eager so push that I didn't review the cherry-picked commits of the old version :/

darosior:
Added two commits for handling the max_fee multiplier `bcli`-side.

darosior:
https://travis-ci.org/github/ElementsProject/lightning/jobs/661160239?utm_medium=notification&utm_source=github_status keeps timeouting...

niftynei:
#3542 is related to this; see comment regarding desired API for specifying a default feerate policy for `fundchannel`

lightningd/bitcoind.c
```diff
@@ -133,118 +141,110 @@ static void bitcoin_plugin_send(struct bitcoind *bitcoind,
...
+ *   - `delayed_to_us` is used for resolving our output from our unilateral close
+ *   - `htlc_resolution` is used for resolving onchain HTLCs
+ *   - `penalty` is used for resolving revoked transactions
+ *   - `min` is the minimum acceptable feerate
+ *   - `max` is the maximum acceptable feerate
+ *
  * Plugin response:
  * {
- *	"feerate": <fee per KB>,
+ *	"opening": <btc per KB>,
```
darosior: It's wrong, it's sat not BTC, will push a fix

darosior:
Corrected the wrong comments (BTC/kVB => sat/kVB)

lightningd/chaintopology.c
darosior: This was outdated, I should have removed the comment.

plugins/libplugin.c
```diff
@@ -797,6 +797,19 @@ char *u64_option(const char *arg, u64 *i)
 	return NULL;
 }
 
+char *u32_option(const char *arg, u32 *i)
+{
+	char *endp;
+
+	errno = 0;
+	*i = strtoul(arg, &endp, 0);
```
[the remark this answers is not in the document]
darosior: Ok

plugins/libplugin.c
```diff
@@ -797,6 +797,19 @@ char *u64_option(const char *arg, u64 *i)
...
+	if (*endp || !arg[0])
+		return tal_fmt(NULL, "'%s' is not a number", arg);
+	if (errno)
+		return tal_fmt(NULL, "'%s' is out of range", arg);
```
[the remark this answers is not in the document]
darosior: Oops, thanks done

darosior:
Corrected, and rebased.

vasild:
ACK dcaf75f, light code review

darosior:
Re-rebased after #3572 was merged.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/eff72417d046120a6e89049c5bca39200da5ed93

PR #3573, opened 2020-03-06
Experimental onion message support

rustyrussell:
See https://github.com/lightningnetwork/lightning-rfc/pull/755

This is a pre-requisite (in some form!) for offers, a.k.a more powerful invoices, but it's also much more efficient than abusing HTLCs to send messages (unless you want to send a payment as well anyway).


rustyrussell:
CLosing, we're going to use a different approach for messaging...

PR #3574, opened 2020-03-06
travis: do not cache ./external

vasild:
Caching the `./external` directory means that any changes to
`external/Makefile` or any updates of the submodules in `./external/*`
would not be picked up and travis would run with old instances of those,
possibly "activating" such changes after weeks or months when the cache
is pruned eventually.

Changelog-None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,6 @@ env:
 cache:
   directories:
     - dependencies
-    - external
 
 script:
   .travis/build.sh


vasild:
https://docs.travis-ci.com/user/caching/#things-not-to-cache

> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source

So maybe this PR will not slow down travis at all.

vasild:
> So maybe this PR will not slow down travis at all.

I did not realize that `./external/` also contains the build artifacts of jsmn, libbacktrace, libsodium and libwally-core and caching it means that they will not be compiled every time.

cdecker:
> > So maybe this PR will not slow down travis at all.
> 
> I did not realize that `./external/` also contains the build artifacts of jsmn, libbacktrace, libsodium and libwally-core and caching it means that they will not be compiled every time.

That was the main aim of making `external` a cached directory. FWIW each PR gets its own branch, so we'd just need to blow those caches away if an existing PR changes `external` in any way. That being said I have a terrible track record of remembering to do so :-)

cdecker:
We could be more specific and just cache `external/*.la` or stash them in a separate cache directory. Then they'd get auto-updated if the source changes underneath.

vasild:
I think caching just the build artifacts will not work as intended because [timestamps of files that are checked out from git are "now"](https://travis-ci.org/ElementsProject/lightning/builds/660308698#L670), whereas [timestamps of files from the cache are preserved to the last time they were updated in the cache](https://travis-ci.org/ElementsProject/lightning/builds/660308698#L594). Ie anything that comes from the cache (build artifacts) will look older compared to anything that is checked out from git (source code) and so everything will get recompiled.

Related: https://git.wiki.kernel.org/index.php/Git_FAQ#Why_isn.27t_Git_preserving_modification_time_on_files.3F

I think the best approach would be to install jsmn, libbacktrace, libsodium, and libwally-core externally, independently of building c-lightning, similarly to what we do with [bitcoin](https://github.com/ElementsProject/lightning/blob/c92e782/.travis/build.sh#L24) and expect them to be already built before building c-lightning. We already do something like that with libsodium - if it is present in the system, then use that one and don't bother compiling the one from `external/libsodium`.

vasild:
Compile time when `./external` is:
* cached: [1m](https://api.travis-ci.org/v3/job/660773776/log.txt)
* not cached: [2m9s](https://api.travis-ci.org/v3/job/660961545/log.txt)

cdecker:
Ok, for 1m speedup it's probably not really useful to keep this heartache around. Let's just remove the caching.

vasild:
Ok, I ditched the last commit from this PR (which was just for experimenting).

In the long term we may consider:
> [... to install jsmn, libbacktrace, libsodium, and libwally-core externally ...](https://github.com/ElementsProject/lightning/pull/3574#issuecomment-596743887)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3574/commits/a6a6a76ba98a0d12710a1deb8e466fa65460f5a4

PR #3575, opened 2020-03-06
Save channel_id when gossipd asks for txout

yaslama:
Save in the database the short channel ids queried by gossip.
The data can be used for instance to send gossip_outpoint_spent messages to gossipd only for channel transactions.
Generally, using the data already present in the db, it's a way to have the full information about the channels without needing to read the gossip_store.


rustyrussell:
Hi @yaslama !

This would indeed save the short_channel_ids which gossipd asked for, however your patch doesn't actually use the information yet?  It seems like it would be more complicated to do that; @cdecker might have thoughts here?

yaslama:
Hi @rustyrussell 
Thanks for your review.
I wanted do use the information in another PR, if this one is accepted, but I can add it in this PR.
My though was to add in the function gossipd_notify_spend() from gossipd_control.c a call to a query in wallet.c which would do something similar to:
```
SELECT channel_id
 FROM utxoset, short_channels_ids
 WHERE
  spendheight IS NOT NULL AND
  (channel_id >> 40 = blockheight) AND
  ((channel_id >> 16) & x'00FFFFFF'::bigint = txindex) AND
  (channel_id & x'FFFF'::bigint = outnum);
```
What do you think?

cdecker:
While saving us some bandwidth between master and `gossipd` it has a couple of downsides:

 - Added processing on `lightningd`: since we now need to match up spent outputs with the list of "used" short channel IDs we'd either need to query the DB or have an in-memory set to check against, both are suboptimal, costing us time to process a block slowing blockchain catchup, or increasing our memory footprint
 - Potentially racy: marking a short channel ID as used and processing a block spending the corresponding output might be racy and cause us to miss a spend.

yaslama:
> While saving us some bandwidth between master and `gossipd` it has a couple of downsides:
> 
>    * Added processing on `lightningd`: since we now need to match up spent outputs with the list of "used" short channel IDs we'd either need to query the DB or have an in-memory set to check against, both are suboptimal, costing us time to process a block slowing blockchain catchup, or increasing our memory footprint

Currently gossipd_notify_spend() (which is the function sending the message to gossipd informing it that a channel was closed) is called just after wallet_outpoint_spend() is called. And wallet_outpoint_spend() not only access the db but also update the db. Imho, the cost query to check if the short channel id is in present in another simple table is very low when compared to the update.
> 
>    * Potentially racy: marking a short channel ID as used and processing a block spending the corresponding output might be racy and cause us to miss a spend.

The race will occur if the short channel id is not known by lightningd when it knows that the tx is spent. If it's not known, that's because gossipd still didn't send to lightningd the message get_txout. But if get_txout is sent after the tx is already spent, lightningd will send an error (in get_txout() from gossip_control.c), so gossipd will not try to use the channel. Do I miss something?
> 
> While I agree that our current "send everything" approach is a bit heavy handed, I think this approach is too simplistic.

Do you have any suggestion to improve this approach. Do you agree that in any case we need to store in the db (at least at the lightningd level) the list of current channels?

cdecker:
Excellent points, I had overlooked that since we have no parallelism the race I described above cannot actually happen. So we're left with a bit of processing overhead, which you correctly pointed out is minimal, since we already have an in-memory filter for outpoints we are interested in.

It occurs to me however that there is a 1-to-1 relationship between the `utxoset` table and your new table `short_channels_ids`, so why not combine the two, i.e., add a flags column to the `utxoset` table. We also already query that table for the changed utxo entry:

https://github.com/ElementsProject/lightning/blob/24984ec680a2e614c43cef8bc1b6da666c8208f0/wallet/wallet.c#L2889-L2908

So it'd be rather easy to add the new column to the existing query and pass that back as well to know whether `gossipd` was interested in the past.

yaslama:
> It occurs to me however that there is a 1-to-1 relationship between the `utxoset` table and your new table `short_channels_ids`, so why not combine the two, i.e., add a flags column to the `utxoset` table. We also already query that table for the changed utxo entry:
> 
> https://github.com/ElementsProject/lightning/blob/24984ec680a2e614c43cef8bc1b6da666c8208f0/wallet/wallet.c#L2889-L2908
> 
> So it'd be rather easy to add the new column to the existing query and pass that back as well to know whether `gossipd` was interested in the past.

Thanks for your review. The problem is that sometimes when lightningd receives the query from gossipd, the tx record is still not in the `utxoset` table (and the txid cannot be null in this table). In this case, `bitcoind_getfilteredblock()` is called to answer later to gossipd. The flag can could be set just before sending the reply to gossipd, but it seems to me that we can miss some channels if we are not very careful. What do you think?

cdecker:
Yes, that's indeed a bit more complex than a simple function call, however like many internal API endpoints, `bitcoind_getfilteredblock` allows passing a context variable or `struct` to the callback that is called once the filtered block was successfully retrieved. We have a single call currently:

https://github.com/ElementsProject/lightning/blob/f8a21f16c9f826c7de5b2a8813cfc890a4af3c4c/lightningd/gossip_control.c#L126

This calls `got_filteredblock` if successful, and passes the `short_channel_id` that triggered the request initially. So at least that one can be marked as queried.

For other `short_channel_id`s we'd either query for them after the filtered block has been returned (happy path, we just resolve in the DB and mark as queried) or while the filtered block is still in flight, in which case I think we just queue another query for the filtered block and process the callback once we get that copy.

In all fairness we could optimize for the concurrent announcement in the same block case, but that is unlikely to happen, since this is backfilling only anyway (blocks that are being processed fully will populate the `utxoset` table as we process them).

So I think all the pieces are there, and we should always either have no UTXO entries for a given block, and we can defer marking until we have it, or we should have them all, in which case a simple marking is sufficient.

cdecker:
After benchmarking this I came to the conclusion that this change slows down the processing of blocks due to the added queries for short channel IDs more than the fire-and-forget method of just telling `gossipd` about all the spent potential channel outpoints is.

Since a performance gain was the main motivation for this, and the additional complexity and storage requirements, I think it's best to shelve this for now.

PR #3576, opened 2020-03-06
Use of null

dscotese:
I added a hint how to specify source but not short channel id.

Changelog-None


darosior:
Thanks, don't forget the changelog line (In your case it's Changelog-None and can be part of the PR description), as well as running `make` (which will call `make doc-all`) to update the manpages along with the markdown file ;)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3576/commits/646e7ac43c3887279f4c89b13aa269a2814b4d23

PR #3582, opened 2020-03-08
Pass typed plugin option data to plugin in init

niftynei:
We're only passing string values of plugin options to plugins, despite soliticting a type from the plugin during the manifest pass. 

In this PR, I will now pass the values down to plugins as the type they've specified.

Additionally, makes the input option parsing stricter and adds a new testing feature where we can capture stderr logs to verify error messages in. If you want to.

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -250,6 +250,9 @@ class Plugin(object):
                 "Name {} is already used by another option".format(name)
             )
 
+        if opt_type not in ["string", "int", "bool"]:
+            raise ValueError('{} not in supported type set (string, int, bool)')
+
         self.options[name] = {
             'name': name,
             'default': default,
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1128,11 +1128,13 @@ plugin_populate_init_request(struct plugin *plugin, struct jsonrpc_request *req)
 		name = opt->name + 2;
 		if (opt->value->as_bool) {
 			json_add_bool(req->stream, name, *opt->value->as_bool);
-			continue;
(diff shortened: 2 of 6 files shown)


contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -250,6 +250,9 @@ def add_option(self, name, default, description, opt_type="string"):
                 "Name {} is already used by another option".format(name)
             )
 
+        if opt_type not in ["string", "int", "bool"]:
```
darosior: [commit msg]
> not sure if this is totally necessary, should we jsut let clightning
enforce the input?

Might be a bit redundant but an advantage is that one can see its mistake just by trying its plugin with `python3 myplugin.py`..

lightningd/plugin.c
```diff
@@ -466,13 +466,37 @@ struct io_plan *plugin_stdout_conn_init(struct io_conn *conn,
...
+		 * is a bigger type). */
+		if (*popt->value->as_int != l)
+			return tal_fmt(tmpctx, "%s does not parse as type %s (overflowed)",
+				       popt->value->as_str, popt->type);
+	} else if (streq(popt->type, "bool")) {
+		/* valid values are 'true', 'True', '1', '0', 'false', 'False', or '' */
+		if (streq(arg, "true") || streq(arg, "True") || streq(arg, "1")) {
+			*popt->value->as_bool = true;
+		} else if (streq(arg, "false") || streq(arg, "False")
+				|| streq(arg, "0") || streq(arg, "")) {
```
niftynei: on second thought, i think that an option flag without a body is a true, i.e. `--daemon` is generally parsed as 'yes daemon'. will submit second PR to fix this.

PR #3584, opened 2020-03-09
bitcoind: check that Bitcoin plugin is alive before sending a request

darosior:
On plugin failure, this is cleaner than having a backtrace to `plugin_request_send`.

Changelog-None


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3584/commits/ed0c5c8d7b1ad69701bd39bab0d17fd8439bac25

PR #3586, opened 2020-03-11
quick fix: a blank plugin option should be true, not false

niftynei:
Fix a syntactic mistake i made in #3582 

Also fix the option interop test so that it actually verifies the log messages.

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -250,8 +250,8 @@ class Plugin(object):
                 "Name {} is already used by another option".format(name)
             )
 
-        if opt_type not in ["string", "int", "bool"]:
-            raise ValueError('{} not in supported type set (string, int, bool)')
+        if opt_type not in ["string", "int", "bool", "flag"]:
+            raise ValueError('{} not in supported type set (string, int, bool, flag)')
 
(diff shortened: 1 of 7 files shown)


rustyrussell:
Yes, as @darosior notes, pretty sure this doesn't work, since --foo will not get a callback with opt_register_arg.

We'd almost need a new type to mean a 'simple' flag argument, say `simple`?  It would use opt_register_noarg instead.

We deliberately don't allow optional arguments, since they're ambigious (is `--foo --bar` to be interpreted as `--foo=--bar` or two options?).

niftynei:
why does this test pass then?

https://github.com/ElementsProject/lightning/blob/f66f1a8f0c0be942b67777e5d87be64ad3f5faa3/tests/test_plugin.py#L66-L74

niftynei:
> Didn't test yet, but we stil use opt_register_arg() to register plugin options. With an empty string passed is it equivalent to opt_unregister_noarg() ?

Hm so the tests seem to pass, but I'm not sure that the command line arguments will work the same as the python mediated options? I removed it entirely, I think it's fine to force specification.

> We'd almost need a new type to mean a simple flag argument, say flag? It would use opt_register_noarg instead.

This seemed like a fun thing to add, so I went ahead and added this as a type.

PR #3587, opened 2020-03-12
Onion tool and sphinx minor cleanups

rustyrussell:
Prep work for messages.

Changelog-None

### DIFF
--- a/common/onion.c
+++ b/common/onion.c
@@ -143,6 +143,7 @@ u8 *onion_final_hop(const tal_t *ctx,
 /* Returns true if valid, and fills in type. */
 static bool pull_payload_length(const u8 **cursor,
 				size_t *max,
+				bool has_realm,
 				enum onion_payload_type *type,
 				size_t *len)
 {
@@ -163,7 +164,7 @@ static bool pull_payload_length(const u8 **cursor,
 	 *   length. In this case the `hop_payload_length` is defined to be 32
 	 *   bytes.
 	 */
-	if (*len == 0) {
+	if (has_realm && *len == 0) {
 		if (type)
 			*type = ONION_V0_PAYLOAD;
 		assert(*cursor - start == 1);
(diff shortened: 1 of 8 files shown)


devtools/onion.c
```diff
@@ -174,16 +169,17 @@ static void do_decode(int argc, char **argv, const u8 assocdata[ASSOC_DATA_SIZE]
...
 {
-	if (!hex_decode(arg, strlen(arg), assocdata, ASSOC_DATA_SIZE))
+	*assocdata = tal_hexdata(NULL, arg, strlen(arg));
+	if (!*assocdata)
 		return "Bad hex string";
 	return NULL;
 }
 
-static void opt_show_ad(char buf[OPT_SHOW_LEN], const u8 *assocdata)
+static void opt_show_ad(char buf[OPT_SHOW_LEN], u8 *const *assocdata)
```
niftynei: ooh this is fancy. what's a `*const`?

common/onion.c
```diff
@@ -176,10 +177,15 @@ static bool pull_payload_length(const u8 **cursor,
 	 *   case the `hop_payload_length` is equal to the numeric value of
 	 *   `length`.
 	 */
-	if (*len > 1) {
+	if (!has_realm || *len > 1) {
```
niftynei: what if `*len == 1`?
```suggestion
	if (!has_realm || *len > 0) {
```

niftynei: oh i just saw this comment ... ignore this.

```
 identified by any length over **`1`**
```

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3587/commits/f192a23a429c1b3c2e0a4b5323f824a512a11557

figured out where the `1` case is :)

devtools/onion.c
cdecker: I think `const u8 *` is a constant pointer (so the address it points to can't be modified) while `u8 *const *` is a pointer to a const (so the pointee cannot be modified). I've never seen this before either, so maybe @rustyrussell can shed some light on this?

common/onion.h
```diff
@@ -39,6 +39,7 @@ u8 *onion_final_hop(const tal_t *ctx,
  * onion_payload_length: measure payload length in decrypted onion.
  * @raw_payload: payload to look at.
  * @len: length of @raw_payload in bytes.
+ * @has_realm: used for HTLCs, where first byte 0 is magical.
```
cdecker: `0` and `1` are both special

PR #3599, opened 2020-03-20
Sphinx cleanups

rustyrussell:
No real changes, just using modern types and extracting hmac into another file for reuse.

Changelog-None

### DIFF
--- a/common/hmac.c
+++ b/common/hmac.c
@@ -1,0 +1,36 @@
+#include <bitcoin/privkey.h>
+#include <ccan/array_size/array_size.h>
+#include <ccan/mem/mem.h>
+#include <common/hmac.h>
+#include <wire/wire.h>
+
+void hmac(const void *src, size_t slen,
+	  const void *key, size_t klen,
+	  struct hmac *hmac)
+{
+	crypto_auth_hmacsha256_state state;
+
+	crypto_auth_hmacsha256_init(&state, memcheck(key, klen), klen);
+	crypto_auth_hmacsha256_update(&state, memcheck(src, slen), slen);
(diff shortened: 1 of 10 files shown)


common/hmac.c
```diff
@@ -0,0 +1,36 @@
+#include <bitcoin/privkey.h>
+#include <ccan/array_size/array_size.h>
+#include <ccan/mem/mem.h>
+#include <common/hmac.h>
+#include <wire/wire.h>
+
+void hmac(const void *src, size_t slen,
+	  const void *key, size_t klen,
```
cdecker: Seeing the previous cleanup would we ever want to use anything other than a `struct secret`?

common/sphinx.c
```diff
@@ -198,43 +197,29 @@ static void xor_cipher_stream(void *dst, const struct secret *k, size_t dstlen)
 	crypto_stream_chacha20_xor(dst, dst, dstlen, nonce, k->data);
 }
 
-static bool compute_hmac(
-	void *dst,
-	const void *src,
-	size_t len,
-	const void *key,
-	size_t keylen)
+static void compute_hmac(const u8 *src, size_t slen,
```
cdecker: This wrapper doesn't do much, maybe we can remove it as well?

common/hmac.c
rustyrussell: Sigh, unfortunately the hmac construction used by Sphinx uses the string as the *key* not src. :(

Found this out when I tried cleaning up the first time.

common/sphinx.c
rustyrussell: Will do in another PR.

rustyrussell: Actually, it is kinda useful since it takes a secret.

PR #3601, opened 2020-03-23
Enable watchtower plugins via the `commitment_revocation` hook

### LINKED ISSUE #1353: Feature Request: RPC endpoint for Watch Tower (Simple)

### LINKED ISSUE #3422: Add plugin hook for penalty transactions on revocation

## Plugin interface

``` json
{
	"to_self_delay": 1234
}
```

## Steps

cdecker:
Implements the `commitment_revocation` hook for watchtowers. Commitments are
forwarded from `openingd` and `channeld`, and tracked in `lightningd`. Upon
receiving a revocation `lightningd` generates the penalty transaction and
calls the hook so plugins can postprocess it, e.g., encrypt it, and then push
it to the watchtower.

The result of the hook is currently not used, and we don't actually wait for
the hook call to complete before telling `channeld` that it can continue. This
matches the save path in the DB, since we'd simply be replaying the
commitment/revocation if we reconnect anyway. Should synchronous operation be
required we can move the `channel_got_revoce_reply` to hook callback, but for
the sake of simplicity I left it were it is for the time being.

The penalty transaction is generated in `lightningd`, not `channeld` which
required forwarding information from `channeld` about the
`commitment_transaction` to `lightningd`, and requires `lightningd` to call
out to `hsmd` for the signature, instead of `channeld`. The decision to have
`lightningd` create the penalty transaction has the following reasons:

 - `channeld` may not have all the information about the previous commit,
   i.e., in case we come from `openingd` the first commitment is negotiated in
   `openingd` so we'd have to funnel that information through `lightningd`
   anyway.
 - `lightningd` will also deliver the hook call, and it can avoid building and
   signing the penalty tx if there are no subscribers, avoiding the tx
   generation and the request to `hsmd` altogether (not yet implemented).

Included in the `commitment_revocation` hook call is the fully signed penalty
transaction, as well as the commitment transaction hash the penalty
transaction is spending from, i.e., the transaction whose broadcast should
trigger the release of the penalty transaction. The commitment transaction
hash could also be extracted from the penalty transaction (it's its only input
after all), but this allows treating the penalty transaction as an opaque blob
and avoids parsing the transaction in the plugin.

**Not included** are the `htlc_success` and `htlc_failure` transactions that
may also be spending `commitment_tx` outputs. This is because these
transactions are much more dynamic and have a predictable timeout, allowing
wallets to ensure a quick checkin when the CLTV of the HTLC is about to
expire.

## Open Questions

 - Naming: as always finding good names is hard, and I wanted to make the hook
   name descriptive for what it does, not what it can be used for, so calling
   it `watchtower` seemed inapropriate. I'm open for better suggestions though
   :wink:
 - What extra information would be good to have?
 - Would it make sense to also include `htlc_success` and `htlc_timeout`
   transactions, despite not having the full information regarding fees and
   aggregations available?
 - Since we don't use any of the specialties of hooks in this case, would it
   make sense to define a notification instead? I went with the hook for now
   mainly so we can eventually make it synchronous.

Closes #3422
Closes #1353
Closes #3645 
Closes #3659

### DIFF
--- a/channeld/channel_wire.csv
+++ b/channeld/channel_wire.csv
@@ -100,9 +100,12 @@ msgdata,channel_fail_htlc,failed_htlc,failed_htlc,
 msgtype,channel_got_funding_locked,1019
 msgdata,channel_got_funding_locked,next_per_commit_point,pubkey,
 
+#include <common/penalty_base.h>
+
 # When we send a commitment_signed message, tell master.
 msgtype,channel_sending_commitsig,1020
 msgdata,channel_sending_commitsig,commitnum,u64,
+msgdata,channel_sending_commitsig,pbase,?penalty_base,
 msgdata,channel_sending_commitsig,fee_states,fee_states,
 # SENT_ADD_COMMIT, SENT_REMOVE_ACK_COMMIT, SENT_ADD_ACK_COMMIT, SENT_REMOVE_COMMIT
 msgdata,channel_sending_commitsig,num_changed,u16,
@@ -113,6 +116,11 @@ msgdata,channel_sending_commitsig,htlc_sigs,secp256k1_ecdsa_signature,num_htlc_s
 
 # Wait for reply, to make sure it's on disk before we send commit.
 msgtype,channel_sending_commitsig_reply,1120
+
+# When we reconnect and resend the commitment we also need to tell the master
+# the commitment details so it can create the penalty once we get the revoke.
+msgtype,channel_resending_commitsig,1130
+msgdata,channel_resending_commitsig,pbase,?penalty_base,
 
 # When we have a commitment_signed message, tell master to remember.
 msgtype,channel_got_commitsig,1021
--- a/channeld/commit_tx.c
+++ b/channeld/commit_tx.c
@@ -94,6 +94,7 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 			     struct amount_msat other_pay,
 			     const struct htlc **htlcs,
 			     const struct htlc ***htlcmap,
+			     struct wally_tx_output *direct_outputs[NUM_SIDES],
 			     u64 obscured_commitment_number,
 			     enum side side)
 {
@@ -102,7 +103,8 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 	struct bitcoin_tx *tx;
 	size_t i, n, untrimmed;
 	u32 *cltvs;
-
+	struct htlc *dummy_to_local = (struct htlc *)0x01,
+		*dummy_to_remote = (struct htlc *)0x02;
 	if (!amount_msat_add(&total_pay, self_pay, other_pay))
 		abort();
 	assert(!amount_msat_greater_sat(total_pay, funding));
@@ -215,7 +217,8 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 		struct amount_sat amount = amount_msat_to_sat_round_down(self_pay);
 
 		bitcoin_tx_add_output(tx, p2wsh, amount);
-		(*htlcmap)[n] = NULL;
+		/* Add a dummy entry to the htlcmap so we can recognize it later */
+		(*htlcmap)[n] = direct_outputs ? dummy_to_local : NULL;
 		/* We don't assign cltvs[n]: if we use it, order doesn't matter.
 		 * However, valgrind will warn us something wierd is happening */
 		SUPERVERBOSE("# to-local amount %s wscript %s\n",
@@ -248,7 +251,7 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 		 */
 		int pos = bitcoin_tx_add_output(tx, p2wpkh, amount);
 		assert(pos == n);
-		(*htlcmap)[n] = NULL;
+		(*htlcmap)[n] = direct_outputs ? dummy_to_remote : NULL;
 		/* We don't assign cltvs[n]: if we use it, order doesn't matter.
 		 * However, valgrind will warn us something wierd is happening */
 		SUPERVERBOSE("# to-remote amount %s P2WPKH(%s)\n",
@@ -304,6 +307,20 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 	 */
 	u32 sequence = (0x80000000 | ((obscured_commitment_number>>24) & 0xFFFFFF));
 	bitcoin_tx_add_input(tx, funding_txid, funding_txout, sequence, funding, NULL);
+
+	/* Identify the direct outputs (to_us, to_them). */
+	if (direct_outputs != NULL) {
+		direct_outputs[LOCAL] = direct_outputs[REMOTE] = NULL;
+		for (size_t i = 0; i < tx->wtx->num_outputs; i++) {
+			if ((*htlcmap)[i] == dummy_to_local) {
+				(*htlcmap)[i] = NULL;
+				direct_outputs[LOCAL] = tx->wtx->outputs + i;
+			} else if ((*htlcmap)[i] == dummy_to_remote) {
+				(*htlcmap)[i] = NULL;
+				direct_outputs[REMOTE] = tx->wtx->outputs + i;
+			}
+		}
+	}
 
 	bitcoin_tx_finalize(tx);
 	assert(bitcoin_tx_check(tx));
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -985,6 +985,28 @@ This will ensure backward
 compatibility should the semantics be changed in future.
 
 
+### `commitment_revocation`
+
+The `commitment_revocation` hook is called whenever a peer has sent a
+revocation for a prior channel state. The hook includes the commitment
+transaction hash that was revoked as well as a matching penalty transaction
+that can be used to punish the peer if it ever attempts to publish the revoked
+commitment. The primary use-case for this hook is to implement watchtower
+plugins that hand the penalty transaction to a remote watchtower that will
+monitor the blockchain during our absence.
+
+The payload has the following format:
+
+```json
+{
+	"commitment_txid": "df5ffe895c778e10f7742a6c5b8a0cefbe9465df58b92fadeb883752c8107c8f",
+	"penalty_tx": "0200000000010115360fd932a9c60e2f7f[...]f0ce1f271291714e5752ae993ed620"
+}
+```
+
+The result returned by the plugin is currently ignored, but should always
+return `{"result": "continue"}` in case additional handling is added in the
+future.
 
 ## Bitcoin backend
 
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -590,6 +590,45 @@ there's a member `error_message`, that member is sent to the peer
 before disconnection.
 
 
+### `commitment_revocation`
+
+This hook is called whenever a channel state is updated, and the old state was
+revoked. State updates in Lightning consist of the following steps:
+
+ 1. Proposal of a new state commitment in the form of a commitment transaction
+ 2. Exchange of signatures for the agreed upon commitment transaction
+ 3. Verification that the signatures match the commitment transaction
+ 4. Exchange of revocation secrets that could be used to penalize an eventual misbehaving party
+
+The `commitment_revocation` hook is used to inform the plugin about the state
+transition being completed, and deliver the penalty transaction. The penalty
+transaction could then be sent to a watchtower that automaticaly reacts in
+case one party attempts to settle using a revoked commitment.
+
+The payload consists of the following information:
+
+```json
+{
+	"commitment_txid": "58eea2cf538cfed79f4d6b809b920b40bb6b35962c4bb4cc81f5550a7728ab05",
+	"penalty_tx": "02000000000101...ac00000000"
+}
+```
+
+Notice that the `commitment_txid` could also be extracted from the sole input
+of the `penalty_tx`, however it is enclosed so plugins don't have to include
+the logic to parse transactions.
+
+Not included are the `htlc_success` and `htlc_failure` transactions that
+may also be spending `commitment_tx` outputs. This is because these
+transactions are much more dynamic and have a predictable timeout, allowing
+wallets to ensure a quick checkin when the CLTV of the HTLC is about to
+expire.
+
+The `commitment_revocation` hook is a chained hook, i.e., multiple plugins can
+register it, and they will be called in the order they were registered in.
+Plugins should always return `{"result": "continue"}`, otherwise subsequent
+hook subscribers would not get called.
+
 ### `db_write`
 
 This hook is called whenever a change is about to be committed to the database.
--- a/lightningd/watchtower.c
+++ b/lightningd/watchtower.c
@@ -1,0 +1,128 @@
+#include "watchtower.h"
+
+#include <bitcoin/feerate.h>
+#include <bitcoin/script.h>
+#include <bitcoin/signature.h>
+#include <bitcoin/tx.h>
+#include <common/htlc_tx.h>
+#include <common/key_derive.h>
+#include <common/keyset.h>
+#include <hsmd/gen_hsm_wire.h>
+#include <lightningd/channel.h>
+#include <lightningd/peer_control.h>
+#include <wire/wire_sync.h>
+
+static const u8 ONE = 0x1;
+
+const struct bitcoin_tx *
+penalty_tx_create(const tal_t *ctx, struct lightningd *ld,
+		  const struct channel *channel,
+		  const struct secret *revocation_preimage,
+		  const struct bitcoin_txid *commitment_txid,
+		  s16 to_them_outnum, struct amount_sat to_them_sats)
+{
+	u8 *wscript;
+	struct bitcoin_tx *tx;
+	struct keyset keyset;
+	size_t weight;
+	u8 *msg;
+	struct amount_sat fee, min_out, amt;
+	struct bitcoin_signature sig;
+	u32 locktime = 0;
+	bool option_static_remotekey = channel->option_static_remotekey;
+	struct pubkey final_key;
+	u8 **witness;
+	u32 remote_to_self_delay = channel->channel_info.their_config.to_self_delay;
+	const struct amount_sat dust_limit = channel->our_config.dust_limit;
+	u32 feerate_per_kw = try_get_feerate(ld->topology, FEERATE_PENALTY);
+	BUILD_ASSERT(sizeof(struct secret) == sizeof(*revocation_preimage));
+	const struct secret remote_per_commitment_secret = *revocation_preimage;
+	struct pubkey remote_per_commitment_point;
+	struct basepoints basepoints[NUM_SIDES];
+	u64 channel_dbid = channel->dbid;
+	basepoints[LOCAL] = channel->local_basepoints;
+	basepoints[REMOTE] = channel->channel_info.theirbase;
+
+	if (to_them_outnum == -1 ||
+	    amount_sat_less_eq(to_them_sats, dust_limit)) {
+		log_unusual(channel->log,
+			    "Cannot create penalty transaction because there "
+			    "is no non-dust to_them output in the commitment.");
+		return NULL;
+	}
+
+	if (!pubkey_from_secret(&remote_per_commitment_secret, &remote_per_commitment_point))
+		fatal("Failed derive from per_commitment_secret %s",
+		      type_to_string(tmpctx, struct secret,
+				     &remote_per_commitment_secret));
+
+	if (!bip32_pubkey(ld->wallet->bip32_base, &final_key,
+			  channel->final_key_idx)) {
+		fatal("Could not derive onchain key %" PRIu64,
+		      channel->final_key_idx);
+	}
+
+	if (!derive_keyset(&remote_per_commitment_point,
+			   &basepoints[REMOTE],
+			   &basepoints[LOCAL],
+			   option_static_remotekey,
+			   &keyset))
+		abort(); /* TODO(cdecker) Handle a bit more gracefully */
+	wscript = bitcoin_wscript_to_local(tmpctx, remote_to_self_delay,
+					   &keyset.self_revocation_key,
+					   &keyset.self_delayed_payment_key);
+
+	tx = bitcoin_tx(ctx, chainparams, 1, 1, locktime);
+	bitcoin_tx_add_input(tx, commitment_txid, to_them_outnum, 0xFFFFFFFF,
+			     to_them_sats, NULL);
+
+	bitcoin_tx_add_output(tx, scriptpubkey_p2wpkh(tx, &final_key),
+			      to_them_sats);
+
+	/* Worst-case sig is 73 bytes */
+	weight = bitcoin_tx_weight(tx) + 1 + 3 + 73 + 0 + tal_count(wscript);
+	weight = elements_add_overhead(weight, 1, 1);
+	fee = amount_tx_fee(feerate_per_kw, weight);
+
+	if (!amount_sat_add(&min_out, dust_limit, fee))
+		log_broken(channel->log,
+			      "Cannot add dust_limit %s and fee %s",
+			      type_to_string(tmpctx, struct amount_sat, &dust_limit),
+			      type_to_string(tmpctx, struct amount_sat, &fee));
+
+	if (amount_sat_less(to_them_sats, min_out)) {
+		/* FIXME: We should use SIGHASH_NONE so others can take it */
+		fee = amount_tx_fee(feerate_floor(), weight);
+	}
+
+	/* This can only happen if feerate_floor() is still too high; shouldn't
+	 * happen! */
+	if (!amount_sat_sub(&amt, to_them_sats, fee)) {
+		amt = dust_limit;
+		log_broken(channel->log,
+			   "TX can't afford minimal feerate"
+			   "; setting output to %s",
+			   type_to_string(tmpctx, struct amount_sat, &amt));
+	}
+	bitcoin_tx_output_set_amount(tx, 0, amt);
+	bitcoin_tx_finalize(tx);
+
+	u8 *hsm_sign_msg =
+	    towire_hsm_sign_penalty_to_us(ctx, &remote_per_commitment_secret, tx,
+					  wscript, *tx->input_amounts[0],
+					  &channel->peer->id, &channel_dbid);
+
+	if (!wire_sync_write(ld->hsm_fd, take(hsm_sign_msg)))
+		log_broken(channel->log, "Writing sign request to hsm");
+
+	msg = wire_sync_read(tmpctx, ld->hsm_fd);
+	if (!msg || !fromwire_hsm_sign_tx_reply(msg, &sig)) {
+		fatal("Reading sign_tx_reply: %s", tal_hex(tmpctx, msg));
+	}
+
+	witness = bitcoin_witness_sig_and_element(tx, &sig, &ONE, sizeof(ONE),
+						  wscript);
+
+	bitcoin_tx_input_set_witness(tx, 0, take(witness));
+	return tx;
+}
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -1200,3 +1200,75 @@ def test_replacement_payload(node_factory):
         l1.rpc.pay(inv)
 
     assert l2.daemon.wait_for_log("Attept to pay.*with wrong secret")
+
+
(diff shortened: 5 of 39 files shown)


cdecker:
This is an initial proposal to brainstorm around, and I am happy to see what else is missing from this.

Ping @ZmnSCPxj, @NicolasDorier and @sr-gi since you were involved in either #1353 or #3422.

 - I currently only expose the `penalty_tx`, and not the `commitment_tx` like @ZmnSCPxj proposed in the discussion of #3422 since this is just the initial version, and adding information as the need arises is easier than to overshoot the target.

doc/PLUGINS.md
```diff
@@ -547,6 +547,44 @@ there's a member `error_message`, that member is sent to the peer
 before disconnection.
 
 
+### `commitment_revocation`
+
```
ZmnSCPxj: It seems the other hook descriptions always start with `This hook is called when ...`.  For consistency suggest adding the following one-sentence paragraph:

    This hook is called whenever a channel is updated.

doc/PLUGINS.md
```diff
@@ -547,6 +547,44 @@ there's a member `error_message`, that member is sent to the peer
...
+Not included are the `htlc_success` and `htlc_failure` transactions that
+may also be spending `commitment_tx` outputs. This is because these
+transactions are much more dynamic and have a predictable timeout, allowing
+wallets to ensure a quick checkin when the CLTV of the HTLC is about to
+expire.
+
+The `commitment_revocation` hook is a chained hook, i.e., multiple plugins can
+register it, and they will be called in the order they were registered in.
+Plugins should always return `{"result": "continue"}`, otherwise subsequent
+hook subscribers would not get called.
```
ZmnSCPxj: What would the use-case be for interactivity of hooks?  Is there any reason for a plugin to somehow affect state update?  Would not `htlc_accepted` (and maybe a future `ptlc_accepted`) be sufficient as a hook, since we will largely be manipulating HTLC/PTLCs in the future only?  It seems to me more sensible to use a notification for `commitment_revocation` event, and use a hook for `htlc_accepted`.

We can only emit this event (whether notification or hook) *after* the counterparty sends a `revoke_and_ack` to us.  This is pretty much the last step in the update dance (I think we send our own `revoke_and_ack` even before we know of the counterparty `revoke_and_ack`, as otherwise it would be a chicken-and-egg problem), so I do not see *what* a hook can affect for this state: it would be "too late" at this point to do anything about this state, so what can a plugin affect for this state?  Even if we disconnect at this point, as far as the counterparty is concerned the new state is now the latest (and it will re-send the revocation data on channel re-establish).

ZmnSCPxj: Hmm.

What it *could* affect would be dependent processing: i.e. we do not consider any HTLCs introduced by the new state to be "irrevoably committed" until the most recent `commitment_revocation` hook returns.  This implies adding a new state in the database: we know the revocation of a previous state, but have not yet confirmed with `commitment_revocation` that all the plugins have confirmed they have processed this revocation correctly, in which case on a restart we will re-invoke `commitment_revocation` on all plugins.  It seems safer to implement this first if we are going to make this a hook anyway.

ZmnSCPxj: So, in case of a failure in `commitment_revocation` (i.e. it returns anything other than a non-error result `{"result": "continue"}`), then what we should do is to store the failed `commitment_revocation` on-database, then set the channel to "always fail" mode for now.

Then future updates on the channel, we:

* We receive `revoke_and_ack`.
* Append the new `commitment_revocation` to the list in he database.
* For every `commitment_revocation` in the in-database list (in order of appending!):
  * Send `commitment_revocation` to the plugin hook chain.
  * If it succeeds, remove that from the in-database list and continue with the next item in the list.
  * If it fails, exit this loop and mark all live incoming HTLCs as failing with `temporary_node_failure`.
* If the above loop completely succeeded, the in-database list would be empty.
* If the above loop completely succeeded, then we can safely handle (forward or claim) any incoming HTLCs.

On startup we perform the above loop as well, and handle incoming HTLCs as normal if the loop completes.

This allows us to keep the node "alive" in some way if its watchtower dies.

sr-gi:
```
[...] but this allows treating the penalty transaction as an opaque blob
and avoids parsing the transaction in the plugin. [...]

What extra information would be good to have?
```

If we are treating the penalty transaction as a blackbox, then the `to_self_delay`  requested to the peer would be nice to have. It may be useful if accountability is to be offered to define early rejections. I'm open to discuss this as well since a minimum `to_self_delay` could be defined by the protocol and everything below that won't count towards accountability, allowing us to get rid of this field altogether. 

```
- Would it make sense to also include htlc_success and htlc_timeout
transactions, despite not having the full information regarding fees and
aggregations available?
```

I had two approaches in mind regarding second stage htlc transactions:

- Sending them as an additional field when sending an appointment to the tower (part of the encrypted blob).

If we sent them within the encrypted blob then the tower could start monitoring them after a breach + decryption. The main issue I can see with this approach is that a big chunk of data will be potentially replicated in the tower. Every new commitment will have it's pending htlcs attached to it, so several commitment may share htlcs. This makes the user waste some of the tower space he paid for.

-  Treating them in the same way commitment transactions are.

Second stage htlc transaction could be sent to the tower following the same approach as commitment-penalty transactions pairs are. The `htlc_sucess` / `htlc_timeout` transaction would follow the `commitment_tx` role, whereas a transaction spending from it (`refund_tx`?) will follow the `penalty_tx` role. Therefore, `refund_tx` can be encrpyted using a `locator` derived from `htlc_sucess` / `htlc_timeout`, and the tower will react if that's seen on the chain.

This approach has the advantage that each `htlc` is backed up only once.

ZmnSCPxj:
> This approach has the advantage that each `htlc` is backed up only once.

Not quite: since the `htlc_timeout` / `htlc_success` commits to a particular `commitment_tx` via its input spend, the `htlc_timeout`/`htlc_success` transactions will change per commitment transaction, and if you do not want to leak too much info to the watchtower then we would have to re-give them at each commitment tx as well, so ---

sr-gi:
> Not quite: since the htlc_timeout / htlc_success commits to a particular commitment_tx via its input spend, the htlc_timeout/htlc_success transactions will change per commitment transaction, 

Oh yeah, you're right. I forgot about the htlc_sigs exchanged on every update.

The first approach also has the issue of how the blobs grow between different updates, that may leak the number of htlcs in the blob though. 

> and if you do not want to leak too much info to the watchtower then we would have to re-give them at each commitment tx as well, so ---

Not sure I follow you here, this seems an issue for the first approach given how the blob grows, but not for this one. Even if an htlc is added/removed between updates and the user sends one appointment more/less in the next round that it did in the last round, given that all locators will be new, that should be indistinguishable from multiple channels being updated at the same time.

ZmnSCPxj:
> Not sure I follow you here

I was still commenting on your "This approach has the advantage that each `htlc` is backed up only once.", i.e. you still need to re-send the HTLC-timeout / HTLC-success at each commitment update, so you still need to re-send a lot of data per update, regardless if each HTLC has its own appointment or not.

In any case giving each HTLC-commitment pair its own appointment seems better than putting them all in a single var-length appointment,  we can give an upper bound on blob size and the watchtower can charge based on that.

I was trying to devise a safe alternative where we use synthetic locators for the HTLC blobs as well but there is no safe alternative, sigh; it would require giving the watchtower the ability to sign for any penalty on the HTLC, which would allow the watchtower to be bribed by the counterparty by making the penalty on the HTLC split between the coutnerparty and the watchtower, thus not trustless.

doc/PLUGINS.md
cdecker: > What would the use-case be for interactivity of hooks? Is there any reason for a plugin to somehow affect state update? Would not `htlc_accepted` (and maybe a future `ptlc_accepted`) be sufficient as a hook, since we will largely be manipulating HTLC/PTLCs in the future only? It seems to me more sensible to use a notification for `commitment_revocation` event, and use a hook for `htlc_accepted`.

The reason I like it being a hook is that if we decide that notifying watchtowers is as critical as it's assumed to be, we can make sure that the watchtower gets the `penalty_tx` before we make progress. This can be achieved by splitting the `peer_got_revoke` function, then sending the `towire_channel_got_revoke_reply` and save to DB only after the watchtower plugin gave us the all-clear. If the watchtower is currently unreachable it is perfectly fine for a plugin to return a `{"result": "fail"}` signalling that we must abort since proceeding could mean that our counterparty publishes this specific state when we're offline, and the watchtower cannot save us. Aborting and restarting means we get another chance later, since our peer will resend the last round of commitment and revocation, at which point the watchtower hopefully is operational again.

I'm not sure myself whether the watchtower hook is that important, but it's nice to have the option :-) 'm not proposing any other state modification, since those are better suited for `htlc_accepted` and `ptlc_accepted` as you correctly point out.

cdecker:
> If we are treating the penalty transaction as a blackbox, then the `to_self_delay` requested to the peer would be nice to have. It may be useful if accountability is to be offered to define early rejections. I'm open to discuss this as well since a minimum `to_self_delay` could be defined by the protocol and everything below that won't count towards accountability, allowing us to get rid of this field altogether.

Is having access to this really useful at all? It's a deadline by which the transaction has to be broadcast, and there is little to no leeway in how the transaction is being broadcast (there is no way of adding more fees at the moment). So isn't broadcasting as soon as the revoked commitment is being seen the only thing that can possibly be done? Even if it's too late, what else can a watchtower do?

> I had two approaches in mind regarding second stage htlc transactions:
> 
> * Sending them as an additional field when sending an appointment to the tower (part of the encrypted blob).

From this I gather that your watchtower would indeed work if we provide only the `penalty_tx` for now, and we can discuss including the `htlc_*` second stage txs in a later update, correct? I'm happy to do this as a first simple step, and then add incremental changes as we learn more about the uses and behaviours.

> This disappears if we are able to switch to `SIGHASH_NOINPUT`/`SIGHASH_ANYPREVOUT` for the second-stage HTLCs, but now you need to change your locator to be based on what `SIGHASH_*` signs instead of the txid that appears onchain.

If we have either of those we get almost perfect watchtowers for free. eltoo's `update_tx` are constant in size, and allow late fee-binding and fee-bumping. You'll know this, but I just thought I'd reiterate this for the benefit of the casual reader :wink: 

We could also emulate `SIGHASH_NOINPUT` by just sending a new signature along with a reference to a prior `htlc_*` appointment. The watchtower could then just rebind the transaction and plug in the signature. It's not free like the `SIGHASH_NOINPUT` variant, but at least we'd minimize the transferred and stored data. 

> The first approach also has the issue of how the blobs grow between different updates, that may leak the number of htlcs in the blob though.

A simple "solution" would be to pad. However that's uncompressible data that the watchtower needs to store in addition to the simple penalty. We could always just treat `htlc_success` and `htlc_penalty` transactions identically to the `penalty_tx`, assuming the watchtower just treats the `penalty_tx` as a pre-canned reaction to a commitment tx (`htlc_timeout` would require a delay to be passed along though).

sr-gi:
> Is having access to this really useful at all? It's a deadline by which the transaction has to be broadcast, and there is little to no leeway in how the transaction is being broadcast (there is no way of adding more fees at the moment). So isn't broadcasting as soon as the revoked commitment is being seen the only thing that can possibly be done? Even if it's too late, what else can a watchtower do?

Yeah, not much to do for now. I was specially thinking on fee bumping and priority queues here, but I guess we could leave that for the future.

> A simple "solution" would be to pad. However that's uncompressible data that the watchtower needs to store in addition to the simple penalty. We could always just treat htlc_success and htlc_penalty transactions identically to the penalty_tx, assuming the watchtower just treats the penalty_tx as a pre-canned reaction to a commitment tx (htlc_timeout would require a delay to be passed along though).

Agreed, I think that treating them in the same way as commitment/penalties is the best way to go for now.

cdecker:
> Yeah, not much to do for now. I was specially thinking on fee bumping and priority queues here, but I guess we could leave that for the future.

If you expect it to be set I'll just add it, since it is quite easy to add anyway. Just seemed like a rather random fact to use in this interface :wink: 



> Agreed, I think that treating them in the same way as commitment/penalties is the best way to go for now.

If you're using the commitment txid as a locator and decryption key all HTLCs could still get collated this way, but we could spread the HTLCs over a number of watchtowers, since they're likely less valuable than the `their_unilateral/to_them` output which'd be claimed by the penalty tx,

ZmnSCPxj:
> If we have either of those we get almost perfect watchtowers for free. eltoo's `update_tx` are constant in size, and allow late fee-binding and fee-bumping. You'll know this, but I just thought I'd reiterate this for the benefit of the casual reader :wink:

Yes, but some might strongly prefer the punitive Poon-Dryja mechanism rather than the Decker-Russell-Osuntokun mechanism even post `SIGHASH_ANYPREVOUT`.

> We could also emulate `SIGHASH_NOINPUT` by just sending a new signature along with a reference to a prior `htlc_*` appointment. The watchtower could then just rebind the transaction and plug in the signature. It's not free like the `SIGHASH_NOINPUT` variant, but at least we'd minimize the transferred and stored data. 

That seems like a workable approach as well.

For a consistent interface, we could divide blobs into two classes:

* Commitment-tx blobs.
* HTLC-timeout/-success tx blobs.

An HTLC-timeout/-success tx blob would:

* Have a randomly-generated 128-bit locator, which is a truncation of a randomly-generated 256-bit encryption key.
  * The node can use a hash of the HTLC details (hash, timelock, amount), as encryption key, then truncate to 128-bits for the locator.
* Contain an HTLC-timeout/-success tx with a single input:
  * A "dummy" input txid of 00...
  * A "dummy" input txout of 65535 (or whatever maximum number of outputs is allowed on the commitment tx).
  * A "dummy" input witness stack containing a DER-encoding of a (0xFF..., 0xFFF...) invalid signature (or something that can be DER-compressed-encoded) plus the HTLC script.
* Contain a penalty tx with a single input:
  * A "dummy" input txid of 00...
   * This will be filled in with the HTLC-timeout/-success txid.
  * A 0 input txnum.
  * A "dummy" input witness stack containing an invalid signature,  plus the script output of the HTLC-timeout/HTLC-success.

A commitment-tx blob would:

* Have a 128-bit locator which is the truncation of a 256-bit encryption key, which is the commitment txid.
* Contain a penalty tx blah blah.
* Contain a set of HTLC references, each of which is:
  * A 256-bit random encryption key that matches an existing HTLC-success/timeout blob locator.
  * A signature that would make the corresponding HTLC-success/timeout transaction valid.
  * A signature that would make the corresponding HTLC-success/timeout penalty transaction valid.

Still, this is less elegant than the simple transaction-reaction that we could do with the proposed scheme where every commitment-tx and every HTLC-success/timeout-tx gets its own blob stored in the watchtower.

sr-gi:
> Still, this is less elegant than the simple transaction-reaction that we could do with the proposed scheme where every commitment-tx and every HTLC-success/timeout-tx gets its own blob stored in the watchtower.

Agreed. Furthermore, from the tower perspective using the same approach for both makes the design cleaner, since any kind of transaction goes trough the same logic, and triggers (either old commitments or htlc-success/timeout) are threaten the same way.

cdecker:
> Yes, but some might strongly prefer the punitive Poon-Dryja mechanism rather
> than the Decker-Russell-Osuntokun mechanism even post `SIGHASH_ANYPREVOUT`.

I've definitely heard a few people expressing that preference, however I think
it is rather short-sighted, since the cost for all participants is complexity
(state updates need to be carefully choreographed, and fees commitments months
ahead of time), safety (backups are dangerous with LN-penalty), and inherent
limitations (limited to 2 parties). But I guess it's not you I need to be
convincing here :-)

cdecker:
Ok, since no objections have been raised I think we can consider this a working solution. Rebased on top of `master` and marked ready for review :wink:

doc/PLUGINS.md
rustyrussell: There's no reason for an intermediate state.  If we receive the revocation and don't complete all the hooks, we will ask the peer for it again, just as if we'd never received it.

rustyrussell:
I was playing around with trying to neaten this a little, and I think your implementation is wrong?

Seems like openingd needs to tell lightnignd two things for the penalty tx: the remote tx, and which output is the to-self output for that.  However, you didn't add the remote tx, you're still just giving lightningd the *local* tx, which doesn't work.

The penalty tx creation just needs remote commit txid, outpoint, and amount, and we need to add this to the db if we want to replay penalty creation (or the penalty tx directly, but saving this information is cheaper).

1. I simplified your infra work: we only need the to-local out which makes things simpler.
2. I implemented a penalty_base struct which we can hand around.

From here:
1. Get openingd to return that struct (iff remote has an output).
2. Hand it to channeld.
3. Have channeld return that struct (iff remote has an output) for each revoke_and_ack.
4. Save it to the database.

Now, once we've done all this, we might as well get channeld to generate the penalties as well, which means it needs the penalty feerate (and it needs to be kept up-to-date like the normal feerate).

This avoids some of your patches, and I think keeps it simple.

I'll push my work for you to look at...

channeld/commit_tx.c
```diff
@@ -306,6 +308,18 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
...
+	for (size_t i = 0; i < tx->wtx->num_outputs; i++) {
+		if ((*htlcmap)[i] == dummy_to_local) {
+			(*htlcmap)[i] = NULL;
+			output_index[LOCAL] = i;
+		} else if ((*htlcmap)[i] == dummy_to_remote) {
+			(*htlcmap)[i] = NULL;
+			output_index[REMOTE] = i;
+		}
+	}
+
```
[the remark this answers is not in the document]
cdecker: I know, right? xD

cdecker:
> Seems like openingd needs to tell lightnignd two things for the penalty tx: the remote tx, and which output is the to-self output for that. However, you didn't add the remote tx, you're still just giving lightningd the _local_ tx, which doesn't work.

That's a dumb mistake on my end, I'll fix that asap.

> The penalty tx creation just needs remote commit txid, outpoint, and amount, and we need to add this to the db if we want to replay penalty creation (or the penalty tx directly, but saving this information is cheaper).

I agree that wrapping this information in a struct would cut down on the changes in the inter-daemon protocol and it'd be a much nicer bundle to hand around. I'll cherry pick the struct into the PR.

I'm not sure limiting the output categorization to the `to_local` output is too specific, and we'd need to re-add the `to_remote` categorization. Would it maybe make sense to encapsulate the `commit_tx` into a struct that fully annotates the transaction, and we can just unwrap the transaction once we don't need that information anymore. That would have avoided all the state drilling through function layers I had to do to surface these annotations to where we needed them. I'm ok with going the ad-hoc way for now, but let's keep the wrapper in mind should we need to shuttle more information to the outside.

> From here:
> 
> 1. Get openingd to return that struct (iff remote has an output).
> 2. Hand it to channeld.
> 3. Have channeld return that struct (iff remote has an output) for each revoke_and_ack.
> 4. Save it to the database.
> 
> Now, once we've done all this, we might as well get channeld to generate the penalties as well, which means it needs the penalty feerate (and it needs to be kept up-to-date like the normal feerate).

Like I wrote in the PR message I chose to construct the penalty tx in `lightningd` for two reasons: 

 - Let information flow up the daemon tree, aggregating it in `lightningd` where we can make an informed decision about whether the penalty tx needs to be built (and signed) in the first place. This potentially saves us interactions with the `hsmd` if there is no watchtower plugin. It also minimizes shuttling information back and forth.
 - Don't have `channeld` require the `hsmd` capability to sign on-chain transactions.

Also if we need to store the commitment `to_local` outpoint we'd need to pass that back to `channeld` on init.

Re: storing the information in the database: I was unsure whether we'd require this, but I see now that there is a gap in my coverage, namely when the remote peer sent the revoke, but we didn't process it, then we won't get the previous commitment on reconnect, which I is what I had mistakenly assumed. Too bad I was hoping to avoid having to drag state around further.

cdecker:
Clean rebase on top of `master`, now cherry-picking things from Rusty's proposal (keeping penalty creation on `lightningd` for now, while we discuss the tradeoffs).

lightningd/watchtower.c
```diff
@@ -0,0 +1,127 @@
...
+	bitcoin_tx_add_output(tx, scriptpubkey_p2wpkh(tx, &final_key),
+			      to_them_sats);
+
+	/* Worst-case sig is 73 bytes */
+	weight = bitcoin_tx_weight(tx) + 1 + 3 + 73 + 0 + tal_count(wscript);
+	weight = elements_add_overhead(weight, 1, 1);
+	fee = amount_tx_fee(feerate_per_kw, weight);
+
+	/* Result is trivial?  Spend with small feerate, but don't wait
+	 * around for it as it might not confirm. */
```
niftynei, line 88: this portion of this comment doesn't make sense out of context, since it's referring to how onchaind tracks output resolution

>but don't wait around for it as it might not confirm

lightningd/watchtower.c
```diff
@@ -0,0 +1,127 @@
...
+
+	/* This can only happen if feerate_floor() is still too high; shouldn't
+	 * happen! */
+	if (!amount_sat_sub(&amt, to_them_sats, fee)) {
+		amt = dust_limit;
+		log_broken(channel->log,
+			   "TX can't afford minimal feerate"
+			   "; setting output to %s",
+			   type_to_string(tmpctx, struct amount_sat, &amt));
+	}
```
niftynei, line 108: in the case that the penalty output is dust, is it still worth issuing a penalty tx? I'm guessing that the watchtower needs to be updated to the new state *every* commitment so it still needs something.

tests/test_plugin.py
```diff
@@ -1164,3 +1164,79 @@ def test_feature_set(node_factory):
...
+
+    # Now make sure it is out of date
+    l1.rpc.pay(l2.rpc.invoice(25000000, 'lbl2', 'desc2')['bolt11'])
+
+    # l2 stops watching the chain, allowing the watchtower to react
+    l2.stop()
+
+    # Now l1 cheats
+    bitcoind.rpc.sendrawtransaction(tx)
+    import time
```
niftynei, line 1198: shouldn't all imports be at the top of the file?

tests/test_plugin.py
```diff
@@ -1164,3 +1164,79 @@ def test_feature_set(node_factory):
...
+            break
+
+    # Need this to check that l2 gets the funds
+    penalty_meta = bitcoind.rpc.decoderawtransaction(penalty)
+
+    time.sleep(1)
+    bitcoind.generate_block(1)
+
+    # l1 onchaind should detect the unilateral, and then crash because it cheated
+    # l1.daemon.wait_for_log(r'Could not find resolution for output 0')
```
niftynei, line 1223: delete unused code?

doc/PLUGINS.md
```diff
@@ -590,6 +590,45 @@ there's a member `error_message`, that member is sent to the peer
 before disconnection.
 
 
+### `commitment_revocation`
+
+This hook is called whenever a channel state is updated, and the old state war
```
niftynei, line 595: `war -> was`

lightningd/watchtower.c
sr-gi: I was also wondering about this. I'm trying it atm and realised that until the dust limit is not passed, no penalty is created. 

```
2020-04-07T11:01:08.782Z DEBUG 020e188b565427822990c7c599f772f68e095d7f93dfacaaacfc6c8239ad37ef08-chan#2: Balance 500000000msat -> 499999948msat
2020-04-07T11:01:08.783Z UNUSUAL 020e188b565427822990c7c599f772f68e095d7f93dfacaaacfc6c8239ad37ef08-chan#2: Cannot create penalty transaction because there is no non-dust to_them output in the commitment.
```

The commitments are update anyway aren't they? If so this could be a problem since the tower would not be covering some cases.

sr-gi:
> I currently only expose the penalty_tx, and not the commitment_tx like @ZmnSCPxj proposed in the discussion of #3422 since this is just the initial version, and adding information as the need arises is easier than to overshoot the target.

Coming back to this, it may be worth exposing the commitment_tx if compiled with DEBUG flags to test breaches.

lightningd/watchtower.c
cdecker: This should only be an issue until the fundee goes above the reserve value, before then it makes little sense to create a penalty transaction at all, since the counterparty doesn't have funds to lose in the first place. The reserve value guarantees that there is at least one non-dust outputs once both parties have surpassed the reserve value:

[BOLT 02](https://github.com/lightningnetwork/lightning-rfc/blob/f068dd0d8dfa5ae75feedd99f269e23be4777381/02-peer-protocol.md#requirements):
> MUST set channel_reserve_satoshis greater than or equal to dust_limit_satoshis.

So we only need to worry about `reserve_value < dust + minfeerate * penaltytx_weight`. In which case we'd still get the majority of funds (otherwise the penalty would actually be worth it) and we're ok to give our peer a couple of breadcrumbs that would have been his in that state anyway. The loss therefore is limited to `minfeerate * penatlytx_weight` :-)

But it also means that there may be state updates that intermittently do not result in a penalty_tx being created, which as far as I can see is part of the protocol and would happen in all implementations.

tests/test_plugin.py
cdecker: They should, yes :wink:

cdecker:
> Coming back to this, it may be worth exposing the commitment_tx if compiled with DEBUG flags to test breaches.

The problem with this is that we only ever get a half-signed commitment TX, since the breaching party doesn't share his signature on the `commitment_tx`. If this were not the case we could publish the transaction, pretending our counterparty cheated, and immediately penalize him for misbehaving :-)

sr-gi:
> > Coming back to this, it may be worth exposing the commitment_tx if compiled with DEBUG flags to test breaches.
> 
> The problem with this is that we only ever get a half-signed commitment TX, since the breaching party doesn't share his signature on the `commitment_tx`. If this were not the case we could publish the transaction, pretending our counterparty cheated, and immediately penalize him for misbehaving :-)

Yeah, of course 🤦‍♂

cdecker:
What you can do is use `lightning-cli dev-sign-last-tx` to sign the last commitment and return it from the breaching party, then perform an update and release the signed transaction. That's how we test internally.

cdecker:
I folded @rustyrussell's proposal for a `struct penalty_base` in and (hopefully) fixed up the issue with `openingd` returning the wrong penalty base  (our LOCAL, instead of their LOCAL). Since managing the `fixups!` got really tedious I went and re-split the final diff into logical chunks (that compile and pass tests individually).

Now working on saving the penalty base in the DB, so we can restart / reconnect and immediately produce valid penalty transactions.

cdecker:
I added two more commits to the end which check some more constraints
regarding the state transitions we track when committing, and to save/load the
`struct penalty_base`s for previous and next commitments to the `channels`
table.

I spent a bit of time trying to figure out whether I could skip either the
`prev_commitment` or the `next_commitment`, but couldn't convince myself the
we don't need either. Maybe @rustyrussell has an idea for or against this
sledgehammer approach, and whether we can skip one of the two. It's not a lot
of data we pull from and push to the DB (48 bytes) but still.

At the moment I'm stress testing this, in the hopes of causing one of the
constraint checks to fail, but so far it seems ok.

cdecker:
Rebased on top of #3626 since that broke some of the `commit_txs` calls.

Should be good to go now, tested it against all configurations in my build matrix, no new issues (just the ones addressed in #3636).

ZmnSCPxj:
> The problem with this is that we only ever get a half-signed commitment TX

We could provide the commitment tx without witness data, all we needed was to see the commitment tx outputs in detail.

channeld/channel_wire.csv
```diff
@@ -113,6 +116,11 @@ msgdata,channel_sending_commitsig,htlc_sigs,secp256k1_ecdsa_signature,num_htlc_s
 # Wait for reply, to make sure it's on disk before we send commit.
 msgtype,channel_sending_commitsig_reply,1120
 
+# When we reconnect and resend the commitment we also need to tell the master
+# the commitment details so it can create the penalty once we get the revoke.
+msgtype,channel_resending_commitsig,1130
+msgdata,channel_resending_commitsig,pbase,?penalty_base,
```
rustyrussell, line 122: No, this is in the wrong place, leading to us having to keep two values inside lightningd.

See 74c8a21a2a2d45a3ef47f58bff6c4def262495d7

Have *channeld* keep the two values, give it at the right time (when we receive the revoke), and we don't need that logic in lightningd.  We just need lightningd to save (the single, last pbase) to the db so it can hand it back to channeld when channeld restarts.

cdecker:
Removed the milestone since this isn't ready yet.

cdecker:
There is now an alternative implementation of this (without changing the public interface) in PR #3645

PR #3604, opened 2020-03-25
Opening fail fixes

rustyrussell:
This is why I run a node.  Of course, I should have checked it two weeks ago!

### DIFF
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -401,8 +401,8 @@ void peer_start_channeld(struct channel *channel,
 					   take(&hsmfd), NULL));
 
 	if (!channel->owner) {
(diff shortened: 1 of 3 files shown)


lightningd/channel_control.c
```diff
@@ -401,8 +401,8 @@ void peer_start_channeld(struct channel *channel,
 					   take(&hsmfd), NULL));
 
 	if (!channel->owner) {
-		log_unusual(channel->log, "Could not subdaemon channel: %s",
-			    strerror(errno));
+		log_broken(channel->log, "Could not subdaemon channel: %s",
```
cdecker: `subdaemon` as in a verb? Had to re-read it a couple of times :-)

PR #3611, opened 2020-03-31
Plugin implementation of `keysend` spontaneous payments

cdecker:
The `keysend` plugin implements the ability to receive spontaneous payments by
including the `payment_preimage` in the onion payload, thus giving the
recipient the ability to claim the payment without having to exchange invoices
first. This sometimes erroneously called a sphinx-send, and can lead to
confusion with the onion routing protocol based on the sphinx paper.

The plugin has the ability to extract the preimage from the payload and will
tell `lightningd` to resolve the HTLC with the provided preimage. As proposed
by @rustyrussell, in order to keep track of funds, `lightningd` will create an
ad-hoc invoice that is immediately fulfilled via the resolved HTLC. This
ensures that we have an entry in the incoming payments, even though we didn't
really have a matching invoice. The ad-hoc invoice is generated whenever a
plugin tells `lightningd` to resolve an HTLC, so that these payments
terminated by plugins now always leave an accounting entry in the database.

Due to the uniqueness constraint on `payment_hash` in the `invoices` table we
cannot always insert an invoice that matches the resolved HTLC, in which case
we will simply not add an entry. This should happen rarely since re-using a
`payment_hash` is inherently insecure, e.g., it allows any node involved in
routing the first payment to grab the second payment. Alternatively we could
also reject the payment due to this protocol violation on the sender side,
however that'd be handing the funds to a (so far) well behaved node that
forwarded the payment despite being able to grab it, leaving the sender out of
pocket, and the recipient without the funds it could have gotten.

Since ad-hoc invoices do not need a `features` fields and `bolt11` string,
which in turn would require asking the `hsmd` for a signature, we made those
two fields nullable (changelog warning added).

Sending support for `keysend` payments will be added in a separate pull
request since it requires abstracting the internals of the `pay` plugin so we
can reuse them here.

As a side note: the keysend plugin will make sure it understands all the
mandatory TLV fields, and issue a `continue` otherwise. This was done in order
to allow alternative plugins to terminate keysends that have additional
information. One example is the `noise` plugin that terminates `keysend`
payments that are sent along a chat message (which has a mandatory TLV-type)
and associates the message with the incoming payment.

### DIFF
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -10,6 +10,9 @@ PLUGIN_FUNDCHANNEL_OBJS := $(PLUGIN_FUNDCHANNEL_SRC:.c=.o)
 PLUGIN_BCLI_SRC := plugins/bcli.c
 PLUGIN_BCLI_OBJS := $(PLUGIN_BCLI_SRC:.c=.o)
 
+PLUGIN_KEYSEND_SRC := plugins/keysend.c
+PLUGIN_KEYSEND_OBJS := $(PLUGIN_KEYSEND_SRC:.c=.o)
+
 PLUGIN_LIB_SRC := plugins/libplugin.c
 PLUGIN_LIB_HEADER := plugins/libplugin.h
 PLUGIN_LIB_OBJS := $(PLUGIN_LIB_SRC:.c=.o)
@@ -56,11 +59,13 @@ plugins/fundchannel: common/addr.o $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_LIB_OBJS)
 
 plugins/bcli: bitcoin/chainparams.o $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
+plugins/keysend: bitcoin/chainparams.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
+
 $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS): $(PLUGIN_LIB_HEADER)
 
 # Make sure these depend on everything.
-ALL_PROGRAMS += plugins/pay plugins/autoclean plugins/fundchannel plugins/bcli
-ALL_OBJS += $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS)
+ALL_PROGRAMS += plugins/pay plugins/autoclean plugins/fundchannel plugins/bcli plugins/keysend
+ALL_OBJS += $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS)
 
 check-source: $(PLUGIN_PAY_SRC:%=check-src-include-order/%) $(PLUGIN_AUTOCLEAN_SRC:%=check-src-include-order/%) $(PLUGIN_FUNDCHANNEL_SRC:%=check-src-include-order/%) $(PLUGIN_BCLI_SRC:%=check-src-include-order/%)
 check-source-bolt: $(PLUGIN_PAY_SRC:%=bolt-check/%) $(PLUGIN_AUTOCLEAN_SRC:%=bolt-check/%) $(PLUGIN_FUNDCHANNEL_SRC:%=bolt-check/%) $(PLUGIN_BCLI_SRC:%=bolt-check/%)
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -59,7 +59,7 @@ plugins/fundchannel: common/addr.o $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_LIB_OBJS)
 
 plugins/bcli: bitcoin/chainparams.o $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
-plugins/keysend: bitcoin/chainparams.o wire/gen_onion_wire.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
+plugins/keysend: bitcoin/chainparams.o wire/tlvstream.o wire/gen_onion_wire.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
 $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS): $(PLUGIN_LIB_HEADER)
 
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -59,7 +59,7 @@ plugins/fundchannel: common/addr.o $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_LIB_OBJS)
 
 plugins/bcli: bitcoin/chainparams.o $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
-plugins/keysend: bitcoin/chainparams.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
+plugins/keysend: bitcoin/chainparams.o wire/gen_onion_wire.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
 $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS): $(PLUGIN_LIB_HEADER)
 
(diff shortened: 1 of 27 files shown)


cdecker:
In order to avoid conflicts I [self-assigned featurebit `55`](https://github.com/lightningnetwork/lightning-rfc/issues/605#issuecomment-606679798) since whatsat does not seem to be announcing support at all.

cdecker:
Having a bit of trouble with these changes due to a missing dependency from `json_tok.h` to `wire/gen_onion_wire.h`:

```
In file included from plugins/keysend.c:2:
In file included from ./plugins/libplugin.h:16:
In file included from ./common/param.h:5:
In file included from ./common/json_tok.h:8:
./common/sphinx.h:13:10: fatal error: 'wire/gen_onion_wire.h' file not found
#include <wire/gen_onion_wire.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
<builtin>: recipe for target 'plugins/keysend.o' failed
make: *** [plugins/keysend.o] Error 1
```

Trying to debug that at the moment, but hints would be appreciated :wink:

darosior:
> Trying to debug that at the moment, but hints would be appreciated wink

But.. Didn't the common objects depends on the wire/ objects ? Could build with :
```patch
diff --git a/common/Makefile b/common/Makefile
index 9c24168f3..e04fc0b3a 100644
--- a/common/Makefile
+++ b/common/Makefile
@@ -78,7 +78,7 @@ COMMON_HEADERS_NOGEN := $(COMMON_SRC_NOGEN:.c=.h)     \
        common/jsonrpc_errors.h                         \
        common/overflows.h                              \
        common/status_levels.h
-COMMON_HEADERS_GEN := common/gen_htlc_state_names.h common/gen_status_wire.h common/gen_peer_status_wire.h
+COMMON_HEADERS_GEN := common/gen_htlc_state_names.h common/gen_status_wire.h common/gen_peer_status_wire.h wire/gen_onion_wire.h
 
 COMMON_HEADERS := $(COMMON_HEADERS_GEN) $(COMMON_HEADERS_NOGEN)
 COMMON_SRC := $(COMMON_SRC_NOGEN) $(COMMON_SRC_GEN)
```

cdecker:
> > Trying to debug that at the moment, but hints would be appreciated wink
> 
> But.. Didn't the common objects depends on the wire/ objects ? Could build with :
> 
> ```diff
> diff --git a/common/Makefile b/common/Makefile
> index 9c24168f3..e04fc0b3a 100644
> --- a/common/Makefile
> +++ b/common/Makefile
> @@ -78,7 +78,7 @@ COMMON_HEADERS_NOGEN := $(COMMON_SRC_NOGEN:.c=.h)     \
>         common/jsonrpc_errors.h                         \
>         common/overflows.h                              \
>         common/status_levels.h
> -COMMON_HEADERS_GEN := common/gen_htlc_state_names.h common/gen_status_wire.h common/gen_peer_status_wire.h
> +COMMON_HEADERS_GEN := common/gen_htlc_state_names.h common/gen_status_wire.h common/gen_peer_status_wire.h wire/gen_onion_wire.h
>  
>  COMMON_HEADERS := $(COMMON_HEADERS_GEN) $(COMMON_HEADERS_NOGEN)
>  COMMON_SRC := $(COMMON_SRC_NOGEN) $(COMMON_SRC_GEN)
> ```

No luck, I really tried adding the missing header almost everywhere, and the behaviour doesn't change.

darosior:
I deleted this comment after making use of my brain (not about linking objects but finding headers), sorry :sweat_smile:.

cdecker:
Interestingly this doesn't happen with `make -j 1`, but with increased parallelism it starts to happen (`make -j 8` is my usual command because I'm impatient). 

BTW this seems to work, but I have no idea why:

```
diff --git a/plugins/Makefile b/plugins/Makefile
index a15a1ff54..06bf35264 100644
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -51,6 +51,8 @@ PLUGIN_COMMON_OBJS :=				\
 	wire/fromwire.o				\
 	wire/towire.o
 
+plugins/keysend.o: wire/gen_onion_wire.h
+
 plugins/pay: bitcoin/chainparams.o $(PLUGIN_PAY_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
 plugins/autoclean: bitcoin/chainparams.o $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
```

darosior:
I experienced this too, both for the non-aggressive build and for the addition of the header file in plugins/Makefile (or common/Makefile).

plugins/Makefile
```diff
@@ -56,10 +59,12 @@ plugins/fundchannel: common/addr.o $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_LIB_OBJS)
...
 plugins/bcli: bitcoin/chainparams.o $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
 
+plugins/keysend: bitcoin/chainparams.o $(PLUGIN_KEYSEND_OBJS) $(PLUGIN_LIB_OBJS) $(PLUGIN_COMMON_OBJS) $(JSMN_OBJS) $(CCAN_OBJS)
+
 $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS): $(PLUGIN_LIB_HEADER)
 
 # Make sure these depend on everything.
-ALL_PROGRAMS += plugins/pay plugins/autoclean plugins/fundchannel plugins/bcli
+ALL_PROGRAMS += plugins/pay plugins/autoclean plugins/fundchannel plugins/bcli plugins/keysend
 ALL_OBJS += $(PLUGIN_PAY_OBJS) $(PLUGIN_AUTOCLEAN_OBJS) $(PLUGIN_FUNDCHANNEL_OBJS) $(PLUGIN_BCLI_OBJS) $(PLUGIN_LIB_OBJS)
```
darosior, line 68: Got it. You forgot to add `$(PLUGIN_FUNDCHANNEL_OBJS)` here. `ALL_OBJS` will later be set to depend on WIRE_HEADERS !

darosior: "later" <==> in the main Makefile

tests/test_pay.py
```diff
@@ -2880,3 +2880,25 @@ def test_reject_invalid_payload(node_factory):
 
     with pytest.raises(RpcError, match=r'WIRE_INVALID_ONION_PAYLOAD'):
         l1.rpc.waitsendpay(inv['payment_hash'])
+
+
+def test_keysend(node_factory):
+    # Use a temporary python plugin until we implement a native one
```
darosior, line 2886: Why not testing the real one directly when it's implemented ?

plugins/keysend.c
```diff
@@ -131,4 +131,7 @@ int main(int argc, char *argv[])
 	setup_locale();
 	plugin_main(argv, init, PLUGIN_RESTARTABLE, commands, ARRAY_SIZE(commands),
 	            NULL, 0, hooks, ARRAY_SIZE(hooks), NULL);
+	plugin_main(argv, init, PLUGIN_RESTARTABLE, NULL, commands,
+		    ARRAY_SIZE(commands), NULL, 0, hooks, ARRAY_SIZE(hooks),
```
darosior: Duplicated

plugins/keysend.c
```diff
@@ -3,6 +3,7 @@
 #include <wire/gen_onion_wire.h>
 
 #define PREIMAGE_TLV_TYPE 5482373484
+#define KEYSEND_FEATUREBIT 0x81
```
darosior, line 6: Why isn't it [55](https://github.com/lightningnetwork/lightning-rfc/issues/605#issuecomment-606679798) ?

plugins/keysend.c
darosior: nvm, resolved innext commit

darosior:
Maybe this deserves the `optech make me famous` tag ? :)

plugins/Makefile
cdecker: Whoa, that's awesome work, thanks a million :+1:

tests/test_pay.py
darosior: That was the dumbest comm ever

plugins/keysend.c
niftynei: I presume it's because 'bit 55' corresponds to 0x81 in hex.

rustyrussell:
Can we just do https://github.com/ElementsProject/lightning/pull/3628 and use that instead please?

Then we can remove the tlv option we don't want lightningd to see, and the rest is simple?

plugins/keysend.c
rustyrussell: .... that doesn't make sense, @niftynei ?  0x81 == 129.  I assume this was supposed to be 0x37?

cdecker:
Ok, I went ahead and changed the plugin to create the invoice itself, modify
the payload, and then telling `lightningd` to `continue` instead of
`resolve`. I dropped the nullability patch for some `invoice` fields which is
great, the downsides are that we now have to guess a bit more to give the
invoice a unique label (I went for a time-based suffix), and there is no way
for us to accept a payment with a duplicate `payment_hash`, potentially
leaving it to another node on the path (shouldn't happen, since the node could
have grabbed that payment on the forward path, but still...).

@rustyrussell's patch didn't work for me, so I disabled checking the payment
was terminated by looking at the invoice, but it's easy to re-enable once the
patch is working again.

cdecker:
Rebased on top of `master`, added @rustyrussell's commit allowing payload modifications, and fixed up the test (I forgot about the payload having a length prefix again). Should be good now. I'll just roll in the spurious fixup into the correct commit and then this is finally done.

darosior:
Succesfully [tested](https://twitter.com/darosior/status/1248332763683487745?s=20) this against Eclair's implementation (as a plugin, too!).

plugins/keysend.c
```diff
@@ -13,26 +14,58 @@ static void init(struct plugin *p, const char *buf UNUSED,
...
 	response = jsonrpc_stream_success(cmd);
+
 	json_add_string(response, "result", "continue");
+	if (payload) {
+		binpayload = tal_arr(cmd, u8, 0);
+		towire_tlvstream_raw(&binpayload, payload->fields);
+		rawpayload = tal_arr(cmd, u8, 0);
+		towire_bigsize(&rawpayload, tal_bytelen(binpayload));
+		towire(&rawpayload, binpayload, tal_bytelen(binpayload));
+		json_add_string(response, "payload", tal_hex(cmd, rawpayload));
```
rustyrussell, line 31: It seems a bit weird to prepend the length here, doesn't it?  Seems like we're exposing a (bad) internal detail?

Ah, never mind, this allows us to put in legacy payloads instead I guess.  Should we remove this ability though, since it's going to be v. ugly when legacy gets removed?

cdecker: Either works for me, legacy payloads need to die alright :+1:

cdecker:
Current status:

 - Fixed the use-after-free issue due to sibling allocations being free'd in an unexpected order
 - I'll fix up the payload replacement, should we decide to drop the length prefix, which is really a bit weird, but allows us to replace with legacy payloads

rustyrussell:
Some timeouts, retrying (it could be that adding YA plugin is making us too slow under VALGRIND).

rustyrussell:
OK, I have pushed one final change, to clean up the payload API.

rustyrussell:
Fixed up test broken by my final patch, and trivial rebase onto master

rustyrussell:
Also, reordered slightly: the previous series had us using the "payload" return from the hook before implementing it.

cdecker:
Rebased on top of #3647, hopefully now it doesn't fail when replacing the payload anymore.

rustyrussell:
Trivial rebase.

Ack ea931f09219b93cab1f5d2c86cd8befcd177a6e1

whitslack:
Should this plugin have been added to `PLUGINS` in `Makefile`? As is, the `keysend` binary is not installed into `$(plugindir)`.

PR #3612, opened 2020-04-01
Wumbo channel support

rustyrussell:
This adds a `--wumbo` (or, more prosiacly) a `--large-channels` option (off by default) which means:

1. We advertize the new `option_support_large_channel` from the spec.
2. We will open, and allow others to open, channels >= 2^24 satoshis if they also offer the feature.
3. Fundchannel 'all' will also not cap funds, if we and the peer offer this feature.

Before merging, someone should interop this with Eclair: in particular, it's the first feature we advertize in the channel_announcement, and we can't advertize the channel if we don't both agree on how that's represented!


common/features.c
```diff
@@ -203,9 +153,11 @@ static void clear_feature_bit(u8 *features, u32 bit)
  *  - MUST set `len` to the minimum length required to hold the `features` bits
  *  it sets.
  */
-u8 *get_agreed_channelfeatures(const tal_t *ctx, const u8 *theirfeatures)
+u8 *get_agreed_channelfeatures(const tal_t *ctx,
+			       const struct feature_set *ours,
+			       const u8 *theirfeatures)
```
cdecker, line 162: Nit: inconsistent naming, should be `theirs` and `ours, or `theirfeatures` and `ourfeatures`.

plugins/libplugin.c
```diff
@@ -710,11 +719,37 @@ static struct io_plan *rpc_conn_init(struct io_conn *conn,
...
+		if (json_tok_streq(buf, t, "init"))
+			p = INIT_FEATURE;
+		else if (json_tok_streq(buf, t, "node"))
+			p = NODE_ANNOUNCE_FEATURE;
+		else if (json_tok_streq(buf, t, "channel"))
+			p = CHANNEL_FEATURE;
+		else if (json_tok_streq(buf, t, "invoice"))
+			p = BOLT11_FEATURE;
+		else
+			continue;
```
cdecker, line 741: This should likely at least produce a warning that the plugin is not in sync with `lightningd`

rustyrussell: I figure it's always legal to add  a field.  Though perhaps libplugin should consider itself more tightly bound?

rustyrussell:
OK, I fixed some tests, and did a sweep to rename all the feature sets to "our_features" or "their_features" where appropriate.

rustyrussell:
Ack 6ea51f071c492ce25e2585a93cefac4402cf4173

PR #3603, opened 2020-03-25
v082 deprecations

rustyrussell:
Not everything was in the Deprecated section of the changelog, so did a grep sweep.

PR #3615, opened 2020-04-02
Fix fundchannel in deprecated-apis=false mode ('satoshi' param in fundchannel_start)

darosior:
Noticed while testing #3611 https://github.com/ElementsProject/lightning/pull/3603/commits/49e21538cc99e8f7340d174f02b853cc582dfd2a kept `satoshi` but did not make `fundchannel_start` accept it again.

Note that we run the functional tests with deprecated-apis=false so we did not catch this, but users run with deprecated-apis=true..

### DIFF
--- a/plugins/fundchannel.c
+++ b/plugins/fundchannel.c
@@ -270,8 +270,6 @@ static struct command_result *fundchannel_start(struct command *cmd,
 
 	json_add_string(req->js, "id", node_id_to_hexstr(tmpctx, fr->id));
 
-	if (deprecated_apis)
-		json_add_string(req->js, "satoshi", fr->funding_str);
 	json_add_string(req->js, "amount", fr->funding_str);
 
 	if (fr->feerate_str)


darosior:
> Note that we run the functional tests with deprecated-apis=false so we did not catch this, but users run with deprecated-apis=true..

Cf #3616

rustyrussell:
No, it was the *output* parameter satoshi we can't deprecate. I definitely meant to deprecate the input.

Was there a reason to defer this?

darosior:
Ah, ok then it becomes way simpler :-) Amended the commit.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3615/commits/eced481f2edf0d98c32cd63dad8b40ad983d8e6c

cdecker:
Notice that the help text in the json command further down still mentions `satoshi` (wrongfully since the param was always called `amount`). I'll add that to my cleanups for this weak.

darosior:
> Notice that the help text in the json command further down still mentions satoshi (wrongfully since the param was always called amount). I'll add that to my cleanups for this weak.

Thanks

PR #3616, opened 2020-04-02
pyln-testing: Also test against deprecated apis

darosior:
This adds a new environment var, `DEPRECATED_APIS`, to allow to quickly run the functional tests against the `lightningd` API users will likely use.

I wanted to also add a Travis job, but am not sure if it's possible to run another one at all. At least we can still set this locally.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/ab4d0de25bced8837d03f9d59dc978619c96659a

PR #3620, opened 2020-04-04
doc: document the plugin interface to gather Bitcoin data

### LINKED ISSUE #3354: Bitcoin backend plugins planning

darosior:
This finally closes #3354 by making the current API final.

There are still things that can get added to the process (namely, notifications for a more optimized polling) but I don't think we'd need to break the interface again (soon).

I have resigned from using WU for transaction sizes as most of the Bitcoin data source seem to actually prefer round ups, and that would have resulted in boutique transformation in plugins to pass from btc/kWU to btc/kVB..

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -939,6 +939,81 @@ This will ensure backward
 compatibility should the semantics be changed in future.
 
 
+
+## Bitcoin backend
+
+C-lightning communicates with the Bitcoin network through a plugin. It uses the
+`bcli` plugin by default but you can use a custom one, multiple custom ones for
+different operations, or write your own for your favourite Bitcoin data source!
+
+Communication with the plugin is done through 5 JSONRPC commands, `lightningd`
+can use from 1 to 5 plugin(s) registering these 5 commands for gathering Bitcoin
+data. Each plugin must follow the below specification for `lightningd` to operate.
+
+
+### `getchainfo`
+
+Called at startup, it's used to check the network `lightningd` is operating on and to
+get the sync status of the backend.
+
+The plugin must respond to `getchainfo` with the following fields:
+    - `chain` (string), the network name as introduced in bip70
+    - `headercount` (number), the number of fetched block headers
+    - `blockcount` (number), the number of fetched block body
+    - `ibd` (bool), whether the backend is performing initial block download
+
+
+### `estimatefees`
+
+Polled by `lightningd` to get the current feerate, all values must be passed in BTC/kVB.
+
+If fee estimation fails, the plugin must set all the fields to `null`.
+
+The plugin, if fee estimation succeeds, must respond with the following fields:
+    - `opening` (number), used for funding and also misc transactions
+    - `mutual_close` (number), used for the mutual close transaction
+    - `unilateral_close` (number), used for unilateral close (/commitment) transactions
+    - `delayed_to_us` (number), used for resolving our output from our unilateral close
+    - `htlc_resolution` (number), used for resolving HTLCs after an unilateral close
+    - `penalty` (number), used for resolving revoked transactions
+    - `min` (number), used as the minimum acceptable feerate
+    - `max` (number), used as the maximum acceptable feerate
+
+
+### `getrawblockbyheight`
+
+This call takes one parameter, `height`, which determines the block height of
+the block to fetch.
+
+The plugin must set all fields to `null` if no block was found at the specified `height`.
+
+The plugin must respond to `getrawblockbyheight` with the following fields:
+    - `blockhash` (string), the block hash as a hexadecimal string
+    - `block` (string), the block content as a hexadecimal string
+
+
+### `getutxout`
+
+This call takes two parameter, the `txid` (string) and the `vout` (number)
+identifying the UTXO we're interested in.
+
+The plugin must set both fields to `null` if the specified TXO was spent.
+
+The plugin must respond to `gettxout` with the following fields:
+    - `amount` (number), the output value in **sats**
+    - `script` (string), the output scriptPubKey
+
+
+### `sendrawtransaction`
+
+This call takes one parameter, a string representing a hex-encoded Bitcoin
+transaction.
+
+The plugin must broadcast it and respond with the following fields:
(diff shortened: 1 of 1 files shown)


fiatjaf:
I very much like the `estimatefees` call. That means I can decide to accept if a peer for any reason wants to pay a very high or low fee by just returning a very large number in `max` or `min`?

What happens if I set all fields to `null`?

darosior:
> That means I can decide to accept if a peer for any reason wants to pay a very high or low fee by just returning a very large number in max or min?

Indirectly, kind of: you can decide of the flexibility you give to your peer regarding transaction fees, but you *should* prefer using `--ignore-fee-limits` for this usecase.

> What happens if I set all fields to null?

That means your plugin [could not estimate](https://github.com/ElementsProject/lightning/pull/3620/commits/152aa26b8afd99a7c11877afb6114d6361dbaf97#diff-ee450b3fa6bc707a38ce6dfb6ce7611eR970) the fees.

doc/PLUGINS.md
```diff
@@ -939,6 +939,81 @@ This will ensure backward
...
+    - `script` (string), the output scriptPubKey
+
+
+### `sendrawtransaction`
+
+This call takes one parameter, a string representing a hex-encoded Bitcoin
+transaction.
+
+The plugin must broadcast it and respond with the following fields:
+    - `succes` (boolean), which is `true` if the broadcast succeeded
```
niftynei, line 1013: succes -> success

fiatjaf:
> > What happens if I set all fields to null?
>
> That means your plugin [could not estimate](https://github.com/ElementsProject/lightning/pull/3620/commits/152aa26b8afd99a7c11877afb6114d6361dbaf97#diff-ee450b3fa6bc707a38ce6dfb6ce7611eR970) the fees.

The link doesn't lead anywhere. What happens when the plugin cannot estimate the fees?

darosior:
> The link doesn't lead anywhere. What happens when the plugin cannot estimate the fees?

Yep, because force push. When the plugin cannot estimate fees the feerate is set to:
- Floor (253perkw) if mainnet and not in developer mode
- 0 if testnet and not developer mode
- 0 if developer mode

PR #3622, opened 2020-04-05
lightningd: don't log BROKEN when we don't have a channel update.

rustyrussell:
This happened on my testnet node because I've been failing to reconnect to
a node which created a channel and never exchanged announcement sigs.

Changelog-None (change was introduced since last release)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3622/commits/2357c43630b74565a35467cc2decca39f85849d7

PR #3623, opened 2020-04-05
HTLC blinded route support (EXPERIMENTAL)

rustyrussell:
This is an implementation of https://github.com/lightningnetwork/lightning-rfc/blob/route-blinding/proposals/route-blinding.md (only with --enable-experimental-features).

In this method, the vendor provides a blinded path similar to route hints, but the node ids are blinded, and instead of short_channel_ids for each hop, it provides an encrypted blob.  This can be used for (finally, genuinely!) private channels as well as fully private vendor paths.  Every node on the blinded path needs to be upgraded, however.

This is the minimum implementation to allow interop testing (e.g. with @t-bast when he gets Eclair done).  There are five things remaining to be done:

1. We need to advertize a feature bit (simple to fix)
2. We currently use a dummy short_channel_id field, which is unnecessary in blinded paths (a bit more complex, since our internals assume a short_channel_id field in every hop).
3. We don't propagate errors properly (we need to decide how we'll do this).
4. We haven't yet defined a bolt11 encoding for blinded paths.
5. We don't validate padded entries in the blinding (we could add them, but we need to check they're valid, otherwise the payer can corrupt them and discover if it was a dummy entry).

It's probably premature to add those until we have a solid spec (esp. how to encode blinded routes in bolt11; presumably this will involve emojis!).

Changelog-None: Experimental only


t-bast:
I did a quick pass on this, nothing looks too weird but my C skills are quite rusty (bad pun intended).
I'll get a prototype working on eclair this week.

On the proposal itself, there are three main areas on which I'd like more eyes on:

- the crypto itself, and whether we're missing an abstraction (it's really a kind of reverse sphinx)
- whether we can live with some `update_add_htlc` that have a different length than other `update_add_htlc` (which kind of gives away the blinded path if you're watching the network)
- error management: how do we ensure errors don't leak the blinded path? Obfuscation + delayed responses may do the trick, but is it enough (and is it necessary)?

I'll turn my blinded path into a draft PR so that others can start having a look at it.

t-bast:
I have a very rough prototype for eclair here: https://github.com/ACINQ/eclair/tree/route-blinding
It should be enough for some interop testing, I can generate a shared test vector if you want.
It made me realize that there are a few places where the spec proposal is vague, and implementation could slightly differ.

t-bast:
> commit 8c960cd is showing as empty/contentless for me?

Because [the best code is no code at all](https://github.com/kelseyhightower/nocode)

rustyrussell:
> commit [8c960cd](https://github.com/ElementsProject/lightning/commit/8c960cde4cff824f9b94917b3b2b0aec51807e29) is showing as empty/contentless for me?

Huh... not sure how that happened.  That's why backups FTW, one moment...

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3623/commits/74c1f9e5e43f3acf5665554c2ddf4e9eeca92d09

niftynei:
experimental means no changelog required.

PR #3627, opened 2020-04-07
devtools/decodemsg: don't crash if we have multiple tlv options.

rustyrussell:
We call tal_count(msg) after we've moved msg, and that causes an abort.


niftynei:
ACK 05cd3ec427951349d8a7cb719d0716b5cbfd0018

niftynei:
no changelog needed.

Issue #3607, opened 2020-03-28
`pay` crashes because "cannot find <unannounced peer>"


fiatjaf:
### Issue and Steps to Reproduce

You can use the invoice below to reproduce (but be sure to apply the parameters):

```
~> lightning-cli pay -k bolt11=lnbc3252500p1p087he3pp57t693zw9adv0a8lvv99zwvs7cl8ed0pkgul27jp09cra355vme9qhp5cv9ldqyn558mp063ays4n2af99cx2w787mlaqy0re75mn4mkpfeqxqyz5vqrzjq2ldrqfd8qj00nzve5ud5htx520ulmq5dl54ufkd9cxnlycdv5ag6zx804nsx6grqqqqqyy6qqqqqqqqqgsp5qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq9qy9qqqmxrzcgwy7qadlmgv3dmrrsl2e87dcxg7wdhpfs5vevkl4sy34ljpajmq7u7cjvj2r3cxxvxcktputmhmed7908jxqwfsxhgzluqnrcqp74xjh9 riskfactor=3 maxfeepercent=0.4 exemptfee=3000
```

That causes the following error:

```
2020-03-28T14:30:25.570Z INFO gossipd: find_route: cannot find 032cbe7d2b0147b494424505e17e6878dfe18bc6e6e68d7b33ed89e525094c6cdc
pay: FATAL SIGNAL 11 (version v0.8.0-371-g8c984fb)
0x55e5f542511b send_backtrace
        common/daemon.c:39
0x55e5f54251c1 crashdump
        common/daemon.c:52
0x7f83e1561f1f ???
        ???:0
0x55e5f5426b92 json_get_member
        common/json.c:299
0x55e5f5412e2e maybe_exclude
        plugins/pay.c:701
0x55e5f541350c getroute_done
        plugins/pay.c:793
0x55e5f5417c7c handle_rpc_reply
        plugins/libplugin.c:542
0x55e5f5418244 rpc_read_response_one
        plugins/libplugin.c:645
0x55e5f5418335 rpc_conn_read_response
        plugins/libplugin.c:664
0x55e5f5438de7 next_plan
        ccan/ccan/io/io.c:59
0x55e5f5439964 do_plan
        ccan/ccan/io/io.c:407
0x55e5f54399a2 io_ready
        ccan/ccan/io/io.c:417
[...]

PR #3630, opened 2020-04-07
Fixup pay crashes

niftynei:
Fix a couple little crashes found by @fiatjaf

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -262,7 +262,7 @@ static struct command_result *waitsendpay_expired(struct command *cmd,
 	for (size_t i = 0; i < tal_count(pc->ps->attempts); i++) {
 		json_object_start(data, NULL);
 		if (pc->ps->attempts[i].route)
-			json_add_member(data, "route", false, "%s",
+			json_add_member(data, "route", true, "%s",
 					 pc->ps->attempts[i].route);
 		json_out_add_splice(data->jout, "failure",
 				    pc->ps->attempts[i].failure);
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -698,6 +698,9 @@ static bool maybe_exclude(struct pay_command *pc,
 {
 	const jsmntok_t *scid, *dir;
 
+	if (!route)
+		return false;
+
 	scid = json_get_member(buf, route, "channel");
 
 	if (node_or_channel_in_routehint(pc->plugin,


plugins/pay.c
```diff
@@ -262,7 +262,7 @@ static struct command_result *waitsendpay_expired(struct command *cmd,
 	for (size_t i = 0; i < tal_count(pc->ps->attempts); i++) {
 		json_object_start(data, NULL);
 		if (pc->ps->attempts[i].route)
-			json_add_member(data, "route", false, "%s",
+			json_add_member(data, "route", true, "%s",
```
rustyrussell, line 265: Hmm, this looks like json_add_string() would be better?  add_member is pretty low-level...

rustyrussell:
But I don't understand *how* this happens.

I tried the example in #3607 (and hacked my node to ignore expiry).  It didn't crash.

It took a long time, however: the routehint itself requires more fee than we're prepared to pay, and we flail through excluding everything else until we run out of routes, but it does eventually.

In particular, find_worst_channel() can only return NULL if there is less than two hops in the route.  Which implies it could only happen if the node was a peer, and for some reason we didn't select it first due to the routeboost.  Let me try that...

rustyrussell:
OK, I can't reproduce this.  In theory, we could add a channel between pay attempts, but that wouldn't be reproducible like this seems to be.

rustyrussell:
OK, @fiatjaf  notes that he *is* a peer in #3607 and I finally reproduced.  Now to track down...

rustyrussell:
@niftynei OK, your fix is correct.  I've got another PR for the test though.

rustyrussell:
Ack 069ffc9f5ace94fc95b5125901368ef9a5eda539

plugins/pay.c
darosior: +1, but fixed in https://github.com/ElementsProject/lightning/pull/3632

fiatjaf:
Oops, sorry for the irreproducibility, I forgot invoices expired. I tried to reproduce myself (without hacking my node to ignore the expiry) and failed too.

Also sorry for implying that I was a peer. I wasn't a peer of the unannounced channel, the unannounced channel doesn't really exist. In fact I was a peer of the last peer before the unannounced channel, but I didn't think this was important when creating the issue. And it seems Rusty found that out before I could even realize.

Anyway, thank you all for looking at this super corner case.

Issue #3341, opened 2019-12-13
lnd ignores update_fee on new connection?


rustyrussell:
I funded a channel with 03634bda49c9c42afd876d8288802942c49e58fbec3844ff54b46143bfcb6cdfaf (Zap Technologies) which seems to be an lnd node based on its features.

We create the commit tx at low feerate (1006):

0200000001d1051af7c4e2218cf21f1b450015148f134626cf4018fcc1580d898157dbc41501000000000e5f108001683f0f000000000016001456c9b57de7610ed4a7721cb5a7797e13ce53d9f47e5a5c20 

All goes well, funding tx gets mined and we exchange funding_locked:

```
channeld-chan#17314: peer_in WIRE_FUNDING_LOCKED
channeld-chan#17314: peer_out WIRE_FUNDING_LOCKED
```
Then we immediately send update_fee (3767), followed by the signature on the updated tx:
0200000001d1051af7c4e2218cf21f1b450015148f134626cf4018fcc1580d898157dbc41501000000000e5f10800199370f000000000016001456c9b57de7610ed4a7721cb5a7797e13ce53d9f47f5a5c20

```
channeld-chan#17314: peer_out WIRE_UPDATE_FEE
channeld-chan#17314: peer_out WIRE_COMMITMENT_SIGNED
```

LND rejects it, using a tx with the *old* feerate:

0200000001d1051af7c4e2218cf21f1b450015148f134626cf4018fcc1580d898157dbc41501000000000e5f108001683f0f000000000016001456c9b57de7610ed4a7721cb5a7797e13ce53d9f47f5a5c20

Complete logs below:
```
2019-12-13T06:07:06.998Z DEBUG 03634bda49c9c42afd876d8288802942c49e58fbec3844ff54b46143bfcb6cdfaf-connectd: Connect OUT
2019-12-13T06:07:07.001Z DEBUG 03634bda49c9c42afd876d8288802942c49e58fbec3844ff54b46143bfcb6cdfaf-connectd: peer_out WIRE_INIT
2019-12-13T06:07:07.224Z DEBUG 03634bda49c9c42afd876d8288802942c49e58fbec3844ff54b46143bfcb6cdfaf-connectd: peer_in WIRE_INIT
2019-12-13T06:07:07.310Z DEBUG 03634bda49c9c42afd876d8288802942c49e58fbec3844ff54b46143bfcb6cdfaf-openingd-chan#17314: pid 17015, msgfd 56
[...]

Issue #3596, opened 2020-03-18
Multiple peers unilaterally closing immediately upon new channel confirmation: "rejected commitment"


whitslack:
### Issue and Steps to Reproduce
I opened twelve channels in quick succession, and five of them were unilaterally closed by the peer as soon as they reached 3 confirmations. The debug logs for all five contain a message that says `rejected commitment` with an `invalid_commit_sig`. Also I see `peer_in WIRE_ERROR`, which to me says that the peer didn't like something my node did.

Here are the debug logs from four of the five failed channels. (The fifth log contains many megabytes of channel update messages.) I'm not sure why there are so many `SKIPPED` log events scattered throughout them, but maybe there is enough intact information across all of them to astertain what went wrong.

* [`03d480ac7ad220371f1b9c2560be7eb8da6335d6a0d8cb1bc27db1a3958fd8c73f.log`](https://github.com/ElementsProject/lightning/files/4346846/03d480ac7ad220371f1b9c2560be7eb8da6335d6a0d8cb1bc27db1a3958fd8c73f.log)
* [`029edbb304c7a00fbe9c912eebb8a865afe96a19015e0548e8be1691f965f1ca97.log`](https://github.com/ElementsProject/lightning/files/4346847/029edbb304c7a00fbe9c912eebb8a865afe96a19015e0548e8be1691f965f1ca97.log)
* [`0351876efe01167a1f81f988ef470311eca17661dd1994bad01861d4a54565bc5c.log`](https://github.com/ElementsProject/lightning/files/4346848/0351876efe01167a1f81f988ef470311eca17661dd1994bad01861d4a54565bc5c.log)
* [`03271338633d2d37b285dae4df40b413d8c6c791fbee7797bc5dc70812196d7d5c.log`](https://github.com/ElementsProject/lightning/files/4346849/03271338633d2d37b285dae4df40b413d8c6c791fbee7797bc5dc70812196d7d5c.log)

For the sake of making this searchable, here's a summary of the sequence of events I see in these logs. These might be slightly out of order because I had to piece them together from multiple logs due to all the `SKIPPED` events.

1. `Funding tx … depth 3 of 3`
1. `State changed from CHANNELD_AWAITING_LOCKIN to CHANNELD_NORMAL`
1. `peer_out WIRE_FUNDING_LOCKED`
1. `Setting REMOTE feerate to 17890`
[...]

PR #3634, opened 2020-04-08
channeld: defer first update_fee until we have an HTLC to send.

rustyrussell:
Sending update_fee immediately after channel establishment seems to
upset LND, so work around it by deferring it.  The reason we increase
the fee after establishment is because now we might need to close the
channel in a hurry due to htlcs, but until there are htlcs that's
unnecessary.

Workaround for #3596

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -2749,7 +2749,12 @@ static void handle_feerates(struct peer *peer, const u8 *inmsg)
 	 */
 	if (peer->channel->funder == LOCAL) {
 		peer->desired_feerate = feerate;
-		start_commit_timer(peer);
+		/* Don't do this for the first feerate, wait until something else
+		 * happens.  LND seems to get upset in some cases otherwise:
+		 * see https://github.com/ElementsProject/lightning/issues/3596 */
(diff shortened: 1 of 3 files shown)


darosior:
I think  this fixes https://github.com/ElementsProject/lightning/issues/3341 too.

cdecker:
Seems some more tests require porting to the new behavior:

 - `test_update_fee_reconnect`: this one seems to actually crash c-lightning

darosior:
> test_update_fee_reconnect: this one seems to actually crash c-lightning

iirc in this one we make `bitcoind` fail to answer fee est, so I guess because we wait for the `UPDATE_FEE` log `bcli` will timeout on failed attempts (and throw a lovely backtrace) before `TIMEOUT`.

rustyrussell:
Hmm, somehow I missed that failure.  Easy to fix, just added a HTLC to make update_fee work.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3634/commits/462e62e9c7a43a3464892f0d5a36d90c7b0daada

PR #3635, opened 2020-04-08
pyln-testing: specify fallbackfee for newer versions of bitcoind

darosior:
The last versions (I think since 0.19.0.1 but not sure) of `bitcoind` disabled the fallbackfee by default, it screwed up my scripts but I just noticed `pyln` too now I installed it globally..

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -1,3 +1,4 @@
+from bitcoin.core import COIN
 from bitcoin.rpc import RawProxy as BitcoinProxy
 from pyln.client import RpcError
 from pyln.testing.btcproxy import BitcoinRpcProxy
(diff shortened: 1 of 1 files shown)


rustyrussell:
I'm not sure: do we want bitcoind lying to us about what the feerate is?  Perhaps we do this for this release, then handle it properly in the next release?

darosior:
> do we want bitcoind lying to us about what the feerate is?

It's not lying to us about the feerate (even if we already do this ourselves by mocking `estimatesmartfee`), this is really internal to `bitcoind` : it won't give us this feerate when asking for fee estimate but will use it in last resort, for example, for `sendtoaddress`.

This is the old default behaviour, removed in https://github.com/bitcoin/bitcoin/pull/16524. I just thought putting it to the old default value was unnecessarily high for regtest.

> Perhaps we do this for this release, then handle it properly in the next release?

"Cleaner" to `fallbackfee` would be to fill `bitcoind`'s fee estimation buckets : it's quite a burden, and I'm not sure it's necessary ?

rustyrussell:
Ah, cool, I missed that this was just for testing.

Thanks!

Ack https://github.com/ElementsProject/lightning/pull/3635/commits/dbe1ecdf0193b4eb9552166f86cba2411942523a

PR #3636, opened 2020-04-08
Weekly cleanups

cdecker:
- Fixes a wrongly interpreted multiplier-less amount
 - Allow directory reuse in `node_factory` (needed by some plugins that prepare the node directory before the node first starts, e.g., the backup plugin dropping a `backup.lock` file so the backend is ensured to stay the same).
 - Remove the `satoshi` parameter from the `fundchannel_start` docs
 - Add a demo of how the wallet DSN looks like for postgresql

### DIFF
--- a/common/bolt11.c
+++ b/common/bolt11.c
@@ -616,7 +616,7 @@ struct bolt11 *bolt11_decode(const tal_t *ctx, const char *str,
 		 */
 		b11->msat = NULL;
 	} else {
-		u64 m10 = 10;
+		u64 m10 = 10 * MSAT_PER_BTC; /* Pico satoshis in a Bitcoin */
 		u64 amount;
 		char *end;
 
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -182,6 +182,23 @@ Identify the location of the wallet. This is a fully qualified data source
 name, including a scheme such as `sqlite3` or `postgres` followed by the
 connection parameters.
 
+The default wallet corresponds to the following DSN:
+
+```
+--wallet=sqlite3://$HOME/.lightning/bitcoin/lightningd.sqlite3
+```
+
+The following is an example of a postgresql wallet DSN:
+
+```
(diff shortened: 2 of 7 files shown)


doc/lightningd-config.5.md
```diff
@@ -182,6 +182,23 @@ Identify the location of the wallet. This is a fully qualified data source
...
+```
+
+The following is an example of a postgresql wallet DSN:
+
+```
+--wallet=postgres://user:pass@localhost:5432/db_name
+```
+
+This will connect to a the DB server running on `localhost` port `5432`,
+authenticate with username `user` and password `pass`, and ten use the
```
niftynei, line 198: ten -> then

common/bolt11.c
```diff
@@ -616,7 +616,14 @@ struct bolt11 *bolt11_decode(const tal_t *ctx, const char *str,
...
 	} else {
-		u64 m10 = 10;
+		/* BOLT #11:
+		 *
+		 * 2. amount: optional number in that currency, followed by an
+		 * optional `multiplier` letter. The unit encoded here is the
+		 * 'social' convention of a payment unit -- in the case of
+		 * Bitcoin the unit is 'bitcoin' NOT satoshis.
+		 */
+		u64 m10 = 10E12; /* Pico satoshis in a Bitcoin */
```
niftynei, line 626: It seems like you've got an off by 10x issue here. 
```
>       assert(b11['msatoshi'] == 7*10**11)
E       assert 7000000000000 == 700000000000
E         -7000000000000
E         +700000000000
```

rustyrussell: Yes, this should be `10 * MSAT_PER_BTC` (which follows the format of the `multipliers` table).

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3636/commits/fab8ad6788bc92dd1eac52e74d96685f347d0c2e

niftynei:
missing changelog for bolt decoding change?

cdecker:
Added changelog for the bolt11 fix.

PR #3637, opened 2020-04-08
python: Consolidate requirements.txt files in a single place

### LINKED ISSUE #3518: requirements.txt files should be consolidated
@cdecker points out we should have one to rule them all.

cdecker:
We had them split according the separate use-cases:

 - testing
 - doc-gen
 - wire-gen

But that was causing new contributors to miss some dependencies when they
first got hacking. So this consolidates all of our own dependencies in a root
requirements.txt, with the notable exception of `pyln-client`, `pyln-testing`
and `pyln-proto` which are distributed as PyPI modules and therefore have
their own dependencies that need to be tracked in the module root.

Fixes #3518

Changelog-None


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3637/commits/5fab2233bdbb423e1b55a57d468dfb65d99431d3

jb55:
ACK, I remember getting tripped up on this a few times

PR #3639, opened 2020-04-09
plugin: Ensure RPC passthrough calls are terminated when plugin dies

### LINKED ISSUE #3496: Hooks abandon their payloads when plugins stopped

In particular, this can lead to stuck HTLCs:

```python
def test_hook_chaining_dynamic(node_factory, bitcoind):
    """Test what happens when we remove a plugin while a hook in use"""
    l1, l2 = node_factory.line_graph(2, opts={'may_reconnect': True})

    # We want this to fail, so we can keep retrying.
    routestep = {'msatoshi': 124, 'id': l2.info['id'], 'delay': 5, 'channel': l1.get_channel_scid(l2)}
    l2.rpc.plugin_start(os.path.join(os.path.dirname(__file__), "plugins/hold_htlcs.py"))

    l1.rpc.sendpay([routestep], '00' * 32)
    l2.daemon.wait_for_log('hold_htlcs.py: Holding onto an incoming htlc for 10 seconds')
    l2.rpc.plugin_stop('hold_htlcs.py')

    # Restarting gets things unstuck!  Uncomment to see expected HTLC failure
    # l2.restart()

    # The effect is that we hold onto the HTLC forever.  Nasty, but we should do
    # OK if we have to abandon it.
    bitcoind.generate_block(9)
    l1.daemon.wait_for_log('hit deadline')
    
    bitcoind.generate_block(5, wait_for_mempool=1)
    
    with pytest.raises(RpcError, match=r"WIRE_PERMANENT_CHANNEL_FAILURE"):
        l1.rpc.waitsendpay('00' * 32)
```

cdecker:
We now track all pending RPC passthrough calls, and terminate them with an
error if the plugin dies.

Changelog-Fixed: JSON-RPC: Pending RPC method calls are now terminated if the handling plugin exits prematurely.

Closes #3496

### DIFF
--- a/common/jsonrpc_errors.h
+++ b/common/jsonrpc_errors.h
@@ -25,6 +25,9 @@ static const errcode_t PARAM_DEV_ERROR = -2;
 /* Plugin returned an error */
 static const errcode_t PLUGIN_ERROR = -3;
 
+/* Plugin terminated while handling a request. */
+static const errcode_t PLUGIN_TERMINATED = -4;
+
 /* Errors from `pay`, `sendpay`, or `waitsendpay` commands */
 static const errcode_t PAY_IN_PROGRESS = 200;
 static const errcode_t PAY_RHASH_ALREADY_USED = 201;
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -23,6 +23,15 @@
  * `getmanifest` call anyway, that's what `init` is for. */
 #define PLUGIN_MANIFEST_TIMEOUT 60
 
+/* A simple struct associating an incoming RPC method call with the plugin
+ * that is handling it and the downstream jsonrpc_request. */
+struct plugin_rpccall {
+	struct list_node list;
+	struct command *cmd;
+	struct plugin *plugin;
+	struct jsonrpc_request *request;
+};
+
 #if DEVELOPER
 static void memleak_help_pending_requests(struct htable *memtable,
 					  struct plugins *plugins)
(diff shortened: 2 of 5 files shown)


darosior:
ut ACK 256f0652ac6a04c537acdb2a785f61721a729716

lightningd/plugin.c
```diff
@@ -49,8 +58,17 @@ struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
...
 	plugin_hook_unregister_all(p);
 	list_del(&p->list);
+
+	/* Terminate all pending RPC calls with an error. */
+	list_for_each(&p->pending_rpccalls, call, list)
+	{
+		was_pending(command_fail(
+		    call->cmd, PLUGIN_PENDING_REQUEST_DIED,
+		    "Plugin terminated before replying to RPC call."));
+	}
```
niftynei, line 71: does the p->pending_calls list need to be cleaned up?

cdecker: No, it's allocated off of the plugin, and therefore will be freed as a dependency when we free the `plugin` (which happens to be our signal to die).

lightningd/plugin.c
```diff
@@ -49,8 +58,17 @@ struct plugins *plugins_new(const tal_t *ctx, struct log_book *log_book,
 
 static void destroy_plugin(struct plugin *p)
 {
+	struct plugin_rpccall *call;
 	plugin_hook_unregister_all(p);
 	list_del(&p->list);
+
+	/* Terminate all pending RPC calls with an error. */
+	list_for_each(&p->pending_rpccalls, call, list)
+	{
```
rustyrussell, line 67: Minor: formatting is usually `{` at end of the above line.

common/jsonrpc_errors.h
```diff
@@ -64,4 +64,6 @@ static const errcode_t INVOICE_WAIT_TIMED_OUT = 904;
 /* Errors from HSM crypto operations. */
 static const errcode_t HSM_ECDH_FAILED = 800;
 
+static const errcode_t PLUGIN_PENDING_REQUEST_DIED = 1000;
```
rustyrussell, line 67: I don't think we can make this a normal error; I think PLUGIN_ERROR or maybe a new PLUGIN_DIED (with associated negative code) is more appropriate?

cdecker:
Addressed all comments, renamed the error name and number to `PLUGIN_TERMINATED` and `-4` so it is a bit nicer now.

PR #3640, opened 2020-04-09
pyln-client: send proper JSONRPC2 errors on failed call

### LINKED ISSUE #3638: Result for missing required arguments in plugins

The output when a Python plugin misses a required argument does not follow the same UI as with built-in RPC method. For example:

```
l1-cli plugin-method
"Error while processing plugin-method: TypeError(\"missing a required argument: 'arg'\")"
```

While you would expect something similar to:

```
{
   "code": -32602,
   "message": "missing a required argument: arg"
}
```

From https://github.com/ElementsProject/lightning/blob/b51772ff68c237627f526be579f0b99c90082183/contrib/pyln-client/pyln/client/plugin.py#L83-L89

The `error` field is set as a string, with the `exc` set using `repr`. Setting the error as a `dict` and using `str` instead of `repr` would generate an output closer to the built-ins:

```
self.exc = exc
self._write_result({
    'jsonrpc': '2.0',
    'id': self.id,
    "error": {"error": "Error while processing {method}: {exc}".format(
        method=self.method, exc=str(exc)
    ),
}})
```

Output:

```
{
   "error": "Error while processing plugin-method: missing a required argument: 'arg'"
}
```
### `getinfo` output

```
{
   "id": "02a59ee414c69266f9779925af18299fb05c0915a2a5ef22c449ac4f9a96498245",
   "alias": "HOPPINGNET",
   "color": "02a59e",
   "num_peers": 1,
   "num_pending_channels": 0,
   "num_active_channels": 1,
   "num_inactive_channels": 1,
   "address": [],
   "binding": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 6060
      }
   ],
   "version": "v0.8.0rc2-447-gc99a0e8-modded",
   "blockheight": 557,
   "network": "regtest",
   "msatoshi_fees_collected": 0,
   "fees_collected_msat": "0msat",
   "lightning-dir": "/tmp/l1-regtest/regtest"
}
```

darosior:
fixes #3638

Added @sr-gi as a co author as he did all the search and found the cause.

### DIFF
--- a/contrib/plugins/helloworld.py
+++ b/contrib/plugins/helloworld.py
@@ -19,6 +19,12 @@ def hello(plugin, name="world"):
     return s
 
 
+@plugin.method("bye")
+def bye(plugin, name, **kwargs):
+    """This methods requires {name} to be set by the caller !"""
+    return "Bye {}".format(name)
+
+
 @plugin.init()
 def init(options, configuration, plugin, **kwargs):
     plugin.log("Plugin helloworld.py initialized")
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -83,9 +83,12 @@ class Request(dict):
         self._write_result({
             'jsonrpc': '2.0',
             'id': self.id,
-            "error": "Error while processing {method}: {exc}".format(
-                method=self.method, exc=repr(exc)
-            ),
+            "error": {
+                "code": -32600,  # "Invalid Request"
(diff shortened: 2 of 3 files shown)


sr-gi:
LGTM.

 I'd only argue that `str` may be nicer than `repr` for the cmd return, since you may want to avoid debugging info on the cmd. 

As discussed in #3638, the full traceback is left in the logs anyway, so it may just create nicer returns.

darosior:
Agree, will amend
-------- Original Message --------

PR #3645, opened 2020-04-14
watchtower: Alternative implementation with `channeld` in control

### LINKED ISSUE #1353: Feature Request: RPC endpoint for Watch Tower (Simple)

### LINKED ISSUE #3422: Add plugin hook for penalty transactions on revocation

cdecker:
This is an alternative implementation of the watchtower hook presented in
#3601, with a couple more things shifted towards `channeld`. The API interface
remains unchanged, but `lightningd` now manages fewer moving pieces, acting
solely as a passthrough between `channeld` and `lightningd`, and it creates
the penalty tx.

flow

Closes #3601
Closes #3659

Fixes #3422
Fixes #1353

### DIFF
--- a/channeld/Makefile
+++ b/channeld/Makefile
@@ -69,6 +69,7 @@ CHANNELD_COMMON_OBJS :=				\
 	common/onionreply.o			\
 	common/peer_billboard.o			\
 	common/peer_failed.o			\
+	common/penalty_base.o			\
 	common/per_peer_state.o			\
 	common/permute_tx.o			\
 	common/ping.o				\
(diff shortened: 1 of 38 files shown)


rustyrussell:
Yes, I kept thinking that we didn't need mutliple pbases per peer, but of course we do.  This looks good so far (though obv `hsmd: Allow `lightningd` to specify the channel and node ID` won't be needed when channeld is doing the penalty creation).

cdecker:
> Yes, I kept thinking that we didn't need mutliple pbases per peer, but of course we do. This looks good so far (though obv `hsmd: Allow `lightningd` to specify the channel and node ID` won't be needed when channeld is doing the penalty creation).

Exactly, but the penalty creation needs some rewiring. Should I go ahead and do that?

cdecker:
Found a lingering memleak.

PR #3655, opened 2020-04-20
changelog: use correct format for label

niftynei:
The reproducible build won't work with `-`'s in the version label.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3655/commits/4cd38a5f5dd59bc9f916810ea2fa52f75a8698ee

PR #3659, opened 2020-04-21
Watchtowers attempt 3: the full channelding

### LINKED ISSUE #1353: Feature Request: RPC endpoint for Watch Tower (Simple)

### LINKED ISSUE #3422: Add plugin hook for penalty transactions on revocation

cdecker:
This is the third variant of the watchtower hook, the channelding. The main
difference with respect to #3645 is that the penalty transaction is now
created by `channeld` which already has a contextualized connection to `hsmd`,
so we can now drop the changes to the `hsmd` wire protocol, and just add the
penalty creation capability to the connection between `hsmd` and
`channeld`. It also removes the need to synchronously talk to the `hsmd` from
`lightningd` which was a bit ugly.

flow

Closes #3601
Closes #3645

Fixes #3422
Fixes #1353

### DIFF
--- a/channeld/watchtower.c
+++ b/channeld/watchtower.c
@@ -1,0 +1,120 @@
+#include "watchtower.h"
+
+#include <bitcoin/feerate.h>
+#include <bitcoin/script.h>
+#include <bitcoin/signature.h>
+#include <bitcoin/tx.h>
+#include <common/htlc_tx.h>
+#include <common/key_derive.h>
+#include <common/keyset.h>
+#include <common/status.h>
+#include <common/type_to_string.h>
+#include <hsmd/gen_hsm_wire.h>
+#include <wire/wire_sync.h>
+
+static const u8 ONE = 0x1;
+#define HSM_FD 6
+
+const struct bitcoin_tx *
+penalty_tx_create(const tal_t *ctx,
+		  const struct channel *channel,
+		  u32 penalty_feerate,
+		  u8 *final_scriptpubkey,
+		  const struct secret *revocation_preimage,
+		  const struct bitcoin_txid *commitment_txid,
+		  s16 to_them_outnum, struct amount_sat to_them_sats)
+{
+	u8 *wscript;
+	struct bitcoin_tx *tx;
+	struct keyset keyset;
+	size_t weight;
+	u8 *msg;
+	struct amount_sat fee, min_out, amt;
+	struct bitcoin_signature sig;
+	u32 locktime = 0;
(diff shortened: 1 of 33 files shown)


channeld/watchtower.c
```diff
@@ -0,0 +1,120 @@
...
+	    towire_hsm_sign_penalty_to_us(ctx, &remote_per_commitment_secret, tx,
+					  wscript, *tx->input_amounts[0]);
+
+	if (!wire_sync_write(HSM_FD, take(hsm_sign_msg)))
+		status_broken("Writing sign request to hsm");
+
+	msg = wire_sync_read(tmpctx, HSM_FD);
+	if (!msg || !fromwire_hsm_sign_tx_reply(msg, &sig)) {
+		status_broken("Reading sign_tx_reply: %s", tal_hex(tmpctx, msg));
+		abort();
```
rustyrussell, line 112: status_failed(STATUS_FAIL_HSM_IO, ....

lightningd/peer_htlcs.c
```diff
@@ -1986,14 +1986,22 @@ static void commitment_revocation_hook_serialize(
...
+	return true;
+}
+
+
+REGISTER_PLUGIN_HOOK(commitment_revocation,
+		     commitment_revocation_hook_deserialize,
 		     commitment_revocation_hook_cb,
-		     struct commitment_revocation_payload *,
 		     commitment_revocation_hook_serialize,
 		     struct commitment_revocation_payload *);
```
rustyrussell: This looks like a fixup for the previous commit, not part of ` pytest: Add a test for the commitment_revocation hook ` ?

lightningd/peer_htlcs.c
```diff
@@ -2121,7 +2129,7 @@ void peer_got_revoke(struct channel *channel, const u8 *msg)
 	payload->wallet = ld->wallet;
 	payload->channel_id = channel->dbid;
 	payload->commitnum = pbase->commitment_num;
-	plugin_hook_call_commitment_revocation(ld, payload, payload);
+	plugin_hook_call_commitment_revocation(ld, payload);
 }
```
rustyrussell: This too...

rustyrussell:
OK, I rebased to avoid the (trivial) conflicts and reordered commits so it compiles and passes tests at all points.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/af88a72e3a517c321e8b289935874488fcd26270

PR #3663, opened 2020-04-24
feerates: de-dupe min_ max_ printing twice when `feerates` called

niftynei:
We want to compare to `i`, the enum counter, not feerates[i], the
feerate value.

Without this fix, `feerates` appears as:
```
{
   "perkw": {
      "opening": 253,
      "mutual_close": 253,
      "unilateral_close": 253,
      "delayed_to_us": 253,
      "htlc_resolution": 253,
      "penalty": 253,
      "min_acceptable": 253,
      "max_acceptable": 2500,
      "min_acceptable": 253,
      "max_acceptable": 4294967295,
      "urgent": 253,
      "normal": 253,
      "slow": 506
   },
   "onchain_fee_estimates": {
      "opening_channel_satoshis": 177,
      "mutual_close_satoshis": 170,
      "unilateral_close_satoshis": 151,
      "htlc_timeout_satoshis": 167,
      "htlc_success_satoshis": 177
   }
}
```

bug introduced in "chaintopology: better feerate targets differentiation"

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -516,8 +516,7 @@ static struct command_result *json_feerates(struct command *cmd,
 	response = json_stream_success(cmd);
 	json_object_start(response, json_feerate_style_name(*style));
 	for (size_t i = 0; i < ARRAY_SIZE(feerates); i++) {
-		if (!feerates[i] || feerates[i] == FEERATE_MIN
-		    || feerates[i] == FEERATE_MAX)
+		if (!feerates[i] || i == FEERATE_MIN || i == FEERATE_MAX)
 			continue;
 		json_add_num(response, feerate_name(i),
 			     feerate_to_style(feerates[i], *style));


niftynei:
Also lol at `slow` appearing as 2x the `urgent` and `normal` rates. This is from testnet though so no wonder things are a bit off.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3663/commits/54fda66d07008758898742619ec24e7ff38c03ba

Depending on the JSON parser this can actually cause some trouble :-)

darosior:
Ok I bugged a bit at the (deprecated) `slow` field being 2 * the minRelayFee, but after double checking actually it's because it displays 2 * feerate_min (slow doesn't exist anymore, and feerate_min is ECONOMICAL/100 divided by 2, floored at minRelayFee (253)).

niftynei:
Yeah it might be reasonable to set normal as the ceiling of slow but it's definitely an edge case.

PR #3665, opened 2020-04-25
doc: bitcoin backend must return sat/kVB not btc/kVB on `estimatefees`

darosior:
Thanks to @SimonVrouwe !


cdecker:
No need to roll this into the release, since it'll deployed independently.

PR #3671, opened 2020-04-26
pyln-proto: Use only coincurve for libsecp256k1 bindings

mrostecki:
secp256k1 Python library is not maintained anymore and coincurve was
already used in the `wire` module.

Changelog-None
Signed-off-by: Michal Rostecki <mrostecki@mailfence.com>

### DIFF
--- a/contrib/pyln-proto/pyln/proto/invoice.py
+++ b/contrib/pyln-proto/pyln/proto/invoice.py
@@ -6,7 +6,7 @@ import base58
 import bitstring
 import hashlib
 import re
-import secp256k1
+import coincurve
 import time
 import struct
 
@@ -180,14 +180,14 @@ class Invoice(object):
 
     def __str__(self):
         return "Invoice[{}, amount={}{} tags=[{}]]".format(
-            hexlify(self.pubkey.serialize()).decode('utf-8'),
+            hexlify(self.pubkey.format()).decode('utf-8'),
             self.amount, self.currency,
             ", ".join([k + '=' + str(v) for k, v in self.tags])
         )
 
     @property
     def hexpubkey(self):
-        return hexlify(self.pubkey.serialize()).decode('ASCII')
+        return hexlify(self.pubkey.format()).decode('ASCII')
 
     @property
     def hexpaymenthash(self):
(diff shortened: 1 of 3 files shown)


cdecker:
Thanks @mrostecki, incidentally I was looking for a maintained version of the sec256k1 bindings (and contacted the maintainer of the PyPI secp256k1 package, but didn't get a reply). coincurve seems like it's pulling in more dependencies than necessary, which I find rather scary, given it'll mostly be used by cryptocurrency related projects and therefore an interesting target for code swaps.

Would it make sense to just create our own package directly out of the secp256k1 tree instead?

mrostecki:
@cdecker As far as I see, coincurve pulls only one dependency more than secp256k1.

```
❯ pipdeptree -p coincurve
coincurve==13.0.0
  - asn1crypto [required: Any, installed: 1.3.0]
  - cffi [required: >=1.3.0, installed: 1.14.0]
    - pycparser [required: Any, installed: 2.20]
```

Is there something wrong with asn1crypto which I'm not aware of? If that's case, then sure, creating our own package based on secp256k1 and using it in `wire.py` sounds like a good idea.

An another problem with secp256k1-py is that by default it pulls an old tarball of secp256k1, which apart from being outdated, doesn't build with OpenSSL 1.1. Something to fix if we decide to fork - either by bumping the tarball version to bundle or shared linking by default.

cdecker:
You might be right, I was misremembering this apparently, but I felt a bit uneasy just pulling that lib in. 
Might want to do a quick review of the lib in my spare time.

darosior:
If I might weigh in I gave it a quick look some time ago and it's a pure wrapper to libsecp256k1 and it seemed up to date. It's used by Python Bitcoin applications (such as [joinmarket](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/jmbitcoin/setup.py)).

ofek:
FYI I plan to drop asn1crypto (only used for der/pem serialization)

cdecker:
@mrostecki can I mark this as "ready for review" and merge it or were you planning on adding to his PR?

mrostecki:
@cdecker I was planning to add the test for `encode` method, to ensure/prove that this change doesn't break anything. I will take care of that soon.

cdecker:
No worries, just wanted to check in whether the draft was on purpose ^^

mrostecki:
@cdecker Ready for review :)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3671/commits/32f571f94d9c67490d6fce10ab8bf13d15cd2972

PR #3673, opened 2020-04-27
libplugin: It's featurebits, not features

cdecker:
The documentation was wrong, and I copied my mistake to `libplugin` where it
was then ignored instead of ORed into the node's featurebits. This fixes both.


niftynei:
closed by #3674

Issue #3571, opened 2020-03-05
Improvement: fail fast if the Bitcoin backend fails


darosior:
#3565 introduces a timeout to not wait the Bitcoin backend's response to the `init` message for ever.

As an improvement, it would be nice to fail directly if the plugin dies.

It could be a good first issue ?
In this case some hints:
The Bitcoin backend plugin will signal its readiness by completing [the handshake](https://lightning.readthedocs.io/PLUGINS.html#a-day-in-the-life-of-a-plugin) with `lightningd`. This handshake is completed by responding to [the `init` message](https://lightning.readthedocs.io/PLUGINS.html#the-init-method).

If the plugin dies for whatever reason (for example `bcli` cannot connect to `bitcoind`), the connection will be dropped and `plugin_conn_finished` will be called:
https://github.com/ElementsProject/lightning/blob/ac5cba22434c8133934b86f1dcf4f79956341b32/lightningd/plugin.c#L435-L444

It frees the plugin struct so maybe an (untested) idea would be to attach a custom destructor to Bitcoin plugins ?

ZmnSCPxj:
Why not fold this into #3213? Make `bcli` a "crucial" plugin which, if it is not loaded and initialized properly, we just die.

Then to replace the plugin we can make it non-crucial and suppress loading it and then make a new crucial plugin.

whitslack:
A couple of weeks ago, I discovered that my Lightning node had not been interacting with the blockchain for over a day because `bcli` had failed. The only way I noticed was that I had a channel that had been in `CHANNELD_AWAITING_LOCKIN` state for hours even though the peer was connected and the funding transaction was many blocks deep in the chain. When I examined the process tree, I noticed that `bcli` was in the defunct/zombie state. If this condition had continued for longer than my `their_to_self_delay` in any of my channels, my peers could have successfully cheated me. Obviously no good! So I appreciate the move toward making `bcli` a "crucial" plugin, without which `lightningd` will terminate.

darosior:
Some logs would be very welcome as we would die on the next request (every 30s at most) if the connection to `bcli` is dropped. So `bcli` was in zombie state but still kept its connection opened ?

whitslack:
[...]

PR #3675, opened 2020-04-28
Bitcoin: fail immediately if we lose connection to our backend

darosior:
I think this fixes \ #3571, but I'd like to double check with @whitslack before auto closing the issue.

This comports two drive-by fixes in addition which I remarked while debugging https://github.com/ElementsProject/lightning/issues/3571.

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -40,20 +40,20 @@ static const char *methods[] = {"getchaininfo", "getrawblockbyheight",
                                 "sendrawtransaction", "getutxout",
                                 "estimatefees"};
 
+static void bitcoin_destructor(struct plugin *p)
+{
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3675/commits/2c1f300f7df8b4a147313a8b62317ff57b123722

darosior:
@cdecker I'm not sure about the "Next release" milestone ? Since the behaviour experienced in #3571 is concerning : without new blocks during 14 hours, he could have been cheated.
However we can assume `bcli` running OOM can be rare, but still..

rustyrussell:
In general, I think we should die if a static plugin dies.

WRT release, it's @niftynei's call here, as Release Captain for 0.8.2.  IMHO, if we were doing another rc anyway, I'd put this in, but wouldn't do an rc4 for this alone.

PR #3679, opened 2020-04-29
Dynamic plugin cleanups

rustyrussell:
This patch series is about unifying the dynamic and startup plugin paths.  In doing so it clears up some latent bugs, and reduces code duplication.


lightningd/jsonrpc.c
```diff
@@ -850,11 +846,15 @@ parse_request(struct json_connection *jcon, const jsmntok_t tok[])
 	/* Duplicate since we might outlive the connection */
 	rpc_hook->buffer = tal_dup_talarr(rpc_hook, char, jcon->buffer);
 	rpc_hook->request = tal_dup_talarr(rpc_hook, jsmntok_t, tok);
-	/* Prevent a race between was_pending and still_pending */
-	new_reltimer(c->ld->timers, rpc_hook, time_from_msec(1),
-	             call_rpc_command_hook, rpc_hook);
```
darosior, line 855: :tada:

lightningd/plugin.c
```diff
@@ -190,6 +190,7 @@ void plugin_kill(struct plugin *plugin, char *fmt, ...)
 {
```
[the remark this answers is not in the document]
darosior: Ok I came across this comment in another commit, so I'll precise: this a comment for the commit message of 62f0916bcee9ec07bc4df0a88009be4fd839775a

lightningd/plugin.c
```diff
@@ -456,7 +495,12 @@ static struct io_plan *plugin_write_json(struct io_conn *conn,
 static void plugin_conn_finish(struct io_conn *conn, struct plugin *plugin)
 {
 	plugin->stdout_conn = NULL;
+	if (plugin->start_cmd) {
+		plugin_cmd_succeeded(plugin->start_cmd, plugin);
```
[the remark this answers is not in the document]
darosior: outdated: modified in a following commit

darosior:
Travis TL;DR :  
```patch
index ddfa141d7..2013501be 100644
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -123,8 +123,8 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
         * is a nod to keeping it minimal and explicit: we need this code for
         * testing, but its existence means we're not actually testing the
         * same exact code users will be running. */
-       ld->dev_debug_subprocess = NULL;
 #if DEVELOPER
+       ld->dev_debug_subprocess = NULL;
        ld->dev_disconnect_fd = -1;
        ld->dev_subdaemon_fail = false;
        ld->dev_allow_localhost = false;
```

lightningd/plugin.c
```diff
@@ -1240,7 +1384,7 @@ void plugins_config(struct plugins *plugins)
 {
 	struct plugin *p;
 	list_for_each(&plugins->plugins, p, list) {
-		if (p->plugin_state == UNCONFIGURED)
+		if (p->plugin_state == NEEDS_INIT)
 			plugin_config(p);
```
darosior, line 1388: I can't add a comment directly in the function, but you don't set the `AWAITING_INIT_RESPONSE` in `plugin_config()`, after sending the init request.

darosior:
nit : [You never set the `AWAITING_INIT_RESPONSE` state after sending the init request](https://github.com/ElementsProject/lightning/pull/3679#discussion_r417852591).

rustyrussell:
Yep, fixed...

tests/plugins/fail_by_itself.py
```diff
@@ -0,0 +1,31 @@
...
+
+class FailThread(threading.Thread):
+    def __init__(self):
+        super().__init__()
+        self.start()
+
+    def run(self):
+        time.sleep(1)
+        print("Exiting!")
+        os._exit(1)
```
cdecker, line 18: Interesting, I usually use `sys.exit(1)` for these, but the docs explicitly suggest using `os._exit(1)` after a `fork` call.

tests/plugins/fail_by_itself.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@plugin.init()
+def init(options, configuration, plugin):
+    FailThread()
```
cdecker, line 23: The same could be achieved by the following:

```python
    threading.Timer(1, lambda: os._exit(1))
```

or if the `print` is actually needed we can do this instead:

```python 
    def die():
        print("Exiting!")
        os._exit(1)
    threading.Timer(1.0, die)
```

:wink:

tests/test_plugin.py
```diff
@@ -1200,3 +1200,18 @@ def test_replacement_payload(node_factory):
         l1.rpc.pay(inv)
 
     assert l2.daemon.wait_for_log("Attept to pay.*with wrong secret")
+
+
+def test_plugin_fail(node_factory):
+    """Test that a plugin which fails (not during a command)"""
+    plugin = os.path.join(os.path.dirname(__file__), 'plugins/fail_by_itself.py')
+    l1 = node_factory.get_node(options={"plugin": plugin})
+
```
cdecker, line 1259: We might want to check that `failcmd` is in help before the plugin spontaneously catches fire :-)

lightningd/plugin.h
```diff
@@ -196,6 +194,18 @@ bool plugin_paths_match(const char *cmd, const char *name);
...
+ * Kick of initialization of a plugin.
+ */
+bool plugin_send_getmanifest(struct plugin *p);
+
+/**
+ * Kick of initialization of all plugins which need it/
+ *
+ * Return true if any were started.
+ */
+bool plugins_send_getmanifest(struct plugins *plugins);
```
cdecker: Why does this need to be public?

lightningd/plugin.c
```diff
@@ -201,8 +202,18 @@ void plugin_kill(struct plugin *plugin, char *fmt, ...)
...
+	while ((opt = list_top(&plugin->plugin_opts, struct plugin_opt, list))) {
+		if (!opt_unregister(opt->name))
+			fatal("Could not unregister %s from plugin %s",
+			      opt->name, plugin->cmd);
+		list_del_from(&plugin->plugin_opts, &opt->list);
+	}
+
+	if (plugin->start_cmd) {
 		plugin_cmd_killed(plugin->start_cmd, plugin, msg);
+		plugin->start_cmd = NULL;
```
cdecker, line 212: Aren't we about to free anyway, which cleans up this dangling pointer?

lightningd/plugin_control.c
```diff
@@ -27,32 +27,13 @@ static struct command_result *plugin_dynamic_list_plugins(struct command *cmd,
 		json_add_string(response, "name", p->cmd);
 		json_add_bool(response, "active",
 		              p->plugin_state == INIT_COMPLETE);
+		json_add_u32(response, "state", p->plugin_state);
```
cdecker, line 30: This is the numeric state which won't be very helpful for users looking at it (especially should we ever add intermediate states), can we have a mapping from enum to a human readable string here?

lightningd/plugin_control.c
```diff
@@ -71,143 +52,27 @@ struct command_result *plugin_cmd_all_complete(struct plugins *plugins,
...
-	io_new_conn(p, stdout, plugin_stdout_conn_init, p);
-	io_new_conn(p, stdin, plugin_stdin_conn_init, p);
-	req = jsonrpc_request_start(p, "getmanifest", p->log,
-	                            plugin_dynamic_manifest_callback, dp);
-	jsonrpc_request_end(req);
-	plugin_request_send(p, req);
-	p->plugin_state = AWAITING_GETMANIFEST_RESPONSE;
-	return command_still_pending(dp->cmd);
-}
-
```
cdecker: Very nice cleanup :+1:

lightningd/plugin_control.c
```diff
@@ -71,143 +52,27 @@ struct command_result *plugin_cmd_all_complete(struct plugins *plugins,
...
 /**
  * Called when trying to start a plugin through RPC, it starts the plugin and
  * will give a result 20 seconds later at the most.
  */
 static struct command_result *
 plugin_dynamic_start(struct command *cmd, const char *plugin_path)
 {
-	struct dynamic_plugin *dp;
```
cdecker, line 180: The comment still references 20 seconds.

cdecker: Couldn't add the review comment on that line since it'd confuse GH to have review comments on unchanged lines.

lightningd/plugin.c
```diff
@@ -339,6 +339,7 @@ static const char *plugin_response_handle(struct plugin *plugin,
 	 * plugin is parent), so detect that case */
 	if (!was_plugin_destroyed(pd))
 		tal_free(request);
+	return NULL;
```
cdecker, line 341: This likely is part of the previous commit.

lightningd/plugin.c
```diff
@@ -482,7 +483,8 @@ static struct io_plan *plugin_read_json(struct io_conn *conn,
 
 		if (err) {
 			plugin_kill(plugin, "%s", err);
-			return io_close(plugin->stdout_conn);
+			/* plugin_kill frees plugin */
+			return io_close(NULL);
```
cdecker, line 485: This is interesting, what's the effect of `io_close(NULL)`? Is it just intended as a no-op?

lightningd/plugin.c
```diff
@@ -1277,9 +1277,12 @@ bool plugins_send_getmanifest(struct plugins *plugins)
 
 	/* Spawn the plugin processes before entering the io_loop */
 	list_for_each_safe(&plugins->plugins, p, next, list) {
+		const char *err;
+
 		if (p->plugin_state != UNCONFIGURED)
 			continue;
-		if (plugin_send_getmanifest(p)) {
+		err = plugin_send_getmanifest(p);
+		if (!err) {
```
cdecker, line 1282: Could this leak `err`?

cdecker: Almost: it's allocated off of the plugin which we'd be freeing if there is an error.

lightningd/jsonrpc.c
```diff
@@ -850,11 +846,15 @@ parse_request(struct json_connection *jcon, const jsmntok_t tok[])
...
 
-	return command_still_pending(c);
+	db_begin_transaction(jcon->ld->wallet->db);
+	completed = plugin_hook_call_rpc_command(jcon->ld, rpc_hook);
+	db_commit_transaction(jcon->ld->wallet->db);
+
+	/* If it's deferred, mark it (otherwise, it's completed) */
+	if (!completed)
+		return command_still_pending(c);
+	return NULL;
```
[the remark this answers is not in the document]
rustyrussell: No, it's waiting for the plugin to respond...

lightningd/plugin.c
rustyrussell: Yes, just like the one above it, it's a cheats way of getting an io object to return

lightningd/plugin.c
rustyrussell: At various points through this series, this is tested in different paths. I think eventually this may be unnecessary, but I'll check

lightningd/plugin.c
rustyrussell: Yeah, I use this pattern  through this code quite a bit.

lightningd/plugin_control.c
rustyrussell: Oops, this was a debugging thing I forgot to remove!!

lightningd/plugin.c
rustyrussell: Ack, fixed!

tests/test_plugin.py
rustyrussell: Timing issue though, so I didn't bother...

lightningd/plugin.h
rustyrussell: Called for dynamic plugins too (startdir).

rustyrussell:
Self-ack for @cdecker feedback minor changes (and rebase)

Ack https://github.com/ElementsProject/lightning/pull/3679/commits/b2ec70af0263361e13b1e554c0edb298c33ab049

PR #3680, opened 2020-04-29
Small set of dual-funding related changes

niftynei:
@rustyrussell had the idea to start pulling out parts of the dual-funding PR into more manageable sets; this is the set of the easiest commits to pull out. more coming soon


wallet/db.c
```diff
@@ -596,6 +598,7 @@ static struct migration dbmigrations[] = {
      * Turn anything in transition into a WIRE_TEMPORARY_NODE_FAILURE. */
     {SQL("ALTER TABLE channel_htlcs ADD localfailmsg BLOB;"), NULL},
     {SQL("UPDATE channel_htlcs SET localfailmsg=decode('2002', 'hex') WHERE malformed_onion != 0 AND direction = 1;"), NULL},
+    {SQL("ALTER TABLE channels ADD our_funding_satoshi INTEGER DEFAULT 0;"), migrate_our_funding},
```
rustyrussell, line 601: BIGINT I think?

wallet/db.c
```diff
@@ -1072,6 +1075,51 @@ static void migrate_pr2342_feerate_per_channel(struct lightningd *ld, struct db
...
 }
 
+/* We've added a column `our_funding_satoshis`, since channels can now
+ * have funding for either channel participant. We need to 'backfill' this
+ * data, however. We can do this using the fact that our_funding_satoshi
+ * is the same as the funding_satoshi for every channel where we are
+ * the `funder`
+ */
+static void migrate_our_funding(struct lightningd *ld, struct db *db)
+{
```
rustyrussell, line 1085: Pretty sure this is easy to do in SQL?  Since you set the default to 0, we just need to fix up the others?

"UPDATE channels SET our_funding_satoshi = funding_satoshi WHERE funder = 0;" /* 0 == LOCAL */

common/utxo.c
```diff
@@ -29,16 +34,22 @@ void towire_utxo(u8 **pptr, const struct utxo *utxo)
...
 	utxo->amount = fromwire_amount_sat(ptr, max);
 	utxo->keyindex = fromwire_u32(ptr, max);
 	utxo->is_p2sh = fromwire_bool(ptr, max);
 
-	/* No need to tell hsmd about the scriptPubkey, it has all the info to
-	 * derive it from the rest. */
-	utxo->scriptPubkey = NULL;
+	script_len = fromwire_u16(ptr, max);
+	utxo->scriptPubkey = tal_arr(utxo, u8, script_len);
+	fromwire_u8_array(ptr, max, utxo->scriptPubkey, script_len);
```
rustyrussell, line 48: We now have `utxo->scriptPubkey = fromwire_tal_arrn(ptr, max, fromwire_u16(ptr, max));`

common/utxo.c
```diff
@@ -29,16 +34,22 @@ void towire_utxo(u8 **pptr, const struct utxo *utxo)
...
+
+	scriptsig_len = fromwire_u16(ptr, max);
+	utxo->scriptSig = tal_arr(utxo, u8, scriptsig_len);
+	fromwire_u8_array(ptr, max, utxo->scriptSig, scriptsig_len);
```
rustyrussell, line 52: Here too

wallet/db.c
cdecker: For `sqlite3` it doesn't matter, but testing `postgres` it'd be triggering an `assert` about field lenght.

bitcoin/tx.c
```diff
@@ -228,6 +228,12 @@ const u8 *bitcoin_tx_output_get_script(const tal_t *ctx,
 	return res;
 }
 
+void bitcoin_tx_output_get_amount_sat(struct bitcoin_tx *tx, int outnum,
+				      struct amount_sat *amount)
+{
+	amount->satoshis = tx->wtx->outputs[outnum].satoshi;  /* Raw: type conversion */
+}
+
```
cdecker, line 236: This should likely use `bitcoin_tx_output_get_amount`, `assert` that we're dealing with the fee-asset, and then return the converted value in `amount_sat` instead. As it is this'd fail attempting to access `.satoshi` on elements since that field is not populated.

common/utxo.h
```diff
@@ -41,6 +41,9 @@ struct utxo {
 
 	/* The scriptPubkey if it is known */
 	u8 *scriptPubkey;
+
+	/* scriptSig. Only for P2SH outputs */
+	u8 *scriptSig;
```
cdecker, line 46: And if we know it at all :-)

wallet/db.c
niftynei: Nice. Wishing I had written a test for this though >.< lol

niftynei:
Ok I think all comments have been addressed? tested the upgrade works as expected.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/7ae6046f6b880fbff33f52c2bbf842262b6ff0cb

PR #3685, opened 2020-05-01
Offers preparation

rustyrussell:
1. Some tooling improvements.
2. Add missing "channel_announce" features to listchannels.


gossipd/gossip_store.c
```diff
@@ -116,14 +116,44 @@ static bool append_msg(int fd, const u8 *msg, u32 timestamp,
...
+{
+	assert(can_upgrade(oldversion));
+
+	/* We only need to upgrade this */
+	if (fromwire_peektype(*msg) == WIRE_GOSSIPD_LOCAL_ADD_CHANNEL) {
+		/* Append two 0 bytes, for (empty) feature bits */
+		tal_resizez(msg, tal_bytelen(*msg) + 2);
+	}
+	return true;
+}
```
cdecker, line 136: Patching these on the fly seems a bit strange to me:  couldn't we have optional fields at the end, maybe even make the switch to TLV encode the `local_add_channel` instead? That'd allow us to add/remove optional fields without bumping the `gossip_store` version later. It'd be better future proofing imho.

rustyrussell: Well, these messages are unusual: usually we don't need to worry about compatibility at all. I only did this because it was easy: I don't want all the callers to have to worry about missing fields.

rustyrussell: Oh, BTW, these are patched during our "offline compaction" run which always runs at startup and rewrites the files.

So they're not converted every time!

PR #3688, opened 2020-05-01
Weekly cleanups

cdecker:
Just two tiny fixes:

 - Using `liquid-regtest` we were having some issues checking amounts in the penalty tests. Now instead we make sure that we have the outputs that we expect instead of just checking the sum of their values. This is more fine-grained and more likely to detect a missing output being swept into our wallet.
 - The testing framework was breaking when running against some branches because the fee estimation was asking for configurations that do not match what we expect. Now we return a constant (42, what else) and queue a warning which gets collected by `pytest` and highlighted in the final summary. This allows us to still be warned, but not break the tests right away.

Cheers 😊

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -21,6 +21,7 @@ import struct
 import subprocess
 import threading
 import time
+import warnings
 
 BITCOIND_CONFIG = {
     "regtest": 1,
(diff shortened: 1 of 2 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3688/commits/211618610c7b3b8a03ecccee1d49c73a8f1a0725

Issue #3684, opened 2020-05-01
`pay` tries to pay an invoice with a wrong amount


kirelagin:
Let me say right away that this is happening with v0.8.1. I haven’t tried with v0.8.2 yet, but I went through the git log, and couldn’t find anything that looked like it could have fixed this.

The problem is that when I try to pay an invoice, even though it is being parsed correctly, the node tries to send a transaction with an incorrect amount. Here is an example with a 20 msat invoice:

```text
$ lightning-cli pay 'lntb200p1p02kekypp5gvs5cp5wwfcjv93xzqaa6p7evljggw2ajdyt6rl982xdnkadj9csdp0d35kw6r5de5kueedvdhkuarjdakzqarjv9h8xctrw35k7msxqyjw5qcqp2sp5ncqsf0ej3qaxn3pmtw29ensrswr5mf0csfnk6ac7fac2mdjk2zjqrzjqgcjvfla7plmm4l9mkcnvcgmmh5mqrfxsgw3fk2gjyu4g5hk0teysxwhgqqqqfsqqqqqqqlgqqqqqqgq9q9qy9qsqxgey4samkm0fq03xxjmj55y3pql3evxq2xzaxtdm5v86d2h5ws0y3lgcmh6upywazvrw3ggxv6yhyds84v90nx2al6r6wfk7f02r6lcqqjyst6'
{
   "code": 203,
   "message": "failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)",
   "data": {
      "id": 48,
      "payment_hash": "43214c068e7271261626103bdd07d967e484395d9348bd0fe53a8cd9dbad9171",
      "destination": "021c38ebb30aa17901aec609e98da063b4e433bfbe45656bddb34ea688192bd734",
      "msatoshi": 1020,
      "amount_msat": "1020msat",
      "msatoshi_sent": 1020,
      "amount_sent_msat": "1020msat",
      "created_at": 1588291657,
      "status": "pending",
[...]

PR #3693, opened 2020-05-02
pay: respect maxfeepercent when choosing a shadow route

darosior:
~~We claim to do so in the manpage.~~ [I was wrong](https://github.com/ElementsProject/lightning/pull/3693#issuecomment-623145123)

fixes https://github.com/ElementsProject/lightning/issues/3684

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -85,6 +85,8 @@ struct pay_command {
 
 	/* How much we're paying, and what riskfactor for routing. */
 	struct amount_msat msat;
+	/* Blank amount to pay, without fees and shadow route(s). */
+	struct amount_msat initial_msat;
 	/* riskfactor 12.345% -> riskfactor_millionths = 12345000 */
 	u64 riskfactor_millionths;
 	unsigned int final_cltv;
@@ -1046,6 +1048,12 @@ static struct command_result *add_shadow_route(struct command *cmd,
 	size_t i;
 	u64 sample = 0;
 	struct route_info *route = tal_arr(NULL, struct route_info, 1);
+	struct amount_msat fees, maxfees;
+	/* Don't go above this. Note how we use the initial amount to get the percentage
+	 * of the fees, or it would increase with the addition of new shadow routes. */
+	if (!amount_msat_fee(&maxfees, pc->initial_msat, 0, pc->maxfee_pct_millionths))
+		plugin_err(cmd->plugin, "Overflow when computing maxfees for "
+					"shadow routes.");
 
 	json_for_each_arr(i, chan, channels) {
 		u64 v = pseudorand(UINT64_MAX);
@@ -1067,6 +1075,11 @@ static struct command_result *add_shadow_route(struct command *cmd,
 			               &route[0].fee_base_msat);
 			json_to_number(buf, json_get_member(buf, chan, "fee_per_millionth"),
 			               &route[0].fee_proportional_millionths);
+
+			if (!amount_msat_fee(&fees, pc->initial_msat, route[0].fee_base_msat,
+					    route[0].fee_proportional_millionths)
+			    || amount_msat_greater_eq(fees, maxfees))
+				continue;
 
 			best = chan;
 			sample = v;
@@ -1313,13 +1326,13 @@ static struct command_result *json_pay(struct command *cmd,
 			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 					    "msatoshi parameter unnecessary");
 		}
-		pc->msat = *b11->msat;
+		pc->msat = pc->initial_msat = *b11->msat;
 	} else {
 		if (!msat) {
 			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 					    "msatoshi parameter required");
 		}
-		pc->msat = *msat;
+		pc->msat = pc->initial_msat = *msat;
 	}
 
 	/* Sanity check */
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -452,6 +452,35 @@ def test_payment_duplicate_uncommitted(node_factory, executor):
     fut2.result(10)
 
 
+@pytest.mark.xfail(strict=True)
+@unittest.skipIf(not DEVELOPER, "Too slow without --dev-fast-gossip")
+def test_pay_maxfee_shadow(node_factory):
+    """Test that we respect maxfeepercent for shadow routing."""
+    l1, l2, l3 = node_factory.line_graph(3, fundchannel=True,
+                                         wait_for_announce=True)
+    # We use this to search for shadow routes
+    wait_for(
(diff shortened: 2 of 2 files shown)


kirelagin:
I gave the docs a more thorough read, and, it turns out, they are actually correct:

> **Route randomization** will never exceed `maxfeepercent` of the payment. **Route randomization and shadow routing** will not take routes that would exceed `maxdelay`.

So, it only claims that route randomisation respects `maxfeepercent`, not shadow routing.

Anyway, docs aside, there are two things here:

1. No way of disabling shadow routing. As I understand now, [this was intentional](https://github.com/ElementsProject/lightning/pull/3212#discussion_r342386943), however I think it would still be good to have an option for disabling it (e.g. my use-case is that I control multiple nodes connected to each other and I would like to be able to distribute liquidity among them through direct channels, and it is easier when the amounts being transferred are not fuzzed).
2. Fuzzing small amounts breaks the transfers as the payee rejects them.

I think that this PR is definitely an improvement on (1), as it makes the behaviour more predictable.

It also resolves (2), assuming that `maxfeepercent` cannot be set to more than 100. The receiving node [rejects the transfer if the amount is twice that of the invoice](https://github.com/ElementsProject/lightning/blob/9a2a09efd64aec614e6d74fa7b68ac5c96a940fd/lightningd/invoice.c#L333). However, I am worried that it effectively completely disables shadow routing for small amounts, although, I suppose, it can be desirable in certain cases. Therefore, probably, not as part of this PR, but I think it is worth considering lifting the restriction on “twice the amount” on the receiving side for small payments (the RFC only says SHOULD, not MUST fail).

darosior:
> So, it only claims that route randomisation respects maxfeepercent, not shadow routing.

Right, amended the OP.

> No way of disabling shadow routing.

Actually you can in developer mode this [PR hid a disabling parameter (`use_shadow`) behind #if DEVELOPER](https://github.com/ElementsProject/lightning/blob/3cfafa81ebdf133d5c30f2ee60da19c4d95ac89f/plugins/pay.c#L1286) :-).

> It also resolves (2), assuming that maxfeepercent cannot be set to more than 100

I think it can, but It does default to 50%.

> I am worried that it effectively completely disables shadow routing for small amounts, although

It doesn't if the receiving peer has channel signaling a `base_fee` which is less than `amount * maxfeepercent`.
However it does in most cases and I'm looking forward to Rusty and/or Christian feedback on this : it seems a lesser evil than effectively paying *more than twice* the requested amount.

> I think it is worth considering lifting the restriction on “twice the amount” on the receiving side for small payments

Maybe, or put a treshold for it to take effect ? Like >10sat.

darosior:
Don't know if I introduced it with this patch, but `test_peerinfo` seems flaky on Travis : I had to restart the `
VALGRIND=1 ARCH=64 DEVELOPER=1 COMPILER=gcc TEST_GROUP=3 TEST_GROUP_COUNT=12 SOURCE_CHECK_ONLY=false` job twice (and another one once).

darosior:
Hmm I'm thinking that maybe we should still add the CLTV delta in this case, just not the amount.

rustyrussell:
OK, so the test sometimes passes before the changes anyway, so let's make it firmer.

Otherwise, excellent.  Needs "Changelog-Fixed: pay: don't send occasional wrong amount for tiny amounts" and "Changelog-Changed: pay: respect maxfeepercent when creating shadow routes" though.

```patch

diff --git a/tests/test_pay.py b/tests/test_pay.py
index c1373652f..af37eb0f9 100644
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -462,18 +462,22 @@ def test_pay_maxfee_shadow(node_factory):
         lambda: len(l1.rpc.listchannels(source=l2.info["id"])["channels"]) > 1
     )
 
-    # A tiny amount, we must not add the base_fee between l2 and l3
-    amount = 2
-    bolt11 = l2.rpc.invoice(amount, "tiny", "tiny")["bolt11"]
-    pay_status = l1.rpc.pay(bolt11)
-    assert pay_status["amount_msat"] == Millisatoshi(amount)
-
-    # A bigger amount, shadow routing could have been used but we set a low
-    # maxfeepercent.
-    amount = 20000
-    bolt11 = l2.rpc.invoice(amount, "bigger", "bigger")["bolt11"]
-    pay_status = l1.rpc.pay(bolt11, maxfeepercent="0.000001")
-    assert pay_status["amount_msat"] == Millisatoshi(amount)
+    # shadow routes are a bit random, so run multiple times.
+    for i in range(5):
+        # A tiny amount, we must not add the base_fee between l2 and l3
+        amount = 2
+        bolt11 = l2.rpc.invoice(amount, "tiny.{}".format(i), "tiny")["bolt11"]
+        pay_status = l1.rpc.pay(bolt11)
+        assert pay_status["amount_msat"] == Millisatoshi(amount)
+
+    # shadow routes are a bit random, so run multiple times.
+    for i in range(5):
+        # A bigger amount, shadow routing could have been used but we set a low
+        # maxfeepercent.
+        amount = 20000
+        bolt11 = l2.rpc.invoice(amount, "big.{}".format(i), "bigger")["bolt11"]
+        pay_status = l1.rpc.pay(bolt11, maxfeepercent="0.000001")
+        assert pay_status["amount_msat"] == Millisatoshi(amount)
 
 
 def test_sendpay(node_factory):
```

rustyrussell:
>     1. No way of disabling shadow routing. As I understand now, [this was intentional](https://github.com/ElementsProject/lightning/pull/3212#discussion_r342386943), however I think it would still be good to have an option for disabling it (e.g. my use-case is that I control multiple nodes connected to each other and I would like to be able to distribute liquidity among them through direct channels, and it is easier when the amounts being transferred are not fuzzed).

You can always use the lowlevel `getroute` and `sendpay` APIs if you want finer control (they can literally be chained together), or if they're direct peers you can make a route pretty easily without `getroute` at all?

darosior:
> OK, so the test sometimes passes before the changes anyway, so let's make it firmer.

Ok, I usually comment out the randomization to test shadow routing :)
https://github.com/ElementsProject/lightning/blob/34ed2093acbc57719d8c63fb33ace41b505d7253/plugins/pay.c#L1107-L1108

kirelagin:
> You can always use the lowlevel `getroute` and `sendpay` APIs if you want finer control (they can literally be chained together),

Yes, that’s what I was considering, but I would also need to `waitsendpay` and, in addition to that, I’m not sure if there are any potential unexpected transient failures that `pay` would retry for me.

> or if they're direct peers you can make a route pretty easily without `getroute` at all?

For that I'd need to know the parameters of the channel, and rather than keeping track of them, it’s much easier to just cal `getroute` to be honest.

rustyrussell:
Yes, pay will retry, but via a different path (if any available) which is probably not what you want?  It won't wait for peers to come online or anything.

PR #3707, opened 2020-05-06
db: unregister sqlite3 trace callback also in error case

theStack:
For sqlite3 versions < 3.14 (i.e. HAVE_SQLITE3_EXPANDED_SQL is not set),
tracing is used to dump statements. The function db_sqlite3_exec()
registers a tracing callback in the beginning and unregisters it at the
end to "avoid it accessing the potentially stale pointer to stmt".
However, the unregistering so far only happened in the success case,
i.e. if the prepare or step calls failed, the callback was still set!

Running the test wallet/test/db-run with sqlite 3.11 leads to a
segmentation fault in the last call to db_commit_transaction():
the tested transaction contains an invalid statement and the (still
registered) trace callback is triggered then by sqlite3_exec() in
db_sqlite3_commit_tx(), leading to a segfault in db_changes_add()
(according to gdb), where it tries to access "stmt->query->readonly".

Changelog-None

### DIFF
--- a/wallet/db_sqlite3.c
+++ b/wallet/db_sqlite3.c
@@ -100,6 +100,7 @@ static bool db_sqlite3_query(struct db_stmt *stmt)
 static bool db_sqlite3_exec(struct db_stmt *stmt)
 {
 	int err;
+	bool success;
 #if !HAVE_SQLITE3_EXPANDED_SQL
 	/* Register the tracing function if we don't have an explicit way of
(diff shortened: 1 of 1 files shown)


theStack:
@rustyrussell: Thanks for reviewing and the hint! I even read STYLE.md days ago but it seems my brain has already forgotten ;) Updated with a forced-push.

PR #3708, opened 2020-05-06
tools/generate-wire: no more lonely messages!

rustyrussell:
When we have only a single member in a TLV (e.g. an optional u64),
wrapping it in a struct is awkward.  This changes it to directly
access those fields.

This is not only more elegant (60 fewer lines), it would also be
more cache friendly.  That's right: cache hot singles!

[I blame @niftynei for the nomenclature which lead to these awful puns, FWIW!]


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3708/commits/edcf57cfaf100ddce52c320994090ffacf4e3063

PR #3721, opened 2020-05-11
pytest: extend timeout for test_payment_duplicate_uncommitted

rustyrussell:
We've been seeing some Travis timeouts under VALGRIND, with the
10 second timeout here: use TIMEOUT as per standard.

Changelog-None


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3721/commits/6aea8bd506c735acadafad6059893afdd592d2ab

PR #3719, opened 2020-05-10
Fix bash completion

mb300sd:
At some point lightning-cli help defaulted to human readable format, the additional -H broke the bash completion.

```
lightning-cli -H help
help=command=autocleaninvoice [cycle_seconds] [expired_by]
category=plugin
description=Set up autoclean of expired invoices.
verbose=Perform cleanup every {cycle_seconds} (default 3600), or disable autoclean if 0. Clean up expired invoices that have expired for {expired_by} seconds (default 86400).
command=balance [unit]
category=plugin
description=List detailed onchain and channel balance information
verbose=List detailed onchain and channel balance information
command=channelstats
category=plugin
description=List detailed channel stats for local node
verbose=List detailed channel stats for local node
command=check command_to_check
category=utility
description=Don't run {command_to_check}, just verify parameters.
verbose=check command_to_check [parameters...]
```

PR #3722, opened 2020-05-11
lightning-cli cleanups

rustyrussell:


cdecker:
How does this interact with https://github.com/ElementsProject/lightning/pull/3719? Could it be that this also addresses the issue in https://github.com/ElementsProject/lightning/pull/3719?

rustyrussell:
Yes: the problems addressed by #3719 are what made me look at this again.  But #3719 is still needed, and serves as a nice backport since it will work with existing c-lightning  nodes so I think we should apply that as is.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3722/commits/a7ee4924994ac769d637a9e4c66d71ac7124bb96

PR #3727, opened 2020-05-15
wally: override wally to use tal-context

niftynei:
For the psbt's, we'd like to be able to tal_steal psbt objects onto the bitcoin_tx they're associated with. To that end, we overwrite the libwally malloc/free functions to use tal.


common/daemon.c
```diff
@@ -119,6 +121,21 @@ static void add_steal_notifiers(const tal_t *root)
 }
 #endif
 
+static void *wally_tal(size_t size)
+{
+	return tal_arr_label(wally_tal_ctx, u8, size, "wally");
```
rustyrussell, line 126: Pretty sure that we changed this later to just call it "wally_notleak"?

common/daemon.c
```diff
@@ -153,7 +170,11 @@ void daemon_setup(const char *argv0,
 
 	/* We handle write returning errors! */
 	signal(SIGPIPE, SIG_IGN);
+
+	/* We set up Wally, the bitcoin wallet lib */
+	wally_tal_ctx = notleak_with_children(tal(NULL, char));
```
rustyrussell, line 175: I think we ran into problems later with this,so chose to use a '_notleak' tal label instead?

common/daemon.c
niftynei: ah yes, updated!

niftynei:
Added the required updates for the unit-tests, plus a patch to get make check-source to work as expected (since we're calling `common_setup()` instead of `setup_locale` one of the checks breaks)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3727/commits/c751810d9a82a3b05a047b9739de3ea3afeb63a9

PR #3728, opened 2020-05-15
pyln: fix incorrect python syntax?

rustyrussell:
```python
contrib/pyln-proto/pyln/proto/bech32.py:120
  /home/rusty/devel/cvs/lightning/contrib/pyln-proto/pyln/proto/bech32.py:120: SyntaxWarning: "is not" with a literal. Did you mean "!="?
    assert decode(hrp, ret) is not (None, None)
```

I think this warning is correct (though I don't see the warning once I installed coincurve:
are we suppressing warnings?)

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3728/commits/01ca2707de3ddfaa1d9724903c2a4723954059cd

PR #3729, opened 2020-05-15
MOVEONLY: Distribute json and wire routines closer to files where they're defined.

rustyrussell:
We used to put all bitcoin primitives into wire/ because otherwise our primitive generation templates wouldn't see them.  Now we have `--include` we can fix this.

The path was a bit tortuous, but it is neater in the end.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3729/commits/0707ba35858ec3d5c0cea1aa7a20aecb8ca9ddff

PR #3732, opened 2020-05-17
Config: revive --commit-fee

darosior:
It has been ignored for quite some time apparently (a quick `git log -S` tells me we don't use it since `88bb38f63b46eeac6ea3f3a29b71dd9aca6cacd0`, but not sure though) and the doc was wrong.


niftynei:
Should we add a pyln test for this flag so that it doesn't get lost again?

darosior:
Yes, I added a commit.

PR #3734, opened 2020-05-18
contrib: allow stderr printing for startup_regtest

niftynei:
Moves the 'daemon'ization from c-lightning to the process level, so that
stderr print messages appear in the terminal. Easier debugging!


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3734/commits/8e935ea90c0ef124cff84f1574c391c78a22b1dc

PR #3738, opened 2020-05-19
Initial PSBT support 

niftynei:
First pass at including PSBT structs into the transaction mechanics. We now can carry witness scripts around in the psbt instead of serializing them separately as `witscripts`, so we strip the `witscript` support.

Finally, as a showpiece, we return the PBST version of the transaction from `txprepare`.


niftynei:
depends on https://github.com/ElementsProject/libwally-core/pull/190 to pass VALGRIND tests

rustyrussell:
Rebased and added latest Wally version.

Ack https://github.com/ElementsProject/lightning/commit/83d13f284996afdb3f8f814a322fcdd8635f7cd5

PR #3740, opened 2020-05-22
PSBT, continued

niftynei:
More internal updates. Here we pull out the input_amount structure and instead always populate the utxo 'parent' of an input with the amount and relevant script pubkey.

This means that we need to start serializing the `last_tx` as a psbt so that we can preserve input amount data for onchaind's signatures, which we do here.

Changelog-None


lightningd/lightningd.c
```diff
@@ -851,6 +851,11 @@ int main(int argc, char *argv[])
 	if (ld->config.keypass)
 		sodium_munlock(ld->config.keypass->data, sizeof(ld->config.keypass->data));
 
+	/* for giggles, we do this here */
+	db_begin_transaction(ld->wallet->db);
+	migrate_last_tx_to_psbt(ld, ld->wallet->db);
+	db_commit_transaction(ld->wallet->db);
+
```
cdecker, line 858: Don't we have the ability to add migration functions to the big array in `db.c`? That way they'd be run only once, and we'd remember that we applied them.

niftynei: We do, yes. The problem is a dependency cycle (sort of) -- the migration code relied on HMSd to fetch the local pubkey for the funding tx but the HSM isn't initialized until after the database has been migrated, so the migration fails.

niftynei: actually i can fix this dependency.

niftynei:
Valgrind errors are caused by a bug I found in libwally-core earlier this week; there's a patch to fix libwally here https://github.com/ElementsProject/libwally-core/pull/200

bitcoin/tx.c
```diff
@@ -164,15 +165,30 @@ int bitcoin_tx_add_input(struct bitcoin_tx *tx, const struct bitcoin_txid *txid,
...
 	assert(wally_err == WALLY_OK);
 	input->features = chainparams->is_elements ? WALLY_TX_IS_ELEMENTS : 0;
 	wally_tx_add_input(tx->wtx, input);
 	psbt_add_input(tx->psbt, input, i);
 
-
+	if (input_wscript)
+		/* Add the prev output's data into the PSBT struct */
+		psbt_input_set_prev_utxo_wscript(tx->psbt, i, input_wscript, amount);
+	else if (scriptPubkey) {
```
cdecker, line 200:
```suggestion
	if (input_wscript) {
		/* Add the prev output's data into the PSBT struct */
		psbt_input_set_prev_utxo_wscript(tx->psbt, i, input_wscript, amount);
	} else if (scriptPubkey) {
```

Nit: just makes it clearer that the comment is not what's guarded by the `if`-statement :-)

bitcoin/tx.h
```diff
@@ -77,9 +77,16 @@ int bitcoin_tx_add_output(struct bitcoin_tx *tx, const u8 *script,
...
+ * For P2WSH inputs, we'll also store the wscript and/or scriptPubkey
+ * Passing in just the {input_wscript}, we'll generate the scriptPubkey for you.
+ * In some cases we may not have the wscript, in which case the scriptPubkey
+ * should be provided. We'll check that it's P2WSH before saving it */
 int bitcoin_tx_add_input(struct bitcoin_tx *tx, const struct bitcoin_txid *txid,
-			 u32 outnum, u32 sequence,
-			 struct amount_sat amount, u8 *script);
+			 u32 outnum, u32 sequence, const u8 *scriptSig,
+			 struct amount_sat amount, const u8 *scriptPubkey,
+			 const u8 *input_wscript);
```
cdecker, line 95: I'm sure I will get those 3 `u8[]` params wrong at some point xD

bitcoin/tx.c
```diff
@@ -479,6 +479,38 @@ char *bitcoin_tx_to_psbt_base64(const tal_t *ctx, struct bitcoin_tx *tx)
...
+	struct wally_psbt *tmppsbt;
+	struct bitcoin_tx *tx = bitcoin_tx(ctx, chainparams,
+					   psbt->tx->num_inputs,
+					   psbt->tx->num_outputs,
+					   psbt->tx->locktime);
+	wally_tx_free(tx->wtx);
+
+	/* We want the 'finalized' tx since that includes any signature
+	 * data, not the global tx. But 'finalizing' a tx destroys some fields
+	 * so we 'clone' it first and then finalize it */
```
cdecker, line 511: Wat? :facepalm:

bitcoin/psbt.c
```diff
@@ -20,6 +20,18 @@
 	memmove((arr) + (pos), (arr) + (pos) + 1,		\
 		sizeof(*(arr)) * ((num) - ((pos) + 1)))
 
+/* FIXME: someday this will break, because it's been exposed in libwally */
+int wally_psbt_clone(struct wally_psbt *psbt, struct wally_psbt **output)
```
cdecker, line 24: ~Isn't this an exact copy of the one found in `tx.c` added in commit 08a677de5e90b01bdb55f6b8a4b0a2a2f839d44c?~

Edit: Nope, it isn't `psbt` and `tx` are different primitives and warrant their own clone function :+1:

cdecker:
Seems your patch made it into libwally, should we pull in the new version now or do you prefer bundling the changes and pull them in all at once?

bitcoin/psbt.c
```diff
@@ -141,19 +141,17 @@ void psbt_rm_output(struct wally_psbt *psbt,
...
-		u8 *pbt_bytes = tal_arr(NULL, u8, room);
-		size_t bytes_written;
-		if (wally_psbt_to_bytes(psbt, pbt_bytes, room, &bytes_written) == WALLY_OK) {
-			towire_u32(pptr, bytes_written);
-			towire_u8_array(pptr, pbt_bytes, bytes_written);
-			tal_free(pbt_bytes);
-			return;
-		}
-		tal_free(pbt_bytes);
+	size_t room = 1024 * 1000;
```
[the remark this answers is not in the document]
niftynei: I can't remember the exact context for this, but libwally has a memory leak that causes valgrind errors if we loop through to discover the correct length; the `wally_psbt_get_length` method is not exposed external to the library at the time this method was written. https://github.com/ElementsProject/libwally-core/pull/191 might fix this?

bitcoin/psbt.c
```diff
@@ -137,6 +141,72 @@ void psbt_rm_output(struct wally_psbt *psbt,
...
+		assert(is_p2wsh(scriptPubkey, NULL) || is_p2wpkh(scriptPubkey, NULL)
+		       || is_p2sh(scriptPubkey, NULL));
+		scriptpk = cast_const(u8 *, scriptPubkey);
+	} else {
+		/* Adding a NULL scriptpubkey is an error, *however* there is the
+		 * possiblity we're spending a UTXO that we didn't save the
+		 * scriptpubkey data for. in this case we set it to an 'empty'
+		 * or zero-len script */
+		scriptpk = tal_arr(psbt, u8, 1);
+		scriptpk[0] = 0x00;
```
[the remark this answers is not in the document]
niftynei: Since this is an Absolute Hack, I prefer the more explicit version (you can easily tell at a glance that we're setting the script to 0x00)

bitcoin/tx.h
niftynei: Hm yeah a type system might be nice here ;)

wallet/db.c
```diff
@@ -1111,6 +1115,78 @@ static void migrate_our_funding(struct lightningd *ld, struct db *db)
...
+		get_channel_basepoints(ld, &peer_id, db_column_u64(stmt, 0),
+				       &local_basepoints, &local_funding_pubkey);
+
+		funding_wscript = bitcoin_redeem_2of2(stmt, &local_funding_pubkey,
+						      &remote_funding_pubkey);
+		/* FIXME: input index for funding will not always be zero
+		 * but at this current update juncture, it is! */
+		psbt_input_set_prev_utxo_wscript(last_tx->psbt,
+						 0, funding_wscript,
+						 funding_sat);
```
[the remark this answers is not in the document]
niftynei: good point. removed

wallet/db.c
```diff
@@ -1111,6 +1115,78 @@ static void migrate_our_funding(struct lightningd *ld, struct db *db)
...
+
+		if (!db_column_signature(stmt, 5, &last_sig.s))
+			abort();
```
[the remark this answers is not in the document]
niftynei: I'm not sure what you're asking here. It should definitely not ever be NULL -- we don't open a channel until we've saved the last_sig for the original commit tx.

tests/test_db.py
```diff
@@ -140,6 +143,23 @@ def test_scid_upgrade(node_factory, bitcoind):
     assert l1.db_query('SELECT failchannel from payments;') == [{'failchannel': '103x1x1'}]
 
 
+@unittest.skipIf(not COMPAT, "needs COMPAT to convert obsolete db")
+@unittest.skipIf(os.getenv('TEST_DB_PROVIDER', 'sqlite3') != 'sqlite3', "This test is based on a sqlite3 snapshot")
+@unittest.skipIf(TEST_NETWORK != 'regtest', "The network must match the DB snapshot")
+def test_last_tx_psbt_upgrade(node_factory, bitcoind):
+    bitcoind.generate_block(12)
+
```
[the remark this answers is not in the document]
niftynei: Yes! It takes the code used to generate the DB 12 blocks to get all of its channels opened

niftynei:
I've got a patch to update libwally that'll add to this PR.

PR #3741, opened 2020-05-22
listtransactions: represent txids as little endian

darosior:
This made my copy/paste to Esplora cumbersome :/

Was this intended ? If so maybe the changelog line is wrong and the old behaviour should be guarded behind `deprecated_apis` ...


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3741/commits/b9097711ba3e3ed752f27e48565a76e66620e339

This is technically a break, but yech, obviously nobody was using the field before, so it's ok!

PR #3745, opened 2020-05-22
Added hsm_new_channel and hsm_ready_channel messages.

ksedgwic:
This PR adds two new messages to the `hsm_wire.csv`.  The first message, `hsm_new_channel` is sent from `lightningd` when an uncommited channel is created.  The second message, `hsm_ready_channel` is sent by `openingd` when the complete set of static channel parameters has been negotiated.

Since these parameters are static for the channel, the upcoming remote signer API will require them early.  This PR provides them early to the HSM daemon process so that this can be achieved.

### DIFF
--- a/hsmd/hsm_wire.csv
+++ b/hsmd/hsm_wire.csv
@@ -21,6 +21,16 @@ msgtype,hsm_init_reply,111
 msgdata,hsm_init_reply,node_id,node_id,
 msgdata,hsm_init_reply,bip32,ext_key,
 
+# Declare a new channel.
+msgtype,hsm_new_channel,25
+# Which identity to use for requests
+msgdata,hsm_new_channel,id,node_id,
+# Database id for this client.
+msgdata,hsm_new_channel,dbid,u64,
+
+# No value returned.
+msgtype,hsm_new_channel_reply,125
+
 # Get a new HSM FD, with the specified capabilities
 msgtype,hsm_client_hsmfd,9
 # Which identity to use for requests
@@ -41,6 +51,26 @@ msgdata,hsm_get_channel_basepoints,dbid,u64,
 msgtype,hsm_get_channel_basepoints_reply,110
 msgdata,hsm_get_channel_basepoints_reply,basepoints,basepoints,
 msgdata,hsm_get_channel_basepoints_reply,funding_pubkey,pubkey,
+
+# Provide channel parameters.
+msgtype,hsm_ready_channel,24
+msgdata,hsm_ready_channel,is_outbound,bool,
+msgdata,hsm_ready_channel,channel_value,amount_sat,
+msgdata,hsm_ready_channel,push_value,amount_msat,
+msgdata,hsm_ready_channel,funding_txid,bitcoin_txid,
+msgdata,hsm_ready_channel,funding_txout,u16,
+msgdata,hsm_ready_channel,local_to_self_delay,u16,
+msgdata,hsm_ready_channel,local_shutdown_script_len,u16,
+msgdata,hsm_ready_channel,local_shutdown_script,u8,local_shutdown_script_len
+msgdata,hsm_ready_channel,remote_basepoints,basepoints,
+msgdata,hsm_ready_channel,remote_funding_pubkey,pubkey,
+msgdata,hsm_ready_channel,remote_to_self_delay,u16,
+msgdata,hsm_ready_channel,remote_shutdown_script_len,u16,
+msgdata,hsm_ready_channel,remote_shutdown_script,u8,remote_shutdown_script_len
+msgdata,hsm_ready_channel,option_static_remotekey,bool,
+
+# No value returned.
+msgtype,hsm_ready_channel_reply,124
 
 # Return signature for a funding tx.
 #include <common/utxo.h>
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -1456,6 +1456,74 @@ static struct io_plan *pass_client_hsmfd(struct io_conn *conn,
 			     send_pending_client_fd, c);
 }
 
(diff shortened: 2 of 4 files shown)


rustyrussell:
This implies a degree of state in the HSM.  A stateless approach would be to have the hsm return a signed blob, which gets handed back in all the HSM calls. HSM is @cdecker 's baby, so I'll let him comment on this.

cdecker:
Good point @rustyrussell, I'd really like to keep the hsm as stateless as possible, so it might be a good idea to just have the HSM return an encrypted representation of the information in `new_channel` and `ready_channel` back as a reply to the call, and then pass that back in into the HSM on subsequent calls. If that's not required, the HSM can just return an empty blob, and there would be no loss in functionality, but it keeps our options open for hardware wallet devices and multi-tenant setups in which statelessness is a security measure.

devrandom:
I believe it's impossible to have the HSM be stateless while keeping it secure.  With a stateless HSM, you can supply it with an old state and cause it to breach by signing a revoked commitment transaction.  In order for it to detect such a rollback, it needs at least the current commitment number in its own secure storage, per channel.  To do robust policy checks it also needs to store the hash of the latest commitment transaction (which covers balances, HTLCs in flight, etc).

I am assuming here that the goal of the HSM is to be able to prevent channel fund loss even if the other components of the node are compromised.  That's certainly the goal of the Lightning Signer project.

So given that the HSM does need some state, the question is whether it might as well store all of its state, or whether it can be slimmer by externalizing some of its state to the node.  That seems to be something that should be up to the designer of the particular HSM implementation.  Definitely agree that providing some facility to externalize state can provide flexibility to the implementer.

Now, it seems that any call which advances state should consume and output this encrypted/signed state blob.  So this would include signing of commitment transactions and closing transactions.

Does the above sound reasonable?  i.e. that complete statelessness should not be a goal? and that all state advancing methods get to use the state blob facility?

cdecker:
> I believe it's impossible to have the HSM be stateless while keeping it secure. With a stateless HSM, you can supply it with an old state and cause it to breach by signing a revoked commitment transaction. In order for it to detect such a rollback, it needs at least the current commitment number in its own secure storage, per channel. To do robust policy checks it also needs to store the hash of the latest commitment transaction (which covers balances, HTLCs in flight, etc).

You are right, I forgot to include the details of how such an HSM would work. You are correct that a completely stateless HSM is not able to distinguish a replay from a new version, and hence would be vulnerable to host takeover.

Let me describe the minimal requirements I had in mind for the HSM: let's assume we have an HSM that has a single persistent integer counter (for the sake of simplicity let's just assume it can't wrap around, in practice it'd just be a 64 bit integer, which should be large enough for our purposes). Each time the client asks the HSM for some action it passes along an encrypted or signed blob. At the end of each operation the HSM increments the internal counter, builds a blob containing all its internal state, including the counter, encrypts it, and sends it back to the caller (to which all of this data is opaque). On the next call the caller will pass the latest blob back to the HSM, which will decrypt the blob or verify its integrity, and rehydrate its internal state with the information from the blob.

With this interface we have minimal requirements when it comes to the hardware of the HSM and we make the least assumptions about the HSM's capabilities. It is far easier for an HSM manufacturer to include volatile memory that can be rehydrated from the blob, than including loads of non-volatile persistent memory on the hardware. If the HSM happens to have sufficient internal persistent storage then it can just return an empty blob since it can fully store everything internally. If on the other hand the HSM has but one persistent counter it can externalize the persistence on the client.

I think it's important to minimize assumptions, especially when it comes to hardware, and there is a large difference between having sufficient internal persistent storage for the entire state, and having a single `u64` persistent storage. In addition this setup can allow us to setup multiple HSMs with the same keys initially, and swap faulty HSMs out for a new one with the identical secret with the internal counter being a high water mark (in this case a replay would require physical access and a replay of an old blob/request tuple).

devrandom:
That sounds like a reasonable plan.

I think there are a few further details to consider:

- which calls should accept the state blob as an input
- which calls should return a new state blob
- storage efficiency

It seems like any call that applies a signature should be considered a state modifier.  Any signing of a commitment transaction or a closing transaction should be memorized.  Even signing an invoice should probably be memorized, because that could be used to make policy decisions inside the signer.

As to storage efficiency - we should consider separating node state from channel state.  If we have thousands of channels, we may not want to re-parse and update the entire state, but just update the channel state that moved forward.

In any case, this seems like a significant code change by itself and requires some design thought.  Should Ken and I implement this as a followup PR?

cdecker:
Excellent question, I think we need to indeed pass in the context into most of the `hsmd` calls that end up signing anything. It might be worth splitting the context into channel-specific and daemon-specific bundles rather than having a global one that needs to be shared between all daemons. If we were to have a global one we'd end up having to serialize through `lightningd` in order to make sure that they are in sync.

Daemon- / channel-specific appears to be good enough to me, since it can still ensure we don't go backwards on any channel, but it might require multiple counters. An alternative could also be that we just add a proxy in the middle that linearizes calls to the `hsmd` and re-establishes the single-counter requirement. With the interposed proxy we can also move the blob-handling outside of `lightningd` altogether and potentially have custom protocols/transports to talk to the actual `hsmd`.

ksedgwic:
Rebased.

cdecker:
Kicked Travis due to flaky test.

ksedgwic:
So should the proxy be a follow-on effort?  What changes are needed for the current PR?

cdecker:
Sorry for not replying sooner. I think for now the proxy is the better approach, allowing us not to commit to stateful hsmd implementations first, which is simpler. Should we decide to add state to the `hsmd` it'll be easy to take up the experiences we have from the proxy and fold them back into the `hsmd`.

PR #3749, opened 2020-05-26
Ensure txs across the wire always have input amounts.

rustyrussell:
This is a gradual move towards splitting `tx` into two types: psbts which have complete information required for signing and `tx_parts` which may even be missing inputs or outputs which are not relevant.

This helps the move to psbts, which are better for dual funding and other manipulations, but also helps moving to a future where HTLC txs (and with eltoo, even commitment txs) are combined with unrelated things so onchaind doesn't need to know the entire tx.

Changelog-None


bitcoin/tx_parts.c
```diff
@@ -0,0 +1,396 @@
...
+	u8 *script;
+	struct wally_tx_witness_stack *ws;
+	int ret;
+
+	fromwire_bitcoin_txid(cursor, max, &txid);
+	index = fromwire_u32(cursor, max);
+	sequence = fromwire_u32(cursor, max);
+	script = fromwire_tal_arrn(tmpctx,
+				   cursor, max, fromwire_u32(cursor, max));
+	/* libwally doesn't like non-NULL ptrs with zero lengths. */
```
niftynei, line 170: :sweat_smile:

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3749/commits/d15410b0a89fd6a4fbcfbb446ca020f844327101

PR #3752, opened 2020-05-29
sort listinvoices and listsendpays by order of creation.

fiatjaf:
I don't know about SQLite, but Postgres doesn't guarantee that rows are ordered by the id or insertion order, so I'm constantly getting bizarre sortings (like `125346`) that cause confusion when listing invoices and payments. This is an easy fix.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3752/commits/5db61888b06bea062c03a61233bd38d0d8a548ca

PR #3753, opened 2020-05-29
paymod: Payments reimagined (part 01)

cdecker:
_This is only a partial implementation of the new payment flow, but given the
looming release deadline and the sheer size of the change, I thought it's best
to get some code reviews in now_

## Design

The current payment flow is a multi-step process in the `pay` plugin (and some
others) that grew organically as logic and more features were added over the
time. This PR is my attempt to reimagine and consolidate the payment flow, to
consolidate some commonalities, better compartmentalize the existing addons,
and facilitate future modifications of the flow.

In order to get to this lofty goal I separated the logic into two separate
parts: the `payment` which consists of the core functionality required to
perform a payment, and `payment_modifier`s which have associated data and are
called at each step of the payment flow to allow them to modify the payment
behavior. The core payment flow consists of the following steps:

base-payment-state-machine

<details>

```
digraph {
  initialized -> got_route -> sendonion -> failed;
  sendonion -> success;
}
```

</details>

After each step is complete all modifiers registered with the flow will be
called, and they are allowed to modify their local data and the data of the
payment in order to fulfill their purpose. In particular they are allowed to
set a payment into any state, including two states (`retry` and `split`) that
are not part of the core flow. These are indications that the payment failed,
but was retried with different parameters or was split into multiple
sub-payments, e.g., in MPP.

This brings us to the structure of payments: payments and their sub-payments
are organized in a tree, with the inner nodes representing failed attempts
that have been retried in some form, and leafs representing the currently
active or final attempts. The root of the tree is the original attempt and
serves as a coordination point for all sub-payments, aggregating information
that is shared among all attempts such as a list of `channel_hint`s used to
compute exclusions, and the invoice information that initiated the payment
flow.

This structure allows modifiers to implement almost arbitrary logic, while
keeping the core logic clean. Modifiers (both planned and implemented)
include:

 - `final_cltv`: the invoice can specify a final CLTV offset that just adds to
   the route CLTV deltas, so calling after `got_route` we can adjust them to
   match.
 - `routehints`: these change the destination for which we call `getroute` and
   once in `got_route` we can append the current routehint to form complete
   route.
 - `waitblockheight`: if a node reports that our CLTV is too low we can infer
   the height we should be at and then wait to reach that height before
   re-attempting.
 - `shadow_route`: since this is just an offset applied to the values we can
   compute the shadow route and then adjust the amount passed to the
   `getroute` call.
 - `retry`: if a payment fails we can just retry for a fixed number of times
   with different routes.
 - `split`: this is the core functionality of MPP, splitting a payment into
   multiple sub-payments if it fails.
 - `keysend` is just adjusting the TLV payload passed to `createonion` to
   include the preimage.
   
There are certainly some suboptimal solutions, however I think this structure
can be used to better experiment and implement future extensions rather than
having to weave the functionality into the existing flow (keysend sending
functionality for example would not have any need for the `routehints`
modifier, but the rest is shared).

## Testing

The functionality is implemented in parallel to the existing `json_pay`
infrastructure, but is planned to become a drop-in replacement, allowing us to
eventually remove the old implementation and just use the new one. To test the
new flow you can use `lightning-cli paymod [args]` instead of `lightning-cli
pay [args]` though some modifiers are not implemented yet, and some are in the
next parts of this PR.

I am testing feature parity by changing `pay.c` to use `json_paymod` instead
of `json_pay`, allowing me to run the existing (unmodified) tests against the
new logic.

## The ugly parts:

I'm calling these out because I am currently not addressing them and they
aren't a high priority, but we should clean these up eventually, and they may
cause reviewers to stumble a bit:

 - Modifiers need to call `payment_continue(p)` to advance the state-machine,
   and I forgot that a couple of times. Something like we do with
   `command_still_pending` and friends would be nice.
 - The RPC calls being bound to an originating `command` instance is rather
   strange, even though it doesn't do much (you can pass `NULL` to it), since
   we now potentially have multiple concurrent RPC calls in flight, if one of
   them were to respond to the original `cmd` it'd break things. Making the
   RPC interface independent of having a `command` instance would be a good
   idea.
 - I started actually parsing the JSON-RPC request and responses into
   `struct`s to facilitate modifying them, however the functions and struct
   definitions are spread all over the place. I'd like to consolidate them
   under `common/` eventually, so we can use them on both the `lightningd` and
   the plugin side, and not have to replicate the extraction of data from JSON
   responses all over the place.

## Next up

I have most of the functionality implemented, and we're down to 12 failing
tests in my [development branch][pay-reorg-branch]. I'll publish the next part
as soon as it settles down, i.e., doesn't get changed anymore, and I'll defer
cleanups and renaming to a last cleanup part in order not to delay the overall
progress and release.

Given the very new and unproven nature I'd propose we release this for
`keysend` sending support directly, but gate the replacement of the `pay`
command behind either a `!COMPAT_090` or the `DEVELOPER` flag. This would
minimize the impact to the majority of users, but still allow us to collect
experience from `keysend` and users that opted into testing it. However it'd
also mean that `mpp-pay` is a separate command or a developer-only command for
this release.

Changelog-None

[pay-reorg-branch]: https://github.com/cdecker/lightning/tree/pay-reorg

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1,4 +1,5 @@
 #include <plugins/libplugin-pay.h>
+#include <stdio.h>
 
 struct payment *payment_new(tal_t *ctx, struct command *cmd,
 			    struct payment *parent,
@@ -144,3 +145,19 @@ void payment_continue(struct payment *p)
 	 * `payment_continue` after the final state. */
 	abort();
 }
+
+static inline struct dummy_data *
+dummy_data_init(struct payment *p)
+{
+	return tal(p, struct dummy_data);
+}
+
+static inline void dummy_step_cb(struct dummy_data *dd,
+				 struct payment *p)
+{
+	fprintf(stderr, "dummy_step_cb called for payment %p at step %d\n", p, p->step);
+	payment_continue(p);
+}
+
+REGISTER_PAYMENT_MODIFIER(dummy, struct dummy_data *, dummy_data_init,
+			  dummy_step_cb);
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -579,8 +579,12 @@ static void payment_finished(struct payment *p)
 	assert((result.leafstates & PAYMENT_STEP_SUCCESS) == 0 ||
 	       result.preimage != NULL);
 
-	if (p->parent == NULL) {
-		assert(p->cmd != NULL);
+	if (p->parent == NULL && cmd == NULL) {
+		/* This is the tree root, but we already reported success or
+		 * failure, so noop. */
+		return;
+
+	}  else if (p->parent == NULL) {
 		if (payment_is_success(p)) {
 			assert(result.treestates & PAYMENT_STEP_SUCCESS);
 			assert(result.leafstates & PAYMENT_STEP_SUCCESS);
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -37,11 +37,44 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 	return p;
 }
 
+/* Generic handler for RPC failures that should end up failing the payment. */
+static struct command_result *payment_rpc_failure(struct command *cmd,
+						  const char *buffer,
+						  const jsmntok_t *toks,
+						  struct payment *p)
+{
+	plugin_log(p->cmd->plugin, LOG_DBG,
+		   "Failing a partial payment due to a failed RPC call: %.*s",
+		   toks->end - toks->start, buffer + toks->start);
+
+	p->step = PAYMENT_STEP_FAILED;
+	payment_continue(p);
+	return command_still_pending(cmd);
+}
+
+static struct command_result *payment_getinfo_success(struct command *cmd,
+						      const char *buffer,
+						      const jsmntok_t *toks,
+						      struct payment *p)
+{
+	const jsmntok_t *blockheighttok =
+	    json_get_member(buffer, toks, "blockheight");
+	json_to_number(buffer, blockheighttok, &p->start_block);
+	payment_continue(p);
+	return command_still_pending(cmd);
+}
+
 void payment_start(struct payment *p)
 {
 	p->step = PAYMENT_STEP_INITIALIZED;
 	p->current_modifier = -1;
-	payment_continue(p);
+
+	/* TODO If this is not the root, we can actually skip the getinfo call
+	 * and just reuse the parent's value. */
+	send_outreq(p->cmd->plugin,
+		    jsonrpc_request_start(p->cmd->plugin, NULL, "getinfo",
+					  payment_getinfo_success,
+					  payment_rpc_failure, p));
 }
 
 static bool route_hop_from_json(struct route_hop *dst, const char *buffer,
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -174,3 +174,39 @@ static inline void dummy_step_cb(struct dummy_data *dd,
 
 REGISTER_PAYMENT_MODIFIER(dummy, struct dummy_data *, dummy_data_init,
 			  dummy_step_cb);
+
+static struct retry_mod_data *retry_data_init(struct payment *p);
+
+static inline void retry_step_cb(struct retry_mod_data *rd,
+				 struct payment *p);
+
+REGISTER_PAYMENT_MODIFIER(retry, struct retry_mod_data *, retry_data_init,
+			  retry_step_cb);
+
+static struct retry_mod_data *
+retry_data_init(struct payment *p)
+{
+	struct retry_mod_data *rdata = tal(p, struct retry_mod_data);
+	struct retry_mod_data *parent_rdata;
+	if (p->parent != NULL) {
+		parent_rdata = payment_mod_retry_get_data(p->parent);
+		rdata->retries = parent_rdata->retries - 1;
+	} else {
+		rdata->retries = 10;
+	}
+	return rdata;
+}
+
+static inline void retry_step_cb(struct retry_mod_data *rd,
+				 struct payment *p)
+{
+	struct payment *subpayment;
+	struct retry_mod_data *rdata = payment_mod_retry_get_data(p);
+	if (p->step == PAYMENT_STEP_FAILED && rdata->retries > 0) {
+		subpayment = payment_new(p, p->cmd, p, p->modifiers);
+		payment_start(subpayment);
+		p->step = PAYMENT_STEP_RETRY;
+	}
+
+	payment_continue(p);
+}
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -162,7 +162,61 @@ static void payment_getroute(struct payment *p)
 
 static void payment_compute_onion_payloads(struct payment *p)
 {
+	struct createonion_request *cr;
+	size_t hopcount;
+	static struct short_channel_id all_zero_scid;
+	struct createonion_hop *cur;
 	p->step = PAYMENT_STEP_ONION_PAYLOAD;
+	hopcount = tal_count(p->route);
+
+	/* Now compute the payload we're about to pass to `createonion` */
+	cr = p->createonion_request = tal(p, struct createonion_request);
+	cr->assocdata = tal_arr(cr, u8, 0);
+	towire_sha256(&cr->assocdata, p->payment_hash);
+	cr->session_key = NULL;
+	cr->hops = tal_arr(cr, struct createonion_hop, tal_count(p->route));
+
+	/* Non-final hops */
+	for (size_t i = 0; i < hopcount - 1; i++) {
+		/* The message is destined for hop i, but contains fields for
+		 * i+1 */
+		cur = &cr->hops[i];
+		cur->style = p->route[i].style;
+		cur->pubkey = p->route[i].nodeid;
+		switch (cur->style) {
+		case ROUTE_HOP_LEGACY:
+			cur->legacy_payload =
+			    tal(cr->hops, struct legacy_payload);
+			cur->legacy_payload->forward_amt =
+			    p->route[i + 1].amount;
+			cur->legacy_payload->scid = p->route[i + 1].channel_id;
+			cur->legacy_payload->outgoing_cltv =
+			    p->start_block + p->route[i + 1].delay;
+			break;
+		case ROUTE_HOP_TLV:
+			/* TODO(cdecker) Implement */
+			abort();
+		}
+	}
+
+	/* Final hop */
+	cur = &cr->hops[hopcount - 1];
+	cur->style = p->route[hopcount - 1].style;
+	cur->pubkey = p->route[hopcount - 1].nodeid;
+	switch (cur->style) {
+	case ROUTE_HOP_LEGACY:
+		cur->legacy_payload = tal(cr->hops, struct legacy_payload);
+		cur->legacy_payload->forward_amt =
+		    p->route[hopcount - 1].amount;
+		cur->legacy_payload->scid = all_zero_scid;
+		cur->legacy_payload->outgoing_cltv =
+		    p->start_block + p->route[hopcount - 1].delay;
+		break;
+	case ROUTE_HOP_TLV:
+		/* TODO(cdecker) Implement */
+		abort();
+	}
+
 	/* Now allow all the modifiers to mess with the payloads, before we
 	 * serialize via a call to createonion in the next step. */
 	payment_continue(p);
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1,0 +1,146 @@
+#include <plugins/libplugin-pay.h>
+
+struct payment *payment_new(tal_t *ctx, struct command *cmd,
+			    struct payment *parent,
+			    struct payment_modifier **mods)
+{
+	struct payment *p = tal(ctx, struct payment);
+	p->children = tal_arr(p, struct payment *, 0);
+	p->parent = parent;
+	p->modifiers = mods;
+	p->cmd = cmd;
+	p->start_time = time_now();
+	p->partid = partid++;
+
+	/* Copy over the relevant pieces of information. */
+	if (parent != NULL) {
+		tal_arr_expand(&parent->children, p);
+		p->destination = p->getroute_destination = parent->destination;
+		p->amount = parent->amount;
+		p->payment_hash = parent->payment_hash;
+	}
+
+	/* Initialize all modifier data so we can point to the fields when
+	 * wiring into the param() call in a JSON-RPC handler. The callback
+	 * can also just `memcpy` the parent if this outside access is not
+	 * required. */
+	p->modifier_data = tal_arr(p, void *, 0);
+	for (size_t i=0; mods[i] != NULL; i++) {
+		if (mods[i]->data_init != NULL)
+			tal_arr_expand(&p->modifier_data,
+				       mods[i]->data_init(p));
+		else
+			tal_arr_expand(&p->modifier_data, NULL);
+	}
+
+	return p;
+}
+
+void payment_start(struct payment *p)
+{
+	p->step = PAYMENT_STEP_INITIALIZED;
+	p->current_modifier = -1;
+	payment_continue(p);
+}
+
+static void payment_getroute(struct payment *p)
+{
+	p->step = PAYMENT_STEP_GOT_ROUTE;
+	return payment_continue(p);
+}
+
+static void payment_compute_onion_payloads(struct payment *p)
+{
+	p->step = PAYMENT_STEP_ONION_PAYLOAD;
+	/* Now allow all the modifiers to mess with the payloads, before we
+	 * serialize via a call to createonion in the next step. */
+	payment_continue(p);
+}
+
+static void payment_sendonion(struct payment *p)
+{
+	p->step = PAYMENT_STEP_FAILED;
+	return payment_continue(p);
+}
+
+/* Mutual recursion. */
+static void payment_finished(struct payment *p);
+
+/* Function to bubble up completions to the root, which actually holds on to
+ * the command that initiated the flow. */
+static struct command_result *payment_child_finished(struct payment *parent,
+						     struct payment *child)
+{
+	/* TODO implement logic to wait for all parts instead of bubbling up
+	 * directly. */
+
+	/* Should we continue bubbling up? */
+	if (parent->parent == NULL) {
+		assert(parent->cmd != NULL);
+		return payment_finished(parent);
+	} else {
+		return payment_child_finished(parent->parent, parent);
+	}
+}
+
+/* This function is called whenever a payment ends up in a final state, or all
+ * leafs in the subtree rooted in the payment are all in a final state. It is
+ * called only once, and it is guaranteed to be called in post-order
+ * traversal, i.e., all children are finished before the parent is called. */
+static void payment_finished(struct payment *p)
+{
+	/* TODO If this is a success bubble it back up through the part
+	 * tree. If it is a failure, decide here whether to split, retry or
+	 * split (maybe in a modifier instead?). */
+	if (p->parent == NULL)
+		return command_fail(p->cmd, JSONRPC2_INVALID_REQUEST, "Not functional yet");
+	else
+		return payment_child_finished(p->parent, p);
+}
+
+void payment_continue(struct payment *p)
+{
+	struct payment_modifier *mod;
+	void *moddata;
+	/* If we are in the middle of calling the modifiers, continue calling
+	 * them, otherwise we can continue with the payment state-machine. */
+	p->current_modifier++;
+	mod = p->modifiers[p->current_modifier];
+
+	if (mod != NULL) {
+		/* There is another modifier, so call it. */
+		moddata = p->modifier_data[p->current_modifier];
+		return mod->post_step_cb(moddata, p);
+	} else {
+		/* There are no more modifiers, so reset the call chain and
+		 * proceed to the next state. */
+		p->current_modifier = -1;
+		switch (p->step) {
+		case PAYMENT_STEP_INITIALIZED:
+			payment_getroute(p);
+			return;
+
+		case PAYMENT_STEP_GOT_ROUTE:
+			payment_compute_onion_payloads(p);
+			return;
+
+		case PAYMENT_STEP_ONION_PAYLOAD:
+			payment_sendonion(p);
+			return;
+
+		case PAYMENT_STEP_SUCCESS:
+		case PAYMENT_STEP_FAILED:
+			payment_finished(p);
+			return;
+
+		case PAYMENT_STEP_RETRY:
+		case PAYMENT_STEP_SPLIT:
+			/* Do nothing, we'll get pinged by a child succeeding
+			 * or failing. */
+			return;
+		}
+	}
+	/* We should never get here, it'd mean one of the state machine called
+	 * `payment_continue` after the final state. */
+	abort();
+}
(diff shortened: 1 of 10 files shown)


plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+			cur->legacy_payload->forward_amt =
+			    p->route[i + 1].amount;
+			cur->legacy_payload->scid = p->route[i + 1].channel_id;
+			cur->legacy_payload->outgoing_cltv =
+			    p->start_block + p->route[i + 1].delay;
+			break;
+		case ROUTE_HOP_TLV:
+			cur->tlv_payload = tlv_tlv_payload_new(cr->hops);
+			tlvstream_set_tu64(
+			    &cur->tlv_payload->fields, 2,
```
niftynei, line 478: are there conts/enums we can use for the type number instead of the straight int?

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+		agg |= payment_aggregate_states(p->children[i]);
+
+	return agg;
+}
+
+/* A payment is finished if a) it is in a final state, of b) it's in a
+ * child-spawning state and all of its children are in a final state. */
+static bool payment_is_success(struct payment *p)
+{
+	return (payment_aggregate_states(p) & PAYMENT_STEP_SUCCESS) != 0;
```
niftynei, line 594: this seems like it'll return true if any of the child leaves are successful -- which may not indicate that all of its children are in a final state (as the code comment above indicates)

should you also check that the payment_is_finished in conjunction with the aggregate state check?

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+{
+	if (!payment_is_finished(p))
+		return command_still_pending(NULL);
+
+	/* Should we continue bubbling up? */
+	return payment_finished(p);
+}
+
+/* This function is called whenever a payment ends up in a final state, or all
+ * leafs in the subtree rooted in the payment are akk in a final state. It is
```
niftynei, line 610:
```suggestion
 * leafs in the subtree rooted in the payment are in a final state. It is
```

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+{
+	struct payment_tree_result result = payment_collect_result(p);
+	struct json_stream *ret;
+	struct command *cmd = p->cmd;
+
+	p->end_time = time_now();
+
+	/* Either none of the leaf attempts succeeded yet, or we have a
+	 * preimage. */
+	assert((result.treestates & PAYMENT_STEP_SUCCESS) == 0 ||
```
niftynei, line 623: your comment says 'leaf attempts' but you're checking the 'treestates' here?

niftynei:
The modifier design is really nice and modular! very cool :sunglasses:

plugins/libplugin-pay.h
```diff
@@ -0,0 +1,217 @@
...
+	struct channel_status *peer_channels;
+
+	/* The blockheight at which the payment attempt was
+	 * started.  */
+	u32 start_block;
+
+	struct timeabs start_time, end_time;
+	struct timeabs deadline;
+
+	struct amount_msat extra_budget;
```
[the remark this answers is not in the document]
cdecker: Isn't used just yet, but it is supposed to be our fee and fuzz allowance for the (partial) payment. It'll be computed as percentage of the amount sent to the destination, and fails routes if it is exceeded. It's here so modifiers can change it if required (a `exemptfee` modifier will changes is to 100% of the amount if the value is below the exemption threshold for example).

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+			    struct payment_modifier **mods)
+{
+	struct payment *p = tal(ctx, struct payment);
+	static u32 partid = 1;
+	p->children = tal_arr(p, struct payment *, 0);
+	p->parent = parent;
+	p->modifiers = mods;
+	p->cmd = cmd;
+	p->start_time = time_now();
+	p->partid = partid++;
```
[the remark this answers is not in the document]
cdecker: Good catch, removed old `static` part ID :+1:

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+			    p->plugin,
+			    "Number overflow summing partial payments: %s + %s",
+			    type_to_string(tmpctx, struct amount_msat,
+					   &res.sent),
+			    type_to_string(tmpctx, struct amount_msat,
+					   &cres.sent));
+
+		/* Bubble up the first preimage we see. */
+		if (res.preimage == NULL && cres.preimage != NULL)
+			res.preimage = cres.preimage;
```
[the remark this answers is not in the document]
cdecker: Definitely, however to maintain the current interface I had to come up with some way of subsuming the tree-result in a single preimage, which for now works, and if it doesn't anymore we'll just remove it from the result.

The `payment_tree_result` is just for display purposes mostly.

plugins/libplugin-pay.c
```diff
@@ -0,0 +1,780 @@
...
+static u8 *tal_towire_legacy_payload(const tal_t *ctx, const struct legacy_payload *payload)
+{
+	const u8 padding[] = {0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+			      0x00, 0x00, 0x00, 0x00, 0x00, 0x00};
+	/* Prepend 0 byte for realm */
+	u8 *buf = tal_arrz(ctx, u8, 1);
+	towire_short_channel_id(&buf, &payload->scid);
+	towire_u64(&buf, payload->forward_amt.millisatoshis); /* Raw: low-level serializer */
+	towire_u32(&buf, payload->outgoing_cltv);
+	towire(&buf, padding, ARRAY_SIZE(padding));
```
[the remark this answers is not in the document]
cdecker: They happen to coincide in this case, the code is a copy-paste from the `pay` plugin anyway, and is going to be deprecated eventually.

plugins/libplugin-pay.c
cdecker: Ouch, good catch :+1:

plugins/libplugin-pay.c
cdecker: The rationale behind this is that we want to show success as soon as possible, and not have to wait for some potentially stuck HTLCs for hours. In the non-MPP case the two considerations are identical, but for MPP we might have some in flight, while others are already settled with a preimage. I prefer showing success as soon as we have a preimage simply because it'll likely match what the recipient is seeing, i.e., the recipient has seen all pieces, and released the preimage, so from his point of view all is good.

This is also reflected in the lines below where we add a warning to that effect:

https://github.com/ElementsProject/lightning/blob/8bb9edc6dbb6c6906e93dbbf2568e40de391c8ea/plugins/libplugin-pay.c#L647-L652

plugins/libplugin-pay.c
cdecker: I added the code to generate an `enum` in the `gen_onion_wire.h` file, so we can use the names. Thanks for the suggestion ^^

ZmnSCPxj:
One thing I have been thinking of is to separate pathfinding into a separate module.  What I mean is, there is a module within `pay` whose purpose is to call `getroute`  repeatedly and cache the resulting routes, speculatively excluding channels/nodes it believes are high-risk (low capacity, historically unreliable, cannot be pinged, etc).  This module would run continuously in the background for a particular payment until it reached some number of candidate routes, then sit idle until the number of valid candidate routes drops below some threshold, after which it would trigger again and search more routes.

Periodically the pay logic would inform the pathfinding module that some node/channel failed a payment of XX amount, and the pathfinding module would rebox routes that include that channel / node into lower-amount boxes.

This allows us to amortize the time we are waiting for HTLCs to be resolved or failed into (speculative) pathfnding, and there may be more efficient pathfinding algorithms that can generate multiple paths in a single run rather than just a single path,

So roughly, the interface between the pay logic and the pathfinder module would be:

* `new_pathfinding_module(destination, routehints)` - construct a new pathfinding module for a particular target destination.
* `pathfinding_module_get(pathfinding_module, amount)` - deliver a path to the destination, for the specified amount.
* `pathfinding_module_inform_failed(pathfiding_module, amount, node_or_channel)` - tell the pathfinding module that the node or channel failed while delivering an amount.

This seems vastly different from your plan, though.

In any case, it would be possible to expose the above proposed pathfinding module as a set of commands provided by a built-in plugin (separate from `pay`).  We would need to bound its lifetime (and `pay` has a timeout parameter anyway that it could pass as the max lifetime of the pathfinding module), and allow early destruction of the pathfinding module object in case of definite success.  This would provide a relatively common framework for interacting with a plan to send funds to the destination.

plugins/libplugin-pay.c
```diff
@@ -142,6 +142,19 @@ void payment_continue(struct payment *p)
 	abort();
 }
 
+void *payment_mod_get_data(const struct payment *p,
+			   const struct payment_modifier *mod)
+{
+	for (int i = 0; p->modifiers[i] != NULL; i++)
```
rustyrussell, line 723: Minor: prefer size_t btw.

plugins/libplugin-pay.h
```diff
@@ -129,12 +132,22 @@ struct payment_modifier {
 			     void (*)(data_type, struct payment *), step_cb),  \
 	};
 
+/* The UNUSED marker is used to shut some compilers up. */
+#define REGISTER_PAYMENT_MODIFIER_HEADER(name, data_type)                      \
+	extern struct payment_modifier name##_pay_mod;                         \
+	UNUSED static inline data_type *payment_mod_##name##_get_data(         \
+	    const struct payment *p)                                           \
+	{                                                                      \
+		return (data_type *)payment_mod_get_data(p, &name##_pay_mod);  \
```
rustyrussell, line 194: cast is unnecessary here...

plugins/libplugin-pay.c
```diff
@@ -170,3 +170,39 @@ static inline void dummy_step_cb(struct dummy_data *dd,
...
+{
+	struct payment *subpayment;
+	struct retry_mod_data *rdata = payment_mod_retry_get_data(p);
+	if (p->step == PAYMENT_STEP_FAILED && rdata->retries > 0) {
+		subpayment = payment_new(p, p->cmd, p, p->modifiers);
+		payment_start(subpayment);
+		p->step = PAYMENT_STEP_RETRY;
+	}
+
+	payment_continue(p);
```
rustyrussell, line 782: I see what you mean about payment_continue() being easy to miss.

Perhaps this function should simply return, and the "payment_continue" should be in the caller?

plugins/pay.c
```diff
@@ -1704,8 +1704,9 @@ static void init(struct plugin *p,
 }
 
 #if DEVELOPER
-struct payment_modifier *paymod_mods[2] = {
+struct payment_modifier *paymod_mods[3] = {
```
rustyrussell, line 1709: explicit size not required here, BTW

plugins/libplugin-pay.c
```diff
@@ -67,28 +67,40 @@ static void payment_sendonion(struct payment *p)
...
 static struct command_result *payment_finished(struct payment *p);
 
+/* A payment is finished if a) it is in a final state, of b) it's in a
+ * child-spawning state and all of its children are in a final state. */
+static bool payment_is_finished(struct payment *p)
+{
+	bool running_children = false;
+	if (p->step == PAYMENT_STEP_FAILED || p->step == PAYMENT_STEP_SUCCESS)
+		return true;
+	else if(p->step == PAYMENT_STEP_SPLIT || p->step == PAYMENT_STEP_RETRY) {
```
rustyrussell, line 574: minor: space after 'if'.

plugins/libplugin-pay.c
```diff
@@ -67,28 +67,40 @@ static void payment_sendonion(struct payment *p)
 /* Mutual recursion. */
 static struct command_result *payment_finished(struct payment *p);
 
+/* A payment is finished if a) it is in a final state, of b) it's in a
+ * child-spawning state and all of its children are in a final state. */
+static bool payment_is_finished(struct payment *p)
+{
+	bool running_children = false;
```
rustyrussell, line 571: Would move this under the branch it is used?

plugins/libplugin-pay.c
```diff
@@ -67,28 +67,40 @@ static void payment_sendonion(struct payment *p)
 /* Mutual recursion. */
 static struct command_result *payment_finished(struct payment *p);
 
+/* A payment is finished if a) it is in a final state, of b) it's in a
+ * child-spawning state and all of its children are in a final state. */
+static bool payment_is_finished(struct payment *p)
```
rustyrussell, line 569: const struct payment *p?

plugins/libplugin-pay.c
```diff
@@ -162,7 +162,61 @@ static void payment_getroute(struct payment *p)
...
+
+	/* Now compute the payload we're about to pass to `createonion` */
+	cr = p->createonion_request = tal(p, struct createonion_request);
+	cr->assocdata = tal_arr(cr, u8, 0);
+	towire_sha256(&cr->assocdata, p->payment_hash);
+	cr->session_key = NULL;
+	cr->hops = tal_arr(cr, struct createonion_hop, tal_count(p->route));
+
+	/* Non-final hops */
+	for (int i = 0; i < hopcount - 1; i++) {
```
rustyrussell, line 457: size_t i?

plugins/libplugin-pay.c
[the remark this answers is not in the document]
rustyrussell: ARRAY_SIZE is nice because if it were to be refactored such that padding were a pointer, compilation would break...

plugins/libplugin-pay.c
```diff
@@ -178,7 +178,38 @@ static u8 *tal_towire_legacy_payload(const tal_t *ctx, const struct legacy_paylo
...
+
+	json_for_each_arr(i, cursectok, secretstok) {
+		if (cursectok->type != JSMN_STRING)
+			goto fail;
+		json_to_secret(buffer, cursectok, &resp->shared_secrets[i]);
+	}
+	return resp;
+
+fail:
+	return tal_free(resp);
```
rustyrussell, line 289: ... error messages on these failure paths?

plugins/libplugin-pay.c
```diff
@@ -188,6 +219,43 @@ static struct command_result *payment_createonion_success(struct command *cmd,
...
 
+static struct command_result *payment_createonion_success(struct command *cmd,
+							  const char *buffer,
+							  const jsmntok_t *toks,
+							  struct payment *p)
+{
+	struct out_req *req;
+	struct route_hop *first = &p->route[0];
+	struct secret *secrets;
+	p->createonion_response = tal_createonion_response_from_json(p, buffer, toks);
```
rustyrussell, line 397: .. and handle NULL response here?  crash is fine, as long as we get that msg out first!

plugins/libplugin-pay.c
```diff
@@ -209,16 +210,80 @@ tal_createonion_response_from_json(const tal_t *ctx, const char *buffer,
...
+		senttok = json_get_member(buffer, datatok, "amount_sent_msat");
+		statustok = json_get_member(buffer, datatok, "status");
+	}
+
+	/* Initial sanity checks, all these fields must exist. */
+	if (idtok == NULL || idtok->type != JSMN_PRIMITIVE ||
+	    hashtok == NULL || hashtok->type != JSMN_STRING ||
+	    senttok == NULL || senttok->type != JSMN_STRING ||
+	    statustok == NULL || statustok->type != JSMN_STRING) {
+		fprintf(stderr, "XXX missing field\n");
```
rustyrussell, line 321: ....

plugins/libplugin-pay.c
```diff
@@ -575,8 +575,12 @@ static struct command_result *payment_finished(struct payment *p)
 	assert((result.leafstates & PAYMENT_STEP_SUCCESS) == 0 ||
 	       result.preimage != NULL);
 
-	if (p->parent == NULL) {
-		assert(p->cmd != NULL);
+	if (p->parent == NULL && cmd == NULL) {
+		/* This is the tree root, but we already reported success or
+		 * failure, so noop. */
+		return command_still_pending(NULL);
```
rustyrussell, line 632: Yech, that's ugly!  I'm not convinced payment_finished should retrurn command_result...

wire/tlvstream.h
```diff
@@ -37,4 +36,19 @@ void towire_tlvs(u8 **pptr,
 /* Given any tlvstream serialize the raw fields (untyped ones). */
 void towire_tlvstream_raw(u8 **pptr, const struct tlv_field *fields);
 
+
+
+/* Generic primitive setters for tlvstreams. */
+void tlvstream_set_raw(struct tlv_field **stream, u64 type, u8 *value STEALS);
```
rustyrussell, line 42: TAKES, not STEALS (impl has this correct)

plugins/libplugin-pay.c
```diff
@@ -457,8 +471,19 @@ static void payment_compute_onion_payloads(struct payment *p)
...
+			    p->route[i + 1].amount.millisatoshis); /* Raw: TLV payload generation*/
+			tlvstream_set_tu32(&cur->tlv_payload->fields,
+					   TLV_TLV_PAYLOAD_OUTGOING_CLTV_VALUE,
+					   p->start_block +
+					       p->route[i + 1].delay);
+			tlvstream_set_short_channel_id(
+			    &cur->tlv_payload->fields,
+			    TLV_TLV_PAYLOAD_SHORT_CHANNEL_ID,
+			    &p->route[i + 1].channel_id);
 		}
```
rustyrussell: Can we generate this?  This is very unpretty... :(

plugins/libplugin-pay.c
rustyrussell: Ah, I see payment_continue gets called from other callbacks, making this more difficult.  OK.

cdecker: Yes, it's pretty similar to the `command_still_pending`/`command_fail`/`command_success` construction, so I though it'd be nice to build something similar to that. I can do that on top of `paymod-03` if you agree that it's a good idea.

plugins/pay.c
cdecker: I'll omit them going forward, it was one of the main causes of conflict while rebasing actually :-)

plugins/libplugin-pay.c
cdecker: Calling `plugin_err` so we exit with a message.

plugins/libplugin-pay.c
cdecker: Handled in the caller. Since we're parsing an internal RPC method I think we can leave the error message generic.

plugins/libplugin-pay.c
cdecker: Oops, removed in favor of logging in the caller.

plugins/libplugin-pay.c
cdecker: I could move it into a separate function for starters, not truly generated, but should clean this up nicely.

plugins/libplugin-pay.c
cdecker: The reason it is like this is that `payment_finished` encapsulates the logic to determine whether a command is pending or not. It is the place where we call `command_fail` and `command_finished`. Those two functions are marked as `warn_unused_result`, so I thought it easiest to hand them up and then silence the result.

We can of course silence it by doing something else, but I have no good ideas on how to do that.

cdecker: After some more googling it seems like we can opt-out of the `warn_unused_result` using a cast to void (???):

```c
(void)command_finished(cmd, ret);
```

Still not nice, but I added what they'd look like in the last few `fixup!`-commits (they clash with other things so I haven't reordered them just yet).

cdecker: Hm, `gcc` (7.5) doesn't seem to like this trick, and still warns about the unused result.

cdecker:
I pushed an update that attempts to fixup all the feedback. Here's the [range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3753/range_diff/dd4e95f43ef147023f380e0de5eba743f9855e8c..e581e3c8c1baa3bd5a79f7e895347cb915521425), no non-fixup commit was touched, so reviewing the fixups should be sufficient.

Not squashing just yet, since there is the "return command_result" change that needs to be acknowledged before I start working on the conflicts that reordering causes :-)

Ping @rustyrussell :-)

cdecker:
Autosquashed and rebased on top of `master`

 - [range-diff with fixups squashed](http://46.101.246.115:8888/ElementsProject/lightning/pull/3753/range_diff/dd4e95f43ef147023f380e0de5eba743f9855e8c..ae6d53e86538d8278fe7ffe4222df3a581e23b62)
 - [range-diff with fixups mapped to amended commits](http://46.101.246.115:8888/ElementsProject/lightning/pull/3753/range_diff/e581e3c8c1baa3bd5a79f7e895347cb915521425..ae6d53e86538d8278fe7ffe4222df3a581e23b62)

cdecker:
Rebase on top of `master` required by a tiny conflict in `test_pay.py`:

```
 1:  5da37043e =  1:  6cc7a3f45 plugin: Add a data backend for the new payment state-machine
 2:  7f53f548b =  2:  57a48da7d paymod: Add base state-machine for payment flow
 3:  ada0fab08 =  3:  b23b31da5 paymod: Implement base state-machine transitions
 4:  3236f4d9b !  4:  1b6a98bbd paymod: Add a simple test-command to test the paymod state-machine
    @@ -158,8 +158,8 @@
     +++ b/tests/test_pay.py
     @@
      
    -     inv = invs[0]
    -     assert(inv['msatoshi_received'] >= amt)
    +     # l1 should still be alive afterwards.
    +     assert l1.rpc.getinfo()['id'] == l1id
     +
     +
     +@unittest.skipIf(not DEVELOPER, "paymod is only available to developers for now.")
 5:  1ac7a41e0 =  5:  3b10bf487 paymod: Generate type-safe accessor functions for modifier data
 6:  18e81ad28 =  6:  ad06cd95a paymod: Add a retry modifier that retries payments on failure
 7:  e09c5ab61 =  7:  a13d42a95 paymod: Implement bubbling results up the hierarchy of payments
 8:  e73d7ab6d =  8:  faa7f6c00 paymod: Implement getroute call
 9:  b0bb7987c =  9:  b432683e3 paymod: Add a `getinfo` call on payment_start to get the blockheight
10:  4047dea37 = 10:  0b18936b9 doc: Fix a documentation error in sendonion
11:  bb37a497e = 11:  f5bf73dff paymod: Implement legacy onion payload computation
12:  bd0a1a3d7 = 12:  8e68aadc6 paymod: Call `createonion` to generate the onion and associated data
13:  b91ac6782 = 13:  388e4d468 paymod: Parse createonion result and call sendonion
14:  20b12bf22 = 14:  4cd094937 paymod: Call waitsendpay on the sendonion we just queued
15:  06adaa6e5 = 15:  4e191d569 paymod: Parse and store the waitsendpay result
16:  947340a12 = 16:  76f824981 paymod: Collect and return results of a tree of partial payments
17:  50a69ddb4 = 17:  c144d88ce paymod: Use reasonable partids
18:  0f35f8ed2 = 18:  2bc1c8bd3 paymod: Maintain a list of current and past payments
19:  fcb46ade0 = 19:  20c22daab paymod: Do not reply to the same JSON-RPC command multiple times
20:  3112bd645 = 20:  eaed425c3 paymod: Do not pass cmd to sub-payments, plugin suffices
21:  6de4eaf02 = 21:  529faefcc tlv: Add generic getters and setters for tlvstream
22:  133b0d32c = 22:  9459e2d18 wiregen: Add enums for TLV types so we can call them by their name
23:  1787df06b = 23:  ace3b9e9f paymod: Implement TLV onion payload generation
24:  ae6d53e86 = 24:  0e9839d02 paymod: Move onion payload construction into its own function
```

plugins/libplugin-pay.c
rustyrussell: No, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66425 :(

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3753/commits/3412a69488293e663d85a8727d8bfdc7ae0c9962

PR #3755, opened 2020-05-29
tools-make: also be quiet if --quiet is flagged

niftynei:
Missed a update-mocks.sh call


cdecker:
This commit was also part of #3756, so it has already been merged ^^

PR #3758, opened 2020-06-02
hsmd: remove unused sign_funding_tx.

rustyrussell:
We always treat it as a withdrawl.

Reported-by: @niftynei

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -1601,49 +1601,7 @@ static void sign_all_inputs(struct bitcoin_tx *tx, struct utxo **utxos)
 	}
 }
 
-/*~ lightningd asks us to sign the transaction to fund a channel; it feeds us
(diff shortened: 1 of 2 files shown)


hsmd/hsmd.c
```diff
@@ -1601,49 +1601,7 @@ static void sign_all_inputs(struct bitcoin_tx *tx, struct utxo **utxos)
...
-			satoshi_out, &local_pubkey, &remote_pubkey,
-			change_out, changekey,
-			NULL);
-
-	sign_all_inputs(tx, utxos);
-	return req_reply(conn, c, take(towire_hsm_sign_funding_reply(NULL, tx)));
-}
-
-/*~ lightningd asks us to sign a withdrawal; same as above but in theory
+/*~ lightningd asks us to sign a withdrawal or funding as above but in theory
```
niftynei, line 1604: technically this comment is no longer true since there is no "as above" any more

PR #3760, opened 2020-06-05
paymod: Slow but steady progress (part 02)

cdecker:
_This PR merges into the branch of the first part `paymod-01` and not
`master`. This should reduce the number of commits that are to be reviewed,
but I'm still unclear if merging the first part will re-address this PR to
point to `master`. It's an experiment, so let's see if this works for stacking
multiple PRs :wink:_

~We start by switching over to `json_paymod` if we are not running in
`COMPAT`-mode, as mentioned in the first part of this series.~ We then do a lot
of the grunt work required to get `pay` and `paystatus` return values to
include the information from our new payment flow. It's not perfect and I
decided not to go for 100% result compatibility, opting instead of guarding
the change behind the `COMPAT_V090` flag, allowing for users to test before
making the switch.

No earth shattering changes here, just incremental work, but a couple of that might be interesting to look at:

 - `local_channel_hints` modifier: fill in the `channel_hints` from
   `listpeers` so we don't try to use a channel we should know won't work
   anyway.
 - Collecting results: iterate through the attempt tree and collect the
   information reported to the caller of `pay` and `paystatus`. It's also used
   to determine whether the payment is still in progress, and triages the
   errors to bubble up the most important.

Edit: I had to disable the `COMPAT`-flag switchover, since Travis is testing with `COMPAT=0`, causing a number of tests to fail. We'll re-arm it once we have fixed all failures and fixed up the tests to match.

Changelog-None


plugins/pay.c
```diff
@@ -1504,20 +1507,35 @@ static struct command_result *json_paystatus(struct command *cmd,
 static bool attempt_ongoing(const char *b11)
 {
 	struct pay_status *ps;
+	struct payment *root;
 	struct pay_attempt *attempt;
+	struct payment_tree_result res;
+	enum payment_step diff,
+	    finalized = PAYMENT_STEP_FAILED | PAYMENT_STEP_SUCCESS;
```
niftynei, line 1514: nit: i'd call `finalized` -> `final_states`

plugins/libplugin-pay.h
```diff
@@ -164,7 +209,25 @@ struct payment {
...
+	struct node_id *excluded_nodes;
+
 	struct payment_result *result;
+
+	/* Did something happen that will cause all future attempts to fail?
+	 * This usually means that the final node reported that it can't be
+	 * reached, or in MPP payments there are no more paths we can
+	 * attempt. Modifiers need to leave failures alone once this is set to
+	 * true. Set only on the root payment. */
+	bool abort;
```
niftynei, line 226: is `abort` a keyword? it's definitely a stdlib method name. (github's syntax highlighter is showing it as a keyword)

plugins/libplugin-pay.c
```diff
@@ -768,16 +990,147 @@ retry_data_init(struct payment *p)
...
+	case WIRE_AMOUNT_BELOW_MINIMUM:
+	case WIRE_CHANNEL_DISABLED:
+	case WIRE_EXPIRY_TOO_SOON:
+	case WIRE_FEE_INSUFFICIENT:
+	case WIRE_FINAL_INCORRECT_CLTV_EXPIRY:
+	case WIRE_FINAL_INCORRECT_HTLC_AMOUNT:
+	case WIRE_INCORRECT_CLTV_EXPIRY:
+	case WIRE_INVALID_ONION_HMAC:
+	case WIRE_INVALID_ONION_KEY:
+	case WIRE_PERMANENT_CHANNEL_FAILURE:
```
niftynei, line 1030: i'm assuming 'retry the payment' means along a new route?

plugins/libplugin-pay.c
```diff
@@ -226,12 +224,35 @@ static struct command_result *payment_getroute_error(struct command *cmd,
...
 	p->step = PAYMENT_STEP_FAILED;
+	p->route = NULL;
 	payment_continue(p);
 
 	/* Let payment_finished_ handle this, so we mark it as pending */
 	return command_still_pending(cmd);
 }
 
+/* Iterate through the channel_hints and exclude any channel that we are
+ * confidet will not be able to handle this payment. */
```
niftynei, line 235:
```suggestion
 * confident will not be able to handle this payment. */
```

plugins/libplugin-pay.c
```diff
@@ -356,16 +447,89 @@ static struct command_result *
...
+	case WIRE_UNKNOWN_NEXT_PEER:
+	case WIRE_REQUIRED_CHANNEL_FEATURE_MISSING:
+		/* All of these result in the channel being marked as disabled. */
+		assert(*p->result->erring_index < tal_count(p->route));
+		hop = &p->route[*p->result->erring_index];
+		hint.enabled = false;
+		hint.scid.scid = hop->channel_id;
+		hint.scid.dir = hop->direction;
+		hint.estimated_capacity = AMOUNT_MSAT(0);
+		tal_arr_expand(&root->channel_hints, hint);
```
niftynei, line 477: is it possible this channel hint is already in the `channel_hints` list?

plugins/libplugin-pay.c
```diff
@@ -356,16 +447,89 @@ static struct command_result *
...
+		break;
+	case WIRE_TEMPORARY_CHANNEL_FAILURE:
+		/* These are an indication that the capacity was insufficient,
+		 * remember the amount we tried as an estimate. */
+		assert(*p->result->erring_index < tal_count(p->route));
+		hop = &p->route[*p->result->erring_index];
+		hint.enabled = true;
+		hint.scid.scid = hop->channel_id;
+		hint.scid.dir = hop->direction;
+		hint.estimated_capacity.millisatoshis = hop->amount.millisatoshis * 0.75; /* Raw: Multiplication */
```
niftynei, line 487: should we add a msat multiplication helper?

plugins/libplugin-pay.c
```diff
@@ -356,16 +447,89 @@ static struct command_result *
...
+	case WIRE_FEE_INSUFFICIENT:
+	case WIRE_INCORRECT_CLTV_EXPIRY:
+	case WIRE_FINAL_INCORRECT_CLTV_EXPIRY:
+		/* These are issues that are due to gossipd being out of date,
+		 * we ignore them here, and wait for gossipd to adjust
+		 * instead. */
+		break;
+	case WIRE_FINAL_INCORRECT_HTLC_AMOUNT:
+		/* These are symptoms of intermediate hops tampering with the
+		 * payment. */
```
niftynei, line 528: should we log these as unusual or info?

plugins/libplugin-pay.c
```diff
@@ -661,9 +825,86 @@ static struct command_result *payment_finished(struct payment *p)
...
-					    "Not functional yet");
+		} else if (result.failure == NULL || result.failure->failcode < NODE) {
+			/* This is failing because we have no more routes to try */
+			ret = jsonrpc_stream_fail(cmd, PAY_ROUTE_NOT_FOUND,
+						  NULL);
+			json_add_string(
+			    ret, "message",
+			    tal_fmt(cmd,
+				    "Ran out of routes to try after "
+				    "%d attempt%s: see paystatus",
```
niftynei, line 836:
```suggestion
				    "%d attempt%s: see `paystatus`",
```

plugins/libplugin-pay.c
```diff
@@ -661,9 +825,86 @@ static struct command_result *payment_finished(struct payment *p)
...
+			json_add_num(ret, "created_at", p->start_time.ts.tv_sec);
+			json_add_string(ret, "message", p->result->message);
+			json_add_node_id(ret, "destination", p->destination);
+			json_add_sha256(ret, "payment_hash", p->payment_hash);
+
+			if (result.leafstates & PAYMENT_STEP_SUCCESS) {
+				/* If one sub-payment succeeded then we have
+				 * proof of payment, and the payment is a
+				 * success. */
+				json_add_string(ret, "status", "complete");
```
niftynei, line 869: might this be confusing for MPP payments that aren't fully terminated? ie the result comes back 'complete' but we havent technically finished collecting all the parts yet. possible i'm misunderstanding what'd going on here.

plugins/libplugin-pay.c
```diff
@@ -234,6 +234,28 @@ static struct command_result *payment_getroute_error(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
+/* Iterate through the channel_hints and exclude any channel that we are
+ * confidet will not be able to handle this payment. */
+static void payment_getroute_add_excludes(struct payment *p,
+					  struct json_stream *js)
+{
+	struct payment *root = payment_root(p);
+	struct channel_hint *hint;
```
niftynei, line 240: the `channel_hint` name is a tad bit confusing since there are also *invoice* channel hints, no? i don't have a good alternative suggestion though.

plugins/pay.c
```diff
@@ -1269,6 +1269,9 @@ static struct pay_status *add_pay_status(struct pay_command *pc,
 	return ps;
 }
 
+#ifndef COMPAT_V090
+UNUSED
```
niftynei, line 1273: this seems like a no-op? am i missing something?

plugins/libplugin-pay.c
cdecker: Yes, retry the entire flow with a new sub-payment, and with updated channel hints which can then lead to the failing channel being excluded :+1:

plugins/libplugin-pay.h
cdecker: Indeed, it seems to be a function name defined in `stdlib.h`. Not sure what the downsides of re-using the name, but I'll look for something else.

plugins/libplugin-pay.c
cdecker: Indeed it is, if we have two successive attempts with smaller amounts (requires MPP) we could end up adding a new one. We'd exclude due to the smaller one, but adding it twice is suboptimal. Will change this into an `update_or_create`.

plugins/libplugin-pay.c
cdecker: Hm, the `amount_msat` interfaces annoyed me quite a couple of times tbh, especially when updating channel hints, where we can end up with negative values. So I didn't spend too much time on the `amount_msat_*` ops, but we can do that.

plugins/libplugin-pay.c
cdecker: We could split them out, but we need to react to it automatically anyway.

plugins/libplugin-pay.c
cdecker: You're correct in assuming that this can happen with some payments in flight. However this is likely to correspond to the recipient's state much more closely than if we wait for all parts to complete.

If we received at least one preimage it means the recipient has accepted the payment. Assuming they implement MPP correctly it also indicates that all other parts got at least to the recipient. Now if we still have some pending, they are either slow, or even stuck, and that can potentially last for a long time, which is a bad user-experience imho. This is why I prefer showing it as complete as soon as we get the preimage through on at least one part, so recipient and sender agree on the state almost always, rather than the occasional wait for a stuck payment that we know will eventually clear :-)

plugins/libplugin-pay.c
cdecker: Good point, I always call the ones in the invoice `route_hint`, and the ones that are just a collection of infos we know about channels `channel_hint`.

plugins/pay.c
cdecker: It basically just marks the function as `UNUSED` iff we are not in compat mode, to get the compiler's warning about unused functions to shut up :-)

plugins/libplugin-pay.c
niftynei: right, my reason for suggesting to log them is that they might be information of interest to the node operator -- not immediately actionable but maybe something you'd want to have notes about

plugins/libplugin-pay.c
niftynei: this seems to suggest that we might add a new state to clearly delineate between "your funds have definitely arrived" and "your funds are definitely arriving"

plugins/libplugin-pay.c
```diff
@@ -828,9 +828,86 @@ static struct command_result *payment_finished(struct payment *p)
...
+						  failure->message);
+
+			json_add_u64(ret, "id", p->result->id);
+
+			json_add_u32(ret, "failcode", result.failure->failcode);
+			json_add_string(ret, "failcodename",
+					result.failure->failcodename);
+			json_add_num(ret, "code", result.failure->code);
+
+			json_object_start(ret, "data");
```
rustyrussell, line 857: That's weird.  jsonrpc_stream_fail() already starts the data obj.  So we return `"error": {"code": 7, "message": "some message", "data": { "id": 1, "failcode": 4103, "failcodename": "temporary_channel_failure", "code": ???, "data": { ...`

plugins/libplugin-pay.c
rustyrussell: Well, Christian is right in that "the recipient has acknowleged payment".  There could be an orthogonal warning ("but we're waiting for some parts to be resolved") which could also apply for the "we've given up" case (assuming that the parts still pending are not sufficient to complete payment, and so the recipient won't take them).

cdecker:
Rebased on top of paymod-01:

 - [range-diff for @rustyrussell](http://46.101.246.115:8888/ElementsProject/lightning/pull/3760/range_diff/00bbed18a3438adbeadb1fad90270d2b8674228e..020dbc4ed981b9a32f554562840682ac353e9ce8)

plugins/libplugin-pay.c
cdecker: Added a comment about this to clarify.

plugins/libplugin-pay.c
cdecker: I'll add the a field to that effect in a cleanup commit :+1:

plugins/libplugin-pay.c
cdecker: Added logging message for this case.

plugins/libplugin-pay.c
niftynei: Oh right this is "send payment" not "receive payment" side of the transaction -- ok makes sense to me. Apologies for the slow response.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3760/commits/a7540de3e2c5fb5fda637e03e52a758a65cc00d2

rustyrussell:
Needs annotations in a couple of places for direct amount access.

cdecker:
Rebased on top of `master` since #3753 has been merged, will address the missing annotations now.

cdecker:
Re-applying Rusty's ACK since the requested changes are minimal

> Ack [a7540de](https://github.com/ElementsProject/lightning/commit/a7540de3e2c5fb5fda637e03e52a758a65cc00d2)

ACK https://github.com/ElementsProject/lightning/pull/3760/commits/cf3dfa73835fee3e41fec4227d7b47dd57843b61

PR #3761, opened 2020-06-08
common/json.c: Check that JSMN result is well-formed.

ZmnSCPxj:
xref: https://lists.ozlabs.org/pipermail/c-lightning/2020-June/000188.html

### DIFF
--- a/common/json.c
+++ b/common/json.c
@@ -301,6 +301,235 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
 	return NULL;
 }
 
+/*-----------------------------------------------------------------------------
+JSMN Result Validation Starts
+-----------------------------------------------------------------------------*/
+/*~ LIBJSMN is a fast, small JSON parsing library.
+ *
+ * "Fast, small" means it does not, in fact, do a
+ * lot of checking for invalid JSON.
+ *
+ * For example, by itself it would accept the strings
+ * `{"1" "2" "3" "4"}` and `["key": 1 2 3 4]` as valid.
+ * Obviously those are not in any way valid JSON.
+ *
+ * This part of the code performs some filtering so
+ * that at least some of the invalid JSON that
+ * LIBJSMN accepts, will be rejected by
+ * json_parse_input.
+ */
+
+/*~ These functions are used in JSMN validation.
+ *
+ * The calling convention is that the "current" token
+ * is passed in as the first argument, and after the
+ * validator, is returned from the function.
+ *
+ *      p = validate_jsmn_datum(p, end, valid);
+ *
+ * The reason has to do with typical C ABIs.
+ * Usually, the first few arguments are passed in via
+ * register, and the return value is also returned
+ * via register.
+ * This calling convention generally ensures that
+ * the current token pointer `p` is always in a
+ * register and is never forced into memory by the
+ * compiler.
+ *
+ * These functions are pre-declared here as they
+ * are interrecursive.
+ * Note that despite the recursion, `p` is only ever
+ * advanced, and there is only ever one `p` value,
+ * thus the overall algorithm is strict O(n)
+ * (*not* amortized) in time.
+ * The recursion does mean the algorithm is O(d)
+ * in memory (specifically stack frames), where d
+ * is the nestedness of objects in the input.
+ * This may become an issue later if we are in a
+ * stack-limited environment, such as if we actually
+ * went and used threads.
+ */
+/* Validate a *single* datum.  */
+static const jsmntok_t *
+validate_jsmn_datum(const jsmntok_t *p,
+		    const jsmntok_t *end,
+		    bool *valid);
+/*~ Validate a key-value pair.
+ *
+ * In JSMN, objects are not dictionaries.
+ * Instead, they are a sequence of datums.
+ *
+ * In fact, objects and arrays in JSMN are "the same",
+ * they only differ in delimiter characters.
+ *
+ * Of course, in "real" JSON, an object is a dictionary
+ * of key-value pairs.
+ *
+ * So what JSMN does is that the syntax "key": "value"
+ * is considered a *single* datum, a string "key"
+ * that contains a value "value".
+ *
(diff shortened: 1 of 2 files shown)


ZmnSCPxj:
Probably deserves some kind of test though.

ZmnSCPxj:
Modified the `common/test/run-json.c`, should be OK for merging now.

common/json.c
```diff
@@ -301,6 +301,90 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
...
+			++p;
+		else
+			p = find_jsmn_bounds(p + 1, e, p->size);
+	}
+
+	return p;
+}
+
+/* Validate that the JSMN result is not weird, e.g.
+ { "" "" "" }, which JSMN will accept incorectly.
```
cdecker, line 323:
```suggestion
 { "" "" "" }, which JSMN will accept incorrectly.
```

common/json.c
```diff
@@ -301,6 +301,90 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
...
+			if (p->size != 0 && objend == p + 1)
+				return false;
+
+			pp = p + 1;
+			while (pp < objend) {
+				/* Check the key.  */
+				if (pp->type != JSMN_STRING)
+					return false;
+				if (pp->size != 1)
+					return false;
```
cdecker, line 364: Don't quite follow why this needs to be `1`, while `JSMN_STRING` instances in the top-level must have size `0` (see line 336).

common/json.c
```diff
@@ -301,6 +301,90 @@ const jsmntok_t *json_get_arr(const jsmntok_t tok[], size_t index)
...
+	}
+
+	return p;
+}
+
+/* Validate that the JSMN result is not weird, e.g.
+ { "" "" "" }, which JSMN will accept incorectly.
+*/
+static bool
+validate_jsmn_result(const jsmntok_t *toks, const jsmntok_t *toks_end)
```
cdecker, line 326: This is supposedly not only true for results but arbitrary JSON-encoded objects.

common/json.c
ZmnSCPxj: libjsmn parses the text `"foo": 2` as:

```
[0]
        type = JSMN_STRING
        start = 1
        end = 4
        size = 1
[1]
        type = JSMN_PRIMITIVE
        start = 7
        end = 8
        size = 0
```

In fact, an "object" in libjsmn is not a dictionary.  It is a *sequence* of *key-value pairs* that is delimited by `{` and `}`.  A "key-value pair" is a `JSMN_STRING` key that contains (`size == 1`) a value.

Indeed, you could write `{ "", "" }` in libjsmn and it would accept that, creating:

```
[0]
        type = JSMN_OBJECT
        start = 0
        end = 10
        size = 2
[1]
        type = JSMN_STRING
        start = 3
        end = 3
        size = 0
[2]
        type = JSMN_STRING
        start = 7
        end = 7
        size = 0
```

Thus, to differentiate between `{ "", ""}` and the slightly-more-correct `{"": 1, "": 2}`, you need to check that you are getting a sequence of key-value pairs, not just strings.

In fact libjsmn is insane enough that if you parse `{ "key": 1 2 3 }` you get a key with multiple values.

```
[0]
        type = JSMN_OBJECT
        start = 0
        end = 16
        size = 1
[1]
        type = JSMN_STRING
        start = 3
        end = 6
        size = 3  # <--------------- notice this
[2]
        type = JSMN_PRIMITIVE
        start = 9
        end = 10
        size = 0
[3]
        type = JSMN_PRIMITIVE
        start = 11
        end = 12
        size = 0
[4]
        type = JSMN_PRIMITIVE
        start = 13
        end = 14
        size = 0
```

common/json.c
ZmnSCPxj: Well it protects against certain infelicities of the JSMN parser.  There are invalid JSON that JSMN accepts; this function is intended to filter *some* of those out (for example, it will still accept objects with duplicate keys).  Presumably a more feature-complete JSON parser would outright reject those invalid JSON.

ZmnSCPxj:
> Am I correct in assuming this is `O(n)` despite the recursive calls, since we advance the pointers by the parsed amounts in the recursive calls?

Yes, it *should* be.

ZmnSCPxj:
>> Am I correct in assuming this is `O(n)` despite the recursive calls, since we advance the pointers by the parsed amounts in the recursive calls?
>
>Yes, it *should* be.

Ah, no, no.  Deeply-nested objects/arrays would trigger the function being called over sub-ranges of an earlier call, so it does not achieve O(n).  Personally I think correctness is more important, we can always rewrite it faster later.

ZmnSCPxj:
Sorry, completely rewrote the code to be O(n) on the number of tokens validated, and requiring O(d) stack space (d being nestedness of objects/arrays in the JSON input).

common/json.c
cdecker: No, I meant this check should be applied to all JSON objects we handle, whether they are requests or results. The name of the function suggests it's for results only.

ZmnSCPxj: The other objects we handle tend to be in `json_stream` or `json_out` forms, which have their own issues, but are *not* issues *this* function can detect.  This function handles LIBJSMN infelicities, specifically the `jsmn_parse` weirdnesses; other issues in `json_stream` / `json_out` are not handled by this function.  So it makes no sense to apply this to other JSON objects; it should only apply to the result of `jsmn_parse`, because that is *all* it actually checks.

In theory, if we had a decent library instead of a small fast one we would not need this function.  It is strictly due to the use of LIBJSMN, and it is the output result of LIBJSMN that it checks.

How about `validate_jsmn_parse_output` instead?

cdecker: That'd be perfect, and you're right, we mostly use jsmn for results.

ZmnSCPxj:
Renamed the validation function to `validate_jsmn_parse_output`.

rustyrussell:
Nice work.  Yes, I hit a similar issue previously, but that one was fixed upstream.  Good to validate generally for this reason.

PR #3762, opened 2020-06-10
PSBTs, Meet the HSM

niftynei:
Updates the HSM withdrawal to take a PSBT and utxo set. We sign the inputs that we've got utxos for, and return a psbt with the attached signatures.

This 'fully' integrates the PSBTs with our HSM for the first time, and gets us ready for a `signpsbt` command (coming soon).

Note that this is built atop #3740 
(new content begins at 
965014d), and requires https://github.com/ElementsProject/libwally-core/pull/203 to be merged in order to pass tests (as there's a number of P2SH-P2WPKH bugs in the existing libwally-psbt implementation)

Changelog-None

### DIFF
--- a/bitcoin/psbt.c
+++ b/bitcoin/psbt.c
@@ -5,7 +5,9 @@
 #include <bitcoin/signature.h>
 #include <ccan/cast/cast.h>
 #include <ccan/ccan/array_size/array_size.h>
+#include <ccan/tal/str/str.h>
 #include <common/amount.h>
+#include <common/type_to_string.h>
 #include <common/utils.h>
 #include <string.h>
 #include <wally_psbt.h>
@@ -309,6 +311,28 @@ struct amount_sat psbt_input_get_amount(struct wally_psbt *psbt,
 
 	return val;
 }
+
+bool psbt_from_b64(const char *b64str, struct wally_psbt **psbt)
+{
+	int wally_err;
+	wally_err = wally_psbt_from_base64(b64str, psbt);
+	return wally_err == WALLY_OK;
+}
+
+char *psbt_to_b64(const tal_t *ctx, const struct wally_psbt *psbt)
+{
+	char *serialized_psbt, *ret_val;
+	int ret;
+
+	ret = wally_psbt_to_base64(cast_const(struct wally_psbt *, psbt),
+				   &serialized_psbt);
+	assert(ret == WALLY_OK);
+
+	ret_val = tal_strdup(ctx, serialized_psbt);
+	wally_free_string(serialized_psbt);
+	return ret_val;
+}
+REGISTER_TYPE_TO_STRING(wally_psbt, psbt_to_b64);
 
 const u8 *psbt_get_bytes(const tal_t *ctx, const struct wally_psbt *psbt,
 			 size_t *bytes_written)
--- a/bitcoin/psbt.c
+++ b/bitcoin/psbt.c
@@ -116,10 +116,32 @@ struct wally_psbt_input *psbt_add_input(struct wally_psbt *psbt,
 {
 	struct wally_tx *tx;
 	struct wally_tx_input tmp_in;
+	u8 *script;
+	size_t scriptlen = 0;
+	struct wally_tx_witness_stack *witness = NULL;
 
 	tx = psbt->tx;
 	assert(insert_at <= tx->num_inputs);
+
+	/* Remove any script sig or witness info before adding it ! */
+	if (input->script_len > 0) {
+		scriptlen = input->script_len;
+		input->script_len = 0;
+		script = (u8 *)input->script;
+		input->script = NULL;
+	}
+	if (input->witness) {
+		witness = input->witness;
+		input->witness = NULL;
+	}
 	wally_tx_add_input(tx, input);
+	/* Put the script + witness info back */
+	if (scriptlen > 0) {
+		input->script_len = scriptlen;
+		input->script = script;
+	}
+	if (witness)
+		input->witness = witness;
 
 	tmp_in = tx->inputs[tx->num_inputs - 1];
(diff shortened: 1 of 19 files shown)


bitcoin/tx.c
```diff
@@ -486,15 +486,7 @@ void bitcoin_tx_finalize(struct bitcoin_tx *tx)
...
-	char *serialized_psbt, *ret_val;
-	int ret;
-
-	ret = wally_psbt_to_base64(tx->psbt, &serialized_psbt);
-	assert(ret == WALLY_OK);
-
-	ret_val = tal_strdup(ctx, serialized_psbt);
-	wally_free_string(serialized_psbt);
-	return ret_val;
+	return psbt_to_b64(ctx, tx->psbt);
```
cdecker: This is now a pure alias, we could just `#define` it, and save the call.

bitcoin/psbt.c
```diff
@@ -116,15 +116,39 @@ struct wally_psbt_input *psbt_add_input(struct wally_psbt *psbt,
...
+	}
+	if (witness)
+		input->witness = witness;
 
 	tmp_in = tx->inputs[tx->num_inputs - 1];
 	MAKE_ROOM(tx->inputs, insert_at, tx->num_inputs);
+
 	tx->inputs[insert_at] = tmp_in;
 
+
```
cdecker, line 151: Whitespace only change, two consecutive whitespaces.

bitcoin/psbt.h
```diff
@@ -49,7 +49,7 @@ void psbt_rm_output(struct wally_psbt *psbt,
 void psbt_input_add_pubkey(struct wally_psbt *psbt, size_t in,
 			   const struct pubkey *pubkey);
 
-void psbt_input_set_partial_sig(struct wally_psbt *psbt, size_t in,
+bool psbt_input_set_partial_sig(struct wally_psbt *psbt, size_t in,
```
cdecker, line 54: Should likely add the `WARN_UNUSED_RESULT` flag to be sure we don't accidentally fail, but believe it worked.

bitcoin/psbt.h
```diff
@@ -32,6 +32,8 @@ struct wally_psbt *new_psbt(const tal_t *ctx,
  */
 bool psbt_is_finalized(struct wally_psbt *psbt);
 
+struct wally_tx *psbt_finalize(const struct wally_psbt *psbt, bool finalize_in_place);
```
cdecker, line 35: Why is the `psbt` argument declared `const` if we can modify it at all in-place? Shouldn't we rather communicate the non-constness to the callers, rather than working around it being a const here?

Could also be that my mental model of `const` is wrong :-)

hsmd/hsmd.c
```diff
@@ -1512,87 +1512,63 @@ static void hsm_key_for_utxo(struct privkey *privkey, struct pubkey *pubkey,
...
+			 * and be done with it. We can't do that though
+			 * because any UTXO that's derived from channel_info
+			 * requires the HSM to find the pubkey, and we
+			 * skip doing that until now as a bit of a reduction
+			 * of complexity in the calling code */
+			psbt_input_add_pubkey(psbt, j, &pubkey);
+
+			if (wally_sign_psbt(psbt, privkey.secret.data,
+					    sizeof(privkey.secret.data)) != WALLY_OK)
+				status_broken("Recieved wally_err attempting to "
```
cdecker, line 1544:
```suggestion
				status_broken("Received wally_err attempting to "
```

hsmd/hsmd.c
```diff
@@ -1512,87 +1512,63 @@ static void hsm_key_for_utxo(struct privkey *privkey, struct pubkey *pubkey,
...
-		struct pubkey inkey;
-		struct bitcoin_signature sig;
+		struct utxo *utxo = utxos[i];
+		for (size_t j = 0; j < psbt->num_inputs; j++) {
+			struct privkey privkey;
+			struct pubkey pubkey;
+
+			if (!wally_tx_input_spends(&psbt->tx->inputs[j],
+						   &utxo->txid, utxo->outnum))
+				continue;
```
cdecker, line 1527: Am I correct in assuming this is the case where the UTXO is not under the control of the node? I.e., this output doesn't belong to us and hence we skip it. So this is the case that is alluded to in the commit message?

So this `continue` has two reasons: either we are pairing up the UTXO with the wrong input, or the caller didn't provide the UTXO because it is not under the control of this node.

niftynei: Yes your assumption is correct. We only have `utxo` data for internal-wallet utxos -- which is where these come from.

This check is looking for our inputs on this transaction. Note that we could skip the check/iteration and just call 'straight' `psbt_sign` except that we have to add the public key information to inputs (we could do this elsewhere except for inputs that originated from a close transaction).

niftynei:
>  Travis doesn't really seem to be happy with this PR, other than that I have only minor comments.

Travis is unhappy because we require ElementsProject/libwally-core#203 to be merged -- there's a number of P2SH-P2WPKH bugs in the existing libwally-psbt implementation, that we now hit.

bitcoin/tx.c
niftynei: This method actually gets removed later :)

bitcoin/psbt.h
niftynei: Oh good point. I guess I meant it more as a "hey you better really be done with this PSBT" marker, but as you point out, since the `finalize in place` option actually updates it internally that's not really *enforceable*.

niftynei: I'll remove it, I don't think the usage corresponds to the intention of `const`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3762/commits/37811fd35a30c01cc31cabb0d76612e1c0b12246

PR #3763, opened 2020-06-10
Multifundchannel

ZmnSCPxj:
Fixes: #1936

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -846,6 +846,27 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("listsendpays", payload)
 
+    def multifundchannel(self, destinations, feerate=None, minconf=None, utxos=None, best_effort=False, **kwargs):
+        """
+        Fund channels to an array of {destinations},
+        each entry of which is a dict of node {id}
+        and {amount} to fund, and optionally whether
+        to {announce} and how much {push_msat} to
+        give outright to the node.
+        You may optionally specify {feerate},
+        {minconf} depth, and the {utxos} set to use
+        for the single transaction that funds all
+        the channels.
+        """
+        payload = {
+            "destinations": destinations,
+            "feerate": feerate,
+            "minconf": minconf,
+            "utxos": utxos
+        }
+        payload.update({k: v for k, v in kwargs.items()})
+        return self.call("multifundchannel", payload)
+
     def newaddr(self, addresstype=None):
         """Get a new address of type {addresstype} of the internal wallet.
         """
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -846,7 +846,7 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("listsendpays", payload)
 
-    def multifundchannel(self, destinations, feerate=None, minconf=None, utxos=None, best_effort=False, **kwargs):
+    def multifundchannel(self, destinations, feerate=None, minconf=None, utxos=None, minchannels=None, **kwargs):
         """
         Fund channels to an array of {destinations},
         each entry of which is a dict of node {id}
@@ -862,7 +862,8 @@ class LightningRpc(UnixDomainSocketRpc):
             "destinations": destinations,
             "feerate": feerate,
             "minconf": minconf,
-            "utxos": utxos
+            "utxos": utxos,
+            "minchannels": minchannels,
         }
         payload.update({k: v for k, v in kwargs.items()})
         return self.call("multifundchannel", payload)
--- a/doc/lightning-multifundchannel.7.md
+++ b/doc/lightning-multifundchannel.7.md
@@ -4,7 +4,7 @@ lightning-multifundchannel -- Command for establishing many lightning channels
 SYNOPSIS
 --------
 
-**multifundchannel** *destinations* \[*feerate*\] \[*minconf*\] \[*utxos*\]
+**multifundchannel** *destinations* \[*feerate*\] \[*minconf*\] \[*utxos*\] \[*minchannels*\]
 
 DESCRIPTION
 -----------
@@ -68,6 +68,11 @@ outputs should have. Default is 1.
 *utxos* specifies the utxos to be used to fund the channel, as an array
 of "txid:vout".
 
+*minchannels*, if specified, will re-attempt funding as long as at least
+this many peers remain (must not be zero).
+The **multifundchannel** command will only fail if too many peers fail
+the funding process.
+
 RETURN VALUE
 ------------
 
@@ -76,12 +81,24 @@ transaction is returned.
 This command opens multiple channels with a single large transaction,
 thus only one transaction is returned.
 
+If *minchannels* was specified and is less than the number of destinations,
+then it is possible that one or more of the destinations
+do not have a channel even if **multifundchannel** succeeded.
+
 An array of *channel\_ids* is returned;
 each entry of the array is an object,
  with an *id* field being the node ID of the peer,
  an *outnum* field being the output number of the transaction
   that anchors this channel,
  and *channel_id* field being the channel ID with that peer.
+
+An array of *failed* is returned,
+which contains the destinations that were removed
+due to failures (this can only happen on success if *minchannels* was specified).
+Each entry of the array is an object,
+ with an *id* field being the node ID of the removed peer,
+ *method* field describing what phase of funding the peer failed,
+ and *error* field of the exact error returned by the method.
 
 On failure, none of the channels are created.
 
--- a/doc/lightning-multifundchannel.7.md
+++ b/doc/lightning-multifundchannel.7.md
@@ -1,0 +1,127 @@
+lightning-multifundchannel -- Command for establishing many lightning channels
+==============================================================================
+
+SYNOPSIS
+--------
+
+**multifundchannel** *destinations* \[*feerate*\] \[*minconf*\] \[*utxos*\]
+
+DESCRIPTION
+-----------
+
+The **multifundchannel** RPC command opens multiple payment channels
+with nodes by committing a single funding transaction to the blockchain
+that is shared by all channels.
+
+If not already connected, **multifundchannel** will automatically attempt
+to connect; you may provide a *@host:port* hint appended to the node ID
+so that c-lightning can learn how to connect to the node;
+see lightning-connect(7).
+
+Once the transaction is confirmed, normal channel operations may begin.
+Readiness is indicated by **listpeers** reporting a *state* of
+`CHANNELD_NORMAL` for the channel.
+
+*destinations* is an array of objects, with the fields:
+
+* *id* is the node ID, with an optional *@host:port* appended to it
+  in a manner understood by **connect**; see lightning-connect(7).
+  Each entry in the *destinations* array must have a unique node *id*.
+* *amount* is the amount in satoshis taken from the internal wallet
+  to fund the channel.
+  The string *all* can be used to specify all available funds
+  (or 16,777,215 satoshi if more is available and large channels were
+  not negotiated with the peer).
+  Otherwise it is in satoshi precision; it can be
+   a whole number,
+   a whole number ending in *sat*,
+   a whole number ending in *000msat*, or
+   a number with 1 to 8 decimal places ending in *btc*.
+  The value cannot be less than the dust limit, currently 546 satoshi
+  as of this writing, nor more than 16,777,215 satoshi
+  (unless large channels were negotiated with the peer).
+* *announce* is an optional flag that indicates whether to announce
+  the channel with this, default `true`.
+  If set to `false`, the channel is unpublished.
+* *push\_msat* is the amount of millisatoshis to outright give to the
+  node.
+  This is a gift to the peer, and you do not get a proof-of-payment
+  out of this.
+
+There must be at least one entry in *destinations*;
+it cannot be an empty array.
+
+*feerate* is an optional feerate used for the opening transaction and as
+initial feerate for commitment and HTLC transactions. It can be one of
+the strings *urgent* (aim for next block), *normal* (next 4 blocks or
+so) or *slow* (next 100 blocks or so) to use lightningd’s internal
+estimates: *normal* is the default.
+
+Otherwise, *feerate* is a number, with an optional suffix: *perkw* means
+the number is interpreted as satoshi-per-kilosipa (weight), and *perkb*
+means it is interpreted bitcoind-style as satoshi-per-kilobyte. Omitting
+the suffix is equivalent to *perkb*.
+
+*minconf* specifies the minimum number of confirmations that used
+outputs should have. Default is 1.
+
+*utxos* specifies the utxos to be used to fund the channel, as an array
+of "txid:vout".
+
+RETURN VALUE
+------------
+
+On success, the *tx* and *txid* of the signed and broadcsted funding
+transaction is returned.
+This command opens multiple channels with a single large transaction,
+thus only one transaction is returned.
+
+An array of *channel\_ids* is returned;
+each entry of the array is an object,
+ with an *id* field being the node ID of the peer,
+ an *outnum* field being the output number of the transaction
+  that anchors this channel,
+ and *channel_id* field being the channel ID with that peer.
+
+On failure, none of the channels are created.
+
+The following error codes may occur:
+* -1: Catchall nonspecific error.
+- 300: The maximum allowed funding amount is exceeded.
+- 301: There are not enough funds in the internal wallet (including fees) to create the transaction.
+- 302: The output amount is too small, and would be considered dust.
+- 303: Broadcasting of the funding transaction failed, the internal call to bitcoin-cli returned with an error.
+
+Failure may also occur if **lightningd** and the peer cannot agree on
+channel parameters (funding limits, channel reserves, fees, etc.).
+See lightning-fundchannel\_start(7) and lightning-fundchannel\_complete(7).
+
+There may be rare edge cases where a communications failure later in
+the channel funding process will cancel the funding locally, but
+the peer thinks the channel is already waiting for funding lockin.
+In that case, the next time we connect to the peer, our node will
+tell the peer to forget the channel, but some nodes (in particular,
+c-lightning nodes) will disconnect when our node tells them to
+forget the channel.
+If you immediately **multifundchannel** with that peer, it could
+trigger this connect-forget-disconnect behavior, causing the
+second **multifundchannel** to fail as well due to disconnection.
+Doing a **connect** with the peers separately, and waiting for a
+few seconds, should help clear this hurdle;
+running **multifundchannel** a third time would also clear this.
+
+AUTHOR
+------
+
+ZmnSCPxj <<ZmnSCPxj@protonmail.com>> is mainly responsible.
+
+SEE ALSO
+--------
+
+lightning-connect(7), lightning-listfunds(), lightning-listpeers(7),
+lightning-fundchannel(7)
+
+RESOURCES
+---------
+
+Main web site: <https://github.com/ElementsProject/lightning>
--- a/plugins/multifundchannel.c
+++ b/plugins/multifundchannel.c
@@ -1,0 +1,1933 @@
+#include <assert.h>
+#include <bitcoin/chainparams.h>
+#include <bitcoin/psbt.h>
+#include <bitcoin/script.h>
+#include <bitcoin/tx.h>
+#include <ccan/array_size/array_size.h>
+#include <ccan/compiler/compiler.h>
+#include <ccan/crypto/sha256/sha256.h>
+#include <ccan/str/str.h>
+#include <ccan/take/take.h>
+#include <common/addr.h>
+#include <common/amount.h>
+#include <common/features.h>
+#include <common/json.h>
+#include <common/json_stream.h>
+#include <common/json_tok.h>
+#include <common/jsonrpc_errors.h>
+#include <common/node_id.h>
+#include <common/pseudorand.h>
+#include <common/type_to_string.h>
+#include <common/utils.h>
+#include <plugins/libplugin.h>
+#include <stdarg.h>
+#include <string.h>
+
+/* Current state of the funding process.  */
+enum multifundchannel_state {
+	/* We have not yet performed `fundchannel_start`.  */
+	MULTIFUNDCHANNEL_START_NOT_YET = 0,
+	/* The `connect` command failed.  `*/
+	MULTIFUNDCHANNEL_CONNECT_FAILED,
+	/* The `fundchannel_start` command succeeded.  */
+	MULTIFUNDCHANNEL_STARTED,
+	/* The `fundchannel_start` command failed.  */
+	MULTIFUNDCHANNEL_START_FAILED,
+	/* The `fundchannel_complete` command failed.  */
+	MULTIFUNDCHANNEL_COMPLETE_FAILED,
+	/* The transaction might now be broadcasted.  */
+	MULTIFUNDCHANNEL_DONE
+};
+
+/* The object for a single destination.  */
+struct multifundchannel_destination {
+	/* The overall multifundchannel command object.  */
+	struct multifundchannel_command *mfc;
+
+	/* The overall multifundchannel_command contains an
+	array of multifundchannel_destinations.
+	This provides the index within the array.
+
+	This is used in debug printing.
+	*/
+	unsigned int index;
+
+	/* ID for this destination.  */
+	struct node_id id;
+	/* Address hint for this destination, NULL if not
+	specified.
+	*/
+	const char *addrhint;
+	/* The features this destination has.  */
+	const u8 *their_features;
+
+	/* Whether we have `fundchannel_start`, failed `connect` or
+	`fundchannel_complete`, etc.
+	*/
+	enum multifundchannel_state state;
+
+	/* The actual target script and address.  */
+	const u8 *funding_script;
+	const char *funding_addr;
+
+	/* The amount to be funded for this destination.
+	If the specified amount is "all" then the `all`
+	flag is set, and the amount is initially 0 until
+	we have figured out how much exactly "all" is,
+	after the dryrun stage.
+	*/
+	bool all;
+	struct amount_sat amount;
+
+	/* The output index for this destination.  */
+	unsigned int outnum;
+
+	/* Whether the channel to this destination will
+	be announced.
+	*/
+	bool announce;
+	/* How much of the initial funding to push to
+	the destination.
+	*/
+	struct amount_msat push_msat;
+
+	/* The actual channel_id.  */
+	const char *channel_id;
+
+	/* Any error messages.  */
+	const char *error;
+	errcode_t code;
+};
+
+/* The object for a single multifundchannel command.  */
+struct multifundchannel_command {
+	/* A unique numeric identifier for this particular
+	multifundchannel execution.
+
+	This is used for debug logs; we want to be able to
+	identify *which* multifundchannel is being described
+	in the debug logs, especially if the user runs
+	multiple `multifundchannel` commands in parallel, or
+	in very close sequence, which might confuse us with
+	*which* debug message belongs with *which* command.
+
+	We actually just reuse the id from the cmd.
+	Store it here for easier access.
+	*/
+	u64 id;
+
+	/* The plugin-level command.  */
+	struct command *cmd;
+	/* An array of destinations.  */
+	struct multifundchannel_destination *destinations;
+	/* Number of pending parallel fundchannel_start or
+	fundchannel_complete.
+	*/
+	size_t pending;
+
+	/* The feerate desired by the user.  */
+	const char *feerate_str;
+	/* The minimum number of confirmations for owned
+	UTXOs to be selected.
+	*/
+	u32 minconf;
+	/* The set of utxos to be used.  */
+	const char *utxos_str;
+
+	/* The PSBT of the funding transaction we are building.
+	Prior to `fundchannel_start` completing for all destinations,
+	this contains an unsigned incomplete transaction that is just a
+	reservation of the inputs.
+	After `fundchannel_start`, this contains an unsigned transaction
+	with complete outputs.
+	After `fundchannel_complete`, this contains a signed, finalized
+	transaction.
+	*/
+	struct wally_psbt *psbt;
+	/* The actual feerate of the PSBT.  */
+	u32 feerate_per_kw;
+	/* The expected weight of the PSBT after adding in all the outputs.
+	 * In weight units (sipa).  */
+	u32 estimated_final_weight;
+	/* Excess satoshi from the PSBT.
+	 * If "all" this is the entire amount; if not "all" this is the
+	 * proposed change amount, which if dusty should be donated to
+	 * the miners.
+	 */
+	struct amount_sat excess_sat;
+
+	/* A convenient change address. NULL at the start, filled in
+	 * if we detect we need it.  */
+	const u8 *change_scriptpubkey;
+	/* Whether we need a change output.  */
+	bool change_needed;
+	/* The change amount.  */
+	struct amount_sat change_amount;
+
+	/* The txid of the final funding transaction.  */
+	struct bitcoin_txid *txid;
+
+	/* The actual tx of the actual final funding transaction
+	that was broadcast.
+	*/
+	const char *final_tx;
+	const char *final_txid;
+};
+
+extern const struct chainparams *chainparams;
+
+/* Flag set when any of the destinations has a value of "all".  */
+static bool has_all(const struct multifundchannel_command *mfc)
+{
+	for (size_t i = 0; i < tal_count(mfc->destinations); i++) {
+		if (mfc->destinations[i].all)
+			return true;
+	}
+	return false;
+}
+
+/*-----------------------------------------------------------------------------
+Command Cleanup
+-----------------------------------------------------------------------------*/
+
+/*~
+We disallow the use of command_fail and forward_error directly
+in the rest of the code.
+
+This ensures that if we ever fail a multifundchannel, we do cleanup
+by doing fundchannel_cancel and unreserveinputs.
+*/
+
+/* TODO: This is lengthy enough to deserve its own source file,
+clocking in at 240 loc.
+*/
+
+/* Object for performing cleanup.  */
+struct multifundchannel_cleanup {
+	size_t pending;
+	struct command_result *(*cb)(void *arg);
+	void *arg;
+};
+
+/* Cleans up a PSBT.  */
+static void
+mfc_cleanup_psbt(struct command *cmd,
+		 struct multifundchannel_cleanup *cleanup,
+		 struct wally_psbt *psbt);
+/* Cleans up a `fundchannel_start`ed node id.  */
+static void
+mfc_cleanup_fc(struct command *cmd,
+	       struct multifundchannel_cleanup *cleanup,
+	       struct multifundchannel_destination *dest);
+/* Run at completion of all cleanup tasks.  */
+static struct command_result *
+mfc_cleanup_complete(struct multifundchannel_cleanup *cleanup);
+
+/* Core cleanup function.  */
+static struct command_result *
+mfc_cleanup_(struct multifundchannel_command *mfc,
+	     struct command_result *(*cb)(void *arg),
+	     void *arg)
+{
+	struct multifundchannel_cleanup *cleanup;
+	unsigned int i;
+
+	plugin_log(mfc->cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": cleanup!", mfc->id);
+
+	cleanup = tal(mfc, struct multifundchannel_cleanup);
+	cleanup->pending = 0;
+	cleanup->cb = cb;
+	cleanup->arg = arg;
+
+	if (mfc->psbt) {
+		plugin_log(mfc->cmd->plugin, LOG_DBG,
+			   "mfc %"PRIu64": unreserveinputs task.", mfc->id);
+
+		++cleanup->pending;
+		mfc_cleanup_psbt(mfc->cmd, cleanup, mfc->psbt);
+	}
+	for (i = 0; i < tal_count(mfc->destinations); ++i) {
+		struct multifundchannel_destination *dest;
+		dest = &mfc->destinations[i];
+
+		/* If not started, nothing to clean up.  */
+		if (dest->state != MULTIFUNDCHANNEL_STARTED)
+			continue;
+
+		plugin_log(mfc->cmd->plugin, LOG_DBG,
+			   "mfc %"PRIu64", dest %u: "
+			   "fundchannel_cancel task.",
+			   mfc->id, dest->index);
+
+		++cleanup->pending;
+		mfc_cleanup_fc(mfc->cmd, cleanup, dest);
+	}
+
+	if (cleanup->pending == 0)
+		return mfc_cleanup_complete(cleanup);
+	else
+		return command_still_pending(mfc->cmd);
+}
+static struct command_result *
+mfc_cleanup_done(struct command *cmd,
+		       const char *buf UNUSED,
+		       const jsmntok_t *res UNUSED,
+		       struct multifundchannel_cleanup *cleanup)
+{
+	--cleanup->pending;
+	if (cleanup->pending == 0)
+		return mfc_cleanup_complete(cleanup);
+	else
+		return command_still_pending(cmd);
+}
+/* Cleans up a txid by doing `txdiscard` on it.  */
+static void
+mfc_cleanup_psbt(struct command *cmd,
+		 struct multifundchannel_cleanup *cleanup,
+		 struct wally_psbt *psbt)
+{
+	struct out_req *req = jsonrpc_request_start(cmd->plugin,
+						    cmd,
+						    "unreserveinputs",
+						    &mfc_cleanup_done,
+						    &mfc_cleanup_done,
+						    cleanup);
+	json_add_psbt(req->js, "psbt", psbt);
+	send_outreq(cmd->plugin, req);
+}
+/* Cleans up a `fundchannel_start` by doing `fundchannel_cancel` on
+the node.
+*/
+static void
+mfc_cleanup_fc(struct command *cmd,
+	       struct multifundchannel_cleanup *cleanup,
+	       struct multifundchannel_destination *dest)
+{
+	struct out_req *req = jsonrpc_request_start(cmd->plugin,
+						    cmd,
+						    "fundchannel_cancel",
+						    &mfc_cleanup_done,
+						    &mfc_cleanup_done,
+						    cleanup);
+	json_add_node_id(req->js, "id", &dest->id);
+
+	send_outreq(cmd->plugin, req);
+}
+/* Done when all cleanup operations have completed.  */
+static struct command_result *
+mfc_cleanup_complete(struct multifundchannel_cleanup *cleanup)
+{
+	tal_steal(tmpctx, cleanup);
+	return cleanup->cb(cleanup->arg);
+}
+#define mfc_cleanup(mfc, cb, arg) \
+	mfc_cleanup_(mfc, typesafe_cb(struct command_result *, void *, \
+				      (cb), (arg)), \
+		     (arg))
+
+/* Use this instead of command_fail.  */
+static struct command_result *
+mfc_fail(struct multifundchannel_command *, errcode_t code,
+	 const char *fmt, ...);
+/* Use this instead of forward_error.  */
+static struct command_result *
+mfc_forward_error(struct command *cmd,
+		  const char *buf, const jsmntok_t *error,
+		  struct multifundchannel_command *);
+/* Use this instead of command_finished.  */
+static struct command_result *
+mfc_finished(struct multifundchannel_command *, struct json_stream *response);
+/* Use this instead of command_err_raw.  */
+static struct command_result *
+mfc_err_raw(struct multifundchannel_command *, const char *json_string);
+
+/*---------------------------------------------------------------------------*/
+
+/* These are the actual implementations of the cleanup entry functions.  */
+
+struct mfc_fail_object {
+	struct multifundchannel_command *mfc;
+	struct command *cmd;
+	errcode_t code;
+	const char *msg;
+};
+static struct command_result *
+mfc_fail_complete(struct mfc_fail_object *obj);
+static struct command_result *
+mfc_fail(struct multifundchannel_command *mfc, errcode_t code,
+	 const char *fmt, ...)
+{
+	struct mfc_fail_object *obj;
+	const char *msg;
+	va_list ap;
+
+	va_start(ap, fmt);
+	msg = tal_vfmt(mfc, fmt, ap);
+	va_end(ap);
+
+	obj = tal(mfc, struct mfc_fail_object);
+	obj->mfc = mfc;
+	obj->cmd = mfc->cmd;
+	obj->code = code;
+	obj->msg = msg;
+
+	return mfc_cleanup(mfc, &mfc_fail_complete, obj);
+}
+static struct command_result *
+mfc_fail_complete(struct mfc_fail_object *obj)
+{
+	plugin_log(obj->mfc->cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": cleanup done, failing.", obj->mfc->id);
+	return command_fail(obj->cmd, obj->code, "%s", obj->msg);
+}
+
+struct mfc_err_raw_object {
+	struct multifundchannel_command *mfc;
+	const char *error;
+};
+static struct command_result *
+mfc_err_raw_complete(struct mfc_err_raw_object *obj);
+static struct command_result *
+mfc_err_raw(struct multifundchannel_command *mfc, const char *json_string)
+{
+	struct mfc_err_raw_object *obj;
+
+	obj = tal(mfc, struct mfc_err_raw_object);
+	obj->mfc = mfc;
+	obj->error = tal_strdup(obj, json_string);
+
+	return mfc_cleanup(mfc, &mfc_err_raw_complete, obj);
+}
+static struct command_result *
+mfc_err_raw_complete(struct mfc_err_raw_object *obj)
+{
+	plugin_log(obj->mfc->cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": cleanup done, failing raw.", obj->mfc->id);
+	return command_err_raw(obj->mfc->cmd, obj->error);
+}
+static struct command_result *
+mfc_forward_error(struct command *cmd,
+		  const char *buf, const jsmntok_t *error,
+		  struct multifundchannel_command *mfc)
+{
+	plugin_log(cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": forwarding error, about to cleanup.",
+		   mfc->id);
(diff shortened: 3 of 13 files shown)


ZmnSCPxj:
Added missing changelog entries.

ZmnSCPxj:
Snuck in a revamp of the `plugins/Makefile`.

Why no reviews yet?

ZmnSCPxj:
Corrected passthrough of `utxos`.

plugins/libplugin_spark.h
```diff
@@ -0,0 +1,234 @@
...
+ * will continue; the command output will be
+ * logged at debug level.
+ * If you need to clean up memory, you should
+ * tal-allocate them from the command.
+ * If you need to clean up something more
+ * complicated, you should install a
+ * tal-destructor on the command or an object
+ * tal-allocated from the command.
+ */
+struct plugin_spark;
```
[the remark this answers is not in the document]
ZmnSCPxj: Not sure what those are, having never programmed in Golang or Java.  It is an entry in a command-level group, and once the command is finished for any reason, all sparks are cancelled.  As the plugin is running single-threaded (an implicit assumption of libplugin and many of our plugin codes) a command can only be completed if every other spark of the command is blocked on something, so all sparks are effectively just `struct out_req`s that happen to be pending right now.

plugins/multiconnect.c
```diff
@@ -0,0 +1,223 @@
...
+
+static struct command_result *
+multiconnect_done(struct command *cmd,
+		  struct connect_multi *cm)
+{
+	size_t i;
+	struct json_out *out;
+
+	out = json_out_new(NULL);
+	json_out_start(out, NULL, '{');
```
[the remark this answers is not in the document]
ZmnSCPxj: I saw `command_success` in `libplugin.h` first, and did not know we should be using `command_finished`.  Probably a cleanup of `libplugin.h` is in order at some point, to reduce what is published there, else this occassional issue will come up again.

plugins/multiconnect.c
```diff
@@ -0,0 +1,223 @@
...
+
+	/* Retrieve the `id` from the base command.
+	 * This automatically removes any @host:port
+	 * for this connection.  */
+	idtok = json_get_member(buf, result, "id");
+	if (!idtok)
+		plugin_err(cmd->plugin, "'connect' missing 'id' field");
+
+	/* Replace whatever string was originally in here.  */
+	css->id = tal_free(css->id);
```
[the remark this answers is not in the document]
ZmnSCPxj: I like it for cleanliness; future code modifications might eventually change how we use `css->id`, so it is better if we are definitely sure no pointer to the thing remains after we free it.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1643 @@
...
+static struct command_result *
+mfc_cleanup_(struct multifundchannel_command *mfc,
+	     struct command_result *(*cb)(void *arg),
+	     void *arg)
+{
+	struct multifundchannel_cleanup *cleanup;
+	unsigned int i;
+
+	plugin_log(mfc->cmd->plugin, LOG_DBG,
+		   "mfc %p: cleanup!", mfc);
```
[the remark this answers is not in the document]
ZmnSCPxj: It is useful in case somebody is running multiple such commands in parallel; that way we have a unique identifier to separate which debug log goes to which command.  Of course, given that malloc implementation will reuse memory, it is not so perfectly unique; perhaps an autoincrement id would do better.

plugins/libplugin.c
```diff
@@ -1266,3 +1309,221 @@ void plugin_main(char *argv[],
...
+plugin_wait_spark_(struct command *cmd,
+		   struct plugin_spark **pspark,
+		   struct command_result *(*cb)(struct command *cmd,
+						void *arg),
+		   void *arg)
+{
+	/* Move responsibility for the spark to this function.  */
+	struct plugin_spark *spark = *pspark;
+	*pspark = NULL;
+
```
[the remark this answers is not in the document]
ZmnSCPxj: True.  Depends on how you code, I suppose.  I like having a "single-owner" rule, which is what `tal` implicitly encodes, though I am avoiding `tal`-ness in the public interface in case in the future we need to change how sparks are stored (i.e. they may become sub-structures of a larger structure, in which case they are not necessarily `tal`-abble).

plugins/libplugin.c
```diff
@@ -1266,3 +1309,221 @@ void plugin_main(char *argv[],
...
+	adaptor = tal(cmd, struct plugin_spark_timer_adaptor);
+	adaptor->cmd = cmd;
+	adaptor->cb = cb;
+	adaptor->arg = arg;
+
+	timer = plugin_timer(cmd->plugin, time_from_sec(0),
+			     &plugin_spark_timer_adaptor_exec,
+			     adaptor);
+	tal_steal(adaptor, timer);
```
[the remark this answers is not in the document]
ZmnSCPxj: Indeed, timers seemed to be the neatest (and least-effort) solution for implementing simple concurrency; all commands share a single `io_conn`, and sparks share a command.

ZmnSCPxj:
> The handling of `all` should definitely be done based on the `listfunds` output, and doesn't warrant the use of `txprepare`. The reservation is a better reason, but still dangerous. If we want to keep the reservation I think we need to add a dummy input that makes the dummy tx uncommittable, and we must create the real transaction from scratch instead of amending the dummy tx in order to avoid leaving stale information.

Seems a bit more complicated change; would it be okay to put that in a separate PR?

plugins/multiconnect.c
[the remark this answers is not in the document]
ZmnSCPxj: Though I suppose given the next statement also assigns to the same field..... eh.  Maybe future code change will change `css`?  LOL.

plugins/multifundchannel.c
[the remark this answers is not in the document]
ZmnSCPxj: OR getting the `struct command::id`; this is unique for the plugin, right?  That might be a better tag for knowing *which* `multifundchannel` we are talking about on the logs.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1643 @@
...
+		dest->all = amount_sat_eq(*amount, AMOUNT_SAT(-1ULL));
+		dest->amount = dest->all ? AMOUNT_SAT(0) : *amount;
+		dest->announce = *announce;
+		dest->push_msat = *push_msat;
+		dest->spark = NULL;
+		dest->error = NULL;
+
+		/* Only one destination can have "all" indicator.  */
+		if (dest->all) {
+			if (*has_all)
```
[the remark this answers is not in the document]
ZmnSCPxj: Cannot -- notice the succeeding `*has_all = true`.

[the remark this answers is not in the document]
ZmnSCPxj: I decided to explciitly put an `else` branch to the `*has_all = true` to be more explicit.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1643 @@
...
+			   result->end - result->start, buf + result->start);
+	mfc->txid = tal(mfc, struct bitcoin_txid);
+	if (!bitcoin_txid_from_hex(buf + txid_tok->start,
+				   txid_tok->end - txid_tok->start,
+				   mfc->txid))
+		plugin_err(mfc->cmd->plugin,
+			   "Unable to parse 'txid' from txprepare: "
+			   "%.*s",
+			   txid_tok->end - txid_tok->start,
+			   buf + txid_tok->start);
```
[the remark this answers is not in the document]
ZmnSCPxj: Or `json_strdup` with a `tmpctx`.  But we consistently use this style in e.g. `plugins/pay.c`, `lightningd/pay.c`, and elsewhere, when reporting missing fields or incorrect types from JSON.

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,110 @@
...
+that is shared by all channels.
+
+If not already connected, **multifundchannel** will automatically attempt
+to connect; you may provide a *@host:port* hint appended to the node ID
+so that C-lightning can learn how to connect to the node;
+see lightning-connect(7).
+
+Once the transaction is confirmed, normal channel operations may begin.
+Readiness is indicated by **listpeers** reporting a *state* of
+CHANNELD\_NORMAL for the channel.
```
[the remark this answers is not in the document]
ZmnSCPxj: Copied from the `fundchannel` manpage.  Also modified the `fundchannel` manpage to match.

ZmnSCPxj:
Force-pushed requested changes, except for the transaction preparation changes which would require some more thought.

ZmnSCPxj:
> The handling of `all` should definitely be done based on the `listfunds` output, and doesn't warrant the use of `txprepare`. The reservation is a better reason, but still dangerous.

Atomicity is the main advantage of having a single command that selects inputs, computes `all`, and reserves the inputs and generates the transaction.  If the node has a lot of other RPC commands being sent to it, then by the time you compute `all` based on `listfunds` the inputs could have been reserved for another parallel command invocation.  So I think we should keep all computations of `all` in `txprepare`, because of the atomicity this guarantees you.

See https://github.com/ElementsProject/lightning/issues/3415#issuecomment-644525866.  We could have dummy transactions have a far future `nLockTime`, then have a `txmodify` which  discards the dummy transaction and re-prepare the real transaction to be broadcasted (as a single atomic step).

ZmnSCPxj:
I got some strange errors, reports from valgrind about errors arising from PSBT code that I never touched.  Rebased to before the recent PSBT changes.

cdecker:
Yes, it seems that libwally's psbt code has some minor valgrind issues. @niftynei has filed a number of PRs with them to address them.

plugins/multiconnect.h
```diff
@@ -0,0 +1,12 @@
+#ifndef LIGHTNING_PLUGINS_MULTICONNECT_H
+#define LIGHTNING_PLUGINS_MULTICONNECT_H
+#include "config.h"
```
darosior, line 3: Hmm libplugin.h already includes this ?

plugins/multiconnect.c
```diff
@@ -0,0 +1,218 @@
...
+						NULL);
+
+		if (idtok->size == 0)
+			return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
+						"Empty 'id' array: "
+						"nothing to connect.",
+						NULL);
+
+		json_for_each_arr(i, t, idtok) {
+			if (t->type != JSMN_STRING)
```
darosior, line 162: Maybe also check if the string length is `>= 32` so that we exit early in case of a mis-copy/paste

plugins/multiconnect.h
ZmnSCPxj: `config.h` is included in nearly every other header file.

plugins/multiconnect.c
ZmnSCPxj: The string can have an `@host:port` appended to the end.  We delegate parsing that to `connect`.  We *could* add the parsing for `@host:port` here but that would violate DRY unless we refactor the parsing code over in `lightningd/connect_control.c` and put it in a new common module.

ZmnSCPxj:
> Our library does already support parallel requests, so I wonder why prefering this whole new spark system over what we already do, e.g. for starting plugins.

Code reuse.  Do I want to keep rewriting the same code to synchronize multiple parallel command executions over and over?

Note also that sparks specifically allow any spark to complete the command, which stops all other suspended sparks (and since libplugin-using plugins are really single-threaded, the only way for one spark to be running and able to complete a command is if every other spark is suspended waiting on a command).  I added some extra code in `libplugin.c` to support `libplugin_spark.h` specifically so I can use `&forward_error` in a parallel execution.  Without that support code you run the risk of a use-after-free when multiple parallel command executions call `forward_error`.

ZmnSCPxj:
Bleah, the new plugins were not being installed.  Snuck in more improvements to the `Makefile` installation targets.

rustyrussell:
Actually, multiconnect is wrong, either implementation.

If one connect fails, we it seems strange to forward the error (with no information on which it was!) and ignore the successful ones.  We probably need to annotate exactly which ones failed though.

plugins/multiconnect.h
rustyrussell: Yes, it's checked in check-source, too!

rustyrussell:
OK, so I've refactored multiconnect, using @darosior 's patch as a starting point.

It's much clearer than the spark version, but I also fixed other issues:

1. We use the same flow for multi and single connect.
2. We keep addrhint explicitly, so we can display it on return.
3. A single error doesn't cause the entire call to fail.
4. The JSON output is more useful.
5. We do more of the parameter parsing in the param callback directly.  This will help when plugins finally get `check` support.
6. We don't use a gratuitous array for connect_command.
7. The init function can be NULL, so remove it.

The resulting file is only 40 lines longer, most in improved error handling:

```patch
diff --git a/plugins/multiconnect.c b/plugins/multiconnect.c
index 5f5002ad8..7fa170578 100644
--- a/plugins/multiconnect.c
+++ b/plugins/multiconnect.c
@@ -1,25 +1,32 @@
 #include "multiconnect.h"
 #include <ccan/array_size/array_size.h>
 #include <ccan/compiler/compiler.h>
+#include <ccan/tal/str/str.h>
 #include <common/json.h>
 #include <common/json_stream.h>
 #include <common/json_tok.h>
 #include <common/jsonrpc_errors.h>
 #include <common/utils.h>
-#include <plugins/libplugin_spark.h>
 
-struct connect_single_spark {
-	/* The id for this connection attempt.  */
-	const char *id;
+struct connect_single {
+	/* The id for this connection attempt (may include @... initially). */
+	char *id;
+	/* Optional address hint. */
+	char *addrhint;
+
+	/* Exactly one of these is set on completion: */
 	/* The features for this connection attempt.  */
 	const char *features;
+	/* If it failed, this is the raw JSON error. */
+	const char *error_json;
 
-	/* The completion of this spark.  */
-	struct plugin_spark_completion *completion;
+	/* The container for this */
+	struct connect_multi *cm;
 };
+
 struct connect_multi {
 	/* The individual connect_single commands.  */
-	struct connect_single_spark *subcommands;
+	struct connect_single **subcommands;
 };
 
@@ -31,25 +38,50 @@ multiconnect_done(struct command *cmd,
 
 	out = jsonrpc_stream_success(cmd);
 
-	json_array_start(out, "id");
-	for (i = 0; i < tal_count(cm->subcommands); ++i)
-		json_add_string(out, NULL, cm->subcommands[i].id);
-	json_array_end(out);
-
-	json_array_start(out, "features");
+	json_array_start(out, "peers");
 	for (i = 0; i < tal_count(cm->subcommands); ++i) {
-		json_add_string(out, NULL, cm->subcommands[i].features);
+		json_object_start(out, NULL);
+		json_add_string(out, "id", cm->subcommands[i]->id);
+		if (cm->subcommands[i]->addrhint)
+			json_add_string(out, "addresshint",
+					cm->subcommands[i]->addrhint);
+		if (cm->subcommands[i]->error_json) {
+			json_add_bool(out, "connected", false);
+			json_add_jsonstr(out, "error", cm->subcommands[i]->error_json);
+		} else {
+			json_add_bool(out, "connected", true);
+			json_add_string(out, "features", cm->subcommands[i]->features);
+		}
+		json_object_end(out);
 	}
 	json_array_end(out);
 
 	return command_finished(cmd, out);
 }
 
+static bool all_complete(struct connect_multi *cm)
+{
+       for (size_t i = 0; i < tal_count(cm->subcommands); i++)
+               if (!cm->subcommands[i]->error_json && !cm->subcommands[i]->features)
+                       return false;
+
+       return true;
+}
+
+static struct command_result *single_completed(struct command *cmd,
+					       struct connect_single *connect)
+{
+	if (all_complete(connect->cm))
+		return multiconnect_done(cmd, connect->cm);
+
+	return command_still_pending(cmd);
+}
+
-connect_single_spark_done(struct command *cmd,
-			  const char *buf,
-			  const jsmntok_t *result,
-			  struct connect_single_spark *css)
+connect_single_done(struct command *cmd,
+		    const char *buf,
+		    const jsmntok_t *result,
+		    struct connect_single *connect)
 {
 	const jsmntok_t *idtok;
 	const jsmntok_t *featurestok;
@@ -62,157 +94,161 @@ connect_single_spark_done(struct command *cmd,
 		plugin_err(cmd->plugin, "'connect' missing 'id' field");
 
 	/* Replace whatever string was originally in here.  */
-	tal_free(css->id);
-	css->id = json_strdup(cmd, buf, idtok);
+	tal_free(connect->id);
+	connect->id = json_strdup(connect, buf, idtok);
 
 	/* Retrieve the `features` from the base command.  */
 	featurestok = json_get_member(buf, result, "features");
 	if (!featurestok)
 		plugin_err(cmd->plugin, "'connect' missing 'features' field");
-	css->features = json_strdup(cmd, buf, featurestok);
+	connect->features = json_strdup(connect, buf, featurestok);
 
-	return plugin_spark_complete(cmd, css->completion);
+	return single_completed(cmd, connect);
 }
-static struct command_result *
-connect_single_spark_start(struct command *cmd,
-			   struct plugin_spark_completion *completion,
-			   struct connect_single_spark *css)
+
+static struct command_result *connect_single_error(struct command *cmd,
+						   const char *buf,
+						   const jsmntok_t *error,
+						   struct connect_single *connect)
 {
-	struct out_req *req;
+	connect->error_json = json_strdup(connect, buf, error);
+	return single_completed(cmd, connect);
+}
 
-	css->completion = completion;
+static void
+connect_single_start(struct command *cmd,
+		     struct connect_single *connect)
+{
+	struct out_req *req;
 
 	req = jsonrpc_request_start(cmd->plugin, cmd, "connect",
-				    &connect_single_spark_done,
-				    &forward_error,
-				    css);
-	json_add_string(req->js, "id", css->id);
-	return send_outreq(cmd->plugin, req);
+				    &connect_single_done,
+				    &connect_single_error,
+				    connect);
+
+	json_add_string(req->js, "id", tal_fmt(tmpctx, "%s%s",
+					       connect->id,
+					       connect->addrhint
+					       ? connect->addrhint : ""));
+	send_outreq(cmd->plugin, req);
 }
 
-static struct command_result *multiconnect(struct command *cmd,
-					   const char *buf,
-					   const jsmntok_t *ids)
+static struct command_result *add_single(struct command *cmd,
+					 struct connect_multi *cm,
+					 const char *buffer, const jsmntok_t *idtok)
 {
-	size_t i;
-	const jsmntok_t *t;
-	struct connect_multi *cm;
-	struct plugin_spark_group *sparks;
+	struct connect_single *connect = tal(cm, struct connect_single);
+	char *atptr;
 
-	cm = tal(cmd, struct connect_multi);
-	cm->subcommands = tal_arr(cm, struct connect_single_spark, ids->size);
+	if (idtok->type != JSMN_STRING)
+		return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
+					tal_fmt(tmpctx,
+						"id: expected string, not '%.*s'",
+						idtok->end - idtok->start,
+						buffer + idtok->start),
+					NULL);
 
-	sparks = plugin_make_spark_group(cm, cmd);
+	connect->features = NULL;
+	connect->error_json = NULL;
+	connect->cm = cm;
+	connect->id = json_strdup(connect, buffer, idtok);
+	atptr = strchr(connect->id, '@');
+	if (atptr) {
+		connect->addrhint = tal_strdup(connect, atptr);
+		*atptr = '\0';
+	} else
+		connect->addrhint = NULL;
+	tal_arr_expand(&cm->subcommands, connect);
+	return NULL;
+}
 
-	/* We know at this point that ids is an array of strings.  */
-	json_for_each_arr(i, t, ids) {
-		struct plugin_spark *spark;
-		cm->subcommands[i].id = json_strdup(cm, buf, t);
-		spark = plugin_start_spark(cmd,
-					   &connect_single_spark_start,
-					   &cm->subcommands[i]);
-		plugin_spark_group_add(sparks, spark);
-	}
+static struct command_result *param_ids(struct command *cmd, const char *name,
+					const char *buffer, const jsmntok_t *tok,
+					struct connect_multi *cm)
+{
+	struct command_result *err;
+
+	/* Is id a string?  Treat it as a single-entry array */
+	if (tok->type == JSMN_ARRAY) {
+		size_t i;
+		const jsmntok_t *t;
+
+		/* Empty array is probably a bug. */
+		if (tok->size == 0)
+			return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
+						"Need at least one id", NULL);
 
-	return plugin_wait_spark_group(cmd, sparks,
-				       &multiconnect_done, cm);
+		json_for_each_arr(i, t, tok) {
+			err = add_single(cmd, cm, buffer, t);
+			if (err)
+				break;
+		}
+	} else
+		err = add_single(cmd, cm, buffer, tok);
+
+	return err;
 }
 
 static struct command_result *json_connect(struct command *cmd,
 					   const char *buf,
 					   const jsmntok_t *params)
 {
-	const jsmntok_t *idtok;
+	struct connect_multi *cm = tal(cmd, struct connect_multi);
 	const char *host;
 	u32 *port;
-	struct out_req *req;
 
+	cm->subcommands = tal_arr(cm, struct connect_single *, 0);
 	if (!param(cmd, buf, params,
-		   p_req("id", param_tok, &idtok),
+		   p_req("id", param_ids, cm),
 		   p_opt("host", param_string, &host),
 		   p_opt("port", param_number, &port),
 		   NULL))
 
-	/* Is id an array?  If so use multi-connect.  */
-	if (idtok->type == JSMN_ARRAY) {
-		size_t i;
-		const jsmntok_t *t;
-
-		if (host)
+	/* For simplicity, we support host and port iff one entry. */
+	if (host || port) {
+		struct connect_single *s = cm->subcommands[0];
+		plugin_log(cmd->plugin, LOG_UNUSUAL, "host = %s, port = %p (%u)",
+			   host, port, port ? *port : 0);
+		if (tal_count(cm->subcommands) != 1)
 			return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
 						"Cannot specify parameter "
-						"'host' when 'id' parameter "
+						"'host' or 'port' when 'id' parameter "
 						"is an array.",
 						NULL);
-		if (port)
+		if (s->addrhint)
 			return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
 						"Cannot specify parameter "
-						"'port' when 'id' parameter "
-						"is an array.",
+						"'host' or 'port' when 'id' contains "
+						" @ hint already",
 						NULL);
+		s->addrhint = tal_strdup(s, "@");
+		if (host)
+			tal_append_fmt(&s->addrhint, "%s", host);
+		if (port)
+			tal_append_fmt(&s->addrhint, ":%u", *port);
+	}
 
-		if (idtok->size == 0)
-			return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
-						"Empty 'id' array: "
-						"nothing to connect.",
-						NULL);
-
-		json_for_each_arr(i, t, idtok) {
-			if (t->type != JSMN_STRING)
-				return command_done_err(cmd,
-							JSONRPC2_INVALID_PARAMS,
-							"All items in 'id' "
-							"array "
-							"must be strings.",
-							NULL);
-		}
-
-		return multiconnect(cmd, buf, idtok);
-	} else if (idtok->type != JSMN_STRING)
-		return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
-					"'id' must be either a string or an "
-					"array of strings.",
-					NULL);
+	for (size_t i = 0; i < tal_count(cm->subcommands); i++)
+		connect_single_start(cmd, cm->subcommands[i]);
 
-	/* If ID is string, just forward to connect_single command.  */
-	req = jsonrpc_request_start(cmd->plugin, cmd, "connect",
-				    &forward_result, &forward_error,
-				    NULL);
-	json_add_string(req->js, "id", json_strdup(tmpctx, buf, idtok));
-	if (host)
-		json_add_string(req->js, "host", host);
-	if (port)
-		json_add_u32(req->js, "port", *port);
-	return send_outreq(cmd->plugin, req);
+	return command_still_pending(cmd);
 }
 
-const struct plugin_command connect_commands[] = {
-	{
-		"multiconnect",
-		"network",
-		"Connect to {id} at {host} (which can end in ':port' if not default). "
-		"{id} can also be of the form id@host[:port].",
-		"Alternately, {id} can be an array of strings of the form id[@host[:port]] "
-		"to connect to multiple peers simultaneously "
-		"(and you should not specify {host} or {port}).",
-		json_connect
-	}
+static const struct plugin_command connect_command = {
+	"multiconnect",
+	"network",
+	"Connect to id, or [id, ...] to connect to multiple.",
+	"{id} can also be of the form id@host[:port].",
+	json_connect
 };
-const unsigned int num_connect_commands = ARRAY_SIZE(connect_commands);
-
-void connect_init(struct plugin *plugin UNUSED,
-		  const char *buf UNUSED, const jsmntok_t *config UNUSED)
-{
-	/* Do nothing.  */
-}
 
 int main(int argc, char **argv)
 {
 	setup_locale();
 	plugin_main(argv,
-		    &connect_init, PLUGIN_RESTARTABLE,
+		    NULL, PLUGIN_RESTARTABLE,
 		    NULL,
-		    connect_commands, num_connect_commands,
+		    &connect_command, 1,
 		    NULL, 0, NULL, 0, NULL);
 }
```

rustyrussell:
@ZmnSCPxj you need to be far more reluctant to create new infrastructure!

multifundchannel is a great idea, but it really wants an "best effort" option which will create as many of the channels as it can.  And it doesn't need multiconnect: it can more easily do that itself.  It doesn't need a new parallel dispatch infrastructure.

It needs to be the *simplest possible* code, so we can all review it and make sure it is as robust as possible.  Not 600 new lines of new concept infrastructure followed by 1700 lines of a new plugin.

I will continue my review tomorrow...

ZmnSCPxj:
Removed sparks.

ZmnSCPxj:
okay.

ZmnSCPxj:
Removed multiconnect.  Will implement `best_effort`, please wait.

ZmnSCPxj:
Implemented `best_effort` flag.

cdecker:
Sorry for being slow at reviewing, will give it a stab today :-)

plugins/multifundchannel.c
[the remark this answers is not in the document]
cdecker: Well, `json_strdup` is not equivalent since it actually does an allocation and copies. We use a mix of `end - start` + `buf + start` and the accessors because we are slowly migrating from the former to the latter, so let's not introduce new instances of the ones we are migrating away from :-)

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1866 @@
...
+	const u8 *funding_script;
+	const char *funding_addr;
+
+	/* The amount to be funded for this destination.
+	If the specified amount is "all" then the `all`
+	flag is set, and the amount is initially 0 until
+	we have figured out how much exactly "all" is,
+	after the dryrun stage.
+	*/
+	bool all;
```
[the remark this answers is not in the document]
ZmnSCPxj: By convention `json_to_sat_or_all` uses the privileged value `AMOUNT_SAT(-1ULL)` to mean "all" but I found it a bit wordy to check `dest->amount == AMOUNT_SAT(-1ULL)` instead of just `dest->all`.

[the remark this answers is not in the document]
ZmnSCPxj: Also that is `amount_sat`.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1866 @@
...
+	/* The overall multifundchannel_command contains an
+	array of multifundchannel_destinations.
+	This provides the index within the array.
+
+	This is used in debug printing.
+	*/
+	unsigned int index;
+
+	/* ID for this destination.  */
+	const char *id;
```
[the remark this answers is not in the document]
ZmnSCPxj: For other parameters:

* `feerate` - Feerate indicator strings require access to infrastructure inside `lightningd`.
* `utxos` - `param_utxos` require access to infrastructure inside `lightningd`.

It seems to me to be not *simplest possible code* if I will have to expose the infrastructure needed above to plugins.

Other parameters than `id` are already in the expected type.

plugins/multifundchannel.c
```diff
@@ -1766,9 +1807,162 @@ multifundchannel_finished(struct multifundchannel_command *mfc)
...
+/* Return true if this destination failed, false otherwise.  */
+static bool dest_failed(struct multifundchannel_destination *dest)
+{
+	switch (dest->state) {
+	case MULTIFUNDCHANNEL_START_NOT_YET:	return false;
+	case MULTIFUNDCHANNEL_CONNECT_FAILED:	return true;
+	case MULTIFUNDCHANNEL_STARTED:		return false;
+	case MULTIFUNDCHANNEL_START_FAILED:	return true;
+	case MULTIFUNDCHANNEL_COMPLETE_FAILED:	return true;
+	case MULTIFUNDCHANNEL_DONE:		return false;
```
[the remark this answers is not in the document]
ZmnSCPxj: That would actually take more SLOC, and I find this "tabular" form easier to audit.  *shrug*.  minor detail.

ZmnSCPxj:
CI build has some very weird "permission denied" errors where some sub-daemon is apparently no longer executable, weird.

ZmnSCPxj:
Rebased, because github is being weird...

ZmnSCPxj:
bump!

ZmnSCPxj:
Adapted to #3787, rebased.

plugins/multifundchannel.c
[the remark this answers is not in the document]
ZmnSCPxj: The advantage of `json_strdup` is that it is DRY and shorter.

     print("%.*s", json_tok_full_len(tok), json_tok_full(buf, tok));
     printf("%s", json_strdup(tmpctx, buf, tok));

ZmnSCPxj:
Used `json_tok_full` and `json_tok_full_len` as well.

ZmnSCPxj:
bumpity bump bump

rustyrussell:
Sorry, this has gotten complex :(

I have patches which rework everything in terms of three new clean primitives:
1. reserveinputs and unreserveinputs.  These take a PSBT.
2. fundpsbt: this takes an amount, feerate, minconf, and returns a PSBT with sufficient inputs that you can have a output of `amount`.

I then rework txprepare, as a plugin, on top of this (for future deprecation, once your multifundchannel goes in, as the current fundchannel uses txprepare).

Unfortunately, this is blocked by libwally, which doesn't allow PSBTs with no outputs.  @niftynei is working on a merge of all the PSBT fixes we need into one libwally branch we can move to, in case the master branch of libwally doesn't have sufficient fixes before release.

I am hoping to get this PR up ASAP...

ZmnSCPxj:
I would suggest merging this in for now, and apply the move to `reserveinputs`/`unreserveinputs`/`fundpsbt`/`signpsbt`/`sendpsbt` later.

ZmnSCPxj:
Rebased now that #3823 is merged.

What is the story around `fundpsbt` at this point?

ZmnSCPxj:
Rebased to handle conflicts.

Now rewriting on top of `fundpsbt` et al.

ZmnSCPxj:
Blocked by #3845

ZmnSCPxj:
Rebased.  Now rewriting for `fundpsbt`/`utxopsbt`/`signpsbt`/`sendpsbt`.

ZmnSCPxj:
Updated to use PSBTs.  Ready for review.  Please notice @rustyrussell @darosior @niftynei

ZmnSCPxj:
Rebased, also added missing checks for dust limit.

ZmnSCPxj:
bumpity bumpity bump bump bumpity bumpity bump bump bumpity bumpity bump bump

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+{
+	bool ok = true;
+	const jsmntok_t *field;
+
+	plugin_log(mfc->cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": %s done.",
+		   mfc->id, mfc->utxos_str ? "utxopsbt" : "fundpsbt");
+
+	field = ok ? json_get_member(buf, result, "psbt") : NULL;
+	ok = ok && field;
```
niftynei, line 935: you can give better error messages if you stop at the problem and return a message there, rather than this "`ok` flag check for every field" strategy.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+}
+
+/* If one of the destinations specified "all", figure out how much that is.  */
+static struct command_result *
+compute_mfc_all(struct multifundchannel_command *mfc)
+{
+	size_t all_index = SIZE_MAX;
+	struct multifundchannel_destination *all_dest;
+
+	assert(mfc->has_all);
```
niftynei, line 974: this check is redundant, we *just* branched on this value

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+		    || dest->state == MULTIFUNDCHANNEL_START_FAILED);
+
+		if (dest->state != MULTIFUNDCHANNEL_START_FAILED)
+			continue;
+
+		/* One of them failed, oh no.
+		Forward the error.
+		*/
+		plugin_log(mfc->cmd->plugin, LOG_DBG,
+			   "mfc %"PRIu64", dest %u: "
```
niftynei, line 1343: print the failing destination node_id here!

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+Those initial commitment transactions are the
+unilateral close (force-close) transactions
+for each channel.
+With unilateral opportunity to close, we can then
+safely broadcast the tx, so that in case the
+peer disappears, we can recover our funds.
+
+The `fundchannel_complete` command performs the
+negotiation with the peer to sign the initial
+commiteent transactions.
```
niftynei, line 1503:
```suggestion
commitment transactions.
```

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+		   mfc->id, dest->index,
+		   node_id_to_hexstr(tmpctx, &dest->id));
+
+	channel_id_tok = json_get_member(buf, result, "channel_id");
+	if (!channel_id_tok)
+		plugin_err(cmd->plugin,
+			   "fundchannel_complete no channel_id: %.*s",
+			   json_tok_full_len(result),
+			   json_tok_full(buf, result));
+	dest->channel_id = json_strdup(mfc, buf, channel_id_tok);
```
niftynei, line 1593: since you don't change the state here, you wouldn't catch if (for some reason) a command failed to complete when you check for 'finishedness' in `after_fundchannel_complete`

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+
+	out = jsonrpc_stream_success(mfc->cmd);
+	json_add_string(out, "tx", mfc->final_tx);
+	json_add_string(out, "txid", mfc->final_txid);
+	json_array_start(out, "channel_ids");
+	for (i = 0; i < tal_count(mfc->destinations); ++i) {
+		json_object_start(out, NULL);
+		json_add_node_id(out, "id", &mfc->destinations[i].id);
+		json_add_string(out, "channel_id", mfc->destinations[i].channel_id);
+		json_add_num(out, "outnum", mfc->destinations[i].outnum);
```
niftynei, line 1875:
```suggestion
		json_add_num(out, "funding_outnum", mfc->destinations[i].outnum);
```
This `outnum` is the `funding_tx`'s outnum

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+The **multifundchannel** RPC command opens multiple payment channels
+with nodes by committing a single funding transaction to the blockchain
+that is shared by all channels.
+
+If not already connected, **multifundchannel** will automatically attempt
+to connect; you may provide a *@host:port* hint appended to the node ID
+so that c-lightning can learn how to connect to the node;
+see lightning-connect(7).
+
+Once the transaction is confirmed, normal channel operations may begin.
```
niftynei, line 21:
```suggestion
Once the transaction is confirmed and both sides have exchanged `funding_locked` messages, normal channel operations may begin.
```
technically 'confirmation' isn't really the open-gate for 'normal channel ops'

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+   a whole number ending in *sat*,
+   a whole number ending in *000msat*, or
+   a number with 1 to 8 decimal places ending in *btc*.
+  The value cannot be less than the dust limit, currently 546 satoshi
+  as of this writing, nor more than 16,777,215 satoshi
+  (unless large channels were negotiated with the peer).
+* *announce* is an optional flag that indicates whether to announce
+  the channel with this, default `true`.
+  If set to `false`, the channel is unpublished.
+* *push\_msat* is the amount of millisatoshis to outright give to the
```
niftynei, line 46: is this field optional?

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+  to fund the channel.
+  The string *all* can be used to specify all available funds
+  (or 16,777,215 satoshi if more is available and large channels were
+  not negotiated with the peer).
+  Otherwise it is in satoshi precision; it can be
+   a whole number,
+   a whole number ending in *sat*,
+   a whole number ending in *000msat*, or
+   a number with 1 to 8 decimal places ending in *btc*.
+  The value cannot be less than the dust limit, currently 546 satoshi
```
niftynei, line 40:
```suggestion
  The value cannot be less than the dust limit, 546 satoshi
```
redundant with 'as of this writing'

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+  as of this writing, nor more than 16,777,215 satoshi
+  (unless large channels were negotiated with the peer).
```
niftynei, line 42: give the option name for this?

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+  If set to `false`, the channel is unpublished.
+* *push\_msat* is the amount of millisatoshis to outright give to the
+  node.
+  This is a gift to the peer, and you do not get a proof-of-payment
+  out of this.
+
+There must be at least one entry in *destinations*;
+it cannot be an empty array.
+
+*feerate* is an optional feerate used for the opening transaction and as
```
niftynei, line 54: i thought we revised feerate strings to be a wider set than just `urgent`, `normal`, `slow`, which are now considered deprecated?

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+*minconf* specifies the minimum number of confirmations that used
+outputs should have. Default is 1.
+
+*utxos* specifies the utxos to be used to fund the channel, as an array
+of "txid:vout".
+
+RETURN VALUE
+------------
+
+On success, the *tx* and *txid* of the signed and broadcsted funding
```
niftynei, line 80:
```suggestion
On success, the *tx* and *txid* of the signed and broadcast funding
```

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+transaction is returned.
+This command opens multiple channels with a single large transaction,
+thus only one transaction is returned.
+
+An array of *channel\_ids* is returned;
+each entry of the array is an object,
+ with an *id* field being the node ID of the peer,
+ an *outnum* field being the output number of the transaction
```
niftynei, line 92:
```suggestion
 an *funding_outnum* field being the output number of the transaction
```
as per previous suggested update

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+  that anchors this channel,
```
niftynei, line 93: i'm not sure i'd use the term `anchor` here as it has significance elsewhere -- you can just say 'output number of the funding transaction.'

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+ and *channel_id* field being the channel ID with that peer.
+
+On failure, none of the channels are created.
```
niftynei, line 105:
```suggestion
On failure, no channels are created.
```

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+
+The following error codes may occur:
+* -1: Catchall nonspecific error.
+- 300: The maximum allowed funding amount is exceeded.
+- 301: There are not enough funds in the internal wallet (including fees) to create the transaction.
+- 302: The output amount is too small, and would be considered dust.
```
niftynei, line 111:
```suggestion
- 302: A destination's proposed output amount is too small, and would be considered dust.
```

doc/lightning-multifundchannel.7.md
```diff
@@ -0,0 +1,127 @@
...
+- 303: Broadcasting of the funding transaction failed, the internal call to bitcoin-cli returned with an error.
```
niftynei, line 112:
```suggestion
- 303: Broadcasting the funding transaction failed, the internal call to the bitcoin backend returned with an error.
```
We may not be using bitcoin-cli?

tests/test_connection.py
```diff
@@ -1252,21 +1252,153 @@ def test_multifunding_one(node_factory, bitcoind):
...
+    destinations = [{"id": '{}@localhost:{}'.format(l2.info['id'], l2.port),
+                     "amount": 50000},
+                    {"id": '{}@localhost:{}'.format(l3.info['id'], l3.port)}]
+    with pytest.raises(RpcError):
+        l1.rpc.multifundchannel(destinations)
+    destinations = [{"amount": 50000},
+                    {"id": '{}@localhost:{}'.format(l3.info['id'], l3.port),
+                     "amount": 50000}]
+    with pytest.raises(RpcError):
+        l1.rpc.multifundchannel(destinations)
```
niftynei, line 1401: missing tests for : dust + insufficient funds

tests/test_connection.py
```diff
@@ -1401,6 +1401,69 @@ def test_multifunding_param_failures(node_factory):
...
+                     "amount": 50000}]
+
+    for i, d in enumerate(disconnects):
+        # Should succeed due to best-effort flag.
+        l1.rpc.multifundchannel(destinations, best_effort=True)
+        bitcoind.generate_block(6, wait_for_mempool=1)
+
+        # Only l3 should fail to have channels.
+        for node in [l1, l2, l4]:
+            node.daemon.wait_for_log(r'to CHANNELD_NORMAL')
```
niftynei, line 1433: assert that l3 indeed failed, and in the correct state/location?

plugins/multifundchannel.c
```diff
@@ -1836,9 +1877,167 @@ multifundchannel_finished(struct multifundchannel_command *mfc)
...
+			tal_arr_expand(&mfc->removeds, removed);
+		} else {
+			plugin_log(mfc->cmd->plugin, LOG_DBG,
+				   "mfc %"PRIu64", dest %u: "
+				   "succeeded.",
+				   mfc->id, dest->index);
+
+			/* Reset its state.  */
+			dest->state = MULTIFUNDCHANNEL_START_NOT_YET;
+			/* We do not mess with `dest->index` so we have
```
niftynei, line 2000: using the node_id instead of index will fix this bit.

plugins/multifundchannel.c
```diff
@@ -1836,9 +1877,167 @@ multifundchannel_finished(struct multifundchannel_command *mfc)
...
+	tal_free(redo);
+
+	/* Clean up previous funding transaction if any.  */
+	mfc->psbt = tal_free(mfc->psbt);
+	mfc->txid = tal_free(mfc->txid);
+
+	/* Filter out destinations.  */
+	old_destinations = tal_steal(tmpctx, mfc->destinations);
+	mfc->destinations = NULL;
+	new_destinations = tal_arr(mfc, struct multifundchannel_destination,
```
niftynei, line 1970: why not just use `mfc->destinations`?

niftynei:
Found a crash case while trying it out on command line, pushed a patch to fix it up.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,2146 @@
...
+#include <common/type_to_string.h>
+#include <common/utils.h>
+#include <plugins/libplugin.h>
+#include <stdarg.h>
+#include <stdlib.h>
+#include <string.h>
+
+/* State of whether we have done `fundchannel_start`.  */
+enum multifundchannel_state {
+	/* We have not yet performed `fundchannel_start`.  */
```
darosior, line 29:
```suggestion
/* Current state of the funding process.  */
enum multifundchannel_state {
	/* We have not yet performed `fundchannel_start`.  */
```

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+	const char *feerate_str;
+	u32 *minconf;
+	const char *utxos_str;
+
+	struct multifundchannel_command *mfc;
+
+	struct command_result *result;
+
+	if (!param(cmd, buf, params,
+		   p_req("destinations", param_tok, &json_destinations),
```
rustyrussell, line 2062: Always prefer to do inline parsing where possible (rahter than using param_tok and parsing later).  That way the "check" command is far more effective (which always exits this function after param).

Of course, IIRC we haven't implemented check for plugins, but the theory applies!

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -862,7 +862,8 @@ def multifundchannel(self, destinations, feerate=None, minconf=None, utxos=None,
             "destinations": destinations,
             "feerate": feerate,
             "minconf": minconf,
-            "utxos": utxos
+            "utxos": utxos,
+            "best_effort": best_effort
```
rustyrussell, line 866: I changed my mind.  This parameter should be "minchannels", to indicate how many channels are acceptable.  Default is all of them, otherwise it proceeds unless there fewer than "minchannels" channels remaining.  0 is invalid.

rustyrussell (changes requested):
I really like this. There are some cosmetic things that @niftynei and I both commented on, and I changed my mind on `best_effort` as a flag (sorry!).

I've marked it for this release, hope we can get it in!

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+		plugin_err(mfc->cmd->plugin,
+			   "Overflow dust limit and change fee.");
+
+	/* Is the excess over the limit?  */
+	if (amount_sat_greater(mfc->excess_sat, change_min_limit)) {
+		bool ok = amount_sat_sub(&mfc->change_amount,
+					 mfc->excess_sat, change_fee);
+		assert(ok);
+		mfc->change_needed = true;
+		if (!mfc->change_scriptpubkey)
```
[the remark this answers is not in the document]
ZmnSCPxj: Yes.  A previous `best_effort` attempt may load this, and it seems wasteful to keep asking for more `newaddr` for each best-effort attempt, just reuse the previous.  So, it is indeed possible if a previous best-effort succeeded.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+perform_multiconnect(struct multifundchannel_command *mfc)
+{
+	unsigned int i;
+
+	plugin_log(mfc->cmd->plugin, LOG_DBG,
+		   "mfc %"PRIu64": multiconnect.", mfc->id);
+
+	mfc->pending = tal_count(mfc->destinations);
+
+	for (i = 0; i < tal_count(mfc->destinations); ++i)
```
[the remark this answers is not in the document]
ZmnSCPxj: preincrement operator.  I prefer this due to habit from very old, very lousy compilers where `++i` is compiled at `-O0` more efficiently than `i++`: the latter causes the compiler to create a useless move to a temporary register that is later reused.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+/* Object for performing cleanup.  */
+struct multifundchannel_cleanup {
+	size_t pending;
+	struct command_result *(*cb)(void *arg);
+	void *arg;
+};
+
+/* Cleans up a PSBT.  */
+static void
+mfc_cleanup_psbt(struct command *cmd,
```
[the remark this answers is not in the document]
ZmnSCPxj: I absolutely detest the reverse ordering that the "usage pattern" induces, I find it more confusing than the occasional decaration that you can easily pass over while reading the code more sensibly in the correct order.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+#include <plugins/libplugin.h>
+#include <stdarg.h>
+#include <string.h>
+
+/* State of whether we have done `fundchannel_start`.  */
+enum multifundchannel_state {
+	/* We have not yet performed `fundchannel_start`.  */
+	MULTIFUNDCHANNEL_START_NOT_YET = 0,
+	/* The `connect` command failed.  `*/
+	MULTIFUNDCHANNEL_CONNECT_FAILED,
```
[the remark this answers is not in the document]
ZmnSCPxj: Seems better IMO to know *which* part failed.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+	struct amount_sat change_amount;
+
+	/* The txid of the final funding transaction.  */
+	struct bitcoin_txid *txid;
+
+	/* The actual tx of the actual final funding transaction
+	that was broadcast.
+	*/
+	const char *final_tx;
+	const char *final_txid;
```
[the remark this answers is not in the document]
ZmnSCPxj: This is the return value from `sendpsbt`.

plugins/multifundchannel.c
ZmnSCPxj: Leads to much larger code that checks at each point (about 33% more, I tried with that already), and in practice giving the entire result when this errors helps debugging well enough.

plugins/multifundchannel.c
ZmnSCPxj: I do not understand.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+{
+	size_t all_index = SIZE_MAX;
+	struct multifundchannel_destination *all_dest;
+
+	assert(mfc->has_all);
+
+	for (size_t i = 0; i < tal_count(mfc->destinations); ++i) {
+		struct multifundchannel_destination *dest;
+		dest = &mfc->destinations[i];
+		if (dest->all) {
```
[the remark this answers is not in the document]
rustyrussell: Meh, need to loop anyway, so this doesn't help.

plugins/multifundchannel.c
```diff
@@ -0,0 +1,1939 @@
...
+	field = ok ? json_get_member(buf, result, "estimated_final_weight") : NULL;
+	ok = ok && field;
+	ok = ok && json_to_u32(buf, field, &mfc->estimated_final_weight);
+
+	/* msat LOL.  */
+	field = ok ? json_get_member(buf, result, "excess_msat") : NULL;
+	ok = ok && field;
+	ok = ok && parse_amount_sat(&mfc->excess_sat,
+				    buf + field->start,
+				    field->end - field->start);
```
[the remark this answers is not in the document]
ZmnSCPxj: This is still shorter:

```c
field = ok ? son_get_member(buf, result, "psbt") : NULL;
ok = ok && field;
mfc->psbt = ok ? psbt_from_b54(mfc, buf + field->start, field->end - field->start) : NULL;
ok = ok && mfc->psbt;
...

if (!ok)
    plugin_err(...);
```

*shrug*.

rustyrussell:
rebased onto master, put some fixmes on top.  Needs merging in properly, but this is better for review.

rustyrussell:
OK, I merged in the fixmes into one smooth progression.  Will apply once Travis is happy.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/0f24eb80337ac168f9285977dae68ecf633e22c3

PR #3764, opened 2020-06-10
listfunds: also list reserved outputs

niftynei:
Currently 'listfunds' lies, a teensy eeinsy bit, in that it doesn't list
all of the funds in a wallet (it omits reserved wallet UTXOs). This
change makes the reserved outputs visible by listing them in the
'outputs' section along with a new field, 'reserved', which denotes the
UTXO's state

Changelog-Changed: JSON-RPC: `listfunds` 'outputs' now includes reserved outputs, designated as 'reserved' = true

### DIFF
--- a/doc/lightning-listfunds.7
+++ b/doc/lightning-listfunds.7
@@ -20,6 +20,7 @@ channels\.
 
 Each entry in \fIoutputs\fR will include:
 
+.RS
 .IP \[bu]
 \fItxid\fR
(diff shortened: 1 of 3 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3764/commits/bcd3a4b0f15e27a146f29a1dfebddee3e827b378

PR #3766, opened 2020-06-12
Weekly cleanups 2020/23

cdecker:
Just a minor typo and a test exclusion I couldn't fit in anywhere else

Changelog-None


cdecker:
Since I was at it I thought I'd also bump pyln-proto to v0.8.2, to be in sync with the binary distribution (had a couple of users wonder why the discrepancy) and I also added Makefile targets to release the version on test.pypi (including testing the published version) and prod PyPI.

PR #3769, opened 2020-06-15
Merge multifundchannel and fundchannel plugins, make fundchannel a thin wrapper around multifundchannel

ZmnSCPxj:
Requres #3763


ZmnSCPxj:
Update based on new #3763

ZmnSCPxj:
Updated based on latest #3763

cdecker:
Blocked by #3763

tests/test_closing.py
```diff
@@ -2318,11 +2318,13 @@ def test_option_upfront_shutdown_script(node_factory, bitcoind, executor):
...
     keyidx = int(l1.db_query("SELECT intval FROM vars WHERE name='bip32_max_index';")[0]['intval'])
 
-    # Expect 1 for change address, 1 for the channel final address,
-    # which are discarded as the 'scratch' tx that the fundchannel
-    # plugin makes, plus 1 for the funding address of the actual
+    # Expect 1 for change address, plus 1 for the funding address of the actual
     # funding tx.
-    addr = l1.rpc.call('dev-listaddrs', [keyidx + 3])['addresses'][-1]
+    addr = l1.rpc.call('dev-listaddrs', [keyidx + 2])['addresses'][-1]
+    # the above used to be keyidx + 3, but that was when `fundchannel`
```
niftynei, line 2324: this comment is unnecessary, as the commit history will preserve the deleted comment + value.

plugins/spender/fundchannel.c
```diff
@@ -0,0 +1,124 @@
...
+	txid = ok ? json_get_member(buf, result, "txid") : NULL;
+	ok = ok && txid;
+	channel_ids_array = ok ? json_get_member(buf, result, "channel_ids") : NULL;
+	ok = ok && channel_ids_array;
+	channel_ids_obj = ok ? json_get_arr(channel_ids_array, 0) : NULL;
+	ok = ok && channel_ids_obj;
+	channel_id = ok ? json_get_member(buf, channel_ids_obj, "channel_id") : NULL;
+	ok = ok && channel_id;
+	outnum = ok ? json_get_member(buf, channel_ids_obj, "outnum") : NULL;
+	ok = ok && outnum;
```
niftynei, line 110: i hate this. just return an error at the point which it fails.

plugins/spender/fundchannel.c
```diff
@@ -0,0 +1,124 @@
...
+
+	if (!ok)
+		plugin_err(cmd->plugin,
+			   "Unexpected result from nultifundchannel: %.*s",
```
niftynei, line 114:
```suggestion
			   "Unexpected result from multifundchannel: %.*s",
```

niftynei:
cool, nice to see the code get consolidated

plugins/.gitignore
```diff
@@ -1,5 +1,6 @@
 autoclean
 bcli
 fundchannel
-pay
 multifundchannel
+pay
+spenderp
```
rustyrussell, line 4: spenderp?

common/json.c
```diff
@@ -838,7 +838,12 @@ void json_add_tok(struct json_stream *result, const char *fieldname,
 		if (json_tok_is_num(buffer, tok)) {
 			json_to_int(buffer, tok, &i);
 			json_add_num(result, fieldname, i);
-		}
+		} else if (buffer[tok->start] == 't' || buffer[tok->start] == 'T')
```
rustyrussell, line 841: Garbage in, garbage out.

If they feed in an invalid primitive, we should either pass it through, or get upset with the passer.

But I can't understand why json_add_tok exists.  Why are we trying to normalize tokens?  We should just append them as we're asked to.

@darosior ?

rustyrussell (approved):
Other than minor cleanup suggested by @niftynei looks good.

common/json.c
darosior: > But I can't understand why json_add_tok exists. Why are we trying to normalize tokens? We should just append them as we're asked to.

Using `json_add_member` ? I don't recall exactly but it seems cleaner to use our already-present helpers.

Anyway, i 1) don't know how comes i only handled the num primitive and 2) am pretty sure `Null` `False` and `True` here and below are invalid JSON (maybe just use `json_add_literal` in the `else` branch).

rustyrussell: It also turns strings into bools, etc.  It's just *weird*.

I'm testing the simplest variant now:
```C
void json_add_tok(struct json_stream *result, const char *fieldname,
                  const jsmntok_t *tok, const char *buffer)
{
	char *space;
	assert(tok->type != JSMN_UNDEFINED);

	space = json_member_direct(result, fieldname, tok->end - tok->start);
	memcpy(space, buffer + tok->start, tok->end - tok->start);
}
```

rustyrussell:
Fairly simple rebase.  Removed the json_add_tok nonsense in favor of a simple copy.

PR #3770, opened 2020-06-15
pyln.proto.message minor fixes

rustyrussell:
My tests weren't exhaustive, and now I'm hitting things in more complex testing of the spec.
Changelog-None

### DIFF
--- a/contrib/pyln-proto/pyln/proto/message/array_types.py
+++ b/contrib/pyln-proto/pyln/proto/message/array_types.py
@@ -1,5 +1,5 @@
 from .fundamental_types import FieldType, IntegerType, split_field
-from typing import List, Optional, Dict, Tuple, TYPE_CHECKING, Any
+from typing import List, Optional, Dict, Tuple, TYPE_CHECKING, Any, Union
 from io import BufferedIOBase
 if TYPE_CHECKING:
     from .message import SubtypeType, TlvStreamType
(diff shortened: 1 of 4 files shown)


contrib/pyln-proto/pyln/proto/message/array_types.py
```diff
@@ -41,6 +41,13 @@ def val_to_str(self, v: List[Any], otherfields: Dict[str, Any]) -> str:
         s = ','.join(self.elemtype.val_to_str(i, otherfields) for i in v)
         return '[' + s + ']'
 
+    def val_to_py(self, v: Any, otherfields: Dict[str, Any]) -> Union[str, List[Any]]:
```
cdecker, line 44: Why is the `to_py()` method not part of the instances themselves, but rather the type class? Seems like a weird indirection.

cdecker: It'd also mean that the instance type (type of `v`) can extend the native python type and just work as one, without having to have an explicit conversion.

rustyrussell:
It's because we create a class instance representing a bolt type.  It would be cleaner to make it a real type, then dynamically generate the other classes, but that's a bit advanced for me.

And in practice it doesn't make a huge difference to class users, who rarely interact below the Message level.

I would like to rewrite it, but time....

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3770/commits/600b56eca47a5185abddf2ad2f6fd83d8bb34f69

PR #3772, opened 2020-06-16
An RPC connection fix and some vbyte fun

### LINKED ISSUE #3771: Onchain fee roundoff
We seem to be rounding off fees in a way that is incompatible with `bitcoind`: https://github.com/ElementsProject/lightning/issues/3591#issuecomment-641466505 by @whitslack

> I was withdrawing with `feerate=1000perkb`, which should be the minimum-allowed fee rate. Indeed, `bitcoin-cli getmempoolinfo` reports:
>
> ```
> {
>   "loaded": true,
>   "size": 15097,
>   "bytes": 9207924,
>   "usage": 32831760,
>   "maxmempool": 64000000,
>   "mempoolminfee": 0.00001000,
>   "minrelaytxfee": 0.00001000
> }
>```

darosior:
90bce56: I regularly hit this corner case on a bloated-memory node

9230907 and 2fbd490 fix #3771.

Fixes #3771

### DIFF
--- a/lightningd/json.c
+++ b/lightningd/json.c
@@ -1,6 +1,7 @@
 #include <arpa/inet.h>
 #include <bitcoin/address.h>
 #include <bitcoin/base58.h>
+#include <bitcoin/feerate.h>
 #include <bitcoin/script.h>
 #include <ccan/json_escape/json_escape.h>
 #include <ccan/mem/mem.h>
@@ -159,6 +160,8 @@ struct command_result *param_feerate(struct command *cmd, const char *name,
 
 	*feerate = tal(cmd, u32);
 	**feerate = feerate_from_style(num, style);
+	if (**feerate < FEERATE_FLOOR)
+		**feerate = FEERATE_FLOOR;
 	return NULL;
 }
 
--- a/lightningd/jsonrpc.c
+++ b/lightningd/jsonrpc.c
@@ -597,8 +597,10 @@ static struct command_result *command_exec(struct json_connection *jcon,
 	if (res == &pending)
 		assert(cmd->pending);
 
-	list_for_each(&jcon->commands, cmd, list)
-		assert(cmd->pending);
+	/* The command might outlive the connection. */
+	if (jcon)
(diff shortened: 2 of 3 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3772/commits/2fbd49061c5b2e3cefea25e9cef2573bd99991ed

lightningd/chaintopology.c
```diff
@@ -474,7 +474,7 @@ u32 feerate_from_style(u32 feerate, enum feerate_style style)
 	case FEERATE_PER_KBYTE:
 		/* Everyone uses satoshi per kbyte, but we use satoshi per ksipa
 		 * (don't round down to zero though)! */
-		return (feerate + 3) / 4;
+		return feerate / 4 + 3;
```
ZmnSCPxj, line 477: Would this lead to us computing a higher feerate than "Everyone" later on?

darosior: I hesitated to use a "return FEERATE_FLOOR if < FEERATE_FLOOR ", but I think we were actually computing *lower* feerate than bitcoind before this.

ZmnSCPxj: Okay.  Will ACK this.  Unless you want to change the second commit message.

ZmnSCPxj:
ACK 2fbd49061c5b2e3cefea25e9cef2573bd99991ed

lightningd/chaintopology.c
darosior: Will need to fixup the tests, so will amend the second commit message while at it.

darosior:
This messed up all feerate-related tests : now I regret not to have made the API in WU for `estimatefees`...

darosior:
Ok, so I went for the "don't go below `FEERATE_FLOOR` approach instead of monkey-patching all the hardcoded feerates in the tests.

ZmnSCPxj:
The linked travis build is passing, but it seems github is not noticing it....

PR #3773, opened 2020-06-16
paymod: Feature parity, finally (part 03)

cdecker:
_This PR merges into the branch of the second part `paymod-02` and not
`master`. This should reduce the number of commits that are to be reviewed,
but I'm still unclear if merging the first part will re-address this PR to
point to `master`. It's an experiment, so let's see if this works for stacking
multiple PRs :wink:_

This is the final part of the large refactor into a new payment flow (see
#3753 and #3760 for the previous parts). It implements most of the missing
functionality as `payment_modifier`s and reaches feature parity with the old
`pay` command.

The PR is not as clean as it could be, in particular some things get moved
around a bit in commits to consolidate their location, but rebasing to get
them into their final position right away would be more work than its worth.

 - I moved some of the deserialization functions of RPC results into
   `common/rpcclient.[ch]` in order to cut down on future need to manually
   deserialize those results. It parses a bit more than any single use might
   need, but it frees us from having to juggle JSON dict keys and the like in
   the future.

## New Features

 - I added a new `direct_pay_mod` which overrides the call to `getroute`
   altogether if we know we have a direct channel that can be attempted for
   the payment, thus disabling any form of fuzzing or shadow-routing in cases
   we definitely don't need it. On failure we will still fall back to the
   normal behavior.
   
 - We now track `channel_hints` by applying routes when we start an attempt,
   and un-applying them if the attempt fails. While we do not share these
   hints across payments due to concerns of acting on stale information, these
   updated hints will become very important once we have an MPP modifier which
   will result in multiple concurrent attempts needing up-to-date estimates on
   the state of a channel. These hints are still only best effort estimates,
   and we will update them as we learn more information on capacity and other
   constraints.

 - We move from a "compute-constraints-on-the-fly" system to a precomputed
   constraints model, in which we precompute the fee and cltv budget at the
   beginning and can then split that budget for retries and parallel
   attempts. This reduces the number of places in which the fees and CTLV
   deltas get computed considerably.
   
## Activation

For testing purposes the new implementation is guarded behind the `COMPAT`
flag, we are discussing however whether we just activate the new
implementation by default, and provide an escape hatch in the form of a CLI
flag to use the old implementation. We can likely gather some feedback once
merged into `master` since the plugin repository has nightly CI runs
configured to run against `COMPAT=0` on `master`.

## Next steps

And now for the reason we started out on this long and tortuous journey in the
first place. We can now implement `keysend` sending support and MPP sending
support with relative ease, by implementing them as modifiers :wink: I will
queue up those as separate pull requests.

Changelog-None


ZmnSCPxj:
> we are discussing however whether we just activate the new
implementation by default, and provide an escape hatch in the form of a CLI
flag to use the old implementation.

I say go for it, and keep the old pay as the command `pay_old`, which we warn as deprecated.

cdecker:
Rebased on top of paymod-02: [range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3773/range_diff/53764fa427ee92825f3ea85519481699cc558f5a..5aba80d14455bc4ce41165483db3224e6196f1d6)

lightningd/htlc_set.c
```diff
@@ -181,7 +181,7 @@ void htlc_set_add(struct lightningd *ld,
 		return;
 	}
 
-	if (amount_msat_eq(set->so_far, total_msat)) {
+	if (amount_msat_greater_eq(set->so_far, total_msat)) {
```
rustyrussell, line 184: This implies we violate the spec.  @joostjager really wanted it to be exact, IIRC.  This might fail against other implementations, and the spec certainly allows them to fail.

But remember, with MPP, fuzzing the amount is unnecessary (or certainly much less).

```
    - MUST ensure that the total `amount_msat` of the HTLC set which arrives at the payee
      is equal to `total_msat`.
```

common/rpcclient.c
```diff
@@ -0,0 +1,192 @@
+#include <common/rpcclient.h>
+
+#include <assert.h>
```
rustyrussell, line 3: These seem very much a part of libplugin, TBH.  Maybe that's because we currently don't supply any other C clients, but I'm not sure what we win by moving this here?

cdecker: Happy to move them into `libplugin`, and move them back out again once we have a user for it.

lightningd/htlc_set.c
cdecker: I double-checked with @t-bast on IRC, and they also said they were checking for `>=` and not `==`. And at least for some of the simpler splitter strategies MPP still leaks information (e.g., the simple divide and conquer strategy employed by lnd can still leak the amount transferred).

plugins/libplugin-pay.c
```diff
@@ -1577,3 +1579,170 @@ static void exemptfee_cb(struct exemptfee_data *d, struct payment *p)
 REGISTER_PAYMENT_MODIFIER(exemptfee, struct exemptfee_data *,
 			  exemptfee_data_init, exemptfee_cb);
 
+/* BOLT #7:
+ *
+ * If a route is computed by simply routing to the intended recipient and
+ * summing the `cltv_expiry_delta`s, then it's possible for intermediate nodes
```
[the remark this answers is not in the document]
cdecker: Argh, fixups gone awry. Will attempt to split them again.

lightningd/htlc_set.c
rustyrussell: Yes, you're allowed to accept it, but you're not supposed to send it!

And sure, dumb divide in half will leak, but no more than non-MPP.

plugins/pay.c
```diff
@@ -1499,6 +1572,34 @@ static struct command_result *json_paystatus(struct command *cmd,
...
+	list_for_each(&payments, p, list) {
+		assert(p->parent == NULL);
+		if (b11str && !streq(b11str, p->bolt11))
+			continue;
+
+		json_object_start(ret, NULL);
+		if (p->bolt11)
+			json_add_string(ret, "bolt11", p->bolt11);
+		json_add_u64(ret, "msatoshi",
+			     p->amount.millisatoshis); /* Raw: JSON */
```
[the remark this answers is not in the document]
rustyrussell: Yes, you need to use "json_add_msat_only()" and call it "amount_msat".

The rule for JSON output is:
1. Currency amounts must be exposed in millisatoshis.
2. Currency values must be as strings ending in "msat".
3. Currency fieldnames must end in "_msat".

If we'd realized that "millisatoshis everywhere" was a good idea from early on, we could have just done rule 1.
But since we didn't, we need rule 2.  And to help JSON conversion, we need rule 3.

lightningd/htlc_set.c
t-bast: > some of the simpler splitter strategies MPP still leaks information (e.g., the simple divide and conquer strategy employed by lnd can still leak the amount transferred)

On that note we recently changed our splitting algorithm in eclair, and it's behaving quite well so far (see https://github.com/ACINQ/eclair/pull/1427). This algorithm doesn't leak total amounts like lnd's divide and conquer does and requires less initial failed attempts in practice (but costs more CPU though). I'm curious what c-lightning is doing for MPP-split at the moment, and what algorithms you're considering for a (potential) upgrade.

cdecker:
> > we are discussing however whether we just activate the new
> > implementation by default, and provide an escape hatch in the form of a CLI
> > flag to use the old implementation.
> 
> I say go for it, and keep the old pay as the command `pay_old`, which we warn as deprecated.

I wonder if we should even provide a `pay_old` at all, or whether we just add a `legacy` flag to the `pay` command that then switches over to the legacy code, directly after parsing the options. Could get us into trouble if we add params, but for now it works.

plugins/libplugin-pay.c
```diff
@@ -1210,3 +1211,208 @@ static void local_channel_hints_cb(void *d UNUSED, struct payment *p)
...
+			if (!route_msatoshi(&dest_amount, p->amount,
+				    routehint + i + 1,
+					    tal_count(routehint) - i - 1)) {
+				/* Just let it fail, since we couldn't stitch
+				 * the routes together. */
+				return payment_continue(p);
+			}
+
+			hop.nodeid = *route_pubkey(p, routehint, i + 1);
+			hop.style = ROUTE_HOP_TLV;
```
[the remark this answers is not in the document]
cdecker: You're probably right, we can't tell whether this is a legacy or a TLV payload hop. The legacy code was leaving the field out, which defaults to the legacy hop payload format. But this is unlikely to cause any issues imho and it'd accelerate the migration to TLV payloads a bit: non-TLV peers that are in routehints will stop working.

I'll make this a legacy payload for now, but at some point we'll want to deprecate that format completely so we'll have to switch it back later.

plugins/libplugin-pay.c
```diff
@@ -220,10 +222,42 @@ static struct command_result *payment_getroute_result(struct command *cmd,
...
+		payment_fail(p);
+		return command_still_pending(cmd);
+	}
+
+	if (amount_msat_greater(fee, p->fee_budget)) {
+		plugin_log(p->plugin, LOG_INFORM,
+		    "Fee exceeds our fee budget: %s > %s, discarding route",
+		    type_to_string(tmpctx, struct amount_msat, &fee),
+		    type_to_string(tmpctx, struct amount_msat, &p->fee_budget));
+		payment_fail(p);
```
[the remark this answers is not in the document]
cdecker: It i indeed passed  to `payment_fail` in a later commit, which is then surfaced to the user through the results of `pay` and `paystatus`.

plugins/libplugin-pay.c
```diff
@@ -220,10 +222,42 @@ static struct command_result *payment_getroute_result(struct command *cmd,
...
+		return command_still_pending(cmd);
+	}
+
+	if (p->route[0].delay > p->cltv_budget) {
+		plugin_log(p->plugin, LOG_INFORM,
+			   "CLTV delay exceeds our CLTV budget: %d > %d",
+			   p->route[0].delay, p->cltv_budget);
+		payment_fail(p);
+		return command_still_pending(cmd);
```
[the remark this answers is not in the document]
cdecker: Yep, passed to `payment_fail` and then to the caller as the permanent failure reason for this attempt.

lightningd/htlc_set.c
cdecker: I will add an opt-out flag to the amount-fuzzing in shadow routes and the `getroute` request, if used in combination with MPP. Until we have MPP implemented (I'm working on that right now) we shouldn't disable fuzzing.

The commit in question is on the receiving end, where we should indeed be more permissive than what we send out. I propose I fixup the fuzzing in the MPP PR and leave it as is here where it is still useful since we have a single part at most here.

common/rpcclient.c
cdecker: Moved to `libplugin.[ch]`, we can move them again once we have a C API that we want to expose outside of the plugins.

lightningd/htlc_set.c
cdecker: Thanks for sharing the MPP logic with us @t-bast, let me try to rephrase your strategy, to see if I understood it correctly.

 - You split only amounts above 15ksat, anything below that is considered small enough to go through eventually.
 - You only split into up to 6 parts
 - You compute a number of routes initially using Yen's k-shortest paths and draw from those routes
 - When trying a route you update your estimated capacities for edges in all routes in order to avoid overloading edges.
 - Amounts are randomized in the range [0.2, 1] of the estimated capacity of the smallest channel on the route

It seems pretty similar to what my code does:
 - We have two splitter phases: a `presplit` modifier which splits large amounts before even starting a first attempt, and an `adaptive_splitter` which reacts to new information about edges and failed payments (mainly capacity estimates and disabled channels)
 - The `presplit` modifier will split anything larger than our threshold (10ksat at the moment) so that's pretty much the opposite of what you're doing. This is derived from a recent round of probing I did (see results below).
 - When splitting we fuzz the amount by +/- 10% uniformly around the target split amount (10ksat in the presplit case, or 1/2 previous attempt in the adaptive splitter).
 - We track estimates and new information we learn about channels and nodes as `channel_hints` in the root of the payment tree
 - We iteratively ask for new routes using `getroute` for each attempt, this is a bit more costly than doing it in batch with Yen's algorithm, but we can provide up to date info on the channel state, and reduce our in-memory state-management during a payment.

About the capacity probes I did: I implemented a simple binary search on the payment amount and then iterated through all nodes and tested the maximum single-part payment that will reach the destination. After testing about 35% of the network the histogram of successful capacities looks like this:

probed-channel-capacities

The 80th percentile is marked in black, and the 50th percentile marked in red. With 10ksat for the pre-splitter we end up with parts that are ~1USD in size, and that will succeed in 80% of cases without requiring further splits. Our goal here is to be rather aggressive, not optimize for fees, but for speed. At 10ksat the proportional fee still dominates the overall cost, so we're confident that adding a couple more parts is not going to be too costly.

Notice that these measurements use my usual worst-case assumption of any sender-receiver pair being equally likely, which we pretty much know is too pessimistic. Not surprisingly the bottlenecks on the path are clustered around the 100ksat and 500ksat sizes, which are the most popular total capacity sizes in the network today.

We also plan to keep measuring and adapting these constants as we learn new things and the network evolves.

So overall we seem to agree on the strategy, but have chosen different parameters, let's maybe open an issue discussing these strategies on https://github.com/lightningnetwork/lightning-rfc to discuss tradeoffs and experiences :-)

t-bast: Thanks for all the details! I agree that it will be great to discuss at length heuristics we're trying and how well they seem to behave on real payments, as all simulations we do will never accurately reflect the reality of the unknown remote balances and the effectiveness of retries.

Our strategies are very similar indeed. Your summary of mine is correct, except maybe one point to clarify: I run Yen's k-shortest path not with the amount requested but instead with 15k sats, and then as you understood I "fill" these routes depending the amount they're able to carry (+ some randomization). I think this makes it even closer to your strategy of pre-splitting (but with the limitation that I impose a maximum number of parts of 6 - although configurable by the user at runtime if necessary - and this upper bound is ignored when paying a direct peer, which is a very useful heuristic when coupled with trampoline).

One thing that's always been bothering me: eclair considers that you can send up to the channel's capacity through a link, I've always wondered if using instead `channel capacity / 2` instead would yield better results. I've been holding off on implementing this change as channel capacities are (hopefully) much bigger than payment amounts so it probably doesn't matter, but I'm curious to know if you've given it some thoughts.

Thanks for sharing your graph, my analysis was only based on channel capacity, not on actual success of sending payments, so my numbers may not be very accurate.

plugins/libplugin-pay.c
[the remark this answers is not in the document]
cdecker: After further looking into it, the legacy code assumes that routehints support TLVs, which is tested in `test_tlv_or_legacy`, so it appears we are already gently nudging users to upgrade to TLV.

plugins/libplugin-pay.c
```diff
@@ -233,25 +233,54 @@ static struct command_result *payment_getroute_error(struct command *cmd,
...
+
+static const struct node_id *payment_get_excluded_nodes(const tal_t *ctx,
+						  struct payment *p)
+{
+	struct payment *root = payment_root(p);
+	return root->excluded_nodes;
+}
+
+/* Iterate through the channel_hints and exclude any channel that we are
+ * confidet will not be able to handle this payment. */
```
rustyrussell, line 422: typo: confidet -> confident

plugins/pay.c
```diff
@@ -1833,7 +1833,8 @@ static void init(struct plugin *p,
 	maxdelay_default = atoi(field);
 }
 
-struct payment_modifier *paymod_mods[3] = {
+struct payment_modifier *paymod_mods[4] = {
```
rustyrussell: Still don't need `[4]` here: `[]` is sufficient and correct.

plugins/libplugin-pay.c
```diff
@@ -1236,10 +1255,11 @@ void payment_continue(struct payment *p)
 	abort();
 }
 
-void payment_fail(struct payment *p)
+void payment_fail(struct payment *p, const char *reason)
 {
 	p->end_time = time_now();
 	p->step = PAYMENT_STEP_FAILED;
+	p->failreason = tal_steal(p, reason);
```
rustyrussell, line 1243: Stealing without warning the caller is an error magnet; in this case, reason could be a string literal and it should work.

~~I'd suggest tal_dup_arr(...) here and label `reason` TAKES.~~

Actually, looking at the user: make payment_fail take a fmt string instead?

lightningd/htlc_set.c
rustyrussell: > The commit in question is on the receiving end, where we should indeed be more permissive than what we send out. I propose I fixup the fuzzing in the MPP PR and leave it as is here where it is still useful since we have a single part at most here.

No, let's not encourage other implementations to break the spec.  If you disagree, then this needs to be debated on the lightning-rfc spec: please open an issue.

Meanwhile, taking into account our responsibility as joint custodians of the spec, I think we need to be strict (unless it's proven that we need to allow this).

plugins/libplugin-pay.c
```diff
@@ -1943,3 +1943,94 @@ static struct direct_pay_data *direct_pay_init(struct payment *p)
...
+   	 * * [`u32`:`height`]
+	 *
+	 */
+	type = fromwire_u16(&raw_message, &raw_message_len); /* type */
+	if (type != WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS)
+		return 0;
+
+	(void) fromwire_u64(&raw_message, &raw_message_len); /* htlc_msat */
+
+	return fromwire_u32(&raw_message, &raw_message_len); /* height */
```
rustyrussell, line 1992: Why open-code  fromwire_incorrect_or_unknown_payment_details?

plugins/libplugin-pay.c
```diff
@@ -1943,3 +1943,94 @@ static struct direct_pay_data *direct_pay_init(struct payment *p)
...
+		return payment_continue(p);
+
+	if (failcode == 17)
+		blockheight = p->start_block + 1;
+	else
+		blockheight = get_remote_block_height(raw_message);
+
+	/* If we were unable to parse the target blockheight just continue. */
+	if (blockheight == 0)
+		return payment_continue(p);
```
rustyrussell, line 2028: How about:

```
if (failcode != 17 /* Former final_expiry_too_soon */)
    blockheight = p->start_block + 1;
else {
    /* If it's incorrect_or_unknown_payment_details, that tells us what height they're at */
    struct amount_msat unused;
    const void *ptr = raw_message;
    if (!fromwire_incorrect_or_unknown_payment_details(ptr, &unused, &blockheight))
        return payment_continue(p);
}
```

plugins/libplugin-pay.c
```diff
@@ -1943,3 +1943,94 @@ static struct direct_pay_data *direct_pay_init(struct payment *p)
...
+
+	plugin_log(p->plugin, LOG_INFORM,
+		   "Remote node appears to be on a longer chain, which causes "
+		   "CLTV timeouts to be incorrect. Waiting up to %" PRIu64
+		   " seconds to catch up to block %d before retrying.",
+		   time_to_sec(remaining), blockheight);
+
+	assert(blockheight >= p->start_block);
```
rustyrussell, line 2043: BOOM!

Untrusted data...

plugins/libplugin-pay.c
```diff
@@ -2018,6 +2018,13 @@ static void waitblockheight_cb(void *d, struct payment *p)
 	if (blockheight == 0)
 		return payment_continue(p);
 
+	/* If we are already at the desired blockheight there is no point in
+	 * waiting, and it is likely just some other error. Notice that
+	 * start_block gets set by the initial getinfo call for each
+	 * attempt.*/
+	if (blockheight <= p->start_block)
+		return payment_continue(p);
```
rustyrussell, line 2035: Note that this subsumes the previous == 0 check, too....

plugins/libplugin-pay.c
```diff
@@ -1606,3 +1607,170 @@ static void exemptfee_cb(struct exemptfee_data *d, struct payment *p)
...
+
+			best = tal_dup(tmpctx, struct route_info, &curr);
+			best_fee = fee;
+			sample = v;
+		}
+	}
+
+	if (best != NULL) {
+		bool ok;
+		/* Ok, we found an extension, let's add it and roll the dice again. */
```
rustyrussell, line 1792: s/and roll the dice again/ since that's done after this code block?

plugins/libplugin-pay.c
```diff
@@ -1606,3 +1607,170 @@ static void exemptfee_cb(struct exemptfee_data *d, struct payment *p)
...
+		 * modifiers don't do it either. */
+		ok &= payment_constraints_update(&p->constraints, best_fee,
+						 best->cltv_expiry_delta);
+
+		/* And now the thing that caused all of this: adjust the call
+		 * to getroute. */
+		ok &= amount_msat_add(&p->getroute->amount, p->getroute->amount,
+				      best_fee);
+		p->getroute->cltv += best->cltv_expiry_delta;
+		assert(ok);
```
rustyrussell, line 1808: Hmm, this seems like if we the shadow route gets too long, we assert()?  Shouldn't we stop (set best = NULL and fall through?).

rustyrussell: Also, what happens if our shadow route makes our onion too large?

plugins/pay.c
cdecker: Went through the commits and removed all explicit size arguments.

plugins/libplugin-pay.c
cdecker: I added an early escape to continue with the next iteration if `payment_update_constraints` on the modifier (which is guaranteed to have 1/4th of the overall constraints). If that update fails we've exhausted our shadow budget and need to continue right away. Since the first constraint update succeeded (which has <= 1/4th of the overall constraints) we're guarded here, and the assert is just here to express that logic.

The onion is not touched at all by the shadow route (they are just amount and cltv offsets to the `getroute` calls, which then is what gets turned into the onion). So we could technically have an arbitrarily long shadow route, and the onion would not explode :-)

lightningd/htlc_set.c
cdecker: I removed the offending commit, so now we're following the RFC to the letter again :-)

plugins/libplugin-pay.c
rustyrussell: Good point!

rustyrussell:
Ack f45753e69aae640e6e1cd4459d71fcd67983a831

cdecker:
Thanks @rustyrussell, still looking into two failing tests, but will merge asap.

cdecker:
Rebased on top of `master`

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/b606f46cfbb5321cef4c3ad2e214dfa104980cf6

PR #3774, opened 2020-06-16
pyln.proto: fix package list in setup.py

rustyrussell:


rustyrussell:
Closing, we should use a separate setup.py for each of these, so the versions can be different.

PR #3777, opened 2020-06-18
pyln minor fixes, move bolts to pyln.spec.bolt*

### LINKED ISSUE #3767: Missing pyln/proto/message/bolts in pyln-proto
While testing the release target in the `pyln-proto` `Makefile` I noticed that it fails with a missing package.

The `setup.py` specifies that there should be a package, but it is missing. Seems https://github.com/ElementsProject/lightning/pull/3733 just dropped it on the floor :-)

Probably @rustyrussell just needs to add the files he already has :-)

rustyrussell:
Fixes #3767


contrib/pyln-spec/bolt1/setup.py
```diff
@@ -0,0 +1,23 @@
...
+          author='Rusty Russell',
+          author_email='rusty@rustcorp.com.au',
+          license='MIT',
+          packages=['pyln.spec.bolt{}'.format(boltnum)],
+          scripts=[],
+          zip_safe=True,
+          install_requires=requirements)
+
+
+do_setup(1, __version__, desc)
```
cdecker: nit: The indirection through a `do_setup` function that we call is unnecessary as far as I can see.

contrib/pyln-spec/requirements.txt
```diff
@@ -1 +1 @@
-pyln.proto.message
+pyln.proto==0.8.4
```
cdecker, line 1: This version doesn't exist on the test or prod PyPI, so installing this package with this requirement would fail.

contrib/pyln-spec/bolt1/setup.py
rustyrussell: I originally had that one level up, and used symlinks.  I would like to restore that, but it had other problems.

contrib/pyln-spec/requirements.txt
rustyrussell: Yes, but it will once this is merged :)

It's a real dependency, since this needs the latest message module.

rustyrussell:
I really do think we want individual packages for the bolts.  They will change fast, but this way they only change if something actually changes, not because of an unrelated repository change.

Perhaps an even better option is to number like so: 1.0.<csvversion>.<gitversion>, and bump csvversion only when the csv changes.  That allows you to specify ==1.0.42.* if you don't care about any textual changes?

rustyrussell:
OK, since it can't fetch pyln.proto 0.8.3 from the real repo yet, I uploaded it to the test repo, and applied the following workaround for testing (downloads previous version to get all the deps, then gets the latest from testing repo):

```patch
diff --git a/contrib/pyln-spec/Makefile b/contrib/pyln-spec/Makefile
index 95b00fc22..081c6443d 100755
--- a/contrib/pyln-spec/Makefile
+++ b/contrib/pyln-spec/Makefile
@@ -58,7 +58,8 @@ test-release-bolt%: $(ARTEFACTS)
        # pyln-proto already installed).
        virtualenv testpypi-$* --python=/usr/bin/python3 --download --always-copy --clear
        # Install the requirements from the prod repo, they are not being kept up to date on the test repo
-       testpypi-$*/bin/python3 -m pip install -r requirements.txt pytest flaky pytest-timeout
+       testpypi-$*/bin/python3 -m pip install pyln.proto pytest flaky pytest-timeout
+       testpypi-$*/bin/python3 -m pip install -I --index-url https://test.pypi.org/simple/ --no-deps pyln.proto
        testpypi-$*/bin/python3 -m pip install -I --index-url https://test.pypi.org/simple/ --no-deps pyln-bolt$*
        testpypi-$*/bin/python3 -c "from pyln.spec import bolt$* as bolt;assert(bolt.__version__ == '$(call version,$*)')"
        testpypi-$*/bin/pytest bolt$*/tests
```

rustyrussell:
Trivial rebase to fix a typo and an accidental libwally revert!

contrib/pyln-spec/bolt1/setup.py
cdecker: Ok, we can revisit this later (or not, since this is really nothing more than a nit).

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3777/commits/39a2a9d3bc2d609609f045c3c42932489e3ee23f

Issue #3757, opened 2020-05-31
Taken pointer in lightning_gossipd triggered by sendinvoiceless plugin


sumBTC:
### Issue and Steps to Reproduce

See: https://bitkoins.nl/clightning/out.txt
See: https://bitkoins.nl/clightning/gossip_store.corrupt

EDIT: `lightning_gossipd: Outstanding taken pointers: (nil)`
doesn't sound good to me ;-)

### `getinfo` output

"version": "v0.8.2-178-g8d98b99"

cdecker:
This is very strange indeed. The code that is reporting an error is this:

https://github.com/ElementsProject/lightning/blob/8d98b99bde55ad0844761b584311543c9051b6f7/common/daemon.c#L81-L92

(permalink to the commit you reported, but it is unchanged since then)

The code tests whether we have any taken pointers, assigning to `t`, checks whether `t` is `NULL`, and if not it aborts with an error message. From the error message it seems that the pointer is `NULL` / `(nil)`, so it should not have triggered the abort, but it does.

I have no idea how that could happen. Paging Dr @rustyrussell :-)

ZmnSCPxj:
It could be `taken_any` returns a non-`NULL` pointer to the string `"(nil)"`.  So possibly we may need to look inside at `taken_any` as well.

ZmnSCPxj:
Looking at `taken_any`, it does a `sprintf(buffer, "%p", takenarr[0])`.    So it could indeed end up pointing to a string `"(nil)"`, if `takenarr[0]` is NULL, `num_taken` is not 0, and there is no `labelarr`.  This implies somebody did a `take` on a `NULL`, I think.

sumBTC:
Haha, indeed:

```
#include <stdio.h>

int main() {
  const char *t;

  t = "nil";

  if (t)
    printf("Outstanding taken pointers: %s", t);

  return 0;
}
```

sumBTC:
By the way, [here](https://Bitkoins.nl/clightning/out.txt_crash_2) is another one from some time ago which maybe gives a little more information after the "(nil)".

ZmnSCPxj:
A part that might be the culprit:

https://github.com/ElementsProject/lightning/blob/5db69f1b414f04c204f086424c96d10c20a8be39/gossipd/gossipd.c#L1499

The `channel_update` is loaded from this:

[...]

PR #3781, opened 2020-06-22
common/utils.c: Correctly handle NULL `take`n pointer in `tal_dup_talarr`.

ZmnSCPxj:
Fixes: #3757

Reported-by: @sumBTC

### DIFF
--- a/common/utils.c
+++ b/common/utils.c
@@ -142,9 +142,12 @@ void tal_arr_remove_(void *p, size_t elemsize, size_t n)
     tal_resize((char **)p, len - elemsize);
 }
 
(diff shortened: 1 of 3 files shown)


ZmnSCPxj:
Added a test which seems to replicate the original bug report.  Let us see if it fixes the problem.

ZmnSCPxj:
Modified the test somewhat as it was still failing, but not due to the bug, after the fix.

Here is a run of the test before the fix is added, note that the test is marked `xfail` at this point so passing means it actually failed (which is expected because the bug is not fixed yet): https://travis-ci.org/github/ZmnSCPxj/lightning/builds/701091542

PR #3782, opened 2020-06-22
wire: update to latest version of the spec.

rustyrussell:
The main change here is that the previously-optional open/accept
fields and reestablish fields are now compulsory (everyone was
including them anyway).  In fact, the open/accept is a TLV
because it was actually the same format.

For more details, see lightning-rfc/f068dd0d8dfa5ae75feedd99f269e23be4777381


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3782/commits/6530f9f25eb508be9da75c8aa922be6c2bf238c8

PR #3783, opened 2020-06-23
pre-work to get elements PSBTs to work

niftynei:
two bugs and a few regtest test fixes

Changelog-None


bitcoin/tx_parts.c
```diff
@@ -181,29 +181,48 @@ static struct wally_tx_input *fromwire_wally_tx_input(const tal_t *ctx,
 		blinding_nonce = fromwire_tal_arrn(tmpctx,
 						   cursor, max,
 						   fromwire_u32(cursor, max));
+
+		/* libwally doesn't like non-NULL ptrs with zero lengths. */
```
rustyrussell, line 185: I wonder if we should just make fromwire_tal_arrn return NULL for zero length, instead.  We do that in our generated code already...

cdecker: ~That'd be a bit strange if you ask me: when I get an array of `u8` for example I expect to be able to `tal_arr_expand` or `tal_resize` it, which doesn't work for `NULL` (because `tal_count` on `NULL` will likely not work).~

Edit: After checking with `tal.c` it appears that `tal_bytelen(NULL)` and `tal_count(NULL)` do the right thing :tm: :+1:

cdecker:
Oh, and we have a tiny `check-source` issue:

```
Include(s) from bitcoin/block.h duplicated in bitcoin/block.c:
961#include <ccan/tal/tal.h>
962
963Makefile:355: recipe for target 'check-includes' failed
964make: *** [check-includes] Error 1
```

niftynei:
> Oh, and we have a tiny check-source issue:

Yeah I thought I had pushed up a fix for this a while back, I'll see what happened to it

cdecker:
Seems we now have a new crash error, I'll see what's causing this.

rustyrussell:
Rebased on master, and added simple fix to avoid tal_resize() on NULL:

```patch
diff --git a/common/features.c b/common/features.c
index 375196e5d..ff193c203 100644
--- a/common/features.c
+++ b/common/features.c
@@ -195,8 +195,9 @@ u8 *get_agreed_channelfeatures(const tal_t *ctx,
                max_len = (i / 8) + 1;
        }
 
-       /* Trim to length */
-       tal_resize(&f, max_len);
+       /* Trim to length (unless it's already NULL). */
+       if (f)
+               tal_resize(&f, max_len);
        return f;
 }
 
```

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3783/commits/0f6c90b717dc8d7f1d8410510df52020aaa0cd10

Issue #3785, opened 2020-06-23
`fundchannel_cancel` `channel_id` infelicities


ZmnSCPxj:
Let me describe the current infelicity:

Supposedly `fundchannel_cancel` has an optional `channel_id` parameter.

If you have a `fundchannel_start` but have *not* yet done `fundchannel_complete` with a peer, then you can just give the peer ID without giving the `channel_id` at all.  In fact `channel_id` is ignored, you can literally put valid JSON garbage into it and it will not affect the operation of `fundchannel_cancel` at all.

    ---[fundchannel_start $peer]-->[fundchannel_cancel $peer "cat /dev/random"]-->[fundchannel_complete $peer]

On the other hand, after you have done `fundchannel_complete`, you *should* give a `channel_id`:

    ---[fundchannel_start $peer]-->[fundchannel_complete $peer]-->[fundchannel_cancel $peer $channel]

(actually if you have not recently closed a channel with that peer you *do not* need the channel ID, because `fundchannel_cancel` will only fail if there is ambiguity, but if you are trying to make a robust plugin that implements `multifundchannel` you *have to* give the channel ID, because you cannot be sure of the previous history with that peer, and it is far simpler (==robuster) to just always give the channel ID all the time than check if we are still remembering a recently-closed channel with that peer)

This translates to a kind of inconsistent use of `fundchannel_cancel`: you have to give a channel ID after `fundchannel_complete`, but you cannot get a channel ID from `fundchannel_start` --- a channel ID is derived from inputs to `fundchannel_complete`, which is itself derived from outputs of `fundchannel_start`.

--

So, to get a consistent usage of `fundchannel_cancel`, we can do a variety of things.

I am not certain what is the best, so let me present the two options I can think of for now.

* Embrace our one-open-channel-with-each-peer nature.  `fundchannel_cancel` removes its `channel_id` parameter as unnecessary.
  * If `fundchannel_cancel` is done after `fundchannel_complete`, it autosearches for the channel that is currently `CHANNELD_AWAITING_LOCKIN`; because there can only be one active channel at a time, only up to one channel can be in `CHANNELD_AWAITING_LOCKIN`.
[...]

PR #3787, opened 2020-06-24
lightningd/opening_control.c: `fundchannel_cancel` no longer requires a `channel_id` argument.

ZmnSCPxj:
Fixes: #3785

Changelog-Changed: `fundchannel_cancel` no longer requires its undocumented `channel_id` argument after `fundchannel_complete`.

Note: this implies changes in #3763.

### DIFF
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -752,7 +752,8 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
 
 	if (txout != NULL) {
 		for (size_t i = 0; i < tal_count(cancel->forgets); i++)
-			was_pending(command_fail(cancel->forgets[i], LIGHTNINGD,
+			was_pending(command_fail(cancel->forgets[i],
+				    FUNDING_CANCEL_NOT_SAFE,
 				    "The funding transaction has been broadcast, "
 				    "please consider `close` or `dev-fail`! "));
 		tal_free(cancel->forgets);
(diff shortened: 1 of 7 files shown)


cdecker:
> Note: this implies changes in #3763.

How is the dependency on #3763? Is this assuming #3763 to be merged or would merging this cause incompatibilities in #3767 (and thus requiring that to be amended)?

Just wondering about the order in which to review / merge.

ZmnSCPxj:
The sequence `fundchannel_start` `fundchannel_complete` `fundchannel_cancel` requires that the cancel command have a `channel_id` argument, because if there are recently-closed channels that have not yet passed the 100-block limit at which we forget them, `fundchannel_cancel` will fail with a message "Multiple channels".  We only *need* to  provide `channel_id` if there are recently-closed channels, but detecting those is strictly more complex.  So #3763 just always provide `channel_id` after `fundchannel_complete`.

Since this PR completely removes `channel_id`, the `fundchannel_cancel` command execution in #3763 will fail with an unknown keyword (or at least I think it should, do we not fail on seeing an unknown keyword on params?).

I am wary of retaining the `channel_id` parameter in `deprecated_apis` since it is not a documented parameter anyway, and it only applies for `multifundchannel`, because the argument only is needed after a *successful* `fundchannel_complete`, and a single-peer `fundchannel` will never fail after a successful `fundchannel_complete` (only `multifundchannel` could, if one of the other peers failed `fundchannel_complete` but this one did, we need to still cancel it even after success).

We can merge in any order, though if we merge one, the other has to be modified before merging.

cdecker:
Hm, I'm not too familiar with the `fundchannel_*` semantics, I think @niftynei can say more about the internals and whether it makes sense to remove the `channel_id`, or whether we should just document it properly.

ZmnSCPxj:
Well, `fundchannel_start` implicitly starts an uncommitted channel with the peer.  This uncommitted channel has no channel ID (it has a temporary one that `fundchannel_start` automatically creates, the `temporary_channel_id` described in BOLT2 `open_channel`/`accept_channel`/`funding_created`, but `fundchannel_start` does not return it).  So when cancelling between `fundchannel_start` and `fundchannel_complete`, the `channel_id` argument to `fundchannel_cancel` makes no sense (you can literally put a garbage string into it).  But after `fundchannel_complete`, you *have to* provide the `channel_id` otherwise previous closed-but-not-forgotten channels will trigger the "Multiple channels" error, yet currently we only support one open channel per peer anyway, so "Multiple channels" makes no sense.

ZmnSCPxj:
Rebased, because github says so.

ZmnSCPxj:
@niftynei bump!!

lightningd/channel_control.c
```diff
@@ -767,44 +767,34 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
...
+		 * ever implement multiple channels per peer.
+		 */
+		if (channel->state != CHANNELD_AWAITING_LOCKIN)
+			continue;
+		cancel_channel = channel;
+		break;
 	}
+	if (!cancel_channel)
+		return command_fail(cmd, LIGHTNINGD,
+				    "No channels matching that peer_id");
```
niftynei, line 794:
```suggestion
				    "No channels awaiting lock-in for peer_id %s", ...);
```

There might be a channel, but in the wrong state.

ZmnSCPxj: Since `fundchannel_cancel` can cancel not just channels awaiting lockin, but also uncommitted channels (i.e. `OPENGIND` state), I reword instead: "No channels being opened or awaiting lock-in for peer_id".

ZmnSCPxj:
I also took the liberty of giving a specific error code for `fundchannel_cancel` refusing to cancel if the channel is already too advanced to safely cancel (i.e. funding tx known to be broadcast, peer initiated channel, channel has HTLCs).

Issue #3731, opened 2020-05-16
Dynamic plugins don't get any options.


fiatjaf:
If a plugin is restarted with `plugin stop` then `plugin start` it gets empty `"options"` in the `init` method call, doesn't matter if the options exist or if they are required.

I tried to circumvent this by calling `listconfigs` and trying to pick up the options that should be on `init` from there, but they had vanished from that list too after `plugin stop`.

cdecker:
Kind of hard to say that this isn't the intended behavior: we free all resources that were used by a plugin if the plugin exits or is stopped by us, this includes the options.

Command line options don't really mesh well with dynamic plugins at all, and I'm tempted to make non-dynamic the default, so developers need to opt-in to the more complex dynamic behavior. FWIW it's not that complicated to implement a shim that acts like a normal (non-dynamic) plugin towards `lightningd` but allows restarting the underlying plugin, while keeping the configuration including the options active (and more importantly buffering notifications and hook calls). You can find an example of how this works here: https://github.com/lightningd/plugins/tree/master/autoreload

Another trick that the `autoreload` plugin does is to look at the parent process' command line and extract the options from there. That's rather an extreme workaround, but it works.

darosior:
I don't see this as an unexpected behaviour. It seems to me to be natural for a non startup-only plugin not to *rely* on a startup option.  
In addition, we don't allow startup-only (static) plugins to be stop and we don't allow unrecognized startup options :
```
$ lightningd --aaaa
lightningd: --aaaa: unrecognized option
```
So I think it's (most of the time) bad design for a dynamic plugin to use startup options which are maybe here, maybe not.

> I tried to circumvent this by calling listconfigs and trying to pick up the options that should be on init from there, but they had vanished from that list too after plugin stop.

We free the dangling option when ,on startup-only plugins (the default) are `plugin stop`ed.

fiatjaf:
Ok, this may not be unexpected, but it certainly is undesirable!

darosior:
Why is it undesirable ? Why cannot you use a command line parameter in place of the option ?

I can see the autoclean plugin as a counter-example but it can probably be made a static plugin.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

fiatjaf:
So if you pass a parameter on the command line when starting `lightningd` it will be persisted and always accessible from dynamic plugins?

I thought the cli parameters and the config file were treated as exactly the same.

[...]

PR #3788, opened 2020-06-24
plugins: make the autoclean plugin static

darosior:
The autocleaning will only happen if the autocleaninvoice-cycle startup
option is passed, which cannot happen if the plugin is started
post-startup.
Thus, it's less misleading for users to restrict its usage to startup.

Changelog-Added: plugins: The `autoclean` plugin is now static (you cannot manage it with the `plugin` RPC command anymore).

Related to  #3731.

### DIFF
--- a/plugins/autoclean.c
+++ b/plugins/autoclean.c
@@ -88,7 +88,7 @@ static const struct plugin_command commands[] = { {
 int main(int argc, char *argv[])
 {
 	setup_locale();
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3788/commits/13c3e75e3985449084210c3f685dcd3346605fc9

PR #3789, opened 2020-06-25
Add a 'reservation expiration' mechanism

niftynei:
It's useful to be able to 'retire' reservations on utxos after a certain time lapse. It's also useful to be able to RBF transactions. This PR lets us do both now, by adding two new parameters to `reserveinputs`. 

- `expire_after` which will functionally 'unreserve' a utxo after that many blocks have passed, and
- `allow_rbf` which if set will create a psbt that has all of its inputs set to be rbf'able. 

RBF'ability forced a few not so nice state breaks in the output status (you have to be able to select utxos that are marked as 'spent' but not in a block yet, etc). But hey, RBF, here it is!

Built ontop of the as of yet unmerged #3775 (new commits start at https://github.com/ElementsProject/lightning/commit/37ba229563002e2962e0d18348c2d6509f1895fa)

### DIFF
--- a/common/wallet_tx.c
+++ b/common/wallet_tx.c
@@ -91,11 +91,6 @@ struct command_result *param_utxos(struct command *cmd,
 	tal_free(txids);
 	tal_free(outnums);
 
-	if (!*utxos)
-		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
-							"Could not decode all of the outpoints. The utxos"
-							" should be specified as an array of "
-							" 'txid:output_index'.");
 	if (tal_count(*utxos) == 0)
 		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 							"No matching utxo was found from the wallet. "
--- a/common/withdraw_tx.c
+++ b/common/withdraw_tx.c
@@ -12,6 +12,7 @@
 
 struct bitcoin_tx *withdraw_tx(const tal_t *ctx,
 			       const struct chainparams *chainparams,
+			       bool allow_rbf,
 			       const struct utxo **utxos,
 			       struct bitcoin_tx_output **outputs,
 			       const struct ext_key *bip32_base,
@@ -19,10 +20,21 @@ struct bitcoin_tx *withdraw_tx(const tal_t *ctx,
 {
 	struct bitcoin_tx *tx;
 	int output_count;
+	/*
+	 * BIP-125: Opt-in Full Replace-by-Fee Signaling
+	 *     https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki
+	 * A transaction is considered to have opted in to
+	 * allowing replacement of itself if any of its
+	 * inputs have an nSequence number less than (0xffffffff - 1).
+	 */
+	/* A sequence of (0xffffffff - 1) signals to use locktime */
+	u32 sequence = BITCOIN_TX_DEFAULT_SEQUENCE - 1;
+	if (allow_rbf)
+		sequence--;
 
 	tx = tx_spending_utxos(ctx, chainparams, utxos, bip32_base,
 			       false, tal_count(outputs), nlocktime,
-			       BITCOIN_TX_DEFAULT_SEQUENCE - 1);
+			       sequence);
 
 	output_count = bitcoin_tx_add_multi_outputs(tx, outputs);
 	assert(output_count == tal_count(outputs));
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -571,6 +571,20 @@ def test_reserveinputs(node_factory, bitcoind, chainparams):
     outputs = l1.rpc.listfunds()['outputs']
     assert len([x for x in outputs if not x['reserved']]) == 12
 
+    # try setting the default reserved to 5 blocks
+    reserved = l1.rpc.reserveinputs([{addr: 'all'}], expire_after=5)
+    bitcoind.generate_block(5)
+    sync_blockheight(bitcoind, [l1])
+    outputs = l1.rpc.listfunds()['outputs']
+    assert len([x for x in outputs if not x['reserved']]) == 12
+
+    # try turning off the default reservation
+    reserved = l1.rpc.reserveinputs([{addr: 'all'}], expire_after=0)
+    bitcoind.generate_block(144)
+    sync_blockheight(bitcoind, [l1])
+    outputs = l1.rpc.listfunds()['outputs']
+    assert len([x for x in outputs if x['reserved']]) == 12
+
 
 def test_sign_and_send_psbt(node_factory, bitcoind, chainparams):
     """
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -586,6 +586,59 @@ def test_reserveinputs(node_factory, bitcoind, chainparams):
     assert len([x for x in outputs if x['reserved']]) == 12
 
 
+def test_reserve_rbf(node_factory, bitcoind, chainparams):
+    """
+    We now default everything to RBF. I'm not really sure how to test this
+    """
+    amount = 1000000
+    total_outs = 12
+    coin_mvt_plugin = os.path.join(os.getcwd(), 'tests/plugins/coin_movements.py')
+    l1 = node_factory.get_node(options={'plugin': coin_mvt_plugin},
+                               feerates=(7500, 7500, 7500, 7500))
+    l2 = node_factory.get_node()
+    addr = chainparams['example_addr']
+
+    # Add a medley of funds to withdraw later, bech32 + p2sh-p2wpkh
+    for i in range(total_outs // 2):
+        bitcoind.rpc.sendtoaddress(l1.rpc.newaddr()['bech32'],
+                                   amount / 10**8)
+        bitcoind.rpc.sendtoaddress(l1.rpc.newaddr('p2sh-segwit')['p2sh-segwit'],
+                                   amount / 10**8)
+    bitcoind.generate_block(1)
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == total_outs)
+
+    # Make a PSBT out of our inputs
+    reserved = l1.rpc.reserveinputs(outputs=[{addr: Millisatoshi(3 * amount * 1000)}])
+    assert len([x for x in l1.rpc.listfunds()['outputs'] if x['reserved']]) == 4
+
+    signed_psbt = l1.rpc.signpsbt(reserved['psbt'])['signed_psbt']
+    broadcast_tx = l1.rpc.sendpsbt(signed_psbt)
+    assert bitcoind.rpc.getmempoolinfo()['size'] == 1
+
+    # Try to do it again?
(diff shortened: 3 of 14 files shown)


wallet/walletrpc.c
```diff
@@ -1210,9 +1210,9 @@ static struct command_result *json_reserveinputs(struct command *cmd,
 	 * around, so we remove the auto-cleanup that happens
 	 * when the utxo objects are free'd */
 	wallet_persist_utxo_reservation(cmd->ld->wallet, utx->wtx->utxos);
-	/* Set the lease on these utxos to expire in ~4 hours,
+	/* Set the lease on these utxos to expire in ~24 hours,
 	 * or 24 blocks */
-	u32 expires_at = cmd->ld->topology->tip->height + 6 * 4;
+	u32 expires_at = cmd->ld->topology->tip->height + 6 * 24;
```
cdecker: The commit message needs an update if we change this default value.

wallet/walletrpc.c
```diff
@@ -160,6 +160,147 @@ static struct command_result *broadcast_and_wait(struct command *cmd,
...
+	if (result)
+		return result;
+
+	/* Because of the max limit of AMOUNT_SAT(-1ULL),
+	 * `(*utx)->wtx->all_funds` won't change in `wtx_select_utxos()` */
+	if ((*utx)->wtx->all_funds)
+		outputs[0]->amount = (*utx)->wtx->amount;
+
+	/* Add the change as the last output */
+	if (!amount_sat_eq((*utx)->wtx->change, AMOUNT_SAT(0))) {
```
cdecker, line 271: Does wally not set `wtx->change` if it is below dust? If that's not the case we should probably compare against the dust limit in `chainparams`.

wallet/walletrpc.c
```diff
@@ -160,6 +160,147 @@ static struct command_result *broadcast_and_wait(struct command *cmd,
...
+		struct bitcoin_tx_output *change_output;
+
+		changekey = tal(tmpctx, struct pubkey);
+		if (!bip32_pubkey(cmd->ld->wallet->bip32_base, changekey,
+				  (*utx)->wtx->change_key_index))
+			return command_fail(cmd, LIGHTNINGD, "Keys generation failure");
```
cdecker, line 277:
```suggestion
			return command_fail(cmd, LIGHTNINGD, "Change key generation failed");
```

wallet/walletrpc.c
```diff
@@ -1243,14 +1243,15 @@ static struct command_result *json_reserveinputs(struct command *cmd,
 	struct unreleased_tx *utx;
 	const struct utxo **chosen_utxos;
 	struct bitcoin_tx_output **outputs;
-	u32 *feerate_per_kw, *minconf;
+	u32 *feerate_per_kw, *minconf, *expire_at;
```
cdecker, line 1248: Just to match the param name:

```suggestion
	u32 *feerate_per_kw, *minconf, *expire_after;
```

wallet/walletrpc.c
```diff
@@ -1275,13 +1276,12 @@ static struct command_result *json_reserveinputs(struct command *cmd,
...
-	u32 expires_at = cmd->ld->topology->tip->height + 6 * 24;
-	for (size_t i = 0; i < tal_count(utx->wtx->utxos); i++)
-		wallet_output_reservation_update(cmd->ld->wallet,
-						 utx->wtx->utxos[i],
-						 expires_at);
+	if (*expire_at > 0) {
+		for (size_t i = 0; i < tal_count(utx->wtx->utxos); i++)
+			wallet_output_reservation_update(cmd->ld->wallet,
+							 utx->wtx->utxos[i],
+							 *expire_at);
```
cdecker, line 1287: Seems to me like this uses the relative expiry as absolute blockheight. Should probably be something like the following:

```suggestion
		for (size_t i = 0; i < tal_count(utx->wtx->utxos); i++)
			wallet_output_reservation_update(cmd->ld->wallet,
							 utx->wtx->utxos[i],
							 cmd->ld->topology->tip->height + *expire_at);
```

common/withdraw_tx.c
```diff
@@ -12,17 +12,29 @@
...
+	 * BIP-125: Opt-in Full Replace-by-Fee Signaling
+	 *     https://github.com/bitcoin/bips/blob/master/bip-0125.mediawiki
+	 * A transaction is considered to have opted in to
+	 * allowing replacement of itself if any of its
+	 * inputs have an nSequence number less than (0xffffffff - 1).
+	 */
+	/* A sequence of (0xffffffff - 1) signals to use locktime */
+	u32 sequence = BITCOIN_TX_DEFAULT_SEQUENCE - 1;
+	if (allow_rbf)
+		sequence--;
```
cdecker, line 33: Isn't this now `0xffffffff - 2` because `BITCOIN_TX_DEFAULT_SEQUENCE` is `0xffffffff` and we subtract twice? That'd mean we always opt in.

common/wallet_tx.c
```diff
@@ -91,11 +91,6 @@ struct command_result *param_utxos(struct command *cmd,
 	tal_free(txids);
 	tal_free(outnums);
 
-	if (!*utxos)
-		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
-							"Could not decode all of the outpoints. The utxos"
-							" should be specified as an array of "
-							" 'txid:output_index'.");
```
cdecker, line 98: This looks like an unrelated change.

tests/test_wallet.py
```diff
@@ -586,6 +586,59 @@ def test_reserveinputs(node_factory, bitcoind, chainparams):
...
+    psbt = bitcoind.rpc.decodepsbt(reserved['psbt'])
+    utxos = []
+    unreserve_utxos = []
+    for vin in psbt['tx']['vin']:
+        utxos.append(vin['txid'] + ":" + str(vin['vout']))
+        unreserve_utxos.append({'txid': vin['txid'], 'vout': vin['vout'], 'sequence': vin['sequence']})
+    unreserve_psbt = bitcoind.rpc.createpsbt(unreserve_utxos, [])
+    unres = l1.rpc.unreserveinputs(unreserve_psbt)
+
+    # let's not set the fee high enough, should explodes
```
cdecker, line 628:
```suggestion
    # let's not set the fee high enough, should explode
```

tests/test_wallet.py
```diff
@@ -586,6 +586,59 @@ def test_reserveinputs(node_factory, bitcoind, chainparams):
...
+    signed_psbt = l1.rpc.signpsbt(reserved['psbt'])['signed_psbt']
+    with pytest.raises(RpcError, match=r'insufficient fee, rejecting replacement'):
+        l1.rpc.sendpsbt(signed_psbt)
+
+    # now let's do it for reals
+    unres = l1.rpc.unreserveinputs(unreserve_psbt)
+    reserved = l1.rpc.reserveinputs([{addr: Millisatoshi(amount * 0.5 * 1000)}], feerate='10000perkw', utxos=utxos)
+    signed_psbt = l1.rpc.signpsbt(reserved['psbt'])['signed_psbt']
+    l1.rpc.sendpsbt(signed_psbt)
+    assert bitcoind.rpc.getmempoolinfo()['size'] == 1
```
cdecker, line 639: Should we check that the new tx is in the mempool? Just checking for the size of the mempool might pass despite the old tx still being in there. `bitcoind.rpc.getrawmempool()` will give you the txids.

rustyrussell:
Note: after discussing with @niftynei I agree with this concept, but she didn't take it far enough.

Because we didn't expire utxo reservations, we had multiple kludges:
1. On startup, we unreserve all of them.
2. We reserve them, but use a destructor to unreserve them.
3. We mark them *spent* before we see the transaction in the blockchain, leading to missing funds and rescan if it never confirms.

On digging further, I noticed that with a simple change, we can move utxo.h entirely into the wallet dir. This further allows us to simplify our utxo handling.

So I've cherry picked part of the "reservations expire" commit and started ripping things apart...

wallet/wallet.c
```diff
@@ -229,6 +245,79 @@ bool wallet_update_output_status(struct wallet *w,
...
+	stmt = db_prepare_v2(w->db, SQL("SELECT COUNT(*) FROM outputs "
+				"WHERE status=? AND prev_out_tx=? "
+				"AND prev_out_index=?"));
+	db_bind_int(stmt, 0, output_status_in_db(state));
+	db_bind_txid(stmt, 1, txid);
+	db_bind_int(stmt, 2, outnum);
+
+	db_query_prepared(stmt);
+	if (db_step(stmt)) {
+		res = db_column_u64(stmt, 0) > 0;
```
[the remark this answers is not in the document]
niftynei: per their [docs](https://www.postgresql.org/docs/current/functions-aggregate.html), it returns a `bigint` which is a signed 8-byte int

wallet/wallet.c
```diff
@@ -355,10 +448,89 @@ struct utxo *wallet_utxo_get(const tal_t *ctx, struct wallet *w,
...
+	struct utxo **reserved, **available;
+	size_t rcount;
+
+	reserved = expired_reserved_utxos_get(ctx, w);
+	available = wallet_get_utxos(ctx, w, output_state_available);
+
+	rcount = tal_count(reserved);
+	tal_resize(&reserved, rcount + tal_count(available));
+	for (size_t i = 0; i < tal_count(available); i++)
+		reserved[rcount + i] = tal_steal(reserved, available[i]);
```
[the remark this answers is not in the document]
niftynei: yes but then we'd have to clean them up on every new block. Another option would be to remove the 'output_state_available'/ status altogether and just use the reserved_til as a marker of whether or not the utxo is still reserved.

common/withdraw_tx.c
niftynei: Right, the intention is to opt in to RBF if 'allow_rbf' is true.

`0xffffffff` is no rbf, no locktime
`0xffffffff - 1` is no rbf, allow locktime (which we set on all withdraws, and now fundchannels)
`0xffffffff - 2` is allow rbf, allow locktime

wallet/walletrpc.c
niftynei: Ah yes, good spot.

wallet/walletrpc.c
niftynei: `wtx_select_utxos` does the dust comparison and sets this to zero if it's less than that, so this is an accurate check.

niftynei:
This will largely be subsumed by work @rustyrussell is doing on the utxo infrastructure, but updated with comments for good measure.

cdecker:
> This will largely be subsumed by work @rustyrussell is doing on the utxo infrastructure, but updated with comments for good measure.

Thanks for the heads up :+1:

ZmnSCPxj:
>> This will largely be subsumed by work @rustyrussell is doing on the utxo infrastructure, but updated with comments for good measure.
>
>Thanks for the heads up :+1:

I thought this was #3798 , which was withdrawn?  The nearest equivalent is #3821, which does not look like it has reservation expiration.  What is the status?  How does this affect #3763 which apparently need to be blocked for the new UTXO things?

PR #3794, opened 2020-06-26
build: Make it possible to build without gcc

laanwj:
The external Makefile hardcodes gcc to get the machine spec. This should use the configured C compiler instead.

This bug was introduced in 601464416b7d437409ceb8f7373115d2f14b21d1.

### DIFF
--- a/external/Makefile
+++ b/external/Makefile
@@ -9,7 +9,7 @@ ifdef BUILD
 CROSSCOMPILE_OPTS := --host="$(MAKE_HOST)" --build="$(BUILD)"
 TARGET_DIR := external/"$(MAKE_HOST)"
 else
-TARGET_DIR := external/"$(shell gcc -dumpmachine)"
+TARGET_DIR := external/"$(shell ${CC} -dumpmachine)"
 endif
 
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
Needs a `Changelog` entry.  Suggest adding to commit message:

    Changelog-Fixed: Make it possible to build `external/` even without `gcc` installed.

rustyrussell:
> Needs a `Changelog` entry. Suggest adding to commit message:
> 
> ```
> Changelog-Fixed: Make it possible to build `external/` even without `gcc` installed.
> ```
No, this wasn't broken in last release?

ZmnSCPxj:
So `Changelog-None` instead?

ZmnSCPxj:
I cannot push into @laanwj branch, so need to wait for OP to add `Changelog-None` to commit message.

rustyrussell:
> I cannot push into @laanwj branch, so need to wait for OP to add `Changelog-None` to commit message.

Putting it in the GH comment is sufficient, BTW (putting changelog-None in commits can seem weird).

laanwj:
> Needs a Changelog entry. Suggest adding to commit message:

Thanks. I'll keep it in mind for next time.

PR #3796, opened 2020-06-29
external/Makefile: fix spurious rebuilds.

rustyrussell:
Quote marks are not special to make: as it can't find
external/"x86_64-linux-gnu"/libwally-core-build/src/libwallycore.la
it always insists on rebuilding it (which rebuilds the world).

If we have spaces in TARGET_DIR, we're in trouble already.


ZmnSCPxj:
ACK 4e452517

Reapplied previous ACK.  Only change is `Changelog-None`.

PR #3797, opened 2020-06-29
wallet/walletrpc.c: `txprepare`d transactions now use current tip blockheight by default.

ZmnSCPxj:
Changelog-Changed: `txprepare` now prepares transactions whose `nLockTime` is set to the tip blockheight, instead of using 0. `fundchannel` will use `nLockTime` set to the tip blockheight as well.

### DIFF
--- a/wallet/walletrpc.c
+++ b/wallet/walletrpc.c
@@ -152,7 +152,7 @@ static struct command_result *json_prepare_tx(struct command *cmd,
 	const u8 *destination = NULL;
 	size_t out_len, i;
 	const struct utxo **chosen_utxos = NULL;
-	u32 locktime = 0;
+	u32 locktime;
 
 	*utx = tal(cmd, struct unreleased_tx);
 	(*utx)->wtx = tal(*utx, struct wallet_tx);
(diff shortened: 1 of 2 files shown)


wallet/walletrpc.c
```diff
@@ -288,8 +294,36 @@ static struct command_result *json_prepare_tx(struct command *cmd,
 			   p_opt("feerate", param_feerate, &feerate_per_kw),
 			   p_opt_def("minconf", param_number, &minconf, 1),
 			   p_opt("utxos", param_utxos, &chosen_utxos),
+			   /* Note, withdraw does not have `locktime`
+			    * parameter.
+			    */
```
cdecker, line 299: Why not? Wouldn't it make sense to have most transactions as homogeneous as possible (yes, I know we can look at the output types and get an idea, but with taproot we wouldn't need to change).

wallet/walletrpc.c
```diff
@@ -288,8 +294,36 @@ static struct command_result *json_prepare_tx(struct command *cmd,
...
-		return command_param_failed();
+			return command_param_failed();
+	}
+
+	if (locktime) {
+		struct abs_locktime b_locktime;
+		/* FIXME: Allow providing some kind of absolute time
+		 * argument or date/time format.
+		 * strptime?
+		 */
```
cdecker, line 309: No, please just keep it simple.

wallet/walletrpc.c
niftynei: `withdraw` populates the locktime, using the same logic as bitcoind's locktime selection

niftynei:
If there's not a reason for wanting control over setting it explicitly, I'd suggest re-using the logic we use for `withdraw`, which mimic's the bitcoind logic also.

https://github.com/ElementsProject/lightning/blob/master/wallet/walletrpc.c#L331-L334

ZmnSCPxj:
Well, we could use `locktime` and set it to 499999999 when doing the initial `txprepare` (the one that gets `txdiscard`ed later) in `fundchannel`/`multifundchannel`, as a belt-and-suspenders protection to prevent accidentally `txsend`ing the initial `txprepare`d transaction.  Though input reservation would do just as well (if only there was a way to atomically get reserved UTXOs into a `txprepare`....)

@niftynei yes, that code just gets shifted around in this PR.

ZmnSCPxj:
Removed `locktime` argument.

cdecker:
Need to kick Travis a bit more, but LGTM

ACK https://github.com/ElementsProject/lightning/pull/3797/commits/51c40f98e4571f127ea188e516faca508f5851be

PR #3798, opened 2020-06-29
WIP: utxo selection cleanup

rustyrussell:
This is uploaded so @niftynei can see where I'm up to...


wallet/reservation.c
```diff
@@ -0,0 +1,282 @@
...
+					cmd->ld->topology->tip->height,
+					&diff,
+					*feerate_per_kw,
+					utxos);
+		if (utxo) {
+			tal_arr_expand(&utxos, utxo);
+			continue;
+		}
+
+		/* If they said "any", we expect to run out of utxos. */
```
ZmnSCPxj, line 253: s/"any"/"all"/

wallet/reservation.c
```diff
@@ -0,0 +1,282 @@
...
+{
+	struct json_stream *response;
+	struct utxo **utxos;
+	u32 *feerate_per_kw;
+	u32 *minconf;
+	struct amount_sat *amount, diff;
+
+	if (!param(cmd, buffer, params,
+		   p_req("satoshis", param_sat_or_all, &amount),
+		   p_req("feerate", param_feerate, &feerate_per_kw),
```
ZmnSCPxj, line 232: Should we not specify as well the number and type of expected outputs?  To get a better idea of how much fees will end up being?

wallet/reservation.c
```diff
@@ -0,0 +1,282 @@
...
+				    type_to_string(tmpctx,
+						   struct amount_sat,
+						   amount),
+				    tal_count(utxos),
+				    type_to_string(tmpctx,
+						   struct amount_sat,
+						   &diff));
+	}
+
+	response = json_stream_success(cmd);
```
ZmnSCPxj, line 268: Is my understanding correct, that this command does not reserve the selected inputs?  Is it not better for it to atomically reserve the inputs as well, so that plugins do not have to do a `fundinputs`-`reserveinputs` loop?

Consider the case where two plugins need to get 0.1BTC each and the node owns two 0.2 BTC UTXOs, if we do not reserve atomically within `fundinputs`, then the plugins might end up getting the same UTXO from `fundinputs`, then reserving them with `reserveinputs`.  To be robust against this the plugins have to loop if `reserveinputs` fails, and only exit this loop if `fundinputs` fails or `reserveinputs` succeeds.

ZmnSCPxj: Yes, a more detailed look at `wallet_find_utxo` shows that if multiple parallel `fundinputs` are invoked, they can indeed return the same UTXO sets in the resulting PSBTs.  Multiple plugins asking for similar amounts can end up having overlapping/same UTXO inputs.  So I strongly suggest to make `fundinputs` do reservation as well.  Otherwise plugins need to do a livelockable loop where they do `fundinputs`-`reserveinputs` in a loop until `reserveinputs` succeeds or `fundinputs` fails.

rustyrussell:
Started again...

PR #3800, opened 2020-06-29
Simple utxo code cleanups

rustyrussell:
Changelog-None


wallet/walletrpc.c
```diff
@@ -807,9 +807,9 @@ static const struct json_command listaddrs_command = {
 };
 AUTODATA(json_command, &listaddrs_command);
 
-static struct command_result *json_outputs(struct command *cmd,
-					   struct json_stream *response,
-					   struct utxo **utxos)
+static void json_add_utxos(struct command *cmd,
```
ZmnSCPxj: Other `json_add_*` functions get `(struct json_stream *response, const char *field, const _whatever_ *value)` as arguments, would it not be better to use similar convention?

cdecker: Agreed, with the one exception that we need access to the log here, which is why `cmd` is being passed in as well. Would be better to pass in the more specific `log` instead of the `cmd` instance, which gives access to loads of stuff that this function shouldn't have access to.

tests/test_closing.py
```diff
@@ -1569,6 +1569,34 @@ def try_pay():
     assert account_balance(l2, channel_id) == 0
 
 
+def test_listfunds_after_their_unilateral(node_factory, bitcoind):
+    """We keep spending info around for their unilateral closes.
+
+Make sure we the address.
```
cdecker, line 1575: I'll buy a verb :-)

tests/test_closing.py
```diff
@@ -1569,6 +1569,34 @@ def try_pay():
...
+    """
+    coin_mvt_plugin = os.path.join(os.getcwd(), 'tests/plugins/coin_movements.py')
```
cdecker, line 1577:
```suggestion
    coin_mvt_plugin = os.path.join(os.path.dirname(__file__), 'plugins', 'coin_movements.py')
```

`os.path.join` will add the `/` as separator, having the elements contain `/` defeats the point (not all that useful since we only run on systems that have `/` as pat-separator, but nevertheless a good habit to have)

wallet/wallet.h
```diff
@@ -333,14 +333,6 @@ struct wallet_transaction {
  */
 struct wallet *wallet_new(struct lightningd *ld, struct timers *timers);
 
-/**
- * wallet_add_utxo - Register an UTXO which we (partially) own
- *
- * Add an UTXO to the set of outputs we care about.
- */
```
cdecker, line 340: Move the comment over to the implementation?

bitcoin/tx.c
```diff
@@ -726,3 +726,73 @@ wally_tx_output_get_amount(const struct wally_tx_output *output)
...
 }
+
+/* Various weights of transaction parts. */
+size_t bitcoin_tx_core_weight(size_t num_inputs, size_t num_outputs)
+{
+	size_t weight;
+
+	/* version, input count, output count, locktime */
+	weight = (4 + varint_size(num_inputs) + varint_size(num_outputs) + 4)
+		* 4;
```
niftynei, line 737: missing `1 + 1` for segwit marker + flag?

bitcoin/tx.c
```diff
@@ -726,3 +726,73 @@ wally_tx_output_get_amount(const struct wally_tx_output *output)
...
+
+	/* A couple of things need to change for elements: */
+	if (chainparams->is_elements) {
+                /* Each transaction has surjection and rangeproof (both empty
+		 * for us as long as we use unblinded L-BTC transactions). */
+		weight += 2 * 4;
+
+		/* An elements transaction has 1 additional output for fees */
+		weight += (8 + 1) * 4; /* Bitcoin style output */
```
niftynei, line 746: call `bitcoin_tx_weight_output` here instead

bitcoin/tx.c
```diff
@@ -726,3 +726,73 @@ wally_tx_output_get_amount(const struct wally_tx_output *output)
...
+
+	/* Input weight: txid + index + sequence */
+	weight = (32 + 4 + 4) * 4;
+
+	/* We always encode the length of the script, even if empty */
+	weight += 1 * 4;
+
+	/* P2SH variants include push of <0 <20-byte-key-hash>> */
+	if (p2sh)
+		weight += 23 * 4;
```
niftynei, line 788: are reedeemscripts prefixed with a length in addition to the scriptSig length? if not, this overcounts by 1 (23 is 20 + 1 pushdata + 00 + 1 len)

wallet/wallet.c
```diff
@@ -404,42 +404,20 @@ static const struct utxo **wallet_select(const tal_t *ctx, struct wallet *w,
 	const struct utxo **utxos = tal_arr(ctx, const struct utxo *, 0);
 	tal_add_destructor2(utxos, destroy_utxos, w);
 
-	/* version, input count, output count, locktime */
-	weight = (4 + 1 + 1 + 4) * 4;
-
-	/* Add segwit fields: marker + flag */
-	weight += 1 + 1;
```
niftynei, line 411: this gets left out above

tests/test_closing.py
rustyrussell: Agreed.  I cut and paste from the previous, so I'll do a separate cleanup sweep PR.

bitcoin/tx.c
rustyrussell: Agreed, good catch!  We already count the length above.

wallet/walletrpc.c
rustyrussell: Actually, this arg is replaced in the next patch with "wallet", where we do appropriate logging.

rustyrussell: But I've moved it to the end, for consistency as @ZmnSCPxj points out.

rustyrussell:
OK, rebased with fixups (and a nasty inversion in the COMPAT code: we would print out address if we *had* closeinfo, not if we *didn't have* closeinfo.

Didn't changelog, even though txs will be ever-so-slightly cheaper from now on :)

tests/test_closing.py
```diff
@@ -1569,6 +1569,34 @@ def try_pay():
...
+    l2.daemon.wait_for_log('onchaind complete, forgetting peer')
+    l1.daemon.wait_for_log('onchaind complete, forgetting peer')
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == 2)
+    assert all(['address' in o for o in l1.rpc.listfunds()['outputs']])
+
+    # Verify accounting for l1 & l2
+    assert account_balance(l1, channel_id) == 0
+    assert account_balance(l2, channel_id) == 0
+
+
```
cdecker, line 1599: CI is complaining that there is an empty line containing whitespace :-(

bitcoin/tx.c
niftynei: I went and looked this up, `23` is actually correct, as you do prefix redeemscripts with their length (in addition to prefixing the scriptSig with it's length).

Example:

`23 22 00209c9c8e1519dacba326d9c2987e15944e9ff5d19315d4af33d2511385d66e5755`

rustyrussell: Yes, you're right.  Code review FTW! (I simply moved the code).

rustyrussell:
OK, more test fixups and whitespace remove to make Travis happier.  I hope!

cdecker:
Restarted 3 flaky tests and `clang` doesn't like that `COMPAT_V072` is not defined. Coin moves now seem to be fixed.

rustyrussell:
> Restarted 3 flaky tests and `clang` doesn't like that `COMPAT_V072` is not defined. Coin moves now seem to be fixed.

Oops, COMPAT needs to be tested with #ifdef not #if like we test everything else.  Fixed, trivial rebase...

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3800/commits/8ffd1d4d030643bc4a7526dddc7b59a435ec3ebb

PR #3801, opened 2020-06-29
db: Retrieve peer ID if it exists or create the peer if not

cdecker:
We were assuming `wallet_channel_insert` that there cannot be a matching peer
if our in-memory representation isn't bound to it (`dbid == 0`). If we then
attempt to create the peer, and we already had one it'd cause a unique
constraint violation. As far as I can tell this could end up happening if we
have an uncommitted channel, and then exited without cleanup (`tal_destructor`
on the uncommitted channel not running). This could then leave the peer in the
DB. This is because the constraint that every peer has at least one channel is
not enforce at DB level, but rather in destructors that may or may not run.

I tried to reproduce the issue in pytest, but was unsuccessful, suggesting I don't
fully capture the causes yet.

Notice that an alternative to this would be adding yet another cleanup-on-start
but that feels even more heavy handed, and if implemented wrongly might 
itself result in inconsistencies, which is why I went for the load-or-create method.

Changelog-Fixed: Fixed a failing assertion if we reconnect to a peer that we had a channel with before, and then attempt to insert the peer into the DB twice.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -1535,22 +1535,47 @@ void wallet_channel_save(struct wallet *w, struct channel *chan)
 	tal_free(stmt);
 }
 
+static void wallet_peer_save(struct wallet *w, struct peer *peer)
+{
+	const char *addr =
+	    type_to_string(tmpctx, struct wireaddr_internal, &peer->addr);
+	struct db_stmt *stmt =
+	    db_prepare_v2(w->db, SQL("SELECT id FROM peers WHERE node_id = ?"));
+
+	db_bind_node_id(stmt, 0, &peer->id);
+	db_query_prepared(stmt);
+
+	if (db_step(stmt)) {
+		/* So we already knew this peer, just return its dbid */
+		peer->dbid = db_column_u64(stmt, 0);
+		tal_free(stmt);
+
+		/* Since we're at it update the wireaddr */
+		stmt = db_prepare_v2(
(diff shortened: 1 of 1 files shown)


wallet/wallet.c
```diff
@@ -1535,22 +1535,48 @@ void wallet_channel_save(struct wallet *w, struct channel *chan)
...
-		db_bind_text(stmt, 1,
-			     type_to_string(tmpctx, struct wireaddr_internal,
-					    &chan->peer->addr));
+		db_bind_node_id(stmt, 0, &peer->id);
+		db_bind_text(stmt, 1,addr);
 		db_exec_prepared_v2(stmt);
-		chan->peer->dbid = db_last_insert_id_v2(take(stmt));
+		peer->dbid = db_last_insert_id_v2(take(stmt));
 	}
+	return peer->dbid;
```
rustyrussell, line 1571: This returns the peer dbid, but nobody cares? Maybe peer_into_db is a better name (returning void)?

cdecker: Yeah, it started out as a simple fetch or store and return the DB ID, but then I noticed I have the location I want to store it in.

This actually matches the other `wallet_*_save` functions, so I renamed it, and removed the return.

cdecker:
[Range-diff here](http://46.101.246.115:8888/ElementsProject/lightning/pull/3801/range_diff/ed57b5d74ad4adacc7deac445dd7710f79e924f5..57825968406849c41fcfb1f683ba715ea8f8593c)

PR #3809, opened 2020-07-03
Implement multi-part payments (MPP) sending support

cdecker:
- We have two splitter phases: a presplit modifier which splits large amounts
   before even starting a first attempt, and an adaptive_splitter which reacts
   to new information about edges and failed payments (mainly capacity
   estimates and disabled channels)
 - The presplit modifier will split anything larger than our threshold (10ksat
   at the moment) so that's pretty much the opposite of what you're
   doing. This is derived from a recent round of probing I did (see results
   below).
 - When splitting we fuzz the amount by +/- 10% uniformly around the target
   split amount (10ksat in the presplit case, or 1/2 previous attempt in the
   adaptive splitter).
 - We track estimates and new information we learn about channels and nodes as
   channel_hints in the root of the payment tree
 - We iteratively ask for new routes using getroute for each attempt, this is
   a bit more costly than doing it in batch with Yen's algorithm, but we can
   provide up to date info on the channel state, and reduce our in-memory
   state-management during a payment.

About the capacity probes I did: I implemented a simple binary search on the
payment amount and then iterated through all nodes and tested the maximum
single-part payment that will reach the destination. After testing about 35%
of the network the histogram of successful capacities looks like this:

probed-channel-capacities

The 80th percentile is marked in black, and the 50th percentile marked in
red. With 10ksat for the pre-splitter we end up with parts that are ~1USD in
size, and that will succeed in 80% of cases without requiring further
splits. Our goal here is to be rather aggressive, not optimize for fees, but
for speed. At 10ksat the proportional fee still dominates the overall cost, so
we're confident that adding a couple more parts is not going to be too costly.

Notice that these measurements use my usual worst-case assumption of any
sender-receiver pair being equally likely, which we pretty much know is too
pessimistic. Not surprisingly the bottlenecks on the path are clustered around
the 100ksat and 500ksat sizes, which are the most popular total capacity sizes
in the network today.

We also plan to keep measuring and adapting these constants as we learn new
things and the network evolves.

Depends-On: #3826


plugins/libplugin-pay.c
```diff
@@ -2047,3 +2047,105 @@ static void waitblockheight_cb(void *d, struct payment *p)
...
+			tlvstream_set_tlv_payload_data(
+			    fields, root->payment_secret,
+			    root->amount.millisatoshis); /* Raw: onion payload */
+		}
+	} else if (p == root && p->step == PAYMENT_STEP_INITIALIZED) {
+		/* The presplitter only acts on the root and only in the first
+		 * step. */
+		/* We need to opt-in to the MPP sending facility no matter
+		 * what we do. That means setting all partids to a non-zero
+		 * value. */
```
rustyrussell, line 2146: Where do we test that the final destination supports MPP?

doc/lightningd-config.5.md
```diff
@@ -309,6 +309,14 @@ up space in the database.
...
+
+ **disable-mpp**=*BOOL*
+Disable the multi-part payment sending support in the `pay` plugin. By default
+the MPP support is enab led, but it can be desirable to disable in situations
+in which each payment should result in a single HTLC being forwarded in the
+network.
+
 ### Networking options
 
 Note that for simple setups, the implicit *autolisten* option does the
```
rustyrussell, line 322: Need to `make doc-all`?

doc/lightningd-config.5.md
```diff
@@ -309,6 +309,14 @@ up space in the database.
 Control how long invoices must have been expired before they are cleaned
 (if *autocleaninvoice-cycle* is non-zero).
 
+Payment control options:
+
+ **disable-mpp**=*BOOL*
+Disable the multi-part payment sending support in the `pay` plugin. By default
+the MPP support is enab led, but it can be desirable to disable in situations
```
rustyrussell, line 316: 'enab led'....

rustyrussell: I really hate the double-negative.  Should be just "**disable-mpp**", with no BOOL option?  Or *enable-mpp* with default true...

cdecker:
> Ack [7f78707](https://github.com/ElementsProject/lightning/commit/7f7870765f78d463a6dc732b917c4c9ca66b4279)
> 
> Just minor: please invert the disable-mpp flag...

Re-applying after fixups

ACK https://github.com/ElementsProject/lightning/pull/3809/commits/4e9054bb6977daaa72408ad9ee9a0520c064a486

PR #3810, opened 2020-07-03
Cleanups

### LINKED ISSUE #3765: Docker compose: /usr/bin/env: 'python3': No such file or directory
Hi.

I have the following c-lightning service in a docker-compose.yml file:

```
  clightning_bitcoin:
    image: elementsproject/lightningd
    container_name: testnetln
    command:
      - --network=testnet
      - --bitcoin-rpcconnect=testnetnode
      - --bitcoin-rpcuser=xxx
      - --bitcoin-rpcpassword=yyy
      - --bind-addr=0.0.0.0:9901
      - --alias=bappTestnet
      - --log-level=debug
      - --plugin=/root/.lightning/ln_testnet.py

    environment:
      EXPOSE_TCP: "true"
    expose:
      - "9901"
    ports:
      - "0.0.0.0:9901:9901"
    volumes:
      - "/media/r2d2/ssd_2/testnet/ln:/root/.lightning"
      - "/media/r2d2/ssd_2/testnet/bitcoin:/etc/bitcoin"
    links:
      - testnetnode
```

As you can see I am running a python plugin: `/root/.lightning/ln_testnet.py`

When upping the docker-compose I get error:
` /usr/bin/env: 'python3': No such file or directory`

I have seen this kind of errors when running docker on windows because EOL but I am running on linux-mint 19.2

Thanks.

UPDATE: I think the problem is that there no python3 installed on the image.
I have open a bash and tried to run "apt-get install python3" but it doesn't find the package I can't figure it out why.

cdecker:
Just my weekly cleanups. The main change is the last one where I enable `pytest-rerunfailures` which reruns any failures up to five times. Since that'd risk masking flaky tests, I also added `pytest-sentry` which will report all failures to a sentry instance, so we can track flakes and failures. This should considerably reduce the number of times we need to restart Travis.

Fixes #3765

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -10,6 +10,7 @@ export SOURCE_CHECK_ONLY=${SOURCE_CHECK_ONLY:-"false"}
 export COMPAT=${COMPAT:-1}
 export PATH=$CWD/dependencies/bin:"$HOME"/.local/bin:"$PATH"
 export PYTEST_PAR=2
+export PYTEST_SENTRY_ALWAYS_REPORT=1
 
 # If we're not in developer mode, tests spend a lot of time waiting for gossip!
 # But if we're under valgrind, we can run out of memory!
@@ -40,6 +41,10 @@ pip3 install --user -U --quiet --progress-bar off \
      -r contrib/pyln-proto/requirements.txt \
      -r contrib/pyln-testing/requirements.txt
 
+pip3 install --user -U --quiet --progress-bar off \
+     pytest-sentry \
+     pytest-rerunfailures
+
 echo "Configuration which is going to be built:"
 echo -en 'travis_fold:start:script.1\\r'
 ./configure CC="$CC"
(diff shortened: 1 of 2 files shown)


tests/test_misc.py
```diff
@@ -476,6 +476,10 @@ def is_p2wpkh(output):
     assert only_one(fundingtx['vin'])['txid'] == res['wallettxid']
 
 
+@unittest.skipIf(
+    os.environ.get('TEST_DB_PROVIDER') == 'postgres',
+    "Postgres does not maintain insertion order on queries on its own"
+)
```
rustyrussell, line 482: That's strange?  Where does this fail?  It's one thing if the test relies on ordering, but another if the actual code does.

check_coin_moves seems to read from a json file written from the notifier, so I can't see how that gets wrong order...

cdecker: You're right, it's a commit I've had for a while and all of the sudden I can't reproduce the issue anymore. Will drop that commit from the PR.

cdecker:
Dropped commit that marked `test_withdraw_misc` as skip on postgres, thanks @rustyrussell for checking :+1:

PR #3813, opened 2020-07-05
Tweak docker entrypoint to support regtest, testnet

joemphilips:
Before this, docker image will never detects that
`lightning-rpc` was created if it is running in regtest
or testnet, because the file will be created under
subfolder for each network name.
~By recursively checking the file creation, now docker
can be run in any network configuration.~
I had to deal the path for argument to the `socat` , so I added a new `ENV` entry in Dockerfile to detect the correct network.

### DIFF
--- a/tools/docker-entrypoint.sh
+++ b/tools/docker-entrypoint.sh
@@ -2,18 +2,20 @@
 
 : "${EXPOSE_TCP:=false}"
 
+networkdatadir="${LIGHTNINGD_DATA}/${LIGHTNINGD_NETWORK}"
+
 if [ "$EXPOSE_TCP" == "true" ]; then
     set -m
     lightningd "$@" &
 
     echo "C-Lightning starting"
     while read -r i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
-    < <(inotifywait  -e create,open --format '%f' --quiet "$LIGHTNINGD_DATA" --monitor)
+    < <(inotifywait -e create,open --format '%f' --quiet "${networkdatadir}" --monitor)
     echo "C-Lightning started"
     echo "C-Lightning started, RPC available on port $LIGHTNINGD_RPC_PORT"
 
-    socat "TCP4-listen:$LIGHTNINGD_RPC_PORT,fork,reuseaddr" "UNIX-CONNECT:$LIGHTNINGD_DATA/lightning-rpc" &
+    socat "TCP4-listen:$LIGHTNINGD_RPC_PORT,fork,reuseaddr" "UNIX-CONNECT:${networkdatadir}/lightning-rpc" &
     fg %-
 else
-    exec lightningd "$@"
+    exec lightningd --network="${LIGHTNINGD_NETWORK}" "$@"
 fi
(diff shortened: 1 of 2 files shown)


tools/docker-entrypoint.sh
```diff
@@ -2,18 +2,20 @@
 
 : "${EXPOSE_TCP:=false}"
 
+datadir="${LIGHTNINGD_DATA}/${LIGHTNINGD_NETWORK}"
```
ZmnSCPxj, line 5: Suggest renaming to `networkdatadir`.

tools/docker-entrypoint.sh
```diff
@@ -2,18 +2,20 @@
...
+    < <(inotifywait -r  -e create,open --format '%f' --quiet "$datadir" --monitor)
     echo "C-Lightning started"
     echo "C-Lightning started, RPC available on port $LIGHTNINGD_RPC_PORT"
 
-    socat "TCP4-listen:$LIGHTNINGD_RPC_PORT,fork,reuseaddr" "UNIX-CONNECT:$LIGHTNINGD_DATA/lightning-rpc" &
+    socat "TCP4-listen:$LIGHTNINGD_RPC_PORT,fork,reuseaddr" "UNIX-CONNECT:${datadir}/lightning-rpc" &
     fg %-
 else
-    exec lightningd "$@"
+    exec lightningd --network=${LIGHTNINGD_NETWORK} "$@"
```
ZmnSCPxj, line 20: `shellcheck` wants quote marks around the variable.

ZmnSCPxj:
Also commit title should probably change as well.

joemphilips:
Updated according to the review. I also dropped `-r` option in`docker-entrypoint.sh` since it is unnecessary after I introduced `LIGHTNING_NETWORK` env variable.

joemphilips:
> Should not commit message say "take LIGHTNING_NETWORK env variable in Dockerfile"?

Right, I was careless. updated.

ZmnSCPxj:
Also probably deserves a Changelog entry in the commit message:

    Changelog-Added: Docker build now includes a `LIGHTNINGD_NETWORK` variable you can override to `regtest` or `testnet` or any valid argument to `--network`.

Though wording might be changed depending on how exactly Docker terminology is, I am not a Docker expert.

joemphilips:
Changelog added. I couldn't find any past changelog which uses newline. So I avoided using it.

ZmnSCPxj:
Rebased on master.  Reapplying ACK:

ACK 5092f334

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3813/commits/5092f3342cc8b8f54f4efc23f9e8e86787992ad1

tools/docker-entrypoint.sh
```diff
@@ -2,18 +2,20 @@
...
+networkdatadir="${LIGHTNINGD_DATA}/${LIGHTNINGD_NETWORK}"
+
 if [ "$EXPOSE_TCP" == "true" ]; then
     set -m
     lightningd "$@" &
 
     echo "C-Lightning starting"
-    while read -r i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
-    < <(inotifywait  -e create,open --format '%f' --quiet "$LIGHTNINGD_DATA" --monitor)
+    while read i; do if [ "$i" = "lightning-rpc" ]; then break; fi; done \
```
ZmnSCPxj, line 12: Hmm, this seems to be different.  `-r` for `read` means "ignore backslashes on input", which is recommended.  `shellcheck` is also complaining about this as well.  Maybe should not be removed?

joemphilips: Yes, it is my fault. fixed.

PR #3814, opened 2020-07-05
Do not use FEERATE_MAX in 0 starting loop.

joemphilips:
This will cause to loop 1 time more than necessary.
It was causing the following error in clang build
in macos.

```
lightningd/chaintopology.c:351:29: error: result of comparison of constant 8 with expression of type 'enum feerate' is always true
      [-Werror,-Wtautological-constant-out-of-range-compare]
        for (enum feerate f = 0; f < NUM_FEERATES; f++) {
```

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -180,7 +180,7 @@ static void estimatefees_callback(const char *buf, const jsmntok_t *toks,
 				     "estimatefees",
 				     "bad 'result' field");
 
-	for (enum feerate f = 0; f < FEERATE_MAX; f++) {
+	for (enum feerate f = 0; f < NUM_FEERATES; f++) {
 		feeratetok = json_get_member(buf, resulttok, feerate_name(f));
 		if (!feeratetok)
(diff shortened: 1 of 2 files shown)


rustyrussell:
Hi!  Your fix is correct (thanks!), but your explanation is wrong: we in fact were iterating one too few, since FEERATE_MAX is the last value, we want to iterate while less than NUM_FEERATES.

Travis got upset for unrelated flakiness reasons AFAICT, kicking now...

Ack https://github.com/ElementsProject/lightning/pull/3814/commits/52f9b26b8ae9eaaa446941ed46163dbfd56f052f

rustyrussell:
Rebasing on master which fixes unrelated Travis issue...

rustyrussell:
OK, I had to fix some of the other loops which were also incorrect.

cdecker:
Uhm, the diff is now empty. Seems the second commit just reverts the changes from the first.

rustyrussell:
Ah, OK, I was confused.

This patch was in fact incorrect all along.  The warning/error from clang is definitely incorrect (for C: in C++ the warning might be valid).

PR #3817, opened 2020-07-06
pytest: fix test breakage added by d0c85033d2e5fe83861af7badb577e7cef86a310

rustyrussell:
By setting nLocktime to the current block, the reorg test
"test_funding_reorg_remote_lags" actually drops the funding transaction
entirely when a reorg happens.

Except the 1 in 10 cases where nLocktime is randomly set to 1-10
blocks earlier.

This implies, strongly, that we hit "restart" too often on Travis.

Changelog-None

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1214,9 +1214,11 @@ def test_funding_reorg_remote_lags(node_factory, bitcoind):
 
     l2.daemon.rpcproxy.mock_rpc('getblockhash', no_more_blocks)
 
-    # Reorg changes short_channel_id 103x1x0 to 103x2x0, l1 sees it, restarts channeld
-    bitcoind.simple_reorg(102, 1)                   # heights 102 - 108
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
ACK 928f0ec

rustyrussell:
> Hmm okay, but I thought the reorg would change it to the same blockheight...? Strange.

Yes, but we reorg by invalidating block 103, which puts txs back in mempool.  But not that one with locktime 103.  Then we add replacement blocks.

ZmnSCPxj:
Ah, okay.  So:

* At blockheight 102, the 103-`nLockTime`d tx is accepted into mempool.
* At blockheight 103, we reorg down to blockheight 101.
* We are now at blockheight 101, so the 103-`nLockTime`d tx is dropped from mempool as "too far".
* The 103-`nLockTime`d transaction never confirms.

PR #3823, opened 2020-07-07
plugins/Makefile: Change plugins build procedure.

ZmnSCPxj:
Changelog-None

Extracted the first commit from #3763 , because #3763 is stuck until we can get 0-output PSBTs working, for some reason.  Anyway *this* commit should be uncontroversial since it does not involve speculative new frameworks for plugin execution, and also fixes some source files that were not included in our standard source file checks.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3823/commits/cd5d43eee274a663814e71e0dc609c99aeb2242d

PR #3821, opened 2020-07-07
UTXO selection cleanups and preps

rustyrussell:
These are the simple reorgs and a fix, to expose more JSON routines to plugins.  They should be non-controversial.

Issue #475, opened 2017-12-27
Feature Request: Speed up channel opening via CPFP+RBF


ZmnSCPxj:
An idea, is to ensure as much as possible that channel funding transactions have a "change" output that is controlled by the built-in onchain wallet.

Then, if the funding transaction takes a long time to confirm on mainnet (due to the well-known congestion problems there), it would be possible to accelerate the channel funding transaction by creating a child transaction from the "change" output.  This child transaction would have a larger-than-usual feerate (CPFP) to encourage mining of the channel funding transaction, and pay out to yet another address of the built-in onchain wallet.  In addition, the child transaction would be marked RBF so that further increases of the fee can be done.

Using CPFP to speed up the channel funding transaction has the advantage that it requires no change to the Lightning BOLT protocol.  Using RBF on the channel funding transaction directly requires that the both ends of the channel sign new versions of the commitment transaction.

While not a priority, I suspect this feature would be necessary before we can practically deploy to mainnet.

---

At the JSON-RPC level, I imagine:

    # open the channel but do not wait to confirm
    tx=$(lightning-cli fundchannelasync $node $satoshis)
    sleep 1h
    # Not yet confirmed even after one hour...?
    if [$(lightning-cli fundchannelconfirmed $tx) = False]; then
      # If we accelerate, how much more will we spend?
      moresatoshis=$(lightning-cli fundchannelaccelerateestimate $tx)
      # so expensive! can we afford it?
      if [$moresatoshis < $satoshithreshhold]; then
        # fine, pay more to speed it up.
        lightning-cli fundchannelaccelerate $tx
      fi
    fi

Further calls to fundchannelaccelerate will be possible and will RBF the child transaction instead.

The fundchannelaccelerate command will silently fail without an error if the specified channel is already confirmed at least once.

---

Implementation-wise, we would need a new table for pending channel funding transactions.  Entries in this table will be removed when the peer enters `CHANNELD_NORMAL`.

    CREATE TABLE funding_tx
      ( id INTEGER PRIMARY KEY REFERENCES channel(id) ON DELETE CASCADE
      , funding_tx_change_outnum INTEGER
      , cpfp_feerate INTEGER
      );

[...]

PR #3825, opened 2020-07-08
Reserve/unreserve/fundpsbt

rustyrussell:
~~(Note: this fails due to libwally failing to marshal 0-output psbts, but should otherwise be complete).~~ -- Fixed in master by @niftynei !

Based on #3821

This changes the (newly-introduced!) reserveinputs and unreserveinputs to be far more primitive, and adds fundpsbt which is a low-level psbt creator.  Unfortunately, the rework I had planned which migrates everything else on top of these will not be finished before the release.

Nonetheless, it's better to ship with these APIs now, so progress can continue: shortly after 0.9.0 I will complete the rewrite which makes txprepare/txdiscard/txcancel a plugin on top of these, and @ZmnSCPxj can do the same for multiwithdraw and multifundchannel!

Closes: #3789 
Fixes: #3415

### DIFF
--- a/doc/lightning-fundpsbt.7.md
+++ b/doc/lightning-fundpsbt.7.md
@@ -1,0 +1,69 @@
+lightning-fundpsbt -- Command to populate PSBT inputs from the wallet
+================================================================
+
+SYNOPSIS
+--------
+
+**fundpsbt** *satoshi* *feerate* \[*minconf*\] \[*reserve*\]
+
+DESCRIPTION
+-----------
+
+`fundpsbt` is a low-level RPC command which creates a PSBT using unreserved
+inputs in the wallet, optionally reserving them as well.
+
+*satoshi* is the minimum satoshi value of the output(s) needed (or the
+string "all" meaning use all unreserved inputs).  If a value, it can
+be a whole number, a whole number ending in *sat*, a whole number
+ending in *000msat*, or a number with 1 to 8 decimal places ending in
+*btc*.
+
+You calculate the value by starting with the amount you want to pay
+and adding the fee which will be needed to pay for the base of the
+transaction plus that output, and any other outputs and inputs you
+will add to the final transaction.
+
+*feerate* is a number, with an optional suffix: *perkw* means the
+number is interpreted as satoshi-per-kilosipa (weight), and *perkb*
+means it is interpreted bitcoind-style as
+satoshi-per-kilobyte. Omitting the suffix is equivalent to *perkb*.
+
+*minconf* specifies the minimum number of confirmations that used
+outputs should have. Default is 1.
+
+*reserve* is a boolean: if true (the default), then *reserveinputs* is
+called (successfully, with *exclusive* true) on the returned PSBT.
+
+RETURN VALUE
+------------
+
+On success, returns the *psbt* containing the inputs, and
+*excess_msat* containing the amount above *satoshi* which is
+available.  This could be zero, or dust.  If *satoshi* was "all",
+then *excess_msat* is the entire amount once fees are subtracted
+for the weights of the inputs.
+
+If *reserve* was true, then a *reservations* array is returned,
+exactly like *reserveinputs*.
+
+On error the returned object will contain `code` and `message` properties,
+with `code` being one of the following:
+
+- -32602: If the given parameters are wrong.
+- -1: Catchall nonspecific error.
+- 301: Insufficient UTXOs to meet *satoshi* value.
+
+AUTHOR
+------
+
+Rusty Russell <<rusty@rustcorp.com.au>> is mainly responsible.
+
+SEE ALSO
+--------
+
+lightning-reserveinputs(7), lightning-unreserveinputs(7).
+
+RESOURCES
+---------
+
+Main web site: <https://github.com/ElementsProject/lightning>
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -486,6 +486,64 @@ def test_reserveinputs(node_factory, bitcoind, chainparams):
     assert not any(o['reserved'] for o in l1.rpc.listfunds()['outputs'])
 
 
+def test_fundpsbt(node_factory, bitcoind, chainparams):
+    amount = 1000000
+    total_outs = 4
+    l1 = node_factory.get_node()
+
+    outputs = []
+    # Add a medley of funds to withdraw later, bech32 + p2sh-p2wpkh
+    for i in range(total_outs // 2):
+        txid = bitcoind.rpc.sendtoaddress(l1.rpc.newaddr()['bech32'],
+                                          amount / 10**8)
+        outputs.append((txid, bitcoind.rpc.gettransaction(txid)['details'][0]['vout']))
+        txid = bitcoind.rpc.sendtoaddress(l1.rpc.newaddr('p2sh-segwit')['p2sh-segwit'],
+                                          amount / 10**8)
+        outputs.append((txid, bitcoind.rpc.gettransaction(txid)['details'][0]['vout']))
+
+    bitcoind.generate_block(1)
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == total_outs)
+
+    feerate = '7500perkw'
+
+    # Should get one input, plus some excess
+    funding = l1.rpc.fundpsbt(amount // 2, feerate, reserve=False)
+    psbt = bitcoind.rpc.decodepsbt(funding['psbt'])
+    assert len(psbt['tx']['vin']) == 1
+    assert funding['excess_msat'] > Millisatoshi(0)
+    assert funding['excess_msat'] < Millisatoshi(amount // 2 * 1000)
+
+    # Cannot afford this one (too much)
+    with pytest.raises(RpcError, match=r"not afford"):
+        l1.rpc.fundpsbt(amount * total_outs, feerate)
+
+    # Nor this (depth insufficient)
+    with pytest.raises(RpcError, match=r"not afford"):
+        l1.rpc.fundpsbt(amount // 2, feerate, minconf=2)
+
+    # Should get two inputs.
+    psbt = bitcoind.rpc.decodepsbt(l1.rpc.fundpsbt(amount, feerate, reserve=False)['psbt'])
+    assert len(psbt['tx']['vin']) == 2
+
+    # Should not use reserved outputs.
+    psbt = bitcoind.rpc.createpsbt([{'txid': out[0], 'vout': out[1]} for out in outputs], [])
+    l1.rpc.reserveinputs(psbt)
+    with pytest.raises(RpcError, match=r"not afford"):
+        l1.rpc.fundpsbt(amount // 2, feerate)
+
+    # Will use first one if unreserved.
+    l1.rpc.unreserveinputs(bitcoind.rpc.createpsbt([{'txid': outputs[0][0], 'vout': outputs[0][1]}], []))
+    psbt = l1.rpc.fundpsbt(amount // 2, feerate)['psbt']
+
+    # Should have passed to reserveinputs.
+    with pytest.raises(RpcError, match=r"already reserved"):
+        l1.rpc.reserveinputs(psbt)
+
+    # And now we can't afford any more.
+    with pytest.raises(RpcError, match=r"not afford"):
+        l1.rpc.fundpsbt(amount // 2, feerate)
+
+
 @pytest.mark.xfail(strict=True)
 def test_sign_and_send_psbt(node_factory, bitcoind, chainparams):
     """
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -307,7 +307,6 @@ def test_txprepare(node_factory, bitcoind, chainparams):
     wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == 10)
 
     prep = l1.rpc.txprepare(outputs=[{addr: Millisatoshi(amount * 3 * 1000)}])
-    assert prep['psbt']
     decode = bitcoind.rpc.decoderawtransaction(prep['unsigned_tx'])
     assert decode['txid'] == prep['txid']
     # 4 inputs, 2 outputs (3 if we have a fee output).
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -438,130 +438,52 @@ def test_txprepare(node_factory, bitcoind, chainparams):
 
 
 def test_reserveinputs(node_factory, bitcoind, chainparams):
-    """
-    Reserve inputs is basically the same as txprepare, with the
-    slight exception that 'reserveinputs' doesn't keep the
-    unsent transaction around
-    """
     amount = 1000000
     total_outs = 12
     l1 = node_factory.get_node(feerates=(7500, 7500, 7500, 7500))
-    addr = chainparams['example_addr']
 
+    outputs = []
     # Add a medley of funds to withdraw later, bech32 + p2sh-p2wpkh
     for i in range(total_outs // 2):
-        bitcoind.rpc.sendtoaddress(l1.rpc.newaddr()['bech32'],
-                                   amount / 10**8)
-        bitcoind.rpc.sendtoaddress(l1.rpc.newaddr('p2sh-segwit')['p2sh-segwit'],
-                                   amount / 10**8)
+        txid = bitcoind.rpc.sendtoaddress(l1.rpc.newaddr()['bech32'],
+                                          amount / 10**8)
+        outputs.append((txid, bitcoind.rpc.gettransaction(txid)['details'][0]['vout']))
+        txid = bitcoind.rpc.sendtoaddress(l1.rpc.newaddr('p2sh-segwit')['p2sh-segwit'],
+                                          amount / 10**8)
+        outputs.append((txid, bitcoind.rpc.gettransaction(txid)['details'][0]['vout']))
 
     bitcoind.generate_block(1)
     wait_for(lambda: len(l1.rpc.listfunds()['outputs']) == total_outs)
 
-    utxo_count = 8
-    sent = Decimal('0.01') * (utxo_count - 1)
-    reserved = l1.rpc.reserveinputs(outputs=[{addr: Millisatoshi(amount * (utxo_count - 1) * 1000)}])
-    assert reserved['feerate_per_kw'] == 7500
-    psbt = bitcoind.rpc.decodepsbt(reserved['psbt'])
-    out_found = False
+    assert not any(o['reserved'] for o in l1.rpc.listfunds()['outputs'])
 
-    assert len(psbt['inputs']) == utxo_count
-    outputs = l1.rpc.listfunds()['outputs']
-    assert len([x for x in outputs if not x['reserved']]) == total_outs - utxo_count
-    assert len([x for x in outputs if x['reserved']]) == utxo_count
-    total_outs -= utxo_count
-    saved_input = psbt['tx']['vin'][0]
+    # Try reserving one at a time.
+    for out in outputs:
+        psbt = bitcoind.rpc.createpsbt([{'txid': out[0], 'vout': out[1]}], [])
+        l1.rpc.reserveinputs(psbt)
 
-    # We should have two outputs
-    for vout in psbt['tx']['vout']:
-        if chainparams['elements'] and vout['scriptPubKey']['type'] == 'fee':
-            continue
-        if vout['scriptPubKey']['addresses'][0] == addr:
-            assert vout['value'] == sent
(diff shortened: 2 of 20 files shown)


wallet/reservation.c
```diff
@@ -173,27 +182,30 @@ static struct command_result *json_fundpsbt(struct command *cmd,
...
-	u32 locktime, maxheight;
+	bool all, *reserve;
+	u32 locktime, maxheight, current_height;
 	struct bitcoin_tx *tx;
 
 	if (!param(cmd, buffer, params,
 		   p_req("satoshi", param_sat_or_all, &amount),
 		   p_req("feerate", param_feerate_val, &feerate_per_kw),
 		   p_opt_def("minconf", param_number, &minconf, 1),
+		   p_opt_def("reserve", param_bool, &reserve, true),
```
ZmnSCPxj, line 197: I approve

wallet/reservation.c
```diff
@@ -164,3 +165,115 @@ static const struct json_command unreserveinputs_command = {
...
+	/* Eventually fuzz it too. */
+	if (locktime > 100 && pseudorand(10) == 0)
+		locktime -= pseudorand(100);
+
+	/* FIXME: tx_spending_utxos does more than we need, but there
+	 * are other users right now. */
+	tx = tx_spending_utxos(cmd, chainparams, utxos,
+			       cmd->ld->wallet->bip32_base,
+			       false, 0, locktime,
+			       BITCOIN_TX_RBF_SEQUENCE);
```
ZmnSCPxj, line 273: Why RBF-flagged?  Am I able to *actually* RBF now??  How does RBF work?

ZmnSCPxj:
Does this effectively fix #3415 ?  The underlying issue in #3415 is that we want to have an atomic reservation of funds all the way to sign+broadcast without using the current discard+re-prepare trick in `fundchannel`.  My understanding is that it does, in which case this should be tagged "fixes #3415"?

wallet/reservation.c
rustyrussell: Yes, we can RBF once everyone uses this...

rustyrussell:
> Does this effectively fix #3415 ? The underlying issue in #3415 is that we want to have an atomic reservation of funds all the way to sign+broadcast without using the current discard+re-prepare trick in `fundchannel`. My understanding is that it does, in which case this should be tagged "fixes #3415"?

Yes, good point.

wallet/reservation.c
ZmnSCPxj: Okay, let me see if my understanding is correct.

To implement RBF, I would have to go all the way up to `signpsbt`, but I would have to *independently* extract the transaction and broadcast it, such as by the `bcli` `sendrawtransaction` method.  Then if I decide to RBF, I would mutate the PSBT directly, then call `signpsbt` again with the RBFed PSBT, then `sendrawtransaction` again.

(we cannot call `sendpsbt` more than once, since it marks consumed coins as spent, with an assertion that the consumed coin must currently be reserved)

Then, if any of the transactions get confirmed onchain, this automatically marks the consumed inputs as spent.  Does that happen?

Looking at `lightningd/chaintopology.c` it  seems we have the rough framework to do so, but while it calls into `wallet_outpoint_spend`, that function does not update the `status` of UTXOs.  It also calls into `wallet_extract_owned_outputs`, which will get any unknown outputs, but not what a transaction spends.

`wallet_update_update_output_status` is otherwise only called from within `wallet/wallet.c`, or from `dev-rescan-outputs`.

I don't see any changes in `wallet/wallet.c` that would allow the above flow to be used.

ZmnSCPxj: And then there is the issue where one version of the transaction is seen by one miner, the other version is seen by the other miner, then we see the block created by the first, but ultimately the second miner gets a second block, so now we have to rewind our chaintopology and destroy UTXOs in that chaintopology and recreate it as we go to the winning branch.  Do we have code that handles that?  Because if not, we are not RBF safe.

wallet/reservation.c
```diff
@@ -164,3 +165,115 @@ static const struct json_command unreserveinputs_command = {
...
+						   &needed));
+	}
+
+	/* Setting the locktime to the next block to be mined has multiple
+	 * benefits:
+	 * - anti fee-snipping (even if not yet likely)
+	 * - less distinguishable transactions (with this we create
+	 *   general-purpose transactions which looks like bitcoind:
+	 *   native segwit, nlocktime set to tip, and sequence set to
+	 *   0xFFFFFFFE by default. Other wallets are likely to implement
```
niftynei, line 258: nota bene: we set the sequence to 0xFFFFFFFD below

wallet/walletrpc.c
```diff
@@ -1248,8 +1248,12 @@ static struct command_result *match_psbt_inputs_to_utxos(struct command *cmd,
 		wally_tx_input_get_txid(&psbt->tx->inputs[i], &txid);
 		utxo = wallet_utxo_get(*utxos, cmd->ld->wallet,
 				       &txid, psbt->tx->inputs[i].index);
-		if (!utxo)
+		if (!utxo) {
+			log_debug(cmd->ld->log, "utxo %zu DNE", i);
```
niftynei, line 1252: these seem extraneous?

rustyrussell:
@ZmnSCPxj for some reason I can't quote you on mobile...

Yes, it also requires the change to reserve, rather than mark spent, utxos when we sendpsbt.  (This is the Right Thing anyway, in case the tx never confirms).

This is also good, since reserving them twice means we'll extend the reservation, which makes sense.

ZmnSCPxj:
@rustyrussell Okay but *does* spending a utxo we own without passing the spend through `wallet/walletrpc.c` interfaces work?

One way to implement #475 would be to keep the PSBT around until *some* transaction confirms.  That means keeping the reservation around until the target transaction confirms.

doc/lightning-fundpsbt.7.md
```diff
@@ -0,0 +1,67 @@
...
+-----------
+
+`fundpsbt` is a low-level RPC command which creates a PSBT using unreserved
+inputs in the wallet, optionally reserving them as well.
+
+*satoshi* is the minimum satoshi value of the output(s) needed (or the
+string "all" meaning use all unreserved inputs).  If a value, it can
+be a whole number, a whole number ending in *sat*, a whole number
+ending in *000msat*, or a number with 1 to 8 decimal places ending in
+*btc*.
```
ZmnSCPxj, line 19: We should probably mention that:

* `fundpsbt` factors in the fees needed for each input it proposes to add.
* the given `satoshi` value should include the fee for the outputs, plus the fee for the common parts (`nLockTime`, `nVersion` etc) of the transaction.

The latter point implies that plugins need access to the fee estimates.  While plugins can call the `bcli` `estimatefees` directly, I think it is best if `lightningd/chaintopology` also exposes the actual feerates it would use, especially since `estimatefees` is allowed to return `null` and `chaintopology`, to my understanding, papers over that case. (I think? @darosior )

Alternately, since e.g. Liquid apparently has explicit fee outputs, maybe we can add another parameter, `outputs`, which is an array of strings or `null`s:

* An entry can be an address, in which case `fundpsbt` allocates fees enough for that address.
* An entry can be any of the following strings, indicating `fundpsbt` allocates fees enough for that *kind* of address:
  * `p2pkh`
  * `p2pk` - or not who uses this nowadays
  * `p2sh`
  * `p2wpkh`
  * `p2wsh`
  * `op-return-#` where # is a number from 0 to 64 (or whatever the `OP_RETURN` max is).
* An entry can be `null`, in which case `fundpsbt` will allocate fees for the largest known valid address type `scriptPubKey`..

Then `fundpsbt` can consider whether there is some excess and see if it is large enough to (1) pay for its existence as a change output and (2) remains large enough to keep itself above the dust limit.  If not, it adds the excess to fees (which it adds as an explicit output for Liquid txes).

With this `outputs` parameter we can also have `fundpsbt` consider the common parts of the transaction (`nLockTime` and friends).  This removes any remaining reasons that a robust plugin would want to loop around `fundpsbt`.

darosior: > The latter point implies that plugins need access to the fee estimates. While plugins can call the bcli estimatefees directly, I think it is best if lightningd/chaintopology also exposes the actual feerates it would use, especially since estimatefees is allowed to return null and chaintopology, to my understanding, papers over that case. (I think? @darosior )

Right, better to use the `feerate` command which returns a human error in case of estimation failure and smoothed estimation otherwise.

`estimatefees` could fail we allow it to, while `feerates` would still return a result based on our cached history.

ZmnSCPxj: Ah, right, `feerates`.  Will go document it for now.

rustyrussell: > We should probably mention that:
> 
>     * `fundpsbt` factors in the fees needed for each input it proposes to add.
> 
>     * the given `satoshi` value should include the fee for the outputs, plus the fee for the common parts (`nLockTime`, `nVersion` etc) of the transaction.

The next paragraph does exactly this:

You calculate the value by starting with the amount you want to pay
and adding the fee which will be needed to pay for the base of the
transaction plus that output, and any other outputs and inputs you
will add to the final transaction.

> With this `outputs` parameter we can also have `fundpsbt` consider the common parts of the transaction (`nLockTime` and friends). This removes any remaining reasons that a robust plugin would want to loop around `fundpsbt`.

Indeed.  My draft `txprepare-as-plugin` patch does exactly this, and it's pretty simple (now we have the helpers in place!).

rustyrussell:
OK, addressed minor feedback from @niftynei  and rebased onto master now libwally is fixed...

doc/lightning-fundpsbt.7.md
ZmnSCPxj: > Indeed. My draft `txprepare-as-plugin` patch does exactly this, and it's pretty simple (now we have the helpers in place!).

Is the code in `txprepare` or in `fundpsbt`?  I cannot find `txprepare-as-plugin` branch in your repo or in the ElementsProject repo.  My concern is that I believe you mentioned deprecation of `txprepare`, and in any case `fundchannel`/`multifundchannel` will want to move away from `txprepare` anyway.  So a command which wraps `fundpsbt` and handles as well the common and output weight fees would be quite convenient; if `txprepare` will be deprecated later, it would be useful for both `withdraw` and `fundchannel` (and their `multi` versions) to use that rather than the existing base `fundpsbt`.

doc/lightning-fundpsbt.7.md
```diff
@@ -0,0 +1,67 @@
+lightning-fundpsbt -- Command to populate PSBT inputs from the wallet
+================================================================
+
+SYNOPSIS
+--------
+
+**fundpsbt** *satoshi* *feerate* \[*minconf* *reserve*\]
```
cdecker, line 7:
```suggestion
**fundpsbt** *satoshi* *feerate* \[*minconf*\] \[*reserve*\]
```

rustyrussell:
> I'm not exactly clear how `all` and the bumping of reservations interact. If I have a wallet with unreserved UTXOs `a`, `b`, and `c`, then I reserve `a`, will a call to `fundpsbt` bump the previous `a` reservation, or will it skip `a`?
> 
> I'd expect the latter (`a` is already reserved for another purpose, and I just want to gather all other UTXOs), but I could not convince myself that that's the case.

The UTXO selection only considers unreserved UTXOs, so it will skip `a`.

wallet/wallet.c
```diff
@@ -418,6 +421,84 @@ void wallet_confirm_utxos(struct wallet *w, const struct utxo **utxos)
 	}
 }
 
+static void db_set_utxo(struct db *db, const struct utxo *utxo)
```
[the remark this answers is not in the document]
rustyrussell: Yes, but that's silly.  Particularly for an internal function which literally hits the db.  Note that many of these functions are going to be removed once the users are gone (i.e. once fundchannel, withdraw and txprepare are reworked on top of fundpsbt/reservepbst/unreservepsbt).

wallet/wallet.c
```diff
@@ -418,6 +421,84 @@ void wallet_confirm_utxos(struct wallet *w, const struct utxo **utxos)
...
+
+	return true;
+}
+
+void wallet_unreserve_utxo(struct wallet *w, struct utxo *utxo, u32 current_height)
+{
+	if (utxo->status == output_state_reserved)
+		assert(utxo->reserved_til);
+	else
+		assert(!utxo->reserved_til);
```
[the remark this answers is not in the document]
rustyrussell: Actually, I think this can potentially crash if you use the old code to reserve, and then call this manually.  Will check...

wallet/reservation.c
```diff
@@ -0,0 +1,154 @@
...
+	for (size_t i = 0; i < psbt->tx->num_inputs; i++) {
+		struct bitcoin_txid txid;
+		struct utxo *utxo;
+		enum output_status oldstatus;
+		u32 old_res;
+
+		wally_tx_input_get_txid(&psbt->tx->inputs[i], &txid);
+		utxo = wallet_utxo_get(cmd, cmd->ld->wallet,
+				       &txid, psbt->tx->inputs[i].index);
+		if (!utxo || utxo->status != output_state_reserved)
```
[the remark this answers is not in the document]
rustyrussell: Hmm, well if you didn't have a reservation in the first place, unreserve ignores it.  So outputting something here would be weird.

wallet/wallet.c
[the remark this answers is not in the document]
ZmnSCPxj: It helps when naming functions and structs to prepend them with a convenient namespace.  I get annoyed that we have multiple contradictory `struct channel`s, for example, and I have to do extra digging to figure out which one is being used in a particular piece of code.

wallet/wallet.c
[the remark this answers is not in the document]
rustyrussell: Indeed, fixed.

rustyrussell:
OK, fixups pushed for those issues.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3825/commits/01672a9c0436b4a36ffb8530f5743f11cfef80c2

cdecker:
Rebased on `master`

cdecker:
Sqashed the fixups, had to manually merge a couple since they were conflicting, but the [end-to-end diff is identical now](http://46.101.246.115:8888/ElementsProject/lightning/pull/3825).

ACK https://github.com/ElementsProject/lightning/pull/3825/commits/095f4e6b543a4e61ef4c0e73fd21f66b5e861ef6

cdecker:
Sadly the rebase seems to have broken something somewhere. I'm now getting these errors:

```
       # Join and add an output
        joint_psbt = bitcoind.rpc.joinpsbts([l1_funding['psbt'], l2_funding['psbt'],
>                                            output_pbst])
tests/test_wallet.py:644: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
contrib/pyln-testing/pyln/testing/utils.py:311: in f
    return proxy._call(name, *args)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <bitcoin.rpc.RawProxy object at 0x7fd9948c7b90>
service_name = 'joinpsbts'
args = (['cHNidP8BAAsCAAAAAAAAaAAAAAA=', 'cHNidP8BAK4CAAAABI3tfktaNA4PBnGH8ZPYfNsRFTjclLsA41ICPDFzo2HOAQAAAAD9////ri3g6ZylJh2...tWBzibHhwEEFgAUoyPCgqmhl/wRjsCc2vIcorsbiHIA', 'cHNidP8BACkCAAAAAAHAxi0AAAAAABYAFMkJbUP0COpSYCAmLM2tfIUWuSqBAAAAAAAA'],)
postdata = '{"version": "1.1", "method": "joinpsbts", "params": [["cHNidP8BAAsCAAAAAAAAaAAAAAA=", "cHNidP8BAK4CAAAABI3tfktaNA4PBn...EFgAUoyPCgqmhl/wRjsCc2vIcorsbiHIA", "cHNidP8BACkCAAAAAAHAxi0AAAAAABYAFMkJbUP0COpSYCAmLM2tfIUWuSqBAAAAAAAA"]], "id": 1}'
headers = {'Authorization': b'Basic cnBjdXNlcjpycGNwYXNz', 'Content-type': 'application/json', 'Host': 'localhost', 'User-Agent': 'AuthServiceProxy/0.1'}
response = {'error': {'code': -22, 'message': 'TX decode failed Size of value was not the stated size: iostream error'}, 'id': 1, 'result': None}
    def _call(self, service_name, *args):
        [...snip...]
        response = self._get_response()
        if response['error'] is not None:
>           raise JSONRPCError(response['error'])
E           bitcoin.rpc.JSONRPCError: {'code': -22, 'message': 'TX decode failed Size of value was not the stated size: iostream error'}
```

Any ideas what this might be @rustyrussell ?

niftynei:
That’s s PSBT decode error, are you running the test on regtest?

cdecker:
This is directly from travis [here](https://travis-ci.org/github/ElementsProject/lightning/jobs/707881030) with the configuration as follows:

```
$ export VALGRIND=1
$ export ARCH=64
$ export DEVELOPER=1
$ export COMPILER=gcc
$ export TEST_GROUP=12
$ export TEST_GROUP_COUNT=12
$ export SOURCE_CHECK_ONLY=false
```

tests/test_wallet.py
```diff
@@ -623,15 +627,22 @@ def test_sign_and_send_psbt(node_factory, bitcoind, chainparams):
...
 
     # Try to get L1 to sign it
     with pytest.raises(RpcError, match=r"No wallet inputs to sign"):
-        l1.rpc.signpsbt(l2_reserved['psbt'])
+        l1.rpc.signpsbt(l2_funding['psbt'])
 
     # Add some of our own PSBT inputs to it
-    l1_reserved = l1.rpc.reserveinputs(outputs=[{addr: Millisatoshi(3 * amount * 1000)}])
-    joint_psbt = bitcoind.rpc.joinpsbts([l1_reserved['psbt'], l2_reserved['psbt']])
+    l1_funding = l1.rpc.fundpsbt(satoshi=Millisatoshi(0),
```
niftynei, line 638: Is this intended to be a zero msat fund call? the PSBT that it's returning is invalid?

`cHNidP8BAAsCAAAAAAAAZgAAAAA=`

rustyrussell: Oops, that was me.  Weird, here's the fix I'm applying:
```diff
diff --git a/tests/test_wallet.py b/tests/test_wallet.py
index 5afbb518b..8fabd83b0 100644
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -636,7 +636,7 @@ def test_sign_and_send_psbt(node_factory, bitcoind, chainparams):
         l1.rpc.signpsbt(l2_funding['psbt'])
 
     # Add some of our own PSBT inputs to it
-    l1_funding = l1.rpc.fundpsbt(satoshi=Millisatoshi(0),
+    l1_funding = l1.rpc.fundpsbt(satoshi=Millisatoshi(3 * amount * 1000),
                                  feerate=7500,
                                  reserve=True)
 
@@ -696,8 +696,11 @@ def test_sign_and_send_psbt(node_factory, bitcoind, chainparams):
         {'type': 'chain_mvt', 'credit': 0, 'debit': 3000000000, 'tag': 'withdrawal'},
         {'type': 'chain_mvt', 'credit': 0, 'debit': 1000000000, 'tag': 'chain_fees'},
         {'type': 'chain_mvt', 'credit': 0, 'debit': 0, 'tag': 'spend_track'},
-        {'type': 'chain_mvt', 'credit': 0, 'debit': 1000000000, 'tag': 'withdrawal'},
-        {'type': 'chain_mvt', 'credit': 0, 'debit': 0, 'tag': 'chain_fees'},
+        {'type': 'chain_mvt', 'credit': 0, 'debit': 0, 'tag': 'spend_track'},
+        {'type': 'chain_mvt', 'credit': 0, 'debit': 0, 'tag': 'spend_track'},
+        {'type': 'chain_mvt', 'credit': 0, 'debit': 0, 'tag': 'spend_track'},
+        {'type': 'chain_mvt', 'credit': 0, 'debit': 3000000000, 'tag': 'withdrawal'},
+        {'type': 'chain_mvt', 'credit': 0, 'debit': 1000000000, 'tag': 'chain_fees'},
     ]
 
     check_coin_moves(l1, 'wallet', wallet_coin_mvts, chainparams)
```

rustyrussell:
Ack d1627391ceb99d9ec678cb33718420954cbdb3ee

PR #3792, opened 2020-06-25
Implement `keysend` sending support for spontaneous payments

cdecker:
This has been a long time in the making, but we finally are able to build new features on top of the paymod series.

This modifier is rather simple, but that's all it takes: add the payment preimage, that we generate randomly, into the onion payload for the final hop, so a recipient that knows to look for it can claim the incoming payment.

## Todo
 - [x] Write the man-page for `lightning-cli keysend`

PR #3826, opened 2020-07-08
paymod: Activate the new payment flow for good (part 04)

cdecker:
This is the last step in activating the new payment flow.

 - It removes all the `compat` changes to tests done in the previous parts and now uses the updated RPC everywhere
 - Added a `legacy` option to `pay` (and `legacypay` in order not to cause param failures) to fall back on the old implementation should there be an issue.
 - Fixed a tiny regression with flapping payment status if we happen to wait for the blockheights to match up.

We informally decided to activate the new payment flow for everybody and provide an escape hatch, for users that require the legacy payment model.

## Open questions
 - [x] Do we need a `legacypay` exposed via JSON-RPC at all, or should we only expose via `legacy` flag on the `pay` command?

Changelog-None

### DIFF
--- a/.travis/build.sh
+++ b/.travis/build.sh
@@ -12,9 +12,8 @@ export PATH=$CWD/dependencies/bin:"$HOME"/.local/bin:"$PATH"
 export PYTEST_PAR=2
 export PYTEST_SENTRY_ALWAYS_REPORT=1
 
-# If we're not in developer mode, tests spend a lot of time waiting for gossip!
-# But if we're under valgrind, we can run out of memory!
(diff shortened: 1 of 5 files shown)


rustyrussell:
>     * [ ]  Do we need a `legacypay` exposed via JSON-RPC at all, or should we only expose via `legacy` flag on the `pay` command?

Removing parameters is hard.  Let's just call it legacypay and remove it later.

cdecker:
I removed the `legacy` parameter from `pay` and kept the old `pay` code in `legacypay` unchanged. This way we'll just have to phase out `legacypay` once `COMPAT_090` gets removed.

cdecker:
Rebased on top of #3792 now in `master` and added some more minor fixes.

PR #3828, opened 2020-07-12
doc/lightning-feerates.7.md: Document `feerates` command.

ZmnSCPxj:
Makes sense to openly expose this command once we have `fundpsbt`, as `fundpsbt` requires the caller to compute the output feerates by itself.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3828/commits/6de30095ab774edc03f36ef2978626dde6efd4f3

PR #3837, opened 2020-07-16
CHANGELOG.md: Update security changes for 0.7.1.

ZmnSCPxj:
The CVE was fully disclosed, so we can safely add it to the Security
field for the 0.7.1 changelog.

Also removed the "No security changes were necessary" text.
If we do this for releases, then either we lie about a CVE-level problem,
or we leak that a release fixes a CVE-level problem.

Changelog-None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Security
 
-No security related changes were necessary.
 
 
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
Alternately, we can write a shasum for "No security changes were necessary" text followed by a hexadecimal salt, which we change to an actual message if the release is old enough.  More troublesome to track and store though.  *shrug* not a serious proposal.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3837/commits/95932ea03942ab5d90f078a9825890aa3b112d69

PR #3838, opened 2020-07-16
UTXO Selection Cleanup

rustyrussell:
This reworks everything on top of `reservepsbt`, `unreservepsbt` and `fundpsbt`.  It's a little awkward: I think we should deprecate `txprepare` and friends once we've exposed enough helpers to make using `fundpsbt` more bearable.

Might split this into a few PRs for proper review, but meanwhile this shows the end goal.  Some tests may need more work, but this shows where I'm up to.


ZmnSCPxj:
I strongly suggest moving much of the processing involved around `txprepare` `utxos` and interfacing with `fundpsbt` to a new `preparepsbt` method.

The issue is that `multifundchannel`/`fundchannel` would really want to get access at the code for handling `utxos` and interfacing with `fundpsbt`.  But as #3415 points out, the `txprepare` and friends interface is not well-suited, as we do not know the actual P2WSH addresses to fund the channels yet.

So here is the spec for `preparepsbt` I am working on:

----

lightning-preparepsbt -- Command to reserve inputs to initially populate a PSBT
===============================================================================

SYNOPSIS
--------

**preparepsbt** *outputs* \[*feerate*\] \[*minconf*\] \[*utxos*\] \[*pay\_common\_fees*\]

DESCRIPTION
-----------

The **preparepsbt** RPC command is a low-level RPC command which
reserves inputs in the wallet.
This command also allocates funds to pay for fees;
see the section FUND ALLOCATION.

The *outputs* is an array of *destination* and *amount* pairs
(\{*destination*: *amount*\}).
The format is like:
\[\{*destination1*: *amount1*\}, \{*destination2*: *amount2*\}\].

The *destination* can be:

* A valid P2PKH address: e.g. 1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2
  on Bitcoin mainnet.
  An output for that address and amount will be added to the returned PSBT,
  and fees for that output will be allocated.
* A valid P2SH address: e.g. 3J98t1WpEZ73CNmQviecrnyiWrnqRhWNLy
  on Bitcoin mainnet.
  An output for that address and amount will be added to the returned PSBT,
  and fees for that output will be allocated.
* A valid P2WPKH address: e.g. bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq
  on Bitcoin mainnet.
  An output for that address and amount will be added to the returned PSBT,
  and fees for that output will be allocated.
* A valid P2WSH address: e.g. bc1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3qccfmv3
  on Bitcoin mainnet.
  An output for that address and amount will be added to the returned PSBT,
  and fees for that output will be allocated.
* The string *\_allocate*.
  Note the underscore before the word.
  Funds will be allocated such that you can spend this in other outputs,
  but fees will *not* be added for any such an output(s).
  On Bitcoin, if you do not add (and pay for) an output that spends this
  amount, they will be put as fees!
  On Elements, you will need to later adjust the fee output to handle this
  extra amount.
* The string *\_p2pkh*.
  Fees for a P2PKH output, plus the specified amount,
  will be allocated,
  but the output will not be added to the PSBT (since address is unknown).
* The string *\_p2sh*.
  Fees for a P2SH output, plus the specified amount,
  will be allocated,
  but the output will not be added to the PSBT (since address is unknown).
* The string *\_p2wpkh*.
  Fees for a P2WPKH output, plus the specified amount,
  will be allocated,
  but the output will not be added to the PSBT (since address is unknown).
* The string *\_p2wsh*.
  Fees for a P2WSH output, plus the specified amount,
  will be allocated,
  but the output will not be added to the PSBT (since address is unknown).
* The string *\_unknown*.
  Fees for the largest known address output (P2WSH as of this writing)
  will be allocated, but the output will not be added to the PSBT
  (since address is unknown).

The *amount* of an output is the amount to be allocated from the wallet
for that output.
This can be the string *all* to specify that all funds
(that are not allocated to fees and other outputs)
should be allocated to this output.
Alternately, it can be
 a whole number,
 a whole number ending in *sat*,
 a whole number ending in *000msat*, or
 a whole number with 1 to 8 decimal places ending in *btc*.

At most one of the *outputs* can have an *amount* with the string *all*.
The *outputs* array must not be empty.

*feerate* is an optional feerate to use. It can be one of the strings
*urgent* (aim for next block), *normal* (next 4 blocks or so) or *slow*
(next 100 blocks or so) to use lightningd’s internal estimates: *normal*
is the default.

Otherwise, *feerate* is a number, with an optional suffix: *perkw* means
the number is interpreted as satoshi-per-kilosipa (weight), and *perkb*
means it is interpreted bitcoind-style as satoshi-per-kilobyte. Omitting
the suffix is equivalent to *perkb*.

*minconf* specifies the minimum number of confirmations that used
outputs should have. Default is 1.

*utxos* specifies the utxos to be used to fund the transaction, as an array
of "txid:vout". These must be drawn from the node's available UTXO set.

*pay\_common\_fees* specifies whether to allocate fees for the common areas
of the transaction, and is a Boolean *true* or *false*.
On Elements, the fee output is considerd a common area.
If unspecified, defaults to *true*.

RETURN VALUE
------------

On success, an object with a *psbt* field is returned.
This is an unsigned PSBT containing
 the inputs,
 a change output if appropriate,
 any outputs with a known address, and,
 on Elements, the fee output.
You may further manipulate this *psbt* before passing it to
lightning-signpsbt(7) and then lightning-sendpsbt(7),
or you may cancel your spend by passing it to
lightning-unreserveinputs(7).

If any of the outputs specified *all*,
then an *all* field is returned as well,
with the amount in satoshis terminated by the string *000msat*.
This is the amount that the *all* output should have.

Finally, a field *feerate* is also returned which contains the feerate,
with the suffix *perkw*, in units of 1 satoshi per 1000 weight units.
This is the feerate that was used by **preparepsbt**.
Note that any *perkb* feerate will be rounded up to the nearest
multiple of 4 during conversion to a *perkw* feerate;
C-lightning always uses *perkw* computations internally.

On error the returned object will contain *code* and *message* fields,
with *code* being one of the following:

- -32602: If the given parameters are wrong.
- -1: Catchall nonspecific error.
- 301: Insufficient UTXOs to meet *satoshi* value.
- 320: Feerate estimation for your given *feerate* not available.
  This will only be returned if you specified *urgent*, *normal*,
  or *slow*.

FUND ALLOCATION
---------------

**preparepsbt**, while processing its *outputs* parameter,
maintains two "bins",
 a fee bin and
 an amounts bin.
Both are initially 0 satoshi.

If *pay\_common\_fees* is *true*, then the fee bin is initialized to
add fees for common parts of the transaction,
and, on Elements, the fee outpuut.

For each output, if that output is not *all*, the amount is added to the
amounts bin.
If the output is *all* then the amounts bin is set to *all*,
which is treated as an infinity.
Then, if the *destination* is not *\_allocate*,
the fees for the destination type is added to the fees bin.

If you specified a *utxos* argument, all of them are added to the PSBT
(and their fees are added to the fee bin).

If *utxos* is not specified,
inputs are selected and added to the PSBT.
The fee to spend the input is added to the fee bin.
Once enough inputs have been allocated
that exceed the sum of the fee bin and the amounts bin,
input selection ends.

If the inputs exceed the sum of the fee bin and the amounts bin,
this excess will be evaluated whether it can be put in change.
If the excess, minus the fee for an extra output, exceeds the dust limit,
we add the fee for the change output to the fee bin
and the excess minus change output fee
to the amounts bin
and add a change output to the PSBT.
Otherwise, we add the entire excess amount to the fee bin.

Finally, on Elements, a fee output is added to the PSBT, equal to the fee bin.

AUTHOR
------

ZmnSCPxj < <ZmnSCPxj@protonmail.com> > is mainly responsible.

SEE ALSO
--------

lightning-fundpsbt(7), lightning-unreserveinputs(7),
lightning-signpsbt(7), lightning=sendpsbt(7),
lightning-txprepare(7).

RESOURCES
---------

Main web site: <https://github.com/ElementsProject/lightning>

----

With `preparepsbt`, we have an almost-plug-in replacement of `txprepare` and friends:

* `txprepare` -> `preparepsbt`
* `txdiscard` -> `unreserveinputs`
* `txsend` -> `signpsbt` + `sendpsbt`

While `withdraw`/`multiwithdraw` can be built on top of `txprepare`, `fundchannel`/`multifundchannel` would really appreciate working with PSBTs.

So I propose moving the `utxos` handling and `fundpsbt` interfacing to a new `preparepsbt` specced above, then implementing `txprepare` in terms of `preparepsbt` instead of in terms of `fundpsbt` and `reserveinputs`.

We can even add a flag to `preparepsbt`, `reject_non_address`, default `false`, and if `true` will make `preparepsbt` error if a destination is an `_allocate`, `_p2pkh` etc keyword instead of an address, so that `txprepare` can just forward its args to `preparepsbt` and all `txprepare` and friends do is maintain a mapping of `txid` to `psbt`.

rustyrussell:
> I strongly suggest moving much of the processing involved around `txprepare` `utxos` and interfacing with `fundpsbt` to a new `preparepsbt` method.
> 
> The issue is that `multifundchannel`/`fundchannel` would really want to get access at the code for handling `utxos` and interfacing with `fundpsbt`. But as #3415 points out, the `txprepare` and friends interface is not well-suited, as we do not know the actual P2WSH addresses to fund the channels yet.

That's OK, because multifund knows the number of outputs, and that's sufficient for fundpsbt.

We *do* need to give access to our growing library of psbt-manipulation APIs, because advanced uses *will* need to manipulate them (e.g. merging HTLC txs, batching transactions, and dual funding).

This work is useful as the world convert to PSBTs anyway.

There are more changes to this series: in particular, I can eliminate feerate lookup and the listfunds steps...

ZmnSCPxj:
> That's OK, because multifund knows the number of outputs, and that's sufficient for fundpsbt.

But `fundpsbt` does not handle `utxos`, which is a feature that `fundchannel` currently supports; so `fundchannel` needs to get at the the handling of `utxos`, which requires doing `listfunds` and since `listfunds` does not have key indexes (which are needed to properly build inputs spending p2sh-p2wpkh) also do `dev-listaddrs` to match up addresses to key indexes.

rustyrussell:
Closing in favour of updated version.

PR #3840, opened 2020-07-17
fundpsbt: make parameters more usable.

rustyrussell:
fundpsbt forces the caller to manually add their weight * feerate
to the satoshis they ask for.  That means no named feerates.

Instead, create a startweight parameter and do the calc for them
internally, and return the feerate we used (and, while we're at it,
the estimated final weight).

This API change is best done now, as it would otherwise have to
be appended as a parameter.

Changelog-None

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -506,33 +506,45 @@ def test_fundpsbt(node_factory, bitcoind, chainparams):
     feerate = '7500perkw'
 
     # Should get one input, plus some excess
-    funding = l1.rpc.fundpsbt(amount // 2, feerate, reserve=False)
+    funding = l1.rpc.fundpsbt(amount // 2, feerate, 0, reserve=False)
     psbt = bitcoind.rpc.decodepsbt(funding['psbt'])
     assert len(psbt['tx']['vin']) == 1
     assert funding['excess_msat'] > Millisatoshi(0)
     assert funding['excess_msat'] < Millisatoshi(amount // 2 * 1000)
+    assert funding['feerate_per_kw'] == 7500
+    assert 'estimated_final_weight' in funding
+    assert 'reservations' not in funding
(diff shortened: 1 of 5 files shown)


wallet/reservation.c
```diff
@@ -203,6 +204,11 @@ static struct command_result *json_fundpsbt(struct command *cmd,
 
 	current_height = get_block_height(cmd->ld->topology);
 
+	/* Can overflow if amount is "any" */
```
ZmnSCPxj, line 207: Do you mean "all"?

rustyrussell: Oops, yes!

rustyrussell:
Agreed, but we need to leave that for @cdecker as release captain.

tests/test_wallet.py
```diff
@@ -506,41 +506,52 @@ def test_fundpsbt(node_factory, bitcoind, chainparams):
...
     # Should get one input, plus some excess
-    funding = l1.rpc.fundpsbt(amount // 2, feerate, reserve=False)
+    funding = l1.rpc.fundpsbt(amount // 2, feerate, 0, reserve=False)
     psbt = bitcoind.rpc.decodepsbt(funding['psbt'])
     assert len(psbt['tx']['vin']) == 1
     assert funding['excess_msat'] > Millisatoshi(0)
     assert funding['excess_msat'] < Millisatoshi(amount // 2 * 1000)
+    assert funding['feerate_per_kw'] == 7500
+    assert 'estimated_final_weight' in funding
+    assert not 'reservations' in funding
```
ZmnSCPxj, line 516: `check-source` is triggering on this:

>     tests/test_wallet.py:516:12: E713 test for membership should be 'not in'

rustyrussell:
.... and fix the other test which uses it.

PR #3843, opened 2020-07-17
Misc internal cleanups (nothing urgent)

rustyrussell:
Changelog-None

### DIFF
--- a/bitcoin/psbt.c
+++ b/bitcoin/psbt.c
@@ -361,11 +361,20 @@ struct wally_tx *psbt_finalize(struct wally_psbt *psbt, bool finalize_in_place)
 	return NULL;
 }
 
-bool psbt_from_b64(const char *b64str, struct wally_psbt **psbt)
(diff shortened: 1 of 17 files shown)


rustyrussell:
Trivial rebase...

bitcoin/psbt.c
```diff
@@ -10,6 +10,8 @@
 #include <common/amount.h>
 #include <common/type_to_string.h>
 #include <common/utils.h>
+/* FIXME: this is ugly! */
+#include <external/libwally-core/src/ccan/ccan/base64/base64.h>
```
ZmnSCPxj, line 14: Any reason to not just, you know --- include `ccan/base64` in our `ccan/` dir?

rustyrussell: (Had to think back why I did it this way!)

I think because this is an ugly hack anyway.  We *really* don't want to be doing this.  Let me see if I can clean up further...

rustyrussell: Fixed,

PR #3845, opened 2020-07-17
utxopsbt command

rustyrussell:
OK, after pondering @ZmnSCPxj 's feedback, I came up with a new helper to make life easier.

~It's on top of #3844 so see last commit...~ Now ready to go!

### DIFF
--- a/doc/lightning-utxopsbt.7.md
+++ b/doc/lightning-utxopsbt.7.md
@@ -1,0 +1,60 @@
+lightning-utxopsbt -- Command to populate PSBT inputs from given UTXOs
+================================================================
+
+SYNOPSIS
+--------
+
+**utxopsbt** *satoshi* *feerate* *startweight* *utxos* \[*reserve*\]
+
+DESCRIPTION
+-----------
+
+*utxopsbt* is a low-level RPC command which creates a PSBT using unreserved
+inputs in the wallet, optionally reserving them as well.
+
+It deliberately mirrors the parameters and output of
+lightning-fundpsbt(7) except instead of an optional *minconf*
+parameter to select unreserved outputs from the wallet, it takes a
+compulsory list of outputs to use.
+
+*utxos* must be an array of "txid:vout", each of which must be
+reserved or available: the total amount must be sufficient to pay for
+the resulting transaction plus *startweight* at the given *feerate*,
+with at least *satoshi* left over (unless *satoshi* is **all**, which
+is equivalent to setting it to zero).
+
+RETURN VALUE
+------------
+
+On success, returns the *psbt* containing the inputs, *feerate_per_kw*
+showing the exact numeric feerate it used, *estimated_final_weight* for
+the estimated weight of the transaction once fully signed, and
+*excess_msat* containing the amount above *satoshi* which is
+available.  This could be zero, or dust.  If *satoshi* was "all",
+then *excess_msat* is the entire amount once fees are subtracted
+for the weights of the inputs and *startweight*.
+
+If *reserve* was true, then a *reservations* array is returned,
+exactly like *reserveinputs*.
+
+On error the returned object will contain `code` and `message` properties,
+with `code` being one of the following:
+
+- -32602: If the given parameters are wrong.
+- -1: Catchall nonspecific error.
+- 301: Insufficient UTXOs to meet *satoshi* value.
+
+AUTHOR
+------
+
+Rusty Russell <<rusty@rustcorp.com.au>> is mainly responsible.
+
+SEE ALSO
+--------
+
+lightning-fundpsbt(7).
+
+RESOURCES
+---------
+
+Main web site: <https://github.com/ElementsProject/lightning>
--- a/doc/lightning-utxopsbt.7.md
+++ b/doc/lightning-utxopsbt.7.md
@@ -4,7 +4,7 @@ lightning-utxopsbt -- Command to populate PSBT inputs from given UTXOs
 SYNOPSIS
 --------
 
-**utxopsbt** *satoshi* *feerate* *startweight* *utxos* \[*reserve*\]
+**utxopsbt** *satoshi* *feerate* *startweight* *utxos* \[*reserve*\] \[*reservedok*\]
 
 DESCRIPTION
 -----------
@@ -22,6 +22,9 @@ reserved or available: the total amount must be sufficient to pay for
 the resulting transaction plus *startweight* at the given *feerate*,
 with at least *satoshi* left over (unless *satoshi* is **all**, which
 is equivalent to setting it to zero).
+
+Unless *reservedok* is set to true (default is false) it will also fail
+if any of the *utxos* are already reserved.
 
 RETURN VALUE
 ------------
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -48,6 +48,7 @@ MANPAGES := doc/lightning-cli.1 \
 	doc/lightning-txdiscard.7 \
 	doc/lightning-txsend.7 \
 	doc/lightning-unreserveinputs.7 \
+	doc/lightning-utxopsbt.7 \
 	doc/lightning-waitinvoice.7 \
 	doc/lightning-waitanyinvoice.7 \
 	doc/lightning-waitblockheight.7 \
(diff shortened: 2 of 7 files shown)


doc/lightning-fundchannel.7.md
```diff
@@ -52,7 +52,7 @@ private.
 *minconf* specifies the minimum number of confirmations that used
 outputs should have. Default is 1.
 
-*utxos* specifies the utxos to be used to fund the channel, as an array
+*utxos* specifies the utxos to be used to fund the channel,  as an array
```
ZmnSCPxj, line 55: useless space change?

doc/lightning-utxopsbt.7.md
```diff
@@ -0,0 +1,60 @@
...
+*utxopsbt* is a low-level RPC command which creates a PSBT using unreserved
+inputs in the wallet, optionally reserving them as well.
+
+It deliberately mirrors the parameters and output of
+lightning-fundpsbt(7) except instead of an optional *minconf*
+parameter to select unreserved outputs from the wallet, it takes a
+compulsory list of outputs to use.
+
+*utxos* must be an array of "txid:vout", each of which must be
+reserved or available: the total amount must be sufficient to pay for
```
ZmnSCPxj, line 21: I do not see a use-case for allowing `reserved` state?  If a UTXO is reserved for a different spend, we should not allow spending it: xref how `fundchannel` today works, if you `fundchannel` in parallel with the same UTXO, one will error due to being unable to reserve.

doc/lightning-utxopsbt.7.md
```diff
@@ -0,0 +1,60 @@
...
+the resulting transaction plus *startweight* at the given *feerate*,
+with at least *satoshi* left over (unless *satoshi* is **all**, which
+is equivalent to setting it to zero).
+
```
ZmnSCPxj, line 25: `*reserve*` parameter exists but is not described here.

doc/lightning-utxopsbt.7.md
```diff
@@ -0,0 +1,60 @@
...
+RETURN VALUE
+------------
+
+On success, returns the *psbt* containing the inputs, *feerate_per_kw*
+showing the exact numeric feerate it used, *estimated_final_weight* for
+the estimated weight of the transaction once fully signed, and
+*excess_msat* containing the amount above *satoshi* which is
+available.  This could be zero, or dust.  If *satoshi* was "all",
+then *excess_msat* is the entire amount once fees are subtracted
+for the weights of the inputs and *startweight*.
```
ZmnSCPxj, line 38: On Elements, do either `fundpsbt` or `utxopsbt` add a fee output?  Or does the caller have to do that?

ZmnSCPxj: If caller has to do that, I suggest adding this explicitly to the documentation, and recommending that if the output is below a dust limit (which should be documented as well) then it should be added to the fee output as well.

doc/lightning-utxopsbt.7.md
niftynei: If you're attempting an RBF, you'll very likely be spending a still `reserved` UTXO. You could require a user to 'unreserve' it first but that seems like a roundabout way to achieve the same result.

ZmnSCPxj: If you are attempting an RBF, you could just mutate the PSBT without messing with the reservations, right?  Like what we are planning with `fundchannel`/`multifundchannel`.  We get a partial transaction pre-`fundchannel_start`, then once we have the addresses we mutate that PSBT without changing reservations.  That works, right?

ZmnSCPxj: Contrasting with `fundpsbt`, `fundpsbt` does not select UTXOs that are reserved.  So I do not think it is safe to allow selection of reserved UTXOs either.

Consider the case where we have a `fundchannel`.  The user invokes it twice with two different peers but with the same UTXO set.  To protect against this, `fundchannel` has to go check the reservations by checking by `listfunds`, but this runs into atomicity problems (it could get reserved between when it queries `listfunds` and it performs `utxopsbt` by the other `fundchannel` --- or other operations like `withdraw` or novel plugins --- running in parallel).  Then the two parallel `funchannel` processes proceed, but at `sendpsbt` time only one of those will confirm, and the other channel is negotiated with the peer and useless.  The user has to manually `fundchannel_cancel` the channel that did not get in.  Now maybe you can argue that the user should not do that, but if the user is running multiple plugins on top of their `lightningd` (and we should *want* them to do that, what would be the point of implementing and maintaining plugins if we did not want them to do that?) then such conflicts are harder to handle.

See, reservations are locks.  If you violate a lock, parallel operations will screw with you.  Basic multithreading.  So we should really be a lot more serious about respecting locks.  At least we do not block when a resource (UTXO) is currently locked by somebody else, but instead fail immediately, so that we can handle this case.

rustyrussell: Agreed it's great to alter the PSBT if you have it, but if you only have the tx, you want to get the PSBT, and `utxopsbt` will do that for you.

There's a good reason that reserved-only is the default: otherwise it's up to the caller to ensure that they are entitled to extend the reservation.  See last commit.

> Consider the case where we have a fundchannel. The user invokes it twice with two different peers but with the same UTXO set. To protect against this, fundchannel has to go check the reservations by checking by listfunds, but this runs into atomicity problems (it could get reserved between when it queries listfunds and it performs utxopsbt by the other fundchannel --- or other operations like withdraw or novel plugins --- running in parallel). Then the two parallel funchannel processes proceed, but at sendpsbt time only one of those will confirm, and the other channel is negotiated with the peer and useless. The user has to manually fundchannel_cancel the channel that did not get in.

No, I wrote this code (before I wrote utxopsbt): it did listfunds and then tried reservepsbt and then could try again if it failed.  However, this was quite complex, and utxopsbt made it much simpler.

> See, reservations are locks. If you violate a lock, parallel operations will screw with you. Basic multithreading. So we should really be a lot more serious about respecting locks. At least we do not block when a resource (UTXO) is currently locked by somebody else, but instead fail immediately, so that we can handle this case.

No, they are probabolistic datastructures.  Reservations are "likely to work" inputs for creating new transactions.

This leaves significant other uses, including deliberately conflicting transactions (such as dual funding where you may prefer to do that than to be DoS out of inputs), combining transactions, RBF and cancelling transactions by deliberately double-spending.

doc/lightning-utxopsbt.7.md
rustyrussell: I don't know?  @cdecker and @niftynei have been tracking this...

ZmnSCPxj: `bitcoin_tx_finalize` does this automatically, but third parties outside of `lightningd` should probably know.

niftynei: @niftynei finally checked in on this and discovered that no, `fundpsbt` does not add a fee output for the elements case and it definitely should. made an issue #3940 to track this (it currently breaks one of our CI tests because of this)

doc/lightning-utxopsbt.7.md
rustyrussell: It's described below...

wire/Makefile
```diff
@@ -57,7 +57,8 @@ wire/gen_peer_wire_csv: wire/extracted_peer_wire_csv $(EXPERIMENTAL_PEER_PATCHES
 	@set -e; trap "rm -f $@.$$$$" 0; cp $< $@.$$$$; for exp in $(EXPERIMENTAL_PEER_PATCHES); do patch $@.$$$$ $$exp >/dev/null ; done; mv $@.$$$$ $@
 
 wire/gen_onion_wire_csv: wire/extracted_onion_wire_csv $(EXPERIMENTAL_ONION_PATCHES)
-	@set -e; trap "rm -f $@.$$$$" 0; cp $< $@.$$$$; for exp in $(EXPERIMENTAL_ONION_PATCHES); do patch $@.$$$$ $$exp; done >/dev/null ; mv $@.$$$$ $@
+	echo $(EXPERIMENTAL_ONION_PATCHES)
```
niftynei, line 60: debugging print statement?

rustyrussell: Oops, fixed!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/4d2457d363e1474332ee028bffdad5643b7aa101

Due to trivial rebase where I removed Makefile noise accidentally committed.

Issue #3855, opened 2020-07-19
Pay plugin stop to work after call "pay an_bolt11"


vincenzopalazzo:
Hi all,

I'm testing the 0.9.0-rc2, in particular my configuration include one node with 0.9.0-rc2 on desktop and two node 0.8.2 on android. I'm testing the to open channels and pay bolt11 from different node but today I run the last rc version from the code and I receive the following error

```json
{
   "code": -4,
   "message": "Plugin terminated before replying to RPC call."
}
```

With the following stakstrace
```
2020-07-19T13:00:24.463Z DEBUG 036c78a7959a62640a5f1e871d00ae0a47bb97907fc64da220dc0841d1fadd5d0c-gossipd: Received channel_update for channel 1781122x121x0/1 now ACTIVE
pay: plugins/libplugin-pay.c:1431: local_channel_hints_listpeers: Assertion `spendsats != NULL && scid != NULL && dir != NULL' failed.
pay: FATAL SIGNAL 6 (version v0.9.0rc2-15-g660a743)
0x5644f5402999 send_backtrace
	common/daemon.c:38
0x5644f5402a3f crashdump
	common/daemon.c:51
0x7f7f0cb85fcf ???
	???:0
0x7f7f0cb85f47 ???
	???:0
0x7f7f0cb878b0 ???
	???:0
0x7f7f0cb77429 ???
	???:0
0x7f7f0cb774a1 ???
	???:0
0x5644f53f1299 local_channel_hints_listpeers
	plugins/libplugin-pay.c:1431
0x5644f53e94a6 handle_rpc_reply
	plugins/libplugin.c:547
0x5644f53e9b16 rpc_read_response_one
	plugins/libplugin.c:662
0x5644f53e9c07 rpc_conn_read_response
	plugins/libplugin.c:681
0x5644f5419e25 next_plan
	ccan/ccan/io/io.c:59
0x5644f541a9a2 do_plan
	ccan/ccan/io/io.c:407
0x5644f541a9e0 io_ready
	ccan/ccan/io/io.c:417
0x5644f541cba6 io_loop
	ccan/ccan/io/poll.c:445
0x5644f53ebd9b plugin_main
	plugins/libplugin.c:1284
0x5644f53e8057 main
	plugins/pay.c:2010
0x7f7f0cb68b96 ???
	???:0
0x5644f53e1549 ???
	???:0
0xffffffffffffffff ???
	???:0
[...]

PR #3856, opened 2020-07-19
Fixed assertion in pay plugin

vincenzopalazzo:
Hi all,

I don't know if this type of PR is welcome, if no, I'm so sorry.

This PR includes the fix discussed on PR #3855. This fix was tested with the use case described inside the issue and worked.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1428,7 +1428,8 @@ local_channel_hints_listpeers(struct command *cmd, const char *buffer,
 			spendsats = json_get_member(buffer, channel, "spendable_msat");
 			scid = json_get_member(buffer, channel, "short_channel_id");
 			dir = json_get_member(buffer, channel, "direction");
-			assert(spendsats != NULL && scid != NULL && dir != NULL);
+			if(spendsats == NULL || scid == NULL || dir == NULL)
+				continue;
 
 			json_to_bool(buffer, connected, &h.enabled);
 			json_to_short_channel_id(buffer, scid, &h.scid.scid);
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -3143,3 +3143,38 @@ def test_mpp_adaptive(node_factory, bitcoind):
     from pprint import pprint
     pprint(p)
     pprint(l1.rpc.paystatus(inv))
+
+
+@pytest.mark.xfail(strict=True)
+def test_pay_fail_unconfirmed_channel(node_factory, bitcoind):
+    '''
+    Replicate #3855.
+    `pay` crash when any direct channel is still
+    unconfirmed.
+    '''
+    l1, l2 = node_factory.get_nodes(2)
+
+    amount_sat = 10 ** 6
+
+    # create l2->l1 channel.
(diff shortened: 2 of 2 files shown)


ZmnSCPxj:
It is acceptable of course.

You need a `Changelog` entry.  Since this fixes a bug that does not exist in previous version, it would be `Changelog-None`.  Just add to commit message (use `git commit --amend` to edit the commit message) then push `--force` to your branch.

You also probably want to add the text "Fixes: #3855" to the commit message and the original comment you put above, so that merging this change auto-closes #3855

I have a test to tickle this bug.

ZmnSCPxj:
Hi @vincenzopalazzo I have a branch with the proposed fixes to your commit message, as well as the test, which I have confirmed catches this bug and also confirms the fix is working. https://github.com/ZmnSCPxj/lightning/commit/1e62ac0e20496be1f8a8bdc2646ca0af5e4356df

Will push into this PR.

ZmnSCPxj:
Force pushed.  Let us leave to @cdecker for review.

vincenzopalazzo:
Hi @ZmnSCPxj 

Thanks and sorry if I missed the test. In addition, Should I add the message that you suggest in the message below?

>You need a Changelog entry. Since this fixes a bug that does not exist in previous version, it would be Changelog-None. Just add to commit message (use git commit --amend to edit the commit message) then push --force to your branch.

ZmnSCPxj:
Already added when I force-pushed into your PR.  Moar tests == better as well.

cdecker:
Excellent work @vincenzopalazzo and @ZmnSCPxj, this is very much appreciated, but I could hit myself for not thinking of this myself :-)

Anyway, this works greate and I'll ACK it as soon as CI reports that it is working :+1:

ZmnSCPxj:
CI seems to catch another bug, where we somehow trigger the paymod system to split the payment into tiny, tiny, tiny parts and then things start collapsing due to the load of handling lots of tiny tiny parts.

ZmnSCPxj:
I suggest adding a reset of some kind to `paymod`, where after we split too many times into tiny tiny pieces, we just clear the entire payment tree, erase our memory of excluded channels and nodes, and retry with the parent payment.

cdecker:
Having a split limiter is definitely a good idea, though we already have that indirectly via the fee and cltv budgets that get split into smaller and smaller pieces.

I think it's probably best not to automatically retry, since that would most likely lead down the same path. Rather we should optimize the first tree in such a way that any retry of the root would be superfluous anyway. There's just loads of parameters we can tweak here, and we're just getting started :-)

cdecker:
Rusty and I are working on fixing those other issues, reported by testers, and fix up the CI runs btw, so I'll rebase this PR on top of those so we can make sure this works as expected (which I'm pretty sure looking at the change anyway)

ZmnSCPxj:
Suggest cherry-picking this on top of #3857 and then closing this then, seems we need a bit more tests and so on for `paymod`.

cdecker:
Will do :+1:

cdecker:
Cherry-picked on top of #3857 as suggested

Issue #3832, opened 2020-07-14
`onchaind` should scorch-earth our penatly txes


ZmnSCPxj:
We should resolve these transaction by bumping their fees up at each block they are not confirmed, to counteract atempts at bribing miners to not confirm them before timeout.

My reading of the current `onchaind` code is that we do not do so.

My plan is:

* Make `steal_to_them` and `steal_htlc` use 0xFFFFFFFD (RBF) ~~`nLockTime`~~ `nSequence` wtf.
* Factor out the transaction creation code from `steal_to_them` and `steal_htlc`.
* Write an alternative to `propose_resolution` which changes the `tx` and depth but does not install an existing proposal, just reuses it.
  * Re-proposal should broadcast tx as well.
* Whenever the depth of a  transaction increases, if a  `tracked_output` for that transaction is `THEIR_REVOKED_UNILATERAL`, we re-propose with a higher feerate depending on the ~~`to_self_delay[REMOTE]`~~ ~~`to_self_delay[LOCAL]`~~ `to_self_delay[REMOTE]` (because `onchain_control.c` swaps them already) because we impose our own `to_self_delay` on the counterparty.  The easiest is just to double the feerate each time.

It seems to me that since the output tracking code of `onchaind` already needs to be resilient to changing feerate estimations, that should be sufficient to implement scorched-earth policy.

Thoughts?

ZmnSCPxj:
One of the effects here is that we can generate a transaction with high fees, possibly violating the "high fee" limit imposed by `bitcoind`.

Unfortunately the API for `sendrawtransaction` was changed in [0.19.0.1](https://bitcoin.org/en/release/v0.19.0.1#updated-rpcs):

> * `sendrawtransaction` and `testmempoolaccept` no longer accept a `allowhighfees` parameter to fail mempool acceptance if the transaction fee exceeds the value of the configuration option `-maxtxfee`. Now there is a hardcoded default maximum feerate that can be changed when calling either RPC using a `maxfeerate` parameter. ([#15620](https://github.com/bitcoin/bitcoin/pull/15620))

I ***think*** what we can do would be:

* Add a `allowhighfees` parameter to the `bcli` `sendrawtransaction`.
* If the `allowhighfees` parameter in `bcli` is true, add a second parameter that is simply `1`.
   * pre-0.19.0.1 would interpret this as `allowhighfees=true`.
[...]

PR #3870, opened 2020-07-23
`onchaind` now scorches the earth when penalizing a revoked transaction

ZmnSCPxj:
Fixes: #3832


cdecker:
This looks great, can't wait to dive in and review it, once the release is out the door 🙂

ZmnSCPxj:
Rebased, also increased test coverage a little by also checking that an RBFed transaction is considered resolved by the `onchaind`.

ZmnSCPxj:
An issue is that this changes the interface of `snedrawtransaction`. Older plugins which provide `sendrawtransaction` will not be compatible after this.  However, a possible change would be to check if `sendrawtransaction` with two arguments results in a `JSONRPC_INVALID_PARAMS` error, and if so, try again with just a single argument, and if this is an RBFed transaction, to print a warning in the log that the plugin is using the old interface and the RBFed transaction may not get forwarded due to too high fee.

ZmnSCPxj:
Did what I suggested, which is, if `deprecated_apis`, try to use a single-arg `sendrawtransaction` if the two-arg `sendrawtransaction` errors complaining about parameters.

ZmnSCPxj:
Rebased.  Also, changed the Changelog regarding two-argument `sendrawtransaction` to "deprecated" (it was "removed"), since we now continue to support the one-argument `sendrawtransaction` if `--deprecated-apis` is set.

ZmnSCPxj:
Rebased on top of b934a92 from #3889

ZmnSCPxj:
Rebased, especially with #3889 merged in.

ZmnSCPxj:
***bump***

ZmnSCPxj:
Rebased on master.  Also, ***bump***.

ZmnSCPxj:
Rebased.  Also, fixed the capstone --- if we make a penalty transaction that donates the money to miners (because the locktime has been reached) then it previously would make a transaction smaller than the 82-byte limit. We now use a padded `OP_RETURN` with a pointless 20-byte random commitment to get over the 82-byte limit.

cdecker:
Assigning @rustyrussell since `onchaind` is his brainchild :-)

rustyrussell:
I am going to apply this, though the "real answer" is to have a general engine which makes these kind of "deadline" decisions.  That will require careful design, and meanwhile this fixes a real issue.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/ee9d85eb8703ca35729eaef9eb3ff34a59ef7a6c

rustyrussell:
(If @ZmnSCPxj doesn't get to it first, I'll rebase and apply...)

rustyrussell:
Fairly trivial rebase.

rustyrussell:
... another trivial rebase.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/8d6442a5417df9e0ef4ff5886f9e538a5d3d9cdf

PR #3871, opened 2020-07-23
Iterate through all the routehints in the specified

cdecker:
This PR re-establishes the old behavior of testing each routehint in the order they were specified, but keeps the routehints immutable, so each sub-payment decides whether to skip one or not:

 - The root payment performs an initial `getroute` test for 1sat to see if the destination could be reached directly, if that's the case it add a `NULL` routehint to the end of the routehint array. This way using or not using a routehint is handled identically.
 - Each payment has an offset internally that indexes into the root routehint array. When starting a new attempt it'll select the next working routehint by skipping the ones that might be known to fail due to a channel hint.
   - On retry the child payment increments the parent's offset by 1 and continues from there
   - On split the child payment reset the offset to 0, so it and its children iterate from the start, considering all routehints since the constraints have changed (amount has changed)
 - Since the retry mod has a circuit breaker that does not retry if the call to `getroute` failed we have to signal that we have more options to try before retries should terminate.

Since I was at it I also added a bit more debug information to see what is going on, and the tests finally work :+1:

Changelog-None

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1820,6 +1820,33 @@ static u32 route_cltv(u32 cltv,
 	return cltv;
 }
 
+/* Change the destination and compute the final msatoshi amount to send to the
+ * routehint entry point. */
+static void routehint_pre_getroute(struct routehints_data *d, struct payment *p)
+{
+	d->current_routehint = next_routehint(d, p);
+	if (d->current_routehint != NULL) {
+		if (!route_msatoshi(&p->getroute->amount, p->amount,
+				    d->current_routehint,
+				    tal_count(d->current_routehint))) {
+		}
+		d->final_cltv = p->getroute->cltv;
+		p->getroute->destination = &d->current_routehint[0].pubkey;
+		p->getroute->cltv =
+		    route_cltv(p->getroute->cltv, d->current_routehint,
+			       tal_count(d->current_routehint));
+		plugin_log(
+		    p->plugin, LOG_DBG, "Using routehint %s (%s) cltv_delta=%d",
+		    type_to_string(tmpctx, struct node_id,
+				   &d->current_routehint->pubkey),
+		    type_to_string(tmpctx, struct short_channel_id,
+				   &d->current_routehint->short_channel_id),
+		    d->current_routehint->cltv_expiry_delta);
+	} else {
+		plugin_log(p->plugin, LOG_DBG, "Not using a routehint");
+	}
+}
+
 static struct command_result *routehint_getroute_result(struct command *cmd,
 							const char *buffer,
 							const jsmntok_t *toks,
@@ -1836,7 +1863,7 @@ static struct command_result *routehint_getroute_result(struct command *cmd,
 	if (d->destination_reachable)
 		tal_arr_expand(&d->routehints, NULL);
 
-	d->current_routehint = next_routehint(d, p);
+	routehint_pre_getroute(d, p);
 
 	plugin_log(p->plugin, LOG_DBG,
 		   "The destination is%s directly reachable %s attempts "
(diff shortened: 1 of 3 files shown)


rustyrussell:
i.e. this seems to work fine:

```diff
From d55529a048c90c265b57eca9e57c5fbea6b7fead Mon Sep 17 00:00:00 2001
From: Rusty Russell <rusty@rustcorp.com.au>
Date: Fri, 24 Jul 2020 10:06:04 +0930
Subject: [PATCH] plugins/pay: remove route probe, just always add NULL routehint at end.

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

diff --git a/plugins/libplugin-pay.c b/plugins/libplugin-pay.c
index 27669b125..392ef70eb 100644
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1874,58 +1874,6 @@ static void routehint_pre_getroute(struct routehints_data *d, struct payment *p)
        }
 }
 
-static struct command_result *routehint_getroute_result(struct command *cmd,
-                                                       const char *buffer,
-                                                       const jsmntok_t *toks,
-                                                       struct payment *p)
-{
-       struct payment *root = payment_root(p);
-       const jsmntok_t *rtok = json_get_member(buffer, toks, "route");
-       struct routehints_data *d = payment_mod_routehints_get_data(root);
-
-       /* If there was a route the destination is reachable without
-        * routehints. */
-       d->destination_reachable = (rtok != NULL);
-
-       if (d->destination_reachable)
-               tal_arr_expand(&d->routehints, NULL);
-
-       routehint_pre_getroute(d, p);
-
-       plugin_log(p->plugin, LOG_DBG,
-                  "The destination is%s directly reachable %s attempts "
-                  "without routehints",
-                  d->destination_reachable ? "" : " not",
-                  d->destination_reachable ? "including" : "excluding");
-
-       /* Now we can continue on our merry way. */
-       payment_continue(p);
-
-       /* Let payment_finished_ handle this, so we mark it as pending */
-       return command_still_pending(cmd);
-}
-
-static void routehint_check_reachable(struct payment *p)
-{
-       struct out_req *req;
-       /* Start a tiny exploratory getroute request, so we
-        * know whether we stand any chance of reaching the
-        * destination without routehints. This will later be
-        * used to mix in attempts without routehints. */
-       req = jsonrpc_request_start(p->plugin, NULL, "getroute",
-                                   routehint_getroute_result,
-                                   routehint_getroute_result, p);
-       json_add_node_id(req->js, "id", p->destination);
-       json_add_amount_msat_only(req->js, "msatoshi", AMOUNT_MSAT(1000));
-       json_add_num(req->js, "maxhops", 20);
-       json_add_num(req->js, "riskfactor", 10);
-       send_outreq(p->plugin, req);
-       plugin_log(p->plugin, LOG_DBG,
-                  "Asking gossipd whether %s is reachable "
-                  "without routehints.",
-                  type_to_string(tmpctx, struct node_id, p->destination));
-}
-
 static void routehint_step_cb(struct routehints_data *d, struct payment *p)
 {
        struct route_hop hop;
@@ -1946,9 +1894,9 @@ static void routehint_step_cb(struct routehints_data *d, struct payment *p)
                                   "After filtering routehints we're left with "
                                   "%zu usable hints",
                                   tal_count(d->routehints));
-                           /* Do not continue normally, instead go and check if
-                            * we can reach the destination directly. */
-                           return routehint_check_reachable(p);
+
+                       /* Append a direct attempt for last one */
+                       tal_arr_expand(&d->routehints, NULL);
                }
 
                routehint_pre_getroute(d, p);
@@ -1996,7 +1944,6 @@ static struct routehints_data *routehint_data_init(struct payment *p)
        d->current_routehint = NULL;
        if (p->parent != NULL) {
                pd = payment_mod_routehints_get_data(payment_root(p));
-               d->destination_reachable = pd->destination_reachable;
                d->routehints = pd->routehints;
                if (p->parent->step == PAYMENT_STEP_RETRY)
                        d->offset = pd->offset + 1;
diff --git a/plugins/libplugin-pay.h b/plugins/libplugin-pay.h
index b6e3a52be..74321912b 100644
--- a/plugins/libplugin-pay.h
+++ b/plugins/libplugin-pay.h
@@ -314,9 +314,6 @@ struct routehints_data {
        /* We modify the CLTV in the getroute call, so we need to remember
         * what the final cltv delta was so we re-apply it correctly. */
        u32 final_cltv;
-
-       /* Is the destination reachable without any routehints? */
-       bool destination_reachable;
 };
 
 struct exemptfee_data {
```

ZmnSCPxj:
> It seems to me that you can just append NULL to the routehints unconditionally? If you can't route, you'll find out then. What am I missing?

I agree as well.

PR #3879, opened 2020-07-26
pay: always send onion error message to gossipd.

rustyrussell:
There were no channel updates in my log; because sendonion doesn't know the
actual node_ids or channel_ids, we can't tell gossipd what node/channel it was
so it can no longer remove them on PERM errors.

However, we can tell it the error message so it can apply the update.

Fixes: #3877
Changelog-None


gossipd/routing.c
```diff
@@ -2762,96 +2762,6 @@ struct route_hop **get_route(const tal_t *ctx, struct routing_state *rstate,
 	return hops;
 }
 
-void routing_failure(struct routing_state *rstate,
```
ZmnSCPxj, line 2765: Well that is a lot of code removed.  I suppose these are just holdovers of the old days when `pay` was far less sophisticated.

cdecker:
Thanks @rustyrussell, this looks great. I'm surprised that the old `sendpay` path doesn't touch these removes parts anymore.

Issue #3880, opened 2020-07-26
listpays rpc method could expose the payment_hash if the bolt11 is null (keysend)


vincenzopalazzo:
Hi all.

During my work on introducing a new command `delpay` (good fist issue), I noted that after the evolution of pay plugin, the method RPC `listpays` could return payment without `id`, such as the `bolt11`  value null if is used keysend plugin.

In cases of `bolt11` value null, I can use the command `listsendpays` to have a result with more details. However, if I don't have a fault on my idea the command `listpays` could be able to return the `bolt11` in cases of payment with invoice and `payment_hash` in cases of payment without invoice. 

An example of output could be the json below

### keysend payment

```json
{
      "payment_hash": "7cdf931c7a23efb6815ff239517ef6d373bb8d040caefba96314368afd93252d",
      "status": "complete",
      "preimage": "9a4f8576e0380394ab791b35e08793533b7d1d3303c383d91fda14a13978e7db",
      "amount_sent_msat": "1000msat"
  },
```

### invoice payment

```json
      {
         "bolt11": "lntb10n1p03cskapp5f730rvqpqelvqmtljkuxjku2ek00qnqmf5g3pcaefc06q6rmk8sqdp9ve5hsatsta3x7mr5xyckuatvd30hgam9d3mx2xqyjw5qcqp2sp5nh2swvkuup2qrekvlmstswh8kdjjh7chyuelp7sgy0hvtz9dzhqqrzjqdsdeghn2vmdxqmy8flmzu46vxzmjzr258aavp36z3rs2redmg8cwxe0fyqqq2qqqqqqqqqpqqqqqzsqqc9qy9qsqu7l2057sh8zvpw56ftu87semua2a6a4ftjscvvfewz0gj92dsylx7whv30heqy0zr6jy6udtad57m93q3pmyezfcf4pntjqstks0v5qpedegh0",
         "status": "complete",
         "preimage": "7516ea2e790e5e8edb02709682e445d38aea07c7e772db9441b5091908d19c23",
         "amount_sent_msat": "1000msat"
      }
```

The content above make sense? or use a different json format inside the same rpc command and it is not correct?

reference implementation ~https://github.com/vincenzopalazzo/lightning/commit/7c13b0be79ff659597e345ad202b6fb692c5f57a~
https://github.com/vincenzopalazzo/lightning/commit/b59f72b201cde197f4b94e53e8d4df1f44348451

vincenzopalazzo:
[...]

PR #3878, opened 2020-07-25
A proposed solution to fix json null value for bolt11 key when is used the pay plugin

vincenzopalazzo:
This PR proposed a solution to fix the problem mentioned at the following PR https://github.com/ElementsProject/lightning/pull/3876 

I can have some idea bug in my mind, so if this PR is wrong, I'm so sorry.

The json below is an example of use cases when I used the command keysend and pay (json_paymod).

```json
{
   "pays": [
      {
         "bolt11": "lnbcrt100n1p03chsnpp5n72j2e0ykwz7ue37e60qn623z9elq4fguv52ny7jjqmv990zajeqdql2pshjmt9de6zqamfw35zqurp09kk2eqxqyjw5qcqp2sp5h0renl745kd7c86z2g49lkuhua3j66an80szd58qdpeyzhdkrl3srzjqfnup6dhq3xls5v5pw0ksgrx0gcerqpr2pd6y6ckhstql6qfncta6qq24qqqqqgqqqqqqqqpqqqqqzsqqc9qy9qsqfy8e5x6aya75n3ayd5e8rsvkmepgq9gwf8v9ram560y5vlrapk9pwjxcfewfpvvnw97v49xzlce6zqmc5znv0ke6e3un5lm4hk8uewqq5dug0s",
         "status": "complete",
         "preimage": "c0c09fb65814b71f56c0a0faa9bf223b1d9667df90b55648a7c023e200c66b70",
         "amount_sent_msat": "10000msat"
      },
      {
         "bolt11": "(null)",
         "status": "failed",
         "label": "Pay with keysend",
         "amount_sent_msat": "0msat"
      },
      {
         "bolt11": "lnbcrt50n1p0s5zvvpp5ylm3y040va7wu08tjw7kegzq0ku2zqav99u6tzfdd7mh6w8elecsdpsg3jhvgz5v4ehggrwv4mjqunsvvsxxmmdd4skuepqw3jhxapnxqyjw5qcqp2sp5jxrwha4g0rjdxf3cs8gc9rqs5zunu3670v5mpppw0zajw7f4yrsq9qy9qsqf043ey32l952hq6rry34r9vugh2kdydr7hnuujpfadt5gxykguf5lsj8rczr7c3kpc6e0ldjw3r3nwehetwfd2f4djwtnjztjwzrnjqqaj8fg2",
         "status": "complete",
         "preimage": "391f288bc4425328742aafe26a01d02a07ce8b7420003aacd3ef4b2741284523",
[...]

PR #3881, opened 2020-07-27
Annotate payments initiated using sendonion with the amount at the destination

cdecker:
Since we use `sendonion` in the new payment flow we end up not knowing the amount received by the destination, only the amount we sent (which includes fees).

This PR adds an optional annotation that tells us what the intended amount to be sent is.

Notice that it is still possible to call `sendonion` without the annotation and readers calling `listpays` or `listsendpays` should fall back to the amount being sent if the amount at the destination is unknown. The amount_sent is always guaranteed to be known and set.

Closes https://github.com/shesek/spark-wallet/issues/146
Changelog-None

### DIFF
--- a/doc/lightning-sendonion.7.md
+++ b/doc/lightning-sendonion.7.md
@@ -4,7 +4,7 @@ lightning-sendonion -- Send a payment with a custom onion packet
 SYNOPSIS
 --------
 
-**sendonion** *onion* *first_hop* *payment_hash* \[*label*\] \[*shared_secrets*\] \[*partid*\] \[*bolt11*\]
+**sendonion** *onion* *first_hop* *payment_hash* \[*label*\] \[*shared_secrets*\] \[*partid*\] \[*bolt11*\] \[*msatoshi*\]
 
 DESCRIPTION
 -----------
@@ -77,6 +77,9 @@ partial payments with the same *payment_hash*.
 
 The *bolt11* parameter, if provided, will be returned in
 *waitsendpay* and *listsendpays* results.
+
+The *msatoshi* parameter is used to annotate the payment, and is returned by
+*waitsendpay* and *listsendpays*.
 
 RETURN VALUE
 ------------
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1179,6 +1179,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 	struct lightningd *ld = cmd->ld;
 	const char *label, *b11str;
 	struct secret *path_secrets;
+	struct amount_msat *msat;
 	u64 *partid;
 
 	if (!param(cmd, buffer, params,
@@ -1189,6 +1190,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt("shared_secrets", param_secrets_array, &path_secrets),
 		   p_opt_def("partid", param_u64, &partid, 0),
 		   p_opt("bolt11", param_string, &b11str),
+		   p_opt_def("msatoshi", param_msat, &msat, AMOUNT_MSAT(0)),
 		   NULL))
 		return command_param_failed();
 
@@ -1201,7 +1203,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 				    failcode);
 
 	return send_payment_core(ld, cmd, payment_hash, *partid,
-				 first_hop, AMOUNT_MSAT(0), AMOUNT_MSAT(0),
+				 first_hop, *msat, AMOUNT_MSAT(0),
 				 label, b11str, &packet, NULL, NULL, NULL,
 				 path_secrets);
 }
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1661,6 +1661,11 @@ struct pay_mpp {
 	size_t num_nonfailed_parts;
 	/* Total amount sent ("complete" or "pending" only). */
 	struct amount_msat amount_sent;
+
+	/* Total amount received by the recipient ("complete" or "pending"
+	 * only). Null if we have any part for which we didn't know the
+	 * amount. */
+	struct amount_msat *amount;
 };
 
 static const struct sha256 *pay_mpp_key(const struct pay_mpp *pm)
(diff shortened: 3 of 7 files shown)


vincenzopalazzo:
I agree with changs inside this PR, the normal user (like me) can feel a little confused when see the amount propriety null.

wallet/wallet.c
```diff
@@ -2749,8 +2749,16 @@ static struct wallet_payment *wallet_stmt2payment(const tal_t *ctx,
 	else
 		payment->failonion = NULL;
 
-	db_column_amount_msat(stmt, 14, &payment->total_msat);
-	payment->partid = db_column_u64(stmt, 15);
+	if (!db_column_is_null(stmt, 14))
+		db_column_amount_msat(stmt, 14, &payment->total_msat);
+	else
+		payment->total_msat =AMOUNT_MSAT(0);
```
ZmnSCPxj, line 2755: Style nit: space after `=`.

lightningd/pay.c
```diff
@@ -1192,6 +1193,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt("shared_secrets", param_secrets_array, &path_secrets),
 		   p_opt_def("partid", param_u64, &partid, 0),
 		   p_opt("bolt11", param_string, &b11str),
+		   p_opt_def("msatoshi", param_msat, &msat, AMOUNT_MSAT(0)),
```
ZmnSCPxj, line 1196: New param deserves documenting in `doc/lightning-sendonion.7.md`.

cdecker: Righto, will amend

cdecker:
> Looks good to me, though i did not follow the latest change to our pay logic closely enough to ack :-)

I should probably add a bit more context: since we switched from the `pay` plugin using `sendpay` to using `sendonion` the main daemon no longer has a way to infer the amount received by the recipient purely from the route (which in `sendpay` is passed in as cleartext). This fills that gap, allowing us to return the amount received by the recipient from this (partial-)payment, if the payment was successful.

One of the things we broke downstream was spark-wallet (https://github.com/shesek/spark-wallet/issues/146) which uses `listsendpay` to list recent payment (it should switch to `listpays` as well see https://github.com/shesek/spark-wallet/issues/147), and it was using that field to display the amount in the details and list for the payment.

vincenzopalazzo:
I can share a my generic think, I'm working to test it but I think is a good idea share it in this PR.

You mentioned the command `listpays`, I noted that the command return a bolt11 null when is used the keysend command. Do you think is a problem? or is it only a fault in my mind?

My complete idea shared [here](https://github.com/ElementsProject/lightning/issues/3880)

plugins/pay.c
```diff
@@ -1708,6 +1739,11 @@ static void add_new_entry(struct json_stream *ret,
 		json_add_tok(ret, "label", pm->label, buf);
 	if (pm->preimage)
 		json_add_tok(ret, "preimage", pm->preimage, buf);
+
+	if (pm->amount != NULL)
+		json_add_string(ret, "amount_msat",
+				fmt_amount_msat(tmpctx, &pm->amount_sent));
```
rustyrussell, line 1745: Hmm, this will print 0 for failed payments (since it's allocated and never updated).

Probably need to gate this on "status" != "failed"?

doc/lightning-sendonion.7.md
```diff
@@ -4,7 +4,7 @@ lightning-sendonion -- Send a payment with a custom onion packet
 SYNOPSIS
 --------
 
-**sendonion** *onion* *first_hop* *payment_hash* \[*label*\] \[*shared_secrets*\] \[*partid*\] \[*bolt11*\]
+**sendonion** *onion* *first_hop* *payment_hash* \[*label*\] \[*shared_secrets*\] \[*partid*\] \[*bolt11*\] \[*msatoshi*\]
```
rustyrussell, line 7: Needs make doc-all.

rustyrussell:
We also need to stage the removal of the null field, since it's an API break...  I can do this too.

lightningd/pay.c
```diff
@@ -97,9 +97,6 @@ void json_add_payment_fields(struct json_stream *response,
 	if (amount_msat_greater(t->msatoshi, AMOUNT_MSAT(0)))
 		json_add_amount_msat_compat(response, t->msatoshi, "msatoshi",
 					    "amount_msat");
-	else
-		json_add_null(response, "amount_msat");
-
```
rustyrussell, line 102: Needs a compat gate, and 6 month deprecation cycle.

vincenzopalazzo:
@rustyrussell if you mean the bolt11 null, I wrote a little code and I'm waiting to make a test with my testnet network. https://github.com/vincenzopalazzo/lightning/commit/b59f72b201cde197f4b94e53e8d4df1f44348451

rustyrussell:
> @rustyrussell if you mean the bolt11 null, I wrote a little code and I'm waiting to make a test with my testnet network. [vincenzopalazzo@b59f72b](https://github.com/vincenzopalazzo/lightning/commit/b59f72b201cde197f4b94e53e8d4df1f44348451)

That's different (though a good idea for next release): I was referring to the old `"amount_msat": null` which this series removed.  That's an API break, and we require a 6 month notification for those by policy.  I am worried it will (further!) break Spark...

Thanks!

rustyrussell:
OK, I dropped the commit which removed the ": null" ( json-rpc: Do not return a null amount_msat to listpays ) because it really should be done via a deprecation cycle, which means we should really pass "allow-deprecated-apis" to getmanifest, and that's too big a change for an -rc.  I'll queue as a separate PR...

rustyrussell:
Also, I prefer not to push to branches on this repo, because Travis tests them twice...

d4amenace:
Hello all...

So to my understanding,  sendpay is being used instead of pay call for the forseeable future? or temporarily?  (This has also broken my front end for my app in development)

ZmnSCPxj:
No, `pay` should still exist.  If it does not exist, it probably means the `pay` plugin crashed,  Try checking for `plugin-pay` in your `lightningd` logs.  It should mention that the `plugin-pay` started up, and if it did not, something is incomplete in your install.  Alternately, it might mention `plugin-pay` started up, but for some reason crashed, and we would want to know the reason for the crash.

d4amenace:
> No, `pay` should still exist. If it does not exist, it probably means the `pay` plugin crashed, Try checking for `plugin-pay` in your `lightningd` logs. It should mention that the `plugin-pay` started up, and if it did not, something is incomplete in your install. Alternately, it might mention `plugin-pay` started up, but for some reason crashed, and we would want to know the reason for the crash.

resolved issue with pay plugin but......

d4amenace:
Ok, so new node. Neither listsendpays nor listpays give complete payment info...now missing bolt11 invoice on listpays....and on listsendpays also no bolt 11 invoice??

minipc@minipc-desktop:$ lightning-cli listsendpays
{
"payments": [
{
"id": 1,
"payment_hash": "3c43486e64284cd5d17e8a38bee76b42310bf195adf93d56cf88658929516a0e",
"amount_msat": null,
"msatoshi_sent": 10000000,
"amount_sent_msat": "10000000msat",
"created_at": 1596002463,
"status": "complete",
"payment_preimage": "7fe29f11eff1960ddfcc58268ad5dc6dedc057d909796f875563c1e6cc3619a1"
}
]
}
minipc@minipc-desktop:$ lightning-cli listpays
{
"pays": [
{
"bolt11": "(null)",
"status": "complete",
"preimage": "7fe29f11eff1960ddfcc58268ad5dc6dedc057d909796f875563c1e6cc3619a1",
"amount_sent_msat": "10000000msat"
}
]
}

Older node gives full information under listsendpays:
{
"id": 8,
"payment_hash": "0bf817e93e8dfe0fe24593c85e2c14923f6c14747679ff5dd8cb6628793a2c80",
"destination": "02dd238ecc1d1c1798025a78caac1e20c5945e97836bdf01b25449d0fef776f456",
"msatoshi": 5001050,
"amount_msat": "5001050msat",
"msatoshi_sent": 5001050,
"amount_sent_msat": "5001050msat",
"created_at": 1596002609,
"status": "complete",
"payment_preimage": "eb0a83d46aea5acd872e154115c112345f4393f4e7b0afca885aedf1d3379afb",
"bolt11": "lnbc50u1p0jzyg6pp5p0up06f73hlqlcj9j0y9utq5jglkc9r5weul7hwcednzs7f69jqqdqcve5hyum5d9hxvun0d45x7mt9xqyjw5qcqp2sp5skxag5u3v54d6j3px6clldpyzpeyntsh69fhf6xa548mlljx6mpsrzjqtwgtyxavad4h7yud00eamkhvu5sk0tq2er9ukcpxat0v4sk60fhyzwfqvqqxlcqqyqqqqlgqqqqqzsqpc9qy9qsqevvddx89zceda2nf3qecxuvzeljy8a937v5n4gs8fu9h908c5uq53l4czysqcq7yzy306x5epgkn6568axxzzd29hdw83nvsshud6lgpfxdalp"
}
]
}

Either listsendpays or listpays MUST include bolt11,destination and payment hash. App is using listsendpays to list all payments sent...cannot identify without any identifiers.

Need to call listsendpays or listpays with bolt 11 parameter....neither tracks so how can lookup particular invoice? (used by app to confirm payments after executing pay call)

ZmnSCPxj:
This should have been fixed by #3878 which was merged into master two days ago.  What version are you using?

d4amenace:
> This should have been fixed by #3878 which was merged into master two days ago. What version are you using?

0.9.0 rc3

ZmnSCPxj:
rc3 is known to be buggy.  It seems the bug you are reporting is already fixed in master.  Can you try uninstalling and reinstalling again with master?

d4amenace:
> rc3 is known to be buggy. It seems the bug you are reporting is already fixed in master. Can you try uninstalling and reinstalling again with master?

Is v0.9.0rc3-10-g14d08b7  current master?

PR #3882, opened 2020-07-27
pay: Be less aggressive when estimating channel capacity

cdecker:
We'd previously take the failed attempt and estimate the failing channel's
capacity at 3/4 of the attempted amount, which is rather aggressive. This
reduces this aggressiveness to use the exact amount tried, but excluding on
equality. This still skips attempting the same route with the same amount, but
also permits attempts that are in the range [3/4, 1] of the failed attempt
amount to still be attempted.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -186,7 +186,7 @@ void payment_start(struct payment *p)
 static void channel_hints_update(struct payment *p,
 				 const struct short_channel_id scid,
 				 int direction, bool enabled, bool local,
-				 struct amount_msat *estimated_capacity,
+				 const struct amount_msat *estimated_capacity,
 				 u16 *htlc_budget)
 {
 	struct payment *root = payment_root(p);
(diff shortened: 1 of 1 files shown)


cdecker:
> Code is OK, though I have my reservations on being so sanguine as to consider the lowest failed value as a limit. Still, it is not as if there has been actual experimental research on the optimal splitting of payments into multiple parts, so in the spirit of experimentation, let us do it.

Agreed that this might be too permissive, and result in more failed attempts than necessary, however I went with a strict interpretation of "hint" here: the estimate doesn't say "we know that this value will work" which was my previous interpretation, but rather "we know that the real current capacity is below this". Any sort of setting of the estimate to a lower value is essentially guessing.

Notice also that this doesn't mean we will attempt with `estimate - 1msat`, but rather the split steps are determined by the other payment mods. This is more permissive in that we don't accidentally exclude a channel that might have worked, which would result in more attempts for smaller sizes in the best case, or an aborted payment in the worst case.

PR #3886, opened 2020-07-29
Backfill missing scriptpubkeys

niftynei:
PSBTs require us to have the scriptpubkeys, which we use to fill in the witness-utxo objects on the inputs (which we then use as the transport mechanism for amounts).

Prior to 0.7.3 we weren't caching the scriptpubkey in the database, so any node attempting to spend a UTXO recorded prior to this release will crash.

To fix this, we backfill the scriptpubkey entries in the database, which in turn allows us to remove the NULL handling for them throughout.

Reported-By: @nothingmuch
Changelog-None


wallet/test/run-wallet.c
```diff
@@ -980,6 +983,8 @@ static bool test_wallet_outputs(struct lightningd *ld, const tal_t *ctx)
 	u.close_info->channel_id = 42;
 	u.close_info->peer_id = id;
 	u.close_info->commitment_point = NULL;
+	u.scriptPubkey = tal_arrz(w, u8, 20);
```
ZmnSCPxj, line 986: Why `talarrz` if we are going to fill the entire array with `1`s later?

rustyrussell:
> ACK [920939a](https://github.com/ElementsProject/lightning/commit/920939ae8b01e2344256911daa7681cf9df1a024)
> 
> Minor style nit tho.
> 
> Also, no `Changelog-` entry.

Yes, if this makes it into -rc4 then it's a fix for a regression from 0.8.3.  If it has to wait for 0.9.1, it needs a Changelog?

wallet/test/run-wallet.c
rustyrussell: Trivial fix pushed.

nothingmuch:
I'm still wrapping my head around the code so can't meaningfully ACK, but it does make sense to me, and it has resolved my crashes when attempting to fund a channel.

wallet/test/run-wallet.c
ZmnSCPxj: Hmmmm does it fix this? confused

ZmnSCPxj: Ah, your fix only fixes one instance of this, this is another one.

rustyrussell: Damn, will patch once kids in bed...

cdecker:
Verified by updating my own node and all missing scriptpubkeys were correctly backfilled :+1:

ACK https://github.com/ElementsProject/lightning/pull/3886/commits/0a3ff50d65ef52506e0423cf7df2e5b20c67fddf

wallet/test/run-wallet.c
cdecker: Fixed :wink:

PR #3887, opened 2020-07-29
pytest: fix flake in tests/test_pay.py::test_pay_exclude_node

rustyrussell:
```
        # Excludes channel, then ignores routehint which includes that, then
        # it excludes other channel.
>       assert len(status) == 2
E       assert 1 == 2
E         -1
E         +2
```

The invoice we use at the end has a routehint: 50% of the time it's
to l2 (which fails), 50% to l5 (which succeeds).

Change it to create invoice before channel with l5 so it does the
retry like we expect here.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3887/commits/ac1c0e1287c1958367a5673c91af85c51cba30c5

PR #3888, opened 2020-07-29
Update format response from listpay is the payment was did with keysend

vincenzopalazzo:
Hi all.

This PR proposes an update to the response of the `listpay` command, discussed inside [this issue](https://github.com/ElementsProject/lightning/issues/3880). In particular, the `listpays` could return a `bolt11: (null)` propriety if the payment was made with `keysend` or if the payment was didi during the [rc3 bug](https://github.com/ElementsProject/lightning/pull/3878).

In addition, this PR proposes the change inside the parameters accepted by `listpay`. In fact, the command could lose the meaning if I want to use `listpay` to filter the payment that I made with `keysend`. In other words, I can not require with `listpays` a single payment if it is made with `keysend`.

### keysend payment

```json
{
      "payment_hash": "7cdf931c7a23efb6815ff239517ef6d373bb8d040caefba96314368afd93252d",
      "status": "complete",
      "preimage": "9a4f8576e0380394ab791b35e08793533b7d1d3303c383d91fda14a13978e7db",
      "amount_sent_msat": "1000msat"
  },
```

### invoice payment

```json
      {
         "bolt11": "lntb10n1p03cskapp5f730rvqpqelvqmtljkuxjku2ek00qnqmf5g3pcaefc06q6rmk8sqdp9ve5hsatsta3x7mr5xyckuatvd30hgam9d3mx2xqyjw5qcqp2sp5nh2swvkuup2qrekvlmstswh8kdjjh7chyuelp7sgy0hvtz9dzhqqrzjqdsdeghn2vmdxqmy8flmzu46vxzmjzr258aavp36z3rs2redmg8cwxe0fyqqq2qqqqqqqqqpqqqqqzsqqc9qy9qsqu7l2057sh8zvpw56ftu87semua2a6a4ftjscvvfewz0gj92dsylx7whv30heqy0zr6jy6udtad57m93q3pmyezfcf4pntjqstks0v5qpedegh0",
         "status": "complete",
         "preimage": "7516ea2e790e5e8edb02709682e445d38aea07c7e772db9441b5091908d19c23",
         "amount_sent_msat": "1000msat"
      }
```

In addition, I see inside the format description of the documentation, that reported an old description for the old payment. I think that this description could be removed or ~signed~.

P.S: Maybe, in this PR the signature `Changelog-None` is not correct?

## Update

After a first review, the response format is changed, now the `payment_hash` is printed every time and the bolt11 is printed only if the value is not `null`

### DIFF
--- a/doc/lightning-listpays.7.md
+++ b/doc/lightning-listpays.7.md
@@ -4,13 +4,13 @@ lightning-listpays -- Command for querying payment status
 SYNOPSIS
 --------
 
-**listpays** \[bolt11\]
+**listpays** \[bolt11\] \[payment_hash\]
 
 DESCRIPTION
 -----------
 
 The **listpay** RPC command gets the status of all *pay* commands, or a
-single one if *bolt11* is specified.
+single one if either *bolt11* or *payment_hash* was specified.
 
 RETURN VALUE
 ------------
@@ -18,16 +18,19 @@ RETURN VALUE
 On success, an array of objects is returned. Each object contains:
 
  *bolt11*
-the *bolt11* argument given to *pay* (see below for exceptions).
+the *bolt11* invoice if provided to `pay`.
+
+ *payment_hash*
+the *payment_hash* of the payment.
 
  *status*
 one of *complete*, *failed* or *pending*.
 
  *payment\_preimage*
-(if *status* is *complete*) proves payment was received.
+if *status* is *complete*.
 
  *label*
-optional *label*, if provided to *pay*.
+optional *label*, if provided to *pay* or *sendonion*.
 
  *amount\_sent\_msat*
 total amount sent, in "NNNmsat" format.
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1178,6 +1178,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 	struct sha256 *payment_hash;
 	struct lightningd *ld = cmd->ld;
 	const char *label, *b11str;
+	struct node_id *destination;
 	struct secret *path_secrets;
 	struct amount_msat *msat;
 	u64 *partid;
@@ -1191,6 +1192,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt_def("partid", param_u64, &partid, 0),
 		   p_opt("bolt11", param_string, &b11str),
 		   p_opt_def("msatoshi", param_msat, &msat, AMOUNT_MSAT(0)),
+		   p_opt("destination", param_node_id, &destination),
 		   NULL))
 		return command_param_failed();
 
@@ -1204,7 +1206,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 
 	return send_payment_core(ld, cmd, payment_hash, *partid,
 				 first_hop, *msat, AMOUNT_MSAT(0),
-				 label, b11str, &packet, NULL, NULL, NULL,
+				 label, b11str, &packet, destination, NULL, NULL,
 				 path_secrets);
 }
 
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1669,6 +1669,9 @@ struct pay_mpp {
 
 	/* Timestamp of the first part */
 	u32 timestamp;
+
+	/* The destination of the payment, if specified. */
+	struct node_id *destination;
 };
 
 static const struct sha256 *pay_mpp_key(const struct pay_mpp *pm)
@@ -1739,6 +1742,9 @@ static void add_new_entry(struct json_stream *ret,
 	if (pm->b11)
 		json_add_string(ret, "bolt11", pm->b11);
 
+	if (pm->destination)
+		json_add_node_id(ret, "destination", pm->destination);
+
 	json_add_sha256(ret, "payment_hash", pm->payment_hash);
 	json_add_string(ret, "status", pm->status);
 	json_add_u32(ret, "created_at", pm->timestamp);
@@ -1785,13 +1791,15 @@ static struct command_result *listsendpays_done(struct command *cmd,
 	ret = jsonrpc_stream_success(cmd);
 	json_array_start(ret, "pays");
 	json_for_each_arr(i, t, arr) {
-		const jsmntok_t *status, *b11tok, *hashtok, *createdtok;
+		const jsmntok_t *status, *b11tok, *hashtok, *destinationtok, *createdtok;
 		const char *b11 = b11str;
 		struct sha256 payment_hash;
+		struct node_id destination;
 		u32 created_at;
 
 		b11tok = json_get_member(buf, t, "bolt11");
 		hashtok = json_get_member(buf, t, "payment_hash");
+		destinationtok = json_get_member(buf, t, "destination");
 		createdtok = json_get_member(buf, t, "created_at");
 		assert(hashtok != NULL);
 		assert(createdtok != NULL);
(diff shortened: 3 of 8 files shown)


doc/lightning-listpays.7.md
```diff
@@ -4,47 +4,51 @@ lightning-listpays -- Command for querying payment status
...
+- *label*:
 optional *label*, if provided to *pay*.
 
- *amount\_sent\_msat*
+- *amount\_sent\_msat*:
 total amount sent, in "NNNmsat" format.
 
-For old payments (pre-0.7) we didn’t save the *bolt11* string, so in its
-place are three other fields:
+~For old payments (pre-0.7) we didn’t save the *bolt11* string, so in its
```
cdecker, line 38: Please don't use strikethrough in the documentation, it should always reflect the latest state, and not carry old / outdated information at all. Also I'm pretty sure we should keep these warnings unchanged for now, until we really remove them.

plugins/pay.c
```diff
@@ -1733,7 +1733,11 @@ static void add_new_entry(struct json_stream *ret,
 			  const struct pay_mpp *pm)
 {
 	json_object_start(ret, NULL);
-	json_add_string(ret, "bolt11", pm->b11);
+	if(pm->b11){
+		json_add_string(ret, "bolt11", pm->b11);
+	}else{
```
cdecker, line 1738: Nit: whitespace arount the parentheses: 

```suggestion
	if (pm->b11) {
		json_add_string(ret, "bolt11", pm->b11);
	} else {
```

plugins/pay.c
```diff
@@ -1733,7 +1733,11 @@ static void add_new_entry(struct json_stream *ret,
...
+		json_add_sha256(ret, "payment_hash", pm->payment_hash);
```
cdecker, line 1739: I think we should be printing the `payment_hash` irregardless of whether we have a `bolt11` string or not, they are about the only way to recognize a payment.

tests/test_pay.py
```diff
@@ -3218,3 +3218,27 @@ def test_bolt11_null_after_pay(node_factory, bitcoind):
     pays = l2.rpc.listpays()["pays"]
     assert(pays[0]["bolt11"] == invl1)
     assert('amount_msat' in pays[0] and pays[0]['amount_msat'] == amt)
+
+
+def test_listpay_resutl_with_paymod(node_factory, bitcoind):
```
cdecker, line 3223:
```suggestion
def test_listpay_result_with_paymod(node_factory, bitcoind):
```

doc/lightning-listpays.7.md
vincenzopalazzo: I will apport this change inside the next commit. Thanks for this review

plugins/pay.c
vincenzopalazzo: I agree if we have the payment hash as result `listpays` is more simple to identify the payment inside the client app.

I will apport this change inside the next commit, and I remove the parenthesis to resolve also this review. https://github.com/ElementsProject/lightning/pull/3888#discussion_r462258558

tests/test_pay.py
vincenzopalazzo: Mh! sorry about that, I apologize to do typing error. This event bring the people to thinks that, I don't pay attention to when I typing (I would think so). But I-m dyslexic and for me is very difficult to find this error.

Thanks for this fix and I will try to pay attention in the future.

vincenzopalazzo:
Thanks @cdecker to review this PR and to discover my mistakes. 

I'm running the local test and after it I will push the new commit.  Only one thing that I need to clarify, I should insert a different Changelog? for instance Changelog-listpays?  Maybe is possible to add a couple of lines inside the Haking.md to explain these rules at the people that want to start and don't know it?

cdecker:
Oh right, I meant to comment on that as well: since you add a new optional parameter to the RPC call I'd say it needs to be a `Changelog-Added: JSON-RPC: ...` :-)

vincenzopalazzo:
For the moment all done @cdecker.

`Changelog-Added: JSON-RPC: listpays now accept a second optional parameter such as payment_hash`

ZmnSCPxj:
`destination` is probably a lot more relevant when reporting a `keysend` payment.

vincenzopalazzo:
@ZmnSCPxj, this is a good point, yep I agree with this update! I will try to improve the format of this feature.

ZmnSCPxj:
This probably means we need to add yet another argument to `sendonion`, an optional `destination`, which we record in the database, and which we modify `libplugin-pay.c` to also provide.

vincenzopalazzo:
@ZmnSCPxj I wrote (https://github.com/ElementsProject/lightning/pull/3888/commits/74d2eeed990cce085e3c30f6e5e15b829c0ea3f1) some code to implement your suggestion and return the destination when the bolt11 is null.

At this point output is like the json below:

```json
{
   "pays": [
      {
         "bolt11": "lntb10n1p0jga20pp53h7k2w8wkvuprjg3ff6l0y4pgdeg6lc9vsln3s74wnfsjl5fzrqqdqdw3jhxazldahx2xqyjw5qcqp2sp5wut5jnhr6n7jd5747ky2g5flmw7hgx9yjnqzu60ps2jf6f7tc0us9qy9qsqu2a0k37nckl62005p69xavlkydkvhnypk4dphffy4x09zltwh9437ad7xkl83tefdarzhu5t30ju5s56wlrg97qkx404pq3srfc425cq3ke9af",
         "payment_hash": "8dfd6538eeb33811c9114a75f792a143728d7f05643f38c3d574d3097e8910c0",
         "status": "complete",
         "preimage": "35bd4e2b481a1a84a22215b5372672cf81460a671816960ddb206464359e1822",
         "amount_msat": "1000msat",
         "amount_sent_msat": "1000msat"
      },
      {
         "destination": "0219f8900ee78a89f050c24d8b69492954f9fdbabed753710845eb75d3a75a5880",
         "payment_hash": "9c87f80bacf06a9ca08a9dcef90cf7f8b68f7b920df6020f3a0a04f35ab6c204",
         "status": "complete",
         "preimage": "3078e0c8574f756a27e7998138fe624c8d806f7075df95caee7e97efb785d108",
         "amount_msat": "1000msat",
         "amount_sent_msat": "1000msat"
      }
   ]
}
```

Did I understand your suggestion?

In addition, I passed a couple of hours to discover a timeout exception inside the test travis https://travis-ci.com/github/vincenzopalazzo/lightning/jobs/367116591

Any suggestion about how I can discover this bug inside the code that I wrote?

P.S: When @cdecker and @ZmnSCPxj have time, I'm happy to receive a review on the new code :smile: .

lightningd/pay.c
```diff
@@ -1190,6 +1191,7 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt("shared_secrets", param_secrets_array, &path_secrets),
 		   p_opt_def("partid", param_u64, &partid, 0),
 		   p_opt("bolt11", param_string, &b11str),
+		   p_opt("destination", param_node_id, &destination),
```
ZmnSCPxj, line 1194: Need documentation in `doc/lightning-sendonoin.7.md` as well,

plugins/pay.c
```diff
@@ -1733,7 +1735,11 @@ static void add_new_entry(struct json_stream *ret,
 			  const struct pay_mpp *pm)
 {
 	json_object_start(ret, NULL);
-	json_add_string(ret, "bolt11", pm->b11);
+	if (pm->b11)
+		json_add_string(ret, "bolt11", pm->b11);
+	else if (pm->destination)
+		json_add_node_id(ret, "destination", pm->destination);
```
ZmnSCPxj, line 1741: Need documentation in `lightning-listpays.7.md` as well.

vincenzopalazzo:
@ZmnSCPxj Yes is very strange, now all tests are ok, on my branch for 3 consecutive times they were red. any why, all good for the moment.

cdecker:
Rebased on top of `master`

lightningd/pay.c
ZmnSCPxj: Also, you should not insert arguments, only append them.  This will conflict if the user uses an array as parameter instead of a JSON object.

vincenzopalazzo: @ZmnSCPxj, mh! can clarify this case? I'm missing something here

cdecker:
Fixed up and squashed.

lightningd/pay.c
ZmnSCPxj: You should never insert a new parameter between existing ones (the exception is for dev-only parameters gated by `DEVELOPER`, which should always be at the end of the parameter list, and new non-dev parameters are inserted before that).  You can only append a new parameter.  So you should have put `destination` *after* `msatoshi`, not before `msatoshi`.  Latest code does this already.

vincenzopalazzo: Oh! Thanks to clarify it. The next time I will pay more focus about it.

Issue #3858, opened 2020-07-20
Planning for SPV backend


ZmnSCPxj:
First off: SPV is worse security than fullnodes.  Please use fullnodes if you can.

Currently, the interface to `bcli` inherently assumes the backend is a fullnode of some kind.  This fullnode is always fully trusted to not lie, and we always download every block from the backend.

When doing block synchronization, what we can do is to have a `txfilter` containing:

* UTXOs which, if spent, we want to know about.
* Addresses which, if spent from or received into, we want to know about.

Now, it is important that we keep track of UTXOs that we want to track spendedness of channels.  However, each such UTXO has its own address as well.

Most SPV protocols (eg BIP158, Electrum) are address-centric.  So, we should provide just addresses to the backend.  The backend should inform us

So I propose for the `bcli` `getrawblockbyheight`:

* Add a new required parameter, `txfilter`, which is an array of `scriptPubKey` hexdumps.
* It may return a result without a `block` field (but must still return `blockhash`).  This is used to inform `lightningd` that the block at the given height does not involve any of the addresses in the `txfilter`.
* It has to return a separate `blockheader` field regardless.  This is needed so that chaintopology can recognize reorgs.

Then:

* For Electrum, we do `blockchain.scripthash.get_history` on every listed `scriptPubKey`, and if any of them has a transaction whose confirmed `height` equals the height being queried, then we should download that block (elsewhere) and provide it to the `lightningd`.
  * But we should be wary of reorgs.  Thus, we should do a `blockchain.block.header` and save the header, then `blockchain.scripthash.get_history` for a *single* scriptpubkey, then a `blockchain.block.header` and compare if it is the same as previous; if not the same, we should restart the querying again, since the alternate block might now contain the previous `scriptpubkey`.  This is not resilient against very fast ABA problems but hopefully those are rare.
* For BIP158, the plugin maintains a header chain, and we download the version `0x00` Neutrino filter for the block header hash at the requested height. Then we check if anything in our `txfilter` matches the Neutrino filter.

------

Ideally, our fullnode-trusting `bcli` should **also** apply the `txfilter` to blocks it returns.  This is to ensure that any bugs in maintaining the `lightningd`-side `txfilter` can be caught by our test suite without having to add SPV plugins of our own.

Unfortunately, to determine if an address is *spent from*, we would require a `txindex`, or have the plugin maintain its own transaction index.  Both are obviously undesirable.

However, in practice:

* We only care about addresses we *receive from*.
[...]

PR #3889, opened 2020-07-29
Explictly check at startup that we don't support `bitcoind`'s `blocksonly` option.

darosior:
This adds a sanity check function to `bcli` which we could extend with other sanity checks in the future.

This was [brought up](https://github.com/ElementsProject/lightning/issues/3858#issuecomment-665619664) by @ZmnSCPxj.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Don't hesitate to reach out to us on IRC at [#lightning-dev @ freenode.net][irc1
 
 ## Getting Started
 
-c-lightning only works on Linux and Mac OS, and requires a locally (or remotely) running `bitcoind` (version 0.16 or above) that is fully caught up with the network you're testing on.
+c-lightning only works on Linux and Mac OS, and requires a locally (or remotely) running `bitcoind` (version 0.16 or above) that is fully caught up with the network you're running on, and relays transactions (ie with `blocksonly=0`).
 Pruning (`prune=n` option in `bitcoin.conf`) is partially supported, see [here](#pruning) for more details.
 
 ### Installation
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -1026,8 +1026,8 @@ class NodeFactory(object):
 
     def get_node(self, node_id=None, options=None, dbfile=None,
                  feerates=(15000, 11000, 7500, 3750), start=True,
-                 wait_for_bitcoind_sync=True, expect_fail=False,
-                 cleandir=True, **kwargs):
+                 wait_for_bitcoind_sync=True, may_fail=False,
+                 expect_fail=False, cleandir=True, **kwargs):
 
         node_id = self.get_node_id() if not node_id else node_id
         port = self.get_next_port()
@@ -1043,7 +1043,8 @@ class NodeFactory(object):
         db = self.db_provider.get_db(os.path.join(lightning_dir, TEST_NETWORK), self.testname, node_id)
         node = self.node_cls(
             node_id, lightning_dir, self.bitcoind, self.executor, db=db,
-            port=port, options=options, **kwargs
+            port=port, options=options, may_fail=may_fail or expect_fail,
+            **kwargs
         )
 
         # Regtest estimatefee are unusable, so override.
--- a/plugins/bcli.c
+++ b/plugins/bcli.c
@@ -40,6 +40,9 @@ struct bitcoind {
 	/* -datadir arg for bitcoin-cli. */
 	char *datadir;
 
+	/* bitcoind's version, used for compatibility checks. */
+	u32 version;
+
 	/* Is bitcoind synced?  If not, we retry. */
 	bool synced;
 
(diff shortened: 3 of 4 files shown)


ZmnSCPxj:
bfc73eb4f52c61aa22934039f75a777c4c32ee76 in #3870 also does a `getnetworkinfo`, but rather than a separate call, it replaces the `echo` in `wait_for_bitcoind`, what do you think?  Since we need to call `getnetworkinfo` anyway I think replacing the `echo` would be better (though this is arguably a tiny optimization).

ZmnSCPxj:
Also we should probably update `README.md` as well, nobody reads anything in `doc/` anyway.

darosior:
I updated the README in the first commit.

Regarding replacing `echo`, I agree it's cleaner after having read your commit. Will amend!

darosior:
Adapted part of https://github.com/ElementsProject/lightning/commit/bfc73eb4f52c61aa22934039f75a777c4c32ee76 in https://github.com/ElementsProject/lightning/pull/3889/commits/b934a927457dbce46b9ef93dbb0c00f2996b60f9 and built upon it for transaction relay sanity checks. Added a trivial check for `bitcoind` minimal supported version too. Also, this is a bit more restrictive than your initial version, as we fail to start if the call fails. I think it makes sense to not go further if a simple `getnetworkinfo` does not succeed.

README.md
```diff
@@ -31,7 +31,7 @@ Don't hesitate to reach out to us on IRC at [#lightning-dev @ freenode.net][irc1
 
 ## Getting Started
 
-c-lightning only works on Linux and Mac OS, and requires a locally (or remotely) running `bitcoind` (version 0.16 or above) that is fully caught up with the network you're testing on.
+c-lightning only works on Linux and Mac OS, and requires a locally (or remotely) running `bitcoind` (version 0.16 or above) that is fully caught up with the network you're runing on, and relays transactions (ie with `blocksonly=0`).
```
ZmnSCPxj, line 34: s/runing/running/

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -1043,7 +1043,8 @@ def get_node(self, node_id=None, options=None, dbfile=None,
         db = self.db_provider.get_db(os.path.join(lightning_dir, TEST_NETWORK), self.testname, node_id)
         node = self.node_cls(
             node_id, lightning_dir, self.bitcoind, self.executor, db=db,
-            port=port, options=options, **kwargs
+            port=port, options=options, may_fail=may_fail or expect_fail,
```
ZmnSCPxj, line 1046: What we usually do in existing tests is `may_fail=True, expect_fail=True` but I agree this is better and makes sense.

darosior: Yeah, and `LightningNode()` does not have an `expect_fail` parameter either.

darosior:
Corrected the grammar error.

plugins/bcli.c
```diff
@@ -834,14 +837,62 @@ static void bitcoind_failure(struct plugin *p, const char *error_message)
...
+	 * estimates. */
+	relaytok = json_get_member(buf, result, "localrelay");
+	if (!relaytok || !json_to_bool(buf, relaytok, &tx_relay))
+		plugin_err(p, "No 'localrelay' in '%s' ? Got '%s'. Can not"
+			      " continue without proceeding to sanity checks.",
+			      gather_args(bitcoind, "getnetworkinfo", NULL), buf);
+
+	if (!tx_relay)
+		plugin_err(p, "The 'blocksonly' mode of bitcoind, or any option "
+			      "deactivating transaction relay is not supported.");
```
niftynei, line 880: Maybe offer a suggestion of how to fix it? i.e. 

```suggestion
			      "deactivating transaction relay is not supported. Remove -blocksonly from your bitcoind config");
```

plugins/bcli.c
```diff
@@ -834,14 +837,62 @@ static void bitcoind_failure(struct plugin *p, const char *error_message)
...
+static void parse_getnetworkinfo_result(struct plugin *p, const char *buf)
+{
+	const jsmntok_t *result, *versiontok, *relaytok;
+	bool valid, tx_relay;
+
+	result = json_parse_input(NULL,
+				  buf, strlen(buf),
+				  &valid);
+	if (!result || !valid)
+		plugin_err(p, "No or invalid response to '%s' ? Got '%s'. Can not"
```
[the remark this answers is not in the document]
darosior: Done

plugins/bcli.c
```diff
@@ -834,14 +837,62 @@ static void bitcoind_failure(struct plugin *p, const char *error_message)
...
+			      " continue without proceeding to sanity checks.",
+			      gather_args(bitcoind, "getnetworkinfo", NULL), buf);
+
+	if (!json_to_u32(buf, versiontok, &bitcoind->version))
+		plugin_err(p, "Invalid 'version' in '%s' ? Got '%s'. Can not"
+			      " continue without proceeding to sanity checks.",
+			      gather_args(bitcoind, "getnetworkinfo", NULL), buf);
+
+	if (bitcoind->version < 160000)
+		plugin_err(p, "Unsupported bitcoind version, you need to update"
```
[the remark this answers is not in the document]
darosior: Done

plugins/bcli.c
darosior: Seems a bit redundant ?..
> The 'blocksonly' mode of bitcoind [...] is not supported. Remove -blocksonly from your bitcoind config

ZmnSCPxj: Agree as well.

PR #3890, opened 2020-07-30
Important plugin

### LINKED ISSUE #3213: "Crucial" plugins
~~For now, if a plugin dies before `lightningd` for whatever reason while having registered a hook, `lightningd` will hang forever when sending  a "hook request" to the plugin. We might want to rely more on plugin for more crucial operations (for example, RPC wallet locking, Bitcoin network interface), hence a special kind of plugins -"crucial" plugins- might be defined so that if they die, `lightningd` does too.~~
`lightningd` isnt stuck anymore as of https://github.com/ElementsProject/lightning/issues/3496, and https://github.com/ElementsProject/lightning/issues/3213#issuecomment-568973022 is overall a better proposal.

ZmnSCPxj:
Closes #3213

Builtin plugins are automatically made important.

### DIFF
--- a/lightningd/lightningd.h
+++ b/lightningd/lightningd.h
@@ -271,6 +271,9 @@ struct lightningd {
 	/* Total number of coin moves we've seen, since
 	 * coin move tracking was cool */
 	s64 coin_moves_count;
+
+	/* If non-NULL, contains the exit code to use.  */
+	int *exit_code;
 };
 
 /* Turning this on allows a tal allocation to return NULL, rather than aborting.
@@ -288,5 +291,12 @@ void test_subdaemons(const struct lightningd *ld);
 
 /* Notify lightningd about new blocks. */
 void notify_new_block(struct lightningd *ld, u32 block_height);
+
+/* Signal a clean exit from lightningd.
+ * NOTE! This function **returns**.
+ * This just causes the main loop to exit, so you have to return
+ * all the way to the main loop for `lightningd` to exit.
+ */
+void lightningd_exit(struct lightningd *ld, int exit_code);
 
 #endif /* LIGHTNING_LIGHTNINGD_LIGHTNINGD_H */
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -348,7 +348,7 @@ static char *opt_add_plugin(const char *arg, struct lightningd *ld)
 		log_info(ld->log, "%s: disabled via disable-plugin", arg);
 		return NULL;
 	}
-	plugin_register(ld->plugins, arg, NULL);
+	plugin_register(ld->plugins, arg, NULL, false);
 	return NULL;
 }
 
@@ -366,6 +366,16 @@ static char *opt_add_plugin_dir(const char *arg, struct lightningd *ld)
 static char *opt_clear_plugins(struct lightningd *ld)
 {
 	clear_plugins(ld->plugins);
+	return NULL;
+}
+
+static char *opt_important_plugin(const char *arg, struct lightningd *ld)
+{
+	if (plugin_blacklisted(ld->plugins, arg)) {
+		log_info(ld->log, "%s: disabled via disable-plugin", arg);
+		return NULL;
+	}
+	plugin_register(ld->plugins, arg, NULL, true);
 	return NULL;
 }
 
@@ -770,6 +780,10 @@ static void register_opts(struct lightningd *ld)
 	opt_register_early_arg("--disable-plugin", opt_disable_plugin,
 			       NULL, ld,
 			       "Disable a particular plugin by filename/name");
+
+	opt_register_early_arg("--important-plugin", opt_important_plugin,
+			       NULL, ld,
+			       "Add an important plugin to be run (can be used multiple times). Die if the plugin dies.");
 
 	/* Early, as it suppresses DNS lookups from cmdline too. */
 	opt_register_early_arg("--always-use-proxy",
@@ -1278,6 +1292,9 @@ static void add_config(struct lightningd *ld,
 			json_add_opt_log_levels(response, ld->log);
 		} else if (opt->cb_arg == (void *)opt_disable_plugin) {
 			json_add_opt_disable_plugins(response, ld->plugins);
+		} else if (opt->cb_arg == (void *)opt_important_plugin) {
+			/* Do nothing, this is already handled by
+			 * opt_add_plugin.  */
 		} else if (opt->cb_arg == (void *)opt_add_plugin_dir
 			   || opt->cb_arg == (void *)plugin_opt_set
 			   || opt->cb_arg == (void *)plugin_opt_flag_set) {
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -566,6 +566,10 @@ static void dev_register_opts(struct lightningd *ld)
 	opt_register_noarg("--dev-no-version-checks", opt_set_bool,
 			   &ld->dev_no_version_checks,
 			   "Skip calling subdaemons with --version on startup");
+	opt_register_early_noarg("--dev-builtin-plugins-unimportant",
+				 opt_set_bool,
+				 &ld->plugins->dev_builtin_plugins_unimportant,
+				 "Make builtin plugins unimportant so you can plugin stop them.");
 }
 #endif /* DEVELOPER */
 
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1526,6 +1526,16 @@ struct log *plugin_get_log(struct plugin *plugin)
 	return plugin->log;
 }
 
+void plugins_set_builtin_plugins_dir(struct plugins *plugins,
+				     const char *dir)
+{
+	/*~ The builtin-plugins dir does not need to exist, but
+	 * we would error those anyway for our important built-in
+	 * plugins.
+	 */
+	add_plugin_dir(plugins, dir, true);
+}
+
 struct plugin_destroyed {
 	const struct plugin *plugin;
 };
@@ -1535,6 +1545,7 @@ static void mark_plugin_destroyed(const struct plugin *unused,
 {
 	pd->plugin = NULL;
 }
+
 
 struct plugin_destroyed *plugin_detect_destruction(const struct plugin *plugin)
 {
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -16,6 +16,9 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+/* Only this file can include this generated header! */
+# include <gen_list_of_builtin_plugins.h>
+
 /* How many seconds may the plugin take to reply to the `getmanifest`
  * call? This is the maximum delay to `lightningd --help` and until
  * we can start the main `io_loop` to communicate with peers. If this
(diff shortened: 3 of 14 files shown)


ZmnSCPxj:
Fix documentation (we need to say `--unimportant-plugin` before `--disable-plugin`) and also make sure test for `bcli` replacement uses `--unimportant-plugin`.

lightningd/lightningd.h
```diff
@@ -289,4 +292,11 @@ void test_subdaemons(const struct lightningd *ld);
 /* Notify lightningd about new blocks. */
 void notify_new_block(struct lightningd *ld, u32 block_height);
 
+/* Signal a clean exit from lightningd.
+ * NOTE! This function **returns**.
+ * This just causes the main loop to exit, so you have to return
+ * all the way to the main loop.
```
niftynei, line 298:
```suggestion
 * all the way to the main loop for `lightningd` to exit
```

This is a bit tautological, as written.

lightningd/options.c
```diff
@@ -771,6 +786,15 @@ static void register_opts(struct lightningd *ld)
...
 			       "Disable a particular plugin by filename/name");
 
+	opt_register_early_arg("--important-plugin", opt_important_plugin,
+			       NULL, ld,
+			       "Refuse to run if the indicated plugin dies or "
+			       "is not found.");
+	opt_register_early_arg("--unimportant-plugin", opt_unimportant_plugin,
+			       NULL, ld,
+			       "Cancel a previous important-plugin; "
+			       "continue runing even if it dies or is not found.");
```
niftynei, line 796:
```suggestion
			       "continue running even if it dies or is not found.");
```

lightningd/plugin.c
```diff
@@ -16,6 +16,9 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 
+/* Only this file can include this generated header! */
+# include <list_of_builtin_plugins.h>
```
niftynei, line 20: Generated files should be prefixed with `gen-`?

lightningd/plugin.c
```diff
@@ -203,6 +236,32 @@ bool plugin_blacklisted(struct plugins *plugins, const char *name)
...
 
+void plugin_important(struct plugins *plugins, const char *name)
+{
+	tal_arr_expand(&plugins->important,
+		       tal_strdup(plugins->important, name));
+}
+
+bool plugin_unimportant(struct plugins *plugins, const char *name)
+{
+	size_t l = tal_count(plugins->important);
```
niftynei, line 247: it'd be easier to read if `l` had a more descriptive name (`l` and `i` are also visually similar)

lightningd/plugin.c
```diff
@@ -1300,6 +1359,31 @@ void plugins_init(struct plugins *plugins)
...
+				break;
+			}
+		}
+
+		if (!found) {
+			log_broken(plugins->log,
+				   "An --important-plugin=%s was not found, "
+				   "exiting lightningd.",
+				   name);
+			fatal("An --important-plugin=%s was not found, "
```
niftynei, line 1381: why are we repeating `fatal` and `log_broken`?

lightningd/plugin.c
```diff
@@ -1300,6 +1359,31 @@ void plugins_init(struct plugins *plugins)
...
+		list_for_each(&plugins->plugins, p, list) {
+			if (plugin_paths_match(p->cmd, name)) {
+				found = true;
+				break;
+			}
+		}
+
+		if (!found) {
+			log_broken(plugins->log,
+				   "An --important-plugin=%s was not found, "
```
niftynei, line 1378:
```suggestion
				   "--important-plugin=%s was not found, "
```

Since we're giving the plugin name, no need to articalize it.

lightningd/plugin.c
```diff
@@ -1536,6 +1643,7 @@ static void mark_plugin_destroyed(const struct plugin *unused,
 	pd->plugin = NULL;
 }
 
+
```
niftynei, line 1646: erraneous addition?

lightningd/plugin.h
```diff
@@ -103,6 +103,12 @@ struct plugins {
 
 	/* Blacklist of plugins from --disable-plugin */
 	const char **blacklist;
+
+	/* Array of plugins from --important-plugin */
```
niftynei, line 107:
```suggestion
	/* List of plugins marked --important-plugin */
```

tests/test_plugin.py
```diff
@@ -1104,7 +1106,8 @@ def test_bitcoin_backend(node_factory, bitcoind):
     """
     This tests interaction with the Bitcoin backend, but not specifically bcli
     """
-    l1 = node_factory.get_node(start=False, options={"disable-plugin": "bcli"},
+    l1 = node_factory.get_node(start=False, options=OrderedDict([("unimportant-plugin", "bcli"),
+                                                                 ("disable-plugin", "bcli")]),
```
niftynei, line 1110: it's weird that you have to specify `unimportant` and `disable`. if you're explicitly `disabling` a plugin, we should automatically mark it as unimportant, as the desire to disable it obviously means "I want this thing off".

tests/test_plugin.py
```diff
@@ -1534,3 +1537,83 @@ def test_3847_repro(node_factory, bitcoind):
...
+
+    # Check we fail if we cannot find the important plugin.
+    n = node_factory.get_node(options={"important-plugin": "nonexistent"},
+                              may_fail=True, expect_fail=True,
+                              allow_broken_log=True)
+    assert not n.daemon.running
+    assert n.daemon.is_in_stderr('An --important-plugin=nonexistent was not found')
+
+    # Check we exit if the important plugin dies.
+    n = node_factory.get_node(options={"important-plugin": "fail_by_itself.py",
```
niftynei, line 1554: it seems like you should be able to specify the plugin as either `important-plugin` or `plugin`, so here you'd only need to do

```
"important-plugin": "os.path.join(pluginsdir, "fail_by_iteself.py")
```

This does break the 'important-' / 'unimportant-' symmetry though hmm.

tests/test_plugin.py
```diff
@@ -1534,3 +1537,83 @@ def test_3847_repro(node_factory, bitcoind):
...
+                              may_fail=True, expect_fail=True,
+                              allow_broken_log=True)
+
+    n.daemon.wait_for_log('Plugin marked as important by --important-plugin=fail_by_itself.py')
+    wait_for(lambda: not n.daemon.running)
+
+    # Check if the important plugin is made unimportant,
+    # we go back to ignoring it.
+    n = node_factory.get_node(options=OrderedDict([("important-plugin", "fail_by_itself.py"),
+                                                   ("unimportant-plugin", "fail_by_itself.py"),
```
niftynei, line 1565: is `unimportant-plugin` necessary? it seems like `disable` might work better. i.e. are there cases where you'd want to run an 'important' plugin but not having it be fatal? if not, it seems like you could do away with the `unimportant` option and use `disable` as the override.

tests/test_plugin.py
```diff
@@ -1534,3 +1537,83 @@ def test_3847_repro(node_factory, bitcoind):
...
+    # And the PIDs of plugins are only seen at plugin startup.
+    TailableProc.start(n.daemon)
+    assert n.daemon.running
+    # Extract the pid of pay.
+    r = n.daemon.wait_for_log(r'started([0-9]*).*plugins/pay')
+    pidstr = re.search(r'.*started\(([0-9]*)\)', r).group(1)
+    # Kill pay.
+    os.kill(int(pidstr), signal.SIGKILL)
+    # node should die as well.
+    n.daemon.wait_for_log('Plugin marked as important by --important-plugin=pay')
```
niftynei, line 1602: Should this say `Plugin marked as important by --important-plugin=pay died`?

tests/test_plugin.py
ZmnSCPxj: You could be developing `pay`, for example, and want to recompile `plugins/pay` without having to restart `lightningd`.  You do this by marking it `--unimportant-plugin` and then using the `lightning-cli plugin restart pay` command.

Might be too edge case however...

ZmnSCPxj:
Updated as per @niftynei feedback.

rustyrussell:
Looks good, but Travis is deeply unhappy...

tests/test_plugin.py
niftynei: one alternative would be to have `--plugin` do the same work as as `--unimportant-plugin`. in other words, to mark an important plugin as unimportant you'd just need to specify it as `--plugin`? that seems straightforward enough.

ZmnSCPxj: Certainly an idea.  However, how do we implement `add-plugin-dir`? Currently this is just equivalent to iterating `--plugin` over all executables in the dir,   But one thing a user might want to do would be to `add-plugin-dir` a directory containing all plugins they have, then later, if the decide that a plugin in that dir is important, use `important-plugin` on the "blessed" plugin.  In the current version of the PR the `important-plugin` upgrades the plugin to important, and a later `--plugin` or `--add-plugin-dir` will not downgrade it to unimportant.  Hmmm...........

ZmnSCPxj: Alternatively, we should note that only the built-in plugins get default-important.  And the only ones interested in making built-in plugins unimportant are we the developers of `lightningd` (and not, say, developers building on top of `lightningd`).  So maybe better would be a `--dev-builtin-plugins-unimportant` for `DEVELOPER=1` builds?

ZmnSCPxj:
Fixed the valgrind bug that was making travis unhappy (why did travis keep rerunning the test tho???) and added `--dev-builtin-plugins-unimportant` option.

lightningd/plugin.c
```diff
@@ -130,6 +154,9 @@ struct plugin *plugin_register(struct plugins *plugins, const char* path TAKES,
 		if (streq(path, p_temp->cmd)) {
 			if (taken(path))
 				tal_free(path);
+		 	/* If added as "important", upgrade to "important".  */
+			if (important)
+				p_temp->important = true;
```
darosior, line 159: @ZmnSCPxj why ? I don't think we want to do this. *(sorry for the post-merge review)*

ZmnSCPxj: A user might have a directory-full of plugins added by `add-plugin-dir`.  Then they decide later that one of those plugins is so vital that `lightningd` must die rather than move on if the plugin dies, so marks it as `important-plugin`.  It would be difficult otherwise to promote a plugin to important.

darosior: Yeah but this loop is to check the path of the plugin we register against already registered one:
```C
/* Don't register an already registered plugin */
	list_for_each(&plugins->plugins, p_temp, list) {
		if (streq(path, p_temp->cmd)) {
			if (taken(path))
				tal_free(path);
		 	/* If added as "important", upgrade to "important".  */
			if (important)
				p_temp->important = true;
                            return NULL;
		}
	}
```
You should not make the cursor used to loop `important` if the plugin you want to register is important.

ZmnSCPxj: It is correct.  Consider the case where `lightningd.conf`:

```
plugin-dir=/path/to/plugins/
important-plugin=/path/to/plugins/really-important-plugin.sh
```

`--plugin-dir` calls into `add_plugin_dir`, which iterates over all executable files in the dir, and calls `plugin_register` for each file.  This causes `plugin_register` to construct a `struct plugin` for each file in the dir, and if the dir contains `really-important-plugin.sh`, then a `struct plugin` will be added to the `plugins->plugins`, but (since it has not processed `important-plugin` yet!) marked as unimportant.

Then the *later* `important-plugin` will again call into `plugin_register`, which will enter the loop, and find the existing `/path/to/plugins/really-important-plugin.sh` `struct plugin`.  So it now does the upgrade of this already-existing plugin to `important`.

Otherwise we would require that all important plugins be listed before all `plugin-dir` indications, and since we have a default `plugin-dir`, we cannot store important plugins there.

In the case where the `important-plugin` is not already existing in the list of plugins, we initialize `p->important = important` in the later allocation code of the function.

darosior: What i mean is the above code leads to *all currently registered plugins being marked as important* if an important plugin is started twice and happens to be at the end of the list. Will open a PR to discuss it, but maybe i'm wrong and i'll remark it by testing it ..

ZmnSCPxj: The `if (streq(path, p_temp->cmd))` and `return NULL` both prevent this from running on all registered plugins.

darosior: Ok, got it, sorry for insisting!

PR #3892, opened 2020-07-31
Route randomization

ZmnSCPxj:
Currently working on reimplementing route randomization, which we currently do not have in any significant due to nonimplementation of large fuzzing in legacy pay and continued non-implementation of any kind of fuzzing in paymod.

The given test shows that route randomization is so nonexistent that I get 100% all payments in the path l1->l2->l3 in the test.  That is just... not doing route randomization.  Replacing `l1.rpc.pay(invoice)` with `l1.rpc.call("legacypay", [invoice])` has the same result, 100% of all payments go through l1->l2->l3.

Not really sure best way to implement route randomization for now.

An idea is to introduce a concept of "route suffix", which is basically a single pre-selected route starting at an entry point, then going to the destination.  At the start of a payment attempt the route suffix is cleared.  Then the `routehint` module adds the current routehint to the route suffix.  Then a new module, `randomizer`, adds a 1 to 3 more hops at random to the route suffix, mutating the entry point.  Is it possible to enforce a particular order to when modules trigger @cdecker?

The `randomizer` module maintains a single `listchannels_cache` for the entire payment, which, as its name suggests, caches calls to `listchannels` for a payment.  The `randomizer` module, on the `INITIALIZED` step callback (which has to execute after `routehints`), then looks up channels connected to the entry point of the suffix.  It filters out channels with insufficient capacity for the payment, and filters out channels/nodes that are in the exclusion set, then randomly selects one of the channels and extends the route suffix.  It then rolls a dice to determine if it goes for another turn.  The `listchannels_cache` ensures that even if multiple sub-payments have the same routehint, or a payment repeats due to some failure, then we do not spam `lightningd` with `listchannels` requests.

What do you think @cdecker , is this doable?

This operation is similar to ideas that @rustyrussell has considered for route randomization before.  For example, we can emulate the Rusty Russell Route Randomization Ralgorithm by creating a short random-walk route from the destination, then doing a "normal" A\* from the source to the point where the random walk stops.

If I get too annoyed at the paymod I might just implement a `getrandomroute` plugin that has the same interface as `getroute` but does the random walk from the destination, implementing a form of the Rusty Russell Route Randomization Ralgorithm.

### DIFF
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -6,7 +6,7 @@ from pyln.client import RpcError, Millisatoshi
 from pyln.proto.onion import TlvPayload
 from utils import (
     DEVELOPER, wait_for, only_one, sync_blockheight, SLOW_MACHINE, TIMEOUT,
-    VALGRIND, EXPERIMENTAL_FEATURES
+    VALGRIND, EXPERIMENTAL_FEATURES, get_tx_p2wsh_outnum
 )
 import copy
 import os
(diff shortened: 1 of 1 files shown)


rustyrussell:
I've got a standalone random routing implementation, which I'm currently evaluating. It seems to give a nice amount of unpredictability, but of course fees are higher.

It builds on the gossip store mapping work from earlier, so route calculations can easily be boutique. It will work well with @cdecker's plans to persist information about route failures, too.

ZmnSCPxj:
Okay, will wait for that PR instead of implementing route randomization myself.

A thing that always comes up in route randomization is "how much is the fee budget I can use to gain more privacy"?  In current `getroute` implementations, this was represented by the `fuzzfactor`.  I think a similar mechanism, or an explicit passing-in of `fee_budget` and `cltv_budget`, is necessary to make `pay` dynamically provide privacy based on network fees and your `maxfeepercent` setting.

ZmnSCPxj:
We should probably remove the text in our manpages re: route randomization though, it hasn't worked for `pay` since 0.7.x or something.

ZmnSCPxj:
Another idea for route randomization:

* Start with the shortest path.
* Select a channel on the current path at random.  Replace that channel with a `permuteroute` (i.e. look for a node connected to both endnodes of that channel and splice it into the path, replacing the original channel).
  * If `permuteroute` fails, just return the current path.
  * If the resulting path exceeds our fee/cltv budget, revert the change and return the previous path.

This gives very aggressive randomization, filling up the fee budget as much as it can

I suggest having a separate RPC method for getting a randomized path, and providing `maxfee` and `maxdelay` arguments to it, which are soft limits: the method will try to keep within those bounds, but if the shortest path is already above those bounds then there is nothing we can do further..

felipelalli:
uACK

cdecker:
This PR predates the rewrite of the `pay` plugin and is likely no longer applicable (though no actual code, just a test is contained in this PR).

Closing for now, feel free to reopen should anyone decide to pick it up again.

PR #3897, opened 2020-08-03
channel: Base the channel forget timeout on the headercount

cdecker:
We were basing the 2016 block timeout on the blockchain height that we had
processed at the time completed the funding, which could lag considerably
behind the network-wide blockheight. For example this could happen if we
started with `--rescan` from a low height, taking some time to catch up.

This means that we could end up forgetting channels even before reaching the
network blockheight. This patch instead uses the headercount reported by the
backend plugin if we aren't caught up yet. While the chances of this happening
are still there, the window it might happen are much reduced, since headers
can be synced in a couple of minutes.

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -932,6 +932,15 @@ u32 get_block_height(const struct chain_topology *topo)
 	return topo->tip->height;
 }
 
+u32 get_network_blockheight(const struct chain_topology *topo)
+{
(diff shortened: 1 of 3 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/3897/commits/8b65f4b5fc0ce932d652f8b022b11a4ef2150cbf

Issue #1995, opened 2018-10-07
Create the equivalent of "delinvoice" for payments


sumBTC:
As far as I know, there isn't a method for deleting failed or completed payments, like there is for invoices (like delinvoice, and delexpiredinvoice). The aim is to reduce the output of 

    lightning-cli listpayments

darosior:
I think this can be marked good first issue ?

sumBTC:
Now, I delete **all** payments via:

`sqlite3 -header -line lightningd.sqlite3 'DELETE FROM payments'`
`sqlite3 -header -line lightningd.sqlite3 'DELETE FROM invoices'`

in `.lightning/bitcoin`

vincenzopalazzo:
@cdecker Maybe this Issue can be close by this PR https://github.com/ElementsProject/lightning/pull/3899

cdecker:
Definitely, thanks for looking this us @vincenzopalazzo

sumBTC:
Tested it and it works. Thanks to all involved but in particular @vincenzopalazzo!

PR #3899, opened 2020-08-03
delpay: introduced a new rpc method to delete a payment by hash

vincenzopalazzo:
Hi all.

This Draft implement a new RPC command to delete the payment with the status `complete` `failed` marked like good first issue https://github.com/ElementsProject/lightning/issues/1995

However, this PR depends from the PR where I proposed to update the result of listpays command with `payment_hash` inside the result https://github.com/ElementsProject/lightning/pull/3888. In addition, with the @cdecker suggestion to add the `payment_hash` in all cases of types of payments, the command `delpay` can use only the `payment_hash` to identify the payment to remove and not `bolt11`.

This PR is a draft yet because I have some point to clarify with the team, for instance, the type of error that the command should be returned in cases of error. In other words, I see that the `LIGHTNINGD` error is marked with a `FIXME` because each command should use the correct error code. I'm not able to find the payment error codes inside the code. maybe we could define the payment error code?

- [x] Docs update with the change proposed in this Draft
- [X] Fix error code inside the command implementation

Changelog-Added: JSON-RPC: delpay a new method to delete the payment completed or failed.

### DIFF
--- a/doc/lightning-delpay.7.md
+++ b/doc/lightning-delpay.7.md
@@ -4,17 +4,16 @@ lightning-delpay -- Command for removing a completed or failed payment
 SYNOPSIS
 --------
 
-**delpay** *payment\_hash* \[status\]
+**delpay** *payment\_hash* *status*
 
 DESCRIPTION
 -----------
 
-The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
-status. However, the command doesn't permit to remove a pending payment.
+The **delpay** RPC command deletes a payment with the given `payment_hash` if its status is either `complete` or `failed`. Deleting a `pending` payment is an error.
 
-- *payment\_hash*: Rapresents the unique identifier of a payment. To find it, you can run **listpays** or **listsendpays**;
-- *status* is the expected status of the payment. It can be *complete* or *failed*. 
-Only delete if the payment status matches. If not specified, defaults to *complete*.
+- *payment\_hash*: The unique identifier of a payment.
+- *status*: Expected status of the payment. 
+Only deletes if the payment status matches.
 
 EXAMPLE JSON REQUEST
 ------------
@@ -32,16 +31,16 @@ EXAMPLE JSON REQUEST
 RETURN VALUE
 ------------
 
-On success, the command will return a payment object, such as the **listsendpays**. In addition, if the payment is a MPP (Multi part payment) the command return a list of 
-payments; a payment object for each partid.
+If successful the command returns a payment object, in the same format as **listsendpays**. If the payment is a multi-part payment (MPP) the command return a list of 
+payments will be return -- one payment object for each partid.
 
-On failure, an error is returned and any payment is deleted. If the lightning process fails before responding, the
+On failure, an error is returned. If the lightning process fails before responding, the
 caller should use lightning-listsentpays(7) or lightning-listpays(7) to query whether this payment was deleted or not.
 
 The following error codes may occur:
 
-- -32602: Some parameter missed or some parameter is malformed;
-- 211: Payment with payment\_hash have a wrong status. To check the correct status run the command **paystatus**;
+- -32602: Parameter missed or malformed;
+- 211: Payment status mismatch. Check the correct status via **paystatus**;
 - 208: Payment with payment\_hash not found.
 
 EXAMPLE JSON RESPONSE
--- a/doc/lightning-delpay.7.md
+++ b/doc/lightning-delpay.7.md
@@ -1,0 +1,84 @@
+lightning-delpay -- Command for removing a completed or failed payment
+============================================================
+
+SYNOPSIS
+--------
+
+**delpay** *payment\_hash* \[status\]
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. However, the command doesn't permit to remove a pending payment.
+
+- *payment\_hash*: Rapresents the unique identifier of a payment. To find it, you can run **listpays** or **listsendpays**;
+- *status* is the expected status of the payment. It can be *complete* or *failed*. 
+Only delete if the payment status matches. If not specified, defaults to *complete*.
+
+EXAMPLE JSON REQUEST
+------------
+```json
+{
+  "id": 82,
+  "method": "delpay",
+  "params": {
+    "payment_hash": "4fa2f1b001067ec06d7f95b8695b8acd9ef04c1b4d1110e3b94e1fa0687bb1e0",
+    "status": "complete"
+  }
+}
+```
+
+RETURN VALUE
+------------
+
+On success, the command will return a payment object, such as the **listsendpays**. In addition, if the payment is a MPP (Multi part payment) the command return a list of 
+payments; a payment object for each partid.
+
+On failure, an error is returned and any payment is deleted. If the lightning process fails before responding, the
+caller should use lightning-listsentpays(7) or lightning-listpays(7) to query whether this payment was deleted or not.
+
+The following error codes may occur:
+
+- -32602: Some parameter missed or some parameter is malformed;
+- 211: Payment with payment\_hash have a wrong status. To check the correct status run the command **paystatus**;
+- 208: Payment with payment\_hash not found.
+
+EXAMPLE JSON RESPONSE
+-----
+```json
+{
+   "payments": [
+      {
+         "id": 2,
+         "payment_hash": "8dfd6538eeb33811c9114a75f792a143728d7f05643f38c3d574d3097e8910c0",
+         "destination": "0219f8900ee78a89f050c24d8b69492954f9fdbabed753710845eb75d3a75a5880",
+         "msatoshi": 1000,
+         "amount_msat": "1000msat",
+         "msatoshi_sent": 1000,
+         "amount_sent_msat": "1000msat",
+         "created_at": 1596224858,
+         "status": "complete",
+         "payment_preimage": "35bd4e2b481a1a84a22215b5372672cf81460a671816960ddb206464359e1822",
+         "bolt11": "lntb10n1p0jga20pp53h7k2w8wkvuprjg3ff6l0y4pgdeg6lc9vsln3s74wnfsjl5fzrqqdqdw3jhxazldahx2xqyjw5qcqp2sp5wut5jnhr6n7jd5747ky2g5flmw7hgx9yjnqzu60ps2jf6f7tc0us9qy9qsqu2a0k37nckl62005p69xavlkydkvhnypk4dphffy4x09zltwh9437ad7xkl83tefdarzhu5t30ju5s56wlrg97qkx404pq3srfc425cq3ke9af"
+      }
+   ]
+}
+
+```
+
+
+AUTHOR
+------
+
+Vincenzo Palazzo <<vincenzo.palazzo@protonmail.com>> is mainly responsible.
+
+SEE ALSO
+--------
+
+lightning-listpays(7), lightning-listsendpays(7), lightning-paystatus(7).
+
+RESOURCES
+---------
+
+Main web site: <https://github.com/ElementsProject/lightning>
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -64,9 +64,11 @@ static const char *payment_status_to_string(const enum wallet_payment_status sta
 		return "complete";
 	case PAYMENT_FAILED:
 		return "failed";
-	default:
+	case PAYMENT_PENDING:
 		return "pending";
 	}
+	//This should never happen
+	abort();
 }
 
 
@@ -1528,35 +1530,34 @@ static struct command_result *json_delpay(struct command *cmd,
 
 	if (!param(cmd, buffer, params,
 		p_req("payment_hash", param_sha256, &payment_hash),
-		p_opt("status", param_string, &status_str),
+		p_req("status", param_string, &status_str),
 		NULL))
 		return command_param_failed();
 
-	if (!status_str)
-		status_str = "complete";
-
 	if (!string_to_payment_status(status_str, &status))
-		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
-						"Unrecognized status: %s", status_str);
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Unrecognized status: %s", status_str);
 
-	if (status == PAYMENT_PENDING)
-		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Invalid status: %s",
-							payment_status_to_string(status));
+	switch(status){
+		case PAYMENT_COMPLETE:
+		case PAYMENT_FAILED:
+			break;
+		case PAYMENT_PENDING:
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Invalid status: %s",
+				payment_status_to_string(status));
+	}
 
 	payments = wallet_payment_list(cmd, cmd->ld->wallet, payment_hash);
 
 	if (tal_count(payments) == 0)
-		return command_fail(cmd, PAY_NO_SUCH_PAYMENT,
-			"Unknown payment with payment_hash: %s",
+		return command_fail(cmd, PAY_NO_SUCH_PAYMENT, "Unknown payment with payment_hash: %s",
 				    type_to_string(tmpctx, struct sha256, payment_hash));
 
 	for (int i = 0; i < tal_count(payments); i++) {
 		if (payments[i]->status != status) {
-			return command_fail(cmd, PAY_STATUS_UNEXPECTED,
-					"Payment with hash %s has %s status but it should be %s",
+			return command_fail(cmd, PAY_STATUS_UNEXPECTED, "Payment with hash %s has %s status but it should be %s",
 					type_to_string(tmpctx, struct sha256, payment_hash),
 					payment_status_to_string(payments[i]->status),
-				        payment_status_to_string(status));
+					payment_status_to_string(status));
 		}
 	}
 
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -42,6 +42,34 @@ struct sendpay_command {
 	struct command *cmd;
 };
 
+static bool string_to_payment_status(const char *status_str, enum wallet_payment_status *status)
+{
+	if (streq(status_str, "complete")) {
+		*status = PAYMENT_COMPLETE;
+		return true;
+	} else if (streq(status_str, "pending")) {
+		*status = PAYMENT_PENDING;
+		return true;
+	} else if (streq(status_str, "failed")) {
+		*status = PAYMENT_FAILED;
+		return true;
+	}
+	return false;
+}
+
+static const char *payment_status_to_string(const enum wallet_payment_status status)
+{
+	switch (status) {
+	case PAYMENT_COMPLETE:
+		return "complete";
+	case PAYMENT_FAILED:
+		return "failed";
+	default:
+		return "pending";
+	}
+}
+
+
 static void destroy_sendpay_command(struct sendpay_command *pc)
 {
 	list_del(&pc->list);
@@ -1485,6 +1513,73 @@ static const struct json_command listsendpays_command = {
 	"Show sendpay, old and current, optionally limiting to {bolt11} or {payment_hash}."
 };
 AUTODATA(json_command, &listsendpays_command);
+
+
+static struct command_result *json_delpay(struct command *cmd,
+						const char *buffer,
+						const jsmntok_t *obj UNNEEDED,
+						const jsmntok_t *params)
+{
+	struct json_stream *response;
+	const struct wallet_payment **payments;
+	const char *status_str;
+	enum wallet_payment_status status;
+	struct sha256 *payment_hash;
+
+	if (!param(cmd, buffer, params,
+		p_req("payment_hash", param_sha256, &payment_hash),
+		p_opt("status", param_string, &status_str),
+		NULL))
+		return command_param_failed();
+
+	if (!status_str)
+		status_str = "complete";
+
+	if (!string_to_payment_status(status_str, &status))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						"Unrecognized status: %s", status_str);
+
+	if (status == PAYMENT_PENDING)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Invalid status: %s",
+							payment_status_to_string(status));
+
+	payments = wallet_payment_list(cmd, cmd->ld->wallet, payment_hash);
+
+	if (tal_count(payments) == 0)
+		return command_fail(cmd, PAY_NO_SUCH_PAYMENT,
+			"Unknown payment with payment_hash: %s",
+				    type_to_string(tmpctx, struct sha256, payment_hash));
+
+	for (int i = 0; i < tal_count(payments); i++) {
+		if (payments[i]->status != status) {
+			return command_fail(cmd, PAY_STATUS_UNEXPECTED,
+					"Payment with hash %s has %s status but it should be %s",
+					type_to_string(tmpctx, struct sha256, payment_hash),
+					payment_status_to_string(payments[i]->status),
+				        payment_status_to_string(status));
+		}
+	}
+
+	wallet_payment_delete_by_hash(cmd->ld->wallet, payment_hash);
+
+	response = json_stream_success(cmd);
+	json_array_start(response, "payments");
+	for (int i = 0; i < tal_count(payments); i++) {
+		json_object_start(response, NULL);
+		json_add_payment_fields(response, payments[i]);
+		json_object_end(response);
+	}
+	json_array_end(response);
+	return command_success(cmd, response);
+}
+
+static const struct json_command delpay_command = {
+	"delpay",
+	"payment",
+	json_delpay,
+	"Delete payment with {payment_hash} and {status}",
+};
+AUTODATA(json_command, &delpay_command);
 
 static struct command_result *json_createonion(struct command *cmd,
 						const char *buffer,
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -3263,6 +3263,107 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
     l1.rpc.pay(inv)
 
 
+def test_delpay_argument_invalid(node_factory, bitcoind):
+    """
+    This test includes all possible combination of input error inside the
+    delpay command.
+    """
+
+    l1, l2 = node_factory.get_nodes(2)
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay()
+
+    # invoice unpayed
+    inv = l1.rpc.invoice(10 ** 5, 'inv', 'inv')
+    payment_hash = inv["payment_hash"]
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash)
+
+    # payment unpayed with wrong status (pending status is a illegal input)
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'pending')
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'invalid_status')
+
+    l2.rpc.connect(l1.info['id'], 'localhost', l1.port)
+    l2.fund_channel(l1, 10 ** 6)
+
+    bitcoind.generate_block(6)
+    sync_blockheight(bitcoind, [l1, l2])
+
+    wait_for(lambda: len(l2.rpc.listchannels()['channels']) == 2)
+
+    l2.rpc.pay(inv['bolt11'])
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'failed')
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'pending')
+
+    assert len(l2.rpc.listpays()['pays']) == 1
+
+    # test if the node is still ready
+    payments = l2.rpc.delpay(payment_hash, 'complete')
+
+    assert payments['payments'][0]['bolt11'] == inv['bolt11']
+    assert len(payments['payments']) == 1
+    assert len(l2.rpc.listpays()['pays']) == 0
+
+
+def test_delpay(node_factory, bitcoind):
+    """
+    This unit test try to catch some error inside the command
+    delpay when it receives the correct input from the user
+    """
+
+    l1, l2 = node_factory.get_nodes(2)
+
+    amount_sat = 10 ** 6
+
+    # create l2->l1 channel.
+    l2.fundwallet(amount_sat * 5)
+    l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
+    l2.rpc.fundchannel(l1.info['id'], amount_sat * 3)
+
+    # Let the channel confirm.
+    bitcoind.generate_block(6)
+    sync_blockheight(bitcoind, [l1, l2])
+
+    invl1 = l1.rpc.invoice(Millisatoshi(amount_sat * 2 * 1000), "j", "j")
+    l2.rpc.pay(invl1["bolt11"])
+
+    before_del_pay = l2.rpc.listpays()
+
+    l2.rpc.delpay(invl1["payment_hash"])
+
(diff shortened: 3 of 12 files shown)


doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,58 @@
+lightning-delpay -- Command for removing a completed payment
+============================================================
+
+SYNOPSIS
+--------
+
+**delpay** *payment_hash* \[status\] (is not specify the status is complete)
```
ZmnSCPxj, line 7: You have a `partid` as well that is not documented.

Defaults should be described in "DESCRIPTION" section, not in the synopsis.

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,58 @@
...
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. However, the command doesn't permit to remove the pending payment.
+
```
ZmnSCPxj, line 14: DESCRIPTION section needs to describe every parameter, and describe defaults for that parameter as well if it is optional and not specified.

ZmnSCPxj:
In addition, it seems to me better to just delete *all* `partid`s, not delete them one-by-one.  Loss of one `partid` of a multi-part payment is already a loss of information anyway, so just delete them all in one database transaction.

vincenzopalazzo:
Hi @ZmnSCPxj, thanks to review this PR.

I will make the changes inside the docs and inside the JSON RPC errors, sorry about that.

In addition, about your suggestion.

>In addition, it seems to me better to just delete all partid, not delete them one-by-one. Loss of one partid of a multi-part payment is already a loss of information anyway, so just delete them all in one database transaction.

For my code comprehension, to make this work, so, to delete the complete payment and not each payment by `partid`, I need a new function inside the wallet.c to run the query "DELETE FROM payments WHERE payment_hash=?"

I'm losing something?

Maybe I can use the `function wallet_payment_list` and run for each element inside the list `wallet_payment_delete` but I think that the first solution is cleaner but I don't know if it is safe to run on the lightning node the delete query by `hash`?

ZmnSCPxj:
`DELETE FROM TABLE payments WHERE payment_hash=?` should be fine.  You might want to do some check to check if this results in any changes (I think we have a function to check the number of changes a query makes) and if it is 0 that is probably fine.

But you want to check the state of the *entire* payment first. The logic for a purported "payment state":

* If any part has succeeded, then the entire payment succeeded.
* If any part is pending, then the entire payment is pending.
* Otherwise, all parts are failed, so the entire payment is failed.

It is safe to delete a payment if the entire payment is failed or succeeded, and we should probably disallow deleting pending payments (which might still be operated by `pay` over in the `pay` plugin).

Note that if you do not pass control to a function with a callback, then the entire processing is done inside a database transaction, so you are assured that between the check and the deletion, the database state is consistent.  But if you pass control to a function with a callback, the database transaction will end at that point, and by the time the callback is called, you are in a new transaction.  So do not call into any callback-using functions between checking the state of the payment and deleting the whole payment.

vincenzopalazzo:
Thanks for the suggestion @ZmnSCPxj.

>You might want to do some check to check if this results in any changes (I think we have a function to check the number of changes a query makes) and if it is 0 that is probably a "payment not found" error.

I make the sanity test of input before to call the database function, but I can improve the check maybe I can make the following test:

1. payment not found: I used the `wallet_payment_by_hash` to find the payment, I think that is possible use this function to get the payment with hash and *partid=0*, if exist this part, the payment exists, right?

2. payment status not valid: from your previous message I understand the payment status is decided from the status of all part, so, I should check the status of all payment find with `wallet_payment_list` and check the status for each payment and make the correct decision.

of course, there are all checks on the input but, they are not relevant into this description.

>It is safe to delete a payment if the entire payment is failed or succeeded, and we should probably disallow deleting pending payments (which might still be operated by pay over in the pay plugin).

Yep, I agree, with my actual logic inside the `delpay` command you can remove only the payment with the status `complete` or `failed`, but of course, with your suggestion the logic change a little.

ZmnSCPxj:
`partid` 0 is for non-split payments.  If we split payments at all, there might not be a `partid` 0.

vincenzopalazzo:
Rebased https://github.com/ElementsProject/lightning/commit/01a82d38f769a410e4f2d127c9f09e8f8b296554 and updated https://github.com/vincenzopalazzo/lightning/commit/3a5f0be11e8f9efe6bc7796a0fd3289e83aebc52

doc/lightning-delpay.7.md
vincenzopalazzo: Updaed

doc/lightning-delpay.7.md
vincenzopalazzo: Updated but the doc needs a second review in the next hours (Caused from my Dyslexia problem :smile: )

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,83 @@
...
+**delpay** *payment_hash* \[status\]
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. However, the command doesn't permit to remove a pending payment.
+
+- *payment_hash*: Rapresents the unique identifier of a payment. To find it, the caller can run **listpays** or **listsendpays**;
+- *status*: Represents a payment status. Otherwise, if it is not specified it is equal to *complete*.
```
ZmnSCPxj, line 16: You should list here the possible values.

```
*status* is the expected status of the payment.
It can be *complete* or *pending*.
Only delete if the payment status matches.
If not specified, defaults to *complete*.
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,83 @@
...
+On success, the command will return a payment object, such as the **listsendpays**. In addition, if the payment is a MPP (Multi part payment) the command return a list of 
+payments; a payment object for each partid.
+
+On failure, an error is returned and any payment is deleted. If the lightning process fails before responding, the
+caller should use lightning-listsentpays(7) or lightning-listpays(7) to query whether this payment was deleted or not.
+
+The following error codes may occur:
+
+- -32602: Some parameter missed or some parameter is malformed;
+- 211: Payment with payment_hash have a wrong status. To check the correct status run the command **paystatus**;
```
ZmnSCPxj, line 43:
```
*payment\_hash*
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,83 @@
+lightning-delpay -- Command for removing a completed or failed payment
+============================================================
+
+SYNOPSIS
+--------
+
+**delpay** *payment_hash* \[status\]
```
ZmnSCPxj, line 7: Escape the underscore (as well all instances)

```
*payment\_hash*
```

lightningd/pay.c
```diff
@@ -1496,6 +1521,77 @@ static const struct json_command listsendpays_command = {
...
+	enum wallet_payment_status status;
+	struct sha256 *payment_hash;
+
+	if (!param(cmd, buffer, params,
+		p_req("payment_hash", param_sha256, &payment_hash),
+		p_opt("status", param_string, &status_str),
+		NULL))
+		return command_param_failed();
+
+	if (!payment_hash)
```
ZmnSCPxj, line 1542: Check is unnecessary, `p_req` will check already.

lightningd/pay.c
```diff
@@ -1496,6 +1521,77 @@ static const struct json_command listsendpays_command = {
...
+{
+	struct json_stream *response;
+	const struct wallet_payment **payments;
+	const char *status_str;
+	enum wallet_payment_status status;
+	struct sha256 *payment_hash;
+
+	if (!param(cmd, buffer, params,
+		p_req("payment_hash", param_sha256, &payment_hash),
+		p_opt("status", param_string, &status_str),
```
ZmnSCPxj, line 1538: Can use instead:

```c
p_opt_def("status", param_string, &status_str, "complete")
```

lightningd/pay.c
```diff
@@ -1496,6 +1521,77 @@ static const struct json_command listsendpays_command = {
...
+		NULL))
+		return command_param_failed();
+
+	if (!payment_hash)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						"missing required parameter: payment_hash");
+
+
+	status = status_str == NULL ? PAYMENT_COMPLETE : string_to_payment_status(status_str);
```
ZmnSCPxj, line 1547: See above for `p_opt_def`.

lightningd/pay.c
```diff
@@ -1496,6 +1521,77 @@ static const struct json_command listsendpays_command = {
...
+
+	if (status == -1)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						"status insert %s wrong", status_str);
```
ZmnSCPxj, line 1551: "status insert"?  Better "Unrecognized status: %s".

lightningd/pay.c
```diff
@@ -42,6 +42,31 @@ struct sendpay_command {
 	struct command *cmd;
 };
 
+static enum wallet_payment_status string_to_payment_status(const char *status_str)
```
ZmnSCPxj, line 45: The usual convention for parsers that can fail is something like this:

```c
bool string_to_payment_status(const char *status_str, enum wallet_payment_status *)
```

Then later in use:

```
if (!string_to_payment_status(status_str, &status))
    return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Unrecognized status '%s'", status_str);
```

wallet/wallet.c
```diff
@@ -2682,6 +2682,17 @@ void wallet_payment_delete(struct wallet *wallet,
 	db_exec_prepared_v2(take(stmt));
 }
 
+void wallet_payment_delete_by_hash(struct wallet *wallet,
+				   const struct sha256 *payment_hash)
+{
+	struct db_stmt *stmt;
+	stmt = db_prepare_v2(
+		wallet->db, SQL("DELETE FROM payments WHERE payment_hash = ?"));
```
ZmnSCPxj, line 2690: Looks correct but @cdecker should check as well, I am no longer familiar with how we interact with the db.

lightningd/pay.c
```diff
@@ -1496,6 +1521,77 @@ static const struct json_command listsendpays_command = {
...
+
+
+	status = status_str == NULL ? PAYMENT_COMPLETE : string_to_payment_status(status_str);
+
+	if (status == -1)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						"status insert %s wrong", status_str);
+
+	if (status == PAYMENT_PENDING)
+		return command_fail(cmd, PAY_STATUS_UNEXPECTED, "invalid status: %s",
```
ZmnSCPxj, line 1554: This is an error in the argument, not in the status of the actual payment, so should be `JSONRPC2_INVALID_PARAMS`.

ZmnSCPxj (dismissed):

wallet/wallet.c
cdecker: This should work just fine. The SQL subsystem just takes a valid query and rewrites it for various dialects. That works well for this type of simple queries, but may be a bit more complex for things like upserts or last inserted IDs, which we use in some places.

lightningd/pay.c
vincenzopalazzo: @ZmnSCPxj, mh! maybe I'm losing something here, param_string accept only a `const char *str` and the function `o_opt_def` can not assign a new value at `status_str` variable.

With this comment, you mean that I need to declare some function that accepts a not const `char*`?

I think that I'm losing something, sorry about that

ZmnSCPxj: String constants *are* `const char *`.  You can do something like `const char *foo = "foo";`.  You can even do something like `const char *foo = "foo"; foo = "bar"`.

The *correct* way to write `const` is *after* the type.  So actually `const char *` should be written as `char const *`.  Then you read types from right to left.  So `char const *` is read as "pointer (`*`) to constant (`const`) character (`char`)".  Note that the pointer itself is not `const`, so you can modify the pointer, which is why `char const* foo; foo = "bar"; foo = "quux"` is valid.  What you cannot do is `char const* foo = "foo"; *foo = '0';`, because what is pointed *to* is constant.

lightningd/pay.c
vincenzopalazzo: Changed in https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

lightningd/pay.c
vincenzopalazzo: Changed https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

lightningd/pay.c
vincenzopalazzo: Changed https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

lightningd/pay.c
vincenzopalazzo: Changed https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

doc/lightning-delpay.7.md
vincenzopalazzo: Changed https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

doc/lightning-delpay.7.md
vincenzopalazzo: Changed https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

doc/lightning-delpay.7.md
vincenzopalazzo: Thanks for this suggestion. modify made into https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934

lightningd/pay.c
vincenzopalazzo: Thanks for this wonderful informations. I made a couple of change inside https://github.com/ElementsProject/lightning/commit/627a374820dd03876eae9e4a99d661cb39f63934 but I want to have more detail about these changes. I like your suggestion to use p_opt_def.

lightningd/pay.c
```diff
@@ -42,6 +42,34 @@ struct sendpay_command {
 	struct command *cmd;
 };
 
+static bool string_to_payment_status(const char *status_str, enum wallet_payment_status *status)
```
vincenzopalazzo, line 45: @ZmnSCPxj I understood your suggestions inside this function? or I'm introducing some mistake?

ZmnSCPxj: looks ok, but no time for in-depth review now, please wait....

tests/test_pay.py
```diff
@@ -3242,6 +3242,107 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+    l2.rpc.fundchannel(l1.info['id'], amount_sat * 3)
+
+    # Let the channel confirm.
+    bitcoind.generate_block(6)
+    sync_blockheight(bitcoind, [l1, l2])
+
+    invl1 = l1.rpc.invoice(Millisatoshi(amount_sat * 2 * 1000), "j", "j")
+    l2.rpc.pay(invl1["bolt11"])
+
+    before_del_pay = l2.rpc.listpays()
```
niftynei, line 3317: this test could easily be rolled into the previous one

lightningd/pay.c
```diff
@@ -1498,6 +1526,73 @@ static const struct json_command listsendpays_command = {
...
+
+	if (status == PAYMENT_PENDING)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Invalid status: %s",
+							payment_status_to_string(status));
+
+	payments = wallet_payment_list(cmd, cmd->ld->wallet, payment_hash);
+
+	if (tal_count(payments) == 0)
+		return command_fail(cmd, PAY_NO_SUCH_PAYMENT,
+			"Unknown payment with payment_hash: %s",
```
niftynei, line 1562: line this up with `cmd` on the previous line?

lightningd/pay.c
```diff
@@ -1498,6 +1526,73 @@ static const struct json_command listsendpays_command = {
...
+		p_opt("status", param_string, &status_str),
+		NULL))
+		return command_param_failed();
+
+	if (!status_str)
+		status_str = "complete";
+
+	if (!string_to_payment_status(status_str, &status))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						"Unrecognized status: %s", status_str);
```
niftynei, line 1552: needs to be aligned with `cmd`

lightningd/pay.c
```diff
@@ -1498,6 +1526,73 @@ static const struct json_command listsendpays_command = {
...
+
+	if (tal_count(payments) == 0)
+		return command_fail(cmd, PAY_NO_SUCH_PAYMENT,
+			"Unknown payment with payment_hash: %s",
+				    type_to_string(tmpctx, struct sha256, payment_hash));
+
+	for (int i = 0; i < tal_count(payments); i++) {
+		if (payments[i]->status != status) {
+			return command_fail(cmd, PAY_STATUS_UNEXPECTED,
+					"Payment with hash %s has %s status but it should be %s",
```
niftynei, line 1568: align with `cmd`?

lightningd/pay.c
```diff
@@ -42,6 +42,34 @@ struct sendpay_command {
...
+}
+
+static const char *payment_status_to_string(const enum wallet_payment_status status)
+{
+	switch (status) {
+	case PAYMENT_COMPLETE:
+		return "complete";
+	case PAYMENT_FAILED:
+		return "failed";
+	default:
```
niftynei, line 67: it's best practice not to use `default` for a switch statement. instead, name all of the cases.

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+SYNOPSIS
+--------
+
+**delpay** *payment\_hash* \[status\]
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. However, the command doesn't permit to remove a pending payment.
```
niftynei, line 13:
```suggestion
status. Deleting a `pending` payment is an error.
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+
+- *payment\_hash*: Rapresents the unique identifier of a payment. To find it, you can run **listpays** or **listsendpays**;
```
niftynei, line 15:
```suggestion
- *payment\_hash*: The unique identifier of a payment. Find in **listpays** or **listsendpays**;
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+- *status* is the expected status of the payment. It can be *complete* or *failed*.
```
niftynei, line 16:
```suggestion
- *status*: Expected status of the payment. Valid values are *complete* or *failed*. 
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. However, the command doesn't permit to remove a pending payment.
+
+- *payment\_hash*: Rapresents the unique identifier of a payment. To find it, you can run **listpays** or **listsendpays**;
+- *status* is the expected status of the payment. It can be *complete* or *failed*. 
+Only delete if the payment status matches. If not specified, defaults to *complete*.
```
niftynei, line 17:
```suggestion
Only deletes if the payment status matches. If not specified, defaults to *complete*.
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+    "payment_hash": "4fa2f1b001067ec06d7f95b8695b8acd9ef04c1b4d1110e3b94e1fa0687bb1e0",
+    "status": "complete"
+  }
+}
+```
+
+RETURN VALUE
+------------
+
+On success, the command will return a payment object, such as the **listsendpays**. In addition, if the payment is a MPP (Multi part payment) the command return a list of
```
niftynei, line 35:
```suggestion
On success, the command will return a payment object, such as the **listsendpays**. If the payment is an MPP (multi-part payment) a list of 
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+    "status": "complete"
+  }
+}
+```
+
+RETURN VALUE
+------------
+
+On success, the command will return a payment object, such as the **listsendpays**. In addition, if the payment is a MPP (Multi part payment) the command return a list of 
+payments; a payment object for each partid.
```
niftynei, line 36:
```suggestion
payments will be returned -- one payment object for each partid.
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+
+On failure, an error is returned and any payment is deleted. If the lightning process fails before responding, the
```
niftynei, line 38:
```suggestion
On failure, an error is returned. If the lightning process fails before responding, the
```

we don't delete the payment on failure?

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+caller should use lightning-listsentpays(7) or lightning-listpays(7) to query whether this payment was deleted or not.
+
+The following error codes may occur:
+
+- -32602: Some parameter missed or some parameter is malformed;
```
niftynei, line 43:
```suggestion
- -32602: Parameter missing or malformed;
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+- 211: Payment with payment\_hash have a wrong status. To check the correct status run the command **paystatus**;
```
niftynei, line 44:
```suggestion
- 211: Payment status mismatch. Check the correct status via **paystatus**;
```

niftynei:
Mostly grammatical/spelling fixes

vincenzopalazzo:
Rebase and update with suggestions.

Thanks @niftynei about the fixes and sorry for the spelling errors. I'm still searching somethings where I'm good to do but I'm sure that I'm very bad to write sentences :smile:

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+SYNOPSIS
+--------
+
+**delpay** *payment\_hash* \[status\]
+
+DESCRIPTION
+-----------
+
+The **delpay** RPC command removes a payment as given in **listsendpays** or **listpays** with a complete or failed
+status. Deleting a `pending` payment is an error.
```
cdecker, line 13:
```suggestion
The **delpay** RPC command deletes a payment with the given `payment_hash` if its status is either `complete` or `failed`. Deleting a `pending` payment is an error.
```

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+
+- *payment\_hash*: The unique identifier of a payment. Find it, you can run **listpays** or **listsendpays**;
+- *status*: Expected status of the payment. Valid values are *complete* or *failed*. 
+Only deletes if the payment status matches. If not specified, defaults to *complete*.
```
cdecker, line 17: Shouldn't an unspecified value for `status` mean we don't care, and either `complete` or `failed` will be removed?

doc/lightning-delpay.7.md
```diff
@@ -0,0 +1,84 @@
...
+    "payment_hash": "4fa2f1b001067ec06d7f95b8695b8acd9ef04c1b4d1110e3b94e1fa0687bb1e0",
+    "status": "complete"
+  }
+}
+```
+
+RETURN VALUE
+------------
+
+On success, the command will return a payment object, such as the **listsendpays**. If the payment is aa MPP (multi part payment) the command return a list of
```
cdecker, line 35:
```suggestion
If successful the command returns a payment object, in the same format as **listsendpays**.
If the payment is a multi-part payment (MPP) the command return a list of 
```

lightningd/pay.c
```diff
@@ -1498,6 +1528,70 @@ static const struct json_command listsendpays_command = {
...
+
+	if (!status_str)
+		status_str = "complete";
+
+	if (!string_to_payment_status(status_str, &status))
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Unrecognized status: %s", status_str);
+
+	if (status == PAYMENT_PENDING)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS, "Invalid status: %s",
+							payment_status_to_string(status));
```
cdecker, line 1557: Probably it's better to check if we have one of the valid statuses instead of not having the third. If we ever add a new status this is would not cover that fourth state. If you use a `switch` statement then the compiler will make sure we cover all cases exhaustively.

wallet/wallet.c
```diff
@@ -2669,7 +2669,7 @@ void wallet_payment_store(struct wallet *wallet,
 	if (payment->bolt11 != NULL)
 		db_bind_text(stmt, 10, payment->bolt11);
 	else
-		db_bind_null(stmt, 10);
+	       	db_bind_null(stmt, 10);
```
cdecker, line 2672: Spurious whitespace change.

wallet/wallet.h
```diff
@@ -1026,6 +1026,15 @@ void wallet_payment_delete(struct wallet *wallet,
...
 			   u64 partid);
 
+/**
+ * wallet_payment_delete_by_hash - Remove a payment
+ *
+ * Removes the payment from the database by hash; if it is a MPP payment
+ * it remove all parts with a single query.
+ */
+void wallet_payment_delete_by_hash(struct wallet *wallet,
+				   const struct sha256 *payment_hash);
```
cdecker, line 1036: Nit: indentation of the followup line should match the `struct` keyword above (might be a GH display issue though).

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
     l1.rpc.pay(inv)
 
 
+def test_delpay_argument_invalid(node_factory, bitcoind):
+    """
+    This test includes all possible combination of input error inside the
```
cdecker, line 3266:
```suggestion
    This test includes all possible combinations of input error inside the
```

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+    l1, l2 = node_factory.get_nodes(2)
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay()
+
+    # invoice unpaid
+    inv = l1.rpc.invoice(10 ** 5, 'inv', 'inv')
+    payment_hash = inv["payment_hash"]
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash)
```
cdecker, line 3279: Well, really this matches the "no payment yet" case, since `l2` doesn't know anything about that payment. So you could just use `payment_hash = "AA"*32` here.

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+
+    # payment unpaid with wrong status (pending status is a illegal input)
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'pending')
```
cdecker, line 3283: Payment was no attempted at all, so it'd fail for that reason already.

cdecker: Basically this is testing our sanity check order, and not the underlying `pending` vs. `non-pending` cases.

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'invalid_status')
+
+    l2.rpc.connect(l1.info['id'], 'localhost', l1.port)
+    l2.fund_channel(l1, 10 ** 6)
```
cdecker, line 3289: You could have used `l1, l2 = node_factory.line_graph(2)` above since so far the channel has not been used. That'd save a couple of test-lines

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+
+    bitcoind.generate_block(6)
+    sync_blockheight(bitcoind, [l1, l2])
+
+    wait_for(lambda: len(l2.rpc.listchannels()['channels']) == 2)
```
cdecker, line 3294: These would have also been subsumed by `line_graph`

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+
+    l2.rpc.pay(inv['bolt11'])
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'failed')
+
+    with pytest.raises(RpcError):
+        l2.rpc.delpay(payment_hash, 'pending')
```
cdecker, line 3302: This is redundant with the test on line 3283

tests/test_pay.py
```diff
@@ -3261,6 +3261,77 @@ def test_mpp_presplit_routehint_conflict(node_factory, bitcoind):
...
+def test_delpay_payment_split(node_factory, bitcoind):
+    """
+    Test behavior of delpay with an MPP
+    """
+    MPP_TARGET_SIZE = 10**7  # Taken from libpluin-pay.c
+    amt = 5 * MPP_TARGET_SIZE
+
+    l1, l2, l3 = node_factory.line_graph(
+        3, fundamount=10**8, wait_for_announce=True,
+        opts={'wumbo': None}
```
cdecker, line 3323: Wumbo is not needed here: `MPP_TARGET_SIZE` is denominated in msats, whereas the `fundamount` is denominated in sats. So this creates a channel that is 4 orders of magnitude larger than the amount being transferred:

 - Channel capacity: 10**8 * 1000 = 10**11
 - Payment amount: 5 * 10**7

Doesn't hurt though :wink:

doc/lightning-delpay.7.md
ZmnSCPxj: Since `delpay` is supposed to mirror `delinvoice`, which has the `*status*` as a required arg, maybe it should be a required arg here as well?

vincenzopalazzo: Yep I agree with this point! thanks for clarifying!

lightningd/pay.c
vincenzopalazzo: Thanks for this suggestion, I made the change from if to switch!

wallet/wallet.c
vincenzopalazzo: Ops! thanks

tests/test_pay.py
vincenzopalazzo: This comment brings me to reflect on the sens of this test!

I moved the pay command to the top of this sanity check, so, with this new situation the unit test has more sense to me because it does the sanity check and the command behavior checks

tests/test_pay.py
vincenzopalazzo: In these cases, I could be used the following line of code to have the channels ready to work `node_factory.line_graph(2, fundamount=10**5, wait_for_announce=True)` , right?

tests/test_pay.py
vincenzopalazzo: >Doesn't hurt though

Yes, it is, but it is also my logic error, I discovered only now that `fundamount` is denominated in sats. I think that `10**5` is enough

vincenzopalazzo:
Rebase and update with review https://github.com/ElementsProject/lightning/commit/3314a059779c4769771a4d086b772b9c40827c14

Waiting to Travis test result

cdecker:
Rebased on top of `master`

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/3899/commits/56a5dc93d43bbadaa64b5be9dd0c07189f5dab27

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3899/commits/db5323522e23bb5f918ec0d96790316212758621

PR #3833, opened 2020-07-15
Jon wally update

jgriffiths:
Update to latest wally candidate

This will require re-working of #3902 somewhat, since that code uses the old PSBT interfaces. I'll comment there on what needs updating.

Changelog-None

PR #3902, opened 2020-08-04
Dual funding, PSBT work

niftynei:
This PR contains just the PSBT utilities needed for the dual-funding accepter PR. 

This allows us to:
- add a serial id to psbt inputs + outputs
- sort psbt inputs + outputs by serial_id
- produce a 'diff' between two psbts (must have serial_ids assigned to inputs + outputs though) 

Changelog-None


bitcoin/psbt.c
```diff
@@ -544,6 +545,119 @@ struct amount_sat psbt_output_get_amount(struct wally_psbt *psbt,
 	assert(amount_asset_is_main(&asset));
 	return amount_asset_to_sat(&asset);
 }
+
+static void add(u8 **key, const void *mem, size_t len)
+{
+	size_t oldlen = tal_count(*key);
+	tal_resize(key, oldlen + len);
+	memcpy(*key + oldlen, mem, len);
```
rustyrussell, line 548: Let's add a memcheck() around this, so if there's something added which is uninitialized, valgrind backtraces to the caller (not later when we use it):

`memcpy(*key + oldlen, memcheck(mem, len), len)`

bitcoin/psbt.c
```diff
@@ -544,6 +545,119 @@ struct amount_sat psbt_output_get_amount(struct wally_psbt *psbt,
...
+			abort();
+	}
+	memcpy(data, memcheck(value, value_len), value_len);
+	result = wally_add_new_unknown(in->unknowns,
+				       cast_const(unsigned char *, key),
+				       tal_bytelen(key),
+				       data, value_len) == WALLY_OK;
+
+	tal_free(data);
+	return result;
```
rustyrussell, line 613: You never document what `result` means; can this actually happen?  If so, document what it means in the header.  If not, don't return bool from this func.

bitcoin/psbt.c
```diff
@@ -544,6 +545,119 @@ struct amount_sat psbt_output_get_amount(struct wally_psbt *psbt,
...
+		       size_t *val_len)
+{
+	if (!map)
+		return NULL;
+
+	size_t key_len = tal_bytelen(key);
+	for (size_t i = 0; i < map->num_items; i++) {
+		struct wally_unknowns_item *item = &map->items[i];
+		if (key_len == item->key_len &&
+			memcmp(key, item->key, key_len) == 0) {
```
rustyrussell, line 627: memeq()

bitcoin/psbt.c
```diff
@@ -544,6 +545,119 @@ struct amount_sat psbt_output_get_amount(struct wally_psbt *psbt,
...
+			     size_t value_len)
+{
+	bool result;
+	u8 *data = tal_arr(NULL, u8, value_len);
+	if (!out->unknowns) {
+		if (wally_unknowns_map_init_alloc(1, &out->unknowns) != WALLY_OK)
+			abort();
+	}
+
+	memcpy(data, memcheck(value, value_len), value_len);
```
rustyrussell, line 647: Kudos for the memcheck(), but why does wally_add_new_unknown need a copy?

common/psbt_open.c
```diff
@@ -0,0 +1,378 @@
...
+bool psbt_get_serial_id(struct wally_unknowns_map *map, u16 *serial_id)
+{
+	u8 *key = psbt_make_key(tmpctx, PSBT_TYPE_SERIAL_ID, NULL);
+	size_t value_len;
+	void *result = psbt_get_unknown(map, key, &value_len);
+	if (!result)
+		return false;
+	assert(value_len == sizeof(*serial_id));
+	memcpy(serial_id, result, value_len);
+	return true;
```
rustyrussell, line 19: This seems a bit round-about, allocating a key to do the comparison.  Either we could implement psbt_get_lightning(), or if we want to be generic, have psbt_get_unknown() take the prefix and the keyvalue?

common/psbt_open.c
```diff
@@ -0,0 +1,378 @@
...
+#include <common/utils.h>
+
+bool psbt_get_serial_id(struct wally_unknowns_map *map, u16 *serial_id)
+{
+	u8 *key = psbt_make_key(tmpctx, PSBT_TYPE_SERIAL_ID, NULL);
+	size_t value_len;
+	void *result = psbt_get_unknown(map, key, &value_len);
+	if (!result)
+		return false;
+	assert(value_len == sizeof(*serial_id));
```
rustyrussell, line 17: I worry this is a vector by which bad PSBTs can crash us.  Perhaps return false in this case?

common/psbt_open.c
```diff
@@ -0,0 +1,378 @@
...
+
+bool psbt_input_get_max_witness_len(struct wally_psbt_input *input,
+				    u16 *max_witness_len)
+{
+	u8 *key = psbt_make_key(tmpctx, PSBT_TYPE_MAX_WITNESS_LEN, NULL);
+	size_t value_len;
+	void *result = psbt_get_unknown(input->unknowns, key, &value_len);
+	if (!result)
+		return false;
+	assert(value_len == sizeof(*max_witness_len));
```
rustyrussell, line 339: Here too?

common/psbt_open.h
```diff
@@ -0,0 +1,144 @@
...
+void psbt_sort_by_serial_id(struct wally_psbt *psbt);
+
+/* psbt_has_diff - Returns set of diffs btw orig + new psbt
+ *
+ * Note that diff is based on *serial_ids*. All inputs+outputs
+ * MUST have a serial_id field present before calling this.
+ *
+ * If a serial_id has remained the same, but the underlying data
+ * has changed, this WILL NOT detect it.
+ *
```
rustyrussell, line 54: Not true any more?

common/psbt_open.h
```diff
@@ -0,0 +1,144 @@
...
+ * @ctx - allocation context for returned diffsets
+ * @orig - original psbt
+ * @new - updated psbt
+ * @added_ins - inputs added to {new}
+ * @rm_ins - inputs removed from {new}
+ * @added_outs - outputs added to {new}
+ * @rm_outs - outputs removed from {new}
+ *
+ * Returns true if no changes are found.
```
rustyrussell, line 63: No it doesn't!

Also probably needs warning about sets simply referring into orig and new here, not on implementation?

jgriffiths:
Assuming #3833 is merged (as the wally changes are in master), this PR will need updating to reflect the current wally interface changes. The major ones are:

- Maps are all the same type now, see the wally_map_xx functions in wally_psbt.h
- You don't need to create new maps if they are empty, just call wally_map_add and it will do the right thing
- wally_map_find should replace your hand-rolled search loops
- You can now add tx inputs and outputs directly to the psbt at any position and it will update the tx in step.

Other comments:
- Per #3833 in psbt_append_input you don't need to allocate an input to add it, but you should use the elements init function rather than messing with the features bits yourself (they are an impl detail its better if you don't poke around in)
- #define WALLY_PSBT_PROPRIETARY 0xFC you can just use PSBT_PROPRIETARY_TYPE  from wally_psbt.h

niftynei:
Ok, updated to the latest things. There's a failing unit test I'll have to address tomorrow.

bitcoin/psbt.c
```diff
@@ -80,6 +110,28 @@ struct wally_psbt_input *psbt_add_input(struct wally_psbt *psbt,
...
+	insert_at = psbt->num_inputs;
+
+	if (wally_tx_input_init_alloc(txid->shad.sha.u.u8,
+				      sizeof(struct bitcoin_txid),
+				      outnum, sequence, NULL, 0, NULL,
+				      &tx_in) != WALLY_OK)
+		abort();
+
+	tx_in->features = chainparams->is_elements ? WALLY_TX_IS_ELEMENTS : 0;
+	input = psbt_add_input(psbt, tx_in, insert_at);
```
jgriffiths, line 126: features is an internal impl detail and we may end up using it for more than just elements. please use wally_tx_elements_input_init_alloc for elements instead.

common/psbt_open.c
```diff
@@ -0,0 +1,389 @@
...
+			return false;
+
+		/* Required because we send the full tx over the wire now */
+		if (!input->utxo)
+			return false;
+
+		/* If is P2SH, redeemscript must be present */
+		const u8 *outscript =
+			wally_tx_output_get_script(tmpctx,
+				&input->utxo->outputs[psbt->tx->inputs[i].index]);
```
jgriffiths, line 382: I think you want to assert this is in bounds for the utxo outputs before dereferencing here, just to be sure.

jgriffiths:
Note that your lightning keys and values in maps need to be stored endian-independent - so rather than memcpy for your serial id you'll need to use cpu_to/from_<storage endianess> before copying, or things will go badly wrong when swapping PSBTs with an instance on a difference arch.

common/psbt_open.c
niftynei: ack, updated

bitcoin/psbt.c
niftynei: ack, updated

common/psbt_open.c
```diff
@@ -89,8 +91,12 @@ static const u8 *linearize_output(const tal_t *ctx,
 
 	if (wally_tx_add_output(psbt->tx, tx_out) != WALLY_OK)
 		abort();
+
 	psbt->outputs[0] = *out;
 	psbt->num_outputs++;
+	/* Blank out unknowns. These are unordered and serializing
+	 * them might have different outputs for identical data */
+	psbt->outputs[0].unknowns.num_items = 0;
```
jgriffiths, line 102: wally_map_sort(&psbt->outputs[0].unknowns) can be used now that its been merged to wally master.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/3902/commits/f99dbfa87191cdd92d95e20553e7870effd9ea05

PR #3904, opened 2020-08-05
amount_sat/amount_msat audit results

rustyrussell:
In #3902 @niftynei added a new helper: I took the opportunity to audit all our amount `/* Raw` comments and created a few new intuitive helpers.  In particular, importing amounts isn't really an issue, though exporting them is still something I prefer people to look at carefully.

Changelog-None

### DIFF
--- a/common/amount.c
+++ b/common/amount.c
@@ -297,6 +297,18 @@ WARN_UNUSED_RESULT bool amount_msat_add_sat(struct amount_msat *val,
 	return amount_msat_add(val, a, msatb);
 }
 
+WARN_UNUSED_RESULT bool amount_msat_scale(struct amount_msat *val,
+					  struct amount_msat sat,
+					  double scale)
+{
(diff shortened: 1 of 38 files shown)


common/amount.c
```diff
@@ -401,22 +401,20 @@ bool amount_msat_to_u32(struct amount_msat msat, u32 *millisatoshis)
 	return true;
 }
 
-void amount_msat_from_u64(struct amount_msat *msat, u64 millisatoshis)
+struct amount_msat amount_msat(u64 millisatoshis)
```
ZmnSCPxj, line 416: I like this better than the `void foo(out* out, int input)` style.  This is in fact what I thought `AMOUNT_SAT`/`AMOUNT_MSAT` did back when they were introduced, and was disappointed to learn they were just constexpr, boo.

The compiler might be able to optimize these better if we `static inline` these in the headers.

PR #3912, opened 2020-08-07
plugin: don't mark all plugins as important when registering one

darosior:
I skimmed through #3890 after merge and i think this is wrong as it would mark some or all of the already registered plugins as important when we register a new important plugin.

This follows the discussion with @ZmnSCPxj here https://github.com/ElementsProject/lightning/pull/3890/files#r466741870.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -154,9 +154,6 @@ struct plugin *plugin_register(struct plugins *plugins, const char* path TAKES,
 		if (streq(path, p_temp->cmd)) {
 			if (taken(path))
 				tal_free(path);
-		 	/* If added as "important", upgrade to "important".  */
-			if (important)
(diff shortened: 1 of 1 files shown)


lightningd/plugin.c
```diff
@@ -154,9 +154,6 @@ struct plugin *plugin_register(struct plugins *plugins, const char* path TAKES,
 		if (streq(path, p_temp->cmd)) {
```
niftynei, line 154: unless i'm mistaken, this line will filter out non-matching plugins. only plugins with an identical path to the desired registree will be marked as important (if so indicated).

darosior: Yeah you are right.

darosior:
Ok so brainfart on my end :sweat_smile:. Closing this.

PR #3908, opened 2020-08-06
Routehints get ignored if invoice >10,000 sats

ZmnSCPxj:
This PR fixes that problem.

Issue #3894, opened 2020-08-02
Thoughts on improving MPP


ZmnSCPxj:
EDIT: also put this long treatise in my blog: https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-August/002775.html

First, let us consider some C-Lightning node who, for some reason, only has unpublished channels to the network (note: in my opinion, ***unpublished channels delenda est!***).  Let us suppose this node wishes to receive some amount of funds, from multiple different payers.

Now, the payee node knows MPP exists, and thus, it is not very concerned that it has insufficient incoming capacity in a ***single*** channel to accept one of the payments.  Reasoning that MPP exists, it considers instead the *sum total* of all incoming capacities.  It sees that it has sufficient *sum total* incoming capacity that it can accept all payments from all payers, with some to spare.

However, at the channel-level detail, *none* of the channels has sufficient capacity to accept a single one of the payments the C-Lightning node now expects to receive, in full.  So, what should the C-Lightning node put as routehints in a generated invoice?

Another, is that we should note that I described "payers" with an "s" at the end.  In the English language as commonly used by puny humans, adding an "s" at the end of a noun is a common (but not universal) convention implying that the subject under discussion is actually composed of more than one item.  So not only must the C-Lightning node (all of whose channels are unpublished, thus cannot be paid unless it reveals at least some of the channels as routehints in the invoice) make an invoice with sufficient routehints to get an MPP delivered, it must make *multiple* invoices, such that the invoices it issues have as little overlap in their routehints as possible

Otherwise, if we provide the **same** set of routehints in invoices to two **different** payers, both of them will interfere with each other, possibly to the point that neither can pay to the payee (since e.g. some of the available capacity is taken up by parts of one payment while the rest is taken up by parts of the other payment, and neither payment can then complete in full due to this deadlock).

This is important in case the C-Lightning node receiver is completely unannounced (i.e. has only unpublished channels).  The payers cannot fall back to just trying directly routing on a different incoming channel of the receiver, as they may not have any knowledge of such channels.

I suggest, then, that the C-Lightning node that wants to get paid, and uses only unpublished channels, should use a Round-Robin strategy for selecting routehints.

To implement this, the `invoice` command maintains a persistent list of all channels the node has.  When a newly funded channel is locked-in by both sides (transitions to `CHANNELD_NORMAL`) it is simply appended to this list.  When a channel leaves `CHANNELD_NORMAL`it is simply removed from this list.  Let us call this the round-robin list.

[...]

PR #3913, opened 2020-08-08
Invoice round robin

ZmnSCPxj:
This implements part of my suggestions in #3894.

The drawback is that unpublished nodes with this commit could potentially be harder to pay by other implementations that do not have any MPP at all yet.

This change basically embraces MPP and holds onto it tight and whispers lovingly into its ears "I want the whole world to have MPP".  On the other hand it does show that our MPP impl seems to be working nicely enough.

It was implementing this that revealed #3908 , if you are curious.  Yet another reason to abolish unpublished channels....

There is yet another bug in our MPP that this code also reveals, which is basically that if we get a blockheight disagreement between payee and payer, we are too aggressive with advancing the routehint pointer.  I will make a new test and fix for that latent bug.

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 from decimal import Decimal
 from ephemeral_port_reserve import reserve
 from pyln.client import LightningRpc
+from pyln.client import Millisatoshi
 
 import json
 import logging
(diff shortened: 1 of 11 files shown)


ZmnSCPxj:
The errors occurring, while they occur in `test_mpp_interference_2`, occur at the `fundchannel` stage where we are waiting for certain logs to appear, and before any attempt at MPP payment.  Weird.  `COMPAT=0` tho....

ZmnSCPxj:
Disabled the test at `DEVELOPER=0`, turns out it takes 10 minutes to set up the channels if `DEVELOPER=0`.

ZmnSCPxj:
Added the routehint randomization suggested here: https://github.com/ElementsProject/lightning/pull/3914#issuecomment-671056394

My initial non-exhaustive tests suggest it improves the success rate and speed of `test_mpp_interference_2`, but not quite enough yet to remove `@flaky` from it.

A proper round-robin allocation of routehints would, I think, improve this even more, but the random-base idea is "good enough" in practice IMO and is very very simple to implement.

ZmnSCPxj:
Rebased now that #3914 is on master.

ZmnSCPxj:
Rebased again.

rustyrussell:
I like this, just want to combine some logic.  We measure the incoming capacity both here and in the listpeers code (which does a more thorough job), so we should combine those.

I also dislike the duplication between the mpp and non-mpp, so I'll refactor that as well while I'm at it.

Just waiting for Travis, which I had to kick...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/6972e1fff4983e5e0c64f4ecfb732491a2da0c8e

PR #3914, opened 2020-08-09
Waitblockheight MPP bugs

ZmnSCPxj:
So I stumbled on a bug where the payee has a longer chain than the payer while developing #3913.

I wrote this little test plus fix, but it looks like we still have issues with being too aggressive in returning a *transient* error permanently!

Here are logs on the payer, after the (partial fix) is done on the payer side:

```
b'2020-08-09T07:12:07.387Z DEBUG plugin-pay: Added a channel hint for 103x1x1/1: enabled true, estimated capacity 4294967295msat'
b"2020-08-09T07:12:07.388Z DEBUG plugin-pay: After filtering routehints we're left with 1 usable hints"
b'2020-08-09T07:12:07.388Z DEBUG plugin-pay: Asking gossipd whether 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d is reachable without routehints.'
b'2020-08-09T07:12:07.391Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:07.391Z DEBUG plugin-pay: The destination is not directly reachable excluding attempts without routehints'
b'2020-08-09T07:12:07.391Z INFO plugin-pay: Split into 3 sub-payments due to initial size (20000000msat > 10000000msat)'
b'2020-08-09T07:12:07.395Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:07.396Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:07.396Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:07.699Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:07.700Z INFO plugin-pay: Remote node appears to be on a longer chain, which causes CLTV timeouts to be incorrect. Waiting up to 59 seconds to catch up to block 113 before retrying.'
b'2020-08-09T07:12:07.700Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:07.700Z INFO plugin-pay: Remote node appears to be on a longer chain, which causes CLTV timeouts to be incorrect. Waiting up to 59 seconds to catch up to block 113 before retrying.'
b'2020-08-09T07:12:07.700Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:07.700Z INFO plugin-pay: Remote node appears to be on a longer chain, which causes CLTV timeouts to be incorrect. Waiting up to 59 seconds to catch up to block 113 before retrying.'
b'2020-08-09T07:12:08.226Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:08.227Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:08.227Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T07:12:08.558Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:08.559Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:08.559Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T07:12:08.678Z INFO plugin-pay: Killing plugin: Plugin exited before completing handshake.'
```

And here are the relevant logs on the payee:

```
b'2020-08-09T07:12:07.558Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/peer_htlcs.c:503'
b'2020-08-09T07:12:07.562Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/peer_htlcs.c:503'
b'2020-08-09T07:12:07.566Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/peer_htlcs.c:503'
b'2020-08-09T07:12:08.421Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/htlc_set.c:113'
b'2020-08-09T07:12:08.425Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/htlc_set.c:113'
b'2020-08-09T07:12:08.429Z DEBUG lightningd: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS: lightningd/htlc_set.c:113'
```

I find it very strange that payee still thinks there is some problem with the incoming payment.  So it looks like we have yet another bug.... LOL

@cdecker for help at the payer side (I think it is a genuine bug as I find the logic of the `d->offset` fairly weird) and @rustyrussell for help at the payee side (I have no idea about how the `htlc_set` works, and why the source of the error is different in the initial case (where the payer genuinely has the wrong blockheight) then in the second case (where the payer should have advanced its view of the blockheight already)).

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -384,6 +384,7 @@ static struct command_result *payment_getroute_result(struct command *cmd,
 	/* Ensure that our fee and CLTV budgets are respected. */
 	if (amount_msat_greater(fee, p->constraints.fee_budget)) {
 		payment_exclude_most_expensive(p);
+		p->route = tal_free(p->route);
 		payment_fail(
 		    p, "Fee exceeds our fee budget: %s > %s, discarding route",
 		    type_to_string(tmpctx, struct amount_msat, &fee),
@@ -393,9 +394,11 @@ static struct command_result *payment_getroute_result(struct command *cmd,
 	}
 
 	if (p->route[0].delay > p->constraints.cltv_budget) {
+		u32 delay = p->route[0].delay;
 		payment_exclude_longest_delay(p);
+		p->route = tal_free(p->route);
 		payment_fail(p, "CLTV delay exceeds our CLTV budget: %d > %d",
-			     p->route[0].delay, p->constraints.cltv_budget);
+			     delay, p->constraints.cltv_budget);
 		return command_still_pending(cmd);
 	}
 
@@ -1760,12 +1763,17 @@ static struct route_info **filter_routehints(struct routehints_data *d,
 	return tal_steal(d, hints);
 }
 
+static bool route_msatoshi(struct amount_msat *total,
+			   const struct amount_msat msat,
+			   const struct route_info *route, size_t num_route);
+
 static bool routehint_excluded(struct payment *p,
 			       const struct route_info *routehint)
 {
 	const struct node_id *nodes = payment_get_excluded_nodes(tmpctx, p);
 	const struct short_channel_id_dir *chans =
 	    payment_get_excluded_channels(tmpctx, p);
+	const struct channel_hint *hints = payment_root(p)->channel_hints;
 
 	/* Note that we ignore direction here: in theory, we could have
 	 * found that one direction of a channel is unavailable, but they
@@ -1779,6 +1787,41 @@ static bool routehint_excluded(struct payment *p,
 		for (size_t j = 0; j < tal_count(chans); j++)
 			if (short_channel_id_eq(&chans[j].scid, &r->short_channel_id))
 				return true;
+
+		/* Skip the capacity check if this is the last hop
+		 * in the routehint.
+		 * The last hop in the routehint delivers the exact
+		 * final amount to the destination, which
+		 * payment_get_excluded_channels uses for excluding
+		 * already.
+		 * Thus, the capacity check below only really matters
+		 * for multi-hop routehints.
+		 */
+		if (i == tal_count(routehint) - 1)
+			continue;
+
+		/* Check our capacity fits.  */
+		struct amount_msat needed_capacity;
+		if (!route_msatoshi(&needed_capacity, p->amount,
+				    r + 1, tal_count(routehint) - i - 1))
+			return true;
+		/* Why do we scan the hints again if
+		 * payment_get_excluded_channels already does?
+		 * Because payment_get_excluded_channels checks the
+		 * amount at destination, but we know that we are
+		 * a specific distance from the destination and we
+		 * know the exact capacity we need to send via this
+		 * channel, which is greater than the destination.
+		 */
+		for (size_t j = 0; j < tal_count(hints); j++) {
+			if (!short_channel_id_eq(&hints[j].scid.scid, &r->short_channel_id))
+				continue;
+			/* We exclude on equality because we set the estimate
+			 * to the smallest failed attempt.  */
+			if (amount_msat_greater_eq(needed_capacity,
+						   hints[j].estimated_capacity))
+				return true;
+		}
 	}
 	return false;
 }
@@ -2044,9 +2087,14 @@ static struct routehints_data *routehint_data_init(struct payment *p)
 		pd = payment_mod_routehints_get_data(payment_root(p));
 		d->destination_reachable = pd->destination_reachable;
 		d->routehints = pd->routehints;
-		if (p->parent->step == PAYMENT_STEP_RETRY)
-			d->offset = pd->offset + 1;
-		else
+		pd = payment_mod_routehints_get_data(p->parent);
+		if (p->parent->step == PAYMENT_STEP_RETRY) {
+			d->offset = pd->offset;
+			/* If the previous try failed to route, advance
+			 * to the next routehint.  */
+			if (!p->parent->route)
+				++d->offset;
+		} else
 			d->offset = 0;
 		return d;
 	} else {
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -721,12 +721,50 @@ static void report_tampering(struct payment *p,
 	}
 }
 
+static bool
+failure_is_blockheight_disagreement(const struct payment *p,
+				    u32 *blockheight)
+{
+	struct amount_msat unused;
+
+	assert(p && p->result);
+
+	if (p->result->failcode == 17 /* Former final_expiry_too_soon */)
+		*blockheight = p->start_block + 1;
+	else if (!fromwire_incorrect_or_unknown_payment_details(
+			p->result->raw_message,
+			&unused, blockheight))
+		/* If it's incorrect_or_unknown_payment_details, that tells us
+		 * what height they're at */
+		return false;
+
+	/* If we are already at the desired blockheight there is no point in
+	 * waiting, and it is likely just some other error. Notice that
+	 * start_block gets set by the initial getinfo call for each
+	 * attempt.*/
+	if (*blockheight <= p->start_block)
+		return false;
+
+	return true;
+}
+
 static struct command_result *
 handle_final_failure(struct command *cmd,
 		     struct payment *p,
 		     const struct node_id *final_id,
 		     enum onion_type failcode)
 {
+	u32 unused;
+
+	/* Need to check for blockheight disagreement case here,
+	 * otherwise we would set the abort flag too eagerly.
+	 */
+	if (failure_is_blockheight_disagreement(p, &unused)) {
+		plugin_log(p->plugin, LOG_DBG,
+			   "Blockheight disagreement, not aborting.");
+		goto nonerror;
+	}
+
 	/* We use an exhaustive switch statement here so you get a compile
 	 * warning when new ones are added, and can think about where they go */
 	switch (failcode) {
@@ -805,8 +843,10 @@ error:
 	p->result->code = PAY_DESTINATION_PERM_FAIL;
 	payment_root(p)->abort = true;
 
+nonerror:
 	payment_fail(p, "%s", p->result->message);
 	return command_still_pending(cmd);
+
 }
 
 
@@ -2508,9 +2548,7 @@ static void waitblockheight_cb(void *d, struct payment *p)
 	struct out_req *req;
 	struct timeabs now = time_now();
 	struct timerel remaining;
-	u32 blockheight = p->start_block;
-	int failcode;
-	const u8 *raw_message;
+	u32 blockheight;
 	if (p->step != PAYMENT_STEP_FAILED)
 		return payment_continue(p);
 
@@ -2521,27 +2559,10 @@ static void waitblockheight_cb(void *d, struct payment *p)
 	if (time_after(now, p->deadline))
 		return payment_continue(p);
 
-	failcode = p->result->failcode;
-	raw_message = p->result->raw_message;
 	remaining = time_between(p->deadline, now);
 
-	if (failcode == 17 /* Former final_expiry_too_soon */) {
-		blockheight = p->start_block + 1;
-	}  else {
-		/* If it's incorrect_or_unknown_payment_details, that tells us
-		 * what height they're at */
-		struct amount_msat unused;
-		const void *ptr = raw_message;
-		if (!fromwire_incorrect_or_unknown_payment_details(
-			ptr, &unused, &blockheight))
-			return payment_continue(p);
-	}
-
-	/* If we are already at the desired blockheight there is no point in
-	 * waiting, and it is likely just some other error. Notice that
-	 * start_block gets set by the initial getinfo call for each
-	 * attempt.*/
-	if (blockheight <= p->start_block)
+	/* *Was* it a blockheight disagreement that caused the failure?  */
+	if (!failure_is_blockheight_disagreement(p, &blockheight))
 		return payment_continue(p);
 
 	plugin_log(p->plugin, LOG_INFORM,
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1076,6 +1076,7 @@ static void payment_add_hop_onion_payload(struct payment *p,
 	u32 cltv = p->start_block + next->delay + 1;
 	u64 msat = next->amount.millisatoshis; /* Raw: TLV payload generation*/
 	struct tlv_field **fields;
+	struct payment *root = payment_root(p);
 	static struct short_channel_id all_zero_scid = {.u64 = 0};
 
 	/* This is the information of the node processing this payload, while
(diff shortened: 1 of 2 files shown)


cdecker:
Quite a scare you gave me there: it seems like after the `waitblockheight` modifier we add the *current* payment's amount to the onion payload, instead of the *total* payment amount. This is easily fixed by changing the following `msat` to `payment_root(p)->amount.millisatoshis` here:

https://github.com/ElementsProject/lightning/blob/723b7223b76e8bb5ab16f55018abd8a8833d4f9e/plugins/libplugin-pay.c#L1105-L1109

What bothers me more is that the existing code works for most MPP payments, just not in combination with `waitblockheight` which is strange. Still debugging where the difference comes from. The offset stuff is unlikely to be involved, and the receiver side is correct.

cdecker:
Ok, found the issue: there are three places we add the TLV payload, which may include the `total_msat` value. All use the the `root->amount`, except this one:

https://github.com/ElementsProject/lightning/blob/723b7223b76e8bb5ab16f55018abd8a8833d4f9e/plugins/libplugin-pay.c#L1105-L1109

Which is the only invocation that isn't mediated through the MPP modifiers. Using the root amount is correct always, however I still don't quite get why the MPP modifiers doesn't just overwrite the TLV field here:

https://github.com/ElementsProject/lightning/blob/723b7223b76e8bb5ab16f55018abd8a8833d4f9e/plugins/libplugin-pay.c#L2613-L2617

and here

https://github.com/ElementsProject/lightning/blob/723b7223b76e8bb5ab16f55018abd8a8833d4f9e/plugins/libplugin-pay.c#L2763-L2767

Still looking into that, but the fix should indeed just be using the `root->amount`.

ZmnSCPxj:
@cdecker the reason why I think the `d->offset` computations for `routehint_data_init` are wrong is because of these logs which appear if I do not change the `d->offset` compputations:

```
b'2020-08-09T13:20:26.702Z DEBUG plugin-pay: Added a channel hint for 103x1x1/1: enabled true, estimated capacity 4294967295msat'
b"2020-08-09T13:20:26.703Z DEBUG plugin-pay: After filtering routehints we're left with 1 usable hints"
b'2020-08-09T13:20:26.703Z DEBUG plugin-pay: Asking gossipd whether 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d is reachable without routehints.'
b'2020-08-09T13:20:26.705Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T13:20:26.706Z DEBUG plugin-pay: The destination is not directly reachable excluding attempts without routehints'
b'2020-08-09T13:20:26.706Z INFO plugin-pay: Split into 2 sub-payments due to initial size (20000000msat > 10000000msat)'
b'2020-08-09T13:20:26.708Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T13:20:26.708Z DEBUG plugin-pay: Using routehint 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 (105x1x1) cltv_delta=6'
b'2020-08-09T13:20:27.048Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T13:20:27.049Z INFO plugin-pay: Remote node appears to be on a longer chain, which causes CLTV timeouts to be incorrect. Waiting up to 59 seconds to catch up to block 113 before retrying.'
b'2020-08-09T13:20:27.049Z INFO plugin-pay: failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)'
b'2020-08-09T13:20:27.049Z INFO plugin-pay: Remote node appears to be on a longer chain, which causes CLTV timeouts to be incorrect. Waiting up to 59 seconds to catch up to block 113 before retrying.'
b'2020-08-09T13:20:27.629Z DEBUG plugin-pay: Not using a routehint'
b'2020-08-09T13:20:27.629Z DEBUG plugin-pay: Not using a routehint'
b'2020-08-09T13:20:27.633Z INFO plugin-pay: Error computing a route to 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d: \\"Could not find a route\\" (205)'
b'2020-08-09T13:20:27.633Z INFO plugin-pay: Error computing a route to 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d: \\"Could not find a route\\" (205)'
b'2020-08-09T13:20:27.807Z INFO plugin-pay: Killing plugin: Plugin exited before completing handshake.'
```

As you can see above, the sub-payments used to have a routehint, but after the `WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS` the routehint is forgotten.  And the `d->offset` initialization is wonky.  First, it checks if the parent is in RETRY state, but gets the offset from the ***root*** payment offset.  And it adds +1 instead of retrying the same offset as the retried parent.  Which is why I am asking to check it, since it looks wrong to me, and as seen above it leads to forgetting the routehint "early".

This is particularly important since completely unpublished payees can only be accessed via the routehints they put in the invoice, so forgetting routehints should be strongly discouraged.

With both my changes to the `routehint_data_init` function and the change to use root amount, the test passes.  Will push it now.

cdecker:
Hm, that is kind of bad tbh: without the `d->offset = pd->offset + 1` we will always only consider the first routehint, we will not even try the other ones, and worse we'll not try without routehints despite that being possible sometimes (if the initial `getroute` check tells us the destination is reachable directly).

We should probably add a signal to the `struct payment` to allow reusing a routehint in a child that came from the `waitblockheight` modifier, but that is the kind of cross-talk that may end up undoing some of the nice compartmentalization. 

To be honest I don't think the `waitblockheight` case is going to hit us too often, and I'd hate to be adding cross-modifier signaling just for this rare case. Let's see if we can somehow work around it.

ZmnSCPxj:
> we will not even try the other ones

But ***we do***!  See: https://github.com/ElementsProject/lightning/blob/723b7223b76e8bb5ab16f55018abd8a8833d4f9e/plugins/libplugin-pay.c#L1797-L1801

We only try the other ones if the current one is now banned by `routehint_excluded`, but IMO that is ***correct***: if the problem does not lie in the *current* routehint, why change the current routehint?

What we should do instead is what I suggest in #3894, i.e. give each sub-payment its own permutation of the order of routehints, so that each sub-payment tries a different routehint if possible.

cdecker:
Hm, you're right, we can rotate the routehints without assuming the last retry was due to the routehint being bad.

I like your idea of giving each split its own iteration order, but think that could also be achieved simply by offsetting the iteration randomly into the routehints list? Basically use `d->offset` to iterate through the array, and have a `d->base = RANDOM` be the starting point for that iteration.

Our current routehint then just becomes `d->routehints[(d->offset + d->base) % num_routehints]`. If instead of assigning a random value to `d->base` we can make just increment them on each split, spreading partial payments evenly across all routehints.

cdecker:
Nice find that we don't need to increment routehing-offset and retry counter in lock-step btw :-)

ZmnSCPxj:
hmm so something like `d->offset = 0` always, and `d->base = pseudorand(tal_count(d->routehints))`?

cdecker:
That sounds reasonable to me, it'd mean that we always use the first non-excluded routehint, starting from a random position in the array. `d->offset` is then just used to determine if we have more routehints we can try before we need to split.

ZmnSCPxj:
It is less optimal than making a strong effort to ensure that different sub-payments have as diverse ordering as possible, but is very simple to implement.

ZmnSCPxj:
Rebased to fix merge conflicts, also add changelogs.  Will go put the randomization we suggested, please wait.

ZmnSCPxj:
@cdecker there is a good reason for us to advance the `d->update`, and it is: https://travis-ci.org/github/ElementsProject/lightning/jobs/716299819#L1943

Basically, if my understanding of the code is right:

* Exceeding the budget needs to be signalled by setting `p->route = NULL`: the error message is "Fee exceeds our budget ... discarding route" even, but we do not, in fact, discard the route.
* `routehint_step_cb` needs to check for failure, and if it is because of an inability to route, should check routehints.  If there are more routehints beyond the current one, it should advance the routehint pointer `d->offset` and retry.

ZmnSCPxj:
The issue with the above is that when we exceed the constraints, we also exclude the most expensive channel.  And the reason for that is that we used to have an actual route randomization, and sometimes it would allow really expensive channels due to randomization.  So we need to factor in route randomization (once implemented) in our decision to advance or not advance the routehints pointer.  If the reason for route rejection is due to route randomization going overboard, we should consider whether to advance the routehints pointer, or reduce the randomization, but not both (we should advance the routehints pointer first, and if that runs out, reduce randomization and reset the routehints pointer).


@rustyrussell notice interaction with route randomization!

ZmnSCPxj:
Okay, went with "advance the routehint only if routing failed", and promote constraint violation to a "routing failed". @cdecker thoughts?  I think constraint violation should be consider a routing failure.

ZmnSCPxj:
Yet another reason to advance the routehint: https://travis-ci.org/github/ElementsProject/lightning/jobs/716418486#L2348

In this scenario there exists two routehints, in this order: l3->l4->l5, and l3->l5, l5 is the payee.  l4 happens to be down.

This is what happens in the payer:

```
b'2020-08-10T03:04:25.353Z DEBUG plugin-pay: Added a channel hint for 103x1x1/1: enabled true, estimated capacity 957757992msat'
b'2020-08-10T03:04:25.353Z INFO plugin-pay: Payment fee constraint 500msat is below exemption threshold, allowing a maximum fee of 5000msat'
b"2020-08-10T03:04:25.355Z DEBUG plugin-pay: After filtering routehints we're left with 2 usable hints"
b'2020-08-10T03:04:25.355Z DEBUG plugin-pay: Asking gossipd whether 032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e is reachable without routehints.'
b'2020-08-10T03:04:25.357Z DEBUG plugin-pay: Using routehint 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d (110x1x1) cltv_delta=6'
b'2020-08-10T03:04:25.357Z DEBUG plugin-pay: The destination is not directly reachable excluding attempts without routehints'
b'2020-08-10T03:04:25.671Z DEBUG plugin-pay: Added a channel hint for 110x1x1/0: enabled true, estimated capacity 101001msat'
b'2020-08-10T03:04:25.671Z INFO plugin-pay: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (reply from remote)'
b'2020-08-10T03:04:25.671Z DEBUG plugin-pay: Retrying d7b7f51de4b0c08aa1276b174f6b154612605d0f3337bfa489c2f7e2b3c47f8c/1 (100000msat), new partid 2. 9 attempts left\\n'
b'2020-08-10T03:04:25.672Z DEBUG plugin-pay: Using routehint 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d (110x1x1) cltv_delta=6'
b'2020-08-10T03:04:25.997Z INFO plugin-pay: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (reply from remote)'
b'2020-08-10T03:04:25.997Z DEBUG plugin-pay: Retrying d7b7f51de4b0c08aa1276b174f6b154612605d0f3337bfa489c2f7e2b3c47f8c/2 (100000msat), new partid 3. 8 attempts left\\n'
b'2020-08-10T03:04:25.998Z DEBUG plugin-pay: Using routehint 035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d (119x1x0) cltv_delta=6'
b'2020-08-10T03:04:26.660Z INFO plugin-pay: Killing plugin: Plugin exited before completing handshake.'
```

So on the first `WIRE_TEMPORARY_CHANNEL_FAILURE` we added a channel hint for `110x1x1/0`.  However, the routehint paymod does *not* consider the channel hints in `next_routehint`.  This is because `routehint_excluded` does not consider channel hints.  The correct solution here is to have `routehint_excluded` check channel hints.

ZmnSCPxj:
Fixed the above.  Hopefully this is the final added fix.

ZmnSCPxj:
YAY IT PASSES

@cdecker re: the routehint randomization, I will put in #3913 instead, as that adds multi-routehint invoices.  I will rebase 3913 on top of this one, I think the `flaky` in 3913 would be resolved by the changes here.

ZmnSCPxj:
I now realized that the `waitblockheight` condition would often cause the paymod system to give up too early due to setting the `abort` flag in `handle_final_failure`.

ZmnSCPxj:
Rebased.  Also, ***bump***

cdecker:
Rebased on top of `master`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3914/commits/ce05d3f6ba23fbaab4a1d764f78ff81ea8a3794a

PR #3917, opened 2020-08-10
plugins/libplugin-pay.c: Give cmd id and partid for each log message.

ZmnSCPxj:
Changelog-None: internal debugging

Makes it easier to debug payments with tons of splits.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -13,6 +13,9 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 			    struct payment_modifier **mods)
 {
 	struct payment *p = tal(ctx, struct payment);
+
+	static u64 next_id = 0;
+
 	p->children = tal_arr(p, struct payment *, 0);
 	p->parent = parent;
 	p->modifiers = mods;
@@ -46,6 +49,7 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 		p->deadline = parent->deadline;
 
 		p->invoice = parent->invoice;
+		p->id = parent->id;
 	} else {
 		assert(cmd != NULL);
 		p->partid = 0;
@@ -53,6 +57,7 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 		p->plugin = cmd->plugin;
 		p->channel_hints = tal_arr(p, struct channel_hint, 0);
 		p->excluded_nodes = tal_arr(p, struct node_id, 0);
+		p->id = next_id++;
 	}
 
 	/* Initialize all modifier data so we can point to the fields when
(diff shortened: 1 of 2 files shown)


ZmnSCPxj:
Also make the splitting operations actually print what their children are.

plugins/libplugin-pay.c
```diff
@@ -78,6 +78,41 @@ struct payment *payment_root(struct payment *p)
...
+
+	va_start(ap, fmt);
+	txt = tal_vfmt(tmpctx, fmt, ap);
+	va_end(ap);
+
+	plugin_log(p->plugin, l,
+		   "id %"PRIu64" partid %"PRIu32": %s",
+		   (root->cmd && root->cmd->id) ? *root->cmd->id : UINT64_MAX,
+		   p->partid,
+		   txt);
```
cdecker, line 96: Printing `UINT64_MAX` seems very hacky to me. Could we add a `(null)` to indicate that this was not initiated by `cmd` or have a global counter of payments and just increment that for each `payment_new` we call?

ZmnSCPxj: The `id` is supposed to identify the command that initiated the call, so that e.g. if we have two commands in parallel, we do no get confused which log belongs to which command.  Using the `cmd` id is better since it correlates with a command coming from `lightningd`, though of course `lightningd` never prints that.

ZmnSCPxj: Modified to print `cmd N` if there is a command with an id, fall back to `id N` and a generated ID if not exist.

ZmnSCPxj:
Rebased, also printed some more details I would have really liked to see in the logs from @thestick613 .

plugins/libplugin-pay.c
```diff
@@ -8,6 +8,8 @@
 
 #define DEFAULT_FINAL_CLTV_DELTA 9
 
+static u64 id_alloc = 0;
+
```
cdecker, line 12: FWIW this could also be done as a `static u64 next_id = 0` in the `payment_new` function. It's a static variable, initialized to 0 only the first time, and can be incremented from there. That'd avoid having a global variable here, and it'd be clearer where it is being used.

plugins/libplugin-pay.c
```diff
@@ -79,6 +83,52 @@ struct payment *payment_root(struct payment *p)
...
+	 * We only fall back to a unique id per root payment if there
+	 * is no command with an id associated with this payment.
+	 */
+	if (root->cmd && root->cmd->id)
+		return tal_fmt(ctx, "cmd %"PRIu64" partid %"PRIu32,
+			       *root->cmd->id, p->partid);
+	else
+		return tal_fmt(ctx, "id %"PRIu64" partid %"PRIu32,
+			       root->id, p->partid);
+}
```
cdecker, line 103: This results in an extra allocation for the string, where it might be preferable to integrate this into the `paymod_log` logic, which is still rather small.

cdecker: We could have a simple tuple returned with the components:

```
struct paymod_log_meta {
    const char *type;
    const u64 payment_id;
    const u32 part_id;
}
```

And then pass that to the `plugin_log` function. If we return this by value rather than by reference it should get allocated on the caller stack (`paymod_log` or `paymod_err`) and we save the extra cost of allocating.

plugins/libplugin-pay.c
```diff
@@ -251,16 +251,38 @@ static void channel_hints_update(struct payment *p,
...
+
+			paymod_log(p, LOG_DBG,
+				   "Updated a channel hint for %s: "
+				   "enabled %s, estimated capacity %s",
+				   type_to_string(tmpctx,
+						  struct short_channel_id_dir,
+						  &hint->scid),
+				   hint->enabled ? "true" : "false",
+				   type_to_string(tmpctx, struct amount_msat,
+						  &hint->estimated_capacity));
```
cdecker, line 285: imho it's preferable to have the non-negated `if` surround the log, rather than having an early exit in the form of that `return`

plugins/libplugin-pay.c
```diff
@@ -1254,12 +1278,21 @@ static void payment_compute_onion_payloads(struct payment *p)
...
 	/* Final hop */
 	payment_add_hop_onion_payload(
 	    p, &cr->hops[hopcount - 1], &p->route[hopcount - 1],
 	    &p->route[hopcount - 1], true, root->payment_secret);
+	tal_append_fmt(&routetxt, "%s",
+		       type_to_string(tmpctx, struct short_channel_id,
+				      &p->route[hopcount - 1].channel_id));
+
+	paymod_log(p, LOG_DBG,
+		   "Created outgoing onion for route: %s", routetxt);
```
cdecker, line 1295: Nice change, this should help alot with debugging (and is way easier to grok than the `getroute` output). We might even want to keep this short format around attached to the `struct payment` to add context to later logs.

ZmnSCPxj:
This seems to have stumbled on a latent bug, where `assign_blame` is somehow returning a node address to uninitialized data.

ZmnSCPxj:
Found it.  We are not propagating `p->local_id` from the parent!!

ZmnSCPxj:
This used to be benign since we only ever printed the erring node in case of a strange failure code, and we would never emit a strange failure code from our local node.  However, since we now print the erring node in all failure codes, we stumbled into this latent bug.  Added checks that `local_id` is initialized.  This is not an existing remote-triggerable crash, since only `local_id` is crashing and we would never emit strange failure codes ourselves.

cdecker:
Excellent, thanks for the fixups @ZmnSCPxj :+1: (I reviewed the [`range-diff` between 344d9  and bbc54](http://46.101.246.115:8888/ElementsProject/lightning/pull/3917/range_diff/344d9d9fd8709d16316707a5ad7db6da8f3cbe63..bbc54dcd3e9ccf8849b7f5290ec0f8d906f44b78)

ACK https://github.com/ElementsProject/lightning/pull/3917/commits/bbc54dcd3e9ccf8849b7f5290ec0f8d906f44b78

PR #3924, opened 2020-08-11
Paymod micro optimization

ZmnSCPxj:
Some pointless micro-optimizations which you are never going to feel anyway.

`payment_root` is actually optimized very well by GCC 9 in `-O2`, it gets transformed into a `while` loop as a tail-recursion optimization, and is therefore not touched by this PR.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -145,10 +145,11 @@ struct payment_tree_result payment_collect_result(struct payment *p)
 	return res;
 }
 
-static struct command_result *payment_getinfo_success(struct command *cmd,
-						      const char *buffer,
-						      const jsmntok_t *toks,
-						      struct payment *p)
+static struct command_result *
+payment_getblockheight_success(struct command *cmd,
+			       const char *buffer,
+			       const jsmntok_t *toks,
+			       struct payment *p)
 {
 	const jsmntok_t *blockheighttok =
 	    json_get_member(buffer, toks, "blockheight");
@@ -157,7 +158,10 @@ static struct command_result *payment_getinfo_success(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
-void payment_start(struct payment *p)
+#define INVALID_BLOCKHEIGHT UINT32_MAX
+
+static
+void payment_start_at_blockheight(struct payment *p, u32 blockheight)
 {
 	struct payment *root = payment_root(p);
 	p->step = PAYMENT_STEP_INITIALIZED;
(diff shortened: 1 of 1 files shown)


cdecker:
I'm at a bit of a loss here: I usually really like optimizations, however this hardly seems worth it. The first commit is borderline but still ok, though I hardly think it's necessary to optimize around call-stack operations when we have RPC calls over stdin / stdout in the same code (syscalls >> call-stack).

The second commit I don't get why add the very specific `payment_start_at_blockheight` when we can just reach into the `payment` and set the blockheight from the waitblockheight callback (we can do that even after calling `payment_start` since that just queues operations to be executed once the parent payment yields control).

Would be interesting to see what others think, does code cleanliness trump performance when the performance gain is small enough?

@ZmnSCPxj do you have any measurements as to how the performance changes before and after applying these optimizations?

ZmnSCPxj:
> @ZmnSCPxj do you have any measurements as to how the performance changes before and after applying these optimizations?

None at all.

ZmnSCPxj:
>  though I hardly think it's necessary to optimize around call-stack operations when we have RPC calls over stdin / stdout in the same code (syscalls >> call-stack).

The motivation here was less call-stack operations but rather call-stack size.

Assume our largest payment size is 2^31 , to be below the wumbo limit, and assuming we throttle the presplit, which seems necessary, so reaching 2^31 for a really large payment is possible.  In order to get payments below the 100,000 limit where adaptive splitter gives up, we have to reach down to 2^16 = 65,536.  So that's 16 splits.  Each split can have up to 10 retries from the retry mod, so that's up to 160 calls deep in the worst case.

Each stack frame requires at least a code pointer to return to, on a 64-bit system that's 8 bytes.  We have a `size_t i`, which cannot be saved in callee-saved register (because the recursion means the callee also needs to create a `size_t i`), so that is another 8 bytes. The same argument exists for the `struct payment *p` argument: sure we can pass it in a register but when you recurse, you have to save your `struct payment *p` on the stack before you replace the value of that register with the specific child.  So it is 24 bytes per frame, times 160 frames depth:  3,840 bytes.

All right, that is not a very large number, not even an entire vmem page, but the optimization potentially reduces this size by a factor of 10 (as it short-circuits the up-to-10 retries of the retry paymod), so I thought it would be worth it for the memory use.

cdecker:
Paging Dr. @rustyrussell: is this really a concern?

ZmnSCPxj:
If your main concern with the first commit is just the `goto`, then the below will be compiled the same when using gcc at `-O2`, because that is the level at which tail-recursion optimization is enabled for GCC:

```c
        /* Not bother with looping if just one child. */
        if (num_children == 1)
                return payment_finished(p->children[0]);
```

But since C does not require tail-recursion optimization or tail-call optimization, unlike Scheme, this is not reliable at lower optimization levels or for other C compilers.

cdecker:
Honestly I care more about readability rather than optimizations, but I think we're good in this case. The `goto` tail recursion for the single-child case just requires looking twice.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3924/commits/e6f2d1a23e31bf42db38f6999d033817da74f262

PR #3927, opened 2020-08-11
Fix aggregation of partial payments in the pay-plugin

### LINKED ISSUE #3915: crashed lightningd
Attempted an MPP payment, which stopped responding after 5 minutes, so i Ctrl+C the `pay` command.
I then attempted to close a channel, which outputed `cannot co-op close channel w/ active htlcs` in the log.
Then i ran `listpays` to see the status of the MPP payments, which resulted in this crash:

```
pay: FATAL SIGNAL 11 (version v0.9.0-44-g01a82d3)
0x4183e7 send_backtrace
        common/daemon.c:38
0x418471 crashdump
        common/daemon.c:51
0x7f70415144af ???
        ???:0
0x7f7041568d26 ???
        ???:0
0x403a68 attempt_ongoing
        plugins/pay.c:1638
0x403e8b listsendpays_done
        plugins/pay.c:1820
0x4092ba handle_rpc_reply
        plugins/libplugin.c:551
0x40938f rpc_read_response_one
        plugins/libplugin.c:666
0x409417 rpc_conn_read_response
        plugins/libplugin.c:685
0x42729a next_plan
        ccan/ccan/io/io.c:59
0x42775a do_plan
        ccan/ccan/io/io.c:407
0x427785 io_ready
        ccan/ccan/io/io.c:417
0x428f20 io_loop
        ccan/ccan/io/poll.c:445
0x409ee2 plugin_main
        plugins/libplugin.c:1303
0x407188 main
        plugins/pay.c:2063
0x7f70414ff82f ???
        ???:0
0x402e08 ???
        ???:0
0xffffffffffffffff ???
        ???:0
2020-08-09T11:32:04.220Z INFO plugin-pay: Killing plugin: Plugin exited before completing handshake.
2020-08-09T11:32:04.221Z **BROKEN** plugin-pay: Plugin marked as important, shutting down lightningd!
2020-08-09T11:32:04.271Z INFO plugin-autoclean: Killing plugin: Plugin exited before completing handshake.
2020-08-09T11:32:04.273Z INFO plugin-bcli: Killing plugin: Plugin exited before completing handshake.
2020-08-09T11:32:04.273Z INFO plugin-fundchannel: Killing plugin: Plugin exited before completing handshake.
2020-08-09T11:32:04.274Z INFO plugin-keysend: Killing plugin: Plugin exited before completing handshake.
2020-08-09T11:32:07.170Z INFO plugin-bcli: bitcoin-cli initialized and connected to bitcoind.

```

### LINKED ISSUE #3931: Unknown command 'listpays'
Sometimes (this happened a few times now) and maybe when interacting with `spark-wallet`, `c-lightning` loses `listpays` and responds:

```
{
   "code": -32601,
   "message": "Unknown command 'listpays'"
}

```
After a restart it knows about `listpays` again :-)

### `getinfo` output

version: "v0.9.0-2-ge7d89cd"

cdecker:
This fixes #3915, which could crash if the presplit modifier skipped the root payment, which was also the only payment annotated with the bolt11 string. It is addressed in two ways: we now annotate the first level of children in presplit (all other splits or skips act after the attempt was sent to `lightnignd` and thus there is at least one partial payment with the bolt11 annotation), and secondly we now aggregate via the `payment_hash` which is guaranteed to be non-null for all current payments, whereas the bolt11 string is only non-null if the payment was annotated with it.

Fixes #3915 
Fixes #3931

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -2661,6 +2661,11 @@ static void presplit_cb(struct presplit_mod_data *d, struct payment *p)
 			struct payment *c =
 			    payment_new(p, NULL, p, p->modifiers);
 
+			/* Annotate the subpayments with the bolt11 string,
+			 * they'll be used when aggregating the payments
+			 * again. */
+			c->bolt11 = tal_strdup(c, p->bolt11);
+
 			/* Get ~ target, but don't exceed amt */
 			c->amount = fuzzed_near(target, amt);
 
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -36,6 +36,7 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 		tal_arr_expand(&parent->children, p);
 		p->destination = parent->destination;
 		p->amount = parent->amount;
+		p->label = parent->label;
 		p->payment_hash = parent->payment_hash;
 		p->partid = payment_root(p->parent)->next_partid++;
 		p->plugin = parent->plugin;
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -5,6 +5,7 @@
 #include <ccan/htable/htable_type.h>
 #include <ccan/intmap/intmap.h>
 #include <ccan/json_out/json_out.h>
+#include <ccan/str/hex/hex.h>
 #include <ccan/tal/str/str.h>
 #include <common/amount.h>
 #include <common/bolt11.h>
@@ -80,6 +81,8 @@ struct pay_status {
 
 	/* Array of payment attempts. */
 	struct pay_attempt *attempts;
+
+	struct sha256 payment_hash;
 };
 
 struct pay_command {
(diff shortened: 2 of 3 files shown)


plugins/libplugin-pay.c
```diff
@@ -36,6 +36,7 @@ struct payment *payment_new(tal_t *ctx, struct command *cmd,
 		tal_arr_expand(&parent->children, p);
 		p->destination = parent->destination;
 		p->amount = parent->amount;
+		p->label = parent->label;
```
ZmnSCPxj, line 39: Should it not also initialize this if there is no parent?  Consider the case where a new plugin does payments in the background without any commands or a concept of labels.

cdecker: Where would we get that label from? It's passed in from the JSON-RPC as an annotation, so we allow the caller to optionally add it to the root from where it'll get inherited, otherwise it defaults to `NULL`.

Issue #3926, opened 2020-08-11
MPP pay splits into too many subpayments


thestick613:
I have an outbound channel with an BTC to Lightning exchange worth 2.000.000 satoshi, with enough capacity.
I try to swap my 2.000.000 satoshi for the same amount, but on-chain.
Lightningd splits the payment into too many subpayments, and eventually fails, even though the destination is only one hop away, and has enough capacity.

ZmnSCPxj:
Can you provide logs?  In particular a message "Split into " in the inform-level logs (though debug-level logs would be best".

thestick613:
First time i tried this, i ended up with #3846, then with #3848 and then with #3851, and then with #3915. The logs always looked like this:
- Detected large payment, splitting into 100 sub-payments, and then a log of entries about too many HTLCs or timeouts.
I'm willing to do another test in a few days.

ZmnSCPxj:
I observe something similar in the failure cases of #3913 as well, which I think are mitigated by #3914.  Are you paying to a node that is unpublished (i.e. is not visible on 1ml.com)?

ZmnSCPxj:
The reason the logs are important is that we have to know if it is the presplit splitter that is too aggressive, or the adaptive splitter.

ZmnSCPxj:
Finally: if you do not want to try paying again for a while due to the extant bugs: how many channels do you have?

ZmnSCPxj:
And if possible: if the destination is published, how many channels does it have?  If the destination is not published, how many routehints were on the invoice?

@cdecker maybe we should consider as well the HTLC budget on the payee end, not just on the payer end?  In #3916 @thestick613 claims to have over 100 HTLCs pending, if those are from a single `pay` attempt that suggests the payer node has ~10 channels or so I think, perhaps it is the payee that is not able to handle that many incoming HTLCs due to having fewer channels?

thestick613:
Logs incoming:

```
2020-08-12T12:53:06.135Z INFO plugin-pay: Split into 51 sub-payments due to initial size (499907000msat > 10000000msat)
2020-08-12T12:53:06.687Z UNUSUAL plugin-pay: Could not update the channel hint for 605257x1187x0/1. Could be a concurrent `getroute` call.
2020-08-12T12:53:06.688Z UNUSUAL plugin-pay: Could not update the channel hint for 605257x1187x0/1. Could be a concurrent `getroute` call.
[...]

PR #3936, opened 2020-08-13
MPP: limit the number of HTLCs based on destination connectivity

ZmnSCPxj:
Fixes: #3926

(probably)

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1,6 +1,7 @@
 #include <bitcoin/preimage.h>
 #include <ccan/array_size/array_size.h>
 #include <ccan/tal/str/str.h>
+#include <common/json_helpers.h>
 #include <common/json_stream.h>
 #include <common/pseudorand.h>
 #include <common/random_select.h>
@@ -2925,9 +2926,6 @@ static u32 payment_max_htlcs(const struct payment *p)
 	return res;
 }
 
-/* Temporary comment out else GCC will complain that it is unused;
- * will be used in a later commit.  */
-#if 0
 /** payment_lower_max_htlcs
  *
  * @brief indicates that we have a good reason to believe that
@@ -2958,7 +2956,6 @@ static void payment_lower_max_htlcs(struct payment *p, u32 limit,
 		root->max_htlcs = limit;
 	}
 }
-#endif
 
 static bool payment_supports_mpp(struct payment *p)
 {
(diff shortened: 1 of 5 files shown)


ZmnSCPxj:
Rebased

ZmnSCPxj:
Currently wondering if I should instead put the new code in a new paymod instead of mixing it with the presplitter.

And additionally make the presplitter less aggressive, say if it would generate more HTLCs than half the `payment_max_htlcs` it would limit itself to half the `payment_max_htlcs`, given that we can very easily split later but have no way to re-merge payments.

ZmnSCPxj:
Turns out the routehints paymod takes over ownership of the invoice routes, but we need to know the number of invoice routes being maintained, because if the payee is unpublished that is the only indication we have of the incoming channels of the payee and therefore the limit on the destination connectivity.

whitslack:
This seems extremely fragile. You're making some wild assumptions about the HTLC capacities of the nodes in the network. You say, "Assume 30, since that is what C-lightning itself uses, and we have probably the lowest such limit among implementations in common use," but 30 HTLCs is a rather high limit. My C-Lightning node has automatically force-closed a channel with only 4 outstanding HTLCs, and it cost me over $5 in chain fees, once all was said and done, so I surely don't want to allow **30** outstanding HTLCs on any channel; that'd be a recipe for going broke. I have set `max-concurrent-htlcs=5` to limit my potential losses. If other node operators are adopting the same strategy, then your estimate of 30 HTLCs per channel doesn't hold water. Besides, you're assuming that no *other* multi-part payments are in progress to the destination node concurrently, but that is unlikely to be true for high-volume destinations.

Shouldn't C-Lightning strive to *minimize* the number of concurrent HTLCs that any given multi-part payment generates? Each additional HTLC costs additional money (base fee) and carries a non-zero (and rather high) probability of triggering a channel force-closure and costing the user many dollars in chain fees. Your current strategy is just too expensive.

ZmnSCPxj:
30 here is the *maximum* we will split *later*.  See `ASSUMED_REASONABLE_HTLCS_PER_CHANNEL` instead for the limit on the *presplit*.  This is the initial limit on the number of splits, and comes to 30 / 3 = 10.  We do the presplit for up to 10 splits per channel, which gives us some slack later for adaptive splitting.

ZmnSCPxj:
Moved into a new paymod, since @whitslack brings up that maybe we should not really split early, so maybe we might decide to remove presplitter later.  Also, toned down the `ASSUMED_MAX_HTLCS_PER_CHANNEL`.

plugins/libplugin-pay.c
```diff
@@ -2671,16 +2672,61 @@ static struct presplit_mod_data *presplit_mod_data_init(struct payment *p)
...
 	struct channel_hint *h;
 	u32 res = 0;
 	for (size_t i = 0; i < tal_count(p->channel_hints); i++) {
 		h = &p->channel_hints[i];
 		if (h->local && h->enabled)
 			res += h->htlc_budget;
 	}
+	root = p;
+	while (root->parent)
+		root = root->parent;
```
cdecker, line 2686:
```suggestion
```

plugins/libplugin-pay.c
```diff
@@ -2671,16 +2672,61 @@ static struct presplit_mod_data *presplit_mod_data_init(struct payment *p)
 
 static u32 payment_max_htlcs(const struct payment *p)
 {
+	const struct payment *root;
```
cdecker, line 2676:
```suggestion
	const struct payment *root = payment_root(p);
```

plugins/libplugin-pay.c
```diff
@@ -2994,3 +2991,111 @@ static void adaptive_splitter_cb(struct adaptive_split_mod_data *d, struct payme
...
+	json_for_each_arr (i, chantok, channelstok) {
+		/* We might get up to 2 channel objects, one for each
+		 * direction.  */
+		sourcetok = json_get_member(buf, chantok, "source");
+		if (!sourcetok)
+			continue;
+		if (!json_to_node_id(buf, sourcetok, &source))
+			continue;
+		if (!node_id_eq(&source, p->destination))
+			continue;
```
cdecker, line 3054: No, `source=mynodeid` will only return channels for which `mynodeid` is the source, the other direction would have it as `destination` and be filtered out. Using the `node_id` as source as a proxy for the channels whose destination the `node_id` is should be ok.

cdecker (approved):
This is rather less convoluted than what I expected to be honest, the paymod structure seems to be working :-)
I was about to argue that it'd be better if we limit the overall number of HTLCs the presplit modifier is allowed to create (using either the fibonacci series in #3951 or the simpler power of 16 proposal that doesn't have code yet), but since this is a rather nicely self-contained limiter I think we should keep it.

Might come in handy to have another early abort signal.

plugins/libplugin-pay.c
ZmnSCPxj: `payment_root` accepts and returns non-`const` pointers to `struct payment`, which is why I open-coded this here.

Since `payment_max_htlcs` does not modify the payment it makes sense for it to accept a non-`const` pointer.  I notice that I am probably the only one who consistently considers `const`-correctness.

As currently only `payment_max_htlcs` uses `const struct payment *`, it seems a bit wasteful to factor out a new `cpayment_root` that accepts and returns `const struct payment *`.

ZmnSCPxj:
> I was about to argue that it'd be better if we limit the overall number of HTLCs the presplit modifier is allowed to create

I would argue that this is different from the presplit:

* The presplit limit is for the *expected network connectivity*, i.e. the average connectivity of hop nodes.
* This limit is for the *destination connectivity*.

We already have a limit for the source connectivity, that is the only limit we have.

We should limit to the lowest of all three limits.

The source or the destination could be unusually low-connectivity compared to the network, which i why the presplit limit (which represents the expected limit of the network) is different from the destination limit (which we derive from the number of channels we know to the destination).

plugins/libplugin-pay.c
cdecker: > `payment_root` accepts and returns non-`const` pointers to `struct payment`, which is why I open-coded this here.

`payment_root` should accept non-`const`, that's my bad.

> it seems a bit wasteful to factor out a new `cpayment_root` that accepts and returns `const struct payment *`.

Definitely don't do this. I lieu of a better solution we can keep it as it is.

plugins/libplugin-pay.c
```diff
@@ -2637,16 +2673,57 @@ static struct presplit_mod_data *presplit_mod_data_init(struct payment *p)
...
+static void payment_lower_max_htlcs(struct payment *p, u32 limit,
+				    const char *why)
+{
+	struct payment *root = payment_root(p);
+	if (!root->max_htlcs) {
+		root->max_htlcs = tal(root, u32);
+		*root->max_htlcs = limit;
+		plugin_log(p->plugin, LOG_INFORM,
+			   "Initial limit on max HTLCs: %"PRIu32", %s",
+			   limit, why);
```
cdecker, line 2718: Just make this a `u32` (not a `u32*`) and initialize it to `UINT32_MAX`, that still maintains the correct semantics without extra allocations.

cdecker: At which point this boils down to `min()` + `plugin_log` and we can just do that in the caller, saving us the indirection of constructing a `why` string and then combine that into a `plugin_log` call.

ZmnSCPxj: I think converting to `u32` and initializing to `UINT32_MAX` is a good idea, but I disagree with manually inlining the `payment_lower_max_htlcs` function into the sole caller.

The intent is that `payment_lower_max_htlcs` and `payment_max_htlcs` are common, shared parts of the code, that *any* paymod can call.

If we were to manually inline `payment_lower_max_htlcs`, it would imply that the `max_htlcs` field is owned by the `payee_incoming_limit` paymod.  But the intent is that the max-htlcs field is a *shared* field, with a decent monoidal combination (i.e. `min` is a monoid, its zero is `UINT32_MAX`) of multiple paymods poking at it, if in the future we discover even more reasons to limit the number of HTLCs we should emit (and given our recent discoveries, we should probably prepare for other reasons to limit the number of HTLCs).  Thus, rather than hard-inline the min operation, it should be encapsulated away.

Smart compilers will inline anyway.

ZmnSCPxj:
Rebased (especially with `paymod_log`), converted `max_htlcs` to `u32`.  I still think separating the code out is good encapsulation, even if it currently only has one caller.  The `why` string is required as well, as this is the justification we have to limit the number of HTLCs we will try.

cdecker:
Rebased on top of `master`. LGTM :+1:

ACK https://github.com/ElementsProject/lightning/pull/3936/commits/db979fe228c8cd74749e7159018c413e255b8d13

rustyrussell:
Rebased onto  #4035

cdecker:
ACK 08a22b73e948feaa864326ab318431bdc811ae17

PR #3941, opened 2020-08-13
Add a new param to listsendpays and listpays to only return last n entries

fiatjaf:
The amount of things returned from these commands can grow so much that it becomes very slow to get a response. Specially now that `listpays` is doing heavy computation for each `listsendpays` entry it is very very slow.

This change would also benefit wallets like Spark that want to list only the latest payments in their UI.

This PR is just a naïve proof of concept, because I don't know what is the best way to achieve what I'm asking or if it is a desired change. If the idea is good it could also be done on `listinvoices`, `listforwards`.

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -786,7 +786,7 @@ send_payment_core(struct lightningd *ld,
 	bool have_complete = false;
 
 	/* Now, do we already have one or more payments? */
-	payments = wallet_payment_list(tmpctx, ld->wallet, rhash);
+	payments = wallet_payment_list(tmpctx, ld->wallet, rhash, NULL);
 	for (size_t i = 0; i < tal_count(payments); i++) {
 		log_debug(ld->log, "Payment %zu/%zu: %s %s",
 			  i, tal_count(payments),
@@ -1450,10 +1450,12 @@ static struct command_result *json_listsendpays(struct command *cmd,
 	struct json_stream *response;
 	struct sha256 *rhash;
 	const char *b11str;
+	u32 *limit_reverse;
 
 	if (!param(cmd, buffer, params,
 		   p_opt("bolt11", param_string, &b11str),
 		   p_opt("payment_hash", param_sha256, &rhash),
+		   p_opt("limit_reverse", param_number, &limit_reverse),
 		   NULL))
 		return command_param_failed();
 
@@ -1475,7 +1477,7 @@ static struct command_result *json_listsendpays(struct command *cmd,
 		rhash = &b11->payment_hash;
 	}
 
-	payments = wallet_payment_list(cmd, cmd->ld->wallet, rhash);
+	payments = wallet_payment_list(cmd, cmd->ld->wallet, rhash, limit_reverse);
 
 	response = json_stream_success(cmd);
 
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1876,11 +1876,13 @@ static struct command_result *json_listpays(struct command *cmd,
 	const char *b11str;
 	struct sha256 *payment_hash;
 	struct out_req *req;
+	u32 *limit_reverse;
 
 	/* FIXME: would be nice to parse as a bolt11 so check worked in future */
 	if (!param(cmd, buf, params,
 		   p_opt("bolt11", param_string, &b11str),
 		   p_opt("payment_hash", param_sha256, &payment_hash),
+		   p_opt("limit_reverse", param_number, &limit_reverse),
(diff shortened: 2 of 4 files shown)


vincenzopalazzo:
I want to add my opinion, maybe is not relevant and maybe it is wrong.

If I understand correctly the code in this PR, you want to limit the result of the query to the last `x` payments. I agree with the idea because with the MPP we can have a lot of payment as a result and the DB must load a lot of payment that the user don't want in some cases.

The only thing that I want to ask if the user doesn't know how many payments the command pay divided the single payment. In another world, how I can know how many payments I should require to have the complete list of "sub payments"?

In my knowledge at the moment, there isn't a method to ask at lightningd, how many payments the command pays had created, the only method to know that is to makes a query to the db and run `SELECT COUNT(*) FROM payment WHERE payment_hash = ? ` but this is equal to make a query by `payment_hash` or `bolt11`, right?

In addition, another thing that I asked to myself is, if I don't know into haw many "sub-payments" created the command pay, I can make a query with a random number and have a partial result from the command `listsendpays`?

fiatjaf:
Yes, that was a concern by me too. That is one reason why I said this PR was naïve.

One possible solution would be to, if the maximum number of shards of an MPP payment are fixed, to query always `specified_limit + max_shard_count` and filter from that. But that seems messy.

Another possible solution is to do this filtering at the SQL level, although it can get ugly.

fiatjaf:
Immediately after writing the previous comment I came up with the best solution: to query by timestamp. A query by timestamp will most certainly get all clusters of MPP shards, and it will not cause `listpays` to return the wrong number of results.

(Since `listpays` aggregates the MPP entries returned from `listsendpays` into a single payment, this PR currently will cause it to return always less than the limit specified in the parameter.)

rustyrussell:
Hmm, how slow is `listpays` for you?  It's pretty efficient AFAICT.  I created 1000 invoices here and:

```
num sendpays = 1000
listsendpays time = 0.0570063591003418
listpays time = 0.1589813232421875
```

I can do some speedups, but how large should I be looking?

---

OK, this refreshes an older conversation about how our APIs should do pagination.  Here's my proposal:

1. Every list* command output object has one (or more!) order fields.  These monotonically increase.
2. Every list command takes parameters "start_order" and "max_num".
3. There's a new `wait` JSON command which takes "listname" "orderfield" ["prevorder"] ["nonblock"] which waits until there's an listname.orderfield >= orderfield, and returns the maximum orderfield (with "nonblock" it just returns the maximum orderfield immediately).

This should let you poll on every list command, and also do caching, catchup and paging reliably.  We would not support it on all list commands at once (in particular, listchannels and listnodes are harder).

Examples:
* listsendpays.order_created - set once when a sendpay order is created, never changed.
* listsendpays.order_ended - set once it's succeeded/failed.
* listinvoices.order_created - set once when an invoice is created, never changed.
* listinvoices.order_ended - set once it's succeeded/failed.

These replace waitsendpays and waitinvoice/waitanyinvoice.   We can refine it by allowing filtering args to 'wait', depending on the order field.

fiatjaf:
`listsendpays` is fine:

```
~> lightning-cli listsendpays | jq '.payments | length'
23627
~> lightning-cli listsendpays > /dev/null
0.28user 0.04system 0:01.16elapsed 28%CPU (0avgtext+0avgdata 30068maxresident)k
0inputs+0outputs (0major+7184minor)pagefaults 0swaps
```

but `listpays`...

```
~> lightning-cli listpays | jq '.pays | length'
22671
~> time lightning-cli listpays > /dev/null
0.19user 0.03system 0:17.58elapsed 1%CPU (0avgtext+0avgdata 24988maxresident)k
0inputs+0outputs (0major+5948minor)pagefaults 0swaps
```

rustyrussell:
OK, I have 50,000 payments in my test setup now, and I'm fixing the slowdown.  Still debugging, maybe in 24 hours.

rustyrussell:
OK, what's your number on master now?

fiatjaf:
```
~> time lightning-cli listsendpays > /dev/null
0.30user 0.03system 0:00.82elapsed 40%CPU (0avgtext+0avgdata 30328maxresident)k
0inputs+0outputs (0major+7283minor)pagefaults 0swaps
~> time lightning-cli listpays > /dev/null
0.17user 0.01system 0:01.43elapsed 12%CPU (0avgtext+0avgdata 25216maxresident)k
0inputs+0outputs (0major+5987minor)pagefaults 0swaps
```

Amazing.

PR #3942, opened 2020-08-14
plugins/libplugin-pay.c: Make presplitter less aggressive in general.

ZmnSCPxj:
Changelog-Changed: pay: Limit presplitting to a smaller number of HTLCs.

EDIT:

Closes #3951

I prefer #3951 over this one.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -2623,6 +2623,24 @@ REGISTER_PAYMENT_MODIFIER(waitblockheight, void *, NULL, waitblockheight_cb);
 #define MPP_TARGET_SIZE (10 * 1000 * 1000)
 #define PRESPLIT_MAX_HTLC_SHARE 3
 
+/* Intermediate network nodes have some limit on the number of HTLCs
+ * they can forward.
+ * Payments which get presplit into several dozen payments are likely
+ * to just flood the network and ultimately fail due to resource
+ * exhaustion.
+ * Even if the payee had the theoretical capacity to accept all those
+ * HTLCs, intermediate hops might not be able to do so, and the public
+ * network is a shared resource as well with other payers.
+ * So we need to limit at least the initial presplit to a number of
+ * sub-payments at the start; the adaptive splitter can always split it
+ * further later on.
+ * Thus, to get this number, we need to gather statistics on the number
+ * of channels a typical hop node has, and gather information from users
+ * on their max-concurrent-htlcs values; the number below is thus derived
+ * ex asino.
+ */
+#define REASONABLE_MAX_PRESPLIT 20
+
 static struct presplit_mod_data *presplit_mod_data_init(struct payment *p)
 {
 	struct presplit_mod_data *d;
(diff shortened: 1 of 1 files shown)


cdecker:
Looking at the number of reports of people sening 1'000'000+ satoshis I'm coming around to the idea of changing the presplitter. I still think that there is value in clamping to common shared values, instead of outright clamping to a constant that'd then allow reversing the split.

What do you think about splitting to `2^p * 10,000` where `p` is the smallest power for which the result is above the total amount of the payment and `p % 4 == 0`. Seems weird I know, but this would result in at most 16 parts, and all payments in the range of `[2^(p-4) * 10,000; 2^p * 10,000]` would result in the same amount side per part better hiding the initial amount.

The added advantage is that with the adaptive splitter splitting the amount in half, the post-adaptive splitter amount would then re-align with a lower class of presplits as well.

I think ranges of 1x - 16x are large enough to provide some obfuscation, while not overwhelming channels like our current simple presplitter does. I'm not sure whether keeping the 10,000 coefficient is a good choice or whether we should go powers of two all the way in order to have the presplitter sizes mix with the adaptive splitter sizes, but we can certainly discuss that :-)

ZmnSCPxj:
I would suggest that powers of two would work better.  Maybe the lowest bin could be 8,388,608 msat (2^23) and the next higher bin is 2^27 then 2^31 .  This mixes well with the 2^32 wumbo limit we still currently have, and 8,388,608 is just slightly smaller than your initial 10,000,000 msats bin.

ZmnSCPxj:
Ultimately, the  cost of each HTLC is determined ultimately by onchain fees.  Even `max-concurrent-htlcs` is just a proxy of the cost of each HTLC, as noted by @whitslack in https://github.com/ElementsProject/lightning/pull/3936#issuecomment-673654383 .  So it might do us better to dynamically select a higher starting bin size (fewer splits) if onchain fees are currently high.  Some of the logs from @thestick613 show that some of the errors are not due to `max-concurrent-htlcs` but because of the HTLC addition fee, so any assumptions about `max-concurrent-htlcs` may be invalidated in a high-fee climate.

Another thing to consider would be to do splits in a fibonacci sequence rather than powers of two.  So you might want to initially split by some fibonacci index, and adaptive splitter splits by fibonacci.  This might work better with the expected distribution that there will be some number of channels with high capacity while a majority are low-capacity.  Unless I am now betraying the fact that I am completely not a mathist and spouting nonsense at this point.

whitslack:
> So it might do us better to dynamically select a higher starting bin size (fewer splits) if onchain fees are currently high.

If you do that, then you also need to dynamically scale (down) `max-concurrent-htlcs` if the HTLC fee rate is high. Otherwise, I'm just going to continue setting the minimum allowed `max-concurrent-htlcs` to try to minimize my monetary losses due to malicious/buggy nodes that stop responding with HTLCs in flight.

whitslack:
Seems like it might be more useful to cap the amount of fees that must be paid to unilaterally close a channel rather than capping the number of HTLCs that may concurrently be in flight through each channel.

ZmnSCPxj:
> Seems like it might be more useful to cap the amount of fees that must be paid to unilaterally close a channel rather than capping the number of HTLCs that may concurrently be in flight through each channel.

True, though seems a bit OT here in `pay`-land.  An issue is what happens if there already are HTLCs in-flight, and then a fee spike occurs?  We cannot simply drop an HTLC, so we would just have to prevent future HTLCs but we cannot remove current ones.  Thus it will be at best a soft limit (which might be acceptable).  Further, we can easily set a number-of-HTLCs limit to the peer (because it is already existant in the spec), but adding a limit to the fees would require a spec change; if we are at the limit already and the peer wants to add a new HTLCs, the best we can do is to accept it (thus temporarily going over the limit) and either immediately succeed it (if we are the payee AND it completes an MPP or is a single-part payment)  or just fail it ASAP.

PR #3947, opened 2020-08-15
Implement a factory for paymod-based payments

ZmnSCPxj:
This gives us the nice capability to disable specific paymods at the command line for testing and experimentation.  For example, instead of a `use_shadow` parameter guarded by `DEVELOPER=1`, which requires our test harness to provide a `dev_pay` RPC, we can just `--dev-disable-paymods=shadowroute` to disable shadowroutes completely.

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -512,6 +512,36 @@ static char *opt_force_channel_secrets(const char *optarg,
 	return NULL;
 }
 
+static char *opt_dev_disable_paymods(const char *optarg, struct lightningd *ld)
+{
+	const char *end = optarg + strlen(optarg);
+	const char *step = optarg;
+	const char *next;
+	char *name;
+
+	/* Comma-separated list of paymod names.  */
+	while (step < end) {
+		while (*step == ',') {
+			++step;
+			if (step == end)
+				goto finished;
+		}
+
+		next = strchr(step, ',');
+		if (!next)
+			next = end;
+		assert(step != next);
+
+		name = tal_strndup(ld->dev_disable_paymods,
+				   step, next - step);
+		tal_arr_expand(&ld->dev_disable_paymods, name);
+
+		step = next;
+	}
+finished:;
+	return NULL;
+}
+
 static void dev_register_opts(struct lightningd *ld)
 {
 	/* We might want to debug plugins, which are started before normal
@@ -570,6 +600,10 @@ static void dev_register_opts(struct lightningd *ld)
 				 opt_set_bool,
 				 &ld->plugins->dev_builtin_plugins_unimportant,
 				 "Make builtin plugins unimportant so you can plugin stop them.");
+
+	opt_register_early_arg("--dev-disable-paymods",
+			       opt_dev_disable_paymods, NULL, ld,
+			       "Disable the comma-separated list of paymod names.");
 }
 #endif /* DEVELOPER */
 
@@ -1305,6 +1339,16 @@ static void add_config(struct lightningd *ld,
 			/* FIXME: We actually treat it as if they specified
 			 * --plugin for each one, so ignore these */
 #if DEVELOPER
+		} else if (opt->cb_arg == (void*)opt_dev_disable_paymods) {
+			/* We normally do not print dev settings, but
+			 * dev-disable-paymods is really handled by
+			 * plugins, so need to print it here so all plugins
+			 * can access it.  */
+			json_array_start(response, name0);
+			for (size_t i = 0; i < tal_count(ld->dev_disable_paymods); ++i)
+				json_add_string(response, NULL,
+						ld->dev_disable_paymods[i]);
+			json_array_end(response);
 		} else if (strstarts(name, "dev-")) {
 			/* Ignore dev settings */
 #endif
--- a/plugins/libplugin-paymake.c
+++ b/plugins/libplugin-paymake.c
@@ -1,0 +1,246 @@
+#include "libplugin-paymake.h"
+#include <ccan/array_size/array_size.h>
+#include <ccan/str/str.h>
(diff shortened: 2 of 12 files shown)


plugins/libplugin-paymake.c
```diff
@@ -0,0 +1,214 @@
...
+	pm->paymod_list = NULL;
+
+	p = payment_new(ctx, pm->cmd, NULL, mods);
+	/* Give responsibility of the payment modifiers array to the
+	 * payment.  */
+	tal_steal(p, mods);
+
+	/* Modify the amount_fuzz field of shadowroute, if appropriate.  */
+	if (remove_fuzz_amount)
+		payment_mod_shadowroute_get_data(p)->fuzz_amount = false;
```
niftynei, line 165: this seems kind of hacky?

plugins/libplugin-paymake.c
```diff
@@ -0,0 +1,214 @@
...
+
+	return p;
+}
+
+void paymake_disable_all_paymods(struct paymake *pm)
+{
+	if (!pm->paymod_list)
```
niftynei, line 172: iiuc if `paymod_list` is NULL tal_count will return zero

plugins/libplugin-paymake.c
```diff
@@ -0,0 +1,214 @@
...
+		/* Misuse of API.  */
+		abort();
+
+	for (size_t i = 0; i < tal_count(pm->paymod_list); ++i)
+		pm->paymod_list[i].disabled = true;
+}
+
+void paymake_disable_paymod(struct paymake *pm, const char *paymod_name)
+{
+	if (!pm->paymod_list)
```
niftynei, line 182: same as above, pretty sure you can delete this

plugins/libplugin-paymake.c
ZmnSCPxj: That is already how it is currently implemented in current code.  I agree it is hacky, but this seems to be the only way to feed configurations into paymods.

plugins/libplugin-paymake.c
ZmnSCPxj: Does not matter.  It is a misuse of the API to use a `struct paymake` after `paymake_create`, and `paymake_create` frees this `pm->paymod_list` and sets it to NULL, that is what we are detecting here.

plugins/libplugin-paymake.c
ZmnSCPxj: Again, this is API misuse guard.

lightningd/options.c
```diff
@@ -570,6 +600,10 @@ static void dev_register_opts(struct lightningd *ld)
 				 opt_set_bool,
 				 &ld->plugins->dev_builtin_plugins_unimportant,
 				 "Make builtin plugins unimportant so you can plugin stop them.");
+
+	opt_register_early_arg("--dev-disable-paymods",
```
cdecker, line 604: No need to make this an early mod if I understand correctly.

ZmnSCPxj:
Paymods are shared between `keysend` and `pay` plugins, and nothing prevents us from in the future having e.g. a `rebalance` plugin that also uses the paymod system.  So we need a single set of disabled paymods that affects all paymod-using plugins, really.  Else we need `--dev-disable-pay-paymod` `--dev-disable-keysend-paymod`.

The intent is that the set of paymods is canonically stored.  Each individual plugin MUST NOT include their own list of paymods in their source code.  This is because we already know that some paymods have a strong ordering relation with each other --- e.g. routehints MUST execute before presplit, and the payee-limiter MUST execute before presplit as well.  Once more plugins use the paymod system, any similar bugs on the ordering of paymod execution become annoying: you have to modify the order of paymods in every plugin.  And creating new general-purpose paymods, like payee-limiting, becomes annoying as well, as you have to update the paymod list of every plugin.  Thus, plugins MUST NOT maintain their own list of paymods in their source code, only blacklist specific paymods they cannot use, and prepend/append their own custom paymods.

How about this?

* Plugins store a global `static struct paymake *factory` that they initialize in their `init` functions.
  * They then indicate the paymods they want to enable, disable, and this is where the `--dev-disable-paymods` and `--disable-mpp` flags are processed.
  * After they indicate the paymods the plugins modifies they finalize the factory, making it ready for construction.  This allocates the list of paymods.
* Then, at command-invocation time, it just uses the existing factory allocated by the `init` function.

Would that work better?

cdecker:
> Paymods are shared between `keysend` and `pay` plugins, and nothing prevents us from in the future having e.g. a `rebalance` plugin that also uses the paymod system. So we need a single set of disabled paymods that affects all paymod-using plugins, really. Else we need `--dev-disable-pay-paymod` `--dev-disable-keysend-paymod`.

Oh I see why you're parsing them in `lightningd` rather than in the plugin, because there are multiple plugins. Still, we can easily have just the `pay` plugin register the option and other plugins reading it out using your `listconfigs` call. That'd free us from having to do anything in the main daemon, which doesn't have any idea what a paymod is.

I still think that having a `disable` flag in the statically instantiated `struct payment_modifier` that can be toggled based on the command line option is far simpler. We can add an `autodata` registry for paymoads, then upon startup fetch the config which paymods to disable, then iterate over the registry and disable the ones that were listed. Finally we just skip the paymods that were disabled in the `payment_continue` dispatch function. It's cleaner, easier and impossible to misuse.

ZmnSCPxj:
Unfortunately paymods are sensitive to the order in which they are registered, so you need some single canonical place to indicate this order, so using `AUTODATA` would not be easy (it could be done, maybe if you can indicate "this goes before that paymod" siomehow).

ZmnSCPxj:
Am working on a complete reimplementation of a payment framework anyway, so ---

PR #3952, opened 2020-08-17
Updates proto to use coincurve for Public and Private keys

### LINKED ISSUE #3751: pyln.proto.wire should use coincurve PublicKey/PrivateKey types
Now we use coincurve elsewhere, seems a bit like it's reinventing the wheel here?

sr-gi:
Updates wire for pyln-proto to use coincurve in the `PublicKey` and `PrivateKey` classes.

Still unsure about https://github.com/ElementsProject/lightning/issues/3751#issuecomment-674788115

I may add some additional tests before moving from draft to an actual PR.

Close #3751

### DIFF
--- a/contrib/pyln-proto/pyln/proto/wire.py
+++ b/contrib/pyln-proto/pyln/proto/wire.py
@@ -5,7 +5,6 @@ from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305
 from cryptography.hazmat.primitives.kdf.hkdf import HKDF
-from cryptography.hazmat.primitives import serialization
 from hashlib import sha256
 import coincurve
 import os
@@ -59,18 +58,19 @@ def decryptWithAD(k, n, ad, ciphertext):
 
 class PrivateKey(object):
     def __init__(self, rawkey):
-        assert len(rawkey) == 32 and isinstance(rawkey, bytes)
+        if not isinstance(rawkey, bytes):
+            raise TypeError(f"rawkey must be bytes, {type(rawkey)} received")
+        elif len(rawkey) != 32:
+            raise ValueError(f"rawkey must be 32-byte long. {len(rawkey)} received")
+
         self.rawkey = rawkey
-        rawkey = int(hexlify(rawkey), base=16)
-        self.key = ec.derive_private_key(rawkey, ec.SECP256K1(),
-                                         default_backend())
+        self.key = coincurve.PrivateKey(rawkey)
 
     def serializeCompressed(self):
-        return self.key.private_bytes(serialization.Encoding.Raw,
-                                      serialization.PrivateFormat.Raw, None)
+        return self.key.secret
 
     def public_key(self):
-        return PublicKey(self.key.public_key())
+        return PublicKey(self.key.public_key)
 
 
 class Secret(object):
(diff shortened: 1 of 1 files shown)


contrib/pyln-proto/pyln/proto/wire.py
```diff
@@ -59,18 +58,19 @@ def decryptWithAD(k, n, ad, ciphertext):
...
+        if not isinstance(rawkey, bytes):
+            raise TypeError(f"rawkey must be bytes, {type(rawkey)} received")
+        elif len(rawkey) != 32:
+            raise ValueError(f"rawkey must be 32-byte long. {len(rawkey)} received")
+
         self.rawkey = rawkey
-        rawkey = int(hexlify(rawkey), base=16)
-        self.key = ec.derive_private_key(rawkey, ec.SECP256K1(),
-                                         default_backend())
+        self.key = coincurve.PrivateKey.from_int(int(hexlify(rawkey), base=16))
```
sr-gi, line 67: Regarding the initialisation of key, this could be done as is using the `from_int` method (bytes -> hexlified bytes -> int) or using the `from_hex` method (bytes ->  hexlified bytes + utf-8 decoding):

self.key = coincurve.PrivateKey.from_hex((hexlify(rawkey).decode('utf-8')))

I was not sure if you'd rather an additional cast vs a decoding, so I left it as is, but it can be changed.

niftynei:
concept ACK

contrib/pyln-proto/pyln/proto/wire.py
rustyrussell: How about:
```
self.key = coincurve.PrivateKey(rawkey)
```
?

contrib/pyln-proto/pyln/proto/wire.py
```diff
@@ -59,18 +58,19 @@ def decryptWithAD(k, n, ad, ciphertext):
...
 
     def serializeCompressed(self):
-        return self.key.private_bytes(serialization.Encoding.Raw,
-                                      serialization.PrivateFormat.Raw, None)
+        return hexlify(self.key.to_hex())
```
rustyrussell, line 70: self.key.format()?

rustyrussell (changes requested):
Nice idea, but some of this can be simplified.

Also, hexlify is unnecessary in modern python: bytes.from_hex() and .hex() work on bytes.

contrib/pyln-proto/pyln/proto/wire.py
sr-gi: There's no format for PrivateKey, only for PublicKey in coincurve. 

I can do ``bytes.fromhex(self.key.to_hex())``, but that'll be equivalent to ``self.rawkey`` and ``self.key.secret``.

Is this the expected return though, the current version of the method doesn't work AFAICT so not sure if the raw key is expected or a base58 encoding. If so, any of the last two should do.

contrib/pyln-proto/pyln/proto/wire.py
sr-gi: That'll do indeed.

PS: I actually didn't know about bytes.from_hex() and .hex(), TIL!

sr-gi: Sorted in 259330f60f3cf3195d1234a46601e31a80d5ca98

contrib/pyln-proto/pyln/proto/wire.py
sr-gi: I've gone for `self.key.secret` in 0cbe96287c932134eb7f94a97e92765536e1821c, but can rebase and change it if the expected return format is different

contrib/pyln-proto/pyln/proto/wire.py
sr-gi: I've also changed all the other uses of `hexlify` in 89088e29f071e385634cd06cd09639eed7b39d89 and removed `binascii`.

contrib/pyln-proto/pyln/proto/wire.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, rawkey):
             raise ValueError(f"rawkey must be 32-byte long. {len(rawkey)} received")
 
         self.rawkey = rawkey
-        self.key = coincurve.PrivateKey.from_int(int(hexlify(rawkey), base=16))
+        self.key = coincurve.PrivateKey(rawkey)
```
niftynei, line 66: this commit is unnecessary, since it reverts a change in the prior commit in this same patchset. You should fold these into the same commit.

contrib/pyln-proto/pyln/proto/wire.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, rawkey):
         self.key = coincurve.PrivateKey(rawkey)
 
     def serializeCompressed(self):
-        return hexlify(self.key.to_hex())
```
niftynei: this commit is unnecessary, since it reverts a change in the first commit in this same patch. You should fold these into the same commit.

contrib/pyln-proto/pyln/proto/wire.py
niftynei: you should fold in your later changes, such that this patch does not introduce any new uses of hexlify.

sr-gi:
Rebased

rustyrussell (approved):

rustyrussell:
Thanks!

PR #3954, opened 2020-08-18
fundpsbt/utxopsbt: let caller specify locktime, signpsbt: signing restrictions.

rustyrussell:
Required for dual funding where the opener sets it.

Changelog-Added: JSON-RPC: fundpsbt takes a new `locktime` parameter

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -1140,6 +1140,21 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("fundpsbt", payload)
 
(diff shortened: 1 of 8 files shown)


ZmnSCPxj:
Is this strictly necessary?  Does not the caller have the ability, to just mutate the `nLockTime` of the returned PSBT, prior to merging with the PSBT from the opener?

rustyrussell:
> Is this strictly necessary? Does not the caller have the ability, to just mutate the `nLockTime` of the returned PSBT, prior to merging with the PSBT from the opener?

You could, but I think this is neater.  We'll eventually want a "version" flag, too.

I've rebased and appended some more modifications, too.

Issue #2662, opened 2019-05-22
fuzz testing


jb55:
[fuzz testing the linux kernel](https://www.youtube.com/watch?v=qrBVXxZDVQY) has cropped thousands of bugs, mostly in interfaces where you are encoding and decoding stuff. [rust-lightning](https://github.com/rust-bitcoin/rust-lightning/tree/master/fuzz) has been building fuzzers for lightning. It would be interesting to see how badly we can break c-lightning with these kinds of tests.

niftynei:
If these count as 'fuzz' testing then we're doing some amount of this already, but could definitely improve.

https://github.com/ElementsProject/lightning/blob/master/tests/test_connection.py#L1494
https://github.com/ElementsProject/lightning/blob/master/wire/test/run-peer-wire.c#L904

ZmnSCPxj:
What fuzz framework is easy to integrate with a  networking application written in C?

jb55:
The go-to appears to be afl (american fuzzy loop), I haven't tried it yet.

ZmnSCPxj:
american fuzzy lop, not loop.
It seems, named after a dog breed, whatever a "dog" is, I am sure most humans find it delicious.

From what I understand, it requires that a program input a file.
Possibly we can compile the individual channel-level daemons with some way of forcing their inputs to arrive from a file rather than from "real" `lightningd` or other peers.
Though I admit I have not investigated much.

ZmnSCPxj:
For the individual channel-level daemons, possibly we can add a `--afl-fuzz` flag that causes them to instead read messages from `STDIN_FILENO` and throw any output messages to `/dev/null`.

Possibly we can have messages from `STDIN_FILENO` mode to include whether the message is supposed to come from `lightningd` or from the network, then have an internal forwarder emulate the sockets.  Not sure how to emulate forwarding of fds, maybe just close them when transferring out, and have a special code byte in `STDIN_FILENO` to feed a faked fd in.

`onchaind` is probably simplest as it only interacts with `lightningd`.

For the more complex daemons such as `gossipd`, `connectd`, and `lightningd` itself..................

PR #3957, opened 2020-08-20
JSON optimizations, cleanups and listsendpays optimizations

rustyrussell:
@fiatjaf reported in #3941 that `listspays` was slow, and indeed, it was.

I created 50,000 payments, and ran some tests.

* With no optimization:
    * `listsendpays` takes 0.983 seconds
    * `listpays` takes 52.415 seconds
* With `-O3 -flto`:
    * `listsendpays` takes 0.628 seconds
    * `listpays` takes 43.104 seconds

After these optimizations (mostly FIXMEs!) the results are:

* With no optimization:
    * `listsendpays` takes 0.676 seconds
    * `listpays` takes 1.545 seconds.
* With `-O3 -flto`:
    * `listsendpays` takes 0.416 seconds
    * `listpays` takes 0.971 seconds.

### DIFF
--- a/lightningd/test/run-jsonrpc.c
+++ b/lightningd/test/run-jsonrpc.c
@@ -132,11 +132,12 @@ static int test_json_filter(void)
 	struct json_stream *result = new_json_stream(NULL, NULL, NULL);
 	jsmntok_t *toks;
 	const jsmntok_t *x;
-	bool valid;
+	bool valid, complete;
 	int i;
 	char *badstr = tal_arr(result, char, 256);
 	const char *str;
 	size_t len;
+	jsmn_parser parser;
 
 	/* Fill with junk, and nul-terminate (256 -> 0) */
 	for (i = 1; i < 257; i++)
@@ -150,9 +151,11 @@ static int test_json_filter(void)
 	str = json_out_contents(result->jout, &len);
 	str = tal_strndup(result, str, len);
 
-	toks = json_parse_input(str, str, strlen(str), &valid);
+	toks = toks_alloc(str);
+	jsmn_init(&parser);
+	valid = json_parse_input(&parser, &toks, str, strlen(str), &complete);
 	assert(valid);
-	assert(toks);
+	assert(complete);
 
 	assert(toks[0].type == JSMN_OBJECT);
 	x = json_get_member(str, toks, "x");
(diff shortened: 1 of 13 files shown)


plugins/libplugin.c
```diff
@@ -651,9 +651,11 @@ static bool rpc_read_response_one(struct plugin *plugin)
 	bool valid;
 	const jsmntok_t *toks, *jrtok;
 
-	/* FIXME: This could be done more efficiently by storing the
-	 * toks and doing an incremental parse, like lightning-cli
-	 * does. */
+	/* For our convenience, lightningd always ends JSON requests with
```
ZmnSCPxj: ~~This forces us to print `\n\n` forevermore.  Maybe acceptable tho.  I think the incremental-parse idea would prevent us from effectively making `\n\n` part of the contract with the plugin.~~

~~If we are going to officially depend on this detail, I suggest adding a note in `doc/PLUGINS.md` or other documentation that our command stream acts this way.~~

~~I suppose my concern is that independent projects might attempt to run plugins by emulating the `lightningd` interface.  For example, the vaunted "fuzz testing" #2662.~~

EDIT: Nevermind you change it in a later commit.  Makes me wonder if **this** particular change needs to exist, tho.

ZmnSCPxj (approved):
ACK 94058b41b8710a12499848c83160d9b06c464d68

Minor quibble though.

plugins/libplugin.c
rustyrussell: Turns out we already rely on this and publish it: in particular, the Python plugin library relies on it.

That's because most languages do not offer an interface for continuous parsing, and so are grossly inefficient :(

rustyrussell: Oh, and reworking it all is messy and means I have to rerun all my benchmarks... again :)

rustyrussell:
Fixed core dump in test-json, which I hadn't fixed up to the new API (at once stage I used NULL args to json_parse_input instead of implementing a separate json_parse_simple).

darosior (approved):
ACK 7471a830224a6b3ea6b55cc7dbadcfbee1937e4f

Impressive!

lightningd/test/run-jsonrpc.c
```diff
@@ -150,9 +151,10 @@ static int test_json_filter(void)
 	str = json_out_contents(result->jout, &len);
 	str = tal_strndup(result, str, len);
 
-	toks = json_parse_input(str, str, strlen(str), &valid);
+	toks = toks_alloc(str);
+	jsmn_init(&parser);
+	valid = json_parse_input(&parser, &toks, str, strlen(str), NULL);
```
darosior: NULL will be dereferenced, maybe `json_parse_input_simple` ?

darosior: Same below

rustyrussell:
... and the other test where I made the same damn mistake...

lightningd/test/run-jsonrpc.c
darosior: (did not submit the review)...

rustyrussell: YEah, originally instead of the json_parse_simple() API I allowed NULL args for parser and complete.  But I reverted that, and of course, missed this :(

Fixed in the obvious way now...

PR #3959, opened 2020-08-20
BOLT updates to latest

rustyrussell:
Which actually have some side effects:
1. We increase our default delays as per new recommendations.
2. We prune channels which haven't been updated by *both* peers in 2 weeks.


darosior (approved):
untested ACK 6fd1cfb3c1e7484cbaa99715d6de17ea4423375b

Issue #3550, opened 2020-02-26
feature request: decommission


m-schmoock:
## Description
Add a new pair of `lightning-cli` RPC commands that allow the permanent shutdown of a node. The commands may be called `decommission` and `recommission`. This was discussed in #3499 

These commands can be used to close all channels and prevent acceptance of new funded channels by remote peers. Since this is a basic operation (any node goes offline eventually) and also security related (operator may decide its host maybe compromised) we should not have this as an optional python plugin but maybe as an integrated C plugin.

Decommissioning a node for operational reasons should focus on closing channels cooperatively to keep a low onchain footprint. Decommissioning a node on a hurry for security reasons requires a shorter force-close timeout, an external address and likely higher onchain fees.

---
#### Immediately
- If `[destination]` is given, try to send out available channel liquidity to `[destination]` pubkey via `keysend` for a given (short) timeout, e.g. 5 minutes.
- Close all responsive channels.
- Send any immediately available on-chain funds to an optional `[address_or_xpub]`, if specified.
- Start reject opening of new channels by the operator with a proper error message indicating the decommissioned state of this node.
- Stop accepting newly funded channels from remote peers.
#### In the meantime
- Force-close all uncooperative or offline channels after given `timeout`, defaulting to a _slow  decommission_ `timeout` of 24hr to give remote counterparts a chance to get online in the meantime for cooperative close. If the user is on a hurry, he can repeat the command with a faster `timeout` value.
- Redirect any settled and released on-chain funds resulting from closed channels to an optional `[address_or_xpub]`, if specified.
- Redirect other received on-chain funds to `[address_or_xpub]`, if specified.
- If we are in slow mode, we can wait and aggregate a batch onchain transaction to save on fees.
#### When done
- Move the `hsm_secret` file to a backup location, so the node can't accidentally be started after the decommission completes.
- Stop the node

### `lightning-cli recommission`

Cancels an ongoing `decommission` process by:
- Restore the moved `hsm_secret` backup, if prior `decommission` was finished.
- Allow again creation of new channels by the operator
- Accept remote funding again
- Stop redirecting released funds to external wallet

### `lightning-cli commissionstate`

Show the user the state an process of an ongoing decommissioning operation:
 - state
 - redirect destination (if set)
 - channels closed
 - channels remaining
 - time remaining until force-close timeout
 - ...

-------

## Technical Requirements and Stepstones

- [ ] integrated C plugin with  json rpc commands and descriptions
[...]

PR #3960, opened 2020-08-20
openchannel: make openchannel hook chainable

m-schmoock:
This change is required to make the decomission feature possible: https://github.com/ElementsProject/lightning/issues/3550

**Status: Done.** The code is functional and does not seem to have bugs. There is a strange workaround in  `tests/test_plugin.py` `test_openchannel_hook_chaining` where it needs to check for actual hook execution order in order to have the test stable. this is caused by the way hooks are currently registered, see #4005 for details about this.

**Question:** Currently this change will just log `UNUSUAL` if a second hook also tries to set a `close_to` address and continues normal operation. Question: Is this sufficient or should rather want to `fatal` such plugin setups.


rustyrussell:
Definitely want this!

lightningd/opening_control.c
```diff
@@ -804,13 +806,83 @@ static void openchannel_payload_remove_openingd(struct subd *openingd,
...
+
+	const jsmntok_t *t = json_get_member(buffer, toks, "result");
+	if (!t)
+		fatal("Plugin returned an invalid response to the"
+		      " openchannel hook: %.*s",
+		      toks[0].end - toks[0].start,
+		      buffer + toks[0].start);
+
+	/* reject */
+	// TODO: hooks do not seem to be called serialized but parallel!
```
niftynei, line 831: Is this what's left on this PR?

m-schmoock: @niftynei It seems I have caused some valgrind stuff that needs to be addressed.

m-schmoock: The TODO was maybe caused by indeterminate code flow because of that. after that I will check if tests are finally stable.
I will close this draft and submit a fresh PR when done.

m-schmoock:
Valgrind issues are fixed now, tests still unstable (at least locally). Need to investigate that final piece...

lightningd/opening_control.c
```diff
@@ -820,65 +821,87 @@ static void openchannel_hook_cb(struct openchannel_hook_payload *payload STEALS,
...
+	if (json_tok_streq(buffer, t, "reject")) {
+		payload->errmsg = "";
+		t = json_get_member(buffer, toks, "error_message");
+		if (t)
+			payload->errmsg = json_strdup(payload, buffer, t);
+		log_debug(openingd->ld->log,
+			  "openchannel_hook rejects and says '%s'",
+			  payload->errmsg);
+		t = json_get_member(buffer, toks, "close_to");
+		if (t)
```
cdecker, line 860: Nit: I usually prefer to have separate variables to the tokens and search for all of them initially, so we don't get confused by the same variable name assuming different meanings, and making composite checks simpler to grok.

cdecker: Can be addressed in a cleanup

tests/plugins/openchannel_hook_accept.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+"""Plugin to test openchannel_hook
+
+Will simply acceppt any channel. Useful fot tetsing chained hook.
```
cdecker, line 4:
```suggestion
Will simply accept any channel. Useful for testing chained hook.
```

tests/plugins/openchannel_hook_reject.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python3
+"""Plugin to test openchannel_hook
+
+Will simply reject any channel with message "reject on principle".
+Useful fot tetsing chained hook.
```
cdecker, line 5:
```suggestion
Useful for testing chained hook.
```

tests/test_plugin.py
```diff
@@ -574,6 +574,82 @@ def test_openchannel_hook(node_factory, bitcoind):
         l1.rpc.fundchannel(l2.info['id'], 100001)
 
 
+def test_openchannel_hook_error_handling(node_factory, bitcoind):
+    """ l2 uses a plugin that should fatal() crash the node.
+
+    This is because the plugin rejects a channel while
+    also setting a close_to address which isn't allowed.
+    """
+    opts = {'plugin': os.path.join(os.getcwd(), 'tests/plugins/openchannel_hook_accepter.py')}
```
cdecker, line 583:
```suggestion
    opts = {'plugin': os.path.join(os.path.dirname(__file__), '..', 'tests/plugins/openchannel_hook_accepter.py')}
```
This is more stable and independent of the directory you run the tests from.

tests/test_plugin.py
```diff
@@ -574,6 +574,82 @@ def test_openchannel_hook(node_factory, bitcoind):
...
+    addr = l1.rpc.newaddr()['bech32']
+    txid = bitcoind.rpc.sendtoaddress(addr, 10)
+    numfunds = len(l1.rpc.listfunds()['outputs'])
+    bitcoind.generate_block(1, txid)
+    wait_for(lambda: len(l1.rpc.listfunds()['outputs']) > numfunds)
+
+    # next fundchannel should fail fatal() for l2
+    with pytest.raises(RpcError, match=r'Owning subdaemon openingd died'):
+        l1.rpc.fundchannel(l2.info['id'], 100004)
+    assert l2.daemon.is_in_log("Plugin rejected openchannel but also set a close_to")
```
cdecker, line 602: We have a helper that does all this:

```suggestion
    l1.connect(l2)
    
    # The plugin used by l2 causes an error by setting a close_to and rejecting
    with pytest.raises(RpcError, match=r'Owning subdaemon openingd died'):
        l1.fundchannel(l2, 100004)
    assert l2.daemon.is_in_log("Plugin rejected openchannel but also set a close_to")
```

cdecker (commented):
Just some minor nits and cleanups. Otherwise good work :+1:

cdecker (approved):

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3960/commits/dbbad65c7f4fb2c1169c8caa948ed12dbdb1935c

tests/test_plugin.py
m-schmoock: Thx for pointing out, fixed that for all three openchannel hook tests.

tests/test_plugin.py
m-schmoock: didn't know that, though it looks ugly ;)

lightningd/opening_control.c
m-schmoock: great, looks a lot better that way

cdecker:
Thanks for the fixup @m-schmoock, excellent work :+1: 

ACK https://github.com/ElementsProject/lightning/pull/3960/commits/e006d19dbcde4575c346b95fc35fd0d3afe023bb

m-schmoock:
Do we want to tag this for the upcoming release?

PR #3963, opened 2020-08-21
listpeers: add `features` array using BOLT9 names.

rustyrussell:
It's actually not possible to currently tell if you're using anchor_outputs
with a peer (since it depends on whether you both supported it at *channel open*).

Changelog-added: JSON-RPC: `listpeers` shows `features` list for each channel.

### DIFF
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -155,6 +155,18 @@ void json_add_uncommitted_channel(struct json_stream *response,
 		json_add_amount_msat_compat(response, total,
 					    "msatoshi_total", "total_msat");
 	}
+
+	json_array_start(response, "features");
+	if (feature_negotiated(uc->peer->ld->our_features,
+			       uc->peer->their_features,
+			       OPT_STATIC_REMOTEKEY))
+		json_add_string(response, NULL, "option_static_remotekey");
+
+	if (feature_negotiated(uc->peer->ld->our_features,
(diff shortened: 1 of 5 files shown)


lightningd/opening_control.c
```diff
@@ -155,6 +155,18 @@ void json_add_uncommitted_channel(struct json_stream *response,
 		json_add_amount_msat_compat(response, total,
 					    "msatoshi_total", "total_msat");
 	}
+
+	json_array_start(response, "features");
+	if (feature_negotiated(uc->peer->ld->our_features,
```
ZmnSCPxj, line 160: Would it not be better (and more easy to extend with new features in the future) to have an array of structs like so?

```c
static const struct { const char *name; size_t feature; } known_features[] = {
        { "option_static_remotekey", OPT_STATIC_REMOTEKEY},
        { "option_anchor_outputs", OPT_ANCHOR_OUTPUTS}
};
```

Then just iterate:

```c
for (size_t i = 0; i < ARRAY_SIZE(known_features); ++i)
        if (feature_negotiated(our_features, their_features, known_features[i].feature))
                json_add_string(response, NULL, known_features[i].name);
```

Also, is not `their_features` set on `init` message, and the peer could be initially an old software, create channel with us pre-anchor-commitments, then upgrade so they now both support anchor-commitments but the actual channel is not using anchor commitments yet?

Issue #3608, opened 2020-03-28
Channel closed because WIRE_CHANNEL_REESTABLISH is an unknown message?


fiatjaf:
### Issue

My BLW had a channel open to my c-lightning node, and amidst many connection and connection losses suddenly this log entry appeared: `03da3b917a178f46472f55ba2dfc192431286d562200a95e9e6479126e94bc30bd-chan#10877: Peer permanent failure in CHANNELD_NORMAL: channeld: sent ERROR Peer sent unknown message 136 (WIRE_CHANNEL_REESTABLISH)`

Then the channel was closed.

```
2020-03-28T17:24:15.869Z INFO 03da3b917a178f46472f55ba2dfc192431286d562200a95e9e6479126e94bc30bd-channeld-chan#10877: Peer connection lost
2020-03-28T17:24:15.869Z INFO 03da3b917a178f46472f55ba2dfc192431286d562200a95e9e6479126e94bc30bd-chan#10877: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
2020-03-28T17:26:53.880Z UNUSUAL 03da3b917a178f46472f55ba2dfc192431286d562200a95e9e6479126e94bc30bd-chan#10877: Peer permanent failure in CHANNELD_NORMAL: channeld: sent ERROR Peer sent unknown message 136 (WIRE_CHANNEL_REESTABLISH)
2020-03-28T17:26:53.889Z INFO 03da3b917a178f46472f55ba2dfc192431286d562200a95e9e6479126e94bc30bd-chan#10877: State changed from CHANNELD_NORMAL to AWAITING_UNILATERAL
```

Here's the message on BLW screen (as received from lightningd):
![photo_2020-03-28_17-02-51](https://user-images.githubusercontent.com/1653275/77832548-ad104d00-7115-11ea-9910-caaad4570e6a.jpg)

### `getinfo` output
```
{
   "id": "02bed1812d3824f7cc4ccd38da5d66a29fcfec146fe95e26cd2e0d3f930d653a8d",
   "alias": "etleneum.com",
   "color": "c2f1bb",
   "num_peers": 7,
   "num_pending_channels": 0,
   "num_active_channels": 6,
   "num_inactive_channels": 1,
   "address": [
      {
         "type": "ipv6",
         "address": "2a01:4f8:c0c:7671::1",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv6",
         "address": "::",
[...]

PR #3964, opened 2020-08-21
channeld/channeld.c: Log and ignore repeated WIRE_CHANNEL_REESTABLISH.

ZmnSCPxj:
Fixes: #3608

Changelog-Changed: protocol: Ignore (and log as "unusual") repeated `WIRE_CHANNEL_REESTABLISH` messages, to be compatible with buggy peer software that sometimes does this.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -1975,6 +1975,60 @@ static void send_onionmsg(struct peer *peer, const u8 *msg)
 }
 #endif /* EXPERIMENTAL_FEATURES */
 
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
No idea how to test this tho.

whitslack:
It honestly seems bonkers to me that the Lightning peer-wire protocol doesn't allow NAK'ing protocol messages in any way that allows the channel to remain open. How can we expect to achieve interoperability when implementations are utterly unable to say, "I disagree with (or did not understand) the message you just sent, so I will take no action, and you need not either"?

PR #3967, opened 2020-08-23
`donateutxo` to donate dust-attack UTXOs to miners

ZmnSCPxj:
What does ZmnSCPxj do when he is bored of stacking sats? Stack PRs, that is what!

This implements a command you can use to donate UTXOs in case you are dust-attacked by evil blockchain analysis companies.

### DIFF
--- a/bitcoin/script.c
+++ b/bitcoin/script.c
@@ -9,6 +9,7 @@
 #include <ccan/endian/endian.h>
 #include <ccan/mem/mem.h>
 #include <common/utils.h>
+#include <sodium/randombytes.h>
 
 /* Some standard ops */
 #define OP_0		0x00
@@ -219,6 +220,16 @@ u8 *scriptpubkey_opreturn(const tal_t *ctx)
 	u8 *script = tal_arr(ctx, u8, 0);
 
 	add_op(&script, OP_RETURN);
+	return script;
+}
+u8 *scriptpubkey_opreturn_padded(const tal_t *ctx)
+{
+	u8 *script = tal_arr(ctx, u8, 0);
+	u8 random[20];
+	randombytes_buf(random, sizeof(random));
+
+	add_op(&script, OP_RETURN);
+	script_push_bytes(&script, random, sizeof(random));
 	return script;
 }
 
--- a/bitcoin/script.h
+++ b/bitcoin/script.h
@@ -30,6 +30,11 @@ u8 *scriptpubkey_p2pkh(const tal_t *ctx, const struct bitcoin_address *addr);
 
 /* Create a prunable output script */
 u8 *scriptpubkey_opreturn(const tal_t *ctx);
+/* Create a prunable output script with 20 random bytes.
+ * This is needed since a spend from a p2wpkh to an `OP_RETURN` without
+ * any other outputs would result in a transaction smaller than the
+ * minimum size.  */
+u8 *scriptpubkey_opreturn_padded(const tal_t *ctx);
 
 /* Create an input script which spends p2pkh */
 u8 *bitcoin_redeem_p2pkh(const tal_t *ctx, const struct pubkey *pubkey,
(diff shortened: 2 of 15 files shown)


ZmnSCPxj:
CI error is complaining about **`libsodium`** of all things, WTF.

ZmnSCPxj:
Okay, tried including a different header file for libsodium use, seems to work.

cdecker (changes requested):
I don't think this is a good idea at all:

1. We should not expose this kind of edge case handling to all of our users. If they want something like this they should use the PSBT-related API methods to construct this kind of thing themselves. Adding it by default will result in more users being confused about its use than users that actually could use this.
2. In almost all cases it is uneconomical to spend outputs that have such a low value that a user would ever consider donating them, and these outputs should not be moved, costing the network more than what an individual miner makes by collecting them in terms of network capacity.

ZmnSCPxj:
So how should we handle dust attacks instead?  Can we just delete them from the UTXO database so that we never spend them?

ZmnSCPxj:
Never spending the dust means all fullnodes need to maintain the UTXO in the UTXO set forever, whereas donating them to miners means they can remove it from the UTXO set (because it becomes an unspendable `OP_RETURN`).  Either way, resources are wasted.  Dust attacks are not something we can control, only react to, because people spying on other people is a lucrative business we cannot stop dust attacks anyway

I disagree that we should just let users handle PSBTs manually to handle dust outputs: they could make a mistake, and.then they lose more than their privacy.

ZmnSCPxj:
An input spend is 272 sipa:

    4x (32 + 4) txid vout
    4x 4 nSequence
    4x 1 scriptSig length
    1 witness length
    73 signature
    1 pubkey push
    33 pubkey

Assuming no blockchain congestion, we could have as low as 253 sat/ksipa, so spending even a 547-sat output costs only 68.816 satoshi, leaving 478.184 satoshi we can still spend.

Suppose we perform a `fundpsbt`.  `fundpsbt` selects inputs at random, so the first input it selects could be a 547-sat dust input that was specifically seeded by evil blockchain analysis companies.  Then it will look for other inputs it can spend to fulfill the requested amount.  This leads to linking the address of the 547-sat dust input to other addresses the node controls, which is exactly the privacy break that blockchain analysis companies want to exploit in a dusting attack.

Dusting can potentially victimize any user, so this should be convenient and easy to perform for users, thus MUST be provided out of the box by `lightningd`.

So we either:

* Donate the UTXO to miners by sending to an `OP_RETURN` output (this PR).
* Donate the UTXO to HODLers by deleting the UTXO from our spendable UTXO set in our wallet.

Donating the UTXO to miners is not so onerous a load on the blockchain.  A single-input single-output donate transaction spending from a P2WPKH to a 20-byte `OP_RETURN` (this PR) consists of:

     272 (quoted above) input
     42 common core
     124 output
         4x 8 amount
         4x 1 script length
         4x 1 OP_RETURN
         4x 1 push-20
         4x 20 20 random bytes

Resulting in a 438-sipa transaction.  547sat / 0.438ksipa = 1,248 sat/ksipa, or more traditionally as 4.992 sat/vbyte.  That is still a good amount above what happens on weekends where 1sat/vbyte txes start getting put onchain.

cdecker:
> So how should we handle dust attacks instead? Can we just delete them from the UTXO database so that we never spend them?

We just improve our coin selection to not consider inputs that are uneconomical to spend, and dusting attacks (a generally overvalued and dumb attack) are no longer a problem. Really this is the last thing we should be worrying about, given our very simplistic coin selection, which doesn't take any privacy enhancing strategies into account.

Furthermore, dusting attacks in the context of LN are likely not an issue, because we already link our addresses to our node ID when funding and announcing channels. We then have routed payments to improve privacy on the layer built on top.

ZmnSCPxj:
We *already* reject inputs that are uneconomical to spend.  The problem is that the 547sat dust is ***still*** economical to spend in a 253sat/ksipa environment, as I pointed out.

Anyway, fine, I am closing this PR.

PR #3972, opened 2020-08-24
pyln-proto: use vals for subtype parsing

niftynei:
There's more work to do on subtypes here, but this fixes one small problem with their parsing.

Changelog-None.


rustyrussell (approved):
Ack 65a8e278ce6cbff61a17d2fadcea4f52b8ef78e7

PR #3971, opened 2020-08-24
Small PSBT updates, precursor for dual-funding

niftynei:
Few small things I found while working on the dual-funding accepter branch.

Changelog-None.

PR #3973, opened 2020-08-25
Dual-funding, accepter side only

niftynei:
Hello, it hath arrived. This is *just* the accepter side of the "anyone can pay" [channel open protocol](https://github.com/niftynei/lightning-rfc/pull/1). It includes a half (literally) working implementation of the protocol, the 'accepter' side. 

The spec is still in draft, so this is liable to change in future revisions. The option flag is currently not signaled (even if built in experimental), as the opener side does not exist.

Still to come:
- the opener side implementation
- PODLEs
- RBF part of the protocol

Edit: Requires https://github.com/ElementsProject/lightning/pull/3971 to build as --enable-experimental


openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+			    "channel ids don't match. expected %s, got %s",
+			    type_to_string(tmpctx, struct channel_id, orig_id),
+			    type_to_string(tmpctx, struct channel_id, id_in));
+}
+
+static struct amount_sat total_funding(struct state *state)
+{
+	struct amount_sat total;
+	if (!amount_sat_add(&total, state->opener_funding, state->accepter_funding))
+		abort();
```
rustyrussell, line 303: Note that they control one of these sides: are you sure this can't fail?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+{
+	struct amount_sat reserve;
+
+	/* BOLT-FIXME #2
+	 *
+	 * The channel reserve is fixed at 1% of the total channel balance
+	 * rounded down (sum of `funding_satoshis` from `open_channel2` and `accept_channel2`)
+	 * or the `dust_limit_satoshis`, whichever is greater.
+	 */
+	reserve.satoshis = total_funding(state).satoshis / 100; /* Raw: rounding */
```
rustyrussell, line 317: amount_sat_div

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	else
+		state->remoteconf.channel_reserve = reserve;
+
+	if (amount_sat_greater(state->localconf.dust_limit, reserve))
+		state->localconf.channel_reserve = state->localconf.dust_limit;
+	else
+		state->localconf.channel_reserve = reserve;
+}
+
+static bool is_openers(const struct wally_map *unknowns)
```
rustyrussell, line 330: BOLT quote!

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	weight = (32 + 4 + 4) * 4;
+	weight +=
+		(psbt->inputs[in].redeem_script_len +
+			(varint_t) varint_size(psbt->inputs[in].redeem_script_len)) * 4;
+
+	ok = psbt_input_get_max_witness_len(&psbt->inputs[in],
+					    &max_witness_len);
+	assert(ok);
+	weight += max_witness_len;
+	return weight;
```
rustyrussell, line 357: This should probably be spelled out in the BOLT, too.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+
+	/* Find the total input and output sums for this participant */
+	input_amt = AMOUNT_SAT(0);
+	side_input_amt = AMOUNT_SAT(0);
+	for (size_t i = 0; i < psbt->num_inputs; i++) {
+		struct amount_sat amt = psbt_input_get_amount(psbt, i);
+
+		/* Add to total balance check */
+		ok = amount_sat_add(&input_amt, input_amt, amt);
+		assert(ok);
```
rustyrussell, line 383: I think some of these assert shouold be return false...

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+
+
+/*~ This is the key function that checks that their configuration is reasonable:
+ * it applied for both the case where they're trying to open a channel, and when
+ * they've accepted our open. */
+static bool check_configs(struct state *state,
+			  const struct channel_config *remoteconf,
+			  bool am_opener)
+{
+	struct amount_sat capacity;
```
rustyrussell, line 436: This function recently changed in openingd (anchor outputs).  Might need to extract that and share?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+}
+
+static bool is_segwit_output(struct wally_tx_output *output,
+			     const u8 *redeemscript)
+{
+	u8 *wit_prog;
+	if (tal_bytelen(redeemscript) > 0)
+		wit_prog = cast_const(u8 *, redeemscript);
+	else
+		wit_prog = cast_const(u8 *, wally_tx_output_get_script(tmpctx, output));
```
rustyrussell, line 586: why isn't `wit_prog` declared `const u8 *`?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+		char *err;
+		if (!fromwire_dual_open_fail(msg, msg, &err))
+			master_badmsg(msg_type, msg);
+
+		peer_failed(state->pps, &state->channel_id, "%s", err);
+	} else if (msg_type == WIRE_DUAL_OPEN_PSBT_CHANGED) {
+		struct channel_id unused;
+		if (!fromwire_dual_open_psbt_changed(state, msg, &unused, &updated_psbt)) {
+			master_badmsg(msg_type, msg);
+			return NULL;
```
rustyrussell, line 615: master_badmsg does not return?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	struct wally_psbt *updated_psbt;
+
+	/* Go ask lightningd what other changes we've got */
+	msg = towire_dual_open_psbt_changed(NULL, &state->channel_id, psbt);
+
+	wire_sync_write(REQ_FD, take(msg));
+	msg = wire_sync_read(tmpctx, REQ_FD);
+
+	msg_type = fromwire_peektype(msg);
+	if (msg_type == WIRE_DUAL_OPEN_FAIL) {
```
rustyrussell, line 605: I generally prefer to just call fromwire_xxx and if it fails, try the other case.  The first thing fromwire does is check the type...

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+static bool send_next(struct state *state, struct wally_psbt **psbt)
+{
+	u8 *msg;
+	bool finished = false;
+	struct wally_psbt *updated_psbt;
+	struct psbt_changeset *cs = state->changeset;
+
+	/* First we check our cached changes */
+	msg = changeset_get_next(tmpctx, &state->channel_id,
+				 &cs);
```
rustyrussell, line 643: This fits

on one line.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	set->rm_ins = tal_arr(set, struct input_set, 0);
+	set->added_outs = tal_arr(set, struct output_set, 0);
+	set->rm_outs = tal_arr(set, struct output_set, 0);
+
+	return set;
+}
+
+#if EXPERIMENTAL_FEATURES
+static u8 *changeset_get_next(const tal_t *ctx, struct channel_id *cid,
+			      struct psbt_changeset **set)
```
rustyrussell, line 147: Does not need to take a `**`, can just take a `*`.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	if (tal_bytelen(redeemscript) > 0)
+		wit_prog = cast_const(u8 *, redeemscript);
+	else
+		wit_prog = cast_const(u8 *, wally_tx_output_get_script(tmpctx, output));
+
+	return is_p2wsh(wit_prog, NULL) || is_p2wpkh(wit_prog, NULL);
+}
+
+static struct wally_psbt *
+fetch_next_moves(struct state *state, const struct wally_psbt *psbt)
```
rustyrussell, line 592: fetch_next_moves seems a weird name.  fetch_next_psbt maybe?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+			    "Unable to determine next tx update");
+
+	if (psbt_has_diff(cs, *psbt, updated_psbt,
+			  &cs->added_ins, &cs->rm_ins,
+			  &cs->added_outs, &cs->rm_outs)) {
+
+		*psbt = tal_steal(state, updated_psbt);
+		msg = changeset_get_next(tmpctx, &state->channel_id,
+					 &state->changeset);
+		assert(msg);
```
rustyrussell, line 663: We don't need psbt_has_diff: call it psbt_get_changeset(), have it allocate and return the changeset, and then assign it to state->changeset.  It will be empty if no changes, so:

```C
state->changeset = tal_free(state->changeset);

state->changeset = psbt_get_changeset(state, *psbt, updated_psbt);
msg = changeset_get_next(tmpctx, &state->channel_id, &state->changeset);
if (msg)
     goto sendmsg;
```

rustyrussell: In fact, this simplifies entire logic: this entire branch goes under `if (!msg)` after `changeset_get_next`.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+	msg = towire_tx_complete(tmpctx, &state->channel_id);
+	finished = true;
+
+sendmsg:
+	sync_crypto_write(state->pps, msg);
+
+	return finished;
+}
+
+static bool find_txout(struct wally_psbt *psbt, const u8 *wscript, u16 *funding_txout) {
```
rustyrussell, line 679: Brace on next line please!

psbt is also const.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+					    type_to_string(tmpctx, struct bitcoin_tx, tx));
+
+			/*
+			 * BOLT- #2:
+			 *  - MUST NOT re-transmit inputs it has already received from the peer
+			 *  ...
+			 * - MUST fail the transaction collaboration if:
+			 *   ...
+			 *  - it receives a duplicate input to one it sent previously
+			 */
```
rustyrussell, line 899: Do we care?  Why?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+				peer_failed(state->pps, &state->channel_id,
+					    "Unable to add input");
+
+			/*
+			 * FIXME: do this
+			 * BOLT- #2:
+			 * - MUST fail the transaction collaboration if:
+			 *   ...
+			 * - it receives an unconfirmed input
+			 */
```
rustyrussell, line 911: Why?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+
+			/*
+			 * BOLT- #2:
+			 * The receiving node:
+			 *  - MUST add all received inputs to the funding transaction
+			 *
+			 *  ...
+			 *
+			  * - The input's sequence number will be set to 0xFDFFFFFF
+			  *   (little endian), which signals RBF.
```
rustyrussell, line 921: Assume this is fixed later?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+					    "Parsing tx_remove_input %s",
+					    tal_hex(tmpctx, msg));
+
+			check_channel_id(state, &cid, &state->channel_id);
+
+			for (size_t i = 0; i < psbt->num_inputs; i++) {
+				u16 input_serial;
+				if (!psbt_get_serial_id(&psbt->inputs[i].unknowns,
+							&input_serial)) {
+					peer_failed(state->pps, &state->channel_id,
```
rustyrussell, line 960: Wrong msg, isn't this an internal error if there's somehow no serial?

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+			if (serial_id % 2 != 0)
+				peer_failed(state->pps, &state->channel_id,
+					    "Invalid serial_id rcvd. %u", serial_id);
+			/*
+			 * BOLT-FIXME #2:
+			 * - MUST fail the transaction collaboration if:
+			 *   ...
+			 *  - it recieves a duplicate `serial_id`
+			 */
+			if (psbt_has_serial_input(psbt, serial_id))
```
rustyrussell, line 866: Instead, implement psbt_find_serial_input() which returns -1 if not found, otherwise an index.

This makes remove much simpler, too.

And similarly for _output.

openingd/dualopend.c
```diff
@@ -95,8 +123,1303 @@ struct state {
...
+				peer_failed(state->pps, &state->channel_id,
+					    "Duplicate serial_id rcvd. %u", serial_id);
+
+			/* Convert tx_bytes to a tx! */
+			len = tal_bytelen(tx_bytes);
+			tx = pull_bitcoin_tx(state, &tx_bytes, &len);
+			if (!tx)
+				peer_failed(state->pps, &state->channel_id,
+					    "Invalid tx sent.");
```
rustyrussell, line 875: If len != 0 also fail (implies junk at the end)

contrib/pyln-spec/bolt2/pyln/spec/bolt2/text.py
```diff
@@ -95,7 +95,7 @@
 - The `nLocktime` for the transaction has already been communicated, or is
   established via convention.
 - The transaction version is 2.
-- The each input's sequence number will be set to 0xFEFFFFFF (little endian), which signals RBF.
+- The each input's sequence number will be set to 0xFDFFFFFF (little endian), which signals RBF.
```
rustyrussell, line 98: This is not true, right?  And weird... 0xFFFFFFFE is the bitcoind default, and anyway this is now set as per tx_add_input?

And endian probably shouldn't be mentioned here anyway: you're describing the value, not the representation.

rustyrussell (changes requested):

wallet/db.c
```diff
@@ -1223,6 +1228,49 @@ void fillin_missing_scriptpubkeys(struct lightningd *ld, struct db *db,
 	tal_free(stmt);
 }
 
+/*
+ * V2 channel open has a different channel_id format than v1. prior to this, we
+ * could simply derive the channel_id whenever it was required, but since there
+ * are now two ways to do it, we save the derived channel id.
+ */
+static void fillin_missing_channel_id(struct lightningd *ld, struct db *db,
+				      const struct ext_key *bip32_base)
```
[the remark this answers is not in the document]
niftynei: > I'm sure you could implement derive_channel_id in SQL!

:worried:

openingd/dualopend.c
niftynei: because i was under the (mistaken, it appears) assumption that you can't assign to a const value after it's declaration. :upside_down_face:

openingd/dualopend.c
niftynei: it exits, which is a super-return. will update.

openingd/dualopend.c
niftynei: We get a new psbt but it may or may not be updated. Really we're fetching the updated set of changes. I like `fetch_psbt_changes`.

openingd/dualopend.c
niftynei: loops. the peer *should not* be sending us back inputs we already sent them.

openingd/dualopend.c
niftynei: This mattered when we weren't receiving the whole tx, as we looked up the tx on-chain to assure segwitness. Now that we're sending the whole tx, you're right, we can relax it.

contrib/pyln-spec/bolt2/pyln/spec/bolt2/text.py
niftynei: This should have gone away when I added the sequence to the spec, will update.

lightningd/dual_open_control.c
```diff
@@ -0,0 +1,1032 @@
...
+psbt_to_witness_stacks(const tal_t *ctx, struct wally_psbt *psbt, enum side opener)
+{
+	size_t stack_index;
+	u16 serial_id;
+	const struct witness_stack **stacks
+		= tal_arr(ctx, const struct witness_stack *, psbt->num_inputs);
+
+	/* Sort first so stacks are ordered correctly */
+	psbt_sort_by_serial_id(psbt);
+
```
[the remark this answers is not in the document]
niftynei: Good point -- if we permute the PSBT at this point it'll invalidate any signatures we've gotten. It shouldn't be updated/mutated by the signer. We can check for this.

rustyrussell (approved):

rustyrussell:
Hmm, Travis fails because it tries to build the dual_openingd wiregen file as !EXPERIMENTAL.

I'll think of something.

rustyrussell:
Let me now just do the rebase...

openingd/dualopend.c
```diff
@@ -95,8 +125,1188 @@ struct state {
...
+	bool finished = false;
+	struct wally_psbt *updated_psbt;
+	struct psbt_changeset *cs = state->changeset;
+
+	/* First we check our cached changes */
+	msg = changeset_get_next(tmpctx, &state->channel_id, cs);
+	if (msg)
+		goto sendmsg;
+
+	/* If we don't have any changes cached, go ask Alice for
```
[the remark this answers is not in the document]
niftynei: the plugin, in this case! lol

PR #3980, opened 2020-08-26
Minor invoice cleanups

rustyrussell:
Reuse the more complete incoming calculation we use in the JSON code to determine availability.

Changelog-None


rustyrussell:
@ZmnSCPxj See the final commit, which fixes a flake.  I think it's OK, as rotation is merely a heuristic...

lightningd/channel.h
```diff
@@ -138,7 +138,7 @@ struct channel {
 	struct command **forgets;
 
 	/* Our position in the round-robin list.  */
-	struct list_node rr_list;
+	u64 rr_number;
```
ZmnSCPxj, line 141: Cute trick, wish I had thought of that.

lightningd/lightningd.h
```diff
@@ -276,7 +276,7 @@ struct lightningd {
 	int *exit_code;
 
 	/* The round-robin list of channels, for use when doing MPP.  */
```
ZmnSCPxj, line 278: Comment deserves rewording?

tests/test_pay.py
```diff
@@ -3460,9 +3460,22 @@ def test_mpp_interference_2(node_factory, bitcoind, executor):
...
+    # 74094000, 52314000, 30318000, 19318000
+
+    # We *always* rotate through, since we have no published channels,
+    # but we can select badly and get an overlap. e.g. first invoice
+    # takes 30318000, 19318000 and 74094000.  Second will then take
+    # 52314000, and have to reuse 30318000, which gets exhausted by the
+    # first payer, thus leaving them unable to pay 66000000.
+
+    # So we re-do this until we only have 4 or fewer routehints.
+    while True:
```
ZmnSCPxj, line 3473: In principle this should be fixed in our running code, not in our test.

Suggest tagging a FIXME in our actual code pointing to this problem and say we have a minor sub-goal to remove this code in the test --- in principle we *should* eventually switch to something as smart as an estimator that will make this test pass without this extra code.

ZmnSCPxj (approved):
ACK 7cbe3cc307bb56b0b233b23cf887482df2f5b408

tests/test_pay.py
rustyrussell: I wasn't sure quite how to fix this.  We *could* run a fuzzy estimator of how much is remaining, but in practice that seems overkill for something which is so marginal.

lightningd/lightningd.h
rustyrussell: I left it because the spirit is correct.

tests/test_pay.py
ZmnSCPxj: Hmm.  If we agree that the issue is marginal (i.e. the issue only matters when the total incoming invoices is at or almost at the total incoming capacity) then maybe the proper test-fix is to increase the incoming channel capacities instead so that most of the invoices generated across all runs have 2 or fewer routehints.

ZmnSCPxj: i.e. if we are going to say "if the total value of invoices you are going to issue is near to the total incoming capacity you have, try increasing your incoming capacity instead", then the proper thing to do in the test is to increase the incoming capacities of the channels going to the payee.

PR #3981, opened 2020-08-26
gossipd: Make gossipd shut up

cdecker:
We are logging way too much from gossipd, causing noisy logs. This PR reduces
logs for incoming messages to those that actually caused a change in our
internal state (duplicate and old messages are just dropped silently now).

Most of the changes are related to tests breaking due to the reduced logging,
and the fixups required to get them working again. We fix them by polling for
the desired resulting state rather than observing the ephemeral state changes
represented by the logs. This may slow down some of the tests, since we now
poll, but that time is likely going to be used by concurrent tests anyway, and
we end up with more stable tests, that don't rely on implicit log ordering..

Changelog-Changed: gossipd: The `gossipd` is now a lot quieter, and will log only when a message changed our network topology.

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -759,18 +759,9 @@ class LightningNode(object):
                                 get_tx_p2wsh_outnum(self.bitcoin, tx, amount))
 
         if wait_for_active:
-            # We wait until gossipd sees both local updates, as well as status NORMAL,
-            # so it can definitely route through.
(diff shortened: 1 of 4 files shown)


rustyrussell (approved):
Ack 7c2a9f182f66c2245e12556ca86448a03d71ea28

For reference, in my logfile of 6.1M lines:
1. Ignoring redundant update: 2M lines
2. Ignoring redundant nannounce: 96k lines
3. Received channel_update: 2.9M lines
4. Ignoring update timestamp: 40k lines.

PR #3983, opened 2020-08-26
travis: Reorganize travis test configurations, add postgres config and elements config

cdecker:
We reorganize the tests slightly in order not to repeat the unit-tests under
all `valgrind` configurations, instead adding them to the lightweight source
checks. This saves us quite some CPU time overall, and giving the `valgrind`
configurations some more breathing room, without having to fork off yet
another config to the test matrix.

We then also add two lightweight smoke-test configs that use the `elementsd`
backend and another that uses `postgres` as the DB backend to avoid
regressions in the future.

Changelog-None


cdecker:
Having a bit of trouble with the new `NO_PYTHON` config:

```
  File "tools/generate-wire.py", line 24, in <module>
    from mako.template import Template
ModuleNotFoundError: No module named 'mako'
Makefile:268: recipe for target 'common/status_wiregen.c' failed
make: *** [common/status_wiregen.c] Error 1
  File "tools/generate-wire.py", line 24, in <module>
    from mako.template import Template
ModuleNotFoundError: No module named 'mako'
Makefile:268: recipe for target 'common/peer_status_wiregen.c' failed
make: *** [common/peer_status_wiregen.c] Error 1
```

It seems to try to generate the protocol implementations despite them being checked in, and as far as I can tell, up to date.

@rustyrussell am I doing something wrong here?

cdecker:
Seems to be all good now that the build issues have been addressed.

rustyrussell (approved):
Ack 05c77c01a31861963897bfba9de242485450c40c

PR #3985, opened 2020-08-26
blocksync: Some performance improvements on top of #3957

cdecker:
Inspired by #3957 I took some time to benchmark and improve the block sync
time as well. It's not as impressive as the 30x that #3957 achieved but it has
some incremental gains. The benchmarks use the following code, with a local
`bitcoind` and the default `bcli` backend:

```python3
from pyln.client import LightningRpc
import time
import subprocess

startheight = 600000
numblocks = 1000
targetheight = startheight + numblocks
proc = subprocess.Popen([
    'lightningd/lightningd',
    '--network=bitcoin',
    '--log-level=debug',
    '--rescan=-{}'.format(startheight),
    '--lightning-dir=/tmp/ll'
])
time.sleep(5)
rpc = LightningRpc('/tmp/ll/bitcoin/lightning-rpc')

def height():
    return rpc.getinfo()['blockheight']

def rate(first, last):
    if first == last:
        return 0
    return (last[1] - first[1]) / (last[0] - first[0])

start = (time.time(), height())
heights = []
while True:
    heights.append((time.time(), height()))
    if len(heights) > 30:
        heights = heights[-30:]
    first = heights[0]
    last = heights[-1]

    print("Speed: {:2f} blocks / s, overall {:2f} blocks / s".format(
        rate(first, last),
        rate(start, last)
    ))
    time.sleep(1)

    if last[1] >= targetheight:
        break

rpc.stop()
proc.wait()
```

The following are the performance gains achieved by individual commits (and compared to #3957):

| Scenario            | TTS 1000 blocks | Blocks / second | Speedup |
|---------------------|-----------------|-----------------|---------|
| pre-3957 (bf5e994)  |            8:01 |        2.073498 |       - |
| post-3957 (128adf0) |            7:28 |        2.224472 |      7% |
| commit 1 (68f3f30)  |            5:42 |        2.923359 |     41% |
| commit 2 (dc0859e)  |            5:00 |        3.321625 |     60% |
| commit 3 (0e4c6b6)  |            3:07 |        5.357775 |    158% |

And here are the flamegraphs that I used to guide the changes:
flamegraph-pre-3957
flamegraph-post-3957
flamegraph-68f3f30
flamegraph-dc0859e
flamegraph-0e4c6b6

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -491,30 +491,42 @@ static struct io_plan *plugin_read_json(struct io_conn *conn,
 					struct plugin *plugin)
 {
 	bool success;
+	bool have_full;
 
 	log_io(plugin->log, LOG_IO_IN, NULL, "",
 	       plugin->buffer + plugin->used, plugin->len_read);
+
+	/* Our JSON parser is pretty good at incremental parsing, but
+	 * `getrawblock` gives a giant 2MB token, which forces it to re-parse
+	 * every time until we have all of it. However, we can't complete a
+	 * JSON object without a '}', so we do a cheaper check here.
+	 */
+	have_full = memchr(plugin->buffer + plugin->used, '}',
+			   plugin->len_read);
 
 	plugin->used += plugin->len_read;
 	if (plugin->used == tal_count(plugin->buffer))
 		tal_resize(&plugin->buffer, plugin->used * 2);
 
 	/* Read and process all messages from the connection */
-	do {
-		bool destroyed;
-		const char *err;
-		err = plugin_read_json_one(plugin, &success, &destroyed);
+	if (have_full) {
+		do {
+			bool destroyed;
+			const char *err;
+			err =
+			    plugin_read_json_one(plugin, &success, &destroyed);
 
-		/* If it's destroyed, conn is already freed! */
-		if (destroyed)
-			return io_close(NULL);
+			/* If it's destroyed, conn is already freed! */
+			if (destroyed)
+				return io_close(NULL);
 
-		if (err) {
-			plugin_kill(plugin, err);
-			/* plugin_kill frees plugin */
-			return io_close(NULL);
-		}
-	} while (success);
+			if (err) {
+				plugin_kill(plugin, err);
+				/* plugin_kill frees plugin */
+				return io_close(NULL);
+			}
+		} while (success);
+	}
 
 	/* Now read more from the connection */
 	return io_read_partial(plugin->stdout_conn,
--- a/bitcoin/block.c
+++ b/bitcoin/block.c
@@ -208,9 +208,11 @@ bitcoin_block_from_hex(const tal_t *ctx, const struct chainparams *chainparams,
 
 	num = pull_varint(&p, &len);
 	b->tx = tal_arr(b, struct bitcoin_tx *, num);
+	b->txids = tal_arr(b, struct bitcoin_txid, num);
 	for (i = 0; i < num; i++) {
 		b->tx[i] = pull_bitcoin_tx(b->tx, &p, &len);
 		b->tx[i]->chainparams = chainparams;
+		bitcoin_txid(b->tx[i], &b->txids[i]);
 	}
 
 	/* We should end up not overrunning, nor have extra */
--- a/bitcoin/block.h
+++ b/bitcoin/block.h
@@ -36,6 +36,7 @@ struct bitcoin_block {
 	struct bitcoin_block_hdr hdr;
 	/* tal_count shows now many */
 	struct bitcoin_tx **tx;
+	struct bitcoin_txid *txids;
 };
 
 struct bitcoin_block *
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -91,7 +91,7 @@ static void filter_block_txs(struct chain_topology *topo, struct block *b)
 		}
 
 		owned = AMOUNT_SAT(0);
-		bitcoin_txid(tx, &txid);
+		txid = b->txids[i];
 		if (txfilter_match(topo->bitcoind->ld->owned_txfilter, tx)) {
 			wallet_extract_owned_outputs(topo->bitcoind->ld->wallet,
 						     tx->wtx, &b->height, &owned);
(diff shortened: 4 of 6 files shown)


lightningd/plugin.c
```diff
@@ -491,30 +491,41 @@ static struct io_plan *plugin_read_json(struct io_conn *conn,
...
 
 	log_io(plugin->log, LOG_IO_IN, NULL, "",
 	       plugin->buffer + plugin->used, plugin->len_read);
 
+	/* Peek into the buffer increment we just read (including a tiny
+	 * overlap if we're not at the start) to see if we have a '\n\n'
+	 * separator. If not we can skip the JSON attempt. */
+	have_full = memmem(plugin->buffer + plugin->used - offset,
+			   plugin->len_read + offset, "\n\n", 2);
+
```
rustyrussell, line 505: Plugins are *not* obliged to do a double CR.

This doesn't work.

rustyrussell (changes requested):

lightningd/plugin.c
rustyrussell: I also don't see how it would help, since plugin_read_json_one() is not O(n^2) any more?

cdecker: But it is, just look at the last two flamegraphs, with the `jsmn_parse_string` all but disappearing. I expect jsmn continues parsing from the last complete token, but will re-parse any token that was truncated. If we have really large fields (like a 2MB hex-encoded string) it'll parse that several times over and over.

I think we should make `\n\n` mandatory, just for this reason alone.

cdecker: Alternatively we can also match on `}\n` which might trigger a bit more often for nested structures (which as generally smaller) but is guaranteed to terminate a message ("a JSON object terminated by a newline").

cdecker:
Fixup switches from searching `\n\n` to searching `}\n` which makes no additional assumptions on the behavior of the plugins, instead using the existing "JSON object terminated by a newline" semantics.

cdecker:
Performance remains the same for block sync, if not better:

flamegraph-newsep

| Scenario            | TTS 1000 blocks | Blocks / second | Speedup |
|---------------------|-----------------|-----------------|---------|
| pre-3957 (bf5e994)  |            8:01 |        2.073498 |       - |
| post-3957 (128adf0) |            7:28 |        2.224472 |      7% |
| commit 1 (68f3f30)  |            5:42 |        2.923359 |     41% |
| commit 2 (dc0859e)  |            5:00 |        3.321625 |     60% |
| commit 3 (0e4c6b6)  |            3:07 |        5.357775 |    158% |
| new separator (1634f45)  |            3:01 |       5.609808  |    170% |

rustyrussell:
Hmm, my response to this got lost.

Anyway, I can't find *any* requirement on json.org or jsonrpc.org that a JSON RPC end with '\n'.  Even if it said that, my ignorance suggests it's dangerous to rely on it.

We can, however, insist that we have just read a '}'.  In the normal case, this is harmless, but in the case here "huge token", it's an effective hack.  Needs a huge comment though:

```C
/* Our JSON parser is pretty good at incremental parsing, but `getrawblock` gives a giant 2MB token, which forces it to re-parse every time until we have all of it.  However, we can't complete a JSON object without a '}', so we do a cheaper check here. */
```

cdecker:
> Hmm, my response to this got lost.
> 
> Anyway, I can't find _any_ requirement on json.org or jsonrpc.org that a JSON RPC end with '\n'. Even if it said that, my ignorance suggests it's dangerous to rely on it.

That is because the JSON-RPC spec is (mostly) transport agnostic, and the usual suspects (HTTP/HTTPS/HTTP2.0) provide built-in framing support, i.e., either the header contains the number of bytes to read, or the connection is closed after submitting the payload. In our case we don't have that which required us to come up with a different end-of-message marker, and so we did with a very inefficient end-of-json = end-of-message which requires either re-parsing the message on every increment, or having a resumable parser like `jsmn`. Notice that resumable parsers are very rare, and we got lucky. For consumers such as the `pyln-client` library we do provide stronger guarantees like the `\n\n` message separator, but I think we should slowly move to consolidate the different interfaces, and start requiring `\n\n` over time (gated behind a `COMPAT` flag).

I'm pretty sure that requiring a newline after the message is safe, even if not explicitly documented right now, but happy to start enforcing that only in conjunction with a `COMPAT` flag as well.

> We can, however, insist that we have just read a '}'. In the normal case, this is harmless, but in the case here "huge token", it's an effective hack. Needs a huge comment though:
> 
> ```c
>     /* Our JSON parser is pretty good at incremental parsing, but `getrawblock` gives a giant 2MB token, which forces it to re-parse 
>     every time until we have all of it.  However, we can't complete a JSON object without a '}', so we do a cheaper check here. */
> ```

I went ahead and implemented the single `}` check:

| Scenario            | TTS 1000 blocks | Blocks / second | Speedup |
|---------------------|-----------------|-----------------|---------|
| pre-3957 (bf5e994)  |            8:01 |        2.073498 |       - |
| post-3957 (128adf0) |            7:28 |        2.224472 |      7% |
| commit 1 (68f3f30)  |            5:42 |        2.923359 |     41% |
| commit 2 (dc0859e)  |            5:00 |        3.321625 |     60% |
| commit 3 (0e4c6b6)  |            3:07 |        5.357775 |    158% |
| new separator (1634f45)  |            3:01 |       5.609808  |    170% |
| single `}` sep (eea01b0721b7677f0d08eeea3bd715345f8f72a5) | 2:46 | 5.966785 | 188% |

<details>
flamegraph-newsep
</details>

So it seems like checking for `}` is just as good as the longer separator. At least we don't need to be more restrictive in what we accept yet :wink: 

I added the details on why we filter to the comment, and if this looks good I can squash it all into place.

lightningd/plugin.c
```diff
@@ -498,8 +498,14 @@ static struct io_plan *plugin_read_json(struct io_conn *conn,
...
+	 * overlap if we're not at the start) to see if we have a '}'
+	 * separator. If not we can skip the JSON attempt.
+	 *
+	 * Our JSON parser is pretty good at incremental parsing, but
+	 * `getrawblock` gives a giant 2MB token, which forces it to re-parse
+	 * every time until we have all of it. However, we can't complete a
+	 * JSON object without a '}', so we do a cheaper check here.
+	*/
 	have_full = memmem(plugin->buffer + plugin->used - offset,
 			   plugin->len_read + offset, "}", 1);
```
rustyrussell, line 510: This can be memchr :)

And we no longer need offset.  I'll add a fixup (and seems like a file needs regen, too...)

rustyrussell:
Changed to a straight memchr, squashed fixups and rebased on (Travis-fixed) master...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/e45b22d167f75f5689a939328425ced317966775

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3985/commits/64f2dcb01153aeed6fe0d7395f912b3b3f8c85a5

PR #3986, opened 2020-08-26
paymod: Scale presplit amounts to support larger payments while maintaining size buckets for privacy

### LINKED ISSUE #3926: MPP pay splits into too many subpayments

cdecker:
This proposal is a simple solution to #3926 and an alternative to
#3942/#3951. It is very much in the spirit of the original `presplit` payment
modifier in that it will create parts with a fixed size (not considering the
amount fuzzing to obfuscate the remainder), mostly independently of the total
amount. I think this is import to make it harder for observers to invert the
split and derive the original amount.

The change consists in adjusting the target amount for the individual parts
upwards should the current split result in too many parts. We use powers of 16
as a step size between buckets, resulting in at most 16 parts, and all
payments whose total amount is in the range of `x < amount <= 16*x` result in
the same part amount.

Why 16 as the upper limit of parts? Because it is approximately 1/2 of the
HTLCs available to a single channel, giving us some room to use adaptively
split should that be necessary. At the same time it is large enough such that
the resulting partial amounts are not too specific to the total amount, and
it's a power of two which means after log2(16) adaptive splits the amounts of
the parts mix with the a smaller payment's parts again. This is mostly in
preparation for PTLCs which give us decorrelation, but I think we can already
expect a benefit in amounts snapping to buckets.

Notice that none of the parameters are set in stone, and I'm happy to adjust
should better parameters be found.

Closes #3951
Closes #3942
Fixes #3926

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -2796,6 +2796,12 @@ REGISTER_PAYMENT_MODIFIER(waitblockheight, void *, NULL, waitblockheight_cb);
 #define MPP_TARGET_SIZE (10 * 1000 * 1000)
 #define PRESPLIT_MAX_HTLC_SHARE 3
 
+/* How many parts do we split into before we increase the bucket size. This is
+ * a tradeoff between the number of payments whose parts are identical and the
+ * number of concurrent HTLCs. The larger this amount the more HTLCs we may
+ * end up starting, but the more payments result in the same part sizes.*/
+#define PRESPLIT_MAX_SPLITS 16
+
 static struct presplit_mod_data *presplit_mod_data_init(struct payment *p)
 {
 	struct presplit_mod_data *d;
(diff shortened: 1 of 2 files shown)


plugins/libplugin-pay.c
```diff
@@ -2888,10 +2903,10 @@ static void presplit_cb(struct presplit_mod_data *d, struct payment *p)
 			return payment_fail(
 			    p, "Cannot attempt payment, we have no channel to "
 			       "which we can add an HTLC");
-		} else if (p->amount.millisatoshis / MPP_TARGET_SIZE > htlcs) /* Raw: division */
+		} else if (p->amount.millisatoshis / target_amount > htlcs) /* Raw: division */
 			target.millisatoshis = p->amount.millisatoshis / htlcs; /* Raw: division */
 		else
-			target = AMOUNT_MSAT(MPP_TARGET_SIZE);
+			target.millisatoshis = target_amount; /* Raw: init */
```
rustyrussell, line 2909: Simpler as `amount_msat(target_amount)`

plugins/libplugin-pay.c
```diff
@@ -2888,10 +2903,10 @@ static void presplit_cb(struct presplit_mod_data *d, struct payment *p)
 			return payment_fail(
 			    p, "Cannot attempt payment, we have no channel to "
 			       "which we can add an HTLC");
-		} else if (p->amount.millisatoshis / MPP_TARGET_SIZE > htlcs) /* Raw: division */
+		} else if (p->amount.millisatoshis / target_amount > htlcs) /* Raw: division */
 			target.millisatoshis = p->amount.millisatoshis / htlcs; /* Raw: division */
```
rustyrussell, line 2907: `target = amount_msat_div(p->amount, htlcs);`

rustyrussell (approved):
Ack 9239153ec8bab8c752479456590090c86e78ef0b

rustyrussell:
Minor cleanups only, will push post.

Issue #3989, opened 2020-08-27
test_mpp_interference_2 broken?


rustyrussell:
It's failing.  It's not failing more often because it's marked !VALGRIND and slow_test.

I've disabled it for now, please fix!

cdecker:
I added a fix to this in the `cleanups` branch, in PR #3996. The issue seems
to come down to a lack of synchronization to the blockchain, which results in
retries, which may cause an `MPP_TIMEOUT` on the recipient.

PR #3990, opened 2020-08-27
pytest: disable test_mpp_interference_2.

rustyrussell:
We're failing this too often: we'd fail it more but it's disabled
with VALGRIND (it shouldn't be: @slow_test removes VALGRIND if SLOW_MACHINE
is set).

See #3989


cdecker:
Tiny fixup ([range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/3990/range_diff/7da3d2f980a301db57dcd515a3a67fc89fb2c7d1..56c793bb3602970e542576dffbc888af7f43814c)), otherwise :+1:

PR #3992, opened 2020-08-28
contrib: start_ln now creates variable number of nodes

niftynei:
Re-write start_ln such that we can create up to 10 nodes locally for
testing. Useful for scenarios where more than two nodes are needed


rustyrussell (approved):
Ack bd091fb2b0eff06f15326ccec7cfbe3f9c3b58aa

PR #3995, opened 2020-08-28
configure: Use pg_config to locate the library location too

whitslack:
On some systems (e.g., Gentoo), there can be multiple major versions of the PostgreSQL client library installed in parallel. There exists a separate `pg_config` and `libpq.so*` for each major version. This Pull Request adds support to C-Lightning's build system to allow the user to choose a specific version of the PostgreSQL client library to build and link against.

**Summary of proposed changes:**

* Introduce a `POSTGRES_LDLIBS` config variable, analogous to the existing `POSTGRES_INCLUDE` variable, and use it in place of `-lpq`, both in `Makefile` and in the Configurator script.
* When running `configure`, allow the user to override the name of `pg_config` by setting the `PG_CONFIG` environment variable. If unset, the default continues to be `pg_config`, which preserves the existing behavior.
* The above changes have the happy side-effect that setting `PG_CONFIG` to an empty string will now cause Configurator's test for PostgreSQL to fail, so C-Lightning will be built without PostgreSQL support, even if a PostgreSQL client library is present on the build system. (The Gentoo ebuild has been hacking the Configurator test code by inserting an `#error` directive to force the test to fail when the user has requested that C-Lightning not support PostgreSQL. That hack is no longer necessary with the changes in this PR.)

### DIFF
--- a/configure
+++ b/configure
@@ -230,8 +230,10 @@ fi
 require 'python3-mako' "You need the mako module for python3: see doc/INSTALL.md" python3 -c 'import mako'
 
 POSTGRES_INCLUDE=""
-if command -v pg_config 2> /dev/null; then
-    POSTGRES_INCLUDE="-I$(pg_config --includedir)"
+POSTGRES_LDLIBS=""
+if command -v "${PG_CONFIG-pg_config}" >/dev/null; then
+    POSTGRES_INCLUDE="-I$("${PG_CONFIG-pg_config}" --includedir)"
+    POSTGRES_LDLIBS="-L$("${PG_CONFIG-pg_config}" --libdir) -lpq"
 fi
 
 rm -f $CONFIG_VAR_FILE.$$
@@ -290,7 +292,7 @@ int main(void)
 var=HAVE_POSTGRES
 desc=postgres
 style=DEFINES_EVERYTHING|EXECUTE|MAY_NOT_COMPILE
-link=-lpq
+link=$POSTGRES_LDLIBS
 code=
 #include <libpq-fe.h>
 #include <stdio.h>
@@ -360,6 +362,7 @@ add_var CWARNFLAGS "$CWARNFLAGS"
 add_var CDEBUGFLAGS "$CDEBUGFLAGS"
 add_var COPTFLAGS "$COPTFLAGS"
 add_var POSTGRES_INCLUDE "$POSTGRES_INCLUDE"
+add_var POSTGRES_LDLIBS "$POSTGRES_LDLIBS"
 add_var VALGRIND "$VALGRIND"
 add_var DEVELOPER "$DEVELOPER" $CONFIG_HEADER
 add_var EXPERIMENTAL_FEATURES "$EXPERIMENTAL_FEATURES" $CONFIG_HEADER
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -230,8 +230,10 @@ fi
 require 'python3-mako' "You need the mako module for python3: see doc/INSTALL.md" python3 -c 'import mako'
 
 POSTGRES_INCLUDE=""
-if command -v pg_config 2> /dev/null; then
-    POSTGRES_INCLUDE="-I$(pg_config --includedir)"
+POSTGRES_LDLIBS=""
+if command -v "${PG_CONFIG-pg_config}" >/dev/null; then
```
whitslack, line 234: Note, I fixed the redirection here. (`command -v` prints to stdout, not stderr.)

configure
```diff
@@ -230,8 +230,10 @@ fi
...
+    POSTGRES_INCLUDE="-I$("${PG_CONFIG-pg_config}" --includedir)"
+    POSTGRES_LDLIBS="-L$("${PG_CONFIG-pg_config}" --libdir) -lpq"
```
cdecker, line 236: We could set the `$PG_CONFIG` at the top of the script to avoid having to use the get-or-default construct here.

cdecker (commented):
> * The above changes have the happy side-effect that setting `PG_CONFIG` to an empty string will now cause Configurator's test for PostgreSQL to fail, so C-Lightning will be built without PostgreSQL support, even if a PostgreSQL client library is present on the build system. (The Gentoo ebuild has been hacking the Configurator test code by inserting an `#error` directive to force the test to fail when the user has requested that C-Lightning not support PostgreSQL. That hack is no longer necessary with the changes in this PR.)

This applies only to users explicitly setting it to an empty string, not the default behavior, correct? I.e., if the user has the libs installed and doesn't customize the `PG_CONFIG` envvar, then postgresql support will be built in. Just want to make sure that that's still the default behavior :-)

whitslack:
> This applies only to users explicitly setting it to an empty string, not the default behavior, correct? I.e., if the user has the libs installed and doesn't customize the `PG_CONFIG` envvar, then postgresql support will be built in. Just want to make sure that that's still the default behavior :-)

You have to explicitly set it to empty if you want to forcibly disable PostgreSQL support even though you have the PostgreSQL library installed. If you leave `PG_CONFIG` unset, then you'd get the same automagic detection that the existing script implements.

whitslack:
(Actually you can set `PG_CONFIG` to any nonexistent path name to disable PostgreSQL support. It just so happens that `command -v ''` fails too, so empty is a useful value for this purpose.)

configure
whitslack: I've already left the computer for the night (going to bed), but I'll make this change within the day.

cdecker:
> You have to explicitly set it to empty if you want to forcibly disable PostgreSQL support even though you have the PostgreSQL library installed. If you leave `PG_CONFIG` unset, then you'd get the same automagic detection that the existing script implements.

Excellent, thanks for putting my mind at ease :-)

configure
whitslack: [Implemented.](https://github.com/ElementsProject/lightning/pull/3995/commits/a29632dfbf4353b4868489a9a5e324e3da89a8e6)

rustyrussell (approved):
Ack a29632dfbf4353b4868489a9a5e324e3da89a8e6

PR #3997, opened 2020-08-28
IO-level logging for plugins

darosior:
Just small things that i found useful when debugging a plugin crash:
- Allow plugins to log as `io` (i was pretty sure it was allowed)
- Log incoming requests in libplugin and pyln, we log in lightningd/plugin the requests coming from the plugins. However we cannot log the outgoing ones in lightningd/plugin without moving around the `jout` struct: instead, just log the incoming requests from the point of view of the plugin.

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -505,9 +505,12 @@ class Plugin(object):
         Returns the last partial message that was not complete yet.
         """
         for payload in msgs[:-1]:
+            raw_req = payload.decode("utf8")
+            self.log(raw_req, level="io")
+
             # Note that we use function annotations to do Millisatoshi conversions
             # in _exec_func, so we don't use LightningJSONDecoder here.
-            request = self._parse_request(json.loads(payload.decode('utf8')))
+            request = self._parse_request(json.loads(raw_req))
 
             # If this has an 'id'-field, it's a request and returns a
             # result. Otherwise it's a notification and it doesn't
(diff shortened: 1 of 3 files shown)


lightningd/plugin.c
```diff
@@ -284,6 +284,8 @@ static const char *plugin_log_handle(struct plugin *plugin,
 		level = LOG_UNUSUAL;
 	else if (json_tok_streq(plugin->buffer, leveltok, "error"))
 		level = LOG_BROKEN;
+	else if (json_tok_streq(plugin->buffer, leveltok, "io"))
+		level = LOG_IO_IN;
```
rustyrussell, line 288: These levels are special, you need to call log_io() not log_.  And you need a direction, so you should probably make the strings 'io-in' and 'io-out'?

contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -505,9 +505,12 @@ def _multi_dispatch(self, msgs):
         Returns the last partial message that was not complete yet.
         """
         for payload in msgs[:-1]:
+            raw_req = payload.decode("utf8")
+            self.log(raw_req, level="io")
```
rustyrussell, line 509: Hmm, doesn't that mean logging twice, since we log from the lightningd side?  And won't we be slowing things down a *lot* to do this?

I expect io logging for plugins which tallk to other things (e.g. webservices(, but they'll have to wire it up themselves...

darosior: Regarding logging twice, we don't log it lightningd side.

Regarding the slowdown, maybe this could be made optional ?

darosior: Just realized that it does not condition the write on the `io` level setting.. Right, this may not be a brilliant idea.

PR #3999, opened 2020-08-30
Allow building without SQLite3

whitslack:
**Summary of proposed changes:**

* Introduce `SQLITE3_CFLAGS` and `SQLITE3_LDLIBS` config variables, whose values are automagically populated by calling out to `pkg-config`. If `pkg-config` is not found, then the new variables are defaulted as `SQLITE3_CFLAGS=` and `SQLITE3_LDLIBS=-lsqlite3`, which allows `configure` to continue working as before on systems lacking `pkg-config`.
    * When running `configure`, allow the user to override the name of `pkg-config` by setting the `PKG_CONFIG` environment variable. If unset, the default is `pkg-config`, which means to use the system default (i.e., in `$PATH`) copy of `pkg-config`. Cross-compiling users would set the `PKG_CONFIG` environment variable to the path of a `pkg-config` binary that provides information about packages built for the target system.
* Pass `$SQLITE3_CFLAGS` to Configurator, and add it to `CFLAGS` in `Makefile`.
* Pass `$SQLITE3_LDLIBS` in place of `-lsqlite3`, both in `Makefile` and in the Configurator script.
* Tweak a few source files so that they compile successfully when `HAVE_SQLITE3=0`.
* Omit `devtools/checkchannels` from build when `HAVE_SQLITE3=0`, as that tool requires SQLite3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -205,7 +205,7 @@ BOLT_DEPS := $(BOLT_GEN)
 ALL_PROGRAMS =
 
 CPPFLAGS += -DBINTOPKGLIBEXECDIR="\"$(shell sh tools/rel.sh $(bindir) $(pkglibexecdir))\""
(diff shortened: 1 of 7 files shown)


rustyrussell (approved):
Ack 38cd1a8573482836018d61ddb1430d45584aafcf

Nice work!

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/3999/commits/38cd1a8573482836018d61ddb1430d45584aafcf

PR #4000, opened 2020-08-31
libplugin-pay: fix default CLTV.

rustyrussell:
This was changed recently, but without a bolt quote, we didn't find
this.


niftynei:
Needs changelog?

rustyrussell:
> Needs changelog?

We have one for the prev PR:

    Changelog-Changed: config: the default CLTV expiry is now 34 blocks, and final expiry 18 blocks as per new BOLT recommendations.

PR #4007, opened 2020-09-04
Cleanups to derived resouces

cdecker:
If I understand the `maintainer` / `developer` split correctly it should be possible for a developer not to have all the python dependencies to generate some of the derived files that are now checked in. Contrary to this the generated DB files where cleaned up with `make clean`, so I moved them to the `maintainer-clean` target instead (and fixed the `wallet-maintainer-clean` target which had a typo).

I then added the derived files to the `.gitattributes` file which will collapse these derived diffs in the Github interface, to cut down on he noise when reviewing changes.

Finally, `maintainer-clean` was deleting the `deployable-lightning.pdf` file which would not get regenerated, and has heavy dependencies to rebuild. I removed that cleanup of a document that hasn't seen any changes for years.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,4 +7,8 @@ configure text eol=lf
 # The following files are generated and should not be shown in the
 # diffs by default on Github.
 doc/lightning*.7 linguist-generated=true
+wallet/db_*_sqlgen.c linguist-generated=true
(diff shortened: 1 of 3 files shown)


rustyrussell (approved):
Ack e39dc0ec293afa3499049d37b22053a022c0ae30

PR #4010, opened 2020-09-06
Added proxy information to plugin configuration

vincenzopalazzo:
Hi all!

With this PR, I'm proposing my idea to add a couple of the information to the proxy during the plugin configuration.

Inside my test to develop a new plugin, I noted the necessity to know if the node has the proxy enabled, the solution was to add a plugin option but I think this is redundant propriety during the node setup!

Another solution is to use the `listconfigs` to set up the connection inside the plugin but in some cases the method plugin call before to start some plugin methods! For instance, an alternative backend bitcoin/element that overrides the plugin `bcli` 

I hope that I'm losing nothings.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -214,6 +214,14 @@ simple JSON object containing the options:
         "channel": "",
         "invoice": "028200"
     }
+    "startup": true,
+    "proxy": {
+        "type": "ipv4",
+        "address": "127.0.0.1",
+        "port": 9050
+    }
+    "torv3-enabled": true,
+    "use_proxy_always": false
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4010/commits/f45c7726233c4f24c4161c18db1cf86f8c3a2328

rustyrussell:
Please add a "Changelog-Added: Plugins: Proxy information now provided in `init.configuration`." to the first commit (and maybe squash them all into one commit?)

Thanks!

vincenzopalazzo:
@rustyrussell Done!

Thanks for review this!

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4010/commits/e0e82a3f1d3b81eefe43f8ba049c2c9f88625f64

PR #4011, opened 2020-09-07
log: align level tags IO DEBUG INFO and UNUSUAL

m-schmoock:
This aligns all log TAGs except \*\*BROKEN\*\* which would require too many spaces for all the other logs.
This makes reading the logs with human eyes easier as the log message payload always starts at the same character of a line.

Additionally, I suggest replacing "UNUSUAL" with "WARN" and "\*\*BROKEN\*\*" with "FATAL" which seems more standard and is shorter.

Changelog-None


darosior (approved):
ACK 47b2634a03f40301a1e0d9a2d341706dece97987

Nice!

PR #4014, opened 2020-09-07
Use the DB utxoset to detect spends for gossipd

cdecker:
This all started with the big idea of not keeping a large (~140k entries) UTXO
filter in memory, along with all the associated overhead in the hope of
reducing the memory footprint, but failed miserably. The savings come down to
about 4MB in RSS, but the block sync loses almost 20% of speed (from a
baseline of 5.26 blocks/s down to 4.16 blocks/s).

I still like some of the changes, like simplifying the detection of spends
based solely on the DB and looking at the `spendheight`, allowing us to do
some other changes, like batching the notification to `gossipd` instead of
notifying for each spent short-channel-id individually (not yet implemented).

I'll let the review gods decide the fate of this one :wink:

<details>

## Baseline

Maximum resident set size (kbytes): 226884
speed 5.260800463705923 blocks / second

## Combining the batch notification with the in-memoy filter

Maximum resident set size (kbytes): 226376
speed 5.41619091963417 blocks / second

## DB-only utxoset

Maximum resident set size (kbytes): 222228
speed 4.166378023473524 blocks / second

</details>

----

#

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -741,7 +741,7 @@ log_fee:
  */
 static void topo_update_spends(struct chain_topology *topo, struct block *b)
 {
-	const struct short_channel_id *scid;
+	const struct short_channel_id *spent_scids;
 	for (size_t i = 0; i < tal_count(b->full_txs); i++) {
 		const struct bitcoin_tx *tx = b->full_txs[i];
 		bool our_tx = true, includes_our_spend = false;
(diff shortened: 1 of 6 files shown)


rustyrussell (approved):
Ack 3d65c9e4886397cfb1e0ab7c261a98a7992333cb

Nice simplification, even if it does cause a slight slowdown.

cdecker:
Actually that second to last commit removed one of the major speed blocks for on-disk only utxoset: querying for the spent scid is no longer needed in the tight loop. Removing that we get back most of the performance. Will add the on-disk-only commit again :-)

rustyrussell:
Rebasing now...

cdecker:
Rebased and added last commit which removes the in-memory set. Finalizing the benchmarks, but it's looking good for now.

cdecker:
Ok, with the last commit, which removes the in-memory filter we indeed get the same performance as before:

Maximum resident set size (kbytes): 222316
speed 5.172151624192423 blocks / second

Turns out the select statement was causing the vast majority of the slowdown. Not limitiing to entries that we knew were present meant that we went from 20 select queries to 4000 select queries which was slow, whereas going from 20 update to 4000 update statements seems not to have the same effect. My best guess is that the select was flushing the updates to disk, whereas keeping them in flight until the batch select at the end removed the perfomance loss.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/6bf81db18e89d93ae76f7faf889f6c76f31a7a6c

PR #4015, opened 2020-09-07
Fix EXPERIMENTAL_FEATURES interference with checked-in files.

rustyrussell:


niftynei (approved):
ACK 98f96ab86d349744700047432c1d320bbbbb57b4

PR #4019, opened 2020-09-08
Avoid compile error on macos mojave

Moller40:
Fixing the following error by changing 'enum feerate' to int.

lightningd/bitcoind.c:183:29: error: result of comparison of constant 8 with expression of type 'enum feerate' is always true [-Werror,-Wtautological-constant-out-of-range-compare]
        for (enum feerate f = 0; f < NUM_FEERATES; f++) {

### DIFF
--- a/lightningd/bitcoind.c
+++ b/lightningd/bitcoind.c
@@ -180,7 +180,7 @@ static void estimatefees_callback(const char *buf, const jsmntok_t *toks,
 				     "estimatefees",
 				     "bad 'result' field");
 
-	for (enum feerate f = 0; f < NUM_FEERATES; f++) {
+	for (int f = 0; f < NUM_FEERATES; f++) {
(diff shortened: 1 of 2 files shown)


rustyrussell:
No, the compiler is *wrong* here.  (It's true for C++, not C).

My preferred fix is to change `f` to an `int`, i.e.:

```C
for (int f = 0; f < NUM_FEERATES; f++) {
```

rustyrussell (changes requested):
Please just change f to an int, check if that fixes compile?

Moller40:
Yes, changing f to an int also fixes it.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4019/commits/ea61f4867bd192680a3b9325802fa71da944dd0f

rustyrussell:
Thanks!

Issue #4018, opened 2020-09-08
feature request: add plugin event notification channel_state_changed


m-schmoock:
## Summary
Currently there's no hook or notification that triggers a plugin when a channel `state` is changed.
Since this can be useful for monitoring plugins and such we might want to have this.

### Example Format
```json
{
    "channel_state_changed": {
        "peer_id": "03bc9337c7a28bb784d67742ebedd30a93bacdf7e4ca16436ef3798000242b2251",
        "channel_id": "a2d0851832f0e30a0cf778a826d72f077ca86b69f72677e0267f23f63a0599b4",
        "short_channel_id" : "561820x1020x1",
        "old_state": "CHANNELD_NORMAL",
        "new_state": "AWAITING_UNILATERAL"
    }
}
```

PR #20, opened 2016-03-31
fix compile issue: missing braces around initializer

bitkevin:
If you compile with gcc(version 4.8.2), it'll cause failure. Error is "missing braces around initializer".

PR #4020, opened 2020-09-08
feat: notification channel_state_changed

m-schmoock:
## Summary
This adds a new plugin notification message `channel_state_changed` that is called when a channel changes its state.

This PR contains code, test and doc.
Addresses #4018 

## Example message
```json
{
    "channel_state_changed": {
        "peer_id": "03bc9337c7a28bb784d67742ebedd30a93bacdf7e4ca16436ef3798000242b2251",
        "channel_id": "a2d0851832f0e30a0cf778a826d72f077ca86b69f72677e0267f23f63a0599b4",
        "short_channel_id" : "561820x1020x1",
        "old_state": "CHANNELD_NORMAL",
        "new_state": "CHANNELD_SHUTTING_DOWN"
    }
}
```

## Future thoughts
@fiatjaf added the question if it would be possible to have some kind of a 'reason' for a state changed operation. From a user or plugin developer perspective, this is mostly the distinction if a channel was closed by remote or by the user or for other (internal) reasons. We could add an string indication to this event, i.e. `"reason" : "USER"` or `"reason" : "REMOTE"`  or `"reason" : "OTHER"` . Also, this only makes sens for state changes from CHANNELD_NORMAL to some CLOSING state. Maybe AWAITING_LOCKIN can also be caused by the user if its a local funding thats ongoing.

## Note
When unilaterally closing a channel it happens that the state changes two times to the same value `AWAITING_UNILATERAL`.
I don't think this is intended, but current test reflect that.

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -11,6 +11,7 @@ from utils import (
     check_coin_moves, first_channel_id, check_coin_moves_idx, EXPERIMENTAL_FEATURES
 )
 
+import ast
 import json
 import os
 import pytest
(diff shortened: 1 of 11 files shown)


fiatjaf:
How impossible would it be to have the reason of a channel state change? Is there such a concept?

m-schmoock:
> How impossible would it be to have the reason of a channel state change? Is there such a concept?

I think it would not be nice code-wise. Maybe this can be added after the basic functionality is done.

cdecker:
> How impossible would it be to have the reason of a channel state change? Is there such a concept?

We currently only produce logs with the specifics, but we should be able to enumerate the possible causes and pass them in. I think this is a great idea, both for this hook as well as for the status in `listpeers`.

cdecker:
Opened #4027 to track this, and brainstorm how these details could look like.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4020/commits/cfdeab6c2748ffb79fabd14245fed831197ab708

tests/test_plugin.py
```diff
@@ -631,6 +632,130 @@ def test_openchannel_hook_chaining(node_factory, bitcoind):
...
+
+    bitcoind.generate_block(100)  # so it gets settled
+
+    msg = l2.daemon.wait_for_log("channel_state_changed.*new_state.*")
+    event = ast.literal_eval(re.findall(".*({.*}).*", msg)[0])
+    assert(event['peer_id'] == peer_id)
+    assert(event['channel_id'] == cid)
+    assert(event['short_channel_id'] == scid)
+    assert(event['old_state'] == "AWAITING_UNILATERAL")  # this actually happens
+    assert(event['new_state'] == "AWAITING_UNILATERAL")  # note: same states
```
rustyrussell, line 740: OK, let's filter these out for the moment (i.e. don't send an event if it's a noop).

We should probably find and kill the place this is calling, but not sure how invasive that would be?

rustyrussell (changes requested):

tests/test_plugin.py
m-schmoock: k. will make that change soooonish

m-schmoock: addressed by 6fc4f2f

cdecker:
LGTM

ACK https://github.com/ElementsProject/lightning/pull/4020/commits/6fc4f2f63ca43b645f3429f82f310f1e117201c5

m-schmoock:
Pipeline #20 sometimes ends up in infinite output termination....

rustyrussell (approved):

rustyrussell (approved):
Ack 6fc4f2f63ca43b645f3429f82f310f1e117201c5

PR #4021, opened 2020-09-08
Repro builds

cdecker:
- Added the release link to the changelog tool since I kept forgetting it...
 - Added ubuntu 20.04 as a repro-build config
 - Rewrote the repro-build documentation to the new docker-based system.

One thing we have though is the Fedora build which is included in the `SHA256SUMS` file but is not built reproducibly (yet), so we should likely not list it in the manifest at all, otherwise we might get complaints that the hashes do not match. What do you think @rustyrussell?

The huge advantage we have with docker is that this can all be automated now, and we can setup a nightly build machine that can feed into a repository.

### DIFF
--- a/devtools/changelog.py
+++ b/devtools/changelog.py
@@ -116,7 +116,9 @@ ${group(groups['removed']) | trim}
 ${h3} Fixed
 ${group(groups['fixed']) | trim}
 ${h3} Security
-${group(groups['security']) | trim}""")
+${group(groups['security']) | trim}
+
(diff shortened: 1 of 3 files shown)


rustyrussell (changes requested):
Can we add a 

Changelog-Added: New reproducible build system now uses docker: try it at home with doc/REPRODUCIBLE.md!

cdecker:
> Can we add a
> 
> Changelog-Added: New reproducible build system now uses docker: try it at home with doc/REPRODUCIBLE.md!

Will do :+1:

rustyrussell (approved):

PR #4025, opened 2020-09-09
contrib: add ability to start/stop elements nodeset

niftynei:
sometimes it's handy to be able to get access to an elements daemon and
nodeset also

Changelog-None.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4025/commits/e71ef796d11afd71dbdc14f5a35ac4c704483e3d

PR #4013, opened 2020-09-07
Re-enable ASAN when building with clang

darosior:
I'm looking at including some fuzz targets, and will probably use clang along with its address sanitizer (and hopefully many others!).

This removes the restriction on only using gcc with ASAN, by using suppressions instead when using clang (see #2285 for the reason).
This also enables ASAN on non-already-under-valgrind Travis jobs.

PR #4030, opened 2020-09-09
4013 fixups

darosior:
The fixups from #4013 (a mem fix detected by ASAN and a build cleanup), so they are part of 0.9.1 even if it takes more time to make Travis happy with ASAN builds.

### DIFF
--- a/lightningd/channel_control.c
+++ b/lightningd/channel_control.c
@@ -735,10 +735,10 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
 	/* Peer could have errored out while we were waiting */
 	peer = peer_by_id(bitcoind->ld, &cc->peer);
 	if (!peer)
-		return;
+		goto cleanup;
 	cancel = find_channel_by_id(peer, &cc->cid);
 	if (!cancel)
-		return;
+		goto cleanup;
 
 	if (txout != NULL) {
 		for (size_t i = 0; i < tal_count(cancel->forgets); i++)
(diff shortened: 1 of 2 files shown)


darosior:
Looks like i cannot ask for review directly, so ping @rustyrussell

lightningd/channel_control.c
```diff
@@ -748,20 +748,24 @@ static void process_check_funding_broadcast(struct bitcoind *bitcoind,
...
+	tal_free(cc);
+	return;
 }
 
 struct command_result *cancel_channel_before_broadcast(struct command *cmd,
 						       struct peer *peer)
 {
 	struct channel *cancel_channel;
-	struct channel_to_cancel *cc = tal(cmd, struct channel_to_cancel);
+	struct channel_to_cancel *cc = tal(NULL, struct channel_to_cancel);
```
cdecker, line 768: Doesn't this now leak if we return with any of the `command_fail` clauses below? If the goal is to have this survive `cmd` it should likely be stolen onto `NULL` and marked as `notleak()` on line 826 instead.

cdecker (changes requested):
Leaks `cc` in `cancel_channel_before_broadcast`.

lightningd/channel_control.c
darosior: Argh good catch thanks..

darosior:
Thanks, amended

cdecker (approved):

PR #4031, opened 2020-09-09
pyln: Add safe fallback results for hooks

### LINKED ISSUE #3748: pyln-client plugin: Exception in htlc_accepted hook handler causes payment to hang indefinitely
It appears that an exception in an `htlc_accepted` hook handler (on receiver side) causes payments to hang indefinitely. This seems to be due to that hook expecting errors in a different format (which is correctly documented [here](https://lightning.readthedocs.io/PLUGINS.html#htlc-accepted)), and the Plugin returns the error as {"error":"error contents"}.

I believe other hooks fail on the c-lightning side when unexpected results are returned -- possibly the same could be done here?

Expected result: payment should fail if `htlc_accepted` hook handler raises an exception.

Actual result: payment hangs indefinitely

#### Repro Plugin
Here is a simple plugin and tests that should reproduce this issue:

Plugin:

```python
#!/usr/bin/env python3
from pyln.client import Plugin

plugin = Plugin()
  
@plugin.hook("htlc_accepted")
def on_htlc_accepted(plugin: Plugin, **kwargs):
  action = plugin.get_option("action")
  if action == "hang":
    # The `pay` rpc will hang forever (as far as I can tell)
    raise ValueError("Error!")
  elif action == "fail":
    # This is how the plugin expects a failure to look -- indeed the payment will fail.
    return {
      "result": "fail",
      "failure_message": "2002"
    }
  elif action == "succeed":
    # The payment will succeed.
    return {"result": "continue"}

plugin.add_option('action', "fail", "Whether to 'fail', 'succeed' or 'hang' the payment")

plugin.run()

```

Tests:

```python
from pyln.testing.fixtures import *

pluginopts = {'plugin': os.path.join(os.path.dirname(__file__), "fail.py")}

def test_fail(node_factory: NodeFactory):
    l1, l2 = node_factory.get_nodes(2, {**pluginopts, **{"action": "fail"}})

    l1.connect(l2)
    l1.openchannel(l2, 1_000_000)
    invoice = l2.rpc.invoice(9000000, "test", "test")
    with pytest.raises(Exception):
      l1.rpc.pay(invoice["bolt11"])

def test_succeed(node_factory: NodeFactory):
    l1, l2 = node_factory.get_nodes(2, {**pluginopts, **{"action": "succeed"}})

    l1.connect(l2)
    l1.openchannel(l2, 1_000_000)
    invoice = l2.rpc.invoice(9000000, "test", "test")
    l1.rpc.pay(invoice["bolt11"])

def test_hang(node_factory: NodeFactory):
    l1, l2 = node_factory.get_nodes(2, {**pluginopts, **{"action": "hang"}})

    l1.connect(l2)
    l1.openchannel(l2, 1_000_000)
    with pytest.raises(Exception):
      l1.rpc.pay(invoice["bolt11"])
```

#### Results
The first two tests succeed, but the last test hangs for (as far as I can tell) forever

```
fail/test_fail.py::test_fail PASSED                                      [ 33%]
fail/test_fail.py::test_succeed PASSED                                   [ 66%]
fail/test_fail.py::test_hang
```

cdecker:
Hooks do not tolerate failures at all. If we return a JSON-RPC error to a hook
call the only thing the main daemon can really do is to crash. This commit
adds a mapping of error to a safe fallback result, including a warning to the
node operator that this should be addressed in the plugin. The warning is
reported as a `**BROKEN**` message, and should therefore fail any testing done
on the plugin.

The alternatives would have been

 - Allow plugins to return errors on hook calls, requiring a similar mapping inside `lightningd`. But I don't like relaxing the requirements on this. It feels magic enough as it is, but prevents us from accidentally hanging due to a plugin.
 - Killing the plugin outright, which is basically just going to use the default action, which in case of security related hooks is bad.

Fixes #3748

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -1743,7 +1743,6 @@ def test_dev_builtin_plugins_unimportant(node_factory):
     n.rpc.plugin_stop(plugin="pay")
 
 
-@pytest.mark.xfail(strict=True)
 def test_htlc_accepted_hook_crash(node_factory, executor):
     """Test that we do not hang incoming HTLCs if the hook plugin crashes.
 
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -1741,3 +1741,33 @@ def test_important_plugin(node_factory):
 def test_dev_builtin_plugins_unimportant(node_factory):
     n = node_factory.get_node(options={"dev-builtin-plugins-unimportant": None})
     n.rpc.plugin_stop(plugin="pay")
+
+
+@pytest.mark.xfail(strict=True)
+def test_htlc_accepted_hook_crash(node_factory, executor):
+    """Test that we do not hang incoming HTLCs if the hook plugin crashes.
+
+    Reproduces #3748.
+    """
+    plugin = os.path.join(os.getcwd(), 'tests/plugins/htlc_accepted-crash.py')
+    l1 = node_factory.get_node()
+    l2 = node_factory.get_node(
+        options={'plugin': plugin},
+        allow_broken_log=True
+    )
+    l1.connect(l2)
+    l1.fund_channel(l2, 10**6)
+
+    i = l2.rpc.invoice(500, "crashpls", "crashpls")['bolt11']
+
+    # This should still succeed
+
+    f = executor.submit(l1.rpc.pay, i)
+
+    l2.daemon.wait_for_log(r'Crashing on purpose...')
+    l2.daemon.wait_for_log(
+        r'Hook handler for htlc_accepted failed with an exception.'
+    )
+
+    with pytest.raises(RpcError, match=r'failed: WIRE_TEMPORARY_NODE_FAILURE'):
+        f.result(10)
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -97,6 +97,26 @@ class Request(dict):
         self.plugin._write_locked(result)
 
 
+# If a hook call fails we need to coerce it into something the main daemon can
+# handle. Returning an error is not an option since we explicitly do not allow
+# those as a response to the calls, otherwise the only option we have is to
+# crash the main daemon. The goal of these is to present a safe fallback
+# should the hook call fail unexpectedly.
+hook_fallbacks = {
+    'htlc_accepted': {
+        'result': 'fail',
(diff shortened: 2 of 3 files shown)


tests/test_plugin.py
```diff
@@ -1741,3 +1741,32 @@ def get_logfile_match(logpath, regex):
 def test_dev_builtin_plugins_unimportant(node_factory):
     n = node_factory.get_node(options={"dev-builtin-plugins-unimportant": None})
     n.rpc.plugin_stop(plugin="pay")
+
+
+def test_htlc_accepted_hook_crash(node_factory, executor):
+    """Test that we do not hang incoming HTLCs if the hook pluin crashes.
```
niftynei, line 1747:
```suggestion
    """Test that we do not hang incoming HTLCs if the hook plugin crashes.
```

tests/test_plugin.py
```diff
@@ -1741,3 +1741,32 @@ def get_logfile_match(logpath, regex):
...
+    # This should still succeed
+
+    f = executor.submit(l1.rpc.pay, i)
+
+    l2.daemon.wait_for_log(r'Crashing on purpose...')
+    l2.daemon.wait_for_log(
+        r'Hook handler for htlc_accepted failed with an exception.'
+    )
+
+    with pytest.raises(RpcError, match=r'failed: WIRE_TEMPORARY_NODE_FAILURE'):
```
niftynei, line 1771: should we also test for one of the 'non-recoverable' hooks, `db_write` or `commitment_revocation`

niftynei (commented):
ACK b7e3829a5316aa45c15192c61d88e13962468d1b

cdecker:
> ~ACK [b7e3829](https://github.com/ElementsProject/lightning/commit/b7e3829a5316aa45c15192c61d88e13962468d1b)~
> 
> > '/home/travis/build/ElementsProject/lightning/tests/plugins/htlc_accepted-crash.py': opening pipe: No such file or directory
> 
> Looks like you forgot to check in `htlc_accepted-crash.py`

Good catch, fixed up :-)

tests/test_plugin.py
cdecker: Sure I can add a test for it :-)

rustyrussell:
Ack 7f0db3431d67ef60a3a9081f484d2fec723eb327

Issue #3968, opened 2020-08-23
keysend only working with other c-lightning nodes


Fonta1n3:
I have been testing `keysend` and it quite reliably only seems to work with other c-lightning nodes.

This recipient for example has keysend enabled and is able to send me keysends but I am unable to return the favor and generally get an error:

```
error =     {
        "amount_msat" = 1000msat;
        "amount_sent_msat" = 0msat;
        code = 203;
        "created_at" = 1598181774;
        destination = 02f7dcc28e15c0b55238a0f7db02b2082b1c2fa5adf66ff96e79db8f7ae17fe7d3;
        "erring_index" = 4;
        "erring_node" = 02f7dcc28e15c0b55238a0f7db02b2082b1c2fa5adf66ff96e79db8f7ae17fe7d3;
        failcode = 16399;
        failcodename = "WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS";
        id = 343;
        message = "failed: WIRE_INCORRECT_OR_UNKNOWN_PAYMENT_DETAILS (reply from remote)";
        msatoshi = 1000;
        "msatoshi_sent" = 0;
        "payment_hash" = 1887be8ddb26b3336494899f0a58d068f4343ba4160c38d05ee28986426744ca;
        "raw_message" = 400f00000000000003e80009d777;
        status = pending;
    }
```

rustyrussell:
Hmm.  Keysend is undocumented, so perhaps we implemented it wrong? @cdecker

cdecker:
Indeed there is no official specification for keysend, so I had to reverse engineer it. It was working for a while, but I can't guarantee it hasn't changed on the lnd side. I'll see what I can do 👍

d4amenace:
Don't know if this is helpful, but I was trying to do a keysend to my Eclair Mobile wallet version 0.4.12 and it to is failing.

Fonta1n3:
From my experience it literally only works c-lightning to c-lightning although i haven’t tested since latest release.

cdecker:
The error message got me thinking that the recipient might not support keysend at all, and checking with my local gossip (which might look different now), it doesn't seem the destination supports it:

```json
{
  "nodeid": "02f7dcc28e15c0b55238a0f7db02b2082b1c2fa5adf66ff96e79db8f7ae17fe7d3",
  "alias": "Satoshi Nakamoto",
  "color": "000000",
  "last_timestamp": 1606207493,
  "features": "02a2a1",
[...]

PR #4035, opened 2020-09-10
pay: fix handling of legacy vs tlv encoding.

rustyrussell:
As revealed by the failure of tests in #3936, where we ended up trying
to send a partial payment using legacy style, we are not handling
style properly.

1. BOLT9 has features, so we can *know* that the destination supports
   MPP.  We may not have seen a node_announcement.
2. We can't assume that nodes inside routehints support TLV.
3. We can't assume direct peers support TLV.

The keysend code tried to fix this up, so I'm not sure that this caused
the issue in #3968, though.


cdecker:
ACK 06cc8384bb7908e580995a05b51824b94ec119e2

PR #4037, opened 2020-09-10
pyln: Bump proto version to match pyln-bolt*

cdecker:
Since pyln-bolt* specify the 0.8.4 version which we didn't upload, and the
requirements.txt specify `==0.8.4`, we need to backfill that version, even if we
could just bump it directly to 0.9.1.

Changelog-None


rustyrussell (approved):
Ack 8907062c9fec8784dd1f0602d9027f11d90d9d62

PR #4039, opened 2020-09-11
Connection timeout

rustyrussell:
Found this because multifundchannel hung.  Want it for rc2.

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -593,6 +593,9 @@ static void dev_register_opts(struct lightningd *ld)
 				 "Make builtin plugins unimportant so you can plugin stop them.");
 	opt_register_arg("--dev-force-features", opt_force_featureset, NULL, ld,
 			 "Force the init/globalinit/node_announce/channel/bolt11 features, each comma-separated bitnumbers");
+	opt_register_arg("--dev-timeout-secs", opt_set_u32, opt_show_u32,
+			 &ld->config.connection_timeout_secs,
+			 "Seconds to timeout if we don't receive INIT from peer");
 }
 #endif /* DEVELOPER */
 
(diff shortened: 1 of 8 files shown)


lightningd/options.c
```diff
@@ -694,6 +700,9 @@ static const struct config mainnet_config = {
 
 	/* Allow to define the default behavior of tor services calls*/
 	.use_v3_autotor = true,
+
+	/* 10 minutes should be enough for anyone! */
+	.connection_timeout_secs = 600,
```
darosior, line 705: This is *very* conservative ? I'd rather not start funding a channel with a peer who took 9 minutes to respond to init..

tests/test_connection.py
```diff
@@ -2634,3 +2634,18 @@ def test_nonstatic_channel(node_factory, bitcoind):
...
     l1.rpc.close(l2.info['id'])
+
+
+@unittest.skipIf(not DEVELOPER, "needs --dev-timeout-secs")
+def test_connection_timeout(node_factory):
+    # l1 hears nothing back after sending INIT, should time out.
+    l1, l2 = node_factory.get_nodes(2,
+                                    opts=[{'dev-timeout-secs': 1,
+                                           'disconnect': ['0WIRE_INIT', '0WIRE_INIT']},
+                                          {}])
```
darosior, line 2645: Interesting, i wondered why you disconnected l1 to test the timeout on l1 and started testing the other way around: this would not trigger a timeout but a connection reset error.

Why specifying two times though ?

darosior (approved):
ACK 8d336e6a340d5408b540aa8925535244136a88c0

Nice ! I'd be in favor of a more realistic timeout but not urgent for this release i guess

cdecker (approved):
10 minutes sounds way too long, but better start with a relaxed default 👍

ACK 8d336e6a340d5408b540aa8925535244136a88c0

tests/test_connection.py
rustyrussell: Wanted to test incoming timeout didn't corrupt or leak memory.  But this doesn't actually do that (since the incoming fails with Connection reset as soon at it hits the dev-disconnect line.

Added some minor logging so we can make sure it actually fired.

PR #4040, opened 2020-09-11
pyln: Sets remote_pubkey for LightningConnection if node is not the initiator

sr-gi:
In the current version of `LightningConnection`,  `remote_pubkey` is set to `None` at the begining of the handshake if the node is not initiator. However, it is never set to the counterparty pubkey after the handshake is completed. 

This sets `remote_pubkey` to ``rs`` in act three from the receiver side so the pubkey of the initiator is known by the receiver after the handshake.


cdecker (commented):
ACK 8f804de68824a6dae95a9a82ee9fcef06e4fcc4f

cdecker (approved):

PR #4047, opened 2020-09-14
feat: log to stdout if not --daemon

m-schmoock:
### Summary
If `--daemon` is not set, we should also log to stdout, regardless of the `log_file` config settings.

### Rationales:
 - This matches (at least my) expectation, I found it strange that the `stdout` output without `--daemon` was suppressed because of certain values of in the default user config file laying around.
 - Better for `systemd` scripts and such, that usually catch `stdout` per default.
 - Better for debugging environments inside IDE's and such.

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -1,6 +1,4 @@
 #include "log.h"
-#include <backtrace-supported.h>
-#include <backtrace.h>
 #include <ccan/array_size/array_size.h>
 #include <ccan/err/err.h>
 #include <ccan/htable/htable_type.h>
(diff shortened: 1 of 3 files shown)


rustyrussell:
Hmm, this is not my expectation at all!

IMHO what we should do is:

* If nothing specified, log to stdout.
* If log-file, log to that file.
* If specified more than once, log to each one
* File "-" means stdout.

For bonus points, support different log levels using level: prefix :)

m-schmoock:
I was just messing around...

PR #4053, opened 2020-09-15
Sort input filenames to SHA256STAMP

whitslack:
Make's `wildcard` function returns matching file paths in an unpredictable order, thereby causing false mismatches in the produced SHA256STAMPs from system to system. This pull request sorts the file paths given to `cat` to make the stamps deterministic.

This change is crucial if you really want people to be able to build C-Lightning without having Mako installed.


cdecker (approved):
ACK bcfdcaa0e634da668ddeb3151a1af46f635950aa

rustyrussell:
It's weird, I thought about this but couldn't find anywhere it was a problem.

Even so, this is harmless, so applied!

Thanks!

whitslack:
> It's weird, I thought about this but couldn't find anywhere it was a problem.


@rustyrussell: I noticed it while working on the Gentoo ebuilds for C-Lightning. My main system would build 0.9.1 without running `generate-wire.py`, but a chroot system I have for testing the stable Gentoo environment was always running `generate-wire.py`. I noticed that the only changes to the generated sources were the SHA256STAMPs, and I had a hunch about the ordering of the dependencies, so I sha256sum'd every permutation of the dependency files and found both the checked-in SHA256STAMP and the one that my build had stamped, so this told me my hunch was right.

Issue #4043, opened 2020-09-12
Time out on non-responsive opens


rustyrussell:
### Issue and Steps to Reproduce
On my attempt to open channels to all of testnet, two nodes simply failed to respond to open_channel.

As a result, the multifundchannel stalled until I liked those openingd manually.  We should probably use our same 1 minute connect timeout for completing funding

ZmnSCPxj:
Maybe this should be implemented at `openingd`.  Determine an absolute time at which we fail the connection (from now + relative timeout).  Then after we send the `open_channel`, additionally wait for this timeout --- this means creating variants of `opening_negotiate_msg` and the common `peer_or_gossip_sync_read` that *additionally* have an absolute time limit.  This gets complicated for `peer_or_gossip_sync_read` as we need to know the lower of the gossip time limit and the given time limit and act accordingly.  If the given time limit is reached then we just fail and exit.

In fact we probably do not need a timeout variant for `opening_negotiate_msg`, that function should always require a timeout, since the remote might never respond to *any* of our opening messages (`funding_created` -> `funding_signed`, `accept_channel`->`funding_created`), not just `open_channel` -> `accept_channel`.

SrijanShovit:
Hi @rustyrussell Is this issue still open for contribution or anything that I could try?

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: Very old issue (5.4 years old) no recent activity

If you believe this issue should remain open, please comment and we can reopen it.

PR #4055, opened 2020-09-16
Opener side of the draft dual-funding spec

niftynei:
A few more commits, to make the opener side of the dual-funding piece work.

The opener side requires 3 new RPC calls:
 - openchannel_init
 - openchannel_update
 - openchannel_signed

Each of these take a PSBT. An opener initiates an v2 open by calling `openchannel_init` with a PSBT that includes enough funding for their desired feerate and a funding output, which will be added to the provided PSBT with the amount the caller specified plus however much the peer adds. This method will return a PSBT with partial additions from the peer.

Next, you'll update the PSBT with any additions/changes you have, then call `openchannel_update` with the updated (or identical) PSBT. Once both sides have stopped updating the PSBT, `openchannel_update` will return with `commitments_secured` set to true. (Until this point it will return `commitments_secured` as false).

The next step is to pass a PSBT with partial_signatures for every input you're contributing to this funding transaction to `openchannel_signed`, which will collect the transaction signatures from the peer, build the final transaction and broadcast it.

For a working example of this see `tests/plugins/df_opener.py`.

FIXME: the df_opener plugin uses `wallycore` which isn't part of the build/pip install process, which is why the provided test is flagged off. @jgriffiths any word on when it might appear on pypi?

~FIXME: needs docs for new RPCs~


jgriffiths:
> @jgriffiths any word on when it might appear on pypi?

We have an internal issue open for this, in the meantime you can use something like the following in your requirements.txt file - this is for wally 0.7.8, change the urls and update/remove the hashes for 0.7.9. You can add more platform checks to use the binary builds, this example will build wally from source only for for non linux/osx boxen.

````
https://github.com/ElementsProject/libwally-core/releases/download/release_0.7.8/wallycore-0.7.8-cp36-cp36m-linux_x86_64.whl; 'linux' in sys_platform and python_version == '3.6' --hash=sha256:b973fbabf7c30952d6ab5f626ea5231f06200c5f4f35f07bea19308ab87c8162
https://github.com/ElementsProject/libwally-core/releases/download/release_0.7.8/wallycore-0.7.8-cp37-cp37m-linux_x86_64.whl; 'linux' in sys_platform and python_version == '3.7' --hash=sha256:77bebee316219c4be5069722e9aa08043278934641c7f677627194b4dfeabfd0
https://github.com/ElementsProject/libwally-core/releases/download/release_0.7.8/wallycore-0.7.8-cp37-cp37m-macosx_10_14_x86_64.whl; 'darwin' in sys_platform and python_version == '3.7' --hash=sha256:bae6656aa86ed1197e636f69905e10987ef4fd3ee8ba18d16e0f62fcd8aaaaf1
https://github.com/ElementsProject/libwally-core/archive/release_0.7.8.tar.gz --hash=sha256:b36a9e4e7db155d5a435b43c08e74ed9d812b05ea2becce454700a18217c2296
````

I think you probably want to move to requiring hashes for your python deps FWIW.

File: doc/lightning-fundpsbt.7.md [hunk 1]
```diff
@@ -46,18 +46,18 @@ Let's assume the caller is trying to produce a 100,000 satoshi output.
...
-It calls "*fundpsbt* 100000sat slow 206", which succeeds, and returns
+It calls "*fundpsbt* 100000sat slow 166", which succeeds, and returns
 the *psbt* and *feerate_per_kw* it used, the *estimated_final_weight*
 and any *excess_msat*.
 
 If *excess_msat* is greater than the cost of adding a change output,
 the caller adds a change output randomly to position 0 or 1 in the
 PSBT.  Say *feerate_per_kw* is 253, and the change output is a P2WPKH
-(weight 164), that would cost the cost is around 41 sats.  With the
+(weight 124), that would cost the cost is around 31 sats.  With the
```
jgriffiths, line 58: `s/that would cost //g`

niftynei: moved over to #4059

lightningd/peer_control.c
```diff
@@ -324,6 +335,72 @@ register_close_command(struct lightningd *ld,
...
+	assert(channel);
+
+	oc = tal(cmd, struct open_command);
+	list_add_tail(&ld->open_commands, &oc->list);
+	oc->cmd = cmd;
+	oc->channel = channel;
+	tal_add_destructor(oc, &destroy_open_command);
+	tal_add_destructor2(channel,
+			    &destroy_open_command_on_channel_destroy,
+			    oc);
```
rustyrussell, line 398: No, this is the wrong level!  Just remove the timeout for now.

There should be a timeout on the *channel construction* which does this, and more; actually broadcasts the unilateral or simply forgets (if we have no inputs).  This is a general fix which solves #4043 too!

doc/lightning-openchannel_signed.7.md
```diff
@@ -0,0 +1,65 @@
...
+DESCRIPTION
+-----------
+
+`openchannel_signed` is a low level RPC command which concludes a channel
+open with the specified peer. It uses the v2 openchannel protocol, which
+allows for interactive transaction construction.
+
+This command should be called after a call to `openchannel_update`
+with a return value that includes a true *commitments_secured*.
+
```
rustyrussell, line 18: IOW, "should be called after `openchannel_update` returns *commitments_secured* `true`."

doc/lightning-openchannel_update.7.md
```diff
@@ -0,0 +1,59 @@
+lightning-openchannel\_update -- Command to update a collab channel open
+========================================================================
+
+SYNOPSIS
+--------
+
+**openchannel_update** *id* *psbt*
```
rustyrussell, line 7: Maybe these APIs should use channelid (in which case, figure out how we know that?)

lightningd/dual_open_control.c
```diff
@@ -980,6 +982,134 @@ static void accepter_got_offer(struct subd *dualopend,
...
+	/* Add serials to PSBT */
+	psbt_add_serials(psbt, LOCAL);
+	if (!psbt_has_required_fields(psbt))
+		return command_fail(cmd, FUNDING_PSBT_INVALID,
+				    "PSBT is missing required fields %s",
+				    type_to_string(tmpctx, struct wally_psbt, psbt));
+
+	peer->uncommitted_channel->fc->cmd = cmd;
+
+	memset(&chan_id_unused, 0, sizeof(chan_id_unused));
```
rustyrussell, line 1104: ?

channeld/channeld.c
```diff
@@ -1977,6 +1977,73 @@ static void send_onionmsg(struct peer *peer, const u8 *msg)
...
+
+	if (!fromwire_tx_signatures(tmpctx, msg, &cid, &txid, &ws))
+		peer_failed(peer->pps,
+			    &peer->channel_id,
+			    "Bad tx_signatures %s", tal_hex(msg, msg));
+
+	/* Check that we're in the right spot for this channel to have gotten
+	 * this message */
+	if (peer->funding_locked[LOCAL] || peer->funding_locked[REMOTE]) {
+		/* FIXME: should we fail here instead? */
```
rustyrussell: Yes, please do.  Or at very least, log BROKEN so our test harness will pick it up if we hit it!

channeld/channeld.c
```diff
@@ -1977,6 +1977,73 @@ static void send_onionmsg(struct peer *peer, const u8 *msg)
...
+			       "for channel %s, ignoring: %s",
+			       type_to_string(tmpctx, struct channel_id,
+					      &peer->channel_id),
+			       tal_hex(tmpctx, msg));
+		return;
+	}
+
+	if (!peer->psbt) {
+		/* FIXME: should we fail here instead? */
+		status_unusual("Got WIRE_TX_SIGNATURES with no PSBT "
```
rustyrussell, line 2028: Here too

lightningd/peer_control.h
```diff
@@ -90,6 +99,10 @@ void channel_watch_funding(struct lightningd *ld, struct channel *channel);
 
 struct amount_msat channel_amount_receivable(const struct channel *channel);
 
+/* Find the open command that was registered for this channel */
+struct open_command *find_open_command(struct lightningd *ld,
+				       struct channel *channel);
```
rustyrussell, line 104: const struct channel *?

lightningd/channel_control.c
```diff
@@ -315,6 +315,97 @@ static void handle_error_channel(struct channel *channel,
...
+		if (oc)
+			was_pending(command_fail(oc->cmd,
+						 FUNDING_BROADCAST_FAIL,
+						 "Error broadcasting funding "
+						 "tx: %s. Unsent tx discarded "
+						 "%s.",
+						 msg,
+						 type_to_string(tmpctx,
+								struct wally_tx,
+								wtx)));
```
rustyrussell, line 356: We probably should log this for posterity...

channeld/channeld.c
```diff
@@ -2886,7 +2926,6 @@ static void handle_funding_depth(struct peer *peer, const u8 *msg)
 		peer->short_channel_ids[LOCAL] = *scid;
 
 		if (!peer->funding_locked[LOCAL]) {
-
```
rustyrussell, line 2815: This level of subtlety *requires* a comment!

Suggest: /* This line intentionally left blank */

common/psbt_open.c
```diff
@@ -434,6 +435,32 @@ int psbt_find_serial_output(struct wally_psbt *psbt, u16 serial_id)
...
+{
+	const u64 serial_space = 10000000;
+	u16 serial_id;
+
+	while ((serial_id = pseudorand(serial_space)) % 2 != role ||
+		psbt_find_serial_input(psbt, serial_id) != -1) {
+		/* keep going; */
+	}
+
+	return serial_id;
```
rustyrussell, line 448: How about `(pseudorand(1 << 15) << 1) | role`?  Yeah, we still need to loop, and if will get inefficient once we've got about 65,000 inputs, but I think that's impossible anyway.

rustyrussell: Even better, make function `get_random_serial(role)` and reuse below...

tests/plugins/df_opener.py
```diff
@@ -0,0 +1,86 @@
...
+
+    funding = plugin.rpc.fundpsbt(amount, feerate, funding_output_weight + core_weight)
+    psbt_obj = psbt_from_base64(funding['psbt'])
+
+    excess = Millisatoshi(funding['excess_msat'])
+    # FIXME: convert feerate ?!
+    change_cost = Millisatoshi(change_output_weight * feerate_val // 1000 * 1000)
+    dust_limit = Millisatoshi(feerate_val * 1000)
+    if excess > (dust_limit + change_cost):
+        addr = plugin.rpc.newaddr()['bech32']
```
rustyrussell, line 62: I see a new return val from newaddr (`scriptpubkey`?) in our future....

tests/test_connection.py
```diff
@@ -819,6 +819,26 @@ def test_funding_toolarge(node_factory, bitcoind):
     l1.rpc.fundchannel(l2.info['id'], amount)
 
 
+@unittest.skipIf(True, "df_opener.py requires wallycore")
```
rustyrussell, line 822: Probabably also skipIf not EXPERIMENTAL...

rustyrussell (approved):
Mainly minor stuff, but more than enough to keep you busy I'm sure!

doc/lightning-openchannel_update.7.md
niftynei: Good idea, will update!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4055/commits/66f1475cf6445aba6601c029f43f2aa2d0a3f146

common/psbt_open.c
```diff
@@ -471,3 +471,17 @@ bool psbt_has_required_fields(struct wally_psbt *psbt)
 
 	return true;
 }
+
+#if EXPERIMENTAL_FEATURES
+void psbt_input_set_final_witness_stack(struct wally_psbt_input *in,
+					struct witness_element **elements)
```
rustyrussell, line 507: const struct witness_element **?

common/psbt_open.h
```diff
@@ -14,6 +14,9 @@ struct wally_psbt;
 struct wally_psbt_input;
 struct wally_psbt_output;
 struct wally_map;
+#if EXPERIMENTAL_FEATURES
+struct witness_element;
+#endif /* EXPERIMENTAL_FEATURES */
```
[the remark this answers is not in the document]
rustyrussell: #if still unnecessary.

lightningd/dual_open_control.c
```diff
@@ -923,6 +952,155 @@ static void accepter_got_offer(struct subd *dualopend,
...
+
+#if EXPERIMENTAL_FEATURES
+	if (!feature_negotiated(cmd->ld->our_features,
+			        peer->their_features,
+				OPT_DUAL_FUND)) {
+		return command_fail(cmd, FUNDING_V2_NOT_SUPPORTED,
+				    "v2 openchannel not supported "
+				    "by peer");
+	}
+#endif /* EXPERIMENTAL_FEATURES */
```
rustyrussell, line 1223: Don't need this #if, since this entire file is experimental-only.

rustyrussell (approved):
Ack beeb6bcd423dac3febe9f12c81963260207ea5bf

Trivial issues only, but you need to fold fixmes anyway.

common/psbt_open.c
niftynei: sort of a rough thing because the struct these are embedded in is generated code and doesn't include the const (but probably should?).

PR #4058, opened 2020-09-16
memleak: 'generify' the memleak calling to code for openingd/dualopend

niftynei:
Switch on name of subd(aemon) as to whether to call dualopend or
openingd for memleak results


rustyrussell (approved):
Ack 862217a369cc9a53c1d92d052482e9ced367c725

PR #4059, opened 2020-09-16
doc and contrib fixes/nits

niftynei:
Update the docs for `fundpsbt` to have the correct numbers; make bitcoin polling and gossip propagation faster for the regtest utility daemons.


rustyrussell (approved):
Ack b6adb28bbfc5e038488fcdd3d4c99ecc655e2a7c

startup_regtest now requires developer build, but that's probably OK.

PR #4060, opened 2020-09-16
listpeers: show correct amounts for starting funding allocation

niftynei:
With dual-funded channels, it's possible for the funding allocation of
the non-opener to not be zero.


rustyrussell (commented):
Ack a94d8a58a7efcfb16072f035e4c17a90a7f861f5

PR #4061, opened 2020-09-16
small dual-funding related nits

niftynei:
A bunch of small memory errors/leaks/documentation nits/spacing fixes for dual-funded related code.


lightningd/dual_open_control.c
```diff
@@ -679,7 +679,9 @@ wallet_commit_channel(struct lightningd *ld,
 		return NULL;
 	}
 
-	channel_info->fee_states = new_fee_states(uc, opener, &feerate);
+	channel_info->fee_states = take(new_fee_states(uc,
+						       opener,
+						       &feerate));
```
rustyrussell, line 683: That is almost certainly wrong.  Were you thinking `tal_steal`?  In which case, change `uc` to `channel_info`.

rustyrussell (changes requested):
Just one weirdness needs fixing.

I also flinch when I see 'fixup' since that has a specific git meaning, prehaps 'cleanup'.  Or more commonly, label the part you're changing, eg 'lightningd:'

lightningd/dual_open_control.c
niftynei: we call `new_channel` below, which if `taken`, will properly tal_steal it onto the newly created `channel` instance. updated to make this clearer, and set the context to `NULL`.

niftynei:
> I also flinch when I see 'fixup' since that has a specific git meaning, prehaps 'cleanup'. Or more commonly, label the part you're changing, eg 'lightningd:'

Great point, updated the comments to not have 'fixup' in the name.

rustyrussell:
Trivial rebase

rustyrussell:
Trivial whitespace fix...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4061/commits/d00ad9ea10a45e6ee5df8d26b47e3f2193026e50

rustyrussell (approved):

Issue #4054, opened 2020-09-16
Build error on ppc64le


grubles:
### Issue and Steps to Reproduce
Running `make` on Ubuntu 18.04 ppc64le encounters an error.

```
ld lightningd/lightningd
wallet/db.o:(.toc+0x0): undefined reference to '__stop_xautodata_db_backends'
wallet/db.o:(.toc+0x8): undefined reference to '__start_xautodata_db_backends'
collect2: error:  ld returned 1 exit status
Makefile:481: recipe for target 'lightningd/lightningd' failed
```

rustyrussell:
Did you somehow compile with neither sqlite3 nor postgres support?

grubles:
I did indeed not have `libsqlite3-dev` installed. I figured `./configure` would yell at me for missing a dependency.

After installing `libsqlite3-dev`, the build completed successfully.

PR #4063, opened 2020-09-17
configure: refuse to compile without any database.

rustyrussell:
Fixes: #4054
Changelog-None

### DIFF
--- a/configure
+++ b/configure
@@ -165,7 +165,7 @@ add_var()
     else
         echo "$1 not found"
     fi
-    echo "$1=$2" >> $CONFIG_VAR_FILE
+    echo "$1=$2" >> $CONFIG_VAR_FILE.$$
     [ -z "$3" ] || echo "#define $1 $2" >> "$3"
 }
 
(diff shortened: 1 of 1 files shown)


cdecker:
Best error message eva!

ACK https://github.com/ElementsProject/lightning/pull/4063/commits/7725f6c89a39724ee6f17728dfef265ab5ad064d

whitslack:
Is there no way to have `configure` complain about this rather than `make`? (I would have implemented such a check in my PR to make SQLite3 optional, but I didn't see any straightforward way to access the variables emitted by Configurator from within the `configure` script.)

rustyrussell:
There is... Let's do that then.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4063/commits/b76e158633cbd5abfe980e85059eb0ad060c05b9

configure
```diff
@@ -396,3 +394,16 @@ add_var HAVE_PYTHON3_MAKO "$HAVE_PYTHON3_MAKO"
...
+
+# Now we set them all and check.
+while IFS='=' read VAR VAL; do
+    eval $VAR=\"$VAL\"
+done < $CONFIG_VAR_FILE.$$
+
+if [ "$HAVE_SQLITE3" = 0 -a "$HAVE_POSTGRES" = 0 ]; then
+    # I have no database yet I must schema!)
+    echo "*** We need a database, but neither sqlite3 nor postgres found" >&2
+    exit 1
```
whitslack, line 406: Exiting here would leave the temporary file `$CONFIG_VAR_FILE.$$` in place. Is that what you want?

rustyrussell: Oops, I assumed there was a "trap" to clean this up (there are other cases we can exit early, due to `set -e`).  Will fix that separately...

PR #4064, opened 2020-09-18
db: Fix size mismatch on postgres in a migration

### LINKED ISSUE #4050: Startup failure "s32 field doesn't match size: expected 4, actual 8" with PostgreSQL

Attempting to start C-Lightning 0.9.1 using the PostgreSQL backend results in the following fatal error:
```
s32 field doesn't match size: expected 4, actual 8
```

Tracing with GDB reveals that this message is coming from the new `fillin_missing_channel_id` function at [line 1256](https://github.com/ElementsProject/lightning/blob/v0.9.1/wallet/db.c#L1256).

The `channels.id` column is of type `BIGSERIAL` (i.e., `BIGINT`), but the [`db_postgres_column_int`](https://github.com/ElementsProject/lightning/blob/v0.9.1/wallet/db_postgres.c#L167) function enforces that the column width is 4 bytes. This obviously fails.

### Workaround
1. Change the `channels.id` column type to `INTEGER`.
    ```
    ALTER TABLE channels ALTER COLUMN id TYPE INTEGER;
    ```
1. Start lightningd so that the database upgrade can occur. The process will eventually fail to start with the following fatal error:
    ```
    u64 field doesn't match size: expected 8, actual 4
    ```
1. Change the `channels.id` column type back to `BIGINT`.
    ```
    ALTER TABLE channels ALTER COLUMN id TYPE BIGINT;
    ```
1. Start lightningd as normal.

cdecker:
Fixes #4050


fiatjaf:
I don't understand what is happening, but this fix works and saved my life.

rustyrussell (approved):
Ack bdab7febe03c96561eebe5aa7cd6d7d3a05ec000

rustyrussell:
Pushed regen fix

cdecker:
> I don't understand what is happening, but this fix works and saved my life.

It's a minor detail that we are checking for the postgres DB backend: postgres has well-defined datatypes, each with a given size on disk. This is different from sqlite3 which aliases all integer types to it's single, variable sized, int type.

To protect against truncation we check that we access fields of a given size only with the matching accessors. In this case one slipped through: we were reading a u64 with an accessory for s32, which doesn't have any impact in this case, but better be safe than sorry 🙂

Issue #3717, opened 2020-05-08
`hsm_secret` generation from a seed-phrase 


positiveblue:
Hello,

I saw that c-lightning was participating Lightning HackSprint May 2020 and I was interested in helping with one of the "challenges". 

The goal of my contribution would be `the creation of a launcher or plugin that takes a seed-phrase and generates the associated hsm_secret file.`

I am pretty new on the topic so I would like to summarize what I understood after doing a bit of research + reading part of the c-lightning code and try to move forward from there.

When a new lightning node starts up, it launches a daemon ([hsmd.c](https://github.com/ElementsProject/lightning/blob/master/hsmd/hsmd.c)) which creates/loads an `hms_secret file`. That file contains the 256 bits (32 bytes) used for creating the master private key of a HD Wallet as it is specified in the [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki). During the creation of the key the bytes are randomly generated using libsodium's cryptographic randomness routine. By now there is no way to derivative those random bytes from a seed in order to backup the wallet.

The idea is to implement some process that allow the node to derive those 32 bytes using a seed-phrase like input.

There are some widely approaches to this like using a 24 word list as specified in the [BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki). However, other lightning implementations like [lnd](https://github.com/lightningnetwork/lnd)  use a similar approach  (24 word list) but with another procedure called the [aezeed cipher seed scheme](https://github.com/lightningnetwork/lnd/blob/5ba630f02b09c9b6b01273e8bd7e4d8f5343f38d/aezeed/README.md).

The one based on aez comes with some sweet features like wallet timestamp and versioning in the seed but I am not sure if those are needed in this case, they are a nice have, or they are not needed at all.

My idea is to discuss which scheme is better for this use case and implement it as an external plugin (I did not investigate yet how plugins are loaded in the program).

Not sure if someone else is already taking care of this, there is no issue for it. I will add @cdecker who seems to be the coordinator for this.

positiveblue:
@cdecker 

[...]

Issue #3687, opened 2020-05-01
hsmd: migrate to BIP32 derivation for all keys


darosior:
*In the middle of writing this I started really doubting about the feasibility/tradeoffs of it.*

*Changing the derivation would require users to close their channels and overall introduce a big mess (I think) between channel using old derivation and channels using the new one. I'm not sure the benefits outweigh the complications..*

*I'm leaving the below text for context, but do you think it's worth fixing this `FIXME` at all or is it too late ? Do you have a better automagic way of updating existing channels ?*

-----

I'd like to gather some feedback on how this can get accomplished before getting started on the implementation.

### Rationale

As noted in the following `FIXME`, LND has a [nicer BIP32-based and versioned derivation](https://github.com/lightningnetwork/lnd/blob/master/keychain/derivation.go#L48). Migrating to it would allow to share the same and compatible backup methods (for example, even if controversed, I personally find BIP39 mnemonics really good for user experience).
https://github.com/ElementsProject/lightning/blob/10f47b41fa3192638442ef04d816380950cc32c9/hsmd/hsmd.c#L375-L376

On the same subject, I think it would be a good opportunity to expand the `hsmtools` helper, for example with @jb55's [clightning-dumpkeys](https://github.com/jb55/clightning-dumpkeys) along with mnemonic-related stuff. It could also be [installed (and documented..)](https://github.com/ElementsProject/lightning/issues/3492#issuecomment-583931721).

Finally, I don't know what derivation method do ACINQ and R-L use, but it would be great for user experience to all use the same method (and LND's one seems a good candidate).

## Our current derivation

We currently use unhardened derivation only for onchain stuff.

We derive keys from `m/0/0` with the root seed being :
```hkdf_sha256(hsm_secret, salt=NULL, info="bip32 seed")```

We derive our node id with : 
```node_privkey = hkdf_sha256(hsm_secret, salt=0, info="nodeid")``` 

We derive our basepoints with:
```
channel_seed = hkdf_sha256(hkdf_sha256(hsm_secret, salt=NULL, info="peer seed"), salt=peer_id | dbid, info="per-peer seed")
[funding, revocation, payment, htlc, delayed, shaseed] = expanded_hkdf_sha256(channel_seed, salt=NULL, info="c-lightning")
```
With `expanded_hkdf_sha256` which outputs a `6 * 32` bytes-long key with each 32 bytes part used (in this order) for each basepoint.

### A versioned `hsm_secret`

[...]

PR #4065, opened 2020-09-18
`hsm_secret` generation from a seed-phrase

positiveblue:
Add `generatehsm` method to hsmtool to derivate BIP32 seeds from a
mnemonic using the BIP39 standard.

The new method uses libwally for the BIP39 to BIP32 derivation. It also
fails if an hsm_secret file already exists, so we do not overwrite
someone else's wallet without noticing.

It allows the use of passphrases, the ECHO mode in the terminal is
disable for higher security.

### DIFF
--- a/doc/lightning-hsmtool.8
+++ b/doc/lightning-hsmtool.8
@@ -51,6 +51,10 @@ and is usually no greater than the number of channels that the node has
 ever had\.
 Specify \fIpassword\fR if the \fBhsm_secret\fR is encrypted\.
 
+
+\fBgeneratehsm\fR \fIhsm_secret_path\fR
+Generates a new hsm_secret using BIP39\.
+
 .SH BUGS
 
 You should report bugs on our github issues page, and maybe submit a fix
@@ -76,4 +80,4 @@ Note: the modules in the ccan/ directory have their own licenses, but
 the rest of the code is covered by the BSD-style MIT license\.
 Main web site: \fIhttps://github.com/ElementsProject/lightning\fR
 
-\" SHA256STAMP:11b3e6c050eb31e7faac10e8b75c3f7b7d57269f7f8c47c67f6d115b7a479c09
+\" SHA256STAMP:918981692d3840344e15c539b007b473d5ea0ad481145eccff092bf61ec6ddb0
--- a/tools/hsmtool.c
+++ b/tools/hsmtool.c
@@ -1,4 +1,5 @@
 #include <bitcoin/privkey.h>
+#include <ccan/array_size/array_size.h>
 #include <ccan/crypto/hkdf_sha256/hkdf_sha256.h>
 #include <ccan/err/err.h>
 #include <ccan/noerr/noerr.h>
@@ -13,15 +14,19 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
+#include <limits.h>
 #include <sodium.h>
 #include <sys/stat.h>
+#include <termios.h>
 #include <unistd.h>
+#include <wally_bip39.h>
 
 #define ERROR_HSM_FILE errno
 #define ERROR_USAGE 2
 #define ERROR_LIBSODIUM 3
 #define ERROR_LIBWALLY 4
 #define ERROR_KEYDERIV 5
+#define ERROR_LANG_NOT_SUPPORTED 6
 
 static void show_usage(const char *progname)
 {
@@ -33,6 +38,7 @@ static void show_usage(const char *progname)
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
 	printf("	- guesstoremote <P2WPKH address> <node id> <tries> "
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
+	printf("	- generatehsm <path/to/new//hsm_secret>\n");
 	exit(0);
 }
 
@@ -368,6 +374,145 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
 	return 1;
 }
 
+static void get_words(struct words **words) {
+	struct wordlist_lang {
+		char *abbr;
+		char *name;
+	};
+
+	struct wordlist_lang languages[] = {
+		{"en", "English"},
+		{"es", "Spanish"},
+		{"fr", "French"},
+		{"it", "Italian"},
+		{"jp", "Japanese"},
+		{"zhs", "Chinese Simplified"},
+		{"zht", "Chinese Traditional"},
+	};
+
+	printf("Select your language:\n");
+	for (size_t i = 0; i < ARRAY_SIZE(languages); i++) {
+		printf("  %zu) %s (%s)\n", i, languages[i].name, languages[i].abbr);
+	}
+	printf("Select [0-%zu]: ", ARRAY_SIZE(languages));
+
+	char *selected = NULL;
+	size_t size = 0;
+	size_t characters = getline(&selected, &size, stdin);
+	if (characters < 0)
+		errx(ERROR_USAGE, "Could not read line from stdin.");
+
+	/* To distinguish success/failure after call */
+	errno = 0;
+	char *endptr;
+	long val = strtol(selected, &endptr, 10);
+	if (errno == ERANGE || (errno != 0 && val == 0) || endptr == selected || val < 0 || val >= ARRAY_SIZE(languages))
+        errx(ERROR_USAGE, "Invalid language selection, select one from the list [0-6].");
+
+	bip39_get_wordlist(languages[val].abbr, words);
+}
+
+static void get_mnemonic(char *mnemonic) {
+	char *line = NULL;
+	size_t line_size = 0;
+
+	printf("Introduce your BIP39 word list separated by space:\n");
+	size_t characters = getline(&line, &line_size, stdin);
+	if (characters < 0)
+		errx(ERROR_USAGE, "Could not read line from stdin.");
+	line[characters-1] = '\0';
+	strcpy(mnemonic, line);
+	free(line);
+}
+
+static void read_mnemonic(char *mnemonic) {
+	/* Get words for the mnemonic language */
+	struct words *words;
+	get_words(&words);
+
+	/* Get mnemonic */
+	get_mnemonic(mnemonic);
+
+	if (bip39_mnemonic_validate(words, mnemonic) != 0) {
+		errx(ERROR_USAGE, "Invalid mnemonic: \"%s\"", mnemonic);
+	}
+}
+
+static void read_passphrase(char **passphrase) {
+	struct termios current_term, temp_term;
+	printf("Warning: remember that different passphrases yield different "
+	       "bitcoin wallets.\n");
+	printf("If left empty, no password is used (echo is "
+	       "disabled now).\n");
+	printf("Enter your passphrase: \n");
+
+	/* Change terminal options so we do not echo the passphrase */
+	if (tcgetattr(fileno(stdin), &current_term) != 0)
+		errx(ERROR_USAGE, "Could not get current terminal options.");
+	temp_term = current_term;
+	temp_term.c_lflag &= ~ECHO;
+	if (tcsetattr(fileno(stdin), TCSAFLUSH, &temp_term) != 0)
+		errx(ERROR_USAGE, "Could not disable passphrase echoing.");
+	/* If we don't flush we might end up being buffered and we might seem
+	 * to hang while we wait for the password. */
+	fflush(stdout);
+
+	size_t passphrase_size = 0;
+	size_t characters = getline(passphrase, &passphrase_size, stdin);
+	if (characters < 0)
+		errx(ERROR_USAGE, "Could not read passphrase from stdin.");
+
+	/* Newline is not part of the valid passphrase */
+	if ( (*passphrase)[characters-1] == '\n' ) {
+		(*passphrase)[characters-1] = '\0';
+	}
+
+	/* If the user did not introduce any password, we want to set passphrase
+	 * to NULL not to '\0' for libwally */
+	if (strlen(*passphrase) == 0) {
+		free(*passphrase);
+		*passphrase = NULL;
+	}
+
+	if (tcsetattr(fileno(stdin), TCSAFLUSH, &current_term) != 0)
+		errx(ERROR_USAGE, "Could not restore terminal options.");
+}
+
+static int generate_hsm(const char *hsm_secret_path)
+{
+	char mnemonic[BIP39_WORDLIST_LEN];
+	read_mnemonic(mnemonic);
+
+	char *passphrase = NULL;
+	read_passphrase(&passphrase);
+
+	u8 bip32_seed[BIP39_SEED_LEN_512];
+	size_t bip32_seed_len;
+
+	if (bip39_mnemonic_to_seed(mnemonic, passphrase, bip32_seed, sizeof(bip32_seed), &bip32_seed_len) != WALLY_OK)
+		errx(ERROR_LIBWALLY, "Unable to derive BIP32 seed from BIP39 mnemonic");
+
+	int fd = open(hsm_secret_path, O_CREAT|O_EXCL|O_WRONLY, 0400);
+	if (fd < 0) {
+		errx(ERROR_USAGE, "Unable to create hsm_secret file");
+	}
+	if (!write_all(fd, bip32_seed, bip32_seed_len))
+		errx(ERROR_USAGE, "Error writing secret to hsm_secret file");
+
+	if (fsync(fd) != 0)
+		errx(ERROR_USAGE, "Error fsyncing hsm_secret file");
+
+	/* This should never fail if fsync succeeded. But paranoia is good, and bugs exist */
+	if (close(fd) != 0)
+		errx(ERROR_USAGE, "Error closing hsm_secret file");
+
+	printf("New hsm_secret file created at %s\n", hsm_secret_path);
+	printf("Use the `encrypt` command to encrypt the BIP32 seed if needed\n");
+
+	free(passphrase);
+	return 0;
+}
+
 int main(int argc, char *argv[])
 {
 	const char *method;
@@ -411,6 +556,21 @@ int main(int argc, char *argv[])
 			errx(ERROR_USAGE, "Bad node id");
 		return guess_to_remote(argv[2], &node_id, atol(argv[4]),
 		                       argv[5], argc >= 7 ? argv[6] : NULL);
+	}
+
+	if (streq(method, "generatehsm")) {
+		if (argc != 3)
+			show_usage(argv[0]);
+
+		char *hsm_secret_path = argv[2];
+
+		/* if hsm_secret already exists we abort the process
+		 * we do not want to lose someone else's funds */
+		struct stat st;
+		if (stat(hsm_secret_path, &st) == 0)
+			errx(ERROR_USAGE, "hsm_secret file at %s already exists", hsm_secret_path);
+
+		return generate_hsm(hsm_secret_path);
 	}
 
 	show_usage(argv[0]);
--- a/doc/lightning-hsmtool.8.md
+++ b/doc/lightning-hsmtool.8.md
@@ -48,6 +48,9 @@ and is usually no greater than the number of channels that the node has
 ever had.
 Specify *password* if the `hsm_secret` is encrypted.
 
+**generatehsm** *hsm\_secret\_path*
+Generates a new hsm_secret using BIP39.
+
 BUGS
 ----
 


positiveblue:
Related to the issue created during the Lightning HackSprint May 2020 #3717 

@ZmnSCPxj @darosior I guess that you are the right people for reviewing this

darosior (commented):
Nice, will review soon (:tm:) but strong concept ack :)

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	/* If we don't flush we might end up being buffered and we might seem
+	 * to hang while we wait for the password. */
+	fflush(stdout);
+
+	size_t passphrase_size;
+	if (getline(&passphrase, &passphrase_size, stdin) < 0)
+		errx(ERROR_USAGE, "Could not read passphrase from stdin.");
+	printf("passphrase size: %ld\n", passphrase_size);
+	/* Replace the last character which will always be '\n' */
+	passphrase[strlen(passphrase) - 1] = '\0';
```
darosior, line 417: Not always:

```C
#include <inttypes.h>
#include <stdio.h>
#include <string.h>

int main(int argc, char **argv) {
	char *test = NULL;
	size_t len;

	getline(&test, &len, stdin);
	printf("%s\n", test[strlen(test) - 1] == '\n' ? "true" : "false");
}
```

```
darosior@darosior:~$ gcc a.c
darosior@darosior:~$ ./a.out 
test_pass
true
darosior@darosior:~$ printf passphrase_from_elsewhere |./a.out 
false
```

tools/hsmtool.c
```diff
@@ -33,6 +35,7 @@ static void show_usage(const char *progname)
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
 	printf("	- guesstoremote <P2WPKH address> <node id> <tries> "
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
+	printf("	- generatehsm <path/to/hsm_secret>\n");
```
darosior, line 38:
```suggestion
	printf("	- generatehsm <path/to/new//hsm_secret>\n");
```
To imply it's not existing yet ?

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+		errx(ERROR_USAGE, "Error writing secret to hsm_secret file");
+
+	if (fsync(fd) != 0)
+		errx(ERROR_USAGE, "Error fsycning hsm_secret file");
+
+	/* This should never fail if fsync succeeded. But paranoia is good, and bugs exist */
+	if (close(fd) != 0) 
+		errx(ERROR_USAGE, "Error closing hsm_secret file");
+	
+	printf("New hsm_secret file created. Use hsmtool to encrypt the BIP32 seed if needed");
```
darosior, line 454:
```suggestion
	printf("New hsm_secret file created. Use the `encrypt` command to encrypt the BIP32 seed if needed");
```

tools/hsmtool.c
```diff
@@ -413,5 +501,20 @@ int main(int argc, char *argv[])
...
+		if ( argc != 3)
+			show_usage(argv[0]);
+
+		char *hsm_secret_path = argv[2];
+
+		/* if hsm_secred already exists we abort the process
+		 * we do not want to lose someone else's funds */
+		struct stat st;
+		if (stat(hsm_secret_path, &st) == 0)
+			errx(ERROR_USAGE, "hsm_secret file already exists");
```
darosior, line 514: Nice sanity check :)

darosior (commented):
Just a quick pass as Travis does not pass yet, but wanted to give it a look!

tools/hsmtool.c
positiveblue: I am not used to that nomenclature but is it common for new files to be described as that?

tools/hsmtool.c
positiveblue: This is a really good catch because I did not think about that... Thanks @darosior!

tools/hsmtool.c
darosior: Absolutely not, entirely baked by myself (though i'm almost sure to have read that somewhere as well) because it's clearer in my mind. :)
Was more an opinion than a request!

tools/hsmtool.c
positiveblue: Ok, after some time playing with this I think that we have a problem here... I do not see a way to know if the last character is noise or was introduced by the user. The reason is that `getline` returns the number of characters that where read, including the `\n` if present but not putting it in the read bytes. Then, if we know that we readed 2 chars, we do not know if `test[1]` is noise that was there because the user pushed enter `\n` or if the user executed `printf "xc" | ./myprogram` and then test[1] is the "c"

I think that the best is going to be to assume that the user is using this interactively and always replace the last character by `\0` as it was the enter and was not supposed to be part of the passphrase.

If not, we have that


```
#include <inttypes.h>
#include <stdio.h>
#include <string.h>

int main(int argc, char **argv) {
	char *test = NULL;
	size_t len;

	size_t characters = getline(&test, &len, stdin);
        printf("read chars: %ld\n", characters);
	printf("%s\n", test[strlen(test) - 1] == '\n' ? "true" : "false");
}
```

Then...

```
~/tmp ~$ ./a.out 
a
read chars: 2
true
~/tmp ~$ printf "ab" | ./a.out
read chars: 2
false
```

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+
+	if (tcsetattr(fileno(stdin), TCSAFLUSH, &current_term) != 0)
+		errx(ERROR_USAGE, "Could not restore terminal options.");
+}
+
+static int generate_hsm(const char *hsm_secret_path)
+{
+	char mnemonic[BIP39_WORDLIST_LEN];
+	read_mnemonic(mnemonic);
+	if (bip39_mnemonic_validate(NULL, mnemonic) != 0)
```
niftynei: does this correctly validate non-English word list sets? You might try the test cases from https://github.com/bip32JP/bip32JP.github.io/blob/master/test_JP_BIP39.json

tools/hsmtool.c
niftynei: Maybe also echo the path that the hsm_secret file was created at? 

New hsm_secret file created. -> New /path/to/hsm_secret created.

tools/hsmtool.c
```diff
@@ -413,5 +501,20 @@ int main(int argc, char *argv[])
 		                       argv[5], argc >= 7 ? argv[6] : NULL);
 	}
 
+	if (streq(method, "generatehsm")) {
+		if ( argc != 3)
```
niftynei, line 524:
```suggestion
		if (argc != 3)
```

tools/hsmtool.c
```diff
@@ -413,5 +501,20 @@ int main(int argc, char *argv[])
...
+			show_usage(argv[0]);
+
+		char *hsm_secret_path = argv[2];
+
+		/* if hsm_secred already exists we abort the process
```
niftynei, line 529:
```suggestion
		/* if hsm_secret already exists we abort the process
```

tools/hsmtool.c
niftynei: I'd echo the path that already exists here also e.g.
```suggestion
		/* if hsm_secred already exists we abort the process
		 * we do not want to lose someone else's funds */
		struct stat st;
		if (stat(hsm_secret_path, &st) == 0)
			errx(ERROR_USAGE, "/path/to/hsm_secret already exists. Aborting.");
```

doc/lightning-hsmtool.8
```diff
@@ -76,4 +80,4 @@ Note: the modules in the ccan/ directory have their own licenses, but
 the rest of the code is covered by the BSD-style MIT license\.
 Main web site: \fIhttps://github.com/ElementsProject/lightning\fR
```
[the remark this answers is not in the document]
positiveblue: yes, I did not squash it because I saw so many files changed (and not added in the gitignore that it was weird...) I will do it at the end

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
 	return 1;
 }
 
+static void read_mnemonic(char *mnemonic) {
+	printf("Introduce the BIP39 24 word list:\n");
+	char word[50];
+	uint n_words;
+	for (n_words = 0; n_words < 24; n_words++) {
```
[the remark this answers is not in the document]
positiveblue: Should we support all of them? How do you want the user to define how long their mnemonic is? another input asking for the number? two empty lines in a row?

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+		scanf("%s", word);
+		/* We need to concat the strings using ' ' so libwally can
+		 * validate it */
+		if (n_words == 0) {
+			strcpy(mnemonic, word);
+		} else {
+			strcat(mnemonic, " ");
```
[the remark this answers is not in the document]
positiveblue: I will update it to support all the languages

tools/hsmtool.c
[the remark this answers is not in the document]
niftynei: Well if you do 'readline" for an entire line which includes the complete passphrase, there's no problem with word counting? I'm not sure how robust libwally's validation method is; you might want to validate that that works as expected

tools/hsmtool.c
darosior: Why not doing it as we do it for `--encrypted-hsm` ?
https://github.com/ElementsProject/lightning/blob/cafaad741b39806a8bd9fff7fe71100488cf4bed/lightningd/options.c#L417-L418

positiveblue:
@niftynei I updated the PR. Now it reads the mnemonic all at once using `getline()` and use that directly.

If the list of words is not provided in the right (standard) format (`word1 word2 word3...`) it fails. Because we use libwally under the hood, we can validate and derive the seed without any problem, supporting out of the box all the mnemonic length 12, 18, 24...) and multiple languages.

I tried to compare the obtained results with the ones from [other bip39 implementations](https://iancoleman.io/bip39) and I got the same results (with and without passphrase) for english, spanish, italian, chinese... including non ASCII characters too.

However, there is one still failing. Japanese mnemonics do not pass the libwally validation. I wonder if it has something to do with the terminal byte encoding or something like that because with Chinese characters I did not have any problem... do you know if you have to do something special with JP characters? Even so, I can just show and error if Japanese is selected (or not show the option) and leave it as a TODO for the next PR.

tools/hsmtool.c
```diff
@@ -368,6 +371,145 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+			strcpy(lang, "jp");
+			break;
+		case 5:
+			strcpy(lang, "zhs");
+			break;
+		case 6:
+			strcpy(lang, "zht");
+			break;
+		default:
+			errx(ERROR_USAGE, "Invalid language selection, select one from the list [0-6].");
```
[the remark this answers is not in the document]
darosior: You could use `strtol` or similar

tools/hsmtool.c
```diff
@@ -368,6 +371,145 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+		errx(ERROR_USAGE, "Could not restore terminal options.");
+}
+
+static int generate_hsm(const char *hsm_secret_path)
+{
+	char mnemonic[BIP39_WORDLIST_LEN];
+	read_mnemonic(mnemonic);
+
+	char *passphrase = NULL;
+	read_passphrase(&passphrase);
```
darosior, line 485: If `read_passphrase` set `passphrase`, it's never freed

darosior (changes requested):
One main request on my side: we should support `passphrase == NULL` instead of `passphrase == '\0'`.

doc/lightning-hsmtool.8
```diff
@@ -51,6 +51,10 @@ and is usually no greater than the number of channels that the node has
 ever had\.
 Specify \fIpassword\fR if the \fBhsm_secret\fR is encrypted\.
 
+
+\fBgeneratehsm\fR \fIhsm_secret_path\fR
```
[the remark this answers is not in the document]
positiveblue: The documentation says that it generates the key following the BIP39 and also the text prompted to the user during the generation... if you prefer one of those names I am happy to change it but bip39 should be the standard by now (even when lnd nodes use aezeed). In the future, if we want to support multiple "standards", we can add a flag: `generatehsm --alg={aezeed, bip39,..}`

tools/hsmtool.c
```diff
@@ -368,6 +371,145 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	 * to hang while we wait for the password. */
+	fflush(stdout);
+
+	size_t passphrase_size = 0;
+	size_t characters = getline(passphrase, &passphrase_size, stdin);
+	if (characters < 0)
+		errx(ERROR_USAGE, "Could not read passphrase from stdin.");
+
+	/* Replace the last character because we assume it was the newline '\n' */
+	(*passphrase)[characters-1] = '\0';
```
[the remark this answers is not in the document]
positiveblue: > Why do we assume it was ? Couldn't we just check it is ?

well, the user needs to introduce a value an press enter (`getline()`). I can add a check and fail in case something weird is going on.

> Also we should support `passphrase == NULL`, for the very common "no passphrase" case !

The empty word and NULL have the [same effect](https://github.com/ElementsProject/libwally-core/blob/c9c45b68c5a18ca1b5bca41f3c3c8350a23b1bd8/src/bip39.c#L209) in libwally. I doubt that this never changes but I can substitute it by NULL when the user does not add any character to their passphrase.

 > Also, characters can in theory be 0

How do we get the line then? ctrl-d or something? In that case I think it's ok to just `segfault`

tools/hsmtool.c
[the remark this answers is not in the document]
positiveblue: >  This cannot be reached, `atoi` will give you `0` if throw it garbage

This is true, and I did not know it! Maybe the safest thing to do then is to take the first byte from the user's input and subtract '0' instead of doing any str to number conversion.

tools/hsmtool.c
positiveblue: Thanks for this,  is there any automatic tool that checks for this silly cases that I forgot? (so I can use it before future contributions and you do not have to lose your time on "simple" memory leaks)

darosior: Just went looking for missing `free`, but you can test with valgrind or clang with leak sanitizer enabled (`configure --enable-address-sanitizer`) to at least make sure the frequent paths don't leak.

tools/hsmtool.c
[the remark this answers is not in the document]
darosior: Hmm i'd say the safest thing here to go for `strtol`, which errors on --well-- error. From the manpage:
> The program shown below demonstrates the use of strtol().  The first command-line argument specifies a string from which strtol() should parse a number.  The second (optional) argument  specifies  the
       base to be used for the conversion.  (This argument is converted to numeric form using atoi(3), a function that performs no error checking and has a simpler interface than strtol().)  Some examples of
       the results produced by this program are the following:

           $ ./a.out 123
           strtol() returned 123
           $ ./a.out '    123'
           strtol() returned 123
           $ ./a.out 123abc
           strtol() returned 123
           Further characters after number: abc
           $ ./a.out 123abc 55
           strtol: Invalid argument
           $ ./a.out ''
           No digits were found
           $ ./a.out 4000000000
           strtol: Numerical result out of range

tools/hsmtool.c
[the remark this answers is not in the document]
darosior: > well, the user needs to introduce a value an press enter (getline()). I can add a check and fail in case something weird is going on.

https://github.com/ElementsProject/lightning/pull/4065#discussion_r493036721 ... I know it's harder to get there with the two consecutive `getline`, but that's a tool and it's going to be used in scripts and stuff.
Moreover, you are overriding the user passphrase here! This definitely needs a sanity check (maybe a belt and suspenders, but at least the belt will do for now!).
If you are really confident about this, make it an assert:
```C
assert((*passphrase)[characters-1] == '\n');
(*passphrase)[characters-1] = '\0';
```
But i think it's a bad idea and can be triggered. Don't know what's wrong with just doing:
```C
if ((*passphrase)[characters-1] == '\n')
    (*passphrase)[characters-1] = '\0';
```

> The empty word and NULL have the same effect in libwally. I doubt that this never changes but I can substitute it by NULL when the user does not add any character to their passphrase.

Oh, i was not aware of that :) I just check the doc comment stating "if no passphrase then set to NULL". So yeah better to rely on documented behaviour i guess!

tools/hsmtool.c
darosior: Even better: [write a test](https://github.com/ElementsProject/lightning/blob/d151d554b37e8cbf117d3a85f1503588830da8a8/tests/test_wallet.py#L988) to trigger those codepaths :)

tools/hsmtool.c
```diff
@@ -368,6 +372,148 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	printf("Select [0-6]: ");
+
+	char *selected = NULL;
+	size_t size = 0;
+	size_t characters = getline(&selected, &size, stdin);
+	if (characters < 0)
+		errx(ERROR_USAGE, "Could not read line from stdin.");
+
+	int selection = selected[0] - '0';
+	if (characters != 2 || selection < 0 || selection > 6)
```
cdecker, line 393: `scanf` would be more appropriate here, rather than checking the length of the string and doing the char-to-int conversion.

tools/hsmtool.c
```diff
@@ -368,6 +372,148 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+			bip39_get_wordlist("es", words);
+			break;
+		case 2:
+			bip39_get_wordlist("fr", words);
+			break;
+		case 3:
+			bip39_get_wordlist("it", words);
+			break;
+		case 4:
+			errx(ERROR_LANG_NOT_SUPPORTED, "Sorry, the language selected is not supported yet.");
```
cdecker, line 410: It's strange to be offering an option and then declining the selection.

tools/hsmtool.c
```diff
@@ -368,6 +372,148 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
 	return 1;
 }
 
+static void get_words(struct words **words) {
+	printf("Select your language:\n");
```
cdecker, line 376: There seem to be quite a number of repetitions in the code here, that might be better suited for a list of structs that contain the various pieces of information:

```c
struct wordlist_lang {
	char *abbr;
	char *name;
};

struct wordlist_lang languages[] = {
	{"en", "English"},
	{"es", "Spanish"},
	{"fr", "French"},
	{"it", "Italian"},
	{"jp", "Japanese"},
	{"zhs", "Chinese Simplified"},
	{"zht", "Chinese Traditional"},
};
```
Then you can use `ARRAY_SIZE` to determine the size of the array at compile time, iterate through them when printing the menu, and indexing with the result after selection to find the abbreviation to use when fetching the wordlist from `libwally` instead of having to update the `printf`, index checks and switch-statement below.

tools/hsmtool.c
```diff
@@ -368,6 +372,148 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+		default:
+			errx(ERROR_USAGE, "Invalid language selection, select one from the list [0-6].");
+	}
+}
+
+static void get_mnemonic(char *mnemonic) {
+	char *line = NULL;
+	size_t line_size = 0;
+
+	printf("Introduce your BIP39 word list:\n");
```
cdecker, line 428: I'd add `separated by space`, so people don't attempt to write one per line, which is what I'd have done :-)

cdecker:
Excellent addition @positiveblue, small nits aside, I think this is a great feature, and I'm positive we'll merge it before the next release :+1:

tools/hsmtool.c
positiveblue: I changed it to `getline()` because `scanf` left the `\n` char and sequential `getline()`failed. What is the best way to deal with that?

tools/hsmtool.c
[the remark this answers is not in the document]
positiveblue: Check added

tools/hsmtool.c
positiveblue: Updated

tools/hsmtool.c
```diff
@@ -368,6 +371,91 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	}
+	/* scanf() leaves the newline character in the input buffer */
+	getchar();
+}
+
+static void read_passphrase(char *passphrase) {
+	struct termios current_term, temp_term;
+	printf("Warning: remember that different passphrase yield different "
+	       "bitcoin wallets.\n");
+	printf("Leave it empty for using the value by default (echo is "
```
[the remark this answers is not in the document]
darosior: +1, probably better to say "if left empty, no password is used".

tools/hsmtool.c
```diff
@@ -368,6 +374,144 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	// newline is not part of the valid passphrase
+	if ( (*passphrase)[characters-1] == '\n' ) {
+		(*passphrase)[characters-1] = '\0';
+	}
+
+	// If the user did not introduce any password, we want to set passphrase
+	// to NULL not to '\0' for libwally
+	if (strlen(*passphrase) == 0) {
+		free(*passphrase);
+	}
```
darosior, line 474: You don't set it to NULL here
```c
free(*passphrase);
*passphrase = NULL;
```

darosior (commented):
Just two small things left, starting to look pretty good :-)

niftynei (commented):
Few more nits, otherwise looks good.


It occurred to me while looking into this that the "normal" BIP39 to BIP32 pathway isn't true for c-lightning. Typically you'd pass the seed generated by BIP39's derivation directly to `bip32_key_from_seed`. However, in hsmd.c, we add an intermediate derivation (using an hkdf), passing the BIP39 seed into this function to get *yet another* seed, which we then pass to `bip32_key_from_seed`.

This isn't a blocker for this patch. However, it will prevent "seed portability". You can't use a mnemonic here and anticipate being able to "seamlessly" import your lightning funds elsewhere using the same mnemonic.

If we want 'full compliance' with the BIP39 to BIP32 pathway, we'll need to update the seed derivation to remove this secondary step. This would be a good opportunity to more fully embrace BIP32 pathways for different key paths -- I hear LND has an (undocumented) pathway scheme for key generation and usage that might be worth porting over to c-lightning ...

darosior:
@positiveblue feel free to ping me as soon as you force push, would like to ACK this before we cut the release :-)

> This would be a good opportunity to more fully embrace BIP32 pathways for different key paths -- I hear LND has an (undocumented) pathway scheme for key generation and usage that might be worth porting over to c-lightning ...

Hehe .. https://github.com/ElementsProject/lightning/issues/3687 :)

positiveblue:
@darosior I see the problems that you mention in #3687 

Because many projects are already using c-lightning we can not simple ask them to close all the channels and open them again. If we really want the project to support BIP39, Azeed (lnd) and other derivations we should support the old way as a legacy feature and introduce the new `third-party compatible` derivations as default.

For Azeed is easy, but for BIP39 is hard taking into account that we can not use the current `populate_secretstuff` function form `hsmd.c`.

Right now people can only recover their wallet by copy the `hsm_secret` file. This PR could be helpful for people creating their `hsm_secret` in a reproducible way for the future. However, at some point the project should just use the default bip39 derivation path and add a flag for all wallets (which would use the current `hkdf_sha256` step)

What do you think about it?

niftynei:
ACK 7cc067479046ae82a2de53f0ddd34bc1dad2ce0c

darosior:
post-merge ACK 7cc0674 . Thanks for having worked on this @positiveblue !

> Right now people can only recover their wallet by copy the hsm_secret file.

This a not a good way to backup and/or recover the wallet, without the db you may (are going to?) be missing utxos. See https://github.com/ElementsProject/lightning/pull/4171 for more discussions about that

darosior:
Hmm so i'm pretty sure to have tested this with `encrypt` and `decrypt`, and even just that it yields an actual `hsm_secret`. But it so happens that it creates a 512-bytes `hsm_secret` !
I tested both master and this branch, and neither does it not seem to be libwally as it didn't change much: https://github.com/ElementsProject/libwally-core/blob/master/src/bip39.c ...

Issue #4029, opened 2020-09-09
investigate: channel state changes two times to AWAITING_UNERLATERAL on force close


m-schmoock:
### Issue and Steps to Reproduce
When a channel is closed unilaterally the channel changes the state to AWAITING_UNERLATERAL two times, which means
the second time the channel is already in that state.

This was discovered when implementing tests for https://github.com/ElementsProject/lightning/pull/4020

See https://github.com/ElementsProject/lightning/commit/160c564e5f5815d7eea2ac68b160ba5240043f62 for the 'fix' to suppress the duplicated event notification.

m-schmoock:
Since we currently save the channel state to the database on each state change (for good reason), we can't skip duplicated updates without selectively checking the database state. The duplicated state changes came from the onchain watchers in `onchain_control.c` and the`channel_fail_permanent` that is used throughout the code.

Addressing this seems over-engineering to me, so we are better of by simply suppressing reaching out duplicated events to the API.

PR #4069, opened 2020-09-19
fix: duplicated channel_state_changed events

m-schmoock:
This fixes occurrences where current code paths were setting the same channel state twice.
This happened on:
 - `tests/test_plugin.py::test_channel_state_changed_unilateral`  => `AWAITING_UNILATERAL`
 - `tests/test_connection.py::test_fail_unconfirmed`  => `FUNDING_SPEND_SEEN`

It also adds a \*\*BROKEN\*\* log if this happens for any other reason in the future.
I also checked that the (now skipped) `wallet_channel_save` did produce exactly the same statements twice for these two occurrences. Maybe we should implement a solution to `/* TODO(cdecker) Selectively save updated fields to DB */` (channel.c line 402) or something like a `channel_is_dirty` function.

**Update**: 
While I was thinking about this I'm now a bit unsure if this is the right approach, specifically since `channel_set_state` also saves channel and config to DB. Suppressing duplicated state updates (and thus the DB update) may be problematic on regorgs while the node was offline and i.e. spending or closing TX has changed blockheight while `lightningd` was waiting for enough confirmations...

I think a better way would be if we detect and allow (and not log BROKEN) duplicated updates if DB state differed in any way.
What do you think?

Addresses #4029

### DIFF
--- a/lightningd/channel.c
+++ b/lightningd/channel.c
@@ -387,11 +387,15 @@ void channel_set_state(struct channel *channel,
 {
 	struct channel_id cid;
 
+	/* log and sanity checks */
 	log_info(channel->log, "State changed from %s to %s",
 		 channel_state_name(channel), channel_state_str(state));
 	if (channel->state != old_state)
 		fatal("channel state %s should be %s",
 		      channel_state_name(channel), channel_state_str(old_state));
+	if (state == old_state)
+		log_broken(channel->log, "Set same state multiple times: %s",
+					 channel_state_str(state));
 
 	channel->state = state;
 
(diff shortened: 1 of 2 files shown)


m-schmoock:
Obviously, because I **BROKEN**ed the event, now the testcases find other locations now where this occurs... will have a look at `test_fail_unconfirmed`

m-schmoock:
Note to myself: check if given state is recent

m-schmoock:
Im a bit unsure if this is the right approach, specifically since `channel_set_state` also saves channel and confg to DB. Suppressing duplicated state updates (and DB save) may be problematic on regorgs while the node was offline and i.e. spending TX has changed blockheight while clnd was waiting for enought confirmations...

I think I might add a offline/reorg tetcase for both occurences to be sure this is save, alternatively wie could allow (not BROKEN) duplicated updates if DB state differed.

What do you think?

PR #4071, opened 2020-09-21
Close PSBT leaks found by Christian, fixes to detect them in future

rustyrussell:
Interfacing with libwally without wrapping everything in our own struct is a pain, but this now manages it.

As an aside, there are cleanups and clarifications to the memleak interface.

### DIFF
--- a/bitcoin/psbt.c
+++ b/bitcoin/psbt.c
@@ -359,8 +359,8 @@ void psbt_elements_normalize_fees(struct wally_psbt *psbt)
 	}
 }
 
-bool psbt_has_input(struct wally_psbt *psbt,
-		    struct bitcoin_txid *txid,
+bool psbt_has_input(const struct wally_psbt *psbt,
+		    const struct bitcoin_txid *txid,
 		    u32 outnum)
 {
 	for (size_t i = 0; i < psbt->num_inputs; i++) {
@@ -388,7 +388,7 @@ bool psbt_input_set_redeemscript(struct wally_psbt *psbt, size_t in,
 	return wally_err == WALLY_OK;
 }
 
-struct amount_sat psbt_input_get_amount(struct wally_psbt *psbt,
+struct amount_sat psbt_input_get_amount(const struct wally_psbt *psbt,
 					size_t in)
 {
 	struct amount_sat val;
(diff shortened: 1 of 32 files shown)


bitcoin/psbt.c
```diff
@@ -32,8 +32,12 @@ static struct wally_psbt *init_psbt(const tal_t *ctx, size_t num_inputs, size_t
 	else
 		wally_err = wally_psbt_init_alloc(0, num_inputs, num_outputs, 0, &psbt);
 	assert(wally_err == WALLY_OK);
+	tal_steal(ctx, psbt);
+	/* If both of these are zero, no sub-allocations were done */
+	if (num_inputs || num_outputs)
```
cdecker, line 37: Is there any harm in calling it unconditionally?

common/daemon.c
```diff
@@ -81,11 +81,20 @@ void send_backtrace(const char *why)
...
 	if (t)
 		errx(1, "Outstanding taken pointers: %s", t);
 
+	wally_leak = tal_first(wally_tal_ctx);
+	if (wally_leak) {
+		/* Trigger valgrind to tell us about this! */
+		tal_free(wally_leak);
+		*wally_leak = 0;
+		errx(1, "Outstanding wally allocations");
+	}
```
cdecker, line 96: Very nice, this will report the location the non-collected child was allocated at in `valgrind`, correct?

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/4071/commits/5677d0af4b15134cf8490595e6e64b4a8785bf62

cdecker:
After checking I can confirm that the memleaks are indeed gone:

memdump-after

However it seems like I produced a (slow) memory leak myself when precomputing the txids in the speedup PR, I'll need to look into dropping the `b->txids` at an opportune time. Afterwards we'll have a 115MB memory footprint, 104 of which is logs :-)

common/daemon.c
rustyrussell: Yes, that's how I found them!

bitcoin/psbt.c
rustyrussell: No *harm*, but we have a debug check for the moment that we don't call tal_gather_wally unless there's something to gather.  Found some places where I was being overzealous.

PR #4072, opened 2020-09-21
db: Add support for key-value pair DSNs in postgresql

cdecker:
It turns out that it isn't always possible to express the connection
parameters in a URI-style DSN. The postgresql primitives provide an
alternative in the form of key-value DSNs [1], which allow for a far wider
range of possible configuration parameters. Specifically in my case I wanted
to have mutual authentication using TLS/SSL server and client certificates, on
top of password authentication.

With this change we still use the `scheme://` portion of the DSN to select
which backend to use, while we attempt to parse the connection string without
it first (which is the key-value option) and if that fails we fall back to the
full URI connection. Since we only attempt to parse, and don't directly
connect we don't actually connect twice.

So this is now a possibility:

``` text
postgres://sslmode=verify-ca \
	sslrootcert=server-ca.pem \
	sslcert=client-cert.pem \
	sslkey=client-key.pem \
	host=12.345.678.9 \
	port=5432 \
	user=postgres \
	dbname=cln \
	password=helloworld"
```

[1]: https://www.postgresql.org/docs/9.1/libpq-connect.html

### DIFF
--- a/tests/test_db.py
+++ b/tests/test_db.py
@@ -281,3 +281,20 @@ def test_optimistic_locking(node_factory, bitcoind):
         l1.rpc.newaddr()
 
     assert(l1.daemon.is_in_log(r'Optimistic lock on the database failed'))
+
+
+@unittest.skipIf(os.environ.get('TEST_DB_PROVIDER', None) != 'postgres', "Only applicable to postgres")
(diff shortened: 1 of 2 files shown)


rustyrussell (approved):
Ack e234bd924098c95a5255d8127afc3e8ec284d1ba

PR #4074, opened 2020-09-22
fix: openchannel_hook log BROKEN on duplicated close_to

m-schmoock:
... As discussed in yesterdays meeting, log BROKEN here is better, as a reliant plugin may behave incorrectly if that happens.


jsarenik:
utACK 12609f6

rustyrussell (approved):
Ack https://github.com/ElementsProject/lightning/pull/4074/commits/12609f6026e56f32e7c8c4866fcd232a43c6edd9

Issue #4075, opened 2020-09-22
Makefile hardcodes sha256sum, not available on FreeBSD


laanwj:
The following error appears on FreeBSD during build: `/bin/sh: sha256sum: not found`.

I think it comes from:
```
Makefile:SHA256STAMP_CHANGED = [ x"`sed -n 's/.*SHA256STAMP://p' $@ 2>/dev/null`" != x"`cat $(sort $(filter-out FORCE,$^)) | sha256sum | cut -c1-64`" ]
Makefile:SHA256STAMP = echo '$(1) SHA256STAMP:'`cat $(sort $(filter-out FORCE,$^)) | sha256sum | cut -c1-64` >> $@
```
Unfortunately, FreeBSD  doesn't have `sha256sum` but a tool `sha256` which works slightly differently. There's no real standard for this. See e.g. the [install_db4.sh](https://github.com/bitcoin/bitcoin/blob/master/contrib/install_db4.sh#L32) script in bitcoin which attempts to detect this.

It doesn't actually prevent anything from being built so I think it's a low-priority problem.

rustyrussell:
Yes, it only stops development files from being regenerated.  Still, worth fixing, thanks!

Fonta1n3:
I noticed the same thing on macOS recently fwiw. 0.9.1

PR #4081, opened 2020-09-23
Makefile: work with shasum -a as well as sha256sum.

rustyrussell:
At least on my Ubuntu box, they're compatible. If they're not, we need
to disable regeneration altogether.

Fixes: #4075
Changelog-None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -264,9 +264,9 @@ endif
 # Git doesn't maintain timestamps, so we only regen if sources actually changed:
 # We place the SHA inside some generated files so we can tell if they need updating.
 # Usage: $(call SHA256STAMP_CHANGED)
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4081/commits/3ffe4d04784e1c82d4be0dad5108c1318d34de78

laanwj:
Tested ACK, thanks.

kristapsk:
`sha256` could be added too, for FreeBSD. That's how we handle it in JoinMarket - https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/77422231207c7c3e984a88e944b6c715cece46b5/install.sh#L14.

PR #4082, opened 2020-09-23
Libwally: neaten tal capture.

rustyrussell:
This captures *all* libwally allocations, and also maintains logical structure around psbt and tx allocations (i.e. tx owns it parts, psbt owns its parts).

Changelog-None

### DIFF
--- a/bitcoin/tx.c
+++ b/bitcoin/tx.c
@@ -34,6 +34,7 @@ struct wally_tx_output *wally_tx_output(const tal_t *ctx,
 	struct wally_tx_output *output;
 	int ret;
 
+	tal_wally_start();
 	if (chainparams->is_elements) {
 		u8 value[9];
 		ret = wally_tx_confidential_value_from_satoshi(satoshis, value,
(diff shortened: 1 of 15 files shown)


cdecker:
ACK 4a1b08caf961f9252d6f77a299f0c406c4c4bcbd

bitcoin/tx.c
```diff
@@ -514,11 +529,16 @@ struct bitcoin_tx *bitcoin_tx_with_psbt(const tal_t *ctx, struct wally_psbt *psb
 					   psbt->tx->num_outputs,
 					   psbt->tx->locktime);
 	wally_tx_free(tx->wtx);
-	tx->wtx = psbt_finalize(tx, psbt, false);
+
+	psbt_finalize(psbt);
```
niftynei, line 533: iirc we don't want to actually finalize the `psbt` here as that'll update the passed in psbt. we want to extract the final_tx, which requires finalization but ideally on a different structure than the passed in one.

PR #4087, opened 2020-09-24
chore: reduce log buffer from 100MB to 10MB

m-schmoock:
The 100MB log buffer has been the biggest memory footprint for the daemon.
Keeping 10MB for emergency log dumps seems sufficient.
This has been mentioned in the last developer meeting.

**Update:** As @rustyrussell points out the log buffer always contains everything regardless of the config (io, esp and debug, ...). An average log entry has about 170Bytes, lets say 200Bytes with certain overhead. A well connected node may produce something like up to ~2 events per second depending on usage. Thus a 10MB buffer will last for ~8Hrs.

Changelog-Changed: In-memory log buffer reduced from 100MB to 10MB

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -180,7 +180,7 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
 	 * book to hold all the entries (and trims as necessary), and multiple
 	 * log objects which each can write into it, each with a unique
 	 * prefix. */
-	ld->log_book = new_log_book(ld, 100*1024*1024);
+	ld->log_book = new_log_book(ld, 10*1024*1024);
 	/*~ Note the tal context arg (by convention, the first argument to any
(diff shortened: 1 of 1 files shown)


rustyrussell:
Except the log holds *everything*, esp debug logs and io logs (to subdaemons and all JSON RPC).  

My log is 970266 seconds old, and it has 28886 entries (ignoring the "skipped" entries), and has eliminated 1150070. This equates to 1.2 log entries per second.

So I agree we should reduce this by 10, or even 100.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4087/commits/b8cff9db428b7bfb663d1b4b7b25827d2e8bb6ab

m-schmoock:
Right, I just checked my average log entry as ~170 chars (this was w/o the debug and io). So for 1.2 logs per second a 10MB buffer still covers about 14 hours.

PR #4090, opened 2020-09-24
fix: SQL statement expansion for plugin `db_write` hook

m-schmoock:
I discovered this accidentally when using the `tests/plugins/dblog.py`
plugin on another testcase: `tests/test_connection.py::test_fail_unconfirmed`
There the plugin/hook crashes because it can't execute the statement:
```json
{
  "jsonrpc": "2.0",
  "id": 34,
  "error": {
    "code": -32600,
    "message": "Error while processing db_write: unrecognized token: \"174WHERE\"",
    "traceback": "Traceback (most recent call last):\n  File \"/home/will/projects/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 535, in _dispatch_request\n    result = self._exec_func(method.func, request)\n  File \"/home/will/projects/lightning.git/contrib/pyln-client/pyln/client/plugin.py\", line 520, in _exec_func\n    return func(*ba.args, **ba.kwargs)\n  File \"/home/will/projects/lightning.git/tests/plugins/dblog.py\", line 45, in db_write\n    plugin.conn.execute(c)\nsqlite3.OperationalError: unrecognized token: \"174WHERE\"\n"
  }
}
```

I used the following change to have the testcase execute the plugin:
```
diff --cc tests/test_connection.py
index 02780a233,02780a233..76789afa4
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@@ -2351,8 -2351,8 +2351,11 @@@ def test_restart_many_payments(node_fac
  def test_fail_unconfirmed(node_factory, bitcoind, executor):
      """Test that if we crash with an unconfirmed connection to a known
      peer, we don't have a dangling peer in db"""
++    dbfile = os.path.join(node_factory.directory, "dblog.sqlite3")
++    opts={'plugin': os.path.join(os.getcwd(), 'tests/plugins/dblog.py'), 'dblog-file': dbfile}
++
      # = is a NOOP disconnect, but sets up file.
--    l1 = node_factory.get_node(disconnect=['=WIRE_OPEN_CHANNEL'])
++    l1 = node_factory.get_node(options = opts, disconnect=['=WIRE_OPEN_CHANNEL'])
      l2 = node_factory.get_node()
 
      # First one, we close by mutual agreement.
``` 

**Update:** I pulled in @cdecker changes from #4092 that fixes this for another location and adds a TEST_CHECK_DBSTMTS=1 option and a Travis Job so this does not happen again.


cdecker:
This has been in the codebase for quite some time, since https://github.com/ElementsProject/lightning/commit/50ff0b26fd48e0b495990ac55a2b72d1735873d9, but is only triggered when we run a plugin that uses the `db_write` hook. I think it must be that the query gets parsed and executed correctly as long as we use prepared statements where the parser does not combine the `?` placeholder with the `WHERE` that immediately follows it. During the expansion for the plugin we then perform a string replacement where the timestamp gets juxtaposed to the `WHERE` which is no longer syntactically valid, which then causes any attempt to execute the expanded statement to fail.

Impacted RPC calls (including any plugins that make use of these calls):

-   `reserveinputs`
-   `fundpsbt` if a reservation was specified which is true
-   `utxopsbt` same as `fundpsbt`
-   `sendpsbt`

So the crash is triggered only if a) we have a `db_write` hook plugin, b) the plugin crashes while trying to parse/execute the statement and c) call any of the above RPC calls. This is pretty much only the `dblog.py` plugin.

m-schmoock:
@cdecker I pulled in your commits from #4092 and fixed the missing generated SQL files for NO_PYTHON.

wallet/wallet.c
```diff
@@ -414,7 +414,7 @@ static void db_set_utxo(struct db *db, const struct utxo *utxo)
 		assert(!utxo->reserved_til);
 
 	stmt = db_prepare_v2(
-		db, SQL("UPDATE outputs SET status=?, reserved_til=?"
+		db, SQL("UPDATE outputs SET status=?, reserved_til=? "
 			"WHERE prev_out_tx=? AND prev_out_index=?"));
```
jsarenik, line 418: What about `" WHERE …"` instead of adding the space to end of previous line?

See lines [304](https://github.com/ElementsProject/lightning/blob/9b02c7827c8a4e864b8df1fa24f27325f18cf6ba/wallet/wallet.c#L304), 341, 518, 1219, etc.

m-schmoock: Both styles are used in this file. See lines: 75, 206, 214, 304, 553, 722, 734, 771, 1360, 1519, 1529, ... 
I can't decide whats better, since I don't see a difference. In general I would prefer to have the same style.
It would require a general decision and cleanup, which is not intended in this PR.

cdecker: It's pretty much personal preference, and we don't have a guideline. FWIW the same query would look like this if I were writing this in python:

```sql
UPDATE
  outputs
SET
  status=?,
  reserved_til=?
WHERE
  prev_out_tx=? AND
  prev_out_index=?
```

Which I find very readable and allows you to jump to any part of the query/statement you want. But multiline strings in C don't work.

jsarenik:
utACK 9b02c7827c8a4e864b8df1fa24f27325f18cf6ba

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4090/commits/9b02c7827c8a4e864b8df1fa24f27325f18cf6ba

PR #4093, opened 2020-09-25
Pay: use gossmap

rustyrussell:
This is a simple translation, but it involves a routing fix and some enhancements.

Building on this to get more route information is a job for @cdecker.

(Moving getroute et-al out to a plugin is still in my todo list, as is random routing!)

### DIFF
--- a/common/dijkstra.c
+++ b/common/dijkstra.c
@@ -22,6 +22,9 @@ struct dijkstra {
 	/* NULL means it's been visited already. */
 	const struct gossmap_node **heapptr;
 
+	/* How we decide "best" */
+	u64 score;
+
 	/* We could re-evaluate to determine this, but keeps it simple */
 	struct gossmap_chan *best_chan;
 };
@@ -67,9 +70,9 @@ static int less_comparer(const void *const ctx,
 			 const void *const b)
 {
 	return get_dijkstra(global_dijkstra, global_map,
-			    *(struct gossmap_node **)a)->distance
+			    *(struct gossmap_node **)a)->score
 		> get_dijkstra(global_dijkstra, global_map,
-			       *(struct gossmap_node **)b)->distance;
+			       *(struct gossmap_node **)b)->score;
 }
 
 static void item_mover(void *const dst, const void *const src)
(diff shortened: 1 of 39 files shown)


rustyrussell:
Trivial rebase...

rustyrussell:
Ya trivial rebase.

common/gossmap.c
```diff
@@ -835,3 +848,70 @@ u8 *gossmap_node_get_announce(const tal_t *ctx,
...
+	/* Note that first two bytes are message type */
+	const size_t feature_len_off = 2 + (64 + 64 + 64 + 64);
+	size_t feature_len;
+
+	feature_len = map_be16(map, c->cann_off + feature_len_off);
+
+	if (map_feature_set(map, OPTIONAL_FEATURE(fbit),
+			    c->cann_off + feature_len_off + 2, feature_len))
+		return OPTIONAL_FEATURE(fbit);
+	if (map_feature_set(map, COMPULSORY_FEATURE(fbit),
```
cdecker, line 883: We could let `map_feature_set` return both optional and compulsory flags, in order to avoid calling it twice.

cdecker (approved):
Excellent change, I particularly like that since it is now synchronous we no longer end up trampling the `channel_hints`.

ACK 0bb0e54

Still seems a bit suboptimal to compute a route on a static map, and only afterwards considering things like route length, CLTV and fee constraints. We could add a view-layer on top of the gossmap, created for each `dijkstra`-call that keeps track of these parameters, and sets the distance to unreachable if any of these constraints is violated. If we add the `struct payment` to the scoring function we could even do this in the scoring function itself, not requiring us to change `gossmap` at all. But all that's for another time :-)

common/dijkstra.c
```diff
@@ -22,6 +22,9 @@ struct dijkstra {
 	/* NULL means it's been visited already. */
 	const struct gossmap_node **heapptr;
 
+	/* How we decide "best" */
```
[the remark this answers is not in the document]
rustyrussell: Good idea, I'll do a followon cleanup commit.

common/gossmap.c
rustyrussell: Yes, it's fairly cheap, but since this is the only way we call it...

cdecker:
Tests using the `gossip_store` appear to be failing on Travis:

 - `test_gossip_store_load`
 - `test_gossip_store_load_announce_before_update`

Otherwise looking good :+1:

rustyrussell:
Ah, tests without COMPAT enabled.  OK, fixed.

PR #4094, opened 2020-09-25
pyln: Plugin methods and hooks refuse to set results twice

cdecker:
We had a couple of instances where a plugin would be killed by `lightningd`
because we were returning a result of an exception twice, and it was hard to
trace down the logic error in the user plugin that caused that. This patch
adds a traceback the first time we return a result/exception, and raise an
exception with a stacktrace of the first termination when a second one comes
in.

This can still terminate the plugin, but the programmer gets a clear
indication where the result was set, and can potentially even recover from it.

Changelog-Added: pyln: Plugin method and hook requests prevent the plugin developer from accidentally setting the result multiple times, and will raise an exception detailing where the result was first set.

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -69,6 +69,7 @@ class Request(dict):
         self.plugin = plugin
         self.state = RequestState.PENDING
         self.id = req_id
+        self.termination_tb: Optional[str] = None
 
     def getattr(self, key: str) -> Union[Method, Any, int]:
         if key == "params":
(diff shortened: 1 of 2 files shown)


contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -84,21 +85,37 @@ def getattr(self, key: str) -> Union[Method, Any, int]:
...
+                    tb=tb
+                ))
         self.result = result
         self._write_result({
             'jsonrpc': '2.0',
             'id': self.id,
             'result': self.result
         })
+        self.state = RequestState.FINISHED
+        self.termination_tb = traceback.extract_stack()
```
m-schmoock, line 105: See a2a64a362 as a suggestion for compacter code and more readability

contrib/pyln-client/tests/test_plugin.py
```diff
@@ -362,3 +362,25 @@ def test1(msat: Millisatoshi):
...
+        test1(*ba.args)
+
+    def test2(request):
+        request.set_exception(1)
+        request.set_exception(1)
+
+    req = Request(p, req_id=1, method="test2", params=[])
+    ba = p._bind_kwargs(test2, {}, req)
+    with pytest.raises(ValueError, match=r'current state is RequestState.FAILED'):
+        test2(*ba.args)
```
m-schmoock, line 386: I would improve on the testcase like in db7f6cdf9  
 - more cases: set_exception before set_result and other way around
 - checks for the occurrence of the stack with a MARKER comment

m-schmoock (commented):
lgtm
I added some improvement proposals... feel free to take or merge.

m-schmoock:
Ack 47cfb59

cdecker:
Folded in @m-schmoock's excellent fixups :+1:

m-schmoock:
Ack f042f27

(still)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4094/commits/8dee4eb108d7f623e8104d2b5ca95b269fe37b41

PR #4095, opened 2020-09-25
Add mypy typing verification to `make check-python`

### LINKED ISSUE #4085: mypy should be run by check-python
Otherwise we'll keep breaking all those type annotations!

cdecker:
This adds `mypy` verification to `pyln-client` and `pyln-testing`, while also restructuring the Makefiles a bit.

I added the verification to the pyln module Makefiles and call it from the main Makefile in order to maintain logical separation between the main codebase and the python modules, which we might end up splitting out into a separate repository eventually.

I started annotating `pyln-proto` but the `messages` and `invoice` module turned out to be a rather deep rabbit-hole that need more refactoring. It is integrated into the main Makefile and progress can be checked with `make check-pyln-proto`, so I'll spin this out into a separate issue.

I chose to ignore any external dependency that doesn't have annotation stubs in the [typeshed](https://github.com/python/typeshed) repository since creating those stubs would be another major time drain.

`pyln-spec` doesn't have any type-annotations yet, so we don't run them either :wink:

Fixes #4085
Changelog-None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -386,14 +386,20 @@ check-spelling:
 
 PYSRC=$(shell git ls-files "*.py" | grep -v /text.py) contrib/pylightning/lightning-pay
 
-check-python:
+# Some tests in pyln will need to find lightningd to run, so have a PATH that
+# allows it to find that
+PYLN_PATH=$(shell pwd)/lightningd:$(PATH)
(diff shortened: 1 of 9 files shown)


rustyrussell (approved):
Ack e373817eeb9c0573118202a54e9269c2c69a784a

PR #4097, opened 2020-09-25
Re-enable postgres test configuration after addressing issue 4080

### LINKED ISSUE #4080: Postgres CI fails
See #4079 

Forcing rerunning enough sometimes works, but we never get any output since it just RERUNs everything until it times out.

So if there are issues, they won't be fixed anyway.

cdecker:
This should help locate and fix any issue that was casing the `TEST_DB_PROVIDER=postgres` test configuration on Travis CI to fail. In order to do this we reduce `--reruns` from 5 to 2, potentially showing more flaky tests, but we should address them anyway, and we fail on the first error we get, since the rest of the test run is doomed anyway, and the extra information about which other tests would also fail is not worth the extra wait time.

The `db_provider` fixture goes from being session-scoped to being test-scoped, so we start and stop a separate instance of postgresql on each test, but measuring the performance impact showed a mere slowdown of 20 seconds on a 1250 second overall test-time, so negligible. On the upside it massively improves test isolation, and we no longer have to worry about scheduling two tests with huge numbers of nodes concurrently exhausting the connection backlog of the shared postgresql instance. This change has absolutely no impact on `TEST_DB_PROVIDER=sqlite3` configurations.

It also turns out that we weren't cleaning the temporary test directory in postgres configurations because the postgres server was leaving its data directory in there, blocking the cleanup. This is now fixed.

Fixes #4080
Changelog-None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,8 +36,7 @@ env:
 
   # Various other configurations
   - VALGRIND=0 ARCH=64 DEVELOPER=1 NETWORK=liquid-regtest
-  # FIXME: this fails almost all the time!
-#  - VALGRIND=0 ARCH=64 DEVELOPER=1 DB=postgres
+  - VALGRIND=0 ARCH=64 DEVELOPER=1 DB=postgres
   - VALGRIND=0 ARCH=arm32v7 DEVELOPER=1 TARGET_HOST=arm-linux-gnueabihf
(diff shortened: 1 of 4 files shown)


rustyrussell (approved):
Ack 66ee483c943c8deebb786ce87f717069d92abd90

PR #4113, opened 2020-10-07
Change mutual close feerate to be much less aggressive.

rustyrussell:
Using bitcoind's `estimatesmartfee 4 ECONOMICAL` is extremely aggressive for mutual close; particularly since you can use CPFP if you really want to speed it along.  Drop to `100 ECONOMICAL` (i.e. "slow") by default.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4113/commits/3042cdeb0f086447975d82bbd46540f87400688e

PR #4115, opened 2020-10-08
pytest: fix flaky test.

rustyrussell:
l2 only gives up 100 blocks after it ignores the HTLC, so sometimes
this test fails.


niftynei (approved):
ACK 854df62bfb589f46308dc9a8de3bd20a3a4bedc2

Issue #4108, opened 2020-10-04
`fundchannel_start` when we are after `WIRE_ACCEPT_CHANNEL` but before `WIRE_FUNDING_SIGNED` hangs and will mess up our knowledge of who the funder is


whitslack:
### Issue and Steps to Reproduce
I had someone open a channel to my node using a multi-open transaction `1c65afac07fdb769086f2020c4884e99bad75c69e0fce3728a6ca61d66ce28ce`. This is the first time I've seen a `short_channel_id` with a last component that wasn't 0 or 1, since the channel was funded by an output that wasn't the first or second output of the funding transaction.

Oddly, `listpeers` shows that *my node* was the one that funded the channel. That is absolutely incorrect. But, given that it thought so, it may not be surprising that it rejected the real channel opener's `update_fee` message and unilaterally closed the channel.

Is it possible that there is some logic somewhere for determining who opened the channel that is incorrect when the channel was opened as part of a multi-open?

```
{
   "peers": [
      {
         "id": "02…0c",
         "connected": true,
         "netaddr": [
            "[::…]:25694"
         ],
         "features": "0aa2a1",
         "channels": [
            {
               "state": "ONCHAIN",
               "scratch_txid": "…",
               "owner": "onchaind",
               "short_channel_id": "651234x1949x…",
               "direction": 1,
               "channel_id": "…",
               "funding_txid": "1c65afac07fdb769086f2020c4884e99bad75c69e0fce3728a6ca61d66ce28ce",
               "close_to_addr": "bc1…",
               "close_to": "…",
               "private": false,
               "features": [
                  "option_static_remotekey"
               ],
               "funding_allocation_msat": {
                  "02…0c": 0,
                  "02…2f": 5000000000
               },
               "funding_msat": {
                  "02…0c": "0msat",
                  "02…2f": "5000000000msat"
               },
               "msatoshi_to_us": 5000000000,
               "to_us_msat": "5000000000msat",
               "msatoshi_to_us_min": 5000000000,
               "min_to_us_msat": "5000000000msat",
               "msatoshi_to_us_max": 5000000000,
               "max_to_us_msat": "5000000000msat",
               "msatoshi_total": 5000000000,
               "total_msat": "5000000000msat",
               "dust_limit_satoshis": 546,
               "dust_limit_msat": "546000msat",
               "max_htlc_value_in_flight_msat": 18446744073709551615,
               "max_total_htlc_in_msat": "18446744073709551615msat",
[...]

PR #4116, opened 2020-10-08
openingd/: Fail `fundchannel_start` if we already are, or will become, the fundee.

ZmnSCPxj:
Fixes: #4108

Requesting review from nifty and rusty, since my grasp of the openingd is not good.

As well, there may be a similar issue in dual-funding, so please consider as well for that point-of-view, I am much more ignorant about the dual-funding flow than openingd flow.

### DIFF
--- a/lightningd/opening_common.c
+++ b/lightningd/opening_common.c
@@ -55,6 +55,8 @@ new_uncommitted_channel(struct peer *peer)
 	uc->peer->uncommitted_channel = uc;
 	tal_add_destructor(uc, destroy_uncommitted_channel);
 
+	uc->got_offer = false;
+
 	return uc;
 }
 
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -104,6 +104,11 @@ wallet_commit_channel(struct lightningd *ld,
 	bool option_static_remotekey;
 	bool option_anchor_outputs;
 
+	/* We cannot both be the fundee *and* have a `fundchannel_start`
+	 * command running!
+	 */
+	assert(!(uc->got_offer && uc->fc));
+
 	/* Get a key to use for closing outputs from this tx */
 	final_key_idx = wallet_get_newindex(ld);
 	if (final_key_idx == -1) {
@@ -538,6 +543,14 @@ failed:
 	tal_free(uc);
 }
 
+/* These two functions used to be a single function, pre-declaring
+ * here in order to reduce review load since the original function
+ * is not really changed at all in its operation, just need access
+ * to the cleanup part in a separate piece of code.
+ */
+static void
+opening_funder_failed_cancel_commands(struct uncommitted_channel *uc,
+				      const char *desc);
 static void opening_funder_failed(struct subd *openingd, const u8 *msg,
 				  struct uncommitted_channel *uc)
 {
@@ -554,6 +567,16 @@ static void opening_funder_failed(struct subd *openingd, const u8 *msg,
 		return;
 	}
 
+	opening_funder_failed_cancel_commands(uc, desc);
+}
+static void
+opening_funder_failed_cancel_commands(struct uncommitted_channel *uc,
+				      const char *desc)
+{
+	/* If no funding command(s) pending, do nothing.  */
+	if (!uc->fc)
+		return;
+
 	/* Tell anyone who was trying to cancel */
 	for (size_t i = 0; i < tal_count(uc->fc->cancels); i++) {
 		struct json_stream *response;
@@ -568,12 +591,17 @@ static void opening_funder_failed(struct subd *openingd, const u8 *msg,
 		was_pending(command_fail(uc->fc->cmd, LIGHTNINGD, "%s", desc));
 
 	/* Clear uc->fc, so we can try again, and so we don't fail twice
-	 * if they close. */
+	 * if they close.
+	 * This code is also used in the case where we turn out to already
+	 * be the fundee, in which case we should not have any `fc` at all,
+	 * so we definitely should clear this.
+	 */
 	uc->fc = tal_free(uc->fc);
 }
 
 struct openchannel_hook_payload {
 	struct subd *openingd;
+	struct uncommitted_channel* uc;
 	struct amount_sat funding_satoshis;
 	struct amount_msat push_msat;
 	struct amount_sat dust_limit_satoshis;
@@ -631,6 +659,7 @@ openchannel_hook_final(struct openchannel_hook_payload *payload STEALS)
 	struct subd *openingd = payload->openingd;
 	const u8 *our_upfront_shutdown_script = payload->our_upfront_shutdown_script;
 	const char *errmsg = payload->errmsg;
+	struct uncommitted_channel* uc = payload->uc;
 
 	/* We want to free this, whatever happens. */
 	tal_steal(tmpctx, payload);
@@ -640,6 +669,16 @@ openchannel_hook_final(struct openchannel_hook_payload *payload STEALS)
 		return;
 
 	tal_del_destructor2(openingd, openchannel_payload_remove_openingd, payload);
+
+	if (!errmsg) {
+		/* Plugins accepted the offer, cancel any of our
+		 * funder-side commands.  */
+		opening_funder_failed_cancel_commands(uc,
+						      "Have in-progress "
+						      "`open_channel` from "
+						      "peer");
+		uc->got_offer = true;
+	}
 
 	subd_send_msg(openingd,
 		      take(towire_openingd_got_offer_reply(NULL, errmsg,
@@ -739,6 +778,7 @@ static void opening_got_offer(struct subd *openingd,
 
 	payload = tal(openingd, struct openchannel_hook_payload);
 	payload->openingd = openingd;
+	payload->uc = uc;
 	payload->our_upfront_shutdown_script = NULL;
 	payload->errmsg = NULL;
 	if (!fromwire_openingd_got_offer(payload, msg,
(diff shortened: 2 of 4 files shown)


lightningd/opening_common.c
```diff
@@ -53,6 +53,8 @@ new_uncommitted_channel(struct peer *peer)
 	uc->peer->uncommitted_channel = uc;
 	tal_add_destructor(uc, destroy_uncommitted_channel);
 
+	uc->is_fundee = false;
```
niftynei, line 56: I'd suggest calling this `got_offer`

openingd/openingd.c
```diff
@@ -974,6 +974,14 @@ static u8 *fundee_channel(struct state *state, const u8 *open_channel_msg)
 	if (!state->upfront_shutdown_script[LOCAL])
 		state->upfront_shutdown_script[LOCAL] = no_upfront_shutdown_script(state, state);
 
+	/* Just before we accept, we tell the master that we are now
+	 * going to be a fundee, so it should fail any pending
+	 * `fundchannel_start` on this peer, and fail all future ones
+	 * as well.
+	 */
+	msg = towire_openingd_will_be_fundee(NULL);
```
niftynei, line 982: You don't need a new wire for this, see `opening_got_offer` in `lightningd/opening_control.c`.

lightningd/opening_control.c
```diff
@@ -1086,6 +1121,10 @@ static struct command_result *json_fund_channel_start(struct command *cmd,
 		return command_fail(cmd, LIGHTNINGD, "Already funding channel");
 	}
 
+	if (peer->uncommitted_channel->is_fundee) {
+		return command_fail(cmd, LIGHTNINGD, "We are the fundee");
```
niftynei, line 1125:
```suggestion
		return command_fail(cmd, LIGHTNINGD, "Have in-progress `open_channel` from peer");
```

niftynei (changes requested):
Nice debugging work on this! You don't need a new wire call to fix this. We already do a similar/exact call a few lines above, `opening_got_offer`, which is used for the `openchannel` hook. We should toggle this flag there.

ZmnSCPxj:
The reason why I did not reuse `got_offer` is because of this:

https://github.com/ElementsProject/lightning/blob/60be62a3bbc25dc849430baaaed4d61c1b4720b5/openingd/openingd.c#L965-L972

We would need to add more code in `lightningd` to clear the `is_fundee`/`got_offer` flag in case we end up rejecting it, I think (not sure, `opening` flow is unfamiliar to me).

Looking over at `lightningd/opening_control.c`, it seems to me I have to set the `got_offer` flag *after* this check, so if we got into this part of the code, we reject before we set the `got_offer` flag:

https://github.com/ElementsProject/lightning/blob/60be62a3bbc25dc849430baaaed4d61c1b4720b5/lightningd/opening_control.c#L729-L736

Then later, if a plugin tells us to fail it, we should add a check in this function:

https://github.com/ElementsProject/lightning/blob/60be62a3bbc25dc849430baaaed4d61c1b4720b5/lightningd/opening_control.c#L625-L644

Unfortunately, it seems we do not have access to the `struct uncommitted_channel` here to clear the `got_offer` flag, so I should modify `openchannel_hook_payload` to include that, is that correct?

lightningd/opening_control.c
```diff
@@ -415,6 +420,17 @@ static void opening_funder_finished(struct subd *openingd, const u8 *resp,
 	tal_free(fc->uc);
 }
 
+static void
+opening_funder_failed_cancel_commands(struct uncommitted_channel *uc,
```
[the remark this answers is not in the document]
ZmnSCPxj: I think `opening_fundee_started` is just fine where it is, and the other one is also just fine where it is.  A small declaration to me is just small and easy to skip over with proper folding.  *shrug*.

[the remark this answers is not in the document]
ZmnSCPxj: Also, modern vim understands etags just fine and we have a `make TAGS` that uses etags, declarations do not confuse etags the way they confuse ctags.

ZmnSCPxj:
Removed the new wire.  Cancel a conflicting `fundchannel_start` after plugins all agree everything is kosher.  Renamed `is_fundee` to `got_offer`.  Did ***not*** move function just to avoid pre-declaration, I think avoiding pre-declaration at all costs leads to ***awful*** code layout where later-executed functions are read first before earlier-executed functions, I think the minor cost of a few lines for pre-declaration is far cheaper than the mental gymnastics needed to understand code written in a non-straightforward order.  If I had moved `opening_funder_failed_cancel_commands` above `opening_funder_failed` just to avoid a pre-declaration, then it would have obscured that the new function is just the latter part of the original single function with no modifications.

niftynei:
So, not to get too "putting on my arm-chair psychologist hat" here, but as you did bring up the topic of "mental gymnastics" I think we should discuss the laundry list of rationales you've given so far as to why it's ok to disregard the request to do the reorg needed to remove the `static` function declaration.

Rationales offered: 
- "opening_fundee_started is just fine where it is"
- "the other one is also just fine where it is"
- "A small declaration to me is just small and easy to skip over"
- "ctags is really just a lousy beta-level software"
- "Twisting your codebase because of lousy tools is worse than just not having that tool."
- "I think avoiding pre-declaration at all costs leads to awful code layout"
- "the minor cost of a few lines for pre-declaration is far cheaper than the mental gymnastics needed"

These are all stated in the pull request comments. However, I find it peculiar that none of these are mentioned in the actual code comment you added (and vice versa, that the reason given there isn't mentioned elsewhere), which is

```
/* These two functions used to be a single function, pre-declaring
 * here in order to reduce review load since the original function
 * is not really changed at all in its operation, just need access
 * to the cleanup part in a separate piece of code.
 */
```

Which can be summarized as
- "reduce review load"

I'm not going to address the rationales given here, I merely wish to point out that you've triggered my "methinksth thou dost protesth too much" heuristic of argumentation. Usually this is an indication that there's actually another, unspoken and perhaps unconscious, rationale for as to why you'd like to keep the code the way it is, rather than following the fairly small request, I think at this point, to change it.

What do you think?

ZmnSCPxj:
Well, I could put a long treatise in the comments like so, if that is what you would prefer:

```c
/* Often, in programming, we find we need to split functions into smaller pieces,
 * simply as a consequence of refactoring code for code reuse, or in a server
 * context, simply as a consequence of having to do callback-style coding
 * in order to block for a task but not block the entire server.
 * Now, ideally you want those functions to be in a reasonable order, that is,
 * in the same order you would have read the lines if they were a single
 * large function instead of broken up into tiny pieces.
 *
 * However, C requires functions to be predeclared before first use.
 * So if you have to split up a function into two pieces, for refactoring, or
 * to make a callback for a blocking operation, and the first piece has to
 * refer to the second piece (to call it or to provide as callback function)
 *
 * - Either you predeclare the second piece and add a few lines.
 * - You twist your code so that the second piece is written before the first.
 *
 * Now, there are other reasons to predeclare, or not predeclare.
 * 
 * - If you have to do mutual recursion, no choice but to predeclare.
 * - If you want to isolate a common operation off into its own function,
 *   you should not predeclare, or should consider moving it off into
 *   its own source file.
 *   (And if it is off into its own source file, it is effectively predeclared
 *   in its own header).
 *
 * On the other hand, maintaining the proper order for executing
 * code should generally be a reason to predeclare code.
 * Predeclaration is just a few lines.
 *
 * But really, the issue is that C requires predeclaration *at all*.
 */
```

Does that satisfy your psychologist?  Or can we get on with fixing this bug?

ZmnSCPxj:
Removed the predeclaration as a separate commit so that the code-move can be reviewed separately from the code-modification.

lightningd/opening_control.c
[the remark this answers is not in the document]
rustyrussell: @niftynei is right here; C is always written bottom to top.  Sometimes judicious use of pre-declaration can reduce diff size, but best practice is usually a separate preparatory commit to reorder.

lightningd/opening_control.c
```diff
@@ -595,6 +569,27 @@ opening_funder_failed_cancel_commands(struct uncommitted_channel *uc,
 	 */
 	uc->fc = tal_free(uc->fc);
 }
+static void
+opening_funder_failed_cancel_commands(struct uncommitted_channel *uc,
```
niftynei, line 573: I think you meant to remove this?

tests/test_connection.py
```diff
@@ -2695,3 +2695,19 @@ def test_htlc_retransmit_order(node_factory, executor):
...
+def test_fundchannel_start_alternate(node_factory, executor):
+    ''' Test to see what happens if two nodes start channeling to
+    each other alternately.
+    Issue #4108
+    '''
+    l1, l2 = node_factory.get_nodes(2)
+
+    l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
+
+    l1.rpc.fundchannel_start(l2.info['id'], 100000)
```
niftynei, line 2709: if you call 'fundchannel' here, instead of `fundchannel_start`, then at some point in the future (soonTM), it'll also test the v2 channel open code.

niftynei:
bump @ZmnSCPxj

tests/test_connection.py
ZmnSCPxj: It has to be `fundchannel_start`, since the race condition being tested is when both peers perform `fundchannel_start`.  Otherwise once `fundchannel` returns, the state is different and a different non-buggy part of the code is triggered.

Adding the changes below results in the test passing even without code changes, so the test *has to* use `fundchannel_start`:

```diff
     '''
     l1, l2 = node_factory.get_nodes(2)
+    l1.fundwallet(10**6 + 1000000)
 
     l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
 
-    l1.rpc.fundchannel_start(l2.info['id'], 100000)
+    l1.rpc.fundchannel(l2.info['id'], 100000)
 
     fut = executor.submit(l2.rpc.fundchannel_start, l1.info['id'], 100000)
     with pytest.raises(RpcError):
```

niftynei (approved):
ACK 1ace2ad

Rebased on master to resolve conflict with (merged) #4126

PR #4117, opened 2020-10-08
pyln-client has wrong `listpayments` method, change to `listpays`

fiatjaf:


rustyrussell:
This is actually left over from the old `listpayments`.  Thanks!

PR #4127, opened 2020-10-14
Add sha256 check to configure for OpenBSD 

grubles:
OpenBSD's `sha256sum` equivalent is `sha256`. This adds support in `configure` for it.


kristapsk:
This should work for other BSD's too, at least for FreeBSD. See https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/03075a047b2cd09d3aee97fd525935a6004a9c72/install.sh#L14.

PR #4129, opened 2020-10-15
Fuzz testing integration

darosior:
This introduces coverage-guided fuzz testing using [libfuzzer](https://www.llvm.org/docs/LibFuzzer.html).

## Why

We've been[ talking about it](https://github.com/ElementsProject/lightning/issues/2662) for some time now, and i'm sure it could help us to uncover trivial bugs with minimal overhead, more sneaky ones more easily than with unit tests, and finally to avoid regressions by maintaining a seed corpus to ran patches against.

Going further, [structure-aware fuzzing](https://github.com/google/fuzzing/blob/master/docs/structure-aware-fuzzing.md) could help us with [Bitcoin Script creation](https://github.com/jonasnick/bitcoinconsensus_testcases) or even with the state machine  which while more intrusive because crypto seems plausible because of the daemons separation and messages.

For now, this PR only introduces the basis for integration with primitives and rudimentary targets (i focused on `common/`). Still, this found some bugs (eg amounts string parsing / formatting) and new targets could help with e.g. our JSON parser or the [gossip store](https://github.com/ElementsProject/lightning/issues?q=is%3Aissue+corrupt+gossip_store).

## Integration

This extends our `tests/` directory by moving the Python functional tests to the `tests/functional/` subdirectory (sorry for the conflicts..) and creates a new sub: `tests/fuzz/`.

Inspired by the unit tests, all targets are named `fuzz-[target_name]`.
Not inspired by the unit tests, all targets are built in `tests/fuzz/`. Even though there are only `common/`-specific targets for now, i think it makes sense to keep them contained in a single directory.

A `libfuzz.h` abstracts out the interface to libFuzzer for potential future support for other fuzzers, and `run.py` allows to manage them in batches. This was inspired by the integration in bitcoin-core but i guess it's a pretty common pattern.

## What next

- a)
I propose to create a new repository (eg in https://github.com/lightningd/ ?) for tracking the seed corpus that we could update if we generate new coverage locally.
This could be pulled in a new CI job that checks PRs against each [target + seeds] once to check for regression. What do you think about using Github Action for this ?
- b)
Write more targets
- c)
Generate more coverage
- d)
Go to b) :)

### DIFF
--- a/common/amount.c
+++ b/common/amount.c
@@ -33,6 +33,9 @@ const char *fmt_amount_msat_btc(const tal_t *ctx,
 				const struct amount_msat *msat,
 				bool append_unit)
 {
+	if (msat->millisatoshis == 0)
+		return tal_fmt(ctx, append_unit ? "0btc" : "0");
+
 	return tal_fmt(ctx, "%"PRIu64".%011"PRIu64"%s",
 		       msat->millisatoshis / MSAT_PER_BTC,
 		       msat->millisatoshis % MSAT_PER_BTC,
@@ -49,6 +52,9 @@ const char *fmt_amount_sat_btc(const tal_t *ctx,
 			       const struct amount_sat *sat,
 			       bool append_unit)
 {
+	if (sat->satoshis == 0)
+		return tal_fmt(ctx, append_unit ? "0btc" : "0");
+
 	return tal_fmt(ctx, "%"PRIu64".%08"PRIu64"%s",
 		       sat->satoshis / SAT_PER_BTC,
 		       sat->satoshis % SAT_PER_BTC,
@@ -80,7 +86,8 @@ static bool breakup(const char *str, size_t slen,
 	*suffix_len = 0;
 
 	for (i = 0;; i++) {
-		if (i >= slen)
+		/* The string may be null-terminated. */
+		if (i >= slen || str[i] == '\0')
 			return i != 0;
 		if (cisdigit(str[i]))
 			(*whole_number_len)++;
@@ -93,7 +100,7 @@ static bool breakup(const char *str, size_t slen,
 		*post_decimal_ptr = str + i;
 		for (;; i++) {
 			/* True if > 0 decimals. */
-			if (i >= slen)
+			if (i >= slen || str[i] == '\0')
 				return str + i != *post_decimal_ptr;
 			if (cisdigit(str[i]))
 				(*post_decimal_len)++;
@@ -168,14 +175,18 @@ bool parse_amount_msat(struct amount_msat *msat, const char *s, size_t slen)
 
 	if (!post_decimal_ptr && !suffix_ptr)
 		return from_number(&msat->millisatoshis, s, whole_number_len, 0);
-	if (!post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "msat"))
+	if (!post_decimal_ptr && memstarts_str(suffix_ptr, suffix_len, "msat"))
 		return from_number(&msat->millisatoshis, s, whole_number_len, 0);
-	if (!post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "sat"))
+	if (!post_decimal_ptr && memstarts_str(suffix_ptr, suffix_len, "sat"))
 		return from_number(&msat->millisatoshis, s, whole_number_len, 3);
-	if (post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "btc"))
-		return from_numbers(&msat->millisatoshis,
-				    s, whole_number_len, 11,
-				    post_decimal_ptr, post_decimal_len);
+	if (memstarts_str(suffix_ptr, suffix_len, "btc")) {
+		if (post_decimal_len > 0)
+			return from_numbers(&msat->millisatoshis,
+					    s, whole_number_len, 11,
+					    post_decimal_ptr, post_decimal_len);
+		return from_number(&msat->millisatoshis, s, whole_number_len, 11);
+	}
+
 	return false;
 }
 
@@ -198,21 +209,24 @@ bool parse_amount_sat(struct amount_sat *sat, const char *s, size_t slen)
 
 	if (!post_decimal_ptr && !suffix_ptr)
 		return from_number(&sat->satoshis, s, whole_number_len, 0);
-	if (!post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "sat"))
+	if (!post_decimal_ptr && memstarts_str(suffix_ptr, suffix_len, "sat"))
 		return from_number(&sat->satoshis, s, whole_number_len, 0);
-	if (!post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "msat")) {
+	if (!post_decimal_ptr && memstarts_str(suffix_ptr, suffix_len, "msat")) {
 		if (!memends(s, whole_number_len, "000", strlen("000"))) {
-			if (memeqstr(s, whole_number_len, "0"))
+			if (memstarts_str(s, whole_number_len, "0"))
 				return from_number(&sat->satoshis, s,
 						   whole_number_len, 0);
 			return false;
 		}
 		return from_number(&sat->satoshis, s, whole_number_len - 3, 0);
 	}
-	if (post_decimal_ptr && memeqstr(suffix_ptr, suffix_len, "btc"))
-		return from_numbers(&sat->satoshis,
-				    s, whole_number_len, 8,
-				    post_decimal_ptr, post_decimal_len);
+	if (memstarts_str(suffix_ptr, suffix_len, "btc")) {
+		if (post_decimal_len > 0)
+			return from_numbers(&sat->satoshis,
+					    s, whole_number_len, 8,
+					    post_decimal_ptr, post_decimal_len);
+		return from_number(&sat->satoshis, s, whole_number_len, 8);
+	}
 
 	return false;
 }
--- a/common/test/run-amount.c
+++ b/common/test/run-amount.c
@@ -128,7 +128,6 @@ int main(void)
 	PASS_MSAT(&msat, "1.234567890btc", 123456789000);
 	PASS_MSAT(&msat, "1.2345678901btc", 123456789010);
 	PASS_MSAT(&msat, "1.23456789012btc", 123456789012);
-	FAIL_MSAT(&msat, "1btc");
 	FAIL_MSAT(&msat, "1.000000000000btc");
 	FAIL_MSAT(&msat, "-1.23456789btc");
 	FAIL_MSAT(&msat, "-1.23456789012btc");
@@ -177,7 +176,6 @@ int main(void)
 	PASS_SAT(&sat, "1.2345678btc", 123456780);
 	PASS_SAT(&sat, "1.23456789btc", 123456789);
 	FAIL_SAT(&sat, "1.234567890btc");
-	FAIL_SAT(&sat, "1btc");
 	FAIL_SAT(&sat, "-1.23456789btc");
 
 	/* Overflowingly big. */
--- a/doc/FUZZING.md
+++ b/doc/FUZZING.md
@@ -1,0 +1,69 @@
+# Fuzz testing
+
+C-lightning currently supports coverage-guided fuzz testing using [LLVM's libfuzzer](https://www.llvm.org/docs/LibFuzzer.html)
+when built with `clang`.
+
+The goal of fuzzing is to generate mutated -and often unexpected- inputs (`seed`s) to pass
+to (parts of) a program (`target`) in order to make sure the codepaths used:
+- do not crash
+- are valid (if combined with sanitizers)
+The generated seeds can be stored and form a `corpus`, which we try to optimise (don't
+store two seeds that lead to the same codepath).
+
+For more info about fuzzing see [here](https://github.com/google/fuzzing/tree/master/docs),
+and for more about `libfuzzer` in particular see [here](https://www.llvm.org/docs/LibFuzzer.html).
+
+
+## Build the fuzz targets
+
+In order to build the C-lightning binaries with code coverage you will need a recent
+[clang](http://clang.llvm.org/). The more recent the compiler version the better.
+
+Then you'll need to enable support at configuration time. You likely want to enable
+a few sanitizers for bug detections as well as experimental features for an extended
+coverage (not required though).
+
+```
+DEVELOPER=1 EXPERIMENTAL_FEATURES=1 ASAN=1 UBSAN=1 VALGRIND=0 FUZZING=1 CC=clang ./configure && make
+```
+
+The targets will be built in `tests/fuzz/` as `fuzz-` binaries.
+
+
+## Run one or more target(s)
+
+You can run each target independently. Pass `-help=1` to see available options, for
+example:
+```
+./tests/fuzz/fuzz-addr -help=1
+```
+
+Otherwise, you can use the Python runner to either run the targets against a given seed
+corpus:
+```
+./tests/fuzz/run.py fuzz_corpus -j2
+```
+Or extend this corpus:
+```
+./tests/fuzz/run.py fuzz_corpus -j2 --generate --runs 12345
+```
+
+The latter will run all targets two by two `12345` times.
+
+If you want to contribute new seeds, be sure to merge your corpus with the main one:
+```
+./tests/fuzz/run.py my_locally_extended_fuzz_corpus -j2 --generate --runs 12345
+./tests/fuzz/run.py main_fuzz_corpus --merge_dir my_locally_extended_fuzz_corpus
+```
+
+
+## Write new fuzzing targets
+
+In order to write a new target:
+ - include the `libfuzz.h` header
+ - fill two functions: `init()` for static stuff and `run()` which will be called
+     repeatedly with mutated data.
+ - read about [what makes a good fuzz target](https://github.com/google/fuzzing/blob/master/docs/good-fuzz-target.md).
+
+A simple example is [`fuzz-addr`](tests/fuzz/fuzz-addr.c). It setups the chainparams and
+context (wally, tmpctx, ..) in `init()` then bruteforces the bech32 encoder in `run()`.
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -13,6 +13,7 @@ FUZZ_COMMON_OBJS :=					\
 	common/base32.o					\
 	common/base64.o					\
 	common/bech32.o					\
+	common/bip32.o					\
 	common/bigsize.o				\
 	common/json.o					\
 	common/json_stream.o				\
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -15,6 +15,7 @@ FUZZ_COMMON_OBJS :=					\
 	common/bech32.o					\
 	common/bip32.o					\
 	common/bigsize.o				\
+	common/channel_id.o				\
 	common/json.o					\
 	common/json_stream.o				\
 	common/wireaddr.o				\
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -7,12 +7,7 @@ FUZZ_TARGETS_SRC := $(wildcard tests/fuzz/fuzz-*.c)
 FUZZ_TARGETS_OBJS := $(FUZZ_TARGETS_SRC:.c=.o)
 FUZZ_TARGETS_BIN := $(FUZZ_TARGETS_SRC:.c=)
 
-FUZZ_COMMON_OBJS :=			\
-	common/utils.o
-$(FUZZ_TARGETS_OBJS): $(COMMON_HEADERS) $(WIRE_HEADERS) $(COMMON_SRC)
-$(FUZZ_TARGETS_BIN): $(LIBFUZZ_OBJS) $(FUZZ_COMMON_OBJS) $(BITCOIN_OBJS)
-
-tests/fuzz/fuzz-addr:					\
+FUZZ_COMMON_OBJS :=					\
 	common/amount.o					\
 	common/addr.o					\
 	common/base32.o					\
@@ -22,9 +17,12 @@ tests/fuzz/fuzz-addr:					\
 	common/json_stream.o				\
 	common/wireaddr.o				\
 	common/type_to_string.o				\
+	common/utils.o					\
 	wire/fromwire.o					\
 	wire/onion_wiregen.o				\
 	wire/towire.o
+$(FUZZ_TARGETS_OBJS): $(COMMON_HEADERS) $(WIRE_HEADERS) $(COMMON_SRC)
+$(FUZZ_TARGETS_BIN): $(LIBFUZZ_OBJS) $(FUZZ_COMMON_OBJS) $(BITCOIN_OBJS)
 
 ALL_C_SOURCES += $(FUZZ_TARGETS_SRC) $(LIBFUZZ_SRC)
 ALL_FUZZ_TARGETS += $(FUZZ_TARGETS_BIN)
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -1,0 +1,30 @@
+LIBFUZZ_SRC := tests/fuzz/libfuzz.c
+LIBFUZZ_HEADERS := $(LIBFUZZ_SRC:.c=.h)
+LIBFUZZ_OBJS := $(LIBFUZZ_SRC:.c=.o)
+
+
+FUZZ_TARGETS_SRC := $(wildcard tests/fuzz/fuzz-*.c)
+FUZZ_TARGETS_OBJS := $(FUZZ_TARGETS_SRC:.c=.o)
+FUZZ_TARGETS_BIN := $(FUZZ_TARGETS_SRC:.c=)
+
+FUZZ_COMMON_OBJS :=			\
+	common/utils.o
+$(FUZZ_TARGETS_OBJS): $(COMMON_HEADERS) $(WIRE_HEADERS) $(COMMON_SRC)
+$(FUZZ_TARGETS_BIN): $(LIBFUZZ_OBJS) $(FUZZ_COMMON_OBJS) $(BITCOIN_OBJS)
+
+tests/fuzz/fuzz-addr:					\
+	common/amount.o					\
+	common/addr.o					\
+	common/base32.o					\
+	common/bech32.o					\
+	common/bigsize.o				\
+	common/json.o					\
+	common/json_stream.o				\
+	common/wireaddr.o				\
+	common/type_to_string.o				\
+	wire/fromwire.o					\
+	wire/onion_wiregen.o				\
+	wire/towire.o
+
+ALL_C_SOURCES += $(FUZZ_TARGETS_SRC) $(LIBFUZZ_SRC)
+ALL_FUZZ_TARGETS += $(FUZZ_TARGETS_BIN)
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -11,6 +11,7 @@ FUZZ_COMMON_OBJS :=					\
 	common/amount.o					\
 	common/addr.o					\
 	common/base32.o					\
+	common/base64.o					\
 	common/bech32.o					\
 	common/bigsize.o				\
 	common/json.o					\
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -15,10 +15,13 @@ FUZZ_COMMON_OBJS :=					\
 	common/bech32.o					\
 	common/bip32.o					\
 	common/bigsize.o				\
+	common/close_tx.o				\
 	common/channel_id.o				\
+	common/permute_tx.o				\
 	common/json.o					\
 	common/json_stream.o				\
 	common/wireaddr.o				\
+	common/setup.o					\
 	common/type_to_string.o				\
 	common/utils.o					\
 	wire/fromwire.o					\
--- a/tests/fuzz/Makefile
+++ b/tests/fuzz/Makefile
@@ -7,7 +7,7 @@ FUZZ_TARGETS_SRC := $(wildcard tests/fuzz/fuzz-*.c)
 FUZZ_TARGETS_OBJS := $(FUZZ_TARGETS_SRC:.c=.o)
 FUZZ_TARGETS_BIN := $(FUZZ_TARGETS_SRC:.c=)
 
-FUZZ_COMMON_OBJS :=					\
+FUZZ_COMMON_OBJS := \
 	common/amount.o					\
 	common/addr.o					\
 	common/base32.o					\
@@ -15,18 +15,40 @@ FUZZ_COMMON_OBJS :=					\
 	common/bech32.o					\
 	common/bip32.o					\
 	common/bigsize.o				\
+	common/channel_config.o				\
 	common/close_tx.o				\
 	common/channel_id.o				\
+	common/daemon.o					\
+	common/daemon_conn.o				\
+	common/derive_basepoints.o			\
+	common/fee_states.o				\
+	common/htlc_state.o				\
 	common/permute_tx.o				\
+	common/initial_channel.o			\
+	common/initial_commit_tx.o			\
 	common/json.o					\
 	common/json_stream.o				\
+	common/key_derive.o				\
+	common/keyset.o					\
+	common/msg_queue.o				\
+	common/memleak.o				\
+	common/node_id.o				\
 	common/wireaddr.o				\
 	common/setup.o					\
+	common/status.o					\
+	common/status_wire.o				\
+	common/status_wiregen.o				\
 	common/type_to_string.o				\
 	common/utils.o					\
+	common/version.o				\
 	wire/fromwire.o					\
 	wire/onion_wiregen.o				\
-	wire/towire.o
+	wire/peer_wire.o				\
+	wire/peer_wiregen.o				\
+	wire/towire.o					\
+	wire/wire_io.o					\
+	wire/wire_sync.o
+
 $(FUZZ_TARGETS_OBJS): $(COMMON_HEADERS) $(WIRE_HEADERS) $(COMMON_SRC)
 $(FUZZ_TARGETS_BIN): $(LIBFUZZ_OBJS) $(FUZZ_COMMON_OBJS) $(BITCOIN_OBJS)
 
--- a/tests/fuzz/fuzz-amount.c
+++ b/tests/fuzz/fuzz-amount.c
@@ -1,0 +1,61 @@
+#include <assert.h>
+#include <stdint.h>
+#include <tests/fuzz/libfuzz.h>
+
+#include <ccan/tal/tal.h>
+#include <common/amount.h>
+#include <common/utils.h>
+
+void init(int *argc, char ***argv)
+{
+}
+
+void run(const uint8_t *data, size_t size)
+{
+	struct amount_msat msat;
+	struct amount_sat sat;
+	char *string;
+	uint8_t *buf;
+	const char *fmt_msat, *fmt_msatbtc, *fmt_sat, *fmt_satbtc;
+
+
+	/* We should not crash when parsing any string. */
+
+	string = tal_arr(NULL, char, size);
+	for (size_t i = 0; i < size; i++)
+		string[i] = (char) data[i] % (CHAR_MAX + 1);
+
+	parse_amount_msat(&msat, string, tal_count(string));
+	parse_amount_sat(&sat, string, tal_count(string));
+	tal_free(string);
+
+
+	/* Same with the wire primitives. */
+
+	buf = tal_arr(NULL, uint8_t, 8);
+
+	msat = fromwire_amount_msat(&data, &size);
+	towire_amount_msat(&buf, msat);
+	sat = fromwire_amount_sat(&data, &size);
+	towire_amount_sat(&buf, sat);
+
+	tal_free(buf);
+
+
+	/* Format should inconditionally produce valid amount strings according to our
+	 * parser */
+
+	fmt_msat = fmt_amount_msat(NULL, &msat);
+	fmt_msatbtc = fmt_amount_msat_btc(NULL, &msat, true);
+	assert(parse_amount_msat(&msat, fmt_msat, tal_count(fmt_msat)));
+	assert(parse_amount_msat(&msat, fmt_msatbtc, tal_count(fmt_msatbtc)));
+	tal_free(fmt_msat);
+	tal_free(fmt_msatbtc);
+
+	fmt_sat = fmt_amount_sat(NULL, &sat);
+	fmt_satbtc = fmt_amount_sat_btc(NULL, &sat, true);
+	assert(parse_amount_sat(&sat, fmt_sat, tal_count(fmt_sat)));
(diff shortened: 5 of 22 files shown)


common/amount.c
```diff
@@ -49,6 +52,9 @@ const char *fmt_amount_sat_btc(const tal_t *ctx,
 			       const struct amount_sat *sat,
 			       bool append_unit)
 {
+	if (sat->satoshis == 0)
+		return tal_fmt(ctx, append_unit ? "0btc" : "0");
```
niftynei, line 56: since these aren't runtime variable length strings, you don't need `tal_fmt` here

```
return append_unit ? "0btc" : "0";
```

common/amount.c
```diff
@@ -33,6 +33,9 @@ const char *fmt_amount_msat_btc(const tal_t *ctx,
 				const struct amount_msat *msat,
 				bool append_unit)
 {
+	if (msat->millisatoshis == 0)
+		return tal_fmt(ctx, append_unit ? "0btc" : "0");
```
niftynei, line 37: same as other comment, no need for `tal_fmt`

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
+# Fuzz testing
+
+C-lightning currently supports coverage-guided fuzz testing using [LLVM's libfuzzer](https://www.llvm.org/docs/LibFuzzer.html)
+when built with clang.
+
+
+## Build the fuzz targets
+
+In order to build the C-lightning binaries with code coverage you will need a recent
+[clang](http://clang.llvm.org/). The younger the compiler version the better.
```
niftynei, line 10: what is a "younger compiler version"?

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+
+Then you'll need to enable support at configuration time. You likely want to enable
+a few sanitizers for bug detections as well as experimental features for an extended
+coverage (no required though).
```
niftynei, line 14:
```suggestion
coverage (not required though).
```

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+```
+DEVELOPER=1 EXPERIMENTAL_FEATURES=1 ASAN=1 UBSAN=1 VALGRIND=0 FUZZING=1 CC=clang ./configure && make
+```
+
+The targets will be built in `tests/fuzz/` as `fuzz-` binaries.
+
+
+## Run one or more target(s)
+
+You can run each target independently. Pass `-help=1` to see available options, for
```
niftynei, line 25: is this a library flag? if not you should use two dashes for flags! (we are not bitcoind). also it'd be nicer if you can make this not require a value.
```suggestion
You can run each target independently. Pass `--help` to see available options, for
```

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+corpus:
+```
+./tests/fuzz/run.py fuzz_corpus -j2
+```
+Or extend this corpus:
+```
+./tests/fuzz/run.py fuzz_corpus -j2 --generate --runs 12345
+```
+
+The latter will run all targets two by two `12345` times.
```
niftynei, line 41: i don't understand what `two by two 12345 times` means.

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Run one or more target(s)
+
+You can run each target independently. Pass `-help=1` to see available options, for
+example:
+```
+./tests/fuzz/fuzz-addr -help=1
+```
+
+Otherwise, you can use the Python runner to either run the targets against a given seed
```
niftynei, line 31: what is a `seed`?

tests/functional/test_plugin.py
```diff
@@ -202,7 +202,7 @@ def test_plugin_slowinit(node_factory):
     n = node_factory.get_node()
 
     with pytest.raises(RpcError, match='failed to respond to \'init\' in time, terminating.'):
-        n.rpc.plugin_start(os.path.join(os.getcwd(), "tests/plugins/slow_init.py"))
+        n.rpc.plugin_start(os.path.join(os.getcwd(), "tests/functional/plugins/slow_init.py"))
```
niftynei, line 205: moving all existing tests seems incredibly unnecessary. what's wrong with having a `fuzzing` directory in along with all the other functional tests?

tests/fuzz/fuzz-bech32.c
```diff
@@ -0,0 +1,49 @@
...
+
+#include <common/bech32.h>
+
+void init(int *argc, char ***argv)
+{
+}
+
+void run(const uint8_t *data, size_t size)
+{
+	const char hrp_inv[5] = "lnbc\0", hrp_addr[3] = "bc\0";
```
niftynei, line 13: do these constants not exist elsewhere in the project?

tests/fuzz/fuzz-bech32.c
```diff
@@ -0,0 +1,49 @@
...
+
+	data_out = malloc(size);
+
+	/* This is also used as part of sign and check message. */
+	data_out_len = 0;
+	bech32_convert_bits(data_out, &data_out_len, 8, data, size, 5, 1);
+	data_out_len = 0;
+	bech32_convert_bits(data_out, &data_out_len, 8, data, size, 5, 0);
+
+	addr = malloc(73 + strlen(hrp_addr));
```
niftynei, line 39: where does this magic number `73` come from?

tests/fuzz/fuzz-bip32.c
```diff
@@ -0,0 +1,52 @@
...
+	uint8_t *wire_buff;
+	const uint8_t **xkey_chunks, **ver_chunks, *wire_ptr;
+	size_t wire_max;
+
+	if (size < BIP32_SERIALIZED_LEN)
+		return;
+
+	xkey_chunks = get_chunks(NULL, data, size, BIP32_SERIALIZED_LEN);
+	for (size_t i = 0; i < tal_count(xkey_chunks); i++) {
+		wire_max = tal_count(xkey_chunks[i]);
```
niftynei, line 27: `tal_bytelen`?

tests/fuzz/fuzz-bip32.c
```diff
@@ -0,0 +1,52 @@
...
+
+void init(int *argc, char ***argv)
+{
+}
+
+void run(const uint8_t *data, size_t size)
+{
+	struct ext_key xkey;
+	struct bip32_key_version version;
+	uint8_t *wire_buff;
```
niftynei, line 18: why eschew the short_types??

tests/fuzz/fuzz-channel_id.c
```diff
@@ -0,0 +1,70 @@
...
+	uint16_t vout;
+	const uint8_t **v1_chunks, **v2_chunks, **marshal_chunks;
+	const uint8_t *wire_ptr;
+	size_t wire_max;
+	uint8_t *wire_buf;
+
+	if (size < 34)
+		return;
+
+	/* 32 + 2 */
```
niftynei, line 30: what is `32 + 2` referring to?

tests/fuzz/fuzz-channel_id.c
```diff
@@ -0,0 +1,70 @@
...
+		fromwire_pubkey(&wire_ptr, &wire_max, &basepoint_1);
+		if (!wire_ptr)
+			continue;
+
+		wire_max = PUBKEY_CMPR_LEN;
+		fromwire_pubkey(&wire_ptr, &wire_max, &basepoint_2);
+		if (!wire_ptr)
+			continue;
+
+		derive_channel_id_v2(&chan_id, &basepoint_1, &basepoint_2);
```
niftynei, line 56: :heart_decoration:

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	struct pubkey *pk1, *pk2;
+
+	/* create_close_tx wants:
+	 * - 3 scripts: 3 * N bytes
+	 * - 1 txid: 32 bytes
+	 * - 1 u32: 4 bytes
+	 * - 4 amount_sat: 4 * 8 bytes
+	 *
+	 * Since the output scripts size is not very restricted (and the shutdown_scriptpukeys
+	 * are not by the spec either), we also try to vary their length.
```
niftynei, line 40: what does 'not by the spec' mean?

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	const uint8_t *our_script, *their_script, *funding_script;
+	struct pubkey *pk1, *pk2;
+
+	/* create_close_tx wants:
+	 * - 3 scripts: 3 * N bytes
+	 * - 1 txid: 32 bytes
+	 * - 1 u32: 4 bytes
+	 * - 4 amount_sat: 4 * 8 bytes
+	 *
+	 * Since the output scripts size is not very restricted (and the shutdown_scriptpukeys
```
niftynei, line 39:
```suggestion
	 * Since the output script size is not restricted (and the shutdown_scriptpukeys
```

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	 * are not by the spec either), we also try to vary their length.
+	 * Therefore, we allocate the entire remaining bytes to scripts.
+	 */
+	min_size = 8 * 3 + 4 + 32;
```
niftynei, line 43: these numbers don't match your above `create_close_tx` comment. assuming this is 4 amount_sat + u32 + txid, you should have 

```
min_size = 8 * 4 + 4 + 32;
```

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	if (size < min_size + 2)
+		return;
+
+	script_size = (size - min_size) / 2;
+	wire_ptr = data;
+
+	wire_max = 8;
+	to_us = fromwire_amount_sat(&wire_ptr, &wire_max);
+	wire_max = 8;
```
niftynei, line 52: should we verify the wire_max value is ok before resetting it?

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	to_them = fromwire_amount_sat(&wire_ptr, &wire_max);
+	wire_max = 8;
+	dust_limit = fromwire_amount_sat(&wire_ptr, &wire_max);
```
niftynei, line 55: you said `4 amount_sat` above, but only use 3 here

tests/fuzz/libfuzz.c
```diff
@@ -0,0 +1,35 @@
+#include <tests/fuzz/libfuzz.h>
+
+#include <ccan/tal/tal.h>
+#include <common/utils.h>
+#include <string.h>
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
```
niftynei, line 7: camel case is not a kosher way to write a function name in C? Is there something i'm missing about these names?
```suggestion
int llvm_fuzzer_test_one_input(const uint8_t *data, size_t size);
```

tests/fuzz/libfuzz.c
```diff
@@ -0,0 +1,35 @@
...
+int LLVMFuzzerInitialize(int *argc, char ***argv);
```
niftynei, line 8:
```suggestion
int llvm_fuzzer_initialize(int *argc, char ***argv);
```

tests/fuzz/libfuzz.h
```diff
@@ -0,0 +1,18 @@
+#ifndef LIGHTNING_TESTS_FUZZ_LIBFUZZ_H
+#define LIGHTNING_TESTS_FUZZ_LIBFUZZ_H
+
+#include <stddef.h>
+#include <stdint.h>
+
+/* Called once before running the target. Use it to setup the testing
```
niftynei, line 7: do we also need a teardown function for doing cleanup?

tests/fuzz/run.py
```diff
@@ -0,0 +1,78 @@
...
+import logging
+import os
+import re
+import subprocess
+
+from concurrent.futures import ProcessPoolExecutor, as_completed
+
+
+def job(command):
+    """Takes a list of str, and runs it as a subprocess."""
```
niftynei, line 12: what does a 'list of str' mean?
```suggestion
    """Takes a list of command strings??, and runs it as a subprocess."""
```

tests/fuzz/Makefile
```diff
@@ -7,12 +7,7 @@ FUZZ_TARGETS_SRC := $(wildcard tests/fuzz/fuzz-*.c)
 FUZZ_TARGETS_OBJS := $(FUZZ_TARGETS_SRC:.c=.o)
 FUZZ_TARGETS_BIN := $(FUZZ_TARGETS_SRC:.c=)
 
-FUZZ_COMMON_OBJS :=			\
```
niftynei: this feels like it belongs in the previous commit

common/test/run-amount.c
```diff
@@ -128,7 +128,6 @@ int main(void)
 	PASS_MSAT(&msat, "1.234567890btc", 123456789000);
 	PASS_MSAT(&msat, "1.2345678901btc", 123456789010);
 	PASS_MSAT(&msat, "1.23456789012btc", 123456789012);
-	FAIL_MSAT(&msat, "1btc");
```
niftynei, line 131: why are these cases removed?

tests/fuzz/fuzz-initial_channel.c
```diff
@@ -0,0 +1,79 @@
...
+#include <common/initial_channel.h>
+#include <common/setup.h>
+#include <common/status.h>
+#include <common/utils.h>
+#include <stdio.h>
+#include <wire/wire.h>
+
+void init(int *argc, char ***argv)
+{
+	common_setup("fuzzer");
```
niftynei, line 23: if you had a teardown function you could call the other `common_teardown` (or whatever the name of it is) there

niftynei (commented):
Cool! Bunch of small nits. I think my biggest complaint is that there's no real reason to move all the existing CI tests down a directory. There's no reason we can't have `tests/fuzzing` exist while also leaving all the other tests where they are.

Didn't really give the Makefiles a good looksee, as that's not really my particular wheelhouse.

doc/FUZZING.md
darosior: A frenchism i guess: i meant it by opposition to older version :)

doc/FUZZING.md
darosior: Library flag :/

> we are not bitcoind

x)

doc/FUZZING.md
darosior: An input (raw bytes) given to the target, we store seeds that increase code coverage.

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+corpus:
+```
+./tests/fuzz/run.py fuzz_corpus -j2
+```
+Or extend this corpus:
```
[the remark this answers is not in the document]
darosior: A set of seeds.

tests/functional/test_plugin.py
darosior: Yeah it might seem overkill but it looked like the right thing to me: `tests` right now contains our functional tests, and mixing up fuzz testing and functional tests seems to be a mess..

tests/fuzz/fuzz-amount.c
```diff
@@ -0,0 +1,62 @@
...
+	sat = fromwire_amount_sat(&data, &size);
+	towire_amount_sat(&buf, sat);
+
+	tal_free(buf);
+
+
+	/* Format should inconditionally produce valid amount strings according to our
+	 * parser */
+
+	fmt_msat = fmt_amount_msat(NULL, &msat);
```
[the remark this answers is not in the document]
darosior: I opportunistically used tmpctx for the last targets as i had to `setup_common()` anyways. No particular reason.

tests/fuzz/fuzz-bech32.c
```diff
@@ -0,0 +1,49 @@
...
+void run(const uint8_t *data, size_t size)
+{
+	const char hrp_inv[5] = "lnbc\0", hrp_addr[3] = "bc\0";
+	char *bech32_str, *hrp_out, *addr;
+	uint8_t *data_out;
+	size_t data_out_len;
+	int wit_version;
+
+	/* Buffer size is defined in each function's doc comment. */
+	bech32_str = malloc(size + strlen(hrp_inv) + 8);
```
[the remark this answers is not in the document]
darosior: ?

doc/FUZZING.md
niftynei: So each process does 12345 runs? Maybe something like 

```
Each process runs all targets X times, so `-j2 --runs 12345` will result in 24690 total fuzz attempts.
```

tests/fuzz/fuzz-bech32.c
darosior: As commented above:
> Buffer size is defined in each function's doc comment.

doc/FUZZING.md
niftynei: If i'm understanding this correctly, this is counter-intuitive to how most parallelization works. I'd expect the total runs to remain constant, with the work divided amongst the different processes (so speeding it up by half instead of doubling the work).

tests/fuzz/fuzz-close_tx.c
darosior: Am i wrong that the spec does not restrict the size of the `shutdown_scriptpubkeys` ?

doc/FUZZING.md
niftynei: Where do the seeds come from if we don't store them? You might add a short description to this document that explains the general idea of how fuzzing works.

doc/FUZZING.md
```diff
@@ -0,0 +1,59 @@
...
+If you want to contribute new seeds, be sure to merge your corpus with the main one:
+```
+./tests/fuzz/run.py my_locally_extended_fuzz_corpus -j2 --generate --runs 12345
+./tests/fuzz/run.py main_fuzz_corpus --merge_dir my_locally_extended_fuzz_corpus
+```
+
+
+## Write new fuzzing targets
+
+In order to write a new target:
```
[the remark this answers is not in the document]
niftynei: These are all good definitions and 100% belong in this document.

doc/FUZZING.md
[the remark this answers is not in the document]
niftynei: Add this definition to this doc.

tests/functional/test_plugin.py
niftynei: You're not mixing them though, the fuzz tests all live in `tests/fuzz`

tests/fuzz/fuzz-close_tx.c
niftynei: afaik that is correct. I'd suggest rewording this to 

```
Since output script (and shutdown_scriptpubkey) size is not restricted...
```

cdecker (commented):
Awesome PR @darosior, I've long wanted to look into fuzzing, and with this I have almost a guided tour to help me learn. Haven't looked too closely yet, but I'll grab a coco and start learning asap.

However, I share @niftynei's feeling re moving all tests down one directory level:

 1. It breaks almost all PRs
 2. Those are likely the most edited files in the whole repo, hiding them in ya directory means some more typing 😜

doc/FUZZING.md
darosior: Actually the work isn't divided, each process is really independant. The reason to have a multiprocess pool is just that each process can use only one core and be CPU bound (usually RAM usage on fuzz targets is kept low), so you can run many of them in parallel (`X` times each).

doc/FUZZING.md
darosior: > Where do the seeds come from if we don't store them?

From the ~~blackbox~~ library which uses genetic algorithms to derive mutation of the data so that it reaches more code paths (themselves detected using compile time marks).

> You might add a short description to this document

I feel like anything i could add would be vague and not necessarily helpful to someone who wants to write new targets, probably better to follow the link at the top ? (which gives a good introduction, applied to the tool we are using)

tests/fuzz/fuzz-close_tx.c
darosior: Yeah i need to update that comment. Thanks!

tests/fuzz/libfuzz.c
darosior: > camel case is not a kosher way to write a function name in C?

Agreed

> Is there something i'm missing about these names?

Yeah: this is `libFuzzer`'s API :-)

tests/fuzz/run.py
darosior: `str` is the string type in Python, and most Python commands treating shell commands take it as `["path_to_bin", "--param-a", "param-b"]` so i did the same

tests/fuzz/run.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
darosior: :p

common/test/run-amount.c
darosior: This belongs to the next commit in which i change the parsing

tests/fuzz/libfuzz.c
```diff
@@ -14,3 +18,18 @@ int LLVMFuzzerInitialize(int *argc, char ***argv) {
...
 	return 0;
 }
+
+const uint8_t **get_chunks(const void *ctx, const uint8_t *data,
+			  size_t data_size, size_t chunk_size)
+{
+	const uint8_t **chunks = tal_arr(ctx, const uint8_t *, 0);
+
+	for (size_t i = 0; i + chunk_size < data_size; i += chunk_size) {
+		uint8_t *chunk = tal_arr(chunks, uint8_t, chunk_size);
```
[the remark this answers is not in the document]
darosior: Nice, thanks

darosior:
Ok, will remove the move!

doc/FUZZING.md
darosior:
```suggestion
[clang](http://clang.llvm.org/). The more recent the compiler version the better.
```

doc/FUZZING.md
niftynei: I understand how it works, thanks for the explainer. My point is that it *shouldn't* work this way -- the total runs should be divided amongst the parallel processes. This is how -n=X works for any other test suite.

doc/FUZZING.md
niftynei: > I feel like anything i could add would be vague and not necessarily helpful 

This is incredibly incorrect. All of the small definitions you've now added to this PR as comments would have been immeasurably helpful for me as I reviewed the code. It's true that someone *could* follow the link at the top, but in general it is *always* a good idea to include basic definitions and a brief overview of how things work when adding documentation.

And then saying "for more info <link>".

doc/FUZZING.md
darosior: > I understand how it works, thanks for the explainer.

.. Ok i didn't get what you proposed. I think that does not make sense: you want to be able to run each targets `X` times to generate the most coverage possible. Having a constant total runs across all the targets seems really counter-intuitive: what if you add two new targets ? You need to recompute the ratio to be able to execute each target the wanted number of times ? Same if we add an `--exclude` argument [like bitcoind](https://github.com/bitcoin/bitcoin/blob/9ad7cd2887abf5f91495337a2458560de10b4f69/test/fuzz/test_runner.py#L34-L38) to exclude targets that are too slow to generate coverage, you need to recompute your global value again.
This is really just an exposure of the `-runs` argument of each target.

tests/fuzz/fuzz-close_tx.c
darosior: Yeah, the above comment is about what close_tx wants, but we get the funding amount out of the to_us and to_them amounts

tests/fuzz/fuzz-close_tx.c
```diff
@@ -0,0 +1,90 @@
...
+	 * - 1 txid: 32 bytes
+	 * - 1 u32: 4 bytes
+	 * - 4 amount_sat: 4 * 8 bytes
+	 *
+	 * Since the output scripts size is not very restricted (and the shutdown_scriptpukeys
+	 * are not by the spec either), we also try to vary their length.
+	 * Therefore, we allocate the entire remaining bytes to scripts.
+	 */
+	min_size = 8 * 3 + 4 + 32;
+	if (size < min_size + 2)
```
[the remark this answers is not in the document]
darosior: To at least have two scripts of length one

tests/fuzz/fuzz-close_tx.c
darosior: This should not fail, will add an assert on wire_ptr

tests/fuzz/fuzz-channel_id.c
darosior: `txid` + `vout` (u16 ala LN)

tests/fuzz/fuzz-bip32.c
darosior: We were already using it for the function's signature so..

tests/fuzz/fuzz-bip32.c
darosior: Why is it better ? (not bikeshedding, doing it but just to get the rationale)

common/amount.c
darosior: But we promise to return a tal-allocated value on `ctx` here ?

common/test/run-amount.c
darosior: Ah no, they were part of the right commit! Well because it was just added above, and as outlined in the commit message..
>  Allows `.*btc` amounts (without post-decimal)

tests/fuzz/libfuzz.h
darosior: Most fuzzing engines offer this API:
- init
- run
I'd rather not differ too much from this pattern

tests/fuzz/fuzz-bip32.c
niftynei: It's more explicit imo -- you're looking for the byte count here

tests/fuzz/fuzz-channel_id.c
niftynei: add this to the comment lol

common/amount.c
rustyrussell: yeah, caller might tal_free it...

tests/fuzz/fuzz-bech32.c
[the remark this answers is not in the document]
ZmnSCPxj: We usually use `tal` in C-Lightning, so a random `malloc` is something to note.  Is there a justification for using `malloc` instead of `tal`?  E.g. "we inadvertently end up testing the `tal` code instead of whatever this test is testing if we use `tal`, so use `malloc` instead."

tests/fuzz/run.py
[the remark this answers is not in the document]
darosior: I initially wanted to joke with "inb4 Zmn comes in and gives a good argument for writing it in C" x)

tests/fuzz/fuzz-bech32.c
darosior: Didn't find them

tests/fuzz/fuzz-bech32.c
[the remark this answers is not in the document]
darosior: I only used `tal` when required, here this interfaces with a header that has no concept of `tal` (even gives the size to allocate in headers)

tests/fuzz/fuzz-channel_id.c
darosior: Done

tests/fuzz/run.py
```diff
@@ -0,0 +1,78 @@
...
+    )
+
+    args = parser.parse_args()
+
+    logging.basicConfig(level=logging.DEBUG)
+
+    target_dir = os.path.abspath(os.path.dirname(__file__))
+    targets = [os.path.join(target_dir, f) for f in os.listdir(target_dir)
+               if re.compile(r"^fuzz-[\w-]*$").findall(f)]
+    with ProcessPoolExecutor(max_workers=args.par) as pool:
```
cdecker, line 57: I wonder if using `multiprocessing.Pool(args.par).map` or a `ThreadPoolExecutor.map` wouldn't be faster and less bookkeeping overall.

doc/FUZZING.md
cdecker: Actually `--help` works like a charm :-)

cdecker (commented):
Just some minor performance tips :-)

tests/fuzz/fuzz-amount.c
```diff
@@ -0,0 +1,62 @@
...
+	uint8_t *buf;
+	const char *fmt_msat, *fmt_msatbtc, *fmt_sat, *fmt_satbtc;
+
+
+	/* We should not crash when parsing any string. */
+
+	string = tal_arr(NULL, char, 0);
+	for (size_t i = 0; i < size; i++)
+		if (data[i] <= CHAR_MAX)
+			tal_arr_expand(&string, (char)data[i]);
```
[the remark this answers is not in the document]
darosior: Nice catch, thanks, applied.

```patch
commit e3fdb759be3867416cbfb62ef18b8a23f2c6b930 (HEAD -> fuzzing)
Author: Antoine Poinsot <darosior@protonmail.com>
Date:   Mon Oct 19 16:02:55 2020 +0200

    fixup! fuzz: add a fuzz target for amounts parsing
    
diff --git a/tests/fuzz/fuzz-amount.c b/tests/fuzz/fuzz-amount.c
index b82e5b970..644c9b0b7 100644
--- a/tests/fuzz/fuzz-amount.c
+++ b/tests/fuzz/fuzz-amount.c
@@ -21,10 +21,9 @@ void run(const uint8_t *data, size_t size)
 
        /* We should not crash when parsing any string. */
 
-       string = tal_arr(NULL, char, 0);
+       string = tal_arr(NULL, char, size);
        for (size_t i = 0; i < size; i++)
-               if (data[i] <= CHAR_MAX)
-                       tal_arr_expand(&string, (char)data[i]);
+               string[i] = (char) data[i] % (CHAR_MAX + 1);
 
        parse_amount_msat(&msat, string, tal_count(string));
        parse_amount_sat(&sat, string, tal_count(string));
```

I discarded pre-allocation for this because of my tests with `get_chunks`..

tests/fuzz/libfuzz.c
```diff
@@ -14,3 +18,18 @@ int LLVMFuzzerInitialize(int *argc, char ***argv) {
...
+
+const uint8_t **get_chunks(const void *ctx, const uint8_t *data,
+			  size_t data_size, size_t chunk_size)
+{
+	const uint8_t **chunks = tal_arr(ctx, const uint8_t *, 0);
+
+	for (size_t i = 0; i + chunk_size < data_size; i += chunk_size) {
+		uint8_t *chunk = tal_arr(chunks, uint8_t, 0);
+		for (size_t j = 0; j < chunk_size; j ++)
+			tal_arr_expand(&chunk, data[j]);
```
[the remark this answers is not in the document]
darosior: Done

> Alternatively we could also just make **chunks an array of pointers into data, avoiding copying at all

Callers want tal objects unfortunately :/

tests/fuzz/libfuzz.c
```diff
@@ -14,3 +18,18 @@ int LLVMFuzzerInitialize(int *argc, char ***argv) {
 
 	return 0;
 }
+
+const uint8_t **get_chunks(const void *ctx, const uint8_t *data,
+			  size_t data_size, size_t chunk_size)
+{
+	const uint8_t **chunks = tal_arr(ctx, const uint8_t *, 0);
```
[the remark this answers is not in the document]
darosior: Done heres the fixup
```patch
commit 4acf898faca5713e7b13c1529c9c837927eede9e (HEAD -> fuzzing)
Author: Antoine Poinsot <darosior@protonmail.com>
Date:   Mon Oct 19 16:59:42 2020 +0200

    fixup! fuzz: add a fuzz target for common/bigsize
    
diff --git a/tests/fuzz/libfuzz.c b/tests/fuzz/libfuzz.c
index 53eee8e0b..273de5523 100644
--- a/tests/fuzz/libfuzz.c
+++ b/tests/fuzz/libfuzz.c
@@ -22,14 +22,12 @@ int LLVMFuzzerInitialize(int *argc, char ***argv) {
 const uint8_t **get_chunks(const void *ctx, const uint8_t *data,
                          size_t data_size, size_t chunk_size)
 {
-       const uint8_t **chunks = tal_arr(ctx, const uint8_t *, 0);
-
-       for (size_t i = 0; i + chunk_size < data_size; i += chunk_size) {
-               uint8_t *chunk = tal_arr(chunks, uint8_t, 0);
-               for (size_t j = 0; j < chunk_size; j ++)
-                       tal_arr_expand(&chunk, data[j]);
-               tal_arr_expand(&chunks, chunk);
-       }
+       size_t n_chunks = data_size / chunk_size;
+       const uint8_t **chunks = tal_arr(ctx, const uint8_t *, n_chunks);
+
+       for (size_t i = 0; i < n_chunks; i++)
+               chunks[i] = tal_dup_arr(chunks, const uint8_t,
+                                       data + i * chunk_size, chunk_size, 0);
 
        return chunks;
 }
```

tests/fuzz/fuzz-initial_channel.c
```diff
@@ -0,0 +1,79 @@
...
+	status_setup_sync(STDERR_FILENO);
```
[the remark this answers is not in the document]
darosior: Yeah was useful for testing (no binary prefixes iirc) though it should not be pointing to stderr anymore :)

```patch
commit b2a28a199cfbbfb3fa1d896b90bb728812099451 (HEAD -> fuzzing)
Author: Antoine Poinsot <darosior@protonmail.com>
Date:   Mon Oct 19 17:14:10 2020 +0200

    fixup! fuzz: add a fuzz target for initial_channel
    
diff --git a/tests/fuzz/fuzz-initial_channel.c b/tests/fuzz/fuzz-initial_channel.c
index 441993610..8217b4007 100644
--- a/tests/fuzz/fuzz-initial_channel.c
+++ b/tests/fuzz/fuzz-initial_channel.c
@@ -1,7 +1,7 @@
 #include <assert.h>
+#include <fcntl.h>
 #include <stdint.h>
 #include <string.h>
-#include <unistd.h>
 #include <tests/fuzz/libfuzz.h>
 
 #include <bitcoin/pubkey.h>
@@ -21,7 +21,8 @@
 void init(int *argc, char ***argv)
 {
        common_setup("fuzzer");
-       status_setup_sync(STDERR_FILENO);
+       int devnull = open("/dev/null", O_WRONLY);
+       status_setup_sync(devnull);
        chainparams = chainparams_for_network("bitcoin");
 }
```

tests/fuzz/run.py
darosior: Will try (made sense to me though to have a **process** per target)

darosior:
Thanks for the reviews so far! Amended as per latest one.

(Will check about the process pool soon :tm:)

tests/fuzz/run.py
cdecker: Well, if we are using a `ProcessPoolExecutor` we are actually using two processes for each fuzzer:

 - The worker process that communicates with the main process, and spawns the subprocess
 - The subprocess which `execv`s the fuzz binary.

We don't really gain anything from spinning up the worker process.

darosior: Yeah yeah definitely

darosior: Made it a `ThreadPoolExecutor`

darosior:
What do you think about having a repo in lightningd/ to store corpora ?

cdecker:
> What do you think about having a repo in lightningd/ to store corpora ?

Sounds like something we could consider. From what I've seen so far the corpus is not going to be huge (<= 1GB by my estimate) so that can work. I wonder however if the version control brings anything new to the table, given that the corpus consists mostly of binary files, so maybe a linear storage model can work as well? The big advantage of using git + github is the reuse of infra and processes to collaborate on building up a corpus.

darosior:
Yeah, really just thinking of copying the workflow at github.com/bitcoin-core/qa-assets :-)

jsarenik:
@darosior that qa-assets repo looks ugly to me. Exactly because of what @cdecker wrote earlier.

Here is my offer: If you agree to put it on IPFS, i will happily help with any and all automation required (which will be properly kept is a git repository; e.g. building the corpuses, adding them to IPFS, changing `dnslink` TXT record, etc.). Seems to me like a better way to cope with binary blobs.

Yes, i am freshly IPFSed and strongly biased against any binary blob in repositories designed for source code*. See my goHugo generated site running via IPFS at https://pokus.bublina.eu.org. And [more can be done](http://docs.ipfs.io.ipns.localhost:8080/how-to/host-git-style-repo/) if needed.

* Having personally obliterated terabytes (!) of binary blobs in CVS-like repositories of an unnamed corporation (greetings).

jsarenik:
Just for reference here is one repository that uses binary fuzz files, even inside its main repository: https://github.com/rust-bitcoin/rust-bitcoin/tree/master/fuzz. The proposed separate repository (like they have in https://github.com/bitcoin-core/qa-assets) is a better idea compared to in-tree blobs.

cdecker:
> @darosior that qa-assets repo looks ugly to me. Exactly because of what @cdecker wrote earlier.
> 
> Here is my offer: If you agree to put the binary data out of git, i will happily help with any and all automation required (which will be itself properly kept in a git repository; e.g. building the corpuses, uploading them somewhere (or sharing them somehow), etc.).
> 
> Yes, i learned a bit about Torrent in last weeks and am strongly biased against any binary blob in repositories designed for source code.
> 
> https://git-annex.branchable.com/ or https://git-lfs.github.com/ can be used to track the binary blobs.
> 
> EDIT: This offer is valid also for serving the binary data over [BitTorrent](https://en.wikipedia.org/wiki/BitTorrent). No cloud needed.

I think the main issue is collecting and managing the corpus, the serving is secondary (happy to host the corpus on GCS for a couple of cents each month). For the code review and addition we have a really good process, that's no longer true if we start managing things outside the git tree.

I absolutely share your dislike of binary artifacts in git, however it might be easiest to get started with an in-tree solution, and move that out at a later point in time. Or the separate corpus repo could be a cleaner solution that doesn't taint the main repo with these binaries.

Going a bit back and forth between the solutions tbh.

darosior:
Yeah, management on a server would be awful, discussions around it would be mixed up. I can already imagine having to upload my new corpus to (a VPS / a torrent seeder / whatever) and bug cdecker on IRC for him to pull it and eventually merge it with the existing one on its GCS for it to be pulled by our CI... Wow. Let's just use Github?

cdecker:
Ok, let's use a corpus repo then on the `lightningd` org. That unblocks this PR which I'd like to merge, and keeps our main tree clean, while allowing us to come up with a better solution at a later point in time.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4129/commits/acef7ea45c7e9934eef7483126f5776ec89fbe12

jsarenik:
ACK acef7ea (have learned a lot about Fuzzing in the process, thank you!)

darosior:
@rustyrussell could you look at the amount parsing commit ? iirc i stumbled upon something like that in the past and you corrected my fix so just would like your input on this one.

jsarenik:
This PR has been merged. Is your comment relevant or is it here by mistake? Any link to that commit you mention?

PR #4131, opened 2020-10-15
openingd: patch test_opening_tiny_channel under EXP_FEAT

niftynei:
`test_opening_tiny_channel` fails if EXPERIMENTAL_FEATURES is on because
we don't include the anchor in our reserve if we're the channel opener.

Seems fine to include in all cases?


rustyrussell (approved):
Ack ec3aaf4cb57ce0a4a3486f8252725a84bbb1007d

PR #4134, opened 2020-10-16
cleanups for some dual-funding related code, II of III

niftynei:
This one's mostly small fixes found while running some tests. Plus we completely rework `check_balances`, the method that needed some help anyway.

Based on #4133, start at 3262c3a

PR #4135, opened 2020-10-16
cleanups for some dual-funding related code, III of III

niftynei:
This one's actually a big logical change. Prior to this (in #4069), we would wait until the user provided the signatures for `openchannel_signed` before we opened up a channeld and sent them to our peer. 

The problem with this is that we wouldn't receive the peer's sigs until after we've sent our sigs.  To make this work with `multifundchannel` (coming soon, watch this spaceTM), we need to be able to collect all of our peer's sigs first, collate them for all the other peer's we're signing with, and then present the collected sigs (ours plus all the other nodes) to the peers.

To do this, we go ahead and jump directly into channeld after we've collected the `commitment_sigs` from our peer. We can now call `openchannel_signed` at our leisure -- ideally as soon as you've got all the signatures you need!

Note that this impacts our reconnect logic in channeld; we'll send the tx_sigs *and* funding_locked on reconnect until we've gotten a funding_locked from the peer.

Right, so to summarize

- `openchannel_signed` is no longer a blocker for starting channeld
- There's some logic in `channeld` now around whether we've sent sigs to our peer yet or not and whether we've gotten them back from the peer (or not)
- If the peer sends us `funding_locked` but we haven't gotten `tx_sigs` from them yet, we'll close the channel
- There's now a notification when we receive sigs from the peer

Built on top of #4134, this commit set starts at 22bf555

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -512,6 +512,20 @@ static void announce_channel(struct peer *peer)
 	send_channel_update(peer, 0);
 }
 
+#if EXPERIMENTAL_FEATURES
+static enum tx_role our_tx_role(const struct peer *peer)
+{
+	return peer->channel->opener == LOCAL ?
+		TX_INITIATOR : TX_ACCEPTER;
+}
+#endif
+
+static enum tx_role their_tx_role(const struct peer *peer)
+{
+	return peer->channel->opener == LOCAL ?
+		TX_ACCEPTER : TX_INITIATOR;
+}
+
 static void channel_announcement_negotiate(struct peer *peer)
 {
 	/* Don't do any announcement work if we're shutting down */
@@ -592,6 +606,13 @@ static void handle_peer_funding_locked(struct peer *peer, const u8 *msg)
 	if (peer->shutdown_sent[LOCAL])
 		return;
 
+	if (peer->psbt && !psbt_side_finalized(peer->psbt,
+					       their_tx_role(peer)))
+		peer_failed(peer->pps,
+			    &peer->channel_id,
+			    "Rcvd `funding_locked` from peer but "
+			    "have not received `tx_signatures`");
+
 	peer->old_remote_per_commit = peer->remote_per_commit;
 	if (!fromwire_funding_locked(msg, &chanid,
 				     &peer->remote_per_commit))
@@ -608,6 +629,7 @@ static void handle_peer_funding_locked(struct peer *peer, const u8 *msg)
 					   &peer->channel_id));
 
 	peer->funding_locked[REMOTE] = true;
+	peer->psbt = tal_free(peer->psbt);
 	wire_sync_write(MASTER_FD,
 			take(towire_channeld_got_funding_locked(NULL,
 						&peer->remote_per_commit)));
@@ -2037,8 +2059,7 @@ static void handle_tx_sigs(struct peer *peer, const u8 *msg)
 	const struct witness_stack **ws;
 
 	size_t j = 0;
-	enum tx_role role = peer->channel->opener == REMOTE
-		? TX_INITIATOR : TX_ACCEPTER;
+	enum tx_role their_role = their_tx_role(peer);
 
 	if (!fromwire_tx_signatures(tmpctx, msg, &cid, &txid,
 				    cast_const3(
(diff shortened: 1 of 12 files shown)


channeld/channeld.c
```diff
@@ -2003,6 +2003,9 @@ static void handle_tx_sigs(struct peer *peer, const u8 *msg)
 	struct witness_stack **ws;
 	const struct wally_tx *wtx;
 	size_t j = 0;
+	enum tx_role role = peer->channel->opener == REMOTE
+		? TX_INITIATOR : TX_ACCEPTER;
```
rustyrussell: This is *their role*, right?  If they opened the channel, they're the initiator...

channeld/channeld.c
```diff
@@ -2738,11 +2737,13 @@ static void peer_reconnect(struct peer *peer,
 
 #if EXPERIMENTAL_FEATURES
 	/* Send our tx_sigs again */
-	if (peer->psbt && !peer->funding_locked[REMOTE]) {
+	enum tx_role role = peer->channel->opener == LOCAL
+		? TX_INITIATOR : TX_ACCEPTER;
```
rustyrussell: Hmm, maybe make an inline function `enum tx_role our_role(enum side opener)`?

channeld/channeld.c
```diff
@@ -1996,12 +1996,47 @@ static void send_onionmsg(struct peer *peer, const u8 *msg)
...
 }
 
+static void handle_send_tx_sigs(struct peer *peer, const u8 *msg)
+{
+	const struct wally_psbt *psbt;
+	struct bitcoin_txid txid;
+
+	if (!fromwire_channeld_send_tx_sigs(tmpctx, msg,
+					    cast_const2(struct wally_psbt **,
+						       &psbt)))
```
rustyrussell, line 2027: just de-const the declaration?

lightningd/channel_control.c
```diff
@@ -386,21 +388,18 @@ static void sendfunding_done(struct bitcoind *bitcoind UNUSED,
...
-	if (!fromwire_channeld_funding_tx(tmpctx, msg, &cs->wtx)) {
-		channel_internal_error(channel,
-				       "bad channeld_funding_tx: %s",
-				       tal_hex(tmpctx, msg));
-		return;
-	}
+	if (taken(wtx))
+		cs->wtx = tal_steal(cs, wtx);
+	else
+		cs->wtx = wtx;
```
rustyrussell, line 403: If we're not supposed to take it, we should almost certainly clone it?

lightningd/channel_control.c
```diff
@@ -412,6 +411,43 @@ static void handle_funding_tx(struct channel *channel,
...
+	if (!fromwire_channeld_funding_sigs(tmpctx, msg, &psbt)) {
+		channel_internal_error(channel,
+				       "bad channeld_funding_sigs: %s",
+				       tal_hex(tmpctx, msg));
+		return;
+	}
+
+	tal_wally_start();
+	if (wally_psbt_combine(cast_const(struct wally_psbt *,
+					  channel->psbt),
```
rustyrussell, line 430: This breaks the rules; channel->psbt must not be const if we want to alter it.  Casting away just removes the warning, but it's still no legal to alter it.

channeld/channeld.c
```diff
@@ -512,6 +512,18 @@ static void announce_channel(struct peer *peer)
 	send_channel_update(peer, 0);
 }
 
+static enum tx_role our_tx_role(struct peer *peer)
```
rustyrussell, line 515: const...

channeld/channeld.c
```diff
@@ -512,6 +512,18 @@ static void announce_channel(struct peer *peer)
...
+{
+	return peer->channel->opener == LOCAL ?
+		TX_INITIATOR : TX_ACCEPTER;
+}
+
+static enum tx_role their_tx_role(struct peer *peer)
```
rustyrussell, line 521: const...

channeld/channeld.c
```diff
@@ -608,6 +627,8 @@ static void handle_peer_funding_locked(struct peer *peer, const u8 *msg)
 					   &peer->channel_id));
 
 	peer->funding_locked[REMOTE] = true;
+	if (peer->psbt)
+		peer->psbt = tal_free(peer->psbt);
```
rustyrussell, line 631: Don't need to `if (peer->psbt)` here.  tal_free(NULL) is a noop.

rustyrussell (approved):
Ack 362123a8714c8d486ea433d1556a72224d35ed0b

Minor cleanups only!

channeld/channeld.c
niftynei: i hate const.

lightningd/channel_control.c
niftynei: Good point!

lightningd/channel_control.c
niftynei: why is const. so unhappy. all the time.

niftynei (commented):
Updated the things; added a small commit https://github.com/ElementsProject/lightning/pull/4135/commits/eeec16c34bbee9faaaf0ba31225b031f9b3c676f to remove the const-ness from `channel.psbt`, since we're now updating it in the next patch.

PR #4138, opened 2020-10-17
chore: adds env NO_PYTHON check to Makefiles

m-schmoock:
This adds an environment variable $NO_PYTHON check to Makefiles so that:
 - It checks and runs into an defined error instead of some python hickup:
   `ModuleNotFoundError: No module named 'mako'`
 - makes it possible to manually export this environment variable NO_PYTHON=1
   to run the same testcase that travis is running on job 1 which causes
   so much pain ;)

Note: also we can use this to sure by bisecting each commit of a branch the correct SQL generations are being committed.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4138/commits/c795b54e61804f566f4bc2cb00c8b0979d16705c

PR #4139, opened 2020-10-19
Add a `commitment feerate` parameter to multifundchannel

niftynei:
This lets a caller of `multifundchannel` specify a separate feerate for the channel's initial commitment transaction. This separates it from the feerate used for the funding transaction.

Writing the test for this uncovered an off by one in how we were calculating signature lengths.

Note that this conforms with the interface for `openchannel_init`, the new RPC that will initiate a v2 channel open (which contains two separate feerates, one for the funding tx and another for the commitment txs)


bitcoin/tx.c
```diff
@@ -806,10 +806,10 @@ size_t bitcoin_tx_output_weight(size_t outscript_len)
 	return weight;
 }
 
-/* We grind signatures to get them down to 71 bytes (+1 for sighash flags) */
+/* We grind signatures to get them down to 71 bytes */
 size_t bitcoin_tx_input_sig_weight(void)
 {
-	return 1 + 71 + 1;
+	return 1 + 71;
```
darosior, line 812: I think you overlooked the length prefix.

length prefix + sig + sighash_flag

darosior (changes requested):

bitcoin/tx.c
niftynei: sighash_flag is included in the 71 count, see commit message

darosior: Oh, right!

rustyrussell (approved):
Ack a45b7a2da85d62b7fba7b66ab30bb619dcb14855

PR #4142, opened 2020-10-20
pyln: Adds some improvements to pyln.proto.zbase32

sr-gi:
Before this PR, `zbase32.encode` could only be used for messages with length multiple of 5. This is because zbase32 encoding uses blocks of 5-bit words for the encoding. However, arbitrary length messages should be encodable as well by properly padding them to match the expected length.

The additions to this PR could be summarised as follows:

- Add helper functions to encode / decode arbitrary length data by padding / cutting the provided data (`_message_to_bitarray` and `_bitarray_to_message`).
- Mark any function that is not supposed to be used publicly as "private".
- Adds input checking to the public functions (`encode`, `decode`, `is_zbased32_encoded`) and raise exceptions accordingly.
- Add short descriptions to each function.
- Add tests for the whole module.

Changelog-None

### DIFF
--- a/contrib/pyln-proto/pyln/proto/zbase32.py
+++ b/contrib/pyln-proto/pyln/proto/zbase32.py
@@ -24,6 +24,7 @@ zbase32_revchars = [
 
 
 def _message_to_bitarray(message):
+    """Encodes a message as a bitarray with length multiple of 5."""
     barr = bitstring.ConstBitStream(message)
     padding_len = 5 - (len(barr) % 5)
     if padding_len < 5:
@@ -33,6 +34,7 @@ def _message_to_bitarray(message):
 
 
 def _bitarray_to_message(barr):
+    """Decodes a bitarray with length multiple of 5 to a byte message (removing the padded zeros if found)."""
     padding_len = len(barr) % 8
     if padding_len > 0:
         return bitstring.Bits(bin=barr.bin[:-padding_len]).bytes
@@ -41,6 +43,7 @@ def _bitarray_to_message(barr):
 
 
 def _bitarray_to_u5(barr):
+    """Converts a bitarray in a list of uint5."""
     ret = []
     while barr.pos != barr.len:
         ret.append(barr.read(5).uint)
@@ -48,6 +51,7 @@ def _bitarray_to_u5(barr):
 
 
 def _u5_to_bitarray(arr):
+    """Converts a list of uint5 values to a bitarray."""
     ret = bitstring.BitArray()
     for a in arr:
         ret += bitstring.pack("uint:5", a)
@@ -55,6 +59,7 @@ def _u5_to_bitarray(arr):
 
 
 def is_zbase32_encoded(message):
+    """Checks if a message is zbase32 encoded."""
     if isinstance(message, str):
         message = message.encode("ASCII")
     elif not isinstance(message, bytes):
@@ -63,6 +68,7 @@ def is_zbase32_encoded(message):
 
 
 def encode(message):
+    """Encodes a message (str or bytes) to zbase32."""
     if isinstance(message, str):
         message = message.encode('ASCII')
     elif not isinstance(message, bytes):
@@ -75,6 +81,7 @@ def encode(message):
 
 
 def decode(message):
+    """Decodes a message (str or bytes) from zbase32."""
     if isinstance(message, str):
         message = message.encode('ASCII')
     elif not isinstance(message, bytes):
--- a/contrib/pyln-proto/pyln/proto/zbase32.py
+++ b/contrib/pyln-proto/pyln/proto/zbase32.py
@@ -23,34 +23,68 @@ zbase32_revchars = [
 ]
 
 
-def bitarray_to_u5(barr):
-    assert len(barr) % 5 == 0
+def _message_to_bitarray(message):
+    barr = bitstring.ConstBitStream(message)
+    padding_len = 5 - (len(barr) % 5)
+    if padding_len < 5:
+        # The bitarray length has to be multiple of 5. If not, it is right-padded with zeros.
+        barr = bitstring.ConstBitStream(bin="{}{}".format(barr.bin, '0' * padding_len))
+    return barr
+
+
+def _bitarray_to_message(barr):
+    padding_len = len(barr) % 8
+    if padding_len > 0:
+        return bitstring.Bits(bin=barr.bin[:-padding_len]).bytes
+    else:
+        return barr.bytes
+
+
+def _bitarray_to_u5(barr):
     ret = []
-    s = bitstring.ConstBitStream(barr)
-    while s.pos != s.len:
-        ret.append(s.read(5).uint)
+    while barr.pos != barr.len:
+        ret.append(barr.read(5).uint)
     return ret
 
 
-def u5_to_bitarray(arr):
+def _u5_to_bitarray(arr):
     ret = bitstring.BitArray()
     for a in arr:
         ret += bitstring.pack("uint:5", a)
     return ret
 
 
-def encode(b):
-    uint5s = bitarray_to_u5(b)
+def is_zbase32_encoded(message):
+    if isinstance(message, str):
+        message = message.encode("ASCII")
+    elif not isinstance(message, bytes):
+        raise TypeError("message must be string or bytes")
+    return set(message).issubset(zbase32_chars)
+
+
+def encode(message):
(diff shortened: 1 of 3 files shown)


contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -23,34 +23,69 @@
...
 
-def bitarray_to_u5(barr):
-    assert len(barr) % 5 == 0
+def _message_to_bitarray(message):
+    barr = bitstring.ConstBitStream(message)
+    filling_len = len(barr) % 5
+    if filling_len:
+        # The bitarray length has to be multiple of 5. If not, it is right-filled with zeros.
+        filling = '0'*(5-filling_len)
+        barr = bitstring.ConstBitStream(bin=f"{barr.bin}{filling}")
```
cdecker, line 33: Not sure we have f-string support in the lightning codebase yet. We only specified a minimum version of python in the `lightningd/plugins` repository. That being said, I'd be happy to specify one explicitly after checking with everybody.

contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -23,34 +23,69 @@
...
+    return barr
+
+
+def _bitarray_to_message(barr):
```
cdecker, line 37: Type annotations here would be useful.

contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -34,6 +35,7 @@ def _message_to_bitarray(message):
 
 
 def _bitarray_to_message(barr):
+    """Decodes a bitarray with length multiple of 5 to a byte message (removing the trailing zeros if necessary)."""
```
cdecker, line 38: Should the trimming assert that the trimmed part is all 0s?

contrib/pyln-proto/tests/test_zbase32.py
```diff
@@ -0,0 +1,106 @@
...
+def test_is_zbase32_encoded():
+    for message in zbase_encoded_messages:
+        assert zbase32.is_zbase32_encoded(message)
+
+    for message in not_zbase_encoded_messages:
+        assert not zbase32.is_zbase32_encoded(message)
+
+
+def test_encode():
+    for i, message in enumerate(messages):
```
cdecker, line 80: You can iterate over both arrays `messages` and `zbase_encoded_messages` using `zip(messages, zbase_encoded_messages)`. This way you can just loop like this:

```
for mess, expected in zip(messages, zbase_encoded_messages):
     got = zbase32.encode(message)
     assert(got == expected)
```

cdecker: Alternatively you can also make the messages a `dict`, mapping from the raw message to the zbase32-encoded messages and then iterate over `messages.items()`.

cdecker (commented):
Just some minor points and some linting issues:

```
contrib/pyln-proto/pyln/proto/zbase32.py:32:22: E226 missing whitespace around arithmetic operator
contrib/pyln-proto/pyln/proto/zbase32.py:32:25: E226 missing whitespace around arithmetic operator
contrib/pyln-proto/tests/test_primitives.py:1:1: F401 'binascii.hexlify' imported but unused
contrib/pyln-proto/tests/test_primitives.py:2:1: F401 'pyln.proto.zbase32' imported but unused
Makefile:405: recipe for target 'check-python' failed
make: *** [check-python] Error 1
```

contrib/pyln-proto/pyln/proto/zbase32.py
sr-gi: I can change it for `.format` otherwise.

contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -23,34 +23,69 @@
...
+    if filling_len:
+        # The bitarray length has to be multiple of 5. If not, it is right-filled with zeros.
+        filling = '0'*(5-filling_len)
+        barr = bitstring.ConstBitStream(bin=f"{barr.bin}{filling}")
+    return barr
+
+
+def _bitarray_to_message(barr):
+    filling_len = len(barr) % 8
+    if filling_len:
```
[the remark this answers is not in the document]
sr-gi: Sorry, I'm too used to Python falsy checks :S. I'll change it.

contrib/pyln-proto/pyln/proto/zbase32.py
sr-gi: I was assuming the input is zbase32 encoded, in which case it must be. Given the function is supposed to be internal I didn't care about checking the input for correctness, but I can add the check and except if you think it's worth having it.

contrib/pyln-proto/tests/test_zbase32.py
sr-gi: Yeah, indeed. That's way nicer.

sr-gi:
Added some fixes: 

- Fixes linting errors
- Added type annotations
- Replaced falsy checks
- Improved code style for tests

Pending:

- f-strings
- input checking for internal functions

contrib/pyln-proto/pyln/proto/zbase32.py
cdecker: Maks sense.

cdecker:
Excellent, thanks for the fixes. There's one more failing python lint:

```
contrib/pyln-proto/pyln/proto/zbase32.py:33:27: E226 missing whitespace around arithmetic operator
```

`make check-python` will run these locally (though in the past I had differing results depending on the version of `flake8`).

sr-gi:
> Excellent, thanks for the fixes. There's one more failing python lint:
> 
> ```
> contrib/pyln-proto/pyln/proto/zbase32.py:33:27: E226 missing whitespace around arithmetic operator
> ```
> 
> `make check-python` will run these locally (though in the past I had differing results depending on the version of `flake8`).

Yeah, looks like we have different versions. Should be fixed now.

jsarenik:
The key word which should be used somewhere in the code (so that someone can `git grep` years later) is **padding**. I.e. "filling" the provided data is called "padding" them.

jsarenik:
So generally what i propose is `%s/filling/padding/g` (in vim syntax). Of course i may be wrong and my earlier comment may just be a false-positive caused by quick-reading the issue for the first time yesterday.

sr-gi:
> So generally what i propose is `%s/filling/padding/g` (in vim syntax). Of course i may be wrong and my earlier comment may just be a false-positive caused by quick-reading the issue for the first time yesterday.

Yeah, I agree, padding is the proper term. I'll rebase to change it.

contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -23,34 +24,76 @@
 ]
 
 
-def bitarray_to_u5(barr):
-    assert len(barr) % 5 == 0
+def _message_to_bitarray(message: bytes) -> bitstring.ConstBitStream:
+    """Encodes a message as a bitarray with length multiple of 5."""
+    barr = bitstring.ConstBitStream(message)
+    padding_len = len(barr) % 5
```
cdecker, line 30: Technically padding is `5 - (len(barr) % 5)`, but that's splitting hairs here :-)

contrib/pyln-proto/pyln/proto/zbase32.py
```diff
@@ -23,34 +24,76 @@
...
+    if padding_len > 0:
+        # The bitarray length has to be multiple of 5. If not, it is right-padded with zeros.
+        padding = '0' * (5 - padding_len)
+        barr = bitstring.ConstBitStream(bin=f"{barr.bin}{padding}")
```
cdecker, line 34: Just one missing f-string.

cdecker (commented):
One f-string and a tiny nit. Looking good, and can merge if the f-string is addressed :-)

contrib/pyln-proto/pyln/proto/zbase32.py
sr-gi: Actually yeah, it used to be called `rem` for remainder by after all the name changing now it's rather not representing its meaning anymore. Will update to fit the description.

sr-gi:
Should be good now.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4142/commits/883527ae3de1f2ed6193eda12961126062842760

PR #4145, opened 2020-10-21
Make check fixups

rustyrussell:
We weren't doing include checks over all files, but fixing it revealed that there's no neat way of separating out gen_ files so I had to rename them...


niftynei (approved):
ACK a27032f8e49eeb6803dc409f346d4ea785b39e36

PR #4146, opened 2020-10-22
dual_open: correct silent enum conversions

darosior:
We were silently converting a side enum (3 variants) to a tx_role enum
(2 variants).

Changelog-None

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -273,15 +273,16 @@ static enum tx_role our_tx_role(const struct peer *peer)
 }
 
 static const u8 *psbt_to_tx_sigs_msg(const tal_t *ctx,
-				     struct channel *channel,
+				     const struct peer *peer,
 				     const struct wally_psbt *psbt)
 {
-	const struct witness_stack **ws =
-		psbt_to_witness_stacks(tmpctx, psbt,
-				       channel->opener);
+	const struct witness_stack **ws;
 
-	return towire_tx_signatures(ctx, &channel->cid,
-				   &channel->funding_txid,
+	ws = psbt_to_witness_stacks(tmpctx, psbt,
+				    our_tx_role(peer));
+
+	return towire_tx_signatures(ctx, &peer->channel->cid,
+				   &peer->channel->funding_txid,
 				   ws);
 }
 #endif /* EXPERIMENTAL_FEATURES */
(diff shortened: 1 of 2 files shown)


channeld/channeld.c
```diff
@@ -264,9 +264,20 @@ static const u8 *psbt_to_tx_sigs_msg(const tal_t *ctx,
...
+	enum tx_role role;
+
+	switch (channel->opener) {
+	case LOCAL:
+		role = TX_INITIATOR;
+		break;
+	case REMOTE:
+		role = TX_ACCEPTER;
+		break;
+	default:
```
niftynei, line 277: FYI, adding a `default` totally undoes the reason for using a `switch` statement here, which is to fail to compile if/when the number of enums changes.

channeld/channeld.c
```diff
@@ -264,9 +264,20 @@ static const u8 *psbt_to_tx_sigs_msg(const tal_t *ctx,
 				     struct channel *channel,
 				     const struct wally_psbt *psbt)
 {
-	const struct witness_stack **ws =
-		psbt_to_witness_stacks(tmpctx, psbt,
-				       channel->opener);
```
niftynei, line 269: Since this will always be our sigs, we can use `our_tx_role`. (takes a peer, you can just pass the peer in to this instead of channel)

niftynei (commented):
Nice. Just needs a small change.

channeld/channeld.c
darosior: There is already a third alternative (`NUM_SIDE`), the purpose of this switch was to assert that we are never passed `NUM_SIDE` (or, as unlikely as it could be, a new alternative value) since it does not make sense for the 2-alternatives enum.

channeld/channeld.c
darosior: Ok, will do tomorrow :)

channeld/channeld.c
rustyrussell: Yeah, NUM_SIDES should be a #define, not within the enum.  This is the pattern we used elsewhere.

niftynei: Just checked this. `NUM_SIDES` is a #define, and has been since https://github.com/ElementsProject/lightning/commit/2618ef10c38b6437cb3b37094f0fb4c0cb569fb1

niftynei (commented):
Updated with suggested changes.

cdecker:
ACK 51e7dcfe4c5f8fc326e20df710fc9fd7f44a6d90

darosior:
ACK 1c3cb34

Thanks for taking care of this. Had to rebase on master to build properly.

darosior:
Could someone restart https://travis-ci.org/github/ElementsProject/lightning/jobs/740598643 ? The failure seems (very) unrelated and I can no longer do it

PR #4149, opened 2020-10-22
Add second format for 'dev-force-features'

niftynei:
Allows `dev-force-features` to also accept either '+231' or '-19' as a way to flag on or off a single feature.

This is probably the most likely usecase.

Changelog-None

### DIFF
--- a/common/features.c
+++ b/common/features.c
@@ -121,6 +121,16 @@ static const struct dependency feature_deps[] = {
 #endif
 };
 
+static void clear_feature_bit(u8 *features, u32 bit)
+{
+	size_t bytenum = bit / 8, bitnum = bit % 8, len = tal_count(features);
+
+	if (bytenum >= len)
+		return;
+
+	features[len - 1 - bytenum] &= ~(1 << bitnum);
+}
+
 static enum feature_copy_style feature_copy_style(u32 f, enum feature_place p)
 {
 	for (size_t i = 0; i < ARRAY_SIZE(feature_styles); i++) {
@@ -202,16 +212,6 @@ static bool test_bit(const u8 *features, size_t byte, unsigned int bit)
 {
 	assert(byte < tal_count(features));
 	return features[tal_count(features) - 1 - byte] & (1 << (bit % 8));
-}
-
-static void clear_feature_bit(u8 *features, u32 bit)
-{
-	size_t bytenum = bit / 8, bitnum = bit % 8, len = tal_count(features);
-
-	if (bytenum >= len)
-		return;
-
-	features[len - 1 - bytenum] &= ~(1 << bitnum);
 }
 
 /* BOLT #7:
--- a/common/features.c
+++ b/common/features.c
@@ -121,6 +121,21 @@ static const struct dependency feature_deps[] = {
 #endif
 };
 
+static void trim_features(u8 **features)
+{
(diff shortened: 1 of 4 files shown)


rustyrussell (approved):
Minor cleanup needed

niftynei:
@rustyrussell did you leave a comment on this somewhere? I can't find the clean up you're referring to.

common/features.c
```diff
@@ -121,14 +121,26 @@ static const struct dependency feature_deps[] = {
...
+	/* What is the new feature array length? */
+	new_len = len;
+	while (new_len > 0 && (*features)[len - new_len] == 0)
+		new_len--;
+
+	u8 *newarr = tal_arr(tal_parent(*features), u8, new_len);
+	if (new_len != 0)
+		memcpy(newarr, *features + (len - new_len), new_len);
+	tal_free(*features);
+	*features = newarr;
```
rustyrussell, line 143: This definitely confused me, because it looks like an open-coded tal_resize().

But ofc the features are big-endian, so we're stripping off the *front*:
```
/* Big-endian bitfields are weird, but it means we trim from the front: */
for (size_t trim = 0; trim < len; trim++) {
   if (features[trim])
       break;
}
memmove(*features, *features + trim, len - trim);
tal_resize(features, len - trim);
```

/* Trim features
memmove(*features, *features + (len - new_len), new_len)

rustyrussell (changes requested):
Sorry, obv my comment got lost somewhere.  It was wrong anyway, since I thought you simply open-coded realloc.  So really needs a comment...

common/features.c
cdecker: Would it make sense to extract this into a separate function `features_strip()` that minimizes the featurebit array as per spec? We're calling `clear_feature_bit` in a loop in the next commit potentially resizing the array on each iteration, which potentially results in intermediate allocations.

lightningd/options.c
```diff
@@ -516,8 +516,29 @@ static char *opt_force_featureset(const char *optarg,
...
+			return "Invalid feature number";
+
+		f = feature_set_for_feature(NULL, n);
+		if (strstarts(optarg, "-")
+		    && !feature_set_sub(ld->our_features, take(f)))
+			return "Feature unknown";
+
+		if (strstarts(optarg, "+")
+		    && !feature_set_or(ld->our_features, take(f)))
+			return "Feature already flagged-on";
```
cdecker, line 538: This could lead to accidentally breaking configurations on upgrade. Should probably be a warning log message, and not a direct failure in the options parsing.

niftynei: Yes, the break is intentional. It's specifically aimed at breaking tests.  Since this is a dev- option, it's probably fine to notify anyone using it (via a start-failure) that their option defaults have changed.

common/features.c
niftynei: I like this idea, of having it as a separate function. I do worry that it creates a 'non-obvious' dependency, in that every 'clear-feature-bit' *must* call `strip`. Guess it's not really a problem since it's all internal/private to this class, though...

niftynei:
Updated! Broke out the `trim_features` into a new method and updated the trim logic to use `memove` + `tal_resize`.

rustyrussell (approved):
Ack bdad89b622a2f6f6347d0221f00bd776b8200f85

PR #4150, opened 2020-10-22
Prelim patches for dual-funding/multifundchannel

niftynei:
Few small things before we get into "dual-funding over multifundchannel" fun.

- 'finalize' the redeemscript (if present) at the same time that we set final witness-scripts. libwally `finalize` ignores unfinalized redeemscripts if `final_witness_stack` field is set.
- correctly handle dev-memleak calls that are made while we're doing tx output/input updates.

Changelog-None


rustyrussell (approved):
Ack ded2dbe8f9feceb798d6f26cf778730b48f7eed1

PR #4153, opened 2020-10-23
chore: optimize postgres binary lookup

m-schmoock:
On my Archlinux system the current way of looking up the PostgreSQL binaries does not work.
I replaced the current hardwired way of looking up binaries by using the `pg_config --bindir` tool that should
work better on most distributions.

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyln/testing/db.py
+++ b/contrib/pyln-testing/pyln/testing/db.py
@@ -1,12 +1,10 @@
 from ephemeral_port_reserve import reserve  # type: ignore
-from glob import glob
 
 import itertools
 import logging
 import os
 import psycopg2  # type: ignore
 import random
-import re
 import shutil
 import signal
 import sqlite3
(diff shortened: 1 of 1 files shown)


jsarenik:
```
$ pg_config | grep BIN
BINDIR = /usr/lib/postgresql/12/bin
```

jsarenik:
No idea how to do `pg_config` magic in Python, but try it for yourself. I guess there should be no path guessing.

Anyway it's a good catch. But fixing hardwired paths by adding more hardwired paths does not seem right.

m-schmoock:
> Anyway it's a good catch. But fixing hardwired paths by adding more hardwired paths does not seem right.

We could add the hardwired path of `pg_config` for that :D

jsarenik:
No hardwired path here. If `pg_config` is not in `$PATH` then it is not around. Yes, it is possible to have many versions of Postgres installed and then their particular `pg_config`s would be in their particular locations. Let's fix the world just one small commit at a time.

jsarenik:
And the correct usage (without my previous `grep`) is:

```
$ pg_config --bindir
/usr/lib/postgresql/12/bin
```

m-schmoock:
@jsarenik thx, will update this PR tomorrow using a proper `shutil.which('pg_config')` lookup of `pg_config` to get the `--binpath` to look for the postres/initdb binaries.

jsarenik:
Rather the `PG_CONFIG` environment variable should be honored as it is used
already by `lightning/configure` script. (do `git grep pg_config`)

m-schmoock:
@jsarenik 
> Rather the `PG_CONFIG` environment variable should be honored as it is used already by `lightning/configure` script. (do `git grep pg_config`)

I changed it now to using the `pg_config --bindir` util directly rather than looking for `$PG_CONFIG` in environment, which may only be set for compile time. Another reason is that, if this particular code is called, the decision to use PostgreSQL has been made somewhere else in the code/config already (e.g. by setting `TEST_DB_PROVIDER=postgres`)  so we must make sure we can deliver a working setup.

jsarenik:
> @jsarenik
> 
> > Rather the `PG_CONFIG` environment variable should be honored as it is used already by `lightning/configure` script. (do `git grep pg_config`)
> 
> I changed it now to using the `pg_config --bindir` util directly rather than looking for `$PG_CONFIG` in environment, which may only be set for compile time. Another reason is that, if this particular code is called, the decision to use PostgreSQL has been made somewhere else in the code already (e.g. by config or by setting `TEST_DB_PROVIDER=postgres`) so we must make sure we can deliver a working setup at this point and not insist on a `$PG_CONFIG` variable and abort.

Fair enough. OK for me. Thanks.

contrib/pyln-testing/pyln/testing/db.py
```diff
@@ -117,28 +116,24 @@ def __init__(self, directory):
...
-            postgres = os.path.join(v, 'bin', 'postgres')
-            if os.path.isfile(initdb) and os.path.isfile(postgres):
-                logging.info("Found `postgres` and `initdb` in {}".format(os.path.join(v, 'bin')))
+        # Use `pg_config` to determine correct PostgreSQL installation
+        pg_config = shutil.which('pg_config')
+        if not pg_config:
+            raise ValueError("Could not find `pg_config` to determine PostgreSQL binaries. Is PostgreSQL installed?")
+
+        proc = subprocess.Popen([pg_config, '--bindir'], stdout=subprocess.PIPE)
+        bindir = io.TextIOWrapper(proc.stdout).readline().rstrip()
```
cdecker, line 125: The easier way to get the output of a subprocess is to use `check_output`:
```suggestion
        bindir = subprocess.check_output([pg_config, '--bindir'], stdout=subprocess.PIPE).decode('ASCII').rstrip()
```

m-schmoock: Thx. I modified it slightly by skipping the `stdout=subprocess.PIPE` as it works without it for `check_output`. I think this is only required for error pipes etc. Also I left the `decode()` without argument, the default is UTF8 for that. Or do you think we need to decode it as ASCII for a specific reason? Maybe some very strange distribution is using path names that contain UTF8 chars for `pg_config` dirs...

rustyrussell (approved):
Ack 798ebca88978800c78d82262937119da6eb04402

PR #4155, opened 2020-10-27
gossmap: fix return of gossmap_xxx_has_feature, rename.

rustyrussell:
1. One place returned false instead of -1.
2. The names implied it returned a bool, and it doesn't.

Fix both, and curse C's loose typing a little.

Changelog-None


jsarenik:
ACK 3b69dce00

Tested.

PR #4157, opened 2020-10-27
Updates to dual-funding spec

niftynei:
Here, we incorporate two new changes to the draft spec:
- `serial_id` goes from a u16 to u64 (this is to avoid collisions when doing multifundchannels)
- We update the feerate protocol. Instead of the opener setting a 'firm' feerate, we allow the accepter to choose a feerate within a provided range. This PR includes updates turn the current 'implicit' acceptable range to be explicit (in the role of the opener). In the accepter case, we merely delegate the decision to the plugin via the `openchannel2` hook.

Changelog-None.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -2114,7 +2114,7 @@ static void handle_tx_sigs(struct peer *peer, const u8 *msg)
 	for (size_t i = 0; i < peer->psbt->num_inputs; i++) {
 		struct wally_psbt_input *in =
 			&peer->psbt->inputs[i];
-		u16 in_serial;
+		u64 in_serial;
 		const struct witness_element **elem;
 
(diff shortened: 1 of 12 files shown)


jsarenik:
Just an idea: would the change of `serial_id` from `u16` to `u64` require any SQL migration? In other words, are the 16-bit values stored in the db (SQLite or Postgres) with a particular bit-width?

jsarenik:
The one failed build on Travis seems to be unrelated.

niftynei:
> Just an idea: would the change of serial_id from u16 to u64 require any SQL migration? In other words, are the 16-bit values stored in the db (SQLite or Postgres) with a particular bit-width?

These values don't get written to the database, so that's not a concern for this patchset. :+1:

rustyrussell (approved):
Ack 91c1dfd9a064c3d988c3a4ed6bd40965249ef00b

PR #4162, opened 2020-10-27
lightningd: note whether payment secret is used or not.

rustyrussell:
(Somehow I missed pushing this up; was meant to be in previous release)

This will allow nodes (with log-level=debug) to gather how many payments
are made without payment_secrets.  We need to know this so we know when
we can make them compulsory.

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4162/commits/ade87d92427707d497bd8edd92bd384c6d22cf9b

Issue #4005, opened 2020-09-03
Plugin hook ordering


cdecker:
@m-schmoock points out that the order in which hooks are added to the chain,
and processed is not guaranteed to match the order in which the plugins are
specified on the command line, and may be randomly interleaved.

The current theory is that this happens due to the concurrent call to
`getmanifest` and the order in which the response gets processed, i.e., the
order in which the hooks get added to the chain. In order to address this
issue we should ensure that the hook order matches the order in which the
corresponding plugin was specified on the command line. This can be achieved
in several ways:

 - Call the `getmanifest` sequentially. Pros: simple, cons: delays startup time.
 - Process the `getmanifest` results in the plugin order. Pros: also
   relatively simple, cons: requires stashing of the results and processing in
   a non-request-response manner, memory overhead.
 - Assign each hook registration a priority based on the plugin order, and
   sort before calling `init` on the plugin. Pros: simple, fast. Cons: adds a
   tiny bit of computation and memory overhead.
   
I personally tend to using the third option, since it is quick and has low
overhead. In addition we can use the priority in future to either allow users
to specify explicit priorities on the command line, or the plugins to specify
priorities for each hook independently, freeing us from the cli order and
adding flexibility. However these additions are likely out of scope for this
issue and need to be refined.

#### Details

 - [IRC logs] of the initial discussion.
 
 
[IRC logs]: http://gnusha.org/c-lightning/2020-09-03.log

rustyrussell:
Explicit ordering is always better, in future some plugins may need to be ordered with respect to each other. There are several ways of doing this. Simplest is a priority number. Best is before and after lists.

So, the ideal is:
1. Allow before and after lists in hook registration.
2. Do partial order sort by those.
3. Secondary sort by command line order, then config order.

PR #4168, opened 2020-10-29
Order specifications for hooks

rustyrussell:
This allows hooks to specify they must be before/after some other plugins.  Either by basename or full pathname.

As our plugin ecosystem grows and gets more sophisticated, such conflicts will inevitably occur where multiple plugins want to ensure they are called in certain order.  

We could implement a priority number system, but that's usually just window-dressing on what people really want (though, unlike this, would allow you to specify that you must be first, or last: we could add wildcards if we wanted to).

Getting this in now means that after a few releases, plugin authors will simply be able to assume its existence.

Fixes: #4005 (thanks @m-schmoock for pointing that out!)

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -57,6 +57,8 @@ class Method(object):
         self.desc = desc
         self.long_desc = long_desc
         self.deprecated = deprecated
+        self.before: List[str] = []
+        self.after: List[str] = []
 
 
 class Request(dict):
@@ -405,7 +407,9 @@ class Plugin(object):
         return decorator
 
     def add_hook(self, name: str, func: Callable[..., JSONType],
-                 background: bool = False) -> None:
+                 background: bool = False,
+                 before: Optional[List[str]] = None,
+                 after: Optional[List[str]] = None) -> None:
         """Register a hook that is called synchronously by lightningd on events
         """
         if name in self.methods:
@@ -427,15 +431,23 @@ class Plugin(object):
 
         method = Method(name, func, MethodType.HOOK)
         method.background = background
+        method.before = []
+        if before:
+            method.before = before
+        method.after = []
+        if after:
+            method.after = after
         self.methods[name] = method
 
-    def hook(self, method_name: str) -> JsonDecoratorType:
+    def hook(self, method_name: str,
+             before: List[str] = None,
+             after: List[str] = None) -> JsonDecoratorType:
         """Decorator to add a plugin hook to the dispatch table.
 
         Internally uses add_hook.
         """
         def decorator(f: Callable[..., JSONType]) -> Callable[..., JSONType]:
-            self.add_hook(method_name, f, background=False)
+            self.add_hook(method_name, f, background=False, before=before, after=after)
             return f
         return decorator
 
@@ -689,7 +701,9 @@ class Plugin(object):
                 continue
 
             if method.mtype == MethodType.HOOK:
-                hooks.append(method.name)
+                hooks.append({'name': method.name,
+                              'before': method.before,
+                              'after': method.after})
                 continue
 
             doc = inspect.getdoc(method.func)
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -688,11 +688,15 @@ The call semantics of the hooks, i.e., when and how hooks are called, depend
 on the hook type. Most hooks are currently set to `single`-mode. In this mode
 only a single plugin can register the hook, and that plugin will get called
 for each event of that type. If a second plugin attempts to register the hook
-it gets killed and a corresponding log entry will be added to the logs. In
-`chain`-mode multiple plugins can register for the hook type and they are
-called in any order which meets their `before` and `after` requirements
-if a matching event is triggered. Each plugin can then
-handle the event or defer by returning a `continue` result like the following:
+it gets killed and a corresponding log entry will be added to the logs.
+
+In `chain`-mode multiple plugins can register for the hook type and
+they are called in any order they are loaded (i.e. cmdline order
+first, configuration order file second: though note that the order of
+plugin directories is implementation-dependent), overriden only by
+`before` and `after` requirements the plugin's hook registrations specify.
+Each plugin can then handle the event or defer by returning a
+`continue` result like the following:
 
 ```json
 {
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -93,8 +93,8 @@ example:
     "disconnect"
   ],
   "hooks": [
-    "openchannel",
-    "htlc_accepted"
+    { "name": "openchannel", "before": ["another_plugin"] },
+    { "name": "htlc_accepted" }
   ],
   "features": {
     "node": "D0000000",
@@ -678,13 +678,20 @@ declares that it'd like to be consulted on what to do next for certain
 events in the daemon. A hook can then decide how `lightningd` should
 react to the given event.
 
+When hooks are registered, they can optionally specify "before" and
+"after" arrays of plugin names, which control what order they will be
+called in.  If a plugin name is unknown, it is ignored, otherwise if the
+hook calls cannot be ordered to satisfy the specifications of all
+plugin hooks, the plugin registration will fail.
+
 The call semantics of the hooks, i.e., when and how hooks are called, depend
 on the hook type. Most hooks are currently set to `single`-mode. In this mode
 only a single plugin can register the hook, and that plugin will get called
 for each event of that type. If a second plugin attempts to register the hook
 it gets killed and a corresponding log entry will be added to the logs. In
 `chain`-mode multiple plugins can register for the hook type and they are
-called sequentially if a matching event is triggered. Each plugin can then
+called in any order which meets their `before` and `after` requirements
+if a matching event is triggered. Each plugin can then
 handle the event or defer by returning a `continue` result like the following:
 
 ```json
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -1078,9 +1078,12 @@ static const char *plugin_hooks_add(struct plugin *plugin, const char *buffer,
 			name = json_strdup(tmpctx, buffer, nametok);
 			beforetok = json_get_member(buffer, t, "before");
 			aftertok = json_get_member(buffer, t, "after");
-		} else {
+		} else if (deprecated_apis) {
 			name = json_strdup(tmpctx, plugin->buffer, t);
 			beforetok = aftertok = NULL;
+		} else {
+			return tal_fmt(plugin,
+				    "hooks must be an array of objects");
 		}
 
 		hook = plugin_hook_register(plugin, name);
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -117,7 +117,6 @@ static void destroy_plugin(struct plugin *p)
 {
 	struct plugin_rpccall *call;
 
-	plugin_hook_unregister_all(p);
 	list_del(&p->list);
 
 	/* Terminate all pending RPC calls with an error. */
@@ -1057,20 +1056,43 @@ static const char *plugin_subscriptions_add(struct plugin *plugin,
 static const char *plugin_hooks_add(struct plugin *plugin, const char *buffer,
 				    const jsmntok_t *resulttok)
 {
-	const jsmntok_t *hookstok = json_get_member(buffer, resulttok, "hooks");
+	const jsmntok_t *t, *hookstok, *beforetok, *aftertok;
+	size_t i;
+
+	hookstok = json_get_member(buffer, resulttok, "hooks");
 	if (!hookstok)
 		return NULL;
 
-	for (int i = 0; i < hookstok->size; i++) {
-		char *name = json_strdup(tmpctx, plugin->buffer,
-					 json_get_arr(hookstok, i));
-		if (!plugin_hook_register(plugin, name)) {
+	json_for_each_arr(i, t, hookstok) {
+		char *name;
+		struct plugin_hook *hook;
+
+		if (t->type == JSMN_OBJECT) {
+			const jsmntok_t *nametok;
+
+			nametok = json_get_member(buffer, t, "name");
+			if (!nametok)
+				return tal_fmt(plugin, "no name in hook obj %.*s",
+					       json_tok_full_len(t),
+					       json_tok_full(buffer, t));
+			name = json_strdup(tmpctx, buffer, nametok);
+			beforetok = json_get_member(buffer, t, "before");
+			aftertok = json_get_member(buffer, t, "after");
+		} else {
+			name = json_strdup(tmpctx, plugin->buffer, t);
+			beforetok = aftertok = NULL;
+		}
+
+		hook = plugin_hook_register(plugin, name);
+		if (!hook) {
 			return tal_fmt(plugin,
 				    "could not register hook '%s', either the "
 				    "name doesn't exist or another plugin "
 				    "already registered it.",
 				    name);
 		}
+
+		plugin_hook_add_deps(hook, plugin, buffer, beforetok, aftertok);
 		tal_free(name);
 	}
 	return NULL;
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -83,24 +83,46 @@ void plugins_free(struct plugins *plugins)
 	tal_free(plugins);
 }
 
-static void check_plugins_resolved(struct plugins *plugins)
+/* Once they've all replied with their manifests, we can order them. */
+static void check_plugins_manifests(struct plugins *plugins)
 {
-	/* As startup, we break out once all getmanifest are returned */
-	if (plugins->startup) {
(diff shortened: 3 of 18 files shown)


m-schmoock:
This addresses https://github.com/ElementsProject/lightning/issues/4005

lightningd/plugin.c
```diff
@@ -1057,20 +1056,43 @@ static const char *plugin_subscriptions_add(struct plugin *plugin,
...
+		if (t->type == JSMN_OBJECT) {
+			const jsmntok_t *nametok;
+
+			nametok = json_get_member(buffer, t, "name");
+			if (!nametok)
+				return tal_fmt(plugin, "no name in hook obj %.*s",
+					       json_tok_full_len(t),
+					       json_tok_full(buffer, t));
+			name = json_strdup(tmpctx, buffer, nametok);
+			beforetok = json_get_member(buffer, t, "before");
```
niftynei, line 1079: it took me a second to fully grasp the implication of 'before' and 'after'. Renaming them to "run_before" and "run_after" might be more immediately obvious?

niftynei (approved):
ACK 0c1d19b4c8ec4730a969551806224399faf0bae

lightningd/plugin.c
rustyrussell: It's more obvious in context, than when reading in a patch, e.g.:

```json
"hooks": [
    {"name": "htlc_accepted",
     "before": ["someplugin.py"],
     "after": ["foo.py"]}
]
```

m-schmoock:
concept ack

@rustyrussell  as you wrote in 4005
> 3\. Secondary sort by command line order, then config order.

I think it is also very helpful to give the user an option for plugin ordering (unless otherwise specified by the devs via before/after) by checking commandline and config order of plugins. The rationale behind is it the following: Before and after list are to be maintained by the plugin developers. Before that happens, a user typically has the problem first and needs to wait for the devs to to make the changes so his specific setup works. Since the devs will often not know all the plugin combinations out in the wild this can take a while or, in the worst case, never happens.

So I think we might also add the commandline/config order as 'last resort/default' if no before/after ordering has been specified for a plugin.

rustyrussell:
The current implementation has this effect, in fact. Technically it's not quite a stable sort, as dependencies can pull in things out of order, but we could fix that (I doubt anyone will notice, though).

For example, assume plugins in alphabetical order (they're in directory-sort order usually BTW). A has to be before D, and B has to be before C.

In this case, algorithm will go A B D C. Easy to fix though.

rustyrussell:
OK, fixed to make the default ordering equal to the specified one.

Also, removed a problem where we'd warn about missing dependencies since we evaluated them *as we loaded*.  This was more invasive, but I think simplifies the code a little anyway.

m-schmoock:
Thanks @rustyrussell 

Super wired, I stress tested your changes a little, but it seems there is an issue with the default commandline/config order (unless otherwise by before/after specified).

### How to reproduce
- check out ece24e5c9
- compile
- apply a test change, that now should work (see below)
- run the tests in parallel:  `py.test --count 50 -x -n 16 -v tests/test_plugin.py -k test_openchannel_hook_chaining`

### What happens
The flaky test can fail. When we look in the logs of l2 we see that the 'reject for a reason' hook is not called but 'reject by design', which should never happen if correct default order would be applied.

### The patch for the test that uses the new feature
```
diff --git a/tests/test_plugin.py b/tests/test_plugin.py
index d2628cd25..9f35a4dc5 100644
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -606,24 +606,15 @@ def test_openchannel_hook_chaining(node_factory, bitcoind):
     l1, l2 = node_factory.line_graph(2, fundchannel=False, opts=opts)
     l1.fundwallet(10**6)
 
-    hook_msg = "openchannel_hook rejects and says '"
     # 100005sat fundchannel should fail fatal() for l2
     # because hook_accepter.py rejects on that amount 'for a reason'
     with pytest.raises(RpcError, match=r'They sent error channel'):
         l1.rpc.fundchannel(l2.info['id'], 100005)
 
-    # Note: hook chain order is currently undefined, because hooks are registerd
-    #       as a result of the getmanifest call, which may take some random time.
-    #       We need to workaround that fact, so test can be stable
-    correct_order = l2.daemon.is_in_log(hook_msg + "reject for a reason")
-    if correct_order:
-        assert l2.daemon.wait_for_log(hook_msg + "reject for a reason")
-        # the other plugin must not be called
-        assert not l2.daemon.is_in_log("reject on principle")
-    else:
-        assert l2.daemon.wait_for_log(hook_msg + "reject on principle")
-        # the other plugin must not be called
-        assert not l2.daemon.is_in_log("reject for a reason")
+    hook_msg = "openchannel_hook rejects and says '"
+    assert l2.daemon.wait_for_log(hook_msg + "reject for a reason")
+    # the third plugin must now not be called anymore
+    assert not l2.daemon.is_in_log("reject on principle")
 
     # 100000sat is good for hook_accepter, so it should fail 'on principle'
     # at third hook openchannel_reject.py
```

rustyrussell:
Argh, good catch: final patch actually fixes the problem, and for some reason I didn't push it!

rustyrussell:
And I turned your patch into a commit, credited you and appended. Thanks!

m-schmoock:
works like a charm

m-schmoock:
Regarding IRC chat:
> `00:22 < rusty> m-schmoock: how do you feel about getting rid of all those single hooks?  Or at least as many of them as possible...`

Can you elaborate on this?

lightningd/plugin.c
```diff
@@ -125,10 +147,16 @@ static void destroy_plugin(struct plugin *p)
...
-	/* Don't call this if we're still parsing options! */
-	if (p->plugin_state != UNCONFIGURED)
-		check_plugins_resolved(p->plugins);
+	/* If this was last one manifests were waiting for, handle deps */
+	if (p->plugin_state == AWAITING_GETMANIFEST_RESPONSE)
+		check_plugins_manifests(p->plugins);
+
+	/* If this was the last one init was waiting for, handle cmd replies */
+	if (p->plugin_state == AWAITING_INIT_RESPONSE)
+		check_plugins_initted(p->plugins);
```
cdecker, line 160: This seems to be a weird place to perform these checks. We look at a random plugin being destroyed and if that was in a specific state we assume it must have been the last one in this state? We should kick this up to the set of plugins and perform the check there after removing the plugin to be destroyed. No need to replicate the last-one-turns-off-the-light test here :-)

lightningd/plugin_hook.c
```diff
@@ -434,6 +434,21 @@ static struct hook_node *find_hook(struct hook_node *graph, const char *name)
...
+	size_t pos, s_num;
+
+	s_num = tal_count(*s);
+	for (pos = 0; pos < s_num; pos++) {
+		if ((*s)[pos]->hook->plugin->index > n->hook->plugin->index)
+			break;
+	}
+	tal_resize(s, s_num + 1);
+	memmove(*s + pos+1, *s + pos, sizeof(**s) * (s_num - pos));
+	(*s)[pos] = n;
```
cdecker, line 449: I don't quite follow why this is working: we're striking dependencies off of the graph node until we finally find that it has no more, then add it to the list. However, before this change we added it to the end of the list, which was always safe (no dependency could jump over us). After this change we suddenly use insertion sort on the `index` at a potentially earlier position in the list, meaning we could end up being inserted before one of our dependencies. What am I missing here?

lightningd/plugin_hook.c
```diff
@@ -392,3 +392,125 @@ void plugin_hook_add_deps(struct plugin_hook *hook,
...
+			graph[i].num_incoming++;
+		}
+	}
+
+	/* Populate array of ready-to-go nodes. */
+	s = tal_arr(graph, struct hook_node *, 0);
+	for (size_t i = 0; i < tal_count(graph); i++) {
+		if (graph[i].num_incoming == 0)
+			tal_arr_expand(&s, &graph[i]);
+	}
```
cdecker: This pre-populating is not necessary since it's just a special case for the below logic. The later topological sort will also make sure that this isn't required to maintain order for hooks without dependencies.

cdecker (commented):
Some minor questions for my own understanding, and a tiny python bug, otherwise a very solid PR :+1:

lightningd/plugin_hook.c
rustyrussell: You're right!

We should instead gather all the things we are going to add from one edge, and order *those* then add as a group.  That means if B and C depend on A, we append A, then sort (B,C) and append those.

I will write a test which triggers this first...

contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -405,7 +407,9 @@ def decorator(f: Callable[..., JSONType]) -> Callable[..., JSONType]:
         return decorator
 
     def add_hook(self, name: str, func: Callable[..., JSONType],
-                 background: bool = False) -> None:
+                 background: bool = False,
+                 before: List[str] = [],
+                 after: List[str] = []) -> None:
```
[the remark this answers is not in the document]
rustyrussell: Oh, nice catch, thanks!

lightningd/plugin.c
rustyrussell: No, we've already unlisted the plugin at this point.  We could, and probably should simply call check_plugins_xxx() unconditionally here.

I wonder if we should have a plugin_set_state() which calls these internally.  Less efficient, but clearer?

lightningd/plugin_hook.c
rustyrussell: No, we need it because we only add the other candidates when we look at their count:

```
			if (--n->outgoing[i]->num_incoming == 0)
				add_to_s(&s, n->outgoing[i]);
```

Nodes which start with no incoming are not caught here.

rustyrussell:
I ended up reworking and thus simplifying the algo.  It's less efficient (O(N^2)) but nobody cares.

lightningd/plugin.c
cdecker: Similar to the `payment_continue` function that is used to continue in a state machine? I'm all for it ^^

lightningd/plugin_hook.c
cdecker: Ah, gotcha, I had not seen the issue at first.

contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -431,13 +431,15 @@ def add_hook(self, name: str, func: Callable[..., JSONType],
...
+            method.before = before
+        if after:
+            method.after = after
         self.methods[name] = method
 
     def hook(self, method_name: str,
-             before: List[str] = [],
-             after: List[str] = []) -> JsonDecoratorType:
+             before: List[str] = None,
+             after: List[str] = None) -> JsonDecoratorType:
```
cdecker, line 442: This now becomes an `Optional[List[str]]`

contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -431,13 +431,15 @@ def add_hook(self, name: str, func: Callable[..., JSONType],
 
         method = Method(name, func, MethodType.HOOK)
         method.background = background
-        method.before = before
-        method.after = after
+        if before:
+            method.before = before
+        if after:
+            method.after = after
```
cdecker, line 437: We should always set these, otherwise we'd have to check with `hasattr` whether the hook has that field.

```
method.before = [] if before else before
method.after = [] if after else after
```

lightningd/plugin_hook.c
```diff
@@ -401,25 +402,8 @@ void plugin_hook_add_deps(struct plugin_hook *hook,
...
- *         if m has no other incoming edges then
- *             insert m into S
- *
- * if graph has edges then
- *     return error   (graph has at least one cycle)
- * else
- *     return L   (a topologically sorted order)
- */
 struct hook_node {
+	bool finished;
```
cdecker, line 406: What are the semantics of `finished`? It's only used in sorting so I'd assume it means that it is in its final position.

lightningd/plugin_hook.c
```diff
@@ -434,31 +418,33 @@ static struct hook_node *find_hook(struct hook_node *graph, const char *name)
...
-	for (pos = 0; pos < s_num; pos++) {
-		if ((*s)[pos]->hook->plugin->index > n->hook->plugin->index)
-			break;
+	for (size_t i = 0; i < tal_count(graph); i++) {
+		if (graph[i].finished)
+			continue;
+		if (graph[i].num_incoming != 0)
+			continue;
+		if (!best
+		    || best->hook->plugin->index > graph[i].hook->plugin->index)
```
cdecker, line 432: We're never moving the plugin in the `graph`, so under no circumstance should this ever be true (a later iteration should always have a higher `index`). Since that's the case I think we can remove the `index` altogether, and break/return after finding a first `best`

contrib/pyln-client/pyln/client/plugin.py
cdecker: Sorry for forgetting this in my first round of comments :-)

m-schmoock:
Hm, I restarted a travis flake which I couldn't reproduce locally also not by running it 100 times and parallel: `tests/test_plugin.py::test_plugin_command`

It was complaining about unclean node shutdown, but it succeeded upon first restart/retry.

m-schmoock:
@rustyrussell still (sorry for bugging):
> regarding IRC chat:
> 
> > `00:22 < rusty> m-schmoock: how do you feel about getting rid of all those single hooks?  Or at least as many of them as possible...`
> 
> Can you elaborate on this?
> 
> I was thinking that it would be sufficient to have the `before` and `after` lists on a per plugin level and not per hook. It would reduce (configuration) complexity. Having it on a per hook basis brings the option to have two different plugins with two (same) hooks each order their execution differently depending on what hook is called. This seems a bit overkill. Having plugin order rather than hook order would just prioritize one plugin over another, which will likely do the job.

lightningd/plugin_hook.c
rustyrussell: That's making an assumption that graph is in index order.  It (I'm fairly sure) starts that way, but can be perturbed by previous calls to plugin_hook_make_ordered.  If plugin B says it has to be before A and after C, we'll go C B A.  If B gets removed, we should return to A C.

Does anyone care?  Probably not, but this way is clear, at least.

rustyrussell:
> @rustyrussell still (sorry for bugging):
> 
> > regarding IRC chat:
> > > `00:22 < rusty> m-schmoock: how do you feel about getting rid of all those single hooks?  Or at least as many of them as possible...`
> > 
> > 
> > Can you elaborate on this?

We should make it so every hook can be registered by multiple plugins.

> > I was thinking that it would be sufficient to have the `before` and `after` lists on a per plugin level and not per hook. It would reduce (configuration) complexity. Having it on a per hook basis brings the option to have two different plugins with two (same) hooks each order their execution differently depending on what hook is called. This seems a bit overkill. Having plugin order rather than hook order would just prioritize one plugin over another, which will likely do the job.

I agree, at the moment.  But hooks are the only place where plugins interact, so I'd rather see explicitly what they care about.  If we ever get order conflicts, it's easier to see what they want when they have priorities on specific hooks.

lightningd/plugin_hook.c
rustyrussell: Included in sorted results. I'll add a comment

lightningd/plugin_hook.c
rustyrussell: On further thought, if we define it as "minimum perturbation to achieve required ordering" which is a totally reasonable goal, we can indeed do this simplification.

rustyrussell: ... Nope.  Hooks start in "who returned from getmanifest" order which is not the same as "order specified by config" order.

rustyrussell:
Trivial rebase to fold in fixups and add a comment on `finished` var.

m-schmoock:
>>> 00:22 < rusty> m-schmoock: how do you feel about getting rid of all those single hooks? Or at least as many of them as possible...

>> Can you elaborate on this?

> We should make it so every hook can be registered by multiple plugins.

absolutely yes, I missed the point that you were referring to 'unchained' hooks.

doc/PLUGINS.md
```diff
@@ -678,14 +678,25 @@ declares that it'd like to be consulted on what to do next for certain
 events in the daemon. A hook can then decide how `lightningd` should
 react to the given event.
 
+When hooks are registered, they can optionally specify "before" and
+"after" arrays of plugin names, which control what order they will be
```
m-schmoock, line 682: General question about 'plugin names': Is this same as for `cli plugin start/stop` commands, meaning either full path (useless for this feature) or basename? If so, we may note that the plugin name includes its extension, meaning full filename.

contrib/pyln-client/pyln/client/plugin.py
rustyrussell: Hmm, mypy didn't like that, so went for an open-coded variant.

doc/PLUGINS.md
rustyrussell: Yes, filename or full path name.

m-schmoock:
ACK 7d41997

Issue #4110, opened 2020-10-06
Feature Request: Is it possible to retrieve the xPub of the on-chain wallet?


saubyk:
Is it possible to retrieve the xPub of the on-chain wallet?

- This can help simplify fund management on implementations like BTCPayserver, where currently the Lightning implementations run as a separate hot wallet from the main one.
- Enable other privacy improvements which BTCPayserver is interested in experimenting with (cc: @kukks)

darosior:
That's on my TODO list fwiw.. (as part of outputting [descriptors for the onchain wallet in `hsm_tool`](https://github.com/ElementsProject/lightning/wiki#implementation-1))

cdecker:
> That's on my TODO list fwiw.. (as part of outputting [descriptors for the onchain wallet in `hsm_tool`](https://github.com/ElementsProject/lightning/wiki#implementation-1))

Should I assign this to you then?

darosior:
Yeah let's do it :-)

PR #4171, opened 2020-10-30
Hsmtool: enable dumping output descriptors of onchain wallet

darosior:
This adds a new `dumponchaindescriptors` command to `hsmtool` which can:
- Dump 2 `xpub`-based descriptors (one for each scriptPubKey type we create)
    - Useful for external service (see https://github.com/ElementsProject/lightning/issues/4110 for more details)
- ~~Possibly dump 2 `xpriv`-based descriptors (one for each scriptPubKey type we create)~~ Removed: footgun with anchor outputs (thanks niftynei!)
    - ~~Useful for backing up / restoring the onchain wallet easily~~

TODO:
- [x] A functional test restoring a descriptor on `bitcoind` asserting we retrieve all the funds
- [x] Manpage and an extended doc on how to restore the onchain wallet using `bitcoind`
- [x] Maybe add a network parameter finally ?.. ~~EDIT: maybe output both main and test extended keys ?..~~ EDIT 2: added a `testnet` parameter

Fixes https://github.com/ElementsProject/lightning/issues/4110
Ping @jb55 : finally got to https://github.com/jb55/clightning-dumpkeys/issues/2 :-)

### DIFF
--- a/common/descriptor_checksum.c
+++ b/common/descriptor_checksum.c
@@ -1,0 +1,82 @@
+#include <ccan/short_types/short_types.h>
+#include <common/descriptor_checksum.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+
+static const char CHECKSUM_CHARSET[] = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+
+static const char INPUT_CHARSET[] =
+	"0123456789()[],'/*abcdefgh@:$%{}"
+	"IJKLMNOPQRSTUVWXYZ&+-.;<=>?!^_|~"
+	"ijklmnopqrstuvwxyzABCDEFGH`#\"\\ ";
+
+static inline int charset_find(char ch) {
+	for (size_t i = 0; i < sizeof(INPUT_CHARSET); i++) {
+		if (INPUT_CHARSET[i] == ch)
+			return i;
+	}
+	return -1;
+}
+
+static u64 polymod(u64 c, int val)
+{
+	u8 c0 = c >> 35;
+	c = ((c & 0x7ffffffff) << 5) ^ val;
+	if (c0 & 1) c ^= 0xf5dee51989;
+	if (c0 & 2) c ^= 0xa9fdca3312;
+	if (c0 & 4) c ^= 0x1bab10e32d;
+	if (c0 & 8) c ^= 0x3706b1677a;
+	if (c0 & 16) c ^= 0x644d626ffd;
+	return c;
+}
+
+
+bool descriptor_checksum(const char *descriptor, int desc_size,
+			 struct descriptor_checksum *checksum)
+{
+	checksum->csum[0] = 0;
+
+	int j;
+	u64 c = 1;
+	int cls = 0;
+	int clscount = 0;
+
+	for (int i = 0; i < desc_size; i++) {
+		char ch = descriptor[i];
+		int pos = charset_find(ch);
+		if (pos == -1) {
+			checksum->csum[0] = 0;
+			return false;
+		}
+		/* Emit a symbol for the position inside the group, for every
+		 * character. */
+		c = polymod(c, pos & 31);
+
+		/* Accumulate the group numbers */
+		cls = cls * 3 + (pos >> 5);
+
+		if (++clscount == 3) {
+			c = polymod(c, cls);
+			cls = 0;
+			clscount = 0;
+		}
+	}
+
+	if (clscount > 0)
+		c = polymod(c, cls);
+
+	/* Shift further to determine the checksum. */
+	for (j = 0; j < DESCRIPTOR_CHECKSUM_LENGTH; ++j)
+		c = polymod(c, 0);
+
+	/* Prevent appending zeroes from not affecting the checksum. */
+	c ^= 1;
+
+	for (j = 0; j < DESCRIPTOR_CHECKSUM_LENGTH; ++j)
+		checksum->csum[j] = CHECKSUM_CHARSET[(c >> (5 * (7 - j))) & 31];
+
+	checksum->csum[DESCRIPTOR_CHECKSUM_LENGTH] = 0;
+
+	return true;
+}
--- a/common/descriptor_checksum.h
+++ b/common/descriptor_checksum.h
@@ -1,0 +1,16 @@
+#ifndef LIGHTNING_COMMON_DESCRIPTOR_CHECKSUM_H
+#define LIGHTNING_COMMON_DESCRIPTOR_CHECKSUM_H
+#include "config.h"
+#include <stdbool.h>
+
+/* https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md#reference */
+#define DESCRIPTOR_CHECKSUM_LENGTH 8
+
+struct descriptor_checksum {
+	char csum[DESCRIPTOR_CHECKSUM_LENGTH + 1];
+};
+
+bool descriptor_checksum(const char *descriptor, int desc_size,
+			 struct descriptor_checksum *checksum);
+
+#endif /* LIGHTNING_COMMON_DESCRIPTOR_CHECKSUM_H */
--- a/contrib/pyln-testing/pyln/testing/fixtures.py
+++ b/contrib/pyln-testing/pyln/testing/fixtures.py
@@ -132,6 +132,14 @@ def bitcoind(directory, teardown_checks):
         raise ValueError("bitcoind is too old. At least version 16000 (v0.16.0)"
                          " is needed, current version is {}".format(info['version']))
 
+    # Make sure we have a wallet, starting with 0.21 there is no default wallet
+    # anymore.
+    # FIXME: if we update the testsuite to use the upcoming 0.21 release we
+    # could switch to descriptor wallets and speed bitcoind operations
+    # consequently.
+    if not bitcoind.rpc.listwallets():
+        bitcoind.rpc.createwallet("lightningd-tests")
+
     info = bitcoind.rpc.getblockchaininfo()
     # Make sure we have some spendable funds
     if info['blocks'] < 101:
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -1045,6 +1045,36 @@ def test_hsmtool_secret_decryption(node_factory):
     assert node_id == l1.rpc.getinfo()["id"]
 
 
+@unittest.skipIf(TEST_NETWORK == 'liquid-regtest', '')
+def test_hsmtool_dump_descriptors(node_factory, bitcoind):
+    l1 = node_factory.get_node()
+    l1.fundwallet(10**6)
+
+    # Get a tpub descriptor of lightningd's wallet
+    hsm_path = os.path.join(l1.daemon.lightning_dir, TEST_NETWORK, "hsm_secret")
+    cmd_line = ["tools/hsmtool", "dumponchaindescriptors", hsm_path, "",
+                "testnet"]
+    out = subprocess.check_output(cmd_line).decode("utf8").split("\n")
+    descriptor = [l for l in out if l.startswith("wpkh(tpub")][0]
+
+    # Import the descriptor to bitcoind
+    # FIXME: if we update the testsuite to use the upcoming 0.21 we could use
+    # importdescriptors instead.
+    bitcoind.rpc.importmulti([{
+        "desc": descriptor,
+        # No need to rescan, we'll transact afterward
+        "timestamp": "now",
+        # The default
+        "range": [0, 99]
+    }])
+
+    # Funds sent to lightningd can be retrieved by bitcoind
+    addr = l1.rpc.newaddr()["bech32"]
+    txid = l1.rpc.withdraw(addr, 10**3)["txid"]
+    bitcoind.generate_block(1, txid)
+    assert len(bitcoind.rpc.listunspent(1, 1, [addr])) == 1
+
+
 # this test does a 'listtransactions' on a yet unconfirmed channel
 def test_fundchannel_listtransaction(node_factory, bitcoind):
     l1, l2 = node_factory.get_nodes(2)
--- a/tools/hsmtool.c
+++ b/tools/hsmtool.c
@@ -6,8 +6,10 @@
 #include <ccan/read_write_all/read_write_all.h>
 #include <ccan/str/str.h>
 #include <ccan/tal/path/path.h>
+#include <ccan/tal/str/str.h>
 #include <common/bech32.h>
 #include <common/derive_basepoints.h>
+#include <common/descriptor_checksum.h>
 #include <common/node_id.h>
 #include <common/type_to_string.h>
 #include <common/utils.h>
@@ -39,6 +41,7 @@ static void show_usage(const char *progname)
 	printf("	- guesstoremote <P2WPKH address> <node id> <tries> "
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
 	printf("	- generatehsm <path/to/new//hsm_secret>\n");
+	printf("	- dumponchaindescriptors <path/to/hsm_secret> <password>\n");
 	exit(0);
 }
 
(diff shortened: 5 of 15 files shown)


darosior:
Ok, added a test, documented (and rebased on https://github.com/ElementsProject/lightning/pull/4146 so that i can build locally)

darosior:
Travis failure is an unrelated timeout

darosior:
Rebased on master after #4146 merge, added a missing changelog.

common/descriptor_checksum.c
```diff
@@ -0,0 +1,84 @@
...
+	if (c0 & 2) c ^= 0xa9fdca3312;
+	if (c0 & 4) c ^= 0x1bab10e32d;
+	if (c0 & 8) c ^= 0x3706b1677a;
+	if (c0 & 16) c ^= 0x644d626ffd;
+	return c;
+}
+
+
+int descriptor_checksum(const char *descriptor, int desc_size, char *checksum,
+			int checksum_capacity)
```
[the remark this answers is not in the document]
niftynei: Further, since this only ever returns `0` or `8`, you could update this to return a boolean.

tools/hsmtool.c
```diff
@@ -368,6 +371,55 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	u32 salt = 0;
+	struct ext_key master_extkey;
+	char *enc_xpub, *descriptor, checksum[DESCRIPTOR_CHECKSUM_LENGTH + 1];
+
+	if (passwd)
+		get_encrypted_hsm_secret(&hsm_secret, hsm_secret_path, passwd);
+	else
+		get_hsm_secret(&hsm_secret, hsm_secret_path);
+
+	/* We use m/0/0/k as the derivation tree for onchain funds. */
```
niftynei, line 390: Needs reference. Where in the codebase do we do this?

tools/hsmtool.c
```diff
@@ -368,6 +371,55 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	/* The root seed is derived from hsm_secret using hkdf.. */
+	do {
+		hkdf_sha256(bip32_seed, sizeof(bip32_seed),
+			    &salt, sizeof(salt),
+			    &hsm_secret, sizeof(hsm_secret),
+			    "bip32 seed", strlen("bip32 seed"));
+		salt++;
+		/* ..Which is used to derive m/ */
+	} while (bip32_key_from_seed(bip32_seed, sizeof(bip32_seed),
+				     /* An xpub can easily be converted to a tpub */
```
niftynei: ~No. This has potential to be *extremely* confusing. We need to either be told what chain to output it for or figure it out from the directory structure.~

updated later.

tools/hsmtool.c
```diff
@@ -368,6 +371,55 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	if (passwd)
+		get_encrypted_hsm_secret(&hsm_secret, hsm_secret_path, passwd);
+	else
+		get_hsm_secret(&hsm_secret, hsm_secret_path);
+
+	/* We use m/0/0/k as the derivation tree for onchain funds. */
+
+	/* The root seed is derived from hsm_secret using hkdf.. */
+	do {
+		hkdf_sha256(bip32_seed, sizeof(bip32_seed),
```
niftynei, line 394: Duplicating this code from `hsmd/hsmd.c#populate_secretstuff` is a bad idea. Needs to be a shared routine that both hsmd and here use.

tools/hsmtool.c
```diff
@@ -368,6 +371,55 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
...
+	} while (bip32_key_from_seed(bip32_seed, sizeof(bip32_seed),
+				     /* An xpub can easily be converted to a tpub */
+				     BIP32_VER_MAIN_PRIVATE,
+				     0, &master_extkey) != WALLY_OK);
+
+	if (bip32_key_to_base58(&master_extkey, BIP32_FLAG_KEY_PUBLIC, &enc_xpub) != WALLY_OK)
+		errx(ERROR_LIBWALLY, "Can't encode xpub");
+
+	/* Now we format the descriptor strings (we only ever create P2WPKH and
+	 * P2SH-P2WPKH outputs). */
```
niftynei, line 407: iirc, anchor outputs break this assumption

tools/hsmtool.c
```diff
@@ -417,6 +419,25 @@ static int dumponchaindescriptors(const char *hsm_secret_path, const char *passw
...
+	/* If they also want xpriv break a line and output them as well. */
+	if (private) {
+		printf("\n");
+		if (bip32_key_to_base58(&master_extkey, BIP32_FLAG_KEY_PRIVATE, &enc_xpub) != WALLY_OK)
+			errx(ERROR_LIBWALLY, "Can't encode xpub");
+
+		descriptor = tal_fmt(NULL, "wpkh(%s/0/0/*)", enc_xpub);
+		descriptor_checksum(descriptor, strlen(descriptor), checksum, sizeof(checksum));
+		printf("%s#%s\n", descriptor, checksum);
+		tal_free(descriptor);
```
niftynei, line 430: this is the 3rd time you're repeating this code (and you do it again below).

tools/hsmtool.c
```diff
@@ -417,6 +419,25 @@ static int dumponchaindescriptors(const char *hsm_secret_path, const char *passw
 
 	wally_free_string(enc_xpub);
 
+	/* If they also want xpriv break a line and output them as well. */
+	if (private) {
+		printf("\n");
+		if (bip32_key_to_base58(&master_extkey, BIP32_FLAG_KEY_PRIVATE, &enc_xpub) != WALLY_OK)
```
niftynei, line 424: you're writing the `enc_xpriv` to a variable named `enc_xpub`

tools/hsmtool.c
```diff
@@ -373,11 +373,12 @@ static int guess_to_remote(const char *address, struct node_id *node_id,
 }
 
 static int dumponchaindescriptors(const char *hsm_secret_path, const char *passwd,
-				  const bool private)
+				  const bool private, const bool testnet)
 {
 	struct secret hsm_secret;
 	u8 bip32_seed[BIP32_ENTROPY_LEN_256];
 	u32 salt = 0;
+	u32 version = testnet ? BIP32_VER_TEST_PRIVATE : BIP32_VER_MAIN_PRIVATE;
```
niftynei, line 381: Can we infer this from the directory structure or some other info source?

niftynei (changes requested):
Cool addition. I'm not sure how anchor outputs affects the assumption that all outputs are some variant of `p2wpkh`, may not be relevant for this patch.

darosior:
> I'm not sure how anchor outputs affects the assumption that all outputs are some variant of p2wpkh, may not be relevant for this patch.

You are right that we actually don't sweep the `1 CSV`-obstructed `to_remote`s in `onchaind` and just adverise them to the wallet, so i don't think i should actually sell this command as a "backup-and-recovery mechanism" in the FAQ..
I was also in doubt wrt dumping `xprivs` at all then, but it might turns to be useful (if i remove all mentions that it is exhaustive from the doc).

tools/hsmtool.c
darosior: No, as we can use all commands of the `hsmtool` on a `hsm_secret` without depending on its environment at all
and this would be quite convoluted too..

niftynei:
This functionality is Cool and Interesting, but until miniscript is widely adopted I think it's best to advertise this as a *subset* of your lightning funds tracking, at best. There's definitely "outputs we own, eventually" that haven't been swept to a 'recognizable wallet output' that will be missed by this -- if for some reason you take your lightning node down before these are swept there's a possibility that they'll be lost.

With these considerations in mind, I'd be tempted to remove the xpriv, with the rationale that it gives the false impression of 'total wallet funds control' when that's really not the case.

tools/hsmtool.c
```diff
@@ -469,7 +490,9 @@ int main(int argc, char *argv[])
 		if (argc < 3)
 			show_usage(argv[0]);
 
-		return dumponchaindescriptors(argv[2], argc > 3 ? argv[3] : NULL);
+		return dumponchaindescriptors(argv[2],
+					      argc > 3 && !streq(argv[3], "") ? argv[3] : NULL,
+					      argc > 4 && streq(argv[4], "private"));
```
niftynei: it's weird to me that you have to physically type in the phrase "private" and not look for a true/false here.

tools/hsmtool.c
```diff
@@ -492,7 +491,8 @@ int main(int argc, char *argv[])
 
 		return dumponchaindescriptors(argv[2],
 					      argc > 3 && !streq(argv[3], "") ? argv[3] : NULL,
-					      argc > 4 && streq(argv[4], "private"));
+					      argc > 4 && streq(argv[4], "private"),
+					      argc > 5 && streq(argv[5], "testnet"));
```
niftynei, line 495: same as previous comment, this seems like it should be checking for 'true/false' not "testnet"

niftynei (commented):
few more things I noticed while testing.

tools/hsmtool.c
darosior: That's how we've created tool since now ?..

common/descriptor_checksum.h
```diff
@@ -0,0 +1,11 @@
+#ifndef LIGHTNING_COMMON_DESCRIPTOR_CHECKSUM_H
+#define LIGHTNING_COMMON_DESCRIPTOR_CHECKSUM_H
+#include "config.h"
+
+/* https://github.com/bitcoin/bitcoin/blob/master/doc/descriptors.md#reference */
+#define TOOL_DESCRIPTOR_CHECKSUM_LENGTH 8
+
+int descriptor_checksum(const char *descriptor, int desc_size, char *checksum,
```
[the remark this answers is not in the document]
darosior: Apart from taking bitcoind's implementation (which is virtually this) and asserting they are equel in the fuzz target, i don't see how there could so not specifically for the checksum result.
However the test added later on would fail to import the descriptor if the checksum was incorrect, so there is already that.

common/descriptor_checksum.c
```diff
@@ -0,0 +1,84 @@
...
+		c = polymod(c, cls);
+
+	// Shift further to determine the checksum.
+	for (j = 0; j < 8; ++j)
+		c = polymod(c, 0);
+
+	// Prevent appending zeroes from not affecting the checksum.
+	c ^= 1;
+
+	for (j = 0; j < 8; ++j)
```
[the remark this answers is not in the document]
darosior: Yeah i messed up my history (again...)

common/descriptor_checksum.c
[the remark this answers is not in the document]
darosior: > and remove the usage of the 0 as a null terminator in this function

Argh actually it's a terrible idea: the checksum *is* a string.

> The Right Way to do this would be to introduce a new struct desc_checksum e.g.
struct desc_checksum {
        u8 csum[8];
}

Actually the right way is using `char`s

tools/hsmtool.c
darosior: This commit is now removed.

tools/hsmtool.c
darosior: This commit is now removed.

tools/hsmtool.c
darosior: I find it better: at usage typing `testnet` is way more meaningful than `true`

darosior:
- Used a proper API for the imported descriptor checksum file (use a struct instead and return a bool) as requested
- Addressed misc nits
- Removed the "output xprivs" functionality altogether, as @niftynei agrees it's a footgun. (We now use a watchonly wallet in the functional test).
- Rebased on master

tools/hsmtool.c
cdecker: We could end up with anything if we use the `close_to` arguments, that includes addresses that we don't even control ourselves. The tool makes it clear we only show on-chain addresses that were generated by the internal wallet, so I don't think this'll be an issue.

tools/hsmtool.c
cdecker: We have all of these parameters in `chainparams.c`, so why not use them? We can even steal the `--network` option from `lightningd` itself :-)

contrib/pyln-testing/pyln/testing/fixtures.py
```diff
@@ -132,6 +132,14 @@ def bitcoind(directory, teardown_checks):
         raise ValueError("bitcoind is too old. At least version 16000 (v0.16.0)"
                          " is needed, current version is {}".format(info['version']))
 
+    # Make sure we have a wallet, starting with 0.21 there is no default wallet
+    # anymore.
+    # FIXME: if we update the testsuite to use the upcoming 0.21 release we
+    # could switch to descriptor wallets and speed bitcoind operations
+    # consequently.
+    if not bitcoind.rpc.listwallets():
+        bitcoind.rpc.createwallet("lightningd-tests")
```
cdecker, line 141: This is slightly nicer than 90f4ea62e856fdac3adf1b59235bb5f69eb609d1

tests/test_wallet.py
```diff
@@ -1045,6 +1045,36 @@ def test_hsmtool_secret_decryption(node_factory):
     assert node_id == l1.rpc.getinfo()["id"]
 
 
+@unittest.skipIf(TEST_NETWORK == 'liquid-regtest', '')
```
cdecker, line 1048: Add a comment saying that this is only supported for `mainnet` and `testnet` (technically we'd require it to be `regtest` but that appears to match `testnet`).

darosior: Yeah descriptors for regtest are the same than for testnet (but addresses the same as mainnet!! :sob: )

tools/hsmtool.c
darosior: Yeah, now we don't promote it as a recovery mechanism anyways :-)

tools/hsmtool.c
darosior: This'd mean bringing the `opt()` system here: that's a much bigger change :/

tests/test_wallet.py
cdecker: Hurray for consistency 🤦

tools/hsmtool.c
niftynei: I don't follow. We really want this code to be shared so that if one changes, the other also gets updated.

tools/hsmtool.c
```diff
@@ -33,6 +35,7 @@ static void show_usage(const char *progname)
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
 	printf("	- guesstoremote <P2WPKH address> <node id> <tries> "
 	       "<path/to/hsm_secret> [hsm_secret password]\n");
+	printf("	- dumponchaindescriptors <path/to/hsm_secret> <password>\n");
```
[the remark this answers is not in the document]
niftynei: We shouldn't be doing it for other commands. This is a security hole.

tools/hsmtool.c
niftynei: One of these comments is wrong then 

>  	 *  - m/iH/0/k corresponds to the k'th keypair of the external chain of 
 	 * account number i of the HDW derived from master m. 

is not the same is.

>  We use m/0/0/k as the derivation tree for onchain funds. 

For the record, the intention here is that you'd update the code comment so other readers know where you're getting the info from, not just post a reply in github.

tools/hsmtool.c
niftynei: Update the parameter name then. As written, the parameter `testnet` suggests to say 'flag true if testnet'. 

Same thing for `private`.

niftynei:
I opened an issue for the password inputs for all of the `hsmtool` commands, #4186. I think it's ok to merge this as is. We should address this issue though, as it's a potential security hole.

rustyrussell:
You should use ccan/opt, now that cmdline parsing is non-trivial.  But can be added after....

niftynei:
Updated the interface to use "network" instead of 'testnet', rebased on master now that #4065 is merged.

darosior:
ACK 29e576459967e963ceb2dbb4de463b04dc71f479

Thanks..

PR #4173, opened 2020-11-02
chore: fixes a cppcheck false positive

m-schmoock:
Just applied the same suppression as rusty in 6635fe12e4 

My cppcheck was complaining about the same issue in the following functions.
I wonder why travis does not care though, maybe its having another version.

Also, I must admit that I can't see why this is a 'false' positive :thinking: .
Update: Figured out myself :D 

Changelog-None


niftynei (approved):
ACK 2f23791ebcec4193d6aa299729f9e4e940f7aee3

cdecker:
ACK 2f23791

PR #4187, opened 2020-11-07
optional "failure_onion" in reply to htlc_accepted hook.

fiatjaf:
I need this feature to be able to use "shadow" last hops with fake nodes on [@lntxbot](https://t.me/lntxbot) and [Etleneum](https://etleneum.com) and return proper `incorrect_or_unknown_payment_details` for the pre-pay probes with the wrong `payment_hash` that Zap, Strike and BoS wallets send. These wallets will not attempt the actual payment if the pre-pay probe fail with any other error -- and it must come from the final destination, which means that in my case I must parse `onion.next_onion` and extract the ephemeral key, compute the shared secret then encode the onion and return it in the special field `"failure_onion"` this PR introduces.

The above paragraph is just for the context.

In reality a feature like this is also necessary if we are to write plugins that interface between two different Lightning Networks or similar networks, since a bridge node must be able to forward a failure onion from one network to the other and this PR (hopefully, maybe) enables that kind of thing.

Disclaimer: I have no idea of what I'm doing, but it works and is live on [@lntxbot](https://t.me/lntxbot). So maybe someone who knows what they're doing can think of a better way to implement this.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -1053,7 +1053,25 @@ onion fields which a plugin doesn't want lightningd to consider.
 ```
 
 `fail` will tell `lightningd` to fail the HTLC with a given hex-encoded
-`failure_message` (please refer to the [spec][bolt4-failure-messages] for details: `incorrect_or_unknown_payment_details` is the most common).
+`failure_message` (please refer to the [spec][bolt4-failure-messages] for
+details: `incorrect_or_unknown_payment_details` is the most common).
+
+
+```json
+{
+  "result": "fail",
+  "failure_onion": "[serialized error packet]"
+}
+```
+
+Instead of `failure_message` the response can contain a hex-encoded
+`failure_onion` that will be used instead (please refer to the
+[spec][bolt4-failure-onion] for details). This can be used, for example,
+if you're writing a bridge between two Lightning Networks. Note that
+`lightningd` will apply the obfuscation step to the value returned here
+with its own shared secret (and key type `ammag`) before returning it to
+the previous hop.
+
 
 ```json
 {
@@ -1263,6 +1281,7 @@ The plugin must broadcast it and respond with the following fields:
 [jsonrpc-notification-spec]: https://www.jsonrpc.org/specification#notification
 [bolt4]: https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md
 [bolt4-failure-messages]: https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#failure-messages
+[bolt4-failure-onion]: https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#returning-errors
 [bolt2-open-channel]: https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#the-open_channel-message
 [sendcustommsg]: lightning-dev-sendcustommsg.7.html
 [oddok]: https://github.com/lightningnetwork/lightning-rfc/blob/master/00-introduction.md#its-ok-to-be-odd
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -888,7 +888,7 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
 	struct htlc_in *hin = request->hin;
 	struct lightningd *ld = request->ld;
 	struct preimage payment_preimage;
-	const jsmntok_t *resulttok, *paykeytok, *payloadtok, *failoniontok;
+	const jsmntok_t *resulttok, *paykeytok, *payloadtok;
 	u8 *payload, *failonion;
 
 	if (!toks || !buffer)
@@ -929,19 +929,9 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
 
 	if (json_tok_streq(buffer, resulttok, "fail")) {
 		u8 *failmsg;
-		const jsmntok_t *failmsgtok, *failcodetok;
+		const jsmntok_t *failoniontok, *failmsgtok, *failcodetok;
 
-		failmsgtok = json_get_member(buffer, toks, "failure_message");
-		if (failmsgtok) {
-			failmsg = json_tok_bin_from_hex(NULL, buffer,
-							failmsgtok);
-			if (!failmsg)
-				fatal("Bad failure_message for htlc_accepted"
-				      " hook: %.*s",
-				      failmsgtok->end - failmsgtok->start,
-				      buffer + failmsgtok->start);
-		} else if ((failoniontok = json_get_member(buffer, toks,
-							       "failure_onion"))) {
+		if ((failoniontok = json_get_member(buffer, toks, "failure_onion"))) {
 			failonion = json_tok_bin_from_hex(NULL, buffer, failoniontok);
 			if (!failonion)
 				fatal("Bad failure_onion for htlc_accepted"
@@ -949,6 +939,17 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
 				      failoniontok->end -  failoniontok->start,
 				      buffer + failoniontok->start);
 			fail_in_htlc(hin, take(new_onionreply(tmpctx, failonion)));
+			return false;
+		}
+		if ((failmsgtok = json_get_member(buffer, toks, "failure_message"))) {
+			failmsg = json_tok_bin_from_hex(NULL, buffer,
+							failmsgtok);
+			if (!failmsg)
+				fatal("Bad failure_message for htlc_accepted"
+				      " hook: %.*s",
+				      failmsgtok->end - failmsgtok->start,
+				      buffer + failmsgtok->start);
+			local_fail_in_htlc(hin, take(failmsg));
 			return false;
 		} else if (deprecated_apis
 			   && (failcodetok = json_get_member(buffer, toks,
@@ -961,10 +962,13 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
 				      - failcodetok->start,
 				      buffer + failcodetok->start);
 			failmsg = convert_failcode(NULL, ld, failcode);
-		} else
+			local_fail_in_htlc(hin, take(failmsg));
+			return false;
+		} else {
 			failmsg = towire_temporary_node_failure(NULL);
-		local_fail_in_htlc(hin, take(failmsg));
-		return false;
+			local_fail_in_htlc(hin, take(failmsg));
+			return false;
+		}
 	} else if (json_tok_streq(buffer, resulttok, "resolve")) {
 		paykeytok = json_get_member(buffer, toks, "payment_key");
 		if (!paykeytok)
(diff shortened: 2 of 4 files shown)


fiatjaf:
I don't know how to write a test for this in Python. I already spent 4 days trying to make my Go code work and I made some heavy usage of lnd's libraries. Here's the part I parse the onion and build the error reply, if anyone is interested: https://github.com/fiatjaf/lntxbot/blob/bdfd13257334ec21411280ad06c9847f7ca4b0bf/htlc_accepted.go#L79-L137

jsarenik:
Wow. This is interesting! And a nice small patch. @m-schmoock have a look please.

rustyrussell:
This is great!! All it needs is:

1. doc/PLUGINS.md so people know it exists.
2. Changelog-Added: htlc_accepted hook can now return custom failure_onion.

lightningd/peer_htlcs.c
```diff
@@ -940,6 +940,16 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
 				      " hook: %.*s",
 				      failmsgtok->end - failmsgtok->start,
 				      buffer + failmsgtok->start);
+        } else if ((failoniontok = json_get_member(buffer, toks,
```
m-schmoock, line 943: Nit: indentations are tabs in our project. Note: We use a visual tab-width of 8 spaces.

doc/PLUGINS.md
```diff
@@ -1001,7 +1001,23 @@ onion fields which a plugin doesn't want lightningd to consider.
...
+
+
+```json
+{
+  "result": "fail",
+  "failure_onion": "[292bytes of a serialized error packet]"
+}
+```
+
+If instead of `failure_message` the response contain a hex-encoded
```
m-schmoock, line 1015:
```suggestion
Instead of `failure_message` the response can contain a ...
```

m-schmoock (changes requested):
Great!
Just some nits and I would like to have a pytest blackbox test for it in e.g. `tests/test_plugin.py` where we do all of the plugin API checks.
If you need assistance ping me.

doc/PLUGINS.md
```diff
@@ -1001,7 +1001,23 @@ onion fields which a plugin doesn't want lightningd to consider.
...
+`failure_onion` that will be used instead (please refer to the
+[spec][bolt4-failure-onion] for details). Note that `lightningd` will
+apply the obfuscation step to the value retrned here with its own shared
+secret (and key type `ammag`) before returning it to the previous hop.
+
```
m-schmoock, line 1020: Maybe you could also point out the 'LN bridging nodes' usecase for this here.

lightningd/peer_htlcs.c
fiatjaf: In which circumstances are spaces used? I do see them in many lines mixed with the tabs.

m-schmoock: Ah, you can refer to `doc/STYLE.md` (Tabs and indentation=

m-schmoock: Basically its tabs with a visual length of 8 spaces. If you break up e.g. a function call to align the arguments each line then you can use spaces (after the tabs) to make them start at exactly the indentation you need.

fiatjaf: Thank you.
These are rules very hard to follow, but I like them.

doc/PLUGINS.md
```diff
@@ -1001,7 +1001,25 @@ onion fields which a plugin doesn't want lightningd to consider.
...
+  "result": "fail",
+  "failure_onion": "[292bytes of a serialized error packet]"
+}
+```
+
+Instead of `failure_message` the response can contain a hex-encoded
+`failure_onion` that will be used instead (please refer to the
+[spec][bolt4-failure-onion] for details). This can be used, for example,
+if you're writing a bridge between two Lightning Networks. Note that
+`lightningd` will apply the obfuscation step to the value retrned here
```
cdecker, line 1019:
```suggestion
`lightningd` will apply the obfuscation step to the value returned here
```

fiatjaf:
Tried to add a test and discovered that:

1. Returning an invalid onion causes `lightningd` to crash, which is more-or-less expected, but I don't know how to account for this in the tests.
2. There is a memory leak on `lightningd` upon received the `failure_onion`:

```
E           ValueError:
E           Node errors:
E           Global errors:
E            - Node /tmp/ltests-lpfkw05m/test_htlc_accepted_hook_failonion_1/lightning-2/
 has memory leaks: [
E               {
E                       "ccan/ccan/tal/tal.c:442 (tal_alloc_)",
E                       "ccan/ccan/tal/tal.c:471 (tal_alloc_arr_)",
E                       "common/json.c:244 (json_tok_bin_from_hex)",
E                       "lightningd/peer_htlcs.c:946 (htlc_accepted_hook_deserialize)",
E                       "lightningd/plugin_hook.c:184 (plugin_hook_callback)",
E                       "lightningd/plugin.c:367 (plugin_response_handle)",
E                       "lightningd/plugin.c:473 (plugin_read_json_one)",
E                       "lightningd/plugin.c:518 (plugin_read_json)",
E                       "ccan/ccan/io/io.c:417 (io_ready)",
E                       "ccan/ccan/io/poll.c:445 (io_loop)",
E                       "lightningd/io_loop_with_timers.c:24 (io_loop_with_timers)",
E                       "lightningd/lightningd.c:1015 (main)"
E                   ],
E                   "label": "common/json.c:244:u8[]",
E                   "parents": [],
E                   "value": "0x5594791ec358"
E               }
E           ]
```

I therefore give up.

lightningd/peer_htlcs.c
```diff
@@ -940,6 +940,16 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
...
+		} else if ((failoniontok = json_get_member(buffer, toks,
+							       "failure_onion"))) {
+			failonion = json_tok_bin_from_hex(NULL, buffer, failoniontok);
+			if (!failonion)
+				fatal("Bad failure_onion for htlc_accepted"
+				      " hook: %.*s",
+				      failoniontok->end -  failoniontok->start,
+				      buffer + failoniontok->start);
+			fail_in_htlc(hin, take(new_onionreply(tmpctx, failonion)));
+			return false;
```
cdecker, line 952: The only difference between `failure_onion` and `failure_msg` is that the latter gets serialized with a prefix and HMAC'd. Would it make sense to pull `local_fail_in_htlc` into this function (basically just the `create_onionreply` call) and handle both cases the same way? That'd mean that we take an eventual `failure_msg`, serialize it using the `create_onionreply` and then store it in `failure_onion`. 

Also I think that `failure_onion` needs to take precedence over `failure_msg`, since more work went into building that one. Ideally we'd panic if both are set, but since this is a hook call, our options are limited, so logging an error is likely the best we can do here.

doc/PLUGINS.md
```diff
@@ -1001,7 +1001,25 @@ onion fields which a plugin doesn't want lightningd to consider.
...
 `fail` will tell `lightningd` to fail the HTLC with a given hex-encoded
-`failure_message` (please refer to the [spec][bolt4-failure-messages] for details: `incorrect_or_unknown_payment_details` is the most common).
+`failure_message` (please refer to the [spec][bolt4-failure-messages] for
+details: `incorrect_or_unknown_payment_details` is the most common).
+
+
+```json
+{
+  "result": "fail",
+  "failure_onion": "[292bytes of a serialized error packet]"
```
cdecker, line 1011: The error onion has a variable size, so it'd be better not allude to a fixed size here by putting the `292bytes` here.

cdecker (commented):
Very nice, I had missed that `failure_msg` is not serialized and lacking the HMAC when injecting a previous onion, so this is indeed needed.

Just some minor cleanups.

lightningd/peer_htlcs.c
m-schmoock: Check the `.editorconfig` in the repos basedir. My VIM editor respects it.

doc/PLUGINS.md
fiatjaf: Does it? I was pretty sure it had a fixed size.

This proves I shouldn't be writing these things.

m-schmoock: You are for sure not more or less talented than I am :)
It's about learning not knowing

fiatjaf: Why has it a variable size? The spec says it is 260 bytes plus an HMAC of 32 bytes, then it is XORed against a cipherstream of the same size on every hop. I don't understand how it can be different than 292 bytes.

lightningd/peer_htlcs.c
fiatjaf: What about this? https://github.com/fiatjaf/lightning/blob/9af91a8493be3eb3b41d422973f2faf6209a02d2/lightningd/peer_htlcs.c#L930-L988

It's not cool, but I think it is clear. And once the deprecated branch is removed it will be much better.

niftynei:
fixed build error, addressed two outstanding review comments.

cdecker:
ACK 51c3441d27a5cf5f1cdeebe0bf51b8780789a89a

lightningd/peer_htlcs.c
```diff
@@ -929,9 +929,29 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
...
 	if (json_tok_streq(buffer, resulttok, "fail")) {
 		u8 *failmsg;
-		const jsmntok_t *failmsgtok, *failcodetok;
+		const jsmntok_t *failoniontok, *failmsgtok, *failcodetok;
 
+		failoniontok = json_get_member(buffer, toks, "failure_onion");
 		failmsgtok = json_get_member(buffer, toks, "failure_message");
+
+		if (failoniontok) {
+			failonion = json_tok_bin_from_hex(tmpctx, buffer,
```
niftynei, line 938: @fiatjaf the memory leak was caused by this line

```original
			failonion = json_tok_bin_from_hex(NULL, buffer,
```

By allocating it to `NULL` this means you're planning to manually clean it up; instead we use the magic allocation context `tmpctx`, which gets cleaned up, via magic.

lightningd/peer_htlcs.c
```diff
@@ -929,9 +929,29 @@ static bool htlc_accepted_hook_deserialize(struct htlc_accepted_hook_payload *re
...
+				      " hook: %.*s",
+				      failoniontok->end -  failoniontok->start,
+				      buffer + failoniontok->start);
+
+			if (failmsgtok)
+				log_broken(ld->log, "Both 'failure_onion' and"
+					   "'failure_message' provided."
+					   " Ignoring 'failure_message'.");
+
+			fail_in_htlc(hin, take(new_onionreply(NULL,
```
niftynei, line 951: when you `take` an object, then is a good time to allocate on `NULL`, as the `take` will reparent it to something else.

doc/PLUGINS.md
fiatjaf: Here: https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#returning-errors

It says the packet consists of `hmac(32) + failure_len(2) + failure + pad_len(2) + pad` such that `failure + pad = 256`. The `channel_update` and other things I imagine are contained in the `failure` element above.

lightningd/peer_htlcs.c
fiatjaf: I think I get it now, kinda.
Thank you.

niftynei:
ACK 6aa61e3cc3edb35deeff583ce321fdcc119716e4

m-schmoock:
Really awesome :D :+1:

PR #4188, opened 2020-11-07
chore: protocol error on htlc_minimum_msat check

m-schmoock:
`lightningd/peer_htlcs.c` `channel_added_their_htlc()`
```
	/* BOLT #2:
	 *
	 *  - receiving an `amount_msat` equal to 0, OR less than its own `htlc_minimum_msat`:
	 *    - SHOULD fail the channel.
	 */
```
By using `channel_fail_permanent()` here instead of `channel_internal_error()
we can pass the reason argument and set it to 'protocol'. Since this a
pure BOLTSPEC protocol error, I think the wrong function was called here.

Because of that I switched instead of adding a `reason` to internal
errors. They can still always reolve to `local` resons for now.

Changelog-None

### DIFF
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -1775,7 +1775,8 @@ static bool channel_added_their_htlc(struct channel *channel,
 	 */
 	if (amount_msat_eq(added->amount, AMOUNT_MSAT(0))
 	    || amount_msat_less(added->amount, channel->our_config.htlc_minimum)) {
-		channel_internal_error(channel,
+		channel_fail_permanent(channel,
+				       REASON_PROTOCOL,
 				       "trying to add HTLC amount %s"
 				       " but minimum is %s",
 				       type_to_string(tmpctx,


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4188/commits/ec7e7deb023fd3849e234110fda6d7e2c6ad5538

rustyrussell:
It's an internal error because channeld does these checks already:

see channeld/full_channel.c:542

```
	/* BOLT #2:
	 *
	 * A receiving node:
	 *  - receiving an `amount_msat` equal to 0, OR less than its own
	 *    `htlc_minimum_msat`:
	 *    - SHOULD fail the channel.
	 */
	if (amount_msat_eq(htlc->amount, AMOUNT_MSAT(0))) {
		return CHANNEL_ERR_HTLC_BELOW_MINIMUM;
	}
	if (amount_msat_less(htlc->amount, channel->config[recipient].htlc_minimum)) {
		return CHANNEL_ERR_HTLC_BELOW_MINIMUM;
	}
```

m-schmoock:
@rustyrussell  @cdecker 
I wrote a test and reproduced this in a hacky way just to know what the nodes will do. This is what I saw:

 - The receiving node force closes the channel onchain when he receives the 0msat HTLC payment, that's correct.
 - The receiving nodes `state_change` closure reason is `local` and not `protocol`, we should change that if possible.
 - The sending node sees the force close as reason `onchain`, that's correct.
 - The sending node sees the listpeers `closer` as `null`, which is what currently happens for any `onchain` closes. We should change that `onchain` closes will be blamed as `remote` closer, which is likely correct in 99.999% of the cases.

All not that bad, just edge cases nobody relies on just yet. No crashes etc.

m-schmoock:
> 4. The sending node sees the listpeers closer as null, which is what currently happens for any onchain closes. We should change that onchain closes will be blamed as remote closer, which is likely correct in 99.999% of the cases.

This is addressed by https://github.com/ElementsProject/lightning/pull/4198

Issue #4176, opened 2020-11-03
hang: lightning-cli plugin stop and then rescan


m-schmoock:
### Issue and Steps to Reproduce
1. Using a plugindir e.g. `~/.lightning/plugins` with some python plugins in it
2. Stop the python plugin by e.g. `lightning-cli plugin stop "summary.py"`
3. Rescan the plugindir by `lightning-cli plugin rescan`

This will cause the cli to hang with no result and log messages.

m-schmoock:
I looked into it a bit, this is likely caused by the fact that nobody is currently calling `was_pending(command_success(cmd, response));` within `lightningd/plugin_control.c`.

PR #4194, opened 2020-11-11
fix: hangup in plugin rescan

m-schmoock:
This adds a missing `plugins_send_getmanifest()` call in the rescan function
that lead to a RPC hangup. Not sure though if this is the proper fix.

Addresses https://github.com/ElementsProject/lightning/issues/4176

Changelog-None

### DIFF
--- a/lightningd/plugin_control.c
+++ b/lightningd/plugin_control.c
@@ -141,6 +141,7 @@ plugin_dynamic_rescan_plugins(struct command *cmd)
 	if (res)
 		return res;
 
+	plugins_send_getmanifest(cmd->ld->plugins);
 	return command_still_pending(cmd);
 }
 


darosior (approved):
ACK bb6c786ab4f76104d58c1c4f8ad4cbce7978cc1f

Good catch, it's the right fix as `plugins_add_default_dir` is only registering the plugin but not sending the getmanifest. If it did not register any, we'd have returned on line 142.

cdecker:
Just needed to make sure that we don't ask plugins for the manifest trice, but this put my mind at ease:

https://github.com/ElementsProject/lightning/blob/cefb64cd683ff4a4b5b1f30624c0a9068dc4daa5/lightningd/plugin.c#L1446-L1459

ACK https://github.com/ElementsProject/lightning/pull/4194/commits/bb6c786ab4f76104d58c1c4f8ad4cbce7978cc1f

PR #4195, opened 2020-11-11
pay: Avoid clobbering the channel_hints at all costs

cdecker:
Due to the asynchronous `getroute` and `createonion` calls we were issuing multiple concurrent `getroute` calls for multi-part payments. This is particularly painful with the `presplit` modifier which causes all of the initial parts to be scheduled at the same time, competing for a route.

Since #4093 we no longer have an asynchronous call to `getroute` which alleviates the problem, but doesn't resolve it: `createonion` is still asynchronous and we were only updating the channel_hints after it was called. In addition plugins can yield control of the `io_loop` if they perform any asynchronous RPC call, resulting in the same issue again.

This PR moves the application of the route to the channel_hints up to just before the `createonion` call, which makes the route creation and postprocessing synchronous unless a plugin performs an RPC call.

In order to address the final venue for colliding route computations we introduce a new state, that serves only to repeat the getroute call, but nothing else (theoretically a plugin could hook into it as well, but that's unlikely to be of use).

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -444,6 +444,21 @@ payment_constraints_update(struct payment_constraints *cons,
 	return true;
 }
 
+static struct channel_hint *payment_chanhints_get(struct payment *p,
+						  struct route_hop *h)
+{
+	struct payment *root = payment_root(p);
+	struct channel_hint *curhint;
+	for (size_t j = 0; j < tal_count(root->channel_hints); j++) {
+		curhint = &root->channel_hints[j];
+		if (short_channel_id_eq(&curhint->scid.scid, &h->channel_id) &&
+		    curhint->scid.dir == h->direction) {
+			return curhint;
+		}
+	}
+	return NULL;
+}
+
 /* Given a route and a couple of channel hints, apply the route to the channel
  * hints, so we have a better estimation of channel's capacity. We apply a
  * route to a channel hint before calling `sendonion` so subsequent `route`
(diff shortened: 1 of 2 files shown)


cdecker:
As far as I can see the `test_mpp_adaptive` test is the only one failing at the moment, will need to dig into that one.

m-schmoock:
@cdecker

I'm now running this on mainnet and will report if this works better in the next days...

m-schmoock:
Tried today with a 150$ payment (this is impossible with unpatched code for me on mainnet).
With this patch it worked in 5 seconds on the first attempt.

I captured stdout/info/debug incase someone needs to look...

rustyrussell:
@niftynei I think this may need consideration for rc2?

cdecker:
> @niftynei I think this may need consideration for rc2?

:+1: I'll try to fix up the failing `mpp_adaptive` test asap

cdecker:
Ok, now ready to review. Added the following changes:

 - Fixed the route applicability check (I had a parenthesis in the wrong place)
 - Split the applicability condition into multiple, documented, parts in order to avoid the above.

See [end-to-end diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/4195/range_diff/8404851010c0132a9432f9e6c0f36e516f6b721c..80ed69f5a409581581e07e60a17215535a1918bb) between the last version and current version.

plugins/libplugin-pay.c
```diff
@@ -453,57 +468,88 @@ payment_constraints_update(struct payment_constraints *cons,
...
 	struct channel_hint *curhint;
 	struct payment *root = payment_root(p);
 	assert(p->route != NULL);
+
+	/* First round: make sure we can cleanly apply the update. */
 	for (size_t i = 0; i < tal_count(p->route); i++) {
+		/* No need to check for applicability if we increase
+		 * capacity and budgets. */
+		if (remove)
+			continue;
```
m-schmoock, line 484: The function argument `remove` is never written to within this function. Thus, if `remove` is `false` it can be expected that this`if (remove) continue;` statement leads to a pointless(?) iteration through this route and it would be the same as `if (remove) break;` just a little slower. Is this intended?

m-schmoock (commented):
Today I retried this both versions (this and also the one from yesterday) on mainnet in order to get a payment from my node to my Phoenix wallet (not connected to my node, phoenix has their own channel management). I tried first `0.01btc` then `0.006btc` and then `0.003btc` (~50$). Unfortunately all of these attempts failed for both Git revisions today.

Last time, when I paid a bitrefill invoice of ~150$, it worked almost instantly (which does not work on master for me). We should test this version and the master version internally between the devs to see if one or the other is really better.

plugins/libplugin-pay.c
cdecker: I know, I just didn't want to indent the whole `for`-loop into an `if` :-)

cdecker:
> Today I retried this both versions (this and also the one from yesterday [23ab1c5](https://github.com/ElementsProject/lightning/commit/23ab1c569b16c675e19bf0be8305e8ac26a53428) ) on mainnet in order to get a payment from my node to my Phoenix wallet (not connected to my node, phoenix has their own channel management). I tried first `0.01btc` then `0.006btc` and then `0.003btc` (~50$). Unfortunately all of these attempts failed for both Git revisions today.
> 
> Last time, when I paid a bitrefill invoice of ~150$, it worked almost instantly (which does not work on master for me). We should test this version and the master version internally between the devs to see if one or the other is really better.

Kind of hard to tell where the issue is. It could be local, but it could also be in the network (min-cut across all paths from you to the destination doesn't have sufficient capacity), or it might be the recipient not having incoming capacity (though Phoenix opens channels on-demand iirc).

I think the PR itself is an improvement, since we reduce the time to completion in all cases (not attempting parts/routes that are doomed to fail). While I was initially thrilled to see that you managed to perform a payment with such a high value, I don't think that or the more recent negative result are representative or indicative of the full effect this PR might have or not.

As such I'd judge this PR more on its own merits to increase correctness and efficiency, and consider improvements to the success rate as anecdotal until we can perform really representative measurements.

Regarding representative measurements, an idea that came to mind was to create a fake-MPP receiver plugin, that'll pretend unknown HTLCs destined for the local node are part of an MPP payment and hold on to the parts for the 60 second timeout just like a real MPP recipient would, and return different errors depending of whether the MPP payment would have been successful. This way we could use random payment hashes, that are not redemable, to test MPP payments. With a number of devs running this plugin, we could perform MPP tests without actually sending any funds back and forth (always a contentious issue among friends) :wink:

plugins/libplugin-pay.c
m-schmoock: then why not `break` instead of `continue`? But not that important ;)

m-schmoock:
I made sure the payment would have fit in the channel I already have on phoenix (you can see your dynamic channel list). In any case your right we can't compare the cases, since bitrefill != phoenix and the situation channel wise might have been different suboptimal this time.

niftynei (approved):
ACK  72efbbb8d063c27f530e65fba3fdbd09efd811c2

Turned the `continue` to a `goto` for maximum readability.

PR #4198, opened 2020-11-12
fix: channel closer on unilateral to remote

m-schmoock:
Currently we set the `lightning-cli listpeers` channel `closer` to `null` if we have been offline and detected a force closed channel on us. This was because in theory it could have been ourself that used some scratching/recovery script for our private keys. Since this is topmost unlikely scenario, especially since someone doing this would certainly not fire up the daemon again and expect it to do something useful, we are better of to set the `closer` to `remote` as it should be.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -2757,6 +2757,10 @@ def test_channel_opener(node_factory):
 
     # close and check for 'closer'
     l1.rpc.close(l2.rpc.getinfo()["id"])
+    l1.daemon.wait_for_log("State changed from CHANNELD_NORMAL to")
+    l2.daemon.wait_for_log("State changed from CHANNELD_NORMAL to")
+
+    # 'closer' should now be set accordingly
     assert(l1.rpc.listpeers()['peers'][0]['channels'][0]['closer'] == 'local')
     assert(l2.rpc.listpeers()['peers'][0]['channels'][0]['closer'] == 'remote')
 
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -2744,27 +2744,6 @@ def test_htlc_retransmit_order(node_factory, executor):
     # If order was wrong, we'll get a LOG_BROKEN and fixtures will complain.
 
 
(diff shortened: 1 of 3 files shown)


jsarenik:
Sounds reasonable. What do you mean by "scratching/recovery script"? Example?

m-schmoock:
> Sounds reasonable. What do you mean by "scratching/recovery script"? Example?

Nothing that I specifically know of, but its something that I can imagine users will come up to check for unclaimed funds in an old backup or something.

m-schmoock:
The `test_channel_closer_unilateral` seems flaky, will fix that...

m-schmoock:
Hopefully the test is now stable ...

tests/test_connection.py
```diff
@@ -2757,10 +2757,36 @@ def test_channel_opener(node_factory):
...
+    # restart and wait
+    l1.restart()
+    bitcoind.generate_block(1)      # gets force close confirmed
+    l1.daemon.wait_for_log("State changed from CHANNELD_NORMAL to")
+
+    # Check for 'closer':
+    #  - l2 has closer 'local' because he forced closed it
+    #  - l1 has closer 'remote' because he just saw the onchain tx
+    assert(l1.rpc.listpeers()['peers'][0]['channels'][0]['closer'] == 'remote')
+    assert(l2.rpc.listpeers()['peers'][0]['channels'][0]['closer'] == 'local')
```
niftynei, line 2787: A faster/more CI friendly thing to do would be to find a test where we *already* do a unilateral close and add this check to that.

niftynei:
Reasonable. If someone's going to all the work to close a channel out-of-band, that definitely counts as remote. For what it's worth, we know definitively who did it in onchaind.

tests/test_connection.py
m-schmoock: Thx.
Did so and also integrated the `test_channel_opener` into the `state_change` tests...

cdecker:
ACK 47415ca14593bfcd0ef460991308b25ae0b1c674

PR #4204, opened 2020-11-17
plugins: undeprecate old form of hooks.

rustyrussell:
This effectively reverts ac93b780d5a457f8f304d658e80b4d531fb3b0e4.

Christian points out that plugins need time before we deprecate
the old options (probably 6 months) as they need to work with
both old and new.

Changelog-Deprecated: **UNDO** plugins: hooks should now be specified using objects, not raw names.
Suggested-by: @cdecker

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,6 @@ This release named by Sergi Delgado Segura.
 
 Note: You should always set `allow-deprecated-apis=false` to test for changes.
 
- - Plugins: hooks should now be specified using objects, not raw names. ([4168](https://github.com/ElementsProject/lightning/pull/4168))
  - cli: scripts should filter out '^# ' or use `-N none`, as commands will start returning notifications soon ([4046](https://github.com/ElementsProject/lightning/pull/4046))
 
 ### Removed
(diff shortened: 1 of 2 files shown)


niftynei (approved):
ACK bf69a91d225978bc081705dc457ca556af59666e

ZmnSCPxj:
*Should* we ever deprecate the "string only" form?  If we have no particular need to run this hook earlier or later than some other plugin does, additional keys are unneeded anyway.

rustyrussell:
Good q.  It's a pretty trivial API change for plugins, and generally I prefer not to keep this kind of baggage around though.

ZmnSCPxj:
This makes it difficult for plugins to support multiple versions of C-Lightning.

As a counterproposal, maybe the `getmanifest` command can give an extra parameter, `hooks-must-contain-objects` set to `true`, if the C-Lightning requires that `hooks` array contains objects instead of strings.  That way, a plugin that wants to support pre-0.9.2 C-Lightning and could not care less about hook ordering can check for the existence of this parameter and whether it is `true`.  If it does not exist or it exists but is not `true`, the plugin emits `"string"`, otherwise it emits `{"name": "string"}`.  For 0.9.2 we can add this parameter now, but set to `false` --- `doc/PLUGINS.md` says plugins should accept and ignore unrecognized parameters to `getmanifest`.

rustyrussell:
I'd prefer to have a long enough deprecation cycle that all plugins can simple change over any nobody notices?

ZmnSCPxj:
Seems okay.

Issue #4200, opened 2020-11-12
Cannot restart lightningd db since had to do fsck of hdd


d4amenace:
I recently had to run an fsck and fix damaged sectors of my drive.  Since I did this (and it was the only way to get my ubuntu running again), I am getting the following error message upon running lightningd:

Default-string:~$ lightningd
2020-11-12T23:09:10.885Z INFO    database: Creating database
Error executing statement: wallet/db.c:1023: CREATE TABLE version (version INTEGER): database disk image is malformed

What are my options?

ZmnSCPxj:
Unfortunately, if you do not have a continuous backup system like PostgreSQL cluster or the @cdecker backup plugin or ldadm/btrfs/ZFS RAID-1, you have no options.  Sorry.

If you have an old backup of the database file, you can risk replacing the database with the old one, and if your peers are nice, they will close channels on you with the latest state (if they are not nice, they can steal your channels, but that is probably better than not having any chance at all of getting any of your funds back).  Hopefully your peers support `option_static_remotekey` -- if not, you will lose your channel funds (but the *chance* of getting *some* of your funds is better than losing *all* of them IMO).

Unfortunately due to the offchain nature of Lightning, there is nobody else who has a copy of your financial data (unlike the onchain case where every archival fullnode has a copy of every user financial data).

d4amenace:
Dammit.  Are LVM snapshots any good to avoid this? Or rdiff-backup?

ZmnSCPxj:
Not sure.  Can you recover the uncorrupted db file in an LVM snapshot?

Assuming "rdiff-backup" is similar to a process doing "rsync" regularly, you would strongly prefer to use  something like the backup plugin or a PostgreSQL cluster.  Stale backups are dangerous and *could* lose your channel funds, but then not having *any* backups is a definite complete and total loss of ***all*** your channel funds (whereas if your channeled peers have `option_dataloss_protect` then a stale backup will recover your funds, but if you have channels without `option_dataloss_protect` then a stale backup will definitely lose your in-channel funds with them), so it might be better than nothing.

In any case, ***if*** you have backups, you should strongly prefer the continuous backups (PostgreSQL cluster, backup plugin on an NFS mount) rather than snapshot backups, as snapshot backups can lose your money (but no backups definitely will).

d4amenace:
> PostgreSQL cluster, backup plugin on an NFS moun

Do you think you could point me in the right directions to find out more about these solutions.  Links or web pages?

[...]

Issue #4181, opened 2020-11-05
Backup solutions 


LightningHelper:
Hi, this is just a general question.
After my precocious 1BTC lnd node got corrupted and died (although all hardware was new and bought specifically for create a node), https://github.com/lightningnetwork/lnd/issues/4720  I wonder what are the backup solutions that are available in c-lightning. I read somewhere that it is possible to do replication or something like that, but at the same time SCB is not supported so I can potentially loss all my money (#3083). Is there a place where I can read a bit about what you support? 
(I guess the corruption happen so I can keep my promise ;-) https://github.com/ElementsProject/lightning/issues/2400#issuecomment-474346091)

@rustyrussell 
Thanks 

**EDIT**
For whoever is intrested:
[Christian Decker on the latest state of LN backups](https://www.youtube.com/watch?v=kGQF3wtzr04)

ZmnSCPxj:
With a PostgreSQL backend you should be able to do DB-level replication.  I am unsure how to actually set that up, but you might ask @cdecker .  In addition, there is also the backup plugin which replicates to another file: https://github.com/lightningd/plugins/blob/master/backup/backup.py though documentation on it is very sparse.  This is intended to be the basis of a "better" backup strategy, such as using an NFS or other network mount for the backup file.

You can also use filesystem replication RAID-1 such as ZFS or btrfs; for example the Raspiblitz setup can set up a btrfs RAID-1 between a partition of your 1Tb HDD/SSD and a 32Gb flash disk.

CLDCB is not yet ready for production use yet (I see you filed an issue there), I am currently working on CLBOSS.

Generally in C-lightning we consider a continuous dynamic backup like the above backup plugin (or CLDCB when I get around to completing it, sorry) or replicating a PostgreSQL database to be superior to SCBs.  Continuous dynamic backups subsume SCBs and you do not have to remember to make an SCB for each channel you create.

MrManPew:
Thank you for the detailed information.
OffTopic
Why did you decided to use C++ for CLBOSS? isn't a scripting language (Python :)) will be easier if speed is not important?

ZmnSCPxj:
[...]

PR #4207, opened 2020-11-17
doc/BACKUP.md: Document backup strategies for `lightningd`.

ZmnSCPxj:
Requested by @d4amenace here: https://github.com/ElementsProject/lightning/issues/4200#issuecomment-727642350

@cdecker to finish up the document re: PostgreSQL.  I remember somebody made a medium article on C-Lightning PostgreSQL but I did not save the link...

Changelog-None

### DIFF
--- a/doc/BACKUP.md
+++ b/doc/BACKUP.md
@@ -1,0 +1,440 @@
+# Backing Up Your C-Lightning Node
+
+Lightning Network channels get their scalability and privacy benefits
+from the very simple technique of *not telling anyone else about your
+in-channel activity*.
+This is in contrast to onchain payments, where you have to tell everyone
+about each and every payment and have it recorded on the blockchain,
+leading to scaling problems (you have to push data to everyone, everyone
+needs to validate every transaction) and privacy problems (everyone knows
+every payment you were ever involved in).
+
+Unfortunately, this removes a property that onchain users are so used
+to, they react in surprise when learning about this removal.
+Your onchain activity is recorded in all archival fullnodes, so if you
+forget all your onchain activity because your storage got fried, you
+just go redownload the activity from the nearest archival fullnode.
+
+But in Lightning, since *you* are the only one storing all your
+financial information, you ***cannot*** recover this financial
+information from anywhere else.
+
+This means that on Lightning, **you have to** responsibly back up your
+financial information yourself, using various processes and automation.
+
+The discussion below assumes that you know where you put your
+`$LIGHTNINGDIR`, and you know the directory structure within.
+By default your `$LIGHTNINGDIR` will be in `~/.lightning/${COIN}`.
+For example, if you are running `--mainnet`, it will be
+`~/.lightning/bitcoin`.
+
+## `hsm_secret`
+
+`/!\` WHO SHOULD DO THIS: Everyone.
+
+You need a copy of the `hsm_secret` file regardless of whatever backup
+strategy you use.
+
+The `hsm_secret` is created when you first create the node, and does
+not change.
+Thus, a one-time backup of `hsm_secret` is sufficient.
+
+This is just 32 bytes, and you can do something like the below and
+write the hexadecimal digits a few times on a piece of paper:
+
+    cd $LIGHTNINGDIR
+    xxd hsm_secret
+
+You can re-enter the hexdump into a text file later and use `xxd` to
+convert it back to a binary `hsm_secret`:
+
+    cat > hsm_secret_hex.txt <<HEX
+    00: 30cc f221 94e1 7f01 cd54 d68c a1ba f124
+    10: e1f3 1d45 d904 823c 77b7 1e18 fd93 1676
+    HEX
+    xxd -r hsm_secret_hex.txt > hsm_secret
+    chmod 0400 hsm_secret
+
+Notice that you need to ensure that the `hsm_secret` is only readable by
+the user, and is not writable, as otherwise `lightningd` will refuse to
+start.
+Hence the `chmod 0400 hsm_secret` command.
+
+Alternately, if you are deploying a new node that has no funds and
+channels yet, you can generate BIP39 words using any process, and
+create the `hsm_secret` using the `hsmtool generatehsm` command.
+If you did `make install` then `hsmtool` is installed as
+`lightning-hsmtool`, else you can find it in the `tools/` directory
+of the build directory.
+
+    lightning-hsmtool generatehsm hsm_secret
+
+Then enter the BIP39 words, plus an optional passphrase.
+
+You can regenerate the same `hsm_secret` file using the same BIP39
+words, which again, you can back up on paper.
+
+Recovery of the `hsm_secret` is sufficient to recover any onchain
+funds.
+Recovery of the `hsm_secret` is necessary, but insufficient, to recover
+any in-channel funds.
+To recover in-channel funds, you need to use one or more of the other
+backup strategies below.
+
+## `backup` Plugin And Remote NFS Mount
+
+`/!\` WHO SHOULD DO THIS: Casual users.
+
+You can get the `backup` plugin here:
+https://github.com/lightningd/plugins/tree/master/backup
+
+The `backup` plugin requires Python 3.
+
+* `cd` into its directory and install requirements.
+  * `pip3 install -r requirements.txt`
+* Figure out where you will put the backup files.
+  * Ideally you have an NFS or other network-based mount on your system,
+    into which you will put the backup.
+* Stop your Lightning node.
+* `/path/to/backup-cli init ${LIGHTNINGDIR} file:///path/to/nfs/mount`.
+  This creates an initial copy of the database at the NFS mount.
+* Add these settings to your `lightningd` configuration:
+  * `important-plugin=/path/to/backup.py`
+* Restart your Lightning node.
+
+It is recommended that you use a network-mounted filesystem for the backup
+destination.
+For example, if you have a NAS you can access remotely.
+
+Do note that files are not stored encrypted, so you should really not do
+this with rented space ("cloud storage").
+
+Alternately, you *could* put it in another storage device (e.g. USB flash
+disk) in the same physical location.
+
+To recover:
+
+* Re-download the `backup` plugin and install Python 3 and the
+  requirements of `backup`.
+* `/path/to/backup-cli restore file:///path/to/nfs/mount ${LIGHTNINGDIR}`
+
+If your backup destination is a network-mounted filesystem that is in a
+remote location, then even loss of all hardware in one location will allow
+you to still recover your Lightning funds.
+
+However, if instead you are just replicating the database on another
+storage device in a single location, you remain vulnerable to disasters
+like fire or computer confiscation.
+
+## Filesystem Redundancy
+
+`/!\` WHO SHOULD DO THIS: Filesystem nerds, data hoarders, home labs,
+enterprise users.
+
+You can set up a RAID-1 with multiple storage devices, and point the
+`$LIGHTNINGDIR` to the RAID-1 setup.
+That way, failure of one storage device will still let you recover
+funds.
+
+You can use a hardware RAID-1 setup, or just buy multiple commodity
+storage media you can add to your machine and use a software RAID,
+such as (not an exhaustive list!):
+
+* `mdadm` to create a virtual volume which is the RAID combination
+  of multiple physical media.
+* BTRFS RAID-1 or RAID-10, a filesystem built into Linux.
+* ZFS RAID-Z, a filesystem that cannot be legally distributed with the Linux
+  kernel, but can be distributed in a BSD system, and can be installed 
+  on Linux with some extra effort, see
+  [ZFSonLinux](https://zfsonlinux.org).
+
+RAID-1 (whether by hardware, or software) like the above protects against
+failure of a single storage device, but does not protect you in case of
+certain disasters, such as fire or computer confiscation.
+
+You can "just" use a pair of high-quality metal-casing USB flash devices
+(you need metal-casing since the devices will have a lot of small writes,
+which will cause a lot of heating, which needs to dissipate very fast,
+otherwise the flash device firmware will internally disconnect the flash
+device from your computer, reducing your reliability) in RAID-1, if you
+have enough USB ports.
+
+### Example: BTRFS on Linux
+
+On a Linux system, one of the simpler things you can do would be to use
+BTRFS RAID-1 setup between a partition on your primary storage and a USB
+flash disk.
+The below "should" work, but assumes you are comfortable with low-level
+Linux administration.
+If you are on a system that would make you cry if you break it, you **MUST**
+stop your Lightning node and back up all files before doing the below.
+
+* Install `btrfs-progs` or `btrfs-tools` or equivalent.
+* Get a 32Gb USB flash disk.
+* Stop your Lightning node and back up everything, do not be stupid.
+* Repartition your hard disk to have a 30Gb partition.
+  * This is risky and may lose your data, so this is best done with a
+    brand-new hard disk that contains no data.
+* Connect the USB flash disk.
+* Find the `/dev/sdXX` devices for the HDD 30Gb partition and the flash disk.
+  * `lsblk -o NAME,TYPE,SIZE,MODEL` should help.
+* Create a RAID-1 `btrfs` filesystem.
+  * `mkfs.btrfs -m raid1 -d raid1 /dev/${HDD30GB} /dev/${USB32GB}`
+  * You may need to add `-f` if the USB flash disk is already formatted.
+* Create a mountpoint for the `btrfs` filesystem.
+* Create a `/etc/fstab` entry.
+  * Use the `UUID` option instad of `/dev/sdXX` since the exact device letter
+    can change across boots.
+  * You can get the UUID by `lsblk -o NAME,UUID`.
+    Specifying *either* of the devices is sufficient.
+  * Add `autodefrag` option, which tends to work better with SQLITE3
+    databases.
+  * e.g. `UUID=${UUID} ${BTRFSMOUNTPOINT} btrfs defaults,autodefrag 0 0`
+* `mount -a` then `df` to confirm it got mounted.
+* Copy the contents of the `$LIGHTNINGDIR` to the BTRFS mount point.
+  * Copy the entire directory, then `chown -R` the copy to the user who will
+    run the `lightningd`.
+  * If you are paranoid, run `diff -R` on both copies to check.
+* Remove the existing `$LIGHTNINGDIR`.
+* `ln -s ${BTRFSMOUNTPOINT}/lightningdirname ${LIGHTNINGDIR}`.
+  * Make sure the `$LIGHTNINGDIR` has the same structure as what you
+    originally had.
+* Add `crontab` entries for `root` that perform regular `btrfs` maintenance
+  tasks.
+  * `0 0 * * * /usr/bin/btrfs balance start -dusage=50 -dlimit=2 -musage=50 -mlimit=4 ${BTRFSMOUNTPOINT}`
+    This prevents BTRFS from running out of blocks even if it has unused
+    space *within* blocks, and is run at midnight everyday.
+    You may need to change the path to the `btrfs` binary.
+  * `0 0 * * 0 /usr/bin/btrfs scrub start -B -c 2 -n 4 ${BTRFSMOUNTPOINT}`
+    This detects bit rot (i.e. bad sectors) and auto-heals the filesystem,
+    and is run on Sundays at midnight.
+* Restart your Lightning node.
+
+If one or the other device fails completely, shut down your computer, boot
+on a recovery disk or similar, then:
+
+* Connect the surviving device.
+* Mount the partition/USB flash disk in `degraded` mode:
+  * `mount -o degraded /dev/sdXX /mnt/point`
+* Copy the `lightningd.sqlite3` and `hsm_secret` to new media.
+  * Do **not** write to the degraded `btrfs` mount!
+* Start up a `lightningd` using the `hsm_secret` and `lightningd.sqlite3`
+  and close all channels and move all funds to onchain cold storage you
+  control, then set up a new Lightning node.
+
+If the device that fails is the USB flash disk, you can replace it using
+BTRFS commands.
+You should probably stop your Lightning node while doing this.
+
+* `btrfs replace start /dev/sdOLD /dev/sdNEW ${BTRFSMOUNTPOINT}`.
+  * If `/dev/sdOLD` no longer even exists because the device is really
+    really broken, use `btrfs filesystem show` to see the number after
+    `devid` of the broken device, and use that number instead of
+    `/dev/sdOLD`.
+* Monitor status with `btrfs replace status ${BTRFSMOUNTPOINT}`.
+
+More sophisticated setups with more than two devices are possible.
+Take note that "RAID 1" in `btrfs` means "data is copied on up to two
+devices", meaning only up to one device can fail.
+You may be interested in `raid1c3` and `raid1c4` modes if you have
+three or four storage devices.
+BTRFS would probably work better if you were purchasing an entire set
+of new storage devices to set up a new node.
+
+## PostgreSQL Cluster
+
+`/!\` WHO SHOULD DO THIS: Enterprise users, whales.
+
+`lightningd` may also be compiled with PostgreSQL support.
+PostgreSQL is generally faster than SQLITE3, and also supports running a
+PostgreSQL cluster to be used by `lightningd`, with automatic replication
+and failover in case an entire node of the PostgreSQL cluster fails.
+
+Setting this up, however, is more involved.
+
+By default, `lightningd` compiles with PostgreSQL support **only** if it
+finds `libpq` installed when you `./configure`.
+To enable it, you have to install a developer version of `libpq`.
+On most Debian-derived systems that would be `libpq-dev`.
+To verify you have it properly installed on your system, check if the
+following command gives you a path:
+
+    pg_config --includedir
+
+Versioning may also matter to you.
+For example, Debian Stable ("buster") as of late 2020 provides PostgreSQL 11.9
+for the `libpq-dev` package, but Ubuntu LTS ("focal") of 2020 provides
+PostgreSQL 12.5.
+Debian Testing ("bullseye") uses PostgreSQL 13.0 as of this writing.
+PostgreSQL 12 had a non-trivial change in the way the restore operation is
+done for replication.
+You should use the same PostgreSQL version of `libpq-dev` as what you run
+on your cluster, which probably means running the same distribution on
+your cluster.
+
+Once you have decided on a specific version you will use throughout, refer
+as well to the "synchronous replication" document of PostgreSQL for the
+**specific version** you are using:
+
+* [PostgreSQL 11](https://www.postgresql.org/docs/11/runtime-config-replication.html)
+* [PostgreSQL 12](https://www.postgresql.org/docs/12/runtime-config-replication.html)
+* [PostgreSQL 13](https://www.postgresql.org/docs/13/runtime-config-replication.html)
+
+You then have to compile `lightningd` with PostgreSQL support.
+
+* Clone or untar a new source tree for `lightning` and `cd` into it.
+  * You *could* just use `make clean` on an existing one, but for the
+    avoidance of doubt (and potential bugs in our `Makefile` cleanup rules),
+    just create a fresh source tree.
+* `./configure`
+  * Add any options to `configure` that you normally use as well.
+* Double-check the `config.vars` file contains `HAVE_POSTGRES=1`.
+  * `grep 'HAVE_POSTGRES' config.vars`
+* `make`
+* If you install `lightningd`, `sudo make install`.
+
+If you were not using PostgreSQL before but have compiled and used
+`lightningd` on your system, the resulting `lightningd` will still
+continue supporting and using your current SQLITE3 database;
+it just gains the option to use a PostgreSQL database as well.
+
+If you just want to use PostgreSQL without using a cluster (for
+example, as an initial test without risking any significant funds),
+then after setting up a PostgreSQL database, you just need to add
+`--wallet=postgresql://${USER}:${PASSWORD}@${HOST}:${PORT}/${DB}`
+to your `lightningd` config or invocation.
+
+To set up a cluster for a brand new node, follow this (external)
+[guide by @gabridome][gabridomeguide].
+
+[gabridomeguide]: https://github.com/gabridome/docs/blob/master/c-lightning_with_postgresql_reliability.md
+
+The above guide assumes you are setting up a new node from scratch.
+It is also specific to PostgreSQL 12, and setting up for other versions
+**will** have differences; read the PostgreSQL manuals linked above.
+
+If you want to continue a node that started using an SQLITE3 database,
+note that we do not support this.
+You should set up a new PostgreSQL node, move funds from the SQLITE3
+node to the PostgreSQL node, then shut dwon the SQLITE3 node
+permanently.
+
+There are also more ways to set up PostgreSQL replication.
+In general, you should use [synchronous replication (13)][pqsyncreplication],
+since `lightningd` assumes that once a transaction is committed, it is
+saved in all permanent storage.
+This can be difficult to create remote replicas due to the latency.
+
+[pqsyncreplication]: https://www.postgresql.org/docs/13/warm-standby.html#SYNCHRONOUS-REPLICATION
+
+## Database File Backups
+
+`/!\` WHO SHOULD DO THIS: Those who already have at least one of the
+other backup methods, those who are #reckless.
+
+This is the least desirable backup strategy, as it *can* lead to loss
+of all in-channel funds if you use it.
+However, having *no* backup strategy at all *will* lead to loss of all
+in-channel funds, so this is still better than nothing.
+
+This backup method is undesirable, since it cannot recover the following
+channels:
+
+* Channels with peers that do not support `option_dataloss_protect`.
+  * Most nodes on the network already support `option_dataloss_protect`
+    as of November 2020.
+  * If the peer does not support `option_dataloss_protect`, then the entire
+    channel funds will be revoked by the peer.
+  * Peers can *claim* to honestly support this, but later steal funds
+    from you by giving obsolete state when you recover.
+* Channels created *after* the copy was made are not recoverable.
+  * Data for those channels does not exist in the backup, so your node
+    cannot recover them.
+
+Because of the above, this strategy is discouraged: you *can* potentially
+lose all funds in open channels.
+
+However, again, note that a "no backups #reckless" strategy leads to
+*definite* loss of funds, so you should still prefer *this* strategy rather
+than having *no* backups at all.
+
+Even if you have one of the better options above, you might still want to do
+this as a worst-case fallback, as long as you:
+
+* Attempt to recover using the other backup options below first.
+  Any one of them will be better than this backup option.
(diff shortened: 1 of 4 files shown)


jsarenik:
Let me link also https://github.com/ElementsProject/lightning/issues/4181 for reference.

Thank you @ZmnSCPxj for working on it!

ZmnSCPxj:
Clarified location of `$LIGHTNINGDIR`.

ZmnSCPxj:
Corrected misspelling of `lighningd`.

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+
+While `lightningd` is not running, just copy the `lightningd.sqlite3` file
+in the `$LIGHTNINGDIR` on backup media somewhere.
+
+To recover, just copy the backed up `lightningd.sqlite3` into your new
+`$LIGHTNINGDIR` together with the `hsm_secret`.
+
+You can also use any automated backup system as long as it includes the
+`lightningd.sqlite3` file (and optionally `hsm_secret`, but note that
+as a secret key, hackers getting a copy of your backups may allow them
```
jsarenik, line 94: hackers? Not sure we should jump on the train calling thieves hackers.

http://www.catb.org/jargon/html/index.html

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+* Channels created *after* the copy was made are not recoverable.
+  * Data for those channels does not exist in the backup, so your node
+    cannot recover them.
+
+Because of the above, this strategy is discouraged: you *can* potentially
+lose all funds in open channels.
+
+However, again, note that a "no backups #reckless" strategy leads to
+*definite* loss of funds, so you should still prefer this strategy rather
+than not have **any** backups at all.
```
jsarenik, line 121: Emphasis on word **any** is a bit misleading. There is no comparision (with other emphasized word nearby). I would prefer something like "so you should still prefer this strategy rather than having no backups at all."

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+
+Even if you have the better options below, you might still want to do
+this as a fallback, as long as you ensure:
```
jsarenik, line 124: as long as you: (and see below)

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+
+* Attempt to recover using the other backup options below first.
```
jsarenik, line 126: This is good as it is. Actually it does not go well when there is the "ensure" word above.

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+  Any one of them will be better than this backup option.
+* Recovering by this method ***MUST*** always be the ***last*** resort.
```
jsarenik, line 128: Change to something like "Use this method **ONLY** as a ***last*** resort" (goes well with "as long as you:" above)

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+
+More sophisticated setups with more than two devices are possible.
+Take note that "RAID 1" in `btrfs` means "data is copied on up to two
+devices", meaning only up to one device can fail.
+You may be interested in `raid1c3` and `raid1c4` modes if you have
+three or four storage devices.
+BTRFS would probably work better if you were purchasing an entire set
+of new storage devices to set up a new node.
+
+On BSD you can use a ZFS RAID-Z setup, which is probably better than BTRFS
```
jsarenik, line 246: The general advice (like first paragraph in this section also) could be moved together, BTRFS could be turned into its own sub-section so that it is optically skippable (sp?).

BTW this review comes from a point of view of a non-native speaker. So take it with a grain of salt.

doc/BACKUP.md
```diff
@@ -0,0 +1,312 @@
...
+* Re-download the `backup` plugin and install Python 3 and the
+  requirements of `backup`.
+* `/path/to/backup-cli restore file:///path/to/nfs/mount ${LIGHTNINGDIR}`
+
+If your backup destination is a network-mounted filesystem that is in a
+remote location, then even loss of all hardware in one location will allow
+you to still recover your Lightning funds.
+
+However, if instead you are just replicating the database on another
+storage device on a single location, you remain vulnerable to disasters
```
jsarenik, line 299: in a single location?

jsarenik (changes requested):
Wonderful documentation (except the BTRFS ad :-P)
Thank you!

doc/BACKUP.md
ZmnSCPxj: Right, "in" is better than "on" here.

doc/BACKUP.md
ZmnSCPxj: "you should still prefer **this** strategy rather than have **no** backups at all." reads better?

jsarenik: Sure. That works well.

doc/BACKUP.md
ZmnSCPxj: Okay.  I guess we can list software methods of RAID-1, such as:

* `mdadm`
* BTRFS
* ZFS

ZmnSCPxj:
Updated as per @jsarenik feedback.

doc/BACKUP.md
ZmnSCPxj: Went with "Recover by this method **ONLY** as a ***last*** resort"

cdecker:
I think you're referring to @gabridome's excellent tutorial here: https://github.com/gabridome/docs/blob/master/c-lightning_with_postgresql_reliability.md

ZmnSCPxj:
Add `autodefrag` to BTRFS instructions, minor wording tweak, expand "PostgreSQL" section.

darosior:
Would be worse to link it from https://lightning.readthedocs.io/FAQ.html#how-to-backup-my-wallet i think

ZmnSCPxj:
Mention the SQLITE3 backup API, add link from FAQ.md to BACKUP.md

ZmnSCPxj:
Mention using the `devid` in BTRFS.

doc/BACKUP.md
```diff
@@ -0,0 +1,391 @@
...
+
+Instead, the **proper** way to do this would be:
+
+* `echo "VACUUM INTO 'backup.sqlite3';" | sqlite3 lightningd.sqlite3`
+* `sync`
+* `mv backup.sqlite3 ${YOURBACKUPLOCATION}/lightningd.sqlite3`
+
+This creates a consistent snapshot of the database, sampled in a
+transaction, that is assured to be openable later by `sqlite3`.
+The operation of the `lightningd` process will be paused while `sqlite3`
```
ZmnSCPxj, line 152: ***THIS IS FACTUALLY INCORRECT***.

I did a stress test where a program continuously makes many small transactions to update a database.  When you do something like `.backup 'backup.sqlite3'` or `VACUUM INTO `backup.sqlite3';` in a *separate* `sqlite3`, then either the main program gets an `SQLITE_BUSY`, ***or*** the separate `sqlite3` returns a "Database is locked" error.

I suspect that an `SQLITE_BUSY` will cause `lightningd` to crash, so this is probably strongly not recommended to use a separate process to back up.

We can do:

* Expose an `sqlite3_backup` command which performs the `VACUUM INTO` query inside `lightningd`.
* Call into `sqlite3_busy_timeout` with a "reasonable" large value, say 5000 (5 seconds). (Equivalently, call `PRAGMA busy_timeout = 5000;` query at the same time we do `PRAGMA foreign_keys=on;`).  Then, even if the `.backup` or `VACUUM INTO` takes up to 5 seconds to back up, `lightningd` will just wait.

The latter solution is easier but there is always the possibility that the backup process will take more than whatever timeout we select.  The earlier solution is more reliable and there is no timeout, but note that if say the target location is slow (e.g. an NFS mount) then `lightningd` will suspend indefinitely.

Thoughts?

ZmnSCPxj:
Remove the factually incorrect text mentioned in https://github.com/ElementsProject/lightning/pull/4207#discussion_r528206037

ZmnSCPxj:
Teach how to quickly check if a backup database is corrupted.

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+any in-channel funds.
+To recover in-channel funds, you need to use one or more of the other
+backup strategies below.
+
+Database File Backups
+---------------------
+
+This is the least desirable backup strategy, as it *can* lead to loss
+of all in-channel funds if you use it.
+However, having *no* backup strategy at all *will* lead to loss of all
```
MrManPew, line 83: Not sure it is true, as eventually the other user will force-close the channels and you will get the BTC back to your address? right?

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+    well-documented open-source code.
+* Channels created *after* the copy was made are not recoverable.
+  * Data for those channels does not exist in the backup, so your node
+    cannot recover them.
+
+Because of the above, this strategy is discouraged: you *can* potentially
+lose all funds in open channels.
+
+However, again, note that a "no backups #reckless" strategy leads to
+*definite* loss of funds, so you should still prefer *this* strategy rather
```
MrManPew, line 120: again, what if the other user close the channels? wouldn't you get the money back to your wallet?

MrManPew (commented):
Good info

doc/BACKUP.md
ZmnSCPxj: If the peer does not have `option_static_remotekey` then the address it goes to is not on the path derived from `hsm_secret`, but requires the derivation keys stored in the db.  Though lots of nodes today already support `option_static_remotekey`.

ZmnSCPxj:
Okay, we can actually implement an sqlite3-backup-snapshot in a plugin, I think.

Basically, `lightningd` performs transactions this way:

* `BEGIN;` to the sqlite3 connection.
* Send queries to the connection.  Write queries are stored in a log maintained by the `db_write` hook.
* Just before committing, trigger the `db_write` hook.
  * If it returns something other than `{"result":"continue"}`, exit the entire program, which implicitly rolls back the transaction.
* `COMMIT;` to the sqlite3 connection.

If so, a plugin that hooks into `db_write` can delay the `COMMIT;` query by the simple act of not responding immediately.  And the write(s) to the database file are triggered by `COMMIT;`; without the `COMMIT;` everything is in the `-journal` file.

So it looks to me that we can have a C plugin which hooks into `db_write`, and the `db_write` handler just responds with `{"result":"continue"}`, and offers a `sqlite3-snapshot` command.  The handler of the `sqlite3-snapshot` just uses synchronous code (which implicitly causes any `db_write` hook to block) to copy the `lightningd.sqlite3` file elsewhere.  After it does the copy it responds and unblocks to the mainloop, letting any pending `db_write` through (it could "simply" unblock after the copy and do `VACUUM` on the copy, possibly also deleting entries in non-vital tables like invoices, and non-pending payments to reduce the size of the snapshot).  The plugin can check if the `lightningd` is on `sqlite3` by checking `wallet` in `listconfigs`  I probably need to write some stress-test experimental code to simulate this.

In WAL mode we need to copy the `-wal` file as well.  It is possible to set your `lightningd.sqlite3` to WAL mode by stopping `lightningd`, then `sqlite3 lightningd.sqlite3` and issue `PRAGMA journal_mode=WAL`, then `.exit` the `sqlite3` and restarting `lightningd`.  You can revert with `PRAGMA journal_mode=DELETE;`

doc/BACKUP.md
ZmnSCPxj: You also need to wait for the other side to actually force close, and that could take a while.

cdecker: No, if a channel was opened without the static key option the payout address is dependent on the last state (address is tweaked with the channel state), and the wallet will not be able to recognize the address as its own.

Worse: relying on the peer to do the right thing might expose you to fraud if they decide they prefer an older state where they had a bigger share of the channel capacity.

doc/BACKUP.md
jsarenik: Thank you @ZmnSCPxj for ideas and investigating possible issues with running backup in separate `sqlite3` process!

Exposing the `sqlite3_backup` command would be great! It sounds like a very nice and clear way to do backup.

ZmnSCPxj:
> Okay, we can actually implement an sqlite3-backup-snapshot in a plugin, I think.

Not actually, because the `doc/PLUGINS.md` specifically says that a plugin that hooks into `db_write` should not handle any other hooks or commands.  Since we have this in our doc, we should really not violate it ourselves, as otherwise future contributors might provide code that assumes no decent plugin will handle `db_write` and another command.  So it would still be best to either add a decent busy timeout to our db connection so a third-party can snapshot the file in a separate `sqlite3` connection, or implement something inside the `lightningd` process to do the snapshot.

d4amenace:
> > ....... or implement something inside the `lightningd` process to do the snapshot.

This actually made sense to me.  There could be a configurable timeout period in the conf file to denote the period of snapshots.  At each interval writing would be paused for the db so that corrupted snapshots would be avoided.  There would be a standardized snapshot folder in the users .lighting folder which would be the most all-encompassing solutions of all the implementations.

ZmnSCPxj:
> There could be a configurable timeout period in the conf file to denote the period of snapshots.

All that `lightningd` has to expose at the most basic level would be the ability to make *one* snapshot at the current time.  Regular snapshots like what you want can be done by a separate program, xref. `autocleaninvoices` which has the scheduled invoice cleaning in a plugin, not in `lightningd` (`lightningd` just exposes a command to clean old invoices using a single db query, the auto-cleaning is done by the plugin).

In particular, such a regular snapshotting would be ***vastly*** inferior to `db_write`:

* During times when there are lots of updates to your channel state, your regular snapshotting would miss them.  If the node crashes at that point, you risk losing funds.
* During times when there is nothing happening, your regular snapshotting will copy files over and over, wasting space and time on substantially the same data.
  * If you want something "smart" like say "oh just save the changes between this snapshot and the previous one", ***that is what `db_write` already is***, *plus* it triggers at each update, so just use `db_write`, via the existing `backup.py` plugin, already.

Regular snapshotting of the db say once or twice a day would be a good "backup of a backup" --- your primary backup should still be a `db_write` plugin OR a PostgreSQL cluster OR at the absolute bare minimum some kind of RAID filesystem setup, and preferably all of them.  Recovering from a snapshot would be the sort of thing you would do as a last resort, only-use-this-if-Godzilla-attacks recovery in case all your *other* backups are broken.  And you would want that regular snapshotting to, say, be uploaded somewhere *else*, like encrypted on some cloud server or at least on a computer you own located in the house of a friend.  Since that probably involves some kind of custom arrangement, it would not be easy to standardize such an interface inside `lightningd`; so a single-shot snapshot would be enough in C-Lightning, I think, and you build your custom arrangement on top with `crontab`.

gabridome:
> Requested by @d4amenace here: [#4200 (comment)](https://github.com/ElementsProject/lightning/issues/4200#issuecomment-727642350)
> 
> @cdecker to finish up the document re: PostgreSQL. I remember somebody made a medium article on C-Lightning PostgreSQL but I did not save the link...

Sorry for the late reply. https://github.com/gabridome/docs/blob/master/c-lightning_with_postgresql_reliability.md

Available for any question, comment or help.

ZmnSCPxj:
@gabridome does the PostgreSQL section I made make sense?

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+Setting this up, however, is more involved.
+
+By default, `lightningd` compiles without PostgreSQL support.
+To enable it, you have to install a developer version of `libpq`.
+On most Debian-derived systems that would be `libpq-dev`.
+To verify you have it properly installed on your system, check if the
+following command gives you a path:
+
+    pg_config --includedir
+
```
gabridome, line 346: I didn't know it. When I built it, everything worked out of the box. Pretty useful to know. 
I probably had already the library installed.

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+For example, Debian Stable ("buster") as of late 2020 provides PostgreSQL 11.9
+for the `libpq-dev` package, but Ubuntu LTS ("focal") of 2020 provides
+PostgreSQL 12.5.
+Debian Testing ("bullseye") uses PostgreSQL 13.0 as of this writing.
+PostgreSQL 12 had a non-trivial change in the way the restore operation is
+done for replication.
+You should use the same PostgreSQL version of `libpq-dev` as what you run
+on your cluster, which probably means running the same distribution on
+your cluster.
+
```
gabridome, line 357: Personally I have added on the cluster machines the Postgres .deb repository in this way I can keep the versions aligned.
I didn't encounter any problem but maybe it is not wise to exit from the Postgres version installed with the Debian distribution. What you guys think?

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+If you want to continue a node that started using an SQLITE3 database,
+use the [tool by @fiatjaf][fiatjafmcldsp] to convert your SQLITE3 database
+to a PostgreSQL database.
+Follow up to just *before* "Connect to the database" on the @gabridome
+guide, then use the @fiatjaf `mcldsp` tool with the `postgres` specification
+you generated, then continue with the rest of the @gabridome guide
+(though you should probably do some more double-checking and tire-kicking
+in the "Connect to the database" stage you resume at, such as checking if
+`listpeers` still lists the same channels as you had, and so on).
+
```
gabridome, line 391: Wise suggestion IMO.

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+    pg_config --includedir
+
+Versioning may also matter to you.
+For example, Debian Stable ("buster") as of late 2020 provides PostgreSQL 11.9
+for the `libpq-dev` package, but Ubuntu LTS ("focal") of 2020 provides
+PostgreSQL 12.5.
+Debian Testing ("bullseye") uses PostgreSQL 13.0 as of this writing.
+PostgreSQL 12 had a non-trivial change in the way the restore operation is
+done for replication.
+You should use the same PostgreSQL version of `libpq-dev` as what you run
```
gabridome, line 354: I encountered the problem myself. I would add to always read the official synchronous replication PostgresQL guide for the **specific version** you are using.

doc/BACKUP.md
```diff
@@ -0,0 +1,400 @@
...
+If you were not using PostgreSQL before but have compiled and used
+`lightningd` on your system, the resulting `lightningd` will still
+continue supporting and using your current SQLITE3 database;
+it just gains the option to use a PostgreSQL database as well.
+
+To set up a cluster for a brand new node, follow this (external)
+[guide by @gabridome][gabridomeguide].
+
+[gabridomeguide]: https://github.com/gabridome/docs/blob/master/c-lightning_with_postgresql_reliability.md
+
```
gabridome, line 380: The guide is not maintained as the versions of Postgres evolve. Please always check also the Postgres Guide about synchronous replication for your specific version.

gabridome (changes requested):

gabridome:
> @gabridome does the PostgreSQL section I made make sense?

Absolutely. Thank you for mentioning my guide.

doc/BACKUP.md
ZmnSCPxj: Okay, maybe the Lightning+PostgreSQL can be a separate PR?

doc/BACKUP.md
ZmnSCPxj: I would have used the same distro throughout just to reduce inter-version problems, which is why I suggested the above.  Mixing deb repositories has gotten me into trouble before when I wanted to upgrade my OS (but upgrading an OS is always fraught, sometimes it is best to just have two small OS partitions and alternate installing OS's between them rather than upgrading an existing install...)

doc/BACKUP.md
ZmnSCPxj: Okay, I will add that.

doc/BACKUP.md
gabridome: The document you mean?
Something like `doc/postgresQL.md`?

Anyway This part in the backup strategies seems pretty right to me. At least to mention a more extended part in the doc directory.  @fiatjaf Has thought about that?

ZmnSCPxj: Yes, the document I mean.

ZmnSCPxj:
Rebased, emphasized checking your PostgreSQL version, explicated more about tire-kicking after SQLITE3->PG conversion.

ZmnSCPxj:
No ACKs?  People seem to want this.  Is there any issue or quibble that prevents this from being merged?

gabridome:
> No ACKs? People seem to want this. Is there any issue or quibble that prevents this from being merged?

FWIW conceptACK.

I plaude loudly this incredible work on the most important topic of the LN that has been delayed for too long.

I frankly hope that these strategies will become more and more user friendly as the time passes. Until then, the paradox is that the ones less aware of the problem, are the one more in need of a good solution...

gabridome:
Hi,
there's a new possibility for the backup of `hsm_secret` file.
I find it useful because it converts to xpriv format, so you can import into one descriptor based wallet in Bitcoin Core or wherever xpriv are accepted.
https://github.com/domegabri/lightning-secret

ZmnSCPxj:
@gabridome how do we recover the `hsm_secret` from the xprv?  The program seems to only convert from `hsm_secret` to xprv.

gabridome:
Mmh... You're right. Gotta check.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
+Backing Up Your C-Lightning Node
+================================
+
+Lightning Network channels get their scalability and privacy benefits
+from the very simple technique of *not telling anyone else about your
+in-channel activity*.
+This is in contrast to onchain payments, where you have to tell everyone
+about each and every payment and have it recorded on the blockchain,
+leading to scaling problems (you have to push data at everyone) and
```
cdecker, line 9:
```suggestion
leading to scaling problems (you have to push data to everyone, everyone needs to verify every transaction) and
```

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+
+Unfortunately, this removes a property that onchain users are so used
+to, they react in surprise when learning about this removal.
+Your onchain activity is recorded in all archival fullnodes, so if you
+forget all your onchain activity because your storage got fried, you
+just go redownload the activity from the nearest archival fullnode.
+
+But in Lightning, since *you* are the only one storing all your
+financial information, you ***cannot*** recover this financial
+information anywhere else.
```
cdecker, line 20:
```suggestion
information from anywhere else.
```

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+
+This means that on Lightning, you **have to** responsibly back up your
```
cdecker, line 22:
```suggestion
This means that on Lightning, **you have to** responsibly back up your
```
:wink:

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+For example, if you are running `--mainnet`, it will be
+`~/.lightning/bitcoin`.
+
+`hsm_secret`
+------------
+
+You need a copy of the `hsm_secret` file regardless of whatever backup
+strategy you use.
+This is just 32 bytes, and you can do something like the below and
+write the hexadecimal digits a few times on a piece of paper:
```
cdecker, line 37: Maybe mention that this file does not change after its creation, so backing it up when first setting up the node is sufficient and no ongoing backup is required, in contrast to the DB which requires continuous backups.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+
+Recovery of the `hsm_secret` is sufficient to recover any onchain
+funds.
+Recovery of the `hsm_secret` is necessary, but insufficient, to recover
+any in-channel funds.
+To recover in-channel funds, you need to use one or more of the other
+backup strategies below.
+
+Database File Backups
+---------------------
```
cdecker, line 79: IMHO the methods should be sorted in descending order of desirability, putting this type first gives absolutely the wrong impression.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+to steal your funds, even in-channel funds).
+
+This backup method is undesirable, since it cannot recover the following
+channels:
+
+* Channels with peers that do not support `option_dataloss_protect`.
+  * Most nodes on the network already support `option_dataloss_protect`
+    as of November 2020.
+  * If the peer does not support this option, then the entire channel
+    funds will be revoked by the peer.
```
cdecker, line 104: Technically not true: the peer may be well-behaved, and receive back his share of the funds, but the user's funds will be lost forever into an unrecoverable address, due to the tweak.

Even if the peer is not well-behaved, it can at most grab the funds that belonged to it some time in the past, not all funds. The peer would then go back to that point in time, grab it's now bigger share, and still sinkhole the user's share.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+  * Specially-coded peers can pretend to support this option, but end
+    up taking the entire channel funds (or otherwise more than what it
+    should have) when you try to recover.
+    Standard releases of Lightning Network node software do not do this
+    (they just support this option according to the standard) but note
+    that anyone can modify their node software, especially with our
+    well-documented open-source code.
```
cdecker, line 111: Doesn't change the behavior above: the peer can prevent the user from recovering its share of the funds by not sharing the commitment point.

cdecker: Generally speaking I'd reduce the level of detail here slightly, to "if the peer is non-cooperative you won't be able to recover your funds".

doc/BACKUP.md
cdecker: Just to add a bit more complexity: if we use `option_upfront_shutdown_script` we may actually get the funds back to our wallet if the peer closes, because that removes the tweak to the their_unilateral/to_us transaction.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+financial information yourself, using various processes and automation.
+
+The discussion below assumes that you know where you put your
+`$LIGHTNINGDIR`, and you know the directory structure within.
+By default your `$LIGHTNINGDIR` will be in `~/.lightning/${COIN}`.
+For example, if you are running `--mainnet`, it will be
+`~/.lightning/bitcoin`.
+
+`hsm_secret`
+------------
```
cdecker, line 32: Mix of `----` and `###` headings makes it more difficult to judge the overall structure. Please use `###` prefixes wherever possible.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+  Any one of them will be better than this backup option.
+* Recover by this method **ONLY** as a ***last*** resort.
+* Recover using the most recent backup you can find.
+  Take time to look for the most recent available backup.
+
+Again, this strategy can lead to only ***partial*** recovery of funds,
+or even to complete failure to recover, so use the other methods below
+first to recover!
+
+### Backing Up While `lightningd` Is Running
```
cdecker, line 136: Ok, while the above should be the last option presented, this one should be even laster! :wink:

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+storage media you can add to your machine and use a software RAID,
+such as (not an exhaustive list!):
+
+* `mdadm` to create a virtual volume which is the RAID combination
+  of multiple physical media.
+* BTRFS RAID-1 or RAID-10, a filesystem built into Linux.
+* ZFS RAID-Z, a filesystem that cannot be legally distributed with the Linux
+  kernel, but can be distributed in a BSD system, and can be installed 
+  on Linux with some extra effort, see
+  [ZFSonLinux](https:://zfsonlinux.org).
```
cdecker, line 196:
```suggestion
  [ZFSonLinux](https://zfsonlinux.org).
```

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+  * `pip3 install -r requirements.txt`
+* Figure out where you will put the backup files.
+  * Ideally you have an NFS or other network-based mount on your system,
+    into which you will put the backup.
+* Stop your Lightning node.
+* `/path/to/backup-cli init ${LIGHTNINGDIR} file:///path/to/nfs/mount`.
+  This creates an initial copy of the database at the NFS mount.
+* Add these settings to your `lightningd` configuration:
+  * `important-plugin=/path/to/backup.py`
+  * `backup-destination=file:///path/to/nfs/mount`
```
cdecker, line 302: This is no longer needed, because we now write a `lock` file into the `$LIGHTNINGDIR` in order to be functional right-away, and avoid accidentally changing the backup location.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+storage device in a single location, you remain vulnerable to disasters
+like fire or computer confiscation.
+
+PostgreSQL Cluster
+------------------
+
+`lightningd` may also be compiled with PostgreSQL support.
+PostgreSQL is generally faster than SQLITE3, and also supports running a
+PostgreSQL cluster to be used by `lightningd`, with automatic replication
+and turnover in case an entire node of the PostgreSQL cluster fails.
```
cdecker, line 335:
```suggestion
and failover in case an entire node of the PostgreSQL cluster fails.
```

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+
+Setting this up, however, is more involved.
+
+By default, `lightningd` compiles without PostgreSQL support.
```
cdecker, line 339: We compile with postgresql by default if `./configure` detect `libpq`, and our releases are configured to ship with postgres enabled.

doc/BACKUP.md
cdecker: Agreed, there is little point in replicating the official doc here, it may just end up with stale information, requiring updates from us. If there is an authoritative source, link to it.

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+it just gains the option to use a PostgreSQL database as well.
+
+To set up a cluster for a brand new node, follow this (external)
+[guide by @gabridome][gabridomeguide].
+
+[gabridomeguide]: https://github.com/gabridome/docs/blob/master/c-lightning_with_postgresql_reliability.md
+
+The above guide assumes you are setting up a new node from scratch.
+It is also specific to PostgreSQL 12, and setting up for other versions
+**will** have differences; read the PostgreSQL manuals linked above.
```
cdecker, line 391: I'm somehow missing the `--wallet-dsn=postgres://...` pointer to switch over to the postgres DB.

cdecker (commented):
Sorry for the delay reviewing this. It looks quite good, but reordering might be good, to emphasize which mechanisms are preferred, and for whom it might be worth it:

 - Backup `hsm_secret`: static, for all users
 - Backup plugins for end-users
 - Postgresql replication for enterprises
 - File-backup: if no other option, and with appropriate warnings
   - Backup while offline
   - Hot backups while the node is running

svewa:
Seeing the discussion on how to backup the sqlite file, would it not be best to not filesystem-copy it, but sqlite3 .dump it? https://www.sqlitetutorial.net/sqlite-dump/ or would this not guarantee consistency?

doc/BACKUP.md
```diff
@@ -0,0 +1,413 @@
...
+
+If you want to continue a node that started using an SQLITE3 database,
+use the [tool by @fiatjaf][fiatjafmcldsp] to convert your SQLITE3 database
+to a PostgreSQL database.
```
darosior, line 395: Hmmm .. Are you confident with it ? Even @fiatjaf [does not seem to be](https://github.com/fiatjaf/mcldsp#beware-this-may-destroy-your-life-and-make-you-lose-all-your-money-use-at-your-own-risk).  If we are confident we are able to provide such a **crucial** tool it shoudl probably be part of the main repo...

darosior (commented):
Really nice! I now have a reference to point people to instead of handwaving the different possibilities each time i'm asked :)

Just a small comment re the db migration tool which i'm not sure is safe.

darosior:
> Hi,
there's a new possibility for the backup of hsm_secret file.
I find it useful because it converts to xpriv format, so you can import into one descriptor based wallet in Bitcoin Core or wherever xpriv are accepted.
https://github.com/domegabri/lightning-secret

FWIW, it was already part of `hsmtool` since last release..
Descriptor are really neat but ya footgun in our case, and yea, as Zmn points out you cannot recover the `hsm_secret` from the xpriv.

doc/BACKUP.md
ZmnSCPxj: How about "up to the entire channel funds can be revoked by the peer"?

ZmnSCPxj:
> Seeing the discussion on how to backup the sqlite file, would it not be best to not filesystem-copy it, but sqlite3 .dump it? https://www.sqlitetutorial.net/sqlite-dump/ or would this not guarantee consistency?

It locks the file while doing the dump.  The `VACUUM INTO` query is similar.  If `lightningd` accesses the database while your backup process is `.dump`ing or `VACUUM INTO`ing, it will crash `lightningd`, which is bad.

doc/BACKUP.md
ZmnSCPxj: No, have never used it, will remove the link instead.

doc/BACKUP.md
ZmnSCPxj: No, wait, this is true if the `option_dataloss_protect` is *not* supported, which is what qualifies the statement.  Without `option_dataloss_protect` we would get confused with the `reestablish_connection` and try to broadcast our obsolete revoked commitment transaction, and the peer will react by revoking the revoked commitment transaction.

doc/BACKUP.md
ZmnSCPxj: I mildly dislike the `###` prefixes since to me "more `#` = more heavier" i.e. `###` should map to `<h1>`, `#` should map to `<h3>`.  *shrug*.  Changed anyway.

doc/BACKUP.md
ZmnSCPxj: It is `--wallet`, and it is mentioned in the gabridome guide, but will add a random reference to it here.

ZmnSCPxj:
Rebased, reordered sections as per @cdecker, changed to use `#` for section headings, various changes as per @cdecker , remove link to /fiatjaf/mcdlsp , warn against `.dump` and `VACUUM INTO`, mention use of multiple USB flash disks for RAID1.

darosior (approved):
ACK 948489ae08e735ea31c750a5bc68f6a82bd39e07
thanks for taking the time to write this up.

cdecker:
Excellent work @everyone, this was very much needed, and thanks in particular to @ZmnSCPxj for taking the initiative :+1:

doc/BACKUP.md
```diff
@@ -0,0 +1,440 @@
...
+lose all funds in open channels.
+
+However, again, note that a "no backups #reckless" strategy leads to
+*definite* loss of funds, so you should still prefer *this* strategy rather
+than having *no* backups at all.
+
+Even if you have one of the better options above, you might still want to do
+this as a worst-case fallback, as long as you:
+
+* Attempt to recover using the other backup options below first.
```
ZmnSCPxj, line 364: s/below/above/

LOL I did not fix it up after rearranging the text!

domegabri:
> > Hi,
> > there's a new possibility for the backup of hsm_secret file.
> > I find it useful because it converts to xpriv format, so you can import into one descriptor based wallet in Bitcoin Core or wherever xpriv are accepted.
> > https://github.com/domegabri/lightning-secret
> 
> FWIW, it was already part of `hsmtool` since last release (EDIT: and i noticed Riccardo Casatta did a new one last week too ! ) ..
> Descriptor are really neat but ya footgun in our case, and yea, as Zmn points out you cannot recover the `hsm_secret` from the xpriv.

Hi @darosior , I think the last release allows to dump public version of the descriptors. But I needed xpriv because I had a corrupt db and wanted to scan for coins and recover tehm

darosior:
@domegabri oh right we removed it in later stage as it was a nice footgun! https://github.com/ElementsProject/lightning/pull/4171#issue-513075486

PR #4209, opened 2020-11-18
htlc.h: move NUM_SIDES to define, not enum member

niftynei:
Drive by fixup.

Changelog-None

### DIFF
--- a/common/htlc.h
+++ b/common/htlc.h
@@ -10,10 +10,10 @@
 #include <ccan/short_types/short_types.h>
 #include <ccan/str/str.h>
 
+#define NUM_SIDES (REMOTE + 1)
 enum side {
 	LOCAL,
 	REMOTE,
-	NUM_SIDES
 };
 
 /* What are we doing: adding or removing? */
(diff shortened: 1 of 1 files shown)


common/htlc.h
```diff
@@ -10,10 +10,10 @@
 #include <ccan/short_types/short_types.h>
 #include <ccan/str/str.h>
 
+#define NUM_SIDES (REMOTE + 1)
 enum side {
 	LOCAL,
 	REMOTE,
-	NUM_SIDES
```
ZmnSCPxj, line 16: I believe the intent here is that this is always the last entry in this enum, so that if we ever get more than two sides to a channel (fat chance), we just insert them before the `NUM_SIDES` entry.  Standard C assigns enum values monotonically IIRC.

With `NUM_SIDES` as a `#define` you have to insert the new side to the `enum`, ***and*** update the `#define`.

But in any case there is going to be a lot more changes if we ever have more than two sides to a channel so....

niftynei: Using this comment as context for this change https://github.com/ElementsProject/lightning/pull/4146#discussion_r512341287

cdecker: The rationale for not having it as part of the `enum` but rather have a `#define` is that exhaustive matching requires us to add a `case NUM_SIDES` to all `switch` statements (and technically `if` statements) if we keep it as part of the `enum`. +1 for moving it to a `#define`.

cdecker (approved):
ACK c82462354f81e67a9f953c48cc87386a83c5376c

PR #4210, opened 2020-11-18
min_witness_weight param for fundpsbt/utxopsbt

niftynei:
Add a `min_witness_weight` param to our 'fund-/utxo-psbt' methods. This is to support the dual funding spec, which requires a floor of witness fees to be paid at 110 weight.

In theory, if we ever get more complicated UTXOs, we'll want to update the witness weight calculation to encompass the correct weight for these, also.

### DIFF
--- a/wallet/reservation.c
+++ b/wallet/reservation.c
@@ -386,7 +386,7 @@ static struct command_result *json_fundpsbt(struct command *cmd,
 {
 	struct utxo **utxos;
 	u32 *feerate_per_kw;
-	u32 *minconf, *weight;
+	u32 *minconf, *weight, *min_witness_weight;
 	struct amount_sat *amount, input, diff;
 	bool all, *reserve;
 	u32 *locktime, maxheight;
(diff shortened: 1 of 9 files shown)


rustyrussell:
Since the difference is small, I wonder if we should just set this to 110 always, and note that this is a requirement to comply with the df spec?

wallet/reservation.c
```diff
@@ -427,7 +429,9 @@ static struct command_result *json_fundpsbt(struct command *cmd,
 						    "impossible UTXO value");
 
 			/* But also adds weight */
-			*weight += utxo_spend_weight(utxo);
+			fprintf(stderr, "min_witness_weight is %u\n",
+					*min_witness_weight);
```
cdecker, line 433: Debugging statement.

niftynei:
> Since the difference is small, I wonder if we should just set this to 110 always, and note that this is a requirement to comply with the df spec?

We could, but at the cost (haha) of precision in feerate on non-dual-funded transactions.

niftynei:
Merged elsewhere.

PR #4211, opened 2020-11-18
Dual-funding over multifundchannel fixups, etc

niftynei:
Starts at https://github.com/ElementsProject/lightning/commit/60c89018a678b4daca752614c1baa8e65eba25f2. Includes commits that are in #4210.

This PR does a few things
- We rename the `dual_open` wires to `dualopend`, as this conforms to the other daemon naming scheme.
- We remove some unnecessary test code, since `multifundchannel` now handles it
- We set the minimum weight on an input to 110, iif we're finding the fee/funding for a opening tx that will have some `opt_dual_fund` participants.


rustyrussell (approved):
Ack 5b8ea32cc23266f16a761fc404dacfc2c01f2745

PR #4213, opened 2020-11-18
lightningd: remove unused pid_fd member in struct lightningd

siburu:
Remove `pid_fd` from `struct lightningd` since `pid_fd` is not used anywhere.

### DIFF
--- a/lightningd/lightningd.h
+++ b/lightningd/lightningd.h
@@ -83,8 +83,6 @@ struct lightningd {
 	 * -1. */
 	int daemon_parent_fd;
 
-	int pid_fd;
-
 	/* Our config basedir, network directory, and rpc file */
 	char *config_basedir, *config_netdir;
 


ZmnSCPxj (approved):
ACK b631711606872449c41f0b364494b41b0869b10b

This was used in the past when we had different `fork` behavior on `daemonize`, but I believe the current daemonize code is fairly stable now.

siburu:
@ZmnSCPxj Sorry if I have misunderstood the code.

This was used for explicitly closing a PID file, wasn't it?
The PID file is still used whether lightning is daemonized or not.
However, in the current implementation, the PID file is not closed explicitly but closed implicitly at process exit.
Therefore `pid_fd` is not needed/used now, I think.

ZmnSCPxj:
Yes, but in the past we opened it in the starting process, then post-daemonization we wrote the PID once in the daemon process.

siburu:
I see, I understand. Thank you!

PR #4225, opened 2020-11-24
libwally: update to latest which uses libsecp-zkp submodule

rustyrussell:
And get rid of the now-obsolete PYTHON_VERSION hack.

Changelog-None.


niftynei:
ACK 4d276cf7e65b1b5efcedc5de07d205fc73b45403z

PR #4226, opened 2020-11-25
invoice: make invoice_payment hook a multi-user hook.

rustyrussell:
We register on it for offers, and without this nobody else can.

Changelog-Changed: plugins: more than one plugin can now register invoice_payment hook.

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -233,42 +233,20 @@ static const u8 *hook_gives_failmsg(const tal_t *ctx,
 }
 
 static void
-invoice_payment_hook_cb(struct invoice_payment_hook_payload *payload STEALS,
-			const char *buffer,
-			const jsmntok_t *toks)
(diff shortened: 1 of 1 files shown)


lightningd/invoice.c
```diff
@@ -393,6 +397,10 @@ void invoice_try_pay(struct lightningd *ld,
 	payload->set = set;
 	tal_add_destructor2(set, invoice_payload_remove_set, payload);
 
+	/* We notify here to benefit from the payload and because the hook callback is
```
ZmnSCPxj, line 400: Comment no longer quite matches?  This is not the hook callback anymore, right?

rustyrussell: Good point!

ZmnSCPxj (approved):
ACK 2dc52dbe47ceb16f73ddcb335132d3cebbf551c7

rustyrussell:
Trivial fold-fixup-and-rebase-on-master.

Issue #4222, opened 2020-11-23
Crash in lightning_connectd on master on FreeBSD


laanwj:
I upgraded from 711133e3c to 66b2bb9a9 and get a crash soon after launch now (every time):
```
~/src/clightning % gdb ./lightningd/lightning_connectd /coredumps/1001.lightning_connectd.33684.core
GNU gdb (GDB) 9.2 [GDB v9.2 for FreeBSD]
[New LWP 101203]
Core was generated by `/usr/home/user/src/clightning/lightningd/lightning_connectd'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  fmt_wireaddr_internal (ctx=0x800a76088, a=0x76130) at common/wireaddr.c:212
212             switch (a->itype) {
(gdb) bt
#0  fmt_wireaddr_internal (ctx=0x800a76088, a=0x76130) at common/wireaddr.c:212
#1  0x0000000000252c3a in type_to_string_ (ctx=0x800a76088, typename=0x2052b5 "wireaddr_internal", u=...) at common/type_to_string.c:35
#2  0x000000000024868c in destroy_io_conn (conn=<optimized out>, connect=0x800ba9168) at connectd/connectd.c:703
#3  0x000000000027157d in io_close_taken_fd (conn=0x800b50268) at ccan/ccan/io/io.c:463
#4  0x0000000000245aaf in peer_connected (conn=0x800b50268, daemon=<optimized out>, id=<optimized out>, addr=0x800bae854, cs=0x800bae960, their_features=0x800a76928 "\002\242\241\001") at connectd/connectd.c:511
#5  0x0000000000248a46 in peer_init_received (conn=0x800b50268, peer=<optimized out>) at connectd/peer_exchange_initmsg.c:94
#6  0x0000000000270cec in next_plan (conn=0x800b50268, plan=<optimized out>) at ccan/ccan/io/io.c:59
#7  0x00000000002713ef in io_ready (conn=0x800b50268, pollflags=1) at ccan/ccan/io/io.c:417
#8  0x00000000002726b2 in io_loop (timers=0x800b48090, expired=0x7fffffffe9a0) at ccan/ccan/io/poll.c:445
#9  0x000000000024618b in main (argc=<optimized out>, argv=<optimized out>) at connectd/connectd.c:1703
(gdb) print *a
Cannot access memory at address 0x76130
```
It looks like an invalid address is being passed (use after free?).

[...]

PR #4228, opened 2020-11-25
connectd: clean up io_finish for incoming conn

niftynei:
Need to figure out why this isn't the same as the outgoing->conn

Fixes: #4222

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -280,6 +280,9 @@ static void connected_to_peer(struct daemon *daemon,
 {
 	struct connecting *outgoing;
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
This is correct, FWIW.  We should be removing the destructor on *both* connections, when we have simultaneous connects.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4228/commits/8b8e2f3e413e4c3387e8e9d6a8ec738a833636c0

rustyrussell:
Closed, since #4222 was also closed.

PR #4230, opened 2020-11-26
tx: Remove wally workaround that is no longer needed

jgriffiths:
Just a minor nit now that wally supports clearing structs on PSBT members.

Changelog-None.


rustyrussell (approved):
Ack 240e8c8165c08330d7a51c668aee137935d088f8

PR #4235, opened 2020-11-30
pyln: default fundamount for fundchannel and openchannel

m-schmoock:
Since we are already using `10**6` as default amount on e.g. `line_graph` we could also use this for fundchannel and openchannel. This makes testcases where we e.g.  link a circle using `line_graph` a tiny bit simpler.


rustyrussell (approved):
Ack 1e7e0d4c3b01f063e29a0ce8e2abef049fae0d54

PR #4244, opened 2020-12-02
invoice: fix potential race where invoice is paid/expired while we're calling hook

rustyrussell:
There's actually a (very unlikely) race here: we would previously have
crashed with an assertion in invoices_resolve.

Changelog-None

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -250,11 +250,17 @@ invoice_payment_hooks_done(struct invoice_payment_hook_payload *payload STEALS)
 		return;
 	}
 
+	/* Paid or expired in the meantime. */
+	if (!wallet_invoice_resolve(ld->wallet, invoice, payload->msat)) {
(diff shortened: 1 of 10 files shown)


ZmnSCPxj (approved):
ACK 7d44660ebba0b9c5c8af260e6e80c22bf637c8aa

Changes to `wallet/wallet` and `wallet/invoices` looks fine to me, do not quite grok the changes in the hook code in `lightningd/invoice` tho.

rustyrussell:
> ACK [7d44660](https://github.com/ElementsProject/lightning/commit/7d44660ebba0b9c5c8af260e6e80c22bf637c8aa)
> 
> Changes to `wallet/wallet` and `wallet/invoices` looks fine to me, do not quite grok the changes in the hook code in `lightningd/invoice` tho.

If you look just above the change, you'll see we send the same error as before, on failure (makes sense, since invoice has already been paid).

PR #4245, opened 2020-12-02
Offers Part 1: Groundwork

rustyrussell:
Some reshuffles which make detaching invoices from bolt11 format easier, and I cleaned up generation a little too.

Changelog-None


niftynei:
ACK e27ace4991e34c3bbaf168e6658f01a34fb17c2f

Issue #4221, opened 2020-11-23
Add `getchannelfees` command


0e319b6:
### Motivation
Currently there is no good way to view channel fees, especially all in one condensed view.  I have a hacky bash alias for this: `lightning-cli summary | awk '{print $2}' | egrep [0-9]{5}x[0-9]+x[0-9]); do echo $scid: ; lightning-cli listchannels $scid | grep fee; done` but that's obviously suboptimal.

### Proposed command format
```
getchannelfees [id]
```
where the optional argument `id` is either a peerid pubkey of one of our channel peers, a channel id, or a short channel ID.

### Proposed RPC format
I propose this command returns an object with a field named `channel_fees` that is a list of objects of the following format:
```
{
    "destination": <peer id pubkey>,
    "short_channel_id": <scid>,
    "from_destination_base_fee_millisatoshi": <int>,
    "from_destination_fee_per_millionth": <int>,
    "to_destination_base_fee_millisatoshi": <int>,
    "to_destination_fee_per_millionth": <int>
}
```

For each such object listed, if our node participates in the corresponding channel, `destination` shall be our corresponding channel peer's peerid.  In the alternative, `destination` may be the peerid of either of the channel peers; a potential `source` field can be omitted as it is implied by `destination` and `short_channel_id`.  The other fields should hopefully be self-explanatory.

In case of an error, the `channel_fees` list will be empty.

### Optional extensions
We could also allow `id` to represent a peerid of a node who is *not* one of our node's channel peers.  In this case, we can run Dijkstra's algorithm on the channel graph and return the lowest cumulative fees between our node and `id`.

A further extension would additionally allow an optional `[amount]` argument that, if `id` represents a peerid to which we are not peered, represents a minimum channel size for the above Dijkstra's algorithm.  During the operation of that search, any channel with a capacity less than `amount` would be ignored.

ZmnSCPxj:
> We could also allow `id` to represent a peerid of a node who is not one of our node's channel peers. In this case, we can run Dijkstra's algorithm on the channel graph and return the lowest cumulative fees between our node and `id`.

Because of the proportional fee, this requires that you provide a reference amount to be sent to the designated peer.  We cannot "just" accumulate the fees along the way due to how the math works out.

And if you want a Dijkstra algorithm, that already exists in `getroute` with `fuzzpercent=0` (default is 5).  You need to provide a reference amount, but the total fees would be the amount in the first hop minus the reference amount.

[...]

PR #4247, opened 2020-12-02
feat: fees on listpeers rpc output

m-schmoock:
As mentioned in the last developer meeting and additional to #4221 we want to have the local half-channel fees (the ones that are applied when a payment is routed though our node) as part of the `listpeers` output. Motivation is that this data is non-gossip and currently only available via gossip `listchannels`.

Example
```
{
    ...
    "fee_base_msat" : "100msat",
    "fee_proportinal_millionths" : 1000,
    ...
}
```

This PR adds feature, test and doc.

### DIFF
--- a/doc/lightning-listpeers.7.md
+++ b/doc/lightning-listpeers.7.md
@@ -100,6 +100,10 @@ The objects in the *channels* array will have at least these fields:
   a number followed by a string unit.
 * *total\_msat*: A string describing the total capacity of the channel;
   a number followed by a string unit.
+* *fee_base_msat*: The fixed routing fee we charge for forwards going out over
+  this channel, regardless of payment size.
+* *fee_proportional_millionths*: The proportional routing fees in ppm (parts-
+  per-millionths) we charge for forwards going out over this channel.
 * *features*: An array of feature names supported by this channel.
 
 These fields may exist if the channel has gotten beyond the `"OPENINGD"`
@@ -238,7 +242,8 @@ Michael Hawkins <<michael.hawkins@protonmail.com>>.
 SEE ALSO
 --------
 
-lightning-connect(7), lightning-fundchannel\_start(7)
+lightning-connect(7), lightning-fundchannel\_start(7),
+lightning-setchannelfee(7)
 
 RESOURCES
 ---------
--- a/doc/lightning-listpeers.7
+++ b/doc/lightning-listpeers.7
@@ -138,6 +138,12 @@ a number followed by a string unit\.
 \fItotal_msat\fR: A string describing the total capacity of the channel;
 a number followed by a string unit\.
 .IP \[bu]
+\fIfee_base_msat\fR: The fixed routing fee we charge for forwards going out over
+this channel, regardless of payment size\.
+.IP \[bu]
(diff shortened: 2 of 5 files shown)


doc/lightning-listpeers.7.md
```diff
@@ -100,6 +100,8 @@ The objects in the *channels* array will have at least these fields:
   a number followed by a string unit.
 * *total\_msat*: A string describing the total capacity of the channel;
   a number followed by a string unit.
+* *fee_base*: The applied routing fees for this half channel as msat string.
```
ZmnSCPxj, line 103: The term "half channel" is never otherwise used in this, or any other, doc.  Perhaps better "the routing fee we will charge for forwards going out over this channel, regardless of payment size".

doc/lightning-listpeers.7.md
```diff
@@ -100,6 +100,8 @@ The objects in the *channels* array will have at least these fields:
...
+* *fee_ppm*: The applied proportional ppm fees for this half channel as a number.
```
ZmnSCPxj, line 104: Similarly "the routing fee we will charge for forwards going out over this channel, in millisatoshi for every 1 million millisatoshi ('parts per million') of payment size."

ZmnSCPxj: Also suggest adding `lightning-setchannelfee` in the "SEE ALSO" section at the end of the file.

ZmnSCPxj (approved):
ACK 8dbe6f15d38858ab06ea01d5ba5ac17dbf0137f5

Minor quibble about docs but not enough to block this.

doc/lightning-listpeers.7.md
m-schmoock: Ack, but I skipped the 'millisatoshi' part, as this is not relevant for a proportional value if its 'millisatoshi', 'satoshi', 'bitcoin' or 'finney'. Acutally the `setchannelfee` doc has the same issue and quotes ppm as 'in satoshi' :D ... however I wrote this manpage two years ago.

rustyrussell:
OK, so minor feedback:

1. We prefer to use spec names for fields (unless we've already misuse elsewhere).  That makes it  `fee_base_msat` and `fee_proportional_millionths`.
2. msat field names should end in _msat so that the Python JSON decoder can automagically turn them into Millisatoshi amounts.

m-schmoock:
@rustyrussel
Thx for the feedback. I changed the names.

I figured that in the current `listchannels` output `base_fee_millisatoshi` is currently not having a `_msat` pendant. Should we add this in another PR? If we want to eventually get rid of non-msat fields we need to add this.

m-schmoock:
> 1. We prefer to use spec names for fields (unless we've already misuse elsewhere).  That makes it  `fee_base_msat` and `fee_proportional_millionths`.

Note: Just checked, `listchannels` gives me `fee_per_millionth` ...

m-schmoock:
@rustyrussell 
> 1. We prefer to use spec names for fields (unless we've already misuse elsewhere).  That makes it  `fee_base_msat` and `fee_proportional_millionths`.

Note: Just checked, `listchannels` gives me `fee_per_millionth` ...

cdecker:
ACK 54af4c9fde80abde7c7339c9608286ac1ded3c2c

PR #4249, opened 2020-12-02
pytest: recreate wallet on bitcoind restart (for master branch of bitcoind)

rustyrussell:
Upgraded builder, and it doesn't work any more.  Suggested by @achow101 that we make older versions behave like new ones, with `-nowallet`.

Also found an unrelated taken() leak while testing.

Changelog-None


cdecker (approved):
ACK ca6d5559fbe0edbec5c71c61ad15c353cfe6bb82

PR #4250, opened 2020-12-02
Reduce mem and cpu usage of onchaind for long-lived channels

rustyrussell:
Noticed my node's onchaind was chewing memory and CPU (I run under valgrind!), caught the IO from onchaind and created this replay test.  It will break as soon as we change onchaind's wire format, but it's a good test for now.


onchaind/onchaind.c
```diff
@@ -3640,6 +3647,12 @@ static void handle_unknown_commitment(const struct tx_parts *tx,
 	wait_for_resolved(outs);
 }
 
+static int cmp_htlc_cltv(const struct htlc_stub *a,
+			 const struct htlc_stub *b, void *unused)
+{
+	return a->cltv_expiry - b->cltv_expiry;
```
ZmnSCPxj, line 3692: `cltv_expiry` is unsigned 32-bit.  If `int` is not 32-bit, this might not work correctly (it might not sign-extend if `int` is 64-bit).  I suggest casting to `int` before doing the subtraction, or better yet, just being explicit about the compare --- on most CPUs the cost of a subtract is the same as the cost of a compare (some use the same instruction for it!), and most compilers can compile `if (a < b) return -1; else if (a == b) return 0; else return 1;` into a compare/subtract instruction and extraction from flags register.

rustyrussell: Yes, will fix.

Good catch!

onchaind/onchaind.c
```diff
@@ -152,6 +152,10 @@ static void trim_maximum_feerate(struct amount_sat funding,
 	size_t weight;
 	struct amount_sat fee = funding;
 
+	/* FIXME: This doesn't work for elements? */
+	if (chainparams->is_elements)
+		return;
```
cdecker, line 157: Good question, I don't think it should be too much of an issue, given that we currently only have one main-asset input, which also makes the below `if (!amount_asset_is_main(&amt))` always false, but better keep it in in case we eventually add multi-input and multi-asset channels.

Was there a concrete error that was triggered on elements, or is this more of a precaution?

cdecker (approved):
Aside from the elements question I think this is good to go :+1:

ACK 40931f17b166c9e54964eb7557d2a1edccda5530

niftynei (approved):
ACK 40931f17b166c9e54964eb7557d2a1edccda5530

onchaind/onchaind.c
rustyrussell: Yes, we had a -1ULL satoshis value. Then added this test and this function caused an onchaind failure (presumably by giving the wrong result).

rustyrussell:
Trivial rebase, plus changelog additions.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4250/commits/ea4d6dc0e82648e2a723341883dd3ea9f753fd2f

PR #4255, opened 2020-12-04
Offers: Part 1, Spec import and basic infrastructure

rustyrussell:
BOLT 12 import, and various helpers to support that (offers, invoice_request, invoice) including signing infrastructure.

Almost all is EXPERIMENTAL_FEATURES, though we'll probably move to a runtime option before release.

Changelog-None

PR #4256, opened 2020-12-04
Offers Part 2: db and invoice infrastructure.

rustyrussell:
~(Built on #4255: first 10 commits)~ That's merged!

This is the db changes for offers, and linking them to payments or invoices, especially so we can enforce atomicity (for single-use offers).


PapauloGamerOfc:
The Travis CI bot is failing on every PR!

rustyrussell:
Rebase on master

rustyrussell:
Rebase on master x2

lightningd/pay.c
```diff
@@ -1213,6 +1285,9 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt("bolt11", param_string, &b11str),
 		   p_opt_def("msatoshi", param_msat, &msat, AMOUNT_MSAT(0)),
 		   p_opt("destination", param_node_id, &destination),
+#if EXPERIMENTAL_FEATURES
+		   p_opt("local_offer_id", param_sha256, &local_offer_id),
```
cdecker, line 1300: I'd prefer not to introduce more underscore-delimited arguments, and shorter is better (`offerid` seems to fit our prior naming schema).

lightningd/offer.c
```diff
@@ -0,0 +1,209 @@
...
+				const char *b12,
+				const struct json_escape *label,
+				enum offer_status status)
+{
+	json_add_sha256(response, "offer_id", offer_id);
+	json_add_bool(response, "active", offer_status_active(status));
+	json_add_bool(response, "single_use", offer_status_single(status));
+	json_add_string(response, "bolt12", b12);
+	if (status == OFFER_USED)
+		json_add_bool(response, "used", true);
```
cdecker, line 24: Shouldn't we always include the `used` flag? Checking for existence and then checking the value on the client can be simplified if we do

lightningd/invoice.c
```diff
@@ -1441,3 +1475,162 @@ static const struct json_command decodepay_command = {
...
+	struct json_stream *response;
+	struct bolt11 *b11;
+	struct sha256 hash;
+	u5 *sig;
+	bool have_n;
+	char *fail;
+
+	if (!param(cmd, buffer, params,
+		   p_req("invstring", param_string, &invstring),
+		   p_req("preimage", param_preimage, &preimage),
```
cdecker, line 1518: Can we make the `preimage` optional? That way we can build arbitrary invoice handling structures using `htlc_accepted` hooks, that do not require internal resolution, but still require a valid signature from the node?

I can think of two scenarios that this would be useful right away:
 - Create fake MPP-probe invoices that we do not have preimages for (and want to prove that's the case by using a derivation that doesn't involve a preimage such as "xor two hashes with each other")
 - Have an external invoice management system, in which the node is untrusted and requires an incoming payment be acknowledged by the external system (accounting, compliance, ....)

lightningd/invoice.c
```diff
@@ -1441,3 +1475,162 @@ static const struct json_command decodepay_command = {
...
+		   p_req("label", param_label, &label),
```
cdecker, line 1519: If we make `preimage` optional, then `label` needs to move up in order to have all required params be first.

lightningd/invoice.c
```diff
@@ -1441,3 +1475,162 @@ static const struct json_command decodepay_command = {
...
+
+		if (!wallet_invoice_create(cmd->ld->wallet,
+					   &invoice,
+					   b11->msat,
+					   label,
+					   b11->expiry,
+					   b11enc,
+					   b11->description,
+					   b11->features,
+					   preimage,
```
cdecker, line 1548: I guess it not being verified that the preimage matches the hash is a form of my desired situation above, but if that's intended we should make sure that we handle preimage / hash mismatches correctly.

lightningd/invoice.c
```diff
@@ -1441,3 +1475,162 @@ static const struct json_command decodepay_command = {
...
+			} else
+				local_offer_id = NULL;
+
+			if (inv->amount)
+				msat = amount_msat(*inv->amount);
+
+			if (inv->relative_expiry)
+				expiry = *inv->relative_expiry;
+			else
+				expiry = 7200;
```
cdecker, line 1591: Pull this constant out into a `#define` so we can find it again later :-)

lightningd/pay.c
```diff
@@ -1282,7 +1292,8 @@ static struct command_result *json_sendonion(struct command *cmd,
 		   p_opt("label", param_escaped_string, &label),
 		   p_opt("shared_secrets", param_secrets_array, &path_secrets),
 		   p_opt_def("partid", param_u64, &partid, 0),
-		   p_opt("bolt11", param_string, &b11str),
+		   /* FIXME: paramter should be invstring now */
```
cdecker, line 1295: Nit: `paramter` -> `parameter`

lightningd/pay.c
```diff
@@ -1436,7 +1447,8 @@ static struct command_result *json_sendpay(struct command *cmd,
 		   p_req("payment_hash", param_sha256, &rhash),
 		   p_opt("label", param_escaped_string, &label),
 		   p_opt("msatoshi", param_msat, &msat),
-		   p_opt("bolt11", param_string, &b11str),
+		   /* FIXME: paramter should be invstring now */
```
cdecker, line 1450: Nit: `paramter` -> `parameter`

lightningd/pay.c
```diff
@@ -1546,31 +1558,43 @@ static struct command_result *json_listsendpays(struct command *cmd,
 	const struct wallet_payment **payments;
 	struct json_stream *response;
 	struct sha256 *rhash;
-	const char *b11str;
+	const char *invstring;
 
 	if (!param(cmd, buffer, params,
-		   p_opt("bolt11", param_string, &b11str),
+		   /* FIXME: paramter should be invstring now */
```
cdecker, line 1564: Nit: `paramter` -> `parameter`

wallet/wallet.c
```diff
@@ -3961,3 +3961,169 @@ void wallet_penalty_base_delete(struct wallet *w, u64 chan_id, u64 commitnum)
...
+		else
+			*label = db_column_json_escape(ctx, stmt, 1);
+	}
+	if (status)
+		*status = offer_status_in_db(db_column_int(stmt, 2));
+	tal_free(stmt);
+	return bolt12;
+}
+
+struct db_stmt *wallet_offer_first(struct wallet *w, struct sha256 *offer_id)
```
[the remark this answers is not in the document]
rustyrussell: Yes: I copied the invoice iterator which was a bad idea.

wallet/db.c
```diff
@@ -664,6 +664,8 @@ static struct migration dbmigrations[] = {
 	 ", status INTEGER"
 	 ", PRIMARY KEY (offer_id)"
 	 ");"), NULL},
+    /* A reference into our own offers table, if it was made from one */
+    {SQL("ALTER TABLE invoices ADD COLUMN local_offer_id BLOB DEFAULT NULL;"), NULL},
```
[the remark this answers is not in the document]
rustyrussell: Looks like appending "REFERENCES offers(offer_id)" is sufficient.

lightningd/pay.c
rustyrussell: offerid is a bit confusing though, since most payments are made in reference to an offer (and we'll all be driving flying cars, obv).

But this is only to link it to a *local* offer id.  I can elide underscores and make it localofferid?

lightningd/invoice.c
rustyrussell: That's a bigger change; we'd have to allow the preimage entry to be null in the db.  However, I'll reorder the args so we can make it optional in future more easily?

lightningd/invoice.c
rustyrussell: No, I imagine if we don't get an internal error, we'll at least lose the channel when we serve our peer an incorrect preimage.  I will enforce it here with a check.

lightningd/pay.c
rustyrussell: Did a sweep, found 4 of them!

wallet/db.c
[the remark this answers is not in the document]
cdecker: Seems postgresql likes it as well :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4256/commits/01c92fe6259179434a7689ed52333a518abb3d4f

lightningd/invoice.c
cdecker: Thanks for reordering, I'll try to tackle the `p_opt("preimage", ...)` next ^^

PR #4259, opened 2020-12-04
Improve validation of `txprepare` arguments

### LINKED ISSUE #4238: fundchannel fails with lnd node
I'm trying to open a channel with

```
fundchannel 029dd81c32e3ea839b4b2f8edcc9bea227c40fa2c6509878034401a8ce2a8bac05 all fast true 0 [TXID:N]
```
And I get an error which I do not understand.

```
lightning-cli: Incorrect 'id' in response: null
```

### `getinfo` output

```
{
   "id": "ID",
   "alias": "test",
   "color": "ff0000",
   "num_peers": 10,
   "num_pending_channels": 1,
   "num_active_channels": 6,
   "num_inactive_channels": 1,
   "address": [
      {
         "type": "ipv4",
         "address": "46.101.127.98",
         "port": 51486
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 51486
      }
   ],
   "version": "v0.9.2",
   "blockheight": 659508,
   "network": "bitcoin",
   "msatoshi_fees_collected": 16029,
   "fees_collected_msat": "16029msat",
   "lightning-dir": "/home/lnadmin/.lightning/bitcoin"
}
```

### LINKED ISSUE #4258: txprepare destinationAddress:all slow 1 "[\"yourOutpoint\"]" sends everything to the change address
I have tried to use `txprepare` to send the whole amount of one UTXO to and address of my personal wallet.

I have used the syntax on the subject:
```
lightning-cli txprepare myPersonalWalletAddress:all slow 1 "[\"nodeOutpoint\"]"

{
   "unsigned_tx": "0200000......",
   "txid": "cd5..."
}
bitcoin-cli decoderawtransaction 020000...
{
  "txid": "cd5...",
  "hash": ".....",
  "version": 2,
  "size": 82,
  "vsize": 82,
  "weight": 328,
  "locktime": 659783,
  "vin": [
    {
      "txid": "fkkb...",
      "vout": 1,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "sequence": 4294967293
    }
  ],
  "vout": [
    {
      "value": 0.0029888929,
      "n": 0,
      "scriptPubKey": {
        "asm": "0 8...",
        "hex": "00...",
        "reqSigs": 1,
        "type": "witness_v0_keyhash",
        "addresses": [
          "bc1...." <<<<<<<<<<<<<<<<<<<<< not myPersonalWalletDestinationAddress
        ]
      }
    }
  ]
}

```
The destination address in the output is one of the change addresses of my node.

### `getinfo` output
```
{
   "id": "02....",
   "alias": "mynode",
   "color": "27ckhs",
   "num_peers": 0,
   "num_pending_channels": 0,
   "num_active_channels": 0,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "torv3",
         "address": "onionaddress.onion",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 9735
      }
   ],
   "version": "v0.9.2",
   "blockheight": 659883,
   "network": "bitcoin",
   "msatoshi_fees_collected": 838,
   "fees_collected_msat": "838msat",
   "lightning-dir": "/home/username/.lightning/bitcoin"
}
```

cdecker:
There are two issues we address in this PR:
 - Broken JSON strings when forwarding params as raw strings to further JSON-RPC calls, causing the infamous `id is null` error because the side receiving the mangled request cannot extract the `id` and returns a generic error instead, which doesn't match up with our expectations.
 - Interpreting a string param as an array, resulting in an empty array, and thus only adding the change output to a prepared tx.

In a couple of places we accept arrays of strings and don't validate
them. If we forward them, e.g., call a JSON-RPC method from the
plugin, we end up embedding the unverified string in the JSON-RPC
call without escaping, which then leads to invalid JSON being passed
on.

We were not checking that outputs is indeed an array, and just going
ahead creating the array of outputs. Since `tok->size` for a string is
0 we ended up ignoring the argument altogether and thus the created
transaction would end up only with a single change output.

Since I was looking at the cli params already I also added some more feedback
to `lightning-cli`. It now parses the request before sending it to `lightningd`, so if
a malformed param results in a malformed request we catch it early and tell the
user to check the formatting of the params.

Fixes #4238 
Fixes #4258

### DIFF
--- a/bitcoin/tx.h
+++ b/bitcoin/tx.h
@@ -21,6 +21,12 @@ struct wally_psbt;
 struct bitcoin_txid {
 	struct sha256_double shad;
 };
+
+struct bitcoin_outpoint {
+	struct bitcoin_txid txid;
+	u16 n;
+};
+
 /* Define bitcoin_txid_eq */
 STRUCTEQ_DEF(bitcoin_txid, 0, shad.sha.u);
 
--- a/bitcoin/tx.h
+++ b/bitcoin/tx.h
@@ -24,7 +24,7 @@ struct bitcoin_txid {
 
 struct bitcoin_outpoint {
 	struct bitcoin_txid txid;
-	u16 n;
+	u32 n;
 };
 
 /* Define bitcoin_txid_eq */
--- a/common/json_tok.c
+++ b/common/json_tok.c
@@ -501,3 +501,32 @@ struct command_result *param_psbt(struct command *cmd,
 	return command_fail_badparam(cmd, name, buffer, tok,
 				     "Expected a PSBT");
 }
+
+struct command_result *param_outpoint_arr(struct command *cmd,
+					  const char *name,
+					  const char *buffer,
+					  const jsmntok_t *tok,
+					  struct bitcoin_outpoint **outpoints)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	if (tok->type != JSMN_ARRAY) {
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "Could not decode the outpoint array for %s: "
+				    "\"%s\" is not a valid outpoint array.",
+				    name, json_strdup(tmpctx, buffer, tok));
+	}
+
+	*outpoints = tal_arr(cmd, struct bitcoin_outpoint, tok->size);
+
+	json_for_each_arr(i, curr, tok) {
+		struct bitcoin_outpoint op;
+		if (!json_to_outpoint(buffer, curr, &op))
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					    "Could not decode outpoint \"%.*s\", "
+					    "expected format: txid:output",
+					    json_tok_full_len(curr), json_tok_full(buffer, curr));
+		(*outpoints)[i] = op;
+	}
+	return NULL;
+}
--- a/common/json_tok.c
+++ b/common/json_tok.c
@@ -520,13 +520,11 @@ struct command_result *param_outpoint_arr(struct command *cmd,
 	*outpoints = tal_arr(cmd, struct bitcoin_outpoint, tok->size);
 
 	json_for_each_arr(i, curr, tok) {
-		struct bitcoin_outpoint op;
-		if (!json_to_outpoint(buffer, curr, &op))
+		if (!json_to_outpoint(buffer, curr, &(*outpoints)[i]))
 			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 					    "Could not decode outpoint \"%.*s\", "
 					    "expected format: txid:output",
 					    json_tok_full_len(curr), json_tok_full(buffer, curr));
-		(*outpoints)[i] = op;
 	}
 	return NULL;
 }
--- a/common/json_tok.h
+++ b/common/json_tok.h
@@ -11,6 +11,7 @@
 struct amount_msat;
 struct amount_sat;
 struct bitcoin_txid;
+struct bitcoin_outpoint;
 struct channel_id;
 struct command;
 struct command_result;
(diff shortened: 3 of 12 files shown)


niftynei (commented):
ACK 10f842e0b85565759f01d788d84cc7ecc1c1bb96

Awesome work finding this!

common/json_tok.c
```diff
@@ -501,3 +501,58 @@ struct command_result *param_psbt(struct command *cmd,
...
+	json_for_each_arr(i, curr, tok)
+	{
+		if (curr->type != JSMN_STRING) {
+			return command_fail(
+			    cmd, JSONRPC2_INVALID_PARAMS,
+			    "Invalid argument for %s: element "
+			    "\"%s\" at position %zu is not a string",
+			    name, json_strdup(tmpctx, buffer, curr), i);
+		}
+		(*strings)[i] = json_strdup(tmpctx, buffer, curr);
```
rustyrussell, line 529: tmpctx? This needs to be `*strings`.

common/json_tok.c
```diff
@@ -501,3 +501,58 @@ struct command_result *param_psbt(struct command *cmd,
...
+		return res;
+
+	for (size_t i = 0; i < tal_count(*strings); i++) {
+		const char *curr = (*strings)[i];
+		if (!tal_strreg(tmpctx, curr, re)) {
+			return command_fail(
+			    cmd, JSONRPC2_INVALID_PARAMS,
+			    "Invalid element in %s: element %zu is not a valid "
+			    "outpoint \"%s\"",
+			    name, i, curr);
```
rustyrussell, line 554: Yesh, can we implement json_to_outpoint and use this instead?

common/json_tok.h
```diff
@@ -172,4 +172,24 @@ struct command_result *param_psbt(struct command *cmd,
...
+ * Parse a list of `txid:output` outpoint strings.
+ *
+ * The elements are kept as strings in case we want to forward them,
+ * however they are verified to be syntactically valid.
+ */
+struct command_result *param_string_outpoint_arr(struct command *cmd,
+						 const char *name,
+						 const char *buffer,
+						 const jsmntok_t *tok,
+						 const char ***strings);
```
rustyrussell, line 194: Despite the inefficiency, I prefer converting rather than leaving strings.  Maybe we need a struct outpoint as a separate patch though.

rustyrussell (approved):
I will fix the others as a separate series.

Ack 6c308bfee1c76b6060bed0081bc5a28837348156

common/json_tok.h
cdecker: I was unsure whether the added effort and conversions would cause some pushback, but generally I agree that we should always verify as much as possible by doing all conversions, even when doing passthru of arguments (which caused this issue in the first place).

cdecker:
Ok, took two steps back and one forward: we now have a `struct bitcoin_outpoint` (distinct from `outpoiint` in wallet which also contains metadata used when spending), `json_to_outpoint`  and `param_outpoint_arr`. We now fully parse the `txprepare` utxos and report errors correctly.

cdecker:
Couple of force-pushes to remove some drive-by fixes.

bitcoin/tx.h
```diff
@@ -21,6 +21,12 @@ struct wally_psbt;
 struct bitcoin_txid {
 	struct sha256_double shad;
 };
+
+struct bitcoin_outpoint {
+	struct bitcoin_txid txid;
+	u16 n;
```
rustyrussell, line 27: This is a u32 in bitcoin AFAICT?

rustyrussell:
Added a couple of cleanups for you feedback

PR #4260, opened 2020-12-04
bitcoin/chainparams.c: increase max_funding and max_payment

PapauloGamerOfc:
When I tried to create an invoice with a high value just for testing, I received a warning saying that the maximum payment was 4294967295 millisatoshis. This PR is to change it, as it is a very low value, and cannot be used to buy expensive things, such as computers.

### DIFF
--- a/bitcoin/chainparams.c
+++ b/bitcoin/chainparams.c
@@ -25,6 +25,17 @@ static u8 liquid_regtest_fee_asset[] = {
     0x05, 0x71, 0x49, 0x9c, 0x03, 0x62, 0x8a, 0x38, 0x51, 0xb8, 0xce,
 };
 
+static const u64 SAT_PER_COIN = 100 * 1000 * 1000;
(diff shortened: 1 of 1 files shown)


PapauloGamerOfc:
The Travis CI bot is failing on every PR!

cdecker:
NACK

These parameters are directly derived from the specification and modifying them unilaterally means we'd be breaking compatibility with the specification, other implementations and even with our earlier releases. If you want channels that are larger than the `2^24` satoshis, you must make sure that both endpoints support `option_support_large_channel`, a.k.a. wumbo. If that's the case we will actually ignore `.max_funding`.

As for the `.max_payment` that is per HTLC, and with MPP you can transfer many times that amount in a single payment. The limit here comes out to ~800USD per HTLC / part so with 22 HTLCs you can already transfer a full Bitcoin worth, and have a much higher success probability than attempting to pipe that through a single channel.

cdecker:
Also you need to `#define` these constants, since we need const-initializers for this statically defined array.

PR #4262, opened 2020-12-05
peer_control: Implement `listchannelfee` command

### LINKED ISSUE #4221: Add `getchannelfees` command

0e319b6:
This PR implements a `listchannelfee` command to complement `setchannelfee`.  Code is mostly cribbed from surrounding functions.

Fixes #4221.

Does not yet implement any test or documentation, so it is a draft PR for now.  Also I have not at all tried the pyln changes so careful review there would be good.

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -1093,6 +1093,18 @@ class LightningRpc(UnixDomainSocketRpc):
 
         return _sendpay(route, payment_hash, *args, **kwargs)
 
+    def listchannelfee(self, id=None):
+        """
+        Get local routing fees for a channel/peer {id} (or all channels if {id} is not specified).
+        """
+        if id is not None:
+            payload = {
+                "id": id
+            }
+            return self.call("listchannelfee", payload)
+        else:
+            return self.call("listchannelfee")
+
     def setchannelfee(self, id, base=None, ppm=None):
         """
         Set routing fees for a channel/peer {id} (or 'all'). {base} is a value in millisatoshi
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1914,6 +1914,110 @@ static const struct json_command waitblockheight_command = {
 };
 AUTODATA(json_command, &waitblockheight_command);
 
+static struct command_result *param_channel(struct command *cmd,
+					     const char *name,
+					     const char *buffer,
+					     const jsmntok_t *tok,
+					     struct channel **channel)
+{
+	struct command_result *res;
+	struct peer *peer;
+
+	/* Find channel by peer_id */
+	peer = peer_from_json(cmd->ld, buffer, tok);
+	if (peer) {
+		*channel = peer_active_channel(peer);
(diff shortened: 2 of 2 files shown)


0e319b6:
@niftynei It's not clear to me how concerned I should be about CI failures.  On this PR, some configs pass and some fail, like [Job #10630.25](https://travis-ci.org/github/ElementsProject/lightning/jobs/747796143) with a linker error:
```
/home/travis/build/ElementsProject/lightning/lightningd/test/../peer_control.c:1983: undefined reference to `json_add_amount_msat_only'

/home/travis/build/ElementsProject/lightning/lightningd/test/../peer_control.c:2001: undefined reference to `json_add_amount_msat_only'
```

But from what I understand, `json_add_amount_msat_only` is defined in `common/json_helpers.c` right next to the other helper functions, which don't cause a linker error.  I'm not sure why other symbols would resolve but not that one.  So maybe this is something known broken with that build configuration.  Further, many of the failures appear to be due to `VALGRIND=1` being set, which AFAIK always fails.

So, can you please point out to me which of these failures I should be concerned about and which are spurious?  Thanks.

m-schmoock:
Also, rebase and check for CI test errors (e.g. `peer_control.c:2001: undefined reference to json_add_amount_msat_only`)

rustyrussell:
Since listpeerchannels does this, I'm closing.

Feel free to open a new PR if someone wants this!

PR #4264, opened 2020-12-06
Fix freebsd valgrind err

rustyrussell:
As seen in #4222 (thanks @laanwj !)

Changelog-None


cdecker (approved):
ACK b3e3552b02fdf3bb2365a69600985c344734fe5f

PR #4268, opened 2020-12-08
Enhance common/sphinx.c to support variable lengths

rustyrussell:
For now it still needs to be 1366 bytes, but the caller is now responsible for that.  This doesn't change anything yet: the onion_message code will use it, but changing that now would conflict badly with the code move in the offers branch.

The first few patches are optimizations, but they are really just to make it easier to avoid static-sized buffers.

Changelog-None


common/sphinx.c
```diff
@@ -513,14 +531,16 @@ struct onionpacket *create_onionpacket(
 	/* Note that this is just hop_payloads: the rest of the packet is
 	 * overwritten below or above anyway. */
 	subkey_from_hmac("pad", sp->session_key, &padkey);
-	generate_cipher_stream(packet->routinginfo, &padkey, ROUTING_INFO_SIZE);
+	generate_cipher_stream(packet->routinginfo, &padkey, fixed_size);
 
-	generate_header_padding(filler, sizeof(filler), sp, params);
+	filler = tal_arr(tmpctx, u8, fillerSize);
+	generate_header_padding(filler, tal_bytelen(filler), fixed_size, sp, params);
```
cdecker, line 534: We could move the allocation into `generate_header_padding`.

cdecker (commented):
ACK 65ea5c49d77d5f871c9d155e11a467c6bab6a791

PR #4257, opened 2020-12-04
Offers: Part 3.  Offers and the plugins who love them!

rustyrussell:
This contains plugins both for creating offers and servicing invoice requests / invoices associated with them, and sending requests based on others' offers.

Changelog-None

PR #4271, opened 2020-12-09
Offers Part 4: send_invoice offers, allow messages from non-channel peers, documentation and API tweaks

rustyrussell:
(based on #4257)

Changelog-None: experimental work


cdecker:
Clean rebase on top of `master` now that #4257 was merged ([rangediff][rd])

[rd]: http://46.101.246.115:8888/ElementsProject/lightning/pull/4271/range_diff/a9af414c670ad8878cb9530652f459727813fbe0..2bbd1046c76f227e824752b972a3fddecf0c86bb

plugins/fetchinvoice.c
```diff
@@ -620,6 +611,31 @@ static struct command_result *send_message(struct command *cmd,
...
+{
+	struct json_out *details = json_out_new(sent);
+
+	json_out_addstr(details, "invstring", invoice_encode(tmpctx, sent->inv));
+	/* This will free sent! */
+	discard_result(command_done_err(sent->cmd, OFFER_TIMEOUT,
+					"Timeout waiting for response"
+					" (but use waitinvoice if invoice_timeout"
+					" was greater)",
+					details));
```
cdecker, line 649: This is pretty unergonomic, better would be to `waitinvoice` internally if we know that there should be a payment coming.

cdecker: We can use the CLI notifications support to deliver these intermediate results.

plugins/fetchinvoice.c
```diff
@@ -851,6 +867,54 @@ static struct command_result *json_fetchinvoice(struct command *cmd,
 	return send_outreq(cmd->plugin, req);
 }
 
+/* FIXME: Using a hook here is not ideal: technically it doesn't mean
```
cdecker, line 948: It's a notification which doesn't block and supports multiple subscribers unlike some hooks.

plugins/fetchinvoice.c
```diff
@@ -851,6 +867,54 @@ static struct command_result *json_fetchinvoice(struct command *cmd,
...
+	list_for_each(&sent_list, i, list) {
+		struct json_stream *out;
+
+		if (!i->inv)
+			continue;
+		if (!preimage_eq(&preimage, &i->inv_preimage))
+			continue;
+
+		/* It was paid!  Success. */
+		/* FIXME: Return as per waitinvoice */
```
cdecker, line 984: We could just call `waitinvoice` since that's supposed to return immediately on paid invoices, and it'd allow us to avoid copy-pasting details here.

plugins/fetchinvoice.c
```diff
@@ -1105,6 +1175,34 @@ static struct command_result *json_sendinvoice(struct command *cmd,
...
+	sent->inv->timestamp = tal(sent->inv, u64);
+	*sent->inv->timestamp = time_now().ts.tv_sec;
+
+	/* If they don't specify an invoice_timeout, make it the same as we're
+	 * prepare to wait. */
+	if (!invoice_timeout)
+		invoice_timeout = timeout;
+	else if (*invoice_timeout < *timeout)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "invoice_timeout %u must be >= timeout %u",
```
cdecker, line 1269: `invoice_timeout %u must be >= timeout %u` -> `invoice_timeout %u must be >= offer timeout %u`

plugins/offers.c
```diff
@@ -140,7 +140,7 @@ static const struct plugin_command commands[] = {
 	    "offer",
 	    "payment",
 	    "Create an offer",
-            "Create an offer for invoices of {amount} with {destination}, optional {vendor}, {quantity_min}, {quantity_max}, {absolute_expiry}, {recurrence}, {recurrence_base}, {recurrence_paywindow}, {recurrence_limit} and {single_use}",
+            "Create an offer for invoices of {amount} with {description}, optional {vendor}, {quantity_min}, {quantity_max}, {absolute_expiry}, {recurrence}, {recurrence_base}, {recurrence_paywindow}, {recurrence_limit} and {single_use}",
```
cdecker: Drive-by fix that should be in an earlier commit.

doc/lightning-fetchinvoice.7.md
```diff
@@ -0,0 +1,90 @@
...
+DESCRIPTION
+-----------
+
+The **fetchinvoice** RPC command contacts the issuer of an *offer* to get
+an actual invoice that can be paid.  It highlights any changes between the
+offer and the returned invoice.
+
+The offer must not contain *send_invoice*; see lightning-sendinvoice(7).
+
+*msatoshi* is optional: it is required if the *offer* does not specify
```
cdecker, line 20: `is optional: it is required` is contradictory -> `*msatoshi* is required if the *offer* does not specify, otherwise it MUST be omitted`

cdecker: Same for the fields below.

doc/lightning-offer.7.md
```diff
@@ -6,15 +6,14 @@ SYNOPSIS
 
 *EXPERIMENTAL_FEATURES only*
 
-**offer** *amount* *description* \[*send_invoice*\] \[*label*\] \[*vendor*\] \[*quantity_min*\] \[*quantity_max*\] \[*absolute_expiry*\] \[*recurrence*\] \[*recurrence_base*\] \[*recurrence_paywindow*\] \[*recurrence_limit*\] \[*refund_for*\] \[*single_use*\]
+**offer** *amount* *description* \[*vendor*\] \[*label*\] \[*quantity_min*\] \[*quantity_max*\] \[*absolute_expiry*\] \[*recurrence*\] \[*recurrence_base*\] \[*recurrence_paywindow*\] \[*recurrence_limit*\] \[*single_use*\]
```
cdecker, line 9: Why was `vendor` swapped?

I totally agree btw that `offer` was overloaded. We could make the distinction even clearer by calling an offer and subsequent invoice payment either a `push payment` or a `pull payment` :-)

plugins/offers.c
```diff
@@ -111,6 +119,552 @@ static const struct plugin_hook hooks[] = {
...
+					       invreq->recurrence_signature))
+			json_add_string(js, "warning_invoice_request_invalid_recurrence_signature",
+					"Bad recurrence_signature");
+	} else if (invreq->recurrence_counter) {
+		json_add_string(js, "warning_invoice_request_missing_recurrence_signature",
+				"invoice_request requires recurrence_signature");
+	}
+}
+
+static struct command_result *json_decode(struct command *cmd,
```
cdecker, line 643: This needs to move into the core, independently from offers.

cdecker (changes requested):
NACK on moving `onion_message` handling into `gossipd`: that daemon is
completely unrelated and doesn't even have a direct connection to the
peer at any point in time. If we just need a convenient place to
scatter the messages to peers we should consider merging the daemons
that are already talking to the peer, rather than moving per-peer
functionality into `gossipd`. Then again the same is true for `ping`,
`pong`, which I think we should move into the per-peer daemons as
well.

rustyrussell:
> NACK on moving `onion_message` handling into `gossipd`: that daemon is
> completely unrelated and doesn't even have a direct connection to the
> peer at any point in time. If we just need a convenient place to
> scatter the messages to peers we should consider merging the daemons
> that are already talking to the peer, rather than moving per-peer
> functionality into `gossipd`. Then again the same is true for `ping`,
> `pong`, which I think we should move into the per-peer daemons as
> well.

Yeah, it's a little ugly.  I copied the logic for ping, for simplicity.  Reworking that logic is a much bigger change, though.

The idea of merging the daemons is interesting; their roles with dual funding, and splicing, will become less distinct than they are now (one reason I like the separation is that they have distinct HSM needs, but that changes quite a bit with splicing).

We could put the message handling in each daemon (i.e. in common/) which would be a much less invasive change.

plugins/fetchinvoice.c
```diff
@@ -372,6 +371,54 @@ static struct command_result *recv_onion_message(struct command *cmd,
...
+						 const jsmntok_t *params)
+{
+	const jsmntok_t *om, *blindingtok;
+	struct sent *sent;
+	struct pubkey blinding;
+	struct command_result *err;
+
+	plugin_log(cmd->plugin, LOG_INFORM, "Received onion message: %.*s",
+		   json_tok_full_len(params),
+		   json_tok_full(buf, params));
```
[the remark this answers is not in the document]
rustyrussell: Agreed.

plugins/fetchinvoice.c
rustyrussell: The idea was that you can explicitly set invoice timeout > the waiting timeout.  But in practice it's a little silly, since we don't do retransmission or anything.  So I'll move it back to a single timeout value.

plugins/fetchinvoice.c
rustyrussell: Yeah, but this gets fired before it hit the db which can cause an issue in theory.

plugins/fetchinvoice.c
rustyrussell: Ah, good idea!

plugins/offers.c
rustyrussell: But that commit has already got into master :(

cdecker:
> Yeah, it's a little ugly. I copied the logic for ping, for simplicity. Reworking that logic is a much bigger change, though.

Yep, it's going to be a gargantuan refactoring, that'll break all in-flight PRs while merging. I'd like to see a `peerd` that is the combination of the per-peer daemons. We could start by merging all daemons with an async loop, rename `main` and select the correct `main` depending on what `init` message we get (from there on the daemons behave as if two different binaries were started). We'd then still kill and start the same binary multiple times as we progress through the states, but that'd be simpler to just skip if we already are in the correct daemon :-) Mixing sync and async daemons is more challenging, but should be doable as well. The last step then is to merge all the `io_loop`, deduplicate functionality and merging structs that represent the same concepts in different daemons.

> We could put the message handling in each daemon (i.e. in common/) which would be a much less invasive change.

With the top-down approach above we could technically skip this extraction into `common`, but would need to see how many clashes it produces that need to be addressed first.

plugins/fetchinvoice.c
cdecker: Ah I see the issue. But aren't notifications delivered only once we hand off to `io_loop` again? That'd mean we always commit before the notification is sent to the plugin, which is what we want.

doc/lightning-offer.7.md
rustyrussell: vendor is associated with description (they're both strings the user sees which talk about the item).  Since API isn't fixed it makes sense to reorder them while we rework the args anyway.

Naming these things is really hard!  Currently:

A: offer
B: fetchinvoice, pay.

A: offerout
B: sendinvoice
A: (automatically) pay

Not sure how we'd use push-pull here?

plugins/offers.c
rustyrussell: Well, it could also be its own plugin.  It doesn't need anything in core, but I'm in two minds here.

doc/lightning-offer.7.md
cdecker: Right, sounded better in my head.

A: offerpull (user A is offering to pull some funds from B)
B: fetchinvoice

A: offerpush (user A is offering to push some funds to B)
B: sendinvoice
A: pay

But we can always add aliasesat a later stage.

plugins/fetchinvoice.c
rustyrussell: Ah, yes, your point about it being a notification not a hook!

rustyrussell:
Trivial rebase, and appended fixes...

Didn't move decode into core, but that can definitely be done in follow up PR.

plugins/offers.c
cdecker: Ok, we can reconcile this later again.

cdecker:
Rebased and squashed. Merging once CI is happy :+1:

ACK https://github.com/ElementsProject/lightning/pull/4271/commits/2b10f2712a8c4e7356664d5f026729fb2e1969b9

cdecker:
Had to fix the docs which were still referencing `invoice_timeout` which is no longer a thing.

ACK https://github.com/ElementsProject/lightning/pull/4271/commits/0d8efcbc4f25255d78538e690f43cb0fc0faceaf

cdecker (approved):

cdecker:
Rebased on top of #4321 in order to shut `test_lockup_drain` up...

PR #4273, opened 2020-12-10
pyln: fix div with Millisatoshi itself

m-schmoock:
Current implementation does not allow dividing two msat's to get the ratio between them.
It fails with `unsupported operand type(s)`:
```python
from pyln.client import Millisatoshi
foo = Millisatoshi(42) / Millisatoshi(2)
```

- This PR changes this to return the `float` ratio between the two `msat's` for `__truediv__` and an `int` for `__floordiv__`. 
- Also adds an exception and a check that `Millisatoshi(float)` initialization is currently not allowed, except for whole float numbers, e.g. `42.0` this is currently required by some existing testcases: e.g. `tests/test_wallet.py::test_txprepare`
See https://github.com/ElementsProject/lightning/pull/4273#discussion_r540369093).

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -176,13 +176,19 @@ class Millisatoshi:
     def __sub__(self, other: 'Millisatoshi') -> 'Millisatoshi':
         return Millisatoshi(int(self) - int(other))
 
-    def __mul__(self, other: int) -> 'Millisatoshi':
+    def __mul__(self, other: Union[int, float]) -> 'Millisatoshi':
+        if isinstance(other, Millisatoshi):
+            raise TypeError("Resulting unit msat^2 is not supported")
         return Millisatoshi(floor(self.millisatoshis * other))
 
-    def __truediv__(self, other: Union[int, float]) -> 'Millisatoshi':
+    def __truediv__(self, other: Union[int, float, 'Millisatoshi']) -> Union['Millisatoshi', float]:
+        if isinstance(other, Millisatoshi):
+            return self.millisatoshis / other.millisatoshis
         return Millisatoshi(floor(self.millisatoshis / other))
 
-    def __floordiv__(self, other: Union[int, float]) -> 'Millisatoshi':
+    def __floordiv__(self, other: Union[int, float, 'Millisatoshi']) -> Union['Millisatoshi', int]:
+        if isinstance(other, Millisatoshi):
+            return self.millisatoshis // other.millisatoshis
         return Millisatoshi(floor(self.millisatoshis // float(other)))
 
     def __mod__(self, other: Union[float, int]) -> 'Millisatoshi':
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -68,6 +68,10 @@ class Millisatoshi:
 
         elif int(v) == v:
             self.millisatoshis = int(v)
+
+        elif isinstance(v, float):
+            raise TypeError("Millisatoshi by float is currently not supported")
+
         else:
             raise TypeError(
                 "Millisatoshi must be string with msat/sat/btc suffix or int"
--- a/contrib/pyln-client/tests/test_units.py
+++ b/contrib/pyln-client/tests/test_units.py
@@ -312,3 +312,26 @@ def test_div():
     assert amount == 14
     amount = Millisatoshi(42) // Millisatoshi(4)
     assert amount == 10
+
+
+def test_init():
+    # Note: Ongoing Discussion, hence the `with pytest.raises`.
+    # https://github.com/ElementsProject/lightning/pull/4273#discussion_r540369093
+    #
+    # Initialization with a float should be possible:
+    # Millisatoshi(5) / 2 currently works, and removes the half msat.
+    # So Millisatoshi(5 / 2) should be the same.
+    amount = Millisatoshi(5) / 2
+    assert amount == Millisatoshi(2)
+    with pytest.raises(TypeError, match="Millisatoshi by float is currently not supported"):
+        assert amount == Millisatoshi(5 / 2)
+
+    ratio = Millisatoshi(8) / Millisatoshi(5)
+    assert isinstance(ratio, float)
+    with pytest.raises(TypeError, match="Millisatoshi by float is currently not supported"):
+        assert Millisatoshi(ratio) == Millisatoshi(8 / 5)
+
+    # Check that init by a round float is allowed.
+    # Required by some existing tests: tests/test_wallet.py::test_txprepare
+    amount = Millisatoshi(42.0)
+    assert amount == 42
--- a/contrib/pyln-client/tests/test_units.py
+++ b/contrib/pyln-client/tests/test_units.py
@@ -269,3 +269,46 @@ def test_nonegative():
         Millisatoshi("1sat") // -1
     with pytest.raises(ValueError, match='Millisatoshi must be >= 0'):
         Millisatoshi("1btc") // -1
+
+
+def test_mul():
+    # msat * num := msat
+    amount = Millisatoshi(21) * 2
+    assert isinstance(amount, Millisatoshi)
+    assert amount == Millisatoshi(42)
+    amount = Millisatoshi(21) * 2.5
+    assert amount == Millisatoshi(52)
+
+    # msat * msat := msat^2  (which is not supported)
+    with pytest.raises(TypeError, match="not supported"):
+        Millisatoshi(21) * Millisatoshi(2)
+
+
+def test_div():
+    # msat / num := msat
+    amount = Millisatoshi(42) / 2
+    assert isinstance(amount, Millisatoshi)
+    assert amount == Millisatoshi(21)
+    amount = Millisatoshi(42) / 2.6
+    assert amount == Millisatoshi(16)
+
+    # msat / msat := num   (float ratio)
+    amount = Millisatoshi(42) / Millisatoshi(2)
+    assert isinstance(amount, float)
+    assert amount == 21.0
+    amount = Millisatoshi(8) / Millisatoshi(5)
+    assert amount == 1.6
+
+    # msat // num := msat
+    amount = Millisatoshi(42) // 2
+    assert isinstance(amount, Millisatoshi)
+    assert amount == Millisatoshi(21)
+
+    # msat // msat := num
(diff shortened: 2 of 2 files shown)


contrib/pyln-client/tests/test_units.py
```diff
@@ -269,3 +269,17 @@ def test_nonegative():
...
+    assert amount == Millisatoshi(42)
+    amount = Millisatoshi(21) * Millisatoshi(2)
+    assert amount == Millisatoshi(42)
+
+
+def test_div():
+    amount = Millisatoshi(42) / 2
+    assert amount == Millisatoshi(21)
+    amount = Millisatoshi(42) / Millisatoshi(2)
+    assert amount == Millisatoshi(21)
```
gallizoltan, line 285: The division between Millisatoshis makes sense, but can we check the type here? For example in this context:
```
our_liquidity = Millisatoshi(5)
channel_size = Millisatoshi(10)
ratio = our_liquidity/channel_size
```
The return type should be a `float`, not a `Millisatoshi`, because it is a ratio, not an amount.

m-schmoock: Then what about multiplication? Keep in mind that a 'division with x' is the same as 'multiplication with 1/x' ... So to be consistent, we need to always return a Decimal/Float something (not Millisatoshi) for __mul__ __truediv__ and __floordiv__

m-schmoock: Meaning  Millisatoshi(5) * 3 := 15    (not msat)

gallizoltan: Millisatoshi(15) / Millisatoshi(5) := 3 (not msat)
So I multiple both side of the equation with Millisatoshi(5), and I get:
Millisatoshi(15) := 3 * Millisatoshi(5)
In your example:
Millisatoshi(5) * 3 := Millisatoshi(15)

m-schmoock: Sorry, I'm not quite getting it. In your opinion what should be the result of
`Millisatoshi(21) * Millisatoshi(2)`  be? Should it be just `42` or `Millisatoshi(42)? Currently it fails with an exception...

gallizoltan: The multiplication between Millisatoshis is difficult. In theory, it should be Millisatoshi square, Millisatoshi**2, Millisatoshi^2 or something like that.
I'm not sure, we need multiplication, maybe the exception is good. But I don't know.

gallizoltan: Also, can we write some more test? This should not raise exceptions:
- x = min(10, Millisatoshi(10))
- x = min(5 / 2, Millisatoshi(10))
- x = Millisatoshi(5 / 2)
- x = Millisatoshi(5) / 2

m-schmoock: > This should not raise exceptions:
> * x = Millisatoshi(5 / 2)

Not too sure about this. We ask it to initialize from `2.5` which, if no exception is raised, would likely remove the half msat. 

Maybe this whole class needs a rewrite/different approach. @cdecker already mentioned a couple of days ago that we might want to have a Decimal internally `self.millisatoshis` and only round down/floor when we do `self.to_json()`. This will keep precision when we do multiple div/mul operations after another... In this case we could go for the 2.5 in the constuctor as well :D

m-schmoock: `Millisatoshi(5) / 2` currently works. and the first two examples: `min(10, Millisatoshi(10))` and `min(5 / 2, Millisatoshi(10))`) should work, yes

gallizoltan: `Millisatoshi(5) / 2` currently works, and removes the half msat. So, I think `Millisatoshi(5 / 2)` should be the same.

m-schmoock: Now that you put it this way, Agreed.

cdecker:
Hm, no I think we should not allow msat^2: if both operands are msat, multiplication doesn't work, because msat^2 doesn't have a well-defined meaning. And as @gallizoltan points out:

 -  dividing msat by an int is msat
 - dividing msat by msat is int
 - dividing int by msat is not representable in this case (we don't have an msat^-1 type) and unlikely to be useful

m-schmoock:
@gallizoltan @cdecker
After your comments:
- Enabled msat division `msat / msat := float` 
- Same for __floordiv__   `msat // msat := int` 
- Allows initialization by float or decimal  (https://github.com/ElementsProject/lightning/pull/4273#discussion_r540369093)
- Updated tests

@gallizoltan I'm not too sure about the min(int, msat) comparisons, because if I allow them by changing the `__le__  __lt__ 
 __ge__  ...` methods , the result will be that we don't have a defined return type on `min`/`max` functions:
```python
min(10, Millisatoshi(10))     # := 10
min(Millisatoshi(10), 10)     # := 10msat
```

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -43,7 +43,7 @@ class Millisatoshi:
     Many JSON API fields are expressed in millisatoshis: these automatically
     get turned into Millisatoshi types. Converts to and from int.
     """
-    def __init__(self, v: Union[int, str, Decimal]):
+    def __init__(self, v: Union[int, str, float, Decimal]):
```
cdecker, line 46: Not sure we want `float` anywhere neat monetary amounts. It's a pretty big source of errors in financial software, and asking the user to explicitly round is safer imho.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -68,6 +68,10 @@ def __init__(self, v: Union[int, str, Decimal]):
 
         elif int(v) == v:
             self.millisatoshis = int(v)
+
+        elif isinstance(v, float) or isinstance(v, Decimal):
+            self.millisatoshis = int(v)
```
cdecker, line 73: Hm, makes me wonder if we should just go the full mile and maintain the internal value as `Decimal` and round on read, with `.round(), `.floor()` and `.ceil()` accessors so the caller can specify what is correct in their context.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -176,13 +180,17 @@ def __add__(self, other: 'Millisatoshi') -> 'Millisatoshi':
...
-    def __mul__(self, other: int) -> 'Millisatoshi':
+    def __mul__(self, other: Union[int, float]) -> 'Millisatoshi':
         return Millisatoshi(floor(self.millisatoshis * other))
 
     def __truediv__(self, other: Union[int, float]) -> 'Millisatoshi':
+        if isinstance(other, Millisatoshi):
+            return self.millisatoshis / other.millisatoshis
         return Millisatoshi(floor(self.millisatoshis / other))
 
     def __floordiv__(self, other: Union[int, float]) -> 'Millisatoshi':
```
cdecker, line 191:
```suggestion
    def __floordiv__(self, other: Union[int, float]) -> Union['Millisatoshi', int]:
```

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -176,13 +180,17 @@ def __add__(self, other: 'Millisatoshi') -> 'Millisatoshi':
     def __sub__(self, other: 'Millisatoshi') -> 'Millisatoshi':
         return Millisatoshi(int(self) - int(other))
 
-    def __mul__(self, other: int) -> 'Millisatoshi':
+    def __mul__(self, other: Union[int, float]) -> 'Millisatoshi':
```
cdecker, line 183: Should probably raise an exception if we try to multiply msat with msat. Typing checks are not performed at runtime and devs often don't use `mypy` yet.

cdecker (commented):
> @gallizoltan I'm not too sure about the min(int, msat) comparisons, because if I allow them by changing the `__le__ __lt__ __ge__ ...` methods , the result will be that we don't have a defined return type on `min`/`max` functions:

Comparison between unit-less amounts and amounts with units is not well-defined. Does the `int` represent msat, btc, USD? We should not allow it due to safety concerns, because whatever convention we chose is likely to cause confusion at some point.

Excellent PR otherwise, just some minor syntactical things I found while reviewing the changes.

contrib/pyln-client/pyln/client/lightning.py
m-schmoock: @cdecker Yes I thought that too (first), but on the same topic we can decide to allow or disallow division that rounds down / floors to int:   `8msat / 5   :=  int(1)`. I find @gallizoltan is right in that this is inconsistent.

contrib/pyln-client/pyln/client/lightning.py
m-schmoock: Yup, good catch, will add that

gallizoltan:
> Comparison between unit-less amounts and amounts with units is not well-defined. Does the `int` represent msat, btc, USD? We should not allow it due to safety concerns, because whatever convention we chose is likely to cause confusion at some point.

@cdecker My problem is `channel["msatoshi_total"]` is `int`, and I cannot compare that with some `Millisatoshi`. I don't know what would be the cleanest solution here. I have to cast, which is ugly.

cdecker:
> > Comparison between unit-less amounts and amounts with units is not well-defined. Does the `int` represent msat, btc, USD? We should not allow it due to safety concerns, because whatever convention we chose is likely to cause confusion at some point.
> 
> @cdecker My problem is `channel["msatoshi_total"]` is `int`, and I cannot compare that with some `Millisatoshi`. I don't know what would be the cleanest solution here. I have to cast, which is ugly.

Use `total_msat` which gets automatically parsed into a `Millisatoshi` instance? The other field is there for backward compatibility only, and should not be relied upon.

contrib/pyln-client/pyln/client/lightning.py
m-schmoock: t.b.d. in a separate PR. I remove the init by float for now ...

gallizoltan:
> Use `total_msat` which gets automatically parsed into a `Millisatoshi` instance? The other field is there for backward compatibility only, and should not be relied upon.

@cdecker I didn't know that. Thank you very much, it works for me!

cdecker:
> @cdecker This is done from my side.
> Note: the `VALGRIND=1` testflake on `tests/test_pay.py::test_lockup_drain` is unrelated to this PR.

Yep, fixed it in my cleanups branch ^^

ACK https://github.com/ElementsProject/lightning/pull/4273/commits/5d4afe370708291a8af2b121d28eb62864b559c3

PR #4275, opened 2020-12-10
make: have `git submodule update` be recursive

niftynei:
libwally added libsecp256k1 as a submodule, so we need to recursively
update the submodules now

Changelog-None.


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4275/commits/eca437cea81c6a12e9ec3abcb534731ceb42997f

PR #4276, opened 2020-12-11
tests: Update mocks after breaking them via interleaved merges

cdecker:
Seems that I broke the mocks by merging two PRs that both modified the tests, thus overwriting each others set of mocks.

Changelog-None


cdecker:
https://github.com/ElementsProject/lightning/pull/4276/commits/2eb5c19b6c60add346b943606e1ea266c79e9010

PR #4280, opened 2020-12-16
Offers: the bonus bits!

rustyrussell:
Based on #4271, this is an ongoing effort to fix the FIXMEs.

See "onion_message: support variable-length onion messages" onwards (last 4 commits)

Changelog-None: experimental stuff...

### DIFF
--- a/lightningd/onion_message.c
+++ b/lightningd/onion_message.c
@@ -138,11 +138,11 @@ void handle_onionmsg_forward(struct lightningd *ld, const u8 *msg)
 	struct short_channel_id *next_scid;
 	struct node_id *next_node;
 	struct pubkey *next_blinding;
-	u8 onion[TOTAL_PACKET_SIZE(ROUTING_INFO_SIZE)];
+	u8 *onion;
 
 	if (!fromwire_gossipd_got_onionmsg_forward(msg, msg, &next_scid,
 						   &next_node,
-						   &next_blinding, onion)) {
+						   &next_blinding, &onion)) {
 		log_broken(ld->log, "bad got_onionmsg_forward: %s",
 			   tal_hex(tmpctx, msg));
 		return;
@@ -398,6 +398,7 @@ static struct command_result *json_send_onion_message(struct command *cmd,
 	struct onionpacket *op;
 	struct secret *path_secrets;
 	struct node_id first_id;
+	size_t onion_size;
 
 	if (!param(cmd, buffer, params,
 		   p_req("hops", param_hops, &hops),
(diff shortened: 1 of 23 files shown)


cdecker:
Looks alright, but the parent PR needs to be adjusted before this can get a verdict.

rustyrussell:
Rebased on latest.

lightningd/onion_message.c
```diff
@@ -432,7 +432,12 @@ static struct command_result *json_send_onion_message(struct command *cmd,
 				hops[i].rawtlv, tal_bytelen(hops[i].rawtlv));
 		sphinx_add_hop(sphinx_path, &hops[i].id, take(tlv_with_len));
 	}
+	/* BOLT-offers #4:
+	 * - SHOULD set `len` to 1366 or 32834.
+	 */
 	op = create_onionpacket(tmpctx, sphinx_path, ROUTING_INFO_SIZE, &path_secrets);
+	if (!op)
+		op = create_onionpacket(tmpctx, sphinx_path, 32768, &path_secrets);
```
cdecker, line 440: Technically we could compute the payload size using `sphinx_path_payloads_size` to determine which format will be required:

https://github.com/ElementsProject/lightning/blob/25463bd5a755d9617c262dbbe1827d1097b0d969/common/sphinx.c#L491-L492

This would save us two allocations and de-allocations:

https://github.com/ElementsProject/lightning/blob/25463bd5a755d9617c262dbbe1827d1097b0d969/common/sphinx.c#L502-L506

plugins/offers_invreq_hook.c
```diff
@@ -476,10 +468,36 @@ static struct command_result *invreq_base_amount_simple(struct command *cmd,
...
+	 *     - MUST fail the request if its `amount` is less than the *base invoice amount*.
+	 *     - MAY fail the request if its `amount` is much greater than the *base invoice
+	 *     amount*.
+	 *     - MUST use the request's `amount` as the *base invoice amount*.
+	 */
+	if (ir->offer->amount && ir->invreq->amount) {
+		if (amount_msat_less(amount_msat(*ir->invreq->amount), base_inv_amount)) {
+			return fail_invreq(cmd, ir, "Amount must be less than %s",
+					   type_to_string(tmpctx, struct amount_msat,
+							  &base_inv_amount));
```
cdecker, line 486: Isn't this inverted? I.e., shouldn't we be allowed to request an invoice with higher amounts than the offer?

plugins/offers_invreq_hook.c
```diff
@@ -476,10 +468,36 @@ static struct command_result *invreq_base_amount_simple(struct command *cmd,
...
+		}
+		/* Easier to divide by 5 and compare, than multiply by 5 */
+		if (amount_msat_greater(amount_msat_div(amount_msat(*ir->invreq->amount), 5),
+					base_inv_amount)) {
+			return fail_invreq(cmd, ir, "Amount vastly exceeds %s",
+					   type_to_string(tmpctx, struct amount_msat,
+							  &base_inv_amount));
+		}
```
cdecker, line 494: Feels a bit like a magic limitation. Is there a BOLT-quote to associate with this?

cdecker (commented):
Two minor comments and an optimization :rocket:

plugins/offers_invreq_hook.c
rustyrussell: Pretty sure this is correct, but I will add a test for it.

The base inv amount is the amount of the offer * quantity.  If the invoice request is < this amount, it's invalid.

plugins/offers_invreq_hook.c
rustyrussell: Yes, it's deliberately vague, but I'll add it.

plugins/offers_invreq_hook.c
rustyrussell: Ah, I do see that the error msg is backwards though! (tests FTW!)

rustyrussell:
Thanks!  Trivial rebase, and two fixups appended based on your feedback.

cdecker:
Fixups folded in, waiting for CI to be green :+1:

ACK https://github.com/ElementsProject/lightning/pull/4280/commits/f43a448ccb2ee662abeb6c9fb6c01764727b5ef6

PR #4284, opened 2020-12-18
Weekly cleanups

cdecker:
Just the usual bunch of drive-by fixes, mostly while trying to get Cirrus-CI to pass.

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -321,7 +321,16 @@ class SimpleBitcoinProxy:
         proxy = BitcoinProxy(btc_conf_file=self.__btc_conf_file__)
 
         def f(*args):
-            return proxy._call(name, *args)
+            logging.debug("Calling {name} with arguments {args}".format(
+                name=name,
+                args=args
+            ))
+            res = proxy._call(name, *args)
+            logging.debug("Result for {name} call: {res}".format(
+                name=name,
+                res=res,
+            ))
+            return res
 
         # Make debuggers show <function bitcoin.rpc.name> rather than <function
         # bitcoin.rpc.<lambda>>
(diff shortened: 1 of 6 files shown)


contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -1038,10 +1040,59 @@ def passes_filters(hmsg, filters):
...
+        with self.lock.acquire(poll_intervall=0.250):
+            # We just got the lock, assume someone else just released it
+            self.current_load = 100
+            while self.load() >= self.target:
+                time.sleep(1)
+
+            delay = time.time() - start_time
+            with open("/tmp/ltest-throttler.csv", "a") as f:
+                f.write("{}, {}, {}, {}\n".format(time.time(), self.load(), self.target, delay))
+            self.current_load = 100  # Back off slightly to avoid triggering right away
```
darosior, line 1084: Read this comment a few times, but i still don't get it :sweat_smile: . What would a `< 75` load would trigger ? And why setting it to `100` after getting each node would avoid the trigger ?

darosior (approved):
LGTM, just a question about the throttler

contrib/pyln-testing/pyln/testing/utils.py
cdecker: The thing is that if we leave the `current_load` at 75 percent a followup call, e.g., another `lightningd` wanting to start, could end up seeing 75% and decide to immediately schedule its startup (maybe the first one hasn't had time to start up and push the load up yet), thus we might end up clearing the queue despite only the first one getting the all-clear. 

Setting this to 100% means the ewma smoothing needs a couple of rounds before allowing the next one to go through, avoiding the flooding scenario above.

darosior: Ok, i had it the other way around, thanks!

darosior (approved):
ACK 336e972

PR #4289, opened 2020-12-20
How to calculate channel fees using the RPC ?

hosiawak:
Rationale: I'm working on a tool for C-Lightning and need to calculate on chain channel opening/closing/other fees. I've added `getrawtransaction` to `bcli.c` that lets me do that. It should come in handy in other situations too.

Please review my code if you have a minute and let me know if you see any issues.

### DIFF
--- a/plugins/bcli.c
+++ b/plugins/bcli.c
@@ -13,6 +13,7 @@
 #include <ccan/tal/grab_file/grab_file.h>
 #include <ccan/tal/path/path.h>
 #include <ccan/tal/str/str.h>
+#include <common/json_stream.h>
 #include <common/json_helpers.h>
 #include <common/memleak.h>
 #include <common/utils.h>
@@ -431,6 +432,26 @@ static struct command_result *process_getblockchaininfo(struct bitcoin_cli *bcli
 	return command_finished(bcli->cmd, response);
 }
 
+static struct command_result *process_getrawtransaction(struct bitcoin_cli *bcli)
+{
+	const jsmntok_t *tokens = NULL;
+	struct json_stream *response;
+
+	tokens = json_parse_simple(bcli->output,
+				   bcli->output, bcli->output_bytes);
+	if (!tokens) {
+		return command_err_bcli_badjson(bcli, "cannot parse");
+	}
+
+	if (tokens[0].type != JSMN_OBJECT) {
+		return command_err_bcli_badjson(bcli, "non-object");
+	}
+
+	response = jsonrpc_stream_success(bcli->cmd);
+
+	json_stream_append(response, bcli->output + tokens[0].start + 1, tokens[0].end - 2);
+	return command_finished(bcli->cmd, response);
+}
 
 struct estimatefees_stash {
 	/* FIXME: We use u64 but lightningd will store them as u32. */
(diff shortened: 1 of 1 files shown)


plugins/bcli.c
```diff
@@ -431,6 +432,26 @@ static struct command_result *process_getblockchaininfo(struct bitcoin_cli *bcli
...
 }
 
+static struct command_result *process_getrawtransaction(struct bitcoin_cli *bcli)
+{
+	const jsmntok_t *tokens = NULL;
+	struct json_stream *response;
+
+	tokens = json_parse_simple(bcli->output,
+				   bcli->output, bcli->output_bytes);
+	if (!tokens) {
```
ZmnSCPxj, line 442: We usually elide `{` `}` for single-statement `if`/`while`/`for`.

plugins/bcli.c
```diff
@@ -959,6 +1002,14 @@ static const struct plugin_command commands[] = {
 		"",
 		getutxout
 	},
+	{
+		"getrawtransaction",
```
ZmnSCPxj, line 1006: Any reason why it needs to go through `lightningd` instead of directly to `bitcoind`?  This would increase effort for #3858.  On the other hand, if #3858 pushes through, then it is reasonable for a plugin to prefer this interface rather than directly calling into `bitcoind`, as the backend might not be `bitcoind` anymore.

hosiawak: The tool I'm working on is web based node dashboard (I haven't found anything decent for C-Lightning). The reason why I'd like to avoid setting up connection details to both `lightningd` and `bitcoind` is better user experience when setting them up. Currently I'm running C-Lightning node on a remote machine whilst the tool is running locally so I'm using `socat` on the remote + `socat` on localhost to get a local socket that will work with [lightning-client for node](https://www.npmjs.com/package/lightning-client). It's not ideal if I have to do this for `bitcoind` as well and expect the users to do the same. 

If #3858 goes through couldn't we disable `getrawtransaction` if it's not available in the backend or cannot be implemented ?

plugins/bcli.c
hosiawak: I mimicked the style from `*process_getblockchaininfo` and `*process_getutxout` etc.

plugins/bcli.c
ZmnSCPxj: What transactions specifically are you extracting?  My understanding is that if the `bitcoind` backend itself is not `txindex=1`, it would not be able to provide arbitrary transactions.  We do not inform the backend of any pubkeys to watch as well (we just download every block as it comes in and iterate over each tx and check for those we are interested in), so without `txindex=1` I am uncertain what a `bitcoind` would be able to provide.

If the transactions you are extracting are those that the LN node wants to track as well, maybe it would be better to expose those in our RPC (which is probably a lot more effort over here in C-Lightning, but feels like the "right thing").

hosiawak: I'm interested in channel opening and closing transactions only and specifically their on chain fees. After your comment I realised that this PR wouldn't work without a `txindex=1` and that I can probably use `getrawblockbyheight` to get the data I want. 

I still don't know how to calculate the value of the total inputs having just their txids and block data though. Does fee calculation still require `txindex=1` in core and there's no way around it currently ?

I noticed that [getblock will return tx fees when run in verbose mode soon](https://github.com/bitcoin/bitcoin/pull/18772) (when it gets merged into core) so maybe I could use that instead and modify `getrawblockbyheight` to accept an optional tx position (index) parameter which will make it return only that transaction data (with calculated fee included because verbosity will be set to 2).

Is it a good idea that has a chance of being accepted ?

cdecker: If we are the funders of the channel we'll have both the funding tx as well as the `outputs` corresponding to the funding's inputs, so for these we could compute them ourselves quite easily. The downside of this is that the data is pretty spotty (only covers our own transactions), but that may match your exact use-case if all you want to do is show users the fees they paid themselves.

hosiawak: @cdecker Yes, I'm interested in showing how much we paid for channel txs. Do you know if it's possible if it's possible to get the outputs corresponding to funding's inputs via the JSON RPC somehow and if it's not possible where I can look in the C-Lightning code and learn how to expose this via the RPC ?

cdecker: We currently don't have a way to list spent outputs which is what you'd require to compute fees. We can easily add a `spent=True` flag to `listfunds` to also include spent outputs.

From there on you'd just
 - call `lightning-cli listtransactions`, stash the result
 - call `lightning-cli listfunds spent=true`

And then merge the two results by:

 - Iterate through transactions
 - For each transaction:
   - Iterate over inputs
   - Look up matching output
   - Compute input amounts, and subtract output amounts.

Adding the flag is a good first issue, and if you'd like to go that way I'll open an issue tracking progress :-)

hosiawak: Thanks @cdecker I'll do that. I guess this PR can be closed then ?

cdecker: Sounds good to me :+1:

I'll open the tracking issue and close this PR then. Would you like to work on the `spent=true` param yourself (in which case I'll earmark the issue with your name)?

hosiawak: @cdecker Yes, I'll pick it up. Thanks.

PR #4293, opened 2020-12-21
Wait for 'FUNDING_LOCKED' in dualopend, not channeld 

niftynei:
Prework for allowing RBF of dual funded channel opens. We need to wait for FUNDING_LOCKED in dualopend, so that handling a request to RBF the channel can transition smoothly back to the 'opening' message flow.

To do this, we move all of the tx_sigs + psbt handling out of channeld, back into dualopend. We also add handling for reconnections, shutdown requests, and waiting for FUNDING_LOCKED to dualopend.

This looks like a big PR (because it is), but most of the stuff is small tweaks to the test suite to have tests pass for dual-funded channel flows. These aren't turned on here; waiting for the move over to dual-funding being a runtime flag (instead of compile flag).

Update: now based on top of #4294.

Changelog-None


tests/test_plugin.py
```diff
@@ -563,18 +563,43 @@ def test_openchannel_hook(node_factory, bitcoind):
...
+            'dust_limit_satoshis': '546000msat',
+            'feerate_per_kw': '7500',
+            'funding_satoshis': '100000000msat',
+            'htlc_minimum_msat': '0msat',
+            'id': l1.info['id'],
+            'max_accepted_htlcs': '483',
+            'max_htlc_value_in_flight_msat': '18446744073709551615msat',
+            'push_msat': '0msat',
+            'to_self_delay': '5',
+        }
```
cdecker, line 598: You can group the common elements into a single dict, and then update with the differences:

```suggestion
    expected = {
            'channel_flags': '1',
            'dust_limit_satoshis': '546000msat',
            'htlc_minimum_msat': '0msat',
            'id': l1.info['id'],
            'max_accepted_htlcs': '483',
            'max_htlc_value_in_flight_msat': '18446744073709551615msat',
            'to_self_delay': '5',
    }
    if l2.config('experimental-dual-fund'):
        # openchannel2 var checks
        expected.update({
            'commitment_feerate_per_kw': '750',
            'feerate_our_max': '150000',
            'feerate_our_min': '1875',
            'funding_feerate_best': '7500',
            'funding_feerate_max': '150000',
            'funding_feerate_min': '1875',
            'locktime': '.*',
            'their_funding': '100000000msat',
        })
    else:
        expected.update({
            'channel_reserve_satoshis': '1000000msat',
            'feerate_per_kw': '7500',
            'funding_satoshis': '100000000msat',
            'push_msat': '0msat',
        })
```

tests/test_plugin.py
```diff
@@ -604,9 +629,9 @@ def test_openchannel_hook_error_handling(node_factory, bitcoind):
     l1.fundwallet(10**6)
 
     # next fundchannel should fail fatal() for l2
-    with pytest.raises(RpcError, match=r'Owning subdaemon openingd died'):
+    with pytest.raises(RpcError, match=r'Owning subdaemon (openingd|dualopend) died'):
         l1.rpc.fundchannel(l2.info['id'], 100004)
-    assert l2.daemon.is_in_log("BROKEN.*Plugin rejected openchannel but also set close_to")
+    assert l2.daemon.is_in_log("BROKEN.*Plugin rejected openchannel2? but also set close_to")
```
cdecker, line 634: Just to make it clear what the `?` applies to:

```suggestion
    assert l2.daemon.is_in_log("BROKEN.*Plugin rejected openchannel[2]? but also set close_to")
```

wallet/db.c
```diff
@@ -668,6 +668,7 @@ static struct migration dbmigrations[] = {
     {SQL("ALTER TABLE invoices ADD COLUMN local_offer_id BLOB DEFAULT NULL REFERENCES offers(offer_id);"), NULL},
     /* A reference into our own offers table, if it was made from one */
     {SQL("ALTER TABLE payments ADD COLUMN local_offer_id BLOB DEFAULT NULL REFERENCES offers(offer_id);"), NULL},
+    {SQL("ALTER TABLE channels ADD tx_sigs_remote INTEGER DEFAULT 0;"), NULL},
```
cdecker, line 671: Can we make this `funding_tx_remote_sigs_received`, otherwise I know I'll forget this eventually :-)

wallet/wallet.c
```diff
@@ -1135,60 +1136,62 @@ static bool wallet_channels_load_active(struct wallet *w)
...
+					", minimum_depth" // 9
+					", next_index_local" // 10
+					", next_index_remote" // 11
+					", next_htlc_id" // 12
+					", funding_tx_id" // 13
+					", funding_tx_outnum" // 14
+					", funding_satoshi" // 15
+					", our_funding_satoshi" // 16
+					", funding_locked_remote" // 17
+					", tx_sigs_remote" // 18
```
cdecker, line 1157: Please add new fields at the bottom of the statement, that avoids a whole cascade of index updates and minimizes the diff.

wallet/db.c
cdecker: The only exception to this should be if the field is conditionally set with some other field, in which case grouping makes sense.

cdecker (commented):
Verified up until commit 9e151f4e60baa9179e34d9498661c1cee2d7eecd.

Since we change the state machine to use two new states that might have been persisted in the DB, I'm a bit worried that we might break channels that were initiated before the upgrade, then loaded after the upgrade and no longer functional. Specifically I'm thinking about `CHANNELD_AWAITING_LOCKIN` vs. `DUALOPEND_AWAITING_LOCKIN`.

Can you confirm that this has been taken care of and works as expected?

I'll finish up the review asap, but thought some early feedback might be good 🎄🎅

niftynei:
Lots of great feedback here, thanks!

>Since we change the state machine to use two new states that might have been persisted in the DB, I'm a bit worried that we might break channels that were initiated before the upgrade, then loaded after the upgrade and no longer functional. Specifically I'm thinking about CHANNELD_AWAITING_LOCKIN vs. DUALOPEND_AWAITING_LOCKIN.
>
> Can you confirm that this has been taken care of and works as expected?

I don't believe this is a necessary concern, given that as it stands currently in `master`, there's no way for you to have opened a channel using `dualopend` unless you've manually edited the source code to add `OPT_DUAL_FUND` to the options added in `lightningd`. The code is only built if you're using `EXPERIMENTAL_FEATURES` but even then dual-funding hasn't been flagged on.  So the possibility of a user having a channel in a state that'd need handling for this is very low and won't be a common case.

Given that the code is 'still dark', do you think it's reasonable to add handling/test for this case?

lightningd/options.c
```diff
@@ -844,6 +852,13 @@ static void register_opts(struct lightningd *ld)
 				 opt_set_wumbo, ld,
 				 "Allow channels larger than 0.16777215 BTC");
 
+	opt_register_early_noarg("--experimental-dual-fund",
+				 opt_set_dual_fund, ld,
+				 "EXPERIMENTAL!!! Advertise dual-funding"
```
cdecker, line 857: `EXPERIMENTAL!!! ` -> `experimental: `

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -1020,6 +1020,13 @@ def passes_filters(hmsg, filters):
             out = out[2 + length:]
         return msgs
 
+    def config(self, config_name):
+        try:
+            opt = self.rpc.listconfigs(config_name)
+            return opt[config_name]
```
cdecker, line 1039: We can also have a default value, that'd also avoid the element not found issue here:

```
return opt.get(config_name, def)
```
Where `def` is just another argument above with `def = None`, so we have the same semantics.

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -1020,6 +1020,13 @@ def passes_filters(hmsg, filters):
...
+        except RpcError:
```
cdecker, line 1040: Not sure we should swallow what amounts to a communication issue.

tests/test_connection.py
```diff
@@ -2322,7 +2324,8 @@ def test_dataloss_protection(node_factory, bitcoind):
 
     # l2 must NOT drop to chain.
     l2.daemon.wait_for_log("Cannot broadcast our commitment tx: they have a future one")
-    assert not l2.daemon.is_in_log('sendrawtx exit 0')
+    assert not l2.daemon.is_in_log('sendrawtx exit 0',
+                                   start=l2.daemon.logsearch_start)
```
cdecker, line 2427: This is semantically the same as `wait_for_log` which starts at `logsearch_start`.

tests/test_misc.py
```diff
@@ -1161,8 +1161,10 @@ def test_funding_reorg_private(node_factory, bitcoind):
 
     l1.rpc.fundchannel(l2.info['id'], "all", announce=False)
     bitcoind.generate_block(1)                      # height 106
+
+    state = 'DUALOPEND' if l1.config('experimental-dual-fund') else 'CHANNELD'
```
cdecker, line 1168: Should this be `daemon` instead?

tests/test_plugin.py
```diff
@@ -1620,8 +1620,12 @@ def test_feature_set(node_factory):
     l1 = node_factory.get_node(options={"plugin": plugin})
 
     fs = l1.rpc.call('getfeatureset')
-    assert fs['init'] == expected_peer_features()
-    assert fs['node'] == expected_node_features()
+    extra = []
+    if l1.config('experimental-dual-fund'):
+        extra.append(223)
```
cdecker, line 1664:
```
extra = [223] if l1.config('experimental-dual_fund') else []
```

cdecker (commented):
Three quarters through and it's looking good :+1:

cdecker:
Reviewed up to 204f2a5fa1a0f17f05ff0d5763eed88bcb409d3e of PR version 96510adab2043f92af899538b32aac05080e0492. Will finish up asap.

cdecker (changes requested):
Turns out that the remainder wasn't that much, mostly tests and infra for tests. 
Just some minor feedback that needs to be addressed, otherwise LGTM :+1:

cdecker:
Thanks @niftynei for the update, I went through the [range-diff](http://46.101.246.115:8888/ElementsProject/lightning/pull/4293/range_diff/96510adab2043f92af899538b32aac05080e0492..2d827657818f701fe92ac50172dc76867d4227d1) and marked the addressed feedback as resolved.

The only ones that are not addressed yet are:

 - [Keeping a TLV around when all we need is one field](https://github.com/ElementsProject/lightning/pull/4293#discussion_r551404031)
 - [Reporting a missing config option when we fail to communicate](https://github.com/ElementsProject/lightning/pull/4293#discussion_r551280498)

Could you comment on those?

cdecker:
btw, the size of the PR made me update my range-diff tool so I can navigate it easier and with colors xD

niftynei (commented):
Forgot to submit these with the updates :facepalm:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4293/commits/2d827657818f701fe92ac50172dc76867d4227d1

cdecker (approved):

cdecker:
Rebased on top of `master` to get rid of the conflicts on generated files.

ACK https://github.com/ElementsProject/lightning/pull/4293/commits/cbcddf2ad8224dc8f84293476ba91f4235bb6565

cdecker:
I'm afraid that merging #4271 caused a conflict in `channeld.c` that requires rebasing and fixing before we can merge. @niftynei can you do that, I failed miserably since it conflicts inside of a `switch` and nested `#if` statement xD

niftynei:
Rebased :+1:

cdecker:
Tiny fixup to address a python lint. Merging asap :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4293/commits/92b03372a253534a17c598855d1d66206505bc5c

PR #4294, opened 2020-12-21
mfc: bug fixes for v2 opens

niftynei:
Fixes a few problems found while running tests against the v2 open pathways.

Update: this now depends on #4295, as we use the `take()` semantic for `json_add_psbt` in this patchset. First 'new commit' is `6cb23fb`

Changelog-None.

### DIFF
--- a/plugins/spender/multifundchannel.c
+++ b/plugins/spender/multifundchannel.c
@@ -868,7 +868,7 @@ static struct command_result *
 handle_mfc_change(struct multifundchannel_command *mfc)
 {
 	size_t change_weight;
-	struct amount_sat change_fee;
+	struct amount_sat change_fee, fee_paid, total_fee;
 	struct amount_sat change_min_limit;
 
 	/* Determine if adding a change output is worth it.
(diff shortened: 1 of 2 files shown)


cdecker:
Looking good, but need to settle #4295 first.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -86,10 +86,14 @@ def __repr__(self) -> str:
         """
         return str(self.millisatoshis) + "msat"
 
-    def to_satoshi(self) -> Decimal:
+    def to_satoshi(self, whole=False) -> Decimal:
         """
-        Return a Decimal representing the number of satoshis.
+        Return a Decimal representing the number of satoshis;
+        rounded up to the nearest satoshi iff whole=True
```
rustyrussell, line 92: This is deeply weird?  We usually expect truncation.

More importantly, Decimal has entire routines dedicated to this: to_integral_value().  It supports all kinds of rounding.

rustyrussell (approved):
Ack, modulo the weird rounding change.

contrib/pyln-client/pyln/client/lightning.py
cdecker: We discussed this in https://github.com/ElementsProject/lightning/pull/4295 and came to the conclusion that we need to expose the rounding behavior correctly.

In the end we punted on the full rounding functionality, and just added a function that rounds up in https://github.com/ElementsProject/lightning/pull/4295/commits/72550f42b092c12be5ceb9c2ee25440a6fe58ac7

cdecker:
Rebased on top of #4295, ready to go.

ACK https://github.com/ElementsProject/lightning/pull/4294/commits/e1938cd0ab0ed1d5b28894b6bbf39ac0804e742a

PR #4295, opened 2020-12-21
Small fixes: dual-funding

niftynei:
A few small things found while testing the dual-funding pathways.

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -92,6 +92,13 @@ class Millisatoshi:
         """
         return Decimal(self.millisatoshis) / 1000
 
+    def to_whole_satoshi(self) -> int:
+        """
+        Return an int respresenting the number of satoshis;
+        rounded up to the nearest satoshi
+        """
+        return (self.millisatoshis + 999) // 1000
+
     def to_btc(self) -> Decimal:
         """
         Return a Decimal representing the number of bitcoin.
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -1244,7 +1244,7 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("unreserveinputs", payload)
 
-    def fundpsbt(self, satoshi, feerate, startweight, minconf=None, reserve=True, locktime=None):
+    def fundpsbt(self, satoshi, feerate, startweight, minconf=None, reserve=True, locktime=None, min_witness_weight=None):
         """
         Create a PSBT with inputs sufficient to give an output of satoshi.
         """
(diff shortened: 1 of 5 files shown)


contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -88,9 +88,10 @@ def __repr__(self) -> str:
 
     def to_satoshi(self) -> Decimal:
         """
-        Return a Decimal representing the number of satoshis.
+        Return a Decimal representing the number of satoshis,
+        rounded up to the nearest satoshi
         """
-        return Decimal(self.millisatoshis) / 1000
+        return (Decimal(self.millisatoshis) + 999) // 1000
```
cdecker, line 94:
```python
import math
Decimal(math.ceil(self.millisatoshi / 1000))
```

cdecker: I'm not sure that we should just be rounding up or down on a whim, we need to keep the `Millisatoshi` as precise as possible, and then let the final step in the computation specify whether we want to round up or down.

Any attempt to do so implicitly will likely cause issues because what might seem intuitive now, may not be when looked at from another side.

openingd/dualopend.c
```diff
@@ -585,7 +585,13 @@ static char *check_balances(const tal_t *ctx,
 				      initiator_weight);
 
 	if (!amount_sat_greater_eq(accepter_diff, accepter_fee)) {
-		return "accepter fee not covered";
+		return tal_fmt(ctx, "accepter fee %s not covered %s",
```
cdecker, line 588: Would be clearer by showing the comparison directly:

```
"accepter fee not covered (need %s > have %s)"
```

cdecker (commented):
The rounding change bothers me, because it directly inverts the implicit semantics. Should really push the decision down to the caller about which direction to round to.

contrib/pyln-client/pyln/client/lightning.py
niftynei: Imo it's a mistake to return a non-rounded response from `to_satoshi` -- a partial value satoshi isn't *correct*. Taking a millisatoshi to a satoshi *should* be the last step in any millisatoshi operations?

niftynei: Maybe the better solution here is to change the interface to accept an optional `whole` argument.

e.g.
```
def to_satoshi(self, whole=False)
```

niftynei:
Updated.

contrib/pyln-client/pyln/client/lightning.py
cdecker: Agreed, however you're also inverting the semantics of the rounding of an existing API, which may break callers.

I was thinking more of deprecating `to_satoshi` as is and then add `floor()` and `ceil()` methods to the `Millisatoshi` class that returns the msat amount rounded in the desired direction.

cdecker: I think if `to_satoshi` returns a `Decimal` we should not round internally at all, and defer the rounding to the caller.

niftynei: You're right, modifying `to_satoshi` is the wrong approach, especially given that it does return the `Decmial` type.

Added a new method `to_whole_satoshi`, which returns an `int` type.

niftynei: Fwiw I wasn't a hugely into adding a call to `import math`, but if you think it's better than the inline math version I'd be happy to update/change it.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4295/commits/5617305b555ab46bba5c4b72c794fd798d615387

PR #4296, opened 2020-12-22
Added spent option to listfunds

### LINKED ISSUE #4291: Add spent funds to the `listfunds` method
There are a number of cases that require having access to earlier
owned funds, such as computing the fee for the transactions we sent
from a plugin or app. In order to do that we can add a simple
`spent=true` flag (semantically: "include spent outputs as well as
unspent ones in the `outputs` array) to the `listfunds` method.

This is a good first issue, and should be limited to:

 - Adding the new parameter to the JSON-RPC method
 - Adjusting the call to `wallet_get_utxos` in the handler function
   `json_listfunds` to use `OUTPUT_STATE_ANY` as third argument if the
   parameter is set to `true`.
   
This was first brought up by @hosiawak in #4289

hosiawak:
Added a `spent` option (a boolean) to `listfunds` which adds spent outputs to the `outputs` if set to true.

I didn't see any `test_listfunds` Python test so I thought I'd write one but then I couldn't figure out how to actually run these Python tests after reading `HACKING.md` :) 

* **blackbox tests** - These tests setup a mini-regtest environment and test
  lightningd as a whole.  They can be run individually:

  `PYTHONPATH=contrib/pylightning:contrib/pyln-client:contrib/pyln-testing:contrib/pyln-proto py.test -v tests/`

What's `py.test` ?

I checked `travis.yml` to see how it does that and it sets up bitcoind and elements and it's Linux specific but I'm using a Mac. If anyone knows how to run `tests/test_misc.py` on a Mac and could share this info with me I'd appreciate it.

Fixes #4291

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -851,11 +851,16 @@ class LightningRpc(UnixDomainSocketRpc):
         """
         return self.call("listforwards")
 
-    def listfunds(self):
+    def listfunds(self, spent=False):
         """
-        Show funds available for opening channels.
+        Show funds available for opening channels
+        or both unspent and spent funds if {spent} is True.
         """
-        return self.call("listfunds")
+
+        payload = {
+            "spent": spent
+        }
+        return self.call("listfunds", payload)
 
     def listtransactions(self):
         """
--- a/doc/lightning-listfunds.7
+++ b/doc/lightning-listfunds.7
@@ -3,13 +3,17 @@
 lightning-listfunds - Command showing all funds currently managed by the c-lightning node
 .SH SYNOPSIS
 
-\fBlistfunds\fR
+\fBlistfunds\fR [\fIspent\fR]
 
 .SH DESCRIPTION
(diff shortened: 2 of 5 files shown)


ZmnSCPxj:
On some distros you can install a `pytest` package and that is invoked as `pytest`, so just use `pytest`.  Depends on your Python setup though.  You may need to install it separately.

ZmnSCPxj (approved):
Code mostly okay.  Would appreciate a test though.

hosiawak:
@ZmnSCPxj I'm not a C or Python coder for the most part so thanks for the hint, helps me navigate these unknown waters. 

I managed to add the `listfunds` test and run it.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4296/commits/fe101e82728ceebae3b7ba020a8ad9df9f217bf7

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4296/commits/f31a27e865c3e8888dcd6eea932c5b35dc9e0d80

Issue #4186, opened 2020-11-06
Remove <hsm password> entry on command line from hsmtool


niftynei:
Inputting passwords into command line options is a security hole -- it gets entered into your bash command history.

A better way to do it would be to disable ECHO, as we do in #4065.

darosior:
Yeah, one way i thought of doing this in a not-so-backward-incompatible manner was to:
1) completely disregard the password field 
2) detect an encrypted hsm_secret (as we do in hsm_control.c) and interactively ask for a password in this case
3) in three releases, completely remove the disregarded `<hsm password>` field

ZmnSCPxj:
Not just on `bash` history --- `ps` and `top` can show it to ***anyone*** on the same machine as you, if they happen to run those at a very opportune time.

PR #4303, opened 2021-01-02
Take hsm_secret password from stdin, even for hsmtool!

darosior:
Fixes https://github.com/ElementsProject/lightning/issues/4186, based on #4284 

Went for the procedure [here](https://github.com/ElementsProject/lightning/issues/4186#issuecomment-723579122) for the API break. Should we instead support both for 3 releases ? It seemed overkill to me as it's a command-line tool and worse case the user has to input their password twice.


darosior:
Travis failure is just a one job timeout https://travis-ci.org/github/ElementsProject/lightning/jobs/752638311

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4303/commits/dc1132dce0162f486a67ead01bf2a5e42a218c09

Issue #4302, opened 2021-01-02
hsmtool: `generatehsm` does not generate a valid hsm_secret (anymore?)


darosior:
Didn't tackle it down yet, noticed it while working on #4186 . See the comment on the original PR https://github.com/ElementsProject/lightning/pull/4065#issuecomment-753539921

PR #4304, opened 2021-01-03
hsmtool: fix hsm_secret generation

darosior:
Based on #4284 , fixes https://github.com/ElementsProject/lightning/issues/4302 .

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -1097,7 +1097,6 @@ def test_hsmtool_dump_descriptors(node_factory, bitcoind):
     assert len(bitcoind.rpc.listunspent(1, 1, [addr])) == 1
 
 
-@pytest.mark.xfail(strict=True)
 @unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
 def test_hsmtool_generatehsm(node_factory):
     l1 = node_factory.get_node()
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -1097,6 +1097,39 @@ def test_hsmtool_dump_descriptors(node_factory, bitcoind):
     assert len(bitcoind.rpc.listunspent(1, 1, [addr])) == 1
 
 
+@pytest.mark.xfail(strict=True)
+@unittest.skipIf(VALGRIND, "It does not play well with prompt and key derivation.")
+def test_hsmtool_generatehsm(node_factory):
+    l1 = node_factory.get_node()
+    l1.stop()
+    hsm_path = os.path.join(l1.daemon.lightning_dir, TEST_NETWORK,
+                            "hsm_secret")
+    master_fd, slave_fd = os.openpty()
+
+    hsmtool = HsmTool("generatehsm", hsm_path)
+    # You cannot re-generate an already existing hsm_secret
+    hsmtool.start(stdin=slave_fd, stdout=subprocess.PIPE,
+                  stderr=subprocess.PIPE)
+    assert hsmtool.proc.wait(5) == 2
+    os.remove(hsm_path)
+
+    # We can generate a valid hsm_secret from a wordlist and a "passphrase"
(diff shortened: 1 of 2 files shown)


tests/test_wallet.py
```diff
@@ -985,6 +985,14 @@ def test_hsm_secret_encryption(node_factory):
     assert id == l1.rpc.getinfo()["id"]
 
 
+class HsmTool(TailableProc):
+    """Helper for testing the hsmtool as a subprocess"""
+    def __init__(self, *args):
+        TailableProc.__init__(self)
+        assert hasattr(self, "env")
+        self.cmd_line = ["tools/hsmtool", *args]
```
cdecker, line 993: This will clash with #4303, so merging either will require a rebase :-)

tools/hsmtool.c
```diff
@@ -503,7 +503,9 @@ static int generate_hsm(const char *hsm_secret_path)
 	if (fd < 0) {
 		errx(ERROR_USAGE, "Unable to create hsm_secret file");
 	}
-	if (!write_all(fd, bip32_seed, bip32_seed_len))
+	/* Write only the first 32 bytes, length of the (plaintext) seed in the
+	 * hsm_secret. */
+	if (!write_all(fd, bip32_seed, 32))
```
cdecker, line 508: Out of curiosity, what are the remaining 32 bytes returned by `bip39_mnemonic_to_seed`?

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/4304/commits/6455c908fd0d00e1f1b66e9816baa8f81745403d

tests/test_wallet.py
darosior: Yeah, i worked on #4303 first but didn't want it to block the fix for whatever reason :)

cdecker: No problem, the changes just seemed eerily familiar xD

tools/hsmtool.c
darosior: It's the output of a SHA512 https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#from-mnemonic-to-seed .
I'm still confused about how i could overlook this in my initial review...

darosior: Thinking more about it i'm not sure it's cryptographically sound. Would `sha256`ing the output be preferable ? This would completely break the link with previously-generated seeds, but hey they were not even valid in the first place.

cdecker: Hm, switching hash function seems the wrong way to go here. Let's stay as close to bip39 as possible. Can we even still express our derivation paths as BIP32 HD keys? Would be quite inconvenient if we break compatibility with all existing recovery tools.

darosior: We don't have compatibility with existing recovery tools already, as we use bip39 to derive the seed which we'll later derive the actual BIP32 seed from (with our homebrewed HKDF).

cdecker:
Rebased on top of #4303

ACK https://github.com/ElementsProject/lightning/pull/4304/commits/dbbcc277326c7ac2fcee3db834640576369c46d9

tools/hsmtool.c
```diff
@@ -530,7 +532,9 @@ static int generate_hsm(const char *hsm_secret_path)
 	if (fd < 0) {
 		errx(ERROR_USAGE, "Unable to create hsm_secret file");
 	}
-	if (!write_all(fd, bip32_seed, bip32_seed_len))
+	/* Write only the first 32 bytes, length of the (plaintext) seed in the
```
jsarenik, line 535: plaintext?

jsarenik (commented):
Sorry I noticed just now, but the `hsm_secret` contains 32 bytes of binary, not plaintext.

darosior:
It was meant by opposition to 'ciphertext'

-------- Original Message --------

PR #4311, opened 2021-01-05
pyln: Add command notifications support

cdecker:
Adds command notification support to both the RPC client library as
well as the plugin library. This allows RPC clients to register a
notification callback that gets called for each notification, and
allows plugin methods to report status or progress using the
`Request.notify` method.

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -141,6 +141,47 @@ class Request(dict):
     def _write_result(self, result: dict) -> None:
         self.plugin._write_locked(result)
 
+    def _notify(self, method: str, params: JSONType) -> None:
+        """Send a notification to the caller.
+
+        Can contain a variety of things, but is usually used to report
+        progress or command status.
+
+        """
+        self._write_result({
+            'jsonrpc': '2.0',
+            'params': params,
+            "method": method,
+        })
+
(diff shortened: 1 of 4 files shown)


contrib/pyln-client/pyln/client/plugin.py
```diff
@@ -141,6 +141,24 @@ def set_exception(self, exc: Union[Exception, RpcException]) -> None:
     def _write_result(self, result: dict) -> None:
         self.plugin._write_locked(result)
 
+    def notify(self, message: str, level='DEBUG') -> None:
+        """Send a notification to the caller.
```
rustyrussell, line 145: We already have this?  It's split into the two kinds of notifications we have currently defined though: notify_message and notify_progress.

cdecker: Oh, I didn't see that it's in the `Plugin` not the `Request`, that's a really weird place for it, if we require access to the request anyway. I'll change `Plugin.notify_message` to become a wrapper of `Request.notify` and move that code there instead.

rustyrussell: Ah, good point.  This is a new API, so just move it.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4311/commits/81a73b528b4653e74dcbd9a37910f5b0f7735e21

PR #4312, opened 2021-01-05
json-rpc: Add support for querying by hash or bolt11 to `listinvoices`

cdecker:
A user reported that it is sometimes cumbersome to search an invoice
based on the payment hash or the bolt11 string in the full list, which
may be required when we don't have the label available.

This adds support for querying / filtering based on the `payment_hash`
or `bolt11` string.

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -875,12 +875,17 @@ class LightningRpc(UnixDomainSocketRpc):
         """
         return self.call("listtransactions")
 
-    def listinvoices(self, label=None):
-        """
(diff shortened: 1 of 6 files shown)


rustyrussell (approved):
Ack 7f128b1ade09b74485372c2071917af5f00e2906

Only one minor thing: maybe rename 'bolt11' to 'invstring', as suggested by the offers branch?  It's going to be an anachronism one day.

cdecker:
Re-applying Rusty's ACK after renaming bolt11 to invstring.

ACK https://github.com/ElementsProject/lightning/pull/4312/commits/23ab13a04cbd022e4545ce494dba6f31ea7e1334

PR #4314, opened 2021-01-06
channeld: stop taken() leak.

rustyrussell:
I saw this message:

```
lightning_channeld: outstanding taken(): channeld/channeld.c:3087:blinding
lightning_channeld: outstanding taken(): channeld/channeld.c:3087:blinding
lightning_channeld: outstanding taken(): channeld/channeld.c:3087:blinding
```

The caller does take(blinding), but blinding can be NULL.  We should
move the code around to do the take() earlier anyway.

Changelog-None

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -496,6 +496,18 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	htlc->amount = amount;
 	htlc->state = state;
 
+	htlc->rhash = *payment_hash;
+	if (blinding)
+		htlc->blinding = tal_dup(htlc, struct pubkey, blinding);
+	else {
+		/* Can be taken, even if NULL. */
+		taken(blinding);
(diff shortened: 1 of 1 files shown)


cdecker:
I don't quite see why this is necessary? I thought `take()` sets a flag in the `tal` header, which doesn't exist for NULL pointers, does it? Shouldn't `take(NULL)` just be a noop? Looks good otherwise :+1:

ACK https://github.com/ElementsProject/lightning/pull/4314/commits/def2594bcf998dac9f5f25163c6f637cc75fa965

cdecker:
Found the answer:

https://github.com/ElementsProject/lightning/blob/0726d2912e1a3ac2bf27f03f97d8d1c6f7b9b863/ccan/ccan/take/take.c#L58-L67

`NULL` is a valid pointer that can be added to the `takenarr`, and needs to be unmarked correctly as well.

PR #4315, opened 2021-01-06
json_scan(): less boilerplate for dealing with JSON

rustyrussell:
json_delve was cute, but inefficient and still left the caller to do the conversion.  During my break, I thought of a better way to do this, and here it is.

There are more places we could clean up (grep for json_get_member), but I didn't want to make too many conflicts.

Changelog-None


plugins/bcli.c
```diff
@@ -833,14 +790,10 @@ static void parse_getnetworkinfo_result(struct plugin *p, const char *buf)
...
-		plugin_err(p, "No 'version' in '%s' ? Got '%s'. Can not"
-			      " continue without proceeding to sanity checks.",
-			      gather_args(bitcoind, "getnetworkinfo", NULL), buf);
-
-	if (!json_to_u32(buf, versiontok, &bitcoind->version))
-		plugin_err(p, "Invalid 'version' in '%s' ? Got '%s'. Can not"
+	if (!json_scan(buf, result, "{version:%,localrelay:%}",
+		       JSON_SCAN(json_to_u32, &bitcoind->version),
+		       JSON_SCAN(json_to_bool, &tx_relay)))
+		plugin_err(p, "No 'version' or localrelay in '%s' ? Got '%s'. Can not"
```
cdecker, line 796: This error message can appear when `localrelay` is not present, which is confusing.

plugins/pay.c
```diff
@@ -512,57 +504,53 @@ static struct command_result *waitsendpay_error(struct command *cmd,
...
 		/* Otherwise, add erring channel to exclusion list. */
-		tal_arr_expand(&pc->excludes,
-			       tal_fmt(pc->excludes, "%.*s",
-			       nodeidtok->end - nodeidtok->start,
-			       buf + nodeidtok->start));
+		tal_arr_expand(&pc->excludes, tal_steal(pc->excludes, idstr));
 	} else {
+		struct short_channel_id scid;
+		u32 dir;
+		const char *scidstr = short_channel_id_to_str(tmpctx, &scid);
```
cdecker, line 533: `scid` is uninitialized here.

plugins/pay.c
```diff
@@ -512,57 +504,53 @@ static struct command_result *waitsendpay_error(struct command *cmd,
...
+
+		if (!json_scan(buf, error, "{data:{erring_channel:%,erring_direction:%}}",
+			       JSON_SCAN(json_to_short_channel_id, &scid),
+			       JSON_SCAN(json_to_number, &dir)))
+			plugin_err(cmd->plugin, "waitsendpay error no erring_channel/direction '%.*s'",
+				   error->end - error->start, buf + error->start);
```
cdecker, line 539: `json_tok_full_len` + `json_tok_full`.

plugins/libplugin.c
```diff
@@ -491,17 +491,18 @@ static struct json_out *start_json_request(const tal_t *ctx,
 }
 
 /* Synchronous routine to send command and extract single field from response */
```
cdecker, line 493: This comment is now outdated :-)

cdecker (commented):
Very intriguing approach to parsing, I like it. Processing a JSON buffer with a JSON-like syntax (missing quotes) reminds me of XML + XSLT, and I would have preferred a `jq`-like syntax, but that's probably just what I'm accustomed to, and `jq` doesn't allow extracting multiple fields at the same time.

One issue that would be great if it were addressed are error messages (can also be a followup PR): we should maintain a stack of our current position, and collect all parse-failures we encounter with their position, so we can at least give syntactic feedback about the issues we encounter, rather than just a catch all "scan failed" error.

plugins/pay.c
rustyrussell: Oops!  Thanks

plugins/libplugin.c
rustyrussell: Fixed!

rustyrussell:
Agree with your complaints about error handling.  Appending a new commit...

plugins/pay.c
rustyrussell: I simply moved the code, but I've fixed it.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4315/commits/26f631090d8255bef7f2be536923cc39c9b092f5

cdecker:
Rebased and squashed on top of `master`

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4315/commits/636c2d0ef47ccdab344bbf7d9e55db2ef1b605f5

PR #4319, opened 2021-01-07
Revert "wallet: Remove in-memory utxoset filter"

cdecker:
This reverts commit c239a7161bb6080ad1e7a9b98ec240d256df3ce9.

The goal of c239a716 was to reduce the memory footprint of our
internal UTXO set tracking, and testing against the sqlite3 backend
showed no performance impact. We have since found that the added
roundtrips to the DB server with postgres was having a considerable
performance impact, and backups would also bloat due to the increased
number of queries.

Undoing this change skips the noop updates that were causing this
regression.

Changelog-Fixed: db: Fixed a performance regression during block sync, resulting in many more queries against the DB than necessary.

### DIFF
--- a/wallet/db_postgres_sqlgen.c
+++ b/wallet/db_postgres_sqlgen.c
@@ -1053,6 +1053,12 @@ struct db_query db_postgres_queries[] = {
          .readonly = true,
     },
     {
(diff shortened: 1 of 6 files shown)


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4319/commits/8026a9589621f3348512caad32170370418a9d74

PR #4324, opened 2021-01-10
don't error "pay" when network is different.

fiatjaf:
The code in this PR is just a suggestion.

I'm working on a plugin that will enable seamless payments between Bitcoin and Liquid Lightning Networks, and it requires the Liquid node to perform some temporary magic (at least until Trampoline is available) to get a route from itself to the payee on the Bitcoin side.

It could work by using `rpc_command` to replace the `getroute` calls done from the default `pay` plugin, but this is not possible currently because `pay` will error if passed an invoice that isn't to the same network.

The other option is to rewrite `pay` entirely just to get this patch into `getroute` working, which would be very wasteful, and since `rpc_command` exists and fits perfectly this use case it might be better remove these checks from `pay`.

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1320,18 +1320,10 @@ static struct command_result *json_pay(struct command *cmd,
 		return command_param_failed();
 
 	b11 = bolt11_decode(cmd, b11str, plugin_feature_set(cmd->plugin),
-			    NULL, chainparams, &fail);
+			    NULL, NULL, &fail);
 	if (!b11) {
 		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
 				    "Invalid bolt11: %s", fail);
-	}
-
-	if (!b11->chain) {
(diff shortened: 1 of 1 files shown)


fiatjaf:
Other ideas of how to handle this on `pay` (obviously not implemented here) are:

1. Upon failing a `getroute`, check if the invoice is for the same network and return a proper error to the user (instead of just "couldn't find route" or "unknown peer").
2. Do a preliminary `getroute` with a small amount and see if it returns _anything_ before attempting to presplit the payment (which would then cause multiple `getroute` attempts that would all fail in case of a different network).

vincenzopalazzo:
Hi,

Another opportunity is to use another parameter in the pay plugin to jump the check of the network, like the idea under some Javascript framework that you need to specify that you need to run out of the sandbox.

rustyrussell:
You need to intercept the pay command, and if it's a different chain, redirect it to some other ("payotherchain"?) command.

Issue #3578, opened 2020-03-07
Can't call an RPC method from inside an `rpc_command` hook


fiatjaf:
### Issue and Steps to Reproduce

I've been testing a plugin that replaces `getroute` using the `rpc_command` command hook. The plugin reads information from `listchannels` RPC plus opens a database stored in a file inside `lightning-dir`.

I tried it a few times and it was working, but now all of a sudden it started crashing when I try to `pay`. The crash looks like this:

```
lightningd: FATAL SIGNAL 11 (version v0.8.0-316-gac5cba2)
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: /home/fiatjaf/lightning/lightningd/lightningd: Too many open files
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
/home/fiatjaf/lightning/lightningd/lightningd: libbacktrace: failed to read executable information
[...]

PR #4326, opened 2021-01-10
add failing test for rpc_command calling another rpc method.

fiatjaf:
Actually this doesn't fail, it hangs forever.

Related to https://github.com/ElementsProject/lightning/issues/3578.

### DIFF
--- a/tests/plugins/rpc_command.py
+++ b/tests/plugins/rpc_command.py
@@ -19,11 +19,15 @@ def on_rpc_command(plugin, rpc_command, **kwargs):
         return {"return": {"result": ["Custom result"]}}
     elif request["method"] == "sendpay":
         # Don't allow this command to be executed
-        return {"return": {"error": {"code": -1,
-                                     "message": "You cannot do this"}}}
+        return {"return": {"error": {"code": -1, "message": "You cannot do this"}}}
     elif request["method"] == "help":
         request["method"] = "autocleaninvoice"
         return {"replace": request}
+    elif request["method"] == "listnodes" and request["params"]:
+        # Fetch results and modify
+        info = plugin.rpc.call("getinfo")
+        info["id"] = request["params"][0]
+        return {"return": {"result": info}}
     return {"result": "continue"}
 
 
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -1323,6 +1323,12 @@ def test_rpc_command_hook(node_factory):
     # Test command redirection to a plugin
     l1.rpc.call('help', [0])
 
+    # Test calling other RPC methods from inside an rpc_command hook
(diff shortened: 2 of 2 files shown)


rustyrussell:
Erk, this is reentrant of course.  Not sure the best way to handle this!

cdecker:
Yeah, this is not something we should or will ever support. `rpc_command` is a synchronous call blocking incoming RPC processing, if we then call anything on the RPC the expected behavior is to block indefinitely. Don't do this.

fiatjaf:
Yes, it makes sense, better not do this. I'll find another way.

But please add a note to the documentation because I was misled by this thinking it would work multiple times.

gijswijs:
I think it's actually worse than described in this issue and there *is* something wrong: Just using the `rpc_command` hook in your pkugin makes any rpc call from that plugin impossible, regardless of whether the call comes from inside that hook. That seems unwanted behavior.

Steps to reproduce:

Run this plugin:

```python
#!/usr/bin/env python3
from pyln.client import Plugin

plugin = Plugin()
plugin.our_node_id = None

@plugin.init()
    plugin.log("Plugin dpchannels.py initialized")
    plugin.our_node_id = plugin.rpc.getinfo()["id"]
    plugin.log("Our node id {}".format(plugin.our_node_id))

@plugin.hook("rpc_command")
def on_rpc_command(plugin, rpc_command, **kwargs):
    return {"result": "continue"}

plugin.run()
```

This plugin will hang upon initialization, because the `plugin.rpc.getinfo()` will never return a result. If you remove the 'rpc_command' hook, it does work.

It seems to me this should not have to be blocking, but somehow it is. 

@fiatjaf could you reopen this issue, or should I submit a new one?

fiatjaf:
I can't reopen, the button is inactive.

And I think it's better to open a new one anyway, as this is a different and more serious issue.

cdecker:
You can run the `getinfo` in a separate thread to avoid the mutual deadlock, it is an issue with pyln since that is singlethreaded. A fix to this could be to have a ThreadPoolExecutor that handles incoming method and hook calls.

gijswijs:
Can confirm @cdecker 's solution works:

Same plugin, now with threading

```
#!/usr/bin/env python3
from pyln.client import Plugin
import threading

plugin = Plugin()
plugin.our_node_id = None

def get_node_id():
    plugin.log("Thread started")
    plugin.our_node_id = plugin.rpc.getinfo()["id"]
    plugin.log("Our node id from thread{}".format(plugin.our_node_id))

@plugin.init()
    plugin.log("Plugin dpchannels.py initialized")
    thread = threading.Thread(target=get_node_id, args=())
    thread.start()
    plugin.log("Our node id from init {}".format(plugin.our_node_id))

@plugin.hook("rpc_command")
def on_rpc_command(plugin, rpc_command, **kwargs):
    return {"result": "continue"}

plugin.run()
```

PR #4332, opened 2021-01-14
fetchinvoice: don't abort() if we try to fetch an invoice from oursel…

rustyrussell:
…ves.

blindedpath.c does:

```
	size_t num = tal_count(route);

	if (!num)
		abort();
```

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4332/commits/9f8f3ccf60e41ceca42c3ab321be40e86ff35f95

PR #4350, opened 2021-01-25
lightningd: disallow --daemon without --log-file.

rustyrussell:
From #clightning:

     (11:24:10) andytoshi: hiya, i'm trying to set up a new lightningd node, and when i run lightningd --network=bitcoin --log-level=debug --daemon
     (11:24:17) andytoshi: i get errors of the form fetchinvoice: Malformed JSON reply '2021-01-25T00:51:16.655Z DEBUG   plugin-offers: disabled itself at init: offers not enabled in config
     (11:24:43) andytoshi: there are a couple variants of this, but always some form of "something: failed to parse <a log line> as json"

Indeed, we close stdout, and it ends up being reused for some plugin.
But the real problem is that we log to stdout by default, which doesn't
make sense.  If they really want to discard logs, they can use
--log-file=/dev/null.

### DIFF
--- a/doc/lightningd-config.5
+++ b/doc/lightningd-config.5
@@ -210,7 +210,8 @@ as well\.
 
 
  \fBdaemon\fR
-Run in the background, suppress stdout and stderr\.
+Run in the background, suppress stdout and stderr\.  Note that you need
+to specify \fBlog-file\fR for this case\.
 
 
  \fBconf\fR=\fIPATH\fR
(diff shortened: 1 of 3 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4350/commits/0efec239fba7ee91e2bcdbf5f201f13e9773ac9d

PR #4356, opened 2021-01-26
pyln: Bump pyln version to 0.9.3 before publishing on PyPI

cdecker:
Let's just keep pyln synched up with the c-lightning version

Changelog-None


darosior (approved):
ACK 3b61e16877949d8cd8bf0be9f18da9ee3f64b578

PR #4357, opened 2021-01-26
ci: Switch to Github Actions and remove Travis CI

cdecker:
With Travis CI refusing to build any of our configurations it's time to switch to something hopefully a bit more stable.


tests/test_pay.py
```diff
@@ -1271,8 +1271,11 @@ def test_forward_stats(node_factory, bitcoind):
     l1, l2, l3, l4, l5 = node_factory.get_nodes(5, opts=[{}] * 4 + [{'may_fail': True}])
     node_factory.join_nodes([l1, l2, l3], wait_for_announce=False)
     l2.openchannel(l4, 10**6, wait_for_announce=False)
-    l2.openchannel(l5, 10**6, wait_for_announce=True)
+    l2.openchannel(l5, 10**6, wait_for_announce=False)
 
+    sync_blockheight(bitcoind, [l1, l2, l3, l4, l5])
+    bitcoind.generate_block(1)
+    sync_blockheight(bitcoind, [l1, l2, l3, l4, l5])
```
rustyrussell, line 1278: Why sync before *and* after? I think first one can be omitted.

rustyrussell (approved):
Great improvements in here! Pretty epic, in fact.

I will test on my local machine to see if any of the test fixes broke things for me.

.github/workflows/ci.yaml
```diff
@@ -3,8 +3,47 @@ name: Continuous Integration
 on: [push, pull_request]
 jobs:
   build-and-test:
+    name: Test DEV=${{ matrix.developer }} VAL=${{ matrix.VALGRIND }} EXP=${{ matrix.EXPERIMENTAL_FEATURES }}
+    env:
```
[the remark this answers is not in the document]
cdecker: Seems not to matter too much, fixed up for consistency.

tests/test_pay.py
cdecker: Had a bit of a switching around of line here. The problem was slow chain sync on some nodes (valgrind slowing them down), so we'd blast past the 6 blocks (confirmation + 5 to get to the announce height) on the announcing node, but the node receiving the announcement was still stuck at height - 6, and then they'd complain about a future channel. Syncing after the confirmation should be enough though, so I removed the first one.

rustyrussell:
Tried rerunning, and it breaks in different ways every time.  I can't see how to run a single job, it only lets me do them all :(

cdecker:
> Tried rerunning, and it breaks in different ways every time. I can't see how to run a single job, it only lets me do them all :(

Yep, that's a common feature request on Github's own tracker. Hopefully they'll implement it some time soon. I'll continue stabilizing tests after this is merged, making it more reliable and reducing the number of reruns we need. So far I have a failure rate of about 1/4, but each fixed test has a big effect on this.

Since we have the reports generated as artifacts we can easily integrate them into my flaky test tracking tool (http://46.101.246.115:5001/) and then we can hunt them down more easily.

cdecker:
I should have mentioned that I have been rescheduling the runs a couple of times to see if new flakes show up, and to fix known ones, so not surprisingly some of the runs will stick to a failed state because I follow them up with a fix :-)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4357/commits/f2b8355604478b4bd98a46174e278c6340008641

PR #4365, opened 2021-02-01
Gossmap overlay RFC

rustyrussell:
This is a simple change designed to let us create transitory additions (i.e. from routehints) into the map.  The idea is to add the changes, perform routing, then clear them.

There needs to be other infrastructure to make it easy to persistently track channel capacities, but that can be done at a higher level.

Changelog-None


tests/test_misc.py
```diff
@@ -64,7 +64,7 @@ def test_names(node_factory):
         ('022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59', 'SILENTARTIST', '022d22'),
         ('035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d', 'HOPPINGFIRE', '035d2b'),
         ('0382ce59ebf18be7d84677c2e35f23294b9992ceca95491fcf8a56c6cb2d9de199', 'JUNIORFELONY', '0382ce'),
-        ('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e', 'SOMBERFIRE', '032cf1'),
+        ('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e', 'SOMBER1FIRE', '032cf1'),
```
cdecker, line 67: This is the break in CI.

cdecker (commented):
Code looks good, except that tiny typo in the test, that is causing all the CI failures.

I'm not quite sure that actually modifying the gossmap instance to provide a mutable view is the way to go though. As you note in the comment this means that we may end up not correctly restoring channels and nodes to their former state when cleaning up.

I was expecting something more like a struct that is backed by a real (singleton) gossmap and that'd overlay local changes when querying, handing the queries it doesn't have to modify to an unmodified gossmap. This would then mean that once we're done with the view we can just `tal_free` it and don't have to go through the cleanup. Furthermore we could then have multiple concurrent views active that act isolated from each other, since they'd each only be writing to their own view, and not the backing gossmap.

On the other hand, your implementation accommodates all our use-cases today, and adding complexity like my proposal would just for the sake of it isn't worth it. wdyt?

tests/test_misc.py
rustyrussell: Oops!

rustyrussell:
> Code looks good, except that tiny typo in the test, that is causing all the CI failures.
> 
> I'm not quite sure that actually modifying the gossmap instance to provide a mutable view is the way to go though. As you note in the comment this means that we may end up not correctly restoring channels and nodes to their former state when cleaning up.
> 
> I was expecting something more like a struct that is backed by a real (singleton) gossmap and that'd overlay local changes when querying, handing the queries it doesn't have to modify to an unmodified gossmap. This would then mean that once we're done with the view we can just `tal_free` it and don't have to go through the cleanup. Furthermore we could then have multiple concurrent views active that act isolated from each other, since they'd each only be writing to their own view, and not the backing gossmap.
> 
> On the other hand, your implementation accommodates all our use-cases today, and adding complexity like my proposal would just for the sake of it isn't worth it. wdyt?

It's also hard to do a proper non-intrusive overlay.  The current datastructures are extremely efficient (arrays), and we'd have to add an explicit lookup for local additions (e.g. when iterating through channels from a node for dijkstra). We could probably hide the current modifications behind an "apply_changes/unapply_changes" kind of API though.

rustyrussell:
OK, I reworked the API to be better defined and more efficient.  You still have to apply and unapply, though.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4365/commits/47f04fc6d76557dfb0b78bc217f98398f1cb3e7c

PR #4366, opened 2021-02-01
ci: Split the test configurations into separate jobs and stabilize some more tests

cdecker:
We split the test configurations into 3 parts:

 - Smoke-tests: quick tests running in under 5 minutes, used to gate the other two sets
 - Normal-tests: tests different compilation configurations and architecture
 - Valgrind-tests: heavy duty tests that are really a single configuration split into test groups to make it bearable

The smoke-tests gate the other two so we don't waste resources on what will ultimately be a failed run.

~Dividing into multiple jobs also allows us to rerun just jobs that have a failed config. While not even close to the desired "rerun-a-single-test", it's better.~

My bad: re-running reruns the entire workflow, not the job. So in order to have selective reruns we'd have to split the `ci.yaml` file into multiple ones, which would not allow coordination anymore...


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4366/commits/c34d9e7ad041019b71575a9c0f2e52fadc6cabbb

Though it's up to you if you want to split these instead, to allow single restarts.  Perhaps that's more useful than smoke-test separation, since it seems we have to choose one :(

cdecker:
> Though it's up to you if you want to split these instead, to allow single restarts. Perhaps that's more useful than smoke-test separation, since it seems we have to choose one :(

We'd have to write and maintain 23 different YAML files to get individual restartability, which seems ludicrous. My hope is Github eventually wakes up and implements the ability to restart individual configurations, and in the meantime we improve our tests to remove flakyness. Lack of the former motivates us to do the latter I guess :wink:

cdecker:
I'll let this run a couple more times on GA, to suss out some more flaky tests :+1:

cdecker:
I have one last `valgrind` flake I'd like to address quickly and then I'll undraft this PR :-)

cdecker:
Added a minor check to the smoke tests: verify that no generated / derived file checked in changes after calling `make`. This should help us get a grip on drift in the sqlgen, docs and other gen files which are causing later failures down the line with the `NO_PYTHON` config.

PR #4368, opened 2021-02-03
Add `excess_as_change` flag to fundpsbt/utxopsbt

niftynei:
For convenience, it's easier to allow the caller to request that a change output be added to a PSBT.


rustyrussell (approved):
Ack f036b2a87661c9698fff2bd9525b6d6a506ee6ce

We could argue that the caller should do this, but it *does* handle a common case.

PR #4369, opened 2021-02-03
chore: pyln testing improvements

m-schmoock:
Tiny changes that can improve testing.

## About `wait_for` changes
When thinking about it, it does not make it potentially faster, but when timeout occurred, it does a final success() check to see if the last `time.sleep()` lead to a situation where `success()` became `True`. Before that, it did a `time.sleep()` and didn't re-check `success()`.

We could further improve this by checking that the final `sleep` will be finished right after timeout would occur so we are more precise on the final sleep and do not waste any time.


m-schmoock:
Closed for now. I dislike it :D

PR #4372, opened 2021-02-04
DRAFT: RBF implementation, for reals

niftynei:
Work in Progress, probably should be split into segments.

Adds two new messages to the protocol:
- INIT_RBF
- ACK_RBF


openingd/dualopend.c
```diff
@@ -2349,6 +2346,178 @@ static void opener_start(struct state *state, u8 *msg)
...
+	}
+
+	if (!fromwire_dualopend_got_rbf_offer_reply(state, msg,
+						    &tx_state->accepter_funding,
+						    &tx_state->psbt))
+		master_badmsg(WIRE_DUALOPEND_GOT_RBF_OFFER_REPLY, msg);
+
+	if (!tx_state->psbt)
+		tx_state->psbt = create_psbt(tx_state, 0, 0,
+					     tx_state->tx_locktime);
```
rustyrussell, line 2784: I think this should be assigned unconditionally?  Aren't we otherwise just reusing the last one (directly; we don't copy it!).

openingd/dualopend.c
```diff
@@ -1462,20 +1462,222 @@ static bool run_tx_interactive(struct state *state,
...
+				      type_to_string(tmpctx, struct wally_psbt,
+						     tx_state->psbt));
+		return NULL;
+	}
+
+	/* Wait for the peer to send us our commitment tx signature */
+	// FIXME: use rbf_failed instead?
+	msg = opening_negotiate_msg(tmpctx, state);
+	if (!msg)
+		return NULL;
```
rustyrussell, line 1559: `*err_reason` unset here?

lightningd/dual_open_control.c
```diff
@@ -1463,6 +1649,46 @@ static void accepter_psbt_changed(struct subd *dualopend,
...
+
+	if (!fromwire_dualopend_got_rbf_offer(msg,
+					      &payload->channel_id,
+					      &payload->their_funding,
+					      &payload->funding_feerate_per_kw,
+					      &payload->locktime)) {
+		log_broken(channel->log, "Malformed dualopend_got_rbf_offer %s",
+			   tal_hex(msg, msg));
+		// FIXME: is this the right thing to do here?
+		tal_free(dualopend);
```
rustyrussell: Yes.

wallet/wallet.c
```diff
@@ -1620,33 +1608,28 @@ void wallet_channel_save(struct wallet *w, struct channel *chan)
...
+	db_bind_int(stmt, 27, chan->feerate_base);
+	db_bind_int(stmt, 28, chan->feerate_ppm);
+	db_bind_talarr(stmt, 29, chan->remote_upfront_shutdown_script);
+	db_bind_int(stmt, 30, chan->option_static_remotekey);
+	db_bind_int(stmt, 31, chan->option_anchor_outputs);
+	db_bind_talarr(stmt, 32, chan->shutdown_scriptpubkey[LOCAL]);
+	db_bind_int(stmt, 33, chan->closer);
+	db_bind_int(stmt, 34, chan->state_change_cause);
+	db_bind_u64(stmt, 35, chan->dbid);
 	db_exec_prepared_v2(take(stmt));
```
rustyrussell, line 1633: Should this commit remove those fields from the db migrations as well?

lightningd/dual_open_control.c
```diff
@@ -54,6 +54,23 @@ unsaved_channel_disconnect(struct channel *channel,
 }
 
 
+/* FIXME: remove when used */
+struct channel_inflight *
+channel_current_inflight(struct channel *channel);
```
rustyrussell: Nit: You can leave this static, and use "UNNEEDED".

lightningd/dual_open_control.c
```diff
@@ -1662,6 +1662,54 @@ static void handle_validate_rbf(struct subd *dualopend,
...
+	struct channel_id *cid;
+	struct channel *channel;
+	struct amount_sat *amount;
+	struct wally_psbt *psbt;
+	struct open_attempt *oa;
+
+	if (!param(cmd, buffer, params,
+		   p_req("channel_id", param_channel_id, &cid),
+		   p_req("amount", param_sat, &amount),
+		   p_req("initialpsbt", param_psbt, &psbt),
```
rustyrussell, line 1863: Um, what is "amount" here?  That's previously been used for "amount to open the channel with", so is this the same?

(I guess I'll find out when I get to the docs?)

lightningd/dual_open_control.c
```diff
@@ -2486,11 +2556,19 @@ static const struct json_command openchannel_bump_command = {
 	"Attempt to bump the fee on {channel_id}'s funding transaction."
 };
 
+static const struct json_command openchannel_next_feerate_command = {
+	"openchannel_next_feerate",
+	"channels",
+	json_openchannel_next_feerate,
+	"What's the next feerate for an RBF of this {channel_id}?"
+};
```
rustyrussell, line 2692: Well, that's up to the caller isn't it?  They just need to know the current feerate, and they can calculate the granularity for themselves.  Shouldn't we just put this info in listpeers?

tests/test_opening.py
```diff
@@ -0,0 +1,62 @@
...
+    startweight = 42 + 172  # base weight, funding output
+    initpsbt = l1.rpc.utxopsbt(chan_amount, next_feerate, startweight,
+                               prev_utxos, reservedok=True,
+                               min_witness_weight=110,
+                               excess_as_change=True)
+
+    # Do the bump
+    bump = l1.rpc.openchannel_bump(chan_id, chan_amount, initpsbt['psbt'])
+
+    update = l1.rpc.openchannel_update(chan_id, bump['psbt'])
```
rustyrussell, line 46: It seems weird to me that openchannel_update doesn't take an explicit feerate?

rustyrussell (approved):
Generally good, a few (perhaps naive?) questions.  It's probably OK to break bisect like this, rather than doing makework, but depends if it can be done easily as a side-effect of things you have to do.

lightningd/dual_open_control.c
niftynei: As proposed, the spec allows you to update the total funding amount on an RBF attempt -- it's the only thing you really can change between attempts, all other channel parameters remain the same between the original `open_channel2` message and `init_rbf`.

lightningd/dual_open_control.c
niftynei: Per proposed spec, we use an automatically increasing feerate. Every RBF attempt bumps the feerate by 25% from the last *successful* attempt.

We need to know the feerate to build a new psbt though, so we expose it via an RPC.

niftynei: I feel like you commented elsewhere that we can/should do this via `listpeers`'s channel listing, which I agree is much better than this approach!

tests/test_opening.py
niftynei: The `openchannel_update` refers to updating the *PSBT* that's in progress -- it's how you can add/remove inputs and outputs to the existing PSBT while the channel funding tx negotiation is ongoing.

Feerate's already set at this point, we'd have to add a new spec message to allow you to update it here...

As is, your options for changing a feerate would be to:
  - if first funding tx attempt, abort open, re-try openchannel
  - if first funding tx, finish open, init RBF (with automatic 25% fee bump)
  - if RBF, finish open, init RBF (with automatic 25% fee bump)
  - if RBF, abort / close channel by publishing the commitment tx, reinit channel with openchannel

wallet/wallet.c
niftynei: removing fields from the db (in sqlite) requires 
- copy existing table to temp table
- deleting existing table
- re-creating table without fields
- copying data back from temp table
- deleting temp table

Doable, but probably best to execute in a separate PR with extensive tests.

lightningd/channel.c
```diff
@@ -143,6 +144,45 @@ void get_channel_basepoints(struct lightningd *ld,
...
+	struct channel_inflight inf_init = {
+		.channel = channel,
+		.funding_txid = funding_txid,
+		.funding = funding,
+		.funding_outnum = funding_outnum,
+		.funding_feerate = funding_feerate,
+		.our_funds = our_funds,
+		.remote_tx_sigs = false,
+	};
+	memcpy(inflight, &inf_init, sizeof(struct channel_inflight));
```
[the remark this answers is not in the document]
niftynei: I did this so that I could set the const fields. Is there another way to do this?

niftynei:
Ended up removing the FAIL_RBF message from spec and using the new "warning" message/hanging up instead. It's much cleaner!

Working on getting a few good reconnection tests done, will update with tests! shortly

niftynei:
Moving elsewhere...

PR #4374, opened 2021-02-06
Added regtest, example, k and null explanation

keblek:
network was missing regtest as an option. 

The example was too simple, so I added a second one so show how the k option works.

The explanation of how k works should let the user know that in some cases lightningd will interpret the position of an argument which can lead to a weird response from lightningd.

let the users know that using null is not encouraged

@cdecker said "I think we should definitely not encourage null, since that is very brittle and forces us to keep argument ordering, and deprecating arguments is practically impossible."

Changelog-None

### DIFF
--- a/doc/lightning-cli.1.md
+++ b/doc/lightning-cli.1.md
@@ -25,6 +25,7 @@ Sets configuration file (default: **lightning-dir**/*config* ).
  **--mainnet**
  **--testnet**
  **--signet**
+ **--regtest**
 Sets network explicitly.
 
  **--rpc-file**=*FILE*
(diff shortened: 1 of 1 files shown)


doc/lightning-cli.1.md
```diff
@@ -25,6 +25,7 @@ Sets configuration file (default: **lightning-dir**/*config* ).
  **--mainnet**
  **--testnet**
  **--signet**
+ **--regtest**
```
rustyrussell, line 28: This option does not exist though!  (Maybe it should, but it doesn't!)

keblek: Whoops!

I guess the CLI happily ignores this option because we've been using it for [this docker build](https://github.com/lnbook/lnbook/tree/develop/code/docker/c-lightning) which enables the option through the conf file.

keblek (commented):
removed regtest

doc/lightning-cli.1.md
cdecker: It's just the shorthand `--regtest` that isn't available, `--network regtest` works fine :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4374/commits/ed1db32eaf5cd67a6a91e0212104604a55042af0

Issue #4363, opened 2021-01-28
broken database


gallizoltan:
### Issue and Steps to Reproduce

Every time I start `lightningd`, I see a `**BROKEN** database` message in the log.
Actually everything works fine, but I worry about that message.

```
2021-01-23T12:22:20.116Z INFO    plugin-bcli: bitcoin-cli initialized and connected to bitcoind.
2021-01-23T12:22:20.370Z **BROKEN** database: Accessing a null column 9 in query SELECT  id, channel_htlc_id, msatoshi, cltv_expiry, hstate, payment_hash, payment_key, routing_onion, failuremsg, malformed_onion, origin_htlc, shared_secret, received_time, we_filled FROM channel_htlcs WHERE direction= ? AND channel_id= ? AND hstate != ?
2021-01-23T12:22:20.758Z INFO    lightningd: --------------------------------------------------
2021-01-23T12:22:20.758Z INFO    lightningd: Server started with public key 03d67f36c4f81789e2fe425028bacc96b199813eae426c517f589a45f1136c1fe5, alias Jubilee (color #dc42f4) and lightningd 0.9.3
```
TBH it is not a new thing. I checked my logs, and its first appearance was in December 2019, when I upgraded to `v0.8.0`. Since then, it is in the log at every start.

### `getinfo` output
```
{
   "id": "03d67f36c4f81789e2fe425028bacc96b199813eae426c517f589a45f1136c1fe5",
   "alias": "Jubilee",
   "color": "dc42f4",
   "num_peers": 31,
   "num_pending_channels": 0,
   "num_active_channels": 23,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv4",
         "address": "212.96.34.143",
         "port": 9735
      },
      {
         "type": "torv2",
         "address": "yqygrtvwihl3exb2.onion",
         "port": 9735
      },
      {
         "type": "torv3",
         "address": "ihhsjdncang7ve2rvhuunvy45bxvmh2qwz3vh7mohbouizd36ywnxeid.onion",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv6",
         "address": "::",
         "port": 9735
      },
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "0.9.3",
   "blockheight": 668035,
   "network": "bitcoin",
   "msatoshi_fees_collected": 310625379,
   "fees_collected_msat": "310625379msat",
[...]

PR #4378, opened 2021-02-08
db: Guard access to channel_htlcs.malformed_onion with a NULL check

cdecker:
We fix it twice: once when accessing the field, and a second time by
inserting the default value `0` on insert. The latter is likely to be
preferred, but the latter former is also need to fix existing wallets.

Fixes: #4363

Reported-by: Zoltán Gálli <@gallizoltan>
Changelog-Fixed: db: Fixed an access to a NULL-field in the `channel_htlcs` table and resulting warning.

### DIFF
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -674,6 +674,7 @@ static struct migration dbmigrations[] = {
      * it cuts down the time by 80%.  */
     {SQL("CREATE INDEX forwarded_payments_out_htlc_id"
 	 " ON forwarded_payments (out_htlc_id);"), NULL},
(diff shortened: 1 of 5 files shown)


rustyrussell:
This is a classic case for COMPAT, or even a db migration?

cdecker:
Good point, the migration allows us not to use the `COMPAT` flag and frees us from having to check it as well. In combination with the setting of the default value we now always have the value of `htlc->badonion` set.

rustyrussell (approved):
Ack 1033989dfbdea6f16c25a1b051def26bb382e165

PR #4391, opened 2021-02-19
EXPERIMENTAL support for option_shutdown_anysegwit

rustyrussell:
See https://github.com/lightningnetwork/lightning-rfc/pull/672

Changelog-None: Experimental only

### DIFF
--- a/bitcoin/script.c
+++ b/bitcoin/script.c
@@ -11,35 +11,8 @@
 #include <common/utils.h>
 #include <sodium/randombytes.h>
 
-/* Some standard ops */
-#define OP_0		0x00
+/* To push 0-75 bytes onto stack. */
 #define OP_PUSHBYTES(val) (val)
-#define OP_PUSHDATA1	0x4C
-#define OP_PUSHDATA2	0x4D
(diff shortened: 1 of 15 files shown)


cdecker:
Code looks good, CI had a timeout, restarted :+1:

ACK https://github.com/ElementsProject/lightning/pull/4391/commits/bbe17fc21791a4c981b4322e9727f983ab83c7f3

cdecker:
> It's timing out, even though we disable valgrind.

Ouch, I must've missed that exception (`slow_test -> valgrind=False`) going in, otherwise I'd have objected to it: this is actively harmful, since it suggests we are running with valgrind and that we'd be catching any memleak or invalid access, when in reality we have sneakily disabled valgrind for the test.

We are already testing without valgrind in another configuration, so there is no added value by just letting it run and pretend to be a valgrind test in this config. This test should just be marked as disabled under valgrind.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/5905b530302c2c232811983b847464098885ce29

(Since the slow_test changes are a separate issue)

PR #4394, opened 2021-02-23
Strip the type and length prefix from the custommsg hook call

cdecker:
We were always prefixing the `message` field with the internal type
prefix 0x0407, followed by the length prefix. Neither is needed since
the type being constant is of no interest to the plugin and the length
being implicit due to the JSON-encoding.

Reported-by: Ilya Evdokimov


lightningd/peer_control.c
```diff
@@ -2454,7 +2454,16 @@ static void custommsg_final(struct custommsg_payload *payload STEALS)
 static void custommsg_payload_serialize(struct custommsg_payload *payload,
 					struct json_stream *stream)
 {
+#ifdef COMPAT_V093
```
rustyrussell, line 2457: This is not a COMPAT_V093 thing!  This is a deprecated_apis thing.

COMPAT_V093 means "if you first started this node at 0.9.3, we will still work".  It's mainly for database fields, but can also apply to existing channels with peers.

tests/plugins/custommsg_a.py
```diff
@@ -5,9 +5,9 @@
 
 
 @plugin.hook('custommsg')
-def on_custommsg(peer_id, message, plugin, **kwargs):
+def on_custommsg(peer_id, payload, plugin, message=None, **kwargs):
     plugin.log("Got custommessage_a {msg} from peer {peer_id}".format(
-        msg=message,
+        msg=payload,
         peer_id=peer_id
```
rustyrussell, line 11: This should probably be simply rolled into the previous one, since otherwise there's bisect breakage.

It's pretty small...

rustyrussell (changes requested):

cdecker:
Removed the `COMPAT_V093` flag, replaced it with `deprecated_apis`, and squashed the two commits into one :+1:

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/96e9d79233348754011fb8af26e972788dbb389e

PR #4398, opened 2021-02-24
RBF: Middle part

niftynei:
These patches have been indescriminately split into 3 parts to make it a bit easier(?) to review. This patch starts at 
f08326b.

This patchset finishes the migration to 'unsaved' `channels`, and continues the work on RBF flows.

Changelog-None.


lightningd/dual_open_control.c
```diff
@@ -1252,21 +1251,29 @@ static void handle_peer_tx_sigs_sent(struct subd *dualopend,
 		return;
 	}
 
-	if (psbt_finalize(cast_const(struct wally_psbt *, psbt))) {
-		wtx = psbt_final_tx(NULL, psbt);
+	inflight = channel_current_inflight(channel);
+	if (psbt_finalize(cast_const(struct wally_psbt *,
+				     inflight->funding_psbt))
```
rustyrussell, line 1270: psbt_finalize takes a non-const because it changes it.  That means inflight->funding_psbt needs to be non-const.

casting-away const is required sometimes because the function being called doesn't actually need const (e.g. strstr()).  But it's technically illegal to modify something via a const pointer, even if you cast it.

openingd/dualopend.c
```diff
@@ -1467,6 +1467,52 @@ static bool run_tx_interactive(struct state *state,
...
+	/* We've already checked this */
+	if (!amount_sat_to_msat(&our_msats,
+				state->our_role == TX_INITIATOR ?
+					tx_state->opener_funding :
+					tx_state->accepter_funding))
+		abort();
+
+	/* If this is the first time through, we can't do anything */
+	if (!tx_state)
+		return;
```
rustyrussell, line 1493: We crash if this was true though?

rustyrussell (approved):

lightningd/dual_open_control.c
niftynei: Nice, we can just remove the `const_cast` here then, `inflight->funding_psbt`'s already 'non-const'.

openingd/dualopend.c
niftynei: Right, in fact it's always set to something (we set the original in `main`).

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/4c3df84bf157b0508cff23a462b187528452bee0 after trivial rebase now part 1 is merged.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/4398/commits/4c3df84bf157b0508cff23a462b187528452bee0 from @rustyrussell

PR #4402, opened 2021-02-26
test utils: dont let memleak calls to dead/frozen nodes hang

niftynei:
We time out if the memleak doesn't finish up?! 

I marked `test_multifunding_v2_exclusive` as `@pytest.mark.xfail` in #4397 for the duration of the refactoring job, but it started hanging weirdly on memleak. 

I say "weirdly" because it seems odd to me that the nodes would even get instantiatiated if the test is marked as `xfail`.

I don't know if this is a reasonable solution, but it fixes the problem.

Changelog-None.


niftynei:
Doesn't seem to work as intended in all cases...

PR #4403, opened 2021-02-26
Weekly cleanups

cdecker:
The last few commits may be a bit controversial but I think we should rather skip a test in a valgrind configuration, rather than pretending to run it and then disable valgrind in a roundabout way.

### DIFF
--- a/common/test/run-gossmap_local.c
+++ b/common/test/run-gossmap_local.c
@@ -436,4 +436,5 @@ int main(int argc, char *argv[])
 	/* Now we can refresh. */
 	assert(write(fd, "", 1) == 1);
 	gossmap_refresh(map);
+	common_shutdown();
 }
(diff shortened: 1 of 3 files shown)


rustyrussell:
No: we don't run these under valgrind iff SLOW_MACHINE. 

I run them locally, however.

cdecker:
> No: we don't run these under valgrind iff SLOW_MACHINE.
> 
> I run them locally, however.

I see, however shouldn't we then use `@unittest.skipIf(SLOW_MACHINE and VALGRIND, "...")` and make it explicit?

rustyrussell:
> > No: we don't run these under valgrind iff SLOW_MACHINE.
> > I run them locally, however.
> 
> I see, however shouldn't we then use `@unittest.skipIf(SLOW_MACHINE and VALGRIND, "...")` and make it explicit?

No, this was explicitly introduced to downgrade them because some tests we were not running under CI *at all* due to this.  

They probably get caught by our postgres test these days, though?

(Note, this was all in the commit message which added the slow_test marker, as it should be.)

cdecker:
Ok, I reverted the `pytest.mark.slow_test` removal. Still don't agree that we should be pretending to run them in a valgrind configuration, but without valgrind at all, but ok.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/026a2ef5d356497a1600b1c784cb73621f5f5fe9

Issue #4299, opened 2020-12-27
Keysend enabled on LND but feature bit 55 missing in node announcement


hMsats:
### Issue and Steps to Reproduce

`keysend` is working for c-lightning version `v0.9.2-77-g23d4e53` but not for version `v0.9.2-126-gd971e3d` because it generates the error `feature bit 55 missing in node announcement`.

So it seems that on LND `keysend` is enabled but the feature bit 55 is nonetheless missing. It's not clear to me if this is a c-lightning or LND error or shortcoming.

cdecker:
We signal support for features in the node announcements in the gossip. It takes a while for the node announcement to be sent (not sure how often lnd sends them) and it takes a while for it to reach every key on the network. So this might be a simple issue of a missing update, which causes us to report the error on the preflight checks so to speak.

Could you include the info from `lightning-cli listnodes [recipient_id]` which should include the features.

hMsats:
The node is on 24/7, it outputs:

```
{
   "nodes": [
      {
         "nodeid": "021779d56f6992431e959991a664d678a2f41054073adb48c5a351d791c9cb0781",
         "alias": "Banana!🍌",
         "color": "ffe135",
         "last_timestamp": 1609084278,
         "features": "8252a1",
         "addresses": [
            {
               "type": "ipv4",
               "address": "212.51.132.226",
               "port": 9735
            }
         ]
      }
   ]
}
```

cdecker:
Indeed bit 55 which we interpret as keysend as reverse engineered from lnd is not set:

```
(0x8252a1 >> 55) &0x01 = 0
```

So that warning is correct. It may still be stale information because we haven't seen a newer node announcement, but if this persists then the recipient doesn't signal support correctly. This of course assumes that lnd didn't change the feature bit without telling us.

hMsats:
Well same thing for the node with alias "BALANCER-Keysend-Enabled" (036bd7addc94c938d9db556d967929d8a531a4187e6a8b3aee2dc6be76f50c20bc). Keysend works with c-lightning version `v0.9.2-77-g23d4e53` but same error message for `v0.9.2-126-gd971e3d`. Listnodes gives `"features": "02a2a1"` which of course doesn't have feature bit 55 set to 1. So maybe it's better for the time being to display the error message but try the keysend payment anyway.

cfromknecht:
> This of course assumes that lnd didn't change the feature bit without telling us.

[...]

PR #4404, opened 2021-02-26
`pay` optimizations round 1

cdecker:
This started out as a quick fix for #4299, but while debugging I
stumbled over a couple of easy optimizations:

 - Keysend no longer requires featurebit 55 to be signaled by the
   destination, instead requiring only TLV. This is because apparently
   keysend is not formally proposed to the spec, and the featurebit is
   not reserved. We now will report a more meaningful error message if
   the recipient returns an `INVALID_ONION_PAYLOAD` error.
 - We now filter routehints whose entrypoints we cannot reach, since
   there is no point in trying them if we can't even get to the
   entrypoint. This considers two different scenarios: entrypoint is
   unknown and entrypoint is unreachable.
 - Added `payment_abort` to set a top-level error message in the `pay`
   result, which should make it easier for modifiers to be helpful and
   give a human-readable error message.
   
I have some more improvements planned, but this seemed like a good
initial candidate set.

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1521,6 +1521,11 @@ static void paystatus_add_payment(struct json_stream *s, const struct payment *p
 	/* TODO Add routehint. */
 	/* TODO Add route details */
 
+	if (p->step < PAYMENT_STEP_SPLIT)
+		json_add_string(s, "state", "pending");
+	else
+		json_add_string(s, "state", "completed");
+
 	if (p->step == PAYMENT_STEP_SPLIT) {
 		/* Don't add anything, this is neither a success nor a failure. */
 	} else if (p->result != NULL) {
@@ -1530,7 +1535,7 @@ static void paystatus_add_payment(struct json_stream *s, const struct payment *p
 			json_object_start(s, "failure");
 		json_add_sendpay_result(s, p->result);
 		json_object_end(s);
-	} else {
+	} else if (p->step >= PAYMENT_STEP_SPLIT) {
 		json_object_start(s, "failure");
 		json_add_num(s, "code", PAY_ROUTE_NOT_FOUND);
 		json_add_string(s, "message", "Call to getroute: Could not find a route");
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1961,7 +1961,7 @@ static struct command_result *json_paymod(struct command *cmd,
 	struct payment *p;
 	const char *b11str;
 	struct bolt11 *b11;
-	char *fail;
+	char *b11_fail, *b12_fail;
 	u64 *maxfee_pct_millionths;
 	u32 *maxdelay;
 	struct amount_msat *exemptfee, *msat;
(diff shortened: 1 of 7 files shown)


cdecker:
It seems the test I added is flaky, I'll have a look before undrafting it again.

plugins/pay.c
```diff
@@ -2003,31 +2004,45 @@ static struct command_result *json_paymod(struct command *cmd,
 	p = payment_new(cmd, cmd, NULL /* No parent */, paymod_mods);
 	p->invstring = tal_steal(p, b11str);
 
-	b11 = bolt11_decode(cmd, b11str, plugin_feature_set(cmd->plugin),
-			    NULL, chainparams, &fail);
-	if (b11) {
+	is_b12 = strlen(b11str) > 3 &&
+		 (strstarts(b11str, "lni") || strstarts(b11str, "lno") ||
+		  strstarts(b11str, "lnr"));
```
rustyrussell, line 2009: Hmm, I think it's clearer to use two fail ptrs (esp since this test doesn't work correctly for upper case, and I worry about it breaking if we allow lightning:// prefixes etc)?

Then prefer to the bolt11 failstr if both fail.  Later we can apply some kind of heuristic like this if necessary (maybe create a helper like looks_like_bolt12_invoice())?

rustyrussell: Untested:

```diff
diff --git a/common/bolt12.c b/common/bolt12.c
index 9cff0ab88..174c66430 100644
--- a/common/bolt12.c
+++ b/common/bolt12.c
@@ -289,6 +289,11 @@ struct tlv_invoice *invoice_decode_nosig(const tal_t *ctx,
 	return invoice;
 }
 
+bool smells_like_bolt12_invoice(const char *str)
+{
+	return strstarts(str, "lni1") || strstarts(str, "LNI1");
+}
+
 static void add_days(struct tm *tm, u32 number)
 {
 	tm->tm_mday += number;
diff --git a/common/bolt12.h b/common/bolt12.h
index 797c5b64b..e9edd9da2 100644
--- a/common/bolt12.h
+++ b/common/bolt12.h
@@ -99,6 +99,9 @@ struct tlv_invoice *invoice_decode_nosig(const tal_t *ctx,
 					 const struct chainparams *must_be_chain,
 					 char **fail);
 
+/* For better error msgs when we get a bolt11 or bolt12. */
+bool smells_like_bolt12_invoice(const char *str);
+
 /* Check a bolt12-style signature. */
 bool bolt12_check_signature(const struct tlv_field *fields,
 			    const char *messagename,
```

cdecker:
After a quick detour decoding both and then deciding on which failure was set, I noticed that we'd be masquerading b12 failures with a generic b11 "wrong prefix" error. So I went and added the `bolt12_has_*_prefix` functions to be able to tell them apart from b11 strings, and now it seems to work.

rustyrussell:
Ack 858d473959a6846b09a986dad2d95eee272c5cb6

Issue #4400, opened 2021-02-25
dev-* options in contrib/startup_regtest.sh cause script to fail


jharveyb:
### Issue and Steps to Reproduce

With release 0.9.3 and bitcoind 0.21, running contrib/startup_regtest.sh fails, due to to lightningd startup failing.
The error message: `lightningd: /tmp/l1-regtest/config line 5: dev-fast-gossip: unrecognized option`

A similar error appears after removing the `dev-fast-gossip` option, with reference to `dev-bitcoind-poll=5`.
Removing the `dev-fast-gossip` and `dev-bitcoin-poll=5` options on lines 89 & 90 fixed the issue.

Alternatively, building with the `--enable-developer` flag also resolves this issue, but that requirement isn't mentioned in the main README.

renepickhardt:
can confirm the issue and that both solutions work. 

Is it useful to have those too flags when working in regtest? In that case I would recommend to extend the readme. otherwise one could just remove those two entries to the configfile

niftynei:
Good point, this is definitely a bug. These options turned on makes using/testing things with `startup_regtest` much much faster (blocks are processed much more quickly, the nodes exchange gossip faster, so routing payments across the nodes is possible much sooner).

Will submit a patch to fix shortly.

PR #4408, opened 2021-03-03
added the default bitcoin path for macos to startup_regtest.sh

renepickhardt:
The `contrib/startup_regtest.sh` script does not work out of the box on mac os as the default bitcoin dir in mac os is not `~/.bitcoin` but rather `~/Library/Application Support/Bitcoin/` so I added that

### DIFF
--- a/contrib/startup_regtest.sh
+++ b/contrib/startup_regtest.sh
@@ -54,6 +54,8 @@ fi
 if [ -z "$PATH_TO_BITCOIN" ]; then
 	if [ -d "$HOME/.bitcoin" ]; then
 		PATH_TO_BITCOIN="$HOME/.bitcoin"
+	elif [ -d "$HOME/Library/Application Support/Bitcoin/" ]; then
+		PATH_TO_BITCOIN="$HOME/Library/Application Support/Bitcoin/"
 	else
 		echo "\$PATH_TO_BITCOIN not set to a .bitcoin dir?" >&2
 		return


renepickhardt:
While at this we might want to fix at least the readme of the file to address https://github.com/ElementsProject/lightning/issues/4400

renepickhardt:
Also I don't understand why the smoketests fail when just touching a helper script that has barely anything todo with c-lightning in itself.

contrib/startup_regtest.sh
```diff
@@ -53,10 +53,14 @@ fi
 
 if [ -z "$PATH_TO_BITCOIN" ]; then
 	if [ -d "$HOME/.bitcoin" ]; then
-		PATH_TO_BITCOIN="$HOME/.bitcoin"
+	    PATH_TO_BITCOIN="$HOME/.bitcoin"
 	else
+	    if [ -d "$HOME/Library/Application Support/Bitcoin/" ]; then
```
cdecker, line 58: Maybe better avoid an extra level of indentation using `elif` instead of `else if`.

cdecker (changes requested):

cdecker:
Regarding CI failing, I'll have to give it a got tomorrow, might just be that an error was grandfathered in and manifests now.

cdecker:
Nope, I missed it at first, but it seems you updated the submodule version of libwally-core. Reverting that should work again.

niftynei:
concept ACK, modulo @cdecker's most excellent suggestions for improvement.

thanks for this!

rustyrussell:
OK, did those minor cleanups, repushed.

PR #4411, opened 2021-03-05
hsmd: Ensure the hsmd is initialized before anything else is called

cdecker:
Just a security measure to avoid alternative use-cases of the hsmd
running into the issue that they need to send a `WIRE_HSMD_INIT`
message as first message. If that is not done, the `secretstuff` won't
get initialized and we'd be producing signatures from uninitialized
memory, which are completely useless.

Changelog-None: Internal change only

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -69,6 +69,9 @@ static struct {
 	secp256k1_keypair bolt12;
 } secretstuff;
 
+/* Have we initialized the secretstuff? */
+static bool initialized = false;
+
 /* Version codes for BIP32 extended keys in libwally-core.
  * It's not suitable to add this struct into client struct,
  * so set it static.*/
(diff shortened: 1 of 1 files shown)


hsmd/hsmd.c
```diff
@@ -2133,6 +2149,10 @@ int main(int argc, char *argv[])
 	status_setup_async(status_conn);
 	uintmap_init(&clients);
 
+	/* We start off uninitialized, and require the first message
+	 * to be WIRE_HSMD_INIT. */
+	initialized = false;
```
rustyrussell, line 2154: Generally prefer an explicit initialization at declaration, so it's not reliant on what order things are called in.

(In this case it's implicitly initialized to false, but I still prefer explicit).

rustyrussell (approved):

hsmd/hsmd.c
cdecker: Wan't sure which rule prevails here (don't init until you need, or the init early), but it makes sense here, since we rely on it being set from the very beginning.

PR updated.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4411/commits/4abe16e914a7154ff1d73d67891a94eca32272bb

PR #4413, opened 2021-03-05
plugin: Do not send the internal framed message over the wire

cdecker:
Well this is embarassing 🙄

Looks like #4394 treated a symptom but not the root cause. We were
actually sending the message framed with the WIRE_CUSTOMMSG_OUT and
the length prefix over the encrypted connection to the peer. It just
happened to be a valid custommsg...

This fixes the issue, and this time I made sure we actually send the
raw message over the wire. However for backward compatibility we
needed to imitate the faulty behavior which is 90% of this patch :-)

Changelog-Fixed: plugin: `dev-sendcustommsg` included the type and length prefix when sending a message.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -2881,7 +2881,10 @@ static void handle_dev_memleak(struct peer *peer, const u8 *msg)
  * just forward it here. */
 static void channeld_send_custommsg(struct peer *peer, const u8 *msg)
 {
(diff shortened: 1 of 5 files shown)


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/4413/commits/a7966d673a9a2392c03fa9760765c5a4688a1a0a

PR #4414, opened 2021-03-08
Cache local basepoints in the database

cdecker:
We are currently re-generating the local basepoints and the funding pubkey for each active channel at startup from the `hsmd`. This means that we may end up deriving a lot of keys (5 per channel) when we really could just cache them in the DB alongside the remote ones we already need to store anyway. This also allows us to defer any key-related operation until the channel connection is re-established.

The last commit disables the `NO_PYTHON` CI configuration because it kept complaining about a file that I didn't touch being out of date. IMHO there is little point in singling out python as an optional build dependency, but happy to discuss this further.


rustyrussell:
> The last commit disables the `NO_PYTHON` CI configuration because it kept complaining about a file that I didn't touch being out of date. IMHO there is little point in singling out python as an optional build dependency, but happy to discuss this further.

I understand the frustration, but this patch would also need to update the docs, AFAICT, and should actually remove the NO_PYTHON code.

Hmm, wait, it still seems to require python-mako anyway.  OK, let's disable the test for now, and we'll revisit later (we can avoid checking in generated files, too).

rustyrussell (approved):
Ack bc64a9975fefbdf1aea81a80dad3fbea181ba275

PR #4417, opened 2021-03-09
Weekly cleanups

cdecker:
- We improve the detection of pytest to prefer `python3 -m pytest`
   over the other executables since it is checked for just
   earlier. Also `pytest-3` must not be given special treatment since
   it refers to version 3 of pytest (ancient nowadays) not the python3
   version of pytest.
   
 - Found a regression in `pyln`: the `spent` parameter to the
   `listfunds` call would be always included, which would lead
   c-lightning versions 0.9.2 and older to complain about an unknown
   parameter. Always set these to `None` so pyln can remove them from
   the call, I didn't catch his in my review.

### DIFF
--- a/configure
+++ b/configure
@@ -56,10 +56,17 @@ default_cwarnflags()
 
 default_pytest()
 {
-    PYTEST_BINS="pytest py.test pytest3 pytest-3"
+    # Since we just checked that we have python3 we give that one the
+    # most priority and then fall back to some common aliases.
+    PYTEST_BINS="python3 -m pytest,pytest,py.test,pytest3,pytest-3"
+    IFS=','
     for p in $PYTEST_BINS; do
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -56,10 +56,17 @@ default_cwarnflags()
 
 default_pytest()
 {
-    PYTEST_BINS="pytest py.test pytest3 pytest-3"
+    # Since we just checked that we have python3 we give that one the
+    # most priority and then fall back to some common aliases.
+    PYTEST_BINS="python3 -m pytest,pytest,py.test,pytest3,pytest-3"
+    IFS=','
```
niftynei, line 62: Do we need to revert/unset IFS after this?

niftynei (approved):
ACK adbe20dc243d3cbe841f9d71a785f7016f2eb51e

PR #4419, opened 2021-03-09
pyln: Plugins can be nice if invoked from CLI too

cdecker:
I had way too much fun with this and got a bit carried away with the
letter writing. The idea is to be helpful when users start the plugin
from the command line, rather than run it under the control of
lightningd. We also print detailed information about the user-visible
things such as the methods and options exposed by the plugin.

Changelog-Added: pyln: Plugins that are run from the command line print helpful information on how to configure c-lightning to include them and print metadata about what RPC methods and options are exposed.
Suggested-by: Rusty Russell <@rustyrussell>

### DIFF
--- a/contrib/pyln-client/pyln/client/plugin.py
+++ b/contrib/pyln-client/pyln/client/plugin.py
@@ -724,7 +724,101 @@ class Plugin(object):
 
         return msgs[-1]
 
+    def print_usage(self):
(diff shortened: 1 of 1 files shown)


rustyrussell (approved):
Ack f947faf39d03696ba2804c4c5ba152770a78d018

PR #4420, opened 2021-03-09
Invert multifundchannel.c

niftynei:
Originally written in an inverted style to allow for a more 'natural reading', which relies heavily on forward declarations and allows for an inversion in the normal flow of C code.

This PR removes all the forward declarations, and inverts the code to conform with the style of every other C file in this project.


rustyrussell:
I had to write some scripts to check that it hadn't changed the object file, but it worked.

Good job!

Ack https://github.com/ElementsProject/lightning/pull/4420/commits/cdf57a13cf8848c42638a7f0f43f37985ce7d200

PR #4424, opened 2021-03-11
Dual-funding: handle failures

niftynei:
Prior to this, multifundchannels that involved v2 opens were left in a bit of a lurch if any of the v2 peers had issues while opening.

In theory, we *should* be able to 'live-action' update the PSBT for the remaining peers, if any has an issue, however this would require quite a bit of re-writing of the redo code in `multifundchannel`. 

Plus, it's only really an option is something happens while we're still in the `update` mode.

More easily, we can merely signal to the peer that we're going to restart the negotiation, by sending an error which fails the in-progress negotiation, and then cleanly restart.

For peers that have reached the 'commitment exchange' already, there's no saving these without spending an input so we just ignore them/mark as failed and retry.


rustyrussell (approved):
Ack 1becab47f3f797b6f619819fa20d387a18d067ff

rustyrussell:
Trivial conflict resolve rebase.

Issue #3953, opened 2020-08-17
question about CHANNEL_ERR_TOO_MANY_HTLCS


mloop1:
Hi,
one of my channel with a stable and reliable peer has been unilaterally closed by my node with this status:
`CHANNELD_NORMAL:Bad peer_add_htlc: CHANNEL_ERR_TOO_MANY_HTLCS`
I was wondering under which circustamces this could happen?

I'm running **"version": "v0.9.0-1"** on mainnet.

rustyrussell:
It should not.  It closed on you because you sent more HTLCs than it allows (which you should never do).

I've tried various scenarios and I cannot reproduce this :(  Any more information on what you were doing?

mloop1:
> It should not. It closed on you because you sent more HTLCs than it allows (which you should never do).
> 
> I've tried various scenarios and I cannot reproduce this :( Any more information on what you were doing?

Hi Rusty, thanks for getting back to me.
I was doing nothing while this happened.
Here some more logs, hope this helps, let me know if any other detail could help:

`2020-08-17T11:04:33.650Z UNUSUAL 03a503d8e30f2ff407096d235b5db63b4fcf3f89a653acb6f43d3fc492a7674019-channeld-chan#39763: Adding HTLC too slow: killing connection
2020-08-17T11:04:33.651Z INFO 03a503d8e30f2ff407096d235b5db63b4fcf3f89a653acb6f43d3fc492a7674019-chan#39763: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channe
ld died (9)
2020-08-17T11:05:24.837Z UNUSUAL 03a503d8e30f2ff407096d235b5db63b4fcf3f89a653acb6f43d3fc492a7674019-chan#39763: Peer permanent failure in CHANNELD_NORMAL: channeld: sent ERROR Bad peer_
add_htlc: CHANNEL_ERR_TOO_MANY_HTLCS
2020-08-17T11:05:24.841Z INFO 03a503d8e30f2ff407096d235b5db63b4fcf3f89a653acb6f43d3fc492a7674019-chan#39763: State changed from CHANNELD_NORMAL to AWAITING_UNILATERAL`

whitslack:
My C-Lightning v0.9.0-1 node did the same thing a few days ago.

```
[...]

PR #4432, opened 2021-03-16
channeld: allow remote node to exceed their own HTLC count limits.

rustyrussell:
We try not to exceed either side, but the spec still allows them to
(we don't, but older nodes would have, as could other implementations).

Fixes: #3953
Changelog-Fixed: protocol: overzealous close when peer sent more HTLCs than they'd told us we could send.

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -489,7 +489,6 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	enum side sender = htlc_state_owner(state), recipient = !sender;
 	const struct htlc **committed, **adding, **removing;
 	const struct channel_view *view;
-	u32 min_concurrent_htlcs;
 
 	htlc = tal(tmpctx, struct htlc);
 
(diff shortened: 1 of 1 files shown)


cdecker:
Code seems fine, just for clarification: this is preventing us from adding a new HTLC to an outgoing channel, caused by a forwarded payment, this is not meant to protect against the peer with whom we share the channel from adding more HTLCs than we'd like to have to the channel (which we might be funder of and thus paying fees for), correct?

ACK https://github.com/ElementsProject/lightning/pull/4432/commits/260032bcd0c0f26615a5bfce66097c3fe5463461

rustyrussell:
> Code seems fine, just for clarification: this is preventing us from adding a new HTLC to an outgoing channel, caused by a forwarded payment, this is not meant to protect against the peer with whom we share the channel from adding more HTLCs than we'd like to have to the channel (which we might be funder of and thus paying fees for), correct?
> 
> ACK [260032b](https://github.com/ElementsProject/lightning/commit/260032bcd0c0f26615a5bfce66097c3fe5463461)

No, they could never add more than we allowed.  But they could add more than they allowed!  That's how the spec was written.  This behavior change should have been codified in the spec (that you take the min of the two values), but until it is and everyone is upgraded, we'll break channels if it happens.

PR #4431, opened 2021-03-14
typo: information is an uncountable mass noun

jsarenik:
See https://en.wikipedia.org/wiki/Information

In libplugin.c also the word "details" was added (without removing
the 'information').

Changelog-None

Discussion welcome.

FYI: @darosior

Nice discussion of the topic is at [Englishforums](https://www.englishforums.com/English/InformationOrInformations/gzdmw/post.htm)

PR #4433, opened 2021-03-16
Dual-funding: more reconnection fail

niftynei:
Better testing of packet-drops and reconnects during/when we're doing the signature exchange. This was hard to test exactly because of the damn reconnect logic which resends sigs on reconnect.

"Fixes" the "race" found in #4431's tests.


rustyrussell:
Mechanical rebase to fix generated files.

Ack https://github.com/ElementsProject/lightning/commit/16eb887c1848a19adc256886c0fa9affba1accc4

Issue #4185, opened 2020-11-06
lightning-cli connect to WRONG IP misrepresents success


dscotese:
### Issue and Steps to Reproduce
I misused c-clightning to test whether or not my laptop node is connected with `lightning-cli connect ID@IP:port` from a different machine on a different LAN.  It reports success no matter what IP address I use.  This suggests to me that "connect" is broken, and probably because it's using cached information to "answer the instruction to connect" which is very misleading.

### `getinfo` output
```
{ 
   "id": "03e1b9e6ca5d912254060927794efd3b96a8149f9d360bbec68e693fe7c47a8235",
   "alias": "MemeracingLN",
   "color": "68f442",
   "num_peers": 17,
   "num_pending_channels": 0,
   "num_active_channels": 17,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv4",
         "address": "172.81.183.236",
         "port": 9735
      },
      {
         "type": "torv3",
         "address": "7porym5sb6d6yxil4egvxshzuy2vjgwxhdaei2ri2xvlodtzdcwyolid.onion",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "0.9.0-1",
   "blockheight": 655688,
   "network": "bitcoin",
   "msatoshi_fees_collected": 363316,
   "fees_collected_msat": "363316msat",
   "lightning-dir": "/home/ubuntu/.lightning/bitcoin"
}
```
## Expected Behavior ##
If the node is already connected, output should be prefaced with:
`Already connected to [URI to which it's connected, ID@IP:PORT]:`
This would make it clear when the user has requested to connect on one IP:PORT but c-lightning is NOT doing that but still reporting whatever it reports.

## Current Behavior ##
It currently reports that it is connected without any indication of the IP or port through which is it connected, giving the impression that the IP:PORT provided by the user was successful even when it has not been.

rustyrussell:
Yes, it will try the hint, then fallback to other sources.  We should accurately report the address which *worked*, IMHO.

dscotese:
Simply because I see a principle here that I believe is difficult to recognize, I wanted to highlight my "misuse."  I recognized as misuse while I was doing it and so no problems resulted.  A user "flying blind" may try to remember the correct IP of his node and discover that his first guess was correct!  OR so it seems.  That could create an extremely confusing and unfortunate situation.

[...]

PR #4436, opened 2021-03-16
connect: return address we actually connected to.

rustyrussell:
Otherwise, we might find an address other than the one given and
the user might think that address worked.

Fixes: #4185


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4436/commits/1ecbb2276fe1fb3b2dedfc6ff80c2986d4f2fae8

PR #4452, opened 2021-03-25
Differentiate outgoing vs incoming addresses.

rustyrussell:
This needs tests, but it's actually important: we display addresses, but without knowing if they're incoming or outgoing we can't really tell if they're reliable (particularly if they're from Tor, where address will be a local socket!).


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4452/commits/4c47ed2dfad021202ec2e9779ac5573e5aba5c73

Issue #4376, opened 2021-02-07
Can't pay invoice to Phoenix (ACINQ) custodial wallet with direct channel open to Phoenix node


tukac:
### Issue and Steps to Reproduce

Can't pay invoice to Phoenix wallet (Android, v. 1.4.6-21). I have direct channel open to Phoenix node (02e2670...).

Initiated via plugin RTL 0.10.0. Ends in lightningd crash.

```
2021-02-07T12:49:34.786Z INFO    plugin-plugin.js: pay initiated...
2021-02-07T12:49:34.786Z INFO    plugin-plugin.js: #%d --> %s
2021-02-07T12:49:34.786Z INFO    plugin-plugin.js: [\"lnbc123450n1psplctlpp5dvlnc8rz49at6vlua9km0zektmst8d6kgy34ntxlwtjxl88vj30qdqqxqyjw5q9qtzqqqqqq9qsqsp53wr5vk6g6sml882ntqn52jnf7p9czvfgl0zf8fncrlhkp50vf6psrzjqwryaup9lh50kkranzgcdnn2fgvx390wgj5jd07rwr3vxeje0glcllmxd05mmyvgxsqqqqlgqqqqqeqqjqd7s2pee2pgmype3uztmp9twzldqacqh7m78k0n2xx0hvr7saeha9vhnrrvrteans2m0x2qsat33z6ax2fv5dd0mhyps8fwf9dc69z8cptn7838\",null,null,null,null,null,null,null]
2021-02-07T12:49:34.792Z DEBUG   plugin-pay: cmd 144 partid 0: Added a channel hint for 570805x1483x0/1: enabled false, estimated capacity 25563193msat
2021-02-07T12:49:34.792Z DEBUG   plugin-pay: cmd 144 partid 0: Added a channel hint for 570890x839x1/1: enabled false, estimated capacity 809020547msat
2021-02-07T12:49:34.792Z DEBUG   plugin-pay: cmd 144 partid 0: Added a channel hint for 601291x1345x1/1: enabled false, estimated capacity 262980491msat
2021-02-07T12:49:34.792Z DEBUG   plugin-pay: cmd 144 partid 0: Added a channel hint for 633415x1637x1/1: enabled true, estimated capacity 4294967295msat
2021-02-07T12:49:34.792Z DEBUG   plugin-pay: cmd 144 partid 0: Added a channel hint for 633416x356x0/1: enabled true, estimated capacity 421547340msat
[...]

Issue #4288, opened 2020-12-20
Pay plugin is crashing C-Lightning


saubyk:
### Issue and Steps to Reproduce
```
lightning-cli pay bolt11=lntb120n1p0alpsvpp505fj6wte456trvlrs66zehmv49xf8r2wn5qm4rgqc2nylzxrp9nqdqsxyezqumpwssxxmr5cqzpgxqrrsssp5jlgayagt7z0jj675k2rug545a7pjheqvxn2gk7dp5gjcj0da2yns9qy9qsq67jrj05knzkp2g3h2hqx8e3pfdy09yyf6y8wyjmjp0888gc57mr5addj32m0lu76dmme7jfejguddklea4gxct06rnh6h9rj0sjd2mcq39epl9
```
### Output on the command prompt:
```
lightning-cli: reading response: socket closed
```
### C-Lightning Log:
```
2020-12-20T17:56:26.395Z INFO    lightningd: Server started with public key XXXX, alias XXXXX (color #008000) and lightningd v0.9.2
2020-12-20T17:56:26.414Z INFO    03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134-chan#2: State changed from FUNDING_SPEND_SEEN to ONCHAIN
2020-12-20T17:59:26.017Z INFO    0292f04ef62d32d30c3413d759c851f03726f696a5874b19c070b444789361165d-channeld-chan#81: Peer connection lost
2020-12-20T17:59:26.018Z INFO    0292f04ef62d32d30c3413d759c851f03726f696a5874b19c070b444789361165d-chan#81: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
2020-12-20T18:00:25.976Z INFO    0292f04ef62d32d30c3413d759c851f03726f696a5874b19c070b444789361165d-channeld-chan#81: Peer connection lost
2020-12-20T18:00:25.977Z INFO    0292f04ef62d32d30c3413d759c851f03726f696a5874b19c070b444789361165d-chan#81: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
2020-12-20T18:01:26.129Z INFO    0200a7f20e51049363cb7f2a0865fe072464d469dca0ac34c954bb3d4b552b6e95-channeld-chan#97: Peer connection lost
[...]

PR #4453, opened 2021-03-25
gossipd: fix rolling corruption.

rustyrussell:
If gossip_store is an incorrect version, we will recreate it: with an incorrect
version!  This means we never get persistent gossip, *and* the pay plugin will
fail to map the gossip_store.  Everyone will be sad.

Debugged-by: Matt Whitlock
Fixes: #4376
Fixes: #4288
Typing-by: Rusty Russell <rusty@rustcorp.com.au>


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4453/commits/756ceffec67b9b1c18ddaea5355f4c1840b7de4a

PR #4454, opened 2021-03-25
pyln-proto: write subtype arrays out correctly

niftynei:
We weren't correctly adding the length of an array type in a subtype to the message when writing them out.

We need to iterate through the fields of the subtype to do this, as well as 'descending' through the otherfield vals as we go.

(The `witness_stack` subtype includes two layers of variable length nested subtypes.)

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4454/commits/6e464a0f18719f9041dad9cde695a1ffc7f4578f

PR #4464, opened 2021-03-31
pay: Use gossmap for reachability check

cdecker:
We can simplify the `pay` plugin by not relying on `getroute` anymore and instead use the faster and more flexible `gossmap` implementation. This removes the one last use of `getroute` in the `libplugin-pay` library, and simplifies the reachability check to juts be locally executed.

There is one remaining `getroute` use in `plugins/pay.c`, however that is accessed exclusively through the `legacypay` JSON-RPC method, which is scheduled for removal in 4 months anyway, so I didn't bother to actually remove it. Looking forward to removing `legacypay` :-)

Changelog-None


rustyrussell (approved):
Ack 67ec0d24e6231d0d219f7ff085723041e6b33b35

PR #4471, opened 2021-04-08
subd version messages

rustyrussell:
This catches the "oops, per-peer daemons have been upgraded underneath me!" problem; it still logs a BROKEN message, but it tried to re-exec itself.

We still have the version check at startup, so we don't infinite loop if we have a half-installed setup somehow.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -3040,8 +3040,6 @@ static void init_channel(struct peer *peer)
 
 	assert(!(fcntl(MASTER_FD, F_GETFL) & O_NONBLOCK));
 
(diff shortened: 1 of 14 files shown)


cdecker (approved):
Just a minor question about the ordering in `channeld` where we seem to happily process the `channeld_init` message before we send the version, which seems to contradict our intended goal. Also the `exec` logic is scary, but I can't see how it could fail.

ACK https://github.com/ElementsProject/lightning/pull/4471/commits/92131a01c83d148fe4730e5cd1cf65c480977166

PR #4478, opened 2021-04-14
Dual-funding fixes

niftynei:
Fixes for problems found while writing protocol level tests for dual-funding (https://github.com/rustyrussell/lnprototest/pull/9)

There's a partial patch for https://github.com/ElementsProject/libwally-core/pull/273; partial because a 0-input any-output will fail still. Hmm.


rustyrussell (approved):
Ack 2f24eb50731964c7a9ee45a25f66430418769787

Issue #4152, opened 2020-10-22
Bad commit_sig signature


darosior:
`az0re` on IRC just reported it. Unfortunately Github won't let them get here because they use Tor.
Told them to send the (private) signature, transaction and feerate by mail.

cdecker:
Very interesting indeed, looking at the transaction whose signature failed it seems like the transaction that ultimately made it on-chain looks identical to the one which failed the signature. The fact that the unilateral close was ours suggests that the two consecutive states look identical, narrowing this down to a feerate being the cause for the new commit. However we also add and remove an HTLC that are identical except the `payment_hash`:

```json
       "htlcs": [
	  {
	     "direction": "out",
	     "id": 95,
	     "msatoshi": 125000125,
	     "amount_msat": "125000125msat",
	     "expiry": 653923,
	     "payment_hash": "hashA",
	     "state": "SENT_ADD_ACK_REVOCATION"
	  },
	  {
	     "direction": "out",
	     "id": 96,
	     "msatoshi": 125000125,
	     "amount_msat": "125000125msat",
	     "expiry": 653923,
	     "payment_hash": "hashB",
	     "state": "RCVD_REMOVE_HTLC"
	  }
       ]
```

So it could be that the unilateral close was really the one with `hashB`, and the new (failed commitment) state was the one with `hashA`. The counterparty being an lnd node, which we had synchronization issues in the past with rapid successions of HTLCs suggests that we might have de-synched inbetween the two commits.

0e319b6:
This happened again with a different peer, this time with an empty `htlcs` field, using git cd7d5cdff9e5efc0dcfb5fdc91e8c80a11daebed.  I will send another email with the `listpeers` object and relevant log excerpt.

0e319b6:
This happened yet again. This bug has cost me a ton of money, and fixing it is clearly not a priority. Despite my significant investment in C-Lightning infrastructure, I am starting to migrate to an LND node; I am not willing to keep burning satoshis like this.

I will also warn other people not to use C-Lightning on mainnet until this bug is fixed so they don't get burned, either.

rustyrussell:
OK, the good news is I've caught one of these in my logs, too!  Finally...

@cdecker's idea that it's around feerate changes seems to be borne out here.  Seems like we were updating fees, both sides were on different levels, and then when we tried updating fees again (while they were still in flux), something b0rked.

[...]

PR #4480, opened 2021-04-18
channeld: don't ever send back-to-back feerate changes, fix feerate logic.

rustyrussell:
There are several reports of desynchronizaion with LND here; a simple
approach is to only have one feerate change in flight at any time.

But I also found that our fee state machine was a bit screwed (thanks to CI flake!), so that may also have caused this problem.

Fixes: #4152

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -1052,7 +1052,9 @@ static void send_commit(struct peer *peer)
 	 */
 	changed_htlcs = tal_arr(tmpctx, const struct htlc *, 0);
 	if (!channel_sending_commit(peer->channel, &changed_htlcs)) {
-		status_debug("Can't send commit: nothing to send");
+		status_debug("Can't send commit: nothing to send, feechange %s (%s)",
+			     want_fee_update(peer, NULL) ? "wanted": "not wanted",
+			     type_to_string(tmpctx, struct fee_states, peer->channel->fee_states));
 
 		/* Covers the case where we've just been told to shutdown. */
 		maybe_send_shutdown(peer);
@@ -1403,8 +1405,13 @@ static void handle_peer_commit_sig(struct peer *peer, const u8 *msg)
 	status_debug("Received commit_sig with %zu htlc sigs",
 		     tal_count(htlc_sigs));
 
-	return send_revocation(peer,
-			       &commit_sig, htlc_sigs, changed_htlcs, txs[0]);
+	send_revocation(peer,
+			&commit_sig, htlc_sigs, changed_htlcs, txs[0]);
+
+	/* This might have synced the feerates: if so, we may want to
+	 * update */
+	if (want_fee_update(peer, NULL))
+		start_commit_timer(peer);
 }
 
 /* Pops the penalty base for the given commitnum from our internal list. There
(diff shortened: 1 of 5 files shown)


common/fee_states.c
```diff
@@ -73,6 +73,16 @@ u32 get_feerate(const struct fee_states *fee_states,
...
 }
 
+/* Are feerates all agreed by both sides? */
+bool feerate_changes_done(const struct fee_states *fee_states)
+{
+	size_t num_feerates = 0;
+	for (size_t i = 0; i < ARRAY_SIZE(fee_states->feerate); i++) {
+		num_feerates += (fee_states->feerate[i] != NULL);
+	}
+	return num_feerates == 1;
```
0e319b6, line 83: Is it really necessary to iterate over the entire array? At the very least can't we short-circuit the loop if there is more than one non-NULL entry?

    if (num_feerates == 0 && fee_states->feerate[i] != NULL)
        num_feerates = 1;
    else if (num_feerates == 1 && fee_states->feerate[i] != NULL)
        return false;
    return true

channeld/channeld.c
```diff
@@ -944,6 +944,25 @@ static struct bitcoin_signature *unraw_sigs(const tal_t *ctx,
...
+
+	max = approx_max_feerate(peer->channel);
+	*target = peer->desired_feerate;
+
+	/* FIXME: We should avoid adding HTLCs until we can meet this
+	 * feerate! */
+	if (*target > max)
+		*target = max;
+
+	return *target != channel_feerate(peer->channel, REMOTE);
```
0e319b6, line 963: I guess this behavior is probably defined in a BOLT somewhere but shouldn't we be using *our* desired feerate, negotiating with the peer to keep within *their* min and max, and also ensuring we never go above *our* max? Rather than just using *their* desired feerate (for a commitment tx *we're* paying for, for Pete's sake!) and merely ensuring that it doesn't exceed our max feerate?

common/fee_states.c
rustyrussell: I prefer the simpler version, TBH; it's pretty marginal for a 16-entry array...

channeld/channeld.c
rustyrussell: As funder, we set the feerates; the comparison here is simply "the feerate they are currently using".

We don't know the peer's allowable range; we just know what we want, and what we can afford (approx_max_feerate).

common/fee_states.c
0e319b6: OK fair enough. 16 entries is no big deal; I didn't realize it was so small.

It's just nitpicking, but IMO a pure Boolean logic implementation is simpler and clearer. Also it avoids the risk of integer overflow (not going to happen with 16 elements but generally speaking):
```
bool has_inflight_feerate = false;
// ...
for (size_t i = 0; i < ARRAY_SIZE(fee_states->feerate); i++) {
    if (fee_states->feerate[i] != NULL) {
        if (!has_inflight_feerate)
            has_inflight_feerate = true;
        else
            return false;
    }
}
return true;
```

channeld/channeld.c
0e319b6: Oh, I see. That makes sense :).

cdecker:
ACK https://github.com/ElementsProject/lightning/commit/16ea05f5f2323efadcc2b876cdbe517689562563

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4480/commits/bf7947bd1cafe06d6f3f4d886de314cd3a437088

PR #4489, opened 2021-04-22
add a `funder` plugin

niftynei:
Plugin to democratize a peer's ability to contribute funds to incoming openchannel request.

Current knobs include:
 - three kinds of policy: match (% of their funding), available (% of your available funds), and fixed (sat value)
 - minimum and maximum funding from peer
 - minimum and maximum that you'll put in a channel
 - fuzzing (defaults to 5%)
 - fund_probability. only funds X% of incoming requests
 - reserve-tank. never* spends available funds beneath this amount

*assuming races never happen

Future work: we don't take feerates into account right now we calculating the total available funds to add.


rustyrussell:
Here's proposed patch for the tmp_channel_id issue: spec should probably be updated too?

```diff
dualopend: don't use final channel_id for accepter_start2

The other side doesn't know it until *after* it parses this msg.  We
add a quick hack to still allow old nodes to work (for now!).

This also fixes a bug (spotted by @niftynei) where any errors we sent
before accepter_start2 would have the new (unknowable!) channel_id
rather than the temp one.

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

diff --git a/openingd/dualopend.c b/openingd/dualopend.c
index ad8d5420e..d753dff8e 100644
--- a/openingd/dualopend.c
+++ b/openingd/dualopend.c
@@ -1897,7 +1897,6 @@ static void accepter_start(struct state *state, const u8 *oc2_msg)
 	struct bitcoin_blkid chain_hash;
 	struct tlv_opening_tlvs *open_tlv;
 	char *err_reason;
-	struct channel_id tmp_chan_id;
 	u8 *msg;
 	struct amount_sat total;
 	enum dualopend_wire msg_type;
@@ -1939,7 +1938,7 @@ static void accepter_start(struct state *state, const u8 *oc2_msg)
 	 * `open_channel2`), a temporary `channel_id` should be found
 	 * by using a zeroed out basepoint for the unknown peer.
 	 */
-	derive_tmp_channel_id(&tmp_chan_id,
+	derive_tmp_channel_id(&state->channel_id,
 			      &state->their_points.revocation);
 	if (!channel_id_eq(&state->channel_id, &tmp_chan_id))
 		negotiation_failed(state, "open_channel2 channel_id incorrect."
@@ -1949,12 +1948,6 @@ static void accepter_start(struct state *state, const u8 *oc2_msg)
 				   type_to_string(tmpctx, struct channel_id,
 						  &state->channel_id));
 
-	/* Everything's ok. Let's figure out the actual channel_id now */
-	derive_channel_id_v2(&state->channel_id,
-			     &state->our_points.revocation,
-			     &state->their_points.revocation);
-
-
 	/* Save feerate on the state as well */
 	state->feerate_per_kw_funding = tx_state->feerate_per_kw_funding;
 
@@ -2105,6 +2098,11 @@ static void accepter_start(struct state *state, const u8 *oc2_msg)
 				     &state->first_per_commitment_point[LOCAL],
 				     a_tlv);
 
+	/* Everything's ok. Let's figure out the actual channel_id now */
+	derive_channel_id_v2(&state->channel_id,
+			     &state->our_points.revocation,
+			     &state->their_points.revocation);
+
 	sync_crypto_write(state->pps, msg);
 	peer_billboard(false, "channel open: accept sent, waiting for reply");
 
@@ -2479,6 +2477,23 @@ static void opener_start(struct state *state, u8 *msg)
 		open_err_fatal(state,  "Parsing accept_channel2 %s",
 			       tal_hex(msg, msg));
 
+	if (!channel_id_eq(&cid, &state->channel_id)) {
+		struct channel_id future_chan_id;
+		/* FIXME: v0.10.0 actually replied with the complete channel id here,
+		 * so we need to accept it for now */
+		derive_channel_id_v2(&future_chan_id,
+				     &state->our_points.revocation,
+				     &state->their_points.revocation);
+		if (!channel_id_eq(&cid, &future_chan_id)) {
+			peer_failed_err(state->pps, &cid,
+					"accept_channel2 ids don't match: "
+					"expected %s, got %s",
+					type_to_string(msg, struct channel_id,
+						       &state->channel_id),
+					type_to_string(msg, struct channel_id, &cid));
+		}
+	}
+
 	if (a_tlv->option_upfront_shutdown_script) {
 		state->upfront_shutdown_script[REMOTE]
 			= tal_steal(state,
@@ -2492,14 +2507,6 @@ static void opener_start(struct state *state, u8 *msg)
 			     &state->our_points.revocation,
 			     &state->their_points.revocation);
 
-	if (!channel_id_eq(&cid, &state->channel_id))
-		peer_failed_err(state->pps, &cid,
-				"accept_channel2 ids don't match: "
-				"expected %s, got %s",
-				type_to_string(msg, struct channel_id,
-					       &state->channel_id),
-				type_to_string(msg, struct channel_id, &cid));
-
 	/* Check that total funding doesn't overflow */
 	if (!amount_sat_add(&total, tx_state->opener_funding,
 			    tx_state->accepter_funding))
```

niftynei:
Ok, patches have been applied (and already folded in mea culpa).

plugins/libplugin.h
```diff
@@ -219,6 +219,10 @@ struct command_result *forward_result(struct command *cmd,
  * send_req() path. */
 struct command_result *timer_complete(struct plugin *p);
 
+/* Signals that we've completed a command. Useful for when
+ * there's no `cmd` present */
+struct command_result *command_done(void);
+
```
rustyrussell, line 225: Drop this commit.  The one place you end up using it, you do indeed have a command.  You just don't use it!

plugins/funder.c
```diff
@@ -335,17 +339,135 @@ psbt_fund_failed(struct command *cmd,
...
+				JSON_SCAN(json_to_sat, &val),
+				JSON_SCAN_TAL(cmd, json_strdup, &status),
+				JSON_SCAN(json_to_bool, &is_reserved));
+		if (err)
+			plugin_err(cmd->plugin,
+				   "`listfunds` payload did not scan. %s: %*.s",
+				   err, json_tok_full_len(result),
+				   json_tok_full(buf, result));
+
+		/* we skip resrved funds */
```
rustyrussell, line 380: Need to buy a vwl?

plugins/funder_policy.c
```diff
@@ -89,6 +89,32 @@ default_funder_policy(enum funder_opt policy,
 				 100);
 }
 
+char *funder_check_policy(struct funder_policy *policy)
```
rustyrussell, line 92: *const* struct funder_policy?

plugins/funder.c
```diff
@@ -586,6 +586,138 @@ static void json_channel_open_failed(struct command *cmd,
...
+		current_policy.reserve_tank = *reserve_tank;
+	if (fuzz_factor)
+		current_policy.fuzz_factor = *fuzz_factor;
+	if (fund_probability)
+		current_policy.fund_probability = *fund_probability;
+
+	err = funder_check_policy(&current_policy);
+	if (err)
+		return command_done_err(cmd, JSONRPC2_INVALID_PARAMS,
+					err, NULL);
```
rustyrussell, line 765: I think you need to modify a copy, then assign it back to current_policy at the end. 

Otherwise, setting a bad value causes an error, but sticks around...

plugins/funder.c
```diff
@@ -586,6 +586,138 @@ static void json_channel_open_failed(struct command *cmd,
...
+
+	res = jsonrpc_stream_success(cmd);
+	policy_to_json(res, &current_policy);
+	return command_finished(cmd, res);
+}
+
+const struct plugin_command commands[] = { {
```
rustyrussell, line 772: *static* const struct plugin_command commands?

doc/PLUGINS.md
```diff
@@ -1157,7 +1157,8 @@ requests an RBF for a channel funding transaction.
     "funding_feerate_per_kw": 7500,
     "feerate_our_max": 10000,
     "feerate_our_min": 253,
-    "locktime": 2453,
+    "channel_max": "200000msat",
```
rustyrussell, line 1160: It will never be this value though, so it's misleading.

lightningd/dual_open_control.c
```diff
@@ -206,6 +209,8 @@ rbf_channel_hook_serialize(struct rbf_channel_payload *payload,
 		     payload->feerate_our_min);
 	json_add_num(stream, "funding_feerate_per_kw",
 		     payload->funding_feerate_per_kw);
+	json_add_amount_sat_only(stream, "channel_max",
+				 payload->channel_max);
```
rustyrussell, line 213: Since this converts to msat, and that fails for UINT64_MAX, this channel_max will actually only appear for non-wumbo channels.  Which is probably OK, but should be documented.

Field should also be called "channel_max_msat".

rustyrussell (commented):
Minor fixes only...

plugins/libplugin.h
niftynei: I don't want to actually do anything with the command though. There's no response to send on our side when this completes, it's a 'silent no-op'?

lightningd/dual_open_control.c
niftynei: ACK. also updated the `channel_max` field that exist on `openchannel2` hook.

niftynei:
Updated, as per comments.

plugins/libplugin.h
rustyrussell: OK, this is because the "cmd" is actually a notification, which doesn't want a response.  It does, however, want to be freed!

I'll submit a followup PR which cleans this up everywhere...

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/de9102a270814b6e57592420bfc1a2c30b6dca9e

PR #4496, opened 2021-04-29
plugin: Allow plugins to publish and subscribe to custom notifications

cdecker:
So far plugins could only subscribe to native notifications that originated in `lightningd` and which were limited to a few topics directly built into the daemon. With this PR we enable plugins to announce their intention to publish notifications to custom topics, and for other plugins to subscribe to those topics and receive the notifications.

Custom topics are required to be announced as part of the `getmanifest` call, allowing us to verify that subscriptions match the announced topics at startup, and print a warning message should a plugin subscribe to a topic that wasn't announced, and therefore will never receive a notification. Trying to send a notification with an unannounced topic will result in a warning being logged, and the message _not_ being delivered to subscribers.

To allow subscribers to identify the sender we wrap the notification payload in an outer object with the sender metadata alongside the payload. So the `sender` sending the first message below, will result in the second message being delivered to all subscribers:

```json
{
  "jsonrpc": "2.0",
  "method": "mycustomnotification",
  "params": {
    "key": "value",
	"message": "Hello fellow plugin!"
  }
}
```

is delivered as

```json
{
  "jsonrpc": "2.0",
  "method": "mycustomnotification",
  "params": {
    "origin": "sender",
    "payload": {
      "key": "value",
      "message": "Hello fellow plugin!"
    }
  }
}

```

Notification subscribers should be lenient in what they accept, given that multiple plugins may publish to the same topic, potentially with different payload structures. This in turn allows for both fanout and fanin communication patterns.

We also instrument the `pay` plugin, reporting a minimal `pay_success` or `pay_failure` notification, which includes the `payment_hash`, the `bolt11` string as well as the failure message for failures. The remaining fields may be added if required or can be looked up by querying `listpays`.

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -417,6 +417,18 @@ static const char *plugin_notify_handle(struct plugin *plugin,
 	return NULL;
 }
 
+/* Check if the plugin is allowed to send a notification of the
+ * specified topic, i.e., whether the plugin has announced the topic
+ * correctly in its manifest. */
+static bool plugin_notification_allowed(const struct plugin *plugin, const char *topic)
+{
+	for (size_t i=0; i<tal_count(plugin->notification_topics); i++)
+		if (streq(plugin->notification_topics[i], topic))
+			return true;
+
+	return false;
+}
+
 /* Returns the error string, or NULL */
 static const char *plugin_notification_handle(struct plugin *plugin,
 					      const jsmntok_t *toks)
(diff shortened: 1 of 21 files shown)


lightningd/plugin.c
```diff
@@ -103,6 +104,25 @@ void plugins_free(struct plugins *plugins)
...
+/* Check that all the plugin's subscriptions are actually for known
+ * notification topics. Emit a warning if that's not the case, but
+ * don't kill the plugin. */
+static void plugin_check_subscriptions(struct plugins *plugins,
+				       struct plugin *plugin)
+{
+	if (plugin->subscriptions == NULL)
+		return;
+
+	for (size_t i = 0; i < tal_count(plugin->subscriptions); i++) {
```
rustyrussell, line 115: Nit: tal_count(NULL) == 0, so this loop doesn't need a precursor check.

lightningd/plugin.c
```diff
@@ -1273,6 +1273,44 @@ static void plugin_manifest_timeout(struct plugin *plugin)
...
+
+	if (!notifications)
+		return NULL;
+
+	if (notifications->type != JSMN_ARRAY)
+		return tal_fmt(plugin,
+			       "\"result.notifications\" is not an array");
+
+	for (size_t i = 0; i < notifications->size; i++) {
+		obj = json_get_arr(notifications, i);
```
rustyrussell, line 1320: `json_for_each_arr(i, obj, notifications)` does these two lines, BTW (and slightly more efficiently).

plugins/libplugin-pay.c
```diff
@@ -1834,6 +1834,21 @@ static void payment_json_add_attempts(struct json_stream *s,
 	json_array_end(s);
 }
 
+static void payment_notify_failure(struct payment *p, const char *error_message) {
+	struct payment *root = payment_root(p);
+	struct json_stream *n;
+			n = plugin_notification_start(p->plugin, "pay_failure");
```
rustyrussell, line 1840: Nit: genuinely weird indent?

rustyrussell (approved):
Ack 9faa9071d91a50913d51ddc029956df579ff3db2

Minor cleanups I will send as separate PR.

PR #4497, opened 2021-04-29
libhsmd: Extract the core of `hsmd` into a separate library

cdecker:
We have seen a couple of projects trying to provide alternative `hsmd` implementations for c-lightning, such as remote signers, threshold based signers, etc. They usually have to go and reimplement the entire subdaemon just to make their changes, so I wanted to help by turning the core functionality of `hsmd` into a library that can be used from a couple of other languages such as Rust, Python and Javascript.

This is the first part of that, separating the IO handling and details
for handing out client connections from the core logic of the
`hsmd`. Language bindings will also follow soon.


rustyrussell (approved):
Ack https://github.com/ElementsProject/lightning/pull/4497/commits/e0065b0ba129631f678fe8ae47e71251af2e31d3

rustyrussell:
I'm sure I reviewed this before, but obv didn't hit commit.  My only comment was on externs, and they are mostly removed by the end.

PR #4498, opened 2021-04-29
libhsmd: Python bindings for `libhsmd`

cdecker:
The title pretty much sums it up. Not sure if there is a better way to
compile than to list all the C files that are involved, but it works
:-)

Depends #4497


hsmd/hsmd.c
```diff
@@ -61,29 +62,20 @@
...
-} secretstuff;
-
-/* Have we initialized the secretstuff? */
-static bool initialized = false;
-
 /* Version codes for BIP32 extended keys in libwally-core.
  * It's not suitable to add this struct into client struct,
  * so set it static.*/
-static struct  bip32_key_version  bip32_key_version;
+extern struct  bip32_key_version  bip32_key_version;
```
rustyrussell: extern in a C file is an anti-pattern; we usually want it in a header so both the user and implementer are *definitely* using the same type...

rustyrussell (approved):
Ack 3ffe97bb960861a9489ba122e4650a26be7cbcfb

I closed my eyes through the SWIG part, but I think I was supposed to.  My complaint about externs is only a few dev flags by the end, so that's fine.

rustyrussell:
Needs a changelog entry I think!

hsmd/hsmd.c
cdecker: Yeah, I was hoping I could just move everything across and then remove the `extern` from `hsmd.c` but since we still do some operations in `hsmd.c` like overriding the privkey we need to keep a way to reference it. I could move it to the `libhsmd.h` file, but that breaks the encapsulation. Maybe we find a good way to move it into libhsmd.

cdecker:
> I closed my eyes through the SWIG part, but I think I was supposed to.

Yep, should have mentioned that, it's generated code, which is also why we need to ignore it for `cppcheck` and exclude the `libhsmd.py` file from linting.

cdecker:
> Needs a changelog entry I think!

Changelog added to both PRs :+1:

rustyrussell (approved):
Ack bfa52af97cfd1655d9838395a89712a25e49fd1f

PR #4499, opened 2021-04-30
Expand dual-funded path test coverage

niftynei:
Now that we have the `funder` plugin, we can get rid of the hacky df-accepter test harness plugin.

While we're at it, let's expand test coverage to run the tests with both funding pathways.

Starts at bdf984b.

Includes missed (incredibly hacky) patch for libwally's 'empty PSBT' parsing bug (https://github.com/ElementsProject/libwally-core/pull/274) and a new convenience marker for marking a test as "DEVELOPER=1"


rustyrussell (commented):
Youch, that just doubled our test matrix size.  Worth it?

niftynei:
No, I dont think so.

Better to annotate the subset that are worth running for both and have
those run when the v2 gets run, instead of flagging everything.

Will update

rustyrussell (approved):
Ack https://github.com/ElementsProject/lightning/commit/a705b6eaa9958749e1a31e519f32e8fb7440bd8d

PR #4286, opened 2020-12-18
Make some command docs clearer and more descriptive

fiatjaf:
I was very confused with these descriptions that said only "a string that represents a psbt" and no examples of what that string looked like. I wanted to know if it was a base64 string or what (although maybe someone experienced with PSBT stuff would know already), so I made these things more explicit.

PR #4501, opened 2021-05-01
JSON schemas in doc/ (only some commands so far)

rustyrussell:
This is the result of my json schema adventures so far; we'll see what it looks like when I've done commands ~D~ H through Z :)

It both checks the schema validity on every command during a test run, and generates documentation for us.

Inspired by #4286


cdecker (approved):
This is a great improvement, I don't envy you for having to dig through all these commands, but the final result is just awesome :+1:

ACK https://github.com/ElementsProject/lightning/pull/4501/commits/a65fd7f968d8b8f3c2d50bebc86e54d3819996a4

rustyrussell:
Rebase and fixed obv typo in Makefile.

Ack https://github.com/ElementsProject/lightning/pull/4501/commits/b22e8ac17cae08dd96152640f754019a4cb91475

PR #4502, opened 2021-05-01
plugins/bcli: strip trailing whitespace appropriately

### LINKED ISSUE #4001: bcli: Bitcoin backend assumes Linux-style newlines
A user reported this error on [reddit]:

> So I have confirmed that the bitcoin-cli.exe is being invoked, with the arguments shown. But I am not sure why this error is coming up. If I run bitcoin-cli.exe from command line, I can easily get replies for command like getblockchaininfo, which confirms that bitcoind is running.
>
> Here is what I get as output of lightningd command:
>
> ```
> 2020-08-30T17:35:53.634Z INFO plugin-bcli: bitcoin-cli initialized and connected to bitcoind.
> /usr/local/bin/../libexec/c-lightning/plugins/bcli error: bad response to getrawblockbyheight (bad 'result' field), response was {"jsonrpc":"2.0","id":8,"error":{"code":400,"message":"bitcoin-cli.exe -datadir=G:\\\\.bitcoin -rpcconnect=127.0.0.1 -rpcport=8332 -rpcuser=... -rpcpassword=... getblockhash 646000: bad JSON: bad blockhash (00000000000000000001ef19be9c9879c6e9aa6241a096f543109e2a34397936\r\n)"} }
> ```
>
> Not sure why I am seeing this problem.
> 
> Update: I got it running. Now I see this
>
> ```
> 2020-08-30T23:45:47.611Z INFO plugin-bcli: bitcoin-cli initialized and connected to bitcoind.
> 2020-08-30T23:45:48.728Z INFO lightningd: --------------------------------------------------
> 2020-08-30T23:45:48.729Z INFO lightningd: Server started with public key 03c7505d1ac56441f234025cd348743cddff147f9356e6296820d5e95d1a56b29d, alias GREENTOLL (color #03c750) and lightningd v0.9.0-197-gdd8cd81-modded
> ```
> I've no clue what I am doing, but will continue digging more.

From the `datadir` specified in the `bitcoin-cli` call it seems like this is
running on Windows (cygwin or MingW), and the error message seems to indicate
like it got a windows-style `\r\n` newline rather than the `\n` newline we
expect:

https://github.com/ElementsProject/lightning/blob/71e713c72211fcb3de019ce1f926ca67ff71cd62/plugins/bcli.c#L657-L662

We should likely strip whitespace at the end in a more reliable fashion.

nalinbhardwaj:
Fixes #4001

### DIFF
--- a/plugins/bcli.c
+++ b/plugins/bcli.c
@@ -320,6 +320,15 @@ start_bitcoin_cli(const tal_t *ctx,
 	next_bcli(bcli->prio);
 }
 
+static void strip_trailing_whitespace(char *str, size_t len)
+{
+	size_t stripped_len = len;
+	while (stripped_len > 0 && cisspace(str[stripped_len-1]))
+		stripped_len--;
+
+	str[stripped_len] = 0x00;
+}
+
 static struct command_result *command_err_bcli_badjson(struct bitcoin_cli *bcli,
 						       const char *errmsg)
 {
@@ -595,8 +604,7 @@ static struct command_result *process_getrawblock(struct bitcoin_cli *bcli)
 	struct json_stream *response;
 	struct getrawblock_stash *stash = bcli->stash;
 
-	/* -1 to strip \n and steal onto the stash. */
-	bcli->output[bcli->output_bytes-1] = 0x00;
+	strip_trailing_whitespace(bcli->output, bcli->output_bytes);
 	stash->block_hex = tal_steal(stash, bcli->output);
 
 	response = jsonrpc_stream_success(bcli->cmd);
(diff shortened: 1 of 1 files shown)


plugins/bcli.c
```diff
@@ -595,8 +595,12 @@ static struct command_result *process_getrawblock(struct bitcoin_cli *bcli)
 	struct json_stream *response;
 	struct getrawblock_stash *stash = bcli->stash;
 
-	/* -1 to strip \n and steal onto the stash. */
-	bcli->output[bcli->output_bytes-1] = 0x00;
+	/* Strip trailing whitespace and steal onto the stash. */
+	size_t striplen = bcli->output_bytes;
+	while (isspace(bcli->output[striplen-1]))
```
rustyrussell, line 600: We have a cisspace helper which is sightly better here, BTW.

Can we extract this into a "static void strip_trailing_whitespace(char *str, size_t len)" 
which also makes sure it doesn't underflow if it's all whitespace?

Thanks!

rustyrussell (changes requested):

plugins/bcli.c
nalinbhardwaj: Done.

rustyrussell (approved):
Ack bd59929d91f72dfd080f9861edc795d0443868fc

PR #4509, opened 2021-05-04
Funder fees

niftynei:
Not accounting for our fees when we're calculating the total available for funding bit @fiatjaf 

If a utxo isn't economical, we shouldn't add it to the transaction.

### DIFF
--- a/plugins/funder.c
+++ b/plugins/funder.c
@@ -350,7 +350,7 @@ listfunds_success(struct command *cmd,
 		  const jsmntok_t *result,
 		  struct open_info *info)
 {
-	struct amount_sat available_funds;
+	struct amount_sat available_funds, est_fee;
 	const jsmntok_t *outputs_tok, *tok;
 	struct out_req *req;
 	size_t i;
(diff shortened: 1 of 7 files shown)


plugins/funder.c
```diff
@@ -219,16 +220,20 @@ json_openchannel2_sign_call(struct command *cmd,
...
+		if (psbt_input_is_ours(&psbt->inputs[i])) {
 			json_add_num(req->js, NULL, i);
+			count++;
+		}
 	}
 	json_array_end(req->js);
 
 	plugin_log(cmd->plugin, LOG_DBG,
-		   "calling `signpsbt` for channel %s",
+		   "calling `signpsbt` for channel %s for %zu input%s",
```
vincenzopalazzo, line 233: Maybe there is missing space in `input%s`?

vincenzopalazzo (changes requested):
ack https://github.com/ElementsProject/lightning/commit/d2ab0b14b15cd3566acdd0fbbda967e97d30f927 with only a space missing I think

plugins/funder.c
niftynei: The last `%s` is for gramamtical reasons. It'll print "1 input" or "2 inputs".

vincenzopalazzo: My big mistake, sorry I only take into consideration the line and not the meaning! (bad job vincent :) ).

vincenzopalazzo (approved):

rustyrussell:
Ack 1d1d64ea6a6caf87db0deaa4cec64d17a80f5e01

PR #4507, opened 2021-05-04
Reduce default fees

rustyrussell:

PR #4513, opened 2021-05-06
Suggested code cleanup by TODO comment

vincenzopalazzo:
This PR ~~(it is not ready there is a new TODO, but I use this to check better the test with Github action)~~ introduces a code cleanup suggested by the TODO comment in the code (I don't know from who, but I guess @cdecker ).

Basically, it moves the code from the if-else statement to a switch statement without the default case. I used the basic idea of the code used in PR #4507.

I'm not sure that this code is better than the code before, also because I'm learning C, so this PR could be a big mistake

~~Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>~~


wallet/wallet.c
```diff
@@ -381,7 +406,6 @@ bool wallet_unreserve_output(struct wallet *w,
 					   OUTPUT_STATE_RESERVED,
 					   OUTPUT_STATE_AVAILABLE);
 }
-
```
cdecker, line 384: Spurious change.

wallet/wallet.c
```diff
@@ -31,6 +31,31 @@
 /* 12 hours is usually enough reservation time */
 #define RESERVATION_INC (6 * 12)
 
+/* Possible channel state */
+enum channel_state_db {
```
cdecker, line 35: Nit: the `_db` suffix suggested to me that this is something we store in the DB, which requires extra care since those values are persistent. However it's just an internal value that doesn't reach the DB, so I suggest dropping the `_db` suffix.

wallet/wallet.c
```diff
@@ -1368,6 +1392,16 @@ bool wallet_init_channels(struct wallet *w)
 	return wallet_channels_load_active(w);
 }
 
+static enum channel_state_db get_state_channel_db(const char *dir, const char *type)
+{
+	for (int i = 0; i < CHANNEL_STATE_DB_SIZE; i++) {
+		struct channel_state_param state = channel_state_params[i];
+		if (streq(state.dir_key, dir) && streq(state.type_key, type))
+			return state.state;
+	}
```
cdecker: You can do one better. If you look at the above order you'll see that the `enum` has the structure you can use:

```
enum channel_state = IN_OFFERED;
if (streq(dir, "out"))
  channel_state += 2
if (streq(type, "fulfilled"))
  channel_state += 1
```

This works because of the order of your `enum`: least significant bit indicates state, most significant bit indicates direction. If you decide you want to do this I also suggest making this a bit more explicit by changing the above definition to this:

```
enum channel_state_db {
	IN_OFFERED = 0,
	IN_FULLFILLED = 1,
	OUT_OFFERED = 2,
	OUT_FULLFILLED = 4,
};
```

wallet/wallet.c
cdecker: Maybe a more descriptive name would be `channel_stat_bucket`?

wallet/wallet.c
```diff
@@ -31,6 +31,31 @@
...
+	const char *type_key;
+	const enum channel_state_db state;
+};
+
+/* possible channels states */
+struct channel_state_param channel_state_params[] = {
+	{"in", "offered", IN_OFFERED}, {"in", "fulfilled", IN_FULLFILLED},
+	{"out", "offered", OUT_OFFERED}, {"out", "fulfilled", OUT_FULLFILLED},
+};
+
```
cdecker: If you decide to go with the bit trick below you can remove this struct array, since we'll just need 2 if-statements to cover all cases.

wallet/wallet.c
```diff
@@ -1368,6 +1392,16 @@ bool wallet_init_channels(struct wallet *w)
...
 
+static enum channel_state_db get_state_channel_db(const char *dir, const char *type)
+{
+	for (int i = 0; i < CHANNEL_STATE_DB_SIZE; i++) {
+		struct channel_state_param state = channel_state_params[i];
+		if (streq(state.dir_key, dir) && streq(state.type_key, type))
+			return state.state;
+	}
+	// Should never happen
+	fatal("Unknown stats key %s %s", dir, type);
```
cdecker: This also isn't required if we build up bitwise.

cdecker (commented):
Looking good, just some minor comments on how we could make this even cleaner :-)

wallet/wallet.c
vincenzopalazzo: Concept Ack.

`channel_stat_bucket` much better

wallet/wallet.c
vincenzopalazzo: Much much better! Thanks 😃

wallet/wallet.c
vincenzopalazzo: The solution with enum is much better I think! Thanks

wallet/wallet.c
vincenzopalazzo: Yep this is a great solution, thanks to suggesting it

vincenzopalazzo:
Rebase https://github.com/ElementsProject/lightning/commit/6dc954bb9156fd291bbdff27fc0b98d913717b07 and made changes as suggested.

Thanks, @cdecker to have shown me the trick with the enum.

wallet/wallet.c
```diff
@@ -1388,6 +1412,16 @@ bool wallet_init_channels(struct wallet *w)
...
```
niftynei:
```suggestion
	fatal("Unknown channel state key (direction %s, type %s)", dir, type);
```

wallet/wallet.c
```diff
@@ -1397,33 +1431,39 @@ void wallet_channel_stats_incr_x(struct wallet *w,
...
 			    "     , out_msatoshi_fulfilled = COALESCE(out_msatoshi_fulfilled, 0) + ?"
 			    " WHERE id = ?;");
-	} else {
-		fatal("Unknown stats key %s %s", dir, typ);
+		break;
 	}
 
+	// TODO(vincenzopalazzo): Better here or in the default case of switch stmt?
+	if (!query)
+		fatal("Unknown stats key %s %s", dir, type);
```
niftynei:
```suggestion
	fatal("Unknown channel state key (direction %s, type %s)", dir, type);
```

wallet/wallet.c
```diff
@@ -1397,33 +1431,39 @@ void wallet_channel_stats_incr_x(struct wallet *w,
...
 		query = SQL("UPDATE channels"
 			    "   SET out_payments_fulfilled = COALESCE(out_payments_fulfilled, 0) + 1"
 			    "     , out_msatoshi_fulfilled = COALESCE(out_msatoshi_fulfilled, 0) + ?"
 			    " WHERE id = ?;");
-	} else {
-		fatal("Unknown stats key %s %s", dir, typ);
+		break;
 	}
 
+	// TODO(vincenzopalazzo): Better here or in the default case of switch stmt?
```
niftynei: this is a perfect place to put this. in general we don't use `default` statements in switches.

wallet/wallet.c
```diff
@@ -47,14 +47,11 @@ struct channel_state_param {
 };
 
 /* possible channels states */
-struct channel_state_param channel_state_params[] = {
+static struct channel_state_param channel_state_params[] = {
```
niftynei: it'd be better to fold this change in with the commit that originally made the line.

wallet/wallet.c
```diff
@@ -1414,7 +1411,7 @@ bool wallet_init_channels(struct wallet *w)
 
 static enum channel_state_db get_state_channel_db(const char *dir, const char *typ)
 {
-	for (int i = 0; i < CHANNEL_STATE_DB_SIZE; i++) {
+	for (size_t i = 0; i < ARRAY_SIZE(channel_state_params); i++) {
```
niftynei: it'd be better to fold this change in with the commit that originally made the line.

wallet/wallet.c
```diff
@@ -1460,7 +1457,7 @@ void wallet_channel_stats_incr_x(struct wallet *w,
 		break;
 	}
 
-	// TODO(vincenzopalazzo): Better here or in the default case of switch stmt?
+	// Sanity check!
```
niftynei, line 1453: it'd be better to fold this change in with the commit that originally made the line.

wallet/wallet.c
```diff
@@ -1409,15 +1403,14 @@ bool wallet_init_channels(struct wallet *w)
 	return wallet_channels_load_active(w);
 }
 
-static enum channel_state_db get_state_channel_db(const char *dir, const char *typ)
+static enum channel_state_bucket get_state_channel_db(const char *dir, const char *typ)
```
niftynei, line 1406: nice! this should be folded in to the original commit at some point before we merge this into master, so that the original commit has the complete/perfect changeset ;)

niftynei (commented):
LGTM as is, my only nit would be to fold all the changes into a single, perfect commit.

vincenzopalazzo:
@niftynei Rebased and fold all the commits in only one, sorry for the missing steps.

I'm thinking, to put in the .gitignore the generated files like "wallet/db_postgres_sqlgen.c", but I'm not sure if I'm missing some important steps.

Let's wait for Github actions 🪑

rustyrussell (approved):
Ack ee987943784c97a6f7220e3683e0d9134600f41b

PR #4518, opened 2021-05-09
listpays: sort output payments

### LINKED ISSUE #4481: `listpays` is not sorted

```
lightning-cli listpays
```

Returns an unsorted list of payments.

### `getinfo` output

```
{
   "id": "02c16cca44562b590dd279c942200bdccfd4f990c3a69fad620c10ef2f8228eaff",
   "alias": "@lntxbot",
   "color": "296683",
   "num_peers": 37,
   "num_pending_channels": 2,
   "num_active_channels": 22,
   "num_inactive_channels": 2,
   "address": [
      {
         "type": "ipv4",
         "address": "5.2.67.89",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "v0.9.3-107-ga020602",
   "blockheight": 679798,
   "network": "bitcoin",
   "msatoshi_fees_collected": 32991436,
   "fees_collected_msat": "32991436msat",
   "lightning-dir": "/home/fiatjaf/.lightning1/bitcoin"
}
```

nalinbhardwaj:
Fixes #4481.

### DIFF
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -3176,7 +3176,8 @@ wallet_payment_list(const tal_t *ctx,
 						  ", partid"
 						  ", local_offer_id"
 						  " FROM payments"
-						  " WHERE payment_hash = ?;"));
+						  " WHERE payment_hash = ?"
+						  " ORDER BY id;"));
 		db_bind_sha256(stmt, 0, payment_hash);
 	} else {
 		stmt = db_prepare_v2(wallet->db, SQL("SELECT"
--- a/doc/lightning-listpays.7
+++ b/doc/lightning-listpays.7
@@ -59,6 +59,9 @@ If \fBstatus\fR is "failed":
 \fBerroronion\fR (hex, optional): the error onion returned on failure, if any\.
 
 .RE
+
+The returned array is ordered by increasing \fBcreated_at\fR fields\.
+
 .SH AUTHOR
 
 Rusty Russell \fI<rusty@rustcorp.com.au\fR> is mainly responsible\.
@@ -71,4 +74,4 @@ Rusty Russell \fI<rusty@rustcorp.com.au\fR> is mainly responsible\.
 
 Main web site: \fIhttps://github.com/ElementsProject/lightning\fR
 
-\" SHA256STAMP:e27d57394bef9bdaf9b99ae0d9050c9044c194ab66f6c94c43b532a86e1a0031
+\" SHA256STAMP:4037a4dd9746b5dfc91ef2f5b4fa9000a334689157d9ac86dc4dba5c82628cfe
--- a/doc/lightning-listpays.7.md
+++ b/doc/lightning-listpays.7.md
@@ -37,6 +37,8 @@ If **status** is "failed":
   - **erroronion** (hex, optional): the error onion returned on failure, if any.
 [comment]: # (GENERATE-FROM-SCHEMA-END)
 
+The returned array is ordered by increasing **created_at** fields.
+
 AUTHOR
 ------
 
--- a/doc/lightning-listsendpays.7
+++ b/doc/lightning-listsendpays.7
@@ -18,7 +18,7 @@ command per \fIpay\fR, so this command should be used with caution\.
 
 .SH RETURN VALUE
 
-On success, an array of objects is returned\. Each object contains:
+On success, an array of objects is returned, ordered by increasing \fIid\fR\. Each object contains:
 
 
  \fIid\fR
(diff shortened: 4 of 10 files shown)


wallet/wallet.c
```diff
@@ -3090,17 +3091,27 @@ void wallet_payment_set_failinfo(struct wallet *wallet,
...
+	struct wallet_payment **payments;
 	struct db_stmt *stmt;
 	struct wallet_payment *p;
 	size_t i;
 
-	payments = tal_arr(ctx, const struct wallet_payment *, 0);
+	payments = tal_arr(ctx, struct wallet_payment *, 0);
 	if (payment_hash) {
 		stmt =
 		    db_prepare_v2(wallet->db, SQL("SELECT"
```
vincenzopalazzo, line 3117: Any particular reason to use the asort algorithm instant of the ORDER BY SQL keyword?

vincenzopalazzo (commented):
I have only a small curiosity and I would like to know more about that.

I'm not able to catch what Is the reason to collect the result from the database and only after sort it, instant to ask the database the list of payment sorted.

The only motivation that I have in mind is that the databases used from c-lightning use a different dialect, but I'm not able to prove that because the `ORDER BY` should be a basic construct of the SQL language

rustyrussell:
OK, this is a great start!  Your code is really nice.

Some feedback:

1. This actually orders the low-level command `listsendpays`. Which is great, but another commit is needed for the `listpays` command to order *that* (which uses listsendpays).
2. This commit should have a line like ```Changelog-Changed: JSON: `listsendpays` output is now ordered by `id`.```
3. As @vincenzopalazzo points out, simply adding "ORDER BY id" is a much simpler change.
4. Create a new commit, which can use your asort code to the plugin/pay.c, except order it by `created_at` maybe? And you'll need to marshal it into a temporary array to sort it.

Thanks!
Rusty.

nalinbhardwaj:
Thanks for the note @vincenzopalazzo @rustyrussell. I realised that after making this PR (which is why I ended up marking it as a draft PR). I'll make both the noted changes soon.

nalinbhardwaj:
@rustyrussell @vincenzopalazzo ready for review.

wallet/wallet.c
```diff
@@ -3143,7 +3143,8 @@ wallet_payment_list(const tal_t *ctx,
 						  ", partid"
 						  ", local_offer_id"
 						  " FROM payments"
-						  " WHERE payment_hash = ?;"));
+						  " WHERE payment_hash = ?"
+						  " ORDER BY id;"));
```
vincenzopalazzo, line 3147: Do you think that sort by timestamp it is a good idea here?

vincenzopalazzo (changes requested):
utack https://github.com/ElementsProject/lightning/pull/4518/commits/1459ab0457eb5cd25cfbb48c20ea2aa47a7fe9e6

With two small nit to me, but @rustyrussell can undo my comments if they are wrong.

- If you sort (ORDER by timestamp ASC | DESC) by timestamp you can avoid the sorting operation, this moves the work from lightning code to database code, and the database can (or should/may) have better performance in sorting things.
- Use the query to sort the element you can avoid using the additional array, but it is in contrast with point 4 of Rusty comment, so in this case, I can missing some things.

PS: I would like to have an additional test in test_pay that makes sure that the array is really sorted, to avoid crazy debbugging in some case in the future.

wallet/wallet.c
rustyrussell: id is the order it was placed in the db, which is probably right here.

rustyrussell:
> utack [1459ab0](https://github.com/ElementsProject/lightning/commit/1459ab0457eb5cd25cfbb48c20ea2aa47a7fe9e6)
> 
> With two small nit to me, but @rustyrussell can undo my comments if they are wrong.
> 
>     * If you sort (ORDER by timestamp ASC | DESC) by timestamp you can avoid the sorting operation, this moves the work from lightning code to database code, and the database can (or should/may) have better performance in sorting things. In addition, this change simplifies the code in the pay plugin.

Unfortunately, `listpays` gathers `listsendpays` into a hashtable, losing order.  So explicit sorting is still required here.

> PS: I would like to have an additional test in test_pay that makes sure that the array is really sorted, to avoid crazy debugging in some case in the future.

More tests would always be nice, of course!  I will have a go at writing one, since this needs a trivial rebase anyway...

rustyrussell:
@nalinbhardwaj Sorry for the delay.  Did a trivial rebase to fix the merge conflict with db files, and added tests and docs.

What do you think?

nalinbhardwaj:
@rustyrussell Looks great! Thanks for the rebase and the doc+tests updates. Let me know if there's anything else I missed before.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/4518/commits/0d622272ceeea4218111093218c4881da54e1f92

PR #4523, opened 2021-05-10
Improve documentation for calls that use psbt

dscotese:
Added FAQ on getting a PSBT.  Also when I read the RETURN VALUE for utxopsbt the first time, I didn't realize this command produces (or recreates) the psbt that I need.  The change I made makes that more obvious when it's presented as a hit by GH for a search for psbt.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -164,3 +164,9 @@ Here is an example in Python checking if [one of the `option_static_remotekey` b
 >>> bool(0x02aaa2 & ((1 << 12) | (1 << 13)))
 True
 ```
+
+## Technical Questions
+
+### How do I get the `psbt` for RPC calls that need it?
+
+A `psbt` is created and returned by a call to [`utxopsbt` with `reservedok=true`](https://lightning.readthedocs.io/lightning-utxopsbt.7.html?highlight=psbt).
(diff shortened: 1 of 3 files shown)


doc/FAQ.md
```diff
@@ -164,3 +164,9 @@ Here is an example in Python checking if [one of the `option_static_remotekey` b
 >>> bool(0x02aaa2 & ((1 << 12) | (1 << 13)))
 True
 ```
+
+## Technical Questions
```
vincenzopalazzo, line 168: Ops! This change make angry the GitHub action https://github.com/ElementsProject/lightning/pull/4523/checks?check_run_id=2548263926

Maybe a good place where put the FAQ it is the following one https://github.com/ElementsProject/lightning/blob/master/doc/FAQ.md ?

This make consistent all the documentation pages

cdecker: Yes, that just means that `make doc-all` needs to be called and the generated files need to be checked in. We used to not require the generated files to also be checked in, but it would result in stale docs being published.

dscotese:
> Concept ACK. Please remove the merge commit, [b3569aa](https://github.com/ElementsProject/lightning/commit/b3569aa1218f36ebea51dd5d0cd24d201d0f6a29), from your PR.

I have looked around how to do this, but since I went through Github to create both the merge commit and the PR, I don't know how to do it and haven't found a way.

rustyrussell:
I can do it, one second (rebase -i onto master lets you only choose the commits you want).

rustyrussell (approved):
Ack e7da67c6cea1dde9d457dc2e2bda2cbb7201177e

PR #4526, opened 2021-05-11
A fuzz target for `libhsmd` message handling and some bugfixes

darosior:
The work on libhsmd in #4497 made it simpler to fuzz the message handling of `hsmd`. This adds a simple target to test that `hsmd` is not going to crash even if sent (smartly malleated) garbage. This uncovered a few missing NULL checks in our fromwire routines.

I'm opening this as draft, and will open another PR for the bugfixes as i spent the last days tracking down a segfault in `wally_tx_from_bytes` and i won't be able to continue.
I was also planning to generate some seeds by (ab)using the functional tests but i couldn't even get to there yet.

PR #4527, opened 2021-05-11
Add some missing parsing checks

darosior:
This is part of the finds of #4526 PR'ed separately as i don't have a fix for all of them yet.

Changelog-None


rustyrussell:
The first commit seems wrong?

Only some functions return success: some return void. Generally it's fine (by design!) to ignore fromwire() failures for a while as long as the top-level caller checks that cur is not NULL.  Presumably there are some places that check is missing?

darosior:
Hmm, the first commit only adds this check to the `fromwire`s that return an allocated ([off the passed `ctx`](https://github.com/ElementsProject/lightning/pull/4527/commits/a6ba4d7a1a319a3942d255ed3cb7facd0adaf62f#diff-96140fbbf3b28243cf5822b043d9e46a703f1abd9cef6a5b292bed55c837e6e7R182-R184)) value. I just manually checked and it doesn't seem wrong (every place the check is implemented, if the result is NULL there was an error).

However it's probably misguided, i figured we should not continue to read if there was an error but i can replace this with adding the missing caller checks!

rustyrussell:
> Hmm, the first commit only adds this check to the `fromwire`s that return an allocated ([off the passed `ctx`](https://github.com/ElementsProject/lightning/pull/4527/commits/a6ba4d7a1a319a3942d255ed3cb7facd0adaf62f#diff-96140fbbf3b28243cf5822b043d9e46a703f1abd9cef6a5b292bed55c837e6e7R182-R184)) value. I just manually checked and it doesn't seem wrong (every place the check is implemented, if the result is NULL there was an error).
> 

They mostly return bool though.  Because C, returning NULL "works", but should have no effect since each of these functions checks *cursor != NULL at the end.

> However it's probably misguided, i figured we should not continue to read if there was an error but i can replace this with adding the missing caller checks!

It's specifically designed that way, to simplify error checking, so ideally this commit should have no effect...

darosior:
> They mostly return bool though. Because C, returning NULL "works", but should have no effect since each of these functions checks *cursor != NULL at the end.

Sorry, i just went through all of the places where the check is added in https://github.com/ElementsProject/lightning/pull/4527/commits/a6ba4d7a1a319a3942d255ed3cb7facd0adaf62f (again!), and none of the function it checks the result of returns `bool`. Anyways, will change to add check that cursor != NULL instead as soon as i can unstuck #4526 !

> ideally this commit should have no effect...

I'm pretty sure it had effect, but i can't reproduce anymore using the fuzz target as i'm stuck by https://github.com/ElementsProject/lightning/pull/4526#issuecomment-842256900 . Will modify to use a check further down the call stack as soon as i can test that it actually fixes the previously encountered issue..

rustyrussell:
The first commit seems wrong, the second is redundant (new_psbt cannot fail), and the third is already merged.

PR #4529, opened 2021-05-12
Dual-funding, last of the CI tests

niftynei:
In 0f79b1bd805ba6b1d2aaf0cd0737f8aef830ffd9, we mark a bunch of tests as 'v1' only. Here we translate as many of those as possible/logical to have 'v2' counterparts.

Changelog-None


rustyrussell (approved):
Ack e4309c5e2d2433df3147b06d07c09754be03b60a

PR #4535, opened 2021-05-16
remove checking for unused "direction" property in `sendpay`.

fiatjaf:
The need for this `direction` property has always puzzled and confused me, as it should be implied from the route direction and the channel. Today I learned it is being required but not being used at all, so I removed it.

This seems to work fine, but maybe I'm making a huge mistake somewhere.

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1187,13 +1187,12 @@ static struct command_result *
 param_route_hop(struct command *cmd, const char *name, const char *buffer,
 		const jsmntok_t *tok, struct route_hop **hop)
 {
-	const jsmntok_t *idtok, *channeltok, *directiontok, *amounttok, *delaytok;
+	const jsmntok_t *idtok, *channeltok, *amounttok, *delaytok;
 	struct route_hop *res;
 
 	res = tal(cmd, struct route_hop);
 	idtok = json_get_member(buffer, tok, "id");
 	channeltok = json_get_member(buffer, tok, "channel");
-	directiontok = json_get_member(buffer, tok, "direction");
 	amounttok = json_get_member(buffer, tok, "amount_msat");
 	delaytok = json_get_member(buffer, tok, "delay");
 
@@ -1202,11 +1201,6 @@ param_route_hop(struct command *cmd, const char *name, const char *buffer,
 		return command_fail(
 		    cmd, JSONRPC2_INVALID_PARAMS,
 		    "Either 'id' or 'channel' is required for a route_hop");
-
-	if (channeltok && !directiontok)
-		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
-				    "When specifying a channel you must also "
-				    "specify the direction");
 
 	if (!amounttok)
 		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
utACK https://github.com/ElementsProject/lightning/pull/4535/commits/898d459ff4786e3985e4919aa407c8d2f668626f

Test failure seems unrelated

rustyrussell:
Yes, it's only for symmetry with getroute, which gives the direction as a boolean.  Thus it should be allowed (sure, why not) but not required.

rustyrussell:
Suggested fix is a partial revert:

```diff
diff --git a/lightningd/pay.c b/lightningd/pay.c
index 2f79539ae..5ac07522b 100644
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1333,7 +1333,7 @@ static struct command_result *param_route_hops(struct command *cmd,
 		struct amount_msat *msat, *amount_msat;
 		struct node_id *id;
 		struct short_channel_id *channel;
-		unsigned *delay;
+		unsigned *delay, *direction;
 		struct pubkey *blinding;
 		u8 *enctlv;
 		enum route_hop_style *style, default_style;
@@ -1346,6 +1346,8 @@ static struct command_result *param_route_hops(struct command *cmd,
 			   p_opt("id", param_node_id, &id),
 			   p_opt("delay", param_number, &delay),
 			   p_opt("channel", param_short_channel_id, &channel),
+			   /* Allowed (getroute supplies it) but ignored */
+			   p_opt("direction", param_number, &direction),
 			   p_opt("style", param_route_hop_style, &style),
 			   p_opt("blinding", param_pubkey, &blinding),
 			   p_opt("enctlv", param_bin_from_hex, &enctlv),
```

cdecker:
To illistrate the issue, I fished this error out of the CI tests:

```
E           pyln.client.lightning.RpcError: RPC call failed: method: sendpay, payload: {'route': [{'id': '035d2b1192dfba134e10e540875d366ebc8bc353d5aa766b80c090b39c3a5d885d', 'channel': '103x3x1', 'direction': 1, 'msatoshi': 50001002, 'amount_msat': 50001002msat, 'delay': 21, 'style': 'tlv'}, {'id': '022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59', 'channel': '103x2x1', 'direction': 1, 'msatoshi': 50000501, 'amount_msat': 50000501msat, 'delay': 15, 'style': 'tlv'}, {'id': '0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518', 'channel': '103x1x0', 'direction': 0, 'msatoshi': 50000000, 'amount_msat': 50000000msat, 'delay': 9, 'style': 'tlv'}], 'payment_hash': '14d652a3a20a13c2eb7cedd088fc94180ba46cb3089c2378c34a3b48ccef3255'}, error: {'code': -32602, 'message': 'unknown parameter: direction, this may be caused by a failure to autodetect key=value-style parameters. Please try using the -k flag and explicit key=value pairs of parameters.'}
```

As expected the direction being specified causes a failure here. It'll not be sufficient to adjust the pyln code by the way, since there are plugins that are using older versions of pyln or other plugin libs. Rusty's suggestion should work :+1:

cdecker (dismissed):
Make `direction` optional instead.

niftynei:
FYI @fiatjaf we're looking to cut a release sometime next week, and I'd love to get this change in before we do

fiatjaf:
Applied Rusty's patch.

cdecker:
ACK ddae6ec31d9cafc8fa7937d53558a9c92a4ea464

PR #4537, opened 2021-05-16
stop requiring channel and direction on `sendonion` since they're not used.

fiatjaf:
Does this make sense?

Untested, wanted to make sure it made sense first.

### DIFF
--- a/plugins/libplugin-pay.c
+++ b/plugins/libplugin-pay.c
@@ -1505,8 +1505,6 @@ static struct command_result *payment_createonion_success(struct command *cmd,
 	json_add_hex_talarr(req->js, "onion", p->createonion_response->onion);
 
 	json_object_start(req->js, "first_hop");
-	json_add_short_channel_id(req->js, "channel", &first->scid);
-	json_add_num(req->js, "direction", first->direction);
 	json_add_amount_msat_only(req->js, "amount_msat", first->amount);
 	json_add_num(req->js, "delay", first->delay);
 	json_add_node_id(req->js, "id", &first->node_id);
--- a/doc/lightning-sendonion.7
+++ b/doc/lightning-sendonion.7
@@ -39,7 +39,6 @@ to add an HTLC for 1002 millisatoshis and a delay of 21 blocks on top of the cur
 .RS
 {
   "id": "022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59",
-  "direction": 1,
   "amount_msat": "1002msat",
   "delay": 21,
 }
@@ -121,4 +120,4 @@ Christian Decker \fI<decker.christian@gmail.com\fR> is mainly responsible\.
 
 Main web site: \fIhttps://github.com/ElementsProject/lightning\fR
 
-\" SHA256STAMP:19d7317d0c50409db6a09b1e4411fc9697976254be3555cbf3bcf2a28984529c
+\" SHA256STAMP:1d7ceae27ac4ac51597b92de50cf7bc917c59bdf59f3280efc06491ed4efaffc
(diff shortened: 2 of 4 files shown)


plugins/libplugin-pay.c
```diff
@@ -1424,8 +1424,6 @@ static struct command_result *payment_createonion_success(struct command *cmd,
 	json_add_hex_talarr(req->js, "onion", p->createonion_response->onion);
 
 	json_object_start(req->js, "first_hop");
-	json_add_short_channel_id(req->js, "channel", &first->channel_id);
-	json_add_num(req->js, "direction", first->direction);
```
rustyrussell, line 1428: This needs to be made dependent on `if (!deprecated_apis)` with a Changelog-Deprecated: JSON-RPC: `createonion` `first_hop` `channel` and `direction` fields removed.

Then we can actually rip it out in 6 months.

fiatjaf: But they are being ignored!

Or do you mean we can't remove because then the calls will fail for everybody that is currently passing these value?

cdecker: That's exactly what the `deprecated_apis` flag is for. If enabled it must replicate the old behavior, if disabled it should report the error.

fiatjaf: Sorry, but let me insist on this point I'm not understanding: this PR doesn't change the current behavior at all. It just stops passing two parameters (channel and direction) that were not being used at all. The behavior should be the exact same as before.

In this case, it doesn't make sense to me to guard the changes with a deprecation notice. What am I getting wrong about this?

cdecker: Let me point you towards the sea of red that the CI tests currently are. This PR breaks a lot of tests (if done in isolation). Please get the tests working again. The failures will point you towards the issue described by Rusty, since `sendonion` still requires the fields you removed.

cdecker (changes requested):
Changing just the requesting side breaks the receiving side. Please make the fields optional on the receiving side before removing them here.

plugins/libplugin-pay.c
rustyrussell: Huh, I misread.  This patch is fine: it just takes them out of the internal call that pay uses.

Of course, CI says that breaks everything, so we should fix that! 

```patch
diff --git a/lightningd/pay.c b/lightningd/pay.c
index d96686beb..7c34b4d87 100644
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1026,7 +1026,7 @@ send_payment_core(struct lightningd *ld,
 					   "peer found");
 
 		json_add_routefail_info(data, 0, WIRE_UNKNOWN_NEXT_PEER,
-					&ld->id, channel->scid,
+					&ld->id, NULL,
 					node_id_idx(&ld->id, &first_hop->nodeid),
 					NULL);
 		json_object_end(data);
```

rustyrussell:
> Changing just the requesting side breaks the receiving side. Please make the fields optional on the receiving side before removing them here.

This is wrong: patch doesn't quite do what the title says. 

It actually does two things:
1. Fix sendonion to handle missing 'channel' and 'direction' properly.
2. Remove "channel" and "direction" from the call used by libplugin-pay.

We both saw the second one and thought JSON output fields were being removed.  Too hasty!

fiatjaf:
I think the sea of red tests was due to a variable name typo I had. Now locally there is only one test failing which seems unrelated or I couldn't understand it. Pushing temporarily anyway while I don't manage to find that out.

plugins/libplugin-pay.c
fiatjaf: Why this change? Shouldn't we pass the actual channel we're using here (`channel->scid`)?

rustyrussell: channel is NULL here, it's inside an "if (!channel)".

fiatjaf: Oops, thank you.

cdecker (approved):

cdecker:
The problematic changes were in the other PR, sorry for holding up the progress here.

rustyrussell:
I squashed this into one commit, and tacked on a Changelog-Changed line.

Thanks!

PR #4543, opened 2021-05-18
libhsmd(python): Use proper tmpctx management and fix embarassing oversights

cdecker:
We were:

 - Not setting up and clearing `tmpctx` correctly.
 - Not using `common_setup`
 - Using the message, instead of the `peer_id` to derive secrets :facepalm: 

Changelog-None


contrib/libhsmd_python/setup.py
```diff
@@ -40,13 +40,18 @@ def build_make(self, ext):
                     "clone",
                     "--recursive",
                     '--branch=libhsmd-python',
-                    "https://github.com/cdecker/lightning.git",
+                    "https://github.com/cdecker/lightning.git"
```
niftynei, line 43: comma missing here seems like a mistake?

rustyrussell: Good catch!

rustyrussell (changes requested):

rustyrussell (approved):

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/f796e58108f49023874e6d607be2926cb97ff42d

Issue #4539, opened 2021-05-16
Ubuntu 20.04 server - lightningd: error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory


urza:
I downloaded the latest binary release (0.10.0) on fresh Ubuntu server 20.04.2

After copying files, running lightningd command then resulted in this error:
`lightningd: error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory`

This can be resolved by installing libpq-dev:

`sudo apt-get install libpq-dev`

I wanted to document it here in hope that it might help if someone else encounters this, feel free to close this issue.

Issue #4538, opened 2021-05-16
Improve experience with installation on new Ubuntu 20.04 - ppa:bitcoin/bitcoin no longer supported


urza:
I just installed c-lightning (for the first time, no prior experiences) on fresh Ubuntu Server 20.04.2

Installing from Ubuntu PPA by following the instructions from [readme](https://github.com/ElementsProject/lightning#installation):
```
sudo apt-get install -y software-properties-common
sudo add-apt-repository -u ppa:bitcoin/bitcoin
sudo add-apt-repository -u ppa:lightningnetwork/ppa
sudo apt-get install bitcoind lightningd
```
--> Installation of bitcoin core is no longer not supported this way. The command results in:

![image](https://user-images.githubusercontent.com/189804/118406964-abd61280-b67e-11eb-805a-9bd169a718ad.png)

and

![image](https://user-images.githubusercontent.com/189804/118407008-e63faf80-b67e-11eb-9485-9e387c89a7c3.png)

It seems that official supported way how to install bitcoin core from packages on ubuntu is by snap packages:

https://github.com/bitcoin-core/packaging
by command `sudo snap install bitcoin-core`

I recommend either removing the PPA installation from the main readme.md and leaving it up to user to install bitcoind in some other way, or changing the apt install method for snap install.

PR #4544, opened 2021-05-18
readme: fix ubuntu installation instructions

urza:
I have verified that this results in working bitcoin core and c-lightning on fresh Ubuntu 20.04 server installation.

This is also consistent with instructions in doc/INSTALL.md

Fixes: #4539
Fixes:  #4538
Reported-By: @urza
Changelog-None: just readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,9 +48,10 @@ For the impatient here's the gist of it for Ubuntu:
 
 ```bash
 sudo apt-get install -y software-properties-common
-sudo add-apt-repository -u ppa:bitcoin/bitcoin
 sudo add-apt-repository -u ppa:lightningnetwork/ppa
-sudo apt-get install bitcoind lightningd
+sudo apt-get install lightningd snapd
+sudo snap install bitcoin-core
+sudo ln -s /snap/bitcoin-core/current/bin/bitcoin{d,-cli} /usr/local/bin/
 ```
 
 ### Starting `lightningd`


jsarenik:
ACK 64c0c34

README.md
```diff
@@ -48,9 +48,10 @@ For the impatient here's the gist of it for Ubuntu:
 
 ```bash
 sudo apt-get install -y software-properties-common
-sudo add-apt-repository -u ppa:bitcoin/bitcoin
 sudo add-apt-repository -u ppa:lightningnetwork/ppa
-sudo apt-get install bitcoind lightningd
+sudo apt-get install lightningd snapd
+sudo snap install bitcoin-core
```
vincenzopalazzo, line 53: I would like to have two sections to handle this information here. Not all people are a fun of snap (like me).

Maybe maintain two versions of this command is better?

vincenzopalazzo (changes requested):

README.md
vincenzopalazzo: Maybe it is good to add also this information about the postgress dependencies `sudo apt-get install libpq-dev` in the description https://github.com/ElementsProject/lightning/issues/4539#issue-892725143 

Usually, I need to install in all my machine before being able to use c-lightning

vincenzopalazzo: In addition, I read your issue on the app, and I think we are a dependencies fork like before ubuntu-20, after ubuntu 20.

In fact, on my docker sandbox, I fixed with this solution (without snap)

```
add-apt-repository ppa:luke-jr/bitcoincore
apt-get update  && apt-get install -y bitcoind
add-apt-repository -u ppa:lightningnetwork/ppa
apt-get update  &&; apt-get install -y lightningd
```

Maybe we need to make a different section for Ubuntu 20?

urza: I also needed to install libpq-dev, but only when I installed c-lightning from binary release. With installation of lightningd from apt-get installed it automatically.

urza: Ad snap - I just followed official bitcoin core webpage where it links to the snap package as official way how to install if from package on ubuntu https://bitcoincore.org/en/download/

It the luke's ppa works, that is also fine I guess.

vincenzopalazzo: >I also needed to install libpq-dev, but only when I installed c-lightning from binary release. With installation of lightningd from apt-get installed it automatically.

Yes, sorry, I mean we can add this library in the Library requirement in the same page https://github.com/ElementsProject/lightning/blob/master/doc/INSTALL.md#library-requirements

>Ad snap - I just followed official bitcoin core webpage where it links to the snap package as official way how to install if from package on ubuntu

Yes, the only thing that I don't agree is to remove the bitcoin ppa, in my vision is better to have the choice. Any way it is fine for me

jsarenik: @vincenzopalazzo There are multiple ways to run `bitcoind` on Ubuntu. I use the [pre-compiled binaries](https://bitcoincore.org/bin/bitcoin-core-0.21.1/bitcoin-0.21.1-x86_64-linux-gnu.tar.gz) for example. Those can be easily extracted and run as unprivileged user without root access. To install a PPA or snap, root access is needed. The question is how far the rabbit-hole you want our first-time users to go. Personally I would totally love to see everyone compiling from scratch in a musl-libc chroot (on any system, also Ubuntu) but that may be an over-kill for many.

Generally I think it would be much better if exact steps like these would be completely left out of the documentation. It is distribution-specific and not related to C-Lightning. For example, I tried to search [Linux Kernel documentation for a mention of apt](https://www.kernel.org/doc/html/latest/search.html?q=apt&check_keywords=yes&area=default) and did not find anything (which is correct IMHO).

C-Lightning is not Ubuntu-specific despite it is now widely accepted as reproducible-build system. The real reproducible build in my opinion is `-no-pie` statically musl-libc-linked executable which can actually be run on almost _any_ Linux running on the chosen architecture, with no further dependencies.

Excuse me for rant :-)

urza:
> Good work updating the docs. Could you remove the `.vs` directory from the commit?

Ups sorry about that, it's embarrassing :) Fixed.

PR #4549, opened 2021-05-20
Onion v2 deprecation

darosior:
This prevents user from setting or connecting to v2 hidden services. This is behind `deprecated_apis` but since they are going to break in a month and a half on the Tor backend we should probably break before?

I still need to update the doc and i wonder if we shoudl do anything about the `node_announcement`.


kristapsk (approved):
utACK 293a7f84b56ccbd39f8dae6e146647412f896c0d

rustyrussell:
Second commit simply needs:

```
Changelog-Deprecated: lightningd: `enable-autotor-v2-mode` option and v2 Tor addresses.  Use v3.  See https://blog.torproject.org/v2-deprecation-timeline.
```

darosior:
Done, and also updated the doc. `doc/TOR.md` is not really clear so it's really best effort here :/

rustyrussell (approved):
Ack d9434eaa0d0fc75d01d8c11f49ad05e53af847ad

rustyrussell:
YA trivial rebase for conflicts on generated files

PR #4552, opened 2021-05-22
pytest: disable mpp test which is blocking merges.

rustyrussell:
I'm still diagnosing exactly why this fails, but it will need Christian's
help at this stage.

Changelog-None


cdecker:
ACK 3ca2e4921bee21c720786154243e8f6e4f0e0667

rustyrussell:
Sigh.  It's unittest.skipIf not pytest.skipIf...

PR #4554, opened 2021-05-23
added Litestream Replication

Alex-CodeLab:
Litestream.io enables easy SQLite replication for continuous backups


Alex-CodeLab:
> Litestream requires periodic but short write locks on the database in order to sync changes. SQLite will return an error by default if your application tries to obtain a write lock at the same time. To prevent this, your application will need to set the busy_timeout pragma when it initializes the database connection.


How does c-lightning handle write-locks now?

rustyrussell:
I think this will break c-lightning: we do not expect db operations to fail at all. We would definitely want to block in this case.

Alex-CodeLab:
> I think this will break c-lightning: we do not expect db operations to fail at all. We would definitely want to block in this case.

I noticed the same issue was discussed here already https://github.com/ElementsProject/lightning/pull/4207#discussion_r528206037
Setting `PRAGMA busy_timeout` is a reasonable solution, but still lock errors could be handled correctly. 
There are other scenarios where this could also happen, for example when someone queries lightningd.sqlite3 using some sqlite-browser or other application on a running system.

rustyrussell:
OK, I (trivially) rebased on master, and added commits so that we set the busy timeout.  I also remove the copy operation in our tests, which should actually exercise this a little.

rustyrussell:
Trivial rebase.

PR #4557, opened 2021-05-25
pytest: temporarily disable hanging test.

rustyrussell:
We're trying to reproduce it now, but it's blocking other PRs.

Changelog-None


rustyrussell:
Closed in favor of disabling all DF tests.

PR #4559, opened 2021-05-26
Negotiate close even if we already consider the channel closed.

rustyrussell:
This was initially found by CI: one peer got the final signature and the other didn't, so one was closing and the other wasn't.  This is generally harmless (the one who thinks it's closed broadcasts, so the peer will see it eventually), but it's painful if they lost information and were relying on the reconnection to provide it.

This adds a test for that, but also allows us to renegotiate closing even if we consider the channel completely closed (in which case, openingd will receive the REESTABLISH msg).  I didn't support this for dualopend yet, since there the potential channel and the connection are quite closely tied, and it's hard to dispose of the (uncreated) channel without closing the connection.

It's a nice-to-have, but not vital.


cdecker:
Rebased on top of `master` to fix conflict on generated SQL files.

rustyrussell:
This opened a longstanding issue with "I consider your channel closed so I will no longer talk to you about it!".  I will fix that on top of this too, since it's been a flaw for a while.

rustyrussell:
OK, so some surprising tests expected the old behavior.  Hmm, will look next week.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4559/commits/015ab604c1df0b4464b7c5d7b5262e25bf037339

PR #4560, opened 2021-05-26
option_will_fund: part I of ?, set liquidity_ads + broadcast via nodeann

niftynei:
Imports the option_will_fund spec; allows you to set/update/broadcast the rate card out through your node_announcement.

Note that the 'canonical' data for the rate-card lives in `funder`, and is polled by lightningd every ~5min.
The options + command implemented by `funder` form the interface required to be implemented by any plugin that handles channel lease requests.

We do this so that there's no API introduced onto lightningd that might be called by accident, i.e. you cannot set the lease terms/rate card unless there is a plugin that supports them.

Unfortunately this means we poll + do some weird parsing of plugin options at start so that gossipd gets the correct initial rate card (as we can't query a plugin for them until after gossipd starts up).


wire/extracted_peer_liquid_ads.patch
```diff
@@ -0,0 +1,48 @@
...
+@@ -249,6 +253,12 @@
+ msgdata,node_announcement,alias,byte,32
+ msgdata,node_announcement,addrlen,u16,
+ msgdata,node_announcement,addresses,byte,addrlen
++msgdata,node_announcement,tlvs,node_ann_tlvs,
++tlvtype,node_ann_tlvs,option_will_fund,1
++tlvdata,node_ann_tlvs,option_will_fund,funding_fee_proportional_basis,u16,
++tlvdata,node_ann_tlvs,option_will_fund,funding_fee_base_sat,u32,
++tlvdata,node_ann_tlvs,option_will_fund,channel_fee_proportional_basis,u16,
++tlvdata,node_ann_tlvs,option_will_fund,channel_fee_base_msat,tu32,
```
rustyrussell, line 45: This is missing "funding_fee_weight_charge,u32,"?

wire/test/run-peer-wire.c
```diff
@@ -928,6 +932,37 @@ static bool update_add_htlc_eq(const struct msg_update_add_htlc *a,
...
+		bo = b->tlvs->option_will_fund;
+
+		if (ao->funding_fee_proportional_basis != bo->funding_fee_proportional_basis)
+			return false;
+		if (ao->funding_fee_base_sat != bo->funding_fee_base_sat)
+			return false;
+		if (ao->channel_fee_proportional_basis != bo->channel_fee_proportional_basis)
+			return false;
+		if (ao->channel_fee_base_msat != bo->channel_fee_base_msat)
+			return false;
```
rustyrussell, line 962: Suggest implememt tlv_node_ann_tlvs_option_will_fund_eq().

common/liquidity_ad.c
```diff
@@ -0,0 +1,26 @@
+#include "liquidity_ad.h"
+#include <wire/wire.h>
+
+void towire_liquidity_ad(u8 **pptr,
+			 const struct liquidity_ad *ad)
+{
+	towire_u16(pptr, ad->lease_basis);
+	towire_u16(pptr, ad->channel_fee_basis);
+	towire_amount_sat(pptr, ad->lease_base_sat);
+	towire_amount_msat(pptr, ad->channel_base_msat);
```
rustyrussell, line 10: You need the weight here too...

gossipd/gossip_generation.c
```diff
@@ -25,7 +26,8 @@
  * between the dummy creation and the call with a signature. */
 static u8 *create_node_announcement(const tal_t *ctx, struct daemon *daemon,
 				    const secp256k1_ecdsa_signature *sig,
-				    u32 timestamp)
+				    u32 timestamp,
+				    struct liquidity_ad *ad)
```
rustyrussell, line 32: const

gossipd/gossipd.c
```diff
@@ -198,6 +199,21 @@ static bool get_node_announcement(const tal_t *ctx,
...
+	if (na_tlv->option_will_fund) {
+		*ad = tal(ctx, struct liquidity_ad);
+		(*ad)->lease_basis =
+			na_tlv->option_will_fund->funding_fee_proportional_basis;
+		(*ad)->lease_base_sat =
+			amount_sat(na_tlv->option_will_fund->funding_fee_base_sat);
+		(*ad)->channel_fee_basis =
+			na_tlv->option_will_fund->channel_fee_proportional_basis;
+		(*ad)->channel_base_msat =
+			amount_msat(na_tlv->option_will_fund->channel_fee_base_msat);
```
rustyrussell, line 213: This is getting messy.  Can we add a subtype to the spec and use that everywhere?

gossipd/gossipd_wire.csv
```diff
@@ -15,6 +15,11 @@ msgdata,gossipd_init,announcable,wireaddr,num_announcable
 msgdata,gossipd_init,dev_gossip_time,?u32,
 msgdata,gossipd_init,dev_fast_gossip,bool,
 msgdata,gossipd_init,dev_fast_gossip_prune,bool,
+msgdata,gossipd_init,has_offer,bool,
+msgdata,gossipd_init,lease_proportional_basis,u16,
+msgdata,gossipd_init,lease_base_sat,amount_sat,
+msgdata,gossipd_init,channel_proportional_basis,u16,
+msgdata,gossipd_init,channel_base_msat,amount_msat,
```
rustyrussell, line 22: Yech.  The alternate approach is to implement setliquidyoffer RPC, and plumb through to gossipd.

Then change gossipd:
1. change update_own_node_announcement() to take a "bool startup" flag.
2. change nannounce_different to take a `bool *only_missing_liquidityoffer`, which it sets if it returns true but the only difference is that the proposed nannounce is missing a liquidity offer.
3. If "startup" and the only difference is that we don't have a liquidity offer yet, set a timer to call update_own_node_announcement(startup=false) in 60 seconds (in case they restart without the liquidity plugin, we'll then send out a new node ann).
3. Call update_own_node_announcement(startup=false) when receiving a new liquidity ann.

This seems much cleaner (ok, except for the startup delay).

lightningd/gossip_control.c
```diff
@@ -195,12 +197,105 @@ static void gossip_topology_synced(struct chain_topology *topo, void *unused)
...
+	/* This is how the manpage says to do it.  Yech. */
+	errno = 0;
+	l = strtoul(arg, &endp, 0);
+	if (*endp || !arg[0])
+		return tal_fmt(NULL, "'%s' is not a number", arg);
+	*u = l;
+	if (errno || *u != l)
+		return tal_fmt(NULL, "'%s' is out of range", arg);
+	return NULL;
+}
```
rustyrussell, line 343: Move with the other opt_set_?

lightningd/gossip_control.c
```diff
@@ -195,12 +197,105 @@ static void gossip_topology_synced(struct chain_topology *topo, void *unused)
...
+	const char *err;
+
+	opt1 = plugin_opt_find(p, OPT_LEASE_FEE_BASE,
+			      strlen(OPT_LEASE_FEE_BASE));
+	opt2 = plugin_opt_find(p, OPT_LEASE_CHAN_FEE_BASE,
+			      strlen(OPT_LEASE_CHAN_FEE_BASE));
+	opt3 = plugin_opt_find(p, OPT_LEASE_FEE_BASIS,
+			      strlen(OPT_LEASE_FEE_BASIS));
+	opt4 = plugin_opt_find(p, OPT_LEASE_CHAN_FEE_BASIS,
+			      strlen(OPT_LEASE_CHAN_FEE_BASIS));
```
rustyrussell, line 361: Ugh, no please no...

lightningd/gossip_control.c
```diff
@@ -512,6 +512,23 @@ static void json_getnodes_reply(struct subd *gossip UNUSED, const u8 *reply,
...
+			json_add_amount_sat_only(response,
+						 "lease_fee_base_msat",
+						 nodes[i]->ad->lease_base_sat);
+			json_add_num(response,
+				     "channel_fee_basis_max",
+				     nodes[i]->ad->channel_fee_basis);
+			json_add_amount_msat_only(response,
+						  "channel_fee_base_max_msat",
+						  nodes[i]->ad->channel_base_msat);
+
```
rustyrussell, line 529: Needs weight.

plugins/funder.c
```diff
@@ -679,6 +680,24 @@ static void json_channel_open_failed(struct command *cmd,
 		unreserve_psbt(open);
 }
 
+static void rate_card_to_json(struct json_stream *stream,
+			      struct rate_card *card)
```
rustyrussell, line 684: const

plugins/funder.c
```diff
@@ -699,6 +718,8 @@ static void policy_to_json(struct json_stream *stream,
 				 policy->reserve_tank);
 	json_add_num(stream, "fuzz_percent", policy->fuzz_factor);
 	json_add_num(stream, "fund_probability", policy->fund_probability);
+
+	rate_card_to_json(stream, policy->rates);
```
rustyrussell, line 722: Usu called "json_add_rate_card"

plugins/funder.c
```diff
@@ -752,64 +773,179 @@ json_funderupdate(struct command *cmd,
...
+	policy->fuzz_factor = *fuzz_factor;
+	policy->fund_probability = *fund_probability;
+	policy->leases_only = *leases_only;
+
+	/* No rates if we're turning them off */
+	if (leases_avail && !*leases_avail) {
+		new_rates = tal_free(new_rates);
+		plugin_log(cmd->plugin, LOG_DBG,
+			   "leases are now off");
+	}
```
rustyrussell, line 846: Hmm, this is awkward.  Normally we'd set defaults, and let them alter them.

For example, if you are offering all zero leases, we should consider that "don't advertize lease" right?  Maybe that's too awkward: we could do a single `lease=base/basis/weight/feebasemax/feeratemax`, and if you omit a number it's the default?  (Pick a sane default!)

plugins/funder.c
```diff
@@ -752,64 +773,179 @@ json_funderupdate(struct command *cmd,
...
+
+		if (lease_fee_basis)
+			new_rates->lease_fee_basis = *lease_fee_basis;
+		else if (current_policy->rates)
+			new_rates->lease_fee_basis =
+				current_policy->rates->lease_fee_basis;
+
+		policy->rates = tal_steal(policy, new_rates);
+	} else
+		policy->rates = NULL;
```
rustyrussell, line 899: Hmm, policy->rates = tal_free(policy->rates)?

plugins/funder_policy.h
```diff
@@ -19,6 +19,13 @@ enum funder_opt {
 	FIXED,
 };
 
+struct rate_card {
+	u32 channel_basis_max;
+	struct amount_msat channel_base_max;
+	u32 lease_fee_basis;
+	struct amount_sat lease_fee_base;
+};
```
rustyrussell, line 27: I think we should put this as a subtype in the spec.  With weight, ofc.

rustyrussell (changes requested):
Good start, but yes, I think it should be done differently :)

gossipd/gossipd_wire.csv
niftynei: Right, I decided against that because I didn't want to expose an RPC to set the liquidityoffer on lightningd, as that's misleading. Your node physically *can't* fulfill any liquidity offer that you set on lightningd.

This design isn't perfect but it does have the advantage that the *only* way to set/update a liquidity offer is via a plugin that supports fulfilling them. This heads off any potential confusion that might come from looking at the RPCs lightningd supports and thinking you can call `setliquidityoffer` and get the ball rolling...

lightningd/gossip_control.c
niftynei: Mwahahaha :smiling_imp:

lightningd/gossip_control.c
niftynei: I don't think including a max weight in the offer card is a good idea -- a smart node is *always* going to have their peer cover their weight of opening the tx, and that's going to change depending on what their current UTXO set is + the offer requested by the peer.

Consider the case where your advertised weight is 100vbytes, but it's going to cost you 120vbytes to fund a channel for the amount a peer has requested. You're not going to cover that extra 20vbytes. The logical thing would be to update the `lease_fee_base` by the difference to that peer to cover the additional cost.

Pre-committing to the weight means that the other numbers won't be reliable, which will make it harder to do an apples to apples comparison of offers across the board. (which, tbf aren't going to be apples to apples until you get their weights back, but you can't know what that'll be until you ask....). For this reason, we should leave the weight out of the offer advertisement, and allow it to be 'discoverable' at open time -- then it's just a matter of setting sane limits on the opener side for an acceptable weight you're willing to pay for.

otoh it *is* a lot easier to evaluate offers if you know the cost of them all upfront ........

Ok fine you've talked me into it. I'll add a "max weight charge" to the rate card, with the idea that a node should never ask for *more* than this. Good talk!

plugins/funder.c
niftynei: >  lease=base/basis/weight/feebasemax/feeratemax,

This is the sort of problem that YAML solves :wink: 

I don't have anything against defaults, but I did want users to explicitly consider their values (and know how to change them if they wanted to). Forcing them to be set upfront means you have to consider the entire picture before you can enable them -- you can update them singly after the fact but by providing all of them you have convinced me that you've put a little bit of thought into it (and that you know how to change them later.. maybe).

I need to work on the documentation for funder, which will include a sample configuration. This will be the source of the default config values then.

niftynei:
Ok, so I definitely agree that the max_weight_charge belongs in the spec/rate card. Will update. 

Not so convinced that adding an RPC to lightningd to update the rate-card is a good idea. This isn't an amazing solution but it prevents some amount of ham-handedness on the user end.

niftynei (commented):
closed in favor of ##4566

PR #4561, opened 2021-05-27
Reject min channels

niftynei:
I realized after doing this that we have `min-capacity-sat` which lets us reject a channel open who's effective capacity will be below a certain limit.

This is less useful for v2 channels, since we're adding funds to it as well, but you can simply set `min-their-funding` to the same amount and achieve roughly the same thing.

Making this a draft since I don't think we should add it, but open to second opinions.

### DIFF
--- a/plugins/funder.c
+++ b/plugins/funder.c
@@ -550,6 +550,23 @@ json_openchannel2_call(struct command *cmd,
 		return command_hook_success(cmd);
 	}
 
+	/* Now we want to reject any channel that's below our
+	 * `min_their_funding` */
+	if (amount_sat_less(info->their_funding,
+			    current_policy.min_their_funding)) {
+		const char *err
+			= tal_fmt(tmpctx,
+				  "Rejecting channel open from %s."
+				  " Their open is %s, our minimum is %s",
+				  type_to_string(tmpctx, struct node_id, &info->id),
+				  type_to_string(tmpctx, struct amount_sat,
+						 &info->their_funding),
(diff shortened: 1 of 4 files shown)


wtogami:
> I realized after doing this that we have min-capacity-sat which lets us reject a channel open who's effective capacity will be below a certain limit.

This PR might seem to be redundant given the current effect of  `min-capacity-sat` but that older option is **confusing** from a user point of view in that it isn't an exact number. Options:
* It wouldn't be harmful to instead redefine the meaning of `min-capacity-sat`?
* Could also consider deprecation of the old option. Conceptually `min-their-funding` is clearer in intended meaning for both open1 and open2?

What error message does it tell the other side currently if `min-capacity-sat` is not satisfied?

niftynei:
The error we send to the peer is

```
				      "channel capacity with funding %s,"
				      " reserves %s/%s,"
				      " max_htlc_value_in_flight_msat is %s,"
				      " channel capacity is %s, which is below %s",
```

wtogami:
> The error we send to the peer is
> 
> ```
> 				      "channel capacity with funding %s,"
> 				      " reserves %s/%s,"
> 				      " max_htlc_value_in_flight_msat is %s,"
> 				      " channel capacity is %s, which is below %s",
> ```

That seems a bit too wordy? Some human wants to read it and know how big of a channel they are expected. They don't care about the reserves size or max_htlc_value_in_flight at this stage?

PR #4566, opened 2021-05-29
option_will_fund: setting, viewing, + broadcasting lease details

niftynei:
Imports the option_will_fund spec; allows you to set/update/broadcast the rate card out through your node_announcement.

Note that the 'canonical' data for the rate-card lives in funder, and is pushed into lightningd on update.

Supercedes #4560


plugins/funder.c
```diff
@@ -744,85 +749,192 @@ param_policy_mod(struct command *cmd, const char *name,
...
+	json_out_add(jout, "lease_fee_basis", false,
+		     "%d", ad->lease_basis);
+	json_out_add(jout, "lease_weight_max", false,
+		     "%d", ad->funding_weight_max);
+	json_out_addstr(jout, "lease_channel_fee_base_max_msat",
+			type_to_string(tmpctx, struct amount_msat,
+				       &ad->channel_base_msat));
+	json_out_add(jout, "lease_channel_fee_basis_max", false,
+		     "%d", ad->channel_fee_basis);
+	json_out_end(jout, '}');
```
rustyrussell, line 932: How is this different from json_add_liquidity_ad ?

plugins/funder.c
```diff
@@ -864,6 +999,30 @@ static char *amount_option(const char *arg, struct amount_sat *amt)
...
+	return u16_option(arg, &policy->ad->lease_basis);
+}
+
+static char *option_lease_weight_max(const char *arg,
+				     struct funder_policy *policy)
+{
+	if (!policy->ad)
+		policy->ad = liquidity_ad_default(policy);
+	return u16_option(arg, &policy->ad->funding_weight_max);
+}
```
rustyrussell, line 1024: This seems to make option behaviour depend on ordering? I think you need to stash these values and override policy after parsing complete

rustyrussell (commented):
Good bones, minor nits, much about change ordering which you can ignore if too much work.

This breaks my gossipd refac, but this is higher priority so that's all good.

plugins/funder.c
niftynei: I don't see how the ordering matters? Whichever one gets called first will populate the `policy->ad` with the default values, and then update the default to whatever was passed in.

The order shouldn't matter, first one to get called should populate the `ad`, if none gets called then it'll remain NULL.

plugins/funder.c
niftynei: `json_add_liquididty_ad` operates on a `json_stream` object, this is a raw `json_out`

gossipd/gossip_generation.c
```diff
@@ -96,13 +96,14 @@ bool cupdate_different(struct gossip_store *gs,
 		|| !memeq(oparts[1], osizes[1], nparts[1], nsizes[1]);
 }
 
-/* Get non-signature, non-timestamp parts of (valid!) node_announcement */
+/* Get non-signature, non-timestamp parts of (valid!) node_announcement,
+ * with TLV broken out separately  */
```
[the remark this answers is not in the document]
niftynei: Hmm, yes I should have re-ordered this...

lightningd/gossip_control.c
```diff
@@ -460,6 +461,66 @@ static const struct json_command getroute_command = {
...
+			 &lease_base_sat),
+		   p_req("lease_fee_basis", param_number,
+			 &lease_basis),
+		   p_req("lease_weight_max", param_number,
+			 &weight_max),
+		   p_req("lease_channel_fee_base_max_msat", param_msat,
+			 &channel_base_msat),
+		   p_req("lease_channel_fee_basis_max", param_number,
+			 &channel_fee_basis_max),
+		   NULL))
```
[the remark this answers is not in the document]
niftynei: Kind of, yeah. There's some logic to the madness here -- I wanted the config option names to match the RPC exposed names; I wanted the config option names to be namespaced so they're grouped together on sight.

gossipd/gossip_generation.c
```diff
@@ -132,7 +132,8 @@ static void get_nannounce_parts(const u8 *node_announcement,
 /* Is this node_announcement different from prev (not sigs and timestamps)? */
 bool nannounce_different(struct gossip_store *gs,
 			 const struct node *node,
-			 const u8 *nannounce)
+			 const u8 *nannounce,
+			 bool *only_missing_liquidityad)
```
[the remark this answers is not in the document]
niftynei: Correct, need to reorder these commits.

plugins/funder.c
rustyrussell: Then why not set policy->ad before parsing?

niftynei:
It only gets set if one of the options is passed in?

niftynei:
Ok re-wrote a good bit of stuff; got rid of the `liquidity_ad` struct in favor of just using the `lease_rates` subtype everywhere. Also imports/uses the most recent version of the spec proposal.

Issue #4482, opened 2021-04-19
`listpays` item without a `bolt11` entry


fiatjaf:
### Issue and Steps to Reproduce

Can't reproduce this since it's a very weird and obscure bug that sometimes happens on https://t.me/lntxbot.

I have the following flow for when bot users want to pay invoices to the external world:

1. Add an entry to the bot database with the outgoing payment with invoice details (`payment_hash`,  `amount` etc) and `pending=true`.
2. Call `pay` on `lightning-rpc` with the invoice given by the user.
3. Listen for `sendpay_success` and `sendpay_failure` events through lightningd's plugin interface.
4. If a `sendpay_success comes for the given `payment_hash`, change the database entry to `pending=false`.
5. If a `sendpay_failure` comes for the given `payment_hash`, do an immediate check on `listpays -k payment_hash=...` and if the result is `failed`, delete the payment entry from the database; if it's `pending` do nothing or if it's `complete` (won't happen, probably) set it to `pending=false` as above.
6. Also whenever a user sends an invoice to be paid we check if there is an entry on the bot database for that same `payment_hash`, so the user can't have two pending `pay`s with the same hash.

However, something weird is going on and I can't figure out.

Apparently in some very rare cases `listpays` is returning `failed` and so the bot database entry is being deleted (this is the log line from the lntxbot server with the full `listpays` output):

```
lntxbot 2:09PM DBG payment marked as failed after inspection hash=8b8eabe7dda4aba1e5671ffba7d7e323d3f0672819eb4373cea089455df306e3 listpays="{\"pays\":[{\"bolt11":\"lnbc2140u1ps8exqgpp53w82he7a5j46ret8rla604lry0flqeegr845xu7w5zy52h0nqm3sdqqxqyjw5q9qtzqqqq
[...]

PR #4567, opened 2021-05-31
Fix sendpay aggregation and ordering in `listpays`

cdecker:
This is still work in progress, but I wanted to share the current
state. The main goal is to address #4481 and #4482, i.e., fix the
ordering and address flip-flopping states on supposedly finished
payments. It takes a bit of a detour to get there. The idea is to add
a grouping key `groupid` to `sendpay` instances, such that all
`sendpay` instances that originate from the same `pay` invocation
share the same `groupid`. This way `listpays` can group them into the
correct group when aggregating `sendpays` for each `pay`
invocation. However, while implementing this simple approach I noticed
we actually delete prior attempts automagically when retrying them,
which still results in flip-flopping, and is strange if you ask me: if
I called `pay` 2 times, I should see it 2 times in `listpays`.

So we now no longer delete prior attempts, but it also means we need
to add `groupid` to any RPC call that may refer to a specific
`sendpay` call: `waitsendpay`, `sendonion`, `sendpay`, etc. In
addition we now need to annotate HTLCs we have in the DB with which
`(payment_hash, partid, groupid)` triple they refer to in order to
correctly route results back to the caller.

I'm still working on some failing tests, but it should just be a
matter of time now :-)


plugins/pay.c
```diff
@@ -1936,6 +1936,76 @@ static const char *init(struct plugin *p,
...
+		const jsmntok_t *status, *grouptok;
+		grouptok = json_get_member(buf, t, "groupid");
+		json_to_u64(buf, grouptok, &groupid);
+
+		/* New group, reset what we collected. */
+		if (last_group != groupid) {
+			completed = false;
+			pending = false;
+			last_group = groupid;
+		}
```
rustyrussell: This assumes that listsendpays is listed in increasing groupid order?   That assumption does not seem documented?

lightningd/pay.c
```diff
@@ -1487,16 +1495,21 @@ static struct command_result *json_waitsendpay(struct command *cmd,
...
 		   p_opt("timeout", param_number, &timeout),
 		   p_opt_def("partid", param_u64, &partid, 0),
+		   p_opt("groupid", param_u64, &groupid),
 		   NULL))
 		return command_param_failed();
 
-	res = wait_payment(cmd->ld, cmd, rhash, *partid);
+	if (groupid == NULL) {
+		groupid = tal(cmd, u64);
+		*groupid = wallet_payment_get_groupid(cmd->ld->wallet, rhash);
```
rustyrussell, line 1551: Would prefer this function to be called "wallet_payment_get_last_groupid" to be clearer? (or max_groupid)?

plugins/pay.c
```diff
@@ -1979,6 +1990,30 @@ payment_listsendpays_previous(struct command *cmd, const char *buf,
 			completed = false;
 			pending = false;
 			last_group = groupid;
+
+			parts = 1;
+			json_scan(tmpctx, buf, t,
```
rustyrussell, line 2204: Hmm, is there no group_id 0?  Since you initialize last_group to 0, it seems you won't run this for the first group.  Since msat is not initialized elsewhere, this means you will use it uninitialized below?

plugins/pay.c
cdecker: This is a `listsendpays` call limited to the `payment_hash` of the invoice. There cannot be multiple groups in flight concurrently. So ordering by `id` which we do internally in `lightningd/pay.c` will result in the desired ordering of `groupid`. I'll add a comment to that effect 👍

plugins/pay.c
cdecker: Could either make it a pointer to make that explicit, but the msats field is only used if we detect that there was a prior success or failure and replicate the result for idempotency, so not having a prior attempt means that we'll not use msats at al

PR #4568, opened 2021-05-31
Give the possibility to return the list of closed channel in the database

### LINKED ISSUE #4425: RPC - List of closed channels
I looked at the docs and the issues in GH but couldn't figure out how to get a list of closed channels.

I know they're in the db because I can see them.

Is this something that could find its way to one of the RPC methods ? If so which one ?

vincenzopalazzo:
This is a working in progress PR, basically because I don't like the actual status of my code.

- [ ] Define the correct channels JSON format;
- [ ] Give a unique method to add the channel in the JSON response;
- [ ] Give some sorting type of channels;
- [ ] Add more tests to cover all the corned cases.
- [ ] Improve C quality

Fixes #4425

### DIFF
--- a/lightningd/gossip_control.c
+++ b/lightningd/gossip_control.c
@@ -497,7 +497,7 @@ struct listchannels_info {
 	struct json_stream *response;
 	struct short_channel_id *id;
 	struct node_id *source;
-	u32 *closed;
+	bool *closed;
 };
 
 /* Called upon receiving a getchannels_reply from `gossipd` */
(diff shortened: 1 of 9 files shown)


rustyrussell:
Ah, I have a half-working PR which removes this code from gossipd and puts it into a new plugin.  That will conflict with this work!

However, you don't actually want to change listchannels, I think, but listpeers?

vincenzopalazzo:
Ops! Thanks for pinging me @rustyrussell, I will close it and I will move the work in the `wallet.c` in another PR.

>However, you don't actually want to change listchannels, I think, but listpeers?

Much better, you are right. Make sense to me

I will subscribe to the PR event and when you finish your work with the refactoring I will start the work on this command.

Thanks!

LightningHelper:
@rustyrussell is there no way to see all of my node closed channels?
thanks

vincenzopalazzo:
@LightningHelper this is one of my goals in the next release 0.10.2, we will work on it. For the moment you can only query the database.

LightningHelper:
Thanks.
I have a question if you may, I assume the db does not keep track on all closed channels in the network, but how does it make sure to keep track on the closed channels of my node?

vincenzopalazzo:
Hi @LightningHelper,

>I have a question if you may, I assume the db does not keep track on all closed channels in the network, but how does it make sure to keep track on the closed channels of my node?

I did check the code right now, but in this case, the node has a record in the database, and this record is updated when the status of the channel changes. In this way, you are sure that all the channels are inside the database. However, it is possible that the database may contain all the channels if the node gives the possibility to remove it with some command that I think it is not implemented right now in the actual implementation of c-lightning

I hope I answer your question.

PR #4572, opened 2021-06-01
tests: cleanup any unfinished/hanging channel opens

niftynei:
Otherwise it hangs on cleanup, as l2 is waiting for l1 to respond (and
it's not going to), so the memleak call hangs.

Changelog-None.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4572/commits/89bdbea1b5e1f23456e48b3c64079cc7bae2381c

PR #4574, opened 2021-06-01
Revert "pytest: Skip hsm encryption test if we don't have a TTY"

darosior:
I don't see any reason for checking whether the stdout is a tty. In addition i'm pretty sure `pytest` captures `stdout` and that means this test was actually never run.

 I did a first dry-run on my fork's Github Action, and it doesn't need it either.

### DIFF
--- a/tests/test_wallet.py
+++ b/tests/test_wallet.py
@@ -12,7 +12,6 @@ from utils import (
 import os
 import pytest
 import subprocess
-import sys
 import unittest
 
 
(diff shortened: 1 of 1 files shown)


cdecker:
The reason to check whether `sys.stdin`/`sys.stdout` is a TTY in the python test is that this will also be the case for the underlying `hsmtool`. Since the `hsmtool` does a bit of TTY magic to disable echo and may not flush the line prompting the user to enter the passphrase (which is the trigger for us to enter) we can end up hanging indefinitely in that test. Happy to re-enable if that problem is not longer present, but it was causing a lot of noise on the CI for considerable time, so we better make sure it doesn't happen before re-enabling.

rustyrussell:
Ack b228f94988f33b228c9ac27a799fd4bad5506215

I'll keep an eye on it and see if it makes things worse...

PR #4582, opened 2021-06-04
pyln.client gossmap support

rustyrussell:
This is a minimal implementation; it probably wants some more helpers to make it less painful to use.

@m-schmoock  ?


m-schmoock:
Thanks, will have a look..

cdecker:
Rebased on top of `master` to see if the CI failure was caused by the merge-base.

cdecker:
I have absolutely no idea how changes in the pyln libs can cause `peer_exp_wiregen.h` to be regenerated... Tried it a couple of times locally and could reproduce.

m-schmoock:
@cdecker
Strange. Found that in the logs too:
(it trieds to get Gossmap which pulls in `pyln.spec.bolt7 import (channel_announcement, channel_update, ..)`
Still I dont get the wiregen update...

```
+ pip3 install --user -U -r requirements.txt
Processing ./contrib/pyln-client
  DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
   pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
    ERROR: Command errored out with exit status 1:
     command: /opt/hostedtoolcache/Python/3.6.13/x64/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-53ma3np9/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-53ma3np9/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-usrh6h8g
         cwd: /tmp/pip-req-build-53ma3np9/
    Complete output (9 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-53ma3np9/setup.py", line 2, in <module>
        from pyln import client
      File "/tmp/pip-req-build-53ma3np9/pyln/client/__init__.py", line 3, in <module>
        from .gossmap import Gossmap
      File "/tmp/pip-req-build-53ma3np9/pyln/client/gossmap.py", line 3, in <module>
        from pyln.spec.bolt7 import (channel_announcement, channel_update,
    ModuleNotFoundError: No module named 'pyln.spec'
    ----------------------------------------
WARNING: Discarding file:///home/runner/work/lightning/lightning/contrib/pyln-client. Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

rustyrussell:
Actually, experimental builds always regen (we don't ship the exp_ files).

It's the pip install failure that causes it to abort, not installing mako, and get upset.

fiatjaf:
Does this mean we can add new channels to the gossip map while gossipd is running and they will be picked up automatically?

m-schmoock:
@rustyrussell 
I haven't forgot about this, looking forward to use this in rebalance :)

contrib/pyln-client/setup.py
```diff
@@ -9,8 +8,14 @@
 with io.open('requirements.txt', encoding='utf-8') as f:
     requirements = [r for r in f.read().split('\n') if len(r)]
 
+# setup shouldn't try to load module, so we hack-parse __init__.py
+with io.open('pyln/client/__init__py', encoding='utf-8') as f:
```
m-schmoock, line 12: You missed a '.' (dot) character:
```python
with io.open('pyln/client/__init__.py', encoding='utf-8') as f:
```

m-schmoock (commented):
A thing I noted...

m-schmoock:
@rustyrussell
Sorry for not being able to run this yet. This is what I did:
I manged to install bolt specs manually by going toi their directories, i.e.  
```bash
cd contrib/pyln-spec/bolt1
python3 setup.py install --user
```

This was only doable after upgrading to newest pip `sudo pip install --upgrade pip`. Still the command errors in the end, but the bolt package installs anyway:
```
....
Copying pyln_bolt1-1.0.1.137-py3.9.egg to /home/will/.local/lib/python3.9/site-packages
pyln-bolt1 1.0.1.137 is already the active version in easy-install.pth

Installed /home/will/.local/lib/python3.9/site-packages/pyln_bolt1-1.0.1.137-py3.9.egg
Processing dependencies for pyln-bolt1==1.0.1.137
Searching for pyln.proto>=0.9.3
Reading https://pypi.org/simple/pyln.proto/
No local packages or working download links found for pyln.proto>=0.9.3
error: Could not find suitable distribution for Requirement.parse('pyln.proto>=0.9.3')
```

This can be ignored as I am able to install pyln-proto also manually by source. Not nice though. What I find strange that under the quoted URL https://pypi.org/simple/pyln.proto/ there is a version 0.9.3  °_° ?

When I then run the gossmap testcase it still complains about pyln.spec module missing maybe because I only have the bolts and not the pyln.spec 'parent module' ?:
```
../../cln/contrib/pyln-client/pyln/client/__init__.py:3: in <module>
    from .gossmap import Gossmap, GossmapNode, GossmapChannel, GossmapNodeId
../../cln/contrib/pyln-client/pyln/client/gossmap.py:3: in <module>
    from pyln.spec.bolt7 import (channel_announcement, channel_update,
E   ModuleNotFoundError: No module named 'pyln.spec'
```

m-schmoock:
Ok, I got pyln-spec 'working' by manually copying it into my python environment. The `setup.py` does not work for me or I don't know how...

m-schmoock:
@rustyrussell
Ok, I got pyln-spec 'working' by manually copying it into my python environment. The setup.py does not work for me or I don't know how...

I messed around a bit with it. It seems like a meta information is missing like i.e. capacity, fees, delay ...
How is a plugin supposed to know that without querying RPC for each channel and node?

m-schmoock:
@rustyrussell 
I Added some additions by PR https://github.com/rustyrussell/lightning/pull/7 into your repo

rustyrussell:
> Again I added more stuff in coop/wip/pr [rustyrussell#8](https://github.com/rustyrussell/lightning/pull/8)
> 
> Also:
> 
> > Does this mean we can add new cannels to the gossip map while gossipd is running and they will be picked up automatically?
> 
> @rustyrussell you thumb-upped this question. Are we really going to push back changes to gossmap back to the gossipd daemon?

No, I misread!  You can use `addgossip` to send gossip messages to gossipd, and it will add it to the store if it accepts them though.

rustyrussell:
> Question:
> Do we want the structure like currently implemented: Gossmap -> list[node] -> list[channel] -> [half_channel_1, half_channel_2]
> Or do we want it more like flat structure of `lightning-cli listchannels` list[half_channel] or do we want both by keeping current and adding iterator functions?

The current structure is more useful: the JSON style of listchannels makes for a lot of redundancy.

rustyrussell:
Once CI is happy, I will apply this to master.

Also, @m-schmoock you now have write access to the repo (though after it's included in master you can simply open new PRs).

m-schmoock:
Yep, I don't think its done, but since nobody relies on it yet we can add more later

contrib/pyln-client/requirements.txt
```diff
@@ -1 +1,2 @@
 recommonmark>=0.7.*
+pyln-bolt7==1.0.1.*
```
vincenzopalazzo, line 2: better avoid fixing the version to a minimum requirement version?

Maybe it is better use the `pyln-bolt7>=1.0.1.7` to avoid that pip said "hey there is a dependence that wants a new package".

vincenzopalazzo (commented):
I'm the last person that touches these dependents, and now I strongly agree with the sentence "when somethings works don't touch it".

I make some experiments on my branch, now the problem it is the missing module `mako.template ` ([see here](https://github.com/vincenzopalazzo/lightning/runs/3512538537#step:5:1691)) and I'm just asking, I'm looking in the right version of the branch?

contrib/pyln-client/requirements.txt
vincenzopalazzo: Just thinking too late, maybe the dependences are skipped (not sure what dependences), because there is this type of conflict?

```
The conflict is caused by:
    The user requested pyln-proto 0.10.1 (from /home/runner/work/lightning/lightning/contrib/pyln-proto)
    pyln-bolt7 1.0.1.137 depends on pyln.proto==0.8.4
```

https://github.com/ElementsProject/lightning/pull/4582/checks?check_run_id=3512149678#step:5:222

vincenzopalazzo:
>I'm the last person that touches these dependents, and now I strongly agree with the sentence "when somethings works don't touch it".

>I make some experiments on my branch, now the problem it is the missing module mako.template (see here) and I'm just asking, I'm looking in the right version of the branch?

Auto answer, mako is skipped because of this https://github.com/ElementsProject/lightning/pull/4582#discussion_r702258102

contrib/pyln-client/requirements.txt
vincenzopalazzo: Here we go, it is my fault, we upgrade the pyln-proto >= 0.9.3 https://github.com/ElementsProject/lightning/pull/4679, this pull from pip the last version, so pyln-proto 0.10.1. However, on the pip repository, we miss the last version of the module, as @cdecker points out in the last meeting last week.

They are also a couple of messages on the [IRC log](https://gnusha.org/c-lightning/2021-08-23.log) where is mentioned this problem of dependencies missing.


```
04:15 < vincenzopalazzo> Guys stupid question, the new version of the package pyln-sec it is update with the new c-lightning version? because I'm getting the following warning `ERROR: pyln-bolt1 1.0.1.137 has requirement pyln.proto==0.8.4, but you'll have pyln-proto 0.10.1 which is incompatible.` but it think this is over with the following PR https://github.com/ElementsProject/lightning/pull/4679/files
04:19 < cdecker[m]> That's because the version on pypi is really old, and still predates that PR. I'll look into automating deployments of the pypi artifacts via GH actions
04:31 -!- rusty [~rusty@103.93.169.121] has quit [Quit: Leaving.]
04:33 < vincenzopalazzo> Christian Decker: Thanks for this :) I'm try to fix the warning in lnprototest
```

Sorry about that, a solution that I found to make CI happy again, at least of python dependencies is to remove the dependencies in this `requirements.txt` file, and all look fine again, but only with CI (see https://github.com/vincenzopalazzo/lightning/runs/3512820849#step:5:2194).

rustyrussell:
Updated on top of #4763

contrib/pyln-client/requirements.txt
```diff
@@ -1 +1,2 @@
 recommonmark>=0.7.*
+pyln-bolt7
```
vincenzopalazzo, line 2: It is better to fix the version here with >= latest one? Maybe pip can take the decision to use the old one if in the system there is an older version?

So we are putting here some dependencies that can cause errors in production,  I will preferer to have a `pyln-bolt7>=1.0.2.186`

vincenzopalazzo (approved):
only a small nit, by the way, excited to have the gossip map in here.

Fancy stuff can happens.

contrib/pyln-client/requirements.txt
m-schmoock: Hm, theres something about the pyln-spec dependency.

I put the `gossip_store_channel_amount` into the bolt7 spec package, not sure if I was supposed to, see  32cb69ef. but the one that will be installed with pip is still the old one which results in:

```
ImportError: cannot import name 'gossip_store_channel_amount' from 'pyln.spec.bolt7' 
```

So we need to put `bolt7 >= 1.0.2.187` (last digit incremented) since the 186 does not have the `gossip_store_channel_amount` in its csv

m-schmoock: 187 is not released yet, but it will be the one with the changes we made in this branch... not sure how we handle something like this

m-schmoock:
Added another testcase for the half channel stuff... no bugs detected.

contrib/pyln-client/requirements.txt
rustyrussell: I did the pypi upload, even though we hadn't merged, because I needed it to fix lnprototest...

rustyrussell: Oh, I missed that change, it's completely wrong!

I will fix that a different way tomorrow...

rustyrussell:
OK, I removed the change to the spec package, re-running CI now.

m-schmoock:
just re-added the commit for the half channel testcase that got missed by rebasing....

m-schmoock:
ACK 56cac767

PR #4585, opened 2021-06-05
plugins/topology: remove topology functionality from gossipd.

rustyrussell:
After this, gossipd's main responsibility is just to create our own gossip messages, and receive those from others.  As a nice bonus, listchannels, listnodes and getroute don't use gossipd, so they don't block its more important functions.


rustyrussell:
Seems to hit CI flake, fixed in #4596

rustyrussell:
Seems to hit CI flake, fixed in #4596

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4585/commits/a0e10204324f2468bd417cf0e6b4576a72d37e43

PR #4587, opened 2021-06-07
chore: fix compile issue discovered by gcc

m-schmoock:
The local struct `hash` was read before it was initialized.

Changelog-None

### DIFF
--- a/bitcoin/signature.c
+++ b/bitcoin/signature.c
@@ -219,8 +219,8 @@ bool check_tx_sig(const struct bitcoin_tx *tx, size_t input_num,
 	}
 	assert(input_num < tx->wtx->num_inputs);
 
-	dump_tx("check_tx_sig", tx, input_num, script, key, &hash);
 	bitcoin_tx_hash_for_sig(tx, input_num, script, sig->sighash_type, &hash);
+	dump_tx("check_tx_sig", tx, input_num, script, key, &hash);
 
 	ret = check_signed_hash(&hash, &sig->s, key);
 	if (!ret)


bitcoin/signature.c
```diff
@@ -219,7 +219,11 @@ bool check_tx_sig(const struct bitcoin_tx *tx, size_t input_num,
 	}
 	assert(input_num < tx->wtx->num_inputs);
 
+	/* gcc version 11.1.0 gives error: ‘hash’ may be used uninitialized */
+	#pragma GCC diagnostic push
+	#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
 	dump_tx("check_tx_sig", tx, input_num, script, key, &hash);
+	#pragma GCC diagnostic pop
```
rustyrussell, line 226: There's a good rule about pragmas.  And it is never to use them :)

But in this case, gcc is correct!  This line should be moved down one (it's only relevant when compiled with DEBUG, ofc)

rustyrussell (changes requested):

bitcoin/signature.c
m-schmoock: Since different compilers don't seem to like these pragma (i.e. clang, strange thou that it does not ignore a pragma thath has `GCC` as first statement) , isn't it better to solve this by initing with `struct sha256_double 
 hash = {0};` ?

m-schmoock: @rustyrussell  updated to zero initialize, okay? Or do we want compatible pragmas that work for all compilers?

rustyrussell: Please read what I wrote: the compiler has found a bug.

Can't you just move the dump_tx line down under the bitcoin_tx_hash_for_sig?

m-schmoock: whoops, I thought your comment was related to the pragma line, which did'nt make sense to me.
Right, extracting the hash before using it makes more sense :D

rustyrussell: Sorry, my fault: I was on mobile, so couldn't quote the details easily, sorry.

Ack https://github.com/ElementsProject/lightning/commit/fe7f62cd226e9703b7422f1227ec9770c412d8de

rustyrussell (approved):

rustyrussell (approved):
Ack fe7f62cd226e9703b7422f1227ec9770c412d8de

PR #4589, opened 2021-06-08
EXPERIMENTAL: wumbo htlc support

rustyrussell:
Implements https://github.com/lightningnetwork/lightning-rfc/pull/877


rustyrussell:
We won't have a feature bit it seems: we'll simply wait for electrum to relax its recv-side restriction (since other implementations will send without checking anyway) and once that is widely deployed we'll remove our send-side restriction.

Issue #4551, opened 2021-05-21
Race condition in test harness teardown


niftynei:
Some CI tests fail at teardown?

```
==================================== ERRORS ====================================
_______________ ERROR at teardown of test_backfill_scriptpubkeys _______________
[gw2] linux -- Python 3.6.13 /opt/hostedtoolcache/Python/3.6.13/x64/bin/python3

request = <SubRequest 'directory' for <Function test_backfill_scriptpubkeys>>
test_base_dir = '/tmp/ltests-du69g5u_'
test_name = 'test_backfill_scriptpubkeys'

    @pytest.fixture
    def directory(request, test_base_dir, test_name):
        """Return a per-test specific directory.
    
        This makes a unique test-directory even if a test is rerun multiple times.
    
        """
        global __attempts
        # Auto set value if it isn't in the dict yet
        __attempts[test_name] = __attempts.get(test_name, 0) + 1
        directory = os.path.join(test_base_dir, "{}_{}".format(test_name, __attempts[test_name]))
        request.node.has_errors = False
    
        if not os.path.exists(directory):
            os.makedirs(directory)
    
        yield directory
    
        # This uses the status set in conftest.pytest_runtest_makereport to
        # determine whether we succeeded or failed. Outcome can be None if the
        # failure occurs during the setup phase, hence the use to getattr instead
        # of accessing it directly.
        rep_call = getattr(request.node, 'rep_call', None)
        outcome = 'passed' if rep_call is None else rep_call.outcome
        failed = not outcome or request.node.has_errors or outcome != 'passed'
    
        if not failed:
            try:
>               shutil.rmtree(directory)

contrib/pyln-testing/pyln/testing/fixtures.py:92: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/shutil.py:486: in rmtree
    _rmtree_safe_fd(fd, path, onerror)
/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/shutil.py:428: in _rmtree_safe_fd
    onerror(os.rmdir, fullname, sys.exc_info())
[...]

PR #4595, opened 2021-06-11
Support the the filtering option by status in the listpays and listsendpays

### LINKED ISSUE #4588: feature request: list of recently-failed payments
Got a request to be able to see what the most recent payment failures were for a node, as a way to help decide where to place channels.

vincenzopalazzo:
This is a working in progress feature and it is a proposal to give the user to see the possibility to see the payment filtered by status.

Fixes #4588

- [X] Support new propriety in the last command that is status
- [X] Sanity check in the status propriety
- [X] Adding a test to make sure that the filter option works
- [X] Update docs

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -935,7 +935,7 @@ class LightningRpc(UnixDomainSocketRpc):
         """
         Show outgoing payments, regarding {bolt11} or {payment_hash} if set
         Can only specify one of {bolt11} or {payment_hash}. It is possible
-        filter the payments by status.
+        filter the payments by {status}.
         """
         assert not (bolt11 and payment_hash)
         payload = {
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -931,15 +931,17 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("listnodes", payload)
 
-    def listpays(self, bolt11=None, payment_hash=None):
+    def listpays(self, bolt11=None, payment_hash=None, status=None):
         """
         Show outgoing payments, regarding {bolt11} or {payment_hash} if set
-        Can only specify one of {bolt11} or {payment_hash}.
+        Can only specify one of {bolt11} or {payment_hash}. It is possible
+        filter the payments by status.
         """
         assert not (bolt11 and payment_hash)
         payload = {
             "bolt11": bolt11,
-            "payment_hash": payment_hash
+            "payment_hash": payment_hash,
+            "status": status
         }
         return self.call("listpays", payload)
 
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -955,11 +955,12 @@ class LightningRpc(UnixDomainSocketRpc):
         }
         return self.call("listpeers", payload)
 
-    def listsendpays(self, bolt11=None, payment_hash=None):
+    def listsendpays(self, bolt11=None, payment_hash=None, status=None):
         """Show all sendpays results, or only for `bolt11` or `payment_hash`."""
         payload = {
             "bolt11": bolt11,
-            "payment_hash": payment_hash
+            "payment_hash": payment_hash,
+            "status": status
         }
         return self.call("listsendpays", payload)
 
--- a/doc/lightning-listpays.7.md
+++ b/doc/lightning-listpays.7.md
@@ -4,13 +4,14 @@ lightning-listpays -- Command for querying payment status
 SYNOPSIS
 --------
 
-**listpays** \[bolt11\] \[payment_hash\]
+**listpays** \[bolt11\] \[payment_hash\] \[status\]
 
 DESCRIPTION
 -----------
 
 The **listpay** RPC command gets the status of all *pay* commands, or a
 single one if either *bolt11* or *payment_hash* was specified.
+It is possible filter the payments also by status.
 
 RETURN VALUE
 ------------
--- a/doc/lightning-listpays.7.md
+++ b/doc/lightning-listpays.7.md
@@ -11,7 +11,7 @@ DESCRIPTION
 
 The **listpay** RPC command gets the status of all *pay* commands, or a
 single one if either *bolt11* or *payment_hash* was specified.
-It is possible filter the payments also by status.
+It is possible filter the payments also by *status*.
 
 RETURN VALUE
 ------------
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -3348,103 +3348,43 @@ wallet_payment_list(const tal_t *ctx,
 	size_t i;
 
 	payments = tal_arr(ctx, const struct wallet_payment *, 0);
-        //FIXME: Make the if-else smaller
-	// A possible solution is divided the string in two part (pre-where and post-where) and
-	// use a small if else to set the remain string, with this method we can have small code.
-	if (payment_hash && status) {
-		stmt =
-			db_prepare_v2(wallet->db, SQL("SELECT"
-						      "  id"
-						      ", status"
-						      ", destination"
-						      ", msatoshi"
-						      ", payment_hash"
-						      ", timestamp"
-						      ", payment_preimage"
-						      ", path_secrets"
-						      ", route_nodes"
-						      ", route_channels"
-						      ", msatoshi_sent"
-						      ", description"
-						      ", bolt11"
-						      ", failonionreply"
-						      ", total_msat"
-						      ", partid"
-						      ", local_offer_id"
-						      " FROM payments"
-						      " WHERE payment_hash = ? AND status = ?"
-						      " ORDER BY id;"));
+
+	u8 enable_payment_hash = payment_hash != NULL ? 0 : 1;
+	u8 enable_status = status != NULL ? 0 : 1;
+
+	stmt = db_prepare_v2(wallet->db, SQL("SELECT"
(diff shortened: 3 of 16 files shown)


vincenzopalazzo:
The GitHub action failure seems unrelated https://github.com/ElementsProject/lightning/runs/2821322348?check_suite_focus=true#step:5:2397

It looks like related to #4551.

I retry to be lucky with GitHub actions, lets see 💺

vincenzopalazzo:
The failure in CI seems to be unrelated.

What do you think about the point` refactoring in the wallet.c to make the if-else smaller and more readable` @rustyrussell do your think that is a good refactoring in the wallet.c file?

I want to make an append operation from a default string before making the query. in this way, we have a smaller if-else, but I don't know if it is an antipattern in C

wallet/wallet.c
```diff
@@ -3366,6 +3367,32 @@ wallet_payment_list(const tal_t *ctx,
 						  " WHERE payment_hash = ?"
 						  " ORDER BY id;"));
 		db_bind_sha256(stmt, 0, payment_hash);
+	} else if (status) {
+		// TODO(vincenzopalazzo): Missing the filter options are both not null
+		// A possible solution is divided the string in two part (pre-where and post-where) and
+		// use a small if else to set the remain string, with this method we can have small code.
```
cdecker: That would not work with our prepared statement + SQL dialect specialization. Better to have a generic query like this:

`SELECT * FROM payments WHERE (status = ? OR ?) AND (payment_hash = ? OR ?)`

This allows us to disable either restriction by passing `0` or `1` depending on whether we want to filter by that clause or not.

cdecker: So the bindings look like this:

 - `NULL, 1, NULL, 1` for no filtering
 - `complete, 0, NULL, 1` for filtering by `status`
 - `NULL, 1, aaaaaaaa..., 0` for filtering by `payment_hash`
 - `complete, 0, aaaaaaa., 0` for combined filtering by both
 
Prevents dynamic query building, while still enabling full flexibility.

wallet/wallet.c
```diff
@@ -3343,7 +3343,35 @@ wallet_payment_list(const tal_t *ctx,
 	size_t i;
 
 	payments = tal_arr(ctx, const struct wallet_payment *, 0);
-	if (payment_hash) {
+        //FIXME: Make the if-else smaller
+	// A possible solution is divided the string in two part (pre-where and post-where) and
+	// use a small if else to set the remain string, with this method we can have small code.
```
cdecker, line 3348: Or we can do this :-) Might not work if we extend the query args much more to have the full matrix of combinations.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -930,15 +930,17 @@ def listnodes(self, node_id=None):
         }
         return self.call("listnodes", payload)
 
-    def listpays(self, bolt11=None, payment_hash=None):
+    def listpays(self, bolt11=None, payment_hash=None, status=None):
         """
         Show outgoing payments, regarding {bolt11} or {payment_hash} if set
-        Can only specify one of {bolt11} or {payment_hash}.
+        Can only specify one of {bolt11} or {payment_hash}. It is possible
+        filter the payments by status.
```
cdecker, line 937: `status` -> `{status}` to match the formatting of the other params.

doc/lightning-listpays.7.md
```diff
@@ -4,13 +4,14 @@ lightning-listpays -- Command for querying payment status
...
 
-**listpays** \[bolt11\] \[payment_hash\]
+**listpays** \[bolt11\] \[payment_hash\] \[status\]
 
 DESCRIPTION
 -----------
 
 The **listpay** RPC command gets the status of all *pay* commands, or a
 single one if either *bolt11* or *payment_hash* was specified.
+It is possible filter the payments also by status.
```
cdecker, line 14: Nit: `status` -> `*status*`

cdecker (approved):
Just some minor formatting issues in the docs

ACK 4d3ed04fd46522395ec60669a5be87e28156d94e

contrib/pyln-client/pyln/client/lightning.py
vincenzopalazzo: Thanks for this fix @cdecker and sorry for the delay in fixing it.

wallet/wallet.c
vincenzopalazzo: This looks like more of a hack solution instead of a real solution to refactoring the code. I will make some experiments with your feedback in the generic query.

Thanks.

wallet/wallet.c
vincenzopalazzo: It feedback was amazing, thank @cdecker!

vincenzopalazzo:
Rebased on top of master, and add missing fix around the code, like formatting fix and doc fixes

contrib/pyln-client/pyln/client/lightning.py
cdecker: No worries, thank you for the time you are investing ^^

wallet/wallet.c
```diff
@@ -3351,104 +3351,41 @@ wallet_payment_list(const tal_t *ctx,
...
-						     ", total_msat"
-						     ", partid"
-						     ", local_offer_id"
-						     " FROM payments"
-						     " ORDER BY id;"));
-	}
-	db_query_prepared(stmt);
+
+	uint8_t enable_payment_hash = payment_hash != NULL ? 0 : 1;
+	uint8_t enable_status = status != NULL ? 0 : 1;
```
cdecker, line 3356: I think we have a `u8` alias for this type, just for consistency sake, but ideally we'd have a `bool` type, which I assume you were not using here due to it missing in `db_bind_*`, is that correct?

wallet/db.c
```diff
@@ -1697,7 +1697,10 @@ void db_bind_preimage(struct db_stmt *stmt, int pos, const struct preimage *p)
 
 void db_bind_sha256(struct db_stmt *stmt, int pos, const struct sha256 *s)
 {
-	db_bind_blob(stmt, pos, s->u.u8, sizeof(struct sha256));
+	if(!s)
+		db_bind_null(stmt, pos);
+	else
+		db_bind_blob(stmt, pos, s->u.u8, sizeof(struct sha256));
```
cdecker, line 1703: NACK, the DB API is very simple on purpose and should not decide the nullability of a field on its own. That's for the `wallet.c` to disentangle where we actually have more information about the schema we're dealing with. This could potentially introduce some really subtle bugs.

wallet/wallet.c
```diff
@@ -3377,15 +3377,15 @@ wallet_payment_list(const tal_t *ctx,
 				      " WHERE (payment_hash = ? OR ?) AND (status = ? OR ?)"
 				      " ORDER BY id;"));
 
-        db_bind_sha256(stmt, 0, payment_hash == NULL ? NULL : payment_hash);
+	db_bind_sha256(stmt, 0, payment_hash);
```
cdecker: This is exactly the right place to disambiguate whether the value is `NULL` or not, no need to push it down the stack.

cdecker (changes requested):
The last commit needs to be removed, since at the raw DB layer we don't have the required context on the schema to even know if the field is nullable in the first place. There are several updates to the `lnprototest` submodule which I think are unwranted, could you filter those out?

Otherwise excellent proposal and rework :+1:

wallet/wallet.c
vincenzopalazzo: I agree, is more clear to read also from a user viewpoint.

Thanks @cdecker

wallet/db.c
vincenzopalazzo: I see, and I agree with your design choice. this check usually is also scary in the API.

Thanks 😸

wallet/wallet.c
vincenzopalazzo: Oh, this is a great this, sorry if I use it, I think I bring inside some C++ convention here, sorry. But in reality, I will never understand if there is some difference from uint8_t u8 :)

I didn't use the boolean here, because I'm using SQLite in one of my side projects with android and flutter, and I know that sqlfile has no support for boolean type.

I just typed on google and I receive the following answer https://www.sqlite.org/datatype3.html

However, I don't know if I'm missing some things, I think have a boolean type is very cool, and I can change it if is supported.

vincenzopalazzo (commented):
>There are several updates to the lnprototest submodule which I think are unwranted, could you filter those out?

@cdecker I think this message was referring to an old commit, I check the actual file change history and there are no lnprototest changes here, but I remember that I have a bad life some day ago with the lnprototest update.

There is other stuff there I need to do regarding to lnprototest?

wallet/db.c
```diff
@@ -1697,7 +1697,7 @@ void db_bind_preimage(struct db_stmt *stmt, int pos, const struct preimage *p)
 
 void db_bind_sha256(struct db_stmt *stmt, int pos, const struct sha256 *s)
 {
-	db_bind_blob(stmt, pos, s->u.u8, sizeof(struct sha256));
+        db_bind_blob(stmt, pos, s->u.u8, sizeof(struct sha256));
```
cdecker, line 1700: This is a no-op change, could you remove it just for cleanliness' sake? :sunglasses:

cdecker (approved):

wallet/db.c
vincenzopalazzo: ups! I think my emacs has some problem  with the space indentation (?) I didn't see it before on my editor.

Anyway, it is a good catch, space is important. Thanks!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/72c13d70b93e9d902c7728a150cbd4b84426f70a

PR #4596, opened 2021-06-13
pytest: fix flake in test_misc.py::test_funding_reorg_*

rustyrussell:
If l2 doesn't have the funding locked in, the rest of the test fails
(we got a timeout on `wait_for(lambda: [c['active'] for c in l2.rpc.listchannels('103x1x0')['channels']] == [False, False])`)

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4596/commits/b1b955f2cbbc8358a2bc9426753b8ad6592027ed

PR #4571, opened 2021-06-01
hsm_encryption: read from STDIN if not in a TTY

openoms:
fixes: #4570 as discussed.

The temporary terminal with ECHO disabled is not created if the input is piped instead of typed in a terminal.

Passwords are still not visible when typed, but allows for automation.

PR #4598, opened 2021-06-14
hsmtool: test password input from stdin

darosior:
This includes a test from https://github.com/ElementsProject/lightning/pull/4571#issuecomment-861047543 for roundtrip between stdin and non-stdin, as well as a cleanup in the line reader to avoid any possibility of divergence between the two.

Changelog-None

### DIFF
--- a/common/hsm_encryption.c
+++ b/common/hsm_encryption.c
@@ -80,6 +80,18 @@ void discard_key(struct secret *key TAKES)
 		tal_free(key);
 }
 
+/* Read a line from stdin, do not take the newline character into account. */
+static bool getline_stdin_pass(char **passwd, size_t *passwd_size)
+{
+	if (getline(passwd, passwd_size, stdin) < 0)
+		return false;
+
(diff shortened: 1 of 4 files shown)


cdecker:
Seems tailing a process that is about to terminate can be racy. We could wait for it to exit and then look at its accumulated logs?

darosior:
Yes i've a patch that i'm about to push :)

darosior:
Added a commit to only check if a process died after having read all the logs in `wait_for_logs` to avoid a race condition.

This is a patch i wanted to upstream for a long time since i stole C-lightning test suite last year for Revault, and @danielabrozzoni patched it for us downstream (https://github.com/revault/revaultd/commit/c152273).

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4598/commits/f31403f217fb52200c4f5905ab524611a5a1db80

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4598/commits/f31403f217fb52200c4f5905ab524611a5a1db80

Issue #4597, opened 2021-06-14
Tor settings: default port is advertised despite specific bind-addr setting


openoms:
### Issue
`getinfo` should show the real port used to listen which is not the case if using a unique port specified with `bind-addr=127.0.0.1:9736` and not adding `\torport=9736` to `addr=statictor:127.0.0.1:9051`.

### Steps to Reproduce
Using a Tor-only node setting with a unique port on latest c-lightning master.

config:
```
proxy=127.0.0.1:9050
bind-addr=127.0.0.1:9736
addr=statictor:127.0.0.1:9051
always-use-proxy=true
```
### `getinfo` output
shows the Tor `"port": 9736`, but it is only connectable on 9735 (default) 

(tested with `connect` and `torify telnet .onion PORT`)

```
$ cl getinfo
{
   "id": "03b344deff71fd852ec8978e4777de06a516c6a09af93d02e94794d659f3d0e43c",
   "alias": "VIOLETAUTO",
   "color": "03b344",
   "num_peers": 0,
   "num_pending_channels": 0,
   "num_active_channels": 0,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 9736
      },
      {
         "type": "torv3",
         "address": "vct2pziq53ym7reum4bu7jr4bfgvpferkknwxorai67sp4o7mpce6gad.onion",
         "port": 9736
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 9736
      }
   ],
   "version": "v0.10.0-324-g3aafe8c",
   "blockheight": 687525,
   "network": "bitcoin",
   "msatoshi_fees_collected": 0,
   "fees_collected_msat": "0msat",
   "lightning-dir": "/home/bitcoin/.lightning/bitcoin"
}
```

When the `addr=statictor:127.0.0.1:9051\torport=9736` is specified in the config the 
```
  "address": "vct2pziq53ym7reum4bu7jr4bfgvpferkknwxorai67sp4o7mpce6gad.onion",
         "port": 9736
 ```
 becomes connectable.
 
 The confusing thing is that `getinfo` show the `"port": 9736` even if in reality it is using the default port for the .onion address.
 
 Reference in the docs: https://github.com/ElementsProject/lightning/blob/master/doc/lightningd-config.5.md#networking-options

rustyrussell:
[...]

PR #4603, opened 2021-06-16
Autotor torport fix

rustyrussell:
We were accidentally using the port that the tor service was connecting to, not the /torport the user said to use.

Fixes: #4597


Saibato:
hmm... would that not also apply to autotor? I wonder what i reasoned about to use local->port and not port for static and blobs and autotor. Maybe some undesired privacy side effects?
I guess to demand defining the advertised and used port with /torport or to more precise the doc might also fix this.
Tor can be configured external bind to any port or the default and use only internal the local.
It should have been possible in the issue case to reach the node over 9735, obliviously we should not announce the intern port, hmm...

vincenzopalazzo (approved):
LGTM.

ack https://github.com/ElementsProject/lightning/pull/4603/commits/5943319e0e47f6d467c24f49487c9d1a4ccb68ee

rustyrussell:
> hmm... would that not also apply to autotor? I wonder what i reasoned about to use local->port and not port for static and blobs and autotor. Maybe some undesired privacy side effects?

Yes, this fix covers both.

> I guess to demand defining the advertised and used port with /torport or to more precise the doc might also fix this.
> Tor can be configured external bind to any port or the default and use only internal the local.
> It should have been possible in the issue case to reach the node over 9735, obliviously we should not announce the intern port, hmm...

Yes, /torport was what we told Tor to listen on, but we didn't actually advertize that!  So this is clearly wrong.

PR #4581, opened 2021-06-03
pay: Do not attempt to filter routehints if we can't find ourselves

cdecker:
There are a couple of ways this can happen, but we shouldn't crash.

Fixed #4488
Fixes #4533

PR #4608, opened 2021-06-17
pay: Fix another instance of a crash when we can't find ourselves

cdecker:
Found another instance of the same issue that we addressed in #4581.

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4608/commits/4dd7a2fbba7c61153907cbd75e8751a3e5561c5c

PR #4610, opened 2021-06-18
experimental: Add extra-tlv support to keysend and cli option to allow extra TLV types

cdecker:
This is an experimental change to the way we handle unknown TLVs in the main daemon and the `keysend` plugin:

 - We add a command line flag that can be used to allow additional even TLV types to be accepted by the main daemon despite it not being able to handle them internally.
 - Based on this allowlist we then add the extra TLVs that were received by with the first HTLC in a payment (this being keysends we will most likely only ever have one total HTLC anyway) to the `invoice_payment` hook. This allows plugins to extract and react to the extra TLV fields that were not handled internally yet.
 - We also extend the `keysend` plugin to inject these extra TLV fields into the destination TLV, representing the write path of this change
 - Finally we also change the way that `keysend` reacts to unknown even TLV types: previously it'd throw up its arms and give up, now it does its preimage extraction and invoice creation logic, then hands off to other plugins. This allows us to incrementally build things like messaging without burdening each plugin to implement this logic over and over again. More in the way of a chain of responsibility rather than simple fail-over

All of this is still guarded behind `--experimental-features` flag for now since it changes a bit of the logic, but it should only ever impact the `noise` plugin which was doing this itself.

I kept the two fixups unmerged, since they are just the gen-files caused by the previous commit, to facilitate reviews.


lightningd/invoice.c
```diff
@@ -167,6 +168,49 @@ struct invoice_payment_hook_payload {
...
+							 field->numtype));
+			/* Finally add this element. */
+			json_add_hex_talarr(stream, NULL, field->value);
+			prevtype = field->numtype;
+		}
+	}
+	/* If we opened an array at all, now is the time to close it. */
+	if (prevtype != 0)
+		json_array_end(stream);
+	json_object_end(stream);
```
rustyrussell, line 210: BTW, JSON schemas has taught me that variable member names are an anti-pattern.  So make extra_tlvs an array of objects, with each object "type": u64, "data": hex?  Code will be simpler too...

plugins/keysend.c
```diff
@@ -150,6 +163,9 @@ static struct command_result *json_keysend(struct command *cmd, const char *buf,
 		   p_opt_def("exemptfee", param_msat, &exemptfee, AMOUNT_MSAT(5000)),
 #if DEVELOPER
 		   p_opt_def("use_shadow", param_bool, &use_shadow, true),
+#endif
+#if EXPERIMENTAL_FEATURES
+		   p_opt("extra-tlvs", param_extra_tlvs, &extra_fields),
```
rustyrussell, line 173: We usually use _ not - in JSON, or omit it altogether?

rustyrussell (approved):
Minor feedback only: mainly that this is 'extra-onion-tlvs', and some JSON formatting.  The idea is fine, AFAICT.

lightningd/invoice.c
cdecker: The main issue here is that the spec technically allows a field to be present multiple times, hence the TLV-type-to-array-of-values indirection. More than happy not to support this :-)

rustyrussell: Sure, then you just have duplicate fields:

```
[ {"type": 8, "data": "001122"}, {"type": 8, "data": "00112233"} ]
```

rustyrussell: Actually, they *can* but it's invalid:

```
 - MUST order `tlv_record`s in a `tlv_stream` by strictly-increasing `type`,
   hence MUST not produce more than a single TLV record with the same `type`
```

And:

```
 - if decoded `type`s are not strictly-increasing (including situations when
   two or more occurrences of the same `type` are met):
   - MUST fail to parse the `tlv_stream`.
```

cdecker: Right, I was going by the validation code, should have checked the spec itself :-)

cdecker: Will amend :+1:

plugins/keysend.c
cdecker: Removed altogether since I don't like the underscores at all, especially when using the command line.

cdecker:
Ok, addressed the comments by removing the dash in the option and by making the dict-of-lists into a list-of-dicts which I agree is much more flexible and easier to code as well.

See the range-diff [here](http://46.101.246.115:8888/ElementsProject/lightning/pull/4610/range_diff/0c407ab4bda706ef97be7c814daaf96ad64ff874..e0d702e182ca35b1076a85e49d82b77b8b1982ec)

rustyrussell:
Not sure this should be tagged for v10.0.1 since it's a bugfix, but pinging @niftynei now?

rustyrussell (approved):
Ack 0cabe48bf3dac34d8cdd9e792cbdde2a72d3eda8

rustyrussell:
Trivial rebase to fix conflict.

Ack 984776b4ab27a2c6e5a50cf74edca31c9538b467

PR #4611, opened 2021-06-18
keysend: Add routehints argument to the keysend command

cdecker:
So far we could not reach non-publicly reachable nodes with the
keysend command since we couldn't compute a route to them. With this
change we can add some routehints that we may have gotten from an
address book or a previous invoice, and make it more likely that we
can reach the destination.

Builds on top of #4610 to avoid rebase conflicts, but I can separate them out if that's preferred. This PR consists only of the last 3 commits.

### DIFF
--- a/common/json_tok.c
+++ b/common/json_tok.c
@@ -584,3 +584,73 @@ struct command_result *param_extra_tlvs(struct command *cmd, const char *name,
 	*fields = temp;
 	return NULL;
 }
+
+struct command_result *param_routehint(struct command *cmd, const char *name,
+				       const char *buffer, const jsmntok_t *tok,
+				       struct route_info **ri)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	const char *err;
+
+	if (tok->type != JSMN_ARRAY) {
+		return command_fail(
+		    cmd, JSONRPC2_INVALID_PARAMS,
+		    "Routehint %s (\"%s\") is not an array of hop objects",
+		    name, json_strdup(tmpctx, buffer, tok));
+	}
+
+	*ri = tal_arr(cmd, struct route_info, tok->size);
+	json_for_each_arr(i, curr, tok) {
+		struct route_info *e = &(*ri)[i];
+		struct amount_msat temp;
+
+		err = json_scan(tmpctx, buffer, curr,
+				"{id:%,scid:%,feebase:%,feeprop:%,expirydelta:%}",
+				JSON_SCAN(json_to_node_id, &e->pubkey),
+				JSON_SCAN(json_to_short_channel_id, &e->short_channel_id),
+				JSON_SCAN(json_to_msat, &temp),
+				JSON_SCAN(json_to_u32, &e->fee_proportional_millionths),
+				JSON_SCAN(json_to_u16, &e->cltv_expiry_delta)
+			);
+		e->fee_base_msat =
+		    temp.millisatoshis; /* Raw: internal conversion. */
+		if (err != NULL) {
+			return command_fail(
+			    cmd, JSONRPC2_INVALID_PARAMS,
+			    "Error parsing routehint %s[%zu]: %s", name, i,
+			    err);
+		}
+	}
+	return NULL;
+}
+
+struct command_result *
+param_routehint_array(struct command *cmd, const char *name, const char *buffer,
+		      const jsmntok_t *tok, struct route_info ***ris)
+{
+	size_t i;
+	const jsmntok_t *curr;
+	char *element_name;
+	struct command_result *err;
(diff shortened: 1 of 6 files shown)


common/json_tok.c
```diff
@@ -584,3 +584,76 @@ struct command_result *param_extra_tlvs(struct command *cmd, const char *name,
...
+		    cmd, JSONRPC2_INVALID_PARAMS,
+		    "Routehint %s (\"%s\") is not an array of hop objects",
+		    name, json_strdup(tmpctx, buffer, tok));
+	}
+
+	*ri = tal_arr(cmd, struct route_info, 0);
+	json_for_each_arr(i, curr, tok) {
+		struct route_info e;
+		struct amount_msat temp;
+		element_name = tal_fmt(cmd, "%s[%zu]", name, i);
```
rustyrussell, line 608: Creating this var here seems weird?  It's only used once, and inside a format string?

common/json_tok.c
```diff
@@ -584,3 +584,76 @@ struct command_result *param_extra_tlvs(struct command *cmd, const char *name,
...
+	char *element_name;
+	struct command_result *err;
+	if (tok->type != JSMN_ARRAY) {
+		return command_fail(
+		    cmd, JSONRPC2_INVALID_PARAMS,
+		    "Routehint array %s (\"%s\") is not an array",
+		    name, json_strdup(tmpctx, buffer, tok));
+	}
+
+	*ris = tal_arr(cmd, struct route_info *, 0);
```
rustyrussell, line 646: Here too, tok->size?

plugins/keysend.c
```diff
@@ -128,11 +128,12 @@ static const char *init(struct plugin *p, const char *buf UNUSED,
 	return NULL;
 }
 
-struct payment_modifier *pay_mods[8] = {
+struct payment_modifier *pay_mods[9] = {
```
rustyrussell, line 131: Maybe just change this to `[]`?

plugins/keysend.c
```diff
@@ -183,7 +186,7 @@ static struct command_result *json_keysend(struct command *cmd, const char *buf,
 	p->destination_has_tlv = true;
 	p->payment_secret = NULL;
 	p->amount = *msat;
-	p->routes = NULL;
+	p->routes = hints;
```
rustyrussell, line 189: Do we need to tal_steal() this onto p?

tests/test_pay.py
```diff
@@ -3108,13 +3108,64 @@ def test_keysend_extra_tlvs(node_factory):
...
 
 
+def test_keysend_routehint(node_factory):
+    """Test whether we can deliver a keysend by adding a routehint on the cli
+    """
+    amt = 10000
+    l1, l2 = node_factory.line_graph(
+        2,
+        wait_for_announce=True,
+        opts=[{}, {}]
```
rustyrussell, line 3124: That opts seems redundant?

tests/test_pay.py
```diff
@@ -3108,13 +3108,64 @@ def test_keysend_extra_tlvs(node_factory):
...
+    with pytest.raises(RpcError):
+        l1.rpc.call("keysend", payload={'destination': dest,'msatoshi': amt, 'routehints': routehints[1:]})
+
+    res = l1.rpc.call("keysend", payload={'destination': dest,'msatoshi': amt,'routehints': routehints})
+    invs = l3.rpc.listinvoices()['invoices']
+    assert(len(invs) == 1)
+
+    inv = invs[0]
+    assert(inv['msatoshi_received'] >= amt)
+    import pdb;pdb.set_trace()
```
rustyrussell: Guessing that's removed soon?

rustyrussell (approved):
Ack 553fd6190a1bd3016184d32eb61fdedf7f164c21

Trivial changes only.

Have no firm opinion on rebase or not, that's @niftynei ....

common/json_tok.c
cdecker: Copy pasta from the outer loop, will remove :+1:

common/json_tok.c
cdecker: I tried this but there are just too many levels of indirection and my head started hurting: we're passing a pointer of position in an array, which should then be initialized in the callee, and is itself a variable length array. What would the canonical way of doing this be? And is there a trick to make this clearer in the code (parentheses, ...)?

rustyrussell:
Ack: https://github.com/ElementsProject/lightning/pull/4611/commits/b3b3f538c81738839f8e01737d83bd91a271812a

niftynei:
ACK 4b42c3c9608ee8b01ef3122c3aee5b3df0c5cdb2

PR #4615, opened 2021-06-24
Pyln fixes and misc cleanups

rustyrussell:
This pairs well with the lnprototest updates for testing channel_types:  https://github.com/rustyrussell/lnprototest/pull/15

Changelog-None

PR #4616, opened 2021-06-24
channel type support

rustyrussell:
Based on #4615


rustyrussell:
Trivial rebase now #4615 is merged

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4616/commits/2c138ea454a31a0a92de80937339dddc59591f14

I think we need to add only the feature `supports_open_accept_channel_types` in the python test.

rustyrussell:
This is updated to match the new spec, finally!

t-bast:
I've tested https://github.com/ElementsProject/lightning/pull/4616/commits/e34036452311ec5cc1cca00d6876a4e25ac4230b against eclair (https://github.com/ACINQ/eclair/pull/1867) and I've found some issues (sorry!).

I have activated anchor outputs in both c-lightning and eclair to be able to test a wide range of cases.
When either eclair or c-lightning proposes the anchor outputs channel type, everything works well.
However, when eclair proposes a different channel type, it doesn't work:

- If eclair proposes only `static_remotekey` (feature bit 12), c-lightning rejects the `open_channel` and says that this is an unknown channel type
- If eclair proposes a standard channel (empty feature bytes, but `channel_type` tlv is present), c-lightning acts as if the tlv was missing and sends an `accept_channel` with the anchor outputs channel type, which doesn't match the channel type in `open_channel`.

t-bast:
@rustyrussell do you think you will have time this week to fix the issues I reported earlier? It would be great to finalize compatibility testing and be able to merge https://github.com/lightningnetwork/lightning-rfc/pull/880 during the next spec meeting!

rustyrussell:
>     * If eclair proposes only `static_remotekey` (feature bit 12), c-lightning rejects the `open_channel` and says that this is an unknown channel type

This is now fixed.  Initially we only allowed the default type, now we allow downgrades.

>     * If eclair proposes a standard channel (empty feature bytes, but `channel_type` tlv is present), c-lightning acts as if the tlv was missing and sends an `accept_channel` with the anchor outputs channel type, which doesn't match the channel type in `open_channel`.

This is also fixed: we correctly respond with the empty channel (at least for now: we may required static_remotekey soon).

channeld/full_channel.c
```diff
@@ -100,15 +101,14 @@ struct channel *new_full_channel(const tal_t *ctx,
 				 u32 lease_expiry,
 				 struct amount_sat funding,
 				 struct amount_msat local_msat,
-				 const struct fee_states *fee_states,
+				 const struct fee_states *fee_states TAKES,
```
vincenzopalazzo, line 104: small question related to my knowledge, what it means this `TAKES` here?

rustyrussell: It means that the caller can use the take() marker to indicate that the fee_states parameter should be owned by the new_full_channel call.

Normally the caller would have to free it, but if it passes take(fee_states) it must not, since the call will take ownership.

We use this pattern in several places, to avoid copying.

There's another marker, STEALS which means the call will *always* take ownership.

lightningd/opening_control.c
```diff
@@ -305,12 +303,12 @@ static void opening_funder_start_replied(struct subd *openingd, const u8 *resp,
...
 {
-	u8 *funding_scriptPubkey;
 	bool supports_shutdown_script;
 
-	if (!fromwire_openingd_funder_start_reply(resp, resp,
-						 &funding_scriptPubkey,
-						 &supports_shutdown_script)) {
+	if (!fromwire_openingd_funder_start_reply(fc, resp,
+						  &fc->funding_scriptpubkey,
+						  &supports_shutdown_script,
```
cdecker, line 310: I take it that the upfront shutdown script is negotiated independently of the channel_type? Or could it be removed and passed via the channel_type instead?

cdecker (approved):
Very nice cleanup, I'm starting to like the `channel_type` proposal after all :wink: 

ACK 707164fb700845a5b8351aeee9d70b17bf826007

Issue #4495, opened 2021-04-28
Crash


reingit:
Didn't notice till much later it crashed but was most likely editing my /etc/fstab at the time it happend:

```
2021-04-27T12:17:57.582Z **BROKEN** lightningd: /home/bitcoin/lightning/bin/../libexec/c-lightning/plugins/bcli error: bad response to estimatefees (bad 'result' field), response was {"jsonrpc":"2.0","id":80213,"error":{"code":400,"message":"/snap/bin/bitcoin-core.cli estimatesmartfee 2 CONSERVATIVE: bad JSON: cannot parse (2021/04/27 14:17:57.222526 system_key.go:129: cannot determine nfs usage in generateSystemKey: cannot parse /etc/fstab: expected between 3 and 6 fields, found 1\n{\n  \"feerate\": 0.00180929,\n  \"blocks\": 2\n}\n)"} }
2021-04-27T12:17:57.662Z **BROKEN** lightningd: FATAL SIGNAL 6 (version v0.10.0-27-gbbfcae6)
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: common/daemon.c:44 (send_backtrace) 0x55fb0d2a2668
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: common/daemon.c:52 (crashdump) 0x55fb0d2a26ba
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7fb9a290120f
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7fb9a290118b
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7fb9a28e0858
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: lightningd/log.c:803 (fatal) 0x55fb0d26c507
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: lightningd/bitcoind.c:120 (bitcoin_plugin_error) 0x55fb0d251481
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: lightningd/bitcoind.c:180 (estimatefees_callback) 0x55fb0d251507
2021-04-27T12:17:57.662Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:460 (plugin_response_handle) 0x55fb0d284969
[...]

PR #4620, opened 2021-06-29
plugins/bcli: don't try to parse stderr as JSON.

rustyrussell:
Just let it go through to the main stderr output.

Fixes: #4495
Changelog-Changed: plugins: bcli will not try to parse bitcoin-cli stderr.

### DIFF
--- a/plugins/bcli.c
+++ b/plugins/bcli.c
@@ -302,7 +302,7 @@ static void next_bcli(enum bitcoind_prio prio)
 	if (!bcli)
 		return;
 
-	bcli->pid = pipecmdarr(&in, &bcli->fd, &bcli->fd,
+	bcli->pid = pipecmdarr(&in, &bcli->fd, &pipecmd_preserve,
 			       cast_const2(char **, bcli->args));
 
 	if (bitcoind->rpcpass)


plugins/bcli.c
```diff
@@ -277,7 +277,7 @@ static void next_bcli(enum bitcoind_prio prio)
 	if (!bcli)
 		return;
 
-	bcli->pid = pipecmdarr(NULL, &bcli->fd, &bcli->fd,
+	bcli->pid = pipecmdarr(NULL, &bcli->fd, &pipecmd_preserve,
```
vincenzopalazzo, line 280: I'm not able to see what the `pipecmd_preserve` is in the Github UI, but from the Github actions, U assumes that we need to check also the value of `pipecmd_preserve` in case we will request a block height not in the blockchain yet?

vincenzopalazzo (commented):
LGTM, but I have a small curiosity about the new tmp variable.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4620/commits/c06439afcdb194e3c3b69b4fd6467b740301b69c

cdecker:
Rebase on top of `master` to fix GH Actions failures

cdecker:
This seems to make the breakage worse. According to the CI runs we're now mixing `bcli` errors with `bitcoin-cli` stderr much more, and any test that has a `bitcoin-cli` command fail will now fail too.

cdecker (changes requested):
Seems we now mix outputs even more.

PR #4622, opened 2021-06-29
fix docs

ndalliard:
changelog-none

### DIFF
--- a/doc/lightning-feerates.7
+++ b/doc/lightning-feerates.7
@@ -16,6 +16,7 @@ C-lightning will also smoothen feerate estimations from the backend\.
 
 \fIstyle\fR is either of the two strings:
 
+
 .RS
 .IP \[bu]
 \fIperkw\fR - provide feerate in units of satoshis per 1000 weight\.
@@ -25,6 +26,7 @@ C-lightning will also smoothen feerate estimations from the backend\.
 .RE
 
 Bitcoin transactions have non-witness and witness bytes:
+
 
 .RS
 .IP \[bu]
@@ -55,6 +57,7 @@ which will override the recommended feerates returned by \fBfeerates\fR\.
 .SH RETURN VALUE
 
 On success, an object is returned, containing:
+
 
 .RS
 .IP \[bu]
@@ -122,6 +125,7 @@ On success, an object is returned, containing:
 
 The following warnings may also be returned:
 
+
 .RS
 .IP \[bu]
 \fBwarning_missing_feerates\fR: Some fee estimates are missing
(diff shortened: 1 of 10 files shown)


doc/lightning-cli.1
```diff
@@ -127,7 +127,7 @@ pretty printing of results isn’t pretty\.
 
 .SH AUTHOR
 
-Rusty Russell \fI<rusty@rustcorp.com.au\fR> is mainly to blame\.
+Rusty Russell <\fIrusty@rustcorp.com.au\fR> is mainly to blame\.
```
vincenzopalazzo, line 130: mh the file without `.md` should be autogenerated from the `make doc-all` command, and here I'm missing the change in the `md `file

vincenzopalazzo (commented):
LGTM.

The only nit is the message as a comment, I'm missing the change the the `md` file.

P.S: The failure in the github action is not related to this PR.

ndalliard:
hmm, i just realised from your nit, that the manpage files are autogenerated in a wrong way

should be:
````man
<\fIrusty@rustcorp.com.au\fR>
````
generated:
````man
\fI<rusty@rustcorp.com.au\fR>
````

i will just squash the commits into one commit with the other fixes and leave the above as it is, it probably has something to do with how `mrkd` generates the manpages, but i am not familiar with it

vincenzopalazzo:
>i will just squash the commits into one commit with the other fixes and leave the above as it is, it probably has something to do with how mrkd generates the manpages, but i am not familiar with it

ack!

Yes, the pages are autogenerated by the python module `mrkd`.

I think if you swash your change in a unique commit it is good too

ndalliard:
are you okay with me adding the generated files into the `.gitignore`?

vincenzopalazzo:
>are you okay with me adding the generated files into the .gitignore?

I think is not a good choice because the command `make` doesn't call the `make doc-all`.

ndalliard:
> > are you okay with me adding the generated files into the .gitignore?
> 
> I think is not a good choice because the command `make` doesn't call the `make doc-all`.

okay, the `gmake doc-all` now generated a lot of empty lines, leave them like this or revert to how it was?

vincenzopalazzo:
>okay, the gmake doc-all now generated a lot of empty lines, leave them like this or revert to how it was?

I would like to have the previous status with your typo fix in the` lightning-config` (that is my fault when I made the doc page)

ndalliard:
updated - kept the other changes, the empty lines are strange, but if i remove them i guess the SHA256 wouldn't be correct

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4622/commits/f7bd80f6a5d4242b1a4f78618e9621fb5bbb02b4

rustyrussell:
We check in the generated files so that you don't need mrkd to build them, which is why they're not .gitignore'd.

Thanks for finding this!

PR #4630, opened 2021-07-03
A little more offers

rustyrussell:
First fix a crash where we decode a signatureless offer.  Doh!  Then add a developer-only option to override the fetchinvoice logic with a custom secret payer key.

Compat break with merkle (i.e. signature, offer_id) calculation


niftynei:
ACK 225a598d7656928a11a26cb9e92cbd8a92f7289e

PR #4638, opened 2021-07-09
Schema and other fixes

niftynei:
Seems like there's a few bugs in master with schemas; also the coin-moves test is off by 1695sats for the dual-funding tests.


rustyrussell (approved):
Ack c497a4a62fdd1b59a9ed6b23574e43c3bd449749

PR #4646, opened 2021-07-12
Make payment secret compulsory on new invoices, and assume unknown nodes support TLV onions

rustyrussell:
Most of this work is removing our assumptions that payment_secret is unnecessary (i.e. making us use modern TLV onion for the terminal everywhere!).

We also switch our assumptions about nodes for which we don't have a node_announcement: we assume they *do* support TLV onions.  With the next release, we might be able to remove legacy sending support altogether (and then finally, legacy receive/forward).


plugins/keysend.c
```diff
@@ -259,17 +261,61 @@ struct keysend_in {
...
+	tal_arr_expand(&ki->payload->fields, field);
+
+	asort(ki->payload->fields, tal_count(ki->payload->fields),
+	      tlvfield_cmp, NULL);
+	for (size_t i = 0; i < tal_count(ki->payload->fields); i++) {
+		plugin_log(cmd->plugin, LOG_DBG,
+			   "field[%zu] = %zu:%s",
+			   i, ki->payload->fields[i].numtype,
+			   tal_hex(tmpctx, ki->payload->fields[i].value));
+	}
```
cdecker, line 317: Better to add a line saying what we're printing here and indenting slightly, otherwise we have random values appearing in the logs without any context.

cdecker (commented):
Making things compulsory always feels dangerous but in this case it looks ok

ACK https://github.com/ElementsProject/lightning/pull/4646/commits/0b885f8e70699244cd20e916abed997ac4e97993

cdecker (approved):

plugins/keysend.c
rustyrussell: Oops, debug!

rustyrussell: Removed, thanks!

niftynei:
reapplying ACK https://github.com/ElementsProject/lightning/pull/4646/commits/0433cec3038a9ee43536840cd8baf64141c18ce5

PR #4648, opened 2021-07-12
dual-fund: no feestep

niftynei:
The `fee_step` portion of the spec was cute but not really servicable, especially given that there's no guarantee that bumping the fee by 25% will put you over the required min_relay fee bump.

Instead, we just let the opener specify a new feerate, and require that it's at least 1/64th more than the last "successful" one (that means we've swapped tx_sigs for it successfully, it might still fail to broadcast).

Includes a fix for estimating the additional fee that a change output adds to the tx; we weren't considering the total weight which means it was possible to get off-by-one errors on the rounding.

This is a breaking change with prior dual-funding implementations.

PR #4652, opened 2021-07-14
Update lnprototest to latest

vincenzopalazzo:
With this PR that updates the latest version of lnprototest the  #4648 needs just a rebase on top of the new master.

Depending on #4648


niftynei:
This change has to be folded into the same commit that introduces the new fee_step pattern, otherwise the tests don't pass. I've pulled this into the commit that introduces the change in #4648. Thanks!

PR #4657, opened 2021-07-18
added minimum word list needed

LightningHelper:
Not every user know the minimum word list that is needed.
Below 12 words the output is:
`lightning-hsmtool: Invalid mnemonic:`


rustyrussell (approved):
Ack 36befa5d63d23c31e55b11f75c02bed5219cdf78

Thanks, good catch!

PR #4662, opened 2021-07-19
Github actions: reduce lnprototest.

rustyrussell:
Setting VALGRIND=1 actually does nothing here; reduce it to two cases,
covering gcc and clang, sqlite3 and postgres.


niftynei:
ACK 7e1159fc25a76ba00e72ea7d1e4d9ec003b5dc5f

PR #4663, opened 2021-07-20
pytest: fix bad gossip flake.

rustyrussell:
If l3 is too slow, it can get channel_announcement after channel
is closed, so it gets upset at the node_announcement which follows:

   022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59-gossipd: Updated pending announce with update 103x1x1/1
   022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59-gossipd: channel_announcement: no unspent txout 103x1x1
   022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59-gossipd: Bad gossip order: WIRE_NODE_ANNOUNCEMENT before announcement 0266e4598d1d3c415f572a8488830b60f7e744ed9235eb0b1ba93283b315c03518

### DIFF
--- a/tests/test_closing.py
+++ b/tests/test_closing.py
@@ -1657,13 +1657,18 @@ def test_onchain_middleman_their_unilateral_in(node_factory, bitcoind):
     l2.rpc.connect(l1.info['id'], 'localhost', l1.port)
     l2.rpc.connect(l3.info['id'], 'localhost', l3.port)
 
-    l2.fundchannel(l1, 10**6)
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4663/commits/9a6a401e29dd45022dd43af91e86ec4c1c65ffd9

I think that there is some stuck process in the wait status, but I'm not able to understand what it is the problem.

PR #4668, opened 2021-07-20
Fixed order parameter in the listforwards command

vincenzopalazzo:
From a report in IRC channel, I noted that the order of the parameters are wrong in the rpc command.

I assumed that the code take the wrong order because for us (programmers) write documentation is costly more than code :-)

In addition, I think is more natural to have as the first parameter the status, in my opinion, it is more frequently have a filter by the status instant of the filter by channel id.

In conclusion, I can revert the change and modify the docs, or if this change is good, I'm happy to know if the place where I wrote the test is a correct place (test_peers.py or test_rpc.py file is missed).

### DIFF
--- a/tests/test_gossip.py
+++ b/tests/test_gossip.py
@@ -1975,7 +1975,14 @@ def test_parms_listforwards(node_factory):
     it is simple and not useful, but it is good to have to avoid
     simile errors in the future.
     """
-    l1, _ = node_factory.line_graph(2)
+    l1, l2 = node_factory.line_graph(2)
 
-    forwards = l1.rpc.listforwards("settled")["forwards"]
-    assert len(forwards) == 0
+    l2.stop()
+    l2.daemon.opts['allow-deprecated-apis'] = True
+    l2.start()
+
+    forwards_new = l1.rpc.listforwards("settled")["forwards"]
+    forwards_dep = l2.rpc.call("listforwards", {"in_channel": "0x1x2", "out_channel": "0x2x3", "status": "settled"})["forwards"]
+
+    assert len(forwards_new) == 0
+    assert len(forwards_dep) == 0
--- a/tests/test_gossip.py
+++ b/tests/test_gossip.py
@@ -1964,3 +1964,18 @@ def test_addgossip(node_factory):
 
     with pytest.raises(RpcError, match='Bad signature'):
         l3.rpc.addgossip(badupdate)
+
+
+def test_parms_listforwards(node_factory):
+    """
+    Simple test to ensure that the order of the listforwards
+    is correct as describe in the documentation.
+
+    This test is written by a issue report in the IR channel,
+    it is simple and not useful, but it is good to have to avoid
+    simile errors in the future.
+    """
+    l1, _ = node_factory.line_graph(2)
+
+    forwards = l1.rpc.listforwards("settled")["forwards"]
+    assert len(forwards) == 0
--- a/lightningd/peer_htlcs.c
+++ b/lightningd/peer_htlcs.c
@@ -2661,13 +2661,30 @@ static struct command_result *json_listforwards(struct command *cmd,
 	const char *status_str;
 	enum forward_status status = FORWARD_ANY;
 
+	// TODO: We will remove soon after the deprecated period.
+	if (params && deprecated_apis && params->type == JSMN_ARRAY) {
+		struct short_channel_id scid;
+		/* We need to catch [ null, null, "settled" ], and
+		 * [ "1x2x3" ] as old-style */
+	        if ((params->size > 0 && json_to_short_channel_id(buffer, params + 1, &scid)) ||
+		    (params->size == 3 && !json_to_short_channel_id(buffer, params + 3, &scid))) {
+			if (!param(cmd, buffer, params,
+				   p_opt("in_channel", param_short_channel_id, &chan_in),
+				   p_opt("out_channel", param_short_channel_id, &chan_out),
+				   p_opt("status", param_string, &status_str),
+				   NULL))
+				return command_param_failed();
+			goto parsed;
+		}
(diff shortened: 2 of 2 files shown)


rustyrussell:
I think we need to fix the documentation, not the code unfortunately, as this was already released in 0.10.0 so people may rely on it.

vincenzopalazzo:
@rustyrussell I will open a new PR with the doc fixing :) thanks

vincenzopalazzo:
Applied the suggestion of @rustyrussell and test it locally. However, I did not find a method to test the deprecated API function in the python test.

tests/test_gossip.py
```diff
@@ -1964,3 +1964,18 @@ def test_addgossip(node_factory):
...
+def test_parms_listforwards(node_factory):
+    """
+    Simple test to ensure that the order of the listforwards
+    is correct as describe in the documentation.
+
+    This test is written by a issue report in the IR channel,
+    it is simple and not useful, but it is good to have to avoid
+    simile errors in the future.
+    """
+    l1, _ = node_factory.line_graph(2)
```
rustyrussell, line 1978: Just do node_factory.get_node(options={'allow-deprecated-apis': True}) ?

rustyrussell: ANd even better would be to add it to test_listforwards rather than having another test.

You could even restart l2 in that test with deprecated-apis enabled, to test?

    l2.stop()
    l2.daemon.opts['allow-deprecated-apis'] = True
    l2.start()

vincenzopalazzo: Thanks, rusty to point me out this detail. I was thinking that maybe it is better to write the test for the RPC interface in a new file test_rpc? In this way, we can track also the deprecated API and the test could fail in the deprecated period it is over.

What do you think?

In this case, do we need to have some convention to remove the code that it is over?

vincenzopalazzo: Now should be fixed! Thanks @rustyrussell

cdecker:
Wasn't the resolution for this to update the docs instead of reordering the arguments in code? As is this is pretty much a breaking change as @rustyrussell pointed out.

vincenzopalazzo:
Hi @cdecker, Yes you have right, but we discussed with @rustyrussell on IRC https://gnusha.org/c-lightning/2021-07-20.log and this change make the API of this command more user friendly. All the code inside this PR is made by rusty suggestion.

cdecker:
Ok, sounds good. Will merge it then.

cdecker:
ACK e560b69786aa5f88df807d5731fae38041c54381

cdecker:
Rebased on top of `master` to ensure we havent accidentally broken things.

cdecker:
Had to move the autodata declaration to avoid having a linenumber conflict...

cdecker:
This is the IRC log extract for posterity:

```
11:01 < nathanael> hmm `lightning-cli listforwards settled` should work imo - synopsis according to the manpage: listforwards [status] [in_channel] [out_channel] - i get an error saying it needs a short_channel_id
12:39 < vincenzopalazzo> <nathanael "hmm `lightning-cli listforwards "> I think that the order of the parameter is wrong, or maybe the description? However, this PR should be contains a solution https://github.com/ElementsProject/lightning/pull/4668
13:09 -!- rich- is now known as yakshaver_
13:11 -!- yakshaver_ is now known as yakshaver
14:32 < rusty> nathanael: nice catch, yes, doc is wrong.  Unf it was in previous release so changing order to match docs is a breaking change (vs. changing docs).  Commented on PR vincenzopalazzo
14:33 < vincenzopalazzo> rusty: Ops, I will close the PR and open a new one. i missed to check the changelog, sorry
14:34 < rusty> vincenzopalazzo: I used git blame to get commit 8a8f81175d, then git describe --contains 8a8f81175d returned v0.10.0rc1~166
14:35 < rusty> I mean, we *could* change order, but it's messy :(  I agree that status first is probably nicest.  lightning-cli listforwards status=settled is less nice.
14:38 < vincenzopalazzo> rusty: Yes agree to have the status before, but if something is already in production is better to leave it as is. angry people are bad for API change are bad :)
14:39 < rusty> We have a 2-release / 6 month window for deprecating old APIs.  And we can detect this (status and short-channel-id are easy to distinguish).  But it's a bit messy.  nathanael, vincenzopalazzo: thoughts?
14:40 < rusty> (API changes are painful in the short term, but bad APIs are painful forever!)
14:41 < vincenzopalazzo> rusty: the first think that this quotes need to be in a book (API changes are painful in the short term, but bad APIs are painful forever!)
14:41 < vincenzopalazzo> for the api, I think we can open an issue and change it later?
14:42 < vincenzopalazzo> in another major version? like 0.11.0
14:42 < rusty> Well, if we change it right now, it's only been that way for one release.
14:42 < rusty> (And that means *today*, since we are about to release rc1!)
14:42 < rusty> niftynei: ?
14:43 < vincenzopalazzo> this is also true
14:44 < rusty> Let me quickly hack something together for review, while we think if this is worthwhile...
14:45 < vincenzopalazzo> So, I will reopen the PR
15:09 < rusty> vincenzopalazzo: untested, but something like this? https://0bin.net/paste/aEXqITFF#t1vJmCnoeWKneWjpU6GbLswFBWqFooRZ5ffYftJC47F
15:09 < rusty> (gtg, my kids are waking up!)
15:17 < vincenzopalazzo> <rusty "vincenzopalazzo: untested, but s"> ack, to deprecate the API and move on with a API change. i you want I can work on your draft and integrate it in the PR?
16:26 < rusty> vincenzopalazzo: please do!
16:27 < rusty> (Test might be nice, ofc! :)
```

rustyrussell (approved):
Ack 791a4ea671790beb4a5742132ff0843beaecf70e

PR #4674, opened 2021-07-22
Datastore

### LINKED ISSUE #4525: Brainstorming: Storing plugin metadata in the database
This came up while debugging #4482, where we have the issue that we
try to reconstruct more context, i.e., group of `sendpay` commands
that were initiated by the same `pay` command, but don't really have
enough information to do so. In #4482's case we group by the
`payment_hash` which is too aggressive in grouping `sendpay`s into the
same group if `pay` was called multiple times with the same invoice /
`payment_hash`, resulting in a potential flip-flop of the state
displayed in the `listpays` output. The solution here is to add an
explicit grouping key that is used to group only the ones together
that are part of the same group, but we faced this issue a number of
times: plugins wanting to store some metadata, in otherwise unrelated
contexts. As an example, `sendpay` has no concept of an invoice, it
just deals with routes, HTLCs and their status, the invoice we
shoehorn into the first `sendpay` of a `pay` is there just because we
don't have a better place for it.

I'd like to propose that we add a new table `metadata` with just two
columns: `id BIGINT AUTOINCREMENT` and `data BLOB`. We then two new
RPC methods `createmetadata` and `getmetadata` which can be used to
create a new entry in the table, and get the entry based on its
`id`. Then from various tables such as the `payments` table we have a
nullable reference to the `metadata` table allowing us to attach extra
data to them if desired. `pay` could then `createmetadata` for each
invocation, storing `invoice` and other metadata in the table, it'd
get back the new row's `id`, which is then passed to all `sendpay`
invocations. `listpay` can then collect and group the `listsendpays`
result, based on the `metadata_id` not on the `payment_hash` which
we'll lose as grouping key once we get PTLCs anyway. Furthermore it
can enrich its own result with the metadata stored by `pay` in a
well-defined location.

To begin with the metadata is immutable, and can only be created.

rustyrussell:
Closes #4525

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -2624,3 +2624,56 @@ def test_force_feerates(node_factory):
         "penalty": 6666,
         "min_acceptable": 1875,
         "max_acceptable": 150000}
+
+
+def test_datastore(node_factory):
+    l1 = node_factory.get_node()
+
+    # Starts empty
+    assert l1.rpc.listdatastore() == {'datastore': []}
+    assert l1.rpc.listdatastore('somekey') == {'datastore': []}
+
+    # Add entries.
+    somedata = b'somedata'.hex()
+    assert l1.rpc.datastore('somekey', somedata) == {'key': 'somekey',
+                                                     'hex': somedata}
+
+    assert l1.rpc.listdatastore() == {'datastore': [{'key': 'somekey',
+                                                     'hex': somedata}]}
+    assert l1.rpc.listdatastore('somekey') == {'datastore': [{'key': 'somekey',
+                                                              'hex': somedata}]}
+    assert l1.rpc.listdatastore('otherkey') == {'datastore': []}
+
+    otherdata = b'otherdata'.hex()
+    assert l1.rpc.datastore('otherkey', otherdata) == {'key': 'otherkey',
+                                                       'hex': otherdata}
+
+    assert l1.rpc.listdatastore('somekey') == {'datastore': [{'key': 'somekey',
+                                                              'hex': somedata}]}
+    assert l1.rpc.listdatastore('otherkey') == {'datastore': [{'key': 'otherkey',
+                                                              'hex': otherdata}]}
+    assert l1.rpc.listdatastore('badkey') == {'datastore': []}
+
+    ds = l1.rpc.listdatastore()
+    # Order is undefined!
+    assert (ds == {'datastore': [{'key': 'somekey', 'hex': somedata},
+                                 {'key': 'otherkey', 'hex': otherdata}]}
+            or ds == {'datastore': [{'key': 'otherkey', 'hex': otherdata},
+                                    {'key': 'somekey', 'hex': somedata}]})
+
+    assert l1.rpc.deldatastore('somekey') == {'key': 'somekey',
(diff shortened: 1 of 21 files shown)


tests/test_misc.py
```diff
@@ -2623,3 +2623,56 @@ def test_force_feerates(node_factory):
...
+    
+    assert l1.rpc.listdatastore('somekey') == {'datastore': [{'key': 'somekey',
+                                                              'hex': somedata}]}
+    assert l1.rpc.listdatastore('otherkey') == {'datastore': [{'key': 'otherkey',
+                                                              'hex': otherdata}]}
+    assert l1.rpc.listdatastore('badkey') == {'datastore': []}
+
+    l = l1.rpc.listdatastore()
+    # Order is undefined!
+    assert (l == {'datastore': [{'key': 'somekey', 'hex': somedata},
```
vincenzopalazzo, line 2658: I think this is the place where the code check will fail. 

I a little bit difficult to guess what is the correct format of the `{}` from the error message

wallet/wallet.c
```diff
@@ -4706,3 +4706,92 @@ void wallet_offer_mark_used(struct db *db, const struct sha256 *offer_id)
...
+	stmt = db_prepare_v2(w->db, SQL("SELECT 1"
+					"  FROM datastore"
+					" WHERE key = ?;"));
+	db_bind_text(stmt, 0, key);
+	db_query_prepared(stmt);
+
+	if (db_step(stmt)) {
+		tal_free(stmt);
+		return false;
+	}
```
vincenzopalazzo, line 4724: I don't know if I'm missing some things in the design of this functionality, but if a plugin needs to override functionality is it now possible with the current API of the wallet.h, right?

Do you think that is necessary use some additional parameters to the RPC command to give the possibility to override the propriety with key?

Somethings like `lightning-cli datastore "fooo" "some" true`

vincenzopalazzo (commented):
LGTM, I have only a few comments

wallet/wallet.c
rustyrussell: Yes, since this isn't going to be in this release, I can revise this properly.

I am going to allow both hex and string arguments, and also allow a mode argument:
- "must-create", "must-replace", or "create-or-replace", "must-append", "create-or-append".

vincenzopalazzo: I like this idea Rusty.

rustyrussell:
See also https://github.com/lightningd/plugins/pull/275 which provides this same functionality for previous versions!

tests/test_misc.py
rustyrussell: Yes, but it's always better to be *explicit* in tests, than loose.  And it works :)

vincenzopalazzo (approved):
ack https://github.com/rustyrussell/plugins/commit/58eb86880f3f03af749cb3c1bf160b2edc3fe636

rustyrussell:
Updated to add generation field, on suggestion from @shesek !

rustyrussell:
Trivial rebase to fix flake.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/22da787e7724081eba82f034ec4c0ffc58db4400

PR #4676, opened 2021-07-23
pyln-client: document and test that init can self-disable.

rustyrussell:
mypy says it returns None, but it actually doesn't have to!

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4676/commits/d7cf0555f18606414893117f29c70d72535f2a39

cdecker:
Amended the `mypy` signature to reflect that we may return a `dict` to self-disable.

ACK https://github.com/ElementsProject/lightning/pull/4676/commits/8b397ce9b88caaf3e74969a01cbfea26f2c5729b

PR #4678, opened 2021-07-23
increase timeout in the wait_for_log method in the test that fails

vincenzopalazzo:
In several PR I noted that the wait_for_log creates a timeout error, and this PR tries to achieve a possible fixed point to find a solution.

The error that tries to resolve this PR is the following one https://github.com/ElementsProject/lightning/pull/4676/checks?check_run_id=3139499262

It is a good point or am I'm working on some random error and it is not a cause? If it is can be a solution, maybe we can increase the default timeout error?


vincenzopalazzo:
it looks like not a solution :/ only a lucky run on my side.

PR #4680, opened 2021-07-23
channel lease: only send update_blockheight iff has channel lease

niftynei:
otherwise the peer may not know to expect it.

Changelog-None.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -2451,9 +2451,11 @@ static void resend_commitment(struct peer *peer, struct changed_htlc *last)
 					channel_feerate(peer->channel, REMOTE));
 		sync_crypto_write(peer->pps, take(msg));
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
There's nothing *wrong* with this, but since it's an odd message I can't see why a peer would reject it (unless someone else used 137 for some incompatible message?).

I'll apply for rc2 anyway, once CI happy.

rustyrussell (approved):
Ack 94a6d3c3d4718d22dbd0e964f1a7c10ea8e4b7ea

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4680/commits/94a6d3c3d4718d22dbd0e964f1a7c10ea8e4b7ea

rustyrussell:
Hmm, OK, my logs also show peers hanging up after we send this.  So it's believable?

PR #4683, opened 2021-07-25
Restore dual-funding GitHub CI

rustyrussell:
I just realized this isn't being tested!  Did some quick fixes, but it still fails.


niftynei:
dual-funding is fixed, flakes are just flakes 

ACK https://github.com/ElementsProject/lightning/pull/4683/commits/a19b6fb7dbad8dd651b0a48dd710b6e0d08d035e

PR #4686, opened 2021-07-27
pyln: Fix backward compat issue in `LightningRpc.close`

cdecker:
The issue is that the new keyword `force_lease_closed` was being set
even if the user didn't specify it, which results in breakage if this
new pyln version talks to older c-lightning nodes that don't have this
keyword yet. By setting the default to `None` it gets filtered out if
the user has not explicitly set it, but still retains the `True` /
`False` values if they did.

Changelog-None Issue is not present in released versions


rustyrussell (approved):
Ack ecd143c948e397ba0dc7c408a90b080151663d0a

vincenzopalazzo (commented):
ack https://github.com/ElementsProject/lightning/pull/4686/commits/ecd143c948e397ba0dc7c408a90b080151663d0a

PR #4700, opened 2021-08-05
Default to lease-only for funder plugin

niftynei:
Quick patch to default to only 'lease-only' channel leases. Effectively shuts off putting money into channels on RBF attempts in the default case.


rustyrussell (approved):
Ack dc1fa814ca9808c84fd20e601fe48bbde57df54a

PR #4713, opened 2021-08-10
common/tests: these test are no longer experimental-only.

rustyrussell:
Changelog-None


vincenzopalazzo (changes requested):
There is some missing stuff here?

rustyrussell:
> There is some missing stuff here?

No, we literally move the test names: exp-run-xxx.c is only run when experimental mode is on, run-xxx.c is always run.

Of course, it revealed a warning with -O3 which needed fixing.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4713/commits/cb88e706e0528466554b24d07c404bb2c8ab1c5b

PR #4716, opened 2021-08-12
chaintopology: tell gossipd that channels no longer exist on reorg.

rustyrussell:
This actually caused the flake in test_funding_reorg_private, where
l1 and l2 might not mark the original channel disabled.  In fact, they
should *remove* it as it gets reorged out.


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4716/commits/49b5785fde923c0c6426866efa483a56239dc389

PR #4724, opened 2021-08-18
missing --lightning-dir and exact backup file

LightningHelper:
It seems that the plugin require `--lightning-dir` , using without `--lightning-dir` result with an error.
Also the exact backup file has to be stated.


rustyrussell:
Ack https://github.com/LightningHelper/lightning/commit/794c646a932658b6f11a703f10b63df07377dff1

This is indeed correct, and matches the README.md in plugins/backup.

PR #4731, opened 2021-08-20
Improve connectd to try non-tor connection first and filter duplicates

m-schmoock:
This does:
- remove duplicate connection attempt if addr_hint was set and also added by gossip.
- try non-tor connections first as they are usually faster and more reliable.
- renames variable "use_proxy_always" to "always_use_proxy" to match config.

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -1459,9 +1459,18 @@ static void add_seed_addrs(struct wireaddr_internal **addrs,
 	}
 }
 
+static bool wireaddr_int_equals_wireaddr(struct wireaddr_internal *addr_a,
+					 struct wireaddr *addr_b)
+{
+	if (!addr_a || !addr_b)
+		return false;
+	return wireaddr_eq(&addr_a->u.wireaddr, addr_b);
+}
+
 /*~ This asks gossipd for any addresses advertized by the node. */
 static void add_gossip_addrs(struct wireaddr_internal **addrs,
-			     const struct node_id *id)
+			     const struct node_id *id,
+			     struct wireaddr_internal *addrhint)
 {
 	u8 *msg;
 	struct wireaddr *normal_addrs;
@@ -1483,6 +1492,24 @@ static void add_gossip_addrs(struct wireaddr_internal **addrs,
 
 	/* Wrap each one in a wireaddr_internal and add to addrs. */
 	for (size_t i = 0; i < tal_count(normal_addrs); i++) {
+		/* add TOR addresses in a second loop */
+		if (normal_addrs[i].type == ADDR_TYPE_TOR_V2 ||
+		    normal_addrs[i].type == ADDR_TYPE_TOR_V3)
+			continue;
+		if (wireaddr_int_equals_wireaddr(addrhint, &normal_addrs[i]))
+			continue;
+		struct wireaddr_internal addr;
+		addr.itype = ADDR_INTERNAL_WIREADDR;
+		addr.u.wireaddr = normal_addrs[i];
+		tal_arr_expand(addrs, addr);
+	}
+	/* so connectd prefers direct connections if possible. */
+	for (size_t i = 0; i < tal_count(normal_addrs); i++) {
+		if (normal_addrs[i].type != ADDR_TYPE_TOR_V2 &&
+		    normal_addrs[i].type != ADDR_TYPE_TOR_V3)
+			continue;
+		if (wireaddr_int_equals_wireaddr(addrhint, &normal_addrs[i]))
+			continue;
 		struct wireaddr_internal addr;
 		addr.itype = ADDR_INTERNAL_WIREADDR;
 		addr.u.wireaddr = normal_addrs[i];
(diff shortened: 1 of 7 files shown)


m-schmoock:
@rustyrussell
I debugged `connectd` a bit and found out that my old TOR issues didn't apply to recent versions anymore and `always_use_proxy` seems to work as intended. Anyway I tweaked code a bit...

lightningd/plugin.c
```diff
@@ -1821,7 +1821,7 @@ plugin_populate_init_request(struct plugin *plugin, struct jsonrpc_request *req)
 	if (ld->proxyaddr) {
 		json_add_address(req->stream, "proxy", ld->proxyaddr);
 		json_add_bool(req->stream, "torv3-enabled", ld->config.use_v3_autotor);
-		json_add_bool(req->stream, "use_proxy_always", ld->use_proxy_always);
```
rustyrussell, line 1824: Needs to go through a deprecation cycle...

connectd/connectd.c
```diff
@@ -1483,6 +1492,24 @@ static void add_gossip_addrs(struct wireaddr_internal **addrs,
 
 	/* Wrap each one in a wireaddr_internal and add to addrs. */
 	for (size_t i = 0; i < tal_count(normal_addrs); i++) {
+		/* add TOR addresses in a second loop */
```
rustyrussell, line 1496: Weird that you put "Changelog-fixed: connectd: do not try address hint twice" instead of "Peter to connect to non-Tor addresses"? (Or both, but the former seems like a trivial change?)

rustyrussell (changes requested):
Minor comments: would you like to also add code to send pings to Tor connections which we initiated, every rand(30) + 30 seconds? Seems it would help with connectivity...

lightningd/plugin.c
m-schmoock: k, I added `if (deprecated_apis) json_add_bool(req->stream, "use_proxy_always" ...` . Is that sufficient?

m-schmoock:
@rustyrussell

I'm not too sure about the each 30secs ping on TOR connections just yet as there are some open questions:
 - What do we do if we don't get a response in time?
 - Should we kill the connection and retry?
    - If yes, after which timeout?
    - If no, what does the ping achieve? (Just hope that TOR does better internal magic?)
 - Is 30secs a good value for TOR?
 
Also Bolt#1 tells us using frequent lightning protocol pings can raise a problem:
```
	/* BOLT #1:
	 *
	 * A node receiving a `ping` message:
	 *  - SHOULD fail the channels if it has received significantly in
	 *    excess of one `ping` per 30 seconds.
	 */
```
When a connection gets laggy and say its TCP socket suddenly spits out 2minutes of 'old remote data' to the daemon  (which happens on TCP over TOR), there may be implementations that then `fail the channels` as it would receive 4 pings at once and they can't know that they were actually distributed 'evenly' over the past 2 minutes.

m-schmoock:
@rustyrussell
Also do we want to prefer IPv6 addresses over IPv4 if a node announces both?

lightningd/plugin.c
rustyrussell: Weird, because they *should* be!

Yes, that's sufficient

rustyrussell:
> @rustyrussell
> 
> I'm not too sure about the each 30secs ping on TOR connections just yet as there are some open questions:

It's a separate PR, but I think it would be useful.

>     * What do we do if we don't get a response in time?
> 
>     * Should we kill the connection and retry?

My advice: only send a ping if you made the connection (no point both sides doing it, might as well establish a convention).

>       * If yes, after which timeout?
>       * If no, what does the ping achieve? (Just hope that TOR does better internal magic?)

I think if you're trying to send the next ping and you haven't got the reply from the last, you close the connection.

>     * Is 30secs a good value for TOR?

Well, if we can't get an HTLC committed (add, commit) in 30 seconds we already kill the connection and fail the HTLC, and we won't send the commit unless we've heard from the peer in the last 30 seconds (in which case we ping).

So if your connection is slower than that, c-lightning isn't going to work.  And while we could extend that, it's pretty unfriendly to the rest of the network.

> Also Bolt#1 tells us using frequent lightning protocol pings can raise a problem:
> 
> ```
> 	/* BOLT #1:
> 	 *
> 	 * A node receiving a `ping` message:
> 	 *  - SHOULD fail the channels if it has received significantly in
> 	 *    excess of one `ping` per 30 seconds.
> 	 */
> ```
> 
> When a connection gets laggy and say its TCP socket suddenly spits out 2minutes of 'old remote data' to the daemon (which happens on TCP over TOR), there may be implementations that then `fail the channels` as it would receive 4 pings at once and they can't know that they were actually distributed 'evenly' over the past 2 minutes.

No, because you are also not supposed to send one until you've got the previous reply.

Note that channeld already has some ping logic, and it's probably a good place to start with this for Tor.

rustyrussell (approved):
Ack da0440e574804fbeca053c75d4354800a4e4c527

PR #4736, opened 2021-08-22
pytest: remove only_one() duplicate

theStack:
The function is already provided in contrib/pyln-testing/pyln/testing/utils.py (which is also imported in this module), so there is no need to define it a second time.


rustyrussell (approved):
Ack 089feee36acaf2800a9925f87a7f0195d816414a

Issue #4721, opened 2021-08-16
Huge memory usage of topology process: memory leak?


whitslack:
### Issue and Steps to Reproduce

Is it expected that the `topology` plugin process should be using multiple gigabytes of RAM?

```
# grep -E '^(Vm|Rss)' "/proc/$(pidof topology)/status"
VmPeak:  2626832 kB
VmSize:  2626832 kB
VmLck:         0 kB
VmPin:         0 kB
VmHWM:   1212888 kB
VmRSS:       284 kB
RssAnon:               0 kB
RssFile:             284 kB
RssShmem:              0 kB
VmData:  2505364 kB
VmStk:       132 kB
VmExe:       492 kB
VmLib:      4288 kB
VmPTE:      5200 kB
VmSwap:  2505060 kB
```

Almost all of the process's memory is swapped out, suggesting that the process is not actively referencing those pages.

Immediately after a call to `lightning-cli listnodes; lightning-cli listchannels; lightning-cli listincoming`, the process has only swapped back in a small (61MiB) portion of its huge (2583MiB) memory footprint.

**I suspect a memory leak. What would be the best way of determining if that is indeed occurring?**

### `getinfo` output
This is the release version 0.10.1.

AutonomousOrganization:
I also am experiencing memory leak ish instability after upgrading to 10.1 that has led me to disable the plugin I am using (clboss) for now.

rustyrussell:
I think this is a false positive: you're seeing the mmap of the gossip store?

If it grows significantly over time, that's an issue...

whitslack:
@rustyrussell: Memory-mapped gossip store pages wouldn't be accounted in `VmSwap`. Since they're file-backed, they'd simply be evicted from RAM, to be re-fetched from file as needed. Only anonymous pages (and COW'd private copies of file-backed pages) go into `VmSwap`.

rustyrussell:
Hmm, usage here is much lighter, but jumped 200M the first time I called (lightning-cli listnodes; lightning-cli listchannels; lightning-cli listincoming) > /dev/null.

(Then nothing moved it again).
I think I know what it must be, let me see if I'm right...

whitslack:
[...]

PR #4737, opened 2021-08-23
memleak hook: find leaks in plugins

rustyrussell:
And indeed, it found some.  None as major as the one in #4721 but that seems to be when the plugin doesn't read the entire JSON stream at once, and leaks the command struct, which doesn't happen in our tests (at least here).

This is based on #4754 thanks to @cdecker suggesting a generic shutdown notification!


rustyrussell:
Trivial fix for non-DEVELOPER unit test builds

plugins/libplugin.c
```diff
@@ -644,6 +645,11 @@ handle_getmanifest(struct command *getmanifest_cmd,
 		}
 		json_object_end(params);
 	}
+#if DEVELOPER
+	json_object_start(params, NULL);
+	json_add_string(params, "name", "memleak");
+	json_object_end(params),
+#endif /* DEVELOPER */
```
cdecker, line 657: This presumably forces a plugin to be recompiled if the `DEVELOPER` option doesn't match the one we compiled `lightningd` with. This seems suboptimal if we ever want to distribute pre-compiled plugins, for example as part of the reckless plugin manager.

cdecker: Would it make sense to always register this, but have stubs that do nothing if we aren't in `DEVELOPER=1` mode?

plugins/libplugin.c
```diff
@@ -1448,12 +1448,20 @@ static struct plugin *new_plugin(const tal_t *ctx,
 	p->in_timer = 0;
 
 	p->commands = commands;
+	if (taken(commands))
+		tal_steal(p, commands);
```
cdecker, line 1467: I'm surprised that this works with non-tal allocated arrays too. I would have guessed that `taken()` on `commands` from `keysend.c` for example, which is statically allocated, would fail when looking up the tal header.

cdecker:
Just a minor comment about the `DEVELOPER` flag being implicitly forced onto plugins that may have been compiled in another environment.

plugins/libplugin.c
cdecker: Alternatively we could add a `shutdown` hook that is always present, allows plugins to flush their state instead of killing them outright, and piggyback some information back to `lightningd` (in this case the memleak info, getting ignored if `lightningd` wasn't compiled with `DEVELOPER=1`)

plugins/libplugin.c
rustyrussell: ?  The only one which passes take(commands) is spender().

But yes, taken() works on *any* pointer: take(x) actually puts x into a hash table...

plugins/libplugin.c
rustyrussell: Yeah, I like a shutdown hook better.  Let me do that instead!

plugins/libplugin.c
cdecker: I see, I was under the impression that `take` was using the `tal_header`. Thanks for the clarification.

plugins/libplugin.c
```diff
@@ -1182,8 +1182,7 @@ static void PRINTF_FMT(1,2) log_memleak(const char *fmt, ...)
 	va_list ap;
 
 	va_start(ap, fmt);
-	/* FIXME: This is LOG_DEBUG until we fix leaks! */
-	plugin_logv(memleak_plugin, LOG_DEBUG, fmt, ap);
+	plugin_logv(memleak_plugin, LOG_BROKEN, fmt, ap);
```
cdecker, line 1185: I see, this is why the `shutdown` isn't a hook.

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/4737/commits/c8f4d31b687a5168a1cc1c8917f11207e37c8dfb

cdecker:
Reasing to address conflict in test_gossip.py

cdecker:
Had a minor fixup to apply: `plugin_exit` in `DEVELOPER=1` mode was missing the `plugin` argument.

cdecker:
Unused variable in `DEVELOPER=0`, added more compile guards to drop `has_shutdown_notif` completely in that config.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4737/commits/9315cee442a6a19477da9c9f959c8e8901458dcb

PR #4738, opened 2021-08-23
Move to minimum required mypy dependeces

vincenzopalazzo:
Fixes the following error https://github.com/vincenzopalazzo/lnprototest/runs/3401023486?check_suite_focus=true#step:3:5839 on `lnprototest`, we can safely require the minimum version instead of a fixed oneversion.

### DIFF
--- a/contrib/pyln-client/Makefile
+++ b/contrib/pyln-client/Makefile
@@ -25,7 +25,7 @@ check-pytest:
 	pytest tests
 
 check-mypy:
-	MYPYPATH=$(PYTHONPATH) mypy --namespace-packages tests pyln
+	MYPYPATH=$(PYTHONPATH) mypy --namespace-packages --follow-imports=skip tests pyln
 
 $(SDIST_FILE):
 	python3 setup.py sdist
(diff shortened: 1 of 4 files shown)


rustyrussell:
Obviously untested?

This works:

```
diff --git a/contrib/pyln-client/Makefile b/contrib/pyln-client/Makefile
index c36116896..748551157 100644
--- a/contrib/pyln-client/Makefile
+++ b/contrib/pyln-client/Makefile
@@ -25,7 +25,7 @@ check-pytest:
 	pytest tests
 
 check-mypy:
-	MYPYPATH=$(PYTHONPATH) mypy -p --namespace-packages tests pyln
+	MYPYPATH=$(PYTHONPATH) mypy --namespace-packages -p pyln
 
 $(SDIST_FILE):
 	python3 setup.py sdist
```

vincenzopalazzo:
Ops @rustyrussell, GH Actions was very slow last night and I start to do other stuff. However, the problem that I was trying to resolve it the following one https://github.com/vincenzopalazzo/lightning/runs/3402999779#step:5:2214

This looks like that in some new version of `mypy` there is some new check, and it throws an error. I'm not able to catch right now. Any idea?

rustyrussell (approved):
Ack cbd587cb8b4e4a9b389ff1143253ae88c74abd51

PR #4741, opened 2021-08-25
pylightning: fix requirements to work with pyln-client

m-schmoock:
Without this, we cannot pip install pylightning and pyln-client in the
same environment anymore, as it tries to pull in an incompatible version
of pyln-client.

Changelog-None


cdecker:
ACK https://github.com/m-schmoock/lightning/commit/ff70c7f3f7d17598f962c92be2b90914b21e73a7

PR #4742, opened 2021-08-26
Required jsonrpc string propriety inside the JSON rpc method request.

vincenzopalazzo:
I discovered that my java RPC wrapper has some missing requirements with the jsonrpc 2.0 protocol, and the unusual thing is that it is accepted inside c-lightning a quest without `"jsonrpc":"2.0"` and as specified in [the specification protocol](https://www.jsonrpc.org/specification) the string must be present inside the request.

- [X] First run it is only for testing.


rustyrussell (approved):
Ack fe294dcc43b0477081422e2e585ed5691dbdf1c9

rustyrussell:
Thanks everyone, this is great!

PR #4745, opened 2021-08-27
Give the possibility to delete all the inactive offers.

### LINKED ISSUE #4744: Add command to clean up inactive offers

```
$ lightning-cli listoffers | grep active | grep false | uniq -c
      5          "active": false,
```

I would like to get rid of the inactive offers similarly to `delexpiredinvoice`.

### `getinfo` output

https://ln.bublina.eu.org/

vincenzopalazzo:
Fixes #4744

There is a better way to delete all the offers from the DB @rustyrussell, I fall in the enum trick and maybe I'm missing some things here?

### DIFF
--- a/lightningd/offer.c
+++ b/lightningd/offer.c
@@ -528,3 +528,27 @@ static const struct json_command payersign_command = {
 	"Sign {messagename} {fieldname} {merkle} (a 32-byte hex string) using public {tweak}",
 };
 AUTODATA(json_command, &payersign_command);
+
+static struct command_result *json_delinactiveoffers(struct command *cmd,
+						     const char *buffer,
+						     const jsmntok_t *obj UNNEEDED,
+						     const jsmntok_t *params) {
+	struct json_stream *response;
+
+	if (!param(cmd, buffer, params, NULL))
+		return command_param_failed();
+
+	wallet_offer_delete_inactive(cmd->ld->wallet);
+
+	//TODO: it is better put in the response the list of deleted offers?
+	response = json_stream_success(cmd);
+	return command_success(cmd, response);
+}
+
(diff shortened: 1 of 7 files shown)


rustyrussell:
Yes, better to do in a single db statement rather than iterating like this.

Moreover, we should not delete ones referred to by elsewhere, I think.  The payments and invoices tables have local_offer_id fields which can refer into here.  

Something like "DELETE FROM offers WHERE (status & ?) == 0 AND offer_id NOT IN (SELECT local_offer_id FROM invoices WHERE local_offer_id IS NOT NULL) AND offer_id NOT IN (SELECT local_offer_id FROM payments WHERE local_offer_id IS NOT NULL)" then `db_bind_int(stmt, 0, OFFER_STATUS_ACTIVE_F)`.

Aside: autoclean should probably clean up old payments which have failed, too, if they're old enough.  That means this offer cleaning will eventually work if the only payments referring to it were failed ones.

jsarenik:
Looks good to me, thanks @vincenzopalazzo !

vincenzopalazzo:
Ops sorry for the delay here @rustyrussell.

>Moreover, we should not delete ones referred to by elsewhere, I think. The payments and invoices tables have local_offer_id fields which can refer to here.

Mh right, I was thinking if a user wants to delete an offer he wants to be deleted also an invoice generated by it or a payment? we need to maintain track of it in some way right? I can sell my phone with offers, and when it is over I want to delete the offers from my DB. However, I want to maintain track of an invoice/payment received or send. (I'm making confusion here with send received, sorry)

Thanks @jsarenik to ping me on this issue :)

vincenzopalazzo:
this required more love

PR #4751, opened 2021-08-31
Detect and fix stack overflows!

rustyrussell:
First patch simple enables CI to find the bug, second fixes it.  If there's a third, that means we found more!

### DIFF
--- a/common/initial_commit_tx.c
+++ b/common/initial_commit_tx.c
@@ -1,5 +1,6 @@
 #include <bitcoin/script.h>
 #include <bitcoin/tx.h>
+#include <ccan/array_size/array_size.h>
 #include <ccan/endian/endian.h>
 #include <common/initial_commit_tx.h>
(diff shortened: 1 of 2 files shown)


cdecker:
> First patch simple enables CI to find the bug, second fixes it. If there's a third, that means we found more!

Seems to be missing the second patch

niftynei:
The bug was being triggered in a EXPERIMENTAL_DUAL_FUND flow, perhaps the EXPERIMENTAL_FEATURES flag doesn't trigger it?

rustyrussell:
OK, so I changed it to turn on stronger stack detection always, in configure.  Indeed, it broke as expected.  Then I pushed the fix, and now it's good.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4751/commits/401d0c57c68abde3a4ed3b39d6045b8a77f73dce

PR #4752, opened 2021-08-31
PSBT comparison problems

niftynei:
@grubles found a bug where our PSBT 'is identical' code was failing -- the updated version included a witness script, whereas the original  did not. `libwally` pushes a type-byte onto the serialized version if there's any bytes at all -- setting the bytes to null avoids this.


common/psbt_open.c
```diff
@@ -491,3 +491,16 @@ bool psbt_has_our_input(const struct wally_psbt *psbt)
 
 	return false;
 }
+
+bool psbt_contribs_changed(const tal_t *ctx,
+			   struct wally_psbt *orig,
+			   struct wally_psbt *new)
+{
```
rustyrussell, line 498: This is an antipattern!  It doesn't return an allocated object, so why pass in a ctx?

I think you meant to use tmpctx internally.

rustyrussell (commented):
Ack ad1bd9df20214f013c84cda9defc326ffed037ed

Minor change to remove ctx arg only.

common/psbt_open.c
niftynei: ACK, yes i can do that. Since this is common it seems .. better? to use an internal allocation/free cycle -- is that the wrong intuition?

rustyrussell: Hmm, tmpctx is pretty universal now.  Either way.

common/psbt_open.h
```diff
@@ -178,4 +178,13 @@ bool psbt_input_is_ours(const struct wally_psbt_input *input);
...
 bool psbt_has_our_input(const struct wally_psbt *psbt);
+
+/* psbt_contribs_changed - Returns true if the psbt's inputs/outputs
+ *                         have changed.
+ *
+ * @orig - originating psbt
+ * @new  - 'updated' psbt, to verify is unchanged
+ */
+bool psbt_contribs_changed(struct wally_psbt *orig,
+			   struct wally_psbt *new);
```
rustyrussell, line 189: These should also be const struct wally_psbt, no?

rustyrussell (approved):
Ack 1fb4e58697798bd1987dc10f76da1cffa2d94eea

PR #4755, opened 2021-09-01
wire: fix up lease_rates patch to what was actually applied

rustyrussell:
Running `make extract-peer-csv` gives the wrong result, so fix the patch.


niftynei:
ACK https://github.com/rustyrussell/lightning/commit/0745f9dbc38262f74808fd5cfa68c9783e0c64b4

afaict this commit brings the patch up to date with the `peer_wire.csv`, as well as the current proposed draft. Seems like I updated the peer_wire.csv without patching the patchfile that it should be generated from. Yikes.

https://github.com/lightningnetwork/lightning-rfc/pull/878/commits/6751029c946f4b224de27cb480da17e80f6d89e0#diff-ed04ca2c673fd6aabde69389511fa9ee60cb44d6b2ef6c88b549ffaa753d6afeR989

PR #4758, opened 2021-09-02
fix: crash when using a lot of hooks

m-schmoock:
I wrote a test plugin that excessively uses the notification and hook APIs in order to slow down the daemon just some 10 milliseconds  using `time.sleep()` on each hook/notification. The test `xfails`, because it runs into database synchronization issues:

```
2021-09-02T12:05:50.421Z **BROKEN** lightningd: Already in transaction from lightningd/bitcoind.c:392
2021-09-02T12:05:50.426Z **BROKEN** lightningd: FATAL SIGNAL 6 (version v0.10.1-49-gf5972dc)
2021-09-02T12:05:50.426Z **BROKEN** lightningd: FATAL SIGNAL 11 (version v0.10.1-49-gf5972dc)
```

Did I use the API in a bad way or is there something wrong with the code?
The plugin can set different delay values. On my (fast) machine, the crashes start to occur when using values of about 3ms.

### DIFF
--- a/tests/plugins/hook_everything.py
+++ b/tests/plugins/hook_everything.py
@@ -1,0 +1,149 @@
+#!/usr/bin/env python3
+
+#
+# This plugin hooks everything possible
+# and slightly slows execution down by sleeping some milliseconds.
+#
+# rationale: discover thread safetyness issues
+
+from pyln.client import Plugin
+import time
+
+RPC_CONTINUE = {'result': 'continue'}
+
+plugin = Plugin()
+plugin.delay = 10
+
+
+def foo():
+    time.sleep(plugin.delay / 1000)
+
+
+@plugin.subscribe("channel_opened")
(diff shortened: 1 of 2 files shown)


m-schmoock:
I can limit the problem to only hooking `db_write` AND i.e. `peer_connected`. When I just go for one of them the problem goes away...

ofc the subscription notifications are pointless in this testcase

rustyrussell:
db_write is special.  You cannot do *anything* else (like call back into lightningd) while inside this hook.  We should probably put more protections in place for this though.

It means that you need to take care combining db_write with other hooks: you must immediately service the db_write request whenever it comes in.

m-schmoock:
Yep, my bad. Do not use `db_write` with other hooks. This even happens on directly returning db_write when there is also other hooks for the same plugin

PR #4760, opened 2021-09-03
feat: add  channel_id and commitnum to  commitment_revocation hook

m-schmoock:
I thought having the `channel_id` and `commitnum` might be a useful addition to the  `commitment_revocation` hook


rustyrussell (approved):
Ack 845c8ed310e53396238499fcdb708b5874a2a2fd

cdecker:
Good catch, I was about to suggest doing this PR myself for a non-encrypting watchtower :-)

PR #4779, opened 2021-09-09
plugin: register `shutdown` instead of `plugin_shutdown` in notifications list

### LINKED ISSUE #4778: python plugin: topic 'shutdown' is not a known notification topic
Experimenting with a python [plugin](https://github.com/lightningd/plugins/pull/296) to catch the "shutdown" notification for cleanup, but plugin is never notified and CL has following in log:
```
plugin-thread_plugin.py: topic 'shutdown' is not a known notification topic
```

The plugin uses `from pyln.client import Plugin` (pyln-client 0.10.1) and subscribes to "shutdown" using the decorator `@plugin.subscribe("shutdown")`

### `getinfo` output
**edit**  ~`"version": "v0.10.1"`~ `"version": "v0.10.1-123-g84957be"

SimonVrouwe:
fixes #4778

My `make check` is broken, so hope for the best.

### DIFF
--- a/lightningd/notification.c
+++ b/lightningd/notification.c
@@ -555,7 +555,7 @@ void notify_channel_open_failed(struct lightningd *ld,
 	plugins_notify(ld->plugins, take(n));
 }
 
-REGISTER_NOTIFICATION(plugin_shutdown, NULL);
+REGISTER_NOTIFICATION(shutdown, NULL);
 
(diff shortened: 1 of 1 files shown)


rustyrussell (approved):
Ack abccb7dc49e390e4ae28f022a125a8a1d120db5f

PR #4789, opened 2021-09-14
Implementing a new node startup notification

vincenzopalazzo:
This PR is still a draft but I want to know more about the team feedback.

In several plugins, I had to need to get some information when the plugin is inited by the c-lightning node, like the getinfo. However, I can do this with some hack inside the code, but I think from an API viewpoint we can do better.

P.S: Commit after commit I really love all the API around c-lightning, but the following code scary me because I don't knwo the life cicle. Howe the method is generated? there is some ccan compiler?

```
#define REGISTER_NOTIFICATION(topic, serialize)                                  \
	struct notification topic##_notification_gen = {                         \
		stringify(topic),                                                \
		serialize,                                                       \
	};                                                                       \
	AUTODATA(notifications, &topic##_notification_gen);
```


SimonVrouwe:
From the [documentation](https://lightning.readthedocs.io/PLUGINS.html#a-day-in-the-life-of-a-plugin)

 - `init` is called after the command line options have been
   parsed and passes them through with the real values (if specified). This is also
   the signal that `lightningd`'s JSON-RPC over Unix Socket is now up
   and ready to receive incoming requests from the plugin.

Once those two methods were called `lightningd` will start passing
through incoming JSON-RPC commands that were registered and the plugin
may interact with `lightningd` using the JSON-RPC over Unix-Socket
interface.

vincenzopalazzo:
>From the documentation
>init is called after the command line options have been
parsed and passes them through with the real values (if specified). This is also
the signal that lightningd's JSON-RPC over Unix Socket is now up
and ready to receive incoming requests from the plugin.
Once those two methods were called lightningd will start passing
through incoming JSON-RPC commands that were registered and the plugin
may interact with lightningd using the JSON-RPC over Unix-Socket
interface.

(auto) nack, thanks for pinging me with the doc. I think I have some problem in code

PR #4798, opened 2021-09-21
New onion_message support (1/3): the cleanups

rustyrussell:

PR #4799, opened 2021-09-21
New onion_message support 2/3: importing the spec and creating helpers

rustyrussell:
(based on #4798)

Changelog-None


common/blindedpath.c
```diff
@@ -147,38 +147,15 @@ static bool blind_node(const struct privkey *blinding,
 	return true;
 }
 
-bool blinded_node_alias(const struct privkey *blinding,
```
cdecker: Looks like this commit could be merged with the prior one.

hsmd/hsmd_wire.csv
```diff
@@ -21,6 +21,7 @@ msgtype,hsmd_init_reply,111
 msgdata,hsmd_init_reply,node_id,node_id,
 msgdata,hsmd_init_reply,bip32,ext_key,
 msgdata,hsmd_init_reply,bolt12,pubkey32,
+msgdata,hsmd_init_reply,onion_reply_secret,secret,
```
cdecker, line 24: This is a constant that we could just as well store in the DB.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4799/commits/65694bfa77dd6282229c1d2f1882590f4d331168

common/blindedpath.c
rustyrussell: Yes, I forgot to fold it in.  Fixing.

hsmd/hsmd_wire.csv
rustyrussell: We *could*, but it is a secret which can be (ab)used to reveal our identity.  We need to derive it deterministically anyway, so it makes sense to do it here.

rustyrussell:
Trivial rebase (and merged fixup commit) now #4798 merged.

PR #4801, opened 2021-09-21
pytest: ensure hsmtool tests do complete writes.

rustyrussell:
This may be causing the test flakiness we sometimes see in CI.

Changelog-None


darosior (commented):
utACK 674016420f823c020bc86dca1ab6dc525563be28

Issue #4708, opened 2021-08-09
CI Integrations for FreeBSD/MacOS?


rustyrussell:
It would be good to get these into CI somehow.

PR #4802, opened 2021-09-21
Add missing headers for compiling on FreeBSD 13

laanwj:
- `netinet/in.h`: struct sockaddr_in, struct sockaddr_in6
- `unistd.h`: close()

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -7,6 +7,7 @@
 #include <common/type_to_string.h>
 #include <common/wireaddr.h>
 #include <netdb.h>
+#include <netinet/in.h>
 #include <wire/wire.h>
 
 bool wireaddr_eq(const struct wireaddr *a, const struct wireaddr *b)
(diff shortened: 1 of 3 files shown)


common/wire_error.c
```diff
@@ -2,6 +2,7 @@
 #include <ccan/tal/str/str.h>
 #include <common/type_to_string.h>
 #include <common/wire_error.h>
+#include <netinet/in.h>
```
rustyrussell, line 5: I can't see why this one is needed?

rustyrussell: I'm going to apply it without this change.

I tried to build on FreeBSD but it was beyond my skill and doc/INSTALL doesn't say how to do it.

laanwj: Huh, I'll try building it without this change and see what happens.

laanwj:
I must have misread an error. Pushed without the change in `wire_error.c`. Good catch. I re-checked that the other changes were necessary.

As for FreeBSD build instructions, if there's interest in that I could have a go at writing them, though I have a weird custom setup here so to make generally applicable ones I'd need to create a new VM too.

cdecker:
As part of #4708 I'm using https://github.com/vmactions/freebsd-vm to create a CI workflow, almost there, but would love to have canonical compile-from-source instructions, since I had to do a lot via trial and error.

PR #4803, opened 2021-09-21
Minor cleanups

cdecker:
Had these laying around for some time, and finally bundled them up.

 - Using wrong enum for pay status
 - Simplify the state aggregation by using orthogonal enum values
 - Silently drop the -27 already in chain failure in `bcli`
 - Error message not matching the verification

Especially the silencing of -27 should finally stabilize the lnprototests, but I'd like to have either @niftynei or @rusty check that it's actually ok to do.


rustyrussell (approved):
Ack dd638994838c9b7b0e3b7282fd50d0b63d2e83d6

PR #4804, opened 2021-09-22
Rework ping logic, have channeld ping regularly.

rustyrussell:
This might help stability of Tor connections, but I made it universal since it's easier (in particular, it's hard to tell which *incoming* connections are Tor, since they usually come from localhost).

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -49,6 +49,15 @@
 #define MASTER_FD STDIN_FILENO
 #define HSM_FD 6
 
+enum pong_expect_type {
+	/* We weren't expecting a ping reply */
+	PONG_UNEXPECTED = 0,
+	/* We were expecting a ping reply due to ping command */
+	PONG_EXPECTED_COMMAND = 1,
+	/* We were expecting a ping reply due to ping timer */
+	PONG_EXPECTED_PROBING = 2,
+};
+
 struct peer {
 	struct per_peer_state *pps;
 	bool funding_locked[NUM_SIDES];
@@ -104,7 +113,7 @@ struct peer {
 	struct oneshot *ping_timer;
 
 	/* Are we expecting a pong? */
-	bool expecting_pong;
+	enum pong_expect_type expecting_pong;
 
 	/* The feerate we want. */
 	u32 desired_feerate;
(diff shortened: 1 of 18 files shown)


channeld/channeld.c
```diff
@@ -2252,6 +2267,19 @@ static void peer_in(struct peer *peer, const u8 *msg)
...
+			handle_ping_reply(peer, msg);
+			/* fall thru */
+		case PONG_EXPECTED_PROBING:
+			peer->expecting_pong = PONG_UNEXPECTED;
+			return;
+		case PONG_UNEXPECTED:
+			status_debug("Unexpected pong?");
+			return;
+		}
+		abort();
```
vincenzopalazzo, line 2282: Why do we need to about with we receive the `PONG_EXPECTED_COMMAND` status?

tests/test_misc.py
```diff
@@ -254,7 +254,7 @@ def mock_getblock(r):
 
 
 def test_ping(node_factory):
-    l1, l2 = node_factory.line_graph(2, fundchannel=False)
+    l1, l2 = node_factory.line_graph(2)
 
     def ping_tests(l1, l2):
         # 0-byte pong gives just type + length field.
```
vincenzopalazzo, line 260:
```suggestion
        # 0-byte pong gives just type + length field.
        
        l1.rpc.ping(l2.info['id'])
```

Maybe we can test also if the node responds correctly to the ping that should be run without any problem?

vincenzopalazzo (changes requested):
LGTM https://github.com/ElementsProject/lightning/pull/4804/commits/212011def00c402def3cdb3b7f71e3041700ffae

I have only a small comment, that I was not able to answer myself.

channeld/channeld.c
rustyrussell: We process the reply, then fall through to the next statement: `peer->expecting_pong = PONG_UNEXPECTED;`

cdecker:
Rebased on `master` since this seems to have tripped up on #4811

rustyrussell:
Rebased again, fixed accidental delete of member initialization.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4804/commits/e9551cf9b886f8c704b4f34b064f66ae2cef2d7e

Issue #4781, opened 2021-09-09
Test that error replies containing channel_update get correctly treated.


rustyrussell:
https://twitter.com/SeverinAlexB/status/1435868545083183104

This is just a note to look over our tests and make sure we're correctly testing and handling this case.

PR #4808, opened 2021-09-22
pytest: test that we properly apply channel_update.

rustyrussell:
Spoiler: we don't, *if* it's in a routehint!

See also: https://github.com/ElementsProject/lightning/issues/4781

Pay should probably retry in this case?

### DIFF
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -5360,3 +5360,30 @@ def test_sendpay_dual_amounts(node_factory):
 
     with pytest.raises(RpcError, match=r'No connection to first peer found'):
         l1.rpc.sendpay(route=route, payment_hash="00" * 32)
+
+
+@pytest.mark.developer("Too slow without dev-fast-gossip")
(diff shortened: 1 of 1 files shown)


cdecker:
Ok, debugging this further it took me a while, but it appears `l3` is serving a stale `channel_update` with the old fees, which `l1` happily stores, and then tries the exact same path again (being the only one available). The gossmap is refreshed just before computing any route, by calling `get_gossmap` which refreshes internally, and I made sure we are seeing the `channel_update` alright, just the fees don't match our expectation.

@rustyrussell: I'm not exactly up to speed on the `channel_update` caching logic we use to defer updates in some cases, could this be related to that? I'd have expected the `setchannelfee` command to synchronously create a new update and stash it with `lightningd` which can then serve it, but we seem to have a stale one in there.

ShahanaFarooqui:
Closing all Draft PRs inactive for more than a year. Please feel free to reopen if required.

Issue #4469, opened 2021-04-06
Bump pyln-client version


zoedberg:
I've noticed that `pyln-client==0.10.0` is not available on PyPI and that the version in the `__init__.py` file needs to be bumped. Could you please bump the version and publish the updated package?

PR #4809, opened 2021-09-22
Test and production deployments to PyPI from Github Actions

cdecker:
Totally overengineered answer to #4469.

 - [x] Need to latch onto the version being tagged and then publishing to PyPI
Changelog-None


cdecker:
The events still need to be testing, but that'll obviously have to wait until we merge this. I think the triggers are set up correctly, but will monitor this during the release process.

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,72 @@
+---
+name: Publish Python 🐍 distributions 📦 to PyPI and TestPyPI
+on:
+  # Only deploy if we're the result of a PR being merged
+  push:
+    branches:
+      - master
```
vincenzopalazzo, line 7: Do you think that maybe it is better to use a label checker here? like this one? https://github.com/marketplace/actions/label-checker-for-pull-requests

We can make only push if there is some change in the pyln modules?

vincenzopalazzo (commented):
LGTM https://github.com/ElementsProject/lightning/pull/4809/commits/5a449836ce0531d4ec717dd4e21c7c318d2ecb06

There is a recursion inside the[ dependencies resolution](https://github.com/ElementsProject/lightning/runs/3688356495?check_suite_focus=true) maybe this is not a problem because we need to release before some package?

or `pyln.proto>=0.9.3` will cause this recursion?

vincenzopalazzo (changes requested):
I don't know why I can not put the comment inside the change, BTW I think that the version of the package `pyln.proto` could have a minimum requirement version like `pyln.proto>= 0.9.1`in the bolt7 (`contrib/pyln-spec/bolt7/requirements.txt`)

Or there is some reason that I'm missing here?

cdecker:
> I don't know why I can not put the comment inside the change, BTW I think that the version of the package `pyln.proto` could have a minimum requirement version like `pyln.proto>= 0.9.1`in the bolt7 (`contrib/pyln-spec/bolt7/requirements.txt`)
> 
> Or there is some reason that I'm missing here?

Had to remove the minimum version temporarily due to the way we auto-generate the version in shallow git clones we would get v0.0.0.post10000 as pyln-bolt7 version thus failing the version check. I have since started checking out the tags to generate the version correctly

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4809/commits/adfbb33fa24d99ffbb4508c591e2191be2d51a8b

PR #4811, opened 2021-09-23
pytest: fix flake in test_gossip.py::test_addgossip

rustyrussell:
We can miss it in both logs, so wait for it instead:

```
2021-09-22T07:25:59.1582950Z >       l3.rpc.addgossip(ann.split()[3])
2021-09-22T07:25:59.1583911Z E       AttributeError: 'NoneType' object has no attribute 'split'
```

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4811/commits/028697fde0d8ca32f68b458301b02775bbdca339

PR #4813, opened 2021-09-23
gci: Add macOS configuration to test compatibility

cdecker:
First part of https://github.com/ElementsProject/lightning/issues/4708, building Mac OS on every push and pull request, to make sure we don't break it accidentally.

We don't test since `make pytest` on Github Actions + Mac OS is very flaky, I'll continue working on enabling testing, but that might take a while.

Changelog-None

### DIFF
--- a/.github/workflows/macos.yaml
+++ b/.github/workflows/macos.yaml
@@ -1,0 +1,94 @@
+---
+name: Mac OS pytest
+on: [push, pull_request]
+jobs:
+  smoke-test:
+    name: Smoke Test macOS
+    runs-on: macos-latest
+    env:
+      DEVELOPER: 1
+      VALGRIND: 0
+      EXPERIMENTAL_FEATURES: 1
+      COMPAT: 0
+    strategy:
+      fail-fast: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.0.0
+
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
+
+      - name: Install dependencies
+        run: |
+          export BITCOIN_VERSION=0.20.1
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yaml
```diff
@@ -0,0 +1,94 @@
...
+            # running tests easier/feasible on CI (and pytest which
+            # keeps breaking the rerunfailures plugin).
+          pip3 install -U \
+            flaky \
+            blinker \
+            pytest-sentry \
+            pytest-test-groups==1.0.3 \
+            pytest-custom-exit-code==0.3.0 \
+            pytest-json-report \
+            mypy==0.790
```
vincenzopalazzo, line 76: This can cause a dependencies resolution error? since we now support also version `>=0.790`?

https://github.com/ElementsProject/lightning/blob/dd93f5dd2929b7ab8d97304de1f667e27b0bb064/contrib/pyln-proto/requirements.txt#L5

vincenzopalazzo (commented):
LGTM https://github.com/ElementsProject/lightning/pull/4813/commits/cb67b1faf085c167091ab060b274af70aa319362

I have only a small question, but I'm sure that I'm missing some things!

.github/workflows/macos.yaml
cdecker: No, those two restrictions are compatible, and will resolve to just exactly 0.790 which we know works. When we go for >=0.800 then we will have to relax this, but for now it works, and any PR changing that will fails its CI tests with the appripriate warnings.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4813/commits/cb67b1faf085c167091ab060b274af70aa319362

PR #4814, opened 2021-09-24
Wire fixes

devrandom:


cdecker:
Just to add a bit of context, I think this came out of a discussion we had about the generic message parsing framework @devrandom is building [here](https://gitlab.com/lightning-signer/rust-lightning-signer/-/merge_requests/171/diffs#bbae886b525c6e525c45ff38249b31451cf114ae_0_1) which separates the message serialization from application level parsing of the transaction. The length prefix may appear redundant since the bitcoin TX is self-contained and encodes its own length, however it requires the bitcoin transaction logic to be implemented in the transport layer, which is a layering violation.

devrandom:
oops, this was supposed to be against the lightning-signer fork, for internal discussion.  but might as well.

devrandom:
Wrote up a description, and will clean up the extra commit / look into the failing protocol test.

bitcoin/tx.c
```diff
@@ -700,6 +700,7 @@ struct bitcoin_tx *fromwire_bitcoin_tx(const tal_t *ctx,
 {
 	struct bitcoin_tx *tx;
 
+	fromwire_u32(cursor, max);
```
devrandom, line 703: is there a way to constrain `pull_bitcoin_tx` to use the length we just deserialized?  or at least check that those many bytes were consumed after the fact?

devrandom: Added a post-parse check

cdecker: We could also replace the `max` pointer with a pointer to the read prefix length, this way the tx parsing code itself will abort if a read would go beyond the supposed end of the tx stream. Would still need to have a check that the length was consumed completely (`max` gets decremented at each read and if it matches we should end up with a `max` of 0).

cdecker:
Just wanted to rebase in order to get rid of the flaky tests that have been stabilized in the meantime, but the PR doesn't allow maintainers to push to it, so you'll have to rebase to get rid of the stale warnings :_)

PR #4815, opened 2021-09-24
Remove from Github :evergreen_tree: autogen files.

vincenzopalazzo:
Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

PR #4816, opened 2021-09-25
misc: Add generated files to .gitignore

cdecker:
Slightly more precise than #4815 with less risk of accidentally ignoring a file we actually want.

Changelog-None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,10 @@ contrib/pyln-*/pyln_*.egg-info/
 release/
 tests/plugins/test_selfdisable_after_getmanifest
 .DS_Store
+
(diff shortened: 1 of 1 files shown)


vincenzopalazzo:
This looks like a duplicate https://github.com/ElementsProject/lightning/pull/4815 😄

vincenzopalazzo:
This looks like that the only missing here is the Changelog-None in the commit message

cdecker:
> This looks like that the only missing here is the Changelog-None in the commit message

Changelog-None is special since it is allowed in the OP of the PR as well, so we don't pollute the commit messages.

Issue #4729, opened 2021-08-20
listpeers channel list should be split into listpeerchannels, channels[] deprecated.


rustyrussell:
listpeers has grown: it lists not only the peers, but all the channels each of them has.  This is getting very unwieldy (and unnecessarily slow, see #4718 ), and should be split into a `listpeerchannels` which does that.

Steps:
- [ ] Create a listpeerchannels command next to the listpeers implementation, which returns a "channels" array.
- [ ] Write the doc/schemas/listpeerchannels.schema.json file (copy existing ones)
- [ ] Write some tests (steal from existing listpeers tests), put Changelog-Added: JSON-RPC: `listpeerchannels` new command.
- [ ] If deprecated_apis is false, don't include "channels" in listpeers.
- [ ] Update listpeers documentation to refer to listpeerchannels and remove channels[] from doc/schemas/listpeers.schema.json
- [ ] Make sure to put a Changelog-Deprecated: in the commit msg for the commit which removes from listpeers.
- [ ] Celebrate!

rl-d:
I'd like to take this issue.

vincenzopalazzo:
@rl-d I think just open a PR, maybe it is also easy to include the following issue inside this RPC refactoring https://github.com/ElementsProject/lightning/issues/4734

rl-d:
Status update: code and doc changes are in my fork and have been tested manually, but I'm having trouble getting "make check" to work in order to begin work on the tests...it's failing on what seems to be an unrelated test. [newbie]Is there a developer Telegram group or somewhere I can chat with somebody about it?[/newbie]

vincenzopalazzo:
@rl-d Telegram https://t.me/lightningd and you can use also a libera IRC channel with the name #c-lightning.

rl-d:
Thanks @vincenzopalazzo. Let's leave issue #4734 separate from this (per discussion in Telegram about making a plugin for listing closed channels).

vincenzopalazzo:
Mh @rl-d, wandering to give the possibility to a plugin to access from the lightning db, I think it is not safe

rustyrussell:
Could incorporate #4734 too...

vincenzopalazzo:
If someone is working on this issue, please react in some way 😄 (@rl-d) otherwise I can put in my queue this task and @rustyrussell can assign to me this issue to try to have this feature in the next release.

rl-d:
[...]

PR #4821, opened 2021-09-27
Move channels out of listpeers, issue #4729

cdecker:
Second attempt at fixing #4729


rl-d:
A huge thanks to @cdecker for recovering this PR!

Status: the error in topology.c reported by CI is a direct result of the deprecation of channels from listpeers, and the fix is pending the implementation of GraphQL queries. (As soon as the GraphQL PR is open, I will link it here.)

PR #4823, opened 2021-09-27
pytest: Stabilize `test_addgossip`

cdecker:
It was incredibly flaky due to the potential for l2 announcing the
channel before l1 could get to it, thus suppressing the outgoing
announcement which we were looking for. This now checks either
direction.

Before this fix the failure rate was 24% (out of 100 runs), afterwards
it's 0%.

Changelog-None


vincenzopalazzo (approved):
ack fe8e5e633b86e15aaf7a26d2ddebe59314e78066

PR #4825, opened 2021-09-28
ci: Add BSD CI configuration

### LINKED ISSUE #4708: CI Integrations for FreeBSD/MacOS?

cdecker:
Just like the Mac OS build config this currently just tests that we can build. It's a bit convoluted, but should ensure we remain buildable on FreeBSD. Enabling tests may take a bit more work.

Fixes #4708
Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4825/commits/53994cfdd0d027bff62022204371fcc98cc7bc64

cdecker:
This cannot be merged since the actions image is not from this repo, or from Github, so we first need to find a way to address that, otherwise all builds will be marked as failed.

PR #4829, opened 2021-09-29
bolt7: dns support

m-schmoock:
This implements RFC https://github.com/lightningnetwork/lightning-rfc/pull/911 to allow annoucements of DNS hostnames.
The feature can only be used with `EXPERIMENTAL`.
The bold quotes are disabled in the topmost commit to please CI.

Adds:
 - related fixes and changes (i.e. announce up to two addresses of the same type)
 - feature itself
 - and tests

### DIFF
--- a/common/json_helpers.c
+++ b/common/json_helpers.c
@@ -281,6 +281,10 @@ void json_add_address(struct json_stream *response, const char *fieldname,
 		json_add_string(response, "type", "torv3");
 		json_add_string(response, "address", fmt_wireaddr_without_port(tmpctx, addr));
 		json_add_num(response, "port", addr->port);
+	} else if (addr->type == ADDR_TYPE_DNS) {
+		json_add_string(response, "type", "dns");
+		json_add_string(response, "address", fmt_wireaddr_without_port(tmpctx, addr));
+		json_add_num(response, "port", addr->port);
 	} else if (addr->type == ADDR_TYPE_WEBSOCKET) {
 		json_add_string(response, "type", "websocket");
 		json_add_num(response, "port", addr->port);
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -287,8 +287,8 @@ REGISTER_TYPE_TO_STRING(wireaddr, fmt_wireaddr);
  * Returns false if it wasn't one of these forms.  If it returns true,
  * it only overwrites *port if it was specified by <number> above.
  */
-static bool separate_address_and_port(const tal_t *ctx, const char *arg,
-				      char **addr, u16 *port)
+bool separate_address_and_port(const tal_t *ctx, const char *arg,
+			       char **addr, u16 *port)
 {
 	char *portcolon;
 
@@ -318,6 +318,81 @@ static bool separate_address_and_port(const tal_t *ctx, const char *arg,
 		char *endp;
 		*port = strtol(portcolon + 1, &endp, 10);
 		return *port != 0 && *endp == '\0';
+	}
+	return true;
+}
+
+bool is_ipaddr(const char *arg)
+{
+	struct in_addr v4;
+	struct in6_addr v6;
+	if (inet_pton(AF_INET, arg, &v4))
+		return true;
+	if (inet_pton(AF_INET6, arg, &v6))
+		return true;
+	return false;
+}
+
+bool is_toraddr(const char *arg)
+{
+	size_t i, arglen;
+	arglen = strlen(arg);
+	if (!strends(arg, ".onion"))
+		return false;
+	if (arglen != 16 + 6 && arglen != 56 + 6)
+		return false;
+	for (i = 0; i < arglen - 6; i++) {
+		if (arg[i] >= 'a' && arg[i] <= 'z')
+			continue;
+		if (arg[i] >= '0' && arg[i] <= '9')
+			continue;
+		return false;
+	}
+	return true;
+}
+
+/* Rules:
+ *
+ * - not longer than 255
+ * - segments are separated with . dot
+ * - segments do not start or end with - hyphen
+ * - segments must be longer thant zero
+ * - lowercase a-z and digits 0-9 and - hyphen
+ */
+bool is_dnsaddr(const char *arg)
+{
+	size_t i, arglen;
+	int lastdot;
+
+	if (is_ipaddr(arg) || is_toraddr(arg))
+		return false;
+
+	/* now that its not IP or TOR, check its a DNS name */
+	arglen = strlen(arg);
+	if (arglen > 255)
+		return false;
+	lastdot = -1;
+	for (i = 0; i < arglen; i++) {
+		if (arg[i] == '.') {
+			/* segment must be longer than zero */
+			if (i - lastdot == 1)
+				return false;
+			/* last segment can not end with hypen */
+			if (i != 0 && arg[i-1] == '-')
+				return false;
+			lastdot = i;
+			continue;
+		}
+		/* segment cannot start with hyphen */
+		if (i == lastdot + 1 && arg[i] == '-')
+			return false;
+		if (arg[i] >= 'a' && arg[i] <= 'z')
+			continue;
+		if (arg[i] >= '0' && arg[i] <= '9')
+			continue;
+		if (arg[i] == '-')
+			continue;
+		return false;
 	}
 	return true;
 }
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -37,6 +37,11 @@ bool fromwire_wireaddr(const u8 **cursor, size_t *max, struct wireaddr *addr)
 	case ADDR_TYPE_TOR_V3:
 		addr->addrlen = TOR_V3_ADDRLEN;
 		break;
+	case ADDR_TYPE_DNS:
+		addr->addrlen = fromwire_u8(cursor, max);
+		memset(&addr->addr, 0, sizeof(addr->addr));
+		addr->addr[addr->addrlen] = 0;
+		break;
 	case ADDR_TYPE_WEBSOCKET:
 		addr->addrlen = 0;
 		break;
@@ -52,6 +57,8 @@ bool fromwire_wireaddr(const u8 **cursor, size_t *max, struct wireaddr *addr)
 void towire_wireaddr(u8 **pptr, const struct wireaddr *addr)
 {
 	towire_u8(pptr, addr->type);
+	if (addr->type == ADDR_TYPE_DNS)
+		towire_u8(pptr, addr->addrlen);
 	towire(pptr, addr->addr, addr->addrlen);
 	towire_u16(pptr, addr->port);
 }
@@ -211,6 +218,7 @@ bool wireaddr_is_wildcard(const struct wireaddr *addr)
 		return memeqzero(addr->addr, addr->addrlen);
 	case ADDR_TYPE_TOR_V2_REMOVED:
 	case ADDR_TYPE_TOR_V3:
+	case ADDR_TYPE_DNS:
 	case ADDR_TYPE_WEBSOCKET:
 		return false;
 	}
@@ -259,6 +267,8 @@ char *fmt_wireaddr_without_port(const tal_t * ctx, const struct wireaddr *a)
 	case ADDR_TYPE_TOR_V3:
 		return tal_fmt(ctx, "%s.onion",
 			       b32_encode(tmpctx, a->addr, a->addrlen));
+	case ADDR_TYPE_DNS:
+		return tal_fmt(ctx, "%s", a->addr);
 	case ADDR_TYPE_WEBSOCKET:
 		return tal_strdup(ctx, "websocket");
 	}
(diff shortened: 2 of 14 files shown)


m-schmoock:
I think this is ready for review now.
still not 100% happy with the way I changed `opt_add_addr_withtype` in `lightningd/options.c`, maybe you see a better way.

m-schmoock:
Note: the BOLT checker complains about my not yet merged changes on BOLT7. Not sure how we solve this chicken and egg when implementing upcoming protocol changes.
```
* `5`: DNS hostname; data = `[byte:len][len*byte:hostname][u16:port]` (length up to 258)
```

m-schmoock:
Thx for the tip, looks like we can set the BOLTVERSION in the makefile for that.

lightningd/options.c
```diff
@@ -228,23 +229,26 @@ static char *opt_add_announce_addr(const char *arg, struct lightningd *ld)
 		return tal_fmt(NULL, "address '%s' is not announcable",
 			       arg);
 
-	/* gossipd will refuse to announce the second one, sure, but it's
-	 * better to check and fail now if they've explicitly asked for it. */
+	/* Limit the number of addresses we can announce via options
+	 * by 2 per type to prevent spam. */
```
rustyrussell, line 259: I've removed this limit altogether in a separate PR (#4902), which I think is better?

I aim to apply that today (it also removed tor v2 support), so just drop this on rebase?

connectd/connectd.c
```diff
@@ -973,8 +981,46 @@ static void try_connect_one_addr(struct connecting *connect)
...
+					wireaddr_from_ipv6(&addrhint.u.wireaddr,
+							   &sa6->sin6_addr,
+							   addr->u.wireaddr.port);
+				} else {
+					/* skip unsupported ai_family */
+					continue;
+				}
+				tal_arr_expand(&connect->addrs, addrhint);
+				/* fixes VALGRIND complains for next iteration */
+				addr = &connect->addrs[connect->addrnum];
```
rustyrussell, line 1022: Gah, no, since you are now altering and potentially moving connect->addrs, valgrind is alerting you that addr can be stale!  But it's a real bug!

rustyrussell: (Better might be to save portnum in a var, and never reference addr in this loop)

connectd/connectd.c
```diff
@@ -973,8 +981,46 @@ static void try_connect_one_addr(struct connecting *connect)
...
+			}
+			freeaddrinfo(ais);
+			/* ... and continue recursion */
+			connect->addrnum++;
+			try_connect_one_addr(connect);
+			return;
```
rustyrussell, line 1028: `goto next` please...

common/json_helpers.c
```diff
@@ -281,10 +281,12 @@ void json_add_address(struct json_stream *response, const char *fieldname,
 		json_add_string(response, "type", "torv3");
 		json_add_string(response, "address", fmt_wireaddr_without_port(tmpctx, addr));
 		json_add_num(response, "port", addr->port);
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 	} else if (addr->type == ADDR_TYPE_DNS) {
```
rustyrussell: (GH won't let me comment on a *commit*, so I'm commenting on these lines).

I would remove the Changelog-Added from the prior commit, and use Changelog-EXPERIMENTAL: here instead.

common/wireaddr.c
```diff
@@ -37,11 +37,13 @@ bool fromwire_wireaddr(const u8 **cursor, size_t *max, struct wireaddr *addr)
...
 		addr->addrlen = TOR_V3_ADDRLEN;
 		break;
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 	case ADDR_TYPE_DNS:
 		addr->addrlen = fromwire_u8(cursor, max);
 		memset(&addr->addr, 0, sizeof(addr->addr));
 		addr->addr[addr->addrlen] = 0;
 		break;
+#endif
 	case ADDR_TYPE_WEBSOCKET:
```
rustyrussell: This is too much, IMHO.  Disable option parsing, sure, but don't disable it internally!

common/wireaddr.c
```diff
@@ -267,8 +273,10 @@ char *fmt_wireaddr_without_port(const tal_t * ctx, const struct wireaddr *a)
 	case ADDR_TYPE_TOR_V3:
 		return tal_fmt(ctx, "%s.onion",
 			       b32_encode(tmpctx, a->addr, a->addrlen));
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 	case ADDR_TYPE_DNS:
 		return tal_fmt(ctx, "%s", a->addr);
+#endif
```
rustyrussell, line 279: I would leave this too.  You might as well display the value?

common/wireaddr.h
```diff
@@ -152,11 +158,13 @@ struct wireaddr_internal {
 bool separate_address_and_port(const tal_t *ctx, const char *arg,
 			       char **addr, u16 *port);
 
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 bool is_ipaddr(const char *arg);
 
 bool is_toraddr(const char *arg);
 
 bool is_dnsaddr(const char *arg);
+#endif
```
rustyrussell, line 166: These functions are logical without EXPERIMENTAL_FEATURES, even if unused.

connectd/connectd.c
```diff
@@ -874,10 +874,12 @@ static struct io_plan *conn_init(struct io_conn *conn,
 			      "Can't connect to forproxy address");
 		break;
 	case ADDR_INTERNAL_WIREADDR:
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 		/* DNS should have been resolved before */
 		assert(addr->u.wireaddr.type != ADDR_TYPE_DNS);
-		/* If it was a Tor address, we wouldn't be here. */
 		assert(!is_toraddr((char*)addr->u.wireaddr.addr));
+#endif
+		/* If it was a Tor address, we wouldn't be here. */
```
rustyrussell: Yeah, leave this assert in even if !EXPERIMENTAL

contrib/pyln-spec/bolt7/pyln/spec/bolt7/text.py
```diff
@@ -286,7 +286,6 @@
          onion service addresses; Encodes:
          `[32:32_byte_ed25519_pubkey] || [2:checksum] || [1:version]`, where
          `checksum = sha3(".onion checksum" | pubkey || version)[:2]`.
-   * `5`: DNS hostname; data = `[byte:len][len*byte:hostname][u16:port]` (length up to 258)
```
rustyrussell: Use a separate comment for this section, i.e:

```
/* BOLT-bolt7-hostnames #7: 
 *     * `5`: DNS hostname; data = `[byte:len][len*byte:hostname][u16:port]` (length up to 258)
 */
```

Then you can check just those quotes with:
```
make check-source-bolt BOLTVERSION=bolt7-hostnames
```

(Or use a real git-id instead of a branch name, if you prefer!)

rustyrussell (changes requested):
Minor changes only, and rebase on top of torv2 removal...

connectd/connectd.c
ZmnSCPxj: Reseating `addr` here should fix the bug, but yes, it is a real bug, not a Valgrind weirdness.  Nevertheless, doing this is still risky especially if the code is modified later, so it is better to just avoid it and save portnum in a new var.

lightningd/options.c
m-schmoock: okay, will do rebase when TOR2 is removed on master.

common/wireaddr.c
m-schmoock: Okay, also lookup abd connect to DNS announced nodes without experimental then?
Or skip DNS announcements without experimental?

connectd/connectd.c
m-schmoock: @rustyrussell @ZmnSCPxj 
Okay, I set a correct comment. I didn't get what you mean with "portnum" since this "portnum" is not even used in this file.

connectd/connectd.c
```diff
@@ -976,8 +983,43 @@ static void try_connect_one_addr(struct connecting *connect)
 			af = AF_INET6;
 			break;
 		case ADDR_TYPE_DNS:
-			// TODO: resolve with getaddrinfo and set af
-			break;
+			/* Resolve with getaddrinfo */
+			memset(&hints, 0, sizeof(hints));
```
rustyrussell, line 992: This misses two import cases:

always-use-proxy -> don't resolve, just hand it to the proxy.
!use->dns -> skip this entry.

connectd/connectd.c
```diff
@@ -976,8 +983,43 @@ static void try_connect_one_addr(struct connecting *connect)
...
+			hints.ai_protocol = 0;
+			hints.ai_flags = AI_ADDRCONFIG;
+			gai_err = getaddrinfo((char *)addr->u.wireaddr.addr,
+					      tal_fmt(tmpctx, "%d",
+						      addr->u.wireaddr.port),
+					      &hints, &ais);
+			if (gai_err != 0)
+				status_failed(STATUS_FAIL_INTERNAL_ERROR,
+					      "DNS with getaddrinfo gave: %s",
+					      gai_strerror(gai_err));
```
rustyrussell, line 1004: Err, no!  If it doesn't resolve for some reason, don't shutdown lightningd!  Just skip it (maybe LOG_DEBUG the error?).

connectd/connectd.c
rustyrussell: We were suggesting keeping the port number in a local var, so you don't have to keep addr up-to-date in the loop (you access addr->u.wireaddr.port).  But it doesn't really help, since addr is used after the loop anyway...

connectd/connectd.c
```diff
@@ -874,8 +874,10 @@ static struct io_plan *conn_init(struct io_conn *conn,
 			      "Can't connect to forproxy address");
 		break;
 	case ADDR_INTERNAL_WIREADDR:
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 		/* DNS should have been resolved before */
 		assert(addr->u.wireaddr.type != ADDR_TYPE_DNS);
+#endif
```
rustyrussell: Don't need #ifdef around this, right?

rustyrussell (changes requested):

connectd/connectd.c
m-schmoock: Ah, that's why I didn't understand it.
I thought about you meaning the `wireaddr.port` but it didn't make sense because addr is used for other stuff as well so I dismissed it.

connectd/connectd.c
m-schmoock: Whops sure, fixed this. Now I use `status_debug` and `goto next;`. Question is if I need to do a `freeaddrinfo(ais)` when we encountered a `gai_error` from `getaddrinfo`. I think not, because that would be bad design.

connectd/connectd.c
m-schmoock: Sure, since we have the enum type `ADDR_TYPE_DNS` this precompiler `#if` is pretty pointless.

connectd/connectd.c
m-schmoock: good catch... will fix

m-schmoock: Do we want testcase that checks DNS is really skipped or passed to proxy if configured or is my recent change sufficient?

m-schmoock:
Currently CI fails on `wallet/test/run-db.c:209` which is very likely not related to this PR. I can even reproduce this on my local machine. So I guess there is some condition in master branch that causes it...

ZmnSCPxj:
> Currently CI fails on `wallet/test/run-db.c:209` which is very likely not related to this PR.

Kicked CI, just in case this is just test flakiness.

m-schmoock:
> Kicked CI, just in case this is just test flakiness.

@ZmnSCPxj The problem remains in `wallet/test/run-db.c:209`

connectd/connectd.c
m-schmoock: Hm, yes I need a testcase because the `if (!use_dns) goto next;` does not work as `bool use_dns = connect->daemon->use_dns;`  seems to always false.

Will fix that than also add a test...

m-schmoock: So I investiaged a bit:
`disable-dns` is set by pytest utils `contrib/pyln-testing/pyln/testing/utils.py` line 46. This makes the testcase fail now, as pytest tells any lightningd it spawns not to use DNS at all.

So the code is right, but the test can't work.
What should I do about it?

Second minor issue I saw is that the connect command (for a DNS only node) fails correctly but with an empty error message.

m-schmoock: @rustyrussell
I fixed that and added a working testcase for DNS enabled and disabled.

rustyrussell:
Ack 930d25bf3865c8d9229034b25552fb1825b9648d

Rebased onto master, fixed trivial conflicts.

rustyrussell (approved):

m-schmoock:
Seems like rust-lightning is going to add support for this as the first other implementation:  https://github.com/lightningdevkit/rust-lightning/pull/1329

PR #4830, opened 2021-09-30
channeld: handle reestablish from previous release with EXPERIMENTAL_FEATURES

rustyrussell:
We switched channel_types from optional to compulsory bits in
cb22015b2a19e4a4f9a259df9227dbd5c199872a.

The result is infinite reconnects against older nodes.

Changelog-None: bug introduced in this version.


rustyrussell:
This does not work, BTW, since the other peer gets upset!  I will alter the spec to change the numbers not to clash...

cdecker:
Drafted and removed from milestone since it cannot be merged as is, due to failure scenarios that weren't considered.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4830/commits/1c03b199d3c28ce22d938676171145e6f1a1f13e

PR #4831, opened 2021-09-30
Ancient databases nonstatic unilateral close fix

rustyrussell:
Found when I closed my oldest channel (since I've not seen the peer in several years, AFAICT).

### DIFF
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -750,6 +750,14 @@ static struct migration dbmigrations[] = {
      NULL},
     {SQL("CREATE INDEX channel_state_changes_channel_id"
 	 " ON channel_state_changes (channel_id);"), NULL},
+    /* Funny story!  Back in 2019, before postgresql, we used FALSE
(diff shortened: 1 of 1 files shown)


rustyrussell:
I confirmed that this fixed the problem on my node.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4831/commits/3fe7838f7876f996c929e6865deafffd41b63c57

P.S: I review the CI action error in different time today, but the error looks unrelated to me.

rustyrussell:
DO NOT SO APPLY THIS BROKE MY NODE!!!!

vincenzopalazzo:
>DO NOT SO APPLY THIS BROKE MY NODE!!!!

Mh, so this is the reason that your node is down all day yesterday. one thing that jumps to my mind, is that this cast is legal to migrate an old table to a new one. But, I have no answer here

PR #4833, opened 2021-10-01
update readme

swami-raj:


README.md
```diff
@@ -74,7 +74,7 @@ you've configured your node to expose the developer options, e.g.
 ```
 
 #### Mainnet Option
-To test with real bitcoin,  you will need to have a local `bitcoind` node running:
+To test with the  real bitcoin,  you will need to have a local `bitcoind` node running:
```
vincenzopalazzo, line 77:
```suggestion
To test with the bitcoin,  you will need to have a local `bitcoind` node running:
```

NIT, IMO it doesn't sound good.

README.md
```diff
@@ -25,7 +25,7 @@ c-lightning is a lightweight, highly customizable and [standard compliant][std]
 [![Documentation Status](https://readthedocs.org/projects/lightning/badge/?version=docs)][docs]
 
 This implementation has been in production use on the Bitcoin mainnet since early 2018, with the launch of the [Blockstream Store][blockstream-store-blog].
-We recommend getting started by experimenting on `testnet` (or `regtest`), but the implementation is considered stable and can be safely used on mainnet.
+We recommend getting started by experimenting on `testnet` (or `regtest`), but this implementation is considered stable and can be safely used on mainnet.
```
vincenzopalazzo, line 28: why we need to switch from `the` to `this`? IMO, `this` assumes that there are other implementations not stable, but the doc tells us only that the c-lightning can be considered stable.

vincenzopalazzo (commented):
Again, not the correct person to review the docs change.

This is only my opinion, I usually run out of grammar.

README.md
cdecker: Yeah, it's bitcoins the currency, not bitcoin the network / protocol

cdecker:
Sadly #4832 was quicker to get the `the` -> `this` fix in and the change in bitcoin to the bitcoin doesn't sound right.

PR #4782, opened 2021-09-10
Fixes make check, and adding a doc check on Gh actions :books:

vincenzopalazzo:
Fixes #4773
Fixes #4904

I think that this is the cause of make check failure, but this command continues to fail. I'm not able to fully understand the make script, but maybe @rustyrussell can give me some tips?

- [X] Fix `make check` command;
- [X] Adding a not dependent check of docs in the GH to make safe also the merge operation.
~- [ ] Adding missing parameters well documented in the `lightningd.config.5.md`~
~- [ ] Adding missing parameters (that was added in the previous point) into the `jsonschema`.~
- [X] Skip experimental proprieties

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

PR #4837, opened 2021-10-04
Add htlc dust ceiling

niftynei:
Implements https://github.com/lightningnetwork/lightning-rfc/pull/919


cdecker:
Excellent fix @niftynei  :+1: 

ACK https://github.com/ElementsProject/lightning/pull/4837/commits/ff381556e83b5501a6d421225352669048793576

cdecker:
There seem to be a couple of failing tests here:

 - Elements fees seem to behave slightly differently, which causes the following test failures:
   - `test_pay_peer` to slam into the dust limiter ("Too much dust to update fee"), which in a second run was obscured by a `dev_pay` failing with "ran out of routes" due to the channel being killed, but the cause is the same
   - `test_mpp_presplit` also fails with "Too much dust to update fee"
   - `test_delpay_payment_split` which I'm still looking into, runs out of routes, but doesn't have the dust warning
 - Normal valgrind seems to not like `test_lockup_drain` which admittedly was disabled due to falkyness. The last line before the crash however suggests that we are running into dust + feerate issues here too: "Feerate: RCVD_ADD_REVOCATION->RCVD_ADD_ACK_COMMIT LOCAL now 35000"

niftynei:
> There seem to be a couple of failing tests here:
> 
> * Elements fees seem to behave slightly differently, which causes the following test failures:
>   
>   * `test_pay_peer` to slam into the dust limiter ("Too much dust to update fee"), which in a second run was obscured by a `dev_pay` failing with "ran out of routes" due to the channel being killed, but the cause is the same
>   * `test_mpp_presplit` also fails with "Too much dust to update fee"
>   * `test_delpay_payment_split` which I'm still looking into, runs out of routes, but doesn't have the dust warning
> * Normal valgrind seems to not like `test_lockup_drain` which admittedly was disabled due to falkyness. The last line before the crash however suggests that we are running into dust + feerate issues here too: "Feerate: RCVD_ADD_REVOCATION->RCVD_ADD_ACK_COMMIT LOCAL now 35000"

Hmm. These elements failures are a good illustration of how this dust-bucket limit interacts badly with MPP + high feerate environments.

`test_pay_peer` fails because we attempt to bump the feerate above our 25% buffer; this would make our dust-bucket exceed the limit so we fail the channel. If I update the buffer to 50%, the channel no longer fails but the MPP payment fails since it eats up the entire bucket budget (at the higher feerate), so we start rejecting HTLC parts.

It might be worthwhile to take this into account when routing MPPs, so as not to route too many 'dusty' payments through the same peer (esp relevant during high fee periods).

I raised the dust-bucket limit for the test to avoid this and changed the channel to "warn" instead of fail if a proposed feerate update will put you above the dust-bucket (though it's worth noting that this will render this channel unusable until feerates go back down). Not sure if this is better or worse than just dropping to chain to be honest -- this problem is hard lol.

rustyrussell:
Sorry, AFAICT the whole concept of this change is wrong.  Maybe because the spec PR is an unreadable mess, but I've left comments there, too.

cdecker:
Turns out that the elements test is failing correctly, with a feerate adjustment after the htlcs have been added, pushing them into the dust bucket (despite 10% safety margin). So we should adjust the test.

@rustyrussell how would you solve this if we could do a complete greenfield implementation?

whitslack:
@niftynei: Shouldn't this PR have added `max-dust-htlc-exposure-msat` to `doc/lightningd-config.5.md`? I'm seeing that the 0.10.2 release adds this new option without having documented it.

vincenzopalazzo:
Right @whitslack,

This  https://github.com/ElementsProject/lightning/pull/4782/commits/272fb7245da55be082eb158d3384f4a19c73a22f should fix the `make check` and fixed the regression on this side by adding the check on the doc with the integration testing.

PR #4838, opened 2021-10-05
plugins/libplugin.c: Allow freeing notification `struct command *`.

ZmnSCPxj:
We always allocate a new `struct command` when we get a full JSON
object from stdin:

https://github.com/ElementsProject/lightning/blob/b2df01dc73ea7a51ae2a495281bf4d775eafa0a4/plugins/libplugin.c#L1229-L1233

If it happens to be a notification, we pass the `struct command` to
the handler, and not free it ourselves:

https://github.com/ElementsProject/lightning/blob/b2df01dc73ea7a51ae2a495281bf4d775eafa0a4/plugins/libplugin.c#L1270-L1275

There are only nine points in `plugins/libplugin.c` where we `tal_free`
anything, and only one of them frees a `struct command`:

https://github.com/ElementsProject/lightning/blob/b2df01dc73ea7a51ae2a495281bf4d775eafa0a4/plugins/libplugin.c#L224-L234

The above function `command_complete` is not appropriate for
notification handlers; the above function sends out a response
to our stdout, which a notification handler should not do.

However, as-is, it does mean that notification handling leaks
`struct command` objects, which can be problematic if we ever
have future built-in plugins which are significantly more
dependent on notifications.

This commit changes notification handlers to return
`struct command_result *`, because possibly in the future
notification handlers may want to perform `send_outreq`, so we
might as well use our standard convention for callbacks, and
to encourage future developers to check how to properly
terminate notification handlers (and free up the
`struct command`).

We also now provide a `notification_handled` function which a
notification handler must eventually call, as well as a
`notification_handler_pending` which is just a snowclone of
`command_still_pending`.

### DIFF
--- a/plugins/funder.c
+++ b/plugins/funder.c
@@ -708,9 +708,9 @@ json_rbf_channel_call(struct command *cmd,
 	return send_outreq(cmd->plugin, req);
 }
 
-static void json_disconnect(struct command *cmd,
-			    const char *buf,
-			    const jsmntok_t *params)
+static struct command_result *json_disconnect(struct command *cmd,
+					      const char *buf,
+					      const jsmntok_t *params)
 {
 	struct node_id id;
 	const char *err;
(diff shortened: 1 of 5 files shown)


vincenzopalazzo (approved):
ACK e509b84644c2b1a61487716220e6497e8403531c

CI looks unrelated

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/e509b84644c2b1a61487716220e6497e8403531c

Agreed, using a sink function like this is neatest, and matches the other patterns.

PR #4841, opened 2021-10-05
pylightning: Make it clear pyln-client should be used instead

michaelfolkson:
I don't think there are any contributing guidelines so hopefully this is ok :)


vincenzopalazzo (approved):
LGTM, and I agree to add a little but more content here :) Thanks!

ack https://github.com/ElementsProject/lightning/pull/4841/commits/3f7758fbd45d3d904a0f720cdef8b2793a2e2920

PR #4849, opened 2021-10-07
Update offers to latest spec

rustyrussell:
The only change of significance is the change of `chains` to `chain` in invoices and invoice_requests.  That's papered over for now with compatibility code.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4849/commits/faa2a8718c1ab12e8f6fc9fdf4073c846f0e13d2

PR #4854, opened 2021-10-09
Small Mac OS X python test fixes

ddustin:
* A memory leak detected (in debug mode only)
* Fix raised error testing regex to work on Mac OS X in a few places

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -96,7 +96,7 @@ def test_reconnect_channel_peers(node_factory, executor):
     wait_for(lambda: not only_one(l1.rpc.listpeers(l2.info['id'])['peers'])['connected'])
 
     # Now should fail.
-    with pytest.raises(RpcError, match=r'Connection refused'):
+    with pytest.raises(RpcError, match=r'(Connection refused|Bad file descriptor)'):
         l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
 
     # Wait for exponential backoff to give us a 2 second window.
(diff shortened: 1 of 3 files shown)


rustyrussell (approved):
Ack 7a8941f5c2e3fa6bd3eaf84a65de827f5465f280

tests/test_connection.py
```diff
@@ -96,7 +96,7 @@ def test_reconnect_channel_peers(node_factory, executor):
     wait_for(lambda: not only_one(l1.rpc.listpeers(l2.info['id'])['peers'])['connected'])
 
     # Now should fail.
-    with pytest.raises(RpcError, match=r'Connection refused'):
+    with pytest.raises(RpcError, match=r'(Connection refused|Bad file descriptor)'):
         l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
```
rustyrussell, line 100: Looks like this broke CI. Python regex issue, not sure this syntax is correct?

Also, we now run CI on MacOS and have not seen this? I cannot obviously see why it would get Bad file descriptor unless we have a bug?

tests/test_connection.py
```diff
@@ -2040,7 +2040,7 @@ def get_funded_channel_scid(n1, n2):
     # With 2 down, it will fail to fund channel
     l2.stop()
     l3.stop()
-    with pytest.raises(RpcError, match=r'Connection refused'):
+    with pytest.raises(RpcError, match=r'(Connection refused|Bad file descriptor'):
```
rustyrussell, line 2043: Actually, this is the problem: missing close ')'.  Pushing fix now...

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4854/commits/8e634a758ca2c6dbebf411f0586a3a0c7f7806b6

tests/test_connection.py
cdecker: The Mac OS CI task doesn't run tests at the moment, exactly because there is a lot of breakage in the tests when we do. So this type of fix will eventually get us to running tests on Mac OS too.

PR #4862, opened 2021-10-12
Add GraphQL capability

kandycoder:
Quick 6-min intro to this PR:
https://youtu.be/atAgRLvcBFA

For background, here are notes collected in the transcript of the c-lightning developer call of 2021-10-04:

c-lightning GraphQL prototype
Robert Dickinson demo: https://www.youtube.com/watch?v=pVkyRxs-QTI

On my side I’ve been working on the GraphQL implementation for the RPC call. It is exciting but on the other hand it is on the fringes, what does this have to do with c-lightning or Lightning in general.

I am really excited about it. GraphQL has been something we’ve been talking about adding for years. Having someone working on it is really great.

It is very powerful. I can see this really changing the face of how we interact with the RPC layer.

Are you adding it as a layer on top of existing RPCs or are you adding a new RPC that takes GraphQL syntax?

It is a new RPC. I started with listpeers because I wanted to break out the channels part. Instead of doing listpeers now you can just do graphql and give it straight GraphQL.

So good, I’m excited.

The cool thing is you could do 5 RPC calls in one, in theory, I have not implemented it for all RPC calls of course. That is where it could lead. It is very powerful in that sense.

In my earlier career I worked as a mobile dev. I never worked on an API that offered GraphQL stuff but it is super powerful for anyone who is adding client interfaces on top of backend. It is really great.

Did you discuss with Christian, you said on Telegram that you’d prefer to have some GraphQL stuff in the core of c-lightning rather than just a plugin? I think you said it was potentially more powerful. Did anything come of that discussion? Is it going to be a plugin?

This is not even in a PR yet. I don’t know what direction it is going to take. As far as what I’ve done in my demo, that is implemented directly in lightningd. That allows something that you can’t do with a plugin, you can optimize the processing to create the data. Using the listpeers example, when you do listpeers it gathers all the channel data and presents that in the JSON output. With GraphQL implemented in lightningd you are able to avoid going into the channels array unless it was actually requested. As an example it allows that level of optimization. Whereas if it was done in a plugin then you would be doing the whole RPC call and then filtering out. This is not as beneficial. There is still a lot that could be done with plugins as well and I can imagine that we will have a plugin for additional features also, for example for accessing the database.

vincenzopalazzo: With a plugin you can run different commands in parallel like listpeers and listnodes. If you were to display the data on some website you need to join these commands. In listpeers you have some information and in listnodes there is other information. With a plugin you can make some queries and run parallel commands and after join the result. I think a plugin could be very powerful also. I don’t know how you implemented the command line with GraphQL but another idea for a plugin could be this one.

If anyone is curious to look at the code, it is in a branch called listpeers2. You are welcome to check that out if you like. For the week going forward I will be working on doing tests for it and hopefully after that opening a PR.

Awesome, looking forward to it, that sounds great. It sounds like it probably won’t hit this release but the next one will be really exciting.

Probably too aggressive to get it into this release. I am still new to the project, new to GitHub and everything. It will take some time to get it perfect.

Telegram discussion on the GraphQL prototype

“This is neither a mere filter, nor a direct-to-database approach although the idea started with both of those two ways. But what I showed in the demo was a re-write of json_listpeers() and subfunctions so that instead of building the JSON output in the normal fixed way it builds only the parts that are requested by the supplied GraphQL pattern”

“Absolutely loving it. As you mention it can save us from doing unneeded work when all we care about is a subset of information, and we can still keep a default query to maintain the pattern we currently use. Is your demo implemented in C or is it built on top as a PoC? We’d likely want to integrate it with the way we query internally and build the responses in order to reap the performance benefits. Also this would be an awesome solution to filtering as well, as it currently stands each filtering option takes a separate CLI argument.”

“Agreed it is pure C, I will push it to GitHub so the code can be inspected.”

“So cool! The only question that I have is about the GraphQL property for each RPC command. Maybe it would be better to have a plugin like graphql.c that exposes a method to query the node with GraphQL? With GraphQL we can join the result of different RPC commands in a single query I think.”

“I tried the plugin approach and it has advantages and disadvantages.

Advantages: It is a plugin not touching the core code. Any other plugin could use its GraphQL RPC. It could provide DB access. It could provide convenient access to existing RPCs.

Disadvantages: No optimization of existing lightningd RPCs. No performance benefit when used to request data from existing RPCs.

I’d love to do the plugin approach to get its advantages for DB access but for RPC data I think it’s not the best approach. As you mention I would also like to consider alternatives to adding a property to each RPC command, maybe one new lightningd RPC for accessing everything the GraphQL way?”


cli/lightning-cli.c
```diff
@@ -286,11 +304,11 @@ static bool is_literal(const char *arg)
 }
 
 static void add_input(char **cmd, const char *input,
-		      int i, int argc)
+		      int i, int argc, enum escape escape)
 {
```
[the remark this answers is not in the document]
rl-d: Thank you. If we make this an experimental feature, then we would not be able to use it to finish PR #4821 until much later...just something to keep in mind.

The name `info` was chosen because in GraphQL it must be the same as the field name in the result set. It would be strange to have `getinfo` as the name in the result set.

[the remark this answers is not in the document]
vincenzopalazzo: >Thank you. If we make this an experimental feature, then we would not be able to use it to finish PR #4821 until much later...just something to keep in mind.

I'm not able to see any information on how you want to try to use this PR in the #4821, but anyway. I'm missing some things.

>The name info was chosen because in GraphQL it must be the same as the field name in the result set. It would be strange to have getinfo as the name in the result set.

It looks like a fork of opinion, but it makes sense.

Thanks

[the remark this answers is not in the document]
rl-d: > > Thank you. If we make this an experimental feature, then we would not be able to use it to finish PR #4821 until much later...just something to keep in mind.
> 
> I'm not able to see any information on how you want to try to use this PR in the #4821, but anyway. I'm missing some things.

The topology plugin currently calls `listpeers` and needs some channel info from the call, to know if each peer has an appropriate channel or not. For that reason, deprecating channels from `listpeers` breaks the topology plugin. To build the fix on top of this PR, the topology plugin can just call `graphql {peers{id,connected,channels{state}}}` and get the exact data it cares about. But ok, there are other ways to fix the topology plugin, even if it costs more RPC calls... it was just an idea to take advantage of this PR as a base.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+
+#define CONSUME_ONE { list_add(used, (struct list_node *)list_pop(tokens, struct graphql_token, list)); }
+#define RESTORE_ONE { list_add(tokens, (struct list_node *)list_pop(used, struct graphql_token, list)); }
+#define ROLLBACK(args) { while (list_top(tokens, struct graphql_token, list) != rollback_top) { RESTORE_ONE; } }
+#define OR if (!*err) goto exit_label; *err = NULL;
+#define REQ if (*err) { ROLLBACK(args); goto exit_label; }
+#define OPT *err = NULL;
+#define WHILE_OPT while(!*err); *err = NULL;
+#define LOOKAHEAD(args, tok) struct graphql_token *tok = list_top(tokens, struct graphql_token, list);
+#define MSG(msg) if (*err) *err = msg;
```
rustyrussell, line 52: So, Linus' rule is "Always return the error".  It's pretty good advice, and it reduces the need for this magic.

Since you can't return NULL AFAICT unless error is set, you could simply use that as a flag, rather than *err.  But stylistically I think returning err and taking a correctly-typed arg is the win.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+RET parse_directive_location(PARAMS);
+RET parse_executable_directive_location(PARAMS);
+RET parse_type_system_directive_location(PARAMS);
+
+RET parse_keyword(PARAMS, const char *keyword, const char *errmsg);
+RET parse_punct(PARAMS, int punct);
+RET parse_name(PARAMS);
+RET parse_int(PARAMS);
+RET parse_float(PARAMS);
+RET parse_string(PARAMS);
```
rustyrussell, line 144: Most of these should be static?

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+}
+
+/* The following parser functions follow special rules:
+ *	- The declaration is standardized with RET and PARAMS
+ *	- The "err" argument is assumed to be NULL upon entrance
+ *	- The "err" argument is set on failure
+ *	- If the function fails to parse, then "tokens" shall be as it was upon entrance
+ *	- INIT and EXIT macros are used
+ *	- Macros such as REQ and OPT facilitate readability and conciseness
+ */
```
rustyrussell, line 169: I think you need to be clearer; that the following code is not really C, it's designed for parsing clarity rather than C style; think of it as something generated rather that something to read.

It's tempting to separate it into a separate file for that reason, but having all one file is also nice.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
+/* MIT (BSD) license - see LICENSE file for details */
+#include "graphql.h"
+
+#include "ccan/tal/str/str.h"
+#include "ccan/utf8/utf8.h"
+
+
+// GraphQL character classes
+
+#define SOURCE_CHAR(c) ((c) == 9 || (c) == 10 || (c) == 13 || ((c) >= 32 && (c) <= 65535))
```
rustyrussell, line 10: 65535?  This is a very strange thing to ask of a char?

rustyrussell: Ah, I see you have 0x2026.  Which really needs a #define, I think.  And this test still doesn't make sense AFAICT?

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+			goto newchar;
+		}
+
+		// Consume other ignored tokens.
+		if (COMMA(c) || WHITE_SPACE(c)) {
+			c = *p++;
+			goto newchar;
+		}
+		if (COMMENT(c)) {
+			while ((c = *p++) != EOF && !EOF_CHAR(c) && COMMENT_CHAR(c))
```
rustyrussell, line 652: EOF check is wrong here, it's a nul-terminated string right?

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+			}
+
+			tok = tal(tok_list, struct graphql_token);
+			list_add_tail(tok_list, &tok->list);
+			tok->token_type = c;
+			tok->token_specific = c;
+			tok->token_string = NULL;
+			tok->source_line = line_num;
+			tok->source_column = start - line_beginning + 1;
+			tok->source_len = p - start;
```
rustyrussell, line 686: Tempting to suggest a new_tok(tok_list, ...) function?

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+#define COMMA(c) ((c) == 44)
+#define EOF_CHAR(c) ((c) == 0 || (c) == 4)
+#define PUNCTUATOR(c) (strchr("!$&().:=@[]{|}", c))
+#define HEX_DIGIT(c) (DIGIT(c) || ((c) >= 0x61 && (c) <= 0x66) || ((c) >= 0x41 && (c) <= 0x46))
+#define DIGIT(c) ((c) >= 0x30 && (c) <= 0x39)
+#define NAME_START(c) (((c) >= 0x61 && (c) <= 0x7A) || ((c) >= 0x41 && (c) <= 0x5A) || (c) == 0x5F)
+#define NAME_CONTINUE(c) (NAME_START(c) || DIGIT(c))
+
+// Safe copy helper
+#define cpystr(d,s) { char *cpystr_p; char *cpystr_q; for(cpystr_p = (s), cpystr_q = (d); *cpystr_p;) *cpystr_q++ = *cpystr_p++; *cpystr_q++ = *cpystr_p++; }
```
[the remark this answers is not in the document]
rustyrussell: Ah, from usage I intuit that this is safe where d and s overlap.  Subtle: definitely needs a comment!

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+							break;
+						case 't':
+							*rewrite_dest++ = '\t';
+							cpystr(rewrite_dest, q--);
+							break;
+						case 'u': {
+								// Insert escaped character using UTF-8 multi-byte encoding.
+								char buf[] = {*q++, *q++, *q++, *q++, 0};
+								int code_point = strtol(buf, 0, 16);
+								int bytes = utf8_encode(code_point, rewrite_dest);
```
rustyrussell, line 912: This can fail!  And return 0.

ccan/ccan/graphql/graphql.h
```diff
@@ -0,0 +1,230 @@
...
+#ifndef __GRAPHQL_H__
+#define __GRAPHQL_H__ 1
+
+#include <stdio.h>
+
+#include <ccan/list/list.h>
+#include <ccan/tal/tal.h>
+
+// Coding constants
+#define GRAPHQL_SUCCESS 0
```
rustyrussell, line 11: I think defining this as ((const char *)NULL) is a little safer for the callers, so they compare it correctly?

ccan/ccan/graphql/graphql.h
```diff
@@ -0,0 +1,230 @@
...
+};
+
+struct graphql_document {
+	struct graphql_definition *first_def;
+};
+
+
+struct graphql_token {
+	struct list_node list;
+	unsigned int token_type;
```
rustyrussell, line 234: Probably best to use an enum here, to give the caller a clue as to what values are valid.

ccan/ccan/graphql/graphql.h
```diff
@@ -0,0 +1,230 @@
...
+	unsigned int token_specific;
```
rustyrussell, line 235: It's unclear to me how this field is used?  It seem to be equal to token_type, or unset altogether?

ccan/ccan/graphql/graphql.h
```diff
@@ -0,0 +1,230 @@
...
+
+/* The lexer.
+ * INPUTS:
+ *	input - string to parse
+ *	ctx - parent tal context or NULL
+ *	tokens - a variable to receive the resulting token list
+ * RETURN:
+ *	GRAPHQL_SUCCESS or an error string.
+ */
+const char *graphql_lex(const char *input, const tal_t *ctx, struct list_head **tokens);
```
rustyrussell, line 252: I really prefer ctx arguments first, by convention.

ccan/ccan/graphql/graphql.h
```diff
@@ -0,0 +1,230 @@
...
+
+/* The parser.
+ * INPUTS:
+ *	tokens - the list produced by the lexer
+ *	doc - a variable to receive the resulting abstract syntax tree (AST)
+ * RETURN:
+ *	GRAPHQL_SUCCESS or an error string.
+ */
+const char *graphql_parse(struct list_head *tokens, struct graphql_executable_document **doc);
```
rustyrussell, line 261: You should probably document that the tokens list is emptied by this function?

rustyrussell (commented):
Hi!  Since this is your first major contribution to c-lightning (congratulations and welcome!) I'm going through with a fine-tooth comb making as many detailed comments as I can.

Feel free to push back on them where you disagree, of course!

rustyrussell (commented):
OK, I did a cursory review of the remainder.  The if/else if streq() branches are too ugly to live; this needs to be table driven.  The macros to create the functions take away more than they add, too.

But the core concept is very solid: a decent GraphQL implementation needs to reach into lightningd like this and we need to refactor our JSON-producing output in a more abstract way!

rustyrussell:
Having slept in it, could we pass the graphql arg (or a processed version?) to all the "to_json" helpers. They would then omit the field if filtered out, and use alias of necessary.

Trickier to handle object/array start and end, but we can manipulate the object to track what we're inside I guess, and importantly, make them return bool, so we know if they return false we should skip entire contents as an optimisation?

rustyrussell:
> Interesting idea. The complication would likely be the parameter parsing and error checking, which would still have to be done in a preparatory pass.

Yes.  I also realized we could reduce churn by hanging the filtering info off json_stream...

rl-d:
> > Interesting idea. The complication would likely be the parameter parsing and error checking, which would still have to be done in a preparatory pass.
> 
> Yes. I also realized we could reduce churn by hanging the filtering info off json_stream...

That sounds great... the more I mull it over—with the table approach to the preparatory pass in mind—the more I think this is the way to go!

rl-d:
Oops, one thing we sacrifice, however, is compliance with the GraphQL spec, which specifies that the order of the results should be the order specified by the caller.

rustyrussell:
Having slept in it, could we pass the graphql arg (or a processed version?) to all the "to_json" helpers. They would then omit the field if filtered out, and use alias of necessary.

Trickier to handle object/array start and end, but we can manipulate the object to track what we're inside I guess, and importantly, make them return bool, so we know if they return false we should skip entire contents as an optimisation?

> Oops, one thing we would sacrifice, however, is compliance with the GraphQL spec per [7.2.2](https://spec.graphql.org/draft/#sec-Serialized-Map-Ordering), which specifies that the order of the results should be the order specified by the caller.

But any problem can be solved with another layer of abstraction! :)

We can produce a tree, rather than directly creating the JSON. This conflicts somewhat with our desire to stream results as they are generated, but TBH that's only really useful for list* functions, in particular listnodes and listchannels. And guess what: we moved them out to plugins recently. So as long as we still stream plugin results, we can probably stop streaming internal results.

(Obv I'm punting on the question of how we route graphql to plugins, but that's tomorrow's problem anyway).

Note that this basically frees us from JSON. Is there a better transport to use for graphql? Or is JSON still the norm? (Tooling matters...)

rl-d:
I would have to defer to your experience on whether it would be wise to abandon streaming in lightningd. I was under the impression that even listpeers could return potentially hundreds of thousands of entries for a large node.

> Note that this basically frees us from JSON. Is there a better transport to use for graphql? Or is JSON still the norm? (Tooling matters...)

I gather that JSON is pretty much the norm.

rustyrussell:
> I would have to defer to your experience on whether it would be wise to abandon streaming in lightningd. I was under the impression that even listpeers could return potentially hundreds or thousands of entries for a large node.

If you have hundreds of peers, the memory use here isn't an issue though.  And I don't think you'd even notice the latency, either.

rustyrussell:
Ping @rl-d ?  What's the status of this, did you want some help reworking our JSON functions to produce an intermediate representation?

rl-d:
> Ping @rl-d ? What's the status of this, did you want some help reworking our JSON functions to produce an intermediate representation?

Thanks for the ping... So far I have worked out about 95% of a table-driven approach (with the remaining 5% having to do with some corner cases). I plan to finish that up in the coming days so I can at least push that piece up to github in a fully working form. I regret that things are moving slower than I wish due to spreading myself too thin. In that sense, help would certainly be a dream come true. However, I'm not sure if it is a good idea regarding the topic of the JSON functions. I have been growing uneasy about the non-callback approach, to the point of "pushing back" (as you allowed me to do in your remarks in an early review comment) for a couple of additional reasons:

1. GraphQL allows the same field to be requested multiple times with different aliases and/or different arguments (could be used e.g. to query the same amount in both sats and btc). Of course this too could be solved with the additional layer of abstraction suggested for field ordering, but it gets more complicated because there isn't a 1:1 correspondence between available fields vs. requested fields. I wonder how many other "gotchas" there will be and whether maybe sticking with the callback approach and just cleaning them up would be better, because:

2. GraphQL is "unapologetically driven" (as they say) by the user/client needs, and the callback function architecture is intended to facilitate that. It is even the paradigm in which the spec is written, with remarks such as the following:

"Fields are conceptually functions which return values, and occasionally accept arguments which alter their behavior. These arguments often map directly to function arguments within a GraphQL service’s implementation."

Of course that doesn't mean it must be implemented that way, but a different way could become problematic. There are a lot of capabilities GraphQL offers that might not necessarily be needed or used by us today, and if we create a middle layer, I fear we might limit and/or overcomplicate things and cause ourselves grief in the long run. I was hoping to give the non-callback approach a fair chance by at least trying to implement such a middle layer myself to discover whether my fears are perhaps unfounded, but if we speak of someone else helping in that area, then as one who has dedicated some time to the GraphQL spec, I felt I better mention my concern right now before anyone else steps into it unsuspectingly.

ccan/ccan/graphql/graphql.c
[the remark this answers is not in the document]
rl-d: Intuition correct; comment added:
// Helper for copying an overlapping string, since strcpy() is not safe for that

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+#define COMMENT_CHAR(c) (SOURCE_CHAR(c) && !LINE_TERMINATOR(c))
+#define STRING_CHAR(c) (SOURCE_CHAR(c) && !LINE_TERMINATOR(c) && (c)!='"' && (c)!='\\')
+#define BLOCK_STRING_CHAR(c) (SOURCE_CHAR(c))
+#define COMMA(c) ((c) == 44)
+#define EOF_CHAR(c) ((c) == 0 || (c) == 4)
+#define PUNCTUATOR(c) (strchr("!$&().:=@[]{|}", c))
+#define HEX_DIGIT(c) (DIGIT(c) || ((c) >= 0x61 && (c) <= 0x66) || ((c) >= 0x41 && (c) <= 0x46))
+#define DIGIT(c) ((c) >= 0x30 && (c) <= 0x39)
+#define NAME_START(c) (((c) >= 0x61 && (c) <= 0x7A) || ((c) >= 0x41 && (c) <= 0x5A) || (c) == 0x5F)
+#define NAME_CONTINUE(c) (NAME_START(c) || DIGIT(c))
```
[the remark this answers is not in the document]
rl-d: The point was to capture the definitions in the spec with these macros (and I have now added a comment to that effect). I also changed to char literals like '#' for readability.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+#define cpystr(d,s) { char *cpystr_p; char *cpystr_q; for(cpystr_p = (s), cpystr_q = (d); *cpystr_p;) *cpystr_q++ = *cpystr_p++; *cpystr_q++ = *cpystr_p++; }
+
+// Parser shorthands
+
+#define RET void *
+#define PARAMS struct list_head *tokens, struct list_head *used, const char **err
+#define ARGS tokens, used, err
+#define INIT(type) \
+	struct graphql_token *rollback_top = list_top(tokens, struct graphql_token, list); \
+	struct graphql_##type *obj = tal(tokens, struct graphql_##type); memset(obj, 0, sizeof(struct graphql_##type)); \
```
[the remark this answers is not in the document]
rl-d: Thanks for the talz tip...I used that here and in some other places. RET and PARAMS/ARGS actually started as a coding crutch, because I wasn't sure from the start how the prototype would turn out and I didn't want to spend time search-and-replacing while working out the details :D That was the most important mileage I was after... but now, there's not much reason to keep them. However, I didn't remove them (yet) because of the comment regarding Linus' rule of returning the error (below), which is still to be addressed.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+
+#define EXIT \
+	goto exit_label; \
+	exit_label: \
+	rollback_top = rollback_top; \
+	if (*err) obj = tal_free(obj); \
+	return obj; \
```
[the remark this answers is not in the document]
rl-d: Added comments and used (void)rollback_top in INIT.

ccan/ccan/graphql/graphql.c
```diff
@@ -0,0 +1,1031 @@
...
+
+#define CONSUME_ONE { list_add(used, (struct list_node *)list_pop(tokens, struct graphql_token, list)); }
+#define RESTORE_ONE { list_add(tokens, (struct list_node *)list_pop(used, struct graphql_token, list)); }
```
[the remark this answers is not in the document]
rl-d: Nice tip regarding {}. I applied it, and also the cleaner syntax for the list_pop result.

ccan/ccan/graphql/graphql.h
rl-d: Updated to the better ((const char *)NULL)

ccan/ccan/graphql/graphql.h
rl-d: Good idea...enum added

ccan/ccan/graphql/graphql.h
rl-d: Deleted - it was just accidentally leftover from the development phase.

ccan/ccan/graphql/graphql.h
rl-d: I switched the argument positions per the convention.

ccan/ccan/graphql/graphql.h
rl-d: Good idea, done.

lightningd/peer_control.c
```diff
@@ -923,6 +924,689 @@ static void json_add_channel(struct lightningd *ld,
...
+					local_feerate,
+					channel->our_config.dust_limit, LOCAL,
+					channel_has(channel, OPT_ANCHOR_OUTPUTS))?
+			true : false);
+	} else if (streq(name, "status")) {
+		if (in) {
+			if (hin->status != NULL)
+				json_add_string(response, alias, hin->status);
+			else
+				json_add_null(response, alias);
```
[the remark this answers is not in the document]
rl-d: No, we don't do this in the old code. This came about because with GraphQL, if the user specifies a field, then that field must appear in the result. (i.e. section 2.5 Fields ....All GraphQL operations must specify their selections down to fields which return scalar values to ensure an unambiguously shaped response....)

lightningd/peer_control.c
```diff
@@ -923,6 +924,689 @@ static void json_add_channel(struct lightningd *ld,
...
+		}
+	} else
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "unknown field");
```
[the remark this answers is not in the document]
rl-d: I have since added some error codes as well as the human-friendly info.

lightningd/peer_control.c
```diff
@@ -1454,33 +2138,238 @@ static void json_add_peer(struct lightningd *ld,
...
+						     params, queryerr);
+		if (!doc)
+			return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+					    "invalid GraphQL executable document");
+
+		response = json_stream_success(cmd);
+		for (def = doc->first_def; def; def = def->next_def) {
+			if (!def->op_def || def->frag_def)
+				return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+						    "fragments not supported");
```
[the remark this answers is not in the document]
rl-d: Yes, this code came from my original overnight prototype, and the next day I realized (after a few core dumps) that failing the stream after starting to stream successfully is not allowed. (That is what led me to the two-pass approach, which corrected this problem.)

lightningd/peer_control.c
```diff
@@ -2333,7 +2333,7 @@ static struct command_result *json_listpeers(struct command *cmd,
 	if (!param(cmd, buffer, params,
 		   p_opt("id", param_node_id, &specific_id),
 		   p_opt("level", param_loglevel, &ll),
-		   p_opt("query", param_string, &querystr),
+		   p_opt("query", param_escaped_string, &querystr),
```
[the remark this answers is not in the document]
rl-d: This has also since changed. Now we have a separate "graphql" method. My only hesitation with conditional-compiling (the "graphql" method) would be that we wouldn't be able to take advantage of any of it in other parts of our non-experimental code, but I would certainly defer to others on that.

rl-d:
Quick follow-up to my above remarks: I've pushed changes including a table-drive approach and most of the little improvements that were recommended. I also threw in a few lines of code to demonstrate querying a subset of a list (of peers in this case) instead of the whole list (see attached picture).

image

At the moment, I'm pretty happy with the current approach to having the prep functions prepare the tree, which is then traversed to generate the output.

ccan/ccan/graphql/graphql.c
rl-d: Yes, fixed.

ccan/ccan/graphql/graphql.c
rl-d: Agreed. I included it in the comments now.

ccan/ccan/graphql/graphql.c
rl-d: The number 65535 comes from the spec, and yes, it's indeed a bit strange. The spec does not assume a particular encoding of the input text (encoding is implementation-specific) and it restricts the meaningful input characters to a subset of the 7-bit ASCII range anyway, so in theory, this number can only come into play in certain comments that are allowed to contain arbitrary code points according to the implementation's chosen encoding. But since we treat input as chars (i.e. UTF-8) without validating multibyte code points, this ends up not being used. And since the comments are stripped anyway, there's really not much point in validating multibyte code points in the input. The only real benefit might be for more accurately tracking "character position" for error reporting, but does one cause an error in a comment, anyway? So in the end, yeah, it's kinda just a useless strange thing which is there only because the macro is intended to translate what the spec says, more or less verbatim (an intent which I've now added to the comments).

As for 0x2026, this constant is now defined in token_type_enum as PUNCT_SPREAD to signify the GraphQL spread operator (...). (The constant itself was "arbitrarily" chosen to be the Unicode code point for the ellipsis character, something out of the ASCII range.)

ccan/ccan/graphql/graphql.c
rl-d: Yes, good catch. (In an early incarnation, this function was using fgetc which of course can return EOF.)

ccan/ccan/graphql/graphql.c
rl-d: Good idea. (Not yet implemented.)

ccan/ccan/graphql/graphql.c
rl-d: I added a comment to indicate that this was taken into consideration and that a zero size will cleanly result in omitting the offending character.

rustyrussell:
OK, I'd like to start merging this.  It's going to be a little messy, but first thing I've done is update ccan with your latest graphql from this PR, and I'll import the latest version with graphql.

That's in https://github.com/ElementsProject/lightning/pull/4975

rl-d:
Awesome. Sorry about the ccan module changes in this PR (I realize that wasn't following the rule). If there's anything else that would make it less messy the next time, I'm happy to learn.

Blade272008 (approved):
Thanks

rustyrussell:
No, this was done the right way for review.  It was always going to get a bit messy to merge, but that's just a matter of typing...

rustyrussell:
I started to rebase this onto master and split into chunks (by taking the entire diff and splitting it into separate commits, e.g. docs, cli, utils...). But unfortunately I think it's the wrong approach.

I will try to find some hacking time on this.  My idea is:
1. Add json_omit_field() wherever we currently don't emit a field.  This does nothing for now, but allows for null in graphql queries.
2. We *don't* alter the high-level json_add_xxx() routines (or any of the current code), but change the lowlevel ones.
2. The lowlevel json_add routines are replaced by routines which build an in-memory tree in the now-misnamed "struct json_stream".
3. When in graphql mode, we suppress production of parts of the tree we don't want.
4. At the end, we serialize the tree for output.  For graphql mode, we reorder as required.

We can then optimize this: make all the json_object_start() and json_array_start() routines return bool, which returns false if we're in graphql mode and not interested.  Callers than then omit things entirely.  We probably need a json_field_wanted() which can similarly be used to avoid expensive queries if they're unwanted, too.

I have no idea how the filtering will work :(  I'll hack something, see if you can glue it together....

rl-d:
As you wish! I can see the appeal although it will also have its drawbacks. But hey, such are decisions beyond my purview at this point... so let's go for it and see what we can accomplish. It might not easily deliver all that GraphQL promises, but maybe that's not as important as other factors. In any case, thank you for your efforts to bring the PR forward and your care for the implementation!

vincenzopalazzo:
>I will try to find some hacking time on this. My idea is:

I like Rusty's idea here, we can just put a small chunk of code inside the commit that implements some functionality that may be useless inside the code, but prepare only some starting point. I didn't read yet the code for the PR, but if you want I can offer some of my hacking time to work on this splitting :) just let me know, I'm always up and running around.

>I have no idea how the filtering will work :( I'll hack something, see if you can glue it together....

This can be a good problem to pass a week to think into it, IMO the real improvement from the end-user and also the user that implements somethings in bash it is the possibility to filter or in a more general way to define inside the query the order or some logic to implementing the filtering option. Again, I would like to share some time to solve this problem, because I would like to have this inside the command line.

rl-d:
Maybe I'm missing something, but in the end, I guess this will give us the option of GraphQL syntactic sugar with very little of GraphQL's power. In my opinion it's kinda not worth it and a pretty far departure from the original presentation of this PR. Personally, I see myself losing interest, but I hope I'm proven wrong for the project's sake!

rl-d:
Did you consider that we might want to eventually implement mutations?

Did you consider that we might want to eventually auto-generate stub functions from the documentation XML schemas?

When I hear "no idea how the filtering will work," I start to wonder if this is really an attempt at implementing GraphQL at all...or just an excuse to hack around and re-factor the code as if there's nothing better to do. Life is too short and Lightning is too important to approach the task that way IMO. But maybe I'm missing something. If there are goals besides GraphQL that justify this, then by all means go for it, but probably better not to bother with GraphQL in that case and just focus on what is important.

rustyrussell:
> I guess this will give us the option of GraphQL syntactic sugar with very little of GraphQL's power.

I understand your frustration!  However, this approach gets us GraphQL on everything, fast, without forcing a rewrite and risk breakage of all our code.  That's about as good as any tradeoff gets!

The ability to filter and select fields in output *alone* is a concrete improvement.

Concretely, consider the json_close JSON generation code, vs. a table-driven method:
```
	bitcoin_txid(cc->channel->last_tx, &txid);

	json_add_tx(result, "tx", cc->channel->last_tx);
	json_add_txid(result, "txid", &txid);
	if (cooperative)
		json_add_string(result, "type", "mutual");
	else
		json_add_string(result, "type", "unilateral");
```

Now, if the caller doesn't want the txid we wasted time calculating it, and we call into json_add_txid() only to have it return immediately.  This is less efficient than it could be!  Similarly, if they don't want "type" we wasted time in a branch.

While we can optimize this with a new `json_field_wanted(result, "txid")`, we still have a cost in permuting the resulting tree to match the GraphQL query order.  This is not theoretically worse than the permuting of a field table, however, so it's just a practical issue.

> Did you consider that we might want to eventually implement mutations?

No I did not.  I'm learning GraphQL by reading your PR, so I had to google what mutations are!

However, I think this would be a second step anyway: our APIs are not really organized for mutation. e.g. changing the alias using getinfo?  That's a weird API name now; we'd want a new "info" for this.  Similarly for channel settings, we've got an API split between setters and getters which doesn't make sense, and would be cleanest with a new GraphQL-only API (perhaps we would formalize get/set pairs for non-graphql and convert those into GraphQL internally?  Still, a whole new API set).

> Did you consider that we might want to eventually auto-generate stub functions from the documentation XML schemas?

I don't think we will.  Our internal datastructures don't tend to map directly onto our API; there are enough differences that we would end with a hybrid approach, where some code would be inside the XML(?) and some elsewhere.  It would be hard to read, hard to maintain.

rl-d:
> > I guess this will give us the option of GraphQL syntactic sugar with very little of GraphQL's power.
> 
> I understand your frustration! However, this approach gets us GraphQL on everything, fast, without forcing a rewrite and risk breakage of all our code. That's about as good as any tradeoff gets!

Thank you for the rationale, I appreciate that. As one new to the project I am certainly sensitive to the risk of breaking things, and that is why in this PR I added the GraphQL implementation as separate logic to mitigate the risk of breaking things. I also don't like having to rewrite (or duplicate, in this case), but in essence that is where the benefit comes from (beyond the syntactic sugar). Regarding speed, I would personally implement things faster once I see acceptance, but I am reluctant to barrel ahead till the approach is accepted and the code is satisfactorily in line with the project's standards. Getting this first PR merged would be my cue that all is clear to go forward.

> The ability to filter and select fields in output _alone_ is a concrete improvement.

I disagree. The ability to filter and select in the output alone could (and arguably should) be done in a plugin...without risk, rewrite, or delay. In fact, there is already a GraphQL plugin in the plugins repo (written in Go) that might already do that. (Disclaimer: I haven't investigated exactly what that plugin does or if it's still working, but if to filter and select fields is all that's wanted, someone should definitely investigate that.) The only reason in my mind to get into this part of the code for GraphQL is to reap the benefits of full integration at the core level.

> Concretely, consider the json_close JSON generation code, vs. a table-driven method:

In my current approach, this is already about as efficient as it gets at run time, albeit with the noted rewrite/duplication of code. The table just defines what _can_ be requested, but only the selected fields are ever iterated over.

> > Did you consider that we might want to eventually implement mutations?
> 
> No I did not. I'm learning GraphQL by reading your PR, so I had to google what mutations are!

This is what I suspected and it is good so...that is why you need other team members: to offload the work of having to dive deep into every detail. But on the other hand, for that to work you have to trust that your helpers (me in this case) are doing their research well and looking out for the future and avoiding the traps.

> However, I think this would be a second step anyway: our APIs are not really organized for mutation. e.g. changing the alias using getinfo? That's a weird API name now; we'd want a new "info" for this. Similarly for channel settings, we've got an API split between setters and getters which doesn't make sense, and would be cleanest with a new GraphQL-only API (perhaps we would formalize get/set pairs for non-graphql and convert those into GraphQL internally? Still, a whole new API set).

I agree that this would be a second step, but the key observation here is that that second step will indeed eventually need a new API set, in order to flex the GraphQL muscle. Without that divergence, we only have "syntactic sugar" by definition.

Changing the alias and retrieving the updated node info at the same time would look something like this:
```
mutation changeAlias(newAlias: "NEWNAME") {
  info { id, alias, ... }
}
```
or just
```
changeAlias(newAlias: "NEWNAME") {
  info { id, alias, ... }
}
```
which is not too weird when you realize that it's basically a function call with the (optional) query afterward (re)using the same syntax as a normal query.

> > Did you consider that we might want to eventually auto-generate stub functions from the documentation XML schemas?
> 
> I don't think we will. Our internal datastructures don't tend to map directly onto our API; there are enough differences that we would end with a hybrid approach, where some code would be inside the XML(?) and some elsewhere. It would be hard to read, hard to maintain.

What I was mainly referring to here but didn't express well is the fact that the new tables (defined where GQLCB_TABLE_TYPES_DECL is called in the current iteration of this PR) could be auto-generated (along with the typedef for the new json_add_xxxxx() functions). I tend to agree with "hard to read, hard to maintain" but the table definitions are not completely intuitive either, and in this case I think it would make it easier to keep the documentation in sync with the GraphQL API, especially if/when the GraphQL API diverges significantly from the existing API. But it was just a thought... let's cross that bridge if/when we get to it.

Thanks for the dialogue. For my part, I'm open to further refinements if the code is still too ugly or whatever, but I don't want to spend time on a drastically different approach that I am convinced will not give us much potential (even if it's quicker to implement).

PR #4863, opened 2021-10-13
offers: fix overzealous BROKEN log (test flake)

rustyrussell:
sendonionmessage can fail when sending a reply, either because
the reply had a bad first peer, or because it went offline.  The
latter happens in CI, which is how I found this.

Also fixed typo "onio" -> "onion".


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4863/commits/6d8bc3bae494356fe1da49856ea22df541d4048c

great catch 🛩️

PR #4864, opened 2021-10-13
bolt1: init tlvs set remote address

m-schmoock:
This implements RFC https://github.com/lightningnetwork/lightning-rfc/pull/917 that
will add the IP `remote_addr` to init tlvs on incoming IP connections.
The other node may decide to use that for public IP detection behind a NAT.
Currently it only logs this information.

It has already been cross-tested with https://github.com/ACINQ/eclair/pull/1973

I think we should decouple this from the actual PR that does use that new information,
as several options are thinkable, and also because we can speed up cross testing this
way to get the RFC done quicker.

## Next Steps: What do we do when we receive a `remote_addr`?
 - Have n multiple peers or a minimal percentage of peers report the same `remote_addr` to consider it valid.
 - Or verify the `remote_addr` by making a test connection to ourself (can work, must not).
 - Or making a test conenction to a random node known from gossip to check if he also reports that address.
 - If a `remote_addr` considered valid and not yet known, insert this into an `node_annoucment` update, if:
   - No other remote_address of the same type (IPv4/IPv6) has been used that way.
   - If same type already exist, drop the one that is reported by fewer of out peers.


connectd/peer_exchange_initmsg.c
```diff
@@ -40,12 +40,27 @@ static bool contains_common_chain(struct bitcoin_blkid *chains)
...
+			u8 addr[4];
+			u16 port;
+		} ipv4;
+		struct {
+			u8 addr[16];
+			u16 port;
+		} ipv6;
+	} u;
+}__attribute__((packed));
+
```
rustyrussell, line 47: Nope.  You need to demashall properly; this makes unwarrented assumptions about the alignment of the packet, esp. the u16.

tests/test_connection.py
```diff
@@ -48,6 +48,9 @@ def test_connect(node_factory):
     assert len(l1.rpc.listpeers()) == 1
     assert len(l2.rpc.listpeers()) == 1
 
+    if EXPERIMENTAL_FEATURES:
+        l1.daemon.wait_for_log("Peer reports remote_addr: 127.0.0.1:[0-9]{5}")
```
rustyrussell, line 52: You can actually check this is correct, with l1.port?

rustyrussell: Maybe also change message to say "Peer tells us it sees our our address as "? which is clearer for normal users...

connectd/peer_exchange_initmsg.c
```diff
@@ -110,10 +114,24 @@ static struct io_plan *peer_init_received(struct io_conn *conn,
...
-				  remote_addr_any->type);
+		if (remote_addr_any->type == ADDR_TYPE_IPV4) {
+			remote_addr_v4 = (struct remote_addr_ipv4 *)remote_addr_any;
+			remote_addr_wi->itype = ADDR_INTERNAL_WIREADDR;
+			remote_addr_wi->u.wireaddr.type = ADDR_TYPE_IPV4;
+			remote_addr_wi->u.wireaddr.port = remote_addr_v4->port;
+			remote_addr_wi->u.wireaddr.addrlen = 4;
+			memcpy(&remote_addr_wi->u.wireaddr.addr,
+			       &remote_addr_v4->addr, 4);
+		}
```
rustyrussell, line 130: If you did proper marshal/demarshal you wouldn't have a buffer overrun here in the case where the buffer is malformed.

"remote_addr" is a weird name in general, it's "address peer sees me as", or something...

connectd/connectd_wire.csv
```diff
@@ -59,6 +59,7 @@ msgdata,connectd_connect_failed,addrhint,?wireaddr_internal,
 msgtype,connectd_peer_connected,2002
 msgdata,connectd_peer_connected,id,node_id,
 msgdata,connectd_peer_connected,addr,wireaddr_internal,
+msgdata,connectd_peer_connected,remote_addr,wireaddr_internal,
```
rustyrussell, line 62: Make this optional?  You can use `wireaddr_internal?`

connectd/peer_exchange_initmsg.c
```diff
@@ -206,6 +210,32 @@ struct io_plan *peer_exchange_initmsg(struct io_conn *conn,
...
+		if (addr->u.wireaddr.type == ADDR_TYPE_IPV6) {
+			remote_addr_v6 = tal(tlvs, struct remote_addr_ipv6);
+			remote_addr_v6->type = 2;
+			remote_addr_v6->port = addr->u.wireaddr.port;
+			memcpy(&remote_addr_v6->addr,
+			       addr->u.wireaddr.addr, 16);
+			tlvs->remote_addr = (u8*)remote_addr_v6;
+		}
+	}
```
rustyrussell, line 266: You need to check this is a public address, otherwise you're telling them about your internal network gateway.  See address_routable...

connectd/connectd_wire.csv
rustyrussell: Actually, make that `wireaddr?` since this isn't an internal thing.

rustyrussell (dismissed):

connectd/peer_exchange_initmsg.c
m-schmoock: Yep will do the demarshalling functions. Regarding the name: I called it this way because the RFC also uses this name: https://github.com/lightning/bolts/pull/917  which makes sense from the protocol viewpoint, but not so much from the receiving peer.

connectd/connectd_wire.csv
m-schmoock: Yes, I will change it to wireaddr. Already had that in mind

connectd/peer_exchange_initmsg.c
m-schmoock: I can add a `address_routeable()` check but allow localhost for testing. Actually a prior version of the PR had `address_routeable()` in it, but I removed it as part of the discussion in the RFC. Maybe we should add a line to the RFC that says "SHOULD NOT send non-public addresses for remote_addr".

m-schmoock: @rustyrussell See latest discussion on https://github.com/lightning/bolts/pull/917#issuecomment-1007311216

For a normal home NAT router this should not happen as a a node should see the public address, right?

m-schmoock: Update: I added `SHOULD NOT set private addresses as 'remote_addr'.` to the RFC and will add a check in the code but allow for localhost.

connectd/peer_exchange_initmsg.c
```diff
@@ -177,6 +238,34 @@ struct io_plan *peer_exchange_initmsg(struct io_conn *conn,
...
+			tlvs->remote_addr = (u8*)remote_addr_v4;
+		}
+		if (addr->u.wireaddr.type == ADDR_TYPE_IPV6) {
+			remote_addr_v6 = tal(tlvs, struct remote_addr_ipv6);
+			remote_addr_v6->type = 2;
+			remote_addr_v6->port = addr->u.wireaddr.port;
+			memcpy(&remote_addr_v6->addr,
+			       addr->u.wireaddr.addr, 16);
+			tlvs->remote_addr = (u8*)remote_addr_v6;
+		}
```
vincenzopalazzo, line 265: is it better use the switch case here? and abort in case of not match?

From the spec https://github.com/lightning/bolts/pull/917/commits/efe9d2d151b80b1992f0ed65fead2b36d2908721 I understand that here can be possible only two type of values

vincenzopalazzo (commented):
LGTM, just a small comment

tests/test_connection.py
m-schmoock: I can't check for l1 listening port, because we see the source port of the outbound connection which is dynamically assigned by the kernel. I added a question to the RFC about that: https://github.com/lightning/bolts/pull/917#issuecomment-1008276220

m-schmoock: Changed the wording though. Keep in mind that `remote_addr` is a protocol field.

m-schmoock:
@rustyrussell I now use `wireaddr.h` for marshalling, as the RFC will use the same data structure for this field. This leads to a lot of mock/include changes. If you want we could split up `wireaddr.h` into `wireaddr.h` (actual structure and wire parsing) and `wireaddr_internal.h` (the _internal struct) and `wireaddr_utils.h` (also the utils like `is_toraddr` ...)

m-schmoock:
Just a testflake in tests/test_plugin.py::test_coin_movement_notices

m-schmoock:
Note: Bolt https://github.com/lightning/bolts/pull/917 has been merged yesterday.
Cross testing has been made with this branch, not sure though if that's the perfectly correct order.

m-schmoock:
suqashed some commits and removed unused struct.

m-schmoock:
Resolved conflicts and rebased on current master

connectd/connectd.c
```diff
@@ -264,6 +267,9 @@ static struct io_plan *peer_reconnected(struct io_conn *conn,
 	pr->id = *id;
 	pr->cs = *cs;
 	pr->addr = *addr;
+	pr->remote_addr = NULL;
+	if (remote_addr != NULL)
+		pr->remote_addr = remote_addr;
```
cdecker, line 272: Isn't a direct assignment pretty much the same here?

lightningd/peer_control.c
```diff
@@ -1151,6 +1157,14 @@ void peer_connected(struct lightningd *ld, const u8 *msg, int peer_fd)
 	if (!hook_payload->channel)
 		hook_payload->channel = peer_unsaved_channel(peer);
 
+	/* Log remote_addr for now */
+	if (hook_payload->remote_addr && (
+	    hook_payload->remote_addr->type == ADDR_TYPE_IPV4 ||
+	    hook_payload->remote_addr->type == ADDR_TYPE_IPV6)) {
+		log_info(ld->log, "Peer tells it sees our address as: %s",
```
cdecker, line 1164: `tells` -> `says` :-)

cdecker (approved):
ACK bcce20f4e2346387a6b024dd27c9d131b8d8aa84

connectd/connectd.c
m-schmoock: yup, good catch. don't know how that got in there :D

m-schmoock:
I just force pushed @cdecker 's changes/remarks

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/7baae67c706c9678e23e651f143c4d41cf5ad9ec

Issue #4860, opened 2021-10-12
`litestream` recommendation may lead to crashes?


ZmnSCPxj:
In `BACKUP.md` we have a recommendation to use `litestream`.  However, I snuck a look at the [official `litestream` online docs](https://litestream.io/how-it-works/#the-shadow-wal):

> Litestream works by effectively taking over the checkpointing process. It starts a long-running read transaction to prevent any other process from checkpointing and restarting the WAL file.

In filesystem terms, that "read transaction" is an advisory shared lock on the database file.  When the SQLITE instance running inside `lightningd` wants to write to the database file (and we write ***all*** the time, you should see the blinkenlights on the SSD on my Pi), however, it demands an advisory exclusive lock.

SQLITE will auto-retry the advisory lock for some time (I believe we actually explicitly set this timeout somewhere...? need to dig).  However after a while it will quit with "database is locked" and then `lightningd` kills itself.

Note that in `BACKUP.md` we recommend *against* using `sqlite3` `.dump` or `VACUUM INTO` (the last paragraph).  Litestream is actually similar to an efficient `.dump` (it only copies (part of) the WAL file instead of dumping the entire database), but this is a difference in quantity and not quality: the backup process read-locks the file, and we write-lock very often, and on slow HDDs or high-load systems the timeout on the write-lock retry can be reached.  If we disrecommend `sqlite3` `.dump` I think we should disrecommend against `litestream` as well (or if the quantity is fairly low enough, maybe put up warnings at the very least, and move `litestream` to *after* the "copy the file" section).

Issue #4857, opened 2021-10-11
crashes with "database is locked"


x3nb63:
### Issue and Steps to Reproduce

version is 0.10.1

but 0.9.3 did the same, maybe more often.

`lightningd` emergency exits itself cause the database says its locked during INSERT or UPDATE statements:

```
$ grep 'database is locked' crash.log.*
crash.log.20211009101318:+498.816259929 lightningdBROKEN: Error executing statement: wallet/wallet.c:3786: INSERT INTO transactions (  id, blockheight, txindex, rawtx) VALUES (?, ?, ?, ?);: database is locked
crash.log.20211009175028:+27421.465744630 lightningdBROKEN: Error executing statement: wallet/wallet.c:768: UPDATE shachains SET num_valid=?, min_index=? WHERE id=?: database is locked
crash.log.20211009175054:+17.143080552 lightningdBROKEN: Error executing statement: wallet/wallet.c:3540: INSERT INTO blocks (height, hash, prev_hash) VALUES (?, ?, ?);: database is locked
crash.log.20211009222937:+16702.140267710 lightningdBROKEN: Error executing statement: wallet/wallet.c:768: UPDATE shachains SET num_valid=?, min_index=? WHERE id=?: database is locked
crash.log.20211010034113:+18687.834439849 lightningdBROKEN: Error executing statement: wallet/wallet.c:768: UPDATE shachains SET num_valid=?, min_index=? WHERE id=?: database is locked
crash.log.20211011032432:+85389.796158637 lightningdBROKEN: Error executing statement: wallet/wallet.c:768: UPDATE shachains SET num_valid=?, min_index=? WHERE id=?: database is locked
```

Don't know how to reproduce. It happens few times a day, presumably related to how busy the node is, but that is a guess already - I failed to trigger it.

### My Digging

Looking generally I find SQLite has [a story](https://www.arysontechnologies.com/blog/fix-sqlite-error-database-locked/) with locks [or here](https://www.convertertools.org/blog/sqlite-database-is-locked-error-code-5/) - I can not say whether this applies.

Naturally I suspect _"race-condition because of threads"_  - but can't say anything either, not knowing lightningd's thread model...

On the hardware I can say its rather slow: 2 x 7200rpm HDD combined as zpool mirror. That is low latency compared to SSD. So, if lightningd is rarely used / never stress-tested on such disks, I guess such a thing can go unnoticed.

ZmnSCPxj:
[...]

PR #4867, opened 2021-10-15
Change `sqlite3_busy_timeout` to 60 seconds, add warnings re: `litestream`.

ZmnSCPxj:
Fixes: #4860 

Fixes: #4857 if @x3nb63 can confirm

We need to specify versions in `BACKUP.md` since it is published on https://lightning.readthedocs.io/BACKUP.html and users might assume the recommendation applies to all versions; `litestream` will reliably work after 5bf2b5c0ba **and** on reasonably fast media/low-load conditions.

### DIFF
--- a/doc/BACKUP.md
+++ b/doc/BACKUP.md
@@ -242,31 +242,6 @@ three or four storage devices.
 BTRFS would probably work better if you were purchasing an entire set
 of new storage devices to set up a new node.
 
-## SQLite Litestream Replication
-`/!\` WHO SHOULD DO THIS: Casual users
-
-One of the simpler things on any system is to use Litestream to replicate the SQLite database.
-It continuously streams SQLite changes to file or external storage - the cloud storage option 
-should not be used. 
-Backups/replication should not be on the same disk as the original SQLite DB.
-
-/etc/litestream.yml : 
-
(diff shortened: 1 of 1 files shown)


wallet/db_postgres.c
```diff
@@ -1,5 +1,6 @@
 #include "db_postgres_sqlgen.c"
 #include <ccan/ccan/tal/str/str.h>
+#include <ccan/compiler/compiler.h>
```
vincenzopalazzo, line 3: unable to understand the motivation that we need the `ccan/compiler` here

ZmnSCPxj: `UNUSED` macro.

rustyrussell:
Let's just increase the default to 30 or 60 seconds. I mean, we could do better, but the user surely prefers a delay to a crash!

Providing knobs like this is a last resort, IMHO.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4867/commits/2434e4d67f14f9261203a5e16482da2102cc17e5

rustyrussell (approved):
Ack 2434e4d67f14f9261203a5e16482da2102cc17e5

PR #4872, opened 2021-10-19
Use `closefrom` and remove our fds max limit.

### LINKED ISSUE #4868: **BROKEN** Could not subdaemon channel: Too many open files

My node was exhibiting some erroneous behavior due to failing to launch subdaemons due to hitting the open file descriptor resource limit. I tried raising the limit using `ulimit` as suggested by @cdecker in https://github.com/ElementsProject/lightning/issues/4534#issuecomment-845204561, but that had no effect. (The `lightningd` process still had both soft and hard limits of 1024 open files.) Come to find out, these limits are actually [hardcoded in `lightning/lightning.c`](https://github.com/ElementsProject/lightning/blob/v0.10.1/lightningd/lightningd.c#L863).

Please implement a configuration option to override this insufficient limit.

> [“640KB ought to be enough for anybody.” —Christian Decker](https://github.com/ElementsProject/lightning/commit/07e6f43e1da13975074bb84e63cb4bc0cdc82e3a)

Thanks to @TheBlueMatt for bringing the failure to my attention.

### `getinfo` output
This is C-Lightning 0.10.1.

ZmnSCPxj:
Fixes #4868

This also inadvertently fixes a latent bug: before this patch, in the 
`subd` function in `lightningd/subd.c`, we would close `execfail[1]`
*before* doing an `exec`.
We use an EOF on `execfail[1]` as a signal that `exec` succeeded (the
fd is marked CLOEXEC), and otherwise use it to pump `errno` to the 
parent.
The intent is that this fd should be kept open until `exec`, at which
point CLOEXEC triggers and close that fd and sends the EOF, *or* if
`exec` fails we can send the `errno` to the parent process vua that
pipe-end.

However, in the previous version, we end up closing that fd *before*
reaching `exec`, either in the loop which `dup2`s passed-in fds (by 
overwriting `execfail[1]` with a `dup2`) or in the "close everything"
loop, which does not guard against `execfail[1]`, only
`dev_disconnect_fd`.

Ping @whitslack.

Also ping @NicolasDorier and @cdecker --- on Linux < 5.9, this new code uses `/proc/$$/fd` to see the open file descriptors (and avoid looping all the way to 1048576), but I do not know if `/proc` is limited or inaccessible within Docker, and the original #2977 was triggered by Docker on root.  Cursory search *suggests* it should be accessible but I would prefer expert opinion.

We can argue that we can just "kick the can" and bump up the max limit to 4096, but at some point @whitslack is going to have 10,000 channels all by himself, and besides `closefrom` is "the future" and should be implementable on more systems at some point, maybe.

Created as draft for now; in particular the code is only minimally tested on Linux 5.11 (and I tried disabling `HAVE_NR_CLOSE_FROM` to test the `/proc` scanning, which should work on a lot of systems that are not running `lightningd` in a `chroot` jail).  It should be tested on more systems and container types, and while I can try to figure out how to get the tests running on FreeBSD in a VM, I do not know if I can legally run a MacOS VM if I refuse to purchase anything from Apple.


ccan/ccan/closefrom/_info
```diff
@@ -0,0 +1,65 @@
...
+ *
+ *	if (closefrom_may_be_slow()) {
+ *		// If being emulated, then we might end up
+ *		// looping over a large _SC_OPEN_MAX
+ *		// (Some systems have it as INT_MAX!)
+ *		// If so, closefrom_may_be_slow() will return
+ *		// true and we should keep _SC_OPEN_MAX down
+ *		// by the below:
+ *		struct rlimit nofile = {1024, 1024};
+ *		setrlimit(RLIMIT_NOFILE, &nofile);
```
rustyrussell, line 36: I think this logical should all be inside the ccan module: "closefrom_limit()" perhaps?  And it should *check* _SC_OPEN_MAX and save it at this point (before any other changes), if it's > 64k, say, then call RLIMIT.

Anyway, I prefer it returning "bool" since all compilers support that now..

rustyrussell (approved):

rustyrussell:
I think increasing to 4k in the worst case is also the Right Thing, which means this won't make things worse,

I say move it out of draft (with the simple fixes preferably) and we use the -rcs for further testing...

ZmnSCPxj:
Right right, `closefrom` returns `void` and I was doing `int res = closefrom(STDERR_FILENO + 1);` in the configurator.  Hope the FreeBSD smoke test shows `closefrom` is detected now!

Also did the `closefrom_limit` as suggested, and use `bool` everywhere it made sense.  `closefrom_may_be_slow` is still exposed so we can print something to logs.

ZmnSCPxj:
Configurator test on FreeBSD now works! https://github.com/ElementsProject/lightning/pull/4872/checks?check_run_id=3958492505#step:3:4385

Will bring out of draft.

rustyrussell:
Included in upstream ccan/ trivial rebase.

ZmnSCPxj:
CI failure seems unrelated, it is pointing at `hsmtool`, which should not be affected by this (does `hsmtool` use `pipecmd`?).  Is there a way to restart just one test?  The only button I can find seems to restart all of them.

I have code now to implement the fix to the *really* edgy edge case in the previous comment, but am wary of adding it currently, because I think we will not hit that edge case in practice anyway (because if so `lightningd` would already be currently failing a lot).  Maybe.  Hmmmmmm.

rustyrussell:
> > Is there a way to restart just one test? The only button I can find seems to restart all of them.
> 
> yep, I think there is any method to restart only one check

No, you have to restart the entire thing :(  known GH actions limitation it seems...

NicolasDorier:
`/proc` seems to work fine inside containers

ZmnSCPxj:
@NicolasDorier thanks!

PR #4874, opened 2021-10-21
Makefile: Avoid overriding `PYTHONPATH`.

ZmnSCPxj:
On some distributions (e.g. Gnu Guix) Python packages are not installed in
some standard directory, rather they are installed in different places and
the `PYTHONPATH` variable is modified to include the different places.
So, we must not use the name `PYTHONPATH` in our `Makefile` since `make`
will replace the `PYTHONPATH` environment variable, preventing e.g.
`tools/generate-wire.py` from finding `python-mako` installed on such
distributions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ endif
 
 # (method=thread to support xdist)
 PYTEST_OPTS := -v -p no:logging $(PYTEST_OPTS)
-PYTHONPATH=$(shell pwd)/contrib/pyln-client:$(shell pwd)/contrib/pyln-testing:$(shell pwd)/contrib/pyln-proto/:$(shell pwd)/external/lnprototest:$(shell pwd)/contrib/pyln-spec/bolt1:$(shell pwd)/contrib/pyln-spec/bolt2:$(shell pwd)/contrib/pyln-spec/bolt4:$(shell pwd)/contrib/pyln-spec/bolt7
+MY_CHECK_PYTHONPATH=$${PYTHONPATH}$${PYTHONPATH:+:}$(shell pwd)/contrib/pyln-client:$(shell pwd)/contrib/pyln-testing:$(shell pwd)/contrib/pyln-proto/:$(shell pwd)/external/lnprototest:$(shell pwd)/contrib/pyln-spec/bolt1:$(shell pwd)/contrib/pyln-spec/bolt2:$(shell pwd)/contrib/pyln-spec/bolt4:$(shell pwd)/contrib/pyln-spec/bolt7
 # Collect generated python files to be excluded from lint checks
 PYTHON_GENERATED=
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -80,7 +80,7 @@ endif
 
 # (method=thread to support xdist)
 PYTEST_OPTS := -v -p no:logging $(PYTEST_OPTS)
-PYTHONPATH=$(shell pwd)/contrib/pyln-client:$(shell pwd)/contrib/pyln-testing:$(shell pwd)/contrib/pyln-proto/:$(shell pwd)/external/lnprototest:$(shell pwd)/contrib/pyln-spec/bolt1:$(shell pwd)/contrib/pyln-spec/bolt2:$(shell pwd)/contrib/pyln-spec/bolt4:$(shell pwd)/contrib/pyln-spec/bolt7
+MY_CHECK_PYTHONPATH=$(shell pwd)/contrib/pyln-client:$(shell pwd)/contrib/pyln-testing:$(shell pwd)/contrib/pyln-proto/:$(shell pwd)/external/lnprototest:$(shell pwd)/contrib/pyln-spec/bolt1:$(shell pwd)/contrib/pyln-spec/bolt2:$(shell pwd)/contrib/pyln-spec/bolt4:$(shell pwd)/contrib/pyln-spec/bolt7
```
cdecker, line 83: Wouldn't augmenting the `PYTHONPATH` instead not be better?

```suggestion
PYTHONPATH=${PYTHONPATH}:$(shell pwd)/contrib/pyln-client:$(shell pwd)/contrib/pyln-testing:$(shell pwd)/contrib/pyln-proto/:$(shell pwd)/external/lnprototest:$(shell pwd)/contrib/pyln-spec/bolt1:$(shell pwd)/contrib/pyln-spec/bolt2:$(shell pwd)/contrib/pyln-spec/bolt4:$(shell pwd)/contrib/pyln-spec/bolt7
```

This way we could keep OS specific paths in the search, overriding it when we actually need it.

ZmnSCPxj: A simplified example seems to fail though:

```
$ cat Makefile
PYTHONPATH=${PYTHONPATH}:/example

.PHONY: show-pythonpath

show-pythonpath:
        PYTHONPATH=$(PYTHONPATH) echo $$PYTHONPATH
$ make show-pythonpath
Makefile:1: *** Recursive variable 'PYTHONPATH' references itself (eventually).  Stop.
```

Seems just overall safer to outright use a completely different name.

ZmnSCPxj: Okay, so `echo $PYTHONPATH` is wrong since the `$PYTHONPATH` gets expanded "early", but still:

```
$ cat Makefile
PYTHONPATH=$$PYTHONPATH:/example

.PHONY: show-pythonpath

show-pythonpath:
        PYTHONPATH=$(PYTHONPATH) env | grep PYTHONPATH
$ make show-pythonpath 
PYTHONPATH=$PYTHONPATH:/example env | grep PYTHONPATH
PYTHONPATH=$PYTHONPATH:/example:/example
```

ZmnSCPxj: This one seems to work out better, and works compatibly on systems which do *not* set `PYTHONPATH`, either:

```$ cat Makefile
MY_CHECK_PYTHONPATH=$${PYTHONPATH}$${PYTHONPATH:+:}/example

.PHONY: show-pythonpath

show-pythonpath:
        PYTHONPATH=$(MY_CHECK_PYTHONPATH) env | grep PYTHONPATH
$ make show-pythonpath
PYTHONPATH=${PYTHONPATH}${PYTHONPATH:+:}/example env | grep PYTHONPATH
PYTHONPATH=/home/zmnscpxj/.guix-profile/lib/python3.8/site-packages:/example
$ PYTHONPATH= make show-pythonpath
PYTHONPATH=${PYTHONPATH}${PYTHONPATH:+:}/example env | grep PYTHONPATH
PYTHONPATH=/example
```

ZmnSCPxj:
Test failure seemed unrelated, kicked the CI.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/894e9e89e872b5da2cee273259f7bcd9dee6047f

PR #4876, opened 2021-10-21
Return the correct `channel_update` when erring

cdecker:
This is a spinoff from #4808 which includes the fix for outdated `channel_updates` when one is already queued for broadcast.


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4876/commits/18d18ded2ea29bcfebf05c5abb890e38f9739a71

PR #4877, opened 2021-10-22
Cleanup fd shuffle

rustyrussell:
Builds on #4872 

Changelog-None


lightningd/subd.c
```diff
@@ -19,53 +19,72 @@
 #include <wire/common_wiregen.h>
 #include <wire/wire_io.h>
 
-static bool move_fd(int from, int to)
+/* Carefully move fd *@from to @to: on success *from set to to */
+static bool move_fd(int *from, int to)
```
ZmnSCPxj, line 23: Was thinking of doing this as well.

lightningd/subd.c
```diff
@@ -191,65 +210,32 @@ static int subd(const char *path, const char *name,
...
-		if (childmsg[1] != STDIN_FILENO) {
-			/* Do we need to move STDIN out the way? */
-			stdin_is_now = dup(STDIN_FILENO);
-			if (!move_fd(childmsg[1], STDIN_FILENO))
-				goto child_errno_fail;
-		}
-
-		/* Dup any extra fds up first. */
+		/* msg = STDIN */
+		fds[0] = childmsg[1];
```
ZmnSCPxj, line 221: How about `STDOUT_FILENO` and `STDERR_FILENO`?  The old version started the passed-in `fd`s at index 3 (xref `fdnum`) aka `STDERR_FILENO + 1`.

lightningd/subd.c
```diff
@@ -191,65 +210,32 @@ static int subd(const char *path, const char *name,
...
+		if (dev_disconnect_fd != -1)
+			tal_arr_expand(&fds, dev_disconnect_fd);
+
+		/* Finally, the fd to report exec errors on */
+		tal_arr_expand(&fds, execfail[1]);
 
-		/* Move execfail[1] *after* the fds we will pass
-		 * to the subdaemon.  */
-		if (!move_fd(execfail[1], fdnum))
+		if (!shuffle_fds(fds, tal_count(fds)))
```
ZmnSCPxj, line 234: This changes the `fd` number of `execfail[i]` and `dev_disconnect_fd`, but does not seem to update those variables?  Maybe `shuffle_fds` should accept an array of `int *` instead of an array of `int`?

lightningd/subd.c
```diff
@@ -19,53 +19,72 @@
...
+		int in_the_way;
+
+		/* Already in the right place?  Great! */
+		if (fds[i] == i)
+			continue;
+		/* Is something we care about in the way? */
+		in_the_way = fd_used(fds + i, num_fds - i, i);
+		if (in_the_way != -1) {
+			/* Find a high-numbered unused fd. */
+			while (fd_used(fds + i, num_fds - i, next_free_fd) != -1)
```
ZmnSCPxj, line 75: `fd_used` is O(n), and this `while` loop potentially loops up to n times as well, and all this is in a loop that is called n times, too.

How about just using `dup`?  Presumably the OS has a fast O(1) or O(log n) structure for looking up file descriptor numbers.  i.e. retain `move_fd_any` and do something like:

```c
if (in_the_way != -1)
    /* Trick: in_the_way is offset by i! */
    move_fd_any(&fds[i + in_the_way]);
```

This *could* move the fd to another low index, but we will do another check on the next loop iteration anyway.  Assuming the OS has an O(1) structure for `dup`, then this drops the loop down to O( n^2 ) instead of O( n^3 ).  Though that may not matter overmuch since we do not expect to pass a lot of `fd`s around...

rustyrussell: N is so low here that I don't think it's worth *any* optimization.  Simplest wins.

rustyrussell:
Fixed from @ZmnSCPxj feedback and rebased...

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4877/commits/a7d35c46e4a1b614bc7798ce7e86342daa23bc80

ZmnSCPxj (approved):
ACK a7d35c46e4a1b614bc7798ce7e86342daa23bc80

Just needs a `ChangeLog-None`.

PR #4881, opened 2021-10-23
pytest: fix test_pluginconnected_hook_chaining

m-schmoock:
The last line of the testcase was checking on the wrong node l3
instead of l1. l3 didn't had the plugins configured that would
produce the log entry we were looking for not to be present.

Changelog-None

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -448,7 +448,7 @@ def test_plugin_connected_hook_chaining(node_factory):
         return peers == [] or not peers[0]['connected']
 
     wait_for(check_disconnect)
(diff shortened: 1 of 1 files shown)


rustyrussell:
Ack 467a91031

Rebase onto master, should fix flakes.

rustyrussell:
Another rebase, This Time For Sure!

Ack f9ecefa05

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4881/commits/f9ecefa05fbf03e5da4a223fa6a7249fcc0f47fb

PR #4893, opened 2021-10-28
Guess to remote: support anchor channel commitments

niftynei:
A user recently had a catastrophic loss of their backups and in the process of working through recovery of funds discovered that they were unable to recover funds from 'dual-funded channels'.

This is ultimately because we use a different to-remote script, as these are anchored by default.

### DIFF
--- a/bitcoin/psbt.c
+++ b/bitcoin/psbt.c
@@ -4,6 +4,7 @@
 #include <bitcoin/psbt.h>
 #include <bitcoin/pubkey.h>
 #include <bitcoin/script.h>
+#include <bitcoin/shadouble.h>
 #include <ccan/ccan/array_size/array_size.h>
 #include <ccan/ccan/mem/mem.h>
 #include <ccan/tal/str/str.h>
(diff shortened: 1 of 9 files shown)


endothermicdev (commented):
-

tests/test_wallet.py
```diff
@@ -1195,6 +1195,43 @@ def test_hsmtool_dump_descriptors(node_factory, bitcoind):
...
+    # unilateral close channels l1<->l2
+    l1.stop()
+    l2.rpc.close(l1.info['id'], 1)
+
+    # Wait til to_self_delay expires, l1 should claim to_local back
+    bitcoind.generate_block(10, wait_for_mempool=1)
+    sync_blockheight(bitcoind, [l2])
+    log = l2.daemon.wait_for_log(r'Script to-them: ')
+    m = re.search(r'Script to-them: ([a-f0-9]{44})', log)
+    onchain_addr = bitcoind.rpc.decodescript(m.group(1))['address']
```
endothermicdev, line 1221: Test is failing with a keyError on line 1221. Dict returned looks like:

`{'asm': '0 [error]', 'type': 'nonstandard', 'p2sh': '2MuBEKJ9vRmy64SSSRheWzf7np7Tcjj1Zgw', 'segwit': {'asm': '0 29e481065c0f83c94cec07ee31c833471c151989d405e77a0c39812048931bb5', 'hex': '002029e481065c0f83c94cec07ee31c833471c151989d405e77a0c39812048931bb5', 'address': 'bcrt1q98jgzpjup7pujn8vqlhrrjpnguwp2xvf6sz7w7sv8xqjqjynrw6sxvm489', 'type': 'witness_v0_scripthash', 'p2sh-segwit': '2NBGWnH9LN5ZqdQ5sU1QPdkjfEhM1Wfb7kS'}}`

should this be `bitcoind.rpc.decodescript(m.group(1))['segwit']['address']`?

endothermicdev: hsmtool invokes okay with this change:
`DEBUG:root:Starting 'tools/hsmtool guesstoremote bcrt1q98jgzpjup7pujn8vqlhrrjpnguwp2xvf6sz7w7sv8xqjqjynrw6sxvm489 022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59 100 /tmp/ltests-_vq9dgd1/test_hsmtool_guess_to_remote_non_anchor_1/lightning-1/regtest/hsm_secret'`
But returns
`Could not find any basepoint matching the provided witness programm.
Are you sure that the channel used option_static_remotekey ?`

bitcoin/psbt.c
```diff
@@ -441,6 +442,47 @@ bool psbt_input_set_redeemscript(struct wally_psbt *psbt, size_t in,
...
+       u64 input_val_sats;
+       struct amount_sat input_amt;
+       int flags = WALLY_TX_FLAG_USE_WITNESS;
+
+       input_amt = psbt_input_get_amount(psbt, in);
+       input_val_sats = input_amt.satoshis; /* Raw: type conversion */
+
+       /* Wally can allocate here, iff tx doesn't fit on stack */
+       tal_wally_start();
+       if (is_elements(chainparams)) {
```
jgriffiths, line 459: This should use `wally_psbt_get_input_signature_hash` instead of calculating the hash itself from psbt internals. In particular, the wally call handles taproot for you, which this doesn't, and it will correctly match missing sighashes to `SIGHASH_DEFAULT`/`SIGHASH_ALL` for you (this code will just produce the wrong hash).

tools/recover_anchor.c
```diff
@@ -0,0 +1,117 @@
...
+	wally_tx_witness_stack_add(stack, witness_script,
+				   tal_bytelen(witness_script));
+
+	wally_psbt_input_set_final_witness(&psbt->inputs[psbt_input], stack);
+	tal_wally_end(psbt);
+	/* Calls tal_wally internally */
+	psbt_input_set_witscript(psbt, psbt_input, witness_script);
+
+	/* Set the sequence number to match the csv */
+	psbt->tx->inputs[psbt_input].sequence = sequence;
```
jgriffiths, line 110: Only v0 psbts will have a `tx` member. For v2 this should be `wally_psbt_input_set_sequence` (ideally wally would handle this difference internally, I'll fix that in wally soon).

rustyrussell:
Closing this under the "too old" rule...  Feel free to fix up and reopen!

Issue #4883, opened 2021-10-24
plugins: hooks are removed too early in shutdown


SimonVrouwe:
This is current master branch 09459a97c7df not in anything released yet.

From the [documentation](https://lightning.readthedocs.io/PLUGINS.html)
> Hooks are a primitive that allows plugins to be notified about internal events in lightningd and alter its behavior or inject custom behaviors.

a plugin can be marked [important](https://lightning.readthedocs.io/lightningd-config.5.html#lightning-plugins) 
> important-plugin=PLUGIN Speciy a plugin to run as part of C-lightning. This can be specified multiple times to add multiple plugins. Plugins specified via this option are considered so important, that if the plugin stops for any reason (including via lightning-plugin(7) stop), C-lightning will also stop running. This way, you can monitor crashes of important plugins by simply monitoring if C-lightning terminates. Built-in plugins, which are installed with lightningd(8), are automatically considered important.

This _custom behavior_ injected by an important plugin should be maintained also during shutdown, as long as any relevant _internal events_ can happen. This is currently not the case because the shutdown of plugin's (even important ones) silently removes their hooks, but an associated internal event may happen because subdaemons are still running.

Below a simple test to demonstrate (or see this [branch](https://github.com/ElementsProject/lightning/commit/835174898e6873b65db2cadeba7f26548aa67cff))
```
def test_htlc_accepted_hook_shutdown(node_factory, executor):
    """Hooks semantics of important-plugins should be maintained until the very end
    """
    l1, l2 = node_factory.line_graph(2, opts=[
        {'may_reconnect': True, 'log-level': 'info'},
        {'may_reconnect': True, 'log-level': 'debug',
         'plugin': [os.path.join(os.getcwd(), 'tests/plugins/delay_shutdown.py')],
         'important-plugin': [os.path.join(os.getcwd(), 'tests/plugins/fail_htlcs.py')]}
    ])

    i1 = l2.rpc.invoice(msatoshi=1000, label="inv1", description="desc")['bolt11']

    # fail_htlcs.py makes payment fail
    with pytest.raises(RpcError):
        l1.rpc.pay(i1)

    f_stop = executor.submit(l2.rpc.stop)
    l2.daemon.wait_for_log(r'plugin-delay_shutdown.py: delaying shutdown with 10s')

    # Should still fail htlc while shutting down
    with pytest.raises(RpcError):
        l1.rpc.pay(i1)
```
the `delay_shutdown.py`:
```
[...]

Issue #4785, opened 2021-09-13
lightningd question: Can we expect db_write's after a "shutdown" notification?


SimonVrouwe:
While working on the `backup` plugin [here](https://github.com/lightningd/plugins/pull/296) to use the "shutdown" notification for cleanup, a questions arose: Can there be `db_write`'s after the "shutdown" notification?

The code below suggest it can:
https://github.com/ElementsProject/lightning/blob/cdb93eefec597b7fa42dc9ef175a6412caa5ce95/lightningd/lightningd.c#L1150-L1159

The separate freeing of `ld->jsonrpc` inside a db transaction was added long time ago in PR #1755 for a particular case of un-reserving the utxos (in destructor `destroy_utxos`) of a fundchannel command that didn't return when the remote peer hanged.

But since `v0.9.1` (PR #3867) that destructor (`destroy_utxos`) is not used anymore, AFAIU utxos are now unreserved by blockheight.

Are there other destructors hanging on `ld->jsonrpc` that do db transaction? If not, then PR #1755 can be reverted?

Maybe the [documentation](https://lightning.readthedocs.io/PLUGINS.html#shutdown) can be more explicit what plugins can expect, such as no rpc calls (e.g. `datastore`) after "shutdown".

vincenzopalazzo:
> Can there be db_write's after the "shutdown" notification?

Yes, lightning doesn't wait for an answer here, so at this point, you can receive all the types of sequences, and it is not deterministic at all. The `shutdown_subdaemons` will make a transaction on db, as you correctly said.

>Maybe the documentation can be more explicit what plugins can expect, such as no rpc calls (e.g. datastore) after "shutdown".

I'm wondering if we need to admit all the types of calls during the "shutdown", I mean, if we have a datatore, in my opinion, the plugin needs to have time to store the data (but not much time here, imagine that the plugin have a while true {}, it is not good at all).

SimonVrouwe:
**TLDR**; I think plugins that subscribe to `db_write` hook and are `important-plugin` should be killed separately, after all the other ones (and their callbacks handled), i.e. after the last potential db write.

[...]

PR #4897, opened 2021-11-01
lightningd: make shutdown smoother and safer

SimonVrouwe:
Another proof of concept (should I make this a draft PR?), to address issues raised in #4785 #4790 and  #4883 

The point of shutdown is to break-down or reduce activity. Starting a new `io_loop` in order to write `shutdown` notifications to plugins and wait for them to terminate, should not trigger new activity other then logging and (maybe?) notifications.

So before restarting this `io_loop` we shutdown subdaemons and disable handling of all JSON RPC requests and responses with an "id" field in it. Hook callback's will abandon already and without subdaemons or JSON RPC no new hooks can be called. All of that should prevent any `db_write`'s and we can safely start the `io_loop` and shutdown all plugins.

Also cleans up some dead code related to unused `destroy_utxos` destructor.

TODO:
- [x] documentation

### DIFF
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -1168,6 +1168,8 @@ int main(int argc, char *argv[])
 	 *  shut down.
 	 */
 	assert(io_loop_ret == ld);
+
+	/* Fail JSON RPC requests and ignore plugin's responses */
 	ld->state = LD_STATE_SHUTDOWN;
 
 	stop_fd = -1;
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -757,13 +757,14 @@ static int setup_sig_handlers(void)
 
 /*~ This removes the SIGCHLD handler, so we don't try to write
  * to a broken pipe. */
-static void remove_sigchild_handler(void)
+static void remove_sigchild_handler(struct io_conn *sigchld_conn)
 {
 	struct sigaction sigchild;
 
 	memset(&sigchild, 0, sizeof(struct sigaction));
 	sigchild.sa_handler = SIG_DFL;
 	sigaction(SIGCHLD, &sigchild, NULL);
+	io_close(sigchld_conn);
 }
 
 /*~ This is the routine which sets up the sigchild handling.  We just
@@ -854,6 +855,7 @@ int main(int argc, char *argv[])
 	struct htlc_in_map *unconnected_htlcs_in;
 	struct ext_key *bip32_base;
 	int sigchld_rfd;
+	struct io_conn *sigchld_conn;
 	int exit_code = 0;
 	char **orig_argv;
 	bool try_reexec;
@@ -1099,10 +1101,8 @@ int main(int argc, char *argv[])
 	 * "funding transaction spent" event which creates it. */
 	onchaind_replay_channels(ld);
 
-	/*~ Now handle sigchld, so we can clean up appropriately.
-	 * We don't keep a pointer to this, so our simple leak detection
-	 * code gets upset unless we mark it notleak(). */
-	notleak(io_new_conn(ld, sigchld_rfd, sigchld_rfd_in, ld));
+	/*~ Now handle sigchld, so we can clean up appropriately. */
+	sigchld_conn = notleak(io_new_conn(ld, sigchld_rfd, sigchld_rfd_in, ld));
 
 	/*~ Mark ourselves live.
 	 *
@@ -1189,7 +1189,7 @@ int main(int argc, char *argv[])
 	stop_topology(ld->topology);
 
 	/* We're not going to collect our children. */
-	remove_sigchild_handler();
+	remove_sigchild_handler(sigchld_conn);
 	shutdown_subdaemons(ld);
 
 	/* Tell plugins we're shutting down, closes the db for write access. */
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -1193,12 +1193,8 @@ int main(int argc, char *argv[])
 	/* Tell plugins we're shutting down, closes the db for write access. */
 	shutdown_plugins(ld);
 
-	/* Clean up the JSON-RPC. This needs to happen in a DB transaction since
-	 * it might actually be touching the DB in some destructors, e.g.,
-	 * unreserving UTXOs (see #1737) */
-	db_begin_transaction(ld->wallet->db);
+	/* Cleanup JSON RPC separately: destructors assume some list_head * in ld */
 	tal_free(ld->jsonrpc);
-	db_commit_transaction(ld->wallet->db);
 
 	/* Clean our our HTLC maps, since they use malloc. */
 	htlc_in_map_clear(&ld->htlcs_in);
--- a/lightningd/lightningd.c
+++ b/lightningd/lightningd.c
@@ -242,6 +242,7 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
 	 */
 	ld->plugins = plugins_new(ld, ld->log_book, ld);
 	ld->plugins->startup = true;
+	ld->plugins->shutdown = false;
 
 	/*~ This is set when a JSON RPC command comes in to shut us down. */
 	ld->stop_conn = NULL;
@@ -1187,10 +1188,10 @@ int main(int argc, char *argv[])
 
 	/* We're not going to collect our children. */
 	remove_sigchild_handler();
-
-	/* Tell plugins we're shutting down. */
-	shutdown_plugins(ld);
 	shutdown_subdaemons(ld);
+
+	/* Tell plugins we're shutting down, closes the db for write access. */
+	shutdown_plugins(ld);
 
 	/* Clean up the JSON-RPC. This needs to happen in a DB transaction since
 	 * it might actually be touching the DB in some destructors, e.g.,
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -3541,6 +3541,9 @@ def test_upgrade_statickey_onchaind(node_factory, executor, bitcoind):
     l1.daemon.wait_for_log('option_static_remotekey enabled at 1/1')
 
     # Pre-statickey penalty works.
+    # FIXME: Without this sleep, l1 will broadcasts one tx more compared to good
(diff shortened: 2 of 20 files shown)


SimonVrouwe:
Rebased.

A little back and forth on what to do with RPC command in shutdown: failing them results in some ** BROKEN ** logs by plugins that subscribed to `shutdown` (all build-in plugins in DEVELOPER=1) which happen to receive the result of a (previous) RPC command when waiting in the 30s shutdown loop. This could be solved by lightningd not returning any result (`command_still_pending` forever) when in shutdown. But that would make RPC calls made (maybe just by chance) in the 30s shutdown loop hang forever which is not nice in my opinion.

So I choose to return error code `-32601` with message `lightningd is shutting down` and will see if its possible to **not** make our tests fail on that particular error.

@cdecker BTW: I see #4883 was closed, but that test (same as 9aeeae5 in this PR) about consistent hook-semantics-in-shutdown still fails on [v0.10.2rc2](https://github.com/ElementsProject/lightning/commit/000508ca1250aac63306ff60a253536501edfc18)

cdecker:
> @cdecker BTW: I see #4883 was closed, but that test (same as 9aeeae5 in this PR) about consistent hook-semantics-in-shutdown still fails on v0.10.2rc2

Sorry about that, i must've tagged the wrong issue in the PR.

SimonVrouwe:
Considered all these troubles and the fact that `shutdown` notification is currently only used for build-in plugin's to do a memleak_check:
https://github.com/ElementsProject/lightning/blob/12689674c7546a6b6ee70839e3867e04dcd33a2f/plugins/libplugin.c#L1265-L1267

No intention to hold up the release but if there is no time, maybe the `shutdown` notification can be made a **developer-only** option (for now)? Before plugin developers get the wrong impression what they can or should do with it.

ZmnSCPxj:
> No intention to hold up the release but if there is no time, maybe the `shutdown` notification can be made a **developer-only** option (for now)? Before plugin developers get the wrong impression what they can or should do with it.

I object!  It should be a **developer-only** option **forever**.  I think plugins should really just handle EOF on their input, as that handles the case where `lightningd` crashes unexpectedly.

lightningd/lightningd.c
```diff
@@ -757,13 +757,15 @@ static int setup_sig_handlers(void)
...
-static void remove_sigchild_handler(void)
+static void remove_sigchild_handler(int sigchld_rfd)
 {
 	struct sigaction sigchild;
 
 	memset(&sigchild, 0, sizeof(struct sigaction));
 	sigchild.sa_handler = SIG_DFL;
 	sigaction(SIGCHLD, &sigchild, NULL);
+	close(sigchld_wfd);
+	close(sigchld_rfd);
```
rustyrussell, line 768: Better to io_close() the io_conn associated with sigchld_rfd rather than just mug the fd.

i.e. save the pointer here:

```
notleak(io_new_conn(ld, sigchld_rfd, sigchld_rfd_in, ld));
```

And pass *that* into remove_sigchild_handler() to close.

lightningd/lightningd.c
```diff
@@ -783,9 +783,13 @@ static struct io_plan *sigchld_rfd_in(struct io_conn *conn,
...
 	static size_t len;
+	pid_t pid;
 
-	/* Reap the plugins, since we otherwise ignore them. */
-	while (waitpid(-1, NULL, WNOHANG) != 0);
+	/* Reap any terminated child we can; for the rest we ignore these */
+	while ((pid = waitpid(-1, 0, WNOHANG)) != 0)
+		/* No child processes: this causes a hang sometimes, so break that */
+		if (pid == -1 && errno == ECHILD)
+			break;
```
rustyrussell, line 793: I don't think this is what you're seeing; I suspect the OOM killer is hitting us.

rustyrussell (changes requested):
Probably drop final commit?

ZmnSCPxj:
How about deprecating the `shutdown` notification and instead create a new `clean-shutdown` chainable hook?

Then the `shutdown` command first triggers the hook and waits for all registered plugins to handle the hook before it *actually* sets the internal flag.

Using a hook provides cleaner semantics: we know exactly when plugins have finished whatever clean-shutdown-related work they have, because then they would respond with `{"result": "continue"}`.  No more phasing and no more heuristic-based waiting --- what if the node scales up and clean shutdown requires more than 30 seconds for a plugin to handle?

It also strongly implies that it only works for a *clean* shutdown, and not for all shutdown conditions.  Plugins are still strongly encouraged to not break if `lightningd` dies from under them.

SimonVrouwe:
Still one failing test `test_hsmtool_secret_decryption`.
I managed to reproduce the hang, but it only happens occasionally, but by setting `TIMEOUT` in pyln.testing.utils to large value and then:
`while pytest tests/ -s -k 'test_hsmtool_secret_decryption' --maxfail=1; do echo "ok"; done`
after a while one of the runs seems to hang, only printing lines `gossipd: seeker: no peers, waiting`
(NOTE: I was also running a full `pytest /tests` in another terminal, not sure if relevant.)

But it seems to hang on
`tools/hsmtool decrypt /tmp/ltests-x5nqrv48/test_hsmtool_secret_decryption_1/lightning-1/regtest/hsm_secret`

connect gdb to that, show backtrace:
```
0x00007f109fc1a461 in __GI___libc_read (fd=0, buf=0x559ab854d1e0, nbytes=1024) at ../sysdeps/unix/sysv/linux/read.c:26
26      ../sysdeps/unix/sysv/linux/read.c: No such file or directory.
(gdb) bt
#0  0x00007f109fc1a461 in __GI___libc_read (fd=0, buf=0x559ab854d1e0, nbytes=1024) at ../sysdeps/unix/sysv/linux/read.c:26
#1  0x00007f109fbac670 in _IO_new_file_underflow (fp=0x7f109fceba00 <_IO_2_1_stdin_>) at libioP.h:839
#2  0x00007f109fba0e38 in _IO_getdelim (lineptr=0x7fff42ec6d10, n=0x7fff42ec6d18, delimiter=10, fp=0x7f109fceba00 <_IO_2_1_stdin_>) at iogetdelim.c:73
#3  0x0000559ab66f317b in getline_stdin_pass (passwd=0x7fff42ec6d10, passwd_size=0x7fff42ec6d18) at common/hsm_encryption.c:83
#4  0x0000559ab66f3316 in read_stdin_pass (reason=0x7fff42ec6de0) at common/hsm_encryption.c:111
#5  0x0000559ab66c0340 in decrypt_hsm (hsm_secret_path=0x7fff42ec8192 "/tmp/ltests-x5nqrv48/test_hsmtool_secret_decryption_1/lightning-1/regtest/hsm_secret") at tools/hsmtool.c:172
#6  0x0000559ab66c191c in main (argc=3, argv=0x7fff42ec7008) at tools/hsmtool.c:599
```
is somewhere here in our test
https://github.com/ElementsProject/lightning/blob/efeb1bc65b0a1429023b2d26ef338db226efa499/tests/test_wallet.py#L1083-L1090

Well at least we know where to look now.
Maybe the test framework can be tweaked, so that in case the TimeoutError is raised, it looks around for such hangs and show back traces in log report?

lightningd/lightningd.c
SimonVrouwe: yes, dropped it

SimonVrouwe:
Rebased, modified commit messages but no code changes except 2 suggestion made by rustyrussel
and added a commit that fixes the `hsmtool` hang by using `TSCANOW` flag (for the interested, have a look at this [demo](https://github.com/SimonVrouwe/lightning_testing/commit/d70943ddde88fb64503d09f1cd69d0091fb58d3a) )

Added documentation.

SimonVrouwe:
I think this can be merged.

> How about deprecating the shutdown notification and instead create a new clean-shutdown chainable hook?

@ZmnSCPxj  Not sure if it needs to be depricated, this PR makes it safe I think but it needs documentation.

I agree with a `clean-shutdown` hook, but how many use cases are there? Some of such functionality can be also accomplished with a _method_ `shutdown_clean` such as in this [example](https://github.com/SimonVrouwe/lightning-tools/tree/2021-10-31_demo_custom_shutdown/clightning-plugins/gracefulshutdown).

SimonVrouwe:
Bugs keep coming.

Rebased it again, mostly a change to fix the (previous) [failure](https://github.com/ElementsProject/lightning/runs/4322633599?check_suite_focus=true), by adding checks around `io_break` calls.
https://github.com/ElementsProject/lightning/blob/ff0d59112db0ecc647c2b3d2ebdc2a09b942aa6b/lightningd/jsonrpc.c#L980-L982
https://github.com/ElementsProject/lightning/blob/ff0d59112db0ecc647c2b3d2ebdc2a09b942aa6b/lightningd/plugin.c#L194
https://github.com/ElementsProject/lightning/blob/ff0d59112db0ecc647c2b3d2ebdc2a09b942aa6b/lightningd/plugin.c#L2089
https://github.com/ElementsProject/lightning/blob/ff0d59112db0ecc647c2b3d2ebdc2a09b942aa6b/lightningd/plugin.c#L2119-L2120

Maybe something similar is useful for #4936?

tests/test_connection.py
```diff
@@ -3541,6 +3541,9 @@ def test_upgrade_statickey_onchaind(node_factory, executor, bitcoind):
     l1.daemon.wait_for_log('option_static_remotekey enabled at 1/1')
 
     # Pre-statickey penalty works.
+    # FIXME: Without this sleep, l1 will broadcasts one tx more compared to good
+    # case, causing `wait_for_onchaind_broadcast` to timeout.
+    time.sleep(5)
```
rustyrussell, line 3546: Thanks, I reproduced this issue, and have the proper fix I think... https://github.com/ElementsProject/lightning/pull/4941

rustyrussell (approved):
Ack ff0d59112db0ecc647c2b3d2ebdc2a09b942aa6b

rustyrussell:
OK, I have a fix for the leak valgrind discovered, in my other PR (https://github.com/ElementsProject/lightning/pull/4921) so I'll apply this after that one...

rustyrussell:
Actually, applying this now since it fixes the hsm encryption flake which is hitting our other PRs!

PR #4899, opened 2021-11-01
Database column operations by fieldname

rustyrussell:
Our field number accesses leave a lot to be desired, particularly when changing queries.  After trying several approaches, this was the cleanest, and I think most readable.

(I was doing some db manipulations in another patch, and changing large SELECT statements is very painful to get right!).

Changelog-None: internal cleanup


wallet/db.c
```diff
@@ -2165,6 +2165,323 @@ u8 *db_column_talarr(const tal_t *ctx, struct db_stmt *stmt, int col)
 			   db_column_bytes(stmt, col), 0);
 }
 
+/* Modern variants: by name */
+u64 db_col_u64(struct db_stmt *stmt, const char *colname)
+{
+	size_t col = db_query_colnum(stmt, colname);
+
+	if (db_column_is_null(stmt, col)) {
```
ZmnSCPxj, line 2230: Why is this more complex than `return db_column_u64(stmt, db_query_colnum(stmt, colname));`?  The existing `db_column_`* functions do the same checks.  Ditto for all the other `db_col_`* functions.

wallet/db.c
```diff
@@ -934,14 +955,26 @@ struct db_stmt *db_prepare_v2_(const char *location, struct db *db,
...
+	}
+#endif
+	return ret;
 }
 
 u64 db_column_u64(struct db_stmt *stmt, int col)
 {
 	if (db_column_is_null(stmt, col)) {
 		log_broken(stmt->db->log, "Accessing a null column %d in query %s", col, stmt->query->query);
+		abort();
```
ZmnSCPxj, line 987: Why an abort here?

rustyrussell: Oops, left on from debugging!!!

Good catch

wallet/db.c
rustyrussell: I cut and paste: I'll be removing the old functions

rustyrussell: Done!

wallet/db.c
rustyrussell: Fixed

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4899/commits/f83cf8784888e0abf5d83a2b0e5ddb1d3d9fcf1d

failure GI is unrelated, and it looks like a `lnprotocoltest` failure, I will look into t[his failure](https://github.com/ElementsProject/lightning/runs/4160658040?check_suite_focus=true) and looks if we can avoid it

rustyrussell:
Trivial rebase onto fresh master.

ZmnSCPxj:
Kicked CI in case it is just test flakiness

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/5539d69e5baab4a191522e4cca66b2608cf7b138

after rebase

PR #4900, opened 2021-11-01
Set default port according to network

jsarenik:
The idea is to have different default ports for different networks.
    
Current default port is `9735` for everything. Let's use it for
the mainnet and reuse the difference added to the default port
from `rpc_port` values in `bitcoin/chainstate.c`.
    
Testnet would be `19735` (adding rpc_port - 8332 = `10000`).
    
Signet would be `39735` (adding rpc_port - 8332 = `30000`).
    
Regtest would be `19846` (adding rpc_port - 8332 = `10111`).
    
With Vincenzo's pair-programming help.
    
Changelog-Changed: If the port is unspecified, the default port is chosen according to used network similarly to Bitcoin Core.


jsarenik:
Overall idea is that command-line specified port always takes precedence, but if it is not set then the port `lightningd` will listen to depends on the network.

From Bitcoin Core `-help` we see this:

```
  -port=<port>
       Listen for connections on <port>. Nodes not using the default ports
       (default: 8333, testnet: 18333, signet: 38333, regtest: 18444)
       are unlikely to get incoming connections. Not relevant for I2P
       (see doc/i2p.md).
```

And that means the differences to the default mainnet port are what the new variable `port_add` contains.

jsarenik:
@rustyrussell I need help. Shell-scripting mindset seems to be of little use here :)

Does the `port_add` make sense to you? What other way would you recommend me to go?

jsarenik:
Vincenzzo says this change may need some change in the specs.

vincenzopalazzo:
>UPDATE: Got help from @vincenzopalazzo

Just suggested how to fix the segmentation fault :) nothing else

>Vincenzzo says this change may need some change in the specs.

I like the idea to have some standards also on different ports, like testnet, request but (maybe not for liquid or different network like litecoin?)

bitcoin/chainparams.c
```diff
@@ -34,6 +34,7 @@ const struct chainparams networks[] = {
 				      0x5a, 0x08, 0x9c, 0x68, 0xd6, 0x19, 0x00,
```
vincenzopalazzo, line 34: I really like to have this type of offset for the network port, and I think that using this technique is pretty good. However, I things in this case we are mixing the bitcoin config with only one propriety related to c-lightning.

However, we can derive your offset from the following formula `lightning_port = defautl_lightning_port + (bitcoin_network_port - bitcoin_mainet_port)`. As an example we can have the following example:

`lightning_port = 9735 + (18332 - 8332)` where we will have that the lightning port is 19735.

What do you think?

lightningd/lightningd.c
```diff
@@ -965,6 +965,7 @@ int main(int argc, char *argv[])
 
 	/*~ Handle options and config. */
 	handle_opts(ld, argc, argv);
+	ld->portnum = DEFAULT_PORT + chainparams->port_add;
```
vincenzopalazzo, line 968:
```suggestion
      if (ld->portnum == DEFAULT_PORT )
	    ld->portnum = DEFAULT_PORT + chainparams->port_add;
```

vincenzopalazzo (changes requested):
LGTM.

I like your proposal, I have only a question on how to derive the lightning port from the bitcoin port standard, and also add a check to avoid the overriding of a custom port.

bitcoin/chainparams.c
jsarenik: So do you recommend to do `port_add = 18332 - 8332` instead of `port_add = 10000`?

lightningd/lightningd.c
jsarenik: The condition looks nice, but in practice it is not needed. I have verified locally that command-line takes precedence still. No condition needed.

jsarenik: Or is [the failed test](https://github.com/ElementsProject/lightning/runs/4092052054?check_suite_focus=true) caused by the missing condition?

jsarenik: To me the failed test looks like any other randomly failed test (since 2018) without reason. I am happy the `lightningd` is not failing like that in the wild and is actually very stable and reliable. But by-standers can be mislead when they see such failing tests.

jsarenik: It only needs more feed-back now :) so the condition is not important.

jsarenik:
@cdecker please have a look and comment.

bitcoin/chainparams.c
jsarenik: Check now. bbc94acef

jsarenik: What's the corner case? I tested it on mainnet and signet. And also with a port `12345` specified on the command line. Works well.

jsarenik: And even better, see b525259b0.

jsarenik: And now with a comment: 17a66f421

Enough. Will wait for feed-back now.

jsarenik:
This may be a breaking change for anyone who allowed port `9735` on their firwall, but used it for any other network than mainnet.

bitcoin/chainparams.c
jsarenik: See where one test was fixed along with the `rpc_port` for `regtest` network.

jsarenik:
This is going to be bigger than few-lines-patch :)

```
$ git grep -w 9735
CHANGELOG.md: - options: Allow the Tor inbound service port differ from 9735 ([3155](https://github.com/ElementsProject/lightning/pull/3155))
Dockerfile:ENV LIGHTNINGD_PORT=9735
Dockerfile:EXPOSE 9735 9835
README.md:(and optional `<port>`, if not 9735) and has the node ID `<node_id>`:
common/test/run-ip_port_parsing.c:	assert(parse_wireaddr("[::1]:9735", &addr, 500, false, NULL));
common/test/run-ip_port_parsing.c:	assert(addr.port == 9735);
common/test/run-ip_port_parsing.c:	assert(streq(ip, "[::1]:9735"));
common/test/run-wireaddr.c:	assert(parse_wireaddr("127.0.0.1:9735", &expect->u.wireaddr, 0, NULL, &err));
common/wireaddr.h: * The default TCP port is 9735. This corresponds to hexadecimal
common/wireaddr.h:#define DEFAULT_PORT 9735
contrib/linuxarm32v7.Dockerfile:ENV LIGHTNINGD_PORT=9735
contrib/linuxarm32v7.Dockerfile:EXPOSE 9735 9835
contrib/linuxarm64v8.Dockerfile:ENV LIGHTNINGD_PORT=9735
contrib/linuxarm64v8.Dockerfile:EXPOSE 9735 9835
contrib/pyln-proto/examples/listen.py:listener.bind(('0.0.0.0', 9735))
contrib/pyln-proto/pyln/proto/wire.py:def connect(local_privkey, node_id, host: str, port: int = 9735,
contrib/pyln-spec/bolt1/pyln/spec/bolt1/text.py:The default TCP port is 9735. This corresponds to hexadecimal `0x2607`: the Unicode code point for LIGHTNING.<sup>[1](#reference-1)</sup>
contrib/pyln-testing/pyln/testing/utils.py:    def __init__(self, lightning_dir, bitcoindproxy, port=9735, random_hsm=False, node_id=0):
doc/PLUGINS.md:    "addr": "34.239.230.56:9735",
doc/TOR.md:bind-addr=127.0.0.1:9735
doc/TOR.md:2.  `bind-addr` informs C-Lightning to bind itself to port 9735.
doc/TOR.md:    9735 is the normal Lightning Network port, so this setting may already be present.
doc/TOR.md:bind-addr=127.0.0.1:9735
doc/TOR.md:9735 port.
doc/TOR.md:Generally `--bind-addr=127.0.0.1:9735` should work fine.
doc/TOR.md:persistent .onion address will be 9735, but you can change this by
doc/TOR.md:HiddenServicePort 1234 127.0.0.1:9735
doc/TOR.md:line is `HiddenServicePort 1234 127.0.0.1:9735` the .onion address will be reachable at
doc/TOR.md:.onion address through the 9735 port.
doc/TOR.md:be able to connect to it through the 9735 port.
doc/lightning-connect.7.md:If not specified, the *port* defaults to 9735.
doc/lightning-listconfigs.7.md:   "bind-addr": "127.0.0.1:9735",
doc/lightning-listconfigs.7.md:   "announce-addr": "fp463inc4w3lamhhduytrwdwq6q6uzugtaeapylqfc43agrdnnqsheyd.onion:9735",
doc/lightning-listnodes.7.md:               "port": 9735
doc/lightningd-config.5.md:port 9735 on IPv4 and IPv6, and will announce it to peers if it seems
doc/lightningd-config.5.md:If 'PORT' is not specified, the default port 9735 is used for mainnet
doc/lightningd-config.5.md:and this will be used to configure a Tor hidden service for port 9735
doc/lightningd-config.5.md:not specified, 9735 is used.
doc/lightningd-config.5.md:is distinguished by ending in *.onion*. *PORT* defaults to 9735.
hsmd/libhsmd.c:	/* Clearly, we should use 9735, the unicode point for lightning! */
hsmd/libhsmd.c:				  BIP32_INITIAL_HARDENED_CHILD|9735,
```

jsarenik:
Tests ran manually on gitpod, commit aaf1dbfa9c0f9ebdf6042f1d30e6d7bbf8b364b2, are passing:

```
$ .github/scripts/setup.sh
...
$ .github/scripts/build.sh
...
==================== 429 passed, 45 skipped, 1 warning in 1104.16s (0:18:24) ====================
```

michaelfolkson:
Concept ACK

Having distinct default ports for testnet, signet etc makes sense to me

> From Bitcoin Core -help we see this:
>
>```
> -port=<port>
>       Listen for connections on <port>. Nodes not using the default ports
>       (default: 8333, testnet: 18333, signet: 38333, regtest: 18444)
>       are unlikely to get incoming connections. Not relevant for I2P
>       (see doc/i2p.md).
>```

This appears to be changing with https://github.com/bitcoin/bitcoin/pull/23306. Bitcoin Core seems to be moving in a direction where non-default ports are fully supported but that doesn't impact this PR's intention.

bitcoin/chainparams.c
```diff
@@ -63,7 +63,7 @@ const struct chainparams networks[] = {
 				      0x5b, 0xbf, 0x28, 0xc3, 0x4f, 0x3a, 0x5e,
 				      0x33, 0x2a, 0x1f, 0xc7, 0xb2, 0xb7, 0x3c,
 				      0xf1, 0x88, 0x91, 0x0f}}}},
-     .rpc_port = 18332,
+     .rpc_port = 18443,
```
rustyrussell, line 66: Actually, we don't use this field (which is why this bug does nothing)  Please delete it altogether!

rustyrussell (changes requested):
I think you'll need to `make doc` to regenerate the man page, too?

bitcoin/chainparams.c
jsarenik: Thanks for comment, I see your point, just that I use `rpc_port` now in that network-dependent-port code which I added into `lightningd.c`. And just wanted to keep the mask exactly as it is in Bitcoin Core (10111 is the difference between the mainnet and regtest).

jsarenik: 10111 decimal, though it may look binary :)

jsarenik:
Pushed a rebase to current master.

jsarenik:
@rustyrussell ah, as for `make doc` we have recently moved to no-generated-code in repo state so that is not needed and updating the `.md` file is enough.

jsarenik:
This morning, half-dreamig still, a `git bisect` was running in my head and the result is that three commits of this PR were squashed into one in order to make sure no one ever lands in the half-baked state.

bitcoin/chainparams.c
jsarenik: https://github.com/jsarenik/lightning/blob/jsn/default_ports/lightningd/lightningd.c#L954

doc/lightningd-config.5.md
```diff
@@ -357,27 +358,31 @@ Set an IP address (v4 or v6) or automatic Tor address to listen on and
...
-first IPv4 or IPv6 address we bind to and is by default unique to
-your nodes id. You can add the text '/torblob=BLOB' followed by up to
+and this will be used to configure a static Tor hidden service.
+You can add the text '/torblob=BLOB' followed by up to
 64 Bytes of text to generate from this text a v3 onion service
 address text unique to the first 32 Byte of this text.
 You can also use an postfix '/torport=TORPORT' to select the external
 tor binding. The result is that over tor your node is accessible by a port
-defined by you and possible different from your local node port assignment
+defined by you and possible different from your local node port assignment.
```
michaelfolkson, line 385: nit: s/possible/possibly

tests/test_gossip.py
```diff
@@ -130,8 +130,8 @@ def test_announce_address(node_factory, bitcoind):
     l1.daemon.wait_for_log(r"\[OUT\] 0101.*47"
                            "010102030404d2"
                            "017f000001...."
-                           "02000000000000000000000000000000002607"
-                           "04e00533f3e8f2aedaa8969b3d0fa03a96e857bbb28064dca5e147e934244b9ba50230032607")
+                           "02000000000000000000000000000000004d86"
+                           "04e00533f3e8f2aedaa8969b3d0fa03a96e857bbb28064dca5e147e934244b9ba50230034d86")
```
michaelfolkson, line 134: Can you explain what these are and why they need to be changed?

doc/lightningd-config.5.md
```diff
@@ -342,8 +342,9 @@ network.
 ### Networking options
 
 Note that for simple setups, the implicit *autolisten* option does the
-right thing: it will try to bind to port 9735 on IPv4 and IPv6, and will
-announce it to peers if it seems like a public address.
+right thing: for the mainnet (bitcoin) network it will try to bind to
```
michaelfolkson, line 345: Initially thought the "bitcoin" was unnecessary but appears you do specify mainnet is Bitcoin on this project :)

michaelfolkson (commented):
ACK 8b3eb6bb8f135ea072b05ffa9856b175a51c301b

(LGTM but unfamiliar with the codebase)

doc/lightningd-config.5.md
jsarenik: Thanks! Fixed in 6336d3317

doc/lightningd-config.5.md
jsarenik: Thanks for noticing that! Yes, it is part of history here that mainnet == bitcoin, but maybe I am just too word-exact here (with alternatives). So what change would you suggest @michaelfolkson ? I will be happy to change it.

michaelfolkson: I guess it just depends whether the implementation either today or in future could support Bitcoin-like altcoins (or Liquid maybe?). If it could it makes sense to keep referring to Bitcoin. If it doesn't just mainnet would be enough as implicitly that means Bitcoin mainnet. As I said there are lots of references to "Bitcoin mainnet" in the codebase so this isn't out of place :)

jsarenik: Liquid is already supported with C-Lightning. As well as Litecoin and Litecoin-testnet. So it makes sense for that network reference to be there (and even more now because the default port mentioned is valid _only_ for Bitcoin mainnet).

Thanks for spending time thinking about it. In a way I find this issue similar to Bitcoin Core's https://github.com/bitcoin/bitcoin/pull/23306, just approached backwards, i.e. C-Lightning supported different ports early, but did not use different default port according to the selected network.

tests/test_gossip.py
jsarenik: The commit message says this (and I agree it is very terse)

```
       hex: 2607 is now 4d86
       dec: 9735 is now 19846
```

So in that test, the regtest network is being used and previously even regtest used the default port of `9735` when the port was not set, but now for regtest there is `9735 + 18443 - 8332` (the default port `plus` [rpc_port specific for network](https://github.com/jsarenik/lightning/blob/jsn/default_ports/bitcoin/chainparams.c#L66) `minus` default RPC port (8332, used as a mask here)).

jsarenik: @michaelfolkson notice those old lines ending with `2607` and new lines with `4d86`.

tests/test_gossip.py
```diff
@@ -938,7 +938,7 @@ def test_gossip_addresses(node_factory, bitcoind):
         {'type': 'ipv4', 'address': '127.0.0.1', 'port': 9735},
         {'type': 'ipv6', 'address': '::', 'port': 3},
         {'type': 'ipv6', 'address': '::', 'port': 9735},
```
vincenzopalazzo, line 940:
```suggestion
        {'type': 'ipv6', 'address': '::', 'port': 19846},
```
Not sure if it is the correct port number, but this causes the failure in the CI

tests/test_gossip.py
```diff
@@ -938,7 +938,7 @@ def test_gossip_addresses(node_factory, bitcoind):
         {'type': 'ipv4', 'address': '127.0.0.1', 'port': 9735},
```
vincenzopalazzo, line 938: Same here

tests/test_gossip.py
vincenzopalazzo, line 940: and here

vincenzopalazzo (changes requested):

tests/test_gossip.py
jsarenik: I have changed these now. But they were OK as they were set to 9375 manually and setting port to anything still works.

jsarenik:
@rustyrussell please revisit

lightningd/lightningd.c
```diff
@@ -197,7 +197,7 @@ static struct lightningd *new_lightningd(const tal_t *ctx)
 	 * NULL.  So we start with a zero-length array. */
 	ld->proposed_wireaddr = tal_arr(ld, struct wireaddr_internal, 0);
 	ld->proposed_listen_announce = tal_arr(ld, enum addr_listen_announce, 0);
-	ld->portnum = DEFAULT_PORT;
+	// ld->portnum moved after handle_early_opts
```
vincenzopalazzo, line 200: I think if it is only a comment, we can remove it. Usually the comment describe somethings, and a refactoring it is not considered "somethings" :)

vincenzopalazzo (approved):
If I'm not missing any additional test, LGTM and I like the derivation port from the default one for the other network 🛩️ 

ack https://github.com/ElementsProject/lightning/pull/4900/commits/632dda34b6bd1e4e19cbdde10a6072a1ea673f2d

lightningd/lightningd.c
jsarenik: OK, fair enough. Will remove it tomorrow. Thanks for feed-back.

jsarenik: Done.

jsarenik:
When testing locally I get following error:

```
**BROKEN** gossipd: gossip_store: Checksum verification failed (0101cf5d870bc7ecabcb7cd16898ef66891e5f0c6c5851bd85b670f03d325bc44d7544d367cd852e18ec03f7f4ff369b06860a3b12b07b29f36fb318ca11348bf8ec00005aab817c03f113414ebdc6c1fb0f33c99cd5a1d09dd79e7fdf2468cf1fe1af6674361695d23974b250757a7a6c6549544300000000000000000000000000000000000000000000000007010566933e4d86). Moving to gossip_store.corrupt and truncating
```

So it seems some checksum needs to be fixed when the port is changed to hex `4d86`. Please help me to figure it out.

jsarenik:
@cdecker the FreeBSD test ends with following:

```
/usr/bin/ssh -t freebsd
Pseudo-terminal will not be allocated because stdin is not a terminal.
ssh: connect to host localhost port 2222: Connection refused
Error: The process '/usr/bin/ssh' failed with exit code 255
```

cdecker:
Yep, that's a flake, given the VM has some timing issues. Kicking CI 👍

jsarenik:
@cdecker please kick it once again. Just one Normal Test time-outed now.

tests/test_gossip.py
```diff
@@ -1036,10 +1036,10 @@ def test_gossip_addresses(node_factory, bitcoind):
     nodes = l2.rpc.listnodes(l1.info['id'])['nodes']
     assert len(nodes) == 1 and nodes[0]['addresses'] == [
         {'type': 'ipv4', 'address': '127.0.0.1', 'port': 2},
-        {'type': 'ipv4', 'address': '127.0.0.1', 'port': 9735},
+        {'type': 'ipv4', 'address': '127.0.0.1', 'port': 19846},
         {'type': 'ipv6', 'address': '::', 'port': 3},
-        {'type': 'ipv6', 'address': '::', 'port': 9735},
-        {'type': 'torv3', 'address': 'vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion', 'port': 9735},
+        {'type': 'ipv6', 'address': '::', 'port': 19846},
+        {'type': 'torv3', 'address': 'vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion', 'port': 19846},
```
vincenzopalazzo, line 1042: Here we need to check the network, in fact, some tests are failing because are running on elements blockchain.

Maybe we need a property before the assert (line 1037) that looks like the following

`port = 19846 if ln.info["network"] is "regtest" else 20735`

I would love to have it at the beginning of the file a list of constant like `LIQUID_PORT = xXX` to make the code more readable

vincenzopalazzo (changes requested):
The other [CI failure](https://github.com/ElementsProject/lightning/runs/4405481792?check_suite_focus=true#step:5:3501) looks unrelated, but I'm not sure

tests/test_gossip.py
```diff
@@ -1036,10 +1033,7 @@ def test_gossip_addresses(node_factory, bitcoind):
     nodes = l2.rpc.listnodes(l1.info['id'])['nodes']
     assert len(nodes) == 1 and nodes[0]['addresses'] == [
         {'type': 'ipv4', 'address': '127.0.0.1', 'port': 2},
-        {'type': 'ipv4', 'address': '127.0.0.1', 'port': 9735},
         {'type': 'ipv6', 'address': '::', 'port': 3},
-        {'type': 'ipv6', 'address': '::', 'port': 9735},
-        {'type': 'torv3', 'address': 'vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion', 'port': 9735},
```
vincenzopalazzo, line 1042: Mh, I did not catch why you are removing the test here.

vincenzopalazzo (commented):
LGTM overall but I have a small missing in the `test_gossip.py`

tests/test_gossip.py
jsarenik: Because this was caused by lines with address and no port a couple of lines earlier:

```diff
@@ -1018,10 +1018,7 @@ def test_gossip_addresses(node_factory, bitcoind):
     l1 = node_factory.get_node(options={
         'announce-addr': [
             '[::]:3',
-            '[::]',
             '127.0.0.1:2',
-            '127.0.0.1',
-            'vww6ybal4bd7szmgncyruucpgfkqahzddi37ktceo3ah7ngmcopnpyyd.onion',
             '4acth47i6kxnvkewtm6q7ib2s3ufpo5sqbsnzjpbi7utijcltosqemad.onion:123
4'
         ],
     })
```

And since each different network would get a different default port when the port is not specified, I have removed all these default-port lines. I think it makes sense to not replicate the same logic which already is in `lightningd.c` into the Python test since it would just be repetition IMHO.

tests/test_gossip.py
jsarenik: Instead of defining the ports for different networks (which would need to be manually changed every time something happens to that code. Yes, I have seen QA departments doing exactly this (rewriting the same code from C to Python for the tests), I think we are understaffed to hurt ourselves this way with no actual reason. The test is still testing all the real things, it just does not care about what default port is chosen when not specified.

tests/test_gossip.py
vincenzopalazzo: Mh this can introduce some regression in the future. I think introducing a duplication logic in the test is fine because we are testing if the code is really working as expected.

I like the idea to have this part of the code covered by a integration test, and maybe this can be a solution https://github.com/ElementsProject/lightning/pull/4900#discussion_r761823282

jsarenik:
The last CI run failed on this: https://github.com/ElementsProject/lightning/runs/4421365349?check_suite_focus=true. Re-running.

jsarenik:
This is not funny. CI failed again on a "Flaky Test", whatever that means. Re-running.

jsarenik:
It seems to me that "Flaky Tests" should be taken out of the regular CI runs and run separately on regular basis with opt-in reporting (for whoever likes to receive Flaky Test failures). This way it would not block or confuse anyone and still whoever wants the results will get them and may use them for what they are worth.

rustyrussell:
> It seems to me that "Flaky Tests" should be taken out of the regular CI runs and run separately on regular basis with opt-in reporting (for whoever likes to receive Flaky Test failures). This way it would not block or confuse anyone and still whoever wants the results will get them and may use them for what they are worth.

No, flaky tests should be *fixed*.  But it's often an ongoing battle, to detect whether it's a test flake, or a real intermittent issue.  And it also doesn't happen on my test boxes :(

rustyrussell:
I will apply this, as it's been sitting here too long, and restore the default-port-is-correct test as a separate commit.

Ack: https://github.com/ElementsProject/lightning/commit/51ce5bb4561aef1a51b5dfa44db69db6095218d8

tests/test_gossip.py
jsarenik: https://github.com/ElementsProject/lightning/pull/4958

michaelfolkson:
> Vincenzo says this change may need some change in the specs.

Just a reminder that this needs a change to BOLT 1 (maybe worth opening a spec PR and/or discussing in the spec meeting @vincenzopalazzo @jsarenik?)

vincenzopalazzo:
>Just a reminder that this needs a change to BOLT 1 (maybe worth opening a spec PR and/or discussing in the spec meeting @vincenzopalazzo @jsarenik?)

This is something that I think for a while, but I don't have time to know how bitcoin core remove the default port https://github.com/bitcoin/bitcoin/pull/23306 I will check this week,

Thanks for pinging on this point!

vincenzopalazzo:
Opened a draft on the Spec https://github.com/lightning/bolts/pull/968 I will review the text in the next few days to avoid different typos e improve the English form.

PR #4902, opened 2021-11-04
Remove deprecated things for next release

rustyrussell:
Also reworked TOR.md for more clarity (since we remove v2).


vincenzopalazzo (approved):
LGTM https://github.com/ElementsProject/lightning/pull/4902/commits/d48ef07125075c99ce6e170aefa791a522edfc78

Seems only that there is some python test related to the onion `addr`

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4902/commits/99cf8bab4b9aad6ce22bcadcbd14e28f51cad51e

lightningd/options.c
```diff
@@ -226,24 +237,6 @@ static char *opt_add_announce_addr(const char *arg, struct lightningd *ld)
...
-		assert(ld->proposed_wireaddr[i].itype == ADDR_INTERNAL_WIREADDR);
-		wi = &ld->proposed_wireaddr[i].u.wireaddr;
-
-		if (wn->type != wi->type)
-			continue;
-		return tal_fmt(NULL, "Cannot announce address %s;"
-			       " already have %s which is the same type",
-			       type_to_string(tmpctx, struct wireaddr, wn),
-			       type_to_string(tmpctx, struct wireaddr, wi));
-	}
```
vincenzopalazzo, line 246: I'm missing why we are removing this one?

vincenzopalazzo (commented):
I have only a small comment

lightningd/options.c
m-schmoock: Because it is just preventing the user from doing something he must have done consciously. If the user gives multiple addresses he most likely has a reason for it and since it is allowed from the protocol spec, why prevent that?

rustyrussell: Yes, it used to be forbidden, but spec was updated to allow it (and other implementations allowed it).

vincenzopalazzo: Ah, got it :) thanks and sorry for the missing

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4902/commits/8d0d4dccfc06c6948c67ff3a97f61f54fd00945f

PR #4907, opened 2021-11-05
Update install docs and add some files to gitignore

andrewtoth:
In my quest to get `make -j12 full-check PYTEST_PAR=24 DEVELOPER=1 VALGRIND=0` to work on Ubuntu from [HACKING.md](https://github.com/ElementsProject/lightning/blob/master/doc/HACKING.md#testing), I came across some directories not ignored and had to install some dependencies. Updated the deps install docs and added some dirs to gitignore.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -37,9 +37,10 @@ Get dependencies:
 
     sudo apt-get update
     sudo apt-get install -y \
-      autoconf automake build-essential git libtool libgmp-dev \
-      libsqlite3-dev python3 python3-mako net-tools zlib1g-dev libsodium-dev \
+      autoconf automake build-essential git libtool libgmp-dev libsqlite3-dev \
+      python3 python3-mako python3-pip net-tools zlib1g-dev libsodium-dev \
       gettext
+    pip3 install --user mrkd
 
 If you don't have Bitcoin installed locally you'll need to install that
 as well. It's now available via [snapd](https://snapcraft.io/bitcoin-core).
@@ -57,8 +58,10 @@ Clone lightning:
     
 For development or running tests, get additional dependencies:
 
-    sudo apt-get install -y valgrind python3-pip libpq-dev
-    sudo pip3 install -r requirements.txt --use-feature=in-tree-build
+    sudo apt-get install -y valgrind libpq-dev shellcheck cppcheck \
+      libsecp256k1-dev
+    pip3 install --upgrade pip
+    pip3 install --user -r requirements.txt
 
 Build lightning:
 
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,8 @@ contrib/pylightning/pylightning.egg-info/
 contrib/pyln-*/build/
 contrib/pyln-*/dist/
 contrib/pyln-*/pyln_*.egg-info/
+contrib/pyln-*/.eggs/
+contrib/pyln-*/pyln/*/__version__.py
 release/
 tests/plugins/test_selfdisable_after_getmanifest
 .DS_Store
--- a/doc/HACKING.md
+++ b/doc/HACKING.md
@@ -163,12 +163,10 @@ Build and Development
 Install `valgrind` and the python dependencies for best results:
 
 ```
-sudo apt install valgrind cppcheck shellcheck libsecp256k1-dev
-pip3 install --user \
-         -r requirements.txt \
-         -r contrib/pyln-client/requirements.txt \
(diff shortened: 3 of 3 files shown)


contrib/pyln-proto/requirements.txt
```diff
@@ -1,6 +1,6 @@
 base58 ~= 2.0.1
 bitstring ~= 3.1.6
-coincurve ~= 13.0
+coincurve ~= 15.0.1
```
vincenzopalazzo, line 3: What is the error that you receive here? can you add a little bit more content?

```suggestion
coincurve >= 13.0
```

However, we preferer to use the `>=` instead of the `~=`, because  `~=` will postpone in the future a future error.

vincenzopalazzo (changes requested):
>- add max-dust-htlc-exposure-msat to lightningd-config.5.md, fetchinvoice-noconnect to listconfigs.schema.json and lightningd-listconfigs.7.md.
>- add =*PORT* to experimental-websocket-port in lightningd-config.5.md. 
>- don't require documenting options with funder/lease/channel-fee in check-manpage.sh.

This is a duplicate fix, a PR with fix proposed in https://github.com/ElementsProject/lightning/pull/4782 with a discussion on how formatting json stuff also

contrib/pyln-proto/requirements.txt
andrewtoth: It was because I didn't have `libsecp256k1-dev`. Added to docs.

andrewtoth:
@vincenzopalazzo fixed, updated PR title and description. This is only updating gitignore and docs now.

doc/INSTALL.md
```diff
@@ -57,8 +58,8 @@ Clone lightning:
     
 For development or running tests, get additional dependencies:
 
-    sudo apt-get install -y valgrind python3-pip libpq-dev
-    sudo pip3 install -r requirements.txt --use-feature=in-tree-build
+    sudo apt-get install -y valgrind python3-pip libpq-dev shellcheck cppcheck libsecp256k1-dev
+    pip3 install --user -r requirements.txt
```
vincenzopalazzo, line 62: In version 21 of pip3 the feature `--use-feature` is still required, so I think we need to keep it until we are sure that there is no pip3 version 21 around
```suggestion
    pip3 install --user -r requirements.txt --use-feature=in-tree-build
```

vincenzopalazzo (approved):
LGTM, I have only a small comment.

doc/INSTALL.md
andrewtoth: Hmm I am using version 20.3.4 and it simply breaks with `no such option: --user-feature`. Are we sure which versions require this?

vincenzopalazzo: Ah ok your pip version is an old pip3 version https://github.com/pypa/pip/tags

Anyway, I think we need to document both in this case, and not just remove stuff because another user will put this change of a different env.

andrewtoth: I upgraded to 21.3.1 and it shows
```
DEPRECATION: In-tree builds are now the default. pip 22.1 will enforce this behaviour change. A possible replacement is to remove the --use-feature=in-tree-build flag.
```
So 20 breaks with `in-tree-build` and 21 already defaults to it. Which versions need this option to run?

vincenzopalazzo: I suggest to documenting Before 21 and 21 because version 22 is the newest and we can stay safe with the warning.

andrewtoth: Ok, so looking into this more. Installing the contrib deps actually does not work with pip 20.3.4, because the `in-tree-build` option doesn't exist. Building it without that option causes the error `LookupError: setuptools-scm was unable to detect version for '/'.`. This is because without in-tree-builds the pyln-* dirs are copied to temp dirs before building, which makes them unable to detect version. I don't know why it worked for me initially before. Installation instructions need to upgrade pip first before installing requirements, which then will use in-tree-build by default. I've updated the instructions accordingly.

jsarenik:
ACK ee188085147fe3b4139c9424a39e0f8b1db23f9b

Tested.

```
============ 429 passed, 45 skipped, 1 warning in 1160.62s (0:19:20) ============
```

vincenzopalazzo (approved):
untested on a clean machine, but I think that @jsarenik did it on a git pot clean machine. So,

ack https://github.com/ElementsProject/lightning/pull/4907/commits/6360d5589316807cb733f6d978333b36e83a41c8

PR #4911, opened 2021-11-08
paystatus: remove doubled amount_msat.

rustyrussell:
It's always been there:

```
{
   "pay": [
      {
         "bolt11": "...",
         "amount_msat": "1000msat",
         "amount_msat": "1000msat",
         "destination": "03...",
         "attempts": [
            {
               "strategy": "Initial attempt",
               "start_time": "2021-11-06T04:20:20.135Z",
               "age_in_seconds": 229032,
               "end_time": "2021-11-06T04:20:27.792Z",
               "state": "completed",
               "success": {
                  "id": 31994,
                  "payment_preimage": "..."
               }
            }
         ]
      }
   ]
}
```

Reported-by: denis2342 on IRC
Changelog-Fixed: JSON-RPC: `paystatus` entries no longer have two identical `amount_msat` entries.

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1598,9 +1598,6 @@ static struct command_result *json_paystatus(struct command *cmd,
 		if (p->invstring)
 			json_add_invstring(ret, p->invstring);
 		json_add_amount_msat_only(ret, "amount_msat", p->amount);
-		json_add_string(
-		    ret, "amount_msat",
-		    type_to_string(tmpctx, struct amount_msat, &p->amount));
 
 		json_add_node_id(ret, "destination", p->destination);
 


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4911/commits/105b0968cdf56a909c1defaff707c93fa24d7257

denis2342:
my proposal would be to replace both these amount_msat line with the following line:

json_add_amount_msat_compat(ret, p->amount, "msatoshi", "amount_msat");

This generates two fields, one as string and one as integer, both msat.

rustyrussell:
> my proposal would be to replace both these amount_msat lines with the following line:
> 
> json_add_amount_msat_compat(ret, p->amount, "msatoshi", "amount_msat");
> 
> This generates two fields, one as string and one as integer, both msat.

The "msatoshi" fields are being slowly deprecated, though.  The intention is that all amounts are explicitly "msat" (we had errors early on when amounts were specified differently in different places).

denis2342:
> > my proposal would be to replace both these amount_msat lines with the following line:
> > json_add_amount_msat_compat(ret, p->amount, "msatoshi", "amount_msat");
> > This generates two fields, one as string and one as integer, both msat.
> 
> The "msatoshi" fields are being slowly deprecated, though. The intention is that all amounts are explicitly "msat" (we had errors early on when amounts were specified differently in different places).

So no integer field is wanted?

ZmnSCPxj:
> So no integer field is wanted?

Correct, the intent is to always use a string with an explicit unit.

rustyrussell:
Trivial rebase on master.

Ack 9fbcc168a200fec6b165c66b8cc374bf01e508fb

PR #4919, opened 2021-11-18
wallet: fix sqlite3 column renaming since db backup merge.

rustyrussell:
Because db->conn is a void *, changing it (from a direct pointer to
a pointer to a pair of pointers) did not break compile if one place hadn't
been update.

The result was a confusing failure: sqlite3 complaining about API misuse,
since the db->conn pointer was not a valid db handle any more.

This is one case where avoiding a void * is hard: we might not even
have the postgresql types, since it might not be installed.  But a union
would have been superior here.

Changelog-None

### DIFF
--- a/wallet/db_sqlite3.c
+++ b/wallet/db_sqlite3.c
@@ -462,15 +462,17 @@ static char **prepare_table_manip(const tal_t *ctx,
 	char *cmd, *bracket;
 	char **parts;
 	int err;
+	struct db_sqlite3 *wrapper = (struct db_sqlite3 *)db->conn;
 
 	/* Get schema. */
-	sqlite3_prepare_v2(db->conn, "SELECT sql FROM sqlite_master WHERE type = ? AND name = ?;", -1, &stmt, NULL);
+	sqlite3_prepare_v2(wrapper->conn, "SELECT sql FROM sqlite_master WHERE type = ? AND name = ?;", -1, &stmt, NULL);
 	sqlite3_bind_text(stmt, 1, "table", strlen("table"), SQLITE_TRANSIENT);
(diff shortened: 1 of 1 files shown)


niftynei:
ACK a2dff65b31f8d01690fca16a0b15605ad3013d3c

wallet/db_sqlite3.c
```diff
@@ -489,22 +491,26 @@ static char **prepare_table_manip(const tal_t *ctx,
...
 		      tablename, tablename);
-	sqlite3_prepare_v2(db->conn, cmd, -1, &stmt, NULL);
+	sqlite3_prepare_v2(wrapper->conn, cmd, -1, &stmt, NULL);
 	if (sqlite3_step(stmt) != SQLITE_DONE)
 		goto sqlite_stmt_err;
 	sqlite3_finalize(stmt);
 
+	/* Make sure we do the same to backup! */
+	replicate_statement(wrapper, "PRAGMA foreign_keys = OFF;");
+	replicate_statement(wrapper, cmd);
```
ZmnSCPxj, line 508: ***WAIT A MINUTE!***  If you are calling `replicate_statement` directly here, that means you are *not* calling `db_sqlite3_changes_add`.  And that means the queries are not propagated to `db_write` hook, meaning it is not propagated to *all* backup methods.  Would this not cause problems if the user were using the `backup` plugin?  Some queries sent to the local copy are not sent to the `backup` plugin?

ZmnSCPxj: Fortunately it seems we do not actually use `delete_columns` or `rename_columns` *yet*, so we can still fix this, probably by rewriting in terms of `db_sqlite3_exec` (which gets the proper unwrapping of `db->conn` *and* replication to *all* backup processes we have, for free).

PR #4921, opened 2021-11-18
Onion Updates to Match Latest Spec (and test vectors!)

rustyrussell:
Once again, we support old and new onions.  First I remove the old-old spec versions.  Then I rename the new to obs2.  Then add the new ones.  Then make sure we send and receive both OK.

This should comply with the latest route blinding https://github.com/lightning/bolts/pull/765 and onion message https://github.com/lightning/bolts/pull/759 proposals.

Thanks to @t-bast and @thomash-acinq and @joostjager for all the feedback!


gossipd/test/run-onion_message.c
```diff
@@ -0,0 +1,580 @@
...
+	rs = process_onionpacket(tmpctx, op, &onion_ss, NULL, 0, false);
+	assert(rs);
+
+	printf("\"onion contents\": {\n");
+	json_strfield("raw", tal_hex(tmpctx, rs->raw_payload));
+
+	cursor = rs->raw_payload;
+	max = tal_bytelen(rs->raw_payload);
+	maxlen = fromwire_bigsize(&cursor, &max);
+	json_strfield("length", tal_fmt(tmpctx, "%"PRIu64, maxlen));
```
vincenzopalazzo, line 463: Not sure that it can be an optimal solution
```suggestion
	json_strfield("length", tal_fmt(tmpctx, "%lu", maxlen));
```

vincenzopalazzo (commented):
LGTM, Github action point me out an error, where I don't know if there are any better solutions than the ones that I proposed in the comment.

Anyway, it is only a comments

tests/test_pay.py
```diff
@@ -4367,7 +4367,7 @@ def test_fetchinvoice_3hop(node_factory, bitcoind):
     l4.rpc.connect(l3.info['id'], 'localhost', l3.port)
 
     l1.rpc.call('fetchinvoice', {'offer': offer1['bolt12']})
-
+
```
vincenzopalazzo, line 4370:
```suggestion

```

vincenzopalazzo (changes requested):
I think here the mistake it is the space and not the end line, if I'm not wrong python have 2 end-line conventions between function.

So, in this case, looks good the change, but I think in the file there is a space " "

rustyrussell:
Rebase to fix flakes...

gossipd/test/run-onion_message.c
rustyrussell: %zu is the correct format for size_t types, actually...

vincenzopalazzo:
I think we have a list part to migrate in the new spec version, that it is `test_pay.py` https://github.com/ElementsProject/lightning/blob/35c6f90666e64ff229217b136f4d05f0d545acdd/tests/test_pay.py#L3284-L3301

thomash-acinq:
This PR is compatible with https://github.com/ACINQ/eclair/commit/59b403559bdfc0ba884503bb12a14846b7aa5d2d, eclair and c-lightning can send and relay messages to each other.

rustyrussell:
> I think we have a list part to migrate in the new spec version, that it is `test_pay.py`
> 
> https://github.com/ElementsProject/lightning/blob/35c6f90666e64ff229217b136f4d05f0d545acdd/tests/test_pay.py#L3284-L3301

Yes.  While HTLCs with blinded routing are not completely specced, they're pretty obvious.  I'll see if it's better to implement or just disable for now...

rustyrussell:
Re-rebased on master.

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4921/commits/afb91fdbb7fc34c3cb2c61ce1e1060877b682e52

lightningd/onion_message.c
```diff
@@ -173,6 +173,7 @@ void handle_onionmsg_to_us(struct lightningd *ld, const u8 *msg)
 	 * means we created the path it's using. */
 	if (!self_id || !secret_eq_consttime(self_id, &ld->onion_reply_secret))
 		payload->our_alias = tal_free(payload->our_alias);
+	tal_free(self_id);
```
SimonVrouwe, line 176: Are you sure it is `self_id` that is leaking? The report points to
```
"label": "gossipd/gossipd_wiregen.c:528:struct secret",
```

SimonVrouwe: Yes its the same, I didn't look.

PR #4922, opened 2021-11-19
Update bitcoin-cli to 0.21.1

johnongit:
Bitcoin-cli deployed in the docker image is a bit old (0.18.1).
We can update it to 0.21.1

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,16 +17,16 @@ RUN wget -qO /opt/tini "https://github.com/krallin/tini/releases/download/v0.18.
     && echo "12d20136605531b09a2c2dac02ccee85e1b874eb322ef6baf7561cd93f93c855 /opt/tini" | sha256sum -c - \
     && chmod +x /opt/tini
 
-ARG BITCOIN_VERSION=0.18.1
+ARG BITCOIN_VERSION=22.0
 ENV BITCOIN_TARBALL bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
 ENV BITCOIN_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/$BITCOIN_TARBALL
-ENV BITCOIN_ASC_URL https://bitcoincore.org/bin/bitcoin-core-$BITCOIN_VERSION/SHA256SUMS.asc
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN wget -qO /opt/tini "https://github.com/krallin/tini/releases/download/v0.18.
     && echo "12d20136605531b09a2c2dac02ccee85e1b874eb322ef6baf7561cd93f93c855 /opt/tini" | sha256sum -c - \
     && chmod +x /opt/tini
 
-ARG BITCOIN_VERSION=0.18.1
+ARG BITCOIN_VERSION=0.21.1
```
vincenzopalazzo, line 20:
```suggestion
ARG BITCOIN_VERSION=0.22
```

Why not upgrade to the last version at this point?

vincenzopalazzo (changes requested):
Hey thanks for your contribution  :)

Great work, I have only a small question that it is not a change request, but I would like to have also your opinion.

the only thing that I would like to ask is if you can squash the comment in the only one, if you are new with git [here](https://stackoverflow.com/questions/35703556/what-does-it-mean-to-squash-commits-in-git) there is a description

Dockerfile
johnongit: Thx for your response.
I have not upgraded to 22.0 because the signature verification process is pretty different.
I can propose a new PR with the last version

vincenzopalazzo: I don't see any reason to make the dev life more complex, I'm happy with v0.21.1 :) but also I'm happy to review your PR for 22.0 support :)

johnongit:
This my new PR. I have squashed my commits and updated bitcoin-cli to 22.0.
As you can see, SHA256SUMS.asc has been renamed to SHA256SUMS from https://bitcoincore.org/bin/bitcoin-core-22.0/.
Thanks for your review.

vincenzopalazzo (approved):
LGTM

utack https://github.com/ElementsProject/lightning/pull/4922/commits/4cf314981c24011ce0301758357867c903555043

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4922/commits/4cf314981c24011ce0301758357867c903555043

Issue #4901, opened 2021-11-04
Adding partial index of unresolved HTLCs drastically reduces startup time


whitslack:
### Issue and Steps to Reproduce
Starting C-Lightning with a database containing a large number of heavily used channels can take a very long time, as the database queries in [`wallet_htlcs_load_in_for_channel`](https://github.com/ElementsProject/lightning/blob/v0.10.2/wallet/wallet.c#L2676-L2695) and [`wallet_htlcs_load_out_for_channel`](https://github.com/ElementsProject/lightning/blob/v0.10.2/wallet/wallet.c#L2724-L2744) can at best make use of the `channel_htlcs_channel_id_channel_htlc_id_direction_key` unique index to find the rows for each channel but must then read each such row to check its `hstate` value. This can require a very large number of random reads, and it only grows worse as the channels accumulate more historical HTLC records with age.

Introducing the following partial index drastically reduces startup time by requiring the HTLC restoring queries only to read the rows corresponding to *unresolved* HTLCs. The reduction in random reads can be by more than an order of magnitude. Maintaining the index is trivial, as it is always very small.

```
CREATE INDEX channel_htlcs_unresolved_idx
    ON channel_htlcs (channel_id)
    WHERE direction = 0 AND hstate <> 19 OR direction = 1 AND hstate <> 9;
```

### Empirical observations

The following observations record the elapsed time between the first "Restored __ incoming HTLCs" and the last "Restored __ outgoing HTLCs" during each startup occasion.

**Without new index**

* 2021-09-06T19:31:31.964Z − 2021-09-06T19:27:38.928Z = 233.036 s
* 2021-09-13T21:30:15.448Z − 2021-09-13T21:24:35.578Z = 339.870 s
* 2021-09-16T21:32:47.081Z − 2021-09-16T21:24:53.665Z = 473.416 s
* 2021-10-07T05:20:05.785Z − 2021-10-07T05:10:52.015Z = 553.770 s

**With new index**

* 2021-10-08T19:15:56.126Z − 2021-10-08T19:15:32.214Z = 23.912 s
* 2021-10-17T00:40:34.488Z − 2021-10-17T00:40:13.479Z = 21.009 s
* 2021-10-17T00:43:01.231Z − 2021-10-17T00:42:58.043Z = 3.188 s (hot caches)
* 2021-10-17T00:51:15.229Z − 2021-10-17T00:51:04.583Z = 10.646 s (warm caches)
[...]

PR #4925, opened 2021-11-23
wallet/db.c, wallet/wallet.c: Add a partial index to speed up startup.

ZmnSCPxj:
Closes: #4901

Tested by `EXPLAIN QUERY PLAN` on sqlite3; #4901 shows the result from
@whitslack doing a similar partial index on PostgreSQL on his ~1000 chan
node.

### DIFF
--- a/wallet/db.c
+++ b/wallet/db.c
@@ -2,8 +2,10 @@
 
 #include <bitcoin/script.h>
 #include <ccan/array_size/array_size.h>
+#include <ccan/build_assert/build_assert.h>
 #include <ccan/mem/mem.h>
 #include <ccan/tal/str/str.h>
+#include <common/htlc_state.h>
 #include <common/key_derive.h>
 #include <common/onionreply.h>
 #include <common/version.h>
@@ -858,6 +860,14 @@ static struct migration dbmigrations[] = {
 
     /* Issue #4887: reset the payments.id sequence after the migration above. Since this is a SELECT statement that would otherwise fail, make it an INSERT into the `vars` table.*/
     {SQL("/*PSQL*/INSERT INTO vars (name, intval) VALUES ('payment_id_reset', setval(pg_get_serial_sequence('payments', 'id'), COALESCE((SELECT MAX(id)+1 FROM payments), 1)))"), NULL},
+
+    /* Issue #4901: Partial index speeds up startup on nodes with ~1000 channels.  */
+    {&SQL("CREATE INDEX channel_htlcs_speedup_unresolved_idx"
+	 "    ON channel_htlcs(channel_id, direction)"
+	 " WHERE hstate NOT IN (9, 19);")
+	[BUILD_ASSERT_OR_ZERO( 9 == RCVD_REMOVE_ACK_REVOCATION) +
+	 BUILD_ASSERT_OR_ZERO(19 == SENT_REMOVE_ACK_REVOCATION)],
+     NULL},
 };
 
 /* Leak tracking. */
--- a/wallet/wallet.c
+++ b/wallet/wallet.c
@@ -2734,10 +2734,16 @@ bool wallet_htlcs_load_in_for_channel(struct wallet *wallet,
 					     " FROM channel_htlcs"
 					     " WHERE direction= ?"
 					     " AND channel_id= ?"
-					     " AND hstate != ?"));
+					     " AND hstate NOT IN (?, ?)"));
 	db_bind_int(stmt, 0, DIRECTION_INCOMING);
 	db_bind_u64(stmt, 1, chan->dbid);
-	db_bind_int(stmt, 2, SENT_REMOVE_ACK_REVOCATION);
+	/* We need to generate `hstate NOT IN (9, 19)` in order to match
+	 * the `WHERE` clause of the database index; incoming HTLCs will
+	 * never actually get the state `RCVD_REMOVE_ACK_REVOCATION`.
+	 * See https://sqlite.org/partialindex.html#queries_using_partial_indexes
+	 */
+	db_bind_int(stmt, 2, RCVD_REMOVE_ACK_REVOCATION); /* Not gonna happen.  */
+	db_bind_int(stmt, 3, SENT_REMOVE_ACK_REVOCATION);
 	db_query_prepared(stmt);
 
 	while (db_step(stmt)) {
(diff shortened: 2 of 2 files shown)


wallet/db.c
```diff
@@ -858,6 +858,15 @@ static struct migration dbmigrations[] = {
 
     /* Issue #4887: reset the payments.id sequence after the migration above. Since this is a SELECT statement that would otherwise fail, make it an INSERT into the `vars` table.*/
     {SQL("/*PSQL*/INSERT INTO vars (name, intval) VALUES ('payment_id_reset', setval(pg_get_serial_sequence('payments', 'id'), COALESCE((SELECT MAX(id)+1 FROM payments), 1)))"), NULL},
+
+    /* Issue #4901: Partial index speeds up startup on nodes with ~1000 channels.  */
+    {SQL("CREATE INDEX channel_htlcs_speedup_unresolved_idx"
+	 "    ON channel_htlcs(channel_id, direction)"
+	 " WHERE hstate NOT IN (9, 19);"),
+	 /*  9 = RCVD_REMOVE_ACK_REVOCATION
+	  * 19 = SENT_REMOVE_ACK_REVOCATION
```
ZmnSCPxj, line 867: It would be really nice if we could put in the constants here instead of hardcoding the constant values 9 and 19.  Any ideas how to do this?  Could use a function for this migration so we can `db_bind_int`, but in principle a function is not necessary and would be too heavyweight...

vincenzopalazzo: maybe it is something stupid and I'm missing the point, but we could define a `#define RCVD_REMOVE_ACK_REVOCATION 9` somewhere?

Just worry if it is a valid type of `db_bind_int`

ZmnSCPxj: They are already defined here: https://github.com/ElementsProject/lightning/blob/a9e261fc20b22d5a77a55276be442ede5baf4fef/common/htlc_state.h#L22 https://github.com/ElementsProject/lightning/blob/a9e261fc20b22d5a77a55276be442ede5baf4fef/common/htlc_state.h#L36

The `#define` would duplicate the above declaration, and if not carefully ordered relative to the above, would cause a syntax error.  We also prefer to use the above `enum` declaration if possible.

whitslack: Rather than asserting the values of these constants in a comment, you could use a compile-time assertion. I see you have a `BUILD_ASSERT` macro presumably for that purpose. `wallet/wallet.h` does contain assertions of the numeric values of these enumerators, but you could duplicate those assertions here in `wallet/db.c` at zero cost to make absolutely certain that the enumerators retain the values you're expecting.

ZmnSCPxj: It is not really about *asserting* the values, it is more that stylistically, I would prefer not to see hardcoded numbers in code.  Ideally I would be able to do something like `tal_fmt("CREATE INDEX blah blah WHERE hstate NOT IN (%d, %d);;", RCVD_REMOVE_ACK_REVOCATION, SENT_REMOVE_ACK_REVOCATION)` but that cannot work as this is a static data structure and this is C, sigh.  C++ initializers...

I think `BUILD_ASSERT` would have to be outside the `db_migrations` array, too.  And the point is not to check the numbers (it is already done elsewhere, as you note) but as a reminder to future readers of the code what these hardcoded `9` and `19` are.  If `BUILD_ASSERT` has to be done outside of `db_migrations`, then as `db_migrations` grows, the explanation of the hardcoded `9` and `19` becomes more distant from the query.  In that respect a comment is superior as it can sit right next to that query in perpetuam (and the check is already done elsewhere).

whitslack: Then you should at least put a comment adjacent to the assertions in `wallet/wallet.h` to help future coders understand what they will be breaking if they violate the assertions.

Are you certain you can't put an assertion inside your `const char *[]` initializer? What about using a compound statement expression for the initializer value? `({ BUILD_ASSERT(…); "string constant here"; })`

ZmnSCPxj: Does not work, not allowed in initializers.  Like I said, C++ initializers allow this, but not C, because C++ actually hooks an extra function into `main` that calls all initializers, but this it not available in C.

```c
/*example.c*/
#define BUILD_ASSERT(cond) \
        do { (void) sizeof(char [1 - 2*!(cond)]); } while(0)

struct example {
        const char *string;
};

struct example db_migrations[] = {
        {({BUILD_ASSERT(1); "example"; })},
};
```

```
$ gcc example.c
example.c:10:3: error: braced-group within expression allowed only inside a function
   10 |  {({BUILD_ASSERT(1); "example"; })},
      |   ^
```

ZmnSCPxj: > Then you should at least put a comment adjacent to the assertions in `wallet/wallet.h` to help future coders understand what they will be breaking if they violate the assertions.

We already have a warning comment in `common/htlc_state.h` for this, which is more important since this is where the constants are actually defined: https://github.com/ElementsProject/lightning/blob/a9e261fc20b22d5a77a55276be442ede5baf4fef/common/htlc_state.h#L5-L8

ZmnSCPxj: `BUILD_ASSERT_OR_ZERO` works:

```c
/* example.c */
#include<stddef.h>

#define BUILD_ASSERT_OR_ZERO(cond) \
        (sizeof(char [1 - 2*!(cond)]) - 1)

enum example {
        EXAMPLE_ZERO,
        EXAMPLE_ONE
};

struct db_entry {
        const char *query;
        const char *foo;
};

struct db_entry db_migrations[] = {
        { "string"
        + BUILD_ASSERT_OR_ZERO(EXAMPLE_ZERO == 0)
        + BUILD_ASSERT_OR_ZERO(EXAMPLE_ONE == 1),
          NULL
        },
};

int main() {
        return 0;
}
```

```
$ gcc -c example.c
```

Still suboptimal though since we do not really want to *assert*, we want to not have to use hardcoded numbers in the source code...

ZmnSCPxj:
Modified to use `BUILD_ASSERT_OR_ZERO`  Still non-ideal since ideally the numbers `9` and `19` should never appear in the source code.

wallet/db.c
ZmnSCPxj: Except clang hates it...

```
wallet/db.c:868:2: error: adding 'unsigned long' to a string does not append to the string [-Werror,-Wstring-plus-int]
        + BUILD_ASSERT_OR_ZERO( 9 == RCVD_REMOVE_ACK_REVOCATION)
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
wallet/db.c:868:2: note: use array indexing to silence this warning
        + BUILD_ASSERT_OR_ZERO( 9 == RCVD_REMOVE_ACK_REVOCATION)
        ^
        [                                                       ]
```

bleah.....

whitslack: What about `&"string"[BUILD_ASSERT_OR_ZERO(…)]`?

ZmnSCPxj: Current version already does that.

cdecker:
ACK https://github.com/ZmnSCPxj/lightning/commit/7c63ce43bfed4d2dd39b93236bc1ba52fa6ac8f4

PR #4926, opened 2021-11-23
doc/PLUGINS.md: Clarify that `db_write` hook only works on SQLITE3.

ZmnSCPxj:
I noticed that `wallet/db_postgres.c` never actually calls `db_changes_add`.

PostgreSQL arguably has a better replication system (a PostgreSQL cluster)
than what `db_write` hook can offer, so rather than make `db_write` work on
PostgreSQL, just document that it does not actually work there.

ChangeLog-none

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -939,7 +939,11 @@ hook subscribers would not get called.
 
 ### `db_write`
 
-This hook is called whenever a change is about to be committed to the database.
+This hook is called whenever a change is about to be committed to the database,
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4926/commits/cdae84d9a8d8be04176acd2be1ff3831afe36c0f

Good information to have in the doc.

ZmnSCPxj:
Eh, merging in despite the CI fail, this is docs, not code.

Issue #4924, opened 2021-11-22
Use lntbs for invoices on signet


vertiond:
Invoices generated on signet have the prefix `lntb`.  LND for example doesn't recognize these invoices as signet and will not pay them.

[Bolt #11](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md#human-readable-part) defines the HRP prefix for invoices as `lntbs`.  Previous discussion [here](https://github.com/lightning/bolts/pull/844).

PR #4929, opened 2021-11-24
bitcoin/chainparams.c: Change `signet` BIP173 name to `tbs`.

ZmnSCPxj:
Fixes: #4924

ChangeLog-Changed: `signet` addresses and invoices now use `tbs` instead of `tb`.

### DIFF
--- a/bitcoin/chainparams.c
+++ b/bitcoin/chainparams.c
@@ -79,7 +79,7 @@ const struct chainparams networks[] = {
 			   .bip32_privkey_version = BIP32_VER_TEST_PRIVATE},
      .is_elements = false},
     {.network_name = "signet",
-     .bip173_name = "tb",
+     .bip173_name = "tbs",
      .bip70_name = "signet",
      // 00000008819873e925422c1ff0f99f7cc9bbb232af63a077a480a3633bee1ef6
      .genesis_blockhash = {{{.u.u8 = {0xf6, 0x1e, 0xee, 0x3b, 0x63, 0xa3, 0x80,


ZmnSCPxj:
This changes onchain bech32 addresses too.  I could not find anything on the Internet about the "standard" bech32 address for signet, unfortunately.  If the signet onchain addresses must still be `tb` while Lightning invoices are `tbs`, that is going to be a larger effort to change.

bitcoin/chainparams.c
```diff
@@ -79,7 +79,7 @@ const struct chainparams networks[] = {
 			   .bip32_privkey_version = BIP32_VER_TEST_PRIVATE},
      .is_elements = false},
     {.network_name = "signet",
-     .bip173_name = "tb",
+     .bip173_name = "tbs",
```
vincenzopalazzo, line 82: From the issue

>LND for example doesn't recognize these invoices as signet and will not pay them.

I read some time ago a conversation in IRC where if I remember correctly the cause is a name collation between an lnd team decision and others impl.

Maybe this is acceptable behavior and the error here is the nomenclature in lnd?

I don't know if it is right at all, or I remember somethings not true

vincenzopalazzo (commented):
Just a small comment

cdecker:
Well, seems we lost the battle to keep the prefixes in sync with bitcoin, which explicitly chose to reuse the `tb` prefix and the collision.

ACK https://github.com/ElementsProject/lightning/pull/4929/commits/3cbe624e0efc015832ae210b51f56957c70e331f

andrewtoth:
err... not sure this should have been merged. Signet is unusable now on master. No signet faucet or wallet will recognize `tbs`, so unless you already had signet coins in the wallet already you can't use signet :/.

PR #4931, opened 2021-11-24
Fix slow memleak in DEVELOPER mode

rustyrussell:
1G of backtraces seems a bit excessive after 1 month of running!


lightningd/log.c
```diff
@@ -55,12 +55,34 @@ struct log_book {
...
+	lp->refcnt = 1;
+	lp->prefix = tal_strdup(lp, prefix);
+	return lp;
+}
+
+static void log_prefix_drop(struct log_prefix *lp)
+{
+	if (--lp->refcnt == 0)
+		tal_free(lp);
+}
```
vincenzopalazzo, line 77: Looks like Valgrind considered this as invalid behavior?

rustyrussell: Yes, and it's right: access after free. We need to do this a bit differently, will address tomorrow!

vincenzopalazzo: For sure this is a very stupid question, but I'm missing the point of the code that makes valid happy again.

I know, it is stupid (maybe)

vincenzopalazzo (approved):
I can not wait until tomorrow to review this PR, this memory stuff is so cool.

ack https://github.com/ElementsProject/lightning/pull/4931/commits/867b7da7f769c75521999f6328900b61da5f6e02

Valgrind is happy again, and CI looks like an unrelated failure, some daemon dead during the test.

Wondering if we try to downgrade the number of parallel tests in the Ci and see if the memory consumption is the problem?

ZmnSCPxj:
>  memory stuff is so cool.

_\*shivers in 20 years of Assembly, C, and C++ coding\*_

PR #4935, opened 2021-11-27
Reduce mem usage for CI with valgrind

rustyrussell:
This should reduce the flakes we're getting due to kill -9, which I suspect is OOM killer.

I printed out the total %mem, mem and vsz every 5 seconds and ran the full testsuite with VALGRIND=1 SLOW_MACHINE=1 (non-parallel).  The largest mem users were those creating many nodes (of course), and this should help with that.

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -640,6 +640,7 @@ class LightningNode(object):
                  allow_bad_gossip=False,
                  db=None, port=None, disconnect=None, random_hsm=None, options=None,
                  jsonschemas={},
+                 valgrind_plugins=True,
                  **kwargs):
         self.bitcoin = bitcoind
         self.executor = executor
@@ -690,11 +691,14 @@ class LightningNode(object):
         if dsn is not None:
             self.daemon.opts['wallet'] = dsn
         if valgrind:
+            trace_skip_pattern = '*python*,*bitcoin-cli*,*elements-cli*'
+            if not valgrind_plugins:
+                trace_skip_pattern += ',*plugins*'
             self.daemon.cmd_prefix = [
                 'valgrind',
                 '-q',
                 '--trace-children=yes',
-                '--trace-children-skip=*python*,*bitcoin-cli*,*elements-cli*',
+                '--trace-children-skip={}'.format(trace_skip_pattern),
                 '--error-exitcode=7',
                 '--log-file={}/valgrind-errors.%p'.format(self.daemon.lightning_dir)
             ]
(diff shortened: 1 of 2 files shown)


contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -690,11 +691,14 @@ def __init__(self, node_id, lightning_dir, bitcoind, executor, valgrind, may_fai
         if dsn is not None:
             self.daemon.opts['wallet'] = dsn
         if valgrind:
+            trace_pattern = '*python*,*bitcoin-cli*,*elements-cli*'
+            if valgrind_plugins is False:
```
cdecker, line 695: `if not valgrind_plugins`

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -690,11 +691,14 @@ def __init__(self, node_id, lightning_dir, bitcoind, executor, valgrind, may_fai
...
+                trace_pattern += ',*plugins*'
```
cdecker, line 696: Maybe this should be called `trace_skip_pattern`

cdecker (approved):
Just some naming nits and CI is sad due to a lint, but looking good otherwise

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -1280,12 +1284,20 @@ def get_nodes(self, num_nodes, opts=None):
 
         assert len(opts) == num_nodes
 
+        # Only trace one random node's plugins, to avoid OOM.
+        if SLOW_MACHINE:
+            valgrind_plugins = [False] * num_nodes
+            valgrind_plugins[random.randint(0, num_nodes-1)] = True
```
vincenzopalazzo, line 1290:
```suggestion
            valgrind_plugins[random.randint(0, num_nodes - 1)] = True
```

I suspect that this is the cause of the CI failure (sadly)

vincenzopalazzo (approved):
LGTM

Today I was looking in some tools for Github action that prints some env information at the end of the test. This could be useful I think, but there are nothings around.

rustyrussell:
Thanks, pushed those fixes!

Hope this will unstick a heap of other PRs

ZmnSCPxj (approved):
ACK 7f6276461115dcc02679ce3f00c5c31b9e6634cd

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4935/commits/7f6276461115dcc02679ce3f00c5c31b9e6634cd

PR #4941, opened 2021-11-30
pytest: fix flake in testing.

rustyrussell:
As noted in https://github.com/ElementsProject/lightning/pull/4897/commits/0a406230d051a83f9f029054efe2dd24e87887a8#diff-5871d4c569454db5e625383975462132da0bd03d32df145d8d72d8fafd86d952R3544-R3546

Turns out we sometimes hang up before l2 sees the previous tx revoked,
so we get a normal unilateral close, not a cheat.

Reported-by: Simon Vrouwe
Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4941/commits/23bad330a8cf4c153f990d5a83336a0e08a3bcfa

Mh, I see some things similar with lnprototest in the protocol tests, maybe there is the same things. I will check. Thanks.

Issue #4939, opened 2021-11-29
v0.10.2 permission issue during install


iamthen0nce:
When building c-lightning v0.10.2 from source, we run into this issue:

```
/bin/sh: common/htlc_state_names_gen.h: Permission denied
```

when executing `make` as unprivileged user.

Not sure if it's a bug or if the doc just has to be adapted to tell users to do `sudo make` instead.

```
Build lightning:

./configure
make
sudo make install
```
should probably become
```
Build lightning:

./configure
sudo make
sudo make install
```

vincenzopalazzo:
adding `sudo make` to compile the sources doesn't make sense IMO, instead, run the command `sudo make install` make sense because we are asking to copy some files in some special directory of the system.

In addition, the command make will generate some source code files so the file `common/htlc_state_names_gen.h` is autogenerated, one thing that I have in mind is if you have on your machine the write permission.

iamthen0nce:
Okay; wasn't sure if I could just ignore that error and run `sudo make install` directly. But normal `make` threw this error that's all I can tell you. Wasn't like that in the past.

cdecker:
I'm almost certain that running with `sudo` isn't what you want. This looks like it is trying to execute a header file, potentially because the tool that we actually wanted to call (i.e., `/path/to/tool.sh common/htlc_state_names_gen.h`) can't be found, leaving it blank, and thus causing us to instead attempting to execute the argument instead (i.e., `./common/htlc_state_names_gen.h`). This fails because the file being passed as an argument is not marked as executable (and that's a good thing).

We should rather be checking if the tool exists and if not report it as an error.

andrewtoth:
> wasn't sure if I could just ignore that error and run `sudo make install` directly.

`sudo make install` copies the compiled binaries into places where they will be found by other system users' paths. If `make` fails then the binaries aren't compiled successfully, so you will be installing the previous version again.

Does `ccan/ccan/cdump/tools/cdump-enumstr common/htlc_state.h` output some C code in the terminal for you, or does it return an error?

ZmnSCPxj:
`sudo make` is almost always a mistake and it would be a bug to require that.

Can you provide the *full* logs of `configure` and `make`?  More likely something else is broken.

iamthen0nce:
> > wasn't sure if I could just ignore that error and run `sudo make install` directly.
> 
[...]

PR #4944, opened 2021-12-02
make: remove generated files when running clean

andrewtoth:
This updates `make clean` to also remove all generated source and header files. This would have resolved https://github.com/ElementsProject/lightning/issues/4939 with a `make clean`, but instead the user ended up nuking the repo and cloning again.

There is a comment about only removing generated files with `make maintainer-clean`:
> This command is intended for maintainers to use; it deletes files that may need special tools to rebuild.

I checked and all generated files are not checked in source control, so anything that succeeds in building already has all available tools. Not sure if that comment is out of date or if there is some history behind it that I'm not aware of, but it seems that it is safe to remove these generated source and header files on `make clean`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -610,7 +610,6 @@ distclean: clean
 maintainer-clean: distclean
 	@echo 'This command is intended for maintainers to use; it'
 	@echo 'deletes files that may need special tools to rebuild.'
-	$(RM) $(ALL_GEN_HEADERS) $(ALL_GEN_SOURCES)
 
 # We used to have gen_ files, now we have _gen files.
 obsclean:
@@ -618,6 +617,7 @@ obsclean:
 
 clean: obsclean
 	$(RM) $(CCAN_OBJS) $(CDUMP_OBJS) $(ALL_OBJS)
+	$(RM) $(ALL_GEN_HEADERS) $(ALL_GEN_SOURCES)
 	$(RM) $(ALL_PROGRAMS)
 	$(RM) $(ALL_TEST_PROGRAMS)
 	$(RM) $(ALL_FUZZ_TARGETS)
--- a/wallet/Makefile
+++ b/wallet/Makefile
@@ -46,7 +46,8 @@ wallet/db_%_sqlgen.c: wallet/statements_gettextgen.po devtools/sql-rewrite.py $(
 		$(call VERBOSE,"sql-rewrite $@",devtools/sql-rewrite.py wallet/statements_gettextgen.po $* > $@ && $(call SHA256STAMP,//,)); \
 	fi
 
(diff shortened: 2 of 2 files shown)


ZmnSCPxj:
`maintainer-clean` is mandated by GNU coding standards: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html

Of course, we no longer actually target GNU coding standards.  Nevertheless, distro packaging tools often expect packages to support at least some of the GNU coding standards.

Arguably, things like `mako` and `mrkd` or even `automake` and `autoconf` should *not* be required to merely *build* a standard source release of a package; they are strictly developer tools.  Unfortunately we do not use autotools, which means that we require users to install `mako` and `mrkd` and `automake` and `autoconf`. Our externals, quite rationally, use autotools, but we ourselves do not, so we cannot have autotools automatically include our upstream libraries with the `configure` and `Makefile.in` and etc etc files already generated, instead we have our own way of generating the tarball which does *not* include those.

Thus, ironically CLBOSS (which *does* use autotools) actually does not require `autoconf` or `automake` to build from the standard source release (because the autotools are *not* a build tool, they are a tool for building build tools that will be included in a standard source release), because even though it depends on libsecp256k1, it can include the generation of the `configure` of the dependency with its own generation of `configure`.

In practice, for `git`-based projects, `maintainer-clean` is the same as `git clean -fxd`.

andrewtoth:
@ZmnSCPxj thanks for the detailed response! I learned a lot. Reading the GNU standards you linked, under `clean` it states:
```
Delete all files in the current directory that are normally created by building the program. Also delete files in other directories if they are created by this makefile. However, don’t delete the files that record the configuration. Also preserve files that could be made by building, but normally aren’t because the distribution comes with them.
```
Looking at the source release distribution, the generated headers and sources are not included, nor do they record configuration. So it does seem appropriate that these files are removed by `make clean`. I did update this to keep the `maintainer-clean` comment though.

cdecker:
FWIW I usually differentiate between source tarball (as the tar file distros build their binaries from) and the source tree in git (where developers will have to handle eventual conflicts).

I have a script that builds tarballs with generated files, without dependencies on `mako`, `mrkd` and `git` (and with all submodules), allowing building without network connection (which Launchpad requires fwiw).

rustyrussell:
This change is now accurate, since we don't check in generated files any more.

Ack: https://github.com/ElementsProject/lightning/commit/a0cd4773002184da38af3e963628ffe3c5a94a7b

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4944/commits/a0cd4773002184da38af3e963628ffe3c5a94a7b

PR #4948, opened 2021-12-02
status: do not log ping and pong packets

m-schmoock:
Debug logging of ping and pong packets turned out to be
too excessive since this is now done on all channels
and without config option.

Changelog-None

### DIFF
--- a/common/status.c
+++ b/common/status.c
@@ -98,13 +98,25 @@ static void status_io_full(enum log_level iodir,
 	status_send(take(towire_status_io(NULL, iodir, peer, who, p)));
 }
 
+static bool status_peer_io_filter_packettype(const u8 *p)
+{
+	int msg_type = fromwire_peektype(p);
+	switch (msg_type) {
+	case WIRE_PING:
+	case WIRE_PONG:
+		return true;
+	}
+	return false;
+}
+
 static void status_peer_io_short(enum log_level iodir,
 				 const struct node_id *peer,
 				 const u8 *p)
 {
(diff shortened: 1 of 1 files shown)


common/status.c
```diff
@@ -98,13 +98,28 @@ static void status_io_full(enum log_level iodir,
...
+	int msg_type = fromwire_peektype(p);
+	switch (msg_type) {
+	case WIRE_PING:
+	case WIRE_PONG:
+		return true;
+	default:
+		break;
+	}
+	return false;
+
```
vincenzopalazzo, line 112:
```suggestion
	int msg_type = fromwire_peektype(p);
	switch (msg_type) {
	case WIRE_PING:
	case WIRE_PONG:
		return true;
	}
	return false;

```

Why not this?

m-schmoock: Jup, I was under the false impression that our code always has the default case, which doesn't seem to be true. So yes, your code looks better ;)

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4948/commits/52292d54a86d7c5f2f5260ba53b256754f8eedf8

cdecker:
I think a more durable solution would be to introduce a `trace` level, that is more verbose than `debug`, and have all these spammy messages be in there. Wire in/out would then just be on that level, since even ping/pong messages can be important to debugging (I found a couple of issues this way with saturated connections where pings weren't getting replies).

rustyrussell:
Meh, we have log_io if you really want to see everything.  Gotta stop somewhere.

PR #4949, opened 2021-12-02
hsmd: Add hsmd_new_channel and hsmd_ready_channel messages for validating signers

ksedgwic:
This PR adds two messages (& replies) to `hsmd_wire.csv`. 

Summary of changes needed to support fully validating signers here:
https://gitlab.com/-/snippets/2179274

### DIFF
--- a/hsmd/libhsmd.c
+++ b/hsmd/libhsmd.c
@@ -91,6 +91,7 @@ bool hsmd_check_client_capabilities(struct hsmd_client *client,
 
 	case WIRE_HSMD_GET_PER_COMMITMENT_POINT:
 	case WIRE_HSMD_CHECK_FUTURE_SECRET:
+	case WIRE_HSMD_READY_CHANNEL:
 		return (client->capabilities & HSM_CAP_COMMITMENT_POINT) != 0;
 
 	case WIRE_HSMD_SIGN_REMOTE_COMMITMENT_TX:
@@ -124,6 +125,7 @@ bool hsmd_check_client_capabilities(struct hsmd_client *client,
 	case WIRE_HSMD_CUPDATE_SIG_REPLY:
 	case WIRE_HSMD_CLIENT_HSMFD_REPLY:
 	case WIRE_HSMD_NEW_CHANNEL_REPLY:
+	case WIRE_HSMD_READY_CHANNEL_REPLY:
 	case WIRE_HSMD_NODE_ANNOUNCEMENT_SIG_REPLY:
 	case WIRE_HSMD_SIGN_WITHDRAWAL_REPLY:
 	case WIRE_HSMD_SIGN_INVOICE_REPLY:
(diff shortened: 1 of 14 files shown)


hsmd/libhsmd.c
```diff
@@ -293,6 +295,61 @@ static u8 *handle_new_channel(struct hsmd_client *c, const u8 *msg_in)
...
+	/* Stub implementation */
+
+	/* Fail fast if any values are uninitialized or obviously wrong. */
+	assert(amount_sat_greater(channel_value, AMOUNT_SAT(0)));
+	assert(amount_sat_to_msat(&value_msat, channel_value));
+	assert(amount_msat_less_eq(push_value, value_msat));
+	assert(!mem_is_zero(&funding_txid, sizeof(funding_txid)));
+	assert(local_to_self_delay > 0);
+	assert(!mem_is_zero(&remote_basepoints, sizeof(remote_basepoints)));
+	assert(!mem_is_zero(&remote_funding_pubkey, sizeof(remote_funding_pubkey)));
```
rustyrussell, line 347: Definitionally, these are valid points since we've just used fromwire() on them.  However, this assertion doesn't *hurt*...

lightningd/dual_open_control.c
```diff
@@ -645,13 +645,23 @@ openchannel2_hook_cb(struct openchannel2_payload *payload STEALS)
...
 	}
 
+	// Determine the wallet index for our upfront_shutdown_script, UINT32_MAX if not found.
+	u32 our_shutdown_script_wallet_index = UINT32_MAX;
+	bool is_p2sh;
+	wallet_can_spend(
+		dualopend->ld->wallet,
+		payload->our_shutdown_scriptpubkey,
+		&our_shutdown_script_wallet_index,
+		&is_p2sh);
```
rustyrussell, line 655: Relying on internals of wallet_can_spend is a bad idea: if it returns False, set our_shutdown_script_wallet_index to UINT32_MAX.

lightningd/dual_open_control.c
```diff
@@ -2602,9 +2612,19 @@ static struct command_result *json_openchannel_init(struct command *cmd,
...
 			= tal_steal(oa, our_upfront_shutdown_script);
 
+	// Determine the wallet index for our upfront_shutdown_script, UINT32_MAX if not found.
+	u32 our_upfront_shutdown_script_wallet_index = UINT32_MAX;
+	bool is_p2sh;
+	wallet_can_spend(
+		cmd->ld->wallet,
+		oa->our_upfront_shutdown_script,
+		&our_upfront_shutdown_script_wallet_index,
+		&is_p2sh);
```
rustyrussell, line 2622: Here too...

lightningd/dual_open_control.c
```diff
@@ -3059,9 +3079,19 @@ static struct command_result *json_queryrates(struct command *cmd,
...
 	psbt = create_psbt(tmpctx, 0, 0, 0);
 
+	// Determine the wallet index for our upfront_shutdown_script, UINT32_MAX if not found.
+	u32 our_upfront_shutdown_script_wallet_index = UINT32_MAX;
+	bool is_p2sh;
+	wallet_can_spend(
+		cmd->ld->wallet,
+		oa->our_upfront_shutdown_script,
+		&our_upfront_shutdown_script_wallet_index,
+		&is_p2sh);
```
rustyrussell, line 3089: Here three...

lightningd/dual_open_control.c
```diff
@@ -3242,6 +3272,15 @@ void peer_restart_dualopend(struct peer *peer,
...
 				      channel->opener, LOCAL);
 
+	// Determine the wallet index for our upfront_shutdown_script, UINT32_MAX if not found.
+	u32 local_shutdown_script_wallet_index = UINT32_MAX;
+	bool is_p2sh;
+	wallet_can_spend(
+		peer->ld->wallet,
+		channel->shutdown_scriptpubkey[LOCAL],
+		&local_shutdown_script_wallet_index,
+		&is_p2sh);
```
rustyrussell, line 3282: Here four...

lightningd/opening_control.c
```diff
@@ -668,9 +670,18 @@ openchannel_hook_final(struct openchannel_hook_payload *payload STEALS)
 		uc->got_offer = true;
 	}
 
+	/* Determine the wallet index for our upfront_shutdown_script,
+	 * UINT32_MAX if not found. */
+	wallet_can_spend(
+		payload->openingd->ld->wallet,
+		our_upfront_shutdown_script,
+		&upfront_shutdown_script_wallet_index,
+		&is_p2sh);
```
rustyrussell, line 679: etc.

lightningd/dual_open_control.c
rustyrussell: Actually, even better, you can make our_shutdown_script_wallet_index an optional parameter in the wire protocol (using the notation `?u32`) which makes it unambiguous vs a magic value?

rustyrussell (commented):
Generally a fine idea, just some minor changes to make it a bit cleaner...

hsmd/libhsmd.c
devrandom: seems to me that either:
- the caller is using a wrong csv definition file / generated wire code, in which case the packet will be short and will error elsewhere
- the caller is using the right definition file, in which case it's somewhat unlikely that zeros will end up in those fields

so these two checks are unlikely to catch a real error.

ksedgwic: removed

lightningd/dual_open_control.c
ksedgwic: fixed

lightningd/dual_open_control.c
ksedgwic: fixed

lightningd/dual_open_control.c
ksedgwic: fixed

lightningd/dual_open_control.c
ksedgwic: fixed

lightningd/opening_control.c
ksedgwic: fixed all

PR #4950, opened 2021-12-03
[DONTMERGE] check: make sure all files outside contrib/ include "config.h" first.

rustyrussell:
And turn "" includes into full-path (which makes it easier to put
config.h first, and finds some cases check-includes.sh missed
previously).

config.h sets _GNU_SOURCE which really needs to be done before any
'#includes': we mainly got away with it with glibc, but other platforms
like Alpine may have stricter requirements.

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

PR #4951, opened 2021-12-04
Code cleanups

rustyrussell:
This is a rebased version of the #4950 (which is based on the last release, for those trying to build), plus removal of unused functions.

Changelog-None: cleanups


lightningd/plugin.c
```diff
@@ -110,6 +110,30 @@ static void plugin_check_subscriptions(struct plugins *plugins,
 	}
 }
 
+static bool plugins_any_in_state(const struct plugins *plugins,
```
ZmnSCPxj, line 113: Why the code movement?  Why not just insert a forward declare here?

(code movements like this require careful review, it is an easy way to insert an exploit since if the contributor says "just a code move" reviewers might not check every character of every line and validate that it is exactly the same)

(of course, rusty does have commit rights to the repo, so...)

wallet/wallet.c
```diff
@@ -1194,6 +1175,37 @@ static bool wallet_channel_load_inflights(struct wallet *w,
 	return ok;
 }
 
+static bool wallet_channel_config_load(struct wallet *w, const u64 id,
```
ZmnSCPxj, line 1178: Similar.  Why the code movement?  Why not just forward declare here?

lightningd/plugin.c
rustyrussell: Ordering, since it's now static.  I wish we had sufficient tooling to automatically check them though :(

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4951/commits/434b207b1a440dd90ec3210d559ad17a05e5d1ce

PR #4955, opened 2021-12-05
pytest: fix occasional flake in test_announce_address

rustyrussell:
We make sure the gossip msg was sent, but the other node might not
have digested it yet:

```
        # Check other node can parse these
>       addresses = l2.rpc.listnodes(l1.info['id'])['nodes'][0]['addresses']
E       KeyError: 'addresses'
```

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4955/commits/873b9964ebf762b9a225363c265cf947d346846b

PR #4956, opened 2021-12-05
introduce tmpdir_mkstemp to respect $TMPDIR in unit tests

whitslack:
Various unit tests were creating temporary files unconditionally in `/tmp` and were not cleaning up after themselves. This PR: (1) introduces a new variant of `mkstemp(3p)` that respects the `TMPDIR` environment variable and (2) uses it in the offending unit tests. This allows each test run to use a dedicated `TMPDIR` that can be cleaned up after the run.

### Case Study

**Without this patch,** the Gentoo ebuild for C-Lightning, when executed with tests enabled, leaves some stray files in `/tmp`:

```
-rw------- 1 portage portage  3260  5-Dec-2021 00:33 run-gossip_local.rCWUh5
-rw------- 1 portage portage  2527  5-Dec-2021 00:33 run-route-gossipstore.FU9Qkc
-rw------- 1 portage portage 22573  5-Dec-2021 00:33 run-route-overlong.ryaSQq
-rw------- 1 portage portage  2527  5-Dec-2021 00:33 run-route-specific-gossipstore.bJ91Fd
```

**With this patch applied,** these temporary files land in `/var/tmp/portage/net-p2p/c-lightning-…/temp` and thus are automatically nuked by Portage at the completion of the build.


vincenzopalazzo (approved):
LGTM

ack https://github.com/ElementsProject/lightning/pull/4951/commits/434b207b1a440dd90ec3210d559ad17a05e5d1ce

Do you think that we need also to add a remove procedure for each test in this PR?

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4956/commits/d8fc3e09a7f895342619ce402b13a29a450e8e36

whitslack:
> ack [434b207](https://github.com/ElementsProject/lightning/commit/434b207b1a440dd90ec3210d559ad17a05e5d1ce)

@vincenzopalazzo: Was that the wrong commit hash you ACK'd?

> Do you think that we need also to add a remove procedure for each test in this PR?

In general I'd advocate for always being a good steward of your temporary files, even in unit tests. It's less concerning when the files can be steered into a custom temp dir. Your call on whether you want to implement explicit cleanup. Consider that the "right way" to do it is to install a signal handler that will unlink the temporary files even if the unit test is killed by a signal (including `SIGSEGV`, `SIGILL`, or `SIGBUS`!).

The "even righter" way to do it is to use a syscall that creates an *anonymous* temporary file, as such file will be removed from disk by the kernel automatically in every case, even if your process is killed by `SIGKILL`, but then you can't open the same file in another file description without using the `proc` file system, which may not be possible on all platforms that C-Lightning targets.

PR #4957, opened 2021-12-06
doc/Makefile: speed up schema check, only do on check-doc.

rustyrussell:
Don't do this on every build, it takes 4 seconds.  And split it
into individual targets, so make can parallelize (1.6 seconds here).

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4957/commits/1feb8d9c06792e62d45c84fcd244f492ed7f7844

You anticipated my today's task inside my TODO list. Sorry if I missed this part in my original PR, and thanks to catching it

PR #4965, opened 2021-12-08
doc: force mistune==0.8.4 during pip package resolution.

vincenzopalazzo:
This PR will add tips in the docs because on IRC there are some reports on bad resolution versions caused by this bug of `mrkd` https://github.com/refi64/mrkd/pull/6#issue-1073838481

Basically, the package has no fixed version and this gives them the power to pip to pull the last version of the `mistune` dependence, which will break the `mrkd` package.

P.S: The change should be forwarded also to all the docker files, but we hope that release the new version of the package takes only a few days

Changelog-None: doc: force mistune==0.8.4 during pip package resolution.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -40,7 +40,7 @@ Get dependencies:
       autoconf automake build-essential git libtool libgmp-dev libsqlite3-dev \
       python3 python3-mako python3-pip net-tools zlib1g-dev libsodium-dev \
       gettext
-    pip3 install --user mrkd
+    pip3 install --user mrkd mistune==0.8.4
 
 If you don't have Bitcoin installed locally you'll need to install that
 as well. It's now available via [snapd](https://snapcraft.io/bitcoin-core).


cdecker:
Does installing from requirements.lock file work correctly? Could you pin the version there too otherwise?

jsarenik:
@cdecker The version seems to be already pinned in `requirements.lock`: https://github.com/ElementsProject/lightning/blob/master/requirements.lock#L82

Just that whenever someone's shell is outside of `lightning` directory where this `requirements.lock` is located, the newer version gets installed.

ACK 4a75b91510e62d3e3868f515f7f9e2d2f61dd4c5

vincenzopalazzo:
>Does installing from requirements.lock file work correctly? Could you pin the version there too otherwise?

yes @cdecker, the lock file contains these fixed dependencies already, maybe this is the reason that not all the person will see the same error.

cdecker:
Wouldn't pin `mrkd` to a version that doesn't have a broken dependency be better? That way we'd not artificially install an extra dependency manually to work around broken dependencies in an upstream packet.

vincenzopalazzo:
Yes @cdecker, but the `mrkd` required the last version of the library, in fact, there is the following PR open on the package https://github.com/refi64/mrkd/pull/6#issue-1073838481

I think this is the only way to work around to pip dependencies resolution algorithm? of course also use the requirements.lock it is a solution

cdecker:
Yeah, that was what I was afraid of, mrkd just referencing the latest version. Let's pin mistune and revisit this once it's fixed upstream.

PR #4967, opened 2021-12-08
Removed FIXME to match bolts PR 918

lightning-developer:
It seems as if  [PR 918](https://github.com/lightning/bolts/pull/918) will be merged to the bolts so this requirement will be dropped from the spec and we do not need to have a FIXME in the code base.

Yes I know that this is a somewhat silly PR.


cdecker:
ACK 246d0bfba50b7c4cd43a111a817fb36c2bde4d67

rustyrussell:
Thanks!

PR #4968, opened 2021-12-10
Update .gitignore for tmp files

benthecarman:
`.tmp` files are generated from compiling the docs, these should be ignored by git

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.dSYM
 *.rej
 *.pyc
+*.tmp
 .cppcheck-suppress
 .mypy_cache
 TAGS


ZmnSCPxj (approved):
ACK a7b5e820b78a0c10ddad2e742e3f5a6f551841ee

vincenzopalazzo:
M, The .tmp file should be removed by this part of the make file
https://github.com/ElementsProject/lightning/blob/be196b42197b8c119c733233af1b9858ce66134f/doc/Makefile#L115

Maybe are not inside the .gitignore because these files should not be there? Any error received during the doc-all process?

benthecarman:
@vincenzopalazzo I got them from the build failing of the docs. Figured they are a good thing to ignore

ZmnSCPxj:
Also saw those before, which is why I merged in.  Probably they should be cleaned by `make clean` too, just in case.

vincenzopalazzo:
Mh, IMO there ignoring it can hide some future doc build error, instead of ignoring it I'm more the person that adds this file in the `make clean`. In addition, having the`.tmp` file out of the git ignore help to catch easily some problem in the build process.

BTW, it is already on master, it is good now.

ZmnSCPxj:
Same argument would go for `.o` files and the executable binaries.  I am not convinced.  If the `Makefile` was properly designed, then the lack of `.tmp` file would be readily visible as a `make` error in generating the tmpfile.

vincenzopalazzo:
I agree, but the file `.o` and executable binaries are generated in any case of build success, and it is more common that we push some binary in the git history, that we push some code change with some build error like the build of the doc error.

But I think this is more a philosophic point of view right? I think both solutions makes sense

PR #4969, opened 2021-12-11
fix dockerifle

johnongit:
Add mistune==0.8.4

https://github.com/ElementsProject/lightning/pull/4965/files

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,7 +79,9 @@ RUN git clone --recursive /tmp/lightning . && \
 
 ARG DEVELOPER=0
 ENV PYTHON_VERSION=3
-RUN pip3 install mrkd
+RUN apt-get install -y --no-install-recommends python3-dev
+RUN pip3 install -U pip && pip3 install -r requirements.lock
+
 RUN ./configure --prefix=/tmp/lightning_install --enable-static && make -j3 DEVELOPER=${DEVELOPER} && make install
 
 FROM debian:buster-slim as final


Dockerfile
```diff
@@ -79,7 +79,8 @@ RUN git clone --recursive /tmp/lightning . && \
 
 ARG DEVELOPER=0
 ENV PYTHON_VERSION=3
-RUN pip3 install mrkd
+RUN pip3 install mrkd mistune==0.8.4
```
vincenzopalazzo, line 82: nit: Inside the docker, I think to use the requirements.lock it is better, what do you think?

vincenzopalazzo (approved):

Dockerfile
```diff
@@ -17,16 +17,16 @@ RUN wget -qO /opt/tini "https://github.com/krallin/tini/releases/download/v0.18.
     && echo "12d20136605531b09a2c2dac02ccee85e1b874eb322ef6baf7561cd93f93c855 /opt/tini" | sha256sum -c - \
     && chmod +x /opt/tini
 
-ARG BITCOIN_VERSION=0.18.1
+ARG BITCOIN_VERSION=22.0
```
vincenzopalazzo: I think this is a old commit?

Dockerfile
johnongit: I'm not able to catch it :)
This one works well. I've just added like #4965

vincenzopalazzo: Mh, I think in the docs it is fine, how mentioned in #4965 I didn't update the docker file because we hope that the package will be fixed in the next couple of weeks. If this happens, this docker file need another update,

If you use the following command `pip install -r requirements.lock` we can update the requirements.lock only, that it is tested to each commit in the main repository and I considered it well tested.

niftynei:
Thanks for this fixup. Just a few notes on this PR as presented.

- it contains a merge commit; we don't use merge commits in this repo
- it contains changes that are out of the scope of the proposed fix.

Please update the commit history to fix these issues; ideally there'd be a single line change in a single commit for this PR.

cdecker:
I took the liberty of simplifying and future-proofing the PR as suggested by reviewers.

ACK https://github.com/ElementsProject/lightning/pull/4969/commits/41216174c478df4deaba4d9d049b31b380f146a6

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/4969/commits/41216174c478df4deaba4d9d049b31b380f146a6

PR #4970, opened 2021-12-12
pay: Disable channels that are not in normal state

cdecker:
Since we have the exact state of the channels from the `listpeers`
response we can filter out the ones that are not yet normal or not
anymore. This is more of a safety net, given that the `gossip_store`
should contain local disables, but better not be racy :-)

This was found while working on greenlight, where the time between
receiving a local update and it being reflected in the `gossip_store` can
take some time.

Changelog-None


rustyrussell:
We seem to have a problem with lightningd not terminating.  I'd like to try to reproduce it here (I expect it's the recent shutdown changes), but meanwhile I've kicked CI.

rustyrussell (approved):
Ack b2c3909a83ecebe48b3b5b53e4753b6bdcfa8777

rustyrussell:
Trivial rebase on master...

Ack 4218202ce

PR #4974, opened 2021-12-13
wire: add missing spec patch.

rustyrussell:
You have to manually run "make extract-bolt-csv" to recreate the
spec files, so we didn't notice that I forgot to check this in :(

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4974/commits/348b0bb38571a5d48f4ec20105e73451d1b84ef9

PR #4980, opened 2021-12-17
cln-rs: Building c-lightning plugins and clients in Rust

cdecker:
This is our first attempt at integrating Rust into the c-lightning
tree. The goal is to use Rust in some parts, mostly plugins and
self-contained subsystems, allowing us to move quicker. This PR tries
to do a couple of things:

 - Add Rust tooling support to the build system (keeping `rustc` and
   `cargo` as optional dependencies for now, and conditionally
   compiling plugins if they're present).
 - Build out `cln-plugin` which will serve as the basis for plugins
   written in Rust.
 - Build out `cln-rpc`, with partially generated requests and
   responses in order to enable Rust programs to interact with
   c-lightning.
   
In order to build these top level features, we also had to build a bit
of tooling and infrastructure:

 - `contrib/msggen`: A parser for the JSON Schema files. It parses the
   files into an intermediate representation in memory, making it
   source and sink agnostic. The goal is to be the target for all
   source types (JSON Schema and wire CSVs) and from there generate
   artifacts from it. In this case the `tools/gen-rust-rpc.py` tool
   will parse the Schema files and generate Rust modules from it.
   
## Todo
   
This PR is work in progress. The following is a subset of tasks that
need to be done:

 - [X] Conditionally compile the Rust plugins if we have `cargo` and
       `rustc`
 - [X] Write a codec to read and write JSON-RPC requests/responses
 - [X] Build the plugin framework to dispatch incoming hooks and notifications
 - [ ] Build a connection pool for the RPC interface

The `cln-rpc-grpc` plugin will be in a separate pull request to
facilitate reviews and decouple discussion for the framework from the
actual plugin.


plugins/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+            subscriptions: Subscriptions::default(),
+        }
+    }
+
+    #[allow(unused_mut)]
+    pub async fn run(mut self) -> Plugin<S, I, O> {
+        let (plugin, input) = self.build();
+        tokio::task::spawn(plugin.run(input));
+        tokio::time::sleep(std::time::Duration::from_secs(300)).await;
+        unimplemented!()
```
vincenzopalazzo, line 54: Refused?

cdecker: Yeah, it's work in progress, and therefore in draft mode (not meant to be reviewed yet). That will be fixed before undrafting

vincenzopalazzo: @cdecker yep, I was just curious about the introduction of Rust in c-lightning, and the plugin library that I'm designing too for the https://github.com/laanwj/rust-clightning-rpc crate.

I didn't know that GitHub sends back-end commits without pushing the button "Send review". Sorry, I was just reading the code as "relax reading" e not much for the review

jb55:
configure got mad at me for some reason:
`./configure: line 114: [: missing `]'`

```diff
From 327d8a4c2b5f87f817e246c6bfc7616bde252186 Mon Sep 17 00:00:00 2001
From: William Casarin <jb55@jb55.com>
Date: Mon, 20 Dec 2021 18:15:32 -0800
Subject: [PATCH clightning] fixup! rs: Add Rust and Cargo detection in
 configure script

---
 configure | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/configure b/configure
index 4ff179b74..35c33b3e8 100755
--- a/configure
+++ b/configure
@@ -111,7 +111,7 @@ default_valgrind_setting()
 
 default_rust_setting()
 {
-    if cargo --version > /dev/null 2>&1 || [ $? = 0]; then
+    if cargo --version > /dev/null 2>&1 || [ $? = 0 ]; then
 	echo 1
     else
 	echo 0
-- 
2.32.0
```

jb55:
This will greatly complicate packaging on the distro side, and kill any chance of from-source-reproducibility, at least for now. Will this be required for core functionality in the future? Would be a bit dissapointing if so.

cdecker:
> This will greatly complicate packaging on the distro side, and kill any chance of from-source-reproducibility, at least for now. Will this be required for core functionality in the future? Would be a bit dissapointing if so.

No, it'll remain optional until we've figured out how to get reproducible builds working again. On the compiler / distro side I see less of a problem, given Rust's proliferation I think we can expect them to be widely distributed in distros.

PR #4982, opened 2021-12-18
On mac a difference error is thrown on connection refusal

ddustin:
@cdecker

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -53,7 +53,7 @@ def test_connect(node_factory):
         l1.rpc.connect('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e')
 
     # Should get reasonable error if connection refuse.
-    with pytest.raises(RpcError, match=r'Connection establishment: Connection refused'):
+    with pytest.raises(RpcError, match=r'Connection establishment: (Connection refused|Bad file descriptor)'):
         l1.rpc.connect('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e', 'localhost', 1)
 
     # Should get reasonable error if wrong key for peer.


tests/test_connection.py
```diff
@@ -53,7 +53,7 @@ def test_connect(node_factory):
         l1.rpc.connect('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e')
 
     # Should get reasonable error if connection refuse.
-    with pytest.raises(RpcError, match=r'Connection establishment: Connection refused'):
+    with pytest.raises(RpcError, match=r'Connection establishment: (Connection refused|Bad file descriptor)'):
```
vincenzopalazzo, line 56: In this case, I usualy preferer to add a comment that points out that we can receive different err message for different os. However, maybe we could be interested only in the prefix of the error?

rustyrussell:
This hints at an underlying bug, TBH.  ECONNREFUSED is definitely the error we expect...

cdecker:
Testing on a Mac myself, it seems that something is up with these kinds of errors. Every once in a while I seem to be losing the `hsmd` filedescriptor, which is weird.

Still working on making compilation on m1 solid, but will keep an eye open on why this fails. In the meantime if there's any pointers as to how to debug this, they'd be much appreciated.

rustyrussell:
> Testing on a Mac myself, it seems that something is up with these kinds of errors. Every once in a while I seem to be losing the hsmd filedescriptor, which is weird.

Hmm.  This is usually caused by someone closing the wrong fd (less often, by dup2 on top of the fd).  Try replacing close with closex which does `assert(!getenv("HSM_FD") || fd != atoi(getenv("HSM_FD"))` then setenv that to "3" at the start of connectd?

That will at least tell us who is closing it, if not why they're wrong.  If you can make it really reproducible you can debug the subd and poke around, but that's harder if it's intermittent.

whitslack:
FWIW, I saw this on Linux and had the same gut reaction as @rustyrussell. "Bad file descriptor" indicates a logic bug: passing to a syscall an integer that does not specify an open file descriptor.

cdecker:
I think we can close this, the root problem being somewhere else :-)

ddustin:
Wfm 👍

PR #4983, opened 2021-12-18
Add pip requirements install to Mac instructions

ddustin:
@cdecker

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -264,6 +264,7 @@ Configure Python 3.x & get mako:
 
 Build lightning:
 
+    $ pip install -r requirements.txt
     $ ./configure
     $ make
 


vincenzopalazzo (commented):
nit: We have this line in the previous section called `For development or running tests, get additional dependencies:`

Or I'm missing some things that you and @cdecker discussed?

cdecker:
I think `mrkd` (and `mistune` due to the broken dependency) might be enough, `requirements.txt` contains quite a few developer dependencies which are not needed for users (doesn't hurt though as long as there is no package conflict in the python installation).

ddustin:
It was just a step I missed when setting up a new environment from scratch
to test and running make check. Took me a while to figure it out ><. Maybe
the ideal solution is having an extra “contributing” or “developer” section
that comes later.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4983/commits/4c094e91127b5ded2635a221c045ce12293ab937

PR #4987, opened 2021-12-21
ci: introduce in ci workflow the compilation testing on alpine

vincenzopalazzo:
This PR tries to introduce inside the CI the compilation testing thought docker image, and this can be also the first step to create the official c-lightning image for each push on master, and this can be very cool to have also after introducing RPC in c-lightning with rust.

However, for the moment, this runs the docker image inside the ci and check if we can compile the actual version of code with a different OS not supported by Github action.

~In addition, having a docker image can help to avoid nondeterministic behavior when we have test failure because we can run exactly the same conf locally. Introduced also a docker-compose file to speed up the command line config with just a `docker-compose -f <name-file.yml> up`~

ping @jsarenik for alpine config
ping @cdecker  for docker configuration
ping @rustyrussell  for compilation failure, I was not able to catch this failure sorry, but I hope this PR can help you to reproduce in your local env. (Fixed https://github.com/ElementsProject/lightning/pull/5006)


cdecker:
> This PR tries to introduce inside the CI the compilation testing thought docker image, and this can be also the first step to create the official c-lightning image for each push on master, and this can be very cool to have also after introducing RPC in c-lightning with rust.

I'm afraid that with Docker Inc's recent monetization changes this is no longer as easy as it was. Downloading or uploading a docker image may now fail due to the lack of a subscription, so we'd first need to setup a repository to host the docker images first. A middleground could be doing nightly builds, where we at least have an upper bound on the number of images we produce.



> In addition, having a docker image can help to avoid nondeterministic behavior when we have test failure because we can run exactly the same conf locally. Introduced also a docker-compose file to speed up the command line config with just a `docker-compose -f <name-file.yml> up`

That would be a good option if we were running the tests inside a docker image, but this only starts the `lightningd` node as is (which immediately fails due to no bitcoind backend being available).

vincenzopalazzo:
>I'm afraid that with Docker Inc's recent monetization changes this is no longer as easy as it was. Downloading or uploading a docker image may now fail due to the lack of a subscription, so we'd first need to setup a repository to host the docker images first. A middleground could be doing nightly builds, where we at least have an upper bound on the number of images we produce.

Oh good to know this point, I didn't see the last updated.

>That would be a good option if we were running the tests inside a docker image, but this only starts the lightningd node as is (which immediately fails due to no bitcoind backend being available).

The actual PR, is only to reproduce the build, I can work to integrate also the testing in this way if you like the idea to have all with docker.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,34 @@
+FROM alpine:3.14.3
```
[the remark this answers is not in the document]
vincenzopalazzo: The idea that I would like to propose here is to introduce some docker images in the GitHub action to make the code reproducible in the local machine, and move these docker files used for internal stuff inside the `docker` directory to maintain order in case of this docker files will grow in numbers.

But, I can move it inside the con `contrib` direcotry

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,33 @@
...
+
+WORKDIR /build
+
+RUN apk update && \
+    apk add ca-certificates alpine-sdk autoconf automake git libtool \
+    gmp-dev sqlite-dev python3 py3-mako net-tools zlib-dev libsodium gettext su-exec \
+    python3 py3-pip && \
+    apk add --upgrade fortify-headers
+
+RUN mkdir lightning
```
cdecker, line 13: No need to first `mkdir` the copy below should suffice.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,33 @@
...
+COPY . lightning
+
+RUN cd lightning && \
+    # hack to address this issue https://github.com/ElementsProject/lightning/issues/4918
+    #rm -r cli/test/*.c && \
+    ./configure && git submodule update --init --recursive && \
```
cdecker, line 19: The submodules must be initialised before calling `./configure`.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,33 @@
...
+    pip3 install mrkd mistune==0.8.4 && \
+    make -j$(nproc) && \
+    make install
+
+RUN apk --update --allow-untrusted --repository http://dl-4.alpinelinux.org/alpine/edge/community/ add \
+      tor torsocks && \
+    apk --update --allow-untrusted --repository http://dl-4.alpinelinux.org/alpine/edge/testing/ add \
+      obfs4proxy
```
cdecker, line 27: What are these used for? I thought this was intended to be a builder for CI, not the image used in production by users.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,33 @@
...
+
+RUN echo $(lightningd --version)
+
+# TODO: review entry point here, to make this availale for the user
+
+CMD ["lightningd"]
```
cdecker, line 33: This entrypoint/cmd will not work for CI as it blocks indefinitely.

docker/Dockerfile.alpine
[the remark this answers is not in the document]
cdecker: Yeah, these aren't really top-level artefacts, rather we provide them for convenience. Especially if we plan to use these images primarily for CI testing, we should not expose them on the top-level.

.github/workflows/ci_build.yml
```diff
@@ -0,0 +1,18 @@
...
+      fail-fast: false
+      matrix:
+        include:
+          - { OS: alpine }
+    steps:
+      - uses: actions/checkout@v2
+      - name: Integration testing
+        run: |
+          docker build -f docker/Dockerfile.${{matrix.OS}} -t clightning-${{matrix.OS}} .
+          docker run clightning-${{matrix.OS}}
```
cdecker, line 18: With the entry point below, and missing the `bitcoin-cli` this'll just report an error about not being able to contact bitcoin and exit.

docker/Dockerfile.alpine
```diff
@@ -0,0 +1,33 @@
+FROM alpine:3.14.3
+LABEL maintainer="Vincenzo Palazzo (@vincenzopalazzo) vincenzopalazzodev@gmail.com"
```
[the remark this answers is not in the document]
vincenzopalazzo: ops, this maintainer field change again? :/

Thanks to point me out.

docker/Dockerfile.alpine
vincenzopalazzo: For now, I just moved to `["lightning", "--version"]`

docker/Dockerfile.alpine
vincenzopalazzo: Just a copy and paste error, for now, this is only used to test the compilation process, so we don't need this

docker/Dockerfile.alpine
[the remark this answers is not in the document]
vincenzopalazzo: I agree I moved the docker directory inside the contr directory, and I moved all the docker filed inside this directory.

Maybe this can be wrong or not needed? I was thinking that maybe having all the docker-related stuff inside a directory can help to keep the stuff organized.

What do you think?

.github/workflows/ci_build.yml
vincenzopalazzo: We should be able to compile without `bitcoin-cli` or better, we must be able to compile without bitcoin-cli because we don't directly depend on bitcoin core.

In addition, I would like to avoid installing a bitcoin core in the image, because if we add a tor dependence, we could ship a docker image and the user can use it with docker-compose.

vincenzopalazzo:
>I'm afraid that with Docker Inc's recent monetization changes this is no longer as easy as it was. Downloading or uploading a docker image may now fail due to the lack of a subscription, so we'd first need to setup a repository to host the docker images first. A middleground could be doing nightly builds, where we at least have an upper bound on the number of images we produce.

Thanks to reviewing it @cdecker, I made some research and I am not able to catch this part here, why we should now be able to push an image each tag with a process like the following one? https://docs.github.com/en/actions/publishing-packages/publishing-docker-images

docker/Dockerfile.alpine
vincenzopalazzo: Mh, the COPY command should not create a directory, or I-m missing some things?

vincenzopalazzo:
Somethings that can be related to the alpine error https://bugs.scummvm.org/ticket/12405

vincenzopalazzo:
trivial rebase on top https://github.com/ElementsProject/lightning/commit/35eaaf8e63449e737c8a340070bfbcd57bd12540

jsarenik (approved):
Looks good to me.

jsarenik:
Is the CI on this branch using the modified CI configuration files?

vincenzopalazzo:
Yes it is, this is the result https://github.com/ElementsProject/lightning/runs/5300375149?check_suite_focus=true

rustyrussell (approved):
Ack ae3b1c31acc0a0aa1220c4a1a1e350c7e11ccc89

PR #4989, opened 2021-12-22
pay: Disable channels that are not in normal state

cdecker:
Since we have the exact state of the channels from the `listpeers`
response we can filter out the ones that are not yet normal or not
anymore. This is more of a safety net, given that the `gossip_store`
should contain local disables, but better not be racy :-)

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/4989/commits/4218202ce7e4acda18ad0eb4122874185ecd44ea

niftynei (approved):
ACK 4218202ce7e4acda18ad0eb4122874185ecd44ea

PR #4993, opened 2021-12-30
hsmd: Add wallet index metadata to existing messages

ksedgwic:
Add wallet index and ext_key metadata to PSBT fields of transactions so validating signers can confirm change and shutdown addresses are controlled by the wallet.

This is needed for validating signers.

[Summary of all changes for validating signers](https://gitlab.com/lightning-signer/docs/-/wikis/C-Lightning-Modifications)

Changelog-None


ksedgwic:
rebased to fix conflicts w/ master

channeld/commit_tx.c
```diff
@@ -285,6 +285,7 @@ struct bitcoin_tx *commit_tx(const tal_t *ctx,
 	 *    `dust_limit_satoshis`, add a [`to_remote`
 	 *    output](#to_remote-output).
 	 */
+	u8 *redeem = NULL;
```
cdecker, line 288: We never set default values, during declaration, rather we let it uninitialized until we actually decide whether it should be a fixed value or a computed value. In this case it'd be down in the `if-else` statement where we either set it to `NULL` or the actual redeem script.

This helps us pin down accidental paths with `valgrind`.

common/initial_commit_tx.c
```diff
@@ -243,19 +243,19 @@ struct bitcoin_tx *initial_commit_tx(const tal_t *ctx,
 		 */
 		u8 *scriptpubkey;
 		int pos;
+		u8 *redeem = NULL;
```
cdecker, line 246: Same as above.

common/psbt_keypath.c
```diff
@@ -0,0 +1,33 @@
...
+#include <common/psbt_keypath.h>
+#include <common/utils.h>
+#include <wally_bip32.h>
+#include <wally_psbt.h>
+
+void psbt_set_keypath(u32 index, const struct ext_key *ext, struct wally_map *map_in) {
+	u8 fingerprint[BIP32_KEY_FINGERPRINT_LEN];
+	if (bip32_key_get_fingerprint(
+		    (struct ext_key *) ext, fingerprint, sizeof(fingerprint)) != WALLY_OK)
+		abort();
```
cdecker, line 12: Would it make sense to use `fatal("some sensible message")` here?

wallet/walletrpc.c
```diff
@@ -720,6 +750,9 @@ static struct command_result *json_signpsbt(struct command *cmd,
 		return command_fail(cmd, LIGHTNINGD,
 				    "No wallet inputs to sign");
 
+	// Update the keypaths on any outputs that are in our wallet (change addresses).
```
cdecker, line 753: Nit: `//` -> `/* ... */`

cdecker (approved):
Sorry for the (embarassing) delay for this PR. On the plus side I can now give more informed feedback since I started this kind of verification myself :-)

I think this is a perfect first step :+1: One thing to consider is if there is a way we can ensure that changes on our end don't end up breaking your side of things. Do you have a way to run the lightning-signer against the `master` branch and ping us if we accidentally break compatibility? I'd like to minimize work on your end as much as on our :-)

ACK 4604ee7437375dfda53743643c3ec21fb13ca65e

common/psbt_keypath.c
ksedgwic: `fatal` is `undefined reference` when this gets linked into `lightning_{channeld,closingd,onchaind}`.  I don't find any uses of `fatal` in the daemons, mostly `abort`.  I could ...
1. return wally status from these routines and let the callers `abort` or `fatal` depending on context
2. link something into the daemons to provide `fatal`
3. print something to `stderr` and `abort`
4 ...?
5. Profit!

Thoughts?

channeld/commit_tx.c
ksedgwic: fixed

common/initial_commit_tx.c
ksedgwic: fixed

wallet/walletrpc.c
ksedgwic: fixed

ksedgwic:
> One thing to consider is if there is a way we can ensure that changes on our end don't end up breaking your side of things. Do you have a way to run the lightning-signer against the master branch and ping us if we accidentally break compatibility? I'd like to minimize work on your end as much as on our :-)

Agreed, we are investigating CI for this purpose.

common/psbt_keypath.c
cdecker: That's quite alright, I shouldn't review code while on a plane :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4993/commits/81f02b5a8f0aee6f05ca43e892d1618aa303e461

PR #4995, opened 2022-01-01
dns: only allow one dns

m-schmoock:
Since https://github.com/lightning/bolts/pull/911 will contain `- MUST NOT announce more than one "type 5" DNS hostname.` we need to add this.


m-schmoock:
sorry, ignore this, will add correct PR later :)

Issue #911, opened 2018-02-04
"We want to know if htlc 1 is missing (immediate)"


fabianfabian:
lightingd keeps crashing, this is in the log:

```
2018-02-04T23:36:23.636Z lightningd(4284): Adding block 507509: 0000000000000000003b0c2cc3034ea8d6c71dd77d820df8b24cf6e668ac8892
**2018-02-04T23:36:24.989Z lightningd(4284): Owning output 1 300000 (P2SH) txid 20d8e529408a281f43284e216b182e23aaabbe64c7167b9b384065401ed2281a**
2018-02-04T23:36:25.186Z lightningd(4284): Adding block 507510: 0000000000000000005fdf68e6a860c11cad255157504bde4ca090ac910b59b7
2018-02-04T23:36:25.490Z lightningd(4284): Adding block 507511: 00000000000000000028307274c3176ffe9cfe418ac47f27e515b2b19c51d74e
2018-02-04T23:36:25.828Z lightningd(4284): Adding block 507512: 0000000000000000001406f51c9785fb2b07f879f8279da383eba4044c021ba3
**2018-02-04T23:36:27.153Z lightningd(4284): Owning output 0 135980 (SEGWIT) txid cf3b58d23d9773bdff19955add8d009b3ef9c02b895a2fa2c9b4d4459c855c81**
2018-02-04T23:36:27.295Z lightningd(4284): Adding block 507513: 00000000000000000044dac62e6b26827f7f717eda4dfdd32a740d9022b60af5
2018-02-04T23:36:28.811Z lightningd(4284): Adding block 507514: 0000000000000000003d8ccb167602660c5c2f6aca14344f7bc85ca21aa98e07
2018-02-04T23:36:30.307Z lightningd(4284): Adding block 507515: 0000000000000000004e25bb406c70ac566a73bd73af77b517e4cf7f53f2b365
2018-02-04T23:36:30.643Z lightningd(4284): Adding block 507516: 00000000000000000054cfd968c7b6d986f594b24a90c726c44db1c5f1e50c36
2018-02-04T23:36:30.941Z lightningd(4284): Adding block 507517: 0000000000000000005159fc290104ead48aebef3d49a53d9a05b4167cd8072f
2018-02-04T23:36:31.610Z lightningd(4284): Adding block 507518: 000000000000000000505633dbbc2050b349babf3fb22e79b54a4c635c96d24d
[...]

PR #4996, opened 2022-01-01
dns: only allow one DNS announcement

m-schmoock:
Since upcoming https://github.com/lightning/bolts/pull/911 will likely contain `- MUST NOT announce more than one "type 5" DNS hostname.` we need to add this.

Changelog-None

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -196,6 +196,23 @@ static char *opt_set_accept_extra_tlv_types(const char *arg,
 }
 #endif
 
+#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
+/* Returns the number of wireaddr types already announced */
+static size_t num_announced_types(enum wire_addr_type type, struct lightningd *ld)
+{
+	size_t num = 0;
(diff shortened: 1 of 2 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/4996/commits/6df2e9cf7834963965a6557235632b3dd9c45f91

m-schmoock:
@cdecker 
Regardless of the RFC that does not restrict the number of announced IP and TOR addresses, should we (re)add such a restriction that limits it down to i.e. 3 IP addresses and 3 TOR address?

@TheBlueMatt and others mentioning that the network should really limit on gossip content, but currently nobody enforces anything so the limits were removed from the spec. So as a first step we can add a rule that at least our implementation can't be used to spam gossip by commandline.

m-schmoock:
Also, currenlty RFC #911 does forsee to ignore and drop node_updates that contain more than one DNS name.

m-schmoock:
Update: Rust-Lightning seems picking up on this :D
 https://github.com/lightningdevkit/rust-lightning/pull/1329

PR #4998, opened 2022-01-04
LightningRpc should accept pathlib Path

manreo:
Not sure why socket does not accept pathlib, but just added some code to transform the path to string.
It is not the most pretty code, but better than these:
https://stackoverflow.com/questions/58647584/how-to-test-if-object-is-a-pathlib-path

### DIFF
--- a/contrib/pyln-client/pyln/client/lightning.py
+++ b/contrib/pyln-client/pyln/client/lightning.py
@@ -223,7 +223,7 @@ class UnixSocket(object):
     """
 
     def __init__(self, path: str):
-        self.path = path
+        self.path = str(path) if 'pathlib' in str(type(path)) else path
         self.sock: Optional[socket.SocketType] = None
         self.connect()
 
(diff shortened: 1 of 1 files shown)


contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -223,7 +223,7 @@ class UnixSocket(object):
     """
 
     def __init__(self, path: str):
-        self.path = path
+        self.path = str(path) if 'pathlib' in str(type(path)) else path
```
cdecker, line 226: This can simply be this:

```suggestion
        self.path = str(path)
```

This is because `str("hellworld") == "helloworld"`. On the other hand it might be sensible to keep `self.path` a `pathlib.Path` as long as possible, since it makes operating on the object much easier.

cdecker (changes requested):
Very good find, though it can be simplified :-)

manreo:
Thanks, not sure what my intuition was and why not just to use str... 
I also moved it next to the socket

Issue #5002, opened 2022-01-12
Build issues on macOS Big Sur: `INSTALL.md` seems out of date


msgilligan:
I had a handful of issues trying to build on **macOS** 11.6.2 (Big Sur) and **Xcode** 13.2.1.

I didn't do a great job of recording the errors (because I'm new to this project and was learning as I went along, sorry), but I am going to make a PR with my successful mods to `INSTALL.md`

msgilligan:
See PR #5003

rustyrussell:
AI cleanup of old issues.  If wrong, please open a new issue.


Triage decision summary:
macOS Big Sur build documentation issue from January 2022. User reports outdated INSTALL.md and mentions planned PR. Issue is over 2 years old with minimal engagement and likely addressed by subsequent documentation updates.

PR #5003, opened 2022-01-12
Update build instructions (INSTALL.md) for macOS

msgilligan:
These changes are what I needed to get it working on macOS 11.6.2 with
Xcode 13.2.1.

See Issue #5002

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -220,7 +220,7 @@ To Build on macOS
 Assuming you have Xcode and Homebrew installed. Install dependencies:
 
     $ brew install autoconf automake libtool python3 gmp gnu-sed gettext libsodium
-    $ ln -s /usr/local/Cellar/gettext/0.20.1/bin/xgettext /usr/local/opt
+    $ ln -s /usr/local/Cellar/gettext/0.21/bin/xgettext /usr/local/opt
     $ export PATH="/usr/local/opt:$PATH"
 
 If you need SQLite (or get a SQLite mismatch build error):
@@ -239,7 +239,7 @@ If you need Python 3.x for mako (or get a mako build error):
     $ brew install pyenv
     $ echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
     $ source ~/.bash_profile
-    $ pyenv install 3.7.4
+    $ pyenv install 3.7.12
     $ pip install --upgrade pip
 
 If you don't have bitcoind installed locally you'll need to install that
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,4 +1,6 @@
 # Dependencies required to build and test c-lightning
+
+mistune ~= 0.8.4
```
vincenzopalazzo, line 3: This is an indirect dependence, you can solve your issue with the following command `pip3 install -r requirements.lock`

This is a leak of the mrkd https://github.com/refi64/mrkd/issues/7 

For now, we are using the option to install mistune by hand with the command `pip3 install mistune == 0.8.4`

vincenzopalazzo (changes requested):
I have only one comment in the `requirements.txt` and for the OSX conf looks good to me, In addition, its looks like an OSX config and I'm a little bit worried if for all OSX with the same version this configuration is true.

msgilligan:
Thanks @vincenzopalazzo -- I probably should have made this a "draft" PR. I'll make the changes and resubmit.

requirements.txt
msgilligan: Fixed in f238cdcbcaf426af15d25678c3fa5f0c7e2c6068

msgilligan:
@vincenzopalazzo said in a line-comment:

> This is an indirect dependence, you can solve your issue with the following command pip3 install -r requirements.lock

Should we add this to the instructions in `INSTALL.md`? (I'm only an occasional Python user, so I don't really know what this means...)

doc/INSTALL.md
```diff
@@ -259,12 +259,12 @@ Clone lightning:
 
 Configure Python 3.x & get mako:
 
-    $ pyenv local 3.7.4
+    $ pyenv local 3.7.12
     $ pip install mako
```
vincenzopalazzo, line 263: maybe we can add here the mistune dependencies
```suggestion
    $ pip install mako mistune==0.8,4
```

Good suggestion, thanks !

vincenzopalazzo (commented):
Yes, for now, we choose that mistune stay in the install description, and hope that we have the new version soon.

doc/INSTALL.md
msgilligan: Did you mean for that to be `0.8,4` rather than `0.8.4`?

vincenzopalazzo: ops! my bad, yes the syntax correct is with dots, like the following one.
```suggestion
    $ pip install mako mistune==0.8.4
```

In the doc for Unix system, we have already the same type of command, you can also double-check in the doc.

Sorry for my mistake

cdecker: Let's use the requirements.lock file rather than specifying individual dependencies with versions. Otherwise we'll have to touch these docs every time a dependency changes.

vincenzopalazzo: Good idea!

doc/INSTALL.md
```diff
@@ -259,12 +259,12 @@ Clone lightning:
...
 
 Build lightning:
 
-    $ pip install -r requirements.txt
+    $ pip3 install -r requirements.txt
```
cdecker, line 267:
```suggestion
    $ pip3 install -r requirements.lock
```

cdecker (changes requested):

Issue #4918, opened 2021-11-17
compilation error on alpine:3.13.6 in a docker container


vincenzopalazzo:
I'm running the official version of c-lightning v0.10.2 and compiling it on alpine, but I receive the following compilation error

```
cc cli/test/run-large-input.c
In file included from /usr/include/fortify/strings.h:22,
                 from /usr/include/string.h:57,
                 from /usr/include/fortify/string.h:22,
                 from ccan/ccan/str/str.h:5,
                 from ccan/ccan/tal/tal.h:8,
                 from ./common/amount.h:5,
                 from cli/test/run-human-mode.c:2:
/usr/include/fortify/unistd.h:110:1: error: 'read' undeclared here (not in a function); did you mean 'pread'?
  110 | _FORTIFY_FN(read) ssize_t read(int __f, void *__s, size_t __n)
      | ^~~~~~~~~~~
In file included from cli/test/../lightning-cli.c:22,
                 from cli/test/run-human-mode.c:27:
/usr/include/fortify/unistd.h: In function 'read':
/usr/include/fortify/unistd.h:116:9: error: called object '__orig_read' is not a function or function pointer
  116 |  return __orig_read(__f, __s, __n);
      |         ^~~~~~~~~~~
In file included from /usr/include/fortify/strings.h:22,
                 from /usr/include/string.h:57,
                 from /usr/include/fortify/string.h:22,
                 from ccan/ccan/str/str.h:5,
                 from ccan/ccan/tal/tal.h:8,
                 from ./common/amount.h:5,
                 from cli/test/run-human-mode.c:2:
/usr/include/fortify/unistd.h:110:1: note: declared here
  110 | _FORTIFY_FN(read) ssize_t read(int __f, void *__s, size_t __n)
      | ^~~~~~~~~~~
make: *** [Makefile:279: cli/test/run-human-mode.o] Error 1
make: *** Waiting for unfinished jobs....
In file included from /usr/include/fortify/strings.h:22,
                 from /usr/include/string.h:57,
                 from /usr/include/fortify/string.h:22,
                 from ccan/ccan/str/str.h:5,
                 from ccan/ccan/tal/tal.h:8,
                 from ./common/amount.h:5,
                 from cli/test/run-large-input.c:2:
/usr/include/fortify/unistd.h:110:1: error: 'read' undeclared here (not in a function); did you mean 'pread'?
  110 | _FORTIFY_FN(read) ssize_t read(int __f, void *__s, size_t __n)
[...]

PR #5006, opened 2022-01-12
cli: change the redefine logic to read from the stdin

vincenzopalazzo:
Procedure redefinition is a very cool feature but when other libraries redefine the same function can be very tricky to avoid compilation errors.

This PR proposed a change of logic and use a customization function definition to read from the stdin, so we can avoid future errors at compile time.

However, we could be check also the os or compiler that cause the error and redefine the missing function.

Compilation success on alpine is reported here https://github.com/vincenzopalazzo/lightning/runs/4795110552?check_suite_focus=true

If this PR will be merged the #4987 should be compiled on alpine to each new changes

Fixes https://github.com/ElementsProject/lightning/issues/4918 (really fixes)

### DIFF
--- a/cli/config_cli.h
+++ b/cli/config_cli.h
@@ -1,0 +1,15 @@
+#ifndef LIGHTNING_CLI_CONFIG_CLI_H
+#define LIGHTNING_CLI_CONFIG_CLI_H
+
+#include "config.h"
+#include <stdio.h>
+
+#ifndef CLN_TEST
+/* Redefinition procedure is a very cool feature, but
+   if we try to redefine a procedure that is already
+   redefined somewhere (like read in alpine) we can have
+   tricky compilation error */
+#define cli_read read
+#endif
+
+#endif /* LIGHTNING_CLI_CONFIG_CLI_H */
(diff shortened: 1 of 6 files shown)


cli/config_cli.h
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+
+#ifndef CLN_TEST
+// Redefinition procedure it is a very cool feature
+// However, if we try to redefine a feature that is already
+// defined like read in alpine we can have triky compilation error
```
cdecker, line 10: Nit: We use `/* ... */` comments everywhere.

cdecker:
This PR is not bisectable, so I'll squash it if that's ok with you :-)

cli/config_cli.h
vincenzopalazzo: Ops, we are not in rust :)

vincenzopalazzo:
>This PR is not bisectable, so I'll squash it if that's ok with you :-)

Sure, and sorry if I missed it. I squash the commits and add the changelog.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5006/commits/795eee1be2679b5ff88efc2f78db64c6e5836c84

PR #5010, opened 2022-01-18
cln-rpc: Oxidizing c-lightning

cdecker:
`cln-rpc` is the first step towards adding rust as a language to c-lightning. It consists of the following steps:

 - Add `msggen` to generate a variety of messages from the JSON-RPC schemas, and maybe soon from the wire CSV files.
 - Generate structs for the rust side that match requests and responses for the JSON-RPC. These just serve as containers for the data sent to and returned by the JSON-RPC. They are as strongly typed as the schemas would allow at the moment, but further constraints and types could be introduced later.
 - Conditionally compile rust targets when we detect `RUST=1` (whether `cargo` is in the path or not)
 - Implement the codec for encoding and decoding of the JSON-RPC to and from the Rust structs

`msggen` is currently based off of the JSON-RPC schemas, but the schemas are lacking some of the higher level concepts such as methods (binding request and response together) or services (binding multiple methods into a grouping). We could eventually create a new root document and make the JSON-RPC schemas derived from that. This would have the advantage of reducing the very verbose JSON-schemas, and being a superset of the information we need for various formats (grpc field identities for example which must not change despite schema evolutions). However I decided against bikeshedding that format for now, and will revisit it in the future.

We are still missing quite a few request schemas, which is why not all methods are mapped at the moment. It's gruntwork that I'll add in a separate PR.


cln-rpc/src/primitives.rs
```diff
@@ -0,0 +1,132 @@
...
+    ONCHAIN,
+}
+
+#[derive(Copy, Clone, Serialize, Debug, PartialEq)]
+pub enum Amount {
+    Millisatoshi(u64),
+    Satoshi(u64),
+    Millibitcoin(u64),
+    Bitcoin(u64),
+}
```
elsirion, line 37: That type seems cursed, why not just go with some msat-based amount type and convert appropriately? Is re-encoding with syntactical (instead of just semantic) equality really a goal?

cdecker: We've had quite some cases in the past where users of the API misunderstood the unit they were dealing with, and ended up creating channels that were 1000x the size they intended, or transferred 0.001x the amount they were supposed to. Hence this more complex, but also much more strict and user visible type.

elsirion: Do you think it's that much better than e.g. `struct Amount { pub msat: u64 }` with `from_btc(btc: u64)`, `from_sats(sats: u64)` and `from_milli_sats(msats: u64)` as constructors?

Its one thing to misjudge units in the CLI when not supplying units at all, it's a completely different thing working with an API that forces you to specify your intent through either function names or variants, imo both work pretty much the same, this will just be somewhat unergonomic imo, imagine maybe implementing arithmetic on this type :grimacing:

At least there should be a `to_msat(&self) -> u64` function so users don't need to match every time.

cdecker: The main reason we have this rather complex type is because we want to have `Amount::All` and `Amount::Any` variants as well, which we can't map to a `pub struct Amount { millisatoshi: u64, }`

cdecker:
Ok, this should be ready to start reviewing, the discussion started with @elsirion is definitely one point I'd like to dig in. The `Amount` type needs to also get `All` and `Any` variants, which have no nice mapping to an msatoshi amount.

cln-rpc/src/primitives.rs
elsirion: Maybe it would still make sense to do

```rust
enum Amount {
    MilliSat(u64),
    All,
    Any
}
```

Btw, does `All` even exist on the LN side? I only know it from the on-chain wallet. If so, maybe even split into two:

```rust
enum BtcAmount {
    Amount(bitcoin::Amount), // only supports sat granularity
    All
}

enum LnAmount {
    MilliSat(u64),
    Any
}
```

Otherwise there will be weird error conditions depending on context.

cln-rpc/src/model.rs
```diff
@@ -0,0 +1,338 @@
...
+	    pub version: String,
+	    #[serde(alias = "lightning-dir")]
+	    pub lightning_dir: String,
+	    #[serde(alias = "blockheight")]
+	    pub blockheight: u32,
+	    #[serde(alias = "network")]
+	    pub network: String,
+	    #[serde(alias = "fees_collected_msat")]
+	    pub fees_collected_msat: Amount,
+	    #[serde(alias = "address[]")]
```
vincenzopalazzo, line 189:
```suggestion
	    #[serde(alias = "address")]
```

The alias should be the name of the field, so the name should not contain [] ? or I'm missing something?

cln-rpc/src/primitives.rs
vincenzopalazzo: Do you think that dividing the enum in ln and btc makes the API harder to read? I mean, there is any ln amount, just a milli convention. 

I like the @cdecker idea, to have one single wrapper around the amount value, at the end of the day we want that the user can use the library with any knowledge of btc.

contrib/msggen/msggen/rust.py
```diff
@@ -0,0 +1,242 @@
...
+//! this file was generated from
+
+"""
+
+
+def normalize_varname(field):
+    """Make sure that the variable name of this field is valid.
+    """
+    # Dashes are not valid names
+    field.path = field.path.replace("-", "_")
```
vincenzopalazzo, line 58:
```suggestion
    field.path = field.path.replace("-", "_").replace("[]", "")
```

Maybe this is the cause of the serde alias name?

vincenzopalazzo (commented):
LGTM, but I have a small comment in some `autogen` code.

In addition, I have also I'm curious about the main motivation that you choose to use `tokio` as dependencies, I mean way, not just a synchronous library?

cln-rpc/src/primitives.rs
elsirion: @vincenzopalazzo I think encoding invariants in the types makes the API easier to use. If you were to supply an msat granular amount to an operation expecting sat granularity that would need to be an error or panic (both annoying imo). Same goes for the possible `all` and `any` variants.

@cdecker reading the `lightning-cli` docs I noticed that other parameters also take `all` (e.g. the id when setting channel fees). So another idea would be to define the amount type as msat without regard to `all`/`any`, but also defining a `MaybeAll` and `MaybeAny` enum with a custom `serde` impl.

```rust
enum MaybeAll<T> {
    Concrete(T),
    All
}
```

cdecker: I very much like the idea of using `MaybeAll<T>` and `MaybeAny<T>`, however currently it is not possible for us to use these, since this is information is currently not conveyed in the JSON schema at all (flat `msat` type for all fields dealing with amounts).

To fix this we'd have to go through the RPC calls and annotate them with `msat-any` or `msat-all` if one of these variants is to be used. So before going that way it's good to be sure we want that option.

cdecker: After looking a bit harder at the JSON schema, I think we can annotate the very few cases where we'd have `msat|all` and `msat|any`. My count shows that only 3 RPC methods accept `any` and 5 accept `all`, and since I need to write the request schemata anyway, that might be a good time to introduce these. Trying the changes in a separate branch and they look rather nice.

The only thing that is less nice is that we don't remember the unit that the user chose to express in, flattening everything to msats, but that's a presentation issue anyway.

elsirion: > The only thing that is less nice is that we don't remember the unit that the user chose to express in, flattening everything to msats, but that's a presentation issue anyway.

Imo representation should be left to the application (there's probably an application-wide standard anyway, so no need to preserve units used). On an API level having only msat makes a lot of sense because it reduces complexity and computers don't care. I assume having support for different units in the first place was to make life easier for `lightning-cli` users. Thinking of it, was any thought given to impl the unit conversion in `lightning-cli` instead of `lightningd` (or whatever subdaemon it is)?

cdecker: Agreed, I think the `struct Amount { msat: u64 }` is likely far better than what we have so far. I'll test that one and see how much fallout it produces.

cdecker:
Thanks for the review @vincenzopalazzo. I chose `tonic` primarily to minimize the memory footprint, since threads each get their own stack preallocated, and may have considerable overhead. Since we are likely much more IO bound than CPU bound, it seems to me like a good fit. Also it's what we're using in greenlight which is where I'm backporting this from 😉

cdecker:
Following the discussion with @elsirion I went and simplified the `Amount` type and introduced two new types `AmountOrAll` and `AmountOrAny` (I didn't go for a generic type since I couldn't see how it'd be useful for other types at all).

I don't think we need to differentiate off-chain and on-chain amounts for now, but they'd be easy to introduce later on.

The range-diff can be found here: https://bot.bitcoinstats.com/ElementsProject/lightning/pull/5010/range_diff/7e456784ed2fac1871340e60e211f331b461eae6..37fd5c7b9a11fc6e5f0758520ee8feed34b3d8b4

```
 1:  fec9f68cd =  1:  f72fe4615 pyln: Delete psql DBs after testing
 2:  0f25011c3 =  2:  66a40cb8d --- START OF cln-rpc
 3:  cca9b8082 =  3:  636b47996 msggen: Parse JSON-RPC schemas and build the in-memory model
 4:  ba9ffb0a5 =  4:  46a8c3730 msggen: Generate the cln-rpc Rust structs
 5:  466395d3f =  5:  affe8f652 json-rpc: Add request stubs for a couple of calls
 6:  eb7a5d702 =  6:  e25eec08c rust: Add rust detection to configure and a target to add binaries
 7:  afb9d3bf9 =  7:  c06234da0 cln-rpc: Scaffolding for the cln-rpc crate
 -:  --------- >  8:  3bfb36608 fixup! cln-rpc: Scaffolding for the cln-rpc crate
 8:  6b2acc85a =  9:  8ebb0a9b9 pytest: Test the rust bindings from cln-rpc
 9:  246ddd946 = 10:  f88e4cf10 gci: Add rust configuration to Github actions
10:  7e456784e = 11:  42a4a6ddf gci: Limit the RUST=1 config to test rust-related functionality
 -:  --------- > 12:  c7b805038 fixup! cln-rpc: Scaffolding for the cln-rpc crate
 -:  --------- > 13:  c27ba39c6 fixup! 3bfb36608c5159a7ad5f1afeb5c1e0a1ec955fff
 -:  --------- > 14:  7c752aceb fixup! cln-rpc: Scaffolding for the cln-rpc crate
 -:  --------- > 15:  37fd5c7b9 cln-rpc: Add type for AmountOrAll and AmountOrAny
```

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/commit/4c12c507a8811a24e3daed9280cbeb71b168ce3a

Maybe add the `Changelog-Add: ....` can be a good information to have inside the changelog?

configure
```diff
@@ -109,6 +109,15 @@ default_valgrind_setting()
     fi
 }
 
+default_rust_setting()
+{
+    if cargo --version > /dev/null 2>&1 || [ $? = 0 ]; then
```
rustyrussell, line 114: I don't understand `|| [ $? = 0 ]` here?  That seems redundant....

rustyrussell (approved):
Ack 3003a01a0c97ba0af8ae9c5f0109608430232d52

PR #5012, opened 2022-01-18
cln-plugin: Write your plugins in rust

cdecker:
This is likely the least mature crate of the cln-* series. It just implements the bare minimum of a plugin's lifetime such that it stays alive when run by `lightningd`. It doesn't yet have custom RPC methods support, nor config support, but I intend to build these out as we go. It is pretty much just the minimum to get us into a situation we can write the `grpc-plugin` plugin.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,4 +2,5 @@
 members = [
   "cln-rpc",
   "cln-grpc",
+  "plugins",
 ]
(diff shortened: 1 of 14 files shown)


sr-gi:
👀

sr-gi:
👀

rustyrussell (approved):
Ack a6e9b6b32500394d9648dde64e21bd40cb2c7214

PR #5013, opened 2022-01-18
cln-grpc-plugin: Secure access to your node over the network

cdecker:
This is the final PR in the cln-* series. It uses all the primitives we built in the previous 3 PRs and uses them to expose the JSON-RPC over grpc, with mTLS authentication builtin.

### DIFF
--- a/plugins/grpc-plugin/Cargo.toml
+++ b/plugins/grpc-plugin/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+edition = "2021"
+name = "grpc-plugin"
+version = "0.1.0"
+
+[dependencies]
+anyhow = "1.0"
+log = "0.4"
+prost = "0.8"
+rcgen = { version = "0.8", features = ["pem", "x509-parser"] }
+
+[dependencies.cln-grpc]
+path = "../../cln-grpc"
+
+[dependencies.cln-plugin]
+path = "../../plugins"
+
+[dependencies.cln-rpc]
+path = "../../cln-rpc"
+
+[dependencies.tokio]
+features = ["net", "rt-multi-thread"]
+version = "1"
+
+[dependencies.tonic]
+features = ["tls", "transport"]
+version = "^0.5"
--- a/plugins/grpc-plugin/Cargo.toml
+++ b/plugins/grpc-plugin/Cargo.toml
@@ -1,7 +1,11 @@
 [package]
 edition = "2021"
-name = "grpc-plugin"
+name = "cln-grpc-plugin"
 version = "0.1.0"
+
+[[bin]]
+name = "cln-grpc"
+path = "src/main.rs"
 
 [dependencies]
 anyhow = "1.0"
--- a/plugins/grpc-plugin/src/main.rs
+++ b/plugins/grpc-plugin/src/main.rs
@@ -5,10 +5,14 @@ use log::{debug, warn};
 use std::net::SocketAddr;
 use std::path::{Path, PathBuf};
 
+mod tls;
+
 #[derive(Clone, Debug)]
 struct PluginState {
     rpc_path: PathBuf,
     bind_address: SocketAddr,
+    identity: tls::Identity,
+    ca_cert: Vec<u8>,
 }
 
 #[tokio::main]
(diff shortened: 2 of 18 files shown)


plugins/grpc-plugin/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+edition = "2021"
+name = "grpc-plugin"
+version = "0.1.0"
```
vincenzopalazzo, line 4: maybe it is missing a space here between the two sections?

vincenzopalazzo (commented):
I was searching about the plugin architecture, to think about how to implement the `grpc` interface for my dart wrapper. In the meanwhile I left one comment (very very stupid).

Half of the code is the `tls` config, and I'm not the right person to review it

plugins/grpc-plugin/Cargo.toml
cdecker: Fixed up in 56d1820c9

rustyrussell:
Should this get a Changelog-Added entry?  Should it be in the default PLUGINS list for builtin plugins if rust is supported?  (Otherwise make install won't install it).

cdecker:
> Should this get a Changelog-Added entry? Should it be in the default PLUGINS list for builtin plugins if rust is supported? (Otherwise make install won't install it).

I'm tempted to mark this as experimental for this release, since I was unable to extensively test the conversion code from JSON-RPC to grpc and back. Shall I wrap the respective `Makefile` entries with experimental markers? Hence the lack of Changelog entries atm.

rustyrussell:
I think experimental is too harsh: if it requires a config option to activate, that's enough. IMHO.

Plugins generally don't have the same requirements as core functionality IMHO.

cdecker:
@rustyrussell I have your branch merged in and it compiles. Will spend today adding more types since they can be difficult to add later, and add some tests to verify the generated conversions are working ok :+1:

plugins/grpc-plugin/src/main.rs
```diff
@@ -31,13 +31,17 @@ async fn main() -> Result<()> {
...
-            options::Value::Integer(50051),
+            options::Value::Integer(-1),
             "Which port should the grpc plugin listen for incoming connections?",
         ))
         .start()
         .await?;
 
     let bind_port = match plugin.option("grpc-port") {
+        Some(options::Value::Integer(-1)) => {
+            log::warn!("`grpc-port` option is not configured, exiting.");
```
rustyrussell, line 42: This is probably just log::info?  It's a great plugin, but every non-activated plugin should probably not shout to be activated :)

rustyrussell (changes requested):
Trivial change requested... I would merge anyway, but I know more is coming...

cdecker:
Fixed up. The remaining grpc methods are in a separate PR, so feel free to merge this one :-)

rustyrussell (approved):

rustyrussell:
Pushed a simple test race fix.

Ack 343961d1c430eb5072264276ef096d109f5091a1

plugins/grpc-plugin/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+edition = "2021"
+name = "cln-grpc-plugin"
+version = "0.1.0"
```
vincenzopalazzo, line 4: if it is incompleted (request and response missed) maybe we can use the convention like this one?
```suggestion
version = "0.0.1"
```

And release the `0.1.0` only after all req/resp are supported?

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5013/commits/343961d1c430eb5072264276ef096d109f5091a1

(with just some small comment!)

plugins/grpc-plugin/Cargo.toml
cdecker: Good idea :+1:

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5013/commits/a47770dd4e7a8ea65d0e915fba25dc8731160797

PR #5014, opened 2022-01-18
json_add_invoice: fix crash if missing invstring

jb55:
If this field is missing for whatever reason (weird db state?)
clightning will crash when listing invoices.

This started happening when I pulled the latest master. I haven't bisected to see where it was introduced yet, if I even can anymore due to db upgrade.

Changelog-Fixed: jsonrpc: Fixed a crash caused by a missing invoice string in `listinvoices`

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -39,7 +39,8 @@ static void json_add_invoice(struct json_stream *response,
 			     const struct invoice_details *inv)
 {
 	json_add_escaped_string(response, "label", inv->label);
-	json_add_invstring(response, inv->invstring);
+	if (inv->invstring)
+		json_add_invstring(response, inv->invstring);
 	json_add_sha256(response, "payment_hash", &inv->rhash);
 	if (inv->msat)
 		json_add_amount_msat_compat(response, *inv->msat,


m-schmoock:
concept ack, but we should add an `@pytest.mark.xfail` testcase as first commit and remove that annotation along with this fix.
I can run all the `tests/test_invoices.py` and `tests/test_pay.py` just fine on master, which means we have a missing test coverage or something fishy happened on your node that maybe needs more investigation.

Any idea how to reproduce this?

m-schmoock:
Do you have a stacktrace to see how its called?

jb55:
On Wed, Jan 19, 2022 at 01:14:19AM -0800, Michael Schmoock wrote:
>Do you have a stacktrace to see how its called?

I did but unfortunately I didn't save it and it got last in the
backbuffer. all I remember is that there was a BROKEN database message
and that listinvoices was in the backtrace.

vincenzopalazzo:
>This started happening when I pulled the latest master. I haven't bisected to see where it was introduced yet, if I even can anymore due to db upgrade.

Maybe due to the db cleaning? but this is only possible if you are running the old version of c-lightning, not the lates.

In addition, I'm worried about the possibility to hide a bug somewhere with this check.

rustyrussell:
You should have a BROKEN log message about "Accessing a null column" as well, in your logs?

jb55:
On Wed, Jan 19, 2022 at 09:08:20PM -0800, Rusty Russell wrote:
>You should have a BROKEN log message about "Accessing a null column" as well, in your logs?

yes it was along those lines, my logs always spewed errors about this before but now I'm hitting a crash. maybe I should just drop those rows instead...

rustyrussell:
> On Wed, Jan 19, 2022 at 09:08:20PM -0800, Rusty Russell wrote: You should have a BROKEN log message about "Accessing a null column" as well, in your logs?
> yes it was along those lines, my logs always spewed errors about this before but now I'm hitting a crash. maybe I should just drop those rows instead...

No, I will figure out what release and put the check in an #if COMPAT wrapper...

niftynei:
I found a similar db-null error in the inflights while pulling out the database code. Are we missing a test/check for database errors? https://github.com/ElementsProject/lightning/pull/5016/commits/79d3ec20e5a23990d307b3156a268df2b261dfc2

rustyrussell:
Hmm, this is super weird.  This code would crash on a null invstring since 0.10.0; looking at 0.9.0 it would have printed "(null)".

let me look and see if there's a believable way we could have *created* a *new* NULL-invstring entry.  `sqlite3 ~/.lightning/bitcoin/lightningd.sqlite3 'select * from invoices where bolt11 is null'` might help give us a clue, too?

jb55:
oh yeah these are pretty ancient:

```csv
     1	id  state  msatoshi  label           expiry_time       pay_index  msatoshi_received  paid_timestamp    bolt11  description  hex(features)
     2	1   1      10000     10 sats please  2018-01-17 11:05  1          10000              2018-01-20 20:25                       
     3	2   1      1         1msat           2018-01-17 11:52  2          1                  2018-01-20 20:25                       
     4	3   2      100000    1bit            2018-01-17 14:19                                                                       
     5	4   2      10000     rompsats        2018-01-17 18:15
```

lightningd/invoice.c
```diff
@@ -39,7 +39,8 @@ static void json_add_invoice(struct json_stream *response,
 			     const struct invoice_details *inv)
 {
 	json_add_escaped_string(response, "label", inv->label);
-	json_add_invstring(response, inv->invstring);
+	if (inv->invstring)
+		json_add_invstring(response, inv->invstring);
```
vincenzopalazzo, line 43:
```suggestion
	if (inv->invstring) {
	        log_unusual("invstring null this should be cause from a very old invoice (back in 2018 maybe?)")
		json_add_invstring(response, inv->invstring);
         }
```

Not sure that the `log_unusual` syntax is correct, but you got may point

vincenzopalazzo (approved):
Wow, the invoice is a piece of history :-D that it is cool.

In addition, from my point of view, it is good to have this sanity check, also because we can have other people with very old db that are not using the node for a long time.

I agree with Rusty that a log message in this case it is good

P.S: the commit message is missing the changelog, maybe we need to put `Changelog-Fixed: jsonrpc: Fixed a crash caused by a missing invoice string in listinvoices` in the commit message? and add more precise information about the cause of the crash.

rustyrussell:
Ack 665b206432c0e9a7a68fd117b24b1e7b9a46142f

I am tempted to add @vincenzopalazzo 's log msg, but OTOH it'd simply spam jb55's logs, and nobody elses :)

PR #5017, opened 2022-01-21
Database librarization

niftynei:
Wouldn't it be great if we could reuse the database code for other things?

This pulls all the database related stuff up into a separate 'package', sort of.

Note that `db_fatal` now prints directly to stderr which may not be ideal in all cases. Open to suggestions for how to make this less janky.

Changelog-None


lightningd/Makefile
```diff
@@ -132,9 +132,9 @@ LIGHTNINGD_COMMON_OBJS :=			\
 include wallet/Makefile
 
 # All together in one convenient var
-LIGHTNINGD_HEADERS = $(LIGHTNINGD_HEADERS_NOGEN) $(LIGHTNINGD_HEADERS_GEN) $(WALLET_LIB_HEADERS)
+LIGHTNINGD_HEADERS = $(LIGHTNINGD_HEADERS_NOGEN) $(LIGHTNINGD_HEADERS_GEN) $(WALLET_HDRS)
 
-$(LIGHTNINGD_OBJS): $(LIGHTNINGD_HEADERS) $(WALLET_HDRS)
+$(LIGHTNINGD_OBJS): $(LIGHTNINGD_HEADERS)
```
rustyrussell, line 139: Yech, this is a bigger mess.  LIGHTNINGD_HEADERS is defined twice in this Makefile.

We should include wallet/Makefile earlier, then append the WALLET_HDRS to LIGHTNINGD_HDRS.  (We should also have `$(WALLET_OBJS): $(LIGHTNINGD_HEADERS)` after that).

common/utils.c
```diff
@@ -223,3 +224,15 @@ int tmpdir_mkstemp(const tal_t *ctx, const char *template TAKES, char **created)
 
 	return fd;
 }
+
+void err_1(const char *fmt, ...)
+{
```
rustyrussell, line 229: This is called errx, it already exists in ccan/err, and I don't see where you call it?

rustyrussell (commented):
Unfinished?

common/utils.c
niftynei: It's used in the previous patch.. will remove this one and update that one to use `errx`

lightningd/Makefile
niftynei: Ok, updated to fix these other issues :)

db/bindings.h
```diff
@@ -0,0 +1,118 @@
...
+/* Bridge function to get column number from SELECT
+   (must exist) */
+size_t db_query_colnum(const struct db_stmt *stmt, const char *colname);
+
+u64 db_col_u64(struct db_stmt *stmt, const char *colname);
+size_t db_col_bytes(struct db_stmt *stmt, const char *colname);
+const void* db_col_blob(struct db_stmt *stmt, const char *colname);
+char *db_col_strdup(const tal_t *ctx,
+		    struct db_stmt *stmt,
+		    const char *colname);
```
vincenzopalazzo, line 69:
```suggestion
char *db_col_strdup(const tal_t *ctx, struct db_stmt *stmt,
		    const char *colname);
```

Just for consistency

db/bindings.h
```diff
@@ -0,0 +1,118 @@
...
+void db_col_secret(struct db_stmt *stmt, const char *colname, struct secret *s);
+struct secret *db_col_secret_arr(const tal_t *ctx, struct db_stmt *stmt,
+				 const char *colname);
+void db_col_txid(struct db_stmt *stmt, const char *colname, struct bitcoin_txid *t);
+void db_col_channel_id(struct db_stmt *stmt, const char *colname, struct channel_id *dest);
+void db_col_node_id(struct db_stmt *stmt, const char *colname, struct node_id *ni);
+struct node_id *db_col_node_id_arr(const tal_t *ctx, struct db_stmt *stmt,
+				   const char *colname);
+void db_col_pubkey(struct db_stmt *stmt, const char *colname,
+		   struct pubkey *p);
```
vincenzopalazzo, line 85:
```suggestion
void db_col_pubkey(struct db_stmt *stmt, const char *colname,
		       struct pubkey *p);
```

With Github UI is a little bit difficult, but I think here it is missing a tab

db/db_postgres.c
```diff
@@ -93,7 +92,7 @@ static PGresult *db_postgres_do_exec(struct db_stmt *stmt)
 
 		switch (b->type) {
 		case DB_BINDING_UNINITIALIZED:
-			db_fatal("DB binding not initialized: position=%zu, "
+			db_fatal(1, "DB binding not initialized: position=%zu, "
```
vincenzopalazzo, line 95:
```suggestion
			db_fatal(GENERIC_DB_ERROR, "DB binding not initialized: position=%zu, "
```

Do you think that it is the right moment to make a small refactoring in the error code? I started to put the error code in the `common/error.h` file, to make an easy life to catch the reason for failure from a bash script, and I think maybe we can put also some DB error?

db/db_sqlite3.c
```diff
@@ -41,7 +41,7 @@ static void replicate_statement(struct db_sqlite3 *wrapper,
 	sqlite3_finalize(stmt);
 
 	if (err != SQLITE_DONE)
-		db_fatal("Failed to replicate query: %s: %s: %s",
+		db_fatal(1, "Failed to replicate query: %s: %s: %s",
```
vincenzopalazzo, line 44: Same comment about the custom db error.

db/exec.c
```diff
@@ -0,0 +1,163 @@
...
+	s64 res = defval;
+	struct db_stmt *stmt = db_prepare_v2(
+	    db, SQL("SELECT intval FROM vars WHERE name= ? LIMIT 1"));
+	db_bind_text(stmt, 0, varname);
+	if (!db_query_prepared(stmt))
+		goto done;
+
+	if (db_step(stmt))
+		res = db_col_int(stmt, "intval");
+
```
vincenzopalazzo, line 87:
```suggestion
	if (db_query_prepared(stmt) && db_step(stmt))
		res = db_col_int(stmt, "intval");

```

Maybe it is much cleaner? and we can avoid the goto stm?

lightningd/Makefile
```diff
@@ -39,24 +39,26 @@ LIGHTNINGD_SRC :=				\
...
-
 LIGHTNINGD_SRC_NOHDR :=				\
 	lightningd/datastore.c			\
 	lightningd/offer.c			\
 	lightningd/signmessage.c
 
-LIGHTNINGD_HEADERS :=				\
+include wallet/Makefile
+
+LIGHTNINGD_HDRS :=				\
```
vincenzopalazzo, line 49: Why change the `LIGHTNINGD_HEADERS` to this more cryptic `LIGHTNINGD_HDRS` stuff?

vincenzopalazzo (changes requested):
I have for moment review only the code from style point of view, and maybe we need to jump also in the CI error that looks related to the db lock.

```
report saved to: report.json
===Flaky Test Report===

test_optimistic_locking failed (2 runs remaining out of 3).
	<class 'AssertionError'>
	assert None
 +  where None = <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f5810650b70>>('Optimistic lock on the database failed')
 +    where <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f5810650b70>> = <pyln.testing.utils.LightningD object at 0x7f5810650b70>.is_in_log
 +      where <pyln.testing.utils.LightningD object at 0x7f5810650b70> = <fixtures.LightningNode object at 0x7f581063ea90>.daemon
	[<TracebackEntry /home/runner/work/lightning/lightning/tests/test_db.py:306>]
test_optimistic_locking failed (1 runs remaining out of 3).
	<class 'AssertionError'>
	assert None
 +  where None = <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f581014ab70>>('Optimistic lock on the database failed')
 +    where <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f581014ab70>> = <pyln.testing.utils.LightningD object at 0x7f581014ab70>.is_in_log
 +      where <pyln.testing.utils.LightningD object at 0x7f581014ab70> = <fixtures.LightningNode object at 0x7f58101a5a58>.daemon
	[<TracebackEntry /home/runner/work/lightning/lightning/tests/test_db.py:306>]
test_optimistic_locking failed; it passed 0 out of the required 1 times.
	<class 'AssertionError'>
	assert None
 +  where None = <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f580ffafc18>>('Optimistic lock on the database failed')
 +    where <bound method TailableProc.is_in_log of <pyln.testing.utils.LightningD object at 0x7f580ffafc18>> = <pyln.testing.utils.LightningD object at 0x7f580ffafc18>.is_in_log
 +      where <pyln.testing.utils.LightningD object at 0x7f580ffafc18> = <fixtures.LightningNode object at 0x7f580ff85e48>.daemon
```

db/db_postgres.c
niftynei: No I don't think that this is the right time to do refactoring here but a follow up PR would be most welcome.

I'm not sure that pulling the database error code out is the right idea -- it's specifically separate to allow for catching errors in the *database* and therefore isn't a good candidate for addition to `common` -- keeping it in `db` is the correct choice here imo

lightningd/Makefile
niftynei: This name change was requested by @rustyrussell and updated accordingly.

vincenzopalazzo: Ops! My fault I missed this review sorry! Ignore it

wallet/db_queries.c
```diff
@@ -0,0 +1,13 @@
+#include "config.h"
+#include "db_postgres_sqlgen.c"
+
+#if HAVE_POSTGRES
```
cdecker, line 4: Seems this code is duplicated with the one in `wallet/db_queries_postgres.c`.

wallet/wallet.c
```diff
@@ -4677,6 +4677,7 @@ bool wallet_offer_create(struct wallet *w,
 	db_query_prepared(stmt);
 
 	if (db_step(stmt)) {
+		db_col_ignore(stmt, "1");
```
cdecker, line 4680: `SELECT 1` is a strange way of saying `SELECT COUNT(*)` :-) The latter would return a single row, whose only column we're interested in. Anyway, this is a reorg, not a cleanup, so I'll roll this into my weekly cleanup PR ^^

tests/test_db.py
```diff
@@ -303,7 +305,8 @@ def test_optimistic_locking(node_factory, bitcoind):
     with pytest.raises(RpcError, match=r'Connection to RPC server lost.'):
         l1.rpc.newaddr()
 
-    assert(l1.daemon.is_in_log(r'Optimistic lock on the database failed'))
```
cdecker, line 306: We shouldn't switch stream that we report errors on, people might be relying on the current behavior, e.g., not collect `stderr`. See the link-time specialization I mentioned 3 commits earlier.

cdecker (commented):
Making the DB abstraction layer available to other binaries was something that I was looking forward to for a long time, and I enjoyed revisiting the code. Thanks for tackling this and for the cleanups/reorgs.

ACK https://github.com/ElementsProject/lightning/pull/5017/commits/e4b4a541bc8553ad688e309c5d1db7759bc5cffd

wallet/db_queries.c
niftynei: Hmm yeah, I think I can clean this up even more. Let me give it a try...

wallet/wallet.c
niftynei: Hm, I wonder if there's a difference between `COUNT(1)` and `COUNT(*)`. Using stars in queries always makes me a bit uneasy, haha.

tests/test_db.py
niftynei: Agreed, this was definitely sub-optimal! (Though I will say that having the errors print to stderr made debugging problems while developing much much faster ;))

niftynei:
Ok, biggest changes are in ff3ff5bf40e25904940dbd4ed0dfca71452aaff4 and e16f49dd03b9cff8f36ebfc31ceba3fbb586bec3, which is new.

For ff3ff5b, took @cdecker's suggestion for implementing `db_fatal` in the section where it's used. We use a `#ifndef` guard for it in `wallet/wallet.c`, so that the unit test code can still override the declaration.

e16f49d is new. It moves the boilerplate db_query code and puts them into the generated sqlgen files instead.

rustyrussell (approved):
Ack 98e69252e5e146475efc1ed76af65bbccbb3bf94

PR #5019, opened 2022-01-25
`coin_movement` updates

niftynei:
While working on the accounting plugin, I ran into some problems/limitations of our current event emissions.

1) We weren't tracking withdrawals to external wallets (initiated via `withdraw`/`multiwithdraw`). This makes it really difficult to see where our funds are going, as well as confuses the onchain fee tracker that the accountant plugin has. Now, we emit a `coin_movement` for 

2) Funds sent to 'external' accounts don't say which account they originated from. While it's possible to figure this out from the utxo tracking, that's sometimes difficult to reconstruct. Instead, for every `external` deposit, we now record what internal account the payment originated from. For example, wallet payments to an external account will be tagged 'wallet'. Payments to the peer in a channel will 'originate' from the channel's account.

### DIFF
--- a/common/coin_mvt.c
+++ b/common/coin_mvt.c
@@ -261,11 +261,9 @@ struct chain_coin_mvt *new_coin_external_deposit(const tal_t *ctx,
 						 struct amount_sat amount,
 						 enum mvt_tag tag)
 {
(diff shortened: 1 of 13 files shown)


rustyrussell (approved):
Ack 050f37788ca2dccaabaea80418641ee6d103caa9

PR #5021, opened 2022-01-26
misc: Add build targets for the tarball and debian packages

cdecker:
The tarball needs to materialze all submodules, and git needs to be
removed as a build dependency. I thought @rustyrussell might find these
useful, so we can distribute the deb packages via the releases page. The
tarball is conforming to the debuild format, hence the extra work.

Changelog-None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -48,7 +48,24 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
 FROM debian:buster-slim as builder
 
 ENV LIGHTNINGD_VERSION=master
-RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates autoconf automake build-essential git libtool python3 python3-pip python3-setuptools python3-mako wget gnupg dirmngr git gettext
+RUN apt-get update -qq && \
+    apt-get install -qq -y --no-install-recommends \
+        autoconf \
+        automake \
+        build-essential \
+        ca-certificates \
+        dirmngr \
+        gettext \
+        git \
+        gnupg \
+        libtool \
+        python3 \
+        python3-mako \
+        python3-pip \
+        python3-setuptools \
+        wget
+
+RUN pip3 install -U setuptools mrkd mako
 
 RUN wget -q https://zlib.net/zlib-1.2.11.tar.gz \
 && tar xvf zlib-1.2.11.tar.gz \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -48,7 +48,24 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
...
+        git \
+        gnupg \
+        libtool \
+        python3 \
+        python3-mako \
+        python3-pip \
+        python3-setuptools \
+        wget
+
+RUN pip3 install -U setuptools mrkd mako
```
vincenzopalazzo, line 68:
```suggestion
RUN pip3 install -U setuptools mrkd mako mistune==0.8.4
```

Little worried, it here we need to set also the `mistune==0.8.4`, but I didn't check if the dev of `mrkd` pushed the new version on `pypi`

cdecker: Just use requirements.lock which has a known good combination of dependencies resolved to their respective versions. Having to replace all these tiny version pins whenever we upgrade is tedious.

vincenzopalazzo (commented):
The CI looks unrelated in the sense that there is some code style checker that looks in the wrong file maybe?

Dockerfile
vincenzopalazzo: ops! Sorry if I review when you are working on it, concurrency is always bad!

In addition, I like the idea to use `requirements.lock` everywhere, but in this case, I did mention it because I thought that maybe with the release we don't need some dependencies.

But I was wrong :)

cdecker:
Nope, that error is correct, as it is complaining about my use of unquoted variable Expansion, which is unsafe if the variables contain word delimiters. In this case it's safe, however it's always better to use quotes in these cases, so I'll fix it up 👍

vincenzopalazzo (approved):
LGTM,

ack https://github.com/ElementsProject/lightning/pull/5021/commits/bec4f6f995288ca747169eefa03b42686440ca06

configure
```diff
@@ -35,7 +35,7 @@ usage_with_default()
 default_coptflags()
 {
     if [ "$1" = 0 ]; then
-	echo "-Og"
+	echo ""
```
rustyrussell, line 38: This is a change?  We used to build -Og if !DEVELOPER, which is a compromise, but a nice one...

cdecker: I don't remember the exact issue, but at least one compiler on Launchpad was complaining about this flag. Let me see if I can find it again, otgherwise I'm happy to revert.

rustyrussell: Damn, I hit apply automatically :(

I'll do another PR which reverts this, sorry...

Issue #4734, opened 2021-08-22
Feature suggestion: Better inspection of closed channels


urza:
Inspecting closed channels is currently not easy with the cli. There is some information in db, but putting it together this way is obviously not ideal.

I noticed that this was already mentioned somewhere by @vincenzopalazzo but can't find it. 

Ideally I would like to be able to get this with clightning-cli somehow:

- list closed channels and their properties (e.g. was this public or private channel)
- type of close (mutual, unilateral)
- who initiated closing
- closing onchain tx id

vincenzopalazzo:
Hi @urza, I started to work on one PR https://github.com/ElementsProject/lightning/pull/4568 to add this feature, but I think now it is a good moment to put this feature in the next release with this issue https://github.com/ElementsProject/lightning/issues/4729

It looks like already been taken, lets see what happens.

rl-d:
@all, most of the data could be made available with an optional parameter to filter the peer channels by state (and I am happy to add that to issue #4729), but there would be two caveats with that approach:

1. after 100 blocks the channels wouldn't be reported anymore, and
2. the onchain tx id does not currently seem to be part of the peer channel data.

What is your advice?

vincenzopalazzo:
I don't remember what are the data in the database about the closed channels, maybe @cdecker can help here. Or maybe you can tell MySQL to describe the table.

rl-d:
Doing the documentation changes for this, I noticed there is a "scratch" tx id which I guess would be equivalent to @urza's onchain tx id of caveat #2 above. So I think the big points are covered at least.

If I've misunderstood anything, I stand to be corrected... thanks!

rl-d:
It's been recommended that this be made a plugin, and the use case is such that it should list historical data (older than 100 blocks), so it's really out of scope for inclusion with issue #4729.

cdecker:
Small correction: we no longer delete entries in the `channels` table as far as a I can see, so it should always be possible to retrieve at least some of the information in the table. We can definitely add an option to include closed channels as well, and should likely set that to true if we filter for `state=closed` as well.

rl-d:
Thanks for the note @cdecker. I am working on a plugin that will implement a GraphQL interface for querying the data that the user is interested in, which (in light of your note) should satisfy this feature request nicely.

urza:
Has there been any new development? What is the best way to inspect closed channels now? One of my channels was closed and I would like to know (using CLI):

- was it cooperative close?
[...]

PR #5022, opened 2022-01-29
rpc: remove channels from listpeers in listpeerchannels

vincenzopalazzo:
Before the PR https://github.com/ElementsProject/lightning/pull/4821 was open, I started to work on the issue https://github.com/ElementsProject/lightning/issues/4729 because the actual status of the network makes the operation slow, and my rest API is starting to begin frustrating.

So I take the good work made in the #4821 PR and I start to integrate my old branch in it hoping to have this PR read before the next release. 

Fixes https://github.com/ElementsProject/lightning/issues/4729
after that, I will move on this one https://github.com/ElementsProject/lightning/issues/4734

Override https://github.com/ElementsProject/lightning/pull/4821

For now, I will give only a run in the CI to fix some eventual failures before starting to merge my work


vincenzopalazzo:
During my war with the test refactoring I found the following error in the CI that seems undelated

```
lightningd-1: 2022-02-09T16:32:08.501Z INFO    plugin-pay: cmd 73 partid 1: Cannot attempt payment, we have no channel to which we can add an HTLC
DEBUG:root:Received response for pay call: {'jsonrpc': '2.0', 'id': 1, 'error': {'code': 210, 'message': 'Ran out of routes to try after 1 attempt: see `paystatus`', 'attempts': [{'status': 'failed', 'failreason': 'Cannot attempt payment, we have no channel to which we can add an HTLC', 'partid': 1, 'amount': '500000000msat'}]}}
DEBUG:root:{
  "id": 1,
  "method": "listfunds",
  "params": {
    "spent": null
  }
}
DEBUG:root:Calling listfunds with payload {'spent': None}
DEBUG:root:Received response for listfunds call: {'jsonrpc': '2.0', 'id': 1, 'result': {'outputs': [{'txid': '3aadb37fa7a375ec8447876e282f81ad1f1181f6f2ee0192ccafb58517a2ff25', 'output': 0, 'value': 995433, 'amount_msat': 995433000msat, 'scriptpubkey': '0014c2ccab171c2a5be9dab52ec41b825863024c5466', 'address': 'bcrt1qctx2k9cu9fd7nk449mzphqjcvvpyc4rxh6826x', 'status': 'confirmed', 'blockheight': 103, 'reserved': False}], 'channels': [{'peer_id': '022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59', 'connected': True, 'state': 'CHANNELD_NORMAL', 'short_channel_id': '103x1x1', 'channel_sat': 1000000, 'our_amount_msat': 1000000000msat, 'channel_total_sat': 1000000, 'amount_msat': 1000000000msat, 'funding_txid': '3aadb37fa7a375ec8447876e282f81ad1f1181f6f2ee0192ccafb58517a2ff25', 'funding_output': 1}]}}
```

And the pay status tell us the following info

```
DEBUG:root:{
  "id": 1,
  "method": "paystatus",
  "params": {
    "bolt11": null
  }
}
DEBUG:root:Calling paystatus with payload {'bolt11': None}
DEBUG:root:Received response for paystatus call: {'jsonrpc': '2.0', 'id': 1, 'result': {'pay': [{'bolt11': 'lnbcrt5m1p3q8mygpp5yyljt9wvgjldv84ltj2zwh7x2chvae3elyuvs7znm0f6vtj67nnsdqdvfskcctwvdjhyxqyjw5qcqp9sp5j05cdyaz0hawm9gjrendj3h9c38jl8j65jvqtqheeevpr2uxkzzsrzjqdwjkyvjm7apxnssu4qgwhfkd67ghs6n6k48v6uqczgt88p6tky96qqqvuqqqqcqqyqqqqqpqqqqqzsqqc9qyyssqpd5u8xcnaj3sjvag8gp5xh66ah3s20t5xlgmvnswgauguzmygyz9syt7mmsz0r9supmsxk72jad0kjwdkutkzcgggyl8g8je5pjeduqpmldge4', 'amount_msat': 500000000msat, 'destination': '022d223620a359a47ff7f7ac447c85c46c923da53389221a0054c11c1e3ca31d59', 'attempts': [{'strategy': 'Initial attempt', 'start_time': '2022-02-09T16:32:08.493Z', 'age_in_seconds': 0, 'end_time': '2022-02-09T16:32:08.500Z', 'state': 'completed', 'failure': {'code': 205, 'message': 'Call to getroute: Could not find a route'}}]}]}}
```

lightningd/peer_control.c
```diff
@@ -1499,6 +1501,53 @@ static const struct json_command listpeers_command = {
...
+{
+	enum log_level *ll;
+	struct node_id *peer_id;
+	struct peer *peer;
+	struct channel *channel;
+	struct json_stream *response;
+
+	if (!param(cmd, buffer, params,
+		   p_req("id", param_node_id, &peer_id),
+		   p_opt("level", param_loglevel, &ll),
```
rustyrussell: id should be opt: default is all peers.  And level is not useful here...

rustyrussell (approved):
I think this ended up OK, but should really be refactored into two commits:
1. Introduce new listpeerchannels.
2. Deprecate channels field in listpeers and switch all the tests over.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -937,6 +937,17 @@ def listpeers(self, peerid=None, level=None):
         }
         return self.call("listpeers", payload)
 
+    def listpeerchannels(self, peer_id=None, level=None):
+        """
```
rustyrussell, line 941: level should not exist any more!

lightningd/peer_control.c
vincenzopalazzo: Ops! I grab the wrong history range of commits from the https://github.com/ElementsProject/lightning/pull/4821

This should be over with the latest changes.

plugins/topology.c
```diff
@@ -321,76 +296,22 @@ struct listchannels_opts {
...
 
-/* We record which local channels are valid; we could record which are
- * invalid, but our testsuite has some weirdness where it has local
- * channels in the store it knows nothing about. */
-static struct node_map *local_connected(const tal_t *ctx,
-					const char *buf,
-					const jsmntok_t *result)
-{
-	size_t i;
-	const jsmntok_t *t, *peers = json_get_member(buf, result, "peers");
```
vincenzopalazzo, line 332: I removed this because I was thinking that it was introduced for historical reasons.

However, I'm sure that this change will cause [this failure](https://github.com/ElementsProject/lightning/runs/5501615365?check_suite_focus=true#step:5:8659) because the `getroute` use the `listchannels` command, but I'm not able to find a good reason to introduce the function `local_connected` again

vincenzopalazzo (commented):
Changed applied, sorry for the commit confusing, but I was trying to recover an old PR.

In addition, I was stuck with the [following comment](https://github.com/ElementsProject/lightning/pull/5022/files#r824217534) that may be the cause of the CI failure.

Any feedback on this?

plugins/topology.c
rustyrussell: No, you will need to restore this!  It can work pretty well off listpeerchannels, no?

rustyrussell (changes requested):

plugins/topology.c
vincenzopalazzo: Right, My original idea was to add a field to the `listpeerchannels` result to avoid an extra caching operation in the `listchannels`.

I'm reintroducing it with a little bit of rework, lets's see what the CI tell us with this last change, on my machine all the test execution is too slow, so sorry if there future errors in the CI

instagibbs:
nit: https://github.com/ElementsProject/lightning/pull/5022/commits/0dc1342d357d711e6d6bb255822d3bf85ec112cc
claims to be deprecation commit but seems to only be testing changes

vincenzopalazzo:
>claims to be deprecation commit but seems to only be testing changes

Yes, I just write the comment in the wrong way, but in this PR I updated the test, due to the deprecated API.

But this PR is still in draft and all the coming need to be reorganize, because are a little bit messy

doc/schemas/listpeers.schema.json
```diff
@@ -165,954 +164,6 @@
                 }
               ]
             }
-          },
-          "channels": {
```
joemphilips, line 169: Instead of removing everything from schema at once, would you please just mark as "deprecated"?

vincenzopalazzo: it is deprecated in cln, but I do not know any type of way to deprecate it in the schema, maybe @rustyrussell know this?

vincenzopalazzo: But I think this change will make some pain when you will go to generate some type? I think here the best solution is use a general call

joemphilips: I've just want this change to follow a general field-deprecation cycle described in https://github.com/ElementsProject/lightning/blob/master/doc/STYLE.md#changing-json-apis
So maybe create another PR just marking this field as deprecated and merge that first?

If that causes too much pain in msggen... well, I suppose there is no reason to hurry in deprecating this field. So we should brush up msggen first?

joemphilips: Note: I made a PR which changes msggen to allow more gradual deprecation cycle: https://github.com/ElementsProject/lightning/pull/5536

vincenzopalazzo: Wait, I do not follow in this PR the deprecation rules? I think I follow

```
if (deprecated_apis) {
		json_array_start(response, "channels");
		json_add_peerchannels(ld, response, p);
		json_array_end(response);
	}
```

I do not know only how express this inside the json schema, do you have any idea?

vincenzopalazzo: a possibility is add a deprecated prefix in the description of the json schema filed maybe? but I do not know, I do not think deeply to this idea

joemphilips: > a possibility is add a deprecated prefix in the description of the json schema filed maybe?

Yes, that is what I was expecting. For 1. deprecation of rpc implementation, I think you are doing fine. but for 2. deprecation of rpc json schema, there are also guideline to annotate deprecated fields https://github.com/ElementsProject/lightning/blob/master/doc/schemas/WRITING_SCHEMAS.md#how-to-write-a-schema

(I should have posted this link at first, sorry.)

vincenzopalazzo: You are right! thanks for this catch, I was assuming that we have no deprecated in the docs and we can not deprecate a schema!

Thanks!

vincenzopalazzo:
Closing in favor of https://github.com/ElementsProject/lightning/pull/5568

PR #5024, opened 2022-01-31
Update LICENSE

MarnixCroes:
Update LICENSE to 2022


cdecker:
@rustyrussell do we have to have a well-defined range for the copyright at all? Wouldn't `2015 - ` be ok too? Don't want to have to change this every year.

PR #5030, opened 2022-02-05
doc: reintroduce the fmt command to fmt the schema

vincenzopalazzo:
For some reason, there is no fmt schema command anymore, but just a check it.

So, this command now it is introduced by a helper command that needs to be expressed called by the human with `fmt-schemas`, and it help to format a very big JSON schema like the one introduced in this PR https://github.com/ElementsProject/lightning/pull/5022

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5030/commits/792b5786e4421523483b07f4e60659fca6482db7

PR #5031, opened 2022-02-07
Add short_channel_id to listchannels.schema.json

AaronDewes:
This was previously marked as required, but not actually defined in properties.

I found this while working on [c-lightning.ts](https://github.com/runcitadel/c-lightning-client).

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5031/commits/3c5ef82f2be2e1029faf0fbd41c23c04eed29685

rustyrussell:
Huh.  Good catch!

The correct fix is to set  "additionalProperties": false, let me take a look what that breaks!

rustyrussell:
Added that, regenerated the man paged, and did trivial rebase on master (hopefully will fix test flake!)

PR #5034, opened 2022-02-08
gossipd: don't send updates in error messages for unannounced channels.

rustyrussell:
This restores the behaviour prior to `lightningd: use our cached
channel_update for errors instead of asking gossipd.`, where gossipd
would refuse to give us channel_updates for unannounced channels.

@t-bast pinged for opinions!

### DIFF
--- a/gossipd/gossip_generation.c
+++ b/gossipd/gossip_generation.c
@@ -420,9 +420,13 @@ static u8 *sign_and_timestamp_update(const tal_t *ctx,
 		tal_free(unsigned_update);
 
 	/* Tell lightningd about this immediately (even if we're not actually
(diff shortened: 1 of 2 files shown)


t-bast:
What exactly do you send in `error`s that require a `channel_update` then? The `channel_update` with a dummy `short_channel_id`?

In eclair we currently include unannounced `channel_update`s in `error`s (for simplicity's sake). We considered fixing this, but figured it wasn't worth it yet as the privacy of unannounced channels wasn't great anyway. But I think we should do something about it at some point, so if c-lightning was not including these `channel_update`s, I think it's good to keep not including them!

PR #5036, opened 2022-02-10
bolt11: prevent duplicate min_final_cltv_expiry and expiry tags

jb55:
When decoding bolt11 tags, it looks like there are checks (`have_c`, `have_x`,
etc) which prevent duplicate tags. Unfortunately `decode_c` doesn't set
`have_c` unlike the other `decode_` methods. At the start of the function,
`decode_c` checks `have_c` to see if it's set, but it is never set. It seems
like this could allow for duplicate `c` tags, which is probably not intended.

It looks like there's the same issue with the `x` (expiry) tag as well.

### DIFF
--- a/common/bolt11.c
+++ b/common/bolt11.c
@@ -224,7 +224,7 @@ static void decode_h(struct bolt11 *b11,
 static char *decode_x(struct bolt11 *b11,
 		      struct hash_u5 *hu5,
 		      u5 **data, size_t *data_len,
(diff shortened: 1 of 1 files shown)


rustyrussell (approved):
Ack 63a05caa3b8903a55c71ae6768a4e7af9fb2582d

rustyrussell:
This is great! We actually allow duplicates, but we add them to the "unknown" fields.

PR #5043, opened 2022-02-16
More coin_move adjustments

niftynei:
A few more things fell out in the wash, so to speak, while working on the upcoming `bookkeeper` plugin.


rustyrussell (approved):
Ack 995b3df1085ac6841e31be3e14f074d41bbdc40a

PR #5045, opened 2022-02-17
Fix new bitcoind compat

rustyrussell:
New machine, new bitcoind, new API breakage.

Plus it uncovered an error in our Tor test (since my new machine has Tor enabled, this test actually runs).


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5045/commits/e7c2eea3aa48bd447c23c6a344dd5fad222fbcfd

PR #5050, opened 2022-02-22
pytest: remove test_htlc_rexmit_while_closing as too flaky.

rustyrussell:
This is breaking CI too much :(


vincenzopalazzo (approved):
I started to work in the migration in lnprotottest, if I will have problems, I will ping you on IRC, but I agree that this type of test now should be in a better env like lnprototest

ack https://github.com/ElementsProject/lightning/pull/5050/commits/8abd659b6a8deb7a7cd4f4aab94cc62a42e93639

Issue #5035, opened 2022-02-09
Huge memory "leak" in wire/wire_io.c


whitslack:
I've been seeing huge (multiple gigabyte) memory consumption by `lightning_hsmd`, and I've been trying to get to the bottom of it. I tweaked `lightningd` so that it launches `lightning_hsmd` under [Heaptrack][]. I let it run for a bit over three weeks, at which time the heap memory consumption of `lightning_hsmd`, as reported by Heaptrack, had grown to 2.7 GB. Heaptrack's analysis points the finger squarely at `wire/wire_io.c`, though by studying the code, it appears to me that the problem is not really in `wire/wire_io.c` but rather in how it's being used.

`io_read_wire_(…)` [allocates](https://github.com/ElementsProject/lightning/blob/v0.10.2/wire/wire_io.c#L77) a buffer to hold the message header to be received, and `do_read_wire_header(…)` [resizes](https://github.com/ElementsProject/lightning/blob/v0.10.2/wire/wire_io.c#L42) that buffer (if necessary) to hold the message header and payload. That's fine; it then becomes the responsibility of the caller of `io_read_wire(…)` to free the allocated buffer.

When we look in `hsmd/hsmd.c`, we see that `client_read_next(…)` [calls](https://github.com/ElementsProject/lightning/blob/v0.10.2/hsmd/hsmd.c#L171) `io_read_wire(…)`, passing `&c->msg_in` as the address of a pointer that will be set to the address of the buffer that `io_read_wire_(…)` will allocate, and passing `c` (a pointer to the `struct client` instance) as the parent for the new allocation. As long as the `struct client` instance eventually gets freed, the allocated message buffer will be freed too, so there is no *leak* in the strict sense of the term, but the freeing of the buffer may not occur for an arbitrarily long time after the buffer has become disused, and indeed many millions of message buffers may be allocated within the lifetime of one `struct client` instance.

Looking through `hsmd/hsmd.c`, I see nowhere that the `c->msg_in` is ever explicitly freed. What I see is that `handle_client(…)` ultimately hands off the `c->msg_in` to one of several message-type-specific handler functions, and those functions are not `TAKES` or `STEALS` on their message buffer parameters and do not free their message buffer arguments. Consequently, each successive call to `client_read_next(…)` will cause `io_read_wire_(…)` to overwrite the `c->msg_in` pointer with the address of a newly allocated message buffer, and the old buffer will be left dangling off of the `struct client` instance indefinitely.

[...]

PR #5051, opened 2022-02-23
hsmd: don't leak message buffers

whitslack:
`client_read_next(…)` calls `io_read_wire(…)`, passing `&c->msg_in` as the address of a pointer that will be set to the address of the buffer that `io_read_wire_(…)` will allocate, and passing `c` (a pointer to the `struct client` instance) as the parent for the new allocation. As long as the `struct client` instance eventually gets freed, the allocated message buffer will be freed too, so there is no *leak* in the strict sense of the term, but the freeing of the buffer may not occur for an arbitrarily long time after the buffer has become disused, and indeed many millions of message buffers may be allocated within the lifetime of one `struct client` instance.

`handle_client(…)` ultimately hands off the `c->msg_in` to one of several message-type-specific handler functions, and those functions are not `TAKES` or `STEALS` on their message buffer parameters and do not free their message buffer arguments. Consequently, each successive call to `client_read_next(…)` will cause `io_read_wire_(…)` to overwrite the `c->msg_in` pointer with the address of a newly allocated message buffer, and the old buffer will be left dangling off of the `struct client` instance indefinitely.

Fix this by initializing `c->msg_in` to `NULL` in `new_client(…)` and then having `client_read_next(…)` do `c->msg_in = tal_free(c->msg_in)` prior to calling `io_read_wire(…)`. That way, the previous message buffer will be freed just before beginning to read the next message. The same strategy is already employed in `common/daemon_conn.c`, albeit without nulling out `dc->msg_in` after freeing it.

**Fixes:** #5035
**Changelog-Fixed:** hsmd: Fixed a significant memory leak

### DIFF
--- a/hsmd/hsmd.c
+++ b/hsmd/hsmd.c
@@ -168,6 +168,7 @@ static struct io_plan *bad_req(struct io_conn *conn,
  * and then call handle_client with argument 'c' */
 static struct io_plan *client_read_next(struct io_conn *conn, struct client *c)
 {
+	c->msg_in = tal_free(c->msg_in);
 	return io_read_wire(conn, c, &c->msg_in, handle_client, c);
 }
 
@@ -189,6 +190,8 @@ static struct client *new_client(const tal_t *ctx,
 				 int fd)
 {
 	struct client *c = tal(ctx, struct client);
+
+	c->msg_in = NULL;
 
 	/*~ All-zero pubkey is used for the initial master connection */
 	if (id) {


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5051/commits/0a30efe9a5c7e7f078ac91a0e643216c481a0656

rustyrussell:
This is great, but why didn't our memleak detection catch it?

... ah, because I broke it in 6c9b75275191f0efdf1036ec5b26aa334662462b!

Fortunately that's since last release...

whitslack:
> This is great, but why didn't our memleak detection catch it?

@rustyrussell: It's not strictly a memory leak, per se. The buffers have valid ancestry chains up to a notleak root. But that doesn't preclude them from piling up in memory.

I would refer you to this sentence of my commit message:

> As long as the `struct client` instance eventually gets freed, the allocated message buffer will be freed too, so there is no _leak_ in the strict sense of the term, but the freeing of the buffer may not occur for an arbitrarily long time after the buffer has become disused…

rustyrussell:
> > This is great, but why didn't our memleak detection catch it?
> 
> @rustyrussell: It's not strictly a memory leak, per se. The buffers have valid ancestry chains up to a notleak root. But that doesn't preclude them from piling up in memory.
> 
> I would refer you to this sentence of my commit message:
> 
> > As long as the `struct client` instance eventually gets freed, the allocated message buffer will be freed too, so there is no _leak_ in the strict sense of the term, but the freeing of the buffer may not occur for an arbitrarily long time after the buffer has become disused…

Still should be detected by our leak detector!  And, once I fixed it, it is...

PR #5052, opened 2022-02-23
gossipd: use remote addr

m-schmoock:
# What it does:
This is the second part of the "IP discovery by remote peer feature" that actually uses the `remote_addr` field to create `node_annoucement` updates when clightning detects that our peers report the same new `remote_addr`.

Currently implements the most simple algorithm I came up with that just checks:
 - two distinct peers
 - we have a channel with
 - report the same `remote_addr`
 - if so, we send an updated `node_announcement`

There is a failsafe that prevents this code from running when we have `always_use_proxy` enabled. Normally, this code would not be executed for TOR connections as a TOR node would not set the `remote_addr` to something useful other than a TOR string that we would ignore, but a nefarious TOR node may respond (thorugh the proxy) with a init TLV having an IP `remote_addr` set that we would otherwise use.

# NextSteps/Todos:
  - direct probing a random node on the network to confirm a `remote_addr` qickly.
  - documentation about this, so people know they can open the port on the router instead of using TOR.
  - IPv4: make a self connection for confirmation. If it works we should announce as we know for sure this is the right address. Also we don't need to wait for another node to confirm that. Note: This does not apply to IPv6, as a self connection will be handeled locally (because the IPv6 address is bound locally and not at the router) and that does not tell us anything if the router has the ports forwarded or not. Also this will likely not work for a lot of routers
 
# Questions:
 - Do we need a config switch for using `remote_addr` to send updated `node_announcement`?
This is decision has tradeoffs. In general it is not harmful to announce one additional address that is potentially unreachable (if the port at the router is not open). However it would affect everyone. Having a config switch increases the complexity for people using this.
Note: rusty noted we may introduce a config switch to turn that off but default to enabled.
Note: if we have a unset config switch but no announced addresses at all, we may likely also ignore the config
Note: If we have a disabled config switch, but we can make a succesful IPv4 self-connection (see above) we should ignore that config switch and send the update any way, as it will improve the situation.  (not sure if


m-schmoock:
just solved conflicts and rebased on master b5a1715c2

rustyrussell:
I don't think an IPv4 self-connection is reliable either.  At least with older Linux NAT setups it won't work; I'm surprised it does for you!

lightningd/peer_control.c
```diff
@@ -1107,6 +1107,47 @@ peer_connected_hook_deserialize(struct peer_connected_hook_payload *payload,
...
+	if (ld->remote_addr_v4 == NULL) {
+		ld->remote_addr_v4 = tal(ld, struct wireaddr);
+		*ld->remote_addr_v4 = *remote_addr;
+		ld->remote_addr_v4_peer = peer_id;
+	}
+	if (ld->remote_addr_v6 == NULL) {
+		ld->remote_addr_v6 = tal(ld, struct wireaddr);
+		*ld->remote_addr_v6 = *remote_addr;
+		ld->remote_addr_v6_peer = peer_id;
+	}
```
rustyrussell, line 1127: This seems wrong: you set remote_addr_v4 potentially to a v6 address here?

lightningd/peer_control.c
```diff
@@ -1107,6 +1107,47 @@ peer_connected_hook_deserialize(struct peer_connected_hook_payload *payload,
 	return true;
 }
 
+/* Compare and store `remote_addr` and the `peer_id` that reported it.
+ * If new address was reported by at least one other, do node_announcement */
+static void update_remote_addr(struct lightningd *ld,
+			       struct wireaddr *remote_addr,
+			       struct node_id peer_id)
```
rustyrussell, line 1115: *const* struct wireaddr.  And pass node_id by copy, so "const struct node_id *".

lightningd/peer_control.c
```diff
@@ -1107,6 +1107,47 @@ peer_connected_hook_deserialize(struct peer_connected_hook_payload *payload,
...
+			break;
+		*ld->remote_addr_v4 = *remote_addr;
+		ld->remote_addr_v4_peer = peer_id;
+		break;
+	case ADDR_TYPE_IPV6:
+		if (node_id_eq(&ld->remote_addr_v6_peer, &peer_id))
+			break;
+		*ld->remote_addr_v6 = *remote_addr;
+		ld->remote_addr_v6_peer = peer_id;
+		break;
```
rustyrussell, line 1149: So, if the same node gives you a new address, you ignore it?  That seems weird.

I'd expect this code to be:

```
        /* Update any v4/v6 address if they've given one */
        switch (remote_addr->type) {
	case ADDR_TYPE_IPV4:
                tal_free(ld->remote_addr_v4);
                ld->remote_addr_v4 = tal_dup(ld, struct wireaddr, remote_addr);
                ld->remote_addr_v4_peer = *peer_id;
       ...
```

gossipd/gossip_generation.c
```diff
@@ -19,6 +20,16 @@
 #include <hsmd/hsmd_wiregen.h>
 #include <wire/wire_sync.h>
 
+static bool wireaddr_contained(struct wireaddr *was, struct wireaddr *wa)
```
rustyrussell, line 23: Both arguments should be const...

gossipd/gossip_generation.c
```diff
@@ -19,6 +20,16 @@
...
+{
+	size_t i;
+
+	for (i = 0; i < tal_count(was); i++)
```
rustyrussell, line 27: Mildly prefer more compact (modern!) `for (size_t i = ...`

gossipd/gossip_generation.c
rustyrussell: - Maybe rename to "wireaddr_arr_contains"?

gossipd/gossipd.c
```diff
@@ -341,6 +342,53 @@ static void handle_local_private_channel(struct daemon *daemon, const u8 *msg)
...
+
+	switch (remote_addr.type) {
+	case ADDR_TYPE_IPV4:
+		if (daemon->remote_addr_v4 != NULL &&
+		    wireaddr_eq_without_port(daemon->remote_addr_v4,
+					     &remote_addr))
+			break;
+		if (daemon->remote_addr_v4 == NULL)
+			daemon->remote_addr_v4 = tal(daemon, struct wireaddr);
+		*daemon->remote_addr_v4 = remote_addr;
```
rustyrussell, line 365: I prefer a "tal_free and tal_dup" pattern in general.  Sure, it's a bit wasteful to reallocate, but it's clear.

tests/test_connection.py
```diff
@@ -63,6 +63,65 @@ def test_connect_basic(node_factory):
         l1.rpc.connect('032cf15d1ad9c4a08d26eab1918f732d8ef8fdc6abb9640bf3db174372c491304e', 'localhost', l2.port)
 
 
+@pytest.mark.developer("needs DEVELOPER=1 for having localhost remote_addr and fast gossip")
+@unittest.skipIf(not EXPERIMENTAL_FEATURES, "BOLT1 remote_addr #917")
+def test_remote_addr(node_factory, bitcoind):
+    """Check address discovery (BOLT1 #917) init remote_addr works as designed:
+
+       `node_annoucemnet` update must only be send out when:
```
rustyrussell, line 71: typo

tests/test_connection.py
```diff
@@ -63,6 +63,65 @@ def test_connect_basic(node_factory):
...
+    l1.restart()
+    l2.rpc.connect(l1.info['id'], 'localhost', l1.port)
+    l2.daemon.wait_for_log("Peer says it sees our address as: 127.0.0.1:[0-9]{5}")
+
+    # now only l1 sees l2 without announced addresses (disabled by opts)
+    assert(len(l1.rpc.listnodes(l2.info['id'])['nodes'][0]['addresses']) == 0)
+    assert(len(l3.rpc.listnodes(l2.info['id'])['nodes']) == 0)
+    assert not l2.daemon.is_in_log("Update our node_announcement for discovered address: 127.0.0.1:9735")
+
+    # connect second node. This will not yet trigger `node_annocement` update,
```
rustyrussell, line 101: typo

rustyrussell (approved):
Nice series!  Some minor feedback only (and needs trivial rebase).

lightningd/peer_control.c
m-schmoock: Jup, thanks.

lightningd/peer_control.c
m-schmoock: Yes, I need to store the latest address of the same peer.
I rewrote the block.

m-schmoock:
... applied the review outcome.
I still need to add the "disable feature for tor/proxy only nodes" and remove the EXPERIMENTAL before we can merge this.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/dda19f50acecaa0a6aa44c4572673b6db44751a4

Thanks!

heyambob:
> Do we need a config switch for using remote_addr to send updated node_announcement? This is decision has tradeoffs. In general it is not harmful to announce one additional address that is potentially unreachable (if the port at the router is not open). However it would affect everyone. Having a config switch increases the complexity for people using this. Note: rusty noted we may introduce a config switch to turn that off but default to enabled.

I think it might not be a desire behavior for a lot of people due to privacy leak. I use tor + always-use-proxy=false to have some sort of privacy and better performance when peering with clearnet nodes (they should be the only ones who knows my clearnet addr from what I understand). However, when c-lightning announces my address, the entire network knows it, and sites like amboss pick it up make a public record that last forever. Even though my clearnet address is unreachable as I don't NAT, but people now easily knows my home.

It's a smart feature, but if it turns on by default, I can see node operators be in shock after update.

https://github.com/ElementsProject/lightning/issues/5109

m-schmoock:
@heyambob @rustyrussell
You might have a point here. We can discuss this in todays dev meeting on Jitsi.

Currently (without this PR), if you use a TOR but without `always_use_proxy` your IP address is not really a secret, because your node will still make normal connections if possible (and even prefer those over TOR). An attacker 'Charlie' can trick you into disclosing your IP address by:
 - Charlie opens a channel to you via a TOR connection
 - Charlie also announces its IP address
 - Charlie removes TOR from his config and restarts its node
 - Your node will connect to Charlie via IP

Also your node can and will do other requests that revewl your IP without `always_use_proxy` such as DNS seed, plugins discover prices, and so on... **Though, a big difference is that only Charlie knows your IP and not everyone else.**

My assumption is that 99% of the people that use TOR without `always_use_proxy` just do so to get past their NAT router/firewall.

heyambob:
Right. My concern is for people who want to keep low profile, not for those who want to stay anonymous. 

>My assumption is that 99% of the people that use TOR without always_use_proxy just do so to get past their NAT router/firewall.

Initially, people play around, they might not care so much. Once they get comfortable and commit a meaningful liquidity to be a useful node, they could be regret. At least there has to be
 config to turn it off.

m-schmoock:
@heyambob addressed by https://github.com/ElementsProject/lightning/pull/5136

PR #5033, opened 2022-02-07
pyln-client: add constraints dependeces to pyln dep, to avoid recursion

vincenzopalazzo:
In the CI we don't see the recursion, but in some cases, it can cause the recursion.

I had this problem in lnprototest during the poetry integration.

Hope that this can solve the problem

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

PR #5054, opened 2022-02-24
Migrate pyln packages to poetry / PEP 517

cdecker:
The dependency resolution issues have gotten really annoying, so let's address that by instead using `poetry` to handle python packages. This more modern approach has a number of advantages:

 - Support for lock-files: we can use the `poetry.lock` file in the root as a way to reproducibly pin packages to a version we know works. The `pyln` packages should not get a `poetry.lock` since that's discouraged for libraries as it interferes with user-code dependency resolution
 - We have a base `cln-meta-project` configuration in the root folder allowing us to have a single command bootstrap of the environment.
 - PEP 517 allows relative package dependencies, so we no longer have to iterate through `pyln` packages, install them in the correct order, and hope that they don't clash, or fall back on pypi.org due to an incompatibility.

Overall this should reduce if not eliminate the issues we've been encountering.

Changelog-None


contrib/pyln-proto/pyln/proto/__init__.py
```diff
@@ -3,7 +3,8 @@
 from .invoice import Invoice
 from .onion import OnionPayload, TlvPayload, LegacyOnionPayload
 from .wire import LightningConnection, LightningServerSocket
-from .__version__ import __version__
+
+__version__ = "0.10.2"
```
vincenzopalazzo, line 7:
```suggestion
__version__ = "0.10.3"
```

The actual version on pypi is `0.10.2.post0` that should be greater than 0.10.2?

vincenzopalazzo (approved):
CI failure looks unrelated, also the protocol test (I will see into it)

ack https://github.com/ElementsProject/lightning/pull/5054/commits/136e617912174cbe7c8633d0948b134aa0b58e31


This PR will help a lot with all the env, great idea :)

In addition, when we have a standard version common between packaged I think I can finish the following PR https://github.com/rustyrussell/lnprototest/pull/34 in lnprototest because for now, I have a recursion problem during the resolution of one of the pyln.proto dependencies.

P.S This PR should override the following one right? https://github.com/ElementsProject/lightning/pull/5033

contrib/pyln-proto/pyln/proto/__init__.py
cdecker: We probably should just make it posts, since we increment the minor versions in lockstep with cln itself

vincenzopalazzo: just worried about the ability of poetry to resolve no number version, I don't see it in the docs https://python-poetry.org/docs/dependency-specification/ and also in the https://github.com/rustyrussell/lnprototest/pull/34 it cause a recursion problem just trying to resolve the `0.10.2.post0`.

However, I'm not sure what can be the real cause of the recursion here because I was not able to reproduce with other dependencies.

cdecker: The relative dependencies we use inside of this repo doesn't care about version numbers, it just takes the version that is in the repo. But I agree, we should verify that a remote user of the library can resolve things correctly. I'll see if I can port `lnprototest` to use poetry and use a `git`-base dependency so that we always develop against `master`

vincenzopalazzo: Another thing that I noted is the possibility that can cause the recursion is the pyln-proto dependece unfixed inside the pyln-bolt* package

I nothed that with the following poetry file poetry is able to resolve the dependencies
diff --git a/pyproject.toml b/pyproject.toml
index d23e41a..7927dd0 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,11 +10,11 @@ authors = [
```
 [tool.poetry.dependencies]
 python = "^3.8"
 pyln-proto = "^0.10.2"
-pyln-bolt1 = "^1.0.1"
-pyln-bolt2 = "^1.0.2"
-pyln-bolt4 = "^1.0.2"
-pyln-bolt7 = "^1.0.2"
-pyln-testing = "^0.9.3"
+#pyln-bolt1 = "^1.0.1"
+#pyln-bolt2 = "^1.0.2"
+#pyln-bolt4 = "^1.0.2"
+#pyln-bolt7 = "^1.0.2"
+#pyln-testing = "^0.9.3"
 coincurve = "13.0.0"
 python-bitcoinlib = "^0.10.0"
 mypy = '0.790'

```

But if I reintroduce one of the pyln-bolt* packages the recursion happens again.

cdecker: So far no luck, there seems to be a bug in `poetry` failing resolution (might be `git` based dependency + path dependencies in that deps). One easy fix is to either publish the `pyln` packages on pypi, or we checkout the lightning repo as a submodule in `lnprototest`

vincenzopalazzo: Yes, it is very strange how it resolves the dependencies in poetry, IMHO I think the option to pull it as a submodule can solve different problems for local development and with this solution, we can take time to resolve it as PyPI dependencies.

What do you think?

cdecker: Yeah, we'll have to fix this on the lnprototest side I think

contrib/pyln-spec/bolt7/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+packages = [
+  { include = "pyln/spec/bolt7" },
+]
+
+[tool.poetry.dependencies]
+python = "^3.7"
+pyln-proto = { path = "../../pyln-proto" }
+
+[tool.poetry.dev-dependencies]
+
```
vincenzopalazzo, line 17: maybe we can use the power of poetry dependencies to add also some dev dep `poetry add pytest --dev`?

contrib/pyln-testing/pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+authors = ["Christian Decker <decker.christian@gmail.com>"]
+license = "BSD-MIT"
+
+packages = [
+  { include = "pyln/testing" },
+]
+
+[tool.poetry.dependencies]
+python = "^3.7"
+pytest = "^7.0.1"
```
vincenzopalazzo, line 14: maybe dev dependencies for pytest is a better place?

cdecker: Hm, good question. This package is a bit special in that it's going to be a dev-dependency for others most likely.

The way I'm thinking about this is that the package is not functional without `pytest` so it's not a dev-dependency, but rather a normal dependency.

.github/workflows/ci.yaml
```diff
@@ -30,10 +30,7 @@ jobs:
...
         uses: actions/checkout@v2.0.0
 
-      - name: Fetch tags for auto versioning
-        run: git fetch --prune --unshallow --tags -f
-
-      - name: Set up Python 3.6
+      - name: Set up Python 3.7
         uses: actions/setup-python@v2
         with:
           python-version: 3.6
```
vincenzopalazzo:
```
 with:
          python-version: 3.7
```

contrib/pyln-spec/bolt2/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+packages = [
+  { include = "pyln/spec/bolt2" },
+]
+
+[tool.poetry.dependencies]
+python = "^3.7"
+pyln-proto = {path = "../../pyln-proto"}
+
+[tool.poetry.dev-dependencies]
+
```
vincenzopalazzo, line 17: we can add pytest as dev-dependencies also here

contrib/pyln-spec/bolt4/pyproject.toml
```diff
@@ -0,0 +1,20 @@
...
+
+packages = [
+  { include = "pyln/spec/bolt4" },
+]
+
+[tool.poetry.dependencies]
+python = "^3.7"
+pyln-proto = {path = "../../pyln-proto"}
+
+[tool.poetry.dev-dependencies]
```
vincenzopalazzo, line 16: Same here, and so on

pyproject.toml
```diff
@@ -10,14 +10,13 @@ python = "^3.7"
 pyln-client = { path = "contrib/pyln-client" }
 pyln-proto = { path = "contrib/pyln-proto" }
 Mako = "^1.1.6"
-mrkd = "^0.2.0"
+mrkd = { git = "https://github.com/refi64/mrkd.git", rev = "781f05eb9898ca652f18eed29b3c956389e6a2a7" }
```
vincenzopalazzo, line 13: lovelly!

vincenzopalazzo (commented):
LGTM,

I'm so happy to have poetry to manage the python package, I have just a couple of questions regarding the python packages as dev dependeces

contrib/pyln-testing/pyproject.toml
vincenzopalazzo: Sorry for the latency to add the comments here at different time.

>The way I'm thinking about this is that the package is not functional without pytest so it's not a dev-dependency, but rather a normal dependency.

I got it, and I think it is safe this put in the not dev dependencies because it help also to resolve the conflict in some configurations when an older version of `pytest` is used. Sorry if I miss this though during the review

rustyrussell (approved):
Ack 364e930d9b5d0e974a8ab8caee4735057887780b

contrib/pyln-spec/bolt7/pyproject.toml
vincenzopalazzo: Mh maybe this one https://github.com/ElementsProject/lightning/blob/master/contrib/pyln-spec/bolt7/tests/test_bolt7.py  or it is only a mock test?

PR #5056, opened 2022-02-25
pytest: Remove 3 stress-tests

cdecker:
These tests have proven to be:

 - very expensive, as they spin up many nodes, and perform long setup
 - are not testing anything specific, they just fuzz functionality
    that is already tested otherwise
 - have not helped pinpoint any issues in living memory
 - are very flaky, making for really bad signal-to-noise, so much
    that devs usually just restart without even looking at the logs
 - even if we were to look at the logs, we'd be unable to reproduce
    due to the inherent randomness involved in these tests
 - are really noisy neighbors, causing other tests to flake as well,
    further muddying the water

All in all, these tests are a waste of time, and source of
frustration.

Changelog-None

### DIFF
--- a/tests/test_closing.py
+++ b/tests/test_closing.py
@@ -187,68 +187,6 @@ def test_closing_id(node_factory):
     wait_for(lambda: not only_one(l2.rpc.listpeers(l1.info['id'])['peers'])['connected'])
 
 
-@pytest.mark.slow_test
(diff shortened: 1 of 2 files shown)


rustyrussell:
Trivial line removal to make flake8 happy...

Ack 5d570d1aad9ec0ab9b9f76c672a8fcc2283fef79

PR #5059, opened 2022-03-01
Goofy af tor.md systemd syntax update

GoofyAF:
Updated systemd syntax to prevent issues with initial tor systemd service activation and ensure that the systemd service is refreshed when the /etc/tor/torrc config is updated. 

Changelog-None

### DIFF
--- a/doc/TOR.md
+++ b/doc/TOR.md
@@ -9,7 +9,7 @@ If Tor is not installed you can install it on Debian based Linux systems (Ubuntu
 ```bash
 sudo apt install tor
 ```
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
utack

Can you squash the two commits in only one?

cdecker:
> utack
> 
> Can you squash the two commits in only one?

We can squash on merge, so not really necessary for these simple PRs.

ACK https://github.com/ElementsProject/lightning/pull/5059/commits/f019e76a6f2002bfd39e11332834ebbfdac7d918

rustyrussell:
Thanks, applied!

PR #5062, opened 2022-03-02
devtools: Add fund_nodes command to startup_regtest.sh

ddustin:
Also fix an occasional quirk of bitcoin-cli where you have to call loadwallet

### DIFF
--- a/contrib/startup_regtest.sh
+++ b/contrib/startup_regtest.sh
@@ -134,8 +134,11 @@ start_ln() {
 	# Kick it out of initialblockdownload if necessary
 	if bitcoin-cli -regtest getblockchaininfo | grep -q 'initialblockdownload.*true'; then
 		# Modern bitcoind needs createwallet
+		echo "Making \"default\" bitcoind wallet."
 		bitcoin-cli -regtest createwallet default >/dev/null 2>&1
 		bitcoin-cli -regtest generatetoaddress 1 "$(bitcoin-cli -regtest getnewaddress)" > /dev/null
+	else
+		bitcoin-cli -regtest loadwallet default
 	fi
 	alias bt-cli='bitcoin-cli -regtest'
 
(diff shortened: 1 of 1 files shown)


niftynei:
Test fails are coming from `make check-shellcheck` fyi!

Those errors need fixing, otherwise lgtm. Also added a few tweaks. Thanks!

niftynei (approved):
ACK cd346802b01204b6b35d0e5f84e43a3c21a2b3b8

contrib/startup_regtest.sh
```diff
@@ -153,39 +153,58 @@ start_ln() {
...
+
+	if [ -z "$2" ]; then
+		node2=2
+	else
+		node2="$2"
+	fi
+
 	ADDRESS=`bitcoin-cli -regtest getnewaddress`
+
+	echo "minning into address " $ADDRESS
```
rustyrussell: minning? :)

contrib/startup_regtest.sh
```diff
@@ -153,39 +153,58 @@ start_ln() {
...
 	bitcoin-cli -regtest generatetoaddress 50 $ADDRESS
 
-	echo $ADDRESS
+	echo "Mined into $ADDRESS, checking balance"
 	bitcoin-cli -regtest getbalance
 
-	L2_NODE_ID=`$LCLI --lightning-dir=/tmp/l2-regtest getinfo | jq -r .id`
-	L2_NODE_PORT=`$LCLI --lightning-dir=/tmp/l2-regtest getinfo | jq .binding[0].port`
+	L2_NODE_ID=`$LCLI --lightning-dir=/tmp/l$node2-regtest getinfo | jq -r .id`
+	L2_NODE_PORT=`$LCLI --lightning-dir=/tmp/l$node2-regtest getinfo | jq .binding[0].port`
```
rustyrussell: Prefer $LCLI -F here, so you can avoid requiring jq?

contrib/startup_regtest.sh
```diff
@@ -153,39 +153,58 @@ start_ln() {
...
+	else
+		node1="$1"
+	fi
+
+	if [ -z "$2" ]; then
+		node2=2
+	else
+		node2="$2"
+	fi
+
```
rustyrussell, line 167: I think we want a generic "fund_nodes" helper instead?  Which does all nodes if no args, otherwise the listed nodes?

contrib/startup_regtest.sh
```diff
@@ -169,7 +169,7 @@ fund_ln() {
 
 	echo "minning into address " $ADDRESS
 
-	bitcoin-cli -regtest generatetoaddress 50 $ADDRESS
+	bitcoin-cli -regtest generatetoaddress 125 $ADDRESS
```
rustyrussell: I think they always did?  100 is enough though.

contrib/startup_regtest.sh
```diff
@@ -197,6 +197,8 @@ fund_ln() {
 
 	bitcoin-cli -regtest generatetoaddress 24 $ADDRESS
 
+	sleep 6
+
```
rustyrussell, line 201: Spin and sleep on listfunds instead...

contrib/startup_regtest.sh
```diff
@@ -175,7 +175,7 @@ fund_ln() {
 
 	echo "minning into address " $ADDRESS
 
-	bitcoin-cli -regtest generatetoaddress 125 $ADDRESS
+	bitcoin-cli -regtest generatetoaddress 125 $ADDRESS > \dev\null
```
rustyrussell: `/` not `\`?

rustyrussell (changes requested):
Nice, but perhaps just add a fund_nodes instead (perhaps with --wallet=?)

And squash into a single commit :)

contrib/startup_regtest.sh
ddustin: What's the ideal way to parse the -F output -- is using sed appropriate here?

`L2_NODE_ID=$($LCLI -F --lightning-dir=/tmp/l$node2-regtest getinfo | sed -n 's/^id=\(.*\)/\1/p')`

eval almost works but it doesn't like the array notation.

```
$ eval $($LCLI -F --lightning-dir=/tmp/l$node2-regtest getinfo)
-bash: binding[0].type=ipv4: command not found
```

contrib/startup_regtest.sh
ddustin: 😳

contrib/startup_regtest.sh
ddustin: I went with using `sed`

contrib/startup_regtest.sh
ddustin: 100 it is!

contrib/startup_regtest.sh
ddustin: fund_nodes does all nodes or listed ones! Also looks for `--wallet=...`

rustyrussell (approved):

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/51ee61d518017eeeca1df5889f1bdde7457e1bdb

Issue #5037, opened 2022-02-12
Node announcement propagation difficulties after IP address change


whitslack:
### Issue and Steps to Reproduce
1. Have lots of active channels with nodes that you can't connect to. (They have to connect to you.)
2. Have your ISP change your IP address.
3. Have many channels force-closed on you because your peers can no longer find you in the network.

My network's public IPv4 address changed on Tuesday of this week. I immediately restarted my node and verified that the IPv4 address shown in the `address` list of its `getinfo` response was indeed my new address. Fast-forward to today (more than three days later), and I still have 516 channel peers that have not yet reconnected to me. I am connected with 636 channel peers, though over 300 of those reconnected immediately since I am able to connect outbound to them. This means that **less than half** of my channel peers that can only connect to me have done so **after more than three days** since my IP address changed. This is obviously not healthy. Furthermore, I have lost almost 40 channels to unilateral closures so far. This incurs a nontrivial monetary expense.

There appears to be a significant problem in node announcement gossip propagation across the Lightning Network. This may be due to a design flaw in the BOLT specifications, or there may be an implementation fault in C-Lightning. I will note that 1ML.com started showing my new IPv4 address practically immediately after the change, whereas Amboss.space took about a day to notice the new announcement, and LightningNetwork.plus *still* isn't showing my new address.

I usually run with `dev-fast-gossip` enabled, so I experimented with disabling that yesterday and also changed my node's RGB color to try to force a new node announcement to go out, but that achieved only very limited success in getting more of my channel peers to find me.

I am finding it impossible to discern from the debug logs how often my `node_announcement` is being transmitted to my connected peers, but it seems like it's not going out to enough of them.

### `getinfo` output
I'm on version 0.10.2, and my correct public IPv4 address is listed in the `address` array.

whitslack:
I re-enabled `dev-fast-gossip` in my config and restarted C-Lightning yet again, and now I have 879 connected and 276 disconnected channel peers, and LightningNetwork+ does now show the correct addresses for my node. I guess propagation of my node announcement is happening; it's just very slow.

whitslack:
I seem to have plateaued at around 900 connected channels and 250 disconnected channels. :frowning_face:

rustyrussell:
[...]

PR #5068, opened 2022-03-03
Gossip node_announcement more aggressively, allow 2 updates per day for gossip.

rustyrussell:
This should do something to mitigate the propagation issues seen in https://github.com/ElementsProject/lightning/issues/5037, as well as the problem reported that some LND nodes are scheduling daily downtime which causes 2 updates per day.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5068/commits/8e9e580f9237dbdf1931a0d46cb744cb3ab964c3

PR #5077, opened 2022-03-08
rust: remove the auto generated files from the git tree

vincenzopalazzo:
This is a good idea until we don't have rust as required dependencies because some PRs from a dev that has no Rust as developer dependencies can introduce an inconsistency in the rust wrapper.


vincenzopalazzo:
>The files do not depend on Rust at all, and can be generated independently. The CI will still ensure that the generated files are ok, so it'd be even less understandable for a user without Rust.

Mh, if the user has no rust installed the rust files are auto-generated? from this code https://github.com/ElementsProject/lightning/blob/0c4cc782df017eaca62306100fd269d31d8ee705/Makefile#L358-L361 I was assuming no, but for sure I'm missing somethings

vincenzopalazzo:
BTW, I just came to this conclusion because I ran the `make` on the master I receive the actual diff, and also in `model.rs` and` convert.rs`

```
diff --git a/cln-grpc/proto/node.proto b/cln-grpc/proto/node.proto
index 76610e9f6..7a59d8d6b 100644
--- a/cln-grpc/proto/node.proto
+++ b/cln-grpc/proto/node.proto
@@ -31,13 +31,20 @@ message GetinfoResponse {
        uint32 num_inactive_channels = 7;
        string version = 8;
        string lightning_dir = 9;
-       uint32 blockheight = 10;
-       string network = 11;
-       Amount fees_collected_msat = 12;
-       repeated GetinfoAddress address = 13;
-       repeated GetinfoBinding binding = 14;
-       optional string warning_bitcoind_sync = 15;
-       optional string warning_lightningd_sync = 16;
+       uint32 blockheight = 11;
+       string network = 12;
+       Amount fees_collected_msat = 13;
+       repeated GetinfoAddress address = 14;
+       repeated GetinfoBinding binding = 15;
+       optional string warning_bitcoind_sync = 16;
+       optional string warning_lightningd_sync = 17;
+}
+
+message GetinfoOur_features {
+       bytes init = 1;
+       bytes node = 2;
+       bytes channel = 3;
+       bytes invoice = 4;
 }
```

PR #5078, opened 2022-03-09
Multiple channel support.

rustyrussell:
May still need some love, but it Works For Me!


jb55:
+2,742 −3,313 excellent

lightningd/dual_open_control.c
```diff
@@ -3085,15 +3073,8 @@ static struct command_result *json_queryrates(struct command *cmd,
...
-	/* We can't query rates for a peer we have a channel with */
-	channel = peer_active_channel(peer);
-	if (channel)
-		return command_fail(cmd, LIGHTNINGD, "Peer in state %s,"
-				    " can't query peer's rates if already"
-				    " have a channel",
-				    channel_state_name(channel));
-
-	channel = peer_unsaved_channel(peer);
+	/* FIXME: This is wrong! */
```
cdecker, line 3063: I don't quite follow why this would be wrong.

lightningd/peer_control.c
```diff
@@ -1828,32 +1829,66 @@ static struct command_result *json_disconnect(struct command *cmd,
...
-			channel_fail_reconnect(channel,
-					       "disconnect command force=true");
-			goto wait_for_connectd;
-		}
-		return command_fail(cmd, LIGHTNINGD, "Peer is in state %s",
+
+	channel = peer_any_active_channel(peer, NULL);
+	if (channel && !*force) {
+		return command_fail(cmd, LIGHTNINGD,
+				    "Peer has channel in state %s",
```
cdecker, line 1840: `Peer has at least one channel`?

lightningd/peer_htlcs.c
```diff
@@ -621,11 +621,45 @@ static void forward_htlc(struct htlc_in *hin,
...
 
+	/* This is a shortcut for specifying next peer; doesn't mean
+	 * the actual channel! */
+	next = any_channel_by_scid(ld, scid);
+	if (next) {
+		struct peer *peer = next->peer;
+		struct channel *channel;
+		struct amount_msat best_spendable = channel_amount_spendable(next);
+
+		/* Seek largest capacity channel! */
```
cdecker, line 635: A bit of a nomenclature mixup I think: capacity to me always sounds like the total funds in the channel, whereas balance is the current allocation to either, though we don't have any rules around nomenclature.

lightningd/peer_htlcs.c
```diff
@@ -621,11 +621,45 @@ static void forward_htlc(struct htlc_in *hin,
...
+			struct amount_msat spendable;
+			if (!channel_active(channel))
+				continue;
+			spendable = channel_amount_spendable(channel);
+			if (!amount_msat_greater(spendable, best_spendable))
+				continue;
+
+			/* Don't override if fees differ... */
+			if (channel->feerate_base != next->feerate_base
+			    || channel->feerate_ppm != next->feerate_ppm)
```
cdecker, line 646: Could relax the equality to allow forwarding if the channel we're picking would end up cheaper than the one indicated by the scid.

cdecker (approved):
After all the pre-work this PR is great. Congrats pulling this one off. I tried my best, but could only find nits, kudos :+1: Now if only CI were as happy as I am xD

ACK https://github.com/ElementsProject/lightning/pull/5078/commits/346605822918d381072c72ecdc64e43fc18c00fc

lightningd/dual_open_control.c
rustyrussell: We really shouldn't be re-using an unsaved channel here: we should *always* create a new one I think.

lightningd/peer_htlcs.c
rustyrussell: Yes!  I wrote the comment then changed my mind when I wrote the code :)

lightningd/peer_htlcs.c
rustyrussell: Yes; generally though you can use a slight differentiator to remove this heuristic.  I'm not sure what the Right Thing is here, TBH.

rustyrussell:
... more flake fixes...

rustyrussell:
More flakes, and removed an overzealous BROKEN msg.

rustyrussell:
Rebased, and fixed (actually tested!) test cases.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5078/commits/3b1d5a0e3dd096a6ba29b6b05fa849e3d04d8544

jb55:
running this, opened more than one channel to a peer and nothing seems to have caught fire yet. nice!

rustyrussell:
Rebase on new master with setchannel support...

PR #5082, opened 2022-03-10
lnprototest: upgrade the lnprototest version

vincenzopalazzo:
Working today on c-lightning I noted some CI errors related to the lnprototest like [this one](https://github.com/ElementsProject/lightning/runs/5496009149?check_suite_focus=true#step:5:3440) or described in this issue https://github.com/rustyrussell/lnprototest/issues/31

I'm not sure that this will fix the error but it will introduce some change on how lnprototest manages the bitcoin backend, so I think it would be useful given a try.

P.S: Hoping to improve in the near future the upgrade process of lnprototest in the CI


vincenzopalazzo:
For some reason, we have the CI stack on the protocol test, but on the lnprototest repo, the integration testing takes only 12m 19s https://github.com/rustyrussell/lnprototest/runs/5477872454?check_suite_focus=true

rustyrussell:
@cdecker Seems to clash with the change to poetry?

vincenzopalazzo:
>Seems to clash with the change to poetry?

Mh! Seems more a problem related to test execution, in particular, related to the "send and receive" messages event.

With the actual fix on lnprototest https://github.com/rustyrussell/lnprototest/pull/38 we are a little bit slow (2 min) because now we manage correctly the backend and the restart process of all the lnprototest conf and restart it.

So, I will try to give another run to the CI with a rebase and see if I can catch the bug (if I introduce some things) in lnprototest, I don't think that it is related to this repo and to the poetry introduction

rustyrussell (approved):
Ack 00709028799b2eb03e82d143b26c7a52d11ea3c8

Issue #5073, opened 2022-03-05
hsm_secret password prompt appears twice at lightningd startup


mmxe:
### Issue and Steps to Reproduce
When starting c-lightning with --hsm-secret the password is required twice instead of once. Twice makes sense for setting the password to encrypt the secret but doesn't make UX sense when starting the lightningd service.

e.g.
--------
The hsm_secret is encrypted with a password. In order to decrypt it and start the node you must provide the password.
Enter hsm_secret password:
Confirm hsm_secret password:

### `getinfo` output

$ lightning-cli getinfo
{
   "id": "0xxx....xxxxd",
   "alias": "AliasX",
   "color": "222222",
   "num_peers": 0,
   "num_pending_channels": 0,
   "num_active_channels": 0,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv4",
         "address": "xxx.xxx.xxx.xxx",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "0.10.1",
   "blockheight": 726024,
   "network": "bitcoin",
   "msatoshi_fees_collected": 0,
   "fees_collected_msat": "0msat",
   "lightning-dir": "/home/bitcoin/.lightning/bitcoin"
}

Issue #4623, opened 2021-06-29
lightningd with `--encrypted-hsm` requires password confirmation for decryption


openoms:
This is a minor UX issue which for the fix is not entirely clear to me in the code so documenting here first.

* starting with an encryped hsm secret.

* `hsmtool decrypt hsm_secret`
only asks for the password once (as it should):

```
bitcoin@blitzvm:~$ /home/bitcoin/lightning/tools/hsmtool decrypt /home/bitcoin/.lightning/signet/hsm_secret
Enter hsm_secret password:
Succesfully decrypted hsm_secret, be careful now :-).
```
* `hsmtool encrypt hsm_secret`
asks for the password confirmation (as it should):
```
bitcoin@blitzvm:~$ /home/bitcoin/lightning/tools/hsmtool encrypt /home/bitcoin/.lightning/signet/hsm_secret
Enter hsm_secret password:
Confirm hsm_secret password:
Succesfully encrypted hsm_secret. You'll now have to pass the --encrypted-hsm startup option.
```

* The less logical step is when starting lightningd with:
`lightningd --encrypted-hsm`
it should not ask for the password to be confirmed for decryption, but it does:
```
bitcoin@blitzvm:~$ /usr/local/bin/lightningd --conf=/home/bitcoin/.lightning/sconfig --encrypted-hsm
The hsm_secret is encrypted with a password. In order to decrypt it and start the node you must provide the password.
Enter hsm_secret password:
Confirm hsm_secret password:
```

### `getinfo` output
```
{
   "id": "03821bc8e6bc11ad934799bd2e4b581614662cfcf1cb31d9d5edd44b6237a167ec",
   "alias": "ODDSET",
   "color": "03821b",
   "num_peers": 4,
   "num_pending_channels": 0,
   "num_active_channels": 4,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 39736
      },
      {
         "type": "torv3",
         "address": "4zwo6dj3qzgy2ltueads23egy23p5uyxsrhvh7awpkt27y6ug7arzcyd.onion",
         "port": 39736
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "127.0.0.1",
         "port": 39736
      }
   ],
   "version": "v0.10.0-398-g8167af5",
   "blockheight": 44408,
   "network": "signet",
   "msatoshi_fees_collected": 0,
[...]

PR #5085, opened 2022-03-10
cli: make the command line more user friendly.

vincenzopalazzo:
Fixes https://github.com/ElementsProject/lightning/issues/5073
Fixes https://github.com/ElementsProject/lightning/issues/4623

This is one of my first attempts to improve the command line UX when we are asking a password to decrypt an `hsm_secret` already encrypted.

In addition, this introduces the check between the password and password confirmation in the cmd option when the password confirmation is provided

### DIFF
--- a/common/hsm_encryption.c
+++ b/common/hsm_encryption.c
@@ -1,6 +1,8 @@
 #include "config.h"
+#include <ccan/tal/str/str.h>
 #include <common/errcode.h>
 #include <common/hsm_encryption.h>
+#include <errno.h>
 #include <termios.h>
 #include <unistd.h>
 
@@ -77,6 +79,17 @@ bool decrypt_hsm_secret(const struct secret *encryption_key,
 		return false;
 
 	return true;
+}
+
+/* Returns -1 on error (and sets errno), 0 if not encrypted, 1 if it is */
+int is_hsm_secret_encrypted(const char *path)
+{
+	struct stat st;
+
+        if (stat(path, &st) != 0)
+		return -1;
+
+        return st.st_size == ENCRYPTED_HSM_SECRET_LEN;
 }
 
 void discard_key(struct secret *key TAKES)
--- a/common/errcode.h
+++ b/common/errcode.h
@@ -14,5 +14,6 @@ typedef s32 errcode_t;
 #define HSM_ERROR_IS_ENCRYPT 21
 #define HSM_BAD_PASSWORD 22
 #define HSM_PASSWORD_INPUT_ERR 23
+#define ERROR_HSM_FILE 24
 
 #endif /* LIGHTNING_COMMON_ERRCODE_H */
--- a/common/hsm_encryption.h
+++ b/common/hsm_encryption.h
@@ -4,6 +4,7 @@
 #include <bitcoin/privkey.h>
 #include <ccan/tal/tal.h>
 #include <sodium.h>
+#include <sys/stat.h>
 
 /* Length of the encrypted hsm secret header. */
 #define HS_HEADER_LEN crypto_secretstream_xchacha20poly1305_HEADERBYTES
@@ -63,4 +64,6 @@ void discard_key(struct secret *key TAKES);
  */
 char *read_stdin_pass_with_exit_code(char **reason, int *exit_code);
 
+/** Returns -1 on error (and sets errno), 0 if not encrypted, 1 if it is */
+int is_hsm_secret_encrypted(const char *path);
 #endif /* LIGHTNING_COMMON_HSM_ENCRYPTION_H */
--- a/lightningd/hsm_control.c
+++ b/lightningd/hsm_control.c
@@ -96,11 +96,9 @@ struct ext_key *hsm_init(struct lightningd *ld)
 	 * not passed, don't let hsmd use the first 32 bytes of the cypher as the
 	 * actual secret. */
 	if (!ld->config.keypass) {
-		struct stat st;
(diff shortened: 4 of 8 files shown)


common/hsm_encryption.c
```diff
@@ -79,6 +79,24 @@ bool decrypt_hsm_secret(const struct secret *encryption_key,
...
 }
 
+bool hsm_secret_check_if_encrypted(const char *path, struct stat *hsm_stat,
+				   char **reason, int *exit_code)
+{
+	// We accept also NULL value, and for us means
+	// that we don't care about the stat data
+	/* FIXME(vincenzopalazzo): there is some way todo somethings like that? */
+	if (sizeof(hsm_stat) != sizeof(struct stat *))
+		hsm_stat = tal(NULL, struct stat);
```
vincenzopalazzo, line 89: There is some better whey to do this in C @rustyrussell?

common/hsm_encryption.c
```diff
@@ -79,6 +79,23 @@ bool decrypt_hsm_secret(const struct secret *encryption_key,
 	return true;
 }
 
+bool hsm_secret_check_if_encrypted(const char *path, struct stat *hsm_stat,
+				   char **reason, int *exit_code)
+{
+	// We accept also NULL value, and for us means
+	// that we don't care about the stat data
+	if (sizeof(hsm_stat) != sizeof(struct stat *))
```
rustyrussell: Yes, this doesn't work (it's always true, whatever the value of hsm_stat is).

rustyrussell: OK, a few things:

As Linus Torvalds says, "always return the error!".  And in this case, there's only one way it can fail, so the caller knows what the error is.  So:

```
/* Returns NULL if check is OK, an error msg if something went wrong */
char *hsm_secretfile_is_encrypted(const char *path, struct stat *hsm_stat, bool *is_encrypted);
```

But two more refinements:
1. No caller uses the hsm_stat argument, remove it.
2. The string here is not very useful, it needs to incorporate strerror, thus be dynamically allocated.

So:

```
/* Returns NULL if check is OK, an error msg (off ctx) if something went wrong */
char *hsm_secretfile_is_encrypted(const tal_t *ctx, const char *path, bool *is_encrypted)
{
	struct stat st;
	if (stat(path, &st) != 0)
		return tal_fmt(ctx, "Could not stat hsm_secret: %s", strerror(errno));

	*is_encrypted = (st.st_size == ENCRYPTED_HSM_SECRET_LEN);
	return NULL;
}
```

rustyrussell (changes requested):

common/hsm_encryption.c
vincenzopalazzo: Thank you for the code improvement I will add these changes :)

In addition, the only reason that I have used the `hsm_stat` is for a case like that!
```
       bool is_encrypted = hsm_secret_check_if_encrypted(hsm_secret_path, &st,
							  &err_msg, &err_code);
	if (!is_encrypted && st.st_size != 32)
		errx(ERROR_HSM_FILE, "Invalid hsm_secret (neither plaintext "
				     "nor encrypted).");
```

I'm always scared to touch the hsm code and I don't know if we need to include these cases too.

A solution that I have in mind is implementing some things like that

```
/* Returns NULL if check is OK, an error msg (off ctx) if something went wrong */
char *hsm_secretfile_is_encrypted_with_stat(const tal_t *ctx, struct stat *hsm_stat, const char *path, bool *is_encrypted)
{
	*is_encrypted = (hsm_stat.st_size == ENCRYPTED_HSM_SECRET_LEN);
	return NULL;
}

/* Returns NULL if check is OK, an error msg (off ctx) if something went wrong */
char *hsm_secretfile_is_encrypted(const tal_t *ctx, const char *path, bool *is_encrypted)
{
	struct stat st;
	if (stat(path, &st) != 0)
		return tal_fmt(ctx, "Could not stat hsm_secret: %s", strerror(errno));

	*is_encrypted = (st.st_size == ENCRYPTED_HSM_SECRET_LEN);
	return hsm_secretfile_is_encrypted_with_stat(ctx, st, path, is_encypted);
}
```

do you think that this case are not useful? I can safely remove it?

rustyrussell:
Hmm, ok, the hsmtool is a special case.  Let's make a simpler function, and have the hsmtool do more work.  Here's my attempt:

```patch

	char *err_msg;
	int err_code;
	bool is_encrypted;

	err_msg = hsm_secret_check_if_encrypted(tmpctx, hsm_secret_path, &st,
						&is_encrypted);

	if (err_msg)
		errx(err_code, "%s", err_msg);

	if (!is_encrypted && st.st_size != 32)
		errx(ERROR_HSM_FILE, "Invalid hsm_secret (neither plaintext "
				     "nor encrypted).");

	return is_encrypted;
}
```

vincenzopalazzo:
ok I make the changes, thanks!

rustyrussell:
OK, I squashed this into two commits, otherwise it won't bisect and looks weird.

I also fixed up Changelog-Fix: to Changelog-Fixed:

rustyrussell (approved):
Ack 52ede8bfe325ad76d24712bc3fba4f8fc7d7a9b3

rustyrussell:
Fixed random unrelated test flake

Ack https://github.com/ElementsProject/lightning/pull/5085/commits/13bcf268dd49c18e935a5f0f89da878bb64f4f7b

vincenzopalazzo:
>Fixed random unrelated test flake

Good :) thanks!

PR #5095, opened 2022-03-14
tests: memleak catching a plugin that's still grinding test_sendinvoice

niftynei:
We shutdown before the offers plugin infra has finished grinding for a
shadow route, which results in a memleak error.

If we wait to end the test until the node has finished the route calculations
etc, we can avoid running memleak while it's in the middle of something.

Found in CI run of #2649.

Changelog-None

### DIFF
--- a/tests/test_pay.py
+++ b/tests/test_pay.py
@@ -4710,6 +4710,7 @@ def do_test_sendinvoice(node_factory, bitcoind, disable):
     assert 'pay_index' in out
     assert out['msatoshi_received'] == 100000000
     assert out['amount_received_msat'] == Millisatoshi(100000000)
+    l1.daemon.wait_for_log('Payed out 100000000msat for offer')
 
     # Note, if we're slow, this fails with "Offer no longer available",
     # *but* if it hasn't heard about payment success yet, l2 will fail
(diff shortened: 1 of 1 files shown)


cdecker:
ACK ffbb36ab429de50e57b56ef68f44e171f4699b84

rustyrussell:
The commit message is insufficient :(

You need to at least include the backtrace so I can tell what you're fixing!  And so others can google it...

And I can't reproduce it.

niftynei:
There's a backtrace in the commit message with the memleak traces?

https://github.com/ElementsProject/lightning/pull/5095/commits/9e314cd53d3f7fa8f6a84a92e6196fe4d1b47a0c

The logs were pretty big, but my understanding is that the `pay` plugin was in the middle of a shadow_route calculation when the test cleanup happened, which resulted in a flake due to memleak catching still allocated memory in the plugin. By waiting til the shadow_route calculation terminates to cleanup, we avoid the memleak flake.

Since we fixed the memleak detection, I've been seeing a decent number more flakes due to non-terminated RPC calls handlers in plugins being caught in the memleak sweep. This is one of those.

rustyrussell:
I think this weird leak is solved by https://github.com/ElementsProject/lightning/pull/5100

PR #5097, opened 2022-03-15
py: Disentangle pyln dependencies

cdecker:
Turns out that the pyln-proto dependency in the bolt packages is only
needed for testing, not for production. Making it a dev-dependency
means it isn't considered in resolution anymore.

Since the bolt, testing and client packages are to be used outside
from the project we can't use relative dependencies either, so make
then dependent on the version on PyPI. This also means we had to push
a couple of updated to PyPI.

Changelog-None


rustyrussell (approved):
Ack 3f7d451b70b96ba4c37528de8628a5087164fb8a

PR #5098, opened 2022-03-15
pyln: Change the default argument to `exclude` in `pay`

cdecker:
We must use `None` for default arguments since otherwise they aren't
filtered out when serializing the request. In addition default
arguments to functions are initialized once and, if mutable, could
persist internal changes across function calls.

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5098/commits/3befae4553b60d760bc6ccbb8e0caf7df3e5ea84

Good catch!

PR #5099, opened 2022-03-15
cln-grpc: Keep the field numbering constant while the schemas evolve

cdecker:
Protocol Buffers and GRPC are very sensitive to its field numbering changing. With this we add a metadata file that tracks already assigned numbers so we can re-assign them correctly despite fields being added and removed.


rustyrussell:
Ack b13fccb89d94ee2d54b8ae40994cb49487df90a9

PR #5103, opened 2022-03-16
setchannel: new command

rustyrussell:


tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
     # get short channel id
     scid = l1.get_channel_scid(l2)
 
     # feerates should be init with global config
-    db_fees = l1.db_query('SELECT feerate_base, feerate_ppm FROM channels;')
+    db_fees = l1.db_query('SELECT feerate_base, feerate_ppm, htlc_maximum_msat FROM channels;')
     assert(db_fees[0]['feerate_base'] == DEF_BASE)
     assert(db_fees[0]['feerate_ppm'] == DEF_PPM)
+    # This will be the capacity - reserves:
+    assert(db_fees[0]['htlc_maximum_msat'] == MAX_HTLC)
```
m-schmoock, line 1897: This line fails.

First issue here is we compare a number int (msat) with msat value (BUT shifted by 1000, since we did `Millisatoshi(int(FUNDAMOUNT * 0.99)))` which should be `Millisatoshi(int(FUNDAMOUNT * 1000 * 0.99)))`. To prevent that we may change FUNDAMOUNT to FUNDAMOUNT_SAT everywhere. This seems also messed up in other locations and tests in this file: `test_setchannel_routing`, `test_setchannel_restart`, ... search for `FUNDAMOUNT`, don't know why they not fail or why we haven't seen earlier.

Second issue: The DB thinks we can send 99% of the channel capacity (which seems correct to me). Where we test for 98%. Since this is a pristine new channel, with l2 having 0msat, we should be able to send 99% just so we keep our own reserve filled. If l2 decides later to send amounts back he can then only do 98% of the FUNDAMOUNT.

```py
>       assert(db_fees[0]['htlc_maximum_msat'] == MAX_HTLC)
E       assert 990000000 == 980000msat
E         +990000000
E         -980000msat
```

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
+    result = l1.rpc.setchannel(ppm=42, id=scid)
     assert(len(result['channels']) == 1)
     assert(re.match('^[0-9a-f]{64}$', result['channels'][0]['channel_id']))
     assert(result['channels'][0]['short_channel_id'] == scid)
-    result = l1.rpc.setchannelfee(base=42, id=scid)
-    assert(result['base'] == 42)
-    assert(result['ppm'] == DEF_PPM)
+    assert(result['channels'][0]['fee_base_msat'] == 1337)
+    assert(result['channels'][0]['fee_proportional_millionths'] == 42)
+    assert(result['channels'][0]['maximum_htlc_out_msat'] == 50000)
```
m-schmoock, line 1939: DUP: Fails, as you set it to `133337` (msat) before in line 1905.

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
     # also check for updated values in `listpeers`
     peers = l1.rpc.listpeers()['peers']
     assert peers[0]['channels'][0]['fee_base_msat'] == Millisatoshi(1337)
     assert peers[0]['channels'][0]['fee_proportional_millionths'] == 137
+    assert peers[0]['channels'][0]['maximum_htlc_out_msat'] == 133337
 
     # wait for gossip and check if l1 sees new fees in listchannels
     wait_for(lambda: [c['base_fee_millisatoshi'] for c in l1.rpc.listchannels(scid)['channels']] == [DEF_BASE, 1337])
     wait_for(lambda: [c['fee_per_millionth'] for c in l1.rpc.listchannels(scid)['channels']] == [DEF_PPM, 137])
+    wait_for(lambda: [c['htlc_maximum_msat'] for c in l1.rpc.listchannels(scid)['channels']] == [MAX_HTLC, 133337])
```
m-schmoock, line 1930: This also fails. `listchannels` currently spits out `990000000msat` for the other direction. Not sure whats correct here since `listchannels` is a gossip function, maybe it should or should not know about counterparts reserve state.

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
     # check result format
-    assert(result['base'] == 1337)
-    assert(result['ppm'] == 137)
     assert(len(result['channels']) == 1)
     assert(re.match('^[0-9a-f]{64}$', result['channels'][0]['channel_id']))
     assert(result['channels'][0]['peer_id'] == l2.info['id'])
     assert(result['channels'][0]['short_channel_id'] == scid)
+    assert(result['channels'][0]['fee_base_msat'] == 1337)
+    assert(result['channels'][0]['fee_proportional_millionths'] == 137)
+    assert(result['channels'][0]['maximum_htlc_out_msat'] == 50000)
```
m-schmoock, line 1914: Fails, as you set it to 133337 (msat) before in line 1905.

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
+    assert(result['channels'][0]['fee_proportional_millionths'] == 42)
+    assert(result['channels'][0]['maximum_htlc_out_msat'] == 50000)
+
+    result = l1.rpc.setchannel(base=43, id=scid)
+    assert(len(result['channels']) == 1)
+    assert(re.match('^[0-9a-f]{64}$', result['channels'][0]['channel_id']))
+    assert(result['channels'][0]['short_channel_id'] == scid)
+    assert(result['channels'][0]['fee_base_msat'] == 43)
+    assert(result['channels'][0]['fee_proportional_millionths'] == 42)
+    assert(result['channels'][0]['maximum_htlc_out_msat'] == 50000)
```
m-schmoock, line 1947: DUP: Fails, as you set it to 133337 (msat) before in line 1905.

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
     # wait for gossip and check if l1 sees new fees in listchannels
     wait_for(lambda: [c['base_fee_millisatoshi'] for c in l1.rpc.listchannels(scid)['channels']] == [DEF_BASE, 1337])
     wait_for(lambda: [c['fee_per_millionth'] for c in l1.rpc.listchannels(scid)['channels']] == [DEF_PPM, 137])
+    wait_for(lambda: [c['htlc_maximum_msat'] for c in l1.rpc.listchannels(scid)['channels']] == [MAX_HTLC, 133337])
 
     # also test with named and missing parameters
-    result = l1.rpc.setchannelfee(ppm=42, id=scid)
-    assert(result['base'] == DEF_BASE)
-    assert(result['ppm'] == 42)
+    result = l1.rpc.setchannel(ppm=42, id=scid)
```
m-schmoock, line 1933: Fails with `got an unexpected keyword argument 'ppm'`. Should be:
```py
    result = l1.rpc.setchannel(feeppm=42, id=scid)
```

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
     assert(re.match('^[0-9a-f]{64}$', result['channels'][0]['channel_id']))
     assert(result['channels'][0]['short_channel_id'] == scid)
-    result = l1.rpc.setchannelfee(base=42, id=scid)
-    assert(result['base'] == 42)
-    assert(result['ppm'] == DEF_PPM)
+    assert(result['channels'][0]['fee_base_msat'] == 1337)
+    assert(result['channels'][0]['fee_proportional_millionths'] == 42)
+    assert(result['channels'][0]['maximum_htlc_out_msat'] == 50000)
+
+    result = l1.rpc.setchannel(base=43, id=scid)
```
m-schmoock, line 1941: Fails with `got an unexpected keyword argument 'base'`. Should be:
```py
    result = l1.rpc.setchannel(feebase=42, id=scid)
```

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
-        l1.rpc.setchannelfee(scid, -1, -1)
+        l1.rpc.setchannel(scid, -1, -1)
 
     # test if zero fees is possible
-    result = l1.rpc.setchannelfee(scid, 0, 0)
-    assert(result['base'] == 0)
-    assert(result['ppm'] == 0)
-    db_fees = channel_get_fees(scid)
+    result = l1.rpc.setchannel(scid, 0, 0)
+    assert(result['channels'][0]['short_channel_id'] == 0)
```
m-schmoock, line 1964: Fails. Scid is not 0 but resolved to the value of `l1.get_channel_scid(l2)`

tests/test_pay.py
```diff
@@ -1873,131 +1873,144 @@ def test_setchannelfee_usage(node_factory, bitcoind):
...
+    assert(result['channels'][0]['fee_base_msat'] == 0)
+    assert(result['channels'][0]['fee_proportional_millionths'] == 42)
```
m-schmoock, line 1966: Fails, fee_proportional_millionths is 0 as set earlier in line 1963.

m-schmoock (commented):
Nice. We should also add `htlc_minimum_msat` with this PR.

channeld/channeld.c
```diff
@@ -119,8 +119,12 @@ struct peer {
 
 	/* CLTV delta to announce to peers */
 	u16 cltv_delta;
+
+	/* We only really know these because we're the ones who create
+	 * the channel_updates. */
 	u32 fee_base;
 	u32 fee_per_satoshi;
+	struct amount_msat htlc_maximum_msat;
```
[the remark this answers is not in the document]
rustyrussell: > Nice. We should also add `htlc_minimum_msat` with this PR.

Confusingly, there's a min we have agreed with our peer at channel creation, and we get very upset if they try to offer an HTLC less than that.  So the minimum case is weird; we'd have to have a new minimum which is not the channel config minimum.

So I backed away, for now...

tests/test_pay.py
rustyrussell: Oh yes, I pushed this up early.  I've reworked the tests now.

channeld/channeld.c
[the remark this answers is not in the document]
m-schmoock: kk lets do this in another PR

tests/test_pay.py
```diff
@@ -2074,62 +2089,61 @@ def test_setchannelfee_routing(node_factory, bitcoind):
...
-    assert route[0]['msatoshi'] == 5002020
-    assert route[1]['msatoshi'] == 4999999
 
+    # FIXME!
+#    # Should refuse to route this!
+#    with pytest.raises(RpcError, match=r'Could not find a route'):
+#        l1.rpc.getroute(l3.info['id'], 4000001, 1, fuzzpercent=0)["route"]
+
+    # 1337 + 4000000 * 137 / 1000000 = 1885
+    route_ok = l1.rpc.getroute(l3.info['id'], 4000000, 1)["route"]
```
m-schmoock, line 2112: Just so we don't forget it, if you go for `4000000` in this test also update the comment line  2102:
`If l1 were to send 4,999,999 millisatoshi to l3 via l2`

m-schmoock: Maybe also go for 3,999,999 to provoke potential rounding errors

rustyrussell:
OK, large rework of the tests, which fixes them up, and fixes for dual-funding.

tests/test_pay.py
```diff
@@ -2062,73 +2091,119 @@ def test_setchannelfee_routing(node_factory, bitcoind):
...
     wait_for(lambda: [c['fee_per_millionth'] for c in l1.rpc.listchannels(scid)['channels']] == [137, DEF_PPM])
+    wait_for(lambda: [c['htlc_minimum_msat'] for c in l1.rpc.listchannels(scid)['channels']] == [17, MIN_HTLC])
+    wait_for(lambda: [c['htlc_maximum_msat'] for c in l1.rpc.listchannels(scid)['channels']] == [4000000, MAX_HTLC])
 
     # test fees are applied to HTLC forwards
     #
     # BOLT #7:
     # If l1 were to send 4,999,999 millisatoshi to l3 via l2, it needs to
     # pay l2 the fee it specified in the l2->l3 `channel_update`, calculated as
     # per [HTLC Fees](#htlc_fees):  base + amt * pm / 10**6
```
m-schmoock, line 2122: This text still needs update, as the 'Note: ' below mentions we are not testing for 4,999,999msat

m-schmoock:
ACK 7db4743

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5103/commits/148a0d2aa7f5a06cb7d49c4bb997ff45adc714df

Fixing an unrelated leak report which cropped up...

m-schmoock:
@rustyrussell 
I'm working on a follow up PR that will make  `htlc_minimum_msat`  and `htlc_minimum_msat` defaults configurable for newly created channels. It will also contain a 'turn off able' switch for the IP discovery feature.

Whats the timeline for the next release? I think we should bring that in before...

PR #5108, opened 2022-03-17
hsmd: Add validate_commitment_tx

ksedgwic:
Add hsmd_validate_commitment_tx message so validating signers can independently validate the peer's signatures.

[Summary of all changes for validating signers](https://gitlab.com/lightning-signer/docs/-/wikis/C-Lightning-Modifications)

This PR is the last of the backlog of changes required to run the Validating Lightning Signer with c-lightning!


ksedgwic:
kicked CI w/ trivial change

rustyrussell:
Thanks, found another random false positive leak :)  I'm tracking it now...

rustyrussell:
Note: we're freezing for release, but I'm going to add this to the milestone anyway.

cdecker:
Just out of curiosity, will a verifying signer (`lightning-signer`) stall if the node doesn't provide these messages, i.e., will it refuse to proceed with the protocol if it doesn't get these messages? I'm a bit worried about compatibility between node and signer not being updated in lock step.

ACK https://github.com/ElementsProject/lightning/pull/5108/commits/f5888c487fb2c5d39d7fa1e38729b152b830a217

cdecker (approved):

cdecker:
Fwiw, @bitcoin-bot is having issues with renamed repos (the fork was renamed to `c-lightning`). Need to look into this

ksedgwic:
> Just out of curiosity, will a verifying signer (`lightning-signer`) stall if the node doesn't provide these messages, i.e., will it refuse to proceed with the protocol if it doesn't get these messages? I'm a bit worried about compatibility between node and signer not being updated in lock step.

Yes, the signer will not proceed in the protocol if these messages are missing.

There is a set of "foundational" messages which must be present, this is one.  On the other hand, implementation of new policy controls on the signer side would not stop the protocol from advancing (unless the specific new policies were violated)

PR #5110, opened 2022-03-18
More leaking, better diags

rustyrussell:
Found by CI for #5108, but reproduced locally.


cdecker (approved):
ACK 9d5ed19947e005399a490d7fcc480e29c21b4fea

rustyrussell:
Rebased, and fixed crash when we call tal_wally_end_onto() with NULL.

PR #5120, opened 2022-03-23
lightningd: allow 'style' 'legacy' for clboss compat (with deprecated_apis)

rustyrussell:
We keep the parameter around for future use (e.g. PTLC support?), but clboss
master still sets 'style' to "legacy".  Accept, but ignore it.

Reported-by: grubman on #c-lightning (IRC)
Changelog-None: change introduced this release

### DIFF
--- a/lightningd/pay.c
+++ b/lightningd/pay.c
@@ -1,6 +1,7 @@
 #include "config.h"
 #include <ccan/tal/str/str.h>
 #include <common/bolt12_merkle.h>
+#include <common/configdir.h>
 #include <common/json_command.h>
 #include <common/json_helpers.h>
 #include <common/json_tok.h>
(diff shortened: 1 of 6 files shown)


lightningd/pay.c
```diff
@@ -1314,6 +1315,11 @@ static struct command_result *param_route_hop_style(struct command *cmd,
 		return NULL;
 	}
 
+	/* We still let you *specify* this, but we ignore it! */
+	if (deprecated_apis && json_tok_streq(buffer, tok, "legacy")) {
+		return NULL;
+	}
+
```
vincenzopalazzo, line 1322:
```suggestion
	if (deprecated_apis && json_tok_streq(buffer, tok, "legacy"))
		return NULL;

```

vincenzopalazzo (changes requested):
LGTM, only some minor in the style convention.

In addition, is it better to add in this case a `Changelog-Deprecated`?

rustyrussell:
> LGTM, only some minor in the style convention.
> 
> In addition, is it better to add in this case a `Changelog-Deprecated`?

We already documented that we removed legacy, but you're right.

vincenzopalazzo:
>We already documented that we removed legacy, but you're right.

I put on my to-do list that may be necessary to remove this remove legacy line in the Changelog generation.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5120/commits/c4f84a17aec1f66b6d25f6d7f1a358842046a973

PR #5121, opened 2022-03-23
invoice description hash support

rustyrussell:


doc/lightning-invoice.7.md
```diff
@@ -29,8 +29,9 @@ of this invoice.
 
 The *description* is a short description of purpose of payment, e.g. *1
 cup of coffee*. This value is encoded into the BOLT11 invoice and is
-viewable by any node you send this invoice to. It must be UTF-8, and
-cannot use *\\u* JSON escape codes.
+viewable by any node you send this invoice to (unless *deschash* is
```
erikarvstedt, line 32:
```suggestion
viewable by any node you send this invoice to (unless *deschashonly* is 
```

doc/lightning-invoice.7.md
```diff
@@ -68,6 +69,10 @@ payment.
 If specified, *cltv* sets the *min_final_cltv_expiry* for the invoice.
 Otherwise, it's set to the parameter **cltv-final**.
 
+If *deschash* is true (default false), then a hash of the *description*,
```
erikarvstedt, line 72:
```suggestion
If *deschashonly* is true (default false), then a hash of the *description*,
```
Also, the main clause, starting with `then`, is incomplete (probably missing verb and subject).

erikarvstedt (changes requested):
No thorough review, these are just some doc errors that stuck out.

As stated before, I'd prefer if there was an option for hash-only storage.
clightning should be unintrusive, efficient base layer tech. Forcing invoice creators to store descriptions and then deleting them in a separate command is inconvenient and slow.

jb55:
Concept ACK, going to test this with https://github.com/jb55/lnurl-commando

jb55:
I was using the description field for bolt11 "payer notes" on jb55.com, it looks like to properly implement lnurl to spec I would have to drop that feature since a description hash would subsume the description. It seems like the ideal scenario would be to extend the lnurl spec to support merkalized entries in the metadata description string? this complicates it a bit but would allow me to parse out payer notes from the description. this probably won't happen so I guess I'll just drop the lnurl/bolt11 "payer note" feature and stick to bolt12 for that.

jb55:
If we now have a `deschashonly` option why not allow just passing a description hash? Otherwise when generating invoices remotely I have to send the full png over commando every time. not a big deal but it's slightly annoying...

jb55:
> If we now have a `deschashonly` option why not allow just passing a description hash? Otherwise when generating invoices remotely I have to send the full png over commando every time. not a big deal but it's slightly annoying...

ah I see #4892 implements it this way already 🤔

rustyrussell:
> No thorough review, these are just some doc errors that stuck out.
> 
> As stated before, I'd prefer if there was an option for hash-only storage. clightning should be unintrusive, efficient base layer tech. Forcing invoice creators to store descriptions and then deleting them in a separate command is inconvenient and slow.

Then... don't delete them?  If you're regularly issuing invoices you should be deleting expired ones, and if you have thousands of successful payments you can afford a few GB of storage.

doc/lightning-invoice.7.md
rustyrussell: Thanks!  Obviously I didn't finish that thought!

rustyrussell:
To be clear: I really want to see what invoices I have issued by my node.  I suspect that the ability to mint invoices with only a description hash will get abused by plugins pretty badly, and disk space is cheap.

lightningd/invoice.c
```diff
@@ -1207,7 +1209,18 @@ static struct command_result *json_invoice(struct command *cmd,
...
+	 *...
+	 * A writer:
+	 *...
+	 *    - MUST include either exactly one `d` or exactly one `h` field.
+	 */
+	if (*hashonly) {
+		info->b11->description_hash = tal(info->b11, struct sha256);
+		sha256(info->b11->description_hash, desc_val, strlen(desc_val));
+	} else
+		info->b11->description_hash = NULL;
```
vincenzopalazzo, line 1223: Why not just 
```suggestion
         info->b11->description_hash = NULL;
	/* BOLT #11:
	 * * `h` (23): `data_length` 52. 256-bit description of purpose of payment (SHA256).
	 *...
	 * A writer:
	 *...
	 *    - MUST include either exactly one `d` or exactly one `h` field.
	 */
	if (*hashonly) {
		info->b11->description_hash = tal(info->b11, struct sha256);
		sha256(info->b11->description_hash, desc_val, strlen(desc_val));
	}
		
```

There is any convention or particular code style? Just curius

wallet/invoices.c
```diff
@@ -648,9 +665,9 @@ void invoices_waitone(const tal_t *ctx,
 			   false, invoice.id, cb, cbarg);
 }
 
-const struct invoice_details *invoices_get_details(const tal_t *ctx,
-						   struct invoices *invoices,
-						   struct invoice invoice)
+struct invoice_details *invoices_get_details(const tal_t *ctx,
+					     struct invoices *invoices,
+					     struct invoice invoice)
```
vincenzopalazzo, line 670:
```suggestion
					     const struct invoices *invoices,
					     const struct invoice invoice)
```

wallet/invoices.h
```diff
@@ -213,8 +224,8 @@ void invoices_waitone(const tal_t *ctx,
  * @invoice - the invoice to get details on.
  * @return pointer to the invoice details allocated off of `ctx`.
  */
-const struct invoice_details *invoices_get_details(const tal_t *ctx,
-						   struct invoices *invoices,
-						   struct invoice invoice);
+struct invoice_details *invoices_get_details(const tal_t *ctx,
+					     struct invoices *invoices,
+					     struct invoice invoice);
```
vincenzopalazzo, line 229:
```suggestion
					     const struct invoices *invoices,
					     const struct invoice invoice);
```

vincenzopalazzo (commented):
LGTM, the only comment that I have is that some of these procedure parameters may be declared constant!

lightningd/invoice.c
rustyrussell: Yes.  Generally I prefer to avoid the pattern of

```
x = foo
if (cond)
    x = bar
```

Because if you forget something, the compiler won't warn you.  For example:

```
if (cond)
    x = bar
else
   y = foo //  Oops, was meant to be x!

print(x) // Compiler warns here that x may not be initialized!
```

wallet/invoices.h
rustyrussell: The 'struct invoice' is passed by value, so we don't generally const those.  And the invoices is not usually passed const to these functions, though it would be legal: we do change an invoice.

niftynei:
Permitting a node to never be fully aware of the purpose of a payment has some bookkeeping and litigation risks. The tradeoff here is that you, as the node operator, by only issuing invoices for description hashes retain a certain amount of protection in that you never knew what the payment was for, if asked, and can't produce the data if subpoenaed.

Well you can produce the hashes. If it turns out the description that matches the hash was for illicit goods that is a real risk of having accepted payment for something illegal... but you can claim lack of knowledge (or any way of knowing) what the payment originally represented. No idea how useful any of this is for any potential situation where it might be brought to bear, this is not legal advice.

Not having descriptions for payments will also make bookkeeping extremely onerous, for those who do need notes classifying payments they've received.

@rustyrussell can you outline what abuse you have in mind here?

> I suspect that the ability to mint invoices with only a description hash will get abused by plugins pretty badly,

lightningd/invoice.c
vincenzopalazzo: Cool, thanks!

wallet/invoices.h
vincenzopalazzo: ops, just noted now that invoice is a copy and we just don't care.

cdecker:
I have to say that I agree with having the description be part of the invoice data is the correct thing to do here:

 - Consistency: being able to provide the description without referring to an external system is very valuable, especially if those external systems may end up malleating the description (image compression, PDF OCR modifications, editing descriptions) which would unlink the hash from the doc
 - Completeness and auditability: as @niftynei points out, there is a point in making sure that the information is as self-contained as possible. Plugins can rely on the information being present, and it reduces the variations that have to be dealt with.

I think ultimately, the fact that the description string can be used as an escape hatch by just including a URL and a hash should serve the use-cases where the description is managed externally.

jb55:
all good justifications and I see that the description is stored even when `deschashonly` is set which makes sense. I still don't want to permanently store a copy of my lnaddress thumbnail for every invoice requested in an lnaddress payment, so perhaps I should just remove my png to become spec compliant or keep it and become not lnaddress spec compliant. I'm leaning toward the latter since lnlink skips the check anyways...

PR #5122, opened 2022-03-24
Pay enhancements: remove legacy, add maxfee and description params

rustyrussell:
~This builds on #5121 (first three commits).~ merged, rebased

That PR made me realize we violate bolt11 when paying a description_hash invoice, so add a description parameter and deprecate payment of such an invoice without it, and I want to do that this release if possible.

However, there was some more work I wanted to do on the way...


plugins/libplugin-pay.c
```diff
@@ -3559,6 +3563,8 @@ static void presplit_cb(struct presplit_mod_data *d, struct payment *p)
 			 * they'll be used when aggregating the payments
 			 * again. */
 			c->invstring = tal_strdup(c, p->invstring);
+			if (p->description)
+				c->description = tal_strdup(c, p->description);
```
cdecker, line 3567: This is really bad, as it replicates the description in all child-payments, bloating the database with duplicate values. I thought we did the sane thing, and didn't copy the `invstring` either, as that is constant in the entire tree, but apparently it got changed somewhere along the line.

There is no point in duplicating these values, and we can use the convention of attaching them only to the root payment.

doc/lightning-pay.7.md
```diff
@@ -43,6 +43,11 @@ creates an absolute limit on what fee we will pay.  This allows you to
 implement your own heuristics rather than the primitive ones used
 here.
 
+*description* is only required for bolt11 invoices which do not
+contain a description themselves, but contain a description hash.
+*description* is then checked against the hash inside the invoice
+before it will be paid.
```
cdecker, line 49: This requirement is incredibly restrictive: it limits us to invoices with `description_hash` where the `description` itself is representable as a raw JSON value. Anything like a PDF, which is something that people have been asking for, is essentially impossible to do with this, unless we allows a hex-encoded value to be provided instead, which makes this cumbersome to use. I know I voted in favor of requiring the `description` but this version is too limiting imho

cdecker (changes requested):
The requirement for `description` to be provided as a raw JSON value is too much of a restriction, and copying a potentially large description around to every single `sendpay` is a horrible waste of bytes (as well as copying the `invstring`, but that somehow sneaked in while I wasn't looking).

doc/lightning-pay.7.md
rustyrussell: Yes, the limits of json!

But we can add descriptionhex later for this case?

plugins/libplugin-pay.c
rustyrussell: Hmm, fixing this is a significant change.  Let me try... (and add a test!).  I copied the invstring, assuming that was doing The Right Thing.

doc/lightning-pay.7.md
rustyrussell: I looked at this, we need a non-trivial amount of re-engineering to add this.  We cannot handle bolt11 descriptions which are not basically valid strings right now, and changing that would be an API change everywhere we print description.

So I'm leaving that for next commit.

plugins/libplugin-pay.c
rustyrussell: Actually, this was all a mess!  This duplicate is only for the pre-splitter, but it turns out we were often returning listpays() without *any* bolt11.

I've fixed that.

rustyrussell:
(Fixed keysend with the new invstring deduplication: it can still be NULL in this case!

doc/lightning-pay.7.md
```diff
@@ -5,8 +5,8 @@ SYNOPSIS
 --------
 
 **pay** *bolt11* [*msatoshi*] [*label*] [*riskfactor*]
-[*maxfeepercent*] [*retry\_for*] [*maxdelay*] [*exemptfee*]
-[*exclude*]
+[*maxfeepercent*] [*retry_for*] [*maxdelay*] [*exemptfee*]
```
vincenzopalazzo, line 8:
```suggestion
[*maxfeepercent*] [*retry\_for*] [*maxdelay*] [*exemptfee*]
```

doc/lightning-sendpay.7.md
```diff
@@ -5,7 +5,8 @@ SYNOPSIS
 --------
 
 **sendpay** *route* *payment\_hash* [*label*] [*msatoshi*]
-[*bolt11*] [*payment_secret*] [*partid*] [*localofferid*] [*groupid*] [*payment_metadata*]
+[*bolt11*] [*payment_secret*] [*partid*] [*localofferid*] [*groupid*]
+[*payment_metadata*] [*description*]
```
vincenzopalazzo, line 9:
```suggestion
[*bolt11*] [*payment\_secret*] [*partid*] [*localofferid*] [*groupid*]
[*payment\_metadata*] [*description*]
```

this is not a new change, at this point, I'm not really sure that `\` is required!

vincenzopalazzo (commented):
LGTM, just some minor doc format, where I'm not sure that the suggested syntax is required.

ACK https://github.com/ElementsProject/lightning/pull/5122/commits/9d71db264e16cefcb0af6e81958789420be5df8b

doc/lightning-pay.7.md
rustyrussell: No, we've been eliminating those _

vincenzopalazzo: oh ok is the opposite!

Issue #4991, opened 2021-12-24
c-lightning crashed on payment


hMsats:
```
2021-12-24T17:51:13.640Z INFO    plugin-pay: cmd 25104 partid 0: Initial limit on max HTLCs: 2565, Destination 028d98b9969fbed53784a36617eb489a59ab6dc9b9d77fcdca9ff55307cd98e3c4 has 171 channels, assuming 15 HTLCs per channel
2021-12-24T17:51:13.696Z INFO    plugin-pay: cmd 25104 partid 1: Split into 10 sub-payments due to initial size (100609000msat > 10000000msat): new partid 2, new partid 3, new partid 4, new partid 5, new partid 6, new partid 7, new partid 8, new partid 9, new partid 10, new partid 11
2021-12-24T17:51:45.823Z UNUSUAL 03abf6f44c355dec0d5aa155bdbdd6e0c8fefe318eff402de65c6eb2e1be55dc3e-channeld-chan#2: Adding HTLC 1631 too slow: killing connection
2021-12-24T17:51:45.825Z INFO    03abf6f44c355dec0d5aa155bdbdd6e0c8fefe318eff402de65c6eb2e1be55dc3e-chan#2: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (9)
2021-12-24T17:51:45.838Z INFO    plugin-pay: cmd 25104 partid 2: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (Outgoing subdaemon died)
2021-12-24T17:51:45.838Z INFO    plugin-pay: cmd 25104 partid 9: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (Outgoing subdaemon died)
2021-12-24T17:51:45.839Z INFO    plugin-pay: cmd 25104 partid 7: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (Outgoing subdaemon died)
2021-12-24T17:51:45.839Z INFO    plugin-pay: cmd 25104 partid 5: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (Outgoing subdaemon died)
2021-12-24T17:51:45.840Z INFO    plugin-pay: cmd 25104 partid 8: failed: WIRE_TEMPORARY_CHANNEL_FAILURE (Outgoing subdaemon died)
[...]

PR #5124, opened 2022-03-24
plugins/pay: don't crash on malformed time.

rustyrussell:
See: https://github.com/ElementsProject/lightning/issues/4991

We seem to correctly set end_time everywhere, so this looks like
a use-after-free somehow?  But this will fix the crash right here :(

@cdecker can you see if you can see anything?  Some weird stuff happening in that log...
Changelog-None

### DIFF
--- a/plugins/pay.c
+++ b/plugins/pay.c
@@ -1387,9 +1387,15 @@ static struct command_result *json_pay(struct command *cmd,
 static void utc_timestring(const struct timeabs *time, char str[UTC_TIMELEN])
 {
 	char iso8601_msec_fmt[sizeof("YYYY-mm-ddTHH:MM:SS.%03dZ")];
+	struct tm *t = gmtime(&time->ts.tv_sec);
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
Trivial rebase.

plugins/pay.c
```diff
@@ -1387,9 +1387,15 @@ static struct command_result *json_pay(struct command *cmd,
 static void utc_timestring(const struct timeabs *time, char str[UTC_TIMELEN])
 {
 	char iso8601_msec_fmt[sizeof("YYYY-mm-ddTHH:MM:SS.%03dZ")];
+	struct tm *t = gmtime(&time->ts.tv_sec);
```
cdecker, line 1390: Interesting take, I was assuming that this could not possibly return `NULL`, btu the man page says:

> It may return NULL when the year does not fit into an integer

My best guess was this https://github.com/ElementsProject/lightning/tree/issue-4991 assuming that the buffer was too small, but yours seems more likely.

The root cause seems uninitialized memory, will dig further :-)

cdecker (approved):
ACK c938e0e5cd265fc7d457c0b131c0941ed563e930

PR #5125, opened 2022-03-24
Lnprototest update, and timeout CI if it goes awry

rustyrussell:
Changelog-None

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,6 +104,7 @@ jobs:
       EXPERIMENTAL_FEATURES: 1
       COMPAT: 0
       PYTEST_PAR: 2
+      PYTEST_OPTS: "--timeout=300"
       TEST_CMD: "make check-protos"
       TEST_GROUP: 1
       TEST_GROUP_COUNT: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -104,6 +104,7 @@ jobs:
       EXPERIMENTAL_FEATURES: 1
       COMPAT: 0
       PYTEST_PAR: 2
+      PYTEST_OPTS: "--timeout=300"
```
vincenzopalazzo, line 107: it is possible also to use the Github action timeout `timeout-minutes: 300` in order to make stay safe at all level

vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5125/commits/f4fae41c1b44ffdede646d835cab0b361b7b1b79

Good catch, I was looking into reproducing it but without luck!

vincenzopalazzo:
Mh, I think this is a GitHub problem, the same lnprototest is raining on the docker image https://github.com/rustyrussell/lnprototest/runs/5670869353?check_suite_focus=true in 12 minutes including the c-lightning build!

I can make another PR with the migration to Docker? What do you think?

.github/workflows/ci.yaml
rustyrussell: No, I want this inside the test, so we get a backtrace...

vincenzopalazzo: Oh! I tested it and I got what you want!

I introduce in lnprototest also the logger that you can enable by passing to pytest `--log-cli-level=DEBUG`!

vincenzopalazzo:
This PR should be full override with the https://github.com/ElementsProject/lightning/pull/5129 to stop the nondeterministic behavior from Github actions, and also with docker, we have the opportunity to run the same CI env locally!

rustyrussell:
#5125  is better

PR #5126, opened 2022-03-24
test: avoid testing for secret_eq under CI.

rustyrussell:
It's failing under CI, but loadavg is under 1.  Simply remove it.

Changelog-None


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5126/commits/14d0f95e74ccc71bfaa8680d5fcc8d6c96959567

Do you think that there the Github VM can cause problem? maybe we can move it to Docker internally

PR #5128, opened 2022-03-24
invoice: Add description_hash

jsarenik:
Wondering if this would fix the problem in https://github.com/getAlby/lightning-browser-extension/issues/726

The script I use for generating LNURL-pay can be found at https://github.com/jsarenik/bitcoin-faucet-shell/blob/master/cgi-bin/lnurl.sh


jsarenik:
This patch does not work, I am wondering if something similar and working would solve the issue (while hoping it is not a lot of wasted CPU time to add this hash).

fiatjaf:
https://github.com/ElementsProject/lightning/pull/5121

PR #5129, opened 2022-03-24
ci: migrate the protocol test inside the docker image

vincenzopalazzo:
I noted that some protocol tests takes a while and rusty find a bug in lnprototest, but the lnprototest still is stuck https://github.com/ElementsProject/lightning/pull/5125

With this PR I'm trying to migrate the work on the Docker image because in the lnprototest integration testing works like a charm in only 12 minutes https://github.com/rustyrussell/lnprototest/runs/5670869353?check_suite_focus=true

### DIFF
--- a/contrib/docker/scripts/build.sh
+++ b/contrib/docker/scripts/build.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+echo "Running in $(pwd)"
+export ARCH=${ARCH:-64}
+export BOLTDIR=lightning-rfc
+export CC=${COMPILER:-gcc}
+export COMPAT=${COMPAT:-1}
+export TEST_CHECK_DBSTMTS=${TEST_CHECK_DBSTMTS:-0}
+export DEVELOPER=${DEVELOPER:-1}
+export EXPERIMENTAL_FEATURES=${EXPERIMENTAL_FEATURES:-0}
+export PATH=$CWD/dependencies/bin:"$HOME"/.local/bin:"$PATH"
+export PYTEST_OPTS="--maxfail=5 --suppress-no-test-exit-code ${PYTEST_OPTS}"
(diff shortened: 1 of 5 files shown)


vincenzopalazzo:
We win @rustyrussell in just 14m 20s

Thanks, @cdecker for the amazing bash script!

vincenzopalazzo:
It also include the update to lnprototest at the latest version of the master, so this PR will full override https://github.com/ElementsProject/lightning/pull/5125

Let see what the the CI tell us

vincenzopalazzo:
Trivial rebase on master https://github.com/ElementsProject/lightning/commit/20392ae526e4540245f1e0e4ba112c034e6f71d0

rustyrussell (approved):
Ack 6c41b3dddc527ad4640b3c77bd93ee2d7adb1135

contrib/docker/scripts/build.sh
```diff
@@ -21,15 +21,20 @@ export VALGRIND=${VALGRIND:-0}
...
 
+# Used in docker image
+cd /work || return
+ls -la
+
 ./configure CC="$CC"
-cat config.vars
+# cat config.vars
 
 cat << EOF > pytest.ini
```
rustyrussell, line 39: These changes seem weird?

vincenzopalazzo: Yes also to me, However, for some reason, the script was run out of the work dir, and the following line just enforced to be on the right directory, and the `ls -la` is just to make the log richer, but we can remove it, but I will look where the change root directory happens

`cd /work || return`

In addition, the `cat config.vars` looks like it was not present, but it can be a refusal of my test. I will double-check that case too.

vincenzopalazzo: Thanks, I found the bug in my docker config!

Let's give a run to the CI and let's see what happens!

Issue #5131, opened 2022-03-25
Installation: launchpadlib's dependence on testresources causes error.


dscotese:
### Issue and Steps to Reproduce
1. Be running C-Lightning v0.8.1-2401-g4ae9e0c on Ubuntu 20.04.3 LTS \n \l
2. Follow these instructions from the Ubuntu section of the installation page:
```sudo apt-get update
sudo apt-get install -y \
  autoconf automake build-essential git libtool libgmp-dev libsqlite3-dev \
  python3 python3-pip net-tools zlib1g-dev libsodium-dev gettext
pip3 install --upgrade pip
pip3 install --user poetry
```
3. See if you get `ERROR: launchpadlib 1.10.13 requires testresources, which is not installed.`

I believe the fix is (from https://gist.github.com/y56/0540d22a1db40dacc7fbbb93c866821e) to add this line to the instructions above:
`sudo apt install python3-testresources`
(or add it to the apt-get install command so it auto-answers Yes to the prompt(s?) )

vincenzopalazzo:
>Be running C-Lightning v0.8.1-2401-g4ae9e0c on Ubuntu 20.04.3 LTS \n \l

This is a typo here, right? the version of c-lighnting is v0.10?

>I believe the fix is (from https://gist.github.com/y56/0540d22a1db40dacc7fbbb93c866821e) to add this line to the instructions above:
sudo apt install python3-testresources
(or add it to the apt-get install command so it auto-answers Yes to the prompt(s?) )

Concept ACK

dscotese:
I only indicated that I was running that version because it implies some state of my machine and the software on it, which might contribute to the Error.  I don't believe it matters otherwise.  For example, it may be the case that I skipped a version which would have installed what was missing.  In fact, it's probably irrelevant that I was already running c-lightning, as the instructions I was following come before even getting the source code.

rustyrussell:
AI cleanup of old issues.  If wrong, please open a new issue.


Triage decision summary:
Installation issue with launchpadlib dependency on Ubuntu 20.04 from March 2022. Likely resolved in current versions through dependency updates or installation procedure changes.

Issue #5115, opened 2022-03-20
None of the plugins are working since the switch to poetry


ClayShoaf:
TL;DR How am I supposed to install and run plugins now that the project is using poetry? 

I've been working for months now on just trying to get to the point where I could actually contribute some small things to plugins that I thought would be helpful. I finally got to the point where I could actually build c-lightning. Took me forever to figure out that I had to do `poetry run make` .  Is there some new way that I'm supposed to be using plugins? 

At first I was trying to run tests by trying everything I could think of.  
I tried copying the plugin to the `lightning/plugins` directory and using `poetry run pytest` (both before and after hackily making sure I had the right dependencies) to test it. That didn't work.
I tried exporting the poetry dependencies and installing all of them using pip and running pytest natively. That didn't work.
Finally, I just tried to see if the plugin would actually work at all. When I used `contrib/startup_regtest.sh` and tried to start the plugin, I get the following error:

```
[user@prgrmr ~/github/lightning]$ l1-cli plugin start /home/user/github/lightning/plugins/noise/noise.py
Traceback (most recent call last):
  File "/home/user/github/lightning/plugins/noise/noise.py", line 5, in <module>
    from pyln.client import Plugin, RpcError
ModuleNotFoundError: No module named 'pyln'
{
   "code": -3,
   "message": "/home/user/github/lightning/plugins/noise/noise.py: exited before replying to getmanifest"
}
```
Probably has to do with the fact that I'm using pyenv and it's somehow trying to use my system version of python, idk, maybe it's something else. I can't even install any of the pyln stuff because it all used relative paths, and pip doesn't recognize things like `../` as a valid URL. I could copy the `pyln-*/pyln` folders manually to my site-packages folder like I did for the version I'm using in pyenv, but there is no way that this method is sustainable. There has to be something I'm missing. I'm almost at the end of my rope here.

Before this switch to poetry, I was just starting to get things to work on my Arch machine by using `pyenv` to use a compatible version of python. 
Tests were working and I was finally getting ready to work on the stuff that I actually wanted to work on. All I had to do was send a PR to add Arch installation instructions to `doc/INSTALL.md` and I would have been all set to contribute and actually write tests and everything. 

Now it's like I'm completely back to square one. 

Weeks worth of tweaking/testing and figuring things out was all for nothing. This is extremely frustrating.

vincenzopalazzo:
[...]

PR #5133, opened 2022-03-25
docs: adding missing dependencies

vincenzopalazzo:
Fixes https://github.com/ElementsProject/lightning/issues/5131
Fixes https://github.com/ElementsProject/lightning/issues/5115 , but better to have a double check by @cdecker I don't want think that my idea is correct without a feedback


cdecker:
nack, this is a completely unrelated library that I have no idea how it made it into the dependency tree. As commented on the original issue, the commit that they were testing was also outdated, and should not be used to infer dependencies.

vincenzopalazzo:
I assumed that it was a typo in the version reporting, sorry for that.

Closing!

Issue #5092, opened 2022-03-14
websocketd creates many zombies


jb55:
something I noticed on my machine:

```
jb55     1316952  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1317227  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1319353  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1322418  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1383098  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1397453  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1402480  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1403333  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1703541  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     1847968  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3383387  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3429083  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3485778  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3517108  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3518072  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3518288  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
jb55     3518997  0.0  0.0      0     0 pts/1    Z+   Mar13   0:00 [lightning_webso] <defunct>
[...]

PR #5134, opened 2022-03-25
startup_regtest.sh: add connect helper and experimental-offers

jb55:
Here are a few additions to `startup_regtest.sh` to make it easier to connect nodes via a new `connect` command.

Also add `experimental-offers` to the node configs for testing offers.

### DIFF
--- a/contrib/startup_regtest.sh
+++ b/contrib/startup_regtest.sh
@@ -14,16 +14,12 @@
 ##
 ##  $ start_ln 3
 ##
-##  Let's connect the nodes.
+##  Let's connect the nodes. The `connect a b` command connects node a to b.
 ##
-##  $ l2-cli getinfo | jq .id
-##    "02b96b03e42d9126cb5228752c575c628ad09bdb7a138ec5142bbca21e244ddceb"
-##  $ l2-cli getinfo | jq .binding[0].port
-##    9090
-##  $ l1-cli connect 02b96b03e42d9126cb5228752c575c628ad09bdb7a138ec5142bbca21e244ddceb@localhost:9090
-##    {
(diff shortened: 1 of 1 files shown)


jb55:
On Fri, Mar 25, 2022 at 01:23:30PM -0700, Vincenzo Palazzo wrote:
>Maybe is a better naming convention?

I just wanted to make it clear that it takes node numbers

contrib/startup_regtest.sh
```diff
@@ -216,3 +213,12 @@ stop_elem() {
 	unset LN_NODES
 	unalias et-cli
 }
+
+connect() {
+	if [ -z $1 ] || [ -z $2 ]; then
+		printf "usage: connect 1 2\n"
+	else
+		to=$($LCLI --lightning-dir=/tmp/l$2-$network getinfo | jq -r '"\(.id)@\(.binding[0].address):\(.binding[0].port)"')
```
rustyrussell, line 221: Hmm, let's not rely on jq?  Maybe:

```
lightning-cli -F getinfo | grep '^\(id\|address\[0\]\.\(address\|port\)\)' | cut -d= -f2- | tr '\n' ' ' | (read ID ADDR PORT; echo $ID@${ADDR}:$PORT)
```

ddustin:
This PR may be relevant https://github.com/ElementsProject/lightning/pull/5062 as it builds two nodes and connects them as well

niftynei:
I want the `experimental-offers` patch! We're looking to get a RC out this week. If the command in #5092 solves this, maybe we just make this the first patch?

contrib/startup_regtest.sh
jb55: done

jb55:
I think I have appeased the shellcheck gods to a sufficient degree for this change to be acceptable.

rustyrussell (approved):
Ack 8e01c925e272bef3dae4c86a9a1b351b504d78b4

PR #5135, opened 2022-03-26
Update INSTALL.md

ClayShoaf:
changed `make` to `poetry run make` as well as added instructions for adding local binary folder to `$PATH` as it is not there by default in a fresh install of Ubuntu 20.04

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -43,6 +43,11 @@ Get dependencies:
     pip3 install --user poetry
     poetry install
 
+If you are getting errors that `/home/<your username>/.local/bin/` is not in `PATH`, you will need to add it.
+
+    echo -e "\n\n"'export PATH="/home/<your username>/.local/bin:$PATH"' >>~/.bashrc
+    source ~/.bashrc
+
 If you don't have Bitcoin installed locally you'll need to install that
 as well. It's now available via [snapd](https://snapcraft.io/bitcoin-core).
 
@@ -65,7 +70,7 @@ For development or running tests, get additional dependencies:
 Build lightning:
 
     ./configure
-    make
+    poetry run make
     sudo make install
 
 Running lightning:
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -41,12 +41,6 @@ Get dependencies:
       python3 python3-pip net-tools zlib1g-dev libsodium-dev gettext
     pip3 install --upgrade pip
     pip3 install --user poetry
-    poetry install
-
-If you are getting errors that `/home/<your username>/.local/bin/` is not in `PATH`, you will need to add it.
-
-    echo -e "\n\n"'export PATH="/home/<your username>/.local/bin:$PATH"' >>~/.bashrc
-    source ~/.bashrc
 
 If you don't have Bitcoin installed locally you'll need to install that
 as well. It's now available via [snapd](https://snapcraft.io/bitcoin-core).
@@ -69,6 +63,7 @@ For development or running tests, get additional dependencies:
 
 Build lightning:
 
+    poetry install
     ./configure
     poetry run make
     sudo make install


doc/INSTALL.md
```diff
@@ -43,6 +43,11 @@ Get dependencies:
     pip3 install --user poetry
     poetry install
 
+If you are getting errors that `/home/<your username>/.local/bin/` is not in `PATH`, you will need to add it.
+
+    echo -e "\n\n"'export PATH="/home/<your username>/.local/bin:$PATH"' >>~/.bashrc
+    source ~/.bashrc
+
```
vincenzopalazzo, line 50: nit: I think this is information not related to c-lightning install procedure, and this should not stay in the docs IMHO.

vincenzopalazzo (changes requested):
`poetry run make` it is good as well as `poetry shell` and after keep all your command as is.

doc/INSTALL.md
ClayShoaf: I put that in there because I was getting that error with a fresh install of Ubuntu. I was thinking of users that may want to run a node but that aren't as familiar with linux. Now that I'm looking at the main README.md again, I see that people are expected to install with a ppa, so maybe I shouldn't have worried about that little part.

I guess I was thinking of myself early on and how many times I wish there had been something like that in docs that I was reading. I guess they can just type their error message into a search engine and figure it out.

vincenzopalazzo: >I guess they can just type their error message into a search engine and figure it out.

Yes, usually I type the error on google before read 20 lines of docs

rustyrussell:
I don't think this advice works very well?

Poetry will install inside a virtual env, but when you actually run lightningd you won't have installed the python packages: none of the Python plugins will work.

Ideally you want to install the python packages system wide, but we removed requirements.txt and now my brain hurts.

@cdecker is this The Right Thing?

cdecker:
No, after digging into how things work it seems like `pip install .` is what we want (assuming `pip` is recent enough to know how to handle `pyproject.toml` files). Thats also why `poetry` defaults to always use a virtualenv, since pip can be used for installations instead.

ClayShoaf:
I tried pulling the most recent version of this repo and installing with `pip install .` using `pip 22.0.4` and it got hung up thinking that `cln-meta-project` was a package.

I've been using a sort of hacky way to get the requirements by using
```
poetry export --dev --without-hashes | cut -d ';' -f1 >requirements.txt
```
but I still have to manually install `pyln` stuff because pip doesn't understand relative paths. It also has an issue with the `mrkd` url that is exported.

vincenzopalazzo:
In my understanding of this problem is when we are doing `sudo make install`, looks like lightning will run the python instance with some different env from the env that pip user.

But I know and I agree with what @cdecker is saying because poetry use pip under the hood, however, I didn't find a way to proof that what @rustyrussell is saying will not happen.

I will redo some of the tests on some clean machines, but until now I had no success to work around the `sudo make install`


>but I still have to manually install pyln stuff because pip doesn't understand relative paths. It also has an issue with the mrkd url that is exported.


This is a workaround, but we need also to mention that the python plugins has some requirements.txt to be installed with pip, so this problem is caused not by the new poetry commands, but because no one are running `pip3 install -r requirements.txt`

The plugin not in dev mode should never run a not tagged version IMHO

rustyrussell:
OK, Python is a mess.  Surprise!

It seems that modern pip should understand the Poetry file, but it doesn't, because we use relative paths.  So I think this fix is actually the right one: plugins will need to handle their own requirements.txt or whatever.

doc/INSTALL.md
rustyrussell: You actually only need to log in again: if pip created the dir, it wasn't in your path.

PR #5136, opened 2022-03-26
feat: config additions

m-schmoock:
This adds config and commanline switches for HTLC min/max values and announcement of auto discovered IPs

### DIFF
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -278,6 +278,23 @@ values), and the order is "opening", "mutual_close", "unilateral_close",
 You would usually put this option in the per-chain config file, to avoid
 setting it on Bitcoin mainnet!  e.g. `~rusty/.lightning/regtest/config`.
 
+ **htlc-minimum-msat**=*MILLISATOSHI*
+Default: 0. Sets the minimal allowed HTLC value for newly created channels.
+If you want to change the `htlc_minimum_msat` for existing channels, use the
+RPC call lightning-setchannel(7).
+
+ **htlc-maximum-msat**=*MILLISATOSHI*
+Default: unset (no limit). Sets the maximum allowed HTLC value for newly created
+channels. If you want to change the `htlc_maximum_msat` for existing channels,
+use the RPC call lightning-setchannel(7).
+
+ **disable-ip-discovery**
+Turn off public IP discovery to send `node_announcement` updates that contain
+the discovered IP with TCP port 9735 as announced address. If unset and you
+open TCP port 9735 on your router towords your node, your node will remain
+connectable on changing IP addresses.  Note: Will always be disabled if you use
+'always-use-proxy'.
+
 ### Lightning channel and HTLC options
 
  **large-channels**
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -757,6 +757,10 @@ static const struct config testnet_config = {
 	/* Testnet blockspace is free. */
 	.max_concurrent_htlcs = 483,
 
+	/* channel defaults for htlc min/max values */
+	.htlc_minimum_msat = AMOUNT_MSAT(0),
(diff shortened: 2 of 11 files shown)


lightningd/options.c
```diff
@@ -1117,6 +1137,9 @@ static void register_opts(struct lightningd *ld)
 	opt_register_arg("--announce-addr", opt_add_announce_addr, NULL,
 			 ld,
 			 "Set an IP address (v4 or v6) or .onion v3 to announce, but not listen on");
+	opt_register_noarg("--disable-ip-discovery", opt_set_invbool,
+			 &ld->config.announce_discovered_ip,
+			 "Turn off address announcement of detected public IP");
```
rustyrussell, line 1142: Just a note: I usually prefer positive options, not negative, with the default being `!ld->always_use_proxy`.

In theory always-use-proxy could be a normal SOCKS proxy, and for all we know you've set up a reverse tunnel.  But that's a really weird case (and you can just advertize an explicit IP for that).

rustyrussell (approved):
Ack 0768ffc7c3e5cd4858b648ca7b288f898a763344

doc/lightningd-config.5.md
```diff
@@ -278,6 +278,23 @@ values), and the order is "opening", "mutual_close", "unilateral_close",
 You would usually put this option in the per-chain config file, to avoid
 setting it on Bitcoin mainnet!  e.g. `~rusty/.lightning/regtest/config`.
 
+ **htlc_minimum_msat**=*MILLISATOSHI*
```
rustyrussell, line 281: This is not the option name!

doc/lightningd-config.5.md
```diff
@@ -278,6 +278,23 @@ values), and the order is "opening", "mutual_close", "unilateral_close",
...
+Default: 0. Sets the minimal allowed HTLC value for newly created channels.
+If you want to change the `htlc_minimum_msat` for existing channels, use the
+RPC call lightning-setchannel(7).
+
+ **htlc_maximum_msat**=*MILLISATOSHI*
+Default: -1 (no limit). Sets the maximum allowed HTLC value for newly created
+channels. If you want to change the `htlc_maximum_msat` for existing channels,
+use the RPC call lightning-setchannel(7).
+
+ **announce_discovered_ip**=*BOOL*
```
rustyrussell, line 291: Nor is this...

rustyrussell (changes requested):

rustyrussell:
You need to add the names to the doc/schemas/listconfig.schema.json file too...

lightningd/options.c
m-schmoock: Hm, if I make it a positive bool option, then ip announcement can't be enabled by default, otherwise the switch would do nothing. Or do I get somthing wrong? What I could do is rename the internal code variables also to `disable_ip_discovery` which defaults to `false` and use an `opt_set_bool` instead of `opt_set_invbool`. Do you meant that?

m-schmoock: I renamed the internal variable also to `disable_ip_discovery` defaulting to `false`. This is now a switch, not an bool option. The switch effectively turns of ip discovery anouncements by using `opt_set_bool` instead of `opt_set_invbool`. 

Can you re-check?

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5136/commits/0f25b7b37fb2eb814e3e14d0539b1e5034f838f1

rustyrussell (approved):

heyambob:
'''
disable-ip-discovery=true
'''

I put this in the config, and it says

'''
lightningd: /clight/config line 41: disable-ip-discovery: doesn't allow an argument
'''

But passing it on the command line is okay.

m-schmoock:
That's because it's a switch and not an argument. Text without `=true` :)

PR #5141, opened 2022-03-28
tests: add test that pay gets update correct.

rustyrussell:
This is the part which works: we have another old PR (or simply
uncomment the FIXME section) for the routehint case.

Closes: #4781
See-also: #4808


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5141/commits/674b63be9270049516901ad0583f1926dc688be7

PR #5143, opened 2022-03-29
subd: make channel/peer own the subd.

rustyrussell:
We get some memleak reports because ld owns the subd, but once
the peer/channel is freed, there's no reference for the brief time
until the subd exits.

This happens for both opening and closingd.  For openingd, the
peer owns it, for others (including dualopend) the channel owns it.


vincenzopalazzo (approved):
Cool!

ack https://github.com/ElementsProject/lightning/pull/5143/commits/d245338ae4a091aedab752fea53bd13b7dfdff94

PR #5145, opened 2022-03-29
py: Update the PyPI job to use poetry and correct versions

cdecker:
Seems PyPI have recently bumped the `long_description` warning to become an error. This PR addresses that by adding the README.md as the `long_description` where we have one, and using `poetry` for publication as well, instead of mixing tools.

Changelog-None

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -38,35 +38,37 @@ jobs:
       with:
         python-version: 3.7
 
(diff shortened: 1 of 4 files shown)


vincenzopalazzo (approved):
ack https://github.com/ElementsProject/lightning/pull/5145/commits/e8c56f07ffa1fc841997ee2e4fda14eb0a36e1b8

PR #5152, opened 2022-04-01
channeld: send warning, not error if peer has old commitment number.

rustyrussell:
This is the minimal change to meet the desired outcome of https://github.com/lightning/bolts/issues/934
which wants to give obsolete-db nodes a chance to fix things up, before we
close the channel.

We need to dance around a bit here, since we *will* close the channel if
we receive an ERROR, so we suppress that.

### DIFF
--- a/channeld/channeld.c
+++ b/channeld/channeld.c
@@ -3000,12 +3000,14 @@ skip_tlvs:
 		}
 		retransmit_revoke_and_ack = true;
 	} else if (next_revocation_number < peer->next_index[LOCAL] - 1) {
-		peer_failed_err(peer->pps,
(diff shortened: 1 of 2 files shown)


t-bast:
This title of the PR is misleading, shouldn't it be exactly the opposite (ie `send warning, not error`)?

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5152/commits/3532d3d2ef5720f6e793f13fe14bd048355c1dfc

cdecker:
Rebased on top of `master` to get CI to rerun and ready the PR for merge.

PR #5104, opened 2022-03-16
cln-rpc & cln-grpc: Backfill more methods

cdecker:
I now know how @rustyrussell felt filling in the JSON-Schemas for the JSON-RPC responses. This PR mostly consists of me backfilling the JSON-RPC request schemas, and then autogenerating the derived files, occasionally addressing a conversion issue or a missing type.

The PR is based on topp of #5013 and starts with the `msggen: Add meta file to keep the field numbers stable` commit (daa26ff27d7941aefdb709a9dab72b21f1d29c03 at the time of writing).

The following files where changed manually, all others were generated and are unlikely to hold any surprises:
 - `cln-grpc/proto/primitives.proto ` to add the `Feerate`, `Utxo` and `OutputDesc` primitives
 - `cln-grpc/src/pb.rs` to add a number of protobuf <-> JSON-RPC conversions
 - `cln-rpc/Cargo.toml` to add the `hex` dependency
 - `cln-rpc/src/primitives.rs`
 - `contrib/msggen/msggen/__main__.py` to allowlist more RPC methods and add method renaming for `connect` which is already used by grpc internally
 - `contrib/msggen/msggen/grpc.py` to add the fixed mapping from field path to field number required for grpc to work across versions (renumbering fields is not allowed by protobuf and grpc since that's how fields are identified).
 - `contrib/msggen/msggen/model.py` and `contrib/msggen/msggen/rust.py` to add mapped types
 - The following request schemas:
   - `doc/schemas/connect.request.json`
   - `doc/schemas/createinvoice.request.json`
   - `doc/schemas/createonion.request.json`
   - `doc/schemas/datastore.request.json`
   - `doc/schemas/deldatastore.request.json`
   - `doc/schemas/delexpiredinvoice.request.json`
   - `doc/schemas/delinvoice.request.json`
   - `doc/schemas/fundpsbt.request.json`
   - `doc/schemas/invoice.request.json`
   - `doc/schemas/keysend.request.json`
   - `doc/schemas/listdatastore.request.json`
   - `doc/schemas/listinvoices.request.json`
   - `doc/schemas/listnodes.request.json`
   - `doc/schemas/listsendpays.request.json`
   - `doc/schemas/listtransactions.request.json`
   - `doc/schemas/newaddr.request.json`
   - `doc/schemas/pay.request.json`
   - `doc/schemas/sendonion.request.json`
   - `doc/schemas/sendpay.request.json`
   - `doc/schemas/sendpsbt.request.json`
[...]

PR #5153, opened 2022-04-01
Pr 5104

rustyrussell:
This is #5104, but GH seems to think that has conflicts?


cdecker:
That is because it uses the predecessor PR as the merge-base, to simplify reviews, and not have to manage multiple PRs in parallel. The edit button under the title can be used to switch the merge-base back to `master`.

PR #5165, opened 2022-04-06
doc: misc updates, including giant Core Lightning rename

rustyrussell:
In particular, using cargo to install rustfmt doesn't work.

Changelog-None

### DIFF
--- a/cli/lightning-cli.c
+++ b/cli/lightning-cli.c
@@ -276,7 +276,7 @@ static bool is_literal(const char *arg)
 	if (arglen == 0) {
 		return false;
 	}
-	return strspn(arg, "0123456789") == arglen
+	return strspn(arg, "0123456789.") == arglen
 		|| streq(arg, "true")
 		|| streq(arg, "false")
 		|| streq(arg, "null")
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -61,6 +61,10 @@ For development or running tests, get additional dependencies:
     sudo apt-get install -y valgrind libpq-dev shellcheck cppcheck \
       libsecp256k1-dev jq
 
+If you want to build the Rust plugins (currently, cln-grpc):
+
+	sudo apt-get install -y cargo rustfmt
+
 Build lightning:
 
     poetry install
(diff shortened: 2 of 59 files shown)


rustyrussell:
I refer to this, too:

https://github.com/bitcoinops/bitcoinops.github.io/pull/742#discussion_r843360764

cli/lightning-cli.c
```diff
@@ -276,7 +276,7 @@ static bool is_literal(const char *arg)
 	if (arglen == 0) {
 		return false;
 	}
-	return strspn(arg, "0123456789") == arglen
+	return strspn(arg, "0123456789.") == arglen
```
vincenzopalazzo, line 279: Unable to catch the main motivation of this, it is related to some issue that we have?

doc/INSTALL.md
```diff
@@ -61,6 +61,10 @@ For development or running tests, get additional dependencies:
     sudo apt-get install -y valgrind libpq-dev shellcheck cppcheck \
       libsecp256k1-dev jq
 
+If you want to build the Rust plugins (currently, cln-grpc):
+
+	sudo apt-get install -y cargo rustfmt
```
vincenzopalazzo, line 66: not sure if this will install also the rust compiler `rustc`, maybe we can reproduce the official installer guide in rust? https://www.rust-lang.org/tools/install
```suggestion
	curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
	
Can you see the official process [on rust official site for other types of installation](https://www.rust-lang.org/tools/install)
```

Maybe the last line need a rework

vincenzopalazzo (changes requested):

cli/lightning-cli.c
rustyrussell: That commit has the rationale:

The documentation promises that cli will pass through any numbers, including decimal point.  We would have converted to a string.

doc/INSTALL.md
rustyrussell: Yes, it does, I tested on a clean install.

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5165/commits/ad1a2f0519abaa3c044da94203cd2e5ca4b1f73d

Issue #5162, opened 2022-04-05
Not clear how to use gRPC 


kiwiidb:
### Issue and Steps to Reproduce
I've already figured out I should add Rust to my Dockerfile and `make` seems to have picked it up:
```
...
 mrkd doc/lightning-hsmtool.8.md
(cd plugins; cargo build --quiet --examples)
cargo build --quiet --bin cln-grpc
cargo build --quiet --example cln-rpc-getinfo
cd external/libwally-core && ./tools/autogen.sh
...
```
However, `lightningd` won't start:
```
lightningd: --grpc-port=9737: unrecognized option
```

rustyrussell:
Ah!  There are two problems:

1. It's not in the plugins/ toplevel directory, so ./lightningd/lightningd won't find it.
2. It's not in the install target, so even when installed it won't work.

Will fix...

KayBeSee:
Has this been added?

I am running v0.11.0.1 and run `lightning-cli plugin list` but am not seeing it in the list of plugins.

```
{
   "command": "list",
   "plugins": [
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/autoclean",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/bcli",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/funder",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/topology",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/keysend",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/offers",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/pay",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/txprepare",
         "active": true
      },
      {
         "name": "/usr/local/bin/../libexec/c-lightning/plugins/spenderp",
         "active": true
      }
   ]
}
```

I also get the `clightning_1  | lightningd: /data/.lightning/config line 9: grpc-port=12345: unrecognized option` when I put `grpc-port=12345` in my config file.

whitslack:
> Has this been added?

@KayBeSee: You need Rust installed on the system where you built Core Lightning. Check the output of `configure` to see whether Rust was detected.

[...]

PR #5166, opened 2022-04-06
plugins: add grpc-plugin to PLUGINS, so it gets started and installed.

rustyrussell:
Fixes: #5162
Changelog-None

### DIFF
--- a/configure
+++ b/configure
@@ -111,7 +111,7 @@ default_valgrind_setting()
 
 default_rust_setting()
 {
-    if cargo --version > /dev/null 2>&1; then
+    if cargo --version > /dev/null 2>&1 && rustfmt --version >/dev/null 2>&1; then
 	echo 1
     else
 	echo 0
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -155,7 +155,7 @@ clean: doc-clean
 check: check-manpages
 
 # This needs plugins, too.
-check-manpages: all-programs check-config-docs
+check-manpages: all-programs check-config-docs default-targets
 	@tools/check-manpage.sh cli/lightning-cli doc/lightning-cli.1.md
 	@tools/check-manpage.sh "lightningd/lightningd --lightning-dir=/tmp/" doc/lightningd-config.5.md
 
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -220,6 +220,11 @@ If there is no `hsm_secret` yet, `lightningd` will create a new encrypted secret
 If you have an unencrypted `hsm_secret` you want to encrypt on-disk, or vice versa,
 see lightning-hsmtool(8).
 
+ **grpc-port**=*portnum* [plugin `cln-grpc`]
+
+The port number for the GRPC plugin to listen for incoming
+connections; default is not to activate the plugin at all.
+
 ### Lightning node customization options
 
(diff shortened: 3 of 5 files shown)


cdecker:
ACK 9777eb671ff89cbaf18233e0134c381f7cb17830

rustyrussell:
Hmm, now it *is* exiting silently, too soon!  Even caught it under strace, but nothing informative; it's doing an exit(0) *sometimes*.

vincenzopalazzo:
>Hmm, now it is exiting silently, too soon! Even caught it under trace, but nothing informative; it's doing an exit(0) sometimes.

Maybe we shouldn't mark the plugin as important? there is a case where the plugin exit in the normal operation, and if I read correctly the code, this should be the case where we are not setting the `grpc-port` value.

```rust
        Some(options::Value::Integer(-1)) => {
            log::info!("`grpc-port` option is not configured, exiting.");
            return Ok(());
        }
        Some(options::Value::Integer(i)) => i,
        None => return Err(anyhow!("Missing 'grpc-port' option")),
        Some(o) => return Err(anyhow!("grpc-port is not a valid integer: {:?}", o)),
    };
```

instagibbs:
this patch also seems to fix my inability to run:
```
make -j12 full-check PYTEST_PAR=24 DEVELOPER=1 VALGRIND=0
```
from `HACKING.md`

rustyrussell:
... all builtin plugins are "important".  There's a designated way for them to say they don't want to be started: you return `{"disable": "... reason why I want to be disabled"}` from the JSON init call.

I'm going to cut rc2 today without this, but clearly there will need to be an rc3: we can't *not* install this plugin!

vincenzopalazzo:
>... all builtin plugins are "important". There's a designated way for them to say they don't want to be started: you return {"disable": "... reason why I want to be disabled"} from the JSON init call.

I will try to look into that tomorrow!

vincenzopalazzo (changes requested):
The actual code skips an `rpc-plugin` example that it is used inside the tests.

In the PR https://github.com/ElementsProject/lightning/pull/5173 I put this addition to the bottom at line 185 again

rustyrussell:
Rebased on master now that cln-grpc is fixed!

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/commit/6d6c6706833a1c84063c18d7666d026839147ac4

whitslack:
@rustyrussell: Question about the Rust build: should these commands…
```
PYTHONPATH=contrib/msggen python3 contrib/msggen/msggen/__main__.py
```
…be completed before Cargo starts compiling Rust code? It appears to my casual eye that `msggen` generates some Rust code, but I often see it running after Cargo has already started. Maybe the Makefiles are missing a dependency somewhere?

cdecker:
> @rustyrussell: Question about the Rust build: should these commands…
> 
> ```
> PYTHONPATH=contrib/msggen python3 contrib/msggen/msggen/__main__.py
> ```
> 
> …be completed before Cargo starts compiling Rust code? It appears to my casual eye that `msggen` generates some Rust code, but I often see it running after Cargo has already started. Maybe the Makefiles are missing a dependency somewhere?

That should indeed be the case, not sure if there's a dependency between make targets missing.

PR #5172, opened 2022-04-07
Mac monterey build fixes

mdance:
I have been working in the comments with https://github.com/ElementsProject/libwally-core/pull/321 to try and compile lightning on  Darwin 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64

I was able to get it to compile, I am not a c programmer so am not expecting this to make it in, but hopefully it helps someone

### DIFF
--- a/plugins/offers_invreq_hook.c
+++ b/plugins/offers_invreq_hook.c
@@ -433,7 +433,7 @@ static bool check_payer_sig(struct command *cmd,
 
 	if (secp256k1_schnorrsig_verify(secp256k1_ctx,
 					sig->u8,
-					sighash.u.u8, &payer_key->pubkey) == 1)
+					sighash.u.u8, sizeof(sighash.u.u8), &payer_key->pubkey) == 1)
 		return true;
 
 	if (!deprecated_apis)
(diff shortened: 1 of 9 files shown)


plugins/offers_invreq_hook.c
```diff
@@ -447,7 +447,7 @@ static bool check_payer_sig(struct command *cmd,
 
 	return secp256k1_schnorrsig_verify(secp256k1_ctx,
 					   sig->u8,
-					   sighash.u.u8, sizeof(sighash.u.u8), &payer_key->pubkey) == 1;
+					   sighash.u.u8, sizeof(sighash), &payer_key->pubkey) == 1;
```
rustyrussell, line 450: I think it was more correct before these changes?  Thought this shouldn't change anything on any platform we support...

mdance: @rustyrussell that change was suggested https://github.com/ElementsProject/libwally-core/pull/321#issuecomment-1090958568 am happy to revert it out if incorrect

What is the best way for me to test?

I'm just playing around on testnet with lnurl, and a lightning addresses implementation so am not running anything mission critical, happy to be a guinea pig

mdance:
I have tried compiling both versions, and I think there is a problem, been trying to test with the following and it seems like something is broken:

lightning-cli signmessage test

{
   "signature": "[signature]",
   "recid": "[recid]",
   "zbase": "[zbase]"
}

lightning-cli checkmessage test [zbase]

{
   "pubkey": "[pubkey]",
   "verified": false
}

mdance:
Ignore my last comment, I have tested the same command against a arch linux master build and its returning the same thing, I thought the verified: false might have meant the signature verification wasn't working

PR #5173, opened 2022-04-07
rpc-plugin: introduce the possibility to disable important plugins

vincenzopalazzo:
Build on top of https://github.com/ElementsProject/lightning/pull/5166

This required some work to do, but I think this should be one of the general idea.

Another one that can be an alternative is keep the plugin active and maintains add two rpc method like `start_grpc` and `stop_grpc` but for now, I choose the simple think

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -48,10 +48,16 @@ where
     subscriptions: HashMap<String, Subscription<S>>,
 }
 
+pub trait PluginState {
+    fn disable_reason(&self) -> Option<String> {
+        None
+    }
+}
+
 impl<S, I, O> Builder<S, I, O>
 where
     O: Send + AsyncWrite + Unpin + 'static,
-    S: Clone + Sync + Send + Clone + 'static,
+    S: PluginState + Clone + Sync + Send + Clone + 'static,
     I: AsyncRead + Send + Unpin + 'static,
 {
     pub fn new(state: S, input: I, output: O) -> Self {
@@ -274,7 +280,16 @@ where
             }
         }
 
-        Ok(messages::InitResponse::default())
+        match self.state.disable_reason() {
+            Some(reason) => {
+                let init_resp = messages::InitResponse{
+                    disable: Some(reason),
+                };
+                Ok(init_resp)
+            },
+            None => Ok(messages::InitResponse::default()),
+        }
+
     }
 }
 
@@ -538,6 +553,7 @@ where
     pub fn state(&self) -> &S {
         &self.state
     }
+    pub fn mut_state(&mut self) -> &mut S {&mut self.state}
 }
 
 impl<S> Plugin<S>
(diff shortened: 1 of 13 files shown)


vincenzopalazzo:
There is more to do in the docs side, I will check at the end of the CI

```
➜  lightning git:(macros/plugin_init) ✗ make check-manpages
Error: Got unexpected message None from lightningd
diff of command names vs manpage names:
--- /dev/fd/63	2022-04-07 12:24:29.300060715 +0200
+++ /dev/fd/62	2022-04-07 12:24:29.300060715 +0200
@@ -36,7 +36,6 @@
 fetchinvoice-noconnect
 force-feerates=
 funding-confirms=
-grpc-port=
 help
 htlc-maximum-msat=
 htlc-minimum-msat=

```

plugins/src/lib.rs
```diff
@@ -274,7 +280,16 @@ where
...
+        match self.state.disable_reason() {
+            Some(reason) => {
+                let init_resp = messages::InitResponse{
+                    disable: Some(reason),
+                };
+                Ok(init_resp)
+            },
+            None => Ok(messages::InitResponse::default()),
+        }
+
```
vincenzopalazzo, line 292: auto review, this simplification should work

```suggestion
    let init_resp = messages::InitResponse{
             disable: self.state.disable_reason(),
           };
    Ok(init_resp)

```

plugins/src/messages.rs
```diff
@@ -132,6 +132,11 @@ pub(crate) struct GetManifestResponse {
 }
 
 #[derive(Serialize, Default, Debug)]
-pub struct InitResponse {}
+pub(crate) struct InitResponse {
+    /// enable the possibility to disable the plugins
+    /// when are marked as important.
+    #[serde(default)]
```
vincenzopalazzo, line 138: I need to check if `#[serde(default)]` can be removed

vincenzopalazzo (commented):
some a auto-comments

cdecker:
I dislike the idea of forcing users to implement a trait as well, just for this. Let's rather have a callback function we can add to `Builder` that can is called with the `Builder` instance as an argument, at the end of `handle_init`. That way it can set the `disable` flag on its instance and use that to then return the `InitResponse`.

vincenzopalazzo:
>Currently at a conference, making coding and review hard, sorry. But I proposed a simpler way of disabling for this release, while we figure out a cleaner interface going forward.

No problem, I just push the code with the hook as coincidence refactoring, there is a bug, but I think that as a draft it is good.

But yes, I know for now yours is simple, and we can work on this later

vincenzopalazzo:
>I dislike the idea of forcing users to implement a trait as well, just for this. Let's rather have a callback function we can add to Builder that can is called with the Builder instance as an argument, at the end of handle_init. That way it can set the disable flag on its instance and use that to then return the InitResponse.

I push the code for the CI that implements this feature including the bug fixed that I get this afternoon. However, I would like to improve the implementation in the future

vincenzopalazzo:
Closing in favor of https://github.com/ElementsProject/lightning/pull/5175

PR #5174, opened 2022-04-07
cln-grpc: Do not complain when we exit after `getmanifest`

cdecker:
We were logging a message about a missing `init` call when called with `--help`. That's normal, so let's not do that :-)


rustyrussell (approved):
Ack 05684713aa37897de81d4187dfbece17216a3242

PR #5175, opened 2022-04-07
Disable grpc plugin if `grpc-port` is not set

cdecker:
Alternative to #5173. This is relatively simple, and just disables listening, but doesn't exit if `grpc-port` isn't set, so we don't need to disable the plugin itself.
This is a quick fix, until we find a good way of doing `init` callbacks that are clean.

### DIFF
--- a/plugins/grpc-plugin/src/main.rs
+++ b/plugins/grpc-plugin/src/main.rs
@@ -34,28 +34,29 @@ async fn main() -> Result<()> {
             options::Value::Integer(-1),
             "Which port should the grpc plugin listen for incoming connections?",
         ))
-        .start()
+        .configure()
         .await?;
 
     let bind_port = match plugin.option("grpc-port") {
         Some(options::Value::Integer(-1)) => {
             log::info!("`grpc-port` option is not configured, exiting.");
-            None
+            plugin.disable("`grpc-port` option is not configured.").await?;
+            return Ok(());
         }
-        Some(options::Value::Integer(i)) => Some(i),
+        Some(options::Value::Integer(i)) => i,
         None => return Err(anyhow!("Missing 'grpc-port' option")),
         Some(o) => return Err(anyhow!("grpc-port is not a valid integer: {:?}", o)),
     };
 
-    if let Some(bind_port) = bind_port {
-        let bind_addr: SocketAddr = format!("0.0.0.0:{}", bind_port).parse().unwrap();
+    let plugin = plugin.start().await?;
 
-        tokio::spawn(async move {
-            if let Err(e) = run_interface(bind_addr, state).await {
-                warn!("Error running the grpc interface: {}", e);
-            }
-        });
-    }
+    let bind_addr: SocketAddr = format!("0.0.0.0:{}", bind_port).parse().unwrap();
+
+    tokio::spawn(async move {
+        if let Err(e) = run_interface(bind_addr, state).await {
+            warn!("Error running the grpc interface: {}", e);
+        }
+    });
 
     plugin.join().await
 }
--- a/plugins/grpc-plugin/src/main.rs
+++ b/plugins/grpc-plugin/src/main.rs
@@ -40,19 +40,22 @@ async fn main() -> Result<()> {
     let bind_port = match plugin.option("grpc-port") {
         Some(options::Value::Integer(-1)) => {
             log::info!("`grpc-port` option is not configured, exiting.");
-            return Ok(());
+            None
         }
-        Some(options::Value::Integer(i)) => i,
+        Some(options::Value::Integer(i)) => Some(i),
         None => return Err(anyhow!("Missing 'grpc-port' option")),
         Some(o) => return Err(anyhow!("grpc-port is not a valid integer: {:?}", o)),
     };
-    let bind_addr: SocketAddr = format!("0.0.0.0:{}", bind_port).parse().unwrap();
 
-    tokio::spawn(async move {
-        if let Err(e) = run_interface(bind_addr, state).await {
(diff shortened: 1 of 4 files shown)


rustyrussell:
Hmm, the infra is all there in lightningd, you just need to plumb it thru the rust plugin layer?

In particular, return `{"disable": "Port not specified"}` from the init RPC call.

cdecker:
> Hmm, the infra is all there in lightningd, you just need to plumb it thru the rust plugin layer?
> 
> In particular, return `{"disable": "Port not specified"}` from the init RPC call.

I know, but this is very much part of the public API, so we better make sure it is as idiomatic as possible, so we need to allow hooking API user code into the `init` request/response flow.

I had not considered this builtin=important => disable dependency while developing the API, but I'd hate to break the design to shoehorn it back in in a hurry. This is not the final solution, but it allows us to cleanly build support in while not holding up the release.

cdecker:
Ok, this has gotten me interested, and I have written up a new variant: it introduces a midstate that reads the `init` message to get the options populated, but defers replying to it so we can give control back to the user code.

The old interface still works, and if users don't need to disable they can just go from `Builder` -- `start`--> `Plugin`, while if disabling is required we go `Builder` -- `configure`-> `ConfiguredPlugin` -- `start` --> `Plugin`

plugins/grpc-plugin/src/main.rs
```diff
@@ -34,18 +34,22 @@ async fn main() -> Result<()> {
...
             "Which port should the grpc plugin listen for incoming connections?",
         ))
-        .start()
+        .configure()
         .await?;
 
     let bind_port = match plugin.option("grpc-port") {
         Some(options::Value::Integer(-1)) => {
             log::info!("`grpc-port` option is not configured, exiting.");
+            plugin.disable("`grpc-port` option is not configured.").await?;
```
vincenzopalazzo, line 43: So, if I'm understanding the API, this `disable` method gives you the possibility to change the method init.
Does this mean if in the future core lightning will need another parameter in the response, we need another method?

or a chain of methods because we need to join method payloads in only one result, right?

vincenzopalazzo, line 43: So, if I'm understanding the API, this `disable` method gives you the possibility to change the method init.
Does this mean if in the future core lightning will need another parameter in the response, we need another method?

or a chain of methods because we need to join method payloads in only one result, right?

vincenzopalazzo (approved):
About the solution ACK https://github.com/ElementsProject/lightning/pull/5175/commits/f1613dcdb387db74d4c80b1bc0e8b7367b5ae1bb

Some nit about the API, in particular with the actual design of the API you are still unable to run some code during the init method like init or migrating a DB or run some type of logic.

Do you think that having a solution that gives the possibility to write something like that can be more flexible for the user, and also remove a little bit of mantainence work for the future?

```
let plugin = Builder::new(state.clone(), tokio::io::stdin(), tokio::io::stdout())
        .option(options::ConfigOption::new(
            "grpc-port",
            options::Value::Integer(-1),
            "Which port should the grpc plugin listen for incoming connections?",
        ))
        .on_init(on_init)
        .start()
        .await?;
```

plugins/src/lib.rs
```diff
@@ -350,6 +368,67 @@ where
...
+        let input = self.input;
+        let receiver = self.receiver; // Now reply to the `init` message that `configure` left pending.
+        output
+            .lock()
+            .await
+            .send(json!(
+                {
+                    "jsonrpc": "2.0",
+                    "id": self.init_id,
+            "result": crate::messages::InitResponse{disable: None}
```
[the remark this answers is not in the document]
cdecker: It's more that `fmt` doesn't really like working across a macro which `json!` is.

plugins/grpc-plugin/src/main.rs
cdecker: For now we don't have other things to do, and I don't foresee much else being added to `init`, but your point is valid. This is much more about allowing user code to run after getting the `init` call, but before replying. It is equivalent to a callback, but avoids any limitation regarding context for a callback, and having additional state in the `ConfiguredPlugin` (i.e., the `disabled` field in your proposal), while still giving the ability to disable through a call. Notice that I made the `disable` method consume the `ConfiguredPlugin` making sure a user cannot accidentally call anything else after terminating this way.

cdecker:
> About the solution ACK [f1613dc](https://github.com/ElementsProject/lightning/commit/f1613dcdb387db74d4c80b1bc0e8b7367b5ae1bb)
> 
> Some nit about the API, in particular with the actual design of the API you are still unable to run some code during the init method like init or migrating a DB or run some type of logic.
> 
> Do you think that having a solution that gives the possibility to write something like that can be more flexible for the user, and also remove a little bit of mantainence work for the future?
> 
> ```
> let plugin = Builder::new(state.clone(), tokio::io::stdin(), tokio::io::stdout())
>         .option(options::ConfigOption::new(
>             "grpc-port",
>             options::Value::Integer(-1),
>             "Which port should the grpc plugin listen for incoming connections?",
>         ))
>         .on_init(on_init)
>         .start()
>         .await?;
> ```
> 
> Draft solution at #5173 that needs a split of the start process here

I think using a callback is more limiting in that we'd suddently have to add context to that as well, i.e., yet another struct that is passed in as a generic argument if we wanted to use something other than just the plugin info to make our decision. Splitting the state machine the way we do here means we get to return control to user code, between the `configure` and the `start` method, without shoehorning it into a callback.

rustyrussell (approved):
Ack f1613dcdb387db74d4c80b1bc0e8b7367b5ae1bb

PR #5176, opened 2022-04-09
use full path to reference `lightningd.c`

stephenwebel:
The relative path makes for a difficult experience when people are reading on `https://lightning.readthedocs.io/`. Directly linking saves the reader a few clicks hunting down the correct location :)

Current experience:
image
image


rustyrussell:
Thanks!

PR #5177, opened 2022-04-10
cln-plugin: Handle --help invocations better

cdecker:
We now have ternary outcomes for `Builder.configure()` and
`Builder.start()`:

 - `Ok(Some(p))` means we were configured correctly, and can continue
   with our work normally
 - `Ok(None)` means that `lightningd` was invoked with `--help`, we
   weren't configured (which is not an error since the `lightningd` just
   implicitly told us to shut down) and user code should clean up and
   exit as well
 - `Err(e)` something went wrong, user code may report an error and exit.

The ternary outcome is a bit unwieldy, with two nested `.unwrap()`s,
but I think this is better than exiting outright in case the user has
initialized some resources (which we discourage but can't prevent).

Changelog-None: Unreleased, not adding noise to the changelog

### DIFF
--- a/plugins/examples/cln-plugin-startup.rs
+++ b/plugins/examples/cln-plugin-startup.rs
@@ -6,7 +6,7 @@ use cln_plugin::{options, Builder, Error, Plugin};
 use tokio;
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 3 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5177/commits/e7dcc3fcf60f438546114b854b831db077873225

look like that you fall in a unlikely run with the CI

PR #5180, opened 2022-04-11
Changed external/libwally-core to 0.8.5 release

rustyrussell:
Combined with the following commit which is required to
update against changed libsecp256k1 APIs:

Updated deprecated function calls

This is #5172 squished into a single commit (so it doesn't break bisect!) and removing the extraneous commit.
Changelog-None

### DIFF
--- a/common/blinding.c
+++ b/common/blinding.c
@@ -35,6 +35,6 @@ bool blinding_next_privkey(const struct privkey *e,
 			   struct privkey *next)
 {
 	*next = *e;
-	return secp256k1_ec_privkey_tweak_mul(secp256k1_ctx, next->secret.data,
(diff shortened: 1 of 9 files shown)


vincenzopalazzo (approved):
Looks like the CI fails about a mem leak in one case, and in the other case fails regarding the "deamon died".

But the code change looks unrelated (from reading the stack trace), so for me looks good. 

ACK https://github.com/ElementsProject/lightning/pull/5180/commits/6f3a2162a9a202f683dc2aae3c6b752a7601d26a

rustyrussell:
Closes: #5172

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5180/commits/e7cb4f60379c75b3a17df6b8689362334fcffa17

PR #5198, opened 2022-04-19
connectd: avoid use-after-free on reconnect with remote_addr.

rustyrussell:
I was seeing a strange crash:
	Connectd gave bad CONNECT_PEER_CONNECTED message

The message is indeed mangled, around the remote_addr!
A quick review of the code revealed that we were not making a copy
when it was a reconnect, and so the remote_addr pointer was pointing
to memory which was freed.

Changelog-None


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5198/commits/9744e5f512eeaa914446a38cd3524980898f4874

m-schmoock:
Yep. Good catch.

Issue #5196, opened 2022-04-19
node_announcement propagation has gotten worse!


rustyrussell:
Two examples: my node, which is running v0.11.0rc3+, and IRATEMONKEY, which is the same:

https://1ml.com/node/024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605
https://amboss.space/node/024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605 
https://lightningnetwork.plus/nodes/024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605 

https://1ml.com/node/032de5c0f28f9d7d10c0c0b5ec92e83f9bf40def2bf40181c0f4330c57e58a8605
https://amboss.space/node/032de5c0f28f9d7d10c0c0b5ec92e83f9bf40def2bf40181c0f4330c57e58a8605 
https://lightningnetwork.plus/nodes/032de5c0f28f9d7d10c0c0b5ec92e83f9bf40def2bf40181c0f4330c57e58a8605

You can see these (despite upgrading) are still not having their node_announcements propagating.

rustyrussell:
This seems to be an issue with LND not liking the draft websocket address extension.

carnhofdaki:
Thanks!

jsarenik:
@fiatjaf Could this be related to issues with https://ln.fiatjaf.com/node/032de5c0f28f9d7d10c0c0b5ec92e83f9bf40def2bf40181c0f4330c57e58a8605 ?

Same for the other node: https://ln.fiatjaf.com/node/024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605

PR #5200, opened 2022-04-19
connectd: fix and further enhance our own gossip promotion

rustyrussell:
The #5196 issue had me look harder at the propagation improvements, and doing some real life testing.

This is a larger PR than I would have liked, but the previous propagation "improvements" were woefully under tested.  That's my fault, as I figured they couldn't make things worse (true, but they didn't make things better either!).


cdecker:
ACK 5b98323ad9d8f0fc01b5e187e331d9440ef17719

Will see if CI is just flaky or really has some issues.

cdecker:
Unsurprisingly `test_websocket` needs to be adjusted or skipped now that we no longer announce websocket addresses, otherwise this looks like it'd pass CI.

PR #5206, opened 2022-04-20
update bolt spec regeneration text

instagibbs:
Wasn't clear to me where things were being searched for, or how to re-do regeneration from the text there.

Second commit probably breaks something in github actions, but I felt by default picking the name most new checkouts would use is best? Could be dropped.

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -2,7 +2,7 @@
 
 echo "Running in $(pwd)"
 export ARCH=${ARCH:-64}
-export BOLTDIR=lightning-rfc
+export BOLTDIR=bolts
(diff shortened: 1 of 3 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5206/commits/88b2f81194a1657e0f8d2713b79ad4ac01d373e8

The CI failure is only a unlucky run, caused by the following issue https://github.com/rustyrussell/lnprototest/issues/31

rustyrussell (approved):
Ack 88b2f81194a1657e0f8d2713b79ad4ac01d373e8

Issue #5189, opened 2022-04-16
"ERROR: git is required for generating version information".  Stop.


JCbette:
I re-load the project and meet the problem

> Makefile:7: *** "ERROR: git is required for generating version information".  Stop.

When i use the command `$ sudo make install`

vincenzopalazzo:
what system are you using?

try to run the command `git --version`

You should see some things like that

```
➜  ~ git --version      
git version 2.25.1

```

JCbette:
yes, i get the correct output:

> jc@jc:~/lightning$ git --version
> git version 2.25.1

vincenzopalazzo:
you are on arch linux?

JCbette:
i am on a virtual machine with ubuntu 20.04

vincenzopalazzo:
This is strange, somehow the make file at like 9 doesn't evaluate the `shell git` command.

Not sure what is happening, maybe @rustyrussell or @cdecker have a better idea of what is going on on the make file. I'm not an expert!

JCbette:
Anyway, thank you so much!

ohenrik:
I have the same issue on Ubuntu. Tried to install main branch and v0.10.2, both produced the same error

JCbette:
Did you resolve the problem? Could you please tell me the command please? I have a little don't know what is your mean

ohenrik:
If you are asking me, then no. I'll try again later today, but would really like some help on this if its possible.

JCbette:
> If you are asking me, then no. I'll try again later today, but would really like some help on this if its possible.

All right, good luck to you! 
And keep in contact.

vincenzopalazzo:
Can you try to run the following command

```
git config --global --add safe.directory `pwd`
make && make install
```

JCbette:
It still can't work:

> jc@jc:~/lightning$ git config --global --add safe.directory 'pwd'
> jc@jc:~/lightning$ make && make install
[...]

PR #5210, opened 2022-04-21
makefile: improve error messages if git version failed

vincenzopalazzo:
Fixes following error

`Makefile:7: *** "ERROR: git is required for generating version information". Stop.`

that generate

`fatal: unsafe repository ('/Users/runner/work/lightning/lightning' is owned by someone else)`

Some of users reported that also with `--global` doesn't work, because with `sudo make install`, will take the system conf. 

This PR include a better error message that set the safe directory at global level, but not sure if it is the right think to do!

Fixes https://github.com/ElementsProject/lightning/issues/5189

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c\{0,1\}lightning-v\{0,1\}\([0-9a-f.rc\-]*\)$$|\1|gp')
 
 ifeq ($(VERSION),)
-$(error "ERROR: git is required for generating version information")
+$(error "ERROR: git is required for generating version information. If you have git installed, try to run the following command: git config --system --add safe.directory `pwd`'/*'")
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@
 VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c\{0,1\}lightning-v\{0,1\}\([0-9a-f.rc\-]*\)$$|\1|gp')
 
 ifeq ($(VERSION),)
-$(error "ERROR: git is required for generating version information")
+$(error "ERROR: git is required for generating version information. If you have git installed, try to ran the following command: git config --system --add safe.directory `pwd`'/*'")
```
swaptr, line 7:
```suggestion
$(error "ERROR: git is required for generating version information. If you have git installed, try to run the following command: git config --system --add safe.directory `pwd`'/*'")
```

vincenzopalazzo: Good catch! Thanks!

PR #5211, opened 2022-04-21
lightning-cli plugin start - Assume default relative path

bartobri:
When starting a plugin, if the plugin path cannot be found in
absolute context, assume it is a relative path to the default
plugins dir. As a result, the following now works when my_plugin.py
is installed in the default plugins dir:

lightning-cli plugin start my_plugin.py

Changelog-Added: plugin start RPC subcommand now assumes relative path to default plugins dir if the path is not found in absolute context. i.e. lightning-cli plugin start my_plugin.py

	modified:   plugin_control.c

### DIFF
--- a/lightningd/plugin_control.c
+++ b/lightningd/plugin_control.c
@@ -259,8 +259,7 @@ static struct command_result *json_plugin_control(struct command *cmd,
 		}
 		if (access(plugin_path, X_OK) != 0)
 			plugin_path = path_join(cmd,
-					path_join(cmd, cmd->ld->config_basedir, "plugins/"),
-					plugin_path);
+					cmd->ld->plugins->default_dir, plugin_path);
 		if (access(plugin_path, X_OK) == 0)
 			return plugin_dynamic_start(pcmd, plugin_path,
 						    buffer, mod_params);
--- a/lightningd/plugin_control.c
+++ b/lightningd/plugin_control.c
@@ -1,4 +1,5 @@
 #include "config.h"
+#include <ccan/tal/path/path.h>
 #include <ccan/tal/str/str.h>
 #include <common/json_command.h>
 #include <common/json_tok.h>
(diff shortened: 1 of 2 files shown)


vincenzopalazzo (commented):
Concept ACK https://github.com/ElementsProject/lightning/pull/5211/commits/41409be407ea9889ac013ea7c599387d50022fd4

Why not, I think this is a good shortcut to load the plugin at runtime, I need to test it locally!

cdecker:
Notice that there may be more than one `--plugin-dir`, overriding the default one, so this could be improved to essentially build a `PATH` with all `plugin-dir` that have been specified and search in each of them. Not a blocker, but maybe an improvement for another PR?

rustyrussell:
Yes, I like the idea of searching all plugin dirs.  But it's tricky; I don't think we remember them. Perhaps we should just search each dir we found a previous plugin?

lightningd/plugin_control.c
```diff
@@ -256,6 +257,10 @@ static struct command_result *json_plugin_control(struct command *cmd,
 					json_get_member(buffer, mod_params,
 							"plugin") - 1, 1);
 		}
+		if (access(plugin_path, X_OK) != 0)
+			plugin_path = path_join(cmd,
+					path_join(cmd, cmd->ld->config_basedir, "plugins/"),
+					plugin_path);
```
rustyrussell, line 263: `cmd->ld->plugins->default_dir` is the right thing to use here I think?  

And of course, documenting this in `doc/lightning-plugin.7.md` would be great too!

rustyrussell (changes requested):
Great idea!  Minor cleanup, and some documentation would be great!

Thanks!

lightningd/plugin_control.c
bartobri: Thanks @rustyrussell . I didn't know `cmd->ld->plugins->default_dir` was available. And yes I should have updated the documentation as well. These are both done now. See commit [2181627]
(https://github.com/ElementsProject/lightning/pull/5211/commits/2181627377ecc310ef026f201ec4968d246c79ee).

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5211/commits/2181627377ecc310ef026f201ec4968d246c79ee

rustyrussell (approved):

PR #5212, opened 2022-04-21
Slight problem with `./configure help`

ClayShoaf:
I noticed that the help for each configure option was a little out of sync at the bottom and there was no help text for `--enable/disable-fuzzing`. Not sure on the exact language I should use, because I don't know what fuzzing is in this context.

Changelog-None


cdecker:
Doh, my bad sorry.

ACK https://github.com/ElementsProject/lightning/pull/5212/commits/6ac30c2ee79ca9cd57ac743ff0b371d9334494e3

PR #5216, opened 2022-04-23
msggen: introduce chain of responsibility pattern to make `msggen` extensible

vincenzopalazzo:
This abstraction was easy beceuse @cdecker make a great work

So, when mssgen will be released like a library, the user can create their own generator that extends IGenerator, and put it inside the following configuration

```
def gen_dotnet(generator_chain: GeneratorChain):
    fname = "<your path>"
    dest = open(fname, "w")
    generator_chain.add_generator(DotnetGenerator(dest))

if __name__ == "__main__":
  service = load_jsonrpc_service(schema_dir="<json schema path>")
  generator_chain = GeneratorChain()
  gen_dotnet(generator_chain, meta)
  generator_chain.generate(service)
```

Pro:
- We don't need to trust the user to push code that works in the own repository, but now the user can generate and test the code in the repository that depended on the model that is generated;
- The user can export this generator like the library and make available the work as a python package.
- The user is not coupled with a specific library if any are used, but if there is something wrong for a user in the generator implementation he can reimplement the own.

Cons:
- a new package to maintain, but it is better that maintains X programming language with its own release process

### DIFF
--- a/contrib/msggen/msggen/__main__.py
+++ b/contrib/msggen/msggen/__main__.py
@@ -5,7 +5,7 @@ from msggen.gen.generator import GeneratorChain
 from msggen.utils import repo_root, load_jsonrpc_service
 
 
-def gengrpc(generator_chain: GeneratorChain, meta):
+def add_handler_gen_grpc(generator_chain: GeneratorChain, meta):
     """Load all mapped RPC methods, wrap them in a Service, and split them into messages.
     """
     fname = repo_root() / "cln-grpc" / "proto" / "node.proto"
@@ -22,7 +22,7 @@ def gengrpc(generator_chain: GeneratorChain, meta):
     generator_chain.add_generator(GrpcServerGenerator(dest))
 
 
-def genrustjsonrpc(generator_chain: GeneratorChain):
+def add_handler_gen_rust_jsonrpc(generator_chain: GeneratorChain):
     fname = repo_root() / "cln-rpc" / "src" / "model.rs"
     dest = open(fname, "w")
     generator_chain.add_generator(RustGenerator(dest))
@@ -43,8 +43,8 @@ def run():
     meta = load_msggen_meta()
     generator_chain = GeneratorChain()
 
-    gengrpc(generator_chain, meta)
-    genrustjsonrpc(generator_chain)
+    add_handler_gen_grpc(generator_chain, meta)
+    add_handler_gen_rust_jsonrpc(generator_chain)
 
     generator_chain.generate(service)
 
--- a/contrib/msggen/msggen/__main__.py
+++ b/contrib/msggen/msggen/__main__.py
@@ -1,153 +1,31 @@
-from msggen.model import Method, CompositeField, Service
-from msggen.grpc import GrpcGenerator, GrpcConverterGenerator, GrpcUnconverterGenerator, GrpcServerGenerator
-from msggen.rust import RustGenerator
-from pathlib import Path
-import subprocess
 import json
+from msggen.gen.grpc import GrpcGenerator, GrpcConverterGenerator, GrpcUnconverterGenerator, GrpcServerGenerator
+from msggen.gen.rust import RustGenerator
+from msggen.gen.generator import GeneratorChain
+from msggen.utils import repo_root, load_jsonrpc_service
 
 
-# Sometimes we want to rename a method, due to a name clash
-method_name_override = {
-    "Connect": "ConnectPeer",
-}
-
-
-def repo_root():
-    path = subprocess.check_output(["git", "rev-parse", "--show-toplevel"])
-    return Path(path.strip().decode('UTF-8'))
-
-
-def load_jsonrpc_method(name):
-    """Load a method based on the file naming conventions for the JSON-RPC.
-    """
-    base_path = (repo_root() / "doc" / "schemas").resolve()
-    req_file = base_path / f"{name.lower()}.request.json"
-    resp_file = base_path / f"{name.lower()}.schema.json"
-    request = CompositeField.from_js(json.load(open(req_file)), path=name)
-    response = CompositeField.from_js(json.load(open(resp_file)), path=name)
-
-    # Normalize the method request and response typename so they no
-    # longer conflict.
-    request.typename += "Request"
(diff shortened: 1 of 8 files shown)


contrib/msggen/msggen/__main__.py
```diff
@@ -1,153 +1,31 @@
-from msggen.model import Method, CompositeField, Service
```
joemphilips, line 1: I think there should be a document on how to use this package. Or at least this file is a example of how to use.

contrib/msggen/msggen/gen/grpc.py
```diff
@@ -235,7 +236,7 @@ def generate(self, service: Service) -> None:
             self.generate_message(message)
 
 
-class GrpcConverterGenerator:
+class GrpcConverterGenerator(IGenerator):
     def __init__(self, dest: TextIO):
         self.dest = dest
         self.logger = logging.getLogger("msggen.grpc.GrpcConversionGenerator")
```
joemphilips, line 242: `GrpcConversionGenerator` -> `GrpcConverterGenerator`

contrib/msggen/msggen/gen/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .generator import IGenerator, GeneratorChain # noqa
```
joemphilips, line 1: Aren't these classes which a consumer is going to use should be in a top level `__init__.py` instead of under `gen/` ?
(I'm not very familiar with python packaging, so I might be wrong)

contrib/msggen/msggen/__main__.py
```diff
@@ -1,153 +1,31 @@
...
-        # "help",
-    ]
-    methods = [load_jsonrpc_method(name) for name in method_names]
-    service = Service(name="Node", methods=methods)
-    service.includes = ['primitives.proto']  # Make sure we have the primitives included.
-    return service
-
-
-def gengrpc(service, meta):
+def gengrpc(generator_chain: GeneratorChain, meta):
```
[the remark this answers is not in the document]
vincenzopalazzo: Mh! would be cool, but I didn't change the name, let me ping @cdecker for this

contrib/msggen/msggen/gen/generator.py
```diff
@@ -0,0 +1,36 @@
...
+author: https://github.com/vincenzopalazzo
+"""
+from abc import ABC, abstractmethod
+
+from msggen.model import Service
+
+
+class IGenerator(ABC):
+    """
+    Change of responsibility handler that need to be
```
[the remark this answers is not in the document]
vincenzopalazzo: Good catch, thanks!

contrib/msggen/msggen/gen/generator.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    @abstractmethod
+    def generate(self, service: Service):
+        pass
+
+
+class GeneratorChain:
+    """
+    Chain responsibility patter implementation to generalize
```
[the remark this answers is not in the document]
vincenzopalazzo: This is not a type, is one of my deficits, it is all the life that I continue to write `patter`,

Thanks!

contrib/msggen/msggen/__main__.py
vincenzopalazzo: Good idea! I will add an example dir!

contrib/msggen/msggen/gen/grpc.py
vincenzopalazzo: Same here, not my choice @cdecker

contrib/msggen/msggen/gen/__init__.py
vincenzopalazzo: This is something that I'm understanding too because works always.


BTW,  from https://stackoverflow.com/a/4116384/10854225 the __init__ is not something related to the module but is something related to the python search module.

>If you remove the __init__.py file, Python will no longer look for submodules inside that directory, so attempts to import the module will fail.


I avoided using the class to import in the `__init__.py` because I would like to keep the code readable, and I like work with the following imports

`from msggen.model import Service`

instead of

`from msggen import Service`

IMHO this make the code much cleaner!

contrib/msggen/msggen/gen/grpc.py
joemphilips: What I meant here is not the class name itself, but an argument to the `getLogger`

contrib/msggen/msggen/__main__.py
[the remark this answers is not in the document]
joemphilips: You didn't change the name, but you changed the behavior of the function. So what I meant is that you *should* change the name. Original name by @cdecker is fine in the original context but not anymore.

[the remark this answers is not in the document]
vincenzopalazzo: well, the behavior of the function is not changed, the function configures the generator with the resulting path and the metadata.

But as you preferer, it is not important to me since it is an internal function!

contrib/msggen/msggen/gen/grpc.py
vincenzopalazzo: Yep!

cdecker:
I'm not quite sure what you're trying to achieve here, if we release `msggen` as a library, then it is trivial to extend the existing classes, and write a tiny entrypoint that loads and generates. There is no need to wire everything into a single top-level executable which seems to be what you're doing here (and I have done myself as a shorthand).

Let's keep the refactoring and the creation of a top-level interface, and drop the plugin-like system of adding a generator to a set of generators that are to be executed as a block. That is also much more in line with the Makefile semantics.

vincenzopalazzo:
>I'm not quite sure what you're trying to achieve here, if we release msggen as a library, then it is trivial to extend the existing classes, and write a tiny entrypoint that loads and generates. There is no need to wire everything into a single top-level executable which seems to be what you're doing here (and I have done myself as a shorthand).

The chain of responsibility IMHO is a good addition to include in the library, because can be some cases where we can dependencies between steps.

This pattern can make the code cleaner to read and write imho. However, the same solution is achieved with the hardcoded inside the python script from the user!

PR #5220, opened 2022-04-25
connectd: prefer IPv6 when available

m-schmoock:
This will add all but IPv4 and TOR connections first to the connectd loop.
Adds IPv4 and Tor in a second and third loop. Ensures we stay modern and use
IPv6 when possible. Followed by (more stable) IPv4 connections, followed by
(less stable) TOR connections when nothing else is possible (or allowed).

Rationale:
 - IPv6 should nowadays be the preferred direct connection protocol
 - Makes it more likely to also discover our public IPv6 address using peer `remote_addr` feature. Currently IPv4 was preferred by being first in the array. Since most IPv6 nodes also have a IPv4, we almost never reached the point where we made a IPv6 connection that can reveal our public IPv6 address. Now IPv4 will only be used for non IPv6 nodes.

Changelog-Changed: connectd: prefer IPv6 connections when available.

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -1743,7 +1743,36 @@ static bool wireaddr_int_equals_wireaddr(const struct wireaddr_internal *addr_a,
 	return wireaddr_eq(&addr_a->u.wireaddr, addr_b);
 }
 
-/*~ Orders the addresses which lightningd gave us. */
+/*~ Adds just one address type.
+ *
+ * Ignores deprecated and the `addrhint`. */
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (commented):
Concept ACK

I need to test it locally!

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5220/commits/c1144f0aa16fb0e3c1feff9f8dbab42c96b9179d

m-schmoock:
Closed, will be integrated in a general remote_addr refurbish PR soon

Issue #5221, opened 2022-04-25
sudo make install fails due to git config


carboncls:
### Issue and Steps to Reproduce
Tried to upgrade to v0.11.0 by `git pull` and than `checkout tags/v0.11.0`. I followed the instructions on the installation guide, however, upon `sudo make install`, I get the following error:
Makefile:7: *** "ERROR: git is required for generating version information".  Stop.

`sudo git describe` returns:
fatal: unsafe repository ('/home/carlo/lightning' is owned by someone else)
To add an exception for this directory, call:

    git config --global --add safe.directory /home/carlo/lightning

After `sudo git config --global --add safe.directory /home/carlo/lightning && make install`, output is:
mkdir -p /usr/local/bin
mkdir -p /usr/local/libexec/c-lightning
mkdir -p /usr/local/libexec/c-lightning/plugins
mkdir -p /usr/local/share/man/man1
mkdir -p /usr/local/share/man/man5
mkdir -p /usr/local/share/man/man7
mkdir -p /usr/local/share/man/man8
mkdir -p /usr/local/share/doc/c-lightning
fatal: unsafe repository ('/home/carlo/lightning/external/libwally-core' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /home/carlo/lightning/external/libwally-core
Reinitializing submodules src/secp256k1 ...
fatal: unsafe repository ('/home/carlo/lightning/external/libwally-core' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /home/carlo/lightning/external/libwally-core
fatal: unsafe repository ('/home/carlo/lightning/external/libwally-core' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /home/carlo/lightning/external/libwally-core
make: *** [external/Makefile:53: submodcheck] Error 128

`sudo git config --system --add safe.directory `pwd`'/*'` is the same

### `getinfo` output
   "version": "v0.11.0rc1",
   "blockheight": 733480,
   "network": "bitcoin",
   "msatoshi_fees_collected": 97079903,
   "fees_collected_msat": "97079903msat",
   "lightning-dir": "/home/carlo/.lightning/bitcoin",
   "our_features": {
      "init": "80282a69a2",
      "node": "800080282a69a2",
      "channel": "",
      "invoice": "02000020024100"

rustyrussell:
[...]

PR #5225, opened 2022-04-26
Makefile: stumble along if git does not work.

rustyrussell:
In particular, it's started complaining about "sudo make install"
and the .git directory being owned by someone else :(

Fixes: #5221
Fixes: #5189
Changelog-None

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,6 @@
 # Extract version from git, or if we're from a zipfile, use dirname
 VERSION=$(shell git describe --always --dirty=-modded --abbrev=7 2>/dev/null || pwd | sed -n 's|.*/c\{0,1\}lightning-v\{0,1\}\([0-9a-f.rc\-]*\)$$|\1|gp')
 
-ifeq ($(VERSION),)
-$(error "ERROR: git is required for generating version information")
-endif
-
 # --quiet / -s means quiet, dammit!
 ifeq ($(findstring s,$(word 1, $(MAKEFLAGS))),s)
 ECHO := :
@@ -583,9 +579,17 @@ ALL_PROGRAMS += ccan/ccan/cdump/tools/cdump-enumstr
 # Can't add to ALL_OBJS, as that makes a circular dep.
 ccan/ccan/cdump/tools/cdump-enumstr.o: $(CCAN_HEADERS) Makefile
 
+# Without a working git, you can't generate this file, so assume if it exists
+# it is ok (fixes "sudo make install").
+ifeq ($(VERSION),)
+version_gen.h:
+	echo "ERROR: git is required for generating version information" >&2
+	exit 1
+else
 version_gen.h: $(FORCE)
 	@(echo "#define VERSION \"$(VERSION)\"" && echo "#define BUILD_FEATURES \"$(FEATURES)\"") > $@.new
 	@if cmp $@.new $@ >/dev/null 2>&1; then rm -f $@.new; else mv $@.new $@; $(ECHO) Version updated; fi
+endif
 
 # That forces this rule to be run every time, too.
 header_versions_gen.h: tools/headerversions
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (commented):
LGTM, but I don't think this fixes https://github.com/ElementsProject/lightning/issues/5221 because the error is generated by libwally directory.

Git is working on a glob pattern with `path/*` but this is not supported yet, and we have two choice:

- Use the "*" magic path to allow skip the safety check as the docs say https://git-scm.com/docs/git-config#Documentation/git-config.txt-safedirectory
- or write a script in the tools directory that makes easy the inclusion of this directory.

Why do you think?

cdecker:
I think this should very well fix #5221 as we always suggest first calling `make` as normal user, followed by calling `sudo make install` which then doesn't require compilation, and hence doesn't require updating submodules either.

ACK https://github.com/ElementsProject/lightning/pull/5225/commits/c43c635e90284dd259e5f850618801429c5c6799

vincenzopalazzo:
>I think this should very well fix https://github.com/ElementsProject/lightning/issues/5221 as we always suggest first calling make as normal user, followed by calling sudo make install which then doesn't require compilation, and hence doesn't require updating submodules either.

Mh! @cdecker not sure, but maybe I'm missing something wrong, but in a clean machine, the error should happen also with a normal user, and git at the global level has no lib wally as a safe directory. BTW, I'm sure that I'm missing something!

cdecker:
> > I think this should very well fix #5221 as we always suggest first calling make as normal user, followed by calling sudo make install which then doesn't require compilation, and hence doesn't require updating submodules either.
> 
> Mh! @cdecker not sure, but maybe I'm missing something wrong, but in a clean machine, the error should happen also with a normal user, and git at the global level has no lib wally as a safe directory. BTW, I'm sure that I'm missing something!

No, the `safe.directory` warning originates from a different user than the one executing `git` owning the directory, hence as long as the building user is the owner of the directories this warning will not be raised.

vincenzopalazzo:
With the actual master code

```
(cln-meta-project-_Zy9CvTo-py3.8) gitpod /workspace/lightning (master) $ sudo make install
mkdir -p /usr/local/bin
mkdir -p /usr/local/libexec/c-lightning
mkdir -p /usr/local/libexec/c-lightning/plugins
mkdir -p /usr/local/share/man/man1
mkdir -p /usr/local/share/man/man5
mkdir -p /usr/local/share/man/man7
mkdir -p /usr/local/share/man/man8
mkdir -p /usr/local/share/doc/c-lightning
fatal: unsafe repository ('/workspace/lightning' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /workspace/lightning
Reinitializing submodules external/libsodium external/libwally-core external/gheap external/jsmn external/libbacktrace external/lnprototest ...
fatal: unsafe repository ('/workspace/lightning' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /workspace/lightning
fatal: unsafe repository ('/workspace/lightning' is owned by someone else)
To add an exception for this directory, call:

        git config --global --add safe.directory /workspace/lightning
make: *** [external/Makefile:52: submodcheck] Error 128
(cln-meta-project-_Zy9CvTo-py3.8) gitpod /workspace/lightning (master) $ 
```

git log: 

```
(cln-meta-project-_Zy9CvTo-py3.8) gitpod /workspace/lightning (master) $ git log
commit 6ced5558fb7c817375112072d8f670b878356ddd (HEAD -> master, origin/master, origin/HEAD)
Author: Rusty Russell <rusty@rustcorp.com.au>
Date:   Tue Apr 26 13:25:47 2022 +0930

    Makefile: stumble along if git does not work.
    
    In particular, it's started complaining about "sudo make install"
    and the .git directory being owned by someone else :(
    
    Fixes: #5221
    Fixes: #5189

```

PR #5228, opened 2022-04-28
Add psbt routine for finalizing multisig sigs

ddustin:
Takes partial signatures from a PSBT and builds a witness stack for a multisig & finalizes things.

This is needed for splicing so we can sign the old channel tx over to the new one as part of the splice.


ddustin:
It looks like this code won't be needed after all as there appears to be a simpler, safer approach to building these witness stacks.

cdecker:
Thanks @ddustin :+1:

PR #5233, opened 2022-05-02
Fix broken length on show_log_prefix

nepet:
This PR solves an issue that appears when building cln as a nix derivation with the "--enable-developer" config flag set. In this case the build crashes with the following error:
```
In file included from /nix/store/bd9k3zwm0b3p8k3qw0nlk71vzfa5hn8n-glibc-2.34-115-dev/include/string.h:519,
                 from ccan/ccan/str/str.h:5,
                 from ccan/ccan/tal/tal.h:8,
                 from ccan/ccan/io/io.h:4,
                 from lightningd/log.c:3:
In function 'strncpy',
    inlined from 'show_log_prefix' at lightningd/log.c:617:2:
/nix/store/bd9k3zwm0b3p8k3qw0nlk71vzfa5hn8n-glibc-2.34-115-dev/include/bits/string_fortified.h:95:10: error: '__builtin_strncpy' specified bound 80 equals destination size [8;;https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-Wstringop-truncation-Werror=stringop-truncation8;;]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
make: *** [Makefile:286: lightningd/log.o] Error 1
```

Lowering the max number of chars copied by `strncpy` by 1 (as it is also done by `show_log_level` in line 604) fixes this.

### DIFF
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -601,7 +601,8 @@ static void show_log_level(char buf[OPT_SHOW_LEN], const struct log *log)
 		l = *log->lr->default_print_level;
 	else
 		l = DEFAULT_LOGLEVEL;
-	strncpy(buf, log_level_name(l), OPT_SHOW_LEN-1);
+	strncpy(buf, log_level_name(l), OPT_SHOW_LEN - 1);
+	buf[OPT_SHOW_LEN - 1] = '\0';
 }
 
 static char *arg_log_prefix(const char *arg, struct log *log)
--- a/lightningd/log.c
+++ b/lightningd/log.c
@@ -614,7 +614,8 @@ static char *arg_log_prefix(const char *arg, struct log *log)
 
 static void show_log_prefix(char buf[OPT_SHOW_LEN], const struct log *log)
 {
-	strncpy(buf, log->prefix->prefix, OPT_SHOW_LEN);
+	strncpy(buf, log->prefix->prefix, OPT_SHOW_LEN - 1);
+	buf[OPT_SHOW_LEN - 1] = '\0';
 }
 
 static int signalfds[2];
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -1537,8 +1537,8 @@ static void add_config(struct lightningd *ld,
 		if (opt->desc == opt_hidden) {
 			/* Ignore hidden options (deprecated) */
 		} else if (opt->show) {
-			strcpy(buf + OPT_SHOW_LEN, "...");
 			opt->show(buf, opt->u.carg);
+			strcpy(buf + OPT_SHOW_LEN - 1, "...");
 
 			if (streq(buf, "true") || streq(buf, "false")
 			    || strspn(buf, "0123456789.") == strlen(buf)) {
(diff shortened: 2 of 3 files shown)


vincenzopalazzo (commented):
Looks like untested or the solution is a little bit more complex?

nepet:
It does the trick for me now (just using a build with this commit for testing purposes), but yes, I guess it is more complex than I thought of.

Looking at `char * strncpy ( char * destination, const char * source, size_t num );` in more detail, it states:
> No null-character is implicitly appended at the end of destination if source is longer than num. Thus, in this case, destination shall not be considered a null terminated C string (reading it as such would overflow).

So first of all, the warning is reasonable.

The question now is, how to handle prefixes that extend the size `OPT_SHOW_LEN`.


A possible solution could be to truncate the prefix. This could be achieved on different ways.

  
1. Copy 1 less char, such that the last char remains `\0`, terminating the C string. This is what is provided by the commit and seemingly also done a few line up https://github.com/ElementsProject/lightning/blob/bf80565bc5f958be1d562b15ebd18586d8b0820e/lightningd/log.c#L596-L605


2. Copy exact size and override last char with a terminating `\0`:
```
strncpy(buf, log->prefix->prefix, OPT_SHOW_LEN);
buf[OPT_SHOW_LEN] = '\0';
```
In this case the warning would be too strong.

If one of this possible solutions fits the needs, I can try to localize and align the corresponding tests.

cdecker:
Paging doctor @rustyrussell, we need a C semantics authority in here :-)

lightningd/log.c
```diff
@@ -614,7 +614,7 @@ static char *arg_log_prefix(const char *arg, struct log *log)
 
 static void show_log_prefix(char buf[OPT_SHOW_LEN], const struct log *log)
 {
-	strncpy(buf, log->prefix->prefix, OPT_SHOW_LEN);
+	strncpy(buf, log->prefix->prefix, OPT_SHOW_LEN-1);
```
rustyrussell, line 617: Yes, the usage description in ccan/opt/opt.h says:

```
Similarly, if @show is not NULL, it should be of type "void *show(char *, const type *)".  It should write up to OPT_SHOW_LEN bytes into the first argument; unless it uses the entire OPT_SHOW_LEN bytes it should nul-terminate that buffer.
```

So, this should also set `buf[OPT_SHOW_LEN-1] = '\0';`.

rustyrussell (changes requested):

lightningd/log.c
nepet: https://github.com/ElementsProject/lightning/blob/bf80565bc5f958be1d562b15ebd18586d8b0820e/lightningd/log.c#L596-L605

Thanks for the response, same then for `show_log_level` I guess.

rustyrussell (approved):
Ack 461c99e1338be1f71e3d48a290ab9f47e9c3a024

PR #5236, opened 2022-05-02
pytest: websocket test can tolerate ipv6 address

endothermicdev:
test_connection.py::test_websocket was failing due to the advertisement of an ipv6 address by the peer. test_websocket now validates the websocket address irrespective of additional addresses.

### DIFF
--- a/tests/test_connection.py
+++ b/tests/test_connection.py
@@ -3816,8 +3816,8 @@ def test_websocket(node_factory):
             break
 
     # Check node_announcement has websocket
-    assert (only_one(l2.rpc.listnodes(l1.info['id'])['nodes'])['addresses']
-            == [{'type': 'ipv4', 'address': '127.0.0.1', 'port': port2}, {'type': 'websocket', 'port': ws_port}])
+    ws_address = {'type': 'websocket', 'port': ws_port}
+    assert ws_address in only_one(l2.rpc.listnodes(l1.info['id'])['nodes'])['addresses']
 
 
 @pytest.mark.developer("dev-disconnect required")


m-schmoock:
If its related to IP discovery, this can be disabled by commandline
```
lightningd --disable-ip-discovery
```

Or by `'dev-allow-localhost': None` in the opts for the node_factory

tests/test_connection.py
```diff
@@ -3816,8 +3816,8 @@ def recv(self, maxlen):
             break
 
     # Check node_announcement has websocket
-    assert (only_one(l2.rpc.listnodes(l1.info['id'])['nodes'])['addresses']
-            == [{'type': 'ipv4', 'address': '127.0.0.1', 'port': port2}, {'type': 'websocket', 'port': ws_port}])
+    required_addresses = [{'type': 'ipv4', 'address': '127.0.0.1', 'port': port2}, {'type': 'websocket', 'port': ws_port}]
+    assert all(a in only_one(l2.rpc.listnodes(l1.info['id'])['nodes'])['addresses'] for a in required_addresses)
```
vincenzopalazzo, line 3820:
```suggestion
    ws_address = {'type': 'websocket', 'port': ws_port}
    assert ws_address in only_one(l2.rpc.listnodes(l1.info['id'])['nodes'])['addresses'])
```

As the comment points out, we are checking only the web socket, so maybe we can discard to filter other types of address? without disable the IP discovery?

vincenzopalazzo (changes requested):

tests/test_connection.py
m-schmoock: Yep that would be cleaner. Also @endothermicdev the git message needs a `Changelog-None` as the last line for the PR to be mergable.

endothermicdev: Looks good to me.  I only left the ipv4 loopback in the check to preserve the previous functionality, but I agree it seems unnecessary.

m-schmoock:
ACK 89e0ae4

tests/test_connection.py
endothermicdev: I added a new commit to remove the unmatched paren and insert the changelog.

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5236/commits/44486eb7de7d62458d240458dbb5ba38467164b9

m-schmoock:
@endothermicdev I can't push into your branch. Can you sqash those two commits into just one so we can merge it?

PR #5237, opened 2022-05-03
Accept channels without requiring a confirmation (`option_zeroconf`)

cdecker:
This is a work-in-progress implementation of https://github.com/lightning/bolts/pull/910, without `option_scid_alias`.

Choices:
 - Both `--experimental-zeroconf` and `--zeroconf-allow` or
   `--zeroconf-allow-all` must be set for anything to happen (double
   opt-in).
 - We selectively enable `zeroconf` for individual peers, or we allow
   all at startup via the command line options. This is because the
   negotiation happens at connection `init` handshake, not channel
   establishment, and the fundee does not have a way to opt out if
   they signalled support for it before. This may change sometime in
   the future allowing more dynamic negotiation via plugin hooks.
 - We did not implement `option_scid_alias`, i.e., forcing a peer to
   always refer to the channel using the alias as opposed to just
   using an alias until lockin.
 - As soon as we have a real confirmation we switch over to the real
   `short_channel_id` as if the alias had been reorged out, and
   reconfirmed with the actual funding being confirmed. This keeps
   changes to a minimum.

Still needs a couple more changes:
 - [x] Fix the direct payment, which doesn't see a working channel yet
 - [ ] Use the `alias` in `update_channel` messages relating to the channel until we have a `short_channel_id`
 - [ ] Wire the zeroconf paths with `dualopend`
 - [ ] Compat testing with other implementations

See #5275 for a less pedantic version of this PR

Closes #5275


ZmnSCPxj:
> We selectively enable `zeroconf` for individual peers, or we allow
all at startup via the command line options. This is because the
negotiation happens at connection `init` handshake, not channel
establishment, and the fundee does not have a way to opt out if
they signalled support for it before. This may change sometime in
the future allowing more dynamic negotiation via plugin hooks.

Why not disconnect from the peer if it is connected when you change the `zeroconf` mode for that peer?

Then you can just provide:

    **setzeroconf** *nodeid* [*allow*]

    Set whether or not to accept 0-confirmation channel
    offers from the specified *nodeid*..
    Set *allow* to *true* to allow 0-confirmation channels,
    or *false* to reject 0-confirmation channels.
    By default, 0-confirmation channels are rejected for
    all nodes.

    If the *allow* parameter is not specified, do not change
    the setting, but instead return the current setting.

    If the *allow* parameter is provided and the setting
    was changed, and the peer is currently connected,
    this will disconnect the peer.

This allows users to change the setting dynamically over `lightning-cli` without requiring a plugin, for example to do manual 0-conf channels from a friend or something.

Plugins can still trigger on some kind of early-`connect` hook later, e.g. they can handle the hook by invoking `setzeroconf` before returning with `{"result":"continue"}`.

cdecker:
Following a lengthy discussion with @TheBlueMatt I found that indeed I was interpreting things in a very pedantic way. I'll rework this PR to be more flexible in a new PR that I'll link here as an alternative.

PR #5239, opened 2022-05-04
Update routing graph with rate-limited gossip

endothermicdev:
The gossip rate-limiting feature is beneficial for the broader lightning network as a DoS prevention measure, but the practical reality is that all other implementations currently do little-to-no rate-limiting, choosing to benefit their node by utilizing the latest gossip updates.  This PR lets us continue to squelch outbound gossip that has been rate-limited, but not penalize our own node in the process.

The ideal way to handle rate-limited gossip is unclear.  Options are:
1. Skip the subject channel_update or node_announcement entirely until a new, valid one comes in.
2. Rebroadcast the last message (now outdated) that was not rate-limited.

This PR uses option 1, which modifies current node behavior from option 2.

To Do:

- [x] Determine optimal broadcast behavior when rate-limited gossip is encountered.


gossipd/routing.c
```diff
@@ -1208,6 +1210,7 @@ bool routing_add_channel_update(struct routing_state *rstate,
 	struct unupdated_channel *uc;
 	u8 direction;
 	struct amount_sat sat;
+	bool spam;
```
vincenzopalazzo, line 1213:
```suggestion
	bool spam = false;
```

In this way you avoid to several else branch only for the spam, if the code keep growing, the state can be difficult to track

gossipd/routing.c
```diff
@@ -1313,17 +1316,22 @@ bool routing_add_channel_update(struct routing_state *rstate,
...
+					  "Spammy update for %s/%u flagged"
 					  " (last %u, now %u)",
 					  type_to_string(tmpctx,
 							 struct short_channel_id,
 							 &short_channel_id),
 					  direction,
 					  hc->bcast.timestamp, timestamp);
 			/* Ignoring != failing */
-			return true;
+			//return true;
```
vincenzopalazzo, line 1327: looks like meaningless? with the actual `spam = true`?

tests/test_gossip.py
```diff
@@ -1878,13 +1910,27 @@ def channel_fees(node):
...
+
+    messages = serialize_gossipwith_output(l3)
+    for m in messages:
+        print(m)
+    decoded = subprocess.run(['devtools/decodemsg', messages[-1]],
+                             check=True,
+                             timeout=TIMEOUT,
+                             stdout=subprocess.PIPE).stdout.decode('utf8')
+    print("last gossipwith output:")
+    print(decoded)
```
vincenzopalazzo, line 1931: this can be removed? not sure that without enabling the console in the pytest this information is displayed, maybe I'm wrong?

vincenzopalazzo (commented):
Tomorrow I will look in the CI for now I have only a couple of comments

gossipd/routing.c
endothermicdev: I was trying to be cognizant of the style guide, but it probably makes more sense to do it this way for the reader's sake.

tests/test_gossip.py
endothermicdev: Oops, I have some other cleanup in this test as well.  Rushed this one out hoping to get some feedback on optimal broadcast behavior.  I'm leaning toward sending out old gossip instead of skipping entirely.  It's going to complicate indexing and compacting the gossip_store, so I was hoping to determine that's really the best course of action.

gossipd/routing.c
```diff
@@ -1208,6 +1210,7 @@ bool routing_add_channel_update(struct routing_state *rstate,
 	struct unupdated_channel *uc;
 	u8 direction;
 	struct amount_sat sat;
+	bool spam = false;
```
rustyrussell, line 1215: As a rule, I prefer doing this assignment inside an else.

That way, if there's a jump over the code which sets it (or one is introduced in future!), the compiler will warn about it (at least, with optimization set), rather than it silently being left at false.

It's just a preference though: sometimes there's no other obvious place to set it.

gossipd/routing.c
```diff
@@ -1557,6 +1559,7 @@ bool routing_add_node_announcement(struct routing_state *rstate,
 	u8 alias[32];
 	u8 *features, *addresses;
 	struct tlv_node_ann_tlvs *na_tlv;
+	bool spam = false;
```
rustyrussell, line 1606: here too...

gossipd/routing.c
```diff
@@ -1327,14 +1332,38 @@ bool routing_add_channel_update(struct routing_state *rstate,
 		}
 	}
 
-	chan->half[direction].bcast.timestamp = timestamp;
+	if (!spam){
```
rustyrussell, line 1335: Space before `{` (everywhere)

gossipd/routing.c
```diff
@@ -1356,23 +1385,38 @@ bool routing_add_channel_update(struct routing_state *rstate,
...
 				= gossip_store_add_private_update(rstate->gs,
 								  update);
-		} else
+			/* no need to separately track spam for private
+			 * channels */
+			hc->rgraph.index = hc->bcast.index;
+		} else {
 			hc->bcast.index = index;
+			hc->rgraph.index = index;
+		}
```
rustyrussell, line 1394: These two branches are effectively the same?  `rgraph.index = bcast.index`?

gossipd/routing.c
```diff
@@ -1615,6 +1659,9 @@ bool routing_add_node_announcement(struct routing_state *rstate,
 		return true;
 	}
 
+	if(!node->rgraph.index)
+		node->rgraph.timestamp = timestamp;
```
rustyrussell, line 1663: I don't understand this?

contrib/pyln-client/pyln/client/gossmap.py
```diff
@@ -125,7 +125,12 @@ def _update_channel(self,
                                   fields.get('htlc_maximum_msat', None),
                                   fields['fee_base_msat'],
                                   fields['fee_proportional_millionths'])
-        self.half_channels[direction] = half
+        if (self.half_channels[direction] is None):
+            self.half_channels[direction] = half
+        # gossip_store contains both broadcastable and routing-only updates
+        # Check that we're not replacing a newer update with broadcastable
+        elif (half.timestamp > self.half_channels[direction].timestamp):
+            self.half_channels[direction] = half
```
rustyrussell, line 133: I think we should filter when we enter the gossip store (if we don't already?) such that later messages will always have greater timestamp than earlier ones?  So this logic is simply "override previous".

gossipd/routing.c
endothermicdev: I did it this way first, but was conditionally assigning in several different places. @vincenzopalazzo thought this was cleaner for the reader to understand.  I'm okay either way, but I think he has point.

vincenzopalazzo: Yeah this was my suggestion just because there was several else branch introduces just for the `spam = false` assignment, like

```c
if (..) {
  if (...) {
   spam = true
  } else 
   spam = false
} else 
   spam = false
```

this can introduce confusion if we introduce other branches in the future. However, I always think that @rustyrussell is right with these code-style suggestions, and I'm very curious to know why this "not assigned at declaration time" is a preference.

We can learn something new :)

common/gossmap.c
```diff
@@ -436,6 +436,13 @@ static struct gossmap_chan *add_channel(struct gossmap *map,
 	return chan;
 }
 
+static u32 peek_cupdate_timestamp(struct gossmap *map, size_t cupdate_off){
+	const size_t timestamp_off = cupdate_off + 2 + 64 + 32 + 8;
```
vincenzopalazzo, line 440:
```suggestion
       // what is mean
	const size_t timestamp_off = cupdate_off + 2 + 64 + 32 + 8;
```

We can add a small comment on what these numbers mean? just to simplify the life for people like me that don't know more about this topic

common/gossmap.c
```diff
@@ -436,6 +436,13 @@ static struct gossmap_chan *add_channel(struct gossmap *map,
 	return chan;
 }
 
+static u32 peek_cupdate_timestamp(struct gossmap *map, size_t cupdate_off){
```
vincenzopalazzo, line 439:
```suggestion
static u32 peek_cupdate_timestamp(struct gossmap *map, size_t cupdate_off) {
```

gossipd/gossip_store.c
```diff
@@ -585,20 +587,20 @@ bool gossip_store_compact(struct gossip_store *gs)
 }
 
 u64 gossip_store_add(struct gossip_store *gs, const u8 *gossip_msg,
-		     u32 timestamp, bool push,
+		     u32 timestamp, bool push, bool spam,
```
vincenzopalazzo, line 590:
```suggestion
		     u32 timestamp, bool push, 
		     bool spam, const u8 *addendum)
```

vincenzopalazzo (changes requested):
Concept ACK, I have only a couple of smaller comments.

And also we need the data in the binary like `contrib/pyln-client/tests/data/gossip_store-part1.xz`?

gossipd/routing.c
endothermicdev: Reverting to the uninitialized method.

gossipd/routing.c
endothermicdev: Added comments to indicate this behavior for non-spam messages.

common/gossmap.c
endothermicdev: Quoted the spec here for clarification.

contrib/pyln-client/pyln/client/gossmap.py
endothermicdev: That would normally be the case.  I was concerned with an edge case where we have a normal message with timestamp 1, then a rate-limited message with timestamp 3.  Some time later, our rate-limit clears and we receive a message with timestamp 2.  This could replace the original broadcastable message, but would not supersede the later-timestamped gossip that's still useful for the routing graph.
This edge case wouldn't normally occur if we're well connected and the originating node is always incrementing gossip timestamps, but it is a valid case.  We could ignore messages in this edge case and keep the gossip_store in ascending order.  Alternatively, we could discard message 2 and unflag message 3 (now okay to broadcast also.) Would probably be more useful to delete and rewrite it to the store without the spam flag, but same idea.  I kind of like this behavior now that I think about it.

rustyrussell: Yes!  We should be comparing any incoming against the *latest we have*, spam or no.

Don't worry too much about the case where we might consider something "no longer spam" sometime later, either.

gossipd/routing.c
```diff
@@ -1659,26 +1706,54 @@ bool routing_add_node_announcement(struct routing_state *rstate,
...
+				   	       NULL);
+		if (node->bcast.timestamp > rstate->last_timestamp
+		    && node->bcast.timestamp < time_now().ts.tv_sec)
+			rstate->last_timestamp = node->bcast.timestamp;
+		if (!spam) {
+			node->bcast.index = grgraph;
+		}
+		if (timestamp >= node->rgraph.timestamp) {
+			node->rgraph.index = grgraph;
+		}
```
rustyrussell, line 1767: 8 character / tab indents please!

gossipd/routing.c
```diff
@@ -1659,26 +1706,54 @@ bool routing_add_node_announcement(struct routing_state *rstate,
...
+			if (node->rgraph.index != node->bcast.index){
+				/* Harmless if it was never added */
+				gossip_store_delete(rstate->gs, &node->rgraph,
+						    WIRE_NODE_ANNOUNCEMENT);
+			}
+		}
+		gossip_store_delete(rstate->gs, &node->bcast,
+				    WIRE_NODE_ANNOUNCEMENT);
+	}
+	else{
```
rustyrussell: `} else {`

gossipd/routing.c
rustyrussell: Needs a comment at lesat!

common/gossmap.c
rustyrussell: Pretty sure this commit should be omitted, once we don't allow timestamps to go backwards any more.

gossipd/gossip_store.c
```diff
@@ -587,8 +587,8 @@ bool gossip_store_compact(struct gossip_store *gs)
 }
 
 u64 gossip_store_add(struct gossip_store *gs, const u8 *gossip_msg,
-		     u32 timestamp, bool push, bool spam,
-		     const u8 *addendum)
+		     u32 timestamp, bool push,
+		     bool spam, const u8 *addendum)
 {
```
rustyrussell, line 592: This commit needs to be quashed into previous.  Or, TBH, simply removed as not worthwhile.

rustyrussell (changes requested):
OK, some more simplifications possible.
Some weird commits in the middle, cherry-picked from elsewhere.
Some later commits need to be squashed into earlier ones.
Style is still not following doc/STYLE.md (see https://www.kernel.org/doc/html/v4.10/process/coding-style.html too)

rustyrussell (changes requested):
Minor changes only; this now looks superb!

rustyrussell (approved):

gossipd/routing.c
```diff
@@ -1660,33 +1691,50 @@ bool routing_add_node_announcement(struct routing_state *rstate,
...
+	}
 
-	if (index)
-		node->bcast.index = index;
-	else {
-		node->bcast.index
-			= gossip_store_add(rstate->gs, msg,
-					   node->bcast.timestamp,
+	/* Don't add to the store if it was loaded from the store. */
+	if (index) {
```
niftynei, line 1722: I think this accomplishes the same thing more succinctly?

```
if (!index) {
    index = gossip_store_add ...
    if (node->bcast.timestamp...)
}
    
node->rgraph.index = index;

if (!spam)
    node->bcast.index = index
```

tests/test_gossip.py
```diff
@@ -1902,8 +1903,7 @@ def channel_fees(node):
                              check=True,
                              timeout=TIMEOUT,
                              stdout=subprocess.PIPE).stdout.decode('utf8')
-    # Used in routing graph, but not passed to gossip peers.
-    assert("fee_proportional_millionths=1005" not in decoded)
```
niftynei: maybe check both? (e.g. leave this check in?)

niftynei (commented):
this was fun to review w/ @endothermicdev. learned a lot about how gossip store works.

this should be a great improvement over the existing impact of "spam" on our routing graph 

mostly logic nits/readability things!

contrib/pyln-client/pyln/client/gossmap.py
```diff
@@ -9,7 +9,7 @@
 import struct
 
 # These duplicate constants in lightning/common/gossip_store.h
-GOSSIP_STORE_VERSION = 9
+GOSSIP_STORE_VERSION = 0x0a
```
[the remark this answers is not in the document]
endothermicdev: Good idea! Implemented and tested.

rustyrussell:
Ack d7ec151bf1ba1adceadbb46289134d8079e8d0ba

PR #5242, opened 2022-05-06
rpc: use the standard port derivation in connect command

vincenzopalazzo:
adding the port derivation when the port is not specified in the connect command.

Complete implementation of BOLT1 port derivation proposal https://github.com/lightning/bolts/pull/968

Let's see if some tests in CI will be unhappy


lightningd/connect_control.c
```diff
@@ -140,7 +141,7 @@ static struct command_result *json_connect(struct command *cmd,
 		/* Is there a port? */
 		if (!port) {
 			port = tal(cmd, u32);
-			*port = DEFAULT_PORT;
+			*port = DEFAULT_PORT + chainparams->rpc_port - 8332;
```
cdecker, line 144: Why not add the default port to the `chainparams` instead of this derivation? It'd be clearer where the numbers come from.

vincenzopalazzo: Yeah, I was thinking to move this derivation in some place, but we can have a static file in the chainparams.

Thanks!

rustyrussell:
Second commit fixes up mistakes in the first: please clean that up!

Results look good though!

vincenzopalazzo:
>Second commit fixes up mistakes in the first: please clean that up!

Ops! sorry!

However, in this PR there are a few things that are missing related to the default port in the regtest! I will add this by the end of the week

rustyrussell:
Here's a little more documentation:

```diff
diff --git a/doc/lightning-connect.7.md b/doc/lightning-connect.7.md
index c2e9851f7..ebdc3d8c0 100644
--- a/doc/lightning-connect.7.md
+++ b/doc/lightning-connect.7.md
@@ -18,7 +18,11 @@ be of the form *id@host* or *id@host:port*. In this case, the *host* and
 
 *host* is the peer's hostname or IP address.
 
-If not specified, the *port* defaults to 9735.
+If not specified, the *port* depends on the current network:
+- **bitcoin** mainnet: 9735.
+- bitcoin **testnet**: 19735.
+- bitcoin **signet**: 39735.
+- bitcoin **regtest**: 19846.
 
 If *host* is not specified (or doesn't work), the connection will be attempted to an IP
 belonging to *id* obtained through gossip with other already connected
diff --git a/doc/lightningd-config.5.md b/doc/lightningd-config.5.md
index 5fc07bea9..07388d262 100644
--- a/doc/lightningd-config.5.md
+++ b/doc/lightningd-config.5.md
@@ -370,7 +370,8 @@ network.
 Note that for simple setups, the implicit *autolisten* option does the
 right thing: for the mainnet (bitcoin) network it will try to bind to
 port 9735 on IPv4 and IPv6, and will announce it to peers if it seems
-like a public address.
+like a public address (and other default ports for other networks,
+as described below).
 
 Core Lightning also support IPv4/6 address discovery behind NAT routers.
 If your node detects an new public address, it will update its announcement.
```

vincenzopalazzo:
trivial rebase and docs updated

bitcoin/chainparams.h
```diff
@@ -22,6 +22,14 @@ struct chainparams {
 	const char *bip70_name;
 	const struct bitcoin_blkid genesis_blockhash;
 	const int rpc_port;
+	/**
+	 * BOLT 1: The default TCP port depends on the network used. The most common networks are:
```
rustyrussell, line 26: Hmm, our bolt quote detector is really dumb.  Needs just "BOLT #1:" on a line by itself, rest underneath.

rustyrussell (changes requested):
You also need to get rid of "DEFAULT_PORT" throughout the code (many are in tests: the can have a local definition no problems!).

But this shows some places you missed!

bitcoin/chainparams.h
rustyrussell: Once you do this, I can update BOLTVERSION to c74a3bbcf890799d343c62cb05fcbcdc952a1cf3, or you can do that in a separate commit.

You might also need https://github.com/ElementsProject/lightning/pull/5324

devtools/gossipwith.c
```diff
@@ -21,6 +21,7 @@
 #include <stdio.h>
 #include <wire/peer_wire.h>
 
+#define chainparams_get_ln_port simple_get_ln_port
```
rustyrussell, line 24: You shouldn't need to hack this one: the tool sets chainparams correctly, and so will Just Work, whereas with this change it will always try to connect in 9735, even if they specify --network...

rustyrussell (changes requested):
Looks good, one overzealous change found!

rustyrussell (approved):

rustyrussell:
I fixed the last commit (I was wrong: gossipwith did not set chainparams, which already caused problems, since gossipwith -h would crash!).

vincenzopalazzo:
Thanks! Looks like also that the tests related to the network looks good!

Issue #5234, opened 2022-05-02
IP address discovery can announce IPv6 localhost [::1] (on DEVELOPER builds)


m-schmoock:
### Issue and Steps to Reproduce
Some of my peers report an local `remote_addr` of mine being for example:
```
   lightningd: Peer says it sees our address as: [::1]:44386
```
The current code will then lead to a `node_annoucement` announcing `[::1]:9375`.
Seem to happen on IPv6 only. This is not critical, but ugly for the time being.

Update: Can only on `DEVELOPER` builds. Needs to be fixed anyway

m-schmoock:
If one configures with `--enable-developer` the node can send out local addresses for `remote_addr` (for testing). This is only one part of the explanation, because a node should clearly not fill in `::1` for a connection from a public address.

m-schmoock:
@rustyrussell Its not as bad, turns out this can only happen on developer builds, which mine is.
I just need to change that to to --dev-allow-localhost and patch the tests

`connectd/peer_exchange_initmsg.c   peer_init_received  103ff`

```C
#if DEVELOPER		/* ignore private addresses (non-DEVELOPER builds) */
			if (address_routable(tlvs->remote_addr, true))
#else
			if (address_routable(tlvs->remote_addr, false))
#endif /* DEVELOPER */
```

m-schmoock:
Resolved by #5244

PR #5244, opened 2022-05-08
Polish feature remote_addr  

m-schmoock:
Now that the feature is out in the wild, I see several improvements that can be made:

- [x] Address https://github.com/ElementsProject/lightning/issues/5234
- [x] Add detected addresses to `getinfo` RPC. This was missed in prior PR.
- [x] Add nodeid to log message ...'Peer says it sees our address as xyz'. This is a lot more useful
- [x] Prefer IPv6 when making a connection.
- [ ] Maybe filter out `remote_addr` from peers that cannot be confirmed other peers (see https://github.com/ACINQ/eclair/issues/2256 ). Note: This currently does not produce false positives, but 'interrupts' the detection of the correct address.
- [x] Maybe add `remote_addr` field to `listpeers` RPC.
- [ ] Maybe add method to set announced addresses via RPC call

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -899,3 +899,12 @@ int wireaddr_cmp_type(const struct wireaddr *a,
 		return tal_bytelen(a_wire) - tal_bytelen(b_wire);
 	return cmp;
 }
+
+bool wireaddr_arr_contains(const struct wireaddr *was,
+			   const struct wireaddr *wa)
+{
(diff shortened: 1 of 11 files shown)


rustyrussell (approved):
Ack 36baf23990c1c1531eb64004d5ac72df3b6ca48a

rustyrussell:
Nice work!

Issue #5247, opened 2022-05-09
checkmessage support broken in 0.11 - fails with verify: false when the pubkey is omitted


gdassori:
Hello.
I'm experiencing a strange behavior of the `checkmessage` API.

Once checking with other people seems like I'm the only one in the following situation:

### Issue and Steps to Reproduce

`checkmessage` fails with `{"verified": false}` when the pubkey is omitted, even if the signature is correct.

```
guido@vmi82:~$ lightning-cli checkmessage test_message dhx66hpsyydtuowf9bd3yaijq86xhnwbz4h7kqqkfz1atm5iueru1ee6gfitbnbgntbn8smhb5oheqx8ayz8tih46hwumi9sdjx49kib
{
   "pubkey": "02a8f16a02fee11b151d08a204b0111da8d52db044a085e13d9b50896077865df6",
   "verified": false
}
```

And properly validates if the key is passed:
```
guido@vmi82:~$ lightning-cli checkmessage test_message dhx66hpsyydtuowf9bd3yaijq86xhnwbz4h7kqqkfz1atm5iueru1ee6gfitbnbgntbn8smhb5oheqx8ayz8tih46hwumi9sdjx49kib 02a8f16a02fee11b151d08a204b0111da8d52db044a085e13d9b50896077865df6
{
   "pubkey": "02a8f16a02fee11b151d08a204b0111da8d52db044a085e13d9b50896077865df6",
   "verified": true
}
```

@ShahanaFarooqui may want to know that it obv affects the RTL sign\verify process as well when the backend is Core Lightning.

### Version note:

```
 "version": "v0.11.0.1-10-g2ab2061",
```

gdassori:
Hello Vincenzo, thanks for your time.
Actually, the same message verification has different outcomes. To ensure the verification, I have to add the public key after the very same command.

vincenzopalazzo:
>So, I believed it should work even without passing the public key if the message is signed by the same node. Am I wrong?

No, you are right, but if you are not specified the pub key core lightning uses its own public key, but if you are verifying a message signed by another node you must provide the pubkey.

I answer your question? or maybe I'm missing something?

vincenzopalazzo:
On bruce node

```
[...]

PR #5252, opened 2022-05-11
rpc: checkmessage return an error is the pubkey is not found in the graph

vincenzopalazzo:
Returning a warning message when the pub key is not specified and there is no node in the graph.

We try to help people that use core lightning as a signer and nothing else.

Fixes https://github.com/ElementsProject/lightning/issues/5247

Changelog-Added: rpc: checkmessage return an warning msg when no node is found in the graph and no pubkey is provided.

### DIFF
--- a/doc/lightning-checkmessage.7.md
+++ b/doc/lightning-checkmessage.7.md
@@ -29,13 +29,8 @@ RETURN VALUE
 
 [comment]: # (GENERATE-FROM-SCHEMA-START)
 On success, an object is returned, containing:
-- **verified** (boolean): Whether the signature was valid
-
-If **verified** is *true*:
-  - **pubkey** (pubkey): the *pubkey* parameter, or the pubkey found by looking for known nodes
-
-If **verified** is *false*:
-  - **pubkey** (pubkey): the *pubkey* (if any) which could have signed this; this is usually not useful!
+- **verified** (boolean): whether the signature was valid (always *true*)
+- **pubkey** (pubkey): the *pubkey* parameter, or the pubkey found by looking for known nodes
 
 [comment]: # (GENERATE-FROM-SCHEMA-END)
 
@@ -54,4 +49,4 @@ RESOURCES
 
 Main web site: <https://github.com/ElementsProject/lightning>
 
-[comment]: # ( SHA256STAMP:7dcca1fd1708d93b4a0c9b83955630fc4f551c4ffd452fb866c624c72aeaa44d)
+[comment]: # ( SHA256STAMP:af2feeb4eddafc509dff150ec4b11225618f1cbbea06ef81f6d97a1bece3e94c)
--- a/doc/lightning-checkmessage.7.md
+++ b/doc/lightning-checkmessage.7.md
@@ -49,4 +49,4 @@ RESOURCES
 
 Main web site: <https://github.com/ElementsProject/lightning>
 
-[comment]: # ( SHA256STAMP:af2feeb4eddafc509dff150ec4b11225618f1cbbea06ef81f6d97a1bece3e94c)
+[comment]: # ( SHA256STAMP:733247e44d555f9c480a684ceb30440f4f33daf5755253249b5c7b9269c96e49)
--- a/doc/lightning-checkmessage.7.md
+++ b/doc/lightning-checkmessage.7.md
@@ -20,6 +20,10 @@ known node key (as per *listnodes*), and verification succeeds if it
 matches for any one of them.  Note: this is implemented far more
 efficiently than trying each one, so performance is not a concern.
 
+On failure, an error is returned and core lightning exit with the following error code:
+- -32602: Parameter missed or malformed;
+- 1301: *pubkey* not found in the graph.
+
 RETURN VALUE
 ------------
 
--- a/lightningd/signmessage.c
+++ b/lightningd/signmessage.c
@@ -1,5 +1,6 @@
 #include "config.h"
 #include <common/bech32.h>
+#include <common/configdir.h>
 #include <common/json_command.h>
 #include <common/json_param.h>
 #include <errno.h>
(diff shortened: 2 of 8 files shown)


cdecker:
ACK 0c07b7a

rustyrussell:
Ok, this is obviously confusing for people.

How about we just return *an error* in this case? Because we've failed to verify, we're probably best off failing: we know *nothing useful* about the signature on that case.

vincenzopalazzo:
>How about we just return an error in this case? Because we've failed to verify, we're probably best off failing: we know nothing useful about the signature on that case.

Yeah I agree on this, the idea that I have not implemented is to make the public key a requirement for the command, but I was not able to find a historical reason to say that my idea was a good idea.

So I make the question here, what is the reason to have the pub key optional?

rustyrussell:
Yeah, lnd didn't require the pubkey.  I felt it was too easy to misuse, but that was their API.  That's kind of OK (it means use a pubkey from known nodes), but it should fail more cleanly.

OK, let's change it (with `allow-deprecated-apis=false` for now).

So, it will always return valid=true (deprecated), otherwise it will actually error.  Make sure the error includes the public key they would have needed to use to make the signature valid though: if someone really wants that they can dig it out.

vincenzopalazzo:
>So, it will always return valid=true (deprecated), otherwise it will actually error. Make sure the error includes the public key they would have needed to use to make the signature valid though: if someone really wants that they can dig it out.

Ok, the suggestion is landed in the PR! let wait the CI to make sure I didn't put any regression!

tests/test_misc.py
```diff
@@ -2606,3 +2607,18 @@ def test_datastore_keylist(node_factory):
...
+
+
+def test_checkmessage_pubkey_not_found(node_factory):
+    l1 = node_factory.get_node()
+
+    msg = "testcase to check new rpc error"
+    pubkey = "03be3b0e9992153b1d5a6e1623670b6c3663f72ce6cf2e0dd39c0a373a7de5a3b7"
+    zbase = "d66bqz3qsku5fxtqsi37j11pci47ydxa95iusphutggz9ezaxt56neh77kxe5hyr41kwgkncgiu94p9ecxiexgpgsz8daoq4tw8kj8yx"
+
+    with pytest.raises(RpcError):
```
rustyrussell, line 2619: Make sure you check error matches what you expect, using `match="not found in the graph, expected pubkey is {}".format(pubkey)`!  Otherwise anything could trigger the error....

rustyrussell (changes requested):

lightningd/signmessage.c
```diff
@@ -137,6 +138,12 @@ static void listnodes_done(const char *buffer,
 	if (t)
 		t = json_get_member(buffer, t, "nodes");
 
+	if (!deprecated_apis && (!t || t->size == 0)) {
+		was_pending(command_fail(can->cmd, SIGNMESSAGE_PUBKEY_NOT_FOUND,
+					 "pub key not found in the graph, expected pubkey is %s",
```
rustyrussell, line 143: We can actually include this as structured data, rather than a string (and "pub key" is one word!):

```
	struct json_stream *response;
	response = json_stream_fail(cmd, SIGNMESSAGE_PUBKEY_NOT_FOUND,
					    "pubkey not found in the graph");
	json_add_node_id(response, "claimed_key", &can->id);
	json_object_end(response);
	return command_failed(cmd, response);
```

doc/lightning-checkmessage.7.md
```diff
@@ -20,6 +20,10 @@ known node key (as per *listnodes*), and verification succeeds if it
 matches for any one of them.  Note: this is implemented far more
 efficiently than trying each one, so performance is not a concern.
 
+On failure, an error is returned and core lightning exit with the following error code:
+- -32602: Parameter missed or malformed;
+- 1301: *pubkey* not found in the graph.
+
```
rustyrussell, line 26: Edit the schema file, so that `verified` is show to always be `true`:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "required": [
    "verified",
    "pubkey"
  ],
  "additionalProperties": false,
  "properties": {
    "verified": {
      "type": "boolean",
      "enum": [
        true
      ],
      "description": "Whether the signature was valid"
    },
    "pubkey": {
      "type": "pubkey",
      "description": "the *pubkey* parameter, or the pubkey found by looking for known nodes"
    }
  }
}
```

This means you'll need to disable schema checks in test_signmessage, or update the test?

rustyrussell: (Maybe just update the test, by putting the guts of test_checkmessage_pubkey_not_found in there?

rustyrussell (commented):
Since it needs rebasing anyway, some minor nitpicks :)

lightningd/signmessage.c
vincenzopalazzo: Sound good! the only thing that I did not catch is the way we need to change the function type and return an error?

I assumed that this was only an example and I make a partial application of your suggestion.

Let me know if I'm missing somethings

doc/lightning-checkmessage.7.md
vincenzopalazzo: Mh! Does the actual version of the test make sense? or do you want me to change it?

vincenzopalazzo:
@rustyrussell Apologize for the delay here but I missed your review, just jump here today during a tour of my open PR on cln and found that you were waiting for me!

I rebased and resolved the conflicts

rustyrussell (approved):
Ack 6b8a44da4878e81bd97d0cd1f1e7a359acf4bfac

vincenzopalazzo:
Trivial conflict resolution, and rebase on master

rustyrussell:
> Trivial conflict resolution, and rebase on master

Ack https://github.com/ElementsProject/lightning/commit/8d695ab7735d1907d395671918ac2c10b5ff6268

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5252/commits/fe8c3de9d3c5fe847647b5d6582b79d3fe1ae179

valgrind leak report is fixed in master, we can simply apply this once @cdecker acks!

Issue #5240, opened 2022-05-05
Channels shutting down with "bad reestablish revocation_number"


laanwj:
### Issue and Steps to Reproduce

In the last day, on a fairly large node, three channels have shut down on me with these messages:
```
                     "message": "channeld: sent ERROR bad reestablish revocation_number: 7457 vs 7462"
                     "message": "channeld: sent ERROR bad reestablish revocation_number: 271 vs 513"
                     "message": "channeld: sent ERROR bad reestablish revocation_number: 529 vs 543"
```

I am sure two of them are with c-lightning nodes, probably all three. Two of them are fairly recent channels (opened last week).

I have never restored from a backup, nor shut down c-lightning in an unclean way in the last month at least. Also, maybe important: the closes did not happen upon my daemon starting, but while it was running and the channels were active for at least a few hours.

Using c-lightning master as of 43ace0368598fe905ceb05e7e1fc07ec4057d265.

### `getinfo` output

```
{
   "num_peers": 65,
   "num_pending_channels": 0,
   "num_active_channels": 63,
   "num_inactive_channels": 2,
   "version": "v0.11.0.1-6-g43ace03",
   "blockheight": 734998,
   "network": "bitcoin",
   "our_features": {
      "init": "80282269a2",
      "node": "800080282269a2",
      "channel": "",
      "invoice": "02000020024100"
   }
}
```

Plugins and configuration:
```
    --plugin=$HOME/src/plugins/lnurlp/lnurlp.py \
    --plugin=$HOME/src/plugins/rebalance/rebalance.py \
    --plugin=$HOME/src/plugins/summary/summary.py \
    --plugin=$HOME/src/plugins/prometheus/prometheus.py \
    --experimental-dual-fund \
    --experimental-onion-messages \
    --experimental-offers \
```

Logs before close:
```
2022-05-04T10:47:34.373Z INFO    plugin-bcli: bitcoin-cli initialized and connected to bitcoind. (c-lightning started)
⋮ 
2022-05-04T13:07:13.965Z INFO    [X]-channeld-chan#59650: Peer connection lost
2022-05-04T13:07:13.965Z INFO    [X]-chan#59650: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
[...]

Issue #5235, opened 2022-05-02
bad reestablish revocation_number: x vs y


heyambob:
### Issue and Steps to Reproduce

I might have got it around 5 times already less than a year of running. I don't think nobody tried to cheat nobody in these incidents, but what could have caused this to happen? I think mine always happen with my CLN peers.

I have a wild theory:

I used to use the default CLTV(34). I got tired of stuck htlc so I changed it to 40 to match LND long time ago. Once in a while I would get this forced close. Asked around, couldn't fine explanation. "Ok"  it rarely happens, I thought.

A few days ago I read Alex B recommends 144 CLTV to minimize stuck htlc and tried it. Within 24 hrs, I got several strange forced closes ranging from stuck htlc, the "internal error" from LND nodes, and this "bad reestablish revocation_number: 5496 vs 5559" from one of my CLN peers. 

A few days has passed, dust seemed to settle until I received this from another CLN peer of mine "channeld: sent ERROR bad reestablish revocation_number: 12181 vs 16321" . A lot different in numbers this time.

It's not clear who has 12181 and who has 16321. The log, at least, should provide more info.

I talked with another CLN user, said he just got it with an LND peer.

Hm.. is it possible that most CLN users having the CLTV default to 34 may run into this issue with LND users using 40 ? When I cranked it up to 144, it might have provoked it to happen more often.

### `getinfo` output

v0.11.0.1-2-gf40b6d

heyambob:
update: so yesterday I changed CLTV back to 40, it was fine for almost 24hr until I got forced closes with other 3 of my CLN peers with this revocation_number issue. I'm reverting back to the default(34) before I run out of CLN peers.

If it's not CLTV, another thing that I did recently was upgrading to v0.11.0.1-2-gf40b6d from rc5

svewa:
My node also force closes every few days with this error, indicating one peer is just a few revocation_numbers behind. I talked to some of them, they ran LND and did not have any backup/restore situation.

niftynei:
> It's not clear who has 12181 and who has 16321. The log, at least, should provide more info.

Agreed, this message could be more clear. Here's an interpretation from the [code](https://github.com/ElementsProject/lightning/blob/master/channeld/channeld.c#L2975).

```
bad reestablish revocation_number: 5496 vs 5559
```

[...]

Issue #5254, opened 2022-05-12
lightning_connectd: crash with 'tal_count(peer->subds) == 0'


heyambob:
### Issue and Steps to Reproduce

This is the second day my node crashed with this error. So I guess it will keep happening until this issue is addressed

From the console:
```
lightning_connectd: connectd/connectd.c:1905: peer_conn_closed: Assertion `tal_count(peer->subds) == 0' failed.                                                                      
lightning_connectd: FATAL SIGNAL 6 (version v0.11.0.1-8-g80db867)                                                                                                                    
0x560e62482068 send_backtrace                                                                                                                                                        
        common/daemon.c:33                                                                                                                                                           
0x560e62482110 crashdump                                                                                                                                                             
        common/daemon.c:46                                                                                                                                                           
0x7f64d6ddd55f ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x7f64d6e2a34c ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x7f64d6ddd4b7 ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x7f64d6dc7533 ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x7f64d6dc745b ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x7f64d6dd6115 ???                                                                                                                                                                   
        ???:0                                                                                                                                                                        
0x560e62473b7c peer_conn_closed                                                                                                                                                      
        connectd/connectd.c:1905                                                                                                                                                     
0x560e62478ebb destroy_subd                                                                                                                                                          
        connectd/multiplex.c:1112                                                                                                                                                    
0x560e624da224 notify                                                                                                                                                                
        ccan/ccan/tal/tal.c:240                                                                                                                                                      
0x560e624da71f del_tree                                                                                                                                                              
        ccan/ccan/tal/tal.c:402                                                                                                                                                      
0x560e624da774 del_tree
        ccan/ccan/tal/tal.c:412
0x560e624daab4 tal_free
        ccan/ccan/tal/tal.c:486
0x560e624c9901 io_close
        ccan/ccan/io/io.c:450
0x560e6247878a write_to_subd
        connectd/multiplex.c:957
0x560e624c8b7a next_plan
        ccan/ccan/io/io.c:59
0x560e624c9863 io_do_always
        ccan/ccan/io/io.c:435
0x560e624cb51d handle_always
[...]

PR #5256, opened 2022-05-13
connectd: fix accidental handling of old reconnections.

rustyrussell:
We had multiple reports of channels being unilaterally closed because
it seemed like the peer was sending old revocation numbers.

Turns out, it was actually old reestablish messages!  When we have a
reconnection, we would put the new connection aside, and tell lightningd
to close the current connection: when it did, we would restart
processing of the initial reconnection.

However, we could end up with *multiple* "reconnecting" connections,
while waiting for an existing connection to close.  Though the
connections were long gone, there could still be messages queued
(particularly the channel_reestablish message, which comes early on).

Eventually, a normal reconnection would cause us to process one of
these reconnecting connections, and channeld would see the (perhaps
very old!) messages, and get confused.

(I have a test which triggers this, but it also hangs the connect
 command, due to other issues we will fix in the next release...)

Fixes: #5240
Fixes: #5235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 TODO: Insert version codename, and username of the contributor that named the release.
 -->
 
+## [0.11.1] - 2022-05-13: Simon's Carefully Chosen Release Name II
+
+Single change which fixed a bug introduced in 0.11.0 which could cause
+unwanted unilateral closes (`bad reestablish revocation_number: 0 vs 3`)
+
+### Fixed
+
+ - connectd: make sure we don't keep stale reconnections around. ([#5256])
+ - connectd: fix assert which we could trigger. ([#5256])
+
+[#5256]: https://github.com/ElementsProject/lightning/pull/5256
+
 ## [0.11.0.1] - 2022-04-04: Simon's Carefully Chosen Release Name
 
 This release would have been named by Simon Vrouwe, had he responded to my emails!
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -211,36 +211,29 @@ static void peer_connected_in(struct daemon *daemon,
 	tal_free(connect);
 }
 
-/*~ This is an ad-hoc marshalling structure where we store arguments so we
- * can call peer_connected again. */
-struct peer_reconnected {
-	struct daemon *daemon;
-	struct node_id id;
-	struct wireaddr_internal addr;
-	const struct wireaddr *remote_addr;
-	struct crypto_state cs;
-	const u8 *their_features;
-	bool incoming;
-};
-
 /*~ For simplicity, lightningd only ever deals with a single connection per
  * peer.  So if we already know about a peer, we tell lightning to disconnect
  * the old one and retry once it does. */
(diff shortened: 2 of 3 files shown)


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5256/commits/3fab422747facaad34fd8d32c02d75d2718d550d

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5256/commits/3fab422747facaad34fd8d32c02d75d2718d550d

This should fix also https://github.com/ElementsProject/lightning/issues/5235 just to free all the issue :)

ZmnSCPxj:
Looks OK to merge, just a style nit: the checks in our repo look for `Changelog-whatever` tags in commit messages, but this specific PR puts the changelog entries directly into the `CHANGELOG.md` file.

connectd/connectd.c
```diff
@@ -259,6 +252,14 @@ static struct io_plan *peer_reconnected(struct io_conn *conn,
 
 	status_peer_debug(id, "reconnect");
 
+	/* If we have a previous reconnection, we replace it. */
+	pr = peer_reconnected_htable_get(&daemon->reconnected, id);
+	if (pr) {
+		status_broken("Reconnected AGAIN");
+		peer_reconnected_htable_del(&daemon->reconnected, pr);
```
ZmnSCPxj, line 259: If the previous reconnection already had the destructor installed, would it not delete this twice?  Or delete from `htable` is idempotent?

rustyrussell:
Removed accidentally left BROKEN log, and added another assert removal to fix another report.

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5256/commits/b817e0a3ae5ec75f1f8dcf27f6d54d7a7f445e46

frustating I know, but I think this will fix also https://github.com/ElementsProject/lightning/issues/5254

cdecker:
ACK b817e0a3ae5ec75f1f8dcf27f6d54d7a7f445e46

svewa:
`bad reestablish revocation_number: 0 vs 3` in the changelog is maybe a bad example, as a revocation number of `0` usually indicates use of SCB?

CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+## [0.11.1] - 2022-05-13: Simon's Carefully Chosen Release Name II
+
+Single change which fixed a bug introduced in 0.11.0 which could cause
+unwanted unilateral closes (`bad reestablish revocation_number: 0 vs 3`)
+
+### Fixed
+
+ - connectd: make sure we don't keep stale reconnections around. ([#5256])
+ - connectd: fix assert which we could trigger. ([#5256])
```
vincenzopalazzo, line 19:
```suggestion
 - connectd: fix assert which we could trigger. ([#5254])
```

This should be https://github.com/ElementsProject/lightning/issues/5254 ?

rustyrussell: No, these refer to the PR, not the bug itself...

vincenzopalazzo: Ah damnt! Sorry just assuming wrong stuff

PR #4594, opened 2021-06-11
Complete the JSON schemas.

rustyrussell:
Based on #4585, since that cleans up JSON outputs.

This has been a mammoth effort, but it finally works.  So much documentation cleanups, and JSON output fixes in here.

Glad it's finally done!

(If someone wants to write schemas & auto-gen docs for the rpc hooks, go for it!)

PR #5260, opened 2022-05-14
Restore description of `"reserved"` field for `listfunds`

kristapsk:
It was lost in #4594.


vincenzopalazzo (commented):
This file was missed? Mh this case is also tricky to check, for now

Concept ACK.

kristapsk:
Was originally added in #3764.

rustyrussell:
Ack 1cc7ab0cedeb7edd6219e0c088649b94c39b5c10

Good catch!

PR #5261, opened 2022-05-16
Proper connectd cleanups

rustyrussell:
I did some bandaids for the v0.11.1 release, but this fixes the issues properly, as well as addresses some things I found along the way.

1.  connectd was not allowed to simply disconnect peers, but only when it was told to by lightningd.  This means a complex dance if we e.g. get a reconnection: we have to ask lightningd to tell us to disconnect.   It turns out, however, that it's not that hard to have lightningd clean up when told by connectd, which makes the logic in connectd much simpler!
2. When lightningd wanted to talk to a peer, it would ask connectd to allocate file descriptors for that channel id and return them.  This again made a callback dance and added complexity.  Instead, lightningd now allocates the file descriptors and hands them down for connectd to attach.
3. We don't ever tell connectd to explicitly disconnect (except from the "disconnect" RPC command, or some cases where we can't allocate a subdaemon fd): it does so on send/receive of a warning/error already.  This changed a number of tests which assumed disconnection, but so many tests were having to be reworked already.
4. Unless another connection comes in, or we are explictly told to disconnect, we give 5 seconds grace for any remaining messages to/from subds to drain.  This is much more reliable and friendly.

With this rework, we seem to be reliably able to run the tests I used to locate the issues, so those are included at the end.

But disconnection is no longer instantaneous, many tests had to be fixed (under CI; they pass on my build machine!).

Aspirational list of things this should fix:

Fixes: #5353
Fixes: #5366 
Fixes: #5371


rustyrussell:
Rebased, and fixed python extra whitespace.

rustyrussell:
Rebased on master, slowly fixing new flakes I added!

lightningd/peer_control.c
```diff
@@ -1285,6 +1291,14 @@ void peer_connected(struct lightningd *ld, const u8 *msg)
 		peer = new_peer(ld, 0, &id, &hook_payload->addr,
 				hook_payload->incoming);
 
+	/* We track this, because messages can race between connectd and us.
+	 * For example, we could tell it to attach a subd, but it's actually
+	 * already reconnected: we would tell it again when we read the
+	 * "peer_connected" message, and it would get upset (plus, our first
+	 * subd wouldn't die as expected.  So we echo this back to connectd
+	 * on peer commands, and it knows to ignore if it wrong. */
```
niftynei, line 1310: *it's

lightningd/peer_control.c
```diff
@@ -1284,6 +1290,11 @@ void peer_connected(struct lightningd *ld, const u8 *msg)
 		fatal("Connectd gave bad CONNECT_PEER_CONNECTED message %s",
 		      tal_hex(msg, msg));
 
+	/* When a peer disconnects, we give subds time to clean themselves up
+	 * (this lets connectd ensure they've seen the final messages).  But
+	 * nowe it's reconnected, we've gotta force them out. */
```
niftynei, line 1295: nowe -> now

tests/test_opening.py
```diff
@@ -653,14 +653,15 @@ def test_rbf_reconnect_tx_construct(node_factory, bitcoind, chainparams):
         l1.rpc.connect(l2.info['id'], 'localhost', l2.port)
         with pytest.raises(RpcError):
             l1.rpc.openchannel_bump(chan_id, chan_amount, initpsbt['psbt'])
-        assert l1.rpc.getpeer(l2.info['id']) is not None
+        wait_for(lambda: l1.rpc.getpeer(l2.info['id'])['connected'] is False)
```
niftynei, line 660: :+1:

niftynei (commented):
Read through the changes in detail, didn't have a chance to dig into the causes of the test failures though, which makes me think maybe there's something here that I missed.

Overall, the removal of the flag checks in connectd is quite nice, I found the new control flow a lot easier to reason about.

Bunch of very tiny spelling nits, otherwise lgtm

rustyrussell:
I've removed many of my test workarounds, since I discovered that we needed to be looser in terminating connections: give all the subds, and the peer a chance to drain out if we can.

And applied Lisa's sp fixes!  Me typoe!

rustyrussell:
Rebased on master for static channel backups clash...

rustyrussell:
Rebased on master; mainly so I can see the crashes which seem to be happening!

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5261/commits/76b5fc895b9ea33fa67b1c4dee832a993607649b

PR #5275, opened 2022-05-19
zeroconf: Less pedantic interpretation of lightning/bolts#910

cdecker:
Just yolo `channel_ready` messages whenever we feel like it :-)

Closes #5237


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5275/commits/6d3bf090feae7c7de2b809045e002381a000f8c3

cdecker:
> ACK https://github.com/ElementsProject/lightning/pull/5275/commits/6d3bf090feae7c7de2b809045e002381a000f8c3

Why are you asking what is clearly an incomplete draft of the functionality? 😅

vincenzopalazzo:
>Why are you asking what is clearly an incomplete draft of the functionality? sweat_smile

Was thinking that you were divided the json code in a different PR to divide the work that you are doing in https://github.com/ElementsProject/lightning/pull/5237 I got confused by the commit message :) I ignored the PR title! My bad!

cdecker:
Nope, I always publish PRs as drafts in order to get early feedback and CI runs ensuring that my PR is not requiring refactoring once I have all the features in. I'll mark this as ready to review once it's... ready to review :-)

cdecker:
This is ready for review, and has been interop checked last week at the LNdev meetup :+1:

t-bast:
Great, I've been looking at the various implementations over the past few days, and it seems like the interop test was quite light. I'd like to do more extensive interop testing, is this branch the right way to go for cln? Should I just ignore #5237?

cdecker:
> Great, I've been looking at the various implementations over the past few days, and it seems like the interop test was quite light. I'd like to do more extensive interop testing, is this branch the right way to go for cln? Should I just ignore #5237?
> 
> I've tried to test this against eclair and compiled this branch with experimental features activated. My cln node correctly advertises `zero_conf` and `scid_alias`. However, when I try to open a channel cln rejects it saying it doesn't support the `channel_type`. I've tried the following combinations, all were rejected:
> 
> * `anchor_outputs` + `zero_conf` (public)
> * `anchor_outputs` + `scid_alias` (private)
> * `anchor_outputs` + `scid_alias` + `zero_conf` (private)
> 
> Is there something I need to do to make cln accept them?
> 
> I'd also like to test the case where cln is funder: how do I specify a `channel_type` with `zero_conf` and/or `scid_alias` in `fundchannel`?

Following a discussion with @bluematt, I opted to not add the zeroconf at all to the `channel_type` since from his explanation it was pretty much just there for lnd's use. Signalling support for it in `init` is sufficient, and the channel_type should therefore not include neither the `option_scid_alias` nor `option_zeroconf` from my current reading of the spec.

Notice that #5237 did include  negotiation in the `channel_type`, but was deemed too pedantic by other implementors (and less flexible by me, as it required static configuration to opt into zeroconf and scid_alias based on the connection open, rather than at channel negotiation).

We could just be permissive in what we accept in the `open_channel` message, but I'd prefer not to add a quirks mode in the first place, and rather be explicit in what we allow. In this case: `init` tells us that we support `zeroconf`, and then we just send the `channel_ready` (former `funding_locked`) whenever we feel like it, no need to double-opt-in again via the `open_channel`.

t-bast:
> Signalling support for it in init is sufficient, and the channel_type should therefore not include neither the option_scid_alias nor option_zeroconf from my current reading of the spec.

It's totally fine for us if you don't include the feature bit in the `channel_type` you send. If we want to accept doing 0-conf with you, we'll send `channel_ready` early. However, you shouldn't reject a `channel_type` that sets it (if you support the feature), since the spec says:

```
Each basic type has the following variations allowed:
  - `option_scid_alias` (bit 46)
  - `option_zeroconf` (bit 50)
```

cdecker:
But what's the point of including the options in the channel type, they have no discernible effect and if they had they'd be broken since as we discussed on the spec PR, there is no way for the accepter to have a say in the negotiation.

t-bast:
IIRC it's mostly lnd folks who wanted to have these options added to the `channel_type`. We added it to eclair since the spec is currently leaning that way, but I really don't mind if we remove them, our code was cleaner without them and I agree with you that:

- we'll always advertise `scid_alias` and want to use it for private channels whenever our peer also advertises it (no need for a `channel_type` for that
- we'll accept 0-conf channels based on a whitelist that each node operator is free to configure however they want, so we don't really need it in the `channel_type` either

But if we all agree this is the right way forward, we should amend the spec to remove these options from `channel_type` entirely. @roasbeef @crypt-IQ WDYT?

Roasbeef:
IMO removing it just makes code less uniform: everything we have today depends on the channel type to know what type of channel to open/w. I know we've disagreed on this in the past, but imo, having the channel type means things can fail fast: you try to open a zero conf channel to me, it has the channel type, I reject it. Otherwise, you open the channel to me, pay chain fees, then realize that I didn't actually want to have it be zero conf and you're stuck waiting for a confirmation. In the happy case, both sides echo back the same channel type and have their expectations aligned. 

Re the white list: we do something similar, there's a dynamic predicate/hook over a streaming RPC call that users can use to arbitrarily accept/deny channels or use diff policies/params with a given peer. If this isn't registered, then it'll reject all zero conf channels. 

Re `scid_alias`, we support _upgrading_ channel: so a user can have an unadvertised channel, then run w/ the new flag and we'll send the alias over. 

One reason we interpret zero conf vs zero conf +aliases differently, is that these channels have additional implications w.r.t the gossip layer: the node needs to keep track of both the fake alias and the _real_ one, to ensure that one doesn't unnecessarily leak out (I send a failure back but it has my real scid alias...oops it leaked now). There're also some other edge cases we ran into related to validating a new policy announcement from our channel peer, which may or may not use the real scid.

Crypt-iQ:
I really prefer to keep the channel_type(s). It makes our code readable and keeps negotiation in one function rather than having to special-case logic elsewhere in the funding code. 

If the `scid-alias` chantype is removed in favor of making _every_ channel private if the feature bit is negotiated, then you can't open public channels? Maybe I didn't understand but that seems wrong?

t-bast:
> If the scid-alias chantype is removed in favor of making every channel private if the feature bit is negotiated, then you can't open public channels? Maybe I didn't understand but that seems wrong?

That's not what we meant, we just meant that whenever you open a private channel and the feature bit is set in `init`, we should always use `scid_alias` even if it's not set in the `channel_type`. Another option is to ensure that it's always set in the `channel_type`. But given that implementations disagree on what to put inside their `channel_type`, `eclair` will handle both cases to make sure we maximize usage of `scid_alias`.

It's fine by me to allow `scid_alias` and `zeroconf` bits in the `channel_type` if it helps your implementation, in that case we need cln to at least allow setting those bits (LDK does allow setting them).

Crypt-iQ:
> we should always use `scid_alias` even if it's not set in the `channel_type`

I am not sure what this means, like setting an alias ? I don't see a problem with that then

t-bast:
> I am not sure what this means, like setting an alias ? I don't see a problem with that then

Exactly, that's it, we'll always set an alias for the channel.

cdecker:
My point above was mostly about the `channel_type` being redundant with other messages (for these two cases):

 - `option_zeroconf`: just send the `channel_ready` as soon as you are satisfied that it is secure to do so. Once you sent and received the `channel_ready` you can use the channel. The only advantage of signaling at all is that it allows the node to display some expected number of blocks until the channel will become active, but that's already included in the `min_depth` in the `accept_channel` message. So other than simplifying a specific implementation I don't see the advantage (and we're not designing for ease of implementation but for clarity of protocol)
 - `option_scid_alias` has a very similar thing going with the `announcement_signatures` going, either you want it to be annoucned, then you send the message, or you don't. Only once both sides opted in by exchanging the messages will the channel be announced, and that's the only time we actually have a need for the real scid. Though I'll concede that a routing node might want to have some assurances that they'll eventually be able to announce the channel, which would be incompatible with signalling `option_scid_alias` so they could back out at that point.

That being said, it is a trivial change to ignore the variants and just go ahead, so I won't be pushing back too hard here.

cdecker:
Commits https://github.com/ElementsProject/lightning/pull/5275/commits/513c27b17fbd560191846831607c4a2d8ec4d31f and https://github.com/ElementsProject/lightning/pull/5275/commits/e4207dbc77f1bea582575e8715c47ec18b2ab2aa add the code to ignore variants when verifying whether to accept a `channel_type` or not.

I think this should now allow the case `anchor_outputs` + `zero_conf` (public) that @t-bast was testing before. Notice that we currently do not implement `option_scid_alias` enforcement, even though we should be mostly compatible, but may be leaking the actual scid in invoices.

cdecker:
Only took 7 restarts to get past those flaky tests...
Ready for review now ^^

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -703,7 +703,8 @@ def feerates(self, style, urgent=None, normal=None, slow=None):
 
     def fundchannel(self, node_id, amount, feerate=None, announce=True,
                     minconf=None, utxos=None, push_msat=None, close_to=None,
-                    request_amt=None, compact_lease=None):
+                    request_amt=None, compact_lease=None,
+                    mindepth: Optional[int] = None):
```
rustyrussell, line 707: Note that Changelog-Added line is wrong:

1. : should go before JSON-RPC otherwise I'm not sure we'll parse it correctly?
2. It's called "mindepth" not "minconf".

(BTW, I agree mindepth is the right name: prefer spec language, and it's "minimum_depth" there.)

lightningd/opening_control.c
```diff
@@ -1167,6 +1161,15 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
 	peer->uncommitted_channel->fc = tal_steal(peer->uncommitted_channel, fc);
 	fc->uc = peer->uncommitted_channel;
 
+	/* BOLT #2:
+	 *
+	 * The sender:
+	 *   - SHOULD set `minimum_depth` to a number of blocks it considers
+	 *     reasonable to avoid double-spending of the funding transaction.
+	 */
+	assert(mindepth != NULL);
```
rustyrussell, line 1188: I usually skip assert here, and just rely on the next line crashing :)

common/features.c
```diff
@@ -546,3 +563,14 @@ const char *fmt_featurebits(const tal_t *ctx, const u8 *featurebits)
...
 	return fmt;
 }
+
+struct feature_set *feature_set_dup(const tal_t *ctx,
+				    const struct feature_set *other)
+{
+	struct feature_set *res = tal(ctx, struct feature_set);
+	for (size_t i = 0; i < ARRAY_SIZE(res->bits); i++) {
+		res->bits[i] = tal_dup_arr(res, u8, other->bits[i],
+					   tal_bytelen(other->bits[i]), 0);
```
rustyrussell, line 584: We have a tal_dup_talarr() helper for this now.

common/features.c
```diff
@@ -89,6 +89,23 @@ static const struct feature_style feature_styles[] = {
...
+	 * support it for zeroconf */
+	{ OPT_SCID_ALIAS,
+	  .copy_style = { [INIT_FEATURE] = FEATURE_REPRESENT,
+			  [NODE_ANNOUNCE_FEATURE] = FEATURE_REPRESENT,
+			  [BOLT11_FEATURE] = FEATURE_DONT_REPRESENT,
+			  [CHANNEL_FEATURE] = FEATURE_DONT_REPRESENT} },
+
+	/* Always signal zeroconf availability, we'll enable it only
+	 * for allowed nodes in `open_channel` and check in
+	 * `accept_channel` */
```
rustyrussell, line 102: These comments are wrong now?

channeld/channeld.c
```diff
@@ -3234,9 +3235,12 @@ static void handle_funding_depth(struct peer *peer, const u8 *msg)
 				     peer->next_index[LOCAL],
 				     type_to_string(tmpctx, struct pubkey,
 						    &peer->next_local_per_commit));
-			msg = towire_funding_locked(NULL,
-						    &peer->channel_id,
-						    &peer->next_local_per_commit);
+			struct tlv_funding_locked_tlvs *tlvs =
+			    tlv_funding_locked_tlvs_new(tmpctx);
```
rustyrussell: BTW, NULL works the same as an empty TLV.  Though I suspect you want this allocation for a future patch...

wallet/wallet.c
```diff
@@ -1291,6 +1291,22 @@ static struct channel *wallet_stmt2channel(struct wallet *w, struct db_stmt *stm
 		scid = NULL;
 	}
 
+	if (!db_col_is_null(stmt, "alias_local")) {
+		alias[LOCAL] = tal(tmpctx, struct short_channel_id);
+		if (!db_col_short_channel_id_str(stmt, "alias_local", alias[LOCAL]))
```
rustyrussell, line 1296: Don't use db_col_short_channel_id_str in new code; encoding scids as strings in the db was a mistake.

We could create a new one, or (ab)use `db_col_short_channel_id_arr` which uses the wire encoding in a blob.

wallet/wallet.c
```diff
@@ -1930,7 +1952,18 @@ void wallet_channel_save(struct wallet *w, struct channel *chan)
...
 	db_bind_amount_msat(stmt, 43, &chan->htlc_maximum_msat);
-	db_bind_u64(stmt, 44, chan->dbid);
+
+	if (chan->alias[LOCAL] != NULL)
+		db_bind_short_channel_id(stmt, 44, chan->alias[LOCAL]);
+	else
+		db_bind_null(stmt, 44);
+
+	if (chan->alias[REMOTE] != NULL)
+		db_bind_short_channel_id(stmt, 45, chan->alias[REMOTE]);
```
rustyrussell, line 1962: Hmm, note that db_bind_short_channel_id should be renamed db_bind_short_channel_id_str and similarly deprecated...

lightningd/channel_control.c
```diff
@@ -242,11 +243,13 @@ static void peer_got_funding_locked(struct channel *channel, const u8 *msg)
 	if (!channel_on_funding_locked(channel, &next_per_commitment_point))
 		return;
 
+	if (channel->alias[REMOTE] == NULL && alias_remote != NULL)
+		channel->alias[REMOTE] = tal_steal(channel, alias_remote);
+
```
rustyrussell, line 256: tal_steal(ctx, NULL) is a nop, so this can be simplified.

lightningd/gossip_control.c
```diff
@@ -348,15 +348,26 @@ void tell_gossipd_local_private_channel(struct lightningd *ld,
 					struct amount_sat capacity,
 					const u8 *features)
 {
+	/* Which short_channel_id should we use to refer to this channel when
+	 * creating invoices? */
+	struct short_channel_id *scid;
```
rustyrussell, line 353: const...

lightningd/routehint.c
```diff
@@ -106,6 +106,11 @@ routehint_candidates(const tal_t *ctx,
 
 		/* Check channel is in CHANNELD_NORMAL */
 		candidate.c = find_channel_by_scid(peer, &r->short_channel_id);
+
+		/* Try seeing if we should be using a remote alias instead, */
```
rustyrussell, line 110: This confused me, but then I realized this could happen for local channels which don't have a real scid yet: we will put the alias in the "short_channel_id" field of listpeers.  Which is, maybe, a mistake, but it's also reasonable.

Probably deserves a comment here on why listpeers gave us an alias?

lightningd/channel.c
```diff
@@ -956,3 +956,12 @@ void channel_fail_reconnect(struct channel *channel, const char *fmt, ...)
 	err_and_reconnect(channel, tal_vfmt(tmpctx, fmt, ap), 1);
 	va_end(ap);
 }
+
+struct short_channel_id *channel_scid_or_local_alias(const struct channel *chan)
```
rustyrussell, line 961: Prefer `const struct short_channel_id *` unless that breaks something?

lightningd/pay.c
```diff
@@ -1056,10 +1056,10 @@ send_payment_core(struct lightningd *ld,
 			     group, channel, &hout);
 
 	if (failmsg) {
-		fail = immediate_routing_failure(cmd, ld,
-						 fromwire_peektype(failmsg),
-						 channel->scid,
-						 &channel->peer->id);
+		fail = immediate_routing_failure(
+		    cmd, ld, fromwire_peektype(failmsg),
+		    channel->scid ? channel->scid : channel->alias[LOCAL],
```
rustyrussell, line 1061: Use channel_scid_or_local_alias() here?

lightningd/peer_htlcs.c
```diff
@@ -1069,7 +1069,11 @@ static void htlc_accepted_hook_serialize(struct htlc_accepted_hook_payload *p,
 	json_object_end(s);
 
 	json_object_start(s, "htlc");
-	json_add_short_channel_id(s, "short_channel_id", hin->key.channel->scid);
+	if (hin->key.channel->scid != NULL)
+		json_add_short_channel_id(s, "short_channel_id", hin->key.channel->scid);
+	else
+		json_add_short_channel_id(s, "short_channel_id", hin->key.channel->alias[LOCAL]);
+
```
rustyrussell, line 1076: channel_scid_or_local_alias?

rustyrussell (approved):
All this is missing is the scid_alias feature check for forwarding, but that can be done after this.

Feedback was all minor, don't need re-review!

contrib/pyln-client/pyln/client/lightning.py
cdecker: Good catch, mentally I sometimes made the switch between `minconf` (confirmations required for funds to be considered when funding) and `mindepth` (depth required to consider a funded channel active), and I'm sure I won't be the only one, but agreed that it is more in line with the spec nomenclature.

lightningd/opening_control.c
cdecker: Mhe, I find the assert clearer to root-cause (one less indirection) but can change it either way.

cdecker:
Addressed the comments, thanks @rustyrussell. Now waiting for CI to pass and I'll merge.

Summary of changes between PR versions is [here](https://bot.bitcoinstats.com/ElementsProject/lightning/pull/5275/range_diff/23bf97ff0ea9c12d1a8c4664b7eb8141a3daa47d..5cc5370927fb223a6d9c29761c7bfddd9c60034e)

PR #5278, opened 2022-05-20
Fixes #5276 by using debian bullseye instead of buster

### LINKED ISSUE #5276: Update Debian buster to bullseye  for docker images
Hi!

I am testing elementsproject/lightningd:v0.11.1 docker image with [rust-teos watchtower plugin](https://github.com/talaia-labs/rust-teos/tree/cln-plugin/watchtower-plugin) but it looks like that a compilation from this rust project cannot be executed with the current debian buster image.
Error:
```
cln-wt-test    | /opt/rust-teos/watchtower-plugin: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /opt/rust-teos/watchtower-plugin)
```
I've tested with compiling with rust in debian buster and it won't work but it will on bullseye, so a "simple" (maybe not) upgrade of the CLN docker images to bullseye might fix a lot of issues with plugins. Another good point to upgrade the images is that debian buster will be EOL on 2022-08
1. Compile [rust-teos watchtower plugin](https://github.com/talaia-labs/rust-teos/tree/cln-plugin/watchtower-plugin) with Rust official docker image.
2. Copy the content to lightningd docker image and set it as as plugin
3. Launch lightningd and receive the error mentioned above.

cc @sr-gi 

Thanks for all your work!

Jossec101:
This fixes #5276 

I've tested the upgrade with plugins that requires GLIBC 2.29+ when compiled with Rust and it works, shall I test something more for this PR?

This is is my first contribution here, guidance is very welcome

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@
 # * final: Copy the binaries required at runtime
 # The resulting image uploaded to dockerhub will only contain what is needed for runtime.
 # From the root of the repository, run "docker build -t yourimage:yourtag ."
-FROM debian:buster-slim as downloader
+FROM debian:bullseye-slim as downloader
 
 RUN set -ex \
 	&& apt-get update \
@@ -45,7 +45,7 @@ RUN mkdir /opt/litecoin && cd /opt/litecoin \
     && tar -xzvf litecoin.tar.gz $BD/litecoin-cli --strip-components=1 --exclude=*-qt \
     && rm litecoin.tar.gz
 
-FROM debian:buster-slim as builder
+FROM debian:bullseye-slim as builder
 
 ENV LIGHTNINGD_VERSION=master
 RUN apt-get update -qq && \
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
LGTM, looks like the only way to update the library is to update the base docker image, and have error has gone good because unlock the possibility to use rust lib provided in a cool way too!


ACK https://github.com/ElementsProject/lightning/pull/5278/commits/1c618282d19125d603b3b0bae308dc66b3a316a5

Your PRs are fine, the only optional thing is the Changelog-None entry in the commit body!

Jossec101:
I added it as Changelog-Fixed:  Upgrade docker base image from Debian buster to bullseye to work with glibc 2.29+

cdecker:
This needs squashing, and the `Changelog-Fixed` should not be on the first line, rather on the bottom so the commit retains a human-readable commit message.

cdecker:
> @cdecker Squashing done, hope it is ok now.

Perfect 👍 Thank you very much

PR #5279, opened 2022-05-20
cln-plugin: Save "configuration" from "init" method

justinmoon:
`cln-plugin` throws away the "configuration" field from the "init" message.

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -3,7 +3,7 @@ pub use anyhow::{anyhow, Context};
 use futures::sink::SinkExt;
 extern crate log;
 use log::trace;
-use serde::Deserialize;
+use messages::Configuration;
 use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
@@ -210,7 +210,9 @@ where
         let plugin = Plugin {
             state: self.state,
             options: self.options,
-            configuration: self.configuration.unwrap(), // OK to unwrap, set in handle_init
+            configuration: self
+                .configuration
+                .ok_or(anyhow!("Plugin configuration missing"))?,
             wait_handle,
             sender,
         };
@@ -368,7 +370,9 @@ where
 {
     /// The state gets cloned for each request
     state: S,
+    /// "options" field of "init" message sent by cln
     options: Vec<ConfigOption>,
+    /// "configuration" field of "init" message sent by cln
     configuration: Configuration,
     /// A signal that allows us to wait on the plugin's shutdown.
     wait_handle: tokio::sync::broadcast::Sender<()>,
@@ -660,12 +664,6 @@ where
             .await
             .context("error waiting for shutdown")
     }
-}
-
-#[derive(Clone, Debug, Deserialize)]
-pub struct Configuration {
-    #[serde(rename = "lightning-dir")]
-    pub lightning_dir: String,
 }
 
 #[cfg(test)]
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -3,6 +3,7 @@ pub use anyhow::{anyhow, Context};
 use futures::sink::SinkExt;
 extern crate log;
 use log::trace;
+use serde::Deserialize;
 use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
@@ -44,6 +45,7 @@ where
 
     hooks: HashMap<String, Hook<S>>,
     options: Vec<ConfigOption>,
+    configuration: Option<Configuration>,
     rpcmethods: HashMap<String, RpcMethod<S>>,
     subscriptions: HashMap<String, Subscription<S>>,
 }
@@ -62,6 +64,7 @@ where
             hooks: HashMap::new(),
             subscriptions: HashMap::new(),
             options: vec![],
+            configuration: None,
             rpcmethods: HashMap::new(),
         }
     }
@@ -207,6 +210,7 @@ where
         let plugin = Plugin {
             state: self.state,
             options: self.options,
+            configuration: self.configuration.unwrap(), // OK to unwrap, set in handle_init
             wait_handle,
             sender,
         };
@@ -297,6 +301,8 @@ where
             }
         }
 
+        self.configuration = Some(call.configuration);
+
         Ok(messages::InitResponse::default())
     }
 }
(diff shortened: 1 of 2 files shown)


plugins/src/messages.rs
```diff
@@ -63,6 +64,8 @@ pub struct GetManifestCall {}
 #[derive(Deserialize, Debug)]
 pub(crate) struct InitCall {
     pub(crate) options: HashMap<String, Value>,
+    pub(crate) configuration: Value,
+    // pub(crate) configuration: Configuration,
```
justinmoon, line 68: If I try to put `Configuration` in here, then [this match](https://github.com/ElementsProject/lightning/blob/master/plugins/src/lib.rs#L188) isn't triggered. In principle it seems like this should work if everything is `Deserialize` ...

justinmoon:
Added a unittest as separate commit to make sure the "init" message (used [this example message](https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#the-init-method)) can be parsed. Not sure if this is helpful or not.

plugins/src/messages.rs
```diff
@@ -63,6 +64,7 @@ pub struct GetManifestCall {}
 #[derive(Deserialize, Debug)]
 pub(crate) struct InitCall {
     pub(crate) options: HashMap<String, Value>,
+    pub(crate) configuration: Configuration,
```
justinmoon, line 67: Would you prefer I just use `Value` here?

plugins/src/lib.rs
```diff
@@ -653,6 +662,12 @@ where
     }
 }
 
+#[derive(Clone, Debug, Deserialize)]
+pub struct Configuration {
+    #[serde(rename = "lightning-dir")]
+    pub lightning_dir: String,
+}
```
justinmoon, line 669: If we keep this struct, what else should go in it?

plugins/src/messages.rs
vincenzopalazzo: Maybe an HasMap<String, Value> or keep the Configuration here! But if I'm not wrong @cdecker wants this library strongly typed, perhaps the struct is better?

plugins/src/lib.rs
vincenzopalazzo: Here the docs https://lightning.readthedocs.io/PLUGINS.html#the-init-method and if you want to do a copy and paste https://github.com/laanwj/rust-clightning-rpc/blob/92912c8a5318624b2c9684676a0f697b4a38432e/plugin/src/commands/types.rs#L25-L51

plugins/src/lib.rs
```diff
@@ -363,7 +369,7 @@ where
     /// The state gets cloned for each request
     state: S,
     options: Vec<ConfigOption>,
-
+    configuration: Configuration,
```
vincenzopalazzo, line 372:
```suggestion
    /// The cln configuration received after the init msg
    configuration: Configuration,
```

vincenzopalazzo (changes requested):
LGTM, just take a decision if we want use an HasMap or a Struct for the configuration

cdecker:
> overall LGTM, just take a decision if we want use an HasMap or a Struct for the configuration

As you mentioned above making these things we have control over as strongly typed as possible would be a good idea. So I'd vote in favor of the `Configuration` struct 👍

plugins/src/lib.rs
justinmoon: I commented this key and the "options" key

plugins/src/messages.rs
justinmoon: Went with "configuration". I prefer that too.

justinmoon:
Added the rest of the fields to `Configuration` mostly copying from [rust-clightning-rpc](https://github.com/laanwj/rust-clightning-rpc/blob/92912c8a5318624b2c9684676a0f697b4a38432e/plugin/src/commands/types.rs#L25-L51) as suggested by @vincenzopalazzo. I prefer the strongly typed struct too.

vincenzopalazzo:
It is missing the Changelog in the body commit, maybe `Changelog-Added: cln_plugin: take into consideration the cln configuration from init msg` can be a good description?

justinmoon:
Added changelog entry @vincenzopalazzo

rustyrussell:
BTW I've invited you to project as a reader. This should mean CI runs for you...

plugins/src/messages.rs
```diff
@@ -63,6 +63,30 @@ pub struct GetManifestCall {}
...
+    pub proxy: ProxyInfo,
+    #[serde(rename = "torv3-enabled")]
+    pub torv3_enabled: bool,
+    pub always_use_proxy: bool,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct ProxyInfo {
+    #[serde(alias = "type")]
+    pub tup: String,
```
vincenzopalazzo, line 87:
```suggestion
    pub typ: String,
```

This is my typo, sorry!

vincenzopalazzo (changes requested):
Only a small comment, and LGTM.

With the next commit, we will see with the CI will be happy!

plugins/src/messages.rs
justinmoon: Done

justinmoon:
The tests seem to be failing because [this match is failing](https://github.com/ElementsProject/lightning/blob/9a880a0932e7d8575620a1f7b48038c52b4b126f/plugins/src/lib.rs#L187) with my `Configuration` struct. But AFAIKT [my unittest does exactly the same match](https://github.com/ElementsProject/lightning/pull/5279/files#diff-28222d09a6ae0658f211b675fe20ca53e5fde769dcfeef8e3b181123e26951b2R205). 

In my unittest I'm using [exactly this JSON](https://github.com/ElementsProject/lightning/blob/master/doc/PLUGINS.md#the-init-method) and it passes. Why is it failing in the python test in CI?

It does seem that the example JSON in the link above is different from what the CI test is sending. For example, the `always_use_proxy` value, for example, is [missing in CI](https://github.com/ElementsProject/lightning/runs/6965889120?check_suite_focus=true#step:5:2902).

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5279/commits/9e912f85666a7ec8929544342e6952e58b1352a0

PR #5281, opened 2022-05-22
Dealing with flakes so far

rustyrussell:
This removes flaky, and fixes the resulting issues (so far).

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyln/testing/utils.py
+++ b/contrib/pyln-testing/pyln/testing/utils.py
@@ -174,13 +174,20 @@ class TailableProc(object):
     tail the processes and react to their output.
     """
 
-    def __init__(self, outputDir=None, verbose=True):
+    def __init__(self, outputDir, verbose=True):
         self.logs = []
-        self.logs_cond = threading.Condition(threading.RLock())
         self.env = os.environ.copy()
-        self.running = False
         self.proc = None
         self.outputDir = outputDir
+        if not os.path.exists(outputDir):
+            os.makedirs(outputDir)
+        # Create and open them.
+        self.stdout_filename = os.path.join(outputDir, "log")
+        self.stderr_filename = os.path.join(outputDir, "errlog")
+        self.stdout_write = open(self.stdout_filename, "wt")
+        self.stderr_write = open(self.stderr_filename, "wt")
+        self.stdout_read = open(self.stdout_filename, "rt")
+        self.stderr_read = open(self.stderr_filename, "rt")
         self.logsearch_start = 0
         self.err_logs = []
         self.prefix = ""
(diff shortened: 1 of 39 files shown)


vincenzopalazzo (approved):
utACK https://github.com/ElementsProject/lightning/pull/5281/commits/a255e12b6f6a61281c226c0f1294aefe58f5f34e

I have also ready a lnprototest update to remove some coop-test that are disabled right now in the code base.

contrib/pyln-testing/pyln/testing/utils.py
```diff
@@ -146,6 +146,27 @@ def get_tx_p2wsh_outnum(bitcoind, tx, amount):
     return None
 
 
+unused_port_lock = threading.Lock()
+unused_port_set = set()
+
+
+def reserve_unused_port():
```
cdecker, line 153: One alternative thing we could do is to internalize the `ephemeral_port_reserve` lib (it's just a couple of lines), modify the reservation code not to set `REUSEADDR` itself, so it won't re-reserve a port that is in state `TIME_WAIT`, and then have `lightningd` itself leave the port in state `TIME_WAIT` on shutdown. As long as the time a node is down isn't longer than the `/proc/sys/net/ipv4/tcp_fin_timeout` we should not get the same port when calling `reserve` again.

However this method here is likely just as good, as long as we remember not to use `ephemeral_port_reserve.reserve` directly.

lightningd/chaintopology.c
```diff
@@ -1107,7 +1109,9 @@ void setup_topology(struct chain_topology *topo,
 	start_fee_estimate(topo);
 
 	/* Once it gets initial block, it calls io_break() and we return. */
-	io_loop_with_timers(topo->ld);
+	ret = io_loop_with_timers(topo->ld);
+	assert(ret == topo);
+	log_debug(topo->ld->log, "io_loop_with_timers: %s", __func__);
```
cdecker, line 1114: I assume this is ok since a failing `assert` would include `__fund__` anyway.

cdecker (approved):
Very nice, thanks for these fixes :+1:

ACK https://github.com/ElementsProject/lightning/pull/5281/commits/a255e12b6f6a61281c226c0f1294aefe58f5f34e

rustyrussell:
Rebased to fix typo...

rustyrussell:
Rebased on master again: pushing caused new flakes due to updated master.

Ack https://github.com/ElementsProject/lightning/commit/b83eae5a7de503c52e81a9cb357b86159929db2a

rustyrussell:
Rebase again, and two more gossip flakes...

vincenzopalazzo:
Good Job! Thanks to do that!

PR #5285, opened 2022-05-23
Add new request schemas

joemphilips:
Since I wanted to use `fundchannel` and `decodepay` RPC in my plugin.
Some schemas are probably not as strict as it can be, but I think it is good enough for starting point.
(IMO Loose schema is better than nothing since `msggen` will be able to generate methods/DTOs, those are helpful for exploratory programming.)


joemphilips:
I found a way to avoid the problem in my side. Closing this PR for now.

louneskmt:
Hey! Why did you close this PR? Seems pretty useful to have access to these methods from gRPC.

joemphilips:
@louneskmt
1. I found a way to avoid using RPCs (`decodepay`, `fundchannel`) for my own project. So this change was not necessary anymore for my own sake.
2. I forgot the detail, but these RPCs had some miner issues for passing the CI.

So I decided to close it for now.
I agree that these changes are overall fine and useful, you can take after me if you want to.

PR #5287, opened 2022-05-24
Collaborative transaction building

ddustin:
Takes @niftynei's dualopen collaborative transaction building and makes it available for other daemons to use.

This is to make it available for channeld to do splicing and also other code that may want to us it in the future!

### DIFF
--- a/common/interactivetx.c
+++ b/common/interactivetx.c
@@ -1,0 +1,735 @@
+#include "config.h"
+#include <bitcoin/chainparams.h>
+#include <bitcoin/psbt.h>
+#include <bitcoin/script.h>
+#include <ccan/array_size/array_size.h>
+#include <ccan/cast/cast.h>
+#include <ccan/mem/mem.h>
+#include <ccan/tal/str/str.h>
+#include <common/billboard.h>
+#include <common/blockheight_states.h>
+#include <common/crypto_sync.h>
+#include <common/gossip_rcvd_filter.h>
+#include <common/gossip_store.h>
+#include <common/initial_channel.h>
+#include <common/interactivetx.h>
+#include <common/lease_rates.h>
+#include <common/memleak.h>
+#include <common/peer_billboard.h>
+#include <common/peer_failed.h>
+#include <common/psbt_internal.h>
+#include <common/psbt_open.h>
+#include <common/read_peer_msg.h>
+#include <common/setup.h>
+#include <common/status.h>
+#include <common/subdaemon.h>
+#include <common/type_to_string.h>
+#include <common/wire_error.h>
+
+/*
+ * BOLT-f53ca2301232db780843e894f55d95d512f297f9 #2:
+ * The receiving node: ...
+ * - MUST fail the negotiation if: ...
+ *  - if has received 4096 `tx_add_input` messages during this negotiation
+ */
+#define MAX_TX_ADD_INPUT_MSG_RCVD 4096
+/*
+ * BOLT-f53ca2301232db780843e894f55d95d512f297f9 #2:
+ * The receiving node: ...
+ * - MUST fail the negotiation if: ...
+ *  - it has received 4096 `tx_add_output` messages during this negotiation
+ */
+#define MAX_TX_ADD_OUTPUT_MSG_RCVD 4096
+
+/*
+ * BOLT-f53ca2301232db780843e894f55d95d512f297f9 #2:
+ * The receiving node: ...
+ * - MUST fail the negotiation if: ...
+ *  - there are more than 252 inputs
+ *  - there are more than 252 outputs
+ */
+#define MAX_FUNDING_INPUTS 252
+#define MAX_FUNDING_OUTPUTS 252
+
+static struct wally_psbt *default_next_update(const tal_t *ctx,
+					      struct interactivetx_context *ictx)
+{
+	return ictx->desired_psbt;
+}
+
+struct interactivetx_context *new_interactivetx_context(const tal_t *ctx,
+							enum tx_role our_role,
+							struct per_peer_state *pps,
+							struct channel_id channel_id)
+{
+	struct interactivetx_context *ictx = tal(ctx, struct interactivetx_context);
+
+	ictx->ctx = NULL;
+	ictx->our_role = our_role;
+	ictx->pps = pps;
+	ictx->channel_id = channel_id;
+	ictx->tx_add_input_count = 0;
+	ictx->tx_add_output_count = 0;
+	ictx->next_update = default_next_update;
+	ictx->current_psbt = NULL;
+	ictx->desired_psbt = NULL;
+	ictx->pause_when_complete = false;
+	ictx->change_set = NULL;
+
+	return ictx;
+}
+
+static bool is_segwit_output(const tal_t *ctx,
+			     struct wally_tx_output *output,
+			     const u8 *redeemscript)
+{
+	const u8 *wit_prog;
+	if (tal_bytelen(redeemscript) > 0)
+		wit_prog = redeemscript;
+	else
+		wit_prog = wally_tx_output_get_script(ctx, output);
+
+	return is_p2wsh(wit_prog, NULL) || is_p2wpkh(wit_prog, NULL);
+}
+
+/* Return first non-handled message or NULL if connection is aborted */
+static u8 *read_next_msg(const tal_t *ctx,
+			 struct interactivetx_context *state,
+			 char **error)
+{
+	u8 *msg = NULL;
+
+	for (;;) {
+		char *desc;
+		bool warning;
+		struct channel_id actual;
+		enum peer_wire t;
+		bool from_gossipd;
+
+		/* Prevent runaway memory usage from many messages */
+		if (msg)
+			tal_free(msg);
+
+		/* This helper routine polls the peer. */
+		msg = peer_or_gossip_sync_read(ctx, state->pps, &from_gossipd);
+
+		/* Line should be in STFU mode and not receiving gossip */
+		if (from_gossipd) {
+			*error = tal_fmt(ctx, "interactivetx got gossip but"
+					 " should be in STFU mode.");
+
+			tal_free(msg);
+			/* Return NULL so caller knows to stop negotiating. */
+			return NULL;
+		}
+
+		/* BOLT #1:
+		 *
+		 * A receiving node:
+		 *   - upon receiving a message of _odd_, unknown type:
+		 *     - MUST ignore the received message.
+		 */
+		if (is_unknown_msg_discardable(msg))
+			continue;
+
+		/* A helper which decodes an error. */
+		if (is_peer_error(msg, msg, &state->channel_id,
+				  &desc, &warning)) {
+			/* In this case, is_peer_error returns true, but sets
+			 * desc to NULL */
+			if (!desc)
+				continue;
+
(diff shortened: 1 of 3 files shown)


common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+		goto tx_complete;
+
+	if(tal_count(set->added_ins) != 0) {
+
+		const struct input_set *in = &set->added_ins[0];
+		struct bitcoin_outpoint outpoint;
+		u8 *prevtx;
+
+		if (!psbt_get_serial_id(&in->input.unknowns, &serial_id))
+			abort();
```
niftynei, line 202: can't decide if a hard fail is ok or if we should return an error? it's not recoverable, but maybe crashing isn't the best policy?

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+					  NULL);
+
+		tal_arr_remove(&set->added_ins, 0);
+
+	}
+	else if (tal_count(set->rm_ins) != 0) {
+
+		if (!psbt_get_serial_id(&set->rm_ins[0].input.unknowns,
+					&serial_id))
+			abort();
```
niftynei, line 228: same comment as above about abort vs error?

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+		next_psbt = ictx->desired_psbt;
+
+	if(!next_psbt)
+		next_psbt = ictx->current_psbt;
+
+	ictx->change_set = psbt_get_changeset(ctx,
+					     ictx->current_psbt,
+					     next_psbt);
+
+	if(ictx->current_psbt != next_psbt) {
```
niftynei, line 328: this compares pointers right? do we actually want to know if they're the same PSBT or is just a pointer check good enough

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+					       tal_hex(ctx, msg));
+
+			/*
+			 * BOLT-f53ca2301232db780843e894f55d95d512f297f9 #2:
+			 * The receiving node: ...
+			 * - MUST fail the negotiation if: ...
+			 *   - it has received 4096 `tx_add_output`
+			 *   messages during this negotiation
+			 */
+			if (++ictx->tx_msg_count[TX_ADD_OUTPUT] > MAX_TX_MSG_RCVD)
```
niftynei, line 547: where do we initialize the `tx_msg_count` array?

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+	TX_ADD_INPUT,
+	TX_ADD_OUTPUT,
+	TX_RM_INPUT,
+	TX_RM_OUTPUT,
+};
+
+/*
+ * BOLT-f53ca2301232db780843e894f55d95d512f297f9 #2:
+ * The maximum inputs and outputs are capped at 252. This effectively fixes
+ * the byte size of the input and output counts on the transaction to one (1).
```
niftynei, line 41: i don't think we actually check for this in this file? maybe we should, in the tx builder after tx_complete is called?

niftynei (commented):
Bunch of small nits and things, but overall lgtm! Biggest thing is that there's a decent amount of setup needed for the `ictx` -- do you have a `new_ictx` method somewhere that encapsulates this?

cdecker:
This appears to duplicate a lot of code that, when changed, will cause things to slowly drift apart. Any chance of not just extracting this but also to rewire the existing uses to use `interactivetx.c`?

ddustin:
> This appears to duplicate a lot of code that, when changed, will cause things to slowly drift apart. Any chance of not just extracting this but also to rewire the existing uses to use `interactivetx.c`?

Yeah definitely. The plan is to make dual open use this code, cutting out its internal implementation, probably next release ish? Also a good moment for some interim tests to look for behavior changes in the refactor.

Dual open is currently the only user of interactivetx but splicing and dual close will want to use it.

cdecker:
Excellent, thanks @ddustin. Want me to wait for `dualopend` to be migrated on top of `interactivetx.c` or would you like to hand this PR over as is (and create an issue so we don't forget about the migration) and I can shepherd it through CI as is.

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
+#include "config.h"
+#include "interactivetx.h"
```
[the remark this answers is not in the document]
ddustin: It makes the compiler look for the file locally ie current dir but I see this isn't the usual style in the codebase. Will swap it out for `#include <common/interactivetx.h>`

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+#include <common/psbt_internal.h>
+#include <common/psbt_open.h>
+#include <common/read_peer_msg.h>
+#include <common/setup.h>
+#include <common/status.h>
+#include <common/subdaemon.h>
+#include <common/type_to_string.h>
+#include <common/wire_error.h>
+
+#define NUM_TX_MSGS (TX_RM_OUTPUT + 1)
```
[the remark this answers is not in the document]
ddustin: The idea is to NUM_TX_MSGS tied to the number of items in the enum. So if the enum gets changed INTERACTIVETX_NUM_TX_MSGS != NUM_TX_MSGS which will fail the assert on line 298. Sort of a defensive coding style to prevent future potential sneaky issues by making the failure explicit.

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+			return NULL;
+		}
+
+		/*~ We do not support multiple "live" channels, though the
+		 * protocol has a "channel_id" field in all non-gossip messages
+		 * so it's possible.  Our one-process-one-channel mechanism
+		 * keeps things simple: if we wanted to change this, we would
+		 * probably be best with another daemon to de-multiplex them;
+		 * this could be connectd itself, in fact. */
+		if (is_wrong_channel(msg, &state->channel_id, &actual)) {
```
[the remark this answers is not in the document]
ddustin: 👍 Is it worth dropping the check?

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+	bool we_complete = false, they_complete = false;
+	u8 *msg;
+	char *error = NULL;
+
+	if(received_tx_complete)
+		they_complete = *received_tx_complete;
+
+	/* Build change_set and handle PSBT variables */
+
+	ictx->change_set = NULL;
```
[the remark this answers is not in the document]
ddustin: It's meant to be only used internally for a single run of process_interactivetx_updates. The comment in the struct definition hopefully makes that clear
```
	/* Internal cached change set */
	struct psbt_changeset *change_set;
```

Any value already in it would be from a prior run and should be ignored.

common/interactivetx.c
ddustin: It is meant to be a pointer comparison yes.

common/interactivetx.c
ddustin: Oh hm that's a great question

common/interactivetx.h
```diff
@@ -0,0 +1,81 @@
...
+#include <wally_psbt.h>
+
+/* Interactive tx handles the building and updating of a transaction between
+ * two peers. A PSBT is passed back and forth between two peers in steps. In
+ * each step a peer can suggest a single change or signal they're done
+ * updating with WIRE_TX_COMPLETE. Once two steps in a row result in
+ * WIRE_TX_COMPLETE the transaction is considered complete.
+ */
+
+#define INTERACTIVETX_NUM_TX_MSGS 4
```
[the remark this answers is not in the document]
ddustin: The goal of having a different one in the header is to not pollute the global namespace with the values in enum tx_msgs (TX_ADD_INPUT etc).

common/interactivetx.c
ddustin: good idea 👍

common/interactivetx.h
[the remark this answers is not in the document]
ddustin: Okay lets put it in the header

common/interactivetx.h
```diff
@@ -0,0 +1,81 @@
...
+	 *
+	 * If set to NULL, the default implementation will simply return
+	 * ictx->desired_psbt.
+	 *
+	 * The resulting psbt's memory is stolen.
+	 *
+	 * If no more changes are demanded, return NULL or return current_psbt
+	 * unchanged to signal completion.
+	 */
+	struct wally_psbt STEALS *(*next_update)(struct interactivetx_context *ictx);
```
niftynei, line 44: `(const tal_t *ctx, struct interactivetx_context *ictx)`

common/interactivetx.h
```diff
@@ -0,0 +1,81 @@
...
+	 * psbt.
+	 */
+	struct wally_psbt *current_psbt;
+
+	/* Optional field for storing your side's desired psbt state, to be
+	 * used inside 'next_update'.
+	 *
+	 * If returned from next_update (the default) its memory will be stolen
+	 */
+	struct wally_psbt *desired_psbt STEALS;
```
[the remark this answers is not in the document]
niftynei: by default i think the *general* assumption is that alloc'd children of an obj are off of the outer obj...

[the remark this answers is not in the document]
ddustin: Let's put these on the interactivetx_context* instead and make the whole context tal allocated

common/interactivetx.c
```diff
@@ -0,0 +1,687 @@
...
+		if (is_unknown_msg_discardable(msg))
+			continue;
+
+		/* A helper which decodes an error. */
+		if (is_peer_error(msg, msg, &state->channel_id,
+				  &err, &warning)) {
+			/* In this case, is_peer_error returns true, but sets
+			 * err to NULL */
+			if (!err)
+				continue;
```
niftynei, line 94: `*error = tal_fmt(ctx, "They sent a", *warning ? "warning" : "error");`

common/interactivetx.c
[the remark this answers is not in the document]
niftynei: i think yes. need to check...

common/interactivetx.c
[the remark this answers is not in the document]
niftynei:
```suggestion
	ictx->change_set = tal_free(ictx->change_set);
```

common/interactivetx.c
```diff
@@ -0,0 +1,732 @@
...
+{
+	struct interactivetx_context *ictx = tal(ctx, struct interactivetx_context);
+
+	ictx->ctx = NULL;
+	ictx->our_role = our_role;
+	ictx->pps = pps;
+	ictx->channel_id = channel_id;
+	ictx->tx_msg_count[TX_ADD_INPUT] = 0;
+	ictx->tx_msg_count[TX_ADD_OUTPUT] = 0;
+	ictx->tx_msg_count[TX_RM_INPUT] = 0;
```
niftynei: i think we can remove the "RM" counts, as the spec only counts adds (for the reason that you'll hit the add limit first!)

niftynei (approved):
ACK with a few changes! lgtm, let's goo!

cdecker:
> ACK with a few changes! lgtm, let's goo!

Gotcha, I'll shepherd this through CI and merge ASAP ^^

ACK https://github.com/ElementsProject/lightning/pull/5287/commits/6f1f6e788fdf56b3dddd5d8727439345762192c5

PR #5288, opened 2022-05-24
Static Channel Backup.

adi2011:
SCB will enable users to backup channels in an encrypted file `emergency.recover`, and recover funds upon data corruption. I've added some RPCs and modifications to the in-memory channel struct to enable this.

-`staticbackup`: Returns SCB (in the form of a TLV) of all the existing channels.

-`recoverchannel`: Commits channels (with enough data to recover funds and the rest is dummy data) to the DB.

-`makesecret`: Gives out pseudorandom keys (based on the hex input) derived from the HSM.

-`emergencyrecover`: Recover all the stored channels in case of data loss

There's a new TLV message as well (scb_wire) which is used to serialize/deserialize the data for a channel.
The `scb` file is maintained by a plugin.

Thanks, @niftynei, and @rustyrussell for all the valuable inputs and guidance. :heart:


plugins/chanbackup.c
```diff
@@ -0,0 +1,300 @@
+#include "config.h"
```
adi2011, line 1: The plugin needs a few more addition and cleanups...

lightningd/hsm_control.c
```diff
@@ -146,6 +147,43 @@ static struct command_result *json_getsharedsecret(struct command *cmd,
 	return command_success(cmd, response);
 }
 
+static struct command_result *json_getsecret(struct command *cmd,
```
cdecker, line 150: Is this not already achieved by `getsharedsecret`? I'm always a bit hesitant to add generic "give me a secret" methods as they may lead to extracting secrets we don't want to give away.

lightningd/channel.c
```diff
@@ -441,9 +447,14 @@ struct channel *new_channel(struct peer *peer, u64 dbid,
...
 	channel->msat_to_us_max = msat_to_us_max;
-	channel->last_tx = tal_steal(channel, last_tx);
-	channel->last_tx->chainparams = chainparams;
-	channel->last_tx_type = TX_UNKNOWN;
+	if(last_tx){
+		channel->last_tx = tal_steal(channel, last_tx);
+		channel->last_tx->chainparams = chainparams;
+		channel->last_tx_type = TX_UNKNOWN;
+	}
+	else{
```
cdecker, line 455: Nit: whitespace -> `} else {`

lightningd/channel.c
```diff
@@ -441,9 +447,14 @@ struct channel *new_channel(struct peer *peer, u64 dbid,
 	channel->our_msat = our_msat;
 	channel->msat_to_us_min = msat_to_us_min;
 	channel->msat_to_us_max = msat_to_us_max;
-	channel->last_tx = tal_steal(channel, last_tx);
-	channel->last_tx->chainparams = chainparams;
-	channel->last_tx_type = TX_UNKNOWN;
+	if(last_tx){
```
cdecker, line 450: Nit: whitespace: `if (last_tx) {`

lightningd/peer_control.c
```diff
@@ -1754,6 +1755,59 @@ static const struct json_command listpeers_command = {
...
+		list_for_each(&peer->channels, channel, list)
+			json_add_scb(cmd->ld, response, channel);
+
+	json_array_end(response);
+
+	return command_success(cmd, response);
+}
+
+static const struct json_command hex_scb_command = {
+	"hex-scb",
```
cdecker, line 1802: Please don't use `-` in command names, they translate very poorly into other languages, requiring workarounds and renaming. We usually just go with something like `hexscb`. Also it'd be nicer to have `staticbackup` as a name instead, `hex` just being the format, allowing us to iterate in the future, without requiring a rename.

lightningd/channel.c
```diff
@@ -756,6 +756,14 @@ void channel_fail_permanent(struct channel *channel,
 			    const char *fmt,
 			    ...)
 {
+	/* FIXME: should I compare the whole scid for stub channel?
+		I doubt there are actually any channels in the Genesis block(1x1x1) */
+	if(channel->scid->u64 >> 40 == 1){
```
cdecker, line 761: Is `1x1x1` a special value in this case? We're not actually guaranteed to have an `scid` at the time we create the recovery data, since the channel may not have yet confirmed.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+						&nodeid,
+						&addr,
+						false
+						);
+	
+	struct lightningd *ld = cmd->ld;
+	
+
+	struct channel_config our_config;
+	memset(&our_config, 1, sizeof(struct channel_config));
```
cdecker, line 1224: We don't usually initialize structs this way, since it has a tendency of hiding uninitialized memory accesses.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	struct pubkey localFundingPubkey;
+	get_channel_basepoints(ld, 
+						   &nodeid, 
+						   id, 
+						   &basepoints, 
+						   &localFundingPubkey);
+
+	struct channel_info *channel_info = tal(ld->wallet, 
+											struct channel_info);
+	memset(channel_info, 3, sizeof(*channel_info));
```
cdecker, line 1250: Same here.

lightningd/channel.c
cdecker: I see it below, seems a bit strange to use this, and not a way more unreasonable `UINT64_MAX`, but it'd ultimately be the same anyway. Maybe it'd be better to just keep it `NULL`?

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	u32 blockht = 100;
+	
+	struct short_channel_id *scid = tal(cmd, struct short_channel_id);
+	
+	/*To indicate this is an stub channel we keep it's scid to 1x1x1.*/
+	if(!short_channel_id_from_str("1x1x1", 6, scid)){
+		return NULL;
+	}
+	
+	struct channel *channel; //Channel Shell with Dummy data(mostly)
```
cdecker, line 1266: I think we use `/* ... */` almost everywhere instead of `//`

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      LOCAL,
+			      peer->ld->log,
+			      "billboard",
+			      8, &our_config,
+			      101,
+			      1, 1, 1,
+			      &funding,
+			      AMOUNT_SAT(500000000), // FIXME: Is this correct? 
+				  /*Need to keep this very high from trimming the fees 
+				  	in onchaind.c(main).*/
```
cdecker, line 1280: Might be worthwile to add the actual funding amount to the recovery data? It's not going to change anyway, so it'd be safe.

lightningd/opening_control.c
```diff
@@ -1227,3 +1406,13 @@ static const struct json_command fundchannel_complete_command = {
     "with {psbt}. Returns true on success, false otherwise."
 };
 AUTODATA(json_command, &fundchannel_complete_command);
+
+static const struct json_command json_commitchan_command = {
+    "emergencystubchannel",
```
cdecker, line 1411: `recoverchannel`?

wallet/wallet.c
```diff
@@ -1488,7 +1492,15 @@ static struct channel *wallet_stmt2channel(struct wallet *w, struct db_stmt *stm
...
 		return NULL;
 	}
-
+	if(stub_chan){
+		/* FIXME: What should I put in as an error?
+			the peer shouldn't know we have lost the states,
+			that can motivate him to cheat.*/
+		chan->error = towire_errorfmt(w->ld, 
+									  &cid,
+									 "We can't be together anymore.."
```
cdecker, line 1501: I like this error message :+1:

plugins/chanbackup.c
```diff
@@ -0,0 +1,304 @@
...
+						       NULL, 0, 0))
+		return;
+	
+	if (!write_all(fd, final, tal_bytelen(final))) {
+			unlink_noerr("scb");
+	}
+}
+
+/* checks if the SCB file exists, creates a new one in case it doesn't. */
+static void maybe_create_new_scb(struct plugin *p, const char *scb_buf)
```
cdecker, line 84: This overwrites the file in place, which does not guarantee atomic operations. It's better to create a temporary file `recover.tmp`, write to that and then rename `recover.tmp` to `recover`. That guarantees that a partial write (plugin crash, loss of power, etc) doesn't corrupt your recovery file.

cdecker (commented):
Excellent work @adi2011, I saw your PR and got excited to take a look. I know this is just a draft, but it's already looking very good.

Since I was here I added a couple of comments, mostly unimportant stuff like formatting (hadn't realized the Draft status) so feel free to ignore them as you likely would have addressed them before undrafting.

One bikeshedding topic I'd like to tackle though is the name SCB, as it always annoyed me that it is being positioned as a backup. Would you be open to call it "emergency recovery" or something like that?

lightningd/hsm_control.c
adi2011: `getsharedsecret` needs a compressed public key and then it performs ecdh() to give out a key for connection handshake and for onion wrapping.

We needed a pseudorandom key derived from hsm_secret, using hkdf_sha256() taking in a string as info, but I get your point. 
what if instead of taking in an 'info' string from the user I hard code it to "peer backup" to get a specific key used to encrypt the scb file?

lightningd/channel.c
adi2011: I am setting it to 1x1x1 to indicate (while loading the in-memory channel from DB) that it's a stub channel and we need to send the error to the peer, nothing else. It was on my to-do list to set it to 0x0x0 (as Rusty suggested), but it won't make any difference. 

I think setting it to NULL will create confusion with the unconfirmed channels which also don't have any scid? If that's not the case I am happy to set it to NULL.

lightningd/opening_control.c
adi2011: Yeah, makes sense, I'll change these.

lightningd/opening_control.c
adi2011: Sure, I am anyways bad at naming things in general...

adi2011:
> Excellent work @adi2011, I saw your PR and got excited to take a look. I know this is just a draft, but it's already looking very good.
Thank you so much for taking out the time to check this out. I'll soon remove the draft tag after testing notification responses of the plugin, looking forward to more valuable feedback :smile:

> One bikeshedding topic I'd like to tackle though is the name SCB, as it always annoyed me that it is being positioned as a backup. Would you be open to call it "emergency recovery" or something like that?
Yes, I agree with this. Do I need to then change the names of the TLV message and commands as well?

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+								struct channel_id cid,
+								struct bitcoin_outpoint funding,
+								struct wireaddr_internal addr){
+	struct peer *peer;
+	peer = new_peer(cmd->ld,
+						0,
+						&nodeid,
+						&addr,
+						false
+						);
```
rustyrussell, line 1218: Generally prefer `);` on same line as last argument.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
 	return command_still_pending(cmd);
 }
 
+static struct channel *stub_chan(struct command *cmd,
+								u64 id,
+								struct node_id nodeid, 
+								struct channel_id cid,
+								struct bitcoin_outpoint funding,
+								struct wireaddr_internal addr){
```
rustyrussell, line 1211: `{` goes on a line by itself, in column 0.

lightningd/opening_control.c
rustyrussell, line 1224: If we want to make sure we set every field we use, I prefer using dynamic allocation (i.e. `our_config = tal(tmpctx,...`) so valgrind will tell us (valgrind doesn't work well with stack variables).

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	struct lightningd *ld = cmd->ld;
+	
+
+	struct channel_config our_config;
+	memset(&our_config, 1, sizeof(struct channel_config));
+	our_config.id = 0;
+	
+	u32 feerate;
+	/* FIXME: What should be the appropriate value here? */
+	feerate = 192838;
```
rustyrussell, line 1229: Hmm, ideally we would allow NULL, but in this case we can just use FEERATE_FLOOR?

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+
+	struct bitcoin_signature sig;
+	sig.sighash_type = SIGHASH_ALL;
+
+	struct pubkey pk;
+	struct basepoints basepoints;
+	pubkey_from_der(tal_hexdata(ld->wallet, 
+					type_to_string(tmpctx, struct node_id, &nodeid), 66),
+					33,
+					&pk);
```
rustyrussell, line 1239: `if (!pubkey_from_node_id().... fatal("Invalid node id")`.

lightningd/opening_control.c
rustyrussell, line 1250: This memset, too, should be removed.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	channel_info->old_remote_per_commit = pk;
+
+	u32 blockht = 100;
+	
+	struct short_channel_id *scid = tal(cmd, struct short_channel_id);
+	
+	/*To indicate this is an stub channel we keep it's scid to 1x1x1.*/
+	if(!short_channel_id_from_str("1x1x1", 6, scid)){
+		return NULL;
+	}
```
rustyrussell, line 1264:
```
if (!mk_short_channel_id(&scid, 1, 1, 1))
        fatal("Failed to make short channel 1x1x1!");
```

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	
+	struct channel *channel; //Channel Shell with Dummy data(mostly)
+
+	channel = new_channel(peer, id,
+			      NULL, /* No shachain yet */
+			      CHANNELD_NORMAL,
+			      LOCAL,
+			      peer->ld->log,
```
rustyrussell, line 1272: Just use `ld->log`?

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      "billboard",
```
rustyrussell, line 1273: "restored from static channel backup"?  You can see this in listpeers, so it's nice to have....

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      8, &our_config,
```
rustyrussell, line 1274: channel_flags should be 0, not 8.  It's only currently got one bit, which is whether to advertize the channel.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      101,
+			      1, 1, 1,
```
rustyrussell, line 1276: None of these should matter, so minimum_depth can be 0, others maybe 0, but 1 is OK too.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      &funding,
+			      AMOUNT_SAT(500000000), // FIXME: Is this correct?
```
rustyrussell, line 1278: It's definitely not correct.  But it's also not saved in the static channel backup, and maybe it should be?  It doesn't really matter, but it shows up in listpeers...

lightningd/opening_control.c
rustyrussell, line 1280: Does this comment apply to the above amount?  If so, move it up above there?

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      101,
+			      1, 1, 1,
+			      &funding,
+			      AMOUNT_SAT(500000000), // FIXME: Is this correct? 
+				  /*Need to keep this very high from trimming the fees 
+				  	in onchaind.c(main).*/
+			      AMOUNT_MSAT(0),
+			      AMOUNT_SAT(0),
+			      true, /* !remote_funding_locked */
+			      scid, /* no scid yet */
```
rustyrussell, line 1284: Comment isn't correct any more. (`/* no scid yet */`)

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      &cid,
+			      /* The three arguments below are msatoshi_to_us,
+			       * msatoshi_to_us_min, and msatoshi_to_us_max.
+			       * Because, this is a newly-funded channel,
+			       * all three are same value. */
+			      AMOUNT_MSAT(0),
+			      AMOUNT_MSAT(0), /* msat_to_us_min */
+			      AMOUNT_MSAT(0), /* msat_to_us_max */
```
rustyrussell, line 1292: I wonder if the max should be the channel capacity, since we don't know. It's really just for show in listpeers though...

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      channel_info,
+			      new_fee_states(ld->wallet, LOCAL, &feerate),
+			      NULL, /* No shutdown_scriptpubkey[REMOTE] yet */
+			      NULL,
+			      1, false,
+			      NULL, /* No commit sent yet */
+			      /* If we're fundee, could be a little before this
+			       * in theory, but it's only used for timing out. */
+			      get_network_blockheight(ld->topology),
+			      100, 10000,
```
rustyrussell, line 1305: onchaind needs these to be correct, (min and max possible feerate) which is a problem!  It iterates through this range to look for where the signatures match (so it doesn't have to remember each fee value for each tx).

min is easy: FEERATE_FLOOR.  max is harder: if we set it to -1ULL (max u64), onchaind will spin forever: see line 204!  But we can assume it's less than the funding, if we store that in the scb, and we have a MINIMUM_TX_WEIGHT constant, so we can actually set it to:

```
funding.satoshis / MINIMUM_TX_WEIGHT * 1000 /* Raw: convert to feerate */
```
(Note that you're not normally supposed to access fields on amount_sat, so the comment suppresses the warning!).

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      /* We are connected */
+			      false,
```
rustyrussell, line 1307: Comment :)

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+			      &basepoints,
+			   	  &localFundingPubkey,
+			      NULL,
+			      ld->config.fee_base,
+			      ld->config.fee_per_satoshi,
+			      NULL,
+			      0, 0,
+			      channel_type_static_remotekey(NULL),
```
rustyrussell, line 1315: We should remember channel_type in static backup too.  This might be important in future!

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+{
+	u64 id;
+	struct node_id nodeid;
+	struct channel_id cid;
+	struct bitcoin_outpoint funding;
+	struct wireaddr_internal addr;
+	const char *scb;
+
+	if (!param(cmd, buffer, params,
+		p_req("scb", param_string, &scb),
```
rustyrussell, line 1340: 1. Use a JSON array here, rather than a comma-separated string.
2. Do the work inside the param function (say, write param_scb_array)

The first is just more natural for JSON, and the second is better because they can use the "check" command to make sure it parses.

lightningd/opening_control.c
```diff
@@ -1202,6 +1203,184 @@ static struct command_result *json_fundchannel_start(struct command *cmd,
...
+	while(token != NULL){
+
+		u8 *scb_arr = tal_hexdata(cmd, token, strlen(token));
+		if(!fromwire_static_chan_backup(scb_arr, 
+										&id, 
+										&cid, 
+										&nodeid, 
+										&addr, 
+										&funding)){
+			/* FIXME: INSERT BROKEN STATEMENT HERE!*/
```
rustyrussell, line 1355: If you do this inside the param parser, it's natural: return an error that the scb is invalid.

rustyrussell (changes requested):
OK, didn't review plugin yet!  But lots of comments on the other stuff.  Mostly easy things, though!

This is an excellent first PR :)

hsmd/libhsmd.c
```diff
@@ -257,6 +259,22 @@ static void hsm_channel_secret_base(struct secret *channel_seed_base)
...
+
+	if (!fromwire_hsmd_derive_secret(tmpctx, msg_in, &info))
+		return hsmd_status_malformed_request(c, msg_in);
+	
+	struct secret secret;
+
+	hkdf_sha256(&secret, sizeof(struct secret), NULL, 0,
+		    &secretstuff.hsm_secret, sizeof(secretstuff.hsm_secret),
+		    info, tal_bytelen(info));
+	return towire_hsmd_derive_secret_reply(NULL, &secret);
```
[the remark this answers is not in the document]
adi2011: Yes, this makes sense :+1:...

lightningd/opening_control.c
adi2011: Yes, onchaind.c trims fees from this amount, so to make sure it doesn't fail I kept it really high.

lightningd/opening_control.c
adi2011: yes, sorry:sweat_smile:

adi2011:
Thank you so much for the review and sorry for the `NIT:`s (I'll keep them in mind from now onwards), I'll try to incorporate all the changes that you and @cdecker suggested by the end of this week.

lightningd/peer_control.c
```diff
@@ -1754,6 +1755,59 @@ static const struct json_command listpeers_command = {
...
+	"backup",
+	json_hex_scb,
+	"Returns SCB of all the channels currently present in the DB"
+};
```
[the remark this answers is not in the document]
adi2011: Yes, I am completely removing this RPC and adding it in `listpeers`...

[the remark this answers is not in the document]
adi2011: Or maybe keep this just to make it easy for fetching scb by the plugin? Because we don't need SCB for uncommitted and unsaved channels and `listpeers` have them as well, Extracting it in RPC_scan is just too much string-parsing :sweat_smile:

lightningd/peer_control.c
vincenzopalazzo: Maybe `static_backup` is better? it makes the docs more accessible for screen readers too.

But usually, the `_` will translate directly into other languages with the camel case, so it is more difficult for single word command name.

But it is minor if it requires a lot of work!

lightningd/channel.c
vincenzopalazzo: I like the idea to make the code unambiguous for the new readers, maybe we can make a `#DEFINE GENESIS_SHORT_ID "1x1x1"`

P.S: I'm assuming that it is easy to convert into a string the `channel->scid`

lightningd/peer_control.c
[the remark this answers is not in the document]
vincenzopalazzo: we need to think also about the performance of these command`listpeers`, maybe can be improved by #5022 but I don't know yet!

lightningd/channel.c
vincenzopalazzo: I need to remember to read the PR from bottom to top, the Rusty suggestion is better I think https://github.com/ElementsProject/lightning/pull/5288#discussion_r881234472

lightningd/opening_control.c
adi2011: Hmm just to keep the expectations (of the user) low, I think we should keep this to 0 😆...

lightningd/channel.h
```diff
@@ -72,6 +72,23 @@ struct open_attempt {
...
+
+	/* Peer's node ID */
+	struct node_id node_id;
+
+	/* To connect to the peer upon rebuidling */
+	struct wireaddr_internal addr;
+
+	/* This would be useful for keeping a watch on funding txn */
+	struct bitcoin_outpoint funding;
+};
```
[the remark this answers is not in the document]
adi2011: > You can actually get our generation code to generate this for you!
We'd need the PSBT for that I guess...

adi2011:
I hope after sorting out the final commit's FIXME, this would be mergeable 🤞🏻 ...

lightningd/hsm_control.c
ZmnSCPxj: Ummmm just generate a random compressed public key?  Pick one out of a hat, then call `getsharedsecret` and then use the result as the encryption key, AND store the compressed public key in the first 33 bytes of the emergency-recovery file.  This lets you recover it anyway with the same `hsm_secret`; you only need the session pubkey.

plugins/chanbackup.c
ZmnSCPxj: Required reading: [UPDATE IN PLACE: A Poison Apple?](https://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.108.7810)

lightningd/hsm_control.c
ZmnSCPxj: Secondarily, the VLS people have implemented a hardware version of the `hsmd`, and adding a new API to the `hsmd` requires that we at least inform them so theey can implement as well.  So I would be reluctant to add a new `hsmd_derive_secret` message here, we should try to use `getsharedsecret` and just give a random point that you store with the backup file anyway.  The random point will not leak the encryption key because of ECDH magic so it is safe to put it right beside the encrypted blob in the backup file.

plugins/chanbackup.c
```diff
@@ -0,0 +1,329 @@
...
+#include <errno.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sodium.h>
+
+/* Global secret object to keep the derived encryption key for the SCB */
+struct secret secret;
+
+/* Helper to fetch out SCB from the RPC call */
+static bool json_to_str (const char *buffer, const jsmntok_t *tok,
```
vincenzopalazzo, line 27: by not just return the string?

plugins/chanbackup.c
```diff
@@ -0,0 +1,329 @@
...
+							   crypto_secretstream_xchacha20poly1305_HEADERBYTES,
+						       NULL, point,
+						       tal_bytelen(point),
+						       /* Additional data and tag */
+						       NULL, 0, 0))
+		return;
+	
+	if (!write_all(fd, final, tal_bytelen(final))) {
+			unlink_noerr("scb.tmp");
+	}
```
vincenzopalazzo, line 65:
```suggestion
	if (!write_all(fd, final, tal_bytelen(final))) 
			unlink_noerr("scb.tmp");
```

vincenzopalazzo (changes requested):
I make the first review of this, so I can continue later.

basically are only C-style code fixes, and some json rpc command good manners when there is a failure.

plugins/chanbackup.c
adi2011: Tried to follow the convention from `json_helpers.c`. It's done in this way there...

adi2011:
Thank you @vincenzopalazzo  for the review! I've incorporated these changes in the last fixup

plugins/chanbackup.c
```diff
@@ -0,0 +1,328 @@
...
+
+static struct command_result *after_staticbackup(struct command *cmd,
+					 const char *buf,
+					 const jsmntok_t *params,
+					 void *cb_arg UNUSED)
+{
+
+	const char *scb_buf =  json_strdup(tmpctx, buf, params);
+	const jsmntok_t *scb_arr = json_parse_simple(tmpctx, scb_buf, strlen(scb_buf)),
+					*scbs = json_get_member(scb_buf, scb_arr, "scb");
```
rustyrussell, line 247: You can just do `const jmsntok_t *scbs = json_get_member(params, "scb")`.  You are converting the JSON to a string and reparsing it, which is unnecessary!

plugins/chanbackup.c
```diff
@@ -0,0 +1,328 @@
...
+{
+	*t = json_strdup(tmpctx, buffer, tok);
+	return true;
+}
+
+/* This writes encrypted scb in the file */
+static void write_hsm(struct plugin *p, int fd, 
+					 const char *buf)
+{
+	u8 *point = tal_dup_arr(buf, u8, (u8 *)buf, strlen(buf), 0);
```
rustyrussell, line 38: This routine encrypts the JSON, which is inefficient.  Define a message containing an array of scbs in common/scb_wire.csv, and then use "towire_" and "fromwire_" to put it in the file.

plugins/chanbackup.c
```diff
@@ -0,0 +1,328 @@
...
+{
+	const jsmntok_t *notiftok = json_get_member(buf, params, "channel_state_changed");
+	const jsmntok_t *statetok = json_get_member(buf, notiftok, "new_state");
+
+	const char *state = json_strdup(tmpctx, buf, statetok);
+
+	/* FIXME: I wanted to update the file on CHANNELD_AWAITING_LOCKIN,
+	 * But I don't get update for it, maybe because there is no previous_state,
+	 * also apparently `channel_opened` gets published when *peer* funded a channel with us? 
+	 * So, is their no way to get a notif on CHANNELD_AWAITING_LOCKIN? */
```
rustyrussell, line 265: Hmm, file a bug, this seems wrong!

plugins/chanbackup.c
```diff
@@ -0,0 +1,328 @@
...
+	if(!strcmp(state ,"CLOSED") || !strcmp(state ,"CHANNELD_NORMAL")){
```
rustyrussell, line 266: You can use json_tok_streq() BTW. And needs a space before `{`

lightningd/opening_control.c
niftynei: 2,100,000,000,000,001 satoshis?

adi2011: > 2,100,000,000,000,001 satoshis?

Fixed in later commits.

plugins/chanbackup.c
```diff
@@ -0,0 +1,393 @@
...
+}
+
+static const char *init(struct plugin *p,
+			const char *buf UNUSED,
+			const jsmntok_t *config UNUSED)
+{
+	const char *scb_buf;
+	rpc_scan(p, "staticbackup",
+		 take(json_out_obj(NULL, NULL, NULL)),
+		 "{scb:%}", JSON_SCAN(json_to_str, &scb_buf));
```
rustyrussell, line 353: Hmm, converting to a string an back to JSON again is really ugly :(

Can you simply call staticbackup async here?  Maybe create a function to wrap this:

```
		struct out_req *req;
		req = jsonrpc_request_start(plugin, NULL,
                                            "staticbackup",
                                            after_staticbackup,
                                            &forward_error,
                                            NULL);

		return send_outreq(plugin, req);
```
?

plugins/chanbackup.c
```diff
@@ -0,0 +1,393 @@
...
+static struct command_result *after_staticbackup(struct command *cmd,
+					         const char *buf,
+					         const jsmntok_t *params,
+					         void *cb_arg UNUSED)
+{
+	const jsmntok_t *scbs = json_get_member(buf, params, "scb");
+	plugin_log(cmd->plugin, LOG_INFORM, "Updating the SCB");
+
+	update_scb(cmd->plugin, json_strdup(tmpctx, buf, scbs));
+	return notification_handled(cmd);
```
rustyrussell, line 313: Hand this the buf and scbs token, and have it iterate through, rather than covnerting to a string and back...

This is neater, since scbs->size will tell it immediately how many scbs there are.

plugins/chanbackup.c
```diff
@@ -0,0 +1,393 @@
...
+		json_state_changed,
+	}
+};
+
+static const struct plugin_command commands[] = { {
+		"emergencyrecover",
+		"recovery",
+		"Populates the DB with stub channels",
+		"returns stub channel-id's on completion",
+		recover,
```
rustyrussell, line 381: I always like the function for a command to be json_COMMANDNAME, in this case json_emergencyrecover.  That way I can easily find it!

rustyrussell (changes requested):
This is really nice!  No really big feedback, just some style things and naming.

Fix those and we can enable this.  I've marked it for the v0.12 milestone!

tools/generate-wire.py
```diff
@@ -244,7 +244,7 @@ class Type(FieldSet):
         'tx_parts',
         'wally_psbt',
         'wally_tx',
-        'channel_type',
```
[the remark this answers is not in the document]
adi2011: It's a duplicate...

rustyrussell:
> Suggest adding a section to `doc/BACKUP.md` about this. In particular, since this now uses an atomic construct-temporary-file-and-rename, it is very safe for an automated backup process to hardlink the recovery file, copy the hardlinked name, then unlink, and the automated backup process assuredly will never acquire a corrupted copy, either.

It's also safe to just copy :)  Every copy does open, read, close, which has the same semantic.

lightningd/hsm_control.c
```diff
@@ -146,6 +148,47 @@ static struct command_result *json_getsharedsecret(struct command *cmd,
...
+	const char *info;
+	struct json_stream *response;
+	struct secret secret;
+
+	if (!param(cmd, buffer, params,
+		   p_req("info", param_string, &info),
+		   NULL))
+		return command_param_failed();
+
+	u8 *point = tal_dup_arr(cmd, u8, (u8*)info, strlen(info), 0);
```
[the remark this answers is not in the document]
adi2011: Agreed, changing it...

plugins/chanbackup.c
adi2011: I Applied 200% of my brain and created `json_to_scb_chan`, which makes it easier for writing it in the file as well...

plugins/chanbackup.c
```diff
@@ -0,0 +1,396 @@
...
+static bool json_to_scb_chan(const char *buffer,
+                        const jsmntok_t *tok,
+		        struct scb_chan ***channels)
+{
+	size_t i;
+	const jsmntok_t *t;
+	*channels = tok->size ? tal_arr(tmpctx, struct scb_chan *, tok->size) : NULL;
+
+	json_for_each_arr(i, t, tok) {
+		const u8 *scb_tmp = tal_hexdata(tmpctx, json_strdup(tmpctx, buffer, t),
```
adi2011, line 37: Grossly violated the 80 col rule here...

niftynei:
Looks like there's a bug in `test_emergencybackup`

```
2022-07-08T00:53:35.357Z **BROKEN** lightningd: FATAL SIGNAL 11 (version c4fe1d3-modded)
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: common/daemon.c:38 (send_backtrace) 0x564000429f3e
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: common/daemon.c:46 (crashdump) 0x564000429f92
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: /build/glibc-SzIz7B/glibc-2.31/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0 ((null)) 0x7ff33ef5e08f
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ./bitcoin/short_channel_id.h:42 (is_stub_scid) 0x5640003a2bb0
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/channel.c:789 (channel_fail_permanent) 0x5640003a575a
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/onchain_control.c:623 (onchaind_funding_spent) 0x5640003d0496
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/peer_control.c:1638 (funding_spent) 0x5640003e5fa9
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/watch.c:271 (txowatch_fire) 0x5640004002d3
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:87 (filter_block_txs) 0x56400039fc0d
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:758 (add_tip) 0x5640003a1ac8
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:850 (get_new_block) 0x5640003a2027
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/bitcoind.c:388 (getrawblockbyheight_callback) 0x56400039e34f
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:587 (plugin_response_handle) 0x5640003f44b1
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:693 (plugin_read_json_one) 0x5640003f46da
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:738 (plugin_read_json) 0x5640003f48be
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:59 (next_plan) 0x56400049b3ce
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:407 (do_plan) 0x56400049bfd6
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:417 (io_ready) 0x56400049c018
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ccan/ccan/io/poll.c:453 (io_loop) 0x56400049e30b
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/io_loop_with_timers.c:22 (io_loop_with_timers) 0x5640003c2ce4
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: lightningd/lightningd.c:1184 (main) 0x5640003c8fc8
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: ../csu/libc-start.c:308 (__libc_start_main) 0x7ff33ef3f082
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x56400039d38d
2022-07-08T00:53:35.357Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0xffffffffffffffff
2022-07-08T00:53:35.616Z **BROKEN** lightningd: FATAL SIGNAL 11 (version c4fe1d3-modded)
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: common/daemon.c:38 (send_backtrace) 0x562b4b73df3e
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: common/daemon.c:46 (crashdump) 0x562b4b73df92
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: /build/glibc-SzIz7B/glibc-2.31/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0 ((null)) 0x7fe0516bc08f
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ./bitcoin/short_channel_id.h:42 (is_stub_scid) 0x562b4b6b6bb0
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/channel.c:789 (channel_fail_permanent) 0x562b4b6b975a
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/onchain_control.c:623 (onchaind_funding_spent) 0x562b4b6e4496
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/peer_control.c:1638 (funding_spent) 0x562b4b6f9fa9
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/watch.c:271 (txowatch_fire) 0x562b4b7142d3
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:87 (filter_block_txs) 0x562b4b6b3c0d
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:758 (add_tip) 0x562b4b6b5ac8
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/chaintopology.c:850 (get_new_block) 0x562b4b6b6027
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/bitcoind.c:388 (getrawblockbyheight_callback) 0x562b4b6b234f
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:587 (plugin_response_handle) 0x562b4b7084b1
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:693 (plugin_read_json_one) 0x562b4b7086da
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/plugin.c:738 (plugin_read_json) 0x562b4b7088be
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:59 (next_plan) 0x562b4b7af3ce
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:407 (do_plan) 0x562b4b7affd6
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ccan/ccan/io/io.c:417 (io_ready) 0x562b4b7b0018
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ccan/ccan/io/poll.c:453 (io_loop) 0x562b4b7b230b
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/io_loop_with_timers.c:22 (io_loop_with_timers) 0x562b4b6d6ce4
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: lightningd/lightningd.c:1184 (main) 0x562b4b6dcfc8
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: ../csu/libc-start.c:308 (__libc_start_main) 0x7fe05169d082
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x562b4b6b138d
2022-07-08T00:53:35.617Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0xffffffffffffffff
```

niftynei:
test_emergencyrecover is failing under postgres, because it expects a sqlite3 database file to delete

```

deftest_emergencyrecover(node_factory, bitcoind):
"""
    Test emergencyrecover
    """
        l1, l2 = node_factory.get_nodes(2, opts=[{}, {}])
        c12, _ = l1.fundchannel(l2, 10**5)
        l1.stop()
>       os.unlink(os.path.join(l1.daemon.lightning_dir, TEST_NETWORK, "lightningd.sqlite3"))
E       FileNotFoundError: [Errno 2] No such file or directory: '/tmp/ltests-a1vrjh39/test_emergencyrecover_1/lightning-1/regtest/lightningd.sqlite3'
```

adi2011:
> test_emergencyrecover is failing under postgres, because it expects a sqlite3 database file to delete
Solved...

PR #5291, opened 2022-05-29
Mistaken default directory in plugin docs

AutonomousOrganization:
directory .lightningd -> .lightning


rustyrussell:
Thanks, good catch!

Issue #5293, opened 2022-05-30
FATAL on malformed zbase on checkmessage RPC


gdassori:
### Issue and Steps to Reproduce

```
$ lightning-cli checkmessage -k message=msg zbase=wrong
lightning-cli: reading response: socket closed
```
from the logs:
```
lightningd: lightningd/signmessage.c:59: from_zbase32: Assertion `len == tal_bytelen(u8arr)' failed
lightningd: FATAL SIGNAL 6 (version v0.11.1)
0x55b9b1b4e617 send_backtrace
[...]
```

### `getinfo` output
```
   "version": "v0.11.1",
```

vincenzopalazzo:
This happens because of this parameter `zbase=wrong` this should never happens, but I agree that the cmd line should be more user friendly!

I will try to look inside it

PR #5297, opened 2022-05-31
signmessage: improve the UX of the rpc command when zbase is not a valid one

vincenzopalazzo:
This PR improves the UX of the RPC command when `zbase` is not a valid one because we have a crash on bad `zbase`!

Fixes https://github.com/ElementsProject/lightning/issues/5293

### DIFF
--- a/lightningd/signmessage.c
+++ b/lightningd/signmessage.c
@@ -56,8 +56,7 @@ static const u8 *from_zbase32(const tal_t *ctx, const char *msg)
 	if (!bech32_convert_bits(u8arr, &len, 8,
 				 u5arr, tal_bytelen(u5arr), 5, false))
 		return tal_free(u8arr);
-	assert(len == tal_bytelen(u8arr));
-	return u8arr;
+	return len == tal_bytelen(u8arr) ? u8arr : tal_free(u8arr);
 }
 
 static struct command_result *json_signmessage(struct command *cmd,
@@ -235,4 +234,3 @@ static const struct json_command json_checkmessage_cmd = {
 	"Verify a digital signature {zbase} of {message} signed with {pubkey}",
 };
 AUTODATA(json_command, &json_checkmessage_cmd);
-
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1867,6 +1867,9 @@ def test_signmessage(node_factory):
     checknokey = l2.rpc.checkmessage(message="message for you", zbase=zm)
     assert checknokey['pubkey'] == l1.info['id']
     assert checknokey['verified']
+    # check that checkmassage used with a wrong zbase format throws an RPC exception
+    with pytest.raises(RpcError, match="zbase is not valid zbase32"):
+        l2.rpc.checkmessage(message="wrong zbase format", zbase="wrong zbase format")
 
 
 def test_include(node_factory):


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5297/commits/3fbf38a232a17ce50d10cbc76bc40d13a3603fe2

Just needs a changelog entry in the commit :+1:

vincenzopalazzo:
>Just needs a changelog entry in the commit +1

I will never learn this :) added now!

Was thinking that maybe having a unit test can be a good idea to avoid regression with future refactoring.

What do you think? or it is meaningless?

cdecker:
Maybe not a full unit test (startup time vs runtime tradeoff) but if there's a sign message test already that would be a good addition 👍

vincenzopalazzo:
>Maybe not a full unit test (startup time vs runtime tradeoff) 

Good point! I think next time also with this in mind. I expanded a `signmessage` test that was already present

lightningd/signmessage.c
```diff
@@ -56,8 +56,7 @@ static const u8 *from_zbase32(const tal_t *ctx, const char *msg)
 	if (!bech32_convert_bits(u8arr, &len, 8,
 				 u5arr, tal_bytelen(u5arr), 5, false))
 		return tal_free(u8arr);
-	assert(len == tal_bytelen(u8arr));
-	return u8arr;
+	return len == tal_bytelen(u8arr) ? u8arr : NULL;
```
cdecker, line 59: I think this ends up leaking `u8arr` in case the length doesn't match. Better free it with `tal_free` here.

cdecker (changes requested):

lightningd/signmessage.c
vincenzopalazzo: Thanks for this catch, I added it right now and free the array in the case of the invalid length!

rustyrussell (changes requested):
While it's fine to refer the github bugs like you did, you should *also* explain the problem, and not assume that GH is accessible in future.

i.e. mention that we crash on bad zbase arg :)

tests/test_misc.py
```diff
@@ -1867,6 +1867,9 @@ def test_signmessage(node_factory):
     checknokey = l2.rpc.checkmessage(message="message for you", zbase=zm)
     assert checknokey['pubkey'] == l1.info['id']
     assert checknokey['verified']
+    # check that checkmassage used with a wrong zbase format throws an RPC exception
+    with pytest.raises(RpcError):
+        l2.rpc.checkmessage(message="wrong zbase format", zbase="wrong zbase format")
```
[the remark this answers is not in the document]
vincenzopalazzo: Right! I avoid this usually because changing the error message required to change the test, but I did not propose a better solution so I introduce the match on the exception message!

Thanks!

vincenzopalazzo:
>While it's fine to refer the github bugs like you did, you should also explain the problem, and not assume that GH is accessible in future.

Yeah! this was a terrible PR description, take me also some time to remember what was the problem!

Sorry about that!

rustyrussell (approved):

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/fd823468a059d8c4869874d22cf93d907a227987

Issue #5284, opened 2022-05-22
V0.11.1, connectd crash


devastgh:
### Issue and Steps to Reproduce
Might be similar to #5282

Flow of events:
Compacting backup:
```
2022-05-17T23:07:50.314Z INFO    plugin-backup.py: Starting compaction: stats={'before': {'backupsize': 8571037716, 'version_count': 939851}}
2022-05-17T23:26:26.894Z INFO    plugin-backup.py: Adding intial snapshot with 650534912 bytes for version 2345945
2022-05-17T23:26:27.602Z INFO    plugin-backup.py: Compacted 939849 changes, saving 7920501846 bytes, swapping backups
```

One force close, that probably shouldn't have happened:

```
2022-05-17T23:26:31.547Z UNUSUAL xxxxxxxxxxxx-chan#1985: Peer permanent failure in CHANNELD_NORMAL: Offered HTLC 3361 SENT_ADD_ACK_REVOCATION cltv 736843 hit deadline
2022-05-17T23:26:31.549Z INFO    xxxxxxxxxxxxx-chan#1985: State changed from CHANNELD_NORMAL to AWAITING_UNILATERAL
```

Crash:

```
2022-05-17T23:26:31.567Z **BROKEN** connectd: FATAL SIGNAL 6 (version 0.11.1)
2022-05-17T23:26:31.567Z **BROKEN** connectd: backtrace: common/daemon.c:38 (send_backtrace) 0x557c9a97e96c

```
Backtrace from syslog:
```

May 18 01:26:29 xxxxxxxx lightningd[2077]: lightning_connectd: ccan/ccan/tal/tal.c:393: del_tree: Assertion `!taken(from_tal_hdr(t))' failed.
May 18 01:26:29 xxxxxxxx lightningd[2077]: lightning_connectd: FATAL SIGNAL 6 (version 0.11.1)
May 18 01:26:29 xxxxxxxx lightningd[2077]: 0x557c9a97e924 send_backtrace
May 18 01:26:29 xxxxxxxx lightningd[2077]: #011common/daemon.c:33
May 18 01:26:29 xxxxxxxx lightningd[2077]: 0x557c9a97e9be crashdump
May 18 01:26:29 xxxxxxxx lightningd[2077]: #011common/daemon.c:46
May 18 01:26:29 xxxxxxxx lightningd[2077]: 0x7f42b8dd908f ???
May 18 01:26:29 xxxxxxxx lightningd[2077]: #011???:0
[...]

Issue #5282, opened 2022-05-22
v0.11.1, crash in connectd


btweenthebars:
### Issue and Steps to Reproduce
I've been experiencing this crash for a while (every 3~ days) and I know another CLN 0.11.1 user that ran into it.

I think I started seeing it since upgrading from rc5 to 0.11

from console:
```
lightning_connectd: FATAL SIGNAL 6 (version v0.11.1)
0x556ec0ac8381 send_backtrace
	common/daemon.c:33
0x556ec0ac8429 crashdump
	common/daemon.c:46
0x7f60ee66f55f ???
	???:0
0x7f60ee6bc34c ???
	???:0
0x7f60ee66f4b7 ???
	???:0
0x7f60ee659533 ???
	???:0
0x556ec0b20179 call_error
	ccan/ccan/tal/tal.c:93
0x556ec0b2034a check_bounds
	ccan/ccan/tal/tal.c:165
0x556ec0b203a8 to_tal_hdr
	ccan/ccan/tal/tal.c:176
0x556ec0b20e1d tal_steal_
	ccan/ccan/tal/tal.c:498
0x556ec0ab4d6e retry_peer_connected
	connectd/connectd.c:225
0x556ec0b0ee93 next_plan
	ccan/ccan/io/io.c:59
0x556ec0b0fb7c io_do_always
	ccan/ccan/io/io.c:435
0x556ec0b11836 handle_always
	ccan/ccan/io/poll.c:304
0x556ec0b11b2b io_loop
	ccan/ccan/io/poll.c:385
0x556ec0aba611 main
	connectd/connectd.c:2158
0x7f60ee65a30f ???
	???:0
0x7f60ee65a3c0 ???
	???:0
0x556ec0ab08b4 ???
	???:0
0xffffffffffffffff ???
	???:0
lightning_connectd: FATAL SIGNAL (version v0.11.1)
0x556ec0ac8381 send_backtrace
	common/daemon.c:33
0x556ec0ad67bd status_failed
	common/status.c:221
0x556ec0ad6b1a status_backtrace_exit
	common/subdaemon.c:18
0x556ec0ac8432 crashdump
	common/daemon.c:49
0x7f60ee66f55f ???
	???:0
0x7f60ee6bc34c ???
	???:0
0x7f60ee66f4b7 ???
	???:0
0x7f60ee659533 ???
	???:0
0x556ec0b20179 call_error
	ccan/ccan/tal/tal.c:93
[...]

PR #5300, opened 2022-06-01
connectd: avoid use-after-free upon multiple reconnections by a peer

### LINKED ISSUE #5299: Node dies/restarts randomly - probably due to gossipd failing.

The node stops - apparently due to `gossipd` failing and automatically restarts

```
2022-06-01T07:16:33.152Z INFO    028111c2552a2fdd19cde97a18f68f8ece30a8bee45f6070109e6d34115ff9e7ae-chan#4199: Peer transient failure in CHANNELD_NORMAL: Reconnected
2022-06-01T07:16:36.765Z **BROKEN** gossipd: Unknown peer 028873425462e0addf3ad8a2f3ae72ef23b4ea4165d9739134b3ff1a86a0a8eb00 for local_channel_announcement
2022-06-01T07:16:37.189Z **BROKEN** gossipd: Unknown peer 028873425462e0addf3ad8a2f3ae72ef23b4ea4165d9739134b3ff1a86a0a8eb00 for local_channel_announcement
2022-06-01T07:16:37.898Z **BROKEN** gossipd: Unknown peer 036508f7e82bb78bad307cfacf4edf850fc3f20ca071eaa8074d9d5424a9092c0b for local_channel_announcement
2022-06-01T07:16:38.450Z INFO    036aa6d860b1229aa56c4b9422e06bbbfdc160e5d62511187be9a91aedeb9ca309-channeld-chan#7470: Peer connection lost
2022-06-01T07:16:38.450Z INFO    036aa6d860b1229aa56c4b9422e06bbbfdc160e5d62511187be9a91aedeb9ca309-chan#7470: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
2022-06-01T07:16:38.452Z **BROKEN** gossipd: Unknown peer 0225cb6d2874d7cf9ed711479059cf8ff472630188376949879e6253cac00701b0 for local_channel_announcement
2022-06-01T07:16:39.537Z **BROKEN** gossipd: Unknown peer 0343ac307bf48a6500a8529f10e80136c428ac628ec4f7619565dfd447805ccb8c for local_channel_announcement
lightningd: gossipd failed (exit status 2), exiting.
Lost connection to the RPC socket.
Lost connection to the RPC socket.
Lost connection to the RPC socket.
Lost connection to the RPC socket.
Lost connection to the RPC socket.
Lost connection to the RPC socket.
2022-06-01T07:16:41.028Z INFO    connectd: Static Tor service onion address: "xtdo5qvvfwcjaruj6z4acdcw4azagn6tdgac4ajnekjdn4ghr6qw2nqd.onion:9735,0.0.0.0:9735" bound from extern port 9735 
2022-06-01T07:16:41.114Z INFO    plugin-bcli: bitcoin-cli initialized and connected to bitcoind.
2022-06-01T07:16:44.094Z INFO    lightningd: Restarting onchaind for channel 193
```

### `getinfo` output

```
{
   "id": "038fe1bd966b5cb0545963490c631eaa1924e2c4c0ea4e7dcb5d4582a1e7f2f1a5",
   "alias": "zero fee routing | CLN",
   "color": "1c262f",
   "num_peers": 999,
   "num_pending_channels": 2,
   "num_active_channels": 992,
   "num_inactive_channels": 37,
   "address": [
      {
         "type": "ipv4",
         "address": "167.235.3.234",
         "port": 9735
      },
      {
         "type": "torv3",
         "address": "xtdo5qvvfwcjaruj6z4acdcw4azagn6tdgac4ajnekjdn4ghr6qw2nqd.onion",
         "port": 9735
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "v0.11.1",
   "blockheight": 738838,
   "network": "bitcoin",
   // ...
}
```

whitslack:
`peer_reconnected` was freeing a `struct peer_reconnected` instance while a pointer to that instance was registered to be passed as an argument to the `retry_peer_connected` callback function. This caused a use-after-free crash when `retry_peer_connected` attempted to reparent the instance to the temporary context.

Instead, never have `peer_reconnected` free a `struct peer_reconnected` instance, and only ever allow such an instance to be freed after the `retry_peer_connected` callback has finished with it. To ensure that the instance is freed even if the connection is closed before the callback can be invoked, parent the instance to the connection rather than to the daemon.

Absent the need to free `struct peer_reconnected` instances outside of the `retry_peer_connected` callback, there is no use for the `reconnected` hashtable, so remove it as well.

See: https://github.com/ElementsProject/lightning/issues/5282#issuecomment-1141454255
Fixes: #5282
Fixes: #5284
Changelog-Fixed: connectd no longer crashes when peers reconnect.

This may also fix #5299.

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -239,7 +239,8 @@ static struct io_plan *retry_peer_connected(struct io_conn *conn,
 	/*~ Usually the pattern is to return this directly. */
 	return peer_connected(conn, pr->daemon, &pr->id, &pr->addr,
 			      pr->remote_addr,
-			      &pr->cs, take(pr->their_features), pr->incoming);
+			      &pr->cs, take(pr->their_features), pr->incoming,
+			      true);
 }
 
 /*~ If we already know about this peer, we tell lightningd and it disconnects
@@ -334,7 +335,8 @@ struct io_plan *peer_connected(struct io_conn *conn,
 			       const struct wireaddr *remote_addr,
 			       struct crypto_state *cs,
 			       const u8 *their_features TAKES,
-			       bool incoming)
+			       bool incoming,
+			       bool retrying)
 {
 	u8 *msg;
 	struct peer *peer;
@@ -344,9 +346,18 @@ struct io_plan *peer_connected(struct io_conn *conn,
 	bool option_gossip_queries;
 
 	peer = peer_htable_get(&daemon->peers, id);
-	if (peer)
+	if (peer) {
+		/* If we were already retrying, we only get one chance: there
+		 * can be multiple reconnections, and we must not keep around
+		 * stale ones */
+		if (retrying) {
+			if (taken(their_features))
+				tal_free(their_features);
+			return io_close(conn);
+		}
 		return peer_reconnected(conn, daemon, id, addr, remote_addr, cs,
 					their_features, incoming);
+	}
 
 	/* We promised we'd take it by marking it TAKEN above; prepare to free it. */
 	if (taken(their_features))
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -93,6 +93,18 @@ struct connecting {
 	u32 seconds_waited;
 };
 
+/*~ This is an ad-hoc marshalling structure where we store arguments so we
+ * can call peer_connected again. */
+struct peer_reconnected {
+	struct daemon *daemon;
+	struct node_id id;
+	struct wireaddr_internal addr;
+	const struct wireaddr *remote_addr;
+	struct crypto_state cs;
+	const u8 *their_features;
+	bool incoming;
+};
+
 /*~ C programs should generally be written bottom-to-top, with the root
  * function at the bottom, and functions it calls above it.  That avoids
  * us having to pre-declare functions; but in the case of mutual recursion
@@ -230,12 +242,6 @@ static struct io_plan *retry_peer_connected(struct io_conn *conn,
 			      &pr->cs, take(pr->their_features), pr->incoming);
 }
 
-/*~ A common use for destructors is to remove themselves from a data structure */
-static void destroy_peer_reconnected(struct peer_reconnected *pr)
-{
-	peer_reconnected_htable_del(&pr->daemon->reconnected, pr);
-}
-
 /*~ If we already know about this peer, we tell lightningd and it disconnects
  * the old one.  We wait until it tells us that's happened. */
 static struct io_plan *peer_reconnected(struct io_conn *conn,
@@ -252,27 +258,18 @@ static struct io_plan *peer_reconnected(struct io_conn *conn,
 
 	status_peer_debug(id, "reconnect");
 
-	/* If we have a previous reconnection, we replace it. */
-	pr = peer_reconnected_htable_get(&daemon->reconnected, id);
-	if (pr) {
-		peer_reconnected_htable_del(&daemon->reconnected, pr);
-		tal_free(pr);
-	}
-
 	/* Tell master to kill it: will send peer_disconnect */
 	msg = towire_connectd_reconnected(NULL, id);
 	daemon_conn_send(daemon->master, take(msg));
 
 	/* Save arguments for next time. */
-	pr = tal(daemon, struct peer_reconnected);
+	pr = tal(conn, struct peer_reconnected);
 	pr->daemon = daemon;
 	pr->id = *id;
 	pr->cs = *cs;
 	pr->addr = *addr;
 	pr->remote_addr = tal_dup_or_null(pr, struct wireaddr, remote_addr);
 	pr->incoming = incoming;
-	peer_reconnected_htable_add(&daemon->reconnected, pr);
-	tal_add_destructor(pr, destroy_peer_reconnected);
 
 	/*~ Note that tal_dup_talarr() will do handle the take() of features
 	 * (turning it into a simply tal_steal() in those cases). */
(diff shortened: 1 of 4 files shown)


whitslack:
Despite my effort to avoid leaking memory, my `lightning_connectd` process, aside from pegging the CPU at 100% continuously, is also leaking memory like a sieve. After only a day and a half of uptime, the process is using 246 MiB of memory. (`lightningd` is using 1.7 GiB of memory, more than double what v0.10.2 had been using after running for weeks.) I really don't understand why these processes are holding so much state in RAM.

whitslack:
`lightning_connectd` is now using 2.3 GiB of RAM. 😠

Does anyone know if I can undo the database schema changes (manually) and successfully go back to running v0.10.2? Aside from the resource leaks, v0.11.1 is costing me several channels a day due to "internal error" closures. Seems like a really botched release. 😢

rustyrussell:
Re: memleak.  I looked at my node, and the immediate workaround is common/gossip_rcvd_filter.c line 78, change `10000` to `100`.

The question of *why* we're not cleaning this cache correctly will take some diag.

whitslack:
> common/gossip_rcvd_filter.c line 78, change `10000` to `100`.

@rustyrussell: Okay, I'll give that a shot, but I don't see how even 10000 gossip entries in a cache would be consuming 2.5 GiB of memory. That would be an average of 262 KiB per gossip entry.

rustyrussell:
It might not be the only source, but it's definitely the memory hog on my node (it's per-peer, with two caches per peer).

whitslack:
> It might not be the only source, but it's definitely the memory hog on my node (it's per-peer, with two caches per peer).

@rustyrussell: #5312 completely solves it for me. `lightning_connectd` has been using about 13 MiB the entire time it's been running (which is close to a day now).

rustyrussell:
I believe this reintroduces the "multiple connections" bug, where we can have multiple reconnections, and a stale one eventually gets restored (which is what the patch which introduced this htable was avoiding).

One alternative is to have us only try reconnecting once: if we reconnect and it doesn't work, simply free it (by closing the connection).

(I was hoping to get the more significant rewrite in, but it's stalled on a heap of timing changes which broke CI)

whitslack:
@rustyrussell: Let me see if I understand the logic. There can be multiple connections with a given peer but only one "active" connection with any given at any given instant. The first duplicate connection from a peer to arrive will cause the current active connection with that peer to be terminated. While that is occurring, a second duplicate connection from the same peer can arrive. Both duplicate connections will be stalled, waiting for the active connection to be dropped. Once the active connection is dropped, then the first duplicate connection will become the active connection. I think the intention at this point is that this new active connection will then immediately be dropped, and subsequently the second duplicate connection will become the active connection. This effectively forms a FIFO queue, where only the last (i.e., the most recently arrived) duplicate connection is allowed to survive as the active connection.

I am not confident about the timing of the message that is sent to the master daemon to tell it to drop the active connection for the peer. I have a spidey sense that something like the following could occur:

1. First duplicate connection arrives. `connectd` sends message to `lightningd` to disconnect the active connection for this peer.
2. Second duplicate connection arrives. `connectd` sends message to `lightningd` to disconnect the active connection for this peer.
3. `lightningd` receives the message sent in step 1, arranges for the active connection with the peer to be killed, and informs `connectd` that this has been done.
4. `lightningd` receives the message sent in step 2, but there is no active connection for the peer. Whoops.
5. `connectd` receives the confirmation sent in step 3 and turns the first duplicate connection into the active connection.
6. Uhh?? Is the second duplicate connection simply left hanging at this point?

It seems like the intent of the htable was to discard the first duplicate connection (which has not yet become the active connection) when the second duplicate connection arrives, but the code wasn't actually doing that, as it was not actually closing that connection but was merely deallocating the structure that the I/O callback function would later reference.

rustyrussell:
> @rustyrussell: Let me see if I understand the logic. There can be multiple connections with a given peer but only one "active" connection with any given peer at any given instant. The first duplicate connection from a peer to arrive will cause the current active connection with that peer to be terminated. While that is occurring, a second duplicate connection from the same peer can  arrive. Both duplicate connections will be stalled, waiting for the active connection to be dropped. 

So far, so good.

> Once the active connection is dropped, then the first duplicate connection will become the active connection. I think the intention at this point is that this new active connection will then immediately be dropped, and subsequently the second duplicate connection will become the active connection. This effectively forms a FIFO queue, where only the last (i.e., the most recently arrived) duplicate connection is allowed to survive as the active connection.

Well, it's not clear which will actually get in next, and which will go back to waiting.  The intent of the code is that there can only ever be one waiting: if there's one already waiting, take its place.
 
> I am not confident about the timing of the message that is sent to the master daemon to tell it to drop the active connection for the peer. I have a spidey sense that something like the following could occur:
> 
>     1. First duplicate connection arrives. `connectd` sends message to `lightningd` to disconnect the active connection for this peer.
> 
>     2. Second duplicate connection arrives. `connectd` sends message to `lightningd` to disconnect the active connection for this peer.
> 
>     3. `lightningd` receives the message sent in step 1, arranges for the active connection with the peer to be killed, and informs `connectd` that this has been done.
> 
>     4. `lightningd` receives the message sent in step 2, but there is no active connection for the peer. Whoops.
> 
>     5. `connectd` receives the confirmation sent in step 3 and turns the first duplicate connection into the active connection.
> 
>     6. Uhh?? Is the second duplicate connection simply left hanging at this point?

Possibly.  The rewrite eschews all this for a smoother flow of "kill current connection, tell lightningd it's dead, then tell it about new connection like normal".

> It seems like the intent of the htable was to discard the first duplicate connection (which has not yet become the active connection) when the second duplicate connection arrives, but the code wasn't actually doing that, as it was not actually closing that connection but was merely deallocating the structure that the I/O callback function would later reference.

Indeed!  A quick fix would have been to keep the conn pointer in the struct, and free that...

rustyrussell:
I added two minimal fixes to this, which resulted from stress-testing (the test unfortunately hangs frequently, since without other fixes the connect command can race).

This is because I want to do an 11.2 release with this fix and the memleak fix... basically, the "whitslack" point release.

PR #5303, opened 2022-06-02
Expose htlc id to `htlc_accepted` hook

fiatjaf:
I believe this is necessary for low-level stuff that deals with MPP, for example, although I am not entirely sure.

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -1365,6 +1365,8 @@ The payload of the hook call has the following format:
     "next_onion": "[1365bytes of serialized onion]"
   },
   "htlc": {
+    "short_channel_id": "4x5x6",
+    "id": 27,
     "amount": "43msat",
     "cltv_expiry": 500028,
     "cltv_expiry_relative": 10,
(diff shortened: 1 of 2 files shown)


fiatjaf:
Does this make sense? I didn't test yet.

cdecker:
Not quite sure what the use here would be, as `pay` already implements the MPP receiving logic without requiring this extra information. However, I'm also not opposed to including it as it can have use-cases (accounting making sure we didn't miss an HTLC for example?).

fiatjaf:
If one is implementing a bridge between two networks they must be able to relay multiple HTLCs with the same payment hash, and because of that it should be able to know which one to fail/fulfill based on the id, not the payment hash. Or I could have misunderstood everything.

fiatjaf:
But I don't get the `partid` as part of the `htlc_accepted` hook. `partid`s are only visible to the sender and the receiver, no?

Imagine I get two HTLCs for the same hash, cltv expiry and amount from the same channel A to be relayed to the same channel B. The only way I can differentiate the two is by the JSONRPC message I got each, which would be fine except for the fact that if I go offline I'll receive these two `htlc_accepted` again and now I don't know which one is which..

cdecker:
Oh, I see this is related to being able to distinguish them and associate them with downstream effects, without having to differentiate them by amount which may also have a collision. Very good catch, I didn't see this :-)

Sounds good to me ^^

Do you mind adding a tiny `Changelog-Added` to your commit message so it'll show up in the changelog on release too?

ACK https://github.com/ElementsProject/lightning/pull/5303/commits/b3bb188891c18aa139fc8a81ae985e809803b62b

fiatjaf:
OK, done!

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5303/commits/94522c5505f87ce6cecb934ee7b2ef1dec798acd

fiatjaf:
OK, I've also added the `short_channel_id` now as I just realized that wasn't present.

rustyrussell:
Good catch!

Ack https://github.com/ElementsProject/lightning/commit/5404c205444c9482adbd0d39445e850e7f53993f

PR #5304, opened 2022-06-03
Adds option_will_fund info to the RPC listnodes output

litch:
I'm having a lot of trouble with this.  I *have* been able to get the code to run and return the values well, but I am unable to get the schema -> code-generation to create the files correctly.

From what I can tell, I should be able to make a change to the file changed:

doc/schemas/listnodes.schema.json

And that should bubble out first to `.msggen.json`.  That is working correctly.

Then the `contrib/msggen/msggen/__main__.py` loads that file and should output a correct:

- `cln-rpc/src/model.rs`
- `cln-grpc/proto/node.proto`
- `cln-grpc/src/server.rs`
- `cln-grpc/src/convert.rs`

The included changes create new definitions for the `ListnodesNodesOption_will_fund` objects, however, they do not attach that object into the `ListnodesNodes` object.  

I would appreciate any guidance here.

### DIFF
--- a/cln-grpc/proto/node.proto
+++ b/cln-grpc/proto/node.proto
@@ -823,6 +823,15 @@ message ListnodesNodes {
 	repeated ListnodesNodesAddresses addresses = 6;
 }
 
+message ListnodesNodesOption_will_fund {
+	Amount lease_fee_base_msat = 1;
+	uint32 lease_fee_basis = 2;
+	uint32 funding_weight = 3;
+	Amount channel_fee_max_base_msat = 4;
+	uint32 channel_fee_max_proportional_thousandths = 5;
+	bytes compact_lease = 6;
+}
+
 message ListnodesNodesAddresses {
 	// ListNodes.nodes[].addresses[].type
 	enum ListnodesNodesAddressesType {
--- a/cln-grpc/src/convert.rs
+++ b/cln-grpc/src/convert.rs
@@ -618,6 +618,8 @@ impl From<&responses::ListnodesNodes> for pb::ListnodesNodes {
             color: c.color.as_ref().map(|v| hex::decode(&v).unwrap()), // Rule #2 for type hex?
             features: c.features.as_ref().map(|v| hex::decode(&v).unwrap()), // Rule #2 for type hex?
             addresses: c.addresses.as_ref().map(|arr| arr.iter().map(|i| i.into()).collect()).unwrap_or(vec![]), // Rule #3 
+            // Also need codegen'd
+            // option_will_fund: c.option_will_fund.<I'm not certain what goes here>
         }
     }
 }
--- a/cln-rpc/src/model.rs
+++ b/cln-rpc/src/model.rs
@@ -2082,6 +2082,22 @@ pub mod responses {
 	    pub status: PayStatus,
 	}
 
+	#[derive(Clone, Debug, Deserialize, Serialize)]
+	pub struct ListnodesNodesOption_will_fund {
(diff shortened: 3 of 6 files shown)


cln-grpc/proto/node.proto
```diff
@@ -823,6 +823,15 @@ message ListnodesNodes {
 	repeated ListnodesNodesAddresses addresses = 6;
 }
 
+message ListnodesNodesOption_will_fund {
```
vincenzopalazzo, line 826:
```suggestion
message ListnodesNodesOption_will_fund {
```

mh we have a bug in the msg gen? this should be a camel case? or I'm wrong?

cln-rpc/src/model.rs
```diff
@@ -2082,6 +2082,22 @@ pub mod responses {
 	    pub status: PayStatus,
 	}
 
+	#[derive(Clone, Debug, Deserialize, Serialize)]
+	pub struct ListnodesNodesOption_will_fund {
```
vincenzopalazzo, line 2086:
```suggestion
	pub struct ListnodesNodesOption_will_fund {
```

Yeah I think it is a bug

vincenzopalazzo (commented):
>I'm having a lot of trouble with this. I have been able to get the code to run and return the values well, but I am unable to get the schema -> code-generation to create the files correctly.

What is the main difficulty that you have here?

cln-grpc/src/convert.rs
```diff
@@ -618,6 +618,8 @@ impl From<&responses::ListnodesNodes> for pb::ListnodesNodes {
             color: c.color.as_ref().map(|v| hex::decode(&v).unwrap()), // Rule #2 for type hex?
             features: c.features.as_ref().map(|v| hex::decode(&v).unwrap()), // Rule #2 for type hex?
             addresses: c.addresses.as_ref().map(|arr| arr.iter().map(|i| i.into()).collect()).unwrap_or(vec![]), // Rule #3 
+            // Also need codegen'd
+            // option_will_fund: c.option_will_fund.<I'm not certain what goes here>
         }
```
litch, line 623: Thanks @vincenzopalazzo - the main difficulty I'm having is that though we have affordances for the option_will_fund fields (the data structures), I cannot get the codegen process to actually put the option_will_fund onto the ListnodeNode response objects.

cln-rpc/src/model.rs
```diff
@@ -2494,6 +2512,7 @@ pub mod responses {
 
 	#[derive(Clone, Debug, Deserialize, Serialize)]
 	pub struct FeeratesOnchain_fee_estimates {
+		// ^^ Similar naming/casing pattern
```
litch, line 2515: I thought that casing in msggen was weird too, and may be local to my issues, but it looks like several places in the model use the same casing.  Like here.

cln-grpc/src/convert.rs
vincenzopalazzo: I think this is a bug of the `msggen` module because the option should be inside the schema https://github.com/ElementsProject/lightning/blob/74ddc154350ee447841d009de2556b5b6bc11cfb/doc/schemas/listnodes.schema.json#L149-L156 and the `msggen` should translate this as optional parameter, but not familiar on how the msggen works so, maybe I'm wrong

cln-rpc/src/model.rs
vincenzopalazzo: Yes it is an unrelated bug from this PR, maybe we should track with an issue?

Issue #5298, opened 2022-05-31
Non-standard PSBT encoding for partial signatures


devrandom:
### Issue

CLN populates partial signatures with the 64-byte raw signature rather than the actual witness stack entry (DER encoded with sighash type byte appended).  This is non-standard (and fails parsing with `rust-bitcoin` 0.28).

Here is an example of PSBT bytes generated:

`70736274ff0100890200000001c447bec5bd00c02e8db7d725f48767c6e19cd8a2046acc921238d8d52ec1a50e0100000000073bbc80024a01000000000000220020708aae81941aed00dd778008ce5e200ca00b732639d21176324d0de965982938e50e0f0000000000220020e2f341c4c55194e29832b3903d393213dbd5483a850c2fad556dc0ae1c1e26317f693f200001012b40420f0000000000220020e3d09d42ab5170dc9bd6f89027e3a5010ea6dc6ddb89705bb83f145db89e2c5c220202e3bd38009866c9da8ec4aa99cc4ea9c6c0dd46df15c61ef0ce1f271291714e574007fc3f44b61d10f8b26863c739082d75bc56ce99b3506f97d6180ccbb87a3d2bf4358f8bb85f7cc0cc50661e7fc49fd49d3f66a16a10cd3f3c614a98a5c6581e01030401000000010547522102d6cf12d636160228c003a20cacdb80c2e0669ec30792b6627667b80c7b46a2c02102e3bd38009866c9da8ec4aa99cc4ea9c6c0dd46df15c61ef0ce1f271291714e5752ae220602d6cf12d636160228c003a20cacdb80c2e0669ec30792b6627667b80c7b46a2c0082a7a952200000000220602e3bd38009866c9da8ec4aa99cc4ea9c6c0dd46df15c61ef0ce1f271291714e57081abcc1d000000000000101282102d6cf12d636160228c003a20cacdb80c2e0669ec30792b6627667b80c7b46a2c0ac736460b2680001014b6321038e4becfc742f862e0c88c8763fa5fd92beb1aa4f1880ae95083eda5b2092d3286755b2752102dd6877efdb9cac37654aaf7ef3fcf8db24cb88c5ac3277ca5f93ed31315e1be268ac00`

The sequence in question starts with `4007` (length = 0x40 = 64 bytes).

jgriffiths:
@cdecker This isn't a wally bug, please see https://github.com/ElementsProject/lightning/pull/5307 which fixes the encoding from the c-lightning side.

[...]

PR #5307, opened 2022-06-07
psbt: use DER encoded + sighash byte for PSBT_IN_PARTIAL_SIG items

jgriffiths:
This fixes the incorrect signature encoding as reported in https://github.com/ElementsProject/lightning/issues/5298.

I don't see that the old code could ever have worked: whether signing through wally or any other PSBT signer, the resulting signature in the witness would be of an incorrect form and thus should be rejected by core.

Its probably worth getting submission/validation of a tx produced from signing one of these PSBTs under test.


cdecker:
That is likely because we were using the PSBT format as an internal transport format only, and didn't use `libwally` primitives to construct the final transaction. This was discovered when we made the previously internal `hsmd` protocol public to be consumed by VLS.

Thank you very much @jgriffiths, and sorry for causing a false alarm for `libwally`.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5307/commits/d8bde0aee436794b3eacc814c935a93c7ba0da9b

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5307/commits/d8bde0aee436794b3eacc814c935a93c7ba0da9b

jgriffiths:
I've squashed the test commit into the actual change for git-bisectability, no further changes.

cdecker:
Hurray, the bot actually did recognize that there were no content changes between the PR versions and reapplied the ACKs :-)

devrandom:
confirmed compatible with rust-bitcoin 0.28.1

PR #5312, opened 2022-06-08
Connectd memleak fix

rustyrussell:
This may not be the only leak, but it's one.

Reported-by: @whitslack


whitslack:
I'm now running my node with this patch (PR through ea988329cc0b5aa3af1a666d28b9b29657d19ba6), and it's looking good. It may be too early to say yet, but I am not seeing the continual run-up in memory usage by `lightning_connectd` that I was seeing before. It's still pegging a CPU core, though.

whitslack:
It's still looking great. I'd call this one licked.

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/1807be7ccb988f9b2b852d093ea5d33060f7ecb6

Removed one overzealous commit, added Changelog message.

PR #5315, opened 2022-06-09
openingd: Allow users to set the enforced reserve for their counterparty (zeroreserve)

cdecker:
Based on top of #5275, this PR starts at commit f7d0827154302e6920946509df5756a8c79cb017

Until now we always applied the default reserve of 1% of the funding amount, however there has been considerable demand to being able to adjust these values. In particular zeroreserve, i.e., not enforcing any minimal amount in the channel from the counterparty, has been an often requested feature. This PR implements the zeroreserve proposal, in two variants (see second to last commit):
 - The first variant implements the prescribed checks from the spec, ensuring that both the remote and the local fees are above the dust limit, resulting in some dust and at least one forced output for both sides should they have ever reached a sufficiently high balance
 - The second variant skips the prescribed checks, allowing true 0 reserve, and users can drain their channels completely. This is a more trusted setup since the party that gets permission from the other side to fully drain their channel can replay an old state without repercussions

The implementation itself is done via a new `reserve` argument to the single funding RPC commands, and by adding a new `reserve` return argument to the `openchannel` hook. The values are absolute satoshi values, not percentages.

Using arguments and hooks allows the caller or a plugin to selectively opt into allowing the counterparty to drain the channel, based on the nodeid, amount or any other information in the `openchannel` hook (potentially including dynamic lookups). I think this is the best way to allow users to selectively opt-in without them automatically allowing everybody.

Either way a stern warning about the trust involved should be added to the docs.

# Open questions
 - [x] I guess the big one is do we feel ok allowing sub-dust reserves, should it be a user option, and if it's ok, should we propose a spec amendment removing those checks?
   - Amending the spec may require a featurebit since at the moment we actually enforce the sub-dust check for the reserve our peer imposes on us too, making this a breaking change that could fail the channel funding flow. Then again, we presume some level of coordination and trust before attempting to open a zeroreserve channel, so that breakage might be ok.


lightningd/lightningd.h
```diff
@@ -70,6 +70,12 @@ struct config {
 
 	/* EXPERIMENTAL: offers support */
 	bool exp_offers;
+
+	/* Allow dust reserves (including 0) when being called via
+	 * `fundchannel` or in the `openchannel` hook. This is a
+	 * slight spec incompatibility, but implementations do this
+	 * already. */
+	bool allowdustreserve;
```
rustyrussell, line 78: I'm ok with this, but....

We should check we never add a (dust) HTLC which turns all outputs to dust (zero-output txs are invalid).  We should neither propose such a thing (using peer's dust limit), nor accept it.

One simple way is to simply decline to allow opens where the channel size is less than (my-max_accepted_htlcs + their-max_accepted_htlcs + 2) * min(my-dust_limit_satoshis, their-dust_limit_satoshis)?

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -704,7 +704,8 @@ def feerates(self, style, urgent=None, normal=None, slow=None):
     def fundchannel(self, node_id, amount, feerate=None, announce=True,
                     minconf=None, utxos=None, push_msat=None, close_to=None,
                     request_amt=None, compact_lease=None,
-                    mindepth: Optional[int] = None):
+                    mindepth: Optional[int] = None,
+                    reserve: Optional[str] = None):
```
rustyrussell, line 708: needs request schema and doc changes?  Clearly we don't test this by keyword, otherwise CI would barf...

rustyrussell (approved):
OK, generally approve of this, but I think we could replace the dev switches with a generic channel-size check.  You can defer that change until next release if you want though?

We should mention the ability to zeroreserve in the Changelog, if not explicitly called already (I didn't see it, but then GH doesn't exactly highlight commit msgs!)

plugins/libplugin-pay.c
```diff
@@ -1284,11 +1283,21 @@ handle_intermediate_failure(struct command *cmd,
...
 
 	case WIRE_TEMPORARY_CHANNEL_FAILURE: {
+		estimated = errchan->amount;
+
+		/* Subtract one msat more, since we know that the amount did not
+		 * work. This allows us to then allow on equality, this is for
+		 * example necessary for local channels where exact matches
+		 * should be allowed. */
+		if (amount_msat_eq(estimated, AMOUNT_MSAT(0)) ||
+		    !amount_msat_sub(&estimated, estimated, AMOUNT_MSAT(1)))
```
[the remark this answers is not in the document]
cdecker: I might have been overly cautious, yes, since we never send out an HTLC with 0msat as value.

contrib/pyln-client/pyln/client/lightning.py
cdecker: The request schema for `fundchannel` is in #5362, hence why this isn't covered by the test yet.

lightningd/lightningd.h
rustyrussell: No. Without a reserve, *all* our outputs could be dust, which is illegal.

contrib/pyln-client/pyln/client/lightning.py
rustyrussell: 👍

niftynei:
Deferring to next release pending small fixes.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5315/commits/afed054c8750d716341299800c3499a04aa5f8ab

PR #5316, opened 2022-06-09
Warning added to PLUGIN documentation re: `stdin`

litch:
Adds some warning/documentation about the `println!`/`dbg!` causing my bad times as a plugin developer.  =)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5316/commits/8b656dd184054ae06207c455ad71166ffad87d2d

This is a good pint to add, because I had some report that fall exactly into this

PR #5321, opened 2022-06-15
add 'perl' dependency for 'build on fedora'

NicolaLS:
Building on Fedora Linux 36 (Workstation Edition) x86_64 failed with an error pointing me to [this](https://github.com/openssl/openssl/issues/13761#issue-777412186). Installing perl fixed it for me

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -90,6 +90,7 @@ $ sudo dnf update -y && \
                 'C Development Tools and Libraries' \
                 'Development Tools' && \
         sudo dnf install -y \
+	        perl \
                 clang \
                 gettext \
                 git \


vincenzopalazzo (approved):
utACK  [45eb9db](https://github.com/ElementsProject/lightning/pull/5321/commits/45eb9db921e0629c2dd1f7e72f9e3bbc5c0ea224)

doc/INSTALL.md
```diff
@@ -90,6 +90,7 @@ $ sudo dnf update -y && \
                 'C Development Tools and Libraries' \
                 'Development Tools' && \
         sudo dnf install -y \
+	        perl \
```
wtogami, line 93: tested NACK - see notes below

wtogami:
Overall the Fedora build instructions in that doc are old. I'd like to test it with a minimal build environment of Fedora 36, RHEL8 and RHEL9, and do a big cleanup PR.

wtogami:
I'm building CLN just fine on Fedora 35 and 36 without perl installed. If you're referring to `FindBin.pm` in particular like in the link you referenced that is not contained in Fedora's `perl` package but rather `perl-FindBin`.

Let's figure out why my results without perl installed builds CLN and you have different results.

Could you please post the exact build errors you're seeing? Can I see the output of `./configure` before you build?

wtogami (changes requested):

NicolaLS:
> 

Oh ok, I don't know if this is easy reproducible I tried a lot of things which I can't remember so maybe something else fixed it (but I really don't think so). Would it be sufficient to just de-install perl, try again and post the ./configure output ?

NicolaLS:
Ok, so I now uninstalled lightningd and perl and did the build process again and everything worked just fine..really weird because adding perl was really the "last action" that fixed it last time.

Issue #5301, opened 2022-06-01
connectd pegging 100% CPU


whitslack:
### Issue and Steps to Reproduce
Running v0.11.1, `lightning_connectd` is utilizing 100% of a CPU core. Attaching `strace` to the process reveals that it is hammering `pread64` calls on the `gossip_store` file. The reads are all very small, most exactly 12 bytes in size, and none in my sample even approached the page size. Why is connectd not memory-mapping this file for performance? Syscalls are to be minimized wherever possible.

rustyrussell:
First step is not to do dumb things, second step is to optimize dumb things :)

We scan for two reasons: first, when they connect we send them any gossip we made ourselves.  We do this in a naive way, by scanning the entire store.  Fix: put our own gossip in a separate store file, which is what @cdecker wants to share gossip files anyway.  This adds some gossmap complexity, however, which now needs to handle two files.

Second, when they send a gossip_timestamp_filter message, we scan the entire store to see if any match the filter they've given.  But it's not unusual to send a dummy filter to say "I don't want anything from you": LND and CLN both use 0xFFFFFFFF for this, so I've optimized that.

whitslack:
v0.11.2 might be ever so slightly better on this issue, but `lightning_connectd` is still hovering around 90% CPU utilization.

Is it expected to have a constant parade of
```
INFO    ...-channeld-chan#233342: Peer connection lost
INFO    ...-chan#233342: Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died (62208)
```
and
```
INFO    ...-chan#153417: Peer transient failure in CHANNELD_NORMAL: Reconnected
```
in the log? I know I have a lot of peers, but the rate of link flapping still seems excessive. And how is "Reconnected" ever the explanation of a "Peer transient failure"? That seems fishy to me.

Also, I have noticed that after some time, my node stops accepting incoming connections entirely. I thought it was happening due to a file descriptor mixup while running under Heaptrack, but it happened again even with no Heaptrack in the mix.

rustyrussell:
These log messages are overzealous.  I'm seeing the "Peer transient failure in CHANNELD_NORMAL: channeld: Owning subdaemon channeld died" mainly when peers don't respond to pings (which is a message you only get at DEBUG level, but should probably be INFO).

With 21 connected peers, I am seeing 36 of these in 24 hours (yeah, one peer responsible for over half).

[...]

PR #5328, opened 2022-06-20
common/gossip_store: optimize case where entries are filtered out.

rustyrussell:
@whitslack complained of large CPU usage by connectd at startup;
I ran perf record on connectd on my machine (which sees a little spike, only)
and I see the cost of reading and discarding the entries:

```
-   95.52%     5.24%  lightning_conne  lightning_connectd  [.] gossip_store_next
   - 90.28% gossip_store_next 
      + 40.27% tal_alloc_arr_ 
      + 22.78% tal_free
      + 11.74% crc32c
      + 9.32% fromwire_peektype
      + 4.10% __libc_pread64 (inlined)
        1.70% be32_to_cpu
```

Much of this is caused by the search for our own gossip: keeping this separately
would be even better, but this fix is minimal.

### DIFF
--- a/common/gossip_store.c
+++ b/common/gossip_store.c
@@ -77,8 +77,16 @@ u8 *gossip_store_next(const tal_t *ctx,
 			continue;
 		}
 
-		checksum = be32_to_cpu(hdr.crc);
+		/* Skip any timestamp filtered */
 		timestamp = be32_to_cpu(hdr.timestamp);
+		if (!push &&
+		    !timestamp_filter(timestamp_min, timestamp_max,
+				      timestamp)) {
+			*off += r + msglen;
+			continue;
+		}
+
+		checksum = be32_to_cpu(hdr.crc);
 		msg = tal_arr(ctx, u8, msglen);
 		r = pread(*gossip_store_fd, msg, msglen, *off + r);
 		if (r != msglen)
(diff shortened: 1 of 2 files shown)


whitslack:
@rustyrussell: To be clear, it's not just at startup. `lightning_connectd` **never** settles down. I started my current Core Lightning process 171 hours ago, and `lightning_connectd` has used 152 hours of CPU time and is still pegging. That's an average of 88.9% CPU utilization over the lifetime of the process.

rustyrussell:
We will do this scan every time someone reconnects: for me that's worst at startup, but if you have enough peers....

This patch should drop it by a factor of ~10.  We can get more sophisticated (keep our own gossip separately for this case) but that's a more invasive change which risks introducing other bugs.

cdecker:
> This patch should drop it by a factor of ~10. We can get more sophisticated (keep our own gossip separately for this case) but that's a more invasive change which risks introducing other bugs.

That would be great for sharing/copying a single gossip file (for bootstrapping) among multiple nodes as the non-local gossip messages would not be intermingled with the public ones. :+1:

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5328/commits/b28adf1b4ecf1d2f59dee8c58cb1d79d2240115b

vincenzopalazzo (approved):
ACK [b28adf1](https://github.com/ElementsProject/lightning/pull/5328/commits/b28adf1b4ecf1d2f59dee8c58cb1d79d2240115b)

rustyrussell:
Not sure how much this helps #5301, which this should have referred to!

PR #5330, opened 2022-06-21
listforwards localfail out_channel

rustyrussell:
There are more cases where we can populate out_channel; we definitely need to fix the case where we failed locally because the outgoing channel didn't have capacity, but we should also try anywhere that they gave us a channel to fwd to, even if it doesn't exist, for example.


rustyrussell:
Rebase after scid aliases!

rustyrussell:
Fixed final patch: we can't set hout in db if it does not exist, as we get "Foreign constraint failed".

rustyrussell:
Rebased again, removed unrelated fixup which clashed with master rebase.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5330/commits/090044a1ecdf19a470c11f8b2ea05d59e9ed6c03

lgtm, test failure is unrelated.

PR #5335, opened 2022-06-22
pyln: Fix relative path dependencies when publishing to PyPI

### LINKED ISSUE #5246: pyln-client: error install the last version available on pypi 0.10.2.post1
This is what I'm trying on a clean machine

```
(lnprototest-AvqpilGs-py3.10) ➜  clightning git:(master) pip3 install pyln-client            
Requirement already satisfied: pyln-client in /home/vincent/.cache/pypoetry/virtualenvs/lnprototest-AvqpilGs-py3.10/lib/python3.10/site-packages (0.10.2.post1)
WARNING: Error parsing requirements for pyln-testing: Invalid URL: ../pyln-client
(lnprototest-AvqpilGs-py3.10) ➜  clightning git:(master) pip3 install pyln-client==0.10.2
ERROR: Could not find a version that satisfies the requirement pyln-client==0.10.2 (from versions: 0.0.7.3, 0.7.3, 0.8.0, 0.8.2, 0.9.2, 0.9.3, 0.10.0, 0.10.1, 0.10.2.post0, 0.10.2.post1, 0.10.2.post3)
ERROR: No matching distribution found for pyln-client==0.10.2
(lnprototest-AvqpilGs-py3.10) ➜  clightning git:(master) pip3 install pyln-client==0.10.1
Collecting pyln-client==0.10.1
  Downloading pyln_client-0.10.1-py3-none-any.whl (24 kB)
Collecting recommonmark>=0.7.*
  Downloading recommonmark-0.7.1-py2.py3-none-any.whl (10 kB)
Collecting docutils>=0.11
```

In particular, the actual version point to some local path `WARNING: Error parsing requirements for pyln-testing: Invalid URL: ../pyln-client`

cdecker:
So this was quite a journey:

 - We want relative depdendencies (using the `path` argument) whenever
   developing locally. Otherwise we would have to install each
   dependency every time we change a single character, which
   undoubtedly would cause us to waste time trying to debug an issue
   just because we forgot to install.
 - When publishing however we want to rely on the version number,
   since the repo context gets lost upon publishing, and path
   dependencies cause failures.

The solution then it seems is to use `dev-dependencies` (not that
surprising once you find it) with relative paths, so that `poetry
install` uses these over the normal dependencies (no idea how they
dedup them) and use `dependencies` when publishing. The paths are
still in there when publishing, but `pip install` ignores them.

I checked that `poetry install` from an unrelated project doesn't
accidentally use the path dependencies, even when adding them as
dev-dependencies. This should hopefully also allow installing them
as a repo link, though I can't test that right now.

Fixes #5246

### DIFF
--- a/contrib/pyln-client/pyproject.toml
+++ b/contrib/pyln-client/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pyln-client"
-version = "0.11.0"
+version = "0.11.1"
 description = "Client library and plugin library for Core Lightning"
 authors = ["Christian Decker <decker.christian@gmail.com>"]
 license = "BSD-MIT"
@@ -13,10 +13,12 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.7"
 pyln-bolt7 = "^1.0"
-pyln-proto = "^0.10.2"
+pyln-proto = "^0.11"
 
 [tool.poetry.dev-dependencies]
 pytest = "^7.0.1"
+pyln-bolt7 = { path = "../pyln-spec/bolt7", develop = true }
+pyln-proto = { path = "../pyln-proto", develop = true}
 
 [build-system]
 requires = ["poetry-core>=1.0.0"]
--- a/contrib/pyln-proto/pyproject.toml
+++ b/contrib/pyln-proto/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pyln-proto"
-version = "0.11.0"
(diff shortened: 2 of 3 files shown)


vincenzopalazzo (approved):
I have no experience with the relative path and poetry, so for me is an ACK

ACK https://github.com/ElementsProject/lightning/pull/5335/commits/96e884d5faf0ee6d4efa64d1c22c839a5d057539

BTW, we can grab this issue https://github.com/ElementsProject/lightning/issues/5246 to close it?

Issue #5337, opened 2022-06-23
CLN crashes at irregular intervals (peer_control.c?)


zerofeerouting:
# Logs

It looks like, there is something about `lightningd/peer_control.c:1193`. Please let me know if I should provide the complete crashdumps (they're huge) or if I can be of assistance in some other way.

## 2022-06-22 17:12 UTC

```
+46870.883995596 channeld-chan#10988DEBUG: rcvd_commit: HTLC REMOTE 80103 = RCVD_REMOVE_ACK_COMMIT/SENT_REMOVE_ACK_COMMIT FAILED
+46870.884247080 gossipdDEBUG: seeker: disabling gossip
+46870.884586292 hsmdDEBUG: Got WIRE_HSMD_VALIDATE_COMMITMENT_TX
+46870.909893699 lightningdBROKEN: FATAL SIGNAL 6 (version v0.11.1-9-g3a763be)
+46870.909905639 lightningdBROKEN: backtrace: common/daemon.c:38 (send_backtrace) 0x555a3524895d
+46870.909908169 lightningdBROKEN: backtrace: common/daemon.c:46 (crashdump) 0x555a3524899f
+46870.909910799 lightningdBROKEN: backtrace: (null):0 ((null)) 0x7f7881e7872f
+46870.909913019 lightningdBROKEN: backtrace: (null):0 ((null)) 0x7f7881c737bb
+46870.909915509 lightningdBROKEN: backtrace: (null):0 ((null)) 0x7f7881c5e534
+46870.909917929 lightningdBROKEN: backtrace: (null):0 ((null)) 0x7f7881c5e40e
+46870.909920369 lightningdBROKEN: backtrace: (null):0 ((null)) 0x7f7881c6c101
+46870.909923368 lightningdBROKEN: backtrace: lightningd/peer_control.c:1193 (peer_connected) 0x555a3521c44d
+46870.909926058 lightningdBROKEN: backtrace: lightningd/connect_control.c:451 (connectd_msg) 0x555a351fde93
+46870.909934828 lightningdBROKEN: backtrace: lightningd/subd.c:556 (sd_msg_read) 0x555a3522a9bc
+46870.909937878 lightningdBROKEN: backtrace: ccan/ccan/io/io.c:59 (next_plan) 0x555a35295af6
+46870.909940148 lightningdBROKEN: backtrace: ccan/ccan/io/io.c:407 (do_plan) 0x555a35295feb
[...]

PR #5338, opened 2022-06-23
lightningd: remove overzealous assertion.

rustyrussell:
This is being hit: it's possible if connectd and lightningd get desynchronized,
and we'll handle this later when peer is activated.

Fixes: #5337 
Changelog-None


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5338/commits/11ac9802a05bf17059ab0e34964aab9daa9f7674

Issue #5305, opened 2022-06-06
Wrong port is gossiped, in addition to correct one


laanwj:
I'm having an issue where the gossiped port is wrong, possibly part of the time. It's supposed to be 9736, but 9735 ends up in the list too:
```bash
% lcli listnodes [M]
{
   "nodes": [
      {
         "nodeid": "[M]",
…
         "addresses": [
            {
               "type": "ipv4",
               "address": "[A]",
               "port": 9735
            },
            {
               "type": "ipv4",
               "address": "[A]",
               "port": 9736
            },
            {
               "type": "ipv6",
               "address": "[B]",
               "port": 9736
            },
            {
               "type": "torv3",
               "address": "[C].onion",
               "port": 9736
            }
         ]
      }
   ]
}
```
I not only see this in my own `listnodes`, other nodes and services do, as well. I do not have anything running on port 9735, FWIW, there's only one lightningd on this system. I've only ever seen it happen for IPv4.

Relevant configuration is:
```bash
    --proxy 127.0.0.1:9050 \
    --addr '0.0.0.0:9736' \
    --addr '[::]:9736' \
    --announce-addr '[C].onion:9736' \
```

`getinfo` output shows, correctly:
```bash
{
   "id": "[M]",
…
   "address": [
      {
         "type": "ipv4",
         "address": "[A]",
         "port": 9736
      },
      {
         "type": "ipv6",
         "address": "[B]",
         "port": 9736
      },
      {
         "type": "torv3",
         "address": "[C].onion",
         "port": 9736
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9736
      },
      {
         "type": "ipv6",
         "address": "::",
         "port": 9736
      }
   ],
   "version": "v0.11.0.1-36-g74ddc15",
…
}
```

laanwj:
amboss.space has a history for the node that shows 9735 appearing and disappearing, it's really strange:

![amboss](https://user-images.githubusercontent.com/126646/172310341-2282917a-77be-4ae6-817d-424264c83231.png)

None of these are intentional configuration changes.

vincenzopalazzo:
Mh! I did not debug it, and I put here only a pointer that we had changed somethings in the default port but there was some refusal that this PR will remove https://github.com/ElementsProject/lightning/pull/5242

Maybe can help

laanwj:
Thanks @kroese and @vincenzopalazzo .

[...]

PR #5344, opened 2022-06-25
gossipd: Turn off IP discovery when we already have useable addresses.

m-schmoock:
Currently, IP discovery will be active (unless turned off) also when users configure announced addresses manually.
When the manual configuration uses a non-default port, this will lead to the strange behavior of announcing the same address twice. One with the correct and one with the guessed default port.

The idea is that we turn off IP discovery when a usable announcement has been defined via commandline, config or autobind.
Also this PR sets the 'guessed' port depending on what network (Mainnet, Testnet, ...) we are on.

### Followup
- Make the switch `--disable-ip-discovery` obsolete it and maybe add a `--ip-discovery` `true`/`false`(/`auto`) switch to force it on or off or keeping it automatically, if not defined.

### Note:
We could/should also enable IP discovery when we only have a working  TOR address (but without --always-use-proxy ofc). This will give nodes an option to have a bootstrap way to be reachable until IP discovery can do the job in a more stable way.
Keep in mind its easier for most users to punch a hole in a NAT router than to setup TOR correctly.
If/Once we have a `--ip-discovery=true` switch the user can also do that manually if he wants to.

Addresses https://github.com/ElementsProject/lightning/issues/5305


gossipd/gossipd.c
```diff
@@ -350,8 +350,10 @@ static void handle_remote_addr(struct daemon *daemon, const u8 *msg)
 	if (!fromwire_gossipd_remote_addr(msg, &remote_addr))
 		master_badmsg(WIRE_GOSSIPD_REMOTE_ADDR, msg);
 
-	/* current best guess is that we use DEFAULT_PORT on public internet */
-	remote_addr.port = DEFAULT_PORT;
+	/* We can only guess the public port here, so we use DEFAULT_PORT.
+	 * Also the DEFAULT_PORT differs on which network is used.
+	 * So we do it similar to what Bitcoin Core does */
+	remote_addr.port = DEFAULT_PORT + chainparams->rpc_port - 8332;
```
vincenzopalazzo, line 356: We should keep our eyes on https://github.com/ElementsProject/lightning/pull/5242 will be a new chain value of the default port, and remove these tricky calculation :)

vincenzopalazzo (commented):
Concept ACK

rustyrussell:
> This will make the switch '--disable-ip-discovery' more or less pointless (except for edge cases). We can think about adding a '--enable-ip-discovery' switch instead or also.

This is why --enable/disable flags are generally poor design: --ip-discovery=<bool> is better, since it can be overridden later in commandline too.

niftynei:
@m-schmoock we're looking to cut an RC this week, and I'd love to get this added. It looks like this just needs a rebase.

m-schmoock:
> @m-schmoock we're looking to cut an RC this week, and I'd love to get this added. It looks like this just needs a rebase.

Ok, I'll try to finish it. It's a little bit more though because we also need to depreciate the '--disable-ip-discovery' flag, update docs and tests etc.

m-schmoock:
Note: Needed to fix some outdated gossmap tests that broke CI

m-schmoock:
@rustyrussell 
So now IP discovery is only active if no other addresses are defined (or usable). I would like the option to have a TOR address announced but also IP discovery enabled. How can we achieve that?

1. TODO anyway: Deprecate `--disable-ip-discovery`. Whats the proper way of deprecating a paramtere? Just remove it or raise a warning/error?
2. Add a `--ip-discovery=true/false(/auto)` switch. But this can't be a strict bool, because we want IP discovery to kick in as a fallback when no addresses are defined. Therefore we don't want to default to `true` nor do we want to default to `false` but `auto`. Whats the proper way to do that? For example in `lightningd/lightningd.h`:
```C
enum config_autobool {
    AUTOBOOL_AUTO,
    AUTOBOOL_ON,
    AUTOBOOL_OFF
};

// ... plus the corresponding helpers to register this as an arg...
```
3. Alternatively to 2. we can have it as a string arg with value checking. But that's ugly too for something that's more like an `optional<boolean>`.

rustyrussell:
Number 2 is correct.  And default to AUTO.

m-schmoock:
> Number 2 is correct. And default to AUTO.

Okay will add that.
And how about the deprecation of the `--disable-ip-discovery` flag? just drop it or fail/warn?

rustyrussell:
> > Number 2 is correct. And default to AUTO.
> 
> @rustyrussell Okay will add that. And how about the deprecation of the `--disable-ip-discovery` flag? just drop it or fail/warn?

You need to deprecate it: accept it if deprecated_apis is false with a log_unusual() warning, put it in Changelog, remove it from documentation...

rustyrussell (approved):
Ack f02e510172d2b21306facb184e632c8dd05bf025

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5344/commits/f02e510172d2b21306facb184e632c8dd05bf025

PR #5346, opened 2022-06-25
plugin: add check on the type json object during the IO message handling

vincenzopalazzo:
During the debugging of a summer of the bitcoin project with @swaptr, core lightning lies in the error message when receiving a notification with an invalid JSON, it tells that the object did not contains the `jsonrpc` filed, but if I send a string with a valid json inside like `"{valid json rpc 2.0 notification}"` cln should tell us that we are sending somethings that it is not a json object

Having a better error message can speed up the debugging process, and from our point of view it is just an `if`

Still thinking about how to make a test for this, any suggestions?

### DIFF
--- a/lightningd/plugin.c
+++ b/lightningd/plugin.c
@@ -643,6 +643,11 @@ static const char *plugin_read_json_one(struct plugin *plugin,
 		return NULL;
 	}
 
+	if (plugin->toks->type != JSMN_OBJECT)
+		return tal_fmt(
+		    plugin,
+		    "JSON-RPC message is not a valid JSON object type");
+
 	jrtok = json_get_member(plugin->buffer, plugin->toks, "jsonrpc");
 	idtok = json_get_member(plugin->buffer, plugin->toks, "id");
 


rustyrussell:
I don't think it really needs a test (though you could write one in python, of course: just read stdin until you see "\n\n", then send junk, then read until you see "\n\n".  Put everything you read in a temporary file.  Test can check contents?

But rebase on master might fix flakes!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5346/commits/b14d8d061950e1eb406f3a322861aac065bfa057

vincenzopalazzo:
>I don't think it really needs a test (though you could write one in python, of course: just read stdin until you see "\n\n", then send junk, then read until you see "\n\n". Put everything you read in a temporary file. Test can check contents?

You just fix the flakiness :) maybe we can avoid testing this?

lightningd/plugin.c
```diff
@@ -638,6 +638,11 @@ static const char *plugin_read_json_one(struct plugin *plugin,
 		return NULL;
 	}
 
+	if (plugin->toks->type != JSMN_OBJECT)
```
SimonVrouwe, line 641: I am no expert, but isn't this already checked in `json_parse_input` ?

https://github.com/ElementsProject/lightning/blob/b618d662bcb9b7040a2df6d6582fa9fd35073db8/lightningd/plugin.c#L618-L624

vincenzopalazzo: Yes! but the message is wrong! 

We fail to parse the JSON but messages that are not JSON at all and this makes the debug very bad and difficult, as mentioned in the following sentences

>Having a better error message can speed up the debugging process, and from our point of view it is just an if

vincenzopalazzo:
trivial rebase on master

lightningd/plugin.c
SimonVrouwe: Ok I see it now,  `json_parse_input` doesn't catch that error.

Tested with a build-in plugin on the command line (which uses the same [json_parse_input](https://github.com/ElementsProject/lightning/blob/f67ab2a86e5d4543570925d9efafcb2b56a6c8d3/plugins/libplugin.c#L704-L707) and I enter your line:
```
./autoclean 
"{valid json rpc 2.0 notification}"
Malformed JSON-RPC notification missing "method" or "params": "{valid json rpc 2.0 notification}"
```
That error is from a later step, passed  `json_parse_input`.

To test manually, maybe modify [cowsay.sh](https://github.com/ElementsProject/lightning/blob/master/contrib/plugins/cowsay.sh)?

vincenzopalazzo: If you put this line in the command line `"{valid JSON rpc 2.0 notification}"` you are passing a string to the stdio?

BTW the body of the json need to contains a valid json, in this case `\"{jsonrpc: \"2.0\", id: \"1\", \"method\": \"getinfo\", \"params\": \"\"}\"`

The `json_parse_input` parse ok all the JSON received because the string is a valid JSON, and if you repeat the test your test should be ok (my string example may contain typo)

Here the choice to put the check-in there is to make a minimal change to the repository because this is to increase only a debug verbosity to avoid some common mistakes during a plugin API development, and I think that putting this change inside a json_parse_input is a wrong place? maybe we need to have some sanity check file regarding the JSON RPC protocol?

BTW, the string `{any plain content` is the wrong JSON format, and the JSON library is right!

SimonVrouwe: > you are passing a string to the stdio

Yep just copy paste the line into the terminal, also handy for debugging plugins separately.
Forget about the test, `cowsay.sh` was just a loose suggestion (but its an insightful example).

Please carry on.
Ack  90f6df5

PR #5349, opened 2022-06-27
Log filtering by node id.

rustyrussell:


rustyrussell:
Ack 9d0e1d834692a18ab7cf9a5f6e997b495b009cd1

Trivial rebase on master.

lightningd/log.c
```diff
@@ -291,13 +291,15 @@ struct log_book *new_log_book(struct lightningd *ld, size_t max_mem)
 }
 
 static enum log_level filter_level(struct log_book *lr,
-				   const struct log_prefix *lp)
+				   const struct log_prefix *lp,
+				   const struct node_id *node_id)
 {
 	struct print_filter *i;
+	const char *node_id_str = node_id ? node_id_to_hexstr(tmpctx, node_id) : "";
```
cdecker, line 303: I wonder if doing it the other way around, i.e., converting the prefix to binary and matching at byte-level, wouldn't be more efficient, rather than converting the node_id to hex over and over again. That's with a caveat though that the prefix would need to have an even length.

Alternatively we could also keep a list of prior matches.

doc/index.rst
```diff
@@ -114,3 +114,4 @@ Core Lightning Documentation
    lightning-withdraw <lightning-withdraw.7.md>
    lightningd <lightningd.8.md>
    lightningd-config <lightningd-config.5.md>
+   lightningd-json <lightningd-json.7.md>
```
cdecker, line 117: This file does not exist.

cdecker:
One spurious change that points to a non-existing file, and some musings re efficiency. At some point we'll have reinvented the structured logging ideas from a decade ago, and we'll just let the operator give us an expression to filter by and a `printf` string to format the various components xD

cdecker (changes requested):
That spurious change needs to be removed

doc/index.rst
rustyrussell: Yes, was on a different branch, it didn't get re-regenerated when I switched.  Gah! thanks

lightningd/log.c
rustyrussell: We can generate this on-stack pretty fast, but I think optimizing logging CPU to this level is probably overkill for now?

rustyrussell:
Rebase on master, hoping to fix prototest flake!

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5349/commits/81a76990c3cfb0432bdac31fd01a98dd1602e26a

Today looks like no the day of the ssh!

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/5349/commits/81a76990c3cfb0432bdac31fd01a98dd1602e26a

PR #5357, opened 2022-06-29
Fix broken link

bubelov:


doc/lightning-createonion.7.md
```diff
@@ -132,5 +132,5 @@ RESOURCES
 
 Main web site: <https://github.com/ElementsProject/lightning>
 
-[bolt04]: https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md
+[bolt04]: https://github.com/lightning/bolts/blob/master/04-onion-routing.md
```
vincenzopalazzo, line 135: nit: It is not broken the `https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md` 
 will resolve in `https://github.com/lightning/bolts/blob/master/04-onion-routing.md` is only a redirection due the spec rebranding

bubelov: @vincenzopalazzo are you sure?

Screenshot from 2022-06-29 19-45-38

bubelov: Something weird is going on here. The link is clearly broken, but it could be due to static site generator stripping `.md` or something

vincenzopalazzo: yes, it is wrong the link that you click in some way! but  if you click [here](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing) you will be redirected correctly.

Maybe you can use the []() format instead the current format with label

bubelov: > but if you click [here](https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing) you will be redirected correctly

nope, it's 404 =)

bubelov: In fact, most of the links in the doc are currently broken

bubelov:
Maybe URL inlining would work? It works fine on GH, but I'm not sure about `readthedocs` though

bubelov:
Well, at least we tried =) Looks like this problem is a bit deeper

PR #5359, opened 2022-06-30
CI: re-enable checks, by changing errant tab back to spaces.

rustyrussell:
FFS GH!

Changelog-None


rustyrussell:
Fixed better in #5364

PR #5362, opened 2022-06-30
pytest: Add grpc based CLN client to test grpc coverage

cdecker:
The goal here is to extend test coverage to the grpc interface. We can do so by adding a new `LightningGrpc` class and swapping out the `LightningRpc` class if configured to do so. Swapping out then means that instead of talking directly to the JSON-RPC:

```
CLN -- JSON-RPC -- LightningRpc -- pytest
\_____CLN_____/    \_______pytest_______/
```
we now take a bit of a detour:

```
CLN -- JSON-RPC -- cln-rpc -- rpc2grpc converters -- grpc interface -- python grpc client -- python grpc2json converter -- pytest
\_____CLN_____/    \___________cln-grpc-plugin____________________/    \__________________________pytest________________________/
```

It's a bit roundabout, but it allows us to test all of the components along the way without having to change the tests themselves, and after all they should expose the same functionality.

To opt-in to the grpc testing for now you have to set an environment variable `CLN_TEST_GRPC=1`, and the `LightningNode` will swap out the RPC implementation.

**Caveat**: the conversion is sometimes lossy, i.e., some fields aren't converted (yet), and some RPC methods aren't implemented (and some may never be). Hence some tests will simply not be possible to run at the moment, but I am working on extending coverage, both method-wise as well as field-wise, just be warned that some


doc/schemas/stop.request.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "psbt"
+  ],
```
rustyrussell, line 7: I think this is wrong!

rustyrussell: Also, we clearly don't test this properly.  Why did it pass CI?

doc/schemas/fundchannel.request.json
```diff
@@ -40,6 +40,10 @@
       "items": {
         "type": "outpoint"
       }
+    },
+    "mindepth": {
+      "description": "Number of confirmations required before we consider the channel active",
+      "type": "u32"
     }
```
rustyrussell, line 47: THis implies we never tested this parameter (or at least, never as a named parameter!).

rustyrussell (changes requested):
One cut & paste bug, and some questions about our lack of test coverage which allowed these schema issues to slip through CI!

But other than that, looks good!

doc/schemas/fundchannel.request.json
cdecker: The issue here is that the `fundchannel.request.json` file is new from this PR (the `fundchannel.schema.json` is the one that was tested).

When I backfilled it early in the PR we didn't have the `zeroconf` PR merged yet, hence why we needed to add it only later. And indeed CI and `make pytest` surfaced this lack of parameter in the first version of the JSON schema, hence why I added it here.

TL;DR: Things are working as expected, and we even verified that a missing param does trigger CI warnings.

doc/schemas/stop.request.json
cdecker: Curious indeed, so I went digging a bit:

https://github.com/ElementsProject/lightning/blob/6204d70a3729570f019ef7269fca79211c1e6350/contrib/pyln-testing/pyln/testing/utils.py#L655-L658

The `payload` is `None` in this case, hence falsifying the third clause, and we skip schema validation. Even if we did check the schema, we'd run into this:

https://github.com/ElementsProject/lightning/blob/6204d70a3729570f019ef7269fca79211c1e6350/contrib/pyln-testing/pyln/testing/utils.py#L856-L861

So we might want to relax that `if` statement and only catch a specific error type instead.

cdecker: Will fix up the schema though ^^

PR #5374, opened 2022-07-04
fix typo in commando documentation

zerofeerouting:


zerofeerouting (commented):
Fixed a tiny typo.

vincenzopalazzo:
your change is missed inside the quantity of commits, but I can still click on the single commit.

However, do you think that the better place for this would be a core review in the Rusty PR?

rustyrussell:
Found it!  Since it's not committed yet, we can just leave a comment on the PR.  But I've applied it now!

Thanks!

PR #5378, opened 2022-07-05
peer_control: Fix check_funding_details assert

ddustin:
Check funding_outnum validity first to avoid reading invalid outputs.

Changelog-Fixed: Fixed a potential issue if the number of outputs decreases in a dualopen RBF or splice.

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1437,13 +1437,14 @@ static bool check_funding_details(const struct bitcoin_tx *tx,
 				  struct amount_sat funding,
 				  u32 funding_outnum)
 {
-	struct amount_asset asset =
-	    bitcoin_tx_output_get_amount(tx, funding_outnum);
-
-	if (!amount_asset_is_main(&asset))
-		return false;
+	struct amount_asset asset;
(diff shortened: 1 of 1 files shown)


lightningd/peer_control.c
```diff
@@ -1437,13 +1437,14 @@ static bool check_funding_details(const struct bitcoin_tx *tx,
 				  struct amount_sat funding,
 				  u32 funding_outnum)
 {
-	struct amount_asset asset =
-	    bitcoin_tx_output_get_amount(tx, funding_outnum);
+	struct amount_asset asset;
 
-	if (!amount_asset_is_main(&asset))
+	if (funding_outnum >= tx->wtx->num_outputs)
```
vincenzopalazzo, line 1442: do you think that in this case can be necessary to add a log line at debug level? or this check will be pretty clear to debug if happens ?

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5378/commits/275b5340bf1a7719dc878f2ae0ff2eb8680b328d

Flaky tests are flaky...

PR #5381, opened 2022-07-06
notify: `channel_state_changed` now receives notice when channel opens

niftynei:
Previously we wouldn't notify when a channel moves into state
"CHANNELD_AWAITING_LOCKIN", as this is the original state (so there's
no movement btw states). This meant that it's impossible to track when a
channel's commitment txs have been exchanged and we're waiting for
onchain confirmation.

It's useful to have notice of this initialization though, all in one
place so that the `channel_state_changed` notification can successfully
track the entire lifecycle of a channel, from inception to close.

Note that for v2 "dual-funded" channels, we already notify at the same place, at
"DUALOPEND_AWAITING_LOCKIN" (the initial state for a dualopend channel
is "DUALOPEND_OPEN_INIT" -- this is the only state we don't get notified
at now...)

Changelog-Added: Plugins: `channel_state_changed` now triggers for a v1 channel's initial "CHANNELD_AWAITING_LOCKIN" state transition (from prior state "unknown")

### DIFF
--- a/lightningd/opening_control.c
+++ b/lightningd/opening_control.c
@@ -101,6 +101,7 @@ wallet_commit_channel(struct lightningd *ld,
 	u64 static_remotekey_start;
 	u32 lease_start_blockheight = 0; /* No leases on v1 */
 	struct short_channel_id *alias_local;
+	struct timeabs timestamp;
 
 	/* We cannot both be the fundee *and* have a `fundchannel_start`
 	 * command running!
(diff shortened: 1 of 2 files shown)


cdecker:
Rebased on top of `master` to get the broken CI job off the board.

ACK https://github.com/ElementsProject/lightning/pull/5381/commits/f131714c619c37949c21f661ae79a60ff826480e

vincenzopalazzo (approved):
ACK [f131714](https://github.com/ElementsProject/lightning/pull/5381/commits/f131714c619c37949c21f661ae79a60ff826480e)[f131714](https://github.com/ElementsProject/lightning/pull/5381/commits/f131714c619c37949c21f661ae79a60ff826480e)

niftynei:
re-applying ACK https://github.com/ElementsProject/lightning/pull/5381/commits/ae80d5ec6231f75f6cdbf5b2a0b71b37492acf4a

PR #5385, opened 2022-07-07
common: add missing .c file to src list

niftynei:
`make check-common-files` fails on master; it's missing the common/scb_wiregen.c file in `common/Makefile`

Changelog-None.


niftynei:
jk just a me problem...

Issue #5393, opened 2022-07-09
CMP Counter Asserted


nepet:
Scenario: We resolved an invoice and the moment after we got the following error (might not be related to the invoice, or is it? ...).
CLN is version v0.10.2 well connected and running for some time now.
```
...
2022-07-08T15:23:48.314Z INFO    lightningd: Resolved invoice 
...
Assertion failed: a->c->rr_number < b->c->rr_number (lightningd/invoice.c: cmp_rr_number: 623)
lightningd: FATAL SIGNAL 6 (version v0.10.2-modded)
0x5654f1c40061 send_backtrace
        common/daemon.c:33
0x5654f1c400e9 crashdump
        common/daemon.c:46
0x7efd87da6c8a ???
        ???:0
2022-07-08T15:23:48.348Z **BROKEN** lightningd: FATAL SIGNAL 6 (version v0.10.2-modded)
2022-07-08T15:23:48.349Z **BROKEN** lightningd: backtrace: common/daemon.c:38 (send_backtrace) 0x5654f1c400a9
2022-07-08T15:23:48.349Z **BROKEN** lightningd: backtrace: common/daemon.c:46 (crashdump) 0x5654f1c400e9
2022-07-08T15:23:48.349Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7efd87da6c8a
```

rustyrussell:
Annoying that we don't see the full backtrace (or better, a core file!).

This code is called for invoice *generation*, so that's where it crashed.

There are several possible causes for this:
1. We have two channels with the same rr_number. A quick audit shows we always set that rr_number to a unique value (and 64 bits, so wrap is not possible between the release and now!).
2. It's theoretically possible that sort() could compare a value with itself, but that would be really dumb: it doesn't that I've ever seen, but then, we've never seen this assert() hit, either.
3. listincoming has given us the same channel twice.  I don't see how that is possible: we had a race where channels could momentarily vanish, but never be duplicated (emailed crash.log shows no duplicates!).
4. General corruption/freed memory access.  If a channel we've just looked up is gone but still in the hash table, this is possible but would cause lots of random behavior and crashes.

PR #5396, opened 2022-07-10
invoice: turn assert failure into more informative log_broken message.

rustyrussell:
We had this assertion fail, and I can't see a clear reason why.  Remove it
(i.e. don't crash because we're having trouble with creating invoice routehints)
and add logging.

```
Assertion failed: a->c->rr_number < b->c->rr_number (lightningd/invoice.c: cmp_rr_number: 623)
lightningd: FATAL SIGNAL 6 (version v0.10.2-modded)
0x5654f1c40061 send_backtrace
        common/daemon.c:33
0x5654f1c400e9 crashdump
        common/daemon.c:46
0x7efd87da6c8a ???
        ???:0
```

There are several possible causes for this:

1. We have two channels with the same rr_number. A quick audit shows we always set that rr_number to a unique value (and 64 bits, so wrap is not possible between the release and now!).
2. It's theoretically possible that sort() could compare a value with itself, but that would be really dumb: it doesn't that I've ever seen, but then, we've never seen this assert() hit, either.
3. listincoming has given us the same channel twice. I don't see how that is possible: we had a race where channels could momentarily vanish, but never be duplicated (emailed crash.log shows no duplicates!).
4. General corruption/freed memory access. If a channel we've just looked up is gone but still in the hash table, this is possible but would cause lots of random behavior and crashes.

Fixes: #5393

### DIFF
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -617,13 +617,17 @@ static struct route_info **select_inchan(const tal_t *ctx,
 
 static int cmp_rr_number(const struct routehint_candidate *a,
 			 const struct routehint_candidate *b,
-			 void *unused)
+			 struct lightningd *ld)
 {
-	/* They're unique, so can't be equal */
 	if (a->c->rr_number > b->c->rr_number)
 		return 1;
-	assert(a->c->rr_number < b->c->rr_number);
-	return -1;
+	if (a->c->rr_number < b->c->rr_number)
+		return -1;
+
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5396/commits/dd70e9af26f1e0cc8af38c853d4d758010fa4c4e

PR #5397, opened 2022-07-10
cln_plugin: Configure "dynamic" field in "getmanifest" message

justinmoon:
Without this I can't use `lightning-cli plugin stop` etc.


cdecker (approved):
ACK c6a6eb4f06b3b262f490ef6e88c32e4e8cdcdf17

PR #5399, opened 2022-07-11
tests: Fix the canned gossmap

cdecker:
The gossip_store version was bumbed to 10 in PR #5239 and #5375 has
apparently not been rebased on top of that when it was merged, causing
the `run-gossmap_canned` test to fail when parsing the v9 gossip
store.

Changelog-None


endothermicdev:
ACK 83a8343a

Issue #5395, opened 2022-07-10
Crash on gossipd?


Coinomatron:
Upgraded to the latest repository build, overnight my node crashed, here's the first lines of the log output and crash error. 

-------------------------

2022-07-09T22:48:13.322Z DEBUG   0318ac9faa9629e7da08819bc8fe0dd2ae3044d69b1b2283a63479acffeb968483-gossipd: Received channel_update for channel 7417>
2022-07-09T22:48:14.002Z DEBUG   023e09c43b215bd3dbf483bcb409da3322ea5ea3b046f74698b89ee9ea785dd30a-gossipd: Received channel_update for channel 7047>
lightning_gossipd: gossipd/gossip_store.c:644: delete_by_index: Assertion `(be32_to_cpu(belen) & GOSSIP_STORE_LEN_DELETED_BIT) == 0' failed.
lightning_gossipd: FATAL SIGNAL 6 (version v0.11.0.1-207-g9d1b5a6)
0x560d25d22e03 send_backtrace
        common/daemon.c:33
0x560d25d22e9d crashdump
        common/daemon.c:46
0x7fd47eceb08f ??? 
        ???:0
0x7fd47eceb00b ???
        ???:0
0x7fd47ecca858 ???
        ???:0
0x7fd47ecca728 ???
        ???:0
0x7fd47ecdbfd5 ???
        ???:0
0x560d25d166df delete_by_index
        gossipd/gossip_store.c:644
0x560d25d171f8 gossip_store_delete
        gossipd/gossip_store.c:665
0x560d25d1cdc8 routing_add_node_announcement
        gossipd/routing.c:1730
0x560d25d1d113 handle_node_announcement
        gossipd/routing.c:1851
0x560d25d14311 handle_node_announce
        gossipd/gossipd.c:281
0x560d25d15a1a handle_recv_gossip
        gossipd/gossipd.c:507
0x560d25d15b0b connectd_req
        gossipd/gossipd.c:583
0x560d25d22f5e handle_read
        common/daemon_conn.c:31
0x560d25d52309 next_plan
        ccan/ccan/io/io.c:59
0x560d25d527d6 do_plan
        ccan/ccan/io/io.c:407
0x560d25d52873 io_ready
        ccan/ccan/io/io.c:417
0x560d25d54208 io_loop
        ccan/ccan/io/poll.c:453
0x560d25d15d0c main
        gossipd/gossipd.c:1130
0x7fd47eccc082 ???
        ???:0
[...]

Issue #5387, opened 2022-07-08
gossipd: STATUS_FAIL_INTERNAL_ERROR: FATAL SIGNAL


grubles:
### Issue and Steps to Reproduce

Running CLN at 0a4cd9102852a3eb5f105f7787e5b8168fb56335
OS: Ubuntu 22.04
Arch: powerpc64le

Logs:
```
2022-07-08T16:53:26.606Z DEBUG   036ff8f2a910f531295d695211a0600ee6d527ea58359f81ceadf44b658c7c476e-gossipd: Received node_announcement for node 03464119063dc712f1629d3f62035f438ddde2c77828afb0647948ac8
2022-07-08T16:53:26.619Z **BROKEN** gossipd: FATAL SIGNAL 6 (version v0.11.0.1-190-g2927c8c)                
2022-07-08T16:53:26.619Z **BROKEN** gossipd: backtrace: common/daemon.c:38 (send_backtrace) 0xb8fe1f2f38b   
2022-07-08T16:53:26.619Z **BROKEN** gossipd: backtrace: common/daemon.c:46 (crashdump) 0xb8fe1f2f413        
2022-07-08T16:53:26.619Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf386db0443                    
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf3869c867c                                                                                                                           
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf386961e3b            
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf38693d25f
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf386954f13
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: (null):0 ((null)) 0x7bf386954fb3
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: gossipd/gossip_store.c:644 (delete_by_index) 0xb8fe1f1d813
2022-07-08T16:53:26.620Z **BROKEN** gossipd: backtrace: gossipd/gossip_store.c:665 (gossip_store_delete) 0xb8fe1f1e69b
[...]

PR #5400, opened 2022-07-11
gossipd: fix crash from gossip_store v10 changes

endothermicdev:
There were a couple cleanups needed for commit 87a66c1.  Removed channels and nodes were leaving rate-limit flagged gossip behind in the store which caused issues on subsequent loads.  Also updates the check for duplication on gossip_store load which allowed only a single message per channel/node to now allow an additional rate-limited message as well.

Addresses issues #5387, #5395.

Changelog-fixed: gossipd: crash on gossip_store write


cdecker:
Thanks @endothermicdev for the fixes. Two small nits I have:

 - We generally don't add changelog fixed entries for bugs that weren't in a release
 - there are a couple of spurious whitespace changes that aren't around other changes. No biggy but they tend to pollute the diffs 

Otherwise excellent 👍

endothermicdev:
Thanks Christian.  Will clean up both of those.

rustyrussell (approved):
Ack ab94d16cec602be942e74edf637fccba5f4780d6

Normally I'd prefer these as separate commits, but I'm not sure it's actually sanely divisible.

Also note: github responds magically to the syntax:

Fixes: #5387
Fixes: #5395

PR #5356, opened 2022-06-29
ci: introduce the deff check for rust

vincenzopalazzo:
This is a monkey patch, just to be sure that we did not release inconsistent code, but we can evaluate to remove the generated file? or make rust a requirements (we all love rust 🦀  )

Updating also some rust model

Rebased on top of https://github.com/ElementsProject/lightning/pull/5364

Signed-off-by: Vincenzo Palazzo <vincenzopalazzodev@gmail.com>

PR #5415, opened 2022-07-12
make: Add msggen generated files to check-gen-updated target

cdecker:
This is a coutnerproposal to #5356 which uses the existing CI jobs and targets to check we committed all changes.


vincenzopalazzo (approved):
ACK [024bfa1](https://github.com/ElementsProject/lightning/pull/5415/commits/024bfa1ff51207fd3d0faace2d2375edee3d9383) but I admit that I'm not a big fun of mixing required stuff with not required one.

rustyrussell:
Love the idea... it was supposed to fail on master, right?

Weird that there's a python diff in there though... hmm....?

cdecker:
Yep, I'll retire libhsmd_python as noone is using it, and add the latest version of the files before merging.

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5415/commits/711f33901bb9ba090de6bd7451757558cac2821b

@adi2011 noticed that the all but 2 of the `doc/*.7` files aren't checked into git, therefore `check-gen-updated` doesn't fail for them.

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5415/commits/711f33901bb9ba090de6bd7451757558cac2821b

>noticed that the all but 2 of the doc/*.7 files aren't checked into git, therefore check-gen-updated doesn't fail for them.

imho this need to be done inside the docs CI because otherwise there is no possibility to run the CI is we are working in a rpc command update without changing the docs.

Changing rpc command it is not too frequent yes this is true!

rustyrussell:
We should not be checking in any generated files: should git rm those two as well!

PR #5423, opened 2022-07-13
plugin: autoclean fix crash + pyln-testing: print stderr from errlog

SimonVrouwe:
Plugin `autoclean` crashes with below backtrace after disable and re-enable.
Strangely I mostly reproduces with the sequence disable-enable-disable-sleep-enable, see added test.

Also modifies `pyln-testing` to add stderr `errlog` (with backtrace etc.) to CI report after unexpected test failure.
 
I have the fix, but I am just curious if below backtrace shows up in the test results.
```
autoclean: FATAL SIGNAL 6 (version v0.11.0.1-240-gbd239d2)
0x55a26e53adf4 send_backtrace
        common/daemon.c:33
0x55a26e53ae9a crashdump
        common/daemon.c:46
0x7fed8b8d783f ???
        /build/glibc-fWwxX8/glibc-2.28/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
0x7fed8b8d77bb __GI_raise
        ../sysdeps/unix/sysv/linux/raise.c:51
0x7fed8b8c2534 __GI_abort
        /build/glibc-fWwxX8/glibc-2.28/stdlib/abort.c:79
0x55a26e57a98c call_error
        ccan/ccan/tal/tal.c:93
0x55a26e57ab54 check_bounds
        ccan/ccan/tal/tal.c:165
0x55a26e57abb2 to_tal_hdr
        ccan/ccan/tal/tal.c:176
0x55a26e57b53c tal_free
        ccan/ccan/tal/tal.c:479
0x55a26e5213b8 json_autocleaninvoice
        plugins/autoclean.c:55
0x55a26e5255d2 ld_command_handle
        plugins/libplugin.c:1307
0x55a26e525723 ld_read_json_one
        plugins/libplugin.c:1348
0x55a26e525855 ld_read_json
        plugins/libplugin.c:1368
0x55a26e5695d6 next_plan
        ccan/ccan/io/io.c:59
0x55a26e56a176 do_plan
        ccan/ccan/io/io.c:407
0x55a26e56a1b4 io_ready
        ccan/ccan/io/io.c:417
0x55a26e56c3d1 io_loop
        ccan/ccan/io/poll.c:453
0x55a26e5262e9 plugin_main
        plugins/libplugin.c:1565
0x55a26e5215de main
        plugins/autoclean.c:92
0x7fed8b8c409a __libc_start_main
        ../csu/libc-start.c:308
0x55a26e520de9 ???
        ???:0
0xffffffffffffffff ???
        ???:0
autoclean: FATAL SIGNAL 11 (version v0.11.0.1-240-gbd239d2)
0x55a26e53adf4 send_backtrace
        common/daemon.c:33
0x55a26e53ae9a crashdump
        common/daemon.c:46
0x7fed8b8d783f ???
        /build/glibc-fWwxX8/glibc-2.28/signal/../sysdeps/unix/sysv/linux/x86_64/sigaction.c:0
0x0 ???
        ???:0

```

### DIFF
--- a/plugins/autoclean.c
+++ b/plugins/autoclean.c
@@ -47,12 +47,14 @@ static struct command_result *json_autocleaninvoice(struct command *cmd,
 	cycle_seconds = *cycle;
 	expired_by = *exby;
 
+	cleantimer = tal_free(cleantimer);
+
 	if (cycle_seconds == 0) {
 		response = jsonrpc_stream_success(cmd);
 		json_add_bool(response, "enabled", false);
 		return command_finished(cmd, response);
 	}
-	tal_free(cleantimer);
+
 	cleantimer = plugin_timer(cmd->plugin, time_from_sec(cycle_seconds),
 				  do_clean, cmd->plugin);
 
--- a/tests/test_invoices.py
+++ b/tests/test_invoices.py
@@ -559,7 +559,6 @@ def test_waitanyinvoice_reversed(node_factory, executor):
     assert r['label'] == 'inv1'
 
 
-@pytest.mark.xfail(strict=True)
 def test_autocleaninvoice(node_factory):
     l1 = node_factory.get_node()
 
--- a/tests/test_invoices.py
+++ b/tests/test_invoices.py
@@ -559,6 +559,7 @@ def test_waitanyinvoice_reversed(node_factory, executor):
     assert r['label'] == 'inv1'
 
 
+@pytest.mark.xfail(strict=True)
 def test_autocleaninvoice(node_factory):
     l1 = node_factory.get_node()
 
@@ -598,6 +599,13 @@ def test_autocleaninvoice(node_factory):
     # Everything deleted
     assert len(l1.rpc.listinvoices('inv1')['invoices']) == 0
     assert len(l1.rpc.listinvoices('inv2')['invoices']) == 0
+
+    # stress test
+    l1.rpc.autocleaninvoice(cycle_seconds=0)
+    l1.rpc.autocleaninvoice(cycle_seconds=1)
+    l1.rpc.autocleaninvoice(cycle_seconds=0)
+    time.sleep(1)
+    l1.rpc.autocleaninvoice(cycle_seconds=1, expired_by=1)
 
 
 def test_decode_unknown(node_factory):
--- a/contrib/pyln-testing/pyln/testing/fixtures.py
+++ b/contrib/pyln-testing/pyln/testing/fixtures.py
@@ -495,6 +495,28 @@ def node_factory(request, directory, test_name, bitcoind, executor, db_provider,
     map_node_error(nf.nodes, lambda n: n.daemon.is_in_log(r'Accessing a null column'), "Accessing a null column")
     map_node_error(nf.nodes, checkMemleak, "had memleak messages")
     map_node_error(nf.nodes, lambda n: n.rc != 0 and not n.may_fail, "Node exited with return code {n.rc}")
+    if not ok:
+        map_node_error(nf.nodes, prinErrlog, "some node failed unexpected, non-empty errlog file")
+
+
+def getErrlog(node):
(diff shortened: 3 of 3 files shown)


niftynei:
Try `check-python-flake8` and fixing up the whitespace errors, then repushing.

SimonVrouwe:
Yep thanks, the good news is that the correct `test_autocleaninvoice` fails. 
Unfortunately the `FATAL SIGNAL 6` does not show up in the test result and I am pretty sure it was written to stderr.

Perhaps this is caused by recent changes to the pyln-testing in #5281 ?
Since that PR, `stderr` is saved in a file named "errlog"
https://github.com/ElementsProject/lightning/blob/27e5ddc7b766cc8dad607d01c9c1de1fb22ebf45/contrib/pyln-testing/pyln/testing/utils.py#L211-L220

But it seems nothing is done with it unless `logs_catchup` is called (which writes it out to `sys.stderr`) and only if `verbose=True`
https://github.com/ElementsProject/lightning/blob/27e5ddc7b766cc8dad607d01c9c1de1fb22ebf45/contrib/pyln-testing/pyln/testing/utils.py#L247-L250
which is not even the case for `lightningd`
https://github.com/ElementsProject/lightning/blob/27e5ddc7b766cc8dad607d01c9c1de1fb22ebf45/contrib/pyln-testing/pyln/testing/utils.py#L540-L543

So I am not sure what the intended use here is.
Shouldn't the CI test report at least print the content of "errlog" files?

rustyrussell:
> Why not use the python logging inside the lib, and enable the log verbosity with `--log-cli-level={LEVEL_NAME}` in pytest? looks like that write in the stdio make the life harder?

We could, and probably should.  But it still won't fix this: we want stderr to go somewhere.

We used to have python threads handling this, but it seemed to introduce stability issues in our tests (hsmtool tests would fail randomly, for example), and flakes...

SimonVrouwe:
Happy to  drop or modify the pyln-testing [8b0d0a7](https://github.com/ElementsProject/lightning/pull/5423/commits/8b0d0a7bb2860bfb588faff1b106c6e86108c258) commit if needed.

tests/test_invoices.py
```diff
@@ -599,6 +599,13 @@ def test_autocleaninvoice(node_factory):
     assert len(l1.rpc.listinvoices('inv1')['invoices']) == 0
     assert len(l1.rpc.listinvoices('inv2')['invoices']) == 0
 
+    # stress test
+    l1.rpc.autocleaninvoice(cycle_seconds=0)
+    l1.rpc.autocleaninvoice(cycle_seconds=1)
+    l1.rpc.autocleaninvoice(cycle_seconds=0)
```
rustyrussell, line 605: You should put `@pytest.mark.xfail(strict=True)` at the top of this test, and remove it in the final commit.  That avoids breaking bisect...

plugins/autoclean.c
```diff
@@ -47,12 +47,15 @@ static struct command_result *json_autocleaninvoice(struct command *cmd,
 	cycle_seconds = *cycle;
 	expired_by = *exby;
 
+	if (cleantimer)
+		cleantimer = tal_free(cleantimer);
+
```
rustyrussell, line 52: Skip the if?  `cleantimer = tal_free(cleantimer);` is sufficient?

rustyrussell (changes requested):
Ack afd5b27b60d0f3fe1bcf0243b00f4ca5427ea81b

Minor changes only, so I've acked this if you disagree.

plugins/autoclean.c
SimonVrouwe: Yes thanks, `tal_free(NULL)` is always ok.

SimonVrouwe:
Applied all suggestions, force pushed.

rustyrussell (commented):
Ack 32cf9daf68ca4a3fc96d322362ffce610aece96c

rustyrussell (approved):

PR #5425, opened 2022-07-14
lightningd: test that hsm_secret is as expected, at startup.

rustyrussell:
If you get the wrong hsm_secret, your node_id will change, and
peers won't know who you are, bitcoind will reject your transaction
signatures, and other madness.

Catch this as soon as it happens, by storing our node_id in the db.

Suggested-by: @cdecker


lightningd/lightningd.c
```diff
@@ -1048,9 +1048,12 @@ int main(int argc, char *argv[])
...
 	 * the network.  Instead, the db knows what chain it belongs to, and we
-	 * simple barf here if it's wrong. */
-	if (!wallet_network_check(ld->wallet))
-		errx(1, "Wallet network check failed.");
+	 * simple barf here if it's wrong.
+	 *
+	 * We also check that our node_id is what we expect: otherwise a change
+	 * in hsm_secret will have strange consequences! */
+	if (!wallet_sanity_check(ld->wallet))
+		errx(1, "Wallet sanity check failed.");
```
vincenzopalazzo, line 1056:
```suggestion
		errx(WALLET_MISMATCH, "Wallet sanity check failed.");
```

What do you think to add a more accurate error here? this use case looks like a use case that can happens during some scripting stuff (raspiblitz stuff are usually happy with specific error on the cmd)

But maybe there are other reason to return a generic one?

rustyrussell: Actually, this is a cleanup which deserves its own (not 0.12) branch....

niftynei:
*and Suggested-By: @fiatjaf!

rustyrussell:
Trivial rebase on master, added Changelog-Changed line.

rustyrussell:
> *and Suggested-By: @fiatjaf!

Didn't know/remember that, but I've credited him too!

rustyrussell:
Fixed hsm generation test, which barfed because we change secret...

PR #5428, opened 2022-07-14
docs: replace bolt12 with bolt11 in lightning-fetchinvoice command

gautamjajoo:
I am working with @vincenzopalazzo on the project [ln-dashboard](https://github.com/clightning4j/ln-dashboard).
While working on it, I was referring to the docs, and it was super confusing for me while looking into the `fetchinvoice` command.

Then I encountered that this command gives `bolt11` as output, and in the docs, it was written `bolt12`.

### DIFF
--- a/doc/lightning-fetchinvoice.7.md
+++ b/doc/lightning-fetchinvoice.7.md
@@ -50,7 +50,7 @@ RETURN VALUE
 
 [comment]: # (GENERATE-FROM-SCHEMA-START)
 On success, an object is returned, containing:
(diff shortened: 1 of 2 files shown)


vincenzopalazzo (commented):
Sorry if there is confusion here, this should my fault that I leave alone in this invoice world!

rustyrussell:
Wait, what?  fetchinvoice DEFINITELY should return a bolt12!  What are you seeing?

gautamjajoo:
@rustyrussell Sorry there was some confusion, will close the PR!

Issue #5435, opened 2022-07-17
"Peer says it sees our address as:" irrelevant if you aren't using auto-discovery 


wtogami:
```
INFO    lightningd: Peer says it sees our address as: 66.254.114.41:30165
```

- There are several configurations in which you don't want to hear about this because you expect auto-discovery to be turned off.
    - `disable_ip_discovery` definitely
    - `announce-addr=` where you explicitly announce public address of the same type strongly implies that you intended `disable_ip_discovery`?
    -  `always_use_proxy` too? I'm not sure.
- `log-level=broken:<nodeid>` type filtering doesn't work because the peer nodeid isn't associated with the log event.

What could be done?

- Fix it so nodeid filtering works.
- Add conditional scenarios where if auto-discovery is turned off then don't log this?
- Change it to `log_debug()`? Would conditionally be confusing?

Note: Any changes of this type might need the tests that rely on this log event need to be changed.

wtogami:
Closing as #5436 fixed most of annoyance.

PR #5436, opened 2022-07-17
lightningd: reduce log level for remote address reporting.

rustyrussell:
It's available in listpeers() if you want to see it, otherwise it's not
really something users want to see in the normal course of operation.

Changelog-None

### DIFF
--- a/lightningd/peer_control.c
+++ b/lightningd/peer_control.c
@@ -1202,8 +1202,8 @@ void peer_connected(struct lightningd *ld, const u8 *msg)
 
 	/* Log and update remote_addr for Nat/IP discovery. */
 	if (hook_payload->remote_addr) {
-		log_peer_info(ld->log, &id, "Peer says it sees our address as: %s",
-			      fmt_wireaddr(tmpctx, hook_payload->remote_addr));
+		log_peer_debug(ld->log, &id, "Peer says it sees our address as: %s",
+			       fmt_wireaddr(tmpctx, hook_payload->remote_addr));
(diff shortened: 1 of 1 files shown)


lightningd/peer_control.c
```diff
@@ -1202,8 +1202,8 @@ void peer_connected(struct lightningd *ld, const u8 *msg)
 
 	/* Log and update remote_addr for Nat/IP discovery. */
 	if (hook_payload->remote_addr) {
-		log_peer_info(ld->log, &id, "Peer says it sees our address as: %s",
-			      fmt_wireaddr(tmpctx, hook_payload->remote_addr));
+		log_peer_debug(ld->log, &id, "Peer says it sees our address as: %s",
+			       fmt_wireaddr(tmpctx, hook_payload->remote_addr));
```
wtogami, line 1206: extraneous space?

otherwise ACK a5b9830796b4fe94feb55d2a05c51e5fd156dd32

I had been using exactly this on v0.11.2 because #5435 was annoying me. I didn't submit this because I thought people would think this was too blunt an instrument but I'm happy if people are OK with this.

wtogami (approved):

lightningd/peer_control.c
m-schmoock: The space is actually 'correct' :)

m-schmoock (approved):
I thought it was already DEBUG. Thanks

m-schmoock:
ACK a5b9830

PR #5441, opened 2022-07-20
hsmtool: implement checkhsm.

rustyrussell:
This gives a nice way to ensure your secret is the correct one.

Also, we don't need to suppress VALGRIND for this test, now the output
races are fixed.

Changelog-Added: `hsmtool`: new command `checkhsm` to check BIP39 passphrase against hsm_secret.


cdecker:
ACK 27d1479242ce926bd016aeb3c6738106f070e55e

PR #5444, opened 2022-07-20
doc: Fix `wrong_funding` description in manpage and type in schema

kristapsk:
<del>It's already `wrong_funding_txid` in DESCRIPTION, but not SYNOPSIS.</del>

<ins>It's currently `wrong_funding` in SYNOPSIS, but `wrong_funding_txid` in DESCRIPTION.</ins>


doc/lightning-close.7.md
```diff
@@ -4,7 +4,7 @@ lightning-close -- Command for closing channels with direct peers
 SYNOPSIS
 --------
 
-**close** *id* [*unilateraltimeout*] [*destination*] [*fee_negotiation_step*] [*wrong_funding*] [*force_lease_closed*] [\*feerange\*]
+**close** *id* [*unilateraltimeout*] [*destination*] [*fee_negotiation_step*] [*wrong_funding_txid*] [*force_lease_closed*] [\*feerange\*]
```
vincenzopalazzo, line 7: nit: this break the code https://github.com/ElementsProject/lightning/blob/e70729b04befe44d603370b96fbb10dc00bd342d/lightningd/closing_control.c#L627

Maybe we have the json schema wrong?

vincenzopalazzo (commented):
nit: this

doc/lightning-close.7.md
cdecker: The correct way of doing this is accept both for a deprecation period, gating the old one on `deprecated_apis == true`, so it won't show if we run without it, and after 6 months ~= 3 releases we can remove the old one and any merge logic on the two.

kristapsk: @cdecker So you agree this way is better and I just need to add deprecation code?

vincenzopalazzo: I want to clarify my review here, the nit is regarding the docs changes because what you are changing is not true in the code.

However, the suffix `txid` will make more clear the command line options type, and this is good.

cdecker: I do agree that the `txid` suffix makes it much clearer. It's just that renaming things in the public interface sadly comes with the deprecation window requirements, which is why we usually refrain from addressing these types of issues.

Would absolutely accept if this had the necessary deprecation window added.

kristapsk: Hmm... but it's actually outpoint, not txid. Maybe proper name would be `wrong_funding_outpoint`?

cdecker: Also true, man, naming is hard ^^

vincenzopalazzo: ok, why not left as is :) and adding a `FIXME: adding a new cool name` should fit in the requirements of when use a `FIXME`

kristapsk: > ok, why not left as is :) and adding a `FIXME: adding a new cool name` should fit in the requirements of when use a `FIXME`

Renaming and adding deprecation window is not so hard, don't see point in adding some FIXME comment here.

vincenzopalazzo: >Renaming and adding deprecation window is not so hard, don't see point in adding some FIXME comment here.

From ho provide the API yes it is easy, from ho use the API is a pain any way. The important think is that we keep the convention by naming along all the API.

My vote for the naming is the same, summurized:

- `wrong_funding_txid`: could work
- `wrong_funding_outpoint`: good luck if you need to type this string on the keyboard, long name is painful if you think accessibility for people that have some disabilities.
- `wrong_funding`: generic name that include both of previous,  so if there is no agreement I preferer stay with the generic one

kristapsk: Probabaly you're right. Then alternative is to rename `wrong_funding_txid` to `wrong_funding` in DESCRIPTION in manpage and also fix type in `doc/schemas/close.request.json`, which is also wrong (should be `outpoint` not `txid`).

vincenzopalazzo: I think the docs fix is a very improvement, and also if you think that the name is not so clear (and it is not clear IMHO) you could al a small explanation in the docs

kristapsk: It's explained good enough in DESCRIPTION part of manpage. Ok, will just rename it properly there and fix type in schema. That should not be controversial. And if decision is made to rename it, can be done later.

vincenzopalazzo:
I think you should run `make` with rust enable, and I do not see any hash change in the docs so maybe can be good to run also `make doc-all`

kristapsk:
> I think you should run `make` with rust enable, and I do not see any hash change in the docs so maybe can be good to run also `make doc-all`

Done

vincenzopalazzo:
```
(cd plugins; cargo build --quiet --examples)
error[E0308]: mismatched types
    --> cln-grpc/src/convert.rs:1080:28
     |
1080 |             wrong_funding: c.wrong_funding.clone(), // Rule #1 for type outpoint?
     |                            ^^^^^^^^^^^^^^^^^^^^^^^ expected struct `cln_rpc::primitives::Outpoint`, found struct `pb::Outpoint`
     |
     = note: expected enum `Option<cln_rpc::primitives::Outpoint>`
                found enum `Option<pb::Outpoint>`

```

I think you have no rust on your system? I have no permission to push so I can't help you to compile it :/ sorry

kristapsk:
I have, did not have errors locally after I rebuilt it with rustfmt enabled. 1.61.0 on Gentoo Linux.

doc/schemas/close.request.json
```diff
@@ -23,7 +23,7 @@
       "description": ""
     },
     "wrong_funding": {
-      "type": "txid",
+      "type": "outpoint",
```
vincenzopalazzo, line 26:
```suggestion
      "type": "txid",
```

I think this is the compiler error problem, sorry but I did not pay too much attention to the error message :/

There is any reason to change into `output` that I'm missing?

P.S: Code generator are lovely but too magic :)

vincenzopalazzo (changes requested):

doc/schemas/close.request.json
kristapsk: Yes, it's actually outpoint, not txid, see manpage.

vincenzopalazzo: you have right! so at this point I think @cdecker is the best person to make rust happy here

niftynei:
Failing with an error on type mismatch in the rust codes.

```
    --> cln-grpc/src/convert.rs:1080:28
     |
1080 |             wrong_funding: c.wrong_funding.clone(), // Rule #1 for type outpoint?
     |                            ^^^^^^^^^^^^^^^^^^^^^^^ expected struct `cln_rpc::primitives::Outpoint`, found struct `pb::Outpoint`
     |
     = note: expected enum `Option<cln_rpc::primitives::Outpoint>`
                found enum `Option<pb::Outpoint>`
```

doc/schemas/close.request.json
cdecker: Yeah, sadly this is also a breaking change for the rust bindings. Though we don't yet have a compatibility guarantee, so we can break on a whim imho.

cdecker:
Added a commit https://github.com/ElementsProject/lightning/pull/5444/commits/5f6b42ce0cc85c5703d67b9453135a7a14db730a to map the `Option<Outpoint>` in `cln-rpc` and `cln-grpc` conversions. This is the first time we had that type as an argument, hence the lack of conversion so far.

kristapsk:
Have this been just forgotten about?

cdecker:
Nope, adding it to my list of PRs to shepherd through CI

PR #5445, opened 2022-07-20
Remove mrkd (use lowdown, build if not found locally)

rustyrussell:
As requested by @wtogami ...

Changelog-None

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -66,6 +66,8 @@ For development or running tests, get additional dependencies:
     sudo apt-get install -y valgrind libpq-dev shellcheck cppcheck \
       libsecp256k1-dev jq lowdown
 
+If you can't install `lowdown`, a version will be built in-tree.
+
 If you want to build the Rust plugins (currently, cln-grpc):
 
 	sudo apt-get install -y cargo rustfmt
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -64,7 +64,7 @@ Checkout a release tag:
 For development or running tests, get additional dependencies:
 
     sudo apt-get install -y valgrind libpq-dev shellcheck cppcheck \
-      libsecp256k1-dev jq
+      libsecp256k1-dev jq lowdown
 
 If you want to build the Rust plugins (currently, cln-grpc):
 
@@ -174,12 +174,6 @@ fiddle with compile time options:
 
     # cd /usr/ports/net-p2p/c-lightning && make install
 
-mrkd is required to build man pages from markdown files (not done by the port):
-
-    # cd /usr/ports/devel/py-pip && make install
-    $ pip install --user poetry
-    $ poetry install
-
 See `/usr/ports/net-p2p/c-lightning/Makefile` for instructions on how to
 build from an arbitrary git commit, instead of the latest release tag.
 
@@ -209,7 +203,7 @@ pkg_add git python gmake py3-pip libtool gmp
 pkg_add automake # (select highest version, automake1.16.2 at time of writing)
 pkg_add autoconf # (select highest version, autoconf-2.69p2 at time of writing)
 ```
-Install `mako` and `mrkd` otherwise we run into build errors:
+Install `mako` otherwise we run into build errors:
 ```
 pip3.7 install --user poetry
 poetry install
--- a/.github/workflows/bsd.yml
+++ b/.github/workflows/bsd.yml
@@ -39,6 +39,7 @@ jobs:
             bash \
             gettext \
             sqlite3 \
+            lowdown \
             curl
 
           curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain nightly-2021-08-3z1
@@ -63,7 +64,6 @@ jobs:
             blinker \
             flake8 \
             mako \
-            mrkd \
             pytest-sentry \
             pytest-test-groups==1.0.3 \
             pytest-custom-exit-code==0.3.0 \
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,4 @@
 [submodule "external/lowdown"]
 	path = external/lowdown
 	url = https://github.com/kristapsdz/lowdown.git
+	ignore = dirty
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,6 @@
 [submodule "external/lnprototest"]
 	path = external/lnprototest
 	url = https://github.com/rustyrussell/lnprototest.git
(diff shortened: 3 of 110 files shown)


cdecker:
Just a word of caution: `lowdown` is only available in Ubuntu LTS starting from 22.04. So we might need to also add pointers to install it on older versions somehow.

wtogami:
> Just a word of caution: lowdown is only available in Ubuntu LTS starting from 22.04. So we might need to also add pointers to install it on older versions somehow.

It available on Fedora at all yet but I will newly package it and all other build deps for that distro to handle offline builds without pip or poetry.

vincenzopalazzo:
>It isn't available on Fedora at all yet but I will newly package it and all other build deps for that distro to handle offline builds without pip or poetry.

Mh! we really achieve the offline build by kicking off pip and poetry here? or are we just for now? if rust enters as required dependencies (it already is a required dep because you can not change an RPC command without it) the offline problem is open again right? or there is some way to use cargo in an offline mode that I'm not aware of?

I'm more than a pro to kick off `mrkd` because the maintainer is slow and unresponsive, and we can always put `lowdown` as a submodule because we already build C, so should be not a big deal! (and we can always keep a fork if some bug exists, I think we have enough people with good knowledge of C to keep a tiny markdown parser alive :) )

wtogami:
> > It isn't available on Fedora at all yet but I will newly package it and all other build deps for that distro to handle offline builds without pip or poetry.

> Mh! we really achieve the offline build by kicking off pip and poetry here? or are we just solve it for now? if rust enters as required dependencies (it already is a required dep because you can not change an RPC command without it) the offline problem is open again right? or there is some way to use cargo in an offline mode that I'm not aware of?

Distros had long packaged python modules to such a degree that many users never have used `pip`.

https://doc.rust-lang.org/cargo/commands/cargo-vendor.html
The usual solution for offline rust builds is with `cargo vendor`.

> I'm more than a pro to kick off `mrkd` because the maintainer is slow and unresponsive, and we can always put `lowdown` as a submodule because we already build C, so should be not a big deal! (and we can always keep a fork if some bug exists, I think we have enough people with good knowledge of C to keep a tiny markdown parser alive :) )

`lowdown` can easily be a tool packaged in distros. Building and running it as a build dependency is less desirable for reasons like it wouldn't work for cross-compiles.

wtogami:
Manually inspected `diff -u lightning-before/doc lightning-after/doc` and found some oddities.

`lightning-decodepay.7`

### Before
Screenshot from 2022-07-25 20-36-03

### After
Screenshot from 2022-07-25 20-36-14

I prefer the new bullet formatting as cleaner and more information dense. But it mangled `min_final_cltv_expiry` in such a way that man misinterprets some `_` literals as underline.

wtogami:
image

Here's the `diff -u` of particularly the `min_final_cltv_expiry` example from above.

wtogami:
These before and after shots show the typical difference from mrkd to lowdown.

### Before
Screenshot from 2022-07-25 21-02-08
### After
Screenshot from 2022-07-25 21-02-20

* What bothers me about many of the lowdown formatted outputs is the 1st level bullets are not indented.
* Various escaping changes make almost no difference except when it causes errors like in the above `decodepay` example. This page had no problems just the `SEE ALSO` is no longer bold.

wtogami:
`.gitignore`
```
# Ignore generated files
devtools/features
doc/lightning*.[1578]
```

I ran into a confusing issue when switching between before and after branches where dependency tracking failed to recognize it needed to re-process the man pages. I then tried `make clean` and didn't redo these files. Perhaps `make clean` could also additionally delete the same files ignored by `.gitignore`?

wtogami (dismissed):
The only real error I could find is underline format mangling within `lightning-decodepay.7`. The cosmetic differences are otherwise not a big deal. There could be other errors but I didn't look at every man file manually using `man`.

wtogami:
Here's the `diff -u` of the `doc/` directory generated before and after this PR.
https://fedorapeople.org/~wtogami/a/2022/e70729b04befe44d603370b96fbb10dc00bd342d...77f0388b6757bc6c7b9909521b0eadb0c55bd773-manpage.diff

Comparing before and after is tedious but we should also visually inspect the HTML formatted output as a sanity check so it won't be a surprise later when the docs website is regenerated.

prusnak:
For the record, clightning is now broken in nixpkgs/nix-bitcoin because mrkd is now broken in nixpkgs - https://github.com/fort-nix/nix-bitcoin/pull/534#issuecomment-1215815187

wtogami:
The above review found only cosmetic or minor formatting bugs. Build being *broken* is a bigger problem. For that reason maybe we should merge this before 0.12?

rustyrussell:
I don't think this should go in now, at it risks breaking every other platform that now needs to figure out how to get lowdown, update their docs etc.

wtogami:
How about pre-process the man pages, commit and skip that build step for the 0.12 release? That would be the least risky change would maximizing ease of deployment.

rustyrussell:
> image
> 
> Here's the `diff -u` of particularly the `min_final_cltv_expiry` example from above.

OK, I've fixed this now.

rustyrussell:
Rebased, and formatting fix.  The extra blank line before sub-indents is a bit jarring, but seems the best we can do for now?

(See man -l doc/lightning-decode.7)

rustyrussell:
> Just a word of caution: `lowdown` is only available in Ubuntu LTS starting from 22.04. So we might need to also add pointers to install it on older versions somehow.

Latest draft here will build one internally if it's not found by configure.  It seems fairly portable...

doc/INSTALL.md
```diff
@@ -209,7 +205,7 @@ pkg_add git python gmake py3-pip libtool gmp
 pkg_add automake # (select highest version, automake1.16.2 at time of writing)
 pkg_add autoconf # (select highest version, autoconf-2.69p2 at time of writing)
 ```
-Install `mako` and `mrkd` otherwise we run into build errors:
+Install `mako` otherwise we run into build errors:
```
vincenzopalazzo, line 208: for what we need `mako` in this case?

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5445/commits/8d471b40023c7b4ef6459c13c5331cdea7cd52a9

Maybe we can remove also `mako`? or we are using it for some somethings? maybe for the doc website?

doc/INSTALL.md
niftynei: `mako` is a templating language which is used to generate wire files; it's not related to `mrkd`, which is used to convert manpages from markdown into the roff file format which `man` requires.

PR #5451, opened 2022-07-22
DRAFT: Update libwally to begin PSBTv2 support

instagibbs:
It compiles.

The underlying PR in libwally is still moving, so this is mostly to prove out that what we need can be accomplished.

There's a few FIXMEs related to PSETs(which are now v2 only...) and a couple spots I didn't understand.

Eventually proper accessors will be available for some of these psbt.psbt_fields things, but for now I duplicate PSBT key values and use the mapping functionality directly.

Proper Taproot field accessors may also drop before merge of the parent libwally PR, unsure.


instagibbs:
Another update with more endpoints exposed for easier usage of libwally, and one bugfix.

FIXMEs still in place, waiting for some pro feedback for those

instagibbs:
`wally_psbt_input_set_asset` and `wally_psbt_input_set_value` were apparently removed, and there's nothing replacing them. I suggest we disable PSET support until this gets sorted and just focus on bitcoin support @niftynei

cdecker:
> `wally_psbt_input_set_asset` and `wally_psbt_input_set_value` were apparently removed, and there's nothing replacing them. I suggest we disable PSET support until this gets sorted and just focus on bitcoin support @niftynei

Won't that kill L-BTC support too?

instagibbs:
> Won't that kill L-BTC support too?

In other PR going to attempt to keep L-BTC support alive.

PR #5454, opened 2022-07-23
test timing fixes/stderr redirect 

niftynei:
I was getting consistent test failures for these when looking at the `bookkeeper` test fails; mostly these were timing bugs!

### DIFF
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -1132,7 +1132,7 @@ def test_funding_reorg_private(node_factory, bitcoind):
     l2.daemon.wait_for_log(r'Deleting channel')
 
 
-@pytest.mark.developer("needs DEVELOPER=1")
+@pytest.mark.developer("needs DEVELOPER=1", "uses --dev-fast-reconnect")
 @pytest.mark.openchannel('v1')
 @pytest.mark.openchannel('v2')
 def test_funding_reorg_remote_lags(node_factory, bitcoind):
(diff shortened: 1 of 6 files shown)


tests/test_misc.py
```diff
@@ -1727,7 +1727,7 @@ def mock_fail(*args):
     l1.daemon.rpcproxy.mock_rpc('getblockhash', mock_fail)
     l1.daemon.rpcproxy.mock_rpc('estimatesmartfee', mock_fail)
 
-    l1.daemon.start(wait_for_initialized=False)
+    l1.daemon.start(wait_for_initialized=False, stderr_redir=True)
```
rustyrussell, line 1732: I don't think this should make a difference?

Unless these errors are going to stderr, in which case wait_for_logs is not the right thing 

This test flakes for me locally all the time too..

tests/test_misc.py
```diff
@@ -1132,15 +1132,15 @@ def test_funding_reorg_private(node_factory, bitcoind):
...
+@pytest.mark.developer("needs DEVELOPER=1", "uses --dev-fast-reconnect")
 @pytest.mark.openchannel('v1')
 @pytest.mark.openchannel('v2')
 def test_funding_reorg_remote_lags(node_factory, bitcoind):
     """Nodes may disagree about short_channel_id before channel announcement
     """
     # may_reconnect so channeld will restart; bad gossip can happen due to reorg
     opts = {'funding-confirms': 1, 'may_reconnect': True, 'allow_bad_gossip': True,
-            'allow_warning': True}
+            'allow_warning': True, 'dev-fast-reconnect': None}
```
rustyrussell, line 1143: Really? CI runs with TIMEOUT=900 right?

rustyrussell (commented):
Not sure these will actually help, though the fast reconnect is probably a good idea.

rustyrussell:
I'm re-runnign all the CI tests, in case this doesn't actually fix the flake...  Otherwise I'll apply it as it *is* neater and doesn't make things worse...

rustyrussell:
Ack a8e4262d64132b5de7be55ec0100e506d2b8fc98

PR #5458, opened 2022-07-25
decode: fix crash when decoding invalid rune.

rustyrussell:
If rune contains invalid UTF-8, offers (which implements decode) would
produce JSON with invalid UTF-8, which causes lightningd to complain
and kill it, and then die because it's an important plugin.

So don't decode invalid UTF-8!

Reported-by: @jb55
Changelog-None: introduced in this release.


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5458/commits/52c5e0df9628692fad8346b04688af86c447bd74

niftynei:
Something's hanging in the v2 opens path, unrelated to this change. I'm investigating.

PR #5461, opened 2022-07-25
Fix fuzzing build

rustyrussell:
We were supposed to be testing compilation in CI.  We didn't (see second fix), so it bitrotted.

Would be great to get someone to write more fuzzing, and actually run these too!

Changelog-None


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5461/commits/110ee8e5db5ba3d1b281eb5c40a1feeec20f6764

PR #5463, opened 2022-07-25
Test flakes?? (works for me)

niftynei:
Valgrind was choking on some things while running tests locally for the `bookkeeper`; this fixes those up.

Also speeds up a test's reconnect, which would flake locally when TIMEOUT is set low.

### DIFF
--- a/common/wireaddr.c
+++ b/common/wireaddr.c
@@ -337,6 +337,7 @@ bool separate_address_and_port(const tal_t *ctx, const char *arg,
 		*port = strtol(portcolon + 1, &endp, 10);
 		return *port != 0 && *endp == '\0';
 	}
+
 	return true;
 }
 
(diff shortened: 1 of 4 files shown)


rustyrussell:
Ok, you generally shouldn't put valgrind in the issue title, just in the credits!

In this case, the bug is real: port was uninitialised, and tested for zero.

rustyrussell (approved):
Ack 4c8ccb1ed0e2bb397ce84a348abf697cea0f0d57

common/wireaddr.c
```diff
@@ -337,6 +337,7 @@ bool separate_address_and_port(const tal_t *ctx, const char *arg,
 		*port = strtol(portcolon + 1, &endp, 10);
 		return *port != 0 && *endp == '\0';
 	}
+
```
vincenzopalazzo, line 340: unneeded?

rustyrussell:
Trivial rebase on master to stop unrelated flakes.

Ack 86e81d6bda494512643a670e0bb100e49cf90f24

PR #5466, opened 2022-07-26
WIP: Test Fedora 36's default hardening CFLAGS

wtogami:
This is a preliminary test of what happens with typical distro hardening flags when run through CI.

(Nothing like this will be proposed for actual merge. I suspect it doesn't work on some archs or alternate toolchains/musl.)

### DIFF
--- a/configure
+++ b/configure
@@ -47,7 +47,8 @@ usage_with_default()
 default_coptflags()
 {
     if [ "$1" = 0 ]; then
-	echo "-Og"
+	# FORTIFY_SOURCE requires optimization level above zero
+	echo "-O1 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fPIE -fstack-protector-strong"
     fi
 }
 


vincenzopalazzo:
Looks like that breaks alpine https://github.com/ElementsProject/lightning/runs/7526408982?check_suite_focus=true that is the only real compilation testing that we have other than ubuntu I think

wtogami:
> Looks like that breaks alpine

```
mocking out tools/test/test_gen.c.tmp.c:
  CC       crypto_hash/sha512/libsodium_la-hash_sha512.lo
  CC       crypto_hash/sha512/cp/libsodium_la-hash_sha512_cp.lo
  CC       crypto_kdf/blake2b/libsodium_la-kdf_blake2b.lo
/usr/lib/gcc/x86_64-alpine-linux-musl/10.3.1/../../../../x86_64-alpine-linux-musl/bin/ld: /usr/lib/gcc/x86_64-alpine-linux-musl/10.3.1/../../../../lib/Scrt1.o: in function `_start_c':
/home/buildozer/aports/main/musl/src/v1.2.2/crt/crt1.c:18: undefined reference to `main'
collect2: error: ld returned 1 exit status
make[1]: *** [<builtin>: tools/test/test_gen.c.tmp] Error 1
```

It was expected to break with musl due to a different hardening implementation and toolchain, although the particular error message where it failed here is odd.

### Notes:
* Build flags should be customized by the distributor who knows their own toolchain and target architectures. 
* The various submodules built within CLN have their own hard-coded opt flags defaults. Maybe better to leave them that way?
* Should CLN's default CFLAGS include basic hardening? Pretty much all deployed software has hardening enabled these days. However it might make debugging more difficult.

wtogami:
I'd be interested in learning how to replicate the failed test environment locally including Alpine which I've never used. Can anyone explain this in chat?

configure
```diff
@@ -47,7 +47,8 @@ usage_with_default()
 default_coptflags()
 {
     if [ "$1" = 0 ]; then
-	echo "-Og"
+	# FORTIFY_SOURCE requires optimization level above zero
+	echo "-O1 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fPIE -fstack-protector-strong"
```
rustyrussell, line 51: The following should be removed (since this is only optimization flags, debug flags are defined already):

```
g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_GLIBCXX_ASSERTIONS
```

Should add `-Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS` to the 
default_cwarnflags() function (but maybe only for gcc builds...)

PR #5467, opened 2022-07-27
v2 opens and funder flakes

niftynei:
The `funder_contribution_limits` test hang is kind of interesting -- I was assuming that the RPC response to openchannel_update would always beat the notification of tx_signatures arriving since the spec says they MUST always arrive one after the other (and in fact the code doesn't work if they get called out of order). In rare cases however the notification arrives before the RPC response, which causes the hang we've been seeing.

Really easy to fix, but kind of interesting!

The other flake has to do with how fast the connection drops -- if it drops too fast we get an RPC error; otherwise the RPC returns correctly. We don't really care if the RPC works we just need to call it.

"needing to call this RPC no matter if it returns ok or not" has a bit of a smell; I'm planning to do some work on this flow next release. I'm punting on cleaning it up til then.  

Changelog-None

### DIFF
--- a/plugins/spender/openchannel.c
+++ b/plugins/spender/openchannel.c
@@ -621,10 +621,9 @@ funding_transaction_established(struct multifundchannel_command *mfc)
 
 	/* If all we've got is v2 destinations, we're just waiting
 	 * for all of our peers to send us their sigs.
-	 * That callback triggers separately, so we just return
-	 * a 'still pending' here */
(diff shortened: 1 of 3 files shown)


rustyrussell:
Got fix for wumbo flake, restarting...

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5467/commits/ee069a33b783bcb5328a0925736bbdf6ef13969a

PR #5468, opened 2022-07-27
Reconnection simplification, get less aggressive.

rustyrussell:
Also fix a flake I hit testing this locally...

Changelog-None


cdecker:
Looking good, now shepherding through CI

ACK https://github.com/ElementsProject/lightning/pull/5468/commits/a73ddaffa0e2e4dd963af2ea9dd4f03178d84718

rustyrussell:
OK, so backoff was too aggressive for test_feerate_stress test, so have to use dev-fast-reconnect there (we manually disconnect repeatedly, triggering the reconnect backoff logic, which is actually The Right Thing).

Ack https://github.com/ElementsProject/lightning/commit/0c653edf06026f651a4db8217093dcca509f3578

PR #5472, opened 2022-07-28
bkpr: avoid assert failure from subtracting an equal amount of sats

nickfarrow:
If the remote party opened the channel and pushed zero sats, then `amt` and `push_amt` are both zero.
This makes the subtraction `amount_msat_sub` fail, breaking on the next assert(ok).

```
2022-07-28T07:11:13.144Z INFO    plugin-bookkeeper: account 2cdeXXXf12 not found, adding along with new balance
bookkeeper: plugins/bkpr/bookkeeper.c:653: new_missed_channel_account: Assertion `ok' failed.
bookkeeper: FATAL SIGNAL 6 (version v0.11.0.1-498-g8c9fa45)
...
```
Tried to fix by first checking `amount_msat_greater` and conditionally setting `amt = AMOUNT_MSAT(0)` if `amt` is in fact equal or smaller. Not really the nicest, is there a better way to subtract two possibly equal amounts?

haven't had a change to test yet for unrelated reasons

### DIFF
--- a/plugins/bkpr/bookkeeper.c
+++ b/plugins/bkpr/bookkeeper.c
@@ -640,7 +640,7 @@ static bool new_missed_channel_account(struct command *cmd,
 				push_credit = AMOUNT_MSAT(0);
 				push_debit = push_amt;
 			} else {
-				ok = amount_msat_sub(&amt, amt, push_amt);
+				ok = amount_msat_sub(&amt, remote_amt, push_amt);
 				push_credit = push_amt;
 				push_debit = AMOUNT_MSAT(0);
 			}


nickfarrow:
Force pushed to just add missing changelog line to commit msg.

Ran `make check-units` successfully but I don't think this case of `0 - 0` is tested (could add to this PR or a later one)

plugins/bkpr/bookkeeper.c
```diff
@@ -640,7 +640,12 @@ static bool new_missed_channel_account(struct command *cmd,
 				push_credit = AMOUNT_MSAT(0);
 				push_debit = push_amt;
 			} else {
-				ok = amount_msat_sub(&amt, amt, push_amt);
+				if (amount_msat_greater(amt, push_amt)) {
+					ok = amount_msat_sub(&amt, amt, push_amt);
+					assert(ok);
+				} else {
+					amt = AMOUNT_MSAT(0);
+				}
```
vincenzopalazzo, line 648:
```suggestion
				} else
					amt = AMOUNT_MSAT(0);
```

vincenzopalazzo (commented):
LGTM, just a code style comment otherwise it is ready to go

niftynei:
@nickfarrow can you tell me the following values from the channel's `listpeers` result that crashed?

- `funding:{ local_msat:??, remote_msat:??, push_msat:??}`
- and what `opener` is?

nickfarrow:
```
"opener": "remote",
 "funding": {
    "local_msat": "0msat",
    "remote_msat": "2000000000msat",
    "pushed_msat": "0msat"
 },
```

plugins/bkpr/bookkeeper.c
nickfarrow: i see a both of these throughout the repo, when is the latter style preferred? When it's a single statement? (not very used to C)

niftynei:
Do you have any channels where the `push_msat` field > 0msat and the `opener: remote`?
 
`amount_msat_sub` should be robust against equal value subtractions (tested it locally); as it stands there's a bug here in that we should be subtracting from `remote_amt`... but actually we can delete the problematic line entirely and it won't impact anything.

If you want to get it to re-run, I'd suggest deleting the `accounts.sqlite3` database entirely (tho note that you'll lose any routing/invoice data that you've got in it).

niftynei:
fwiw i managed to recreate this locally, patch incoming!!

Thanks for the report!!

nickfarrow:
> Do you have any channels where the `push_msat` field > 0msat and the `opener: remote`?
Yep I do! So it actually crashed on this one?
```
               "opener": "remote",
               "funding": {
                  "local_msat": "0msat",
                  "remote_msat": "800000000msat",
                  "pushed_msat": "400000000msat"
               }, 
```

> `amount_msat_sub` should be robust against equal value subtractions (tested it locally)

ah yeah you're right, works fine with equal values..

PR #5477, opened 2022-07-29
bookkeeper crash fixes

niftynei:
Fixes two crash reports, and adds back some missing functionality for the snapshots in coin-moves.


tests/test_bookkeeper.py
```diff
@@ -336,8 +475,6 @@ def test_bookkeeping_onchaind_txs(node_factory, bitcoind):
 
     # Double check there's no bookkeeper plugin on
     assert l1.daemon.opts['disable-plugin'] == 'bookkeeper'
-    with pytest.raises(RpcError):
-        l1.rpc.listincome()
```
rustyrussell, line 340: Why remove this??

plugins/bkpr/incomestmt.c
```diff
@@ -437,7 +437,7 @@ void json_add_income_event(struct json_stream *out, struct income_event *ev)
 
 const char *csv_filename(const tal_t *ctx, const struct csv_fmt *fmt)
 {
-	return tal_fmt(ctx, "cln_incomestmt_%s_%zu.csv",
+	return tal_fmt(ctx, "cln_incomestmt_%s_%li.csv",
 		       fmt->fmt_name,
 		       time_now().ts.tv_sec);
```
rustyrussell, line 442: That will break on some platforms too!  Use an "unsigned long" cast...

tests/test_bookkeeper.py
niftynei: It fails now b/c the command name moved to `bkpr_listincome`, not because it's not there. It's technically a functional check of the above (deleting `bookkeeper` plugin)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/31e5b8234d3d690c4e26695543aac30e16ea3c92

PR #5483, opened 2022-07-31
bkpr: track channel rebalances, display in listincome

niftynei:
Self-transfers should really be ignored for income reports. This change adds rebalance tracking,
which allows us to report the right thing for `listincome`.

Tracks rebalances, and report income events for them.

Previously `listincome` would report:
	- invoice event, debit, outgoing channel
	- invoice_fee event, debit, outgoing channel
	- invoice event, credit, inbound channel

Now reports:
	- rebalance_fee, debit, outgoing channel
	(same value as invoice_fee above)

Note: only applies on channel events; if a rebalance falls to chain
we'll use the older style of accounting.

Changelog-None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,6 @@ contrib/pyln-*/dist/
 contrib/pyln-*/pyln_*.egg-info/
 contrib/pyln-*/.eggs/
 contrib/pyln-*/pyln/*/__version__.py
-release/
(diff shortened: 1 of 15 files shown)


rustyrussell (approved):
Ack 2df38ddabff7332e8dd6cb22cf3742bf30980a0c

PR #5484, opened 2022-08-01
bkpr: insert obscure 60s pop references.

rustyrussell:
The initial snapshots on an already-running lightningd are expected to
be unbalanced, but this shouldn't cause users to long for the green,
green grass of home.

This controls the Art of Noise.

Changelog-None


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5484/commits/2e853e1961f5ea149306f4f7f23b95d733ee4769

Issue #4928, opened 2021-11-23
wallet/test/run-db and wallet/test/run-wallet crash (SIGSEGV) if built without SQLite support


whitslack:
### Issue and Steps to Reproduce

1. Configure and build C-Lightning on a system lacking the SQLite library (or poison the configurator test to force `HAVE_SQLITE3=0`).
2. Run the unit tests.
3. `wallet/test/run-db` and `wallet/test/run-wallet` will segfault.

#### Backtrace of `wallet/test/run-db` at SIGSEGV
```
Program received signal SIGSEGV, Segmentation fault.
0x00005555555933f4 in db_open (ctx=ctx@entry=0x0, filename=filename@entry=0x5555555b8b48 "sqlite3:///tmp/ldb-0P80Fy") at ./wallet/db.c:1169
1169            if (db->config->setup_fn && !db->config->setup_fn(db))
(gdb) bt
#0  0x00005555555933f4 in db_open (ctx=ctx@entry=0x0, filename=filename@entry=0x5555555b8b48 "sqlite3:///tmp/ldb-0P80Fy")
    at ./wallet/db.c:1169
#1  0x00005555555934d2 in create_test_db () at wallet/test/run-db.c:87
#2  0x000055555555ba02 in test_empty_db_migrate (ld=0x5555555b83d8) at wallet/test/run-db.c:95
#3  main (argc=<optimized out>, argv=0x7fffffffd398) at wallet/test/run-db.c:179
(gdb) print db
$1 = (struct db *) 0x5555555b8bb8
(gdb) print *db
$2 = {filename = 0x5555555b8c38 "sqlite3:///tmp/ldb-0P80Fy", in_transaction = 0x0, conn = 0x0, config = 0x0, 
  changes = 0x5555555b8dc8, pending_statements = {n = {next = 0x5555555b8be0, prev = 0x5555555b8be0}}, error = 0x0, log = 0x0, 
  dirty = false, data_version = 0}
```

#### Backtrace of `wallet/test/run-wallet` at SIGSEGV
```
Program received signal SIGSEGV, Segmentation fault.
0x00005555555a63c6 in db_open (ctx=ctx@entry=0x5555555e7da8, filename=filename@entry=0x5555555e7e38 "sqlite3:///tmp/ldb-oooEEt") at ./wallet/db.c:1169
1169            if (db->config->setup_fn && !db->config->setup_fn(db))
(gdb) bt
[...]

PR #5485, opened 2022-08-01
bkpr: make unit tests not fail if !HAVE_SQLITE3

niftynei:
We rely on sqlite3 being present to run unit tests for some bookkeeper
tests; here we effectively disable these tests if not available

Fixes report in #4928
Reported-By: @whitslack


rustyrussell (approved):
Ack 3d8bac7944152edcec4343b071f15a16619f9f72

PR #5487, opened 2022-08-02
chore: remove EXPERIMENTAL for DNS Bolt7

m-schmoock:
This has been merged https://github.com/lightning/bolts/pull/911
Should we remove EXPERIMENTAL?

Changelog-Changed: DNS annoucenent support is no longer EXPERIMENTAL

### DIFF
--- a/connectd/connectd.c
+++ b/connectd/connectd.c
@@ -766,14 +766,12 @@ static void try_connect_one_addr(struct connecting *connect)
 	bool use_proxy = connect->daemon->always_use_proxy;
 	const struct wireaddr_internal *addr = &connect->addrs[connect->addrnum];
 	struct io_conn *conn;
-#if EXPERIMENTAL_FEATURES /* BOLT7 DNS RFC #911 */
 	bool use_dns = connect->daemon->use_dns;
 	struct addrinfo hints, *ais, *aii;
 	struct wireaddr_internal addrhint;
 	int gai_err;
 	struct sockaddr_in *sa4;
 	struct sockaddr_in6 *sa6;
-#endif
 
 	assert(!connect->conn);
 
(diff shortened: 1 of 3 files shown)


connectd/connectd.c
```diff
@@ -875,12 +872,6 @@ static void try_connect_one_addr(struct connecting *connect)
 				addr = &connect->addrs[connect->addrnum];
 			}
 			freeaddrinfo(ais);
-#endif
-			tal_append_fmt(&connect->errors,
-				       "%s: EXPERIMENTAL_FEATURES needed. ",
-				       type_to_string(tmpctx,
-						      struct wireaddr_internal,
-						      addr));
 			goto next;
```
m-schmoock, line 875: Can someone else also confirm that `goto next;` is correct here? Or do we need a `break;` ?
I'm a bit confused right now :D

niftynei: This `#endif` looks like it was originally done a bit incorrectly; I'd expect there to be an `if` statement around the `experimental features needed` error?

That aside, `goto next` is used when you want to skip the current address info attempt without trying to connect to it.
Based on the fact that the code in the experimental block (if successful) adds a new address to try to the list of addresses, going to the next seems appropriate to me. (Go try the next address which we've now added to the list of addresses to try)

m-schmoock: Tanks that confirms my thoughts. Will do a rebase on new master...

m-schmoock:
Also, we have confirmation that LND now handles unknown descriptor types correctly:
https://github.com/lightningnetwork/lnd/issues/6337#issuecomment-1233315763

m-schmoock:
@niftynei @cdecker 
Just the usual `mrkd` and `mako` test flakes left...

rustyrussell (approved):
Ack 122a588c25c8aa2348e3821c7cfaa5ec5c9291a9

PR #5489, opened 2022-08-03
lightningd: fix funding_locked in channel_opened notification.

rustyrussell:
Previously, "funding_locked" was always "true"!

(It's actually been wrong since its introduction in v0.7.3!)

Changelog-Fixed: Plugins: `channel_opened` notification `funding_locked` field is now accurate: was always `true`.


niftynei:
oh nice! This particularly becomes important for zerconfs eh.

ACK https://github.com/ElementsProject/lightning/pull/5489/commits/bd6e97b0011cbfaa7fdb48b78d2bee99fd36471e

trueptolemy:
Sorry for late response. Good work! :-)

PR #5494, opened 2022-08-03
pay: Fix a memory leak when retrying getroute

cdecker:
If we retry a getroute call we could end up losing track of a prior route. This could happen if we were unable to apply the channel hints due to insufficient balance or number of HTLCs.


niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5494/commits/2a259cd82581859555f695540a96b4ef58006789

PR #5501, opened 2022-08-08
Alias privacy, as per BOLT 2

rustyrussell:
The purpose of short-channel-id aliases is to remove the requirement to expose real short-channel-ids.  Unfortunately, our implementation falls short of this; part of that is because we don't track channel_type explicitly.

This meets the requirements, by assuming all peers who support aliases at open time want us to use them exclusively in private channels.

@niftynei Note the new Changelog lines!


niftynei (commented):
ACK https://github.com/ElementsProject/lightning/pull/5501/commits/c32159635a57b187a15846584dbcd054121d72f7

PR #5503, opened 2022-08-08
bookkeeper: dont skip zero balance channels that are missing

niftynei:
We weren't adding zero balance channels during "catch up starts (accounting for things missing while bookkeeper wasn't running)", which resulted in a crash/restart when any payment got routed through these channels.

Also patches a flake (which really needs a nicer fix than this line/by line patch but yolo) and removes a duplicated log statement.

Changelog-None.

### DIFF
--- a/tests/test_bookkeeper.py
+++ b/tests/test_bookkeeper.py
@@ -539,15 +539,18 @@ def test_bookkeeping_onchaind_txs(node_factory, bitcoind):
     Test for a channel that's closed, but whose close tx
     re-appears in the rescan
     """
+    coin_mvt_plugin = Path(__file__).parent / "plugins" / "coin_movements.py"
     l1, l2 = node_factory.line_graph(2,
                                      wait_for_announce=True,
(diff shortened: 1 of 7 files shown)


rustyrussell (approved):
Ack 751f13bdfb421cbe81520701a51447f002dc2ae5

tests/test_bookkeeper.py
```diff
@@ -145,7 +145,7 @@ def test_bookkeeping_external_withdraws(node_factory, bitcoind):
         assert inc['credit_msat'] == amount_msat
     # The event should show up in the 'bkpr_listaccountevents' however
     events = l1.rpc.bkpr_listaccountevents()['events']
-    assert len(events) == 3
+    assert len(events) == 4
```
niftynei, line 148: we now record an extra event for wallet balances (starting at zero)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/3d9ff537b45a4be6a469b938cb58f310ca66345a

PR #5506, opened 2022-08-08
Don't crash if the PSBT is junk

niftynei:
If you pass in a PSBT that's invalid (no inputs/outputs for example), lightningd blows up in the HSM.

Instead, let's catch bad PSBTs before we get that far.

Also uncovered a kinda fun "double entrance" into `tal_wally` code.

Waiting for confirmation from @fiatjaf


rustyrussell (approved):
Hmm, this fails on an already-signed PSBT?  Seems wrong...

rustyrussell (changes requested):
CI gets upset, since it seems an already-signed PSBT fails this check?

rustyrussell (approved):

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/b7c3b29372a0aef18a1b09ca41a8fd06ef2869e2

PR #5509, opened 2022-08-09
plugins/bcli: load RPC password from stdin instead of an argument

### LINKED ISSUE #3984: Avoid leaking bitcoin auth secrets to other users
On Linux, running `ps -e -o args` frequently enough will eventually show the bitcoin-cli child process spawned by c-lightning including the RPC authentication arguments. The issue is that by default *any* user on the system can get the authentication secrets doing that - not just the user the c-lightning and bitcoin-cli processes run as.

I can reliable reproduce this with a fresh c-lightning datadir, `--network main` and running the following script under a different user:
```
#!/bin/bash

while :
do
    RES=$(ps -e -o args | grep -v grep | grep bitcoin-cli)
    N=$(echo "$RES" | wc -l)
    if [ "$N" -gt 1 ]; then
        echo "$RES"
    fi
done
```

This problem can be mitigated by mounting `/proc` with the `hidepid` option. However, process arguments seem to be generally not considered secret. For example, even with `hidepid` if you run lightningd as a systemd service, other users can see the arguments of bitcoin-cli through `systemctl status` (H/T @nixbitcoin, see https://github.com/systemd/systemd/issues/16825).

A naive idea for how c-lightning could prevent this is if it created a temporary config file for bitcoin-cli with the authentication secrets and provides the file via the `--config` argument.

seberm:
Passing a RPC password via cli argument (`-rpcpassword`) can leak auth secrets to other users on a system (at least on Linux). This PR tries to pass the RPC password through a stdin using a `-stdinrpcpass`

Refs.:
- Fixes #3984
- https://man.archlinux.org/man/bitcoin-cli.1.en

Fixes #3984

### DIFF
--- a/plugins/bcli.c
+++ b/plugins/bcli.c
@@ -4,6 +4,7 @@
 #include <ccan/cast/cast.h>
 #include <ccan/io/io.h>
 #include <ccan/pipecmd/pipecmd.h>
+#include <ccan/read_write_all/read_write_all.h>
 #include <ccan/tal/grab_file/grab_file.h>
 #include <ccan/tal/str/str.h>
 #include <common/json_param.h>
@@ -121,8 +122,10 @@ static const char **gather_argsv(const tal_t *ctx, const char *cmd, va_list ap)
 	if (bitcoind->rpcuser)
 		add_arg(&args, tal_fmt(args, "-rpcuser=%s", bitcoind->rpcuser));
 	if (bitcoind->rpcpass)
-		add_arg(&args,
-			tal_fmt(args, "-rpcpassword=%s", bitcoind->rpcpass));
+		// Always pipe the rpcpassword via stdin. Do not pass it using an
+		// `-rpcpassword` argument - secrets in arguments can leak when listing
+		// system processes.
+		add_arg(&args, "-stdinrpcpass");
 
 	add_arg(&args, cmd);
 	while ((arg = va_arg(ap, char *)) != NULL)
@@ -290,6 +293,7 @@ static void next_bcli(enum bitcoind_prio prio)
 {
 	struct bitcoin_cli *bcli;
 	struct io_conn *conn;
+	int in;
 
 	if (bitcoind->num_requests[prio] >= BITCOIND_MAX_PARALLEL)
 		return;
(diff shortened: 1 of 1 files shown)


plugins/bcli.c
```diff
@@ -298,8 +300,15 @@ static void next_bcli(enum bitcoind_prio prio)
 	if (!bcli)
 		return;
 
-	bcli->pid = pipecmdarr(NULL, &bcli->fd, &bcli->fd,
+	int in;
+	bcli->pid = pipecmdarr(&in, &bcli->fd, &bcli->fd,
 			       cast_const2(char **, bcli->args));
+
+	if (bitcoind->rpcpass)
+		write(in, bitcoind->rpcpass, strlen(bitcoind->rpcpass));
```
rustyrussell, line 308: This needs to be write_all(), as it can be unreliable with partial writes.

plugins/bcli.c
```diff
@@ -298,8 +300,15 @@ static void next_bcli(enum bitcoind_prio prio)
 	if (!bcli)
 		return;
 
-	bcli->pid = pipecmdarr(NULL, &bcli->fd, &bcli->fd,
+	int in;
```
rustyrussell, line 303: Minor: We declare these at the top of the block / function by convention, BTW

plugins/bcli.c
```diff
@@ -863,7 +872,13 @@ static void wait_and_check_bitcoind(struct plugin *p)
 	for (;;) {
 		tal_free(output);
 
-		child = pipecmdarr(NULL, &from, &from, cast_const2(char **,cmd));
+		child = pipecmdarr(&in, &from, &from, cast_const2(char **, cmd));
+
+		if (bitcoind->rpcpass)
+			write(in, bitcoind->rpcpass, strlen(bitcoind->rpcpass));
```
rustyrussell, line 878: here too...

rustyrussell (approved):
Minor changes only...

seberm:
Everything should be fixed now, thanks for your review!

rustyrussell:
Hmm can you add a `Changelog-Fixed: bcli: don't expose bitcoin RPC password on commandline` to the commit message, so we make sure it gets noted in the CHANGELOG.md for next release?

Thanks!
Ack da47e8ea9d51dbeb37b3f2d9f84a5eb4085ca4a7

seberm:
Sure, I've added the changelog info to commit message and I've fixed the order of includes.

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5509/commits/af4335cb797323067123affbca97d35c03607366

vincenzopalazzo:
We can add the magic keyword Fixes https://github.com/ElementsProject/lightning/issues/3984 to avoid to leave the issue open?

seberm:
I've added the `Fixes` keyword into my first comment. It seems the #3984 issue is now linked to close automatically.

rustyrussell:
Trivial rebase, should kick CI.

Ack https://github.com/ElementsProject/lightning/commit/be29c1fb3ed821a206ad4947cef6f5471a13ba39

PR #5510, opened 2022-08-10
Fix derived_secret, use correct size of secretstuff.derived secret

ksedgwic:
I was reproducing this calculation in the VLS code, don't think the CLN version is correct.

This will invalidate some integration test results, ie `tests/test_plugin.py::test_commando_rune`


ksedgwic:
@niftynei, @adi2011

rustyrussell:
!!

rustyrussell:
IOU a beer for finding this before release!  Also, I've invited you to be a collaborator so you can attach milestones, etc to stuff in future!

rustyrussell:
Cancelled CI workflow on this, since we know two tests will fail.  Those are fixed in #5511

adi2011:
Ack `89e6e36` 
:smiling_face_with_tear:

PR #5513, opened 2022-08-10
Postgres crash fix

rustyrussell:
Note new Changelog line!

### DIFF
--- a/tests/test_opening.py
+++ b/tests/test_opening.py
@@ -1205,6 +1205,41 @@ def test_funder_contribution_limits(node_factory, bitcoind):
     assert l3.daemon.is_in_log(r'calling `signpsbt` .* 7 inputs')
 
 
+@pytest.mark.openchannel('v2')
+@pytest.mark.developer("requres 'dev-disconnect'")
+def test_inflight_dbload(node_factory, bitcoind):
+    """Bad db field access breaks Postgresql on startup with opening leases"""
+    disconnects = ["@WIRE_COMMITMENT_SIGNED"]
+    l1, l2 = node_factory.get_nodes(2, opts=[{'experimental-dual-fund': None,
+                                              'dev-no-reconnect': None,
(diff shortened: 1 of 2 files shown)


wallet/wallet.c
```diff
@@ -1133,7 +1134,7 @@ wallet_stmt2inflight(struct wallet *w, struct db_stmt *stmt,
 	if (!db_col_is_null(stmt, "lease_commit_sig")) {
 		lease_commit_sig = tal(tmpctx, secp256k1_ecdsa_signature);
 		db_col_signature(stmt, "lease_commit_sig", lease_commit_sig);
-		lease_chan_max_msat = db_col_int(stmt, "lease_chan_max_msat");
+		lease_chan_max_msat = db_col_u64(stmt, "lease_chan_max_msat");
```
jb55, line 1137: ACK, I was wondering how this works on the sqlite site but it looks like sqlite integers are 8 bytes anyways...

niftynei:
note: test flake is patched in #5503

tests/test_opening.py
```diff
@@ -1205,6 +1205,41 @@ def test_funder_contribution_limits(node_factory, bitcoind):
     assert l3.daemon.is_in_log(r'calling `signpsbt` .* 7 inputs')
 
 
+@pytest.mark.openchannel('v2')
+@pytest.mark.developer("requres 'dev-disconnect'")
+def test_inflight_dbload(node_factory, bitcoind):
+    """Bad db field access breaks Postgresql on startup with opening leases"""
```
niftynei, line 1211: extremely minor nit: technically it only breaks if you try to load inflights from a postgres db, which happens in the case that a channel open isn't resolved at startup

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5513/commits/1da6c4031e050a5a21723d96bca7ef16c4340257

PR #5519, opened 2022-08-12
Fix bsd workflow

gruve-p:


vincenzopalazzo (commented):
What we are fixing here?

gruve-p:
FreeBSD test workflow has been disabled since one month because it keeps failing. This PR fixes that so the workflow can be re-enabled again.

vincenzopalazzo:
mh! it was disabled because was flaky related to some ssh error, but I leave this review to @cdecker because I have no clue for this freebsd CI part!

gruve-p:
> mh! it was disabled because was flaky related to some ssh error, but I leave this review to @cdecker because I have no clue for this freebsd CI part!

yes this PR fixes that SSH error

Issue #5517, opened 2022-08-11
Testing v0.12.0rc2, some bookkeeper related tests fail


SimonVrouwe:
### Issue and Steps to Reproduce
```
git checkout v0.12.0rc2
make clean && ./configure --enable-developer --enable-experimental-features --disable-rust && make
```
Some tests fail: 
```
FAILED tests/test_bookkeeper.py::test_bookkeeping_closing_trimmed_htlcs - assert 0 == 1
FAILED tests/test_bookkeeper.py::test_bookkeeping_closing_subsat_htlcs - assert 0 == 1
FAILED tests/test_closing.py::test_penalty_htlc_tx_timeout - assert False
FAILED tests/test_closing.py::test_penalty_rbf_burn - assert 0 == 1
FAILED tests/test_closing.py::test_onchain_their_unilateral_out - AssertionError
FAILED tests/test_opening.py::test_zeroconf_public - AssertionError
```
```
Bitcoin Core version v23.0.0
```
With a normal build at least `test_bookkeeping_closing_subsat_htlcs` fails, see below for partial log.
What I found so far is that bkpr plugin seems to miss `close_fee`:
```
>       assert len(close_fee) == 1
E       assert 0 == 1
E        +  where 0 = len([])

tests/test_bookkeeper.py:99: AssertionError
```
```
(cldev) simon@gamepc:~/git/SimonVrouwe/lightning$ pytest tests/ -s -k 'test_bookkeeping_closing_subsat_htlcs'
============================================================================================================================= test session starts =============================================================================================================================
platform linux -- Python 3.7.3, pytest-7.1.2, pluggy-1.0.0
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/simon/git/SimonVrouwe/lightning
plugins: xdist-2.5.0, timeout-2.1.0, benchmark-3.2.2, flaky-3.7.0, test-groups-1.0.3, forked-1.4.0, custom-exit-code-0.3.0, rerunfailures-9.1.1
collected 526 items / 525 deselected / 1 selected                                                                                                                                                                                                                             

tests/test_bookkeeper.py Running tests in /tmp/ltests-gecvmt23
Throttler delayed startup for 0.0010685920715332031 seconds
Throttler delayed startup for 0.1016073226928711 seconds
-- snip --
[...]

PR #5522, opened 2022-08-15
Strip "NULLS FIRST" from sqlite3 queries

niftynei:
Keywords were added "recently" (2019) and were causing silent failures in updates on machines using older versions of sqlite3.

These patches strip the keywords out for sqlite3 (which is fine, nulls first *is* the default behavior on sqlite3, it was only problematic on postgres). 

Nicer fix would be to allow for versioning of queries depending on sqlite3 version but yolo?

Changelog-None


rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5522/commits/c9ab02b3158d9bbaf083a318ff6d1eaff688e665

niftynei:
Fixes #5517

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5522/commits/c9ab02b3158d9bbaf083a318ff6d1eaff688e665

The alternative woul be use the `ORDER BY`, but this will make the query more complex i think!

PR #5525, opened 2022-08-16
fix: Zeroconf in combination with multiple channels tries to log a non-existent scid

cdecker:
When forwarding an HTLC we're allowed to pick another channel than the one specified in the onion. We do so if using another channel is advantageous. In this case we'd have a confirmed channel and a zeroconf channel in parallel, and the latter would be picked because it is advantageous (more `spendable_msat`), and we'd try to log its `scid`. However the `scid` for a zeroconf channel is not set until it gets at least one confirmation, causing a NULL-access.

### DIFF
--- a/tests/test_opening.py
+++ b/tests/test_opening.py
@@ -1575,3 +1575,41 @@ def test_scid_alias_private(node_factory, bitcoind):
     route[1]['channel'] = alias23
     l1.rpc.sendpay(route, inv['payment_hash'], payment_secret=inv['payment_secret'])
     l1.rpc.waitsendpay(inv['payment_hash'])
+
+
+@pytest.mark.xfail("Reproducing a crash", strict=True)
+def test_zeroconf_multichan_forward(node_factory):
+    """The freedom to choose the forward channel bytes us when it is 0conf
+
+    Reported by Breez, we crashed when logging in `forward_htlc` when
+    the replacement channel was a zeroconf channel.
+
+    l2 -> l3 is a double channel with the zeroconf channel having a
+    higher spendable msat, which should cause it to be chosen instead.
+
+    """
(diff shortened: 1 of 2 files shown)


tests/test_opening.py
```diff
@@ -1575,3 +1575,41 @@ def test_scid_alias_private(node_factory, bitcoind):
     route[1]['channel'] = alias23
     l1.rpc.sendpay(route, inv['payment_hash'], payment_secret=inv['payment_secret'])
     l1.rpc.waitsendpay(inv['payment_hash'])
+
+
+@pytest.mark.xfail("Reproducing a crash", strict=True)
+def test_zeroconf_multichan_forward(node_factory):
+    """The freedom to chose the forward channel bytes us when it is 0conf
```
endothermicdev, line 1582:
```suggestion
    """The freedom to choose the forward channel bites us when it is 0conf.
```

endothermicdev (approved):
Is there a reason to leave the xfail marker now that the scid alias is used (with the second commit)? Looks good apart from a spelling nit.
ACK 948c287

cdecker:
> Is there a reason to leave the xfail marker now that the scid alias is used (with the second commit)? Looks good apart from a spelling nit. ACK [948c287](https://github.com/ElementsProject/lightning/commit/948c287cc51f6b302a25b8c20c7e27335e7d5137)

Good point, reordering the commits I forgot to un-xfail the test. Will fix up :+1:

cdecker:
Range diff is here: https://bot.bitcoinstats.com/ElementsProject/lightning/pull/5525/range_diff/948c287cc51f6b302a25b8c20c7e27335e7d5137..548697e709a75b0d75f6cf007a91de54e1aba358

niftynei:
ACK https://github.com/ElementsProject/lightning/pull/5525/commits/548697e709a75b0d75f6cf007a91de54e1aba358

PR #5529, opened 2022-08-17
docs: use 'postgres' driver (not postgresql). Include plain url to guide

cdecker:
Apologies @lightningorb, I have no idea what happened to #5521, so I'm reopening the PR with this amended changeset.

### DIFF
--- a/doc/BACKUP.md
+++ b/doc/BACKUP.md
@@ -394,13 +394,13 @@ it just gains the option to use a PostgreSQL database as well.
 If you just want to use PostgreSQL without using a cluster (for
 example, as an initial test without risking any significant funds),
 then after setting up a PostgreSQL database, you just need to add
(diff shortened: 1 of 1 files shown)


lightningorb:
@cdecker thanks for including that tinyurl link. This looks good.

I didn't want to give up on the issue so tried building sphinx docs locally, and the .md link was fine. I then created an RTD of my branch (had to create a .readthedocs.yaml) however the issue was still taking place. I poked around in sphinx for a bit but gave up as i can't repro the issue here.

cdecker:
Yep, RTD assuming that links ending in `.md` seems to be a common occurence, and the short URL should avoid that issue by sidestepping it. It's a bit less clean than fixing the root cause, but we're not sphinx or RTD devs so let them figure this out ^^

PR #5533, opened 2022-08-17
plugin: make the plugin aware of deprecated api flag

vincenzopalazzo:
This PR put the deprecated apis flag inside the plugin init payload to give the possibility to a plugin to be aware of some deprecated API, and more important to adopt the same deprecated workflow of core lightning in case this plugin will proxy some cln command.

Use case covered:

- Give the plugin the possibility to use old API if available, or avoid to use it. eg: splitting listpeers in 2 different command.
- Be able to deprecated some field in the plugin and follow the same rules of cln.

Changelog-Added: plugin: add deprecated api flag in the plugin init method.

Release target > 0.12.0

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -337,7 +337,8 @@ simple JSON object containing the options:
         "port": 9050
     },
     "torv3-enabled": true,
-    "always_use_proxy": false
+    "always_use_proxy": false,
+    "deprecated-api-allowed": false,
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


rustyrussell:
It's already in the "getmanifest" call:

```
	json_add_bool(req->stream, "allow-deprecated-apis", deprecated_apis);
```

Which is often where it's needed (you might not want to register deprecated apis at all).

vincenzopalazzo:
Ah!

>The getmanifest method is required for all plugins and will be called on startup with optional parameters (in particular, it may have allow-deprecated-apis: false, but you should accept, and ignore, other parameters).

I looked only to the payload :/ thanks for he catch!

PR #5536, opened 2022-08-18
Allow msggen to generate `#[deprecated]` attr

joemphilips:
I've been using msggen for generating my package.
I want to generate deprecated fields so that consumers of my package does not face the breaking change when the fields are deprecated, instead, I want them to notice that fields are deprecated by compiler (or IDE) warning.
Since `Field` does not have information about deprecation, there is no way I can do that on my side.

I suppose this happens in the rust client too. so I changed `rust.py` to generate `#[deprecated]` attribute

Note: I'm expecting that in the future, when the schema deprecates a field it should just mark as "deprecate" without deleting other fields in the property.

Changelog-None


joemphilips:
I'm not sure why CI is failing :(

cdecker:
> I'm not sure why CI is failing :(

Don't worry, CI is rather flaky, and we kick it when necessary 🙂

cln-grpc/proto/node.proto
```diff
@@ -71,6 +71,7 @@ message GetinfoResponse {
 	string lightning_dir = 9;
 	uint32 blockheight = 11;
 	string network = 12;
+	optional uint64 msatoshi_fees_collected = 18;
```
cdecker, line 74: Notice that `optional` is no longer supported in proto files, and will cause some compilers to fail.

cdecker: Also was this edited manually? I can't find a place where the `optional` annotation comes from.

cdecker (changes requested):

cln-grpc/proto/node.proto
joemphilips: It is not edited manually, it seems that `optional` has already been used to represent optional fields (see e.g. https://github.com/ElementsProject/lightning/blob/0abe2e3af1ccd9941992266b65f85aa32d3ccb23/cln-grpc/proto/node.proto#L77)

Although this seems not directly related to this PR, I can change `grpc.py` to stop using it if you want me to.
See commit  3e94f7ffb58e7639028dd34ded33ccdc9066fe11 for the change. I will drop if you don't like it.

cdecker: Hm, my bad, let's keep the `optional` for now, otherwise we end up changing more semantics than absolutely required. Sorry :disappointed:

joemphilips: Dropped https://github.com/ElementsProject/lightning/commit/3e94f7ffb58e7639028dd34ded33ccdc9066fe11
And rebased.

cdecker (approved):

cdecker:
Excellent, thanks. I'll make sure this gets through CI and merge asap ^^

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5536/commits/2f301e33d225f533b7de1a4b73dd4d71474c0272

PR #5537, opened 2022-08-18
Skip over saved channels while recovering from emergencyrecover

adi2011:
Currently, If a user uses `emergencyrecover` and they already have the channels stored in the Database, it throws a UNIQUE constraint error and the node goes down. Which is horrible UX. So now, we check for every channel and skip over those which are already stored.

This completely slipped out of my mind while developing, All thanks to @niftynei for being patient with me, and reminding me about this.


niftynei (commented):
ACK 46ddc8e9a89f680ddf2c1ba1968f324b7168ce82

Issue #5496, opened 2022-08-04
Core-Lightning and Eclair interoperability: Channel is not closing


engenegr:
I have several instances of strange behavior:

1. Mainnet issue. Channel between Eclair and CLightning is in closing state for month or so. The closing was started due to fee mismatch and this is known interoperability issue. The concern of this report is why channel is in closing state for one month. What I see from logs:

```
2022-08-03 07:06:23,528 ERROR f.a.e.c.TransportHandler CON n:020fc7e62c9956a34bcdba309a4809ccfd7adcae4b8b644cac81b4dc58d43a2479 - cannot deserialize ByteVector(53 bytes, 0xa4657b2276657273696f6e223a302c22617373657473223a5b22627463225d2c22706565725f616c6c6f776564223a66616c73657d): message: cannot acquire 252176 bits from a vector that contains 392 bits
```

2. Testnet issue. I tried to open a channel from my Simple Bitcoin Wallen which uses some Eclair code but mostly works on Immortan library. The channel was in pending state even some other channel I've opened simultaneously but with Eclair, successfully opened. I tried to close it from CLN and now CLN shows `CHANNELD_SHUTTING_DOWN` but funds aren't moving anywhere. I couldn't retrieve logs from SBW but it looks like I observe similar behavior as reported with Eclair mainnet node.

rustyrussell:
OK, first, our message is backwards: "We've send shutdown, waiting for theirs" (which, shown above, they never send!).

Now I can't see from their error message what message they're upset about.  But we can probably figure it out if you do "log-level:io" (This will produce a *lot* of output, but we can see what we send that might contain "a4657b2276657273696f6e223a302c22617373657473223a5b22627463225d2c22706565725f616c6c6f776564223a66616c73657d"...

sangbida:
This issue is being closed as part of issue cleanup.

**Reason**: Mentioned in git commits (likely fixed); About deprecated feature

If you believe this issue should remain open, please comment and we can reopen it.

PR #5538, opened 2022-08-19
channeld: correct reversed shutdown message in billboard.

rustyrussell:
It was backwards, which made #5496 confusing.

Changelog-Fixed: `listpeers` `status` "They've sent shutdown" and "We've sent shutdown" were backwards.


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5538/commits/1e61eff17d1607c77b6d2786bd3430ce43b40647

Issue #5410, opened 2022-07-11
test_gossip_store_compact_on_load


niftynei:
test_gossip_store_compact_on_load

```
expected: >       wait_for(lambda: l2.daemon.is_in_log(r'gossip_store_compact_offline: [5-8] deleted, 9 copied'))
actual: 2022-07-06T05:04:10.014Z DEBUG   gossipd: gossip_store_compact_offline: 9 deleted, 9 copied
```

endothermicdev:
I haven't been able to replicate this flake locally.  The gossip store should look something like this under the nominal case:

1 DELETED Private channel_announcement
2 DELETED private channel_update
3 DELETED private channel_update
4 Push channel_announcement
5 channel amount
6 DELETED push channel_update
7 DELETED push channel_update
8 Push node_announcement
9 node_announcement
10 Push channel_update
11 private channel_announcement
12 DELETED private channel_update
13 DELETED private channel_update
14 Push channel_update
15 private channel_update
16 private channel_update

I thought this may be related to the gossip spam rewrite, but there were no spam entries in the CI logs, so I haven't found a connection.  For debugging purposes, I would like to add a call to devtools/dump-gossipstore before the compact occurs (if that sounds reasonable.)  I've done this locally, but without encountering the flake.

niftynei:
#5483 hit this. log from the test fail attached.

[5_Build.txt](https://github.com/ElementsProject/lightning/files/9283883/5_Build.txt)

niftynei:
archiving this

PR #5543, opened 2022-08-22
pytest: Add debugging to test_gossip_store_compact_on_load

endothermicdev:
It looks like we still occasionally hit this flake. It has been difficult to reproduce, so let's dump the gossip store to aid in debugging. See issue #5410.

Changelog-None

### DIFF
--- a/tests/test_gossip.py
+++ b/tests/test_gossip.py
@@ -1715,6 +1715,11 @@ def test_gossip_store_load_no_channel_update(node_factory):
 def test_gossip_store_compact_on_load(node_factory, bitcoind):
     l2 = setup_gossip_store_test(node_factory, bitcoind)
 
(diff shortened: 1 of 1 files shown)


cdecker:
There is also a python library to investigate the `gossip_store` file if you want to wait for something specific in the tests. Not that it'd be useful here, but it gives you fine-grained control over things to look for (and we could add a dump method with colors 🌈)

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/b49940fb1d6dc48e67a710d3ea4b01f1b11a29bd

cdecker:
You're right, it's returning some things as unparsed bytes, that's simply because at the time I was investigating a specific issue where the channel amount wasn't interesting (local updates we're getting ignored by the pay plugin for some reason), hence I used the return-as-bytes thing as an escape hatch for types I hadn't implemented yet.

Should just be a matter of implementing them and wiring them into the dispatch table though.

PR #5550, opened 2022-08-25
lightningd: refuse to upgrade db on non-released versions by default.

rustyrussell:
This is a good sanity check that users understand that if they upgrade
to master mid-cycle they can't go back!

Suggested-by: @wtogami
Changelog-Added: Config: `--database-upgrade=true` required if a non-release version wants to (irrevocably!) upgrade the db.

### DIFF
--- a/doc/lightning-listconfigs.7.md
+++ b/doc/lightning-listconfigs.7.md
@@ -62,6 +62,7 @@ On success, an object is returned, containing:
 - **experimental-offers** (boolean, optional): `experimental-offers` field from config or cmdline, or default
 - **experimental-shutdown-wrong-funding** (boolean, optional): `experimental-shutdown-wrong-funding` field from config or cmdline, or default
 - **experimental-websocket-port** (u16, optional): `experimental-websocket-port` field from config or cmdline, or default
+- **database-upgrade** (boolean, optional): `database-upgrade` field from config or cmdline
 - **rgb** (hex, optional): `rgb` field from config or cmdline, or default (always 6 characters)
 - **alias** (string, optional): `alias` field from config or cmdline, or default
 - **pid-file** (string, optional): `pid-file` field from config or cmdline, or default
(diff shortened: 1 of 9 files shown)


cdecker:
Looks like we are getting an access to uninitialized memory in `options.c`: `ld->db_upgrade_ok` seems to be uninitialized.

Sjors:
What about release candidates?

wtogami:
> What about release candidates? (I tend to start those using systemd so it's a slight inconvenience to relaunch with the appropriate argument)

You need only launch it differently once.  You could also  have`database-upgrade=true` in your config file if you rather opt into the old behavior.

rustyrussell:
> What about release candidates? (I tend to start those using `systemd` so it's a slight inconvenience to relaunch with the appropriate argument)

Yeah, tell systemd to stop, do `lightningd --database-upgrade=true --offline` and then kill it and tell systemd to restart.

But really, you can just put `databse-upgrade=true` in your config and be careful upgrading!  I trust you @Sjors!

rustyrussell:
Rebased on master, fixed uninitialized var.

rustyrussell:
That was weird... First CI seemed to have the wrong version, then I seemed to push an empty tree.  But I've rebased onto master now...

cdecker:
Rebased on top of `master`, waiting for CI to complete before merging.

ACK https://github.com/ElementsProject/lightning/pull/5550/commits/262dbf97862a56fc15f302b57fea104221e2a475

cdecker:
Seems we have a false positive for memleak (guessing the bool flag?).

rustyrussell:
> Seems we have a false positive for memleak (guessing the bool flag?).

No, it's a valgrind error.  But I *do* initialize it in this commit (I didn't initialize it previously), and indeed, I checked it's correct locally...

Sjors:
> But really, you can just put `databse-upgrade=true` in your config and be careful upgrading! I trust you @Sjors!

Ok fine...

Scherm­afbeelding 2022-12-02 om 19 04 05

PR #5556, opened 2022-08-28
Fix mrkd

niftynei:
Alternate proposal to #5445.

What if instead of removing it, we bring `mrkd` "in-house" and update it to use the newest/latest and greatest version of `mistune`?

Not really sure what the policy is on updating the `poetry.lock` so did it in two chunks.

Seems to work for almost everything, with the exception that I had to update my local `mistune` install to use 2.0.4; the reproducible builds are getting stuck in the libwally autoconfigure step. Need to debug those.

### DIFF
--- a/.github/workflows/bsd.yml
+++ b/.github/workflows/bsd.yml
@@ -63,7 +63,7 @@ jobs:
             blinker \
             flake8 \
             mako \
-            mrkd \
+            mistune \
             pytest-sentry \
             pytest-test-groups==1.0.3 \
             pytest-custom-exit-code==0.3.0 \
--- a/contrib/docker/Dockerfile.alpine
+++ b/contrib/docker/Dockerfile.alpine
@@ -15,7 +15,7 @@ COPY . lightning
 RUN cd lightning && \
     git submodule update --init --recursive && \
     ./configure && \
-    pip3 install mrkd mistune==0.8.4 && \
+    pip3 install mistune && \
     make -j$(nproc) && \
     make install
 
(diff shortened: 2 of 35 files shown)


cdecker:
Wait, `mrkd` is just one file? 😂

ACK https://github.com/ElementsProject/lightning/pull/5556/commits/4aee424fb8c30a0478623ccc939e923ff971581c

rustyrussell:
But also consider the workarounds we've already needed, and the fact that lowdown seems to actually have maintainers and a user base...

cdecker:
I'd definitely prefer not adding it as something we need to maintain, I was assuming this is mostly to get v0.12.0 out the door, and then go with `lowdown` longer term.

niftynei:
Hm yeah happy to move over to `lowdown` instead (no python, great!). This PR really just moves our dependency up the chain from `mrkd` to `mistune` anyway. (Though it looks very much as though this library won't be re-written any time soon).

Thought this note from `lowdown`'s docs was topical.

> You may be tempted to write [manpages](https://man.openbsd.org/) in Markdown, but please don’t: use [mdoc(7)](https://man.openbsd.org/mdoc), instead — it’s built for that purpose! The man output is for technical documentation only (section 7). [↩](https://kristaps.bsd.lv/lowdown/#fnref1)

rustyrussell:
"Wait, did the lowdown man page just tell me to go fuck myself?"

"Yes. Yes it did."

rustyrussell:
You're still not allowed to adopt a new puppy though!

rustyrussell:
We're close to writing our own documentation generator, but we are not there yet. There's no great solution for producing documentation that fits easily in both web and nroff formats, AFAICT, though.

wtogami:
How large is the mistune code? If small can we embed that alongside to make build more self contained? Otherwise we may run into more problems when mistune upgrades yet again in an incompatible manner.

Regarding lowdown, note in the lowdown PR it has a number of formatting problems if you inspect the before and after. It needs a bit of work upstream to be ready for us.

https://github.com/ElementsProject/lightning/pull/5445#issuecomment-1194873909
Had anyone inspected this the output of old mrkd vs new mrkd with diff and visual inspection?

PR #5568, opened 2022-09-02
move the channels in listpeers to listpeerchannels

vincenzopalazzo:
Now that the release v0.12.x is out I think it is the moment to resurrect https://github.com/ElementsProject/lightning/pull/5022 and finish my work there!

Fixes https://github.com/ElementsProject/lightning/issues/4729
Override https://github.com/ElementsProject/lightning/pull/5022

### Todos

- [x] Add deprecation in the json schema instead of removing the channel from listpeers
- [x] Fixes some nasty test hide inside the CI
- [x] Make sure that all the pay logic is sane after the refactoring


plugins/bkpr/bookkeeper.c
```diff
@@ -916,14 +900,14 @@ listpeers_multi_done(struct command *cmd,
...
 	plugin_log(cmd->plugin, LOG_DBG, "Snapshot balances updated");
 	return notification_handled(cmd);
 }
 
 static char *do_account_close_checks(const tal_t *ctx,
 				     struct chain_event *e,
 				     struct account *acct)
+
 {
 	struct account *closed_acct;
```
rustyrussell, line 912: These two changes are just noise, can be removed.

plugins/bkpr/bookkeeper.c
```diff
@@ -1560,7 +1545,7 @@ parse_and_log_chain_move(struct command *cmd,
 
 		plugin_log(cmd->plugin, LOG_DBG,
 			   "channel event received but no open for channel %s."
-			   " Calling `listpeers` to fetch missing info",
+			   " Calling `listpeerchannls` to fetch missing info",
```
rustyrussell, line 1548: typo

plugins/libplugin.h
```diff
@@ -402,6 +402,8 @@ struct listpeers_channel {
 	int *direction;
 	struct amount_msat total_msat;
 	struct amount_msat spendable_msat;
+	bool peer_connected;
+	struct node_id *peer_id;
```
rustyrussell, line 406: I don't think peer_id needs to be a pointer?

plugins/topology.c
```diff
@@ -271,8 +271,7 @@ static void json_add_halfchan(struct json_stream *response,
 		json_add_num(response, "message_flags", message_flags);
 		json_add_num(response, "channel_flags", channel_flags);
 
-		json_add_bool(response, "active",
-			      c->half[dir].enabled && !local_disable);
+		json_add_bool(response, "active", c->half[dir].enabled && !local_disable);
 		json_add_num(response, "last_update", timestamp);
```
rustyrussell, line 275: Gratuitous reformat?

plugins/topology.c
```diff
@@ -293,32 +292,31 @@ struct listchannels_opts {
...
 
+
 /* We record which local channels are valid; we could record which are
  * invalid, but our testsuite has some weirdness where it has local
  * channels in the store it knows nothing about. */
 static struct node_map *local_connected(const tal_t *ctx,
 					const char *buf,
-					const jsmntok_t *result)
+				        const jsmntok_t *result)
 {
```
rustyrussell, line 302: Gratuitous reformat?

plugins/topology.c
```diff
@@ -383,8 +375,7 @@ static struct command_result *listpeers_done(struct command *cmd,
 			for (size_t i = 0; i < dst->num_chans; i++) {
 				int dir;
 				c = gossmap_nth_chan(gossmap, dst, i, &dir);
-				json_add_halfchan(js, gossmap, connected,
-						  c, 1 << !dir);
+				json_add_halfchan(js, gossmap, connected, c, 1 << !dir);
 			}
```
rustyrussell, line 379: Also doesn't need reformatting?

tests/test_closing.py
```diff
@@ -1192,7 +1196,8 @@ def test_penalty_htlc_tx_fulfill(node_factory, bitcoind, chainparams):
 
     # push some money so that 1 + 4 can both send htlcs
     inv = l2.rpc.invoice(10**9 // 2, '1', 'balancer')
-    l1.rpc.pay(inv['bolt11'])
+    res = l1.rpc.pay(inv['bolt11'])
+    logging.info(f"Result pay command {res}")
```
rustyrussell, line 1200: This seems gratuitous, along with "import logging"?

tests/test_connection.py
```diff
@@ -1682,8 +1682,11 @@ def test_funding_external_wallet(node_factory, bitcoind):
     # Peer should still be connected and in state waiting for funding_txid
     assert peer['id'] == l2.info['id']
     r = re.compile('Funding channel start: awaiting funding_txid with output to .*')
-    assert any(r.match(line) for line in peer['channels'][0]['status'])
-    assert 'OPENINGD' in peer['channels'][0]['state']
+
+    channels = l1.rpc.listpeerchannels(peer['id'])['channels']
+    assert len(channels) == 1, f"Channels for peer {peer['id']} need to be not empty"
```
rustyrussell, line 1687: Wow, I did not know about two-argument assert!!  Thanks! :heart:

tests/test_connection.py
```diff
@@ -4263,10 +4265,22 @@ def test_no_reconnect_awating_unilateral(node_factory, bitcoind):
...
+def test_peer_disconnected_reflected_in_channel_state(node_factory):
+    """
+    Make sure that if a node is disconnected we have the value correct value
+    across listpeer and listpeerchannels.
+    """
+    l1, l2 = node_factory.line_graph(2, opts={'may_reconnect': True})
+    l2.stop()
+
+    wait_for(lambda: only_one(l1.rpc.listpeers(l2.info['id'])['peers'])['connected'] is False)
+    wait_for(lambda: only_one(l1.rpc.listpeerchannels(l2.info['id'])['channels'])['peer_connected'] is False)
```
rustyrussell, line 4286: This last line can be assert, since we should not have to wait?

tests/utils.py
```diff
@@ -438,3 +439,19 @@ def scriptpubkey_addr(scriptpubkey):
...
+    """
+    try:
+        return rpc_pay_call()
+    except RpcError as rpc_err:
+        # Enable the trace with --log-cli-level=INFO
+        # and void to put other info inside the normal console log.
+        logging.info(node.rpc.listfunds())
+        logging.info(node.rpc.paystatus())
+        logging.info(node.rpc.listpeerchannels())
+        raise rpc_err
```
rustyrussell, line 457: If you run pytest as we do in GH CI, as "TEST_DEBUG=1 pytest -p no:logging" you get all the rpc calls and their results already!

doc/schemas/listpeers.schema.json
```diff
@@ -165,1002 +164,1025 @@
...
+                  "type": "hex",
+                  "description": "bitmap of BOLT #9 features from peer's INIT message"
+                }
+              }
+            }
+          }
+        ]
+      },
+      "channels": {
+        "deprecated": "true",
```
rustyrussell, line 217: Ah, we don't normally put deprecated options in the schema at all (if we want to test them, we disable schema checking).

OK that you did though!

rustyrussell (commented):
Changelog lines need some tweaking:

```
Changelog-Added: JSON-RPC: new command `listpeerchannels` now contains information on direct channels with our peers.
Changelog-Deprecated: JSON-RPC: `listpeers` `channels` array: use `listpeerchannels`
```

Some minor comments: generally this is great!
I would probably (not now!) have split this into commits like so:
1. Add the listpeerchannels command, add schema and docs and GRPC bindings.
2. Switch the plugins over to use it.
3. Switch the tests over to use this.
4. Deprecate channels in listpeers, update schema, docs and GRPC bindings again.

Thanks for all this!

plugins/libplugin-pay.c
```diff
@@ -2310,76 +2310,69 @@ REGISTER_PAYMENT_MODIFIER(retry, struct retry_mod_data *, retry_data_init,
...
+	    *dir, *peer_connected, *max_htlc, *htlcs, *state, *alias, *alias_local;
+	size_t j;
+        channels = json_get_member(buffer, toks, "channels");
+	assert(channels);
 
-		connected = json_get_member(buffer, peer, "connected");
+	json_for_each_arr(j, channel, channels) {
+		/* FIXME: we can skip the channel if the peer is not connected in this case? */
+		peer_connected = json_get_member(buffer, channel, "peer_connected");
+		assert(peer_connected);
```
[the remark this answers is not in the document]
vincenzopalazzo: Yesh, this was just a sanity check while I was debugging

plugins/topology.c
vincenzopalazzo: I learned! never active autoformatting on save inside emacs :)

doc/schemas/listpeers.schema.json
vincenzopalazzo: Yeah, but this was requested here https://github.com/ElementsProject/lightning/pull/5022#discussion_r949215155

We should add some policy regarding the auto generated files?

tests/utils.py
vincenzopalazzo: Oh right, this was a hack to trace the pay command, with the logging lib we can have the log print in the terminal and it should be easy to trace in the terminal.

Is there a better way to implement this trace function? I think this will be very useful during the developing lifecycle but I hope we do not need inside the CI when the test are stable

vincenzopalazzo:
>I would probably (not now!) have split this into commits like so:

Yes I'm not happy at all with my committed organization, I refactored them several time next time I will try to be more organized on it

vincenzopalazzo:
Ok, I spend some time on this PR in the last few days and I do not find any direct cause of the refactoring that cause the error CI error.

However, from I think that the problem is from this piece of code

https://github.com/ElementsProject/lightning/blob/master/plugins/bkpr/bookkeeper.c#L1661-L1678

But I'm not sure about that because I do not have to much experiences in the bookkeeper code, but I make a nice log filtering that shows that the log is missing a `coin_move 2 (invoice) 222msat -0msat channel_mvt 1668546274` event https://hackmd.io/@cln/SkOPjtWXo

vincenzopalazzo:
Ok I stole https://github.com/ElementsProject/lightning/pull/5568/commits/33e92e87c1672a5fce3a3b634215d4bf7d7d6ccf from @rustyrussell  to just give a fresh run inside the CI see if all it is ok with this fix.

vincenzopalazzo:
closing in favor of https://github.com/ElementsProject/lightning/pull/5825

Issue #5572, opened 2022-09-05
crash in connectd due to bad gossip_store


rustyrussell:
We hit an assert as we try to send a giant status message (> 128MB), but the reason we're sending it is because a gossip_store entry fails checksum.  I accessed the gossip_store shortly after this, and it seemed fine.

The most likely explanation is that connectd somehow got a bogus offset, and was reading junk from the store.

```
lightning_connectd: FATAL SIGNAL (version v0.12.0)
0x414a56 send_backtrace
        common/daemon.c:33
0x41d10a status_failed
        common/status.c:221
0x41d1fc status_backtrace_exit
        common/subdaemon.c:18
0x414ada crashdump
        common/daemon.c:49
0x7f2a8063c31f ???
        ???:0
0x7f2a8063c2a2 ???
        ???:0
0x7f2a806258a3 ???
        ???:0
0x7f2a80625788 ???
        ???:0
0x7f2a80634a15 ???
        ???:0
0x42d90d wire_sync_write
        wire/wire_sync.c:13
0x414d75 daemon_conn_sync_flush
        common/daemon_conn.c:114
0x41cb31 flush_and_exit
        common/status.c:189
0x41d01a status_send_fatal
        common/status.c:206
0x41d102 status_failed
        common/status.c:223
0x416f16 gossip_store_next
        common/gossip_store.c:156
0x40d138 maybe_from_gossip_store
        connectd/multiplex.c:530
0x40d3fe write_to_peer
        connectd/multiplex.c:966
0x4aee10 next_plan
        ccan/ccan/io/io.c:59
0x4af210 io_do_always
        ccan/ccan/io/io.c:435
0x4aff02 handle_always
        ccan/ccan/io/poll.c:304
0x4b0229 io_loop
        ccan/ccan/io/poll.c:385
0x40b1db main
        connectd/connectd.c:2063
0x7f2a80626b74 ???
        ???:0
0x40481d ???
        ???:0
0xffffffffffffffff ???
        ???:0
lightning_connectd: wire/wire_sync.c:13: wire_sync_write: Assertion `tal_bytelen(msg) < WIRE_LEN_LIMIT' failed.
Lost connection to the RPC socket.Lost connection to the RPC socket.
```

Issue #5570, opened 2022-09-04
`mako` dep broken due module on found


vincenzopalazzo:
I noted this on CI relate to lnprototest, I open the issue just to keep track of this error, and also I make sure that the error is node generate by the `python-mako` ubuntu dep

```
wiregen hsmd/hsmd_wiregen.h
cc ccan/ccan/cdump/tools/cdump-enumstr.c
cc ccan/ccan/cdump/cdump.c
Traceback (most recent call last):
  File "tools/generate-wire.py", line 27, in <module>
    from mako.template import Template
ModuleNotFoundError: No module named 'mako'
make: *** [Makefile:337: hsmd/hsmd_wiregen.h] Error 1
make: *** Waiting for unfinished jobs....
wiregen hsmd/hsmd_wiregen.h
Traceback (most recent call last):
  File "tools/generate-wire.py", line 27, in <module>
    from mako.template import Template
ModuleNotFoundError: No module named 'mako'
make: *** [Makefile:337: hsmd/hsmd_wiregen.h] Error 1
```

I can take a look into it later in the week

niftynei:
caused by a poetry install failure, think we've fixed this now?

vincenzopalazzo:
Yeah this was before the fix, unfortunately, I had not enough permission to manage labels or milestones, but I think this PR could be also take in consideration to fix poetry instead or remove it https://github.com/ElementsProject/lightning/pull/5606

PR #5573, opened 2022-09-05
connectd: better diagnostics on invalid gossip_store entries.

rustyrussell:
Should help diagnose https://github.com/ElementsProject/lightning/issues/5572
which hit the invalid csum on a >64MB entry, if it happens again.

Changelog-None

### DIFF
--- a/common/gossip_store.c
+++ b/common/gossip_store.c
@@ -115,6 +115,7 @@ u8 *gossip_store_next(const tal_t *ctx,
 		      size_t *off, size_t *end)
 {
 	u8 *msg = NULL;
+	size_t initial_off = *off;
 
 	while (!msg) {
 		struct gossip_hdr hdr;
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK [c90d798](https://github.com/ElementsProject/lightning/pull/5573/commits/c90d79870c3dfe6bdc03f8710daab1125cca415a)

CI failure relate to https://github.com/ElementsProject/lightning/issues/5570

endothermicdev (approved):
I guess this will tell us if it's purely an offset issue as opposed to something like a bad length / length mask.  We may also want to validate the message length/CRC before returning the `find_gossip_store_by_timestamp` offset as `gossip_store_next` traversal was a little more thorough.  Maybe that's a another PR.
ACK c90d798

Issue #5563, opened 2022-08-31
Problems with plugins in 0.12.0


unsettledroell404:
recently installed v0.12.0 has issues with plugins.
Summary plugin works, feeadjuster does not: when I run 'lightning-cli feeadjust' it displays a blank line. I also tried balance, but it does not work either.

I am looking in the logs and found some information that may be useful. Maybe there is an issue with pyln client (v.0.12.0)?
Summary does actually work but shows the same error namely "TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType".

2022-08-31T17:17:40.277Z DEBUG   plugin-summary.py: Reopened summary.dat shelve with 0 runs and 6 entries
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py: Traceback (most recent call last):
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:   File \"/usr/local/lib/python3.10/dist-packages/pyln/client/plugin.py\", line 631, in _dispatch_request
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:     result = self._exec_func(method.func, request)
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:   File \"/usr/local/lib/python3.10/dist-packages/pyln/client/plugin.py\", line 616, in _exec_func
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:     return func(*ba.args, **ba.kwargs)
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:   File \"/usr/local/lib/python3.10/dist-packages/pyln/client/plugin.py\", line 957, in _init
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:     return self._exec_func(self.child_init, request)
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:   File \"/usr/local/lib/python3.10/dist-packages/pyln/client/plugin.py\", line 616, in _exec_func
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:     return func(*ba.args, **ba.kwargs)
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:   File \"/data/cln/plugins-available/feeadjuster/feeadjuster.py\", line 298, in init
2022-08-31T17:17:40.277Z INFO    plugin-feeadjuster.py:     config = plugin.rpc.listconfigs()
[...]

PR #5574, opened 2022-09-06
pyln-client: allow 'msat' fields to be 'null'

rustyrussell:
This happens with deprecated-apis and listconfigs, breaking some
python plugins!

Fixes: #5546
Fixes: #5563


jonasnick (commented):
This PR makes the [nix-bitcoin](https://github.com/fort-nix/nix-bitcoin) test suite pass.

Without it, the test suite fails, because the `summary` and `feeadjuster` plugins crash on startup. The test suite is not very in-depth, it just checks that the plugins are loaded and that certain commands don't fail.

vincenzopalazzo (approved):
ACK [ebbf22e](https://github.com/ElementsProject/lightning/pull/5574/commits/ebbf22e97372fcc41558491c8a36707ddc446000)

Issue #5363, opened 2022-06-30
mkfunding tool appears to be broken


instagibbs:
on master, with the example being taken directly from the source file:
```
/devtools/mkfunding 16835ac8c154b616baac524163f41fb0c4f82c7b972ad35d4d6f18d854f6856b 1 0.01btc 253 76edf0c303b9e692da9cb491abedef46ca5b81d32f102eb4648461b239cb0f99 0000000000000000000000000000000000000000000000000000000000000010 0000000000000000000000000000000000000000000000000000000000000020
Aborted (core dumped)
```

For a fix would be nice if all these tools had their example vectors extracted from the source and added to a simple sanity checking test that would at least catch segfaults/crashes/error conditions.

PR #5580, opened 2022-09-08
Issue 5363 mkfunding unit tests

greggzigler:
Issue https://github.com/ElementsProject/lightning/issues/5363 suggests that we create some sanity checking tests for the mkfunding devtool. In this draft pull request, I copied the parameters from the top of mkfunding.c and used them as input into three "unit tests".

A couple obvious problems so far:
- I mocked-out most of the interesting bitcoin/*.c functions that the devtool uses, just to get the three tests to pass; so it doesn't do any interesting work. So far, "pass" simply means it didn't abort after all the mocked functions returned successfully;
- I don't know what use cases these three sets of parameters actually represent; I can't put a meaningful label on any of the test cases.

I'd like to create tests for specific use cases, including segfaults and crashes and error conditions as suggested in the issue. I'd be open to suggestions.

Is this PR headed in the right direction?


niftynei:
Whoa this is a great idea. I think the easiest thing to do would be more "blackbox", so having a shell function that calls `mkfunding` with some parameters and verifies the output.

You could probably do these using the python test framework? SImilar to the ones we have for the `hsmtool`.

instagibbs:
agree with more blackbox approach, it has served me well in the past

greggzigler:
Issue 5363 has been fixed, so closing this pull request

PR #5582, opened 2022-09-09
Don't report redundant feerates to subdaemons

instagibbs:
Currently channeld will get announcements of feerates even if the prior *modified* feerate was the one already reported prior. F.e., the floor is 253, but grabbed rate from bitcoind was 250, but we report it again based on 250(which is never stored!), not 253.

Only report if historical data didn't exist(which is backfilled for smoothing), or if the final computed feerate doesn't match historical.

Changelog-None

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -338,7 +338,7 @@ static void update_feerates(struct bitcoind *bitcoind,
 	 * 2 minutes. The following will do that in a polling interval
 	 * independent manner. */
 	double alpha = 1 - pow(0.1,(double)topo->poll_seconds / 120);
-	bool feerate_changed = false;
+	bool notify_feerate_changed = false;
 
 	for (size_t i = 0; i < NUM_FEERATES; i++) {
 		u32 feerate = satoshi_per_kw[i];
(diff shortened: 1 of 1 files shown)


lightningd/chaintopology.c
```diff
@@ -353,16 +353,14 @@ static void update_feerates(struct bitcoind *bitcoind,
 
 		/* Initial smoothed feerate is the polled feerate */
 		if (!old_feerates[i]) {
-			feerate_changed = true;
+            notify_feerate_changed = true;
```
vincenzopalazzo, line 356:
```suggestion
                       notify_feerate_changed = true;
```

vincenzopalazzo (commented):
Just a TAB missing for me

cdecker (approved):
ACK 3903be15969a81bd3969b85e09bef89c3051149f

cdecker:
Rebased on top of `master` in the hopes that the `mako` issue in lnprototest is fixed.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5582/commits/7a518e2a950587d864239e97b42ee5e4cfd54356

Issue #5557, opened 2022-08-30
[bug]: lease_fee double counted, causes bkpr-listbalances to crash lightningd


chrisguida:
### Issue and Steps to Reproduce
1) Open private dual-funded channel using liquidity ads page on RTL 0.13.0 (this bug was fixed, in 0.13.1 RTL opens public channels, not sure if relevant)
2) Close channel
3) run `bkpr-listbalances`
4) lightningd crashes with this message:

```
2022-08-30T00:06:43.169777Z 2022-08-30T00:06:43.162Z **BROKEN** plugin-bookkeeper: Get account balance returned err for account 58c817f67224a80373ad7f1eafc4c069b6ea41986a2c9404bd531a47f1997e1c: bc channel balance is negative? 23268000msat - 26536000msat
2022-08-30T00:06:43.343931Z 2022-08-30T00:06:43.339Z INFO    plugin-bookkeeper: Killing plugin: exited during normal operation
2022-08-30T00:06:43.350405Z 2022-08-30T00:06:43.339Z **BROKEN** plugin-bookkeeper: Plugin marked as important, shutting down lightningd!
```

### `getinfo` output
```
# lightning-cli getinfo
{
   "num_peers": 26,
   "num_pending_channels": 0,
   "num_active_channels": 21,
   "num_inactive_channels": 0,
   "address": [
      {
         "type": "torv3",
         "port": 9735
      },
      {
         "type": "ipv4",
      }
   ],
   "binding": [
      {
         "type": "ipv4",
         "address": "0.0.0.0",
         "port": 9735
      }
   ],
   "version": "v0.12.0",
   "blockheight": 751861,
   "network": "bitcoin",
   "lightning-dir": "/root/.lightning/bitcoin",
   "our_features": {
      "init": "0200000000000008a080282269a2",
      "node": "0200000000000088a080282269a2",
      "channel": "",
      "invoice": "02000020024100"
   }
}
```

`bkpr-listaccountevents 58c817f67224a80373ad7f1eafc4c069b6ea41986a2c9404bd531a47f1997e1c` shows that the least fee was counted twice.

deleting the duplicate lease_fee entry from the db fixes the issue.

cc @niftynei

niftynei:
Ok got a test case that reproduces this, thinking about a fix. Suspect that you had initiated the channel open and then upgraded to v12 very soon after, but before the channel had opened.

[...]

PR #5587, opened 2022-09-13
Fix for  #5557

niftynei:
Crash found in `bookkeeper` around duplicate lease_fees! Here we fix them.


rustyrussell (approved):
Ack 3e903e63f016a618b30b354263223a6b0e030550

PR #5589, opened 2022-09-13
what if we just didn't use poetry install

niftynei:
What if we use pip install for our builds? Just a thought.

Changelog-None

### DIFF
--- a/contrib/docker/scripts/build.sh
+++ b/contrib/docker/scripts/build.sh
@@ -23,8 +23,8 @@ export LIGHTNINGD_POSTGRES_NO_VACUUM=1
 
 pip3 install --upgrade pip
 pip3 install --user poetry
-poetry config virtualenvs.create false --local
-poetry install
+poetry export --dev --without-hashes -o requirements.txt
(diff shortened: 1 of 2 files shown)


wtogami:
yes please

cdecker:
That should be much less of a problem, if we generate the `requirements.txt` from `pyproject.toml` it will specify exact versions, cutting down on the combinatorial issue that we were having before on CI.

So poetry for dependency definition and resolution, and pip + requirements.txt to avoid having to configure poetry and others. This sounds like a good idea, and we could go one step further and check the `requirements.txt` file in, and treat it as a lock file that if generated and can be checked for changes in CI

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5589/commits/ae2bedfb097b753c2479ec9572e90eb2754f6b7d

rustyrussell (approved):
Ack ae2bedfb097b753c2479ec9572e90eb2754f6b7d

poetry is for devs, pip is for users.

Issue #5565, opened 2022-09-01
**BROKEN** connectd: gossip_store: bad checksum at offset 218892226:


grubles:
CLN v0.12 release
Rocky Linux 9 
ppc64le

Plugins: summary, CLBOSS, peerswap

CLN crashed in the middle of the night so I don't have any steps to reproduce unfortunately. 

stdout log:
```
0x1001a657 send_backtrace
        common/daemon.c:33
0x10025e7f status_failed
        common/status.c:221
0x1001e43f gossip_store_next
        common/gossip_store.c:156
0x1000f9d3 maybe_from_gossip_store
        connectd/multiplex.c:530
0x1000fdff write_to_peer
        connectd/multiplex.c:966
0x100d322b next_plan
        ccan/ccan/io/io.c:59
0x100d3a1b io_do_always
        ccan/ccan/io/io.c:435
0x100d51ef handle_always
        ccan/ccan/io/poll.c:304
0x100d5853 io_loop
        ccan/ccan/io/poll.c:385
0x1000bfe7 main
        connectd/connectd.c:2063
0x7fff9e777ca3 ???
        ???:0
0x7fff9e777e8f ???
        ???:0
0xffffffffffffffff ???
        ???:
Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.Lost connection to the RPC socket.terminate called after throwing an instance of 'std::runtime_error'
```

CLN log file backtrace:

```
02022-09-01T10:40:19.304Z **BROKEN** lightningd: FATAL SIGNAL 11 (version v0.12.0)
2022-09-01T10:40:19.305Z **BROKEN** lightningd: backtrace: common/daemon.c:38 (send_backtrace) 0x100885df
2022-09-01T10:40:19.305Z **BROKEN** lightningd: backtrace: common/daemon.c:46 (crashdump) 0x10088657
2022-09-01T10:40:19.305Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7fffa2200463
2022-09-01T10:40:19.305Z **BROKEN** lightningd: backtrace: (null):0 ((null)) 0x7fffa1c6c210
2022-09-01T10:40:19.305Z **BROKEN** lightningd: backtrace: ccan/ccan/json_out/json_out.c:297 (json_out_addstr) 0x1016fca3
[...]

PR #5591, opened 2022-09-14
Fix for #5572 and #5565

endothermicdev:
This took more effort to debug than it appears, but I believe this is ultimately the source of the gossip_store offset error!  Some of the previously reported node announcement propagation issues may have originated here as well.  Accounting for the gossip_store's leading version byte is a simple fix.


rustyrussell (approved):
Ack 6af04d10bf4ff90cd9bee35ff96e9e73195da73f

Changelog-Fixed should also note that it fixes a crash!

Issue #5578, opened 2022-09-08
crashed by an assertion in common/gossmap.c in v0.12.0


heyambob:
### Issue and Steps to Reproduce

So far it happened once. A little bit after restarted while I had some problem with tor, unsure if related.

```
topology: common/gossmap.c:419: add_channel: Assertion `!private' failed.
topology: FATAL SIGNAL 6 (version v0.12.0)
0x561f49f56244 send_backtrace
	common/daemon.c:33
0x561f49f562ec crashdump
	common/daemon.c:46
0x7f571843e8df ???
	???:0
0x7f571848e36c ???
	???:0
0x7f571843e837 ???
	???:0
0x7f5718428534 ???
	???:0
0x7f571842845b ???
	???:0
0x7f5718437365 ???
	???:0
0x561f49f2e72d add_channel
	common/gossmap.c:419
0x561f49f2ef3d map_catchup
	common/gossmap.c:618
0x561f49f3002f gossmap_refresh
	common/gossmap.c:938
0x561f49f3b5fa get_gossmap
	plugins/topology.c:27
0x561f49f3d389 json_listnodes
	plugins/topology.c:525
0x561f49f42245 ld_command_handle
	plugins/libplugin.c:1452
0x561f49f42398 ld_read_json_one
	plugins/libplugin.c:1493
0x561f49f424ca ld_read_json
	plugins/libplugin.c:1513
0x561f49fdea0b next_plan
	ccan/ccan/io/io.c:59
0x561f49fdf5d6 do_plan
	ccan/ccan/io/io.c:407
0x561f49fdf614 io_ready
	ccan/ccan/io/io.c:417
0x561f49fe18de io_loop
	ccan/ccan/io/poll.c:453
0x561f49f42f63 plugin_main
	plugins/libplugin.c:1710
0x561f49f3dac2 main
	plugins/topology.c:701
0x7f571842928f ???
	???:0
0x7f5718429349 ???
	???:0
0x561f49f2b764 ???
	../sysdeps/x86_64/start.S:115
0xffffffffffffffff ???
	???:0
topology: FATAL SIGNAL 11 (version v0.12.0)
0x561f49f56244 send_backtrace
	common/daemon.c:33
0x561f49f562ec crashdump
	common/daemon.c:46
0x7f571843e8df ???
	???:0
0x0 ???
	???:0
```

### `getinfo` output

"version": "v0.12.0",

PR #5593, opened 2022-09-14
gossmap: don't crash if we see a duplicate channel_announce.

rustyrussell:
Apparently we had two private channel announcements (the !private assert failed).  While this shouldn't happen, don't crash because of it.

Fixes: #5578
Changelog-Fixed: Plugins: topology plugin could crash when it sees duplicate private channel announcements.

### DIFF
--- a/common/gossmap.c
+++ b/common/gossmap.c
@@ -414,11 +414,8 @@ static struct gossmap_chan *add_channel(struct gossmap *map,
 	 * that's the only time we get duplicates */
 	scid.u64 = map_be64(map, cannounce_off + plus_scid_off);
 	chan = gossmap_find_chan(map, &scid);
-	if (chan) {
-		assert(chan->private);
-		assert(!private);
+	if (chan)
 		gossmap_remove_chan(map, chan);
-	}
 
 	/* We carefully map pointers to indexes, since new_node can move them! */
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5593/commits/ccede9df217af22ff3d71beee5206dea936f4a26

vincenzopalazzo (approved):
ACK [ccede9d](https://github.com/ElementsProject/lightning/pull/5593/commits/ccede9df217af22ff3d71beee5206dea936f4a26)

cdecker:
Leaving open for now since it needs to be cherry-picked into v0.12.1

common/gossmap.c
```diff
@@ -414,11 +414,8 @@ static struct gossmap_chan *add_channel(struct gossmap *map,
 	 * that's the only time we get duplicates */
 	scid.u64 = map_be64(map, cannounce_off + plus_scid_off);
 	chan = gossmap_find_chan(map, &scid);
-	if (chan) {
-		assert(chan->private);
-		assert(!private);
```
endothermicdev, line 419: Do we know the cause of the duplicate private channel announcement? Maybe log this case with status_broken for visibility.

endothermicdev (approved):
ACK ccede9d

PR #5594, opened 2022-09-14
autoclean gains SUPER POWERS (and other optimizations)

rustyrussell:
autoclean is generalized to delete old listforwards/listpays/listinvoices (not just expired invoices).  And much optimization is done, so that it takes only 3-6 minutes to delete 1M entries.

This should help with larger nodes, in particular.

There are some new APIs in here!  listhtlcs and delforward in particular...


rustyrussell:
Rebased....

rustyrussell:
Fixed up rust generated stuff, and also updated delpay docs (and added missing request schema!).

lightningd/pay.c
```diff
@@ -37,15 +37,16 @@ struct sendpay_command {
 	struct command *cmd;
 };
 
-static bool string_to_payment_status(const char *status_str, enum wallet_payment_status *status)
+static bool string_to_payment_status(const char *status_str, size_t len,
+				     enum wallet_payment_status *status)
 {
-	if (streq(status_str, "complete")) {
```
cdecker, line 42: Just out of curiosity: how much performance difference does this make to the comparison? I'd have expected this to be pretty close, maybe a bit less branchy due to the predictable length in `memeqstr`

plugins/autoclean.c
```diff
@@ -75,58 +84,103 @@ static const char *datastore_path(const tal_t *ctx,
...
 					     datastore_path(tmpctx, i, "num"),
-					     tal_fmt(tmpctx, "%"PRIu64, num_cleaned[i]),
+					     tal_fmt(tmpctx, "%"PRIu64, total_cleaned[i]),
 					     "create-or-replace", NULL, NULL, NULL);
 
 	}
 
-	return set_next_timer(plugin);
+	/* autoclean-once? */
+	if (cinfo->cmd) {
```
cdecker, line 104: This is safe for races? I.e., if the user closes the connection with the command in flight, `cinfo->cmd` doesn't get reset, correct? Otherwise we might end up starting multiple timer loops.

doc/lightning-batching.7.md
```diff
@@ -0,0 +1,54 @@
+lightning-batching -- Command to allow database batching.
```
cdecker, line 1: How is this safe? There are a number of places where we assume that the DB transaction has been unrevocably been committed to the DB, i.e., HTLC changes. If we permit a DB transaction to linger for longer than the synchronous JSON RPC function, then we cannot guarantee that anymore, or if so how? My fear is that if we start a batch at an inopportune time, then commit an update to a channel, leaving the DB tx dangling, and then kill the daemon, we will have just forgotten about a commit that we should absolutely not forget.

cdecker: Notice that this would not be such an issue if we had support for parallel transactions, where we could have a transaction for RPC and have that distinct from the state machine transaction, but alas sqlite3 does not allow us to do that.

cdecker (changes requested):
I have some doubts regarding the RPC batching support, which I don't see how it can work without jeopardizing correctness.

wallet/wallet.c
```diff
@@ -4622,13 +4622,8 @@ const struct forwarding *wallet_forwarded_payments_get(struct wallet *w,
...
 		}
 
-		if (!db_col_is_null(stmt, "payment_hash")) {
-			cur->payment_hash = tal(ctx, struct sha256);
-			db_col_sha256(stmt, "payment_hash", cur->payment_hash);
-		} else {
-			cur->payment_hash = NULL;
-		}
-
+		/* FIXME: This now requires complex join to determine! */
```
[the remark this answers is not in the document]
rustyrussell: No, because you'd need to match scid too. "id" used to be the db id, it's now the htlc I'd.

wallet/wallet.c
```diff
@@ -3528,47 +3527,58 @@ wallet_payment_list(const tal_t *ctx,
...
+						     " ORDER BY id;"));
 		db_bind_sha256(stmt, 0, payment_hash);
-	else
-		db_bind_null(stmt, 0);
-	db_bind_int(stmt, 1, enable_payment_hash);
-	if (status)
-		db_bind_int(stmt, 2, wallet_payment_status_in_db(*status));
-	else
-		db_bind_null(stmt, 2);
-	db_bind_int(stmt, 3, enable_status);
```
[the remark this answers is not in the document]
rustyrussell: Yes, I got SQLite to explain the statements, and it didn't use the index with this construction :(

Maybe we could convince it to, but I decided naive was good.

plugins/autoclean.c
rustyrussell: Yes, since autoclean-once allocates cinfo dynamically, each command will be a separate one (you might get lots of messages is they try to delete at the same time ofc!)

doc/lightning-batching.7.md
rustyrussell: No, we always close the transaction before blocking. We only batch when we read more than one command at once (which happens *a lot* with autoclean slamming us with deletes).

Enabling batching does not *guarantee* batching: that would be hard to do, indeed!

lightningd/pay.c
rustyrussell: Yes, when we're parsing JSON we don't have a nul-terminated string :(

wallet/db.c
```diff
@@ -1504,3 +1509,40 @@ static void migrate_channels_scids_as_integers(struct lightningd *ld,
...
+			db_col_ignore(stmt, "id");
+			continue;
+		}
+
+		str = db_col_strdup(tmpctx, stmt, "failchannel");
+		if (!short_channel_id_from_str(str, strlen(str), &scid))
+			db_fatal("Cannot convert invalid payments.failchannel '%s'",
+				 str);
+		update_stmt = db_prepare_v2(db, SQL("UPDATE payments SET"
+						    " failscid = ?"
```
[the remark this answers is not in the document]
rustyrussell: I think for the payments table we actually delete the old column. But we should do this for the channels table where we can't, good call...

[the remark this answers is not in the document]
rustyrussell: Done, new commit!

rustyrussell:
Rebased on master, and:

- null-out the now-unused short_channel_id field in channels (use scid!)
- expand documentation of "batching" command:

```
The **batching** RPC command allows (but does not guarantee!) database
commitments to be deferred when multiple commands are issued on this RPC
connection.  This is only useful if many commands are being given at once, in
which case it can offer a performance improvement (the cost being that if
there is a crash, it's unclear how many of the commands will have been
persisted).
```

cdecker (approved):

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/1de8bca51ee8ad2d1bf9a932a0bd589fc4628a78

rustyrussell:
Extended timeouts in test_autoclean: CI is *slow* esp under valgrind!

Ack https://github.com/ElementsProject/lightning/pull/5594/commits/e13433258e8b1185f80214b9fabae168dd3d1f62

rustyrussell:
Trivial rebase for generated code clash (contrib/pyln-testing/pyln/testing/node_pb2.py).

Ack 282c8b2b9

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5594/commits/2ac334d140e52ab19630c1e594c2bc2c9f01d8ea

PR #5598, opened 2022-09-16
gci: Use stable rust instead of nightly

cdecker:
Nightly occasionally breaks, so use stable instead.

Changelog-None


vincenzopalazzo (approved):
ACK [d46cf6d](https://github.com/ElementsProject/lightning/pull/5598/commits/d46cf6d9f3657587c8e6768c50ed96b554e712a8)

I think this is good to land, also if maybe we should report this bug if no one report already it, looks like a rust suggestion bug from the stack-trace

PR #5602, opened 2022-09-17
fix dockerfile build

Evanfeenstra:
Adding `mako` and `mrkd` fixes `docker build` for us

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -108,6 +108,8 @@ RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/inst
     && pip3 install -U wheel \
     && /root/.local/bin/poetry config virtualenvs.create false \
     && /root/.local/bin/poetry install
+RUN pip install Mako
+RUN pip install mistune==0.8.4 mrkd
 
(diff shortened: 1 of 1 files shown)


cdecker:
Good catch, however I think the apropriate place to add these would be `pyproject.toml` and `poetry.lock` so all environments get them?

niftynei:
We removed the `mrkd` and `mistune` dependencies in master, seems wrong to be adding them back?

Dockerfile
```diff
@@ -108,6 +108,8 @@ RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/inst
     && pip3 install -U wheel \
     && /root/.local/bin/poetry config virtualenvs.create false \
```
niftynei, line 109: should we just convert this over to building a requirements.txt and installing w/ pip as well, instead of using `poetry install`?

cdecker: Probably

cdecker:
> We removed the `mrkd` and `mistune` dependencies in master, seems wrong to be adding them back?

Right, in that case we should likely ensure that `mako` and `mrkd` are not needed to build at all, and we need to ensure that all generated files are checked in correctly.

Evanfeenstra:
thanks for looking into it

PR #5604, opened 2022-09-19
devtools: Make fund_nodes compatible w/zsh

ddustin:
Mac is updating to using zsh in general.

The “for i in” reads strings with spaces as a single entry instead of multiple entries as sh did.

Using “while read” … “<<< $var” makes it treat each space as a new entry.

Changelog-None

### DIFF
--- a/contrib/startup_regtest.sh
+++ b/contrib/startup_regtest.sh
@@ -206,7 +206,7 @@ fund_nodes() {
 
 	last_node=""
 
-	for i in $NODES; do
+	echo "$NODES" | while read -r i; do
 
 		if [ -z "$last_node" ]; then
 			last_node=$i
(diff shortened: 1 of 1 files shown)


cdecker:
`shellcheck` seems not to like the changes:

```
In contrib/startup_regtest.sh line 209:
	while read i; do
              ^--^ SC2162: read without -r will mangle backslashes.

make: *** [Makefile:552: check-shellcheck] Error 123

In contrib/startup_regtest.sh line 257:
	done <<< "$NODES"
             ^-^ SC2039: In POSIX sh, here-strings are undefined.

For more information:
  https://www.shellcheck.net/wiki/SC2039 -- In POSIX sh, here-strings are und...
  https://www.shellcheck.net/wiki/SC2162 -- read without -r will mangle backs...
```

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5604/commits/dd76595e91eef82695172caac7e753c58b858f3c

PR #5592, opened 2022-09-14
More BOLT updates, to latest.

rustyrussell:
Includes the 12-block delay, removing non-htlc_maximum_msat updates, and setting the `dont_forward` bit!

Currently breaks lnprototest as expected, I'll have to fix that!

Closes: #5548 
Closes #5562

PR #5605, opened 2022-09-19
lightningd: figure out optimal channel *before* forward_htlc hook, allow override

rustyrussell:
Otherwise what the hook sees is actually a lie, and if it sets it we might override it.

The side effect is that we add an explicit "forward_to" field, and allow hooks to override it.  This lets a *hook* control channel choice explicitly.

Changelod-Added: Plugins: `htlc_accepted_hook` return can specify what channel to forward htlc to.

(Interestingly, this fell out of the blinded payments work, where we can be given the peer, rather than the channel, to forward to).

### DIFF
--- a/doc/PLUGINS.md
+++ b/doc/PLUGINS.md
@@ -1432,7 +1432,8 @@ The payload of the hook call has the following format:
     "cltv_expiry": 500028,
     "cltv_expiry_relative": 10,
     "payment_hash": "0000000000000000000000000000000000000000000000000000000000000000"
-  }
+  },
+  "forward_to": "0000000000000000000000000000000000000000000000000000000000000000"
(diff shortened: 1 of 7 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5605/commits/f0073822e77e11c83739e80a19dcd4a4dc7a9b6e

rustyrussell:
Rebased onto #5625 and fixed "Changelod" to "Changelog" typo in commit message!

cdecker:
Most likely requires #5592 to pick up the BOLT changes

rustyrussell:
Trivial rebase to pick up lnprototest fixes...

Ack b266c304c

PR #5607, opened 2022-09-19
pyln: Reduce dependency strictness for pyln-testing

cdecker:
Pointed out by @rustyrussell we were too restrictive in many cases.


rustyrussell (approved):
Ack 86718bbb553a5e2950a6c99f376da8a4a6bd0031


But see #5613 ?

Issue #5609, opened 2022-09-19
TypeError in pyln-client get_json_id()


MiWCryptAnalytics:
I noticed that a py plugin (rebalance) will silently fail to load with the newer py lib.
From debug logging, I can see that it is raising a TypeError on the str cat operation:
TypeError: unsupported operand type(s) for +: 'int' and 'str'

https://github.com/ElementsProject/lightning/blob/248d60d7bd7e9a1961e70a3579ea93a0d496d6cc/contrib/pyln-client/pyln/client/lightning.py#L338

Can I recommend an fstring instead?
eg:  
`this_id = f'{cmdprefix}/{this_id}'`

rustyrussell:
Are you using the master branch of the python lib with the 0.12.0 version of the code?

MiWCryptAnalytics:
Yes, 0.12.0 client via docker image and I copied the py libs from master branch into my plugin volume to provide runtime dependencies

MiWCryptAnalytics:
❤️ Thanks rusty! f-strings are great!

PR #5611, opened 2022-09-19
pyln-client: use f strings to concatenate JSON ids, handle older integer ids

rustyrussell:
We don't really support using pyln-client with older Core Lightning versions, but this is neater anyway.  I checked: f-strings go back to python 3.6, so we can use them (I think this may be the first!).

Suggested-by: @MiWCryptAnalytics
Fixes: #5609
Changelog-None: introduced since 0.12 with chained ids


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5611/commits/da1501d6d752f304c33a204f38c4018fffd89d69

cdecker (approved):
ACK 21a69b16c70defde12dbf06b3e96fc97548e2d5e

PR #5614, opened 2022-09-20
pyln: Publish 0.12.0.post1 to address non-integer ID error

cdecker:
The non-numeric JSON-RPC request IDs are throwing off the plugins based on pyln-client, let's fix that.

Changelog-None


rustyrussell:
1. The non-numeric ids are only in master lightningd, so though nice to have, this isn't critical.
2. If we publish the version in master without #5611 we break every user *not* using master.
3. More importantly, this includes the fix the 0.12.0 where pyln.client can't parse listconfigs in deprecated-apis mode.
4. Aren't we about to ship 0.12.1, shouldn't we just ship this as 0.12.1?

PR #5620, opened 2022-09-22
build-release: configure before submodcheck

niftynei:
We added a submod dep (lowdown) that requires config to run first.

Changelog-None


rustyrussell (approved):
Ack 2b3e59058cea653f70b5a730603101b373375b2c

PR #5621, opened 2022-09-22
pyln spec update

rustyrussell:
This doesn't really belong in the same repo as CLN, but it's where it is.  So I'll publish this as soon as it's acked, to fix lnprototest master which uses the new field names!


cdecker:
Ignoring the lnprototest CI configs, as that's due to the chicken-and-egg nature of this PR :-)

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5621/commits/22906c55e59a656763841ce35dfde7ae572f653b

cdecker (approved):

PR #5623, opened 2022-09-22
pytest: Stabilize `test_pay_disconnect`

cdecker:
It seems like detecting unreachability is flaky after the migration to `connectd` muxing all connections. So now we wait for the logline, and ensure that even when we detached `sendpay` we wait for it, which should fail just the same way.

Changelog-None


rustyrussell:
Oops, I should have checked... I fixed this properly in #5623!

PR #5624, opened 2022-09-22
onchaind/onchaind_wire.c duplicated in ONCHAIND_SRC

instagibbs:


rustyrussell:
OK, I've fixed the two unrelated flakes in PR #5625 so I'll rebase after that is committed...

PR #5626, opened 2022-09-23
pyln: Bump versions to v0.12.1

cdecker:
This is just before the introduction of `get_json_id`, but has the correct dependency constraints such that all packages can be updated to >=v0.12 and we don't mix minor versions.

This addresses a couple of issues:

 - The `pyln-client=^0.11` constraint would end up picking an older version
 - The `PrettyPrintingLightningRpc` published in `pyln-testing` already uses `get_json_id` which is not defined in `LightningRpc`, but intercepted via `__getattr__` call, and interpreted as a JSON-RPC call, which caused a recursive call.

This fixes both of them by publishing an earlier `pyln-testing` and fixing the constraints in `pyln-testing`.

Changelog-None

### DIFF
--- a/contrib/pyln-testing/pyproject.toml
+++ b/contrib/pyln-testing/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pyln-testing"
-version = "0.12.0.post1"
+version = "0.12.1"
 description = "Test your Core Lightning integration, plugins or whatever you want"
 authors = ["Christian Decker <decker.christian@gmail.com>"]
 license = "BSD-MIT"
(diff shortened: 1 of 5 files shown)


cdecker:
Notice that this is a PR against the hotfix branch, not `master`. I'll cherry-pick this one into `master` once we fixed the immediate problem.

cdecker:
In the meantime I yanked the v0.12.0.post1 which introduced the issue.

contrib/pyln-testing/pyproject.toml
```diff
@@ -17,7 +17,7 @@ ephemeral-port-reserve = "^1.1.4"
 psycopg2-binary = "^2.9.3"
 python-bitcoinlib = "^0.11.0"
 jsonschema = "^4.4.0"
-pyln-client = "^0.11"
+pyln-client = ">=0.12.1"
```
cdecker, line 20: This is the minimal version required so we have `LightningRpc.get_json_id`. Bound expressed as lower bound, so we don't have to bump this every single time.

rustyrussell:
Thanks! This would be great to fix...

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5626/commits/8f06415415946e89869f26e9011e6320d2de810e

rustyrussell:
Needs to be rebased onto master now 0.12.1 was already released, then published?

Issue #5636, opened 2022-09-27
Channel reinit fails: "Could not restore HTLCs"; add_htlc(…) returning CHANNEL_ERR_TOO_MANY_HTLCS


whitslack:
### Issue and Steps to Reproduce
I have an open channel that cannot be reinitialized because it currently has more outstanding HTLCs than the configured limit.

```
2022-09-27T01:03:47.116Z **BROKEN** 02…a8-channeld-chan#248802: out HTLC 28629 failed error 7
2022-09-27T01:03:47.355Z **BROKEN** 02…a8-channeld-chan#248802: Could not restore HTLCs (version 0.11.2-gentoo-r105)
2022-09-27T01:03:47.355Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: common/status.c:221 (status_failed) 0x55…5a
2022-09-27T01:03:47.355Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: channeld/channeld.c:3917 (init_channel) 0x55…bc
2022-09-27T01:03:47.355Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: channeld/channeld.c:3992 (main) 0x55…bc
2022-09-27T01:03:47.355Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: ../sysdeps/nptl/libc_start_call_main.h:58 (__libc_start_call_main) 0x7f…89
2022-09-27T01:03:47.356Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: ../csu/libc-start.c:381 (__libc_start_main_impl) 0x7f…44
2022-09-27T01:03:47.356Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: (null):0 ((null)) 0x55…80
2022-09-27T01:03:47.356Z **BROKEN** 02…a8-channeld-chan#248802: backtrace: (null):0 ((null)) 0xff…ff
2022-09-27T01:03:47.356Z **BROKEN** 02…a8-channeld-chan#248802: STATUS_FAIL_INTERNAL_ERROR: Could not restore HTLCs
```

In `channeld/channeld.c` (`init_channel`), we called `status_failed` from here:
```c
	if (!channel_force_htlcs(peer->channel,
			 cast_const2(const struct existing_htlc **, htlcs)))
		status_failed(STATUS_FAIL_INTERNAL_ERROR,
			      "Could not restore HTLCs");
```

[...]

PR #5640, opened 2022-09-28
channeld: do not enforce max_accepted_htlcs on LOCAL in channel reinit

whitslack:
If a channel goes offline while the count of outstanding outgoing HTLCs exceeds the limit that we enforce against the peer, then the channel could never be brought online again because `add_htlc` called by `channel_force_htlcs` in `channeld/full_channel.c` would return `CHANNEL_ERR_TOO_MANY_HTLCS`. The protocol specification actually does allow us to exceed the limits that we are enforcing against the peer; we are only prohibited from exceeding the limits that the peer is enforcing against us. `add_htlc` takes an `enforce_aggregate_limits` parameter that appears to have been intended for `channel_force_htlcs` to exempt the local node from obeying the limits that it is enforcing against the peer, but this parameter was only being respected for the total HTLC value-in-flight check but not for the HTLC count check. This commit respects the parameter for the HTLC count check as well and resolves the problem of "Could not restore HTLCs".

Fixes: #5636
Changelog-Fixed: channeld: Channel reinitialization no longer fails when the number of outstanding outgoing HTLCs exceeds `max_accepted_htlcs`.

### DIFF
--- a/channeld/full_channel.c
+++ b/channeld/full_channel.c
@@ -629,7 +629,7 @@ static enum channel_add_err add_htlc(struct channel *channel,
 	 * the channel to pay unnecessary onchain fees during a fee
 	 * spike with large commitment transactions.
 	 */
-	if (sender == LOCAL
+	if (enforce_aggregate_limits && sender == LOCAL
 	    && htlc_count + 1 > channel->config[LOCAL].max_accepted_htlcs) {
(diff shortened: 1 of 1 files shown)


cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/5640/commits/2003a38ca64a11905c42c4a8dd01872dcc99ba68

PR #5645, opened 2022-10-03
keysend: just strip even unknown fields.

rustyrussell:
We've had reports of keyends failing, and indeed:

```
plugin-keysend: Experimental: Accepting the keysend payment despite having unknown even TLV type 34349334.
02e19b96774811eedd1ba5b26350950a184a7a0530b118eb653b13d5a026c24ebe-chan#1: Failing HTLC because of an invalid payload
```

Sure, we've made accepting weird TLVs a non-experimental option, but surely it's better to only (and automatically!) accept them in the keysend plugin where the problem is?

Changelog-Fixed: Plugins: `keysend` now removes unknown even (technically illegal!) fields, to try to accept more payments.

### DIFF
--- a/plugins/keysend.c
+++ b/plugins/keysend.c
@@ -248,7 +248,7 @@ struct keysend_in {
 	struct preimage payment_preimage;
 	char *label;
 	struct tlv_tlv_payload *payload;
-	struct tlv_field *preimage_field;
+	struct tlv_field *preimage_field, *desc_field;
 };
 
 static int tlvfield_cmp(const struct tlv_field *a,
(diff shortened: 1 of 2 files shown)


plugins/keysend.c
```diff
@@ -267,11 +267,27 @@ htlc_accepted_invoice_created(struct command *cmd, const char *buf,
...
 {
 	struct tlv_field field;
-	int preimage_field_idx = ki->preimage_field - ki->payload->fields;
 
-	/* Remove the preimage field so `lightningd` knows how to handle
-	 * this. */
-	tal_arr_remove(&ki->payload->fields, preimage_field_idx);
+	/* Remove all unknown fields: complain about unused even ones!
+	 * (Keysend is not a design, it's a kludge by people who
+	 * didn't read the spec, and Just Made Something Work). */
```
cdecker, line 273: 🤣

cdecker (approved):

PR #5646, opened 2022-10-04
Blinded payments spec update and infrastructure for forwarding

rustyrussell:
This does a pile of updates:
- Remove x-only pubkeys from offers.
- Update onion message and blinded payments to latest spec
- Implement forwarding of blinded payments

This PR does not include:
- Generating and actually paying invoices with blinded payments
- All the offers changes after removing x-only pubkeys.


hsmd/hsmd_wire.csv
```diff
@@ -20,7 +20,7 @@ msgdata,hsmd_init,dev_force_channel_secrets_shaseed,?sha256,
 msgtype,hsmd_init_reply,111
 msgdata,hsmd_init_reply,node_id,node_id,
 msgdata,hsmd_init_reply,bip32,ext_key,
-msgdata,hsmd_init_reply,bolt12,point32,
```
cdecker, line 23: This will break VLS compatibility, unless @devrandom has an idea on how we can work around this somehow.

cdecker (commented):
Only got as far as bd75295b18943f4b8802beb105ee720705898e70 for now, but wanted to start discussing the hsm_wire changes so we can avoid breaking last minute merges that could cause VLS and `hsmd` incompatibility.

common/gossmap.c
```diff
@@ -1296,12 +1296,11 @@ void gossmap_guess_node_id(const struct gossmap *map,
...
-	id->k[0] = SECP256K1_TAG_PUBKEY_EVEN;
-	secp256k1_xonly_pubkey_serialize(secp256k1_ctx,
-					 id->k + 1,
-					 &point32->pubkey);
+	struct pubkey pk;
+	pk.pubkey = point32->pubkey;
+	node_id_from_pubkey(id, &pk);
 
-	/* If we don't find this, let's assume it's odd. */
+	/* If we don't find this, let's assume it's the alternate. */
```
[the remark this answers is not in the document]
rustyrussell: This all gets deleted with full pubkeys anyway...

hsmd/hsmd_wire.csv
rustyrussell: Hmm, the dangers of experimental code :(

We can put in a dummy here, and simply append a new field. Or we can leave it and derive the bolt12 key separately?

cdecker: True, VLS is doing a great job at keeping in sync, so shouldn't be much of an issue. Maybe we can add a wire-version field we bump when changing the proto, or a hash of the csv file we generate from? That way we could programmatically detect incompatibilities. Just an idea going forward.

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5646/commits/df9cc4a9a8c72bd8ce1cb2a3c45878ee19fc7759

cdecker (approved):

cdecker:
There seems to be a BOLT-quote mismatch somewhere:

```
2022-10-06T05:55:25.5455402Z common/blindedpay.c:20:mismatch:		/* BOLT #4:
2022-10-06T05:55:25.5470757Z Closest match: - For every node...[: - MUST include `am]
2022-10-06T05:55:25.5484524Z make: *** [Makefile:502: bolt-check/common/blindedpay.c] Error 1
```

hsmd/hsmd_wire.csv
rustyrussell: I think it's cleaner if we use a whole new message each time. Some code duplication, but it's very simple to see what's going on.

I'll add this...

rustyrussell:
> 

Yep, fixed: this is actually a quote from the route-blinding branch only.

rustyrussell:
Trivial rebase on master, those simple typo fixes, and added (as first two commits) a new API system for hsmd_wire.csv

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5646/commits/4c25d7c0d9e638df18962c8a6bb0b795e64b0ae1

Since @cdecker approved and I need to get moving on the next PR :)

PR #5647, opened 2022-10-06
Reckless

endothermicdev:
# reckless

A Core-Lightning plugin manager

## Features

Love Core-Lightning? Want to install and test a CLN plugin with only three words and a command line? Reckless has you covered!

- Install a plugin from the lightningd/plugins repository
- Add additional source repos to search/install
- Automatically install dependencies and verify installation
- Dynamically enable/disable reckless-managed plugins
- Persists plugin state by managing a config file inherited by lightningd

## Commands

`reckless install <plugin>` searches available repositories, if one matches, clones it, installs dependencies, and tests that the plugin executes and exits normally.  If so, the plugin is permanently installed, added to the CLN config, and dynamically activated.

`reckless uninstall <plugin>` disables the plugin, removes the directory.

`reckless search <plugin>` looks through all available sources for a plugin matching this name.

`reckless enable <plugin>`/`reckless disable <plugin>` dynamically enables/disables the reckless-installed plugin and updates the config to match.

`reckless source list` list all plugin repositories.

`reckless source add <repo url>` add another plugin repository for reckless to search or install from.

## Let's Get Reckless!

Running the first time will prompt the user that their lightningd's bitcoin config will be appended (or created) to inherit the reckless config file (this config is specific to bitcoin by default.) Management of plugins will subsequently modify this file.

Start with a simple installation such as `reckless install summary`

## Troubleshooting tips

Plugins must be executable. For python plugins, the shebang is invoked, so `python3` should be available in your environment. This can be verified with `which Python3`. The default reckless directory is /home/<user>/.lightning/reckless and it should be possible for the lightningd user to execute files located here.  If this is a problem, the option flag `reckless -d=<my_alternate_dir>` may be used to relocate the reckless directory from its default. Consider creating a permanent alias in this case.

## For Developers

Tips for making your plugin compatible with reckless install:
- Choose a unique plugin name.
- The plugin entrypoint is inferred.  Naming your plugin executable the same as your plugin name will allow reckless to identify it correctly (file extensions are okay.)
- For python plugins, a requirements.txt is the preferred medium for python dependencies. A pyproject.toml will be used as a fallback, but test installation via `pip install -e .` - Poetry looks for additional files in the working directory, whereas with pip, any references to these will require something like `packages = [{ include = "*.py" }]` under the `[tool.poetry]` section.
- Additional repository sources may be added with `reckless source add https://my.repo.url/here` however, https://github.com/lightningd/plugins is included by default. Consider adding your plugin lightningd/plugins to make installation simpler.
- If your plugin is located in a subdirectory of your repo with a different name than your plugin, it will likely be overlooked.

## To Do

This is the minimum viable product of the utility [outlined by Rusty](https://gist.github.com/rustyrussell/762297e84dc5239fe408c45e6811e1a9).

Still to come:
- Man page
- Support for non-python plugins.
- Install by specifying url/dir directly.
- Possibly support installing a virtual environments for python plugins.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -411,7 +411,8 @@ ALL_NONGEN_SRCFILES := $(ALL_NONGEN_HEADERS) $(ALL_NONGEN_SOURCES)
 BIN_PROGRAMS = \
 	       cli/lightning-cli \
 	       lightningd/lightningd \
-	       tools/lightning-hsmtool
+	       tools/lightning-hsmtool\
+	       tools/reckless
 PKGLIBEXEC_PROGRAMS = \
 	       lightningd/lightning_channeld \
 	       lightningd/lightning_closingd \
--- a/tools/reckless
+++ b/tools/reckless
@@ -105,7 +105,7 @@ def remove_dir(target):
 
 class Config():
     """A generic class for procuring, reading and editing config files"""
-    def obtain_config(self, config_path, default_text):
+    def obtain_config(self, config_path, default_text, warn=False):
         """Return a config file from the desired location. Create one with
         default_text if it cannot be found."""
         if isinstance(config_path, type(None)):
@@ -117,8 +117,11 @@ class Config():
                 config_content = f.readlines()
             return config_content
         print(f'config file not found: {config_path}')
-        confirm = input('press [Y] to create one now.\n')
-        if confirm.upper() != 'Y':
+        if warn:
+            confirm = input('press [Y] to create one now.\n').upper() == 'Y'
+        else:
+            confirm = True
+        if not confirm:
             sys.exit(1)
         parent_path = os.path.split(config_path)[0]
         # Create up to one parent in the directory tree.
@@ -161,11 +164,12 @@ class Config():
                 if not line_exists:
                     conf_write.write(f'\n{addline}')
 
-    def __init__(self, path=None, default_text=None):
+    def __init__(self, path=None, default_text=None, warn=False):
         assert path is not None
         assert default_text is not None
         self.conf_fp = path
-        self.content = self.obtain_config(self.conf_fp, default_text)
+        self.content = self.obtain_config(self.conf_fp, default_text,
+                                          warn=warn)
 
 
 class RecklessConfig(Config):
@@ -199,13 +203,13 @@ class LightningBitcoinConfig(Config):
     """lightningd config specific to the bitcoin network. This is inherited by
     the main lightningd config and in turn, inherits bitcoin-reckless.conf."""
 
-    def __init__(self, path=None, default_text=None):
+    def __init__(self, path=None, default_text=None, warn=True):
         if path is None:
             path = os.path.join(LIGHTNING_DIR, 'bitcoin', 'config')
         if default_text is None:
             default_text = "# This config was autopopulated by reckless\n\n"
         Config.__init__(self, path=str(path),
-                        default_text=default_text)
+                        default_text=default_text, warn=warn)
 
 
 class InferInstall():
--- a/tools/reckless
+++ b/tools/reckless
@@ -232,12 +232,12 @@ class InferInstall():
         raise Exception(f'plugin entrypoint not found in {self.dir}')
 
 
-def help(target):
-    if len(target) > 0:
-        if target[0] in globals() and hasattr(globals()[target[0]], '__doc__'):
-            print(globals()[target[0]].__doc__)
-    else:
+def help_alias(target):
+    if len(target) == 0:
         parser.print_help(sys.stdout)
+    else:
+        print('try "reckless {} -h"'.format(' '.join(target)))
+        sys.exit(1)
 
 
 def verbose(*args):
@@ -697,9 +697,10 @@ if __name__ == '__main__':
     source_rem.add_argument('targets', type=str, nargs='*')
     source_rem.set_defaults(func=remove_source)
 
-    help_cmd = cmd1.add_parser('help', help='display this message')
+    help_cmd = cmd1.add_parser('help', help='for contextual help, use '
+                               '"reckless <cmd> -h"')
     help_cmd.add_argument('targets', type=str, nargs='*')
-    help_cmd.set_defaults(func=help)
+    help_cmd.set_defaults(func=help_alias)
 
     args = parser.parse_args()
 
--- a/tools/reckless
+++ b/tools/reckless
@@ -82,14 +82,14 @@ class InstInfo:
 
 def create_dir(r: int, directory: str) -> bool:
     """Creation of a directory at path `d` with a maximum new dir depth `r`"""
-    if os.path.exists(directory):
+    if Path(directory).exists():
         return True
     elif r <= 0:
         return False
-    elif create_dir(r-1, os.path.split(directory)[0]):
+    elif create_dir(r-1, Path(directory).parent):
         os.mkdir(directory, 0o777)
         print(f'created directory {directory}')
-        assert os.path.exists(directory)
+        assert Path(directory).exists()
         return True
 
 
@@ -116,7 +116,7 @@ class Config():
             raise Exception("Generic config must be passed a config_path.")
         assert isinstance(config_path, str)
         # FIXME: warn if reckless dir exists, but conf not found
-        if os.path.exists(config_path):
+        if Path(config_path).exists():
             with open(config_path, 'r+') as f:
                 config_content = f.readlines()
             return config_content
@@ -127,7 +127,7 @@ class Config():
             confirm = True
         if not confirm:
             sys.exit(1)
-        parent_path = os.path.split(config_path)[0]
+        parent_path = Path(config_path).parent
         # Create up to one parent in the directory tree.
         if create_dir(1, parent_path):
             with open(self.conf_fp, 'w') as f:
@@ -196,14 +196,14 @@ class RecklessConfig(Config):
     def __init__(self, path: Union[str, None] = None,
                  default_text: Union[str, None] = None):
         if path is None:
-            path = os.path.join(LIGHTNING_DIR, 'reckless',
-                                'bitcoin-reckless.conf')
+            path = Path(LIGHTNING_DIR).joinpath('reckless',
+                                                'bitcoin-reckless.conf')
         if default_text is None:
             default_text = '# This configuration file is managed by reckles' +\
                            's to activate and disable\n# reckless-installed' +\
                            ' plugins\n\n'
         Config.__init__(self, path=str(path), default_text=default_text)
-        self.reckless_dir = os.path.split(path)[0]
+        self.reckless_dir = Path(path).parent
 
 
 class LightningBitcoinConfig(Config):
@@ -214,7 +214,7 @@ class LightningBitcoinConfig(Config):
                  default_text: Union[str, None] = None,
                  warn: bool = True):
         if path is None:
-            path = os.path.join(LIGHTNING_DIR, 'bitcoin', 'config')
+            path = Path(LIGHTNING_DIR).joinpath('bitcoin', 'config')
         if default_text is None:
             default_text = "# This config was autopopulated by reckless\n\n"
         Config.__init__(self, path=str(path),
@@ -228,16 +228,16 @@ class InferInstall():
         if name[-3:] == '.py':
             name = name[:-3]
         if name in reck_contents:
-            self.dir = os.path.join(RECKLESS_CONFIG.reckless_dir, name)
+            self.dir = Path(RECKLESS_CONFIG.reckless_dir).joinpath(name)
         else:
             raise Exception(f"Could not find a reckless directory for {name}")
-        plug_contents = os.listdir(os.path.join(RECKLESS_CONFIG.reckless_dir,
-                                                name))
-        for n in py_entry_guesses(name):
-            if n in plug_contents:
-                self.entry = os.path.join(self.dir, n)
-                self.name = n
-                return
+        plug_dir = Path(RECKLESS_CONFIG.reckless_dir).joinpath(name)
+        for guess in py_entry_guesses(name):
+            for content in plug_dir.iterdir():
+                if content.name == guess:
+                    self.entry = str(content)
+                    self.name = guess
+                    return
         raise Exception(f'plugin entrypoint not found in {self.dir}')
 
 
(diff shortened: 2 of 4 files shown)


tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+
+repos = ['https://github.com/lightningd/plugins']
+
+
+def py_entry_guesses(name):
+    return [name, f'{name}.py', '__init__.py']
+
+
+def unsupported_entry(name):
+    return [f'{name}.go', f'{name}.sh']
```
cdecker, line 22: Might be good to document these conventions

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+        self.git_url = git_url  # API access for github repos
+        self.entry = None
+        self.deps = None
+        self.subdir = None
+        self.commit = None
+
+    def __repr__(self):
+        return f'<InstInfo object>\n'\
+               f'name: {self.name}\nrepo: {self.repo}\ngit: {self.git_url}'\
+               f'\nentry:{self.entry}\ndepency source:{self.deps}'
```
cdecker, line 38:
```suggestion
               f'\nentry:{self.entry}\ndependency source:{self.deps}'
```

cdecker: Multiline `__repr__`s are a bit strange, but you can easily achieve multi-line strings this way:

```
f"""<InstInfo object>
name: {self.name}
repo: {self.repo}
git: {self.git_url}
entry:{self.entry}
dependency source:{self.deps}"""
```
That way we can read the output in the code as it would appear in the output. If indentation is an issue then `textwrap.dedent` might be useful.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+    """Creation of a directory at path `d` with a maximum new dir depth `r`"""
+    if os.path.exists(directory):
+        return True
+    elif r <= 0:
+        return False
+    elif create_dir(r-1, os.path.split(directory)[0]):
+        os.mkdir(directory, 0o777)
+        print(f'created directory {directory}')
+        if os.path.exists(directory):
+            return True
```
cdecker, line 92: This can be achieved by `pathlib.Path(directory).mkdir(parents=True, exist_ok=True)`: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+
+def remove_dir(target):
+    try:
+        shutil.rmtree(target)
+        return True
+    except NotADirectoryError:
+        print(f"Tried to remove directory {target} that does not exist.")
+    except PermissionError:
+        print(f"Permission denied removing dir: {target}")
+    return False
```
cdecker, line 103: Shouldn't this bubble the error up, so we can handle it where it was caused by trying to delete the directory?

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+            raise Exception("Generic config must be passed a config_path.")
+        assert isinstance(config_path, str)
+        # FIXME: warn if reckless dir exists, but conf not found
+        if os.path.exists(config_path):
+            with open(config_path, 'r+') as f:
+                config_content = f.readlines()
+            return config_content
+        print(f'config file not found: {config_path}')
+        if warn:
+            confirm = input('press [Y] to create one now.\n').upper() == 'Y'
```
cdecker, line 121: User interaction should most likely be in a function, and called in the command handlers, not buried in a class that one might call from other places too.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+        else:
+            confirm = True
+        if not confirm:
+            sys.exit(1)
+        parent_path = os.path.split(config_path)[0]
```
cdecker, line 126: These types of operations get much easier with `pathlib.Path` :-)

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+        # Create up to one parent in the directory tree.
+        if create_dir(1, parent_path):
+            with open(self.conf_fp, 'w') as f:
+                f.write(default_text)
+                # FIXME: Handle write failure
+                return default_text
+        else:
+            print('could not create the parent directory')
```
cdecker, line 134: The idiomatic thing here would be to raise an exception and handle it where we know how to react.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+    except NotADirectoryError:
+        print(f"Tried to remove directory {target} that does not exist.")
+    except PermissionError:
+        print(f"Permission denied removing dir: {target}")
+    return False
+
+
+class Config():
+    """A generic class for procuring, reading and editing config files"""
+    def obtain_config(self, config_path, default_text, warn=False):
```
cdecker, line 108: Feels a lot like it should be a `@classmethod`, i.e., a constructor that returns the config instance fully configured.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+                f.write(default_text)
+                # FIXME: Handle write failure
+                return default_text
+        else:
+            print('could not create the parent directory')
+            return None
+
+    def editConfigFile(self, addline, removeline):
+        remove_these_lines = []
+        with open(self.conf_fp, 'r') as reckless_conf:
```
cdecker, line 139: Since you just read the file you can also use this:

```
lines = open(self.conf_fp, 'r').readlines()
```

And then use `lines` from thereon, no need for a context manager.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+                        else:
+                            conf_write.write(l.strip())
+                        if addline == l:
+                            # addline is idempotent
+                            line_exists = True
+                if not line_exists:
+                    conf_write.write(f'\n{addline}')
+
+    def __init__(self, path=None, default_text=None, warn=False):
+        assert path is not None
```
cdecker, line 168: No point in declaring `path=None` and then failing if it is `None`. Default args suggest that you can omit them. Same for `default_text`.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+
+class RecklessConfig(Config):
+    """Reckless config (by default, specific to the bitcoin network only.)
+    This is inherited by the lightningd config and contains all reckless
+    maintained plugins."""
+
+    def enable_plugin(self, plugin_path):
+        """Handle persistent plugin loading via config"""
+        self.editConfigFile(f'plugin={plugin_path}',
+                            f'disable-plugin={plugin_path}')
```
cdecker, line 183: We could also just comment them out.

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+
+    def disable_plugin(self, plugin_path):
+        """Handle persistent plugin disabling via config"""
+        self.editConfigFile(f'disable-plugin={plugin_path}',
+                            f'plugin={plugin_path}')
+
+    def __init__(self, path=None, default_text=None):
+        if path is None:
+            path = os.path.join(LIGHTNING_DIR, 'reckless',
+                                'bitcoin-reckless.conf')
```
cdecker, line 193: You didn't fall into the mutable-default-argument trap that makes life of a pythonista interesting, well done :+1:

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+        if default_text is None:
+            default_text = '# This configuration file is managed by reckles' +\
+                           's to activate and disable\n# reckless-installed' +\
```
cdecker, line 196:
```suggestion
            default_text = '# This configuration file is managed by reckless ' +\
                           'to activate and disable\n# reckless-installed' +\
```

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+def verbose(*args):
+    if not IS_VERBOSE:
+        return
+    print(*args)
+
+
+def _search_repo(name, url):
+    """look in given repo and, if found, populate InstInfo"""
+    # Remove api subdomain, subdirectories, etc.
+    repo = url.split('/')
```
cdecker, line 251: `urlparse` has a very robust parsing implementation:

https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+                # Maybe we were passed an api.github.com/repo/<user> url
+                start = start + 1
+            repo_user = repo[start]
+            repo_name = repo[start+1]
+            break
+    # FIXME: Handle non-github repos.
+    # Get details from the github API.
+    if repo_name is not None:
+        api_url = f'https://api.github.com/repos/{repo_user}/' + \
+                  f'{repo_name}/contents/'
```
cdecker, line 269: Python can auto-concatenate string continuations, no need for `+ \`:

```suggestion
        api_url = (
            f'https://api.github.com/repos/{repo_user}/'
            f'{repo_name}/contents/'
        )
```

Gives back some of the characters that it is using on the whitespace indentation :-)

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+    if len(target) > 0:
+        print(globals()[target[0]].__doc__)
+    else:
+        parser.print_help(sys.stdout)
+
+
+def verbose(*args):
+    if not IS_VERBOSE:
+        return
+    print(*args)
```
cdecker, line 245: Python has quite the powerful `logging` library at your disposal: https://docs.python.org/3/library/logging.html

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+    print(f'Unable to locate source for plugin {plugin_name}')
+
+
+def lightning_cli_available():
+    clncli = Popen(['lightning-cli'], stdout=PIPE, stderr=PIPE)
+    clncli.wait(timeout=1)
+    if clncli.returncode == 0:
+        return True
+    else:
+        return False
```
cdecker, line 458: We could also check if we have an executable on the `$PATH`:

```
from distutils.spawn import find_executable
find_executable('lightning-cli') 
```

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+    if not os.path.exists(path):
+        print('cannot find installed plugin at expected path {}'
+              .format(path))
+        sys.exit(1)
+    verbose('activating {}'.format(plugin_name))
+
+    if not lightning_cli_available():
+        # Config update should not be dependent upon lightningd running
+        RECKLESS_CONFIG.enable_plugin(path)
+        return
```
cdecker, line 480: This can be merged with lines 484-486 I think

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+
+    clncli = Popen(['lightning-cli', 'plugin', 'start', path], stdout=PIPE)
```
cdecker, line 482: We most likely want to wrap the `lightning-cli` call in a function, raising an exception if it fails, and make it configurable as well. In my case for example it didn't pick up the `lightning-dir` and failed to start the plugin.

cdecker (commented):
Thanks @endothermicdev, this is a great first iteration of `reckless`. Since this is the first python built-in tool destined for the end-user I went through it with a fine-toothed comb, so don't worry about the many (optional) comments.

Let me know if I can help, or some of the comments aren't clear :hugs:

tools/reckless
endothermicdev: Wow, this is so much nicer!

tools/reckless
```diff
@@ -0,0 +1,684 @@
...
+                        type=str)
+    # This default depends on the .lightning directory
+    parser.add_argument('-d', '--reckless-dir',
+                        help='specify a data directory for reckless to use',
+                        type=str, default=None)
+    parser.add_argument('-l', '--lightning',
+                        help='lightning data directory (default:~/.lightning)',
+                        type=str,
+                        default=Path.home().joinpath('.lightning'))
+    parser.add_argument('-r', '--regtest', action='store_true')
```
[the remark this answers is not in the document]
endothermicdev: I guess I missed this in the docs. Exactly what I was looking for, thanks!

cdecker (approved):
Thanks for the fixups @endothermicdev, I only found a couple of cleanups that could also be done in a separate cleanup PR. I think the user-visible interface looks good, and we can clean up under the covers.

Which brings us to the question of whether you're happy to merge this as is, and I'd add it to the milestone for the release, or whether you'd like to hold it back, and we can include it in two months. WDYT?

tools/reckless
```diff
@@ -463,33 +497,21 @@ def enable(plugin_name: str):
...
-
-    if not lightning_cli_available():
-        # Config update should not be dependent upon lightningd running
-        RECKLESS_CONFIG.enable_plugin(path)
-        return
-
-    cmd = LIGHTNING_CLI_CALL.copy()
-    cmd.extend(['plugin', 'start', path])
-    clncli = Popen(cmd, stdout=PIPE)
-    clncli.wait(timeout=3)
```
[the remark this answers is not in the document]
cdecker: And of course this is in direct contradiction with my [earlier comment](https://github.com/ElementsProject/lightning/pull/5647#discussion_r995980169) :facepalm: Feel free to ignore this :wink:

tools/reckless
```diff
@@ -353,6 +353,7 @@ def _install_plugin(src: InstInfo) -> bool:
     if src.subdir is not None:
         plugin_path = Path(clone_path).joinpath(src.subdir)
     os.chdir(plugin_path)
+    assert os.getcwd() == str(plugin_path)
```
cdecker, line 356: This is likely unexpected for the user since they'll be dropped into a temporary directory if anything goes wrong. You can use the `cwd` argument to `Popen` to have it change the directory in the subprocess before exec'ing the executable, thus leaving the parent process in its current working directory unchanged.

Let's do that in a cleanup PR, since this is just a minor fix.

Makefile
```diff
@@ -801,7 +801,9 @@ install-data: installdirs $(MAN1PAGES) $(MAN5PAGES) $(MAN7PAGES) $(MAN8PAGES) $(
 	$(INSTALL_DATA) $(MAN8PAGES) $(DESTDIR)$(man8dir)
 	$(INSTALL_DATA) $(DOC_DATA) $(DESTDIR)$(docdir)
 
-install: install-program install-data
+install-python: tools/pip-install
```
cdecker, line 804: This file seems to be missing. What is it's purpose?

tools/Makefile
```diff
@@ -28,4 +32,12 @@ tools-clean:
 	$(RM) tools/headerversions
 	$(RM) tools/headerversions.o
 
+tools/pip-install:
+	@if which $(PIP_EXE) >> /dev/null; then \
+		echo "installing python dependencies"; \
+		$(PIP_EXE) $(PIP_ARGS) install -r tools/requirements.txt ; \
+	else \
+		(echo "Pip not installed, skipping python dependencies") ; \
+	fi
```
cdecker, line 41: Ok, I see where this comes from, but it most likely isn't what we want:

 - Developer dependencies are tracked in `pyproject.toml`
 - There is currently no user dependencies (other than libraries from the distro), so we need something here. On the other hand the users will not be the ones to run `Makefile`s, so we better look for some better option. Alternatively we could just use `urllib3` and not have dependencies at all.

Potentially reckless could be a completely separate tool that we distribute on PyPI where it can have its own dependencies, and users would just be a `pip install reckless` away from being able to run it.

cdecker (changes requested):
Good fixes, but I think we probably want to lose that last commit.

tools/Makefile
endothermicdev: I'm certainly open to a better option for installing python dependencies for the user (or at least prompting a `pip install`.)  I think we'll want that at some point.  Let's just switch to `urllib3` for the time being?  Distributing reckless via PyPI is fine, but I would still prefer it to be updated and working by default with the latest CLN release if possible.

cdecker (approved):

PR #5649, opened 2022-10-07
PSBT: wipe scriptSig/witnesses in global tx

niftynei:
We do some hacky stuff to get PSBTs re-built from a tx but forget to wipe out the scriptSig data; which caused the tx to fail to be written to/from the wire successfully. Whoops.

Changelog-None


ddustin:
ACK e9679cd0bcc3ebbdeb218cf2e42359f1dae23781 -- works great on my machine!

cdecker:
That must have been one hell of a debugging session to find this. Good catch :+1:

ACK e9679cd0bcc3ebbdeb218cf2e42359f1dae23781

PR #5652, opened 2022-10-07
update alpine build packages

briancolecoinmetrics:


briancolecoinmetrics:
Also it looks like building releases might require `bash`. Not clear on what's happening there; it doesn't happen on tip of master.

vincenzopalazzo:
Mh we should compile on alpine each time inside the CI, what type of error are you facing? or does the Ci include different packages?

vincenzopalazzo:
This is the configuration that ran in the CI https://github.com/ElementsProject/lightning/blob/master/contrib/docker/Dockerfile.alpine

briancolecoinmetrics:
So that Dockerfile contains this for dependencies:
```
RUN apk update && \
    apk add ca-certificates alpine-sdk autoconf automake git libtool \
    gmp-dev sqlite-dev python3 py3-mako net-tools zlib-dev libsodium gettext su-exec \
    python3 py3-pip #&& \
    #apk add --upgrade fortify-headers
```
The difference from the list of packages in INSTALL.md is that this doesn't include `python`, does include `su-exec` and `py3-pip`, and includes `python3` twice.

I don't know what `su-exec` is for, but removing the extra python stuff from Dockerfile.alpine appears to still produce a working image, so I have implemented that change and bumped the base image to the current version of Alpine and included it into this PR.

briancolecoinmetrics:
> so I have implemented that change and bumped the base image to the current version of Alpine and included it into this PR.

...which brings the dependencies in INSTALL.md and Dockerfile.alpine into sync except for `su-exec` (which I suspect is specifically for something about how the original author runs containers). Realized after I hit "Comment" that I should be explicit.

vincenzopalazzo:
Thanks! I'm really a monkey with alpine the bad docker file is my bad, but bringing the INSTALL.md and the Docker file in sync sounds the way to go to me, feel free to ping me if you have any problem with this.

contrib/docker/Dockerfile.alpine
```diff
@@ -1,12 +1,12 @@
...
+FROM alpine:3.16
 LABEL org.opencontainers.image.authors="Vincenzo Palazzo (@vincenzopalazzo) vincenzopalazzodev@gmail.com"
 
 WORKDIR /build
 
 RUN apk update && \
     apk add ca-certificates alpine-sdk autoconf automake git libtool \
     gmp-dev sqlite-dev python3 py3-mako net-tools zlib-dev libsodium gettext su-exec \
-    python3 py3-pip #&& \
+    ;
```
cdecker, line 9: It's probably sufficient to remove the `\` on line 8 and we can drop this line altogether.

Also notice that the next line is commented out, so we should remove that as well.

cdecker (dismissed):

cdecker:
The `python` dependency used to be required by `libwally` unless a special `configure` flag was set. We now set that flag when building ourselves, so the old `python` (used to be python2.7 btw) can be dropped.

Very nice cleanup, I wasn't aware of the virtual package system in `apk`. There are just some tiny cleanups in the last commit. If you want I can just do them and get the PR through CI :+1:

cdecker:
Trying out a cleaned up version of this PR, turns out it is broken anyway: it is trying to link against gmp without PIE support, so we'll need to address that as well.

cdecker:
My bad, turns out it was picking up some artifacts from the source repository. Better separate the git origin from the build directory to avoid these in the future.

Still have one small issue with the `bitcoin-cli` failing to be called.

vincenzopalazzo:
Why do we need `bitcoin-cli` in this case? the idea was just to compile in order to work with the more strict compiler used in alpine, do you think that installing these extra dependencies could waste time? or are you thinking to run also the test inside?

briancolecoinmetrics:
This version is fine, but it's worth noting that alpine has a `bitcoin-cli` package in their official repos, so you could also just `apk add bitcoin-cli` if you want to support that.

Thanks for the improvements; I was trying to make the minimal changes to make things not broken, but moving to a fully functional docker image is a nice improvement.

cdecker:
Ok, turns out that using `bitcoin-cli` from the alpine repo works like a charm, making this Docker image functional again :+1:

cdecker:
@vincenzopalazzo and @briancolecoinmetrics can you give me an ACK if the intention of the PR has been met?

vincenzopalazzo:
ACK https://github.com/ElementsProject/lightning/pull/5652/commits/972e9e4bb75dfdbd33c32c827a10b8bfb115d928

PR #5658, opened 2022-10-13
poetry: Update python dependencies

cdecker:
Dependabot is complaining that we have an outdated `protobuf` version in our lockfile, let's fix that.


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5658/commits/c212c20a254139da1669a41447bf67fe18136559

cdecker:
Had to trim down the updates, since there seems to be something that gets `lnprototest` stuck once again.

PR #5665, opened 2022-10-18
psbt: Prevent memleak

ddustin:
When calling `psbt_*_set_serial_id` with a NULL ctx the key value gets leaked because it’s just a temporary variable (`psbt_*_set_unknown` makes a copy of key and doesn’t steal it).

Seems to only effect splicing usage but seems like a good general fix to have.

Changelog-None

### DIFF
--- a/common/psbt_open.c
+++ b/common/psbt_open.c
@@ -325,10 +325,11 @@ void psbt_input_set_serial_id(const tal_t *ctx,
 			      struct wally_psbt_input *input,
 			      u64 serial_id)
 {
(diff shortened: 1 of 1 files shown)


rustyrussell:
I don't see how?

It's allocating key off `tmpctx` not `ctx` here?

ddustin:
> I don't see how?
> 
> 
> 
> It's allocating key off `tmpctx` not `ctx` here?

Oh strange -- on my branch two of these were allocating off ctx and two off tmpctx. In making the PR off master I didn't double check the diff.

Either it was something messed up just on my branch or someone fixed it in the meantime 🙈.

PR #5669, opened 2022-10-19
onchaind: Adjust the witness weight estimation to be a bit more conservative

cdecker:
This came up while debugging the VLS signer for Greenlight. In case of the feerate hitting the floor it is important we agree on the size of the transaction, and in particular the witness weight which is estimated, not measured. If the estimation of VLS is larger than the one in CLN we end up with too small a fee according to VLS, and it'll refuse to sign.

The differences we found were due to a) a missing `OP_PUSH33` for the pubkeys, and b) the lower expected signature weight of 72 instead of the spec suggestion of 73.

### DIFF
--- a/tests/test_closing.py
+++ b/tests/test_closing.py
@@ -26,7 +26,7 @@ def test_closing_simple(node_factory, bitcoind, chainparams):
     l1, l2 = node_factory.line_graph(2, opts={'plugin': coin_mvt_plugin})
     chan = l1.get_channel_scid(l2)
     channel_id = first_channel_id(l1, l2)
-    fee = closing_fee(3750, 2) if not chainparams['elements'] else 4263
+    fee = closing_fee(3750, 2) if not chainparams['elements'] else 4278
 
     l1.pay(l2, 200000000)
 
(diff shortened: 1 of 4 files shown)


cdecker:
As discussed here https://github.com/lightningdevkit/rust-lightning/pull/1783#issuecomment-1286277916 the spec is likely a bit too conservative with its 73 byte signatures, given that the low-S rule ensures we don't have 73 bytes, but 72 bytes.

tests/test_closing.py
```diff
@@ -3642,12 +3642,12 @@ def test_close_weight_estimate(node_factory, bitcoind):
...
-    # This could actually be a bit shorter: 1 in 256 chance we get
-    # lucky with a sig and it's shorter.  We have 2 sigs, so that's
-    # 1 in 128.  Unlikely to do better than 2 bytes off though!
+    # To match the signer's estimate we use the pessimistic estimate
+    # of 73bytes / signature, which often ends up overestimating the
+    # actual weight by up to 4 bytes.
     signed_weight = int(bitcoind.rpc.decoderawtransaction(tx)['weight'])
     assert signed_weight <= actual_weight
-    assert signed_weight >= actual_weight - 2
+    assert signed_weight >= actual_weight - 4
```
rustyrussell, line 3650: This is confusing!  actual_weight should be called "final_estimate" since that's what it is: closingd's weight estimate for the final negotiated tx.

And this change is wrong. We are *always* 4 bytes under, since we actually grind to 71 bytes per signature, so first assert should be:

```
        assert signed_weight <= final_estimate - 4
```

But occasionally we get lucky (and fail CI like we did here), indeed, decoding shows we got a 70 byte signature, so the second should be:

```
        assert signed_weight >= final_estimate - 6
```

rustyrussell (changes requested):

tests/test_closing.py
cdecker: Ok, now that makes sense, adjusted as per your comments.

vincenzopalazzo (approved):
utACK https://github.com/ElementsProject/lightning/pull/5669/commits/4cc20e35af0ad37b6051a5ca68146b9cbcb95883

PR #5673, opened 2022-10-21
fd_send: Add explicit ack packet

ddustin:
* Adding an explicit ACK 1 byte packet of the fd_send
* This should hopefully allows us to remove the sleep trick
* And doesn’t appear to bring back the flack’ness issue (at least on my machine)

Would be great to get additional testing from anyone else who’s environment has experience this particular flake.

### DIFF
--- a/ccan/ccan/fdpass/fdpass.c
+++ b/ccan/ccan/fdpass/fdpass.c
@@ -3,6 +3,9 @@
 #include <sys/socket.h>
 #include <errno.h>
 #include <string.h>
+#include <unistd.h>
+
+#define MAGIC_ACK_NUMBER 85
 
 bool fdpass_send(int sockout, int fd)
 {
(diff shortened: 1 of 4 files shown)


ddustin:
Two questions nagging the back of my mind:
1) Is looping in fd_send for the receive ACK (to handle non-blocking sockets) fine?
2) Should we also be looping in fd_recv around the write call incase it's non-blocking and the kernel is backlogged?

cdecker:
Just wanted to relay some of the comments made on Discord to the PR for posterity:

 - The `ccan` code is managed externally and would have to be changed upstream first, and then merged back into the CLN tree.
 - Several doubts came up that this does indeed address the issue, or papers over it. In particular the ACK forces there to be at least one context switch, allowing the recipient to process the incoming message, but it doesn't guarantee the message has been processed. This potentially makes future instances of this issue harder to identify, while not addressing the flakyness.
   - If such semantics are desired we should rather build a task-queue abstraction on top of `fdpass`, where the ACK is returned after the message and FD have been processed correctly.

rustyrussell:
No, this changes the ccan module (which is supposed to simply pass an fd, not be a protocol for fd passing!) for some reason we can't isolate in a larger piece of code.  We'll stick with the workaround we have until someone root-causes this properly.

PR #5674, opened 2022-10-25
msggen: Expose `extratlvs` of `keysend`

cdecker:
It's a bit special in the sense that it is a composite whose keys are numbers and not enumerated (they're custom types).


cdecker:
This seems mostly unrelated to `msggen`. You can switch to `key=value` mode on the command line by adding the `-k` switch, and then you can skip the nulls which are required in positional mode.

dennisreimann:
Got it done, thanks! Turns out 5482373484 shouldn't be part of the payload.

PR #5681, opened 2022-10-31
Field filtering: limit what fields will be returned from (almost) any JSON command.

rustyrussell:
Sometimes you only care about a single field from a request: this especially makes this faster if you're running remotely (RTL, commando, etc).  So we add a simple `_filter` field, which is a template of the same form as the request, but only fields mentioned in this template will be included.

In future, commands may examine the filter to avoid calculations altogether.  And pyln may also be enhanced to support it (plugins which don't support it have it stripped, so no change for them!).


jb55:
very nice

doc/PLUGINS.md
```diff
@@ -136,7 +136,8 @@ example:
 	  "method": "mycustomnotification"
 	}
   ],
-  "dynamic": true
+  "dynamic": true,
+  "fields_filtering: true
```
jb55, line 140: missing " here

doc/PLUGINS.md
```diff
@@ -162,6 +163,10 @@ The `dynamic` indicates if the plugin can be managed after `lightningd`
 has been started using the [plugin][lightning-plugin] JSON-RPC command. Critical plugins that should not be stopped should set it
 to false. Plugin `options` can be passed to dynamic plugins as argument to the `plugin` command .
 
+The `fields_filtering` field indicates if the plugin's rpc calls can
+handle the `_filter` parameter.  If your plugin supplies rpcmthods,
+you should implement this!
+
```
cdecker, line 169: An alternative to this would have been to slightly break with the JSON-RPC spec and allow us to add meta-parameters in the top-level dict, instead of in the `payload` keyed dict. This is would have been nice since the filtering, and eventually paging, is orthogonal to the actual operations we're asking `lightningd` to perform.

That would then also have meant that plugins that do not support such post-processing of the results are simply oblivious to it, and we don't need a magic prefix to differentiate method arguments from result post-processing instructions.

doc/lightningd-rpc.7.md
```diff
@@ -0,0 +1,271 @@
+lightningd-rpc -- Lightning Daemon RPC Protocols
+================================================
```
cdecker, line 2: Good idea to add this to the docs :+1:

cdecker (approved):
Very cool idea. Apart from a minor suggestion to locate the result post-processing separately from the arguments to the method, I think this is good to go :+1:

jb55:
On Tue, Nov 01, 2022 at 10:21:59AM -0700, Christian Decker wrote:
>An alternative to this would have been to slightly break with the JSON-RPC spec and allow us to add meta-parameters in the top-level dict, instead of in the `payload` keyed dict. This is would have been nice since the filtering, and eventually paging, is orthogonal to the actual operations we're asking `lightningd` to perform.

I originally assumed this is how it would have been implemented but I'm
ok with the current implementation as well.

rustyrussell:
Hmm, good point re: a separate field being neater.  However, that also means every layer above needs to know about it.

1. lightning-cli needs a --filter argument
2. pyln.client and grpc need updating either way since we already need to either wrap them all or have a "set_filter" function.
3. commando needs to handle this somehow.
4. c-lightning-rest needs updating.

I'd like @ShahanaFarooqui input on this!

ShahanaFarooqui:
Keeping it as a separate field looks more logical and neat solution, even though it will require some top layer updates initially.

Regarding c-lightning-rest, this update should not take much time to implement.

doc/PLUGINS.md
rustyrussell: OK, I have done this instead!

rustyrussell:
OK, this reworks it to use a (non-standard) "filter" object in the request.  This is simpler, though we now need to have an explicit lightning-cli `--filter` option, and a context manager for pyln to set it.

contrib/pyln-client/pyln/client/lightning.py
```diff
@@ -438,6 +441,22 @@ def fn(message, progress, request, **kwargs):
...
+    @contextmanager
+    def reply_filter(self, filter):
+        """Filter the fields returned from am RPC call (or more than one)..
+
+        This is a context manager and should be used like this:
+
+        ```python
+        with rpc.reply_filter({"transactions": [{"outputs": [{"amount_msat": true, "type": true}]}]}):
+            rpc.listtransactions()
+        ```
```
cdecker, line 453: This is very cool ^^

cdecker (approved):
ACK https://github.com/ElementsProject/lightning/pull/5681/commits/1ae3a4747c561ba3ba98d5fa806088ab5c644399

cdecker:
Source code checker choked on duplicate include. Fixed and force-pushed.

cdecker:
Hm, still fails, this time because with `clang` + `fuzzing` we have a reference to the `json_filter` infrastructure that doesn't get linked in. I tried adding it to the linker, but then we start slowly leaking out into the `lightningd`-specific code, which we likely don't want at all. To isolate the fuzz targets from `lightningd` may take some more time and a bit of rework.

Two options:
 - Disable the fuzzing targets that reference those symbols, and keep this PR in the release
 - Defer this PR until the next release.

What do you think @rustyrussell ?

contrib/pyln-client/pyln/client/lightning.py
rustyrussell: I stole it from your context code I think!!

rustyrussell:
Found a trivial fix: remove the *other* JSON common objects from the fuzz makefile!

Ack 3bbdd359e870084812ea4861a0d387cdc972c860

rustyrussell:
Fuzz fix was trivial (stop including *other* JSON files).

Ack https://github.com/ElementsProject/lightning/commit/a6f38a2c1a47c5497178c199691047320f2c55bc

PR #5685, opened 2022-11-01
Rework the CI system

cdecker:
Just a bunch of changes that I'm working on to speed up and deduplicate work in CI.

### DIFF
--- a/.github/scripts/install-bitcoind.sh
+++ b/.github/scripts/install-bitcoind.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+set -e
+
+DIRNAME="bitcoin-${BITCOIN_VERSION}"
+EDIRNAME="elements-${ELEMENTS_VERSION}"
+FILENAME="${DIRNAME}-x86_64-linux-gnu.tar.bz2"
+EFILENAME="${EDIRNAME}-x86_64-linux-gnu.tar.bz2"
+
+cd /tmp/
+wget "https://storage.googleapis.com/c-lightning-tests/$FILENAME"
+wget -q "https://storage.googleapis.com/c-lightning-tests/${EFILENAME}"
+tar -xaf "${FILENAME}"
+tar -xaf "${EFILENAME}"
+sudo mv "${DIRNAME}"/bin/* "/usr/local/bin"
+sudo mv "${EDIRNAME}"/bin/* "/usr/local/bin"
+
+
+rm -rf "${FILENAME}" "${EFILENAME}" "${DIRNAME}" "${EDIRNAME}"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,7 +51,7 @@ jobs:
     # The unit test checks are not in the critical path (not dependent
     # on the integration tests), so run them with `valgrind`
     name: Run unit tests
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       COMPAT: 1
       VALGRIND: 1
@@ -70,6 +70,7 @@ jobs:
       - name: Install dependencies
         run: |
           bash -x .github/scripts/setup.sh
+          sudo apt-get install -y -qq lowdown
           pip install -U pip wheel poetry
           # Export and then use pip to install into the current env
           poetry export -o /tmp/requirements.txt --without-hashes --with dev
@@ -84,7 +85,7 @@ jobs:
 
   compile:
     name: Compile CLN ${{ matrix.cfg }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       COMPAT: 1
     needs:
@@ -155,7 +156,7 @@ jobs:
 
   integration:
     name: Test CLN ${{ matrix.cfg }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     env:
       COMPAT: 1
       BITCOIN_VERSION: 0.20.1
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,7 +13,7 @@ jobs:
     env:
       RUST: 1
       COMPAT: 1
-      BOLTDIR: /tmp/bolts
+      BOLTDIR: bolts
     strategy:
       fail-fast: true
     steps:
@@ -38,14 +38,45 @@ jobs:
         run: ./configure
       - name: Check source
         run: make -j 4 check-source
-      - name: Check units
-        run: make -j 4 check-units
       - name: Check Generated Files have been updated
         run: make -j 4 check-gen-updated
       - name: Check docs
         run: make -j 4 check-doc
-      - name: Check install
-        run: make -j 4 installcheck
+
+  check-units:
+    # The unit test checks are not in the critical path (not dependent
+    # on the integration tests), so run them with `valgrind`
+    name: Run unit tests
+    runs-on: ubuntu-20.04
+    env:
+      COMPAT: 1
+      VALGRIND: 1
+      BOLTDIR: bolts
+    needs:
+      - prebuild
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.7
+
+      - name: Install dependencies
+        run: |
+          bash -x .github/scripts/setup.sh
+          pip install -U pip wheel poetry
+          # Export and then use pip to install into the current env
+          poetry export -o /tmp/requirements.txt --without-hashes --with dev
+          pip install -r /tmp/requirements.txt
+          # We're going to check BOLT quotes, so get the latest version
+          git clone https://github.com/lightning/bolts.git ../${BOLTDIR}
+
+      - name: Build
+        run: |
+          ./configure
+          make -j $(nproc) check-units installcheck
 
   compile:
     name: Compile CLN ${{ matrix.cfg }}
@@ -166,7 +197,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          echo Nothing to do
+          pip3 install --user pip wheel poetry
+          poetry install
 
 
       - name: Download build
@@ -184,12 +216,7 @@ jobs:
           CFG: ${{ matrix.CFG }}
           SLOW_MACHINE: 1
           PYTEST_PAR: 10
+          TEST_DEBUG: 1
         run: |
-          set -e
-          tar -xvf cln-${CFG}.tar
-
-          pip3 install --user pip wheel poetry
-          poetry export -o requirements.txt --with dev --without-hashes
-          python3 -m pip install -r requirements.txt
-
-          poetry run make pytest
+          tar -xf cln-${CFG}.tar
+          poetry run pytest tests/ -n ${PYTEST_PAR} ${PYTEST_OPTS}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,10 +35,10 @@ jobs:
             EXPERIMENTAL_FEATURES: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
@@ -76,10 +76,10 @@ jobs:
       COMPAT: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
@@ -146,10 +146,10 @@ jobs:
             EXPERIMENTAL_FEATURES: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
@@ -245,10 +245,10 @@ jobs:
             PYTEST_PAR: 3
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
@@ -286,10 +286,10 @@ jobs:
       TEST_CMD: "make -j 8 && pytest -vvv tests/test_cln_rs.py"
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,13 @@ on:
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
-  
+
+env:
+  # Makes the upload-artifact work more reliably at the cost
+  # of a bit of compile time.
+  RUST_PROFILE: release
+  SLOW_MACHINE: 1
+
 jobs:
   prebuild:
     name: Pre-build checks
@@ -138,8 +144,6 @@ jobs:
           CFG: ${{ matrix.CFG }}
         run: |
           set -e
-          export SLOW_MACHINE=1
-
           pip3 install --user pip wheel poetry
           poetry export -o requirements.txt --with dev --without-hashes
           python3 -m pip install -r requirements.txt
@@ -148,11 +152,11 @@ jobs:
           make -j $(nproc) testpack.tar.bz2
 
           # Rename now so we don't clash
-          mv testpack.tar cln-${CFG}.tar
+          mv testpack.tar.bz2 cln-${CFG}.tar.bz2
       - uses: actions/upload-artifact@v3
         with:
-          name: cln-${{ matrix.CFG }}.tar
-          path: cln-${{ matrix.CFG }}.tar
+          name: cln-${{ matrix.CFG }}.tar.bz2
+          path: cln-${{ matrix.CFG }}.tar.bz2
 
   integration:
     name: Test CLN ${{ matrix.cfg }}
@@ -161,6 +165,7 @@ jobs:
       COMPAT: 1
       BITCOIN_VERSION: 0.20.1
       ELEMENTS_VERSION: 0.18.1.8
+      RUST_PROFILE: release  # Has to match the one in the compile step
     needs:
       - compile
     strategy:
@@ -227,7 +232,7 @@ jobs:
       - name: Download build
         uses: actions/download-artifact@v3
         with:
-          name: cln-${{ matrix.CFG }}.tar
+          name: cln-${{ matrix.CFG }}.tar.bz2
 
       - name: Test
         env:
@@ -241,8 +246,8 @@ jobs:
           PYTEST_PAR: 10
           TEST_DEBUG: 1
         run: |
-          tar -xf cln-${CFG}.tar
-          poetry run pytest tests/ -n ${PYTEST_PAR} ${PYTEST_OPTS}
+          tar -xaf cln-${CFG}.tar.bz2
+          poetry run pytest tests/ -vvv -n ${PYTEST_PAR} ${PYTEST_OPTS}
 
   gather:
     # A dummy task that depends on the full matrix of tests, and
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -159,7 +159,7 @@ jobs:
           path: cln-${{ matrix.CFG }}.tar.bz2
 
   integration:
-    name: Test CLN ${{ matrix.cfg }}
+    name: Test CLN ${{ matrix.name }}
     runs-on: ubuntu-22.04
     env:
       COMPAT: 1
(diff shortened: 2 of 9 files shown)


vincenzopalazzo (commented):
Concept Ack.

Another good think about doing in order to speed up the tests is to build the image in stages and run the test on a pre-build docker image.

However, this will required a very deep refactoring, maybe for the future :))

cdecker:
The main problem with stuff running in docker is that you have to store the images in a repository, and docker hub stopped being free a while ago (complicating operations quite a bit).

I think another step I'll attempt to do is to build CLN in a first stage, pass the tarballs over the artifact mechanism in GHA and then test in a separate stage. That'll massively reduce the number of times we build CLN, and cutting down the overall length of reruns.

vincenzopalazzo:
>I think another step I'll attempt to do is to build CLN in a first stage, pass the tarballs over the artifact mechanism in GHA and then test in a separate stage.

Yeah, I was pretty sure that this was possible also with docker image but I never test it so I'm wrong for sure!

>That'll massively reduce the number of times we build CLN, and cutting down the overall length of reruns.

Yeah I think we can zip the time execution in the CI for a lot :) and would be very lovely because Github starts to be slower and slower each time

cdecker:
> > I think another step I'll attempt to do is to build CLN in a first stage, pass the tarballs over the artifact mechanism in GHA and then test in a separate stage.
> 
> Yeah, I was pretty sure that this was possible also with docker image but I never test it so I'm wrong for sure!

It is. As a matter of fact the actions published by Github are all docker images. However, they are storing their own images, whereas we'd have to find a location to store our docker images, when we encapsulate these ourselves. Might be worth looking into this, but might take a bit longer to find a solution.

> > That'll massively reduce the number of times we build CLN, and cutting down the overall length of reruns.
> 
> Yeah I think we can zip the time execution in the CI for a lot :) and would be very lovely because Github starts to be slower and slower each time

Another thing, that is already in there, is segmentation by CI objective, allowing us to unbundle long runnning runs from each other (see lnprototest being split out from integration tests).

cdecker:
Since I spent some time into this PR, I wanted to see how effective it is at saving us CI minutes, and how quickly we get feedback:

```mermaid
gantt
    title New CI waterfall
    dateFormat HH:mm
    axisFormat %H:%M  
    section New
    pre-build :a1, 00:00, 11m55s
    unit-test :after a1, 15m12s
    compile :a2, after a1, 11m44s
    integration :a3, after a2, 49m49s
    integration-valgrind :after a2, 46m3s
    completion :after a3, 1m
```

```mermaid
gantt
    title A Gantt Diagram
    dateFormat HH:mm
    axisFormat %H:%M  
    section Old
    smoke-test :a1, 00:00, 12m1s
    rust-test :after a1, 6m43s
    valgrind-test :after a1, 45m17s
    normal-test :after a1, 39m15s
    postgres-test :after a1, 110m
    doc-check :00:00, 9m37s
```

The old process is slower, but only because the postgres checks are being run slower than necessary (`-n=5` is likely underwhelming the CI runner, and the newer one runs with more parallelism hence the speedup). But the good part is that we get to abort early, and skip a number of things if anything goes wrong:

| Scenario | Old | New | Wasted Runner (old)|
|---|----|---|---|
| Success | ~02:10 | ~01:15| 0 |
|Source error | 00:12 | 00:12 | 1 |
|Unit test error | 02:10 | 00:26| 20 (all tests run in parallel)|
| Doc Error | 09:00 | 00:12 | 1|
| Compilation error | 00:24 | 00:24 | 20|

So the overall effect is that we use way fewer CI runners, despite the end-to-end time not decreasing by too much, we nevertheless have to wait less on CI runners becoming available, which was my main goal for this change.

Going forward we can further speed things up by moving them off of the hot path, so for example we can run various distro builds in a crontab off of `master` and file issues if something breaks, rather than having all possible combinations be listed in a huge matrix.

.github/workflows/ci.yaml
```diff
@@ -7,296 +7,110 @@ on:
...
-            TEST_CMD: "make check-source check-units installcheck check-gen-updated"
-            EXPERIMENTAL_FEATURES: 1
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2.0.0
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v2
```
rustyrussell: These look like accidental reversions?

.github/workflows/ci.yaml
```diff
@@ -7,296 +7,110 @@ on:
...
-            PYTEST_PAR: 3
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v2.0.0
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v2
         with:
```
rustyrussell, line 130: Here too...

.github/workflows/ci.yaml
```diff
@@ -77,10 +77,10 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v2.0.0
+        uses: actions/checkout@v3
 
       - name: Set up Python 3.7
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
```
rustyrussell, line 132: Squash into N-2?  But that's nitpicking...

Makefile
```diff
@@ -803,6 +803,27 @@ install-data: installdirs $(MAN1PAGES) $(MAN5PAGES) $(MAN7PAGES) $(MAN8PAGES) $(
...
+
+# The testpack is used in CI to transfer built artefacts between the
+# build and the test phase. This is necessary because the fixtures in
+# `tests/` explicitly use the binaries built in the current directory
+# rather than using `$PATH`, as that may pick up some other installed
+# version of `lightningd` leading to bogus results. We bundle up all
+# built artefacts here, and will unpack them on the tester (overlaying
+# on top of the checked out repo as if we had just built it in place).
+testpack.tar: $(BIN_PROGRAMS) $(PKGLIBEXEC_PROGRAMS) $(PLUGINS) $(MAN1PAGES) $(MAN5PAGES) $(MAN7PAGES) $(MAN8PAGES) $(DOC_DATA) config.vars $(TESTBINS) $(DEVTOOLS)
+	tar -cvf testpack.tar $^
```
rustyrussell, line 825: Mildly prefer `$@` to repeating `testpack.tar` here.

.github/workflows/ci.yaml
```diff
@@ -6,6 +6,10 @@ on:
       - "master"
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
rustyrussell: TIL!

.github/workflows/ci.yaml
```diff
@@ -138,23 +144,21 @@ jobs:
...
           python3 -m pip install -r requirements.txt
           ./configure CC="$COMPILER"
 
           # We'll need lowdown, not present on ubuntu:20.04 yet
-          make -j $(nproc) doc-all  # Builds lowdown as a dependency
           make -j $(nproc) testpack.tar
 
           # Rename now so we don't clash
-          mv testpack.tar cln-${CFG}.tar
+          bzip2 testpack.tar
```
rustyrussell, line 156: Mildly prefer `make testpack.tar.bz2` and `tar cvfa $@`

.github/workflows/ci.yaml
```diff
@@ -243,8 +248,8 @@ jobs:
           PYTEST_PAR: 10
           TEST_DEBUG: 1
         run: |
-          tar -xf cln-${CFG}.tar
-          poetry run pytest tests/ -n ${PYTEST_PAR} ${PYTEST_OPTS}
+          tar -xvjf cln-${CFG}.tar.bz2
```
rustyrussell, line 251: Use `a` instead of `j`...

.github/scripts/install-bitcoind.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -x
```
rustyrussell, line 1: Change this to:

```
#! /bin/sh

set -e
```

There are no bashisms in here, and you really do want to fail if anything fails!

.github/scripts/install-bitcoind.sh
```diff
@@ -0,0 +1,17 @@
...
+
+DIRNAME="bitcoin-${BITCOIN_VERSION}"
+EDIRNAME="elements-${ELEMENTS_VERSION}"
+FILENAME="${DIRNAME}-x86_64-linux-gnu.tar.bz2"
+EFILENAME="${EDIRNAME}-x86_64-linux-gnu.tar.bz2"
+
+cd /tmp/ || exit
```
rustyrussell, line 8: No longer need `|| exit` with -e.

.github/scripts/install-bitcoind.sh
```diff
@@ -0,0 +1,17 @@
...
+wget "https://storage.googleapis.com/c-lightning-tests/$FILENAME"
+wget -q "https://storage.googleapis.com/c-lightning-tests/${EFILENAME}"
+tar -xjf "${FILENAME}"
+tar -xjf "${EFILENAME}"
```
rustyrussell, line 12: Use `-xaf` instead?

.github/workflows/ci.yaml
```diff
@@ -215,21 +215,7 @@ jobs:
...
-            tar -xjf elements-$ELEMENTS_VERSION-x86_64-linux-gnu.tar.bz2
-            sudo mv bitcoin-$BITCOIN_VERSION/bin/* /usr/local/bin
-            sudo mv elements-$ELEMENTS_VERSION/bin/* /usr/local/bin
-            rm -rf \
-              bitcoin-$BITCOIN_VERSION-x86_64-linux-gnu.tar.gz \
-              bitcoin-$BITCOIN_VERSION \
-              elements-$ELEMENTS_VERSION-x86_64-linux-gnu.tar.bz2 \
-              elements-$ELEMENTS_VERSION
-          )
+        run: bash -x .github/scripts/install-bitcoind.sh
```
rustyrussell, line 231: If you're invoking it this way, you don't need the shebang line in the script.  But use `sh ....` instead of `bash -x ....`?

.github/workflows/ci.yaml
```diff
@@ -301,16 +311,19 @@ jobs:
...
         env:
           COMPAT: 1
           SLOW_MACHINE: 1
           TEST_DEBUG: 1
         run: |
-          tar -xjf cln-${{ matrix.CFG }}.tar.bz2
-          sed -i 's/VALGRIND=0/VALGRIND=1/g' config.vars 
+
+          sed -i 's/VALGRIND=0/VALGRIND=1/g' config.vars
           poetry run pytest tests/ -vvv -n 5 ${PYTEST_OPTS}  ${{ matrix.PYTEST_OPTS }}
```
rustyrussell, line 327: Frobbing config.vars here is a bad smell.... Um, can we do `VALGRIND=1 poetry...`?

tests/test_plugin.py
```diff
@@ -2221,6 +2221,7 @@ def test_htlc_accepted_hook_crash(node_factory, executor):
         f.result(10)
 
 
+@pytest.mark.skip("With newer GCC versions reports a '*** buffer overflow detected ***: terminated'")
```
rustyrussell, line 2224: Wow.... fixed this in #5871.

rustyrussell (approved):
Minor stuff, doesn't all need fixing TBH.

Ack 243d8ba8b7e5b033fb8e8c10276d0a5ba1038073

endothermicdev (approved):
I like the changes - the concurrency feature cancelling an existing run on a new push will itself be a big win.  Also nice to see the memory leak and buffer overflow caught!  I liked @rustyrussell's comments on install-bitcoind.sh (sh vs bash) and explicitly setting VALGRIND=1 in the valgrind test matrix, but pretty minor.  Also, I think the last commit can be dropped now that #5871 has corrected the issue.  Bonus points for the GANTT charts!

ACK 92989dd

.github/workflows/ci.yaml
```diff
@@ -245,3 +245,17 @@ jobs:
...
+    # before merging.
+    name: CI completion
+    runs-on: ubuntu-20.04
+    needs:
+      - integration
+      - check-units
+    steps:
+      - name: Complete
+        run: |
+          echo CI completed successfully
```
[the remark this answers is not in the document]
cdecker: Yep, it'll get marked as failed due to an unsuccessful dependency, and when you restart (either all or just failed) this gets pulled into that set too.

.github/workflows/ci.yaml
```diff
@@ -7,296 +7,110 @@ on:
...
           ./configure
-          make check-doc
-
-  normal-test:
-    name: Normal Test Config ${{ matrix.cfg }}
+          make -j 4 check-source
+          make -j 4 check-units
+          make -j 4 check-gen-updated
+          make -j 4 check-doc
+          make -j 4 installcheck
```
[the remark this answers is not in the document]
cdecker: We could, but we may not be happy with the 2 we get here (4 is just a random value I found worked ok btw), since most tasks are not CPU bound but rather IO-bound it. We can likely tweak these values to get the shortest possible runtime.

I'm also not quite happy that some of these static checks actually end up building the binaries (`installcheck` obviously does). Ideally we'd have very lightweight checks that gate the heavier executions (and I'm counting building the binaries as heavy) and defer any heavy or time-intensive task to later stages.

.github/workflows/ci.yaml
cdecker: Initially we were using envvars in `pytest` exclusively but if I remember correctly at some point we switched to a strange combination of `config.vars` and envvars, so this was the simplest way of making absolutely sure we'd pick the right values for this config.

Happy to simplify if we can :-)

cdecker:
Addressed a couple of comments, and committed fixups. Range-diff [here](https://bot.bitcoinstats.com/ElementsProject/lightning/pull/5685/range_diff/243d8ba8b7e5b033fb8e8c10276d0a5ba1038073..911357107f3373f235341c13b1a0303fc416a5b9), should be autosquashable since the fixups are sorted into their respective position in the stack.

cdecker:
ACtually this [range-diff](https://bot.bitcoinstats.com/ElementsProject/lightning/pull/5685/range_diff/243d8ba8b7e5b033fb8e8c10276d0a5ba1038073..cc70554f06221009a6cd666a5bff5afd6cd5473f) is more useful, since it shows all changes between current HEAD and the reviewed version.

I'm also hunting down some flakyness

.github/workflows/ci.yaml
```diff
@@ -209,162 +139,202 @@ jobs:
...
+      RUST_PROFILE: release  # Has to match the one in the compile step
+      VALGRIND: 1
+      CFG: gcc-dev1-exp1
       DEVELOPER: 1
-      RUST: 1
-      VALGRIND: 0
-      # Run only the rust tests, others are not impacted.
-      TEST_CMD: "make -j 8 && pytest -vvv tests/test_cln_rs.py"
+      EXPERIMENTAL_FEATURES: 1
+      PYTEST_OPTS: --test-group-random-seed=42
```
endothermicdev, line 265: Could you explain the purpose of the random seed? If it's an attempt to achieve better resource balanced between test-groups, I like the concept, but the benefit may be ephemeral as tests are added or as the codebase changes. If this is useful, maybe we could automate this with a cron job to periodically rebalance?  I guess you would quantify time and resource use of each test and then generate random seeds until the balance is relatively equal across test groups.

.github/workflows/ci.yaml
```diff
@@ -209,162 +139,202 @@ jobs:
...
+
+      - name: Test
+        env:
+          COMPAT: 1
+          SLOW_MACHINE: 1
+          TEST_DEBUG: 1
+        run: |
+
+          sed -i 's/VALGRIND=0/VALGRIND=1/g' config.vars
+          poetry run pytest tests/ -vvv -n 5 ${PYTEST_OPTS}  ${{ matrix.PYTEST_OPTS }}
```
endothermicdev, line 326: We previously used PYTEST_PAR=3 here while valgrind was in use. 5 tests might be too ambitious and responsible for the SIGTERM we're getting on test-group 6.  It looks like these were the active tests when it occurs:
tests/test_connection.py::test_funding_close_upfront 
tests/test_pay.py::test_reject_invalid_payload
tests/test_misc.py::test_address
tests/test_pay.py::test_mpp_presplit
tests/test_closing.py::test_invalid_upfront_shutdown_script

endothermicdev (commented):
The changes look great.  However, I notice the CI has been having trouble completing the valgrind runs (group 6 exited with a SIGTERM a couple times.) If we can resolve that, I think this PR will be ready.

vincenzopalazzo (approved):
LGTM!

ACK for me after the SIGTERM fix

.github/workflows/ci.yaml
```diff
@@ -209,162 +139,202 @@ jobs:
...
-          bash -x .github/scripts/setup.sh
+          pip3 install --user pip wheel poetry
+          poetry install
 
-      - name: Build
+      - name: Install bitcoind
+        run: .github/scripts/install-bitcoind.sh
+
+      - name: Download build
+        uses: actions/download-artifact@v3
```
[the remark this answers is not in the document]
cdecker: I think so, I've only seen bad uploads, never bad downloads

.github/workflows/ci.yaml
cdecker: Sure, we can play with the parameters, but it must be the same for all runs. Each group selects it's tests based on this value, and so if we want to run all tests they need to use the same selection criteria.

This can be easily seen when running with `putest-xdist`, i.e., `-n 5` or so. Without specifying the random seed, each process gets to pick it's own, resulting in overlaps in test sets (which xdist catches) but the same issue is true for test groups across multiple ci runners and there we don't have a way to check that tests don't overlap, other than parameterizing them all the same.

.github/workflows/ci.yaml
cdecker: You're right, this might be too aggressive and maybe we're running out of memory causing the fault?

.github/workflows/ci.yaml
```diff
@@ -6,197 +6,127 @@ on:
...
+          # We're going to check BOLT quotes, so get the latest version
+          git clone https://github.com/lightning/bolts.git ../${BOLTDIR}
+      - name: Configure
+        run: ./configure
+      - name: Check source
+        run: make -j 4 check-source
+      - name: Check Generated Files have been updated
+        run: make -j 4 check-gen-updated
+      - name: Check docs
+        run: make -j 4 check-doc
```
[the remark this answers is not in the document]
cdecker: No, as explained here https://github.com/ElementsProject/lightning/pull/5685#discussion_r1062642810 it doesn't make sense to bind the parallelism to the number of CPUs in IO-bound operations. I'd rather experiment with concrete values and hone in on the value that minimizes runtime than just defaulting to the number of processors (we can later parametrize that decision, but we need to find the optimal scale first).

.github/workflows/ci.yaml
```diff
@@ -6,197 +6,127 @@ on:
...
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
 
       - name: Install dependencies
-        run: bash -x .github/scripts/setup.sh
+        run: |
+          bash -x .github/scripts/setup.sh
+          sudo apt-get install -y -qq lowdown
```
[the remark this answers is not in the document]
cdecker: I had a lot of issues compiling `lowdown`, and why should we keep compiling it over and over again in CI? It's a dependency that we compile as a backfill, but we don't want to test `lowdown` do we? So let's take the easy route and just install it from `apt`.

.github/workflows/ci.yaml
cdecker: Reduced to 4, let's see if that works, otherwise I'll reduce it to 3 as well :-)

.github/workflows/prototest.yaml
```diff
@@ -0,0 +1,45 @@
...
+          -e COMPAT=0 \
+          -e PYTEST_PAR=2 \
+          -e PYTEST_OPTS="--timeout=300" \
+          -e TEST_CMD="make check-protos" \
+          -e TEST_GROUP=1 \
+          -e TEST_GROUP_COUNT=1 \
+          cln-ci-ubuntu
+      - name: Upload Unit Test Results
+        if: always()
+        uses: actions/upload-artifact@v2
```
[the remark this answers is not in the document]
cdecker: Amended.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
vincenzopalazzo: yeah, I had the experience where GitHub begin always slower with time.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
vincenzopalazzo: >I had a lot of issues compiling lowdown
 
I had too, I think the issue is inside our makefile.

>why should we keep compiling it over and over again in CI

Do not make any sense, just give us the serenity to be sure that we work correctly when lowdown is not present

>So let's take the easy route and just install it from apt

Agree 👍🏼

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/37e59d50041d1a1d0cdaef42594c91c8a329d1dd

Not sure why it thinks the postgres test has not reported in?

cdecker:
Doh, that is my bad, the required tests in the Repo config goes by name, and after the rework that name no longer matches. It gets added on push, so any auto-resolution doesn't trigger right away, but there is no test behind it that could change the status, so it stays incomplete forever. I'll remove that.

cdecker:
Rebased and removed the required check, so we should get the green checkmark soon :-)

The flow is still a bit flaky, still measuring how many false positives we get, will report back.

@endothermicdev  re: merging, from now it is mostly incremental, so if you can bear with a bit of flakyness we can merge it :-)

cdecker:
> re-ACK https://github.com/ElementsProject/lightning/pull/5685/commits/59113aa8e6a334f7331f4ae9aea30884a56c4152

I'm afraid my bot isn't smart enough to recognize this variant 🥴

endothermicdev:
Excellent. No hurry if you're still putting it through its paces - I'm ready when you're satisfied.
ACK 59113aa

PR #5688, opened 2022-11-03
cln-grpc notification support

orbitalturtle:
This pr does two things:
- It adds support for notifications in msggen so we can generate notification data structures in grpc and rust.
- Adds a rpc for streaming invoicecreation notifications with the grpc rust api. Should be easy to add other notifications later on, but figured it be good to square away an initial one first

### DIFF
--- a/.msggen.json
+++ b/.msggen.json
@@ -512,6 +512,11 @@
             "Invoice.warning_offline": 6,
             "Invoice.warning_private_unused": 8
         },
+        "InvoicecreationnotificationResponse": {
+            "InvoiceCreationNotification.amount_msat": 3,
+            "InvoiceCreationNotification.label": 1,
+            "InvoiceCreationNotification.preimage": 2
+        },
         "KeysendRequest": {
             "KeySend.amount_msat": 10,
             "KeySend.destination": 1,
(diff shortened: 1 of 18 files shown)


cdecker (commented):
Excellent work @orbitalturtle, I know writing code that generates code is always a bit weird, but this works like a charm.

I have added a couple of comments inline in the hopes to make it more extensible in the future. For one I think we should treat notifications as their own entity, and not as a strange variant of methods, and it'd be good to generalize the bookkeeping in the plugin a bit, but consolidating the handling through a single `tokio::sync::broadcast` instead of having one for each notification type.

Happy to discuss tradeoffs though, so don't feel obliged to take my word for it :-)

Otherwise this is looking very good indeed ^^

orbitalturtle:
Hey Christian, thanks for the review on this! Your feedback makes sense to me ~ making Notifications its own object in msggen, etc looks a lot cleaner. I made those changes, whenever you have time for another look.

Couple of questions:
1) I had some trouble with indentation in msggen, so some of the code templating in the python code is indented in weird ways lol, if you have any pointers on that I can try to clean that up
2) I like the idea of adding an "Other" variant. Would this be for if a user wants to subscribe to an event emitted by a plugin or something like that? I'm trying to think through how to create a handler for that, like invoice_handler in /plugins/grpc-plugin/src/main.rs. Is there a way to subscribe on the plugin builder for "other" notifications, and add a handler for that, or what might be the best way to go about doing that?

PR #5697, opened 2022-11-07
lightningd: Add `signinvoice` to sign a BOLT11 invoice. 

dongcarl:
```
Though there's already a `createinvoice` command, there are usecases where a
user may want to sign an invoice that they don't yet have the preimage to. For
example, they may have an htlc_accepted plugin that pays to obtain the preimage
from someone else and returns a `{ "result": "resolve", ... }`.

This RPC command addresses this usecase without overly complicating the
semantics of the existing `createinvoice` command.

Changlog-Added: JSON-RPC: `signinvoice` new command to sign BOLT11
invoices.
```

Also adds a commit to correct `createinvoice`'s `invstring` description. (See the commit message for more details)

One main question for reviewers is whether this PR is acceptable without support for BOLT12 invoices. I'm not familiar enough with them to write code for them. Perhaps it can be left for a followup PR? If so, do we need to make the `bolt11` field in `doc/schemas/signinvoice.schema.json` non-required?

CC: @cdecker

### DIFF
--- a/doc/schemas/signinvoice.request.json
+++ b/doc/schemas/signinvoice.request.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "added": "v23.02",
+  "required": [
+    "invstring"
+  ],
+  "properties": {
+    "invstring": {
+      "type": "string",
+      "description": ""
+    }
+  }
+}
--- a/doc/schemas/signinvoice.schema.json
+++ b/doc/schemas/signinvoice.schema.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "bolt11"
+  ],
+  "properties": {
+    "bolt11": {
+      "type": "string",
+      "description": "the bolt11 string"
+    }
+  }
+}
--- a/lightningd/invoice.c
+++ b/lightningd/invoice.c
@@ -1897,3 +1897,57 @@ static const struct json_command preapprovekeysend_command = {
 	"Ask the HSM to preapprove a keysend payment."
 };
 AUTODATA(json_command, &preapprovekeysend_command);
+
+static struct command_result *json_signinvoice(struct command *cmd,
+						 const char *buffer,
+						 const jsmntok_t *obj UNNEEDED,
+						 const jsmntok_t *params)
+{
+	const char *invstring;
+	struct json_stream *response;
+	struct bolt11 *b11;
+	struct sha256 hash;
+	const u5 *sig;
+	bool have_n;
+	char *fail;
+
+	if (!param(cmd, buffer, params,
+		   p_req("invstring", param_string, &invstring),
+		   NULL))
+		return command_param_failed();
+
+	b11 = bolt11_decode_nosig(cmd, invstring, cmd->ld->our_features,
+				  NULL, chainparams, &hash, &sig, &have_n,
+				  &fail);
+
+	if (!b11)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "Unparsable invoice '%s': %s",
+				    invstring, fail);
+
+	/* This adds the signature */
+	char *b11enc = bolt11_encode(cmd, b11, have_n,
+				     hsm_sign_b11, cmd->ld);
+
+        /* BOLT #11:
+         * A writer:
+         *...
+         *    - MUST include either exactly one `d` or exactly one `h` field.
+         */
+	if (!b11->description && !b11->description_hash)
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "Missing description in invoice");
+
+	response = json_stream_success(cmd);
+	json_add_invstring(response, b11enc);
+	return command_success(cmd, response);
+}
+
+static const struct json_command signinvoice_command = {
+	"signinvoice",
+	"payment",
+	json_signinvoice,
+	"Lowlevel command to sign invoice {invstring}."
+};
+
+AUTODATA(json_command, &signinvoice_command);
--- a/tests/test_invoices.py
+++ b/tests/test_invoices.py
@@ -533,6 +533,19 @@ def test_waitanyinvoice(node_factory, executor):
         l2.rpc.waitanyinvoice('non-number')
 
 
+def test_signinvoice(node_factory, executor):
+    # Setup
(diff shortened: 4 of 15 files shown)


lightningd/invoice.c
```diff
@@ -1788,3 +1788,56 @@ static const struct json_command createinvoice_command = {
...
+	if (!param(cmd, buffer, params,
+		   p_req("invstring", param_string, &invstring),
+		   NULL))
+		return command_param_failed();
+
+	b11 = bolt11_decode_nosig(cmd, invstring, cmd->ld->our_features,
+				  NULL, chainparams, &hash, &sig, &have_n,
+				  &fail);
+
+	if (!b11)
```
vincenzopalazzo, line 1814: >One main question for reviewers is whether this PR is acceptable without support for BOLT12 invoices. I'm not familiar enough with them to write code for them. Perhaps it can be left for a follow-up PR?

```suggestion
        /* FIXME: return an ad-hok error message when the invstring is a BOLT 12 */
	b11 = bolt11_decode_nosig(cmd, invstring, cmd->ld->our_features,
				  NULL, chainparams, &hash, &sig, &have_n,
				  &fail);

	if (!b11)
```

I think we should make a sanity check if you do not want to support BOL12 here

doc/schemas/signinvoice.schema.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "bolt11"
+  ],
```
vincenzopalazzo, line 7: The bolt12 invoice born already signed iirc, so maybe here the options can be:

- Support only the bolt11 and return an error in case of bolt12;
- Support both, and return a warning in case of bolt12 (and if I'm not wrong regarding the bolt12 invoice) that tells that the invoice is already signed. This warning could be useful also in case of bolt11 is already signed?

doc/schemas/signinvoice.request.json
```diff
@@ -0,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "invstring"
```
vincenzopalazzo, line 6: This is not related to this PR but maybe we should write down some naming convention about the string invoice to have the same key around different API 

See https://lightning.readthedocs.io/lightning-decode.7.html

tests/test_invoices.py
```diff
@@ -533,6 +533,19 @@ def test_waitanyinvoice(node_factory, executor):
         l2.rpc.waitanyinvoice('non-number')
 
 
+def test_signinvoice(node_factory, executor):
+    # Setup
```
vincenzopalazzo, line 537:
```suggestion
    """ Make sure that the signinvoice work as expected while re-sign a previous invoice """
    # Setup
```

This is a personal preference, that I find useful in case of refactoring the test in the future

vincenzopalazzo (commented):
Concept ACK

I added some comments that include  some of my personal conventions, also if I'm still not able to catch the use case that this rpc enable, to me makes sense to provide an API to plugins to do somethings that it is not possible today/

Thanks

cdecker:
> One main question for reviewers is whether this PR is acceptable without support for BOLT12 invoices. I'm not familiar enough with them to write code for them. Perhaps it can be left for a followup PR? If so, do we need to make the `bolt11` field in `doc/schemas/signinvoice.schema.json` non-required?

I think we can side-step this issue by calling it `unsigned`, `string`, or `raw`, i.e., not referencing a specific format for now, but keeping it mandatory.

To deduplicate code it might be useful to make `createinvoice` an alias of `signinvoice` (the latter being more memorable imho) and adding optional `preimage` and `save` flags to check the hash is correct, and whether to save the invoice in the DB (requires `preimage` so we can terminate the payment as well).

Wdyt?

rustyrussell:
We just merged `createinvoicerequest` with a `savetodb` parameter (default true).  This could be made to match, though some work would be needed since json_add_invoice_fields assumes a db entry right now (fake one up, or make that take more parameters).

Obv preimage would be optional iff savetodb was false.

dongcarl:
Reading through the comments... Is the general suggestion to instead modify `createinvoice` such that it looks like:

```
**createinvoice** *invstring* *label* [*preimage*] [*savetodb*]
```

?

It seems like this would require me to:
1. Modify `json_add_invoice_fields` to take more parameters
2. Add full support for bolt12

Which is not un-doable but expands on the scope of the PR.

Am I understanding you all correctly? @vincenzopalazzo @cdecker @rustyrussell

dongcarl:
On the open call yesterday @rustyrussell mentioned that he wanted me to just modify the `createinvoicerequest` call to accept bolt11 invoices, however, it seems like `createinvoicerequest` deals with `tlv_invoice_request` structs while `createinvoice` deals with `tlv_invoice` structs, is that okay? What is the difference between an invoice struct and an invoice request struct?

rustyrussell:
On further consideration, this would be a modification to createinvoice, but the API would be ugly: no label, no preimage.  So I take it back, this API is cleaner.

rustyrussell:
Fairly trivial rebase, and two fixups for you to consider?  Sorry for the delay!

lightningd/invoice.c
```diff
@@ -1897,3 +1897,57 @@ static const struct json_command preapprovekeysend_command = {
...
+		return command_fail(cmd, JSONRPC2_INVALID_PARAMS,
+				    "Unparsable invoice '%s': %s",
```
vincenzopalazzo, line 1925:
```suggestion
				    "Invalid invoice '%s': %s",
```

More user friendly message?

vincenzopalazzo (approved):
happy to see the CI up and running again!

Concept Ack, I have only a tiny comment on the error message!

lightningd/invoice.c
dongcarl: This is the same message as in `json_createinvoice` so I think we'd want to keep it consistent?

vincenzopalazzo: Oh yes, sorry I did not note it. So ignore this, we can improve in another moment

rustyrussell:
Typo in commit message, "Changlog"!  Other than that, it's looking great!

rustyrussell:
Ack https://github.com/ElementsProject/lightning/pull/5697/commits/813a93e04131234cb95ff0e74f77870791d4a1db

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5697/commits/813a93e04131234cb95ff0e74f77870791d4a1db

PR #5698, opened 2022-11-08
Handle non-standard onion reply size

rustyrussell:
Turns out we *didn't* handle this correctly, sorry Joost!

See https://github.com/lightning/bolts/pull/1021


cdecker (approved):
ACK f15613ae18126770d7ced173d770254a8b79dd95

PR #5701, opened 2022-11-09
Weekly @cdecker cleanups for 2022/45

### LINKED ISSUE #5693: Build fails with GNU make 4.4: No rule to make target 'FORCE'
Using **GNU Make 4.4**, the build of `v0.12.1` (and `master` at 0d756ff017afa0eee577990e07247772d7f29221) fails with:
```
make: *** No rule to make target 'FORCE', needed by 'wallet/db_sqlite3_sqlgen.c'.  Stop.
```

This may be due to a regression in Make, but it's easy to work around: simply define `FORCE` as a normal rule, not a double-colon rule. It's unclear why it was defined as a double-colon rule in the first place since it's only defined once.

Please change:

    FORCE::

to:

    FORCE:

With that change, all is well with GNU Make 4.4.

cdecker:
Just some drive-by fixes.

Fixes #5693

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.65"
+version = "1.0.66"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "98161a4e3e2184da77bb14f02184cdd111e83bbbcc9979dfee3c44b9a85f5602"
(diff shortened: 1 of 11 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5701/commits/380acdcb594fd93f60ee7ca7c86bfb80818c0415

cdecker:
Getting things to work correctly on PPA is really painful, hence the slow trickle of changes, no point in reviewing yet.

Issue #5576, opened 2022-09-07
listconfigs shows only 80 chars of "lightning-dir" path


SimonVrouwe:
On latest master (c4203e7de6f35ed1cd1658d7c898b8b6bfaac2df), `test_relative_config_dir` fails when `lightning-dir` directory path is longer then 80 characters.

Accidentally ran into this when using the $TEST_DIR variable for the pytests, below example results in `--lightning-dir` path to be 81 chars.

```
TEST_DIR=/tmp/aaaaaaaaaaaaaaaaaa VALGRIND=0 pytest tests/ -k 'test_relative_config_dir'
============================================================================================================================= test session starts =============================================================================================================================
platform linux -- Python 3.9.2, pytest-7.1.2, pluggy-1.0.0
rootdir: /home/simon/git/SimonVrouwe/lightning, configfile: pytest.ini
plugins: custom-exit-code-0.3.0, test-groups-1.0.3, xdist-2.5.0, forked-1.4.0, timeout-2.1.0
collected 528 items / 527 deselected / 1 selected                                                                                                                                                                                                                             

tests/test_misc.py F                                                                                                                                                                                                                                                    [100%]

================================================================================================================================== FAILURES ===================================================================================================================================
__________________________________________________________________________________________________________________________ test_relative_config_dir ___________________________________________________________________________________________________________________________

node_factory = <pyln.testing.utils.NodeFactory object at 0x7fdd97606d60>

    def test_relative_config_dir(node_factory):
        l1 = node_factory.get_node(start=False)
        initial_dir = os.getcwd()
        lndir = l1.daemon.opts.get("lightning-dir")[:-1]
        *root_dir, l1.daemon.opts["lightning-dir"] = lndir.split('/')
        os.chdir('/'.join(root_dir))
        l1.daemon.executable = os.path.join(initial_dir, l1.daemon.executable)
        l1.start()
>       assert os.path.isabs(l1.rpc.listconfigs()["lightning-dir"])
E       AssertionError: assert False
E        +  where False = <function isabs at 0x7fdd9d7cf820>('tmp/aaaaaaaaaaaaaaaaaa/ltests-nhzau93b/test_relative_config_dir_1/lightning-1...')
E        +    where <function isabs at 0x7fdd9d7cf820> = <module 'posixpath' from '/usr/lib/python3.9/posixpath.py'>.isabs
E        +      where <module 'posixpath' from '/usr/lib/python3.9/posixpath.py'> = os.path

[...]

PR #5703, opened 2022-11-10
fix: Raise Error for long TEST_DIR

kevkevinpal:
I added an if statment after we define our test dir path to check if it is larger than 12 characters because otherwise it was throwing an error

this closes: https://github.com/ElementsProject/lightning/issues/5576

### DIFF
--- a/contrib/pyln-testing/pyln/testing/fixtures.py
+++ b/contrib/pyln-testing/pyln/testing/fixtures.py
@@ -25,6 +25,8 @@ __attempts: Dict[str, int] = {}
 @pytest.fixture(scope="session")
 def test_base_dir():
     d = os.getenv("TEST_DIR", "/tmp")
+    if len(d) > 12:
+        raise ValueError("TEST_DIR length is too long please create one with less than x characters")
 
     directory = tempfile.mkdtemp(prefix='ltests-', dir=d)
     print("Running tests in {}".format(directory))


contrib/pyln-testing/pyln/testing/fixtures.py
```diff
@@ -25,6 +25,8 @@
 @pytest.fixture(scope="session")
 def test_base_dir():
     d = os.getenv("TEST_DIR", "/tmp")
+    if len(d) > 12:
+        raise ValueError("TEST_DIR length is too long please create one with less than x characters")
```
vincenzopalazzo, line 29: do you think that this is a hacky approach?

Maybe the solution is to fix the opts parser?

cc: @javiervargas that is working on a C solution

kevkevinpal: ya its a bit hacky would you be able to point me to the opts parser? @javiervargas do you have a branch maybe I could look at?

kevkevinpal: is this where I should be looking?

https://github.com/ElementsProject/lightning/blob/master/lightningd/options.c#L1420

javiervargas:
I've been working on in two possibles solutions.   One could be increase this variable (after compile tests worked well)
[https://github.com/javiervargas/lightning/commit/77079bd376cacf52c7f31d0dc0189b125240e284](url)

`
/* FIXME: */
#define OPT_SHOW_LEN_BIG 128`

But here,  Maybe rusty could precise if this is a good approach.

Or second create a new method that prints arbitrary lenghts.  I'm finishing that code.   Thanks to @vincenzopalazzo 

Issue seems to be here.  

[https://github.com/ElementsProject/lightning/blob/c4203e7de6f35ed1cd1658d7c898b8b6bfaac2df/lightningd/options.c#L1568-L1569](url)

rustyrussell:
BTW, I'm closing this.  The options parser should now be fixed?

vincenzopalazzo:
yeah it is alredy fixed

PR #5707, opened 2022-11-11
doc: Escape non-formatting underscores in Markdown for man pages

whitslack:
The only time underscores aren't special in Markdown is when they appear in preformatted text. We have gotten away with not escaping underscores where an asterisk-enclosed span or the paragraph ends before the next underscore appears, but this is fragile and bad practice. Conversely, there are many places where we have not escaped underscores but needed to.
    
Escape all underscores that do not appear in preformatted blocks or preformatted spans and are not themselves delineating emphasized spans.

Also, add a check in `check-manpages` to help prevent more naughty underscores from creeping in in the future.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -184,6 +184,7 @@ check: check-manpages
 check-manpages: all-programs check-config-docs default-targets
 	@tools/check-manpage.sh cli/lightning-cli doc/lightning-cli.1.md
 	@tools/check-manpage.sh "lightningd/lightningd --lightning-dir=/tmp/" doc/lightningd-config.5.md
+	@awk '/^$$/ { do { getline } while ($$0 ~ /^( {4,}|\t)/) } /^\s*```/ { do { getline } while ($$0 !~ /^\s*```/) } /^([^`_\\]|`([^`\\]|\\.)*`|\b_|_\b|\\.)*\B_\B/ { print "" ; print "Unescaped underscore at " FILENAME ":" NR ":" ; print ; ret = 1 } ENDFILE { NR = 0 } END { exit ret }' doc/*.[0-9].md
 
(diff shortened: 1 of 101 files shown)


cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5707/commits/328e4d194d89c6d0a2f628c733b6a772f7d6fa18

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5707/commits/328e4d194d89c6d0a2f628c733b6a772f7d6fa18

SimonVrouwe:
It looks ok to me, but what exactly does

>Also, add a check in check-manpages

when I modify a single markdown file, for example in `/doc/lightning-fetchinvoice.7.md` by changing "msatoshi" to "amount_msat"
```diff
diff --git a/doc/lightning-fetchinvoice.7.md b/doc/lightning-fetchinvoice.7.md
index fce644ef5..8661ea74b 100644
--- a/doc/lightning-fetchinvoice.7.md
+++ b/doc/lightning-fetchinvoice.7.md
@@ -6,7 +6,7 @@ SYNOPSIS
 
 **(WARNING: experimental-offers only)**
 
-**fetchinvoice** *offer* [*msatoshi*] [*quantity*] [*recurrence\_counter*] [*recurrence\_start*] [*recurrence\_label*] [*timeout*] [*payer\_note*]
+**fetchinvoice** *offer* [*amount_msat*] [*quantity*] [*recurrence\_counter*] [*recurrence\_start*] [*recurrence\_label*] [*timeout*] [*payer\_note*]
 
 DESCRIPTION
 -----------
```
an then run `make check-manpages` it rebuilds everything, which seems weird for above change.

And then run `make check-manpages` again does nothing, but also shows no error.

Issue #5695, opened 2022-11-05
We should double check all addresses we use


rustyrussell:
### Issue and Steps to Reproduce
See https://github.com/lightningnetwork/lnd/issues/7022#issuecomment-1278695682

We need to check that we can create signatures for each one.

PR #5708, opened 2022-11-11
Extra address checks, with new HSM support

rustyrussell:
This is built on top of #5986 (first 9 commits)
Closes: #5695


cdecker:
Ugh, more hsm_wire.csv breaking. Let's not create all those variants of the init message, let's rather move the HSM version up to in the init, and then use that to differentiate. Replicating the `init` message is just weird, and not really helping all that much. It's better to bundle changes so we break less often, but at least we do it cleanly and the `hsmd` can detect it based on the version in the `init` message.

cdecker (changes requested):
Replicating the `init_reply` for each type is weird, a simple version number in the actual `init_reply` followed by optional fields (TLV?) is better backwards compatible imho.

Also longer term we should try to bundle changes to the hsm wire protocol better. I'll squeeze this into v22.11 just because I don't want the signers breaking again in 2 months time.

cdecker:
Thinking about this a bit more, I think we have two options to make the schema evolution for the `hsmd` more practical:

 - Move the `hsmd_version` to the first position in the `init_reply` message, allowing us to peek at it before decoding, and then allowing extensions to the message to be made at the end. This allows backwards compatibility for the case when `hsmd_version >= node_version`, since the `fromwire_` methods can underread the message, ignoring trailing fields.
   - One choice to be made is which part should detect and abort on version incompatibilities: the node can set a max version, however that defeats the signer backward compatibility. Alternatively the signer can abort if the node version can't be satisfied (which is in line with the signer being backward compatible)
 - The other option is to go with TLVs, allowing us to omit fields that have been removed from the schema without having to introduce an incompatibility because the node can no longer read the message because of missing positional fields.

I guess it boils down to whether we expect the signer or the node to be ahead, and whether we care about pruning old fields from the message.

As a quick side-note on why I don't like changing the numeric type of `init_reply`: when attempting to start the node without the signer present (remote signer setup), we need to cache the `init_reply` which is static anyway, in order to be able to start at all. Changing the type means we now need additional logic to recognize a new type and cache that instead, and on upgrades we need to bump the signer and node version in lock-step.

niftynei:
Patch lgtm, utACK.

But I did have a question about the goals of API support. Mainly, how long is one's HSMD impl being out of date supported for? If a user has an HSM with a terribly outdated version from the current CLN version they're running, will it fail to start or is there supposed to be a "limp along"/backwards compatible level of functionality that is maintained?

Just trying to get a better understanding of the design goals re HSM API here.

rustyrussell:
Some of this is a limitation of our wire format. It's simplest and most robust to have separate messages, but we can actually make the v3 a variant on v2 with an extra field appended since we have option logic for that (which off the top of my head, isn't used anywhere else anymore, so I would have liked to remove it, but ok).

Using TLVs here unconstrains the test matrix and is probably a bad idea. The multiple init responses is ugly, but simple. Putting the version first is a good idea, in the hope that all messages can be based on this in future.

ksedgwic:
Looks sensible to me, concept ACK
nit: a comment explaining the implicit v3 semantics might be helpful.

A capability which seems to be coming up is `option_anchors_zero_fee_htlc_tx` support; VLS would like to disallow plain anchor commitments ...
@devrandom

hsmd/hsmd_wire.csv
```diff
@@ -323,3 +323,11 @@ msgdata,hsmd_derive_secret,info,u8,len
 # Reply with the derived secret
 msgtype,hsmd_derive_secret_reply,127
 msgdata,hsmd_derive_secret_reply,secret,secret,
+
+# Sanity check this pubkey derivation is correct.
+msgtype,hsmd_check_pubkey,28
+msgdata,hsmd_check_pubkey,index,u32,
+msgdata,hsmd_check_pubkey,pubkey,pubkey,
+
+# Reply (if it's bad, it simple aborts.)
```
devrandom, line 332: what does this mean exactly?  closes the connection?  it would be difficult to distinguish from a real disconnect due to restart.

hsmd/hsmd_wire.csv
```diff
@@ -323,3 +323,11 @@ msgdata,hsmd_derive_secret,info,u8,len
 # Reply with the derived secret
 msgtype,hsmd_derive_secret_reply,127
 msgdata,hsmd_derive_secret_reply,secret,secret,
+
+# Sanity check this pubkey derivation is correct.
+msgtype,hsmd_check_pubkey,28
+msgdata,hsmd_check_pubkey,index,u32,
```
devrandom, line 329: - are hardened indices allowed (high bit = 1)?
- we plan to deny very high indices, e.g. > 1_000_000, to make recovery sane in case of database loss.  is that OK?

devrandom (commented):
utACK on the versioning scheme.  a couple of small questions on the specific feature added here.

devrandom:
BTW, note that we check L1 destinations at the time of `SignWithdrawal` anyway.  I assume that we are passed the derivation index at that time in the PSBT (@ksedgwic ?).

ksedgwic:
Yes, `Wallet::can_spend` confirms that the index generates the correct script_pubkey

hsmd/hsmd_wire.csv
rustyrussell: OK, this works for a local hsmd, not for remote.  Will make it an explicit reply!

Thanks!

hsmd/hsmd_wire.csv
rustyrussell: No, we don't use hardened derivations.  And the 1M keys limit seems like a sane precaution: if someone exceeds that they should tell us :)

rustyrussell:
Folded in fix which makes hsm explicitly reply, and fixed up hsm_versions.h definition.

rustyrussell:
Trivial rebase on top of master now #5986 is merged.

Ack https://github.com/ElementsProject/lightning/pull/5708/commits/a3a720c70cfb3bbb70232938fc0598bb1b9346a0

rustyrussell:
Ack https://github.com/ElementsProject/lightning/commit/cf04060f28ed7d43e20be82e44c96ba63767e7bc

PR #5709, opened 2022-11-11
docs: remove the table inside the reprobuild docs 

vincenzopalazzo:
This is a bit of a rabbit hole but the table looks like not supported in the docs site https://lightning.readthedocs.io/REPRODUCIBLE.html we should add additional dependencies, but I find the table is not accessible at all so I'm proposing a new list format.

In addition, other small commits to introduce the manpage for reckless and fix a small typo found in my emacs spellchecker

### DIFF
--- a/doc/REPRODUCIBLE.md
+++ b/doc/REPRODUCIBLE.md
@@ -79,7 +79,7 @@ currently support:
   - Distribution Version: 22.04
   - Codename: jammy
 
-Depending on your host OS release you migh not have `debootstrap`
+Depending on your host OS release you might not have `debootstrap`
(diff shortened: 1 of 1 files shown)


vincenzopalazzo:
I forget that we do not include the manpage, so I'm including reckless inside the ignored one!

rustyrussell (approved):
Ack 95cad3ea341fae3582906f0737b408cf7a36683a

rustyrussell:
Removed last commit as redundant (.gitignore fixed in master).

Ack d3a37775efeacc72788b1a467b2ee0d3faabddd4

PR #5715, opened 2022-11-15
ld: Use `htable` to detect own transactions

### LINKED ISSUE #5682: Long periods of unresponsiveness every time a block is added
```
$ gnuplot --persist 3< <(grep -Fa -A1 'Adding block' /var/log/lightningd.log | awk '
    match($0, /^([0-9]{4})-([0-9]{2})-([0-9]{2})T([0-9]{2}):([0-9]{2}):([0-9]{2})\.[0-9]{3}Z/, m) {
        time = mktime(m[1]" "m[2]" "m[3]" "m[4]" "m[5]" "m[6]" 0", 1)
    }
    start { print time - start ; start = 0 }
    /Adding block/ { start = time }
') <<END
set style histogram rowstacked gap 0
set style fill solid 0.5 border lt -1
set xlabel "Duration of quiescence (seconds)"
set ylabel "Number of blocks"
binwidth = 5
set boxwidth binwidth
bin(x) = binwidth * (floor(x / binwidth) + 0.5)
plot "/proc/self/fd/3" using (bin($1)):(1.0) smooth freq with boxes notitle
END
```
chart

This histogram shows the elapsed time from each "Adding block" log message until the immediately succeeding message in the log. My node is very busy and normally spews many pages of log messages every second, but whenever a block is found, the whole world grinds to a standstill, often for upwards of two minutes. During these standstills, all RPC calls through `lightning-cli` hang, and they only return after block processing has finished. I strongly suspect that all HTLC additions and removals in channels are also being deferred until after block processing has finished. If so, this implies a rather huge thorn in the user experience for any Lightning users trying to route payments through CLN nodes. Indeed, I suspect a significant fraction of routing delays in the Lightning Network may be directly attributable to this issue in CLN. I have personally experienced slow payments from my Phoenix wallet in periods immediately following when a new block has been found.

What in the world is `lightningd` doing that takes so darn long whenever a block is found, and why isn't the work being done by a subdaemon? This has an odor reminiscent of doing work in the UI thread in a graphical application. Always do work in a background worker thread, and then post results back to the UI thread when finished. I would think an equivalent notion for CLN, its being a multi-process, single-threaded application, would be to have a subdaemon (`topology`?) do the work and post its findings back to `lightningd`.

### `getinfo` output
I am running [`v0.11.2-gentoo-r107`](https://github.com/whitslack/lightning/tree/v0.11.2-gentoo-r107).

cdecker:
@whitslack did an in-depth analysis of the issue at hand, and it turns out we were doing quadratic lookups on an ever increasing list of transactions we broadcast, with ever rebroadcast doubling the size of the rebroadcast list.

The first commit avoids adding the same TX multiple times, while the following two commit replace the entire list with an `htable` instead, reducing complexity to linear.

Fixes #5682

### DIFF
--- a/lightningd/chaintopology.c
+++ b/lightningd/chaintopology.c
@@ -53,13 +53,7 @@ static void next_topology_timer(struct chain_topology *topo)
 static bool we_broadcast(const struct chain_topology *topo,
 			 const struct bitcoin_txid *txid)
 {
-	const struct outgoing_tx *otx;
-
-	list_for_each(&topo->outgoing_txs, otx, list) {
-		if (bitcoin_txid_eq(&otx->txid, txid))
-			return true;
-	}
-	return false;
+	return outgoing_tx_map_get(&topo->outgoing_txs, txid) != NULL;
 }
 
 static void filter_block_txs(struct chain_topology *topo, struct block *b)
@@ -160,13 +154,16 @@ static void rebroadcast_txs(struct chain_topology *topo)
 	/* Copy txs now (peers may go away, and they own txs). */
 	struct txs_to_broadcast *txs;
 	struct outgoing_tx *otx;
+	struct outgoing_tx_map_iter it;
 
 	txs = tal(topo, struct txs_to_broadcast);
 	txs->cmd_id = tal_arr(txs, const char *, 0);
 
 	/* Put any txs we want to broadcast in ->txs. */
 	txs->txs = tal_arr(txs, const char *, 0);
-	list_for_each(&topo->outgoing_txs, otx, list) {
+
+	for (otx = outgoing_tx_map_first(&topo->outgoing_txs, &it); otx;
+	     otx = outgoing_tx_map_next(&topo->outgoing_txs, &it)) {
 		if (wallet_transaction_height(topo->ld->wallet, &otx->txid))
 			continue;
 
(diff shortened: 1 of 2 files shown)


cdecker:
This is a rather simple PR, however there is one issue that I could not fix on the fly:

https://github.com/ElementsProject/lightning/pull/5715/files#diff-f13ab0f51e9574e49798cd32c0cf2a2476bc1dbd01d1aa31eb4ff9c1d6df6d94R218

Notice the `notleak` here to avoid shutdowns falsely detecting leaked `outgoing_tx` instances, which I don't quite get where these are from. The loop in the destructor should ensure we free all the queued instances, but it seems not to quite get them all. Could this be a concurrent modification of the `htable` (via the destructor added to `outgoing_tx` instances)?

whitslack:
> Could this be a concurrent modification of the `htable` (via the destructor added to `outgoing_tx` instances)?

Hmm, but you're removing the destructor before freeing each `outgoing_tx` in the map. Assuming that's working correctly, there shouldn't be any concurrent modification of the `htable` during the iteration over it.

cdecker:
Yep, that was an attempt to avoid this issue, but it didn't work, which is why I'm asking here :-)

PR #5717, opened 2022-11-17
cln-plugin: Make the configuration in `init` public

cdecker:
It was set to crate level visibility for some reason, not all that helpful :-)

Suggested-by: Sergi Delgado Segura <@sr-gi>

### DIFF
--- a/plugins/src/messages.rs
+++ b/plugins/src/messages.rs
@@ -58,7 +58,7 @@ pub(crate) enum Notification {
 }
 
 #[derive(Deserialize, Debug)]
-pub struct GetManifestCall {}
+pub(crate) struct GetManifestCall {}
 
 #[derive(Deserialize, Debug)]
 pub(crate) struct InitCall {
@@ -93,7 +93,7 @@ pub struct ProxyInfo {
 }
 
 #[derive(Debug)]
-pub enum JsonRpc<N, R> {
+pub(crate) enum JsonRpc<N, R> {
     Request(serde_json::Value, R),
     Notification(N),
     CustomRequest(serde_json::Value, Value),
--- a/plugins/src/messages.rs
+++ b/plugins/src/messages.rs
@@ -63,7 +63,7 @@ pub struct GetManifestCall {}
 #[derive(Deserialize, Debug)]
 pub(crate) struct InitCall {
     pub(crate) options: HashMap<String, Value>,
-    pub(crate) configuration: Configuration,
+    pub configuration: Configuration,
 }
 
 #[derive(Clone, Deserialize, Debug)]
--- a/plugins/src/codec.rs
+++ b/plugins/src/codec.rs
@@ -12,7 +12,7 @@ use std::{io, str};
 use tokio_util::codec::{Decoder, Encoder};
 
 use crate::messages::{Notification, Request};
-pub use crate::messages::JsonRpc;
+use crate::messages::JsonRpc;
 
 /// A simple codec that parses messages separated by two successive
 /// `\n` newlines.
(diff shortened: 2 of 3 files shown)


plugins/src/messages.rs
```diff
@@ -63,7 +63,7 @@ pub struct GetManifestCall {}
 #[derive(Deserialize, Debug)]
 pub(crate) struct InitCall {
     pub(crate) options: HashMap<String, Value>,
```
vincenzopalazzo, line 65:
```suggestion
    pub options: HashMap<String, Value>,
```

We should make pub also this? or there is some API like `get_opt`

cdecker: Not sure, since we have accessors that return typed option values: https://github.com/ElementsProject/lightning/blob/master/plugins/src/lib.rs#L399

Then again, people might want to iterate over all options, so maybe there's no harm in providing a redundant way to access them?

sr-gi:
I may be wrong here, but I think this does not solve the issue.

`cln_plugin::messages` is private, so the types defined there cannot be used elsewhere. This would also include `Configuration` even if it is now defined as `pub` instead of `pub(crate)`, won't it?

My original request was being able to store a variable of type `ProxyInfo` (which is also defined in that module).

cdecker:
Ok, in that case we need to make messages pub. Sorry my brain is quite foggy at the moment due to a cold 🙂

sr-gi:
> Ok, in that case we need to make messages pub. Sorry my brain is quite foggy at the moment due to a cold 🙂

No worries. I don't have my laptop with me ATM so I was wondering if I was missing something 😅

Hope you get better @cdecker

cdecker:
Ok, I went ahead and adjusted the visibility of `cln_plugin::messages` and since there were a couple of ugly internals being exposed to users, I also reduced visibility of things I'd like to keep hidden from users (`codec` and the use of `anyhow` for example).

plugins/src/messages.rs
vincenzopalazzo: >Then again, people might want to iterate over all options, so maybe there's no harm in providing a redundant way to access them?

I do not have any strong feeling about it, let give a run period and if someone need it we make it public?

vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5717/commits/9757db003dea4c2374e32ade50d7e00d30109cbe

the CI looks only unlucky!

sr-gi:
> Ok, I went ahead and adjusted the visibility of `cln_plugin::messages` and since there were a couple of ugly internals being exposed to users, I also reduced visibility of things I'd like to keep hidden from users (`codec` and the use of `anyhow` for example).

I tested this locally and `messages` is now accessible! There is one breaking change though. The interface for rpc methods (and hooks) is `Result<serde_json::Value, anyhow::Error>` if I'm not mistaken, however, `anyhow` is not accessible anymore.

Shouldn't a method (or macro) to generate errors be accessible if the plugin scaffolding is going to be using `Error = anyhow::Error`? Otherwise, any plugin will need to add `anyhow` as a dependency.

cdecker:
Correct, we used to expose `anyhow::Error` as the error type that is returned by `cln_plugin`, whereas we now have a public type alias:

https://github.com/ElementsProject/lightning/blob/9757db003dea4c2374e32ade50d7e00d30109cbe/plugins/src/lib.rs#L32

So now you can use `cln_plugin::Error` which is a type under our control, and I want to eventually migrate to `thiserror` to have strongly typed errors, and not just strings in the form of `anyhow`. Hence the removal of the re-export.

cdecker:
By the way, we're not bound to publish these crates in sync with the CLN versions, so happy to amend outside of the release windows too :-)

PR #5721, opened 2022-11-18
msggen: Include schemas in source and add FundChannel{Start,Complete,Cancel}

cdecker:
`msggen` is intended as a general purpose tool allowing us to generate a number of formats from the abstract descriptions in our repo. In order for external tools to use these we need the package to contain these metadata files in its own subtree.

This PR introduces a script that takes configured methods and builds a schema.json file which contains all the methods, their requests and responses, as well as any future notification and hook schemas.

Along with these changes we also map the `fundchannel_start` / `fundchannel_complete` / `fundchannel_cancel` calls for use in external wallets.

Changelog-None


cdecker:
Superceded by #6918

PR #5722, opened 2022-11-18
Grpc docs

### LINKED ISSUE #5280: gRPC Docs
It would be nice to have docs on how to use the new gRPC server.

Some non-obvious this I have noticied:

No default port, only enabled if port is set
Not packaged with the release
Various `.pem` files are used (needed to read the code to find these lol)

cdecker:
I've gotten a number of requests on how the GRPC interface can be configured, and the what the client needs in order to connect to it. This is a minimal complete example of how to connect.

It also contains an example on how the mTLS identity for the node plugin can be changed in order to accept custom domains too.

Fixes #5280

### DIFF
--- a/plugins/grpc-plugin/README.md
+++ b/plugins/grpc-plugin/README.md
@@ -1,0 +1,144 @@
+# GRPC plugin for Core Lightning
+
+This plugin exposes the JSON-RPC interface through grpc over the
+network. It listens on a configurable port, authenticates clients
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5722/commits/20bc743840bf5d948efbf62d32a21a00ed233e31

PR #5723, opened 2022-11-19
CHANGELOG.md: order into a more user-first ordering.

rustyrussell:
And internal changes don't get CHANGELOG entries, so removed "- Added interactive transaction building routine ([#5287])".

We need to come up with some way of marking rust crate changes, as it's not immediately obvious what "- cln-plugin" means?

Maybe we switch from JSON-RPC to Control as a more general prefix?

Changelog-None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,51 +12,50 @@ TODO: Insert version codename, and username of the contributor that named the re
 
 ### Added
 
- - cli: new `--filter` parameter to reduce JSON output. ([#5681])
- - Documentation: `lightningd-rpc` manual page describes details of our JSON-RPC interface, including compatibility and filtering. ([#5681])
- - pyln: LightningRpc has new `reply_filter` context manager for reducing output of RPC commands. ([#5681])
- - JSON-RPC: `filter` object allows reduction of JSON response to (most) commands. ([#5681])
  - Reckless - a Core Lightning plugin manager ([#5647])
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/ElementsProject/lightning/pull/5723/commits/d9f5ca07b55d21d6cf6da969673f3f5a4f067a72

regarding the rust and python changes what about putting a filter in front of `Changelog-None(rust-grpc)` in this way from the git history, we can generate the changelog by scope. I had a tool that already do this but required some more love to be shipped with cln. But if you want I can spend some time testing it and we will see if it is possible to integrate it. I think should be trivial, I use it in all monorepo repositories and make the job

rustyrussell:
I think the simple prefix is correct, but our prefix choices need work. We're not just JSON any more: perhaps we should change this to simply API?

vincenzopalazzo:
I agree!

>I think the simple prefix is correct, but our prefix choices need work. We're not just JSON any more: perhaps we should change this to simply API?

Yes, I think ad a section regarding the API added and changed would be really interesting for the user that just does not care really about the internal.

I would also propose that we should generate our own changelog for the python client package and rust packages, because right now the changes are a little bit hide in the darkness of cln changes

cdecker:
ACK https://github.com/ElementsProject/lightning/pull/5723/commits/e937c82736a6dfa0ec658bfcb77c564a72ecccc6

Issue #5719, opened 2022-11-18
encrypted-hsm doesn't automatically create secret if it's missing


orbitalturtle:
### Issue and Steps to Reproduce

I'm using the "encrypted-hsm" config setting, but getting this error when I use it: "encrypted-hsm: Could not access 'hsm_secret': No such file or directory" In the docs here, it says "If there is no hsm_secret yet, lightningd will create a new encrypted secret." https://lightning.readthedocs.io/lightningd-config.5.html#lightning-daemon-options So I was expecting to not need the hsm_secret file to exist already?

I've been talking to @vincenzopalazzo who said he will take a look at this :)

vincenzopalazzo:
Sorry for the delay here!

You are right this is a bug, and in the next week I will patch it, for now I reproduced it

```
➜  lightning git:(master) ✗ ./lightningd/lightningd --testnet --disable-plugin bcli --plugin=/home/vincent/Github/clightning4j/btcli4j/btcli4j-gen.sh --encrypted-hsm 
2022-11-20T16:48:24.123Z INFO    plugin-manager: /home/vincent/Github/lightning/plugins/bcli: disabled via disable-plugin
./lightningd/lightningd: --encrypted-hsm: Could not access 'hsm_secret': No such file or directory
➜  lightning git:(master) ✗ 
```

From the docs we had the following assertion!

>If set, you will be prompted to enter a password used to encrypt the hsm_secret. Note that once you encrypt the hsm_secret this option will be mandatory for lightningd to start. If there is no hsm_secret yet, lightningd will create a new encrypted secret. If you have an unencrypted hsm_secret you want to encrypt on-disk, or vice versa, see lightning-hsmtool(8).

vincenzopalazzo:
A workaround is running lightningd without ` -encrypted-hsm`, lightning will read a un encrypted file, then you stop the daemon and rerun the same command with `-encrypted-hsm`


an example

```
➜  lightning git:(master) ✗ ./lightningd/lightningd --testnet --disable-plugin bcli --plugin=/home/vincent/Github/clightning4j/btcli4j/btcli4j-gen.sh                
2022-11-20T16:51:29.639Z INFO    plugin-manager: /home/vincent/Github/lightning/plugins/bcli: disabled via disable-plugin
2022-11-20T16:51:31.583Z INFO    lightningd: Creating database
2022-11-20T16:51:31.736Z UNUSUAL hsmd: HSM: created new hsm_secret file
[...]

PR #5725, opened 2022-11-20
lightningd: do not abort while parsing hsm pwd

vincenzopalazzo:
This is a mistake that I introduced while I implemented the lightningd custom error for hsmd.

In particular, when the command line parser check
if the file is encrypted make an additional check regarding the existence of the file.

This can be a not useful check, but due to the delicate nature of the hsm file, it is better to check if exist and if it doesn't exist an informative line inside the log is emitted that notifies the user that the file does not exist.

This log may return useful while debugging disaster
(that can happen) to understand that there is something strange (eg. the user moves the hsm file somewhere else).

Fixes https://github.com/ElementsProject/lightning/issues/5719

### DIFF
--- a/lightningd/options.c
+++ b/lightningd/options.c
@@ -516,9 +516,15 @@ static char *opt_set_hsm_password(struct lightningd *ld)
 	int is_encrypted;
 
         is_encrypted = is_hsm_secret_encrypted("hsm_secret");
+	/* While lightningd is performing the first initialization
+	 * this check is always true because the file does not exist.
+	 *
+	 * Maybe the is_hsm_secret_encrypted is performing a not useful
+	 * check at this stage, but the hsm is a delicate part,
+	 * so it is a good information to have inside the log. */
 	if (is_encrypted == -1)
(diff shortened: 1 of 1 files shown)


lightningd/options.c
```diff
@@ -516,9 +516,15 @@ static char *opt_set_hsm_password(struct lightningd *ld)
...
+	/* While lightningd is performing the first initialization
+	 * this check is always true because the file does not exist.
+	 *
+	 * Maybe the is_hsm_secret_encrypted is performing a not useful
+	 * check at this stage, but the hsm is a delicate part,
+	 * so it is a good information to have inside the log. */
 	if (is_encrypted == -1)
-		return tal_fmt(NULL, "Could not access 'hsm_secret': %s",
-			       strerror(errno));
+		log_info(ld->log, "'hsm_secret' not exist (%s)",
```
cdecker, line 526:
```suggestion
		log_info(ld->log, "'hsm_secret' does not exist (%s)",
```

orbitalturtle (commented):
Just confirming that testing this, the change worked for me, thanks for taking a look at this @vincenzopalazzo

PR #5727, opened 2022-11-21
Make modern string-based JSON ids opt in, deprecate numeric ones

rustyrussell:
This is not actually as bad as you'd think; we already handle arbitrary ids, it's just that we need to remember if this request id is a string or not.  Most of the code is plumbing, to only set the flag when the plugin says we can.

This fixes clboss!

I get a new rust warning, though, suspect my monkey coding there is wrong:

```
warning: field is never read: `nonnumericids`
  --> plugins/src/lib.rs:49:5
   |
49 |     nonnumericids: bool,
   |     ^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

```

### DIFF
--- a/plugins/src/lib.rs
+++ b/plugins/src/lib.rs
@@ -46,6 +46,7 @@ where
     rpcmethods: HashMap<String, RpcMethod<S>>,
     subscriptions: HashMap<String, Subscription<S>>,
     dynamic: bool,
+    nonnumericids: bool,
 }
 
 /// A plugin that has registered with the lightning daemon, and gotten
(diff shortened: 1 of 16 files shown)


plugins/src/lib.rs
```diff
@@ -46,6 +46,7 @@ where
     rpcmethods: HashMap<String, RpcMethod<S>>,
     subscriptions: HashMap<String, Subscription<S>>,
     dynamic: bool,
+    nonnumericids: bool,
```
cdecker, line 49: The Rust plugin handles the `id` completely internally, so this is not needed at all (also why the warning is being returned during compilation).

cdecker (approved):
We can likely drop the entire last commit, since `cln-plugin` handles the `id` internally.

ACK a25c5d14fe986b67178988e6ebb79610672cc829

